PR #3730, opened 2014-04-04
fixes #706 - possibility to choose cgeo main directory on external sd

### LINKED ISSUE #706: Move /sdcard/.cgeo to /sdcard/Android/data/cgeo.geocaching
I have a lot of directories on my sd-card and sometimes it's really annoying. I suggest to move ```/sdcard/.cgeo``` to ```/sdcard/Android/data/cgeo.geocaching```.

See:
http://developer.android.com/guide/topics/data/data-storage.html#AccessingExtFiles
http://developer.android.com/guide/topics/data/data-storage.html#ExternalCache

Maybe we have to add a new compatibility-class or we're using the "old" methods with a directory-name in any case:

>If you're using API Level 8 or greater, use ```getExternalFilesDir()``` to open a File that represents the external storage directory where you should save your files. [...]
>
>If you're using API Level 7 or lower, use ```getExternalStorageDirectory()```, to open a File representing the root of the external storage. You should then write your data in the following directory: ```/Android/data/<package_name>/files/```

So e.g. all the GC***-directories are in ```/sdcard/Android/data/cgeo.geocaching/cache/```.

The only problem, that I see, is:

> If the user's device is running API Level 8 or greater and they uninstall your application, this directory and all its contents will be deleted.

marco-dev:
Added the possibility to set the main c:geo directory e.g. to an external device. If DB file location is set to cgeo dir, database will be closed and initiated.

I added some @NonNull annotation to places where eclipse shows errors.


eslavko:
The MEGA event is probably ended, but still no news on the sky.

eslavko:
Hello...
Is this project abandoned? The Geocaching page was changed, login from that branch is not possible .. .. I uninstall that and upload default cgeo. the memory is full after some 20 caches loaded. So I badly need that one to store caches to external SD.

Issue #3950, opened 2014-06-05
Feedback needed: Setting Cache Directory


marco-dev:
**I need feedback:**

* Do you / we want the setting "set cache directory" or not?

**If yes:**
* Where should we add the check / warning dialog stating the cache directory path is not accessible any more? AbstractActivity, CgeoApplication? onCreate, onResume?
* Should we transfer the files ourself?
* Should we move or copy the files (user might not be able to access internal cgeo dir)?

Lineflyer:
I would like to put all of these discussions (#3950, #706 and others) into a new issue dealing with the overall file structure of c:geo. We surely have to find a solution for the offline cache (/.cgeo) but IMO as well include those random dirs we are using (e.g. /field-notes / GPX) into that discussion.

A lot of the discussion in those issues was about the right strategy and -even more important- about a migration strategy from the current to a possible new strategy.

Independently of whether there is someone willing to implement this, I would like to ask if you agree with starting from scratch and at first outlining the current situation and our target and afterwards define the ways towards it in a new issue.
The example from Locus posted in https://github.com/cgeo/cgeo/issues/3950#issuecomment-67784999 seems like a very good option also for c:geo.

Feedback?

PR #4034, opened 2014-07-01
fixes #706 - setting to choose offline cache dir

### LINKED ISSUE #706: Move /sdcard/.cgeo to /sdcard/Android/data/cgeo.geocaching

marco-dev:
The third approach now having a warning dialog if offlineCacheDirectory is remove or unreadable.
**Please only coding comments and no discussion now.**

Discussions can go here: #3950, #706

Former pr was #3921

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -78,6 +78,7 @@
     <string name="pref_fakekey_preference_restore">fakekey_preference_restore</string>
     <string name="pref_fakekey_preference_maintenance_directories">pref_fakekey_preference_maintenance_directories</string>
     <string name="pref_dbonsdcard">dbonsdcard</string>
+    <string name="pref_offline_cache_directory">offlineCacheDirectory</string>
     <string name="pref_debug">debug</string>
     <!-- preferences used internally -->
     <string name="pref_temp_twitter_token_secret">temp-token-secret</string>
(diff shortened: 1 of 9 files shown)


Lineflyer:
I would like to put all of these discussions (#3950, #706 and others) into a new issue dealing with the overall file structure of c:geo. We surely have to find a solution for the offline cache (/.cgeo) but IMO as well include those random dirs we are using (e.g. /field-notes / GPX) into that discussion.

A lot of the discussion in those issues was about the right strategy and -even more important- about a migration strategy from the current to a possible new strategy.

Independently of whether there is someone willing to implement this, I would like to ask if you agree with starting from scratch and at first outlining the current situation and our target and afterwards define the ways towards it in a new issue.
The example from Locus posted in https://github.com/cgeo/cgeo/issues/3950#issuecomment-67784999 seems like a very good option also for c:geo.

Feedback?

Issue #3675, opened 2014-03-23
Have "Use GPS only" setting session based


Lineflyer:
Current situation:
If the user enables "Use GPS only" in the compass menu, this setting is persistent. It affects not only the compass but also other directional views (e.g. pointer on live map).
There are quite a number of users on support mail with "compass problems", where it turns out this settings was active (either accidently or because it was once needed for a special purpose and forgotten afterwards).

I suggest to change this settings to be "session based" and defaults back to "Use Compass and GPS" (only if the device has a compass sensor of course).
The other option would be to have a clear indication somewhere, that the magnetic sensor is disabled.

Lineflyer:
@samueltardieu 
Please correct me if I am wrong: The active sensor setting has meanwhile been implemented transient/session based, so we can close this issue?

samueltardieu:
Yes, this has been made transient since dce71fa22dc6ebeeca837edf5db0fe68cf28ed0a.

Issue #432, opened 2011-09-13
Direction in compass-view


SammysHP:
I noticed a strange behavior in compass-view:

- select using of compass for direction
- while standing and walking the compass is used
- while driving (>20 km/h?) the gps is used (rotation of the phone doesn't change the direction of the arrow, on standing it jumps back to compass mode), the selections in the menu stays the same

I don't know if this is a "feature" of android or a bug in our access to the location-service.

Lineflyer:
Quite some time ago a manual selector has been added to the compass view to distinguish between GPS+Compass or only GPS.

PR #4224, opened 2014-08-26
Visual compass state and transient status

samueltardieu:
Discussions in #432 and #3675 as well as frequent reports from users show that:
- the user might not know or understand that the compass has been disabled and that only the GPS will be used for direction;
- the user might forget that they disabled the compass.

This PR, opened for testing, makes the *compass disabled* state transient (instead of saving it into the preferences), and switches to a always-visible toggle on the compass activity.

Please comment on feature and suggest style changes. It might be adapted easily for the `release` branch if needed. I have reused the `attribute_compass_yes` and `attribute_compass_no` strings so as not to require new translations if we decide to add it to the release, it will be changed later for specific strings after the release.

### DIFF
--- a/main/res/layout-land/compass_activity.xml
+++ b/main/res/layout-land/compass_activity.xml
@@ -36,6 +36,19 @@
             android:textColor="?text_color"
             android:textSize="14sp" />
 
+        <ToggleButton
+            android:id="@+id/use_compass"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:layout_alignParentLeft="false"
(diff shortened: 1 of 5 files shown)


Lineflyer:
Shall we close this one for now or is anyone (@rsudev) working on this?

samueltardieu:
Please let it open, I'll revisit this at some point (unless someone beats me to it) and I'd like to rescue the code first.

samueltardieu:
I'll merge this and open a new issue so that the look and feel can be fixed.

Issue #1382, opened 2012-04-07
Heading line


mucek4:
One thing, I always wanted to ask for, is this "heading line" on the map between the current position and the target position (cache, waypoint). I'd find that very helpful. In fact, this is one of the very few features of the official app, I'm missing in c:geo.

Lineflyer:
Recently implemented with  #4637

PR #4395, opened 2014-10-01
Fix #3590: Added compass and distance overlay to map views

### LINKED ISSUE #3590: Additional information on the map (View destination on the map)
I would hope if I ansteuere the destination using the map, nor the information such as the distance and number of degrees as an overlay is shown on the map.

As an example, I have appended a screenshot, where it is seen, as I imagine it.
http://abload.de/img/mapextdatajbr1i.png

kgraefe:
This fixes #3590 which is marked as a duplicate of #1382.

Screenshot


Lineflyer:
@kgraefe 

Reading your last comment I think that we can close this PR? 
Anyway another approach with a heading line would be really cool.

PR #4629, opened 2015-01-23
fixes #4072 -larger preview for larger display

### LINKED ISSUE #4072: Improve Tablet preview
As mentioned in #3967 the preview static map is too small on tablets and less in detail due to zoom factor.

marco-dev:
Just a small change to better fit preview image so it becomes larger.
Zoom level is still 15 - to be checked.


samueltardieu:
What does "to be checked"? Is it ready to be merged as-is, or do you want to wait?

marco-dev:
Ready to merge. Sorry for confusing.

marco-dev:
According to the screenshots in the issue we only need to get an image larger in width instead of a minWidth.

PR #4637, opened 2015-01-27
Add distance and direction overlayto the map view

kgraefe:
This adds the distance in a little box and the direction as a line between the current position and a destination to the map view (see attached screenshot).

cgeo

This addresses #3590 and #1382 and is basically the same as I proposed in #4395 but with the compass arrow replaced by the heading line.

Note that I did not refactor the overlay code as you said I should not invest too much time in that. So there is some duplicated code in the GoogleOverlay and MapsforgeOverlay classes.


Lineflyer:
How is this mode activated?
I guess it is active by default on Navigate - Map from a cache detail listing?

kgraefe:
Yes.

SammysHP:
Haven't read the code yet. Does this create two additional overlays to mapsforge? We're already really, really low on memory, please try to use the existing ones.

rsudev:
This looks definitely like a good start (I second the UI suggestions from SammysHP WRT line thickness and transparency), but in order to make this work for a considerable part of our users the functionality has to be merged into the position overlay, as the current mapsforge implementation allocates a full-screen bitmap for each overlay, which will lead to OOM on many older/smaller devices.
I am fully aware that this is not a clean approach from the software design perspective, but unfortunately inevitable due to the mentioned technical limitations...

samueltardieu:
The comment from @rsudev reinforces my feeling that modifying the maps as of today instead of rewriting them using Google Maps API v2 and a more recent Mapsforge is a mistake.

SammysHP:
+1

Lineflyer:
@samueltardieu 
I think I understand the problem in terms of whether to add something to the existing instead of first rewriting it to a new implementation. But the history shows, that such tasks are probably too big to be done in one step anyhow. So either it has to be split down into several work packages or we need to find someone willing to do the complete rewrite in one step. Right?

Anyhow regarding this feature:
IMHO we could implement that based on the current maps implementation (if the open remarks are solved) or we should otherwise declare a kind of "Feature stop" for the map until the map parts are rewritten. But last time we encouraged @kgraefe to continue, so we can now not just reject it due to a general problem.

samueltardieu:
@Lineflyer By accepting more functionalities into the current maps implementation, we are also discouraging people from rewriting it, as they will have to rewrite a lot to get equivalent functionalities, or users will be unhappy with it. The Maps v1 API will not be there forever, new developers cannot even get a key for it, and our Mapsforge implementation is well behind the current state of the library.

kgraefe:
I don't think that it would generate any extra effort to the guy who is rewriting the map implementation if my code sits within the `PositionAndScaleOverlay`. Furthermore it would eliminate most of the duplicated code I introduced.

SammysHP:
As lot of work is already done in this pull request I see no problem with merging it (after the transparency is added and the layer is merged with the one that is already existing).

kgraefe:
This is what it looks like now:
cgeo

Lineflyer:
The UI looks good for the start. Maybe we get more ideas for optimization once it can be tested in real life.

main/src/cgeo/geocaching/maps/PositionAndScaleOverlay.java
```diff
@@ -16,22 +19,40 @@
...
+                directionDrawer = new DirectionDrawer(bounds.center);
+                distanceDrawer = new DistanceDrawer(mapView, bounds.center);
+            }
+        }
     }
 
-    public void setCoordinates(Location coordinatesIn) {
+    public void setCoordinates(final Location coordinatesIn) {
         positionDrawer.setCoordinates(coordinatesIn);
+        if (directionDrawer != null) {
```
SammysHP: Assume that `distanceDrawer` is also `!= null`.

Well, in this case it should work, but there must be better ways to draw this stuff if necessary.

kgraefe: Yeah, I don't like it either. But earlier in this thread @samueltardieu said that I should not check every single variable for nullness when they are modified at the same time...

samueltardieu: If you don't like using one of the variables to guard the others, just introduce a boolean variable "drawerInitialized" or similar and use that as a guard.

kgraefe: In fact, I don't care very much. It's your software and I'm fine with doing things your way. ;) That was just the explanation why I did it the way I did.

Lineflyer:
How is the status of this PR?

I think @kgraefe updated it according to the comments meanwhile and @SammysHP stated, that it looks good for merging (transparency and merging of overlays has been done).

@kgraefe 
I suggest you squash your PR into a single commit and resubmit it as a PR then as this is common use here (as far as I understand) unless some of the developer team has more important remarks...

SammysHP:
Nope, these are quite clean commits that makes sense (in contrast to useless merge commits). So it would be ok to keep it as it is or rebase it (note: no new PR necessary, this one would be updated automatically).

I won't merge this as I'm not familiar with the maps code. Maybe @rsudev can have a short look at it and merge it then.

rsudev:
I didn't yet look in detail, but can try to do so maybe tomorrow.

main/src/cgeo/geocaching/maps/DirectionDrawer.java
```diff
@@ -0,0 +1,48 @@
...
+public class DirectionDrawer {
+    private Geopoint currentCoords;
+    private final Geopoint destinationCoords;
+    private final MapItemFactory mapItemFactory;
+
+    private Paint line = null;
+
+    public DirectionDrawer(final Geopoint coords) {
+        this.destinationCoords = coords;
+        this.mapItemFactory = Settings.getMapProvider().getMapItemFactory();
```
rsudev: This is a bit problematic, as the correct map provider is defined on start of a specific map activity instance. If the DirectionDrawer is created later (again), the settings might have been changed already by a different map activity instance. While this might not be very likely, we have seen this happening in some edge cases in other contexts already. So it would be necessary to hand down the MapItemFactory from the overlay or activity.
If you want I can also add this 'nonfunctional' change.

rsudev:
In general IMO this changeset looks good. Adjustements to the visual appearance can be easily done after merging.

main/src/cgeo/geocaching/maps/DirectionDrawer.java
kgraefe: That would be great. Thank you!

rsudev: I revisited this topic today while preparing the merge and changed my opinion. While strictly speaking problematic, the same pattern is applied in a number of similar places. Therefore I decided it is likely not worth the added overhead of wiring that through several levels in the responsibility chain at this time.

rsudev:
Merged and rebased locally, made the line thickness depend on the pixel density.
Thanks again for your contribution!

PR #4650, opened 2015-02-02
Export waypoints to Locus when they don't have coordinates

auricgoldfinger:
When exporting geocaches to Locus, some valuable information is lost.
- when waypoints are defined without coordinates, those waypoints aren't exported, causing the loss of the other defined information
- trackables aren't exported
- logs aren't exported.

This patch does export defined waypoints without coordinates, exports the name of trackables (no other information is currently available) and exports the logs.

### DIFF
--- a/main/src/cgeo/geocaching/apps/AbstractLocusApp.java
+++ b/main/src/cgeo/geocaching/apps/AbstractLocusApp.java
@@ -147,18 +147,23 @@ public abstract class AbstractLocusApp extends AbstractApp {
         if (withWaypoints && cache.hasWaypoints()) {
             pg.waypoints = new ArrayList<>();
             for (final Waypoint waypoint : cache.getWaypoints()) {
-                if (waypoint == null || waypoint.getCoords() == null) {
+                if (waypoint == null) {
                     continue;
                 }
+
                 final PointGeocachingDataWaypoint wp = new PointGeocachingDataWaypoint();
-                wp.code = waypoint.getGeocode();
+                wp.code = waypoint.getLookup();
                 wp.name = waypoint.getName();
+                wp.description = waypoint.getNote();
                 final String locusWpId = toLocusWaypoint(waypoint.getWaypointType());
                 if (locusWpId != null) {
                     wp.type = locusWpId;
                 }
-                wp.lat = waypoint.getCoords().getLatitude();
-                wp.lon = waypoint.getCoords().getLongitude();
+
+                if (waypoint.getCoords() != null) {
+                    wp.lat = waypoint.getCoords().getLatitude();
+                    wp.lon = waypoint.getCoords().getLongitude();
+                }
                 pg.waypoints.add(wp);
             }
         }
@@ -275,5 +280,4 @@ public abstract class AbstractLocusApp extends AbstractApp {
                 return null;
         }
     }
-
 }


auricgoldfinger:
Locus crashes because of a TransactionTooLargeException. My test set was only 14 large. 37 made Locus crash with me as well.

I changed the code so that a maximum of 10 logs is exported, but "10" is hardcoded. Please advice about that solution.

SammysHP:
Well, as this is not a cache export, I'd suggest to skip the logs.

Lineflyer:
Or maybe only the most recent one to keep the data amount minimized but still have some clues.

rsudev:
IIRC we had this issue also in the past and limited/batched the transfer to Locus for that reason. Perhaps you can get sime feedback from Locus (menion) on how to best handle this situation?

auricgoldfinger:
[TransactionTooLargeException](http://developer.android.com/reference/android/os/TransactionTooLargeException.html) is an Android exception so it's not really locus that crashes. Transaction can be only 1MB large so large chunks of logs can cause the exception. 

Solution is to not send too much data at once (limit logs? Batch sending?)
However, I'm convinced that 10 logs is enough to have an idea in Locus while caching. You can always go to c:geo to read all logs :-)

I'm more disappointed in not being able to get the necessary trackable information in Locus since that would help logging visit and trackable at once.

Bananeweizen:
The log text itself is HTML, including smileys and all other kinds of markup, if I remember right. Completely stripping that markup away (using our HtmlUtils.extractText()) and retaining only the pure text could also reduce the data amount.

auricgoldfinger:
I removed the export of trackables and logs since they make c:geo and Locus crash when you export a list of 100 or more caches. I will keep searching for a way to export these logs without causing crashes, but that'll be another pull request.

... which means that this pull request only handles the missing waypoints when they don't contain coordinates.

main/src/cgeo/geocaching/apps/AbstractLocusApp.java
```diff
@@ -147,18 +148,23 @@ private static Point getCachePoint(final Geocache cache, final boolean withWaypo
...
-                if (waypoint == null || waypoint.getCoords() == null) {
+                if (waypoint == null) {
                     continue;
                 }
+
                 final PointGeocachingDataWaypoint wp = new PointGeocachingDataWaypoint();
-                wp.code = waypoint.getGeocode();
+                wp.code = waypoint.getLookup();
                 wp.name = waypoint.getName();
+                wp.description = waypoint.getNote();
```
SammysHP: The description is not `getNote()` (or the name is misleading).

main/src/cgeo/geocaching/apps/AbstractLocusApp.java
```diff
@@ -276,4 +282,37 @@ private static String toLocusWaypoint(final WaypointType wt) {
         }
     }
 
+    private static int toLocusLogType(final LogType lt) {
```
SammysHP: This is not needed anymore.

SammysHP:
Please squash your commits.

auricgoldfinger:
Ok, done.

SammysHP:
Thanks for your contribution!

auricgoldfinger:
It's a pleasure :)

PR #4737, opened 2015-03-15
fix #4321: Sort by Event is missing in menu

### LINKED ISSUE #4321: Sort by Event is missing in menu
In the CacheList, I didn't found an entry to sort by EventDate.

I've found where to add the entry, but I'm wondering if this entry should always be visible even if there is no Event Cache in list. What if the list contains mixed types of caches?

### LINKED ISSUE #4335: Sorting selector sometimes not prefilled
2014.09.08.-NB:

Followup of #3914 and #4260 

Steps:
- Clean start c:geo (no sorting set for this session)
- Open list with non-event caches 
- Check sorting menu, sort by distance is active
- Change to a list of event caches
- Default sorting changes to event date (=hidden date) automatically
- Check sorting menu 
--> No entry is selected but should be "Hidden date"

Also related to #4321

kumy:
fix #4321: Sort by Event is missing in menu
fix #4335: Sorting selector sometimes not prefilled

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -293,6 +293,7 @@
     <string name="caches_sort_finds">Finds</string>
     <string name="caches_sort_state">State</string>
     <string name="caches_sort_storage">Date stored on device</string>
+    <string name="caches_sort_eventdate">Event date</string>
     <string name="caches_select_mode">Select mode</string>
     <string name="caches_select_mode_exit">Exit select mode</string>
     <string name="caches_select_invert">Invert selection</string>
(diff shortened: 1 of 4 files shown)


rsudev:
Too me it looks good, but I am not much into the sorting stuff. Any objections?
Otherwise I'll merge it tomorrow evening.

Lineflyer:
IMHO no separate sorting is required for event dates as currently the sorting "Date hidden"  does exactly that job. Or am I wrong?
Or was it implemented like suggested here https://github.com/cgeo/cgeo/issues/4321#issuecomment-54629186 ? Then it sounds good from users perspective.

kumy:
There is already an existing specific `EventDateComparator` class.
I can imagine the comparison tests aren't on the same fields as on `DateComparator`

samueltardieu:
From reading the source code, the difference lies in the way we sort events/caches hidden at the same date. For events, we look at the time as well, for caches, we favor closest ones.

samueltardieu:
Merged, thanks :)

Issue #1536, opened 2012-05-10
Static save button in edit waypoint


ludoo0d0a:
In edit waypoint screen, save should be static at bottom so that will be always visible/"clickable" on the screen.
I got a GalaxyS, I always have to scroll down to acces "save" button in edit waypoint screen.

Lineflyer:
Implicitly solved with #4738

PR #4738, opened 2015-03-15
Fixes #3443, Actionbar for edit waypoint

### LINKED ISSUE #3443: Waypoint improvements
Working with waypoints is more awkward than it should be. I propose the following changes to make the experience 'smoother':
- Remove the save button from edit waypoint, add save and cancel to 'action bar', make 'save' the default action on 'back'
- Swap short and long press in waypoint list. Short-press - edit, long-press - menu

rsudev:
- remove save button, add action bar items
- add a new notification system to notify cache details when the save is done

Implements #1536 as well implicitly.

The core feature is quite straight forward IMO, but I'd like some more eyes on the new underlying notification I had to introduce, to get the change information distributed.


rsudev:
On IRC a broadcast-based solution was suggest, so closing here.

Bananeweizen:
If it were still open, I would have remarked two things: comparison of objects should use equals method instead of equals operator, and abbreviations like "Bg" for background should be avoided. I didn't dive into the actual functionality as you said you will rework it anyway.

rsudev:
Thanks for the feedback! Not sure why the abbreviation 'Bg' entered my mind (because of ui?), normally I don't create names like this.
Regarding the comparison, that was in this case deliberate, as I wanted to explicitly check for instance-equality here. Probably should have commented that then.
In the reworked implementation which I right now prepare both will be gone anyway.

PR #4755, opened 2015-03-22
Fixes #4741, Search result filter

### LINKED ISSUE #4741: Search result filter
I can't use filter to searched results it always return me to saved cache instead of search results.

Steps to reproduce:
Open c:geo
Click search 
Type ex. Warszawa in address
Display results
Use filter

Result:
Filtered stored cache not searched cache

--- System information ---
Device: LG-D802 (g2_open_com, lge)
Android version: 4.4.2
Android build: KOT49I.D80220h
Cgeo version: 2015.03.16-NB-bab02fe
Google Play services: enabled
Low power mode: active
Compass capabilities: yes
Rotation sensor: present
Geomagnetic rotation sensor: present
Orientation sensor: present
Hide own/found: false
Map strategy: detailed
HW-acceleration: enabled (Disabled by default: no)
geocaching.com: logged in (Login OK) / Basic Member
System language: pl_PL
Installed cgeo plugins: none

rsudev:
- do not refresh online lists!

Someone with more insight in the listhandling should take a second look.
My local tests with online and offline lists were successful

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -1058,8 +1058,10 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             final int[] filterIndex = data.getIntArrayExtra(FilterActivity.EXTRA_FILTER_RESULT);
             setFilter(FilterActivity.getFilterFromPosition(filterIndex[0], filterIndex[1]));
         }
-
-        refreshCurrentList();
+        // refresh only offline lists
+        if (search == null) {
+            refreshCurrentList();
+        }
     }
 
     private String getDisplayName(final Uri uri) {


samueltardieu:
Have you tested with other lists such as history and nearest caches? Because another option would be to check the "type" field.

rsudev:
Before committing I checked searches (by address, nearest) and stored lists. Right now also history. So this seemingly works.
An explicit type specifier I didn't see when looking, but I can look again tomorrow.
But if you want you can also adapt my proposal roght away.

samueltardieu:
The `type` field is of type `CacheListType`, which should be clearer in the code that checking `search` (whose nullness is more of a side-effect). I cannot code and test right now unfortunately.

PR #4816, opened 2015-03-31
Fix #4783: Time selector for GK logs does not respect locale settings

### LINKED ISSUE #4783: Time selector for GK logs does not respect locale settings
2015.03.29-NB:
- Open the logging page for a geokret
- The time selector defaults to 12hour format (AM/PM) although my local setting should be 24hour format

kumy:


main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -611,6 +612,10 @@ public static boolean isRegisteredForGeokrety() {
         return getGeokretySecId() != null;
     }
 
+    public static boolean get24HourStyle() {
```
samueltardieu: A method named like that should not return a boolean. Or, rather, should be renamed, into something like `use24HoursStyle`.

Anyway, I think it should go away, see below.

main/src/cgeo/geocaching/ui/dialog/TimeDialog.java
```diff
@@ -45,6 +46,7 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
         picker.setCurrentHour(date.get(Calendar.HOUR_OF_DAY));
         picker.setCurrentMinute(date.get(Calendar.MINUTE));
         picker.setOnTimeChangedListener(new TimePickerListener());
+        picker.setIs24HourView(Settings.get24HourStyle());
```
samueltardieu: No need for this extra method, you can use:

```java
   picker.setIs24HourView(DateFormat.is24HourFormat(getActivity());
```

to refer to the context of the enclosing activity.

PR #4837, opened 2015-04-04
Fix #4833: Write note for GC trackable no longer possible without tracking code

### LINKED ISSUE #4833: Write note for GC trackable no longer possible without tracking code
2015.04.03-NB

- Open any GC trackable (for me it was a TB listed in a cache and I opened it from the inventory)
- Select log type NOTE
- Enter log text
- Try to send the log

Result:
Toast is displayed, that the tracking code is missing.

It seems the tracking code is needed for GKs but it is not needed for GC TB (and was working without before implmentation of GK support)

kumy:
Create a loggingManager.isTrackingCodeNeededToPostNote() function.
Show/hide the tracking code field depending on Trackable properties.

### DIFF
--- a/main/src/cgeo/geocaching/connector/trackable/AbstractTrackableLoggingManager.java
+++ b/main/src/cgeo/geocaching/connector/trackable/AbstractTrackableLoggingManager.java
@@ -26,5 +26,10 @@ public abstract class AbstractTrackableLoggingManager extends AsyncTaskLoader<Li
     public abstract boolean isRegistered();
 
     @Override
+    public boolean isTrackingCodeNeededToPostNote() {
+        return false;
+    }
+
+    @Override
     public abstract boolean postReady();
 }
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/connector/trackable/AbstractTrackableLoggingManager.java
```diff
@@ -26,5 +26,8 @@ public AbstractTrackableLoggingManager(final Context context) {
     public abstract boolean isRegistered();
 
     @Override
+    public abstract boolean isTrackingCodeNeededToPostNote();
```
Bananeweizen: Like I wrote in the other PR, an abstract class with an abstract method overriding an interface method is a bit confusing. I would either leave the abstract method away or (even better) provide a useful default implementation to avoid overriding this is every single concrete subclass.

Bananeweizen:
Looks generally okay to me, but I suggest moving the default implementation into the abstract connector.

kumy:
Is this what you suggested?
I hope it's better right now.

PR #4846, opened 2015-04-05
Fix #4778: Geokrety login UI should be easier

### LINKED ISSUE #4778: Geokrety login UI should be easier
Currently the user enters his login and password and additionally has to select "Retrieve API key".
This could be redesigned to make it look likt e.g. the authorization to Opencaching sites.
Just enter a username and password and select "Authorize" to complete it.

Not yet sure however about how this thing works.

kumy:


main/res/xml/preferences.xml
```diff
@@ -352,50 +352,28 @@
                 android:title="@string/init_geokrety" >
                 <PreferenceCategory android:title="@string/settings_settings" >
                     <CheckBoxPreference
-                        android:defaultValue="true"
```
Bananeweizen: You should separate different issues in different PRs, even if one of them contains only one single line. This line is about having GK not active by default and is completely unrelated to the remaining changes about authorization.

main/src/cgeo/geocaching/connector/trackable/GeokretyAuthorizationActivity.java
```diff
@@ -0,0 +1,77 @@
...
+
+import ch.boye.httpclientandroidlib.HttpResponse;
+
+import android.os.Bundle;
+
+import java.util.regex.Pattern;
+
+public class GeokretyAuthorizationActivity extends TokenAuthorizationActivity {
+
+    private static final Pattern PATTERN_IS_ERROR = Pattern.compile("error ([\\d]+)");
```
Bananeweizen: duplicated code, same constants defined in another activity

main/src/cgeo/geocaching/connector/trackable/TrackableConnector.java
```diff
@@ -19,6 +19,8 @@
  */
 public interface TrackableConnector {
 
+    public int getPreferenceActivity();
```
Bananeweizen: When extending interfaces, it is a good idea to add javadoc. We have so little javadoc (meaning we didn't do a good job at this in the past), and it is hard to understand for other developers what is meant with "isRegistered", "isActive", ... etc.

main/res/values/strings.xml
```diff
@@ -580,6 +583,7 @@
     <string name="settings_open_geokretymap_website">Open GeoKretyMap.org website</string>
     <string name="settings_settings">Settings</string>
     <string name="settings_information">Information</string>
+    <string name="settings_enhancement">Enhancement</string>
```
Bananeweizen: I'm not even sure what this is referring to, but I believe the term might be wrong, as I can't imagine this to appear anywhere in a settings screen. Is this referring to "Extensions", "Advanced settings", or something similar?

main/res/values/strings.xml
```diff
@@ -236,6 +236,7 @@
     <string name="info_stored_image">New image saved to:</string>
     <string name="info_storing_static_maps">Trying to store static maps</string>
     <string name="info_cache_saved">The cache has been stored locally</string>
+    <string name="err_trackable_log_not_anonymous">Logging trackables from c:geo cannot be anonymous. Please connect to the service.</string>
```
Bananeweizen: ...cannot be done anonymously.  or ...cannot be done without user name and password.

"Please connect to the service" is easier to understand when rephrased as "Please add your user name and password in the %s settings", with the service name inserted.

main/src/cgeo/geocaching/connector/trackable/AbstractTrackableConnector.java
```diff
@@ -18,6 +18,9 @@
 public abstract class AbstractTrackableConnector implements TrackableConnector {
 
     @Override
+    public abstract int getPreferenceActivity();
```
Bananeweizen: Overriding an interface method as an abstract method in a class doesn't make sense (and is reported as code inspection issue in IDEA at least). I suggest to make this a concrete method and to remove the all the "return 0;" implementations from the other connectors.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
```diff
@@ -0,0 +1,319 @@
+package cgeo.geocaching.activity;
```
Bananeweizen: What exactly is this class needed for? I thought that we can use our OAuthAuthorization activity for all kinds of OAuth authorization, Geokrety included. I must be missing something here.

main/res/xml/preferences.xml
kumy: I'll try to remember ;)

main/src/cgeo/geocaching/connector/trackable/GeokretyAuthorizationActivity.java
kumy: Are you sure?
This pattern is a specific check for the GK responses from server.
Those types of constant shouldn't be "local", as for strings?

main/src/cgeo/geocaching/connector/trackable/TrackableConnector.java
kumy: I'll have to learn javadoc too...

main/res/values/strings.xml
kumy: I vote +1 for "Advanced settings"

main/res/values/strings.xml
kumy: GK could logged anonymously directly on geokrety.org. But as we use the API, authentication is mandatory. Users used to log anonymously on GK website may be disturbed.

What about "Logging trackables from c:geo cannot be done anonymously. Please fill your credentials in the %s settings"?

main/src/cgeo/geocaching/connector/trackable/AbstractTrackableConnector.java
kumy: I'll check.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
kumy: As said in the related issue, GK service doesn't provide OAuth authentication.
I've tried to be as generic as possible. So next connector that need similar authentication system could be easily implemented. This code is largely inspired from your OAuth implementation.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
```diff
@@ -0,0 +1,319 @@
...
+            }
+            requestTokenDialog.show();
+            startButton.setEnabled(false);
+            startButton.setOnTouchListener(null);
+            startButton.setOnClickListener(null);
+
+            RxUtils.networkScheduler.createWorker().schedule(new Action0() {
+                @Override
+                public void call() {
+                    requestToken(usernameEditText.getText().toString(), passwordEditText.getText().toString());
```
samueltardieu: You are not supposed to access `usernameEditText.getText()` from the network thread. Capture the username and password into `final String` variables right before scheduling the action.

main/src/cgeo/geocaching/connector/trackable/GeokretyAuthorizationActivity.java
samueltardieu: I think too they are better as local variables. I think @Bananeweizen was due to an uncertainty about whether this authentication was derived form OAuth or was completely different, and the comment may be moot now that this has cleared up.

kumy: Thanks. I let it as is.

kumy:
Thanks for review.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
```diff
@@ -0,0 +1,322 @@
...
+            }
+            requestTokenDialog.show();
+            startButton.setEnabled(false);
+            startButton.setOnTouchListener(null);
+            startButton.setOnClickListener(null);
+
+            final String username = usernameEditText.getText().toString();
+            final String password = passwordEditText.getText().toString();
+
+                RxUtils.networkScheduler.createWorker().schedule(new Action0() {
```
samueltardieu: There seems to be an indentation issue, lines 192 and 194 should be indented the same way.

kumy:
Fixed the indent issue

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
```diff
@@ -0,0 +1,322 @@
...
+        startButton.setText(StringUtils.isBlank(getToken()) ? getAuthStart() : getAuthAgain());
+    }
+
+    @Override
+    public void onNewIntent(final Intent intent) {
+        setIntent(intent);
+    }
+
+    @Override
+    public void onResume() {
```
samueltardieu: This identical override is useless, please remove.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
```diff
@@ -0,0 +1,322 @@
...
+    public static final int AUTHENTICATED = 1;
+    private static final int ERROR_EXT_MSG = 2;
+
+    @NonNull private String host = StringUtils.EMPTY;
+    @NonNull private String pathToken = StringUtils.EMPTY;
+    @NonNull private String urlRegister = StringUtils.EMPTY;
+    @NonNull private String fieldUsername = StringUtils.EMPTY;
+    @NonNull private String fieldPassword = StringUtils.EMPTY;
+
+    private boolean https = false;
```
samueltardieu: Better use something that can be identified as `boolean`, such as `useHTTPS`. Nevertheless, why single out http/https and pass it as an extra, rather than getting it from the URL? And why decompose the URL into components at all?

kumy: I'll check it this tonight.
Note: code is inspired from AuthorizationActivity.java
The same remark could be done for https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java#L56 ? ;)

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
kumy: I agree.

main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
samueltardieu: Yes, the same remark can certainly be done for the code you cite. It should probably be cleaned up as well in the future.

kumy:
Removed the https parameter and define the full url for each case.
Code rebased on current master.

Issue #4822, opened 2015-04-01
ConnectorFactory.getTrackableConnector may return wrong connector


samueltardieu:
`ConnectorFactory.getTrackableConnector` returns the first connector that can handle a trackable, regardless of whether it does it or not.

For example, `GK8765` can be handled by both the GeoKrety and the TravelBug connectors. We should determine more precisely which one does handle the trackable, as is done for example in `TrackableActivity.loadTrackable` where all the matching connectors are queried.

kumy:
`ConnectorFactory.getTrackableConnector` could build a list, and request each connector if the Trackable exists.

But how should we handle the case where there are really two responses?

* return the 'most used' connector (TravelBug)
* ask user which one he would like to use/see
* prefer the connector which initials correspond

(I'm not confident on how to code the second proposition.)

rsudev:
I deem this as a very inlikely edge-case and would suugest to not put too much effort into it. Perhaps a toast or message that tells the user 'c:geo cannot handle this trackable' or the like.

Bananeweizen:
No user complained about this, although that situation exists for a long time. That's why I also don't see the necessity. I've run into this multiple times when implementing the above mentioned interface for the connectors. The 80/20 solution which is currently implemented: Have the connectors ordered by usage frequency. That will solve the issue for 99% of our users by returning a TB, in case both a TB and GK are possible.

And if someone really wants to tackle this issue, then I suggest not to do something automatically, but just to present a list selection to the user, where he needs to choose the wanted connector in case of multiple possible ones, and then we continue as of now.

rsudev:
I think the simple boolean ```canHandle``` doesn't fit trackables to well. There are some codes, a connector unambiguously knows are his (TBXXXX for GC, GKXXXX for geokret) and some, which might fit (the secret codes which do not follow any pattern besides a length of six in both cases (if I am not mistaken)).
So we could follow a two step approach (first ask for ```canHandle```, if no one answers for ```mightHandle```) and then query in parallel and present a selection to the user in case of more than one returning a result. What do you think?

Bananeweizen:
[...]

Issue #4779, opened 2015-03-28
Network traffic on access of offline stored cache for GK


Lineflyer:
With the introduction of Geokrety-support we currently have a network access on each loading of a cache from the database.
Shouldn't this be limited to the case where the cache is really loading from network or refreshed?

```
00:17:52.486 Debug cgeo 17545  [ModernAsyncTask #1] GeokretyConnector.searchTrackables: wpt=GC4ZEM5
00:17:52.486 Debug cgeo 17545  [ModernAsyncTask #1] GET http://geokretymap.org/export2.php?wpt=GC4ZEM5
00:17:52.596 Debug cgeo 17545  [main] Loading GC4ZEM5 Blauer See - Dolomitgestein (14) from DB
00:17:53.576 Debug cgeo 17545  [ModernAsyncTask #1] 200 (1149 ms) GET http://geokretymap.org/export2.php?wpt=GC4ZEM5
00:17:53.916 Debug cgeo 17545  [main] Loading GC4ZEM5 Blauer See - Dolomitgestein (14) from DB
```

Lineflyer:
It looks like the GK inventory is not updated on refresh or store but only on access of the cache.
That does not really fit into the current concept if I interpreted the log correctly.

kumy:
This was asked  on 24 Oct 2014.
See #4374 "still needed to be made asynchronous at least, as to not add another roundtrip to the load time of a cache."

I may have misunderstood what was said...

kumy:
I've misunderstood the requirements.
GK should be stored in Datastore. And loaded and merged async.

Observable should be controlled here https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/connector/gc/GCParser.java#L1891-1891

Lineflyer:
@kumy Does that mean you chose another implementation than desired and want to change it?

kumy:
@Lineflyer Store GK in Datastore was in the first implementation 6 month+ ago. It has been dropped after the first reviews as I was confused (or misunderstood) by the remarks... This part should be re-coded. BTW it will help for #4802.

For the second part, "move" the code that load generic trackable, I'm not sure that the GCParser class suggested is a good "destination" choice as it has to be generic to be usable by other connectors... Maybe it was suggested as an example of implementation... ;)

The actual code for loading Generic Trackable is there: 
https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/CacheDetailActivity.java#L339
[...]

PR #4859, opened 2015-04-08
Fix #4858: TravelBug doesn't have good icon in CacheDetailsActivity inventory

### LINKED ISSUE #4858: TravelBug doesn't have good icon in CacheDetailsActivity inventory
The icon for TB in the CacheDetailsActivity inventory is the default one.

It should be identical to the one shown when logging a cache.

kumy:
The list in CacheDetailsActivity is filled from the cache page. In this
page, there is only de guid information, no Trackable serial number.

When trackable are parsed in GCParser, they are stored in the Datastore.
In this parsing, Trackable brand is set.

Save Trackable brand in datastore.
So Travelbug will have the good logo instead of unknown.

### DIFF
--- a/main/src/cgeo/geocaching/DataStore.java
+++ b/main/src/cgeo/geocaching/DataStore.java
@@ -2,6 +2,7 @@ package cgeo.geocaching;
 
 import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.gc.Tile;
+import cgeo.geocaching.connector.trackable.TrackableBrand;
 import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.LoadFlags;
@@ -153,7 +154,7 @@ public class DataStore {
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static SQLiteDatabase database = null;
-    private static final int dbVersion = 68;
+    private static final int dbVersion = 69;
     public static final int customListIdOffset = 10;
     private static final @NonNull String dbName = "data";
     private static final @NonNull String dbTableCaches = "cg_caches";
@@ -299,7 +300,8 @@ public class DataStore {
             + "released long, "
             + "goal text, "
             + "description text, "
-            + "geocode text "
+            + "geocode text, "
+            + "brand integer "
             + "); ";
 
     private static final String dbCreateSearchDestinationHistory = ""
@@ -804,6 +806,15 @@ public class DataStore {
                             db.execSQL("alter table " + dbTableCaches + " add column logPasswordRequired integer default 0");
                         } catch (final Exception e) {
                             Log.e("Failed to upgrade to ver. 68", e);
+
+                        }
+                    }
+                    // Introduces trackableBrand
+                    if (oldVersion < 69) {
+                        try {
+                            db.execSQL("alter table " + dbTableTrackables + " add column brand integer default 0");
+                        } catch (final Exception e) {
+                            Log.e("Failed to upgrade to ver. 69", e);
 
                         }
                     }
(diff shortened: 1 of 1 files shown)


Lineflyer:
@kumy I don't fully understand. Can you post a screenshot of the inventory before and after this change to help me understand it?

kumy:
Yep,

before:
device-2015-04-14-224702


after:
screenshot2

SammysHP:
I'd prefer the first one. Why? Because it shows "Travelbug and Geocoin" – you can't distinguish between them. What we need is this first icon and an icon for Geokretys.

kumy:
@SammysHP If you prefer to use the first one for representing "Travelbug and Geocoin", shouldn't we have to choose/add another icon for "unknown brand".

Here are the actual icons for trackables:
https://github.com/cgeo/cgeo/blob/master/main/res/drawable-hdpi/trackable_all.png
https://github.com/cgeo/cgeo/blob/master/main/res/drawable-hdpi/trackable_geokrety.png
https://github.com/cgeo/cgeo/blob/master/main/res/drawable-hdpi/trackable_swaggie.png
https://github.com/cgeo/cgeo/blob/master/main/res/drawable-hdpi/trackable_travelbug.png

kumy:
I've hacked the existing icon a bit to build the TravelBug.
What about something like that?

Unknown:
trackable_all
TravelBug:
trackable_tb
GeoKrety:
trackable_geokrety
Swaggie:
trackable_swaggie

SammysHP:
What is a Swaggie?! I'd use a question mark for unknown, in your draft unknown and travelbug are rather similar.

kumy:
@SammysHP I didn't know about Swaggie before the last rebase of the GK patches...
Seems to be a type of trackables on http://geocaching.com.au/swaggies/

akkatracker:
@kumy yep that's exactly what they are. Funny how GCA's popping up here and there

Lineflyer:
@SammysHP 
I would currently not like to use a question mark for an unknown brand, because in our current way of loading all GC-trackables are unknown as long as you only look at the inventory of a cache.

@kumy
So "Unknown" means, that it is a GC trackable for sure, but only miss the info whether it is a TB, coin or one of hundreds of individual icons. Right?

kumy:
GC connector set the Brand while parsing the webpage.
However, "Unknown" are displayed while Trackables are loaded from Datastore as Brand is not stored in DB.

Actually, only TravelBugs are stored in Datastore (I have to check what's the code for GK loaded from OC connectors). The main goal of this PR is to store Brand in DB so brand will be displayed directly.

In the future generic Trackables may/should also be stored in DB, see #4779. All Trackables will be treated as "unknown" (another method could be to "find" connector from trackable id to retrieve the brand but we may hit #4822).

Note, if PR is merged, users will see "unknown" for Trackables already present in datastore, until DB is refreshed...

(Do we have a designer in the team? Maybe the discussion about icons should be done in another) thread...

Lineflyer:
I am a little confused right now.
I do see the current "Unkown" icon in each GC cache inventory listing..however sometimes I see the TB icon instead. Not sure under which conditions this difference is there?!

What I want to say:
If the standard GC-only user opens a GC cache it would be OK and correct if he would see this icon IMO:
trackable_tb

When the user activates GK he will anyway see that GKs are displayed different already now.

kumy:
Can't understand how you may have seen TB icon as Cache inventory is always loaded from Datastore and Brand not stored/loaded in DS:
https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/CacheDetailActivity.java#L716-716
https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/SearchResult.java#L243-243

However, I've found another less "intrusive" approach.
This PR will be closed as "won't merge".

The debate about icon is still open...

kumy:
* Fixed by 216b55693
* #4946 opened for icon discussion

PR #4863, opened 2015-04-09
Fix #4862: Wrong strings used in new magnetic compass button

### LINKED ISSUE #4862: Wrong strings used in new magnetic compass button
Looking at 4723c5ad2366e996861bfcab162a7e4fcfcdc6ec wrong strings seem to be used for the new button.
We misuse the stings `attribute_compass_yes` and `attribute_compass_no`which are intended for the text form of cache attributes.
I just changed them to match the new function not recognizing, that they are used also in other places.
I will revert that commit in a minute.

We should either use the original strings `use_gps` and `use_compass` or create new strings for this button.

Lineflyer:


SammysHP:
Huh? We had these strings already. Why are they gone?

Lineflyer:
Because of 4723c5a

PR #4876, opened 2015-04-12
Fix #4854: Direct users to settings if they try to log a GK without login

### LINKED ISSUE #4854: Direct users to settings if they try to log a GK without login
I attended an opencaching event today and presented the upcoming GK feature.
A big challenge for many users was, that they did not know what to do when they tried to log a GK for the first time and ended up in the info window telling that anonymous logging is not possible.

Instead of having just OK as an option we should have a second button to open the geokrety settings directly in that info window.

Furthermore we should slightly adapt the text not only to say "Please register" but "Please register or login with your existing account from geokrety.org".

kumy:
Rework of the message given to the user.
Remove the toast. Display a Yes/No Dialog.


Lineflyer:
Any complaints to this PR or can it be merged?

PR #4959, opened 2015-05-21
Fix #3015 - Mixed use of send2c:geo and send2cgeo

### LINKED ISSUE #3015: Mixed usage of "send2c:geo" and "send2cgeo"
As found in review of #3011 we currently have a mixed usage of these two terms in our strings and also on the website.

How is our tool officially called?

My preference goes to "send2cgeo", but I can't tell why...just subjective

Lineflyer:


Lineflyer:
Will create a new PR towards 'release'

Lineflyer:
No, 'master' is the better choice.

PR #4974, opened 2015-05-25
Convert Serializables Filters to Parcellable

kumy:
As discussed on 6d82792 Parcellable is a better way than Serialize.
Convert Filters to Parcellable and remove all added Serialize stuff
from last commits.
In OriginFilter we wouldn't Parcell the connectors, so we store
only the Connector name.

### DIFF
--- a/main/src/cgeo/geocaching/CgeoApplication.java
+++ b/main/src/cgeo/geocaching/CgeoApplication.java
@@ -20,13 +20,11 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.view.ViewConfiguration;
 
-import java.io.Serializable;
 import java.lang.reflect.Field;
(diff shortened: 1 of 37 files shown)


kumy:
@rsudev here is the diff using int for saving Enum. See SizeFilter and TypeFilter

I understand that the whole thing would be difficult to read, so I've squashed f30b124 1c47dfe ce8b30c 6d82792 5357eec 110cee1  into only one commit -> https://github.com/kumy/cgeo/commit/f709c2d3fd7b0ecb05ac0c760e977a56a0c26110 (only for reading ; not to merge...)

Thanks for review

kumy:
PR updated, constructors are now `protected`. Manually tested.
@rsudev ok to merge?
Thanks

PR #4992, opened 2015-05-30
Fix #4987: Trackables might not be removed from a cache on refresh

### LINKED ISSUE #4987: Trackables might not be removed from a cache on refresh
2015.05.29-RC:

From IRC:
```
00:43:29OneEyed: if (hasBrand && removeTrackable) {
00:43:31OneEyed: iterator.remove();
00:43:33OneEyed: }
00:43:56OneEyed: Former trackables will never be removed if they don't have a brand originally
...
00:45:26OneEyed: kumy: am I correct that if there is a trackable of brand X, you refresh, and there are now zero trackables of brand X, it will never be removed?
00:45:42OneEyed: Geocache.java:1020
```

kumy:
Unit tests passed.
Do you imagine other test case?

### DIFF
--- a/main/src/cgeo/geocaching/Geocache.java
+++ b/main/src/cgeo/geocaching/Geocache.java
@@ -12,6 +12,7 @@ import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.connector.gc.GCConstants;
 import cgeo.geocaching.connector.gc.Tile;
 import cgeo.geocaching.connector.gc.UncertainProperty;
+import cgeo.geocaching.connector.trackable.TrackableBrand;
 import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.LoadFlags;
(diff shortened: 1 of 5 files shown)


kumy:
@samueltardieu may you give me your comments please :)

main/src/cgeo/geocaching/Geocache.java
```diff
@@ -1000,32 +1000,37 @@ public void setInventory(final List<Trackable> newInventory) {
...
+        while (iterator.hasNext()) {
+            final Trackable trackable = iterator.next();
+            boolean removeTrackable = true;
+            boolean hasBrand = false;
+            for (final Trackable newTrackable : newTrackables) {
+                if (trackable.getBrand() == newTrackable.getBrand()) {
+                    hasBrand = true;
+                    if (trackable.getUniqueID().equals(newTrackable.getUniqueID())) {
+                        removeTrackable = false;
+                        break;
```
samueltardieu: Would not it be simpler to just start from a copy of `newTrackables`, and all all items in `inventory` whose brand is not in `processedBrands` (which could be made a set)?

kumy: You're right.
Inverse the logic and things seems more simple ;)

kumy:
PR updated

kumy:
oups something is wrong :(

kumy:
Fixed NPE and add unit test.
It's better now

Lineflyer:
@samueltardieu Ready to merge this?

samueltardieu:
I think it is. I can't do it though.

PR #5039, opened 2015-06-17
Fix #5038 - Request large heap

### LINKED ISSUE #5038: Try to minimize OOM problems by requesting larger heapsize
As discussed on IRC:
I would like to test if requesting a larger heapSize in our manifest could help minimizing the OOM problems (caused by large bitmaps and mapsforge library) for the time being.
I am aware it is no solution but still maybe helpful until we get more memory optimizations (like e.g. a new mapsforge library).

Reference:
http://developer.android.com/reference/androi/R.styleable.html#AndroidManifestApplication_largeHeap

I suggest to change it on `master` and do some comparisons on some devices.

Lineflyer:
Experimental change to test if it has any advantage/disadvantage on memory intense scenarios in c:geo.

Could someone check if the syntax is OK or reply if anyone does see a problem with it?

I will test and provide feedback based on the nightlies so we can decide to keep or revert.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -41,7 +41,8 @@
         android:label="@string/app_name"
         android:theme="@style/cgeo"
         android:hardwareAccelerated="false"
-        android:supportsRtl="true">
+        android:supportsRtl="true"
+        android:largeHeap="true">
         <uses-library
             android:name="com.google.android.maps"
             android:required="false" />


rsudev:
From inspection this probably will not compile as thos attribute conflicts with our minSdk AFAIK

Lineflyer:
Good question.
I have read that it will only work out on Honeycomb+... but I did not read a hard API limit though.
Can you try to compile?

mucek4:
It is possible to compile different versions. Just more work for deployment.

rsudev:
Contrary to my initial short research it seems not to fail the build, therefore I'll merge it for testing. Please cross-check specifically if it creates problems on 2.x devices (shouldn't have any effect there).

Issue #4873, opened 2015-04-10
Allow upload of personal note to OC server


Lineflyer:
This seems to be supported with the OKAPI and should be implemented identically to the gc.com feature:
http://www.opencaching.de/okapi/services/caches/save_personal_notes.html

Natureshadow:
Issue #2930 asked for and implemented download of the personal note and the personal coordinates. This feature request compeltes this work. Please also make changing the personal coordinates work, as for GC.com.

SammysHP:
> This feature request compeltes this work.

That's why we have this issue.

> Please also make changing the personal coordinates work, as for GC.com.

Do you mean upload of corrected coordinates? That is another issues.

rsudev:
> Do you mean upload of corrected coordinates? That is another issues.
Which currently seems not to be supported by the OKAPI, unfortunately.

madankumarbaskaran:
Can I work on this request (Upload of personal note to OC server)?

SammysHP:
Sure!

madankumarbaskaran:
For GC.com, personal notes are enabled for premium users. I believe for opencaching.de does not have basic or premium users, so upload should be enabled to all the users of opencaching.de? Please comment.

SammysHP:
Yes.

rsudev:
@madankumarbaskaran Furthermore does this not only affect opencaching.de, but all OKAPI-enabled sites which support the latest OKAPI version. IN the current code there are two levels of OKAPI-support defined, and the sites initialize the connector with the appropriate level. So for an implementation it is necessary to check:
* which sites currently support which version (resp. which feature-set, the exact version is not so relevant)
* how many support levels do we need in the future (are two still sufficient, because all 'current' sites also support the new feature) or do we need more
* report this through the capabilities in the connector and enable the upload button based on that only.

madankumarbaskaran:
@rsudev two levels of OKAPI-support, do you mean authentication level for OKAPI?. I am beginner to Android and CGEO, please elaborate on the above requirement. Appreciate your help.

rsudev:
[...]

PR #5076, opened 2015-07-05
Upload Personal note for opencaching (Fixes the issue #4873)

madankumarbaskaran:
Allow upload of personal note to OC server #4873 is fixed. Please review my code and share your suggestions.

### DIFF
--- a/main/res/values-fr/strings.xml
+++ b/main/res/values-fr/strings.xml
@@ -86,7 +86,7 @@
   <string name="log_post_geocode_missing">Géocode manquant…</string>
   <string name="log_date_future_not_allowed">Les visites dans le futur sont interdites.</string>
   <string name="log_add">Ajouter</string>
-  <string name="log_repeat">Recopier la dernière entrée</string>
+  <string name="log_repeat">Recopier la dernière visite</string>
   <string name="log_no_rating">Aucune évaluation</string>
   <string name="log_stars_1_description">faible</string>
   <string name="log_stars_15_description">Plutôt faible</string>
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
```diff
@@ -6,6 +6,7 @@
 import cgeo.geocaching.network.Parameters;
 import cgeo.geocaching.utils.CancellableHandler;
 import cgeo.geocaching.utils.CryptUtils;
+import cgeo.geocaching.utils.Log;
```
rsudev: If I am not mistaken, this import is not necessary

main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
```diff
@@ -108,4 +109,24 @@ public int getTokenSecretPrefKeyId() {
     public boolean isSearchForMyCaches(final String username) {
         return false;
     }
+
+    public boolean supportsPersonalNote() {
+        if (this.getApiSupport() == ApiSupport.current) {
```
rsudev: In addition to the Api-level, the user also needs to be authenticated on ```level3``` in order to have the possibility to upload a note (i.e. he needs to be logged in).
While in the gc.com version, this is assumed (you really cannot do anything useful without a login) this is not necessary the case for opencaching. Even without authentication you can download full cache details and go caching.
I suggest to move this change to the ```OCapiLiveConnector``` which handles level3-authenticated communication.

rsudev:
Sorry that it took me so long time to review your contribution.
I did some small remarks, overall it looks fine to me.

PR #5120, opened 2015-07-30
Removed color profile from PNG with warnings

ecornely:
I just used mogrify.exe -strip to remove color profiles from those PNG to remove the warning


Bananeweizen:
Can you give a short explanation of what warning that is? Somewhere in the build process or where can we see it? Just so I know what to watch for next time...

SammysHP:
In the big chunk of image processing output:

```
   [crunch] Processing image to cache: /home/sammyshp/development/eclipse/git/cgeo/main/res/drawable-xhdpi/ic_menu_start_conversation.png => /home/sammyshp/development/eclipse/git/cgeo/main/bin/res/drawable-xhdpi/ic_menu_start_conversation.png
   [crunch]   (processed image to cache entry /home/sammyshp/development/eclipse/git/cgeo/main/bin/res/drawable-xhdpi/ic_menu_start_conversation.png: 0% size of source)
```

ecornely:
As I don't know yet how to code in AndroidStudio and I'm just discovering the project I tried to fix some simple warnings.
I will propose an other pull request with Java Generics problems I solved but anyway for these png, I see the same warnings on images coming from the appcompat-v7-20.0.0.aar.

Can somebody tell me what is done during the build with that appcompat archive and what kind of process on the png is giving the warnings ?

I don't like warnings and wish we could remove all of them so that when a new one arrive we can directly see a potential problem and fix it. If we should consider this warning is not important, there's probably a way to configure the build tool so that this warning is not outputted. Or maybe should we pre/post process the png to remove those iCC profiles...

At least the png that were under git are now without iCC profiles in my trunk removing a few warnings.

SammysHP:
It's the PNG "optimization" process that complains about unknown color profile (I wonder that it respects color profiles at all). We can ignore that warning, the output won't change. Also Google ignores it, see appcompat.

Bananeweizen:
Thanks Eric, I appreciate the explanation and your cleanup work (as I also invest a lot of effort into making everything work without warnings). For appcompat it depends on the build tool: We have an Ant based build, which uses a _copy_ of appcompat in our own repo. For the gradle build I'm not sure if we use the copy in our repo, or if we just depend on the gradle dependency resolution (I don't use gradle).

PR #5152, opened 2015-08-25
Refactor: use Image class instead of local variables.

kumy:
Do you agree with such update?

### DIFF
--- a/main/src/cgeo/geocaching/Image.java
+++ b/main/src/cgeo/geocaching/Image.java
@@ -5,6 +5,9 @@ import cgeo.geocaching.utils.FileUtils;
 import cgeo.geocaching.utils.Log;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
 
 import android.app.Activity;
 import android.content.ActivityNotFoundException;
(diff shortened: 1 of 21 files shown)


main/src/cgeo/geocaching/Image.java
```diff
@@ -15,27 +17,92 @@
...
 
+/**
+ * Represent an Image along with Title and Description.
+ * Mostly used for representing image in/from logs.
+ */
 public class Image implements Parcelable {
-    private final String url;
-    private final String title;
-    private final String description;
+    @NonNull private Uri uri;
```
samueltardieu: Can't this field (and the two other ones) be `final` as well?

samueltardieu:
Except for the note on `final` fields, there is nothing shocking after a cursory glance.

kumy:
Ok, thanks. I'll update then merge.

main/src/cgeo/geocaching/Image.java
kumy: @samueltardieu it's not possible because of setter L210-L252.
Still ok to merge?

kumy: It however cause warnings in `equals()` and `hashCode()`. Any idea on how to solve this? (chicken egg)

samueltardieu: I really *hate* mutable fields. Is `setUrl()` really unavoidable?

kumy: For references http://programmers.stackexchange.com/questions/151733/if-immutable-objects-are-good-why-do-people-keep-creating-mutable-objects

If I understand well the concept, I should create new objects with previous data, and only change the necessary fields in constructor?

Is some sort of function in `Image` class called `ImageBuilder(???)` (or whatever) is the way to go? maybe that's what you say here "we should use mutable builders to build immutable data structures"

samueltardieu: Good link. The approved answer to this Stack Exchange question summarizes my point very well:

> Apart from all these legitimate reasons, alas, the most probable cause why people keep creating mutable objects is inertia of mind, a.k.a. resistance to change. Note that most developers of today have been trained well before immutability (and the containing paradigm, functional programming) became "trendy" in their sphere of influence, and don't keep their knowledge up to date about new tools and methods of our trade - in fact, many of us humans positively resist new ideas and processes. "I have been programming like this for nn years and I don't care about the latest stupid fads!"

Ideally, we should have a static `Image.builder()` create an mutable object, with setters and a `build()` method which then creates an immutable `Image`. Unfortunately, doing so is a very non-interesting-to-code task, but leads to much better code. Once you have an immutable object, coding `equals()` and `hashCode()` is trivial and safe (the fields never change, so the equality results and the hash-code don't either, which makes it safe to include as keys in a map for example).

samueltardieu: (note: I may be influenced by my strong Erlang and Haskell backgrounds where a variable can be assigned only once and never change afterwards)

kumy: >  we should have a static Image.builder() create an mutable object, with setters

is `object` an `ImageBuilder` class in this context?

So it could be ?
```
class ImageBuilder {
    @NonNull private Uri uri;
    @Nullable private String title;
    @Nullable private String description;

    public Image build() {...};
    public void setTitle() {...};
    [then only setters...]
}

class Image {
    @NonNull final private Uri uri;
    @Nullable final private String title;
    @Nullable final private String description;

    public ImageBuilder builder() {...};
    [then current functions...]
}
```
May I put the two classes in the same file?

samueltardieu: Yes: `builder()` would have to be `static`, and the `Builder` class may be a `static` inner class of `Image` (that would be a `Image.Builder` class then.

kumy: Is a `Image.Builer.buildFrom(final Image image)` a good idea as a two lines replacement of actual setters?

The inner class don't have access to privates variables so getters will be necessary for all `Image` privates variables.
If this is ok, I imagine some class will have also protected variables not exposed via getters, so we can't copy the original `Image` entirely. That's a different comportement from just a setter in `Image`.

samueltardieu: Note that I don't insist on having final variables be `private`with a getter: a `public final` field is as safe as a `private final` one with a getter, since it can't be modified.

kumy: Ok for `public`/`private` variables. And what about `protected`? Is everything `public`?

samueltardieu: Do we really need `protected` variables? You can make things semi-public by not qualifying them at all, this will be limited to the current package, which includes the inner classes of course.

kumy: The questions were more for a general pattern, not for this specific case.

kumy:
@samueltardieu PR updated, using Factory pattern.

main/src/cgeo/geocaching/Image.java
```diff
@@ -117,6 +364,68 @@ public boolean isLocalFile() {
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        final Image image = (Image) o;
+
+        if (!uri.equals(image.uri)) {
+            return false;
+        }
+        if (title != null ? !title.equals(image.title) : image.title != null) {
```
samueltardieu: You should use `StringUtils.equals()` for those comparisons.

main/src/cgeo/geocaching/Image.java
```diff
@@ -117,6 +364,68 @@ public boolean isLocalFile() {
...
+            return false;
+        }
+        return !(description != null ? !description.equals(image.description) : image.description != null);
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = uri.hashCode();
```
samueltardieu: Have you looked at class `HashCodeBuilder` from the commons lang library (that we use)? It helps you building a hashcode, and its `append()` method supports `null` so it does not require any explicit testing. It should let you do something like (untested):

```java
   return new HashCodeBuilder().append(uri).append(title).append(description).build();
```

kumy: (Those two functions were auto-generated by Idea.)
Code updated, but untested.

kumy:
@samueltardieu Unit tests ok with the latest changes. Go for a merge?

samueltardieu:
Sure!

samueltardieu:
Wait, why do you call the builder just to call build immediately after? Can't you use the Image constructor?

kumy:
oh, too late, merged. :/

Where?

samueltardieu:
I mean, the builder is useful when you have to mutate field, or build the object bit by bit. If you have all the arguments to give to the constructor, just give them, that will create an immutable object much more easily.

samueltardieu:
> Where?

Everywhere you call `new Image.Builder(arg1, arg2).build()`, can't you call `new Image(arg1, arg2)`?

kumy:
You mean move most constructors back in Image.

samueltardieu:
I hadn't seen you had moved them in the builder. Usually, the builder takes no argument, and has a set of  `Builder setXXX(final XXX xxx)` methods that let you chain them easily. You don't need to move the constructors back in `Image`, but you could have `new Image.Builder().setUrl(…).setTitle(…).build()`. But that's not a big deal.

kumy:
Sorry, I start to feel lost...

Which case should be used?
Ex: from Datastore L1966.
```
a)
log.addLogImage(new Image.Builder().setUrl(cursor.getString(10)).setTitle(cursor.getString(9)).build());

b)
log.addLogImage(new Image(cursor.getString(10), cursor.getString(9)));

c)
log.addLogImage(new Image.Builder(cursor.getString(10), cursor.getString(9)).build());

d)
?
```

samueltardieu:
If you want to add the constructors back, (b). If you would prefer not to, (a), which is a bit more verbose. In any case, both are better than (c), because here the builder here brings nothing (none of its mutability capabilities are used).

kumy:
If I already have an `Image` object, I just need to update the description. Here I would prefer (b). I can't feel the real benefit of the `Builder` class without (c).
 ```
a)
        final Image.Builder imageBuilder = new Image.Builder();
        imageBuilder.setUrl(image.uri);
        imageBuilder.setTitle(image.title);
        imageBuilder.setDescription(descriptionView.getText().toString());
        image = imageBuilder.build();

b)
        image = new Image(image.uri, image.title, descriptionView.getText().toString());

c)
        final Image.Builder imageBuilder = new Image.Builder().setFromImage(image);
        imageBuilder.setDescription(descriptionView.getText().toString());
        image = imageBuilder.build();

 ```

What's the name of this design pattern? I'll read some documentation. Or maybe you have a good link.

Sorry to make you lose your time.

kumy:
hum https://en.wikipedia.org/wiki/Builder_pattern


```
Construct a CarBuilder called carBuilder
carBuilder.setSeats(2)
carBuilder.setSportsCar()
carBuilder.setTripComputer()
carBuilder.unsetGPS()
car := carBuilder.getResult()

Of course one could dispense with Builder and just do this:

car = new Car();
car.seats = 2;
car.type = CarType.SportsCar;
car.setTripComputer();
car.unsetGPS();
car.isValid();
```

> [...] the Builder pattern is more than just a means to limit constructor proliferation. It removes what could be a complex building process from being the responsibility of the actual user of the object that is built. It also allows for inserting new implementations of how an object is built without disturbing the client code.

So "good" implementation would be (a) only (in the last example I've given).

kumy:
@samueltardieu what about https://github.com/kumy/cgeo/commit/9d7f1dd9bbfe1b18149ad4b77f3e467341840398?

samueltardieu:
That's certainly much more pleasant to read :)

kumy:
I'll push.

main/src/cgeo/geocaching/Image.java
kumy: @samueltardieu The angular stone I was looking for in the last questions was [`Image.buildUpon()`](https://github.com/cgeo/cgeo/blob/9c0d069de29e876405baf9c43f600ef2369fa23a/main/src/cgeo/geocaching/Image.java#L195-L206)

Found that pattern while studding [`android.net.Uri`](http://developer.android.com/reference/android/net/Uri.html#buildUpon%28%29)

rsudev: Just a small off-topic note: You mean 'pierre angulaire' here? I think this should rather be a 'cornerstone' then.

kumy: @rsudev Thanks ;)

PR #5157, opened 2015-08-28
navigation: add entry for google maps transit mode

lewurm:
This is more like a feature request, as I didn't manage to setup eclipse properly to build this thing. Can someone with a proper setup have a look?

I want to add another entry in the navigation menu that opens google maps finding a route with public transport. It should be rather straight forward I think.

Another question: How are translations managed? Do I need to add an entry for each language?

Thanks!


samueltardieu:
The code looks and compiles fine (translations are added later by translaters using Crowdin). However, it doesn't switch to transit mode on Google Maps on my Nexus 5 (Android 5.1.1).

However, it looks like this is not specific to your change: every mode triggers the same applications (Google Maps & CityMapper) and seems to ignore the mode.

samueltardieu:
It looks like transit mode is "r", not "t" which means toll-free car mode. I'll update it.

samueltardieu:
Thanks for your contribution.

lewurm:
awesome, thank you!

PR #5168, opened 2015-08-30
Support links encoded in some GeoKrety QRCode

kumy:
@samueltardieu could you please check this when you have time? I used Rx but I'm unsure if I use it right. Thanks

### DIFF
--- a/main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
+++ b/main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
@@ -13,6 +13,7 @@ import cgeo.geocaching.network.Network;
 import cgeo.geocaching.network.Parameters;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.RxUtils;
 import cgeo.geocaching.utils.Version;
 
 import org.apache.commons.collections4.CollectionUtils;
@@ -23,6 +24,7 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.xml.sax.InputSource;
 
 import rx.Observable;
+import rx.functions.Func0;
 import rx.functions.Func1;
 
 import android.content.Context;
@@ -37,6 +39,7 @@ import java.util.regex.Pattern;
 public class GeokretyConnector extends AbstractTrackableConnector {
 
     private static final Pattern PATTERN_GK_CODE = Pattern.compile("GK[0-9A-F]{4,}");
+    private static final Pattern PATTERN_GK_CODE_EXTENDED = Pattern.compile("(GK[0-9A-F]{4,})|([0-9A-Z]{6})");
     private static final String URL = "http://geokrety.org";
     private static final String URLPROXY = "http://api.geokretymap.org";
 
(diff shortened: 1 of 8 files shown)


kumy:
PR updated
Is it too late to put that in `release`?

samueltardieu:
I won't have the time to reread it carefully enough for it to be put in `release` as far as I am concerned.

Lineflyer:
`master` please as we are now in a close to release status.

kumy:
@samueltardieu do have time to reread it?
@Bananeweizen, @rsudev Anyone else would like to review?

samueltardieu:
I didn't plan to reread it. Is it needed?

main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
```diff
@@ -191,6 +217,41 @@ public String getTrackableCodeFromUrl(@NonNull final String url) {
...
+    /**
+     * Lookup Trackable Geocode from Tracking Code.
+     *
+     * @param trackingCode
+     *          the Trackable Tracking Code to lookup
+     * @return
+     *          the Trackable Geocode
+     */
+    @Nullable
+    public static String getGeocodeFromTrackingCode(final String trackingCode) {
```
samueltardieu: It is not called from the UI thread, is it?

kumy: IIRC original call comes from https://github.com/kumy/cgeo/blob/issue-5135/main/src/cgeo/geocaching/TrackableActivity.java#L213 which itself is called from `onCreate()`

samueltardieu: If this is the case, this is **not** OK. @Bananeweizen already introduced such a network call on the UI thread recently, and this should not enter our code base.

kumy:
I'm curious on how @Bananeweizen fixed that. @samueltardieu do you remember the class/function where this was implememted?

samueltardieu:
Nope, I don't remember. I just remember that it is done wrong.

kumy:
Found the other code where this was also implemented: https://github.com/cgeo/cgeo/commit/766d09124dbbddb96a991b0db841500416a6dd1e#commitcomment-13330678

main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
kumy: @samueltardieu I've looked deeper into call stack of this function. Concerned `getGeocodeFromTrackingCode()` function and calling function `searchTrackable()` are running on network thread. 
Here is the call stack from the caller function. Do you think it's ok? Thanks
capture du 2015-10-22 22 59 56

Bananeweizen:

kumy:
(I have to rebase the code on current master).
I'm not sure the network call is on the UI thread, and if it is really, I didn't found a good/clean way to do it. Any help will be appreciated ;)

samueltardieu:
I don't understand why you don't handle that as you handle loading a cache from the network and display it. You can either display minimal data and refresh it as soon as the right information is available, or you can use a "loading" activity which waits for the data to be available.

Lineflyer:
As this PR can no longer be merged I close it.
@kumy If you think different please feel free to reopen, else submit a new PR is still needed.

kumy:
@Lineflyer I'll push a new PR as soon as I'll have time :)

PR #5194, opened 2015-09-15
Icon size in lists different with/without overlay

kumy:
Here is a patch to fix this issue.
It more a workaround than a really clean fix. There is lots of weird things there regarding general icons resolution in different resolutions, ratio etc...

Do you agree to merge in release?


rsudev:
I tried to get back a little bit into this topic again (I also had to fiddle with the sizes once in the past), but this change is not really clear to me.
Meanwhile we also got a complaint that the markers on the map got even smaller for one user (https://geoclub.de/forum/viewtopic.php?f=155&t=67870&start=510#p1231183 German geocaching forum), so I suggest to take a second look here (especially as the release is going smooth otherwise)

Lineflyer:
@rsudev Opened #5195 for the mentioned problem.

kumy:
@rsudev thanks for the "second look" :) let me know if I can help you

samueltardieu:
@kumy, can this be closed now?

kumy:
Based on pictures from the other PR from @rsudev. The "problem" has disapeared. Let s close this.

Issue #5196, opened 2015-09-15
Have confirm selection also in coordinate input on top


Lineflyer:
From support mail:

While in waypoint adding the "OK" has been moved to the top the coordinate input window still ahas an "Ready" button at the bottom which might be blocked by the keyboard.
It would be much more convenient to also have an "Ready" button on the top of the window.

See explanation:
![_20150915_074707](https://cloud.githubusercontent.com/assets/949669/9891774/353ef7a2-5c09-11e5-9833-638e5b4e5edc.JPG)

---

I like this suggestion!

SammysHP:
The difference is that waypoint editing is an activity with an action bar while the coordinate dialog is a dialog that does not have an action bar.

Lineflyer:
Ah, ok. 
But couldn't we place the "Done" box just on the top right and replace "Done" by a tick mark to make it look alike?

rahu:
+1 - don't know how often I cried because of this.

rahu:
what du you think about 9b67fa17cee3a2c3f1539d21470784a98faf8e4e ?

-> looks like this

![screenshot_2015-10-17-13-17-46](https://cloud.githubusercontent.com/assets/937126/10558519/d8e06374-74d1-11e5-8e7c-1b915b271d49.png)

and works great. 
(can someone please tell me the name of the correct drawable...)

rsudev:
In the waypoint edit action bar we use android:icon="@drawable/ic_menu_done_holo_dark" (see edit_waypoint_options.xml in res/menu)

rahu:
thanks.. 

what about adding a cancel button as well?

rsudev:
For consistency it surely is helpful, even though I mostly use the back button anyway in this case.

rahu:
![screenshot_2015-10-17-20-52-09](https://cloud.githubusercontent.com/assets/937126/10560639/216a8f2c-7511-11e5-86b4-ed09ab706104.png)

Lineflyer:
So the Done button could be removed...or shall we keep this redundancy?

rsudev:
Done button should be removed!

rahu:
ok - done button is now hidden. 

BTW: Is it really necessary to check if the NO_TITLE feature is supported? I belief there is no phone outside which isn't supporting this? But I haven't found any further information on this.

Lineflyer:
I can't judge the code, however the screenshot looks good. You should sent a PR to allow review of the changes by other contributors.

[...]

PR #5248, opened 2015-10-20
Issue #5196

rahu:
issue #5196: done button is now on top of the coord's input dialog and removed from bottom end. also added cancel button next to done button


Lineflyer:
Can someone review the code. I deem this a very nice improvement.

PR #5249, opened 2015-10-20
Add missing android prefix.

UniQP:


SammysHP:
The code in `android-support-v7-appcompat` is not from us. We did only include it for some reasons. (Which reasons? After the migration to gradle this might be obsolete.)

Lineflyer:
Is this PR triggered by any issue or aligned with any of the repository admins?

SammysHP:
No. And as I said it's not our code and we should only modify it if it's really necessary.

rsudev:
We include this code (from Google support library) because this is necessary for integrating it into ant/eclipse build environment.
IMO changes should be done when we switch to a new version or a bugfixed version.
I would also be interested in what triggered this PR. Is this just an inconsistency? Does it have any effect in c:geo? Why is it wrong?

SammysHP:
It is wrong because the color attribute is not in the global namespace and has no namespace prefix. The result might be a minor graphical glitch, but nothing serious.

Usually library projects are referenced and not included in the code repo. IIRC we talked about submodules some time ago, but that should be obsolete once we use gradle.

UniQP:
The problem was reported by Android Lint.

After your comments, I have taken a look at the original repository https://android.googlesource.com/platform/frameworks/support . There, the issue was fixed in 39ce6efbd7a2e15374784ebbb8eff9e2eef5f3b6 .

Since you copied the code into your repository, it's your job to maintain it. However, using a git submodule (as also suggested in cgeo bd1f19167c978280432fc21edc82f3e6a2dcacaf) would be the cleaner solution.

@SammysHP: According to the comment of rsudev, the code is not obsolete.

SammysHP:
The copy of the code in our repository as well as submodules will be obsolete.

Lineflyer:
So, shall this be merged or should we better update the complete library to ne newest version?

SammysHP:
Let's update the library. c:geo works with the current version and we shouldn't modify third party code in that case.

UniQP:
@SammysHP Why did you close the issue although the library is not updated yet?

SammysHP:
It is explained above. We do not modify third party code unless it is really necessary. I don't know if we already use the library as a dependency or still a copy in our repository, but this is a pull request that we won't merge.

UniQP:
So what's the process? Should I create a new issue for the corresponding Android Lint warning?

rsudev:
This code is no longer used for building c:geo, as we switched to gradle almost completely (I think the current master will not build at all with ant).
The removal will follow in due course. No need for any further action in this area.

Bananeweizen:
@UniQP That code is only used as long as you are building from Ant or Eclipse. If you use gradle or run Android Studio/IDEA, it is not used at all. In that case, you can basically delete it from your local repo. Right now this project/module has just not been deleted as some devs still use Eclipse for local development. But our CI server does not build from that source anymore.

SammysHP:
Which means that we get a different result depending on build method. We should really delete duplicate dependencies.

samueltardieu:
:+1:

PR #5256, opened 2015-10-24
Issue5253

rahu:
this fixed 5253: edit personal note has done / cancel button on top. you can edit the note from optionsmenu and listing will not jump back to fist line after editing.


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2294,11 +2297,6 @@ protected Void doInBackground(final Void... params) {
                 DataStore.saveCache(cache, EnumSet.of(SaveFlag.DB));
                 return null;
             }
-
-            @Override
-            protected void onPostExecute(final Void v) {
-                notifyDataSetChanged();
-            }
```
samueltardieu: Could you explain why you removed this?

main/src/cgeo/geocaching/CacheMenuHandler.java
```diff
@@ -88,6 +88,7 @@ public static void onPrepareOptionsMenu(final Menu menu, final Geocache cache) {
             return;
         }
         final boolean hasCoords = cache.getCoords() != null;
+		menu.findItem(R.id.menu_edit_fieldnote).setVisible(true);
```
SammysHP: Indentation.

main/src/cgeo/geocaching/CacheDetailActivity.java
SammysHP: To "fix" the problem that the description page jumps back to the top after closing the dialog.

The reason for the jump is that the page is recreated with the new data. I assume that the content of the page won't be updated without `notifyDataSetChanged()`.

rahu:
These causes ui to be redrawn and line one is back on top. You have to scroll down again.. And I haven't found a reason why not simply remove this.

main/src/cgeo/geocaching/CacheMenuHandler.java
rahu: I hate eclipse...

main/src/cgeo/geocaching/CacheDetailActivity.java
rahu: So is there a reason to leave it? In this case I'll find another solution.

SammysHP: Yes. The detail activity must know when the data was changed and when it has to update the UI.

rahu: But data (in this case personal note is updated without calling this).  The only difference I found after changing back is the "note has been saved" info which is not shown any longer  :( I'll fix this and come back...

rahu:
ok guys - I updated the commit. Now scroll position is reset to the last position after updating UI.

Lineflyer:
@rahu Did you see, that this has conflicts to merge? Maybe you should send a new PR based on current master.

Issue #5250, opened 2015-10-20
Moto G compass fails completely


Lineflyer:
From support mail.

A user performed an upgrade to Android 5.2 which was offered by Motorola for his device.
Afterwards the compass of c:geo stopped working.

I first advised to try the setting of using the orientation sensor...He reported back that it is not selectable.

Other compass apps are working all the time before and after the upgrade.

The sys info is as follows:
```
--- System information ---
Device: XT1032 (falcon_retde, motorola)
Android version: 5.1
Android build: LPB23.13-56
Cgeo version: 2015.09.28
Google Play services: disabled
Low power mode: inactive
Compass capabilities: yes
Rotation sensor: absent
Orientation sensor: present
Hide own/found: false
Map strategy: detailed
HW acceleration: enabled (default state) 
System language: de_DE 
Geocaching sites enabled:
- geocaching.com: logged in (Login OK) / Premium
 Installed cgeo plugins: none
--- End of system information ---
```

@samueltardieu 
Is it really possible that the device does not provide rotation sensor but only orientation?

samueltardieu:
Yup, this is possible. I don't know what is wrong.

GSNord:
Having the very same problem since the last Motorola system update. It was not an Android update; still Android 5.1; it was a firmware update for my Moto G device.
Android Version 5.1
Kernel Version 3.4.42-g89906d6
System Version 221.21.56.falcon_umts.Retail.en.DE

I performed a compass recalibration using GPS Status+Tools app. So the compass seems to work using GPS Status+Tools, but with some strange behaviour. It points to north most of the time, but movement is faster since the system update. Also the compass is influenced by the horizontal g-sensor; i.e. if I shake the device left to right the compass starts rotating.

Using cgeo, the compass initially points to the direction of the cache, but it will not move any more when I turn the device.

GSNord:
Checked it with Google Maps today. It is my out-of-the-box never-updated version 9.4.0 Google Maps of 2014.
It rotates the map correctly when I rotate my device. However, it also turns the map for some degrees when I shake the device horizontally.

samueltardieu:
@GSNord Can you please double-check that you haven't activated the GPS-only mode (from the Compass screen)?

GSNord:
I noticed that the new cgeo version can switch between compass+GPS and GPS only in the compass screen. The described behaviour is when I use compass+GPS.
[...]

PR #5265, opened 2015-11-02
Issue #5250 

bck5:
Add a new Provider for cardinal directions to solve issue #5250 (Compass not working on Moto G)

### DIFF
--- a/main/src/cgeo/geocaching/sensors/MagnetometerAndAccelerometerProvider.java
+++ b/main/src/cgeo/geocaching/sensors/MagnetometerAndAccelerometerProvider.java
@@ -1,0 +1,96 @@
+package cgeo.geocaching.sensors;
+
+import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.RxUtils;
+
+import rx.Observable;
+import rx.Observable.OnSubscribe;
+import rx.Subscriber;
+import rx.functions.Action0;
+import rx.subscriptions.Subscriptions;
+
+import android.content.Context;
+import android.hardware.Sensor;
+import android.hardware.SensorEvent;
+import android.hardware.SensorEventListener;
+import android.hardware.SensorManager;
+
+public class MagnetometerAndAccelerometerProvider {
+
+    private MagnetometerAndAccelerometerProvider() {
+        // Utility class, not to be instantiated
+    }
+
+    public static Observable<Float> create(final Context context) {
+
+        final SensorManager sensorManager = (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
+        final Sensor accelerometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
+        final Sensor magenetometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD);
+        if (magenetometerSensor == null || accelerometerSensor == null) {
+            return Observable.error(new RuntimeException("no magenetic or accelerometer sensor"));
+        }
+        Log.d("MagnetometerAndAccelerometerProvider: sensors found");
+        final Observable<Float> observable = Observable.create(new OnSubscribe<Float>() {
+            private final float[] lastAccelerometer = new float[3];
+            private final float[] lastMagnetometer = new float[3];
+            private boolean lastAccelerometerSet = false;
+            private boolean lastMagnetometerSet = false;
+            private final float[] rotateMatrix = new float[9];
+            private final float[] orientation = new float[3];
+            @Override
+            public void call(final Subscriber<? super Float> subscriber) {
+                final SensorEventListener listener = new SensorEventListener() {
+                    @Override
+                    public void onSensorChanged(final SensorEvent sensorEvent) {
+
+                        if(sensorEvent.sensor.equals(accelerometerSensor)) {
+                            System.arraycopy(sensorEvent.values, 0, lastAccelerometer, 0, sensorEvent.values.length);
+                            lastAccelerometerSet = true;
+                        } else if(sensorEvent.sensor.equals(magenetometerSensor)){
+                            System.arraycopy(sensorEvent.values, 0, lastMagnetometer, 0, sensorEvent.values.length);
+                            lastMagnetometerSet = true;
+                        }
+                        if(lastAccelerometerSet && lastMagnetometerSet) {
+                            SensorManager.getRotationMatrix(rotateMatrix, null, lastAccelerometer, lastMagnetometer);
+                            SensorManager.getOrientation(rotateMatrix, orientation);
+                            subscriber.onNext((float) (orientation[0] * 180 / Math.PI));
+                        }
+                    }
+
+                    @Override
+                    public void onAccuracyChanged(final Sensor sensor, final int i) {
+                        /*
+                         * There is a bug in Android, which apparently causes this method to be called every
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/sensors/MagnetometerAndAccelerometerProvider.java
```diff
@@ -0,0 +1,96 @@
...
+            private boolean lastMagnetometerSet = false;
+            private final float[] rotateMatrix = new float[9];
+            private final float[] orientation = new float[3];
+            @Override
+            public void call(final Subscriber<? super Float> subscriber) {
+                final SensorEventListener listener = new SensorEventListener() {
+                    @Override
+                    public void onSensorChanged(final SensorEvent sensorEvent) {
+
+                        if(sensorEvent.sensor.equals(accelerometerSensor)) {
```
samueltardieu: Please use space after `if`.

main/src/cgeo/geocaching/sensors/MagnetometerAndAccelerometerProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private MagnetometerAndAccelerometerProvider() {
+        // Utility class, not to be instantiated
+    }
+
+    public static Observable<Float> create(final Context context) {
+
+        final SensorManager sensorManager = (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
+        final Sensor accelerometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
+        final Sensor magenetometerSensor = sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD);
```
samueltardieu: Typo (should be `magnetometerSensor`).

main/src/cgeo/geocaching/sensors/Sensors.java
```diff
@@ -100,7 +102,16 @@ public void setupDirectionObservable() {
 
         // On some devices, the orientation sensor (Xperia and S4 running Lollipop) seems to have been deprecated for real.
         // Use the rotation sensor if it is available unless the orientatation sensor is forced by the user.
-        final Observable<Float> sensorDirectionObservable = Settings.useOrientationSensor(app) ? OrientationProvider.create(app) : RotationProvider.create(app);
+        // after updating Moto G there is no rotation sensor anymore. Use magnetic field and accelerometer instead.
```
samueltardieu: Please use upper-case `a` at the beginning of the sentence.

samueltardieu:
The patch looks mostly fine to me. There should be some information in `SystemInformation.java` too.

samueltardieu:
Also, please use a terser first line in the commit message, such as:

> fix #5250: compass not working on Moto G
>
> Add a new Provider for cardinal directions.

bck5:
I corrected your remarks and also add information about the used sensors in `SystemInformation.java` .

I also used another (better?) commit message, hope it is ok this time.

Lineflyer:
@bck5 
I am taking care of testing and support for c:geo (as far as I can).
Thus I would be glad if you could provide some explanation about the Moto G problem and how you solved it in "some easy words" to have a basis for testing and feedback to users.
What is the difference in the sensor meachnism and to what devices does it apply?
Is this working around a problem on the devices or a general new approach?

bck5:
The Problem is: after the update for Moto G it seems there is no more rotation sensor provided by the system.It also seems the system do not offer the orientation sensor. Cgeo use one of these sensors, to get the direction right.

But these sensors are not the only one for calculating the cardinal directions. With help of the magnetic field sensor and the accelerometer it is also possible to get the right rotation matrix.[1]

So my fix is a new approach calculating the rotation matrix. The sensor type "TYPE_ROTATION_VECTOR" which is used by cgeo can be hardware or software based (see [2]). Because of this I think it was only a software sensor which does not exists anymore after the Moto G update.  

This fix will give all device the possibility to use the compass if they provided the sensor types "TYPE_MAGNETIC_FIELD" and "TYPE_ACCELEROMETER". These both sensors are supported by android since version 1.5[2]. I don't know which devices provided these two, but not rotation_vector or orientation sensor. If there exists such devices (like the Moto G) this fix will apply to them, too.


To say it in easier words: my fix (more or less) emulates the rotation vector sensor and doing this it brings the compass back. 

hope this helps
bck5


[1]https://developer.android.com/reference/android/hardware/SensorManager.html#getRotationMatrix(float[], float[], float[], float[])
[2]https://developer.android.com/guide/topics/sensors/sensors_overview.html

samueltardieu:
Thanks for your contribution. I will first merge it to `master` and let people field test it for a few days before we merge it to release.

samueltardieu:
Closing the pull request (merged to `master`). Thanks for basing your work on `release` as expected for a bug fix, this will let us put it in this branch after the field test.

Lineflyer:
@bck5 

> This fix will give all device the possibility to use the compass if they provided the sensor types "TYPE_MAGNETIC_FIELD" and "TYPE_ACCELEROMETER". These both sensors are supported by android since version 1.5[2]. I don't know which devices provided these two, but not rotation_vector or orientation sensor. If there exists such devices (like the Moto G) this fix will apply to them, too.

Thanks for explaining...
What is the priority in which the different approaches are now used... or more practical:
Will this change also affect devices which do have the "normal" sensors?

bck5:
It will not.

If devices have the "old" sensors, it will use them, or if the orientation sensor is forced it will use the orientation sensor.
So it will behave like it does before the patch.

samueltardieu:
The orientation sensor is deprecated but may result in a better fusion (for example, it may use the gyroscope if present). Is it absent or is it non-working on Moto G?

samueltardieu:
Ok, from the info I got from a friend trying the NB, it is present but non-working.

```
--- System information ---
Device: MotoG3 (cm_osprey, Motorola)
Android version: 5.1.1
Android build: cm_osprey-userdebug 5.1.1 LMY48W 9030813075 test-keys
Cgeo version: 2015.11.08-NB-4482e03
Google Play services: enabled
Low power mode: active
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: present
Direction sensor used: magnetometer & accelerometer
Hide own/found: false
Map strategy: auto
System language: fr_FR
Geocaching sites enabled:
- geocaching.com: logged in (Connecté) / Basic
Installed cgeo plugins:  none

Maybe we should give the user a choice of the sensor to use in the settings rather than a simple "force use of rotation sensor".
```

bck5:
Yes, I run a stock Android on the MotoG1.
```
--- System information ---
Device: XT1032 (falcon_retde, motorola)
Android version: 5.1
Android build: LPB23.13-56
Cgeo version: 2015.11.10-NB-477e26c
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: present
Direction sensor used: magnetometer & accelerometer
Hide own/found: false
Map strategy: auto
System language: de_DE
Geocaching sites enabled:
- geocaching.com: not logged in (Keine Anmeldedaten gespeichert.)
Installed cgeo plugins:  none
```

The system info says, the orientation sensor is present, but using the orientation sensor for the compass results in a non-working compass.

samueltardieu:
By non-working, you mean that it outputs no value at all?

bck5:
I did not looked at the values it outputs, but it seems it is always outputs the same value. It results in a non moving compass.

Issue #5253, opened 2015-10-22
Have OK/Cancel also on top for personal note editing


Lineflyer:
From FB after users tried our nightly which included the fix for #5196 

Also the personal note input window should have the Done/Cancel selection on top sue to ease of use and design consistency.

@rahu In case you are interested and it is technically possible...

rahu:
I'll look at this. 

Within this context my wife pointed me to 2 other details:
- after writing a field note listing scrolls back to top
- you have to scroll down to write a field not
Usually she want's to write the field note after you have found the solution for a task which is not at the end of the listing. -> you have to scroll down, write the note, an then have to scroll to the last position in the listing again.  I think it's worth to look into it and maybe avoid these scroll actions. (write Fieldnote from context menu and remember position)

rahu:
![screenshot_2015-10-23-22-00-42](https://cloud.githubusercontent.com/assets/937126/10703641/9e6cdc36-79d1-11e5-8224-1494d1219052.png)

just a preview - still need to do some code cosmetics...

PR #5274, opened 2015-11-08
fix issue #5253: have ok / cancel on top of personal note dialog

rahu:


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -738,6 +741,8 @@ private void notifyDataSetChanged() {
             return;
         }
 
+        final ScrollView scroll = ButterKnife.findById(this, R.id.detailScroll);
```
samueltardieu: Why so early instead of near `scrollNew`?

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -761,6 +766,17 @@ private void notifyDataSetChanged() {
         progress.dismiss();
 
         Settings.addCacheToHistory(cache.getGeocode());
+        final ScrollView scrollNew = ButterKnife.findById(this, R.id.detailScroll);
+        if ((scroll != null) && (scrollNew != null)) {
```
samueltardieu: Inner parenthesis are useless.

In which case can this condition be true?

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -761,6 +766,17 @@ private void notifyDataSetChanged() {
...
+            final int lastScrollPos = scroll.getScrollY();
+            scrollNew.post(new Runnable() {
+            @Override
```
samueltardieu: Suspicious indentation.

main/src/cgeo/geocaching/ui/EditNoteDialog.java
```diff
@@ -69,24 +71,36 @@ public Dialog onCreateDialog(final Bundle savedInstanceState) {
...
-                    @Override
-                    public void onClick(final DialogInterface dialog, final int whichButton) {
-                        dialog.dismiss();
-                    }
-                });
         final AlertDialog dialog = builder.create();
+
+        dialog.requestWindowFeature(Window.FEATURE_NO_TITLE);
+        final TextView title = ButterKnife.findById(view, R.id.dialog_title_title);
+        if (title != null) {
```
samueltardieu: In which case can this be true?

main/src/cgeo/geocaching/ui/EditNoteDialog.java
```diff
@@ -69,24 +71,36 @@ public Dialog onCreateDialog(final Bundle savedInstanceState) {
...
+            title.setText(R.string.cache_personal_note);
+            title.setVisibility(View.VISIBLE);
+        }
+        final ImageButton cancel = ButterKnife.findById(view, R.id.dialog_title_cancel);
+        if (cancel != null) {
```
samueltardieu: In which case can this be true?

main/src/cgeo/geocaching/ui/EditNoteDialog.java
```diff
@@ -69,24 +71,36 @@ public Dialog onCreateDialog(final Bundle savedInstanceState) {
...
+            cancel.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(final View view) {
+                    dialog.dismiss();
+                }
+            });
+            cancel.setVisibility(View.VISIBLE);
+        }
+        final ImageButton done = ButterKnife.findById(view, R.id.dialog_title_done);
+        if (done != null) {
```
samueltardieu: In which case can this be true?

main/src/cgeo/geocaching/CacheDetailActivity.java
rahu: good question - I can move this 17 lines down, right before 
reinitializeViewPager();
Can't execute this after this call because I would get the new scrollview.

main/src/cgeo/geocaching/CacheDetailActivity.java
rahu: someone removed / renamed the detailScroll ? 
inner parenthesis are for readability. 
I can remove the statement if you think it's not needed...

rahu:
I removed nearly all of the checks. The remaining one is needed in case the view is created first time where notifyDataSetChanged gets called before the ScrollView is created.

samueltardieu:
Thanks for your contribution. Note that it would have been better to use `fix #5253: …` instead of `fix issue #5253: …`, since it would have closed the issue automatically.

PR #5302, opened 2015-12-01
fix #5296

### LINKED ISSUE #5296: Crash when adding personal note from WP or Images tab
Found this strange issue during Russian localization testing. Here is the test case:

0. Store a cache for offline use.
1. Open Cache details
2. Slide to page with Waypoints or Images
3. Press button with 3 dots
4. Press 'Personal notes" 
5. Enter some text
6. Press "OK"

Application will crash.

Some notes:
- The case is not reproducible from Description or Details or Logbook pages, only from Images and WP
- I found that I cannot add a personal note to a cache if it wasn't stored. Button "Personal notes" just does nothing. I assume this as a separated FR to allow to write personal notes for non-stored caches (or add error message for such case)

My phone is Xperia Z3 Compact, Android 5.1.1
C:geo 2015.11.29-NB1-a4faa8b

rahu:
In #5296 are several issues named:
- crash when editing note from WP or Images Tab
- edit-note button / menu is shown if cache is not saved.

### DIFF
--- a/main/res/menu/cache_options.xml
+++ b/main/res/menu/cache_options.xml
@@ -114,6 +114,6 @@
         android:id="@+id/menu_edit_fieldnote"
         android:icon="@drawable/ic_menu_rotate"
         android:title="@string/cache_personal_note"
-        app:showAsAction="ifRoom">
+        android:visible="false">
     </item>
 </menu>
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -89,6 +89,7 @@ import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.functions.Func0;
 import rx.functions.Func1;
+import rx.schedulers.Schedulers;
 import rx.subscriptions.CompositeSubscription;
 import rx.subscriptions.Subscriptions;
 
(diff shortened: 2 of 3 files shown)


samueltardieu:
This fix is overly complicated. You don't need to pass the boolean parameter around, you can set the personal note outside the `AsyncTask` and call `notifyDataSetChanged()` right after, and do only the saving in the `AsyncTask` (or, better, use a scheduler from `AndroidRxUtils`).

Also, there are several style issues, in the code and in the commit message (the line is too long).

rahu:
I'm not sure, but since `AsyncTask` is a background operation it isn't predictable when it's executed. Imagine I call `notifyDataSetChanged()` right after and slide to Description really fast. I think in this case, maybe in combination with a high system load it's possible the new Details aren't storred because `AsyncTask` hasn't run. That's why I used `onPostExecute` for this.

Btw. do we need a Mutex for `Geocache::personalNote`? It's accessed from two different threads at the same time.

samueltardieu:
The cache will be stored in memory (in the CacheCache), and the description will have been changed there already. Only the save in the database might not have completed, but this won't change what's displayed or the fact that the cache will eventually be saved to the database.

rahu:
Ok, so move
`    cache.setPersonalNote(note);`
`    cache.parseWaypointsFromNote();`
out of the AsyncTask and call `notifyDataSetChanged()` after this would be fine.

samueltardieu:
Yup. Then, afterwards, schedule a save into the database using

``` java
Schedulers.io().createWorker().schedule(new Action0() {
  …
});
```

rahu:
Not working. `notifyDataSetChanged` reloads the Cache from the Database. Doing this before the new note is saved is a bad idea... 
What about using `reinitializeViewPager` instead of  `notifyDataSetChanged`?

rahu:
All known sub issues should be solved now.

Bananeweizen:

samueltardieu:
Thanks for your contribution.

PR #5306, opened 2015-12-02
Fixed  #5231

### LINKED ISSUE #5231: User name for GCVote can not be deleted
2015.10.08-NB:
- Open Settings/Services
- Open GCVote settings
- Click on username and replace the existing user by an empty field
- Select OK
- Select "Verify login"

This will return "Check OK" and I do see in the log that the previously stored username is still in use
- Close c:geo / Force close
- Start c:geo again
- Go to Settings/Services/GCVote

You will see, that the previous username has been restored.

As the username of GCVote and geocaching.com need to be identical we most probably derive it from the settings for geocaching.com.
However couldn't we just disallow editing of the username in the GCVote settings and just show it for informational purposes.

GodfatherThe:
Locked username on GCVote page, added refreshing the preference when changing from GC.com screen.

### DIFF
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -313,6 +313,7 @@
                         android:text="@string/init_gcvote_password_description" />
 
                     <EditTextPreference
+                        android:enabled="false"
                         android:dependency="@string/pref_ratingwanted"
                         android:dialogTitle="@string/init_username"
                         android:hint="@string/init_username"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -462,7 +462,7 @@
     <string name="settings_activate_oc_uk">Activate</string>
     <string name="init_oc_uk_description">Authorize c:geo with opencaching.org.uk to search for caches and access/filter your found caches.</string>
     <string name="init_gcvote">GCvote.com</string>
-    <string name="init_gcvote_password_description">To be able to rate a cache, you need to follow the instructions at GCVote.com and enter your GCVote password here.</string>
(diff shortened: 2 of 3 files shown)


GodfatherThe:
@Bananeweizen @rsudev @Lineflyer could you please take a look?

main/res/xml/preferences.xml
```diff
@@ -313,6 +313,7 @@
                         android:text="@string/init_gcvote_password_description" />
 
                     <EditTextPreference
+                        android:enabled="false"
```
rsudev: Did you consider switching it to a simple (and therefore display-only) Preference? I find the disabled look unfitting.
We likely should also adapt the string to tell that this is in fact the geocaching.com username.

GodfatherThe:
@rsudev If it's usual Preference, then it looks like user is able to change it, but in fact he\she can not, so it can be confusing. However, when it's disabled, it underlines that you're unable to edit this. 
Agree about the string, added. And also I need only login handling, so added a condition.

SammysHP:
rsudev did not mean a normal preference, but a text preference like the one we use for descriptions etc.

Bananeweizen:

SammysHP:
The patch might work, but we are still discussing about the method that is used to show that it is disabled.

SammysHP:
@GodfatherThe Are you still interested in this patch?

rsudev:
ok to test

samueltardieu:
No news from submitter and not mergeable as-is, closing, reopen one if you want to start working on this again.

PR #5331, opened 2015-12-20
crowdin: New translations

rsudev:
downloaded 2015-12-20 22:09:25


rsudev:
Check fail is due to emu not available on ci -merging anyway

PR #5345, opened 2015-12-30
fix #3819: make description text selectable

### LINKED ISSUE #3819: Ability to copy only a part of the description
I often want to copy only a part of the description to use in other apps.
It might be a coord, a word that I want to google, a cipher in a mystery cache that I want to decipher etc.
But when I long press the description there are only an option to copy all of the description.

I would love if there was a possibility to select and copy a part of the description or part of the hint.

Suggestion:
Add a "Select text" option in the longpress menu
Use standard text selection to select a part of the text.
Like in the mockup below.

Select text

rahu:
with these changes it's possible to copy only parts of description. I removed the separation of short and long description to be able to select both in a single selection. 
Testing the new feature I run into https://code.google.com/p/android/issues/detail?id=137509 . The workaround for this works fine.
screenshot_2015-12-30-18-48-22

### DIFF
--- a/main/res/layout/cachedetail_description_page.xml
+++ b/main/res/layout/cachedetail_description_page.xml
@@ -12,10 +12,10 @@
         android:orientation="vertical"
         android:padding="4dip" >
 
-        <!-- Short description -->
+        <!-- Description -->
 
         <cgeo.geocaching.ui.IndexOutOfBoundsAvoidingTextView
-            android:id="@+id/shortdesc"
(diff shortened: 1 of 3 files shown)


Lineflyer:
Will the "Translate" feature still work with your changes? I used to be in the same context menu (or am I wrong)?

rahu:
Translate feature? I didn't notice this before reading your comment. It's broken now. I'll come back with a solution.

Issue #5202, opened 2015-09-24
Clipboard copy of cache description sometimes contains HTML code


Lineflyer:
From FB and also noticed this myself recently with current market release 2015.09.15

When copying the cache description the clipboard content is sometimes plain text (as expected) but sometimes it contains also all HTML codes as well (so its more or less the source code of the description)

rsudev:
Do you have a sample for text only? Or is it varying for the same cache? (All I tried had html in the clipboard...)

Lineflyer:
GC5H5TZ is text only



Von Samsung-Tablet gesendet

-------- Ursprüngliche Nachricht --------
Von: rsudev <notifications@github.com> 
Datum:23.09.2015  22:05  (GMT-07:00) 
An: cgeo/cgeo <cgeo@noreply.github.com> 
Cc: Lars <gc@droescher.eu> 
Betreff: Re: [cgeo] Clipboard copy of cache description sometimes contains
  HTML code (#5202) 

Do you have a sample for text only? Or is it varying for the same cache? (All I tried had html in the clipboard...)

rsudev:
For me even this description contains all the html which is also used in the description on gc.com. It is just less prominent (only some <p> tags). If you don't see those, the question would be what your paste target is.

Bananeweizen:
I'm wondering whether we might need 2 different copy actions, as probably sometimes it is desired to have the text only, sometimes it might be desired to have the HTML. Can we put both to the clipboard or can we decorate the object with a mimetype? At least on Windows, copying rich text to the clipboard would still allow to paste it as plain text. I would assume similar abilitities on Android...

Lineflyer:
@rsudev 
You are right. My example also contained HTML.
It seems we always export HTML here... I am quite sure it was text only somewhen in the past and only recently this behaviour changed. 

I agree to @Bananeweizen that both options can be useful.
I suggest that if something is changed in this context we should also check the possibility to copy only part of the description as mentioned in #3819.
So the final implementation could offer two options:
1. Copy complete description as HTML (as currently implemented)
2. Copy text only but allow text selection

samueltardieu:
I'm not sure we need two options. We could just set the proper mime type on the data we put in the clipboard. If an application can handle HTML, it will get HTML, otherwise I guess it will get plain text data converted from HTML.

samueltardieu:
Or let's implement that for API >= 16 (Android 4.1.x, +80% of our users base).

[...]

PR #5351, opened 2016-01-01
fix #3819 (and #5202): make description text selectable

### LINKED ISSUE #3819: Ability to copy only a part of the description

rahu:
with these changes it's possible to copy, share and translate only parts of description. I removed the separation of short and long description to be able to select both in a single selection.

Testing the new feature I run into https://code.google.com/p/android/issues/detail?id=137509 and https://code.google.com/p/android/issues/detail?id=23381. The workarounds are working fine.

### DIFF
--- a/main/res/layout/cachedetail_description_page.xml
+++ b/main/res/layout/cachedetail_description_page.xml
@@ -12,10 +12,10 @@
         android:orientation="vertical"
         android:padding="4dip" >
 
-        <!-- Short description -->
+        <!-- Description -->
 
         <cgeo.geocaching.ui.IndexOutOfBoundsAvoidingTextView
-            android:id="@+id/shortdesc"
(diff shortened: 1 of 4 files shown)


Bananeweizen:

rsudev:
I can probably take a look at it today evening

rsudev:
ok to test

rsudev:
ok to test

Issue #5264, opened 2015-11-01
Amount of favorite points not updated when awarding a favorite point in logging view


Lineflyer:
2015.10.31-NB:
- Open a cache
- Select "Log Visit"
- Post your log and select (the new) checkbox for "Add to your favorites"

When returning to the cache details afterwards the state is not updated:
- The amount of favorite points for the cache is not incremented
- The button still shows "Add to your favorites"

Expectation:
If posting the log was OK, the information of the (stored) cache should be updated in cache details accordingly to avoid user confusion.

@madankb

madankb:
I will check this issue.

samueltardieu:
@madankb Any news on this?

madankb:
Sorry for the delay.. Got held up in other works.. I will fix it as soon as
possible.

madankb:
Can someone please review my code for the PR#5367?.

PR #5367, opened 2016-01-10
Bug fix for the issue #5264

madankb:
Code fix for the issue "Amount of favorite points not updated when awarding a favorite point in logging view #5264".

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
@@ -1,11 +1,21 @@
 package cgeo.geocaching.connector.gc;
 
-import cgeo.geocaching.storage.DataStore;
-import cgeo.geocaching.models.Geocache;
-import cgeo.geocaching.models.Image;
+import android.os.Bundle;
+import android.support.v4.app.LoaderManager;
+import android.support.v4.content.Loader;
+import android.widget.CheckBox;
+
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
```diff
@@ -14,25 +24,15 @@
 import cgeo.geocaching.enumerations.LogType;
 import cgeo.geocaching.enumerations.StatusCode;
 import cgeo.geocaching.loaders.UrlLoader;
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.models.Image;
+import cgeo.geocaching.models.TrackableLog;
```
SammysHP: Sorting of imports messed up.

madankb:
@Sammy I am using Android studio 2.0 version which I believe sort the headers in this order. Should I need to rearrange in old order? Do I have any other changes in the code? Please confirm.

madankb:
@SammysHP Is my code fine to merge?.

main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
SammysHP: Sorry, it's ok. I am used to see the current package located at the top and was somehow confused by the diff.

SammysHP:
We do not check the result web page, do we? Instead of just assuming that the favorite point was given to the cache successfully it might be better to check the real result.

As we do the same for other properties of a cache, I think this patch is fine and I'll merge it. Thanks for your contribution!

PR #5368, opened 2016-01-10
Support links encoded in some GeoKrety QRCode

kumy:
Follow PR#5168

Code has been rebased.

@samueltardieu I've looked again at the call of `getGeocodeFromTrackingCode()`. The call is from `GeokretyConnector.searchTrackable()` which already deal with network calls. On the other side, `GeokretyConnector.searchTrackable()` is in turn called from inside `AppObservable.bindActivity()` in `TrackableActivity`. So I don't think this function is really problematic. May you please deeper look into this PR?

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -614,6 +614,7 @@
                 <data android:host="geokrety.org" />
                 <data android:host="www.geokrety.org" />
                 <data android:pathPrefix="/konkret.php" />
+                <data android:pathPrefix="/m/qr.php" />
             </intent-filter>
 
             <!-- GeoKretyMap.org -->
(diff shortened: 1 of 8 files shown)


rsudev:
ok to test

rsudev:
@kumy The check result in this case indicate a build fail (you need to look into the build console log to see that). Could you cross-check that?

kumy:
Ho, I'll check...

On January 16, 2016 6:37:19 PM GMT+01:00, rsudev <notifications@github.com> wrote:
>@kumy The check result in this case indicate a build fail (you need to
>look into the build console log to see that). Could you cross-check
>that?
>
>---
>https://github.com/cgeo/cgeo/pull/5368#issuecomment-172234490

-- 

kumy:
@rsudev I've missed to change an include while merging... Should be ok now. Thank you for your vigilance.
BTW the new job for checking the PR is really cool :)

PR #5370, opened 2016-01-10
Fixed #5339 no feedback no personal note upload

### LINKED ISSUE #5339: Upload of Personal Note fails, but says it did
If one tries to upload a personal note, despite not connected to the internet, one still get a success message "Personal note uploaded".

reproducable: always
occurs since: quite a while ago
Device/Android: HTC One m7, Android 5.1 (Google Play Edition)

pstorch:
Evaluating return value from Connector uploadPersonalNote to present correct message.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1555,12 +1555,11 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
                 @Override
                 public void call() {
                     final IConnector con = ConnectorFactory.getConnector(cache);
-                    if (con.supportsPersonalNote()) {
-                        con.uploadPersonalNote(cache);
-                    }
+                    final boolean success = con.uploadPersonalNote(cache);
                     final Message msg = Message.obtain();
                     final Bundle bundle = new Bundle();
-                    bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT, CgeoApplication.getInstance().getString(R.string.cache_personal_note_upload_done));
+                    bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT,
+                            CgeoApplication.getInstance().getString(success ? R.string.cache_personal_note_upload_done : R.string.cache_personal_note_upload_error));
                     msg.setData(bundle);
                     myHandler.sendMessage(msg);
                 }
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -719,6 +719,7 @@
     <string name="cache_personal_note_upload">Upload</string>
     <string name="cache_personal_note_uploading">Uploading personal note</string>
     <string name="cache_personal_note_upload_done">Personal note uploaded</string>
+    <string name="cache_personal_note_upload_error">Error uploading personal note</string>
     <string name="cache_personal_note_upload_cancelled">Personal note upload cancelled</string>
     <string name="cache_description">Description</string>
     <string name="cache_description_long">Long Description</string>


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1553,13 +1553,18 @@ private void uploadPersonalNote() {
...
+                        final boolean success = con.uploadPersonalNote(cache);
+
+                        final Message msg = Message.obtain();
+                        final Bundle bundle = new Bundle();
+                        if (success) {
+                            bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT, CgeoApplication.getInstance().getString(R.string.cache_personal_note_upload_done));
+                        } else {
+                            bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT, CgeoApplication.getInstance().getString(R.string.cache_personal_note_upload_error));
+                        }
+                        msg.setData(bundle);
```
samueltardieu: I would have simply written:

```java
  bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT, CgeoApplication().getInstance().getString(success ?
    R.string.cache_personal_note_upload_done : R.string.cache_personal_note_upload_error));
```

to avoid repeating the same code.

samueltardieu:
Please use a header line such as

```
fix #5339: no feedback no personal note upload
```

in the commit message, so that a simple "git log --pretty=oneline" gives meaningful information.

samueltardieu:
ok to test

main/src/cgeo/geocaching/CacheDetailActivity.java
pstorch: Yes, it's better this way.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1553,13 +1553,14 @@ private void uploadPersonalNote() {
...
                     final IConnector con = ConnectorFactory.getConnector(cache);
                     if (con.supportsPersonalNote()) {
-                        con.uploadPersonalNote(cache);
+                        final boolean success = con.uploadPersonalNote(cache);
+
+                        final Message msg = Message.obtain();
+                        final Bundle bundle = new Bundle();
+                        bundle.putString(SimpleCancellableHandler.MESSAGE_TEXT, CgeoApplication.getInstance().getString(success ? R.string.cache_personal_note_upload_done : R.string.cache_personal_note_upload_error));
+                        msg.setData(bundle);
+                        myHandler.sendMessage(msg);
```
SammysHP: If this method is called and the connector does not support personal notes, no message would be sent, thus the handler would wait infinitely.

IMHO a message should be sent in any case, either with a success or a failure message (like in this patch) or with a "not supported" message as a dirty way of handling this situation in a case that shouldn't happen (because the button to upload the note should not be visible unless it is possible).

pstorch: In CacheDetailActivity the upload button is only shown if the connector supports it.
See line 1474

pstorch: But I can add a Message "just in case" the disabling didn't work. Should I?

SammysHP: IMHO it would be a safer behavior. What do you think, @samueltardieu?

pstorch: What about simply calling `myHandler.cancel()` in the else case?

rsudev: I would suggest a 'not supported' message. This handles the unlikely case in a smooth and understandable fashion without too much effort.

rsudev:
ok to test

main/src/cgeo/geocaching/CacheDetailActivity.java
pstorch: A general "Not supported" message or  specific one like "Upload of personal note not supported"?

rsudev: I'd rather be specific.

pstorch:
I've added a "Upload of personal note not supported" in the case where the upload button could be pressed but the upload is not supported by the connector.
This is hard to test, because the upload button usually prohibits it. I've temporarly removed this check (not commited though) to see the new error message.

samueltardieu:
I don't think we need defensive programming at the point where we schedule a localized error message for a condition that can never happen. In which case could the button be enabled *and* `con.supportsPersonalNote()` be false? Isn't the same condition used to enable the button?

In this case, I'd be in favour to crash plain and simple, this is a logic bug in the application and I'd prefer to know about it from the crash traces.

pstorch:
Enablind the upload button is done with this condition:
```
    if (cache.isOffline() && ConnectorFactory.getConnector(cache).supportsPersonalNote()) {
                personalNoteUpload.setVisibility(View.VISIBLE);
```
So technically it can't happen.

pstorch:
Any agreement what we should do? 
Keep this dialog just in case, silently close the handler or ignore this condition?

samueltardieu:
Since `con.supportsPersonalNote()` is necessarily true, I'd remove the `if` altogether.

pstorch:
Ok, I've removed the `if` completely.

Bananeweizen:
thanks

PR #5381, opened 2016-01-13
Fix #5377: "refresh all" on "all caches" list

### LINKED ISSUE #5377: Do not simply hide the "Refresh all" option on the "All caches" list
...but make it depending on the sum of caches on that list.
Quite often user complain that they are unable to refresh all saved caches. Reason is always, that this menu item is hidden on the all caches list indepently of the number of caches on that list.

This "feature" was initially introduced to avoid unnecessary refresh of a large amount of caches.
However I have seen many cases where users have something like e.g. 6 lists with 5 caches each and would like to refresh all 30 on the all caches list which wont work but require to seperately refresh each list.

I instead suggest to use a hardcoded limit for the amount of caches on the "All caches" list where the "refresh all" option is hidden. My suggestion would be >500 caches and not to simply hide the option but display a info toast when it is selected. On other lists there should stay no restriction.

Another option would be to allow it on all lists but only present a warning if the amount of caches is >500.

Opinions?

pstorch:
As discussed in Issue #5377 I enabled the "refresh all" menu item on the "all caches" list.
I added a Dialog with a warning, that this might cause high network traffic. The same Dialog applies for concrete Lists.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1462,5 +1462,6 @@
     <string name="pq_as_cache_list">View cache list</string>
     <string name="pq_download_and_import">Import Pocket query</string>
     <string name="pq_my_finds">My finds</string>
-    
+    <string name="caches_refresh_all_warning">This action may lead to high network traffic. Are you sure you want to batch refresh all caches?</string>
+
 </resources>
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -659,7 +659,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             setVisible(menu, R.id.menu_create_list, isOffline);
 
             setVisible(menu, R.id.menu_sort, !isEmpty && !isHistory);
-            setVisible(menu, R.id.menu_refresh_stored, !isEmpty && (isConcrete || type != CacheListType.OFFLINE));
+            setVisible(menu, R.id.menu_refresh_stored, !isEmpty && isOffline);
             setVisible(menu, R.id.menu_drop_caches, !isEmpty && isOffline);
             setVisible(menu, R.id.menu_delete_events, isConcrete && !isEmpty && containsPastEvents());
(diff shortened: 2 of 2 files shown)


SammysHP:
Have you tested if it puts all caches in a single list? I'm still the opinion that it had something to do with lost list assignment when we introduced this.

pstorch:
Oh, didn't know that. I'll check.

Lineflyer:
mmh...now that you say I tend to remember something like that.

pstorch:
Ok, created 2 Lists (one with 5 caches and another with 12 caches).
After "Refresh all" on "All caches" all 17 caches were refreshed and every cache was still in the list where it was before.

Lineflyer:
Sounds good.
Another ¨problem¨ was, that IMHO we collect the list to be refreshed at run time...which leads to quite some delay after starting the action before the refreshing starts on huuuge lists. But thats the same also know for huge single lists. So not a problem for this patch.

SammysHP:
> and every cache was still in the list where it was before.

Great!

Maybe we should show the dialog only if more than 50 or 100 caches are affected. For small lists it is no problem to refresh all.

pstorch:
> Maybe we should show the dialog only if more than 50 or 100 caches are affected. For small lists it is no problem to refresh all.
If the user has a bad connection even 20 might hurt. 

On the other hand the time estimation is shown in the progress dialog and the user can cancel at any time.

LeXeRLV:
This is great innovation - I have been waiting for it very long time!
Usually I refresh a lot amount of caches in different lists (such as My
Found caches, Mystery solved caches with final location, some trails in
different lists, archived caches and so on),still I really usually refresh
them all, but to not break my listings do it separately for each list - it
is horrible.

It would be great if cgeo developers add some functionality like refresh
list <all caches> - refresh all caches shown on map through "show as list"
command.
Also don't want use store all offline, which I see on the map, bcs then it
asking me to store them all in new(or choosen) list, but I just want to
refresh them all but also let them stay in list they are.

-- 
Best regards,
LeXeR_LV

rsudev:
Thanks for your contribution!

PR #5387, opened 2016-01-17
Update IntelliJ settings

rsudev:
I started to use IntelliJ as well, but the current settings seem not to be aligned with the current project.
As I am not really experienced with IntelliJ, I'd like someone to cross-check that these are correct and don't create issues.

### DIFF
--- a/cgeo-calendar/cgeo-calendar.iml
+++ b/cgeo-calendar/cgeo-calendar.iml
@@ -81,8 +81,6 @@
       <excludeFolder url="file://$MODULE_DIR$/build/intermediates/rs" />
       <excludeFolder url="file://$MODULE_DIR$/build/intermediates/symbols" />
       <excludeFolder url="file://$MODULE_DIR$/build/outputs" />
-      <excludeFolder url="file://$MODULE_DIR$/build/reports" />
-      <excludeFolder url="file://$MODULE_DIR$/build/test-results" />
       <excludeFolder url="file://$MODULE_DIR$/build/tmp" />
     </content>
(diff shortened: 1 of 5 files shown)


kumy:
@rsudev please find in commit #1b02bfb the diff I get too, They are very similar to yours. As the changes only touch the .iml files, I think it's safe to merge.

rsudev:
@samueltardieu will also atke a look this week. As IntelliJ is still flipping around my project files in one ore two lines occasinally, we hopefully come to a version that is stable to use and doesn't get into the way of other commits with unrelated changes.

Bananeweizen:
Actually I've given up on trying to make those files usable by everyone. I typically have 10 to 12 IDEA settings files different than in git, just after checking out. The differences I have seen are caused by
* different plugins installed in IDEA (and their settings being stored in very central files()
* platform differences (e.g. I see line ending differences on Windows for the Linux files of @samueltardieu )
* developer specific settings being included (the SDK settings of samuel surely don't fit to my Android SDK installation, but are part of some settings file)

So I'm fine with merging any change there, I put the files into an ignored change set group afterwards.

kumy:
The default Idea configuration doesn't include a rule for unit tests. PR #5394 propose a common configuration for our c:geo tests. I think it should also be included by default.

kumy:
Also, the build tasks for `cgeo-contacts` and `cgeo-calendar` are not working from our current config ; complain about `Default activity not found` ; should we move them from `Launch default Activity` to `Do not launch Activity` or `Launch` with `cgeo.contacts.ContactsActivity` / `cgeo.calendar.CalendarActivity` ?

Issue #4267, opened 2014-08-29
Option to download PQ from list


mucek4:
Currently we have PQ reader, that displays 20 at the time.
The most seen usecase I see from users is. They make PQ, download it, import it and then "refresh all" as PQ just does not contains all the informations a normal cacher needs in c:geo.

The request is. Have an option to directly download all caches from PQ reader. Not just 20 at the time.

samueltardieu:
The code from @pstorch is merged. Two minor issues: the prompt text looks a bit long and goes out of the box on my Nexus 5, and the success message at the end references `downloadpq.ashx` which should not be exposed to the user.

pstorch:
Maybe the prompt text can be shortened. The context (pocket queries) should be clear at this point. 
What about:
Title = "Download available" 
Message = "View cache list or import?" 
Buttons = "View list" and "Import" 

The success message is a little bit buried. The name of the PQ is last know when creating the intent for the CacheListActivity. Then only the uri is passed. 
The success message uses the getSourceDisplayName() from ImportGpxZipAttachmentThread which returns LastPathSegment from the uri. 
The ideal solution would be to pass the name along, but would cause some more changes along the call hierarchy. A quick fix would be to filter getSourceDisplayName() for http uris.

Bananeweizen:
Since I have converted the selection dialog to a list, it would now be possible to have dedicated buttons for preview or download directly on each list item, thereby avoiding the first dialog completely.

Lineflyer:
Will relocate the open improvement to a new issue.

PR #5393, opened 2016-01-17
Improve usability of pq download activity

pstorch:
According to the comment from @Bananeweizen at #4267 I made use of the new PocketQueryListActivity to get rid of the awkward dialog.
I made the download icon appear on the right side of the list item, if the pq is downloadable.
A click on the rest of the list item shows the cache list as before. 
No need for the extra dialog anymore.


samueltardieu:
Please do not include any change related to languages other than the default. It will lead to merge errors as it is managed from Crowdin.

Can you amend your patch?

pstorch:
Sure. I'll amend it this evening.

pstorch:
I've removed the language changes and found two missing final modifiers.

samueltardieu:
Merged, thanks.

Issue #5396, opened 2016-01-18
Storing some caches from search result lead to a "refresh all" popup


kumy:
When using `store Offline` menu from `CacheListActivity`, a popup called `Refresh all` is shown. I'm not refreshing any caches, but storing new ones. This is a bit confusing.
The message is displayed even if I choose to store only one cache.
I think it's related to PR #5381.

```
Subject: Refresh all
Message: This action may lead to high network traffic. Are you sure you want to batch refresh all caches?
```

Bananeweizen:
Same here. Just opened the nearest list and tried storing them.

Lineflyer:
Yes, I overlooked that while testing. But at least it does not block the function...just a not matching warning.

pstorch:
Sorry didn't know that the "refresh All" menu item is reused for "store offline".

Lineflyer:
AFAIR this was solved but autoclose failed due to missing # in commit message.

PR #5397, opened 2016-01-18
Fix 5396: removed popup for "store Offline"

pstorch:
Fix for #5396.
Sorry I overlooked this case when implementing the refresh All. The menu item "refresh all" is reused for "store offline".

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -743,14 +743,19 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
                 invalidateOptionsMenuCompatible();
                 return true;
             case R.id.menu_refresh_stored:
-                Dialogs.confirmYesNo(this, R.string.caches_refresh_all, R.string.caches_refresh_all_warning, new DialogInterface.OnClickListener() {
-                    @Override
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -743,14 +743,20 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
             case R.id.menu_refresh_stored:
-                Dialogs.confirmYesNo(this, R.string.caches_refresh_all, R.string.caches_refresh_all_warning, new DialogInterface.OnClickListener() {
-                    @Override
-                    public void onClick(final DialogInterface dialog, final int id) {
-                        refreshStored(adapter.getCheckedOrAllCaches());
-                        invalidateOptionsMenuCompatible();
-                        dialog.cancel();
-                    }
-                });
+                final boolean isOffline = type == CacheListType.OFFLINE;
```
rsudev: Is it really worth it, to introduce an intermediate variable just for this check?

pstorch: No, you're right. I'll remove it and put it directly into the if statment.

samueltardieu:
Merged manually, thanks.

PR #5405, opened 2016-01-19
Fixed #5400: refresh warning only when gt 100

### LINKED ISSUE #5400: Refresh all warning also for low amount of caches on list or selected caches
With version 2016.01 19
I consider this as small issue to be fixed with the next bugfix release

pstorch:
I added an if statement to show the refresh warning message only when more than 100 caches are selected (in the list).
The `case R.id.menu_refresh_stored:` section (line 746) grew so big, which I didn't like. So I refactored the code a bit.  I hope it's ok this way.
Please check.

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -127,6 +127,7 @@ import rx.schedulers.Schedulers;
 public class CacheListActivity extends AbstractListActivity implements FilteredActivity, LoaderManager.LoaderCallbacks<SearchResult> {
 
     private static final int MAX_LIST_ITEMS = 1000;
+    private static final int REFRESH_WARNING_THRESHOLD = 100;
 
     private static final int REQUEST_CODE_IMPORT_GPX = 1;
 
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -1133,6 +1123,20 @@ private String getDisplayName(final Uri uri) {
     }
 
     public void refreshStored(final List<Geocache> caches) {
+        if (type == CacheListType.OFFLINE && caches.size() > REFRESH_WARNING_THRESHOLD) {
+            Dialogs.confirmYesNo(this, R.string.caches_refresh_all, R.string.caches_refresh_all_warning, new DialogInterface.OnClickListener() {
+                @Override
+                public void onClick(final DialogInterface dialog, final int id) {
+                    refreshStoredConfirmed(caches);
+                    dialog.cancel();
```
samueltardieu: Is this `dialog.cancel()` happening soon enough if another dialog (for cache selection) is created from within `refreshStoredConfirmed()`? Can there be an interaction there?

pstorch: Interesting question. I copied the Dialog code from somewhere else in cgeo without fully understanding it.

According to:
https://developer.android.com/guide/topics/ui/dialogs.html#DismissingADialog
> When the user touches any of the action buttons created with an AlertDialog.Builder, the system dismisses the dialog for you.

So is this extra `dialog.cancel()` necessary at all? I don't know the exact lifecycle of the AlertDialog. Maybe the automatic closing is done after `onClick()` has been called. Then it might be necessary to cancel the dialog before continuing the processing to prevent any further Dialogs to stack on top of it.

Should I put it before `refreshStoredConfirmed(caches);`, to be on the safe side?

pstorch: We should keep it the way it is.

I debugged a little bit. 
The next dialog coming up is the ProgressBar with the time estimation. This comes up perfectly on top of the warning dialog.

A problem might arise, when we close the dialog before starting the refresh job. Then the dialog disapears immediately, but now it is possible for an impatient user to click on some cache listing before the progress dialog comes up. This causes the cache listing to open and the progress dialog of the refresh process opens "in the background". He can't see it and can't "cancel" the job as long as the cache listing is on top.

samueltardieu:
This patch should be based onto the `release` branch so that it can be included into the next bugfix release. I'll rebase it manually before putting it onto `master` for testing.

samueltardieu:
As a side note: could you please make your commit message imperative rather than descriptive? This is the style we use in cgeo: "fix #xxx" rather than "fixed #xxx".

samueltardieu:
This has been manually rebased and merged. Thanks for your contribution.

rsudev:
IIRC 'fixed' will also fail to trigger the auto-close mechanism of github

pstorch:
ok, will use 'fix' instead of 'fixed'

samueltardieu:
@rsudev "fixed" will work for closing issues (https://help.github.com/articles/closing-issues-via-commit-messages/)

PR #5413, opened 2016-01-21
Fixed #3221: gpx import encoding problem

### LINKED ISSUE #3221: PQ-import dont work if "include name" is enabled at gc.com and the name contain a german umlaut
Hope this is not a duplicate, - i found nothing in the old issues.

I created a PQ named "Rügen Auswahl" at gc.com and marked the option " Include Pocket Query name in download file name" there. 
But I was not able to import it to c:geo. I get the Error Message "Bad bite at 10".

 It worked with two workarounds: 
- open the zip and rename the gpx-files so that the umlauts are gone
- create a PQ without this name-option.

Here the not-working file: http://db.tt/WUkVQ9Hu

pstorch:
Fix for #3221: replaced java ZipInputStream with Apache Commons Compress ZipArchiveInputStream.


pstorch:
Are the build failures caused by my change?

rsudev:
It looks rather like a fail in accessing a bunch of caching sites (you should be able to follow te link and see the test results without login)

samueltardieu:
It would be great if you could add a test (which fails without your changes and succeeds with your changes).

pstorch:
I had to grant Jenkins access to my github account. Then I can see the failing tests.
Looks like some Okapi Problems.

pstorch:
@samueltardieu I'll look into it.

pstorch:
I'm a little bit confused. Maybe it's already too late.

First I struggled with reproducing the error. I created a new PQ with Umlauts, but this time I got no error. The Umlauts where escaped with HTML entities.
It took some time till I found out. It's not only a PQ with Umlauts. The PQ needs to be based on a Bookmarklist with Umlauts. 

Ok, now I got a new PQ with this CP 437 encoding problem. I added it to the raw resource directory as pq_cp437.zip and added a test method in GPXImporterTest. But I'm not able to execute the tests in this class. It always results in 
> no tests were found
> Empty test suite.

The other test classes (e.g. GPXParserTest or LocParserTest) work fine.
I don't know what's wrong here.

Can one of you execute the GPXImporterTest class? Can someone imagine what might be wrong?

Bananeweizen:
I tested your changes locally and it works fine for an otherwise failing PQ. Regarding your local tests: I'm only aware that the dropdown setting in the build variants window of AndroidStudio should be set to "Android Instrumentation Tests", to run this test on your local emulator.

Bananeweizen:
Just as a side note: You are using a different mail adress in your commits than for you github user account. That leads to you not being shown as a cgeo contributor in the github statistics: https://github.com/cgeo/cgeo/graphs/contributors. I guess that is not intendend, so you may want to register the commit mail adress in your account settings as additional mail adress.

pstorch:
@Bananeweizen thanks for the info. The wrong email was configured locally.

Issue #4686, opened 2015-02-16
First Waypoint shows text of second waypoint


dl1ely:
Sometimes (unfortunately i cannot reproduce it), the text of the second waypoint is duplicated into the first for saved caches. Screenshot for GC4E28K:
![screenshot_2015-01-27-12-18-09](https://cloud.githubusercontent.com/assets/427752/6210507/bd36f306-b5d0-11e4-9b51-795955b55f38.png)

„Der Yogi bewacht C und D“ belongs to the second WP, Stage 1. Text for first WP (Start) is gone. This is different from #3420. The error is permanent and persists across refreshes of the saved cache. Also WP detail view shows the wrong text.

Unfortunately, the error appears infrequently, and the circumstances are unclear to me. I noticed it 4-5 times during the last months, but never was annoyed enough to open an issue. 

Version 2015.02.16-NB, but i suspect the error occurs during saving of the cache, which was quite a long time ago now, and thus happened with an older version.

Lineflyer:
Is it possible that the owner updated the waypoint amount and/or content between your initial save date and the next refresh?
If yes, this is AFAIR a known issue (have to search it first).

Lineflyer:
Here we go: #3650  (Please doublecheck)

dl1ely:
I have no way to find out whether the WPs did change, sorry.

So we close here. Sorry for not finding #3650 on my own.

Bananeweizen:
Can you please export that cache as GPX (as it is now, before trying anything) and send it to our support mail? If I read the cache website right, the first waypoint has a note (about Buchstabenwert...) and that should surely override the existing note when updating the cache.

Lineflyer:
@Bananeweizen If you want to try debugging this (and/or #3650) please let me know. I can help my modifying one of my caches if needed.

dl1ely:
Is there a way to export a single cache? I only see that for lists?

rsudev:
You can use selection mode in the list

dl1ely:
Thanks, i mailed the export…

dl1ely:
I just peeked quickly into the .gpx: The waypoint has the correct text there. But in c:geo it is still wrong. Just a display problem?

dl1ely:
[...]

PR #5417, opened 2016-01-22
Fixed #5366: add ID to own-waypoints prefix

### LINKED ISSUE #5366: Waypoint description added automatically
Create a new waypoint (e.g. stage of a multicache) with a description. Create another waypoint without a description (description field is empty). After the waypoint was stored, it has the description of the first waypoint. Edit the description of the second waypoint.  Create a third waypoint without a description. It will get the description of the first waypoint as well. Create another waypoint with a description. It will keep its description.

Master from today morning.

rahu:
Merging new and existing waypoints needs a unique waypoint prefix.

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -82,7 +82,7 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
 
     private ProgressDialog waitDialog = null;
     private Waypoint waypoint = null;
-    private String prefix = "OWN";
+    private String prefix = "";
     private String lookup = "---";
     private boolean own = true;
     List<String> distanceUnits = null;
--- a/main/src/cgeo/geocaching/models/Waypoint.java
+++ b/main/src/cgeo/geocaching/models/Waypoint.java
@@ -142,11 +142,12 @@ public class Waypoint implements IWaypoint {
         return cachedOrder;
     }
 
+    @NonNull
     public String getPrefix() {
         return prefix;
     }
 
-    public void setPrefix(final String prefix) {
+    public void setPrefix(final @NonNull String prefix) {
         this.prefix = prefix;
         cachedOrder = ORDER_UNDEFINED;
     }
--- a/main/src/cgeo/geocaching/utils/Formatter.java
+++ b/main/src/cgeo/geocaching/utils/Formatter.java
@@ -201,7 +201,7 @@ public final class Formatter {
         if (waypointType != WaypointType.OWN && waypointType != null) {
             infos.add(waypointType.getL10n());
         }
-        if (Waypoint.PREFIX_OWN.equalsIgnoreCase(waypoint.getPrefix())) {
+        if (waypoint.isUserDefined()) {
             infos.add(CgeoApplication.getInstance().getString(R.string.waypoint_custom));
         } else {
             if (StringUtils.isNotBlank(waypoint.getPrefix())) {


main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -201,7 +201,7 @@ public static String formatWaypointInfo(final Waypoint waypoint) {
         if (waypointType != WaypointType.OWN && waypointType != null) {
             infos.add(waypointType.getL10n());
         }
-        if (Waypoint.PREFIX_OWN.equalsIgnoreCase(waypoint.getPrefix())) {
+        if (waypoint.getPrefix().toUpperCase().startsWith(Waypoint.PREFIX_OWN.toUpperCase())) {
```
rsudev: I feel not too comfortable with assuming that getPrefix() never returns null. Could you establish that by adding the necessary NonNull annotations?

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -539,6 +539,9 @@ protected Void doInBackground(final Void... params) {
                     finishHandler.sendEmptyMessage(SAVE_ERROR);
                     return null;
                 }
+                if (own) {
+                    waypoint.setPrefix(waypoint.getPrefix()+String.valueOf(cache.getWaypoints().size()));
```
rsudev: Seems I never finished that comment 4 days ago:
Here I think the prefix is not guaranteed to be unique. If you first add two waypoints, than delete the older one and add another, I assume you end up with a duplicate prefix again.

main/src/cgeo/geocaching/models/Waypoint.java
```diff
@@ -142,11 +142,12 @@ private int order() {
         return cachedOrder;
     }
 
+    @NonNull
     public String getPrefix() {
         return prefix;
     }
 
-    public void setPrefix(final String prefix) {
+    public void setPrefix(final @NonNull String prefix) {
```
rsudev: Did I overlook this 4 days ago, or did you amend the commit?

rahu: I pushed a new commit fixing this. Did github send a notifiaction about adding a commit changing a existing merge request?

rsudev: Not sure if github sent a notification, but looking at the PR again, the ordering of comments and commits also suggest that you changed the commit after my initial comment.
The seem to have changed how they present that, so probably therefore I got confused somewhat.

rahu:
I changed the way to find a unique prefix.

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -539,6 +539,20 @@ protected Void doInBackground(final Void... params) {
...
+                            try {
+                                final int suffix = Integer.parseInt(w.getPrefix().substring(3));
+                                if (suffix > max) max = suffix;
+                            } catch(final NumberFormatException ignored) {
+
+                            }
+                        }
+                    }
+                    waypoint.setPrefix(waypoint.getPrefix()+String.valueOf(max+1));
+                }
```
rsudev: While thinking about this implementation I also scanned our code and cross-checked with gc.com, in order to better understand what we should be doing here.
First of all is the prefix meant to have only to digits or letters, as it replaces the 'GC' part in the geocode when creating a waypoint name, especially for gpx export (PKXXXXX, S1XXXXX) (was not aware of that either).
Therefore already code existed, that was meant to make prefixes unique within the cache (and adhere to the length limit), see GpxSerializer.java,202ff and Geocache.java,1383ff and 1350ff
This was 'damaged' when fixing #4686 but I think the fix to that fix should also adhere to the conceptual limits of the prefix (so cross-checking gpx ex- and import seems a good idea)
On a side-note is there also a gap in this implementation, as you compare the prefix of the other waypoints with "own" (we have a constant for that), but then use the previous prefix to construct the new one (implicitly assuming that "own" was assigned here when constructing the waypoint)

rahu:
I found some time to take a closer look at this. I think the main problem is setting the prefix to "OWN" in EditWaypointActivity. If a prefix is set the existing methods to assing a unique prefix do nothing. So I removed setting the prefix to OWN and use waypoint.isUserDefined instead of testing the prefix starts with "OWN".

At the moment I'm unsure if GpxSerializer lines 230:233 
```
if (StringUtils.isBlank(wp.getPrefix()) || StringUtils.equalsIgnoreCase(Waypoint.PREFIX_OWN, wp.getPrefix())) {
   maxPrefix++;
   wp.setPrefix(StringUtils.leftPad(String.valueOf(maxPrefix), 2, '0'));
}
```
are obsolete.

rsudev:
This version doesn't compile (Jenkins):
```
EditWaypointActivity.java:107: error: cannot find symbol
prefix = waypoint.getPrefix();
```
The prefix variable is needed for the editing case for sure, just initializing it to an empty string (and assigning it while saving!) should do the trick.
WRT the GpxSerializer, I would see it as a fall-back option. There might still be old waypoints around in some databases...

rahu:
Don't know what was wrong yesterday. I would swear I compiled an tested this. :-/ 
However - I fixed the compile error and tested again. Waypoint adding without doubling the note is still working as expected.

rsudev:
Thank you very much for your contribution!

PR #5418, opened 2016-01-22
Fix 5406 and 4970 progress indicator LogCacheActivity / LogTrackableActivity

kumy:
While digging for fixing those two issues, I've found that things around `ActionBarListActivity` could be factorized. I've started by doing that, then fix 5406&4970  issues.
Do you agree with such changes?


main/src/cgeo/geocaching/activity/AbstractActivity.java
```diff
@@ -139,6 +135,9 @@ protected void onCreate(final Bundle savedInstanceState, @LayoutRes final int re
     }
 
     private void initializeCommonFields() {
```
rsudev: If you re-purpose the method, you should change its name as well

rsudev:
I support such refactoring activities in general, but was today merely 'flying over' your changes here. For some more comments from my side, I need another more thorough look through the changes.

kumy:
@rsudev I've updated the PR according to your remark about renaming the function.

kumy:
@rsudev or others, any objection to merge into master?

PR #5423, opened 2016-01-24
fix #5414: import complete message for PQ import

### LINKED ISSUE #5414: Import complete message for PQ import
When importing a PQ I get this message:

> 447 Caches von downloadpq.ashx importiert

It should contain the PQ name instead of the download URL.

pstorch:
This is a proposal on how to fix #5414.
Instead of extracting the SourceDisplayName from the URI, I used the Name of the GPX file within the ZIP Archive. 
Because we have to deal with two encodings (CP 437 and HTML entities) I added code to cleanup the HTML entities, too.

I extended the `testImportGpxZipAttachmentCp437()` test method and added another `testImportGpxZipAttachmentEntities` with an example of HTML entities encoded Umlauts.

Unfortunately I still can't execute the GPXImporterTest class. My build variants drop down shows `Android Instrumentation Tests`. Any other idea or hint is welcome.


pstorch:
Do I see it correctly that also CI is not executing the GPXImporterTest?
http://ci.cgeo.org/job/cgeo%20pull%20request/41/testReport/cgeo.geocaching.files/

Bananeweizen:
Yes, I've disabled it during the creation of the gradle build (see gradle file of main module), cause it always failed. I've not checked whether it works again, or what exactly is different when testing under gradle.

pstorch:
Thanks, that was the reason why the GPXImporterTest wasn't executed locally.
I took some time to investigate the failures in this Test class. It had some threading issues.
I had to instantiate the TestHandler with a Looper and add an appropriet Constructor to the CancelableHandler super class as well. Then I could execute the Test locally but saw the "sometimes" failing tests. Sometimes this, sometimes that. That led me to concurrency issues in the TestHandler.
I guess the terminationMessage wasn't transfered across the Threads anymore.
I first experimented with adding sleep() calls. So I could stabilize the tests, but that was not a good solution.
Finally I decided to send a final explizit TERMINATION_MESSAGE in the `runImportThread()` method and check for it in the `waitForCompletion()` method.
Now the tests are stable on my system.
I would be interessted if this also stabilizes the tests on the CI server.
I kept the exclude in the main gradle file, so you can decide to give it a try.

pstorch:
I would like to see if the CI Server can execute the GPXImporterTest.
Let me try it. I'll amend the commit and enable this test class.

rsudev:
GPXImportTest was successful

pstorch:
Ok, `GPXImporterTest` works on CI, too.
The failing test is not according to my changes `cgeo.watchdog.WatchdogTest.testTrackableWebsites`.

rsudev:
Thanks for your contribution!

PR #5426, opened 2016-01-26
fix #5415: PQ import dialog not intuitive

### LINKED ISSUE #5415: The improved dialog for PQ import is less intuitive
2016.01.21-NB

This is for sure a subjective impression, but for me the previous implementation with showing a dialog after clicking on a PQ name was ¨failsafe¨ and intuitive. WIth the new implementation users might miss to understand that clicking on the name and clicking on the icon triggers different things.
I don't know why Ihave this feeling, maybe because the icon does not really look clickable. I would honestly prefer the old implementation.
Opinions on that?

pstorch:
I'm not an expert in UI design, but I wanted to give it a try.
What do you think about this design change? Now they are 'real' buttons and the options should be more clear to the user.

### DIFF
--- a/main/res/layout/pocketquery_item.xml
+++ b/main/res/layout/pocketquery_item.xml
@@ -12,6 +12,22 @@
     android:paddingTop="7dip"
     tools:context=".ui.PocketQueryListAdapter">
 
+    <Button
+        android:id="@+id/cachelist"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:drawableLeft="@drawable/abc_ic_search"
+        android:gravity="left|center_vertical"
+        android:layout_alignParentRight="true"/>
+
+    <Button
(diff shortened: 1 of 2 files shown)


rsudev:
I am also no UI expert :)
What I like in this change is the clear interaction hint to the user. What I do not like so much is the ragged left the titles form. I think putting the buttons to the right will work better.

pstorch:
Good idea, let's compare it with buttons to the right.

pstorch:
Yes, buttons to the right looks better.

SammysHP:
This does not follow any UI guideline, but c:geo does not follow any UI guideline at all, so it's not a problem at the moment. ;)

What happens when you tap on the title? The same as tapping on the magnifying glass?

pstorch:
Nothing. The listeners are attachted to the buttons only.

PR #5433, opened 2016-02-03
fix #5431: last generated, days remaining for PQ list

### LINKED ISSUE #5431: Pocket Query list: show last generated and remaining days
The PQ list on gc.com has a column with "last generated" and "remaining days".

(https://geoclub.de/forum/viewtopic.php?p=1243198#p1243198)

pstorch:
Read the "last generated Date" and the "days remaining" from the download PQ table.

I've put both values next to the cache count below the PQ name:

Maybe a separator between these three values would be good. A "-" or a "," maybe. I didn't like the parantheses around the "days remaining" like in the original webpage.

Any opinions?


rsudev:
For a seperator you could look at the cache lists - there we have already some kind of dot in the middle. Otherwise the look is OK IMO (didn't check the code yet)

pstorch:
I found the middle dot separator and some Formatting routines.
Looks better, I think:


The parsing got a little bit awkward, because I found out that sometimes there is only a "last generation" date without "days remaining" in parenthesis. Couldn't get the logic into one regex.

pstorch:
I spent some more time to clean up the pattern matching code. Now it's one regex which also handles the optional "days remaining" in parenthesis.

main/src/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1088,13 +1075,36 @@ public int compare(final PocketQuery left, final PocketQuery right) {
...
+
+            int daysRemaining = 0;
+            try {
+                final String daysRemainingString = matcherPocket.group(6);
+                if (daysRemainingString != null && !"last".equals(daysRemainingString)) {
+                    daysRemaining = Integer.parseInt(daysRemainingString);
+                }
+            } catch (final NumberFormatException e) {
+                Log.w("GCParser.getDownloadablePocketQueries: cannot parse PQ days remaining");
+                e.printStackTrace();
```
SammysHP: Please add it to the log.

main/src/cgeo/geocaching/connector/gc/GCConstants.java
```diff
@@ -172,8 +172,8 @@
     final static Pattern PATTERN_USERTOKEN = Pattern.compile("userToken\\s*=\\s*'([^']+)'");
     final static Pattern PATTERN_LIST_PQ = Pattern.compile(Pattern.quote("(") + "(\\d+)" + Pattern.quote(")") + ".+?guid=(.+?)\".*?title=\"(.+?)\"");
 
-    // downloadable PQs,     <a href="/pocket/downloadpq.ashx?g=b1390cb5-88f3-4fcc-b5e9-f3786c30f773&src=web">
-    final static Pattern PATTERN_LIST_PQ_DL = Pattern.compile("downloadpq.ashx\\?g=([a-zA-Z0-9-]+).*?>\\s*(\\d+)\\s*</td>");
+    // downloadable PQs,
+    final static Pattern PATTERN_LIST_PQ_DL = Pattern.compile("downloadpq\\.ashx\\?g=([a-fA-F0-9-]+)[^>]*>\\s*([^<]*)<\\/a>\\s*<\\/td>\\s*<td[^>]*>[^<]*<\\/td>\\s*<td[^>]*>\\s*(\\d+)\\s*<\\/td>[^<]*<td>\\s*([^<(]*)(\\(([^\\s]+)[^<]*)?<");
```
SammysHP: This looks too complex. I'd prefer a simple regex that matches the whole line and then further searches for the additional information (or a complete regex, but more robust against changes in the HTML code).

main/res/values/strings.xml
```diff
@@ -683,6 +683,13 @@
         <item quantity="other">%1$d caches</item>
     </plurals>
 
+    <!-- days remaining -->
+    <plurals name="days_remaining">
+        <item quantity="one">one day remaining</item>
+        <item quantity="other">%1$d days remaining</item>
+    </plurals>
+    <string name="last_day_available">last day available</string>
```
SammysHP: Which is zero, right?

main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -254,4 +255,25 @@ public static String formatMapSubtitle(final Geocache cache) {
...
+    public static String formatPocketQueryInfo(PocketQuery pocketQuery) {
+        if (!pocketQuery.isDownloadable()) {
+            return StringUtils.EMPTY;
+        }
+
+        final List<String> infos = new ArrayList<>(3);
+        final int caches = pocketQuery.getCaches();
+        infos.add(caches >= 0 ? CgeoApplication.getInstance().getResources().getQuantityString(R.plurals.cache_counts, caches, caches) : StringUtils.EMPTY);
+
+        infos.add(pocketQuery.getLastGeneration() != null ? Formatter.formatShortDateVerbally(pocketQuery.getLastGeneration().getTime()) : StringUtils.EMPTY);
```
SammysHP: Does this generate separators if the date is not available?

main/src/cgeo/geocaching/connector/gc/GCParser.java
Bananeweizen: Actually, Log.w(...) should have an overloaded version that takes the exception as additional argument.

main/src/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1088,13 +1075,36 @@ public int compare(final PocketQuery left, final PocketQuery right) {
...
+                lastGeneration = GCLogin.parseGcCustomDate(lastGenerationString);
+            } catch (final ParseException e) {
+                Log.w("GCParser.getDownloadablePocketQueries: cannot parse PQ last generation date");
+                e.printStackTrace();
+            }
+
+            int daysRemaining = 0;
+            try {
+                final String daysRemainingString = matcherPocket.group(6);
+                if (daysRemainingString != null && !"last".equals(daysRemainingString)) {
```
Bananeweizen: Please use StringUtils instead of the direct .equals() method. StringUtils is null safe and the first part of condition vanishes.

main/res/values/strings.xml
pstorch: I had zero in the first place, but zero isn't recognized for some languages like english, because they say it's the same as plural.
See: https://stackoverflow.com/questions/17261290/plural-definition-is-ignored-for-zero-quantity

main/src/cgeo/geocaching/connector/gc/GCConstants.java
pstorch: Parsing HTML with Regex isn't a good thing at all. I would prefer a DOM parser, but this would be a bigger refactoring.
I'll try to split it into smaller Patterns. Or should I make a proposal with a DOM parser?

main/src/cgeo/geocaching/utils/Formatter.java
pstorch: Yes, I should check for a null date before adding the item to the List.

main/src/cgeo/geocaching/connector/gc/GCParser.java
pstorch: Then it would change the meaning. I only can parse an Integer, when the String is not null and not equals "last". If I use (!StringUtils.equals()) it would be true when the String is "last" and when it is null. But I need not null to parse the int. 
Anyway, maybe this code changes when I rework the Regex Pattern.

main/src/cgeo/geocaching/connector/gc/GCConstants.java
pstorch: Some time ago I helped with the pocket query creator app and there the Jericho HTML Parser is used. That would be more stable than using Regex Patterns.

BTW: do you know this? https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454
:)

main/src/cgeo/geocaching/connector/gc/GCParser.java
pstorch: Agree.

pstorch:
Besides fixing the other comments I replaced the overly complex Regex.
I wanted to give an HTML Parser a try. Just to evaluate how it's different to Regex parsing.
If you don't like it, no problem. I still have the Regex Version around and can continue working on that (breaking it down for example).

I'm not sure about the proper error handling here. Checking each element selection would clutter the code with lots of error checking statements. Surrounding the whole parsing block with a try catch might not be specific enough.

pstorch:
Redesigned the error handling. Give feedback to the user, if there was a problem while reading the pq page.

samueltardieu:
ok to test

samueltardieu:
ok to test

PR #5444, opened 2016-02-13
Fixed some issues related to @NonNull

UniQP:
I only replaced org.eclipse.jdt.annotation.NonNull with android.support.annotation.NonNull if the superclasses already have an android.support.annotation.NonNull annotation, but maybe it's a good idea to replace them all?!  I can adapt the patches in such a way, if desired.

### DIFF
--- a/main/src/cgeo/geocaching/files/ProgressInputStream.java
+++ b/main/src/cgeo/geocaching/files/ProgressInputStream.java
@@ -1,5 +1,7 @@
 package cgeo.geocaching.files;
 
+import android.support.annotation.NonNull;
+
 import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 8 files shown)


Bananeweizen:
let me check locally, if the replacement of all annotations leads to issues in Eclipse or not...

Bananeweizen:
When I configured Eclipse to use the support annotations it completely failed to do a correct analysis. So as long as we are still partly developing with Eclipse, I suggest to not change the other annotations, as IDEA can use the Eclipse annotations just fine.

We should recheck this after Eclipse Neon has been released (June), as the null analysis can be configured to use secondary annotations in Eclipse Neon.

UniQP:
@Bananeweizen Can you recheck the NonNull annotations with Eclipse Neon?

Bananeweizen:
I've just converted all the NonNull annotations to android.support.annotation.NonNull for testing. Interestingly enough, there is a difference. Eclipse @NonNull can be applied to variable declarations, while the android one is not allowed there. We should wait for every Eclipse developer switching to Neon before deleting these lines (as pre neon will probably throw errors there).

Bananeweizen:
replaced eclipse annotations by android annotations now. I know rsudev and me are on Neon already, if someone else is using eclipse, you should switch now. :)

samueltardieu:
I know those are just annotations, but I don't like the idea of having `android.*` annotations in code which is independent from Android (from example `RxUtils.java`). This doesn't help in separating the concerns.

PR #5453, opened 2016-02-16
fix #5439: update header coords for letterbox and wherigo

### LINKED ISSUE #5439: Allow to update Header-Coordinates for Letterbox and Wherigo Caches
The finals of a Letterbox and Wherigo Caches could have other coordinates as the header coordinates from their listings. So it should be possible to enter these coordinates so that g:geo treat them as header coordinates. At the moment it is also possible to send these coordinates to geocaching.com as corrected coordinates for the listing. These seems to be a breach in the website. 

The impementation of this feature request is a simple one-liner in the source file EditWaypointActivity, in the current revision the line 207 have to be changed.

pstorch:
Enabled updating of header coords in cgeo for Letterboxes and Wherigos in EditWaypointActivity.


samueltardieu:
ok to test

PR #5458, opened 2016-02-20
Small cleanups and improvements.

UniQP:
Various cleanups/improvements, mostly in main/src/cgeo/geocaching/storage/DataStore.java.


UniQP:
Since the changes are already merged, I will create a new pull request for the isNotEmtpy change (it seems that this is not the only occurrence).

The other change is fine in my opinion.

PR #5459, opened 2016-02-20
Use more precise regular expressions for GC/TB codes.

UniQP:
The codes do not contain the letters I, L, O, S, and U.


kumy:
I'm curious, where did you get this information?

SammysHP:
Maybe it is better to have the definition of the expression only one time.

SammysHP:
@kumy 
https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/connector/gc/GCConstants.java#L210

samueltardieu:
So the failing test must be fixed as well, I guess that was a dummy geocode.

PR #5465, opened 2016-02-21
fix #5451: locus export limit, api update

### LINKED ISSUE #5451: Locus export limit?
Can someone with more insight on this feature help to get an answer on this support mail:

Hallo,
I have small problem if I want to show or export them into Locus (Menu / Show on map / Locus or Menu / Show on map / Export to Locus).
If I use it for ie 10 caches all is OK.
If I use it for ie 200 caches, I haven't tab Listing in Locus.

---

I guess there is some limitation in the amount of data we can hand over?!

pstorch:
I've removed the previous locus-api-4.0.jar from libs folder and replaced it with the source from locus as described here: http://docs.locusmap.eu/doku.php?id=manual:advanced:locus_api:installation

I've removed the .hg directories, because I thought it migth not be a good idea to mix hg with git.

The main change was in the AbstractLocusApp class. Changing of the packages and class names.
I've removed the cursor API to locus. It seams that this is not supported anymore. I couldn't find samples.
The way to handle bigger Lists (> 1000 Caches) is to use a intermediate file and send the reference to it.

With this chane I could export a List of 1060 Caches to Locus.

I don't know if my changes to the gradle settings files are correct and compliant with your way of doing it. So please have a look.


SammysHP:
Is there another way to get the dependency? I don't like to have another library project in our repository.

pstorch:
There is no "official" release of the jar available. We could build the locus-api.jar on our own and put it into the libs folder as before.

rsudev:
This sounds like a feasible idea. We also have a mapsforge repo to build our variant of the library. I suggest something similar for the Locus Api

pstorch:
Ok, can someone create the new Repo for it?

pstorch:
Sorry, hit the wrong button.

SammysHP:
The repo is already there – you'd cloned it. ;) We just need a proper jar file (or better: a dependency in gradle which downloads it automatically).

pstorch:
Misunderstanding from my side.

After some further searching I found the locus-api jar files. Not available via gradle, so I added them into the libs folder.

main/build.gradle
```diff
@@ -486,8 +487,8 @@ task downloadLibs(type:Download) {
             mavenLib('org/apache/commons','commons-lang3','3.4'),
             mavenLib('commons-io','commons-io','2.4'),
             mavenLib('com/jakewharton','butterknife','7.0.1'),
-            mavenLib('org/jsoup','jsoup','1.8.3'),
             mavenLib('com/squareup/okhttp3', 'okhttp3', '3.1.2')
+            mavenLib('org/jsoup','jsoup','1.8.3')
```
samueltardieu: Missing a comma on the previous line

pstorch:
merge error fixed

samueltardieu:
Merged, thanks!

PR #5467, opened 2016-02-22
Replace Apache HTTP client with OkHttp

samueltardieu:
This brings us a lot of advantages:

  - no more Android specific code in Network
  - support for SNI (server name indication) to connect to HTTPS hosts
    using virtual hosts
  - support for HTTP/2 (for Google static maps)
  - easier configuration

At this time, everything has been migrated to OkHttp, but code will be
need to be cleaned up a bit more over time.

Twitter OAUTH mechanism seems to be broken. It may be due to the use
of HTTP/2 by Twitter servers and needs to be investigated over the
following days.


Lineflyer:
Just an idea:
At the moment `master` looks quite stable to me, only some more things to be tested.
As this is a rather huge change, we might consider copying `master` to `release`before merging it in order to speed up the release cycle for the changes done last weeks and have more time for testing the new HTTP libraries.

Of course I don't know if there is still some work in progress on `master` which might need to be completed at first.

samueltardieu:
I agree, `master` looks stable. I'd prefer not to have any change there before merging this PR, because this has been a lot of work and I don't want to handle large merge conflicts.

@Bananeweizen, @SammysHP, @rsudev, @kumy: don't hesitate to (quickly) provide feedback on this PR and on pulling `master` into `release`.

main/build.gradle
```diff
@@ -491,4 +489,4 @@ task downloadLibs(type:Download) {
             mavenLib('org/jsoup','jsoup','1.8.3')
```
Bananeweizen: please also add it here for the Eclipse build

main/src/cgeo/geocaching/network/NameValuePair.java
```diff
@@ -0,0 +1,20 @@
+package cgeo.geocaching.network;
+
+public class NameValuePair {
```
Bananeweizen: you know android.util.Pair?

main/src/cgeo/geocaching/network/Network.java
```diff
@@ -57,40 +50,43 @@
...
-    private final static HttpParams CLIENT_PARAMS = new BasicHttpParams();
-
-    static {
-        CLIENT_PARAMS.setParameter(CoreProtocolPNames.HTTP_CONTENT_CHARSET, CharEncoding.UTF_8);
-        CLIENT_PARAMS.setParameter(CoreConnectionPNames.CONNECTION_TIMEOUT, 30000);
-        CLIENT_PARAMS.setParameter(CoreConnectionPNames.SO_TIMEOUT, 30000);
-        CLIENT_PARAMS.setParameter(ClientPNames.HANDLE_REDIRECTS, true);
-    }
+    private final static OkHttpClient OK_HTTP_CLIENT = new OkHttpClient.Builder()
+            .connectTimeout(30, TimeUnit.SECONDS)
```
Bananeweizen: since the configuration has different settings for read and write timeout here, would it make sense to think about different values? e.g. have a higher value for write, assuming it is more important to save user input (logs, field notes, ...).

Bananeweizen:
Looks okay to me, since at many places the code is still structurally identical. At work my automated checkin tooling would have complained about not concatening single character strings (but character literals instead) and not to catch Exception, but RuntimeException.

samueltardieu:
ok to test

PR #5478, opened 2016-02-23
fix #5477: crash when adding cache to watchlist

### LINKED ISSUE #5477: Crashing when trying to add cache to watchlist
- go to cache details 
- press button "add to watchlist"

C:geo restarts and return to home page 
Cache is not added to watchlist 

Cache should be added to watchlist 

Yes

###### Attach system information if available
--- System information ---
Device: ONE E1003 (OnePlus, OnePlus)
Android version: 5.1.1
Android build: ONE E1003_14_160122
Cgeo version: 2016.02.23-NB-891182e
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
System language: sv_SE
Geocaching sites enabled:
 - geocaching.com: logged in (Inloggning OK) / PREMIUM
Installed cgeo plugins: calendar

Noticed the bug in 2016.02.22 NB

pstorch:
It fixes all POST request without parameter body.

Other examples where this happend: favorites and ignorelist

### DIFF
--- a/main/src/cgeo/geocaching/network/Network.java
+++ b/main/src/cgeo/geocaching/network/Network.java
@@ -164,15 +164,13 @@ public final class Network {
             builder.url(urlBuilder.build());
         } else {
             builder.url(uri);
+            final FormBody.Builder body = new FormBody.Builder();
             if (params != null) {
-                final FormBody.Builder body = new FormBody.Builder();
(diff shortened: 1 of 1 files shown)


samueltardieu:
Thanks.

PR #5481, opened 2016-02-24
fix #5472: error reading pq list, when no downloads available

### LINKED ISSUE #5472: Error reading pocket query list
Since this evening i get an error when I try to retrieve a pocket query. Im on 2016.02.23-NB. Am I the only one? Is it gc related, or has it something to do with cgeo?

pstorch:
Handling of empty download PQ table. Checking available table cells and download link.

I don't have an empty download table at the moment. If someone is able to double check, would be good.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCParser.java
@@ -1079,7 +1079,15 @@ public final class GCParser {
             }
 
             final Elements cells = row.select("td");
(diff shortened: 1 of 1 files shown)


samueltardieu:
Is an empty table something worth a warning? Looks more like DEBUG material to me than WARNING one.

pstorch:
Yes, you are right. The empty table is somewhat expected. Changed it to Log.d().

PR #5482, opened 2016-02-24
Use try-with-resources statements.

UniQP:
The feature is supported since Java SE 7.

### DIFF
--- a/main/src/cgeo/geocaching/files/AbstractImportGpxZipThread.java
+++ b/main/src/cgeo/geocaching/files/AbstractImportGpxZipThread.java
@@ -31,8 +31,7 @@ abstract class AbstractImportGpxZipThread extends AbstractImportGpxThread {
         Collection<Geocache> caches = Collections.emptySet();
         // can't assume that GPX file comes before waypoint file in zip -> so we need two passes
         // 1. parse GPX files
-        final ZipArchiveInputStream zisPass1 = new ZipArchiveInputStream(new BufferedInputStream(getInputStream()), ENCODING);
-        try {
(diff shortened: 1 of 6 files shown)


samueltardieu:
I think the older versions we support only have Java 6. It needs checking.

Bananeweizen:
http://developer.android.com/tools/sdk/eclipse-adt.html

> Added support for Java 7 language features like multi-catch, try-with-resources, and the diamond operator. These features require version 19 or higher of the Build Tools. Try-with-resources requires minSdkVersion 19; the rest of the new language features require minSdkVersion 8 or higher.

Shouldn't be a problem therefore.

samueltardieu:
Well, given that we build cgeo with `minSdkVersion 9`, that certainly would be a problem.

UniQP:
The code base already uses the diamond operator as one of Java 7 features:
git grep "<>()" | wc -l
201

samueltardieu:
The diamond operator is a compiler feature. It does not generate any new opcode and thus runs on JVM 6. Same goes with multi exceptions catch. However, try with resources require a new opcode in the JVM AFAIK, as do lambdas in Java 8 for example.

SammysHP:
> Try-with-resources requires **minSdkVersion 19**; the rest of the new language features require minSdkVersion 8

Bananeweizen:
Okay, I misread my own quoted text, since I totally focused on the build tools version. :/

@UniQP Android was created by implementing the Java 6 API from scratch. Therefore you can write Java 6 code, and the compiler will get this working. However, you cannot use any specific language feature or API method of newer Java versions, unless the Android tooling guys explicitly support that. While you are using the Java language, you are not programming against a Java runtime.

PR #5488, opened 2016-02-27
Fix several typos

UniQP:


Bananeweizen:
Thanks for the cleanups. Keep them coming... :)

PR #5494, opened 2016-02-29
fix #5378: add "hide disabled caches" in map menu

### LINKED ISSUE #5378: Add "Hide disabled caches" to map menu
From support mail:

Additionally to the "Hide found/owned caches" the option "Hide disabled caches" should be available also from the map menu.
See mock up screenshot:
current_map_options

pstorch:
Added "hide disabled caches" into the map menu in the same way as "hide own/found caches". 
It is a shortcut to the same setting in the preferences dialog.


samueltardieu:
Nice, thanks.

Lineflyer:
@samueltardieu @pstorch 
Can you elaborate if this feature is contained on the current `release` branch?
I have a user in support who does not see this item in version 2016.03.06.

PR #5504, opened 2016-03-05
Fix tests

samueltardieu:


samueltardieu:
This introduces a new message that will need to be translated, so I'll merge it into `master` only. But it's based on `release`, feel free to merge commit ef3d326 into `release` if we can deal with the translations in time.

Cc @Lineflyer & @rsudev

PR #5506, opened 2016-03-05
fix #5490: fix selection for LinkMovement enabled TextView

### LINKED ISSUE #5490: Scrolling broken after selection
Select some text of the description. Close selection mode. Try to scroll the view up and down. It jumps around.

rahu:
same fix, now based on release branch. Also removed the no longer needed workaround.

### DIFF
--- a/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
+++ b/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
@@ -1,8 +1,14 @@
 package cgeo.geocaching.ui;
 
+import android.support.annotation.NonNull;
+import android.text.Layout;
+import android.text.Selection;
 import android.text.Spannable;
 import android.text.method.LinkMovementMethod;
+import android.text.method.Touch;
+import android.text.style.ClickableSpan;
 import android.view.MotionEvent;
+import android.view.View;
 import android.widget.TextView;
 
 /**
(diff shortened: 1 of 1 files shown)


rahu:
I modified the code and added @kumy 's findings.

Bananeweizen:
I believe that removing the old fix in IndexOutOfBoundsAvoidingTextView isn't right, since it covers another unrelated issue: https://code.google.com/p/android/issues/detail?id=35466

rahu:
I don't see the overloaded `onSelectionChanged` workaround in this issue?

Am 6. März 2016 08:23:34 MEZ, schrieb Michael Keppler <notifications@github.com>:
>I believe that removing the old fix in IndexOutOfBoundsAvoidingTextView
>isn't right, since it covers another unrelated issue:
>https://code.google.com/p/android/issues/detail?id=35466
>
>---
>https://github.com/cgeo/cgeo/pull/5506#issuecomment-192820079

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

Bananeweizen:
Me neither. I'm just saying that the class itself has that issue in its Javadoc comment as the reason why it was introduced. I cannot really judge, if that issue will or will not occur if that method is removed. Since your PR targets the release branch, I suggest a kind of in-the-middle solution: On the release branch I would leave the onSelectionChanged method in there (and only apply your remaining change) and on the master branch we could remove it, allowing for some more testing. Does that sound reasonable?

However, I have no strong opinion here. If you or other developers are confident that the method can be deleted, then I'll accept that.

kumy:
@Bananeweizen solution seems reasonable for me.

rahu:
for me too, I updated the PR.

Issue #5517, opened 2016-03-08
Selection on Cache description does not provide "translation" menu


kumy:
###### Detailed steps causing the problem:
- Open cache GC5ZZZZ
- Slide to description tab
- Long press to activate selection
- Contextual Popup menu appear

###### Actual behaviour after performing these steps:
The contextual menu display "copy/share/select all"

###### Expected behaviour after performing these steps:
The contextual menu should propose "Share / Translate to English / ... [-> "copy/share/select all"]"

###### Version of c:geo used:
2016.03.06

###### Is the problem reproducible for you:
Yes

###### Attach system information if available (see c:geo menu -> About c:geo -> System):
```
--- System information ---
Device: LG-D855 (g3_global_com, lge)
Android version: 5.0
Android build: LRX21R.A1445306351
Cgeo version: 2016.03.06
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: false
Map strategy: auto
HW acceleration: enabled (default state)
System language: fr_FR
Geocaching sites enabled:
- geocaching.com: logged in (Connecté) / PREMIUM
Installed cgeo plugins:  none
--- End of system information ---
```

###### Other comments and remarks:
Works "**most**" of the time on my phone (Android 6.0.1)

```
--- System information ---
Device: m8 (cm_m8, htc)
Android version: 6.0.1
Android build: cm_m8-userdebug 6.0.1 MMB29U 1921d5ed9d test-keys
Cgeo version: 2016.03.06-223b663 developer build
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: false
Map strategy: auto
HW acceleration: enabled (default state)
System language: en_US
Geocaching sites enabled:
- geocaching.com: logged in (Login OK) / PREMIUM
Installed cgeo plugins: none
--- End of system information ---
```

kumy:
On my phone:

Behavior 1:
- long press to select a work -> contextual menu show "Translate..."
- without closing the context menu, long press on another word -> contextual menu show "Copy..."

-> Expected for the second selection -> contextual menu should show "Translate..."

Behavior 2:
- long press to select a work -> contextual menu show "Translate..."
[...]

PR #5518, opened 2016-03-08
Related #5517: Selection inconcistency with translation menu

kumy:
Fix what is described as "Behavior 1" in the issue.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1939,6 +1939,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
     }
 
     private ActionMode mActionMode = null;
+    private boolean selectionModeActive = false;
     private IndexOutOfBoundsAvoidingTextView selectedTextView;
 
     private class TextMenuItemClickListener implements MenuItem.OnMenuItemClickListener {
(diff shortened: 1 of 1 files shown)


kumy:
Anyone willing to check?

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1949,7 +1950,8 @@ public boolean onMenuItemClick(MenuItem menuItem) {
 
     @Override
     public void onSupportActionModeStarted(ActionMode mode) {
-        if (mActionMode == null && selectedTextView != null) {
+        if (mSelectionModeActive == true && selectedTextView != null) {
```
samueltardieu: Please don't compare to `true`. `if (mSelectionModeActive && selectedTextView != null)` is enough.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1968,7 +1970,9 @@ public void onSupportActionModeFinished(ActionMode mode) {
         mActionMode = null;
         if (selectedTextView != null)
             selectedTextView.setWindowFocusWait(false);
-        selectedTextView = null;
+        if (mSelectionModeActive == false) {
```
samueltardieu: `if (!mselectionModeActive)`

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1934,6 +1934,7 @@ public static void startActivity(final Context context, final String geocode, fi
     }
 
     private ActionMode mActionMode = null;
+    private boolean mSelectionModeActive = false;
```
samueltardieu: We don't usually follow the convention `mXXX` for member fields. I'd prefer that we stay consistent and use `selectionModeActive` instead.

kumy:
Thanks ;)

PR #5520, opened 2016-03-09
fix #5516: LinkMovementMethod needs to be aritrary selectable

### LINKED ISSUE #5516: Can no longer select text on cache page.
On my Nexus 7 running 4.1.2 I can no longer select text from the cache page  for translation/copying.
If I long-press on text in the cache page, nothing happens.
However, if I long-press on the hint text, or on log entries, the two blue arrows appear and I can select that text as expected.

Note that it works fine on my Xperia Z3 running 5.1.1

- Step 1
Use a Nexus 7 running 4.1.2
- Step 2
Open a cache page
- Step 3
Long-press on some text on the cache page.

Nothing.
I expect to be able to select text for translation / copying.

###### Version of c:geo used: 

2016.02.03

Yes

```
--- System information ---
Device: Nexus 7 (nakasi, google)
Android version: 4.1.2
Android build: JZO54K
Cgeo version: 2016.03.06
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
HW acceleration: disabled (default state)
System language: en_GB
Geocaching sites enabled:
 - geocaching.com: logged in (Login OK) / PREMIUM
Installed cgeo plugins:  none
```

Works fine under android 5.1.1

rahu:
for details see #5516

### DIFF
--- a/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
+++ b/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
@@ -83,4 +83,9 @@ public class AnchorAwareLinkMovementMethod extends LinkMovementMethod {
             Selection.setSelection(text, text.length());
         }
     }
+
+    @Override
+    public boolean canSelectArbitrarily () {
+        return true;
+    }
 }


kumy:
Works now as expected with the emulator I used to reproduce.
Thanks @rahu

PR #5521, opened 2016-03-09
Fix #87: store caches on multiple lists

### LINKED ISSUE #87: Allow a stored cache to belong to multiple lists
I have a suggestion.  

My caches are organized in lists like To Do, Events, My Hides, etc.  Every time I import a new GPX file that contains a cache that is in one of my other lists, it pulls that cache out of the list and places it into the Stored list along with the others from that GPX.

Would it be possible to allow importing of a GPX without affecting other stored lists?

Thanks much!  :-D

pstorch:
I couldn't resist and started working on this issue #87 .

It's not ready yet, but as this is a bigger change I would like to have the chance to get some feedback.

I introduced a n:m relationship table between caches and lists. Then I started changed all code which was working with the one ListId of the Cache to be able to handle a Set of ListIds.

The main visible difference is in the CacheDetailActivity. 

It's not fully tested and needs some polishing, too. Any feedback is welcome.

If I'm totally wrong here, please tell me, too. Then it was just a programming exercise for me.
Otherwise I'll continue.


samueltardieu:
As you can see on http://ci.cgeo.org/job/cgeo%20pull%20request/121/console some symbols are missing from resources.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -901,18 +906,30 @@ private void storeCache() {
...
         } else {
-            storeCache(StoredList.TEMPORARY_LIST.id, new StoreCacheHandler(this, progress));
+            if (cache.isOffline()) {
+                // cache already offline, just add to another list
+                DataStore.addToList(Collections.singletonList(cache), StoredList.TEMPORARY_LIST.id);
+                new StoreCacheHandler(CacheDetailActivity.this, progress).sendEmptyMessage(CancellableHandler.DONE);
+            } else {
+                storeCache(StoredList.TEMPORARY_LIST.id, new StoreCacheHandler(this, progress));
+            }
         }
```
samueltardieu: This fragment, which is more a repetition of the fragment above, should be refactored to not be repeated.

pstorch: Yes, there is also a similar code fragment in another file. I'll refactor them.

pstorch:
Now it's compiling, but 3 Unittests are failing. I'll look into them. But maybe not this evening.

samueltardieu:
It looks like you might be trying to insert several times the same cache into the same list (probably the default one, where the cache may also be, while you're removing the other list it was into).

samueltardieu:
Using "Store in list" in cache details let the user choose between the standard list and a new list. Other list names are not listed.

samueltardieu:
ok to test

pstorch:
I refactored the similar code blocks mentioned above.

There was a comment in my mail, but not here on github saying: 
> Using "Store in list" in cache details let the user choose between the standard list and a new list. Other list names are not listed.

Is it obsolete? I couldn't reproduce it. When "Store in list" is clicked, only these Lists are available in the popup which the cache is not yet stored in.

Open Issues I see and I'm still working on:

- Presentation of the List of Lists in the CacheDetails (I don't like the "List:" header much)
- recognizing if a Geocache isOffline() uses the getLists() Collection. This is not 100% correct. Result: if you store a cache, remove it from all lists and then again store it on a list, the cache is downloaded again. Maybe using the DB _id would be better here.
- Count of caches on CacheListsDropDown Menu not working correctly for StandardList.
- Review usage of customListIdOffset (I would like to get rid of it)
- In the CacheList Menu a new Item to copy caches to another list would be good

pstorch:
- Added CopyToListCommand
- Adjusted Menu in CacheDetailActivity to correspond with the buttons in the offline_box
- Count of Caches per List in the CacheListDropDownMenu seems to work now (I restarted with a blank DB). But needs more testing with migration from old db schema

samueltardieu:
ok to test

samueltardieu:
ok to test

samueltardieu:
ok to test

samueltardieu:
Have you tried running `gradle testBasicDebugUnitTest`? It looks like this is failing.

pstorch:
Not yet, I'll check.

pstorch:
I fixed the failing UnitTest. Didn't know that there are other tests to be executed by gradle alone. Had to search a little bit in AndroidStudio to find them. :)

I improved the CacheDetails page a little bit. The "drop" button is back, which removed the Cache from all lists. I'm still not 100% happy with the UI, but it's getting better. Any ideas to improve it are welcome.

Regarding the DB migration: I tested again with current master and 3 different Lists. All caches were transferred correctly in the multi list schema. Only going back is not possible. c:geo crashes on startup with a NPE, because it can't downgrade the DB. I think this is a general problem and not related to this change. Should we address this in a separate Issue?

I looked again into the `customListIdOffset` hack. As mentioned above I would like to remove it, but I'm not sure if there is another **clean** solution. The only idea I have is changing the IDs of the PseudoLists to a negative value, but that would be another hack. So because it's working today, I guess we better keep it that way.

samueltardieu:
Concerning the hack: indeed, the clean solution would be to add another field to indicate whether the list is virtual or real, but if you look at it from a mathematical point of view, this is just an extra bit, which can be factored into the actual list id (even though we don't use a separate bit for this and just reserve some values). It might be cleaner to reserve the highest bit for that though, but we have already reserved some extra ids should we need some new virtual lists.

pstorch:
Finally I adjusted the CacheDetails Popup dialog. Because there is not much room, I just added a comma separated List of Lists at the end. There is not "Remove from list" button for the individual lists. There is only a "Drop" Menu item to remove the Cache from all lists. I think this should be enough. If the user needs more detailed functionality he needs to open the full CacheDetailsPage.

Regarding the `customListIdOffset` I doubt that it is worth changing at the moment.

So right now I don't have any open issues left (the failing tests are because geocaching.com has some problems tonight).
I'm waiting for feedback.

samueltardieu:
ok to test

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1378,20 +1359,37 @@ private void updateFavPointBox() {
...
+                // show listListView
+                listListView.setVisibility(View.VISIBLE);
+                final ListListAdapter adapter = new ListListAdapter(CacheDetailActivity.this, cache, this);
+                listListView.setAdapter(adapter);
+
+                final int numberOfItems = adapter.getCount();
+
+                // Get total height of all items.
+                int totalItemsHeight = 0;
+                for (int itemPos = 0; itemPos < numberOfItems; itemPos++) {
```
samueltardieu: Should be `final`

samueltardieu:
I am not able to try it right now, but could you please confirm that:
- refreshing a cache (from its details or from any list) will keep the cache in all lists it currently belongs to?
- importing a GPX (or a PQ) or importing from web will keep the cache in all lists it currently belongs to if the same cache is found during the import?

If those answers are both yes, I propose to merge the issue if there are no objections.

Ping @Lineflyer, @Bananeweizen, @SammysHP, @rsudev, @kumy, @mucek4

pstorch:
Oh no, I missed the GPX Import. Need to fix it.

kumy:
No objection for me. This next feature release will be a long standing one, from people I know... Will do extending tests once merged.

Lineflyer:
In general ok for me.

Just one remark so far:
> Finally I adjusted the CacheDetails Popup dialog. Because there is not much room, I just added a comma separated List of Lists at the end. There is not "Remove from list" button for the individual lists. 

Why at all does the popup need changes here. As of now there is no reference to any lists.

pstorch:
> Just one remark so far:
> 
> > Finally I adjusted the CacheDetails Popup dialog. Because there is not much room, I just added a comma separated List of Lists at the end. There is not "Remove from list" button for the individual lists.

> Why at all does the popup need changes here. As of now there is no reference to any lists.

Now you can hit "store" all the time. If we don't show the lists you can hardly recognize what's happening. The only clue would be the decreasing list of available lists in the list selection popup.

pstorch:
I still can't believe that I missed the GPX (PQ) Import feature. 
Anyway, now this should also work with multiple lists.

While working on it I found two other places which needed attention:
- the "remove all/selected" menu in the CacheListsActivity removed the caches from all lists, not only from the current list.
- the housekeeping code DataStore.cleanIfNeeded() which is triggered on MainActivity startup was still working on the old "reason" column.

main/res/values/strings.xml
```diff
@@ -1483,4 +1496,8 @@
     <string name="invitation_title">Invite friends to use c:geo</string>
     <string name="invitation_message">Try the c:geo app for geocaching</string>
 
+    <string name="cache_list_remove_from">Remove from list</string>
+    <string name="list_list_headline">Lists: </string>
```
Bananeweizen: please trim that string

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1378,20 +1359,37 @@ private void updateFavPointBox() {
...
+                final int totalDividersHeight = listListView.getDividerHeight() *
+                        (numberOfItems - 1);
 
-                // update text
-                final TextView text = ButterKnife.findById(view, R.id.list_text);
-                final StoredList list = DataStore.getList(cache.getListId());
-                text.setText(res.getString(R.string.cache_list_text) + " " + list.title);
+                // Set list height.
+                final ViewGroup.LayoutParams params = listListView.getLayoutParams();
+                params.height = totalItemsHeight + totalDividersHeight;
```
Bananeweizen: why do we need this manual measuring and pixelwise layouting? Generally, I believe one should avoid that.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -229,6 +237,12 @@ public String call(final Cursor cursor) {
             + "title text not null, "
             + "updated long not null"
             + "); ";
+    private static final String dbCreateCachesLists = ""
+            + "create table " + dbTableCachesLists + " ("
+            + "list_id integer not null, "
+            + "geocode text not null, "
+            + "primary key (list_id, geocode)"
```
Bananeweizen: I'm totally not a database expert, so this might be wrong: I would expect that we should have 2 additional indexes on the list_id and on the geocode, because we know for sure that we have a query that ask for all caches of a single list, and that we have a query which asks for all lists a cache is in. Otherwise those always have to do a full table scan, right?

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -1707,7 +1753,7 @@ public static Geocache loadCache(final String geocode, final EnumSet<LoadFlag> l
 
                 if (loadFlags.contains(LoadFlag.LOGS)) {
                     final Map<LogType, Integer> logCounts = loadLogCounts(cache.getGeocode());
-                    if (MapUtils.isNotEmpty(logCounts)) {
+                    if (logCounts != null && MapUtils.isNotEmpty(logCounts)) {
```
Bananeweizen: that is not necessary. the apache utils are null safe

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -2213,6 +2272,11 @@ public static int getAllHistoryCachesCount() {
             selectionArgs = new String[] { String.valueOf(cacheType.id) };
         }
 
+        selection.append(" and geocode in (SELECT geocode FROM " + dbTableCachesLists + " WHERE list_id ");
```
Bananeweizen: just wondering why none of our static code analysis tools complains here (doing string concatenation inside of stringbuilder append calls is not optimal for performance)

Bananeweizen:
I've only done review, no testing yet. A question out of curiosity: All existing menus and commands have been reworked. I'm wondering if we should have an additional context and options menu on lists to delete the selected/all caches from _all_ lists instead of the current. I really don't have an opinion about that myself yet. It might be confusing for users, or it might be helpful for power users. No clue.

SammysHP:
I thought we decided to use tags. Each cache can have none, one or many tags (aka labels) and "lists" will be replaced by a filter (e.g. checkboxes in the current list dropdown).

main/src/cgeo/geocaching/storage/DataStore.java
pstorch: Got tricked by the IDE warning. I'll remove it.

main/src/cgeo/geocaching/storage/DataStore.java
pstorch: Search by list_id should be handled by the combined primary key. But to assist the search via geocode an index would be good on the geocode column, yes.

main/src/cgeo/geocaching/CacheDetailActivity.java
pstorch: I got a problem when adding lists to the listview. They didn't appear because the height didn't recalculate. So I searched and found this solution. I'll look for another solution. Any help/hint would be appreciated.

samueltardieu:
> I thought we decided to use tags. Each cache can have none, one or many tags (aka labels) and "lists" will be replaced by a filter (e.g. checkboxes in the current list dropdown).

This proposal, which works very well for me so far, is an extension of what we have now. We can still change the wording afterwards (list -> tag) and add the new notion (collection of tags -> list) if we want, but the current step is needed anyway.

main/src/cgeo/geocaching/storage/DataStore.java
samueltardieu: If IDE bugs you about it, just add an assertion inside the "then" clause: `assert(logCounts != null);`. Also, if you're using IntelliJ IDEA, you can add a contract to `MapUtils.isNotEmpty` saying that `null` implies `false`, and you won't be bothered about it any longer.

main/src/cgeo/geocaching/CacheDetailActivity.java
pstorch: Found another solution. Replaced the ListView with a LinearLayout and added the Items as children there.

pstorch:
Fixed some review comments.

Open issues (for discussion):
- "Delete all" menu item in CacheListMenu to delete caches completely from all lists. I could add it, but this menu is already quite full. 
- lists vs. tags: I think it's just a naming thing.

rsudev:
list vs tags is indeed 'just a naming thing', but IMO one that can make a difference when it comes to understanding the feature easily.
When you 'store' something on a list, remove/delete becomes somewhat ambiguous and you need to consider namings like 'delete from all lists' to describe a complete removal.
When you organize your caches with tags, the two operations are distinct. You add/remove tags from a cache or store/delete a cache on/from the device.
That said, I have no objections to merge the feature to make it easier to test and gather feedback before the naming question has a 'final' decision.

SammysHP:
What @rsudev said and:

- "All caches" are all caches with or without a tag. Furthermore it might contain caches without tags at all. This makes an "default" tag/list obsolete.
- You can select/deselect tags, e.g. "vacation" and "must-do". This might be possible with "multiple lists" as well, but is more intuitive.
- Tag selections and filters can be mixed via the same UI, they are more or less on the same level while lists and filters are on different levels.

As this is in some way only a naming issue I'm also the opinion to merge this PR and test it. Before the next release we should think about the naming and possible new ways to use the new system (as written above).

samueltardieu:
Let's merge and make it evolve.

samueltardieu:
Thanks for this painful but very useful work @pstorch.

PR #5531, opened 2016-03-13
Use diamond operator.

UniQP:


Lineflyer:
I am not the right one to be able to have any technical comment to that code, however I have something in mind, that the diamond operator was not supported.
Looking at: https://github.com/cgeo/cgeo/pull/5482

Bananeweizen:
thanks

SammysHP:
@Lineflyer As long as we compile with Java 7 this will work.

PR #5545, opened 2016-03-18
fix #5424: changed pq icon to suitecase with magnifying glass

### LINKED ISSUE #5424: Action bar icon for PQ
I just used c:geo in landscape mode and wondered what the icon with the key means. It is the "Pocket Queries" button. IMHO the icon should be replaced with something better.

pstorch:
There seems to be still no con sense about the best action bar icon for the PQ download in issue #5424.
Here is my self made icon. If someone wants to test it.


main/res/drawable/ic_menu_pocket_query.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
rsudev: Can you use vector drawables even for older platforms (8+)? Are they handled by AppCompat or generated as png by the build in such a case?

pstorch: When I created it AndroidStudio told me it will convert it for older platforms automatically. But I guess this is not good enough. We need it on gradle level, too.
I just searched a little bit and found this article: http://android-developers.blogspot.de/2016/02/android-support-library-232.html
It says that Android Support Library 23.2 can handle it, but we are still on 20.0. So at the moment it would be a problem.

I'll change it to png.

samueltardieu: What would we need to use the support library 23.2? (cc @Bananeweizen)

rsudev: According to my research that would mainly mean, that the map activities will have to use our old fake actionbar for all API versions below 23.

pstorch:
I replace the vector image with PNGs. Safer choice I guess.

kumy:
@pstorch May you also publish the svg file to https://github.com/cgeo/cgeo/tree/master/main/project/rawimages ?

pstorch:
@kumy sure, here it is

kumy:
@pstorch Thanks :)

PR #5548, opened 2016-03-19
fix #5535: right selection handle not working

### LINKED ISSUE #5535: Cannot move right selection anchor - Cache Description
- Open a cache
- Slide to Description tab
- Long press a word
- Try to move right anchor
###### Actual behavior after performing these steps:

The anchor jump to left anchor + 1 character
###### Expected behavior after performing these steps:

The anchor should be freely movable
###### Version of c:geo used:

`2016.03.15-b452a33 developer build`

Yes

```
--- System information ---
Device: m8 (cm_m8, htc)
Android version: 5.1.1
Android build: cm_m8-userdebug 5.1.1 LMY48Y a4d1ba3b73 test-keys
Cgeo version: 2016.03.15-b452a33 developer build
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
System language: en_US
Geocaching sites enabled:
- geocaching.com: logged in (Login OK) / PREMIUM
Installed cgeo plugins: none
```

Left anchor move well

rahu:
@kumy you wanted a hack... :-/

this takes the code from android/widget/TextView.java and adjusts the line in selection mode. It's working fine for me.

### DIFF
--- a/main/src/cgeo/geocaching/ui/IndexOutOfBoundsAvoidingTextView.java
+++ b/main/src/cgeo/geocaching/ui/IndexOutOfBoundsAvoidingTextView.java
@@ -58,6 +58,40 @@ public class IndexOutOfBoundsAvoidingTextView extends TextView {
 		}
 	}
 
+	float convertToLocalHorizontalCoordinate(float x) {
+		x -= getTotalPaddingLeft();
+		// Clamp the position to inside of the view.
+		x = Math.max(0.0f, x);
+		x = Math.min(getWidth() - getTotalPaddingRight() - 1, x);
+		x += getScrollX();
+		return x;
+	}
+
+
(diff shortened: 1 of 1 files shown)


samueltardieu:
Independently of the merits of this patch, it doesn't respect our code style (in particular, `final` are missing everywhere).

rahu:
This is only a hack. I'm still searching for the reason why it returns the wrong offset in selection mode while working fine if selection is not active... 

For this hack - which surely should be removed once a clean solution is available - I copied the code from the android sources and added two lines... 

```
if (shouldWindowFocusWait && (line < getLineCount()-1) 
    line+=1;
```

Feel free to close this without merging - I don't want to fix this

Issue #5546, opened 2016-03-19
Lists panel is not updated when editing Personal note


MrYaz:
###### Detailed steps causing the problem:
- Open a cache that is not currently stored in the device
- Go to the Description tab
- Press "Edit" under Personal note
- The message  "The cache has been stored locally" is shown.
- Enter something in the Personal note field and save the changes by pressing the checkmark
- Go back to the Details tab

###### Actual behaviour after performing these steps:
The cache does not seem to be stored in any List since the Lists panel is empty.
You have to refresh the cache details for it to appear in the List view in the List "Stored"

###### Version of c:geo used:
2016.03.19 NB

###### Is the problem reproducible for you:
Yes

###### Other comments and remarks:
Bug related to the newly implemented feature #87

samueltardieu:
It has been fixed by https://github.com/pstorch/cgeo/commit/a5d3196a8072a7e4675666880b885221817bec9f.

PR #5552, opened 2016-03-20
fix #5543, #5546: upload personal note if cache not saved yet

### LINKED ISSUE #5543: Upload of personal note only if cache is saved
- PM-user
- Open a unsaved cache in c:geo
- Edit the personal note

There is no upload button shown as long as the cache is not saved in c:geo

I don't see a reason why we can't upload the note for unsaved caches?! (Not sure if it worked before)
Else the upload button should be shown and the user should be asked to save the cache before uploading the note (like in other places if data is added).

###### Version of c:geo used:
2016.03.15-NB

Yes

pstorch:
Allow edit and upload of personal cache note if cache has not been saved yet.
Refresh DetailsPage after autosave to default list.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1472,7 +1472,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
                 }
             });
             final Button personalNoteUpload = ButterKnife.findById(view, R.id.upload_personalnote);
-            if (cache.isOffline() && ConnectorFactory.getConnector(cache).supportsPersonalNote()) {
+            if (ConnectorFactory.getConnector(cache).supportsPersonalNote()) {
                 personalNoteUpload.setVisibility(View.VISIBLE);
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2353,11 +2354,9 @@ public void call() {
     }
 
     public static void editPersonalNote(final Geocache cache, final CacheDetailActivity activity) {
-        if (cache.isOffline()) {
```
rsudev: Not sure if this change is really needed here. All execution path call ```ensureSaved``` before, so normally it would be ```isOffline```. Should we leave this check as a safeguard?

pstorch: It doesn't hurt, I can bring it back in.

samueltardieu: @rsudev If we are into defensive programming, I would prefer an `assert()` here in this case since this condition *cannot* happen.

samueltardieu:
@rsudev Feel free to add an extra check if you want, I'll merge this in the meantime.

PR #5555, opened 2016-03-21
Move constants to right-hand side of comparisons.

UniQP:


samueltardieu:
Thanks, I really dislike Yoda talk :-)

PR #5566, opened 2016-03-22
fix #5526

### LINKED ISSUE #5526: Own coords not updated in waypoint screen
- Open s saved cache
- Go to the waypoint tab
- Start creating a new waypoint
- Tap on the coords to open the input dialog
- Start moving from your position so that your own coords are changing
- Press the "Own coordinates" button repeatedly to obtain updated coords

The coords are never changing in the input dialog,
You have to leave the dialog and open it again to get an update.

The button "Own coordinates" should update the fields with the actual coordinates.

###### Version of c:geo used:
2016.03.10-NB (but probably since we introduced the new input dialog)

Yes

samueltardieu:


samueltardieu:
Failure is solely due to opencaching uk.

PR #5571, opened 2016-03-22
Use precise minimum cache distance.

UniQP:
See https://www.geocaching.com/about/guidelines.aspx.

### DIFF
--- a/main/src/cgeo/geocaching/maps/CachesOverlay.java
+++ b/main/src/cgeo/geocaching/maps/CachesOverlay.java
@@ -11,6 +11,7 @@ import cgeo.geocaching.connector.gc.GCMap;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.LoadFlags;
 import cgeo.geocaching.location.Geopoint;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/maps/CachesOverlay.java
```diff
@@ -170,7 +170,7 @@ private int calculateDrawingRadius(final MapProjectionImpl projection) {
         final GeoPointImpl itemGeo = mapItemFactory.getGeoPointBase(itemCoord);
 
         final Geopoint leftCoords = new Geopoint(itemCoord.getLatitude(),
-                itemCoord.getLongitude() - 161 / longitudeLineDistance);
+                itemCoord.getLongitude() - 160.9344f / longitudeLineDistance);
```
samueltardieu: Instead of using this constant, I would prefer that you use `528 * FEET_TO_KILOMETER * 1000` (constant coming from `IConversion`, that will be made `static final` in a few minutes but that you can use right away).

UniQP:
I updated the code accordingly.

PR #5572, opened 2016-03-22
Improve checks for Android versions

UniQP:


samueltardieu:
We usually do not use final dots on the end of single line imperative sentences in the short commit message.

UniQP:
I updated the commit messages.

samueltardieu:
Thanks :-)

PR #5573, opened 2016-03-23
fix #1594: inconsistent use of SearchResult error code

### LINKED ISSUE #1594: Inconsistent use of SearchResult error code
There are a couple places in the code where the following is used to determine if an error has occurred `search.getError() != null`. The code assumes that if a non-null value is returned, then an error has occurred.
The problem is that the error can be set to `StatusCode.NO_ERROR` which breaks the logic above.
I only see one spot (in GCConnector.java) that actually sets it to NO_ERROR.

Here are a couple of ways I see to fix this.
 - Expand all checks to check != null and != NO_ERROR
 - Remove all != null checks and only check != NO_ERROR (requires making sure NO_ERROR is always set)
 - remove NO_ERROR and only use null checks

Thoughts?

(I realized this as there is a small bug dealing with being offline and viewing caches in history that have been dropped from a list)

samueltardieu:


samueltardieu:
restest this please

Issue #5093, opened 2015-07-15
Automatically check login after first credentials supplied


Bananeweizen:
For a new installation, we should automatically trigger the "check login" functionality after the login credentials for geocaching.com have been supplied. The login needs to be executed anyway during the next steps to get some data, but it would be good to detect errors early, while the user is still in the settings.

Hauwertlhaufn:
Imho that is a good idea, 'cause many users may think, that if there is no problem after you saved them, any problem afterwards couldn't be 'caused by the login-information.

madankb:
May I work on this request?

Lineflyer:
@madankb I think so if not @Bananeweizen already started something?!

samueltardieu:
This should be done after #5092 or at the same time.

madankb:
Just to confirm, after user supplies credentials for geocaching, Check login should be triggered automatically. I believe same needs to be done for GCVote and Extremcaching. For #5092, please clarify what is single dialog for both username and password. Appreciate your help.

samueltardieu:
Please ask questions you might have about #5092 in the corresponding issue.

kumy:
@Lineflyer assigned for testing

Lineflyer:
Tested with a clean installation of 2016.04.03-NB to see the sequence for new users.
- App jumps to gc.com login screen
- User selects Authorize
- Enters his credentials (or can click "Create an account")
- Press check authentication will perform a login check automatically.

Issue #5092, opened 2015-07-15
Use one dialog for username and password


Bananeweizen:
For services which need username and password, we should use a single dialog combining both. This should help in getting the credentials right for new users.

Lineflyer:
@madankb 
SIngle dialog means, that the user does not need to select one menu item to input his username and just another menu item for the password, but have one dialog which offers two input fields for username and password and can be ended by the user by pressing "Check Login" (see #5093).

If implemented it should be the same method for all services.

kumy:
@Lineflyer assigned for testing

Lineflyer:
Implemented implicitly with the new authentication screens. Closing.

Issue #4986, opened 2015-05-29
Avatar no longer shown on "Check Login" on some devices


Lineflyer:
Strangely with 2015.05.29-RC the user avatar is no longer loaded when using the function "Check Login" in Menu - Settings - Services - geocaching.com.

With production release 2015.03.31 it is working normal (also tested today...so not website change related)
Edit: Test with production release was done on another device (see comment from 03.06.2015).

Log does not show any error.

samueltardieu:
I had noticed this as well some times ago. It may be due to the increase of SDK version.

samueltardieu:
On my Nexus 5 / CM 12.1 / Android 5.1.1, I get the avatar.

Lineflyer:
I happening for me on my Samsung S4, Android 5.0.1

Lineflyer:
When testing this I do see a `304`error in the log, which should mean, that the picture is already locally stored. Still it is not shown to me.

```
09:36:12.597 Debug cgeo 25380  [network-5] GET http://www.geocaching.com/my/
09:36:13.367 Debug cgeo 25380  [network-5] 200 (734 ms) GET http://www.geocaching.com/my/
09:36:14.568 Debug cgeo 25380  [network-1] GET https://img.geocaching.com/user/large/e3afdf64-7188-464e-ac80-8474501d9352.jpg
09:36:16.098 Warning cgeo 25380  [network-1] 304 [Not Modified] (1580 ms) GET https://img.geocaching.com/user/large/e3afdf64-7188-464e-ac80-8474501d9352.jpg
```

If I manually delete it from `/.cgeo/shared` and try again the picture is loaded again (but still not shown):
```
09:38:03.605 Debug cgeo 25380  [network-10] GET https://img.geocaching.com/user/large/e3afdf64-7188-464e-ac80-8474501d9352.jpg
09:38:05.679 Debug cgeo 25380  [network-10] 200 (2028 ms) GET https://img.geocaching.com/user/large/e3afdf64-7188-464e-ac80-8474501d9352.jpg
```

Lineflyer:
Retested this on my device (Samsung S4, Android 5.0.1) with the current market release 2015.03.31. Also on this device the image is requested and stored locally but not shown.

So at least this is no regression but something else device and/or Android version depending.
Removing "Regression label" and changed title.

kumy:
@Lineflyer assigned for testing

Lineflyer:
[...]

PR #5578, opened 2016-03-23
Concistency between Services settings

kumy:
I have reworked some services setttings. Introduced a new `CredentialsAutorizationActivity` based on other activities from the same kind (`TokenAutorizationActivity`, `OAuthAutorizationActivity`) which ask for a user/password.
Connectors for EC, GC, GCVote are upgraded to `CredentialsAutorizationActivity`. Now all connectors share the same UI.
Some issues are fixed, not necessary how it was proposed (I think about #5092 / #5093)
Due to the new UI, avatar is not longer displayed uppon connexion (see also #4986). I wanted to display it directly in the connector settings, but failed to store the image to show it "permanently". Do you think it's necessary? Can someone help here?

screenshot_2016-03-24-00-27-24
screenshot_2016-03-24-00-26-48
screenshot_2016-03-24-00-27-30
screenshot_2016-03-24-00-27-05
screenshot_2016-03-24-00-28-01


samueltardieu:
> Due to the new UI, avatar is not longer displayed uppon connexion (see also #4986). I wanted to display it directly in the connector settings, but failed to store the image to show it "permanently". Do you think it's necessary?

You can use `HtmlImage.getDrawable` with using `HtmlImage.SHARED` pseudo-geocode as the geocode in the constructor. This way, the image will not be refreshed unless you set `userInitiatedRefresh` (for example, when the user checks its credentials).

kumy:
@samueltardieu thanks for the tip.
screenshot_2016-03-25-01-02-36

kumy:
Any objection to merge?

rsudev:
I didn't check the code, but frontend-wise I like the unification.

samueltardieu:
No objection at all, this is a great step in the good direction!

kumy:
So let's merge :)

PR #5594, opened 2016-03-28
fix #5371: dialogs out of shape on MainActivity

### LINKED ISSUE #5371: Wrong window size or text size not matching in some dialogs
While testing the new PQ functions I realized some design glitches on my device, see screenshots attached.
AFAIR there is at least one other dialog with the same problem, but I could not find it.

```
--- System information ---
Device: GT-I9505 (jfltexx, samsung)
Android version: 5.0.1
Android build: LRX22C.I9505XXUHOJ2
Cgeo version: 2016.01.11-NB1-f1ec8b2
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: true
Map strategy: auto
System language: de_DE
Geocaching sites enabled:
- geocaching.com: logged in (Login OK) / Premium
- opencaching.de: logged in (Login OK)
Installed cgeo plugins: calendar, contacts
```

screenshot_2016-01-10-10-56-20
screenshot_2016-01-10-10-58-04

pstorch:
The attribute android:fitsSystemWindows="true" for the MainActivity styles was the culprit.

See:
https://stackoverflow.com/questions/26599805/android-alert-dialog-not-styled-properly-on-lollipop

I can't see any problems after removing this attribute. I'm on Android 5.1. Should be tested with older versions, too.

### DIFF
--- a/main/res/values-v19/themes.xml
+++ b/main/res/values-v19/themes.xml
@@ -3,7 +3,6 @@
     <style name="cgeo_main" parent="cgeo_main.base">
 
         <!-- KitKat's transparent navigation -->
-        <item name="android:fitsSystemWindows">true</item>
         <item name="android:windowTranslucentNavigation">true</item>
     </style>
 </resources>


UniQP:
Typo in the commit message: shap -> shape.

pstorch:
fixed typo

kumy:
no problem for me, but I don't know the impact on older android versions.
Let's merge, we could revert if necessary.
@Lineflyer may you "watch" against possible impacts when you are testing please.

PR #5596, opened 2016-03-28
Add newline at end of files

UniQP:


samueltardieu:
What's the point of this? Changing 72 files for something which is not in our styleguide and does bring little benefit is dubious.

UniQP:
The POSIX standard states that a line should end with a newline character (which is also required by other programming languages like C): http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

I guess for that reason some text editors (in my case: vim) automatically inserts the newline character at the end of files. Thus, when I change some file without trailing newline, I always got two changes: The actual one and the trailing newline change, which in turn requires a bit more effort to create a clean commit. That's why I fixed them.

Natureshadow:
Text files do have to end with exactly one \n character in all POSIX systems, yes.

SammysHP:
Many code conventions include such a rule. This change help to get a consistent formatting. It might have been enough to fix it whenever a file is touched, but I have no objection to this change.

(Do we have any CRLF file?)

UniQP:
main/thirdparty/org/openintents/intents/FileManagerIntents.java is the only java file with CRLF, followed by 14 xml files.

I can create a pull request to fix this, if desired.

SammysHP:
Let's wait for @samueltardieu's OK first. ;)

samueltardieu:
This is ok with me. However, I would prefer that we do *not* modify any file whose content has been stored as-is (such as the data in `tests/res/raw`).

Bananeweizen:
Do we need any additional settings in the Idea/AS settings files to enforce this? Or should I set core.autocrlf=true on my Windows machines? Maybe even core.whitespace?

Otherwise this might come up again in the future.

PR #5608, opened 2016-03-29
fix #5542: exif target navigation to picture context menu

### LINKED ISSUE #5542: Add exif target navigation to picture context menu
- IOpen a gallery tab of a cache containing georeference pictures (Having our new overlay)
- Long press on the picture

Only some "Open with..." options are shown in the menu

As a click on the overlay can "only" trigger the prefered navigation I would suggest to include the full navigation menu offering all methods in the context menu after long press

###### Version of c:geo used:
2016.03.15-NB

This will also make the feature more usable as the user might fail to recognize that clicking the overlay on the picture triggers an navigation action

pstorch:
Added the Navigation menu entry to the image context menu on long click.

### DIFF
--- a/main/src/cgeo/geocaching/ui/ImagesList.java
+++ b/main/src/cgeo/geocaching/ui/ImagesList.java
@@ -55,7 +55,7 @@ import rx.subscriptions.Subscriptions;
 
 public class ImagesList {
 
-    private BitmapDrawable currentDrawable;
+    private ImageView currentView;
     private Image currentImage;
 
     public enum ImageType {
@@ -81,6 +81,7 @@ public class ImagesList {
      * map image view id to image
      */
     private final SparseArray<Image> images = new SparseArray<>();
+    private final SparseArray<Geopoint> geoPoints = new SparseArray<>();
     private final String geocode;
     private LinearLayout imagesView;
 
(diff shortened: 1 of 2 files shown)


main/src/cgeo/geocaching/ui/ImagesList.java
```diff
@@ -240,12 +243,13 @@ public void onCreateContextMenu(final ContextMenu menu, final View v) {
...
-        final ImageView view = (ImageView) v;
-        currentDrawable = (BitmapDrawable) view.getDrawable();
-        currentImage = images.get(view.getId());
+        currentView = (ImageView) v;
+        currentImage = images.get(currentView.getId());
+        menu.findItem(R.id.menu_navigate).setVisible(geoPoints.get(currentView.getId()) != null);
     }
 
     public boolean onContextItemSelected(final MenuItem item) {
+        BitmapDrawable currentDrawable = (BitmapDrawable) currentView.getDrawable();
```
samueltardieu: Missing `final`.

pstorch:
Added the missing final.

I thought about applying more refactoring here, but resisted to not change too much.
I think the Exif parsing should be in ImageUtils and the Geoposition an Attribute in the Image model class. But the Image class is immutable, so that the Geoposition needs to be parsed ahead and feed into the Image.Builder. That I don't like, because it would affect more than the ImageList.
I like to keep it lazy evaluated. Maybe a transient Geoposition Attribute in the Image class which is loaded if some other class calls getGeoposition() on the Image and then cached.

PR #5609, opened 2016-03-29
Use more specific parameter type to avoid casts

UniQP:


kumy:
@UniQP Sorry I reverted this change. It caused Exception. See #5610

UniQP:
Hm, it's also sad that the problem wasn't recognized by the CI.
Is there a possibility to encounter the problem without testing on the emulator or on an actual device?

samueltardieu:
No.

But you should give your changes at least a cursory check on a device (or an emulator at least) before submitting a pull request.

PR #5625, opened 2016-04-03
fix #5598: reset some preferences at phone or SDK change

### LINKED ISSUE #5598: HW acceleration switched off on Android 6.0
There are numerous users on support mail, who have graphical problems and stability problems.
All of them have in common:
- They either have a new device with Android 6.0 or recently upgraded to Android 6.0
- In all system informations I do see that HW-acceleration is switched off (`HW acceleration: disabled (manually changed)`)
- After enabling it, the problems are gone 

The users do say that they did not manually change this setting before.

Is there any problem in the code, which may lead to the effect, that it is disabled on Android 6.0?
Is this setting contained in any (google) backup and might be wrongly restored on a users new device?

##### Version of c:geo used:
2016.03.06.

samueltardieu:


Bananeweizen:
If I get it right, this means that the value used by the app might be reset once the SDK version changes. However, since that that reset value is never stored, the user will never notice. I.e. his preferences still show the same checkbox settings, even though they might be silently ignored. I don't like that kind of non understandable fixes behind the scenes and suggest to store the value on reset.

samueltardieu:
@Bananeweizen Sorry, didn't see your comment before merging: I think there is no problem here as the `SettingsActivity` calls `initDeviceSpecificPreferences()` in its `onCreate()` which explicitly sets the preferences.

Issue #5536, opened 2016-03-15
Performance issue


Lineflyer:
Just a reminder for me to test this and maybe clean install before retesting.

I noticed today with 2016.03.14-NB that cgeo is laggy in responding to clicks.
Especially after pressing the three dot button in cache details or popup may take a few seconds to open the menu.

Bananeweizen:
Seems not reasonable, looking at the latest commits. I'd suspect this to be a local issue.

Lineflyer:
Well I reinstalled without restoring the DB and the specific problem mentioned above does no longer appear, however c:geo is still laggy in usage.
Might still be my device...I will keep an eye on this and eventually close this issue lateron.

Lineflyer:
I still do see a laggy behaviour.
E.g. open the lists from main menu may take more than a second to open although there are only 10-20 caches on the list.

SammysHP:
Just to confirm this:

> (13:57:57) SammysHP: I have the feeling that the lists in c:geo became slow
(13:58:29) SammysHP: when you open the stored caches from the homescreen it takes almost a second until the list with 5 (!) caches is shown

Lineflyer:
@SammysHP  Can you please check the responsiveness of the three-dot menu while on cache details?
I also have some delayed response there on the first access.

SammysHP:
Not sure if there's a delay. But on my first try, the action bar itself had a short delay (the buttons/menu were missing for a second).

Lineflyer:
I do still have this delays every.
The most noticable delay is when I open the lists coming from the main screen. However I have no clue how to analyze this, at least the log is not showing anything related.

pstorch:
A quick performance analysis when clicking on the **stored** button on the main screen:
![selection_003](https://cloud.githubusercontent.com/assets/5598124/14190198/c08f16b0-f791-11e5-9e26-eb649daa46bb.png)

Looks like `loadSpoilers()` and `loadLists()` are the main cpu-time consumers. The call to `loadLists()` is new, that's for sure. But I'm not sure why we need to load **Spoilers** and **Lists** it at this point in time. I try to understand the `LoadFlags`. Maybe creating a special Select with only the data needed in the CacheListActivity?

Lineflyer:
@pstorch Are you experiencing the delay?

i.e.: There should be a noticable difference between first opening of the list (afer new start of c:geo) and leaving and reopening. (Second access is faster).

Lineflyer:
[...]

PR #5628, opened 2016-04-03
an interim solution for #5536

pstorch:
This is not a complete fix, but should reduce the performance issues of #5536 .
I replaced the N-Selects of the `dbTableCachesLists` Table with one select with a `where in` clause.

Now the main remaining performance issue is `loadSpoilers()`.

selection_004


samueltardieu:
Please retest this.

samueltardieu:
The failing test is unrelated to this change. Merging.

PR #5631, opened 2016-04-04
Merge if blocks with the same condition

UniQP:
Maybe, we also want to combine both warnings?


samueltardieu:
No, that's wrong, `list` can be `null` because `call()` returned an exception, so we cannot issue the message `null result` in this case.

UniQP:
Oops, you're right. Sorry.

Issue #4633, opened 2015-01-25
Spoiler picture link missing


Lineflyer:
2015.01.25-NB (also present in 2015.01.20 as reported on support mail).
Happens also in other scenarios where the cache is opened directly from map.
Not sure whether it is a regression, but I think so as I never noticed it before.

- Search for GC5K2A1
- Open the cache details
- Note that the cache has a spoiler pic which is correctly indicated on the description tab
- Open the spoiler picture, is is shown normally. It is also visible on the image tab
- Select Navigate - Map
- Click on the cache to open popup
- Select more details

Result:
The link to the spoiler pic is missing on the description tab and the picture is not shown on the image tab. Selecting "refresh" will show it again until the next test cycle (unless the cache is saved).

Lineflyer:
Corrected the sequence to reproduce above.

samueltardieu:
Please test for inclusion in `release`.

Lineflyer:
Can you tell which areas are modified by the fix?
Why did the problem only show up in this strange procedure?

Understanding this might help for retesting the sourrounding areas.

samueltardieu:
To complete a cache information to display in the info box, an incomplete view was loaded, without the spoilers. Once merged, this view was saved into the database, thus clearing the spoilers.

Lineflyer:
Using the above mentioned scenario I was able to verify the fix in 2015.02.02.-NB.
Up to now I did not see any regressions coming along, so its good to go for inclusion into `release`.

samueltardieu:
Done.

Bananeweizen:
This one has a bad impact on performance, as I just noticed during measurements. It leads to always loading spoilers. Even in the case of the cache list activity, where they are surely not needed.

samueltardieu:
@Bananeweizen Feel free to fine-tune it. However, that would require remembering not to erase unloaded information from the database when updating a cache (from what I've seen, cache is removed from the database then stored again).

PR #5640, opened 2016-04-05
Improves #5536, partly reverts #4633

pstorch:
Removed the LoadFlag.SPOILERS which were introduced in #4633 and causing performance issues in #5536. 
While debugging the Issue #4633 I realized that the Cache was saved twice in the DB. And the second time there were no Spoilers. The code in `DataStore.saveSpoilersWithoutTransaction()` first deletes all existing Spoilers and then inserts the new ones. 
Now I don't delete Spoilers when the Cache currently beeing saved doesn't contain Spoilers (assuming it was not properly loaded). So the Spoilers are kept in the DB and properly loaded in the last Step when the CacheDetailsActivity is loaded again.
One possible downside might be that when the Spoilers where removed completely from geocaching.com and the Cache is refreshed. Then we would keep the old Spoiler(s).

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -56,6 +56,7 @@ import android.database.sqlite.SQLiteStatement;
 
 import java.io.File;
 import java.io.FilenameFilter;
+import java.sql.PreparedStatement;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -1477,14 +1478,15 @@ public class DataStore {
     }
 
     private static void saveSpoilersWithoutTransaction(final Geocache cache) {
-        final String geocode = cache.getGeocode();
-        database.delete(dbTableSpoilers, "geocode = ?", new String[]{geocode});
+        if (cache.hasSpoilersSet()) {
+            final String geocode = cache.getGeocode();
+            final SQLiteStatement remove = PreparedStatement.REMOVE_SPOILERS.getStatement();
+            remove.bindString(1, cache.getGeocode());
+            remove.execute();
 
-        final List<Image> spoilers = cache.getSpoilers();
-        if (CollectionUtils.isNotEmpty(spoilers)) {
             final SQLiteStatement insertSpoiler = PreparedStatement.INSERT_SPOILER.getStatement();
             final long timestamp = System.currentTimeMillis();
-            for (final Image spoiler : spoilers) {
+            for (final Image spoiler : cache.getSpoilers()) {
                 insertSpoiler.bindString(1, geocode);
                 insertSpoiler.bindLong(2, timestamp);
                 insertSpoiler.bindString(3, spoiler.getUrl());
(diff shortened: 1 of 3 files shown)


samueltardieu:
Couldn't we have a way to distinguish between "spoilers not loaded" and "no spoilers exist"? This would let us clear them if they disappear from the site.

pstorch:
I thought about it, but it is a Collection. I didn't want to use NULL as an indicator that it was not loaded and I also wanted to avoid a separate flag for the spoilers.

pstorch:
Maybe keeping the complete LoadFlags in the Geocache model class to indicate what was loaded and what is missing?

samueltardieu:
That would be a solution. Another one would be to create a particular `final` object (to not use `null`) representing an absence of loaded spoilers (compared with `==`).

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -1477,11 +1477,11 @@ public static boolean deleteWaypoint(final int id) {
     }
 
     private static void saveSpoilersWithoutTransaction(final Geocache cache) {
-        final String geocode = cache.getGeocode();
-        database.delete(dbTableSpoilers, "geocode = ?", new String[]{geocode});
-
         final List<Image> spoilers = cache.getSpoilers();
         if (CollectionUtils.isNotEmpty(spoilers)) {
+            final String geocode = cache.getGeocode();
+            database.delete(dbTableSpoilers, "geocode = ?", new String[]{geocode});
```
Bananeweizen: Just wondering if the deletion performance would also benefit from a preparedStatement. (The existing prepared statements were mostly introduced to avoid that the database query cache has an overflow, not because of performance reasons).

pstorch:
I found another way: for waypoints and attributes there is already a mechanism to support lazy loading with a `LazyInitializedList`. I'm now pretty sure, that this is the way to go.

pstorch:
Oh, maybe I was a little bit over confident here. I first thought I could look how it's done with attributes and waypoints and apply this to spoilers. But they are not handled the same way. I stumbled over `Geocache.gatherMissingFrom()` and `Geocache.isEqualTo()` were waypoints are merged and attributes are just added under some conditions. And I guess there are some bugs, too (check for null but they never are, compare with ==).
Have to think about it.

samueltardieu:
IIRC, `isEqualTo` is more of a shortcut. Comparing some fields with `==` there might give false negatives, but it won't have any adverse effect except miss some opportunity not to resave an identical cache into the database.

main/src/cgeo/geocaching/storage/DataStore.java
pstorch: I tried with PreparedStatement, but the saveSpoilers method takes only 0,1 % of the Cache store operation time. But having it is not a bad thing, so I keep it in.

pstorch:
I tried again with LazyInitializedList for spoilers, but that had other implications. It's somehow tied to the Database. 
I came down to simply have a `Geocache.hasSpoilersSet()` method which checks if spoilers == null to indicate that no Spoilers were loaded.

pstorch:
Please test

Lineflyer:
Please test this

Lineflyer:
Ok to test

PR #5648, opened 2016-04-07
fix #5646: store to default list if list selection is disabled

### LINKED ISSUE #5646: Cache not stored if setting for list selection is disabled
- Disable Settings - Offline Data - "Ask for list"
- Open unsaved cache
- Select "Store"

The cache details do not reflect that the cache has been stored, the cache is not stored to the default list

Cache is stored to the default list

##### Version of c:geo used:
2016.04.07-NB1

Yes

We might also decide to remove the "Ask for list" setting as it is contrary to the desired tag/multi-list feature

pstorch:
Temporary List was used instead of Standard List.


Lineflyer:
@pstorch If you think its fine (and the checks are green in fact) feel free to directly merge such small changes if no one objected wihin a reasonable time.

PR #5657, opened 2016-04-10
Fix #5650: Send2cgeo PIN not showing up

### LINKED ISSUE #5650: Send2cgeo PIN not showing up (finish locale)
- Finish language
- Use "Register" in send2cgeo service to obtain a PIN from our server

Instead of the PIN `###` is shown.
sen2cgeo problem

PIN is shown.

##### Version of c:geo used:
206.03.06

##### System information:
Finish language

I suspect the problem might be that the translation is wrong. The english original contains `####` as a placeholder while finish misses on `#`
Can someone confirm this reason, then I will correct it on Crowdin.

kumy:
How this string change should be handled? @rsudev the crowdin strings be should be uploaded then updated in master soon after merging this, isn't it?


rsudev:
@kumy you are right, will do that.

PR #5662, opened 2016-04-11
Avoid protected fields/methods in final classes

UniQP:


Bananeweizen:
That is one of the changes (and code style recommendations) I completely don't like. I always favour explicit (potentially redundant) modifiers at declarations in classes and interfaces. The reason is quite simple: If you move the declaration into another class/interface during a refactoring, then the modifiers are _not the same_ after the refactoring. I have seen to many people running into the exact situation I described above. First someone removed an implicitly set modifier, then someone else moved the declaration and some time later everyone is wondering why the hell those methods are public and non final, though they be final and protected.

UniQP:
@Bananeweizen Are your sure that you commented the right PR? If yes, I don't see why it's better (or worse) to change protected visibility to public than to change default visibility to public.

PR #5663, opened 2016-04-11
Remove tags

UniQP:
The commits should not change the behaviour of the method. However, I'm not convinced that the method behaves like expected (independent of my changes), e.g. it does not remove the tags for "<span><span>...</span></span>". Maybe some unit tests would be helpful?


Bananeweizen:
Unit tests are available for that method. Simply add more unit tests to test your suspicions. And please commit those test cases, even if they should not prove your thoughts.

Issue #5550, opened 2016-03-20
Finalize multi list UI (first step for release)


Lineflyer:
In order to optimize the usability of the possibility to have a cache on more than on list:

Instead of the new multiple buttons for ¨Remove from list¨ couldn`t we just make the list selection dialog checkbox based and keep the remaining UI elements on cache details as they have been before?

IMHO this would have some advantages:
- Moving a cache from one list to another currently got more complicated than before
- No cluttering of buttons on cache detals page
- Simultanous storing to multiple lists possible in one dialog

Doing it like this would also bring it more into the direction of the discussed ¨tag¨ strategy

Bananeweizen:
I'm thinking more along these lines:
![tags](https://cloud.githubusercontent.com/assets/406876/14233161/d557f86a-f9c0-11e5-9232-e717da1c133c.png)

Replace the buttons I used for the tags with something like the tags from here (and allow users to choose a color, when creating a new tag): https://github.com/mcharmas/android-tagview

pstorch:
I think this could be a good solution. 
Should we open a new tab for this or keep it on the Details page. Showing all available Tags could need some space.

SammysHP:
Keep it on the overview page. This shouldn't take more than one or two lines (most users will use only one, maybe two tags). Editing of tags should open a new activity/popup.

Lineflyer:
In general this is a good way forward.
But I would also like to keep it on the overview (cache details) page if possible.
Also the create new tag should in a first approach still be in the current mdoe (like "Create list") at the place where the user is used to it in the UI.

Lineflyer:
I don't know the efforts of effectively changing the list UI into a tag UI.
In order not to have so long release cycles I suggest in a first step to optimize the multi-list as it is now (e.g. with the initially mentioned changes) and later on move forward to a tag based UI.

What do you think?
IMHO the checkbox based selection would make is more usable for now and might be easier to implement on short term.

pstorch:
I think the Checkboxes can improve the usability a lot. But I don't have a clear picture yet.
Having the buttons as they were before is not 100% suitable.
The old buttons were:
- Refresh: is always there -> no problem
- Move: was only shown when the Cache was stored on a List
- Store / Drop: were exchanged depending on current offline state

I think instead of "Move" and "Store" we would have "Lists" or "Manage Lists" which opens the Checkbox based Listselection dialog.
"Drop" is shown when the Cache is on at least one List. (related to #5638)

[...]

PR #5669, opened 2016-04-13
issue-5627: long press on store button moves cache to the target list…

haarspalter:
obviously I still have some challenges in using git, let's try once again

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -913,7 +913,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
                         }
                     }, true, cache.getLists());
         } else {
-            storeCacheInList(StoredList.TEMPORARY_LIST.id, move);
+            storeCacheInList(StoredList.STANDARD_LIST_ID, move);
         }
     }
 
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1,5 +1,76 @@
 package cgeo.geocaching;
 
+import android.R.color;
+import android.app.AlertDialog;
+import android.app.ProgressDialog;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.Typeface;
+import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.Drawable;
+import android.graphics.drawable.LayerDrawable;
+import android.net.Uri;
+import android.os.AsyncTask;
+import android.os.Build;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Message;
+import android.support.v4.app.FragmentManager;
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -904,20 +909,26 @@ private void storeCache() {
                     new Action1<Integer>() {
                         @Override
                         public void call(final Integer selectedListId) {
-                            storeCacheInList(selectedListId);
+                            storeCacheInList(selectedListId, move);
                         }
                     }, true, cache.getLists());
         } else {
-            storeCacheInList(StoredList.STANDARD_LIST_ID);
+            storeCacheInList(StoredList.TEMPORARY_LIST.id, move);
```
pstorch: should be StoredList.STANDARD_LIST_ID, see #5646

main/src/cgeo/geocaching/CachePopupFragment.java
```diff
@@ -199,6 +198,55 @@ public void call() {
...
+                // let user select list to store cache in
+                new StoredList.UserInterface(getActivity()).promptForListSelection(R.string.list_title,
+                        new Action1<Integer>() {
+                            @Override
+                            public void call(final Integer selectedListId) {
+                                storeCache(selectedListId);
+                            }
+                        }, true, cache.getLists());
+            } else {
+                storeCache(StoredList.TEMPORARY_LIST.id);
```
pstorch: should be `StoredList.STANDARD_LIST_ID`, see #5646

rsudev:
Ok to test

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -2940,6 +2939,41 @@ public static void removeFromList(final Collection<Geocache> caches, final int o
...
+        database.beginTransaction();
+        try {
+            for (final Geocache cache : caches) {
+                remove.bindString(1, cache.getGeocode());
+                remove.execute();
+                add.bindLong(1, listId);
+                add.bindString(2, cache.getGeocode());
+                add.execute();
+
+                cache.getLists().add(listId);
```
pstorch: should be preceeded by `cache.getLists().clear()`

pstorch:
Does it interfere with #5550 ?

main/src/cgeo/geocaching/storage/DataStore.java
haarspalter: That is code from the existing moveToList method, having the old and the new list as parameter. I think this with the new list handling, calling this method might not produce the expected result, as it can only handle a single old list.
Actually this is not part of my pull request, but I would say that it would make sense to replace calls to **moveToList(final Collection<Geocache> caches, final int oldListId, final int newListId) ** with **moveToList(final Collection<Geocache> caches, final int listId)**, or remove the calls/code completely if no longer applicable.

pstorch: The existing `moveToList()` does a `cache.getLists().remove(oldListId);` now you want to remove all old lists, that would be equivalent to `clear()`, wouldn't it?

samueltardieu:
What IDE did you use? There are a lot of gratuitous import order changes.

haarspalter:
AS 2.0

Bananeweizen:
I reviewed this change for quite some time to understand what all the diffs are supposed to do. In the end I re-implemented the functionality, since I had a hard time understanding all details of this PR.

PR #5674, opened 2016-04-16
Fix watchdog tests for gc.com

rsudev:
Introduce a special test url, which does not redirect

there might be better possibilities, therefore as a PR

### DIFF
--- a/main/src/cgeo/geocaching/connector/AbstractConnector.java
+++ b/main/src/cgeo/geocaching/connector/AbstractConnector.java
@@ -172,6 +172,12 @@ public abstract class AbstractConnector implements IConnector {
     }
 
     @Override
+    @NonNull
+    public String getTestUrl() {
+        return getHostUrl();
+    }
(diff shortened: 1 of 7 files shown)


kumy:
Couldn t we do something at our network library level?

rsudev:
The network library as such allows following redirects, but the question is, how elaborate we want to make this simple availability test.
Instead of checking the content, we also could just check that we get a sensible response code (200, 302) or so.
From my point of view this was the fastest and easiest way to stop these annoying 'false failures'. If you have another idea, just go forward.

kumy:
When I looked at this some days ago, I end up to something similar than your fix. But I've just overridden `getHost()` to make it return the "valid" url after the redirects.

As I wasn't confident about my fix, I've not pushed it, hopping someone will go further fixing the root cause...

But I agree with you when you say "stop these annoying 'false failures'".
:+1: from me to merge.

PR #5684, opened 2016-04-21
fix #5682: an absent image file is not fatal

### LINKED ISSUE #5682: Images from intentionally surpressed URLs produce error on image tab
With current master I keep running into a hanging app once I swipe to the image tab of a cache.
It is not clearly reproducible up to now and only happening on my tablet but not on my phone.

What happens is, that the screen freezes in the middle of the ¨swiped¨ tabs  with only frozen loading indicators showing up as placeholder instead of the images.
It keeps frozen until I force close c:geo. Until now I dont even get a ANR warning from the system.

Need to investigate a little more unless  someone else is also able to reproduce.

samueltardieu:


samueltardieu:
Retest this please

Lineflyer:
Retest this please

samueltardieu:
Retest this please

PR #5709, opened 2016-05-01
Fix indentation

UniQP:


Bananeweizen:
Yeah, while the Eclipse auto formatter normally does a very good job, it has some bugs, since it was re-implemented from scratch with Eclipse Mars. :(

PR #5713, opened 2016-05-01
Fix typos in comments

UniQP:
A small rebase exercise for @Bananeweizen ;-)

### DIFF
--- a/main/src/cgeo/geocaching/connector/trackable/TrackableConnector.java
+++ b/main/src/cgeo/geocaching/connector/trackable/TrackableConnector.java
@@ -41,7 +41,7 @@ public interface TrackableConnector {
     String getServiceTitle();
 
     /**
-     * Check whether the connector has URLs corresponding the the trackable.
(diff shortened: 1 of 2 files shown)


Bananeweizen:
Thanks a lot. I like the straight line in my local history view. :+1:

UniQP:
Hm, looks like that email address of the commit changed :-(

Bananeweizen:
I think when doing any operation in the web UI, github will use the name and mail of the user as given in the profile for creating new commits, not as given in the original commit. I see a similar thing in that the merges which I create for pull requests (of others) go by "Michael Keppler", while my own commits all go by "Bananeweizen".

SammysHP:
In case of a rebase this is wrong and should be reported to GitHub. (Commits might come from different people, not necessarily from the one who opened the pull request.)

UniQP:
For a linear history you can still fetch the branch of the PR and rebase/push by hand.

PR #5715, opened 2016-05-03
Do not use deprecated method

UniQP:


tests/src/cgeo/geocaching/files/ProgressInputStreamTest.java
```diff
@@ -9,7 +9,7 @@
 public class ProgressInputStreamTest extends TestCase {
 
     public static void testRead() throws Exception {
-        ProgressInputStream stream = new ProgressInputStream(IOUtils.toInputStream("test"));
+        ProgressInputStream stream = new ProgressInputStream(IOUtils.toInputStream("test", (String) null));
```
samueltardieu: I would prefer that you use an explicit encoding rather than `null` which is the platform default. Even ASCII would be a better choice than `null`.

Bananeweizen: To extend on the remark of @samueltardieu: It's always bad if tests run green under a German locale, but not a French one, etc. Therefore we should avoid relying on any changeable platform defaults, unless the data being operated on also depends directly on that platform default. In this specific case, I'm okay with your change, since "test" will be encoded fine with any locale.

samueltardieu: Thanks for developing @Bananeweizen, that was indeed my point entirely. This is fine here whatever the default is (I think even Asian language contain the basic latin plane), but I would not want to see a change in the string made later with consequences differing depending on the default locale.

UniQP: Short remark: The original method already used default encoding:So 

> Converts the specified CharSequence to an input stream, encoded as bytes using the default character encoding of the platform.

So the PR did not make it worse. However, I will update the PR to use UTF-8.

UniQP:
I added UTF-8 as encoding and rebased to current master.

Bananeweizen:
Oh. Seems that github remembered the merge setting from last time. So this time I did not intend to rebase, but still did. Sorry for that.

PR #5723, opened 2016-05-07
Fix off-by-one error

UniQP:


SammysHP:
I'm not sure if this method makes sense. The if condition can never be false because `i` is initialized with `length` and decremented at least once (with this PR it can be false if the last character is not a whitespace). But it also worked before because the `end` in `subSequence(start, end)` is exclusive and thus is limited by `length`. This means that the if is superfluous.

On the other hand this would fail when `length` is zero (and thus `i = -1`):

> IndexOutOfBoundsException - if start or end are negative, if end is greater than length(), or if start is greater than end

UniQP:
The method calls `subSequence(0, i + 1)`, so if `i = -1`, we still get `subSequence(0, 0)`, where `start` and `end` are both non-negative.

I guess the idea of the if statement is to avoid an unnecessary copy of the original string.

samueltardieu:
Let's wait for @Bananeweizen to comment on this.

SammysHP:
@UniQP Yes, you're right.

Bananeweizen:
Actually, I haven't thought of that end condition at all, since I looked that up. If this were original code of mine, it would not be that "C style". :) I just wrote some test cases, and to me it looks like they all pass with both versions of the trimSpanned code. Of course I might have been wrong with the test cases, therefore I will submit them in a moment.

UniQP:
You shouldn't see a difference in the test cases, since the only difference is whether `subSequence` is called if the string is already trimmed. The condition of the if statement was always true in the old version.

Issue #5581, opened 2016-03-25
Switch IntelliJ project format?


rsudev:
The other day I switched to the offered new version of IntelliJ:

IntelliJ IDEA 2016.1
Build #IC-145.258, built on March 17, 2016

It seems to have a different project format with all files stored in the .idea folder only. Should we switch to this version in general? Should I upload these files in addition?

Bananeweizen:
If I get it right, Android Studio 2.0 is also based on that version. However, Android Studio 2.0 is not yet released (1.5 is current).

On the other hand, I always use just "import from gradle" when checking out on a new machine (and I basically ignore all changes in those many IDE related files).

So, I can't really give advice...

pstorch:
I tried Android Studio 2.0 today, but got this build error:

`Error:A problem was found with the configuration of task ':main:packageBasicDebug'. > File ... specified for property 'resourceFile' does not exist.`

Bananeweizen:
I've switched the gradle build to Android 2.0 on Saturday, and at least the gradle tasks I normally use (testBasicDebug and similar) work fine for me. I've never used the non gradle run configurations in AS before, so I can't say whether they changed for worse or not.

Summary: I can work with AS 2.0 currently.

kumy:
For me, on Idea 2016.1, just opening the project and click `sync gradle` removes `.iml` files and create `.idea/` directory structure... This is quiet annoying while commiting to see all those diff not related to the work going on. For me we should migrate.
I've opened a PR #5656 for others to test with their respective IDE.

kumy:
That will not be easy to have a consistent project layout as we all use different tools. Couldn't we simply not share the .idea directory at all? What would be the consequences?

pstorch:
I can help testing. We could remove and gitignore the `.idea` directory.
But I'm not an AS poweruser. I don't have special settings to keep there.

kumy:
I've just tested to copy my workdir, then remove `.idea` directory and `.iml` files. The project can still be successfully imported into Idea. The "non gradle run list" is empty, but this is expected. Users can add configurations themselves if they prefer to use them. So, it seems technically feasible to not distribute Idea project files and add `.idea/` to `.gitignore`.
What's the global opinion here?

Bananeweizen:
[...]

PR #5731, opened 2016-05-10
Removed idea files and folders, see #5581

pstorch:
As discussed in #5581 I removed the `.idea` folder and the `.iml` files and added them to the `.gitignore`.
Android Studion (and I guess IntelliJ) works without them. It might be necessary to import the project once again.
Ignoring these files and folders simplifies switching between branches, commiting relevant changes, merging and rebasing.


samueltardieu:
Please do not merge right away. I think some files should be kept (the inspection profile for example, as well as the style rules). I'll try to check tomorrow.

pstorch:
I compared the `.idea` directory from `master` and this PR after starting AS once.
These files seem to be specific for cgeo: 

- `inspectionProfiles/`
  - `Project_Default.xml`
  - `c_geo_standards.xml`
  - `profiles_settings.xml`
- `runConfigurations/`
  - `cgeo___tests.xml`
- `scopes/`
  - `cgeo_project_sources.xml`
  - `scope_settings.xml`
- `codeStyleSettings.xml`
- `findbugs-idea.xml`

If these files are essential, how are they synchronized between Eclipse and AS?

samueltardieu:
They are not synchronized automatically. Is this even possible?

pstorch:
I don't know an automatic way. That's why I'm asking. Is someone doing it manually?
Can we move some settings into the Gradle build (e.g. the findbugs config), so that we have a common basis?

samueltardieu:
You can keep files starting with "cgeo" and drop others I would say.

samueltardieu:
Retest this please

pstorch:
That might be a compromise. But I would like to understand what they are used for. 
For example `runConfigurations/cgeo___tests.xml` changes everytime I start AS and when I look into it, I don't see something important which is worth sharing.

pstorch:
I tried to keep these two cgeo specific files:

- `inspectionProfiles/c_geo_standards.xml`
- `scopes/cgeo_project_sources.xml`

and ignored anything else.
The result was that after importing the project into AS it deleted both files and restored all others.

At the moment I think we should try ignoring the complete `.idea` folder as initially planned and see how it works out. If anything feels awkward we can revert the change and restore these files anyway.

pstorch:
Any other objections?

Bananeweizen:
I just merged 5 other PRs, but I have too little insight into the AS file structure to give any valuable feedback here. Someone else?

rsudev:
Right now I am testing a little what is volatile and what fixed in the configs of AS and idea, when you just work with them. Perhaps this gives some insight what to keep and what to ignore

rsudev:
My current research supports the concept of this PR.
- idea (latest version) deletes all .iml files in the projects, creates new ones in .idea/modules and keeps fiddling with them on every run. If there is some stable configuration file where code styles and checks can be placed I would need to find out
- AS (latest version) still uses the .iml files in the projects, also fiddling with them on every run. There seems to be a codeStyle configuration in .idea, but how stable that is I don't know (and it is deleted by idea).

I guess at some point AS will again be aligned with the workspace layout of idea and than it might make more sense to carve out the (hopefully) stable part that contains codestyle settings and checks.

rsudev:
@samueltardieu What version of idea are you using (mine is 2016.1.3 Build #IC-145.1617) and how do you manage the config changes?

samueltardieu:
I don't. I have no objection to removing the IDEA specific files.

rsudev:
Rebased and pushed directly. Thanks!

PR #5772, opened 2016-06-03
Extend patterns by proper typographic characters for minutes and seconds

UniQP:


SammysHP:
I doubt that this format is ever used, but let's do it in the professional way. ;)

PR #5786, opened 2016-06-14
Fix #5779: Avoid implicit toString() call

### LINKED ISSUE #5779: Embedded pictures not shown
- open GC61EE3
- the listing contains embedded pictures (at the end)

For me only a placeholder is shown (small rectangle with OBJ in it)

Pictures should be shown

##### Version of c:geo used:
2016.06.10-NB1

Not clear but happening on several caches for me

##### System information:
```
Attach system information here if available (see c:geo menu -> About c:geo -> System)
```

UniQP:


rsudev:
Very good catch I'd say...
Thanks!

Issue #5338, opened 2015-12-27
Allow switching off coord parsing from personal note


Lineflyer:
..or improve it to make it more reliable.

Several users contact us regularly having problems with strange waypoints automatically created by parsing of the personal note. There seem to be cases, where we parse nonsense.

One example I received today:
Personal note content:

```
   A=6 B=0 C=5 D=4 E=13

   N 48° 53.(A*C*E)+194   E 009° 11.((D*71)-0)-1
   N 48° 53.(6*5*13)+194  E 009° 11.((4*71)-0)-1
   N 48° 53.(390)+194     E 009° 11.((284)-0)-1
   N 48° 53.584           E 009° 11.283
```

c:geo creates the following waypoint as parsing result:

```
N 48° 53,000  E 006° 00,000
```

This means users might see a lot of incorrect waypoints on their map without a possibility to permanentely remove them.

So I suggest to either improve the parsing or add an option to disable it in general so that affected users have a solution.

rahu:
I would expect cgeo solves the formula in the given example. :-)

SammysHP:
A formula solver is an often requested feature. Damn, I hate these "number of stairs = A" caches… :D 

What about a keyword to turn on the parsing explicitly? I heard of users who discovered this feature by pure luck, but it also reacts strange in some situations (like here).

Anyway I'd favor a per-cache solution (e.g. `NOWAYPOINTS` keyword in the note).

rahu:
I would prefer the other way round. A "parse waypoints" button in personal notes section. Maybe next to clear waypoints.

Is there any reason why there is no solver?

SammysHP:
A keyword has the advantage that the user can control the behavior of c:geo from the computer (where the user edits the note).

> Is there any reason why there is no solver?
Time and it's not easy – think about all the different kinds of formulas. It would require a complex and flexible parser and an interpreter.

samueltardieu:
I also think a "parse waypoints from note" button would give the user more flexibility. The user is going to use cgeo anyway to find the cache, and having them in control of the waypoints parsing also lets them check that everything went fine. A keyword would need to be cgeo-specific; if every geocaching application did this, the user would have to include all the keywords in the description instead of pressing a button in the one application they are eventually using to find the cache.

rahu:
for the solver: there are several expression parsers and solvers available. I'd like to evaluate some of them and maybe integrate one into cgeo. Is there a issue for this or should I create a new one?

[...]

Issue #4620, opened 2015-01-16
Wrong location when on "Home coords"


Krakapouet:
Concern 2015.01.12
Now I'm located 3000 km from my home ! My coordinates didn't change in geocaching.com so there is a bug for sure.
More enoying is the fact that when opening cgeo the map shown is not the one on the screen when closing the app.
Better to stay on 2014.12.22 at the moment ;-)

Lineflyer:
Can you please post more details:
1. the system information shown in About c:geo - System
2. What is shown on the lower right in c:geo main screen in the case your location is wrong (e.g. "Fused", "Network", "Last position",...)

Lineflyer:
There is another user on FB and support telling, that she is located in "Ukraine" as soon as GPS is disabled and "Home coords" is shown on the lower right.
I asked to doublecheck the home location setting on gc.com and are waiting for feedback. But she confirmed already before, that the home location is set to a position in England.

Lineflyer:
Ok, I can now somehow reproduce that on 2015.01.22-NB:
- Disable Google Play Location Services, Disable GPS and all other location sources
- Set home coords to Dublin on gc.com
- Start c:geo

Result:
- "GPS disabled" and "home coords" is shown on the mainscreen
- The Geoprovider reports another (random?) location, in my case: "Germany, Essen" (which is also different from my current or last location)
- Nearby search returns results around this wrong location
- Live map opens at this wrong location

From the log you can see that
- Home coords are used due to absence of GPS
- The home location from gc.com is retrieved OK (Dublin coords)
- The following nearby search is started for a wrong location

```
16:07:21.158 Info cgeo 15352  [main] Google Play services are available
16:07:21.158 Debug cgeo 15352  [main] OrientationProvider: sensor found
16:07:21.158 Info cgeo 15352  [looper callbacks] No last known location available, using home location
16:07:21.158 Debug cgeo 15352  [looper callbacks] GeoDataProvider: starting the GPS and network listeners
16:07:21.268 Debug cgeo 15352  [network-1] GET http://status.cgeo.org/api/status.json?version_code=30150122&version_name=2015.01.22-NB-c0bddd3&locale=de_DE
16:07:21.378 Info cgeo 15352  [main] Starting cgeo.geocaching 30150122 a.k.a 2015.01.22-NB-c0bddd3
16:07:21.378 Debug cgeo 15352  [RxCachedThreadScheduler-2] Database clean: started
[...]

Issue #922, opened 2011-12-31
Deleting of automatic generated waypoints


schertha:
I can't delete automatic generated waypoints . The new feature: "waypoints from personal notes" is great, but it made 10 new waypoints, that are all just wrong and I can't delete these...

I would like to delete the "original coordinates"-waypoints, too but that is not so important ;)

PR #5790, opened 2016-06-15
Improve latitude/longitude patterns

UniQP:


SammysHP:
Why did you remove valid formats?

UniQP:
It still supports all formats listed at http://www.cgeo.org/faq.html#coords-format. Do you miss any particular case?

SammysHP:
Yes, I miss the formats that you removed. The FAQ is not a complete reference.

rsudev:
To clarify: The FAQ is more aimed at the formats our input dialog and the display offers than to describe what we understand when parsing them.
But it might be a good idea to specify this explicitly as well.

UniQP:
@SammysHP Can you give me an example, when you used the removed formats?

Since N/S/W/E are currently optional, we have some ugly side effects, like parsing "23m in 45°" as N 23° 00.000′ E 45° 00.000′.

SammysHP:
@rsudev It is documented in the code.

@UniQP Then try to fix that. There may not be any other character than a whitespace between two coordinates.

UniQP:
@SammysHP: As already mentioned in https://github.com/cgeo/cgeo/issues/5338#issuecomment-223641967, our unit test contains strings like `N 49° 43' 57" | E 2 12' 35` or `Station3: N51 21.523 / E07 02.680` that contains non-whitespace characters between the coordinates.

Currently, it's simply underspecified what should be parsed as coordinates and what not. That's why I asked you for examples.

SammysHP:
IMHO both of your examples are not valid.

Valid examples:

```
N 52° 36.123 E 010° 06.456'
N52° 36.123 E010° 06.456
N 52°           E 010° 06.456
N 52° 36.123 E 010°
52 10
52° 10°
52.55123 10,56789
52.55123° 10.56789°
N 52.55123° E 10.56789°
```

The last two ones are not supported at the moment AFAICS.

`52 10` is not an easy one and I'm not sure if that should be supported, but `52° 10°` is ok.

Bananeweizen:
As I wrote in #5790, I suggest to disallow integer numbers as longitude _or_ lattitude. So "52 10" is invalid, as well as "52 10.0", "52 E 10.0" etc. That simple change alone will make us get rid of the majority of wrongly parsed personal notes, which is the original intention of this issue.

kumy:
It would be nice to have the unit tests written before the implementation. Or a clear "expectation"...

UniQP:
I can extend the unit test according to the specification, as soon as we have one :-)

Decision, we should make

1. Are non-whitespace characters between latitude and longitude allowed?
2. Is it allowed to mix different formats, e.g. N 50° 00.000 -30.00000?
3. Is it allowed to omit NSWE?
4. Is it allowed to omit °,′,″?
5. Is it allowed to skip the degree part, see https://github.com/cgeo/cgeo/blob/master/tests/src/cgeo/geocaching/location/GeoPointParserTest.java#L46?

Bananeweizen:
I now had a chance to look at the actual change. I tend to reject it completely. I like the replacement of the empty alternative in the first group, but I don't see any improvement in the possessive quantifiers (I rather like the engine to take care of potentially overlapping patterns, not us). And I also don't understand the removal of the lookahead, since that removes a very valid pattern: "10°" is perfectly fine.

Answering your questions:
1. Yes, but only to a certain amount, e.g. 10 characters.
2. No. That will almost always be an attempt to parse something that is not a coordinate.
3. Of course.
4. Yes, as long as we can be certain that a coordinate is meant. And more specifically this also depends on the number of number groups in the input string (to my mind). DD.DDDDDD is clear without any further punctuaction, DD MM is completely not clear without all punctuation, this could be 2 arbitrary numbers in a text, not a coordinate.
5. No, that is an invalid format for sure. I'd rather like to see the input string be fixed at the place where we get the home location string out of the page (that is, before feeding our own coordinates parser).

And surely, there is no scientifically correct answer to how our pattern should look like by means of looking at official standards for formatting coordinates. Our goal must instead be to parse as many coordinates from the wild (that is, out of existing cache websites, and what users type in manually, normally), while at the same time rejecting as many texts as possible, which are not meant to contain coordinates.

Lineflyer:
>   Are non-whitespace characters between latitude and longitude allowed?

Some people might separate by `,` or other similar characters. Even our own export of coords contains a character between lat and lon. This should be allowed.

> Is it allowed to mix different formats, e.g. N 50° 00.000 -30.00000?

No. This might cause wrong parsings and it is unlikely that users mix between formats in one set of coords.

> Is it allowed to omit NSWE?

Yes, it should. A lot of users might ommit NSWE.

> Is it allowed to omit °,′,″?

At least `°` should be allowed to ommit. 

> Is it allowed to skip the degree part, see https://github.com/cgeo/cgeo/blob/master/tests/src/cgeo/geocaching/location/GeoPointParserTest.java#L46?

Not sure what you mean, but without degree part we could only guess the correct coords. So the degree part should be mandatory IMHO.

SammysHP:
To summarize, these should be allowed:

```
N 52° 36.123'
N 52° 36.123′
N 52° 36.123
N 52° 36,123
N 52 36.123
N 52° 36'
N 52° 36' 25"
N52° 36.123
52.123
52.12345°
-52.123456
```

> Are non-whitespace characters between latitude and longitude allowed?

Yes, but maybe limited to a specific number.

> Is it allowed to mix different formats, e.g. N 50° 00.000 -30.00000?

Currently the parser handle each part individually which should be kept (if we want to parse only half of a coordinate). But for normal parsing pairs of the same format should be expected.

> Is it allowed to omit NSWE?

NSWEO (yes, don't forget the O) might be necessary if a negative number is not expected.

> Is it allowed to omit °,′,″?

Yes, all of them.

For `50.123` it should require at least 3 decimals. We should also check the boundaries (±90° and ±180°).

UniQP:
Thanks for the feedback. I will update the pull request accordingly (might take a while).

Short remark to 5.: The test stems from #4620. I checked whether the bug is still there, and it is. So we need to parse "W ° 17.123" to avoid regressions.

samueltardieu:
Set the `Do not merge` label until this is updated.

UniQP:
I revised the code and it now supports the formats proposed in https://github.com/cgeo/cgeo/pull/5790#issuecomment-226436709 except:

- `N 52° 36' 15" E 010° 06.456` (mixture of different formats)
- `N 52.55123° E 10.56789°` (non-existing format)
- `52 10` (no indication for coordinate)

Unfortunately, it's quite a bit of code, but I think that's necessary to detect a mixture of different formats in a maintainable way.

Before merging the PR (from my side), I like to discussion two points:

1. In my opinion, providing NSWEO for the MinDec and DMS format are mandatory. Parsing `24°16.489 48° 24.131` as `N 24°16.489 E 48° 24.131` seems wrong, because it is simply not specified, whether it is N/S and W/E.
2. I would like to forbid format with different precision for latitude and longitude, like `N 52°           E 010° 06.456`. This would prevent us from parsing `N 12° 34.567 E 008° 09.ABC` as `N 12° 34.567 E 008° 09`.

Bananeweizen:
To 1. I disagree for practical and pragmatic purpose. The 12.234 format uses + and - to indicate NWSE. There the + is optional, and numbers without a sign lead to being interpreted as N,E. For users it is only reasonable that all formats without a sign (or compass direction) are interpreted the same way. Since this has been working without any raised user issue before, I really would like us to continue that. This also follows the [robustness princicple](https://en.wikipedia.org/wiki/Robustness_principle), which I see as one of the corner stones for good software.
To 2. Your suggestion seems okay to me, since that probably leads to more false positives than actually correct coordinates. I think we already had complaints about wrongly interpreted personal note content due to this format.

SammysHP:
To 2. I often omit trailing zeros.

UniQP:
Regarding 1. @Bananeweizen To my knowledge decimal degrees do not use '+' (at least I never saw that in the wild); decimal degrees use '-' or nothing. The DMS format specifies to use N/S and W/E and that's exactly what I want to do. It's a different format with a different specification.
If I'm in Brazil and enter `24° 16.489 48° 24.131`, I would not expect to get coordinates in Saudi Arabia; so there is no sane default. I also have not seen people omitting NSWEO in the wild.

Regarding 2. @SammysHP  I see your point. I'm fine with requiring the same number of coordinate parts (e.g. if the coordinates contain milliminutes in the latitude, they should also contain milliminutes in the longitude). This would forbid `N 12° 34.567 E 008° 09.ABC` but would parse `N 12° 34.567 E 008° 09.3AB` as `N 12° 34.567 E 008° 09.300`. The latter case should be really rare, since the latitude must be fixed and the formula just computes the two least-significant digits.

UniQP:
Well, next round... (sorry for the codacy noise)

Regarding 1: The parser now requires NSWEO except for `00° 00.000′`. As discussed above, omitting NSWEO leads to ambiguous coordinates.

Regarding 2. I added word-boundaries at the end of the coordinates, which prevents parsing `N 12° 34.567 E 008° 09.3AB` but would still parse `N 12° 34.567 E 008° 09.3[A][B]`. That should do the job for most cases.

The parser now rejects `N 34 O 17` (see #922) and some UTM coordinates with too small northing/easting values. I also extended the personal note parsing to check for all coordinate formats.

I'm still not happy with the amount of code and will try to improve the PR in this regard. Nevertheless, feel free to review the code.

UniQP:
I cleaned the code up and the PR is now ready to merge from my side. I would appreciate some comments.

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
```diff
@@ -139,6 +137,264 @@ public static void testFloatingPointNbsp() {
...
+        try {
+            GeopointParser.parse("01° E 000°");
+            fail();
+        } catch (final Geopoint.ParseException e) {
+            // expected
+        }
+    }
+
+    public static void testFormula() {
+        try {
```
pstorch: These expected failures cause a lot of duplicate code blocks. Maybe extract into a `assertParsingFails(final String geopointToParse)` method.

main/src/cgeo/geocaching/location/GeopointParser.java
```diff
@@ -3,193 +3,646 @@
...
+         *            the matcher that holds the matches groups
+         * @param first
+         *            the first group to parse
+         * @param last
+         *            the last group to parse
+         * @return the parsed latitude/longitude, or null if parsing failed
+         */
+        @Nullable
+        private Double parseGroups(@NonNull final MatcherWrapper matcher, final int first, final int last) {
+                final List<String> groups = new ArrayList<>(last - first + 1);
```
pstorch: Identation level one too many?

UniQP:
@pstorch Thanks for your feedback. I've integrated the requested changes and the test are indeed much more readable now.

PR #5799, opened 2016-06-22
Remove trailing spaces in strings

UniQP:


UniQP:
Should I also remove the space in the translations or should this be handled via crowdin?

SammysHP:
This might invalidate the matching translations. Maybe there is a better way. @rsudev?

rsudev:
It will invalidate the translations, but those can be easily re-attached (and checked for spaces at the same time). I can do that right after the import to crowdin.

rsudev:
Rebased and merged locally.

PR #5806, opened 2016-06-30
Fix checkstyle warnings

UniQP:


Bananeweizen:
I'll add checkstyle to the eclipse setup to avoid that.

PR #5808, opened 2016-07-02
fix #5760 wrong marker when canceling store

### LINKED ISSUE #5760: cancelling storing leads to wrong markers
- Start downloading an online search result list
- Cancel at some point in between

The online search result list is refreshed and _all_ caches have the offline marker. But in fact, not all have been stored. When randomly clicking caches from the list now, some open immediately, and some will be downloaded first (the ones we cancelled).

pstorch:
moved the assignment of the (new) lists into the store method

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -112,6 +112,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


pstorch:
A geokrety Test fails. I don't think it is related to my changes. I can have a look tomorrow.

kumy:
@pstorch GKM api seems down. I ll look at it. Thanks for pointing this.

kumy:
CI: test please

kumy:
test this please

pstorch:
@kumy thanks

Issue #5810, opened 2016-07-04
Crash while opening certain cache


Lineflyer:
##### Detailed steps causing the problem:
From Facebook:
- Open GC6661W

##### Actual behavior after performing these steps:
c:geo crashes reproducibly 

##### Expected behavior after performing these steps:
Cache details should be opened

##### Version of c:geo used:
2016.06.09 (and current master)

##### Is the problem reproducible for you?
Yes

##### Other comments and remarks:
```
23:20:10.071 Debug cgeo 18728  [network-8] GeokretyConnector.searchTrackables: wpt=GC6661W
23:20:10.072 Debug cgeo 18728  [OkHttp] GET https://www.geocaching.com/seek/cache_details.aspx?decrypt=y&wp=GC6661W&log=y&numlogs=0
23:20:10.072 Debug cgeo 18728  [main] LocationProvider: requesting most precise locations
23:20:10.291 Debug cgeo 18728  [OkHttp] GET https://api.geokretymap.org/export2.php?wpt=GC6661W
23:20:10.403 Debug cgeo 18728  [OkHttp] 301 [Moved Permanently] (281 ms) GET https://www.geocaching.com/seek/cache_details.aspx?decrypt=y&wp=GC6661W&log=y&numlogs=0 (http/1.1)
23:20:10.403 Debug cgeo 18728  [OkHttp] GET https://www.geocaching.com/geocache/GC6661W_kohikohi-paru?decrypt=y&log=y&numlogs=0
23:20:10.404 Debug cgeo 18728  [OkHttp] 302 [Moved Temporarily] (141 ms) GET https://api.geokretymap.org/export2.php?wpt=GC6661W (http/1.1)
23:20:10.509 Debug cgeo 18728  [OkHttp] GET http://api.gkm.kumy.org/export2.php?wpt=GC6661W
23:20:10.617 Debug cgeo 18728  [OkHttp] 200 (40 ms) GET http://api.gkm.kumy.org/export2.php?wpt=GC6661W (http/1.1)
23:20:10.838 Debug cgeo 18728  [OkHttp] 200 (465 ms) GET https://www.geocaching.com/geocache/GC6661W_kohikohi-paru?decrypt=y&log=y&numlogs=0 (http/1.1)
23:20:12.255 Error cgeo 18728  [network-7] UncaughtException
23:20:12.255 Error cgeo 18728  java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.Worker thread.
[...]

PR #5813, opened 2016-07-06
Fix typo in regular expression (#5810)

UniQP:


Bananeweizen:
Thanks for fixing. I only tested it when I still had 2 capture groups. Replaced the second with a back reference directly before committing. My fault not to test again.

PR #5825, opened 2016-07-13
fix #5710: use toast if no connector is active for search

### LINKED ISSUE #5710: several searches have zero results if no connector active
Disable all cache connectors (I had this by accident after a previous debug session). Search by address, search nearest etc. will return zero results.

I suggest that we show an error message for each search, that does not directly lookup a geocode. Cause to my knowledge we only support geocode lookup with disabled connectors. We could also add a warning on the main activity where the login state is shown, but this would potentially be wrong forusers that want to cache completely offline (e.g. by means of importing GPX or whatever, but never using online search).

samueltardieu:


samueltardieu:
@Bananeweizen Is that what you had in mind? Feel free to merge if it is.

PR #5829, opened 2016-07-13
Prevent subscriptions from capturing their environment

samueltardieu:
It is possible that custom unsubscriptions methods reference an outer
class. In this case, unless the subscription has been itself added to a
CompositeSubscription, it may be kept referenced even after being
unsubscribed.

By nulling or emptying the subscriptions after unsubscribing, we prevent
such references from being captured. Such references were detected by
LeakCanary.


samueltardieu:
Retest this please

samueltardieu:
Retest this please

PR #5835, opened 2016-07-16
Make "Found" status in GPX exports optional

auricgoldfinger:
When you export a cache or a list of caches to GPX, then all data is exported, including the "found" status of the cache, obviously indicating that you have found the cache.

However, when you want to export a cache or a list of caches so that a friend can import them (in any application) then this status is only confusing. It is not guaranteed that he found the same caches as well.

This change adds a checkbox to the "export to GPX" dialog to include/exclude the Found status. When checked (default), the sym tag of the exported GPX will contain the text "Geocache Found" for geocaches that were found. When unchecked, the sym tag of the exported GPX will always show "Geocache" and thus omits the word "Found". 

include_found_status


Bananeweizen:
ok to test

main/res/values/preference_keys.xml
```diff
@@ -128,6 +128,7 @@
     <string translatable="false" name="pref_settingsversion">settingsversion</string>
     <string translatable="false" name="pref_trackableaction">trackableaction</string>
     <string translatable="false" name="pref_shareafterexport">shareafterexport</string>
+    <string translatable="false" name="pref_includefoundstatus">includefoundstatus</string>
```
samueltardieu: Do not include translations in your patch, those are handled through crowdin.net.

main/src/cgeo/geocaching/export/GpxExport.java
```diff
@@ -96,11 +96,15 @@ private Dialog getExportDialog(final String[] geocodes, final Activity activity)
         final CheckBox shareOption = ButterKnife.findById(layout, R.id.share);
         shareOption.setChecked(Settings.getShareAfterExport());
 
+        final CheckBox includeFoundStatus = ButterKnife.findById(layout, R.id.include_found_status);
+        includeFoundStatus.setChecked(Settings.getIncludeFoundStatus());
```
samueltardieu: Note: check if this is needed or if the default state can be set in the layout itelf (same for "share").

main/res/values-nl/strings.xml
```diff
@@ -558,6 +558,7 @@
   <string name="init_maptrail">Pad tonen</string>
   <string name="init_summary_maptrail">Laat pad op kaart zien</string>
   <string name="init_share_after_export">Open deel-menu na GPX export</string>
+  <string name="init_include_found_status">\"Gevonden\" status mee exporteren</string>
```
samueltardieu: Do not include translations in your patch, those are handled through crowdin.net.

auricgoldfinger: Ok, removed.

samueltardieu:
Thanks for your contribution.

PR #5843, opened 2016-07-20
Use == to compare enumerations

UniQP:


UniQP:
The codacy issue is a bug:

- http://stackoverflow.com/questions/30916139/pmd-warns-use-equals-to-compare-object-references-when-comparing-enums
- https://sourceforge.net/p/pmd/bugs/1028/

PR #5846, opened 2016-07-20
Simplify angle normalization

UniQP:


samueltardieu:
What problem does it solve exactly? From what I can tell, no new test has been introduced.

UniQP:
What commit are you talking about?

UniQP:
I added an additional test that would fail with `assertThat`. I hope that's what you asked for.

main/src/cgeo/geocaching/utils/AngleUtils.java
```diff
@@ -35,7 +35,8 @@ public static float difference(final float from, final float to) {
      * @return the same angle in the [0, 360[ range
      */
     public static float normalize(final float angle) {
-        return (angle >= 0 ? angle : (360 - ((-angle) % 360))) % 360;
+        final float mod = angle % 360;
+        return mod >= 0 ? mod : (mod + 360) % 360;
```
samueltardieu: Is the `% 360` needed here?

UniQP: Yes, because -Float.MIN_VALUE + 360 becomes 360, which is not within the desired interval.

samueltardieu: Indeed.

Issue #5839, opened 2016-07-18
Improve list handling in map popup


SammysHP:
Short, because from mobile.
- [x] buttons in map popup are small and the text is truncated
- [ ] storing a cache requires an additional action, first select a list, then press ok (previously only select list)
- [x] you can press ok without selecting a list first, cache is stored, but not in a list

pstorch:
The second point is because it is possible to store a cache on several lists at once. So we can't start the store action with a press on the list item only.

SammysHP:
Yep, but a shortcut (e.g. long press on list) would be nice.

pstorch:
Yes, this could work

Bananeweizen:
fixed third item

pstorch:
I have a look at the first item

pstorch:
I'm still thinking about number 2. The list selection dialog is used in two cases, when the cache is stored the first time and to manage the lists, if it is already stored. 
The solution with the longpress on an item to select it and confirm the dialog would only make sense on the first store, when the cache is not on any list yet. It would be confusing when managing the lists.

But there is another possibility: what about preselecting the default list `Stored` in cases where the cache is not yet on any list. This would save one click for the default case. Just hit `OK`. But the users who want to put it on a different list than the default would need 3 clicks. First deselect the default, select the other list and hit `OK`.

While writing this comment and testing again the current behavior I realized that the **move** feature is also working when the cache is not stored yet. So a long press on the `Store` button allows you to store the Cache with a single click on the target list. @SammysHP does this already support your case? 
It's just a bit misleading that the title of the list selection dialog is **Move to different list** in that case.

SammysHP:
I think this is something that one can get used to. So I'd like to get some more feedback from other users for number 2.

pstorch:
There are some more rough edges on the **move** feature: Besides the title when the cache is not stored yet, the `Drop` button (or Trashcan) is not shown after it is stored.
And the **move** feature is missing on the map popup.
Regardless what we decide here, the **move** feature needs some polishing.

pstorch:
Coming back to this issue. In the meantime it is possible with a long press to store a cache on the last used list. @SammysHP is it good enough to close it?

SammysHP:
Yep, should be fine now.

PR #5873, opened 2016-08-03
Change List handling buttons for #5839

pstorch:
I changed the list handling buttons from text to image. So they have the same size in all languages.
Improves the map popup where the text buttons were to small in some languages.


pstorch:
Here is how it looks with ImageButtons instead of text Buttons:

rps20160803_122839
rps20160803_123023



rps20160803_122914
rps20160803_122953

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -72,79 +146,6 @@
...
-import java.util.Set;
-import java.util.regex.Pattern;
-
-import butterknife.BindView;
-import butterknife.ButterKnife;
-import org.apache.commons.collections4.CollectionUtils;
-import org.apache.commons.lang3.StringEscapeUtils;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.Pair;
```
rsudev: I guess we should get this import re-ording sorted out, otherwise it totally obscures any meaningful changes. Was this done by the IDE because of our changed ordering defines? Which IDE so you use?

rsudev:
I like the image buttons :) +1

main/src/cgeo/geocaching/CacheDetailActivity.java
pstorch: Oh, this wasn't on purpose. I'm using AS. There was a settings file for the import order somewhere, right? I'll look for it and amend the commit.

samueltardieu: See https://github.com/cgeo/cgeo/wiki/Coding-Conventions#imports-order

Lineflyer:
Nice approach...sometimes solutions are so simple.
I also like it.

I would go one step further and replace `Add to Watchlist` and `Add` within the favorite section by a + icon.

Is it possible to make those icons show their original description when the user long clicks on them? That would be the the cherry on the cake.

If this is merged, it should be merged to `release`

pstorch:
@rsudev: fixed import order

@Lineflyer: this way?
rps20160803_225735
This uses the icons we already have (a plus within a circle and a trashcan), but maybe it would be better with a `plus` and `minus` icon without a circle. What do you think?

Should I change the pull request to go to `release` first?

Lineflyer:
Ì don't have a real preference between `minus` and `trash` however `minus` might be slightly more matching and not to be confused with deleting the cache itself. So if its feasible on short term i think a `minus` would be good.
With or without circle I dont care.

SammysHP:
I'm not sure that the wrench icon is suited. Maybe a pencil?

Lineflyer:
> I'm not sure that the wrench icon is suited. Maybe a pencil?

Agree.

pstorch:
Yes, the pencil icon (edit) is better suited then the wrench icon (preferences).
Changed it.

pstorch:
I haven't seen @Lineflyer s comment about the `trash` vs `minus` icon yesterday. I'll try to find some icon which fits better for remove.

pstorch:
Here is a version with a new `remove` icon. It corresponds to the `add` icon and doesn't confuse with the `delete` (drop cache) icon above.
device-2016-08-04-112529

SammysHP:
In my opinion + and x looks too similar.

pstorch:
And a version with a `minus` sign as the remove icon:

device-2016-08-04-120512

Looks better, I think.

SammysHP:
That looks good! :+1:

Lineflyer:
That looks good. From UI perspective it looks good for merge and test (to `release`).

Lineflyer:
Any objections to merge?

pstorch:
Is the pull request correctly addressing `master` branch? Or should I change it to `release` instead?

pstorch:
Light skin:
screen

PR #5876, opened 2016-08-03
New strings for showcase

Lineflyer:
The explained functions are no longer new, therefore I slightly changed the content.


Lineflyer:
Please someone doublecheck and merge if OK.

PR #5881, opened 2016-08-05
Fix #5759, New map: "Store Offline" not shown

### LINKED ISSUE #5759: New map: "Store Offline" not shown
- Enable new map
- Open live map
- wait till all visible caches are loaded
- Open menu

- No "Store Offline" menu entry is shown

- Menu "Store Offline" is shown

##### Version of c:geo used:
Reported via `2016.05.29-NB-fb50e95

Yes

##### System information:
```
Attach system information here if available (see c:geo menu -> About c:geo -> System)
```

Reported by a user via c:geo Fanpage on Facebook. I can reproduce it.

rsudev:
- implemented check for loading activity
- implemented handover of visible caches to storing process

Please give it a short cross-check, if possible.

### DIFF
--- a/main/src/cgeo/geocaching/location/Viewport.java
+++ b/main/src/cgeo/geocaching/location/Viewport.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.models.ICoordinates;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
+import java.util.ArrayList;
 import java.util.Collection;
 
 public final class Viewport {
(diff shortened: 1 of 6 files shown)


rsudev:
I'll check the codacy issue tomorrow

samueltardieu:
Is that supposed to give you a "Store as list" menu? It doesn't on my phone.

rsudev:
You should have a 'Store offline' menu entry (once the loading was completed), which, when selected, would either directly store or ask for a list selection, depending on your settings.

samueltardieu:
I don't.

photo_2016-08-09_14-39-34

Bananeweizen:
Just a side note: Also in the old map this is only available after loading finished.

samueltardieu:
Loading had finished. No menu entry  though.

rsudev:
There is another condition that has to be met to show this item (also on the old map) - there need to be unstored caches on the map. Was that the case for you or had all caches already been stored?

samueltardieu:
Nope. None were stored in a list. However, since think this condition should be removed now that caches can be in multiple lists.

rsudev:
Revisiting your screenshot I caught the reason for the menu not appearing - the caches that are cached in the database, are (falsely) not counted as 'not stored'.
If your suggestion to remove that check reaches consensus, this would be easily solvable...

pstorch:
+1 for always show `Store as list`

Bananeweizen:
Fine with the suggestion of always showing. It was previously disabled to avoid the same confusion like after GPX import with the old lists: Caches from moving from one list to another.

rsudev:
Will resubmit, likely rather against master

PR #5898, opened 2016-08-12
fix #5893 arrangement of list buttons

### LINKED ISSUE #5893: Different arrangement of new icons in cache details
Feedback from beta user:
Icons instead of text buttons are nice as they are bigger and clearer, however IMHO the delete button is at the wrong place (also compared to where the textual buttons have been).
I suggest to have only regresh in upper row and lower row should be edit left and delete right.
This also helps in accidently not tapping delete instead of refresh.

pstorch:
Changed the arrangement of the list buttons (offline box) as proposed in #5893

### DIFF
--- a/main/res/layout/cachedetail_details_page.xml
+++ b/main/res/layout/cachedetail_details_page.xml
@@ -101,13 +101,6 @@
                     android:layout_alignParentRight="true" >
 
                     <ImageButton
-                        android:id="@+id/offline_drop"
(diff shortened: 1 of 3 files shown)


pstorch:
Based on `release` branch. 

DetailsPage new cache:
details_new

DetailsPage saved cache:
details_saved

Popup new cache:
popup_new

Popup saved cache:
popup_saved

Issue #4086, opened 2014-07-18
Cache not added to history if unsaved and logged from popup


Bjurran:
Been running the latest nightly builds for som time now and there's one annoying bug with the ones I've used this week. The history doesn't contain all the offline logs resulting in missing found logs when exporting the feild notes. I noticed this the first time July 16th and it's the same today.
All caches are saved and updated on the phone prior to caching.
The offline logs are saved and I'm able to open the logs from the individual caches, but they simply don't appear in the history.
Today 2 of the caches are not shown, last time it was 3 or 4.
Currently running 2014-07-18-NB1-0822830 and can confirm that the bug is still there.

MrYaz:
I know that there are many people that suffer from this bug and as a consequence doesn't use the History at all. Please change priority of this bug if possible.

Lineflyer:
Raising priority as we get quite some complaints where offline logs have been lost due to this reason!

Lineflyer:
It might even get worse:
A chat with a user made me aware that caches logged offline under that condition might stay forever in the database (making it increasing) without beeing visible to the user in any list (not saved=not in lists, not in history).
Only by explicitly loading them (e.g. by GC-code) the user would bring those caches back into the normal cycle.

Lineflyer:
Complaints keep coming in on support mail. So either users have started using this usecase or another manifestation of this issue has been introcuced recently.

Lineflyer:
Here is a (german) description of that problem from FB.
Maybe the hint that the logged date is missing is the clue for finding the root cause why it is not listed in the history:

[...]normal steht dann wenn man den Cache geloggt hat und das Listing aufmacht als Status "Gespeicherter Log und das DATUM" z.B. Gespeicherter Log 11.08.2015 wenn man den Cache loggt, ohne ihn vorher zu speichern steht nur "Gespeicherter Log" also ohne Datum daher scheint er wohl dann auch im Verlauf nicht gelistet zu werden.

MrYaz:
I have also found that a cache is not added to the History list when long-pressing on it's name in a list and selecting Log visit. 
It does not add the Logged date in that case either.

Cache that was logged via Cache details - with Logged datum:
![image](https://cloud.githubusercontent.com/assets/1811963/9294236/27a7f48c-4448-11e5-9876-3d459bbb0bd9.png)

Cache that was logged via popup menu in List view - missing Logged datum:
[...]

PR #5940, opened 2016-09-01
fix #4699 remove cache from history when offline log is deleted

### LINKED ISSUE #4699: Cache not removed from history after offline log deleted
2015.02.21-NB

- Log a cache offline
- Verify it is added to the history
- Open the cache again and delete the offline log by Menu - Log offline - Delete (or whatever the menu is called in english)

The cache keeps showing up in the history permanently (also after app restart)

pstorch:
As #4086 might take longer, I'll send a fix for this separately.

When removing an offline log it also sets the `visiteddate` to 0 to remove it from History.

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -46,6 +46,8 @@ import android.database.sqlite.SQLiteDatabase.CursorFactory;
 import android.database.sqlite.SQLiteDoneException;
 import android.database.sqlite.SQLiteOpenHelper;
 import android.database.sqlite.SQLiteStatement;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import java.io.File;
 import java.io.FilenameFilter;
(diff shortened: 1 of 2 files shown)


main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -2652,7 +2652,9 @@ public static void clearLogOffline(final String geocode) {
 
         init();
 
-        database.delete(dbTableLogsOffline, "geocode = ?", new String[]{geocode});
+        final String[] geocodeWhereArgs = {geocode};
+        database.delete(dbTableLogsOffline, "geocode = ?", geocodeWhereArgs);
+        database.execSQL(String.format("UPDATE %s SET visiteddate = 0 where geocode = ?", dbTableCaches), geocodeWhereArgs);
```
samueltardieu: `WHERE` should be in uppercase (several instances) as it is a SQL keyword.

pstorch: Fixed these. 
There are some more places in this class where SQL keywords are lowercase. Maybe I go through them in a separate PR in one go.

samueltardieu: That would be great indeed.

samueltardieu:
retest this please

PR #5956, opened 2016-09-13
Remove some code duplication between map and new map

samueltardieu:
This attempts to remove some (not all, far from it) duplications between the existing map code and the "new" map.

Note that `liveEnabled` is still present in `mapOptions` and `mapState`. This mirrors the existing situation.


samueltardieu:
retest this please

samueltardieu:
retest this please

samueltardieu:
please retest this

samueltardieu:
retest this please

SammysHP:
IMHO the new map code should stay independent as much as possible. Then it is easier to rewrite the code for new ideas instead of being stuck to the old implementation.

samueltardieu:
Some common (and reusable) ideas have been extracted. No dependency on the new map from the old maps or on the old maps on the new map have been added.

The idea is to extract what is common to be able to also use it in version 2 of Google Maps API.

samueltardieu:
retest this please

rsudev:
I was initially also a little suspicious, but this refactoring is a very worthwhile improvement for both and does not hinder the new advance independently.
There is only one thing. If I read the change correctly, the maps will now be indistinguishable in the navigation menu and navigation menu settings.

samueltardieu:
@rsudev You're right the label were indistinguishable in the original version. I have changed this one hour ago and updated the PR.

PR #5966, opened 2016-09-15
Fix #5948, Live map crashing

### LINKED ISSUE #5948: Live map crashing
Sine this morning we get complaints about freezing/crashing live map.
I can reproduce this after short usage of the live map.

As it wasnt the case last days I assume we have a server issue one again (or still) where cgeo crashes instead of catching the error and giving a proper error message.
Maybe someone can reproduce and provide a log.
IMHO we should urgently fix this because people blame cgeo for this.

rsudev:
Silently handle failures to get additional info for the popup.

Is this the desired approach? Or should we notify in a toast?
@samueltardieu my implementation is just guessing by analogy. Is this correct?


rsudev:
I will think a little on a better semantics. A toast could be a good idea, perhaps. It allows the popup to open nevertheless and you can then proceed to the details, in case it is just one of the gc.com tile server hiccups.

rsudev:
I created an alternative with catching the error higher up and informing the user in a toast.

Issue #1402, opened 2012-04-14
Unsynchronized access to LeastRecentlyUsedSet instances


samueltardieu:
It looks like as of commit gf5d5d78 we may modify the `caches` list in `CGeoMap.java` in an unsynchronized way (if we move the live map too fast for example, while it is loading).

```
E/AndroidRuntime(21312): java.util.NoSuchElementException
E/AndroidRuntime(21312):    at java.util.LinkedHashMap$LinkedHashIterator.nextEntry(LinkedHashMap.java:353)
E/AndroidRuntime(21312):    at java.util.LinkedHashMap$KeyIterator.next(LinkedHashMap.java:370)
E/AndroidRuntime(21312):    at java.util.AbstractCollection.toArray(AbstractCollection.java:343)
E/AndroidRuntime(21312):    at java.util.ArrayList.<init>(ArrayList.java:93)
E/AndroidRuntime(21312):    at cgeo.geocaching.maps.CGeoMap.countVisibleCaches(CGeoMap.java:337)
E/AndroidRuntime(21312):    at cgeo.geocaching.maps.CGeoMap$1.handleMessage(CGeoMap.java:223)
E/AndroidRuntime(21312):    at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime(21312):    at android.os.Looper.loop(Looper.java:137)
E/AndroidRuntime(21312):    at android.app.ActivityThread.main(ActivityThread.java:4424)
E/AndroidRuntime(21312):    at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime(21312):    at java.lang.reflect.Method.invoke(Method.java:511)
E/AndroidRuntime(21312):    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:787)
E/AndroidRuntime(21312):    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:554)
E/AndroidRuntime(21312):    at dalvik.system.NativeStart.main(Native Method)
```

PR #5979, opened 2016-09-19
Cleanup map code

UniQP:


main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -360,15 +354,11 @@ private void showProgress(final boolean show) {
...
+            if (waitDialog != null) {
+                if (msg.what == UPDATE_PROGRESS) {
                     final int secondsElapsed = (int) ((System.currentTimeMillis() - detailProgressTime) / 1000);
-                    final int secondsRemaining;
-                    if (detailProgress > 0) {
-                        secondsRemaining = (detailTotal - detailProgress) * secondsElapsed / detailProgress;
-                    } else {
-                        secondsRemaining = (detailTotal - detailProgress) * secondsElapsed;
-                    }
+                    assert detailProgress > 0;
```
Bananeweizen: We generally don't use Java assert statements (except for helping the null analysis in some cases). Either you _are_ sure and the code cannot ever reach the deleted else, then the assert is not needed. Or you are not sure, then the refactoring is wrong.

main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -1316,28 +1300,22 @@ private void doDisplayRun() {
 
             // display caches
             final List<Geocache> cachesToDisplay = caches.getAsList();
-            final List<Waypoint> waypointsToDisplay = new ArrayList<>(waypoints);
```
Bananeweizen: This was introduced in #1402 to avoid iterating the original waypoints collection asynchronously. Generally such copy constructors in cgeo should only be removed if one is completely sure there is no more concurrent access to the collection in question.

main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -1487,7 +1465,7 @@ private static synchronized void filter(final Collection<Geocache> caches) {
 
         final List<Geocache> removeList = new ArrayList<>();
         for (final Geocache cache : caches) {
-            if ((excludeMine && cache.isFound()) || (excludeMine && cache.isOwner()) || (excludeDisabled && cache.isDisabled()) || (excludeDisabled && cache.isArchived())) {
+            if ((excludeMine && (cache.isFound() || cache.isOwner())) || (excludeDisabled && (cache.isDisabled() && cache.isArchived()))) {
```
Bananeweizen: superfluous brace in "excludeDisabled && (cache.isDisabled()"

UniQP: Good catch (should be || inside the braces).

main/src/cgeo/geocaching/maps/CGeoMap.java
UniQP: Copying the waypoints does not solve the problem. IIRC we already had this discussion.

UniQP:
I adapted the commit according to @Bananeweizen suggestions.

PR #5987, opened 2016-09-22
webview prototype, fixes #5986

### LINKED ISSUE #5986: webview of cache
For working around limitations of the textviews in our cache details, I suggest to add a webview tab (as the last tab). This could either show the stored HTML page from download (which we currently don't store) or the online page. This browser based tab is then able to show tables, animated gifs and other things which are missing in the normal text view.

I've implemented a prototype, accessing the online page, and removing everything except the cache content. The browser content typically looks very similar to our description (except for defaulting to white background like in the desktop browser).

Example description with broken tables:
2016-09-22_15h52_36

Example webview with correct tables for the same cache:
2016-09-22_15h53_21

What's missing:
* The browser doesn't share the session with our httpclient. Therefore you have to manually login once on the browser tab.
* progress reporting while loading

I've deliberately choosen to add online browsing, and not to show the offline contents (which would need to be stored first). This functionality is similar to opening the external browser via menu item, except for removing everything from the page, which is not actual content.

Bananeweizen:
- shows _online_ cache page in additional tab in cache details
- lazy loading when switching to the page
- need to login manually by now
- after page load finished, everything around cache description is cut
  via javascript

### DIFF
--- a/main/res/layout/cachedetail_webview_page.xml
+++ b/main/res/layout/cachedetail_webview_page.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<WebView  xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/webview"
+    android:layout_width="fill_parent"
+    android:layout_height="fill_parent"
+/>
--- a/main/res/raw/cache_details_webview.js
+++ b/main/res/raw/cache_details_webview.js
@@ -1,0 +1,15 @@
+if (document.getElementById('hlSignIn') == null) {
+    var contents = document.getElementsByClassName('UserSuppliedContent');
+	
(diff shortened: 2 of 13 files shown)


Bananeweizen:
Just in case someone is wondering: I've moved the new classes into a separate package, as I intend to split also the existing classes of the details view into separate top level classes in the same package. Working with a 2500 lines file is not nice.

Lineflyer:
We should in the same step remove that IMHO useless c:geo internal browser.
It wont keep cookies...ans most users rather get confused by this.

samueltardieu:
@Bananeweizen You might want to `git reset --hard samueltardieu/webview` and `git push -f` into your branch to update this PR (I've rebased your changes on top of the current `master`).

PR #5989, opened 2016-09-23
Do not pass null into an Observable-like

samueltardieu:
Version 2.0.0 of RxJava will prohibit using null values in any
Observable-like as mandated by the reactive streams specfication.


Bananeweizen:
retest this please

Bananeweizen:
I see a process crash when running tests locally. Can you please verify this once more @samueltardieu ?

PR #6001, opened 2016-09-27
Removed duplicate SEPARATOR Constant, cleanup of location tests

pstorch:
Removed duplicate Constant `SEPARATOR`.

I first thought about pushing this small improvement directly to master, but decided otherwise.

Not sure about usage of the Constant in `GeoPointFormatterTest`. I replace the hard coded middle dot with the Constant, to be consistent with the other two test methods. But sometimes its better for Unittests to use redundant hard coded values. To ring the bell when someone accidentally changes the Constant.
What do you think?

### DIFF
--- a/main/src/cgeo/geocaching/location/GeopointFormatter.java
+++ b/main/src/cgeo/geocaching/location/GeopointFormatter.java
@@ -1,12 +1,13 @@
 package cgeo.geocaching.location;
 
+import cgeo.geocaching.utils.Formatter;
+
 import java.util.Locale;
 
 /**
  * Formatting of Geopoint.
  */
 public class GeopointFormatter {
-    private static final String SEPARATOR = " · ";
 
     private GeopointFormatter() {
         // utility class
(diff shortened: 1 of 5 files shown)


Bananeweizen:
I'd rather leave the hardcoded literal in the unit tests for multiple reasons:
* Generally, every method under test should be a black box for the test. So the unit test should not be coupled to internals of the class being tested.
* If we were really doing TDD (I admit, I also fail rather often there), it were actually impossible to have the separator constant in the test class, since the assertion would be written first.
* If you feel that this might now lead to code duplication in test methods (i.e. when testing 20 different cases of coordinate formatting), then the better way would be to write one new assertion method assertSomeKindOfFormatting, which contains this (hardcoded) literal just once, and to call that new assertion 20 times.

Independent of that question, I personally don't like the format string instead of the concatenation here. This reduced readability quite a bit, while I see no additional benefit (even though this might save some operations on the CPU, that is neglectable, since this performance difference is not notable by any user). If you don't feel convinced by this, have a look at GCConstants, and find that some of the very huge regular expressions are in fact written as multiple concatenated literals, to at least give some more structure to them.

pstorch:
@Bananeweizen I agree with Tests using hard coded and probably duplicated values. That was my thought after I changed it the other way around.

My point with String.format was not performance, but consistency. We mix two methods here to produce a String. But I can see your point in readability, too.

pstorch:
Ok, to not keep this open too long, I reverted the String concatenation to keep the readability.
I still removed the duplicate SEPARATOR Constant.
I removed the usage of the SEPARATOR Constant in the tests to not couple the tests to the internals of the code under test.
I found some usages of `assertEquals()` and changed it to `assertThat()` in the location test package.

pstorch:
Something wrong with the CI server? 
`com.android.builder.testing.api.DeviceException: com.android.ddmlib.InstallException: Failed to finalize session : INSTALL_FAILED_INSUFFICIENT_STORAGE`

Bananeweizen:
Retest this please is the comment to trigger a rebuild

pstorch:
Still failing, but this time `Failed to establish session`. Something wrong with the emulator? Try again?

pstorch:
retest this please

pstorch:
Now again the `INSTALL_FAILED_INSUFFICIENT_STORAGE` error.
Can @mucek4 or @kumy help here?

kumy:
not at home right now, please try to deactivate failing executors.

On October 1, 2016 10:37:54 PM GMT+02:00, Peter Storch <notifications@github.com> wrote:
>Now again the `INSTALL_FAILED_INSUFFICIENT_STORAGE` error.
>Can @mucek4 or @kumy help here?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/cgeo/cgeo/pull/6001#issuecomment-250936137

-- 

samueltardieu:
I'm on it.

samueltardieu:
retest this please

pstorch:
Thanks

PR #6003, opened 2016-09-28
Fix #5948, Live map crashing

### LINKED ISSUE #5948: Live map crashing

rsudev:
Catch errors requesting popup details and inform user in a toast


Bananeweizen:
:+1: from me, but I don't feel informed enough to judge this change. If I remember right (from the first alternative PR), @samueltardieu should comment on this?

PR #6012, opened 2016-10-01
4086 offline log for unsaved caches

pstorch:
This is still WIP and I'm still testing it.
Don't know all side effects for saving caches which have `detailed = false`.
Looks a bit awkward, too, when such a Cache is viewed in CacheDetailActivity. I guess this can confuse users. Maybe a visible indication would be necessary that the Cache was not properly loaded and a hit on refresh can solve it.

### DIFF
--- a/main/src/cgeo/geocaching/loaders/HistoryGeocacheListLoader.java
+++ b/main/src/cgeo/geocaching/loaders/HistoryGeocacheListLoader.java
@@ -18,7 +18,7 @@ public class HistoryGeocacheListLoader extends AbstractSearchLoader {
 
     @Override
     public SearchResult runSearch() {
-        return DataStore.getHistoryOfCaches(true, coords != null ? Settings.getCacheType() : CacheType.ALL);
+        return DataStore.getHistoryOfCaches(coords != null ? Settings.getCacheType() : CacheType.ALL);
     }
 
 }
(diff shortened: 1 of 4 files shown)


samueltardieu:
retest this please

samueltardieu:
retest this please

SammysHP:
If the cache has no details, it must not be possible to open it. If it is opened, the details must be loaded. If that is not possible, an error will be shown that the details cannot be loaded.

But maybe there's a better way of handling this situation.

pstorch:
The `detailed` property of Geocache is a boolean, but in fact we have several levels of details:

1. when the Map shows the Cache marker with an orange circle it represents the least details we could have. In this situation it's not possible to log offline. It's not even possible to show the Map popup when the user is offline.
1. when the Map shows a gray/white marker, the `detailed` property is still `false`, but it is possible to open the popup when the user is offline. Now (with this PR) it is also possible to log offline. We still miss the Cache description, attributes, logs, ...
1. when the Cache was imported from a GPX. We have the description and some logs, but we still miss favorites, personal notes, images, ...
1. fully loaded, when c:geo itself loads the Cache details via an online connection to GC, OC, ...

With this PR I enabled saving the Cache in state 2) to allow proper offline logging.

pstorch:
rebased both commit into one

pstorch:
retest this please

pstorch:
Can anyone think of special situations which should be tested/ double checked when storing Geocaches with `detailed=false`?

pstorch:
I've removed the `Do not merge` label. I can't find any hard obstacles. 
The only thing is that saved caches without details have much less details on the details page. Obvious for us, but might surprise some users. A hit on the refresh button solves it and loads the missing details.

Lineflyer:
> Obvious for us, but might surprise some users. A hit on the refresh button solves it and loads the missing details.

Indeed some users might get confused. Lets first see how it looks like (after this has been merged) before deciding if other measures are meaningful (e.g. a toast that this cache is not fully loaded and user should refresh to get all info).

PR #6015, opened 2016-10-02
fix #4564 mass upload of personal notes

### LINKED ISSUE #4564: Mass operation for uploading notes
Today I did a cache tour with bonus mystery. So as always I noted down all bonus numbers in each cache personal note, without any uploading while on the tour. Now it would be nice to upload all those notes together.

pstorch:
There is a new Export submenu "Personal note" which can be used to mass upload personal notes.

The only issue is the 500 character warning in the original single upload function. Do you think it is necessary here? I could only scan all personal notes and if one of them is > 500 chars add such a message to the user. But I'm not a fan of it.

I'm not sure about the proper case for the word `notes`. We have `Personal notes`, but `Field Notes`. On geocaching.com I can find both version `Field Notes` and `field notes`.

### DIFF
--- a/main/src/cgeo/geocaching/export/PersonalnoteExport.java
+++ b/main/src/cgeo/geocaching/export/PersonalnoteExport.java
@@ -1,0 +1,86 @@
+package cgeo.geocaching.export;
+
+import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.ActivityMixin;
+import cgeo.geocaching.connector.ConnectorFactory;
+import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.connector.capability.PersonalNoteCapability;
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.utils.AsyncTaskWithProgress;
+import cgeo.geocaching.utils.Log;
+
+import android.app.Activity;
+import android.content.Context;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import java.util.List;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Exports personal notes.
+ */
+public class PersonalnoteExport extends AbstractExport {
+
+    private int persNotesCount = 0;
+
+    public PersonalnoteExport() {
+        super(R.string.export_persnotes);
+    }
+
+    @Override
+    public void export(@NonNull final List<Geocache> cachesList, @Nullable final Activity activity) {
+        final Geocache[] caches = cachesList.toArray(new Geocache[cachesList.size()]);
+        new ExportTask(activity).execute(caches);
+    }
+
+    private class ExportTask extends AsyncTaskWithProgress<Geocache, Boolean> {
+        /**
+         * Instantiates and configures the task for exporting personal notes.
+         */
+        ExportTask(@Nullable final Activity activity) {
+            super(activity, getProgressTitle(), CgeoApplication.getInstance().getString(R.string.export_persnotes));
+        }
+
+        @Override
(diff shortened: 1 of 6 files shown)


samueltardieu:
Side note: you should add a # before the issue number in the commit message, that adds a link to the issue (and a backlink to the commit as well).

pstorch:
@samueltardieu ups, there should be one. I'll amend it.

main/src/cgeo/geocaching/export/PersonalnoteExport.java
```diff
@@ -0,0 +1,99 @@
...
+            }
+            return true;
+        }
+
+        @Override
+        protected void onPostExecuteInternal(final Boolean result) {
+            if (activity != null) {
+                final Context nonNullActivity = activity;
+                if (result) {
+                    ActivityMixin.showToast(activity, nonNullActivity.getString(R.string.export_fieldnotes_upload_success));
```
Bananeweizen: You should not reuse the existing string R.string.export_fieldnotes_upload_success. Even though it might simply say "Success" in English, translators of other languages may refer to field notes in there, since we gave it that id. If you want to use the same string in both cases, you basically have to introduce a new id "R.string.export_success" or similar.

Bananeweizen:
Regarding the capitalization: The original English terms are simply "[field notes](https://en.oxforddictionaries.com/definition/field_notes)" and "personal notes". Whether we capitalize all, some or none of the words depends on our own capitalization strategy. We have never specified that formally, but I believe most of our app uses [sentence capitalization](https://en.wiktionary.org/wiki/sentence_case), i.e. capitalization like in this comment, where "field note" would only be capitalized "Field note" when starting a new sentence and not at all otherwise.

pstorch:
@Bananeweizen thanks for the review. Not grouping by Connector simplifies the code a bit. I over sighted some of the R.strings, because AS shows the text instead of the variable name and there was no "field notes" in it. I now created specific ones for personal notes and used them to create more informative messages. I think we should remove geocaching.com from the field notes texts as well, because GC is not the only platform supporting field notes.

pstorch:
retest this please

pstorch:
retest this please

main/src/cgeo/geocaching/export/PersonalnoteExport.java
```diff
@@ -0,0 +1,84 @@
...
+            super(activity, getProgressTitle(), CgeoApplication.getInstance().getString(R.string.export_persnotes), true);
+        }
+
+        @Override
+        protected Boolean doInBackgroundInternal(final Geocache[] caches) {
+            try {
+                for (final Geocache cache : caches) {
+                    final IConnector connector = ConnectorFactory.getConnector(cache);
+                    if (connector instanceof PersonalNoteCapability && StringUtils.isNotBlank(cache.getPersonalNote())) {
+                        publishProgress(++persNotesCount);
```
Bananeweizen: I didn't notice this on first review: I suggest to publish the progress outside the if. If you have a mixed collection of caches where some do not have personal note capability, the progress would not report these caches, and when it has finished the loop, the progress bar would suddenly jump to the end (e.g. if you have 300 caches with personal note capability and 100 without, the progress would go smoothly to 75% and then suddenly finish).
If progress is always reported, then the progress bar may do small jumps in between, but it will better reflect the overall progress.

pstorch: @Bananeweizen at first I haven't seen no progress bar. But after your comment I looked closer. I've copied the `indeterminate=true` from the `FieldnoteExport.ExportTask`. After changing this to false (or using the default false) I now have a progress bar. Now publishing the progress on every Geocache makes sense. It's not very smooth on my Device, but much more informative to see the overall progress.

PR #6016, opened 2016-10-03
fix #5981 refresh on history not put on default list

### LINKED ISSUE #5981: Refresh caches on history will store them to default list
- Log some caches from different lists
- Delete some of them from all lists, let some remain on the different lists
- Open the history
- Select "Refresh all"

- All refreshed caches are now stored on the default list (whereas caches still stored on a random list will additionally appear on the default list and no longer stored caches will also appear on the default list)

- Refreshing the history should neither store no longer stored caches, nor add them to the default list

##### Version of c:geo used:
2016.08.27

From support mail (only partly retested by myself)

pstorch:
When Caches are refreshed on history list, they should not be put on the default list (additionaly).


pstorch:
retest this please

PR #6019, opened 2016-10-05
fix #5872 sort order for history

### LINKED ISSUE #5872: History sorted incorrectly
History is sorted by the time when the offline log was created, not the date that was set for the log.

pstorch:
When logging the `visiteddate` was set to the current timestamp, but it should be the data/time the user chose in the Log Dialog.


samueltardieu:
retest this please

PR #6027, opened 2016-10-11
fix #6024 recognize log picture upload success

### LINKED ISSUE #6024: Log picture upload ends with error message although OK
- Log a cache find and attach picture with the log (GC.com)
- Send the log

An error message is shown, that sending seems failed and user shall check on website (retry/cancel/offline).

This occured to me on the last 14 days for each picture I attched, so something seems wrong while checking the result.
In all cases log and picture had been uploaded correctly.

##### Version of c:geo used:
Master

Yes (at least it seems like that to me...can't do extensive tests right now)

Feel free to test with my owned caches to check whether reproducible

pstorch:
Updated the pattern to recognize the successful upload of log pictures.

Due to `Regression` I opened the PR against `release` branch, right?


Lineflyer:
Thanks for taking care, release branch is ok.
Feel free to merge, than I might test with an RC.

PR #6038, opened 2016-10-18
Fix modifier orders

samueltardieu:


samueltardieu:
@Bananeweizen I use the PR as a way to run a full job (as this is easy to do and prevents obvious mistakes). Please do not merge my PRs *before* the tests results are available.

Bananeweizen:
I was wondering about that. I merged since the change _obviously_ could not fail tests. Will not do in the future...

samueltardieu:
No problem. I hadn't communicated my way of using the PR clearly and could have done the same thing. I tend to protect myself against committing "obvious" changes that cannot fail if I can have the CI confirm that indeed they don't :-)

PR #6052, opened 2016-10-22
fix #6051: date format parsing incomplete

### LINKED ISSUE #6051: date format parsing might be broken
According to recent FB posts it might be that cgeo does not parse all formats from gc.com.
Can we check?

samueltardieu:


pstorch:
I guess the `GCConstants.PATTERN_CUSTOMDATE` needs to be adjusted as well.

PR #6053, opened 2016-10-22
Adjust date format parsing (#6051)

samueltardieu:


pstorch:
@samueltardieu I think this Pattern is too broad now. There are more `<option selected="selected"` on that page, and the wrong is selected now.

samueltardieu:
Yes, I have tested and restricted it since your remark.

PR #6059, opened 2016-10-23
Fix #6043, Support user modified coords for all cache types

### LINKED ISSUE #6043: Support user modified coords from/to server for all cache types
According to the latest release notes gc.com now supports this function for all cache types, thus c:geo should do.

rsudev:
Just removed the type checking.


rsudev:
Hmm - the test result looks suspicious, will check it.

rsudev:
retest this please

PR #6075, opened 2016-10-30
fix #6065 no logs when parsing gc custom date failed

### LINKED ISSUE #6065: No logs
2016.10.23

Two French-speaking users have reported that they cannot see the logs associated with a cache. The log tab says that there are none.

The one I exchanged mails with (on support) is a basic member.

pstorch:
fixes #6065 in release. Needs to be done differently for master, after @samueltardieu jsoup changes (which I appreciate). This fix shows that Regex parsing of HTML can fail too often.

As suspected some GC profiles don't have a selected default date format. So the Regex fetches an option from another selection later on on the preferences page. 

First I fixed that the Regex to fetch only Date Format patterns, including ones with `&#32;`, and falling back to the default gc custom dateformat when something fails.

There is already the default `MM/dd/yyyy` but c:geo somehow used `dd.MM.yyyy` here, don't know where this comes from.


Lineflyer:
I would like to create a RC-version containing this fix, however I am not sure about the way to merge this as it is incompatible with the `master` branch. Can someone take a look and do the needful?

PR #6076, opened 2016-10-30
fix #6065 no logs when gc custom date not detected

### LINKED ISSUE #6065: No logs

pstorch:
Now this is the fix for branch `master`. Here only a proper fallback needed to be set.

As the PR for `release` can't be merged without conflicts to `master`, I was not sure what the proper way is. So I've sent a separate PR for `master`.

If you prefer I can also do a manual merge from `release` to `master` after the other PR is merged into `release`. Then I close this one.


samueltardieu:
Merged. For reference:
- I ensured that `release` was merged into `master`.
- I merged #6075 into `release`.
- I merged `release` into `master` with `-Xours`. This recorded the merge, but no change actually came in.
- I merged #6076 into `master`.

Issue #5850, opened 2016-07-22
Memoryleak reports


pstorch:
I got some Leaks reported while using c:geo. As usual, I'm on a developer build (currently `589a996`).

One example:
![2016-07-22 21 16 10](https://cloud.githubusercontent.com/assets/5598124/17071674/4a8b9c24-5063-11e6-908a-6527315f63dd.png)

kumy:
got one once. don t know what to do too...

Bananeweizen:
You may want to read http://www.androiddesignpatterns.com/2013/01/inner-class-handler-memory-leak.html. For this specific case, the Handler had a private member of type Activity (given in constructor). Since the handler may live much longer than the referenced activity, it will stop the garbage collector from actually removing the activity after it (the activity) has been finished. I replaced the activity member by a weak reference to avoid that.

However, I'm not certain of my analysis, so some thoughts of @samueltardieu would always be appriciated.

pstorch:
@Bananeweizen good read, thanks. I stumbled over a warning about a non-static Handler once.

I have another report:

```
In cgeo.geocaching:2016.07.11-30194b5 developer build:20160711.
* cgeo.geocaching.CachePopup has leaked:
* GC ROOT thread java.lang.Thread.<Java Local> (named 'network-9')
* references cgeo.geocaching.CachePopupFragment$StoreCacheHandler.this$0
* references cgeo.geocaching.CachePopupFragment.cacheDistance
* references android.widget.TextView.mContext
* leaks cgeo.geocaching.CachePopup instance

```

kumy:
```
In cgeo.geocaching:2016.07.21-621ba36 developer build:20160721.
* cgeo.geocaching.CacheListActivity has leaked:
* GC ROOT static cgeo.geocaching.sensors.Sensors$InstanceHolder.INSTANCE
* references cgeo.geocaching.sensors.Sensors.directionObservable
* references rx.Observable.onSubscribe
* references rx.internal.operators.OnSubscribeRefCount.source
* references rx.internal.operators.OperatorReplay.current
* references java.util.concurrent.atomic.AtomicReference.value
* references rx.internal.operators.OperatorReplay$ReplaySubscriber.producersCache
* references array rx.internal.operators.OperatorReplay$InnerProducer[].[10]
* references rx.internal.operators.OperatorReplay$InnerProducer.child
[...]

PR #6077, opened 2016-11-01
Migration to RxJava 2

samueltardieu:


main/src/cgeo/geocaching/utils/AndroidRx2Utils.java
```diff
@@ -0,0 +1,99 @@
...
+import io.reactivex.functions.Consumer;
+import io.reactivex.functions.Predicate;
+import io.reactivex.internal.schedulers.RxThreadFactory;
+import io.reactivex.schedulers.Schedulers;
+
+public class AndroidRx2Utils {
+
+    public static final Scheduler computationScheduler = Schedulers.computation();
+
+    public static final Scheduler networkScheduler = Schedulers.from(Executors.newFixedThreadPool(10, new RxThreadFactory("network-")));
```
Bananeweizen: Oh. So you create all the memory leaks where a thread called "network-x" is mentioned? ;) See #5850. I always thought those threads are from the framework. With this new knowledge gained, is there anything that we can and should do to avoid the leaks referring to these threads?

samueltardieu: The leak is not here, it happens to be on those threads. Anyway, `bindActivity` in its current form (from the old RxAndroid package)  is going away, and the new one being used when necessary (in `AndroidRx2Utils` as I am refactoring) uses a `WeakReference`.

main/src/cgeo/geocaching/connector/gc/GCMap.java
```diff
@@ -332,16 +332,16 @@ public Void call(final Bitmap bitmap, final String data) {
                                     }
                                 }
 
-                                // release native bitmap memory
-                                if (bitmap != null) {
+                                // release native bitmap memory if we didn't get the placeholder
+                                if (bitmap != ONE_ONE_BITMAP) {
```
samueltardieu: Shut up bot, I mean this comparaison.

samueltardieu:
Merged manually

PR #6081, opened 2016-11-03
Make "some seconds" more natural english

akkatracker:
Noticed this when out caching today - this does not come across as fluent English.


samueltardieu:
ok to test

samueltardieu:
**Note**: merging this will likely invalidate all translations. It should be merged only when there is time to update translations before a release, for example just after a release.

rsudev:
I'd say currently is not a bad point in time to merge it on master, as a call for translations for the next feature release is still due.

samueltardieu:
@rsudev Done.

@akkatracker Thanks for your contribution.

akkatracker:
@samueltardieu no worries - enjoy using the app so it's the least I can do.

PR #6093, opened 2016-11-06
fix #6092 crash when refreshing certain caches

### LINKED ISSUE #6092: cgeo crashes when refreshing certain caches
- Save either GC2MCW3 or GC90D6 to an empty list
- select "Refresh all" caches in this list
- cgeo crashes

cgeo crashes

the caches should get refreshed

##### Version of c:geo used:
2016.11.06-NB-cc2aae8

Yes, crashes occur every time I try to refresh one of the both caches listed above

##### System information:
```
Android 4.4.2
tried on N7100 (Galaxy Note II) and Galaxy Tab 10.1 2014 edition
```

Refreshing both caches works well using release 2016.11.05

pstorch:
Looks like `onNext(bitmap)` was called with a null `bitmap`, which is not allowed.

### DIFF
--- a/main/src/cgeo/geocaching/network/HtmlImage.java
+++ b/main/src/cgeo/geocaching/network/HtmlImage.java
@@ -232,7 +232,9 @@ public class HtmlImage implements Html.ImageGetter {
                         final ImmutablePair<BitmapDrawable, Boolean> loaded = loadFromDisk();
                         final BitmapDrawable bitmap = loaded.left;
                         if (loaded.right) {
-                            emitter.onNext(bitmap);
+                            if (!onlySave) {
+                                emitter.onNext(bitmap);
+                            }
                             emitter.onComplete();
                             return;
                         }


main/src/cgeo/geocaching/network/HtmlImage.java
```diff
@@ -232,7 +232,9 @@ public void run() {
                         final ImmutablePair<BitmapDrawable, Boolean> loaded = loadFromDisk();
                         final BitmapDrawable bitmap = loaded.left;
                         if (loaded.right) {
-                            emitter.onNext(bitmap);
+                            if (bitmap != null) {
```
rsudev: while this looks like an easy fix, I couldn't find a call chain that legitimately would lead  to a null bitmap with a true flag in this context.
So it might be worthwhile to look a little deeper into the root cause.

samueltardieu: There is a case where the pair can be `(null, true)` in `HtmlImage#loadCachedImage()` if the image is present and fresh and we do not want to display it (`onlySave` is `true`).

In order not to mask a problem, I would rather `onNext` the bitmap only if `onlySave` is `false`, and use:

``` java
   if (!onlySave) {
      emitter.onNext(bitmap);
   }
```

so if we have a `null` image popping up while `onlySave` is `false` we get this error again (this should not happen).

pstorch: @samueltardieu you are right, the case was `(null, true)`. I've updated the PR.

pstorch:
Testfailure isn't related to the change. Cache https://coord.info/GC2CJPF has been archived today and breaks some tests.

PR #6099, opened 2016-11-07
fix #6098: do not make rounding errors on fractional parts

### LINKED ISSUE #6098: Rounding error near "minute confluence"?
- Open a cache
- Move to  East xx° yy'.998 from West
- Swipe to waypoints
- "Add Waypoint", select DD MM.MMM format
- Now move further to East

- Display will eventually show xx yy.000 (and a WP taken will be >1 km off)

- Expected: yy gets incremented (and WP reflects current position)

##### Version of c:geo used:
2016.07.09-NB.34c95b1

It's a bit tricky because is apparently only happens at xx yy.999[5-9] and with very good GPS reception

##### System information:
OnePlus (A0001, bacon)
6.0.1 MHC19Q

This looks quite similar to Issue #2624 - but this time, for minutes

samueltardieu:
The latitude and longitude are now stored in micro-degrees. The test
suite has been extended to increase coverage.

### DIFF
--- a/main/src/cgeo/geocaching/location/Geopoint.java
+++ b/main/src/cgeo/geocaching/location/Geopoint.java
@@ -3,14 +3,14 @@ package cgeo.geocaching.location;
 import cgeo.geocaching.R;
 import cgeo.geocaching.models.ICoordinates;
 
-import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


samueltardieu:
retest this please

samueltardieu:
retest this please

PR #6105, opened 2016-11-09
fix #5975 remember last list selection

### LINKED ISSUE #5975: Remember last list selection
Sometimes I store caches one by one from the live map. Every time I have to select the same list(s) again and again.

Previously we had the "ask for list" option, now I suggest to add an option to remember the last list selection for storing a cache.

This is somehow related to the requested long-press in #5839 as it removes the additional work for selecting a list.

pstorch:
Remembering the selected lists for the multi list selection dialog. I've decided to remember the selection in the Settings, similar to last used list.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -108,6 +108,7 @@
     <string translatable="false" name="pref_lastmaplon">maplon</string>
     <string translatable="false" name="pref_livelist">livelist</string>
     <string translatable="false" name="pref_lastusedlist">lastlist</string>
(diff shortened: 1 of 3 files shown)


pstorch:
retest this please

samueltardieu:
I cannot test it right now so I'll ask: does it cover the case where a list belonging to the saved selected lists has been deleted? This can also happen if a preferences file is restored during a reinstallation.

pstorch:
@samueltardieu oh, good point. I'll check.

pstorch:
@samueltardieu now deleted lists are handled

PR #6112, opened 2016-11-12
remove micro degree geopoint constructor

Bananeweizen:
It was introduced basically as clone constructor, but can be inlined.
This avoids the potential confusion with the unused object. Also there
is no client using that constructor outside of the class itself.

### DIFF
--- a/main/src/cgeo/geocaching/location/Geopoint.java
+++ b/main/src/cgeo/geocaching/location/Geopoint.java
@@ -1,8 +1,5 @@
 package cgeo.geocaching.location;
 
-import cgeo.geocaching.R;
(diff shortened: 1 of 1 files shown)


samueltardieu:
I'd rather make the constructor `private` than dividing then multiplying. The whole point of rounding is to get an exact result.

However, maybe the rounding is not useful at all. I'll check that.

samueltardieu:
I'd prefer that you merge #6116 instead. The fix in 567133cecbc325b57b5e07efd2fff3cb1479dfa2 made rounding unnecessary.

Bananeweizen:
Perfectly fine. That's why I created a pull request in the first place, instead of my normal pushes. :)

PR #6116, opened 2016-11-12
Remove integer constructor for Geopoint and rounding method

samueltardieu:


samueltardieu:
Note: we may have a small rounding error there. The (better IMO) alternative would be to make the `Geopoint(int, int)` constructor private.

Bananeweizen:
I'm also fine with leaving the current code in. Honestly, I have not really understood the consequence of either rounding or not rounding. I thought that only rounding when converting to strings for displaying would give best results, but I haven't really looked at where errors propagate through the code.

samueltardieu:
Rounding is used beacuse different formats give different precisions. For example, if you give a "degrees minutes" format with 4 decimals, the precision is `1/(60*1000)` degrees, which is not the same precision that you would have for "degrees minutes seconds" precision with 3 decimals and `1/(60*60*100)` degrees.

It means that it is best to re-round before giving such a representation, otherwise we will end up with some reduced latest decimal in some cases.

I'll make the constructor private.

PR #6141, opened 2016-11-17
Cancel network requests on non-UI thread

samueltardieu:


samueltardieu:
Do not merge yet as it may not be necessary.

main/src/cgeo/geocaching/utils/RxOkHttpUtils.java
```diff
@@ -35,8 +36,16 @@ public void subscribe(final SingleEmitter<Response> singleEmitter) throws Except
                 singleEmitter.setDisposable(Disposables.fromRunnable(new Runnable() {
                     @Override
                     public void run() {
-                        if (!completed.get()) {
-                            call.cancel();
+                        if (!completed.getAndSet(true)) {
+                            // Cancel request on a non-UI thread as operations such as close()
```
Bananeweizen: I suggest to replace this explicit scheduling by a call to a new method, and to mark that method @UiThread (from androidannotations). That is more easy to read and understand.

samueltardieu: I think it will not be merged. Moreover, I would rather have `@NonUiThread` since this must *not* happen on the UI thread. To be more complete, we do not know where the call to `dispose()` will come from, but we want the call to `cancel()` to be outside the UI thread because it has been a problem (strict mode triggering by mistake) in the past (which I think is fixed in the recent versions of OkHttp).

Anyway, I'm investigating more complete solutions.

Issue #6114, opened 2016-11-12
Live map handling became massively instable with latest nightlies 


Lineflyer:
More a reminder to myself to take a closer look later

Either its the server once again or something within cgeo.
E.g. the following makes cgeo crash in a high percentage of cases:
- Open live map
- While map is still loading click on a saved cache
- Click to open details from popup

Using mapsforge beta with nightly 2016.11.12

samueltardieu:
It might be due to the changes of RxJava major version and the code modifications that I had to do. I need traces to assert that though.

Lineflyer:
Will investigate next week.It became more stable after deactivating mapsforge beta AFAICS.


Von meinem Smartphone gesendet.
-------- Ursprüngliche Nachricht --------Von: Samuel Tardieu <notifications@github.com> Datum: 12.11.16  15:47  (GMT+01:00) An: cgeo/cgeo <cgeo@noreply.github.com> Cc: Lars <gc@droescher.eu>, Author <author@noreply.github.com> Betreff: Re: [cgeo/cgeo] Live map handling became massively instable with latest nightlies (#6114) 
It might be due to the changes of RxJava major version and the code modifications that I had to do. I need traces to assert that though. 

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/cgeo/cgeo","title":"cgeo/cgeo","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/cgeo/cgeo"}},"updates":{"snippets":[{"icon":"PERSON","message":"@samueltardieu in #6114: It might be due to the changes of RxJava major version and the code modifications that I had to do. I need traces to assert that though. "}],"action":{"name":"View Issue","url":"https://github.com/cgeo/cgeo/issues/6114#issuecomment-260126128"}}}

Lineflyer:
[...]

PR #6144, opened 2016-11-18
Do not let uncaught OkHttp exceptions crash the application

samueltardieu:
The OkHttp exceptions thrown in background threads are now intercepted
by an UncaughtExceptionHandler.

Also, this handler checks the "debug" settings dynamically rather than
installing or uninstalling itself when they are toggled.

This is part of the fix for #6114.


samueltardieu:
retest this please

samueltardieu:
retest this please

PR #6163, opened 2016-11-27
fix #5359: upgrade to assertj 2.x

### LINKED ISSUE #5359: Test assertJ 2.3
We fixed our dependencies to only use AssertJ 1.x, as 2.x is not compatible with Android. Since the latest release 2.3 it might be compatible again: http://joel-costigliola.github.io/assertj/assertj-core-news.html#assertj-core-2.3.0, so we should try using it.

Bananeweizen:


samueltardieu:
retest this please

Bananeweizen:
unfortunately this needs some additional refactoring. all the tests using isEqualToComparingFieldByField are broken, as far as I can see. Interestingly, the assertion failures are correct behaviour at least in the one case I debugged so far.

PR #6165, opened 2016-11-27
fixes #6164 import loc files on correct list

### LINKED ISSUE #6164: LOC files always imported to standard list
- Create a new list
- Select import and select a loc file (in my case it contained 25 gc benchmarks)

Waypoints are imported to default list

Waypoint should be imported to active list

##### Version of c:geo used:
2016.11.26-NB

Yes

pstorch:


Bananeweizen:
Looks better than what I had in mind before Telekom network died yesterday. :)

PR #6183, opened 2016-12-09
fixes #6181 history not working reliably

### LINKED ISSUE #6181: History not working reliably
This is more or less a reminder to myself (and others who would like to test):

Some user reports for the current RC version (2016.11.28-RC) tell, that there are problems in the history:
- Offline logged caches not visible
- History cannot be emptied
- History is emptied but cache count in title bar still shows x caches on list

I cannot reproduce any of them right now, but as we had a change in this area I think there might be some leftover bug under special conditions.

I will try to gather more information from users and try to test some things myself.

pstorch:
- fix History count
- fix online Logs not visible in History
- fix "clearOfflineLogs" cleared also all visited

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,15 @@ Such a procedure is [described in the wiki](https://github.com/cgeo/cgeo/wiki/Ho
 Make sure to use Java 8 for your IDE and build process. Some of the involved tools require it, even though the source code is Java 7 only.
 
 ### Eclipse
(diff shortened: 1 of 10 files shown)


Lineflyer:
this should go to release instead of master please.

pstorch:
@Lineflyer of course, missed that. Just changed it here on GitHub, but that leads to unrelated commits. I'll rebase it locally.

pstorch:
I'll send a new PR

PR #6190, opened 2016-12-13
Minsdk14

pstorch:
@samueltardieu I fixed the last compile error for the minSdk14 branch. I had to pull in some changes from master because of other compile errors. I rebased, because this is my default strategy. Later on, just before pushing, I realized that I would have to force push on this public branch. And that is something to avoid. I don't want to destroy any work from you. So I decided to push to my own fork and send a PR.
I'm not an expert in Android SDK levels. I just solved the compile error and tested that the map is still working. It's not much change, so if done wrong, it doesn't hurt to throw it away.
But if it's ok, you can merge my PR into the minSdk14 branch.


pstorch:
The map test are still falling, because of the recent icon changes.

pstorch:
I've put my changes on top of minsdk14 branch

PR #6192, opened 2016-12-15
fixes #6185 list count for non detailed caches

### LINKED ISSUE #6185: Non-detailed caches incorrectly counted for amount of stored caches
- Remember the amount of saved caches on your lists as shown on the mainscreen as X
- Open live map
- Click on a non-detailed cache (with orange circle)
- From the popup select Log Visist 
- Enter log text and press back to save the log

- Cache icon has now markers for offline logged and saved(?)
- Saved caches amount on main screen is now x+1
- Cache is not visible in lists
- Cache is visible in history
- Opening this cache in history will show the non-detailed status: Empty description, Empty logbook
- Cache details screen shows "Saved on device 11746 days ago"

I don't know enough about the DB structure to judge the best way of fixing this.
However from a user perspective:
- It should show up in the history (as it does)
- It should not count to the amount of total caches stored in lists as it is not part of any list contained there
- It should not show up with saved icon on map

##### Version of c:geo used:
2016.12.10-NB

Yes

This seems to be **not** caused by the change of @pstorch (ded3fb7c41093590b2586ae7e4d756291c9ee3db) as the behaviour is the same with this fix (fresh RC) and without this fix (current nightly).

pstorch:


pstorch:
Any objection to merge?

pstorch:
I'll open a new PR for release branch.

PR #6193, opened 2016-12-15
fixes #5949 delete caches from history and search result list

### LINKED ISSUE #5949: Deleting geocache in "History" don't work (+related problem)
- open history
- long press on a cache
- press "drop" > Message "Deleted 1 Cache"
- Cache is still stored (in history and everywhere else)
- Cache should be deleted (in all lists)
##### Version of c:geo used:

2016.09.06-RC

Yes
##### System information:

Device: SM-G800F (kminiltexx, samsung)
Android version: 4.4.2
Android build: KOT49H.G800FXXU1ANJ2
Cgeo version: 2016.09.06-RC
Google Play services: enabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: true
Map strategy: auto
System language: de_DE
Log date format: 12/09/2016
Debug mode active: no (cgeo forced to English)
Geocaching sites enabled:
Installed cgeo plugins: calendar

There is an **other** probably related Problem: For normal I delete all caches after export field notes in history list from the device by marking it with select mode but that is also not possible any more. In select mode there is no more "drop" or "delete" for this list. I don't want found caches on the device and I don't like to search for them in other lists every time. (Possible with all lists > filter > with offline logs - but deleting in history was much easier.)

pstorch:
Caches removed via long press from history or search result will be dropped from all lists.

BUT: caches with offline logs will still appear (stay) on history. We still don't have an agreement to delete offline Logs in these cases.

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -1392,16 +1392,20 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
 
         @Override
         protected void doCommand() {
-            if (listId == PseudoList.ALL_LIST.id) {
+            if (appliesToAllLists()) {
                 oldCachesLists.putAll(DataStore.markDropped(getCaches()));
             } else {
                 DataStore.removeFromList(getCaches(), listId);
             }
         }
 
+        private boolean appliesToAllLists() {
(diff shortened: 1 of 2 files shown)


pstorch:
Any objection to merge?

Lineflyer:
From intended functionality point of view I have no objection to merge.
Shouldn't this be better suited to be merged into `release`?

pstorch:
ok, I'll open a new PR for release, then merge into release and master

PR #6210, opened 2016-12-28
fixes #6208 add geochecker gc-apps.com

### LINKED ISSUE #6208: Add geochecker link also for gc-apps.com
Curently we provide a menu item "Open Geochecker" if the cache listing contains a checker link.
Since some months there is a new officially allowed checker of www.gc-apps.com

We should also scan for their links and make the menu item working for that.

Example: GC3EQ4W

pstorch:


Lineflyer:
Should we include this on `release`? 
I think it is uncritical.

pstorch:
@Lineflyer ok, I'll change to to release.

PR #6213, opened 2016-12-28
fixes #6208 add geochecker gc-apps.com

### LINKED ISSUE #6208: Add geochecker link also for gc-apps.com

pstorch:


pstorch:
As it it so trivial, I'll merge it right away.

PR #6215, opened 2016-12-29
fix #6212 icondecoder tests

### LINKED ISSUE #6212: Fix IconDecoder tests
Following the recent changes to `IconDecoder.java` the corresponding tests need to be fixed as they are currently failing.

pstorch:
I've added some new maps with different types and zoom levels.
@mucek4 please have a look if these tests are reasonable


Lineflyer:
Thanks for your effort. Very much appreciated!

tests/src-android/cgeo/geocaching/connector/gc/IconDecoderTest.java
```diff
@@ -57,389 +42,127 @@ private static Geocache parseMapPNG(final Bitmap bitmap, final int x, final int
...
-        assertThat(parseMapPNG(bitmap, 181, 116, 13).getType()).isEqualTo(MYSTERY);
-        assertThat(parseMapPNG(bitmap, 118, 230, 13).getType()).isEqualTo(MULTI);
+    public void testParseMap1() {
+        final Bitmap bitmap = getBitmap(map1_z13);
+        assertThat(parseMapPNG(bitmap, 76, 30, 13).isFound()).isTrue();
+        assertThat(parseMapPNG(bitmap, 137, 98, 13).getType()).isEqualTo(TRADITIONAL);
+        assertThat(parseMapPNG(bitmap, 127, 76, 13).getType()).isEqualTo(TRADITIONAL);
+        assertThat(parseMapPNG(bitmap, 188, 114, 13).getType()).isEqualTo(MULTI);
+        assertThat(parseMapPNG(bitmap, 188, 174, 13).getType()).isEqualTo(MYSTERY);
+        assertThat(parseMapPNG(bitmap, 120, 93, 13).getType()).isEqualTo(UNKNOWN); // disabled TRADITIONAL
```
mucek4: Hm. IconDecored should get the disabled cache too. I wonder why it's unknown.

mucek4:
I say we merge, so tests pass. But there is some work to be done on that for sure.
Disabled caches shall also work (I have to check) and wait for some feedback from users.

Lineflyer:
Current tests show, that disabled caches are not identified and I doubt that this is possible by color matching. They are all grey.

However this is not critical for me. Most times on the map in c:geo only the disabled caches remain as ??-type and after clicking on them the user will see that they are disabled.

PR #6216, opened 2016-12-30
fix #6191: make inventory getter immutable non null

### LINKED ISSUE #6191: Crash when opening logbook of certain cache
- Open GC1YX6F
- Swipe to logbook

Crash

Normal operation

##### Version of c:geo used:
2016.12.09-RC / 2016.12.13-NB

No

Happend for a user on FB.
I can reproduce it on 1 out of 5 devices, sadly only on the one which offers no logging.

Can someone reproduce and provide a log?

Bananeweizen:


Bananeweizen:
retest this please

Issue #6186, opened 2016-12-11
History cannot be emptied


Lineflyer:
##### Detailed steps causing the problem:
- Start with empty database
- Save some caches to your lists
- Enter list, open a cache and write an offline log for one of the saved caches
- Offline log status is visible in lists 
- Cache is shown in history
- Open history and select "Clear history"

##### Actual behavior after performing these steps:
Nothing happens

##### Expected behavior after performing these steps:
Hsistory is emptied

##### Version of c:geo used:
2016.12.11-RC (including ded3fb7c41093590b2586ae7e4d756291c9ee3db)

##### Is the problem reproducible for you?
Yes

##### Comment:
Related to #6181

pstorch:
@Lineflyer for deleting caches with offline logs there is an extra menu entry. "clear offline logs". This is the reason why I think they should be separated.

Lineflyer:
Yes, that has become quite unintuitive. 

For deleting offline logs I would expect the cache to stay in history but only have the logs deleted  (did not check that yet).
For emptying history I would expect to remove them from history but keep the associated log.
Otherwise we should simply remove that emptying of history?!

pstorch:
We had the discussion somewhere, that we must prevent a user from accidentally deleting the offline logs. That's why it is the other way around.  When you clear history, the ones with offline logs stay. But if you by purpose choose to clear offline logs, then the history for those is deleted (resetted), too.

pstorch:
Anyway, it's confusing both ways.

rsudev:
For me the history is much less useful with the current semantics as it was before. I guess the changes were mainly introduced as the behaviour was unclear due to some quirks and bugs.
Previously (in theory), every cache that was logged in c:geo (online or offline, regardless of the log type, maybe not for WA) was shown on the history. It stayed there, until the history was cleared manually. You could upload field-note from there and refresh all or a selection of the caches in the history. And you could delete all offline logs (e.g. after uploading field-notes, logging in the browser and refreshing the caches in question).
To me, that behaviour was useful, as I could easily reference from there to caches I recently visited, and as it was a virtual list, it didn't require additional storage.
I think it would be good to go back to the drawing board and define an overall view how the history should work, before continuing to address the current quirks,

pstorch:
I think it just seamed to be clear in the past, but it was as mixed up as it is now. It was just not visible.
[...]

PR #6228, opened 2017-01-04
fix #6226, fix #6129 cleanup orphaned records in db

### LINKED ISSUE #6129: DataStore.cleanIfNeeded() search for `detailed < timestampString`
While looking through the DataStore claas I stumbled upon this code block in `cleanIfNeeded()` method:

```
                    if (version != Settings.getVersion()) {
                        queryToColl(dbTableCaches,
                                new String[]{"geocode"},
                                "geocode NOT IN (SELECT DISTINCT (geocode) FROM " + dbTableCachesLists + ")",
                                null,
                                null,
                                null,
                                geocodes,
                                GET_STRING_0);
                    } else {
                        final long timestamp = System.currentTimeMillis() - DAYS_AFTER_CACHE_IS_DELETED;
                        final String timestampString = Long.toString(timestamp);
                        queryToColl(dbTableCaches,
                                new String[]{"geocode"},
                                "detailed < ? AND detailedupdate < ? AND visiteddate < ? AND geocode NOT IN (SELECT DISTINCT (geocode) FROM " + dbTableCachesLists + ")",
                                new String[]{timestampString, timestampString, timestampString},
                                null,
                                null,
                                geocodes,
                                GET_STRING_0);
                    }
```

If we have a version upgrade, we select all caches not on any list. If it is the same version we select caches not on any list and not updated or logged since DAYS_AFTER_CACHE_IS_DELETED. Additionally in the second select there is also the `detailed` column which is also compared to the `timestampString`. `detailed` can have the values `0` and `1` and therefore is always lower than a timestamp.
I guess we can remove the `detailed` column from the where clause.

The second question is: shouldn't we always use the second version? Why should we remove temporary caches on a version upgrade faster than normal?

### LINKED ISSUE #6226: Database cleanup of orphaned records
Follow up from https://github.com/cgeo/cgeo/issues/6186#issuecomment-270253650.
It turned out that due to the lack of foreign keys in the DB it happens that geocaches got deleted from `cg_caches` without their related records in dependent tables like `cg_logs`, `cg_waypoints`, ...

The `DataStore.cleanIfNeeded()` is the housekeeping method which regulary (on startup) cleans up the database. This should be extended to also clean up these orphaned records as well.

pstorch:
With this cleanup I could reduce the 68 MB big sqlite file from https://github.com/cgeo/cgeo/issues/6186#issuecomment-270253650 to 48 MB. There were a lot of orphaned records left in the DB.

Adding FK constraints would be a next step.

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -584,8 +584,6 @@ public class DataStore {
                 }
 
                 if (oldVersion > 0) {
-                    db.execSQL("DELETE FROM " + dbTableCaches + " WHERE reason = 0");
-
                     if (oldVersion < 52) { // upgrade to 52
                         try {
                             db.execSQL(dbCreateSearchDestinationHistory);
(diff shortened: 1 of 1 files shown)


pstorch:
OK, let's get some feedback from the nightlies

PR #6237, opened 2017-01-09
fix #6106 remove fastes live map strategy

### LINKED ISSUE #6106: Remove fastest strategy on live map
I suggest to remove the selectable live map strategy for the following reasons:

- Users I spoke with are not even aware of this option 
- Users I spoke with always prefer to see the cache type and get most information out of the map
- Networks have become faster
- There are often users on support and in ratings claiming that they only see ?? caches as they accidentally selected "Fastest" thinking it would be the same but just faster

- Make things simpler, less options

The speed dependent solution (Current default) should be the fixed strategy.

pstorch:


main/src/cgeo/geocaching/maps/LivemapStrategy.java
```diff
@@ -9,7 +9,6 @@
  * Defines the strategy for the Live Map
  */
 public enum LivemapStrategy {
-    FASTEST(1, EnumSet.of(Flag.LOAD_TILES), R.string.map_strategy_fastest),
```
samueltardieu: Maybe a note explaining why we don't (and can't) have a value of 1 anymore would be useful.

Bananeweizen:
@samueltardieu Actually the deleted value will already be replaced (by AUTO) on loading from the Settings (due to the enum lookup encapsulated in getLiveMapStrategy). That's the main reason we have this explicit lookup routine in every enum we serialize. So the code and the UI will see the same.

main/src/cgeo/geocaching/maps/LivemapStrategy.java
pstorch: Yes, I'll add a comment here. First I thought to keep the commented out code there, but decided against it, because it is bad practice. But a comment that there was another option which is not supported anymore would be good.

samueltardieu:
@Bananeweizen Indeed, I didn't remember the use of `getById` at load time.

Lineflyer:
Sorry for the maybe unqualified question:
Isn't there now a leftover in the code with respect to the `FASTEST` strategy at the point where the IconDecoder is called or not (which in fact is the difference between `FAST` and `FASTEST`)

pstorch:
@Lineflyer FAST is parsing the tiles with IconDecoder.

Lineflyer:
> @Lineflyer FAST is parsing the tiles with IconDecoder.

Yes, thats what I meant. Isn't there a code like
If x!=FASTEST then goto Icondecoder.java

I can't see anything like this removed with this PR...still of course it would no longer be called. I just wanted to test if I understand it correctly.

Sorry for my basic like coding syntax example ;)

main/src/cgeo/geocaching/maps/LivemapStrategy.java
pstorch: comment added

PR #6244, opened 2017-01-10
fix #6232 one-click offline log does not save cache

### LINKED ISSUE #6232: One-click offline log does not save cache
I found a difference depending of whether or not "One-click offline log" is active.
Can someone explain:

- Enable "One-click offline log" in Menu - Settings - Logging
- Open live map, set strategy to "Fast"
- Select an orange circled (= non-detailed) cache
- From the popup select "One-click offline log" and select found

- Cache does not get stored

- Should be treated in the same way as when using the online log function (see #6231)

pstorch:
Saving the cache was not a problem, but updating the map popup took me quite some time.
Closing the Dialog of the one-click offline log didn't trigger the `onResume()` method of the `CachePopupFragment`. I didn't want to extend the `LoggingUI` with yet another view to update, so I refactored that class and used an `OnDismissListener` which can be provided. With this I could also remove the WeakReference of the `CacheListAdapter.ViewHolder` which was used to perform an update, too.

Maybe @Bananeweizen and/or @samueltardieu can have a quick look if this solution is ok.

### DIFF
--- a/main/src/cgeo/geocaching/AbstractDialogFragment.java
+++ b/main/src/cgeo/geocaching/AbstractDialogFragment.java
@@ -337,7 +337,12 @@ public abstract class AbstractDialogFragment extends DialogFragment implements C
         if (CacheMenuHandler.onMenuItemSelected(item, this, cache)) {
             return true;
         }
-        if (LoggingUI.onMenuItemSelected(item, getActivity(), cache)) {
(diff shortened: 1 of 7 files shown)


pstorch:
Any objection to merge?

samueltardieu:
None from me.

PR #6245, opened 2017-01-10
fix #6189 store with preselection on long press

### LINKED ISSUE #6189: Distinction between "store" and "manage" buttons
- save a cache to a list
- try to save another cache

- the checkbox in which the previously saved cache was stored is ticked

- in the dialogue, none of the checkboxen is ticked
  (unless the cache was already in any lists, in which case those should be ticked)

##### Version of c:geo used:

nightly build 2016-12-12

note it didn't occur in some previous (november?) nightly build

Yes

##### System information:

Android 4.4 Liquidsmooth; I could give the full info if necessary, but this is a UI issue, and therefore unlikely to be needed.

pstorch:
Changes the store cache preselection of lists from "remember with timeout" to "preselect on long press".

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -610,7 +610,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
                 dropCache();
                 return true;
             case R.id.menu_store_in_list:
-                storeCache();
+                storeCache(false);
                 return true;
             case R.id.menu_refresh:
                 refreshCache();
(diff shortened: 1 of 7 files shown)


SammysHP:
I wonder if there's a solution for the menus. Maybe a "load previous selection" button in the list selection dialog?

pstorch:
I see three possibilities:

- add a long press option for the menu items as well 
do we have that anywhere? If not it would be a new paradigm for menu items.
- add yet another menu item, e.g. "Store on previous selected lists"
there are already a lot of menu items in CacheDetailsActivity
- add a third button on the list selection dialog
that doesn't save you many clicks.

All options have some sort of downside. If not really essential, I would like to limit this special function on the Store button. Opening the menu instead of using the Store button is one click more anyway.

pstorch:
I added a `Do not merge` label, because the discussion in #6189 is not finalized yet.

pstorch:
I close here and send a new PR

PR #6250, opened 2017-01-11
fix #6249 history jumps to top when returning from CacheDetails

### LINKED ISSUE #6249: History list jumps to top after returning from CacheDetails
- have more caches on history than fit on the screen
- scroll down and open one of the cache details
- close the details

- return to history list jumping to the top of the list

- returning to history and stay were you were before

##### Version of c:geo used:
Release branch - 6daf3fc67f40f186128f5c83cb4e15cab6f63638

Yes

pstorch:


Lineflyer:
This equals the history list behaviour to the remaining lists behaviour. Correct?

pstorch:
@Lineflyer yes, the other lists also keep their vertical position.

Lineflyer:
OK from my side to include into release then.

PR #6269, opened 2017-01-17
fix #6259: do not save cookie store with Google

### LINKED ISSUE #6259: Login issue (Unknown error) Code=403
Recently I started asking users on support to do this procedure when they report "Login failed: Unknown error" and confirmed to have checked all other things mentioned in the FAQ:

- Go to c:geo Menu - Settings - Service - geocaching.com
- Disable the service completely (disabling the tick box)
- Reboot device
- Go to c:geo Menu - Settings - Service - geocaching.com
- Enable the service again

In all cases users reported back, that it was now working normal (Login OK).

Can someone imagine any root cause in the c:geo code causing this.
I personally suspect problems in upgrade cases or data restore after reinstallation (Google backup)...IMHO this is somehow enabled in c:geo to store this data in Google.

samueltardieu:
Keeping the cookie store separate from the preferences allows not to
backup it with the backup manager.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -119,7 +119,6 @@
     <string translatable="false" name="pref_memberstatus">memberstatus</string>
     <string translatable="false" name="pref_coordinputformat">coordinputformat</string>
     <string translatable="false" name="pref_gccustomdate">gccustomdate</string>
-    <string translatable="false" name="pref_cookiestore">cookiestore</string>
     <string translatable="false" name="pref_googleplayservices">googleplayservices</string>
     <string translatable="false" name="pref_lowpowermode">lowpowerlocation</string>
     <string translatable="false" name="pref_lastdetailspage">lastdetailspage</string>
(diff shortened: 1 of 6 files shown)


pstorch:
Interesting. Didn't know about the concept of preferences backup via google. Had to look it up. 
But the solution sounds plausible.

pstorch:
Should we also reset the cookies on `CredentialPreference.revokeAuthorization()`. 
The `AbstractLogin` class has a `resetLoginStatus()` method.

Lineflyer:
As written in #6259 I suggest to put this to `master` only (and maybe also discuss the background and further steps).

pstorch:
@Lineflyer the target branch of a PR can also be changed via the `edit` button on top of the page. Ok, this doesn't work in all cases, but most of the time it is not needed to close and open a new PR.

Lineflyer:
Then lets @samueltardieu do the needful. :)

samueltardieu:
> Should we also reset the cookies on `CredentialPreference.revokeAuthorization()`?

Good idea, I'll do that in case of revocation of gc.com authorization.

PR #6270, opened 2017-01-17
Clear cookies when revoking an authorization

samueltardieu:
Suggested in #6269.

### DIFF
--- a/main/src/cgeo/geocaching/settings/CredentialsPreference.java
+++ b/main/src/cgeo/geocaching/settings/CredentialsPreference.java
@@ -7,6 +7,7 @@ import cgeo.geocaching.connector.capability.ICredentials;
 import cgeo.geocaching.connector.ec.ECConnector;
 import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.gcvote.GCVote;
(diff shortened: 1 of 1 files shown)


Lineflyer:
Which user action triggers the mentioned "revoking of authorization"?

pstorch:
It's triggerd by long press on `Authorize..' or your avatar.

main/src/cgeo/geocaching/settings/CredentialsPreference.java
```diff
@@ -135,6 +136,9 @@ protected boolean isAuthorized() {
 
     @Override
     protected void revokeAuthorization() {
+        if (credentialsMapping == CredentialActivityMapping.GEOCACHING) {
+            Cookies.clearCookies();
```
Bananeweizen: Just wondering if the clearing has any negative effect on the other authorizations? But generally I'm completely fine with the change.

samueltardieu: It shouldn't. Cookies store session information for sites not using OAuth. Clearing the cookies will go through the classical login phase (a few extra requests) when data is next requested, but no interaction will need to take place.

PR #6271, opened 2017-01-18
fix #6189 fast-store with preselection on long press

### LINKED ISSUE #6189: Distinction between "store" and "manage" buttons

pstorch:
This (proposed) change adds a third button on the list selection dialog to preselect the lists from last store.
It add a `fast-store` function on the store buttons triggered by a long press. The `fast-store` uses the last remembered list selection and stores immediately without showing a list selection. It falls back to show the dialog if there are no remembered lists.
The timeout for remembering the lists is removed.
The store menu items don't offer the `fast-store` option, see comment https://github.com/cgeo/cgeo/issues/6189#issuecomment-273444179

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -789,6 +789,7 @@
     <string name="cache_list_text">List:</string>
     <string name="cache_list_change">Move</string>
     <string name="cache_list_unknown">Not in a list</string>
+    <string name="cache_list_select_last">Last Selection</string>
     <string name="cache_images">Images</string>
     <string name="cache_waypoints">Waypoints</string>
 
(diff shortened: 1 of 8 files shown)


pstorch:
Updated the fallback to show the dialog instead of using the default `Stored` list.

Lineflyer:
From what you wrote I like the solution. I suggest to try it out (but on `master`, not on `release`).

pstorch:
Ok, I'll change it to `master`

PR #6275, opened 2017-01-19
fix #1249 extract waypoints from cache description

### LINKED ISSUE #1249: Automate copying of waypoints from cache description
Often caches have waypoints described inside cache text but not available as cache waypoint.

For example, have a look at GCM4Y8.
You'll found several coordinates inside cache text description but no official cache waypoints.

It would be nice if c:geo could create waypoints automatically from the coordinates inside the text.

I don't know: tapping and holding over a coordinate inside text will popup a dialog asking for a description (nice to have: description already populated with 3-5 words from text before the coordinates)?

The same feature would be useful also while reading the logs.

pstorch:
Added an item to the text selection context menu. It looks like this:
screenshot_20170119-090001

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractActivity.java
@@ -1,6 +1,7 @@
 package cgeo.geocaching.activity;
 
 import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.Intents;
 import cgeo.geocaching.R;
 import cgeo.geocaching.compatibility.Compatibility;
 import cgeo.geocaching.enumerations.CacheType;
@@ -23,6 +24,7 @@ import android.os.Bundle;
 import android.support.annotation.LayoutRes;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.v4.content.LocalBroadcastManager;
 import android.support.v7.app.ActionBarActivity;
 import android.support.v7.view.ActionMode;
 import android.view.Menu;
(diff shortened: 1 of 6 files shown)


main/src/cgeo/geocaching/activity/AbstractActivity.java
```diff
@@ -179,7 +181,7 @@ protected void buildDetailsContextMenu(final ActionMode actionMode, final Menu m
         menu.findItem(R.id.menu_translate_to_english).setVisible(!copyOnly && !localeIsEnglish);
     }
 
-    protected boolean onClipboardItemSelected(@NonNull final ActionMode actionMode, final MenuItem item, final CharSequence clickedItemText) {
+    protected boolean onClipboardItemSelected(@NonNull final ActionMode actionMode, final MenuItem item, final CharSequence clickedItemText, final Geocache cache) {
```
samueltardieu: Would you mind making the added parameter `@Nullable` so that we know from the prototype that `null` is an acceptable argument?

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1510,10 +1510,14 @@ public Waypoint getWaypointById(final int id) {
      * Detect coordinates in the personal note and convert them to user defined waypoints. Works by rule of thumb.
      */
     public boolean parseWaypointsFromNote() {
+        return parseWaypointsFromText(getPersonalNote(), false);
+    }
+
+    public boolean parseWaypointsFromText(final String text, final boolean updateDb) {
```
samueltardieu: Since we take care of a `null` entry, I'd use `@Nullable` here as well for `text`.

pstorch:
@samueltardieu added `@Nullable` annotations as suggested

PR #6282, opened 2017-01-24
Waypoint naming when parsing text + unit test

pstorch:
Initially I just wanted to change the name of the parsed waypoints from the cache description from "Personal note #" to "Description #" as suggested in  https://github.com/cgeo/cgeo/issues/1249#issuecomment-274693217, but it got a bit more.
I wanted to add this to a UnitTest. Then I stumbled upon the `WaypointTest` class which didn't run. It took a while till I found out it was (accidentally) excluded by disabling the Espresso tests. The exclude pattern was to broad. Then I had to fix the test, because the UTM coordinate parsing caused a false match here. While at it I refactored a bit. Splitting tests, renaming methods to better reflect what they are doing.

Usually I prefer smaller changes, but here I came from one topic to the next.
I hope it's still ok, otherwise please object.

So, here is a summary of the changes:
- Waypoints extracted from Cache description now have "Description #" as name
- `WaypointTest` is reactivated (was excluded together with Espresso tests)
- `WaypointTest` extended and refactored
- `Geocache.parseWaypointsFrom...()` is renamed to `Geocache.addWaypointsFrom...()` to better reflect what it is doing
- `UTMPoint` regex pattern narrowed to prevent false positive from `WaypointTest`
- Added another test to `GeocacheText` which tests that duplicate waypoints don't overwrite waypoints which are already there.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -123,7 +123,9 @@ android {
                 // TODO: espresso causes some dependency conflicts, therefore disable espresso tests temporarily
                 exclude '**/*EspressoTest.java'
                 exclude '**/*LogTrackableActivityTest.java'
-                exclude '**/*Waypoint*Test.java'
+                // there is a valid test WaypointTest which should run, so the pattern must be more specific
+                exclude '**/A*Waypoint*Test.java'
+                exclude '**/EditWaypoint*Test.java'
             }
             resources.srcDirs = ['../tests/src']
             res.srcDirs = ['../tests/res']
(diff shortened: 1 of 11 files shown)


pstorch:
Found another test `cgeo.geocaching.connector.gc.WaypointsTest` which was accidentally disabled. It didn't run (anymore). There was a problem with instantiating the `DisposableHandler`. For the test to run, the handler is not needed, so I replaced it with `null`.

pstorch:
I'll merge this, because I need the working `WaypointTest` in my other task.

PR #6283, opened 2017-01-27
fix #5353, fix #6223 waypoint storage

### LINKED ISSUE #5353: Idea: Change waypoint storage
In order to untangle the current mess with changed waypoint descriptions and such, I would suggest to separate the waypoint description as loaded from the source (website, OKAPI, gpx) from the user additions. Than we can clearly detect website changes and update the local waypoint accordingly without endangering user-entered data.
As a migration path I would suggest to treat the current description as user supplied and remove if it is identical to the website description on next update.
Comments?

### LINKED ISSUE #6223: Waypoint description not updated if new description is shorter
- Open a GC cache which has a waypoint that contains a description
- Check that the waypoint is identical to the listing on the website
- Ask the owner to change the description of the waypoint
- Refresh cache in c:geo

- Waypoint will keep showing the old description

- Waypoint should be updated with the new description

##### Version of c:geo used:
2016.12.28

Yes

I can volunteer with my owned cache to help debugging this

Set to "High" as this might be not the standard case but still can make it impossible for users to find a cache (e.g. given they stored a cache long time ago and don't get the updated info 

Similar to #3650

pstorch:
Here is a proposal to change the waypoint storage to preserve local changes.

- Name (title): only editable if it is a user defined waypoint
- Coords: only editable if it is a user defined waypoint or if no coords were present (like finals)
- Note: only editable if it is a user defined waypoint
- User note: new field which is editable for all waypoints and not affected by server side changes

Migration:
Because we don't know which notes were edited locally, here I copy all `notes` into `user_note` and set the `note` to empty. On refresh the `note` field is set to the server side value and compared with the `user_note` field. If identical, the `user_note` field is emptied.

While testing this, it worked good for me. But of course I would like to have some feedback.

### DIFF
--- a/main/res/layout/editwaypoint_activity.xml
+++ b/main/res/layout/editwaypoint_activity.xml
@@ -71,6 +71,14 @@
             android:layout_height="wrap_content"
             android:hint="@string/waypoint_note"
             android:inputType="textMultiLine|textCapSentences"
+            android:minLines="5"/>
+
+        <EditText
+            android:id="@+id/user_note"
+            style="@style/edittext_full"
+            android:layout_height="wrap_content"
+            android:hint="@string/waypoint_user_note"
+            android:inputType="textMultiLine|textCapSentences"
             android:minLines="5" />
 
         <CheckBox
--- a/main/res/layout/waypoint_item.xml
+++ b/main/res/layout/waypoint_item.xml
@@ -95,6 +95,19 @@
         tools:visiblity="visible"
         tools:text="Note"/>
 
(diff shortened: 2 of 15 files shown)


pstorch:
My idea with the coords does only work on first edit. I need to think about it a bit more.

pstorch:
OK, now a flag indicates if the original coordinates were empty, to control which WP coordinate changes to allow.

pstorch:
Rebased with `master` and added unit tests

PR #6305, opened 2017-02-07
fix #5476 locally changed coords replaced by server coords

### LINKED ISSUE #5476: Locally changed coords replaced by listing coords on refresh
- Open a saved cache
- Create a waypoint, select to change coords locally to listing coords
- Refresh the cache

The local change is lost and replaced again by the listing coords retrieved during refresh

As the user defined change was done, it is not correct that this data is lost by refreshing. 
Not all users will understand this in beforehand and loss of data will appear.
###### Version of c:geo used:

2016.02.23-RC

Yes

Not relevant

Follow up of #5439 

Comment by @blochi79:
Now we have a follow problem. ;-( The local header coordinates are overwritten if the cache is refreshed. The method for refreshing a cache needs also to be adapted for the case that the local header coordinates have been corrected. Then the original coordinates in waypoints must be updated with the Listing coordinates instead of the header coordinates.
[...]
I often perform bulk refresh of mixed lists that contains also traditional caches. In this case I don't want to lose the corrected coordinates of letterbox or Wherigo caches. I did not know that it is the default behavior of c:geo to overwrite the local corrected coordinates with the listing coordinates. I think the current issue can be closed and the behavior of overwriting corrected coordinates while refreshing a cache may be discussed as subject of another issue. Sorry for the confusion.

pstorch:
If coords are only locally changed, the local changes are preserved now. The original coords are updated in the ORIGINAL waypoint.

If coords are changed locally and server side, the server side coords win.

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -281,7 +281,7 @@ public class Geocache implements IWaypoint {
         if (StringUtils.isBlank(getLocation())) {
             location = other.getLocation();
         }
-        coords = UncertainProperty.getMergedProperty(coords, other.coords);
+
         // don't use StringUtils.isBlank here. Otherwise we cannot recognize a note which was deleted on GC
         if (personalNote == null) {
             personalNote = other.personalNote;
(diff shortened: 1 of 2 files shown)


Lineflyer:
> If coords are changed locally and server side, the server side coords win.

+1

Issue #6302, opened 2017-02-06
GPS and Magnetic Compass Used box appears when rotated


FreddieSanchez:
##### Detailed steps causing the problem:
- Have a phone without Magnetometer (Compass) sensor (Moto G4).
- Select a cache
- open the compass
- Do not see the "GPS and Magnetic Compass Used".
- rotate the phone from portrait to landscape

##### Actual behavior after performing these steps:
- The  "GPS and Magnetic Compass Used" box appears when rotated to landscape and remains after rotated back to portrait mode.

##### Expected behavior after performing these steps:
- The  "GPS and Magnetic Compass Used" box should not appear when rotated to landscape and should not appear when rotated back to portrait mode.

##### Version of c:geo used:

2017.01.15

##### Is the problem reproducible for you?
Yes

##### System information:
```
--- System information ---
Device: Moto G (4) (athene, motorola)
Android version: 6.0.1
Android build: MPJ24.139-64
Cgeo version: 2017.01.15
Google Play services: enabled - 10.0.84 (438-137749526)
Low power mode: inactive
Compass capabilities: no
Rotation vector sensor: absent
Orientation sensor: absent
Magnetometer & Accelerometer sensor: absent
Direction sensor used: magnetometer & accelerometer
Hide own/found: false
Map strategy: detailed
HW acceleration: enabled (default state)
System language: en_US
Log date format: 2/6/17
Debug mode active: no
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / PREMIUM
Geocaching.com date format: d.M.yyyy
Installed cgeo plugins:  none
--- End of system information ---
```

Lineflyer:
Sadly I don't have a device for testing this scenario. But due to the precise error description I think it might really be a bug.

FreddieSanchez:
I have this device, and can would like to try to fix it. Recently got the development environment setup.

PR #6310, opened 2017-02-10
Only display the toggle switch if the device has magnetic sensor support

FreddieSanchez:


FreddieSanchez:
This is fixing issue #6302 that I recently opened.

rsudev:
Please test

pstorch:
I don't have such a device either and couldn't find an emulator without magnetic compass to test it.
But from my point of view this change sounds reasonable.
We already hide the button in `onCreate()` if the device doesn't have compass capabilities but in `forceRefresh()` we set the status (which might cause the button to reappear).
So, from my point of view :+1:

pstorch:
@FreddieSanchez please add `fix #6302` to your commit message. Then the issue will be closed automatically when the PR gets merged.

FreddieSanchez:
That's exactly what happens.

When the device is rotated into landscape mode, the new layout is used and
the toggle button's visibility from portrait layout does not persist.

I'll update the commit message shortly. Thanks

rsudev:
Please test this

rsudev:
test this please

pstorch:
@rsudev do you want to trigger a new CI build?

retest this please

FreddieSanchez:
I tested this using the Moto G4. Displaying the compass for a cache, then rotating the device into landscape mode. Verifying the toggle button is not visible. And rotating it back to portrait mode. Again verifying the toggle button is not present. 

I also tested this using the emulator to verify the expected behavior of a device with the compass sensor.

pstorch:
@rsudev CI doesn't show up here, but according to the CI itself the PR was tested: http://ci.cgeo.org/job/cgeo%20pull%20request/722/

I think we can merge it.

rsudev:
Thanks for the contribution!

PR #6320, opened 2017-02-12
fix #6308 "Add Waypoint" button on top

### LINKED ISSUE #6308: Have "Add Waypoint" on top
If you are adding a lot of waypoints for a (multi)cache newly created waypoints are at the bottom of the list with the add button below the list. To add a new waypoint you always have to scroll down (swipe up) first to access the "Add waypoinr"-button. Same goes to access (or use) a newly created waypoint.

Suggestion is to create a setting to show waypoints in reverse chronological order (newest on top), with the "Add waypoint"-button right above the list, or not (as it currently is).

pstorch:


Lineflyer:
After creating #6308 I tend to remember that we had a discussion long time ago and for whatever reason it turned out, that having it on top is not possible....or was it, that we once had it on top and placed it at the end because for some reason.
However..lets try our this simple change, which looks straight forward to me.

pstorch:
Ok, let's try

PR #6321, opened 2017-02-12
Do not hide type by type parameter

UniQP:


pstorch:
I had to research a bit on this "hiding" type warning and what is the correct way to implement `Collection`.
But then I looked where and how we use the `GeoitemLayers` class and the only case where the `Collection` interface comes into play are the for loops in `AbstractCachesOverlay`. We actually don't call `toArray()` anywhere. 
Maybe it is better to change this class to implement `Iterable<GeoitemLayer>` instead of `Collection<GeoitemLayer>`?
@rsudev what do you think?

UniQP:
I followed the idea of @pstorch and changed the implemented interface.

pstorch:
Looks good and clean to me. @rsudev any opinions from your side?

pstorch:
@UniQP thanks

PR #6328, opened 2017-02-16
fix #6319 - Sort trackables in inventory by name or tracking code

### LINKED ISSUE #6319: Sort trackables on log page by code rather than by name
Received by a user:
> [...]would it be possible to list the trackable inventory by tracking code? Right now, it is listed in alphabetical order with the code below it. The only issue, especially when you move many trackables like myself, most of the time the user doesn't know the name of the trackable, only knows the code. So putting the codes alphabetically would be amazing and much faster. 

I never thought about that, but I fully agree. If I put trackables into the cache I always look for its tracking number and sorting by that would indeed make this more easy.
Maybe we could also put it into the first row (bigger font) and the name into the second row (smaller font) or make both lines with same font size.

FreddieSanchez:
This attempts to resolve #6319. I'd like some feedback on this. It feels clunky.

I still need to address the formatting as @Lineflyer suggested.

> Maybe we could also put it into the first row (bigger font) and the name into the second row (smaller font) or make both lines with same font size.


rsudev:
ok to test

pstorch:
It works so far.
I would suggest to remember the sort order in the users preferences (Settings).
I'm thinking about the `Sort By` button: do we find a better place for it? There are already lots of buttons on this page. Maybe put it into the menu? Other opinions?

Lineflyer:
My opinion... No button at all and sort by code plus have the code in first row to make clear its sorted by that.

FreddieSanchez:
Saving the sort method in the setting makes sense. It preserves the setting across all caches (keeps you from having to select the same sorting function for each cache). Moving the button to the menu also makes sense. It will mimic the functionality of the GeoCache sort  in the NearBy activity screen.

@Lineflyer's approach is by far the easiest to implement. Maybe I can just default the sort to the code.

Lineflyer:
If there is no button but just a menu item I am fine as well...my proposal intended to not clutter the view with more buttons.

FreddieSanchez:
OK - thanks for the review. I'll make the suggested changes tonight/tomorrow and push the updates to this branch.

FreddieSanchez:
Updated the PR. Please take another look. 

The only piece missing is saving the selection in the settings.

pstorch:
"Sort by" in the menu instead of the extra button: ok.
Now Code and Name are in one row. It's more compact this way. If it is ok, @Lineflyer has to say.
It looks like the order is wrong (Ascending vs. Descending?).
If sorting by code, then view a trackables details and going back to the log activity, the sort order is resetted (should be kept).

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -227,12 +236,14 @@ public ViewHolder(final View rowView) {
...
 
     private void updateTrackablesList() {
+
         final TrackableLog[] trackablesArray = getSortedTrackables().toArray(new TrackableLog[trackables.size()]);
         final ListView inventoryList = ButterKnife.findById(this, R.id.inventory);
         inventoryList.setAdapter(new TrackableLogAdapter(this, R.layout.logcache_trackable_item, trackablesArray));
         ViewUtils.setListViewHeightBasedOnItems(inventoryList);
 
-        ButterKnife.findById(this, R.id.inventory_box).setVisibility(trackables.isEmpty() ? View.GONE : View.VISIBLE);
+        final LinearLayout inventoryBox = ButterKnife.findById(this, R.id.inventory_box);
```
pstorch: Why introducing a temp variable here?

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -227,12 +236,14 @@ public ViewHolder(final View rowView) {
     }
 
     private void updateTrackablesList() {
+
```
pstorch: Remove blank line at start of the method

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +258,55 @@ public void onClick(final View view) {
         });
     }
 
-    private ArrayList<TrackableLog> getSortedTrackables() {
-        final ArrayList<TrackableLog> sortedTrackables = new ArrayList<>(trackables);
-        Collections.sort(sortedTrackables, new Comparator<TrackableLog>() {
+    private void selectTrackableSort() {
+
```
pstorch: Please remove the extra blank line

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +258,55 @@ public void onClick(final View view) {
...
+        comparators.add(new Comparator<TrackableLog>() {
+
+            @Override
+            public int compare(final TrackableLog lhs, final TrackableLog rhs) {
+                return TextUtils.COLLATOR.compare(lhs.trackCode, rhs.trackCode);
+            }
+        });
+
+
+
```
pstorch: Collapse the blank lines to one

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +258,55 @@ public void onClick(final View view) {
...
+                trackableSort = comparators.get(position);
+                trackableSortPositionSelected = position;
+                updateTrackablesList();
+                dialog.dismiss();
+            }
+        });
+        alert.create().show();
+    }
+
+
```
pstorch: same here, collapse blank lines into one

pstorch:
@FreddieSanchez In the code I only had some minor comments. Please squash your commits into one when you send the final one.

SammysHP:
@pstorch You can use the "start review" function of GitHub for such comments. That will track all comments in a review that is listed at the end of the PR with the status.

pstorch:
@SammysHP OK, will use it next time

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: This was left over from adding the button. I'll remove it.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Agreed. Will remove.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Agreed. Will remove.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Agreed. Will remove.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Agreed. Will remove.

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -1032,8 +1033,26 @@ public static boolean getIncludeFoundStatus() {
     public static void setIncludeFoundStatus(final boolean includeFoundStatus) {
         putBoolean(R.string.pref_includefoundstatus, includeFoundStatus);
     }
+    /**
+     * Get Trackable inventory sort method based on the last Trackable inventory sort method.
+     *
+     * @return
+     *          The last Trackable Action or RETRIEVED_IT
```
pstorch: Copy & Paste error ;)

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +248,48 @@ public void onClick(final View view) {
         });
     }
 
-    private ArrayList<TrackableLog> getSortedTrackables() {
-        final ArrayList<TrackableLog> sortedTrackables = new ArrayList<>(trackables);
-        Collections.sort(sortedTrackables, new Comparator<TrackableLog>() {
+    private void setTrackableSort(int position) {
```
pstorch: add final modifier

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +248,48 @@ public void onClick(final View view) {
...
+        final List<Comparator<TrackableLog>> comparators = new ArrayList<>();
```
pstorch: Create a List of Comparators on each call of this setter? Maybe you find another way to do it.

FreddieSanchez: Agreed. Moved this to static members of the class.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Done.

main/src/cgeo/geocaching/settings/Settings.java
FreddieSanchez: I should stop using copy & paste... All it does is lead to problems.

FreddieSanchez:
I think that last commit resolves all of @pstorch's comments and suggestions. However, it seems to have upset the codacy bot.

rsudev:
I think the remark of the codacy bot is a valid one. @pstorch suggested to move the comparator list from the setter method to somewhere else. From your comment I would deduce that you intended to create a static initializer for this collection (which seems feasible to me, as it is defined compile-time and never changes), but created an instance-initializer instead, which runs on every instantiation of the class.

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -85,6 +85,44 @@
...
+            @Override
+            public int compare(final TrackableLog lhs,  final TrackableLog rhs) {
+                return TextUtils.COLLATOR.compare(lhs.name, rhs.name);
+            }
+        });
+
+        private final int position;
+        private final int label;
+        private final Comparator<TrackableLog> comparator;
+        public static TrackableComparator[] comparators = new TrackableComparator[]
```
pstorch: I think this public mutable array can be omitted. 
Instead add a static method `TrackableComparator.findByPosition(final int position)`

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +285,30 @@ public void onClick(final View view) {
...
     }
 
-    private ArrayList<TrackableLog> getSortedTrackables() {
-        final ArrayList<TrackableLog> sortedTrackables = new ArrayList<>(trackables);
-        Collections.sort(sortedTrackables, new Comparator<TrackableLog>() {
+    private void selectTrackableSort() {
+        final Builder alert = new AlertDialog.Builder(this);
+        alert.setTitle(res.getString(R.string.log_tb_sortby));
+        
+        final String[] tbSortLabels = new String[TrackableComparator.comparators.length];
```
pstorch: `TrackableComparator.values().length`

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +285,30 @@ public void onClick(final View view) {
...
+        for (final TrackableComparator tc: TrackableComparator.comparators) {
```
pstorch: `TrackableComparator.values()`

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -247,15 +285,30 @@ public void onClick(final View view) {
...
+                dialog.dismiss();
             }
         });
+        alert.create().show();
+    }
+
+    private ArrayList<TrackableLog> getSortedTrackables() {
+        final int position = Settings.getTrackableInventorySortMethod();
+        final ArrayList<TrackableLog> sortedTrackables = new ArrayList<>(trackables);
+        Collections.sort(sortedTrackables, TrackableComparator.comparators[position].getComparator());
```
pstorch: `TrackableComparator.findByPosition(position).getComparator()`

main/res/layout/logcache_trackable_item.xml
```diff
@@ -73,4 +75,4 @@
         tools:text="action"
         android:maxLines="1" />
 
-</LinearLayout>
```
UniQP: These whitespace fixes (3 occurrences) are unrelated and should be removed from the commits (you can open a separate PR to tackle them).

UniQP:
For the final version, you should also merge the fixup commits into the commit they fix (it's nice to have them separately for review though).

main/res/layout/logcache_trackable_item.xml
SammysHP: These are added automatically by many editors. IMHO this is ok.

UniQP: It's orthogonal to the idea of the PR/commit, and IMHO it then should be handled separately.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Didn't realize there was a ```.values()``` method. Will update the PR to use this.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Done.

main/src/cgeo/geocaching/log/LogCacheActivity.java
FreddieSanchez: Done.

PR #6331, opened 2017-02-19
Fix #6316: Partially revert 59f584e60ab0a62923f3fbb2ee8ea780976c8d71

### LINKED ISSUE #6316: Cache description can't be shown if one image has an error
- Open Cache https://coord.info/GC59KT4

Error message: `c:geo can't load description.`
On description tab only a loading spinner is shown instead of the description.

Show description with missing image

##### Version of c:geo used:
`master`, 6b4cdbc

Yes

The cache description contains one errornous image tag:
``
It has no URL (no src attribute).

I don't know if we want to fix it, because the error is in the listing. But it would be more user friendly to ignore it (or show a broken image placeholder).

UniQP:
If an  tag has no src attribute getDrawable(null) is called.
Thus, we should not throw an exception in this case (and instead
return a transparent 1x1 image).

### DIFF
--- a/main/src/cgeo/geocaching/network/HtmlImage.java
+++ b/main/src/cgeo/geocaching/network/HtmlImage.java
@@ -168,9 +168,6 @@ public class HtmlImage implements Html.ImageGetter {
     @Nullable
     @Override
     public BitmapDrawable getDrawable(final String url) {
-        if (url == null) {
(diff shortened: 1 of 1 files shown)


UniQP:
The NPath complexity should be tackled in a separate PR.

pstorch:
Yes, let's do it separately. @UniQP thanks.

PR #6332, opened 2017-02-19
Fix method name in log message

UniQP:


UniQP:
Yes, within an anonymous class created in the `fetchDrawableUncached` method.

pstorch:
@UniQP ah, I haven't seen the full file. That's an easy one, thanks :+1:

PR #6338, opened 2017-02-21
Add newline at end of files

UniQP:


SammysHP:
Is this something we can also address in the IDE configuration?

UniQP:
I enabled the corresponding checkstyle warning, but did not know about the IDEs (since I did not use them).

PR #6342, opened 2017-02-21
Fix #6333: Override cache ID

### LINKED ISSUE #6333: c:geo logs wrong cache
I discussed this case on support mail already some moths ago. Now the user contacted me again with some details, and said he had at least 10 cases of this problem on his cache and all confirmed to be c:geo users.

Here is what he writes:
```
Wenn jemand über c:geo den Cache http://coord.info/GC41JNJ loggt,
erfolgt der Log nicht dort, sondern in meinem Cache
https://coord.info/GC15X39 Es ist unabhängig von den Versionen.
```
Short translation:
If a c:geo user composes a log for GC41JNJ the log will be posted to GC15X39 instead.

I contacted the owner of the other cache and asked him whether I can do a test log on his cache.
Will try to reproduce it.

If this is really a bug, what could be the cause? Some mixup with the GUID?

UniQP:
If the cache is already stored on the device, we need to override the
cache ID.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCParser.java
@@ -257,6 +257,7 @@ public final class GCParser {
             cache.setFound(row.contains("/images/icons/16/found.png") || row.contains("uxUserLogDate\" class=\"Success\""));
 
             // infer cache id from geocode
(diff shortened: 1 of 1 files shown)


pstorch:
Let's merge it, so that @Lineflyer can test it.

Lineflyer:
Tested OK! Refresh of the cache and starting the logging acitivity sends a request to the correct URL/ID

PR #6349, opened 2017-02-22
Partially fix #6344: Remove superfluous parameters

### LINKED ISSUE #6344: Lint Warnings may cause crash, display problem and memory leak
Hi,

I am using Lint to check for warnings in cgeo and found some warnings that can cause problems:
- [x] **Multi-catch with these reflection exceptions requires API level 19 (current min is 14) because they get compiled to the common but new super type `ReflectiveOperationException`. As a workaround either create individual catch statements, or catch `Exception`**
I found warnings of this category on [line 127 in FilterActivity.java](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/filter/FilterActivity.java#L127) [line 31 in CustomProgressDiag.java](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/ui/dialog/CustomProgressDialog.java#L31) and [line 150 in SortActionProvider.java](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/sorting/SortActionProvider.java#L150)
This warning may lead to crash as suggested in an [issue report of Android](https://code.google.com/p/android/issues/detail?can=1&q=153406&colspec=ID%20Status%20Priority%20Owner%20Summary%20Stars%20Reporter%20Opened&id=153406) (See comment 2)

- [x] **Format string `auth_register` is not a valid format string so it should not be passed to `String.format`**
This warning can be found in many places in the project like [line 274 in TokenAuthorizationActivity.java](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java#L274).
I have checked [strings.xml](https://github.com/cgeo/cgeo/blob/master/main/res/values/strings.xml#L688). The string is really not a valid string format and the second parameter passed to the method would be ignored.

- [ ] **This Handler class should be static or leaks might occur (anonymous cgeo.geocaching.utils.DisposableHandler)**
This is a possible memory leak issue found [on line 399 in CacheListActivity.java](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/CacheListActivity.java#L399) and [line 165 in GPXImporter](https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/files/GPXImporter.java#L165)
I have found some [related discussions from StackOverflow](http://stackoverflow.com/questions/11407943/this-handler-class-should-be-static-or-leaks-might-occur-incominghandler)

Hope this information can help. :)

UniQP:


Lineflyer:
I do see that this is changing something in the area of authentication. As we are short before merge to `release` (at least I hope) please do only merge to `master` if you consider it save (cant't judge the severity of the change).

pstorch:
I think it is not critical, because the removed parameters were not be used before.
But it won't hurt if a second reviewer double checks.

SammysHP:
These strings were never changed since c:geo is open source. I wonder if this is something that existed even before that.

pstorch:
The strings were there from the beginning (of the open source version), but the extra unused parameter was introduced later in e87c964. When I read it correctly @samueltardieu refactored the OAuth logic from the Twitter class into the abstract base class to be reusable by others (e.g. OC). Then he introduced the extra parameter `getAuthTitle()` to be able to identify the specific OAuth provider in the message produced in the base class. But some messages were not extended to use this parameter.

Source Code archaeology ;)

samueltardieu:
@pstorch This is likely a correct analysis, good job :-)

PR #6392, opened 2017-03-09
Simplify if statements

UniQP:


Lineflyer:
retest this please

samueltardieu:
I'm not sure I like most of the changes either.

Bananeweizen:
I must admit I also dislike the change. Just one example: https://github.com/cgeo/cgeo/pull/6392/files#diff-fbadf9066763d4c81579f2ccddf60a3eR403 (the ProcessPhoenix change). Previously the 2 if conditions were closely together. Now they are separated by 18 lines although they _belong_ together. For me the logic becomes very un-obvious. I also prefer readability over strictly less control structures.

Side note: At work we prioritize [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) over less if statements. In fact the (calculated) complexity of such a method raises and our Sonarqube metrics become worse. But it is totally worth it to have more lines of code there, and to ignore the Sonarqube complexity metric in these cases, since this is rather the opposite of "complex".

UniQP:
I reverted the discussed changes and hope the remaining ones are fine.

PR #6396, opened 2017-03-11
fixes #4957 new approach for local storage

### LINKED ISSUE #4957: New Approach to Move Cache dir to external SD
Hi together,

I would like to try a new approach moving cache dir to external sd like Locus does.
Please vote for or against a new approach.

Choices for user:
- /storage/emulated/0/.cgeo (default)
- /storage/extSdCard/Android/data/cgeo.geocaching/.cgeo (will be removed on uninstall)

The files should be moved by c:geo with progress dialog.
We already have a function to remove garbage from cache directory.
I suggest to remove garbage before moving files.

Open questions, please give your feedback:
- Should I do another approach?
- If the "moving files" dialog is cancel by back button or by sending the app to background, what should be done?
- Should we remove garbage before moving cache directory?
- Any other questions?

pstorch:
Maybe it's time to get some first feedback.

It's implementing the https://github.com/cgeo/cgeo/wiki/Target-disk-usage-structure

:exclamation:  If you want to test it, be careful: it moves your files around. :exclamation:


pstorch:
Looks like codacy has some problems with checking this PR. It doesn't show up here and on codacy.com it checks and checks and checks ...

pstorch:
Codacy is fixed.

pstorch:
Fixed open Issues from https://github.com/cgeo/cgeo/issues/4957#issuecomment-286886863.

Lineflyer:
Will test it again this evening or tomorrow

Lineflyer:
That is strange:
I went the same way I used for the first test:
Installed an older debug.apk retrieved from CI, then wanted to update to [this](http://ci.cgeo.org/job/cgeo%20pull%20request/801/) debug.apk. But installation fails, as the signature is different. Do we use different signatures on CI?

kumy:
If I m not mistaken, release/nigthly/legacy are build on master node using "main" keys. PR are built on sam1/2, keys _may_ differ here.

On March 17, 2017 9:51:20 AM GMT+01:00, Lars <notifications@github.com> wrote:
>That is strange:
>I went the same way I used for the first test:
>Installed an older debug.apk retrieved from CI, then wanted to update
>to [this](http://ci.cgeo.org/job/cgeo%20pull%20request/801/) debug.apk.
>But installation fails, as the signature is different. Do we use
>different signatures on CI?
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/cgeo/cgeo/pull/6396#issuecomment-287299530

-- 

Lineflyer:
retest this please

Lineflyer:
retest this please

pstorch:
retest this please

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -220,12 +223,58 @@ private static void migrateSettings() {
...
+            // move geocache data files
+            final FileFilter geocacheDirectories = new FileFilter() {
+                @Override
+                public boolean accept(final File pathname) {
+                    final String name = pathname.getName();
+                    return pathname.isDirectory() &&
+                            (HtmlImage.SHARED.equals(name) || LocalStorage.GEOCACHE_FILE_PATTERN.matcher(name).find());
+                }
+            };
+            final File[] list = LocalStorage.getLegacyExternalCgeoDirectory().listFiles(geocacheDirectories);
```
Bananeweizen: listFiles can return null. That should not happen, but me may want to be safe.

main/src/cgeo/geocaching/utils/FileUtils.java
```diff
@@ -96,6 +106,232 @@ public static boolean deleteDirectory(@NonNull final File dir) {
...
+     * @param url
+     *            the relative or absolute URL
+     * @return the file extension, including the leading dot, or the empty string if none could be determined
+     */
+    @NonNull
+    public static String getExtension(@NonNull final String url) {
+        final String urlExt;
+        if (url.startsWith("data:")) {
+            // "data:image/png;base64,i53…" -> ".png"
+            urlExt = StringUtils.substringAfter(StringUtils.substringBefore(url, ";"), "/");
```
Bananeweizen: StringUtils.substringBetween(...)?

Bananeweizen:
I've just been wondering about the fault tolerance of that migration. E.g. if one of the operation fails, should we still continue with some of the other? That is, should we try-catch several blocks in the migration code? And how much time will this take, if a user has 5000 caches on disk? I believe we don't show any UI about this, so he might decide to stop the application for any reason. Could that lead to issues?

main/src/cgeo/geocaching/utils/FileUtils.java
samueltardieu: You have to handle `null` if you use `substringBetween` (the current construct returns an empty string) if it doesn't match. I'm not sure the subsequent call to `defaultString` will get better performances, but it may indeed may more readable as

    StringUtils.defaultString(StringUtils.substringBetween("/", ";"));

main/src/cgeo/geocaching/settings/Settings.java
samueltardieu: On the other hand, if `getLegacyExternalCgeoDirectory()` is supposed to be a directory, then we might want to crash if it isn't :)

pstorch:
Regarding fault tolerance: at the moment I'm more in favor of being tolerant and don't stop. The migration and the moving of the Geocache data directory can go wrong easily. The source directory might be missing, some target dirs might already be there. What about no disk space is left? Currently I try to do what is possible and skip over errors, so the user doesn't get stuck. Some cached data might not be copied, but it remains on the other disk and is recoverable. We have to deal with partly migrations anyway, because a filesystem doesn't work with transactions.
This is just my view on this, let's discuss.

The copy of the Geocache data has a UI progress dialog. I don't know how long copying of 5000 caches would take. Unfortunately I don't have a device with a real extSdCard. Maybe I should still order one for testing. Or @Lineflyer might be able to test such a volume.

main/src/cgeo/geocaching/settings/Settings.java
pstorch: As described below, I think I should catch null here to be more fault tolerant.

main/src/cgeo/geocaching/utils/FileUtils.java
pstorch: I haven't thought about this line, because I just moved it here from somewhere else ;)

samueltardieu: Yeah, I recognized this code, but I couldn't tell why I chose `after`+`before` instead of `between`+`default`.

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1869,16 +1869,15 @@ public boolean hasStaticMap() {
     }
 
     /**
-     * Add spoilers stored locally in <tt>/sdcard/GeocachePhotos</tt>. If a cache is named GC123ABC, the
-     * directory will be <tt>/sdcard/GeocachePhotos/C/B/GC123ABC/</tt>.
+     * Add spoilers stored locally in <tt>/sdcard/cgeo/GeocachePhotos</tt>. If a cache is named GC123ABC, the
+     * directory will be <tt>/sdcard/cgeo/GeocachePhotos/C/B/GC123ABC/</tt>.
```
samueltardieu: I think the choice of not using `cgeo` in the path name was because additional data imported from geocaching software could be used by several applications, and did not belong to cgeo.

samueltardieu: I tried to reread all threads on the subject, and it looks like this has no importance after all, so feel free to rename.

pstorch: As far as I could find it, it's from a GSAK macro. I also found the corresponding issues and commits here: https://github.com/cgeo/cgeo/issues/4957#issuecomment-284263660

When I look into the GSAK macro it is actually `Garmin\GeocachePhotos`. So the user of this macro need to copy the subfolder anyhow. I don't know if this directory might be used by some other Apps in parallel. Should we keep it one level up?

samueltardieu: Probably not. As you say, it has to be moved through a file manager or similar anyway, so your proposed change is fine.

PR #6421, opened 2017-03-21
Remove redundant copy operation, use IOUtils instead

samueltardieu:


samueltardieu:
Not to be merged before #6396 gets in as part of this patch is already in.

pstorch:
@samueltardieu thanks for waiting

PR #6424, opened 2017-03-21
Use https for gcvote through hosting provider endpoint

samueltardieu:


Bananeweizen:
The only drawback I can imagine is that our gcvote integration stops working the moment that domain is transferred. otherwise +1

samueltardieu:
Agreed. But fixing it is easy, and moreover losing gcvote temporarily would be almost invisible to the user. I just don't like the user location (through the geocache code) and pseudo to go in the clear on the wire. You never know who's listening, let's do our part correctly :-)

PR #6439, opened 2017-03-22
fix #6426 add percentage progress dialog while moving geocache data

### LINKED ISSUE #6426: Show percentage of completion while moving GeocacheData
As the operation of moving GeocacheData from internal to external might take long, we should show a clear progress.

Test still outstanding how long it actually take with 1k or 5k cache dirs.

pstorch:
Added a progress dialog with percentage while moving the geocache data.

There are some additional ideas from @Lineflyer in https://github.com/cgeo/cgeo/issues/6426#issuecomment-288357770 (warning dialog and disk full handling), they are not included here.

### DIFF
--- a/main/src/cgeo/geocaching/activity/Progress.java
+++ b/main/src/cgeo/geocaching/activity/Progress.java
@@ -7,6 +7,7 @@ import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.os.Message;
+import android.view.Window;
 import android.view.WindowManager;
 
 /**
@@ -16,7 +17,6 @@ public class Progress {
 
     private ProgressDialog dialog;
     private int progress = 0;
-    private int progressDivider = 1;
     private final boolean hideAbsolute;
 
     public Progress(final boolean hideAbsolute) {
@@ -67,8 +67,11 @@ public class Progress {
         }
         dialog.setProgress(0);
         dialog.setCanceledOnTouchOutside(false);
-        dialog.getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
-        this.progress = 0;
+        final Window window = dialog.getWindow();
+        if (window != null) {
+            window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+        }
+        progress = 0;
     }
 
     public synchronized void setMessage(final String message) {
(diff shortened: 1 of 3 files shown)


Lineflyer:
retest this please

pstorch:
Should I put the Warning dialog and the disk full handling into this change? I have the feeling that we have to elaborate on these a bit. Maybe it's better to open a new issue for them.

Lineflyer:
@pstorch Lets make this separate.

main/src/cgeo/geocaching/activity/Progress.java
```diff
@@ -108,4 +108,11 @@ public synchronized int getProgress() {
     public synchronized void setProgressDivider(final int progressDivider) {
         this.progressDivider = progressDivider;
     }
+
+    public void incrementProgressBy(final int increment) {
+        if (isShowing()) {
+            dialog.incrementProgressBy(increment);
```
samueltardieu: Why not call `dialog.setProgress(progress)` after having incremented `progress`? This would ensure that `progress` and `Dialog.progress` stay in sync.

main/src/cgeo/geocaching/activity/Progress.java
```diff
@@ -108,4 +108,11 @@ public synchronized int getProgress() {
...
+        }
+        this.progress += increment;
```
samueltardieu: All those `this.` should be removed from the file :-)

main/src/cgeo/geocaching/storage/LocalStorage.java
```diff
@@ -307,19 +309,21 @@ public static File getLogPictureDirectory() {
     }
 
     public static void changeExternalPrivateCgeoDir(final SettingsActivity fromActivity, final String newExtDir) {
-        if (StringUtils.equals(getExternalPrivateCgeoDirectory().getAbsolutePath(), newExtDir)) {
-            Settings.setExternalPrivateCgeoDirectory(newExtDir);
-            return;
-        }
-        final ProgressDialog dialog = ProgressDialog.show(fromActivity, fromActivity.getString(R.string.init_datadirmove_datadirmove), fromActivity.getString(R.string.init_datadirmove_running), true, false);
+        final Progress progress = new Progress(true);
+        final Resources res = fromActivity.getResources();
```
samueltardieu: Is `res` needed here? Can't `fromActivity.getString(…)` get used instead?

main/src/cgeo/geocaching/activity/Progress.java
pstorch: I'll remove them.

main/src/cgeo/geocaching/storage/LocalStorage.java
pstorch: Yes, can be replaced.

main/src/cgeo/geocaching/activity/Progress.java
pstorch: On a second look I think I understand what this Progress wrapper is doing. 
Can I remove this `progressDevider` logic? It is not used (always 1, setter never called) and it is buggy, see `getProgress()`.

samueltardieu: Sure!

pstorch:
Updated PR according to the comments.

PR #6440, opened 2017-03-23
Fix #6438 - Improve strings for storage settings

### LINKED ISSUE #6438: Improve setting strings for DBonSD
We should change the description to reflect that the DB is (in most cases) only moved between total internal and emulated public dir. 
I will take the action for this.

Lineflyer:
I could not find any string for the caption of the Geocache data section (like e.g. `<string name="init_datadir_title">`. I would like this to be renamed to "Geocache Data Location".

It seems hardcoded right now. Will open a dedicated issue after checking.


Lineflyer:
Edit: Found the missing string.

PR #6441, opened 2017-03-23
fix #6437 add more prefixes to geocache data pattern

### LINKED ISSUE #6437: Not all GeocacheData elements moved
Today I installed 2017.03.23-NB containing the NewDir onto my regular device.
I checked the condition of `/.cgeo/` after migration: It was still existing and contained a directory `TC9KV` with pictures, maps, etc.

This is the additional data of a terracaching cache I have stored on my device (imported by GPX).
It seems the TC-prefix is missing for the move operation (and maybe also on other places in c:geo where this code was derived from).

pstorch:
Added more patterns from other platforms.
Removed pattern matching for orphaned files cleanup (we now have the GeocacheData directory with only cached Data).


Bananeweizen:
:+1:

PR #6442, opened 2017-03-23
fix #6425 improved Settings design for geocache data option

### LINKED ISSUE #6425: Improve setting design
The setting section in Settings - System for "GeocacheData" is not likely to be identified as "clickable".
I suggest to add a field below the description text showing the currently used dir and clicking on that offers the dir selection dialog.

pstorch:
Similar to Settings - Navigation - "Standardnavigation"


Bananeweizen:
:+1:

PR #6444, opened 2017-03-24
Use GeographicLib to compute distance, azimuth and projections

### LINKED ISSUE #6123: Inaccurate projection
- create point N60° 00.000' E024° 00.000'
- project from it 45° and 1000m

- c:geo creates new point at N60° 00.382' E024° 00.763'

The following three implementations of WGS84 ellipsoid projection using different code bases seem to agree with each other and disagree with the c:geo implementation. There is an difference of 0.001'/0.003' which is already a couple of meters even at this short distance.

http://williams.best.vwh.net/gccalc.htm
Great Circle Calculator by Ed Williams (javascript)
Compute lat/lon given radial and distance from a known point, WGS84 earth model
N60:00.3808 E024:00.7605

http://gc.pitko.net/cal/gcal.pl
Spitko's GeoCalc (perl Geo::Ellipsoid)
N60 00.381 E24 00.760

http://xiit.dy.fi/gc/
GeoCalcing² by zuge (GeographicLib http://geographiclib.sourceforge.net/)
Projections
N 60 00.381 E 024 00.760

##### Version of c:geo used:
2016.11.05

Yes, also on earlier releases

##### System information:
--- System information ---
Device: Jolla C (Jolla C, Jolla)
Android version: 4.4.4
Android build: KTU84Q dev-keys
Cgeo version: 2016.11.05
Google Play services: unavailable
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: true
Map strategy: auto
System language: en_GB
Log date format: 13/11/2016
Debug mode active: no
Geocaching sites enabled:
Geocaching.com date format: yyyy-MM-dd
Installed cgeo plugins:  none

samueltardieu:
GeographicLib is released under the MIT/X11 License.

Fix #6123.

### DIFF
--- a/main/src/cgeo/geocaching/speech/TextFactory.java
+++ b/main/src/cgeo/geocaching/speech/TextFactory.java
@@ -1,14 +1,14 @@
 package cgeo.geocaching.speech;
 
-import android.support.annotation.PluralsRes;
-import android.support.annotation.StringRes;
-
 import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.IConversion;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.AngleUtils;
+
+import android.support.annotation.PluralsRes;
+import android.support.annotation.StringRes;
 
 import java.util.Locale;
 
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/speech/TextFactory.java
```diff
@@ -100,16 +100,21 @@ private static String getQuantityString(@PluralsRes final int resourceId, final
...
-        final int degrees = (int) AngleUtils.normalize(bearing - direction);
+        int hours;
 
-        int hours = (degrees + 15) / 30;
-        if (hours == 0) {
+        if (position.distanceTo(target) < 1e-4) {
+            // By convention, distance smaller than ten centimers will be represented as 12 o'clock.
             hours = 12;
+        } else {
+            final int bearing = (int) position.bearingTo(target);
```
UniQP: Shouldn't that be a float?

main/src/cgeo/geocaching/speech/TextFactory.java
```diff
@@ -100,16 +100,21 @@ private static String getQuantityString(@PluralsRes final int resourceId, final
...
+            final int degrees = (int) AngleUtils.normalize(bearing - direction);
+
+            hours = (degrees + 15) / 30;
```
UniQP: I would appreciate the use of Math.round()

UniQP:
@samueltardieu Did you check, whether we can (or should) replace some other calls to `Location.distanceBetween(...)` by calls to GeographicLib?

main/src/cgeo/geocaching/speech/TextFactory.java
samueltardieu: I haven't touch this part except put in into braces. Feel free to open another issue if you think some things should be cleared up.

samueltardieu:
`Location.distanceBetween()` is used to determine map ranges on screen and such. This has little importance, and may prevent allocating a new object.

I could add a `LocationUtils` though, and add a `distanceBetween` there. Let me think about it.

PR #6454, opened 2017-03-25
fix #6452 center move geocache data progress dialog

### LINKED ISSUE #6452: Progress indication should be centered on screen
- Start moving geocache data to/from SD

Progress window appears on top of the screen

Just cosmetic:
The progress window should be centered on the screen

##### Version of c:geo used:
2017.03.25-NB

Yes

##### System information:
Android 5.1.1
Samsung Galaxy Note 10.1 Edition 2014 (SD removeable without power down)

Screenshot

screenshot_2017-03-25-12-21-39

pstorch:


samueltardieu:
Is it a side effect of not using `ActivityDialog.getDialogTheme()` (through `CustomProgressDialog`)?

pstorch:
Don't know actually. I found other progress dialogs which are aligned on top (e.g. GPX import) and others in the middle (e.g. refresh in ListActivity). So I simply chose the option of the latter.
Maybe someone else has more insights here.

samueltardieu:
I was just curious. It looks like the `CustomProgressDialog` (which gets used when you pass `true` to the `Progress` constructor) is aligned at the top.

PR #6457, opened 2017-03-26
Fix #6344: Fix handler leaks

### LINKED ISSUE #6344: Lint Warnings may cause crash, display problem and memory leak

UniQP:


main/src/cgeo/geocaching/MainActivity.java
```diff
@@ -101,56 +102,62 @@
     private ConnectivityChangeReceiver connectivityChangeReceiver;
 
     private final UpdateLocation locationUpdater = new UpdateLocation();
+    private final Handler updateUserInfoHandler = new UpdateUserInfoHandler(this);
+    private final Handler firstLoginHandler = new FirstLoginHandler(this);
 
-    private final Handler updateUserInfoHandler = new Handler() {
+    private static final class UpdateUserInfoHandler extends Handler {
+        private final WeakReference<MainActivity> activityRef;
```
samueltardieu: It looks like it would be best to use `WeakReferenceHandler` here.

main/src/cgeo/geocaching/MainActivity.java
```diff
@@ -195,21 +202,29 @@ public void accept(final Status gpsStatus) {
         }
     };
 
-    private final Handler firstLoginHandler = new Handler() {
+    private static final class FirstLoginHandler extends Handler {
+        private final WeakReference<MainActivity> activityRef;
```
samueltardieu: `WeakReferenceHandler`

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
```diff
@@ -63,47 +64,67 @@
     private ProgressDialog requestTokenDialog = null;
     private ProgressDialog changeTokensDialog = null;
 
-    private final Handler requestTokenHandler = new Handler() {
+    private final Handler requestTokenHandler = new RequestTokenHandler(this);
+    private final Handler changeTokensHandler = new ChangeTokensHandler(this);
+
+    private static final class RequestTokenHandler extends Handler {
+        private final WeakReference<OAuthAuthorizationActivity> activityRef;
```
samueltardieu: `WeakReferenceHandler`

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
```diff
@@ -63,47 +64,67 @@
...
+                    activity.showToast(activity.getErrAuthInitialize());
+                    startButton.setText(activity.getAuthStart());
+                }
             }
         }
+    }
 
-    };
+    private static final class ChangeTokensHandler extends Handler {
+        private final WeakReference<OAuthAuthorizationActivity> activityRef;
```
samueltardieu: `WeakReferenceHandler`

main/src/cgeo/geocaching/MainActivity.java
UniQP: That's not possible, because `MainActivity` does not extend `Activity`, which is expected by `WeakReferenceHandler`.

main/src/cgeo/geocaching/MainActivity.java
UniQP: That's not possible, because `MainActivity` does not extend `Activity`, which is expected by `WeakReferenceHandler`.

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
UniQP: That's not possible, because `OAuthAuthorizationActivity` does not extend `Activity`, which is expected by `WeakReferenceHandler`.

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
UniQP: That's not possible, because `OAuthAuthorizationActivity` does not extend `Activity`, which is expected by `WeakReferenceHandler`.

samueltardieu: Could you update `WeakReferenceHandler` so that it accepts an arbitrary type? There is no good reason to restrict it to an activity AFAICS.

You can rename `getActivity()` to `getReference()` to reflect the type change.

UniQP: Done.

pstorch:
@samueltardieu can you have a look again?
@UniQP a small merge conflict happend meanwhile.

samueltardieu:
I'm testing it on the CI after a merge. Thanks for the heads up @pstorch.

samueltardieu:
Thanks @UniQP!

PR #6475, opened 2017-04-06
fix #6471 allow copying waypoint description text

### LINKED ISSUE #6471: Allow copying waypoint description text
With recent changes we disallowed editing the waypoint text of server waypoints.
At least we should allow copying the text to clipboard (via long click on it), so that the user can reuse e.g. contained formulas, etc.

pstorch:
Allow selecting and copying waypoint name and description text for server defined (non-editable) waypoints.

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -164,8 +164,8 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
                 if (own) {
                     initializeWaypointTypeSelector();
                 } else {
-                    waypointName.setEnabled(false);
-                    note.setEnabled(false);
+                    nonEditable(waypointName);
+                    nonEditable(note);
                     if (!waypoint.isOriginalCoordsEmpty()) {
                         projection.setVisibility(View.GONE);
                     }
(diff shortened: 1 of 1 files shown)


Bananeweizen:
The text becomes editable with these changes, which is probably not what you intended, right?

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -164,8 +165,11 @@ protected void onPostExecute(final Geocache cache) {
                 if (own) {
                     initializeWaypointTypeSelector();
                 } else {
-                    waypointName.setEnabled(false);
-                    note.setEnabled(false);
+                    waypointName.setInputType(EditorInfo.TYPE_NULL);
```
Bananeweizen: change to InputType.TYPE_NULL please.

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -164,8 +165,11 @@ protected void onPostExecute(final Geocache cache) {
...
+                    waypointName.setTextIsSelectable(true);
+
```
Bananeweizen: add waypointName.setKeyListener(null) to make it non-editable again.

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -164,8 +165,11 @@ protected void onPostExecute(final Geocache cache) {
...
+                    note.setInputType(EditorInfo.TYPE_NULL);
+                    note.setTextIsSelectable(true);
```
Bananeweizen: Same for these 2 lines. We now have 2 times 3 lines of identical code, so extracting a method seems reasonable.

PR #6485, opened 2017-04-11
fix #6472 proper truncate directory names

### LINKED ISSUE #6472: Dialog for dir selection truncated
Thats how it looks on my Samsung S4.
You are not able to see the free space on the target dirs:
screenshot_2017-04-03-21-00-12

pstorch:


pstorch:
I don't think the codacy findings are valid here.
In the UnitTest I exactly wanted to test the sdcard names and the other finding is at a place I haven't changed.

main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -324,5 +324,31 @@ public static String formatBytes(final long bytes) {
...
+                if (!directories.get(i).toString().endsWith(newEnding)) {
+                    break LOOP;
+                }
+            }
+            commonEnding = newEnding;
+            offset = commonEnding.length() + 1;
+        }
+        final List<CharSequence> truncatedDirs = new ArrayList<>(directories.size());
+        for (int i = 0; i < directories.size(); i++) {
+            final CharSequence title = directories.get(i);
```
UniQP: You can use an enhanced for-loop here.

main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -324,5 +324,31 @@ public static String formatBytes(final long bytes) {
         return String.format(Locale.getDefault(), "%.1f %sB", bytes / Math.pow(1024, exp), pre);
     }
 
+    public static List<CharSequence> truncateCommonSubdir(final List<CharSequence> directories) {
```
UniQP: Do we have any guarantees on the length of `directories`? With length 0 we get an OutOfBoundsException and with length 1 the while loop will never terminate.

pstorch: This dialog is only shown if the list has a size > 1.

main/src/cgeo/geocaching/utils/Formatter.java
pstorch: Thanks for pointing it out. This was not possible before my last refactoring.

main/src/cgeo/geocaching/utils/Formatter.java
pstorch: I'll add a check to be on the safe side

main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -324,5 +324,33 @@ public static String formatBytes(final long bytes) {
...
+                    break LOOP;
+                }
+            }
+            commonEnding = newEnding;
+            offset = commonEnding.length() + 1;
+        }
+        final List<CharSequence> truncatedDirs = new ArrayList<>(directories.size());
+        for (final CharSequence title : directories) {
+            CharSequence truncated = title;
+            if (commonEnding != null) {
```
UniQP: The condition is loop invariant, so we can test it before the loop and simply return `directories` if `commonEnding` is `null`.

main/src/cgeo/geocaching/utils/Formatter.java
```diff
@@ -324,5 +324,33 @@ public static String formatBytes(final long bytes) {
...
 
+    public static List<CharSequence> truncateCommonSubdir(@NonNull final List<CharSequence> directories) {
+        if (directories.size() < 2) {
+            return directories;
+        }
+        final String firstDir = directories.get(0).toString();
+        String commonEnding = null;
+        int offset = 0;
+        LOOP:
+        while (true) {
```
UniQP: I'm still not happy with this loop. I will try to rewrite it and report back.

pstorch: I'm not happy either. Ideas are welcome. 
I've written a Unit test for it, so refactorings can be verified easily. 
Thanks for your effort.

pstorch:
Thanks @UniQP for simplifying the loop. I've just added another testcase and inlined a temp variable.

PR #6492, opened 2017-04-15
Remove superfluous final modifier from static methods

UniQP:


Bananeweizen:
I'm not sure I can avoid re-adding these (with the Eclipse formatter settings).

PR #6493, opened 2017-04-15
Fix some type mismatches in format strings

UniQP:


Lineflyer:
In order not to break our crowdin process it might be meaningful only to correct the english strings (and the corresponding code) and retrieve the remaining languages via Crowdin. Or am I wrong?

samueltardieu:
@Lineflyer You're right

main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -455,7 +455,7 @@ private void initializeFavoriteCheck() {
         if (ConnectorFactory.getConnector(cache).supportsAddToFavorite(cache, typeSelected)) {
             if (premFavPoints > 0) {
                 favCheck.setVisibility(View.VISIBLE);
-                favCheck.setText(getString(R.string.fav_points_remaining, premFavPoints));
+                favCheck.setText(getString(R.string.fav_points_remaining, String.valueOf(premFavPoints)));
```
samueltardieu: Why not either:

- fix the string to use `%d` instead of `%s`
- do nothing and let the implicit `.toString()` do the job

rather than calling `String.valueOf` explicitly?

main/res/values-ca/strings.xml
```diff
@@ -92,7 +92,7 @@
   <string name="log_posting_twitter">Publicant Tweet…</string>
   <string name="log_posting_gcvote">Publicant GCvote…</string>
   <string name="log_posting_image">Publicant imatge…</string>
-  <string name="log_posting_generic_trackable">Publicant %1$s%2$s/%3$s</string>
+  <string name="log_posting_generic_trackable">Publicant %1$s%2$d/%3$d</string>
```
samueltardieu: Please do not modify strings outside the source (`res/values/strings.xml`). Other ones come from Crowdin.

main/src/cgeo/geocaching/log/LogCacheActivity.java
UniQP: You're right, I will change it to use `%d`.

UniQP:
Regarding Crowdin: For https://github.com/cgeo/cgeo/pull/5891 @rsudev uploaded the translations to Crowdin. Since the PR changes only the format specifiers, we should use a similar approach here.

samueltardieu:
This is equivalent to force pushing changes. If @rsudev wants to take the risk, no problem with me.

UniQP:
Ok, I fixed the other issue in the meantime.

rsudev:
In the case of #5891 it was only one language file to check and upload. To verify and address all 20+ languages in a similar fashion seems much riskier and more tedious to me than to upload the source strings and than go through the changes on crowdin and adapt the translations there.

UniQP:
Ok, then I will update PR accordingly.

UniQP:
I removed the translations from the PR.

UniQP:
@rsudev Can you take care of the translations?

Issue #6427, opened 2017-03-21
Handling of full SD card with GeocacheData on SD active


Lineflyer:
##### Detailed steps causing the problem:
- Have a full SD-card 
- Open nearby list
- Try to store search results

##### Actual behavior after performing these steps:
c:geo is stuck in download progress 0/20

##### Expected behavior after performing these steps:
Some (to be discussed) handling of that problem

##### Version of c:geo used:
http://ci.cgeo.org/job/cgeo%20pull%20request/816/artifact/main/build/outputs/apk/cgeo-debug.apk

##### Is the problem reproducible for you?
Yes

##### System information:
tbc

##### Other comments and remarks:
Just a quick test performed. Better test later.

Log:
```
23:28:08.896 Debug cgeo 10326  [refresh--2] Saving GC5NB80 Adneter Marmor ([]) to DB
23:28:09.006 Error cgeo 10326  [network--6] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--6] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--5] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--4] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--2] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--9] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--7] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--2] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
23:28:09.006 Error cgeo 10326  [network--8] Could not make directories /storage/extSdCard/Android/data/cgeo.geocaching/files/GeocacheData/GC5NB80
[...]

PR #6508, opened 2017-04-20
for #6427 check diskspace on startup

pstorch:
It's not fully fixing the issue, but presenting a warning on startup if the geocache data directory has less then 5 MB of free disk space.
It's only checking this directory because this is the biggest one (with all the images etc.). There could also be a problem with the internal one, but normally there is a warning from the device if it is running low on internal memory.

@Lineflyer please check if this is sufficient for now. Also the wording of the message and the 5 MB threshold.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1581,5 +1581,7 @@
 
     <!-- calendar adder -->
     <string name="event_fail">Failed to add event cache to calendar</string>
+    <string name="init_low_disk_space">Running low on disk space</string>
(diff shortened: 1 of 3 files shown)


pstorch:
Retest this please

Bananeweizen:
Looks good to me. However, I would increase the threshold. 5 MB can be consumed by just taking 1 more picture with the camera app. I would rather go with like 100 MB. Given that SD cards come with multiple GB of space, having 100 MB remaining is still equivalent to 98+ percent full.

pstorch:
Sounds reasonable, 5 MB was a bit low.

PR #6514, opened 2017-04-23
fix #6513 grey out non-selectable target dir

### LINKED ISSUE #6513: Target dir not selectable when free space not suffcient
- Have additional geocache data stored on internal storage (15 MB used)
- Fill up external SD with dummy data until less than 15MB are available (in my case 6MB available)
- Select Menu - Settings - System and open storage selection

- Both dir options are shown (internal and external)
- Free space and used space is shown correctly
- Dialog looks normal but I cannot select the external SD (6 MB free)

I guess this is intentional as the free space on SD (6MB) is not sufficient (15MB used).
This however is not intuitive in this dialog. 
The target which has not sufficient space should be e.g. "greyed out"

##### Version of c:geo used:
2017.04.23-RC

Yes

##### System information:
Samsung Galaxy S4

pstorch:


Bananeweizen:
Not tested locally, but looks okay to me.

PR #6520, opened 2017-04-25
fix #6512 Cyclemap via Thunderforest with API key

### LINKED ISSUE #6512: OSM:Cyclemap requires API key
- Use live map with OSM:Bikemap (or whatever it is called in english)

Tiles have a text overlay API key required

No warning like this

##### Version of c:geo used:
2017.03.29

Yes

pstorch:
For the old map I also had to update https://github.com/cgeo/mapsforge, I'll send a PR there, too.


main/src/cgeo/geocaching/maps/mapsforge/v6/layers/ThunderforestMap.java
```diff
@@ -0,0 +1,39 @@
...
+    public ThunderforestMap(final String[] hostNames, final int port) {
+        super(hostNames, port);
+    }
+
+    public int getParallelRequestsLimit() {
+        return 8;
+    }
+
+    public URL getTileUrl(final Tile tile) throws MalformedURLException {
+        return new URL("https", this.getHostName(), this.port, "/cycle/" + tile.zoomLevel + '/' + tile.tileX + '/' + tile.tileY + ".png?apikey=d6866308a6e24783822338c9a49ceb6c");
```
Bananeweizen: I suggest to remove the key similarly to all other keys we use:
* create a new resource string in keys.xml (and in its template version), including a comment where to register for a new key
* use the resource string in the code
* have @mucek4 put the actual value into the file on the CI server

kumy: if @mucek4 is unavailable I could do it. :)

pstorch: I can externalize the key in the mapsforge V6 version, but not in the V3. This is handled in the other repo and I doubt I can access the keys from there.

pstorch:
@Bananeweizen, @kumy, @mucek4 the key is externalized and CI should replace `@thunderforest.api.key@` in the `templates/keys.xml` file.

kumy:
Key added to `private.properties` on server.

kumy:
@samueltardieu it may also be good to add the key on your executor?

samueltardieu:
@kumy Is that needed? There doesn't seem to be any test using this map, and as long as we don't share other keys and signatures (which we should IMO, as artifacts would be usable by everyone, including on pull requests), this seems useless to me.

kumy:
I was thinking about unit tests, if there is none, then I agree, it's not necessary.

PR #6527, opened 2017-04-26
fix 6521 remove mapsforge_patches

pstorch:


pstorch:
Looks like it's not used anymore, I'll merge it.

PR #6536, opened 2017-05-01
fix #6533 access global type filter in filter bar of cache list

### LINKED ISSUE #6533: Click on active filter opens wrong menu
- Activate a cache type filter in the main menu function (lower right icon)
- Open list of saved caches
- Click on the active filter notification icon on too

The filter menu for list filtering is opened, where the user cannot find a way to reset the active filter

The main menu filter option should open

##### Version of c:geo used:
2017.03.29

Yes

The underlying problem here is once again the filter structure which needs to be reworked. Other issues exist for that.
Still the user should be able to drop or change the filter when clicking on it.

pstorch:
Without starting the big rework of the filter(s), here is a proposed solution for this issue:
If a global type filter is active a click on the filter bar is now showing global type filter dialog. A push on the filter icon in the action bar is always showing the Filter Activity. With this solution a user is still able to combine both filters. It's not ideal because the filterbar behaves differently depending on context, but the best I could think of at the moment.

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
+++ b/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
@@ -1,7 +1,11 @@
 package cgeo.geocaching.ui.dialog;
 
 import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.R;
+import cgeo.geocaching.enumerations.CacheType;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.ImageUtils;
+import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.functions.Action1;
 
 import android.app.Activity;
@@ -25,6 +29,10 @@ import android.widget.EditText;
 import android.widget.ListAdapter;
 import android.widget.TextView;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 
 import butterknife.ButterKnife;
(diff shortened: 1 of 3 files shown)


pstorch:
@UniQP thanks for spotting. I'm wondering why codacy is not checking it.

Bananeweizen:
because for a long time checkstyle was not able to correctly detect imports necessary for javadoc class references, and therefore we probably disabled the rule. Android Studio/IDEA are not removing unused imports with the settings checked into our repository. I've asked multiple times for someone with knowledge to fix this (as I only know how to check in the right settings for eclipse).

Bananeweizen:
change looks good to me

pstorch:
I've removed the unused imports. Since some time I have enabled the "optimize imports on the fly" setting in AS, but it doesn't work reliably as it seams. 
I'll have a look into how to share the project settings for Idea based IDEs.

pstorch:
retest this please

samueltardieu:
> I've asked multiple times for someone with knowledge to fix this (as I only know how to check in the right settings for eclipse)

I think the answer has always been "I don't know how to do this in IDEA", from multiple people.

main/src/cgeo/geocaching/ui/dialog/Dialogs.java
```diff
@@ -500,4 +508,53 @@ public static void dismiss(@Nullable final ProgressDialog dialog) {
...
+        Collections.sort(sorted, new Comparator<CacheType>() {
+            @Override
+            public int compare(final CacheType left, final CacheType right) {
+                return TextUtils.COLLATOR.compare(left.getL10n(), right.getL10n());
+            }
+        });
+
+        cacheTypes.addAll(sorted);
+
+        int checkedItem = cacheTypes.indexOf(Settings.getCacheType());
```
samueltardieu:
``` java
final int checkedItem = Math.max(0, cacheTypes.indexOf(Settings.getCacheType());
```

should convey the intention more clearly (IMO).

UniQP:
Looks like Checkstyle also discovered the unused imports: http://ci.cgeo.org/job/cgeo%20pull%20request/902/checkstyleResult/new/
Do we need to upgrade from warning to error to reject the PR in such a case?

main/src/cgeo/geocaching/ui/dialog/Dialogs.java
pstorch: I haven't looked much at the code I moved around, but I agree. I'll change it.

pstorch:
@UniQP @Bananeweizen we don't want to have unused imports, so from my point of view it would be good to raise the checkstyle severity for them to error.

pstorch:
Regarding .idea settings, I started reading a bit: https://tips.seebrock3r.me/share-the-settings-with-the-whole-team-android-studio-protip-6-3cce16eb2ea4
Let's see.

Bananeweizen:
> Some plugins tend to write their own settings, and that should be fine, but the problem is that a bug (marked as resolved, but it’s probably a regression) in the IDE causes “unknown” code style nodes to be removed instead of ignored. This means you have to make sure the team is aligned on which plugins to have enabled in the IDE while working on the project.

That is a real problem. We once had some of the idea settings file in version control, and those always changed immediately on just starting the IDE on my system, because they contain installation specific parts and even OS specific parts. So the unix like paths of samuel were immediately changed on my Windows system etc.

PR #6543, opened 2017-05-06
#6533 cleanup global type filter in filter bar of cache list

pstorch:
- Deleting the global cachetype filter didn't properly cleanup the filter bar.
- When the list filter was used then the global cachetype filter was always prepended in the filterbar text, resulting in something illogical like: "All Cache Types, Traditional Cache". So I've removed the global cachetype filter when it is `CacheType.ALL`.

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -92,6 +92,7 @@ import android.support.v4.app.LoaderManager;
 import android.support.v4.content.Loader;
 import android.support.v4.view.MenuItemCompat;
 import android.support.v7.app.ActionBar;
+import android.text.TextUtils;
 import android.view.ContextMenu;
 import android.view.ContextMenu.ContextMenuInfo;
 import android.view.KeyEvent;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -1694,15 +1696,17 @@ public static void startActivityFinder(final Activity context, final String user
     }
 
     private void prepareFilterBar() {
-        if (Settings.getCacheType() != CacheType.ALL || adapter.isFiltered()) {
-            final StringBuilder output = new StringBuilder(Settings.getCacheType().getL10n());
-
+        final boolean cacheTypeFilterActive = Settings.getCacheType() != CacheType.ALL;
+        if (cacheTypeFilterActive || adapter.isFiltered()) {
+            final List<String> filters = new ArrayList<>();
+            if (cacheTypeFilterActive) {
```
UniQP: I don't like the duplication of the conditions.

Can you extract the whole if block into an own function that takes the filter name as parameter? You can then call the new function with `Settings.getCacheType().getL10n()` and `adapter.getFilterName()`, respectively.

pstorch: Yes, thought about the ifs, too. Had various versions of this function. I've extracted the creation of the `filterNames` List, so the original method is just responsible for setting the UI state, wdyt?

Bananeweizen:
nice catch

PR #6548, opened 2017-05-09
Play with gc.com API

samueltardieu:
gc.com uses oauth for some requests from the web browser. It means that we can use it as well for those requests, as we act like a browser.

As an example, I added location retrieval for search results using API requests instead of loc request (which is probably a bad idea for premium members as a single loc request can retrieve a whole bunch of locations).

Do not merge, it is there for illustration only, and the API class may be reused.


samueltardieu:
Closing because of PR #6686.

PR #6557, opened 2017-05-12
fix #6515 remove android wear helper app

### LINKED ISSUE #6515: Remove c:geo wear from recommended apps
The mentioned Android wear app has been depublished as the author stopped maintaining it.
Very sad IMHO.

However we should remove it from our list of recommended apps for now.
I contacted the author, if we could takeover the code.

If someone does the needed changes: Please apply them to `release` branch.

pstorch:


pstorch:
@FUSSEL4 thanks, I've removed one occurence of these string from `preference_keys.xml`, but we don't remove messages from `strings.xml`.

PR #6567, opened 2017-05-17
 Fix #6546 (2): Proposal for a new structure of the cache list menu

FUSSEL4:
### Base of the PR
The proposal provided here suggests a new structure of the cache list menu like described in [Issue-6546](https://github.com/cgeo/cgeo/issues/6546). 

This PR is a updated version of [#6551](https://github.com/cgeo/cgeo/pull/6551) which couldn't be rebased/merged due to conflicts.

### Scope of the PR
The scope of the PR is to provide a much shorter main cache list menu by introducing sub menus. 

### Considerations
* A restructuring of the cache list menu seems necessary because new functionality often leads to an additionally menu item (see [#6528](https://github.com/cgeo/cgeo/issues/6528) or  [#6473](https://github.com/cgeo/cgeo/issues/6473)) .
* Two main menu items are added: "Manage caches", "Manage lists".

### Screenshots
| Before | After |
| ----- | ----- |
| device-2017-05-10-230544 | screenshot_20170515-081710 |

### Testing
* Functionality was tested manually (Samsung Galaxy S3 with Kitkat, S6 with Nougat).

PR #6568, opened 2017-05-17
issue-6546: changes the cache list menu structure

pstorch:
- new menu structure of the cache list menu
- hides empty caches sub menu if there are no caches in the list


Lineflyer:
Looks good now. Thanks @pstorch for helping out here.
AFAIR we discussed whether wwe should set the menus disabled now instead of invisible, which I would also prefer in order to minimize the "magically" vanishing menus. Its more clear for the user to disable instead of hide.

But we can anyway do this in a follow up PR.

FUSSEL4:
for detailed description see [#6567](https://github.com/cgeo/cgeo/pull/6567).

If this PR is merged I'll close the original one.

FUSSEL4:
@pstorch Thank you for helping out.

pstorch:
@FUSSEL4 thanks for your contribution

PR #6574, opened 2017-05-21
Fix #6573: Parse GPX original coordinates as double values

### LINKED ISSUE #6573: Cannot Parse Latitude - GPX load failure
- Load caches from GPX file in Dropbox

- Get a message saying "Import failed - unexpected error   Cannot Parse Latitiude"
- this consistently happens after 1678 caches (of the 4500 in the gpx)

- I have used this method for years and never had a problem.

##### Version of c:geo used:
- 2017.05.09

Yes

##### System information:
```
```

- I have rebuilt the GSAK database but it hasn't helped

UniQP:
Parsing the coordinates is not necessary because we already know their
format in advance. Futhermore, directly parsing the coordinates as
double values also support integer values, which would not be recognized
as valid latitude/longitude by the coordinate parser.

### DIFF
--- a/main/res/values/changelog_release.xml
+++ b/main/res/values/changelog_release.xml
@@ -5,6 +5,7 @@
     <b>Next bugfix release:</b>\n
     · Fix: Avoid crash during file operation\n
     · Fix: Avoid crash when checking for Locus installation\n
+    · Fix: Avoid failure when parsing original coordinates of a GPX file\n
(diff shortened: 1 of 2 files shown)


UniQP:
Now featuring a change log entry ;-)

pstorch:
Verified. Fix works. Thanks for your contribution.

PR #6593, opened 2017-06-04
fix #6591 raise personal notes character limit

### LINKED ISSUE #6591: Raise note length limit to 2500 characters
-Download  or write  a personal Note  ( with morge the 500x charactars
- edit it, make it longer, what do you want.
- reupload it too the server

When the note is longer then 500x characters 
c:geo open a popup window and tell you that the personal note will cut off by the geocaching.com server after 500x characters .

the popup window should only open when he note is longer then 2500x characters  and then to tell you that the note will be cut off by the geocaching.com server after 2500x characters.

##### Version of c:geo used:
2017.04.27-RC
and
2017.05.22

Yes

Groundspeak has made an update to the personal note system and raised the max charakters from 500x  to 2500x charakter.

Is it possible to show the counted charaters of the personalnote field in the top of the edit field ( beside the close/save icon)  like (1478 / 2500)  ?

greetings from germany
Marco / Wulfman_Do

pstorch:
It also differentiates different limits for different connectors.


pstorch:
retest this please

PR #6598, opened 2017-06-07
Fix checkstyle warning: Remove unused import

UniQP:


pstorch:
Thanks for spotting it. I'm still puzzled why 1) my AS isn't cleaning it up (I've configured organize imports on the fly) and 2) why codacy doesn't report it?

PR #6601, opened 2017-06-10
Sort by number of votes in case of equal GCVote rating

UniQP:


SammysHP:
A cache with one 5-star rating will be ranked higher than a cache with 20 4.5-star ratings. While this commit solves a specific problem, it is not the solution for a correct rating. We get average rating and number of votes, this should be enough to sort caches in a better way.

http://www.evanmiller.org/how-not-to-sort-by-average-rating.html

UniQP:
@SammysHP I think this discussion is independent of the PR and should be moved into a new issue.

pstorch:
Maybe we should first open an issue for this and discuss possible solutions.

Bananeweizen:
Merging, as this just makes the existing sorting stable, independent of any of the problems mentioned by @SammysHP.

PR #6602, opened 2017-06-13
fix #6578 hide menu items by list type, disable items by current situation

### LINKED ISSUE #6578: Disable menu items instead of hiding them
Follow up of #6546:

@SammysHP wrote:
> With a shorter menu we could also think about disabling menu items instead of hiding them. It's easier to use a GUI if elements are always at the same place. What do you think about that? @Lineflyer @FUSSEL4 @pstorch

I think we have +3 so far for this proposal looking at the replies in #6546 and reactions on the posting.

pstorch:
~~As discussed in #6578 hide/show toplevel menu items and enable/disable submenu items.~~

As discussed in https://github.com/cgeo/cgeo/pull/6602#issuecomment-309776332:
Hide menu items not applicable for current list type.
Disable menu items not applicable for current situation.
Except "invert selection" which is only shown in select mode.

Search Result main menu:
screenshot_20170622-212345

Search Result manage caches submenu:
screenshot_20170621-223344

Concrete List main menu:
screenshot_20170622-212154

Concrete List manage caches submenu:
screenshot_20170621-223406

Concrete List manage lists submenu:
screenshot_20170621-224006

Concrete List selection mode:
screenshot_20170622-212202

All caches main menu:
screenshot_20170622-212253

All caches manage lists submenu:
screenshot_20170621-223559

History main menu:
screenshot_20170622-212303

Default list (empty) main menu:
screenshot_20170622-212243

Default list (empty) manage caches submenu:
screenshot_20170621-233234

Default list (empty) manage caches submenu:
screenshot_20170621-225959

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -695,10 +695,27 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         return true;
     }
 
+    private static void setVisibleEnabled(final Menu menu, final int itemId, final boolean visible, final boolean enabled) {
+        final MenuItem item = menu.findItem(itemId);
+        item.setVisible(visible);
+        item.setEnabled(enabled);
+    }
+
     private static void setVisible(final Menu menu, final int itemId, final boolean visible) {
         menu.findItem(itemId).setVisible(visible);
     }
 
+    private static void setEnabled(final Menu menu, final int itemId, final boolean enabled) {
+        menu.findItem(itemId).setEnabled(enabled);
+    }
+
+    /**
+     * Menu items which are not at all usable with the current list type should be hidden.
+     * Menu items which are usable with the current list type but not in the current situation should be disabled.
+     *
+     * @param menu
+     * @return
+     */
     @Override
     public boolean onPrepareOptionsMenu(final Menu menu) {
         super.onPrepareOptionsMenu(menu);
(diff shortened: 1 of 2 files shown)


Bananeweizen:
I think we should not change _every_ menu hiding to disabling. The menus are used in very different contexts, and some items may not be useful _at all_ in one or another of those contexts, therefore those should still be hidden instead of disabled.

"Clear history" is what sparked that. It should not be disabled, but really be hidden unless we are in a history list. And there it might be enabled depending on whether the list is empty.

pstorch:
That's the reason I made screenshots, to make it visible :)

"Clear history" is is only used in one kinde of lists, so that's pretty clear.
What about "Remove current list", "Rename current list" these are only allowed in user defined lists?

As a general Rule should we only disable items which might be enabled on the current list, but not applicable at the moment?

Bananeweizen:
> As a general Rule should we only disable items which might be enabled on the current list, but not applicable at the moment?

Sounds good to me.

pstorch:
Looks a bit different now.

main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -707,43 +717,38 @@ public boolean onPrepareOptionsMenu(final Menu menu) {
...
-            setVisible(menu, R.id.menu_copy_to_list, isOffline && !isEmpty);
-            setVisible(menu, R.id.menu_remove_from_history, !isEmpty && isHistory);
-            setVisible(menu, R.id.menu_clear_offline_logs, !isEmpty && (isHistory || isOffline) && containsOfflineLogs());
-            setVisible(menu, R.id.menu_import, isOffline);
-            setVisible(menu, R.id.menu_import_web, isOffline);
-            setVisible(menu, R.id.menu_import_gpx, isOffline);
-            setVisible(menu, R.id.menu_export, !isEmpty && (isHistory || isOffline));
-            setVisible(menu, R.id.menu_make_list_unique, !isEmpty && isOffline && listId != PseudoList.ALL_LIST.id);
+            setEnabled(menu, R.id.menu_switch_select_mode, !isEmpty);
+            setEnabled(menu, R.id.menu_invert_selection, adapter.isSelectMode());
```
Bananeweizen: This one might lead to confusion. A lot of people never use the selection mode at all, but will always see a menu item "Invert selection" now. I would exchange enabled for visible here.

pstorch:
Yes, I think we have to make an exception from the general rule here. I first wanted to make it consistent everywhere. But this is special.

pstorch:
Made an exception for "invert selection" item which is only shown in select mode.

PR #6603, opened 2017-06-16
fix #6530 event list sorting improved

### LINKED ISSUE #6530: Event lists can become very slow
Version: 2017.04.28-NB-e001def
Severity: important
Justification: serious user performance degradation

After upgrading to an NB with the new data format, I observe that newly created lists (e.g. from PQ import) are **very** slow to load, whereas older pre-existing lists (even with _more_ caches in them) load almost instantly.

The difference is notable even on a modern device (Note 4 “trltexx”); old lists are there in the blink of an eye, new lists (169 caches!) take about **three seconds** to display.

MichielK:
Fixes #6530 - Changed 2 things to improve sorting:
- store calculated value (eventTimeMinutes) so it doesn't need to be recalculated for every compare in EventDateComparator
- store regex-patterns in a static list so they don't have to be evaluated _every time_ for _every cache_

### DIFF
--- a/main/src/cgeo/geocaching/calendar/CalendarEntry.java
+++ b/main/src/cgeo/geocaching/calendar/CalendarEntry.java
@@ -45,7 +45,7 @@ class CalendarEntry {
                 StringUtils.defaultString(cache.getPersonalNote()),
                 cache.getName(),
                 cache.getCoords() == null ? "" : cache.getCoords().format(GeopointFormatter.Format.LAT_LON_DECMINUTE_RAW),
-                cache.guessEventTimeMinutes());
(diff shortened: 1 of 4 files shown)


pstorch:
ok to test

pstorch:
No, I'm not allowed to command the CI server. 😀

pstorch:
ok to test

pstorch:
@MichielK I think the cached value eventTimMinutes needs to be invalidated if the cache description or the short description changes.

MichielK:
@pstorch good catch and same holds true if cachetype changes - I committed a fix.

PR #6604, opened 2017-06-18
upgrade to gradle 4.0

Bananeweizen:


Bananeweizen:
Does someone understand the errors in the build output regarding mocked tests, like http://ci.cgeo.org/job/cgeo%20pull%20request/957/testReport/junit/cgeo.geocaching.location/GeopointTest/testParcelable/

If I get it right, we have 7 test cases, which are in the (non Android) src test folder, which actually invoke some Android methods (and should therefore be in the other test source folder for Android tests). How comes we did not notice this before?

Of course, the simple solution is moving these test cases to the Android test folder, but I still don't understand how this was working before...

PR #6608, opened 2017-06-20
fix #6605 enable TLS v1.2 Support for Android 4.1 - 4.4

### LINKED ISSUE #6605: no images in decription
trying to watch a cache description. Online or offline (already stored on the device).
The "download images" checkbox is set
This happens only with some caches. E.g.: GC6ZPRE; GC6BHT6

I simply see no image of the cache description. Neither in the cache description tab nor in the image tab.

See the pictures on my Android device

##### Version of c:geo used:
2017.05.22

Yes

##### System information:
several SAMSUNG devices, including GALAXY TAB 2 with Android 4.2.2 and S3 mini with Androis 4.1.2

same problem as #5848??

pstorch:


Bananeweizen:
This is completely not my area of knowledge. So while the change looks okay from the formatting point of view, I cannot say anything about the content as such. Sorry.

pstorch:
We have to test it in the nightlies anyway.

PR #6613, opened 2017-06-22
Externalize mapsforge v3 thunderforrest (OpenCycleMap) ApiKey

pstorch:
To be able to exchange the Thunderforrest ApiKey via our CI, it's necessary to externalize the key also for the legacy mapsforge v3.

It's not the ideal solution, but I hope we can phase out the old mapsforge implementation in the near future.


Bananeweizen:
:+1:

PR #6615, opened 2017-06-24
Fix #6606: Use weighted arithmetic mean to sort GCVote ratings

### LINKED ISSUE #6606: Improve sort by rating
Follow-up of #6601:

A cache with one 5-star rating will be ranked higher than a cache with 20 4.5-star ratings. While this commit solves a specific problem, it is not the solution for a correct rating. We get average rating and number of votes, this should be enough to sort caches in a better way.

http://www.evanmiller.org/how-not-to-sort-by-average-rating.html
https://stackoverflow.com/questions/1411199/what-is-a-better-way-to-sort-by-a-5-star-rating

UniQP:
Intended as a basis for discussion.

Not tested yet.

### DIFF
--- a/main/res/values/changelog_master.xml
+++ b/main/res/values/changelog_master.xml
@@ -4,7 +4,7 @@
   <string name="changelog_master" translatable="false">
     <b>Next feature release:</b>\n
     · New: Use abbreviation for name of geocaching service on main screen\n
-    · New: Caches with the same GCVote rating are sorted by the number of votes\n
+    · New: GCVote ratings are now sorted by a weighted arithmetic mean that also considers the number of votes\n
     \n
     \n
   </string>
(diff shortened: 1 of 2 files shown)


Bananeweizen:
2.5 is not the average (and that was already wrong before). It should be 3, since that is the average between 1 and 5, which are the limits of the rating (and even the website uses "average" as textual value for 3).

In fact, I would even more like it to have a random sample of 10000 caches and their votes to calculate on actual in the field average rating, but I'm to lazy. :)

UniQP:
It was 3.3995 for GC50*** (6782 caches with votes) → I changed it to 3.4.

Bananeweizen:
:+1: I'm fine with merging.

UniQP:
I finally got my emulator working again to test the PR. Seems to work as intended: 4.3 (16 votes) > 5.0 (2 votes) > 4.1 (9 votes).

pstorch:
I don't really use GCVote and I haven't thought much about 5-star-rating-sorting, but I just verified this PR.

With this change the sorting of my homezone PQ was as follows:

1. 5.0 with 4 votes

1. 4.0 with 9 votes

1. 4.8 with 2 votes

1. 4.2 with 3 votes

1. 5.0 with 1 vote

1. 4.0 with 3 votes

1. 3.8 with 5 votes

1. 4.5 with 1 vote

1. 4.0 with 2 votes

1. 3.8 with 3 votes


Looks reasonable to me.

Issue #6610, opened 2017-06-22
Remove OpenCycleMap


SammysHP:
Since a while OpenCycleMap requires an API key to monitor and limit the use of their tiles. In #6512 we added that API key in c:geo.

Now we were contacted by Andy that we have reached the limit of the free plan. Based on the number of requested tiles we would need the "large business" plan for £395 per month. I'm still surprised that there are one million tiles requested per days on average.

We are not able to pay that money so the only solution would be to remove OpenCycleMap completely.

There would be another way, but I'm not sure if that is allowed: Make the key dynamic and let users enter their own key.

SammysHP:
PS: I wonder how many tiles are used from OpenStreetMap. There is a chance that we produce a lot (or too much) of traffic there as well.

mucek4:
I looked at OpenCycleMap. On my 1080p phone we request like 5x8 tiles. So 40 tiles per screen.

pstorch:
Is it an option to provide a setting for the user to use individual keys? I mean is it possible/feasible to let the user apply for their own key? Then it is up to the user.

SammysHP:
It is possible, sure. But I don't think that it is allowed by the terms of use (although I can't find anything about this topic in it).

pstorch:
At least the hardcoded api key in the mapsforge v3 library is externalized. Now we can decide to:
- make a new release (from release or master) with a new api key (configured on CI server). 
- leave it empty and let the user configure a personal api key in settings (a new setting needs to be implemented)
- or remove OpenCycleMap completely

Lineflyer:
What abou just asking them via reply to their mail how to proceed....telling we are a non profit opensource app?We could also ask about allowance of individual keys.
We should also tell them, that we are surprised by the high traffic and that it takes us some weeks to remove it from our app in tbe market if there is no solution.
null

SammysHP:
I'd try the second option (especially after the fact that it is easy to "steal" the key) but I'm sure Andy wouldn't be happy with it.

Not really sure if it makes sense to change the key. Maybe someone uses the old key outside of c:geo, but that might happen again. Or c:geo really creates so many requests, then we have to remove it anyway.

@Lineflyer 
I'm already in contact with Andy but there's no special offer for open-source apps. He knows that we were surprised, but will not wait for several weeks. If there's no change by the end of next week he will just block our key probably.

pstorch:
[...]

PR #6616, opened 2017-06-24
Remove Thunderforest API key and provide user setting

pstorch:
Thunderforest API key for OSM Cyclemap is not provided by c:geo anymore, but supports it via a user setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,6 @@ Request your personal API key for the various [OpenCaching](http://www.opencachi
 * [opencaching.us OKAPI signup](http://www.opencaching.us/okapi/signup.html)
 * [opencache.uk OKAPI signup](http://www.opencache.uk/okapi/signup.html)
 
-Request API key for OpenCycleMap from [Thunderforest](https://manage.thunderforest.com/users/sign_up?plan_id=5).
-
 ### Building with gradle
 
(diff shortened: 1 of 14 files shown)


Bananeweizen:
Will this automatically switch an existing user with Thunderforest active to another map? I've seen that a check is done in the settings activity to avoid empty keys with your change, but I'm thinking of a user who just upgrades and never opens settings. Or does the current code in the mapsforge provider already fall back to the default source then?

pstorch:
To be honest, I have not tested this case. I have tested the SettingsActivity forth and back. It drove me crazy 😂. Till it updated the list of MapSources correctly. There I tested that if the Cyclemap is chosen and you remove the api key, the list above is reduced and the default is selected. 
I have to test the migration case explicitly. I think it will fall back to the default, but I'm not sure.

pstorch:
Ok, tested migration successfully. If the Cyclemap was the chosen map source before, c:geo falls back to the default (Mapnik) after this change.

Lineflyer:
Let me set "Do not merge" in order to first discuss the further proceeding in #6610

pstorch:
I'll close here, due to better ideas came up in https://github.com/cgeo/cgeo/issues/6610#issuecomment-315747226

PR #6619, opened 2017-06-28
fix #6550 samsung compass

### LINKED ISSUE #6550: Compass on Samsung Galaxy S6 fails to rotate after Android 7 update
Placeholder (to be completed with more details from support mail).

pstorch:
Change from SENSOR_DELAY_NORMAL to SENSOR_DELAY_GAME


pstorch:
Doesn't fix the problem

Issue #2731, opened 2013-05-15
zero caches on live map.


stayhuman:
I've tried everything! Even removed the app and reinstalled it. App worked great for the first two months, now my live map will only show stored caches. Plenty of ways around this, but it still sucks.

PR #6624, opened 2017-07-01
fix #6611 identify c:geo in user agent

### LINKED ISSUE #6611: Identify in user agent
In #1025 we decided to use a generic user agent because we didn't want to be tracked by Groundspeak and to force a desktop version of the website (for parsing).

Meanwhile we [send an "app=cgeo" parameter](https://github.com/cgeo/cgeo/blob/7b8ac7e898055878a6d645913bc2451ff2836ce2/main/src/cgeo/geocaching/connector/gc/GCMap.java#L69) for GC tile access, so we might add "cgeo" to the user agent as well. That would allow other services to measure the load that c:geo creates on their servers.

pstorch:
- use Androids WebView to determine system user agent
- append `[cgeo]` to the user agent string
- set userAgent in mapsforge mapdownloder (V3 and new map)
- invert setting (useNativeUa with useDesktopUa) default is androids webview user agent
- update desktop user agent string to a more recent version of FF


Bananeweizen:
Do I get it right, that the userAgent value is not cached right now (I may have missed something)? Since in the worst case this involves creating a new webview, it seems reasonable to me to only calculate this once, but not every time the method is invoked. Besides that, it looks good to me.

And a question from the curious: Did you add the annotation regarding UIThread because you read about it, because it was in some code from stackoverflow or because Android Studio or some other tool actively suggested it? Just wondering if we should use this more...

pstorch:
> Do I get it right, that the userAgent value is not cached right now (I may have missed something)? 
Oh yes, you are right. I got tricked with the single instance of the OkHttpClient. But the the Interceptor gets called on every request. I'll add caching for the userAgent string. Just need to think about the cache invalidation if the Setting changes.

> And a question from the curious: Did you add the annotation regarding UIThread because you read about it, because it was in some code from stackoverflow or because Android Studio or some other tool actively suggested it? 
Again you caught me ;) 
I copied it from SO as it sounded reasonable to me. I'll take it as an opportunity for me to read about it.

pstorch:
Now with caching of the userAgent string. I also added the version info and extended the system information to show the current userAgent.

Regarding `@UIThread`: https://github.com/androidannotations/androidannotations/wiki/WorkingWithThreads explains it. In general the Androidannotations looks like a really helpful library http://androidannotations.org/. I should have know about it earlier. This can help to write cleaner code. It also seams like Butterknife is using it internally.

Lineflyer:
This might directly impact the answer we get from the server. How do you consider the risk of this change?

I am thinking about e.g. that for some user agents the server response might be different (e.g. mobile optimized scheme). Or is this something completey different/independent?

At least from testing perspective I am a little unhappy with this change as it might lead to different behaviour of the app depending on the device used.

SammysHP:
I also vote for a fixed mobile user agent. c:geo is not a webview, it is an independent browser. Also it would remove some complexity of the code.

pstorch:
I understand the concerns and we should at least take some time to test it on master. 
But I think taking the user agent from the web view is more correct (it better identifies the device) than pretending to be an arbitrary desktop browser.

pstorch:
retest this please

pstorch:
As `master` is merged on `release` do we want to test this on the nightlies?

pstorch:
As long as we have some time on master before the next release, I really would like to test this.

Lineflyer:
@pstorch
If you like you could merge it...I can try to keep an eye on it, whether I see any problems.
I am only afraid, that I might not be able to identify possible underlying problems coming with it.

And - to mention it again - it might result in different server repsonses thus lead to different behaviour of the app depending on the device used. A testers nightmare.

Anyway If you think merging and testing in nightly might bring a benefit, we could merge but should revert it at some point unless we have a high percentage of confidence, that it does not bring any negative aspects.

BTW: I still do not understand the advantage in this...can someone explain it to me? Or is it just, that the server has better statistics (which I don't care to much about).

pstorch:
@Lineflyer the reasoning behind it was discussed in #6611 

Before that change c:geo was only identified when requesting map tiles from GC in #2731 (interesting discussion btw, also in http://irclogs.cgeo.org/?year=2013&month=5&day=16 )

Lineflyer:
Lets continue in #6611 then.

pstorch:
Let's close this one for now, there will be a new PR.

PR #6625, opened 2017-07-01
fix #6509 relative URLs in cache description

### LINKED ISSUE #6509: Non working link in oc caches
Load OC10406 or any other safari cache. All hyperlinks in the cache description are working, except for the 'Suche' link.

pstorch:
Iterates over the links in the description, looks for relative URLs and prepends them with the HostUrl from the Connector.


Bananeweizen:
I would rather like to fix this in the parser, to avoid running code lately in the UI all the time. But admittedly, when I looked into the GPX parser for how to approach this, it is also rather complex, since URLs don't exist there and regex-ing through all the GPX contents is also not nice.
Therefore :+1:

PR #6640, opened 2017-07-17
fix #6610 remove opencyclemap

### LINKED ISSUE #6610: Remove OpenCycleMap

pstorch:
Maybe @Lineflyer can check/improve the changelog with the link to the FAQ


pstorch:
The "experimental" build pipeline failed, but the normal was successfull.

Retest this please.

pstorch:
retest this please

Lineflyer:
I will adapt the change log later.

Issue #2038, opened 2012-09-12
Provide calculator in edit waypoint activity


triakcz:
Hi,

when solving multis, it is sometimes to needed to calculate coordinates, so I put cgeo into background, launch calculator, then remember counted number and then open cgeo back ... I'm proposing special button ... which will launch calculator (if it is possible to do it this way, I will check). And after pressing OK there, It could return number back to cgeo focused field.

Hadatko:
Hi,
how its going with this feature? I saw many issues related to this but i am not sure if is any progress on it. If not, i could do it for us ;)

SammysHP:
There are calculators that can be put on top of an app, e.g. the one in Cyanogenmod.

Hadatko:
I was thinking about the MultiSolver thing.

S-Bartfast:
Hello, I have used c:geo for many years but am just now getting into Android programming and have been thinking about this feature a lot.  As an exercise I decided to start working to a design to do exactly this.  I have no idea if I will be able to produce anything of value to the community at large but I thought I would let you all know I am at least having a go at things here.

At present I am thinking to add an extra button to the Coordinate Input Dialogue which will open a new dialogue with simple tools for calculating waypoints for miltis:
![coordinateinputdialogue](https://cloud.githubusercontent.com/assets/6573522/22402004/1d58ab10-e637-11e6-930f-8672c7f6ef68.png)

Here is a mockup of what I am thinking the calculator itself could look like:
![geocalc](https://cloud.githubusercontent.com/assets/6573522/22402005/249a9640-e637-11e6-8f60-89221ca98ace.png)

There are ofcourse many details still to be worked out but this it the general direction of how I am thinking.

Could someone please post an image of what HandyGeocaching waypoint calculator looks like?  I've had a search but can't seem to find it.

Lineflyer:
I somehow like it, but problems may arise in the detail:
- In a typical multi cache you wont have all values initially. So you could only use the function at the very end to calculate the final, or you need the ability to store it wihout filling all variables and editing it later on.
- A simple variable per digit does not always fulfill the calculation requirement. I know many caches which use something like N51 2(A-1).(B+4)(C)(D/2) ...

S-Bartfast:
Thanks for those points Lineflyer.

I have thought about both your points but haven't as of yet decided on a solution.

In terms of your first point:
[...]

PR #6637, opened 2017-07-16
Added a Coordinate Calculator: Issue #2038

S-Bartfast:
At present I am happy with how the calculator functions.  For further details see the issue discussion: https://github.com/cgeo/cgeo/issues/2038

My only concern is that the state of the calculate is saved in the waypoint 'ContentValues' as an SQL 'BUNDLE', which may not be acceptable in the c:geo ethos.

PR #6645, opened 2017-07-23
Added a Coordinate Calculator: Issue #2038

S-Bartfast:
Here we go again...

### DIFF
--- a/main/res/layout-land/coordinatescalculate_dialog.xml
+++ b/main/res/layout-land/coordinatescalculate_dialog.xml
@@ -1,0 +1,90 @@
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+              xmlns:tools="http://schemas.android.com/tools"
+              android:layout_width="match_parent"
+              android:layout_height="match_parent"
+              android:orientation="vertical"
+              tools:context=".ui.dialog.CoordinatesCalculateDialog" >
+
+    <include layout="@layout/dialog_title_ok_cancel" />
+
+    <ScrollView
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        android:id="@+id/scroller"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:fillViewport="true">
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:baselineAligned="false">
+
+            <LinearLayout
+                android:id="@+id/scroller_child"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_weight="1"
+                android:orientation="vertical" >
+
+                <include layout="@layout/coordinates_calculate" />
+                <include layout="@layout/coordinates_equations" />
+
+                <TextView
+                    android:id="@+id/latRes"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+
+                    style="@style/edittext_dialog"
+                    android:focusable="false"
+                    android:focusableInTouchMode="false"
+                    android:clickable="false"
+
+                    android:gravity="right"
+                    android:textAlignment="gravity"
+                    android:text="@string/lat_point_min_format"/>
+
+                <TextView
+                    android:id="@+id/lonRes"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+
+                    style="@style/edittext_dialog"
+                    android:focusable="false"
+                    android:focusableInTouchMode="false"
+                    android:clickable="false"
+
+                    android:paddingRight="16dp"
+                    android:gravity="end"
+                    android:textAlignment="gravity"
+                    android:text="@string/lon_point_min_format"/>
+
+                <Button
+                    android:id="@+id/done"
+                    style="@style/button_full"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:text="@string/waypoint_done" />
+            </LinearLayout>
+
+            <LinearLayout
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:layout_weight="1"
+                android:orientation="vertical" >
+
+                <EditText
+                    android:id="@+id/notes_text"
+                    android:layout_width="match_parent"
+                    android:layout_height="match_parent"
+                    style="@style/edittext_full"
+                    android:textSize="16dp"
+                    android:singleLine="false"
+                    android:overScrollMode="always"
+                    android:scrollbarStyle="insideInset"
+                    android:scrollbars="vertical"
+
+                    android:hint="@string/waypointcalc_notes_prompt" />
+            </LinearLayout>
+        </LinearLayout>
+    </ScrollView>
+</LinearLayout>
--- a/main/res/layout/coordinates_calculate.xml
+++ b/main/res/layout/coordinates_calculate.xml
@@ -1,0 +1,204 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+       xmlns:tools="http://schemas.android.com/tools"
+       android:layout_width="match_parent"
+       android:layout_height="match_parent" >
+
+    <Spinner
+        android:id="@+id/spinnerCoordinateFormats"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginLeft="10dp"
+        android:layout_marginRight="10dp"
+        tools:listitem="@android:layout/simple_spinner_item" />
+
+    <!-- EditTexts for 'Plain' input -->
+    <LinearLayout
+        android:id="@+id/PlainFormat"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical">
+        <EditText
+            android:id="@+id/PlainLat"
+            style="@style/edittext_full"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:inputType="text|textCapCharacters"
+            android:hint="@string/latitude" />
+
+        <EditText
+            android:id="@+id/PlainLon"
+            style="@style/edittext_full"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:inputType="text|textCapCharacters"
+            android:hint="@string/longitude" />
+    </LinearLayout>
+
+    <!-- H DDD°MM'SS.PPPPP" -->
+    <GridLayout
+        android:id="@+id/coordTable"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:columnCount="17">
+
+        <Button
+            android:id="@+id/ButtonLatHem"
+            android:layout_width="33dp"
+            android:layout_gravity="center_horizontal|center_vertical"
+            style="@style/button"
+
+            android:layout_marginLeft="1dip"
+            android:layout_marginRight="5dip"
+            android:layout_marginTop="2dip"
+            android:layout_marginBottom="2dip" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatDeg_100"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatDeg_010"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatDeg_001"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LatDegChar"
+            android:text="°"
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatMin_10"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatMin_01"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LatMinChar"
+            android:text="'"
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatSec_10"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatSec_01"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LatPointChar"
+            android:text="."
+            android:textSize="22sp"
+            android:layout_gravity="center_vertical|left"
+            android:layout_marginLeft="0dp"
+            android:layout_marginRight="0dp"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatPnt_10000"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatPnt_01000"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatPnt_00100"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatPnt_00010"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLatPnt_00001"
+            style="@style/button_digit" />
+        <TextView
+            android:id="@+id/LatLastUnitsChar"
+            android:text='"'
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+
+        <Button
+            android:id="@+id/ButtonLonHem"
+            android:layout_width="33dp"
+            android:layout_gravity="center_horizontal|center_vertical"
+            style="@style/button"
+
+            android:layout_marginLeft="1dip"
+            android:layout_marginRight="5dip"
+            android:layout_marginTop="2dip"
+            android:layout_marginBottom="2dip" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonDeg_100"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonDeg_010"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonDeg_001"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LonDegChar"
+            android:text="°"
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonMin_10"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonMin_01"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LonMinChar"
+            android:text="'"
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonSec_10"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonSec_01"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LonPointChar"
+            android:text="."
+            android:textSize="22sp"
+            android:layout_gravity="center_vertical"
+            tools:ignore="HardcodedText" />
+
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonPnt_10000"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonPnt_01000"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonPnt_00100"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonPnt_00010"
+            style="@style/button_digit" />
+        <cgeo.geocaching.ui.CalculateButton
+            android:id="@+id/ButtonLonPnt_00001"
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LonLastUnitsChar"
+            android:text='"'
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+    </GridLayout>
+</merge>
--- a/main/res/values/dimens.xml
+++ b/main/res/values/dimens.xml
@@ -9,4 +9,6 @@
     <dimen name="app_defaultsize_w">632.0dip</dimen>
     <dimen name="app_defaultsize_h">598.0dip</dimen>
 
+    <dimen name="equation_name_width">40dip</dimen>
+
 </resources>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -239,6 +239,7 @@
     <string name="warn_rendertheme_missing">Map theme not found.</string>
     <string name="warn_pocket_query_select">No Pocket query selected.</string>
     <string name="warn_no_pocket_query_found">No Pocket query found online.</string>
+    <string name="warn_invalid_character">Character must be a LETTER, DIGIT or SPACE</string>
     <string name="err_read_pocket_query_list">Error reading Pocket query list.</string>
     <string name="info_log_posted">c:geo successfully submitted the log.</string>
     <string name="info_log_saved">c:geo successfully saved the log.</string>
@@ -252,6 +253,8 @@
     <string name="info_stored_image">New image saved to:</string>
     <string name="info_storing_static_maps">Trying to store static maps</string>
     <string name="info_cache_saved">The cache has been stored locally</string>
+    <string name="info_calculator_state_saved">Current state of the calculator has been saved to waypoint.</string>
+    <string name="warn_calculator_state_save">Be sure to save waypoint if you want to keep the calculator state.</string>
     <string name="err_trackable_log_not_anonymous">Logging %1$s from c:geo cannot be done anonymously. Do you want to open the settings to fill in your credentials from %2$s?</string>
     <string name="err_trackable_no_preference_activity">Sorry, cannot find a suitable preference screen.</string>
     <string name="warn_discard_changes">Unsaved waypoint changes discarded!</string>
@@ -991,6 +994,7 @@
         <item>@string/waypoint_duplicate</item>
     </array>
     <string name="waypoint_cache_coordinates">Cache coordinates</string>
+    <string name="waypoint_calculate_coordinates">Calculate coordinates</string>
     <string name="waypoint_custom">Custom</string>
     <string name="waypoint_my_coordinates">My coordinates</string>
     <string name="waypoint_bearing">Bearing in °</string>
@@ -1025,6 +1029,7 @@
     <string name="waypoint_duplicate">Duplicate waypoint</string>
     <string name="waypoint_duplicated">Waypoint duplicated</string>
     <string name="waypoint_copy_of">Copy of</string>
+    <string name="waypointcalc_notes_prompt">Enter notes here…</string>
     <string name="search_history">History</string>
     <string name="search_history_empty">No previous destinations</string>
     <string name="search_remove_destination">Destination removed</string>
@@ -1590,5 +1595,11 @@
     <string name="init_low_disk_space_message">The geocache data directory is running low on storage space. This can cause malfunctions of c:geo. Please free up some space.</string>
     <string name="init_confirm_debug">c:geo is running in debug mode!</string>
     <string name="list_confirm_debug_message">The debug mode causes instabilities! Do you want to deactivate the debug mode?</string>
+    <string name="equation_hint">a^2 + b^2</string>
+    <string name="free_variable_hint">2 + 2</string>
+    <string name="empty_equation_result">_</string>
+    <string name="equation_error_result">*</string>
+    <string name="lat_point_min_format">N DDD° MM.PPP</string>
+    <string name="lon_point_min_format">E DDD° MM.PPP</string>
 
 </resources>
--- a/main/src/cgeo/geocaching/models/CalcState.java
+++ b/main/src/cgeo/geocaching/models/CalcState.java
@@ -1,0 +1,118 @@
+package cgeo.geocaching.models;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+import cgeo.geocaching.settings.Settings.CoordInputFormatEnum;
+import cgeo.geocaching.ui.CalculateButton;
+import cgeo.geocaching.ui.CalculatorVariable;
+import cgeo.geocaching.ui.JSONAble;
+import cgeo.geocaching.ui.JSONAbleFactory;
+
+/**
+ * This class is designed to capture the current state of the coordinate calculator such that it can be preserver for latter use.
+ *
+ * All the relevant information is in a serializable form such that it can be stored as a bundle in waypoint's 'ContentValues'.
+ */
+
+public class CalcState implements Serializable {
+    public static final char ERROR_CHAR = '#';
+    public static final String ERROR_STRING = "???";
+
+    public final CoordInputFormatEnum format;
+    public final String plainLat;
+    public final String plainLon;
+    public final char latHemisphere;
+    public final char lonHemisphere;
+    public final List<? extends JSONAble> buttons;
+    public final List<? extends JSONAble> equations;
+    public final List<? extends JSONAble> freeVariables;
+    public final List<? extends JSONAble> variableBank;
+    public final String notes; // Note, we have to use a String rather than an Editable as Editable's can't be serialized
+
+    public CalcState(final CoordInputFormatEnum format,
+                     final String plainLat,
+                     final String plainLon,
+                     final char latHem,
+                     final char lonHem,
+                     final List<CalculateButton.ButtonData> buttons,
+                     final List<CalculatorVariable.VariableData> equations,
+                     final List<CalculatorVariable.VariableData> freeVariables,
+                     final List<CalculatorVariable.VariableData> bankVariables,
+                     final String notes) {
+        this.format = format;
+        this.plainLat = plainLat;
+        this.plainLon = plainLon;
+        latHemisphere = latHem;
+        lonHemisphere = lonHem;
+        this.buttons = buttons;
+        this.equations = equations;
+        this.freeVariables = freeVariables;
+        this.variableBank  = bankVariables;
+        this.notes = notes;
+    }
+
+    private CalcState(final JSONObject json) {
+        format = CoordInputFormatEnum.values()[json.optInt("format", 2)];
+        plainLat = json.optString("plainLat");
+        plainLon = json.optString("plainLon");
+        latHemisphere = (char) json.optInt("latHemisphere", ERROR_CHAR);
+        lonHemisphere = (char) json.optInt("lonHemisphere", ERROR_CHAR);
+        buttons       = createJSONAbleList(json.optJSONArray("buttons"),       new CalculateButton.ButtonDataFactory());
+        equations     = createJSONAbleList(json.optJSONArray("equations"),     new CalculatorVariable.VariableDataFactory());
+        freeVariables = createJSONAbleList(json.optJSONArray("freeVariables"), new CalculatorVariable.VariableDataFactory());
+        variableBank = new ArrayList<>(); // "variableBank" intentionally not loaded.
+        notes = json.optString("notes");
+    }
+
+    private static ArrayList<? extends JSONAble> createJSONAbleList(final JSONArray json, final JSONAbleFactory factory) {
+        final int length = json != null ? json.length() : 0;
+        final ArrayList<JSONAble> returnValue = new ArrayList<>(length);
+
+        for (int i = 0; i < length; i++) {
+            returnValue.add(factory.fromJSON(json.optJSONObject(i)));
+        }
+
+        return returnValue;
+    }
+
+    public static CalcState fromJSON(final String json) throws JSONException {
+        if (json == null) {
+            return null;
+        }
+
+        return new CalcState(new JSONObject(json));
+    }
+
+    private JSONArray toJSON(final List<? extends JSONAble> items) throws JSONException {
+        final JSONArray returnValue = new JSONArray();
+
+        for (final JSONAble item : items) {
+            returnValue.put(item.toJSON());
+        }
+
+        return returnValue;
+    }
+
+    public JSONObject toJSON() throws JSONException {
+        final JSONObject returnValue = new JSONObject();
+
+        returnValue.put("format", format.ordinal());
+        returnValue.put("plainLat", plainLat);
+        returnValue.put("plainLon", plainLon);
+        returnValue.put("latHemisphere", latHemisphere);
+        returnValue.put("lonHemisphere", lonHemisphere);
+        returnValue.put("buttons", toJSON(buttons));
+        returnValue.put("equations", toJSON(equations));
+        returnValue.put("freeVariables", toJSON(freeVariables));
+        // "variableBank" intentionally left out.
+        returnValue.put("notes", notes);
+
+        return returnValue;
+    }
+}
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -20,6 +20,7 @@ import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.Viewport;
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
+import cgeo.geocaching.models.CalcState;
 import cgeo.geocaching.models.Destination;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
@@ -83,16 +84,13 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONException;
 
 public class DataStore {
 
     public static final String DB_FILE_NAME = "data";
     public static final String DB_FILE_NAME_BACKUP = "cgeo.sqlite";
     public static final String DB_FILE_CORRUPTED_EXTENSION = ".corrupted";
-
-    private DataStore() {
-        // utility class
-    }
 
     public enum StorageLocation {
         HEAP,
@@ -162,7 +160,7 @@ public class DataStore {
                     "cg_caches.watchlistCount";          // 42
 
     /** The list of fields needed for mapping. */
-    private static final String[] WAYPOINT_COLUMNS = { "_id", "geocode", "updated", "type", "prefix", "lookup", "name", "latitude", "longitude", "note", "own", "visited", "user_note", "org_coords_empty" };
+    private static final String[] WAYPOINT_COLUMNS = { "_id", "geocode", "updated", "type", "prefix", "lookup", "name", "latitude", "longitude", "note", "own", "visited", "user_note", "org_coords_empty", "calc_state" };
 
     /** Number of days (as ms) after temporarily saved caches are deleted */
     private static final long DAYS_AFTER_CACHE_IS_DELETED = 3 * 24 * 60 * 60 * 1000;
@@ -172,7 +170,7 @@ public class DataStore {
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
-    private static final int dbVersion = 72;
+    private static final int dbVersion = 73;
     public static final int customListIdOffset = 10;
 
     @NonNull private static final String dbTableCaches = "cg_caches";
@@ -269,7 +267,8 @@ public class DataStore {
             + "own INTEGER DEFAULT 0, "
             + "visited INTEGER DEFAULT 0, "
             + "user_note TEXT, "
-            + "org_coords_empty INTEGER DEFAULT 0"
+            + "org_coords_empty INTEGER DEFAULT 0, "
+            + "calc_state TEXT"
             + "); ";
     private static final String dbCreateSpoilers = ""
             + "CREATE TABLE " + dbTableSpoilers + " ("
@@ -339,6 +338,10 @@ public class DataStore {
             + "latitude DOUBLE, "
             + "longitude DOUBLE "
             + "); ";
+
+    private DataStore() {
+        // utility class
+    }
 
     private static final Single<Integer> allCachesCountObservable = Single.create(new SingleOnSubscribe<Integer>() {
         @Override
@@ -871,6 +874,14 @@ public class DataStore {
                             Log.e("Failed to upgrade to ver. 72", e);
                         }
                     }
+                    // Adds coord calculator state to the waypoint
+                    if (oldVersion < 73) {
+                        try {
+                            db.execSQL("ALTER TABLE " + dbTableWaypoints + " ADD COLUMN calc_state TEXT");
+                        } catch (final Exception e) {
+                            Log.e("Failed to upgrade to ver. 73", e);
+                        }
+                    }
                 }
 
                 db.setTransactionSuccessful();
@@ -1383,6 +1394,8 @@ public class DataStore {
                 values.put("visited", oneWaypoint.isVisited() ? 1 : 0);
                 values.put("user_note", oneWaypoint.getUserNote());
                 values.put("org_coords_empty", oneWaypoint.isOriginalCoordsEmpty() ? 1 : 0);
+                putCalcState(values, oneWaypoint.getCalculatorStoredState());
+
                 if (oneWaypoint.getId() < 0) {
                     final long rowId = database.insert(dbTableWaypoints, null, values);
                     oneWaypoint.setId((int) rowId);
@@ -1418,6 +1431,16 @@ public class DataStore {
     private static void putCoords(final ContentValues values, final Geopoint coords) {
         values.put("latitude", coords == null ? null : coords.getLatitude());
         values.put("longitude", coords == null ? null : coords.getLongitude());
+    }
+
+    private static void putCalcState(final ContentValues values, final CalcState calcState) {
+        if (calcState != null) {
+            try {
+                values.put("calc_state", calcState.toJSON().toString());
+            } catch (final JSONException e) {
+                Log.e("Unable to write calculator state information", e);
+            }
+        }
     }
 
     /**
@@ -1463,6 +1486,8 @@ public class DataStore {
             values.put("own", waypoint.isUserDefined() ? 1 : 0);
             values.put("visited", waypoint.isVisited() ? 1 : 0);
             values.put("org_coords_empty", waypoint.isOriginalCoordsEmpty() ? 1 : 0);
+            putCalcState(values, waypoint.getCalculatorStoredState());
+
             if (id <= 0) {
                 final long rowId = database.insert(dbTableWaypoints, null, values);
                 waypoint.setId((int) rowId);
@@ -2003,6 +2028,11 @@ public class DataStore {
         waypoint.setNote(cursor.getString(cursor.getColumnIndex("note")));
         waypoint.setUserNote(cursor.getString(cursor.getColumnIndex("user_note")));
         waypoint.setOriginalCoordsEmpty(cursor.getInt(cursor.getColumnIndex("org_coords_empty")) != 0);
+        try {
+            waypoint.setCalculatorStoredState(CalcState.fromJSON(cursor.getString(cursor.getColumnIndex("calc_state"))));
+        } catch (final JSONException e) {
+            Log.e("Unable to read calculator state information", e);
+        }
 
         return waypoint;
     }
--- a/main/src/cgeo/geocaching/ui/CalculateButton.java
+++ b/main/src/cgeo/geocaching/ui/CalculateButton.java
@@ -1,0 +1,303 @@
+package cgeo.geocaching.ui;
+
+import android.content.Context;
+import android.graphics.Color;
+import android.graphics.Typeface;
+import android.support.v4.content.ContextCompat;
+import android.text.Editable;
+import android.util.AttributeSet;
+import android.view.View;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.Serializable;
+
+import cgeo.geocaching.R;
+
+import static cgeo.geocaching.models.CalcState.ERROR_CHAR;
+
+/**
+ * This class derives from EditButton and handles all the attributes that are unique to this particular application of button.
+ *
+ * In particular this button stores three values:
+ * 1. The value obtained from the waypoint coordinates that was used at the time the calculator was created
+ * 2. An 'AutoChar' which is the next alphabetic character used when the user clicks on the button
+ * 3. A custom character which is a specific value assigned by the user when performing a long-click.
+ *
+ * The button displays one of these values as appropriate as the text of the button.
+ */
+public class CalculateButton extends EditButton {
+
+    // Flag values used to designate that no AutoChar has been set
+    private static final char EMPTY_CHAR = '-';
+
+    // The three stated the button can be put into
+    private enum ValueType {
+        INPUT_VAL {
+            @Override
+            char getLabel(final ButtonData buttonData) {
+                return buttonData.inputVal;
+            }
+
+            @Override
+            char setAutoChar(final ButtonData buttonData, final char nextChar) {
+                buttonData.autoChar = nextChar;
+                return nextChar;
+            }
+
+        },
+        AUTO_CHAR {
+            @Override
+            char getLabel(final ButtonData buttonData) {
+                return buttonData.autoChar;
+            }
+
+            @Override
+            char setAutoChar(final ButtonData buttonData, final char autoChar) {
+                char nextChar = autoChar;
+                buttonData.autoChar = nextChar++;
+                return nextChar;
+            }
+        },
+        CUSTOM {
+            @Override
+            char getLabel(final ButtonData buttonData) {
+                return buttonData.customChar;
+            }
+
+            @Override
+            char setAutoChar(final ButtonData buttonData, final char autoChar) {
+                char nextChar = autoChar;
+                buttonData.autoChar = nextChar;
+                if ('A' <= buttonData.customChar && buttonData.customChar <= 'Z') {
+                    nextChar = buttonData.customChar;
+                    nextChar++;
+                }
+
+                return nextChar;
+            }
+        };
+
+        abstract char getLabel(ButtonData buttonData);
+        abstract char setAutoChar(ButtonData buttonData, char nextChar);
+    }
+
+    // Every button (but the last) has a reference to the next button in the calculator.
+    // This reference is used to determine the alphabetic progression of the 'AutoChars'.
+    private CalculateButton nextButton = null;
+
+    // The buttons own state is stored in a 'state' object to facilitate easy saving and restoring.
+    private ButtonData buttonData;
+
+    // Data used to capture the state of this particular button such that it can be restored again later.
+    public static class ButtonData implements Serializable, JSONAble {
+        ValueType type = ValueType.INPUT_VAL;
+        char inputVal;                  // Value obtained from CoordinateInputDialog.
+        char autoChar = EMPTY_CHAR;     // Character obtained by automatically 'counting up' variable names.
+        char customChar;                // User defined character.
+
+        public ButtonData() { }
+
+        public ButtonData(final JSONObject json) {
+            type = ValueType.values()[json.optInt("type", 0)];
+            inputVal   = (char) json.optInt("inputVal",   ERROR_CHAR);
+            autoChar   = (char) json.optInt("autoChar",   ERROR_CHAR);
+            customChar = (char) json.optInt("customChar", ERROR_CHAR);
+        }
+
+        public JSONObject toJSON() throws JSONException {
+            final JSONObject returnValue = new JSONObject();
+
+            returnValue.put("type", type.ordinal());
+            returnValue.put("inputVal", inputVal);
+            returnValue.put("autoChar", autoChar);
+            returnValue.put("customChar", customChar);
+
+            return returnValue;
+        }
+    }
+
+    public static class ButtonDataFactory implements JSONAbleFactory {
+        @Override
+        public JSONAble fromJSON(final JSONObject json) {
+            return new ButtonData(json);
+        }
+    }
+
+    private class CoordDigitClickListener implements View.OnClickListener {
+        @Override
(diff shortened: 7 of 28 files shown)


SammysHP:
OK to test.

main/res/values/strings.xml
```diff
@@ -1025,6 +1029,7 @@
     <string name="waypoint_duplicate">Duplicate waypoint</string>
     <string name="waypoint_duplicated">Waypoint duplicated</string>
     <string name="waypoint_copy_of">Copy of</string>
+    <string name="waypointcalc_notes_prompt">Enter notes here...</string>
```
SammysHP: Please use a proper ellipsis character `…`.

main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -547,7 +547,7 @@ public void onSaveInstanceState(final Bundle savedInstanceState) {
     /**
      * Action bar spinner adapter. {@code null} for list types that don't allow switching (search results, ...).
      */
-    CacheListSpinnerAdapter mCacheListSpinnerAdapter;
+    CacheListSpinnerAdapter cacheListSpinnerAdapter;
```
SammysHP: While it is an improvement, it is not related to this pull request.

main/src/cgeo/geocaching/activity/AbstractListActivity.java
```diff
@@ -14,12 +14,12 @@ protected AbstractListActivity(final boolean keepScreenOn) {
         super(keepScreenOn);
     }
 
-    private ListView mListView;
+    private ListView listView;
```
SammysHP: While it is an improvement, it is not related to this pull request.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public CalcState(final JSONObject json) {
+        format = Settings.CoordInputFormatEnum.values()[json.optInt("format", 2)];
+        plainLat = json.optString("plainLat");
+        plainLon = json.optString("plainLon");
+        latHemisphere = (char) json.optInt("latHemisphere", ERROR_CHAR);
+        lonHemisphere = (char) json.optInt("lonHemisphere", ERROR_CHAR);
+
+        {
```
SammysHP: Is something missing here? A lot of groups without any control structure.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,135 @@
...
+            for (int i = 0; i < arrayLength; i++) {
+                freeVariables.add(new CalculatorVariable.VariableData(array.optJSONObject(i)));
+            }
+        }
+
+        notes = json.optString("notes");
+    }
+
+    public static CalcState fromJSON(final String json) throws JSONException {
+
```
SammysHP: Superfluous empty line.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,135 @@
...
+        CalcState rv = null;
+        if (json != null) {
+            rv = new CalcState(new JSONObject(json));
+        }
+
+        return rv;
```
SammysHP:
```java
if (json != null) {
    return new CalcState(new JSONObject(json));
}

return null;
```

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public JSONObject toJSON() throws JSONException {
+        final JSONObject rv = new JSONObject();
```
SammysHP: `rv` is not really descriptive.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,135 @@
...
+
+        rv.put("format", format.ordinal());
+        rv.put("plainLat", plainLat);
+        rv.put("plainLon", plainLon);
+        rv.put("latHemisphere", latHemisphere);
+        rv.put("lonHemisphere", lonHemisphere);
+
+        {
```
SammysHP: Again unnecessary blocks.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -83,17 +84,14 @@
...
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONException;
 
 public class DataStore {
 
     public static final String DB_FILE_NAME = "data";
     public static final String DB_FILE_NAME_BACKUP = "cgeo.sqlite";
     public static final String DB_FILE_CORRUPTED_EXTENSION = ".corrupted";
 
-    private DataStore() {
```
SammysHP: While it is an improvement, it is not related to this pull request. But IMHO this change can stay because you were working in this area.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -1990,6 +2016,7 @@ public Waypoint call(final Cursor cursor) {
 
     @NonNull
     private static Waypoint createWaypointFromDatabaseContent(final Cursor cursor) {
+
```
SammysHP: Superfluous empty line.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,266 @@
...
+        ButtonData() { }
+
+        public ButtonData(final JSONObject jason) {
+            type = ValueType.values()[jason.optInt("type", 0)];
+            inputVal   = (char) jason.optInt("inputVal",   ERROR_CHAR);
+            autoChar   = (char) jason.optInt("autoChar",   ERROR_CHAR);
+            customChar = (char) jason.optInt("customChar", ERROR_CHAR);
+        }
+
+        public JSONObject toJASON() throws JSONException {
```
SammysHP: Typo (`toJSON`).

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,266 @@
...
+            final JSONObject rv = new JSONObject();
```
SammysHP: `rv` not descriptive.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,266 @@
...
+            rv.put("inputVal", inputVal);
+            rv.put("autoChar", autoChar);
+            rv.put("customChar", customChar);
+
+            return rv;
+        }
+    }
+
+    private class CoordDigitClickListener implements View.OnClickListener {
+
```
SammysHP: Superfluous empty line.

main/src/cgeo/geocaching/ui/CalculatorVariable.java
```diff
@@ -0,0 +1,223 @@
...
+        this.cacheDirty = cacheDirty;
+    }
+
+    // Method to returned the cached values as a String
+    // This is used to display the result in the UI.
+    public String getCachedString() {
+        final String rv;
+
+        if (variableData.expression == null || variableData.expression.length() == 0) {
+            rv = "_";
```
SammysHP: Maybe move to constant.

And the `rv`, of course. ;)

main/src/cgeo/geocaching/ui/EditButton.java
```diff
@@ -0,0 +1,153 @@
...
+        addViews(context);
+    }
+
+    public EditButton(final Context context, final AttributeSet attrs) {
+        super(context, attrs);
+        addViews(context);
+    }
+
+    private void addViews(final Context context) {
+
```
SammysHP: We don't leave empty lines at the beginning or end of a class or method.

main/src/cgeo/geocaching/ui/EditNoteDialog.java
```diff
@@ -22,7 +22,7 @@
 
     public static final String ARGUMENT_INITIAL_NOTE = "initialNote";
 
-    private EditText mEditText;
+    private EditText editText;
```
SammysHP: While it is an improvement, it is not related to this pull request.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+    //                 All 'freeVariables' must have a lower-case name.
+
+    List<CalculatorVariable> equations;       // List of equations to be displayed in the calculator.
+    List<CalculatorVariable> freeVariables;   // List of freeVariables to be displayed in the calculator.
+
+    Spinner spinner;
+
+    private String inputLatHem;
+    private Button bLatHem;  // Latitude hemisphere (North or South)
+    private CalculateButton         bLatDeg_100,   bLatDeg_010,   bLatDeg_001,
```
SammysHP: As I suggested in the other pull request: Can you think of a better way to store these variables? They are already formatted as an array, so you could either try `CalculateButton[][]` or a custom class.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+                return;
+            }
+
+            if (gp != null) {
+                ((CoordinatesInputDialog.CoordinateUpdate) getActivity()).updateCoordinates(gp);
+            }
+
+            close();
+        }
+
```
SammysHP: Here is an empty line at the end of the class, but not between classes. ;)

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+    private class InputDoneListener implements View.OnClickListener {
+
+        @Override
+        public void onClick(final View v) {
+
+            // Save calculator state regardless of weather the coordinates are valid or not.
+            ((CoordinatesInputDialog.CalculateState) getActivity()).saveCalculatorState(saveState());
+            displayToast(R.string.info_calculator_state_saved, true);
+
+
```
SammysHP: Superfluous empty line.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+     */
+    public static CoordinatesCalculateDialog getInstance(final Geopoint gp, final CalcState calculationState) {
+
+        final Bundle args = new Bundle();
+
+        if (gp != null) {
+            args.putParcelable(GEOPOINT_ARG, gp);
+        }
+
+        final CoordinatesCalculateDialog cid = new CoordinatesCalculateDialog();
```
SammysHP: `cid` → `dialog`, `gp` → `geopoint`.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+                    break;
+            }
+
+            updateResult();
+        }
+    }
+
+    /**
+     *
+     * @param gp: Geopoint representing the coordinated from the CoordinateInputDialog
```
SammysHP: Typo ("coordinated").

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+            buttonDone.setVisibility(View.GONE);
+        } else {
+            buttonDone.setOnClickListener(inputdone);
+        }
+
+        loadCalcState();
+        resortEquations();
+        updateResult();
+
+        /**
```
SammysHP: Comment not necessary (or if you think that it is necessary, just use a small `//`).

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+    // This method has largely been lifted from the CoordinateInputDialog and is used to validate
+    // coordinate specifications prior to assigning them to the Geopoint.
+    private boolean areCurrentCoordinatesValid(final boolean signalError) {
+        // then convert text to geopoint
+        final Geopoint current;
+        final String lat = getLatResult();
+        final String lon = getLonResult();
+
+        final String[] latTokens = lat.split("[ °'\"\\.]+");
+        final String[] lonTokens = lon.split("[ °'\"\\.]+");
```
SammysHP: Both are the same?

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+            }
+            if (current.isValid()) {
+                gp = current;
+                return true;
+            }
+        } catch (final Geopoint.ParseException ignored) {
+            // Signaled and returned below
+        }
+
+        if (signalError) {
```
SammysHP: I don't understand the purpose of `signalError` and why it is checked so late.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+        return false;
+    }
+
+    private void setCalcState(final CalcState calcState) {
+        this.calcState = calcState;
+    }
+
+    private void loadCalcState() {
+        if (calcState != null) {
+
```
SammysHP: Superfluous empty line.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+        return null;
+    }
+
+    // Resort the equations into the order in which they first appear in the 'buttons' or 'plain-text' fields as appropriate.
+    // This method is also responsible for adding and removing equations from the associated list as appropriate.
+    private void resortEquations() {
+        final List<CalculatorVariable> resortedEquations = new ArrayList<>();
+
+        String chars = "";
```
SammysHP: Which chars?

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1096 @@
...
+                    chars = chars.concat(lon);
+                }
+            }
+        } else {
+            for (final EditButton b : coordButtons) {
+                chars = chars.concat(String.valueOf(b.getLabel()));
+            }
+        }
+
+        for (int i = 0; i<chars.length(); i++) {
```
SammysHP: Probably the missing spaces between `1` and `chars.length()`.

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,122 @@
+package cgeo.geocaching.utils;
+
+public final class CalculationUtils {
+    private CalculationUtils() {
+        // utility class
+    }
+    // This simple evaluation algorithm was created by user 'Boann' and released to the public domain on Stack Overflow:
```
SammysHP: Please use `/** … */`.

SammysHP:
Please fix the `Position literals first in String comparisons` warnings from Codacy as well.

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -96,13 +96,13 @@ boolean eat(final int charToEat) {
                     }
                     final String func = str.substring(startPos, this.pos);
                     x = parseFactor();
-                    if (func.equals("sqrt")) {
```
SammysHP: Do we use Java 6 or 7? In case of 7 this could be replaced by a single switch statement.

S-Bartfast: I've changed it over to use a switch and it seems to compile fine.  It's not relay possible to check if it works the same though as these functions can never be reached in the current implementation.

main/res/values/strings.xml
```diff
@@ -252,6 +253,8 @@
     <string name="info_stored_image">New image saved to:</string>
     <string name="info_storing_static_maps">Trying to store static maps</string>
     <string name="info_cache_saved">The cache has been stored locally</string>
+    <string name="info_calculator_state_saved">Current state of the calculator has been saved to waypoint.</string>
```
S-Bartfast: It's used when one clicks the "tick" (OK) button at the top of the calculator.  It's useful to see that the state has been saved when there's an error in the formulation and clicking the button doesn't appear to do anything else otherwise.

SammysHP: IMHO there should be only a message in case of an error and no explicit save button.

main/src/cgeo/geocaching/models/CalcState.java
SammysHP: This was requested by @samueltardieu in https://github.com/cgeo/cgeo/pull/6637#discussion_r128224281.

main/src/cgeo/geocaching/models/CalcState.java
SammysHP: This was requested by @samueltardieu in https://github.com/cgeo/cgeo/pull/6637#discussion_r128224281.

main/res/layout-land/coordinatescalculate_dialog.xml
```diff
@@ -0,0 +1,96 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:fillViewport="true">
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:orientation="vertical">
+
+
```
S-Bartfast: Done.

main/res/layout/coordinates_calculate.xml
```diff
@@ -0,0 +1,204 @@
...
+            style="@style/button_digit" />
+
+        <TextView
+            android:id="@+id/LonLastUnitsChar"
+            android:text='"'
+            android:textSize="22sp"
+            android:layout_marginRight="1dp"
+            tools:ignore="HardcodedText" />
+    </GridLayout>
+</merge>
```
S-Bartfast: Done.

main/res/values/dimens.xml
```diff
@@ -9,4 +9,7 @@
     <dimen name="app_defaultsize_w">632.0dip</dimen>
     <dimen name="app_defaultsize_h">598.0dip</dimen>
 
+
```
S-Bartfast: Done.

main/res/values/strings.xml
S-Bartfast: Done.

main/src/cgeo/geocaching/CacheListActivity.java
S-Bartfast: I'll revert it.

main/src/cgeo/geocaching/activity/AbstractListActivity.java
S-Bartfast: I'll revert it.

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: It's so I can reuse the "array" variable.

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: Done.

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: I usually do my best to only have one return in a method.

If anything I would return early in event of an error, in which case I would structure the code as:

```
if (json == null) {
    return null;
}

return new CalcState(new JSONObject(json));
```
But I would prefer to keep something as short as this to one return statement.

One reason I prefer to use a variable and have a single return is it makes it easy to put a break on the statement and inspect the result before it's returned.

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: It just means ReturnValue and I thought it was a common practice but after looking it up it appears not to be a common as I thought, so I'll be happy to rename them.  Would I be allowed t use `ofTheJedi` instead? :) :)

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: Same reason.

main/src/cgeo/geocaching/storage/DataStore.java
S-Bartfast: The modification was made as my changes to 'WAYPOINT_COLUMNS' (line 163) and 'dbVersion' (line 173) were raising Codacy warnings because they were coming after the constructor.

main/src/cgeo/geocaching/storage/DataStore.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/CalculateButton.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/CalculateButton.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/EditButton.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/EditNoteDialog.java
S-Bartfast: Reverted.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Now says coordinates.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Changed to `ccd`.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Removed.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: They are both using the same delimiters but the tokens themselves are quite different.  I did actually consider creating a temporary 'delimiters' string and passing that string into the functions instead of using literals but decided against it as I thought typing in the literals was clearer, but I'll do it now.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: That's how it was done in`CoordinatesInputDialog` so I just kind of copied it.

Anyway, it looks like it's a flag to indicate whether the error message should be displayed as a toast or not.  As it is, in `CoordinatesCalculateDialog` it appears to only be called with a 'true' parameter so I'll just get rid of the variable.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: They are the characters that appear on the buttons.  I'll rename it to 'equationNames'.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: "Probably the missing spaces"??

I would normally write this as: `for (int i=0 ; i<chars.length() ; i++)` however Codacy wants spaces around the '='.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: I'll need to have a bit more of a look at it and will see what I can do tomorrow night.  'Tis getting late here now.

main/src/cgeo/geocaching/models/CalcState.java
SammysHP: I know that it means return value, but it would be better to know what it contains when it will be returned.

main/src/cgeo/geocaching/models/CalcState.java
SammysHP: Your proposal is even better. Early returns make it easier to understand: Here it will return, (almost) nothing will be executed after this line.

So please change this to

```java
if (json == null) {
    return null;
}

return new CalcState(new JSONObject(json));
```

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1089 @@
...
+                    equationNames = equationNames.concat(lon);
+                }
+            }
+        } else {
+            for (final EditButton b : coordButtons) {
+                equationNames = equationNames.concat(String.valueOf(b.getLabel()));
+            }
+        }
+
+        for (int i = 0; i<equationNames.length(); i++) {
```
SammysHP: Please add spaces around `<`.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1089 @@
...
+    private void resortFreeVariables() {
+        final List<CalculatorVariable> resortedVariables = new ArrayList<>();
+
+        String chars = "";
+
+        for (final CalculatorVariable equ : equations) {
+            chars = chars.concat(equ.getExpression());
+        }
+
+        for (int i = 0; i<chars.length(); i++) {
```
SammysHP: Same here, please add spaces around `<`.

SammysHP:
IMHO the remaining Codacy warnings can be ignored. Except from some minor issues the code looks fine and I'll approve it in my review once they are fixed. The biggest open issue are the `CalculateButton` variables with the numbers in their name. An array or extra class might be better.

I don't know why the test fails because the fix from master should be included. Does anyone know why it fails?

I'd like to get at least one more review before merging this.

@S-Bartfast 
When you are finished with the open issues you might want to start with a short manual for the new feature which will be added to the FAQ. You can edit the FAQ [here](https://github.com/cgeo/cgeo.github.io/blob/master/_data/faq.yml) and submit a PR afterwards. You made some fantastic videos that should be linked as well.

main/res/values/strings.xml
S-Bartfast: Hmm, yeah.  I see what I said above was a bit ambiguous.

To be clear, what we're talking about is the "Done" button in the top left corner:
done

Normally, when the user clicks the "Done" button, the coordinates are parsed and, if the result is valid, the result is used as the input for the way point and the calculator is closed.  If however the result is not valid a toast is displayed saying "c:geo can't parse latitude or longitude." and the calculator is left open so the user can correct the malformed coordinates and try again.  If the user doesn't want to repair the coordinates they can always click the "Cancel" cross to discard the malformed calculation and return to the waypoint with the values it had before —and this is exactly the same behavior as the CoordinateInputDialog itself uses.

The thing to workout is when should the calculators state itself be saved.  As it is now (which I think is the best way to have it) the calculators state is saved whenever the "Done" button is pressed.  The issue is that if the "Done" button is clicked while the coordinates are invalid the user is informed about the invalid coordinates but there is no indication about the calculators state being saved.

Another option would be to only save the calculators state if the coordinate's format is valid, however it is anticipated that people will often want to store partially computed waypoints and requiring that the coordinates be fully formatted before saving would be unnecessarily burdensome.

The final option would be to save the state every time either the "Done" tick OR the "Cancel" cross is clicked.  This would obviously mean though that a user would not be able cancel their erroneous scratchings and revert to their previously saved calculators state should they be uncertain about things, which doesn't seem like a good option.

There would also be options of allowing the back arrow to save the state and the "Cancel" button not to (or vice versa) but neither of these seem particularly natural and would not be intuitive to the user, and adding a dedicated "Save State" button or menu item seems like even more of a kludge.

After considering all these options (and even implementing most of them) I eventually settled on the "Always save when 'Done' is pressed" methodology and have elected to inform the user when this happens.  As it is the message is shown even when the coordinate format is valid but I actually think it's fine like that as the message is immediately followed by the "Be sure to save waypoint if you want to keep the calculator state." message once the 'Waypoint' screen is reached, which actually does look good.

If you can think of an approach that you believe would be better I'm open to suggestions but I would recommend at least having a go with the current implementation as it does seem pretty good to me.

Lineflyer: >  "Always save when 'Done' is pressed" methodology

I agree, that this sound like the most meaningful approach.
In this case I would change the string to something like:
"The calculation result was invalid, no coordinates could be set but the current state of your calculation has been saved to the waypoint"

SammysHP: > "The calculation result was invalid, no coordinates could be set but the current state of your calculation has been saved to the waypoint"

Too long for a toast. I don't like positive feedback (in terms of UI).

> "Always save when 'Done' is pressed"

👍, but please without a toast. Keep toast for things that aren't expected.

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: All right, it's done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Okay, I've converted them to arrays and while it does look neater I can't help but think it's less clear and more error prone.  But there you have it.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Done.

main/res/values/strings.xml
S-Bartfast: > Keep toasts for things that aren't expected.

But that's my point, it _is_ unexpected!

> "The calculation result was invalid, no coordinates could be set but the current state of your calculation has been saved to the waypoint"

At the moment the messages are displayed as two separate toasts.
When the coordinates _are_ valid it shows:
> Current state of the calculator has been saved to waypoint.

> Be sure to save waypoint if you want to keep the calculator state.

And when the coordinates _aren't_ valid it shows:
> Current state of the calculator has been saved to waypoint.

> c:geo can\'t parse latitude or longitude.

And honestly, it's perfectly fine.

I acknowledge it's somewhat in-elegant to have two toasts display one after the other but it actually plays out perfectly well.  As I said before, I've put a lot of thought into this and have tried several approaches and this is the best I've been able to find and I recommend giving it a go.  If we need to change it we'll change it but it actually works perfectly fine as it is.

Lineflyer: @SammysHP  @S-Bartfast 
I understand all pro and con of this topic. I think I need to see this "in action" and then suggest a change if needed. 
So my proposal would be to leave it as it is for now and review after merge to `master` when testing the implementation on a real day usage.

SammysHP: >When the coordinates are valid it shows:
>
>>    ~Current state of the calculator has been saved to waypoint.~
>
>>    ~Be sure to save waypoint if you want to keep the calculator state.~
>
>And when the coordinates aren't valid it shows:
>
>>    ~Current state of the calculator has been saved to waypoint.~
>
>>    c:geo can't parse latitude or longitude.

main/src/cgeo/geocaching/ui/EditButton.java
```diff
@@ -0,0 +1,152 @@
...
+import cgeo.geocaching.R;
+
+import static android.text.InputType.TYPE_TEXT_FLAG_CAP_CHARACTERS;
+import static android.text.InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+
+/**
+ * This class allows for a user to change the text on a button with a long-click.
+ */
+
```
pstorch: remove blank line

main/src/cgeo/geocaching/ui/EditButton.java
```diff
@@ -0,0 +1,152 @@
...
+    Button butt;    // The actual button used for the most part.
+
+    private class CoordDigitLongClickListener implements View.OnLongClickListener {
+
+        // This implementation was obtained from 'Adithya' via the stack overflow question 'Long-press on Button to Change Text':
+        // https://stackoverflow.com/questions/44858720/long-press-on-button-to-change-text/44859328#44859328
+
+        @Override
+        public boolean onLongClick(final View view) {
+
```
pstorch: remove blank line

main/src/cgeo/geocaching/ui/EditButton.java
```diff
@@ -0,0 +1,152 @@
...
+                    getContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+
+            butt.setVisibility(View.INVISIBLE);
+            edit.setVisibility(View.VISIBLE);
+
+            if (edit.requestFocus()) {
+                imm.showSoftInput(edit, InputMethodManager.SHOW_IMPLICIT);
+            }
+
+            edit.addTextChangedListener(new TextWatcher() {
```
pstorch: Looks like a candidate to extract from anonymous to inner class.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+
+    static final String SYMBOL_DEG = "°";
+    static final String SYMBOL_MIN = "'";
+    static final String SYMBOL_SEC = "\"";
+    static final String SYMBOL_POINT = ".";
+
+    private Geopoint gp;
+    private CalcState savedState;
+
+    // Through out this implementation:
```
pstorch: As this is a comment explaining the implementation of the whole class, maybe it would be better to move it to the class javadoc.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+    private Settings.CoordInputFormatEnum currentFormat = null;
+    private List<CalculateButton> latButtons;
+    private List<CalculateButton> lonButtons;
+    private List<CalculateButton> coordButtons;
+    private List<View> minButtons;
+    private List<View> secButtons;
+    private List<CalculateButton> pointLowButtons;
+    private List<TextView> lastUnits;
+
+
```
pstorch: superfluous blank line

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+        }
+    }
+
+    private class CoordinateFormatListener implements AdapterView.OnItemSelectedListener {
+        // One Shot marker
+        private boolean shot = false;
+
+        @Override
+        public void onItemSelected(final AdapterView<?> parent, final View view, final int pos, final long id) {
+
```
pstorch: no blank line at the beginning of a method

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+            final byte[] bytes = savedInstanceState.getByteArray("calc_state");
+            if (bytes != null) {
+                setSavedState((CalcState) SerializationUtils.deserialize(bytes));
+            }
+        }
+    }
+
+    @Override
+    public View onCreateView(final LayoutInflater inflater, final ViewGroup container, final Bundle savedInstanceState) {
```
pstorch: This is reported as a long method with a high complexity. Can you find some structure which is worth extracting to its own method?

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+        final Dialog dialog = getDialog();
+        final boolean noTitle = dialog.requestWindowFeature(Window.FEATURE_NO_TITLE);
+
+        final View v = inflater.inflate(R.layout.coordinatescalculate_dialog, container, false);
+        final InputDoneListener inputdone = new InputDoneListener();
+
+        if (!noTitle) {
+            dialog.setTitle(R.string.cache_coordinates);
+        } else {
```
pstorch: maybe the else block is a candidate for extraction

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+
+        minButtons = Arrays.asList(tLatDegChar, bLatMin[1], bLatMin[0],
+                                   tLonDegChar, bLonMin[1], bLonMin[0]);
+        secButtons = Arrays.asList(tLatMinChar, bLatSec[1], bLatSec[0],
+                                   tLonMinChar, bLonSec[1], bLonSec[0]);
+        pointLowButtons = Arrays.asList(bLatPnt[4], bLatPnt[3],
+                                        bLonPnt[4], bLonPnt[3]);
+        lastUnits = Arrays.asList(tLatLastUnits, tLonLastUnits);
+
+        bLatDeg[2].setNextButton(bLatDeg[1]);
```
pstorch: There are similar things done twice: once for the lat buttons and another time for the lon buttons. Maybe this can be extracted into a method which is called twice.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -0,0 +1,1101 @@
...
+        resortEquations();
+        updateResult();
+
+        ePlainLat.addTextChangedListener(new PlainWarcher());
+        ePlainLon.addTextChangedListener(new PlainWarcher());
+
+        bLatHem.setOnClickListener(new HemisphereClickListener());
+        bLonHem.setOnClickListener(new HemisphereClickListener());
+
+        for (final EditButton button : coordButtons) {
```
pstorch: candidate for extraction

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            /**
+             * Grammar:
+             **
+             * expression = term | expression `+` term | expression `-` term
+             * term = factor | term `*` factor | term `/` factor
+             * factor = `+` factor | `-` factor | `(` expression `)`
+             *        | number | functionName factor | factor `^` factor
+             **/
```
pstorch: only one `*`

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+                final double x = parseExpression();
+                if (pos < str.length()) {
+                    throw new RuntimeException("Unexpected: " + (char) ch);
+                }
+                return x;
+            }
+
+            /**
+             * Grammar:
+             **
```
pstorch: only one `*`

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+             * expression = term | expression `+` term | expression `-` term
+             * term = factor | term `*` factor | term `/` factor
+             * factor = `+` factor | `-` factor | `(` expression `)`
+             *        | number | functionName factor | factor `^` factor
+             **/
+
```
pstorch: remove blank line

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * This simple evaluation algorithm was created by user 'Boann' and released to the public domain on Stack Overflow:
+     * https://stackoverflow.com/questions/3422673/evaluating-a-math-expression-given-in-string-form
+     *
+     * I added the '%' operator myself.
+     * Note that the 'sqrt', 'cos', 'sin' and 'tan' functions don't work in this implementation as
+     * expected as the individual letters are automatically replaced with their corresponding
+     * expressions as they are interpreted as being individual variables.
+     **/
```
pstorch: only one `*`

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+    public static double eval(final String str) {
+        return new Object() {
```
pstorch: This looks like a strange concept too me. If you refactor this anonymous class to an inner class with a proper name, then I think it's more readable and you takle the npath complexity reported by codacy.

pstorch: Or just refactor this class from a utility class `CalculationUtil` to an instantiate-able class `Calculator`, put the expression `str` into the constructor and store it in a field. Then have `eval()` without argument and extract all the methods from the anonymous class into the outer class `Calculator`.

main/src/cgeo/geocaching/utils/CalculationUtils.java
```diff
@@ -0,0 +1,128 @@
...
+             **
+             * expression = term | expression `+` term | expression `-` term
+             * term = factor | term `*` factor | term `/` factor
+             * factor = `+` factor | `-` factor | `(` expression `)`
+             *        | number | functionName factor | factor `^` factor
+             **/
+
+            double parseExpression() {
+                double x = parseTerm();
+                for (;;) {
```
pstorch: this expression makes me think, don't know if `while (true)` is better.

pstorch:
If I click on the buttons unordered, the variables get mixed up.

pstorch:
I think we should warn or prevent the user from changing the coordinate format. All variables get lost!

pstorch:
If I change one variable to a static value, all variables after the changed one get reassigned and emptied.

I think this calculator is a good start, but there are some UX pitfalls which can causes some frustration among the users. Don't know if we should sort them out first or takle them once the PR is merged.

S-Bartfast:
I'm sure what's happening with the buttons getting re-ordered, that definitely sounds wrong.  I'll have to look at it tonight.

This was working fine before the last commit but when I changed the implementation to use arrays of buttons rather than individually named buttons it became much harder to keep track of which buttons referred to which.  As it was I made many mistakes in the change over and thought I'd sorted them all out, but clearly not.  I'll look at it again to night and workout what I've done.

S-Bartfast:
Closing the CalculationDialog with the "X" is meant to clear the calculator's state.
Closing the CalculationDialog and the EditWaypointDialog with X is meant to discard that session's state and when the calculator is opened again it should be reload with whatever state was saved previously.  Is that not happening?

Again I'll have to check it again tonight.

S-Bartfast:
The reason the variables are cleared when the format is changed is that attempting to carry over variables from previous formats becomes troublesome as different formats have differing numbers of buttons.  It would be possible for a "format change" to do a "best effort" at retaining previous data but to be honest it didn't seem worth it.  If anything being able to 'clear' the calculator with a format change can be kind of useful.

S-Bartfast:
The general philosophy of the form was to work from the top down.

The idea is that the format is chosen first, then the Equations are nominated next, then they are filled in (possibly with the addition of 'free variables'), with each stage only affecting the ones after.  Some effort has been made to allow a little back and forth between the different parts but as a general philosophy the calculator is intended to be used in order.

I'm aware there is the possibility of data loss when 'higher up' modifications are made but trying to retain 'lower down' information when these modifications are made quickly leads to conflicting states and seemingly unintuitive results.  If there are any particular scenarios you think should be addressed please do raise them though and I'll certainly do my best to incorporate them into a design.

pstorch:
> The general philosophy of the form was to work from the top down.

I thought so, but users often don't use the way which was intended :wink:. 
As I wrote, we don't need to tackle all the (possible) issues from the start. We can improve when it is in the master. This is a complex feature, that's why nobody tried to implement it till now. :+1:

main/src/cgeo/geocaching/models/CalcState.java
S-Bartfast: Okay, after yet another evening of hacking I've managed to find a way to implement what I understand to be a "factory" paradigm to get around the blessid 'constructor in the loop' issue.  Hopefully this attempt will make the grade ;).

This is my first attempt at using a technique like this so I'll be more than happy to neaten this up more if anyone can suggest anything that might help ;).

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.ui.CalculateButton;
+import cgeo.geocaching.ui.CalculatorVariable;
+
+/**
+ * This class is designed to capture the current state of the coordinate calculator such that it can be preserver for latter use.
+ **
```
S-Bartfast: Done.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,133 @@
...
+ * All the relevant information is in a serializable form such that it can be stored as a bundle in waypoint's 'ContentValues'.
+ **/
```
S-Bartfast: Done.

main/src/cgeo/geocaching/models/CalcState.java
```diff
@@ -0,0 +1,133 @@
...
+    public JSONObject toJSON() throws JSONException {
+        final JSONObject returnValue = new JSONObject();
+
+        returnValue.put("format", format.ordinal());
+        returnValue.put("plainLat", plainLat);
+        returnValue.put("plainLon", plainLon);
+        returnValue.put("latHemisphere", latHemisphere);
+        returnValue.put("lonHemisphere", lonHemisphere);
+
+        {
```
S-Bartfast: Hopefully that looks alright.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -1420,6 +1433,18 @@ private static void putCoords(final ContentValues values, final Geopoint coords)
...
+        if (calcState != null) {
+            try {
+                values.put("calc_state", calcState.toJSON().toString());
+            } catch (final JSONException e) {
+                Log.e("Unable to write calculator state information", e);
+            }
+        }
+    }
+
+
```
S-Bartfast: Curious, not sure how that happened.
But fixed now.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+import java.io.Serializable;
+
+import cgeo.geocaching.R;
+
+import static cgeo.geocaching.models.CalcState.ERROR_CHAR;
+
+/**
+ * This class derives from EditButton and handles all the attribures that are unique to this particular application of button.
+ **
```
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+ * In particular this button stores three values:
+ * 1. The value obtained from cache coordinates that were at the time the calculater war crreated
+ * 2. An 'AutoChar' which is the next alphabetic character used when the user clicks on the button
+ * 3. A custom character which is a specific value assigned by user by perforning a long-click.
+ *
+ * The button displayed one of these values as appropriate as the text on the button.
+ **/
```
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+
```
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    // The 'Label' is the 'name' that is to be displayed on the button
+    public char getLabel() {
+        final char returnValue;
+        switch (getType()) {
+            case INPUT_VAL: returnValue = buttonData.inputVal;   break;
+            case AUTO_CHAR: returnValue = buttonData.autoChar;   break;
+            case CUSTOM:    returnValue = buttonData.customChar; break;
+            default:        returnValue = '*'; // Should never happen.
```
S-Bartfast: That does sound good actually, but it's 2am and I just can't face it now.  Perhaps later.

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    public void setType(final ValueType type) {
+        buttonData.type = type;
+        setBackgroundColour();
+    }
+
+    // Buttons displaying a custom value are given a 'steel-like' appearance so as to distinguish them from regular buttons.
+    private  void setBackgroundColour() {
+        if (buttonData.type == ValueType.CUSTOM) {
```
S-Bartfast: I like it, perhaps later.

main/src/cgeo/geocaching/ui/CalculatorVariable.java
```diff
@@ -0,0 +1,223 @@
...
+import cgeo.geocaching.R;
+
+import static cgeo.geocaching.models.CalcState.ERROR_CHAR;
+import static cgeo.geocaching.models.CalcState.ERROR_STRING;
+
+/**
+ * Class used to display a variable with an equation, such as:
+ *  X = a^2 = b^2
+ */
+
```
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/EditButton.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/EditButton.java
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/EditButton.java
S-Bartfast: That probably would be neater.  I'll look into it.
It's now down at line 133 by the way ;).

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Perhaps.  The reason it's there though is that it's right above the two variables it's discussing.

I was thinking people may well come across those variables else ware in the code and wonder what exactly they were meant to represent so placing the comment directly above them would make it easy to find.  I wouldn't even know how to find the "class javadoc" myself but perhaps someone who knows more about what they are doing would think to look there first. I wouldn't know ;)

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: I can't find that one.  I probably got rid of it my self.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Already gone.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: I'm not sure where the complexity comes from.  It's long but really is very simple.
I could break it up into separate sections but it wouldn't really do anything other than run one section after the other.  There's basically just nothing much going on in there.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: I'll think about it.

I just copied this section from CoordinateInputDialog by the way.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: It would probably work well to have `setNextButton(...)` return its newly assigned button so one could chain them together, as in: `but1.setNext(but2).setNext(but3).setNext(but4)...`
I'll think on it some more.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: This is looking different now.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Done.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Ooow, my mind just exploded!

As I've said before, this is actually someone else's algorithm and I don't like poking it too much but I'll do my best to get my head around what you just said.

Someone else is more than welcome to take ownership of this algorithm my the way.  I'm kind of not really wanting to be all that responsible for it.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Done.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Done.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Done.

main/src/cgeo/geocaching/utils/CalculationUtils.java
S-Bartfast: Nor do I, hehe ;)

S-Bartfast:
Alright, just gave this another whack.

There was a problem I must have introduced at some stage where by the buttons were refreshing at an inopportune time resulting in the equations being cleared when they most certainly shouldn't have been.  I believe I've now sorted this out again by implementing a couple of callbacks.  Let us know if the problem persists.

As for the "variables getting mixed up" are you able to explain this a little more pstorch?  I don't think I'm seeing the same thing you are.

main/src/cgeo/geocaching/utils/CalculationUtils.java
pstorch: Then leave it as it is for now. We can refactor it later on (hope later will not be never :wink: ). With some Unittests to back the implementation.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
pstorch: But in `CoordinateInputDialog` it's not a problem, because the overall complexity is lower. And I only wanted to hint possible refactoring in case you want to address the complexity of this method.

pstorch:
Hey, only 11 codacy issues left over :v:

pstorch:
Regarding "variable mixup":

Assume the following situation. I've configured 3 variables with formular:
screenshot_20170725-210524

Then I suddenly see I forgot one variable (e.g. the first digit of the north minutes), then this button becomes the new `A` and all other buttons get a newly assigned variable, but keep the old formula which was meant for the other position:

screenshot_20170725-210550

To correct it I have to copy&paste all formulas to their new variable.

Yes, I know, this dialog was meant to be used top to bottom and from left to right. But users can shoot them self in the foot if they don't know/follow this rule. Let's see the others think about this situation. I'm also fine with merging first, try it out on master and the c:geo nightlies and if necessary improve it

Lineflyer:
@pstorch 
The described mixup is surely a thing which I would also have recognizing while testing it acting as a random user. I agree this needs to be made failsafe in one or the other way.

But as I can see, the effort of cleaning the code structure itself is quite a bunch of work, I also think that this might be tackled in a later phase after merge. Lets just remind the problem for now (unless @S-Bartfast has a fix right away).

S-Bartfast:
I certainly can fix it right a way but the way it is behaving is in fact the way I intended it to behave.

The order in which the equations are shown is meant to reference the order in which they appear in the coordinate listing itself.  The algorithm basically starts at the left of the first row and adds every 'equation' it finds as it scans along, adding them to the first column as it goes.  Once it finishes the first row is then scans the second.  It was implemented this way to optimize the common case of `037° AB.CDE'` and `122° EG.HIJ'`.  In this case the user will end up with two columns of equations, the left being for the Latitude values, the right for Longitude.  This is just a little odd in that the two horizontal rows of buttons end up being represented by two vertical roes of equations but really there was no choice so far as I could see.  It works well when there are 6 or more equations but if only 4 of fewer equations are used it would look better if organised differently.  It would be an option to re-arrange the ordering once a 5th Equation is added but I thing the consistency of always having the equations appear in the same order is more important.

I tried a few different ordering ideas before settling on this one, the first and most basic being simply displaying the equations in the order in which the buttons were pressed (i.e. always adding 'new' equations to the end) and also tried disabling the 'auto-increment' feature of buttons that already exist to give a more 'static' experience to the user but ultimately decided against these.  The 'long-press' functionality was ultimately implemented to cater for this.  As it is the "auto-increment" mode is set as the 'default' mode but if people would prefer it the other way I could change it such that "manually-specifying" is the default and move the auto-increment to the long-click behavior.

Another thing I did consider early on but also rejected was to 'rename' equations rather than move them.  The idea of this was to have the 'names' of the equations swap about while keeping there expressions static, (which I think is what you are suggesting would be preferable).  My thought was that given the instructions to a cache are usually fairly explicit as to what expression should be associated with each letter it would be more important to keep the same expressions associated with the same letters than to keep them in the same location, but if people want it the other way I'm sure I could change it.

Finally, one thing that is regrettable though is that when equations are removed their associated expressions are also lost, and when re-added the expression fields do indeed appear empty.  This could be problematic if a user had entered 10 expressions for the letters A-J then for some reason manually specified say an "X" in the very first , top corner.  I can't imagine how someone would accidentally do this but if they were to it would result in most of the other buttons becoming 'Z's (with blank expressions) and once the "X" was clicked again to undo it the other buttons would indeed revert to their previous lettering however their expressions would have been lost.  This clearly would be regrettable and I was considering creating a 'variable bank' such that variables could be recovered in such an event but didn't end up implementing the feature.  I could of course set about implementing it though and it actually shouldn't be as hard as one might think.

S-Bartfast:
Just so I understand, if you could pick anything you wanted, what would you like to have happen when you click the "10's Deg" button in the example above?

S-Bartfast:
Okay, I've now implemented the 'Variable Bank'.

As it is now the 'equations' are still cleared when the format of the coordinates is changed but the 'variable bank' is not.  This means that although the coordinate fields are cleared it is very easy to re-populate the equations again.  This is probably a good compromise between resetting the calculator on a format change and allowing the state to persist.

While this 'variable bank' should persist through screen rotations it is not my intention to preserve the 'variable bank' between sessions.

pstorch:
Had to fiddle around some time until I saw the effect of the 'variable bank'. I think we should think about the usability when more users start using it in the nightlies.
There are some new codacy issues. I think at least the one missing final modifier and the two braces should be fixed.
Then a final squash and I'm fine with merging.

Lineflyer:
> Then a final squash and I'm fine with merging.

Sound nice. Can't wait for testing it.

S-Bartfast:
> I think we should think about the usability when more users start using it in the nightlies.

Huh, it often is surprising how something that seems completely straight forward to the developer can be entirely unintuitive to a user.  I guess this is one of those things.  In what ways is the usability lacking?  I actually can't imagine any other way someone would want it.

I'll fix the new Codacy warnings and review some of the other suggestions you made tonight.

main/src/cgeo/geocaching/ui/CalculateButton.java
S-Bartfast: I think I've done it.

I've never actually see that before, but what I've done seems to be working.
It's a neat trick 👍

main/src/cgeo/geocaching/ui/CalculateButton.java
S-Bartfast: I ended up changing this to a switch rather than overriding methods in the enum as having both the 'INPUT_CHAR' and 'AUTO_CHAR' cases run the same code was a little clunky.  As a switch though it's quite neat :)

main/src/cgeo/geocaching/ui/CalculateButton.java
```diff
@@ -0,0 +1,265 @@
...
+    public void setAutoChar(final char currentChar) {
+        char nextChar = currentChar;
+
+        if (nextChar < 'A') {
+            nextChar = 'A';
+        } else if (nextChar > 'Z') {
+            nextChar = 'Z';
+        }
+
+        switch (getType()) {
```
S-Bartfast: Done.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Think I'll leave it here for now, but happy to move it if others think the class javadoc would be a better place for it.

S-Bartfast:
Watch out guys.  Re-base it on it's way :) :)

SammysHP:
Looks good to me. Codacy should be happy (there is no report for this commit, but I checked the two open issues and there are fixed). There are some open issues (like the feedback or resetting of the calculator), but these can be addressed later.

Btw: I still don't understand what our CI server does. @Bananeweizen Can you check it please?

S-Bartfast:
Sorry, what are the open issues?

UniQP:
Independent of @SammysHP comment, Android Studio still report some issues with the modified/new files (Right click in the editor→ Analyze → Inspect Code).

@S-Bartfast Do you have Android Studio available, or should I add them here?

SammysHP:
There are new problems in CalculateButton.java reported by Codacy.

pstorch:
@S-Bartfast now the codacy report has appeared here https://www.codacy.com/app/cgeo/cgeo/pullRequest?prid=791105&bid=4967060. The number of issues have slightly increased with your last refactoring. Mainly it's about the reassignment of the parameter `nextChar` in the `CalculateButton` class and a line break before a curly brace at line 37. 
I think these are easy to fix.

S-Bartfast:
Okay, have fixed the additional Codacy warnings.

I have also done a brief run through of the Android Studio warnings.  I've run the tool before but found the output so voluminous that I wasn't sure where to begin.  As it is now I've combed through it all and did my best to address the issues that seemed to be related to the work I've done –do let me know if there are any others worth addressing that I may have missed though.  Note that I've intentionally left the "Perhaps 'xxx' could be replaced by a local variable." warnings as they are because I wanted to leave the assignment there with it's mates ;)

And finally I'm about to squish these last couple of commits, so watch-out... another re-base is on it's way (: (:

pstorch:
retest this please

pstorch:
Let's wait for a final test from CI. The experimental build failed, but the other one is still running. Then I'm fine with merging.

pstorch:
@S-Bartfast thank you for your contribution. Was a a lot of work.

S-Bartfast:
No worries, it was a pleasure.

And I'm pleased to serve the community that has made GeoCaching so accessible to me in the first place  👍

Natureshadow:
>Merged #6645.

Yay! I can't await trying it!

Issue #6655, opened 2017-07-29
Variable field vanishing on certain input


Lineflyer:
Just built a new nightly to take a first look at the waypoint calculator. I had no time to test in details, so just a short try where I either found a bug or still have a misunderstanding of the working mode.

##### Detailed steps causing the problem:
- Start with a new calculation
- Assign A to F to the last lat and lon digits
- Enter formulas for A to F where A=a , B=b, C=c-b

##### Actual behavior after performing these steps:
When entering B=b the variable input field for b is shown, when entering C=c-b a new variable field for c is shown but the field for b is gone

##### Expected behavior after performing these steps:
Variable input b should stay

##### Version of c:geo used:
2017.07.29-NB1

##### Is the problem reproducible for you?
Yes

##### System information:
Samsung Galaxy Tab S2
Android 7.0

##### Other comments and remarks:
I added a new label "WP Calc" to get a better overview for issues regarding the new WP calculator.

![screenshot_20170729-112834](https://user-images.githubusercontent.com/949669/28743801-fea15938-7452-11e7-949f-a1ac9f890cb1.png)

S-Bartfast:
That is definitely wrong!
I did some refactoring while implementing the 'variable bank' and suspect I may well have made an mistake during this refactoring.

I'm looking at it now.

S-Bartfast:
Alright, just verified it myself and yes, I get the exact same behavior.
Working on a fix now...

S-Bartfast:
Yup, fixed it.

The issue was indeed related to the 'variable bank' refactoring.

In particular it was caused by the 'findAndRemove(...)' method removing previously seen variables from the 'returnList' when it should have only been looking for them.
The method has been renamed 'getVariable(...) and now accepts a Boolean to indicate whether the found variable should be removed or not:
![6655_fix](https://user-images.githubusercontent.com/6573522/28750357-87ec5efc-7527-11e7-9aa4-82880f2d851e.png)


As it was, adding yet another 'b' to the 'equations' expressions would have made it appear again.  In fact any 'free variables' that had an odd number of mentions would have appeared where as even-numbered occurrences were removed.  This has now been fixed and any variables mentioned in the 'equations' expressions appear in the 'free variables' table (as they should).

A pull request has also been made to incorporate this fix into the main repository.

pstorch:
closed via #6657

PR #6657, opened 2017-07-30
Fixed bug #6655

S-Bartfast:
It was caused by the 'findAndRemove(...)' method removing previously seen variables from the 'returnList' when it should have only been looking for them.
The method has been renamed 'getVariable(...) and now accepts a Boolean to indicate whether the found variable should be removed or not.


pstorch:
ok to test

pstorch:
I still can't tell CI that it is "ok to test". 
@S-Bartfast in the commit message the "Fixed" keyword needs to be followed directly by the issue number. Otherwise Github won't close the issue automatically. 
The code itself looks OK to me.

SammysHP:
OK to test

PR #6663, opened 2017-07-30
Let expression parser fail on unbalanced parentheses

UniQP:


pstorch:
Makes sense. At the moment, it's just ignoring the missing parenthesis, leading to possible errors.

Lineflyer:
@S-Bartfast : Any objections on this (as you contributed the original code)?

pstorch:
I think we can merge this

S-Bartfast:
> Any objections on this (as you contributed the original code)?
Oh sorry, forgot to respond.
No objections at all.

S-Bartfast:
Although most actual calculators I've used (including the Android default) do seem to automatically append closing brackets to the end if they're needed.

Is there any reason we'd want to make our calculator more strict than most?  If anything I guess I'd kind of err on the side of being less strict (ie. more fault tolerant) than most other calculators but I really don't have a strong opinion either way.

pstorch:
That's interesting, indeed. But isn't it dangerous to assume the bracket is missing at the end? How will a user spot a possible mistake?

S-Bartfast:
> But isn't it dangerous to assume the bracket is missing at the end? How will a user spot a possible mistake?

Possibly, but it's a trade off between being strict and being fault tolerant.  As it is, programming environments tend to be strict about such expressions and calculators tend to be more forgiving.  Neither way is necessarily better than the other, it's a trade off, but possibly calculators err on the side of being fault tolerant because, unlike a programming environment, it's much harder to give useful feedback to help a user amend their formulation.

I'm not exactly sure on the reasons and as I said above, I'm not particularly fussed either way, I was just pointing out that automatically closing off brackets is in fact normal for calculators.  Given these expressions are stored and displayed perpetually (unlike calculators where the expression typically disappears when you hit the equals button), possibly a stricter parsing is justified, but I was just pointing out that it's a trade-off and that most calculators seem to take the other approach. ;)

SammysHP:
Most TI and HP calculators that I know require balanced brackets, while at least some Casio calculators accept omitted closing brackets as long as the input can be parsed from left to right without syntax error.

I suggest to add closing brackets automatically during typing (if feasible) and don't accept unbalanced brackets.

S-Bartfast:
As it is I've only used CASIO calculators IRL but the default Android, this [HiPER Calc](https://play.google.com/store/apps/details?id=cz.hipercalc&hl=en) and this [RealCalc](https://play.google.com/store/apps/details?id=uk.co.nickfines.RealCalc&hl=en), along with the Windows default calculator all automatically close off brackets.  I don't doubt there are 'strict' real world calculators out there, I just haven't used one myself ;)

pstorch:
Just found some online calculators which use strict mode:
https://www.calculatorsoup.com/calculators/math/math-equation-solver.php
https://www.wyzant.com/resources/lessons/math/algebra/calculators/expression-factoring?expression=4%2B(3*4
http://calculator-tab.com/#/
http://www.calculatoria.com/#c

These automatically inserts a bracket at the end while typing: 
https://www.mathpapa.com/calc/alg
https://web2.0calc.com/

Google silently assumes a bracket at the end: https://www.google.de/search?q=3%2B%284*4&oq=3%2B%284*4

I guess you can argue both ways. I prefer strict mode at the moment to not mask possible errors. Assume you calculate the coordinates of the next stage and then walk a few km until you arrive nowhere just to find out you missed a bracket in the formula. But I won't insist on this logic. If there is a consensus to be more relaxed here, I'm fine with reverting this change. We can also wait for more feedback from the users.

UniQP:
Is there a standard way to indicate that the content of an input box is not valid? If so, we should use it to mark formulas with unbalanced parentheses.

S-Bartfast:
At the moment an '*' is displayed in the result rather than a number (or the blank '_' character which means no expression).  There's nothing official about this but it seems intuitive enough.  Oh, and the 'name' of the equation is also set to an 'inactive gray' colour.

PR #6668, opened 2017-07-31
fix #6666 parse OC URLs

### LINKED ISSUE #6666: Some OC URLs not parsed correctly
c:geo cannot handle this URL: https://www.opencaching.de/viewcache.php?wp=OC10F41&log=A#log1186982

pstorch:
Read geocode from `wp` URI parameter from OC URLs.

### DIFF
--- a/main/src/cgeo/geocaching/connector/AbstractConnector.java
+++ b/main/src/cgeo/geocaching/connector/AbstractConnector.java
@@ -328,6 +328,10 @@ public abstract class AbstractConnector implements IConnector {
     public void logout() {
     }
 
+    public String getShortHost() {
+        return StringUtils.remove(getHost(), "www.");
+    }
+
     @Override
     @Nullable
     public String getCreateAccountUrl() {
--- a/main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
+++ b/main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
@@ -8,6 +8,7 @@ import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.CryptUtils;
 import cgeo.geocaching.utils.DisposableHandler;
 
+import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
@@ -116,7 +117,7 @@ public class OCApiConnector extends OCConnector implements ISearchByGeocode {
     @Override
     @Nullable
     public String getGeocodeFromUrl(@NonNull final String url) {
-        final String shortHost = StringUtils.remove(getHost(), "www.");
+        final String shortHost = getShortHost();
 
         final String geocodeFromId = getGeocodeFromCacheId(url, shortHost);
         if (geocodeFromId != null) {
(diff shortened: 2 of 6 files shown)


SammysHP:
Why here and not in `ConnectorFactory.getGeocodeFromURL()`?

pstorch:
Ah, good point. Saw only the geocaching.com URI parsing in `CacheDetailActivity` class. Why is this not handled (completely) in the `GCConnector.getGeocodeFromUrl()`?
I'll have a look.

pstorch:
Ok, this looks better.
I was tempted to remove the geocaching.com parsing in https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/CacheDetailActivity.java#L258, because it is also available in https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/connector/gc/GCConnector.java#L345, but not the `guid` logic (which is specific to GC).

pstorch:
@Bananeweizen both inherit from `OCConnector` and call `super.getGeocodeFromUrl(url);`. But both might have a similar problem with extracting the parameter `cacheid`. I'll fix that.

pstorch:
Found some more issues with handling of hostnames.

SammysHP:
Btw: Is the short URL handled correctly?  https://opencaching.de/OC10DA9

pstorch:
Testcase added :)

Bananeweizen:
:+1: 
@pstorch I knew there would be more issues since I had originally written the substring based code you fixed in your change, and I remembered vaguely to have used the same code pattern multiple times.

pstorch:
I'm thinking about some more refactorings, but don't want to take them in this issue. As mentioned in a comment above there is logic in the `CacheDetailsActivity.onCreate()` specifically for GC. This pattern `https://www.geocaching.com/geocache/GCxxx_*` is implemented twice (`CacheDetailsActivity` and `GCConnector`). But the pattern `https://www.geocaching.com/seek/cache_details.aspx?wp=GCxxx` is only implemented in `CacheDetailActivity`. So this test fails at the moment:

```
assertThat(GCConnector.getInstance().getGeocodeFromUrl("https://www.geocaching.com/seek/cache_details.aspx?wp=GC43GWB")).isEqualTo("GC43GWB");
```

Then we have the special handling of the GC `guid`. Which should also be moved into the `GCConnector` but doesn't fit into the `getGeocodeFromUrl()` method.

Then the `getGeocodeFromUrl()` takes a `String`, but sometimes we need an `Uri` and so we create one in the Connectors.  Actually we already have an `Uri` instance in the `CacheDetailsActivity`, but convert it to a `String` when calling this method. So I would like to change the method signature.

Bananeweizen:
Regarding the general moving of that code into the connectors, I'm completely with you. It should be possible to re-arrange that such that it is not necessary again in the details activity. I can only speculate on historic reasons (it surely was there before we introduced the connectors as such).

Regarding guid I'm not so sure. I was able to remove that from a lot of other places where it was sprinkled in, but there were still lot of areas where the guid looked completely wrong, nevertheless I could not easily remove it (e.g. the methods to load by rectangular area, by circle etc.). If you find a good way, I encourage you to tackle that for sure.

And of course feel free to change the method signatures from Strings to Uris, I must admit it never occured to me that we are transforming back and forth.

PR #6676, opened 2017-08-07
fix #6670 cache ellipsis with strike through and color

### LINKED ISSUE #6670: Strickenthrough Cache names not abreviated correctly
In listviews, GC6KGKZ is shown in c:geo as having a name of "Patty's", instead of "Patty's verbandtrommeltje" (though it would probably be abbreviated with ellipsis). See screenshot.

pstorch:
TextView ellipsis doesn't work with Spannable, which we use for styling disabled and archived caches with strike through and color red. An alternative is to use PaintFlags and TextColor on the TextView.

### DIFF
--- a/main/src/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/cgeo/geocaching/ui/CacheListAdapter.java
@@ -24,13 +24,10 @@ import cgeo.geocaching.utils.MapUtils;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.res.Resources;
+import android.graphics.Paint;
 import android.graphics.drawable.BitmapDrawable;
 import android.support.annotation.NonNull;
 import android.support.v4.content.ContextCompat;
-import android.text.Spannable;
-import android.text.Spanned;
-import android.text.style.ForegroundColorSpan;
-import android.text.style.StrikethroughSpan;
 import android.view.GestureDetector;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/ui/CacheListAdapter.java
```diff
@@ -428,23 +426,14 @@ public View getView(final int position, final View rowView, final ViewGroup pare
         compasses.add(holder.direction);
         holder.direction.setTargetCoords(cache.getCoords());
 
-        Spannable spannable = null;
         if (cache.isDisabled() || cache.isArchived() || CalendarUtils.isPastEvent(cache)) { // strike
-            spannable = Spannable.Factory.getInstance().newSpannable(cache.getName());
-            spannable.setSpan(new StrikethroughSpan(), 0, spannable.toString().length(), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+            holder.text.setPaintFlags(holder.text.getPaintFlags() | Paint.STRIKE_THRU_TEXT_FLAG);
```
Bananeweizen: Does this really work in a list containing both kinds of caches? I thought that with the view holder pattern, we always must make sure all relevant flags are set for each element of the list. Therefore I would expect that we need to set the paint flags and the text color to a common standard every time before overriding one of them in the if statements. Otherwise the we might end up e.g. with an old paint flag from one of the previous iterations, even though we now want to render a non-disabled cache.

pstorch:
Oh, on the first glance it looked good. My Testlist had disabled and archived at the end. But you are right. It gets mixed up. 🙁

Bananeweizen:
But then it should really work with just adding
setTextColor(...)
setPaintFlag(getPaintFlag & ~FLAG_STRIKE_THROUGH)
before the if-statement. Your code as such seems fine to me.

pstorch:
I just needed to find out the right color depending on the chosen skin.

PR #6685, opened 2017-08-15
Fix #6682: Failed to detect loggin status

### LINKED ISSUE #6682: Error displayed when do a cache log
- Open a geocache
- Log as "found it"
- When you are submitting it display an error saying "login failed"
- Reload geocache information, you see that the log was submitted successfully

It begin to appear, after new Geocaching.com layout.

Yes/No

##### System information:
```
```

kumy:
GS has changed some style.
Adapt the regex to correspond to the new format.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -76,7 +76,7 @@ public final class GCConstants {
     static final Pattern PATTERN_BACKGROUND_IMAGE = Pattern.compile("<body background=\"(.+?)\"");
 
     // Info box top-right
-    public static final Pattern PATTERN_LOGIN_NAME = Pattern.compile("class=\"li-user-info\"[^>]*>.*?<span>(.*?)</span>", Pattern.DOTALL);
+    public static final Pattern PATTERN_LOGIN_NAME = Pattern.compile("class=\"[^\"]*li-user-info\"[^>]*>.*?<span>(.*?)</span>|<span class=\"user-name\"[^>]*>(.*?)</span>", Pattern.DOTALL);
     /** Use replaceAll("[,.]","") on the resulting string before converting to an int */
     static final Pattern PATTERN_CACHES_FOUND = Pattern.compile(Pattern.quote("cache-count\">") + "\\s*([\\d,.]+)", Pattern.DOTALL);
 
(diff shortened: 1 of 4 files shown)


kumy:
@Lineflyer do you have time to test the build from the CI?

kumy:
There seems to have more changes than expected...

For example, on the `settings/homelocation` page, some html parts are now(?) embedded in `script` tags! Which break `document.select("input.search-coordinates").attr("value")`

```
<script type="text/x-handlebars-template" id="tplSearchCoords">
    <div class="search-coordinates-wrapper">
        <input class="search-coordinates" id="Query" name="Query" placeholder="Location&#32;(city&#32;or&#32;postal&#32;code)" type="text" value="xxx" />
        <span class="btn-wrapper">
            <button id="search" class="btn btn-primary">Search</button>
        </span>
    </div>
</script> 
```

Lineflyer:
Do I interprete your comment and the CI result correct, that you need to recheck for other changes before it makes sense to test?

kumy:
Last (failed) build should at least fix the logging problem with cache and TB. However unit tests reveal more hidden problems.

kumy:
Other problems:
* One cache used for Unit testing is now archived (should not be a big problem)
* Home location failed to be detected, due to html changes...

kumy:
All tests should now be ok.
@Lineflyer ready to test?

kumy:
I think we could ignore the build failed.
CI has two build chains. The one marked as expirimental always fail since some time, whereas the [other one](http://ci.cgeo.org/job/cgeo%20continuous%20integration/1684/) ended with success.

Lineflyer:
IMHO your change looks straight forward enough to merge first and directly test with a RC version.
I will do that in some minutes and provide feedback.

Can you say something about "backwards compatability", i.e. If they roll back to the old elements, will your change still cope with that?

Lineflyer:
Question to the Git experts:
When merging via Github webpage, shall I use "rebase and merge" or "squash and merge"?

kumy:
> Can you say something about "backwards compatibility"

* first commit 846bf53 should not be a problem in case of a rollback on their side (login detection)
* second commit 78fd083 ***will be** problematic* (home coordinates detection)

kumy:
@Lineflyer for this current PR, don't use "squash" method please as I've intentionally separated commits for each error. So "rebase and merge" is the way to go. :)

Issue #6519, opened 2017-04-24
Use new logging page


Lineflyer:
In todays [release notes](http://forums.groundspeak.com/GC/index.php?showtopic=344009) Groundspeak announces, that a new logging page is introduced.
You can see in the video on [this page](https://www.geocaching.com/blog/2017/04/test-a-new-logging-experience-is-coming-to-geocaching-com/) that it looks different to the old one.

I tested logging just now with 2017.04.24-NB and had no problem to post logs with and without attached photo. 
However looking into the browser I do still see the "old" logging page on my account...so might be, that the change is not yet live.

We should monitor this.

SammysHP:
> You will still have the option of returning to the old logging page when creating a new log, but encourage you to try out the new page and provide any feedback.

So we have some time to test the new logging page.

alan666notb:
I get the new logging page on the website every time now,
cgeo still seems to log caches OK.

Bananeweizen:
Try logging a field note via browser. That is _different_ since it allows needs maintenance as _additional_ flag instead of as separate log, according to a thread on OC.de

steve8x8:
Had to do exactly this: after performing owner maintenance, I couldn't submit a matching log using c:geo (03.29), had to offline write note instead and upload as field notes. Web page allowed me to change the log type.

Lineflyer:
No need to use field notes to see the new logging page, its live now, e.g.:
https://www.geocaching.com/play/geocache/gc1pqe6/log

We should use this instead of the old one as it might be retired some day.

Lineflyer:
We should look into using this page for logging.
However at the moment I do not see any urgent need, as the new is still under approval and it seems far away from perfect.

johan974:
Hi, 
The good news is that GC provided the 'opt out' way as a workaround. I hope for very long ... 
When investigating the new logging HTTP sequence I saw some elements that looked like Oauth(2). 
* The old logging page points you via HTTP 302/301 redirects to something like. 
   https://www.geocaching.com/play/geocache/gc3brd4/log. Just like Lineflyer indicated. 
* I see that an access_token is needed int the actual post header. 
* Scanning beyond a lot of intermediate HTTP requests, both GET/POST requests are done for the token. 
   GET "https://www.geocaching.com/account/oauth/token". Why do they need so many of these ;-) ? 
* Then the final POST is done with the Authorization: berear ey* (access token). 
[...]

PR #6686, opened 2017-08-15
6519 new logging page

pstorch:
This is work in progress for the new logging page #6519

It's a mixture of code provided by @johan974 and @samueltardieu 's PR #6548.
It's compiling and I managed to send my first log with it.

Open Issues/Tasks:

- favorites: 
  - [x] reading available favorites in in `GCLoggingManager` via API
  - [x] post favorite with Log

- [x] read possible log types from new log page in `GCLoggingManager`

- [x] Trackables

- [x] Photo(s)

- [x] Handling of user setting (new or old log page), I have the new logpage as default now
- [x] Cleanup code

- [ ] "Report a problem" aka Needs Maintenance and Needs Archived (I guess this needs a rework of the logging UI as well and could also be postponed)

- [x] Posting of coordinates: is only used for cache owners with LogType `Update coordinates`. C:geo doesn't support it. So, we don't need to implement it for the new log page. At least for now.

- [ ] GUID for GCVote (is it needed?)

### DIFF
--- a/main/src/cgeo/geocaching/connector/AbstractLogin.java
+++ b/main/src/cgeo/geocaching/connector/AbstractLogin.java
@@ -101,5 +101,5 @@ public abstract class AbstractLogin {
             actualCachesFound++;
         }
     }
-    
+
 }
--- a/main/src/cgeo/geocaching/connector/AbstractLogin.java
+++ b/main/src/cgeo/geocaching/connector/AbstractLogin.java
@@ -8,9 +8,11 @@ import cgeo.geocaching.network.Network;
 import cgeo.geocaching.settings.Credentials;
 import cgeo.geocaching.settings.DiskCookieStore;
 
-import org.apache.commons.lang3.StringUtils;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+
+
+import org.apache.commons.lang3.StringUtils;
 
 public abstract class AbstractLogin {
 
@@ -94,4 +96,10 @@ public abstract class AbstractLogin {
     @NonNull
     protected abstract StatusCode login(final boolean retry, @NonNull final Credentials credentials);
 
+    public void increaseActualCachesFound() {
+        if (actualCachesFound >= 0) {
+            actualCachesFound++;
+        }
+    }
+    
 }
--- a/main/src/cgeo/geocaching/connector/ec/ECApi.java
+++ b/main/src/cgeo/geocaching/connector/ec/ECApi.java
@@ -127,8 +127,8 @@ final class ECApi {
 
             final String data = Network.getResponseData(response, false);
             if (!StringUtils.isBlank(data) && StringUtils.contains(data, "success")) {
-                if (logType == LogType.FOUND_IT || logType == LogType.ATTENDED) {
-                    ecLogin.setActualCachesFound(ecLogin.getActualCachesFound() + 1);
+                if (logType.isFoundLog()) {
+                    ecLogin.increaseActualCachesFound();
                 }
                 final String uid = StringUtils.remove(data, "success:");
                 return new LogResult(StatusCode.NO_ERROR, uid);
--- a/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
@@ -1,24 +1,11 @@
 package cgeo.geocaching.connector.gc;
 
-import android.os.Bundle;
-import android.support.v4.app.LoaderManager;
-import android.support.v4.content.Loader;
-import android.widget.CheckBox;
-
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import android.support.annotation.NonNull;
-import android.support.annotation.Nullable;
-
-import java.util.Calendar;
-import java.util.Collections;
-import java.util.List;
-
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.connector.AbstractLoggingManager;
 import cgeo.geocaching.connector.ImageResult;
 import cgeo.geocaching.connector.LogResult;
+import cgeo.geocaching.connector.trackable.TrackableBrand;
 import cgeo.geocaching.enumerations.Loaders;
 import cgeo.geocaching.enumerations.StatusCode;
 import cgeo.geocaching.loaders.UrlLoader;
@@ -27,18 +14,30 @@ import cgeo.geocaching.log.LogType;
 import cgeo.geocaching.log.TrackableLog;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
-import cgeo.geocaching.network.Parameters;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.Log;
-import cgeo.geocaching.utils.TextUtils;
+
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v4.app.LoaderManager;
+import android.support.v4.content.Loader;
+import android.widget.CheckBox;
+
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.List;
+
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
 
 class GCLoggingManager extends AbstractLoggingManager implements LoaderManager.LoaderCallbacks<String> {
 
     private final LogCacheActivity activity;
     private final Geocache cache;
 
-    private String[] viewstates;
     @NonNull private List<TrackableLog> trackables = Collections.emptyList();
     private List<LogType> possibleLogTypes;
     private boolean hasLoaderError = true;
@@ -57,7 +56,7 @@ class GCLoggingManager extends AbstractLoggingManager implements LoaderManager.L
             return null;
         }
         activity.onLoadStarted();
-        return new UrlLoader(activity.getBaseContext(), "https://www.geocaching.com/seek/log.aspx", new Parameters("ID", cache.getCacheId()));
+        return new UrlLoader(activity.getBaseContext(), "https://www.geocaching.com/play/geocache/" + cache.getGeocode() + "/log", null);
     }
 
     @Override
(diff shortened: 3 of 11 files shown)


pstorch:
Still work in progress. Started working on the trackable logs, but it gives me a 500 error. Don't know why yet.

pstorch:
It's getting slightly better. I get a bit used to the new web API of GC.
Now Trackables logging alongside a cachelog is working.

Lineflyer:
👍

kumy:
retest this please

kumy:
Jenkins retest this please

johan974:
@pstorch - good work! I mean - not the 'all checks have failed' ;-)
In which branch are you working?

pstorch:
@johan974 look at the top of the PR 😉

pstorch:
Current State:
the GCLoggingManager is still reading the old log page to get the available logtypes. It was also used to get the trackables in your inventory and the available favorite points. The latter two I'm reading via new  (web-)API endpoints. But I still need to parse the new logpage to get the available logtypes.

For the available favorites the API needs the users profile referenceCode `PR...`. Till now we haven't used it anywhere (or I haven't found it). But I found a URL to get it from: `https://www.geocaching.com/play/serverparameters/params`
It's returning a javascript object which can be stripped to a simple json document. I provides some user profile information. We can also read from it the custom date format, user locale, the profile referenceCode, premium state, ...
At the moment I read it in the GCLoggingManager, but I think it should be loaded in GCLogin and then cached as long as the user is logged in. It can also replace some code in GCLogin. But it's a further refactoring. I want to do it step by step.

pstorch:
Ok, now the available logtypes are also read from the new log page. That means that the old log page isn't loaded anymore.
But I found that the cache GUID was also read from the old page. According to the code comments the GUID is needed for GCVote. The new log page doesn't contain the GUID. The caches I tested with had the GUID already loaded, so this code was not needed. But does anyone know in what situations this is not the case and we have to load the GUID to log a GCVote voting?
The next open issue I'll look into is the photo upload. 
The new way of reporting problems (NM and NA-Logs) needs some more refactoring, even a redesign/extension of the UI. So I keep it for later ;)

pstorch:
Successfully posted a log with an image :smile: 
I think I'll clean up a little, then I decide on the next partial task.

pstorch:
retest this please

pstorch:
Looks like it also works for users which have opted out of the new logging page.

So what is left:
- Posting of coordinates: looks like it's not even used by the website itself -> ignore for now
- loading of geocache GUID for GCVote log: not sure if this is needed, when I tested it, the GUID was always already available
- "Report a problem" aka Needs Maintenance and Needs Archived: as I see it, it is not mandatory to have it from the start. So we could decide to implement it now (takes some more time) or later in a new issue. I would prefer to solve it later, to have the base feature earlier in master for test and feedback.
- Cleanup code: yeah, sure ;)

pstorch:
Regarding the GUID for GCVote: when using the live map and you log the cache from the map popup, then it can be that we don't have the GUID and need to retrieve it separately. 
I'm thinking about requiring a cache refresh in case the user wants to make a GCVote rating and the guid isn't available yet. That means for all users not using GCVote a refresh is not necessary. I would also move the code out of the GCLoggingManager, because it is not needed there. It's needed right before the GCVote rating is posted.

pstorch:
I think it's ready for review and test.
The 3 remaining issues:
- posting of coordinates is not even implemented by the website itself, so let's keep it for later
- GUID for GCVote: it's seldom that the guid is missing. Sometimes on the live map, if the cache is not properly loaded (I think we have some other issues in this case). There is no error, just the rating bar is missing. This can be fixed by the user by refreshing/storing the cache properly. 
- NM/NA-Logs: as these are not so frequent log types and would require some more work. I would like to takle this as a separate issue while collecting feedback on the new logging function in master.

pstorch:
retest this please

kumy:
retest this please

pstorch:
retest this please

kumy:
Jenkins test this please

kumy:
Jenkins test this please

main/src/cgeo/geocaching/connector/AbstractLogin.java
```diff
@@ -94,4 +96,9 @@ public StatusCode login(@Nullable final Credentials credentials) {
     @NonNull
     protected abstract StatusCode login(final boolean retry, @NonNull final Credentials credentials);
 
+    public void increaseActualCachesFound() {
+        if (getActualCachesFound() >= 0) {
```
UniQP: Introduce a local variable for getActualCachesFound()?

main/src/cgeo/geocaching/log/LogTypeTrackable.java
```diff
@@ -1,18 +1,19 @@
...
 import android.support.annotation.NonNull;
+import android.support.annotation.StringRes;
+
 import java.util.ArrayList;
 import java.util.List;
 
 public enum LogTypeTrackable {
     DO_NOTHING(1, -1, "", R.string.log_tb_nothing, LogType.UNKNOWN),
-    VISITED(1001, 5, "_Visited", R.string.log_tb_visit, LogType.VISIT),
+    VISITED(1001, 5, "_Visited", R.string.log_tb_visit, LogType.VISIT, 75),
```
UniQP: Can we avoid the additional parameter by using LogType.VISIT.iconName instead?

pstorch:
@UniQP 
> Regarding the GUID: Is it possible to get the GUID via the API? Is there any documentation about the API (how do you know about the coordinatesInput section)?

First I need to clarify that this API is not the "Geocaching.com Live API" everyone is talking about. It is the API they use for their own Web Frontend. Maybe I should rename the API class to WebAPI or something else to make it more distinct.

There is no documentation, not one I'm aware of. My PR is based on the work of @samueltardieu who implemented the basic flow with the Authorization Token in his own PR. Additionally I got some information from @johan974 regarding the new logging page. And some I found out myself. So it is mainly reverse engineering. 

So far I'm not aware of a Geocode to GUID call. The possibility we have at the moment is to load and parse the Cachelisting the old way. Which is something I would like to avoid in this special case.

> I also wonder what happens if we submit a GCVote without the GUID but I'm still waiting for my GCVote password.

If there is no GUID available yet, the GCVote stars are hidden and a voting is not possible. Reading on the GCVote Website, they have based their database on the GUID for some reasons. So I guess it is not possible to send a vote without it.

main/src/cgeo/geocaching/connector/ec/ECApi.java
```diff
@@ -127,8 +127,8 @@ static LogResult postLog(@NonNull final Geocache cache, @NonNull final LogType l
 
             final String data = Network.getResponseData(response, false);
             if (!StringUtils.isBlank(data) && StringUtils.contains(data, "success")) {
-                if (logType == LogType.FOUND_IT || logType == LogType.ATTENDED) {
-                    ecLogin.setActualCachesFound(ecLogin.getActualCachesFound() + 1);
+                if (logType.isFoundLog()) {
+                    ecLogin.increaseActualCachesFound();
```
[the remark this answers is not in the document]
pstorch: Yes, it's a bit unrelated. I'll extract it into an extra commit.

main/src/cgeo/geocaching/connector/AbstractLogin.java
pstorch: I could also reference the field directly

UniQP: Yes, that sounds like a good idea.

main/src/cgeo/geocaching/log/LogTypeTrackable.java
pstorch: I think I know what you mean. You want to avoid the duplication of the number `75`. But somehow it doesn't feel right to use `LogType.VISIT.iconName` here. At this place it is not the name of an icon. Maybe the naming in the `LogType` class is also missleading.

main/src/cgeo/geocaching/connector/AbstractLogin.java
```diff
@@ -94,4 +96,9 @@ public StatusCode login(@Nullable final Credentials credentials) {
     @NonNull
     protected abstract StatusCode login(final boolean retry, @NonNull final Credentials credentials);
 
+    public void increaseActualCachesFound() {
+        if (actualCachesFound >= 0) {
+            setActualCachesFound(actualCachesFound + 1);
```
UniQP: That's actually `actualCachesFound++;`. Sorry, I though you already had this in mind when you proposed to reference the field directly.

pstorch: ah, sure

pstorch:
Regarding the coordinates:
I just wanted to document where I found it and suddenly saw a hint, where it might be used. And then I found it: It is used in the `Update coordinates` LogType for Cache owners.

image

But we don't support this log type. Not in the old log page. So I guess we can tick this open issue off.

kumy:
Jenkins retest this please

johan974:
Congrats to @pstorch pstorch!

pstorch:
What do you think, should we start testing this in the nightlies?

johan974:
@pstorch  - how is it going? I saw your question since 18 days, I guess you asked the (other) members to react on your work.

pstorch:
@johan974 yes, that was a general question to the whole team. But the plan is to make a new feature release first and then start new big changes (including this one) on master.

johan974:
Has this fix for logging been implemented?

pstorch:
Implemented yes, this PullRequest has the changes ;)
But they are not yet merged. The next release is about the new map and I guess we don't want to include more (big) changes. So this has to wait.

pstorch:
@Lineflyer I would like to see this in sooner than later. 
Regarding the two open issues:
- "Report Problem": this needs a redesign of the Logging UI, because at the moment we don't have this as a separate option. And OC also supports something similar which should be done the same way. But it looks like a change on its own. I think we could live without it for a while. What do you think?
- GUID for GCVote: if the cache is not properly loaded we don't have the GUID yet, so we can't add a GCVote. I think this is a rare problem and can be easily spotted and fixed by the user. 

Regarding the stability of this PR. I've not tested it for a while, but I can restest it tonight.

main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
```diff
@@ -102,11 +115,16 @@ public LogResult postLog(@NonNull final LogType logType, @NonNull final Calendar
...
+            final ImmutablePair<StatusCode, String> postResult = GCWebAPI.postLog(cache, logType,
+                    date.getTime(), log, trackableLogs, favCheck.isChecked());
 
             if (postResult.left == StatusCode.NO_ERROR) {
+                DataStore.saveVisitDate(cache.getGeocode(), date.getTime().getTime());
+
+                if (logType.isFoundLog()) {
+                    GCLogin.getInstance().increaseActualCachesFound();
+                }
+
```
UniQP: We can use an `else` here.

pstorch: right

pstorch:
A short test after rebasing it onto the current master: it's still working as I left it. Waiting for an ok to be merged and tested in the nightlies.

Lineflyer:
Regarding the adaptions needed for maintenance notification:
Did you already implement the functions itself so that "only" an UI adaption is needed?
I think we need to a kind of dropdown with the relvant options below the "Add to favorite". Right?

pstorch:
OK, let's try it out in the nightlies.
If I merged it too fast don't hesitate to make comments. 
I'll open another issue to explain the "Report Problem" feature. What I found out and what is missing. Maybe tomorrow.

johan974:
Wishing Success!

Issue #6687, opened 2017-08-16
Show complete variable field in landscape view


Lineflyer:
Of course the new calculator function reaches the limits of the UI.
@S-Bartfast already implemented it quite well with this scrolling variable field. 

However I could imagine the view being optimized in landscape mode:
Currently it will keep the formulas and calculations on the left half of the screen, whereas the right half of the screen is filled with the note field.
For a better overview I would like to see the variable field expanding over the full screen width allowing an easy overview by just turning your device into landscape. 
The note field could just start underneath.

Here is how it currently looks:
![screenshot_20170816-110840](https://user-images.githubusercontent.com/949669/29356753-92f46140-8275-11e7-9afc-b4500e4a1615.png)

SammysHP:
On the other hand notes are helpful if you've copied the formula into it. But a bigger view of the variables would be nice as well...

S-Bartfast:
I'm more than open to suggestions on how this in particular can be improved.

To begin with I did just have the exact same layout in both portrait and landscape but it did look a bit strange on a tablet.  In particular the 'digit-buttons' were very broad and the 'variables' section looks kind of strange when the expressions were short:
![image](https://user-images.githubusercontent.com/6573522/29361081-1b1072ae-82c5-11e7-884f-027690c2b9d9.png)

On a phone it's not too bad:
![image](https://user-images.githubusercontent.com/6573522/29361272-febd3ef6-82c5-11e7-91d8-c31bf8408c2b.png)

But on a tablet it's really not that good, which is why I implemented the side-notes view:
![image](https://user-images.githubusercontent.com/6573522/29362166-f4fa586e-82c9-11e7-947d-efd9d2385c04.png)
Do we have any details on how many people use tablets for caching?

I'll look into exclusively using the 'side-notes' layout on _tablets_ in landscape orientation.


[...]

Issue #6664, opened 2017-07-30
Incorporate usage guide into WP calculator


Lineflyer:
I am wondering if we could add a comprehensive usage guide by just adding a heading description to each "section" of the calculator view.

E.g. the coordinate view could be preceeded by something like:
"You can click on any of the coordinate digits below to replace it by a variable for further calculation"

The variable section could be preceeded with:
"For each upper case variable assigned above you can either provide a numeric value or an equation also containing new lower case variables"

etc.

Lineflyer:
Fixes #6664

Lineflyer:
Fixes #6665

Lineflyer:
Closes #6665

SammysHP:
Once a user is an "expert", such messages are annoying. I'd suggest a <kbd>?</kbd> button next to <kbd>✔</kbd> and <kbd>✘</kbd>.

Lineflyer:
>  I'd suggest a ? button next to ✔ and ✘.

Good idea. This would allow us to put a quick guide and external links (e.g. to demonstration video) on a separate page.
Question would be if @S-Bartfast would be able and willing to implement such a help page as well or we need to find someone else to contribute this?

SammysHP:
There is a very simple solution: Just some HTML in a (almost) fullscreen webview. This could lead to a full integrated manual (assuming we find enough people writing it).

Lineflyer:
Are you refering to an online version (getting the page e.g. from our webpage repo) or something offline?

SammysHP:
Both is possible, but I was talking about an offline version (html file in the APK which is displayed via a webview).

S-Bartfast:
I'll be happy take a look at implementing a help screen.

I'd also be more than happy for the videos to be used as demonstration material but I'm thinking I'll redo them if there going to be used officially, just to make them a little more succinct and polished.

Lineflyer:
@S-Bartfast 
> I'd also be more than happy for the videos to be used as demonstration material but I'm thinking I'll redo them if there going to be used officially, just to make them a little more succinct and polished.

I also posted this in the closed issue, but maybe its better to post it here:
I think redoing the videos would be great for later official use.
However are you ok if I link this video to a german user group on Facebook (there are quite some nightly users there and I would like to get their feedback and test): https://youtu.be/eN2y5mvJul4

S-Bartfast:
As mentioned in #6689 I've been working on this issue.

[...]

PR #6689, opened 2017-08-16
Use improved hints for equation content

Lineflyer:
showing more of the supported calculation operators

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1595,8 +1595,8 @@
     <string name="init_low_disk_space_message">The geocache data directory is running low on storage space. This can cause malfunctions of c:geo. Please free up some space.</string>
     <string name="init_confirm_debug">c:geo is running in debug mode!</string>
     <string name="list_confirm_debug_message">The debug mode causes instabilities! Do you want to deactivate the debug mode?</string>
-    <string name="equation_hint">(a+b)*c/d</string>
-    <string name="free_variable_hint">2+2</string>
+    <string name="equation_hint">(a+b)*c^d/e</string>
+    <string name="free_variable_hint">(5+4-3)*2/3</string>
     <string name="empty_equation_result">_</string>
     <string name="equation_error_result">*</string>
     <string name="lat_point_min_format">N DDD° MM.PPP</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1595,8 +1595,8 @@
     <string name="init_low_disk_space_message">The geocache data directory is running low on storage space. This can cause malfunctions of c:geo. Please free up some space.</string>
     <string name="init_confirm_debug">c:geo is running in debug mode!</string>
     <string name="list_confirm_debug_message">The debug mode causes instabilities! Do you want to deactivate the debug mode?</string>
-    <string name="equation_hint">a^2 + b^2</string>
-    <string name="free_variable_hint">2 + 2</string>
+    <string name="equation_hint">(a+b)*c/d</string>
+    <string name="free_variable_hint">2+2</string>
     <string name="empty_equation_result">_</string>
     <string name="equation_error_result">*</string>
     <string name="lat_point_min_format">N DDD° MM.PPP</string>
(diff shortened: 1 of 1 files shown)


S-Bartfast:
Just wondering what it is about these hints that you think is an improvement.

I mean no offence but I certainly put thought into the original hints and decided on them for a few reasons.  In what way is it that you think these are better?

In particular, I used the "a^2 + b^2" to demonstrate that 'powers' were available.  While I was concerned that the '^' character may appear foreign to some I still elected to use it as I figured that even if the meaning of the character was lost it would still look sufficiently 'maths-ie' to convey the meaning.  I did actually want to show bracket priority as well but ultimately decided against it in favour of the Pythagorean formula as it is such a famous and elegant equation 👍.  Is there any particular advantage to "(a+b)*c/d" other than it showing bracketing?
 
As for the spaces around the '+' signs, that was also intentional; partly to add visual neatness to the layout but also because the 'hint' seems to set the minimum width of the field and there is a subtle bug that seems to exhibit itself on occasion if the fields are too narrow.

Lineflyer:
>  Is there any particular advantage to "(a+b)*c/d" other than it showing bracketing?

The main purpose was to present all available operators. But you are right I forgot `^` to make it complete (and `%` but that would be to special I think).
I came to this thought, because I showed the calculator to a friend and he failed to understand what to fill into the fields. He was not familiar with the `^` operator thus not got the hint, that he can actually insert an equation here.

> As for the spaces around the '+' signs, that was also intentional; partly to add visual neatness to the layout but also because the 'hint' seems to set the minimum width of the field and there is a subtle bug that seems to exhibit itself on occasion if the fields are too narrow.

Good to know...in this case we should not change it but maybe also could show the possible operators here.

>  in favour of the Pythagorean formula as it is such a famous and elegant equation

:) But that was exactly what confused the non-mathematical expert user, whom I showed the calculator.
And to be honest, only very few of the caches I have come across are using powers in their formulas.

S-Bartfast:
> I came to this thought, because I showed the calculator to a friend and he failed to understand what to fill into the fields. He was not familiar with the ^ operator thus not got the hint, that he can actually insert an equation here.

That's a shame.  I was really hoping it would be more or less intuitive.  If it's not though, than sure, I'd advocate making it simpler.

I'm actually not such a fan of trying to turn the hint test into a substitute manual.  If the original was unintuitive I'd lean more toward making it less complicated and instructional rather than more.  If `"a^2 + b^2"` was confusing I'd prefer to simply reduce it to just `"a + b"` (assuming "a + b" is a long enough string to not cause problems).  If "a + b" is too short then possibly we could use something like `"2*(a+b)"` but `"(a+b)*c^d/e"` just offends my sense of aesthetic :p

As for `"(5+4-3)*2/3"`.  I mean really??  I'm not going to put my foot down about it or anything but seriously?  `"2 + 2"` is the very epitome of a simple truth ;)

You know, we're not trying to write a manual here or anything, it's just a hint to suggest what to enter :) :)

Lineflyer:
> You know, we're not trying to write a manual here or anything, it's just a hint to suggest what to enter :) :)

Thanks, got your point. Maybe I was a little overinspirated.

So how about keeping it simple and having
`a + b` and
`2 + 2` ?

That will show that operators can be used on both, and the first one supports variables as well.

S-Bartfast:
That looks good to me 👍 

By the way, I have created a little 'help' button (for #6664) which opens a HTML page where we can be more detailed about accepted operators and so forth.  I'm struggling to workout exactly what to say in the help page but the button itself can actually be seen in the top left corner of the screenshots I uploaded for #6687.  I'll discuss it more fully in #6664 though when I have a page worth showing.

Lineflyer:
> That looks good to me 

OK, I will change the PR accordingly and merge it.

> By the way, I have created a little 'help' button (for #6664) which opens a HTML page where we can be more detailed about accepted operators and so forth. I'm struggling to workout exactly what to say in the help page but the button itself can actually be seen in the top left corner of the screenshots I uploaded for #6687. I'll discuss it more fully in #6664 though when I have a page worth showing.

Feel free to provide it even if not complete.
I can for sure help trying to complete it or draft some idea/structure for it.

Lineflyer:
As squashing is not allowed from Github UI I will close this PR and do changes directly.

PR #6690, opened 2017-08-16
Fix #6687 - Only revert calculator to 'side-note' layout when width is greater than 1200dp.

### LINKED ISSUE #6687: Show complete variable field in landscape view

S-Bartfast:


Lineflyer:
OK to test

Lineflyer:
You should start your PR with "Fix #6687" to allow Github to trigger autoclosure once this is merged. I changed the title for this one accordingly.

Lineflyer:
Any objection to merge this?

Issue #6699, opened 2017-08-20
Multi digit solutions in WP Calc do not support decimals


Lineflyer:
##### Detailed steps causing the problem:
- Today I performed GC28M8N
- This is a multi cache where you have given stages (A to J) and only need to calculate for the final as follows ( in DDD° MM.MMM' format):
```
N51°(23522+A+C+E+G+I)
E06°(45973+B+D+F+H+J)
```

I tried to enter this as follows in the WP calc:
First I assigned the first decimal digit of lat/lon as A/B in DDD° MM.MMM' format:
```
N051° A 
E006° B
``` 
For A and B I entered:
```
A=23522+a+c+e+g+i
B=45973+b+d+f+h+j
```

##### Actual behavior after performing these steps:
Two problems:
- After entering a-j I get a result which is the correct sum but summed up before the decimal digit only.
- When trying to save the app crashes (debug mode active)

##### Expected behavior after performing these steps:
It should be possible to use such multi digit solutions independently of the decimal separator.

##### Version of c:geo used:
2017.08.19-NB

##### Is the problem reproducible for you?
Yes

S-Bartfast:
As it is the calculator intentionally only works with integers.

I thought about trying to handle decimal expansion results while working on it but ultimately decided against implementing it partly to reduce confusion to the user but mainly because it was too hard to think how to handle all the cases sensibly.  I figured multi's requiring decimal calculations were rare and in virtually all cases there would be a way to make do.

Not being able to read the language I find it difficult to follow exactly what the cache is asking but would it be possible to represent `N51°(23522+A+C+E+G+I)` as `N51°A.B` and define 'A' and 'B' as follows:
`A: (23522+a+c+e+g+i)/1000`
`B: (23522+a+c+e+g+i)%1000`

I acknowledge this isn't as neat as it could be but it should get you through.

Lineflyer:
From reading the calculation I think it would work in this situation, but I dont expect any user to easily find this solution.
I am also thinking about what could be a better solution, but I agree with you that this will not be easy to handle (on top of that it needs also different handling depending on the coordinate format the user selected). So lets keep this open until we have a good idea.

Will change the title in a second.

geocachermgo:
I would also like to have a solution. If float is not wanted maybe a special treatment of __.__A or A_.___ is an option?
Just like the first N is not treated as a variable.
[...]

PR #6700, opened 2017-08-21
Unnecessary check

S-Bartfast:
Removed an unnecessary check that was preventing the calculator from opening when a caches location is unknown.

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesInputDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesInputDialog.java
@@ -570,13 +570,6 @@ public class CoordinatesInputDialog extends DialogFragment {
 
         @Override
         public void onClick(final View v) {
-            if (cacheCoords == null) {
-                final AbstractActivity activity = (AbstractActivity) getActivity();
-                activity.showToast(activity.getString(R.string.err_location_unknown));
-                return;
-            }
-
-
             final CalcState newState = ((CalculateState) getActivity()).fetchCalculatorState();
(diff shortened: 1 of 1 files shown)


Lineflyer:
OK to test

Lineflyer:
Was this check introduced by yourself, or was it in place before?
If it was there before, I wonder why we did it like that.

S-Bartfast:
I wrote it before I really understood what it was.

When I first added the 'CalculateListener' I just copied it from the 'CacheListener' above it and while the 'cacheCoords' is valid for the 'CacheListener' it makes no sense to apply it to the 'CalculateListener'.  In fact it was preventing the calculator from opening for cache GC28M8N (as mentioned in #6699) as for some reason this cache has no cacheCoords.

Lineflyer:
> as for some reason this cache has no cacheCoords.

It should have. I assume you are not logged in and/or have no login information stored for geoaching.com?!

However I think it might be in as you are able to select "Cache Coordinates" while entering coords for a new waypoint. Having these not defined might result in unexpected behaviour.

S-Bartfast:
> ...as you are able to select "Cache Coordinates" while entering coords for a new waypoint.

I can't actually:
image

It is true that I am not logged in but I haven't been logged in on most of my emulated devices.  The other caches do seem to have 'cacheCoords' defined though so I'm not sure why this one doesn't.

No idea what's going on but in any event I can't see why 'cacheCoords' would need to be assigned for the calculator to be made available, so I got rid of the check, and it seems to work fine for me ;)

Lineflyer:
Lets test then.

PR #6728, opened 2017-09-08
Fix #6724, Description of GC15GV4 unreadable

### LINKED ISSUE #6724: Description of GC15GV4 unreadable
Some weeks ago I stored GC15GV4 on my device using an older nightly of beginning of August.
I noticed no problem reading the description.

Meanwhile I use a newer nightly 2017.09.03 and noticed the still stored cache had now black text on black background. 
As the caches was not refresehed since original storing it, it might be the visualization might have a bug in newer version.

I might be wrong though.
Can someone check this cache with the current market release?

rsudev:
add detecting problematic text colors in style attributes

As per our normal bug-fix guidelines I based this on release and first propose a merge to master, so that it can be checked on the nightlies but also be applied to release if necessary without hassles.


kumy:
Jenkins retest this please

Lineflyer:
@rsudev Looks fine to me. Tested on some caches. We should merge this to release as you prepared.

PR #6734, opened 2017-09-17
fix #4137 log picture for oc

### LINKED ISSUE #4137: Support log pictures for OC
Users wish:
Support uploading of a picture with the log like for gc.com.

Don't know if OKAPI supports it.

pstorch:
As I was working on the new logging page for GC, I saw the (sad) comment on this issue and I was wandering how image upload works for OC in comparison. Much simpler ;)

The only thing missing right now is the check of the server capabilities on max image size and dimensions. An unscaled image upload was working from my device to opencaching.de.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1602,5 +1602,6 @@
     <string name="equation_error_result">*</string>
     <string name="lat_point_min_format">N DDD° MM.PPP</string>
     <string name="lon_point_min_format">E DDD° MM.PPP</string>
+    <string name="err_select_logimage_upload_size">Image too big, please select a smaler scale.</string>
 
 </resources>
--- a/main/src/cgeo/geocaching/ImageSelectActivity.java
+++ b/main/src/cgeo/geocaching/ImageSelectActivity.java
@@ -50,6 +50,8 @@ public class ImageSelectActivity extends AbstractActionBarActivity {
 
     private static final String SAVED_STATE_IMAGE = "cgeo.geocaching.saved_state_image";
     private static final String SAVED_STATE_IMAGE_SCALE = "cgeo.geocaching.saved_state_image_scale";
+    private static final String SAVED_STATE_MAX_IMAGE_UPLOAD_SIZE = "cgeo.geocaching.saved_state_max_image_upload_size";
+    private static final String SAVED_STATE_MAX_IMAGE_PIXELS = "cgeo.geocaching.saved_state_max_image_pixels";
 
     private static final int SELECT_NEW_IMAGE = 1;
     private static final int SELECT_STORED_IMAGE = 2;
@@ -57,6 +59,8 @@ public class ImageSelectActivity extends AbstractActionBarActivity {
     // Data to be saved while reconfiguring
     private Image image;
     private int scaleChoiceIndex;
+    private long maxImagePixels;
+    private long maxImageUploadSize;
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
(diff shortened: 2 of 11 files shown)


pstorch:
@following5 how critical is the check of the max upload size of an image? Can you test with various OC installations?

following5:
Yeah :-)

> how critical is the check of the max upload size of an image?

If an image is larger than `image_rcmd_max_pixels`, OKAPI will try to downscale it. If the dimensions are too huge to handle it, you will get a [4xx error response](https://www.opencaching.de/okapi/introduction.html#errors).

If an image is larger than `image_max_upload_size`, you will also get a 4xx error response.

pstorch:
Ok, will implement the validation of the image size restrictions.

following5:
>If an image is larger than `image_max_upload_size`, you will also get a 4xx error response.

Sorry, I got this wrong. If the site is configured properly, you will not get a 4xx but a 200 response, with `success`=false and `message` explaining why the image was rejected. (If it is not configured properly, you may still get a 4xx response directly from the Apache.)

If file size is ok but `image_rcmd_max_pixels` is exceeded, it will allways be a 200 response with with `success`=false and explanation `message`.

Those  limits in `services/apisrv/installation` are are provided to help developers in optimizing their uploads: Avoid transferring lots of megabytes though it is known in advance that they will be rejected.

pstorch:
> If file size is ok but image_rcmd_max_pixels is exceeded, it will allways be a 200 response with with success=false and explanation message.

Sure? I uploaded an image exceeding `image_rcmd_max_pixels` but got a `success=true` response. Only the image on the server was downscaled silently. (tested on opencaching.de)

To be on the safe side I validate both restrictions and deny the upload, if one of them is exceeded.

following5:
> Sure? I uploaded an image exceeding image_rcmd_max_pixels but got a success=true response. Only the image on the server was downscaled silently. (tested on opencaching.de)

Sorry again. I should know better, because I wrote the OKAPI image handling code. :-$

`image_rcmd_max_pixel` indicates the maximum size which is stored at the OC site. OKAPI will try to automatically downscale larger pictures. If the image is that huge that OKAPI cannot handle it (not enough PHP memory to process it), then it will be rejected with HTTP 2xx and `success`=false.

c:geo probably has much more memory available than OKAPI methods, so besides of saving upload bandwidth, it still makes sense to do the downscaling locally.

pstorch:
retest this please

pstorch:
retest this please

pstorch:
retest this please

pstorch:
retest this please

pstorch:
Ok, let's test it in the nightlies.

@following5 can you use the next nightly to do some testing?

following5:
Unfortunately, I don't have any device where I could test c:geo. I am just Okapi and Opencaching.de developer and try to support all the Opencaching users out there, whatever applications they use.

But I have posted a request for testing in the [OC.de forum](https://forum.opencaching.de/index.php?topic=4915).

pstorch:
Ok, so what is the expected behavior of an API client regarding `image_rcmd_max_pixels`? Downscale automatically, give a warning to the user or should we just ignore it?

following5:
Best practice is to automatically downscale, second-best to ignore it.

Note that all Okapi submitting methods - like submitting logs or images - return a boolean `success` value and a localized plaintext `message`. Best practice is to evaluate `success` and display the `message` string to the user if `success`=false. But I am not sure if that fits into the c:geo model of calling API methods.

following5:
The message e.g. may tell that the image was too large to process, if the image file was broken, the image format could not be recognized, the caption was missing and so on.

So the overall best algorithm is:
* Test for `image_max_upload_size` and abort if the image file is too large.
* Either downscale to ` image_rcmd_max_pixels` or ignore ` image_rcmd_max_pixels`.
* upload the image
* evaluate `success`
* display `message` if `success` == false

(We should add this to Okapi docs!)

But if evaluating the Okapi result is not feasible to you, noone will complain if you ignore it. ;)

following5:
[Here](https://forum.opencaching.de/index.php?topic=4915.msg61408#msg61408) is another positive test. Users are very happy about this new feature!

PR #6737, opened 2017-09-23
Fix #6736: Parse OC coordinates as double values

### LINKED ISSUE #3736: Wrong image when opening a spoiler with Photos
When clicking on a spoiler image and opening it with the Photos app, another image is displayed (in my case a spoiler from a previously visited cache). Sometimes the other image is displayed during less than 1 second and the actual spoiler is displayed after that but darker than usual.

### LINKED ISSUE #6736: parse error for coordinates on integer longitude/lattitude
Try to store OC0054 or OC533E. Both throw a parse exception. The commonality is that they use coordinates exactly on full longitude/lattitude.

https://forum.opencaching.de/index.php?topic=4912.msg61348#msg61348

UniQP:
Parsing the coordinates is not necessary because we already know their
format in advance. Futhermore, directly parsing the coordinates as
double values also support integer values, which would not be recognized
as valid latitude/longitude by the coordinate parser.

### DIFF
--- a/main/res/values/changelog_release.xml
+++ b/main/res/values/changelog_release.xml
@@ -19,6 +19,7 @@
     · Fix: TLS v1.2 Support for Android 4.1 - 4.4 devices\n
     · Fix: Improve sort of events\n
     · Fix: Raise personal note character limit\n
+    · Fix: Avoid failure when parsing OC coordinates\n
     \n
     \n
 </string>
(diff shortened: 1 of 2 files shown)


pstorch:
I've tested it and it works.
I'm just thinking about if it is worth merging it in `release` first.
I've tried to change the base of the PR to `release`, but then other commits are pulled in, so I changed it back. If we want to have it in `release` first, then I guess it has to be rebased first.

UniQP:
I will change the base to `release`.

UniQP:
done

pstorch:
Thanks

PR #6745, opened 2017-10-02
fix #6744 attach oc log image

### LINKED ISSUE #6744: Log picture not attached to log immediately after posting
- Log an OC cache incl. picture 
- After log is posted, swipe to logbook

No picture shown attached to the log

Picture should be shown

##### Version of c:geo used:
2017.09.30-NB

Only tested once

##### Other comments :
Compare to GC picture upload with c:geo. Picture is shown there after log has been posted

pstorch:
It assumes the following pattern to derive the image URL from the `image_uuid` we get from the API call: 
`https://www.opencaching.de/images/uploads/{image_uuid}.jpg`

At least it works for opencaching.de, but I don't know if we can use this pattern for all OC plattforms.
Maybe @following5 can give some input.

### DIFF
--- a/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -364,7 +364,7 @@ final class OkapiClient {
             }
 
             if (data.get("success").asBoolean()) {
-                return new ImageResult(StatusCode.NO_ERROR, "");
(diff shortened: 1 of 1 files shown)


pstorch:
No, another way should be used, see https://github.com/cgeo/cgeo/issues/6744#issuecomment-333666024

following5:
> At least it works for opencaching.de, but I don't know if we can use this pattern for all OC plattforms.

It currently works on all OC sites, but it's not guaranteed to work in the future.

following5:
> It currently works on all OC sites

... if the image is in JPG format ...  Okapi also accepts and stores PNG and GIF.

PR #6746, opened 2017-10-02
fix #6744 attach oc log image

### LINKED ISSUE #6744: Log picture not attached to log immediately after posting

pstorch:
Until we have the the `image_url` in the response of the post image call as mentioned in https://github.com/cgeo/cgeo/issues/6744#issuecomment-333675587, we have to retrieve the log entry and extract the image url from there.


following5:
@pstorch
Meanwhile all OC sites return the image url directly from services/images/add. So you may simplify this.

Issue #6654, opened 2017-07-26
Only unstored caches on live map view are stored in selected list


MrYaz:
##### Detailed steps causing the problem:
- Zoom into an area on the live map where you have both stored and unstored caches
![image](https://user-images.githubusercontent.com/1811963/28647520-088b4bb2-7268-11e7-9489-4944c59d90e0.png)

- Choose Store Offline in menu
![image](https://user-images.githubusercontent.com/1811963/28647539-2659a580-7268-11e7-999c-587ba290f74c.png)

- Create a new list or select an existing list

##### Actual behavior after performing these steps:
Only the caches that are not already stored in any other lists are added to the new list.
The cache with the offline log is also added to the new list, even if it has a "stored" icon

![image](https://user-images.githubusercontent.com/1811963/28647611-78120232-7268-11e7-82a8-e2d6758b8c82.png)

##### Expected behavior after performing these steps:
All caches in the map view should be added to the new list.
Since several lists can co-exist nowadays the already stored caches should be added to the new list as well.

##### Version of c:geo used:
2017.07.26-NB

##### Is the problem reproducible for you?
Yes

Lineflyer:
I somehow agree, that the behaviour should be changed.
I wouldn't call it a bug, so I labeled it as "Feature Request".

PR #6749, opened 2017-10-06
#6654 Only unstored caches on live map view are stored in selected list

superspindel:
Removed hasUnsavedCaches check from store offline menu creation. Removed isOffline check for caches when storing from viewport.

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -588,7 +588,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory {
 
 
             final Set<String> geocodesInViewport = getGeocodesForCachesInViewport();
-            menu.findItem(R.id.menu_store_caches).setVisible(!isLoading() && CollectionUtils.isNotEmpty(geocodesInViewport) && new SearchResult(geocodesInViewport).hasUnsavedCaches());
+            menu.findItem(R.id.menu_store_caches).setVisible(!isLoading() && CollectionUtils.isNotEmpty(geocodesInViewport));
 
             menu.findItem(R.id.menu_mycaches_mode).setChecked(Settings.isExcludeMyCaches());
             menu.findItem(R.id.menu_disabled_mode).setChecked(Settings.isExcludeDisabledCaches());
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -671,13 +671,7 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
                 if (!isLoading()) {
                     final Set<String> geocodesInViewport = getGeocodesForCachesInViewport();
-                    final List<String> geocodes = new ArrayList<>();
-
-                    for (final String geocode : geocodesInViewport) {
-                        if (!DataStore.isOffline(geocode, null)) {
-                            geocodes.add(geocode);
-                        }
-                    }
+                    final List<String> geocodes = new ArrayList<>(geocodesInViewport);
```
UniQP: Do we need this copy to prevent concurrent modifications on `geocodesInViewport`? Otherwise, we can simply reuse `geocodesInViewport` by adapting the signature of `storeCaches` (`List<String>` → `Collection<String>`).

pstorch:
ok to test

pstorch:
ok to test

pstorch:
Jenkins still doesn't listen to me 🤔

Lineflyer:
test this please

main/src/cgeo/geocaching/maps/CGeoMap.java
superspindel: Can we change the commit message on this pull request to that message or is that set when the pull request is instantiated? 

As for the second question, my guess would be that we do not need that copy, and a redesign of the storeCaches would suffice since it's only use is from this method.

UniQP: If you use `git commit --amend` instead of `git commit` the changes will be added to the current commit and you can change the commit message. Thus, `git commit --amend` without any changes allows you to change the commit message of your current commit.

If you already have multiple commits, you can use interactive rebase to combine them, see https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history.

pstorch:
I've tested the change and it solves the issue :+1: 
But I would also prefer to get rid of the second List and change the `storeCaches` method and the `LoadDetails` inner class to use `Set<String> geocodes`.

superspindel:
Where these changes in line with what you were thinking?

pstorch:
@superspindel just one thing, I didn't realize this morning. You have two commits now. Can you squash them into one?

superspindel:
Oh yeah sorry about that, fixed now.

pstorch:
Perfect, thanks :+1:

PR #6770, opened 2017-10-14
fix #6769 use `image_url` from OC postLog result

### LINKED ISSUE #6769: All OC sites return the image url directly now
As mentioned by @following5 all OC sites return the image url (return value `image_url`) directly from services/images/add. 
So we can simplify our code and omit the second HTTP call.

https://www.opencaching.de/okapi/services/logs/images/add.html

pstorch:


pstorch:
retest this please

PR #6773, opened 2017-10-15
Fixed #6772: Implement Tri-State Calculator Buttons

### LINKED ISSUE #6772: Implement Tri-State Calculator Buttons
Just making this ticket as a way to track the suggestion that @pstorch made in #6704 that:

>I'm thinking about if we should change the button modes. With a short press it toggles between variable and original value. Maybe it should also toggle through the "space" case. So that this feature is more accessible.

If that comment doesn't quite make sense out of context what @pstorch is suggesting is:
Perhaps rather than simply having the calculator buttons toggle between 'Input-value" and "Auto-char" (with the blank option being available by entering SPACE through a long press) that the calculator buttons have three states, being "Input-value", "Auto-char" and "Blank".  Note that the "Custom-char" functionality will still be available through a long press, it just won't be the way you "blank-out" buttons.

S-Bartfast:


main/res/values/strings.xml
```diff
@@ -239,7 +239,7 @@
     <string name="warn_rendertheme_missing">Map theme not found.</string>
     <string name="warn_pocket_query_select">No Pocket query selected.</string>
     <string name="warn_no_pocket_query_found">No Pocket query found online.</string>
-    <string name="warn_invalid_character">Character must be a LETTER, DIGIT or SPACE</string>
+    <string name="warn_invalid_character">Character must be a LETTER or a DIGIT</string>
```
UniQP: Maybe not directly related to this PR, but why are LETTER and DIGIT all upper case?

S-Bartfast: I guess it's just adding emphasis.

Originally it read:
>Character must be in the range  "A-Z" or "0-9"

But then when the space was added it was harder to describe, as in:
>Character must be in the range "A-Z", "0-9" or the " " character

I tried saying:
> space (" ")

and various other things but eventually decided it just looked best with capital letters:
>Character must be a LETTER, DIGIT or SPACE

Now that the space has been removed I did nearly add the word '_either_' to the phrase, as in:
>Character must be either a LETTER or a DIGIT

but didn't end up bothering.

Now that the "space" has been removed I could go back to the original "A-Z" or "0-9" but I actually thought it was good with the capitalised words.

What would you prefer to see?

rsudev: A-Z (if that really is the definition) is less ambiguous than LETTER (which in some places might end up being äöü or even åđžć ...

UniQP: I would prefer to allow all characters accepted by `Character.isLetterOrDigit(...)` and simply use `letter` and `digit` in the message. Is that feasible?

pstorch:
I've tested the solution and it works as expected. Thanks, @S-Bartfast .
Regarding the `letter` and `digit` discussion: at the moment it is ambiguous. Just tried `ö`, which is a letter, and got the above error message. I personally think we should restrict the variable letters to A-Z. If we allow all unicode letters from all languages, then we might run into sorting issues again. So, to avoid that and make things simple, I would suggest A-Z letters only. What do you think?

pstorch:
retest this please

main/res/values/strings.xml
S-Bartfast: `Character.isLetterOrDigit(...)` could be implemented but some thought would have to go into how the 'auto-increment' feature works.  As in if a variable is called `Å`, what should the next button be called.  With 'A' - 'Z' we can just basically use `nextLetter = thisLetter +1` but for broader character sets it's not so clear.

As it is 'Å' has the code "U+00C5", and "U+00C6" is the letter code for "Æ", so does that mean 'Å' should be followed by 'Æ'.  Then 'È' is followed by 'É', which is followed by 'Ê', then by 'Ë'.  Where as 'Ï' is followed by 'Ð', which is followed by 'Ñ'.  If there is a consistent way to solve the 'nest letter' problem I'd be more than happy to use the 'isLetterOrDigit(...)' method.

At the moment I'm working on implementing a solution where by non-standard letter are treated in much the same way as manually specified numbers are treated, as in the alphabetic sequence essentially 'skips' them:
> `A` > `B` > *`Ë`* > `C`

Just as:> `A` > `B` > *`Ë`* > `C`

PR #6778, opened 2017-10-16
Fixed #6775: Allow non-standard letters to be use in equations

### LINKED ISSUE #6775: Allow non-standard letters to be use in equations
This is just being created in order to track the discussion in #6773 about allowing non-standard letters to be used in equations.

@rsudev commented:
>A-Z (if that really is the definition) is less ambiguous than LETTER (which in some places might end up being äöü or even åđžć ...

and @UniQP said:
>I would prefer to allow all characters accepted by Character.isLetterOrDigit(...) and simply use letter and digit in the message. Is that feasible?

And later on @pstorch commented:
>Regarding the letter and digit discussion: at the moment it is ambiguous. Just tried ö, which is a letter, and got the above error message. I personally think we should restrict the variable letters to A-Z. If we allow all unicode letters from all languages, then we might run into sorting issues again. So, to avoid that and make things simple, I would suggest A-Z letters only. What do you think?

S-Bartfast:
Fixed #6775: Allow non-standard letters to be use in equations

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -127,17 +127,23 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
      * Class used for checking that a value is with in a given range.
      * This is used to check for upper-case an lower-case letters.
      */
-    private class Range {
-        final int min;
-        final int max;
+    private class CaseCheck {
+        final boolean useUpper;
 
-        Range(final int min, final int max) {
-            this.min = min;
-            this.max = max;
+        CaseCheck(final boolean upper) {
+            useUpper = upper;
         }
 
-        boolean check(final int ch) {
-            return min <= ch && ch <= max;
+        boolean check(final char ch) {
+
+            boolean returnValue = Character.isLetterOrDigit(ch);
+            if (useUpper) {
+                returnValue &= Character.isUpperCase(ch);
+            } else {
+                returnValue &= Character.isLowerCase(ch);
+            }
+
+            return returnValue;
         }
     }
 
@@ -580,7 +586,6 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     private boolean areCurrentCoordinatesValid() {
         // convert text to geopoint
         final Geopoint current;
-
         final String lat = getLatResult();
         final String lon = getLonResult();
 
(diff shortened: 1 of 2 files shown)


Lineflyer:
retest this please

pstorch:
@S-Bartfast there is a conflict now. And please a more descriptive commit message.

S-Bartfast:
Hmm, I'm not really sure where the conflicts have come from but I've resolved the conflicts and rebased onto master but the change doesn't seem to be coming through here.  I may have to issue another pull request.

S-Bartfast:
Okay, I think it's good now.
That was a bit confusing??

S-Bartfast:
Okay, think I've got it!

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -580,17 +586,34 @@ private void setButtonInputValuesFromGP() {
...
-
         final String lat = getLatResult();
         final String lon = getLonResult();
 
         if (lat.contains("_") || lat.contains("*") || lon.contains("_") || lon.contains("*")) {
             return false;
         }
 
-        try {
-            current = new Geopoint(lat, lon);
```
pstorch: @S-Bartfast wasn't this just changed the other way around?

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -843,7 +866,7 @@ private String getValues(final List<CalculateButton> buttons, final int degDigit
         for (final EditButton button : buttons) {
             // Remove inactive and blank digits from result
             if (button.getVisibility() == View.VISIBLE) {
-                if (button.getLabel() == CalculateButton.ButtonData.BLANK) {
```
pstorch: Same here

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -1100,22 +1123,24 @@ private CalculatorVariable getVariable(final char name, final List<CalculatorVar
...
-                                                   final Range range,
+                                                   final CaseCheck theCase,
                                                    final String hintText,
                                                    final TextWatcher textWatcher) {
         final List<CalculatorVariable> returnList = new ArrayList<>();
 
         final char[] sortedVariables = variableNames.toCharArray();
         Arrays.sort(sortedVariables);
 
-        for (final char ch : sortedVariables) {
```
pstorch: also here

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Yeah, I've definitely managed to get something "arse about face" as we say in Australia.

S-Bartfast: Reverted.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Reverted.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
S-Bartfast: Reverted.

pstorch: Hmm, I still see the wrong version of the loop :thinking:

S-Bartfast: Okay, that should get it.

pstorch:
@S-Bartfast thanks :+1:

PR #6779, opened 2017-10-16
Fixed #6776: Crashed on Rotation

### LINKED ISSUE #6776: Calculator Crashed on Screen Rotation
This was probably introduced with #6759 and seems to be caused by a change to the 'Plain Format' text field.  The change gets triggered sometime between when the 'textListener' is added in `onCreateView(...)` and when the calculator's state is loaded in `onResume()`.

The actual crash is caused by accessing the `currentFormat` variable before it is initialised but I can't for the life of me workout what is causing the text to be changed prior to the `onResume()` call.  In any event simply protecting the reference to `currentFormat` with "if null" checks alleviates the problem but I would have preferred to isolate the cause of the text changed event in the first place but can't seem to find it.

S-Bartfast:
Fixed #6776: Addressed the issue by guarding with 'null checks' rather than identifying the source, but you know, it's a solution.

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -584,7 +584,7 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         final String lat = getLatResult();
         final String lon = getLonResult();
 
-        if (lat.contains("_") || lat.contains("*") || lon.contains("_") || lon.contains("*")) {
+        if (currentFormat == null || lat.contains("_") || lat.contains("*") || lon.contains("_") || lon.contains("*")) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


pstorch:
I've verified the fix successfully. Only one question: can you mention the issue number in the commit message?
If we later browse history and search through the commits, there is no indication which issue was fixed.

pstorch:
@S-Bartfast maybe I should have mentioned you, so that you get a notification about my comment?

S-Bartfast:
> @S-Bartfast maybe I should have mentioned you, so that you get a notification about my comment?

Sorry, I did get a notification about you original comment but was waiting for the rebase once #6781 was accepted rather then changing the history twice (Force-pushes still kind of scare me ;))

I did amend the comment at the top of this thread when I  received the original notification though :)

pstorch:
@S-Bartfast I meant the commit message "Addressed the issue by guarding with 'null checks' rather than identi..."

S-Bartfast:
Okay, I'm about to rebase this one.

So just so I understand, there were two issues with the commit message, yeah?  Firstly that the issue number was missing from the commit message and secondly that the message was different to the branch name.  Is that right?

pstorch:
@S-Bartfast I just wanted to have the issue number and some description in the commit message, which explains what has changed or what was addressed. It doesn't need to be identical to the issue title. And have one of the Github keywords (fixed, fixes, closes,...) in front of the issue number, to automatically close the issue with the merge to master.
Thanks for the update.

PR #6780, opened 2017-10-16
Fixed #6772: Implement Tri-State Calculator Buttons

### LINKED ISSUE #6772: Implement Tri-State Calculator Buttons

S-Bartfast:


pstorch:
@S-Bartfast thanks :+1:

Issue #6781, opened 2017-10-16
Calculator can crash on fast click on coordinate buttons


pstorch:
I could cause an exception by fast clicking on the coordinate buttons:

```
10-16 21:11:28.798 10999-10999/cgeo.geocaching E/AndroidRuntime: FATAL EXCEPTION: main
  Process: cgeo.geocaching, PID: 10999
  java.lang.ArrayIndexOutOfBoundsException: length=3; index=3
  at cgeo.geocaching.ui.dialog.CoordinatesCalculateDialog.areCurrentCoordinatesValid(CoordinatesCalculateDialog.java:600)
  at cgeo.geocaching.ui.dialog.CoordinatesCalculateDialog.updateResult(CoordinatesCalculateDialog.java:961)
  at cgeo.geocaching.ui.dialog.CoordinatesCalculateDialog.onClickCompleteCallback(CoordinatesCalculateDialog.java:148)
  at cgeo.geocaching.ui.EditButton.handleClick(EditButton.java:125)
  at cgeo.geocaching.ui.CalculateButton.handleClick(CalculateButton.java:173)
  at cgeo.geocaching.ui.CalculateButton$CoordDigitClickListener.onClick(CalculateButton.java:156)
  at android.view.View.performClick(View.java:5669)
  at android.view.View$PerformClick.run(View.java:22549)
  at android.os.Handler.handleCallback(Handler.java:751)
  at android.os.Handler.dispatchMessage(Handler.java:95)
  at android.os.Looper.loop(Looper.java:154)
  at android.app.ActivityThread.main(ActivityThread.java:6321)
  at java.lang.reflect.Method.invoke(Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

S-Bartfast:
I'm not able to re-produce this error myself, either on my own device nor an emulated one.

Were you experiencing this issue with the build featuring the "tri-state buttons" or the previous "two-state" ones?

pstorch:
Now I'm on latest master commit including "tri-state buttons" and I can't reproduce either.
So let's close here.

S-Bartfast:
[...]

Issue #6716, opened 2017-08-30
Managing ambiguous results in WP Calc where trailing/leading zeros are involved 


Matze2:
Here is a use case where the decision for trailing zeros is not the expected one in my opinion.

Two problems with this scenario:
- at least in this case leading zeros would be the expected interpretation of that number
- the result area below does not show the final interpretation, this is only outside of the calculator when the page is confirmed

![screenshot_2017-08-30-07-04-56](https://user-images.githubusercontent.com/12542802/29856396-cc396350-8d52-11e7-901b-7122d19ffb8f.png)
![screenshot_2017-08-30-07-02-00](https://user-images.githubusercontent.com/12542802/29856402-d257fb0c-8d52-11e7-9403-7aceb4bff999.png)

Probably interpretation rules are needed. Some examples:
If A should be mapped to DDD or DD then leading zeros seems to me more correct than trailing zeros.
If AB should be mapped to DDD then the solution is only correct if AB is exactly three digits, otherwise it could be 0AB or A0B or with trailing zeros AB0 which I would really call ambiguous. :)
We should also warn if the result is bigger then the expected pattern, e.g. if A should be mapped to DDD and A > 999 (negative is anyway wrong). 

The primary issue for now is that the result area **inside** the calculator should already display the leading/trailing zeros.

Ah - and of course this means that the "Plain" mode also needs a pattern like "Plain DD MM.MMM" if the button issue cannot be solved.

S-Bartfast:
Hmm, interesting.
So the calculator is indeed zero padding with _trailing_ zeros after the decimal point after all.

I take it the result you expected to see would be:
`N 38° 31,A`; `A = 3` -> `N 38° 31.003`
`N 011° 29,B`; `A = 30` -> `N 38° 31.030`

This incidentally _is_ the way the 'button-formats' behave.  The 'incorrect' result is displayed in the calculator (being "N 38° 31.3" and "N 38° 31.30") however the _leading_ zero is inserted once the 'tick' button is clicked.

Note that the regular coordinate input dialogue also inserts _trailing_ zeros when in 'Plain' format, which isn't surprising as the 'plain' format parser employed by the calculator was lifted directly from the regular coordinate input dialogue.

Matze2:
Yes during writing the issue I felt that this is a special problem with "Plain" mode.
My request would only be valid if there is would be a "button-less" mode, but with a given coordinate format.
[...]

PR #6785, opened 2017-10-19
#6781 Fixed: Blank section crash

S-Bartfast:
#6781 Fixed: Blank section crash

### DIFF
--- a/main/src/cgeo/geocaching/location/GeopointFormatter.java
+++ b/main/src/cgeo/geocaching/location/GeopointFormatter.java
@@ -38,13 +38,13 @@ public class GeopointFormatter {
         /** Example: "-0.123456" (unlocalized latitude) */
         LAT_DECDEGREE_RAW,
 
-        /** Example: "N 10° 12.345" */
+        /** Example: "N 01° 02.034" */
         LAT_DECMINUTE,
 
         /** Example: "N 10 12,345" */
         LAT_DECMINUTE_RAW,
 
-        /** Example: "N 10° 12' 34.567" */
+        /** Example: "n 01° 02' 03.045" */
         LAT_DECMINSEC,
 
         /** Example: "W 0.123456°" */
@@ -53,13 +53,13 @@ public class GeopointFormatter {
         /** Example: "-0.123456" (unlocalized longitude) */
         LON_DECDEGREE_RAW,
 
-        /** Example: "W 5° 12.345" */
+        /** Example: "W 001° 02.034" */
         LON_DECMINUTE,
 
         /** Example: "W 5 12,345" */
         LON_DECMINUTE_RAW,
 
-        /** Example: "W 5° 12' 34.567" */
+        /** Example: "W 001° 02' 03.045" */
         LON_DECMINSEC,
 
         /** Example: "32U E 549996 N 5600860" */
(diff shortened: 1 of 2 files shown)


pstorch:
retest this please

Lineflyer:
retest this please

Lineflyer:
I don't understand why Codacy has issues with this PR...and I don't even see issues in their review report.

pstorch:
Interesting. Codacy reports 2 new issues, but the issues tab on the report is empty. But if you open the changed file in the codacy report you see 9 issues and 2 new. When I interpret it correctly it's about the missing space after the comma in lines 585 and 586.

pstorch:
I've sent them a message about this report. Let's see.

S-Bartfast:
Huh, just realised there was indeed a couple of missing spaces in there.  At first I thought it was being silly about the comma in the string argument `","` so just ignored it but now i see it was actually complaining about the comma before the empty string `,""` which I have now fixed.

S-Bartfast:
At first I thought it was being silly about the comma in the string argument `","` so just ignored it but now i see it was actually complaining about the comma before the empty string `,""` which I have now fixed.

S-Bartfast:
Opps, didn't mean to do that. How do I cancel the close?

S-Bartfast:
Okay, got it.

I did click just the "Comment" button first but it took a while to react and I thought nothing happened so I then tried the other one ("Close and comment").  I now see how easy it is for people to accidentally close issues without meaning to 😊

Bananeweizen:
I cannot judge the correctness of the change. However, I'm wondering why we are having the validation code in this class at all. Wouldn't it be more useful to instead invoke one of the parsers inside the GeopointParser class instead? Those have been finetuned a lot and also deal with corner cases. Especially if we already know which format is currently active in this dialog, we could easily invoke the matching parser (MinDecParser etc.) instead of going through the Geopoint constructor based on 6 to 8 String arguments.

S-Bartfast:
>I'm wondering why we are having the validation code in this class at all.

Well, it's there because to begin with I really didn't have a good handle on how things worked so I essentially just copied it directly out of the "Add Waypoint" dialog.  At the time I new it wasn't a 'good' move but didn't want to rearrange anything external to my class.  Even just doing preliminary testing made me aware the parser has many short comings and figured a proper over hall would no doubt be necessary at some stage.  I guess that time is about to come...

>Wouldn't it be more useful to instead invoke one of the parsers inside the GeopointParser class instead?

Almost certainly.  I've tried to follow exactly what goes on inside that parser a few times but each time I've just got so completely lost that I've just thrown my hands up in despair.  I'll create a separate ticket to address parsing more generally.

S-Bartfast:
Okay, I just implemented a better solution which exclusively uses the 'Plain' format parser.

This solution is a lot neater and actually works quite well however, as discussed in #6716, it has the unfortunate side effect of zero padding groups after the decimal point with trailing (rather than leading) zeros:
image

S-Bartfast:
Okay, a fix for #6716 has been submitted, now we'll just have to work out whether it was a "bug or a feature" ;)

pstorch:
@S-Bartfast thanks

PR #6786, opened 2017-10-19
fix #6784 more than 100 waypoints

### LINKED ISSUE #6784: Add own waypoint fails when geocache has already >100 waypoints defined
From german geocaching forum:
https://geoclub.de/forum/viewtopic.php?f=155&t=67870&start=700#p1276867

- open https://coord.info/GC6HH0B 
- goto waypoints
- add an own wapoint
- c:geo crashes

Maybe related to #6243 

I could reproduce the bug, but couldn't yet look into the logs what happens.

pstorch:
Use also characters as waypoint prefixes.

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -88,7 +88,8 @@ import org.apache.commons.lang3.StringUtils;
  */
 public class Geocache implements IWaypoint {
 
-    private static final int OWN_WP_PREFIX_OFFSET = 17;
+    private static final String WP_PREFIX_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
     private long updated = 0;
     private long detailedUpdate = 0;
     private long visitedDate = 0;
@@ -1388,7 +1389,7 @@ public class Geocache implements IWaypoint {
     }
 
     /*
-     * Assigns a unique two-digit (compatibility with gc.com)
+     * Assigns a unique two-character (compatibility with gc.com)
      * prefix within the scope of this cache.
      */
     private void assignUniquePrefix(final Waypoint waypoint) {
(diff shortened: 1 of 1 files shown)


pstorch:
retest this please

kumy:
retest this please

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1398,12 +1399,20 @@ private void assignUniquePrefix(final Waypoint waypoint) {
...
 
-        for (int i = OWN_WP_PREFIX_OFFSET; i < 100; i++) {
-            final String prefixCandidate = String.valueOf(i);
-            if (!assignedPrefixes.contains(prefixCandidate)) {
-                waypoint.setPrefix(prefixCandidate);
-                return;
+        // start with "17"
+        int a = 0;
+        int b = 6;
+        while (a < WP_PREFIX_CHARS.length()) {
```
UniQP: I would prefer to keep the style of the original loop. Rough idea:
```
final int length = WP_PREFIX_CHARS.length();
for (int i = OWN_WP_PREFIX_OFFSET; i < length * length; i++) {
final int a = i / length;
final int b = i % length;
...
```

pstorch: Interesting, most of the time, when I'm unsure with some lines of code, someone will comment on it :wink: 
I'll try your idea, thanks.

Still wondering where this `OWN_WP_PREFIX_OFFSET = 17` comes from. Does anyone know?

pstorch:
I've changed the loop as @UniQP suggested, but without the `OWN_WP_PREFIX_OFFSET`, because we count differently now.
00 - 09, 0A-0Z instead of 00-99, ...
So the initial 17 doesn't make much sense anymore.

pstorch:
retest this please

PR #6799, opened 2017-10-25
Remove Magic Char

S-Bartfast:
Just a little housekeeping


kumy:
jenkins retest this please

pstorch:
@S-Bartfast I don't want to be the one who is picky all the time, but I don't like the commit message. Can't you use the PR title `Remove Magic Char` also as the commit message? That would be great.

S-Bartfast:
> @S-Bartfast I don't want to be the one who is picky all the time, but I don't like the commit message. Can't you use the PR title Remove Magic Char also as the commit message? That would be great.

No problem.  And no offence is taken, in fact I prefer you being picky.  It's the best way to learn ;)

pstorch:
@S-Bartfast thanks a lot ;)

PR #6800, opened 2017-10-25
Fixed #6716: Leading Zero Padding

### LINKED ISSUE #6716: Managing ambiguous results in WP Calc where trailing/leading zeros are involved

S-Bartfast:
Fixed #6716: Leading Zero Padding

### DIFF
--- a/main/res/values/changelog_master.xml
+++ b/main/res/values/changelog_master.xml
@@ -2,22 +2,5 @@
 <resources>
   <!-- changelog for the master branch -->
   <string name="changelog_master" translatable="false">
-    <b>Next feature release:</b>\n
-    · New: Coordinate calculator (New Waypoint - Calculate Coordinates)\n
-    · New: Log image upload for Opencaching\n
-    · New: Detect certitude as checker and support opening it via menu\n
-    · New: Show as list menu now only active if more than one cache is on map\n
-    · New: Support OpenCaching GPX format extensions\n
-    · Fix: Improve parsing of home coordinates\n
-    · Fix: Disabled caches might be shown with truncated title\n
-    · Fix: Some OC cache URL not detected correctly\n
-    · Fix: Some possible crashes on map\n
-    · Fix: Search caches by users with a \'+\' in the name\n
-    · Fix: Correctly display pictures with relative URLS\n
-    · Fix: Disable phone number linking in personal note field\n
-    · Fix: Possible crash when starting directory selection\n
-    · Fix: Improved detection of dark text to allow contrast improvement\n
-    · Fix: Improve creation of nomedia file in geocache data directories\n
-     \n
   </string>
 </resources>


S-Bartfast:
Is that one lien 'while' statement going to be okay to leave:
>`while (pts.length() < 3) { pts = "0".concat(pts); }`

Of course I can break it out into three lines but I actually think it's neat as it is ;)

pstorch:
@S-Bartfast 

> Is that one line 'while' statement going to be okay to leave:
> 
> `while (pts.length() < 3) { pts = "0".concat(pts); }`
> Of course I can break it out into three lines but I actually think it's neat as it is ;)
> 

I would like to stick to our conventions.

pstorch:
retest this please

S-Bartfast:
>We can't parse for example N 49 43.95 and change it to N 49 43.095, that is not correct!

I agree mathematically it's not correct but that is exactly what #6716 is asking for.  I'm more than happy to mark it as "will not fix" but that is what has been requested.

Making this logic unique to the calculator is of course an option as well but, which ever way we decide to go on this leading-trailing zeros thing, I think it would be better to make it universal.  At the moment, in the Coordinates Input Dialog, leading zeros are inserted after the decimal for the DEG, MIN and SEC formats but trailing zeros are appended when in PLAIN format.

pstorch:
@S-Bartfast we are preparing a new feature release. So we just have merged `master` to `release` and are now working on the `release` branch.
If you continue on this PR, it would be nice if you could rebase it on `release` branch.

S-Bartfast:
> @S-Bartfast we are preparing a new feature release.

Cool, thanks for letting me know.

As it is I had actually thought this issue was good to go at the moment though.
The solution currently implemented is to pad all formats with leading zero's after the decimal point for the calculator only.  Is that an acceptable solution or would we prefer something different?

In any event, if the calculator is going into release I would probably advice that this 'fix' go in as well and if it turns out we end up deciding something different later to then modify it accordingly.  My reason for suggesting as much is that I think the current solution, though unique to the calculator, is what most people would want/expect and we would be best of presenting this behaviour from the beginning even if we aren't 100% sure we want to keep it that way.  Without this 'fix' the calculator will pad with trailing zeros (after the decimal point) for all formats and I think this would be contrary to most users expectation.  I appreciate the policy is still up for debate but I think this implementation is more inline with peoples expectations and would be better served if it were included in the initial roll-out of the calculator.

Note, the fix as it is now only affects the calculator and thus should not be causing any test failures.

pstorch:
@S-Bartfast I hoped for the indicators as presented in https://github.com/cgeo/cgeo/issues/6716#issuecomment-339862836 to be included here :wink: 
But I'm fine with handling that in a separate issue/PR.

@Lineflyer what do you think about the leading zeros padding? Should we include it or further refine?

S-Bartfast:
> I hoped for the indicators as presented in #6716 (comment) to be included here.

Yeah, that would be nice, but it's not trivial to implement I'm afraid and is very unlikely to be completed in time for the release.  It should be perfectly fine to handle as a later enhancement though as it won't change behavior.

I think it's more important to include this PR as it is before the release though as including it later _will_ change the behavior of the product.

S-Bartfast:
Oh dear, think I may have messed things up again.

I just tried to rebase onto "Release" and it seems everything disappeared.
There were some minor conflicts, which I thought I resolved but after 'force-pushing' the only changes are the removal of some comments which I defiantly didn't mean to remove.  Is it possible to undo this?

Damn you "--force", you are the bane of my existence!

UniQP:
Don't worry, you can restore the old version.

`git reflog '#6716_Leading_Zero_Padding'` shows all recent versions/SHA1s of the branch `#6716_Leading_Zero_Padding`. What's the output of the command?

S-Bartfast:
>#6716_Leading_Zero_Padding. What's the output of the command?

It all looks like a tangeled mess to me:
>f7e2c837a (origin/#6716_Leading_Zero_Padding, Upstream/release, Origin_Leading_Zero_Padding, #6716_Leading_Zero_Padding) #6716_Leading_Zero_Padding@{0}: rebase finished: refs/heads/#6716_Leading_Zero_Padding onto f7e2c837a6f4637c93b27e979d0bd36648b0fda8
5527550c7 #6716_Leading_Zero_Padding@{1}: rebase -i (finish): refs/heads/#6716_Leading_Zero_Padding onto 5ab94a34f7d4b661f447896e6354b8f565556b1d
a4daefb76 #6716_Leading_Zero_Padding@{2}: commit: Fixed Codacy warnings.
5f276acbb #6716_Leading_Zero_Padding@{3}: rebase finished: refs/heads/#6716_Leading_Zero_Padding onto 5ab94a34f7d4b661f447896e6354b8f565556b1d
73afd1f0c #6716_Leading_Zero_Padding@{4}: rebase -i (finish): refs/heads/#6716_Leading_Zero_Padding onto ea62d1353271e8763757db281c2961b740ac32fc
8cb77e44f #6716_Leading_Zero_Padding@{5}: rebase finished: refs/heads/#6716_Leading_Zero_Padding onto ea62d1353271e8763757db281c2961b740ac32fc
d46e041d1 #6716_Leading_Zero_Padding@{6}: rebase -i (finish): refs/heads/#6716_Leading_Zero_Padding onto 9e534c254ee5980087b67d6249db5ee2025c3d89
56e7f38e4 #6716_Leading_Zero_Padding@{7}: commit: Moved zero padding into calculator
cde047e5b #6716_Leading_Zero_Padding@{8}: rebase -i (finish): refs/heads/#6716_Leading_Zero_Padding onto 9e534c254ee5980087b67d6249db5ee2025c3d89
40071be1f #6716_Leading_Zero_Padding@{9}: commit: Whitespace
f559b4ed3 #6716_Leading_Zero_Padding@{10}: reset: moving to f559b4ed3cb9c97cb774edddb41dd9a87aee9a3a
5b3f58449 #6716_Leading_Zero_Padding@{11}: commit: no message
f559b4ed3 #6716_Leading_Zero_Padding@{12}: reset: moving to f559b4ed3cb9c97cb774edddb41dd9a87aee9a3a
2c64196b0 #6716_Leading_Zero_Padding@{13}: commit: no message
f559b4ed3 #6716_Leading_Zero_Padding@{14}: rebase finished: refs/heads/#6716_Leading_Zero_Padding onto 9e534c254ee5980087b67d6249db5ee2025c3d89
408da9670 #6716_Leading_Zero_Padding@{15}: commit: Fixed #6716: Leading Zero Padding
0bd68cc01 #6716_Leading_Zero_Padding@{16}: branch: Created from #6781_Blank_section_crash

UniQP:
Ok, only the first two lines are interesting for us: f7e2c83 is the current SHA1 and 5527550 the previous one. Use `git checkout '#6716_Leading_Zero_Padding'` to switch to the branch and `git reset --hard 5527550` to reset the branch to the previous SHA1/version.

Afterwards, you can try to rebase on release, again. If you need any help in resolving the conflicts, let us know.

S-Bartfast:
Cool, thanks UniQP.
I'll give it a go.
(:

S-Bartfast:
Oow, look at that!
Back from the dead 👍

S-Bartfast:
Now after having resurrected my branch I take it I need to 'force-push' this resurrected branch to 'undo' the failed rebase on origin, yeah?  Then I can attempt the merge again.

At present it won't let me retry the rebase I think because I haven't pushed the 'resurrection'.

UniQP:
Yes, pushing the resurrected branch is a good idea. By the way, it's a good habit to use `--force-with-lease` instead of `--force`, because `--force-with-lease` also checks that your remote branch did not changed in the meantime.

In general, if you want to remember the old version before the rebase, you can simply create a new branch before the rebase `git branch backup`. Then, you can use `git reset --hard backup` to reset your branch if the rebase fails, or simply delete the branch afterwards if the rebase succeeds: `git branch -D backup`.

S-Bartfast:
Nice.  Think I have it done.

Looks like I'll have to open a new PR as this one inadvertently got closed but looks like we got there (:

PR #6801, opened 2017-10-25
Fixed #6796: Calculator Strings

### LINKED ISSUE #6796: Calculator strings
After importing the latest strings to Crowdin for translation, I have some suggestions:
- I would move the placeholder/hint entries to not translatable (a + b, 2 + 2, _, *, N D°....)
- I'd suggest to change the coordinate pattern to the more common N D° MM.SSS'

comments?

S-Bartfast:
Fixed #6796: Calculator Strings

### DIFF
--- a/main/res/layout/coordinatescalculate_dialog.xml
+++ b/main/res/layout/coordinatescalculate_dialog.xml
@@ -33,8 +33,7 @@
                 android:clickable="false"
 
                 android:gravity="right"
-                android:textAlignment="gravity"
-                android:text="@string/lat_point_min_format"/>
+                android:textAlignment="gravity"/>
 
             <TextView
                 android:id="@+id/lonRes"
@@ -48,8 +47,7 @@
 
                 android:paddingRight="16dp"
                 android:gravity="end"
-                android:textAlignment="gravity"
-                android:text="@string/lon_point_min_format"/>
+                android:textAlignment="gravity"/>
 
             <EditText
                 android:id="@+id/notes_text"
--- a/main/res/values/strings_not_translatable.xml
+++ b/main/res/values/strings_not_translatable.xml
@@ -95,4 +95,8 @@
     <string name="ellipsis" translatable="false">…</string>
     <string name="triple_dash_vertical" translatable="false">┆</string>
 
+    <string name="equation_hint" translatable="false">a + b</string>
+    <string name="free_variable_hint" translatable="false">2 + 2</string>
+    <string name="empty_equation_result" translatable="false">_</string>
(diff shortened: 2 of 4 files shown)


pstorch:
@rsudev as owner of the issue, can you have a look here? I can't judge it.

main/res/values/strings_not_translatable.xml
```diff
@@ -95,4 +95,8 @@
     <string name="ellipsis" translatable="false">…</string>
     <string name="triple_dash_vertical" translatable="false">┆</string>
 
+    <string name="equation_hint">a + b</string>
```
rsudev: Here you need to add `translatable="false"` to all the entries. Placing them in this file takes them out of our workflow fro translation, adding the attribute tells the tools to not complain about missing localisations.

main/res/layout/coordinatescalculate_dialog.xml
```diff
@@ -48,8 +47,7 @@
 
                 android:paddingRight="16dp"
                 android:gravity="end"
-                android:textAlignment="gravity"
-                android:text="@string/lon_point_min_format"/>
```
rsudev: What was the purpose of this text previously? I am just curious.

main/res/values/strings_not_translatable.xml
S-Bartfast: Okay, understood.

rsudev:
Looks like my attempt to solve the conflict in the web was not the best off ideas. @S-Bartfast Can you squash those commits (or rebase your original one and force push to your branch) to really resolve the situation? Otherwise I'll try to sort this out locally at home this evening.

main/res/layout/coordinatescalculate_dialog.xml
S-Bartfast: Oh, only just saw this.

To be honest I'm not exactly sure.  It's "default" text that's being placed in the results field but it never gets seen.  It's never seen by the user because as soon as the calculator opens this test gets replaced with the results of the calculation itself:
image

I don't recall why a gave it a default value but it probably had something to do with my wanting to see what the layout looked like before the actual result code was working.  A bit like how people put "Lorem Ipsum" text in mock layouts to get an idea oh how things look.

S-Bartfast:
Rebase done.

I think the mess you managed to get into was by doing a 'pull' after you rebased.  I spent several failed rebase attempts in the woods doing exactly that but eventually worked out I had to "rebase", then "force-push" WITHOUT doing a "pull".  I've always been very wary of the dreaded 'force' option but it seems that's what's needed.

PR #6804, opened 2017-10-28
fix #6715 merge WP userNotes with Calculator notes

### LINKED ISSUE #6715: Merging the existing waypoint note with the new note in the calculator
From #6704

> Interesting. I thought temporarily about linking the 'Calculator notes' with the 'waypoint notes' (or 'user notes') but didn't end up giving it much thought and just implemented a 3rd 'notes' field exclusively for the calculator. Would you find it more useful to have these notes merged in some way? If so I'll look at it as another ticket.

Yes, I see no benefit having two separate notes. I feel it as an disadvantage to not see the calculator notes in the waypoint overview. If the calculator is used this usually contains information (e.g. variables and formulas) which might be interesting for other waypoints, too.

pstorch:
Current data in the calculator notes is simply ignored. Possible data loss for nightly build users! But the calculator variables are still available, so the loss might not be too hard.

It also removes the calculate button from CoordinatesInputDialog when called from NavigateAnyPointActivity, SearchActivity and LogTrackableActivity. That caused a crash.

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -468,11 +468,12 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
     @Override
     public void saveCalculatorState(final CalcState calcState) {
         this.calcState = calcState;
+        this.userNote.setText(calcState.notes);
     }
 
     @Override
     public CalcState fetchCalculatorState() {
-        return calcState;
+        return new CalcState(calcState, userNote.getText().toString());
     }
 
     /**
--- a/main/res/layout/coordinatesinput_dialog.xml
+++ b/main/res/layout/coordinatesinput_dialog.xml
@@ -34,7 +34,8 @@
             style="@style/button_full"
             android:layout_width="fill_parent"
             android:layout_height="wrap_content"
-            android:text="@string/waypoint_calculate_coordinates" />
+            android:text="@string/waypoint_calculate_coordinates"
+            android:visibility="gone" />
 
         <Button
             android:id="@+id/clipboard"
--- a/main/src/cgeo/geocaching/models/CalcState.java
+++ b/main/src/cgeo/geocaching/models/CalcState.java
@@ -58,6 +58,11 @@ public class CalcState implements Serializable {
         this.notes = notes;
     }
 
+    public CalcState(final CalcState calcState, final String notes) {
+        this(calcState.format, calcState.plainLat, calcState.plainLon, calcState.latHemisphere, calcState.lonHemisphere,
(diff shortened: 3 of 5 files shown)


main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -468,11 +468,12 @@ public void updateCoordinates(final Geopoint gp) {
     @Override
     public void saveCalculatorState(final CalcState calcState) {
         this.calcState = calcState;
+        this.userNote.setText(calcState.notes);
     }
 
     @Override
     public CalcState fetchCalculatorState() {
-        return calcState;
+        return new CalcState(calcState, userNote.getText().toString());
```
S-Bartfast: I'm concerned about this line.
Can we be confident that 'calcState' will always be defined whenever this is called?  Previously I think it was fine for 'fetchCalculatorState()' to return "null" but this change requires that 'calcState' be defined.

pstorch: Good point. I'll have a look.

S-Bartfast: Changing this to:
```    @Override
    public CalcState fetchCalculatorState() {
        if (calcState != null) {
            calcState = new CalcState(calcState, userNote.getText().toString());
        }

        return calcState;
    }
```
Seems to work fine.

S-Bartfast: I see a null check has been implemented to satisfy #6816.

Any reason we're not updating the 'calcState' variable though.

pstorch: This is done when coming back from the calculator in `saveCalculatorState()`, right?

S-Bartfast: Yes, it will be updated when it comes time to save, but is their any reason were keeping the old one around at all?

To be honest I'm still a little hazy as to exactly how references work in Java (I'm a C++ developer by trade) but the fact that 'calcState' still refers to the old object means it can't be collected by the garbage collector until it is replaced, no?  I'm sure it doesn't matter, but once the new CalcState has been constructed we don't have any use for the old one do we?

pstorch: I think garbage collection is not a problem here, because latest when the activity gets finalized, the calcState gets finalized, too.
But I think I've introduced another bug here. I'll check later.

S-Bartfast: Sure, garbage collection is a very minor issue, but I don't see the value in having two different versions of the CalcState object (one that is stored in 'calcState' and one that is returned by 'fetchCalculatorState()').  Should they not be the same?

PR #6815, opened 2017-11-01
#6716 leading zero padding

S-Bartfast:
Replacing #6800

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -41,6 +41,8 @@ import android.widget.TextView;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 
 import butterknife.ButterKnife;
@@ -583,6 +585,40 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         }
     }
 
+    private String addLeadingZerosToDecimal(final String coordinate) {
+        final Pattern wholePatern = Pattern.compile("(.*)[.,]");
+        final Pattern decimalPatern = Pattern.compile("[.,](\\d*)");
+
+        final Matcher wholeMatcher = wholePatern.matcher(coordinate);
+        final Matcher decimalMatcher = decimalPatern.matcher(coordinate);
+
+        if (!wholeMatcher.find() || !decimalMatcher.find()) {
+            return coordinate;
+        }
+
+        final String leadingString = wholeMatcher.group(1);
+        String trailingString = decimalMatcher.group(1);
+
+        final int decimalPoints;
+        switch (currentFormat) {
+            case Deg:
+                decimalPoints = 5;
+                break;
+
+            case Min:
+            case Sec:
+            case Plain:
+            default:
+                decimalPoints = 3;
+        }
+
+        while (trailingString.length() < decimalPoints) {
+            trailingString = "0".concat(trailingString);
+        }
+
+        return leadingString + "." + trailingString;
+    }
+
     /**
      * This method has largely been lifted from the CoordinateInputDialog and is used to validate
      * coordinate specifications prior to assigning them to the Geopoint.
@@ -590,12 +626,16 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     private boolean areCurrentCoordinatesValid() {
         // convert text to geopoint
         final Geopoint current;
-        final String lat = getLatResult();
-        final String lon = getLonResult();
+        String lat = getLatResult();
+        String lon = getLonResult();
 
         if (currentFormat == null || lat.contains("_") || lat.contains("*") || lon.contains("_") || lon.contains("*")) {
             return false;
         }
+
+        // Pad decimal field with leading zeros
+        lat = addLeadingZerosToDecimal(lat);
+        lon = addLeadingZerosToDecimal(lon);
 
         try {
             current = new Geopoint(lat, lon);


kumy:
Jenkins retest this please

main/res/values/changelog_release.xml
```diff
@@ -2,7 +2,7 @@
 <resources>
   <!-- changelog for the release branch -->
   <string name="changelog_release" translatable="false">\n
-    <b>Next beta version:</b>\n
+    <b>2017.11.01-RC:</b>\n
```
Bananeweizen: Looks like a merge error on your side.

S-Bartfast: No worries.

I actually see that while I was merging and thought "2017.11.01-RC" was the correct choice.  I modify it.

S-Bartfast: Hang-on, Lars made that change about an hour before I committed.

What should it be?

pstorch: I fixed it by changing the target branch to `release`, where it should be.

pstorch:
@Bananeweizen, @Lineflyer, @kumy, @rsudev, @SammysHP, @mucek4 as I (personally) have some doubts about the leading zeros for decimals in the calculator, I would like to have your opinions. If the majority of the team is in favor of having this behavior I'm totally fine with merging it.

UniQP:
@pstorch You are talking about the [DegDec](http://www.cgeo.org/faq#coords-format) format, right?
I'm not aware of a standard number of decimal places for coordinates in this format: Google Maps uses 6 decimal places, OSM between 5 and 15. Thus, padding with leading zeros makes no sense in this case.

pstorch:
@UniQP it's mainly about the decimals for the MinDec and DMS formats. See also the discussion on the issue #6716

Bananeweizen:
@pstorch I must admit I have not used the calculator at all yet. I only read the code changes, since I was looking into build issues yesterday. To me the only valid rule is using that of pure math: use trailing zeros for everything after a decimal point, use leading zeros for everything in front of a decimal point. However, I'm aware that not every cache owner is interpreting this the same, so I would not enforce this. 

We probably should put the highest priority on signaling to the user that the result is questionable and to have him be able to choose the "other interpretation" easily. E.g. could we have the result of the calculation not as a single entry (in the textfield right now), but maybe as a list of results, that includes 2 or 3 entries then?

If the "attention, this might not be what you want" is implemented, then I guess we have to wait for feedback from a broader user base anyway, since we are otherwise always creating assumptions, what users like/understand. Let's just get some data from feedback and then rethink it eventually.

pstorch:
@UniQP but good point about DegDec, I wasn't aware of. This solution assumes 5 decimal digits.

main/res/values/changelog_release.xml
S-Bartfast: Just so I understand, how does one set the "Target Branch"?
I take it it's not as simple as just rebasing onto the "release" branch.

pstorch: If you edit the PR on the Website. On the top you have a dropdown with the branch.

S-Bartfast:
@UniQP, I wasn't aware of their being ambiguity about the number of decimal points in DegDec either. That is indeed a concern. I just checked the regular coordinate input dialog and it zero pads with trailing decimals for DegDec format but pads with leading zeros for both MinDec and DMS.  How frustrating!

My intention was to copy the behaviour of the regular coordinate input dialog as best I could.  While this zero padding is definitely of concern for the calculator lets not loose sight of the fact that it is also a concern of the previous coordinate input dialog as well. As I mentioned in the previous discussion (#6800) I would really prefer to have the calculator use the same policy as the coordinate input dialog so if we are to define the policy for the calculator we should probably amend the coordinate input dialog's policy as well.

main/res/values/changelog_release.xml
S-Bartfast: Oh, I see.

So even though I'd rebased onto 'release' the actual request was still being made to 'master'.  So I need to both 'rebase' _then_ update the target branch.  Got it 👍

Lineflyer:
I did not closely follow the trailing/leading zero problem, although I initially started the disussion...sorry for that. Not enough time.

Without knowing the purpose of this PR in detail (maybe someone can sum up the current behaviour and the change planned with this PR?), it is at the end a question of how mathematical correct owners create their fomulas (mainly speaking about MinDec format as this is the most used format).

For example today I had a multi GC4FHTR:
The formula for the final was given as:
```
N 51° 55.X+112
E6° 47.0(Y-36)
```

So I programmed the calc like in this way:
```
N 51° 55.A
E006° 47.0B
```
with
```
A=x+112
B=y-36
```

After finding the needed info I entered it:
```
x=25
y=45
```

The calcuators output was
```
N51° 55.137
E006° 47.090
```
but the cache was located at ```E006° 47.009```.
mmmh..

Expecting a single digit result the owner could have written ```47.00(Y-36)```, but I guess he did not want to give this clue in his formular.

I think there is no clear solution whether to use trailing/leading zero all the time or not or depending on format. I need to think about this looking at more caches and the different coordinate formats once more.
We have an issue where this is discussed in general, do we?

pstorch:
@Lineflyer the corresponding issue is #6716

S-Bartfast:
> Without knowing the purpose of this PR in detail (maybe someone can sum up the current behaviour and the change planned with this PR?)

This PR is to implement leading zero padding after the decimal point:
> WITH this PR the calculator will pad _all_ formats with **leading** zero's after the decimal point.
WITHOUT this PR the calculator will pad _all_ formats with **trailing** zero's after the decimal point.
NOTE that the 'Coordinate Input Dialog' itself pads with **leading** zeros for 'MinDec' and 'DMS' formats but pads with **trailing** zeros for 'Plain' and 'DegDec' formats.

In your instance this PR would have given you the result you sought but whether that should be a general policy is open for debate.

pstorch:
Anyway we do it, it will be wrong in some situations for some users. And we need more feedback anyway.
The question is should we collect more feedback and discuss a solution within the next release with or without this change?

Lineflyer:
THanks for your explanations, I think I will follow up in #6716 with my personal opinion. This might be better than to misuse a PR comment. ;)

pstorch:
I think some rebasing is necessary here.

S-Bartfast:
>I think some rebasing is necessary here.

Really? I rebased it just last night.
Are you saying it should be rebased onto 'release'?

pstorch:
@S-Bartfast yes, rebase onto `release` when we want to merge it to `release`. Because at the moment I see 13 commits.

S-Bartfast:
Okay, I (attempted) the rebase and now have this:
image

Is that correct?

I'm not sure why the  "Upstream/release" tag is 5 commits behind my commit, but it doesn't seem right.

Should I attempt to rebase again on to the commit that has the tag?

S-Bartfast:
Oh, I see.  It has re-grafted everything that has happened on 'master' onto 'release' since the last merger.

Okay, I take another swing at it!

S-Bartfast:
Okay, how's that?

I did keep the #6823 commit in as it was just cleaning up code that was hastily added to 'release' to begin with.

pstorch:
Hmm, the fix for #6823 was already merged to `master`. I don't know what happens if we merge it into `release` now, too and later merge `release` to `master`?

S-Bartfast:
Okay, I'll remove it.

pstorch:
let's go

PR #6819, opened 2017-11-01
fix #4124 html in waypoint popup

### LINKED ISSUE #4124: Dont show HTML tags in waypoint popup
- Map a cache which has waypoints with coordinates
- Open the popup for one of the waypoints
- The waypoint notes are shown with their HTML tags

We should either format the text according to the tags or simply strip them.

Example: GC2K085 - Stage 2
screenshot_2014-08-05-13-03-33

pstorch:


pstorch:
retest this please

PR #6824, opened 2017-11-03
Fixed  #6823: Calculator Notes Cleanup

### LINKED ISSUE #6823: Calculator Notes Cleanup
#6715, #6816 and #6821 are all addressing the calculator notes issue and all were closed before I was able to be involved with the changes, so I thought it would be best to open an issue of my own to provide input.

As it stands the 'calculator notes' have been removed from the 'calculators state' but seem to be tagged onto the state externally with an 'ImmutablePair' object.  In my mind if the notes field is essentially intended to be independent of the calculator itself it would be better to simply access the Waypoint's note data directly than tag it onto the calculator's state and kind of 'carry' them around together.

I have an implementation of this and are mainly creating this issue so as to have a branch in which to commit this implementation.

S-Bartfast:
Fixed  #6823: Calculator Notes Cleanup

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -467,14 +467,13 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
     }
 
     @Override
-    public void saveCalculatorState(final CalcState calcState, final String notes) {
+    public void saveCalculatorState(final CalcState calcState) {
         this.calcState = calcState;
-        this.userNote.setText(notes);
     }
 
     @Override
-    public ImmutablePair<CalcState, String> fetchCalculatorState() {
-        return new ImmutablePair<>(calcState, userNote.getText().toString());
+    public CalcState fetchCalculatorState() {
+        return calcState;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


S-Bartfast:
Oh, I see this was merged into 'master' rather than release'.
What would be the best way to fix this?

pstorch:
This was only a code cleanup, not a Bugfix. So it should go to master, right?

Lineflyer:
@pstorch I would appreciate that, as far as its only a cleanup and does not implicitly contain a bugfix.

S-Bartfast:
I just though it should go in the release as it was neatening up code that was added for the release.

S-Bartfast:
If this could be brought across into release soon I would appreciate it.

I have to keep undoing this change whenever I commit to release but for some reason I'm having trouble undoing it on my #6835 branch. I'm not sure what the problem is with my #6835 branch but suffice to say it wouldn't be a problem if this PR were included already ;)

As it is I am now about to take a swipe at #6843 (Calculator does not save state on pause) and again will be working from a base that includes this commit.

Cheers,
(:

pstorch:
@S-Bartfast mabe you can try to delete your local `release` branch and checkout a clean version from upstream. If this doesn't solve it for you, I can try to cherry-pick it from `master` to `release`. So that we are not blocked with fixing the last two issues for the release.

S-Bartfast:
>mabe you can try to delete your local release branch...

My release branch is fine, it's my "#6835_Plain_Format_Bracketing" that is the problem.
I'm not sure that I want to delete that branch yet as I'm not 100% confident it won't get lost if I do.

Issue #6791, opened 2017-10-23
Problem if waypoint description contains a "<"


Matze2:
##### Detailed steps causing the problem:
Scenario 1 on imported waypoint note:
- Import GC768Y3
- Note for waypoint "Stage 14" is truncated at the "less than" (<) sign

Scenario 2 on waypoint user note:
- Add a waypoint user note containing a "less than" (<) sign
- Only the text before the "less than" (<) sign is saved

##### Actual behavior after performing these steps:
See steps above

##### Expected behavior after performing these steps:
Full waypoint description is imported / saved

##### Version of c:geo used:
2017.10.18-NB-30e87fb

##### Is the problem reproducible for you?
Yes

##### System information:
```
On demand (should not matter)
```

Matze2:
Waypoint calculator note seems not affected.

Lineflyer:
Confirmed using GC768Y3 also in official release 2017.10.23

pstorch:
The description (note) of Stage 14 is recognized as HTML in

https://github.com/cgeo/cgeo/blob/2f1a321510ed2906e7619dc36c19d5c50bca0554/main/src/cgeo/geocaching/utils/TextUtils.java#L163-L173

and then later parsed by `Html.fromHtml()`:
https://github.com/cgeo/cgeo/blob/2f1a321510ed2906e7619dc36c19d5c50bca0554/main/src/cgeo/geocaching/CacheDetailActivity.java#L1871

finally the `<10` is recognized as a tag with name `_10` and forwarded to the `UnknownTagsHandler` with the rest stripped, because the end `>` of the opening tag is not found.

If there is a blank before the number like `< 10` then it is not recognized as a tag and it works.

We would have to find all `<` which are not starting a tag and replace them with `&lt;` before calling `Html.fromHtml()`. 
Any other idea?

Matze2:
Makes sense to me. Browsers probably use the same logic.

pstorch:
My problem is how to recognize correctly a `<` which is not part of a start tag.
Because a tag name can consist of alphanumeric characters:
https://www.w3.org/TR/html51/syntax.html#tag-name
So in our example the `<10` is a valid start of a start tag. 
Then how to recognize the corresponding `>` is missing? Just looking for `>` without another `<` in between is not good enough. There could be attributes in the tag which could have a `<` in the attribute value, so you have to count single and double quotes, too...
[...]

PR #6834, opened 2017-11-08
fix #6826 remove superfluous waypoint note

### LINKED ISSUE #6826: User created waypoint has two editable note areas
- Add a new waypoint
- There are two editable areas "Note" and "User Note"

Two areas where one is enough.

"Note" area which is usually read-only for imported waypoints is not visible because it makes no sense for user created waypoints.

##### Version of c:geo used:
2017.11.05-NB-cc9328b

Yes

##### System information:
On demand.

Not sure if this was always the case.

pstorch:
For user defined waypoints, the note is merged with the userNote and the original note be emptied.

Empty note fields get hidden, even for server defined waypoints.

For server defined notes gets preserved. They are not editable but we stripped html. Now html can be preserved.

Html in userNotes is not supported and "<" characters escaped.

This might solve #6791 as well.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1890,11 +1890,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
             if (StringUtils.isNotBlank(wpt.getUserNote()) && !StringUtils.equals(wpt.getNote(), wpt.getUserNote())) {
                 userNoteView.setOnClickListener(new DecryptTextClickListener(userNoteView));
                 userNoteView.setVisibility(View.VISIBLE);
-                if (TextUtils.containsHtml(wpt.getUserNote())) {
-                    userNoteView.setText(Html.fromHtml(wpt.getUserNote(), new SmileyImage(cache.getGeocode(), userNoteView), new UnknownTagsHandler()), TextView.BufferType.SPANNABLE);
-                } else {
-                    userNoteView.setText(wpt.getUserNote());
-                }
+                userNoteView.setText(wpt.getUserNote());
             } else {
                 userNoteView.setVisibility(View.GONE);
             }
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -13,6 +13,7 @@ import cgeo.geocaching.location.GeopointFormatter;
 import cgeo.geocaching.models.CalcState;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Waypoint;
+import cgeo.geocaching.network.SmileyImage;
 import cgeo.geocaching.sensors.GeoData;
 import cgeo.geocaching.sensors.GeoDirHandler;
 import cgeo.geocaching.sensors.Sensors;
@@ -25,6 +26,7 @@ import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.ClipboardUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.TextUtils;
+import cgeo.geocaching.utils.UnknownTagsHandler;
 
 import android.app.AlertDialog;
 import android.app.ProgressDialog;
@@ -37,6 +39,7 @@ import android.os.Handler;
 import android.os.Message;
 import android.support.annotation.NonNull;
 import android.support.v4.content.LocalBroadcastManager;
+import android.text.Html;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -58,12 +61,12 @@ import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
 
+
 import org.androidannotations.annotations.EActivity;
 import org.androidannotations.annotations.Extra;
 import org.androidannotations.annotations.InstanceState;
 import org.androidannotations.annotations.ViewById;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
 
 @EActivity
 public class EditWaypointActivity extends AbstractActionBarActivity implements CoordinatesInputDialog.CoordinateUpdate, CoordinatesInputDialog.CalculateState {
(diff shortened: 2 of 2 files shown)


Matze2:
I did a quick test. For edit it looks good.
But for a new waypoint it still shows both note fields.

Matze2:
I've added a
`            note.setVisibility(View.GONE);
`
around line 262 in the "else"-Block to hide the note area in case of "Add Waypoint".

Matze2:
Another one: in case of "duplicate" the note is stil HTML when concatenated.
I guess HTML.fromHTML() has to be called before entering the if-else-block but it's too late now. :)

Matze2:
Here is my git diff. Almost untested. Just for illustration.
[diff.txt](https://github.com/cgeo/cgeo/files/1465304/diff.txt)

pstorch:
@Matze2 thanks for testing. The handling of the HTML formatted notes gets a bit tricky now. But I wanted to preserve it as long as possible. I'll have a look.

Matze2:
At least it does not get more complex with my changes. And it seems that now the three use cases (new, edit, duplicate) work.

Matze2:
Tested a little bit more the PR plus my changes that fix the remaining issues with "duplicate" and "new" waypoint. Everything looks fine to me now.

pstorch:
I'm testing your changes, too. The visibility to GONE is working.
With the HTML issue I'm not sure right now. The only problem is the `\n` concatenation is not resulting in a line break, if the note contains HTML. Everything else seams to be fine.
With your changes the EditNotesActivity crashes when I immediately clear the userNote field, save the waypoint and open it again.

Matze2:
About the \n: yes I see that, too. But only for duplicates where HTML is preserved.

Matze2:
It's double-stripped:
first in line 162: Html.fromHtml
second in line 173 if user note contains a "<"

So I reproduced it with the following steps:
- take a waypoint with a note
- add <12 as user note
- duplicate it
- open it (line feed is eaten between note and user note)

Matze2:
With the crash: you're right. I screwed it up. waypoint.getNote() can be null and I now call HTML.fromHtml() unconditionally.

pstorch:
Hmm, the fix for the `<` characters which where interpreted as invalid HTML tags cause some confusion. It causes the entire note to be recognized as HTML (because it is a `&lt;` now) and all line breaks get ignored.

I think we have to properly "clean" the notes: strip html tags while preserving the line breaks somehow. And the `TextUtils.containsHtml()` needs to be more intelligent.

pstorch:
I've updated the PR, but the issue is not finally solved yet. That's why I added the `Do not merge` tag.
I'll continue tomorrow. @Matze2 thanks for your help.

Matze2:
I have fix for the NullPointerException. Will share it later.

pstorch:
@Matze2 The Nullpointer is already handled. But if you like, you can take over and continue on this.

Matze2:
No, is ok. I was just irritated because I still saw an NPE, but with my tainted waypoint with a null note.
As this cannot happen anymore, I am fine.

For the HTML not vs. non-HTML user note, I am not sure I should start on that.
This looks weird - especially together with duplicate - and might require bigger changes as you pointed out.

I will give it a try when you have something new.

Matze2:
Ok. A few comments before I also stop for today:
- Do we really need to check if user note is HTML? How can this happen since user note is only edited in a simple text field? And in duplicate case HTML note is already converted before it gets copied to user note...
- Is it 100% sure that waypoint.getNote() is never null? Otherwise a check might be needed. Something like that (note that I use CharSequence instead of Spanned):
```
                        CharSequence noteFromHTML;
                        if (StringUtils.isNotBlank(waypoint.getNote())) {
                            noteFromHTML = Html.fromHtml(waypoint.getNote(), new SmileyImage(activity.geocode, activity.note), new UnknownTagsHandler());
                        } else {
                            noteFromHTML = StringUtils.EMPTY;
                        }
```

pstorch:
@Matze2 my null handling was this:
```
final Spanned noteFromHTML = Html.fromHtml(StringUtils.trimToEmpty(waypoint.getNote()), new SmileyImage(activity.geocode, activity.note), new UnknownTagsHandler());

````

The `StringUtils.trimToEmpty()` handles null and converts it to empty string. 
Or have you found another line where a null pointer could happen?

Matze2:
You're right. Maybe it was too late yesterday. I might have build a mixture of your latest commit and code from me.
Your changes should work. Sorry about the confusion.

pstorch:
> Sorry about the confusion.
Don't be sorry. I'm glad that someone else is looking at it, too. I regularly miss things ;)

pstorch:
Ok, when we agree on NOT supporting HTML in the `userNote` field then it is much easier (I guess anyway no one likes to edit HTML in source code in an Android textfield).

It looks like I found a way to merge the HTML `note` with a TEXT `userNote`, incl. preservation of line breaks in the TEXT.

Please have a look.

Lineflyer:
> Ok, when we agree on NOT supporting HTML in the userNote field then it is much easier 

My few cents:
I dont think HTML in the user note is being used quite often.
So I agree it could be interpreted as plain text anyway.

Matze2:
Wow! You were really busy. Generally I like it when more code/complexity is removed than added. :)

About HTML in user note: I have no idea how you can use it at all. Entering HTML tags directly in the input area? Is this supported?

Anyway, I like the diff, but I cannot test anything before Friday. But then I will.

Matze2:
I still have some old waypoints (created in previous versions) with null note.
These crash in line EditWaypointActivity:158. Do we need to care about that?

Matze2:
I added another trimToEmpty in line 158 and now this works.
`
                        if (TextUtils.containsHtml(StringUtils.trimToEmpty(waypoint.getNote()))) {
`

Matze2:
So, finally with the correct branch I did some testing.
- Existing waypoints with "<" in user note are now escaped and need to be fixed once.
- Existing waypoints with "<" in note got truncated but after reload everything was fine with them

Besides that I checked open, duplicate save and reopen with notes and without notes (also with <) for
- listing waypoints without note and user note
- listing waypoints with note
- own waypoints
and couldn't find any issue. 👍

pstorch:
@Matze2 thanks again for testing.

pstorch:
@all ready for testing in `master`?

Issue #6802, opened 2017-10-25
Coordinate Parser Needs Work


S-Bartfast:
I keep observing wacky interpretations for things being presented to the coordinate parser, particularly for the 'Plain' format parser, such as this:
![image](https://user-images.githubusercontent.com/6573522/32008762-97d43ddc-b9f4-11e7-8efd-9725af937731.png)

No real idea what's going on there but suffice to say it probably shouldn't be being parsed as a valid GeoPoint.

To play with the parser yourself create a new waypoint and open the calculator, change the format to 'Plain' and anything it considers 'valid' will be displayed in white (like above) and anything invalid will have the result displayed in grey:
![image](https://user-images.githubusercontent.com/6573522/32009009-5574e1e8-b9f5-11e7-9991-7f829ab329fa.png)

UniQP:
I think a good next step is to extract a failing unit test.

S-Bartfast:
Yeah, I haven't even looked at unit tests yet.  I should probably get onto that next.

S-Bartfast:
Sorry, still haven't got to this one.

Just off the top of my head these are additions that need to me made:

* Check for non legal characters
Other than the first letter the only valid characters I can think of are `0-9`, `.`, `,`, `'` ,`"`, `°` and ` ` (space) 
(though we may want to add other min/sec characters).
* There should only be one of these characters (`.`, `,`, `'` ,`"`, `°`)
* There should be at most 3 groups (using `'` ,`"`, `°` and ` ` as delimiters)
 If we want to add `.` and `,` as delimiters there could be up to 4 groups.
* there should be only one group after the decimal point

If any of these conditions are not met the formulation should be considered invalid.

In terms of providing feedback to the user if a formulation is invalid it would be good to know which character was the first to cause it to fail.  I would like to use this information in conjunction with #6842 (Implement "Result Highlighting" in the Calculator) to help guide the user to creating valid formulations.

Matze2:
This list is good. I compiled a similar one while I was experimenting with stricter regexp patterns.
From looking through the GeoPointParser code and its usage the problem seems to be that it combines two modes into one:
- trying to parse coordinates from text; this includes also passing latlon in one line; here it's hard to require the stricter settings
- trying to parse single coordinates (lat and lon separated); here it is possible to be very pedantic which gives the user an immediate feedback if the input is valid or not

So there could be a "precheck" phase for the second case which implements all the additional checks mentioned above.

[...]

Issue #6797, opened 2017-10-24
Next feature release


Lineflyer:
Its once again a big gap between `release` and `master` already, which makes me think about finding a new cut to do a feature release. The major new feature on `master` seems to be the waypoint calculator, which basically works and some issues are known and about to be adressed.  Additionally we have some other features like OC-picture logs.

While using the nightlies I have the slight impression that we got some unstability, which I was not able to reproduce or track down right now. It is just, that c:geo more often crashes randomly.
As I have only limited time to test at the moment I would therefore suggest to use our beta program to let users test based on current `master`, Therefore we should IMHO merge `master` to `release` asap.

We will not have a version which I would consider ready to be released right away, but it allows us to
- Freeze current feature state
- Release a beta version
- Fix found issues on release

I have seen that some PRs are still open (mainly dealing with the calculator), therefore my question would be if we should wait until these are finished and merged or rather align the branches now and merge the PRs to `release` once they are ready?

Waiting for your feedback.

pstorch:
From my point of view I would like to see the current PRs for the calculator be merge first. They address some possible crashes.

I also think #6715 should be fixed before merging to `release`. Because this would cause a DB change (remove a column).

Matze2:
Having tested the WP calculator a lot I agree with @pstorch. Without the current fixes being merged c:geo can be crashed easily.
Same for #6715: I guess users would lose the calculator specific notes after #6715 is done which is probably only acceptable for nightly users.
Without #6715 there are three different waypoint notes with other issues, e.g. this information is not duplicated.

Lineflyer:
Its without doubt, that these fixes shall come in before release or publishing a beta.
My question regarding the PRs was more if they should be finalized on `master` first and then copy to `release` or copy to release immediately and fix on `release`.
From @pstorch answer I learned it should be done on `master` and only then copy to `release`.

Then I  suggest to put no more huge untested changes (other new features) on `master` while it has not yet been copied to `release`.

pstorch:
Just realized that I was wrong about #6715: the calculator note is not a separate column. It's just in the `calc_state` column together with the variables. But nevertheless I think this should be fixed, because it either needs a more or less complicated migration or causes data loss.

pstorch:
From my point of view we can start a new feature release branch.

rsudev:
Just a sec, I'll download the last translations.

Lineflyer:
> From my point of view we can start a new feature release branch

[...]

PR #6836, opened 2017-11-09
Resolves #6835: Allow use of Brackets for the calculator while in 'Plain' format

### LINKED ISSUE #6835: Allow use of Brackets for the calculator while in 'Plain' format
As discussed in the original proposal for the calculator only single character 'labels' fit on the buttons requiring intermediary 'names' to be used:
image

However, in 'Plain' format we do not have such size restrictions, so allowing direct entering of bracketed expressions should be possable:
image

S-Bartfast:
Allow use of Brackets for the calculator while in 'Plain' format

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -63,7 +63,6 @@ import org.androidannotations.annotations.Extra;
 import org.androidannotations.annotations.InstanceState;
 import org.androidannotations.annotations.ViewById;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
 
 @EActivity
 public class EditWaypointActivity extends AbstractActionBarActivity implements CoordinatesInputDialog.CoordinateUpdate, CoordinatesInputDialog.CalculateState {
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -467,14 +467,13 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
     }
 
     @Override
-    public void saveCalculatorState(final CalcState calcState, final String notes) {
+    public void saveCalculatorState(final CalcState calcState) {
         this.calcState = calcState;
-        this.userNote.setText(notes);
     }
 
     @Override
-    public ImmutablePair<CalcState, String> fetchCalculatorState() {
-        return new ImmutablePair<>(calcState, userNote.getText().toString());
+    public CalcState fetchCalculatorState() {
+        return calcState;
     }
 
     /**
@@ -514,6 +513,10 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
         if (currentState != null && isWaypointChanged(currentState)) {
             ActivityMixin.showToast(this, R.string.warn_discard_changes);
         }
+    }
+
+    public EditText getUserNotes() {
+        return userNote;
     }
 
     private ActivityData getActivityData() {
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -5,6 +5,7 @@ import static cgeo.geocaching.R.id.coordTable;
 import static cgeo.geocaching.models.CalcState.ERROR_CHAR;
 import static cgeo.geocaching.ui.dialog.CoordinatesInputDialog.GEOPOINT_ARG;
 
+import cgeo.geocaching.EditWaypointActivity;
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.AbstractActivity;
 import cgeo.geocaching.location.Geopoint;
@@ -70,14 +71,12 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     private static final String SYMBOL_SEC = "\"";
     private static final String SYMBOL_POINT = ".";
     public static final String CALC_STATE = "calc_state";
-    public static final String NOTES = "notes";
 
     private ImageButton doneButton;
     private boolean stateSaved = false;
 
     private Geopoint gp;
     private CalcState savedState;
-    private String savedNotes;
 
     /** List of equations to be displayed in the calculator */
     private List<CalculatorVariable> equations;
@@ -166,9 +165,9 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         public void onClick(final View v) {
             // Save calculator state regardless of weather the coordinates are valid or not.
             final CalcState currentState = getCurrentState();
-            final String currentNotes = notes.getText().toString();
-            setSavedState(currentState, currentNotes);
-            ((CoordinatesInputDialog.CalculateState) getActivity()).saveCalculatorState(currentState, currentNotes);
+            setSavedState(currentState);
+            ((CoordinatesInputDialog.CalculateState) getActivity()).saveCalculatorState(currentState);
+            ((EditWaypointActivity) getActivity()).getUserNotes().setText(notes.getText());
 
             if (!areCurrentCoordinatesValid()) {
                 return;
@@ -311,9 +310,8 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     /**
      * @param gp               Geopoint representing the coordinates from the CoordinateInputDialog
      * @param calculationState State to set the calculator to when created
-     * @param notes            user notes
      */
-    public static CoordinatesCalculateDialog getInstance(final Geopoint gp, final CalcState calculationState, final String notes) {
+    public static CoordinatesCalculateDialog getInstance(final Geopoint gp, final CalcState calculationState) {
         final Bundle args = new Bundle();
 
         if (gp != null) {
@@ -322,7 +320,7 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
 
         final CoordinatesCalculateDialog ccd = new CoordinatesCalculateDialog();
         ccd.setArguments(args);
-        ccd.setSavedState(calculationState, notes);
+        ccd.setSavedState(calculationState);
         return ccd;
     }
 
@@ -342,9 +340,8 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
                 gp = savedInstanceState.getParcelable(GEOPOINT_ARG);
             }
 
-            final String savedNotes = savedInstanceState.getString(NOTES);
             final byte[] bytes = savedInstanceState.getByteArray(CALC_STATE);
-            setSavedState(bytes != null ? (CalcState) SerializationUtils.deserialize(bytes) : null, savedNotes);
+            setSavedState(bytes != null ? (CalcState) SerializationUtils.deserialize(bytes) : null);
         }
     }
 
@@ -431,6 +428,7 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         tLonResult = ButterKnife.findById(v, R.id.lonRes);
 
         notes = ButterKnife.findById(v, R.id.notes_text);
+        notes.setText(((EditWaypointActivity) getActivity()).getUserNotes().getText());
 
         latButtons = Arrays.asList(            bLatDeg[1], bLatDeg[0],
                                                bLatMin[1], bLatMin[0],
@@ -523,9 +521,6 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
             resetCalculator();
         }
 
-        if (savedNotes != null) {
-            notes.setText(savedNotes);
-        }
 
         resortEquations();
         updateResult();
@@ -538,7 +533,6 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     public void onSaveInstanceState(final Bundle outState) {
         super.onSaveInstanceState(outState);
         outState.putByteArray(CALC_STATE, SerializationUtils.serialize(getCurrentState()));
-        outState.putString(NOTES, notes.getText().toString());
     }
 
     private void displayToast(final int message) {
@@ -651,9 +645,8 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         return false;
     }
 
-    private void setSavedState(final CalcState savedState, final String savedNotes) {
+    private void setSavedState(final CalcState savedState) {
         this.savedState = savedState;
-        this.savedNotes = savedNotes;
         stateSaved = true;
 
         if (doneButton != null) {
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -584,18 +584,21 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     }
 
     private String addLeadingZerosToDecimal(final String coordinate) {
-        final Pattern wholePatern = Pattern.compile("(.*)[.,]");
-        final Pattern decimalPatern = Pattern.compile("[.,](\\d*)");
+        final Pattern preDecimalPattern = Pattern.compile("(.*)[.,]");
+        final Pattern decimalPattern = Pattern.compile("[.,](\\d*)");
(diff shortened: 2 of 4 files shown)


pstorch:
@S-Bartfast: as this is a new feature it should address `master` branch.
cc @Lineflyer

pstorch:
I really would like to have this feature. Today I went for a small Multicache with just one final formula.
But this formula was complex enough, that I didn't want to set it up with the individual buttons. At home I checked out this PR and tried it with C&P of the formula in the Plain format. But the result is wrong:

image

It sould be: `N 54° 33.845 E 008° 37.679`
As you can see only the first brackets are evaluated and the rest of the expressions are ignored.

S-Bartfast:
>As you can see only the first brackets are evaluated and the rest of the expressions are ignored.

Yup, this is a problem with the parser rather than an issue with the calculator as such.

Have you tried entering `N 54° 33.125+720` into the regular coordinate parser.  I'm pretty sure you'll get the same result.

It's not entirely clear to me why there are even brackets in that formulation at all.  Usually brackets are used to delineate discrete numbers but in this instance it appears the brackets are intended to assign mathematical priority –though mathematically the expression would be identical whether brackets were used or not.

In this instance though, given it appears you want to group everything after the decimal point into a single number you could simply enclose everything in another set of brackets.
As in:
`N 54° (A-1967).((Z*F)+720)`
`E 008° (K-1963).((330*G)+9)`

Or if it makes it eaiser to follow you could group the number in square brackets such that the round brackets look more 'mathematical'.
As in:
`N 54° [A-1967].[(Z*F)+720]`
`E 008° [K-1963].[(330*G)+(G*E)+9]`

The real issue here though is the fact that `N 54° 33.125+720` is not parsed as being an error.

UniQP:
> The real issue here though is the fact that `N 54° 33.125+720` is not parsed as being an error.

The parse function currently tries to parse different formats and picks the longest match (`N 54° 33.125` in the example). I can add a check that it matches the whole string, if desired.

Might be related to #6802.

S-Bartfast:
>I can add a check that it matches the whole string, if desired.

Yes, that will defiantly be needed.
I'll continue the discussion in #6802 itself.

pstorch:
As mentioned in https://github.com/cgeo/cgeo/issues/6797#issuecomment-346464089 this should go to `release`. 
@S-Bartfast can you rebase onto `release`?
@UniQP #6850 is a prerequisite for this? Then please change to `release` as well.

UniQP:
Done

pstorch:
I've tried it again and merged in #6850 locally into this, but it doesn't solve it. So I guess something else needs to be done additionally.

BTW: adding extra brackets like `N 54° (A-1967).((Z*F)+720)` works, but without them should, too.

S-Bartfast:
>BTW: adding extra brackets like N 54° (A-1967).((Z*F)+720) works, but without them should, too.

I don't think `"N 54° (A-1967).(Z*F)+720"` _should_ be a valid formulation.  The issue with this formulation is that the brackets are being used in a different way than they usually are.  Normally whatever is inside the brackets is replaced by the result of the expression contained inside them and these results are simply concatenated (joined together) to create the final representation.

In this example the brackets (I assume) are defining mathematical priority, which is not normally what brackets are use for in this context.

Sure, we would be able to run an evaluation over each of the groups (essentially automatically wrapping each group in brackets) but I don't think that would be advisable.  Basically I think the calculator should report errors rather than try to 'guess' at what was intended.  Like the example we had before with additional spaces after the decimal point, once we start interpreting what was _meant_ rather than what is actually there  we will quickly run into contradictions.

For instance, if:
`N 54° (A-1967).(Z*F)+720` is equivalent to `N 54° (A-1967).((Z*F)+720)`

Surely `N 54° A-1967.Z*F+720` is equivalent to `N 54° (A-1967).(Z*F)+720` as well (I mean those brackets in this case are meaningless).

And what about `1.5 + 10` is that `1.15`, or `11.5`?  Presumably it wold be `1.15` but it starts to get tricky
How about `1 1. 1.5*4` is that `1 1.6`?
Then what about `1 1.1.5*4` 
You see what I mean.

Best thing to do is to substitute brackets and provide errors.

pstorch:
> > BTW: adding extra brackets like N 54° (A-1967).((Z*F)+720) works, but without them should, too.

> I don't think "N 54° (A-1967).(Z*F)+720" should be a valid formulation.

Hmm, I just copied it from the geocache description as regular users would do. And it is mathematically correct.

pstorch:
> And what about `1.5 + 10` is that `1.15`, or `11.5`?  Presumably it wold be `1.15` but it starts to get tricky
> How about `1 1 . 1.5*4` is that `1 1.6`?
> Then what about `1 1.1.5*4` 

My brain hurts :face_with_head_bandage:

S-Bartfast:
>My brain hurts 🤕

Yeah, that's my point.

What I'm saying is we need delimiters to know which parts of the string to "evaluate".  Normally this is done with brackets but in this case the brackets are not representing 'sections' to be replaced.  I don't doubt `"N 54° (A-1967).(Z*F)+720"` came directly from a cache's description, I'm saying that it's a format we wont be able to accommodate directly.

"Normal" descriptions are of the form `"S48 51.(C-A)(C+D)(C+A)"`, where `(C-A)`, `(C+D)` and `(C+A)` are evaluated and their results joined together.  In the example you have their there are no delimiters and there for no consistent way to know which parts of `"N 54° (A-1967).(Z*F)+720"` should be substituted.  As it is in this instance it appears everything after the decimal point is to be evaluated as a single expression but we can't always assume that.

The rule as it is now is simply:
>* If it's inside brackets it gets evaluated
>* If it's outside brackets it gets sent to the parser directly

If you have a suggestion for a better rule I'll be willing to hear it but I can't think of a better way myself, and it's not due to a lack of trying.

Matze2:
I agree with @S-Bartfast here. The rule "surround every expression with parenthesis" is clear.
And most of the listings I checked follow this rule.

At the end it is all about transparency again:
"N 54° (A-1967).(Z*F)+720" would be evaluated to something like "N 54° 30.200+720" and this should be indicated as an invalid coordinate.

Lineflyer:
>  I don't doubt "N 54° (A-1967).(Z*F)+720" came directly from a cache's description, I'm saying that it's a format we wont be able to accommodate directly.

Most cache listings will IMHO assume the decimal digits being (virtually) a dedicated number. So couldn't we (always) set our own brackets (I mean in terms of being able to calculate it) around the decimal part, where in this example it would be:
`N 54° (A-1967).(Z*F)+720` --> `N 54° (A-1967).((Z*F)+720)`

This would also give a clear rule for the brainhurting examples above, wouldn't it?

Matze2:
With additional parenthesis around will

> GC768Y3, Stage 10
> N 48° 32.(G-K)(E+E)(C) E 11° 27.(K-C)(E+F+G+H)

still work?

S-Bartfast:
>This would also give a clear rule for the brainhurting examples above, wouldn't it?

Possibly, 
Assuming you consider the first '.' or ',' as the decimal point this would make:
`1 1.1.5*4` -> `1 1.(1.5*4)` -> `1 1.20` - `1 1.020`
but that's quite a leap of faith.  I would think it would be safer to give an error and have the user clarify with brackets ;)

And another example:
`N 1° 2.3(A) * 2`, where `A = 4`
Grouping after the decimal would give: `N 1° 2.34 * 2` -> `N 1° 2.68`
But the cache owner could well have meant: `N 1° 2.3[A * 2]` -> `N 1° 2.38`

Their's just no way to be sure and rather than guess it would be safer to simply require a more explicit formulation.

S-Bartfast:
>With additional parenthesis around will
>>GC768Y3, Stage 10
>>N 48° 32.(G-K)(E+E)(C) E 11° 27.(K-C)(E+F+G+H)
>
>still work?

Sure.

If a formulation was valid to begin with, additional parentheses will never make it _not_ work.
The issue is that inserting parentheses will potentially make invalid formulations appear valid.

pstorch:
>  would think it would be safer to give an error and have the user clarify with brackets ;)

When I proposed this for the leading zero problem, it was rejected. Why can't we then "assume" here the most probable interpretation, too? 
Stopping at the first separator (`.` or `,`) and interpret the rest as a whole.

And with the brackets: evaluate each pair, replace with the result. After that look at the whole string and either calculate (if operators are present, like in `(Z*F)+720`) or concatenate (like in `(K-C)(E+F+G+H)`);

S-Bartfast:
>>would think it would be safer to give an error and have the user clarify with brackets ;)
>
>When I proposed this for the leading zero problem, it was rejected.

I'm not sure why your proposal was rejected.  If anything that proposal would agree with my general philosophy for error detection/handling.  Possibly it was rejected as it differed from the behaviour of the Coordinate Input Dialog, something we were more concerned about before.

>Why can't we then "assume" here the most probable interpretation, too?

Partially it's because the most probable interpretation is not as clear (see `N 1° 2.3(A) * 2` for example) but mostly because in this example those brackets have a different meaning to the regular multi-cache brackets.

Normally, in a multi-cache formulation, brackets are used to designate "substitution elements", that is sections that are to be 'swapped' for evaluated values but that's not what they mean here.

Now, we could define our own rules as to what denotes a 'section' to be 'swapped' (and it seems you are advocating "everything after the first decimal point" as one such rule) but I see this as a road that is fraught with difficulties.  Sure, like @Matze2's "removing spaces after the decimal" suggestion in #6835 it seems such a rule would be simple to include but this really is just to tip of the iceberg.  If we were to define an "after the decimal" rule we really should try to implement something similar for before the decimal as well, which is much harder.

We will also need to consider what to do if evaluating an expression changes the number of decimals as well.  For instance consider the formulation: `N 12 A*3` where `A = 1.5`.  Should this be evaluated as `N 12 1.15` (grouping after the decimal) or `N 12 4.5` (multiplying "A" by 3).  And I could go on...

My point being, while we could potentially write our own "lexer" (as it is known) for coordinate strings this is a bigger task than most would at first think and I for one simply don't think it's worth it.

pstorch:
>  I for one simply don't think it's worth it.

You just don't want headaches, right? :innocent:

I don't know how much effort is worth putting into it. It clearly depends on how much you want to put into it. We all do it voluntarily. 

What is the best we can do right now? Start a release with what we have (incl. this PR) and see what users think about it?

Matze2:
Honestly, I don't understand why all these useful pull requests are still hanging in the queue. I don't understand these theoretical discussions if N 1 or N 1.0 is a valid coordinate. We are not university anymore.
I could name at least one cache which is not covered at all by the current implementation but i won't because I would like to try this PR and hopefully also #6842/#6848 in master at least and not create yet another useless discussion about the perfect coordinate scanning.
Me as a nightly user would like to try these features in practice and not discuss about theoretical coordinate patterns that I will never make use of.
Unfortunately, using cgeo-debug.apk is not an option for various reasons (Google Maps is one of them), otherwise I would have put together my own cgeo since some time and not using nightlies that do not fulfill their purpose anymore.

S-Bartfast:
>>I for one simply don't think it's worth it.
>
>You just don't want headaches, right? 😇

Essentially yeah. It's not that I'm not willing as such, it's more that I would imagine writing such a parser would be a extensive task and I'm not convinced that it would make the product any better.  In fact I actually think it would make the product less good.

If we can define a set of rules that govern how 'groups' should be defined than sure, I'll be happy to implement it.  The problem is that it's actually quite difficult to clearly define exactly what should constitute a 'group'. At the moment the 'groups' are defined by "What's between the brackets", which on the whole is the 'normal' convention used in descriptions, and I think it's a good convention. As it is I did actually foresee the possibility that formulations such as this one may arise so implemented the ability to use 'nested' brackets, along with the use of square and curly brackets/braces.  While this isn't immediately obvious and certainly isn't helped by lack of feedback, it was an attempt to cover exactly this type of unexpected formulation.

>I don't know how much effort is worth putting into it. It clearly depends on how much you want to put into it. We all do it voluntarily.

Sure, but to be fair I have put literally 100's of hours into this project over the past few months.  In fact the only reason I've been able to devote as much time as I have is that I am currently out of work and largely devoting all the time I have to the project (which of course I do willingly) but to imply that the reason I prefer the use of brackets rather than creating a parser to define groups is due to laziness or an unwillingness to put in the effort is a bit disingenuous.  As I say, I would be willing to "put in the effort" if it were clear what the solution should be and that it were to improve the product, neither of which I believe is true at the moment.

S-Bartfast:
>What is the best we can do right now? Start a release with what we have (incl. this PR) and see what users think about it?

I think the calculator is pretty functional at the moment and certainly worthy of being used in the main app.  While I would like to be able to improve the error feedback provided to the user, just to help guide them a little more through these type of unexpected situations, that could be a ways off as it really is dependant upon getting more details from the coordinate parser. 

Also, if someone skilled in the art of web design could write a user guide it would be good to include #6664 (Incorporate usage guide into WP calculator) with the release as well.  While the framework for displaying the page has already been done and I would of course be more than happy to contribute to the technical details I'm afraid I'm not really up to the task of writing the guide itself.  So if someone with the know-how could produce a guide, that would make a really good addition to the release.

S-Bartfast:
>I could name at least one cache which is not covered at all...

Oow, which cache would that be :p

To answer your question though...
The reason we keep getting hung up on these issues is that we do not want to put anything into the field that creates a precedent that needs to be rolled back later.  While we could probably be a little looser with allowing experimental concepts into the nightly build the reason we are reluctant to be too free is because once out there it's much harder to bring back.  As I say, we could probably afford to be a  little freer than we currently are for the nightly, but that's the reason for getting hung up on such seemingly pedantic details.

pstorch:
> Honestly, I don't understand why all these useful pull requests are still hanging in the queue.

Yes, that is a difficult situation right now. I don't like either. There are several things that come together. We want to make a release with the first version of the calculator, but don't want to put more changes in the release as necessary to avoid breaking it. Then it looks like most of us core developers don't have time at the moment to review and test the changes. I feel a bit lonely a at the moment with these tasks and I don't want to decide it all by myself. That's why I ask so much and try to discuss where I am uncertain. Unfortunately that also leads to these difficult  edge cases. 
On the other hand I try to keep it going and move forward as much as I can.

pstorch:
>  I have put literally 100's of hours into this project over the past few months.

It might not look like that when reading my comments, but I really appreciate your contributions.

Matze2:
@pstorch Yes I see that you try to keep things rolling (desperately :wink:).

I just wonder why this couldn't be merged to master at least. Currently master and release are regulated in the same way.

Perhaps all others like me are just too tired to read all these posts and just want to try that feature.
I do not see how the PRs 6836 and 6848 make things worse than it is now in master and release.

S-Bartfast:
>It might not look like that when reading my comments, but I really appreciate your contributions.

Yeah, I have no doubt that you do.
And although it may not look like it, I do actually really appreciate your comments ;)

pstorch:
> I just wonder why this couldn't be merged to master at least. Currently master and release are regulated in the same way.

Because we want to have it in the `release`. When we merge it to `master` first, we can't easily merge it to `release` afterwards without pulling in all other changes in `master` in the meantime. Then we would fiddle around with cherry-pick, which I want to avoid.

pstorch:
Tried it again, now I get this:
image

It doesn't give a wrong result anymore as before. It now gives me the intermediate result and some clue what to do. If I add extra brackets it evaluates the rest, too. 

I still think it would be "nice" to do the rest of the calculation as I would interpret it, but I'm totally fine if we start with this version for the `release`. 

If you can sort out the commits (squash into one and rebase on current `release`) I think we can finally merge it.

S-Bartfast:
>If you can sort out the commits (squash into one and rebase on current release) I think we can finally merge it.

I'm trying to.
I'm having a great deal of trouble removing the "clean-up" commit though.
Would it be okay to squash the 'clean-up' commit into this one?

pstorch:
> Would it be okay to squash the 'clean-up' commit into this one?

Let's give it a try.

S-Bartfast:
>>Would it be okay to squash the 'clean-up' commit into this one?
>
>Let's give it a try.

Works for me (:

pstorch:
Still thinking about. Maybe it's better when I cherry-pick this commit to release, because it is needed for two PRs.

S-Bartfast:
>Maybe it's better when I cherry-pick this commit to release, because it is needed for two PRs.

I don't follow?

So, I've merged the 'clean-up' code in to this commit, and have now rebased the #6842 (Highlighting) onto this release, mainly to take advantage of the 'clean-up'.

Are you talking about excluding the 'clean-up' code again?

pstorch:
This is merge now via PR #6848

PR #6838, opened 2017-11-09
fix #5498 only one loader should be active

### LINKED ISSUE #5498: "Nearby" list sometimes forgets number of caches loaded (truncates to 40 entries)
This is another one which is a little bit annoying if you want to explore "Nearby" caches with c:geo.
This issue exists since a long time.
- Open "Nearby" list
- Load at least 60 caches in the list
- Select one entry to see the details
- Go back

List shows only 40 caches and not the original list.
If you select an entry again to see the details and go back you might get back to the original number of entries.

The already loaded list is always kept and not truncated to 40.
###### Version of c:geo used:

Latest NB.

Yes, almost always.

Device: D5503 (D5503, Sony)
Android version: 5.1.1
Android build: 14.6.A.1.236
Cgeo version: 2016.03.03-NB-46936b1
Google Play services: disabled
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: true
Map strategy: auto
System language: de_DE (cgeo forced to English)
Geocaching sites enabled:
geocaching.com: logged in (Login OK) / BASIC Installed cgeo plugins: none 

Matze2:
See #5498 for steps how to reproduce and how to verify the fix.

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -1540,6 +1540,9 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             showFooterLoadingCaches();
 
             getSupportLoaderManager().restartLoader(CacheListLoaderType.NEXT_PAGE.getLoaderId(), null, CacheListActivity.this);
+            if (type == CacheListType.NEAREST) {
+                getSupportLoaderManager().destroyLoader(CacheListType.NEAREST.getLoaderId());
+            }
         }
     }
 


rsudev:
Please test

pstorch:
test this please

Matze2:
@pstorch
Thanks for testing and approving.

pstorch:
Ok, let's merge it. @Matze2 thanks for your contribution.

Matze2:
Thanks. Looking forward to use a nightly again with Google Map support.

PR #6848, opened 2017-11-19
Resolves #6842: Implement "Result Highlighting" in the Calculator

### LINKED ISSUE #6842: Implement "Result Highlighting" in the Calculator.
As mentioned in #6716 (Managing ambiguous results in WP Calc where trailing/leading zeros are involved) it can be necessary to insert additional zeros into the result of a formulation, and in some instances there is contention as to whether these should be 'leading-zeros' or 'trailing-zeros'.  It has been suggested that a method of identifying how a result has been manipulated would be desirable and a suggestion made is to colour the 'manipulated' digits differently to provide this indication:
image

Developer notes:
>Implementing this is not as trivial as one might at first think for a few reasons:
>
>The main concern is that the text to display for a valid location is actually obtained from the waypoint itself rather than from the input strings used to create it.  This is done specifically so that users can be aware of exactly what the resulting waypoint will actually represent rather than their entered value (which may result in something they did not expect).  Although we have access to both the input string and the resulting GeoPoint's string displaying a 'diff' of these two strings is not as trivial as you may think.
>
>However, once it is known what colours should be used for each letter actually displaying the desired results shouldn't be too hard given TextView's support the use of HTML strings.

Further notes:
>Once implemented this "Syntax Highlighting" would be most useful in conveying information to users of what leads to "Parsing Failures" (as related to #6802: Coordinate Parser Needs Work).

S-Bartfast:
Resolves #6842 Implement "Result Highlighting" in the Calculator.

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -5,6 +5,7 @@ import static cgeo.geocaching.R.id.coordTable;
 import static cgeo.geocaching.models.CalcState.ERROR_CHAR;
 import static cgeo.geocaching.ui.dialog.CoordinatesInputDialog.GEOPOINT_ARG;
 
+import cgeo.geocaching.EditWaypointActivity;
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.AbstractActivity;
 import cgeo.geocaching.location.Geopoint;
@@ -46,6 +47,8 @@ import java.util.regex.Pattern;
 
 
 import butterknife.ButterKnife;
+import cgeo.geocaching.utils.CalculationUtils;
+
 import org.apache.commons.lang3.SerializationUtils;
 import org.apache.commons.lang3.StringUtils;
 
@@ -70,14 +73,14 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     private static final String SYMBOL_SEC = "\"";
     private static final String SYMBOL_POINT = ".";
     public static final String CALC_STATE = "calc_state";
-    public static final String NOTES = "notes";
+    static final char BRACKET_OPENINGS[] = {'(', '[', '{'};
+    static final char BRACKET_CLOSEINGS[] = {')', ']', '}'};
 
     private ImageButton doneButton;
     private boolean stateSaved = false;
 
     private Geopoint gp;
     private CalcState savedState;
-    private String savedNotes;
 
     /** List of equations to be displayed in the calculator */
     private List<CalculatorVariable> equations;
@@ -166,9 +169,9 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         public void onClick(final View v) {
             // Save calculator state regardless of weather the coordinates are valid or not.
             final CalcState currentState = getCurrentState();
-            final String currentNotes = notes.getText().toString();
-            setSavedState(currentState, currentNotes);
-            ((CoordinatesInputDialog.CalculateState) getActivity()).saveCalculatorState(currentState, currentNotes);
+            setSavedState(currentState);
+            ((CoordinatesInputDialog.CalculateState) getActivity()).saveCalculatorState(currentState);
+            ((EditWaypointActivity) getActivity()).getUserNotes().setText(notes.getText());
 
             if (!areCurrentCoordinatesValid()) {
                 return;
@@ -311,9 +314,8 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
     /**
      * @param gp               Geopoint representing the coordinates from the CoordinateInputDialog
      * @param calculationState State to set the calculator to when created
-     * @param notes            user notes
      */
-    public static CoordinatesCalculateDialog getInstance(final Geopoint gp, final CalcState calculationState, final String notes) {
+    public static CoordinatesCalculateDialog getInstance(final Geopoint gp, final CalcState calculationState) {
         final Bundle args = new Bundle();
 
         if (gp != null) {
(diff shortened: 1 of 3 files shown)


S-Bartfast:
Note: This PR includes #6716 (fix for #6800: Leading Zero Padding) as it uses code added for that fix.

Matze2:
Gave it a try. The highlighting is a helpful feedback. 👍

pstorch:
retest this please

pstorch:
Hmm, I tried it locally, but it doesn't compile. Am I doing anything wrong? :thinking:

pstorch:
```
> Task :main:compileBasicDebugJavaWithJavac
warning: A class activity declared in the AndroidManifest.xml cannot be found in the compile path: [android.support.v7.widget.TestActivity]
/home/peter/StudioProjects/cgeo/main/src/cgeo/geocaching/EditWaypointActivity.java:481: error: cannot find symbol
    public ImmutablePair<CalcState, String> fetchCalculatorState() {
           ^
  symbol:   class ImmutablePair
  location: class EditWaypointActivity
1 error
1 warning

```

pstorch:
I have to add this import to be able to build it locally: 
```
import org.apache.commons.lang3.tuple.ImmutablePair;
```

Don't know why it works on CI.

The first thing I tested was this:
image
Then I wondered why I don't see the added leading zeros until I found out that my choosen variables cause the minutes to be out of range. So the leading zeros were not added in this case.
Don't know what we should do in this case. A highlight of this error would be nice of course.

With proper variables it works:
image

pstorch:
I've changed the target to `release` no CI reports the same build error I got locally.

S-Bartfast:
The `import org.apache.commons.lang3.tuple.ImmutablePair;` issue is cause by removing the 'code-cleanup' commit.

pstorch:
I see this PR contains now the two commits we want to have, right?
I'll make a new local test. If it's ok, then I'll merge this one and we can close the other PR.

S-Bartfast:
Sounds good.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -1011,21 +1088,40 @@ private void updateResult() {
...
         doneButton.setImageResource(doneIcon);
-        tLatResult.setText(lat);
-        tLonResult.setText(lon);
+        tLatResult.setText(latFormatted);
+        tLonResult.setText(lonFormatted);
         tLatResult.setTextColor(resultColour);
         tLonResult.setTextColor(resultColour);
     }
 
+    private Spannable formatCoordinateString(String coordinateString, int leadingZeros, int paddingColour)
```
pstorch: I don't want to be picky, but my IDE says final modifiers are missing here. And the opening { is on the next line, should be on this.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -989,7 +1062,11 @@ private void updateResult() {
             resultColour = validColour;
             doneIcon = R.drawable.ic_menu_done_holo_dark;
 
+            String lat;
```
pstorch: final modifier

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -989,7 +1062,11 @@ private void updateResult() {
             resultColour = validColour;
             doneIcon = R.drawable.ic_menu_done_holo_dark;
 
+            String lat;
+            String lon;
```
pstorch: final modifier

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -585,19 +589,28 @@ private void setButtonInputValuesFromGP() {
...
+        }
 
         if (!wholeMatcher.find() || !decimalMatcher.find()) {
             return coordinate;
         }
 
         final String leadingString = wholeMatcher.group(1);
-        String trailingString = decimalMatcher.group(1);
+        String decimnalsString = decimalMatcher.group(1);
+        String trailingString = trailingMatcher.find() ? trailingMatcher.group(1) : "";
```
pstorch: final modifier

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -585,19 +589,28 @@ private void setButtonInputValuesFromGP() {
         }
     }
 
-    private String addLeadingZerosToDecimal(final String coordinate) {
-        final Pattern wholePatern = Pattern.compile("(.*)[.,]");
-        final Pattern decimalPatern = Pattern.compile("[.,](\\d*)");
+    private String addLeadingZerosToDecimal(final String coordinate, boolean lat) {
```
pstorch: `lat` should have a final as well

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -585,19 +589,28 @@ private void setButtonInputValuesFromGP() {
...
+            lonLeadingZerosAdded = 0;
+        }
 
         if (!wholeMatcher.find() || !decimalMatcher.find()) {
             return coordinate;
         }
 
         final String leadingString = wholeMatcher.group(1);
-        String trailingString = decimalMatcher.group(1);
+        String decimnalsString = decimalMatcher.group(1);
```
pstorch: typo

pstorch:
@S-Bartfast thanks for your efforts

PR #6852, opened 2017-11-26
Fixes #6843: Calculator does not save state on pause

### LINKED ISSUE #6843: Calculator does not save state on pause
- Add or edit waypoint, open calculator
- dd mm.mmm format, add some variables
- type some numbers
- press home or switch to another app
- go back to c:geo
- calculator still visible, but all variable buttons are numbers again
- when you tap the buttons, they will turn into variables again and still have their assigned content that you typed previously.

Nightly from today.

S-Bartfast:
Fixes #6843: Calc cleared when restored from backgorund


pstorch:
@S-Bartfast I'm glad it was an easy fix for you. It works. I'll merge it, thanks.

PR #6854, opened 2017-11-26
Fix #6851 implement copy waypoint notes action

### LINKED ISSUE #6851: Add support to copy waypoint (user) notes
While using the waypoint calculator I wondered how I can share the variables gathered from stage to stage among the waypoints for easy transfer in the calculator formulas.

My idea was to create a virtual waypoint called "Variables" to collect all stage variables in its user notes.
From there I wanted to copy them into the stages' user notes. The user notes are also visible in the calculator so they can be copied in the variable values of the waypoint calculator.
In the following screenshot you can see virtual waypoint "Variables" from where the variables were copied into the waypoints "Stage 2" and "Stage 4".

screenshot_2017-11-24-00-04-16

Unfortunately there is no way to copy the user notes from a waypoint. :open_mouth:

I did a rough implementation that creates a new waypoint action "Copy Notes". This action strips HTML from the waypoint note, merges with user note and copies both into clipboard.

Before I continue on the code I just wanted to ask in advance if there is interest and a chance to get this feature included.

Matze2:
Changed behavior:
Open the waypoint context menu by long press:
- Waypoint without (server) note and user note -> no menu item "Copy notes" shown
- Waypoint with (server) note only or user note only -> menu item "Copy notes" shown and note is copied
- Waypoint with both notes -> menu item "Copy notes" shown, notes are concatenated as in "Duplicate Waypoint" and copied

### DIFF
--- a/main/res/menu/waypoint_options.xml
+++ b/main/res/menu/waypoint_options.xml
@@ -19,6 +19,10 @@
         android:title="@string/waypoint_copy_coordinates">
     </item>
     <item
+        android:id="@+id/menu_waypoint_copy_notes"
+        android:title="Copy notes">
+    </item>
+    <item
         android:id="@+id/menu_waypoint_duplicate"
         android:title="@string/waypoint_duplicate">
     </item>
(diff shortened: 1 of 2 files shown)


pstorch:
test this please

Matze2:
This PR was based on the assumption that copying of editable text areas is not possible with all devices.
Instead of a long press, Xperia devices need a double tap on a word of the text.
As a consequence this PR lost its purpose.

PR #6856, opened 2017-11-26
fix #6658 add pq import in cachelist menu

### LINKED ISSUE #6658: Add PQ list to import menu
The import menu should contain an option to open a list with all downloadable PQ. The selected PQ should be imported to the active list (or open the dialog with the active list pre-selected).

pstorch:


main/src/cgeo/geocaching/connector/gc/PocketQueryListAdapter.java
```diff
@@ -14,13 +14,12 @@
 import android.widget.Button;
 import android.widget.TextView;
 
-import java.util.List;
```
UniQP: Just cosmetic: Can you also remove one of the empty lines?

pstorch: thanks for spotting, I'll remove one

PR #6860, opened 2017-11-29
Fix #6857 Add support to delete user defined coordinates from waypoints

### LINKED ISSUE #6857: Provide a way to delete coordinates from waypoints
This was already discussed at the end of #6697.

There are various ways you can get unwanted coordinates in a waypoint:
- you just want to add a note there
- you prepare waypoint calculator formulas in advance

In all these cases arbitrary coordinates (usually your current position) are stored into the waypoint and there is no way to get them removed again.

This has different disadvantages:
- you cannot distinguish between a waypoint with correct or incorrect coordinates
- it pollutes the waypoint map with wrongly placed waypoints

This is of course more problematic for caches with lots of stages: you get a wild mixture of valid and arbitrary coordinates. 

Therefore it would be nice to have an action to remove waypoint coordinates from a waypoint.

The implementation for this is trivial. It acts in the same way as the "Delete" action in the waypoint context menu. This means, there is no confirm dialog and there is no different behavior for user created waypoints or listing waypoints. The latter can be restored by a cache refresh again.

I will share my changes tonight.

Matze2:


Matze2:
Clear/Delete is possible on waypoints with user defined coordinates. The following implementation related items are open:
- How to name that action? Clear coordinates or delete coordinates? Or something completely different?
- This is now the 4th AsyncTask with identical onPostExecute method. We could refactor all four cases to use an AsyncTask subclass ModifyWaypointAsyncTask which implements this common method. Or the AsyncTasks get converted to RxJava.

Lineflyer:
ok to test

pstorch:
> How to name that action? Clear coordinates or delete coordinates?

I would prefer "Clear coordinates"

> This is now the 4th AsyncTask with identical onPostExecute method. We could refactor all four cases to use an AsyncTask subclass ModifyWaypointAsyncTask which implements this common method. Or the AsyncTasks get converted to RxJava.

I get a lot of warnings in Android Studio about the non-static AsyncTasks. I would prefer RxJava, as we did in other locations in the code. But changing all AsyncTasks in this class would be a separate refactoring.

> Open a confirm dialog for both actions "waypoint delete" and "waypoint coordinate delete/clear". Once the context menu is open this can be easily clicked and the change is persisted immediately, i.e. important data might get lost.

We are striving towards less dialogs. What about an undo action, like we have in some of the list operations?


I also tested this PR. Works for all three cases. :+1:

Matze2:
Yes, introducing undo sounds good. Looks like a bigger change to be done in a separate issue, doesn't it?

pstorch:
> Yes, introducing undo sounds good. Looks like a bigger change to be done in a separate issue, doesn't it?

It's not that big of a change, but I don't mind doing it separately.

Matze2:
I am currently implementing it. 👍

Matze2:
I've added also an abstract WaypointModificationCommand. Probably we want to convert the Delete action, too.

pstorch:
I like it. If you squash the commits into one, we can merge it.

The delete action can be done in a separate issue / commit.

Matze2:
Ok. Squashed.

pstorch:
Ok, let's merge. Thanks for your contribution.

Matze2:
Thanks for the review and the merge. Maybe I am working on the undo funcionality for waypoint delete in the next days.
Currently I am busy with solving mysteries for our local Adventskalender :sunglasses:

Issue #6861, opened 2017-12-02
Support of NULL coordinates in EditWaypointActivity and don't prefill with user location


pstorch:
To be able to handle invalid coordinates (for #6697) and to prevent the misbehavior found in #6857 (e.g. opening a WP without coords, only updating notes and saving it), I would like to change the EditWaypointActivity to properly support NULL (a.k.a. empty) coordinates in this dialog.

At the moment, as soon as you open this dialog and the WP doesn't have any coords (the field is NULL) the coords are filled with the current location of the device.

I guess this was done for convenience when adding a new stage for a multi cache for example. But in all other cases (e.g. calculating the next stage with the calculator or just editing a user note for a waypoint without coordinates) it is confusing that the coords are set to the users location without intention. 

I propose to set the lat/lon buttons to a text like `NOT SET`, `---` or `N 00° 00.000` to clearly indicate that there are no coordinates set. Only fill the coords by explicitly using the Coordinate Input Dialog, the Calculator or the Clipboard.

SammysHP:
Although highly unlikely, `N 00° 00.000` is a valid format. `N --° --.---` or `N __° __.___` sounds good IMHO.

pstorch:
Yes. Maybe we should avoid all `N ...` variants. Because a fixed hemisphere could be confusing or we would have to dynamically set the hemisphere according to the users location. Which might be unnecessary complexity.

Matze2:
In my opinion this is a big step forward. 👍 

What is still missing in my opinion (which could be also part of other issues):

1. Obvious invalid variable values should make the calculator result invalid too.
Example: E 11° 28.(A-1)(A)0 with A > 9 (leads to 4 digits) or A < 1 (leads to negative digits)

2. If calculator result is invalid exiting the calculator is not intuitive since you need to click the "disc" for saving and then exit it with the "cross". The right button (disc or check) should always save and exit the calculator. Then the "cross" will be a pure cancel button.
If this is not clear I can give more details with some UI screenshots.

pstorch:
Thanks for your feedback.

Regarding 1. and 2. there is another issue #6855 on how the calculator should handle invalid input/results.

Regarding 3. yes, that I've missed, too. Maybe in the CoordsInputDialog or an indicator in the Waypoint itself that it is based on a calculation.

Matze2:
Which brings us back to my opinion that a waypoint is either in "direct input" or in "calculator" mode.
But that's a different story. 😃

[...]

PR #6862, opened 2017-12-02
fix #6844 confirm discard wp changes

### LINKED ISSUE #6844: Confirmation before discard waypoint changes
It's annoying when you have entered a lot of data for a waypoint and accidentally press back. Currently everything is lost. c:geo should show a confirmation instead.

pstorch:
Adding the confirmation dialog was not an issue. But I saw that the calculator state was not included in the detection of changes in `isWaypointChanged(currentState)`. Adding it was a bigger challenge. 
Long story short I decided to go with the json string to compare before and after. Otherwise I would have to implement a deep copy and compare of this object. 
@S-Bartfast: maybe you can give some feedback here. Better ideas are welcome.

Another issue is the prefill of the coordinates if they are NULL which we want to tackle in #6861. At the moment the user needs to confirm an unsaved change here, even if he didn't change anything on purpose.

Sometimes there is a rounding issue in the Geopoint(Parser) class which results in a failing equals test even if the coords haven't been changed and they look similar. In the database we have lat/lon as doubles convert them to a String for the buttons and parse them back.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -257,6 +257,8 @@
     <string name="warn_calculator_state_save">Be sure to save waypoint if you want to keep the calculator state.</string>
     <string name="err_trackable_log_not_anonymous">Logging %1$s from c:geo cannot be done anonymously. Do you want to open the settings to fill in your credentials from %2$s?</string>
     <string name="err_trackable_no_preference_activity">Sorry, cannot find a suitable preference screen.</string>
+    <string name="confirm_unsaved_changes_title">Unsaved changes</string>
+    <string name="confirm_discard_wp_changes">Discard unsaved waypoint changes?</string>
     <string name="warn_discard_changes">Unsaved waypoint changes discarded!</string>
     <string name="err_request_popup_info">c:geo failed to get cache popup info.</string>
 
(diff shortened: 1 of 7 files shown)


pstorch:
Still thinking about the (de)serialization of the calculator state. We only need the Java Object `CalcState` inside the Calculator. Maybe we should just transport it as json String anywhere else. 
Any opinions?

pstorch:
I've decided to go with the JSON String variant in the Waypoint class. It makes the code easier.

pstorch:
Interesting insight by Codacy:

The new method `createWaypointValues()` has an NPath complexity of 625 which counts as a new issue. But on the other hand with this method extraction I resolved NPath complexity of `saveWaypointsWithoutTransaction()` with 2502 and `saveWaypointInternal()` with 10004.

pstorch:
I've added some extra logic to solve the coords rounding problem I described above. As we always convert the coordinates to the DEC_MINUTES format with 3 decimal digits for the buttons, we don't support higher precision than that. If a waypoint had a higher precision then the waypoint was treated as changed, because of the conversion. This could cause confusion because the user might not have changed the coords.

I now compare the coords in the DEC_MINUTES format as visible on the buttons.

pstorch:
Ok, I'll merge here. If someone sees something awkward or has some comments later on, please don't hesitate.

Issue #1236, opened 2012-03-06
Fail to handle error on gc.com


campbeb:
While not logged in I tried to load (by entering geocode) an event cache which had been archived.
I received this:

![Error](http://i.imgur.com/OwTcY.png)

I was not able to reproduce it, so I think it was a one-time error on gc.com.

PR #6867, opened 2017-12-05
fix 6865 waypoints not parsed and 6866 archived/disabled state

pstorch:
There is one failing test: `CgeoApplicationTest.testSearchErrorOccured(CgeoApplicationTest.java:193)`

I'm not sure what this test should do. I suspect something with member status, but it's too late now.


pstorch:
WhitespaceTest? Too late to look into.

pstorch:
WhitespaceTest fixed. Next I'll look at the last failing test.

pstorch:
Looks like the error case which was handled by @Bananeweizen in commit e6a4214b304f3441f46bf75be85b1f538d3e2d7e for issue  #1236 doesn't happen anymore. The geocache which was tested here can be displayed without being logged in.
I've removed this testcase. If I'm wrong, please let me know.

Issue #6833, opened 2017-11-08
Combining Input Coordinates Dialog and the Calculator into a single dialog.


S-Bartfast:
This issue is being forked off from #6823 as #6823 was intended to just be a simple modification to clean up code rather than a thread to discuss improvements of the user experience itself.

The main discussion in #6823 was about the convoluted process one needs to go through in order to reach the calculator and two solutions were suggested.

One, suggested by @Matze2, was to add a checkbox to the waypoint editor which could be used to indicate whether the Coordinate Input Dialog or the Calculator should be opened:
![image](https://user-images.githubusercontent.com/12542802/32419940-15a4ce02-c283-11e7-9c3d-d7402452a845.png)

The other, suggested by @pstorch, was to combine both the Coordinate Input Dialog and the Calculator in to a single Dialog.

Personally I'm in favour of the single dialog idea (which is why I've named this issue as I have) but I'm open to discussion.

Matze2:
Just to explain quickly why I would prefer two use case oriented dialogs:
- Users that do not use the calculator do not need to learn anything new. Of course the default is to enter the Coordinate Input Dialog as before
- I fear a combined dialog gets overloaded and the editor state will be difficult to understand for users and to manage for developers (but maybe you surprise me :) )
- I think that bearing and distance fields should move into both dialogs which makes them even bigger,
- There should be a way to clear the coordinates of a waypoint. Again yet another button needed at least for the "direct input" case.

pstorch:
Actually my idea was to merge the calculator with the edit waypoint activity and get rid of the coordinate input dialog if it is a calculated waypoint.
I haven't made up my mind where to put the bearing/distance fields, yet.

pstorch:
At the moment I see three waypoint edit modes:
- direct input
- calculated
- projection

As mentioned in comments https://github.com/cgeo/cgeo/issues/6861#issuecomment-351431242 and following.

Matze2:
Because this **the** issue about frontend design about the calculator let me mention that I am desperately missing and indicator in the waypoint list if this waypoint is in "calculator mode" or not.
While preparing all formulas for a multi I entered lots of waypoint multiple times because I did not remember if already prepared the formula.
It would be nice to have indication like the visited state directly in the waypoint list.

Billy-Wallace:
I agree that the initial 'Add Waypoint' GUI should not combine the 'Calculator' so as to keep things simple for the majority of users.

[...]

Issue #6697, opened 2017-08-19
Flag invalid calculated coordinates


pstorch:
##### Detailed steps causing the problem:
- Add a waypoint with calculated coordinates
- Enter some formulas/values which result in errors
- The preview below the variables show, that there is something wrong, either with an '*' or with a negative value
- Save the current state and go back to the waypoint

##### Actual behavior after performing these steps:
- Valid coordinates are shown (some digits are 0 but valid)
- There is no indication in the waypoint or in the waypoint list that there is something wrong with the coordinate

##### Expected behavior after performing these steps:
- Flag the waypoint as errornous

##### Version of c:geo used:
- nightly

##### Is the problem reproducible for you?
Yes

##### Other comments and remarks:
Errors in the calculator can be missed easily. When changing a variable value the keyboard is overlayed and the coordinate result is not visible. If the user just clicks the checkmark to close the dialog without first closing the keyboard to check the result, an error is missed. It's not visible in the other dialogs, that this waypoint has a problem.

S-Bartfast:
Yes, the error detection most certainly leaves a bit to be desired.

As it is I simply lifted the exact same checking algorithm as used in the regular coordinate entry validation.  I tried to follow what was going on inside that checking algorithm a few times but got so thoroughly lost I just threw my hands up and assumed "If it's good enough for them..."

While it leaves a lot to be desired try entering the exact same text into the 'plain' format of the regular coordinate input and see if it's accepted by that parser.  If you're getting different behavior I'll look into it more closely but if it's the same than I dare say it's a deeper issue with parsing in general and not specifically related to the calculator as such.

S-Bartfast:
Commit [78e84cf489e51200adf4eb89c1e998b7827ff774](https://github.com/cgeo/cgeo/commit/78e84cf489e51200adf4eb89c1e998b7827ff774) (Implemented showing a "Save" icon to replace the "tick" icon for the "Done" button when the resulting coordinates are invalid) should have gone some of the way to resolving this issue.

Also, as part of the work for #6716 the asterisk ("*") and underscore ("_") characters are now considered invalid characters so far as the calculator is concerned.

Do these fixes address the concerns of this ticket?

pstorch:
Not really. As long as the calculator can't finish its job, the waypoint has "some other" coordinates.
I don't see in the waypoint list, that these coordinates are not "ready". 

We have waypoints without coordinates. Maybe it's better to not show coordinates as long as they are not completely calculated.

[...]

PR #6883, opened 2017-12-12
fix #6861 support null coords in waypoints

### LINKED ISSUE #6861: Support of NULL coordinates in EditWaypointActivity and don't prefill with user location

pstorch:
The EditWaypointActivity is not anymore prefilled with the users location. Placeholders are visible on the latitude and longitude buttons.
If the user just wants to have the current location he has to open the CoordsInputDialog once and confirm.
The projection is only enabled if coordinates are set on the waypoint. Otherwise the projection fields are disabled. We can't do a projection from null coordinates.
I've also added a "Clear Coordinates" button on the CoordsInputDialog for convenience.
Invalid coordinates from the Calculator are now forwarded as null to the waypoint. No longer the users location is used in the meantime. This solves #6697 for me. I don't know if someone needs more visual clue.

Unfortunately the "simple" usecases like recording my current position as waypoint or a simple projection from my location needs two more clicks now.

What gets more apparent now is the fact that the CoordsInputDialog is in between the EditWaypointActivity and the Calculator. You can have null coordinates now in the EditWaypointActivity because of an invalid calculation, but when you navigate through the CoordsInputDialog you see your current location as coordinates but don't use it.

Maybe a future refactoring for e.g. #6833 can simplify the dialog(s).

Null coordinates with placeholder text on the lat/lon buttons:
image

CoordsInputDialog:
image

### DIFF
--- a/main/res/layout/coordinatesinput_dialog.xml
+++ b/main/res/layout/coordinatesinput_dialog.xml
@@ -46,6 +46,14 @@
             android:visibility="gone" />
 
         <Button
+            android:id="@+id/clear"
+            style="@style/button_full"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:text="@string/waypoint_clear_coordinates"
+            android:visibility="gone" />
+
+        <Button
             android:id="@+id/done"
             style="@style/button_full"
             android:layout_width="fill_parent"
--- a/main/res/layout/editwaypoint_activity.xml
+++ b/main/res/layout/editwaypoint_activity.xml
@@ -34,7 +34,8 @@
                 android:id="@+id/bearing"
                 style="@style/edittext_full"
                 android:hint="@string/waypoint_bearing"
-                android:inputType="numberDecimal" />
+                android:inputType="numberDecimal"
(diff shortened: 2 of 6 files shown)


pstorch:
I'll merge here, too. To test this in master.

PR #6893, opened 2017-12-21
fix #6889 treat locked caches as archived

### LINKED ISSUE #6889: Handling of locked caches
Example: https://coord.info/GC2PKP3

image

HTML fragment:
```
<div id="ctl00_ContentBody_lockedMessage" class="alert alert-info">
        This cache has been locked, but it is available for viewing.
    </div>
```

When trying to log, c:geo shows error message `c:geo can't load data to required to log cache` but keeps all logtypes available.

pstorch:
As c:geo doesn't know locked status and this is rather seldom I would propose to treat it like archived. Most of the time it superseeds archived anyway (like in the example of this issue). There are exceptions though.


Lineflyer:
Are there really exceptions? I thought only archived caches might be locked in case of need.

pstorch:
One famous exception: https://coord.info/GC1BE91

But no "normal" player is able to get there :wink:

pstorch:
Any objections to do it this way? Otherwise we would need to introduce a "locked" status type and find out another way to recognize the secondary state "Archived" or "disabled".

Lineflyer:
Lets treat it as archived. If this creates confusion for the user (wait for feedback) we can rethink it, but I doubt this will come to attention.

pstorch:
OK, let's go for it

PR #6894, opened 2017-12-22
6868 present disabled archived locked messages

pstorch:
Here is now a coded version of the idea on how to present disabled, archived and locked messages in the cache details.

I've added `Do not merge`, because it is not ready yet. It is also based on #6893 which is not merged.

The messages can get quite long. The website is truncating them. I have no solution for it yet.

Here are some example screenshots:

screenshot_20171222-215614
screenshot_20171222-215857
screenshot_20171222-220427
screenshot_20171222-220641
screenshot_20171222-221531

Should we go on with this? What do you think?

If you want to try it yourself: beware that it changes the DB schema!

### DIFF
--- a/main/res/drawable/alert_border.xml
+++ b/main/res/drawable/alert_border.xml
@@ -1,0 +1,4 @@
+<shape xmlns:android="http://schemas.android.com/apk/res/android" android:shape="rectangle" >
+    <solid android:color="@android:color/transparent" />
+    <stroke android:width="3dip" android:color="#ff0000"/>
+</shape>
--- a/main/res/layout/cachedetail_details_page.xml
+++ b/main/res/layout/cachedetail_details_page.xml
@@ -16,7 +16,7 @@
             android:id="@+id/alert_message"
             android:autoLink="web|map"
             android:layout_width="fill_parent"
-            android:layout_height="wrap_content"
+            android:layout_height="@dimen/cache_details_alert_height"
             android:focusable="false"
             android:textColor="?text_color"
             android:textSize="14sp"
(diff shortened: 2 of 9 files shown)


SammysHP:
I'm more interested in the details which are moved to the bottom. I still think that it is better to link the status to the corresponding log if available.

pstorch:
> I'm more interested in the details which are moved to the bottom. 

Ok, what about a smaller version which can be expanded/collapsed on tap? 
screenshot_20171223-142847

> I still think that it is better to link the status to the corresponding log if available.

I don't know if we can always match it to the correct log. We only read some of the latest logs.

SammysHP:
Looks good. Do you think it is feasible to add some kind of indicator that the area can be expanded? Like fading out the last line of text.

pstorch:
Fading out was a bit tricky:
screenshot_20171223-213734

But I'm still not happy with the whole thing. At the moment we can read this information only when retrieving the cachelisting from the website. The locked status and the disabled/archived messages are missing from the search and the GPX imports. Then we would have to solve the merging of the incomplete loaded geocache with probably older but complete data from the database. The locked status should be in the database, too. 

I have to think about it a while. Don't know if I continue here.

Lineflyer:
Mmh. I personally think, that the current representation of archived or disabled is quite OK (despite a general overhaulof the complete UI). I also did never receive any complaint about users unawareof archived/disabled representation.
This given, I would have assumed we do only show Locked as a new value in the existing state field and thats it.

pstorch:
I initially thought it would be a good idea to present these messages similar to how GS is doing it, so that important information are visible right away. But after my experiment so far I don't think it is worth the effort.

PR #6900, opened 2018-01-01
fix #6898 coords input zero padding

### LINKED ISSUE #6898: Zeros inserted in coord editor when changing format
- Create new WP
- Open the coordinate editor
- It will come up empty (NULL)
- Change randomly between the possible formats

At some point some fields will be filled with zeros instead staying empty

All field should remain empty (NULL)

##### Version of c:geo used:
2017.12.31-NB

Yes

I cant see any functional impact, so it seems only to be a small UI glitch

pstorch:
I tried to fix this, so that no zero padding happens on format switch when there is no input at all in the fields. But padding on input done (check mark). The only fields I left unpadded are the degree fields. I believe these are the only mandatory fields of a coordinate which shouldn't be set to all zeros when left empty.


pstorch:
retest this please

pstorch:
Let's test it in the nigthlies

PR #6909, opened 2018-01-08
fix #6903 simplify calculator save state

### LINKED ISSUE #6903: Changing only notes in waypoint calculator does not set the dirty state
- Open waypoint calculator; the calculator must be in "incomplete" state where the disc icon is displayed
- Press Save (disc button); disc button gets disabled
- Edit notes within the waypoint calculator

Disc button stays disabled; saving is not possible

It should be possible to edit only calculator notes and to save it.

##### Version of c:geo used:
2018.01.04-NB

Yes

pstorch:


Bananeweizen:
👍

PR #6917, opened 2018-01-19
#6259 reset cookies on GCLogin failure

pstorch:
As discussed in #6259 this PR clears the cookies when a GCLogin failure is happening.

I don't know if this finally solves the issue because I cannot reproduce it myself. As users report that it is happening if c:geo wasn't used for quite some time. So we suspect stale session keys in the cookies.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCLogin.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLogin.java
@@ -76,6 +76,14 @@ public class GCLogin extends AbstractLogin {
     @Override
     @NonNull
     protected StatusCode login(final boolean retry, @NonNull final Credentials credentials) {
+        final StatusCode status = loginInternal(retry, credentials);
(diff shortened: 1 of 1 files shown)


pstorch:
Do we first want to test it in the nightlies for a while or should it be included in the next release?

Lineflyer:
Difficult to test the error case, but we should verify that we have no regression.
As we currently have a beta out we should put it into release, publish a new beta.

We can still also test it in the nightlies in parallel.

Lineflyer:
Questions:
AFAICS it resets on all errors.
Thats fine unless this reset would also delete name/password?!

So if e.g. the password is incorrect, it should not harm to delete the cookie but we should never delete username and password.

pstorch:
> So if e.g. the password is incorrect, it should not harm to delete the cookie but we should never delete username and password.

No, username and password are not deleted. Only the login status is resetted.

Issue #6594, opened 2017-06-06
Make Mapsforge Beta default


Lineflyer:
IMHO we already had such an issue, but I failed to find it.

Despite the [known issues](https://github.com/cgeo/cgeo/issues?q=is%3Aopen+is%3Aissue+label%3A%22New+Map%22) for the Mapsforge Beta implementation, I would like to initiate a discussion about making it live asap.

Main reason is, that user complaints/questions are rising as they load up to date maps (I guess e.g. from openandromaps.org) which do no longer work in c:geo as they changed to a newer mapsforge rendering meanwhile (refer to mail from openandromaps admin I forwarded to cgeo-dev mailing list some weeks ago).

Besides the above mentioned small issues, the only big open point is, that once mapsforge beta is active, the google maps are no longer selectable. In case this can be solved (enabling them again as they are while the new mapsforge is active) I would prefer to switch the library rather soon.

Any opinions?
@rsudev (or anyone capable) Would you have a little time for that (and maybe for some of the small issues, which you consider a drawback)?

rsudev:
Unfortunately I was (and partly still are) very much tied up in work for a long time, so I just managed to follow the discussions superficially.
I have an idea how to integrate the two again (we even now re-start the map activity under the hood when switching), but this will not fill the feature-gap between the two (cache selection pop-up, target-mode, waypoint through long-press) and I think we need a good way to make this distinction more prominent to manage expectations and not create to many questions from users.
I will start to work towards the integration, hopefully someone comes up with a good idea for the presentation.

SammysHP:
The most important feature is still missing: map feature selection. Many maps do not work without it because of too many or overlapping features. There is also a discussion about multi-language. So before the beta map will be made default at least the standard mapsforge features should be implemented.

IIRC the attribution is still missing, we shouldn't make the same mistake again and add it.

Or in other words: We need someone with plenty of time to finish the new map implementation.

rsudev:
Regarding the map feature selection: should we really just insert the huge selection menu tree into our map menu? Or is there a better idea, how to present that?
AFAIK, openandromaps work well without the feature selection, 'freizeitkarte' IIRC not really (which is IMO a design fail)

Lineflyer:
I have no idea about the content of this feature selection. How many items are on it?
Anyway in a first step we can add it to the menu like we do for the theme selection and later improve it.

SammysHP:
10-40 items, depends on the map.

[...]

Issue #6149, opened 2016-11-20
Zoom factor not consistent between maps and map sources


rsudev:
When switching between mapsforge beta offline and other map sources or maps, the zoom factor is not consistent. The mapsforge beta offline is much more zoomed in.

Matze2:
So here are my first findings. Here are three screenshots of the Live Map with the following sources:
- Google Maps
- Freizeitkarte_BAYERN with theme freizeitkarte-v5.xml
- Openandromaps Bayern with theme Elevate_Hiking_XL.xml

Here are my findings:
- As soon as the Map is switched the OSM maps do not show caches anymore
  - even moving or zooming in/out to a limited extend to force a refresh does not help
  - a big zoom out shows the caches again at a certain level
  - Google Map keeps the caches
- the zoom between the OSM based maps stays the same; it differs from Google Maps; 
  - the zoom center seems to be stable

Experimenting with extreme zoom in/out and switching maps leads to some other interesting effects, e.g. caches are shortly shown and then hidden.

Personally I see the inconsistent zoom not that critical. The fact that caches are not shown after switching maps is more annoying.

![screenshot_2018-02-15-08-53-54](https://user-images.githubusercontent.com/12542802/36246204-f1f32b9c-122e-11e8-840a-8d5db9a20f72.png)
![screenshot_2018-02-15-08-53-33](https://user-images.githubusercontent.com/12542802/36246209-f7be80e4-122e-11e8-9034-f0f6e175f220.png)
![screenshot_2018-02-15-08-53-03](https://user-images.githubusercontent.com/12542802/36246211-fb1a5e20-122e-11e8-8905-e3e381d59d12.png)

Matze2:
The issue with different zoom levels are independent of the Mapsforge version.
The disappearance of cache icons on map source switch do not occur with the "Old Mapsforge V3" mode.

Lineflyer:
@Matze2 

> The disappearance of cache icons on map source switch does not occur with the "Old Mapsforge V3" mode.

Thanks for your feedback. Can you open a dedicated issue for this problem please?

Matze2:
Yes, will do tonight.

pstorch:
BTW: we also have the new Google Maps v2 PR open: #5973 
Does it make sense to integrate it first and fix the zoomlevel then? Otherwise there might be another difference, then we have to fix it twice. On the other it might be a bigger risk.

Lineflyer:
[...]

PR #6918, opened 2018-01-20
#6594 use mapsforge beta as default

pstorch:
Moved the BetaMap preference into the MapsPreferences section and reverted its logic. 
Now a user can choose to use the Old Mapsforge API as fallback if something is wrong. The new map is default now.

Google Maps is now selectable in the MapsMenu regardless of the chosen Mapsforge API.


rsudev:
The general approach is spot-on. There are two small issues that I observed when looking briefly.
The zoom levels do not match up between google and mapsforge (switching from google to mapsforge zooms in considerably. this is likely my fault, as the zoom-level definition/calculation/adaption likely is wrong).
Switching back from google to mapsforge creates an empty map for me (no caches loaded), probably due to the missing MapOptions in the intent when coming from there, I can take a deeper look into it later today.

pstorch:
@rsudev thanks for your help. The part with the maps in c:geo is still new to me.

Lineflyer:
> Switching back from google to mapsforge creates an empty map for me (no caches loaded)

IMHO we have a dedicated issue for that.

pstorch:
> > Switching back from google to mapsforge creates an empty map for me (no caches loaded)
>
> IMHO we have a dedicated issue for that.

Found it: #6149

pstorch:
I tried to look into the zoom and missing caches issue reported by @rsudev.

Not much success. I see that the old mapsforge view has an offset of 1 to the google zoomlevel (which we store in the preferences) and the new map uses an offset of 3. If I change it to 1 I see other issues.

Regarding the missing caches when switching back from the google maps: I could only partly reproduce. In non-live-mode it was not an issue. In live mode it heavily depended on how I zoomed and scrolled around on the map. Sometimes the caches vanished and when zoomed or scrolled, they suddenly reappeared. 

The only thing I fixed was a possible nullpointer which I saw in the log while testing.

In the meantime mapsforge is already at release 0.9.1, we are still using 0.8.0. Should we upgrade?

rsudev:
I had also looked into the zoomlevel issue and I have the feeling, that mapsforge and google maps define and use this differently. MAybon google maps in relation to the display and mapsforge in relation to the tiles. The reload issue is likely independent from the map switch as such and of secondary importance IMO.
Hopefully I can give the zoom issue another spin tomorrow...

rsudev:
And yes, we should look into upgrading IMO.

pstorch:
I've included the upgrade to Mapsforge 0.9.1 as a separate commit and I think I found the error that the zoomlevel increased on every change between google maps and mapsforge.
I think the maps from offline lists work quite ok. The main problem for me is the livemap. On the new mapsforge map the markers suddenly disappear. Then you have to zoom or scroll around to get them loaded again. Don't know yet what is causing it.

Matze2:
Just to get some movement back on this PR.
As soon as this one is merged to master I would give it a try.

rsudev:
I already started several attempts to tackle the zoom problem in its various aspects, but did not yet reach anything which worked consistently across screen resolutions and map types.
But I would be fine with giving this a spin on master, that makes it also easier to address the upcoming issues one at a time and broaden the testing as well.

Lineflyer:
+1

PR #6928, opened 2018-02-03
fix #6922 stored months/year ago

### LINKED ISSUE #6922: Replace stored 123 days ago by months and years
We should adapt the "stored X days ago" by months (2 to 11) or years plus months for even longer periods.

pstorch:
Adds the following cases:
- "about 1 month ago"
- "about %d months ago" (from 2-11 months)
- "over a year ago" (if more than 365 days ago)

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -2392,17 +2392,25 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
         offlineRefresh.setOnClickListener(refreshCacheClickListener);
 
         if (cache.isOffline()) {
-            final long diff = (System.currentTimeMillis() / (60 * 1000)) - (cache.getUpdated() / (60 * 1000)); // minutes
+            final long minutes = (System.currentTimeMillis() - cache.getUpdated()) / (60 * 1000);
+            final long days = minutes / (24 * 60);
+            final float daysPerMonth = 365F / 12; // on average
 
             final String ago;
-            if (diff < 15) {
+            if (cache.getUpdated() == 0L) {
+                ago = "";
+            } else if (minutes < 15) {
                 ago = res.getString(R.string.cache_offline_time_mins_few);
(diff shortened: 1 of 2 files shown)


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2395,14 +2395,20 @@ static void updateOfflineBox(final View view, final Geocache cache, final Resour
             final long diff = (System.currentTimeMillis() / (60 * 1000)) - (cache.getUpdated() / (60 * 1000)); // minutes
```
Bananeweizen: performance wise it would be nice to divide only once after subtracting

Bananeweizen: if we rename "diff" to "minutes" and introduce another statement "long days = minutes / (24 * 60);" then...

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2395,14 +2395,20 @@ static void updateOfflineBox(final View view, final Geocache cache, final Resour
...
+                ago = "";
+            } else if (diff < 15) {
                 ago = res.getString(R.string.cache_offline_time_mins_few);
-            } else if (diff < 50) {
+            } else if (diff < 60) {
                 ago = res.getQuantityString(R.plurals.cache_offline_about_time_mins, (int) diff, (int) diff);
             } else if (diff < (48 * 60)) {
                 ago = res.getQuantityString(R.plurals.cache_offline_about_time_hours, (int) (diff / 60), (int) (diff / 60));
-            } else {
+            } else if (diff < (365F / 12 * 24 * 60)) {
```
Bananeweizen: ...all the comparisons and quantity string arguments from here on would become much more readable by simply using the "days" variable and leaving away " x 24 x 60"

pstorch:
@Bananeweizen good suggestions. I'm thinking about if it is worth extracting it into its own method to have it testable.

Bananeweizen:
feel free to extract it for testing, independent of this merge

PR #6941, opened 2018-03-02
fix 6931: provide user actions also for OC

Bananeweizen:


Bananeweizen:
I intend to disable the check for final modifiers, so the codacy result is superfluous.

PR #6942, opened 2018-03-03
remove superfluous code

Bananeweizen:
The fields of the result are unconditionally set again after the removed
pieces of code, therefore the removed code doesn't have any effect on
the result.


Bananeweizen:
I was trying to understand when personal notes can start having 2 versions (and the user sees them merged). So I was reading that part multiple times and suddenly noticed this...

Issue #6933, opened 2018-02-15
Caches disappear on map switch with new mapsforge beta


Matze2:
##### Detailed steps causing the problem:
- Open live map with an offline OSM map (Freizeitkarte or Openandromaps with freizartkarte-v5.xml theme)
- Switch to Google Maps; caches are still there
- Switch back to OSM map

##### Actual behavior after performing these steps:
Caches disappeared. Whether they are saved or not is irrelevant.
The OSM theme used seems to be irrelevant, too.

##### Other strange behavior (not sure if this is all the same problem or we need separate issues)
- Zooming out to "2km" brings the caches back, zooming in to "1km" make them disappear again.
- If automatic rotation is active and the device is rotated then the map zooms out by 3 levels. Rotating back rotates again by 3 levels. At the c:geo crashes.
- Zoom out button is disabled at "10km", but using two fingers it is possible to zoom out to "5000km" or even to "10000km"

All tested using 2018-02-14-NB1.

Lineflyer:
I thought this appears when changing between two offline maps.

The case of switching between Google and mapsforge now simply starts the other activity....but this should also populate the map...mmh.

Matze2:
Just checked. Yes it is just switching. It also happens when switching between two offline maps.

pstorch:
The map crashes could be related to: https://github.com/mapsforge/mapsforge/issues/957
At least this is what I see in the logs when switching the map source.

Lineflyer:
I used the new mapsforge very intensivly to prepare a cache trip last weekend.
I also recognized a problem with disappearing caches (without any switching between map providers).

It is somehow like parts of the map are being populated with caches and a small move of the map makes them disappear in a way which looks randomly. Moving the map again makes them reappear and then again other caches disappear once the loading is finished.

@Matze2 Is this, what you also described above?

Matze2:
As I've wrote in the other issue. I've just updated to the latest NB and will recheck in the next days.

Pure moving without switching was not really a problem for me (at least not that big that I remember it).

Matze2:
I just upgraded to the latest NB and the new Map feels much better now. The issue seems to be gone. 👍

PR #6948, opened 2018-03-05
fix #6311 max zoomlevel for osm:mapnik

### LINKED ISSUE #6311: OSM Online: Max zoom level does not provide map tiles
- Select Mapsforge Beta
- Select OSM:Mapnik or Cycling
- Open live map
- Use zoom control to zoom in

At zoom >200m/division the map disappears

On non-beta 200m/division is maximum. On beta the zoom maybe needs to be limited to that as well I think.

##### Version of c:geo used:
2017.02.10-NB

Yes

pstorch:
It also addresses "Zoom out button is disabled at 10km" as reported in #6933.

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -94,6 +94,7 @@ import org.mapsforge.core.model.LatLong;
 import org.mapsforge.core.util.Parameters;
 import org.mapsforge.map.android.graphics.AndroidGraphicFactory;
 import org.mapsforge.map.android.graphics.AndroidResourceBitmap;
+import org.mapsforge.map.android.input.MapZoomControls;
 import org.mapsforge.map.android.util.AndroidUtil;
 import org.mapsforge.map.layer.Layers;
 import org.mapsforge.map.layer.cache.TileCache;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
```diff
@@ -639,17 +639,23 @@ private void switchTileLayer(final MapSource newSource) {
...
-
+                final MapFile mapDataStore = new MapFile(mapFile);
+                newLayer = new RendererLayer(tileCache, mapDataStore, this.mapView.getModel().mapViewPosition, false, true, false, AndroidGraphicFactory.INSTANCE);
+                final MapFileInfo mapFileInfo = mapDataStore.getMapFileHeader().getMapFileInfo();
+                zoomControls.setZoomLevelMax(mapFileInfo.zoomLevelMax);
+                zoomControls.setZoomLevelMin(mapFileInfo.zoomLevelMin);
             }
         } else {
             this.mapView.getModel().displayModel.setFixedTileSize(256);
             if (newSource.getNumericalId() == MapsforgeMapProvider.MAPSFORGE_MAPNIK_ID.hashCode()) {
```
Bananeweizen: Your change is perfectly okay. I'm just wondering if this condition can be not fulfilled at some point in time. And if we have to do something about it (cause I would expect one more "else" then). Basically, I would expect us to be able to convert the newSource to a provider instance without an if.

pstorch: Valid point. I'll have a look.

PR #6952, opened 2018-03-07
fix #6943 reset usermodified coords

### LINKED ISSUE #6943: Local coord reset not working
- Open a saved cache
- Create a new waypoint and select to use the coords as cache coords locally
- Check that coords are now set as cache coords and another waypoint "Original coords" has been created
- Long click on "Original coords" wayoint and select to reset them as cache coords (locally)

- Cache details page still shows the previously set cache coords (also after closing and opening the cache)
- Refresh of the cache is needed to have the original coords shown again

- Coords should be reset as expected

##### Version of c:geo used:
2018.02.15-NB

Yes

pstorch:


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2271,7 +2271,7 @@ public void run() {
                     cache.setCoords(wpt.getCoords());
                     cache.setUserModifiedCoords(false);
                     cache.deleteWaypointForce(wpt);
-                    DataStore.saveChangedCache(cache);
```
Bananeweizen: I don't have an IDE here, therefore I can't check. Might it be that saveChangedCache and saveUserModifiedCoords are always called together? If so, I would move the new call into saveChangedCache instead of having a separate call afterwards.

pstorch: I'll check

pstorch: There is no `saveChangedCache` anymore, or do I overlook something?

Bananeweizen: Oh, my fault. Seems I did not really see the red color before. :)

PR #6953, opened 2018-03-07
test jenkins PR checkstyle check

pstorch:
Do not merge. I just want so see a failing Jenkins build in case of checkstyle violations.


pstorch:
test done

Bananeweizen:
If I remember right, I configured it only as warning, because it is hard to say what Jenkins takes as "reference build". For a PR you might expect that the commit before the PR is taken as reference, but I think it is just the last green build from the same Jenkins job.
If you want, we can increase the severity in the job configuration to see if that works better for us.

pstorch:
I changed it because Codacy seems to be unreliable, too, when it comes to checkstyle validations.
Last time I missed some checkstyle rules the pull request was green with both checks.

Issue #6812, opened 2017-10-30
Log not stored when only date changed


SammysHP:
- open cache
- menu, log cache
- change date
- press back

=> offline log is not stored.

2017.10.27-NB-ea62d13

Lineflyer:
AFAIR this is a side effect of a change done by @Bananeweizen a while ago to avoid storing offline logs when in fact nothing has been written into the log text field. 
IMHO a changed date or log type should also trigger saving the log offline,

pstorch:
I just had a look at it. We could remember the old date and compare it when going back. Then save if it is changed.
But what about the other things? LogType, Favorite Points, Attached Image, Log Password (in case of OC), Trackable Actions, ...
I know, we don't even save everything in an offline log, but somehow I would expect it. What do you think?

SammysHP:
Try to save as much as possible whenever something changes.

pstorch:
What was the reason to not change the log if the new logtext is empty? Preventing the user from accidentally deleting the logtext?  Does it not contradict to "save every change"?

Maybe a confirmation dialog would be better here (similar to the one proposed for waypoint changes). To make it explicit what the user wants.

pstorch:
This should be fixed by now. Can someone retest?

Lineflyer:
Tested and fixed.

Issue #5539, opened 2016-03-16
Implement 'needs maintenance' for Opencaching


following5:
All Opencaching sites now support 'needs maintenance' logs. Opencaching.de additionally has the feature of a 'needs maintenance' cache state (similar to the GC attribute 42) and of logging 'cache is ok' (does not need maintenance) by any user. All this is available through OKAPI, and it would be nice if it was added to c:geo.

First the basics. There are two different paradigms on 'needs maintenance' logs:
1. The GC way, which uses a dedicated log type "Needs maintenance" and is implemented also on OCPL sites (all OC sites except OC.de) and in c:geo.
2. The OC.de way, which uses a 'needs maintenance' = true / false / not-specified flag, which can be combined with any log type. This is also the way it is implemented in OKAPI, even when submitting logs to OCPL sites; OKAPI then will translate the logs to OCPL/GC logic.

I think the easiest way to implement submitting and displaying OC 'needs maintenance' in c:geo may be this:

**Submit 'needs maintenance'**

Enable log type 'Needs maintenance' for OC, and translate it to `logtype=Comment&needs_maintenance2=true` in the OKAPI request.

**Displaying 'needs maintenance' logs**
- Besides of the log type, also query the field `needs_maintenance2`.
- For logs which are not of type "Needs maintenance" (this is returned only by OCPL sites) and have `needs_maintenance2` = `true`, display an additional needs-maintenance-symbol besides of the log type symbol.
- For any log with `needs_maintenance2` = `false`, display a symbol for "cache is ok" or "cache does not need maintenance" besides of the log type symbol. This is the symbol displayed at OC.de: ![cache is ok](https://www.opencaching.de/resource2/ocstyle/images/log/16x16-needs-maintenance-no.png) This may be omitted for "Ready for search" logs, which imply that the cache is ok and always have `needs_maintenance2` = `false` at OC.de.

**Displaying the 'needs maintenance' cache state**
- Request the field `needs_maintenance`, and handle the value `true` like the GC attribute 42.

**Lose ends**

What is still missing now is submitting "cache is ok" (which is only supported by OC.de, other OC sites will ignore it and return the (localized) message: _Your cache log entry was posted successfully. However, your "does not need maintenance" flag was ignored, because Opencaching.XX does not yet support this feature._)

[...]

Issue #4899, opened 2015-04-20
Only last offline log for each cache is saved and exported when trying to save multiple logs of different types.


MrYaz:
If I first save an offline Found log and then edits and save an NM log it is only the last log for that particular cache that is saved and later on exported via export Field Notes from the History list. In this case only the NM log are uploaded to gc.com

This is only a problem when logging offline like I do most of the time..

If you work online, you first send the Found log, then the NM log. No problems.

You should be able to save offline and export several logs for each individual cache

rsudev:
As the UI very clearly tells does c:geo right now handle exactly one user-log per cache at a time. The recent integration of the offline log into the log view can pave the way for a change here, but it still requires some groundwork on database level and adjustments in the UI interaction to make this happen.

MrYaz:
I suspected that the current architecture does not support my suggestion as of today, but from a strictly users point of view it should be possible to post more then one log type in a row for a cache in offline mode as well.

rsudev:
Ah, now I see. Of course, if you do purely offline logging, nothing tells you that you can't stack several logs for a cache and you only get suspicious when you ask yourself where you would eliminate one of them...

SammysHP:
So let us see this issue as "make clear that only single offline log is possible" as we already have an issue about more than one offline log.

pstorch:
As we now have the possibility to log a cache + report a problem in one go, I think it is good enough for this feature request. Any opinions?

PR #6977, opened 2018-04-13
fix #6973 log "Report Problem" a.k.a NM, NA

### LINKED ISSUE #6973: (Re)implement "Report a problem"
Since the new logging function has been implemented (#6519) we have a regression;
The log types "Needs maintenance" and "Needs archived" is no longer available due to the inline function on the new logging page for these kinds of reports.

What we need is a dropdown menu in the logging activity with the following options:
Heading: Report a problem
Selections: Logbook is full / Container is damaged / Cache should be archived / Other

We should furthermore inform the user in a appropriate way, that he needs to describe the problem in his log, once he selected an option.
Here is the Groundspeak wording for that:
"Add details to your log to help the cache owner and other geocachers understand the problem."

pstorch:
Implements the "Report Problem" for the new GC logging page and also for OC (#5539).

It partly solves #4899 for those cases where a cacher wants to save e.g. a found and a needs maintenance offline or a DNF and a needs archived.

It also solves #6812 because I had to change the logic to save the log when the report problem changed.

### DIFF
--- a/main/res/layout/logcache_activity.xml
+++ b/main/res/layout/logcache_activity.xml
@@ -77,6 +77,14 @@
                     android:maxLines="1" />
             </LinearLayout>
 
+            <Button
+                android:id="@+id/report_problem"
+                style="@style/button_full"
+                android:layout_width="match_parent"
+                tools:text="report problem"
+                android:visibility="gone"
+                />
+
             <include layout="@layout/gcvote_rating_bar"  android:layout_width="wrap_content" android:layout_height="wrap_content"/>
             <CheckBox
                 android:id="@+id/favorite_check"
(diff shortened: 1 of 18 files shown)


pstorch:
I think the reported issue by Codacy regarding the `Use equals() to compare object references.` in the `ReportProblemType.isVisible()` is a false positive. You can compare two Enums with `==`. @Bananeweizen what do you think?

Lineflyer:
@Bananeweizen @Kumy If you have some spare time, it would be nice if you could review this PR (and the relevance of the Codacy issue).

Lineflyer:
@pstorch 
Seems the guys where you requested a review are not available at the moment.
How shall we proceed, I would of course like to test the change in our nightlies, but cannot judge the codacy issue.
Given codacy is right, would it be possible to use `equals()` here or else just put a comment in, that this needs to be investigated/modified later?!

pstorch:
@Lineflyer I checked with other Enums in the codebase and as far as I can see we use `==` to compare them. I've added a Unittest to check that it is working.
I'll merge it now. If someone has a different opinion, please raise your hand.

PR #6978, opened 2018-04-13
fix #6976 show tracking code for TBs

### LINKED ISSUE #6976: Public TB code shown instead of tracking code
Since implementation of the new logging page support the trackable list shows the public TB codes (TBxxxx) instead of the secrect codes.

It should show the secret code instead to have the old behaviour and be consistent with GK in the list.

Background:
We implemented to list the secret code years ago as this is the code printed on the TB.
So it helps when selecting the action while logging as you can identify the TB in your hand which you want to place/visit.

pstorch:


pstorch:
I think there is not much to review, so I've merged it.

PR #6985, opened 2018-04-24
Fix #6965 and #6966

### LINKED ISSUE #6965: Search by owner fails
- open GC48TE6
- click on owner name and search hidden caches

No result

show hidden caches of owner

##### Version of c:geo used:
2018.03.27-NB

Yes, seems to depend on owner name

##### System information:
```
```

### LINKED ISSUE #6966: Open profile or send message fails for certain owner
- Open GC42NEZ
- Click on user name, view profile or send message

404 / Error message

Opening related page

##### Version of c:geo used:
2018.03.27-NB

Yes for this owner

##### System information:
```
```

samueltardieu:


samueltardieu:
In the popup menu for the cacher, we must use the real user (in charge of the cache) instead of the display name chosen by the user (which can include, for example, "adopted from …" or "& some other cacher" or "& team").

PR #6989, opened 2018-04-24
6983 ui design for report problem

pstorch:


pstorch:
retest this please

pstorch:
Hmm, I tried to upload the new strings to Crowdin from master. It shows me that I've deleted 8 Strings. Maybe I've done something wrong :thinking:

pstorch:
Got it. My local HEAD was not updated correctly even I've pulled from upstream. Fixed it now. The new Strings are available on Crowdin.

Lineflyer:
Thanks. Took a look...Question now: The strings GC uses to generate the problem report log are encoded in the log page as far as I know. We have them now as translatable strings. 

Does this mean a "Problem report" (e.g. NM) will now be reported with the strings translated by our translators? While there is no functional problem with that, I ask myself if this would be considered undesireable by GC, because normally a user cannot change the string which the site foresees to be posted. Can you follow my thoughts?

The only alternative would IMHO be, to always keep the english strings...

pstorch:
Exactly, we have now a copy of the texts and can also translate them differently.
I don't know if this is a problem. Using only the English texts is also not a good solution, I think.
We could iterate through all available languages on GC and take exactly their texts and mark them as not translatable afterwards. But this doesn't sound like a good idea either.
On the other hand we could also try to find better texts or make them overwriteable by the users. 😉

Lineflyer:
I would like to avoid making the impression, that we manipulate the site content by replacing the strings predefined by the site.

That would mean iterate through the languages and store all the strings....

I suggest we wait for incoming translations and I volunteer to crosscheck against the predefined texts on Crowdin later on.
That should be enough.

pstorch:
Ok, just take into account that they might change the texts without prior notice.

PR #7005, opened 2018-05-03
fix #6997 report problem added to logbook

### LINKED ISSUE #6997: Posting NM/NA doea not add the corresponding entry to the logbook in c:geo
- Post a found log for a cache and select, that it has a problem
- Send the log after confirming the warning

- The found log is prepended to the logbook of this cache in c:geo
- The NM or Na log is not added but only visible after manual refresh of the cache details

- The NM/NA log should be added to the logbook same as the found log itself

##### Version of c:geo used:
2018.04.29-NB

Yes/No

##### System information:
```
```

pstorch:


pstorch:
@Lineflyer sorry for using your cache again for testing "Report Problem" :wink:

PR #7008, opened 2018-05-06
nonnull logImages in LogEntry.Builder

pstorch:
According to the comment here 
https://github.com/cgeo/cgeo/commit/f69bee474d5fa5631522851c96cd89d73ff3a170#commitcomment-28872323 I reworked the LogEntry.Builder handling of the NonNull logImages.

BTW this is "kind of" funny:
image

The assert is there to make the compiler happy, but my compiler (AndroidStudio) is not happy about it :wink:


main/src/cgeo/geocaching/log/LogEntry.java
```diff
@@ -80,9 +80,9 @@
         /** Friend's log entry */
         private boolean friend;
         /** report problem */
-        private ReportProblemType reportProblem;
+        @NonNull private ReportProblemType reportProblem;
```
UniQP: This looks unrelated to the commit message. Can you put it (and the setReportProblem change) into a separate commit?

pstorch: Sure

PR #7021, opened 2018-05-17
Crowdin

rsudev:
New translations
Added Greek


rsudev:
Somehow seem some translations gone lost on crowdin - need to cross-check first.

PR #7022, opened 2018-05-17
Crowdin

rsudev:
Now translations, added Greek


Lineflyer:
Just a quick remark: 
If I remember correct, we had a problem with the latest added language (turkish) that it was not used in the app although included which was solved later by some additions to configurations.
So might be this is also needed for Greek?!

PR #7046, opened 2018-05-31
fix #6993 backup menu item in main menu

### LINKED ISSUE #6993: have backup menu item in main menu
Several times I have experienced that people have lost their data, and they didn't know about the system preference to create a backup. We need to make that more visible. One possible way would be adding this to the main context menu.

pstorch:


Lineflyer:
Please better target ˋmasterˋ for this one?!

pstorch:
@Lineflyer yes, it's a new feature and should address `master`. It's corrected.

PR #7063, opened 2018-06-19
Transmit GC membership status for statistical purposes

samueltardieu:


samueltardieu:
Primarily for testing at this moment.

Lineflyer:
Fine for me. I only just can't do it via the GitHub UI as this can only merge to master.

UniQP:
@Lineflyer With the edit button on the top, you can also change the base from master to release.

Lineflyer:
Lets try that.

Lineflyer:
@samueltardieu Should I be able to see the new information on the status page already when testing with the new RC build, or does it require a certain amount of users to be displayed at all?

samueltardieu:
There was a typo in the URL parameter (`gc_mmembership` instead of `gc_membership`). I just fixed the name in the `release` branch, and allowed the typo temporarily on the server so that we can use the current RC right now.

Note that while the percentage will always be accurate, the number of users it is based on will be estimated during the first 30 minutes of the server restart (as are all the other numbers).

samueltardieu:
I'll rebuild a new nightly with this merge (and the typo fix). By the way, we might want to exclude development version from those statistics in the future, because the cgeo CI might connect numerous instance of premium and basic users while testing. Tell me if you want such a change by opening an issue on https://github.com/cgeo/cgeo-status.

Lineflyer:
@samueltardieu Thanks for fixing. It seems to work now.
I did not yet do anything with the RC version unless building and testing it myself. I will just trigger a rebuild, so we will have the typo corrected in beta.

PR #7084, opened 2018-07-15
Fix #5900, Waypoints hidden on map even below selected limit

### LINKED ISSUE #5900: Waypoints hidden on map even below selected limit
- Use "Beta Map" as a live map
- Waypoint limit is set to e.g. 20 caches (or even more)
- Open live map showing caches and its waypoints

While scrolling on the map it often happens that the waypoint suddenly vanish even if the selected limit of shown caches is not reached.
Its not clearly reproducible but seems to be related to the cache itself being out of view on the map.

I crosschecked with the "Old map" and its working without a problem on the same occurence (same location and caches/waypoints on map)
##### Version of c:geo used:

2016.08.06-RC1

Not clearly but happening often

rsudev:
- show waypoints in viewport as well

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/caches/AbstractCachesOverlay.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/caches/AbstractCachesOverlay.java
@@ -1,6 +1,7 @@
 package cgeo.geocaching.maps.mapsforge.v6.caches;
 
 import cgeo.geocaching.CgeoApplication;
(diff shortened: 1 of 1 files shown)


rsudev:
I'll give some time for comments, than a merge to test it on master. as its based on release, can be merged back to release as well.

Lineflyer:
The failing tests are caused by my cache GC3FJ5F, which is needed to disable last days.
Can't judge the code itself, but would like to test it in the nightlies.

rsudev:
The branch should be kept for possible merge back to release.

PR #7088, opened 2018-07-29
Fix #6987 Move "Repeat last log" next to "Clear"

### LINKED ISSUE #6987: Lost message log because of executing "Repeat last log" instead of "Save offline"
This is at least the third time that this happened to me:
I create a log in c:geo, but could not finish it. Therefore I wanted to save it offline.
But instead of hitting the "Save offline" menu item I accidentally pressed the neighbour menu item "Repeat last log" which replaced my log irreversibly.

The following options come into my mind:

- Reorder the menu items. Especially items that erase an existing log ("Repeat last log" and "Clear") should be grouped together and moved to the bottom
- Add a confirmation dialog or an Undo toast for these two actions "Repeat last log" and "Clear" if the log text area is not empty

Matze2:
... and away from "Save offline" to prevent accidental activation

### DIFF
--- a/main/res/menu/abstract_logging_activity.xml
+++ b/main/res/menu/abstract_logging_activity.xml
@@ -38,11 +38,6 @@
         app:showAsAction="ifRoom|withText">
     </item>
     <item
-        android:id="@+id/menu_repeat_last"
-        android:title="@string/log_repeat"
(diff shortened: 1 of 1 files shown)


Lineflyer:
retest this please

pstorch:
@Matze2 thanks :+1:

PR #7133, opened 2018-08-28
Fix #7130 - Change Twitter callback URL

### LINKED ISSUE #7130: Twitter login fail
- Settings
- Services
- Social media --> Twitter
-  - [x] Activate
- Authorize c:geo
- Start authorization

Short loading _"Waiting for Twitter"_ and then shows "toast" message _"c:geo failed to initialize authorization process"_

- [x] Yes

##### System information:
```
--- System information ---
Device: LG-D405 (w7n_global_com, lge)
Android version: 5.0.2
Android build: LRX22G.A1454843669
c:geo version: 2018.08.14
Google Play services: enabled - 12.8.74 (020304-204998136)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
System language: cs_CZ (c:geo forced to English)
System date format: M/d/y
Debug mode active: no
System internal c:geo dir: /data/data/cgeo.geocaching (274,4 MB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (274,4 MB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (274,4 MB free) external non-removable
Database: /data/data/cgeo.geocaching/databases/data (1,1 MB) on system internal storage
Geocaching sites enabled:
```

Lineflyer:
as the current prefix is not allowed to be configured at Twitter.

Please doublecheck if this is the only place, which needs changes and if we might run into problems using `http` instead of `callback`

### DIFF
--- a/main/src/cgeo/geocaching/twitter/TwitterAuthorizationActivity.java
+++ b/main/src/cgeo/geocaching/twitter/TwitterAuthorizationActivity.java
@@ -18,7 +18,7 @@ public class TwitterAuthorizationActivity extends OAuthAuthorizationActivity {
             true,
             Settings.getTwitterKeyConsumerPublic(),
             Settings.getTwitterKeyConsumerSecret(),
-            "callback://www.cgeo.org/twitter/");
+            "http://www.cgeo.org/twitter/");
 
     @Override
     protected String getCreateAccountUrl() {


Lineflyer:
retest this please

Lineflyer:
retest this please

SammysHP:
I don't think this will work because we need the callback scheme for the result handling. Otherwise the browser will open with the response, not c:geo.

rsudev:
That was my spontaneous thought as well, when reading this. Maybe we have the chance to register a specific pseudo-url that we can capture in c:geo as well.

PR #7144, opened 2018-09-25
Fixes #6947

### LINKED ISSUE #6947: Show whether TB has already been discovered
From support mail:
Geocaching.com now gives a status note on the top right of a trackable's page letting us know if we've discovered a TB before. It would be wonderful if that same status could be added when we pull TBs up in cgeo.  

I'd be great if in the same basic place you show "Found it" when we pull up a cache page showed "discovered it" on a trackable. We keep logging the same ones over and over between the 2 of us because we'd otherwise have to link to the web browser to see if we've logged it before or not, & that's a pain.

---
I did not yet see this feature on the webpage...seems to be pretty new.

bjdupuis:
Fixes #6947. Shows trackable status for the user, either discovered or retrieved. Tapping the status opens the corresponding log in the browser.

screen shot 2018-09-25 at 5 55 08 pm

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1167,7 +1167,7 @@
     <string name="trackable_geokrety">GeoKrety</string>
     <string name="trackable_geolutins">GeoLutins</string>
     <string name="trackable_status">Status</string>
-    <string name="trackable_found">Retrieved (%1$s)</string>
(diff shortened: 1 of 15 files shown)


SammysHP:
OK to test

bjdupuis:
Bah, meant to reopen my existing PR.

Issue #7148, opened 2018-09-29
Use diffent inputType for GC and Trackable search EditText


lazydays79:
It has always been a PITA to enter gc or trackable codes into the EditText of the search activity.

Just because you have to switch between characters and digits on your screen keyboard, and almost every GC or trackable code consists of both characters and digits.

If you switch the input type from textCapCharacters to textVisiblePassword, you have instantly both characters and digits on your screen keyboard. And gc.com is accepting lower case characters in their search, so nothing will be missed if you don't use cap characters.

textCapCharacters:

![screenshot_1538244934](https://user-images.githubusercontent.com/17769242/46249173-1af68e00-c425-11e8-89ca-540c99abce24.png)

textVisiblePassword:

![screenshot_1538244990](https://user-images.githubusercontent.com/17769242/46249172-1af68e00-c425-11e8-8571-68faec8288c1.png)

I will provide a pull request within the next hour.

Regards,
Michael

rsudev:
Done with 7aff8e5de155af144a76663158365d02df1fb794

PR #7149, opened 2018-09-29
Use textVisiblePassword instead of textCapCharacters

lazydays79:
This fixes issue #7148

### DIFF
--- a/main/res/layout/search_activity.xml
+++ b/main/res/layout/search_activity.xml
@@ -73,7 +73,7 @@
             style="@style/edittext_full"
             android:hint="@string/search_geo"
             android:imeOptions="actionGo"
-            android:inputType="textCapCharacters"
+            android:inputType="textVisiblePassword"
             android:text="GC"
             tools:ignore="HardcodedText" />
 
@@ -160,7 +160,7 @@
             style="@style/edittext_full"
             android:hint="@string/search_tb_hint"
             android:imeOptions="actionGo"
-            android:inputType="textCapCharacters" />
+            android:inputType="textVisiblePassword" />
 
         <Button
             android:id="@+id/display_trackable"
--- a/main/res/values/changelog_master.xml
+++ b/main/res/values/changelog_master.xml
@@ -7,7 +7,8 @@
     · Fix: Reorder menu items on logging screen to prevent accidental deletion of log\n
     · Fix: Waypoints on mapsforge map hidden already below configured limit\n
            Known issue:\n
-           Waypoints from other lists might leak onto the map....we are working on it\n  
+           Waypoints from other lists might leak onto the map....we are working on it\n
+    · New: Better keyboard layout while searching for geocaches or trackables on the search activity\n
     \n
   </string>
 </resources>


SammysHP:
ok to test

lazydays79:
I tried to check why cgeo.geocaching.address.GeocoderTest.testAndroidGeocoder fails (org.junit.ComparisonFailure: [country code for Android .geocoder] expected:<"FR"> but was:<null>), but locally it runs without any failure...

rsudev:
@lazydays79 Sorry for not jumping in earlier: the geocoder might fail, as it uses a service which has a request limit (we currently don't know any which is free and open source compatible).

lazydays79:
Ah, ok. I just was afraid that my PR could be rejected due to a failing build. 😅

rsudev:
I looked thorugh the change (easy enough :) ) and have only one question: Did you check whether the lower case input also works with other caching sites (like oc.de/pl/...) as well?
Looking at the Android dev docs associated with this, it looked like it might even be possible to combine these flags (to perhaps achieve capitalized input alongside numbers available). Did you consider this or have I misinterpret the docs? Might not be overly necessary if the lower case input works for all cache sites, but could still be helpful if users don't know that lower case works.

rsudev:
So, in the meanwhile checked both of my questions. Searching with lower-case works even for OC caches and adding textCapCharacters does not have an effect together with textVisiblePassword.
BTW, textVisiblePassword didn't have the desired effect (keyboard with letters and numbers) on my SDK 22 emulator instance...
Nevertheless a nice addition IMO, the remaining question is, whether this is not better suited on master (our general collection branch for new features and non-urgent bug-fixes).

lazydays79:
Hi!

No, I didn't check it with oc sites.

But combining the flags is a good idea worth trying. I will test that today
or tomorrow.

If it doesn't work, we could upcase the strings in the Text Fields before
sending the requests. Wdyt?

Regards,
Michael

rsudev <notifications@github.com> schrieb am Mo., 1. Okt. 2018, 09:08:

lazydays79:
Oh, didn't see your 2nd reply. 😂

Did you use the standard android keyboard in your emulator?

rsudev <notifications@github.com> schrieb am Mo., 1. Okt. 2018, 20:58:

rsudev:
I guess so (at least I didn't actively change anything in that area). It might be because in that keyboard the numbers are available (and depicted as such) on long press on the upper row of characters.
I personally would suggest (if you don't mind) to re-target the PR against master, then I'll merge it right away.

UniQP:
Maybe a changelog entry would also be nice.

lazydays79:
Sure, master sounds good. 👍

Thanks!


rsudev <notifications@github.com> schrieb am Mo., 1. Okt. 2018, 21:31:

lazydays79:
> I personally would suggest (if you don't mind) to re-target the PR against master, ...

Done

lazydays79:
> Maybe a changelog entry would also be nice.

Done

rsudev:
I know it's kind of tiring, but in order to have an understandable history - could you squash your changes into one commit please?

lazydays79:
I can't get lose at least the merge commit, but all the others have been squashed. If this is not satisfying, I suggest to cherry-pick 7aff8e5de155af144a76663158365d02df1fb794

rsudev:
As the new sqashed commit contained the complete change and was based on latest master, I just pushed it.
So all is done, thanks for your contribution!

PR #7159, opened 2018-10-17
Fix condition to prevent out-of-bounds access

UniQP:


UniQP:
This was found by inspection.

PR #7164, opened 2018-10-22
Fix #7161: Allow country code to be null

### LINKED ISSUE #7161: Failing tests on CI
Currently, the Android Geocoder test always fails on CI, e.g. http://ci.cgeo.org/job/cgeo%20pull%20request/1444/testReport/cgeo.geocaching.address/GeocoderTest/testAndroidGeocoder/

In a [comment](https://github.com/cgeo/cgeo/pull/7149#issuecomment-425712144) @rsudev stated that this is due to a request limit but I don't think that's the case. I think we receive the requested address data but it just does not contain the country code anymore.

So we should check whether the country code is missing in general and either remove the country code check or use other test data that contains a country code.

UniQP:
Let's see whether the CI likes it


Lineflyer:
retest this please

UniQP:
Not necessary anymore.

PR #7178, opened 2018-11-03
clarify "friend" javadoc

okainov:
Current "friend" field and its description was very confusing. I'm not sure whether I should rename it to something else, but at least add some javadoc explaining its purpose.


rsudev:
ok to test

Lineflyer:
restest this please

Issue #7115, opened 2018-08-15
Problem authenticating to OC.de


Lineflyer:
##### Detailed steps causing the problem:
- Open your default web browser and manually login to opencaching.de, select to keep logged in
- Go to Menu - Settings - Services - Opencaching.de
- Start authorization process
- Browser is only opened for a split second (which is normal, as the browser login is still valid)
- c:geo reports successful outcome "c:geo is now authorized..."

##### Actual behavior after performing these steps:
When returning to the main menu, it shows OC.DE - Not authorized

##### Expected behavior after performing these steps:
User should be authorized to OC.DE

##### Version of c:geo used:
2018.08.14 and 2018.08.14-NB1

##### Is the problem reproducible for you?
Yes

The only workaround is to manually log out on your default browser and only then start the authentication procedure in c:geo. This will open the web page, let you log in and authorization is working normal afterwards

@following5 @mambofive
Any server side changes recently, which could cause this?

following5:
mambofive and me don't have access to OC.de servers. Server admins are @teiling88 and @bohrsty.

Lineflyer:
For some reason it looks like we send the token twice, which leads to 401 for the second try of course as it was already used:
```
08-15 10:25:59.933 28611 28611 I cgeo    : [main] Requesting settings backup with settings manager
08-15 10:26:01.356 28611  1463 D cgeo    : [OkHttp] GET https://www.opencaching.de/okapi/services/oauth/request_token?oauth_callback=callback%3A%2F%2Fwww.cgeo.org%2Fopencaching.de%2F&oauth_consumer_key=REMOVED3&oauth_nonce=REMOVED1&oauth_signature_method=HMAC-SHA1&oauth_timestamp=1534321561&oauth_token=&oauth_version=1.0&oauth_signature=REMOVED2
08-15 10:26:01.815 28611  1463 D cgeo    : [OkHttp] 200 (459 ms) GET https://www.opencaching.de/okapi/services/oauth/request_token?oauth_callback=callback%3A%2F%2Fwww.cgeo.org%2Fopencaching.de%2F&oauth_consumer_key=REMOVED3&oauth_nonce=REMOVED1&oauth_signature_method=HMAC-SHA1&oauth_timestamp=1534321561&oauth_token=&oauth_version=1.0&oauth_signature=REMOVED2 (http/1.1)
[...]

PR #7185, opened 2018-11-11
OAuth: fix double exchanging tokens

okainov:
Sometimes exchangeTokens() can be called twice, so it will lead to
failure in exchanging during the second try. Let's remember the last
verifier and if it was already used, skip exchange step

Fixes: #7115

### DIFF
--- a/main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
+++ b/main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
@@ -64,6 +64,8 @@ public abstract class OAuthAuthorizationActivity extends AbstractActivity {
     private ProgressDialog requestTokenDialog = null;
     private ProgressDialog changeTokensDialog = null;
 
+    private String lastVerifier = null;
+
     private final Handler requestTokenHandler = new RequestTokenHandler(this);
     private final Handler changeTokensHandler = new ChangeTokensHandler(this);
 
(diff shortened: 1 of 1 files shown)


Lineflyer:
ok to test

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
```diff
@@ -64,6 +64,8 @@
     private ProgressDialog requestTokenDialog = null;
     private ProgressDialog changeTokensDialog = null;
 
+    private String lastVerifier = "";
```
UniQP:
```suggestion
    private String lastVerifier = null;
```
Otherwise we behave different if the first verifier is `""`.

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
```diff
@@ -181,12 +183,15 @@ public void onResume() {
...
-                changeTokensHandler.sendEmptyMessage(NOT_AUTHENTICATED);
+            if (!verifier.equals(lastVerifier)) {
+                if (StringUtils.isNotBlank(verifier)) {
+                    exchangeTokens(verifier);
+                } else {
+                    // We can shortcut the whole verification process if we do not have a token at all.
+                    changeTokensHandler.sendEmptyMessage(NOT_AUTHENTICATED);
+                }
             }
+            lastVerifier = verifier;
```
UniQP: Can you move this line into the `if` block?

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
okainov: I think in general I got different behavior here. Original code was returning an error in case of empty verifier. SO probably we want to keep it(?). Thus, I changed the code to maintain the same behavior, so it doesn't matter actually what is the initial value here. But changed it to null, don't see any issue here.

main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
okainov: restructured to simplify the whole structure

okainov:
Any update here?

PR #7189, opened 2018-11-11
Fix #4697: TB selections and GC vote lost when returning

### LINKED ISSUE #4697: TB selections and GC vote lost when returning to logging activity
2015.02.21-NB
- Have some TBs in your inventory and GCVote activated in settings
- Start logging activity for a cache
- Write log text
- Select some TB actios
- Enter a GCVote rating
- Open details of a TB by clicking on it OR attach a log image

Once you return from this actions back to the logging activity, the screen still shows GCVOte rating and TB actions as you set it but this is only due to the screen not being updated. Scroll or click somwhere and the rating and the TB-actions are lost.

bekuno:
This Fixes #4697: TB selections and GC vote lost when returning to logging activity

After returning from TB detail view - do not reload the log data when already TB data exist.
This is a little bit dirty, but I think we call this only in this situation.

Extra: deactivate generating the activity in onCreate, ist seems not really useful

### DIFF
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -223,7 +223,7 @@ public class LogCacheActivity extends AbstractLoggingActivity implements DateDia
                     trackable.action = LogTypeTrackable.VISITED;
                 }
             } else { // refresh view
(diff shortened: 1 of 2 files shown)


bekuno:
sorry, this was the false branch

PR #7190, opened 2018-11-11
Add missing `final` modifiers to interfaces

okainov:


UniQP:
`final` modifiers on interface parameters have no effect. Do you see an advantage in adding them?

Currently some interfaces use `final`modifiers, other don't use them. We should at least make this consistent.

okainov:
>final modifiers on interface parameters have no effect. Do you see an advantage in adding them?

Modern IDEs allow you to automatically create all methods when you declare your class as implementing an interface. Thus, to auto-generate methods interface signature is being used. Without `final` in interface you get methods without `final` in your class and then you get style checker complains.

Lineflyer:
ok to test

okainov:
>I tested Android Studio, which explicitly remove final modifiers.

Not really true, more true is "AS does not respect final modifiers in interfaces by default". It can be easily improved by following settings:

image

However its behavior doesn't depend on presence of `final` in interface, correct. It was my initial intention though, I thought that it's smart enough here...

UniQP:
I did a quick check and there are more missing `final` modifiers in interfaces. Is there a checkstyle configuration (or similar), so we get a warning if we miss a `final` modifier? That's especially important to keep this consistent in the future.

Lineflyer:
@okainov @UniQP 
Can you give me a hint, whether this PR is still ongoing or obsolete after your discussion?

UniQP:
It's not obsolete.

okainov:
@UniQP tried to see Checkstyle inspections and found out that `final` in interfaces falls under `RedundantModifierCheck` which is currently ignored. I can put it to `info` and remove all finals from interfaces, so it will be consistent and will be enforced by current jenkins style check for the next commits. Does it look fine?

okainov:
If #7231 looks good & will be merged, then this can be closed as obsolete.

PR #7201, opened 2018-11-27
Do not merge! Test-Pr for style changes only

rsudev:
This PR is only for getting a build of the style changes alone.
When testing (on Android 6 and above) you need to manually check the permissions in the Android App settings, as no requests for them are made.
But the styles and action bars (dark and light) should look OK and no crashes on google map.

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="PROJECT_RELATIVE:$PRJ_DIR$/checkstyle.xml:cgeo" />
+        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
         <entry key="checkstyle-version" value="8.8" />
         <entry key="copy-libs" value="true" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="PROJECT_RELATIVE:$PRJ_DIR$/checkstyle.xml:cgeo" />
+        <entry key="location-2" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
         <entry key="scan-before-checkin" value="true" />
         <entry key="scanscope" value="JavaOnly" />
         <entry key="suppress-errors" value="false" />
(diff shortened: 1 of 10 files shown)


Lineflyer:
@rsudev Will test the build result APK.

Lineflyer:
I installed the debug build on my Android 8 tablet for testing.
Here is what I found:

- In general all main functions are now working again (e.g. live map). Did however not test all possible activities/screens.
- Google map tiles are not loading (Is this a limitation of the debug build?)
- Some menus have a bit strange style (e.g. big font size, text alignment not the same, etc.) but nothing serious, could be optimized later.
- Settings menu is missing title bar on each screen and subscreen, which is confusing and requires you to use the device back button as only way up
- Three-dot menu is appearing on topmost right covering partly overlapping the action bar and the three dot button itself. This is no functional drawback, but I used to close the menu by tapping the three-dot button once again, which is no longer possible now. Instead you can dismiss it by tapping outside of the menu.

Summary:
IMHO good to merge and further work on it. Of course the permission requesting needs to be added to make it a usable version.

@rsudev: I think I will only get more clues when using it in the nightlies in real life. Once this is merged I will generate dedicated isues for the remaining glitches.

rsudev:
The settings screens should now be as before. The non-loading google map tiles are indeed a limitation of the debug build.
Regarding merging to master for the nightlies: If you install this freshly, I assume you will get a direct crash of the app on trying to start already, unless you manually assign the location permission in the Android app settings. As long as this is OK in your opinion, we can merge it (probably after a smoke test of the PR)

Lineflyer:
> The settings screens should now be as before. 

Will test it.

> The non-loading google map tiles are indeed a limitation of the debug build.

Fine, then no problem.

> Regarding merging to master for the nightlies: If you install this freshly, I assume you will get a direct crash of the app on trying to start already, unless you manually assign the location permission in the Android app settings. As long as this is OK in your opinion, we can merge it (probably after a smoke test of the PR)

Well, I know how to handle and for updating nightly users there wont be a problem. But I am wondering whether you cannot just "mix in" (sorry for this term) the implementation of PR #7162 at the same time?

Lineflyer:
> > The settings screens should now be as before.

> Will test it.

Looks perfect now.

rsudev:
The 'mix-in' is the next step which I hopefully can pursue further the next days. There are some resulting peculiarities I rather want to get a grip on by stepwise applying it on top than trying to figure out after the fact.

Lineflyer:
For me merging into nightlies would be ok. People using nightlies have to expect problems and it helps us finding more problems.

rsudev:
Supersed by #7213

PR #7202, opened 2018-11-27
Fix #7200, Map reloading when popup opens

### LINKED ISSUE #7200: Map reloading when popup opens
- Map a list of caches on mapsforge offline map (new libraries)
- Click on a cache to open the popup
- Close the popup
- Repeat the popup opening/closing

After some tries (sometimes already on the first try) the map tiles are reloading when trying to open a popup. In fact it looks like the whole activity is reset/restartet as also the title bars vanish for a moment.

This behavior seems to be caused by a recent change on `master` as it was working normal with a build from approx. 2 weeks ago.

No reload/restart

##### Version of c:geo used:
2018.11.25-NB

Yes, see above

##### System information:
Samsung S7, Android 8

@rsudev Might be related to the new theme selector as this change came in recently?

rsudev:
Do not restart on all preference changes,
only on theme-setting related ones


Lineflyer:
That looks like a reasonable fix and explains #7200

PR #7203, opened 2018-11-28
Fix cgeo#7192: README.md unclear to newbie about API keys

bekuno:
Update README.md to use private.properties for generating and filling out keys.xml.
Additional some minor changes in README.md.
Add rows for the keys in the template file for private.properties.


rsudev:
ok to test

PR #7210, opened 2018-12-10
Update Okapi waypoint types

following5:
poi = point of interest = intesting place; this is not a trailhead. Best mapping is "Reference point".

For trailheads, there is the new OC type "trailhead", e.g. in [OP6FEC](https://opencaching.pl/viewcache.php?cacheid=28657).


rsudev:
OK to test

PR #7212, opened 2018-12-13
Fix #7211 - PMO failure not detected

### LINKED ISSUE #7211: Wrong error message when BM tries to load PM-only cache
- Login as basic member of geocaching.com
- Try opening a PM-only cache, e.g. GC7Z0DM or GC792QY

c:geo shows "Download of cache details failed. Unknown error".

c:geo should show, that this cache is restricted to PM users

##### Version of c:geo used:
2018.10.13 as well as 2018.12.10-NB

Yes, seems to be a regression after site change?!

Lineflyer:
Seems the page shown to basic members when accessing a PMO cache has changed recently.
The used class item seems to be specific for this page, as I could not find it on other cache pages.

Change is not tested yet. Will test it with debug APK from CI.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -165,7 +165,7 @@ public final class GCConstants {
     static final Pattern PATTERN_SESSIONTOKEN = Pattern.compile("sessionToken:'([^']+)'");
 
     static final String STRING_PREMIUMONLY_2 = "Sorry, the owner of this listing has made it viewable to Premium Members only.";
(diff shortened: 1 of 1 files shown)


Lineflyer:
I tested the change using http://ci.cgeo.org/job/cgeo%20pull%20request/1541/artifact/main/build/outputs/apk/basic/debug/cgeo-debug.apk and the issue is solved. Normal caches, found caches, disabled caches are parsed normal.

Looking forward for review by @UniQP or @rsudev

rsudev:
I agree, this pattern should be very specific.

rsudev:
I'll see to merging it to master tomorrow, if that is OK.

PR #7213, opened 2018-12-16
Targetsdk 26 - first step

rsudev:
Seems to work ok now (the memory leak was unrelated, but is now fixed).

Review comments already welcome, but I want to merge manually in any case.

Feature wise missing is the check for storage permission (we fallback to app storage, so to works, but backups will be deleted on uninstall).

Contains the build and theming changes as well.

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="PROJECT_RELATIVE:$PRJ_DIR$/checkstyle.xml:cgeo" />
+        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
         <entry key="checkstyle-version" value="8.8" />
         <entry key="copy-libs" value="true" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="PROJECT_RELATIVE:$PRJ_DIR$/checkstyle.xml:cgeo" />
+        <entry key="location-2" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
         <entry key="scan-before-checkin" value="true" />
         <entry key="scanscope" value="JavaOnly" />
         <entry key="suppress-errors" value="false" />
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -14,7 +14,7 @@ android {
 
     defaultConfig {
         minSdkVersion 14
-        targetSdkVersion 22
+        targetSdkVersion 27
     }
 
     sourceSets {
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,6 +9,3 @@ org.gradle.jvmargs = -Xmx2048m
 
 # Only configure other modules when they are actually needed for the task being executed
 org.gradle.configureondemand=true
-
-# avoid "resource is private" errors. https://stackoverflow.com/questions/45865366/cardslib-error-resource-androidattr-foregroundinsidepadding-is-private-while-ta
-android.enableAapt2=false
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -20,6 +20,8 @@ if (isContinuousIntegrationServer()) {
 android {
     compileSdkVersion "Google Inc.:Google APIs:23"
 
+    buildToolsVersion "27.0.3"
+
     compileOptions {
         // use the diamond operator and some other goodies in Android
         sourceCompatibility JavaVersion.VERSION_1_7
@@ -28,7 +30,7 @@ android {
 
     defaultConfig {
         minSdkVersion 14
-        targetSdkVersion 22
+        targetSdkVersion 27
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)
 
@@ -334,10 +336,10 @@ dependencies {
     implementation "io.reactivex.rxjava2:rxandroid:2.0.2"
 
     // Support Library. Appcompat
-    implementation 'com.android.support:appcompat-v7:20.0.0'
+    implementation 'com.android.support:appcompat-v7:23.4.0'
 
     // Support Library RecyclerView
-    implementation 'com.android.support:recyclerview-v7:22.0.0'
+    implementation 'com.android.support:recyclerview-v7:23.4.0'
     // we don't want to implement decorators on our own
     implementation 'com.yqritc:recyclerview-flexibledivider:1.4.0'
 
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -20,7 +20,7 @@ if (isContinuousIntegrationServer()) {
 android {
     compileSdkVersion "Google Inc.:Google APIs:23"
 
-    buildToolsVersion "27.0.3"
+    buildToolsVersion "28.0.3"
 
     compileOptions {
         // use the diamond operator and some other goodies in Android
--- a/main/res/layout/actionbar_popup.xml
+++ b/main/res/layout/actionbar_popup.xml
@@ -36,7 +36,7 @@
         android:id="@+id/overflowActionBar"
         style="@style/action_bar_action"
         android:longClickable="true"
-        android:src="@drawable/abc_ic_menu_moreoverflow_normal_holo_dark" />
+        android:src="@android:drawable/ic_menu_more" />
 
 
 </LinearLayout>
--- a/main/res/layout/livemapinfo.xml
+++ b/main/res/layout/livemapinfo.xml
@@ -18,7 +18,6 @@
             android:layout_width="fill_parent"
             android:layout_height="wrap_content"
             android:text="@string/live_map_notification"
-            android:textColor="?text_color"
             android:textSize="14sp" />
     </LinearLayout>
 
--- a/main/res/layout/pocketquery_item.xml
+++ b/main/res/layout/pocketquery_item.xml
@@ -29,7 +29,7 @@
             android:id="@+id/cachelist"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:drawableLeft="@drawable/abc_ic_search"
+            android:drawableLeft="@android:drawable/ic_menu_search"
             android:gravity="left|center_vertical"
             android:text="@null"/>
     </LinearLayout>
--- a/main/res/menu/main_activity_options.xml
+++ b/main/res/menu/main_activity_options.xml
@@ -5,7 +5,7 @@
     <!-- TODO: use ic_action_search -->
     <item
         android:id="@+id/menu_gosearch"
-        android:icon="@drawable/abc_ic_search"
+        android:icon="@android:drawable/ic_menu_search"
         android:title="@string/search_bar_hint"
         app:actionViewClass="android.support.v7.widget.SearchView"
         app:iconifiedByDefault="true"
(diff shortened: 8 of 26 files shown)


rsudev:
On issue I meanwhile found is, that the app crashes if it is not running in the background and you jump directly e.g. to the cache details via a link. In that case the main activity is bypassed and therefore the GPS not checked and initialized.

Lineflyer:
> Feature wise missing is the check for storage permission (we fallback to app storage, so to works, but backups will be deleted on uninstall).

Well....for backup its only a glitch (as long as backup/restore functions are still available within c:geo), but I guess functions like GPX-export and field notes might be meaningless if they are saved on a location not accessible with file browsers (which I assume is the case then?!).

What about offline map file access without the permission?

rsudev:
I added permission check and location initialization to the other entry points and added a first draft of requesting storage permissions.
Need to restructure the source a bit, but perhaps could the PR-build already get a short check?

ControlTheBit:
@rsudev thank you for your work here! 👍

Lineflyer:
Yes, thanks very much @rsudev.
Cant wait to test it, let me know when its ready for a first test.

rsudev:
I think the output of the current PR is already test-worthy. Before actually merging it to master I need to give it a second look.

Lineflyer:
ok, will do next year :)

bekuno:
I tested the current apk from the PR build with my Samsung S6 Test-Smartphone. (Android 7.0).
The app starts at top splash with the geocaching.com credential configuration page.
After the back key (nothing set) I see the system splash asking for storage access. In the background lies a c:geo dialog. He is active after closing the storage access question (regardless of the answer).
When I hit "ask again" then the app crashes.

I think that the credential page should not be shown until the permission questions have been answered.
Without the own permission dialog the initialization would work too, I think.
If you want I can make some snapshots to document this.

Lineflyer:
My test result is comparable to the experiences of @bekuno :

- Install
- First start
- Enter GC.com login credentials (or simply exit to main screen, does not make a difference)
- Device asks for storage permission
- After granting this permission a c:geo dialog appears stating that location permission failed.
- Selecting ask again makes c:geo crash

There seems something wrong with the sequence of permission requests.
Maybe it would be cleaner to create a new activity/screen which handles the first start (or in general on startup if it is detected, that a permission is not granted (e.g. removed after granting it))  including permission requests before anything else happens.

bekuno:
I advocate to remove the automatic jump to the settings.
The preselection of geocaching.com is for me also not necessary.

Lineflyer:
> The problem is, that there are several avenues into the application (normal start, cache url, list shortcut, ...) 

That was the reason for my idea sketched above:
Isn't there a possibility to "detour" every entry point into the app into a "subroutine" checking the permissions and in case at least one of them is not granted, handle this with the appropriate dialogues, etc. in this "subroutine" and then continue to the original "route".
Not sure however if there is a easy way for that in Android programming (or maybe even a specialized supportive library available for that)...

rsudev:
Another attempt. If possible, please check.

rsudev:
Just cleaned up the checkstyle issues, no functional changes.

bekuno:
I did a short test. (Android 7)
In my opinion, the initial authorization dialog is very good.

After the app start I see the credential dialog. Should we better present a message that the user should now type his credential to one of the geocaching platforms and the jump to the main services setting page (with all platforms)? Maybe this is worth to be another issue. (@okainov : what do you mean from perspectice of geocaching.su?)

Then the app wants to load the database backup. This question was 3 times. Could this be easily fixes or should this also be another issue?

And, I was surprised that the map shows caches without typing login data to geocaching.com (only activated in the settings).

rsudev:
The database restore is another topic, good catch. BTW, if you did restore the db, no wonder that caches were shown on the map...
Regarding the initial guidance: Currently the gc connector is activated by default. The user will see GC: no credentials stored (or similar) on the main screen. tapping on that should open the services page, which should be enhanced to show the authorization status besides the activation information, to lead the user towards the right settings.

Lineflyer:
Will try to test it next days.

@bekuno The topics of multiple restore dialog, better first credential input (tutorial) and caches without login on map are all known issues already listed here somewhere.

okainov:
Just tried the build out. Some random notes:

- Filesystem permission on startup. May be ok as first version to get it working, but in general seems very suspicious - why? You should ask for filesystem access when saving caches \ making backup \ logging in (maybe). But not during startup
- GPS permission on startup. Needs to be asked only for Live map and Nearby search, not always
- Context menus on Live map look overlapping the main title bar, also nice arrow disappeared

- Button "more details" on CachePopup is now super small, it had ~3 times bigger height

- "Cancel" font is huge

- Personal note test is aligned vertically and has weird blue line

- Crash when clicking on nick of any user in the list of logs (as well as clicking on cache owner nick)
- Click on images on "Images" tab does nothing (Neither does long tap "open as file"). I expect selector "Which app to use to open the image"
- I find "back" arrow on all screens very ugly. Before it was much less visible gray icon.
- When rolling back to my old dev version, I got `    java.lang.RuntimeException: Unable to create application cgeo.geocaching.CgeoApplication: java.lang.NullPointerException: Attempt to invoke virtual method 'android.database.sqlite.SQLiteStatement android.database.sqlite.SQLiteDatabase.compileStatement(java.lang.String)' on a null object reference
` error in AS. Solved by clearing all app data and uninstalling. However could be a good point for "release notes" if someone will try to downgrade

--- System information ---
Device: SM-A320FL (a3y17ltexc, samsung)
Android version: 8.0.0
Android build: R16NW.A320FLXXU3CRH2
c:geo version: 2019.01.08-c5b7140 developer build
Google Play services: enabled - 14.7.99 (040306-223214910)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
System language: en_AU
System date format: d/M/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (1.1 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (1.1 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (1.1 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (268.0 KB) on system internal storage
Geocaching sites enabled:
Geocaching.com date format: d. M. yyyy

Lineflyer:
Today I found some spare time to test the latest build result:

The permission checks are now consistent and OK, both for location and storage.
In a little bit more detail:

- First start after install:
  - Permissions are requested and if one of them is not granted the foreseen dialog appears
  - The "ask again" and "close app" selectors are working as intended for both permissions.

- Trying another "avenues" into the app after manually revoking the permissions:
  - Opening a cache link from mail to get to the cache details:
       - Permission request for location appears with correct dialogues, permission request for storage only appears in case you can go back to the main level in this scenario. Furthermore if the location permission request is denied, there is no dialog "ask again" / "close app".
- There seem to be some other combination(s) of ways into the app and requests granted and/or denied, which lead to crashes. Need to investigate further to find reproducible problem cases.

However:
Basic permission handling is fine, only special cases and regressions caused by the changed target SDK are open. IMHO this is now mergable and go the next test step (nightly builds). This allows getting more practical experience and a slightly broader user base. 

I would then like to transfer all issues listed here into dedicated issues and try to prioritize them.

Lineflyer:
@rsudev 
If this PR still reprents your current code base for this change, feel free to merge from my point of view.

okainov:
> feel free to merge from my point of view

Apart of my other comments which may be improved later (however IMO some of the changes are pretty ugly), this one I'd consider as an important one, app shouldn't crash anyway:

>Crash when clicking on nick of any user in the list of logs (as well as clicking on cache owner nick)

Lineflyer:
> Crash when clicking on nick of any user in the list of logs (as well as clicking on cache owner nick

I have overseen this in your report. Thanks for highlighting again.
Such kind of failures are actually the reason I would like to broaden the test base by merging it. I am pretty sure, there are even more high prio issues still present as side effect....IMHO we just need an extension of testing now.

moving-bits:
@rsudev @Lineflyer How can I (and maybe other "junior c:geo devs") help to bring the sdk26 changes to market? Can we support by doing testing on different devices (emulated or physical), by trying to fix certain issues (which ones in which prio?) or by something else? I'm willing to support, but I am currently sort of lost what would help most right now :-)
(I do have a couple of ideas what I would like to do next feature-wise, but do not want to drain energy or distract focus from getting the changes needed for sdk26 done for now.)

rsudev:
I currently see three problem areas:

* the v3 maps crash
* the inconsistent and partly not working permissions request logic
* the style/ui issues

Currently I would say that I might be able to fix the first near-time, as I have a good idea where it comes from and how to fix.
For the second, there is some conceptual work required to get it straight. I have some ideas I could contribute currently to a discussion, but time doesn't permit to lead it right now. I think we can run without location permission (just ensure to keep location access off and return some 'dummy'). Regarding storage, having a potentially useless backup and weird issues importing/exporting gpx I see more problematic.
For the third, I'd suggest to first do a systematic 'inventory' across the version span (like Api 17, 22, 23, 26) with screenshots of the known issues before and after the change as a reference. Some of the issue might need to be addressed on a more global scale (we currently e.g. use compile-sdk 23. while it might be needed for gmaps v1 it can lead to side-effects in the styling I guess).

Feedback welcome!

moving-bits:
Regarding the permissions request logic I had the following (partial) concept in mind:

IMHO the most important part is the first start of c:geo after installation. There we should ask for the rights needed and stay in main screen until the ones absolutely needed are given. Maybe we could set an internal flag for it, something like "first time rights check successful". You cannot go further than main screen until this flag is set.

Then, for all parts in the application where we need rights, we only need to check if permission is granted. If yes, we can succeed. If no, we just print an error message and return. IMHO we don't need to have to implement an a asynchronous check like "ask for granting a certain right / wait until right granted / continue with function requested".

This would be ok IMHO, since
- on first app start there is only one entry point, and we do request all rights, and we cannot continue until the right absolutely necessary are granted
- later on, you can only run into a rights problem, if you have deliberately chosen to reject a right already given - so you know what you have done and how to undo this (ok, maybe we can print a hint for this - but we don't have to implement additional logic for this)
- for the rights not absolutely needed (like GPS) we can implement (or already have) a dummy logic

Would this approach make things easier to implement?

moving-bits:
As for the "before-after inventory": I will fire up an emulated device with a couple of different API versions and start making screenshots for the 725x/726x UI issues for sdk26.

My approach would be
- to install the latest c:geo release version,
- make a screenshot for each issue of this row,
- with each API you've mentioned in your post
- and attach it to the respective issue.

Then I would do this again for the current master and, if applicable, for current master + fix provided (if not yet merged to master).

Lineflyer:
My few cents:

I tried to give all sdk26 related issue a corresponding label, so it should be easy to see what are the known regressions.

I also tried to assign prio label according to what I think would be the severity for our users focussed on usability.

IMHO the permission thing is realeaseable as is and can also be optimized later, while however the design problems are more relevant as they potentially affect all users.

moving-bits:
> As for the "before-after inventory": I will fire up an emulated device with a couple of different API versions and start making screenshots for the 725x/726x UI issues for sdk26.

Sadly, I'm out of this game for now, as I cannot get the release version of c:geo to run on any emulated device 👎. Error messages (if any) are different from API version to API version, and sometimes the emulated device simply freezes. I've tried several different approaches to get rid of this (like cold boot vm, wipe data, clean & rebuild all etc.), but nothing worked for the release version.
Running the current master still works, though.

moving-bits:
I started making screenshots for reference and layout comparision, see https://www.moving-bits.net/cgeo/targetsdk26/index.php

So far, I've created screenshots for release version with API 23 and 26 - more to come...

PR #7223, opened 2019-01-02
allow jpg, png, gif for log foto uploads (fix #6955)

### LINKED ISSUE #6955: PNG file not selectable for photo uploads
While trying to upload a log photo for OC (same at GC?) I was not able to select a *.png file. It was greyed out in the file selector.

No idea whether this is a phone problem, whether its supported on OKAPI, but uploading this PNG via browser on the OC website worked fine afterwards.

moving-bits:
instead of a fixed MIME filter of "image/jpeg" set intent filter to "image/*" and check the file chosen by the user for MIME types "image/jpeg", "image/png" and "image/gif" (which are uploadeable for both geocaching.com as well as opencaching.* platforms)

### DIFF
--- a/main/src/cgeo/geocaching/ImageSelectActivity.java
+++ b/main/src/cgeo/geocaching/ImageSelectActivity.java
@@ -239,7 +239,7 @@ public class ImageSelectActivity extends AbstractActionBarActivity {
 
     private void selectImageFromStorage() {
         final Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
-        intent.setType("image/jpeg");
+        intent.setType("image/*");
 
         startActivityForResult(Intent.createChooser(intent, "Select Image"), SELECT_STORED_IMAGE);
     }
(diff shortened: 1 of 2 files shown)


Lineflyer:
ok to test

Lineflyer:
@moving-bits 
Codacy still has one issue open for the PR. Can you check and mitigate?

moving-bits:
Is there anything I can do about it?

Lineflyer:
Probably not. I tried to retrigger it on Codacy, but it still uses the outdated commit. However your PR is fine. If no one objects I will merge for testing tomorrow.

UniQP:
I will take a look at the code.

main/src/cgeo/geocaching/ImageSelectActivity.java
```diff
@@ -344,6 +350,10 @@ private void showFailure() {
         showToast(getString(R.string.err_acquire_image_failed));
     }
 
+    private void showFailureUnsupportedImageFormat() {
```
UniQP: I don't think it's worth introducing a new method here (especially since the method is only used once).

Lineflyer:
Thanks @moving-bits for your contribution. Will test as time permits. And thanks @UniQP for review.

PR #7224, opened 2019-01-03
raise wp limit to 200 (fix #6810)

### LINKED ISSUE #6810: Allow higher limit of waypoints shown on map
In the "Settings" menue, sub menue "Map", parameter "Show Waypoints": Actual limit is 50.
When using a large high resolution screen, e.g. on a 10 inch tablet, a higher limit would be needed. Suggestion: limit of 200, pre setting after installation should be 40.

Difficult to get a good overview about an area and a possible walking path if a number of final location have already been entered: If the zoom level fits for the area, the finals are no longer shown.

Just a better ovrview.

##### Version of c:geo used:
2017.10.23

Yes, it's not a bug, it's a nice feature

##### System information:
```
```

moving-bits:
raised max value for waypoints to 200 / default unchanged

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -72,7 +72,7 @@ public class Settings {
     private static final char SEPARATOR_CHAR = ',';
 
     private static final int SHOW_WP_THRESHOLD_DEFAULT = 10;
-    public static final int SHOW_WP_THRESHOLD_MAX = 50;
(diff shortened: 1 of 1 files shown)


Lineflyer:
ok to test

Lineflyer:
@rsudev Do you recall, if there was any technical reason for the upper limit of 50 WPs?

moving-bits:
did a rebase with merge conflict resolution

moving-bits:
rebased to today's (Feb 23) master

PR #7225, opened 2019-01-03
three colum mainscreen with up to two lines of text each (fix #6361)

### LINKED ISSUE #6361: Overhaul of existing home screen
While there are some existing issues to redesign the complete app UI or major parts of it (and still we have no one around capable of doing this) might I suggest to at least to improve the existing UI a little.

Most devices now have a high resolution screen but our main menu is still optimized for low resolution it seems.

I mean:
When looking at the main menu I do see a lot of empty space and six items in the middle.
We could at least place the icons more distinct from each other and modify the text boxes to show the complete text (right now it shows for me e.g. "Gespeich..." where there is enough space for the complete title).

Here is my screen with almost all items descriptions shortened:
screenshot_20170226-222303

moving-bits:
tweaked the layout settings for the mainscreen so that
- icons + text are set into three equally-wide columns
- text can be up to two lines long (but takes up space for only one line if short enough)

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -21,18 +21,19 @@
     <!-- ** -->
 
     <LinearLayout
-        android:layout_width="fill_parent"
+        android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_centerInParent="true"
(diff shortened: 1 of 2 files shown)


Lineflyer:
ok to test

PR #7227, opened 2019-01-04
 Add user option to change max distance up to which brouter is used (fix  #6157)

### LINKED ISSUE #6157: Enable brouter for distance >10km
I tried the new BRouter feature in the latest Beta and noticed it is only working for a short distance navigation. When I try it for caches further than 8-10 km away I get a map showing the destination cache but without guiding lines.

There seems not to be a fixed reproducible distance, I guess it depends on route complexity. I have no Idea if this is a known BRouter limitation or a bug in c:geo implementation but I read that BRouter can provide "long distance cycle routes" so I guess it's the second.

A last general feedback I like to give for BRouter: I love the possibility to route to a cache inside c:geo but I also noticed that BRouter itself don't feel very user friendly in the moment. It uses a unusual UI and users have to read different readme files that can be found inside the brouter folder on the phone storage to get it work like it should.

##### Version of c:geo used:
2016.11.21-RC

Yes

##### System information:
```
--- System information ---
Device: BV6000 (Blackview, Blackview)
Android version: 6.0
Android build: BV6000_Blackview_V01_20161011
Cgeo version: 2016.11.21-RC
Google Play services: enabled
Low power mode: active
Compass capabilities: yes
Orientation sensor: present
Hide own/found: true
Map strategy: auto
System language: de_DE
Log date format: 23.11.16
Debug mode active: no
Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
Geocaching.com date format: d.M.yyyy
Installed cgeo plugins: calendar, contacts
```

moving-bits:
Add a user option to settings/map configuration for max. distance (in km) up to which brouter gets involved.
Default value = 10 / max. value = 120 / 0 = disables route calculation

### DIFF
--- a/main/src/cgeo/geocaching/maps/routing/Routing.java
+++ b/main/src/cgeo/geocaching/maps/routing/Routing.java
@@ -21,7 +21,6 @@ import org.xml.sax.helpers.DefaultHandler;
 
 public final class Routing {
     private static final double UPDATE_MIN_DISTANCE_KILOMETERS = 0.005;
-    private static final double MAX_ROUTING_DISTANCE_KILOMETERS = 10.0;
     private static final double MIN_ROUTING_DISTANCE_KILOMETERS = 0.04;
     private static final int UPDATE_MIN_DELAY_SECONDS = 3;
     private static BRouterServiceConnection brouter;
@@ -86,7 +85,9 @@ public final class Routing {
      */
     @NonNull
     public static Geopoint[] getTrack(final Geopoint start, final Geopoint destination) {
-        if (brouter == null || Settings.getRoutingMode() == RoutingMode.STRAIGHT) {
+        final int maxThresholdKm = Settings.getBrouterThreshold();
+
+        if (brouter == null || Settings.getRoutingMode() == RoutingMode.STRAIGHT || maxThresholdKm == 0) {
             return defaultTrack(start, destination);
         }
 
@@ -98,7 +99,7 @@ public final class Routing {
 
         // Disable routing for huge distances
         final float targetDistance = start.distanceTo(destination);
-        if (targetDistance > MAX_ROUTING_DISTANCE_KILOMETERS) {
+        if (targetDistance > maxThresholdKm) {
             return defaultTrack(start, destination);
         }
 
--- a/main/res/layout/brouter_threshold_preference.xml
+++ b/main/res/layout/brouter_threshold_preference.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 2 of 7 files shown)


moving-bits:
Calculated conversion between metric and imperial units may work, but how can we adjust the text according to c:geo configuration? All strings are configured via XML files (as well as the layout elements are).

Lineflyer:
Ok to test

bekuno:
I hope this could be the right way: 
https://stackoverflow.com/questions/10145190/android-how-to-change-texts-in-the-preferences-activity-dynamically

main/src/cgeo/geocaching/maps/routing/Routing.java
```diff
@@ -21,7 +21,7 @@
 
 public final class Routing {
     private static final double UPDATE_MIN_DISTANCE_KILOMETERS = 0.005;
-    private static final double MAX_ROUTING_DISTANCE_KILOMETERS = 10.0;
+    // private static final double MAX_ROUTING_DISTANCE_KILOMETERS = 10.0;
```
rsudev: Don't forget to remove commented lines finally.

moving-bits:
@bekuno & @rsudev: I will look into both, thanks for the hints.

moving-bits:
Found an even more elegant way:
- text above seekbar is now neutral
- legend for seekbar converts the unit and displays "mi" or "km" according to settings

moving-bits:
Codacy complains about empty but undocumented method body. It's necessary because of abstract method, so I need to add a comment.
Does this comment have to be above the method signature `public void onStartTrackingTouch(final SeekBar seekBar)` or within the method body?

kumy:
Why not raise in this case some sort of "unimplemented" exception?

moving-bits:
> Why not raise in this case some sort of "unimplemented" exception?

The method gets called, actually, so I cannot (should not) raise an exception (it's a "start touch" handler).
I needed to implement this method, even if it's empty, as I'm instantiating a class which has some abstract methods. - So I just need to add a comment about this, which is what Codacy asks for. I only wanted to know where I have to place this comment (within or before this method) - I think I will simply try it, it's a 50:50 chance :-)

Lineflyer:
@rsudev Do you approve this PR?

rsudev:
I did not have the oppoertunity to try it out, but codewise it looks ok.
Needs to be squashed and rebased before merging.

moving-bits:
Ok, I've tried to rebase and squash by using the following commands:

`git checkout master`
`git pull --rebase upstream/master`
`git checkout issue_6157`
`git rebase upstream/master`
`git rebase -i HEAD~4` - leaving the first commit as `pick`, switching the others to `squash` 
`git push --force`

Lineflyer:
As @rsudev approved codewise and the PR has been squashed and rebased meanwhile, I merge this to test it in the nightly builds

PR #7228, opened 2019-01-04
 Add Icon to address search result list to directly jump to map (fix #6400)

### LINKED ISSUE #6400: Center Map to search coords after search by coords/address (Basic member)
- Step 1
Search any address where you don't have saved caches.
- Step 2
Click Map Icon
- ...

Shows last location, not the location you searched

It would be nice that it could center the map on the location we have chosen. This way we could perform live search without having to save list of caches first.

##### Version of c:geo used:
2017.02.28-RC

Yes

moving-bits:
Extends mainscreen's "search for address" functionality by adding a map icon to the result list. Clicking on this map icon directly leads to a map view, centered at the coordinates given by the clicked search result.

### DIFF
--- a/main/src/cgeo/geocaching/address/AddressListActivity.java
+++ b/main/src/cgeo/geocaching/address/AddressListActivity.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.Intents;
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.AbstractActionBarActivity;
 import cgeo.geocaching.location.Geopoint;
+import cgeo.geocaching.maps.DefaultMap;
 import cgeo.geocaching.ui.recyclerview.RecyclerViewProvider;
 import cgeo.geocaching.utils.AndroidRxUtils;
 
@@ -66,4 +67,9 @@ public class AddressListActivity extends AbstractActionBarActivity implements Ad
         finish();
     }
 
+    @Override
+    public void onClickMapIcon(final Address address) {
+        DefaultMap.startActivityGeoCode(this, new Geopoint(address.getLatitude(), address.getLongitude()));
+        finish();
+    }
 }
--- a/main/res/layout/addresslist_item.xml
+++ b/main/res/layout/addresslist_item.xml
@@ -1,42 +1,58 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/button_map"
-    android:layout_width="fill_parent"
+    android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:layout_margin="3dip"
     android:background="?background_color"
-    android:orientation="vertical"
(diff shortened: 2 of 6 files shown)


Lineflyer:
Ok to test

main/src/cgeo/geocaching/address/AddressListActivity.java
```diff
@@ -66,4 +67,9 @@ public void onClickAddress(final Address address) {
         finish();
     }
 
+    @Override
+    public void onClickMapIcon(final Address address) {
+        DefaultMap.startActivityCoords(this, new Geopoint(address.getLatitude(), address.getLongitude()), null, null);
```
rsudev: This method is tailored to display an arbitrary point on a map, not really what your intention is.
As there is currently no method available to achieve what you want, you need to create a new one (which likely includes the creation of a new MapOptions contructor).

moving-bits: Created new intent, maptoptions etc. to start with a live map around the coords. (see commit 42a0377)

Seems to work fine generally, BUT after loading caches (live & stored) they get shown and instantly vanish again ?!

Lineflyer: > Seems to work fine generally, BUT after loading caches (live & stored) they get shown and instantly vanish again ?!

That sounds like one of the regressions we have on the new map.

moving-bits: @rsudev 
There seems to be some racing condition, because if I step through it step by step using breakpoints in
- `DefaultMap#43:startActivityGeoCode(...)` (to dive into) and
- `StoredCachesOverlay#66:if(moved) ...` (to see status) 

it works fine (meaning: loaded caches get displayed and do not vanish instantly [*]).
Sadly I do not know enough about Java and concurrent programming to dive further into this...

[*] There is a follow up regression: If I click on a cache, a popup gets shown (ok). If I close this popup by using the back button, the cache vanishes from map view. This is independent from whether or not I save this cache in between.

moving-bits: > @moving-bits Which kind of map did you use when you saw the mentioned problems (Google, OSM)?

@Lineflyer I can only test with OSM since my Google Maps keys are not working with c:geo.

Lineflyer: Are you sure those problems were not already present without your changes here?

moving-bits: I cannot say for sure - I had the problem with loosing the display of loaded caches already before, frequently, but not always. On the other hand, with my change I could not get the emulator to make the loaded caches stay displayed at all :-( 
(Have not tested the APK on a physical device yet, though, don't have a spare one.)

moving-bits: Ok, I found a physical device to test this on. Works like a charm with the new map.

moving-bits:
rebased on today's (Feb 23) master and squashed

Lineflyer:
@rsudev Your change request is outdated, right? So feel free to merge, if you think its ok now.

Lineflyer:
My question was clarified with @moving-bits , therfore merging now.

PR #7231, opened 2019-01-09
Improve code style checks

okainov:
Following up on #7190 as well as improving a bit beyond that

- Add redundant modifiers check, thus removing all `final` from interfaces
- Add imports order check

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -126,6 +126,13 @@
     <module name="WhitespaceAround">
       <property name="severity" value="info" />
     </module>
+    <module name="ImportOrder">
+      <property name="groups" value="cgeo,android,java,javax,*"/>
+      <property name="ordered" value="true"/>
+      <property name="separated" value="true"/>
+      <property name="option" value="top"/>
+      <property name="sortStaticImportsAlphabetically" value="true"/>
+    </module>
   </module>
 </module>
 
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -94,7 +94,7 @@
       <property name="severity" value="info" />
     </module>
     <module name="RedundantModifier">
-      <property name="severity" value="ignore" />
+      <property name="severity" value="info" />
     </module>
     <module name="RightCurly">
       <property name="severity" value="info" />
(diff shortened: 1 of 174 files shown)


Lineflyer:
Add to whitelist

checkstyle.xml
```diff
@@ -126,6 +126,13 @@
     <module name="WhitespaceAround">
       <property name="severity" value="info" />
     </module>
+    <module name="ImportOrder">
+      <property name="groups" value="cgeo,android,java,javax,*"/>
```
rsudev: When browsing through the changes, i would consider the additional groups butterknife, io.reactivex and org.apache.commons, to highlight the usage of those more uncommon libraries better.
What do you think?

rsudev:
Very nice :)

checkstyle.xml
okainov: I was just following [wiki rules](https://github.com/cgeo/cgeo/wiki/Coding-Conventions) which I got pointed to many times during recent PRs =)

If you want to add more groups, I'm fine with that, however keep in mind that this probably will cause a bit more code style failures in PRs of beginners, especially if they haven't set up checkstyle plugin in AS (I just commented separately in the other issue) But anyway update the wiki first please. Also it would probably worth mentioning there that from now on we enforce alphabetic sorting of imports inside groups.

rsudev: Thanks for highlighting this, I obviously was no longer aware of this page. If we agrree on an extension, I will care for updating the wiki accordingly.
So, back to the question: Would it be sensible/worthwhile to have these new groups from your perspective or does it not add value?

okainov: Disclaimer: I'm not relly a Java developer :)

Honestly I don't really think that extra groups will bring some value. For me it's pure automated\scripted thing, I work in IDE and I don't write any imports manually. It's definitely good to have the order defined to be deterministic. But either it 4 or 5 groups - I will press Ctrl+Alt+O anyway and it will be put in the correct order automatically.

Lineflyer:
retest this please

Lineflyer:
@rsudev Approved?

okainov:
I rebased it several times already, seemed to be fine, but was not merged apparently. Feel free to rebase and resolve conflicts by yourself when you're ready to merge.

rsudev:
Somehow did the online resolution of conflicts not work. So i rebase and resolved conflicts locally and pushed to master.
If everything looks fine on master from your perspective, please close here.

okainov:
@rsudev you could also push directly to my branch / this PR. But anyway, fine for me.

PR #7232, opened 2019-01-09
added check for actual username when logging a cache (fix #7197)

### LINKED ISSUE #7197: Usage of mail address instead of username not handled in c:geo
- Since some time is seems possible to use your mail address associated with geocaching.com instead of the username to login to the platform
- Correspondingly also with c:geo login works when using the mail address as username

While the app works normally, c:geo takes the mail address used to login as the username.
This means, that e.g. after posting logs the mail address appears in the local cache logbook instead of the usernmae. 
Only after cache refresh the correct data (i.e. username) is shown.

If the user uses a valid mail address for login, c:geo should parse the corresponding user name and use it instead.
Alternatively we should either not allow login by mail address or in this case additionally let the user enter his username.

##### Version of c:geo used:
2018.10.13

Yes

moving-bits:
added check for actual username when logging a cache (fix #7197)

### DIFF
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -42,9 +42,11 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.TrackableActivity;
 import cgeo.geocaching.command.AbstractCommand;
 import cgeo.geocaching.connector.ConnectorFactory;
+import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.ILoggingManager;
 import cgeo.geocaching.connector.ImageResult;
 import cgeo.geocaching.connector.LogResult;
+import cgeo.geocaching.connector.capability.ILogin;
 import cgeo.geocaching.connector.trackable.TrackableConnector;
 import cgeo.geocaching.connector.trackable.TrackableLoggingManager;
 import cgeo.geocaching.enumerations.LoadFlags;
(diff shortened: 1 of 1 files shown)


kumy:
Jenkins: add to whitelist

Lineflyer:
@kumy Good idea. Thank you!

rsudev:
Looks good. As I am not sure whether rebaseing on github will do the right thing, can you do that on your side?

moving-bits:
> Looks good. As I am not sure whether rebaseing on github will do the right thing, can you do that on your side?

Never done that before - could someone give me some advice on this, pls?

So far I had
- merged cgeo:master into my local master some does ago (`git pull upstream master` IIRC),
- created a branch for this issue on top of that,
- committed locally and pushed the changes to my fork on github
- and created a PR from that.

What would be the next steps?

okainov:
>Never done that before - could someone give me some advice on this, pls?

As a baseline, let's consider you have repo with both `upstream` and `origin` remotes. 

And you want to rebase `origin/issue_7197` branch on top of `upstream/master`.
Straightforward commands would be:

```
git fetch upstream #Make sure everything is up-to-date
git checkout issue_7197
git rebase upstream/master #May want to add --autostash option
```
However, due to some merge commits that you got (in the future you should not do `git pull upstream master`, but maybe `git pull --rebase upstream master` or even `git fetch` and then explicitly rebasing manually) it may result in painful resolving (or may not). In this case it could be easier to do it a clean way:


```
git fetch upstream #Make sure everything is up-to-date
git checkout -b issue_7197_new upstream/master #Create new clean branch
git cherry-pick d873b96 #Include only the commit you need
git push origin issue_7197_new:issue_7197 --force # Forcibly push new branch to the old branch in your fork
```

moving-bits:
@okainov 
Thanks for the detailed explanation. I'll write that down for reference, I'll probably need that frequently...

I went with the second approach as the first one gave me an error message when I tried to push this into the repo. Let's hope that everything is in a state that the team will be able to merge those changes.

rsudev:
Thanks, that was perfect. Thank you for your conribution!

PR #7242, opened 2019-01-18
restrict brouter threshold setting to a minimum of one km (fix #7239)

### LINKED ISSUE #7239: Two ways of disabling brouter not consistent
- Using the new setting for the max distance for routing it is also possible to disable it (Distance=0km)
- There is another switch for disabling the routing via Map - Menu - Routing - "Direct line"

Those two possibilities currently do not work hand in hand.
This leads to the problem, that if the user selected "Direct line" he will never see a routing although the diance is set >0. Vice versa if the disance is set to 0 and the user selects routing e.g. "Bike" no routing will happen.

Either 0 should not be valid as distance setting (no deactivation via this setting) or we should move the routing method selection from the map menu into the settings.

##### Version of c:geo used:
2019.01.16-NB

Yes

@moving-bits

moving-bits:
- restrict brouter threshold setting slider to a minimum of one km
- removed the check for zero value in routing routine
- remove mention of "0=disable" in settings string

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -583,7 +583,7 @@
     <string name="init_showwaypoints">Show Waypoints</string>
     <string name="init_showwaypoint_description">If less than the given amount of caches are displayed on the map, their waypoints are shown additionally.</string>
     <string name="init_brouterThreshold">Calculate route</string>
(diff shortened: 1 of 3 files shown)


Lineflyer:
Retest this please

Lineflyer:
Thanks @moving-bits

Issue #7235, opened 2019-01-13
Map trail disappears when viewing cache details


mjarmstr:
##### Detailed steps causing the problem:
- Open the map view with trail enabled.
- Walk a while so the trail shows up.
- Tap a cache and open the cache details.
- Exit back to the map.

##### Actual behavior after performing these steps:
- When exiting back to the map the trail is gone. 

##### Expected behavior after performing these steps:
- When exiting back to the map the trail remains.

##### Version of c:geo used:
2018.10.13

##### Is the problem reproducible for you?
Yes

##### System information:
```
--- System information ---
Device: moto x4 (payton_fi, motorola)
Android version: 9
Android build: PPW29.69-26
c:geo version: 2018.10.13
Google Play services: enabled - 14.7.99 (100408-223214910)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: absent
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: true
Map strategy: auto
HW acceleration: enabled (default state)
System language: en (c:geo forced to English)
System date format: M/d/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (36.7 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (36.6 GB free) external non-removable
Geocache data: /storage/0000-0000/Android/data/cgeo.geocaching/files/GeocacheData (55.0 GB free) external removable
Database: /storage/emulated/0/Android/data/cgeo.geocaching/files/databases/data (12.5 MB) on user storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / BASIC
Geocaching.com date format: MM/dd/yyyy
Installed c:geo plugins:  none
--- End of system information --
```

##### Other comments and remarks:
In general the impermanence of the trail is disturbing to me.  It is very useful when navigating from cache to cache using the map and when orienting oneself as you approach a cache.

SammysHP:
Probably duplicate of #2958 (which should be re-opened if we change our opinion of the trail feature).

I assume that the map activity is destroyed due to memory limitations. The trail is not stored permanently, so it will not be loaded again. A short term solution would be to place the trail in the `Bundle` and restore it on resume. A more complex solution would be to make the trail persistent as requested in #2958, but that would require some management functions as well (and should be discussed in #2958).

[...]

PR #7244, opened 2019-01-20
make trail history persistent (fix #2958)

### LINKED ISSUE #2958: Have trail on live map more permanent
From FB:
When one exits the map and enter the cachedescription for logging and then enters the map afterwards the trace is gone.

Request: Track on map log. Would like a track log that shows the track even if one have closed the map and opened it once again. Just like a track log on a real GPS.

Reasson: Have been some places where mapcoverage isn't good and would like the possibilities to "backtrace" ones footsteps.
## 

Also when leaving the live map by using the back key the trail is lost.
It would be could to keep it running/saved as long as c:geo is active with a menu option on the map menu to clear it.

moving-bits:
- log trail history to database
- restore when recreating map (implemented for CGeoMap and NewMap)
- menu option to clear history (implemented for CGeoMap and NewMap)
- limit number of trail points in database automatically (cleanup on open db)

### DIFF
--- a/main/src/cgeo/geocaching/maps/PositionHistory.java
+++ b/main/src/cgeo/geocaching/maps/PositionHistory.java
@@ -1,8 +1,13 @@
 package cgeo.geocaching.maps;
 
+import cgeo.geocaching.storage.DataStore;
+import cgeo.geocaching.utils.AndroidRxUtils;
+
 import android.location.Location;
 
 import java.util.ArrayList;
+
+import io.reactivex.schedulers.Schedulers;
 
 /**
  * Map trail history
@@ -21,6 +26,42 @@ public class PositionHistory {
 
     private ArrayList<Location> history = new ArrayList<>();
 
+    // load data from permanent storage
+    public PositionHistory() {
+        history = DataStore.loadTrailHistory();
+    }
+
+    // save current coords to permanent storage
+    private void saveToStorage(final Location coordinates) {
+        AndroidRxUtils.andThenOnUi(Schedulers.io(), new Runnable() {
+            @Override
+            public void run() {
+                DataStore.saveTrailpoint(coordinates);
+            }
+        }, new Runnable() {
+            @Override
+            public void run() {
+                // nothing to do UI-wise
+            }
+        });
+    }
+
+    // clear position history (in memory and on permanent storage)
+    public void reset() {
+        AndroidRxUtils.andThenOnUi(Schedulers.io(), new Runnable() {
+            @Override
+            public void run() {
+                DataStore.clearTrailHistory();
+                history.clear();
+            }
+        }, new Runnable() {
+            @Override
+            public void run() {
+                // nothing to do UI-wise
+            }
+        });
+    }
+
     /**
      * Adds the current position to the trail history to be able to show the trail on the map.
      */
@@ -32,6 +73,7 @@ public class PositionHistory {
             return;
         }
         if (history.isEmpty()) {
+            saveToStorage(coordinates);
             history.add(coordinates);
             return;
         }
@@ -41,6 +83,7 @@ public class PositionHistory {
             return;
         }
 
+        saveToStorage(coordinates);
         history.add(coordinates);
 
         // avoid running out of memory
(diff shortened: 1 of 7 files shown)


moving-bits:
Could someone have a look at my function that writes the data to the db in the background, please? `saveToStorage()` in PositionHistory.java#34 - It seems to work, but I'm not sure if it's the right way to do. Thanks.

rsudev:
As I just came across the bug that triggered this implementation, I took a short first look.
I would think it should be more efficient to save all track points to the database when we leave the map, not on every move, shouldn't it?
As the PositionHistory is self-initializing now, the Activity-state persisting should then no longer take it into account (but could trigger the save to db).

moving-bits:
ok, will try to adapt my PR after you've checked in the bug fix for #7235

If I understood it correctly, I would:
- change the place where saving gets triggered + save the whole trail (instead of just a single point) up to a certain maximum length
- therefore I do not need to strip the saved trail on DB startup anymore, since max length is limited on save
- keep the DB structure change (new table) as is in the PR
- probably have to change the place where the trail gets loaded (initializing of PositionHistory instead of Activity init)

kumy:
Testing jenkins - related to #7295 

Jenkins: retest this please

kumy:
Jenkins: retest this please

Lineflyer:
retest this please

Lineflyer:
I suggest to postpone merging this new feature until we have "stable" base for the next release and have branched to `release` branch.

moving-bits:
Yes - this PR was not supposed to be merged in a situation like today. I think kumy used it just for testing purposes today.

Lineflyer:
Setting "Do not merge" for the time being to have better overview.

moving-bits:
Hm, this wasn't supposed to be merged right now. Need to talk to @rsudev first to get a grip on how to implement the changes he suggested. Readiness to merge won't be until next week, probably.

Lineflyer:
I did not merge it, just removed the label. But I added the label again and changed the title to make sure no one merges yet :)

moving-bits:
Maybe it was the CI "waiting for input" and taking the removal of the "Do not merge" label as call to action :)
Thanks for re-placing it.

Lineflyer:
> Maybe it was the CI "waiting for input" and taking the removal of the "Do not merge" label as call to action :)

No, it was as I pushed a commit to your branch for this PR (Typo fix).

moving-bits:
> No, it was as I pushed a commit to your branch for this PR (Typo fix).

ah, I see

moving-bits:
> As I just came across the bug that triggered this implementation, I took a short first look.
> I would think it should be more efficient to save all track points to the database when we leave the map, not on every move, shouldn't it?
> As the PositionHistory is self-initializing now, the Activity-state persisting should then no longer take it into account (but could trigger the save to db).

@rsudev: Could you give me a hint or two on:
- where to place the saving of the trailhistory to db after your changes?
- whether loading of the trailhistory from db can stay where it is now?
- whether the way I initiate the db write in background is technically ok?

Thanks!

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -21,6 +25,26 @@
 
     private ArrayList<Location> history = new ArrayList<>();
 
+    // load data from permanent storage
+    public PositionHistory() {
```
rsudev: PositionHistory is only used on CGeoMap if I am not mistaken, so the persistence will only work there, or did I miss something?

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -41,6 +66,7 @@ public void rememberTrailPosition(final Location coordinates) {
             return;
         }
 
+        saveToStorage(coordinates);
```
rsudev: Essentially I would prefer to save the history only when the map activity is stopped, but when I looked through the implementation, it seemed not too easy to get that triggered in a safe fashion, so we might leave it like it is implemented here for the moment.

Lineflyer:
@moving-bits Can you double confirm you consider this change ready to merge?`We just copied `master` to `release`so its the best time for new feature merging onto `master`.

moving-bits:
rebased, retested, squashed and pushed => feel free to merge this one

PR #7268, opened 2019-01-26
adjust fontsizes in cachelist spinner item (fix #7259)

### LINKED ISSUE #7259: List name font size too big on top bar
- Open a list

The font size of the list name is much bigger than it used to be before the targetSDK change and it is also not the same font size as used in the list change dropdown, which opens once the list name is clicked.
Huge font size also limits the characters shown, thus lists names like "Nice caches Düsseldorf" can no longer be diferentiated from "Nice caches Cologne" as only "Nice cache..." is shown on the top bar.

Font sizes should be aligned and smaller

##### Version of c:geo used:
2019.01.23-NB

Yes

##### System information:
Android 8.0

moving-bits:
adjust fontsizes in cachelist spinner item (fix #7259)

### DIFF
--- a/main/res/layout/cachelist_spinneritem.xml
+++ b/main/res/layout/cachelist_spinneritem.xml
@@ -15,6 +15,7 @@
         android:ellipsize="end"
         android:layout_width="match_parent"
         style="?attr/titleTextStyle"
+        android:textSize="16sp"
         android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


Lineflyer:
@moving-bits Thanks. I will look through your PRs and merge all, which I can judge to be easy changes.

PR #7273, opened 2019-01-26
restore actionbar icons on main activity and cache details pages (fix #7256)

### LINKED ISSUE #7256: Icons missing in top bar of several screens
- See actual behaviour

- c:geo main screen top bar does no longer show c:geo icon on top left
- Cache details screen does no longer show cache type icon on top left

Icons should be shown

##### Version of c:geo used:
2019.01.23-NB

Yes

##### System information:
Android 8.0

##### Other remarks:
Considering as `high` priority as in the cache detail view this is somehow influencing the usability as the cache type was previously visible "at first look". Now users have to scan the details list.

moving-bits:
restore actionbar icons on main activity and cache details pages (fix #7256)

### DIFF
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -54,6 +54,7 @@ import android.os.Handler;
 import android.os.Message;
 import android.support.annotation.NonNull;
 import android.support.v4.view.MenuItemCompat;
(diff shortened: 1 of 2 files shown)


moving-bits:
updated naming of local vars, rebased to current master and squashed commits into one

rsudev:
Thanks @moving-bits for the implementation and @okainov for taking a look!

PR #7282, opened 2019-02-02
adjust cachedetails trackable list layout (fix #7281)

### LINKED ISSUE #7281: Icons and text in cache inventory out of alignment
Similar to the problem on "Useful apps" also icons and text of the cache inventory tab are out of alignment.

2019.02.01-NB

Android 8.0

I am on mobile phone only, therefore I had to keep it short.

moving-bits:
fix layout on cachedetail's trackable list page (fix #7281)


moving-bits:
added screenshots for visual reference: https://www.moving-bits.net/cgeo/targetsdk26/index.php

PR #7284, opened 2019-02-05
Change height of "More details" button on cache popup to fixed height (fixes #7252)

### LINKED ISSUE #7252: More details button on cache popup too small
- Open map
- Click on cache to open popup

The "More details" button got smaller compared to before the targetsdk raise

It should be at least the height as before the targetsdk change (i.e. same height as the disk icon below)

##### Version of c:geo used:
2019.01.23-NB

Yes

##### System information:
Android 8.0

Image from @okainov

bekuno:
Change height of "More details" button on cache popup to fixed height 40dip - the same height as the buttons below (offline refresh, ...)

### DIFF
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -31,7 +31,7 @@
                 android:id="@+id/more_details"
                 style="@style/button_small"
                 android:layout_width="fill_parent"
-                android:layout_height="wrap_content"
+                android:layout_height="40dip"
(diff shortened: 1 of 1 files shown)


bekuno:
The code formatting for the TextView was done automatic by the AS.

moving-bits:
added screenshots for visual reference: https://www.moving-bits.net/cgeo/targetsdk26/index.php

PR #7326, opened 2019-02-24
fix #7311 Coord input dialog broken for DMS format

### LINKED ISSUE #7311: Coord input dialog broken for DMS format
- Open "search" menu
- Click on coordinate fields
- Change format to DMS (DDD° MM'SS.SSS'')

The N/E prefixes are not shown, once you click on them the window is resized and shown normal.
Besides that the whole input got smaller than before.

Window should be shown complete initially. 
Size of window and all elements could be a bit higher.

##### Version of c:geo used:
2019.02.17-RC

Yes

##### System information:
Android 8.0

Reported on support mal

bekuno:
fix #7311
set N/S E/W Buttons to fixed height and width

### DIFF
--- a/main/res/layout/coordinates_input.xml
+++ b/main/res/layout/coordinates_input.xml
@@ -24,7 +24,7 @@
 
             <Button
                 android:id="@+id/ButtonLat"
-                style="@style/button_full" />
+                style="@style/button_letter" />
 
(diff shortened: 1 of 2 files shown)


Lineflyer:
Thanks @bekuno

PR #7335, opened 2019-02-25
Add three dot image for cache popup (Fix #7251)

### LINKED ISSUE #7251: Three dot icon replaced by new icon on cache popup
- Open map
- Click on a cache on the map to open the popup

The popup "action bar" now contains a new icon instead of the three-dot icon

Actually the new icon is pretty same style as the other icons.
So instead of havin the three-dot back here, I could also imagine to keep it but then maybe make it consistent in other activities.

##### Version of c:geo used:
2019.01.23-NB

Yes

##### System information:
Android 8.0

Picture from @okainov

bekuno:
Fix #7251
Use also xml file generated in AS


bekuno:
@Lineflyer Now the fresh build apk file from jenkins was fine for me.
Can you test it?
I'm not sure why that did not work without this vector xml.

bekuno:
Rebased, squashed and tested the debug.apk with android 7.
Now the image is ok.
@moving-bits can you test this with some other API Versions with the emulator(s)?

moving-bits:
@bekuno will do

But how can I fetch your branch to my local repo for testing? `git checkout --track bekuno/Three_dot_icon_replaced_by_new_icon_on_cache_popup` did not work (although I had created "bekuno" as alias for remote repo https://github.com/bekuno/cgeo before).

SammysHP:
If you've added bekuno as a remote you can

```
git fetch bekuno Three_dot_icon_replaced_by_new_icon_on_cache_popup:bekuno/Three_dot_icon_replaced_by_new_icon_on_cache_popup
git checkout bekuno/Three_dot_icon_replaced_by_new_icon_on_cache_popup
```

If you have not added the remote yet, you can also directly fetch from GitHub:

```
git fetch https://github.com/bekuno/... Three_dot_icon_replaced_by_new_icon_on_cache_popup:...
```

moving-bits:
@SammysHP Thx for the explanation, will try when back at the dev machine

moving-bits:
I've tested with different API versions on the same (emulated) device - looks good. See screenshots attached to #7251.

bekuno:
Yes, we need them.
I tried all from https://developer.android.com/studio/write/vector-asset-studio .
But this works not realy good. Using the support library it shows this:
screenshot_20190227-234233
Therfore it is the mixed method:
"Android 4.4 (API level 20) and lower doesn't support vector drawables. If your minimum API level is set at one of these API levels, you have two options when using Vector Asset Studio: generate Portable Network Graphic (PNG) files (the default) or use the Support Library.
For backward-compatibility, Vector Asset Studio generates raster images of the vector drawable. The vector and raster drawables are packaged together in the APK. You can refer to vector drawables as Drawable in Java code or @drawable in XML code; when your app runs, the corresponding vector or raster image displays automatically depending on the API level."

bekuno:
> 
> 
> The PNGs for ldpi and xxxhdpi differ significantly from the other icons. I think that's not intended.

You are right.
These were not overwritten. Fixed now.

UniQP:
Can you squash the commits into one?

UniQP:
Sorry, the xxxhdpi icon size is already correct.

bekuno:
> 
> 
>     * The ldpi icon has no transparent background (in contrast to the other icons)
> 
>     * The xxxhdpi icon has the wrong size (should be 96x96)

deleted the ldpi icon.
(after consulting rsudev)

bekuno:
Image problem solved, squashed and successfully tested on my dev device.

PR #7343, opened 2019-03-04
status: add active connectors reporting

### LINKED ISSUE #7214: Add services usage reporting
c:geo part of related https://github.com/cgeo/cgeo-status/issues/20

It would be great to see services usage distribution. No doubts, gc.com is leader, but the interesting point is other services.

Need to add currently enabled services with its state (authorized or not) to cgeo\main\src\cgeo\geocaching\network\StatusUpdater.java

okainov:
c:geo side of https://github.com/cgeo/cgeo-status/issues/20

Sending new "active_connectors" with comma-separated list of active connectors or empty string, if none.

Mention @samueltardieu 

Fixes #7214

### DIFF
--- a/main/src/cgeo/geocaching/connector/ConnectorFactory.java
+++ b/main/src/cgeo/geocaching/connector/ConnectorFactory.java
@@ -168,6 +168,17 @@ public final class ConnectorFactory {
         return liveConns.toArray(new ILogin[liveConns.size()]);
     }
 
+    @NonNull
+    public static List<IConnector> getActiveConnectors() {
+        final List<IConnector> activeConnectors = new ArrayList<>();
+        for (final IConnector conn : CONNECTORS) {
+            if (conn.isActive()) {
+                activeConnectors.add(conn);
+            }
+        }
+        return activeConnectors;
+    }
+
     public static boolean canHandle(@Nullable final String geocode) {
         if (geocode == null) {
             return false;
--- a/main/src/cgeo/geocaching/network/StatusUpdater.java
+++ b/main/src/cgeo/geocaching/network/StatusUpdater.java
@@ -1,6 +1,8 @@
 package cgeo.geocaching.network;
 
 import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.connector.ConnectorFactory;
+import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.gc.GCMemberState;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.AndroidRxUtils;
@@ -10,7 +12,10 @@ import android.app.Application;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.support.annotation.NonNull;
+import android.text.TextUtils;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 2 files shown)


samueltardieu:
Looks good to me. I'll try to find some time to implement the cgeo-status side, unless someone beats me to it.

main/src/cgeo/geocaching/connector/ConnectorFactory.java
```diff
@@ -168,6 +168,17 @@ private ConnectorFactory() {
         return liveConns.toArray(new ILogin[liveConns.size()]);
     }
 
+    @NonNull
+    public static IConnector[] getActiveConnectors() {
```
samueltardieu: Why not return directly the `List<IConnector>`?

main/src/cgeo/geocaching/network/StatusUpdater.java
```diff
@@ -64,6 +66,22 @@ static Status defaultStatus(final Status upToDate) {
...
+            try {
+                activeConnectorsBuilder.append(conn.getNameAbbreviated());
+                activeConnectorsBuilder.append(',');
+            } catch (IllegalStateException ignored) {
+            }
+        }
+        if (activeConnectorsBuilder.length() > 0) {
+            activeConnectorsBuilder.deleteCharAt(activeConnectorsBuilder.length() - 1);
+        }
+        return activeConnectorsBuilder.toString();
```
samueltardieu: This looks overly complicated to me. Any way to use `StringUtils.join()` or similar instead of reinventing the wheel to join strings with a separator? Even constructing an intermediate list would be more readable IMO, and this method will be called once every 30 minutes as far as cgeo-status goes.

main/src/cgeo/geocaching/network/StatusUpdater.java
```diff
@@ -79,10 +97,14 @@ public void run() {
                 } else {
                     gcMembershipParameters = null;
                 }
+
+                final Parameters activeConnectorsParameters = new Parameters("active_connectors",
+                        getActiveConnectorsString());
+
                 Network.requestJSON("https://status.cgeo.org/api/status.json",
                         Parameters.merge(new Parameters("version_code", String.valueOf(Version.getVersionCode(app)),
                                 "version_name", Version.getVersionName(app),
```
samueltardieu: You could insert "active_connectors" and `getActiveConnectorString()` directly without using an intermediate variable. Those intermediate variables are introduced only when we don't know for sure that the parameter will be present (for example the installer can be absent).

main/src/cgeo/geocaching/connector/ConnectorFactory.java
okainov: No preference from my side, just wanted to be consistent with existing `getActiveLiveConnectors` which returns `ILogin[]`. SHould I change to `List<IConnector>`?

main/src/cgeo/geocaching/network/StatusUpdater.java
okainov: done

main/src/cgeo/geocaching/network/StatusUpdater.java
okainov: The problem is that we need to call extra method for every element in collection, so you cannot apply `map` so easily. If you think that intermediate list will be better (I thought exactly opposite and that's why I used StringBuilder), I can replace

main/src/cgeo/geocaching/connector/ConnectorFactory.java
samueltardieu: The simplest the better, so probably.

okainov: Done

Lineflyer:
@samueltardieu 
- Are your comments resolved OK?
- Does it disturb merging this as long as the server side is not prepared?

samueltardieu:
This is good for merging as far as I am concerned. Java 8 would have made some constructions more readable, but it doesn't look supported by all the libraries at the moment.

PR #7346, opened 2019-03-05
add build-cache to .gitignore

okainov:
New folder got created in PR #7339 but it was not added to .gitignore. Let's improve it


rsudev:
Thanks!

Issue #7261, opened 2019-01-24
GPX import dialog looks broken


Lineflyer:
##### Detailed steps causing the problem:
- Open pocket query list
- Select to import an availabe PQ
(Using file import will probably have the same problem)

##### Actual behavior after performing these steps:
The import progress dialog seems to be transparent (background is interfering with dialog) and the dialog is no longer centered on the device screen

##### Expected behavior after performing these steps:
Centered dialog with background

##### Version of c:geo used:
2019.01.23-NB

##### Is the problem reproducible for you?
Yes

##### System information:
Android 8.0

moving-bits:
I managed to get the dialog box back to centered and with a non-transparent background, but one thing is still missing: the button is cyan text on grey background, for whatever reason:

![image](https://user-images.githubusercontent.com/3754370/51806235-d1ee0380-2277-11e9-8db9-0c78c1f28d0f.png)

How can I change this back to "normal"?
So far I haven't found any combination of styling / themes to achieve this :-(

Lineflyer:
This Cyan...this brings me back to my question I had in a dedicated issue regarding the seamingly two different main styles visible in cgeo now. One of them also uses cyan. Maybe we should clarify this general thing before we redesign the faulty parts?!

rsudev:
In most of the UIs we derive from 'ActionBarActivty', which is bound to certain styles which (AFAIK) are also designed to have a consistent look and feel.
Some activities (Google map, preferences) are derived from their own base classes and (if I understood correctly) require a different set of styles, which are more bound to the look and feel of the platform.
I guess this can be understood an addressed, but as a tool we likely need 'before and after' screenshots for the relevant UIs on a relevant number of API versions.

moving-bits:
I've tested `cgeo.ActionBarStyle` as parent, but this did not work:

![image](https://user-images.githubusercontent.com/3754370/51858634-813ce000-2335-11e9-800b-be8a09f094bc.png)

Dialog box moved to the top, has semi-transparent background and the button still has a grey background, although everything else in styles.xml is configured like in my previous post. Only the font size & color has become more consistent with c:geo.

And if I chose `cgeo_main.ActionBarStyle` as parent, the result is like this:

![image](https://user-images.githubusercontent.com/3754370/51858879-0de79e00-2336-11e9-8c6b-4d19d6f2a01f.png)

Lineflyer:
Fixed with #7294

PR #7348, opened 2019-03-07
change alert dialog styling (fix #7342)

### LINKED ISSUE #7342: Buttons to save caches to list not correct aligned
- open live map
- Show caches as list
- Save offline
- The buttons (OK, break, last) are one above the other

The buttons are one above the other

The buttons are side by side

##### Version of c:geo used:
2019.02.23

Yes

reported by Spooky155 on IRC

moving-bits:
- add theming/styling for AlertDialogs to get c:geo standard buttons as defined in styles
- use this theme for manually created AlertDialogs in StoredList (fix #7342)
- also restores white button text color for other AlertDialog popups (like GPX import, which got changed by fix for #7261)

### DIFF
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -11,7 +11,7 @@
         <item name="android:ellipsize">marquee</item>
         <item name="android:textSize">16sp</item>
         <item name="android:textColor">?text_color</item>
-        <item name="android:background">?button</item>
+        <item name="android:background">?android:attr/selectableItemBackground</item>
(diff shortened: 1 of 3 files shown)


Lineflyer:
Can you post a screenshot of the fixed dalog ?

moving-bits:
SDK 26 with font size "extra large":

image

SDK 26 with font size "Standard":

image

Lineflyer:
Thanks, so lets take a look.

Issue #7350, opened 2019-03-10
update references to other APIs


bekuno:
I did some research which of the referenced libraries could be updated.

The result was not great. (I expected nothing less).

Do we have a documentation for the dependencies? I found nothing in the wiki.
I think it would be easier to see what is necessary for Prepare for targetSdkVersion 28 (#7321).
Should I create such an overview, if yes - where?

Lineflyer:
> Should I create such an overview, if yes - where?

I think the wiki might be the right place for such an overview.

bekuno:
Many work is now already done by @Bananeweizen

PR #7351, opened 2019-03-10
API Updates to maximal usable without code changes

bekuno:
API Updates to maximal usable without code changes:
(see also #7350)

- gradle-wrapper.properties:
  - update to the latest gradle plugin of 4.10.x

- main\build.gradle
  - org.apache.commons:
	- commons-collections4:4.2
    - commons-compress:1.18
    - commons-io:2.6
    - commons-lang3:3.8.1
   - org.assertj: Patch 2.9.1
  - com.fasterxml.jackson.core: Patch 2.9.6
  - com.google.android.gms:10.0.1 (latest)
  - io.reactivex.rxjava2: (latest)
	- rxjava:2.2.7
	- rxandroid:2.1.1
  - com.android.support.test:runner:1.0.2 (latest)

- build.gradle
  - Update compatible to AS 3.2.2:
    - com.android.tools.build:gradle:3.3.2'
  - other plugins as max as possible
    - com.github.ben-manes:gradle-versions-plugin:0.21.0'
    - de.mobilej.unmock:UnMockPlugin:0.7.3'
    - com.google.gms:google-services:2.1.2'
    - com.getkeepsafe.dexcount:dexcount-gradle-plugin:0.8.6'
    - com.gradle:build-scan-plugin:1.16'
  - remove some outdated checkes
  - Configuration (dependencyUpdates) for gradle-versions-plugin
	(gradleReleaseChannel starting 0.21.0)


bekuno:
@kumy can you update the environment in the next day for the new packages?

kumy:
@bekuno May you propose a PR on https://github.com/cgeo/cgeo-executor/blob/master/android-packages ?

Looking at the errors, they don't seems to be Android packages, but libraries, no?

bekuno:
Yes, only libraries.
So normaly nothing to do for you?

It is strange that jenkins gives such errors:
  Could not resolve com.android.support:support-annotations:24.2.1
This is a standard condition.
Maybe it is a problem with different library versions. I have also this error:
  Dependency path 'cgeo:main:unspecified' --> 'com.android.support.test:runner:1.0.2' -->   'com.android.support.test:monitor:1.0.2' --> 'com.android.support:support-annotations:27.1.1'
But there was no such error visible in AS. 
I will check that tomorrow.

kumy:
IINM, gradle is responsible for downloading such libraries.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.3-all.zip
```
Bananeweizen: It is not sufficient to just change the URL. You also need to update the wrapper script itself. See https://gradle.org/install/#with-the-gradle-wrapper
When you execute that, both the properties and the wrapper should be updated.

bekuno: What script (gradlew ?) ant which point should be updated?

bekuno: Ok, will do it.

PR #7355, opened 2019-03-10
temporarily disable static maps feature (fix #7332)

### LINKED ISSUE #7332: Remove Google static map feature
Google stopped offering static maps for free. This is the reason, why the feature stopped working in c:geo. Details can be found here: #7150

The non working feature should be removed from c:geo as I don't expect Google to change its mind or c:geo team to be able to pay for such maps.

An alternative source (if there is any) could be implemented at the same step or afterwards.

Please be aware:
Currently all user installations will still send request to Google for static maps as the feature is still in and the default value is "Active". So removing it, might also a bit increase the performance.

moving-bits:
- settings provider always returns false for "static maps" and "static wp maps" to effectively disable static maps feature
- disable two menu entries in settings => offline data

This is a temporary disabling on purpose, so that we can later decide on either removing google static maps provider completely or replacing it by a different static maps provider.
Release note should mention that we have to disable static maps feature temporarily at least (and with google maps for sure) due to the licensing changes on google side.

### DIFF
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -614,11 +614,13 @@
             android:defaultValue="true"
             android:key="@string/pref_offlinemaps"
             android:summary="@string/init_summary_offline"
+            android:enabled="false"
             android:title="@string/init_offline" />
         <CheckBoxPreference
             android:defaultValue="false"
             android:key="@string/pref_offlinewpmaps"
             android:summary="@string/init_summary_offline_wp"
+            android:enabled="false"
             android:title="@string/init_offline_wp" />
         <CheckBoxPreference
             android:defaultValue="false"
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -649,11 +649,15 @@ public class Settings {
     }
 
     public static boolean isStoreOfflineMaps() {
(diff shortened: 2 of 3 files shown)


moving-bits:
retest this please

moving-bits:
Ok, I've _temporarily_ disabled static maps generation in the tests (and checking for static maps as well)

Lineflyer:
Looking into the details of this PR, you did now include all or many commits currently on purpose only present on `master`. I think this was not intentious?!

moving-bits:
No, something with rebasing did not work on my machine. I had the same problem with the other PR yesterday. Will rebase and force-push.

Lineflyer:
IMHO this is valid as a workaround on `release` to avoid unnecessary requests. On the other side we could remove the code completely. For reuse of the code (e.g. for another map provider) we would still have it in the history....or whats the best style to handle such thing?

moving-bits:
Currently it's just a quick "get ready for release branch" way of temporarily disabling the static maps generation (and testing).
=> IMHO it is ok to get this release version out to market.

For master branch I would completely delete those functions if we do not plan to reimplement it based on a different map provider in the near future (which I doubt). If, in any case, we do want to reimplement it later, we can still dig out the old code using Git.
=> Will remove the static maps generating functions in the upcoming days.

PR #7360, opened 2019-03-11
Fix #7359 - Raise minSDK to Android 4.1 (SDK16)

### LINKED ISSUE #7359: Raise minSDK to 16
Could it be, that we discussed in #7078 to raise minSDK from 14 to 16 (due to the not supported TLS versions needed to authorize to geocaching.com) but only adapted the closeout warning but never raised the minsdk accordingly?

In `build.gradle` I do still see the minsdk=14

If someone can doublecheck if I got this right, we should raise it now.
Garmin Monterra and other 4.0.x devices are not working online (with GC.com) anymore.

Lineflyer:
Please doublecheck as described in #7359 whether this is the right file to change minSDK before merging.

I targeted this PR to `release` as I would like to have this change out asap. Android 4.0.x can no longer login and use online functions. A closeout warning has already been given to users since middle of last year.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -29,7 +29,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 14
+        minSdkVersion 16
         targetSdkVersion 27
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)


UniQP:
I think we should also adapt `ext.minSdkVersion` in `build.gradle`.

rsudev:
For c:geo as such only the minSdk in the main/build.gradle should be relevant, therefore I'd only change that in this context.
Checking and re-aligning the sdk boundary conditions of the other components is a separate concern IMO.

PR #7370, opened 2019-03-14
Changed string osm:mapnik to osm:default

HolgerJeromin:
Mapnik is the rendering software, not the style name

The internal names should be adapted, too. But this is not so important.

https://wiki.openstreetmap.org/wiki/Mapnik
Another possible name would be 
Osm:Carto
https://wiki.openstreetmap.org/wiki/Standard_tile_layer
But this would be confusing imo

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -716,7 +716,7 @@
     <!-- map sources -->
     <string name="map_source_google_map">Google: Map</string>
     <string name="map_source_google_satellite">Google: Satellite</string>
-    <string name="map_source_osm_mapnik">OSM: Mapnik</string>
+    <string name="map_source_osm_mapnik">OSM: Default</string>
     <string name="map_source_osm_cyclemap">OSM: Cyclemap</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="init_sendToCgeo">Send to c:geo</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -716,7 +716,7 @@
     <!-- map sources -->
     <string name="map_source_google_map">Google: Map</string>
     <string name="map_source_google_satellite">Google: Satellite</string>
-    <string name="map_source_osm_mapnik">OSM: Default</string>
+    <string name="map_source_osm_mapnik">OSM: Map</string>
     <string name="map_source_osm_cyclemap">OSM: Cyclemap</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="init_sendToCgeo">Send to c:geo</string>


SammysHP:
There is no "default" OSM style. Maybe "openstreetmap.org".

Lineflyer:
We could just name it "Openstreetnmap"?!

SammysHP:
But there are many map providers that use OpenStreetMap data.

bekuno:
This is only the base text. But what about the translations?
There we have:
- fr:        OSM : rendu Mapnik
- ja:        OpenStreetMap:Mapnik
- other:  OSM: Mapnik
There was before the now removed "OSM: Cyclemap" to show a difference.
My favorite is "OpenStreetMap.org" for the current implementation.

@rsudev If I see it right it is based of the old mapsforge implementation. Do you plan to switch to new implementation?

Lineflyer:
@bekuno 
After a string in `/res/values/string.xml` has been either changed or newly added, we just need to trigger the upload to our crowd translation platform Crowdin (AFAIK there are some scripts for it). 
This will trigger the crowd translators to retranslate it and on next download from Crowdin the new translation comes in or at least the outdated translation will be removed (if not yet translated into that specific language).

I agree to "OpenStreetMap" but would recommend to leave out the ".org". We also only use "Google" but not "Google.com"

HolgerJeromin:
> I agree to "OpenStreetMap" but would recommend to leave out the ".org". We also only use "Google" but not "Google.com"

"Google" is the datasource, equivalent to "OSM". We have right now 
````
Google: Map
Google: Satellite
````
Do you really want `OpenStreetMap` next to `OSM: Cyclemap`?
````
Google: Map
Google: Satellite
OpenStreetMap
OSM: Cyclemap
````
This is IMO a total mess. 

But "OpenStreetMap" is far better than the wrong "mapnik".

**Comment by bekuno** [comment]
For CycleMap see #6610. It is removed.

**Comment by Lineflyer** [comment]
I don't want to artificially expand this topic, but while we are on this now:
Given the naming our users are used to (e.g. "Google: Map") and the pattern, which is derived from that, we could use this opportunity to harmonize it through all sources.

My proposal would be:
```
Google: Map
Google: Satellite
OSM: Online  <-- instead of Mapnik
OSM: <file name> <-- instead of <file name> (Offline)
```

Not 100% following the scheme, but most understandable by users IMHO.

rsudev:
'Mapnik' is/has been at least from my point of view a synonym for its default rendering style for more than a decade, so I do not see anything 'wrong' here.
That said I am not against revisiting the naming conventions for the map source menu, but I doubt there is a way to unambigously capture the diverse degrees of freedom this linear menu encompasses (coverage, online/offline, Google/OSM, map/satellite imagery, ...)

HolgerJeromin:
> 'Mapnik' is/has been at least from my point of view a synonym for its default rendering style for more than a decade, so I do not see anything 'wrong' here.

Technically it is wrong. In the past osm.org had one unnamed style with the renderer mapnik and one unnamed style with the renderer osmarenderer. So the people learned to differentiate between the two with the renderer name.

Now osm.org features four styles (which are all rendered with mapnik).
Often people report style issues on the renderer ([example](https://github.com/mapnik/mapnik/issues/4027)).
This will only change if all places uses precise wording :-)

cgeo should not be confused with the service/database/website geocaching.com or the official geocaching app, too.

HolgerJeromin:
I like the proposal of @Lineflyer so i updated the PR.

Lineflyer:
Objections to merge this. Otherwise I will merge tonight.

Issue #7340, opened 2019-03-03
Navigation with maps.me not starting


Lineflyer:
##### Detailed steps causing the problem:
- Have maps.me installed
- Open cache details of a cache
- Select Navigate -> maps.me

##### Actual behavior after performing these steps:
Nothing happens, maps.me not starting

##### Expected behavior after performing these steps:
maps.me should be started with cache coordinates as destination

##### Version of c:geo used:
2019.02.23

##### Is the problem reproducible for you?
Yes (for me) and reported via FB

##### System information:
Tested on my Samsung S7, Android 8.0

Lineflyer:
Reason could(!) be, that targetSDK of the API code is different to the main app:
https://github.com/cgeo/cgeo/blob/master/mapswithme-api/build.gradle

Lineflyer:
Well, the alignment of the sdk versions does not solve it, I changed it nontheless.

Interesting enough it work from the list view (Cache list -> Show on map -> Maps.me) but not from cache details (Cache details -> Navigate -> Maps.me). The debug log simply tells nothing.

Maybe someone with more code insights can take a look regarding the difference between the mentioned menu items/methods.

bekuno:
I did not have this problem on devices Android 7.0 (Samsung S6) and Android 9 (samsung Note 8).
Maybe a system specific problem? Or only when there is no downloaded map in maps.me?

Lineflyer:
> I did not have this problem on devices Android 7.0 (Samsung S6) and Android 9 (samsung Note 8).

Really? Also from cache details Navigation menu, not only from cache list?

Lineflyer:
Addon:
Different from the sequence initially described, there is no need to have maps.me installed to see a problem. Even without this app being installed, there should be a message shown (see the API code of maps.me we are using) but it doesn't. 
So I guess the problem is within c:geo and not a problem of intent sending, etc.

Lineflyer:
Could it be, that this line of code
https://github.com/cgeo/cgeo/blob/1fcabfb076b5b0ccc15992005c12d7013cc53849/main/src/cgeo/geocaching/apps/navi/MapsMeApp.java#L57

is somehow since SDK26 preventing maps.me being called at all?

@bekuno 
You wrote, that you do not have this problem on some devices...can you please doublecheck? I can reproduce the problem on Android 7 and 9.

bekuno:
[...]

PR #7373, opened 2019-03-15
Align SDK versions of mapswithme to main app

Lineflyer:
Might be related to #7340


Lineflyer:
Well, this change does not solve the related issue #7340 but at least it compiles and looks meaningful to align.

PR #7375, opened 2019-03-15
remove all static map functions and settings (fix #7332)

### LINKED ISSUE #7332: Remove Google static map feature

moving-bits:
removes all static map related functions, menu entries, settings, preference keys, strings [english only] and tests


moving-bits:
Reminder to myself (not needed for this PR): The following preference values could be deleted from c:geo internal database: "offlinemaps", "offlinewpmaps", "navigationStatic", "navigationStaticDownload"

Lineflyer:
> Reminder to myself (not needed for this PR): The following preference values could be deleted from c:geo internal database: "offlinemaps", "offlinewpmaps", "navigationStatic", "navigationStaticDownload"

Maybe you should file a dedicated issue for this (if useful) and assign yourself to it.

Lineflyer:
Lets give it a test drive in our nightlies

PR #7376, opened 2019-03-16
add layout captures directory of Android Studio to .gitignore

moving-bits:
add "captures" directory to .gitignore
("captures" directory is used by Android Studio for the Layout Inspector)


Lineflyer:
Looks reasonable to me. Merging.

PR #7380, opened 2019-03-17
Fix #7379 - Change CACHE_URL_LONG to currently used scheme

### LINKED ISSUE #7379: Double app selection needed for "open in Browser"
- Open a gc.com cache in c:geo
- Select Menu - Open in browser

- The system asks which app to use to open https://www.geocaching.com/seek/cache_details.aspx?wp=...
- After seleting the desired browser the system asks again which app to use as the URL above is redirected to https://www.geocaching.com/geocache/....

Shouldn't it be possible to avoid the double selection if we directly open https://www.geocaching.com/geocache/<GCCODE>

##### Version of c:geo used:
2019.03.17-NB

Yes

##### System information:
Android 8.0

Lineflyer:
Not sure if there is any dependency with this variable, which could lead to problems at other places.
Let CI test it first and review appreciated.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConnector.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConnector.java
@@ -58,7 +58,7 @@ public class GCConnector extends AbstractConnector implements ISearchByGeocode, 
     private static final String CACHE_URL_SHORT = "https://coord.info/";
     // Double slash is used to force open in browser
     @NonNull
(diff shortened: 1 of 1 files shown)


Lineflyer:
Tested the debug.apk and this change indeed fixes the behavior described in #7379.
Setting @rsudev for a review.

Lineflyer:
Watchlist just tested and still working. Thanks for the review.

PR #7388, opened 2019-03-17
fix #7249 group trackable logs by 10

### LINKED ISSUE #7249: Logging sporadically has error result with big amount of TB in inventory
Since several months I am chatting with a user on support, who keeps experiencing error messages when trying to send a geocache log to gc.com sporadically. The failure rate is varying from permanent to no problem during a day of caching.
The failure started for him with the release, where we changed to the new logging page (2018.08.14).

The exact problem is:
- Send composed log 
- Sending time is high 
- c:geo reports, that log might not have been sent
- Browser check shows, that log in fact has been posted

We checked and sorted out many causes over time, from what I remember it was:
- Reinstallation did not help
- Factory reset did not help
- Logs with or without picture make no difference
- Favorite point assignment makes no difference
- Parallel usage of other apps/website in browser not existing
- Parallel usage of same account on other device not existing
- Length of log makes no difference
- Not depending on "Report problem" selection

Current questions I gave to the user:
- Any trackables in inventory
- Any special characters in log text or signature

Does anyone have any other idea of
a) what could cause this issue
b) what could be done to track down the root cause

Not any high priority but just that I am clueless....

@pstorch In case you read this: As you implemented this logging, maybe you have some idea?

pstorch:
to avoid timeouts

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
@@ -374,6 +374,9 @@ class GCWebAPI {
     }
 
     /**
+     * Sends trackable logs in groups of 10.
+     * https://github.com/cgeo/cgeo/issues/7249
+     *
      * https://www.geocaching.com/api/proxy/trackable/activities
      * <div>
      *   "postData": {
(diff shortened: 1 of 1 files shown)


Lineflyer:
@pstorch Thanks very much!
I will give it a test drive and report back.

PR #7392, opened 2019-03-18
Make max radius for okapi connector searches configurable (fix #7279)

### LINKED ISSUE #7279: search for address, having multiple caching providers active, gives results hundreds of km away
- I have gc.com, oc.de and oc.nl enabled
- when I do a search for address, I get 20 results for each provider
- some of these results are hundreds of km away

- oc.nl sends me some result at the border of Germany & Netherlands
- although I'm doing a search around, lets say, Cologne or Munich

- feature request: skip results which are more than x <variable setting> km away

##### Version of c:geo used:
- last release and current NB

Yes

moving-bits:
- make max radius for okapi connector searches configurable (fix #7279)
- standardize preference seekbar layouts (replaces brouter & wp seekbar layout + use it for okapi max distance seekbar as well)


okainov:
Please generalize approach to be applicable for other connectors (i.e. SU) as well.

main/src/cgeo/geocaching/settings/OCMaxDistancePreference.java
```diff
@@ -0,0 +1,112 @@
...
+    public OCMaxDistancePreference(final Context context, final AttributeSet attrs, final int defStyle) {
+        super(context, attrs, defStyle);
+        init();
+    }
+
+    private void init() {
+        setPersistent(false);
+    }
+
+    private int km2valueHelper(final double distance) {
```
okainov: Doesn't seem to be belonging to `settings`. There was some `util`-like package, please put it there

main/src/cgeo/geocaching/settings/OCMaxDistancePreference.java
```diff
@@ -0,0 +1,112 @@
...
+import android.view.ViewGroup;
+import android.widget.SeekBar;
+import android.widget.SeekBar.OnSeekBarChangeListener;
+import android.widget.TextView;
+
+import java.util.Locale;
+
+import butterknife.ButterKnife;
+
+public class OCMaxDistancePreference extends Preference {
```
okainov: I think that methods\classes in common `settings` folder should not be tightened to particular service

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -75,6 +75,8 @@
     public static final int SHOW_WP_THRESHOLD_MAX = 200;
     private static final int BROUTER_THRESHOLD_DEFAULT = 10;
     public static final int BROUTER_THRESHOLD_MAX = 120;
+    public static final int OCMAXDISTANCE_DEFAULT = 200;    // km
```
okainov: Apart of being OC-specific (mentioned before), can we use `OC_MAX_DISTANCE_KM_DEFAULT` or something like this to remove this comment and make the units explicit in the name of the constant?

main/res/layout/preference_seekbar.xml
```diff
@@ -10,7 +10,7 @@
     tools:context=".settings.SettingsActivity" >
 
     <SeekBar
-        android:id="@+id/wp_threshold_seekbar"
```
okainov: Was it intended to remove `wp_threshold_preference.xml` file?

main/res/layout/brouter_threshold_preference.xml
```diff
@@ -1,30 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
okainov: I would suggest to separate non-related changes to at least different commit

main/res/layout/preference_seekbar.xml
moving-bits: Yes, we already had two seekbar layout files (for wp_threshold and brouter_threshold), and this PR would have added a third one, although they are basically the same, so I refactored the two other settings to use the same layout file, giving it a generalized name.

main/src/cgeo/geocaching/settings/OCMaxDistancePreference.java
moving-bits: This conversion formula is specific to the seekbar length and the way that specific seekbar accelerates. (Android only offers linear seekbars, but by using those two formulas I made it exponential, which suits better for distance setting over such a large range, IMHO.) That's why I put it into this specific seekbar class.

moving-bits:
> Please generalize approach to be applicable for other connectors (i.e. SU) as well.

I don't know about the SU connector yet, but I had checked it for the GC connector and did not find an reasonable way to implement it for that. As the reason why I implemented this function lays in the constant radius for the OKAPI connectors, I chose to implement it for them only (which is the many opencaching platforms c:geo supports). Should I rename the functions to OKAP_xxx instead of OC_xxx then?

main/res/layout/brouter_threshold_preference.xml
moving-bits: They have been during development, but I normally squash them before finalizing a PR.

okainov:
I would suggest just "searchRadius" with maybe text note (for connectors which support it). Okapi has no difference with "oc" in this context.

ATM there is at least EC connector which also has hardcoded "20". And I hope my SU masterpiece will be merged really soon. 

As more neat but detailed solution, it could be individual per-connector setting (like icons for EC)

moving-bits:
> As more neat but detailed solution, it could be individual per-connector setting (like icons for EC)

ok, I'll look into it - there is some interface for connector specific settings, maybe I can extend this.

moving-bits:
Closing this PR - will rethink the approach according to the issues linked in #7279

PR #7406, opened 2019-03-21
code cleanup (merging layout files)

moving-bits:
Fix for #6157 introduced a new layout file instead of reusing the existing one. This PR merges those two layout files under a generalized name + adjusts their references.


Lineflyer:
Looks straight forward.

PR #7408, opened 2019-03-21
Prefixed name of offline map to "OSM: offline"

HolgerJeromin:
Follow up to #7370

This was a proposal of @Lineflyer 
I missed that this part is solvable in the same file.
So I opened a new PR.


Lineflyer:
That wont work as AFAIK this string is no longer be used but currently only the filename without extension.

HolgerJeromin:
I suspected something like that. But my first tests with offline maps did not worked anymore (had it working a few years ago), so i was not really sure.

PR #7409, opened 2019-03-22
Use try with resources

UniQP:


Bananeweizen:
Now I'm confused. try-with resource is only available from level 19, but we have minimum 16, right? Shouldn't this PR build have failed?

UniQP:
You're right. Maybe the CI doesn't check SDK 16.

Should I revert the commit?

Bananeweizen:
Yes, I suggest to revert. I'm still not sure how to configure Android Studio for this. In Eclipse I got the error message about API level the moment I pulled the change.

UniQP:
I reverted the commit on master.

PR #7410, opened 2019-03-22
fixes for opening map from address search (fix #7386) (fix #7387)

### LINKED ISSUE #7386: Open map from address search shows routing line
Testing the change from #7228 (Add icon to address search to show on map) I noticed that on the opened map the routing line with beginn or target(?) to map center will be shown on the opened live map.

I don't think this is useful and helpful and maybe not even intentionally done?!

### LINKED ISSUE #7387: Live map opened from address search cannot be disabled
- Open search menu
- Search for a location/address
- Click map icon to open the map

The map menu does not contain the menu item "disable live" as it does on the "normal" live map

"Disable live" should be working the same on this map as on the "normal" live map

##### Version of c:geo used:
2019.02.23

Yes

moving-bits:
don't show direction line when called from address search
show "enable live / disable live" menu entry when called from address search

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -463,6 +463,11 @@ public class CGeoMap extends AbstractMap implements ViewFactory {
 
         overlayCaches = mapView.createAddMapOverlay(mapView.getContext(), Compatibility.getDrawable(getResources(), R.drawable.marker));
 
+        final Geopoint savedCoords = mapOptions.coords;  // remember for centerMap
+        if (mapOptions.mapMode == MapMode.LIVE && savedCoords != null) {
+            mapOptions.coords = null;   // no direction line, even if enabled in settings
+            followMyLocation = false;   // do not center on GPS position, even if in LIVE mode
+        }
 
         overlayPositionAndScale = mapView.createAddPositionAndScaleOverlay(mapOptions.coords, mapOptions.geocode);
         if (trailHistory != null) {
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -423,6 +423,12 @@ public void onCreate(final Bundle savedInstanceState) {
 
         ArrayList<Location> trailHistory = null;
 
+        final Geopoint savedCoords = mapOptions.coords;  // remember for centerMap
+        if ((mapOptions.mapMode == MapMode.LIVE) && (mapOptions.coords != null)) {
```
UniQP:
```suggestion
        if (mapOptions.mapMode == MapMode.LIVE && savedCoords != null) {
```

main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -423,6 +423,12 @@ public void onCreate(final Bundle savedInstanceState) {
 
         ArrayList<Location> trailHistory = null;
 
+        final Geopoint savedCoords = mapOptions.coords;  // remember for centerMap
```
UniQP: Is it possible to move the whole block down to fix the codacy issue?

moving-bits: done

main/src/cgeo/geocaching/maps/CGeoMap.java
moving-bits: done

PR #7413, opened 2019-03-23
 Remove superfluous try block

UniQP:


Bananeweizen:
The block might have existed due to the possibility of new GeoPoint("...") to throw an exception, but I also think this is useless.

PR #7415, opened 2019-03-24
remove unused OSM cyclemap strings

Bananeweizen:
The translations cause a warning, since that key does not exist anymore.
Let's remove them from source, even if they may need to be removed from
crowdin additionally (I'm not sure of that).


Lineflyer:
It should be OK to manually remove them. As the source string is no longer existing the incoming translations from Crowdin should also no longer contain them. 
So it should not lead to any conflict if these strings are already gone before the import.

But this PR needs rebasing it seems.

Bananeweizen:
No longer needed due to translation update.

PR #7417, opened 2019-03-24
disable eclipse autosave actions

Bananeweizen:
Since everyone uses AS meanwhile, disable the Eclipse autosave actions
to avoid the import order getting shuffled when actually using Eclipse.


tests/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -40,6 +40,8 @@ sp_cleanup.qualify_static_member_accesses_through_subtypes_with_declaring_class=
 sp_cleanup.qualify_static_member_accesses_with_declaring_class=true
 sp_cleanup.qualify_static_method_accesses_with_declaring_class=false
 sp_cleanup.remove_private_constructors=true
+sp_cleanup.remove_redundant_modifiers=false
+sp_cleanup.remove_redundant_semicolons=false
```
UniQP: These lines looks unrelated to the commit message (and are only added in this file).

Bananeweizen: Good catch. Fixed.

Bananeweizen: Fixed (And what is the button "Resolve conversation" doing, which I see below this input box? I never saw that before.)

UniQP: Resolved conversations are hidden by default.

PR #7424, opened 2019-03-25
fix more warnings

Bananeweizen:
* static imports from class defining the constant
* remove unused members
* javadoc missing arguments documentation
* unnecessarily nested block
* missing override
* empty method


main/src/cgeo/geocaching/utils/ColorUtils.java
```diff
@@ -38,8 +38,7 @@ public static double getContrastRatio(final int rgb1, final int rgb2) {
         final double lumi2 = getLuminance(Color.red(rgb2), Color.green(rgb2), Color.blue(rgb2)) + 0.05;
         if (lumi1 > lumi2) {
             return lumi1 / lumi2;
-        } else {
-            return lumi2 / lumi1;
         }
+        return lumi2 / lumi1;
```
UniQP: `return Math.max(lumi1, lumi2) / Math.min(lumi1, lumi2);`
is shorter and more readable.

Bananeweizen: Done.

PR #7425, opened 2019-03-25
replace deprecations from commons.lang3

Bananeweizen:
Some classes from commons.lang3 are meanwhile in a separate project,
and the old classes are deprecated.


Bananeweizen:
Have to re-learn proguard first, as it looks like...

Bananeweizen:
For me, something seems completely broken in our Android Studio configuration. The import order that both checkstyle and codacy complain about is exactly the order that AS creates (I just deleted all imports to check).
Can someone verify that in his installation, maybe? And if that is actually the case for everyone and not just me, then I'm very much in favor of just disabling the "ImportOrder" check in all tools.

rsudev:
The AS import config is definitely not aligned with the checkstyle definition...
I'll check if this can get fixed.

rsudev:
It looks like it is possible to have a compatible configuration for the import optimization on project level, but the respective file is seemingly not under revision control.

bekuno:
seen in an other .gitignore:
!.idea/codeStyles
Maybe this takes under revision control.

Bananeweizen:
retest this please

Issue #7382, opened 2019-03-17
custom task named 'wrapper' has been deprecated


bekuno:
in "Configure project "

Creating a custom task named 'wrapper' has been deprecated. This is scheduled to be removed in Gradle 5.0. You can configure the existing task using the 'wrapper { }' syntax or create
 your custom task under a different name.
        at build_w5tx7p2bosgsq4d6b4f13ifv.run(C:\StudioProjects\cgeo\build.gradle:186)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)

Lineflyer:
@bekuno 
It would be helpful if you also set an appropriate label when opening an issue. That's not mandatory but helps a bit to keep the overview and sort/filter later on, therefore I would recommend it.
I set label "Build tools" as it seems matching.

Also not forget to assign yourself if you plan to work on it or are already working on it. This helps to avoid someone else picking it up at the same time.

If you can't set labels or assignment, let me know. Then something with your repo access rights is not ok. :)

PR #7427, opened 2019-03-26
fix gradle warnings

Bananeweizen:
* tasks of type wrapper are deprecated (this is only shown with
--warning-mode all)
* variant.install() is deprecated. This is shown in every build. We can
silence the warning by adding one more indirection towards the new
installProvider()

The compile scope warning is still there and not fixed with this change.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,3 +12,6 @@ org.gradle.configureondemand=true
 
 # activate build caching
 org.gradle.caching=true
+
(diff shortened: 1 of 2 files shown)


bekuno:
Your PR will fix #7382.

Bananeweizen:
Removed overlapping parts and reworded commit message to indicate only the variant.install() fix and the release notes suppression.

PR #7430, opened 2019-03-28
upgrade dependencies

Bananeweizen:
Since the switch to gradle 5.x and Java 8 we can again upgrade more
dependencies.


Bananeweizen:
retest this please

Lineflyer:
retest this please

Lineflyer:
For whatever reason this seems to be a real failure on CI despite for the CI having some hickups last days.
Looks a bit like the character encoding was influenced causing the test to fail?!

Bananeweizen:
okhttp behaves differently when upgraded. I reverted that, so tests should be green now

PR #7431, opened 2019-03-28
use assertj

Bananeweizen:
* more readable (left to right)
* more expressive (hasSize(), contains(), isEmpty()...)
* better error messages
* junit assert is deprecated

Really please don't introduce new assertions using plain junit assert.


Lineflyer:
retest this please

Issue #7193, opened 2018-11-17
Raise Target SDK to 26 for contacts plugin (or reintegrate into main app)


Lineflyer:
Hello Google Play Developer, 
In October, we announced updates to our Permissions policy that will limit which apps are allowed to request Call Log and SMS permissions. This policy will impact one or more of your apps. 
Only an app that has been selected as a user's default app for making calls or text messages, or whose core functionality is approved for one of the exception use cases, will be able to request access to Call Log or SMS permissions. 
Action required 
Below, we've listed apps from your catalog which do not meet the requirements for permission requests. Please remove any disallowed or unused permissions from your app's manifest (specified below), migrate to an alternative implementation (e.g. SMS Retriever API for most cases of OTP verification), or evaluate if your app qualifies for an exception. 
Next steps

Read through the Permissions policy and the Play Console Help Center article, which describes intended uses, exceptions, invalid uses, and alternative implementation options for usage of Call Log or SMS permissions.

Update your app or submit a Permissions Declaration Form.

Option 1) If your app does not require access to Call Log or SMS permissions: Make appropriate changes to your app by removing the specified permissions from your app's manifest or migrating to an available alternative implementation by January 9, 2019. 

Option 2) If your app is a default handler or you believe your app qualifies for an exception: Please submit a request via the Permissions Declaration Form. You do not need to have implemented APK changes in order to submit a form. Declaration Forms received by January 9, 2019 may be eligible for additional time to make changes to bring their app(s) into compliance. If you have recently submitted a Permissions Declaration Form, we are in the process of reviewing your information and will respond to your application.

Make sure that your app is otherwise compliant with all other Developer Program Policies to prevent your app from being removed.

Alternatively, you can choose to unpublish the app. 
Our Developer Program Policies are designed to provide a safe and secure experience for our users while also giving developers the tools they need to succeed. That is why we will remove apps that violate our policies. In cases of repeated or serious violations of our policies, we may also terminate your developer account and any related developer accounts. 
We appreciate your willingness to partner with us as we make these improvements to better protect users. 
Affected apps 
Affected apps and permissions are listed below, up to 20; if you have additional apps, please ensure that they are also compliant with the Permissions policy. 

cgeo.contacts READ_CALL_LOG

samueltardieu:
Now that cgeo will have to request permissions as needed, the contacts plugin could be reintegrated in cgeo and removed from the store.

SammysHP:
[...]

PR #7432, opened 2019-03-28
cgeo-contacts: upgrade according to c:geo

bekuno:
see #7193

- build.gradle
  - java 8
  - lib updates
- ContactsActivity.java
  - replace with try with ressources (Java 8)

The READ_CALL_LOG permission is now not more implicit necessary, because minSdkVersion is now 16.
(https://developer.android.com/reference/android/Manifest.permission.html#READ_CALL_LOG)


bekuno:
@Lineflyer What is to do to get now the contacts-addon.apk?

Lineflyer:
The addon is currently taken down from Google Play. Details in #7193

Lineflyer:
The plugin wont work again with only these changes AFAIK
We need to handle the runtime read_contacts permission handling.

Lineflyer:
@Bananeweizen Given the fact that you initially created this plugin AFAIK: Due to the requirements of Google Play we had to depublish it some months ago. We could republish it as it has no the matching targetSDK but it still needs the related permission handling. As the permissions were the reasons to initially publish it as a plugin, it makes even more sense to integrate it into the main app rather than pusblishing the plugin again, as we have runtime permission handling available now.
If you could help getting this function back into the main app, it would be great. At this moment only the existing installations are still out in the field, but new users or users not having installed it previously cannot use the feature.

rsudev:
@Lineflyer: As long as we still target a minsdk that does not yet have runtime permission handling, the initial reason for this plugin still stands, so re-publishing with extension of a (simple) permission handling is IMO the way to go.

Lineflyer:
@rsudev That is a valid argument. But on state-of-the-art android versions we are fine. At the end it depends also on the effort you want to spend to implement permission handling on the seperate plugin.

Issue #330, opened 2011-08-31
Download caches as non blocking service


ludoo0d0a:
Could it be possible to move the download task as a non blocking service ?
Like some file manager, I suggest to add a "hide" button on the popup.

okainov:
Was there any update on the implementation for this? Still would be really nice to have this functionality... @triakcz seemed to have proper implementation years ago, but seems like it was not merged for some reason...

fm-sys:
If someone/I should ever work on this, a `IntentService` running is foreground mode should the way to go... (same as for download receiver)

fm-sys:
Hi, coming back to this very old issue:

We meanwhile offer downloading the routing and map updates in the background, and this process could probably be reused for background cache updates without much additional effort.

However, one general question remains:
When should the updated data be presented to the user?
Currently, the user will see the results as soon as the download is complete and the progress bar gets dismissed. However the UI was blocked while the download was running anyway, so the question didn't arise. 

But how should cgeo behave when the refresh is happening in the background without blocking the UI? Refresh the UI for every single cache, refresh as soon as everything was downloaded or even don't refresh automatically at all to do not interrupt the user? 
The problem is, that a non-blocking download will allow users to navigate to totally different places in the meantime... What if a user views a cache (CacheDetailsActivity) which is refreshed at the same time? 

I would be interested to work on this issue as soon as these things are clarified... (hopefully time will permit)

okainov:
>Currently, the user will see the results as soon as the download is complete and the progress bar gets dismissed. However the UI was blocked while the download was running anyway, so the question didn't arise.

Side note: if user just clicks outside of loading popup, it disappears and user has no idea about the status, if one starts new download, it doesn't work, only restarting c:geo works. So the problem was somehow still there.

As of "when to reload" - IMO at whole batch download finish, not after every cache (because downloads might be of hundreds/thousands of caches).

As a user I would expect there should be some activity/window/screen like "download status" which would show the status of downloads. Probably it might be one screen for all downloads in c:geo: maps, caches, whatever else. And when download finishes, there can be some kind of notification (starting with simple toast and up to actual notification and some "notification" overlay over the download manager icon)


>What if a user views a cache (CacheDetailsActivity) which is refreshed at the same time?

[...]

PR #7438, opened 2019-03-30
add wait for images loading

### LINKED ISSUE #7395: Saved caches from gc.su cannot be opened without loading operation
I will explain the way I found this problem, there might be others or not this might not describe the real problem or be reproducible:
- Have only gc.su geocaching platfrom active
- Open "search" menu
- Search for address "Moscow"
- The search results in 721 caches for me
- Use Menu - Save for offline to save them onto a list (which is BTW really nice and fast)
- Open the resulting list
- Click on a cache on the resulting list

It shows a progress window with "Loading cache details" for at least 20-30 seconds, afterwards the cache is shown.

Cache should be opened without online operation as it is saved offline

##### Version of c:geo used:
2019.03.18-NB1 (first version with gc.su)

Not yet tested in depth

##### System information:
Samsung S7, Android 8.0

From my experience it could be related to additional services which are active (e.g geokrety.org, gcvote.com) ?! Not yet tested
Will look into the debug logs tomorrow

@okainov 
Did not look into the new code in detail:
Did you implement any `Log.d` outputs. Might be helpful to see whats going on like on the other connectors.

okainov:
Images loading was non-blocking leading to incorrect progress report

Fixes #7395

-------------

I'm somehow confused about `return`'s all around, seems like there can be case when wait will not be executed, however worked fine for me, so I assume it's ok...

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -1757,6 +1757,9 @@ public class Geocache implements IWaypoint {
                 return;
             }
 
+            // Need to wait for images loading since HtmlImage.getDrawable is non-blocking here
+            imgGetter.waitForEndCompletable(null).blockingAwait();
+
             cache.setLists(lists);
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.DB));
 


Lineflyer:
I wonder if we should really wait for the image loading. The issue described in #7395 is perfectly explained with the lazy loading, thus not a fault anymore. Or do you see other reasons for this change to be meaningful?

okainov:
@Lineflyer as a user, when progress bar disappears (== "Done) I do expect that loading is really done. I'm sure I already had this issue many times when I save couple of hundreds caches, then I go somewhere without internet and near the GZ I find out that images were not loaded. So for me it is really an issue, when loading continues without any indicators. This for sure can be improved (see #330), but the current behavior is very synchronous, thus I expect proper completion.

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1753,6 +1753,9 @@ public static void storeCache(final Geocache origCache, final String geocode, fi
                 }
             }
 
+            // Need to wait for images loading since HtmlImage.getDrawable is non-blocking here
```
Bananeweizen: I believe it would be sufficient to wait before the final sendEmptyMessage(DONE). In case the handler has been disposed (the user cancelled), we can return without waiting for the images to download (instead we could try to cancel the imgGetter to before those returns).

I'm not sure if that makes much of a difference, however. If at all, probably only in batch mode.

Lineflyer:
> @Lineflyer as a user, when progress bar disappears (== "Done) I do expect that loading is really done. I'm sure I already had this issue many times when I save couple of hundreds caches, then I go somewhere without internet and near the GZ I find out that images were not loaded.

Indeed, you are perfectly right. I guess that was the very same reason where I experienced the underlying issue randomly. I only wonder how this influences the UI blocking time in practice (especially when some resources are not responding, which has AFAIR 30sec timeout in our code).

Looking into the usage of quite some users I know, they are e.g. batch updating 100 caches. As for gc.com we also download externally hosted images it might happen, that for - lets say - 5% of caches at least one resource might be unresponsive (often seen for discontinued visitor counters, etc,). This means 5x30 sec more loading time from users perspective.

Is there any reasonable low effort possibility to e.g. move at least such kind of "hanging downloads" to asynchronous as long as we do not have a complete background solution?

main/src/cgeo/geocaching/models/Geocache.java
okainov: Good point, I was not checking cancel case. And it seems to have bigger problems underneath. Even without my change I'm getting some exception in the console

``` W/cgeo: [OkHttp] Failure (2 ms) GET https://ssl.webpack.de/gcvote.com/getVotes.php?version=cgeo&userName=lega4&password=***&cacheIds=f565d35e-2443-4ce1-b17e-0830804a55e1 (java.io.IOException: Canceled)
 W/System.err: io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | java.io.IOException: Canceled
      at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
```
With it also similar
```
E/cgeo: [refresh--3] Geocache.storeCache
    java.lang.RuntimeException: java.lang.InterruptedException

...and a lot more lines
```

But again, seems to be a bigger issue in general. Also it looks like the way I handled exceptions in suApi is a bit wrong, but since there are no documented guidelines for this... Can we maybe have a chat about proper way to implement handling of cancellations?

As for this PR, I moved this line before `DataStore.saveCache(cache, EnumSet.of(SaveFlag.DB));` (because we need to save images before saving the cache), should be ok.

Issue #7436, opened 2019-03-30
"Sort Caches by Favorite [%]" gives weird result


MagpieFourtyTwo:
##### Detailed steps causing the problem:
- Open Live Map at any point (caches NearBy may end up with same result)
- wait until "Show as List" is active
   (BTW: Anything new concerning the busy spinner in Live Map? - I'm missing it *really* badly, esp. as newly loaded caches still have the tendency to vanish after half a second, so you have to stare at the map all the time, otherwise you miss the moment to realize that loading has finished)
- select "Show as List"
- sort by "Favorite [%]"

##### Actual behavior after performing these steps:
- the order of the list seems to be completely messed up (see attached screenshots)

##### Expected behavior after performing these steps:
- caches with 0 favos should not be listed above caches that already have favorite points (although number of logs is poss. not available at this very moment)
- when there are offline saved caches in the list, these should be listed first, as their favo % can be calculated based on (real) offline data (can not reproduce right now, but I'm pretty sure that I already had offline saved caches somewhere in the middle of the list and even beneath 0 favo caches)

##### Version of c:geo used:
- 2019.03.30-NB1-4c5da32

##### Is the problem reproducible for you?
- Yes

##### System information:
```
--- System information ---
Device: E6853 (E6853, Sony)
Android version: 7.1.1
Android build: 32.4.A.1.54
c:geo version: 2019.03.30-NB1-4c5da32
Google Play services: disabled - 15.0.90 (040408-231259764)
Low power mode: active
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: true
Map strategy: detailed
HW acceleration: enabled (default state)
System language: de_DE
System date format: dd.MM.yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (6,9 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (6,9 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (6,9 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (235,8 MB) on system internal storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
   extremcaching.com: Logged in (Anmeldung OK)
[...]

PR #7439, opened 2019-03-30
do not use 0 as initial fav. point value

okainov:
0 is a valid value for number of Favorite points, it should not be a special case
Also fix displaying of caches with unknown fav points as suggested in #7436

### DIFF
--- a/main/src/cgeo/geocaching/AbstractDialogFragment.java
+++ b/main/src/cgeo/geocaching/AbstractDialogFragment.java
@@ -238,7 +238,10 @@ public abstract class AbstractDialogFragment extends DialogFragment implements C
         }
 
         // favorite count
-        details.add(R.string.cache_favorite, res.getString(R.string.favorite_count, cache.getFavoritePoints()));
+        final int favCount = cache.getFavoritePoints();
+        if (favCount >= 0) {
+            details.add(R.string.cache_favorite, res.getString(R.string.favorite_count, favCount));
+        }
 
         // more details
         final View view = getView();
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1458,9 +1458,10 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
          */
         private void updateFavPointBox() {
             // Favorite counts
-            if (cache.getFavoritePoints() > 0) {
+            final int favCount = cache.getFavoritePoints();
+            if (favCount >= 0) {
                 favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(res.getString(R.string.favorite_count, cache.getFavoritePoints()));
+                favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
             } else {
                 favoriteLine.left.setVisibility(View.GONE);
             }
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -138,8 +138,10 @@ public class Geocache implements IWaypoint {
     private Boolean onWatchlist = null;
     private Boolean logOffline = null;
     private int watchlistCount = -1; // valid numbers are larger than -1
-    private int favoritePoints = 0;
+    private int favoritePoints = -1; // valid numbers are larger than -1
     private float rating = 0; // valid ratings are larger than zero
+    // FIXME: this makes no sense to favor this over the other. 0 should not be a special case here as it is
+    // in the range of acceptable values. This is probably the case at other places (rating etc.) too.
     private int votes = 0;
     private float myVote = 0; // valid ratings are larger than zero
     private int inventoryItems = 0;
@@ -301,9 +303,7 @@ public class Geocache implements IWaypoint {
         if (StringUtils.isBlank(getDescription())) {
             description = other.getDescription();
         }
-        // FIXME: this makes no sense to favor this over the other. 0 should not be a special case here as it is
-        // in the range of acceptable values. This is probably the case at other places (rating, votes, etc.) too.
-        if (favoritePoints == 0) {
+        if (favoritePoints == -1) {
             favoritePoints = other.favoritePoints;
         }
         if (rating == 0) {
--- a/main/src/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/cgeo/geocaching/ui/CacheListAdapter.java
@@ -489,7 +489,8 @@ public class CacheListAdapter extends ArrayAdapter<Geocache> {
             }
         }
 
-        holder.favorite.setText(Integer.toString(cache.getFavoritePoints()));
+        final int favCount = cache.getFavoritePoints();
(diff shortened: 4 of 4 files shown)


main/src/cgeo/geocaching/ui/CacheListAdapter.java
```diff
@@ -489,7 +489,8 @@ public void accept(final BitmapDrawable bitmapDrawable) {
             }
         }
 
-        holder.favorite.setText(Integer.toString(cache.getFavoritePoints()));
+        final int favCount = cache.getFavoritePoints();
+        holder.favorite.setText(favCount < 0 ? "?" : Integer.toString(favCount));
```
UniQP:
```suggestion
        holder.favorite.setText(favCount >= 0 ? Integer.toString(favCount) : "?");
```

I think the inverted condition is more readable.

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -301,9 +301,7 @@ public boolean gatherMissingFrom(final Geocache other) {
         if (StringUtils.isBlank(getDescription())) {
             description = other.getDescription();
         }
-        // FIXME: this makes no sense to favor this over the other. 0 should not be a special case here as it is
-        // in the range of acceptable values. This is probably the case at other places (rating, votes, etc.) too.
```
UniQP: I think the FIXME still holds for rating, etc. Please move it to the rating check below and adapt the text accordingly.

main/src/cgeo/geocaching/AbstractDialogFragment.java
```diff
@@ -227,7 +227,10 @@ protected final void addCacheDetails() {
         }
 
         // favorite count
-        details.add(R.string.cache_favorite, cache.getFavoritePoints() + "×");
+        final int favCount = cache.getFavoritePoints();
+        if (favCount >= 0) {
+            details.add(R.string.cache_favorite, favCount + "×");
```
UniQP: The `"×"` should be part of the resource string.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1471,9 +1471,10 @@ private void updateWatchlistBox() {
          */
         private void updateFavPointBox() {
             // Favorite counts
-            if (cache.getFavoritePoints() > 0) {
+            final int favCount = cache.getFavoritePoints();
+            if (favCount >= 0) {
                 favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(cache.getFavoritePoints() + "×");
+                favoriteLine.right.setText(favCount + "×");
```
UniQP: We should introduce (or reuse?) a resource string for the text.

main/src/cgeo/geocaching/ui/CacheListAdapter.java
UniQP: Lint warning: Number formatting does not take into account local settings. Consider using `String.format` instead.

okainov: Applied, and I'm not sure that we really care about number formatting here. Even if there are caches with more than 1k favs, I still find "1257" readable enough.

main/src/cgeo/geocaching/models/Geocache.java
okainov: done

main/src/cgeo/geocaching/AbstractDialogFragment.java
okainov: This is [intended] to be the only place where "x" is displayed. Since it doesn't have neither translation nor smart logic, I think we can keep it in this only place and should be fine.

main/src/cgeo/geocaching/CacheDetailActivity.java
okainov: "x" will be removed from here in #7441

okainov:
Any update?

main/src/cgeo/geocaching/CacheDetailActivity.java
UniQP: We should use the same style to show favorite counts at all places. Removing the `×`leads to an inconsistent style.

main/src/cgeo/geocaching/AbstractDialogFragment.java
UniQP: You can use `main/res/values/strings_not_translatable.xml` for strings that are not intended for translation.

main/src/cgeo/geocaching/CacheDetailActivity.java
okainov: Actually I was thinking about it, but now it's different style. For CachePopup appearance is not changing, I find just a number there looking a bit ugly.
For CacheDetail though it will have percentage all the time except zero. And zero without x looks fine together will other numbers IMO. And of course when there is percentage, no need in x any more

okainov: If you think that we can remove x from cachepopup as well - can do, just seemed not so nice from my point of view

main/src/cgeo/geocaching/AbstractDialogFragment.java
okainov: It's just one/two numbers. There is no any smartness in formatting. It was not in resources before for years. Maybe it can be added, but it does not belong to this change anyway.

main/src/cgeo/geocaching/CacheDetailActivity.java
UniQP: Don't get me wrong. I think, we should keep the `×`.
My point is that we should use the same style here and in `AbstractDialogFragment.java `.

main/src/cgeo/geocaching/AbstractDialogFragment.java
UniQP: Appending the `×` here and in `CacheDetailActivity.java` should be consistent but currently the code fragments are not connected. That's why introducing a resource improves the code quality.

The resource issue will come up again in #7441 anyway, so starting here would be helpful.
In general, it is always a good thing to clean up the code you are currently changing.

main/src/cgeo/geocaching/CacheDetailActivity.java
okainov: >For CacheDetail though it will have percentage all the time except zero. And zero without x looks fine together will other numbers IMO.

Lineflyer:
@UniQP Feel free to merge if you consider your review to be OK.  We just copied `master` to `release`so its the best time for new feature merging onto `master`.

main/src/cgeo/geocaching/CacheDetailActivity.java
UniQP: Adding percentages is another PR – please focus on this one.

Lineflyer:
Follow Up: 
Right now we are again possibly about to merge branches. To allow more time for testing here, I would like this only to be merged after the next release. Setting "Do not merge" until then.

Lineflyer:
And again:
@UniQP Feel free to merge if you consider your review to be OK. We just copied `master` to `release` so its the best time for new feature merging onto `master`.

UniQP:
@okainov Can you rebase your changes and resolve the conflicts?

okainov:
Done. Haven't tested it after rebasing though.

PR #7440, opened 2019-03-30
add progressdialog styling with border and non transparent background (fix #7434)

### LINKED ISSUE #7434: Cache refresh window has no borders anymore
Since the recent layout change of the cache refresh window (shown after you e.g. selected to refresh a cache) and other similar dialogs their are looking better in general, but they do no longer have any visual border.

This makes them look melted into the (also black) background. A visual seperation in form of a small greyish border would be appropriate.

This is only an optical issue.
2019.03.29-NB
Android 8.1

Can add a screenshot later, if the meaning is unclear.

moving-bits:
add progressdialog styling with border and non transparent background (fix #7434)

### DIFF
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -380,7 +380,7 @@
     <style name="cgeoProgressdialogTheme" parent="Theme.AppCompat.Dialog.Alert">
         <item name="android:textColorPrimary">?text_color</item>
         <item name="android:textColor">?text_color</item>
-        <item name="android:windowBackground">?background_color</item>
+        <item name="android:windowBackground">@drawable/border_straight</item>
         <item name="buttonBarButtonStyle">@style/button_borderless</item>
     </style>
 </resources>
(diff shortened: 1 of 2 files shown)


main/res/values/styles.xml
```diff
@@ -380,7 +380,7 @@
     <style name="cgeoProgressdialogTheme" parent="Theme.AppCompat.Dialog.Alert">
         <item name="android:textColorPrimary">?text_color</item>
         <item name="android:textColor">?text_color</item>
-        <item name="android:windowBackground">?background_color</item>
+        <item name="android:windowBackground">@drawable/border_straight</item>
```
Bananeweizen: How do you debug such stuff typically? I was following the parent declarations of Theme.AppCompat.Dialog.Alert until I was at item name="android:windowBackground">@drawable/abc_dialog_material_background</item and didn't find any reasonable way to continue there...

PR #7441, opened 2019-03-30
Add favorite percentage to cache details view

### LINKED ISSUE #4346: Show favorite point percentage in cache details
From support mail:
It would be nice to show the favorite point percentage (next to the absolute favorite points) on the cache details page. As it is already possible to sort by percentage the info should be there.
## 

As far as I know we use a self calculated value (all logs over favorite points) to calculate the sorting (?!).
We could/should instead parse the value shown on gc.com which is more accurate (i.e. only counting premium logs AFAIK).

okainov:
Fixes #4346 

Also remove "x" from the text on cache details page, and intention is to show fav points even if they're 0 (see #7439)

### DIFF
--- a/main/res/values/strings_not_translatable.xml
+++ b/main/res/values/strings_not_translatable.xml
@@ -85,7 +85,8 @@
 	</string>
 
     <!-- cache details -->
-    <string name="favorite_count" translatable="false">%1$d×</string>
+    <string name="favorite_count" translatable="false">%1$d</string>
+    <string name="favorite_count_percent" translatable="false">%1$d (%2$.0f%%)</string>
 
     <!-- cache menu -->
     <string name="caches_map_mapswithme" translatable="false">Maps.me</string>
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1461,11 +1461,16 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
             final int favCount = cache.getFavoritePoints();
             if (favCount >= 0) {
                 favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
+
+                final int findsCount = cache.getFindsCount();
+                if (findsCount > 0) {
+                    favoriteLine.right.setText(res.getString(R.string.favorite_count_percent, favCount, (float) (favCount * 100) / findsCount));
+                } else {
+                    favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
+                }
             } else {
                 favoriteLine.left.setVisibility(View.GONE);
             }
-
             final LinearLayout layout = ButterKnife.findById(view, R.id.favpoint_box);
             final boolean supportsFavoritePoints = cache.supportsFavoritePoints();
             layout.setVisibility(supportsFavoritePoints ? View.VISIBLE : View.GONE);
--- a/main/src/cgeo/geocaching/AbstractDialogFragment.java
+++ b/main/src/cgeo/geocaching/AbstractDialogFragment.java
@@ -240,7 +240,12 @@ public abstract class AbstractDialogFragment extends DialogFragment implements C
         // favorite count
         final int favCount = cache.getFavoritePoints();
         if (favCount >= 0) {
-            details.add(R.string.cache_favorite, res.getString(R.string.favorite_count, favCount));
+            final int findsCount = cache.getFindsCount();
+            if (findsCount > 0) {
+                details.add(R.string.cache_favorite, res.getString(R.string.favorite_count_percent, favCount, (float) (favCount * 100) / findsCount));
+            } else {
+                details.add(R.string.cache_favorite, res.getString(R.string.favorite_count, favCount));
+            }
         }
 
         // more details
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -1921,6 +1921,7 @@ public class Geocache implements IWaypoint {
 
     /**
      * Get number of overall finds for a cache, or 0 if the number of finds is not known.
+     * TODO: 0 should be a valid value, maybe need to return -1 if the number is not known
      */
     public int getFindsCount() {
         if (getLogCounts().isEmpty()) {


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1473,7 +1473,11 @@ private void updateFavPointBox() {
             // Favorite counts
             if (cache.getFavoritePoints() > 0) {
                 favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(cache.getFavoritePoints() + "×");
+                final StringBuilder favoriteString = new StringBuilder(String.valueOf(cache.getFavoritePoints()));
+                if (cache.getFavoritePoints() > 0 && cache.getFindsCount() > 0) {
+                    favoriteString.append(" (" + Math.round((float) cache.getFavoritePoints() / cache.getFindsCount() * 100) + "%)");
```
Bananeweizen: Using concatenation inside append() is an anti pattern. Wondering that checkstyle didn't trigger here. Change to append().append().append() please.

If you reorder the math as ((float)(favorites*100))/finds, then you do one operation as (faster) integer instead of everything as float.

Also you are repeatedly calling the same accessor methods. You should always store to a temporary local variable, if you need the same method call value more than once (since you never know what is behind that call).

okainov: I still cannot get used to this super ugly append().append() construction even though I know about immutable strings in Java :(

Done, thanks for comments.

Bananeweizen: In plain Java, if you have just _one_ statement containing String concatenation, it doesn't even make a difference, since all current compilers replace such String concatenation by StringBuilder.append() calls during compilation. :)

okainov:
>if (favCount > 0 is called twice.

Yes, I know, see #7439 where first condition is changed. I want to get both changes merged, so in the end it will be fine

>How about using String.format instead the .append(s)?

According to random internet articles and SO, `String.format` is muuuuuch slower than both concatenation and StrinGuilder

SammysHP:
What does this number mean? Usually the fav% is calculated only from logs by premium members.

okainov:
>What does this number mean? Usually the fav% is calculated only from logs by premium members.

For `Geocache` we know amount of fav points and amount of finds. So we divide these numbers to get fav percent. Yes, for some services it will not fully correspond to reality (i.e. if not all users are egible to put favorites), but `Geocache` is service-agnostic, so we just use the data that we have.

bekuno:
> > if (favCount > 0 is called twice.
> 
> Yes, I know, see #7439 where first condition is changed. I want to get both changes merged, so in the end it will be fine

I mean to use write in line 1479:
                if (findsCount > 0) {

> > How about using String.format instead the .append(s)?
> 
> According to random internet articles and SO, `String.format` is muuuuuch slower than both concatenation and StringBuilder

We make no 1000 rounds. 
I love it for readability on such places.
The .append solution is ok for me.

okainov:
> mean to use write in line 1479: if (findsCount > 0) {

I understand. But I don't really want to remove it here and then return this condition again during conflict resolving in #7439 (because the conditions will be different again there). And when I was developing, I did #7439 first, so this one was somehow "following"

UniQP:
If your PR is based on #7439, then let's wait until  #7439 is merged.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1471,9 +1471,17 @@ private void updateWatchlistBox() {
...
+            final int favCount = cache.getFavoritePoints();
+            if (favCount > 0) {
                 favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(cache.getFavoritePoints() + "×");
+                final StringBuilder favoriteString = new StringBuilder(String.valueOf(favCount));
+                final int findsCount = cache.getFindsCount();
+                if (favCount > 0 && findsCount > 0) {
+                    favoriteString.append(" (");
+                    favoriteString.append(Math.round((float) (favCount * 100) / findsCount));
+                    favoriteString.append("%)");
```
UniQP: Please introduce resource strings (two variants: with and without percentages).

okainov:
Rebased & rebased on #7439 to avoid unnecessary work

UniQP:
@okainov Looks like you have to rebase again.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1458,13 +1458,19 @@ private void updateWatchlistBox() {
...
                 favoriteLine.left.setVisibility(View.GONE);
-            }
+            } else {
+                favoriteLine.left.setVisibility(View.VISIBLE);
+                // Special case when cache has no fav points
+                favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
 
+                final int findsCount = cache.getFindsCount();
+                if (favCount > 0 && findsCount > 0) {
+                    favoriteLine.right.setText(res.getString(R.string.favorite_count_percent, favCount, (float) (favCount * 100) / findsCount));
```
UniQP: The format string expects an integer as last argument.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1458,13 +1458,19 @@ private void updateWatchlistBox() {
...
+            if (favCount < 0) {
                 favoriteLine.left.setVisibility(View.GONE);
-            }
+            } else {
+                favoriteLine.left.setVisibility(View.VISIBLE);
+                // Special case when cache has no fav points
+                favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
 
+                final int findsCount = cache.getFindsCount();
+                if (favCount > 0 && findsCount > 0) {
```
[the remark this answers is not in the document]
okainov: I don't find line "0 (0%)" more useful than "0" or "0x". It's even more confusing, that's why I'd prefer just "0" in this case.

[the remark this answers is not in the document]
UniQP: Well, "0 (0%)" says that the cache has been found and is probably not very good, whereas "0" indicates that the cache has not been found, which makes it interesting for me (as FTF hunter).

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1458,13 +1458,19 @@ private void updateWatchlistBox() {
          */
         private void updateFavPointBox() {
             // Favorite counts
-            if (cache.getFavoritePoints() > 0) {
-                favoriteLine.left.setVisibility(View.VISIBLE);
-                favoriteLine.right.setText(res.getString(R.string.favorite_count, cache.getFavoritePoints()));
-            } else {
+            final int favCount = cache.getFavoritePoints();
+            if (favCount < 0) {
```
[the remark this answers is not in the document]
okainov: done

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1458,13 +1458,19 @@ private void updateWatchlistBox() {
...
                 favoriteLine.left.setVisibility(View.GONE);
-            }
+            } else {
+                favoriteLine.left.setVisibility(View.VISIBLE);
+                // Special case when cache has no fav points
+                favoriteLine.right.setText(res.getString(R.string.favorite_count, favCount));
```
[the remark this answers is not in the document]
okainov: why? in "else" we just don't show fav part because it makes no sense.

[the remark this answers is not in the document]
UniQP: I'm talking about adding an else part to the if statement in line 1470.

main/src/cgeo/geocaching/CacheDetailActivity.java
okainov: updated format string in case someone will want to show digits after dot :)

okainov:
>Is it possible to also show the percentatges in AbstractDialogFragment.java?

In theory yes, but I was thinking that at AbstractDialogFragment point we may not have finds number (yet).

UniQP:
At least, we should have find numbers for caches that are already stored on the device.

okainov:
Probably something like this. Also removed "x" from the non-percent recourse since it's now only for zero.

Lineflyer:
@UniQP Ready to merge from your perspective? I do see pending requested changes according to Github but think they are already addressed, are they?

main/res/values/strings_not_translatable.xml
```diff
@@ -85,7 +85,8 @@
 	</string>
 
     <!-- cache details -->
-    <string name="favorite_count" translatable="false">%1$d×</string>
+    <string name="favorite_count" translatable="false">%1$d</string>
```
UniQP: > Probably something like this. Also removed "x" from the non-percent recourse since it's now only for zero.

The resource is also used in other places where the cache count is greater than 0.

main/res/values/strings_not_translatable.xml
```diff
@@ -85,7 +85,8 @@
...
+    <string name="favorite_count_percent" translatable="false">%1$d (%2$.0f\%)</string>
```
UniQP: I think we should add a '×' to make it consistent with the number without percentages.

UniQP:
@Lineflyer What's your opinion on showing percentages for caches with 0 favorites but > 0 founds?

Lineflyer:
> @Lineflyer What's your opinion on showing percentages for caches with 0 favorites but > 0 founds?

That sounds like a valid percentage which is 0%, which should be shown. There will be a lot of default micro caches which probably will have that prcentage forever ;(

Lineflyer:
@UniQP 

What would be your verdict after my last feedback. I did not check how the questioned situation is handled by this PR, thus I cannot judge if changes are required or this can be finally merged?!

UniQP:
@Lineflyer Your last feedback is included. However, as I stated in https://github.com/cgeo/cgeo/pull/7441#pullrequestreview-237446290, the PR is not working as intended. So it's not ready to merge yet.

main/res/values/strings_not_translatable.xml
okainov: This is consistent with number without "x"

main/res/values/strings_not_translatable.xml
okainov: I don't like this weird "x"

okainov:
Fixed escaping.

My code should show percentage on CachePopup if cache was stored before, however due to #7640 this doesn't work right not.

Lineflyer:
@UniQP 
>  However, as I stated in #7441 (review), the PR is not working as intended. So it's not ready to merge yet.

This is resolved, is it?

Lineflyer:
As there was no further feedback from @UniQP or others I consider this ready for testing in nightlies. In worst case it can be reverted.

Lineflyer:
@okainov Thanks for your contribution. I took a brief look on it today and the it seems to work properly.

PR #7446, opened 2019-03-31
lint: hard-coded package name

Bananeweizen:
All resources should use the auto resolution namespace nowadays.


Lineflyer:
retest this please

PR #7448, opened 2019-03-31
fix potential null accesses

Bananeweizen:


tests/src-android/cgeo/geocaching/connector/su/SuParserTest.java
```diff
@@ -265,7 +265,8 @@ public void testCanParseImageText() throws Exception {
         parseCache(cacheJson);
         boolean found = false;
         for (final Image img : cache.getImages()) {
-            if (img.title.contains("Ограда монастыря")) {
+            final String title = img.title;
+            if (title != null && title.contains("Ограда монастыря")) {
```
okainov: Have you got test failure here? This is test based on JSON above, all the images should contain title in this example.

Lineflyer:
retest this please

tests/src-android/cgeo/geocaching/connector/su/SuParserTest.java
Bananeweizen: No, this is only about the null annotations. The title field is `@Nullable`, therefore you see a compiler warning for the `.contains()` in both AS and Eclipse, since the compiler cannot prove that there is no null access. Unfortunately this is only indicated by some light yellow background color in AS, so it is easy to miss.
I'm aware there will be no error in practice, but I want to get rid of all warnings in the IDE.

PR #7455, opened 2019-04-02
Fix #7453: upgrade okhttp to 3.12.2

### LINKED ISSUE #7453: upgrade okhttp to 3.12.2
We could upgrade okhttp to 3.12.2 with our current target settings. However, it behaves differently in encoding query parameters then and ParametersTest.java fails.

@samueltardieu can you check that, since you added explicit encoding in the Parameters.toString() method? I suspect that can be removed with the new version, but I'm not sure.

samueltardieu:


samueltardieu:
I won't have time to check what went wrong with the builder. If someone wants to take over (@Bananeweizen?), please be my guest.

Lineflyer:
Retest this please

samueltardieu:
What version of Java is the CI using?

mucek4:
If you look in https://github.com/cgeo/cgeo-executor/blob/master/Dockerfile it's openjdk-8-jdk-headless for executors.

Lineflyer:
retest this please

PR #7458, opened 2019-04-03
remove AppInvite

Bananeweizen:
AppInvite has been deprecated. The replacement is not much more than sharing a link, therefore the normal "share a cache" functionality is sufficient.


Lineflyer:
@Bananeweizen Can you rebase this change. I assume it has now conflicts due to 50f0582cbde48bd6f7d9d392c922098d20c2f849 which I just merged from `release`

Issue #7466, opened 2019-04-05
NPE onResume in AbstractDialogFragment


Lineflyer:
This seems to be a remainder not covered by the recent fixes?!
@rsudev 

Version 2019.04.04
Android 8.0

```
java.lang.RuntimeException: 
 
  at android.app.ActivityThread.performResumeActivity (ActivityThread.java:3790)
 
  at android.app.ActivityThread.handleResumeActivity (ActivityThread.java:3830)
 
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3038)
 
  at android.app.ActivityThread.-wrap11 (Unknown Source)
 
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1696)
 
  at android.os.Handler.dispatchMessage (Handler.java:105)
 
  at android.os.Looper.loop (Looper.java:164)
 
  at android.app.ActivityThread.main (ActivityThread.java:6944)
 
  at java.lang.reflect.Method.invoke (Native Method)
 
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:327)
 
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1374)
Caused by: java.lang.NullPointerException: 
 
  at cgeo.geocaching.sensors.GeoDirHandler.start (GeoDirHandler.java:94)
 
  at cgeo.geocaching.sensors.GeoDirHandler.start (GeoDirHandler.java:77)
 
  at cgeo.geocaching.AbstractDialogFragment.onResume (AbstractDialogFragment.java:171)
 
  at android.support.v4.app.Fragment.performResume (Fragment.java:2126)
 
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1151)
 
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1290)
 
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1272)
 
  at android.support.v4.app.FragmentManagerImpl.dispatchResume (FragmentManager.java:2159)
 
  at android.support.v4.app.FragmentController.dispatchResume (FragmentController.java:223)
 
  at android.support.v4.app.FragmentActivity.onResumeFragments (FragmentActivity.java:507)
 
  at android.support.v4.app.FragmentActivity.onPostResume (FragmentActivity.java:496)
 
[...]

PR #7467, opened 2019-04-05
Fix NPE onResume in AbstractDialogFragment, #7466

rsudev:
Insert RestartLocationPermissionHandler
Re-assign callback id numbers

This is going to be tricky to test, as there is no direct venue into this code like calling the map from a shortcut. It is used e.g. in cache and waypoint popups


rsudev:
retest this please

PR #7469, opened 2019-04-06
add hint (+ toggle button for it) to cache popup (fix #7468)

### LINKED ISSUE #7468: Make cache popup more flexible
When tapping on a cache icon in map view, a popup window opens. It shows a couple of information and offers action buttons. Generally ok. But...

- It shows a lot of information I either don't need or which could be displayed in a more compact way.
- It misses some pretty important information while being on a caching tour: the hint. (*)

Simply adding the hint to this popup would make it even larger and therefore too big on some devices. So here a some ideas to condense this view a little bit:
- Cache type is already shown by the logo, thus could be left out. The size information could be printed after the distance.
- Difficulty, Terrain and favorites info could be printed in a condensed way after GC code.
- "last saved" info could be condensed into a one-liner
- list info could be left out
- hint could be added besides the the buttons (refresh / edit lists / delete)

I do understand that personal preferences on what infos to show in this popup vary widely. So I suggest to make it somewhat configurable, the defaults being set in a way that the current layout is shown.

---

(*) Why the hint should be available on the cache popup IMHO:

Currently my workflow while being on a longer tour is:
- tap on cache icon on map
- select cache from marker list
- swipe to details page
- swipe down to end of page
- tap on hint
- tap on back to return to map
- tap on cache icon on map again
- select cache from marker list again
- and follow distance info to final

This could be reduced to:
- tap on cache icon on map
- select cache from marker list
- tap on hint

moving-bits:
non intrusive way to show hint on cache popup:
- hint is not shown initially, but activated by pressing the new hint button
- hint button gets shown only if hint is available
- hint button is a toggle

### DIFF
--- a/main/src/cgeo/geocaching/CachePopupFragment.java
+++ b/main/src/cgeo/geocaching/CachePopupFragment.java
@@ -128,7 +128,7 @@ public class CachePopupFragment extends AbstractDialogFragment {
             addCacheDetails();
 
             // offline use
-            CacheDetailActivity.updateOfflineBox(view, cache, res, new RefreshCacheClickListener(), new DropCacheClickListener(), new StoreCacheClickListener(), null, new StoreCacheClickListener());
+            CacheDetailActivity.updateOfflineBox(view, cache, res, new RefreshCacheClickListener(), new DropCacheClickListener(), new StoreCacheClickListener(), new ShowHintClickListener(view), null, new StoreCacheClickListener());
 
             CacheDetailActivity.updateCacheLists(view, cache, res);
         } catch (final Exception e) {
@@ -200,7 +200,7 @@ public class CachePopupFragment extends AbstractDialogFragment {
                 DataStore.saveLists(Collections.singletonList(cache), listIds);
                 CacheDetailActivity.updateOfflineBox(getView(), cache, res,
                         new RefreshCacheClickListener(), new DropCacheClickListener(),
-                        new StoreCacheClickListener(), null, new StoreCacheClickListener());
+                        new StoreCacheClickListener(), new ShowHintClickListener(getView()), null, new StoreCacheClickListener());
                 CacheDetailActivity.updateCacheLists(getView(), cache, res);
             } else {
                 final StoreCacheHandler storeCacheHandler = new StoreCacheHandler(CachePopupFragment.this, R.string.cache_dialog_offline_save_message);
@@ -219,7 +219,7 @@ public class CachePopupFragment extends AbstractDialogFragment {
                         if (view != null) {
                             CacheDetailActivity.updateOfflineBox(view, cache, res,
                                     new RefreshCacheClickListener(), new DropCacheClickListener(),
-                                    new StoreCacheClickListener(), null, new StoreCacheClickListener());
+                                    new StoreCacheClickListener(), new ShowHintClickListener(view), null, new StoreCacheClickListener());
                             CacheDetailActivity.updateCacheLists(view, cache, res);
                         }
                     }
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/CachePopupFragment.java
```diff
@@ -261,6 +261,22 @@ public void onClick(final View arg0) {
         }
     }
 
+    private class ShowHintClickListener implements View.OnClickListener {
+        @Override
+        public void onClick(final View arg0) {
```
Bananeweizen: please avoid auto-generated names arg0, arg1, ...

main/src/cgeo/geocaching/CachePopupFragment.java
```diff
@@ -261,6 +261,22 @@ public void onClick(final View arg0) {
...
+            final View popupView = (View) arg0.getParent().getParent();
```
Bananeweizen: Such parent/child traversal code is dangerous, since every change of the layout XML files can break it, and we will not notice it since there is no compile error or similar. Therefore I'm in favor of always using explicit ID based lookups instead of getParent() or getChild() calls.
(I'm aware we have more code with this pattern, nevertheless we should probably at least not write new code using the pattern).

moving-bits: I would like to use ButterKnife etc. for doing an ID based search, but what can I use as a root anchor for it to start the search from?

main/src/cgeo/geocaching/CachePopupFragment.java
moving-bits: fixed with 4e7e23b

main/src/cgeo/geocaching/CachePopupFragment.java
moving-bits: fixed with 4e7e23b

moving-bits:
Retest this please

moving-bits:
@kumy: Our CI seems to have problems - no lint result files etc., see http://ci.cgeo.org/job/cgeo%20pull%20request/1857/
Can you take a look at it? Thanks

kumy:
Not at home right now. Isn t it because the build failed?

moving-bits:
The build itself seems to be ok. The first error message I can find is
`Error in testSearchByGeocodeSpecialties(cgeo.geocaching.CgeoApplicationTest): org.junit.ComparisonFailure: expected:<"[California, United States]"> but was:<"[]">`,
and somewhat later we have 
`Error in testNoHtmlInHomeLocation(cgeo.geocaching.connector.gc.GCLoginTest): org.junit.ComparisonFailure: expected:<[NO_ERROR]> but was:<[CONNECTION_FAILED]>`.

So maybe it's not the CI but timeout problems with gc.com website (which is terribly slow the last days). Sorry for the confusion.

Lineflyer:
Yes, indeed... As always on weekend evenings while a souvenir rage is ongoing, the site is under like 200% load it seems.

kumy:
Jenkins: retest this please

Lineflyer:
Tests show a new warning, which seems reasonable?!

moving-bits:
fixing the issue was the easy part, but the new button will not show up alongside the "more details" button (or anywhere in the upper part) :(

Lineflyer:
> fixing the issue was the easy part, but the new button will not show up alongside the "more details" button (or anywhere in the upper part) :(

How about a fallback then to just place it as an item in the popup menu (behind three-dot)?

moving-bits:
retest this please

Lineflyer:
@moving-bits 
Can you please post a screenshot showing the popup with these changes or a reference to one of those you put into the issue?

moving-bits:
If a cache has no hint (or hint is not loaded yet), nothing changes visually:

grafik


If a cache has a hint (and hint is loaded already), an additional button is shown beside a (somewhat shrinked) "more details" button:

grafik

If you press this button, the cache popup shows an additional part with the hint at the lower end:

grafik

This button is a toggle, you can switch showing the hint on and off, default is off.

The view is scrollable, so if the hint is extra long, you still see buttons etc. and the beginning of the hint, but can scroll down to see the rest of the hint.

Lineflyer:
That looks good to me implementation wise. Lets give it a try.

PR #7471, opened 2019-04-08
su: add find_date param during log posting

### LINKED ISSUE #7444: SU Logging ignores date
SuApi.postLog(Geocache, LogType, Calendar, String) does not use the given date of log creation. It should either be added to the http request or the parameter should be deleted. Since I don't know the API well enough, I cannot decide.

okainov:
Fixes #7444


Lineflyer:
@okainov This PR came in parallel (more or less) of merging master to release...as it extends the function of the new geocaching.su connector I suggest you target this to `release` branch so we can include it into the next release.

okainov:
@Lineflyer done

PR #7486, opened 2019-04-10
add create table cg_trail_history to onCreate (fix #7482)

### LINKED ISSUE #7482: c:geo crashes with "no such table: cg_trail_history" after fresh install
- fresh install of current (April 10) master brach on a new device
- run c:geo and open Map

- c:geo crashes with log message "no such table: cg_trail_history"

- show map

##### Version of c:geo used:
- today's (April 10) master

Yes

##### System information:
- tested under API 21 (Android 5.0) with Google API enabled

- table "cg_trail_history" got introduced by fix for #2958
- why is the database update routine not called before opening the map / what has to be done that it gets called before?

moving-bits:


main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -560,7 +560,7 @@ public void onCreate(final SQLiteDatabase db) {
             db.execSQL(dbCreateLogsOffline);
             db.execSQL(dbCreateTrackables);
             db.execSQL(dbCreateSearchDestinationHistory);
-
+            db.execSQL(dbCreateTrailHistory);
```
rsudev: Only a minor thing, but I would keep the blank line between the table creation statements and the index call.

Lineflyer:
retest this please

Bananeweizen:
This is an issue for further PRs: Please use the "fix #1234" notation already in the commit message, not only in the title of the pull request. Example from samuel: https://github.com/cgeo/cgeo/pull/7455/commits
If you do that, github automatically links back to the original issue (while it is hard to navigate to the issues of your PRs), and autoclosing may also not work for your PRs: https://help.github.com/en/articles/closing-issues-using-keywords

Lineflyer:
Just for completeness:
Autoclose also works if the "fix #1234" is contained in the PR name, however I agree with the better hyperlinking if it is referenced (also) in the commit message, so we should use it like that.

main/src/cgeo/geocaching/storage/DataStore.java
moving-bits: done

Issue #7289, opened 2019-02-14
Loading progress spinner missing on live map


Lineflyer:
No progress spinner shown on live map title bar any more.

More details when I am back home, but might be self explanatory.

rsudev:
It is clear. I observed it  as well when testing the marker clustering.
I saw it missing on the new mapsforge map. Google maps and old mapsforge map need to be confirmed.

Lineflyer:
Tested: It is missing on all map types.

moving-bits:
@rsudev Seems like progress spinners are deprecated in newer API levels (starting from 24), if I understood the following links correctly:
- https://stackoverflow.com/questions/27788195/setprogressbarindeterminatevisibilitytrue-not-working
- https://developer.android.com/reference/android/view/Window (see parts about several constants for progress spinner, e. g. FEATURE_INDETERMINATE_PROGRESS)

Lineflyer:
BTW: The progress spinner is also missing while the list of search results is being loaded..however in that case there is a spinner in the list window itself. Don't know if we need to address this issue...

okainov:
There should be for sure an indicator that caches loading is not finished. Whatever spinner or not, but it should be in place

moving-bits:
Looks like we need to replace the old actionBar (or AppCompat.actionBar) with the new Toolbar / AppCompat.Toolbar to regain a progress spinner, see here: https://www.vogella.com/tutorials/AndroidActionBar/article.html

Lineflyer:
> ooks like we need to replace the old actionBar (or AppCompat.actionBar) with the new Toolbar / AppCompat.Toolbar to regain a progress spinner, see here: https://www.vogella.com/tutorials/AndroidActionBar/article.html

That sounds like a bigger change required, right?

moving-bits:
> > Looks like we need to replace the old actionBar (or AppCompat.actionBar) with the new Toolbar / AppCompat.Toolbar to regain a progress spinner, see here: https://www.vogella.com/tutorials/AndroidActionBar/article.html
> 
> That sounds like a bigger change required, right?

To my understanding: Yes.

okainov:
Can you make it whatever indication of loading which is simpler to implement? I mean, really whatever - changing title, showing toast "loading completed", changing some color of some UI element. The problem is not a spinner, but indication. Spinner is nice, but if you're saying it requires deeper refactor - can we just get simple fix to eliminate the problem and properly refactor spinner later?

[...]

PR #7503, opened 2019-04-15
workaround for missing progress indicator on live map: show toast after loading is finished

moving-bits:
- does not solve #7289, is just a workaround to give some progress indication
- workaround can be removed after switching from ActionBar to AppCompat.Toolbar (refactoring)
- workaround is for live map only (NewMap and CGeoMap)


Lineflyer:
I would like to have this merged only after the next possible as IMHO it needs some testing (e.g. how it disturbs the usage flow, etc.). Therefore setting "Do not merge" until branches have been adjusted.

moving-bits:
superseded by #7639, therefore closing this one

Issue #7230, opened 2019-01-07
New live map version on gc.om


Lineflyer:
gc.com is trying out a new version of the live map.

I am not sure whether this is user bound or accessible for all.
The new map is on: 
https://www.geocaching.com/play/map

Not sure whether and when they might drop the old live map but it might be worth to check the difference in terms of parsing it.

Blog report about the new map:
https://www.saarfuchs.com/2018/11/die-neue-karte-bei-der-suche-auf-geocaching-com.html

moving-bits:
FYI: Looks like the first access to this map has to be coming through their homepage and conducting a search. In the results list you get the "Go to new map" button you have to click. From then on you can directly jump to https://www.geocaching.com/play/map, the site remembers you having selected the new map.

jonas-koeritz:
This is super interesting, GPX download seems to possible and searching seems to be easy, maybe a new GC connector can be based on that interface?

UniQP:
Today, Groundspeak announced that the old search map will be retired on April 25, 2019: https://forums.geocaching.com/GC/index.php?/topic/351791-pre-release-notes-website-existing-search-map-retirement-april-15-2019/

Lineflyer:
Oh well. This will most probably break our current live map :(

moving-bits:
I do not know how much effort ut is to adapt the parsing (I'm not deep into the current connector's parsing part), but maybe we should write to Bryan that we need more time than the remaining handful of days?

Lineflyer:
@moving-bits 
Thanks to @jonas-koeritz we might be one step closer to solve the problem. If you are interested, feel free to take a look at the related PR.

Lineflyer:
Reading the linked pre-release note in detail, we might have some time buffer:

> As a reminder, we are only removing the old Search map, which looks very similar to what we call the Browse map. [...] The Browse map is what you see today when you click “View Larger Map” on the cache details page, or when you select “Play” > “View Map” from the website header menu. (Those methods are changing. More on that in a minute.) The new Search map and the Browse map will still be available after the old Search map is retired.  

and

> Browse map (Will be accessible with a toggle from the new Search map as of April 25).

This sounds like the old map is still in background but only the defaults change.
Anyway it might be wise to change asap.

Lineflyer:
Implemented by #7514

PR #7514, opened 2019-04-16
Updated Live Map to use new GC.com /play/map

jonas-koeritz:
Referencing #7230, this is using the API used by the new map to load caches for the live map viewport. I'm not that comfortable with this project yet, so please review.

There are still some references to the tiles left in some functions calling my code, I left it intact to just be a drop-in replacement.

Please feel free to comment and give further advice, I'd like to help bring the c:geo project forward.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -137,11 +137,10 @@ public class GCMap {
             searchResult.setUrl(viewport.getCenter().format(Format.LAT_LON_DECMINUTE));
         }
 
-        final GCWebAPI.MapSearchResultSet mapSearchResultSet = GCWebAPI.searchMap(viewport).blockingGet();
+        final GCWebAPI.MapSearchResultSet mapSearchResultSet = GCWebAPI.searchMap(viewport);
         final List<Geocache> foundCaches = new ArrayList<Geocache>();
 
         if (mapSearchResultSet.results != null) {
-
             for (final GCWebAPI.MapSearchResult r : mapSearchResultSet.results) {
                 if (r.postedCoordinates != null) {
                     final Geocache c = new Geocache();
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -167,7 +167,7 @@ public class GCMap {
         return searchResult;
     }
 
-    private static CacheSize containerTypeToCacheSize(int containerType) {
+    private static CacheSize containerTypeToCacheSize(final int containerType) {
         switch (containerType) {
             case 2:
                 return CacheSize.MICRO;
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -153,7 +153,7 @@ public class GCMap {
                     c.setPremiumMembersOnly(r.premiumOnly);
                     c.setFound(r.userFound);
                     c.setDisabled(r.cacheStatus == 1);
-                    if(r.owner != null) {
+                    if (r.owner != null) {
                         c.setOwnerDisplayName(r.owner.username);
                         c.setOwnerUserId(r.owner.username);
                     }
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -7,13 +7,11 @@ import cgeo.geocaching.enumerations.StatusCode;
 import cgeo.geocaching.files.ParserException;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.GeopointFormatter.Format;
-import cgeo.geocaching.location.Units;
 import cgeo.geocaching.location.Viewport;
 import cgeo.geocaching.maps.LivemapStrategy;
 import cgeo.geocaching.maps.LivemapStrategy.Flag;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.network.Parameters;
-import cgeo.geocaching.sensors.Sensors;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.Formatter;
@@ -30,12 +28,9 @@ import java.io.IOException;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -382,13 +377,13 @@ public class GCMap {
         }
 
         GCWebAPI.MapSearchResultSet mapSearchResultSet = GCWebAPI.searchMap(viewport).blockingGet();
-        List<Geocache> foundCaches = new ArrayList<Geocache>();
+        final List<Geocache> foundCaches = new ArrayList<Geocache>();
 
         if (mapSearchResultSet.results != null) {
 
             for (GCWebAPI.MapSearchResult r : mapSearchResultSet.results) {
                 if (r.postedCoordinates != null) {
-                    Geocache c = new Geocache();
+                    final Geocache c = new Geocache();
                     c.setDetailed(false);
                     c.setReliableLatLon(true);
                     c.setGeocode(r.code);
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -30,9 +30,12 @@ import java.io.IOException;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -238,17 +241,10 @@ public class GCMap {
      */
     @NonNull
     public static SearchResult searchByViewport(@NonNull final Viewport viewport, @Nullable final MapTokens tokens) {
-        final int speed = (int) Sensors.getInstance().currentGeo().getSpeed() * 60 * 60 / 1000; // in km/h
-        LivemapStrategy strategy = Settings.getLiveMapStrategy();
-        if (strategy == LivemapStrategy.AUTO) {
-            strategy = speed >= 30 ? LivemapStrategy.FAST : LivemapStrategy.DETAILED;
-        }
-
-        final SearchResult result = searchByViewport(viewport, tokens, strategy);
+        final SearchResult result = searchPlayMapByViewport(viewport);
 
         if (Settings.isDebug()) {
-            final StringBuilder text = new StringBuilder(Formatter.SEPARATOR).append(strategy.getL10n()).append(Formatter.SEPARATOR).append(Units.getSpeed(speed));
-            result.setUrl(result.getUrl() + text);
+            result.setUrl(result.getUrl());
         }
 
         Log.d(String.format(Locale.getDefault(), "GCMap: returning %d caches from search", result.getCount()));
@@ -366,6 +362,51 @@ public class GCMap {
 
         return searchResult;
     }
+
+    /**
+     * Searches the view port on the live map for caches.
+     * The strategy dictates if only live map information is used or if an additional
+     * searchByCoordinates query is issued.
+     *
+     * @param viewport
+     *            Area to search
+     */
+    @NonNull
+    private static SearchResult searchPlayMapByViewport(@NonNull final Viewport viewport) {
+        Log.d("GCMap.searchPlayMapByViewport" + viewport.toString());
+
+        final SearchResult searchResult = new SearchResult();
+
+        if (Settings.isDebug()) {
+            searchResult.setUrl(viewport.getCenter().format(Format.LAT_LON_DECMINUTE));
+        }
+
+        GCWebAPI.MapSearchResultSet mapSearchResultSet = GCWebAPI.searchMap(viewport).blockingGet();
+        List<Geocache> foundCaches = new ArrayList<Geocache>();
+
+        if (mapSearchResultSet.results != null) {
+
+            for (GCWebAPI.MapSearchResult r : mapSearchResultSet.results) {
+                if (r.postedCoordinates != null) {
+                    Geocache c = new Geocache();
+                    c.setDetailed(false);
+                    c.setReliableLatLon(true);
+                    c.setGeocode(r.code);
+                    c.setName(r.name);
+                    c.setCoords(new Geopoint(r.postedCoordinates.latitude, r.postedCoordinates.longitude));
+                    c.setType(CacheType.getByWaypointType(Integer.toString(r.geocacheType)));
+                    c.setPremiumMembersOnly(r.premiumOnly);
+                    c.setFound(r.userFound);
+                    foundCaches.add(c);
+                }
+            }
+        }
+
+        searchResult.addAndPutInCache(foundCaches);
+
+        return searchResult;
+    }
+
 
     /**
      * Creates a list of caches types to filter on the live map (exclusion string)
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -14,16 +14,12 @@ import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.JsonUtils;
 import cgeo.geocaching.utils.Log;
 
-import android.graphics.Bitmap;
-import android.graphics.Bitmap.Config;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import java.io.IOException;
 import java.text.ParseException;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -155,8 +155,12 @@ public class GCMap {
                     c.setName(r.name);
                     c.setCoords(new Geopoint(r.postedCoordinates.latitude, r.postedCoordinates.longitude));
                     c.setType(cacheType);
+                    c.setDifficulty(r.difficulty);
+                    c.setTerrain(r.terrain);
+                    c.setSize(containerTypeToCacheSize(r.containerType));
                     c.setPremiumMembersOnly(r.premiumOnly);
                     c.setFound(r.userFound);
+                    c.setFavoritePoints(r.favoritePoints);
                     c.setDisabled(r.cacheStatus == 1);
                     if (r.owner != null) {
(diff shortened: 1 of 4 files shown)


jonas-koeritz:
So far it seems to be much more performant and the coordinates seem to be reliable again.

Lineflyer:
ok to test

jonas-koeritz:
Should have run the tests...

Lineflyer:
@jonas-koeritz 
First of all, thanks for this contribution. Yesterday evening I had no clue to to proceed in the current situation. Highly appreciated!!

I have seen you already addressed the codays issues.
I triggered our CI to run the tests, and I think there are only some related test need to be adapted (or they show open issues, not sure of it).

I am sadly not the one able to judge the change/implementation itself, but I hope @rsudev can take a look. I assigned him for a review.

jonas-koeritz:
Thanks for your response, the tests are failing unfortunately. I've done some investigation already:

- For the second test (Not logged in) it's totally clear: The map is not available when not logged in.
- For the first test, I can't see why this would not work. The response should be correct and when I open the Live Map on my device I can got to Velbert and its showing GC3FJ5F without a problem at the correct location. If I replay the API request on my PC it's working correctly as well.

So far the loading speed of caches on the map seems to have improved quite a bit. Limiting the request to 500 caches at a time (like the web version does) doesn't lead to strange behaviour as far as I can tell.

PS: The option to choose a "Strategy" might now be obsolete, as all caches in the viewport can be loaded at once with all necessary details.

Lineflyer:
> For the second test (Not logged in) it's totally clear: The map is not available when not logged in.

Yes, IMHO we can drop this test if the new map does not deliver data when not logged in. . I don't see the necessity to validate we are getting data when not logged in anyway. I don't know the history, why it was introduced.

> For the first test, I can't see why this would not work. The response should be correct and when I open the Live Map on my device I can got to Velbert and its showing GC3FJ5F without a problem at the correct location. If I replay the API request on my PC it's working correctly as well.

I am not so familiar with these unit tests. Could it be, that the test does not really run but already has a problem to check/set the prerequisites (e.g. strategy)?

> PS: The option to choose a "Strategy" might now be obsolete, as all caches in the viewport can be loaded at once with all necessary details.

Indeed, this can then be removed in a separate step. We can open a dedicated issue for that, once the PR is implemented.

For my understanding:
When you speak of all necessary details, this also include defined coordinates (no longer estimated graphically) and cache type information (no longer parsed by color)? If this is the case, we can even remove some more functions (unreliable marker).

main/src/cgeo/geocaching/connector/gc/GCMap.java
```diff
@@ -367,6 +358,56 @@ public String apply(final Bitmap bitmap, final String data) {
         return searchResult;
     }
 
+    /**
+     * Searches the view port on the live map for caches.
+     * The strategy dictates if only live map information is used or if an additional
+     * searchByCoordinates query is issued.
+     *
```
Lineflyer: I think the commented is C&P as "strategy" is no longer used in your code AFAICS.

jonas-koeritz: Good catch, I'm gonna do a bit more cleanup.

jonas-koeritz:
> When you speak of all necessary details, this also include defined coordinates (no longer estimated graphically) and cache type information (no longer parsed by color)? If this is the case, we can even remove some more functions (unreliable marker).
Yes, I'm pretty sure that all information provided by the API is accurate, currently we can get the following fields:

- Geocode
- Name
- Coordinates
- Cache Type
- Premium Only
- Found By User
- Disabled (Possibly Archived too, I would need an example of an archived cache for this)
- Owner

For Reference: [GCMap.java:387](https://github.com/cgeo/cgeo/blob/99a11996ab1ffdb7e600ebfe3e525cddf361cec6/main/src/cgeo/geocaching/connector/gc/GCMap.java#L387)

Lineflyer:
> Disabled (Possibly Archived too, I would need an example of an archived cache for this)

If Groundspeak stays with its strategy, as what is shown on the map, then archived should not show up.

But the other items are just a perfect (non-detailed) geocache. Nice.

Lineflyer:
@jonas-koeritz 
Feel free to push new commits to this PR if you got some clues regarding the tests.
The CI should automatically build it here to give you a second result.

Lineflyer:
I just used the cgeo-debug.apk from our CI to test this PR.
Also for me the live map returns 0 caches. I am logged in and can see the requests going out to the WebAPI with no results returned.

@jonas-koeritz 
Did you test this code in the runing app?

Lineflyer:
Debug log (coordinates removed):

```
17:37:15.417 Debug cgeo 12679  [network--5] GCMap.searchPlayMapByViewport(N 51° xx,xxx' · E 006° xx,xxx',N 51° xx,xxx' · E 006° xx,xxx')
17:37:15.417 Debug cgeo 12679  [OkHttp] GET https://www.geocaching.com/api/proxy/web/search?box=51.xxxxxx%2C6.xxxxxx%2C51.xxxxxx%2C6.xxxxxx9&origin=51.xxxxxx%2C6.xxxxxx&take=500&skip=0&asc=true&sort=distance
17:37:16.934 Debug cgeo 12679  [OkHttp] 200 (1544 ms) GET https://www.geocaching.com/api/proxy/web/search?box=51.xxxxxx%2C6.xxxxxx%2C51.xxxxxx%2C6.xxxxxx&origin=51.xxxxxx%2C6.xxxxxx&take=500&skip=0&asc=true&sort=distance (http/1.1)
17:37:16.936 Debug cgeo 12679  [network--5] GCMap: returning 0 caches from search
17:37:16.936 Debug cgeo 12679  [pool-5-thread-1] Filtering out 0 caches: []
```

jonas-koeritz:
Yes I've been testing this using the app on my phone. It's currently working.

I also have been able to reproduce this (no caches loaded at all). This should be the root cause of the tests failing. Is it possible to dump the response contents and debug the JSON deserialization? I just created a version of `getAPI` that accepts parameters to load the data but I don't quite understand how thats working internally.

Lineflyer:
@pstorch You have a bit experience with the WebAPI as such and can probably answer the question of @jonas-koeritz ?!

jonas-koeritz:
> Debug log (coordinates removed):
> 
> ```
> 17:37:15.417 Debug cgeo 12679  [network--5] GCMap.searchPlayMapByViewport(N 51° xx,xxx' · E 006° xx,xxx',N 51° xx,xxx' · E 006° xx,xxx')
> 17:37:15.417 Debug cgeo 12679  [OkHttp] GET https://www.geocaching.com/api/proxy/web/search?box=51.xxxxxx%2C6.xxxxxx%2C51.xxxxxx%2C6.xxxxxx9&origin=51.xxxxxx%2C6.xxxxxx&take=500&skip=0&asc=true&sort=distance
> 17:37:16.934 Debug cgeo 12679  [OkHttp] 200 (1544 ms) GET https://www.geocaching.com/api/proxy/web/search?box=51.xxxxxx%2C6.xxxxxx%2C51.xxxxxx%2C6.xxxxxx&origin=51.xxxxxx%2C6.xxxxxx&take=500&skip=0&asc=true&sort=distance (http/1.1)
> 17:37:16.936 Debug cgeo 12679  [network--5] GCMap: returning 0 caches from search
> 17:37:16.936 Debug cgeo 12679  [pool-5-thread-1] Filtering out 0 caches: []
> ```

Thats exactly the behaviour thats happening during the test (and rarely on my phone). I'm not sure if my code is the problem here.
Can somebody verify my logic on generating the `box` parameter implemented here: [GCWebAPI.java#L350](https://github.com/cgeo/cgeo/blob/99a11996ab1ffdb7e600ebfe3e525cddf361cec6/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java#L350)? I think this might be the, or part of the problem.

jonas-koeritz:
The problem seems to be the deserialization of the `GCWebAPI.MapSearchResultSet` class. I don't think it is as easy as it seems to deserialize into an array of objects, the problem might be only surfacing when there is only one array item? Would be nice if somebody who beter understands Jackson deserialization could have a look.

pstorch:
@jonas-koeritz can you try to debug the `getResponseData` here? https://github.com/cgeo/cgeo/blob/99a11996ab1ffdb7e600ebfe3e525cddf361cec6/main/src/cgeo/geocaching/network/Network.java#L422

Or the `js` parameter in the `apply()` method.

jonas-koeritz:
The `js` parameter contains the JSON data as expected but the call to apply seems to be the problem. The JSON doesn't get parsed into the MapResultSet Object.

pstorch:
It's the first time there is a nested Array in a Json response.
Maybe this helps: https://stackoverflow.com/questions/27895376/deserialize-nested-array-as-arraylist-with-jackson

Lineflyer:
@jonas-koeritz 
I wonder why it working for you sometimes. Maybe depending on the amount of caches returned?

jonas-koeritz:
I think I finally got it by annotating way more JSON properties. I'm currently running the tests on my machine to see if it works reliably.

jonas-koeritz:
The failing test seems to be incorrect, the cache it is looking for is not inside the specified viewport (or am I missing something?):
```
final Viewport viewport = new Viewport(new Geopoint("N 52° 25.369 E 9° 35.499"), new Geopoint("N 52° 25.600 E 9° 36.200"));
```

The Cache is located at: N 52° 25.363 E 009° 36.236.

Lineflyer:
Looking at the two test cases starting at https://github.com/cgeo/cgeo/blob/1fcabfb076b5b0ccc15992005c12d7013cc53849/tests/src-android/cgeo/geocaching/connector/gc/GCConnectorTest.java#L36
I assume the first one should be the negative test (test, that the cache is not contained) but was wrongly implemented. It might be, that due to the limit to zoom level steps on the old map the viewport was not 100% matched with the results producing a false passed test here.

Lineflyer:
@jonas-koeritz 
Feel free to change that test to check, that the cache is not contained if you agree to my assumption.
The second test there however should return the cache.

tests/src-android/cgeo/geocaching/connector/gc/GCConnectorTest.java
```diff
@@ -36,14 +36,14 @@ public static void testGetViewport() {
                 final Viewport viewport = new Viewport(new Geopoint("N 52° 25.369 E 9° 35.499"), new Geopoint("N 52° 25.600 E 9° 36.200"));
                 final SearchResult searchResult = ConnectorFactory.searchByViewport(viewport, tokens);
                 assertThat(searchResult).isNotNull();
-                assertThat(searchResult.isEmpty()).isFalse();
```
Lineflyer: I assume the `isEmpty().isFalse()` was intentional here, to check the the result contains caches (general function) but not contains the mentioned geocode. Otherwise it would pass if the result is empty (false positive result). I would therefore leave it in this testcase rather than moving it to the positive case, where the check for the mentioned cache is enough to verify that the result contains caches.

tests/src-android/cgeo/geocaching/connector/gc/GCConnectorTest.java
```diff
@@ -36,14 +36,14 @@ public static void testGetViewport() {
...
-                assertThat(searchResult.getGeocodes()).contains("GC4ER5H");
+                assertThat(searchResult.getGeocodes()).doesNotContain("GC4ER5H");
             }
 
             {
                 final Viewport viewport = new Viewport(new Geopoint("N 52° 24.000 E 9° 34.500"), new Geopoint("N 52° 26.000 E 9° 38.500"));
                 final SearchResult searchResult = ConnectorFactory.searchByViewport(viewport, tokens);
                 assertThat(searchResult).isNotNull();
+                assertThat(searchResult.isEmpty()).isFalse();
```
Lineflyer: See comment above. This is not explicitly needed here but on the first test IMHO.

Lineflyer:
@jonas-koeritz 
Finally green. Gratulation :)

jonas-koeritz:
> @jonas-koeritz
> Finally green. Gratulation :)

I've been staring at the command line output of jenkins for the whole seven and a half minutes ;)

Lineflyer:
Function wise it works really good and fast on my first trial. Will continue to try it out tonight.

Some instant finding of one regression though:
- The global cache type filter (in main menu) is not applied to the live map (e.g. set to "Mystery Cache" but live map keeps showing all types")

jonas-koeritz:
@Lineflyer yeah I think I forgot the filters...will have a look soon.

Lineflyer:
@jonas-koeritz 
I think there is also one other leftover of the old map communication AFAICS:
```
15:46:08.497 Debug cgeo 23862  [OkHttp] GET https://tiles.geocaching.com/map.details?i=GC64NJR&_=1555508769255&app=cgeo
15:46:10.749 Debug cgeo 23862  [OkHttp] 200 (2208 ms) GET https://tiles.geocaching.com/map.details?i=GC64NJR&_=1555508769255&app=cgeo (http/1.1)
```

This is triggered when you click on a cache on the map to open its popup. The purpose is to load the information which we do not have (which is IMHO Difficulty/Terrain and also formerly used to confirm the cache type of caches parsed unreliable).

I would suggest one of the following:
- If the D/T is available with the same WebAPI request you implemented, we should also request this during map loading.
- If D/T is not available we could also decide to drop this for now...i.e. the popup will no longer contain D/T but the user needs to open(=download) the cache details to see it.
-

Lineflyer:
> @Lineflyer yeah I think I forgot the filters...will have a look soon.

Depending on your time and the effort needed we can also handle this in a dedicated issue/PR after we got this merged. Up to you.

jonas-koeritz:
> > @Lineflyer yeah I think I forgot the filters...will have a look soon.
> 
> Depending on your time and the effort needed we can also handle this in a dedicated issue/PR after we got this merged. Up to you.

Everything should be implemented now.

Jenkins seems to be acting up from time to time and has test runners crashing, the tests all show green when done on my machine and nothing major changed between the last successful build and the current versions.

Lineflyer:
> Everything should be implemented now.

Wow, thanks very much so far.
The team will take a look at the code as time permits and I will for sure give it a test run.

> Jenkins seems to be acting up from time to time and has test runners crashing, the tests all show green when done on my machine and nothing major changed between the last successful build and the current versions.

From the last build the tests on Jenkins are fine AFAICS, but only CheckStyle claims some missing final declarations, which look reasonable:
http://ci.cgeo.org/job/cgeo%20pull%20request/1911/checkstyleResult/new/

rsudev:
I took a first short look and don't see anything up-front, will try to get a more detailed overview tomorrow.
A question for my understanding: Did I see it correctly, that the cache-type filter is applied to the result? So even the web is no longer applying type filters to the query? And did I miss the own/disabled filter?

Also from my side already many thanks for your contribution!

jonas-koeritz:
> Did I see it correctly, that the cache-type filter is applied to the result? So even the web is no longer applying type filters to the query?

I currently do not know how to apply the cache type filter at request time, I don't have a premium account, so I can't record the API requests beeing made to filter the caches. I suppose it is possible to filter at request time but I don't know yet how its done.

> And did I miss the own/disabled filter?

The own/disabled filter works with the current code because I already set the owner and disabled state when loading the cache. Checking and unchecking the options works flawlessly on my phone.

I will have a look at the codacy suggestions tomorrow, does anybody know whats up with jenkins? The tests don't seem to be failing (there has also been a successful run...) but it's currently showing up as `The unit test execution crashed (i.e. outside of the tests).`.

Lineflyer:
> I will have a look at the codacy suggestions tomorrow, does anybody know whats up with jenkins? The tests don't seem to be failing (there has also been a successful run...) but it's currently showing up as The unit test execution crashed (i.e. outside of the tests)..

That seems not to be the case...maybe a default message...as the build failed because of Checkstyle not within the tests. From the summary page it states clearly that Checkstyle made it fail.

Lineflyer:
> Did I see it correctly, that the cache-type filter is applied to the result? So even the web is no longer applying type filters to the query? 

To me, this sounds like a reasonable approach which has less dependency to the website functions. But there might be other reasons to limit the data at the server rather than the client...

jonas-koeritz:
> From the summary page it states clearly that Checkstyle made it fail.

I'm not that proficient in Jenkins, must have missed that. Fixing the few Checkstyle hints shouldn't be a big deal.

Lineflyer:
@rsudev 
Regarding your comments on IRC for the "mightly" build: 
I have seen the job on the CI, but actually I am not sure what to put into the parameter git-commit-id in order to get a mightly for this specific PR? Is this even possible before merging the change?

moving-bits:
@jonas-koeritz Regarding checkstyle: If you're using Android Studio you can add a Checkstyle plugin to the IDE which finds most (but not all) of the checkstyle issues the server highlights during build procedure. Saved me tons of round trips already :-)
See here: https://plugins.jetbrains.com/plugin/1065-checkstyle-idea

jonas-koeritz:
@moving-bits Thank you, I will have a look at that! This will help a lot. I've recently been doing more Go than Android programming so I'm a bit outdated on that.

jonas-koeritz:
There seems to be a problem when logging a cache. It's not marked as found on the map immediately (it takes up to a few minutes). Maybe the API lags behind on that and I have to find a way to not overwrite it when it's already marked as found locally.

Lineflyer:
Logging it in cgeo or on the website?
When its logged with cgeo it should at least be saved into the cache cache or even autosaved on device. Not sure at the moment.

But you did not change any of that data merging algorithms AFAIR. So if there is such merge problem, it should IMHO not be caused by the map implementation change.

jonas-koeritz:
I logged it in c:geo. The new map implementation does set the found property of the cache, could this be a problem? I just observed this once...maybe it was my fault/user error...

Lineflyer:
@rsudev 

I remember this and have seen the tag in the old map requests.
From my understanding we cant use it here as we cannot just append it somewhere. I would drop it for now unless they actively ask for having it.

Lineflyer:
Let me merge this and try it out on  nightlies. After easter days I will create issues for remaining tasks and bugs.

@jonas-koeritz 
Thanks for your contribution and I hope to see you around here also in future  not only for the remaining tasks of this one but maybe also for more new ideas and contributions :)

jonas-koeritz:
If necessary we could append the parameter to the requests, unknown parameters seem to be ignored by the API.
Filtering by found status seems to be unavailable (at least for basic members), the filtering options that are accessible are exactly the ones that can be used through the web GUI.

jonas-koeritz:
@Lineflyer thanks for merging. I will definitely stick around, this project has been really welcoming and friendly to me as a contributor!

rsudev:
> I logged it in c:geo. The new map implementation does set the found property of the cache, could this be a problem? I just observed this once...maybe it was my fault/user error...

If this property is set it will prevail when existing information is merged in (see Geocache.java:221)

rsudev:
> Filtering by found status seems to be unavailable (at least for basic members), the filtering options that are accessible are exactly the ones that can be used through the web GUI.

Allright, that was also true for the old map (therefore it had query-flags and filtering). For 'power-users' could that make a big difference in traffic in their homezone. If I interpret you remark correctly, you are basic member? Than I could try to find the necessary parameter.

rsudev:
Seems to be similar to the previous map:
 hide my finds: &hf=1
 hide my owned: &ho=1

jonas-koeritz:
I can definitely implement the filtering in the next days pretty easily.

If I correctly understand this, the found status should mostly be more recent one the device. Would it be better to not set the found info in the Map loader or to change the merging code?

Lineflyer:
Could you please open a deidcated issue for the found caches handling? Otherwise the discussion might get lost in this PR comments.

mucek4:
> I remember this and have seen the tag in the old map requests.
> From my understanding we cant use it here as we cannot just append it somewhere. I would drop it for now unless they actively ask for having it.

I would step towards them and ask what kind of tag, if any, then would like to have. Let's show some fair play from our side.

jonas-koeritz:
@mucek4 I already added an additional parameter (app=cgeo) in my latest PR (#7532).

PR #7519, opened 2019-04-18
Draw line between two consecutive trail points only up to max distance (fix #7506)

### LINKED ISSUE #7506: Do not draw trail line between consecutive points if distance is high
With todays NB I tried to check out the recent changes to the map trail by @moving-bits .
I noticed, that the history is kept also between several sessions (i.e. close of app and restart of app. I know, that it is intended to work like this, but in normal daily use this has shown some problems for me.

Example:
- Drive to location A, go caching and use c:geo trail function
- Drive to location B, go caching and use c:geo trail function
- On the next day start c:geo at location C
- Know I have a bunch of trail worms on my map connected with a straight line (last and first position after app restart) which I need to clear manually using the "Clear trail" function.

My proposal would be to keep it running while the app is in use but not between "sessions" or autoclear it after 24 hours or alike.

moving-bits:
while drawing historytrail: checks distance of two consecutive points and draws line between them only, if their distance is less than 10 km (stored history itself is untouched)

### DIFF
--- a/main/src/cgeo/geocaching/maps/PositionDrawer.java
+++ b/main/src/cgeo/geocaching/maps/PositionDrawer.java
@@ -22,6 +22,11 @@ import java.util.ArrayList;
 
 public class PositionDrawer {
 
+    /**
+     * maximum distance (in meters) up to which two points in the trail get connected by a drawn line
+     */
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

PR #7522, opened 2019-04-19
first installation: "Getting Started" page, deactivate geocaching.com (fix #7461)

### LINKED ISSUE #7461: Don't activate gc.com connector on install
As mentioned in #4263 we need to improve the "first start flow" anyhow, so this issue is more or less a workaround until then:

Since last release c:geo does not directly jump to the gc.com credentials settings anymore (due to the permission request handling now implemented) but starts up on the main screen.
I already got quite some support mail from first time users (or users reinstalled the app) which have not been aware, that something essential is missing to use c:geo.

The problem here is, that the gc.com connector is still by default set to active even though no credentials are stored. That leads to the "problem", that users can use live map, nearby search, etc. delivering them some unreliable data so they potentially do not notice they are working without being logged in.

If we could - as a workaround - simply disable the gc.com connector at first start it would at least be very visible to the user, that he needs to do something to get it working; i.e. if no connector is active click on "nearby" will return "None of the active services is supporting this function", live map will not show any unreliable caches, etc.

bekuno:
Created a "Getting Started" page to display on first installation, deactivated geocaching.com on start.

fix #7461

I have created a "Getting Started" page that is displayed to the user the first time they start after installation. An initial text is defined. Of course, this is a first idea now. I think that there may be additional links in other languages, e.g. in German language to https://www.cachewiki.de/wiki/Hauptseite.

The page is presented before the permissions questions (on map access). 

@Lineflyer:
The text is contained in the strings.xml file. Can/should it be separated into an additional file, would something need to be configured for crowdin translation?
Please test it with different screen sizes and android versions. Feel free to change the text.
By the way, the documentation for the first steps in manual.cgeo.org is out of date.

@okainov: 
Is this behavior fine for users who only use geocaching.ru?
I think that in the Russian translation a link to this page would be fine.

### DIFF
--- a/main/res/layout/about_starting_page.xml
+++ b/main/res/layout/about_starting_page.xml
@@ -20,22 +20,92 @@
             android:layout_margin="7dip"
             android:paddingLeft="3dip"
             android:text="@string/about_starting_head"
-            android:textColor="?text_color"
+            android:textColor="?text_color_headline"
             android:textStyle="bold"
             android:textSize="16sp" />
 
         <TextView
-            android:id="@+id/about_starting_text"
+            android:id="@+id/about_starting_text_basics_head"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="left"
+            android:layout_margin="7sp"
+            android:paddingLeft="3dip"
+            android:text="@string/about_starting_text_basics_head"
+            android:textColor="?text_color_headline"
+            android:textStyle="bold"
+            android:textSize="14sp" />
+
+        <TextView
+            android:id="@+id/about_starting_text_basics_text"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="left"
             android:layout_margin="7sp"
             android:autoLink="web"
             android:paddingLeft="3dip"
-            android:text="@string/about_starting_text"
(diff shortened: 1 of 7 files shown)


Lineflyer:
I will review after easter.

Regarding the translation. Its no problem to include it to crowdin as all strings are uploaded there. So the new one will come in. However it might be complicated for translators due to the length of that single string with all its html elements. 

I suspect, that it could be better to split it up in several strings with each a few lines only (e.g. each block of text) if technically feasible.
That will also only require one part to be retranslated in case of changes.

Lineflyer:
Addition:
I will also think about the content of this page a bit more while reviewing.

okainov:
Im out on vacation until beginning of May and cannot check this, but if you post screenshot of the new start page, it would be easier to take a quick look

bekuno:
@okainov please have a look at #7461

bekuno:
retest this please

bekuno:
Could someone show me the right way to view the services page?

bekuno:
retest this please

Lineflyer:
@bekuno 
This PR mixes two topics (deactivate GC, starting page) and for some reason also includes an unrelated commit from `release`

Can I ask you the favor to create to separate new PRs and check the unrelated commit contained?

PR #7528, opened 2019-04-19
remove live map warning (fix #7524)

### LINKED ISSUE #7524: Remove live map's warning about unreliable coordinates
Currently we have a warning dialog which is shown on the first 3(?) starts of the live map regarding the unreliable coords on it.

This is not needed anymore and should be removed.

moving-bits:


Bananeweizen:
Oh yes. I was so annoyed by that dialog with my recent new installation of a self compiled version...

PR #7532, opened 2019-04-21
Fix problems introduced with #7514

### LINKED ISSUE #7527: Found status not correctly merged between Live Map and Database
When the Live Map loading code is setting the found status it's value prevails. This leads to caches not getting marked as found on the map when the API the caches are loaded from is lagging behind on the latest status.
The live map code added in #7514 does set the found property of the caches.

There are two possible solutions to this:
- Set the status of the map provides this information and fix it in the merging code
- Don't set the status in the love map code and leave the merging code alone

How should this be handled to get the latest found status possible on the map?

jonas-koeritz:


jonas-koeritz:
Why does this show 18 commits but only one file changed?

Lineflyer:
add to whitelist

Lineflyer:
I think your master is not even with cgeo master....but I am not git expert.

It might be wise to develop patches in a dedicated branch on your local repo and the submit the PR from that branch.

jonas-koeritz:
I already started my next ideas in a feature branch so this wont make any problems in the future anymore.

UniQP:
@jonas-koeritz Maybe the easiest way to proceed here is to create a new feature branch, cherry-pick the relevant commits, and then create a new PR.

jonas-koeritz:
Going to do this in a cleaner branch.

PR #7538, opened 2019-04-23
Fix problems introduced with #7514, Fixes #7527

### LINKED ISSUE #7527: Found status not correctly merged between Live Map and Database

jonas-koeritz:


Lineflyer:
Can you please change the label of the PR (or commit) in a way, that the issue to be fixed is referenced?
If you use "Fix #xxxx" or "Fixes #xxxx" it will also autoclose the issue on merging, allowing a better tracking.

PR #7542, opened 2019-04-23
Don't activate gc.com connector on install (Fix #7461)

### LINKED ISSUE #7461: Don't activate gc.com connector on install

bekuno:
Fix #7461

### DIFF
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -18,7 +18,7 @@
                 <PreferenceCategory android:title="@string/settings_settings"
                     android:layout="@layout/preference_category" >
                     <cgeo.geocaching.settings.CheckBoxWithPopupPreference
-                        android:defaultValue="true"
+                        android:defaultValue="false"
                         android:key="@string/pref_connectorGCActive"
                         android:title="@string/settings_activate_gc"
                         app:text="@string/settings_gc_legal_note"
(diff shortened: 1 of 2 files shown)


Lineflyer:
Just a stupid question from my side:
This change does not affect any active installations while upgrading but only comes into effect on new installations, correct?

bekuno:
Correct. These are only initial values. (xml -> GUI visibility, java -> use in java)

Lineflyer:
OK, lets try it out then.

PR #7543, opened 2019-04-23
Create a "Getting Started" page to display on first installation

bekuno:
fix#4263 (1st impression in application is bad)

As before discussed in #7461 now separated the idea for an initial welcome screen.

I have created a "Getting Started" page that is displayed to the user the first time c:geo start after installation.
It is located in the "about c:geo" section, so the user can read it also later.
The initial text is split in 3 section - general, permission and platform information.
This could be extended if necessary, e.g. in German language to https://www.cachewiki.de/wiki/Hauptseite.

To make the cache platforms recognizable to other services, the corresponding section of the services page has been split.

The page is presented before the permissions questions (on map access).

Screenshot_20190423-233248

### DIFF
--- a/main/res/layout/about_starting_page.xml
+++ b/main/res/layout/about_starting_page.xml
@@ -25,31 +25,6 @@
             android:textSize="16sp" />
 
         <TextView
-            android:id="@+id/about_starting_text_basics_head"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:layout_gravity="left"
-            android:layout_margin="7sp"
-            android:paddingLeft="3dip"
-            android:text="@string/about_starting_text_basics_head"
-            android:textColor="?text_color_headline"
-            android:textStyle="bold"
-            android:textSize="14sp" />
-
-        <TextView
-            android:id="@+id/about_starting_text_basics_text"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:layout_gravity="left"
-            android:layout_margin="7sp"
-            android:autoLink="web"
-            android:paddingLeft="3dip"
-            android:text="@string/about_starting_text_basics_text"
-            android:textColor="?text_color"
-            android:textColorLink="?text_color_link"
-            android:textSize="12sp" />
-
-        <TextView
             android:id="@+id/about_starting_text_device_head"
             android:layout_width="wrap_content"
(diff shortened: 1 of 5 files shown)


Lineflyer:
retest this please

Lineflyer:
I would suggest to focus on "getting the app working" but not on the geocaching basics.
Therefore how about changing the sequence to:
1. c:geo device permissions
2. First steps in c:geo with link to service settings
3. Geocaching basics

I have some small improvement proposals for the text as well. I might probably push them to your branch/to this PR later today/tomorrow.

Anyway the approach is IMHO very good, not so much text but links to further information.

Lineflyer:
I can not push the change to your local branch, probably as it is marked private.

Here is my proposal (base for discussion):
https://github.com/Lineflyer/cgeo/commit/42f06a11131a5b0c05fd2ed798ee210c65bd8620

bekuno:
I'm not sure with the permission info. We use by definition:
    <uses-permission android:name="android.permission.INTERNET" />
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
    <uses-permission android:name="android.permission.NFC" />
but only for location and storage we need the explicit user agreement.
The camera is a hardware-feature.
    <uses-feature
        android:name="android.hardware.location.gps"
        android:name="android.hardware.camera"
        android:name="android.hardware.screen.portrait"
        android:name="android.hardware.nfc"
        android:name="android.hardware.touchscreen"

So I think we should reduce the info to these two points.

I suppose that when "Geocaching Basics" is after the Button to jump to the platform configuration then nobody will read this text. But I will present your sequence as next. How about to write the "Basics" sentence on top without a headline?

Screenshot_20190425-002113

Lineflyer:
retest this please

Lineflyer:
Sure geoaching basics wont maybe readed if it is at last on this page. But from experience I think the platform autorization and permissions are more often asked on our support channels. Maybe because c:geo is not very often a beginners app.

If a potential user has no idea of geocaching, then our basics wont help so much anyway. I think this page is however OK for the time being and can be superseeded by something better and more complete in future.

PR #7545, opened 2019-04-24
Fixes #7539, update disabled status in the database from live map

### LINKED ISSUE #7539: Merging of disabled status between map and saved caches
The new live map implementation now also (correctly) parses disabled status of caches. 
On the old map we did not have the needs to merge anything as the disabled status has not been parsed explicitly.

This means, that we now have the new need to merge the data in some way.
One typical scenario I experienced this weekend with the new map was:

- Have a cache saved (status at time of save was "Enabled")
- Browsing the live map you will see the cache as "Saved"/"Enabled"
- Once you click on the popup it shows "Disabled" in the popup information
- If you open the cache details it will still be "Enabled" there
- if you close details and/or the popup the live map will still show it "Enabled"

What could be a clever way to merge this information and/or make is consistent?
As a user I would not expect, that using the live map alters anything on my saved caches.
At the same time I would like to see the live map with real live information (i.e. the cache should be shown disabled there).

Maybe its the best idea to show the saved caches as disabled on the map (to be implemented) and in the popup (already working) but just keep the saved cache untouched. Once the user clicks on "More Details" in the popup he will see in the cache details, that this information was stored xxx days ago.

jonas-koeritz:
Just had exactly one cache get disabled that was on my lists and it worked flawlessly, might need some more testing though.
I opted to not touch the `storeIntoDatabase` function because I did not want to disturb the `updated` value or any other data stored.
This could possibly be extended to also update the favorite count.

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -1250,6 +1250,7 @@ public class DataStore {
         }
 
         final List<Geocache> toBeStored = new ArrayList<>();
+        final List<Geocache> toBeUpdated = new ArrayList<>();
         // Merge with the data already stored in the CacheCache or in the database if
         // the cache had not been loaded before, and update the CacheCache.
         // Also, a DB update is required if the merge data comes from the CacheCache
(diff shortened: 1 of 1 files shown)


Lineflyer:
retest this please

The failed test looks unrelated and fails from time to time as the geocoder seems to be not available

PR #7546, opened 2019-04-24
Completely remove MapToken

jonas-koeritz:
MapToken is not needed to access the new Live Map.
References #7514.


Lineflyer:
Can you explain (for me as I don't know the details of the code):
Is this only "dead code" which does not harm for the moment or a necessary change to ensure the new map is working?

jonas-koeritz:
This is just dead code that was woven into many places. The change is not mandatory to get the new map implementation working.

PR #7549, opened 2019-04-24
adapt proguard config to work with Android Studio 3.4 (fix #7548)

### LINKED ISSUE #7548: build fails with gradle error ... can't find referenced class ... ActivityCompat21
After upgrading Android Studio to current version 3.4 build fails with error message `android.support.v4.app.FragmentActivity: can't find referenced class android.support.v4.app.ActivityCompat21`

In AS 3.3 this has been a warning only (with no apparent consequences), but now build stops:

image

(I've done "clean" and "rebuild project")

Besides updating AS there has been no change on my side.
Do we have to adjust gradle settings or is there anything else I need to configure on my side?

moving-bits:
for now this is for testing only to see if it fixes all build issues occurring after upgrade to Android Studio 3.4 without unwanted side effects

### DIFF
--- a/main/proguard-project.txt
+++ b/main/proguard-project.txt
@@ -41,6 +41,7 @@
 -dontnote org.apache.http.conn.scheme.HostNameResolver
 -dontnote org.apache.http.conn.ConnectTimeoutException
 -dontnote org.apache.http.params.HttpParams
+-dontwarn android.support.v4.**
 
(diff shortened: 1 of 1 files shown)


moving-bits:
closing, as #7548 seems to work after completely deleting existing installation, followed by fresh install

PR #7561, opened 2019-04-27
check for active connectors before geocode or trackable search (fix #7554) (fix #7555)

### LINKED ISSUE #7554: Search for geocode working without active connectors
- Deactivate all geocaching platforms in c:geo
- Enter search menu
- Search for GC77 /OC1234 (or maybe even other codes related to other supported platforms)

- Cache details for this geocode are shown which probably incorrect or incomplete information (as connector is not active/authorized)

- User should be prompted by a toast informing, that this function is not supported by any active connector (as e.g. for other searches already working in that scenario)

##### Version of c:geo used:
2019.04.25-NB1

Yes

### LINKED ISSUE #7555: Search for trackable code working without active connectors
- Deactivate all geocaching platforms in c:geo
- Enter search menu
- Search for trackable GK1234 / TB59M9P  (or maybe even other codes related to other supported platforms)

- Trackable details for this tracckable are shown which probably incorrect or incomplete information (as connector is not active/authorized)

- User should be prompted by a toast informing, that this function is not supported by any active connector (as e.g. for other searches already working in that scenario)

##### Version of c:geo used:
2019.04.25-NB1

Yes

moving-bits:
- checks for any active connector before conducting a search for geocode (fix #7554)
- checks for any active trackable connector before conducting a search for trackables (fix #7555)
and shows a toast after pressing the search button, if none fitting active connector found

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -248,7 +248,6 @@
     <string name="warn_search_help_keyword">Enter all or part of a geocache\'s name. For example \"Night Cache\".</string>
     <string name="warn_search_help_user">Enter a username from Geocaching.com.</string>
     <string name="warn_search_help_tb">Enter the code for a trackable. For example \"TB29QMZ\".</string>
-    <string name="warn_search_no_active_connectors">No active connectors found for this search</string>
     <string name="warn_log_text_fill">Please input some text for your log.</string>
     <string name="warn_load_images">c:geo failed to load images.</string>
     <string name="warn_invalid_mapfile">The selected map file is not a valid mapsforge version 0.3.0 map file.\nOffline maps are not available.</string>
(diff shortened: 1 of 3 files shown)


main/res/values/strings.xml
```diff
@@ -248,6 +248,7 @@
     <string name="warn_search_help_keyword">Enter all or part of a geocache\'s name. For example \"Night Cache\".</string>
     <string name="warn_search_help_user">Enter a username from Geocaching.com.</string>
     <string name="warn_search_help_tb">Enter the code for a trackable. For example \"TB29QMZ\".</string>
+    <string name="warn_search_no_active_connectors">No active connectors found for this search</string>
```
Lineflyer: There should already be a comparable string available. It is e.g. shown when you search by keyword while no connector is active.

moving-bits: you are right, we could use this one (haven't seen):
`<string name="warn_no_connector">No currently activated geocaching services provide this type of search.</string>`

or we could change the string (at least for geocode search) to something like "cannot search, no connector activated" which points the user more prominently to the root cause

Lineflyer: IMHO for now the existing should be fine.

Bananeweizen: While this issue has already been resolved, please be aware that "connector" is an implementation internal term and should _never_ appear in user messages. When writing new resource strings, they must always fit to the mental model of the user, not to the implementation.

PR #7563, opened 2019-04-28
support new syntax for OC registration (fix #7536)

### LINKED ISSUE #7536: Create account option not working for some OC platforms
- In the menu for each geocaching service we have the option to authorize.
- In the following menu we have a button to create an account for this platform.
- for some opencaching provider then we got the error err_application_no ("c:geo can't find any suitable application.")
- Ok for opencaching.de, .us
- KO for opencaching.pl, .nl, .ro, .uk

##### Version of c:geo used:
current master on Samsung S6 / Android 7
2019.04.16 on Samsung Note 8 / Android 9

Yes

bekuno:
fix #7536 (Create account option not working for some OC platforms)

additional compose URL with site url and registration url

### DIFF
--- a/main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
+++ b/main/src/cgeo/geocaching/connector/oc/OCApiConnector.java
@@ -176,17 +176,34 @@ public class OCApiConnector extends OCConnector implements ISearchByGeocode, IOA
     @Nullable
     public String getCreateAccountUrl() {
         // mobile
-        String url = OkapiClient.getMobileRegistrationUrl(this);
+        String url = checkAccountUrl(OkapiClient.getMobileRegistrationUrl(this));
         if (StringUtils.isNotBlank(url)) {
             return url;
         }
         // non-mobile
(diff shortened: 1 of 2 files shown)


bekuno:
Mmm, this is a bug in OCPL, should be fixed in opencaching/opencaching-pl#2028

So this PR will not be necessary after the OCPL PR.
Or to take up this PR, if such a problem occurs again later?

Lineflyer:
IMHO we can wait as this is probably not a function which is called that often. Let OC-sites fix it. What do you think?

bekuno:
It is no big change at OC site.
So lets wait a little bit for the PR there.

rsudev:
I would also suggest not to introduce code on our site to workaround temporary OKAPI glitches, esp. in such a seldom used area.

Lineflyer:
OK, then lets follow the majority and close this without merging.
@bekuno I hope, this is Ok for you.

bekuno:
is fixed now in OCPL, PR here is not neccessary anymore

PR #7566, opened 2019-04-28
option added to prevent NewMap rotation on device rotation (fix #7565)

### LINKED ISSUE #7565: Add option to prevent map rotation on device rotation
When cycling & geocaching I frequently face the problem that my mobile is slightly off-angle and the screen changes its orientation. This not only takes time for screen refresh but reduces overview. (Same happens when holding my mobile in my hands while moving around etc.)

So I'm looking for an feature to lock map rotation optionally. I know that I can do this system-wide, but I want to have this for c:geo only (even better: map only), and I don't want to have to switch this on and off manually each time.

I've created a PR for this, which works fine in the emulator and on my testing device, but I want to field test this.

moving-bits:
- option added to prevent NewMap rotation on device rotation (fix #7565)
- requires API 18 minimum, useful for API 27 maximum (as API 28+ offeres a system built-in function for this) - checking current API level and offering the menu entry only for API 18-27

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -108,7 +108,13 @@
                 android:visible="false"
                 app:showAsAction="ifRoom|withText">
             </item>
-
+            <item
+                android:id="@+id/menu_map_rotation_locked"
+                android:checkable="true"
+                android:title="@string/map_rotation_lock"
+                android:visible="false"
+                app:showAsAction="ifRoom|withText">
+            </item>
         </menu>
     </item>
     <item
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -82,6 +82,7 @@
     <string translatable="false" name="pref_brouterDistanceThreshold">brouterDistanceThreshold</string>
     <string translatable="false" name="pref_maptrail">maptrail</string>
     <string translatable="false" name="pref_dot_mode">dot_mode</string>
+    <string translatable="false" name="pref_map_rotation_locked">map_rotation_locked</string>
     <string translatable="false" name="pref_defaultNavigationTool">defaultNavigationTool</string>
     <string translatable="false" name="pref_defaultNavigationTool2">defaultNavigationTool2</string>
     <string translatable="false" name="pref_gpxExportDir">gpxExportDir</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1111,6 +1111,7 @@ You can read more about <b>c:geo First steps</b> in our user guide: https://manu
     <string name="map_modes">Map settings</string>
     <string name="map_trail_show">Show trail</string>
     <string name="map_dot_mode">Use compact icons</string>
+    <string name="map_rotation_lock">Lock map rotation</string>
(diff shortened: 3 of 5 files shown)


moving-bits:
For now this is for a field test only, not for merging. Feedback from field test appreciated.

Lineflyer:
Are there some limitations of that feature depending on android version?

moving-bits:
> Are there some limitations of that feature depending on android version?

The option to lock screen rotation is avaible from API level 18 only (which is Android 4.4x IIRC), so the menu entry is show only when c:geo is running on that API level or higher only.

Also, starting with Android Pie (API level 28) Android has a system built-in function to prevent screen rotating app-wise. When turning the device while an activity is open, the screen does not rotate automatically, but a small symbol is shown instead, and the screen only rotates if you tap on this symbol. Therefore you don't need my extension when running on Pie or newer.

bekuno:
This function can be helpful also with the other map implementation.
And also for automatic triggering when rotation brings problems, e.g. when saving caches (from menu: save all, save new).
You shouls externalize as much code as possible from the new map code.

moving-bits:
Yes, will do. But for now it's just a test balloon to field test whether this approach works reasonably reliable. Maybe we can merge it to master after the next release to get a broader testing base.

moving-bits:
Did anyone had the chance to field test this one on a different device?

Lineflyer:
@moving-bits How to activate this?

> Also, starting with Android Pie (API level 28) Android has a system built-in function to prevent screen rotating app-wise.

moving-bits:
> @moving-bits How to activate this?
> 
> > Also, starting with Android Pie (API level 28) Android has a system built-in function to prevent screen rotating app-wise.

It needs Android Pie (9) or newer. Set screen mode to "portrait only".

When you now open c:geo map and rotate the screen to landscape mode, a small symbol appears in the lower right corner (besides back button). If you tap on this symbol, the screen changes to landscape mode despite the general setting is still in "portrait only" mode.
If you rotate the screen back to portrait, the symbol reappears, and the map only rotates back to portrait after tapping the symbol.

Lineflyer:
> When you now open c:geo map and rotate the screen to landscape mode, a small symbol appears in the lower right corner (besides back button)

Indeed...thanks. Learned something new on my phone :)

Lineflyer:
AFAIR we forgot to discuss about this PR in our teamspeak session. I do not see the need for such kind of setting.
IMHO we could let other apps or the OS itself do the job of controlling rotation.

moving-bits:
> AFAIR we forgot to discuss about this PR in our teamspeak session. I do not see the need for such kind of setting.
> IMHO we could let other apps or the OS itself do the job of controlling rotation.

Motivation behind this PR was that mobile devices tend to change screen orientation too frequently, which I would like to prevent app-wise instead of system-wide. The OS itself only supports rotation on or off system-wide for Android version < 9, and I'm not sure, if it's allowed for other apps to influence c:geo's rotation setting (without having your mobile device rooted). So I was looking for a different solution. But anyway, this PR is just some sort of hack, a POC only, trying to bring that feature to c:geo running under Android < v9. - I'm fine with closing this PR.

PR #7571, opened 2019-04-28
Consistently use Java style for array declarations

UniQP:


mapswithme-api/src/com/mapswithme/maps/api/MapsWithMeApi.java
```diff
@@ -46,7 +46,7 @@ OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
   public static void showMapsWithMeUrl(Activity caller, PendingIntent pendingIntent, double zoomLevel, String url)
   {
     final Uri uri = Uri.parse(url);
-    final String latlon[] = uri.getQueryParameter("ll").split(",");
+    final String[] latlon = uri.getQueryParameter("ll").split(",");
```
rsudev: Not sure, but I think this is a file provided by Mapswithme. In that case we shouldn't modify it, I'd say.

UniQP: I also created a PR for Mapswithme: https://github.com/mapsme/api-android/pull/34.

But you're right. We should fix this occurrence by upgrading the the Mapswithme API after the mentioned PR is merged.

Lineflyer: Is there any update available for this API? 
Its not working in cgeo anyway at the moment...see related issue.

UniQP: Well, they have a git repository for the API.

Lineflyer: Seems untouched since several years AFAICS.

PR #7573, opened 2019-04-28
create access to gc.su account creation web site

### LINKED ISSUE #7535: missing option to create account for geocaching.su
In the menu for each geocaching service we have the option to authorize.
In the following menu for every service except geocaching.su we have a button to create an account for this platform. (normaly redirecting to the website with a Sign up dialog)
So we need also such option also for geocaching.su.

bekuno:
fix #7535 (missing option to create account for geocaching.su)

@okainov Please be free to change it to api access if possible


okainov:
Looks fine for me, not sure, how feasible is it to add user signup via API. This should be enough :+1

PR #7575, opened 2019-04-28
Replace simple switch statements by if statements

UniQP:


Bananeweizen:
I'm not really in favor of this one. If a new menu item needs to be added there, this refactoring has to be reverted, while it currently follows the standard pattern of implementing a menu handler. I'm all for applying this in other places than the menu selection code.

Lineflyer:
> If a new menu item needs to be added there, this refactoring has to be reverted, while it currently follows the standard pattern of implementing a menu handler.

IMHO that is a valid argument to keep it then at these places.
@UniQP Shall we disregard this PR then?

UniQP:
> If a new menu item needs to be added there, this refactoring has to be reverted [...]

In general, I'm not a fan of making code less readable, because it *might* change in the future.

In the changed switch statements only case is handled by the switch and all other cases are delegated to another method. Even if new menu items are added, they most likely get handled by the delegated method and not in the method itself.

Finally, it's not a big deal to revert the refactoring.

> [...] while it currently follows the standard pattern of implementing a menu handler.

Let's assume we have the same code twice, once in a menu handler, and once in a method that isn't a menu handler. To me it's inconsistent to apply a refactoring (switch → if) for the second one but not for the first one.

Furthermore, we already use the if variant in menu handlers (e.g., https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/SearchActivity.java#L431).

Lineflyer:
@Bananeweizen What do you think?

Lineflyer:
As there was no reaction from author and requested reviewer and the PR has merge conflicts meanwhile, let me close this for now.
Feel free to rebase / resubmit anytime for further discussion,

PR #7577, opened 2019-04-28
use gradle 5.4.1

Bananeweizen:


Bananeweizen:
The CI build result needs to be checked carefully. Last time we had issues with checkstyle or so, which occured only on the server.

PR #7579, opened 2019-04-28
align more details and hint button to each other

moving-bits:
"more details" and "hint" buttons in cache popup were slightly misaligned, so lets reuse "hint" button's margin settings and apply them mirroredly to the "more details" button

### DIFF
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -36,7 +36,7 @@
                     style="@style/button_small"
                     android:layout_width="fill_parent"
                     android:layout_height="40dip"
-                    android:layout_marginRight="51dip"
+                    android:layout_toLeftOf="@id/offline_hint"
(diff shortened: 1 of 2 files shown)


rsudev:
I have an alternative suggestion for this change, which lets the layout handle it, see https://github.com/rsudev/c-geo-opensource/tree/button_alignment
Please have a look and comment.

Bananeweizen:
I haven't run the code, but I'm very much in favor of the layout XML based solution. Whenever we modify layouts, traverse parents etc. in code that is a code smell to me.

moving-bits:
@rsudev Your solution is way more elegant, thanks for pointing me to this. - `layout_toLeftOf` is what I have been looking for when I created the original PR #7469 (but had not found).

I've adjusted my PR accordingly and did some more code cleanup.

PR #7588, opened 2019-04-30
add confirmation dialog to batch upload modified coords function (fix#7585)

moving-bits:
add confirmation dialog to batch upload modified coords function (fix#7585)


moving-bits:
retest this please

PR #7590, opened 2019-05-01
Update Android gradle plugin for Android Studio 3.4

UniQP:


Bananeweizen:
With that change, for me the Gradle sync inside Android Studio fails somewhere in the test tasks initialization (something about manifestOutputDir or so, don't have that machine around right now). gradle command line builds are fine for me, but the failing sync breaks a lot of stuff in Android Studio. Can some more people try if they have good or bad experience with that change?

UniQP:
I had the same problem but it worked after a rebuild.

Bananeweizen:
I can't check until Sunday, so basically I'm fine with the change if it works for you.

bekuno:
The **gradle sync** fails with
ERROR: No value has been specified for property 'manifestOutputDirectory'.
on :main:processBasicDebugAndroidTestManifest

with
org.gradle.api.tasks.TaskValidationException: A problem was found with the configuration of task ':main:processBasicDebugAndroidTestManifest'.
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.report(ValidatingTaskExecuter.java:67)
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:54)
...
Caused by: org.gradle.api.InvalidUserDataException: No value has been specified for property 'manifestOutputDirectory'.
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.report(ValidatingTaskExecuter.java:64)

A Clean Project or Rebuild Project did not solve the error.

Nevertheless the **build** is running fine in AS after the failed sync.

UniQP:
I did some research on the internet but found no solution to the problem. At least, we are not the only ones facing the problem: https://github.com/luggit/react-native-config/issues/345.

I will set the 'Do not merge' tag until we found a solution.

Bananeweizen:
Workaround https://github.com/luggit/react-native-config/issues/345#issuecomment-489497865

Have not tested yet, still traveling...

UniQP:
The workaround works for me.

Bananeweizen:
Works for me, too. I'll therefore merge.

PR #7601, opened 2019-05-05
 Fix #7560: Use GNSS API if available

### LINKED ISSUE #7560: c:geo shows less satellites
Start c:geo and wait for a fix
Start "gps test" app and wait for a fix
compare the results 

c:geo shows 19/22 satelites
gps test shows 24/32 satelites

c:geo also shows 24/32 satelites

##### Version of c:geo used:
installed yesterday via play store

Yes

##### System information:
Using a Huawei Mate 20 Pro, that utilizes gps, glonass, beidou and galileo satelites.

UniQP:
Setting the 'Do not merge' tag until tested properly.

### DIFF
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -19,8 +19,8 @@ import cgeo.geocaching.permission.PermissionHandler;
 import cgeo.geocaching.permission.PermissionRequestContext;
 import cgeo.geocaching.sensors.GeoData;
 import cgeo.geocaching.sensors.GeoDirHandler;
-import cgeo.geocaching.sensors.GpsStatusProvider;
-import cgeo.geocaching.sensors.GpsStatusProvider.Status;
+import cgeo.geocaching.sensors.GnssStatusProvider;
+import cgeo.geocaching.sensors.GnssStatusProvider.Status;
 import cgeo.geocaching.sensors.Sensors;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.settings.SettingsActivity;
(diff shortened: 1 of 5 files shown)


UniQP:
I tested it and the number of satellites matches the number shown by GPSTest (> 30 with the PR and at most 26 with current nightly).

main/src/cgeo/geocaching/MainActivity.java
```diff
@@ -198,11 +200,13 @@ private static String formatAddress(final Address address) {
...
+    private final Consumer<GnssStatusProvider.Status> satellitesHandler = new Consumer<Status>() {
         @Override
-        public void accept(final Status gpsStatus) {
-            if (gpsStatus.gpsEnabled) {
-                navSatellites.setText(res.getString(R.string.loc_sat) + ": " + gpsStatus.satellitesFixed + '/' + gpsStatus.satellitesVisible);
+        public void accept(final Status gnssStatus) {
+            if (gnssStatus.gnssEnabled) {
+                navSatellites.setText(res.getString(R.string.loc_sat) + ": " + gnssStatus.satellitesFixed + '/' + gnssStatus.satellitesVisible);
+            } else if (VERSION.SDK_INT >= VERSION_CODES.N) {
+                navSatellites.setText(res.getString(R.string.loc_gnss_disabled));
```
Bananeweizen: I'm not sure of this additional new status. I never heard of GNSS before, and I would guess the same for many other of our users. For those "GPS" is what they want, independent of what we use to calculate the status.
Basically, this is the general UX best practice "never use implementation terminology in the user interface, the user has a different mental model than the developer".

main/src/cgeo/geocaching/sensors/GnssStatusProvider.java
```diff
@@ -0,0 +1,152 @@
...
+            public void subscribe(final ObservableEmitter<Status> subscriber) throws Exception {
+                final LocationManager geoManager = (LocationManager) context.getSystemService(Context.LOCATION_SERVICE);
+                final Callback callback = new Callback() {
+                    Status latest = NO_GNSS;
+
+                    @Override
+                    public void onSatelliteStatusChanged(final GnssStatus status) {
+                        final int visible = status.getSatelliteCount();
+                        int fixed = 0;
+                        for (int satellitIndex = 0; satellitIndex < visible; satellitIndex++) {
```
Bananeweizen: satellit_e_Index

main/src/cgeo/geocaching/MainActivity.java
Lineflyer: Yes, lets leave the UI untouched.

 Basically all people only know GPS and still expect, that all other sat systems are used as well.

PR #7608, opened 2019-05-09
upgradeSupportLibrary

Bananeweizen:


Bananeweizen:
checked output on CI system, looks good to me, therefore merging

PR #7611, opened 2019-05-11
switch to Schedulers.io().scheduleDirect from AndroidRxUtils.andThenOnUi

moving-bits:
code cleanup: switch functions related to saving / clearing the history trail to/in database from AndroidRxUtils.andThenOnUi to AsyncTask, since the `andThenOnUi` part is not needed

### DIFF
--- a/main/src/cgeo/geocaching/maps/PositionHistory.java
+++ b/main/src/cgeo/geocaching/maps/PositionHistory.java
@@ -1,7 +1,6 @@
 package cgeo.geocaching.maps;
 
 import cgeo.geocaching.storage.DataStore;
-import cgeo.geocaching.utils.AndroidRxUtils;
 
 import android.location.Location;
 
@@ -33,31 +32,21 @@ public class PositionHistory {
 
     // save current coords to permanent storage
     private void saveToStorage(final Location coordinates) {
-        AndroidRxUtils.andThenOnUi(Schedulers.io(), new Runnable() {
+        Schedulers.io().scheduleDirect(new Runnable() {
             @Override
             public void run() {
                 DataStore.saveTrailpoint(coordinates);
-            }
-        }, new Runnable() {
-            @Override
-            public void run() {
-                // nothing to do UI-wise
             }
         });
     }
 
     // clear position history (in memory and on permanent storage)
     public void reset() {
-        AndroidRxUtils.andThenOnUi(Schedulers.io(), new Runnable() {
+        Schedulers.io().scheduleDirect(new Runnable() {
             @Override
             public void run() {
                 DataStore.clearTrailHistory();
                 history.clear();
-            }
-        }, new Runnable() {
-            @Override
-            public void run() {
-                // nothing to do UI-wise
             }
         });
     }


Bananeweizen:
Looks okay to me, but I don't have much knowledge in this area.

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -33,32 +32,22 @@ public PositionHistory() {
 
     // save current coords to permanent storage
     private void saveToStorage(final Location coordinates) {
-        AndroidRxUtils.andThenOnUi(Schedulers.io(), new Runnable() {
+        AsyncTask.execute(new Runnable() {
```
UniQP: According to the RxJava documentation
```suggestion
        Schedulers.io().scheduleDirect(new Runnable() {
```
seems to more appropriate, because it still uses `Schedulers.io()`, which is dedicated for I/O-bound operations.

Lineflyer:
Besides the requested changes I would suggest only to merge this into `release` after we have the next coming release out as it IMHO requires a bit more testing. Alternatively merge to `master``at your discretion.

main/src/cgeo/geocaching/maps/PositionHistory.java
moving-bits: Today's commit changes both occurrences to `Schedulers.io().scheduleDirect`

moving-bits:
> Besides the requested changes I would suggest only to merge this into `release` after we have the next coming release out as it IMHO requires a bit more testing. Alternatively merge to `master``at your discretion.

changed base to master

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -1,14 +1,14 @@
...
 import cgeo.geocaching.storage.DataStore;
-import cgeo.geocaching.utils.AndroidRxUtils;
 
 import android.location.Location;
 
 import java.util.ArrayList;
 
 import io.reactivex.schedulers.Schedulers;
 
+
```
UniQP: Can you remove the new newline?

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -33,32 +33,16 @@ public PositionHistory() {
...
-            public void run() {
-                DataStore.clearTrailHistory();
-                history.clear();
-            }
-        }, new Runnable() {
-            @Override
-            public void run() {
-                // nothing to do UI-wise
-            }
+        Schedulers.io().scheduleDirect(() -> {
```
UniQP:
```suggestion
        Schedulers.io().scheduleDirect(new Runnable() {
```

main/src/cgeo/geocaching/maps/PositionHistory.java
moving-bits: which one - before or after line 10new? (line 12new is a comment belonging to the class definition)

main/src/cgeo/geocaching/maps/PositionHistory.java
```diff
@@ -33,32 +33,16 @@ public PositionHistory() {
...
-            @Override
-            public void run() {
-                DataStore.saveTrailpoint(coordinates);
-            }
-        }, new Runnable() {
-            @Override
-            public void run() {
-                // nothing to do UI-wise
-            }
+        Schedulers.io().scheduleDirect(() -> {
```
UniQP: For all other calls of scheduleDirect, we use:
```suggestion
        Schedulers.io().scheduleDirect(new Runnable() {
```

moving-bits: It has been like this (with `new Runnable()`) before. But we do have lambda functions now since we support Java 8, so we can make use of the new syntax available (less & cleaner code, in this case).

main/src/cgeo/geocaching/maps/PositionHistory.java
moving-bits: done

main/src/cgeo/geocaching/maps/PositionHistory.java
UniQP: I think it's more consistent to use `new Runnable()` here and add another PR that changes all occurrences to use lambda.

moving-bits: ok, updated accordingly

PR #7620, opened 2019-05-18
Use ExifInterface from support lib, not from android (that has more s…

UniQP:
…ecurity fixes)


main/build.gradle
```diff
@@ -341,6 +341,9 @@ dependencies {
     // Support Library AppCompat
     implementation "com.android.support:appcompat-v7:$supportLibraryVersion"
 
+    // Support Library ExifInterface replaces com.android.media.ExifInterface by com.android.support.media.ExifInterface
+    implementation "com.android.support:exifinterface:$supportLibraryVersion"
```
UniQP: This requires API level 25.1.0 but we still use compileSdk 24. Is this a problem?

bekuno: This means only that this library is included in this version of the support library(s).
We use now 25.4.0, so it is fine.
It has nothing to do with the compileSDK version.

see https://developer.android.com/topic/libraries/support-library/rev-archive#new-and-modified-apis_2

PR #7621, opened 2019-05-18
Add inputtype attributes to text fields

UniQP:


main/res/layout-w1200dp/coordinatescalculate_dialog.xml
```diff
@@ -80,7 +80,7 @@
                     android:overScrollMode="always"
                     android:scrollbarStyle="insideInset"
                     android:scrollbars="vertical"
-
+                    android:inputType="textAutoCorrect"
```
bekuno: I suggest to use
  android:inputType="textAutoComplete"
if I understand the documentation correctly:
* TYPE_TEXT_FLAG_AUTO_COMPLETE means the editor should show an interface for displaying suggestions, but instead of supplying its own it will rely on the Editor to pass completions/corrections.
* TYPE_TEXT_FLAG_AUTO_CORRECT means that the IME will try to auto-correct typos as the user is typing, but does not define whether the IME offers an interface to show suggestions.

main/res/layout/coordinatescalculate_dialog.xml
```diff
@@ -59,7 +59,7 @@
                 android:overScrollMode="always"
                 android:scrollbarStyle="insideInset"
                 android:scrollbars="vertical"
-
+                android:inputType="textAutoCorrect"
```
bekuno: see the other comment

main/res/layout-w1200dp/coordinatescalculate_dialog.xml
Bananeweizen: I disagree. There is no auto-completion in our app and we also cannot implement something like that. It would require us to listen to the typing in the geocaching notes field and then trying to complete things like the name or the date or the user name or... (which is what we can semantically add over plain text completion). However, that would be so much effort for so simple benefit. If someone really wants to tackle that problem at some point of time, he might do so, but with this PR ...correct instead of ...complete is what we want, I believe.

bekuno: My thoughts were in the direction of the used system keyboard and thus the handling of the text recommendations (as e.g. the Samsung keyboard does).
But for both flags a reaction of the application is expected. 
In c:geo neither of the two values has been used so far.

Therefore I suggest to use the value textNoSuggestion instead.

Lineflyer:
I can offer to test the result of this change if you are unsure about the effect.
However in this case someone might help me by explicitly naming the fields which have been changed.

main/res/layout-w1200dp/coordinatescalculate_dialog.xml
Bananeweizen: The 2 places you commented on are for field notes. There it definitely makes sense to have auto correction of the system. And that one does _not_ require anything from the application. Your comment is wrong.

Issue #7057, opened 2018-06-12
Implement location service which works while screen is off


mucek4:
##### Detailed steps causing the problem:
- Open cache
- Start compass
- Select "start talking"
- Turn off screen

##### Actual behavior after performing these steps:
GPS turns off, phone stops talking

##### Expected behavior after performing these steps:
GPS stays on, phone talks distance to cache

##### Version of c:geo used:
2018.05.28-RC

##### Is the problem reproducible for you?
Yes

##### System information:
```
--- System information ---
Device: MHA-L29 (MHA-L29, HUAWEI)
Android version: 8.0.0
Android build: MHA-L29 8.0.0.366(C432)
c:geo version: 2018.05.28-RC
Google Play services: enabled - 12.6.85 (040408-197041431)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: false
Map strategy: auto
HW acceleration: enabled (default state)
System language: en_GB
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (37.4 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (37.4 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (37.4 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (3.1 MB) on system internal storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / PREMIUM
Geocaching.com date format: MM/dd/yyyy
Installed c:geo plugins:  none
--- End of system information ---```


##### Other comments and remarks:
This might not be c:geo's bug directly. Andorid 8 + EMUI is known to stop apps when not needed. Maybe we need to start service when talking is on.

pstorch:
Maybe new energy saving options in Android?

Lineflyer:
Interesting.
My device received the Oreo update some days ago.

It still talks while the screen is off, but it seems GPS is disabled.

Lineflyer:
Setting for retest.

moving-bits:
[...]

PR #7625, opened 2019-05-18
proximity notification for map cache and waypoint popups (fix #417)

### LINKED ISSUE #417: Notification sound while seeking
It would be nice to have a notification sound, while seeking a cache and the distance to the cache goes under 50 or 40 meters.
This is also avaiable in garmin devices.

moving-bits:
Adds a notification sound when approaching the cache or waypoint currently show in the map popup.
There are two distances configurable: One for single beep, one for double beep.

add distance beep feature to map cache and waypoint popups (AbstractDialogFragmentWithBeep)
add configuration settings, strings and preference keys
add AbstractSeekbarPreference
refactor existing BroutherThresholdPreference and WpThresholdPreference to use it

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -596,6 +596,8 @@
     <string name="init_showwaypoint_description">If less than the given amount of caches are displayed on the map, their waypoints are shown additionally.</string>
     <string name="init_brouterThreshold">Calculate route</string>
     <string name="init_brouterThreshold_description">Maximum distance up to which brouter helper app (if installed) will calculate a route.</string>
+    <string name="init_distanceBeep">Distance beep</string>
+    <string name="init_summary_distanceBeep">Play one or two beeps when coming close to the active cache.</string>
     <string name="init_disabled">Exclude Disabled</string>
     <string name="init_summary_disabled">Exclude disabled caches</string>
     <string name="init_save_log_img">Save Images</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -596,8 +596,8 @@
     <string name="init_showwaypoint_description">If less than the given amount of caches are displayed on the map, their waypoints are shown additionally.</string>
     <string name="init_brouterThreshold">Calculate route</string>
     <string name="init_brouterThreshold_description">Maximum distance up to which brouter helper app (if installed) will calculate a route.</string>
-    <string name="init_distanceBeep">Distance beep</string>
-    <string name="init_summary_distanceBeep">Play one or two beeps when coming close to the active cache.</string>
+    <string name="init_distanceBeep">Proximity notification</string>
+    <string name="init_summary_distanceBeep">Play one or two signalling tones when approaching the active cache. Works from cache or waypoint popup in map view. Adjust slider "1" to set the distance for the single tone signal and slider "2" for the double tone signal.</string>
     <string name="init_disabled">Exclude Disabled</string>
     <string name="init_summary_disabled">Exclude disabled caches</string>
     <string name="init_save_log_img">Save Images</string>
--- a/main/res/layout/preference_seekbar.xml
+++ b/main/res/layout/preference_seekbar.xml
@@ -9,22 +9,29 @@
     android:paddingTop="8dp"
     tools:context=".settings.SettingsActivity" >
 
+    <TextView
+        android:id="@+id/preference_seekbar_label_view"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="left"
+        android:gravity="left"/>
+
     <SeekBar
         android:id="@+id/preference_seekbar"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_gravity="center"
         android:layout_weight="1"
-        android:paddingRight="8dp"
+        android:paddingRight="4dp"
         tools:progress="25"/>
 
     <TextView
         android:id="@+id/preference_seekbar_value_view"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:layout_gravity="center"
+        android:layout_gravity="right"
         android:layout_weight="5"
-        android:gravity="center"
+        android:gravity="right"
         tools:text="25"/>
 
 </LinearLayout>
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -80,6 +80,9 @@
     <string translatable="false" name="pref_renderthemepath">renderthemepath</string>
     <string translatable="false" name="pref_showwaypointsthreshold">waypointsthreshold</string>
     <string translatable="false" name="pref_brouterDistanceThreshold">brouterDistanceThreshold</string>
+    <string translatable="false" name="pref_distanceBeep">distanceBeep</string>
+    <string translatable="false" name="pref_distanceBeep1">distanceBeep1</string>
+    <string translatable="false" name="pref_distanceBeep2">distanceBeep2</string>
     <string translatable="false" name="pref_maptrail">maptrail</string>
     <string translatable="false" name="pref_dot_mode">dot_mode</string>
     <string translatable="false" name="pref_defaultNavigationTool">defaultNavigationTool</string>
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -601,6 +601,17 @@
             <cgeo.geocaching.settings.BrouterThresholdPreference
                 android:key="@string/pref_brouterDistanceThreshold"
                 android:layout="@layout/preference_seekbar" />
+            <CheckBoxPreference
+                android:defaultValue="false"
+                android:key="@string/pref_distanceBeep"
+                android:summary="@string/init_summary_distanceBeep"
+                android:title="@string/init_distanceBeep" />
+            <cgeo.geocaching.settings.DistanceBeepPreference1
+                android:key="@string/pref_distanceBeep1"
+                android:layout="@layout/preference_seekbar" />
+            <cgeo.geocaching.settings.DistanceBeepPreference2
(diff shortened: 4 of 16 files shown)


SammysHP:
AFAICS this does not add the feature to the compass activity, correct? I think the compass activity would be the more important location for most users interested in this feature.

moving-bits:
> AFAICS this does not add the feature to the compass activity, correct? I think the compass activity would be the more important location for most users interested in this feature.

Yes, compass activity is unchanged for now, as the cache popup in map activities is the most common use case for me, followed by the waypoint popup. But compass activity may be added later (probably with separate PR).

Lineflyer:
Interesting!!
I am in favor to give it a try if considered OK codewise by others of the team.

okainov:
Looking at #417 I have couple of comments:

Can you describe a use case (-s)? There were a lot of questions from dev team before exactly about use cases, and when I try to think of it, I realize that it's really somehow questionable.

- If you're just wandering around (walking\driving\cycling) with your audio on then probably you expect notifications when *any* cache is nearby. This PR doesn't solve this.
- If you're in the forest looking for the final cache box, they I'd really think that you either navigate yourself by simply looking at the map or using dedicated "compass" activity. Navigating when popup is open is not really the case - you don't see half of the map. This PR doesn't solve this.
- ..but when you're using compass, I assume, you have the phone in your hands and on, so you see the screen and not really blindly following the theoretical beeps. I can hardly imagine that someone walks in the forest to the cache without a screen.
- Just a side note why I'm saying "forest" - because in the cities compass doesn't help due to some GPS lags or bad precision, it's much more efficient to navigate yourself using the map (because you know where you are usually, even without GPS)


> I'm meanwhile in favor of generating speach output like "270 meters, 8 o'clock,...". 

Any comment, why have you decided to go with just beeps? There were a lot of pros in the comments for more detailed sound. Also it makes even more sense when you don't see a compass.

And for sure wording in settings need to be improved, "1" and "2" are not really helping :)

Also do you remember my comment about extracting log-scale-based slider to some reusable "base"? Back then you said WP slider it's the only log-based slider in the app. But now you've got two more. I still think that you should extract the logic to some "AbstractLogBasedSlider" or "ILogSlider" ;)

moving-bits:
> Also do you remember my comment about extracting log-scale-based slider to some reusable "base"? Back then you said WP slider it's the only log-based slider in the app. But now you've got two more. I still think that you should extract the logic to some "AbstractLogBasedSlider" or "ILogSlider" ;)

WP slider is linear. "OC max distance" would have become the only log based slider then, but we decided not to merge the "OC max distance" function.

Now we have two log based sliders for the proximity settings, and I refactored the seekbar preference to `AbstractSeekbarPreference` (which is linear, but extendable) and `AbstractDistanceBeepPreference` (which extends it to be log based).

moving-bits:
> Can you describe a use case (-s)? There were a lot of questions from dev team before exactly about use cases, and when I try to think of it, I realize that it's really somehow questionable.

My use case is quite simple: I'm on a geo trail, cycling from cache to cache of this trail, using the map display. When I'm done with a certain cache, I tap on the next cache on the map and continue.
The same works if you're looking for waypoints during a multi stage cache.

I do understand that there are probably tons of other use cases for a proximity beep, but before never ever getting a proximity beep implemented in c:geo at all I started with the ones most pressing to me :-) (no offense to anyone intended)

main/res/values/strings.xml
```diff
@@ -596,6 +596,8 @@
     <string name="init_showwaypoint_description">If less than the given amount of caches are displayed on the map, their waypoints are shown additionally.</string>
     <string name="init_brouterThreshold">Calculate route</string>
     <string name="init_brouterThreshold_description">Maximum distance up to which brouter helper app (if installed) will calculate a route.</string>
+    <string name="init_distanceBeep">Distance beep</string>
```
bekuno: I suggest not to use the word beep.
What about "Distance signaling"?

main/res/values/strings.xml
```diff
@@ -596,6 +596,8 @@
...
+    <string name="init_summary_distanceBeep">Play one or two beeps when coming close to the active cache.</string>
```
bekuno: And here:
Generates one or two signaling tones when you are approaching the enabled cache.

SammysHP:
As explained in the issue, I think this goes back to the implementation in Garmin GPS receivers. If you navigate to a POI the device beeps a single time when you are close to the selected POI (depending on speed 10-100 m before).

My use case is simple: select a cache, start compass, turn off the display, put phone in pocket and go into the direction of the cache (e.g. along a path or towards a landmark). When it beeps, stop, get the phone and then use the compass to finish the last meters.

moving-bits:
> My use case is simple: select a cache, start compass, turn off the display, put phone in pocket and go into the direction of the cache (e.g. along a path or towards a landmark). When it beeps, stop, get the phone and then use the compass to finish the last meters.

That sounds reasonable and pretty similar to my use case. I'll look into it and see if I can add this as well.

main/res/values/strings.xml
moving-bits: texts for settings title and info texts adjusted

main/res/values/strings.xml
moving-bits: texts for settings title and info texts adjusted

okainov:
>cycling <...>. When I'm done with a certain cache, I tap on the next cache on the map and continue.

Being on a bike, how do you understand where to go? That's what  I meant - two types of beep do not help you to understand that you're getting closer. Dynamic frequency would be more suitable (think of parking sensors or Geiger counter)

moving-bits:
> My use case is simple: select a cache, start compass, turn off the display, put phone in pocket and go into the direction of the cache (e.g. along a path or towards a landmark). When it beeps, stop, get the phone and then use the compass to finish the last meters.

added proximity notification to compass activity with 98971ad

bekuno:
> Being on a bike, how do you understand where to go? That's what I meant - two types of beep do not help you to understand that you're getting closer. Dynamic frequency would be more suitable (think of parking sensors or Geiger counter)

I think that this will be a more complex solution, for a more special use case.

SammysHP:
> Being on a bike, how do you understand where to go? That's what I meant - two types of beep do not help you to understand that you're getting closer. Dynamic frequency would be more suitable (think of parking sensors or Geiger counter)

This should be covered by the TTS output. It reads the direction and distance and is available in the compass activity.

Lineflyer:
> >My use case is simple: select a cache, start compass, turn off the display, put phone in pocket and go into the direction of the cache (e.g. along a path or towards a landmark). When it beeps, stop, get the phone and then use the compass to finish the last meters.
>
> That sounds reasonable and pretty similar to my use case. I'll look into it and see if I can add this as well.

Indeed, that would also be my preferred use case.

Lineflyer:
@moving-bits 
Regarding this use case:
> My use case is simple: select a cache, start compass, turn off the display, put phone in pocket and go into the direction of the cache (e.g. along a path or towards a landmark). When it beeps, stop, get the phone and then use the compass to finish the last meters.

I have seen you implemented this as well for the compass?
Did you check whether your new feature is also affected by #7057 and maybe we could -in case - fix that problem as well?

moving-bits:
> I have seen you implemented this as well for the compass?

Yes

> Did you check whether your new feature is also affected by #7057 and maybe we could -in case - fix that problem as well?

Haven't tested for this yet, but AFAIR we (or Android?) switch off GPS when screen is turned off, which means, that we don't receive any coord updates, which in turn means no proximity notifications, neither in map popups nor in compass mode. This PR does not change this behavior.

moving-bits:
> Haven't tested for this yet, but AFAIR we (or Android?) switch off GPS when screen is turned off, which means, that we don't receive any coord updates [...]

I haven't tested any of the following apps, but according to their description on Google Play they allow to stay GPS connected even if screen is turned off:
- https://play.google.com/store/apps/details?id=com.silentlexx.gpslock
- https://play.google.com/store/apps/details?id=org.bruxo.gpsconnected
- https://play.google.com/store/apps/details?id=com.webmajstr.gpson

(and many others)

Maybe this solves the problem mentioned in #7057?

Lineflyer:
The problem with #7057 is, that it is a regression. It used to work but stopped working at some point. Might be due to targetsdk change or simply due to power management on newer androud devices. I have not tested it on different versions yet.

moving-bits:
> Might be due to [...] power management on newer androud devices.

This might very well be the case, as Android has enforced power management restrictions a lot throughout newer Android versions.

moving-bits:
Setting the "do not merge" label again on this one, as I've implemented a different approach for proximity notification in map's cache and waypoint popups, which uses the same tts navigation as CompassActivity, see PR #7670.
(Proximity "beeps" could be an additional notification as it plays its sounds automatically without further user action necessary, once activated, but maybe #7670 already brings enough functionality for proximity notification and is consistent with compass' behavior. TBD later.)

SammysHP:
I'm not sure I've understood your intention. TTS and proximity beep are two distinct features. While TTS is permanently active and repeats direction and distance, proximity beep is a single event (the first time the distance is low enough after starting navigation). They can be enabled individually.

Btw: The enable status of proximity beep should be persistent.

Lineflyer:
We could use TTS for cache specific proximity notification and a beep for every cache which comes close while live map is running. 
I am just testing this POI alarm with Locus and I have to admit its very nice while on a biking tour to notice, that a cache is on the route.

SammysHP:
Quick example from my Garmin device: https://youtu.be/SkRAnZikhr8

This is what I originally intended with this feature request. There is also a proximity alert that can be configured for all wayoints.

Lineflyer:
In order to keep a bit more overview in this discussion I will continue in #417

moving-bits:
Due to the discussion in #417 PR needs to be rethought. Therefore closing for now.

PR #7626, opened 2019-05-18
upgrade google play services

bekuno:
After upgrade the support library to v.25.x we can now upgrade here.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -323,7 +323,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp:3.12.2'
 
     // Play Services
-    def playServicesVersion = '10.0.1'
+    def playServicesVersion = '11.8.0'
     implementation "com.google.android.gms:play-services-location:$playServicesVersion"
     // somehow there is a transitive play service dependency which we don't want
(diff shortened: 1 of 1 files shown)


Bananeweizen:
Do you have any experience what limits us to upgrade this further? With the support libs I'm aware of all the many resource related problems we typically run into when trying to upgrade, with the play services library I would have expected less such issues (due to less skins, strings and other resources).

bekuno:
Dependencies are all mixed up.
- The compileSdkVersion (24) should not be lower than the targetSdkVersion (27)
- This support library should not use a different version (25) than the compileSdkVersion (24)

But we can't increase the compileSdk version (correct is: "Google Inc.:Google APIs:24") as this is the last version supported by GoogleMaps v1.

Therefore I would have preferred to stay with the support library 24 in master.
But since the already used libraries are on version 25.4.0 there are probably no extra problems with the PR, because the required libs from version 25.2.0 or 25.4.0 are referenced.

However, the branch GoogleMapsV2 should be updated to the current values for targetSdkVersion, compileSdkVersion and support library version. This will finally make it possible to update the other libs as well.

PR #7629, opened 2019-05-19
use lambda functions for Schedulers.io()

moving-bits:
As proposed in https://github.com/cgeo/cgeo/pull/7611#discussion_r285342082 this is a separate PR to make use of lambda functions for `Schedulers.io()`.

### DIFF
--- a/main/src/cgeo/geocaching/AboutActivity.java
+++ b/main/src/cgeo/geocaching/AboutActivity.java
@@ -16,7 +16,6 @@ import android.os.Bundle;
 import android.support.annotation.RawRes;
 import android.support.annotation.StringRes;
 import android.view.View;
-import android.view.View.OnClickListener;
 import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.ScrollView;
@@ -102,13 +101,7 @@ public class AboutActivity extends AbstractViewPagerActivity<AboutActivity.Page>
             } else {
                 changeLogMaster.setMovementMethod(AnchorAwareLinkMovementMethod.getInstance());
             }
-            changeLogLink.setOnClickListener(new OnClickListener() {
-
-                @Override
-                public void onClick(final View v) {
-                    startUrl("https://github.com/cgeo/cgeo/releases");
-                }
-            });
+            changeLogLink.setOnClickListener(v -> startUrl("https://github.com/cgeo/cgeo/releases"));
             return view;
         }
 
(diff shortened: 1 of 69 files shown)


UniQP:
We have a lot more occurrences, where we can replace a `new Runnable()...` by a lambda function. We even have some more occurrences for `scheduleDirect`.

@moving-bits Can you extend the PR and handle (at least) the missing `scheduleDirect` calls?

moving-bits:
I had processed all files below main/src/geocaching and - AFAICS - converted all `scheduleDirect()` calls.

I'm unsure, though, about constructs like the following:

```
AndroidRxUtils.andThenOnUi(Schedulers.io(), new Callable<String>() {
    @Override
    public String call() {
        return DataStore.backupDatabaseInternal();
    }
},
[...]
```
(There are a couple occurrences similar to this, spread over three files.)
How would this have to be converted?

UniQP:
`git grep -l "scheduleDirect(new Runnable"`
```
main/src/cgeo/geocaching/CacheDetailActivity.java
main/src/cgeo/geocaching/EditWaypointActivity.java
main/src/cgeo/geocaching/MainActivity.java
main/src/cgeo/geocaching/activity/OAuthAuthorizationActivity.java
main/src/cgeo/geocaching/activity/TokenAuthorizationActivity.java
main/src/cgeo/geocaching/maps/CGeoMap.java
main/src/cgeo/geocaching/maps/mapsforge/v6/caches/SinglePointOverlay.java
main/src/cgeo/geocaching/models/Geocache.java
main/src/cgeo/geocaching/network/HtmlImage.java
main/src/cgeo/geocaching/network/Send2CgeoDownloader.java
main/src/cgeo/geocaching/playservices/LocationProvider.java
main/src/cgeo/geocaching/utils/RxUtils.java
```

moving-bits:
my bad, I was looking for `Schedulers.io()` - will update the PR

jakedot:
> I had processed all files below main/src/geocaching and - AFAICS - converted all `scheduleDirect()` calls.
> 
> I'm unsure, though, about constructs like the following:
> 
> ```
> AndroidRxUtils.andThenOnUi(Schedulers.io(), new Callable<String>() {
>     @Override
>     public String call() {
>         return DataStore.backupDatabaseInternal();
>     }
> },
> [...]
> ```
> 
> (There are a couple occurrences similar to this, spread over three files.)
> How would this have to be converted?

It all works the same - you convert an occurence of

```
new SingleMethodInterface {
  R method(A a, B b, C c, ...) {
    return r;
  }
}
```

into a Lambda like

`(a,b,c, ...) -> r`

or when containing control structures like if, for, try-catch, ...

```
(a,b,c, ...) -> {
  [...]
  return r;
  [...]
}
```

- throws declations don't need to be declared in the Lambda, it is sufficient that they are contained in the Interface.

So a Callable like your example would become

```
AndroidRxUtils.andThenOnUi(Schedulers.io(), () -> DataStore.backupDatabaseInternal(),
[...]
```

moving-bits:
@jakedot Thanks for your detailed reply. I was a little bit confused about how to convert parameters with additional type information, like this:

`AndroidRxUtils.andThenOnUi(Schedulers.io(), new Callable<String>() {`

But it seems (according to Android Studio's assistant function (Alt-Return)), that the additional type info simply can be left out.

jakedot:
Yes, generic type parameters can usually be left out. There are only a few cases where I think this is not the case, and then you can specify the parameters explicitly like so:

`(String a, Object b, SomeClass c) -> { [...] }`

Lineflyer:
retest this please

jakedot:
I hope that #7668 is not related to this PR, but mentioning it may be in
order. Did you have any other problems apart from understanding the
generics syntax? As mentioned/teasered in the issue, Lambdas are designed
to replace Single Method Interfaces but they're implemented a bit
differently under the hood and choosing target java version or similar
javac options can help but lead to more inperformant bytecode for modern
JVMs...

I have been spending quite some time on learning this in uni and at work
and while I'm in no means an expert, feel free to ask me stuff on this
topic!

PR #7632, opened 2019-05-20
remove outdated support v7 appcompat integration

bekuno:
The source tree contains a very old tree "android-support-v7-appcompat".
I don't see where this is used and makes sense.


Bananeweizen:
It's used by the Eclipse build, but I can manage this locally.

PR #7635, opened 2019-05-21
upgrade dependencies

Bananeweizen:
* 2 libraries have new minor versions
* inline play services version, it is used only one time
* convert support lib version to project extension, otherwise the contacts plugin gradle config easily gets out of sync with the main gradle config

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,9 @@ buildScan {
  * common configuration for the Java projects
  */
 
-ext.minSdkVersion = 14
+ext {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -155,6 +155,7 @@ buildScan {
  */
 
 ext.minSdkVersion = 14
+ext.supportLibraryVersion = '25.4.0'
```
bekuno: please use the block form:
ext {
    minSdkVersion = 14
    supportLibraryVersion = '25.4.0'
}

PR #7639, opened 2019-05-24
add horizontal progress spinner to maps (fix #7289)

### LINKED ISSUE #7289: Loading progress spinner missing on live map

moving-bits:
adds a centered circular progress spinner to CGeoMap and NewMap which is shown while loading caches in live map (and which replaces the no longer supported actionbar based spinner) (fixes #7289)

### DIFF
--- a/main/res/layout/map_google.xml
+++ b/main/res/layout/map_google.xml
@@ -38,6 +38,9 @@
             android:clickable="true"
             android:enabled="true"
             android:keepScreenOn="true" />
+
+        <include layout="@layout/map_progressbar" />
+
     </RelativeLayout>
 
 </LinearLayout>
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -162,6 +162,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory {
     private int detailTotal = 0;
     private int detailProgress = 0;
     private long detailProgressTime = 0L;
+    private ProgressBar spinner;
 
     // views
     private CheckBox myLocSwitch = null;
@@ -323,13 +324,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory {
             if (map == null) {
                 return;
             }
-
-            final ProgressBar progress = ButterKnife.findById(map.activity, R.id.actionbar_progress);
-            if (progress != null) {
-                final int visibility = show ? View.VISIBLE : View.GONE;
-                progress.setVisibility(visibility);
-            }
-            map.activity.setProgressBarIndeterminateVisibility(show);
+            map.spinner.setVisibility(show ? View.VISIBLE : View.GONE);
         }
 
     }
@@ -473,6 +468,9 @@ public class CGeoMap extends AbstractMap implements ViewFactory {
             overlayPositionAndScale.setHistory(trailHistory);
         }
 
+        // prepare circular progress spinner
+        spinner = (ProgressBar) activity.findViewById(R.id.map_progressbar);
+        spinner.setVisibility(View.GONE);
 
         mapView.repaintRequired(null);
 
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -74,6 +74,7 @@ import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
 import android.widget.CheckBox;
 import android.widget.ListAdapter;
+import android.widget.ProgressBar;
 import android.widget.TextView;
 
 import java.io.File;
@@ -136,6 +137,7 @@ public class NewMap extends AbstractActionBarActivity implements XmlRenderThemeM
 
     private ProgressDialog waitDialog;
     private LoadDetails loadDetailsThread;
+    private ProgressBar spinner;
 
     private String themeSettingsPref = "";
 
@@ -198,6 +200,10 @@ public class NewMap extends AbstractActionBarActivity implements XmlRenderThemeM
 
         setContentView(R.layout.map_mapsforge_v6);
         setTitle();
+
+        // prepare circular progress spinner
+        spinner = (ProgressBar) findViewById(R.id.map_progressbar);
+        spinner.setVisibility(View.GONE);
 
         // initialize map
         mapView = (MfMapView) findViewById(R.id.mfmapv5);
@@ -1140,7 +1146,7 @@ public class NewMap extends AbstractActionBarActivity implements XmlRenderThemeM
             if (map == null) {
                 return;
             }
-            map.setProgressBarIndeterminateVisibility(show);
+            map.spinner.setVisibility(show ? View.VISIBLE : View.GONE);
         }
 
     }
--- a/main/res/layout/map_mapsforge.xml
+++ b/main/res/layout/map_mapsforge.xml
@@ -37,6 +37,9 @@
             android:clickable="true"
             android:enabled="true"
             android:keepScreenOn="true" />
+
+        <include layout="@layout/map_progressbar" />
+
     </RelativeLayout>
 
 </LinearLayout>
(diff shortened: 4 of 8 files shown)


main/src/cgeo/geocaching/maps/CGeoMap.java
```diff
@@ -330,6 +331,7 @@ private void showProgress(final boolean show) {
                 progress.setVisibility(visibility);
             }
             map.activity.setProgressBarIndeterminateVisibility(show);
```
Bananeweizen: If I get it right, we can remove the activity.setProgress... calls everywhere, since they are not doing anything anymore?

main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
```diff
@@ -1141,6 +1147,7 @@ private void showProgress(final boolean show) {
                 return;
             }
             map.setProgressBarIndeterminateVisibility(show);
```
Bananeweizen: same here

main/src/cgeo/geocaching/maps/CGeoMap.java
moving-bits: Will update accordingly. There is probably more code specific to the old actionbar progress spinner. Do you know where the layout of the actionbar is defined?

main/res/layout/map_google.xml
```diff
@@ -38,6 +38,16 @@
             android:clickable="true"
             android:enabled="true"
             android:keepScreenOn="true" />
+
+        <ProgressBar style="action_bar_progress"
```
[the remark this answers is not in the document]
moving-bits: done

main/src/cgeo/geocaching/maps/CGeoMap.java
bekuno: For the setProgressBarIndeterminateVisibility deprecation, also 
the initialization with Window.FEATURE_INDETERMINATE_PROGRESS
is deprecated and needs to be removed.

See
https://developer.android.com/reference/android/app/Activity#setProgressBarIndeterminateVisibility(boolean)
https://developer.android.com/reference/android/view/Window.html#FEATURE_INDETERMINATE_PROGRESS

bekuno:
Could not resist to test - here on my S6 (Android 7) on OSM map.

Screenshot_20190525-115452

moving-bits:
I adjusted the spinner size for better visibility:

image

main/src/cgeo/geocaching/maps/CGeoMap.java
moving-bits: done

main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
moving-bits: done

Lineflyer:
That looks perfect

bekuno:
What about also removing
        supportRequestWindowFeature(Window.FEATURE_INDETERMINATE_PROGRESS);
in AbstractActivity?
I don't saw other activities requiring this feature.

moving-bits:
> What about also removing
> supportRequestWindowFeature(Window.FEATURE_INDETERMINATE_PROGRESS);
> in AbstractActivity?
> I don't saw other activities requiring this feature.

The general removal of all references to the ActionBar spinner is a different issue (which will probably require adjustments similar to this PR in other activities as well). I would like to keep that in a separate PR.

okainov:
I also tried, my few cents:

- Sometimes it rotates too fast. It would be much nicer if it's not so crazy fast and rotating in a calm pace.
- Since it's a big something in the middle of the screen, it's kinda annoying especially when I do many short map moves around already loaded area. Suggestion to show the spinner only if loading takes hmm... more than a second, for example? 
- Disable Internet and move around the map. Spinner appears for a few ms and disappears. As a user, I see it as "loading completed, done, all set". However in fact, obviously, no any loading was performed (because I have no Internet). I suppose the root cause is not the spinner itself, but loading logic, however with huge something in the middle of the screen it's much more noticeable now.

- I'm far from UX guy, but still.... isn't there better place than middle of the screen? Upper left corner?

Bananeweizen:
>     * I'm far from UX guy, but still.... isn't there better place than middle of the screen? Upper left corner?

To my understanding this has been deprecated mainly such that developers use the progress bar at the top of the activity window. However, I have no clue how much effort migrating to that component is and whether we can use it with our current version of the support libs. That's why I did not object the PR.
https://material.io/design/components/progress-indicators.html

okainov:
>To my understanding this has been deprecated mainly such that developers use the progress bar at the top 

Yes, I remember you said about progress bar. But what about having the spinner not in the action bar, but just in the corner of the map? 

image

moving-bits:
As for the first three remarks: I have not changed anything in terms of when and how long the spinner gets displayed as I reuse the existing function calls. The only change is size and location of the spinner.

_(ActionBar based spinner is no longer supported starting target SDK 24. One solution to this would be to migrate from ActionBar to Toolbar component which looks like a bigger refactoring to me.)_

I'm open to adjustments in size in placements. What do others think: What is the expected UI behaviour in this case?

bekuno:
My favorite is the linear progress indicator. 
It is not so concise on top of the map (cache symbols).

Lineflyer:
What matters most to me is, to have some kind of loading information back to the UI as this is a regression since the targetSDK change.
So I am fine with this centered circular spinner, but also have to admit, that due to the fact, that it is more intrusive than the old one in the action bar, it might need some improvements of its behaviour (e.g. split second loading as described by @okainov). However this can/should be handled as a different task afterwards.

Regarding the use of the linear progress indicator:
This would be a much better solution for sure, but if the effort is significantly more or not doable by @moving-bits right now, we should go with the circular spinner for now and add another issue to migrate to the linear indicator.

Another question:
Does the spinner block interaction with map elements/caches shown at the same place?

moving-bits:
Switched the large circular, centered spinner to a very unobstrusive horizontal indeterminate progressbar at the top of the map: (notice the light blue line)

grafik

Lineflyer:
Looke more nice than the circular one. Do you consider ready to be merged?

Any problems to be expected on older Android versions...or is this thing compatible to old versions?

bekuno:
It is derived from the AppCompat library, so that should work on old androids.

moving-bits:
> Do you consider ready to be merged?

I squashed the commits into one to make it easier to understand what's finally in this PR.

Have tested with SDK 24 in emulator only (I'm not at my regular dev machine currently), but I don't expect anything unusual to happen - at most no spinner is shown.

Lineflyer:
@moving-bits 
The review of @Bananeweizen has been addressed already, has it?

moving-bits:
> @moving-bits
> The review of @Bananeweizen has been addressed already, has it?

Yes, it is. (I always set the conversations to "resolved" when requested changes are implemented or the cause is solved by other means.)

Lineflyer:
Sorry, I reverted this as I was a bit fast with merging and overlooked, that it targeted the `release` branch.
While this is OK generally, I would like to have some more testing time with this one and rather like to have it on `master` only instead (also looking at the hopefully upcoming bugfix release for the OSM tile loading).

@moving-bits 
Can you target your PR to `master` please?
Sorry for the additional effort.

rsudev:
I could merge the change (based on release) to master first. That would leave the option to put it on release directly later still open.

moving-bits:
> I could merge the change (based on release) to master first. That would leave the option to put it on release directly later still open.

@rsudev Thanks, this would be very helpful to me, as I won't be back at my dev machine before Thursday.

Lineflyer:
> I could merge the change (based on release) to master first. That would leave the option to put it on release directly later still open.

Yes, that would be great.

While you are at it, you could also take a look into merging changes on `release` to `master`. I checked via Github UI but it seems there are conflicts to be resolved.

rsudev:
rebased on latest release and merged to master (merged release to master as well)
could not push to moving-bits, so pushed to upstream.

PR #7641, opened 2019-05-26
api.geokretymap.org replaced by api.geokrety.org (fix #7638)

### LINKED ISSUE #7638: api.geokretymap.org replaced by api.geokrety.org
From @kumy in #7615 & https://github.com/GeoKretyMap/gkm-api-php/issues/13
https://api.geokretymap.org will be deprecated and replaced by https://api.geokrety.org (Fusion of GKM and GK projects)

bekuno:
fix #7638 (api.geokretymap.org replaced by api.geokrety.org)

- change the references to api.geokretymap.org with api.geokrety.org
- change the text on the geokrety settings page
- AndroidManifest: remove the references to geokretymap.org, integrate into geokrety.org

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1586,8 +1586,8 @@
     <string name="init_geokrety_register_ok">Login successful</string>
     <string name="init_geokrety_register_fail">Login failed</string>
     <string name="init_geokrety_description">Authorize c:geo with GeoKrety.org to log GeoKrety trackables.</string>
-    <string name="init_geokrety_cache_description">GeokretyMap.org provides cached datas for GeoKrety trackables.\nAvailable datas contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nDatas are not live!\nPositions: ~30min.\nDescriptions: ~24h.</string>
-    <string name="init_geokrety_cache">Use GeoKretyMap.org</string>
+    <string name="init_geokrety_cache">Use api.GeoKrety.org</string>
+    <string name="init_geokrety_cache_description">api.Geokrety.org provides cached data for GeoKrety trackables.\nAvailable data contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nData is not live!\nPosition: ~30min.\nDescription: ~24h.</string>
     <string name="init_geokrety_userid">User Id: %s</string>
     <string name="trackable_title_log_without_geocode">Geocode not set</string>
     <string name="trackable_log_without_geocode">You\'re about to send a log without a Geocode. It\'s encouraged to set one. Do you want to add it now?</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1586,8 +1586,6 @@
     <string name="init_geokrety_register_ok">Login successful</string>
     <string name="init_geokrety_register_fail">Login failed</string>
     <string name="init_geokrety_description">Authorize c:geo with GeoKrety.org to log GeoKrety trackables.</string>
-    <string name="init_geokrety_cache">Use api.GeoKrety.org</string>
-    <string name="init_geokrety_cache_description">api.Geokrety.org provides cached data for GeoKrety trackables.\nAvailable data contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nData is not live!\nPosition: ~30min.\nDescription: ~24h.</string>
     <string name="init_geokrety_userid">User Id: %s</string>
     <string name="trackable_title_log_without_geocode">Geocode not set</string>
     <string name="trackable_log_without_geocode">You\'re about to send a log without a Geocode. It\'s encouraged to set one. Do you want to add it now?</string>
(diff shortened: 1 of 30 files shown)


main/res/values/strings.xml
```diff
@@ -1586,8 +1586,8 @@
     <string name="init_geokrety_register_ok">Login successful</string>
     <string name="init_geokrety_register_fail">Login failed</string>
     <string name="init_geokrety_description">Authorize c:geo with GeoKrety.org to log GeoKrety trackables.</string>
-    <string name="init_geokrety_cache_description">GeokretyMap.org provides cached datas for GeoKrety trackables.\nAvailable datas contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nDatas are not live!\nPositions: ~30min.\nDescriptions: ~24h.</string>
-    <string name="init_geokrety_cache">Use GeoKretyMap.org</string>
+    <string name="init_geokrety_cache">Use api.GeoKrety.org</string>
+    <string name="init_geokrety_cache_description">api.Geokrety.org provides cached data for GeoKrety trackables.\nAvailable data contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nData is not live!\nPosition: ~30min.\nDescription: ~24h.</string>
```
kumy: As GK and GKM are now one project, api.geokrety.org should be the default.

We may eventually keep the old fashion as a fallback (probably in case of GK api side service failure) or simply remove it.

Lineflyer: My question would go into the same direction:
Do we then still need this additional option to use a second platform to receive the data. I would imagine, we could completely remove the (optional) differentiation between GKM and GK in code and menu?!

kumy: BTW, the timing set in the original message are now completely obsolete, there is just 1 minute max of delay. But we still have cached responses for 15 minutes (now I'm wondering if we should keep this cache or lower it a bit…)

bekuno: > Do we then still need this additional option to use a second platform to receive the data. I would imagine, we could completely remove the (optional) differentiation between GKM and GK in code and menu?!

api.geokrety.org has no direct access to the geokrety database, if I understand the change right.

My test case is
- open geokret (maybe open cache, open log dialog, open geokret from inventory)
- write and sent log
- refresh 
- the log shoud be in the log list

Currently without api.geokrety.org we have no log list, with api I have to refresh some times until the log will be visible. The log can be seen immediately on the geokrety website.
(With a TB I habe the log instantly after refresh)

I think we can remove the extra entry for api.geokrety(map).org from the options as first step in this PR.
In another PR the complete code conversion to api.gekrety org should be realized.

kumy: Effectively, there is a little delay by design, but the cache of 15min may be a big problem. I'll disable it.

bekuno: > As GK and GKM are now one project, api.geokrety.org should be the default.

GKM was already default.
The isGeokretyCacheActive() should now always return true.

main/res/values/strings.xml
```diff
@@ -1586,8 +1586,6 @@
     <string name="init_geokrety_register_ok">Login successful</string>
     <string name="init_geokrety_register_fail">Login failed</string>
     <string name="init_geokrety_description">Authorize c:geo with GeoKrety.org to log GeoKrety trackables.</string>
-    <string name="init_geokrety_cache">Use api.GeoKrety.org</string>
-    <string name="init_geokrety_cache_description">api.Geokrety.org provides cached data for GeoKrety trackables.\nAvailable data contains more GeoKrety informations like \'description\' or \'missing\' status.\n\nData is not live!\nPosition: ~30min.\nDescription: ~24h.</string>
```
kumy: @rsudev, don't we have some special treatment on string deletion? I can't remember exactly…

main/res/values/strings.xml
kumy: GKM cache on select GK endpoint is now disabled. Ex https://api.geokretymap.org/gk/73512 should refresh the date attribute in gkxml element on each page refresh.

bekuno:
@mucek4 We have a problem with jenkins. The last 2 jobs should not fail.

main/res/values/strings.xml
Lineflyer: @kumy Deleting/Editing/Adding in the english strings file is fine. Optionally removals can also be done for the other languages, but anyway Crowdin will sync this on next up/download to Crowdin.

kumy:
```
23:02:57  09:02:59 V/InstrumentationResultParser: 1) testService(cgeo.geocaching.connector.trackable.GeolutinsParserTest)
23:02:57  09:02:59 V/InstrumentationResultParser: java.lang.AssertionError:
23:02:57  09:02:59 V/InstrumentationResultParser: Expecting actual not to be null
```
GeoLutins seems down :(
https://downforeveryoneorjustme.com/geolutins.com

kumy:
> GeoLutins seems down :(

I've dropped a mail to GeoLutins owner, let's wait a bit.

Lineflyer:
I have also seen these failing tests since this morning.
Might it be an idea to make a thread in the github team forum to collect contact persons / addresses for the supported services...because in this case I struggled  bit to find a suitable communication channel to report it.

bekuno:
Retest this please

Issue #7494, opened 2019-04-14
Upgrade to mapsforge 0.11.0


rsudev:
Upgrade to the latest mapsforge library version (0.11.0 from 2019-03-25).
Doing it now on master allows the longest test cycle to find issues/regressions.

rsudev:
As the upgrade was reverted due to unclear performance issues showing up, this has to be addressed again.

Lineflyer:
@rsudev Thanks for reverting.
To be on the safe side I built a fresh nightly and tested again (i.e. current master reverted back to 0.9.1). The problem described in #7523 is gone.

I suggest to close #7523 for now and potentially discuss next steps in this issue instead.

jonas-koeritz:
The currently used Mapsforge 0.9.1 doesn't work on Android 9, couldn't find any issues with 0.11.0 so far.

Issue #7523, opened 2019-04-19
Zooming in is bumpy on mapsforge 0.11


Lineflyer:
I noticed since some days (nightlies), that when pinching or zooming on an offline map the rerendering is not really smooth all the time.

Sometimes it workd normal but sometimes tiles need lot of time to show up in new zoomlevel and cache icons remain too large (on zoomout) until they are new rendered.
I have  never seen this so frequently befor and on my Samsung S10 I would not think its a device performance problem.

From the commit history I suspect the upgrade to mapsforge 0.11 or the trail function (but it happens also if the trail is switched off....so probably not related).

moving-bits:
Lastly I have rendering problems (tile loading problems) as well. In LogCat I can see a couple of error messages, which seem related to host name resolving:

`Unable to resolve host "c.tile.openstreetmap.org": No address associated with hostname java.net.UnknownHostException: Unable to resolve host "c.tile.openstreetmap.org": No address associated with hostname
        at java.net.Inet6AddressImpl.lookupHostByName(Inet6AddressImpl.java:95)
        at java.net.Inet6AddressImpl.lookupAllHostAddr(Inet6AddressImpl.java:74)
        at java.net.InetAddress.getAllByName(InetAddress.java:752)
        at com.android.okhttp.internal.Network$1.resolveInetAddresses(Network.java:29)
        at com.android.okhttp.internal.http.RouteSelector.resetNextInetSocketAddress(RouteSelector.java:187)
        at com.android.okhttp.internal.http.RouteSelector.nextProxy(RouteSelector.java:156)
        at com.android.okhttp.internal.http.RouteSelector.next(RouteSelector.java:98)
        at com.android.okhttp.internal.http.HttpEngine.createNextConnection(HttpEngine.java:345)
        at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:328)
        at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:246)
        at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:457)
        at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:405)
        at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getHeaders(HttpURLConnectionImpl.java:162)
[...]

PR #7653, opened 2019-05-30
set target SDK to 28

moving-bits:
Play store requires all existing apps to target SDK 28 to be still upgradeable beginning November 19 so let's start testing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -155,7 +155,7 @@ buildScan {
  */
 
 ext {
-    supportLibraryVersion = '27.1.1'
+    supportLibraryVersion = '28.0.0'
 }
 
 subprojects {
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -5,7 +5,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion 28
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -14,7 +14,7 @@ android {
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 27
+        targetSdkVersion 28
     }
 
     sourceSets {
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -82,6 +82,9 @@
             android:name="com.google.android.geo.API_KEY"
             android:value="@string/maps_api2_key" />
 
+        <!-- required by Google Maps SDK version 16 and below if targeted to SDK 28 or above -->
+        <uses-library android:name="org.apache.http.legacy" android:required="false" />
+
         <!-- support also very wide screen devices. https://developer.android.com/guide/practices/screens_support.html#MaxAspectRatio -->
         <meta-data
             android:name="android.max_aspect"
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -30,7 +30,7 @@ android {
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 27
+        targetSdkVersion 28
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)
 
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -18,7 +18,7 @@ if (isContinuousIntegrationServer()) {
 }
 
 android {
-    compileSdkVersion 27
+    compileSdkVersion 28
 
     buildToolsVersion "28.0.3"
 
--- a/mapswithme-api/build.gradle
+++ b/mapswithme-api/build.gradle
@@ -7,7 +7,7 @@ android {
 
   defaultConfig {
     minSdkVersion 16
-    targetSdkVersion 27
+    targetSdkVersion 28
   }
 
   sourceSets.main {
--- a/mapswithme-api/build.gradle
+++ b/mapswithme-api/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'android-library'
 android {
 
   // Define these properties in the gradle.properties file in the root project folder
-  compileSdkVersion 27
+  compileSdkVersion 28
 
   defaultConfig {
     minSdkVersion 16


Lineflyer:
Wasnt there the problem, that this will break gmapsv1?

In this case I would suggest another approach:
1. Set target to 28 on gmapsv2 branch and test it
2. Try to tackle the listed bugs of gmapsv2 
3. At some point merge the new gmapsv2 to master

What do you think?

moving-bits:
> Wasnt there the problem, that this will break gmapsv1?

Can't say. It compiles with GMv1, and the visual result is the same for me:

image

That is to say, that the map tiles were not visible for me before as well - my key does not seem to be valid for Google. There are no error messages regarding GM in Logcat for me.

Lineflyer:
Before merging this, let us wait for next branching from `master`to `release` as mentioned in #7665.
Meanwhile I might give the debug.apk a testrun.

bekuno:
Here is the site descriping the "behavior changes: apps targeting API level 28+":
https://developer.android.com/about/versions/pie/android-9.0-changes-28

Lineflyer:
I did not come very far with testing the debug.apk from CI as seemingly these APK do not contain any of our keys (No Google Maps, No authorization towards opencaching). I also tested other debug.apk from CI with the same result. So this is independent of the targetSDK and I am wondering whether this is normal on the CI?

moving-bits:
> Wasnt there the problem, that this will break gmapsv1?

It breaks when you set the "Compile SDK" to 28 - it's not compilable anymore then (which I encountered while testing the "keep GPS going" issue using a foreground service).

Lineflyer:
Shall we merge this to `master` now to check the (unexpected) consequences and as a first step towards GMaps V2?

jonas-koeritz:
Mapsforge does break on targetSDK 28. Upstream Mapsforge 0.11.0 seems to work with only minor modifications. I could create a pull request for that.

Lineflyer:
Oh, I did not expect a problem with target28 and mapsforge. Do you know the cause?

Generally we could use mf0.11 but recently we rolled back to 0.91 due to performance issues. You might want to browse our issues for details.
I assume we will run into this again once we upgrade to 0.11.
Still we could do this, but then we should debug and solve this performance problem.

jonas-koeritz:
See https://github.com/mapsforge/mapsforge/commit/b7b07619a6389886734acd5bcb8ae9b14dfce8c2 for reference, I currently do not experience any performance problems when running on targetSdk 27 and Mapsforge 0.11.0

Lineflyer:
See #7494 for the task of upgrading to mf0.11 (You might want to include "Fix #7494" in your corresponding PR).

Additionally #7523 describes the performance problem with mf0.11

jonas-koeritz:
I would say that the "bumpy zoom" has been there on 0.9.1 too, when zooming in, everything looked just like the screenshot on #7523. I'm going to test c:geo with mapsforge 0.11.0 a bit, currently it seems about as fast as before.

Lineflyer:
retest this please

UniQP:
@moving-bits Can you also increase the `compileSdk`?

moving-bits:
> @moving-bits Can you also increase the `compileSdk`?

Done.

Also added  `<uses-library android:name="org.apache.http.legacy" android:required="false" />` to `AndroidManifest.xml` to fix #7841.

Result is compilable and both Google Maps and OSM show up. (Although a first regression comes up: #7854)

moving-bits:
@mucek4 
Could you once again accept licence agreement for SDK 28, please?

from error log of last build:
 > Failed to install the following Android SDK packages as some licences have not been accepted.
23:15:44       platforms;android-28 Android SDK Platform 28
23:15:44    To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
23:15:44    Alternatively, to transfer the license agreements from one workstation to another, see http://d.android.com/r/studio-ui/export-licenses.html

bekuno:
Before switching to compileSdk/targetSdk 28 we should at first update Upgrade the support libraries.

Cite from
https://google-developer-training.github.io/android-developer-fundamentals-course-concepts-v2/unit-1-get-started/lesson-3-testing,-debugging,-and-using-support-libraries/3-3-c-the-android-support-library/3-3-c-the-android-support-library.html#versions :
~~~
As a general rule, use the most recent version of the support library for the API your app is compiled and targeted for, or a newer version. For example, if your app targets API 26, use the version 26.x.x of the support library.
You can always use a newer support library than the one for your targeted API. For example, if your app targets API 22 you can use version 25 or higher of the support library. The reverse is not true—you cannot use an older support library with a newer API. As a general rule, you should try to use the most up-to-date API and support libraries in your app. 
~~~

Let's start with version 7.1.1 .

mucek4:
I am updating executor 3 with to latest Docker image. When it gets back online (and if it works) I have to build 4 and 5 too.

moving-bits:
@mucek4:

> I am updating executor 3 with to latest Docker image. When it gets back online (and if it works) I have to build 4 and 5 too.

Hi mucek4, I don't know if you'd found the time to update cgeo-3 already, but currently build still fails on that CI with the following message:

> File /home/jenkins/.android/repositories.cfg could not be loaded.
> Checking the license for package Android SDK Platform 28 in /opt/android-sdk-linux/licenses
> Warning: License for package Android SDK Platform 28 not accepted.

(see https://ci.cgeo.org/job/cgeo%20pull%20request/2300/console)

mucek4:
I did. Executor 3 should be the latest. I'll check tomorrow.

moving-bits:
@mucek4

Thanks

bekuno:
folow up the discussion in PR #7861

I got these errors - in the test suite:
~~~
> Task :main:compileBasicDebugAndroidTestJavaWithJavac
C:\Users\admin\StudioProjects\cgeo\tests\src-android\cgeo\CGeoTestCase.java:9: error: package android.test does not exist
import android.test.ApplicationTestCase;
                   ^
C:\Users\admin\StudioProjects\cgeo\tests\src-android\cgeo\CGeoTestCase.java:13: error: cannot find symbol
public abstract class CGeoTestCase extends ApplicationTestCase<CgeoApplication> {
                                           ^
... and so on ....
~~~

bekuno:
A currently failing function here (not automatic tested) 
see https://github.com/cgeo/cgeo/issues/7860#issuecomment-533871209

Bananeweizen:
retest this please

Bananeweizen:
I've accepted the SDK28 license directly in the Jenkins build script (as it was done for all earlier licenses already). That way you can already test, even though the new SDK will be installed repeatedly in the docker image.

moving-bits:
Created an issue regarding the tests failing due to missing package `android.test`: #7866

moving-bits:
This PR has grown too much into a mixup of compileSDK, targetSDK and other changes, IMHO, therefore creating a separate PR, which only switches targetSDK and adds the one lib ref to make it compilable for targetSDK28.

PR #7654, opened 2019-05-30
extend cache popup's hint view by personal note

moving-bits:
When pressing the cache popup's hint button: Do show the personal note as well (if given), not just the hint.


moving-bits:
retest this please

okainov:
Not tested yet but have you checked the case when personal note is big (I.e. like 1-2 full description screens)?

moving-bits:
The view is scrollable.

moving-bits:
Retest this please

SammysHP:
I wonder if we still need the popup as more and more information is integrated in it…

Lineflyer:
> I wonder if we still need the popup as more and more information is integrated in it…

Well yes, its still a quick possibility to get info while staying on the map (and of course its useful for unsaved caches to get basic information)

PR #7656, opened 2019-06-03
Fix #7647, OSM tiles, first step

### LINKED ISSUE #7647: OSM online tiles stopped working
Several user reports and I can confirm, that tiles are not loading for online OSM maps.
I assume the server either has a problem or the service is no longer available?!

rsudev:
- upgrade internal mapsforge v3 to latest version
- set user agent to 'cgeo' for tile downloads


main/src/cgeo/geocaching/maps/mapsforge/MapsforgeMapView.java
```diff
@@ -195,6 +196,11 @@ public void setMapSource() {
 
         final MapGenerator mapGenerator = MapGeneratorFactory.createMapGenerator(newMapType);
 
+        // Check for tile donwloading, set UA
```
jakedot: donwloading -> downloading

main/src/cgeo/geocaching/maps/mapsforge/MapsforgeMapSource.java
```diff
@@ -8,12 +8,15 @@
...
 import org.mapsforge.map.android.graphics.AndroidGraphicFactory;
 import org.mapsforge.map.layer.cache.TileCache;
+import org.mapsforge.map.layer.download.tilesource.AbstractTileSource;
 import org.mapsforge.map.layer.download.tilesource.OpenStreetMapMapnik;
 import org.mapsforge.map.model.MapViewPosition;
 import org.mapsforge.v3.android.maps.mapgenerator.MapGeneratorInternal;
 
 public class MapsforgeMapSource extends AbstractMapSource {
 
+    public static final String MAPNIK_TILE_DOWNLOAD_UA = "cgeo";
```
jakedot: Wasn't there a more complete User Agent agreed to in other issues/PRs? See cgeo#6611

SammysHP: `cgeo` is not a valid user agent. Please read this for details: https://stackoverflow.com/a/2601492

rsudev: If I read the RFC (https://tools.ietf.org/html/rfc7231) correctly is 'cgeo' a valid UA, while the addition of a version could be considered. Even though it would be the only real one to use if the UA string would have been used as it was originally intended, I do not suggest to replace it globally, but just for tile downloads from OSM (if the continue to let us do it in the first place).

jakedot: Why not use the UA SammysHP suggested in cgeo#6611? Are there technical reasons against it?

> What do you think about
> 
> ```
> Mozilla/5.0 (Linux; Android <android-release>) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.0.0 Safari/537.36 cgeo/<cgeo-release>
> ```

Lineflyer: It is IMHO the final target to use such kind of UA throughout the app.
But as this kind of change needs some testing, we change to this simple UA for mapsforge loading only first and later implement a common solution.
At least this would be my proposal.

rsudev: > Why not use the UA SammysHP suggested in #6611? Are there technical reasons against it?
There is no hard reason against it, besides wasting resources on transmission (tiny) and UA parsing on OSM side (likely tiny as well), but there are no reasons for using it in this context either. Did you check the discussing WRT UA-based blocking and the 'Tile Usage Policy' on OSM side (in the underlying issue #7647)?

jakedot: Ok, then this just needs to be tested, correct?

@SammysHP doesn't the SO link you posted also say the `cgeo` string is ok? the version seems to be optional, at least that is my reading of the EBNF ( [ ... ] denoting optional)?

@rsudev Sorry I didn't read it (and the policy) in detail before now. Yes, it seems that `cgeo` is the easiest solution, however the other points in the policy are dubiously fulfilled by c:geo as well (resource usage, attribution), and using the more complete UA is IMHO not faking another app's UA, if the `cgeo` string is included.

So I vote for provisionally accepting your PR (if I have any voting rights at all, being a newbie and all 😅)

The other points are addressed in the issue already aren't they...

SammysHP: @jakedot Yeah, you're right. It might be valid with respect to the RFC but far away from the from the usual `Mozilla/5.0`.

jakedot: tested it, seems to work fine for me

jakedot:
tested it, seems to work fine for me

Lineflyer:
retest this please

PR #7657, opened 2019-06-03
Rework new map to fix #7209 and missing caches

### LINKED ISSUE #7209: Waypoints of other lists leak onto map of a list
- Use mapsforge offline map (new version)

Example scenario 1:
- Have caches with waypoints in same area on two different lists
- Open list 1
- Open map for list 1
- At some point (e.g. after panning, zooming) you will also see the waypoints of caches on list 2 on the map of list 1

Example scenario 2:
- Having a list containing all caches in a certain area
- Having a list containing a subset of the other list (my ToDo)
- While mapping the second list, suddenly WP (but not the cache icons) from caches of the first list appear on the map
- WP limit is set to 30

Included above

Only waypoints of the current list should be shown

##### Version of c:geo used:
2018.11.29-NB

Yes, at least after some tries

Followup of discussion in #5900. See also comments of #5900 for more details.

rsudev:


rsudev:
I wouldn't mind if someone with interest for the map libraries, esp. new mapsforge would try to dig into my changes and give them a review (or just ask questions).
I have used this change now for a while (I think four or five caching occasions, including a two-day hiking multi last weekend) and found it working well. So if there don't come up any requests for waiting a bit, I'd merge it to master in the second half of the week.

Lineflyer:
Appreciate fixing #7209 as this might be the main reason of people falling back to v3. But what exactly do you mean with missing caches?

rsudev:
There were situations when caches loaded on the live map would suddenly disappear again and only reappear after moving the map a bit. I don't think there is a dedicated issue to that, as it is hard to describe and to reproduce (but I could produce it relatively easy).

Narrat:
Moving or zooming the map. Same for me with the disappearing caches. But didn't thought of a bug.

jakedot:
Yes, I also saw this and thought it was an artifact of optimization or something while changing the viewport. All the better if you could fix it!

Lineflyer:
@rsudev
As also planned by you, let me merge this to give it a test in the nightly over the coming long weekend.

PR #7659, opened 2019-06-03
optionally calculate and display route length on map (fix #6017)

### LINKED ISSUE #6017: Brouter: Distance shown is not the routed distance
- New map active
- Open live map
- Select a cache to show its popup
- Press the "route button"

The route is shown instead of the direct line (correct) but the distance shown on the upper right is not the length of the route. 
At this point I would expect to see the routing distance (if available) or we should ommit this information.

Found this today while routing to a cache on the other side of a bay (direct distance 50m, routing distance >1km)

moving-bits:
- calculate and display route length on map (in addition to direct distance) (optional) (fix #6017)
- add settings to enable/disable route length calculation and display

(implemented for GoogleMaps v1, Mapsforge v0.3 and NewMap)

### DIFF
--- a/main/res/layout/map_mapsforge_v6.xml
+++ b/main/res/layout/map_mapsforge_v6.xml
@@ -43,23 +43,35 @@
             android:maxLines="1" />
         
         <TextView
-            android:id="@+id/distance"
+            android:id="@+id/distance1"
+            android:layout_marginTop="-8dp"
             android:layout_alignParentTop="true"
             android:layout_alignParentRight="true"
-            android:layout_marginTop="-8dp"
-            android:layout_marginRight="-8dp"
-            android:paddingTop="4dp"
             android:layout_width="100dp"
-            android:layout_height="30dp"
-            android:background="@drawable/icon_bcg"
-            android:ellipsize="end"
-            android:gravity="center_horizontal"
-            android:lines="1"
-            android:textColor="@color/text_icon"
-            android:textStyle="bold"
-            android:textSize="18sp"
-            android:visibility="gone"
-            android:maxLines="1" />
+            style="@style/map_distanceinfo" />
+
+        <TextView
+            android:id="@+id/distance1info"
+            android:layout_marginTop="-8dp"
+            android:layout_toLeftOf="@id/distance1"
+            android:layout_width="wrap_content"
+            style="@style/map_distanceinfo" />
+
+        <TextView
+            android:id="@+id/distance2"
+            android:layout_marginTop="0dp"
+            android:layout_below="@id/distance1"
+            android:layout_alignParentRight="true"
+            android:layout_width="100dp"
+            style="@style/map_distanceinfo" />
+
+        <TextView
+            android:id="@+id/distance2info"
+            android:layout_marginTop="0dp"
+            android:layout_below="@id/distance1info"
+            android:layout_toLeftOf="@id/distance2"
+            android:layout_width="wrap_content"
+            style="@style/map_distanceinfo" />
 
         <include layout="@layout/map_progressbar" />
 
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -80,6 +80,7 @@
     <string translatable="false" name="pref_renderthemepath">renderthemepath</string>
     <string translatable="false" name="pref_showwaypointsthreshold">waypointsthreshold</string>
     <string translatable="false" name="pref_brouterDistanceThreshold">brouterDistanceThreshold</string>
+    <string translatable="false" name="pref_brouterShowBothDistances">pref_brouterShowBothDistances</string>
     <string translatable="false" name="pref_maptrail">maptrail</string>
     <string translatable="false" name="pref_dot_mode">dot_mode</string>
     <string translatable="false" name="pref_defaultNavigationTool">defaultNavigationTool</string>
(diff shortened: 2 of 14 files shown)


Lineflyer:
> add settings to enable/disable route length calculation and display

Do we need this?
if the users has the routing app installed, he probably want to see the route and the corresponding info.
At least it should be defaulted to "On" in this case IMHO.

Lineflyer:
retest this please

jakedot:
> > add settings to enable/disable route length calculation and display
> 
> Do we need this?
> if the users has the routing app installed, he probably want to see the route and the corresponding info.
> At least it should be defaulted to "On" in this case IMHO.

I would not display two different labels by default, it might be confusing. But maybe a setting to force displaying straight line distance instead of the route distance?

Or did i misunderstand the PR wording and this is exactly what it provides?

Lineflyer:
> I would not display two different labels by default, it might be confusing. But maybe a setting to force displaying straight line distance instead of the route distance?

I would also tend to show the distance which is matching the route shown on display, i.e. when in straight line mode (or above routing distance limit) show the straight line distance, when in routing line mode show the routed distance. 
No need for additional menu items and not consuming additional screen space on the map.

jakedot:
> I would not display two different labels by default, it might be confusing.

Maybe I should be explicit about to what I'm referring to here, it's about the preview image in the issue #6017 which [I'll link to here](https://github.com/cgeo/cgeo/issues/6017#issuecomment-495977010). 

So we're both on the same page, there should only be one distance shown and it should match the current setting?

> Or did i misunderstand the PR wording and this is exactly what it provides?

bekuno:
We know already the direct distance without the routing.
Therfore the calculated distance should be shown everytime.
For me it is fine to show both values as posted in the screenshot. Maybe a symbol before the value indicating the type could be helpful.

Lineflyer:
From a users perspective:
Cureently I do see a red line and a distance and I assume this is the length of the red line.
In #6017 the bug is, that the distance is not the length of the red line and that should be fixed.

I do not see any benefit in showing both and potentially need to introduce icons or text to explain them. 
If we just show the length of the red line (depending on the routing mode selected by the user), that should be straight forward without need of any more complication.

moving-bits:
Showing both distances was part of the discussion in #6017, but I tend to follow the discussion here that this may be confusing and not necessary. (At least we should try it this way and test for edge cases like what happens if brouter does not find a sensible route.)

---

So to summarize the discussion: I will change this PR to
- only show one distance info on the map - either straight distance (if no routing involved) or routing path's calculated distance (if available)
- remove the settings option (so if brouter is installed you'll automatically get the calculated distance)

rsudev:
I would strongly vote for showing both distances normally (because that would be the information I personally want to see) and indicate the meaning of the distance by a simple symbol (to dots connected by a diagonal straight line for the first, four-something dots connected by a curvy line for the second).
Of course should the second value only be displayed if routing is active.

Lineflyer:
OK, so now we have 2 by 1 votes from the team for showing both distances AFAICS. So I give up, but at least it should show symbols as mentioned by @rsudev

moving-bits:
The symbols mentioned by @rsudev are a good idea - I will try to create some visible but non intrusive ones.

As for the discussion whether to show only the calculated or both distances: We could display only the calculated distance by default, and use a setting switch to show both.

jakedot:
> We could display only the calculated distance by default, and use a setting switch to show both.
I think it would be a good compromise.

Lineflyer:
> As for the discussion whether to show only the calculated or both distances: We could display only the calculated distance by default, and use a setting switch to show both.

Good compromise.

I set the "Do not merge" label for now as you are still working on this PR.

moving-bits:
Removing the "WIP" label, as the PR is completed and the only failing thing is a test unrelated to this change (`cgeo.geocaching.connector.oc.OkapiClientTest.testGetOCCache`).

moving-bits:
One additional thought. Current implementation is like this:

image

a) Do we really need the symbols?
- If there are two distances, the higher value always stands for the calculated route's length.
- If there is only one distance shown it's the length of the red routing line, whether it's a straight line or the specific route.

b) Even if we answer a) with "yes": Do we always need to show the symbols or wouldn't it be enough to show the symbols only when there are two distances shown? (Otherwise the outline of the red routing line explains whether this is a straight distance or a calculated route distance.)

jakedot:
I would instinctively say a) yes b) when displaying only one distance it
depends: if it is made configurable (which iiuc is, by selecting direct
line vs some calculated route) I would still prefer to have it visible.
Otherwise, leaving it out would probably be ok.

moving-bits:
> b) when displaying only one distance it depends: if it is made configurable (which iiuc is, by selecting direct line vs some calculated route) I would still prefer to have it visible. Otherwise, leaving it out would probably be ok.

Configuration has changed throughout this PR: When you have brouter installed, the calculated route's distance is shown automatically. You only have the choice to see the straight distance additionally.
(If you don't have brouter installed or the distance is farther than the set max distance for brouter you'll get the straight distance only, but then the red line also is straight.)

jakedot:
Ah, ok, I didn't catch that part of changes. Then I'd say i doesn't need
symbol for single distance.

moving-bits:
retest this please

Lineflyer:
@moving-bits Do you consider this ready to be merged?

I am fine so far with how it looks and works.
We can fine tune things like symbols later on if needed.

moving-bits:
> @moving-bits Do you consider this ready to be merged?

Yes, the PR is feature complete. AFAICS the failed builds are due to the oc.us problem and the Geocoder, nothing related to the PR itself.

PR #7669, opened 2019-06-09
Fix #2558, Filter waypoints for owned caches

### LINKED ISSUE #2558: Waypoints of hidden owned caches shown on live map
Today I have seen that the live map behavior changed.
I am not quite sure about the exact behavior, however I can see the following which is unusual:
- I activated "Hide own/found"
- I use to have my owned caches (and some found as well) saved on list

Today I have seen that the waypoints of my owned caches are shown on live map although the cache itself is not shown.
I am sure, that this constellation never happend before, however I am not sure if we previouly hide also owned/found caches stored locally.
However if the cache is hidden due to setting/filter its waypoints should be hidden as well.

rsudev:
add an cache-owner based filter to waypoint overlay for mapsforge v6

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractActivity.java
@@ -14,7 +14,7 @@ import cgeo.geocaching.utils.ClipboardUtils;
 import cgeo.geocaching.utils.EditUtils;
 import cgeo.geocaching.utils.HtmlUtils;
 import cgeo.geocaching.utils.Log;
-import cgeo.geocaching.utils.MapUtils;
+import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.TranslationUtils;
 
@@ -251,7 +251,7 @@ public abstract class AbstractActivity extends AppCompatActivity implements IAbs
         final ActionBar actionBar = getSupportActionBar();
         if (actionBar != null) {
             actionBar.setDisplayShowHomeEnabled(true);
-            actionBar.setIcon(MapUtils.getCacheMarker(getResources(), cache, CacheListType.OFFLINE));
+            actionBar.setIcon(MapMarkerUtils.getCacheMarker(getResources(), cache, CacheListType.OFFLINE));
         }
     }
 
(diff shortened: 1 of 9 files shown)


rsudev:
The failing test seems to be associated with website issues at oc.us...

Lineflyer:
OC.us is down for some days due to server migration

rsudev:
I was just looking over there, but didn't see any anouncement. So I opened a bug report...

rsudev:
If somebody would like to do a code-review (even not right away but perhaps the next days), please notify here. Otherwise I'll merge tomorrow or Wednesday.

jakedot:
sure, I'll have a look, but probably not this evening, tomorrow would be a
good time.

jakedot:
I want to start on related issues (missing distance circles et al) so your
code will be a good starting point.

Lineflyer:
@rsudev 
This is only fixing it for new mapsforage, isn't it?

Do you think it needs to be tested performance wise (thinking of your comment in the issue, that it might affect performance) ? If yes, what would be a good test?  List(s) with >1000 caches?

jakedot:
I looked over the code and the changes look very promising - clearer code and may be better flow as well - but I didn't look at the actual change in detail.

rsudev:
@Lineflyer: the driving entity for the amount of work now to be done is the number of waypoints in the viewport that would be shown. I would assume that this mostly is in the range < 100, so I do not strongly expect issues.
A limits-checking test case could be, to have ~ 50 caches, each with one waypoint. The waypoints have a sufficient distance to the cache, so that it is outside the extended viewport, and all waypoints are within the viewed (live map) area, but not too many caches, so that waypoints are displayed.
It is necessary to have the flag 'Don't show own/found caches' active (otherwise a shortcut is taken). Whether the waypoints are shown or suppressed does not matter.

Lineflyer:
retest this please

rsudev:
> It is necessary to have the flag 'Don't show own/found caches' active (otherwise a shortcut is taken). Whether the waypoints are shown or suppressed does not matter.

I not be more precise here. It doesn't matter if they are shown or 'suppressed as they are attached to an owned cache'. Suppression due to found status is done on database level as before, so the performance was not changed.

PR #7670, opened 2019-06-09
add tts navigation to cache and waypoint popups

moving-bits:
We have tts navigation already in CompassActivity. This PR adds the same to the map's cache and waypoint popups. Tested with GMv1, MFv0.3 and NewMap.

### DIFF
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -20,12 +20,29 @@
             android:layout_height="wrap_content"
             android:orientation="vertical" >
 
-            <LinearLayout
-                android:id="@+id/details_list"
-                android:layout_width="fill_parent"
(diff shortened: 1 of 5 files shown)


jakedot:
I'd give this PR a heart&rocket if I had proper access to github on this
device as well as test it if I could install any APKs or even play store
apps. It isn't possible because of broken storage config / partitioning I
can't fix without a PC... But have this email comment of approval!

Lineflyer:
See https://github.com/cgeo/cgeo/issues/417#issuecomment-500795837 for further discussion.

Lineflyer:
As this is confirmed to be ready and distinct from a beep notification feature according to https://github.com/cgeo/cgeo/issues/417#issuecomment-500817265 I will merge this for further testing.

PR #7672, opened 2019-06-10
Temporay disable oc.us website check

Lineflyer:
Downtime was expected one day, but now already nearly a week


Lineflyer:
retest this please

PR #7680, opened 2019-06-14
tts menu entry instead of button (fix #7674)

### LINKED ISSUE #7674: Use menu entry for TTS in popup rather than button
Having just tested the new TTS feature on the popup #7670 I am a bit unsure about the activation method:
The activation via the button is simple and easy but maybe still not the best way, because:
- It clutters the relatively small popup window
- The TTS feature might be relatively rarely used compared to the other frequently used buttons like Details, List Management and Refresh.
- It might happen easily to accidentally activate it
- Its not placed consistent with other navigation options from that point

Due to the reasons mentioned above I would suggest to have it as menu option behind the three-dot-button instead (same as in compass activity).

It would be even more consistent to instead have it implemented as an additional generic navigation method. This would have the additional benefit, that it becomes available in all places where the Navigation menu is offered. But I guess, this might require some thinking about a user interface as there has to be a method to stop the navigation.

@moving-bits 
-

moving-bits:
- switch to menu entry for tts on cache and waypoint popups instead of button
- add tts cache details page

Fixes #7674

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1527,6 +1527,7 @@
 
     <!-- text-to-speech for compass view -->
     <string name="tts_service">Talking compass</string>
+    <string name="tts_toggle">Toggle talking</string>
     <string name="tts_start">Start talking</string>
     <string name="tts_stop">Stop talking</string>
     <string name="tts_started">Started speech</string>
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -20,29 +20,12 @@
             android:layout_height="wrap_content"
             android:orientation="vertical" >
 
-            <RelativeLayout
-                android:layout_width="match_parent"
-                android:layout_height="wrap_content">
-
-                <LinearLayout
-                    android:id="@+id/details_list"
-                    android:layout_width="fill_parent"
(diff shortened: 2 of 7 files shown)


moving-bits:
retest this please

main/res/values/strings.xml
```diff
@@ -1527,6 +1527,7 @@
 
     <!-- text-to-speech for compass view -->
     <string name="tts_service">Talking compass</string>
+    <string name="tts_toggle">Toggle talking</string>
```
Lineflyer: Can't we reuse the `tts_start` and `tts_stop` at this point. If you look into the compass it uses these two strings (depending on current status).

moving-bits: Or use "toggle" for compass as well - forgot to change it there.
I don't see a big additional benefit by providing two toggling menu entries (as there is a toast shown which informs the user whether tts started or ended), and it inflates the code unnecessarily IMHO.

Lineflyer: Agreed. You might change it in the compass as well.
The only advantage I see in using the current strings is, that we do not need to wait for translations...but thats minor IMHO.

PR #7690, opened 2019-06-20
proximity notifications

moving-bits:
adds two types of proximity notifications:

a) general proximity notification
- available in map view
- checks distance to closest waypoint or cache
- beeps once when closest distance is below threshold (no repeats)
- is reset when distance > threshold

(a) is the "Garmin like" notification variant.)

b) specific proximity notification
- available in compass view and in map popups for cache or waypoint
- checks distance to currently active geopoint (cache / waypoint)
- supports two notification signals, bound to two different distances (far / near)
- sound is repeated after 10s (far) / 5s (near)
- is reset when distance > far threshold

Settings are on Settings => Map => Proximity notification:
- enable / disable general proximity notification
- enable / disable specific proximity notification
- set far and near distances

Implemented for
- NewMap and CGeoMap (general notification)
- cache popup, waypoint popup and compass view (specific notification)

### DIFF
--- a/main/res/menu/cache_options.xml
+++ b/main/res/menu/cache_options.xml
@@ -17,12 +17,6 @@
         app:showAsAction="ifRoom">
     </item>
     <item
-        android:id="@+id/menu_tts_toggle"
-        android:icon="@drawable/ic_menu_start_conversation"
-        android:title="@string/tts_toggle"
-        app:showAsAction="ifRoom">
-    </item>
-    <item
         android:id="@+id/menu_log_visit_offline"
         android:icon="@drawable/ic_menu_edit"
         android:title="@string/cache_menu_visit_offline"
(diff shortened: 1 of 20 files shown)


Lineflyer:
FYI #7682 has been merged.

moving-bits:
> FYI #7682 has been merged.

Thanks for merging and for notifying me.

Lineflyer:
@moving-bits 
Do you consider ready for merge and test. I am out caching a trail tomorrow, so would be perfect test conditions.

Lineflyer:
> beeps once when closest distance is below threshold (no repeats)
> is reset when distance > threshold

What happens if the next WP is below threshold before the first one is above threshold?

moving-bits:
> What happens if the next WP is below threshold before the first one is above threshold?

So far only the distance gets reported by the cache lists, so in this case the second waypoint would take over without a new notification. Notifications resume when all waypoints and caches are above threshold.

For this scenario you could use the specific proximity notification and manually select the desired waypoint.

Or you could set the notification limits low enough to not run into this scenario :-)

Lineflyer:
> Or you could set the notification limits low enough to not run into this scenario :-)

True :)

moving-bits:
> @moving-bits
> Do you consider ready for merge and test. I am out caching a trail tomorrow, so would be perfect test conditions.

Should be ok for field test IMHO, but I've tested in emulator only so far. Caches work, waypoints in NewMap should be ok as well (as I have explicitly included its overlay), but I'm unsure (and have not yet tested) waypoints in CGeoMap (Google v1 and MF v0.3).

PR #7694, opened 2019-06-22
switch tts in compass view to toggle

moving-bits:
(a followup to https://github.com/cgeo/cgeo/pull/7680#discussion_r296318035)


Lineflyer:
We could also remove  the now unused strings for on/off, can we?

moving-bits:
> We could also remove the now unused strings for on/off, can we?

For now I've left them in intentionally, in case someone wants to have the old handling back. Don't easily want to get rid of translated strings... But thinking of it, we do have a version control system... :-)

Issue #2386, opened 2013-01-12
Replace list of contributors


SammysHP:
Because the list of contributors in our "about" view gets longer and less maintainable I suggest to replace it with the output of

```
$ git shortlog -s -n
```

Currently this is

```
   953  Samuel Tardieu
   922  Bananeweizen
   291  SammysHP
   247  blafoo
   219  rsudev
   181  Lars
   144  mucek4
    98  campbeb
    85  Marco Jacob
    69  bananeweizen
    50  Karsten Priegnitz
    41  RomNexus6
    40  YraFyra
    39  Pavol Babincak
    38  keith.paterson
    35  Torsten Keil
    32  Portree-Kid
    29  Stephan Merker
    28  cachapa
    24  koem
    23  Andreas Scharf
    23  Joachim Wilke
    21  Sven Karsten Greiner
    19  Lineflyer
    18  c:geo
    17  Rainer
    14  Brian
    14  Florian Moesch
    14  Michiel Korthuis
    13  Gerald Barker
    13  M@®k
    13  Ralf Hubert
     8  Nelson Correia
     8  triakcz
     7  Ondřej Kunc
     6  Heiko W. Rupp
     5  BudBundi
     5  LudoO
     5  Ray
     4  Bazsy
     4  Libor Tvrdík
     4  Ricardo Seromenho
     4  italian-geocacher
     3  Bc. Jan Kaláb
     3  Sven Greiner
     3  inkantis
     2  BestHafr
     2  Buuhuu
     2  David Hrbáč
     2  Mieczyslaw
     2  ludoo0d0a
     2  patrick
     2  seromenho
     2  teschi
     1  Fernando Coello
     1  Francesco Sacchi
     1  Gacksi
     1  Kai Stempfle
     1  Marc
     1  Matthias Büchner
     1  Michael Keppler
     1  Stefan Breunig
     1  jenda^^
     1  mieczyslaw
     1  rasch
     1  runehl
     1  serenity182
     1  tm0702
     1  unknown

```

This can be done on Jenkins as we already process dynamic content. A few duplicate entries (because of different names as author) should be ok.

akkatracker:
Any update on this, was going through the outdated contributors list in the app and was wondering if it had been brought up yet.

SammysHP:
I suggest to

- manually filter the current list of contributors in c:geo and remove everyone that is not obviously included in git (we can remove people with the addition "translation" because we do not attribute translators anymore)
[...]

PR #7696, opened 2019-06-23
update contributors page

moving-bits:
Our contributors page has become out of sync lately. I updated the page, adding all contributors from 2018/19 according to github's contributors page, using their respective github user names.

### DIFF
--- a/main/res/layout/about_contributors_page.xml
+++ b/main/res/layout/about_contributors_page.xml
@@ -12,18 +12,76 @@
         android:layout_height="wrap_content"
         android:orientation="vertical" >
 
+        <RelativeLayout style="@style/separator_horizontal_layout" >
+            <View style="@style/separator_horizontal" />
+            <TextView
+                style="@style/separator_horizontal_headline"
+                android:text="carnero" />
+        </RelativeLayout>
+
         <TextView
-            android:id="@+id/contributors"
+            style="@style/about_contributors"
+            android:id="@+id/about_carnerodetails"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_gravity="left"
-            android:layout_margin="7dip"
-            android:linksClickable="true"
-            android:paddingLeft="3dip"
-            android:text="@string/contributors"
-            android:textColor="?text_color"
-            android:textColorLink="?text_color_link"
-            android:textSize="14sp" />
+            android:text="@string/about_contributors_carnerodetails" />
+
+        <RelativeLayout style="@style/separator_horizontal_layout" >
+            <View style="@style/separator_horizontal" />
+            <TextView
+                style="@style/separator_horizontal_headline"
+                android:text="@string/about_contributors_recent" />
+        </RelativeLayout>
+
+        <TextView
+            style="@style/about_contributors"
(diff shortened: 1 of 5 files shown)


bekuno:
I'm not sure if we really should create a list for active contributors.

By the way, at least kumy & SammysHP should be in the actice contributor list.
Also other user where involved in this time, eg. following5.

@Lineflyer et al,
A more generally question: some users are not listed with there github names, e.g. rsudev. Was this planned?

I think that we could also thank to issue-creator and other supporter.

And at least. This is no longer a contributors page. ("strings_not_translatable")
How about to make the page with (translatable) headline?

SammysHP:
See also #2386.

moving-bits:
> See also #2386.

Nice stats on https://www.openhub.net, thanks for the link.
But I'm not too happy about linking to it, as it would create an additional dependency to a separate server (and one which currently has an expired security certificate which creates big colorful warning messages blocking access at first in the browsers I use...).

So maybe we should look for a different solution, preferably one that is included in c:geo, even better an automated one. (On the other hand: as new contributors do not emerge too frequently it might be enough to update this list once per year manually.)

moving-bits:
> I'm not sure if we really should create a list for active contributors.

The current list is pretty much outdated, and if I understood the comments right, it's so since years. So I started to add at least the contributors from the last year and this year and put together a list of "2018/19 contributors" (without deleting the names of the other contributors).

> By the way, at least kumy & SammysHP should be in the actice contributor list.

I started by looking at the github commits from 18 and 19 in the c:geo repo, but for sure other contributors can be added as well.

> Also other user where involved in this time, eg. following5.

is included already

> I think that we could also thank to issue-creator and other supporter.

will add

> And at least. This is no longer a contributors page. ("strings_not_translatable")
> How about to make the page with (translatable) headline?

You mean the headline only, or subheadings as well? The tab label is already translated.

bekuno:
> > And at least. This is no longer a contributors page. ("strings_not_translatable")
> > How about to make the page with (translatable) headline?
> 
> You mean the headline only, or subheadings as well? The tab label is already translated.

Also subheadings, if possible.

Lineflyer:
> @Lineflyer et al,
> A more generally question: some users are not listed with there github names, e.g. rsudev. Was this planned?

In the passt users added theirselves sometimes, or added such kind of links/names after they have been added. So for the existing it is fine, but we might also remove that if we e.g. decide to apply a commong source / automation of this.

moving-bits:
> > > And at least. This is no longer a contributors page. ("strings_not_translatable")
> > > How about to make the page with (translatable) headline?
> > 
> > You mean the headline only, or subheadings as well? The tab label is already translated.
> 
> Also subheadings, if possible.

Done - subheadings, roles and parts of the text are translatable now.
Layout for the parts is the same as on the first "about c:geo" page (version / help).

Lineflyer:
I appreciate, that the contributors list gets updated. While not having a fixed process for it, its IMHO quite OK to decide case by case to add/modify the list for now.

I will therefore merge this and doublecheck afterward based on my knowledge of the project history and maybe modify something.  Also all others: If you miss someone feel free to submit your change.

PR #7698, opened 2019-06-23
more prominent placing of help in about pages + add titles to actionbar

moving-bits:
We have seven subpages in about c:geo which leads to much scrolling and some information being deeply buried. This PR 
- moves the help info to the opening version page, making help info more prominent,
- and places the subpage's title into the action bar to make it more visible.

image

Some more ideas: (TBD)
- one could even leave out "About c:geo" in title at all (or move it to the end)
- the "support us" section is pretty prominent - if we decide that the help links are more important, "support us" could be moved after the help section (another "website" link is part of the help section already)
- c:geo logo could be made a little smaller


Lineflyer:
A real good idea IMHO to have the links on the first page.

Issue #7692, opened 2019-06-21
Images are not stored/restored with backup


okainov:
**To Reproduce:**
1. Save some caches with images
2. Disconnect internet
3. Make a backup
4. Clear database/reinstall cgeo
5. Restore backup
6. Progress reported as succesfull, however no images restored.

Probably images are not backed up at all for some reason.

Lineflyer:
The backup function only performs a backup of the sql databsse, but not of the additional cache data.
Additionally the geocache data dir will be removed by the system upon deinstall.

If you want to keep the additional data you have to manually save it somewhere before deinstall at the moment.

Its not part of the backup as it can be retrieved again by refreshing the caches after restore of the backup, and as it might be several GB depending on the amount of pictures/caches stored.

okainov:
>and as it might be several GB depending on the amount of pictures/caches stored.

Yes, exactly, that's why I want it to be stored with backup :)

SammysHP:
You can create a database backup as the database might not be easily accessible. But there shouldn't be a problem with additional data because it is stored in the same area as the database backup.

Lineflyer:
Yes, AFAIR this was the reason we did it like that. 
For the additional data we are using the system default file directories now, which should even make it compatible for clever third party backup tools.

okainov:
@SammysHP  @Lineflyer as a developer I fully understand you. However as a user I would expect backup to make a backup. Full stop. Either you implement two versions "light backup: text only" and "full backup: with images", or make one full - TBH it doesn't really matter. Saying "users can make backup by going to some filesystem folder and copying it manually" - you as developers should know it better than anyone else - not a single normal user ever would do this, it's too complicated workflow. You're not making app for developers but for usual people ;)

Lineflyer:
I admit it would be convenient. What would you expect to be the result of such kind of backup. A zipped archive of the additional geocache data?

Some of my devices would not be able to handle this, as they do not have enough free disk to store such thing. That would then also require to add some other storing options for such backups IMHO.

okainov:
>A zipped archive of the additional geocache data?

As a user I don't really care :) Yes, I would expect some blob-file like "cgeo_backup_20190101.tar.gz" or something like this. Maybe zip, maybe some binary format. Just one file.

> That would then also require to add some other storing options for such backups IMHO.

That's what I said - it could be just two versions "lite backup" (same as now) and "full backup" with some warning like "Full backup may take up to several GBs. Please make sure you have free space".

[...]

Issue #7665, opened 2019-06-07
Next feature release - 2019.08.18


Lineflyer:
Quote from #7661:

After a review of recent changes on both branches I suggest the following:

- ~Release from current `release` to have the fix for #7647 out on the market asap~
- Merge the following PRs (if considered OK) to `master` next days:
  - [x] PR #7659
  - [x] PR #7657
  - [x] PR #7641 
  - [x] PR #7669
  - @cgeo/team Any others?
- [x] Some days smoke testing with nightly builds
- [x] Merge `master` to `release` and publish beta version

Doing this way we can have a new release (with minimum risk and fix for OSM) asap and the longest possible beta testing period for the reviewed changes and a new playground for fresh changes.

jakedot:
When do you want to do the release? Ok, the listed PRs should probably all be done before the release, but maybe I can patch in the circle stuff in time if it's supposed to be, say, mid to end of next week.

moving-bits:
As for the last step with the "more experimental" PR: What about merging the GMv2 branch as well, to bring it more into our focus and start fixing the open issues?

Lineflyer:
> As for the last step with the "more experimental" PR: What about merging the GMv2 branch as well, to bring it more into our focus and start fixing the open issues?

Well, why not if someone feels a bit engaged to work on it and fix the remaining bugs.
But then in the last step (as described above) to maximize our time for getting it clean.

Lineflyer:
Added PR #7669 to list of PRs to be merged before branching.

What about PRs #7612, #7575, #7441 ?
I labeled them with "Feedback required" and assigned the team members which have outstanding feedback/review there.

moving-bits:
Please replace #7625 by #7690 in your list.

Lineflyer:
All PR mentioned in the initial posting (and some more) have now been merged to `master`.
I suggest to wait for some days of testing in nightlies and maybe fine tune some things (as e.g. mentioned in #7699) and then merge to `release` to be able to issue a beta release.

I did not follow recent activities (e.g. regarding OC attributes), so feel free to mention here if things should come into `master` before branching.

Lineflyer:
Just FYI:
I merged some recent PRs as I am currently on a longer bike trip and have the chance to use and test the newest nightlies.

After my return I will review and propose the further progress on next release and branching.

moving-bits:
@Lineflyer  Thanks & have a nice trip!

Lineflyer:
[...]

PR #7707, opened 2019-06-29
backup / restore program settings (shared prefs)

moving-bits:
backup / restore program settings (shared prefs)

- shared prefs are saved as XML file in downloads dir
- with or without account related data
- backup / restore as separate settings page
- backup / restore as menu entry in main screen

This is to ease the process of uninstalling and reinstalling c:geo, e. g. when switching between nightly and release version.

The new backup settings screen:
image
If we are going to add a backup function for cache images etc. (see #7692) it could be added here as well.


SammysHP:
I think we should mention more explicit that the *full* backup contains unencrypted passwords.

moving-bits:
> I think we should mention more explicit that the _full_ backup contains unencrypted passwords.

There is a dialog box specifically for this, which opens only when you chose the full backup.

Lineflyer:
Thanks for your PR @moving-bits 
Looking at the plan phrased in #7665 I suggest to merge only after next branching from `master` to `release` which is hopefully not so far away from now.

moving-bits:
This would be ok for me. - On the other hand: Merging this PR to master shouldn't be too risky - core backup database functionality is untouched (except moving it to a separate settings/preferences screen), and the new backup/restore program settings is following standards (and even if this would not work it would be no regression).

PR #7711, opened 2019-07-03
Fix #7710 - Use MultiLine text input

### LINKED ISSUE #7710: Personal note in calculator does not support new line
**Describe the bug:**
When entering notes into the waypoint calculator its not possible to add a new line / line break.
This draws this rather large text field almost unusable.
Not sure whether this is a regression from a recent change, but I only just now noticed and I would have noticed earlier if this problem was there from start of that feature.

**To Reproduce:**
1. Open a cache
2. open waypoint tab
3. Create new waypoint
4. Click on coordinates
5. Select "calculate coordinates"
6. Go into the note text field

You can enter text but no line feed
If you copy several lines of text into that field, the line feeds are removed and all text appears in one line

Text field should support line feed and also show copied multi line text in the original formatting

**Version of c:geo used:**
2019.07.02-NB

Yes, since some weeks

Lineflyer:


Lineflyer:
Reference:
https://developer.android.com/reference/android/widget/TextView.html#attr_android:inputType

main/res/layout-w1200dp/coordinatescalculate_dialog.xml
```diff
@@ -79,7 +79,7 @@
                     android:overScrollMode="always"
                     android:scrollbarStyle="insideInset"
                     android:scrollbars="vertical"
-                    android:inputType="textAutoCorrect"
+                    android:inputType="textMultiLine"
```
bekuno: Did you see any advantage with textAutoCorrect compared with plain text?
You can combine the value textMultiLine with other text entries, e.g.
"textMultiLine|textAutoCorrect"
But yy suggestion here is "textMultiLine|textCapSentences", because the text from the coordinate calculator is copied in the user_note field of the editwaypoint_activity, which is defined in this way.

main/res/layout/coordinatescalculate_dialog.xml
```diff
@@ -58,7 +58,7 @@
                 android:overScrollMode="always"
                 android:scrollbarStyle="insideInset"
                 android:scrollbars="vertical"
-                android:inputType="textAutoCorrect"
+                android:inputType="textMultiLine"
```
bekuno: The same as before.

main/res/layout-w1200dp/coordinatescalculate_dialog.xml
Lineflyer: Thanks for the hint. Will merge here first and append your suggestion next week when back at PC. Or feel free to add a PR yourself.

PR #7722, opened 2019-07-08
Update coordinatescalculate_dialog.xml

### LINKED ISSUE #7710: Personal note in calculator does not support new line

bekuno:
fixes #7710 

as discussed in #7711 - suggestion: the text from the coordinate calculator is copied in the user_note field of the editwaypoint_activity, which is defined in this way.

### DIFF
--- a/main/res/layout-w1200dp/coordinatescalculate_dialog.xml
+++ b/main/res/layout-w1200dp/coordinatescalculate_dialog.xml
@@ -79,7 +79,7 @@
                     android:overScrollMode="always"
                     android:scrollbarStyle="insideInset"
                     android:scrollbars="vertical"
-                    android:inputType="textMultiLine"
+                    android:inputType="textMultiLine|textCapSentences"
                     android:hint="@string/waypointcalc_notes_prompt" />
             </LinearLayout>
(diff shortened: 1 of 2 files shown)


Lineflyer:
Thanks, that speeds things up :)

Lineflyer:
Addition: But IMHO both layout should use the same type, shouldnt they?

Lineflyer:
retest this please

Lineflyer:
retest this please

PR #7731, opened 2019-07-19
respect silent mode settings for proximity notification (fix #7724)

### LINKED ISSUE #7724: Temporary disable proximity tone or make it follow device settings
Currently the proximity notification is always enabled, once its enabled in the settings.
In some situations users might want to temporary disable the notification (e.g. in situation where silence is required). 

This could either be done by some kind of silence toggle or (maybe preferably) by letting is follow some device global setting (like ringtone off or "do not disturb" modes).

I have no idea whether this is feasible, and it for sure not an urgent feature/improvement request, but maybe more a base for further discussion if such function is desirable at all.

moving-bits:


SammysHP:
Sorry for the late reply, but I vote against this implementation. We use the media stream (STREAM_MUSIC) which is not related to the silent MODE (that would be STREAM_RING). My ring volume is always set to silent because I use my watch for notifications.

moving-bits:
Ok, thanks for clarifying the technical details. How to continue?

Without this PR c:geo continues signalling, even if the device is set to silent mode, which I think would not be what the user expects. (That's why I proposed to include this PR into the next release.)

With this PR I read out the current silent mode settings and act accordingly.

But I'm open to a different approach. What do you propose?
- I could use STREAM_RING for the notification to make it respect silent mode automatically. (Although it's not really a ringtone)
- I could use a different stream. (But which one is user configurable in terms of volume and still respects silent mode settings?)
- yet another approach
?

Lineflyer:
@SammysHP comment is valid, however I do also not have a better proposal other than having some kind of shortcut to switch it off temporary.

Now this is already part of `release` branch.
I think we should revert this PR on `release` and give some more time for discussing and agreeing whether and how to implement it.
@moving-bits Could you do me the favor and provide a reverting PR targeting on `release` ?

moving-bits:
@Lineflyer  will do so tonight when I'm back at my dev machine

SammysHP:
If a user opens a website with a video it might produce sound even if the ring volume is set to silent. On the other hand a user can reduce the media volume so that c:geo does not beep.

Lineflyer:
> If a user opens a website with a video it might produce sound even if the ring volume is set to silent. On the other hand a user can reduce the media volume so that c:geo does not beep.

Good argument. 
In the beginning it was not coupled to the media volume, only some days ago @moving-bits changed this accordingly. With that change we could IMHO almost close #7724 .

PR #7737, opened 2019-07-23
Update Mapsforge to 0.11.0

### LINKED ISSUE #7494: Upgrade to mapsforge 0.11.0

jonas-koeritz:
Fixes #7494 

Upgrade Mapsforge to 0.11.0.
Mapsforge 0.9.1 fails on Android 9 with targetSdk 28.

See https://github.com/mapsforge/mapsforge/commit/b7b07619a6389886734acd5bcb8ae9b14dfce8c2 for reference.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -303,7 +303,7 @@ dependencies {
     implementation 'com.caverock:androidsvg:1.3'
 
     // Mapsforge new version
-    def mapsforgeVersion = '0.9.1'
+    def mapsforgeVersion = '0.11.0'
     implementation "org.mapsforge:mapsforge-core:$mapsforgeVersion"
     implementation "org.mapsforge:mapsforge-map:$mapsforgeVersion"
     implementation "org.mapsforge:mapsforge-map-android:$mapsforgeVersion"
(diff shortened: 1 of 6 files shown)


Lineflyer:
Retest this please

jonas-koeritz:
I just have been catching a little and have experienced some delays while loading/zooming. I will have a look at the mapsforge project to see if anything has been changed that could influence this.

Lineflyer:
Did you see my videos in #7523 trying to explain the original difference/problem?

If you say we need Mf0.11 anyway for targetSDK28 we could also merge this and continue testing and debugging on `master`. This way I could also test based on the nightlies.

jonas-koeritz:
Yes I watched your videos, the problem seems to be still there. We should
test this in master and open a bug with mapsforge.

Lars <notifications@github.com> schrieb am Do., 25. Juli 2019, 23:24:

jonas-koeritz:
I implemented using the fast parent tile rendering and mapsforge is now rendering **much** faster.

jonas-koeritz:
I recorded a [video](https://drive.google.com/open?id=11WNXl1WuDMEU19ZPTkwSXSGGEm3XlJKr) of the live map performance of this branch. I'm using an offline map file as map source. The clip shows flawless zooming and panning.

Lineflyer:
Thanks @jonas-koeritz 
Lets merge and test in nightlies

Lineflyer:
@jonas-koeritz 
I used today nightly including this PR and it is really fast. Thanks for digging this out.
BTW: Is there any downside in using this fast parent tile rendering?

I also have another finding which for which I will open a dedicated issue.

jonas-koeritz:
I don't think there is a downside to this, I couldn't see the visual
difference that's supposed to be there. The parameter only affects so
called "parent tiles", if I understand mapsforge correctly the parent tile
is the one that you zoomed into, shown right before the new zoom level is
fully rendered.

Lars <notifications@github.com> schrieb am So., 28. Juli 2019, 21:41:

PR #7739, opened 2019-07-23
remove trailing spaces in CacheAttribute

bekuno:
Remove spaces at the end of lines created when splitting lines with multiple attributes into individual lines.


Lineflyer:
Thanks!!

PR #7743, opened 2019-07-24
put settings backup into cgeo's backup folder (fix #7736)

### LINKED ISSUE #7736: Setting backup location not shown and not findable
**To Reproduce:**
1. Go to Menu - Settings - Backup / Restore
2. Click on "Backup settings"

Backup is performed but only a small toast "Settings saved" is shown. There is however no corresponding file in `/cgeo/backup/` where I would have expected it.

We should inform the user about the location of the backup file to give him a chance to manually backup this file. Ideally we should use the same dialog which is used for the (old) backup of the cache database.
We should store the file in `/cgeo/backup/` where also the sqlite backup file is located.

Yes

moving-bits:
- use cgeo backup folder instead of general download dir
- replace toast messages with dialog boxes
- remove superflous Log.e() messages


moving-bits:
retest this please

PR #7753, opened 2019-07-30
Don't restrict the tile size of mapsforge layers. Fixes #7750.

### LINKED ISSUE #7750: Mapsforge - Zoom level inconsistent between online and offline
We have some similar issues listed here somewhere, however while using todays nightly after upgrade to mf0.11 (2019.07.28-NB) I recognized a sudden big difference between online tiles and offline tiles. 
I am pretty sure the following described difference is either new or has become worse with mf0.11:

While the offline map rendering is pretty much as it should be speaking about zoom level, detail level, etc. the online tiles from OSM have a problem with maximum zoom size and/or font size and/or tile resolution.

My assumption would be, that somehow the online tiles are shown with a wrong DPI or there is some kind of mismatch between tile area and zoom control.

That is an area of Cologne shown using mf0.11 and openandromaps offline file for Germany:
Screenshot_20190728-113130_cgeo

This is the same area of Cologne shown using mf0.11 and OSM-online tiles (just changes the map source and zoomed in as much as possible):
Screenshot_20190728-113059_cgeo
I would have expected to see the same detail level and zoomed area. If the server does not provide more detailed tiles, then it should be enough to just enlarge the existing tiles (they are shown quite small, i.e. I cannot read the text).

Google MapsV1 in c:geo with highest zoom as comparison:
Screenshot_20190728-113115_cgeo

OSM online tiles in Locus as comparison:
Screenshot_20190728-122408_Locus Map

jonas-koeritz:
Fixes #7750

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -663,7 +663,6 @@ public class NewMap extends AbstractActionBarActivity implements XmlRenderThemeM
 
         // Exchange layer
         if (newLayer != null) {
-            this.mapView.getModel().displayModel.setFixedTileSize(newLayer.getFixedTileSize());
             final MapZoomControls zoomControls = mapView.getMapZoomControls();
             zoomControls.setZoomLevelMax(newLayer.getZoomLevelMax());
             zoomControls.setZoomLevelMin(newLayer.getZoomLevelMin());


main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
```diff
@@ -663,7 +663,6 @@ private void switchTileLayer(final MapSource newSource) {
 
         // Exchange layer
         if (newLayer != null) {
-            this.mapView.getModel().displayModel.setFixedTileSize(newLayer.getFixedTileSize());
```
rsudev: Mentally this change doesn't align with the title for me.
Originally this line was required to get the switch between on- and offline mapsforge maps working (IIRC, it's been a while) without a crash.
I would be happy to better understand the background and effect.

jonas-koeritz:
When setting a fixed tile size the tiles get "pinned" to be 256px squares. This  results in my screen beeing about 5 tiles across, making everything unreadable when using osm. Removing that line seems to allow mapsforge to scale the tiles based on screen density and makes OSM Online maps and offline maps congruent again. Switching between maps does not lead to big jumps in zoom anymore and I can zoom closer. The problem this seems to fix is better described in #7750.

rsudev:
The problem is clear, I just didn't understand the fix. And I still would say that you avoid setting a fixed tile size for the model, not the layer. But that is not really relevant, it just puzzled me a bit (because I assume the online tiles are still of the same size, they just get mapped to larger display tiles through the now 'freed' model).
I will give it a spin, and if it looks as expected, I would even consider adding it to release.

jonas-koeritz:
You're right, I just change the display model. My confidence with the
mapsforge code is still not very high. Is there some documentation from
mapsforge somewhere?
That line has been created in a refactoring effort and the value has been 0
before, I don't know if there was any specific reasoning for this or if the
tile layer returned some other value before.
It seems to fix the issue, if it's working correctly on other phones than
mine I would call this closed.

rsudev <notifications@github.com> schrieb am Di., 30. Juli 2019, 17:22:

rsudev:
The documentation for mapsforge is mostly in the sources and some samples. I checked my original implementation and also saw the change from the refactoring (which I find a bit suspicious, but did not go deeper).
I'll give it a spin on my phone and set of emulators (with some range of Android versions and display densities)

jonas-koeritz:
Thanks for checking this out!

rsudev <notifications@github.com> schrieb am Di., 30. Juli 2019, 17:43:

rsudev:
On some devices it now looks a bit blurred, but nevertheless better readable. I am fine with it.
As it is based on the 0.11.0 change, I guess we need to keep it on master, right?

jonas-koeritz:
It looks blurred on my device too (Pixel 2 XL) it's stretching the tiles to
a 4k Display, Locus Looks similarly blurry on my screen.

rsudev <notifications@github.com> schrieb am Di., 30. Juli 2019, 21:52:

Lineflyer:
> As it is based on the 0.11.0 change, I guess we need to keep it on master, right?

That would also be my question: Can we change this on release where we use mf0.9?

Lineflyer:
@jonas-koeritz 
> That would also be my question: Can we change this on release where we use mf0.9?

As I might issue another beta version next weekend, we could think about including this change if compatible to mf0.9. Do you have an idea whether we can do that?

jonas-koeritz:
Should be compatible without a problem. Which branch should I do a pull
request from?

Lars <notifications@github.com> schrieb am Do., 1. Aug. 2019, 12:43:

Lineflyer:
`release`

jonas-koeritz:
superseeded by #7765

Issue #7460, opened 2019-04-03
RuntimeException related to CachePopup.onCreate


Lineflyer:
This problem report from Google Play has the highest occurence (besides the NPE on Resume issues, which are hopefully already fixed).

Version: 2019.02.23 (and also beta 2019.03.30-RC1)
```
java.lang.RuntimeException: 
 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2822)
 
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2897)
 
  at android.app.ActivityThread.-wrap11 (Unknown Source)
 
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1598)
 
  at android.os.Handler.dispatchMessage (Handler.java:105)
 
  at android.os.Looper.loop (Looper.java:251)
 
  at android.app.ActivityThread.main (ActivityThread.java:6572)
 
  at java.lang.reflect.Method.invoke (Native Method)
 
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
 
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:767)
Caused by: android.util.AndroidRuntimeException: 
 
  at android.support.v7.app.AppCompatDelegateImplV7.throwFeatureRequestIfSubDecorInstalled (AppCompatDelegateImplV7.java:1683)
 
  at android.support.v7.app.AppCompatDelegateImplV7.requestWindowFeature (AppCompatDelegateImplV7.java:585)
 
  at android.support.v7.app.AppCompatActivity.supportRequestWindowFeature (AppCompatActivity.java:228)
 
  at cgeo.geocaching.activity.AbstractActivity.onCreateCommon (AbstractActivity.java:134)
 
  at cgeo.geocaching.activity.AbstractActivity.onCreate (AbstractActivity.java:114)
 
  at cgeo.geocaching.CachePopup.onCreate (CachePopup.java:40)
 
  at android.app.Activity.performCreate (Activity.java:6975)
 
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1214)
 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2775)
```

Lineflyer:
Still happening on 2019.04.04 , 2019.04.08-RC and probably also with 2019.04.14-RC (will check later).

[...]

PR #7755, opened 2019-07-30
Mitigate crashes due to #7460

rsudev:
On some mobiles, requesting a deprecated indeterminate
progress does not lead to a 'denial', but to a crash instead.
Avoid making the whole app unusable, handle and log the
Exceptions instead.


rsudev:
@Lineflyer Fine with merging this on release?

Lineflyer:
Yes

PR #7759, opened 2019-07-31
Check if specific proximity notification is enabled in CompassActivity (fix #7756)

### LINKED ISSUE #7756: Dont have proximity notification active by default
Users feedback regarding our beta version 2019.07.25-RC on social media:

> I noticed yesterday when using the compass to navigate to a cache, it's now "beeping". I get a beep at a certain distance that switches to three beeps when I get closer? 

> Seit heute gibt mein Handy seltsame Geräusche von sich, wenn ich mich einem Cache nähere und den Kompass an habe. Kann ich das irgendwo abschalten?

Obviously the new proximity notification seems to be active by default.
IMHO we should change this to not being active by default. While this would be nice to make users aware of this feature, I also understand, that it might be annoying if your app behaves unexpected.

moving-bits:


Lineflyer:
retest this please

Issue #7761, opened 2019-08-01
support tags for caches


moving-bits:
Currently the main feature to organize caches is lists. This is working pretty well, but it is one-dimensional. I'm looking for an additional / orthogonal way for organizing caches, which can be combined with lists. Therefore I suggest to add a tagging feature to c:geo.

Tagging would work similar to managing lists:
- tags itself are a flat list of user defined names, no hierarchy
- tags and caches are n:m-related
- unlimited number of tags
- each cache can have an unlimited number of tags, tags can have an unlimited number of caches

The following things would have to be added:
- Cache list
  - [x] Filter cache list by a specific tag - _(done with PR #7764)_
  - [x] Show comma separated list of tags assigned to a cache - _(done with PR #7769)_
- Cache popup and cache detail view
  - [x] Manage assigned tags - _(included with lists management)_
  - [x] Show comma separated list of tags assigned to a cache - _(included with lists management)_
- Tag management (from cache list view)
  - [x] Add tag - _(included with lists management)_
  - [x] Rename tag - _(included with lists management)_
  - [x] Delete tag - _(included with lists management)_

(and maybe more)

Example use case:
- Being on vacation I've put all the caches I would like to do in a separate list.
- As there is currently a promo running for which you need to collect certain clues first, then specific jewels etc., I would like to be able to assign this information to the specific caches (eg: "which cache has which type of jewel") and filter accordingly.

(Today I'm helping myself by saving this information by using a couple of extra lists and additionally storing infos in my personal notes, but handling this is clumsy and personal notes get convoluted.)

(A follow-up feature could be to assign specific symbols to a tag and show this symbol in map view, but that's something for a separate issue - let's do the basic things first.) - _(done with PR #7775)_

rsudev:
Not sure why you want an additional categorisation category. The only thing missing to do what you describe is a filter for list membership.

Lineflyer:
See also #5540

moving-bits:
Thanks for your links. Hm, you are right - lists and tags basically would be the same, just independant from each other. So let's start with a list membership filter and try out how far I'll come with that.

moving-bits:
After the switch to use lists as tags, I created the above listed three PR to implement the different aspects. With those you can:

- create lists to tag caches according to your choice, maybe for "T5 only caches", "must do today" etc.,
- see those "tags" in your cache list
- filter your cache lists by any "tag"
- assign a visual marker to your "tag"
- and display this marker on your map

[...]

PR #7764, opened 2019-08-01
add list membership filter

moving-bits:
Related to (but does not fully implement) #7761: Add a list membership filter to be able to use lists as tag replacement.

grafik

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -401,6 +401,7 @@
     <string name="caches_filter_own_rating">With own rating</string>
     <string name="caches_filter_multi_listing">Multi listing with different state</string>
     <string name="caches_filter_own_waypoint">With user defined waypoints</string>
+    <string name="caches_filter_lists">Assigned to a list</string>
     <string name="caches_removing_from_history">Removing from History…</string>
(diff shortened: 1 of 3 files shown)


Lineflyer:

Lineflyer:
Question as (nearly) unable to test:
Is there a restriction or limit up to how many lists can be handled here?
I just heard from a user, that he has 560 lists in c:geo :)

moving-bits:
> Question as (nearly) unable to test:
> Is there a restriction or limit up to how many lists can be handled here?
> I just heard from a user, that he has 560 lists in c:geo :)

none that I know of (besides memory and CPU power limits, but if that's hitting you, then it should have done this already before in the list selection). Filtering itself is just by one criteria (a single list you selected).

Lineflyer:
> Filtering itself is just by one criteria (a single list you selected).

I mean the filter menu itself, which will then have 560 entries...

moving-bits:
> I mean the filter menu itself, which will then have 560 entries...

I understand, but the filter dialog is technically basically the same as the list selection dialog, so the problem should have arrived there already, if at all.

PR #7766, opened 2019-08-01
Fixes #7750

### LINKED ISSUE #7750: Mapsforge - Zoom level inconsistent between online and offline

jonas-koeritz:


Lineflyer:
Thanks @jonas-koeritz 
I will test it a bit.

PR #7768, opened 2019-08-01
add cache filters for stored within or before last 8h / 24h / 7d / 30d

### LINKED ISSUE #7498: Filter by time of last update (to limit refresh of caches)
Often, I want to refresh “all” caches except those I refreshed within the last
- 8 hours
- 24 hours
- week
- month
… to save traffic. If we could have a filter, that would be easier.

moving-bits:
Fixes #7498

add cache filters for "stored within xxx" and their opposite ("stored before xxx") with xxx = any of 8 hours, 24 hours, 7 days, 30 days


moving-bits:
> While looking over this PR I am a bit confused why those strings above have already been included before.
> AFAIR there was no filter of such kind in the past, was it?

I created this PR in two steps (two commits): First the positive conditions ("saved within xxx") and strings needed for this, then the negative conditions ("not saved within xxx").

Lineflyer:
Agreed to be revisited by @moving-bits due to better sequence in filter menus / sub menus.

moving-bits:
Moved the eight new filters together with the two existing filters "Stored" and "Not stored" into its own filter category "stored" (directly below category "status"):

image

moving-bits:
> Agreed to be revisited by @moving-bits due to better sequence in filter menus / sub menus.

done

PR #7769, opened 2019-08-02
optionally show additional list memberships in cachelist

moving-bits:
Another part related to #7761: Optionally show list memberships on a separate line in cachelist. Only shows lists other than the current list.

This is to help using lists as tags by showing which "tags" are assigned to a cache.

This display is optional and can be toggled via the menu entry "manage lists" => "toggle display of additional lists". The setting gets persisted.

grafik


moving-bits:
retest this please

Lineflyer:

Lineflyer:
@moving-bits Needs conflict resolution in strings due to other PRs just merged.

moving-bits:
> @moving-bits Needs conflict resolution in strings due to other PRs just merged.

done

Lineflyer:
@moving-bits  Needs a small bit of conflict resolution once again please :)

moving-bits:
> @moving-bits Needs a small bit of conflict resolution once again please :)

done

Lineflyer:
After a first quick test:
I would have expected the setting for this feature in our settings rather than in some submenu.
I think we should not distribute the settings into our normal function menus. Same problem BTW for the compact icons.

The list member setting would be expected in Settings - Offline Data (also contains the "Ask for list" e.g.)
The compact icons belongs into Settings - Maps

Lineflyer:
After a first quick test:
I would have expected the setting for this feature in our settings rather than in some submenu.
I think we should not distribute the settings into our normal function menus. Same problem BTW for the compact icons.

The list member setting would be expected in Settings - Offline Data (also contains the "Ask for list" e.g.)
The compact icons belongs into Settings - Maps

moving-bits:
> After a first quick test:
> I would have expected the setting for this feature in our settings rather than in some submenu.
> I think we should not distribute the settings into our normal function menus. Same problem BTW for the compact icons.
> 
> The list member setting would be expected in Settings - Offline Data (also contains the "Ask for list" e.g.)
> The compact icons belongs into Settings - Maps

I tend to disagree here, although I understand that it could be useful to have all settings in preferences part as well. But some settings are settings which should be easily accessible, so that you can quickly turn them on and off.

For those the regular way of 
- completely having to leave the current program context back to the main menu, 
- then go to settings, 
- search find the setting you want to change,
- change what you like to change, 
- and then go back to the original program context, 

is quite some way. :-;

Preferences program part is for longer lasting settings primarily, IMHO. Settings which should be changeable easily should be available in their specific context. Still it's a comfort feature to make those settings permanent. We already have other settings, which work in this way, like the "live map".

Lineflyer:
Well, given the fact, that this is very unintrusive feature I would guess people would like to have it either turned on or off once and keep it like that. I would even suggest to default it to on as I consider it very useful.

There are for sure other options like hide found, compact icons which are better available in the relevant menu than in settings.
But this feature is for me very similar to "Ask for list" which you setup once and keep using it.

moving-bits:
Ah, I see, thanks for sharing this perspective. I had the impression that we always try to keep lists pretty short, so that if I add another line or two at least is has to be not-default and easily switchable.

Given your explanations I'm happy to switch. Shall I prepare a PR for this, or should we wait for user feedback first?

Lineflyer:
No, we should move the setting already now IMHO if you find the time.

Lineflyer:
BTW: I know found the perfect use case for me if I look into the "history" list:

Screenshot_20190822-231516_cgeo

PR #7772, opened 2019-08-03
complete dropDatabase / add missing tables

moving-bits:
While working on a different part I noticed that the dropDatabase method is missing some tables, therefore completed it.


moving-bits:
retest this please

Lineflyer:
Agreed to be merged at later point of time.

Lineflyer:
@moving-bits 
As we now completed the targetsdk28 this is no longer to be considered "Do not merge". Feel free if you think its ready.

Issue #5193, opened 2015-09-15
Icon size in lists different with/without overlay


Lineflyer:
In releas version 2015.09.15 there is a remaining bug in the frontend design:
Icons with overlays are larger as icons without overlays.

kumy:
PR #5194

Lineflyer:
@kumy 
I try to get an overview about recent activities while I was absent.
If I got it right I assume this issue can be closed as also  #5194 is closed?!

kumy:
@Lineflyer if I'm not mistaken, this issue is still present. PR #5194 may be a potential fix even after #5199. What about @rsudev thoughts?

Pitel:
Just noticed it, after upgrading to Nexus 5X. I never noticed it on my Nexus 4 but now I know about it, it's there too, just less noticable. So I guess it's display density related.

![Nexus 4](https://cloud.githubusercontent.com/assets/113248/18352195/b549c870-75dd-11e6-8353-e70e218709f2.png)

![Nexus 5X](https://cloud.githubusercontent.com/assets/113248/18352196/b54a2202-75dd-11e6-88bd-45722100cef9.png)

moving-bits:
Two reasons found so far lead to ragged boundaries:
- icons using up varying space
- marker line to the left (offline logs) shown or not shown

I'm preparing a fix for both. Current state is:
![grafik](https://user-images.githubusercontent.com/3754370/62706227-68516d80-b9ef-11e9-9e08-c870db469546.png)

Next step is to test it with different screen densities.

Issue #3523, opened 2014-01-07
Feature Request - Colour Code Lists/Caches


yellowduck:
I keep many lists to separate out caches into groups. Examples include Church Micros (a very popular series in the UK), specific D/T ratings, date hidden, interesting caches, night caches, etc. When out and about, I use the 'enable live' feature to display these on the map, however, once the map becomes populated, it is very difficult to distinguish the location of these 'special'  caches amongst all the others.
(Note, I also have lists for letterboxes and virtuals, which are easier to see due to the distinctive icons).

My proposal is to be able to apply colour codes to either complete lists or individual caches in order to highlight them as priority targets.

The c:geo icon for a cache has a border of white around it (grey for a disabled or archived cache). The intention is to be able to change this border - only a small number of primary colours would be needed. See mock-up on the attched image.

![2014-01-07 13 23 26](https://f.cloud.github.com/assets/1512344/1859231/14696af8-779f-11e3-811f-2709768870d6.png)

yellowduck:
This has been re-visited with #5798

yellowduck:
Also similar #5037

Kjerkloe:
I just wanted to propose a similar feature and I saw this was already suggested. 

I travel often, and every time I visit a cache-dense area, I'm missing a similar feature suggested here. It would be really usefull to have some sort of tool to mark or highlight certain caches on the map. It doesn't have to be over complicated, but even one distinctive color or any other feature that would make user marked caches (or a list of  caches) to stand out from the others, would make a huge difference and would make this awesome app even more usefull... 

Is there any progress on this topic?

magma1447:
I have for years wanted pretty much the exact same thing as the above request.

Use case example number 1:
I might have lists of D/T caches that I need. If I happen to drive fairly close to them I might want to take that detour. It therefore helps to know what it is by having it color coded, to not confuse it with all other lists one might have stored in the app.

Use case example number 2:
If I go with a group to geocache for the weekend, we usually have a shared list with our primary target of geocaches. This is stored as an offline list. Added to that, I might want to have 1-2 lists myself, that I don't feel is so important, but if we are in the vicinity I would like the group to take that small detour. This might be D/T, hidden month, a mystery I solved or whatever.

My idea has been exactly the same as the original post here. Color code lists, by changing the background color. I understand that it might conflict with the grey/disabled. In my opinion, color coded lists could/should have presence.

[...]

PR #7775, opened 2019-08-07
optionally assign icon marker to lists

moving-bits:
Third part related to #7761: optionally assign icon marker to lists and display those markers on map and in cache lists

assign marker to list or remove it from list:

grafik

markers are shown in cache list...

grafik

... and on the map:

grafik

### DIFF
--- a/main/res/layout/cachelist_listmarker_item.xml
+++ b/main/res/layout/cachelist_listmarker_item.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="horizontal"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content" >
+    <TextView
+        android:layout_width="15dp"
+        android:layout_height="wrap_content" />
+    <ImageView
+        android:id="@+id/listmarker_item_marker"
+        style="?attr/spinnerDropDownItemStyle"
+        android:singleLine="true"
+        android:layout_height="?attr/dropdownListPreferredItemHeight"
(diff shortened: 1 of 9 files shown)


moving-bits:
Hm, looking at the screenshots I notice something strange: Adding an overlay / inset to a drawable seems to change the drawable's size. How come? Might have the same cause as #5193.

moving-bits:
> Hm, looking at the screenshots I notice something strange: Adding an overlay / inset to a drawable seems to change the drawable's size. How come? Might have the same cause as #5193.

I added a fix for this with PR #7779, which needs to get merged before this one. Need to adjust a few things (to resolve merge conflicts) after merging #7779 and before merging this one, therefore setting the "WIP" label here for now.

Lineflyer:
From where do you assign these markers?
Is it only possible from the cache list view or is there also a menu/visualization in cache details?

moving-bits:
> From where do you assign these markers?
> Is it only possible from the cache list view or is there also a menu/visualization in cache details?

Those markers are related to a list, not to an individual cache, so assigning them is done from cache list view: manage lists => set list marker.

Lineflyer:
> Those markers are related to a list, not to an individual cache, so assigning them is done from cache list view: manage lists => set list marker.

Oh, this is indeed a neat method for the tags and pretty straight forward and understandable.

Lineflyer:
Merged and I will test and provide feedback.
Any depending PRs should be adapted accordingly as @moving-bits described in #7761

okainov:
@moving-bits thanks!

BTW have you seen https://github.com/cgeo/cgeo/issues/3523 and https://github.com/cgeo/cgeo/issues/5798? Seems very related :)

PR #7777, opened 2019-08-08
trackable log list: move gc info to main column to give it more space

moving-bits:
Currently the cache name info of trackable logs get cut off, only a small portion is shown. This PR moves the cache information in trackable log list to the main column to give it more space. Cache log list remains unchanged.

grafik

currently the list looks like this:
_(see the cache name "TB - Schlosshotel Rasthof Stillhorn" in the log of "michaundkaro" which got cut off to "TB - Schlossh.")_
grafik


Lineflyer:
Like it!!

PR #7778, opened 2019-08-09
output warning when restore database failed due to corrupt database

moving-bits:
Currently c:geo silently fails when restoring a corrupt database. It even displays "restore successful", which, in this case, only means that copying the backup database file has been successful.

This PR changes this behaviour to perform a check after initializing the freshly copied backup file whether the `newlyCreatedDatabase` flag is set, and outputting a warning message if so.


moving-bits:
retest this please

Lineflyer:

PR #7779, opened 2019-08-11
Fix Icon size in lists different with/without overlay (fix #5193)

### LINKED ISSUE #5193: Icon size in lists different with/without overlay

moving-bits:
Fixes the two causes for ragged display in cache list:
- fix different icon sizes (by calculating inset padding at runtime to respect the actual scaling)
- always reserve space for offline marker in cache list (show transparent marker if needed)

This fix is also a prerequisite for PR #7577.

### DIFF
--- a/main/res/drawable/mark_transparent.xml
+++ b/main/res/drawable/mark_transparent.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle" >
+
+    <solid android:color="#00000000" />
+
+</shape>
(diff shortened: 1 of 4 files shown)


moving-bits:
retest this please

Lineflyer:
Looking at your screenshot in #5193 I do now see, that the overlays are actually shown outside of the cache icon itself and no longer overlapping it. Is this intentional?
I am fine with that in general, I only fear, that the total icon size will increase in lists and map?

moving-bits:
PR adapted to the changes in PR #7775

Lineflyer:
Lets test this out as next step.

PR #7781, opened 2019-08-12
POC: show status message on main screen, if no live connector is active

moving-bits:
Use status updater to display a warning message on the main screen, if no live connector is active. ("live connector active" = connector must be active and support login, which requires validated authentication data)

POC only - there is a lot missing for now, eg:
- setting to disable this warning message (for "offline only" use)
- update this message after a connector was changed
- click on the message should lead to services authentification page
- strings to be put in resources + wordings adapted to "non technical view"
(and maybe more)

For now I would like to get feedback
- if this approach would be reasonable at all
- how to update the status window after a (relevant) change to the connectors
- how to start the authentication intent

image


moving-bits:
replaced by #7883

PR #7785, opened 2019-08-15
rename cacheType's

bekuno:
- change the geocaching.com cache type names to the current names
- implement FIND_BY_PATTERN with the old names for backward compatibility (GPX files)


Lineflyer:

Issue #7695, opened 2019-06-22
Icon not updated on change of waypoint visited state


Lineflyer:
- Open new mf map for a cache with waypoints
- Click on a waypoint
- Select Modify
- Set it to "Visited" (or vice versa)
- Return to the map

Result:
The waypoint icon is not updated (transparent background) but only after leaving reentering the map

Expected:
Icon should be instantly updated
Works on GoogleMaps

moving-bits:
Reopening, as it is fixed only partially: It works the first time you are closing the waypoint editor, but not on subsequent calls.

Lineflyer:
Was unable to reproduce this on 2021.03.01-NB1 with my worst case theme scenario, while it was easily reproducible before. I guess the dependency to ongoing background operation is still there but is now unlikely to occur. 
Lets therefore close until we might get new clues or other scenarios producing such problem.

PR #7791, opened 2019-08-16
add visit state toggle button to map's waypoint popup (fix #2592)

### LINKED ISSUE #2592: Easy setting of visited state to wpt from popup
What about setting visited state directly from wpt popup ? Button or checkbox ?

I did one multi today ... and I felt it uncomfortable to mark wpts by clicking edit, checking that checkbox than saving and getting back to map.

moving-bits:
added a toggle button to the map's waypoint popup:

image

(label switches from "set visited" to "unset visited" according to current state)

### DIFF
--- a/main/res/layout/waypoint_popup.xml
+++ b/main/res/layout/waypoint_popup.xml
@@ -30,14 +30,21 @@
 
             <LinearLayout
                 android:id="@+id/edit_box"
-                android:layout_width="fill_parent"
+                android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:orientation="vertical" >
+                android:layout_gravity="right"
+                android:orientation="horizontal" >
+
+                <Button
+                    android:id="@+id/toggle_visited"
+                    style="@style/button_small"
+                    android:text="toggle visited" />
 
                 <Button
                     android:id="@+id/edit"
                     style="@style/button_small"
                     android:text="@string/waypoint_edit" />
+
             </LinearLayout>
 
             <RelativeLayout style="@style/separator_horizontal_layout" >
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1105,6 +1105,8 @@
     <string name="waypoint_note">Note</string>
     <string name="waypoint_user_note">User note</string>
     <string name="waypoint_visited">Visited</string>
+    <string name="waypoint_set_visited">Set visited</string>
+    <string name="waypoint_unset_visited">Unset visited</string>
     <string name="waypoint_save">Save</string>
     <string name="waypoint_cancel_edit">Cancel</string>
     <string name="waypoint_loading">Loading waypoint…</string>
--- a/main/src/cgeo/geocaching/WaypointPopupFragment.java
+++ b/main/src/cgeo/geocaching/WaypointPopupFragment.java
@@ -30,6 +30,7 @@ import org.apache.commons.lang3.StringUtils;
 public class WaypointPopupFragment extends AbstractDialogFragmentWithProximityNotification {
     @BindView(R.id.actionbar_title) protected TextView actionBarTitle;
     @BindView(R.id.waypoint_details_list) protected LinearLayout waypointDetailsLayout;
+    @BindView(R.id.toggle_visited) protected Button buttonToggle;
     @BindView(R.id.edit) protected Button buttonEdit;
     @BindView(R.id.details_list) protected LinearLayout cacheDetailsLayout;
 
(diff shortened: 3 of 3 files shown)


Lineflyer:
Is it possible to have the new button leftbound, so that the popup does not get such wide?
Or maybe the same checkbox like in the waypoint edit?

moving-bits:
> Is it possible to have the new button leftbound, so that the popup does not get such wide?
> Or maybe the same checkbox like in the waypoint edit?

The two buttons are on a separate row, so if there's not enough space, the button should start below the "note" label or even left to it. ("note" does not belong to the buttons but to the waypoint infos the lines before.)

moving-bits:
> > Is it possible to have the new button leftbound, so that the popup does not get such wide?
> > Or maybe the same checkbox like in the waypoint edit?
> 
> The two buttons are on a separate row, so if there's not enough space, the button should start below the "note" label or even left to it. ("note" does not belong to the buttons but to the waypoint infos the lines before.)

See this screenshot made on a low res (emulated) device:

image

moving-bits:
retest this please

Lineflyer:

okainov:
As per screenshot it feels like really visually huge and attractive button. However `set visited` IMO is not really very often required action. I do click `more details` much-much more often.

Also, it it possible to have checkbox-like visual element instead of button? 

I would suggest to put it below, maybe as separated row after all the info, however it will make it even longer, so not really sure how feasible is it...

Can we make "more details" button to fit the width of the dialog so it's wider?

We definitely need UX guy here and in general for the app :(

Lineflyer:
First tests show its working as expected but now the bug (listed somewhere here) comes in, that the icon is not updated instantly when returning to map. 
This will confuse users for sure as their change is not visible on map (only after closing/opening the map).

moving-bits:
> First tests show its working as expected but now the bug (listed somewhere here) comes in, that the icon is not updated instantly when returning to map.
> This will confuse users for sure as their change is not visible on map (only after closing/opening the map).

This is #7695 - let's continue there for that one. I'll try to look into it.

moving-bits:
> Also, it it possible to have checkbox-like visual element instead of button?

I'll look into it, see, if it can be done with reasonable effort. If possible, I would try to implement it visually in a way similar to the waypoint list (so: checkbox + "visited" label).

> I would suggest to put it below, maybe as separated row after all the info, however it will make it even longer, so not really sure how feasible is it...

I would not like to have it that way, as it would be consuming even more space + adding visual clutter.

> Can we make "more details" button to fit the width of the dialog so it's wider?

We could, but then the "visited" toggle would have to go into a separate line - see before.

moving-bits:
> > Also, it it possible to have checkbox-like visual element instead of button?
> 
> I'll look into it, see, if it can be done with reasonable effort. If possible, I would try to implement it visually in a way similar to the waypoint list (so: checkbox + "visited" label).

It could look like this: (no implementation yet, just a visual prototype)

image

Not too bad, visually.

But: It would be breaking our regular workflows. On the waypoint edit mask, from which the user is used to the checkbox approach shown in the screenshot above, no change is permanent until you have confirmed them using the confirmation button in the upper right (check mark). - Whereas in this PR, if modified to using the checkbox approach, simply checking or unchecking would start the action. Not good...

As I do not want to introduce an extra confirmation button, and also do not want to have setting / unsetting the visited state using the dot menu, I would like to stay with the button implemented so far (as long as no other, better solution comes up). Using a button to directly start an action is standard UI.

Lineflyer:
> But: It would be breaking our regular workflows. On the waypoint edit mask, from which the user is used to the checkbox approach shown in the screenshot above, no change is permanent until you have confirmed them using the confirmation button in the upper right (check mark). - Whereas in this PR, if modified to using the checkbox approach, simply checking or unchecking would start the action. Not good...

Correct me if I am wrong:
But the current implementation of the toggled button does also save directly, doesnt it?

Lineflyer:
And generally:
I like the checkbox more than the button personally.

moving-bits:
> Correct me if I am wrong:
> But the current implementation of the toggled button does also save directly, doesnt it?

Yes, it does. But that's what you would expect from a button. (But not from a checkbox itself, at least not in "older" UI like we use in c:geo - WPF or Win10 UI is different.)

moving-bits:
> And generally:
> I like the checkbox more than the button personally.

Visually: I agree. But...

A different approach could be to create a smaller button having a checkbox icon on it (checked or unchecked according to state). Then it would be smaller, but still be a button.

Lineflyer:
> Yes, it does. But that's what you would expect from a button. (But not from a checkbox itself,...

Mmh..I have no idea of UI concepts but for me its logical. In doubt I would change, close the popup and open again to see its still changed.
My thinking would be: Everything on popup needs no confirmation...so this item might work the same way...independently fro the way its looking.

Other opinions out there?

moving-bits:
Maybe this approach works: use a labelled checkbox (as shown in the screenshot above) and save directly after setting or unsetting the check, but give a visual feedback to the use that the setting has been saved already (using a short toast message). Any opinions?

Issue #7028, opened 2018-05-26
Batch upload modified cache coords to the website


moving-bits:
Feature request:

c:geo has a nice feature to upload corrected coordinates to geocaching.com website. I frequently have the wish to do this for multiple caches at once (eg: after solving a complete trail's puzzles on my mobile device or after importing a gpx file into c:geo). This function would then upload the corrected coordinates cache by cache.

A good place for this function would be (IMHO): stored - dot menu - manage caches and then maybe "upload coordinates".

This could be implemented for personal notes as well. (This is the place where I put information on how I have solved the puzzle.)

Lineflyer:
@moving-bits 
What is your usecase scenario for this?
Do you save the corrected coords but only set them locally and later want to batch upload them to the website? That would for sure be possible as you suggest.
However in this case you could also directly select to set the corrected coords locally and on server...

If you just solve some cache and add some abitrary waypoints to them, how should c:geo judge which of them is the final? That could work out on easy caches and if the user follows some rules (e.g. set the waypoint type to "Final" and make sure it is the only waypoint of this type) but is also not likely to work out for all caches.

moving-bits:
I have two different use cases for this:

a) I solve some mysteries and save the calculated coordinates in c:geo while I'm in offline mode (or the signal is too weak for uploading)
b) I import a gpx file with caches into c:geo (eg some route I've prepared using PC programs)

In both cases I would like having a function in c:geo to batch upload the current coordinates to the website for all selected caches (which would be basically the upload function we have now for a single waypoint, but called for all selected caches in sequence). As for the waypoints: Just use the caches' current coordinates, so no specific waypoints or waypoint guessing.

Lineflyer:
> As for the waypoints: Just use the caches' current coordinates, so no specific waypoints or waypoint guessing.

In that case it would be wise to check if the current coords are modified or not, else such a batch job would also unnecessarily upload all unchanged coords. Doing so, the batch job could also be started for all caches of a list without any harm.

Obilivion:
This long awaited new feature seems to have a bug. Upload no modified coordinates are uploaded, unless you press "Set as cache coordinates in c:geo and on website" for each and any cache.

To reproduce: Import GPX with modified coordinates to c:geo. 
When selecting "Manage caches" --> Upload modified coords (german: "Geänderte Koordinaten hochladen"), even if caches are selected previously, c:geo reports successful upload, but no changes on the gc.com are performed.

[...]

PR #7795, opened 2019-08-18
add function to batch upload coords for all (selected) caches

moving-bits:
This is a follow up PR related to the recent discussion in #7028:

Our current method to batch upload coords for caches relies on the internal flag in c:geo for "modified coords", which gets set by either manually storing a waypoint's coords as new final coords in c:geo or by using a GPX import which has the needed GSAK extensions (and data).

This PR adds a separate menu entry to upload the current coords for all / all selected caches.

To upload caches of a GPX without the proper GSAK extensions you can simply import this GPX to a new list and use the new "upload coordinates for all caches" menu entry.

### DIFF
--- a/main/res/menu/cache_list_options.xml
+++ b/main/res/menu/cache_list_options.xml
@@ -74,6 +74,10 @@
                 android:title="@string/caches_upload_modifiedcoords"
                 app:showAsAction="ifRoom|withText"/>
             <item
+                android:id="@+id/menu_upload_allcoords"
+                android:title="@string/caches_upload_allcoords"
+                app:showAsAction="ifRoom|withText"/>
+            <item
                 android:id="@+id/menu_clear_offline_logs"
                 android:title="@string/caches_clear_offlinelogs"
                 android:enabled="false"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -371,6 +371,8 @@
     <string name="caches_delete_events">Delete past events</string>
     <string name="caches_upload_modifiedcoords">Upload modified coordinates</string>
     <string name="caches_upload_modifiedcoords_warning">Do you want to upload modified coordinates for all (selected) caches and overwrite the existing coordinates on the server? This cannot be undone.</string>
+    <string name="caches_upload_allcoords">Upload coordinates for all caches</string>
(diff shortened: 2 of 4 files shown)


Lineflyer:

Lineflyer:
While this was discussed in our teamspeak meeting, I still do not completely understand the consequence:
- Modified coords are not possible for all cache types and connectors. Is this new function limited to supported cache types and connectors when trying to upload modified coords?
- Is it really a good idea to have such function which might set listing coords as modified coords if applied on caches which have not coords different to listing coords. I dont even know how the gc.com website reacts on that. Does it take this change and flag as modified or does it just ignore if the coords are identical.
- For how many people might this be useful in contrast to for how many people this might mix up their server data? As long as you stay within the supported zone of modified coords (website, c:geo, GSAK extension) it is not needed.

I am really undecided here.

moving-bits:
> Modified coords are not possible for all cache types and connectors. Is this new function limited to supported cache types and connectors when trying to upload modified coords?

It's limited to supporting connectors. I have no way to check whether the specific cache type supports modified coords, but that is for the connector to handle (and returning an error code, if needed).

> I dont even know how the gc.com website reacts on that. Does it take this change and flag as modified or does it just ignore if the coords are identical.

Yes, gc.com happily stores the coords as "new coords", even if they are the original ones. So no problem on this side.

> For how many people might this be useful in contrast to for how many people this might mix up their server data? As long as you stay within the supported zone of modified coords (website, c:geo, GSAK extension) it is not needed.

Yes, but there are more sides, eg. when you receive a GPX with modified coords but without GSAK extensions and want to upload those coords to gc.com. Currently you cannot do this without having to edit each cache and set the current coords as final coords at least locally.

I've set a distinctive title for this function to make clear what it does, and to make it distinguishable from the upload of modified coords: "Upload coordinates for _**all caches**_" vs. "Upload _**modified coordinates**_" and we have a warning message popping up, which you have to confirm before anything gets uploaded.

If a different wording helps I can adapt the strings.

Lineflyer:
Well OK. Lets merge here and test. I will probably change the warning message to a very drastic one ;) later on.

moving-bits:
> I will probably change the warning message to a very drastic one ;) later on.

;-))

PR #7802, opened 2019-08-22
move 'additional lists' setting to preferences

moving-bits:
A followup to #7769:
- move "show additional lists" toggle from CacheList dot menu to preferences screen (offline data section)
- set default to true


moving-bits:
retest this please

PR #7803, opened 2019-08-22
replace 'visited' button on popup by checkbox + toast

moving-bits:
Followup to #2592 / #7791: Replace "set visited" button by checkbox + short toast


Lineflyer:
Lets test it.

Issue #7793, opened 2019-08-18
overlay icons are inconsistent


moving-bits:
c:geo uses a set of icons for:
- 8 base markers (eg: marker_disabled_oc)
- 15 cache types (eg: type_earth)
- 6 waypoint types (eg: waypoint_trailhead)
- 12 overlay markers (eg: marker_found)

### Current inconsistencies:

After the recent changes to the way overlays are put on the map, some inconsistencies in our icon set have become visible:

- Cache type overlays and waypoint type overlays have different sizes, although their base markers have the same size. 
  _Since c:geo does not freely upscale them any longer, the type overlays of waypoints do look much smaller now._

- Size of overlays do not grow in the same way as their base markers. 
  _Eg: Base marker grow from 33px in mdpi to 45px in hdpi, which is about one third), but cache type overlays grow from 22px to 36px, which is more than 60% (similar for waypoint type overlays, which grow from 16px to 26px)._

- Missing cache overlays for some densities: 
  _personalnote, stored and usermodifiedcoords are missing for hdpi, so they get upscaled automatically from mdpi by Android, which is not necessarily done in the same way as our manual scaling. Waypoint overlays, cache types and base markers are missing completely for hdpi._

- Missing cache types for some densities: 
  _Cache type hq icon is missing for hdpi._

- We have icons for three different densities, which should add up to 3x41 = 123 icons, but only 90 of them are implemented.

This all leads to pretty inconsistent look & feel for base markers, cache & waypoint types and cache overlays.

### TODO:
- Make sure the above mentioned icon set is complete for all densities supported (so either every or no icon is implemented for a given density).
- Make sure the relative sizes are consistent.

### QUESTION:
@Lineflyer @rsudev @SammysHP
Do we have high resolution templates for those icons, or do I have to restart from scratch?

Lineflyer:
As just discussed on team speak:
- The best way could be to contact the contributors of the corresponding resources directly. Icons and graphics have been added by different people over a long time. There might have not a high res version of all of them.
- Sources for graphics and icons can also be found in out "About c:geo" contribution list (e.g. the noun project).

bekuno:
A list for cache attribute icons is here:
https://github.com/cgeo/cgeo/blob/master/main/project/attributes/iconlist.txt

and templates for some images here:
https://github.com/cgeo/cgeo/tree/master/main/project/rawimages

There are more icons/infos somewhere in the tree
https://github.com/cgeo/cgeo/tree/master/main/project

[...]

PR #7806, opened 2019-08-23
fix some icon issues

moving-bits:
fixes the most pressing icon issues from #7793:
- rescale waypoint images to match the size of type images
- add three missing xhdpi overlay marker images
- add missing type_hq image for hdpi density


moving-bits:
before:
image

after:
image

PR #7809, opened 2019-08-24
update waypoint icon on change of waypoint visited state (fix #7695)

### LINKED ISSUE #7695: Icon not updated on change of waypoint visited state

moving-bits:


main/src/cgeo/geocaching/maps/mapsforge/v6/caches/CachesBundle.java
```diff
@@ -220,6 +224,18 @@ public void invalidate(final Collection<String> geocodes) {
         if (liveOverlay != null) {
             liveOverlay.invalidate(geocodes);
         }
+        if (wpOverlay != null) {
+            // get waypoint IDs for geocodes and invalidate them
+            final Set<Geocache> baseCaches = DataStore.loadCaches(geocodes, LoadFlags.LOAD_WAYPOINTS);
```
rsudev: To me this looks like something the waypoint overlay should implement and provide as a method to call from here, what do you think?

moving-bits: yes, you're right. See updated PR

rsudev:
looks good now, IMO

PR #7810, opened 2019-08-24
initialize visited checkbox

moving-bits:
followup to #7803: initialize "visited" checkbox according to current "visited" state


moving-bits:
retest this please

PR #7823, opened 2019-09-06
disable Google Maps on Android 10

moving-bits:
Quick fix for Android 10 not supporting Google Maps v1 API by disabling Google Maps on systems running Android 10 (which automatically switches to OSM map provider).


moving-bits:
Retest this please

Lineflyer:
Yes, this seems to be the easiest workaround.

PR #7853, opened 2019-09-14
make trailcolor pref dependent on trailhistory pref

moving-bits:
automatically disable trailcolor preference setting if trailhistory preference is set to disabled


moving-bits:
retest this please

Issue #7866, opened 2019-09-22
test framework changes cause tests to fail (compileSDK28 / targetSDK28)


moving-bits:
**Describe the bug:**
Tests fail when run using compileSDK28. See https://ci.cgeo.org/job/cgeo%20pull%20request/2307/console

eg:
```
14:07:29  > Task :main:compileBasicDebugAndroidTestJavaWithJavac FAILED
14:07:29  63 actionable tasks: 58 executed, 2 from cache, 3 up-to-date
14:07:29  /home/jenkins/slave/workspace/cgeo pull request/tests/src-android/cgeo/CGeoTestCase.java:9: error: package android.test does not exist
```

**Possible reason**
There are relevant changes to the Android testing framework with Android 9 / API 28, see https://developer.android.com/about/versions/pie/android-9.0-changes-all#android-test-changes

Bananeweizen:
I think the AndroidX testing infrastructure can be introduced independently of changing to SDK 28. If I remember right, that is basically the default testing library since about 2 years, where we still use an old pre-AndroidX lib.

moving-bits:
> I think the AndroidX testing infrastructure can be introduced independently of changing to SDK 28.

It depends on **compile**SDK. As soon as we switch compileSDK from 27 to 28, our tests (using the old Android testing framework) fail with `error: package android.test does not exist`.

As we want to keep compileSDK and targetSDK at the same level, updating our tests to the current testing framework become a prerequisite for switching to targetSDK28, although the switch can (and should) be done before.

Who has knowledge of the Android testing framework and can do the update?

UniQP:
> As we want to keep compileSDK and targetSDK at the same level, updating our tests to the current testing framework become a prerequisite for switching to targetSDK28, although the switch can (and should) be done before.

It's fine to update the compileSDK before updating the targetSDK.

Bananeweizen:
What I meant with my previous statement is that the testing lib upgrade can be done _before_ the SDK level changes, separately.

bekuno:
@Bananeweizen done :-)

PR #7868, opened 2019-09-22
set target SDK to 28

moving-bits:
changes target SDK to 28 and adds "uses-library org.apache.http.legacy" to make it compilable with Google Maps

does NOT increase compileSDK to 28 as this requires a couple of further changes

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -83,7 +83,9 @@
             android:value="@string/maps_api2_key" />
 
         <!-- required by Google Maps SDK version 16 and below if targeted to SDK 28 or above -->
-        <uses-library android:name="org.apache.http.legacy" android:required="false" />
+        <uses-library
+            android:name="org.apache.http.legacy"
+            android:required="false" />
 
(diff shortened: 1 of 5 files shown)


main/AndroidManifest.xml
```diff
@@ -82,6 +82,9 @@
             android:name="com.google.android.geo.API_KEY"
             android:value="@string/maps_api2_key" />
 
+        <!-- required by Google Maps SDK version 16 and below if targeted to SDK 28 or above -->
+        <uses-library android:name="org.apache.http.legacy" android:required="false" />
```
bekuno: please write it in multi line form like other code
~~~
        <uses-library
            android:name="org.apache.http.legacy"
            android:required="false" />
~~~

bekuno:
I think that I found a working solution for #7866.
It would be fine to merge this PR here, then I can prepare a PR for #7866.

main/AndroidManifest.xml
moving-bits: formatting updated accordlingly

PR #7874, opened 2019-09-29
create a "Null-Search-Loader" to present the result

### LINKED ISSUE #7865: Unable to start CacheListActivity from map (targetSDK28)
**To Reproduce:**
- use a version with targetSDK28
- open live map
- select "show as list"

c:geo crashes with LogCat as follows:
```
2019-09-22 17:31:13.457 5521-5521/cgeo.geocaching E/AndroidRuntime: FATAL EXCEPTION: main
    Process: cgeo.geocaching, PID: 5521
    java.lang.RuntimeException: Unable to start activity ComponentInfo{cgeo.geocaching/cgeo.geocaching.CacheListActivity}: java.lang.IllegalArgumentException: Object returned from onCreateLoader must not be null
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3270)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3409)
[...]
```

**Version of c:geo used:**
Today's master branch + PR #7653

Yes

**Additional context:**
see https://github.com/cgeo/cgeo/issues/7860#issuecomment-533871209

bekuno:
fix #7865 Unable to start CacheListActivity from map (targetSDK28)

It is not necessary to create a loader for showing the list.
Therefore let's catch only the Exception.

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -43,6 +43,7 @@ import cgeo.geocaching.loaders.FinderGeocacheListLoader;
 import cgeo.geocaching.loaders.HistoryGeocacheListLoader;
 import cgeo.geocaching.loaders.KeywordGeocacheListLoader;
 import cgeo.geocaching.loaders.NextPageGeocacheListLoader;
+import cgeo.geocaching.loaders.NullGeocacheListLoader;
 import cgeo.geocaching.loaders.OfflineGeocacheListLoader;
 import cgeo.geocaching.loaders.OwnerGeocacheListLoader;
 import cgeo.geocaching.loaders.PocketGeocacheListLoader;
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -521,7 +521,12 @@ public void onCreate(final Bundle savedInstanceState) {
             initActionBarSpinner();
         }
 
-        currentLoader = (AbstractSearchLoader) getSupportLoaderManager().initLoader(type.getLoaderId(), extras, this);
+        try {
+            currentLoader = (AbstractSearchLoader) getSupportLoaderManager().initLoader(type.getLoaderId(), extras, this);
+        } catch (final NullPointerException e) {
```
rsudev: I don't think it is an awfully good idea to blindly swallow an exception here. If the loader is 'not needed', we should either detect it and avoid calling the creation.

bekuno: The loader is not needed, but the initialization.
Now I found the (hopefully) a correct implementation.

bekuno:
retest this please

PR #7887, opened 2019-10-07
set cache data to active during tests, if it is online disabled (#7886)

### LINKED ISSUE #7886: Tests with live caches fail if they are disabled
**Describe the bug:**
Tests with live caches fail if they are disabled.

**Samples:**
~~~
1) testSearchByGeocodeBasis(cgeo.geocaching.CgeoApplicationTest)
org.junit.ComparisonFailure: [Cache GC3FJ5F: disabled] expected:<[fals]e> but was:<[tru]e>
at cgeo.test.Compare.assertCompareCaches(Compare.java:27)
at cgeo.geocaching.CgeoApplicationTest.testSearchByGeocodeBasis(CgeoApplicationTest.java:306)
~~~
~~~
2) testSearchFromMap(cgeo.geocaching.connector.gc.GCBaseTest)
org.junit.ComparisonFailure: [Cache GC3FJ5F: disabled] expected:<[fals]e> but was:<[tru]e>
at cgeo.test.Compare.assertCompareCaches(Compare.java:27)
at cgeo.geocaching.connector.gc.GCBaseTest.testSearchFromMap(GCBaseTest.java:26)
~~~

bekuno:
fix #7886 Tests with live caches fail if they are disabled

### DIFF
--- a/tests/src-android/cgeo/geocaching/CgeoApplicationTest.java
+++ b/tests/src-android/cgeo/geocaching/CgeoApplicationTest.java
@@ -303,6 +303,11 @@ public class CgeoApplicationTest extends CGeoTestCase {
             try {
                 mockedCache.setMockedDataUser(Settings.getUserName());
                 final Geocache parsedCache = CgeoApplicationTest.testSearchByGeocode(mockedCache.getGeocode());
+
+                // enable here if cache is disabled (see #7886)
+                if (parsedCache.isDisabled()) {
+                    parsedCache.setDisabled(false);
+                }
                 Compare.assertCompareCaches(mockedCache, parsedCache, true);
             } finally {
                 mockedCache.setMockedDataUser(oldUser);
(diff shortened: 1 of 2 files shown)


Lineflyer:
IMHO this should be only in as workaround as it might also make the test pass if the cache is false parsed as disabled, or not?

bekuno:
> IMHO this should be only in as workaround as it might also make the test pass if the cache is false parsed as disabled, or not?

You are right.
There is no additional way to check if it is realy disabled.
I will prepare a new PR for the moment when the cache is enabled.

moving-bits:
Do we really need to change our tests twice just for a known and temporary problem?
IMHO we can simply ignore those tests _temporarily_ - the tests affected are known, the cause is known, and the test reports quite clearly state the cause, so that we know if they are failing for a different reason.

Lineflyer:
> Do we really need to change our tests twice just for a known and temporary problem?

I agree, that it might be not needed.
I expect that I can enable the cache next weekend. If something like this happens for something out of our control, we might apply a workaround or change the test accordingly. However in this case we can just wait.

moving-bits:
not needed anymore as GC3FJ5F is online again :-)

PR #7888, opened 2019-10-07
deactivate the test of a disabled cache state (see #7886/#7887)

bekuno:
please merge when the cache GC3FJ5F is active again.


moving-bits:
not needed anymore as GC3FJ5F is online again :-)

PR #7895, opened 2019-10-13
fix some trail history issues (fix #7830)

### LINKED ISSUE #7830: Smoothen Trail History
Switching screen off when it's not needed (in order to save battery life) with GPS switching off when screen is off (same goal, of course), produces GPS position jumps, as first GPS reading seems to either give no position at all (that given, c:geo jumps to home position) or a random one (seemingly where GPS has been a short while ago).

These jumps are stored without any filtering in the trail history and may result in a sometimes extremely weird track, which becomes nearly unreadable, even over short periods of time, in extreme situations even the map cannot be read properly anymore.

A GPS position resp. it's movement may only be stored in the trail history, when reliability of GPS position is high enough and distance to last x number of readings is below a predefined or even configurable value. BTW: a trail history, jumping around the home position just because the GPS does not give reliable readings or is even switched off, is not really meaningful anyway. ;)

Alternatively trail history should be smoothed afterwards (using e. g. Ramer–Douglas–Peucker algorithm)

moving-bits:
Fixes the following regressions and missing pieces for trail history:
- error in trail history routines of our Mapsforge map implementation, which prevented a correct distance calculation and also caused the current position to be connected with the beginning of the trail history
- trail history routines of Google Maps v2 implementation did not respect the maximum distance at all (as well as ignored the trail color setting)

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -44,6 +44,7 @@ public class GooglePositionAndHistory implements PositionAndHistory {
     private GoogleMapObjects positionObjs;
     private GoogleMapObjects historyObjs;
     private GoogleMapView mapView;
+    private final int trailColor;
 
     private static Bitmap locationIcon;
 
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

PR #7896, opened 2019-10-13
fix/add missing de translations

moving-bits:
Adds a number of missing de translations + some minor fixes


Lineflyer:
These wont last very long as they will get overwritten by Crowsdin import.
We should really find a way to allow inport from Crowdin in some automized way so more Contributors can do it.

moving-bits:
Sad to hear that... My hope was that we upload our (partial) translations as well.
So the only way is to get a Crowdin account and do our translations there? Can we override certain translations (rare cases)?

PR #7898, opened 2019-10-14
make 'not logged in' warning clickable (fix #7897)

### LINKED ISSUE #7897: Make "not logged in warning" clickable to open account settings
I like this new warning on main screen and from what I have tested it works as intended.

I suggest to change the string to something like this and make the message clickable to open the "First steps" page:

"Not logged in yet. Live map will only show caches stored locally and online functions will not work.\n Either there is no network connection, a server problem or you have to configure your account by clicking this message."

What do you think?

moving-bits:
make 'not logged in' warning clickable (fix #7897)


moving-bits:
Reminder to myself: need to crosscheck on a low res device

PR #7904, opened 2019-10-20
fix Lint/Checkstyle issues

bekuno:
- checkstyle.xml
  - switch to the revised Doctype definition
  - explicitly linefeed as line separator for check NewlineAtEndOfFile
- fix different code annotation warnings
  ("not annotated method overrides method annotated with ..." etc.)

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
@@ -33,9 +33,7 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.List;
 
-
 import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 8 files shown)


bekuno:
retest this please

main/src/cgeo/geocaching/connector/gc/GCLoggingManager.java
```diff
@@ -67,7 +68,7 @@
     }
 
     @Override
-    public void onLoadFinished(final Loader<String> arg0, final String page) {
+    public void onLoadFinished(@NotNull final Loader<String> arg0, final String page) {
```
UniQP: Shouldn't it be `@NonNull` instead of `@NotNull`?

bekuno: You are right. Is fixed.

Issue #7909, opened 2019-10-23
Delay "not logged in" message on first start


Lineflyer:
Beta feedback 2019.10.22-RC2

Many beta users are confused by the new "not logged in" warning as it may be shown on start of c:geo while it is performing the login.

We should target to only show this message in cases where it is likely, that some kind of malfunction or error occured.

I suggest to delay the warning by x seconds after the not logged in status is detected to avoid unnecessary warnings while logging in is still in progress or during short connection interruptions.

x=30 seconds ?

moving-bits:
Hm, the current implementation works the other way around: The "not logged in" warning is presented as default message directly at startup (without any checking) and cleared as soon as a successful login is detected. Don't know how to delay this initial message...

SammysHP:
But there are three states: logging in, not logged in and logged in.

moving-bits:
The current info message is informative, but quite long and blocky. Besides possibly changing the logic behind the message, maybe we can shorten the text to something like "Not yet logged in. c:geo will only show caches already stored offline."?

PR #7920, opened 2019-10-27
shorten 'not logged in' info + add dialog to configure services

moving-bits:
in regard to #7909:
- shorten "not logged in" warning on main screen to make it less obstructive
- tap on this message opens dialog with detailed info; "OK" on dialog leads to service configuration, "cancel" back to main screen

info message if not (yet) logged in:

image

dialog after tapping on above message:

image

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -20,22 +20,6 @@
         tools:layout="@layout/status" />
     <!-- ** -->
 
-    <TextView
-        android:id="@+id/info_notloggedin"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:background="@drawable/helper_bcg"
-        android:layout_centerVertical="true"
-        android:layout_below="@+id/status"
-        android:layout_marginLeft="16dip"
-        android:layout_marginRight="16dip"
-        android:padding="4dip"
-        android:gravity="center"
-        android:textColor="@color/text_icon"
-        android:textIsSelectable="false"
-        android:textSize="14sp"
-        android:text="@string/warn_notloggedin" />
-
     <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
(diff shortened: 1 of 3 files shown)


Lineflyer:
Like it

moving-bits:
@Lineflyer
One additional idea: I could move this info box down to the connectors' section, as it is related. This would make it even less intrusive, and also harder to mistake for our regular info box. What do you think?

Lineflyer:
Can you share a screenshot of how it would look like?

moving-bits:
> Can you share a screenshot of how it would look like?

A possible placement could look like this, but the details could be adjusted:

image

(not yet pushed to the PR)

bekuno:
pro: better relative to the 'services' section. This is fine for users who know this and take a look at this section. 
con: it is not in the main view area. Therefore I assume that most users do not recognize the message.

How about a more eye-catching display, perhaps with a red background?

rsudev:
Please do not forget that the capability to function completely offline is one of the basic design ideas (and for me e.g. the main mode of operation) of c:geo...

moving-bits:
@bekuno 

Thanks for your feedback. I do share the pros and cons you listed.

> How about a more eye-catching display, perhaps with a red background?

I'd rather not introduce a new color scheme, which would have to be interpreted / understood and set in relation to other messages of c:geo in terms of relevance and danger. 
So I suppose to stick with the original PR (which is what I've pushed to this repo, the screenshot from my latter post was done offline).

moving-bits:
@rsudev 

> Please do not forget that the capability to function completely offline is one of the basic design ideas (and for me e.g. the main mode of operation) of c:geo...

Yes, offline capabilities of c:geo have been the key feature which has brought me to c:geo in the first place, actually.

But that is exactly the difficulty we have with this issue - the spread of experience our users have with c:geo, and the spread of different ways c:geo is used:
- Do inform the user that something might not work as expected, because none of the connectors could log in (either intentionally or not),
- but do not stand in the way of anybody for which this is the intended way of working with c:geo
- and do not trigger "false alarms" because someone misreads the above info as being a serious error.

So, as mentioned in my last post, I suggest to stay with the implementation above - unobstructive but informative (IMHO), and if someone needs more info he/she can tap on the message and gets additional info on the background.

Lineflyer:
> 
> So, as mentioned in my last post, I suggest to stay with the implementation above - unobstructive but informative (IMHO), and if someone needs more info he/she can tap on the message and gets additional info on the background.

I agree, but positioning it in the lower section?

Lineflyer:
Addon: Also this PR should be merged to release branch

moving-bits:
> I agree, but positioning it in the lower section?

Hm, I don't understand your comment - shall I leave it at the current position (above the six main buttons), or shall I move it down to the connectors' section?

Lineflyer:
Yes, moving it down was my proposal. Sorry for being unclear.

But be aware people might have 2 or 3 connectors active which occupies more space already.  Theoretically many connectors could be active blocking many lines, but thats not likely.

moving-bits:
> But be aware people might have 2 or 3 connectors active which occupies more space already. Theoretically many connectors could be active blocking many lines, but thats not likely.

Tested on a different (emulated) device with lower resolution and a handful connectors activated:

image

PR #7933, opened 2019-11-03
update README for API keys

bekuno:
add GoogleMaps Key Source
make the text clearer


moving-bits:
@bekuno Sorry, have been AFK for the whole week. Thanks for updating the file and making it clearer.

PR #7938, opened 2019-11-09
list marker enhancements

moving-bits:
list marker enhancements:
- use named labels in marker selection instead of just the marker's id
  (makes it easier to add more markers later which are not just colored dots)
- add black marker


moving-bits:
retest this please

Lineflyer:
@moving-bits I would have rather prefered to not change release branch by pushing functional changes on that branch. Such things should go to master branch instead, even more given the fact that we are in beta phase at this moment.

moving-bits:
@Lineflyer My bad, I'd not considered this a functional change - just an additional icon to an existing set of icons, and displaying color names instead of "marker 1", "marker 2" etc. - I can revert it, if you prefer.

Lineflyer:
After looking into the changes, I principally agree, but as there are string changes involved we have some new untranslated then.
If possible I would suggest to revert and push to master instead.

PR #7941, opened 2019-11-10
expand GMv2 by user defined cache long click

moving-bits:
expand GMv2 by user defined cache long click
(long click on GMv2 map to open dialog to create a user defined cache)

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java
@@ -1,5 +1,7 @@
 package cgeo.geocaching.maps.google.v2;
 
+import cgeo.geocaching.connector.internal.InternalConnector;
+import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.Viewport;
 import cgeo.geocaching.maps.DistanceDrawer;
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

Lineflyer:
@moving-bits Do we have some issue where the concept of this user defined caches is drafted and/or can be discussed. I do see what you did so far, but I don't see the clear goal what we want to achieve with this feature. I am sure it might be helpful for some feature requests, but I do not see it at the moment. If there is no such issue, we should open one for it.

moving-bits:
> @moving-bits Do we have some issue where the concept of this user defined caches is drafted and/or can be discussed. I do see what you did so far, but I don't see the clear goal what we want to achieve with this feature. I am sure it might be helpful for some feature requests, but I do not see it at the moment. If there is no such issue, we should open one for it.

I'd written down some thoughts in #7918. 
Main reason is to be able to save user defined waypoints which are not bound to an existing cache. Next step could be replacing the current "go to" screen by the waypoints page of a predefined cache of the internal connector.

Lineflyer:
How about targeting the long click on map to the very same user defined cache waypoint screen you propose to be used to replace the Goto menu?

Meaning:
Whether the user uses Goto or long clicks on the map, it all comes together into this special user defined cache. So the user has all his cache independent waypoints at one place.

moving-bits:
> How about targeting the long click on map to the very same user defined cache waypoint screen you propose to be used to replace the Goto menu?
> 
> Meaning:
> Whether the user uses Goto or long clicks on the map, it all comes together into this special user defined cache. So the user has all his cache independent waypoints at one place.

I would rather like to keep this separate, as those are two separate use cases IMHO:
- using "go to" if you do have specific coordinates you want to see on the map (and/or calculate), but do not want to save them with an existing or new cache;
- using a "user defined cache" if you want to group some waypoints into a new cache and/or give the collection of waypoints a specific name

This leaves open the variant that you want to have an unspecific waypoint you don't want to have a separate cache for, and for which you want to tap on the map to create it. Don't know how often this happens.
If it's for a relevant amount we probably would need to open a popup menu after long tapping on the map, to choose between "create 'go to' waypoint" and "create user defined cache".

Issue #3288, opened 2013-09-16
Replace "Go to" by waypoint view


Bananeweizen:
Couldn't we replace the any screen by an invocation of the waypoints screen and just put all these waypoints to some artificial cache ID?

This would increase the usability of that feature and at the same time reduce our code.

Lineflyer:
To bring up this again:
In a discussion with some users it came up, that they would prefer to have to `Go to` menu look like the waypoint screen.
Then I found this long existing issue and want to emphasize again, that it would be a good idea to completely replace the `Got to` by a waypoint screen of a virtual cache in our database. That would just be perfect.

Lineflyer:
Implemented by using the user defined cache approach

PR #7946, opened 2019-11-11
replace go to menu by user defined cache and waypoint page

moving-bits:
Issue #3288 proposes replacing the "go to" menu by a waypoint view. With introduction of the internal connector and user defined caches with #7918 this can be implemented easily now.

Here are some screenshots:

a) After pressing "go to" button on mainscreen a waypoint view of the automatically created user defined cache opens:

image

(There is only one user defined cache created for all "go to" targets, so if you press "go to" again you will be directed to the same cache. If you delete this cache (which you are perfectly allowed to do) it will be recreated on next run.)

b) Waypoint handling is the same as you're used to:

image

c) And you may open the compass from waypoint view directly:

image

(And as this is a regular cache you may swipe to the cache's main page and open navigation and much more.)

Current state of this PR is fully functional, but not yet cleaned up. It shows how replacing the "go to" menu by a automatically created user defined cache would look and work, but does not yet remove the current implementation or migrate any history data - the latter would have to be discussed later.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1057,6 +1057,8 @@
     <string name="internal_cache_default_name">User defined cache #%1$d</string>
     <string name="internal_cache_default_description">This is a user defined cache</string>
     <string name="internal_cache_default_owner">You</string>
+    <string name="internal_goto_targets_title">\'Go to\' targets</string>
+    <string name="internal_goto_targets_description">This cache stores your recent \'Go to\' targets</string>
 
     <!-- editor dialog -->
 
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -174,6 +174,8 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
     private static final int MESSAGE_FAILED = -1;
     private static final int MESSAGE_SUCCEEDED = 1;
 
+    private static final String EXTRA_FORCE_WAYPOINTSPAGE = "cgeo.geocaching.extra.cachedetail.forceWaypointsPage";
+
     /**
      * Minimal contrast ratio. If description:background contrast ratio is less than this value
      * for some string, foreground color will be removed and gray background will be used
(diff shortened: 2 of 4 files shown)


pstorch:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- main/src/cgeo/geocaching/CacheDetailActivity.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/cgeo/cgeo/pullRequest?prid=4479648&bid=15117564)

Lineflyer:
@pstorch Was that really you or does our Codacy bot somehow have a split personality? :)

pstorch:
@Lineflyer no, but I was the last one setting up the integration between Codacy and GitHub. I think then Codacy continues using this account.

Lineflyer:
> @Lineflyer no, but I was the last one setting up the integration between Codacy and GitHub. I think then Codacy continues using this account.

I could swear, that Codacy was using an own dedicated user name....not sure.

@moving-bits 
I like this proposal. 
We should only think a bit about the differences to the current solution and if/how we deal with it.
- E.g. waypoints created here might be shown on the live map (which might be a nice addon but might also be undesired by some users?!).
- What about the cache itself, where these waypoints belong to? In your screenshot the user can access it. Should we maybe remove this possibility to swipe to the other tabs in this context?

Lineflyer:
@moving-bits 
As we have time to test and improve on master branch, I would be fine with merging this approach.

However some more thoughts in addition to my comment above:
In order to maintain usability in a similar way, how about starting directly into the waypoint edit mode for a reference point with prefilled current coordinates. If we do that, we come very close to the initial view of the current Goto menu? On the other hand starting with the waypoint tab is more inline with what users know from a cache. I am undecided.

moving-bits:
> E.g. waypoints created here might be shown on the live map (which might be a nice addon but might also be undesired by some users?!).

That's what waypoints are meant for ;-)
Generally it's a desired behavior, at least if you going to cache details -> map.
You may delete either the cache or some of it's waypoints to make them go away.

OTOH I can imagine that it can be confusing for some. Field test to get some feedback on it? Then we can decide to either do it permanently one or the other way, or to make it configurable, if needed.

> What about the cache itself, where these waypoints belong to? In your screenshot the user can access it. Should we maybe remove this possibility to swipe to the other tabs in this context?

I would stay with the "cache details" page being accessible for this cache type to have access to the functions offered there. Other tabs could be disabled, though, probably.

> In order to maintain usability in a similar way, how about starting directly into the waypoint edit mode for a reference point with prefilled current coordinates. If we do that, we come very close to the initial view of the current Goto menu? On the other hand starting with the waypoint tab is more inline with what users know from a cache. I am undecided.

Me too. Currently I tend to the latter, to have it consistent with waypoint editing for regular caches.
(I'm not sure how many people are actually using the "go to" feature and thus would be affected by the change. Personally I haven't used "go to" since a couple of years, I'm always using waypoint editing coming from a cache (and may it be a user defined cache imported via self-created GPX file...))

moving-bits:
> > > @pstorch Was that really you or does our Codacy bot somehow have a split personality? :)

> > @Lineflyer no, but I was the last one setting up the integration between Codacy and GitHub. I think then Codacy continues using this account.

> I could swear, that Codacy was using an own dedicated user name....not sure.

Definitively, using "Codacy Bot" as user name. Received the last one around the beginning of October, so the change (whatever it was) has happened after that.

PR #7947, opened 2019-11-13
Add geochecker geotjek.dk

zagge-cgeo:
** Description of change: **
Add the geochecker geotjek.dk used in e.g. this cache
https://coord.info/GC2P6JZ

### DIFF
--- a/main/src/cgeo/geocaching/utils/CheckerUtils.java
+++ b/main/src/cgeo/geocaching/utils/CheckerUtils.java
@@ -21,6 +21,7 @@ public final class CheckerUtils {
             "geocheck.org/geo_inputchkcoord.php?",
             "geochecker.com/index.php?",
             "geochecker.gps-cache.de/check.aspx?",
+            "geotjek.dk/geo_inputchkcoord.php?",
     };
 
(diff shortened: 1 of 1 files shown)


Lineflyer:
ok to test

Lineflyer:
@zagge-cgeo Can you please provide a geocode of a cache where this is used?

Lineflyer:
@zagge-cgeo  Sorry, did not see it in your comment. Nevermind.

Lineflyer:
@zagge-cgeo Still one question: This looks like the same website as geocheck.org. Do you know whether they have even more aliased, which we could/should add in the same step?

zagge-cgeo:
> @zagge-cgeo Still one question: This looks like the same website as geocheck.org. Do you know whether they have even more aliased, which we could/should add in the same step?

@Lineflyer I do not know that but I currenlty check for checkers not yet supported by cgeo

zagge-cgeo:
I have greped trough the caches I have in my database but found none other one that uses the same path as geochecker.com. Howerver I have added support for the checker of geocache-planer.de that is used quite often.

PR #7950, opened 2019-11-16
update gc-apps.com geochecker url

moving-bits:
Geochecker URL of gc-apps.com has either changed or never worked with "geochecker" path (receiving a 404 error now), but works with "checker" path, so updated it in c:geo utils class.


Lineflyer:
Do you have a geocode at hand which uses this checker?

moving-bits:
> Do you have a geocode at hand which uses this checker?

Eg. GC6F99F or GC80676

Lineflyer:
Thanks, now I see what you mean. 
IMHO they used the other URL in their very early time, but then changed it without properly redirecting the old one. I found one old (no longer working) link on a cache page last days but forgot to write down the code.

Issue #7952, opened 2019-11-20
Strategy for handling of user defined caches


Lineflyer:
Recently @moving-bits implemented "user defined caches" as new feature (see #7918 and #7946). This is a nice feature, especially as replacement for the long time orphaned "Go to" menu. Thanks for that!!

However I think we should discuss some ideas about the implementation details to improve usability.
As base for discussion here are my thoughts after trying the feature in our nightlies (also summarizing some other comments I made for the mentioned PRs):

1. First of all a generic idea: At the moment new user defined caches (from map or "Go to") are stored into the default list. I suggest to use a dedicated new default list for those caches instead (users cann still move/copy afterwards if desired). The reason is mainly the "Go to" cache, which will be automatically created and might confuse the user, when showing up on the default list.

2. Using the waypoint screen of a "user defined" caches as "Go to" screen is really perfect. Our users are used to that UI, so they will most probably be able to understand and use it.
However the "Go to" now also presents all the menus and options, which belong to a cache and this might either be irritating or at least meaningless in that special "Go to" use case. But IMHO the users will get used to that, once they learn about the "user defined cache" feature I guess. Not sure about any optimization possibilities for this special "Go to" case.

3. I guess the "user defined cache" feature is targeting power users/experts, rather than beginners and normal cachers. For this very reason I suggest to discuss again, whether it should be really associated to "long click" on the live map. E.g. it happened to me several times recently, that I accidentally triggered this dialog while clicking or moving the map. 
I could imagine using a setting (in map menu or settings or both (linked)) to optionally turn this behavior on but keeping it off as default. 

4. At the moment other attributes of the cache cannot be assigned (e.g. D/T, size, etc.). Would this be something for future improvement of the feature or not needed?

Additionally some questions to @moving-bits for clarification of the functionality:

- After creating an UDC (nice new abbreviation) from the list menu it has no coordinates. To assign coordinates to it, you have to create a waypoint and set this as listing coords. Is that the intended way or is there any other option I did not see yet. 
While this is perfectly fine as it fits in our normal UI, the only drawback is, that it will always also assign a final marker to the cache.

- How did you implement the geocode assignment? I first guessed you use the next free code in that name space (ZZ*) but also after deleting older UDC the numbers are not reused (or was it only the internal caching after deletion which has tricked me?).

- Shall we keep the prefix with ZZ or any better idea? Are there any reserved prefixes for such use case...is there any regulation for this at all? 

[...]

PR #7957, opened 2019-11-21
put goto target user defined cache into its own list

moving-bits:
part of #7952: Put the pseudo cache created for "goto targets" into its own list instead of standard list. Cache will only be shown if you chose "All caches".

Other user defined caches behave unchanged, will get stored to either the current list (if available) or the standard list.


bekuno:
I propose to put all UDCs initial in this list.

PR #7961, opened 2019-11-22
make trailcolor string translatable (fix #7960)

### LINKED ISSUE #7960: String "Trail color" not translatable
"Trail color" in Settings - Map is not contained in `strings.xml` thus not translated.

moving-bits:
Fixes #7960


Lineflyer:
Thanks for fixing, noticed it while doing the user guide adaptions :)
Will merge, when lights are green.

PR #7971, opened 2019-11-26
crowdin: New translations

bekuno:
downloaded 2019-11-26 04:50:48


bekuno:
the changes in the download script should not be in this PR

PR #7979, opened 2019-11-30
throttle distance calculation to prevent NewMap ANR (fix #7923)

### LINKED ISSUE #7923: Mapping large lists on NewMapOSM leads to ANR
**Describe the bug:**
A user contacted us with this problem on support mail. Not yet retested, but sounds reasonable as the bug is not there, if he uses "Old Mapsforge V3"

**To Reproduce:**
- Use OSM or offline map
- Have a list with (in this case) approx. 1300 caches (same with 800 caches and slow with 560 caches)
- Map this list
- Map headline appears and map tries to load

c:geo does no longer respond and ANR error occurs

c:geo should work normal.
Its working with old mapsforge V3 as well as with Google maps

**Version of c:geo used:**
2019.10.05

Yes

moving-bits:
throttle distance calculation to prevent NewMap ANR (fix #7923)
on large caches/waypoints lists with activated proximity notification

- uses faster implementation of calculating min distance
- ensures a minimum timespan between two calculations

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
@@ -234,16 +234,4 @@ public class MapsforgeCachesList extends AbstractItemizedOverlay {
     public int size() {
         return items.size();
     }
-
-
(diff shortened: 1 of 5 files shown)


Lineflyer:
@rsudev Can you review this PR?

PR #7982, opened 2019-12-01
change CacheNumber on testSearchByOwner

bekuno:
Test error:
~~~
V/InstrumentationResultParser: There was 1 failure:
V/InstrumentationResultParser: 1) testSearchByOwner(cgeo.geocaching.CgeoApplicationTest)
V/InstrumentationResultParser: java.lang.AssertionError:
V/InstrumentationResultParser: Expected size:<8> but was:<9> in:
~~~

Reason:
New mystery cache now by blafoo (GC8G0X7) from 15.11.19.

### DIFF
--- a/tests/src-android/cgeo/geocaching/CgeoApplicationTest.java
+++ b/tests/src-android/cgeo/geocaching/CgeoApplicationTest.java
@@ -230,7 +230,7 @@ public class CgeoApplicationTest extends CGeoTestCase {
             public void run() {
                 final SearchResult search = GCParser.searchByOwner("blafoo", CacheType.MYSTERY);
                 assertThat(search).isNotNull();
-                assertThat(search.getGeocodes()).hasSize(8);
+                assertThat(search.getGeocodes()).hasSize(9);
                 assertThat(search.getGeocodes()).contains("GC36RT6");
             }
         });


moving-bits:
@bekuno  Thanks for fixing the test. (Cache is from Nov 15, but got published today, so tests started failing today). Do you know if blafoo wants to publish a new cache every day for the "Adventskalender"? Then we should either disable this test temporarily or use a different owner.

@Lineflyer Shouldn't we target this PR to release first and merge release to master after that, to have tests fixed on release branch as well?

bekuno:
@moving-bits Don't know, if this is a cache with 24 days or start of a serie.
@Lineflyer Do you know blafoo?

Lineflyer:
@blafoo was a former contributor to c:geo. Most probably he created this test case.

If you want to rewrite the test case you might use my profile instead. It might be, that I publish new caches, but not that frequent.

Lineflyer:
Lets take this one first to get our tests working

blafoo:
I published a new cache. To make the test more stable maybe the condition should be something like size() >= 9

Lomanic:
Wouldn't be more sensible to use a dedicated test account for this (and all other similar tests also), or someone we know isn't playing anymore?

PR #7983, opened 2019-12-01
Link to new offline map tutorial

Lineflyer:
instead of the old user guide page with only a few sentences of explanation


Lineflyer:
retest this please

moving-bits:
The new offline map tutorial is very well done, thanks for your work!

PR #7985, opened 2019-12-01
change testSearchByOwner to use Lineflyer's earth caches

moving-bits:
In case blafoo starts a 24 day Adventskalender (which would break this test again and again): update testSearchByOwner to use Lineflyer's earthcaches instead (5 caches as of today, all non-PMO)


Lineflyer:
OK!
Once you merged this please check merging upstream to master as it will result in a conflict (due to the other PR) which needs to be resolved then.
But targeting `release` is the correct way for fixing tests...I did not pay attention to it when merging the other PR.

PR #7995, opened 2019-12-09
Move commit id computation to dedicated gradle class

olivergondza:
Fixes trivial gradle script TODO

** Description of change: **

Move all commit ID computation to dedicated class.

** Screenshot (if applicable): **

N/A

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -61,7 +61,7 @@ android {
     buildTypes {
         debug {
             // debug build name contains git commit for better reproduction of bugs
-            versionNameSuffix "-" + gitCommitId().substring(0, 7) + " developer build"
+            versionNameSuffix "-" + GitCommit.currentShort() + " developer build"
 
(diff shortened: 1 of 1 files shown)


main/build.gradle
```diff
@@ -440,22 +440,24 @@ static def versionNameFromDate() {
...
-        def stdout = new ByteArrayOutputStream()
-        exec {
-            commandLine 'git', 'rev-parse', 'HEAD'
-            standardOutput = stdout
+    /*
+     * get the most recent git commit ID
+     */
+    static String current() {
+        if (commitId == null) {
+            Process p = ['git', 'rev-parse', 'HEAD'].execute()
```
olivergondza: It seems challenging to run the `exec` task from within a class method (as it resolves the symbol from that context failing to find class field/method of that name). Refactoring to standard groovy process invocation that is not all that non-idiomatic.

PR #8001, opened 2019-12-13
add option to disable controls for manual map rotation on GMv2

### LINKED ISSUE #7986: Google Maps v2: disable screen rotation
This is quite the opposite of  #7507 (linked to that issue at least): I find this an anti-feature, while zooming in or out with two fingers, to have the map rotated quite frequently because my fingers were not moving exactly the right way to have the map not rotated. This of course not only occurs with cgeo and its new Google Maps v2, but with many other applications (original Google Maps, Citymapper for example), and I always have to put the map back to north-oriented afterwards (by clicking the compass in the upper left corner for cgeo).

Also, and this is similar to #7953, when you switch from rotated Google Maps to OSM: Map (online), the orientation is then set to north as it can't (currently?) be rotated. Switching back to Google Maps, the map is then pointing north.

I would like map rotation to be disabled ([simple to achieve according to this SO answer](https://stackoverflow.com/a/13700449)) to keep the same user interactions as before the migration to Google Maps v2, or at least to have a setting to disable or enable this feature.

Toggle switch to enable this feature.

moving-bits:
fixes #7986

Adds menu option "Settings" - "Map" - "Manual rotation" to disable controls for manual rotation of map. (Currently only available for Google Maps v2, as our OSM implementation does not allow rotation at all.)

### DIFF
--- a/main/res/values/changelog_master.xml
+++ b/main/res/values/changelog_master.xml
@@ -5,6 +5,7 @@
     <b>Next feature release:</b>\n
     · New: Ability to create user defined caches\n
     · New: Add support for geocheckers of geocache-planer.de and geotjek.dk\n
+    · New: Option to disable Google Maps controls for manual screen rotation\n
     · Change: Use waypoints of a user defined cache in \"Go to\" menu\n
     · Change: Link from offline map \"More info\" page to the new offline map tutorial\n
     · Fix: Support for geochecker gc-apps.com\n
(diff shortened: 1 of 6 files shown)


pstorch:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/cgeo/cgeo/pullRequest?prid=4676050&bid=15646068)

PR #8008, opened 2019-12-16
Merge branch release

Lineflyer:
@moving-bits 
This needs conflict resolution as you changed relevant files on both branches without merging branches or rebasing your change on `master`.
Can you check the conflicts?

### DIFF
--- a/main/res/values/changelog_release.xml
+++ b/main/res/values/changelog_release.xml
@@ -2,6 +2,10 @@
 <resources>
   <!-- changelog for the release branch -->
   <string name="changelog_release" translatable="false">\n
+    <b>Next bugfix release:</b>\n
+	· Fix: Avoid app freeze when mapping large lists on OSM maps\n
+	\n
+	\n 
     <b>2019.11.28:</b>\n
     · Fix: Update distance info on Google Maps while moving\n
     \n
--- a/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
@@ -234,16 +234,4 @@ public class MapsforgeCachesList extends AbstractItemizedOverlay {
     public int size() {
         return items.size();
     }
-
-
-    public int getClosestDistanceInM(final Geopoint coord) {
-        int minDistance = 50000000;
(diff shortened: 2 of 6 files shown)


Lineflyer:
Better not use the webtool to resolve :)

moving-bits:
@Lineflyer  Hm, I don't understand how this happened - I had merged PR #7979 into release and merged release to master directly after that (#8000), without any conflicts. I have done both using the Github UI - should I've done the merge in a different way?

Lineflyer:
It seems the sequence of commits is not ok AFAICS from my tablet (on vacation at the moment).

How did you merge PR #8000? Looks like you did not use a merge commit but squashed the changes into a new seperate commit. Is that possible?

Lineflyer:
In case thats the culprit you should revert the commit which was created by your merge and then try again to merge branches via PR but select to do it via merge commit.

Not sure if this is enough though....still learning git, but I made some improvements since last time.

moving-bits:
> In case thats the culprit you should revert the commit which was created by your merge and then try again to merge branches via PR but select to do it via merge commit.
> 
> Not sure if this is enough though....still learning git, but I made some improvements since last time.

Same here... What would happen on master if I revert a commit "from the middle" - there have been a couple of more commits to master since #8000.

Lineflyer:
> Same here... What would happen on master if I revert a commit "from the middle" - there have been a couple of more commits to master since #8000.

The reverting commit is just a new commit reverting the changes of the original.
So (I guess) it will work as such but the result might not be functional if the later changes rely on these changes. But this shouldn't be a problem as you merge the changes via branch merge afterwards. Not sure about if any of these actions might report conflicts....mmh.. try it.
As long as you dont force anything you should not cause any unresolvable problem..it only might get more complex.

Lineflyer:
> Can't remember. But commit history visualization for master branch definitely looks different:
> 
> ￼

Yes, thats wrong.
When merging from release please always do a merge commit and do not apply the same changes under a new commit. This breaks the history of branch mergings.

On the Github UI use "Create merge commit" in the dropdown of the merge button.

rsudev:
I could try to take a look tomorrow. 
Sorting out git trouble is something I am somewhat used to.

Lineflyer:
That would be great.
Please share the sequence resolving this for us to learn :)

PR #8011, opened 2019-12-18
DO NOT MERGE: cleanup of squashed merge of release

rsudev:
To sort out the accidentally squashed merge (see #8008), the commit is reverted and than release merged properly again. As there was some other change already done addressing the same area, a conflict resolution was necessary at both stages. This PR is only for inspection by (@Lineflyer, @moving-bits), i would push the changes directly if you are fine with them.
Steps taken:
- revert the commit a72352889f59cbe2dcc203237d38fad0ae7646aa. In the conflict resolution take extra care to remove inconsistencies stemming from the fact that the same code was changed in between
- merge again the same state of release which was the base for the failed merge. Check the result against the current tip of master, as it has to be identical. Amend the merge commit if necessary.
- merge the tip of release
- push

### DIFF
--- a/main/res/values/changelog_release.xml
+++ b/main/res/values/changelog_release.xml
@@ -2,6 +2,10 @@
 <resources>
   <!-- changelog for the release branch -->
   <string name="changelog_release" translatable="false">\n
+    <b>Next bugfix release:</b>\n
+	· Fix: Avoid app freeze when mapping large lists on OSM maps\n
+	\n
+	\n 
     <b>2019.11.28:</b>\n
     · Fix: Update distance info on Google Maps while moving\n
     \n
--- a/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/MapsforgeCachesList.java
@@ -234,16 +234,4 @@ public class MapsforgeCachesList extends AbstractItemizedOverlay {
     public int size() {
         return items.size();
     }
-
-
-    public int getClosestDistanceInM(final Geopoint coord) {
-        int minDistance = 50000000;
-        for (final CachesOverlayItemImpl item : items) {
-            final int distance = (int) (1000 * coord.distanceTo(item.getCoord()));
-            if (distance > 0) {
-                minDistance = Math.min(minDistance, distance);
-            }
-        }
-        return minDistance;
-    }
 }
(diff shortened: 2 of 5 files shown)


Lineflyer:
This PR represents what I would expect in case the original branch merge was done correct. It does only contain the recent changes I made on changelog on the release branch (after the wrong merge happened). 
But where is the correct merge of the fix @moving-bits commited onto the release branch?

Lineflyer:
Dont get me wrong. I do see the several commits of this PR, however the overall result of these commits are only the change of changelog_release.xml

moving-bits:
@rusdev  Thanks for sorting out this, and sorry for the confusion created by my wrong merging approach with #8000.

> Dont get me wrong. I do see the several commits of this PR, however the overall result of these commits are only the change of changelog_release.xml

Is this probably due to merges on master after #8000 are expecting the changes of #7979 (which got merged by #8000) being present on master, so conflict resolution indirectly reintroduced those changes to master?

moving-bits:
Which leads me to two different questions:

In case we do not plan to release from current release branch: What would happen to the two branches if we just reverted those two changes on release branch?

Or else: What would happen if we simply ignored the wrong merge on master? The commit tree for master has no visible connection from release branch to master branch for #8000.

rsudev:
Hmm, maybe I misunderstood what was expected (in that case good that I didn't push directly).
What I understood:
- @moving-bits had the intention to merge release to master to get his fix with the throttling onto master as well
- Accidentally the UI rebased the changes on master instead. This leads to conflicts between release and master and a problem merging future changes from release (and merging master to release for a release)
There are several ways to sort this out. One thing can be, to just merge release and resolve the conflicts once, than your are good again. Another possibility is to revert the accidental commit and than merge again (the approach used here).
Of course are all the changes introduced through the accidental commit again on master after merging release again correctly. It does not change the code, just makes the history more transparent. Unfortunately is this less clean than could be wished for in this case, as conflicting changes were done on master after the accidental commit. But those would have lead to conflicts even when merging correctly after they were done, so that can't be helped and is not a big deal anyway.
If this clarifies everything I would push this state, before other changes landing on master require me to re-do the exercise.

moving-bits:
@rsudev  The intention is as you've described
- get both branches back to a "clean" state (being able to merge in both directions), and
- get the changes of #7979 onto master branch
Either way to reach those goals are fine to me, the cleaner the commit history the better.

Thank you for your work!

rsudev:
Done.

Lineflyer:
Looks perfect. Thanks!

PR #8012, opened 2019-12-18
add support for locationless cache type (fix #8010)

### LINKED ISSUE #8010: Support new locationless cache type
https://coord.info/GC8FR0G

Remark:
At this moment the cache is locked, which is visualized as archived in c:geo (intentional).

moving-bits:
Fixes #8010

Add config to detect caches of the type "Locationless (Reverse) Cache" - the once only reintroduced cache type for GCFR0G. As it's probably only a single new cache, I did not created a new symbol for it but reused the "virtual cache" symbol.

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -35,6 +35,7 @@
     <string name="gchqceleb">Geocaching HQ Celebration</string>
     <string name="gps">GPS Adventures Exhibit</string>
     <string name="block">Geocaching HQ Block Party</string>
+    <string name="locationless">Locationless (Reverse) Cache</string>
     <string name="userdefined">User defined cache</string>
     <string name="unknown">Unknown Type</string>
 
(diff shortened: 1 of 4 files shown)


bekuno:
Comparing with PR ##7785 I think we need also a line like
      FIND_BY_PATTERN.put("Lost and Found Event Cache".toLowerCase(Locale.US), COMMUN_CELEBRATION);

And, if I remember correctly, there are CacheType compares in the code somewhere.

moving-bits:
@bekuno  Integrity cheks and locus connection added. Could not find places where comparing with this individual cache type would be relevant.
As for the possible pattern extension: I don't know if this cache type got renamed, but maybe we can skip this anyway, as the last locations caches have been grandfathered nearly 15 years ago. (But if if has been renamed and someone can send me the old name, I'm happy to add the pattern.)

bekuno:
I found this locationless cache
https://www.geocaching.com/geocache/GCB48F_die-letzte-ihrer-art?guid=9ce22167-5a8d-451c-a69b-b19d5ec4f578
and some Virtual caches without position (traveling geocache), e.g.
https://www.geocaching.com/geocache/GC3507

The current used type for the old locationless cache is the same as today.

The PR looks good to me.

PR #8019, opened 2019-12-22
crowdin: New translations

bekuno:
downloaded 2019-12-22 15:00:35


moving-bits:
@bekuno  Thanks for updating

PR #8023, opened 2019-12-23
Crowdin new languages

bekuno:
- some updates 
- add 6 languages with translations (Approved %)
  - Arabic (78%)
  - Filipino (59%)
  - Hebrew (9%)
  - Icelandic (9%)
  - Indonesian (53%)
  - Tagalog (66%)


Lineflyer:
We used to add new languages only if they received a significant amount of translations. There is some issue here about new languages AFAIR, where we discussed it. However technically I don't see a problem to also add languges with less translation percentage... The only thing could be, that they are then shown as "xxx language" supported but in fact most of it is still english. But maybe this even encourages users to contribute.

Lineflyer:
Addon: For me it would be >50% are OK in any case.

moving-bits:
@Bekuno  Please have a look at the three new lint warnings - they are about the naming of the language directories as well as about one of the translations string's parameters.

bekuno:
Thanks for the info.
I will create a new PR without Hebrew & Icelandic.
And check of cause the lint warnings.

bekuno:
> @bekuno Please have a look at the three new lint warnings - they are about the naming of the language directories as well as about one of the translations string's parameters.

@moving-bits One is 
strings.xml:1024,           StringFormatCount, Priority: Normal
Inconsistent number of arguments in formatting string `file_searching_sdcard_in`; found both 0 and 1

https://ci.cgeo.org/job/cgeo%20pull%20request/2456/androidLintResult/new/package.-963969145/source.-1116759823343971321/#1014

This is a old entry, it was no lint warning before.

@Lineflyer 
The other lint warnings are
The locale folder "`he`" should be called "`iw`" instead; see the `java.util.Locale` documentation
The locale folder "`id`" should be called "`in`" instead; see the `java.util.Locale` documentation

Did we had such warning before?

bekuno:
from java.util.Locale:
ISO 639 is not a stable standard; some of the language codes it defines (specifically "iw", "ji", and "in") have changed. This constructor accepts both the old codes ("iw", "ji", and "in") and the new codes ("he", "yi", and "id"), but all other API on Locale will return only the OLD codes.

PR #8031, opened 2019-12-27
adjust heading pointer to map's bearing (fix #8021)

### LINKED ISSUE #8021: Rotating map not rotating as expected
Recently @moving-bits implemented the rotation for GMapsV2 (see #7507).

Up to now I only had very limited time to test this feature, however I have instantly seen something unusual (at least on my Samsung S10, not yet tested with any other device):

As the map shall rotate according to the device compass orientation I would have expected, that the position arrow on the map (which in fact shows my device orientation related to the map) is now always pointing upwards on the display.  This would be logical as the map is now rotated acc. to my orientation upwards.
However on my device there is a delta between map orientation and pointer orientation most of the time. I cannot yet say what exactly is happening here, however for me it looks like the position arrow rotation is more accurate than the map rotation.

Looking forwards for other testers and feedbacks.

moving-bits:
set rotation of heading marker according to current map's bearing

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -210,7 +210,7 @@ public class GooglePositionAndHistory implements PositionAndHistory {
         positionObjs.addMarker(new MarkerOptions()
                 .icon(BitmapDescriptorFactory.fromBitmap(getLocationIcon()))
                 .position(latLng)
-                .rotation(heading)
+                .rotation(mapRef.get().getCameraPosition().bearing)
                 .anchor(0.5f, 0.5f)
                 .flat(true)
                 .zIndex(ZINDEX_POSITION)


moving-bits:
@Lineflyer  So far I could test this in the emulator only. Do you have a chance to field test this?

Lineflyer:
I can try to if time permits (in our nightlies).
However my experience with the current code is, that the heading marker is more correct than the map bearing. 
But I did not test the rotation feature very intense (and only on one device).

moving-bits:
@Lineflyer  Thanks. I'll try to get another field test with a different setting (heading + bearing) in parallel. Let's see what works better.

Lineflyer:
I merge this to test in nightlies as noted above.

moving-bits:
> I merge this to test in nightlies as noted above.

@Lineflyer  Thanks. I've an alternative version which uses bearing + heading, which I tested yesterday in the field with an S10. Works well most of the time, but has some occasional jumps. Not sure if this is due to the signal or the program code, but at least there has not been any visible fixed anomaly. Looking forward to your readings...

Lineflyer:
Feel free to submit a PR if you have further optimizations to the code in this PR.
What is the codewise difference between bearing and heading?

PR #8036, opened 2019-12-29
crowdin: New translations

bekuno:


Lineflyer:
>  

I am wondering why this strings seem to be new. They should be in the english source since a long time, shouldn't they?

Lineflyer:
Oh, they were introduced with aa4867aa37c442c04c0ac1224729908875254f5e

@moving-bits Why is that?

moving-bits:
> Oh, they were introduced with [aa4867a](https://github.com/cgeo/cgeo/commit/aa4867aa37c442c04c0ac1224729908875254f5e)
> 
> @moving-bits Why is that?

Lint remarked missing hint texts for coordinate calculator input fields, so I extended the layout file accordingly, and as the strings required for that had not been there in strings.xml yet, I added them.

Lineflyer:
Ah, understand.
I think they are not correct. The MinDec (Geocaching standard) format does not have any seconds but fractions of minutes or am I wrong? Also the DegDec format has fractions of degress and not minutes. But please doublecheck, I am not so sure.

moving-bits:
> I think they are not correct. The MinDec (Geocaching standard) format does not have any seconds but fractions of minutes or am I wrong? Also the DegDec format has fractions of degress and not minutes. But please doublecheck, I am not so sure.

Hm, I need to doublecheck if fields are reused with a different meaning - I went by the field names only :-(

PR #8045, opened 2019-12-31
checkstyle optimization

### LINKED ISSUE #6946: checkstyle configs are broken
Unfortunately the checkstyle configuration xml schema sometimes changes in incompatible ways. Our config schema does not fit the more recent version used in codacy by now, therefore the codacy scan seems to ignore it (and reports checks we disabled).

will upgrade the configuration

### LINKED ISSUE #8032: checkstyle configuration for static imports
The checkstyle definition for imports is defind by the module ImportOrder.
See https://checkstyle.org/property_types.html#importOrder

The current checkstyle configuration defines static imports as "top". There is no definition for grouping the static imports. This means that all static imports are in one group on top of all imports, but the non static imports are separated in groups.
I suggest to include the static imports in the non static import section, my favorite is "under".

What do you think?

bekuno:
fix #8032

- configure AS checkstyle to version to 8.26 as used on codacy
- set checkstyle output language fix to English
- change java import order rule (option under, severity info)
- update suppressions.xml doctype definition
- code formatting: xml intent to 4

For testfolders see #8052
maybe containes changes for #6946

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -4,7 +4,7 @@
     <option name="configuration">
       <map>
         <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
-        <entry key="checkstyle-version" value="8.8" />
+        <entry key="checkstyle-version" value="8.26" />
         <entry key="copy-libs" value="true" />
         <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
         <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -4,138 +4,140 @@
     "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
 <module name="Checker">
-  <module name="SuppressionFilter">
-    <property name="file" value="suppressions.xml"/>
-  </module>
-  <module name="FileTabCharacter">
-    <property name="eachLine" value="true"/>
-    <property name="fileExtensions" value="java"/>
-  </module>
-  <module name="NewlineAtEndOfFile">
-    <property name="severity" value="warning" />
-    <property name="lineSeparator" value="lf" />
-  </module>
-  <module name="TreeWalker">
-    <property name="tabWidth" value="1" />
-    <module name="AvoidStarImport">
-      <property name="severity" value="info" />
-      <property name="allowClassImports" value="false" />
-      <property name="allowStaticMemberImports" value="false" />
+    <property name="localeLanguage" value="en" />
+
+    <module name="SuppressionFilter">
+        <property name="file" value="suppressions.xml"/>
     </module>
-    <module name="DefaultComesLast">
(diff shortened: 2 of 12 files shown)


Lineflyer:
Did you check (or adapt if you think its incorrect) our wiki pages which mentions import order:

https://github.com/cgeo/cgeo/wiki/Coding-Conventions

bekuno:
> Did you check (or adapt if you think its incorrect) our wiki pages which mentions import order:

The import order specifically for static imports was different used in the code (mainly in the tests).
Therefore my issue  #8032. As nobody stated to stay on the separation for static imports I created the change in this PR.

I willl also update the document.

The results for codacy, on CI and in my AS look different. I assume we have a separate configuration on codacy that does not use the checkstyle.xml from the code and the folder definition from gradle/suppressions.xml. 
I can't see (or change) the codacy configuration for the team - maybe it is necessary to have the owner role.

moving-bits:
@bekuno  Is it possible to integrate CheckStyle in the local build process with AS in a way that the build fails if not all CheckStyle checks are green? Currently you have to start the checks manually for each file, and so sometimes something slips through and only gets caught during the server build.

bekuno:
retest this please

bekuno:
retest this please

bekuno:
> @bekuno Is it possible to integrate CheckStyle in the local build process with AS in a way that the build fails if not all CheckStyle checks are green? Currently you have to start the checks manually for each file, and so sometimes something slips through and only gets caught during the server build.

Mmm, really?
Please aware that checkstyle != lint.

If you have the checkstyle plugin installed in AS (and configured, see https://github.com/cgeo/cgeo/wiki/Android-Studio), then you can
- on the checkstyle tab: "check all modified files" or "check files in the current change list"
- when committing: activate "scan with Checkstyle"
- start "./gradlew check" or "./gradlew checkstyle"

On CI we start explicitely "./gradlew ... checkstyle lintBasicDebug ".

Bytheway, I changed the severity for ImportOrder to info. I think that is enough to see a problem.

moving-bits:
> > @bekuno Is it possible to integrate CheckStyle in the local build process with AS in a way that the build fails if not all CheckStyle checks are green? Currently you have to start the checks manually for each file, and so sometimes something slips through and only gets caught during the server build.
> 
> Mmm, really?
> Please aware that checkstyle != lint.

I know. I would not make lint mandatory, too many possible false positives due to the nature of this checking. (And currently too many open items.)

> If you have the checkstyle plugin installed in AS (and configured, see [cgeo/cgeo/wiki/Android-Studio](https://github.com/cgeo/cgeo/wiki/Android-Studio)), then you can
> * on the checkstyle tab: "check all modified files" or "check files in the current change list"

Ah, good to know. This toolbar item wasn't visible in my configuration (and the green "play" button too prominent to look for it). Is there a way to move toolbar icons? I've seen something for the built-in toolbars, but not for plugin toolbars.

> * when committing: activate "scan with Checkstyle"
> * start "./gradlew check" or "./gradlew checkstyle"

Thanks for the hint - I've added a pre-commit hook for CheckStyle now.

PR #8047, opened 2020-01-01
refactor to increase utilization of lambda expressions

moving-bits:
straighten and clarify code by increased utilization of lambda expressions


moving-bits:
retest this please

PR #8051, opened 2020-01-03
localization scripts (folder) optimization

### LINKED ISSUE #8050: localization scripts
During the work on crowdin translations and the configuration of the c:geo language I became aware of the folder main/project/localization.
Several points have been noticed/need improvement:
- there is no readme for it
- findmissintraslations does not work with languages whose abbreviation is longer than 2 characters
- findextratranslations finds 2 entries that are correct and therefore should not have been found (auth_connected, settings_reauthorize)

bekuno:
fix #8050

- add readme
- fix: findmissintraslations does not work with languages whose abbreviation is longer than 2 characters
- add as known false positives in readme (wont fix): findextratranslations finds 2 entries that are correct and therefore should not have been found

### DIFF
--- a/main/project/localization/readme.txt
+++ b/main/project/localization/readme.txt
@@ -1,0 +1,31 @@
+Checks for installed languages
+==============================
+
+These scripts are only relevant if you are not sure whether the crowdin translations are up-to-date.
+
+* findmissingtranslations.sh
+
+Search of missing translation strings in the language files other than English.
+Creates a file <lang>.missing for all checked languages in this directory.
+
+Parameter:
+  * all     check all languages
+  * <lang>  check a dedicated language
+
+
+* findextratranslations.sh
+
+Search of redundant strings in
+
+Parameter:
+  * -n      only display
+  * -f:     force unused strings removal
+
+Known false positives:
+* auth_connected:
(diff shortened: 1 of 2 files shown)


main/project/localization/readme.txt
```diff
@@ -0,0 +1,31 @@
+Checks for installed languages
+==============================
+
+These scripts are only relevant when unsure that the crowdin translations are not up-to-date.
```
moving-bits: Suggestion for text:
These scripts are only relevant when unsure whether the crowdin translations are up to date.

main/project/localization/readme.txt
```diff
@@ -0,0 +1,31 @@
...
+
+* findmissingtranslations.sh
+
+Search of missing translation strings in the language files other then english.
```
moving-bits: typos: "then english" => "than English"

main/project/localization/readme.txt
bekuno: Ok, and a litte further changed.

main/project/localization/readme.txt
bekuno: fixed.

PR #8053, opened 2020-01-03
set isStoreEnabled in MapState on map source change (fix #8049)

### LINKED ISSUE #8049: Caches disappear after consecutive  map source change
- Open live map with OSM online or offline map
- Deactivate live function so that only saved caches are shown
- Leave live map
- Reenter live map (still deactivated)
- Saved caches are shown
- Change map source to Google
- Caches still shown
- Change map source to OSM online or offline

Result:
- No caches shown anymore on map. On further change of map source no caches are shown until live is enabled/disabled again

Expected Result:
- Caches should be shown as before regardless of the map source

moving-bits:
Fixes #8049 

tested both cycles:
- MF current => GMv2 => MF current
and 
- MFv0.3 => GMv2 => MFv0.3

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -1017,8 +1017,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

PR #8054, opened 2020-01-03
update contributors page

moving-bits:
Jan 2020 update to about:contributors page


moving-bits:
@Lineflyer  It looks like we have some trouble with our CI - All new PR don't get a fresh build, status stops at "Pending - Build triggered for merge commit."
Can you have a look (or explain me where to look, if I have the rights necessary for that)?

PR #8071, opened 2020-01-13
Create UDC for migrated 'go to' targets during migration (fix #8042, fix #8067)

### LINKED ISSUE #8042: Go to targets lost
I saved three Go to targets in c:geo today.

Some hours later I wanted to revisit them however the list was empty.

Not sure what happened.

### LINKED ISSUE #8067: Migration of GoTo targets not working
I just smoke tested a release candidate (build from now merged release branch).
Before upgrading from version 2019.12.30 I created 7 different GoTo targets and verified, that they have been stored.

After upgrading to the new RC/Beta however the waypoints of ZZ0 are empty.

moving-bits:
- Create UDC for migrated 'go to' targets during migration
- UDC ZZ0 gets stored in standard list (instead of non visible virtual list)

### DIFF
--- a/main/src/cgeo/geocaching/connector/internal/InternalConnector.java
+++ b/main/src/cgeo/geocaching/connector/internal/InternalConnector.java
@@ -8,7 +8,6 @@ import cgeo.geocaching.connector.capability.ISearchByGeocode;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.LoadFlags;
 import cgeo.geocaching.enumerations.StatusCode;
-import cgeo.geocaching.list.PseudoList;
 import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.models.Geocache;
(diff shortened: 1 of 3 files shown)


moving-bits:
retest this please

Lineflyer:
@moving-bits Due you consider this PR ready to be merged? I woudl then redo the smoke test and we see if the failing scenario (#8067) works now.

moving-bits:
> @moving-bits Due you consider this PR ready to be merged?

At least for beta testing it should be fine - just rechecked in the emulator I'd created for this a couple of days ago.

I'm currently storing the ZZ0 UDC into standard list. We once had a discussion where to store it, which resulted in the (alas failing) attempt to store it in a hard coded virtual list. If we don't like storing it to the standard list, we still have the alternative to store it into a different, automatically created, but regular list - although I'm not really comfortable with creating an extra list just for one cache.

But anyway, as ZZ0 for now gets created into default list, a user can move it to a different list if wanted. And we can decide throughout beta phase whether we want to create a separate list for it or not.

> I woudl then redo the smoke test and we see if the failing scenario (#8067) works now.

Thanks for testing 👍

PR #8072, opened 2020-01-13
check if viewport is only a dot before zoomToBounds (fix #8068)

### LINKED ISSUE #8068: Google maps open with worldmap
There are two reports on support mail from nightly users:
Those users experience since some days, that the Google map opens in world view (max zoom out) when opening the map for a cache.

I was unable to reproduce up to now but I guess this is could be somehow related to https://github.com/cgeo/cgeo/pull/8046

moving-bits:
If a cache has no waypoints, the viewport spanned by this cache is just a dot (upper left corner identical to bottom right corner), which leads GMv2 to use world view on `zoomToBounds()`. Precheck for this condition and use the previous zoom instead of `zoomToBounds()` in this condition.


moving-bits:
retest this please

PR #8075, opened 2020-01-14
Add an AppShortcut to directly launch the Live-Map

jonas-koeritz:
Let the user go straight to the Live-Map from their app drawer or home screen.

photo5256059635884993642

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -109,6 +109,8 @@
                 <!-- ProcessPhoenix uses the default flag to recognize which activity to use after restart -->
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
+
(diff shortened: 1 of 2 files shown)


Lineflyer:
Nice.

IMHO we already have such function as a widget, thus why not use all widget functions also here?

jonas-koeritz:
Can somebody point me to where the widget code lives? Had a quick look but didn't find it :P

moving-bits:
@jonas-koeritz 

> Can somebody point me to where the widget code lives? Had a quick look but didn't find it :P

`main/src/cgeo/geocaching/CreateShortcutActivity.java`

jonas-koeritz:
Recommended maximum number of AppShortcuts is 4 so i left out "Go to", wich also has some preconditions that would add complexity to this implementation.

Issue #7631, opened 2019-05-20
Check if icon change is needed due to updated Google specs


Lineflyer:
Google Play informed, that there is a new specification for app icons:
https://developer.android.com/google-play/resources/icon-design-specifications

There will be a legacy (smaller) version used as of June,24th 2019 if nothing is changed. So it is not urgent, and I even do not know if any changes to our icon might be recommended due to these updated specifications.

Maybe a topic for @SammysHP to take a look at?

jakedot:
I have seen some icons in major apps revert to the standard "Anrdoid app" icon on some devices. E.g. the GS Geocaching app on my BV P6000, the WhatsApp on a CAT S30, ... and I think it is related to this change. 

I hope that c:geo is not affected, but I agree it doesn't seem urgent.

Lineflyer:
This is the preview of options on Google Play Console:
![grafik](https://user-images.githubusercontent.com/949669/58128012-fbf87680-7c16-11e9-896f-04e79d43f49e.png)

bekuno:
Did we have any changed icons?
I didn't really notice anything of it.

Lineflyer:
No, otherwise this issue would probably be closed.

moving-bits:
fixed with 37f0a9c

PR #8076, opened 2020-01-15
Implemented a proposal for an adaptive icon.

jonas-koeritz:
I implemented an adaptive icon for c:geo that will work from Android 8.0 onwards on all launchers that support them. The screenshots are all from my Pixel 4 XL running Android 10, switching between the different theme options for icon outline.
This doesn't look too bad, given I just had to modify some parts of the SVGs to get this to work. IMO the 3D effect of the "Map" part of the icon is weakened a little but Google doesn't encourage icons with a distinct outline anymore.
Devices < API Level 26 will still display the old PNG icons.

photo5258281246373555500
photo5260563372951317635
photo5260563372951317636
photo5260563372951317637


bekuno:
see issue #7631

bekuno:
Why did you use these special folder names?

Lineflyer:
It looks nice IMHO. People will recognize it (even withour shadow ;) ).

moving-bits:
Like it 👍

moving-bits:
> Why did you use these special folder names?

By appending a version number to certain ressource folder names you can tell Android to include those ressources only from this version on. Also you can provide alternative implementations for different versions.

SammysHP:
Sorry, not my taste. The detailed map doesn't work as a background with a cutout mask. It also feels like something is missing from that lonely container.

jonas-koeritz:
Somebody with a more designer"ish" background should have a look at what we could come up with.

Lineflyer:
> Sorry, not my taste. The detailed map doesn't work as a background with a cutout mask. It also feels like something is missing from that lonely container.

Shall we then better keep the non-optimized until we get a better proposal or use this as an interim version? Or would you be able to create a better design, IMHO you redesigned the logo already in the past.

jonas-koeritz:
I haven't done any work on the logo before. Let's stick with the current one until we come to some consensus.

davidlb:
What about a single color background or even just white as most of google apps icon?

jonas-koeritz:
That white background is just the default fallback icon when adaptive icons
are not supported. c:geo already looks like that when running on my Pixel 4.

David Le Brun <notifications@github.com> schrieb am Fr., 17. Jan. 2020,
11:36:

SammysHP:
I'd keep the old logo and just add a better background (color or texture).

Lineflyer:
> I'd keep the old logo and just add a better background (color or texture).

Maybe a black/darkgrey color like our homepage background? But I am not at all a good design expert.

Lineflyer:
@SammysHP 
Any proposal from your side?
It does not help to keep this pending. So we should either have a new design propoal, accept this one or close it...

SammysHP:
Then let's close it. If someone has a better idea we can reopen it.

PR #8079, opened 2020-01-17
add cache shortinfo to waypoint compass (fix #4674)

### LINKED ISSUE #4674: Add more info to compass screen for waypoints
Hi.

Would it be possible to display the information for cache size and D/T in the compass view also for other waypoints? Right now it is only shown when using the compass on the main cache coords.

moving-bits:
Fixes #4674


moving-bits:
retest this please

PR #8084, opened 2020-01-20
crowdin: New translations

### LINKED ISSUE #8080: Portuguese  language not working (beta)
We have several users of beta version 2020.01.16-RC which claim, that c:geo is no longer using portuguese language but english.
Did not verify yet.

bekuno:
also fix #8080

now based on release with help of  
~~~
git cherry-pick c260c4eb05cb428ebab89d5a1026eb444a4b3f51
~~~
:-)


moving-bits:
retest this please

Lineflyer:
seems our CI test for signature needs to be adapted to the new decade :)

Lineflyer:
@moving-bits Still a minor issue here as you can see in 140 new lint issues:
https://ci.cgeo.org/job/cgeo%20pull%20request/2518/androidLintResult/new/

This commit added strings which are not contained in the source language (so these are master strings). I have no idea if this will cause problems in the app, but I guess not....so we coule leave it like that (until we merge from master again)?!

Lineflyer:
I have also seen the Replace "..." with ellipsis character (…, &#8230;)  in lint results for values-pt which I (hopefully) now corrected on Crowdin. So it should be solved on next import.

moving-bits:
> @moving-bits Still a minor issue here as you can see in 140 new lint issues:
> [ci.cgeo.org/job/cgeo%20pull%20request/2518/androidLintResult/new](https://ci.cgeo.org/job/cgeo%20pull%20request/2518/androidLintResult/new/)

Not all 140 issues are new, just a handful. "New issues" is always reported relative to the last build, not necessarily to the git ancestor.

> This commit added strings which are not contained in the source language (so these are master strings). I have no idea if this will cause problems in the app, but I guess not....so we coule leave it like that (until we merge from master again)?!

IIRC they get automatically remove by the R8 code shrinker, so they should not cause any harm. Even if they would not be removed, they would not get requested by the app, so they would only "bloat" the app a bit, but not more. - It would be a different case, though, if the same resource name got different values between release and merge branch and the wrong one got merged into the app.

Issue #8082, opened 2020-01-19
Localize name of ZZ0 cache and provide more explanation


Lineflyer:
Some beta testers are confused about the ZZ0 cache showing up on their default list.
While the description explains the meaning (also localized AFAIR), the title " 'Go to' targets" seems to be in english only.

Maybe translating it will help a bit.

Additionally in contrast to the first planning this cache is now on the default list.
If there is no possibility to make it "invisible" we could at least provide a more detailed info text in its cache description (which describes why it is there, how to handle it, and that it can be deleted and will be autocreated if the Go to menu is used).

moving-bits:
> Some beta testers are confused about the ZZ0 cache showing up on their default list.
> While the description explains the meaning (also localized AFAIR), the title " 'Go to' targets" seems to be in english only.
> 
> Maybe translating it will help a bit.

That's the weak spot of migration: During the database operation I cannot access the translated strings (or at least I haven't found a way to do so, as no `Context` object is available to this background task, but is necessary to translate strings).
So if we want to migrate the old history targets, we have to stay with English in the beginning.

If you delete ZZ0, and create new "go to" targets a couple of days (\*) later, a new ZZ0 will be created automatically, and this time it will be in localized strings.
((\*) "A couple of days" being at least three days, as this is our tolerance limit after which a deleted cache gets actually removed from database.)

> Additionally in contrast to the first planning this cache is now on the default list.
> If there is no possibility to make it "invisible" we could at least provide a more detailed info text in its cache description (which describes why it is there, how to handle it, and that it can be deleted and will be autocreated if the Go to menu is used).

Do you have a proposal for the explanation?
Or shall we simply add a link to the online manual for further explanation. (By this it would be easier to change explanation details. OTOH we would have less translations, probably.)

Additionally we can create an individual list for this cache (alas non-translated as well), "ZZ history of 'go to' targets" or something like this.


A more radical way would be to not migrate any historic "go to" targets. By this, ZZ0 would be created in a localized version. (And we can still decide whether to put it into a new list and/or add some more explanation or a link to its description.)

moving-bits:
> Some beta testers are confused about the ZZ0 cache showing up on their default list.
> While the description explains the meaning (also localized AFAIR), the title " 'Go to' targets" seems to be in english only.
> 
> Maybe translating it will help a bit.

[...]

PR #8095, opened 2020-01-22
remove migration of "go to" targets

moving-bits:
Second part of #8082: remove migration of "go to" targets from database scheme routines


Lineflyer:
retest this please

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -165,7 +164,7 @@
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
-    private static final int dbVersion = 78;
+    private static final int dbVersion = 77;
```
Lineflyer: What happens to nightly and RC users with this forced "downgrading"?

Issue #7215, opened 2018-12-16
Add automatic sync of personal notes


okainov:
Seems like there is no automatic personal note sync\upload to the website. So users have options:
- Just forget and do not get their notes uploaded, which is the most popular use case, I believe :(
- Click on small "upload" button every time they're online and there are some changes
- Use batch "export" regularly.

Instead of those, I'd suggest to have somehow automatic scheduled task which should upload all modified personal notes to the gc-service website. For completely automatic upload probably dependent on "dirty" flag suggested in #6590, however it could be done by putting "upload" task into some queue directly after editing, so it can be processed after some time.

A few extra points to consider:

- Is there a need to differentiate connection types? Probably we shouldn't spam metered 3G\4G connection and use auto-upload with Wi-Fi only
- How often upload should be done? I'd suggest "as needed", if there is some queue of tasks, we should instantly put the event there and uplaod handler should check it every couple of minutes when app is active

okainov:
>such an option located under export - personal notes 
That's exactly my third bullet. This is (1) not automatic and (2) will take forever in foreground with 1000+ caches.

> I also opened an issue as I think the export menu is not the right place for it.
I saw it, but it seems to have nothing to do with subj suggestion

okainov:
At least you do automatic status update for status page =)

bekuno:
I would prefer a small solution.
The app should (if onfigured) try to upload the note when saving the note.
And additionally it would be nice to have a state indicator to see if the note was uploaded before. This would also be useful without automatic upload.

okainov:
@bekuno from the one side I agree that it would be better than nothing. However there can be several issues with uploading while saving:
- Offline case - I usually add personal note while travelling w\o Internet. That's 95% of my personal usage, I can easily believe that I'm not unique here.
- Many edits in the short timeframe (i.e. 5-10 edits in the 5-10 minutes). When it hardly noticeable, it's still extra overhead for the user's network and for GC services

bekuno:
@okainov That's the problem with automatics. Many different use cases have to be considered. It will not be easier to implement with background tasks, you have then also to deside what to do on these two points.

PR #8116, opened 2020-02-01
add option to prevent waypoint extraction from note (fix #8078)

### LINKED ISSUE #8078: improve handling of automatic waypoint creation from notes
_(taken from: https://github.com/cgeo/cgeo/issues/8068#issuecomment-575601703 by @MagpieFourtyTwo)_

Nevertheless this "AutoZoom" (i. e. showing all defined waypoints of a cache) still renders the map view useless in some (but regrettably not really rare) cases, like e. g. in combination with c:geo's stubborn way of creating completely senseless waypoints from the personal note.

Example? Just create a personal note like this:

```
12-a-c=7	=> a+c=5	=> a=2 / c=3
36-b-d=26	=> b+d=10	=> b=2 / d=8

a.b.c.d@gmx.de
2.2.3.8@gmx.de (don't try, doesn't exist anyway ;) )

Final (doesn't exist, neither):
N48 51.234 E9 01.234
```
From this personal note two waypoints will be created automatically:
1. N48 51.234 E9 01.234 - which is perfectly fine
2. N02 12.000 E3 48.000 - which is *very* obviously *complete* rubbish

Usually you can suppress the automatic generating by putting a colon after the "N" like 
```
Calculation
N:48 51.234+17 E9 01.234-22
```
but in the example below even colons do not help ... :(

So, can we *please* invent a way of suppressing at least these completely senseless waypoints?
Perhaps it would be at least a quick'n dirty workaroud to suppress automatic generation when the extracted waypoint data gives a result which is more than let's say 10 km away from the base coordinates?  I guess, WPs more than 10 km away are rare, and if there is one, the user can create it using copy'n paste?

Which BTW at least for me would be the preferred way anyway, hence it could also be a solution to be able to switch this automatic extraction completely off by a config flag ...?

moving-bits:
(on per cache basis)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -860,6 +860,7 @@
     <string name="cache_menu_preventWaypointsFromNote">Prevent waypoints from note</string>
     <string name="cache_menu_allowWaypointExtraction">Allow waypoints from note</string>
     <string name="cache_personal_note_edit">Edit</string>
+    <string name="cache_personal_note_preventWaypointsExtraction">Prevent waypoints extraction</string>
     <string name="cache_personal_note_limit">Personal note limit</string>
     <string name="cache_personal_note_truncation">This personal note will be truncated after %d characters by Geocaching.com.</string>
     <string name="cache_personal_note_upload">Upload</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -857,6 +857,8 @@
     <string name="cache_log_images_title">Log images</string>
     <string name="cache_log_image_default_title">Photo</string>
     <string name="cache_personal_note">Personal note</string>
+    <string name="cache_menu_preventWaypointsFromNote">Prevent waypoints from note</string>
+    <string name="cache_menu_allowWaypointExtraction">Allow waypoints from note</string>
     <string name="cache_personal_note_edit">Edit</string>
     <string name="cache_personal_note_limit">Personal note limit</string>
     <string name="cache_personal_note_truncation">This personal note will be truncated after %d characters by Geocaching.com.</string>
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -170,6 +170,7 @@ public class Geocache implements IWaypoint {
     private final EnumSet<StorageLocation> storageLocation = EnumSet.of(StorageLocation.HEAP);
     private boolean finalDefined = false;
     private boolean logPasswordRequired = false;
+    private boolean preventWaypointsFromNote = false;
     private LogEntry offlineLog = null;
     private Integer eventTimeMinutes = null;
 
@@ -1547,7 +1548,7 @@ public class Geocache implements IWaypoint {
      * Detect coordinates in the personal note and add them to user defined waypoints.
      */
     public boolean addWaypointsFromNote() {
-        return addWaypointsFromText(getPersonalNote(), false, CgeoApplication.getInstance().getString(R.string.cache_personal_note));
+        return addWaypointsFromText(getPersonalNote(), false, CgeoApplication.getInstance().getString(R.string.cache_personal_note), false);
     }
 
     /**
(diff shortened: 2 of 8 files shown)


bekuno:
Can you integrate it in edit note dialog?

moving-bits:
> Can you integrate it in edit note dialog?

Which part would you like to be integrated in the edit note dialog?
The menu entry is intentionally put into the cache's menu, so that it is accessible from all cache details pages (especially the waypoint page).

bekuno:
In this frame if possible:

moving-bits:
@Bekuno:  What UI do you propose for this?

A third button or an overflow menu spring to my mind, but I'm not too happy with either of it:
- An overflow menu would sit at the right edge, where now the checkmark button is. - This might confuse some people which are used to confirm the note by pressing the upper right corner of the edit field. Also, this overflow menu would have only a single entry.
- A third button may be placed to the left of the two existing ones, but we would still end up with two different ways to access the same functionality, as IMHO we should have this functionality available at least on the waypoints page as well, where it would be (it is currently) a menu entry in the overflow menu.

(I haven't tested the technical aspects or checked available space yet.)

Maybe you have a good idea for how and where to place this item?

bekuno:
@moving-bits 
We had also a discussion in #7215 for a automatic upload/sync of notes.
This can in my mind on this issue.

How about to add an option frame in this window above or below the text, and within a option button to enable/disable the extraction of coords.
We could also change the small dialog to a full size window like the LogCacheActivity.

moving-bits:
> How about to add an option frame in this window above or below the text, and within a option button to enable/disable the extraction of coords.

I've extended the dialog box:

image

bekuno:
Fine :)

As text I would say "prevent waypoint extraction"
"from note" not necessary, we are in the note dialog. 
".. extraction" - because we want waypoints in notes :)

bekuno:
retest this please

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1557,12 +1558,14 @@ public boolean addWaypointsFromNote() {
      * @param updateDb if true the added waypoints are stored in DB right away
      * @param namePrefix prefix for waypoint names
      */
-    public boolean addWaypointsFromText(@Nullable final String text, final boolean updateDb, @NonNull final String namePrefix) {
+    public boolean addWaypointsFromText(@Nullable final String text, final boolean updateDb, @NonNull final String namePrefix, final boolean forceExtraction) {
```
bekuno: I do not understand of the parameter forceExtraction.
Why do we need them?

main/res/values/strings.xml
```diff
@@ -857,6 +857,8 @@
     <string name="cache_log_images_title">Log images</string>
     <string name="cache_log_image_default_title">Photo</string>
     <string name="cache_personal_note">Personal note</string>
+    <string name="cache_menu_preventWaypointsFromNote">Prevent waypoints from note</string>
+    <string name="cache_menu_allowWaypointExtraction">Allow waypoints from note</string>
```
bekuno: Where is the extra parameter "cache_menu_allowWaypointExtraction" used?

moving-bits: Those two lines are for the cache menu - enabling / disabling waypoint extraction from personal note.

main/src/cgeo/geocaching/models/Geocache.java
moving-bits: `addWaypointsFromText()` is called in two contexts - extracting from listing (needs `forceExtraction = true`) and extracting from personal note (should respect global setting, therefore `forceExtraction = false`).

moving-bits:
> As text I would say "prevent waypoint extraction"

You're right. In my changing the hard coded test-text to a resource I reused the text for the menu entry, which does not fit in this context. Changed the personal note edit label to this:

image

main/res/values/strings.xml
bekuno: but it is a checkbutton, isn't it?

moving-bits: No, the strings starting with "cache_menu_" refer to the new menu entry at the end of the cache menu, whereas the new string "cache_personal_note_preventWaypointsExtraction" is for the checkbox in the personal note edit dialog.

bekuno: Understand, ok.

PR #8117, opened 2020-02-01
add button 'add current location' to waypoint list (fix #8114)

### LINKED ISSUE #8114: Make new "Goto" system less click-intensive
Hi there,

I was out today and encountered the new 'Goto' system for the first time, I use this to record coordinates when I'm out setting caches. I have to say it isn't very nice to use.

To record a new point, I have to
1) tap 'Add waypoint'
2) tap on co-ordinates
3) tap on 'My coordinates'
4) tap on the 'tick'
5) tap on another 'tick'
6) scroll to the bottom on the list
7) tap on navigate to check I've got it right

On the old system, it was (IIRC)
1) tap on 'My coordinates'
2) tap on navigate to check I've got it right (which also saves the waypoint)

Can something be done to reduce the workload?

At least. can new waypoints be named to that they appear at the top of the list instead of having to scroll all the way to the end of the each time.

moving-bits:
Add a button "add current location" to create a new waypoint with current coordinates, standard waypoint type and title.

image

Also moves method `getDefaultWaypointName()` from `EditWaypointActivity` private method to `Waypoint` public static method to make it reusable.

### DIFF
--- a/main/res/layout/cachedetail_waypoints_header.xml
+++ b/main/res/layout/cachedetail_waypoints_header.xml
@@ -1,11 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Button xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/add_waypoint"
-    style="@style/button_full"
-    android:layout_margin="10dip"
-    android:clickable="true"
-    android:focusable="true"
-    android:padding="10dip"
-    android:text="@string/cache_waypoints_add"
-    tools:context=".CacheDetailActivity$WaypointsViewCreator" />
+    android:layout_width="match_parent"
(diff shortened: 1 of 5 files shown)


Lineflyer:
Sorry, already on the move to vacation still two comments:
- Does this only apply for the UDC or all caches waypoint tab?
- In the context of "Go to", what would be the benefit of adding a waypoint with my current location?

But anyway I am out for now.
Generally I would tend to not to modify to much for the "Go to". Sure its different than before, but it has the very same UI as our normal waypoint tab and thats IMHO a benefit. So all further improvements should be judged if they benefit both the "Go to" as well as the normal waypoint tab of a cache.

bekuno:
I would also see that as general solution for all waypoints.

@Lineflyer 
It is always usefull to set a waypoint when on a trip :)

moving-bits:
This is for all waypoints, UDC "go to" and other caches.
It's a shortcut for "save my current location as waypoint of the current cache". I also use this regularly as reminder for "need to come back to this place".

PR #8132, opened 2020-02-22
Allow solid background (fix #1361)

### LINKED ISSUE #1361: Black screen /Solid screen  instead of wallpaper in main-menu
From facebook (don't know if it is worth to implement it):

> Is it possible to put a black (maybe selectable option) background on the apps main screen? I would prefer that instead of seeing my homescreen wallpaper. 

I can understand somehow e.g. if you are using a quite contrast-less wallpaper.

moving-bits:
Add a toggle under settings - appearance to switch between transparent and solid background for mainscreen. Color of solid background is determined by c:geo's skin.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -59,6 +59,7 @@
     <string translatable="false" name="pref_webDeviceName">webDeviceName</string>
     <string translatable="false" name="pref_nativeUa">nativeUa</string>
     <string translatable="false" name="pref_skin">skin</string>
+    <string translatable="false" name="pref_transparentBackground">transparentBackground</string>
     <string translatable="false" name="pref_showaddress">showaddress</string>
     <string translatable="false" name="pref_useenglish">useenglish</string>
     <string translatable="false" name="pref_units_imperial">units</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -580,7 +580,9 @@
     <string name="init_openlastdetailspage">Last Details Page</string>
     <string name="init_summary_openlastdetailspage">Open details with last used page</string>
     <string name="init_skin">Light Skin</string>
-    <string name="init_summary_skin">Use light skin (Restart needed)</string>
+    <string name="init_summary_skin">Use light skin (restart required)</string>
+    <string name="init_transparentBackground">Transparent background</string>
+    <string name="init_summary_transparentBackground">Use transparent background (restart required)</string>
     <string name="init_address">Show Address</string>
     <string name="init_summary_address">Show address instead of coordinates on main screen</string>
     <string name="init_useenglish">Use English</string>
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -477,6 +477,11 @@
             android:title="@string/init_skin" />
         <CheckBoxPreference
             android:defaultValue="true"
+            android:key="@string/pref_transparentBackground"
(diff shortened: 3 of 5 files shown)


moving-bits:
Settings:
image

Solid background in dark skin:
image

Solid background in light skin:
image

moving-bits:
The last screenshot with solid background in light skin leads me to a question about the action bar:

Shouldn't we make the action bar's background solid black always? Looks much better, IMHO.

moving-bits:
> Shouldn't we make the action bar's background solid black always? Looks much better, IMHO.

Example with mainscreen's action bar background in solid black (light skin, c:geo background set to non-transparent):

image

bekuno:
The black/white does not fit to other backgrounds for dark/light theme.
The colors are defined in colors.xml, see
~~~
    <color name="background_dark">#FF000000</color>
    <color name="background_light">#FFFFFFFF</color>
~~~

moving-bits:
> The black/white does not fit to other backgrounds for dark/light theme.

Are you referring to the main screen's background or to the action bar's background?


> The colors are defined in colors.xml, see
> 
> ```
>     <color name="background_dark">#FF000000</color>
>     <color name="background_light">#FFFFFFFF</color>
> ```

FF000000 is solid black, FFFFFFFF solid white, so it's the same colors. (But you're right, I should use c:geo's predefined constants - will adapt the PR accordingly.)

bekuno:
You are right, that's black/white.
But we have some shades of grey that are used in different dialogues.
Unfortunately this is not used in the same way in all dialogs; some have black headline and gray background, some have gray headline and black background and many other color alternatives.
My favorite is the cache overview:
Screenshot_20200225-225421

moving-bits:
@bekuno 
Sorry, I still do not understand what you're trying to tell me :-(

The purpose for now is just to offer a solid background as additional option.

Do you ask me to change my existing PR in some way, or are you talking about something which should be done in a separate PR to enhance the visual experience further?

bekuno:
@moving-bits 
You had a implicit idea for an further change:

> Shouldn't we make the action bar's background solid black always? Looks much better, IMHO.

But here I disagree.
I'm no fan for the solid black/white colors. Therefore I showed this sample.

Back to the PR without this idea.
On the light skin the action bar, the nav bar, some text backgrounds are using a light grey. This looks fine. I checked once more now the colors.xml. It seems to me that 

~~~
    <color name="background_dark_notice">#FF191919</color>
    <color name="background_light_notice">#FFDFDFDF</color>
~~~

are the searched color codes.
What do you think to use these instead?

moving-bits:
> You had a implicit idea for an further change:
> > Shouldn't we make the action bar's background solid black always? Looks much better, IMHO.

Ah, now I understand what you're referring to. Thanks for your reply.

> But here I disagree.
> I'm no fan for the solid black/white colors. Therefore I showed this sample.

"Solid background" would be optional, and if changing the action bar's background to solid as well, it should depend on that setting, too. (So solid only, if option is set.)

> Back to the PR without this idea.
> On the light skin the action bar, the nav bar, some text backgrounds are using a light grey. This looks fine.

I'd chosen solid black, as on OLED screens pixel are switched to off for that, conserving some energy. But I'll look into it anyway and post new screenshots.

moving-bits:
I've tried the new colors:

Dark skin:
image

Light skin:
image

LGTM. What do you think?

Lineflyer:
I like the new approach!

bekuno:
It's not like I was expecting. But it's fine.

PR #8133, opened 2020-02-22
Merge branch release

moving-bits:


moving-bits:
@Lineflyer
Would this work to do a merge from release to master?
- I switched to my fork's local master branch, 
- created a new branch "merge_release" from it, 
- merged upstream/release into that
- and did conflict resolution.
This PR is created from the resulting branch state.

But: Github complains "This branch cannot be rebased due to conflicts" - yes, that's why I resolved them... - Or do I have to do the merge from a local copy of the original repo instead of from my fork?

Lineflyer:
This looks not OK to me, as the target branch for this PR is master.

Your branch merge_release should however now contain the conflict resolution, right? I need to practive anyway so I could try to take this file and use it for conflict resolution when directly merging the upstream branches..

rsudev:
The target branch should be master....

Lineflyer:
> The target branch should be master....

Jup, just edited my comment above.
@rsudev  Could you take a look if the commit sequence is correct. It looks basically OK as the original commit IDs are used (nothing squashed or rebased).

rsudev:
The other thing is, we don't want a rebase for this, so I guess this needs to be done locally and pushed directly

rsudev:
The sequence of commits is correct, but the conflict resolution as such has to be done in the merge commit. I have it ready locally and could push it.

rsudev:
I took a look onto the branch from @moving-bits as well he has some additional corrections in the merge commit (regarding the preference sections). I'd rather see those in a separate commit.

rsudev:
(likely done on release, to complete the change there as well)

moving-bits:
> I took a look onto the branch from @moving-bits as well he has some additional corrections in the merge commit (regarding the preference sections). I'd rather see those in a separate commit.

Ok, I'll prepare a PR for that. It separates the recently grown waypoint settings from "map data" into its own preference group.

moving-bits:
> The sequence of commits is correct, but the conflict resolution as such has to be done in the merge commit. I have it ready locally and could push it.

That would be fine, thanks. But: If I'm going to add the "preference splitting" PR mentioned above on release branch, release branch will be ahead of master again, so another merge is needed.

As I would like to learn how to do the merge, with or without conflict resolution: Can you explain to me what be the correct way to do it would be?

rsudev:
Besides the additional change in the merge commit everything was fine. The only thing is, that you can't apply that through a PR, you need to push the merge directly to master.
So in this case: git push upstream merge_release:master
(under the assumption, that 'upstream' points to the main c:geo repository).
So, after the separate change has landed on release, do the merge locally again. Than you can push to your repo and present as a PR to get it checked, if you want. But than it needs to be pushed directly as outlined above.

moving-bits:
@rsudev 
Thanks for explaining - I'll add that to my personal "git cheat sheet" :-)
I'll merge #8134 into release then, and try to merge release into master after that, following your description.

Issue #7848, opened 2019-09-14
Add "Hide archived" option to map settings


alan666notb:
**Is your feature request related to a problem?**
I like to keep all of my own caches in cgeo, even the ones that have been archived.
However, these are all shown on the map making it unusable.

**Describe the solution you'd like**
Under map-settings, there could be a "Hide archived" option (There is a "Hide disabled" option but I still want to see live caches that are disabled).

**Describe alternatives you've considered**
There could be a flag added to each list to say "Don't show caches in this list on the map".

MagpieFourtyTwo:
Would be great to have filters in live map, anyway. E. g. "Just show offline saved caches" ...

Lineflyer:
Relevant feature request....also refer to linked issues.

PR #8101, opened 2020-01-25
extend map settings to hide archived caches and/or certain waypoints

moving-bits:
Add a separate menu entry "hide" in maps' dot menu for the new options to
- hide archived caches
- hide original waypoints
- hide parking waypoints
- hide visited waypoints

and move the existing options to
- hide own/found caches
- hide disabled caches

to the same menu.

fixes #7848
fixes #5118

For updated screenshots see below.

Change: The impact from "exclude disabled" gets split: Until now it checked for both disabled and archived, now it only checks for disabled, and check for archived is separated. For migration purposes, the default value for "exclude archived" is set to the value of "exclude disabled".

Issue #5118, opened 2015-07-27
Option for hiding specific waypoint types


Lineflyer:
Quite often users ask to add more customization to the live map by being able to hide some types of waypoints:

Original Coordinates:
Most of the time the listing coords of mystery caches are "meaningless". It is good to have them as a waypoint (also for the function to restore them) but they might clutter the map

Parking waypoints:
While those are meaningful if you use a car and watch a single cache map, they might overcrowd the map in a dense city area on live map if you have saved several caches.

Visited waypoints:
If a waypoint of a cache is marked as "Visited" I probably have no need to see them on the map anymore.

I do also see a benefit in doing this, so I suggest the following:
In Setting - Map we do have a section called "Map Content" where also the waypoint limit is set.
Some checkboxes could be added at this point in the menu:
- Hide "Original Coordinates"
- Hide "Parking Waypoints"
- Hide waypoints marked as "Visited"

This setting should then be used on all live maps (but not on other maps e.g. when mapping waypoints of a single cache).

Follow up of #1530 (will be closed),

PR #8134, opened 2020-02-22
move waypoint settings to own preference group within map settings

moving-bits:


Lineflyer:
I stumbled upon this PR, which has recently been merged to `release`.
Isn't that underlying feature (hiding different waypoint types) on `master` only? Or is this something different?

I would like to suggest again(!) to open issues before submitting PRs and use the issue to explain the desired change and only after that submit the corresponding PR (with subject "Fix #xyz). That allows a more easy tracking of whats going on.

moving-bits:
@Lineflyer 
This is #8101 which fixes #7848 and #5118 (both are mentioned in the PR)

Lineflyer:
#8101 is perfectly fine. But does this PR here in any way depend on that PR? That was basically my question as this PR was merged to release and #8101 to master.

moving-bits:
> #8101 is perfectly fine. But does this PR here in any way depend on that PR? That was basically my question as this PR was merged to release and #8101 to master.

It's applicable to both master and release, as it's a follow-up to both #8101 and #8127, so rsudev suggested in https://github.com/cgeo/cgeo/pull/8133#issuecomment-589970698 to put it in release branch first.

Background: Both #8101 and #8127 introduce a new preference regarding waypoints, and together with the existing waypoint options it's clearer if grouped into it's own preference section (which I discovered during merge resolution - should have separated it then in the first place). #8101 is a new feature, therefore started on master branch, #8127 a fix for something already on release branch, therefore I targeted it to release.

Lineflyer:
OK, thanks for clarification and sorry, I missed that discussion during my vacation.
Will smoke test a new RC then.

PR #8136, opened 2020-02-23
downgrade jackson library

### LINKED ISSUE #8135: Nightly not starting on Android 4.3
I do have an Android 4.3 device here, on which an old nightly was working (did not record the version though). I just updated it to the newest nightly and now it fails to start ("Unfortunately c:geo has stopped").

Will try to find debug logs. Can anyone try with an emulator?

bekuno:
fix #8135

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -280,9 +280,10 @@ dependencies {
     implementation 'net.sf.geographiclib:GeographicLib-Java:1.50'
 
     // Jackson XML processing
-    def jacksonVersion = '2.10.2'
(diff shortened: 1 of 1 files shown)


moving-bits:
@bekuno
Thanks for fixing this

moving-bits:
Shouldn't we target this downgrade to release branch to have it in our next release?

Lineflyer:
The upgrade was only done on master.
Release is still on a even lower version.

moving-bits:
Ah, I see - that's fine, then. Thanks for pointing that out.

Issue #7497, opened 2019-04-14
OC Authentication not possible (Android 4.x)


mirabilos:
##### Detailed steps causing the problem:

- start c:geo
- see it cannot login to OC.DE
- try to reauthenticate via Settings → Services
- see it fail

##### Actual behavior after performing these steps:

it fails

##### Expected behavior after performing these steps:

it should login to OC.DE

##### Version of c:geo used:

NB 2019-04-13 375874c

##### Is the problem reproducible for you?
Yes

##### System information:

can do if needed

Lineflyer:
Thanks for your report.
Can you explain in more detail, what failure happens?
Does your browser open to present the login page of oc.de?

What Android version are you using (system information could be indeed helpful here).

mirabilos:
Hi Lars,

>Can you explain in more detail, what failure happens?
>Does your browser open to present the login page of oc.de?

OK, I’ve just upgraded to today’s NB, and the errors are the same.

In Browser, I’m logged into OC. I’m starting c:geo, and the status
line shows that GC is logged in, but OC.DE is “Login failed”.

I go to ⋮ → Settings → Services → Opencaching.de → Authorise c:geo again
then hit the “Start authorisation” button, then there’s a popup saying
“Waiting for opencaching.de…” (I think, it goes away very quickly) and
a… I think it’s called toast… saying “c:geo failed to initialise autho‐
risation process”.

Neither Browser pops up nor Fennec nor the dialogue in which I can
choose which browser to use.

>What Android version are you using (system information could be indeed helpful here).

LiquidSmooth 3.2 (KitKat), with microG (which, for some reason, is
not recognised as Google Play services, despite having UnifiedNLP
positioning and all).

--- System information ---
Device: HTC Desire (htc_bravo, htc_wwe)
Android version: 4.4.4
Android build: liquid_bravo-userdebug 4.4.4 KTU84Q c5fea05051 test-keys
c:geo version: 2019.04.15-NB-017b502
Google Play services: unavailable
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: magnetometer & accelerometer
Hide own/found: true
Map strategy: detailed
HW acceleration: enabled (default state)
System language: en_GB
System date format: yyyy/MM/dd
Debug mode active: no
System internal c:geo dir: /data/data/cgeo.geocaching (824.6 MB free) internal
[...]

PR #8139, opened 2020-02-25
Possible fix for #7497: Allow weaker cipher suites

jonas-koeritz:
Maybe this fixes opencaching.de issues for some ancient android versions that do not support TLSv1.2 or any of the more modern cipher suites proposed by the server.
This explicitly allows all the cipher suites offered by opencaching.de and appends this weaker CipherSpec to the list of available specs to try.

### DIFF
--- a/main/src/cgeo/geocaching/network/Network.java
+++ b/main/src/cgeo/geocaching/network/Network.java
@@ -40,6 +40,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.reactivex.Completable;
 import io.reactivex.Single;
 import io.reactivex.functions.Function;
+import okhttp3.CipherSuite;
 import okhttp3.ConnectionSpec;
 import okhttp3.FormBody;
 import okhttp3.HttpUrl;
@@ -52,6 +53,8 @@ import okhttp3.Request.Builder;
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.TlsVersion;
+import okhttp3.internal.Util;
+
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.lang3.CharEncoding;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

Lineflyer:
IMHO we should not put effort in solving this as long as we are not clear about the side effects on other Android versions/devices and other platforms.

In fact this problem seems limited to some 4.x devices trying to use opencaching.de. That sounds like a very limited amount of users.

moving-bits:
I would rather not allow weaker cipher suites - don't like the idea of weakening security for that. If someone's still using such old (and insecure) devices they will have to live with some limitations, IMHO.

jonas-koeritz:
This does not weaken the security for devices that support the higher security cipher suites significantly. It just allows weaker ciphers and doesn't force them.  I'm not a fan of supporting old Android devices forever. Maybe I should add this for older Android versions (<5) only. They don't have much to loose in terms of security anyways.

SammysHP:
Most browsers remove support for TLS 1.0 and 1.1 this year, so I expect that many servers start to drop support for these versions as well.

Lineflyer:
> Maybe I should add this for older Android versions (<5) only. They don't have much to loose in terms of security anyways.

From risk analysis perspective I would appreciate to limit the devices to only those old ones by a clear code passage (if <5 then), so the behaviour is unchanged on all other devices.

> Most browsers remove support for TLS 1.0 and 1.1 this year, so I expect that many servers start to drop support for these versions as well.

If this happens for gc.com (which a  major part of our users is using) this would be the time to just finally drop the support of Android 4.x as well.

bekuno:
retest this please

moving-bits:
Before merging, can the following be implemented?

> > Maybe I should add this for older Android versions (<5) only. They don't have much to loose in terms of security anyways.
> 
> From risk analysis perspective I would appreciate to limit the devices to only those old ones by a clear code passage (if <5 then), so the behaviour is unchanged on all other devices.

main/src/cgeo/geocaching/network/Network.java
```diff
@@ -52,6 +53,8 @@
 import okhttp3.RequestBody;
 import okhttp3.Response;
 import okhttp3.TlsVersion;
+import okhttp3.internal.Util;
+
```
bekuno, line 57: see https://ci.cgeo.org/job/cgeo%20pull%20request/2611/checkstyleResult/new/
remove extra separation in import group before 'org.apache.commons.compress.utils.IOUtils'

Lineflyer:
Needs to be redone and retargeted to branch `cgeo-legacy` 
See #8602

PR #8145, opened 2020-03-04
Check for userCorrectedCoordinates on Live Map, fixes #8100.

### LINKED ISSUE #8100: Handling of cache with corrected coords on server when using live map
I noticed an interesting difference (triggered by a user on support mail):

- Have a cache where corrected coords are stored on the server side
- This cache is not previously loaded or stored in c:geo

Case A:
- Open the live map on the website for this area. The cache will be shown on the corrected coords (incl. a puzzle marker)

Case B:
- Open the live map in c:geo for this area. The cache will be shown at the listing coords.

With our current live map implementation I would have expected that for case a and case b the cache is shown at the corrected coords. Only the marker in Case A is not supported. However this seems not to be the case.

Therefore the feature request is:
- Use this information about corrected coords on our live map
- If possible also use our corrected coords marker to identify those caches on the map

@jonas-koeritz: Do you have an explanation for this difference?

jonas-koeritz:


moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

Lineflyer:
Looks pretty straight forward to me.
Lets test this in the nightlies.

PR #8152, opened 2020-03-18
set heading pointer direction back to heading, ignore map bearing

### LINKED ISSUE #8150: Arrow on google map not rotating
**Describe the bug:**
Since version 2020.03.07 (where we introduced the rotating google map) some users claim, that the arrow on map does no longer rotate

**To Reproduce:**
- Switch off rotating maps (the affected users don't use this feature)
- Open live map with Google maps
- Rotate the device slowly

The arrow always points to a stationary direction on the map (seems not the identical direction for all affected users) and does not follow the device rotation.
When users switch to OSM the arrow rotation is OK.

The arrow should follow the device rotation/orientation

**Version of c:geo used:**
2020.03.07

No, I can't reproduce on my device. Not sure what prerequisites are needed to have this problem.

**System information:**
Will try to get more detail, but does not seem clearly related to a certain device type or Android version.

**Additional context:**
@moving-bits AFAIR you changed something in this part, due to my problem, that the arrow does not show conclusive direction while the map is rotating. Maybe we should revert and go back to the old code at this place (as we never had such issues reported before).

moving-bits:
(reverts #8031 / fix #8150)

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -211,7 +211,7 @@ public class GooglePositionAndHistory implements PositionAndHistory {
         positionObjs.addMarker(new MarkerOptions()
                 .icon(BitmapDescriptorFactory.fromBitmap(getLocationIcon()))
                 .position(latLng)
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

Lineflyer:
Thanks!.
Lets start from origin here and try to understand why it does not work on some devices.

PR #8157, opened 2020-03-28
Fix logging on geochaching.com

jgoestl:
Add payload as json and use cookie authentication

PR #8161, opened 2020-03-29
Update for AS 3.6.1 on release

### LINKED ISSUE #8141: AS 3.6
The AndroidStudio 3.6 is out.
It needs the gradle plugin 3.6.0.
And for this gradle has to updated at least to 5.6.4.

I think we should stay at the current AS 3.5.3 at least until the next c:geo release is released.

bekuno:
fixes #8141 for release

- Update gradle plugin to 3.6.1
- Update gradle to 5.6.4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:3.5.3'
+        classpath 'com.android.tools.build:gradle:3.6.1'
 
         // un-mocking of portable Android classes
         classpath 'de.mobilej.unmock:UnMockPlugin:0.7.3'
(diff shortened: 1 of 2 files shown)


bekuno:
This should merged to release before PR #8157 to have changes to the infrastructure are split from solving the problem.

bekuno:
@moving-bits 
These are the minimum requirements vor AS 3.6.x. This should be enought for release.
The PR #8141 has more changes.

Lineflyer:
I have seen this PR to late.

I don't have a clue about gradle, but if you want to update it on both branches it might make sense to only apply this to `release` and then merge upstream to `master`?

Otherwise there will be conflicts when merging branches. 
But now you already applied gradle changes on `master` and I already merged #8157 and merged upstream trying to resolve the conflict by taking over Gradle-wrapper.properties from `release`.

Can you check if the config is still fine or whether I destroyed everything?

bekuno:
@Lineflyer I merged #8141 to master but the idea was to sepatate the gradle updates from problem solving in PR #8157. My mistake :-(

bekuno:
On release only sorting in gradle-wrapper.properties is changed.
We can let them so.

PR #8173, opened 2020-04-01
configure undobar

### LINKED ISSUE #8167: Crash when modifying filtered list
To Reproduce:
- Open a list with many caches
- Set a filter so that a subset is remaining (I used: Status=archived)
- On the filtered list, select Menu - Manage caches - Delete all

Result:
Filtered caches are deleted as expected, but app crashes immediately afterwards

Expected:
No crash, filtered caches deleted

Version:
2020.03.29-NB1

bekuno:
fixes #8167 Crash when modifying filtered list

Fixing the problem:
- add metadata configuration for mavenCentral, because undobar has no *.pom
- reset including undobar library from external repositories

updates:
- cleanup the repository sources
- set undobar implementation to api15
- remove nineoldandroids, not more necessary for api15 of undobar
- add api15 aar files to undobar lib folder


bekuno:
retest this please

moving-bits:
@bekuno Thanks for tackling this issue!

PR #8182, opened 2020-04-08
Rename "OSM: Map" to "OpenStreetMap Standard" (fix #8181)

### LINKED ISSUE #8181: Rename OSM menu entry
From https://github.com/openstreetmap/operations/issues/371#issuecomment-598016057

> Could you change the "OSM: Map" to "OpenStreetMap Standard" or something that has the full name?

moving-bits:
per request of OSM operations team, see https://github.com/openstreetmap/operations/issues/371#issuecomment-598016057

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -760,7 +760,7 @@
     <!-- map sources -->
     <string name="map_source_google_map">Google: Map</string>
     <string name="map_source_google_satellite">Google: Satellite</string>
-    <string name="map_source_osm_mapnik">OSM: Map</string>
+    <string name="map_source_osm_mapnik">OpenStreetMap</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="map_source_osm_offline_combined">Combined (Offline)</string>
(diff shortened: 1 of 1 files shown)


Lineflyer:
Do we need to name it standard, while there is no other OSM based map around?Von Samsung-Tablet gesendet
null

moving-bits:
We could call it just "OpenStreetMap" as well. What do you think?

jakedot:
I think the "Standard" was something the OSM guys wanted to have explicitly, so if possible we should honor it?

moving-bits:
> I think the "Standard" was something the OSM guys wanted to have explicitly, so if possible we should honor it?

They requested `Could you change the "OSM: Map" to "OpenStreetMap Standard" or something that has the full name?`, so I guess an "OpenStreetMap" would be ok for them as well.

jakedot:
The question is if that is the full name as defined by them, but I'd guess so. Maybe @pnorman can clarify?

pnorman:
OSM should be expanded to OpenStreetMap. Standard isn't required.

Issue #8180, opened 2020-04-07
make onboarding with offline maps easier for new users of c:geo


moving-bits:
One of the steepest hurdles for new users of c:geo is installing an offline map, IMHO. Thus probably most new users (except technically very savant) stick to Google maps and miss one of the greatest features of c:geo - offline caching.

There have been proposals for downloading maps in the past (eg #6072, #7529), but they are very sophisticated approaches which try to automate everything and cover a wide range of use cases. Sadly, they are also a bit more heavy to implement, not only on our side but on providers' side as well, and nobody has done it yet.

I propose a more basic approach with the focus on simplifying onboarding for new users of c:geo:
- Skip themeing, only support downloading map files for this onboarding feature. Therefore use a map source which does not need a separate theme file. (This would be the hs-esslingen mirror of download.mapsforge.org)
- Direct the user to the relevant directory, let them download the map file using the browser, and make c:geo integrate the downloaded file...
- ... which would include copying the file to c:geo map directory, set the map provider and current map file accordingly.

Implementation wise there are three areas of work:
- [x] Offer a link to the map download site
- [x] Recognize downloaded map files and move them to c:geo maps dir (current map dir, if already set, or new default one)
- [x] Set map handler and current map file accordingly

I've started working on it. Current state is:

a) Initiate download via Settings => Map => Download basic offline maps:
![image](https://user-images.githubusercontent.com/3754370/78711262-fbdc9480-7916-11ea-8c0c-d571932ed6f6.png)

b) Info screen:
![image](https://user-images.githubusercontent.com/3754370/78711297-0a2ab080-7917-11ea-93c9-fc4f9bb403d9.png)

c) Detecting that a map already exists under the same file name:
![image](https://user-images.githubusercontent.com/3754370/79076617-c90e1400-7cfb-11ea-8337-44032687df2a.png)

d) cancellable progress dialog during transfer:
![image](https://user-images.githubusercontent.com/3754370/79076626-e642e280-7cfb-11ea-8df3-d7f4233deaa7.png)

e) confirmation dialog after successful installation:
(recognition of file name depends on Android version and map source; may be a more generic name like "5.map" or "msf:27.map")
![image](https://user-images.githubusercontent.com/3754370/78916095-41bb6900-7a8d-11ea-8539-4db4a0731a2f.png)

Alternatively you can tap on a map file stored locally. This would enter the process shown above at step c).

[...]

PR #8183, opened 2020-04-08
WIP ease offline maps onboarding

moving-bits:
This is a WIP for beta testing only - do not merge

- integrate locally stored OSM map file into c:geo
  - move to map dir (cancellable)
  - set as current map source
- add preference link to download v3 OSM files
- improve handling of local files (forward detected file types internally, show an info message for the rest)
- smoke tested with different file types and Android versions from 4 to 10

For details and current status see #8180.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -230,26 +230,28 @@
             android:name=".HandleLocalFilesActivity"
             android:exported="true"
             android:label="@string/localfile_intenttitle">
-            <!-- accept local files for further filtering -->
+            <!-- Accept local files for further filtering -->
             <!-- Android itself does not support filtering local files by file extension, and MIME Types are mostly unset or generic for the files c:geo uses -->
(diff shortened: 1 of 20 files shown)


moving-bits:
(resolved merge conflicts introduced by moving the non translatable strings)

Lineflyer:
It makes IMHO no sense to make use of V3 maps as these maps are no longer updated since a longer time on the mapsforge servers. That will in return give complaints of outdated maps. 
You probably used that version to be compatible to WhereYouGo, but IMHO it makes more sense to rather make a c:geo solution with newer maps and wait with WhereYouGo until we have support for higher versions in there.

moving-bits:
Ok, shall I make it download a v5 map instead, then, or shall I make this dependent on the "use old Mapsforge v3" setting?

Lineflyer:
IMHO the setting to use old mpasforge is for experienced users (and might hopefully be removed rather soon).
Therefore I suggest to go for V5 as a straight default.

moving-bits:
Rebased to current master and changed map dir to v5 maps.

moving-bits:
Rebased to current master

PR #8186, opened 2020-04-10
Add support for multilingual OSM maps (fix #8185)

### LINKED ISSUE #8185: Support multilingual OpenStreetMap maps
Starting with v4 format OpenStreetMap maps support multiple languages. Labels for city names etc. can be in any of the supported languages.

When opening a map file you can name a preferred language. Labels will be returned in that language, if available, or in default language of that map, else.

It would be nice if c:geo supported this as well.

moving-bits:
If the opened OSM offline map file supports multiple languages, a new menu entry "Map language" gets added to the map's menu, having the supported languages as entries of the sub menu.

image

Selecting one of those languages restarts the map, setting labels accordingly to the set language, if available:

Language set to "fr":
image
(The red box is just for better visualization in this info, it's not part of the actual map rendering!)

Language set to "de":
image

Selected language gets saved to preferences.
If the map does not support the language set in preferences, it switches to default language.
If a map does not support any languages, the menu entry "Map language" is not shown.

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -199,6 +199,18 @@
         app:showAsAction="ifRoom|withText">
     </item>
     <item
+        android:id="@+id/menu_select_language"
+        android:title="@string/map_language"
+        android:visible="false"
+        app:showAsAction="ifRoom|withText">
+        <menu>
+            <group
+                android:id="@+id/menu_group_map_languages"
+                android:checkableBehavior="single" >
+            </group>
+        </menu>
+    </item>
+    <item
         android:id="@+id/menu_as_list"
         android:icon="@drawable/ic_menu_agenda"
         android:title="@string/map_as_list">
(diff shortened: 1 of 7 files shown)


SwissPoPo:
I miss the menu entry for the default language. 
If I have have been selected one of the additional languages of the map then I have no possibility to go back to the default local language.
"default" should be added to the list.

moving-bits:
If the language you have selected currently is not supported by a map, it automatically falls back to the map's default language (whatever language that is - you cannot query the map for that info).

SwissPoPo:
The automatic failback is nice but does not solve the problem that I see here. 
Not every map provider does include the local language as additional language. The maps from mapsforge itself offers only "en,de,fr,es" for every map worldwide.
So, an map from Italy has the "default" language and "en,de,fr,es". No Italian as additional language. In this case the default language is Italian. If I select one of the additional languages  I cannot go back to default languange. That means no local language, no Italian anymore.
An menu entry to reset language selection to the default language is needed. Take a look to Cruiser, Orux, or Locus how they do it.

moving-bits:
@SwissPoPo 
Thanks for explaining the background. I've opened a separate issue for this for easier tracking (see #8814)

PR #8187, opened 2020-04-11
Support other OSM based online map sources (fix #8148)

### LINKED ISSUE #8148: Support other OSM based online map sources
WhereYouGo supports more tile severs then only what we called "OSM:Mapnik":

- OSM:Mapnik
- OSM.de
- CyclOSM (could be a good replacement of the retired OpenCycleMap)
- Public Transport
- Esri world street
- Esri world Imagery (alternative to Google Satellite)

Tile source info can be found in cgeo/WhereYouGo code.

I have no idea whether these tile sources can really be freely used, but if this is the case we should support those (or a subset) also in c:geo.

moving-bits:
Extend our supported OSM based online maps by
- openstreetmap.de and
- cyclosm.org

Avaialble for NewMap only.
Also renamed "OpenStreetMap" to "OpenStreetMap.org" to distinguish from "OpenStreetMap.de".

Example screen shots:

OpenStreetMap.org (aka "OSM:Mapnik")
image

OpenStreetMap.de:
image

CyclOSM.org:
image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -760,7 +760,9 @@
     <!-- map sources -->
     <string name="map_source_google_map">Google: Map</string>
     <string name="map_source_google_satellite">Google: Satellite</string>
-    <string name="map_source_osm_mapnik">OpenStreetMap</string>
+    <string name="map_source_osm_mapnik">OpenStreetMap.org</string>
+    <string name="map_source_osm_osmde">OpenStreetMap.de</string>
+    <string name="map_source_osm_cyclosm">CyclOSM</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="map_source_osm_offline_combined">Combined (Offline)</string>
     <string name="init_sendToCgeo">Send to c:geo</string>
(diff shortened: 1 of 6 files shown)


moving-bits:
"Help wanted":

Current implementation seems to work fine so far, with one caveat which I haven't been able to tackle yet: After switching between two of the three OSM online services the map stays blank until you close and reopen the map. (This does not happen when switching from an online based map to an offline based map and vice versa.) So there seems to be one additional step needed.

SammysHP:
Do you have a permission for us to use these maps?

cyclosm.org:

> A Tile Map Service is available at <code>https://{s}.tile-cyclosm.openstreetmap.fr/cyclosm/{z}/{x}/{y}.png</code> under a <a href="https://operations.osmfoundation.org/policies/tiles/">fairuse tile policy</a>.

https://operations.osmfoundation.org/policies/tiles/:

> Heavy use (e.g. distributing an app that uses tiles from openstreetmap.org) is forbidden without prior permission from the Operations Working Group.

openstreetmap.de has something similar:

https://www.openstreetmap.de/germanstyle.html

> Das Einbinden der Kacheln in eigene Webseites können wir aufgrund der begrenzten Bandbreite leider nur für nichtkommerzielle Zwecke erlauben. Kleinere Webanwendungen, wie z.B. der Anfahrtsplan auf der Firmenwebseite dürfen die Kacheln selbstverständlich verwenden. Bitte beachten sie auch die Tile usage policy.

moving-bits:
> Do you have a permission for us to use these maps?

Both new sources refer to the identical policy we already have for our current OSM online tile server, and attribution notes have been extended to list the two new sources.

moving-bits:
retest this please

SammysHP:
> Both new sources refer to the identical policy we already have for our current OSM online tile server

And nobody said that it is OK what we do. ;)

moving-bits:
> And nobody said that it is OK what we do. ;)

The comment https://github.com/openstreetmap/operations/issues/371#issuecomment-598016057 can be read as a "temporarily ok, until we say otherwise".

This PR intends to reduce load on that server by distributing it over several different sources, hopefully resulting in low/acceptable load for all servers involved.

#8180 tries to help new users to configure offline map sources more easily to make more users of c:geo use offline maps.

moving-bits:
retest this please

PR #8198, opened 2020-04-13
Add helper for WhereYouGo offline map onboarding

moving-bits:
This is the c:geo part for https://github.com/cgeo/WhereYouGo/pull/146


moving-bits:
(resolved branch conflicts introduced by moving the non translatable strings)

PR #8208, opened 2020-04-13
Clean up old crowdin scripts

Lineflyer:
Although these scripts have made good work in the past, I think we can remove them now.
In case we might need anything, they are still in the git version history.


bekuno:
I would not remove the helper scripts in project/localization.
They are helpfull to check if we have unnecessary strings, e.g. deleted in the code.

Lineflyer:
> They are helpfull to check if we have unnecessary strings, e.g. deleted in the code.

Isn't this covered in our lint checks?
It should be AFAIR and once they are deleted from the source, crowdin will remove them from all localization's.

bekuno:
I

> Isn't this covered in our lint checks?

Will check it the next days.

bekuno:
Compiler/AS and the basic lint check are able to find such issues.

PR #8209, opened 2020-04-13
 supported language list

bekuno:
After the crowdin work we need some adjustments.
- languages ar=Arabic, fil=Filipino and in=Indonesian are at >74% translated and will be activated
- add info for other included but not activated languages: is=Icelandic, iw=Hebrew, zh=Chinese Simplified

Configure the same for cgeo-contacts

### DIFF
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -21,6 +21,10 @@ android {
         minSdkVersion 16
         //noinspection OldTargetApi
         targetSdkVersion 28
+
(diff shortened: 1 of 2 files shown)


bekuno:
@Lineflyer There were suggestion to use iw instead of he somwhere in the lint suggestions. (Did not find it now) But this is not the current situaltion in ISO 639 (see https://de.wikipedia.org/wiki/Liste_der_ISO-639-1-Codes). 
Did you configured something for the gradle import to c:geo? We should have test the situation at least when activating the language in build.gradle.

PR #8211, opened 2020-04-14
move non translatable strings to according file and mark them as non-translatable

moving-bits:
- move non translatable strings to according file and mark them as non-translatable
- also format strings in strings_not_translatable to have the `translatable="false"` attribute in front consistently to make reading and editing of that file easier

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -368,7 +368,6 @@
     <string name="caches_move_all">Move all</string>
     <string name="caches_copy_selected">Copy selected</string>
     <string name="caches_copy_all">Copy all</string>
-    <string name="caches_map_locus">Locus</string>
     <string name="caches_map_locus_export">Export to Locus</string>
     <string name="caches_filter">Filter</string>
     <string name="caches_filter_title">Filter by</string>
@@ -502,8 +501,6 @@
 
     <string name="settings_goto_url_button">more …</string>
 
-    <string name="settings_title_gc">Geocaching.com</string>
-    <string name="settings_title_ec">Extremcaching.com</string>
     <string name="settings_activate_gc">Activate</string>
     <string name="settings_activate_ec">Activate</string>
     <string name="settings_gc_legal_note">With using the service of geocaching.com, you accept the Groundspeak Terms of Use.</string>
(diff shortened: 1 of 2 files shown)


moving-bits:
First try on moving all non-translatable string to the according source file and mark them as non-translatable. If you find more strings which should be marked as non-translatable, or strings wrongly sorted to this, please add as comment here, I will update the PR then.

main/res/values/strings.xml
```diff
@@ -1539,13 +1506,10 @@
...
     <!-- next things -->
     <string name="quote">To make geocaching easier, to make users lazier.</string>
-    <string name="powered_by">carnero</string>
     <!-- Note: Links here are just for appearance. See AboutActivity to make changes -->
     <string name="support_title">Support</string>
     <string name="website_title">Website</string>
-    <string name="facebook_title">Facebook</string>
     <string name="facebook_link"><a href="">c:geo page</a></string>
-    <string name="twitter_title">Twitter</string>
     <string name="market_title">Google Play</string>
```
Lineflyer, line 1513: Maybe also Google Play?

main/res/values/strings.xml
```diff
@@ -1651,13 +1614,6 @@
         <item quantity="other">Add this cache to favorites (%d remaining)</item>
     </plurals>
 
-    <!-- distance units -->
-    <string name="unit_m">m</string>
-    <string name="unit_km">km</string>
-    <string name="unit_ft">ft</string>
-    <string name="unit_yd">yd</string>
-    <string name="unit_mi">mi</string>
-
```
Lineflyer, line 1660: For these units I am unsure, whether they could be different in a different alphabet, like e.g. greek, cyrillic, etc?

main/res/values/strings.xml
moving-bits: I had not included Google Play as I was unsure whether it has the same name in all languages. - Just cross checked with current translation files: At least in `values-ko` variant it differs (see string `market_title`).

main/res/values/strings.xml
moving-bits: Ok, will revert this part - most languages use the original abbreviations, but some languages have translated the full names.

moving-bits:
> Mainly looks good to me. See my two comments inline

Thanks for cross-checking. I updated the PR and moved the unit strings back to the translatable strings, but left `market_title` unchanged, as we have at least one translation for it already.

Lineflyer:
OK, I will merge this if tests are green. Thanks for taking this point.

Lets see what Github does with branch ` moving-bits:issue_8206` after merge. Not sure if the new deletion setting also applies to remote repositories.

Lineflyer:
The auto deletion seems only to apply to the cgeo/cgeo repository. Remote branches are not autodeleted (which makes sense as it should be the responsibility of the repo owner).

moving-bits:
> The auto deletion seems only to apply to the cgeo/cgeo repository. Remote branches are not autodeleted (which makes sense as it should be the responsibility of the repo owner).

GitHub has been deleting remote branches earlier, starting at some point in (end of?) 2019, IIRC. But maybe only when I merged my own PR, not sure.

Lineflyer:
>  GitHub has been deleting remote branches earlier, starting at some point in (end of?) 2019, IIRC. But maybe only when I merged my own PR, not sure.

Yes, you should see an option in this PR to delete your local branch. I do not see it.

bekuno:
The unit strings are as
<string translatable="false" name="unit_yd">yd</string>

But this was planned to revert, isn't it?

moving-bits:
> The unit strings are as
> yd

That should be fine, as it's been like that before (and some language files indeed have translations for that).

But you triggered a different check for me, and I discovered a mistake I made while reverting the distance units back to `strings.xml`: I had not removed the `translatable="false"` flag. => Updated my PR now

bekuno:
@moving-bits 
With `back ticks` is my comment better understandable, see now https://github.com/cgeo/cgeo/pull/8211#issuecomment-613405339

bekuno:
> GitHub has been deleting remote branches earlier, starting at some point in (end of?) 2019, IIRC. But maybe only when I merged my own PR, not sure.

Maybe you had activated such autamatic for your repo? 
I have to delete all my merged PR manually.

moving-bits:
> With `back ticks` is my comment better understandable, see now [#8211 (comment)](https://github.com/cgeo/cgeo/pull/8211#issuecomment-613405339)

Thanks. Ok, so we've stumbled upon the same. This is fixed with #8212.

moving-bits:
> > GitHub has been deleting remote branches earlier, starting at some point in (end of?) 2019, IIRC. But maybe only when I merged my own PR, not sure.
> 
> Maybe you had activated such autamatic for your repo?

Not that I know of. When I started end of 2018, I had to delete the no longer needed branches locally (clone of my fork) and in my fork, both manually, same as you. (Therefore I created a little script to do this for me.)

At some point in time in 2019 GitHub started deleting the no longer needed branches in my fork automatically, so I only had them to manually delete locally since then.

moving-bits:
I just tested my theory of "this happens when I merge my own PR" with #8212.

=> Yes, this seems to be the cause.
After confirming the merge, GitHub automatically deleted my fork's branch:

image

bekuno:
Maybe you have forgotten your change at that point in your repo setting:
grafik
I is a nice feature :-)

PR #8210, opened 2020-04-13
Implemented Filtering by Country (fix #5257)

jakedot:
Unconstrained, over all Countries extracted from the Location field on caches.

Fixes #5257.

PR #8218, opened 2020-04-17
Implemented Filtering by Country

### LINKED ISSUE #5257: Filter by Location (country)
From support mail:

Hello guys. Please add filter by Location (country). Could be very usefull to sort stored caches to different lists. Thanks.

jakedot:
Unconstrained, over all Countries extracted from the Location field on all caches.

Fixes #5257 
See #8210 for history.

### DIFF
--- a/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
+++ b/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
@@ -1,0 +1,84 @@
+package cgeo.geocaching.filter;
+
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.storage.DataStore;
+
+import android.os.Parcel;
+
+import androidx.annotation.NonNull;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class CountryFilterFactory implements IFilterFactory {
+
+
+    static class CountryFilter extends AbstractFilter {
+        private String location = null;
+
+        public static final Creator<CountryFilter> CREATOR
+                = new Creator<CountryFilter>() {
+
+            @Override
+            public CountryFilter createFromParcel(final Parcel in) {
+                return new CountryFilter(in);
+            }
+
+            @Override
+            public CountryFilter[] newArray(final int size) {
+                return new CountryFilter[size];
+            }
+        };
+
+        CountryFilter(final String location, final String title) {
+            super(title);
+            this.location = location;
+        }
+
+        protected CountryFilter(final Parcel in) {
+            super(in);
+            location = in.readString();
+        }
+
+        @Override
+        public void writeToParcel(final Parcel dest, final int flags) {
+            super.writeToParcel(dest, flags);
+            dest.writeString(location);
+        }
+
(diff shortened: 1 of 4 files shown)


moving-bits:
ok to test

Lineflyer:
ok to test


@moving-bits @bekuno
We need to look up why you are not allowed to talk to the CI.

main/src/cgeo/geocaching/filter/CountryFilterFactory.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    @NonNull
+    public List<IFilter> getFilters() {
+
+        final Map<String, IFilter> filters = new HashMap<>();
+
+        final String separator = ", ";
+
+        // TODO: use DataStore.getStoredLocationsForSet(Set<String>)
```
jakedot, line 68: Here, I need to somehow get the list of current geocodes.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -3408,6 +3411,86 @@ public static void saveLists(final Collection<Geocache> caches, final Set<Intege
...
+        } catch (final SQLiteDoneException ignored) {
+            // Do nothing, it only means we have no information on the cache
+        } catch (final Exception e) {
+            Log.e("DataStore.getAllStoredLocations", e);
+        }
+
+        return locations;
+    }
+
+    public static Set<String> getStoredLocationsForSet(Set<String> geocodes) {
```
jakedot, line 3460: this should be able to utilize the DB for getting all locations for a set of geocodes, this way we can skip creating Geocache objects

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -3408,6 +3411,86 @@ public static void saveLists(final Collection<Geocache> caches, final Set<Intege
         }
     }
 
+    public static Set<String> getStoredLocationsInList(final int listId) {
+
```
jakedot, line 3415: this could get all locations in a stored list I guess. may be dropped if wholly unnecessary.

moving-bits:
@jakedot 
Please follow the "Details" link of the failed "default - Build finished." line to see the remaining CheckStyle issues (or other things that lead to a broken build).

main/src/cgeo/geocaching/filter/CountryFilterFactory.java
jakedot: @moving-bits any idea?

moving-bits: @jakedot  Nothings springs to my mind immediately, but I haven't looked into the source code yet. Maybe have a look at the position where the filter gets instantiated - can you add the set as a new parameter there?

jakedot:
It should be possible, but I still have no clue where the set can be
retrieved from. The filter is created in the FilterActivity, which doesn't
seem to have it. I guess we would need to pass it in with an Intent?

moving-bits:
So filters are static? Then we need to inject the current set later, when the list is created. (if possible)

PR #8227, opened 2020-04-18
WIP/testing only: reduce calls to online map search

moving-bits:
- caching of map search results
- use factor 3 bounding box for viewport being searched

If you activate debug mode in c:geo you'll see messages for each call to searchByViewport with info whether a cached result could be reused or a new call has been issued.

### DIFF
--- a/main/src/cgeo/geocaching/connector/ConnectorFactory.java
+++ b/main/src/cgeo/geocaching/connector/ConnectorFactory.java
@@ -33,6 +33,7 @@ import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Trackable;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.AndroidRxUtils;
+import cgeo.geocaching.utils.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -109,6 +110,11 @@ public final class ConnectorFactory {
     private static final Collection<ISearchByFinder> SEARCH_BY_FINDER_CONNECTORS = getMatchingConnectors(ISearchByFinder.class);
 
     private static boolean forceRelog = false; // c:geo needs to log into cache providers
+
+    // for viewport searchResult caching
+    private static int counterSearchByViewport = 0;     // @todo delete this line after testing
+    private static Viewport lastViewportUsed = null;
+    private static SearchResult lastSearchResult = null;
 
     private ConnectorFactory() {
         // utility class
@@ -334,7 +340,18 @@ public final class ConnectorFactory {
     /** @see ISearchByViewPort#searchByViewport */
     @NonNull
     public static SearchResult searchByViewport(@NonNull final Viewport viewport) {
-        return SearchResult.parallelCombineActive(searchByViewPortConns, connector -> connector.searchByViewport(viewport));
+        if (null != lastViewportUsed && lastViewportUsed.includes(viewport)) {
+            Log.d("searchByViewport: saved searchResult reused");            // @todo delete this line after testing
+            return lastSearchResult;
+        }
+
+        // @todo delete next two lines after testing
+        counterSearchByViewport++;
+        Log.d("called searchByViewport #" + counterSearchByViewport);
+
+        lastViewportUsed = viewport.resize(3);
+        lastSearchResult = SearchResult.parallelCombineActive(searchByViewPortConns, connector -> connector.searchByViewport(lastViewportUsed));
+        return lastSearchResult;
     }
 
     @Nullable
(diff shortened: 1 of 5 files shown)


Lineflyer:
Short feedback for today only (need to carefully double-check whether I did something wrong):
Without running the testsuite in details, I have to say, that I do see exactly no difference compared to 2020.03.30. Still a lot of requests for short movements.

Note:
I had to use OSM instead of Google as the debug version has no GMaps key, but IMHO OSM behaved like GMaps in 2020.03.30.

Lineflyer:
Only a short feedback for today:
Looks promising. Lot less requests. Could not test the rotation case beause of missing GMaps key.
I ran into a situation (after zooming out a lot and moving the map) that no more new requests are sent at all until I restarted the live map.

Need to take a closer look tomorrow.

moving-bits:
Thanks for smoke testing.

> I ran into a situation (after zooming out a lot and moving the map) that no more new requests are sent at all until I restarted the live map.

There might be one problem with this caching approach if the user zooms out too much: The maximum result set size is 500 IIRC, but due to the new caching there will be no further searching within this large area, even if not all caches in that bounding box have been loaded. - So we need some criteria when to redo a search even if still covered by the cached viewport. Maybe time-based (30 secs), and only if maximum results had been retrieved for gc.com (don't know if we can check this at that stage)?

Lineflyer:
> There might be one problem with this caching approach if the user zooms out too much: The maximum result set size is 500 IIRC, but due to the new caching there will be no further searching within this large area, even if not all caches in that bounding box have been loaded. - So we need some criteria when to redo a search even if still covered by the cached viewport. Maybe time-based (30 secs), and only if maximum results had been retrieved for gc.com (don't know if we can check this at that stage)?

Yes, sounds likely this was the situation I experienced:
Zooming out to Europe I get a bubble of 500 caches in the center (logical result of requesting 500 around the location). As you probably store the viewport requested (which is much larger than the virtual viewport of returned caches) there will probably only new requests once I read America ;)

For me personally c:geo is no tool for such low zoom exploring, but many users do this. So we need to think about a suitable solution. For me short term it could rather be to limit the zoom level or just accept this strange behaviour for now until we have a good solution developed.

bekuno:
We could define a maximum area (eg. 20x20km) on which we will request for caches. If the viewport is greater than this area then we present a message instead to query for caches.

moving-bits:
> We could define a maximum area (eg. 20x20km) on which we will request for caches. If the viewport is greater than this area then we present a message instead to query for caches.

Cache density is quite different around the world, so a fixed size might be difficult. Within some densly populated areas like Berlin / Munich / Ruhrgebiet even 20x20 km might lead to more than 500 caches, whereas in sparely populated areas even 250x250 km might work.

Lineflyer:
I quickly checked the website:
The search maps requires active interaction (i.e. clicking the button "Search geocaches") and then only loads some hundred around the center coords (similar to c:geo in fact).
The browse map uses compact mode and shows more but it seems no realistic view on low zoom but just kind of a interpolation and also just skipping part of the map (e.g. australia when viewing the world centered on Europe).

moving-bits:
Maybe we can combine both approaches:
- If zoom level < 11 (GMv2) / 14 (NewMap)
- and searchResult.length >= 500
- then do a reload when requested viewport has changed
?

moving-bits:
I updated this PR to reset the cache under the condition that
- zoom level is at max 11 (GMv2) or 14 (NewMap), and
- last search result has at least 400 geocodes, and
- viewport has changed

moving-bits:
@jonas-koeritz 
Do you have time for a short review of this PR? Thanks.

jonas-koeritz:
Looks OK to me.
This is another good candidate for beeing done in Room/LiveData ;)

Can somebody invest some brain time into a good caching strategy for live map data? What if we load caches for an area, store them into the (Room) database, store the area we queried and a timestamp in the database as well and check if we already have that inside the database before fetching network data.
The database size would be manageable just dumping the live map data to the database, full caches can still be loaded individually.
Areas that match the currently viewed map could be updated if the data is too old and we could even let users mark an area to keep it up to date at regular intervals in the background (like: tell me about new caches in that area, check once per week).

Lineflyer:
I will merge this and test on a RC during this week.

moving-bits:
> Looks OK to me.

Thanks for cross checking.

> Can somebody invest some brain time into a good caching strategy for live map data? What if we load caches for an area, store them into the (Room) database, store the area we queried and a timestamp in the database as well and check if we already have that inside the database before fetching network data.
> The database size would be manageable just dumping the live map data to the database, full caches can still be loaded individually.
> Areas that match the currently viewed map could be updated if the data is too old and we could even let users mark an area to keep it up to date at regular intervals in the background (like: tell me about new caches in that area, check once per week).

Sounds promising generally, although I have no idea at hand how to track which parts have been updated already. E. g.: Let's say you store an area N50 20.0 E7 10.0 to N50 22.0 E7 10.5 at one point in time. Some time later you view (and store) N50 21.0 E7 10.2 to N50 23.0 E7 11.0 (overlapping areas, but not fully included). Now you would need to split the original viewport marker into separate areas. How to keep track of that, and how to clean up?

PR #8228, opened 2020-04-18
WIP/testing only: set viewport factor to 3.0 on map search

moving-bits:
alternative to #8227, only setting the viewport factors to 3.0 on calls to `searchByViewport()`

(note: in contrast to #8227 without additional debug info)

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -1395,7 +1395,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
         try {
             showProgressHandler.sendEmptyMessage(SHOW_PROGRESS); // show progress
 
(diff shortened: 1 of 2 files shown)


Lineflyer:
Short feedback for today only (need to carefully double-check whether I did something wrong):
Without running the testsuite in details, I have to say, that I do see exactly no difference compared to 2020.03.30. Still a lot of requests for short movements.

Note:
I had to use OSM instead of Google as the debug version has no GMaps key, but IMHO OSM behaved like GMaps in 2020.03.30.

moving-bits:
> I have to say, that I do see exactly no difference compared to 2020.03.30. Still a lot of requests for short movements.

Same here.

But that's probably to be expected as we've only changed the viewport size to be searched within, but no further checks. So it's even worse than currently, as we do not only do the same amount of requests but we do them for a larger area with possibly more results...

moving-bits:
not showing the results wished for, therefore closing this tech study

PR #8232, opened 2020-04-21
Add option to clear the goto history (fix #8224)

### LINKED ISSUE #8224: Provide a way to delete all waypoints on the "Goto" page.
I'd like to be able to remove all (or a large subset) of the waypoints on the "Goto" page.

An option to "Delete all" waypoints, or maybe delete some of them using the "Select mode" feature that appears in other lists in cgeo.

moving-bits:
- Deletes all but the five most recent goto history entries
- Menu option is only shown for the "goto history" UDC (ZZ0)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1005,6 +1005,8 @@
     <string name="init_longtap_udc">Long tap on map</string>
     <string name="init_summary_longtap_udc">Long tap on map to create a user defined cache</string>
     <string name="goto_targets_list">Internal caches</string>
+    <string name="clear_goto_history_title">Clear goto history</string>
+    <string name="clear_goto_history">This will delete all historical waypoints except the five most recent</string>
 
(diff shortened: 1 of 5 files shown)


main/res/values/strings.xml
```diff
@@ -1005,6 +1005,8 @@
     <string name="init_longtap_udc">Long tap on map</string>
     <string name="init_summary_longtap_udc">Long tap on map to create a user defined cache</string>
     <string name="goto_targets_list">Internal caches</string>
+    <string name="clear_goto_history_title">Clear goto history</string>
+    <string name="clear_goto_history">This will delete all history waypoints but the last five</string>
```
bekuno, line 1009: I suggest the wording `This will delete all historical waypoints except the last five`

moving-bits: Thanks for the hint. I changed it to "This will delete all historical waypoints except the five most recent" - should be even clearer.

Issue #8120, opened 2020-02-02
Upgrade minSDK to 21 until end of 2020


moving-bits:
(Edit: We support down to minSDK 16, but the message is basically the same.)

Currently we support down to minSDK 19 (which is Android 4.4). This blocks at least OK.HTTP from upgrading to the v4 branch. v3 branch is deprecated and will get crucial security fixes only, and only until end of 2020 (see https://square.github.io/okhttp/#requirements). (There are probably other dependencies blocked by that, haven't checked, though.)

As of May 2019 the market share of Android 4.4 was down to 6.9%, it should be less now, and even less until end of 2020.
According to the stats Lineflyer sent around in November 2019, it's even less among the c:geo users, a mere 2.3% using Android 4.4 with c:geo.

So I propose to deprecate Android 4.4 for c:geo during the next months, and revoke Android 4.4 support until fall of 2020.

bekuno:
If necessary, we can define a version as "legacy" and only fix website changes on there.

bekuno:
see https://github.com/cgeo/cgeo/issues/8135#issuecomment-589998434

Lineflyer:
AFAIR our minSDK is 16, not 19.

moving-bits:
> AFAIR our minSDK is 16, not 19.

You're right - we currently support minSDK 16, which is Android 4.1

Android versions up to 4.0 have a share of 0.6%, according to Googles distribution dashboard as of May 19. Android versions up to 4.4.x had a share of 10.7% at that time, should be less now.

jonas-koeritz:
Going above 18 would enable us to use the KeyStorw to securely encrypt user credentials in storage as well.

Lineflyer:
Some statistics taken from Google Play as of today:


Android Version | Active installs
-- | --
Android   4.1 | 2637
Android 4.2 | 3076
Android 4.3 | 1585
Android 4.4 | 15287

Total active installations: 800k

So the percentage of Android 4.x is about 2,8%.
BTW: Active installs does not provide any information about usage, just that c:geo is confirmed to be installed on those devices within the last 7 (or 14) days.

mirabilos:
How about asking the OKHTTP people to support 4.4 then?

jonas-koeritz:
@mirabilos I think they couldn't care less as we're not using the latest version of the library. I don't know of their release cycles and stuff but I assume backporting fixes for ancient Android versions is a lot to ask for.

[...]

PR #8238, opened 2020-04-22
adjust styling for API 21+ to fix progressdialog crash (fix #7533)

### LINKED ISSUE #7533: Move data dir to SD crashes the app
- Go to Settings - System
- Select to move the geocache data to the external SD

- Crash

This was reported for the market version 2019.04.16 on Android 4.4.2 but seems also to happen on my Android 9 device

Fix for this should target releas branch.

Not sue if this is an unfixed Targetsdk regression or something else.
Need to take a deeper look next days, unless someone is faster.

moving-bits:
Leave out two attributes for `cgeoProgressdialogTheme` and `cgeoProgressdialogTheme_light` stylings and re-add them only for API level 1 to 20.

Needs some extra testing on different physical devices, see https://github.com/cgeo/cgeo/issues/7533#issuecomment-618023674

### DIFF
--- a/main/res/values-v1/styles.xml
+++ b/main/res/values-v1/styles.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!-- progress dialog theme -->
+    <style name="cgeoProgressdialogTheme" parent="cgeoProgressdialogTheme_Base">
+        <item name="android:textColorPrimary">?text_color</item>
+        <item name="buttonBarButtonStyle">@style/button_borderless</item>
+    </style>
+    <style name="cgeoProgressdialogTheme_light" parent="cgeoProgressdialogTheme_light_Base">
+        <item name="android:textColorPrimary">?text_color</item>
+        <item name="buttonBarButtonStyle">@style/button_borderless</item>
+    </style>
+
+</resources>
--- a/main/res/values-v21/styles.xml
+++ b/main/res/values-v21/styles.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!-- progress dialog theme -->
+    <style name="cgeoProgressdialogTheme" parent="cgeoProgressdialogTheme_Base">
+    </style>
+    <style name="cgeoProgressdialogTheme_light" parent="cgeoProgressdialogTheme_light_Base">
+    </style>
+
+</resources>
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -377,17 +377,13 @@
     </style>
 
     <!-- progress dialog theme -->
-    <style name="cgeoProgressdialogTheme" parent="Theme.AppCompat.Dialog.Alert">
-        <item name="android:textColorPrimary">?text_color</item>
+    <style name="cgeoProgressdialogTheme_Base" parent="Theme.AppCompat.Dialog.Alert">
         <item name="android:textColor">?text_color</item>
         <item name="android:windowBackground">@drawable/border_straight</item>
-        <item name="buttonBarButtonStyle">@style/button_borderless</item>
     </style>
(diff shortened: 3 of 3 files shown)


bekuno:
I think that this is in this way too detailed devided in 3 files.
If we switch to minSdk 21 then we stay with 2 files, but this is not necessary then.
Therefore I suggest to not define the extra "cgeoProgressdialogTheme_Base".
Nevertheless, it's fine that you could test the behavior (what I can not).

bekuno:
retest this please

moving-bits:
> I think that this is in this way too detailed devided in 3 files.

I could have done it in 1, 2 or 3 files, depending on the mileage we want to go:
- If we skip support for API level up to 20, we could simply remove the four offending lines in the original `styles.xml`. Otherwise we will have to distinguish between API level 1-20 and 21+.
- To distinguish between the two versions we could copy the original `styles.xml` to a folder `values-v21` and remove the four lines. This would lead Android to use this version of `styles.xml` for all Android versions with API level 21+. That would require only two files, but the new one being a nearly complete duplicate, thus having to maintain two more or less complete sets of styles.
- The option I chose was to leave as much of the original styles in a common file used by all Android versions - a modified version of today's `styles.xml`, the only change there being the `progressdialogtheme` styles reduced to those attributes that are shared over all Android versions. Also I needed to rename those two styles to be able to derive from them in the next two files without needing to change program code.
Then I added the `values-v1` version with new `progressdialogtheme` styles deriving from the modified base version, adding those attributes only relevant for the older Android versions. This will be taken by Android for all versions starting with API level 1, meaning: all versions.
And to prevent that this version of `styles.xml` gets used in newer Android versions as well, I needed to add a "dummy" `styles.xml` in `values-v21` folder, although that one does nothing more than to supply `progressdialogtheme` styles that derive from the definition in the common `style.xml` file without any change.

(All this is under the assumption that it starts with API level 21, which I'm not yet sure about. If it starts with an earlier version, we simply need to move the `values-v21` version of `styles.xml` to the according folder, e. g. `values-v19`.)

> If we switch to minSdk 21 then we stay with 2 files, but this is not necessary then.

Yes, and I would like to do that the earlier the better, see #8120. How to proceed with that?

moving-bits:
retest this please

bekuno:
> > If we switch to minSdk 21 then we stay with 2 files, but this is not necessary then.
> 
> Yes, and I would like to do that the earlier the better, see #8120. How to proceed with that?

I do also not love a duplication of code. 
My suggestion was __only__ because of our plan in #8120.
Without that I had written "perfect solution". 

> (All this is under the assumption that it starts with API level 21, which I'm not yet sure about. If it starts with an earlier version, we simply need to move the values-v21 version of styles.xml to the according folder, e. g. values-v19.)

I think that v21 is fine.
See the link in https://github.com/cgeo/cgeo/issues/7533#issuecomment-617425109
The corresponding PR https://github.com/heinrichreimer/android-issue-reporter/pull/101
fixes it in the 2-files-method with v21.

moving-bits:
> I think that v21 is fine.

I'm not sure, because #7533 states "Android 4.4.2" in the opening message, which would be API 20 IIRC. I saw this too late, so stayed with API 21 for the time being. Also I cannot test this, as my only older physical testing device is running on Android 5 (API 21).

Lineflyer:
I did not follow your discussion closely, just saying:
The issue is also present on my Samsung S10 running Android 10.

moving-bits:
> I did not follow your discussion closely, just saying:
> The issue is also present on my Samsung S10 running Android 10.

Yes, seems to happen from Android 5 onward, up to current versions. Maybe already starting at some earlier point (API 20 instead of 21?), and maybe not on all devices / not for all manufacturers. - Tests will tell us.

Lineflyer:
Is the failing test `cgeo.geocaching.activity.ProgressTest.testProgressWrapper` related to the change?

moving-bits:
> Is the failing test `cgeo.geocaching.activity.ProgressTest.testProgressWrapper` related to the change?

will look into it tonight

moving-bits:
I started the tests locally again - `testProgressWrapper` does not fail.

Looking at the source, this test only checks if the dialog window itself comes up (which it does).

So should be fine here. Will restart the build.

moving-bits:
retest this please

PR #8268, opened 2020-04-27
allow triggering WhereYouGo from cachedetail's main page (fix #8256)

### LINKED ISSUE #8256: Make triggering WhereYouGo from c:geo more prominent
At the moment the link to the app "WhereYouGo" is implemented as an entry into the list of navigation tools, if the cache description contains a cartridge link.

The placement of this function inside the menus is not so ideal. While it technically is a navigation aid most users might not expect it there or just miss to see it in the whole bunch of options

Therefore I propose to either place the menu entry into the cache details menu itself or (even better) find a good placement directly into the cache detail tab. 
On of the following could be used:
- Make the cache name clickable like a hyperlink triggering WhereYouGo download
- Add a section (e.g. above or below the watchlist section) with a small WhereYouGo logo and the option "Click here to download the cartridge in WhereYouGo". If WhereYouGo is not installed this should be catched and offered to install it together with a hint, that users need to add their credentials before using the download function

moving-bits:
For cache listings with a detected URL to a WhereIGo cartridge an extra line gets added on the cachedetail's main screen with a button to start the cartridge:

image

When the button gets tapped, c:geo checks for availability of the WhereYouGo player. If it's not present, a short info notifies the user and offers installing the player:

image

Otherwise the user gets informed that WhereYouGo will be called with the cartridge:

image


Lineflyer:
Thanks @moving-bits also especially for including those screenshots which are really helpful to evaluate the proposed change and for later testing.

This is exactly what I had in mind when opening the correpsonding issues. 

If WhereYouGo is not installed, we add a note about the need to store user credentials after installation. I tend to say no, but rather solve this problem at the WhereYouGo side in parallel (if possible).

moving-bits:
> If WhereYouGo is not installed, shall we add a note about the need to store user credentials after installation. I tend to say no, but rather solve this problem at the WhereYouGo side in parallel (if possible).

This is handled inside WhereYouGo. After hitting the "OK" button on the last screen, WhereYouGo opens:

image

If you haven't set any credentials, WhereYouGo will ask for them:

image

You can enter username / password from this page:

image

and return to the download screen. Downloading works, then, as does starting the cartridge.

Lomanic:
This is a real life improvement, I think I (almost) never cached WIG caches because it was not obvious to use, this makes it very straightforward!

If I may add, is it really necessary to have these alerts explaining what the app does next? 

As an alternative, the "Start this cartridge in WYG" text could be "(Tap to)? Install companion app WYG to launch this cartridge" if WYG missing, and would redirect to Play Store without confirmation. Users can simply go back to cgeo by tapping the back button, as usual when switching between applications.

Lineflyer:
I would keep the dialog before redirecting to Playstore, but skip the dialog if its already installed.

If the text on the cache detail page can be changed depending on WhereYouGo being installed or not as suggested by @Lomanic I agree, that we could drop the dialoges

moving-bits:
Good ideas. I adapted the PR:

image

image

PR #8271, opened 2020-04-28
add filtering for empty locations to country filter (fix #8258)

### LINKED ISSUE #8258: Country filter list contains empty element
**Describe the bug:**
While trying the new country filter I do see one empty line within the list items.
I guess this might be due to having OC.de caches on my lists, where the location field is not used.

**To Reproduce:**
1. Save some gc.com caches, save some oc.de caches
2. Open the list
3. Select Filter, Country

The list contains the countries matching the caches on my list, but also one empty line
When activating this empty line as a filter, it seems to return all caches

Empty lines should be hidden as they do not deliver a valid filter result

**Version of c:geo used:**
You will find the c:geo version in c:geo Menu -> About c:geo

Yes / No

**Screenshots:**

**System information:**
```
Attach system information here if available (see c:geo Menu -> About c:geo -> Swipe right to System)
```

**Additional context:**

moving-bits:
Selecting the empty country filter line now means filtering for an empty/missing location info.

### DIFF
--- a/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
+++ b/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
@@ -52,7 +52,11 @@ public class CountryFilterFactory implements IFilterFactory {
         public boolean accepts(@NonNull final Geocache cache) {
             final @NonNull String location = cache.getLocation();
 
-            return location != null && location.endsWith(this.location);
+            if (this.location.isEmpty()) {
+                return null == location || location.isEmpty();
+            } else {
+                return null != location && location.endsWith(this.location);
+            }
         }
 
     }


moving-bits:
> if the location of any cache in the db is `null`

I've cross checked this with a manipulated database: Caches with location==null will be treated in the same way as caches with location=="" (at least regarding the country filter).

jakedot:
Just a quick question: Is the filter displayed just an empty line then? This can be corrected to something like '<empty>' e.g.

moving-bits:
> Just a quick question: Is the filter displayed just an empty line then? This can be corrected to something like `<empty>` e.g.

Yes, has been so. But I've just found a way to get a translatable string into filter factory, see #8277.

Issue #8097, opened 2020-01-23
Map rotation unreliable


Lineflyer:
Beta feedback from support channels:

```
When using google.maps the map starts to spin in seemingly random directions.
This doesn't happen if using Google satellite or osm maps.
```

```
On Live map, or viewing stored caches on Google map or satelite all the time is rotating. If i set back to true North, it still changes map direction in random ways. Its constantly changing and rotating on itself. 

Phone - Huawei p20 pro mate. Android 9.1
```

```
The map is rotating randomly and/or continuously.
Sometimes after moving my phone a bit, the rotation stops but while walking it starts rotating back and forth from time to time.
I changed to Google Maps app and let it rotate...its very accurate, no random rotations.
Tried to calibrate my compass (unless it showed good calibration), no change.
```

```
I don't like the rotating map. It often does not work reliable and trying switching it off by pressing the compass rose fails. It aligns north but starts to rotate again. Please make it optional or improve control
```

Conclusion:
- Either on some phones its not reliable, or the algorithms (damping, etc.) are not good enough on some phones
- We should rethink if we switch it on by default
- We should improve the UI to toggle the rotation (as mentioned by myself in another issue where I compared with the Google Maps app)

jakedot:
It's already possible to switch it off in the map settings, but it seems
that clicking the compass rose has no effect. I was confused by this
earlier today too.

Am Do., 23. Jan. 2020 um 22:49 Uhr schrieb Lars <notifications@github.com>:

moving-bits:
Hm, that'll be hard to track down for me - I have only one physical mobile phone, and on that hardware it works fine (as well in the emulator). => We need to investigate further.

For now, I suggest to make map rotation non-default and declare it as "experimental" in the changelog, but not block the release for it. See #8098.

Lineflyer:
Coming back to this:
As can be seen on support channel and in app ratings, still people have the problem that cgeo gets unusable with rotation.

Most of them do not find how to switch it off.
Most try to use the compass rose on the map to turn it off as they are used to from GMaps app, but thats not implemented.

Is it possible to implement it like in GMaps app?

jakedot:
is it too late now to - in the meantime, until it can be fixed, or better yet permanently - make this feature the non-default behavior like @moving-bits suggested?

moving-bits:
> make this feature the non-default behavior like @moving-bits suggested?

[...]

PR #8274, opened 2020-04-29
unify map rotation settings (related to #8097)

moving-bits:
Unify map rotation settings:
- There is a single setting & preference, which can be either "off", "manual" or "auto"
- Setting can be changed via Settings -> Map -> Map rotation
- or via the new menu entry "Map rotation"

image

image

Today's separate settings for "disable manual map rotation" and "disable auto map rotation" are NOT migrated. (Should be noted in release notes.)

Default value is "manual rotation".

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -93,15 +93,27 @@
                 android:title="@string/map_dot_mode"
                 app:showAsAction="ifRoom|withText">
             </item>
-            <item
-                android:id="@+id/menu_map_autorotate"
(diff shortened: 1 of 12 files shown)


jakedot:
> image

Nice! But is this screenshot up-to-date? It should be a radio button group, not a checkbox group I guess?

moving-bits:
> Nice! But is this screenshot up-to-date? It should be a radio button group, not a checkbox group I guess?

You're right. I've updated PR and screenshot:

image

PR #8276, opened 2020-04-30
skip validity check on pressing lat/lon buttons (fix #8254)

### LINKED ISSUE #8254: Misleading toast when entering coords
For version 2020.04.25:

- Open waypoint tab of a cache
- Select "Create Waypoint"
- In the coordinate dialog, tap on the hemisphere indicator N or E to change it to S or W

Result:
An error toast ist shown, that c:geo cannot parse the coords (or similar)

Expected:
I think this results from the fact, the previously the dialog was prefilled with the current location
There should no longer be a check on validity when tapping N/E but only when using OK to close the dialog

moving-bits:
skip validity check on pressing lat/lon buttons in wp editor


moving-bits:
retest this please

PR #8278, opened 2020-05-01
sort countries in country filter alphabetically

moving-bits:


jakedot:
Huh? Didn't I use a TreeSet with a sorting function in the `DataStore` select method? Which case wasn't sorted correctly? It should be possible to just put some comparator other than String::compareTo or what I put in there...

jakedot:
oh sorry. this can all be solved by just using a LinkedHashMap (with insertion order from the iteration over the already sorted TreeSet from the DataStore) or a TreeMap instead of sorting into a new ArrayList.

main/src/cgeo/geocaching/filter/CountryFilterFactory.java
```diff
@@ -67,23 +68,31 @@ public boolean accepts(@NonNull final Geocache cache) {
     @NonNull
     public List<IFilter> getFilters() {
 
-        final Map<String, IFilter> filters = new HashMap<>();
-
+        // which countries do we have?
+        final Map<String, String> tempUnsorted = new HashMap<>();
```
jakedot, line 72: just use a LinkedHashSet, then we don't need to sort again

moving-bits:
> oh sorry. this can all be solved by just using a LinkedHashMap (with insertion order from the iteration over the already sorted TreeSet from the DataStore) or a TreeMap instead of sorting into a new ArrayList.

Thanks, I was looking for something like `LinkedHashMap` to avoid the copying around!

I played around with it a bit, but stumbled upon a problem lying more deeply burrowed: The `TreeSet` from the datastore isn't sorted according to countries. It's only sorted to regions. A short example would be:

```
Berlin, Germany
Brandenburg, Germany
California, United States
Denmark
Drenthe, Netherlands
```

which would result in a country filter of `Germany, United States, Denmark, Netherlands`.

So we either need to get a result sorted by countries from the database (=> I will look into SQLite string functions for that) or we'll have to stay with a solution sorting the extracted countries within the filter.

main/src/cgeo/geocaching/filter/CountryFilterFactory.java
moving-bits: LinkedHashMap is a good hint, thanks for that, but does not solve the problem completely, see comment below.

jakedot:
> It's only sorted to regions.

Oh. I didn't think of that. Well, nothing easier than that, I'll just move the country extraction to the DataStore class?

Or else we can just use another TreeMap instead of the LinkedHashMap. See my PR #8279

jakedot:
> So we either need to get a result sorted by countries from the database (=> I will look into SQLite string functions for that) or we'll have to stay with a solution sorting the extracted countries within the filter.

I think the best and cleanest options is to introduce a function in DataStore that extracts the country, whether it is done by substringing the result of the locations query or in the DB probably isn't all that relevant. So the interface is one getter for location, and one for country.

jakedot:
Could you include the change here into your commit? https://github.com/cgeo/cgeo/pull/8279/commits/4d962f9407bbac5cbfcbd3af1e565d5f346db2a4#diff-ad4d074f482b8b269bb8d4fb558835eb

Or remove the unused variants for this time...

moving-bits:
As I have feared:
- The new solution (not yet pushed upstream) does not work with Android 4.x (due to the supported SQLite version there, which does not understand the string function INSTR).
- Android 5.x is fine, though.

So I propose to either:
- use the current solution of this PR (https://github.com/cgeo/cgeo/pull/8278/commits/67f637bbb2cd03f64b2182a9d4fcdd31e4f63c97) for now and push the more compact solution to master after cutting off the legacy branch
- or wait with this PR for cutting off the legacy branch and push the updated solution then.

moving-bits:
> Could you include the DataStore change here into your commit? [4d962f9#diff-ad4d074f482b8b269bb8d4fb558835eb](https://github.com/cgeo/cgeo/commit/4d962f9407bbac5cbfcbd3af1e565d5f346db2a4#diff-ad4d074f482b8b269bb8d4fb558835eb)
> 
> Or remove the unused variants for this time...

I propose to remove the unused variants (consts and functions) for now and reintroduce them later again, when needed (which would be if we extend the filter system to work on current sets in memory).

jakedot:
#8284 might be a cleaner approach for now.

moving-bits:
ok, let's close this PR in favor of #8284

PR #8279, opened 2020-05-01
Preserve sort order from DB in country filter list

jakedot:
alternative to #8278


moving-bits:
This does not work - see the results from my test data set:

image

But never mind, I've got a working solution now (by tweaking the SQL) - just need to test it with an older version of Android to make sure it's not dependent on some newer SQLite version.

PR #8280, opened 2020-05-01
sort countries in country filter (API 21+ version)

moving-bits:
More elegant version of #8278, but working from API level 21 onward only, so a WIP for now.

Thanks to @jakedot for pointing me to the `LinkedHashMap`!

image


moving-bits:
After merging of #8284 this is no longer needed, therefore closing it.

PR #8294, opened 2020-05-03
remove unused string

moving-bits:


UniQP:
`user_defined_caches` is also unused. What about unused translations?

moving-bits:
> `user_defined_caches` is also unused.

I was coming from the last CI build's lint warning, which had marked `map_autorotate` as no longer in use.

I'd also started "check for unused resources" in AS, but that lead to nothing, even did not point to `map_autorotate`. This had worked in the past, IIRC.

> What about unused translations?

Should be removed automatically with the next PR coming from Crowdin.

UniQP:
For me, the check for unused resources reports both:
image
Since it reports more unused resources for the translated strings, I'm not convinced that they will get removed automatically.

moving-bits:
Strange. What am I doing wrong here?

image

image

image

UniQP:
For me, these steps just work. I'm also using AS 3.6.3.

bekuno:
Short path in AS:
Menu: Refactor
Remove Unused Ressources ...
Preview
grafik

moving-bits:
Thanks for the hints. I had tried both ways, but still no luck - something seems to be wrong in my AS config :-(

Anyway, I've doubled the impact of this PR and included both strings.

UniQP:
Let's see whether the translations are removed automatically.

moving-bits:
> Let's see whether the translations are removed automatically.

They are removed during build process anyway ("warn: removing resource cgeo.geocaching:string/map_autorotate without required default value."), so there's no problem for the generated APKs.

As for the sources coming from Crowdin I guess they will be removed for those languages where a new `strings.xml` is generated, which may be those new translations are coming in. But I'm only guessing here - let's keep an eye on it throughout the next days.

UniQP:
After merging #8296 the translations of the removed strings are gone :-)

moving-bits:
LGTM

Lineflyer:
> After merging #8296 the translations of the removed strings are gone :-)

Thats what I hoped would happen.
So we don't need to care about strings in translated languages. Keeping the source clean, will automatically sync the translations from Crowding. Thanks for proving this.

PR #8306, opened 2020-05-04
fixes for update individual route (fix #8303, fix #8304)

### LINKED ISSUE #8303: Brouter route is lost for individual route when resuming map
2020.05.04-NB

To Reproduce (example case):
- Have brouter installed and configured
- Map a cache which has some waypoints with coordinates
- Use long tap to define a route between several of the waypoints
- Close map to return to cache details
- Map the cache again

Result:
The turn-by-turn route (coming from brouter) is replaced by direct lines between the waypoints

Expected:
Same route as before should be shown

Screenshots:
After defining route
Screenshot_20200504-210453_cgeo

After returning to map
Screenshot_20200504-210509_cgeo

### LINKED ISSUE #8304: Individual route not updated on routing type change
2020.05.04-NB

To Reproduce:
- Have brouter installed and configured
- Open live map with some waypoints/caches on it
- Select Menu - Routing - Walking
- Create individual route by long tapping some waypoints/caches
- The appropriate route routed as pedestrian is shown
- Select Menu - Routing - Driving

Result:
Route does not change (still on foot paths, etc. for walking)

Expected:
Route should be redrawn using the newly selected routing type

moving-bits:
- recalculate and repaint individual route after routing type changes (fix #8304)
- delay route calculation until BRouter connection is established (fix #8303)

### DIFF
--- a/main/src/cgeo/geocaching/maps/AbstractMap.java
+++ b/main/src/cgeo/geocaching/maps/AbstractMap.java
@@ -32,7 +32,7 @@ public abstract class AbstractMap {
 
     public void onCreate(final Bundle savedInstanceState) {
         mapActivity.superOnCreate(savedInstanceState);
-        Routing.connect();
+        Routing.connect(null);
     }
 
     public void onResume() {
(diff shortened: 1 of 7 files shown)


Lineflyer:
retest this please

Lineflyer:
retest this please

PR #8326, opened 2020-05-10
GMv2: disable autorotation on tap on compass rose

moving-bits:
This is a proposal to get a grip on the auto rotation UI:
It catches a tap on the compass rose (if rotation mode is set to `AUTO` only) and ask the user if he/she wants to disable auto rotation.

This is a WIP only - @Lineflyer  Do you have a chance to test it?

Note: Currently it asks every time the user presses the compass rose (if rotation mode is set to `AUTO`) - we should sort of damp this, like only asking once or twice each session. - To be figured out yet what could work.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1156,6 +1156,8 @@
     <string name="map_routing_walk">Walk</string>
     <string name="map_routing_bike">Bicycle</string>
     <string name="map_routing_car">Car</string>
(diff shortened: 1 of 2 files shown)


Lineflyer:
I think we should give this a try for the nightlies and potentially our next beta.
@moving-bits Can you make those string translatable, so we could merge this as potential solution?

PR #8332, opened 2020-05-11
add logcat function to about-system screen (fix #8331)

### LINKED ISSUE #8331: Provide built in logcat function
Inspired by a discussion in https://github.com/cgeo/cgeo/issues/8319#issuecomment-626706647:

>> Hmm. LogCat doesn't work anymore, thus installed MatLog, granted read access to logs, as it demanded (adb shell pm grant com.pluscubed.matlog android.permission.READ_LOGS), but I don't see anything from c:geo - anything else left to be done? Or is there a better way to get the trace?

>Yes, actually using on-board tools to trace has become really difficult. I have to admit, that I did not find any solution up to now due to limited time.

---

Feature request: Provide a built-in function in c:geo to save its own logcat to a user-readable file, maybe even make it sendable via e-mail.

moving-bits:
To make acquiring logcats easier I started implementing a little logcat functionality to c:geo. I tested in different emulated Android versions, and it seems to work starting from API level 24, but not up to 21 - although I'm not explicitly using any function with a higher API level. Android documentation on logcat itself is a bit sparse, though, so it may be due to a parameter not supported under those Android versions. - Therefore setting "WIP" for now to get some feedback first.

Logcat function is added to the c:geo - about - system screen:
(see the new "generate logfile" button at the bottom)

image

After successful write of the logfile to the new "logcat" directory a message confirms if you additionally want to send the file as e-mail:

image

When confirmed, the regular app chooser opens for you to select your preferred e-mail program:

image

### DIFF
--- a/main/res/layout/about_system_page.xml
+++ b/main/res/layout/about_system_page.xml
@@ -49,6 +49,12 @@
             android:textIsSelectable="true"
             android:textSize="12sp"
             tools:text="long list of system properties"/>
+
+        <Button android:id="@+id/logcat"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:text="@string/about_system_write_logcat"
+            android:padding="12dp" />
     </LinearLayout>
 
 </ScrollView>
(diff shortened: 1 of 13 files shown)


moving-bits:
Changes since last version:
- parameters for logcat call optimized to work from Android 4.x to 10
- directory renamed to "logfiles", now also used for memory dump
- added button "Generate logfile" to settings - system
- reorganized into appropriate utility classes

Lineflyer:
Lets test in our nightlies

PR #8355, opened 2020-05-16
show waypoints of owned cache on OSM SINGLE when hideOwn==true (fix #8333)

### LINKED ISSUE #8333: Waypoints of own caches hidden on cache map (OSM)
I noticed the following behaviour with 2020.05.12-NB which I consider to have changed only recently.
(Somehow connected to #8300 )

Steps:
0. Use OSM map
1. Enable hiding of own and found caches in c:geo settings
2. Open an owned cache which has waypoints with coordinates
3. Open map for this cache

Actual result:
Cache icon is shown, waypoints are missing

Expected result:
As also discussed in #8300 I would expect all cache information being shown in the map for a certain cache. With Google Maps this function works as expected, problem only on OSM.
I am pretty sure, that this worked on OSM few weeks ago.

moving-bits:
For OSM in single cache mode: show waypoints of owned caches even when hideOwn is set to true


moving-bits:
retest this please

PR #8364, opened 2020-05-22
refine viewport caching searchresults (fix #8260)

### LINKED ISSUE #8260: Live map caches might be missing on low zoom
This is a follow up issue for a regression caused by the fix for #8225:

While testing the interim solution we made, I ran into situations where the live map was no longer populated with caches at some areas. I do not yet have a clear sequence or description of the problem, but let me try to describe, what I think is happening:

If you zoom out the map on a region with high cache density, it will happen that we reach the maximum amount of caches returned from the server and this ony fills part of the total viewport requested (You will then typically see a circle filled with e.g. 500 caches in the center and empty space around).
If you now zoom in again staying within the area of the previously requested viewport but not within the "circle" of loaded caches, you will see a map without caches and those are not requested from the server as we consider this viewport already loaded.

Maybe we need to dig deeper, if it is that simple.
And we should really carefully analyze how to solve such kind of situation without again sending to many requests. One part of it could be to analyze the area covered by the caches loaded in the request and set this into relation with the total viewport. But this alone will not solve our problem as loading more caches will potentially cause other caches being thrown out of our internal cache cache.

I personally do not consider this a high priority problem as such, as the live map was never designed to visualize thousands of caches at once. So better stick with this regression than find a quick fix by increasing server load.

moving-bits:
Refine viewport caching results: If we retrieved a sufficient amount of caches in searchByViewport (400+) do not longer use the queried viewport as cached area but the bounding box of caches returned.
(as described in https://github.com/cgeo/cgeo/issues/8260#issuecomment-629660662)

### DIFF
--- a/main/src/cgeo/geocaching/connector/ConnectorFactory.java
+++ b/main/src/cgeo/geocaching/connector/ConnectorFactory.java
@@ -33,7 +33,6 @@ import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Trackable;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.AndroidRxUtils;
-import cgeo.geocaching.utils.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


moving-bits:
retest this please

Issue #7732, opened 2019-07-22
API to extremcaching.com not working


Lineflyer:
Reason is, that their website SSL certificate has expired (once again).

I disabled the corresponding watchdos tests already in 9d36811a57d525909624312922fa0dd94c796108 as I expect at least some days (given to the experience of last time this happend) to install a new certificate.

This issue is just to inform and as a reminder to enable those tests again if the API is functional again.

okainov:
Seems certificate is valid now, probably should be re-enabled and closed :)

Lineflyer:
Funny side note, the website is offline, the API works and delivers data.

moving-bits:
I have a PR ready for reactivating, but we should not merge it for now:

Reactivating the tests still break for EC, as their site currently returns a HTTP 503 error - "service temporarily unavailable".

If someone can give me a valid EC cache URL I can adapt the test to use this instead of their homepage.

bekuno:
Currently no login, only this text:
![grafik](https://user-images.githubusercontent.com/32555742/82705695-70c21e80-9c78-11ea-9570-e0ef3444f719.png)
(https://extremcaching.com/index.php)

Lineflyer:
The API was functional although the website landing page was blocked.
But now qdditionally their SSL certificate again expired.

bekuno:
@moving-bits 
Here are some cache URL:
http://extremcaching.com/output-2/1226-t5-baumcache-nr-172
http://extremcaching.com/output-2/1218-01-brenz-14
http://extremcaching.com/output-2/1107-the-yellow-roppe

bekuno:
> But now qdditionally their SSL certificate again expired.

It's now sign by Let's Encrypt. The certificate is valid until "24.9.2020, 19:28:26 (Mitteleuropäische Sommerzeit)"

moving-bits:
@bekuno 
Thanks for the links. The last one is accessible without login, so I chose that one as new test URL for ec adapter, see updated #8365.

moving-bits:
Site is back running. Can we close this issue?

Lineflyer:
I think so.

PR #8365, opened 2020-05-22
reactivate ec website test as last website being tested

moving-bits:
Do not merge for now, due to https://github.com/cgeo/cgeo/issues/7732#issuecomment-632822577

- reactivate ec website test
- but move it to the end of the website tests to not block other website tests

### DIFF
--- a/tests/src-android/cgeo/watchdog/WatchdogTest.java
+++ b/tests/src-android/cgeo/watchdog/WatchdogTest.java
@@ -4,6 +4,7 @@ import cgeo.CGeoTestCase;
 import cgeo.geocaching.SearchResult;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.connector.ec.ECConnector;
 import cgeo.geocaching.connector.internal.InternalConnector;
 import cgeo.geocaching.connector.oc.OCApiConnector;
 import cgeo.geocaching.connector.trackable.TrackableConnector;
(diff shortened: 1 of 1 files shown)


Lineflyer:
Using a non htttps page here does not really make sense.
In fact I would like the test to fail if their certificate is not ok, as this means, that also our users cant use the API.

main/src/cgeo/geocaching/connector/ec/ECConnector.java
```diff
@@ -86,6 +86,15 @@ public String getNameAbbreviated() {
         return "EC";
     }
 
+    @Override
+    @NonNull
+    public String getTestUrl() {
```
bekuno, line 91: The test data should be in the test code.

moving-bits:
> Using a non htttps page here does not really make sense.
> In fact I would like the test to fail if their certificate is not ok, as this means, that also our users cant use the API.

Ok, then I'd misunderstood you. I removed the last commit (non https-URL), so that the PR now only consists of reactivating the EC website test, but as the last website being tested.

main/src/cgeo/geocaching/connector/ec/ECConnector.java
moving-bits: That's the way the `IConnector` interface is implemented currently (and used by the `GCConnector` as well).

But anyway, it's obsolete now, as I I've removed the addition of an individual testUrl for the EC connector due to the change requested by @Lineflyer.

PR #8367, opened 2020-05-22
Code cleanup: Move display metrics functions to DisplayUtils

moving-bits:
We query display metrics in several places, reusing the same duplicated code fragment several times. => Therefore moved it into utility class `DisplayUtils`.


moving-bits:
retest this please

PR #8369, opened 2020-05-22
unify map lines stroke width between GMv2 and OSM maps

moving-bits:
- unify the different stroke widths in GMv2 and OSM
- use a density based width for history line and route

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -186,7 +186,7 @@ public class GooglePositionAndHistory implements PositionAndHistory, Route.Route
 
     private PolylineOptions getDirectionPolyline(final Geopoint from, final Geopoint to) {
         final PolylineOptions options = new PolylineOptions()
-                .width(5f * DisplayUtils.getDisplayDensity())
+                .width(DisplayUtils.getThinLineWidth())
                 .color(0x80EB391E)
                 .zIndex(ZINDEX_DIRECTION_LINE)
                 .add(new LatLng(from.getLatitude(), from.getLongitude()));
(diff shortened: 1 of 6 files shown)


main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
```diff
@@ -286,15 +286,15 @@ private synchronized void drawHistory() {
                 historyObjs.addPolyline(new PolylineOptions()
                     .addAll(points)
                     .color(0xFFFFFFFF)
-                    .width(3)
+                    .width(DisplayUtils.getThinLineWidth() / 2f)
```
UniQP, line 289: Shouldn't this be a separate method that uses a separate constant?

main/src/cgeo/geocaching/utils/DisplayUtils.java
```diff
@@ -9,6 +9,8 @@
 
 public class DisplayUtils {
 
+    private static final float THIN_LINE = 5f;
```
UniQP, line 12: Looking at the code changes above, I would expect the constant to be `7f`.

moving-bits: Four different widths have been used in the past:
- 5f * density
- 8f * density
- 3
- 7
the first two being for the same, but in different map types, the latter two being absolute values instead of density related.

I merged this into `getThinLineWith()` and `getThinLineWidth() / 2` (although agreeable that I could invent two different methods for this - will update the PR accordingly).

main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
moving-bits: resolved

PR #8371, opened 2020-05-23
update waypoint list after nested waypoint delete (fix #8370)

### LINKED ISSUE #8370: Waypoint list not up to date after changes in nested cache calls
**Describe the bug:**
A deletion from waypoint list will not immediately be shown in first cache call, if the waypoint has been deleted in a subsequent [more details] call ...

**To Reproduce:**
1. open any cache
2. change to waypoint list and add a waypoint
3. use compass rose in waypoint list header to display all of them on map
4. tap on cache --> [more details]
5. go to waypoint list
6. delete the previously added waypoint
7. tap BACK (will end up in map view - where BTW the deleted waypoint will still be displayed)
8. tap BACK again (will end up in waypoint list

The previously deleted waypoint *seems* to be reincarnated. But in fact it just seems to be a refresh problem, cause as soon as you tap [add waypoint] and after that the [X] to cancel waypoint addition, the deleted waypoint will now be vanished.

Waypoint list should be updated in all "levels", esp. when a deletion has been done in a nested cache call.

**Version of c:geo used:**
Latest nightly

Yes

**Additional context:**
Things are a little bit different, if in the second cache call you do not delete the waypoint, but just change coordinates like:
1. add a waypoint
2. use compass rose in waypoint list header to display all of them on map
3. tap on cache --> [more details]
4. go to waypoint list
5. change coordinates of the previously added waypoint
6. tap BACK (will end up in map view - where BTW *both* versions of this waypoint will be displayed)
7. tap BACK again (will end up in waypoint list

In this case the waypoint coordinates are the ones you entered in step 5.

Would it be possible that this issue may have any connection to #5854 ...?
Esp. where PNs must be changed twice, sometimes, to make the change permanent?

moving-bits:


moving-bits:
retest this please

PR #8372, opened 2020-05-23
Fix link in Javadoc comment

UniQP:


moving-bits:
Not needed / will be superseded by #8364.

PR #8381, opened 2020-05-25
add mapcache bounding box for GMv2 when in debug mode

moving-bits:
For testing #8260/#8364:
- activate debug mode in c:geo settings
- go to Google Maps live map
- zoom out until max. number of caches got loaded => see red box around it
- red box will move when you start panning the map and caches get loaded additionally

see https://github.com/cgeo/cgeo/issues/8260#issuecomment-633713331 for examples


Lineflyer:
will try to test with the debug APK of this PR.

Lineflyer:
@moving-bits Or shall we just include this code for future debug possibilities? Its not that much and only active in debug mode. So I would be fine, to just include it on `master`.

moving-bits:
> @moving-bits Or shall we just include this code for future debug possibilities? Its not that much and only active in debug mode. So I would be fine, to just include it on `master`.

yes, that should be ok - I'll merge it

PR #8384, opened 2020-05-27
restrict viewport calculation to GC live caches

moving-bits:
See https://github.com/cgeo/cgeo/issues/8260#issuecomment-634888573


Lineflyer:
The failure can be ignored, geolutins.com is down.
Merging for further test.

PR #8391, opened 2020-05-30
adapt mockfile for GC2JVEH to changed listing

moving-bits:
Description in original listing now has an integrated closing tag in the ``-Tag


moving-bits:
errors in testSearchByGeocodeBasis and testParseCacheFromTextWithMockedData due to the changed cache listing are fixed with this, so let's merge

PR #8419, opened 2020-06-07
adapt WIG URL checker to catch more conditions (fix #8414)

### LINKED ISSUE #8414: Wherigo link not detected
For GC7WDB4 we seem to fail parsing the wherigo link thus not showing the "open with WhereYouGo"

moving-bits:
Adapt URL regexp to match the exact parameter part in the URL instead of catching every character until the next closing '"' character.

- catch slightly different syntaxes as ok, if the URL itself is identical (example: https://coord.info/GC7WDB4)
- catch https URL (example: https://coord.info/GC7K2KZ)

Can be targeted to release branch, if needed.


main/src/cgeo/geocaching/apps/cache/WhereYouGoApp.java
```diff
@@ -20,7 +20,7 @@
 import org.apache.commons.lang3.StringUtils;
 
 public class WhereYouGoApp extends AbstractGeneralApp {
-    private static final Pattern PATTERN_CARTRIDGE = Pattern.compile("(" + Pattern.quote("http://www.wherigo.com/cartridge/details.aspx?") + ".*?)" + Pattern.quote("\""));
+    private static final Pattern PATTERN_CARTRIDGE = Pattern.compile("(https?\\:\\/\\/www\\.wherigo\\.com\\/cartridge\\/details\\.aspx\\?[Cc][Gg][Uu][Ii][Dd]=[-0-9a-zA-Z]*)");
```
bekuno, line 23: Are the double "\" really needed for escaping?
Please start with "a href=\"" in the pattern to find the right position. Then it is clear that the next "\">" will end the match.
Maybe you could refactor the code to move the pattern to the other in `cgeo.geocaching.connector.gc.GCConstants`

bekuno: > Are the double "\" really needed for escaping?

Find by myself that this is needed. With `Pattern.quote` is the code better readable.

bekuno: The tests in WhereYouGoAppTest should also be updated.

moving-bits: > > Are the double "" really needed for escaping?
> 
> Find by myself that this is needed. With `Pattern.quote` is the code better readable.

I had tried it with `Pattern.quote` before, as the original pattern also used this, but the mix of unquoted and quoted regexp did not work in this case, so I switched to direct quoting in this case.

moving-bits: > Please start with "a href=\"" in the pattern to find the right position. Then it is clear that the next "\">" will end the match.

I would rather stay without a `a href` part, as the link does not necessarily have to be enclosed in an `a` tag.

> Maybe you could refactor the code to move the pattern to the other in `cgeo.geocaching.connector.gc.GCConstants`

Is that the right place? WIG is from geocaching.com as well, that's true, but it is not related to the GC connector, so I think it should stay separate.

moving-bits: > The tests in WhereYouGoAppTest should also be updated.

added two tests to catch both new cases - https and URL as `a href` description (or anywhere else in the description)

bekuno, line 23: I think that this should be ok:
`Pattern.compile("(https?" + Pattern.quote("://www.wherigo.com/cartridge/details.aspx?") + "[Cc][Gg][Uu][Ii][Dd]=[-0-9a-zA-Z]*)");`

bekuno: > I would rather stay without a `a href` part, as the link does not necessarily have to be enclosed in an `a` tag.

You are right, we are parsing "freetext".

bekuno: > Is that the right place? WIG is from geocaching.com as well, that's true, but it is not related to the GC connector, so I think it should stay separate.

Mmm, yes. If we make a generalized pattern file, then the place should also be central. Let´s as is.

moving-bits: Ok, slightly different than the one I had tried. But this one actually works ;-)
Thanks for pointing me to this, will adapt the PR.

moving-bits:
> Please add also a https test.

It's already included, see `testGetWherIGoURLHttps()`

bekuno:
> It's already included, see `testGetWherIGoURLHttps()`

Ohrjaa - ischwerbleede!!

bekuno:
I would also say - let it target to release.

PR #8434, opened 2020-06-12
remove styles and themes workarounds for API < 21

moving-bits:
Remove a few alternative paths in styles.xml and themes.xml for API level < 21


moving-bits:
Not to be merged before updating min API to 21, therefore setting WIP label

PR #8439, opened 2020-06-13
add summary for show additional lists preference (fix #8437)

### LINKED ISSUE #8437: Change settings layout for init_listdisplay
Taken from portuguese translators issue on Crowdin (and also valid for e.g. german):

Please change this setting layout in the app, adding description label below the option title.
Indeed we should maybe use a short caption and a longer descriptive text here as we do for other setting items.

moving-bits:
image


Lineflyer:
Wow, that was fast :)

PR #8441, opened 2020-06-13
Change closeout warning to an incompatibility note

Lineflyer:
Theoretically people should not be able to install c:geo on a device below Android 5.0 any longer. In case this however happens for some reason, let us just reuse the closeout warning as a note about incompatibility until we have the next closeout phase.


Lineflyer:
> 
> 
> Generally ok for me, but I would leave the "Tap on this message..." part in, as the text is tapable and leads to our website.

Changed

PR #8474, opened 2020-06-19
open links from listing text in separate task (fix #8472)

### LINKED ISSUE #8472: Open links from cache listing in separate task
Taken from https://github.com/cgeo/cgeo/issues/8458#issuecomment-646081436 by @MagpieFourtyTwo:

> BTW: When started from cache detail page, new in release 2020-06-14, it's started as a separate task

Which is BTW good stuff - although when starting the cartridge from c:geo's listing page, WYG will still launch in c:geo's context - and then #8368 strikes again. Don't really get why this behaves different compared to http links: clicking these guys in a listing invokes Firefox correctly (in an own task) ...?

moving-bits:
by setting `Intent.FLAG_ACTIVITY_NEW_TASK` on URL intents

### DIFF
--- a/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
+++ b/main/src/cgeo/geocaching/ui/AnchorAwareLinkMovementMethod.java
@@ -1,11 +1,18 @@
 package cgeo.geocaching.ui;
 
+import android.content.ActivityNotFoundException;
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

Issue #7404, opened 2019-03-21
Wrong field notes encoding


SammysHP:
##### Detailed steps causing the problem:
- Export logs as field notes

##### Actual behavior after performing these steps:
File is encoded as UTF-16 Big Endian.

##### Expected behavior after performing these steps:
File should be encoded as UTF-16 **Little** Endian.

##### Version of c:geo used:
unknown

##### Other comments and remarks:
Report from user with sample field notes file.

Lineflyer:
> File should be encoded as UTF-16 Little Endian.

Just out of curiosity....why should it be little endian?

SammysHP:
Not easy to answer as there is not much documentation available.

https://forums.geocaching.com/GC/index.php?/topic/241595-field-notes-file-format/

We should also check what older Garmin GPS receivers use.

Lineflyer:
From FB:

Just now getting around to posting an issue I ran into back in February involving the Export Field Notes function.

Without going into all the gory details (which can be found in the thread I posted on the GSAK FB page on Feb 10th) it would appear that on about February 10th, c:geo began encoding these exports as Unicode big endian.  

This is probably transparent to most users, but in my case it rendered the files unusable as imports to GSAK, owing to all the 00 bytes (clearly visible when using a hex editor to examine the files). 

At first I suspected it might be the result of an Android update (I have a Galaxy S9) but I was able to recreate the problem using my old Galaxy S5 as well. Finally, I ruled out the Android as the problem when I was able to export files from other apps on the phone and they were not encoded as Unicode big endian. 

If this behavior is indeed the result of a change within c:geo, I hope it can be reversed. The only reason I can think of that it is now this way is to allow emoji or foreign language character support, but I do understand there could be other reasons.

Bananeweizen:
Since I have no debugger at hand right now, I only checked the related files. There are no changes I can correlate. FieldNotes, FieldNoteExport, FileUtils are all unchanged since a long time. commons-io as library doing the file writing also unchanged since 2017. Seems someone needs to really debug into this...

Lineflyer:
Could be related (from support mail today):

With the update to pie, c:geo is no longer exporting a valid Field Notes txt file.  Everything looks like it did yesterday, but it is not recognized by GSAK as a valid txt file.  I have uploaded the Field Notes to GSAK for the last 4479 days in a row and today with the new Android update, it doesn't work.

jakedot:
[...]

PR #8489, opened 2020-06-20
Use standard charsets

UniQP:


bekuno:
checkstyle is unhappy:
https://ci.cgeo.org/job/cgeo%20pull%20request/2989/checkstyleResult/

Lineflyer:
Maybe this implicitly solves the strange problem we have with our UTF-8 Big endian vs. little endian problem.
I was unaware that we used a non-recommended charset tool until now.

Maybe linked to #7404

Issue #8480, opened 2020-06-20
GPX track: Crash on loading invalid GPX track file


moving-bits:
Taken from https://github.com/cgeo/cgeo/issues/2469#issuecomment-646968107:

after choosing an incorrectly formatted file (like the one exported with the previous track export version, with two lat but no lon tag) c:geo first stutters, then hangs, and finally crashes

moving-bits:
@MagpieFourtyTwo:
From a first debug view: `GPXTrackImporter` handles missing `lon` attribute correctly by ignoring those trackpoints. When trying to load one of the misformatted track files produced by the first trail exporter, it simply returns null  (as signal "track could not be loaded").

What you may have encountered is the "infinite loop problem" described in https://github.com/cgeo/cgeo/issues/8481#issuecomment-647008318 and (hopefully) fixed with #8485.

MagpieFourtyTwo:
Yep, seems so - with the latest nightly it works, c:geo does not crash anymore in both scenarios (misformatted and/or missing GPX file).

Would be perfect if a toast can be displayed, cause without a success or failure message you sometimes may feel a bit left alone in the dark. Esp. as even a valid GPX (e. g. found by automatic scan) may result in displaying nothing at all (if it e. g. contains a cache export). So, perhaps the toast can say s/th like "Trail: Loading GPX {failed|succeeded} (x trackpoints)"

moving-bits:
@MagpieFourtyTwo 
Thanks for testing. Closing this issue then.
For the status toast see #8498.

PR #8498, opened 2020-06-21
add load track status toast

moving-bits:
Show a shot toast after call to loadTracks - either "Track loaded" or "Error loading track".
(see https://github.com/cgeo/cgeo/issues/8480#issuecomment-647101943)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1185,6 +1185,8 @@
     <string name="map_hidewp_visited">Hide visited waypoints</string>
     <string name="map_hide_track">Hide track</string>
     <string name="map_load_track">Load track</string>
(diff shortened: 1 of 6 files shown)


moving-bits:
retest this please

MagpieFourtyTwo:
Tested with the mis-formatted and "good" track GPX - this seems to work.
BTW: Why not remove a failed file from the settings immediately ...?

But: Even an import that results in no points at all, shows "Track loaded" (used a locus track export), so it would be nice to know how many trackpoints c:geo is going to display ...

And ... selecting a different directory does not work (if a .gpx has been found in default dest):
After I have selected the directory to use, the map will be shown immediately, without giving me a chance to select a gpx.
Will file an issue for that tomorrow ...

PR #8506, opened 2020-06-21
Simplify code by removing unnecessary factor

moving-bits:
see also http://svn.openstreetmap.org/applications/viewer/jmapviewer/src/org/openstreetmap/gui/jmapviewer/OsmMercator.java (which is linked to from our source):

```
public long getMaxPixels(int aZoomlevel) {
    return tileSize * (1 << aZoomlevel);
}
```

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/Tile.java
+++ b/main/src/cgeo/geocaching/connector/gc/Tile.java
@@ -42,7 +42,7 @@ public class Tile {
     static {
         for (int z = ZOOMLEVEL_MIN; z <= ZOOMLEVEL_MAX; z++) {
             NUMBER_OF_TILES[z] = 1 << z;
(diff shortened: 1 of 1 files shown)


moving-bits:
AFAIK `TILE_SIZE * 1 << z` equals `(TILE_SIZE * 1) << z`, which is not the intended result in this case.

UniQP:
> AFAIK `TILE_SIZE * 1 << z` equals `(TILE_SIZE * 1) << z`

That's correct but still `(TILE_SIZE * 1) << z` and `TILE_SIZE * (1 << z)` compute the same value. We could even simplify this to `TILE_SIZE << z`.

moving-bits:
Ah, you're right - I got distracted by the lint message, because I had some issues with shifting in combination with and-ing an or-ing while working on the `ColorpickerPreference`

PR #8521, opened 2020-06-22
align individual route preference naming to map menu

moving-bits:


Lineflyer:
Couldn't we ommit the word "line" after all?
e.g. "Individual route color" instead of "Individual route line color"

moving-bits:
> Couldn't we ommit the word "line" after all?
> e.g. "Individual route color" instead of "Individual route line color"

This would be fine for color, but IMHO does not work for width: "Individual route width"?

Lineflyer:
I am just a bit afraid, that longer strings (depending on languages) will not fit in the current menu layout.
Another thought:
The meaning might be perfectly clear for us, but how about adding an explanatory text below, so its not that important how the caption reads as long as the text explains.

moving-bits:
@Lineflyer 
Something like this (shown for history trail)?
image

Lineflyer:
Yes, more or less. I would keep the current naming for the captions and use the subtext to add much more info for easy understanding, e.g. as a draft something like "This defines the color of the history trail which shows the history of your movement on the map (if activated in the corresponding setting above)."

moving-bits:
Ok. IMHO the long explanation would be necessary for each color only, and for width we can simply refer to that.

Can you provide me with texts, reflecting what's happening on support?

Lineflyer:
Ok. Will do end of this weekend.

moving-bits:
> This defines the color of the history trail which shows the history of your movement on the map (if activated in the corresponding setting above).

While starting to implement this, two questions came across my mind:
- Shouldn't we use that explanation for the "Show trail" entry as well?
- How to name it - "Show trail", "Show history track", "Show track history", ...?

@Lineflyer  What do you think?

Lineflyer:
> @Lineflyer What do you think?

I also thought about that naming each time I try to translate it.
Honestly as I am no native english speaker I am not sure about an understandable term. IMHO "Show history track" explains what it does and consequently also uses "track" as a term which we also use for the loaded track. Its somehow consistent then.

Go ahead with some string proposals first. Will review them after merge and propose improvements in another PR if needed.

moving-bits:
@Lineflyer 
Ok, have a look at #8567 - I renamed the trail history / history trail to "history track" and the new "track" to "individual track" to make them distinguishable.

moving-bits:
@Lineflyer 
I've added proposals for detailed descriptions for each line type. Can you review?

moving-bits:
rebased to current master, ready to merge after tests completed

PR #8529, opened 2020-06-26
Adapt GC login to changed error message (fix #8528)

### LINKED ISSUE #8528: Wrong error message if login to gc.com fails due to wrong credentials
- Delete stored login information for gc.com
- Start authorizing again, enter correct username but wrong password

Result:
c:geo reports "Unknown error....consult FAQ"

Expected:
c:geo should report "Wrong user name or password"

I assume that our parsing fails due to changes on the website.
This is not such a minor issue, as many people with wrong credentials contact support not knowing that just their password is wrong.
Furthermore I can no longer distinguish between credential problems and other problems.

Tested with 2020.06.24-NB but assumingly also present on market release

moving-bits:


UniQP:
Should this be merged into release instead? @Lineflyer ?

Lineflyer:
Yep, into release branch please.

moving-bits:
see #8530 for a release-based version

Issue #8477, opened 2020-06-20
Default line appearances have changed


Lineflyer:
I appreciate the possibility to customize thw different line colors and widths, however it would be good if the default values would result in the same appearance that users as used to. Otherwise I assume a lot of users will think of it as a fault, beeing unaware of the new settings.

I just now thought the direction line is broken, as it was hard to see after all. From the former big red line its now only a very thin line.

moving-bits:
Are the lines smaller in current release already (due to setting the same settings for both Google Maps and OSM, using a thinner default value, see #8394 and #8369), or did they get smaller after introducing the customization (see #8447)?

Lineflyer:
Google maps does not have a routing line, so it cant be the the former default). I assume you took the trail line prefs as default for all of them?
Could we have the individual route and cache route just have the previous values as defaults?

MagpieFourtyTwo:
As @moving-bits explained a while ago. the current default values do result in the same width - at least for him. ;)

But over here we had to at least double the default values, to make them visible enough to be able to work with them again. First I thought, this may probably be a special Huawei problem (P30 Pro with Android 10.1), so they possibly handle screen density factor different, but after upgrading our Sony Xperia Z5 Premium (Android 7.1.1) this morning, it was the same: Width must at least be 12 to make the lines visible.

moving-bits:
@Lineflyer 
I'm still unsure about what version you're talking - current release or a nightly?
(So what is the "previous value" you're comparing to.)

For current release, the change is in #8394 and #8369. Up until that version Google Maps and OSM had very (!) different line settings (so no common previous value), and I unified them to a Google Maps like style. So compared to the old OSM settings, it's much thinner now. For Google Maps, it should not have changed significantly.

Line color & width customization is on current master branch only, and I've tried to use the same settings which were used after the change described in the last paragraph. But that's tested with the emulator only (and seems to work there). - I will recheck the source, if there are any other parameters which may influence the width.

MagpieFourtyTwo:
> Google maps does not have a routing line

Of course it has - at least in the nightly, which you are using as well, AFAIK.
Over here line width is set to 12 (opacity 128) - and it has the same width in GM and OSM.

Perhaps my current values may help to find appropriate defaults (width/color/opaqueness):
- history line:  18/dark brown/255 (mapline_trailcolor: -16764160)
- direction and individual route:  12/strong blue/128 (mapline_direction/routecolor-2144141108)
[...]

PR #8531, opened 2020-06-27
use path drawing functions on OSM map to avoid "dotted lines"

moving-bits:
Switch to drawing paths instead of single line entities to avoid artifacts as shown in https://github.com/cgeo/cgeo/issues/8477#issuecomment-648646797. (OSM only, as Google Maps already uses paths.)

image


moving-bits:
retest this please

PR #8532, opened 2020-06-27
Update gradle plugin to 4.0.1 for AS 4.0

UniQP:


moving-bits:
Does Gradle plugin 4.0 work for you?

There seems to be an issue with Android Annotations in combination with Gradle plugin 4.0: https://github.com/androidannotations/androidannotations/issues/2273

bekuno:
I did not updated AS because it has some big changes.
I would (personnally) wait for AS 4.0.1 (or AS 4.1) to have a first bugfix release.

moving-bits:
> I did not updated AS because it has some big changes.

I have upgraded, due to some of the big changes (in this case: layout inspector) ;-)

> I would (personally) wait for AS 4.0.1 (or AS 4.1) to have a first bugfix release.

Which in this case seems to be a good decision, because AS 4.0 mostly works, but is buggy in some places, some of them:
- layout inspector did not work for our settings activity (which was the first use case I've tried it with), it always stayed at the main activity. Maybe my fault, need to investigate further...
- window size restores to some fixed size, but not to full screen, even if I close it in full screen mode
- running a program does not work, only debug mode - if you try to run with Shift-F10, emulator shows a message "waiting to attach debugger"
- installing a compiled app does not work on first try in about 30% of all builds
- jumping from word to word with Ctrl + cursor jumps to the end (!) of the word (since 3.6 already) - you can get used to that, but it still irritates me frequently

(and some more minor things)

Also I noticed a significant speed degradation from AS 3.5 to AS 3.6, which does not seem to be resolved with AS 4.0

fm-sys:
gradle 4.0.1 is released now

moving-bits:
Has anyone tested the 4.01 plugin with AndroidAnnotations 4.7 already whether the problem described above still persists?

moving-bits:
> Has anyone tested the 4.01 plugin with AndroidAnnotations 4.7 already whether the problem described above still persists?

Tested it today with 4.0.1 => works
Also mentioned in the original issue for AndroidAnnotations (https://github.com/androidannotations/androidannotations/issues/2273)

Therefore updated the PR to update Gradle plugin to 4.0.1

moving-bits:
@bekuno 
Does the 4.0.1 plugin work with AS 3.6? Or is it AS 4.x only?
(Meaning: Can we upgrade now, or should we wait until AS 4.x is a bit more bug-free?)

bekuno:
This should work.
I switched to AS 4.x before some time, so I can´t check it now.
Lets update now.

PR #8545, opened 2020-06-27
use number comparators

moving-bits:


bekuno:
I have a question.
From the docu of `compare`:
~~~
Returns:
    the value 0 if x == y; a value less than 0 if x < y; and a value greater than 0 if x > y
~~~
The old code returns -1/1 if not equal.
Is in the Comparator code somewhere a check to -1/1 implemented? I did not find something in `SortActionProvider`.

moving-bits:
This is how `Integer.compare()` is implemented:

```
public static int compare(int x, int y) {
    return (x < y) ? -1 : ((x == y) ? 0 : 1);
}
```

So it's equivalent to the old code.

bekuno:
I was unsure and found at https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#compare-int-int-
~~~
  compare

    public static int compare(int x, int y)

    Compares two int values numerically. The value returned is identical to what would be returned by:
        Integer.valueOf(x).compareTo(Integer.valueOf(y))
    Parameters:
        x - the first int to compare
        y - the second int to compare
    Returns:
        the value 0 if x == y; a value less than 0 if x < y; and a value greater than 0 if x > y
    Since:
        1.7
~~~

Shouldn't this adress the reference?
What is your information source?

moving-bits:
I followed Integer.compare() to the source declaration, using Ctrl-B in AS.

bekuno:
I see. The implementation is mostly realized as -1 / 0 / 1.
I found some compare with other result:
~~~
    public static int compare(short x, short y) {
        return x - y;
    }

    public static int compare(char x, char y) {
        return x - y;
    }

    public static int compare(byte x, byte y) {
        return x - y;
    }
~~~

PR #8559, opened 2020-06-28
fix #8558 adaptive icon

### LINKED ISSUE #8558: Adaptive icon
Follow up on:
https://github.com/cgeo/WhereYouGo/pull/190#issuecomment-650730442

I tried to also come up with an adaptive icon for c:geo. The current icon is not really suitable for the new specification with an opaque background and different shapes. 

First I tried it with a solid blue background as for the WhereYouGo icon (white background was too boring):

cgeo-launcher-map-blue

Then I tried something different and removed the shape of the background map and scaled it to fill the full icon background and a bigger box as the foreground:

cgeo-launcher-map-orange

What do you think?

*The GIFs are a bit grainy. On the real device it looks clean.

pstorch:
An adaptive c:geo launcher icon. It's the second version from https://github.com/cgeo/cgeo/issues/8558#issue-646955096


Lineflyer:
retest this please

PR #8563, opened 2020-06-29
skip testFormatStoredAgo on non English device

bekuno:
The test testFormatStoredAgo only works on a device with language EN.
When testing (e.g. on a dedicated physical device) it fails on devices with other languages.
Therefore we should skip this test in this situation.


bekuno:
retest this please

PR #8567, opened 2020-07-01
rename history trail and track to history track and individual track

moving-bits:
Followup to / proposal for https://github.com/cgeo/cgeo/pull/8521#issuecomment-652264737:

- rename "history trail / trail history" to "history track"
- rename "track" to "individual track"
- add a description to "show history track" in map settings

grafik

grafik

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -660,16 +660,16 @@
     <string name="init_gpx_importdir">GPX Import Directory</string>
     <string name="init_dataDir">Select Geocache data directory</string>
     <string name="init_dataDir_note">You may choose to store the additional data for geocaches (spoilers, log images, …) on your external storage (emulated or real external SD card depending on your device).</string>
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

Lineflyer:
@moving-bits Question for this: I am a bit confused now after looking at the new strings on Crowdin. What it the meaning of individual track now (I thought it would be the route you create by long clicking different waypoints), but it seems to be the track, which can be loaded. In this case, what is the wording for the route I described before?

moving-bits:
@Lineflyer 
After this PR we have:
- History track (formerly known as "trail" / "trail history" / "history trail") - which is a collection of GPS coords where you have been
- Individual track (FKA "track") - which is an imported track GPX file (which itself could be an exported history track)
- Individual route (unchanged in naming so far) - which is a collection of waypoints and/or caches you form by long-pressing on caches/waypoints on the map

PR #8569, opened 2020-07-02
add sorting ability for individual route

moving-bits:
Add a menu entry to OSM / Google Maps map: Individual route => Sort individual route.
This opens a simple sorting view, which lists alle waypoints / caches in current order and allows resorting by pressing up/down buttons. After saving the new sort order, the individual route gets recalculated an repainted.

grafik


moving-bits:
retest this please

PR #8580, opened 2020-07-04
make opaqueness value editable in ColorPreference

moving-bits:
As with the several seekbar slider preferences you can now manually enter the opaqueness value in the `ColorPreference`.


moving-bits:
retest this please

PR #8584, opened 2020-07-05
automatically disable/enable fast scroll after grace period (fix #8406)

### LINKED ISSUE #8406: Logs View Scrollbar is buggy
You can press the button in the right bar to scroll through the logs faster.  That's a good thing.

But if you get to the border bar during normal scrolling, the above mentioned button jumps to the corresponding position on this scrollbar.

This is quite buggy, because you always lose the position you were reading.
So only accept clicks on the scrollbutton, not on the hole scrollbar.

Many thanks and best regards

moving-bits:
Mitigation for the unintuitive behavior of fast scroll described in #8406:
- disable fast scroll after one second of not scrolling
- reenable after first scroll

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -102,6 +102,7 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

moving-bits:
edit: implemented for cache list view as a first test

PR #8586, opened 2020-07-06
add compatibility for restore of early settings backup versions

moving-bits:
Settings backup created with older c:geo versions created entries with `<int name="xxx" value="xxx" />` instead of `<integer ... />`, which cannot be restored by current c:geo versions. This PR adds compatibility to those old backups.


moving-bits:
retest this please

Issue #3137, opened 2013-08-17
Alert for new caches


aabeling:
This feature would alarm me when a new cache has been hidden in some predefined area via the normal android notification mechanisms.

This is like a nearby search with sorting by hiding date, newest first.
But executed automatically by the AlarmManager in predefined intervals.
cgeo should store the last check time and only alert about caches hidden after this check time.

PR #8595, opened 2020-07-08
show fastscroll thumbs in cachelist view

moving-bits:
Adds fastscroll thumbs to cache list view (similar to #8583) to make fastscroll in cache list view more visible + more useful.

What gets shown as thumb depends on the currently selected sorting, eg.:
- sort by name: shows the first two letters
- sort by date: shows year + month
- sort by rating: shows rating
etc.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1011,6 +1011,7 @@
     <string name="cache_found_last_30_days">Found last 30 days</string>
     <string name="cache_status_archived">Archived</string>
     <string name="cache_status_disabled">Disabled</string>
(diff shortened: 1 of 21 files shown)


moving-bits:
retest this please

fm-sys:
worked on my device without crashes (Build #3137 - 08.07.2020 11:29:10)

But maybe show the unit (km/mi) when sorting by distance...

moving-bits:
@fm-sys 
Thanks for cross-checking. Added conversion and distance units to thumbs view when in distance mode.

PR #8597, opened 2020-07-08
Fix #8589: Make coordinate parsing more strict

### LINKED ISSUE #8589: Wrongly parsed coords from user note
From support mail for version 2020.07.02

Using this in your personal note results in three waypoints being automatically created:
```
Answer questions
6, 12, 16, 29, 54, 3, 36, 5, 3, 47, 45, 43
```

---

I informed the user that he can turn this feature of for this particular cache, however we should analyze if this can be optimized without any disadvantages while parsing.

UniQP:
We now forbid all special characters of the coordinates to occur between the coordinate parts.

### DIFF
--- a/tests/src/cgeo/geocaching/location/GeoPointParserTest.java
+++ b/tests/src/cgeo/geocaching/location/GeoPointParserTest.java
@@ -161,11 +161,6 @@ public class GeoPointParserTest {
     }
 
     @Test
-    public void testDoubleComma() {
-        assertThat(GeopointParser.parse("47.648883,122.348067").equals(GeopointParser.parse("47.648883 122.348067")));
-    }
-
-    @Test
     public void testGerman() {
         assertThat(GeopointParser.parse("N 47° 38.933 O 122° 20.884")).isEqualTo(GeopointParser.parse("N 47° 38.933 E 122° 20.884"));
     }
@@ -313,4 +308,13 @@ public class GeoPointParserTest {
         }
     }
 
+    @Test
+    public void test8078() {
+        assertParsingFails("2.2.3.8");
+    }
+
+    @Test
+    public void test8589() {
+        assertParsingFails("6, 12, 16, 29");
+    }
 }
--- a/main/src/cgeo/geocaching/location/GeopointParser.java
+++ b/main/src/cgeo/geocaching/location/GeopointParser.java
@@ -294,7 +294,7 @@ public class GeopointParser {
 
         //                                        (   1  )    (    2    )    (       3      )
         private static final String STRING_LON = "([WEO]?)\\s*(\\d++°?|°)\\s*(\\d++[.,]\\d++)\\b['′]?";
-        private static final String STRING_SEPARATOR = "[^\\w'′\"″°]*";
+        private static final String STRING_SEPARATOR = "[^\\w'′\"″°.,]*";
         private static final Pattern PATTERN_LAT = Pattern.compile(STRING_LAT, Pattern.CASE_INSENSITIVE);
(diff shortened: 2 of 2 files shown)


Lineflyer:
Not sure if we should really exclude things like having a separator ( e.g. / or -) between N and E.

bekuno:
We should also build some negative tests. 
This meens to test strings which not should end in coords. (like the problem from #8589)

UniQP:
Changed the PR to exclude only some characters between the coordinate parts. Also added some negative tests as requested by @bekuno.

Lineflyer:
@bekuno @moving-bits 
Please feel free to review and merge if OK.

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
```diff
@@ -313,4 +308,23 @@ public void test6802() {
...
     }
 
+    @Test
+    public void test8078() {
+        try {
+            GeopointParser.parse("2.2.3.8");
+            failBecauseExceptionWasNotThrown(Geopoint.ParseException.class);
+        } catch (final Geopoint.ParseException e) {
+            // expected
+        }
```
bekuno, line 318: this can be replaced by 
`        assertParsingFails("2.2.3.8"); `

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
```diff
@@ -313,4 +308,23 @@ public void test6802() {
...
+    }
+
+    @Test
+    public void test8589() {
+        try {
+            GeopointParser.parse("6, 12, 16, 29");
+            failBecauseExceptionWasNotThrown(Geopoint.ParseException.class);
+        } catch (final Geopoint.ParseException e) {
+            // expected
+        }
```
bekuno, line 328: this can be replaced by
assertParsingFails("6, 12, 16, 29");

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
UniQP: Done.

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
UniQP: Done.

moving-bits:
retest this please

baiti:
> We now forbid all special characters of the coordinates to occur between the coordinate parts.

This is not good. The waypoint dialog's coorindate entry pull down provides a function to take coordinates from the clipboard. A very popular application that can put coordinates into the clipboard is Locus Map, however, Locus Map uses the format:

Nxx° yy.zzz', Exx° yy.zzz'

This format was parsed by c'geo correctly until the referenced change was made. Now since the colon between Lat / Lon part makes the parser  reject these coordinates, the effect is that the menu entry "Use coords from clipboard" is not even displayed anymore.

baiti (commented):
This change breaks coordinate parsing when coordinates come from Locus Map via the clipboard. Please backout this change. Coordinates in the form of "Nxx° yy.zzz', Exx° yy.zzz'" should be parsed corrextly as valid coordinates. The only other option I see is to convince the Locus Map folks to remove that colon as it is "ugly" in the first place and doesn't really make sense, however, as a matter of fact, that colon happens to be there since a very long time and  c'geo was indeed able to process it correctly

moving-bits:
@baiti 
Thanks for your feedback. I have created a new issue for this for easier tracking. (It's hard to track things in issues already closed.)

Issue #987, opened 2012-01-14
Offline found count


rsudev:
Suggestion from the german geoclub forum:
For offline logging it would be nice to have a found count as well. So I suggest to introduce a new variable [OFFLINENUM] that relies on some start value that you manually set in the settings and is automatically inceased with every log. It lays the responsibility o n the user that it is correctly set, but allows for convinient offline caching and logging.

TofLaBeuze:
Hi c:geo team,
up for this feature request that would be nice : do offline logging with the correct find numbers (by incrementing the last online number).

Jorgixhallo:
This feedback has been "buried" here for a long time... I face the same problem and I might have a suggestion (at least partial)...
1: When reopening the log in order to send it, the [NUMBER] gets updated if the user is online.
2: Hide the number (or even the other metadata) and update it only when sending. This could be done in the background and user wouldn't be able to perceive that the "same number" is there while offline.

Technical-13:
This is a five year old feature request?  Is it just a lack of coders keeping this from happening or is there a technical limitation making it more difficult to complete?

Lineflyer:
Well, technically it is not so easy to keep track of offline numbering but feasible. So it is more lack of interested coders (with time to find a good process for it).

SammysHP:
Actually it's not that easy. There are a lot of possibilities how something unexpected can happen.

lippih:
Cachebox has a good solution for this feature: There is a field (variable) that you can update online automatically by clicking it, or you can enter your number manually. This number is used for the founds. If you log a found, this number will increase for +1. If you delet a log it decreases for -1. It's very simple but very useful for offline loggers!  Perhaps Cgeo can insert a feature like this?

Lukasdoe:
So you say, that probably a new counter should be added to the c:geo startscreen? Is this really mandatory, because I don't really see how it would solve the problems that someone could encounter with offline caching, other than having a visual confirmation whether something has happened. Probably it's better adding such a counter in the "Log Visit" screen.

itsbrody:
Coming from GDAK (the app pendant of GSAK), which had a similar feature as described by @lippih  I'd second this, too. Sorrily I'm not a developer, but I'd intensivly test it, when existing.

Regards providing an additional counter for offline logging - I don't see the need - but if this is the cgeo solution, I would suggest the following:
- this counter is editable (part of "Logging" settings; "Log Visit" screen could also be an option)
[...]

PR #8603, opened 2020-07-09
first step for [OFFLINENUM] #987

fm-sys:
My first Pull Request, so please be considerate ;-) 

It uses the `DBExtension` to store the found number for each geocaching service. (needed for #987) 

The values will be updated each time the main activity is called and a connection to a geocaching service is established.


bekuno:
ok to test

Lineflyer:
> It creates the shared preferences value cached_found_num which is the cumulated sum of all founds from all geocaching services. (needed for #987)

Just a side note aside from code review:
IMHO using the sum of all geocaching services is not the way to go. You would probably need and increment it for each geocaching service, as we are also e.g. using it depending on geocaching service for the signature place holder.

fm-sys:
#You would probably need and increment it for each geocaching service

Yes, that's right. I also thought about it... But offline number sounds for me like "all founds I did", I
Independent of the service.
But since most users only use one service anyway (GC.com), most people won't notice any difference.
Besides, this is a completely new LogTemplate.  I can call it for example Offline Number (Beta), so that everyone knows that there are still small improvements to be made

fm-sys:
Based on the feedback from @Lineflyer, all founds are now stored in a separate variable for each service...

bekuno:
retest this please

moving-bits:
Thanks for your contribution, very welcomed!

Please do not hardcode the connectors and their keys, as there might be new ones coming along from time to time. Instead use some connector-specific suffix that comes from the connectors itself, is connector-specific and immutable. (Maybe `conn.getNameAbbreviated()` fulfills those conditions, haven't checked.)

If you want to use shared prefs as storage, you can use a generic pref prefix (like `pref_found_num_` in your code) and add the connector-specific suffix to the key.

If you instead use the new extension table (not yet merged, see #8611), you can use your own class derived from `DBExtensionType` and put the connector-specific part into one of the dynamic fields. - (If you want to implement it using the extension table I can show you how to do this, as the first use-case for the extension table, the new map downloader, is not yet pushed as PR.)

fm-sys:
Yes, I have already seen the currently chosen variant as problematic.

I already had a look at your PR for the extension table and wanted to wait until the offline map onbording is finished to orientate myself. 

Would be cool if you could show me how to implement the ExtensionsDB.
I imagine it like this:

`public static long loadFoundNum (final String nameAbbreviated)`
`public static void actualizeNum (final String nameAbbreviated, final long foundNum)`

moving-bits:
@fm-sys:
Extension table supports type, key, and up to four values (internally labelled long1, long2, string1, string2).
Start by defining a new unique enum value in `DBExtensionType` in file `storage/DataStore.java` (in my case: `DBEXTENSION_PENDING_DOWNLOAD(1)`), then create your own class in `storage/extension` dir, derived from `DBExtension`. 

For map downloader I need to store simple key/value pairs (id and filename), one for each download. So I decided to use the download id as `key`, and the filename as `string1` internally. The outside interface is using the domain specific labels `pendingDownload` (= key/id) and `filename`.

It looks like this (import section stripped):

file `storage/extension/PendingDownloads.java`:
```
public class PendingDownload extends DataStore.DBExtension {

    private static final DataStore.DBExtensionType type = DataStore.DBExtensionType.DBEXTENSION_PENDING_DOWNLOAD;

    private PendingDownload(final DataStore.DBExtension copyFrom) {
        this.id = copyFrom.getId();
        this.key = copyFrom.getKey();
        this.string1 = copyFrom.getString1();
    }

    public String getFilename() {
        return getString1();
    }

    @Nullable
    public static PendingDownload load(final long pendingDownload) {
        final DataStore.DBExtension temp = load(type, String.valueOf(pendingDownload));
        return null == temp ? null : new PendingDownload(temp);
    }

    public static void add(final long pendingDownload, @NonNull final String filename) {
        final String key = String.valueOf(pendingDownload);
        removeAll(type, key);
        add(type, key, 0, 0, filename, "");
    }

    public static void remove(final long pendingDownload) {
        removeAll(type, String.valueOf(pendingDownload));
    }
}

```

For your use case you could use the abbreviated connector identifier as key, and the current value into `long1`. Otherwise it's basically the same - `load()` fulfills your `loadFoundNum()` description, `add()` implements the `actualizeNum()` part. (`add()` actually is for inserting as well as for updating the value for a given `key`).

As soon as we have branched for the upcoming release I will merge the extension functions into master branch.

moving-bits:
@fm-sys Your PR contains conflicting changes, probably due to editing the `DBExtensionType` and manually adding an already existing value. You can try to rebase your PR locally to the current master.

Long form: Standard procedure would be:

one time steps:
- close c:geo repo to your own fork on github.com (you have done this)
- download your fork and work locally on it (you're probably doing this as well)
- setup the original cgeo repo as `upstream` in your local copy of your fork, and `origin` pointing to your fork (see `git remote`)

steps for every new PR you're working on:
- for a new PR create your own branch locally (don't work on master branch - makes it pretty hard to work on two 
things at the same time, which will happen more frequently than you may imagine)
  - `git checkout -b <your local branch>`
  - example: `git checkout -b issue_4711`
- regularly sync your local copy with the changes from cgeo repo - I've written a batchfile for this I'm calling every time (eg: every day) I start working on c:geo:
  - `git fetch upstream`
  - `git checkout master`
  - `git pull --rebase upstream master`
  - this ensures that my local `master` branch is always in sync with c:geo's `master` branch
  - (I'm doing a similar command sequence for `release` branch, because I'm working on `release` as well from time to time)
- before pushing your PR from your local copy to your fork and create a PR, make sure you have the most recent changes from c:geo repo locally, therefore in addition to the batch file above do a rebase locally before pushing a commit:
  - `git checkout <your local branch>`
  - `git rebase upstream/master`
  - then push your changes: `git push -u origin <your local branch>`
- During a rebase conflicts may occur. Git shows your the filenames involved. You will have to resolve those conflicts before pushing your changes.
  - Resolve those conflicts using Android Studio. Conflicting areas are marked with `<<<` and `>>>`
  - manually add the cleaned up files to commit stage using `git add <filename>` or `git add -i`
  - after having resolved all conflict continue the rebase with `git rebase --continue`

fm-sys:
I will close this and create a new pull request from another branch when i'm finished, so that I can continue working on other topics...

fm-sys:
@moving-bits My new PR is not finished yet, so I post my question here: Could it be true that all values in DBExtention are deleted when closing c:geo due to the `deleteOrphanedRecords` Method? And what can I do to keep the values?

moving-bits:
> @moving-bits My new PR is not finished yet, so I post my question here: Could it be true that all values in DBExtention are deleted when closing c:geo due to the `deleteOrphanedRecords` Method?

It certainly should not be the case. I'm trying to delete all entries in the extension table which do not belong to one of the defined ID (currently 0, 1, and 2 with your PR), and then, in a second step, clear all that have the 0 value.

fm-sys:
```database.delete(dbTableExtension, "_key NOT IN (" + keys + ")", null);```
then it should be `_type`, not `_key`... 
I will fix this within my PR ;-)

moving-bits:
> then it should be `_type`, not `_key`...

@fm-sys 
Ah, you're right, good catch!
And thanks for fixing this.

Issue #8610, opened 2020-07-11
Ease onboarding with offline maps, part 2


moving-bits:
This is part two to ease onboarding with offline maps for new users of c:geo, in addition (or as partial replacement) of #8180/#8183:

- add a link to the map selector: "Download offline maps"
- which opens a directory listing of the Mapsforge server (actually its mirror at fh-esslingen.de)
- You can browse through the directories up and down and select a file (map) to download.
- Map file gets downloaded in the background using Android's download manager.
- On completion the map file gets copied into c:geo map directory and activated as current map. (as in #8183).

This is how it looks like:

New menu entry "Download offline map" in map selection menu:
![image](https://user-images.githubusercontent.com/3754370/87226999-05531e00-c398-11ea-9d42-85145273d8b3.png)
(the thin separator line is available on Android 10+ only)

Retrieved the initial listing of the v5 map directory:
![image](https://user-images.githubusercontent.com/3754370/87227020-26b40a00-c398-11ea-9ec2-e234ba27754c.png)

Retrieving a sub directory, "Australia-Oceania" in this case:
![image](https://user-images.githubusercontent.com/3754370/87227029-39c6da00-c398-11ea-829b-9ca4bfc50327.png)

For downloadable map files size and date of last change are shown:
![image](https://user-images.githubusercontent.com/3754370/87227043-4cd9aa00-c398-11ea-86d0-2c00b121423f.png)

When you select a map for downloading, you get a short notification that the download starts in the background. You can continue to work with c:geo until the download has finished:
![image](https://user-images.githubusercontent.com/3754370/87227066-6c70d280-c398-11ea-9ad5-68d65904680c.png)

When the download has finished the file gets copied to c:geo map dir automatically:
![image](https://user-images.githubusercontent.com/3754370/87227074-83afc000-c398-11ea-8080-44cea75b8ea3.png)

Finally you get a confirmation that the download has succeeded:
![image](https://user-images.githubusercontent.com/3754370/87227079-96c29000-c398-11ea-9017-c679019f186c.png)
The newly downloaded map gets activated as current map automatically.

---

Current state of this PR is beta. Next steps are:
- creating and merging the "extension table" PR, which is a prerequisite
- a bit of minor roundup on this, then creating "map download" PR
- testing, testing, testing, ...

Any feedback is very welcome.

fm-sys:
Wouldn't it make more sense to move the downloaded map to the map directory instead of copying it?

This would be faster and save memory space...

PR #8611, opened 2020-07-11
add extension table

moving-bits:
Sometimes there is the need to store a few values, mostly transiently, into the database. During my work on #8610 I've again stumbled upon this, so I created a minimalistic extension framework for this.

Base class is DataStore.DBExtension, which defines a simplistic key/value store with a type (enum), key (String) and two long + two string values (usable individually), and methods to access it. To use this, you should derive you own class from this and expose the parts you need under domain specific names.

Two examples for (possible) usage:
- For my maps downloader I need to store a temporary id from the download managerer + additional information regarding the selected map, so I created a `PendingDownload extends DBExtension` class, which uses the fields `key` (= id of pending download) and `string1` (filename), exposed under that names. (See upcoming PR for map downloader.)
- There is another PR under construction regardings offline counter. This could use the extension table as well, using one of the long fields for the counter, and either a string or a long field to dynamically identify the geo service.

Exposed interface for use case one will be:
```
// return associated filename of a pending download
public String getFilename() { }

// lookup data regarding a certain pending download
public static PendingDownload load(final long pendingDownload) { }

// save data regarding a new pending download
public static void add(final long pendingDownload, @NonNull final String filename) { }

// remove data associated with pending download
public static void remove(final long pendingDownload) {  }
```
completely hiding the underlaying technical things.
Implementation for this use case will be part of the "map download" PR, which builds upon this PR.

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -39,6 +39,7 @@ import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.Version;
 import cgeo.geocaching.utils.functions.Func1;
 import static cgeo.geocaching.settings.Settings.getMaximumMapTrailLength;
+import static cgeo.geocaching.storage.DataStore.DBExtensionType.DBEXTENSION_INVALID;
 
 import android.app.Activity;
 import android.app.ProgressDialog;
@@ -107,6 +108,17 @@ public class DataStore {
         DATABASE,
     }
 
+    public enum DBExtensionType {
+        // values for id must not be changed, as there are database entries depending on it
+        DBEXTENSION_INVALID(0);
+
+        public int id;
+
+        DBExtensionType(final int id) {
+            this.id = id;
+        }
+    }
+
     private static final Func1<Cursor, String> GET_STRING_0 = cursor -> cursor.getString(0);
 
     private static final Func1<Cursor, Integer> GET_INTEGER_0 = cursor -> cursor.getInt(0);
@@ -170,7 +182,7 @@ public class DataStore {
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
-    private static final int dbVersion = 81;
+    private static final int dbVersion = 82;
     public static final int customListIdOffset = 10;
 
     @NonNull private static final String dbTableCaches = "cg_caches";
@@ -187,6 +199,7 @@ public class DataStore {
     @NonNull private static final String dbTableSearchDestinationHistory = "cg_search_destination_history";
     @NonNull private static final String dbTableTrailHistory = "cg_trail_history";
     @NonNull private static final String dbTableRoute = "cg_table_route";
+    @NonNull private static final String dbTableExtension = "cg_table_extension";
     @NonNull private static final String dbTableSequences = "sqlite_sequence";
     @NonNull private static final String dbCreateCaches = ""
             + "CREATE TABLE " + dbTableCaches + " ("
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
To be merged in conjunction with the new "map downloader", probably not in the now to be prepared upcoming release, therefore setting "Do not merge" for now.

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
rebased to current master - let's see, if the CI stays stable then

moving-bits:
@Lineflyer @mucek4 
Can someone explain that build error `com.android.builder.testing.ConnectedDevice.No tests found.`
I've already created a fresh PR (#8680) in case something has been wrong with the original PR, but the same error is coming up.
Can this be related to this PR having a database change in it?

Lineflyer:
retest this please

Lineflyer:
mmh:
```
21:58:59  Tests on android-23(AVD) - 6.0 failed: Instrumentation run failed due to 'java.lang.NullPointerException'
21:58:59  
21:58:59  com.android.builder.testing.ConnectedDevice > No tests found.[android-23(AVD) - 6.0] [31mFAILED [0m
21:58:59  No tests found. This usually means that your test classes are not in the form that your test runner expects (e.g. don't inherit from TestCase or lack @Test annotations).
```

Lineflyer:
Maybe a NPE while starting the app on the emulator?
Did you test on your side with an Android 6 emulator (AFAIK we use Android 6 on CI).

moving-bits:
> Maybe a NPE while starting the app on the emulator?
> Did you test on your side with an Android 6 emulator (AFAIK we use Android 6 on CI).

Not yet, will give it a try. (As this PR contains a DB change I tried to focus on only one of my VMs to keep the others "clean" for ongoing development.)

bekuno:
can reproduce on my machine:
~~~

> Task :main:copyDefaultPreferencesToAndroid
cgeo.geocaching_preferences.xml: 1 file pushed, 0 skipped. 0.1 MB/s (307 bytes in 0.002s)
run-as: Could not set capabilities: Operation not permitted
Process 'command 'C:\Users\admin\AppData\Local\Android\Sdk\platform-tools\adb.exe'' finished with non-zero exit value 255
...
> Task :main:connectedBasicDebugAndroidTest
Starting 0 tests on SM-G920F - 7.0
Tests on SM-G920F - 7.0 failed: Instrumentation run failed due to 'Process crashed.'

com.android.builder.testing.ConnectedDevice > No tests found.[SM-G920F - 7.0] FAILED 
No tests found. This usually means that your test classes are not in the form that your test runner expects (e.g. don't inherit from TestCase or lack @Test annotations).
~~~

moving-bits:
I just fired up an Android 6 emulator and was able to start c:geo.

Only the about dialog stays empty, as https://github.com/cgeo/cgeo/commit/e05863e80186e300d79b0d0abc6b5cee16284a32 seems not to be on master yet, but other than that, everything ran smoothly.

Will try running the test suite next on that VM.

moving-bits:
Tests run fine in the VM.
About 20 tests (out of 514) fail, but that's due to no login data and no OC keys supplied in my config in that VM.

bekuno:
@moving-bits Maybe this problem?
~~~
2020-07-26 22:59:22.074 23065-23065/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: cgeo.geocaching, PID: 23065
    java.lang.RuntimeException: Unable to create application cgeo.geocaching.CgeoApplication: java.lang.NullPointerException: Attempt to invoke virtual method 'android.database.sqlite.SQLiteStatement android.database.sqlite.SQLiteDatabase.compileStatement(java.lang.String)' on a null object reference
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5879)
        at android.app.ActivityThread.-wrap3(ActivityThread.java)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1699)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6682)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1520)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1410)
     Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.database.sqlite.SQLiteStatement android.database.sqlite.SQLiteDatabase.compileStatement(java.lang.String)' on a null object reference
        at cgeo.geocaching.storage.DataStore$PreparedStatement.getStatement(DataStore.java:3841)
        at cgeo.geocaching.storage.DataStore$PreparedStatement.simpleQueryForLong(DataStore.java:3833)
        at cgeo.geocaching.storage.DataStore.getLists(DataStore.java:3268)
        at cgeo.geocaching.CgeoApplication.onCreate(CgeoApplication.java:55)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1032)
~~~

bekuno:
and:
~~~
2020-07-26 23:02:41.005 25227-25227/cgeo.geocaching E/SQLiteLog: (1) no such table: main.cg_table_extension
2020-07-26 23:02:41.030 25227-25227/cgeo.geocaching E/cgeo: [main] DataStore.init: unable to open database for R/W
    android.database.sqlite.SQLiteException: no such table: main.cg_table_extension (code 1): , while compiling: CREATE INDEX IF NOT EXISTS in_extension_key ON cg_table_extension (_key)
    #################################################################
    Error Code : 1 (SQLITE_ERROR)
    Caused By : SQL(query) error or missing database.
    	(no such table: main.cg_table_extension (code 1): , while compiling: CREATE INDEX IF NOT EXISTS in_extension_key ON cg_table_extension (_key))
    #################################################################
~~~

moving-bits:
> @moving-bits Maybe this problem?
> 
> ```
> 2020-07-26 22:59:22.074 23065-23065/? E/AndroidRuntime: FATAL EXCEPTION: main
>     Process: cgeo.geocaching, PID: 23065
>     java.lang.RuntimeException: Unable to create application cgeo.geocaching.CgeoApplication: java.lang.NullPointerException: Attempt to invoke virtual method 'android.database.sqlite.SQLiteStatement android.database.sqlite.SQLiteDatabase.compileStatement(java.lang.String)' on a null object reference
> ```

If I'm not mistaken this is the error message when the database has a higher internal version number than the current code is expecting.
Eg.: You run this PR #8611, it upgrades your database to version 82, and later on you try to run current release version, which does not include this change yet, and still is at DB version 81.
SQLite then tries to downgrade your database, but fails, as we do not have this method implemented.

moving-bits:
@bekuno 
If you want to downgrade database version in your VM after having run this PR follow those steps:
- checkout a version not having this PR included (eg.: master or release branch)
- add the following code to storage/DataStore.java
- compile & run once
- then remove the following lines

```
@Override
public void onDowngrade(final SQLiteDatabase db, final int oldVersion, final int newVersion) {
    if (oldVersion == 82) {
        db.beginTransaction();
        db.execSQL("DROP TABLE IF EXISTS cg_table_extension");
        db.execSQL("DROP INDEX IF EXISTS in_extension_key");
        db.setTransactionSuccessful();
        db.endTransaction();
    } else {
        throw new IllegalStateException("cannot downgrade");
    }
}
```

moving-bits:
Let's see if it works if I leave out index creation for now

bekuno:
> If I'm not mistaken this is the error message when the database has a higher internal version number than the current code is expecting.

I think you are right.
I will test this reset code if I find time. Thanks for sharing it!

> Let's see if it works if I leave out index creation for now

The test ran fine now :-)

moving-bits:
> > Let's see if it works if I leave out index creation for now
> 
> The test ran fine now :-)

That's good news. I missed this, as I ran the code first (which created the table and the index fine), and run the test later (so table did already exist).

There seems to be a difference in test management compared to a regular installation - `createIndices()` called before db upgrade routines during testing, leading to the error. I only wonder when this has changed, as we have earlier db changes with the same order I've used - create new table, then call `createIndices()` during the same update step...

moving-bits:
Ok, let's merge this one to be able to move on with things based on this change, and add a reminder to look into the differences in index creation between test case and regular installation.

Issue #8467, opened 2020-06-18
Limit file types in intents


Lineflyer:
After installing the newer nightlies, my phone offeres "c:geo local file manager" for nearly every file type (e.g. also APK installs).

Sorry, that I did not have time yet to understand why this is needed after all.
What is the purpose of this? Can't we limit it to a certain file type? Is the the use case to open files with it, or just an unwanted side effect?

moving-bits:
> After installing the newer nightlies, my phone offeres "c:geo local file manager" for nearly every file type (e.g. also APK installs).

This is a consequence of #8180 and its implementation with #8183.

> Sorry, that I did not have time yet to understand why this is needed after all.
> What is the purpose of this? 

Reasoning behind this is to ease onboarding for offline maps. For new users of c:geo it's very hard to understand how to find and to download offline map files, and how to integrate that into c:geo.
With the changes of #8183 you can tap on a link from within c:geo, select a map file for download, and after downloading has finished tap on the downloaded file to integrate it into c:geo.
The last step is the reason why c:geo is now offered more frequently for files.

> Can't we limit it to a certain file type? Is the the use case to open files with it, or just an unwanted side effect?

Android does not offer a way to limit this to certain file extensions, only to MIME types. But as "map" files do not have a specific MIME type, the servers do give us unspecific ones (and varying such), which I have to catch all to be able to get c:geo into the line if a user taps on a map file.

This is similar to what we already have for gwc / gwz / zip files, see here: https://github.com/cgeo/cgeo/issues/7663
And some discussion about it: https://github.com/cgeo/cgeo/issues/8180#issuecomment-611165666

OTOH the implementation with #8183 has the wanted side effect, that you can tap on any local map file to integrate it into c:geo.

And alternative and/or extension to this would be to implement #6072/#7529, but nobody has found a way and/or the time to do so yet.

Lineflyer:
I have read those limitations for the intents and hope at least partly understood the problem here.
But anyhow from users perspective (and thinking of the amount of users hitting support mail with questions like "Why is this shown now for all different files") it not reasonable to implement it in that way then.

Additionally for me it does not work. If I click on a valid OSM *.map file, my phone reports "no app installed to open this file type".

My question would be, if it is really worth to have this map import click feature with these side effects? Can't we instead just fix / enhance the scanning for map files in the app?

[...]

PR #8617, opened 2020-07-12
disable #8183, restore old AndroidManifest

moving-bits:
- disables the changes introduced by #8183 
- restores the old AndroidManifest

which disables c:geo's possibility to read locally stored map files, but avoids the side effects described in #8467.

Also disables the menu entry "Download basic offline map" in settings => map, as it doesn't really make sense without the other part. Menu entry can be re-enabled and re-targeted when the new "map downloader" is merged.


moving-bits:
retest this please

PR #8628, opened 2020-07-13
alt. implementation for fast scroll, mostly avoiding onScrollStateChange

moving-bits:
As `onScrollStateChange()` seems to miss a lot of `SCROLL_STATE_IDLE` notifications this alternative implementation (mostly) switches to `onScoll()`, setting a timer for 1s, and checks whether the last scroll position is the same as the current position. If so, fastscroll gets disabled, otherwise the timer is restarted.

Enabling fastscroll is unchanged compared to the current implementation.

Additional change is that the fast scroll bar is set to "always visible", so that as long as fast scroll mode is active, fast scroll bar is visible (instead of fading out, but still being active for some time). If you see the thinner position marker, fast scroll is disabled.


moving-bits:
retest this please

moving-bits:
retest this please

Issue #5854, opened 2016-07-24
Avoid duplication of personal note


Bananeweizen:
Since I again got reports of notes duplicating themselves, I suggest that we strip all whitespace characters when comparing the offline and online version. This might help with proxies or other outside influences when comparing the notes.

Lineflyer:
AFAIR there is an issue about unnecessary duplication of the note...have to look it up as time permits.

MagpieFourtyTwo:
Just to remind: This bug still exists - and is extremely annoying, esp. if you are used to solve Multis and Mysteries using the personal note. Please have a look at #5286 for a more detailed scenario description. And #5456 for even a screenshot.

Perhaps the following add. info and workaround may help finding the bug:
- the duplicate is a perfect copy of either the current or the previously saved personal note
- it does not matter if a personal note has been uploaded to gc.com before, it even happens on newly saved caches which do not have a pn on gc.com at all
- unfortunately, this is still not deterministic, it sometimes takes a bunch of edits until the pn gets duplicated and in some cases it even will not be duplicated at all
- but, once duplicated, it will *always* be duplicated again, when removing the complete duplicated text

My current workaround:
- do *not* remove the *complete* duplicated text, but just remove most of the upper version: leave esp. the last line of the upper version and the separating "--" line untouched - doing so, the note will not be duplicated again.

rsudev:
@MagpieFourtyTwo Thanks for bringing this up again. I played around with it a little and failed to reproduce it (perhaps expected, as it was already described as not deterministic).
That makes it of course quite hard to address, because without a known trigger it is not only hard to find out what is happening but even a problem to verify any mitigation/fix.
From your experience you might be able to give some hints regarding possible triggers.
- Does it seem to relate to the content, some specifics in there?
- The pattern of interaction with the cache (editing note, then wp, than map, than ...)?
- The pattern of interaction with the app as a whole (closing, locking screen, switching to other apps, ..)?
or any combination thereof?
If we can find a way to make this more reproducible, I promise to address it...

rsudev:
Shortly after writing these lines, I found a way to reproduce a duplication.
- open a cache
- edit the PN
- open the cache map
- open the cache from there (popup - details)
- edit the PN
- back to map
- back to cache: the old PN is still shown
- edit the PN
- leave cache
- re-enter cache: PN is duplicated

That should help to trace this issue...

MagpieFourtyTwo:
[...]

PR #8643, opened 2020-07-16
Solving Issue 5854: avoid duplication of personal note 

eddiemuc:
Attempts to solve #5854 
This PR removes **completely** the existing "merge" logic for personal notes when local caches are refreshed from providers and replaces it with a much simpler logic: when a cache is refreshed from provider then:
* the local note is preserved as is
* if a note exists at provider side, it is split using the well-known "\n--\n" separator. For any of the provider site "subnotes" found that way: if it is not found in local note (using `String.contains() `) then it is appended to local note using "\n--\n"

This should IMHO remove the side-effect duplicate creations we experience nowadays while preserving changes on provider side as best as possible

Note that (like before) the upload of user notes from cgeo to provider overwrites the provider note without any mercy or check!


Lineflyer:
ok to test

Lineflyer:
retest this please

moving-bits:
Ok, let's merge and test this one.

@eddiemuc 
Thank you for your contribution, highly appreciated!

MagpieFourtyTwo:
@eddiemuc 
This fix (resp rework), BTW, is working amazingly GREAT!
It's a real relief to have this issue solved.
Therefore, it is an urgent need for me to say thank you as well! :)

eddiemuc:
@MagpieFourtyTwo thank you :-) I really like the way the cgeo Team works together and how constructive and respectful you tread each other. Great spirit!

PR #8644, opened 2020-07-16
Adaptive icon for cgeo-contacts #8558

pstorch:
The cgeo-contacts PlugIn is not really visible in the AppDrawer / Launcher. I only see it in my Android Settings / Apps List.
But I guess the Google Play Console requires an adaptive icon when uploading a new release.

cgeo-contacts-launcher


moving-bits:
retest this please

Lineflyer:
@pstorch Thanks very much. I thi k we do now have new icons for all we need. We do however have a sendtocgeo icon somewhere (at least in Useful apps list of c:geo). We do neither need app icons nor playstore icons however same style would be nice...but feel free to skip that anyway.

Issue #7729, opened 2019-07-17
CI fails constantly for Geocoder test (GRPC failed)


Lineflyer:
As you might have already seen for my PR tests on CI we almost constantly have the problem of the failing geocoder test.

```
java.lang.RuntimeException: java.io.IOException: grpc failed
at io.reactivex.internal.util.ExceptionHelper.wrapOrThrow(ExceptionHelper.java:46)
at io.reactivex.internal.observers.BlockingMultiObserver.blockingGet(BlockingMultiObserver.java:93)
at io.reactivex.Single.blockingGet(Single.java:2835)
at cgeo.geocaching.address.GeocoderTest.testGeocoder(GeocoderTest.java:58)
at cgeo.geocaching.address.GeocoderTest.testAndroidGeocoder(GeocoderTest.java:35)
at java.lang.reflect.Method.invoke(Native Method)
at junit.framework.TestCase.runTest(TestCase.java:168)
at junit.framework.TestCase.runBare(TestCase.java:134)
at junit.framework.TestResult$1.protect(TestResult.java:115)
at android.support.test.internal.runner.junit3.AndroidTestResult.runProtected(AndroidTestResult.java:77)
at junit.framework.TestResult.run(TestResult.java:118)
at android.support.test.internal.runner.junit3.AndroidTestResult.run(AndroidTestResult.java:55)
at junit.framework.TestCase.run(TestCase.java:124)
at android.support.test.internal.runner.junit3.NonLeakyTestSuite$NonLeakyTest.run(NonLeakyTestSuite.java:63)
at junit.framework.TestSuite.runTest(TestSuite.java:243)
at junit.framework.TestSuite.run(TestSuite.java:238)
at android.support.test.internal.runner.junit3.DelegatingTestSuite.run(DelegatingTestSuite.java:103)
at android.support.test.internal.runner.junit3.AndroidTestSuite.run(AndroidTestSuite.java:69)
at android.support.test.internal.runner.junit3.JUnit38ClassRunner.run(JUnit38ClassRunner.java:103)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
[...]

PR #8645, opened 2020-07-16
Fix issue 7729: CI fails constantly for Geocoder test

eddiemuc:
Fixes issue #7729 
With respect to the known open android bug, a workaround is implemented to ignore the "grpc fail" exceptions in test case


Lineflyer:
add to whitelist

For our CI

moving-bits:
@eddiemuc 
Thanks for preparing this workaround

PR #8652, opened 2020-07-18
remove geokrety auth token from limited shared prefs backup

moving-bits:
Auth keys / tokens etc. should not be part of the limited shared prefs backup, therefore exclude the geokrety auth token prefkey from it.

(Cause was: Auth token for geokrety seems to be handled a bit different than the other auth tokens, using a different prefkey structure, so it has been missed so far.)


moving-bits:
retest this please

PR #8657, opened 2020-07-19
apply fastscroll listener to settings view

moving-bits:
Current settings view has the problem, that some buttons cannot be reached due to fast scroll with thumbs being activated. Applying our new fastscroll listener automatically disables fastscroll after one second of (fling mode) scrolling inactivity, which also solves the button problem.


moving-bits:
retest this please

Issue #8664, opened 2020-07-21
[Nightly] crash with any OSM v5


MagpieFourtyTwo:
**Describe the bug:**
c:geo crashes in map view with any OSM v5
Guess it's related to latest history line fix ...?

**Version used**
2020.07.21-NB

[logcat_2020-07-21_03-19.txt](https://github.com/cgeo/cgeo/files/4950729/logcat_2020-07-21_03-19.txt)

MagpieFourtyTwo:
FYI: Downgraded to 2020.07.20-NB-33a57ef (which BTW worked like a charm ;) )

moving-bits:
Very strange. I can see the errors in the logfile, but I cannot find a clear cause, code looks inconspicuous. And cannot reproduce in the emulator - map is stable there and works like a charm.

Can you describe the specific steps you did, until that error occurred?
Does it happen directly on map startup, or when moving pretty fast/slow, or anything other "special"?

Besides that: I adapted the algorithm for Mapsforge NewMap to that from Google Maps v2 - let's see if that's better (see #8665).

MagpieFourtyTwo:
Perhaps because of a different constellation of options and/or data?
Over here there's nothing special, piece of cake:

With OSM activated ...
- open c:geo, open live map - crash.
- open cache detail page, show cache on map - crash.
- set map to GM (in options), open live map, switch to OSM - crash.

Maybe deletion of history track may "fix" this, but I did not try yet, cause a) did not want to destroy test data and b) "old" tracks should work with the new strategy, too, otherwise Joe Public must clear history line prior to update, too ... which will for sure lead to massive increase of support inquiries ... would like to protect @Lineflyer as much as possible ... ;)

moving-bits:
I had tested with both a grown history track as well with a freshly cleared one - made no difference. Maybe it's some timing thing, which does not happen in the emulator, because that one's so slow...

MagpieFourtyTwo:
> I adapted the algorithm for Mapsforge NewMap to that from Google Maps v2 - let's see if that's better (see #8665).

Ok so I should re-check the crashes with this night's nightly?

BTW: You have in mind that #8663 happens with GM, too?

moving-bits:
> Ok so I should re-check the crashes with this night's nightly?

yes, please

> BTW: You have in mind that #8663 happens with GM, too?

no, hadn't. but #8663 confuses me anyway :-)

MagpieFourtyTwo:
[...]

Issue #8663, opened 2020-07-20
[Nightly] History line not drawn properly (or not at all)


MagpieFourtyTwo:
**Describe the bug:**
History line is not drawn properly if map view is closed and re-opened again.

**To Reproduce:**
1. open map view (no matter if live or cache context, tested with GM and OSM v5)
2. move - around corners (to produce edges) ... and watch history line growing ...
3. leave map view (not matter if via switching display off or changing to cache detail page or to a different app)
4. move a little
5. re-open map view ...
6. History line will not be (re-)drawn like it was before, i. e. show the curves you've taken (and which have definitely been recorded in step 2.), but will be a straight bee's line back to a way older point (I would say, it's the one where you have been when you switched to map view the first time). BTW: Previously recorded parts of the history will neverhtheless be shown correctly.
7. move a little ... and watch history line growing again: newly recorded parts will be drawn properly again (while the line from step 6 will remain straight)
8. leave map like in step 3.
9. move a little 
10. re-open map view ...
11. newly recorded parts (step 7) will again not be shown, and there will again just be a straight line from your current position to the position mentioned in step 6.

Instead of steps 8. to 11. it sometimes is even sufficient to switch map products (e. g. from OSM to GM or vice versa).

**Expected behavior/state after performing these steps:**
History line should always display *all* recorded points.

**Version of c:geo used:**
2020.07.20-NB-33a57ef

**Is the problem reproducible:**
Yes

**Additional context:**
Although these newly recorded parts of the history line seem to have vanished, they will suddenly appear again on a fresh start of c:geo (you may even have to remove c:geo from "recent apps").

Moreover: Sometimes, when opening map view after some usage of c:geo, the entire history line will not be displayed at all - until you do a fresh restart, too (and sometimes just switching map context [e. g. from live to map context] may be sufficient) - but most regretfully I was not able to find a deterministic way to reproduce this ... perhaps someone else observed it, too ...?

MagpieFourtyTwo:
@moving-bits wrote in #8664 :
> I adapted the algorithm for Mapsforge NewMap to that from Google Maps v2 - let's see if that's better (see #8665).

[...]

PR #8665, opened 2020-07-21
adapt history draw algorithm from GMv2 to NewMap

moving-bits:
Let's see if this can fix issue #8664


MagpieFourtyTwo:
BTW: The scenario described in #8663 happens with GM, too ...

PR #8667, opened 2020-07-22
work on trail copy to display + return correct trail segment on load (fix #8663)

### LINKED ISSUE #8663: [Nightly] History line not drawn properly (or not at all)

moving-bits:
Using the original trail object lead to strange display artifacts, especially after returning from subview.
Now works on copy of trail for display, in both GMv2 and in NewMap.

Additionally return the correct segment of the trail on loading (most recent part, in correct order).

Attempts to fix #8663.

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -269,7 +269,7 @@ public class GooglePositionAndHistory implements PositionAndHistory, Route.Route
         }
         historyObjs.removeAll();
         if (Settings.isMapTrail()) {
-            final ArrayList<Location> paintHistory = getHistory();
+            final ArrayList<Location> paintHistory = new ArrayList<>(getHistory());
             final int size = paintHistory.size();
             if (size < 2) {
                 return;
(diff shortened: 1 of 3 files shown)


moving-bits:
retest this please

PR #8671, opened 2020-07-25
add new icon + background to about page (fix #8661)

### LINKED ISSUE #8661: Replace logo in About section?
Shall we also replace the logo in "About c:geo" with the new one?
While I also like the "old" it would only be consistent to also use the new one here.

moving-bits:
Uses the new launcher icon (on newer API versions) or scaled standard squircle version on older API versions.
Uses c:geo homepage header image as background for the first section.

image


fm-sys:
Looks great! 

Only one single question: 
Would it be possible to add some more space above the icon, so that it is more in the middle and not so squeezed to the edge of the background image?

moving-bits:
Added a bit more space above the logo:

image

PR #8680, opened 2020-07-26
add extension table

moving-bits:
functionally identical to #8611 - let's see if CI can process this one


moving-bits:
same error on CI as with the original #8611, therefore closing here

Issue #7818, opened 2019-09-03
different usage of userName and displayName 


bekuno:
**Describe the bug:**
Normally the users userName and the displayName are the same.
Sometimes not.
Then the displayName is shown on most places, but not in the cache owner details page.

**To Reproduce:**
Open Cache OC13613.
The owner row shows "team-ringel"
![20190903_230447](https://user-images.githubusercontent.com/32555742/64209039-cf9ff780-ce9f-11e9-9570-6d3ddc17b837.png)

Click on the owner row.
You will see in the head line:
**About 264765**
![20190903_230654](https://user-images.githubusercontent.com/32555742/64209048-d29ae800-ce9f-11e9-88b7-52a187e36fcd.png)

**Reason:**
In UserClickListener.showUserActionsDialog line 61:
```
        builder.setTitle(res.getString(R.string.user_menu_title) + " " + user.userName);
```

**Version of c:geo used:**
own dev, branch master, revision 451444caca8ffdcadbdfc286cd9a539dbb0b4678

**Is the problem reproducible:**
Yes

Lineflyer:
IMHO we also have a related issue, that opening the user profile or some other context option is not working due to this

Lineflyer:
I think your report reveals the root cause of #7125

bekuno:
The problem in this issue is only cosmetic.
The problems in #7125 are not a consequence of this. (but probaly simular)

moving-bits:
reopened after #8703

Issue #7125, opened 2018-08-24
Owner actions not working


Lineflyer:
##### Detailed steps causing the problem:
- Open OC cache (e.g. OC1120A)
- Click on owner name to invoke context menu

##### Actual behavior after performing these steps:
The context title shows an ID rather than the user name
All associated actions (e.g. show hides) return no results

##### Expected behavior after performing these steps:
I guess this should rather use the owner name instead of the user ID to work as intended

##### Version of c:geo used:
2018.08.22-NB

##### Is the problem reproducible for you?
Yes

bekuno:
OK:
- Open profile in browser
- Send message

KO:
(- Open contact card)
- Caches hidden
- Caches found
- Copy to clipboard

bekuno:
I checked all 4 KO points, used GC and some OC sites. 
The points are not defined for SU caches.

moving-bits:
reopened after #8703

PR #8698, opened 2020-07-28
userName/displayName in different meaning in GC and OC

### LINKED ISSUE #7998: "Caches hidden" fails to load with customized "cache owner" name
**Describe the bug:**
When viewing cache details, selecting "Caches hidden" from the CO name menu, fails to load when CO have specified a custom username for the currently displayed cache.

While I'm not sure, it seem as though "Caches hidden" tries to load the caches of a user with the username displayed for av given a cache. The displayed **name** for a CO, isn't necessarily the same as the username associated with a cache. The CO name is a free form name, on a cache by cache basis. All caches are still associated with a specific account/user, this should be used to load the caches hidden.

**To Reproduce:**
1. Open cache details for a cache with customized CO name
2. Click cache owner and select "Caches hidden"

c:geo tries to load caches hidden by a user with the displayed name. 

Caches hidden by the proper CO should be loaded.

**Version of c:geo used:**
2019.11.28

Yes

**Additional context:**
This bug was introduced some time back, maybe a couple of months.

bekuno:
Set to the old behavior - works for GC, works not for OC, unknown for other cache connectors
needs to create a better solution for all connectors

fixes #7998: "Caches hidden" fails to load with customized "cache owner" name
revert #7125: Owner actions not working
revert #7818: different usage of userName and displayName

### DIFF
--- a/main/src/cgeo/geocaching/connector/AbstractConnector.java
+++ b/main/src/cgeo/geocaching/connector/AbstractConnector.java
@@ -274,14 +274,14 @@ public abstract class AbstractConnector implements IConnector {
         final List<UserAction> actions = getDefaultUserActions();
 
         if (this instanceof ISearchByOwner) {
-            actions.add(new UserAction(R.string.user_menu_view_hidden, context -> CacheListActivity.startActivityOwner(context.getContext(), context.displayName)));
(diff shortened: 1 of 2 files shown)


bekuno:
retest this please

bekuno:
retest this please

moving-bits:
retest this please

bekuno:
false base branch - create new PR

PR #8707, opened 2020-07-29
OSM offline map files downloader (fix #8610)

### LINKED ISSUE #8610: Ease onboarding with offline maps, part 2

moving-bits:
For detailed description and screenshots see #8610

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -845,7 +845,15 @@
             android:name="cgeo.geocaching.connector.gc.PocketQueryListActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/search_pocket_title"
-            android:parentActivityName="cgeo.geocaching.MainActivity"            >
+            android:parentActivityName="cgeo.geocaching.MainActivity">
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value="cgeo.geocaching.MainActivity" />
+        </activity>
+        <activity
+            android:name="cgeo.geocaching.settings.MapDownloadSelectorActivity"
+            android:label="@string/downloadmap_title"
+            android:parentActivityName="cgeo.geocaching.MainActivity">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value="cgeo.geocaching.MainActivity" />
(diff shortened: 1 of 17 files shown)


moving-bits:
> copying the downloaded file makes less sense from what I understand. Please move the file to the correct map directory instead. Thats faster and saves disk space. Or isn't that possible due to the android permissions?

I'm not sure regarding permissions, but the main reason I'm not moving the file is that renaming/moving only works as long as source and target files are on the same mount points. To my understanding Anrdoid standard download folder is on internal storage, and many (most?) users will store c:geo data an sdcard (which would be a different mount point). What I can try is renaming first, see if it works, and do a copy only on failure to rename.

Regarding storage space: Downloaded files are managed by Android, and deleted automatically on running out of space (AFAIK).

Lineflyer:
Shouldn't we make this available from map settings menu as well?

moving-bits:
> Shouldn't we make this available from map settings menu as well?

Yes - this is the function I would like to put behind the "map icon" button we've searched a couple of days ago...

Lineflyer:
> Yes - this is the function I would like to put behind the "map icon" button we've searched a couple of days ago...

Excellent...today I was already looking there

moving-bits:
> > Yes - this is the function I would like to put behind the "map icon" button we've searched a couple of days ago...
> 
> Excellent...today I was already looking there

See #8724.

Issue #8699, opened 2020-07-28
Extend map lines reset function to cover line widths as well


moving-bits:
Taken from https://github.com/cgeo/cgeo/issues/8683#issuecomment-665279897 by @MagpieFourtyTwo:

What about widths and opacities?
Have seen in the diff that something happens to line width, too, but had not enough context to see where it came from ...
So, if not done anyway, I would suggest to add a possibility to revert the other values to default, too.

Lineflyer:
I would also have expected a reset of all the manual settings.

And (as I also read somewhere around):
I would also strongly suggest to put this reset switch (IMHO it can be a single reset for all custom values) into the map settingf rather than into the map UI. This is probably nothing you want to change on the fly but when you messed it up before in settings.

moving-bits:
Question is: When / where do you recognize that you have messed up the map line settings? That's why I put it into a map menu instead of settings menu.

MagpieFourtyTwo:
@moving-bits:
> Question is: When / where do you recognize that you have messed up the map line settings? 

Following this argumentation even all the line settings themselves should be placed in the map menu ...  ;)
As for me the rules for placing a menu entry should take into consideration ...
- how often an entry will be used
(the less, the more hidden it can be - and this one I will probably never ever use anyway ;) )
- how significant the evoked changes are
(one tap, a quick confirmation and all your line settings, tweaked in weeks of trial and error, are gone! \*aaargh*)
- and for a reset: which settings will be reset
(to possibly be able to write them [or some] down, just in case you wanted to reset just some of them)

Therefore (at least IMNSHO) a ``Reset`` should never ever be placed at a context or shortcut menu.

@Lineflyer: 
> IMHO it can be a single reset for all custom values

Sorry. but totally disagreed. If implemented as the name suggests, this would (or may) revert *all* individual settings to default - which esp. for John Average would be more horrifying than helpful, as some of the custom ones are important for functionality (like e. g. "map/themes directory" ...) and should be "more permanent", while others (like "map rotation", "display all waypoints" ...) substantially change c:geo's behavior. And then there is the rest which has just been changed because of curiosity and John Average's urge to play ...

Who will say, which is which?  And how do you want to communicate the extent of changes to the one, who will be frivolous enough to really use this button?  ;)

So, *if* there should be a possibility to revert to defaults (which I would generally agree to), it should either be per section or with a dialogue where you can select which ones should be reset and which should stay.

moving-bits:
[...]

PR #8720, opened 2020-07-31
add default function to ColorpickerPreference

moving-bits:
A follow-up to the discussion in https://github.com/cgeo/cgeo/issues/8699#issuecomment-666677719

Adds a button labelled "default" if the preference has an `app:defaultColor` entry in `preference.xml`:
image

On pressing the button the color picker returns to the default color (but does not store it yet - you can still decide to cancel).

### DIFF
--- a/main/res/values/attrs.xml
+++ b/main/res/values/attrs.xml
@@ -48,6 +48,7 @@
     <!-- attributes for ColorpickerPreference -->
     <declare-styleable name="ColorpickerPreference">
         <attr name="showOpaquenessSlider" format="boolean" />
+        <attr name="defaultColor" format="color" />
     </declare-styleable>
 
     <!-- attributes for SeekbarPreference -->
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1799,6 +1799,7 @@
     <string name="preference_colorpicker_select_color_scheme">Select color scheme</string>
     <string name="preference_colorpicker_select_color">Select color</string>
     <string name="preference_colorpicker_select_opaqueness">Select opaqueness</string>
+    <string name="reset_to_default">Default</string>
 
     <!-- number input -->
     <string name="number_input_title">Enter a value between %1$s and %2$s</string>
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -705,8 +705,8 @@
                 android:title="@string/init_trailcolor"
                 android:summary="@string/init_trailcolor_summary"
                 android:key="@string/pref_mapline_trailcolor"
-                android:defaultValue="@color/default_trailcolor"
                 android:dependency="@string/pref_maptrail"
+                app:defaultColor="@color/default_trailcolor"
                 app:showOpaquenessSlider="true" />
             <Preference
                 android:title="@string/init_trailwidth"
@@ -725,7 +725,7 @@
                 android:title="@string/init_directioncolor"
                 android:summary="@string/init_directioncolor_summary"
                 android:key="@string/pref_mapline_directioncolor"
-                android:defaultValue="@color/default_directioncolor"
+                app:defaultColor="@color/default_directioncolor"
                 app:showOpaquenessSlider="true" />
             <Preference
                 android:title="@string/init_directionwidth"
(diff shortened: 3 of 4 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

MagpieFourtyTwo:
Was pretty astonished to notice that e. g. History line's ``Default`` button reverted Navigation line's settings, too.
Was this really intended?  If so, I would suggest to limit the reset to the section, where it has been pressed.

moving-bits:
> Was pretty astonished to notice that e. g. History line's `Default` button reverted Navigation line's settings, too.
> Was this really intended? If so, I would suggest to limit the reset to the section, where it has been pressed.

That's definitely not intended - will look into it.

moving-bits:
Hm, cannot reproduce it quickly. I tried the following:
- fire up the emulator with current master version (API level Android 7 in this case)
- set the color as follows:
  - history track to red, full opaqueness
  - direction line to yellow, full opaqueness
  - individual route to green, full opaqueness
  - individual track to blue, full opaqueness
  - then re-enter individual track color settings, use "default" and "ok"
- return to map view - result looks ok to me:

image

- Also, after returning to map settings, colors shown in settings are fine.

Do you have more info on how to reproduce it?

MagpieFourtyTwo:
Hmm, seems as if things are more complex (hello Murphy), as a simple [DEFAULT] really seems to do nothing wrong.
Nevertheless there is this magic change of colors - and in fact the [CANCEL] button is somehow involved ...
And although it was a little bit tricky to deterministically reproduce the case, as it seems to behave differently every time I try, but eventually I found this route, where it at least "worked" most of the times:

1. set "History line Color" to s/st different from default
   (over here: scheme: [leftmost], color: [top rightmost], opaqueness: 128; line width: 2)
   => [OK]
2. set "Navigation line Color" to s/th different from default
   (over here: scheme: [leftmost], color: [top rightmost], opaqueness: 128; line width: 4)
   => [OK]
3. leave settings, display map ... or whatever
4. re-enter settings and revert "History line Color" to [DEFAULT] => [OK]
5. open "Navigation line Color" picker => and press [CANCEL] (!!!)
6. re-open "Navigation line Color" picker ... and notice that it appears to be changed to ... default?
7. again just click [CANCEL] ... and notice that in the menu item list it's color still appears to be like set in step 2.
8. leave settings, display map ... or whatever
9. re-enter settings and set "History line Color" to the very same color from step 1 => [OK]
10 and *NOW* have a look at the color of "Navigation line Color" in the menu list ... it finally changed completely to the default.

Pls. let me know if you can reproduce it - if not, I will look for an even more deterministic way ...  ;)

MagpieFourtyTwo:
Sometimes it's even sufficient to just open the Navigation line Color picker, click [CANCEL] and then reopen the picker again, to get the default value again - but I was not able to get a clear route to this behavior ... :/
Seems as if this is "possible" after previously having set the History line Color ... but I'm not sure ...

MagpieFourtyTwo:
Hmm, possibly it's just a problem with Navigation line's color picker in conjunction with the [CANCEL] button ...? As I right now just open it's picker now and then, and sometimes it shows the default color without previously having changed anything ... only "constant" in this scenery seems to be the usage of the [CANCEL] button ...

Just opened the settings for good, everything fine, press [CANCEL] ... re-open the picker and \*boom* ... default again.

MagpieFourtyTwo:
Ok, new scenario - shorter, but possibly not that deterministic - just have at try:

1. set "Navigation line Color" to the values mentioned above
2. leave settings, close c:geo and remove it from memory to get a clear base
3. open "Navigation line Color" picker and press [CANCEL]
4. re-open "Navigation line Color" picker => default color is shown.

This at least while my very last three tries ...

MagpieFourtyTwo:
Even more strange if you ...
5. press [CANCEL] ... (individual color will be shown)
6. leave  settings, close c:geo and remove it from memory to get a clear base
7. open map settings and have a look at the color of Navigation line ... it's finally set to default anyway, although we "said" [CANCEL] in step 5

moving-bits:
> Hmm, seems as if things are more complex (hello Murphy), as a simple [DEFAULT] really seems to do nothing wrong.
> [...]

Wow, thanks for the elaborate description!

> Pls. let me know if you can reproduce it - if not, I will look for an even more deterministic way ... ;)

I was able to reproduce it except the last step. But that still includes the wrong behavior in step 6:

> re-open "Navigation line Color" picker ... and notice that it appears to be changed to ... default?

I'll debug into it to see what's happening here...

moving-bits:
Maybe I have found the cause for this: Looks like we have a wrong preset of the `originalColor` value (which is being referred to on pressing "cancel"), therefore I now do set it manually on startup. Let's see how #8743 works.
@MagpieFourtyTwo Thanks for your intensive testing and the detailed description - has been very helpful!

PR #8724, opened 2020-08-01
add map download button to map settings

moving-bits:
Adds a button to map settings to start the new map downloader.

image

After pressing the button a short explanation and warning message is shown:

image

On pressing the "Choose map" button the new map downloader starts, following the same procedure as if started from map selector in map view. (For details on this see #8610.)


moving-bits:
retest this please

Issue #8727, opened 2020-08-01
Individual route: add export / import functions


moving-bits:
I'm pretty regularly using the individual route functions, especially for planning a tour. It would be very helpful if a carefully crafted individual route could be stored as a GPX file (export) and reloaded at a later time (import).

Export:
- Export file should use the standard GPX route format.
- Setting the name of a route point to the waypoint's geocode would allow for reattaching the data to c:geo's internal data structures.
- File name is "route_" (date/time) ".gpx". File is stored in GPX export dir. Optional: ask for file name infix (instead of date/time).

Import:
- Search for files named "route_*.gpx" in gpx dir using the internal file chooser.
- Ask before overwriting the current individual route (if any).
- Try to reattach the waypoints by using the name of the route points.
- If the waypoint cannot be reattached during import, c:geo should use the stored coordinates for this waypoint instead.

moving-bits:
fixed with #8730 and #8744

PR #8730, opened 2020-08-02
export individual route as GPX

moving-bits:
First part for #8727: Export function for individual route
(Individual route gets exported as a regular GPX route file, but follows the convention to use the geocode as the route point's name to be able to import it accordingly later. Import is not part of this PR.)


moving-bits:
Note: Test error due to geokrety.org currently being offline: `Service unavailable - DB ERROR GKIE_5f266b95279e9 - Unable to join database server`

moving-bits:
retest this please

PR #8731, opened 2020-08-02
Issue #8729: fix

### LINKED ISSUE #8729: Cannot store waypoints into empty personal note
**Describe the bug:**
Cannot store wapoints into empty personal note - will receive an error message "Waypoints could not be stored in note due to size limits":

image

**To Reproduce:**
- open a cache (GC2BG6B in my case)
- go to waypoint page
- create a new waypoint (it's the only one in this case)
- return to cache details page and press "Store Waypoints"

Error toast "Waypoints could not be stored in note due to size limits"

Waypoints being stored into note as described in #8608

**Version of c:geo used:**
Today's master including #8608

Yes, have also tried with different waypoint types and numbers.

@eddiemuc 
Can you look into this?

eddiemuc:
this PR contains code which SHOULD fix #8729 (although I couldn't reproduce it I think this is caused by a not correctly initialized maxPersonalNoteSize variable), and in case it doesn't it contains code helping to analyze situation better (now the max size of personal note is added to the error toast)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -902,8 +902,8 @@
     <string name="cache_personal_note_edit">Edit</string>
     <string name="cache_personal_note_storewaypoints">Store Waypoints</string>
     <string name="cache_personal_note_storewaypoints_success">Waypoints were added to personal note</string>
-    <string name="cache_personal_note_storewaypoints_success_limited">Waypoints were added to note but had to be shortened</string>
-    <string name="cache_personal_note_storewaypoints_failed">Waypoints could not be stored in note due to size limits</string>
+    <string name="cache_personal_note_storewaypoints_success_limited">Waypoints were added to note but had to be shortened (max personal note size: %1$d)</string>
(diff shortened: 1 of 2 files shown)


moving-bits:
Works for me in the scenarios described in #8729. Thanks for fixing!

moving-bits:
@eddiemuc 
Looking at the changes: Could you remove the `onDowngrade` part, please? It's not related to fixing issue #8729, and might confuse us keeping track of things.

(BTW: I do not share the assumption that `onDowngrade` gets only called in dev situations IMHO, although this is probably the most prominent part. We do also have a lot of users working with our nightlies, and there are situations where they need to downgrade. DB changes are mostly not downward compatible, which would lead to strange errors. Devs and testers might remember this, but if a "common user" accidentally installs an earlier version we might get stuck.)

eddiemuc:
removed the `onDowngrade`. This is indeed a different topic, including it here was my mistake

moving-bits:
> removed the `onDowngrade`. This is indeed a different topic, including it here was my mistake

no worries, and thanks for fixing

moving-bits:
retest this please

PR #8735, opened 2020-08-02
[Basics, related to Issue:#2553] adding DataStoreUtils

eddiemuc:
While working on Issue #2553 (Offline Log with image upload doesn't save image to log) I realized at some point that I need some more fundamental changes in order to implement this in a future-proof way. This became much too big for one Pull Request. Thus the idea is to push parts of this as independent pull requests and glueing the code pieces together once they are through. This also allows separate discussion with regards of different things I suggest to bring to the code.
This is one of those Pull Request.

#2553 requires the creation of 2 new tables (offline log images and offline log trackabe action settings) and modifying an existing one. I had lots of problems keeping all thing in sync using pure sql approach of DataStore. Thus I created the DataStoreUtils which I propose for merge. If merged, this will be used heavily in #2553

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStoreUtils.java
+++ b/main/src/cgeo/geocaching/storage/DataStoreUtils.java
@@ -1,0 +1,377 @@
+package cgeo.geocaching.storage;
+
+import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.functions.Func1;
+
+import android.content.ContentValues;
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+import android.database.sqlite.SQLiteException;
+
+import androidx.annotation.NonNull;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.commons.lang3.StringUtils;
+
+public final class DataStoreUtils {
+
+    private DataStoreUtils() {
+        //No instance of this class
+    }
+
+    /** Enum for DB types used */
+    public enum DBType {
+        INTEGER, TEXT, FLOAT, DOUBLE, LONG
+    }
+
+    /**
+     * Represents a single DB table. Meant for instantiating as a constant. Provides helper methods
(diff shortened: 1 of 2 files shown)


eddiemuc:
With this PR I am also requesting to change Codacy Util-Class name pattern from `[A-Z][a-zA-Z0-9]+(Utils?|Helper|Constants)` to `[A-Z][a-zA-Z0-9]+(Utils?|Helper|Constants|Table)`

moving-bits:
I'm unsure about what you're trying to accomplish here, so please forgive my questions...
Can you explain a bit how this relates to the existing db routines and the existing db upgrade mechanism?
And is this meant to become a complete rewrite of our db layer at some point in time, or is it an extension?
IIRC we also had some discussion some time ago whether we want to switch to a Android Room based storage. How would that fit in? May this be an alternative?

eddiemuc:
HI @moving-bits , Android Room or similar framework would IMHO be the far better alternative. If you are already considering this, then I will pull back my Pull Request immediately
From the code in DataStore, my impression was that for some reason you don't want to pull in a database abstraction framework into c:geo, because currently everything is mainly pure handwritten SQL there.

moving-bits:
> HI @moving-bits , Android Room or similar framework would IMHO be the far better alternative. If you are already considering this, then I will pull back my Pull Request immediately

Sadly, we don't have an ready-to-go implementation for it. IIRC @jonas-koeritz some time ago started with it, but I don't know the current state. Maybe you two could pair up with this task?

> From the code in DataStore, my impression was that for some reason you don't want to pull in a database abstraction framework into c:geo, because currently everything is mainly pure handwritten SQL there.

Don't think so. It's probably more due to the code base's age - the roots of c:geo are somewhere in 2011 (long before my own active time with c:geo), and nobody has found the time (or had the knowledge) to bring this part up-to-date.

jonas-koeritz:
I've started many months ago and currently stopped in anticipation for updated documentation from Google regarding LiveData  and Room.

eddiemuc:
In any case, I will pull back this PR

Issue #8670, opened 2020-07-24
Add "Delete via point" button to "Individual Route" => "Sort"


MagpieFourtyTwo:
Would be great to be able to [Trash] a via point from the list of via points in "Individual Route" => "Sort".

moving-bits:
Yeah, I'm sort of longing for that myself. I've only been reluctant so far as to where to put that third button. If I put it in the same row as the sorting buttons, screen space for name etc. would get reduced significantly. Not sure, if that would work for all devices. (Other than that probably the easiest to use solution.) 
OTOH I would not like to open a second row of buttons for that, as this would produce a lot of empty space and require much more scrolling.
What do you think? / What would be your favorite solution?

MagpieFourtyTwo:
What about just swiping it away (system conform to the left) ...?
Possibly there even can appear a trash can like with Android's notifications?

moving-bits:
Closed with https://github.com/cgeo/cgeo/commit/fd0728343ceaaee7ecfc5288ae540df52f9e4878

MagpieFourtyTwo:
Well done, works ... although I did not expect that I would have to additionally tap the trash bin.
Android 10's "delete notification swype" is different: You (really) swype the notification to the left, it in fact goes to the left as you do, then a trash appears - if you stop now, you can tap the can, but if you simply continue to swype, it will vanish immediately.
But, this is cosmetic, the real problem is:  It works only if you manage to do the horizontal swype *extremely* exact ...
And to be honest: I get it working only once out of ten.  :/
(BTW; I even have problems sometimes, to get the select mode activated by swyping to the right in a cache list)

So, I would prefer if the left swype will be accepted no matter how exact it is,
In fact a "grab" of the entry => If it goes to the left => delete and, if possible: If it goes up or down, this may be a quick sort move ...?  ;)

moving-bits:
Right now I have implemented the same "swipe" behavior as we have in cache list - having no animation, but enabling/disabling some additional UI element (select box there, delete button here). Without animation it's hard to recognize that you're going to do a significant operation like deleting an element, therefore I decided to go with the delete button first. (There's no built-it swipe functionality (at least AFAIK), so you either have to add another library for that, or to hard-code it (which I would like to avoid).)

Regarding the "have to manage to do the horizontal swipe extremely exact": This is something we can help with :-) as we check the vertical deviation. I could set this to be a bit more tolerant.

MagpieFourtyTwo:
[...]

PR #8737, opened 2020-08-02
add wrapping to title in route sort view

moving-bits:
Follow-up to the discussion in https://github.com/cgeo/cgeo/issues/8670#issuecomment-667132652:

> Unfortunately, a lot of these have looong names and, even worse, with the significant part of the name at the very end (in form of a counter), which makes it really hard to say, which one is which ... so what would you think about a line wrap for cache names that do not fit into the column?

This PR adds wrapping to the title, if needed:

image

### DIFF
--- a/main/res/layout/twotexts_twobuttons_item.xml
+++ b/main/res/layout/twotexts_twobuttons_item.xml
@@ -18,17 +18,13 @@
         android:layout_marginRight="110dip"
         android:layout_alignParentLeft="true"
         android:layout_alignParentTop="true"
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

Lineflyer:
@moving-bits  This change has been merged upstream to `master` now (without conflicts).
Feel free to do this yourself after pushing something to `release` so that we avoid having conflicts later if a commit is not merged upstream on short term.

PR #8738, opened 2020-08-02
remove unused translations

moving-bits:
Remove a couple of translation string which are no longer in use, but haven't been removed automatically by crowdin update yet. This is not a problem per se, and removing them does not change the resulting apk, but erases some warnings during compilation.


moving-bits:
retest this please

moving-bits:
retest this please

PR #8740, opened 2020-08-02
remove duplicated calls to clearIndividualRoute

moving-bits:
PR #8732 should have gone to release instead of master. Try to cherry-pick that commit and add it to release, as it is a prerequisite for #8739.

@Lineflyer 
Will that work? Or will we get a branch conflict anyway?
Or do I need to do a revert on #8732 on master first, then commit this PR to release?

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -896,9 +896,6 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
                 new TrailHistoryExport(activity, this::clearTrailHistory);
                 return true;
             }
-            case R.id.menu_clear_individual_route:
-                clearIndividualRoute();
-                return true;
             case R.id.menu_hint:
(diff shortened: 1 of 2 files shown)


Lineflyer:
You should revert and submit against release. 
But I dont consider it as such important, so why not leave it on master only?

moving-bits:
> But I dont consider it as such important, so why not leave it on master only?

It's not about this PR itself, but another PR depending on it:

I would like to add a confirmation dialog to the "clear individual route", as it has a different position in menu now, directly beneath the "sort individual route" menu entry. To avoid accidentally clearing your route, a confirmation could help.

Problem is, that the two calls I'm trying to remove with this PR (located in `NewMap` and `CGeoMap`) are hiding the confirmation dialog which is located in `IndividualRouteUtils` - they are a leftover which I should have deleted some time ago.

Lineflyer:
> > But I dont consider it as such important, so why not leave it on master only?
> It's not about this PR itself, but another PR depending on it:

I understood this in first place. 
But why not doing all this on master?

PR #8748, opened 2020-08-04
Rename List "Internal caches" to "User Defined Caches"

### LINKED ISSUE #8685: List "Internal caches" should be named "User Defined Caches"
Spin off from #8678 

As also the caches are named like that, we should name their default list (which is the default list where the "Go to" targets are saved to) accordingly "User defined caches"

fm-sys:
fix #8685 as discussed in https://github.com/cgeo/cgeo/issues/8685#issuecomment-668209792


main/res/values/strings.xml
```diff
@@ -1076,7 +1076,7 @@
     <string name="internal_goto_targets_description">This cache stores your recent \"Go to\" targets. It is automatically created once you have used the \"Go to\" function in the main menu. The coordinates entered there will be stored into waypoints of this cache. If you do no longer need these coordinates you can safely delete this cache or any of its waypoints. Also you can move the cache to any other list of your choice.</string>
     <string name="init_longtap_map">Long tap on map</string>
     <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache)</string>
-    <string name="goto_targets_list">Internal caches</string>
+    <string name="goto_targets_list">User defined caches</string>
```
UniQP, line 1079:
```suggestion
    <string name="goto_targets_list">User-defined caches</string>
```

fm-sys: Yes, you're right...

Issue #8761, opened 2020-08-06
Fatal exception of logging


moving-bits:
While playing around a bit with the new LoggingManager I constantly receive the following error already on the first step:

```
2020-08-06 22:36:24.092 31987-31987/cgeo.geocaching E/AndroidRuntime: FATAL EXCEPTION: main
    Process: cgeo.geocaching, PID: 31987
    java.lang.RuntimeException: Unable to start activity ComponentInfo{cgeo.geocaching/cgeo.geocaching.log.LogCacheActivity}: java.lang.IllegalArgumentException: Object returned from onCreateLoader must not be null
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3270)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3409)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2016)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.IllegalArgumentException: Object returned from onCreateLoader must not be null
        at androidx.loader.app.LoaderManagerImpl.createAndInstallLoader(LoaderManagerImpl.java:385)
        at androidx.loader.app.LoaderManagerImpl.initLoader(LoaderManagerImpl.java:421)
        at cgeo.geocaching.connector.gc.GCLoggingManager.init(GCLoggingManager.java:127)
        at cgeo.geocaching.log.LogCacheActivity.onCreate(LogCacheActivity.java:287)
        at android.app.Activity.performCreate(Activity.java:7802)
        at android.app.Activity.performCreate(Activity.java:7791)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1299)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3245)
[...]

PR #8762, opened 2020-08-07
[Issue:#8761] fix 'Fata exception of logging'

eddiemuc:
'onCreateLoader' in GCLoggingManager usd to return null when no GC credentials are stored, this is apparently not allowed.
Now the Loader is returned even if no GC credentials are available. The activity logic will then handle this as if it were a "normal" network connectivity failure


eddiemuc:
retest this please

fm-sys:
retest this please

moving-bits:
Does returning a failing loader instead of `null` have any side effects, e. g. having to wait for timeouts until program continues, or else?

eddiemuc:
The returned loader runs asynchronous, no UI blockage will occur.
The effect to the user is the same as if the server can't bei reached e.g. due to no internet connectivity. I tried it out. After a timeout period (during which user is not blocked), a toast appears saying that data could not be loaded.

moving-bits:
Ah, I see, thanks for explaining. And thanks for the contribution!

PR #8772, opened 2020-08-09
colorAccent in cache list selector

### LINKED ISSUE #8769: Wrong colorAccent in cache list selector
List selector still uses the old blue `colorAccent` - style/theme needs to be adjusted to use the new orange `colorAccent`.

See:
- go to a cache detail page
- click on "pencil" icon to choose lists
- check a check box

alternatively:
- go to a map
- tap on a cache and select the "save" button

fm-sys:
fix #8769


moving-bits:
@fm-sys 
Thanks for fixing this issue!

PR #8776, opened 2020-08-10
Include offline finds in total finds on home screen

### LINKED ISSUE #4900: Include offline finds in total finds on home screen
As you might have noticed by now I am a die hard offline user of c:geo.
That is one of the reasons I want issue #4334 to be dealt with.
I definitely see it as a bug rather than a feature by design.

Anyhow...

I have another feature request that might help us off-liners to keep count of the total finds in another way.

Today it says like this on the home screen:
geocaching.com • MrYaz (2233) • Login OK

What if the offline founds would be listed as well like this:
geocaching.com • MrYaz (2233+48) • Login OK
...where 2233 are the online founds and 48 are the offline founds

Or even:
geocaching.com • MrYaz (2233+48=2281) • Login OK
...to help doing the math for you.

If you don't make any offline logs the found count are listed exactly like it is today.

fm-sys:
fix #4900

### DIFF
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -147,14 +147,28 @@ public class MainActivity extends AbstractActionBarActivity {
                     private void fillView(final TextView connectorInfo, final ILogin conn) {
                         final StringBuilder userInfo = new StringBuilder(conn.getNameAbbreviated()).append(Formatter.SEPARATOR);
                         if (conn.isLoggedIn()) {
-                            userInfo.append(conn.getUserName());
+                            userInfo.append(conn.getUserName()).append(" ");
                             if (conn.getCachesFound() >= 0) {
-                                userInfo.append(" (").append(conn.getCachesFound()).append(')');
                                 FoundNumCounter.updateFoundNum(conn.getName(), conn.getCachesFound());
                             }
-                            userInfo.append(Formatter.SEPARATOR);
                             activity.checkLoggedIn();
                         }
+
+                        final FoundNumCounter f = FoundNumCounter.load(conn.getName());
+                        if (f != null) {
+
+                            userInfo.append("(").append(f.getCounter(false));
+
+                            final int offlinefounds = DataStore.getFoundsOffline(conn.getName());
+                            if (offlinefounds > 0) {
+                                userInfo.append('+').append(offlinefounds);
+                            }
+                            userInfo.append(')').append(Formatter.SEPARATOR);
+
+                        } else if (conn.isLoggedIn()) {
+                            userInfo.append(Formatter.SEPARATOR);
+                        }
+
                         userInfo.append(conn.getLoginStatusString());
 
                         connectorInfo.setText(userInfo);
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -157,11 +157,11 @@ public class MainActivity extends AbstractActionBarActivity {
                         final FoundNumCounter f = FoundNumCounter.load(conn.getName());
                         if (f != null) {
 
-                            userInfo.append("(").append(f.getCounter(false));
+                            userInfo.append('(').append(f.getCounter(false));
 
                             final int offlinefounds = DataStore.getFoundsOffline(conn.getName());
                             if (offlinefounds > 0) {
-                                userInfo.append('+').append(offlinefounds);
+                                userInfo.append(" + ").append(offlinefounds);
                             }
                             userInfo.append(')').append(Formatter.SEPARATOR);
 
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -2828,7 +2828,7 @@ public class DataStore {
         int counter = 0;
 
         try {
-            final Cursor cursor = database.rawQuery("SELECT geocode FROM " + dbTableLogsOffline + " WHERE type = ? OR type = ? OR type = ?",
+            final Cursor cursor = database.rawQuery("SELECT geocode FROM " + dbTableLogsOffline + " WHERE geocode IN (SELECT geocode FROM " + dbTableCaches + ") AND (type = ? OR type = ? OR type = ?)",
                     new String[]{LogType.FOUND_IT.iconName, LogType.ATTENDED.iconName, LogType.WEBCAM_PHOTO_TAKEN.iconName});
             final Set<String> geocodes = cursorToColl(cursor, new HashSet<>(), GET_STRING_0);
 


moving-bits:
Nice! Just one little thing: Can you add spaces around the "+" sign when both online and offline log count is available? Makes it more readable.

moving-bits:
BTW: How is the offline found counter calculated?
I have an imported database having caches from geocaching.com and opencaching.de. c:geo shows "4+228" and "85+1" as found counter, but when I go to "all caches" list (> 6.000 caches) and filter "With personal data => With offline log" it lists only two caches at all. Strange.

fm-sys:
> 
> 
> BTW: How is the offline found counter calculated?
> I have an imported database having caches from geocaching.com and opencaching.de. c:geo shows "4+228" and "85+1" as found counter, but when I go to "all caches" list (> 6.000 caches) and filter "With personal data => With offline log" it lists only two caches at all. Strange.

This is the function I implemented for the new number counter, which is also used here... Mabe you could take a look if there is something wrong?

```
    // get number of offline founds for a specific connector
    public static int getFoundsOffline (final String serviceName) {
        int counter = 0;

        try {
            final Cursor cursor = database.rawQuery("SELECT geocode FROM " + dbTableLogsOffline + " WHERE type = ? OR type = ? OR type = ?",
                    new String[]{LogType.FOUND_IT.iconName, LogType.ATTENDED.iconName, LogType.WEBCAM_PHOTO_TAKEN.iconName});
            final Set<String> geocodes = cursorToColl(cursor, new HashSet<>(), GET_STRING_0);

            for (String geocode : geocodes) {
                if (ConnectorFactory.getConnector(geocode).getName() == serviceName) {
                    counter ++;
                }
            }
        } catch (final Exception e) {
            Log.e("DataStore.getFoundsOffline", e);
        }

        return counter;
    }

moving-bits:
Seems like the "offline log" filter is wrong (or I do not understand it), as by inspecting the database in question I can find the required amount of offline logs, and the caches being referred to are available in the db as well. I will open a separate issue for it.

Lineflyer:
If the PR is considered OK codewise, feel free to merge. As I log offline all the time I go caching I will test it in the nightlies anyway. If @moving-bits prefers to clarify #8781 before that, we could also wait a bit.

moving-bits:
Let's add the spaces around the "+" sign first, and maybe resolve #8781 (which to me, after @Lineflyer's comment, looks like the db query used in this PR needs to be extended to exclude caches not stored in any list).

fm-sys:
I seem to remember copying the query from somewhere and adapting it for my needs. In the original, the query was checking whether the cache is also stored in the database. This seemed unnecessary and I removed it for my function.

I will try to fix with insert this again, let's see if it works

fm-sys:
> which to me, after @Lineflyer's comment, looks like the db query used in this PR needs to be extended to exclude caches not stored in any list

@moving-bits could you please check if the displayed offline founds number is correct now?

moving-bits:
> > which to me, after @Lineflyer's comment, looks like the db query used in this PR needs to be extended to exclude caches not stored in any list
> @moving-bits could you please check if the displayed offline founds number is correct now?

Just tested it - it still displays all caches, even those no longer visible in other places of the app.

According to @Lineflyer's comment you need to check for offline logs, if the referred caches is stored in at least one list (by using something like `SELECT ... FROM cg_logs_offline INNER JOIN cg_caches_lists USING (geocode)` - but use the placeholder variables for table names instead).

Lineflyer:
> According to @Lineflyer's comment you need to check for offline logs, if the referred caches is stored in at least one list (by using something like SELECT ... FROM cg_logs_offline INNER JOIN cg_caches_lists USING (geocode) - but use the placeholder variables for table names instead).

Should we also clarify where these ghost offline logs come from?
Does your database provide any clue about their origin and state? Did you check your history list, whether they show up there?
Or try to save one of them again to see if the offline log is then associated with it?

fm-sys:
> you need to check for offline logs, if the referred caches is stored in at least one list (by using something like SELECT ... FROM cg_logs_offline INNER JOIN cg_caches_lists USING (geocode)

I don't think thats a good idea, because it do not include the history list. You could deleate a caches from the "Stored" list (doesn't work with "All Caches"!), and it will disappear also from the "All caches" list as well. But the cache will stay in the history list, if it has an offline log. 

Screenshot (Aug 12, 2020 6 04 21 PM)

@moving-bits So could you please recheck your **"history" list**, whether they show up there? (Also with filter "With personal data => With offline log")

Maybe we should merge it despite this error. The new [NUMBER] template is also merged to master, which uses the same function. Or provide me your database please, because I cannot reproduce it on my side.

moving-bits:
> @moving-bits So could you please recheck your **"history" list**, whether they show up there? (Also with filter "With personal data => With offline log")

I've never used the history list in all those years, didn't have this on my mind.

You are right, those caches show up on the history list, and if I filter the *history list* to "contains offline logs", the counter seems to be ok.

So sorry for the confusion! And thanks for pointing me into the right direction.

MagpieFourtyTwo:
Just stumbled over the new counter, wondering what this "+ x" could be ... 
and *immediately* got the urge request to *please* make this configurable,
as being confronted with the open number of logs still to do on every single start of the app
leads to massive frustration for some people - even ending up in giving up caching at all. :/

Lineflyer:
@MagpieFourtyTwo 
Do you really think this would be the case? I think it would be rather more helpful for all users to see whats on the ToDo and whats their current real find count.
Anyway, please br reminded to better open a new issue refering to here instead of commenting merged/cloed issues.
So feel free to open up a new one to discuss this :)

MagpieFourtyTwo:
Yes, most definitely.
In this very moment one of them is sitting right by my side.  :) or :/ ...?

MagpieFourtyTwo:
Moreover I know at least three cachers, who stopped logging completely, just writing offline logs or notes in c:geo sometimes, cause they feel max pressure from logging. There are in fact a lot of people who do not like writing (or possibly are not really capable of) and therefore logging in public is a big deal for them. And they for sure do not need an app that confronts them every single day with this lack of capabilities ...

fm-sys:
> Moreover I know at least three cachers, who stopped logging completely, just writing offline logs or notes in c:geo sometimes, cause they feel max pressure from logging

we could add an "= x", so they can be proud of how many caches they already have found

MagpieFourtyTwo:
Good idea, but most regretfully misses my point completely, as there still will be the pressure by reminding them of having to do this f***i^W unbeloved log task ... ;)

MagpieFourtyTwo:
BTW: Esp. these do not cache for numbers, but for being outside and in nature, so a high number gives nothing to them, anyway.

Lineflyer:
Lets continue in the dedicated issue please.

alan666notb:
I've just been reading some FB posts about this...

-- people have no idea what it means, most common guess is 'Adventure lab finds'

-- Some people have a lot, how do they resolve that?

MagpieFourtyTwo:
> Adventure lab finds

AdvLab == Pandora's box.
Completely different story.
Ok, admitted, I may perhaps not be completely objective, as I don't like them that much, but AdvLabs don't fit into anything at all.
And AFAIK there's absolutely *no* way, to integrate them into "traditional" gc.com structures, no possibility to find, manage, log or even display them in a gc map or list other than using the AdvLab app. They don't even integrate into traditional statistics. And therefore I have absolutely no clue, why gc is pushing them so hard. Esp. as there are WIGs, which do pretty much the same. But instead of fixing the WIG issues, they just invent s/th new, with even more problems ...

Nevertheless, IMHO AdvLabs are a completely different story, so I would leave them out of this topic completely, otherwise we ill not even get the most of the "normal" (and realizable) possibilities.

With regard to the AdvLab related questions of the people: I don't see us in the role of gc.com's "explainer", we should not pay for what they have omitted to communicate to the caching community - thus I would most clearly block any request and tell them how it is, i. e. that the app is the only way, an isle game, not integratable into nothing at all. period. And they should complain against gc.com and should not blame c:geo for what gc.com has messed up.

MagpieFourtyTwo:
> people have no idea what it means

So, to get back to our topic over here; What do you exactly mean with "it"?

fm-sys:
> What do you exactly mean with "it"?
 
Probably targeting the "+1" counter introduced with this PR. 

I added the explain message some days ago (see PR #8960 and @alan666notb your issue #8938) which will be appear in the second beta...

So I don't really now, what you mean with your comment https://github.com/cgeo/cgeo/pull/8776#issuecomment-694059214 ...

MagpieFourtyTwo:
> > What do you exactly mean with "it"?

> Probably targeting the "+1" counter introduced with this PR.

Yes, of course - but *what* do they not understand?  Would be valuable to know more about this "it", in order to know, if this be properly explained by the popup (whereas I would have written more detailed, *which* ``+1`` the popup refers to) .

alan666notb:
> > > What do you exactly mean with "it"?
> 
> > Probably targeting the "+1" counter introduced with this PR.
> 
> Yes, of course - but _what_ do they not understand? Would be valuable to know more about this "it", in order to know, if this will be properly explained by the popup (whereas I would have written more detailed, _which_ `+1` the popup refers to) .

Remember -  I'm just relaying the reactions I've seen posted in FB groups.
They see (12345 + 22) when they used to see (12345), and they don't know what it ( + 22) means.
I wouldn't have either, but I read the release note, I'm guessing most users don't.
Like me, people didn't even know they had offline logs, or how to find them.

fm-sys:
Could you please check if the message in PR #8960 explains everything clearly? (there is an image showing this message attached)
If not, please surgest changes in a separate issue. That makes discussion and fixing easier...

Thanks for your help!

alan666notb:
> Could you please check if the message in PR #8960 explains everything clearly? (there is an image showing this message attached)
> If not, please surgest changes in a separate issue. That makes discussion and fixing easier...
> 
> Thanks for your help!

Yes, I think that would solve it.
I've added a suggestion.

PR #8779, opened 2020-08-10
add option to write complete logfile (fix #8777)

### LINKED ISSUE #8777: Implement recording Debug Logs to file (as with Crash Logs)
When it comes to crashes, Debug Logs are (way more than just) pretty helpful. This is where the new function ``Create logfile`` comes into play. Great stuff, but seemingly not helpful, when trying to analyze weird app behavior *with no crash at all*, cause ``Create logfile`` seems to record crash scenarios only. Thus: No crash, no (current) log (at least I was not able to get a usable one while testing for #8632 and #8682).

c:geo should start to build a logcat itself (like it does when clicking ``Create logfile``) as soon as I set ``[X] Activate Debug Log`` and ask for save or share, as soon as I remove the ``[X]`` again.

Had to create the Debug Logs with a third party app, but for Joe Average this is not really a suitable way.

**Additional context**
Maybe a special sequence of setting ``[X] Activate Debug Log`` and pressing ``Create Logfile`` may lead to the same result, but I was in fact not able to get there ... if it was just my mistake, I guess we need a more detailed instruction in the menu ...?

moving-bits:
image

### DIFF
--- a/main/src/cgeo/geocaching/utils/DebugUtils.java
+++ b/main/src/cgeo/geocaching/utils/DebugUtils.java
@@ -43,6 +43,24 @@ public class DebugUtils {
     }
 
     public static void createLogcat(@NonNull final Activity activity) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+            // no differentiation possible on older systems, so no need to ask
+            createLogcatHelper(activity, true);
+        } else {
+            Dialogs.confirmPositiveNegativeNeutral(
+                    activity,
+                    activity.getString(R.string.about_system_write_logcat),
+                    activity.getString(R.string.about_system_write_logcat_type),
+                    activity.getString(R.string.about_system_write_logcat_type_standard),
+                    null,
+                    activity.getString(R.string.about_system_write_logcat_type_extended),
+                    (dialog, which) -> createLogcatHelper(activity, false),
+                    null,
+                    (dialog, which) -> createLogcatHelper(activity, true));
+        }
+    }
+
+    private static void createLogcatHelper(@NonNull final Activity activity, final boolean fullInfo) {
         final AtomicInteger result = new AtomicInteger(LogcatResults.LOGCAT_ERROR.ordinal());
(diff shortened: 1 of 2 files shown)


main/src/cgeo/geocaching/utils/DebugUtils.java
```diff
@@ -52,7 +70,11 @@ public static void createLogcat(@NonNull final Activity activity) {
                 if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
                     builder.command("logcat", "-d", "-f", file.getAbsolutePath());
                 } else {
-                    builder.command("logcat", "-d", "AndroidRuntime:E", "cgeo:D", "cgeo.geocachin:E", "*:S", "-f", file.getAbsolutePath());
+                    if (fullInfo) {
+                        builder.command("logcat", "-d", "*:V", "-f", file.getAbsolutePath());
+                    } else {
+                        builder.command("logcat", "-d", "AndroidRuntime:E", "cgeo:D", "cgeo.geocachin:E", "*:S", "-f", file.getAbsolutePath());
```
Lineflyer, line 76: Shouln`t we add info and warning for c:geo as well?

moving-bits: The `fullInfo` does not exclude anything, the manual logs have the prefix "cgeo", and the system-written messages for c:geo have the package name as prefix (which is "cgeo.geocachin"), so everything c:geo related should be included.

Or have you found log messages prefixed by "c:geo" anywhwere?

moving-bits: I probably misunderstood you - you're not talking about the prefix but about the log level?

Lineflyer: Exactly, I would suggest to add level warning and info for the c:geo package name to the normal logging.

moving-bits: Added info/warning for package related logcat messages

Issue #7587, opened 2019-04-30
Align theme design of GMapsV2 to "rest of app"


Lineflyer:
The GMapsV2 implementation currently still uses a different style (on my Android 8 device) which is different to most other places in c:geo.
Maybe we can use the work on GMapsV2 to align styles/design to what is used e.g. for mapsforge.

I will post some screenshots next weeks to make clear, what I am referring to (e.g. while the map is open it change the color of my status bar (above the app) to light blue and text in top bar is also in this color.

okainov:
Isn't it a duplicate? :) https://github.com/cgeo/cgeo/issues/7470

Lineflyer:
Yes, kind of.
Will close the other one as it might be wise to focus on GMapv2 to fix it rather than on the old stuff, whereas I assume its the same technical wise.

Lineflyer:
Taken from https://github.com/cgeo/cgeo/pull/8782#issuecomment-671879598:

To name just a few things:
- action bar is black instead of grey
- if you open the "three dot menu" it has round edges
- and some menu entries crash into each other (eg.: map settings)
- colorAccent not being applied
- ...

PR #8782, opened 2020-08-10
use themeing for ui.Dialog based dialogs to support colorAccent

moving-bits:
Several dialogs are based on `ui/Dialog.java`, which does not actively use a theme with the new `colorAccent`. This PR changes the dialogs to use the `Alert_Dialog` theme, which includes the new `colorAccent`. Two examples:

image image

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
+++ b/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
@@ -10,7 +10,6 @@ import cgeo.geocaching.utils.functions.Action1;
 
 import android.app.Activity;
 import android.app.AlertDialog;
-import android.app.AlertDialog.Builder;
 import android.app.ProgressDialog;
 import android.content.DialogInterface;
 import android.content.DialogInterface.OnClickListener;
@@ -18,6 +17,7 @@ import android.graphics.drawable.Drawable;
 import android.text.Editable;
 import android.text.InputType;
 import android.text.TextWatcher;
+import android.view.ContextThemeWrapper;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.WindowManager;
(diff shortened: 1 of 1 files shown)


Lineflyer:
Thanks @moving-bits for keeping on to create a consistent UI.
Fun fact BTW:

We had severa users on support appreciating the accents and the new logo...AND...we had a new 1star rating on Google Play with something like "I dont like the orange"...Nevermind :)

Lineflyer:
@moving-bits 
Totally unrelated to this but it just came to my mind (and is AFAIR already documented in some other issue):
While you are on it: Did you see, that GMaps use quite different menu appearance than new map? The GMaps looks different to all other parts of the app.

fm-sys:
> Did you see, that GMaps use quite different menu appearance than new map? The GMaps looks different to all other parts of the app.

Do you mean that black aktion bar at the top or also the buttons for resizing the map? Second is delivered from Google and i'm not sure if we could change it.

moving-bits:
> Fun fact BTW:
> We had severa users on support appreciating the accents and the new logo...AND...we had a new 1star rating on Google Play with something like "I dont like the orange"...Nevermind :)

... yeah, you can't please everyone, or confirm to everyone's taste...
IMHO the new orange fits nicely to the c:geo logo and gives the app sort of a branding.

moving-bits:
> > Did you see, that GMaps use quite different menu appearance than new map? The GMaps looks different to all other parts of the app.
> 
> Do you mean that black aktion bar at the top or also the buttons for resizing the map? Second is delivered from Google and i'm not sure if we could change it AFAIK.

To name just a few things:
- action bar is black instead of grey
- if you open the "three dot menu" it has round edges
- and some menu entries crash into each other (eg.: map settings)
- colorAccent not being applied
- ...

Not that everything of this is bad in itself, but at least it's not consistent.

Lineflyer:
> Not that everything of this is bad in itself, but at least it's not consistent.

Covered in  #7587

PR #8784, opened 2020-08-11
direct logging of PM-only caches from search

### LINKED ISSUE #2288: Allow direct logging of PM-only caches also for basic members
From support mail, I did not verify this but it sounds reasonable:

Groundspeak guidelines generally allow also BM-users to log PM-only caches:

"Basic Members can log finds for Premium Member Only caches when the physical logbook has been signed. Basic Members may have access to PMO caches through various means, including when they geocache with PMs. Groundspeak encourages Cache Owners to allow these logs..."

Currently it is not possible to do this in c:geo because as a BM you are not able to open a PM-cache. However a direct logging possibility could still be offered (e.g. directly from list view or live map popup).

It is possible as a BM to log a PM-cache by using e.g. http://www.geocaching.com/seek/log.aspx?WP=GC2F2VF 
As c:geo already knows the GC-code of the PM-only cache it should be possible for c:geo to post a log in this case.

---

If this information is correct, it could also be used for generally enabling logging without the need to previously open the cache (which is AFAIK the current implementation).

If I currently try to log a cache directly from map it tells me 'c:geo does not have enough information to log...' (or similar).

fm-sys:
fix #2288

it shows folowing dialog when trying to load a PM cache as basic member instead of "c:geo failed to download cache details. This Cache is only available to Geocaching.com Premium Members."

Screenshot (Aug 11, 2020 2 24 02 PM)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -187,6 +187,7 @@
     <string name="err_detail_cache_find_any">c:geo can\'t find any geocaches.</string>
     <string name="err_detail_no_spoiler">c:geo found no spoiler images for this cache.</string>
     <string name="err_detail_still_working">Still working on another task.</string>
+    <string name="err_detail_premium_log_found">This Cache is only available to Geocaching.com Premium Members. Therefore the cache details could not be downloaded. \n\nIf you have already found the cache, you can log it here.</string>
     <string name="err_watchlist_still_managing">Still managing your watchlist.</string>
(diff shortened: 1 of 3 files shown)


Lineflyer:
Like it.
May I suggest the following small change to to the text:
```
This cache is only available to premium members of geocaching.com.
Therefore, the cache information cannot be displayed.

However, if you have already found this cache, you can log it here.
```

moving-bits:
Nice! Thanks for this contribution!

Issue #8786, opened 2020-08-11
Update Playstore Images


fm-sys:
As there where some more or less major UI changes in the last weeks, months and years I would like to update the screenshots shown at Google Play

Some of the UI changes:
- new icon
- new colorAccent
- no more orange circles on live map
- cache details activity - buttons have icons instead of text
- cache details activity - map section is no longer shown
- action bar has no more blue line except main screen, here the color was changed
- and so much more...

Question: In which languages are the screenshots currently available and need to be updated? 
Edit: Looks like they are provided in [German, English and French](https://github.com/cgeo/cgeo/tree/master/main/project/playstore/screenshots)...

And when we're on the go, the [Feature graphic](https://raw.githubusercontent.com/cgeo/cgeo/master/main/project/playstore/icons/cgeo_functional_graphic.jpg) should also be updated as well and could then be integrated also as first (first two) screenshots ...

---
And (if it's OK) I would like to make these pictures a little more modern, like other apps are doing it as well. For example like Threema is doing it:

![Screenshot (Aug 11, 2020 5 37 09 PM)](https://user-images.githubusercontent.com/64581222/89918287-20fd4e80-dbfa-11ea-9b7a-7cf998b3fe7f.png)

Lineflyer:
This would be more than welcome. Great idea.

You can find the existing assets here:
https://github.com/cgeo/cgeo/tree/master/main/project/playstore/screenshots

You can also place your suggestions as a PR in that directories.

fm-sys:
The pictures could look like this...
![c;geo playstore bilder](https://user-images.githubusercontent.com/64581222/89953385-448ebc00-dc2f-11ea-8807-bf4ea2dc2a42.jpg)

What do you think?

Lineflyer:
Looks good in principle. I would tend to use a non transparent background (activate corresponding setting in c:geo) to make it look more neutral.

The file specifications for Google Play are:
JPEG oder 24-Bit-PNG (kein Alpha). Mindestlänge für eine Seite: 320 px. Maximale Länge für eine Seite: 3.840 px.
Minimum 2 Screenshot, maximum 8 screenshots pro Sprache

As it is a too big amount of work to maintain screenshots for all languages, I suggest to strip it down either to use only English or maximum to 3-4 countries with most users. These are Germany, US, Czech Republic, France.

Lineflyer:
While you are on it and if you are able, you could also try to make a new version of the functional graphic:
[...]

PR #8794, opened 2020-08-12
Update functional graphic

fm-sys:
#8786 part 1

cgeo_functional_graphic


fm-sys:
What do you think about it?

Lineflyer:
I would leave out the orange parts.

How about placing our new logo someehere on the left, and leaving text and background unchanged. Of course the new screenshot is needed as in your proposals, as the older was outdated from its content and UI.

fm-sys:
It would look like this...

cgeo_functional_graphic

Lineflyer:
IMHO much better in the second version.
I am however far away from being a designer, but I would place the logo on same height left of the c:geo text.

But let others also provide some feedback first :)

fm-sys:
2020-08-13 14_01_45-PowerPoint-Bildschirmpräsentation  -  c;geo playstore bilder pptx (Benutzerdefiniert)

I'm not sure this looks better...

Lineflyer:
@SammysHP Do you know which font we originally used or with font looks rather similar?
I dont' remember, but I can try to look into the shirt designer.

Lineflyer:
Addon:
Found a resource link for our shirts which also contains two fonts.
@fm-sys Will forward via mail.

SammysHP:
Do you mean the font in the logo? The original font is unknown. Not sure which font I used in my recreation, but I think we can also use a different font.

fm-sys:
| c:geo's font (looks like it's the same as on the logo) | first draft font |
|-|-|
2020-08-13 15_44_59-PowerPoint-Bildschirmpräsentation  -  c;geo playstore bilder pptx (Benutzerdefiniert) | 2020-08-13 14_01_45-PowerPoint-Bildschirmpräsentation  -  c;geo playstore bilder pptx (Benutzerdefiniert)

Lineflyer:
IMHO the left font is the one to use.

fm-sys:
There is one last thing we need to discuss - the font size: 
I would choose the lage font, because the functional graphic is displayed quite small in playstore

| lage font | font size as my pictures before |
|-|-|
Folie7 | Folie6

Lineflyer:
The small font looks a bit lost on the background. I think the larger font is appropriate.

Lineflyer:
BTW: The background is a different one to the old one. The old one has a bite more contrast, thus it would be IMHO better. But I am not sure if we have this graphic somehwere.

Lineflyer:
Another BTW:
We once produced a graphic for a facebook group (not associated with us)....when you are on it and find some time:
grafik

fm-sys:
> The old one has a bite more contrast, thus it would be IMHO better.

I'm afraid I couldn't go along with that. I think it is important to have a certain recognition value and since we have already used this graphic on our website and in c:geo itself, I think it is the only right thing to use the graphic here as well...

Lineflyer:
> I'm afraid I couldn't go along with that. 

Fair enough. Lets stay with your proposal, then everything is aligned.

moving-bits:
Hi @fm-sys, I really like what you're doing here on the graphics side - gives c:geo a more modern touch on its web presences. As for the font: I can look up the font we had used for the t-shirt and send you the font name.

fm-sys:
It already *is* the font from the t-shirts... @Lineflyer had sent them to me...

moving-bits:
@fm-sys 
Is this PR ready to merge from your side?

fm-sys:
Yes, ready to go...

moving-bits:
> Yes, ready to go...

Ok, then let's merge it... Thanks for your contribution!

PR #8803, opened 2020-08-14
remove superflous debug info

moving-bits:
remove two debug messages no longer needed

### DIFF
--- a/main/src/cgeo/geocaching/utils/TrackUtils.java
+++ b/main/src/cgeo/geocaching/utils/TrackUtils.java
@@ -142,14 +142,12 @@ public class TrackUtils {
 
                             latitude += point.getLatitude();
                             longitude += point.getLongitude();
-                            Log.e("lat=" + latitude + ", lon=" + longitude);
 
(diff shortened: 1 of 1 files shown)


Lineflyer:
Just for my understanding:

Some years ago I learned that cgeo will stop on log.e 

Is this still the case?
In that case we should probably avoid log.e in favor of log.d

moving-bits:
> Some years ago I learned that cgeo will stop on log.e
> Is this still the case?

Yes, AFAIK

> In that case we should probably avoid log.e in favor of log.d

I completely agree when it comes to a released version (even nightly) - but...

I use Log.* a lot during development to get some quick info on what's going on. There I prefer Log.e over Log.d/Log.i, as the latter require me to set my filter in Android Studio to a lower level (the message levels are hierarchical). Log window gets overcrowded then, which makes it pretty hard to get the info you need quickly.

So what I would need is something which outputs to log window during development with level "error", but does not output in a release version at all (for the cases I forgot to remove the log statement).

PR #8809, opened 2020-08-15
Update external references, configuration

### LINKED ISSUE #8339: Compile warning since androidannotations-4.7.0
**Describe the bug:**

~~~
Task :main:compileBasicDebugJavaWithJavac
warning: A class activity declared in the AndroidManifest.xml cannot be found in the compile path: [leakcanary.internal.AppWatcherInstaller$MainProcess]
~~~

before with androidannotations-4.6.0:

~~~
Task :main:compileBasicDebugJavaWithJavac
Gradle may disable incremental compilation as the following annotation processors are not incremental: jetified-androidannotations-4.6.0.jar (org.androidannotations:androidannotations:4.6.0)
~~~

bekuno:
- main
  - update dependencies
    - leakcanary to 2.4
    - rxjava to 3.0.5
    - appcompat to 1.2.0
  - add leakcanary:plumber-android, see https://square.github.io/leakcanary/changelog/#plumber-android-is-a-new-artifact-that-fixes-known-android-leaks

- cgeo-contacts
  - update dependencies
    - commons-lang3 to 3.10
	- (let commons-io stay at 2.5)
	- appcompat to 1.2.0

changed:
  - removed leakcanary-object-watcher-android, does not help for #8339
  - remove compileSdkVersion and targetSdkVersion to 29 according for c:geo contacts

### DIFF
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -54,11 +54,14 @@ android {
 
 dependencies {
     // Apache Commons
+    implementation 'org.apache.commons:commons-lang3:3.10'
+    // commons-io 2.6 uses java.nio.file.Path - which is first contained in Android API26
+    // could be replaced by Storage Access Framework
+    //noinspection GradleDependency
     implementation 'commons-io:commons-io:2.5'
-    implementation 'org.apache.commons:commons-lang3:3.5'
 
     // Support Library AppCompat
-    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.appcompat:appcompat:1.2.0'
 
     // Android annotations
     implementation 'androidx.annotation:annotation:1.1.0'
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -284,8 +284,9 @@ dependencies {
     testImplementation 'junit:junit:4.13'
 
     // Leak Canary, memory leak detection
-    def leakCanaryVersion = '2.3'
+    def leakCanaryVersion = '2.4'
     debugImplementation "com.squareup.leakcanary:leakcanary-android:$leakCanaryVersion"
+    implementation "com.squareup.leakcanary:plumber-android:$leakCanaryVersion"
 
     // Locus Maps integration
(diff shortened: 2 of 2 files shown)


bekuno:
fixing #8339 only works on my dev environmant

bekuno:
@moving-bits Ok to set compileSdkVersion and targetSdkVersion to 29 for cgeo-contacts according to c:geo ?

moving-bits:
> @moving-bits Ok to set compileSdkVersion and targetSdkVersion to 29 for cgeo-contacts according to c:geo ?

I have no idea of the internal workings of this plugin, so I can't fully grasp the consequences here.

targetSDK 29 will be a requirement by Play Store beginning November (if they haven't changed this meanwhile), so it will be a must have anyway if we want to do updates to the plugin. - OTOH upgrades to the plugin are only done sporadically, so there's no urge for us to update.

Setting the compileSDK to 29 would have consequences as to what API we can use and how the API behave. Does the contact plugin store data in the filesystem anywhere? Then we would get hit by the "storage framework" problematic as well.

Otherwise no other possible problems come to my mind which could arise from an update of compileSDK / targetSDK.

bekuno:
@moving-bits The plugin does only lookup in the adress book.
The idea of this plugin was to let the necessary permission part of this plugin, not of c:geo.
Nevertheless, I will make a separate PR for the change of compileSdkVersion and targetSdkVersion, so it can be easier to find problems related to this.

Lineflyer:
> The idea of this plugin was to let the necessary permission part of this plugin, not of c:geo.

That reminds me of the issue (which I cannot find at this moment) to reintegrate this function into the main app because the reason for the addon is gone as permissions now need to be accepted at runtime only.

bekuno:
> That reminds me of the issue (which I cannot find at this moment) to reintegrate this function into the main app because the reason for the addon is gone as permissions now need to be accepted at runtime only.

#7193
Raise Target SDK to 26 for contacts plugin (or reintegrate into main app)
We implemented the permission management in the addon

Issue #8783, opened 2020-08-10
Use the new colorAccent in settings


moving-bits:
Settings seems to ignore our new colorAccent:

![image](https://user-images.githubusercontent.com/3754370/89833017-99173600-db60-11ea-8733-6607f950a3f7.png)

Lineflyer:
For me this does not seem to work with todays nightly.

moving-bits:
I had developed and tested this with emulated API 29 device.
What version are you using?

Lineflyer:
Tested on two devices both API 29.
I do still see blue accents in settings.

moving-bits:
It should look like this:

![image](https://user-images.githubusercontent.com/3754370/90328605-e7cc3200-df9d-11ea-89e5-a4afd84bd204.png) ![image](https://user-images.githubusercontent.com/3754370/90328607-ef8bd680-df9d-11ea-8497-2a45c9e91bc7.png)

![image](https://user-images.githubusercontent.com/3754370/90328621-177b3a00-df9e-11ea-8c16-afa670dc80cb.png) ![image](https://user-images.githubusercontent.com/3754370/90328623-1f3ade80-df9e-11ea-9804-3d170aef8ced.png)

Seems like the internal attributes differ between the systems. I'll try something and create a PR for it.

Lineflyer:
I have built a new nightly 2020.08.16NB1-2e91303 to test the newest change from @moving-bits 

On my Samsungs Galaxy S10, Android 10 it looks as follows:
When opening the settings main screen I do already see the old (blue) color in my phone header:
![Screenshot_20200816-220309_cgeo](https://user-images.githubusercontent.com/949669/90342980-e0903d00-e00c-11ea-9ac6-64d5e9547339.jpg)

When entering e.g. the map settings (same in others) it looks blue as well:
![Screenshot_20200816-220317_cgeo](https://user-images.githubusercontent.com/949669/90342986-f1d94980-e00c-11ea-90af-88a514f059fc.jpg)

Just now I recognized (not sure if it was already OK some days ago or only today), that the color picker has the new accents:
![Screenshot_20200816-220325_cgeo](https://user-images.githubusercontent.com/949669/90343000-07e70a00-e00d-11ea-9ed0-0c66e250f908.jpg)

murggel:
I have the blue color-accent in the live-maps-menu, when the google-maps is active. With offline-maps the orange color accent is used (2020.11.22)

Google-Maps
[...]

PR #8811, opened 2020-08-16
apply colorAccent for preferences using AppCompat theme

moving-bits:
Tries to fix https://github.com/cgeo/cgeo/issues/8783#issuecomment-674456981 by adding colorAccent for themes using AppCompat attributes


moving-bits:
@Lineflyer 
Can you give it a try if this extended theme does work for you?

Issue #8457, opened 2020-06-15
Rework storage framework acc. to Android 11 requirements


Lineflyer:
- Install nightly build 2020.06.13-NB-9bbc28e
- Start app and apply permissions for storage and position
- Go to "About c:geo" and check system information tab

Result as expected:
 - `User storage c:geo dir: /storage/emulated/0/cgeo`

- Install nightly build 2020.06.14-NB-e549d24
- Start app and apply permissions for storage and position
- Go to "About c:geo" and check system information tab

Result (wrong):
- `User storage c:geo dir: /data/user/0/cgeo.geocaching`
This fallback should only occur in exceptional cases as this directory cannot be accessed by users (without root)

Remark:
The dir is not changed when upgrading from 2020.06.13 to 2020.06.14, the problem only occurs on installation but not on upgrade

This smells like being related to our change of `targetSDK` I would guess.

moving-bits:
If I'm not mistaken this happens since long time ago. After a fresh install c:geo still points to internal storage, even with permissions granted, and only after completely removing c:geo from RAM (thus forcing a reload with permissions already set) this works. IIRC there is already an issue for this regarding not finding the backup db or so.

MagpieFourtyTwo:
> only after completely removing c:geo from RAM (thus forcing a reload with permissions already set) this works

Not all the time. See my scenario described [here](https://github.com/cgeo/cgeo/issues/8451#issuecomment-644060441) - tried to load a backup several times today, to no avail, although I removed c:geo from memory on every single try.

Although I have to admit (as I realize right now) that I have just removed it from "Recent Apps", but not stopped it via Settings -> Apps. May be this should be done additionally ...?

Lineflyer:
> If I'm not mistaken this happens since long time ago. 

Maybe after first install. But it stays like that even after restart.
I can clearly see a difference between the mentioned versions.

Also normally c:geo always asked to restore existing backups after installation, so it found them.
This is no longer the case now, which is also a followup of this problem.

moving-bits:
> > If I'm not mistaken this happens since long time ago.
> 
> Maybe after first install. But it stays like that even after restart.

Ok, that's new then. Needs to be debugged into.

> Also normally c:geo always asked to restore existing backups after installation, so it found them.

This has not worked reliably for me in the past, at least not in the emulator.

moving-bits:
[...]

PR #8812, opened 2020-08-16
add Android 11 notice about storage spaces

moving-bits:
Android 11 enforces storage spaces for all apps targeted to API 29 (which by itself is an upcoming requirement by Play Store). c:geo does not yet support storage spaces, which will lead to some of its data not being accessible by c:geo if running on Android 11. See #8457 for details.

I propose to add a info notice to our changelog to inform users not to upgrade their system to Android 11 for now (or at least only if they can live with the consequences regarding c:geo):

image

### DIFF
--- a/main/res/values/changelog_master.xml
+++ b/main/res/values/changelog_master.xml
@@ -4,6 +4,9 @@
   <string name="changelog_master" translatable="false"><Data><![CDATA[<div>
 <h1>Developer version:</h1>
 
+<h4>Android 11 notice:</h4>
+Android 11 enforces a new way for storing data, which c:geo does not support yet. If you switch your system to Android 11, c:geo will not be able to access part of its data. This will most likely disable storing offline maps, importing/exporting data as well as backing up/restoring c:geo\'s database. Other effects may occur. We recommend to not upgrade to Android 11 for the time being.
+
 <h4>Map:</h4>
 <ul>
 <li>New: Map downloader for OpenStreetMap offline maps (see map selector menu and settings - map)</li>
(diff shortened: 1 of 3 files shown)


eddiemuc:
I fully support writing this note. I would however add more clearly from a users perspective which data is not accessible any more and which functionality will not work with Android 11.
(Actually I don't know this myself. Will the sqlite DB still be accessible?)

moving-bits:
> I would however add more clearly from a users perspective which data is not accessible any more and which functionality will not work with Android 11.

Can't say yet in detail, as I haven't explored in-depth so far.

As far as I've seen so far, the so called "public external directory" won't be accessible anymore, at least not using the methods we use now - that would be the `cgeo` dir and its subdirs, which are publicly visible today.

> (Actually I don't know this myself. Will the sqlite DB still be accessible?)

The internal database should still be accessible, but you wouldn't be able to backup to / restore from the public directory (`cgeo/backup`).

Let's start with above note for now, and add along the way as we explore what will work and what does not.
(I'm sort of in a hurry regarding this, as public beta versions are available for Android 11 already, and publishing date for the final version is not too far away.)

eddiemuc:
Sugestion: add following sentence:
Android 11 enforces a new way for storing data, which c:geo does not support yet. If you switch your system to Android 11, c:geo will not be able to access part of its data. This will most likely disable storing offline images (listings and logs) as well as backup/restoring c:geo's database. Other effects may occur. We recommend to not upgrade to Android 11 for the time being.

moving-bits:
@eddiemuc Thanks for the extended text. I'm not sure about the offline images (they go to an internal storage, IIRC), but export to GPX will be harmed, also access to offline maps. I'll update the text accordingly, including the other suggestions from you.

moving-bits:
Updated to:
image
(light theme in this case, unrelated to the change)

Lineflyer:
LGTM.

Maybe we should also add a warning if we detect running on Android 11. For example at the same place where missing login is shown on the mainscreen and link to our FAQ (entry to be created)?

moving-bits:
> Maybe we should also add a warning if we detect running on Android 11. For example at the same place where missing login is shown on the mainscreen and link to our FAQ (entry to be created)?

I'll add it. Which anchor name within the FAQ shall I use? (Even if there's no text behind it yet.)

Lineflyer:
> I'll add it. Which anchor name within the FAQ shall I use? (Even if there's no text behind it yet.)

As you like, e.g. `#android11`
Please kindly open a followup on the website repo once done (or even feel free to add it to the FAQ yourself).

moving-bits:
Added a mainscreen warning "You appear to be using Android 11, which c:geo does not support fully yet. Please read our FAQ to get more info on this." if running on Android >= 11. Links to https://www.cgeo.org/faq#android11.

moving-bits:
> Please kindly open a followup on the website repo once done

is this the `cgeo.github.io` repo?

Lineflyer:
> is this the cgeo.github.io repo?

Jepp

PR #8832, opened 2020-08-17
update contacts to api29

bekuno:
set compileSdkVersion and targetSdkVersion to 29 according to c:geo

### DIFF
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -5,7 +5,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion 29
 
     // signing is handled via private.properties
     signingConfigs {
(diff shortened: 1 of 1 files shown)


Lineflyer:
Shall we release a new plugin version before we merge this?

bekuno:
> 
> 
> Shall we release a new plugin version before we merge this?

We can wait with this PR.

Lineflyer:
@bekuno 
Almost forgot about this one.
I just built and released contacts plugin version 2020.10.13 (new versioning similar to main app). That was still targeting API 28.
Now we can merge this (unless it makes no sense as we rather want to integrate the function back into the main app).

bekuno:
> Now we can merge this (unless it makes no sense as we rather want to integrate the function back into the main app).

Lets merge.
It is not urgent to integrate it in the main app. 
But I hope that I will find time to do it sometime this year.

rsudev:
As long as we support Android versions that don't have user-selectable App permissions, we should not integrate it again in order to not require contact permissions generally.

PR #8841, opened 2020-08-18
additional checks for map downloader (fix #8821)

### LINKED ISSUE #8821: Map file downloads fails if maps dir not writeable
**Describe the bug:**
Map download fails with message "Receive map file - Error while receiving map file". I in fact tried this several times, even some time ago, but was always unsuccessful. Just forgot to report as I'm using different sources anyway.

**To Reproduce**
1. Go to map view
2. Open map list
3. Tap "Download offline map"
4. Select one (over here: Europe => Germany => Baden-Wuerttemberg)
5. Download starts and seems to finish
6. Error message appears

**Version of c:geo used:**
 2020.08.17-NB-2c47d82

Yes

**System information:**
```
--- System information ---
Device: VOG-L29 (VOG-L29EEA, HUAWEI)
Android version: 10
Android build: VOG-L29 10.1.0.140(C431E19R2P5)
c:geo version: 2020.08.17-NB-2c47d82
Google Play services: disabled - 20.30.19 (120400-326531024)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide caches: -
Hide waypoints: visited
System language: en_DE
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (76,2 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (76,2 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (76,2 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (284,1 MB) on system internal storage
Geocaching sites enabled:
   extremcaching.com: Logged in (Login OK)
Geocaching.com date format: dd MMM yy
Installed c:geo plugins: contacts
```

**Additional context:**
As the download seems to work (progress bar in notification bar comes completely to it's end), I would guess there maybe either an unzip or storage (access) problem. Question: Where will c:geo put it's map files?

Already generated a log file, internal and with Logcat reader, but there are no errors in it, and even extended log with Debug Log active just contains two entries for mapdownloader, with get token and release token. In Logcat reader I additionally found these entries (while filtering for "mapdownload"):

```
[08-17 14:21:20.319 2103:2442 E/DollieAdapterService]
notifyActivityState pkg:cgeo.geocaching/cgeo.geocaching.settings.MapDownloadSelectorActivity state:2 fg:true mUid:10313

[08-17 14:21:20.423 760:1465 W/BufferQueueProducer]
[cgeo.geocaching/cgeo.geocaching.settings.MapDownloadSelectorActivity#0]:1368: disconnect: not connected (req=1)

[08-17 14:21:20.428 1279:2157 E/InputDispatcher]
Window handle Window{ab00288 u0 cgeo.geocaching/cgeo.geocaching.settings.MapDownloadSelectorActivity} has no registered input channel

[08-17 14:21:20.432 1279:1306 E/WindowManager]
win=Window{ab00288 u0 cgeo.geocaching/cgeo.geocaching.settings.MapDownloadSelectorActivity EXITING} destroySurfaces: appStopped=false win.mWindowRemovalAllowed=true win.mRemoveOnExit=true
```

Any idea how I should create a more informative debug log?

moving-bits:
- do a "target dir writable?" check before opening the map selector
- do a "target dir writable?" check before copying received map file
- print more detailed info if copying received map file fails

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1885,6 +1885,8 @@
     <string name="receivemapfile_intenttitle">Receive map file</string>
     <string name="receivemapfile_success">Map file successfully received as \"%s\" and stored to map directory</string>
     <string name="receivemapfile_error">Error while receiving map file</string>
+    <string name="receivemapfile_error_io_exception">I/O error while receiving map file. Is your map directory \"%s\" writable?</string>
+    <string name="receivemapfile_error_filenotfound_exception">File not found error while receiving map file.</string>
     <string name="receivemapfile_alreadyexists">A file with this name already exists. Overwrite existing file or save new file under a different name?</string>
     <string name="receivemapfile_option_overwrite">Overwrite</string>
     <string name="receivemapfile_option_differentname">Different name</string>
@@ -1897,6 +1899,7 @@
     <string name="downloadmap_choose">Choose map</string>
     <string name="downloadmanager_not_available">Map file cannot be downloaded, system download manager not available</string>
     <string name="download_started">Downloading started in background</string>
+    <string name="downloadmap_target_not_writable">Target \"%s\" is not writable - please select a different map file directory.</string>
 
     <!-- history trail export -->
     <string name="trailhistory">History track</string>
--- a/main/src/cgeo/geocaching/settings/MapDownloadSelectorActivity.java
+++ b/main/src/cgeo/geocaching/settings/MapDownloadSelectorActivity.java
@@ -183,9 +183,15 @@ public class MapDownloadSelectorActivity extends AbstractActionBarActivity {
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState, R.layout.mapdownloader_activity);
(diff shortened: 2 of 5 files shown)


moving-bits:
OK / Cancel does not make sense for an error message => switched from `confirm` to `message` dialog type ("Ok" only):

image

Lineflyer:
Will the downloaded file be kept in such error case?

If yes, we could at least inform the user, where it is located to be able to manually copy it. The user might not be able (due to size) to set the directory to another one.

moving-bits:
> Will the downloaded file be kept in such error case?
> If yes, we could at least inform the user, where it is located to be able to manually copy it. 

I do not delete it explicitly, but the file is managed by the download manager completely, and from my observations the files are not visible to the end-user. (That seems to be the default case for downloads triggered by apps internally, in contrast to downloads started from a browser. If anyone knows a switch to make it "visible" to the end-user, I'm happy to add it.)

PR #8850, opened 2020-08-20
remove orange line from one screenshot which shouldn't be there

fm-sys:
There was a small orange line on the left side of the first german playstore screenshot, which shouldn't be there and looked quite ugly.

@Lineflyer could you please replace the corresponding screenshot with this new version?


Lineflyer:
Will do!

Lineflyer:
Done!

Issue #8752, opened 2020-08-05
App not working (Freeze) on device with Sailfish OS


Lineflyer:
From support mail:
A user contacted us, that c:geo stopped working for him. The problem probably started with an update of his device.
It is a bit of a special case, however maybe someone over here has an idea.

User mail (summary):
- The issue is most probably related to the underlying Android layer. The device
is running SailfishOS with AlienDalvik running the Android apps. But almost all
other Android apps run fine under this system.
- Another additional note: I tried installing old versions of c:geo. I tried all releases from the most current one until the last from 2018, but all of them fail.
- Resetting doesn't help. Same problem after resetting.
- Logfile is empty, I cannot activate debug log as I cannot reach that menu without freeze.
Issue description:
I can install it and run it once. The "Getting started" page opens, I can browse it and switch to other About pages.
When I tap "Configure your geocaching platform(s)", the platform selection opens, but if I tap on any of the platform names, the app freezes. If I then close the app, it won't start at all on the next try. Reinstall or "Clear data" fixes the startup-issue, but not the freezing.
The freezing happens also when trying to exit the "About" page byt tapping the back button.

System information:
```
--- System information ---
Device: Xperia XA2 Dual (AOSP) (aosp_h4113, Sony)
Android version: 8.1.0
Android build: OC test-keys
c:geo version: 2020.07.02
Google Play services: unavailable
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: absent
Magnetometer & Accelerometer sensor: present
Direction sensor used: magnetometer & accelerometer
Hide caches: -
Hide waypoints: -
HW acceleration: enabled (default state)
System language: en_GB
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (3.1 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (3.1 GB free) external non-
removable
Geocache data:
/storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (3.1 GB
free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (4.0 KB) on system
internal storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled: None
Installed c:geo plugins:  none
--- End of system information ---
```

Ticket Reference:
#826588

mmtj:
[...]

PR #8852, opened 2020-08-20
[Related to #8752] turn on debug mode hardcoded - DO NOT MERGE

eddiemuc:
THis PR turns on debug mode hardcoded (and it is not possible to turn it off again).
It is strictly meant for error analysis of #8752 only and shall NOT BE MERGED!


fm-sys:
Can this PR be closed?

Lineflyer:
I think so. The debug.apk will be still available on our CI.

Issue #8848, opened 2020-08-20
Global config option to disable parsing and creation of anything out of a cache listing's personal cache notes


baiti:
Quite a few geocaching applications have started to create waypoints based on coordinates that they have extracted from a cache listing's personal cache notes (PCN). c'geo does that too. PCN is a free form scratch pad for users to take notes for geocaches. There is no defined format of how PCNs are structured and consequently there is no defined way for a program to parse any meaningful information out of the PCN. Some applilcations, for instance project-gc provides a way for the user to mask coordinates in the PCN with a prefixed caret e.g. "^N48 44.444 E08 44.444" and thus prevent project-gc from creating a waypoint (and render it on the map) for coordinates masked in this way. 

This I'd consider a circumvention of a problem but not a true solution, however, it is still better than nothing and if you for any reason cannot implement the functionality I am going to describe below, at least provide the masking capability by parsing for the caret "^" prefix.

A solution would be a system wide configuration option that allows a user to instruct c'geo to completely avoid parsing the PCN data for anything.

I am aware that recent versions of c'geo provide a checkbox underneath the PCN Editor that reads "Prevent waypoint extraction". While I fully understand that checkboxes semantics, it addresses only a part of the problem as I have no way to tell c'geo to stay away from the contents of the PCN field entirely. c'geo's waypoint extraction function happens at the time a geocache listing is downloaded (or refreshed) from geocaching.com and persisted for offline use in cgeo.

Lineflyer:
Fair enough.
Maybe it is possible to implement it in a way, that the global setting (once active) triggers the checkbox in the PN field to be active by default for all caches (so users can deactivate it for single caches, meaning kind of opt-in instead of opt-out).

eddiemuc:
> Fair enough.
> Maybe it is possible to implement it in a way, that the global setting (once active) triggers the checkbox in the PN field to be active by default for all caches (so users can deactivate it for single caches, meaning kind of opt-in instead of opt-out).
> 

I support this idea and I think it should be easy to implement

Lineflyer:
Go for it, if you would like to implement this.
I do however suggest to keep the current "opt-out" behavior as default so the behavior does not change compared to the current market version.

baiti:
> Go for it, if you would like to implement this.
> I do however suggest to keep the current "opt-out" behavior as default so the behavior does not change compared to the current market version.

[...]

PR #8856, opened 2020-08-20
New entry in settings for cache details to disable extraction of waypoints from personal cache notes

baiti:
see issue #8848


main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -194,6 +194,7 @@ private static void migrateSettings() {
             e.putInt(getKey(R.string.pref_defaultNavigationTool), prefsV0.getInt(getKey(R.string.pref_defaultNavigationTool), NavigationAppsEnum.COMPASS.id));
             e.putInt(getKey(R.string.pref_defaultNavigationTool2), prefsV0.getInt(getKey(R.string.pref_defaultNavigationTool2), NavigationAppsEnum.INTERNAL_MAP.id));
             e.putBoolean(getKey(R.string.pref_debug), prefsV0.getBoolean(getKey(R.string.pref_debug), false));
+            e.putBoolean(getKey(R.string.pref_global_wp_extraction_disable), prefsV0.getBoolean(getKey(R.string.pref_global_wp_extraction_disable), false));
```
moving-bits, line 197: IMHO this line is not needed, as this setting is new, so no need to migrate.

Lineflyer:
OK to test (for CI)

main/res/values/strings.xml
```diff
@@ -600,6 +600,8 @@
     <string name="init_summary_disabled">Exclude disabled caches</string>
     <string name="init_archived">Exclude Archived</string>
     <string name="init_summary_archived">Exclude archived caches</string>
+    <string name="init_global_wp_extraction_disable">Waypoint Extraction</string>
+    <string name="init_summary_global_wp_extraction_disable">Disable waypoint extraction from personal cache notes</string>
```
Lineflyer, line 604: I suggest something like:
"Globally disable waypoint extraction from personal note. It can still be enabled for individual caches by deactivating the checkbox below the personal note of the cache."

Lineflyer:
OK to test

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1558,7 +1558,7 @@ public boolean addWaypointsFromNote() {
      */
     public boolean addWaypointsFromText(@Nullable final String text, final boolean updateDb, @NonNull final String namePrefix, final boolean forceExtraction) {
         boolean changed = false;
-        if (forceExtraction || !preventWaypointsFromNote) {
+        if (forceExtraction || (!Settings.isGlobalWpExtractionDisabled() && !preventWaypointsFromNote)) {
```
moving-bits, line 1561: Not sure if I understand this condition correctly. To my understanding we wanted to implement an opt-out system. Wouldn't we need to check this in the `Geocache` class then instead of here? Distinguish on cache-level between a manually set value (true/false), which would override the global setting, and a "not yet set" value, which would activate the global setting for this cache?

Lineflyer: > To my understanding we wanted to implement an opt-out system.

Cannot judge the code, but indeed an opt-out system was the idea:
- If the setting is not active (which should be default) waypoints are extracted and only surpressed if the user activates the checkbox under the specific personal note.
- It the setting is active (set by user, no default) the checkbox should be activated by default for all caches and can be individually deactivated by the user for each cache.

baiti: > > To my understanding we wanted to implement an opt-out system.
> 
> Cannot judge the code, but indeed an opt-out system was the idea:
> 
> * If the setting is not active (which should be default) waypoints are extracted and only surpressed if the user activates the checkbox under the specific personal note.

That is indeed the implemented behaviour

> * It the setting is active (set by user, no default) the checkbox should be activated by default for all caches and can be individually deactivated by the user for each cache.

Ahh ok, that is not what the change does at the moment, however, I agree that is a much better. I understand that this type of change does not go into the place I put it. I have not found out yet, where in the code I could set the checkbox you mentioned but I will investigate that tomorrow.

Once I have it done, would I do another push into the same branch?

moving-bits: > Ahh ok, that is not what the change does at the moment, however, I agree that is a much better. I understand that this type of change does not go into the place I put it. I have not found out yet, where in the code I could set the checkbox you mentioned but I will investigate that tomorrow.

Database change would go into `storage/DataStore`, applying the default value to `models/Geocache` (IIRC).

> Once I have it done, would I do another push into the same branch?

Yes, for the time being. (At then end, when everything is fine, we would squash the individual commits into one commit containing all the changes necessary for this PR.)

Thanks for working on this issue! Very welcomed!

main/res/values/strings.xml
baiti: > I suggest something like:
> "Globally disable waypoint extraction from personal note. It can still be enabled for individual caches by deactivating the checkbox below the personal note of the cache."

Ok, consider it done. Will go up with the next push

main/src/cgeo/geocaching/settings/Settings.java
baiti: I wasn't sure ;) Now I have removed it.

main/src/cgeo/geocaching/models/Geocache.java
eddiemuc: > > * It the setting is active (set by user, no default) the checkbox should be activated by default for all caches and can be individually deactivated by the user for each cache.
> 
> Ahh ok, that is not what the change does at the moment, however, I agree that is a much better. I understand that this type of change does not go into the place I put it. I have not found out yet, where in the code I could set the checkbox you mentioned but I will investigate that tomorrow.

Sorry for late comment..
Imho, the easiest way to achieve the desired behaviour is to set the default setting of `Geocache.preventWaypointsFromNote` to the new setting. Meaning: in `Geocache.java`, change line 176:
`    private boolean preventWaypointsFromNote = false;`

to

`    private boolean preventWaypointsFromNote = Settings.isGlobalWpExtractionDisabled();`

That way the checkbox handling is automatically "done", and also the value would be correctly overridden from Database in case the cache is already stored there with a different setting.

baiti: > > > * It the setting is active (set by user, no default) the checkbox should be activated by default for all caches and can be individually deactivated by the user for each cache.
> > 
> > 
> > Ahh ok, that is not what the change does at the moment, however, I agree that is a much better. I understand that this type of change does not go into the place I put it. I have not found out yet, where in the code I could set the checkbox you mentioned but I will investigate that tomorrow.
> 
> Sorry for late comment..
> Imho, the easiest way to achieve the desired behaviour is to set the default setting of `Geocache.preventWaypointsFromNote` to the new setting. Meaning: in `Geocache.java`, change line 176:
> ` private boolean preventWaypointsFromNote = false;`
> 
> to
> 
> ` private boolean preventWaypointsFromNote = Settings.isGlobalWpExtractionDisabled();`
> 
> That way the checkbox handling is automatically "done", and also the value would be correctly overridden from Database in case the cache is already stored there with a different setting.

Hi Eddie, you wouldn't believe it, that is **exactly** the change I just made ;) I have started to test it and it works as expected and as discussed in this thread yesterday.

I am going to push it shortly. Please have a look at the code and prefereably I'd also want some of you to give it a real try and walk through some usage scenarios on the device.

baiti: I have just committed it into yesterday's pull request. I hope this was the right way doing it from a process point of view.

baiti:
I have actually already started to do so. First I deleted my fork on github, then I deleted my local clone and started all over. I clicked on edit in the README.md and noticed that "the system" edited the README.md in my fork and that everything I do on the Wiki, happens in my fork. Now I was just wondering, when I do a PR at some point in time, will the links I put into the Wiki be automatically converted to point to cgeo/cgeo as opposed to baiti/cgeo when a mergs is done?

Narrat:
> > One additional note: while I too like to work almost completely "manually" and on the command line with git, resolving conflicts is the one thing where I appreciate the help of Android Studio.
> 
> Good to know, will try it the next time. Thanks for the hint!

Worth noting that on the commandline is the command mergetool for those cases. Or if someone don't want to manually search for the conflicts. The diff program can be freely configured (vimdiff, meld and whatnot)

PR #8881, opened 2020-08-25
Support drag & drop for route sort view

### LINKED ISSUE #8836: [individuell route] Better UI implementation of sorting way points
The current UI is not really User friendly, so i whould surgest that Sorting individual route should use an UI that looks similar to this:

Screenshot_20200818_081908
I think the behavior should be clear so far. Click and hold one of the symbols on the right side to drag its waypoint to a new position...

fm-sys:
fix #8836


moving-bits:
On a quick glance this looks good to me, thanks @fm-sys for tackling this issue!

Just add some themeing support, cause when running with light theme activated it is hard to read:

See the white buttons in light theme:
image

Light theme, moving an element - item description gets lost:
image

Dark theme, moving an element:
image

Also the border of the "delete" button is lost in light theme, but when we switch to swipe left gesture for deleting, the button is obsolete in this view, anyway.

Arrow up/down buttons can be deleted, IIRC they are only used in this view.

moving-bits (changes requested):
see https://github.com/cgeo/cgeo/pull/8881#issuecomment-681027007

moving-bits:
Additional info @fm-sys:
- For setting the buttons with `setImageResource()` you can distinguish between light and dark theme by using `Settings.isLightSkin()`.
- For text coloring etc. you can apply a style which sets the `textColor` attribute. (Maybe there is an existing style set you can reuse for this view, or you may have to define new ones.)

fm-sys:
The delete button is the same as in cache details view, so mabe we would have to think about a complete button redesign in light theme, but I think most users aren't using light theme anyway...

Everything else is fixed in light mode (black color for reorder icon and grey float view background color)

moving-bits (commented):
Thanks for the update. UI is fine now in light skin mode, but I've just stumbled upon a different issue:

Can it be that you swap the elements instead of inserting the dropped item at the dropping position?

E.g.:
- Take the bottom most item and move it to the top of the list.
- Now those two entries get swapped instead of the bottom one inserted before the current top element.

fm-sys:
Oh yes. I reused your function which was swapping the items (wich was ok as long as the items get only moved 1 step up or down). I will fix this...

fm-sys:
@moving-bits could you please do a final cross check? As far as I know everything should work properly now...

moving-bits (approved):
Played around a bit with it in the emulator, LGTM.
Tanks @fm-sys for upgrading the view to make it much more user friendly!

PR #8899, opened 2020-08-29
add Chirp Wolf to useful apps list (fix #8883)

### LINKED ISSUE #8883: Add Chirpwolf as useful app (and link to it for beacon caches)
Some while ago we had to remove a useful app, which supported chirp receiving (see #3951).

There is a free app now which works quite good (for me):
https://play.google.com/store/apps/details?id=de.wolffire.chirpwolf

We should probably add this to our useful apps and (if desired) we could add a link to this app into cache details (similar to whereyougo for Wherigo caches) when the cache has the beacon attribute set.

moving-bits:
image


Lineflyer:
To be more precise:
It doen't allow you to play these caches, it just allows you to read Garmin chirp data needed for some geocaches.
IMHO it does not support reading NFC tags.

Proposal:
"Chirp Wolf allows you to read the data from Garmin chirp devices, which is needed for some geocaches."

moving-bits:
Is it only Garmin chirp devices or regular NFC tags?

"Chirp Wolf allows you to read the data from NFC/Garmin chirp tags, which is needed for NFC/chirp geocaches."?

Lineflyer:
AFAICS it only supports ANT+, which means Garmin chirp. NFC is something totally different.

moving-bits:
> AFAICS it only supports ANT+, which means Garmin chirp. NFC is something totally different.

Ok, I'll adapt the text accordingly.

I got confused because IIRC I always have to activate NFC to use ANT+ - but maybe that's just a quirk of the Samsung UI (or I'm mistaken and would not have to activate NFC at all to use ANT+).

PR #8903, opened 2020-08-29
Add configuration options to logging (fix #8851)

### LINKED ISSUE #8851: Introduce configurable Logging framework
Issues #8777 and #8752 got me thinking whether c:geo should provide more possibilities to configure the way it logs information:
* configurable via some configuration file
* log level configurable not only for whole application but on a lower level
* ability to log (temporary) to different targets than logcat/system log

c:geo should look on some default place for a log configuration file on startup and use this configuration. This way, if we need information about a specific problem not easy to reproduce we could ask the user to install a specific log config file and send the result to us. Would be useful e.g. for #8752 and similar things.
For migration, implementation can be hidden behind `cgeo.geocaching.utils.Log`

Coming from a strong java background, my first thought would be slf4j combined with logback-android. 

There are other solutions out there of course, e.g. https://android-arsenal.com/tag/57?sort=updated&category=1. Currently I don't know most of them.

Also interesting: https://www.triology.de/blog/slf4j-und-logback-android (german)

eddiemuc:
This PR adds the option to configure Logging by issuing a config file "log.properties" into the "logfiles" directory. Currently following options are supported:
```
/** Minimum log level. Value should be one of {@link LogLevel} in textual form */
    public static final String PROP_MIN_LOG_LEVEL = "logging.minlevel";
    /** Minimum log level to add callerinfo to log message. Value should be one of {@link LogLevel} in textual form */
    public static final String PROP_MIN_CALLERINFO_LEVEL = "logging.mincallerinfolevel";
    /** Whether to throw an exception when an error is logged. Value should be true or false */
    public static final String PROP_THROW_ON_ERROR_LOG = "logging.throwonerror";
```

This attempts to solve #8851 while NOT going to the overhead of introducing a dedicated logging framework

### DIFF
--- a/main/src/cgeo/geocaching/utils/Log.java
+++ b/main/src/cgeo/geocaching/utils/Log.java
@@ -1,33 +1,59 @@
 package cgeo.geocaching.utils;
 
-import android.os.Environment;
+import cgeo.geocaching.storage.LocalStorage;
 
-import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
-import java.nio.charset.StandardCharsets;
-
-import org.apache.commons.io.IOUtils;
+import java.io.FileReader;
+import java.util.Properties;
 
 public final class Log {
 
     private static final String TAG = "cgeo";
 
+    public enum LogLevel { VERBOSE, DEBUG, INFO, WARN, ERROR, NONE }
+
+    /** Name of File containing Log properties which will be searched for in logfiles-directory */
+    private static final String LOGPROPERTY_FILENAME = "log.properties";
+    /** Minimum log level. Value should be one of {@link LogLevel} in textual form */
+    public static final String PROP_MIN_LOG_LEVEL = "logging.minlevel";
+    /** Minimum log level to add callerinfo to log message. Value should be one of {@link LogLevel} in textual form */
+    public static final String PROP_MIN_CALLERINFO_LEVEL = "logging.mincallerinfolevel";
+    /** Whether to throw an exception when an error is logged. Value should be true or false */
+    public static final String PROP_THROW_ON_ERROR_LOG = "logging.throwonerror";
+
     /**
+     * If the debug flag is set then minimum log level is debug AND an exception is thrown on error logging
      * The debug flag is cached here so that we don't need to access the settings every time we have to evaluate it.
      */
     private static boolean isDebug = true;
-    private static boolean first = true;
 
-    private static final class StackTraceDebug extends RuntimeException {
-        private static final long serialVersionUID = 27058374L;
+    private static LogLevel minLogLevel = LogLevel.WARN;
+    private static boolean logThrowExceptionOnError = false;
+    private static LogLevel minLogAddCallerINfo = LogLevel.NONE;
+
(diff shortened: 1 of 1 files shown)


moving-bits:
@eddiemuc 
Can you provide an example file for logging configuration?

eddiemuc:
> @eddiemuc
> Can you provide an example file for logging configuration?

Sure. The following, if put into a file `log.properties` in folder `logfiles` would trigger logging starting at INFO level with caller info included starting at WARN level and NO throwException on error logging.
(Note that if you set "debug" to true in Settings, this overrides the `log.properties`: it raises the minimum log level to DEBUG and turns on throwException on error logging)

```
logging.minlevel=INFO
logging.mincallerinfolevel=WARN
logging.throwonerror=true
```

moving-bits:
Thanks for the example config - makes it easier for me to understand.

One question, though: Shall this config file always be interpreted, or shall we activate this behavior only when debug mode in c:geo has been activated? I see pros and cons for both ways:
- config file and implicit debug settings clashing may confuse the user
- using defined settings when in debug mode helps against having a config file left over
- OTOH leaving a config file (forgetting it) may lead to constant logging unnecessarily when not coupled to an explicitly activated debug mode

One possible solution:
- write extra debugging info only when debug mode is activated (like today)
- add an extra setting for "use config file for settings" (falling back to defaults if no config file found)

Does this make sense?

And/or we can print the current debug settings to the system info screen.

eddiemuc:
Hi @moving-bits ,

> Shall this config file always be interpreted, or shall we activate this behavior only when debug mode in c:geo has been activated? 

when I implemented this I had things like #8752 in mind, where we have to analyze a situation where the user is even unable to activate debug logging in settings because c:geo crashes/freezes before. (of course, since the user did not react on the issue, this might be a bad example in hindsight...)
In these cases we could send a prepared log.properties to the user, ask him/her to put it into logfiles and get the logs we need.
  
> * config file and implicit debug settings clashing may confuse the user

My idea is that from user perspective nothing changes compared to today. This should be more of a "backdoor" for in-field-support and of course for development.

> * OTOH leaving a config file (forgetting it) may lead to constant logging unnecessarily when not coupled to an explicitly activated debug mode

Thats a valid point. What about adding a "validuntil"-date to the log configuration? Default could be "20 days after log-config-creationdate" After this date the log config will no longer be used.

> One possible solution: 
> * write extra debugging info only when debug mode is activated (like today)
> * add an extra setting for "use config file for settings" (falling back to defaults if no config file found)

As said before, my idea was that for user nothing changes compared to today

moving-bits:
> when I implemented this I had things like #8752 in mind, where we have to analyze a situation where the user is even unable to activate debug logging in settings because c:geo crashes/freezes before. (of course, since the user did not react on the issue, this might be a bad example in hindsight...)
> In these cases we could send a prepared log.properties to the user, ask him/her to put it into logfiles and get the logs we need.

Ah, I see. Very valid use-case.

> > * OTOH leaving a config file (forgetting it) may lead to constant logging unnecessarily when not coupled to an explicitly activated debug mode
> 
> Thats a valid point. What about adding a "validuntil"-date to the log configuration? Default could be "20 days after log-config-creationdate" After this date the log config will no longer be used.

Good idea. Maybe a `validfordays=20` or `validuntil=2020-09-15`?

But let's start with merging this PR now, the `valid*` option can be added later, IMHO.

Thanks for this contribution, I'm looking forward to using this in real life ;-)

Issue #8876, opened 2020-08-24
[Nightly] cgeo aborts when opening map


geocachermgo:
**Describe the bug:**
cgeo aborts when opening map. 

**To Reproduce:**
Open Live Map

**Version of c:geo used:**
2020.08.24-NB-f5f21f0

**Is the problem reproducible:**
Yes
I changed the map from google map to offline,  that did not change something.
If I try to open the map display of a list, the same happens.

**Screenshots:**
![Screenshot_2020-08-24-16-20-02-406_com miui home](https://user-images.githubusercontent.com/43813636/91056216-f8c41580-e625-11ea-84d6-7978306f9f86.jpg)

**System information:**
```--- System information ---
Device: MI 8 (dipper, Xiaomi)
Android version: 10
Android build: QKQ1.190828.002 test-keys
c:geo version: 2020.08.24-NB-f5f21f0
Google Play services: disabled - 20.26.14 (120400-320008519)
Low power mode: active
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide caches: own/found 
Hide waypoints: -
HW acceleration: enabled (default state)
System language: de_DE
System date format: dd.MM.yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (90,0 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (90,0 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (90,0 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (10,8 MB) on system internal storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
Geocaching.com date format: MM/dd/yyyy
Installed c:geo plugins:  none
BRouter connection available: false
--- End of system information ---
```

**Additional context:**
e.g. Reference to other issues, projects, sources, etc.

geocachermgo:
I installed the normal play store release and when I imported the settings: cgeo won't even open but abort when I try to open it.

geocachermgo:
I removed the password from the settings file. Is it then save to upload it here?

Lineflyer:
The setting might be incompatible between play and nightly...not sure.

Lineflyer:
[...]

PR #8906, opened 2020-08-29
throw exception on database upgrade error

moving-bits:
Also related to #8876 (see https://github.com/cgeo/cgeo/issues/8876#issuecomment-681843301):

Throw an exception if any upgrade step fails (instead of just skipping the failing step).

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -53,6 +53,7 @@ import android.content.res.Resources;
 import android.database.Cursor;
 import android.database.DatabaseUtils;
 import android.database.MatrixCursor;
+import android.database.SQLException;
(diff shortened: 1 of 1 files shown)


eddiemuc:
Good idea!
One question though: At least in dev environment / emulator I had the impression that Android automatically restarts app if such an exception occurs - leading to an endless restart cycle. Is there any way to prevent this?

moving-bits:
(removed the test code in upgrade to v76)

moving-bits:
> One question though: At least in dev environment / emulator I had the impression that Android automatically restarts app if such an exception occurs - leading to an endless restart cycle. Is there any way to prevent this?

Not that I know of. But AFAIK after two (?) repeated restarts a popup opens, something like "app is constantly failing. Do you want to quit program?".

Issue #8896, opened 2020-08-28
ribbonizer does not work with adaptive launcher icons (from Android 8.0)


bekuno:
After changing our icon to an adaptive icon it does not show the "debug" overlay created by the Ribbonizer.
This addon seems not to be further developed - latest release is from August 2017.

Therefore I suggest to switch to another plugin.
I found easylauncher-gradle-plugin - should work for now, but also not in active development (latest from October 2018).

An alternative would be to have own images for the build flavers as described in https://stackoverflow.com/questions/19528063/android-app-icon-for-debug-and-release-mode.

Any ideas?

moving-bits:
Or shall we simply leave out the "debug" ribbon at all?
You can still use the "about" page to see whether you are running a debug build (and most people having installed a debug version know that anyway).

bekuno:
The start icon on android is the more interesting one to see before starting c:geo what type is installed.
But you are right, normally you know what you start ....

The now used adaptive-icon can not changed at compile time (has alredy forground and background layer).
So I propose to let the state for older devices until we change the minSdkVersion to 26 or we have another solution.

I will prepare an PR to change the ribbon collor to blue, because the current yellow is bad to distinct from the icon backgroud.
![Screenshot_20200830-171426_1](https://user-images.githubusercontent.com/32555742/91662811-0d505400-eae5-11ea-939e-757a5d9d720c.jpg)
![Screenshot_20200830-171438_1](https://user-images.githubusercontent.com/32555742/91662815-0fb2ae00-eae5-11ea-923d-78b7a55ea726.jpg)

Lineflyer:
Would it make sense to also add it for our nightlies (if possible in the current flavor configuration).
From time to time users started using it without being aware, what it is and that they are still using it after a while. Therefore it could be helpful (also for my testing).

bekuno:
Should be possible for the nightlies.
My short test however was not yet successful.

bekuno:
would this ribbon be ok (RGB FF00FF)? 
![grafik](https://user-images.githubusercontent.com/32555742/91764185-26393200-ebd7-11ea-8c63-9198cc2e380e.png)

Idea for legacy - grayScale:
![grafik](https://user-images.githubusercontent.com/32555742/91767525-8e3e4700-ebdc-11ea-9ebb-574aa881ec23.png)
Or is that too much change?

Should I also implement it for rc?
Are these colors ok?

Lineflyer:
[...]

Issue #8096, opened 2020-01-22
Unattached routing target


Lineflyer:
Please exuse my brevity and missing investigation:

We had the discussion in an issue here, that Google Maps routing line under some circumstance can get unattached from the target. Meaning the red line does not end on the selected cache but somewhere in nowhere.

This has also been reported twice as feedback to our beta version and should be investigated. I would favor if someone could add needed details to this issue from the source where it was reported. I can add more screenshots from support next week if needed.

PR #8916, opened 2020-08-30
change ribbon color

bekuno:
related to #8896

- change the ribbon color to blue
- add reference to the issue #8896 to have the problem visible for other
- remove the extra iconNames because this is already the default from AndroidManifest.xml


moving-bits:
(edited the text to set the reference to #8896 instead of #8096)

Issue #8893, opened 2020-08-27
"Map download" menu entry blank (with GM and GS)


MagpieFourtyTwo:
**Describe the bug:**
In the Map menu (while displaying the Map) you can choose to download an offline map.
This looks pretty fine while using an offline map,
but when you're using Google Map or Google Sat, the slot, where the download entry should be, is just empty.
Nevertheless, it can be clicked and starts to (try to) download maps ...

**To Reproduce:**
1. Open map page with Google Map or Google Sat view
2. Click on 'Map' menu
3. Scroll down to last slot ... it's empty
4. click it ... map download procedure starts

**Expected behavior/state after performing these steps:**
The slot should either be labeled correctly (as map download seems to work anyway), or completely be removed.

**Version of c:geo used:**
2020.08.27-NB-422e2ae

**Is the problem reproducible:**
Yes

moving-bits:
Strange. It's being shown here in both modes:

![image](https://user-images.githubusercontent.com/3754370/91630948-0c7ccc80-e9d6-11ea-8a73-c69bca9dd69d.png)

fm-sys:
I can verify the bug exists:
![Screenshot_20200829_090658_cgeo geocaching](https://user-images.githubusercontent.com/64581222/91631080-2539b200-e9d7-11ea-826c-a528905195fb.jpg)

moving-bits:
@fm-sys 
Can you set a breakpoint at line 85 in `maps/MapProviderFactory` and debug into it?

```
parentMenu.add(Menu.NONE, R.id.menu_download_offlinemap, mapSources.size() + 1, R.string.downloadmap_title);
```

MagpieFourtyTwo:
Possibly depends on Dark Mode?

moving-bits:
> Possibly depends on Dark Mode?

It does work here in both modes. That's why I need to get more info on it from someone experiencing that problem, either system info (API level etc.) and/or by throwing it into the debugger and seeing what's going on.

fm-sys:
Probably it is related to #7587, but I hadn't time to debug into it yet

fm-sys:
Here is a log file where I switched from Mapsforge to GMap and then started the Offline Map downloader. Unfortunately I can't debug it with android studio because I have no api keys for google maps (yet)... 

[logcat_2020-08-30_19-32.txt](https://github.com/cgeo/cgeo/files/5147226/logcat_2020-08-30_19-32.txt)

[...]

PR #8918, opened 2020-08-30
change group setting for download map menu entry

moving-bits:
@fm-sys 
Can you test this APK in your environment?

I've implemented a different group marker for the "download map" menu item - maybe that works in your setting.
Also I've added some `Log.i()` calls to output some debug info about the menu items while adding the map sources.

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -20,9 +20,9 @@
                 android:id="@+id/menu_group_map_sources"
                 android:checkableBehavior="single" >
             </group>
-            <group
-                android:id="@+id/menu_download_offlinemap">
-            </group>
+            <item
+                android:id="@+id/menu_download_offlinemap"
+                android:visible="false" />
         </menu>
     </item>
     <item
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -719,7 +719,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
     public boolean onCreateOptionsMenu(@NonNull final Menu menu) {
         super.onCreateOptionsMenu(menu);
 
-        MapProviderFactory.addMapviewMenuItems(menu);
+        MapProviderFactory.addMapviewMenuItems(activity, menu);
 
         /* if we have an Actionbar find the my position toggle */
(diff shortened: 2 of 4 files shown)


fm-sys:
Unfortunately no success, Map Downloader Entry is still not displayed...

[logcat_2020-08-31_15-52.txt](https://github.com/cgeo/cgeo/files/5150729/logcat_2020-08-31_15-52.txt)

moving-bits:
@fm-sys 
Thanks for testing this scenario.

Ok, as a fallback let's try something different: "Download offline map" now is a hard-coded item, and the different map sources get added dynamically. This leads to something like this:

image

Can you do another test (when the build is finished)?
I've removed the debug info, so no logcat needed this time.
Thanks.

fm-sys:
This time the Google Maps item is hidden... (Again only for google maps/satellite)

Screenshot_20200831_200312_com huawei himovie overseas

moving-bits:
This. is. strange. 

:-(

Thanks for testing. As I still cannot reproduce the problem: Do you have any system specifics for me to look for?
I've tested against emulator with both API 30 (Android 10) and 21 (Android 5), but both work.

fm-sys:
I've tested the apk on my Huawei mate 20 lite running with android 10... So mabe EMUI specific. @MagpieFourtyTwo on which device did the bug occur? (model and android version) 

@moving-bits two ideas (haven't tried):
- isn't it possible to also hardcode the downloader as last element (with string and everything)?
- what about adding all the elements to one group (then we probably wouldn't have this separator line but it would be a try)

moving-bits:
> * isn't it possible to also hardcode the downloader as last element (with string and everything)?

I had tried this before, but dismissed it, because the map sources got added at the end anyway. Haven't checked though, whether that variation helps with #8893.

> what about adding all the elements to one group (then we probably wouldn't have this separator line but it would be a try)

Will try that next. This will be without separator then, because separators are separating different groups only.

moving-bits:
@fm-sys 
Can you test this variant? (Failing check is just a Checkstyle message, irrelevant for testing)
This time all menu entries are created dynamically, and all are in the same menu group.

fm-sys:
Works fantastic!!!!! 

Just mabe add some `< ... >` like is is done for "new list" in the list selector. What do you think? Than the map downloader would be more visible and distinguished...

moving-bits:
What a relief to finally have found a cross-device working solution... Thanks for testing!
Good idea with the `< ... >`, will add that.

PR #8922, opened 2020-08-31
add Chirp Wolf icon to cache detail page for chirp caches (fix #8883)

### LINKED ISSUE #8883: Add Chirpwolf as useful app (and link to it for beacon caches)

moving-bits:
see red box:

image


Lineflyer:
LGTM!

Lineflyer (approved):

PR #8923, opened 2020-08-31
fix db upgrade error

moving-bits:
While doing some tests for #8918 on an older physical device running the release version, I've stumbled upon an database upgrade error, depending on your specific upgrade path. This PR fixes this.


moving-bits:
retest this please

PR #8936, opened 2020-09-03
Handle null-value ratings in stored offline logs (fix #8935)

### LINKED ISSUE #8935: [Nightly] Crash when tap on Saved Log
**Describe the bug:**
After upgrading to latest nightly I can not open any ``Saved Log`` anymore, which has been saved with a previous nightly.
As soon as I tap the displayed log text, c:geo crashes with a NullPointerException - pls. see attached logfile.
These ``Saved Logs`` are created with nightlies dated August, 21st up to yesterday's nightly.
If I create a *new* ``Saved Log`` with the current nightly and reopen it, this works.

**To Reproduce:**
1. open any cache with an already offline ``Saved Log`` (must be from a previous nightly!)
2. click on the ``Saved Log`` text to edit it

c:geo immediately crashes and restarts in previously selected cache list.

Workaround: In order to at least get rid of the non-editable ``Saved Log``, you can use ``Select mode`` and ``Manage Caches`` => ``Clear offline logs`` to delete it (after uploading it to gc.com, of course ;) ).

**Version of c:geo used:**
2020.09.03-NB-868fddb

Yes

**System information:**
```
--- System information ---
Device: VOG-L29 (VOG-L29EEA, HUAWEI)
Android version: 10
Android build: VOG-L29 10.1.0.140(C431E19R2P5)
c:geo version: 2020.09.03-NB-868fddb
Google Play services: disabled - 20.30.19 (120400-326531024)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide caches: own/found 
Hide waypoints: original 
System language: en_DE
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (76,2 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (76,2 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (76,2 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (286,4 MB) on system internal storage
Geocaching sites enabled:
   extremcaching.com: Logged in (Login OK)
Geocaching.com date format: dd MMM yy
Installed c:geo plugins: contacts
BRouter connection available: true
```

**Additional context:**
Remark: The log file moreover contains an ArrayIndexOutOfBoundsException (dated 08-26) - this was with a previous nightly version and most probably happened while switching from OSM to GM - this happens from time to time, but is not reproducible (yet) - perhaps this helps anyway, so I just left it in the log file.

[logcat_2020-09-03_02-32.txt](https://github.com/cgeo/cgeo/files/5165509/logcat_2020-09-03_02-32.txt)

eddiemuc:
fix #8935

### DIFF
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -346,7 +346,7 @@ public class LogCacheActivity extends AbstractLoggingActivity {
         date.setDate(new Date(logEntry.date));
         setLogText(logEntry.log);
         reportProblem.set(logEntry.reportProblem);
-        cacheVotingBar.setRating(logEntry.rating);
+        cacheVotingBar.setRating(logEntry.rating == null ? cache.getMyVote() : logEntry.rating);
         favCheck.setChecked(logEntry.favorite);
         tweetCheck.setChecked(logEntry.tweet);
         logPassword.setText(logEntry.password);


Lineflyer:
Any other of the surrounding values which might return `null`? Please do also think of logs for other platform, which might not support certain features (such as favorite or others)?

eddiemuc:
> Any other of the surrounding values which might return `null`? Please do also think of logs for other platform, which might not support certain features (such as favorite or others)?

Nope, that's the only problematic one

PR #8943, opened 2020-09-03
custom tab implementation instead of internal browser

### LINKED ISSUE #6680: Use Chrome View instead of internal browser
Sorry, that I cannot describe any technical detail, but I just learned from using some apps, that there seems to be a possibility to "embedd" a HTML browser within our app which  works hand in hand with Google Chrome.

You might know this function if you use Google Now and it comes up with search results in a view which looks like Chrome but isn't really Chrome. From there you have the ability to open the same page in Chrome via menu.
It seems also to use Cookies and configuration stored in Chrome, which the internal browser we have implemented is not capable of.

I have now seen more and more apps using this possibility, so it might be there is easy possibility to implement it and get rid of the internal browser which has several problems and is IMHO unsuable.

fm-sys:
Cache detail page - open in browser method:
- removes c:geo internal browser
- uses chrome custom tab if chrome is installed and enabled
- otherwise show intent chooser or start the default browser

Fix #6680


Lineflyer:
@fm-sys Shall we try this on master (if you rebase the PR)?

fm-sys:
I was struggling whether we should use costum tabs at all places or only for viewing caches in the browser. Other places could for example be link user profile page or the message center. 

Anyway, I will probably not be able to finish it before next weekend... But you are right, this PR had been lying on the siding long enough ;-)

fm-sys:
It was a hard way to find a solution that works on all devices with different installed browsers and apps (as far as I could test it). From my side it's ready to merge now...

There is only (IMHO acceptable) regression:
If chrome is installed, the official GC app isn't listed when clicking "Open with..." inside the custom tab as this app only filters for `coord.info` links, not for `www.geocaching.com/geocache/` which is loaded inside the custom tab. Should I add a workaround for converting the link back to `coord.info`? But IMHO it's their job to make their app working correctly. And BTW, why should someone try to open a cache with the official GC app, if he has c:geo installed? xD

Lineflyer:
> If chrome is installed, the official GC app isn't listed when clicking "Open with..." inside the custom tab as this app only filters for coord.info links, not for www.geocaching.com/geocache/ which is loaded inside the custom tab. Should I add a workaround for converting the link back to coord.info? But IMHO it's their job to make their app working correctly. And BTW, why should someone try to open a cache with the official GC app, if he has c:geo installed? xD

Why is this only if chrome is installed?

fm-sys:
> Why is this only if chrome is installed?

If chrome isn't installed, an intent chooser will be shown similar as today with the `coord.info` link as URL...

Lineflyer:
> If chrome isn't installed, an intent chooser will be shown similar as today with the coord.info link as URL...

mmh..Why don't you use the coord.info link in first place then?

fm-sys:
The link for the "open with..." action is request dynamic from the currently shown web page. For example when you navigate inside the webview to a different cache, the link will change (as it would be expected). This behavior comes along with using a user defined action inside costum tab automatically and is not implemented by myself. 

We only have influence on what we do with that link. For example we could convert it back to the `coord.info` scheme as written above before we sent it to the intent chooser.

PR #8948, opened 2020-09-04
add legacy number template

### LINKED ISSUE #8947: Made offline log count in log templates optional
Hey guys,

thanks for the create work with new beta release!

It is possible to made it optional that offline logs count towards the number in log templates?

How I log caches:
1. I create a offline log while a geocache tour.
2. At end of day I write a text as a template.
3. Than I edit each offline log and insert the log template.
4. I submit the log. and do that again with the next geocache.

There is than the wrong number. I want that there is the current find count + 1

Thanks!

fm-sys:
fix #8947 for users who don't like the new template


fm-sys:
Sorry, wrong branch...

Issue #8632, opened 2020-07-14
Saving offline logs takes up to three seconds


MagpieFourtyTwo:
**Describe the bug:**
Saving a (no matter if short or somewhat longer) log text offline takes a *lot* of time, and I don't really get why.  AFAICS, there (may) not happen more than just removing trailing blanks (experienced this from trial&error) and ... what else?

Online connection *should* not be the problem, as it seems to be unnecessary anyway (log type and TB list has already been fetched on opening of the write window and TB state is set to remain unchanged [overe here]), but *may* be an issue (for whatever reason) as it seems to be a little bit faster in WLAN, but in fact there should just be the insertion of the log text into the DB, which should not take more than a fraction of a second ... or am I missing s/th?

Would be great if someone may have a look into this, as waiting several seconds is somehow annoying.

**Version of c:geo used:**
2020.07.13-NB-2b03a7

**Is the problem reproducible:**
Yes

Lineflyer:
Never seen that. How do you save the offline log (pressing back from logging view, Menu -> Save, or using the one click offline logging)?

IMHO it could only take longer than just the system reaction time, AFAIK there is no online action associated with it.

And not reproducible here: 2020.07.25-NB, Android 10, Samsung S10
Whats your device?

MagpieFourtyTwo:
> How do you save the offline log 
Hamburger menu => "Save offline" (cause just leaving via back button sometimes gave problems in the past)

> Whats your device?
Huawei P30 Pro with Android 10.1.0

Lineflyer:
Today I had problem to enter settings, which also took up to 3 seconds.
After device reboot it was gone....maybe we really do have something under the hood, which sometimes eats up resources?

moving-bits:
Do the logfiles show anything useful for this?

Lineflyer:
No, while looking for it I only discovered #8646 which looks unrelated.

It behaves like an ANR, meaning the app is unresponsive during that time.

@MagpieFourtyTwo Can you judge whether c:geo hangs in that waiting time or is just busy doing something but reacts to e.g. back key.

MagpieFourtyTwo:
> Can you judge whether c:geo hangs in that waiting time or is just busy doing something but reacts to e.g. back key.

The answer is a definite "it depends".  But I don't really know on what ...  :/

[...]

PR #8958, opened 2020-09-05
Related to #8632: add verbose logging, remove SaveMode from OfflineLogging

eddiemuc:
First steps to analyze and solve #8632 

Does two things:
* removes duplicate "save" on saving offline log (by removing the "SaveMode" logic)
* adds verbose log information on how often savelog is called (and how much time it takes) as well as whether the class `androidx.appcompat.widget.ShareActionProvider` is accessible


eddiemuc:
For testing

[log.properties.txt](https://github.com/cgeo/cgeo/files/5178101/log.properties.txt)

eddiemuc:
@moving-bits would be good to have this in master

PR #8960, opened 2020-09-06
add popup message explaining the offline finds counter

### LINKED ISSUE #8938: Explain how to find Offline logs
I've just updated to the latest version [ 2020.09.02-RC ] and I see that the count of offline logs are now shown on the first screen. Mine says (11172 + 7).
But I didn't know I had any offline logs - how do I find them to see if they are mistakes or logs I've forgotten to upload.

some way to find Offline logs - maybe this exists and I just can't find it.

**Additional context**
As above, mine says (11172 + 7), according to the web site I do have 11172 finds and when I try to log a new find on cgeo I get find 11180 in my auto-generated siganture.

fm-sys:
fixes #8938

I created a small framework which could be used for other one-time messages as well. The state if the message was already shown is stored in the `ExtentionsDB`. This framework could also be used for showing introduction messages for new users, new important functions to be highlighted or other stuff like that... just thinking ;-) 

main functions:
```
public static DialogStatus getStatus(final DialogType dialogType, final DialogStatus defaultStatus) { }

public static void setStatus(final DialogType dialogType, final DialogStatus status) { }
```

`DialogType` and `DialogStatus` are two enum's which could be amend later on...

---

Screenshot_20200906_155303_cgeo geocaching (Phone)
(of course this message get's only shown one single time)

I tested it quite a lot so no bugs should be expected...

If this is ready to merge textual and codewise, a would like to add a german translation to `strings-de.xml` to have a translation for the country with most of our users.  (Which should be overwritten at master by Crowdin without merge conflicts I hope)


fm-sys:
retest this please

moving-bits:
I like this idea of having such a framework. 👍 

One suggestion: Can you add a "remind me later" button which would close the dialog, but not mark the message as read. (This should probably the default button / "positive" button, whereas the "Ok" could be the "neutral" button, which is relevant for the order in which the buttons get displayed.)

Reasoning behind that is, that sometimes you just don't have the time to read and follow the instructions at the time the program displays such a message, and then it would be helpful to have this message being displayed again. (Whether this is "next day", "next restart of c:geo" or by using a separate "show pending messages" I've no preferences currently.)

What do you think?

fm-sys:
> Reasoning behind that is, that sometimes you just don't have the time to read and follow the instructions at the time the program displays such a message, and then it would be helpful to have this message being displayed again. (Whether this is "next day", "next restart of c:geo" or by using a separate "show pending messages" I've no preferences currently.)
>
> What do you think?

"next restart of c:geo" is the easiest for implementing and testing ;-) 
I will add it...

fm-sys:
@moving-bits for some reason this line `final ArrayList <DataStore.DBExtension> elements = getAll(type, null);` returns an ArrayList of size 0 instead of all elements of this type (which I can find when analysing the DB). 

Can you give me a hint?

moving-bits:
> `getAll` is a function of the Extensions Table. Can you give me a hint why searching without a key does not work?

Strange. This case has been explicitly being taken care of in `getAll` (but not yet in `load`, in case that's relevant). Can you provide me a test db?

fm-sys:
[cgeo.zip](https://github.com/cgeo/cgeo/files/5195608/cgeo.zip) - The database is quite empty, but the extentionsDB contains the "EXPLAIN_OFFLINE_FOUND_COUNTER" key which should be returned by the `getAll` method. But probably I will avoid using the `getAll` method at all. That's the easiest for now... But thanks for looking into this!

fm-sys:
@Lineflyer Can you cross read the information popup string? 
If it's OK I will add a hard coded German string which should be overwritten by Crowdin at master (I hope). So that most of our users will have an translation... 

Offtopic: How many of our users are actual German?

main/res/values/strings.xml
```diff
@@ -262,6 +262,7 @@
     <string name="info_select_logimage_cancelled">Image selection or capture was cancelled.</string>
     <string name="info_stored_image">New image saved to:</string>
     <string name="info_cache_saved">The cache has been stored locally</string>
+    <string name="info_feature_offline_counter">You have offline logs, which are not yet uploaded. Therefore a counter (eg. \"+ 1\") will be displayed. If you don\'t like the counter, you can disable it under settings > logging.\n\nTo easily find remaining offline logs, goto \"history list\" and filter for \"With personal data\" > \"With offline log\"</string>
```
Lineflyer, line 265: The menu name should IMHO be capitalized: Settings and Logging. Else its fine.

bekuno:
> If it's OK so far I will add a hard coded German string which should be overwritten by Crowdin at master (I hope). So that most of our users will have an translation...

You could also pick the other translations back from crowdin.

fm-sys:
> You could also pick the other translations back from crowdin.

see https://github.com/cgeo/cgeo/issues/8938#issuecomment-686426416:
> > We would need to cherry pick and do conflict resolution, which is not that easy. I would prefer not doing this, as last time it totally messed up.

We decided not doing it, because it could lead to some extensive merge conflicts... If there is only one language, the effects can still be kept within limits ;-)

@moving-bits: By the way, this should be ready to merge now...

moving-bits:
> @moving-bits: By the way, this should be ready to merge now...

I had running this in the emulator locally already, works fine here, so let's merge.

Regarding the db query: Opening the file in HeidiSQL and running the SQL directly works (as expected), so this seems to lay somewhere in the `db.query` code. Haven't had the chance to look into this yet, though.

bekuno:
@fm-sys 
To avoid conflicts the string should go to the same possition in the strings.xml as in the version from crowdin.
And, the text has to be the same.
The second is also necessary for the german string. Maybe it is better to wait for the crowdin translation.

alan666notb:
Is it possible to add a "See them now" button?
i.e. go to the history list with the correct filter already set up.

PR #8961, opened 2020-09-06
NoSuchMethodException for ShareActionProvider - Error in #8632

bekuno:
some variants trying to solve #8632

variant 1 - proguard rule
variant 2 - add appcompat-resources
variant 3 - latest appcompat dev-version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,10 +87,10 @@ allprojects {
             resolutionStrategy {
                 componentSelection {
                     all { ComponentSelection selection ->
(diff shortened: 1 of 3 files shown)


bekuno:
see
https://github.com/cgeo/cgeo/issues/8632#issuecomment-686935360
~~~
Cannot instantiate class: androidx.appcompat.widget.ShareActionProvider
java.lang.NoSuchMethodException: androidx.appcompat.widget.ShareActionProvider.<init> [class android.content.Context]
~~~
the error seems not reletad to the initial problem in #8632 

The PR should fix this by adding a proguard rule

PR #8971, opened 2020-09-07
Fix #8845: Enable parsing for Locus Map coordinate format

### LINKED ISSUE #8845: Check coordinate parsing for the format used by Locus Map
Taken from a discussion in https://github.com/cgeo/cgeo/pull/8597#issuecomment-676570462 an follow-up article by @baiti:

>> We now forbid all special characters of the coordinates to occur between the coordinate parts.

> This is not good. The waypoint dialog's coorindate entry pull down provides a function to take coordinates from the clipboard. A very popular application that can put coordinates into the clipboard is Locus Map, however, Locus Map uses the format:
>
> Nxx° yy.zzz', Exx° yy.zzz'
> 
> This format was parsed by c'geo correctly until the referenced change was made. Now since the colon between Lat / Lon part makes the parser  reject these coordinates, the effect is that the menu entry "Use coords from clipboard" is not even displayed anymore.
>
> This change breaks coordinate parsing when coordinates come from Locus Map via the clipboard. Please backout this change. Coordinates in the form of "Nxx° yy.zzz', Exx° yy.zzz'" should be parsed corrextly as valid coordinates. The only other option I see is to convince the Locus Map folks to remove that colon as it is "ugly" in the first place and doesn't really make sense, however, as a matter of fact, that colon happens to be there since a very long time and c'geo was indeed able to process it correctly

UniQP:
We now separately parse the coordinates for each supported delimiter instead of normalizing the delimiters and parsing the coordinates once. Thus, we support parsing of coordinates with separators that contain other delimiters, e.g., the Locus Map format: Nxx° yy.zzz', Exx° yy.zzz'.

### DIFF
--- a/tests/src/cgeo/geocaching/location/GeoPointParserTest.java
+++ b/tests/src/cgeo/geocaching/location/GeoPointParserTest.java
@@ -277,7 +277,7 @@ public class GeoPointParserTest {
         GeopointParser.parse("N52 36.123 E010 06.456");
         GeopointParser.parse("52° 10°");
         GeopointParser.parse("52° -10°");
-        GeopointParser.parse("52.55123 10,56789");
+        GeopointParser.parse("52,55123 10,56789");
         GeopointParser.parse("52.55123° 10.56789°");
     }
 
@@ -320,6 +320,13 @@ public class GeoPointParserTest {
     @Test
     public void test8589() {
         assertParsingFails("6, 12, 16, 29");
+    }
+
+    @Test
+    public void test8845() {
+        final Geopoint ref = new Geopoint("N", "49", "56", "31", "E", "8", "38", "564");
+        final Geopoint point = GeopointParser.parse("N49° 56.031', E08° 38.564'");
+        assertThat(point).isEqualTo(ref);
     }
 
     public void parseMultipleCoordinatesWithCorrectStartEndPositions() {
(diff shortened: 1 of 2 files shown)


bekuno:
This fix should go to release, I would say.

tests/src/cgeo/geocaching/location/GeoPointParserTest.java
```diff
@@ -277,7 +277,7 @@ public void test5790() {
         GeopointParser.parse("N52 36.123 E010 06.456");
         GeopointParser.parse("52° 10°");
         GeopointParser.parse("52° -10°");
-        GeopointParser.parse("52.55123 10,56789");
+        GeopointParser.parse("52,55123 10,56789");
         GeopointParser.parse("52.55123° 10.56789°");
```
bekuno, line 281: We need also a test for something with "," between N and E part, like
`Nxx° yy.zzz', Exx° yy.zzz'`

UniQP: That's covered by test8845 added below.

bekuno: oh, yes.

UniQP:
@bekuno Retargeted the PR to release branch.

PR #8993, opened 2020-09-11
add splash logo (fix #8988)

### LINKED ISSUE #8988: Splash Screen while starting c:geo
Opening c:geo on low budget devices can take 1-2 seconds without any visual feedback, so we should integrate a splash screen which is displayed while android loads the first activity.
Currently, `windowDisablePreview` is (probably) set, that's why we don't see a just black or white empty screen while c:geo is starting up (like in some other apps). 
However, we can also display the c:geo logo on startup, like it is also done by whatsapp or the official gs geocaching app.

Approach is explained here:
https://developer.android.com/topic/performance/vitals/launch-time#themed
https://android.jlelse.eu/the-complete-android-splash-screen-guide-c7db82bce565

Sidenote: Of course, I don't want to prolong the start unnecessarily, but just "beautify" the waiting time that is already there...

moving-bits:
Tries to implement the solution suggested in #8988.
Cannot test here, as even the emulator seems to be fast enough to not show the splash logo, but maybe someone with a slower device can test it?


bekuno:
This would be interesting in the legacy banch.

fm-sys:
It is working partly...

The splash logo is shown when you click on open at the "install was successful" dialog and when restoring a settings backup within c:geo *but not* while starting c:geo from Launcher. (That's probably why you did not saw the splash screen)

The logo needs higher resolution IMHO (see Screenshot attached):
Screenshot_20200912_003526_cgeo geocaching

Also I would use a dark gray (mabe the action bar color?) instead of pure black for the background. And there is a small white line on the left side of the logo...

moving-bits:
@fm-sys  Thanks for testing!

> The splash logo is shown when you click on open at the "install was successful" dialog and when restoring a settings backup within c:geo _but not_ while starting c:geo from Launcher. (That's probably why you did not saw the splash screen)

Strange. I added it to the `MainActivity` which is what I thought gets called on launching c:geo.
Any other activities I should add this behavior to?

> The logo needs higher resolution IMHO (see Screenshot attached):
> Screenshot_20200912_003526_cgeo geocaching

Yes, just made a q&d version of the image. Have to try yet if we can use a scalable image (SVG) here.

> Also I would use a dark gray (mabe the action bar color?) instead of pure black for the background.

I'd set the background to transparent to have it the same as it is on c:geo mainscreen. Do you have an all-black background on the device you've tested this on?

fm-sys:
> Do you have an all-black background on the device you've tested this on?

Yes, but even with changing my background to another image the splash screen stays black. Looks like a transparent background isn't supported...

Lineflyer:
I would set it to a dark background by default and white background if the white theme setting is active.

Lineflyer:
Would still be a nice addon...any idea why its not working on normal startup?

SammysHP:
Can we use the "misty forest" image that we have in our website design and play store?

moving-bits:
> Can we use the "misty forest" image that we have in our website design and play store as a background?

As this splash screen is mainly intended to give a visual feedback during longer load times, and those (should) happen only on less-capable devices, I would keep the splash screen as lightweight as possible. Also c:geo logo is easier to recognize than a complex image, which would help if that splash screen actually just splashes. I have something in mind like that green messenger does - a blank screen with a tiny logo in the center.

Aside from that, we need to find out first, why the splash screen is shown currently only under special circumstances.

moving-bits:
@fm-sys 
I've created an alternative implementation which uses its own little `SplashActivity` that does nothing more than displaying the image and loading our original `MainActivity`. Image etc. is unchanged.

Can you cross-check if that way of implementing does work for you? (I still cannot see a splash screen in the emulator, but before I explicitly `finish()`ed the activity, I could back-step to it, so it gets called.)

PR is #9291

fm-sys:
> I still cannot see a splash screen in the emulator, but before I explicitly `finish()`ed the activity, I could back-step to it, so it gets called.

I cannot see a splash screen either, except the very first launch after installing it... But I have really no idea why the splash screen isn't called :-/

Lineflyer:
Any help in here: https://stackoverflow.com/questions/22679535/splash-screen-not-displaying-in-android

fm-sys:
Looks like the start main activity intent needs to run asynchronous via handler...

moving-bits:
> I cannot see a splash screen either, except the very first launch after installing it... But I have really no idea why the splash screen isn't called :-/

Sad to hear, but thanks for testing 👍

moving-bits:
> Any help in here: [stackoverflow.com/questions/22679535/splash-screen-not-displaying-in-android](https://stackoverflow.com/questions/22679535/splash-screen-not-displaying-in-android)

Thanks for the link. If I understand the example correctly, they introduce an active 3 second-delay to make the splash screen seen. That's not necessarily something we want to do, as we want the splash screen to be shown mainly for slow devices, so something like "load splash screen as fast as you can and display it, then load MainActivity asynchronously and start it when you are ready".

moving-bits:
Superseeeded by #9291, therefore closing this one

Issue #8818, opened 2020-08-17
Significant amount of users report login problem "Username/Password incorrect"


Lineflyer:
Over the last days and weeks (especially this week) we get significant amount of people claiming login problems with reason
"Username and/or password incorrect" although the users doublechecked and credentials should be correct.

Its not like a reproducible problem or a problem which affects all users, still unusual and I cannot explain it.

I will monitor the situation and try to get more clues.

Question:
What was the latest change we did here? Do we still explicitly check for a matching string to identify password problems or could it be a wrong error message and some other login problem applies?

moving-bits:
The last change I can remember was changing the string for login error.

We still do check for a certain string:
```
if (loginData.contains("Your password or username/email is incorrect")) {
    Log.i("Failed to log in Geocaching.com as " + username + " because of wrong username/password");
    return resetGcCustomDate(StatusCode.WRONG_LOGIN_DATA); // wrong login
}
```

Problem may be, that this string is included in the login page itself, without even having tried to login:
```
<script type="text/javascript">
    var resources = {
        "changePasswordSuccess": "You have successfully changed your password."
    }
    var errorResources = [{"emailEmpty":"Your email cannot be empty","emailInvalid":"Your email must be in a valid 
format","emailTaken":"This email is already in use.","passwordEmpty":"Password is required","passwordInvalid":"Your 
password must be at least 5 characters","signinError":"Your password or username/email is 
incorrect","usernameEmpty":"Username is required","usernameInvalid":"Username is invalid","usernameTaken":"The 
selected username already exists."}][0];
</script>
```

Thus, if login fails for whatever reason (`getLoginStatus()` returns `false`) it reaches the check above, finds the text copied beneath, and reasons that login has failed due to "Your password or username/email is incorrect".

But that's only a guess from a short glance at `connectors/gc/GCLogin.java` - would be nice if someone finds the time to debug into this.

moving-bits:
> Over the last days and weeks (especially this week) we get significant amount of people claiming login problems with reason "Username and/or password incorrect" although the users doublechecked and credentials should be correct.

Does this get reported from users of our nightly versions as well? Then we could add some additional debug log infos to get a grasp on it.

Lineflyer:
No, only release users.

But indeed you conclusion could be right.
Let me check the support system, whether there was any user at all having the "unknown reason" problem. From what I remember it could indeed be, that all what was "unknown error" before now returns the password error.

[...]

PR #9005, opened 2020-09-12
add logWriter utility class for separate and extended logfiles

moving-bits:
Since logcat can hold strings up to a certain length only, it is not possible to log whole HTML pages returned by gc.com during the login process.

This PR adds a small logging utils class to log into separate files. Can be combined with the regular `Log` functions in c:geo: use `logWriter.d()` instead of `Log.d()` etc., or `logWriter.log()` to log to separate file only.

~PR also adds specific logging for the login process to (hopefully) catch the differences sometimes happening during login at gc.com. Tries to support research for #8818.~

### DIFF
--- a/main/src/cgeo/geocaching/utils/LogWriter.java
+++ b/main/src/cgeo/geocaching/utils/LogWriter.java
@@ -1,0 +1,71 @@
+package cgeo.geocaching.utils;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+
+public class LogWriter {
+    private PrintWriter logWriter = null;
+    private String prefix;
+
+    public LogWriter(final String prefix) {
+        this.prefix = prefix;
+        checkLogfile();
+    }
+
+    public void log(final String info) {
+        checkLogfile();
+        if (null != logWriter) {
+            logWriter.println("--------------------------------------------------------------------");
+            logWriter.println(CalendarUtils.formatDateTime("yyyy-MM-dd HH:mm:ss.SSS"));
+            logWriter.println(info);
+        }
+    }
+
+    public void d(final String info) {
+        Log.d(info);
+        log(info);
+    }
+
+    public void e(final String info) {
+        Log.e(info);
+        log(info);
+    }
+
+    public void i(final String info) {
+        Log.i(info);
+        log(info);
+    }
+
+    public void v(final String info) {
+        Log.v(info);
+        log(info);
+    }
+
+    public void w(final String info) {
+        Log.w(info);
+        log(info);
+    }
+
+    public void close() {
+        if (null != logWriter) {
+            log("end of logging");
+            logWriter.println("--------------------------------------------------------------------");
+            logWriter.close();
+            logWriter = null;
+        }
+    }
+
+    private void checkLogfile() {
+        if (null == logWriter && Log.isDebug()) {
+            final File logfile = FileUtils.getUniqueNamedLogfile(prefix, "txt");
+            try {
+                logWriter = new PrintWriter(logfile);
+                log("begin logging to file " + logfile.getPath());
+            } catch (FileNotFoundException e) {
+                // ignore error
+            }
+        }
+    }
+
+}


moving-bits:
This is currently for manually testing for #8818 only, therefore setting the "Do not merge" label for now.

main/src/cgeo/geocaching/utils/LogWriter.java
```diff
@@ -0,0 +1,71 @@
...
+            logWriter = null;
+        }
+    }
+
+    private void checkLogfile() {
+        if (null == logWriter && Log.isDebug()) {
+            final File logfile = FileUtils.getUniqueNamedLogfile(prefix, "txt");
+            try {
+                logWriter = new PrintWriter(logfile);
+                log("begin logging to file " + logfile.getPath());
```
eddiemuc, line 64: You probably meant to write "Log.i" here? Because "log" logs to the newly created file, and when you are there the file path is known anyway

main/src/cgeo/geocaching/utils/LogWriter.java
```diff
@@ -0,0 +1,71 @@
...
+        if (null != logWriter) {
+            log("end of logging");
+            logWriter.println("--------------------------------------------------------------------");
+            logWriter.close();
+            logWriter = null;
+        }
+    }
+
+    private void checkLogfile() {
+        if (null == logWriter && Log.isDebug()) {
```
eddiemuc, line 60: Since "checkLogfile" is called on every call to "LogWriter" again: if an exception is catched later (and logWriter stays null) then this check will be repeated again and again, and on every call to LogWriter it will again try to instantiate the logWriter. Maybe a flag should be used so trying to set up file once it failed is not repeated

main/src/cgeo/geocaching/utils/LogWriter.java
```diff
@@ -0,0 +1,71 @@
...
+        checkLogfile();
+        if (null != logWriter) {
+            logWriter.println("--------------------------------------------------------------------");
+            logWriter.println(CalendarUtils.formatDateTime("yyyy-MM-dd HH:mm:ss.SSS"));
+            logWriter.println(info);
+        }
+    }
+
+    public void d(final String info) {
+        Log.d(info);
```
eddiemuc, line 26: My understand is that LogWriter is for logs too long to be handled by logcat? In this case, why log the info additionally to logcat (it won't be able to handle it anyway). Instead, maybe a link to the LogWriter file should be logged.

moving-bits:
@eddiemuc Thanks for your feedback. I'll reply in the separate messages.

main/src/cgeo/geocaching/utils/LogWriter.java
moving-bits: This is mainly used for having a timestamp being logged, therefore calling the internal `log` instead of the "old" `log.xxx` method here. I agree, though, that the filename is not necessarily of use here.

main/src/cgeo/geocaching/utils/LogWriter.java
moving-bits: Doing the repeated call to 'checkLogfile()' is to catch if a user has switched on debug mode throughout the operation. I agree that this may be a rare case, as you can activate debug mode only using settings, which are reachable only if you close any other activity within c:geo, so this extra check can probably left out.

main/src/cgeo/geocaching/utils/LogWriter.java
moving-bits: Strings which may be too long for logcat should probably use `logWriter.log()` instead of the `.i` etc. variants. But even if not, the full string would be logged to the individual log file, whereas the string in logcat would be truncated.

Only those strings logged using one of the `logWriter.i()`, `logWriter.e()` etc. variants also log to the regular logcat. This is to have logcat unchanged compared to not using `logWriter`, and you can additionally turn on the separate logfile, if needed.

main/src/cgeo/geocaching/utils/LogWriter.java
eddiemuc: Ok understood

main/src/cgeo/geocaching/utils/LogWriter.java
eddiemuc: Ok understood

main/src/cgeo/geocaching/utils/LogWriter.java
eddiemuc: Ok understood

eddiemuc:
Ok it all makes sense now. Thanks for the explanations

moving-bits:
Divided the original PR into two separate parts. This PR now only includes the `LogWriter` class to support logfiles separate from system `LogCat`. Further analyses can be based upon this class.
I'll create a separate PR for extended logging for the login problems, once this PR is merged to master.

PR #9014, opened 2020-09-18
redesign onetime dialog and improve explanation text

### LINKED ISSUE #9013: [Nightly] Improve explanation text of "FoundCount +x" popup
Topic deals with the popup, which appears after the first start of c:geo after upgrading to the version which contains the "+x" behind the cache found count in the footer of c:geo's main screen: 

Popup

When I saw this popup the very first time, I thought it could be hard to understand esp. for somebody who never ever had seen the "+" before or even has paid attention to this counter at all. So I would be a bit more descriptive like:

```
You have some caches with an offline saved log,
which are not yet uploaded to geocaching,com.
For your comfort the count of these logs is now
displayed as "+x" behind your number of found
caches in the footer at c:geo's main page.

Moreover these "x" logs will be added to the
"[NUMBER]" template in your log signature.

If you don't like this new counter extension,
you can disable it under "Settings" => "Logging"
and/or use the "[NUMBER (without offline logs)"
template instead.

To easily find remaining offline logs, go to the
list "History" and filter for "With personal data
 => "With offline log".

[_] Don't show this message again     [OK]
```

As most of the users will not know what this will mean, I would add the ``Don't shown again`` marker.
This way they have the chance to see the counter at least once and then read again.

fm-sys:
fix #9013

New string (created from the comments in #9013):
Screenshot_20200918_171103_cgeo geocaching

Could someone please cross read if it sounds fine or any other changes are needed?

### DIFF
--- a/main/res/layout/dialog_text_checkbox.xml
+++ b/main/res/layout/dialog_text_checkbox.xml
@@ -29,5 +29,5 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginLeft="-5dp"
-        android:text="@string/dont_shown_again" />
+        android:text="@string/dont_show_again" />
 </LinearLayout>
--- a/main/res/layout/dialog_text_checkbox.xml
+++ b/main/res/layout/dialog_text_checkbox.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    android:paddingLeft="25dp"
+    android:paddingRight="25dp" >
+
+    <ScrollView
+        android:layout_width="fill_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1"
+        android:layout_marginBottom="10dp"
+        android:id="@+id/text_container">
+
+        <TextView
+            android:id="@+id/message"
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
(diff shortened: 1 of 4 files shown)


fm-sys:
I changed the second paragraph to the following (because many users will *not* have the NUMBER template in their signature):
Screenshot_20200918_192337

I have to fix one small bug that the checkbox gets hidden on small devices when the text is to long, but I should probably get this working until tomorrow, so I should be merged before the second beta IMHO...

Lineflyer:
Thats looks like wrong usage of "eventually" as it has a different meaning in english and german.

fm-sys:
So what about using "possibly" instead?

Lineflyer:
you might use in your....

fm-sys:
I just fixed the last bugs, now it should be ready to merge from my side. Thanks for all your help!

fm-sys:
Should I also add a German translation or do we want to merge this now and then wait for the Crowdin translation at master?

The new dialog:
Screenshot_20200920_201145
The text view will be scrollable if the screen is not large enough.

moving-bits:
> Should I also add a German translation or do we want to merge this now and then wait for the Crowdin translation at master?

Either way would be fine for me. 
BTW: There is a small typo "Don't shown again" (a "n" too much).

fm-sys:
> BTW: There is a small typo "Don't shown again" (a "n" too much).

Thanks for the hint! I missed that...


> > Should I also add a German translation or do we want to merge this now and then wait for the Crowdin translation at master?
>
> Either way would be fine for me.

Then let's merge without translations for now...

fm-sys:
@moving-bits just a quick reminder: please merge `release` into `master`, so that we can collect some translations for this until we merge from `master` again...

PR #9028, opened 2020-09-19
add warning about automatic waypoint extraction (fix #9026)

### LINKED ISSUE #9026: Mention "wp from notes will be recreated" in confirmation dialog
Taken from https://github.com/cgeo/cgeo/issues/8994#issuecomment-695202219 by @geocachermgo:

Remark from testing: Works a little strange if I have have "extract waypoint from note" active:
Everything is removed.
If you switch off "extract" and on again, the waypoints from the note are there again.
If you manually delete a waypoint, it is immediately recreated.
If the note is edited the re extract happens. If you reload the cache too.
Not if you open it for editing and cancel.
Options: 
1) Additionally switch off "extract" and note that in the warning
 2) Keep behavior but note that immediate re extract is blocked but will happen later
3) Allow immeadiate re extract (analogous to manual delete) and mention that in the warning " If you don't want to re-extract you have to switch-off extract or clear note"

My preference would be 3) followed by 1) 2)

moving-bits:
Confirmation dialog extended, if "prevent waypoint extraction" is NOT set for this cache:

image

@eddiemuc 
Or is there an easy way to detect programatically if there would be any waypoints detected from personal note?
Then I would add this check and only show the extended info when "prevent waypoint extraction" is not set AND there would be waypoints detected/created from PN.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -871,6 +871,7 @@
     <string name="cache_offline_drop">Remove</string>
     <string name="cache_delete_userdefined_waypoints">Remove user-defined waypoints</string>
     <string name="cache_delete_userdefined_waypoints_confirm">Remove all user-defined waypoints of this cache?\n\nWaypoints will be deleted immediately, there is no undo.</string>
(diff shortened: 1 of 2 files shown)


eddiemuc:
Should be relatively easy to implement such a check using the parsing function for waypoints in text (I think it is in class GeopointParser or something like this)

PR #9034, opened 2020-09-21
fix #8970: remove title prefix

### LINKED ISSUE #8970: What is the image title prefix?
Out on vacation I tried the new logging sheet with its image feature.

I however got no clue what the "Image title prefix" is. It does not seem to be used on upload.

I am afraid many people will confuse it with the title as it is shown very prominent. If its only informative, why not use the title here?

eddiemuc:
fix #8970: remove title prefix

### DIFF
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -46,8 +46,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.AsyncTask;
 import android.os.Bundle;
-import android.text.Editable;
-import android.text.TextWatcher;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuItem;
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -741,11 +717,7 @@ private String getImageTitle(final Image image, final int position) {
         if (!StringUtils.isBlank(image.getTitle())) {
             return image.getTitle();
         }
-        String titlePrefix = logImageTitlePrefix.getText().toString();
-        if (StringUtils.isBlank(titlePrefix)) {
-            titlePrefix = getString(R.string.log_image_titleprefix);
-        }
-        return titlePrefix + " " + (position + 1);
+        return getString(R.string.log_image_titleprefix) + " " + (position + 1);
```
Lineflyer, line 720: Why does this return still contain `log_image_titleprefix`?

eddiemuc: > Why does this return still contain `log_image_titleprefix`?

Images for which you don't set a caption in the "image" dialog will get the log_title_prefix plus an index as their name (e.g. "Image 1" or "Image 3"). Ths behaviour is similar to the naming of waypoints extracted from PN as "Personal Note 1" and similar (except for images you can explicitely choose a different name in the image dialog if you want)

Lineflyer: Thanks, now I got it. :)

Issue #9029, opened 2020-09-19
unnecessay line breaks in list "All" after "Offline"


geocachermgo:
Still some unnecessay line breaks in list "All" after "Offline". I would say that in case "Offline" should not be there, because there are other lists given.
![IMG_20200919_162811](https://user-images.githubusercontent.com/43813636/93669552-469d3380-fa95-11ea-8dbc-0be4b2e0b552.jpg)
![IMG_20200919_162633](https://user-images.githubusercontent.com/43813636/93669565-63d20200-fa95-11ea-921e-ced7ba4629b8.jpg)

_Originally posted by @geocachermgo in https://github.com/cgeo/cgeo/issues/9007#issuecomment-695220604_

moving-bits:
Just for documentation: Already discarded ideas: (see discussion in #9007)
- abbreviate "Premium" to "$$" or "Prem."
- use symbols instead of textual information
- merge the two info lines into one
- leave out "Offline" completely, if there are further offline lists

---

Current point in discussion: (copied from https://github.com/cgeo/cgeo/issues/9007#issuecomment-695756167)
> I would leave all of this information unchanged. I personally dont like apps which use abbreviations I might not understand.

Generally agreed to. One tiny idea to save some space: We could skip the "Offline" textual info if we add the "stored" symbol (floppy disc) to the cache icon. This would be the same symbol as is used on the map.

Lineflyer:
> Generally agreed to. One tiny idea to save some space: We could skip the "Offline" textual info if we add the "stored" symbol (floppy disc) to the cache icon. This would be the same symbol as is used on the map.

I dont't fully understand the discussion. We came from a good working implementation until the new feature was merged to show long titles in two lines.
There seem to be several regressions caused by that, so either we just solve these regressions (without cutting off older features, working modes, strings) or we might also decide to revert this new feature (I can also live with tilting my phone or opening cache details in case of long names.

bekuno:
The "Stored" symbol is already used in the "History" list.
I think that using it instead od the "Offline" text is fine and worth to become a general use.

geocachermgo:
> I dont't fully understand the discussion. We came from a good working implementation until the new feature was merged to show long titles in two lines.
> There seem to be several regressions caused by that, so either we just solve these regressions (without cutting off older features, working modes, strings) or we might also decide to revert this new feature (I can also live with tilting my phone or opening cache details in case of long names.

The problem was, that the cache names were truncated.
[...]

PR #9035, opened 2020-09-21
widen and shorten found marker line in cache list

moving-bits:
Following the discussion in https://github.com/cgeo/cgeo/issues/9029#issuecomment-696053797:
- double the width of the found marker line from 2dp to 4dp
- add a bottom margin of 10dp to more clearly separate this entry from the next

|Before|After|
|------|-----|
|image|image|

### DIFF
--- a/main/res/layout/cacheslist_item.xml
+++ b/main/res/layout/cacheslist_item.xml
@@ -22,9 +22,11 @@
     <!-- "found" marker, vertical green/red line -->
     <ImageView
         android:id="@+id/log_status_mark"
-        android:layout_width="2dip"
+        android:layout_width="4dip"
         android:layout_height="match_parent"
         android:layout_marginRight="4dip"
+        android:layout_marginTop="5dip"
+        android:layout_marginBottom="5dip"
         android:scaleType="centerCrop"
         android:gravity="left|center_vertical"
         android:src="@drawable/mark_transparent"


bekuno:
I like the change.
Only 2 cent extra: what about make the line symmetric (5 top+bottom)?

Lineflyer:
> Only 2 cent extra: what about make the line symmetric (5 top+bottom)?

That was also my first thought when looking at the "after" screenshot. Maybe even 10 top / 10 bottom?

moving-bits:
This would be the results:


|5/5|10/10|
|------|-----|
|image|image|

I would say 5/5 is the way to go, as it neatly aligns with the icon overlays.

(Although there is no extra passing for the text line, there seems to be some implicit padding added, maybe font-specific, which makes aligning with the text difficult.)

Lineflyer:
My vote: 10/10 but as the majority decides...

MagpieFourtyTwo:
> I would say 5/5 is the way to go, as it neatly aligns with the icon overlays.

Same for me. Looks more harmonious. But in fact it's a matter of taste.
Way more important is the increase of readability.
While I would even spend another pixel in width ...  ;)

bekuno:
Im undecided.
Then go in the middle with 7?

MagpieFourtyTwo:
Ok .. let's haggle. Say you open at one shekel, ahem, pixel. I start at two thousand. We close about eighteen hundred ... ;)

PR #9036, opened 2020-09-21
show floppy symbol instead of text offline in cache lists

moving-bits:
Follow-up to the discussion in #9029:
Replace the text "Offline" by floppy-disk overlay on the cache icon

|Before|After|
|------|-----|
|image|image|

### DIFF
--- a/main/src/cgeo/geocaching/utils/MapMarkerUtils.java
+++ b/main/src/cgeo/geocaching/utils/MapMarkerUtils.java
@@ -413,7 +413,7 @@ public final class MapMarkerUtils {
      *         True if the floppy overlay should be displayed
      */
     private static boolean showFloppyOverlay(@Nullable final CacheListType cacheListType) {
-        return cacheListType != CacheListType.OFFLINE; // also covers null check
+        return cacheListType != null;
     }
 
     private static void readLists() {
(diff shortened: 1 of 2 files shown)


Lineflyer:
Looks good!

And thanks for these before/after screenshots, very helpful to discuss and agree the change.

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -413,7 +413,7 @@ private static boolean showUserModifiedCoords(final Geocache cache) {
      *         True if the floppy overlay should be displayed
      */
     private static boolean showFloppyOverlay(@Nullable final CacheListType cacheListType) {
-        return cacheListType != CacheListType.OFFLINE; // also covers null check
```
Lineflyer, line 416: Just an unqualified question looking at the comment:
What about a NULL case?

moving-bits: Ah, I simply erased the condition and always returned true for testing purposes. When could a `null` value happen - no list assigned? Is this a possible use case? (Search by address from mainscreen, maybe?)

Lineflyer: > Search by address from mainscreen, maybe?

Yes, maybe all the list generated from on demand like search results or lists of live map caches. But I am not deep enough into the code to judge....just guessing, why sometime in the past someone added this comment.

moving-bits: added a null case check

MagpieFourtyTwo:
Great stuff!

MagpieFourtyTwo:
Just installed the current nightly - and really like the list layout improvements.

Thx a lot!

PR #9039, opened 2020-09-22
Properly detect failed signin because of invalid credentials in GC connector

### LINKED ISSUE #8818: Significant amount of users report login problem "Username/Password incorrect"

Lomanic:
See https://github.com/cgeo/cgeo/issues/8818#issuecomment-696409866
The "Your password or username/email is incorrect" string is always there and dynamically added to the div
with id #signup-validation-error via client-side JS, under the condition that this div is present in the page.

Fixes #8818

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCLogin.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLogin.java
@@ -199,7 +199,7 @@ public class GCLogin extends AbstractLogin {
                 return completeLoginProcess();
             }
 
-            if (loginData.contains("Your password or username/email is incorrect")) {
+            if (loginData.contains("id=\"signup-validation-error\"")) {
                 Log.i("Failed to log in Geocaching.com as " + username + " because of wrong username/password");
                 return resetGcCustomDate(StatusCode.WRONG_LOGIN_DATA); // wrong login
             }


Lineflyer:
add to whitelist

moving-bits:
@Lomanic 

Thanks for trying to tackle that issue.
To my understanding comparing against the new string wouldn't work, as the `<div>` element is not part of the HTML returned, but inserted dynamically via Javascript, therefore c:geo will not be able to compare against it.
At least those are my observations from yesterday with the logfiles created with #9025 APK.

```
<div id="signup-validation-error" class="alert alert-error" role="alert" aria-live="assertive">
Your password or username/email is incorrect</div>
```

Lomanic:
https://github.com/cgeo/cgeo/issues/8818#issuecomment-696409866

> This can be confirmed by accessing (GET) the login page (this element is not present) and submitting wrong credentials (POST) (this element is in the response page), even with curl (no JS).

Here's a turn-key little script for you to verify this @moving-bits 

```bash
RequestVerificationToken=$(curl --silent --cookie cookie.txt --cookie-jar cookie.txt https://www.geocaching.com/account/signin | tee /dev/stderr 2>login.html | awk -F'"' '/"__RequestVerificationToken"/ { print $12 }')
grep "id=\"signup-validation-error\"" login.html # no match expected here
curl --silent --cookie cookie.txt --cookie-jar cookie.txt https://www.geocaching.com/account/signin --data "__RequestVerificationToken=$RequestVerificationToken&ReturnUrl=&UsernameOrEmail=aa&Password=aa" | grep "id=\"signup-validation-error\""
```

moving-bits:
@Lomanic 
Just retested this within the emulator with extra logging of #9025 enabled. Seems that I have been wrong, had somehow missed that string in the output. Sorry for the delay & thanks for preparing the PR.

Lomanic:
I'm glad I could finally contribute back some code to the project :)

PR #9040, opened 2020-09-22
sync flag for userdefined waypoint with GSAK

murggel:
GSAK does not recognize with the current export, that a waypoint is user-defined. GSAK has an extension-element which indicates the user-flag.
This PR writes the element to the gpx.file and reads it on import to set the flag for user-defined waypoint.


moving-bits:
ok to test

moving-bits:
@murggel 
Thanks for your contribution!

Issue #9049, opened 2020-09-23
Include even more info into sysinfo


Lineflyer:
When trying to help users on support I do often miss the following config information, as these are typical problems:
- Current value of global cache type filter (from lower right main menu item)
- Current state whether live mode of map is enabled or not
- Current path set for offline maps
- Current path set for offline themes
- Current path set for GPX import
- Current path set for GPX export
- State of backup (existing, date/time stamp)

SammysHP:
We could include the whole configuration (excluding credentials) to cover all c:geo related information and in addition required system information.

moving-bits:
> We could include the whole configuration (excluding credentials) to cover all c:geo related information and in addition required system information.

I would stay with a selected subset in the system view, as displaying the whole set makes it hard to find the details needed. We can still ask the user to backup the settings (without credentials) and send it to support, if needed.

fm-sys:
> State of backup (existing, date/time stamp)

Please wait with the backup state as there will be lager code changes related to #8959...

moving-bits:
> > State of backup (existing, date/time stamp)
> 
> Please wait with the backup state as there will be lager code changes related to #8959...

That's why I had excluded this part from #9055.

PR #9055, opened 2020-09-23
add more system info details (related to #9049)

moving-bits:
add a couple of additional info to about - system information:
- GPX import / export dir
- map file / map theme dir
- live map mode
- global filter mode

image


Lineflyer:
retest this please

Lineflyer (approved):
Thanks @moving-bits

Issue #7953, opened 2019-11-20
Zoom level changing when switching Google/OSM


Lineflyer:
**Describe the bug:**
Some user complaints, that when they switch between Google and OSM/Offline the zoom and/or position of the map are not kept.

**To Reproduce:**
1. Have Google Map selected
2. Open a cache list
3. Map the list
4. Change map source to OSM

**Actual behavior/state after performing these steps:**
Shown area on OSM is significantly higher (lower zoom) than the previously shown on Google Maps

**Expected behavior/state after performing these steps:**
Viewport should be somewhat the same (+/-)

**Version of c:geo used:**
2019.11.11

**Is the problem reproducible:**
Yes 

**Additional context:**
Either it is related to GoogleMapsV2 or to the recent resolution optimization implemented by @jonas-koeritz

andrixnet:
I can confirm this. Really annoying at times.

Lineflyer:
+many from support channel.

There seem to be quite some users frequently changing between OSM and Google which is not really usable at the moment do to the large change of zoom.

jonas-koeritz:
I just worked on my Mapsforge Fragment and noticed that the scaling is off by the dp to px factor. Zoom levels can be mapped 1:1 between Google Maps and our MfMapView if we do this:

```
mapView.getModel().displayModel.setUserScaleFactor(Resources.getSystem().getDisplayMetrics().density)
```

This works at least for raster maps (OSM: Online), I haven't implemented offline maps in my Fragment yet but will report back once it's done.

jonas-koeritz:
Correction: I called that way too early in my Lifecycle, resulting in density defaulting to 1.0. Just calling `setUserScaleFactor()` at all (with 1.0f as factor) seems to resolve the problem for me.

Lineflyer:
I guess it might be easy to solve. This issue is a regression after changing to GMapsV2. So I assume the gmapsv2 code is missing this.

jonas-koeritz:
Can somebody post screenshots of the problem? Maybe I'm not facing the same
issue in my work.

Lars <notifications@github.com> schrieb am Do., 9. Jan. 2020, 19:06:

Birdy2404:
Osm Offline (1km):
![Screenshot_20200109-211924](https://user-images.githubusercontent.com/34218281/72101770-cc24ed00-3325-11ea-9c18-56ba49b15c8b.png)
OSM Online (1km):
[...]

PR #9057, opened 2020-09-24
Do not merge: Test: Remove zoom-level adaption mf<->google

rsudev:
Only to buidl a test version


Lineflyer:
As discussed in #7953 I will merge this experimental change.
Its a simple change, which can easily be reverted if it does not work.

In fact that way is much more easy than setting up a dedicated CI job, etc. to have keys in.

PR #9060, opened 2020-09-24
Focus text input in list creation dialog

### LINKED ISSUE #9059: Focus text input in list creation dialog
According to a user this seems to be a regression of one of the last releases:

- Open lists
- Select Manage Lists -> Create List

Result:
The dialog for entering the list name is opened but another click is needed before name input can be started

Expected:
Focus should already be on text input, so that user can directly start typing.

fm-sys:
fix #9059


Lineflyer:
CI fails for trackable test because this has been fixed on master only (possibly due to the fact, that we plan to merge branches again soon).

PR #9064, opened 2020-09-25
 Fix #7771: Support [TYPE] in signature

### LINKED ISSUE #7771: Support [TYPE] in signature
From support mail:
To add cache type for the log signatures so that it would say something like: Thanks to [OWNER]  for this [TYPE].

bmuzzin:
Add [TYPE] as a signature template, and also as a logging insertion option.

### DIFF
--- a/main/src/cgeo/geocaching/log/LogTemplateProvider.java
+++ b/main/src/cgeo/geocaching/log/LogTemplateProvider.java
@@ -4,6 +4,7 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.capability.ILogin;
+import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Trackable;
 import cgeo.geocaching.settings.Settings;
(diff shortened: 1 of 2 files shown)


bmuzzin:
Will reopen with style fixes.

SammysHP:
Hi @bmuzzin,

thank you for the pull request! To make things easier to understand (when reading the commit history), please reword your commit and PR message to something like `Fix #7771: Support [TYPE] in signature`. This will also close the issue automatically once the PR got merged.

Lineflyer:
ok to test

bmuzzin:
Okay, I've updated the commit comment as you suggested. I'll and start working from a different branch from now on... while I'm not new to git, I'm new to the whole github side of things, so excuse my ignorance on these sorts of things. Thanks!

Lineflyer:
retest this please

main/src/cgeo/geocaching/log/LogTemplateProvider.java
```diff
@@ -285,6 +286,17 @@ public String getValue(final LogContext context) {
                 return StringUtils.EMPTY;
             }
         });
+        templates.add(new LogTemplate("TYPE",  R.string.init_signature_template_type) {
+            @Override
+            public String getValue(final LogContext context) {
+                final Geocache cache = context.getCache();
+                if (cache != null) {
+                    final CacheType cacheType = cache.getType();
+                    return cacheType.pattern;
```
moving-bits, line 295: `CacheType.pattern` is the non-translatable string used by GPX files etc. Should use `CacheType.stringId` instead AFAICS, which is the translated type identifier.

bmuzzin: Looks like the translated stringId is private, but comes through translated with getL10n() - not sure what the function is named for? .. But seems to be used in many places to get the translated name.

fm-sys:
retest this please

main/src/cgeo/geocaching/log/LogTemplateProvider.java
moving-bits: @bmuzzin
Thanks for updating the PR.
BTW: "L10n" stands for "Localization" which starts with a "L", ends with a "n" and has 10 letters in between...

PR #9062, opened 2020-09-24
rename maps.api.key to maps.api2.key also in private.properties template

bekuno:
was forgotten when renaming in the keys.xml in abc8ad7dee9595e71dc5514678d6af2277500843,
see also https://github.com/bekuno/cgeo/blob/master/README.md#api-keys-installation

PR #9067, opened 2020-09-25
fix broken ci

eddiemuc:
CI breaks on CodeStyle check in file MfMapView for all pull requests currently.
(I do not know how one was able to introduce this due to Codacy...)

This very small PR intends to fix that, please merge as soon as possible...

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/MfMapView.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/MfMapView.java
@@ -86,7 +86,7 @@ public class MfMapView extends MapView {
     }
 
     public void setMapZoomLevel(final int zoomLevel) {
-        getModel().mapViewPosition.setZoomLevel((byte) (zoomLevel /* - 3 */));
+        getModel().mapViewPosition.setZoomLevel((byte) zoomLevel /* - 3 */);
     }
 
     public void zoomToViewport(final Viewport viewport) {


Lineflyer:
Fix was already implicitly prepared in #9061. But anyway. :)

eddiemuc:
> Fix was already implicitly prepared in #9061. But anyway. :)

@Lineflyer yes, I saw, but #9061 is labeled with "do not merge" and while this fix is not in master ALL PRs will fail. I figured we don't want to leave this state for too long...
Anyway, how was it even possible to merge a PR introduction such a problem? CI must have failed for this PR too after all.

fm-sys:
> Anyway, how was it even possible to merge a PR introduction such a problem?

Probably it was overlooked as we had CI fails caused of the Trackable change at the same time too...

eddiemuc:
> > Anyway, how was it even possible to merge a PR introduction such a problem?
> 
> Probably it was overlooked as we had CI fails caused of the Trackable change at the same time too...

I meant something different: when you issue a pull request, then two automated checks are performed: Codacy and a Jenkins build. If one of the two fails, you should be unable to merge to master. For the problem introduced in class MfMapView the Jenkins build failed - this is the reason why no PR came through afterwards.
But I am asking myself: how was it possible to introduce this problem into master in the first place, because the pull request with the problem change inside must have failed the CI when the corresponding PR was tried to get merged.

Lineflyer:
> : how was it possible to introduce this problem into master in the first place

It did this on purpose (knowing that extra parenthesis and out-commented code) was in there, to allow easier testing (PR build does not include GMaps key, thus no testing possible). In parallel I submitted a PR to be merged next day to remove the experimental change.

Lineflyer:
> If one of the two fails, you should be unable to merge to master.

This could be set up on Github, but I would rather leave it to the core team member finally merging it to judge, whether the fail has any importance. There might be reasons for CI failing unrelated to the PR.

eddiemuc:
> This could be set up on Github, but I would rather leave it to the core team member finally merging it to judge, whether the fail has any importance. There might be reasons for CI failing unrelated to the PR.

Hi @Lineflyer I agree that there has to be a core team able to overwrite safeguards in CI to be able to react quickly on problems. I also strongly believe that a functioning, trustworthy CI must be available at all times. More precise: a CI shall fail on a PR only on problems introduced with this PR and nothing else. (Also, a green CI must be a sure sign to PR issuer that he/she did not break something seriously, but this is maybe another topic). I also believe that ensuring such a CI must be highest priority of a project at all times.
IMHO with great power comes great responsibility: when bypassing CI rules, one must be extra careful not to introduce "flawed" CI for "normal" developers. Maybe something like this can be at least communicated via developer email? I spent a good hour yesterday searching in my PR for the reason of failed CI before I realized that this was an unrelated thing.

bekuno:
@eddiemuc 
That's why my comment here that other see the problem: https://github.com/cgeo/cgeo/pull/9062#issuecomment-698614794

PR #9044, opened 2020-09-22
TB logbook not parsed - adopt change at website

bekuno:
fix #9038

PR #9079, opened 2020-09-26
provide #9044 for legacy version (fix #9078)

### LINKED ISSUE #9078: Apply trackable website change on legacy
The fix and the related mocked data need to be updated also on legacy.

moving-bits:


Lineflyer:
checks passed on CI, so I guess its OK.

Issue #9015, opened 2020-09-18
Speed up filtering lists for caches not yet found


MagpieFourtyTwo:
**Is your feature request related to a problem?**
When being on the road I most of time would like to see only caches which are saved offline (for a good reason), but are not yet found (as I save found caches, too, to have them handy if s/one needs a TJ - and perhaps this saves online time, too ...?).

As the map still lacks the feature to filter in any way (see e. g. #7278 and #3346), I use list "All caches" and filter it for "Status" => "Not Found". This kind of works - at least most of the times (unless some crashes because of timeout), but it's annooooyingly slow ...

**Describe the solution you'd like**
Result of filter should appear in reasonable time (i. e. less than 1 sec., which is a [widely used quasi standard in GUI development](https://www.nngroup.com/articles/response-times-3-important-limits/)).

**Additional context**
An index á la ``CREATE INDEX in_caches_found ON cg_caches(geocode,found)`` may already solve the problem ...?

eddiemuc:
I see you begin to like the idea of INDEXES speeding up processing :-)
In this case, this might be not so easy though.  I checked the code, and unfortunately it seems to me that c:geo first loads all cache data for the whole list into memory and then applies filtering programmatically (someone please correct me if I misinterpret the code here...)

To further harden (or weaken) my assumption:
* Which part is the slow one? The loading of the "all" list or the filtering for "status" thereafter (or both)?
* Could you try to send me a logcat file with verbose settings on (like for the other issue using the log-properties.txt)? PLease do this after a fresh start of c:geo (so no caches are loaded from db to memory before). I write "try" because it might be that turning on this log level when loading the "all" list will create too many and long messages for logcat to handle.

MagpieFourtyTwo:
> I see you begin to like the idea of INDEXES speeding up processing :-)

Not really. Know about the advantages of indexes for nearly 30 years now ...  ;)
And when I first got in touch with them, they were even more important, as we had to cope with 33 M(!)Hz CPUs and 2 to 4 M(!)B of RAM. Although more would have been possible with i386, but was not really affordable. Moreover there were no such nice things like SQLite, but only plain Oracle (and a bit DB2) these days. Nevertheless, we managed storehouses with thousands of storage places and controlled dozens of pickers, packers and drivers, already using wireless mobile scanners to transfer and control progress of the orders given by our system ... and all this with hardware and cpu power, that is surpassed by any digital wrist watch nowadays. Yeah, these were the days ...! ;)

[...]

PR #9080, opened 2020-09-27
rel to #9015: impr in-memory-filter perf, adjusts datastore-logs, logs logcat command

eddiemuc:
related to #9015: This PR
* improves in-memory-filtering performance for large cache lists
* adjusts logging in class DataStore
* adds logging of logcat command issued to extract c:geo logs via Settings-> System
* adds a helper class to create/remove artificial cache test data into developer databases

### DIFF
--- a/main/src/cgeo/geocaching/filter/AbstractFilter.java
+++ b/main/src/cgeo/geocaching/filter/AbstractFilter.java
@@ -29,13 +29,19 @@ abstract class AbstractFilter implements IFilter {
 
     @Override
     public void filter(@NonNull final List<Geocache> list) {
-        final List<Geocache> itemsToRemove = new ArrayList<>();
+
+        //method must be performant when used with very large lists (e.g. 50000 elements) filtered into very short lists (e.g. 30 elements)
+        //be aware that "list" most likely is an ArrayList, thus "get(i)" is very performant but "remove" is definitely not -> don't use it!
+
+        final List<Geocache> itemsToKeep = new ArrayList<>();
         for (final Geocache item : list) {
-            if (!accepts(item)) {
-                itemsToRemove.add(item);
+            if (accepts(item)) {
+                itemsToKeep.add(item);
             }
         }
-        list.removeAll(itemsToRemove);
+
+        list.clear();
+        list.addAll(itemsToKeep);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


main/src/cgeo/geocaching/filter/AbstractFilter.java
```diff
@@ -29,13 +29,19 @@ protected AbstractFilter(final Parcel in) {
 
     @Override
     public void filter(@NonNull final List<Geocache> list) {
-        final List<Geocache> itemsToRemove = new ArrayList<>();
+
+        //method must be performant when used with very large lists (e.g. 50000 elements) filtered into very short lists (e.g. 30 elements)
```
okainov, line 33: Sorry for random comment, but shouldn't be this method performant in all the cases, including opposite one - filtering very large lists (50k) into a very large list (49k)?

fm-sys: It is performant in all the cases 😉

okainov: Just from the code, existing version must have been faster for my case. Current code is creating the copy-list with almost the same caches leading to some potential memory overhead and then clears and re-adds items again (so feels like ~3x overhead on the initial size of the list).
Previous version would have very short `itemsToRemove` (i.e. 1 element) and then iterate over the existing list once (for 1 element), shrinking it once... I'm not convinced new approach is faster in this case :)

eddiemuc: @okainav you are right, at least I should have tested this case too in unit tests, and also the comment is misleading.
In general I assume filtering large lists into large results should work fine performance-wise since addAll from arraylist to arraylist should be a single arraycopy underneath (performance-wise same thing happening with every single remove-call on such a list). Memory-wise I agree that temporary a second listsl of size 50000 has to be kept in memory, but this only adds 50000 pointers (not deep copies of the cache data) to existing 50000 complete caches stored in memory - so in relation it should not add much to the (huge!) Load of holding so many caches in memory anyway.

eddiemuc: @okainov I created PR #9090 for this

PR #9082, opened 2020-09-27
[cachelist] center vertically and keep found marker allways right

fm-sys:
- moves found marker to the total right also in selection mode
- align distance and favorite (& trackable) count vertically

| before | now |
|-|-|
| Screenshot_20200927_154633_cgeo geocaching (2) | Screenshot_20200927_160709_cgeo geocaching (2)  |

### DIFF
--- a/main/res/layout/cacheslist_item.xml
+++ b/main/res/layout/cacheslist_item.xml
@@ -8,17 +8,6 @@
     android:background="?background_color"
     tools:context=".ui.CacheListAdapter">
 
-    <!-- selection mode checkbox -->
-    <CheckBox
-        android:id="@+id/checkbox"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_gravity="left"
-        android:gravity="left|center_horizontal"
-        android:paddingRight="5dip"
-        android:visibility="gone"
-        tools:visibility="visible"/>
-
     <!-- "found" marker, vertical green/red line -->
     <ImageView
         android:id="@+id/log_status_mark"
@@ -33,6 +22,18 @@
         android:visibility="visible"
         tools:visibility="visible" />
 
+    <!-- selection mode checkbox -->
+    <CheckBox
+        android:id="@+id/checkbox"
+        android:layout_width="wrap_content"
+        android:layout_height="fill_parent"
+        android:layout_gravity="left"
+        android:gravity="left|center_horizontal"
+        android:paddingRight="5dip"
+        android:visibility="gone"
+        tools:visibility="visible"/>
+
+    <!-- cache icon -->
     <ImageView
         android:id="@+id/text_icon"
         android:layout_width="wrap_content"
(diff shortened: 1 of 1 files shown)


Lineflyer:
IMHO the marker is better to activate if it is not somewhere on the display but on the far left. My opinion.
The centering of distance and other fields however looks nice.

fm-sys:
From my point of view, the found marker is more like a colored border in the normal list view. If it moves in the middle for selection mode, it looks more like a design bug IMO.

But I will let the majority decide. Has anyone else an opinion?

okainov:
IMO both have some issues... Cannot really suggest the solution though..

**before**:
+ top alignment for almost all elements, eye scans the line normally-horizontally
-Distance and squared zero are not on one line
+Checkmark clearly indicates the element, checkmark and cache are visually separated which is good (because clicking on them makes different action, right? Click on checkbox will mark, click on cache will open the cache, correct?)
-Vertical alignment seems different for distance and the cache name

**after**:
-Colored line has no margin on the left and looks not professional
-No visual separation for checkbox and the cache
-Eye scan has to change lines in the middle (go down from cache name to distance)



Is the squared zero so important that we show it everywhere? I cannot even remember when I saw non-zero there... Cannot it go to small items under the title? Then the distance can be on the right without any other elements...

Lineflyer:
> 
> Is the squared zero so important that we show it everywhere? I cannot even remember when I saw non-zero there

It is the amount of favorite point / recommendations for platforms supporting it as e.g. geocaching.com and opencaching.*
And yes. IMHO its rather important for many people. 

Coming back to this PR:
For me the horizontal alignment of tbe new approach is fine.
Regarding the position of the selection square I have to admit, that the point of @fm-sys , that it looks like a design bug to move the green line, is indeed true. 
So I revert my first opinion and would love to see and test it as proposed here.

moving-bits:
I do like the vertical alignment of the two boxes at the right, and have no preference regarding the colored found marker. So let's merge and see how it works.

Regarding @okainovs questions:
> Is the squared zero so important that we show it everywhere? I cannot even remember when I saw non-zero there...

It's a checkbox to select one or multiple caches for further action (like export), and is only visible if you go into "selection mode" in cache list.

> Cannot it go to small items under the title? Then the distance can be on the right without any other elements...

Putting the checkbox below the cache icon would make it pretty hard to select.

okainov:
> It's a checkbox to select one or multiple caches for further action (like export), and is only visible if you go into "selection mode" in cache list.

I'm not so dummy, sorry =) I know what checkbox is. My question was about zero in the rectangle on the very right which @Lineflyer  has answered :) And yeah, I'd still say there is no need for amount of FPs in so visible place and it might have gone to the small text under the title (same as D/T/size and so on)

moving-bits:
@okainov 
Sorry, then I misread your posting... Anyway, the information in the rightmost column consists of more parts than just the squared zero you see in the screenshots above: number of favos, GC vote status (if service activated), and trackable inventory count for this cache (if > 0).

PR #9096, opened 2020-09-29
refactor backup system

### LINKED ISSUE #4170: Doubled restore dialog after fresh install
2014.08.13-NB:
- Install c:geo
- Start c:geo
- Enter userdata for gc.com
- Press back to get to the main menu
- Restore dialog for exisiting backup is shown
- Select OK
- Confirmation is shown, select OK
- Restore dialog is shown again

### LINKED ISSUE #8959: Unified handling for backing up/restoring caches & settings
Currently we have two backup functions: `settings` and `caches/waypoints`.

While often switching between developer/nightly build, I always have to do both backups which is quite annoying and also for normal users probably rather confusing.

So I would vote for combining these functions in the UI.

And if there are really users, who only want to import settings without database (or vice-versa) we can add two checkboxes where they could deselect one part of the backup while importing...

fm-sys:
fixes #4170
fixes #8959

Implements a unified handling for backing up/restoring caches & settings. Note: the backup will still be split up into two files and is therefore compatible/identical with our current backup.

Probably some strings are still not perfect, so feel free to provide suggestions... Also UI improvement ideas are more than welcome!

@MagpieFourtyTwo If you have time, it would be very nice if you could test this PR build and provide some "real life feedback". I tested (hopefully) everything, so restore your previous backup & creating a new one should work completely.


fm-sys:
Two additional screenshots showing the UI:

Screenshot_20200929_212730_cgeo geocaching (Phone)

Screenshot_20200929_213014_cgeo geocaching (Phone) (Benutzerdefiniert) (1)

MagpieFourtyTwo:
Looks good.  Although I implicitly (and unfairly, as I have to admit) expected to already get support for backup history.
But this is of course a different issue.  ;)

Ok, back to topic: Dialog and options look perfectly fine to me.
Just one remark: "restoration" is (AFAIK) usually used for e. g. buildings ... I would suggest to stay with "Start restore" (as "restore" is commonly used anyway) or, if you don't want to just repeat the phrase, perhaps "recovery".

> the backup will still be split up into two files and is therefore compatible/identical with our current backup.

IIRC it was general tenor anyway not to change that, as it should be possible to restore data *OR* settings.

fm-sys:
@MagpieFourtyTwo thanks for testing! Today I found time to make some minor fixes and to implement the backup history support. So maybe time for another testing round? ;-)

If there are no typos, etc. it would probably be the best to merge it to master, and then collect some user feedback... From my testing, backing up and restore should run without bugs

MagpieFourtyTwo:
Looking forward ...  ;)

MagpieFourtyTwo:
> backup history support

Maybe I've missed s/th, but after a quick glance it seems as if you're moving the backup to a dir prefixed with "yyyyy-mm-dd" ...?
Not a big deal, but I would have added "_hh-mm", cause sometimes there are more backups needed on one day.
Moreover it sometimes may be helpful to know if it has been created in the morning or in the evening ...

MagpieFourtyTwo:
As it's already merged to master I'll just wait for the nightly - saves the hassle with re-installing two times. ;)

geocachermgo:
Thanks for the improvment, I like it. 3 remarks

1) I find the "Start backup" text a little unobtrusive and the heading would be more correctly "backup settings"
IMG_20201001_090238

I propose a action button as in "add waypoint" to start the backup.
IMG_20201001_084122
Maybe on top "Create backup" "Restore from <date>" and "Restore from old backup" followed by a subheading "settings"?
I would omit the separate section "advance settings" and just add it as a 4th option.

2) In the acknowledgement window the creation and copying of the old_backup is missing:
Screenshot_2020-10-01-08-34-18-485_cgeo geocaching
3) If we create old backups the should be a possibility to restore from old backups cf suggestion of my third button that should result in a selection menue.

fm-sys:
Let's use #9104 for future discussions...

Issue #8659, opened 2020-07-19
Use new logo on other cgeo project websites


bekuno:
The (old) c:geo logo is also used on the github project site, the both crowdin projects, the codacy organization, maybe more.
Do we want to replace the old logo with the new one, create a dedicated project logo or use the old one?

Lineflyer:
The new one everywhere would be my suggestion.
Lets make a list.

Lineflyer:
- [x] Github organization icon
- [x] Avatar of c:geo-ci-bot
- [ ]  what else

IiroP:
[Twitter page](https://twitter.com/android_gc) is also using the old logo

fm-sys:
[Playstore developer page](https://play.google.com/store/apps/dev?id=7945466068179931672) has still the old icon too...

Lineflyer:
@mucek4 I think you need to update the profile icon on Google Play

fm-sys:
I'll open a new issue: #17150

PR #9098, opened 2020-09-30
update widget icons

fm-sys:
related to #8661 and #8659 we should also update our widget icons:

| Screenshot_20200930_132212 | --> | Screenshot_20200930_132118_com huawei android launcher |
|-|-|-|


bekuno:
The icons are created dynamicly - long press the c:geo icon, then you have also the 4 icons for the shortcuts shown. Now long press on such icon - you can create the launcher icon with the new c:geo image (very small in the edge in my case).
I don't know when it was implemented.

By the way, I can not create a "Go To" icon in this way, and the "Verlauf" icon has on my phone the text "Gespeichert".

You are right that the launcher icons needs the refresh.
I asume that we need an update routine, isnt it?

(tested with 2020.08.04)

fm-sys:
Amazing! I did not know this option... But this PR is about the widgets you can insert in your launcher via "widgets > c:geo Shortcut"

But we should probably match the c:geo icon size for the two different types, shouldn't we?

bekuno:
> But we should probably match the c:geo icon size for the two different types, shouldn't we?

That would great.
And also the same icons and text.

For the icon principle - shouldn't it the oposite presentation - as base the c:geo icon and overlay the extra function? (as for contacts plugin) 
@Lineflyer Was there such discussion in the past?

fm-sys:
> For the icon principle - shouldn't it the oposite presentation - as base the c:geo icon and overlay the extra function? (as for contacts plugin)

Would also be fine for me. But note that this could not be done for the shortcuts created as you described above...
> long press the c:geo icon, then you have also the 4 icons for the shortcuts shown. Now long press on such icon - you can create the launcher icon with the new c:geo image (very small in the edge in my case)

Lineflyer:
> @Lineflyer Was there such discussion in the past?

I remember, that we choose this because the user might want to easily identify which widget shortcut he is about to use latest in case he uses more than one.
Secondly it matches the built in shortcuts mentioned above.

moving-bits:
I'm unsure about the state of discussion in this case. Any changes planned for this PR, or is it ready to merge?

bekuno:
From my taste I would have the c:geo icon a little bit smaller.

A update routine can go to another PR, imho.

Lineflyer:
I am fine with this PR.

PR #9099, opened 2020-09-30
add tooltip to PN buttons

fm-sys:
As we still have the currently unused resource strings for the personal note buttons in the `strings.xml`, I thought it would be useful to display them as a long click tooltip...

Screenshot_20200930_161421

---
Further question: Should we display a tooltip for image buttons also in other places?

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -148,6 +148,7 @@ import android.widget.Toast;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.view.ActionMode;
+import androidx.appcompat.widget.TooltipCompat;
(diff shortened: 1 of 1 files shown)


moving-bits:
> As we still have the currently unused resource strings for the personal note buttons in the `strings.xml`, I thought it would be useful to display them as a long click tooltip...

Sounds useful

> Further question: Should we display a tooltip for image buttons also in other places?

We already do have that in some places (eg: "search" and "history" icons on mainscreen), haven't looked up the technical details for those, though.

Lomanic:
> Further question: Should we display a tooltip for image buttons also in other places?

Keep in mind that some icons already have a special action on long-press https://cgeo.droescher.eu/en/hiddenfunctions

fm-sys:
> Keep in mind that some icons already have a special action on long-press

I know... But there are also many without  ;-)

Issue #7634, opened 2019-05-21
Hint button missing when maps live map enabled


bekuno:
##### Detailed steps causing the problem:
- have a cache with hint saved
- open map view
- Set map option to live map disabled
- open the cache preview 
 -> hint button is visible
- Set map option to live map enabled
- open the cache preview 
 -> hint button is not visible

##### Expected behavior after performing these steps:
The hint button should be visible when a hint exist in both modes
(at least for saved caches)

##### Version of c:geo used:
2019.05.20

##### Is the problem reproducible for you?
Yes

##### System information:
Android 9

Lineflyer:
I experienced something like that but not reproducible with the steps mentioned above yet. Need to try again.

However I experienced cases where the hint button was not shown but should have been (cache was saved and has hint).

moving-bits:
I played around with this in c:geo a bit: It seems to be related to "automatic reloading" and synchronizing data while in live mode.

Example:
- While not in live mode: Cache says "saved before 10 months", shows hint button
- After enabling live mode: Cache says "saved a couple of minutes ago", no hint button show - but if you click on "details" the hint itself is shown.
- If you disable live mode again, cache switches back to the first state of "saved before 10 months" and hint button shown.

So it looks like the hint information is lost in memory (while still being available in the database).

SammysHP:
A long time ago we had some kind of lazy loading of caches to improve database performance. When loading a cache you could control whether the full cache should be loaded from the database or just some fields (e.g. for lists). Not sure if it is still done this way.

bekuno:
> While not in live mode: Cache says "saved before 10 months", shows hint button
> After enabling live mode: Cache says "saved a couple of minutes ago",

This is Issue #7593

moving-bits:
Fixed with #9120 as well.

PR #9120, opened 2020-10-02
Fix #9069 - Hint is not displayed on compass view, when launched directly from map

### LINKED ISSUE #9069: Hint is not displayed on compass view, when launched directly from map
**Describe the bug:**
When the compass is directly launched from a cache in map view (icon next to the cache title), and the "Hint" icon is pushed, the hint toast always says "No hint available", even if the cache does have a hint. Note that, if you go into "More Details", and launch the compass from that view, the hint toast is displayed correctly.

**To Reproduce:**
1. Go to map view. 
2. Click on a cache icon, that has a hint (eg. GC2BG6B)
3. Click the compass icon next to the cache title the cache summary dialog.
4. Click the hint icon next to the cache title.

Hint toast is displayed as "No Hint Available"

The hint for the cache should be displayed in the toast (in this case "[Puzzle] <a href=..."

**Version of c:geo used:**
2020.09.25-dee1772 developer build

Yes (on my personal device LG Power X3, but also in the emulator)

**System information:**
```
--- System information ---
Device: Android SDK built for x86 (sdk_gphone_x86, google)
Android version: 8.1.0
Android build: sdk_gphone_x86-userdebug 8.1.0 OSM2.171116.002 4458339 dev-keys
c:geo version: 2020.09.25-dee1772 developer build
Google Play services: unavailable
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide caches: -
Hide waypoints: -
System language: en_US
System date format: M/d/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (567.2 MB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (567.2 MB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (567.2 MB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (212.0 KB) on system internal storage
Offline maps path: null
Map render theme path: 
Live map mode: true
Geocaching sites enabled:
Geocaching.com date format: MMM/dd/yyyy
```

**Additional context:**
Similar to #6091, but in that case, the toast isn't shown at all. In this case the toast is shown, but it always says "No Hint Available", whereas in that issue, the toast doesn't show up at all.

bmuzzin:
When gathering from another cache, if this cache wasn't detailed, and the other is,
the hint, etc. would never be copied. This fixes the case where the cache is stored
in the database.

Also, if there isn't a hint (or it isn't stored), don't show the show-hint icon on
the compass UI.

### DIFF
--- a/main/src/cgeo/geocaching/CompassActivity.java
+++ b/main/src/cgeo/geocaching/CompassActivity.java
@@ -236,7 +236,7 @@ public class CompassActivity extends AbstractActionBarActivity {
     @Override
     public boolean onPrepareOptionsMenu(final Menu menu) {
         super.onPrepareOptionsMenu(menu);
-        menu.findItem(R.id.menu_hint).setVisible(cache != null);
+        menu.findItem(R.id.menu_hint).setVisible(cache != null && StringUtils.isNotEmpty(cache.getHint()));
         return true;
     }
 
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -222,6 +222,16 @@ public class Geocache implements IWaypoint {
             // boolean values must be enumerated here. Other types are assigned outside this if-statement
             reliableLatLon = other.reliableLatLon;
             finalDefined = other.finalDefined;
+
+            if (StringUtils.isBlank(getHint())) {
+                hint = other.getHint();
+            }
+            if (StringUtils.isBlank(getShortDescription())) {
+                shortdesc = other.getShortDescription();
+            }
+            if (attributes.isEmpty() && other.attributes != null) {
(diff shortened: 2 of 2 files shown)


bekuno:
ok to test

moving-bits:
Cool! Also seems to fix #7634.
Thanks @bmuzzin for this fix, highly appreciated!

geocachermgo:
With the current nightly it is still not working.
Im not sure if we have that info, since the cache GC2VZYA is not "loaded".
Compass directly from Live Map
VonKarte
Back to Detail and Compass: with hint
Detail Kompass

Lineflyer:
No, if the cache is not loaded we dont have the info.

PR #9132, opened 2020-10-03
fix some db upgrade pitfalls

moving-bits:
While working on some other issues I recognized that I wasn't able to upgrade my database any longer. During investigation I found some common pitfalls:
- trying to create indices for tables not yet existing
- trying to change a table structure which has been created correctly (as the wrong step-in-between hadn't been executed as I was coming from an even older version)

This PR is created for `master` branch currently, to gain time for testing.
Shall I create it for `release` instead, to have it in the upcoming release?

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -241,7 +241,7 @@ public class DataStore {
     @NonNull private static final String dbTableExtension = "cg_extension";
     @NonNull private static final String dbTableSequences = "sqlite_sequence";
     @NonNull private static final String dbCreateCaches = ""
-            + "CREATE TABLE " + dbTableCaches + " ("
+            + "CREATE TABLE IF NOT EXISTS " + dbTableCaches + " ("
             + "_id INTEGER PRIMARY KEY AUTOINCREMENT, "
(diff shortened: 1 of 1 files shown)


eddiemuc:
sorry...

moving-bits:
> sorry...

No worries. Some of the changes have been my own, and some others are pretty old already.
I guess it's not such a frequent use case to upgrade from eg. version 75 to 87 in one step ;-)

Lineflyer:
> Shall I create it for release instead, to have it in the upcoming release?

Do you consider it an important bugfix for `release`? I lost overview the overview about the database version of current release compared to beta.

moving-bits:
> > Shall I create it for release instead, to have it in the upcoming release?
> 
> Do you consider it an important bugfix for `release`? I lost overview the overview about the database version of current release compared to beta.

People who update each version of c:geo won't notice a difference. The fixes are relevant only to those people skipping a couple of releases and trying to upgrade then.

So it depends on whether we have a relevant number of people complaining they cannot update their old version of c:geo (> 6 month) to the current one. If not, then we should stay with `master` for this PR.

Lineflyer:
Then lets stay on `master` with this PR I would say.

PR #9136, opened 2020-10-04
WIP grant access to base dir

moving-bits:
This is a first step in trying to get a grip on the new storage framework requirements enforced with Android 11.

I've set up a new branch 'storage_framework' to be able to collaborate on this topic. Whether we merge this branch later on into `master` or put the findings into separate PR is to be decided later.

This PR includes the following first steps:
- Enforces using the new storage framework by setting `targetSDK` to 30 and removing the `requestLegacyExternalStorage` workaround.
- Creates a new base dir for public c:geo data beneath the `Download` directory.
- Tries to acquire access rights for all files in that directory and store the grant in settings.
- Tries to use this new base dir as replacement for our current base dir.

The first results are promising in some places, but not everything works yet as expected:
- Step 1 works. As our current base dir (`/sdcard/cgeo`) is no longer accessible, c:geo falls back to the internal dir (`data/user/0/cgeo/geocaching`) as you can see in our system info page.
- The directory created in step 2 does work, and c:geo has access to all files within that directory structure - as long as the files and dirs get created by c:geo! If you move a new file into this structure externally (eg: a map file), c:geo won't see this file at all!
- In step 3 the system dialog to grant rights opens, and after user confirmation returns a uri like `tree/primary:Download/cgeo` which looks ok to me, but I've found no way so far to reuse this uri in further file access (step 4).

Still I'm sharing this intermediate state so that we are able to collaborate upon this.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -50,7 +50,7 @@
         android:usesCleartextTraffic="true"
         android:supportsRtl="true"
         android:largeHeap="true"
-        android:requestLegacyExternalStorage="true">
+        android:requestLegacyExternalStorage="false">
 
         <!-- Samsung Multi-Window support -->
         <uses-library
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -30,7 +30,7 @@ android {
     defaultConfig {
         minSdkVersion 21
         //noinspection OldTargetApi
-        targetSdkVersion 29
+        targetSdkVersion 30
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)
 
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -263,6 +263,13 @@
     <string translatable="false" name="pref_map_routing">map_routing</string>
     <string translatable="false" name="pref_theme_menu">theme_menu</string>
 
+
+
+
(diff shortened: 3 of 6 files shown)


moving-bits:
@Lineflyer or @bekuno 
Can you set up CI integration for the new branch? Thanks.

Lineflyer:
@moving-bits 
Sadly I have no deeper knowledge of our CI system...I always wanted to learn something about it, but did not yet get into it as it also seems to require a lot of knowledge about the build infrastructure.  If anyone has some experience with Jenkins, feel free to help and tell in case something (e.g. access) is needed.

But at least I can see, that it is building your PR. Additionally I just enabled the new branch on Codacy.

moving-bits:
> But at least I can see, that it is building your PR. Additionally I just enabled the new branch on Codacy.

That's what I meant. Thanks for updating the config.

eddiemuc:
I installed this PR locally. First thing I noticed is a warning message from Android Studio in `build.gradle` file that `compileSdkVersion` (29) should not be lower than `targetSdkVersion` (30) . Did you keep `compileSdkVersion` on 29 intentionally?

moving-bits:
> Did you keep `compileSdkVersion` on 29 intentionally?

Not really. Just pushed targetSDK to 30 to enforce SAF rules. Should have updated compileSDK as well...

Issue #8978, opened 2020-09-09
[Log: adding images] add resorting option and reintroduce image title prefix 


eddiemuc:
**Is your feature request related to a problem?**
Adding of images to logs is currently a long image-per-image-procedure involving at least four clicks through 3 dialogs (in log dialog click "add", in image dialog click "stored", in image select intent choose image, in image dialog click "ok"). This is ok in case you want to edit other image properties like title and description anyway. For people simply wanting to add a couple of pics (maybe 2 or 3) however this is quite painful.

**Describe the solution you'd like**
Currently there is a single "add" button in the log. Idea is to place three (icon) buttons there instead: 
- "add from storage": opens image select intend, allows for selection of multiple images, those are added with default title, empty description and scaled down to a predefined size (maybe 640ox)
- "add from camera": opens camera intend, pic is placed in list with same logic as above
-  "add (advanced)": works exactly like today's button (opens c:Geo's own advanced image select dialog)

Additionally I would make tipping of image in a list open a context menu with options "edit" (opens c:geo image dialog) and "remove" (removes entry from image list

**Describe alternatives you've considered**
Alternative is to do nothing. Feature does not add new functionality.

**Additional context**
- Discussion in #8970 shows that existing dialog should stay
- fast select is inspired by other caching apps

Lineflyer:
Why not replacing the own image select dialog completely by your idea of using the image intents. 

But this would mean the log view needs to be extended by subtitle and scaling selector.
Wouldn't this be even more consistent instead of having two selection methods?

Lineflyer:
So each picture would need:
- Title (current prefix field could be used)
- Subtitle/Description 
- Scale selection

eddiemuc:
> Why not replacing the own image select dialog completely by your idea of using the image intents.
> 
> But this would mean the log view needs to be extended by subtitle and scaling selector.
> Wouldn't this be even more consistent instead of having two selection methods?

This would mean that the image title list could get very long/big in the log dialog. Would that be a problem?

The usage you suggest with resp of "title" vs "title prefix" is still not clear to me. Currently you can give each img its own "title" (in the image dialog). For those imgs where this title is empty, the "title prefix" (editable in the log dialog) plus a number is used as title. This can be changed of course. Are you suggesting not to make individual img titles editale any more (just the "prefix" then set for all images)? Or is it something else? Sorry for my slow understanding...

[...]

PR #9138, opened 2020-10-04
related to #8978: fixing non-GUI-parts (img delete, remove exif, move img fct to helper)

eddiemuc:
Contains the NON-GUI-parts for #8978 :
* moves image select functionality in separate helper class `ImageActivityHelper` so it can be used elsewhere too
* ensures that EXIF information in images is ALWAYS removed
* adds delete-logic to ensure that no non-used image data is kept on users device
* some minor fixes (load image thumbnals asynchronously, ensure image default caption is correct on image deletion, center image view in image select activity)

### DIFF
--- a/main/src/cgeo/geocaching/ImageSelectActivity.java
+++ b/main/src/cgeo/geocaching/ImageSelectActivity.java
@@ -3,38 +3,27 @@ package cgeo.geocaching;
 import cgeo.geocaching.activity.AbstractActionBarActivity;
 import cgeo.geocaching.models.Image;
 import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.ui.ImageActivityHelper;
 import cgeo.geocaching.ui.TextSpinner;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.ImageUtils;
-import cgeo.geocaching.utils.Log;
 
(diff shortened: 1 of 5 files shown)


main/src/cgeo/geocaching/ImageSelectActivity.java
```diff
@@ -196,129 +196,38 @@ private void syncEditTexts() {
...
+            loadImagePreview();
+        });
     }
 
     private void selectImageFromStorage() {
-        final Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
-        intent.setType("image/*");
 
-        startActivityForResult(Intent.createChooser(intent, "Select Image"), SELECT_STORED_IMAGE);
+        //TODO: revert to single
```
moving-bits, line 209: something yet to be done, or an outdated comment?

eddiemuc: Danm, I forgot to remove the TODO tag. But the change itself was done. I can remove the false comment tomorrow (don't have access to source right now)

eddiemuc: > something yet to be done, or an outdated comment?

Todo is now removed

moving-bits: thanks

PR #9166, opened 2020-10-07
adapt to removal of old profile page (fix #9150)

### LINKED ISSUE #9150: Retirement of old profile page on gc.com
As listed on [release notes](https://forums.geocaching.com/GC/index.php?/topic/358816-release-notes-website-upcoming-retirement-of-old-profile-page/) gc.com will retire the old profile page.

I am not sure whether we use that one for any purpose (getting find count or whatever else).

Please kindly doublecheck if there is any dependency to avoid upcoming problems.

Old page: https://www.geocaching.com/my/default.aspx
New page: https://www.geocaching.com/account/dashboard

moving-bits:
- updates link for retrieving avatar image
- removes outdated check for login (non-matching already today)


Lineflyer:
release or master?
I would think to release as the removal of that page is upcoming.

fm-sys (commented):
The AvatarUrl can also be loaded from https://www.geocaching.com/play/serverparameters/params which is probably the more efficient way...

Btw this page is already parsed from c:geo at other places, so maybe these requests can also be combined (pure guessing)?

Lineflyer:
> 
> 
> The AvatarUrl can also be loaded from https://www.geocaching.com/play/serverparameters/params which is probably the more efficient way...
> 
> Btw this page is already parsed from c:geo at other places, so maybe these requests can also be combined (pure guessing)?

Interesting, It also delivers other maybe useful information.

PR #9168, opened 2020-10-07
add GitHub link to about page (fix #9164)

### LINKED ISSUE #9164: Easy in-app bug submitting function
> Yes, this problem occurred before the latest beta update. I was hoping the update would fix the problem. I couldn't find a "submit problem" link within the beta version of c:geo. That would be helpful.

_Originally posted by @pictom1 in https://github.com/cgeo/cgeo/issues/9145#issuecomment-704860761_

Any ideas where we could place such a function, to make bug submitting a bit more clear and easy for our users? 

Probably the support email is the right choice for this, aren't it? And/or what about a link to our GitHub repository?

moving-bits:
image

### DIFF
--- a/main/res/layout/about_version_page.xml
+++ b/main/res/layout/about_version_page.xml
@@ -316,6 +316,34 @@
                 <TextView
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
+                    android:text="@string/github_title"
+                    android:textColor="?text_color"
+                    android:textSize="14sp" />
+
+                <TextView
+                    android:id="@+id/github"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_marginBottom="5dip"
+                    android:layout_marginLeft="10dip"
+                    android:layout_marginRight="10dip"
+                    android:clickable="true"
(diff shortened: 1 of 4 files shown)


Lineflyer:
I would rather name it "Source code" or "Bugtracker" as Github will probably mean nothing to the common user.

moving-bits:
I would prefer "Bugtracker" as this would have some meaning to the non-developer. (And for developers both GitHub and Bugtracker would work, IMHO.) => will update the PR

@Lineflyer 
Shall I target this to release as well? Not that important, but helpful and should be unrisky.

Lineflyer:
> @Lineflyer
> Shall I target this to release as well? Not that important, but helpful and should be unrisky.

Lets keep it on `master` to avoid unnecessary new strings in `release`. Merging those is not always easy and IMHO strings should be added to `release` only in case of urgent need. `master` also offers more time for translation.

moving-bits:
> Lets keep it on `master` to avoid unnecessary new strings in `release`.

You're right. Currently the new strings are marked as `translatable="false"`, but with the change to "Bugtracker" this would indeed add a new string.

Issue #9104, opened 2020-10-01
New backup handling - todo list 


fm-sys:
Since #9096 is merged, there will probably be some discussion, UI feedback, bugs, etc. about the new backup/restore implementation (#8959). Let's collect them in this issue. 

## Important to-dos

- [x] always backup database and settings (to keep the UI clear and simple)

## Next Steps for multiple backup support
- [x] don't move the entire folder but just the backup files
- [x] always store the backup in `backup/yyyy-mm-dd hh-mm` (independent of advanced settings)
- [x] multiple backups at one day
- [x] move existing backup into new folder structure
- [x] Advanced settings: I would make the explanation text always visible & integrate it into the "create backup" section

- [x] message, that the old backup has been moved into the new folder structure

#### And later on...
- [x] choose dialog for restore backup from backup history (use c:geo's `SimpleDirChooser` component)
- [x] Add warning to description that backup is only offline (see #9175) 

- [x] Simplify the "backup successful" message to something like `backup was created at <dir path>`
- [x] add the backup state to `system information` (see #9049) 
- [x] Change overwrite notice to something like `This will purge your oldest existing *backup dir* as of dd MMM, hh:mm - continue?`
- [x] "share option" after completing the backup (maybe as a button below the "backup successful" message) #4705

#### Discussed, but outdated or opposed...
- [outdated] we could add meaningful icons on the left side in "create backup" section
- [outdated] Better wording for "Start backup"
- [outdated] Restore settings do not remember currently if I want to restore my settings as well (is it needed?)
- [outdated] After a fresh install, if you try to create a backup, c:geo shows a dialog "database is empty, no backup needed" _(This message was introduced for secure reason, so that users do not overwrite their backup when actually trying to restore it...)_

---

Feel free to complement the list...

 (checked points are done already, pull request will follow if everything works so far)

geocachermgo:
See my comment at  #9096

geocachermgo:
I just had the names of the backup files extended like cgeo_2020-10-01.sqlite in the backup dir. Do we really need a separate directory old_backup with subdirectories for each date of the backup?
We could also put the date subdirs under the backup dir if we do not want to touch the filenames.

MagpieFourtyTwo:
I would add the time in any case, because of three reasons:
1. Consistency
2. Information, as it may be good to know if the backup has been created in the morning or in the evening, e. g. before or after the "Grand Tour" ;)
[...]

PR #9179, opened 2020-10-10
Improve new backup handling

fm-sys:
## Changes - see https://github.com/cgeo/cgeo/issues/9104#issue-712512860:
- [x] always backup database and settings (to keep the UI clear and simple)
- [x] don't move the entire folder but just the backup files
- [x] always store the backup in `backup/yyyy-mm-dd hh-mm` (independent of advanced settings)
- [x] multiple backups at one day
- [x] move existing backup into new folder structure
- [x] Advanced settings: I would make the explanation text always visible & integrate it into the "create backup" section
- [x] choose dialog for restore backup from backup history (use c:geo's `SimpleDirChooser` component)
- [x] warning that backup is only local

---
- info message, that the old backup has been moved into the new folder structure

I'm unsure about this point: Should we really inform the user that his existing backup was moved into the new folder structure, or is it rater confusing? Probably most users will not care in which subdir the backup was exactly moved...  (with this PR, it only writes an info to log file instead)


fm-sys:
Screenshot to make UI discussions easier...
Screenshot_20201011_132623_cgeo geocaching


I personally would have kept the "Maximum stored backups" in a third category at the bottom, but the majority decides...

fm-sys:
As no further feedback occurred, we should merge it for testing it in our nightlies...

moving-bits:
Thanks @fm-sys for the various improvements, like them!

One small suggestion: I find it confusing that "Maximum stored backups" has its own subheading and "include account data" does not have, which makes the latter sort of vanish optically. So I would either swap the two, so that "include account data" is part of the general "Create backup" section, or introduce a separate subheading for it.

Oh, and I would start each option with a capital letter. ("Include account data" in this case)

fm-sys:
@moving-bits done!

Screenshot_20201019_180436_cgeo geocaching

moving-bits:
@fm-sys 
Thumbs up & thanks!

PR #9186, opened 2020-10-13
fix #8978: GUI part of add fast log image

### LINKED ISSUE #8978: [Log: adding images] add resorting option and reintroduce image title prefix

eddiemuc:
fix #8978: GUI part of add fast log image


Lineflyer:
Can you provide a screenshot for this PR? Would be nice to see it before merging.

fm-sys:
@Lineflyer see https://github.com/cgeo/cgeo/issues/8978#issuecomment-707863999

eddiemuc:
Thanks to @fm-sys 

@Lineflyer should I copy screenshots to the PRs rather than into the linked issues? I always try to keep discussions on the issues, since they seem more fit for "long-term-documentation" to me while PRs are mostly short-living things. Is that a wrong assumption?

Lineflyer:
> @Lineflyer should I copy screenshots to the PRs rather than into the linked issues? I always try to keep discussions on the issues, since they seem more fit for "long-term-documentation" to me while PRs are mostly short-living things. Is that a wrong assumption?

I agree that discussion should take place on the issue rather than on the PR. Maybe you could just link (like you did in your comment) to the screenshot.

PR #9196, opened 2020-10-16
fix #9193: hard-code point instead of language-specific separator

### LINKED ISSUE #9193: PN function "Add/Update Waypoints" should use "." instead of "," as decimal minute separator
When adding the waypoints to Personal Note, coordinates are formatted like ``dd° mm,mmm`` which may mathematically be correct, but is pretty unusual - and, even worse, leads to unexpected behavior when feeding these coords to various tools.

Example: This [bearing tool](http://gps-cache.de/geocaching/koordinaten-peilung.htm):
With the default coords "N52 30.871 E013 21.005" and "N52 30.977 E013 22.659" it computes a (correct) bearing of 83,992°.
But when you use dots instead of commata in e. g. the 2nd value "N52 30,977 E013 22,659", you will get 196,708°.
Which in fact took a while until I a) even noticed that there's an error, and b) recognized the reason for the strange result.

This may of course be considered a flaw in the bearing tool, but in fact the dot is the common decimal minute separator and there are a lot of tools out there, relying on it, thus the ``Add Waypoints`` routine should use the dot, too. Esp. as even c:geo uses it everywhere else.

Format saved waypoints like ``dd° mm.mmm`` (i. e. with dot as decimal separator).

eddiemuc:
fix #9193: hard-code point instead of language-specific separator

### DIFF
--- a/main/src/cgeo/geocaching/location/GeopointFormatter.java
+++ b/main/src/cgeo/geocaching/location/GeopointFormatter.java
@@ -30,6 +30,9 @@ public class GeopointFormatter {
         /** Example: "N10 12,345 W5 12,345" */
         LAT_LON_DECMINUTE_SHORT,
 
+        /** Example: "N10 12.345 W5 12.345" (unlocalized) */
+        LAT_LON_DECMINUTE_SHORT_RAW,
+
         /** Example: "N 10° 12.345 W 5° 12.345" */
         LAT_LON_DECMINUTE_RAW,
 
@@ -100,6 +103,11 @@ public class GeopointFormatter {
                 return String.format(Locale.getDefault(), "%c%d %02d%c%03d %c%d %02d%c%03d",
                         gp.getLatDir(), gp.getDecMinuteLatDeg(), gp.getDecMinuteLatMin(), decimalSeparator, gp.getDecMinuteLatMinFrac(),
                         gp.getLonDir(), gp.getDecMinuteLonDeg(), gp.getDecMinuteLonMin(), decimalSeparator, gp.getDecMinuteLonMinFrac());
+
+            case LAT_LON_DECMINUTE_SHORT_RAW:
+                return String.format((Locale) null, "%c%d %02d.%03d %c%d %02d.%03d",
+                    gp.getLatDir(), gp.getDecMinuteLatDeg(), gp.getDecMinuteLatMin(), gp.getDecMinuteLatMinFrac(),
+                    gp.getLonDir(), gp.getDecMinuteLonDeg(), gp.getDecMinuteLonMin(), gp.getDecMinuteLonMinFrac());
 
             case LAT_LON_DECMINUTE_RAW:
                 return String.format((Locale) null, "%c %02d° %06.3f %c %03d° %06.3f",
--- a/main/src/cgeo/geocaching/models/Waypoint.java
+++ b/main/src/cgeo/geocaching/models/Waypoint.java
@@ -562,7 +562,7 @@ public class Waypoint implements IWaypoint {
         if (getCoords() == null) {
             sb.append(PARSING_COORD_EMPTY);
         } else {
-            sb.append(getCoords().format(GeopointFormatter.Format.LAT_LON_DECMINUTE_SHORT));
+            sb.append(getCoords().format(GeopointFormatter.Format.LAT_LON_DECMINUTE_SHORT_RAW));
         }
         //user note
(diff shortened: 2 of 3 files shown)


UniQP:
@eddiemuc Have you seen https://github.com/cgeo/cgeo/issues/9193#issuecomment-710162730 ?

eddiemuc:
> @eddiemuc Have you seen [#9193 (comment)](https://github.com/cgeo/cgeo/issues/9193#issuecomment-710162730) ?

I don't understand this comment. Can you elaborate?

main/src/cgeo/geocaching/location/GeopointFormatter.java
```diff
@@ -98,8 +98,8 @@ public static String format(final Format format, final Geopoint gp) {
 
             case LAT_LON_DECMINUTE_SHORT:
                 return String.format(Locale.getDefault(), "%c%d %02d%c%03d %c%d %02d%c%03d",
```
UniQP, line 100:
```suggestion
                return String.format((Locale) null, "%c%d %02d.%03d %c%d %02d.%03d",
```
and remove the `'.'` arguments.

main/src/cgeo/geocaching/location/GeopointFormatter.java
```diff
@@ -98,8 +98,8 @@ public static String format(final Format format, final Geopoint gp) {
 
             case LAT_LON_DECMINUTE_SHORT:
```
UniQP, line 99: The example in the enum comment also needs to be adapted and should mention that the string in unlocalized.

UniQP (changes requested):
If we used the unlocalized version for DECMINUTE, we should also use it for other formats (DECSECOND, …).

eddiemuc:
> If we used the unlocalized version for DECMINUTE, we should also use it for other formats (DECSECOND, …).

Have you checked whether this has no unwanted side effects on other parts of c:geo?

eddiemuc:
HI @UniQP thanks for the review. You made good points, I decided to approach this a different way based on your feedback.

I now introduced a NEW enum value `LAT_LON_DECMINUTE_SHORT_POINTHARDCODED` and changed the relevant placed to use this one instead of the other value `LAT_LON_DECMINUTE_SHORT` when exporting waypoints to persnoal note. That way we can be sure I don't break existing functionality where maybe the language-specific settings are wanted and needed.

I squashed commits and force-pushed the change, maybe you (or someone) could take a second look?

main/src/cgeo/geocaching/location/GeopointFormatter.java
UniQP: @eddiemuc Also applies to new version

main/src/cgeo/geocaching/location/GeopointFormatter.java
```diff
@@ -30,6 +30,9 @@ private GeopointFormatter() {
         /** Example: "N10 12,345 W5 12,345" */
         LAT_LON_DECMINUTE_SHORT,
 
+        /** Example: "N10 12.345 W5 12.345" (always use point, independent from language settings) */
```
UniQP, line 33:
```suggestion
        /** Example: "N10 12.345 W5 12.345" (unlocalized) */
```

main/src/cgeo/geocaching/location/GeopointFormatter.java
```diff
@@ -30,6 +30,9 @@ private GeopointFormatter() {
...
+        LAT_LON_DECMINUTE_SHORT_POINTHARDCODED,
```
UniQP, line 34:
```suggestion
        LAT_LON_DECMINUTE_SHORT_RAW,
```

UniQP (changes requested):

main/src/cgeo/geocaching/location/GeopointFormatter.java
eddiemuc: done

main/src/cgeo/geocaching/location/GeopointFormatter.java
eddiemuc: done

main/src/cgeo/geocaching/location/GeopointFormatter.java
eddiemuc: done

eddiemuc:
@UniQP I believe I fixed all your comments according to your suggestions (I am a little overwhelmed by GitHub GUI, so I really hope I don't miss anything). If there's anything left, please drop a short hint.

UniQP (approved):

PR #9199, opened 2020-10-16
Call the right superclass method when overriding onRestart

osma:
When investigating #8752 I noticed a case where the onRestart method is overridden in AbstractActivity, but instead of calling the superclass method onRestart, the code calls onResume. 

This looks like a copy-paste error. I'm not sure if this bug has any consequences, but fixing it is surely the prudent thing to do. It looks like the bug was introduced in PR #8967.

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractActivity.java
@@ -324,6 +324,6 @@ public abstract class AbstractActivity extends AppCompatActivity implements IAbs
     @Override
     protected void onRestart() {
         Log.v(logToken + ".onRestart()");
-        super.onResume();
+        super.onRestart();
     }
 }


eddiemuc:
I confirm the bug and the fix (can't merge though, missing permissions) and I am outing myself as the one introducing the bug.

imho this should to to `release `branch though

bekuno:
ok to test

rsudev:
I tried to rebase to release from here - doesn't really work.
I'd suggest to cherry-pick to release and resubmit as PR (can do that as well)

UniQP:
Rebased to `release`.

osma:
Thanks @UniQP for rebasing and sorry for targeting the wrong branch! I'm a newbie in this project and I'm used to the opposite workflow, where bugs are fixed first on the master branch, then backported (often cherry picked) to release branches.

rsudev:
Fixes for regressions we might want to put out soon are rather done against the release branch. For bigger, more risky fixes, we normally create a PR with a change based on release, but against master. By that, we can test it on master but later merge it to release as well and still can merge master to release and vice versa without risks for conflicts that often come from cherry picking changes pushed to master onto release. See also https://github.com/cgeo/cgeo/wiki/How-to-get-a-bug-fix-into-the-release

moving-bits:
Thanks @osma for spotting and fixing this issue!

Issue #9094, opened 2020-09-29
Migrate codebase to Kotlin


eddiemuc:
Sorry, I am pretty sure this has been discussed before, but I couldn't find that discussion and simply have to ask this:

Is there any support/plan/approval for migrating c:geo code base to Kotlin at some point?

bekuno:
There was no such discussion here in the two years (and I think also not before).
But there was this queastion raised here: https://github.com/cgeo/WhereYouGo/issues/204.
My opinion on this is unchanged - I don't see that we have the ressources and necessity to do such a big change for a maybe little advantage.

eddiemuc:
Well, resources is indeed a serious problem because such a change would be big - and thus would need a migration plan of course. On the other hand, maybe using Kotlin would attract more developers to work on c:geo in general?

Frankly, I have to admit that I would really like to learn Kotlin one of these days... I think it is a nice language and imho it keeps the good parts of Java while enhancing it in places where it has some pitfalls. Just a personal opinion.

fm-sys:
> On the other hand, maybe using Kotlin would attract more developers to work on c:geo in general?

I'm not firm with Kotlin and also just a Java hobby programmer. But there are much more tutorials for Java out there then for Kotlin. So I would not be able to contribute any more. And I think nearly every developer is able to write Java code. So I don't really see why a switch would attract more developers. 

Let's put our effort in making c:geo greater instead of wasting time for a codebase migration...

SammysHP:
Porting a software to a different language that you don't know and that you just want to learn doesn't seem wise. It should be the other way around: learn a language, master it and then try to port a complex software.

Oh, and this would be a full-time job for many months. This doesn't make sense. Would be easier to port everything to a framework that can run on Android as well as iOS.

moving-bits:
Although I hear some people naming Kotlin a more modern programming language, and gaining more and more market share for Android apps, I have to agree with @SammysHP's comment. We should direct the majority of our energy at tackling the pressing issues like Android 11 storage framework requirements, several GUI issues, fixing errors (to name just a few), and to make c:geo an even better geocaching application.

That does not mean, though, that we cannot do anything in Kotlin at all, as compiled Kotlin code can interact with Java. So one might start with a small, separate feature. - We should check in advance, though, what modifications would be needed to our toolchain (if any), and what that would mean to compile time and APK size.

okainov:
>Is there any support/plan/approval for migrating c:geo code base to Kotlin at some point?

[...]

PR #9210, opened 2020-10-18
related to #9094: convert two classes to kotlin (GCUtils and CalculateButton)

eddiemuc:
related to #9094: convert two classes to kotlin (GCUtils and CalculateButton).

This is a tryout/attempt on how hard or easy it would be to start allowing kotlin and java in parallel


eddiemuc:
retest this please

eddiemuc:
retest this please

eddiemuc:
I guess it is not very likely that we will see Kotlin in c:geo in the foreseeable future, so I will close this one

PR #9235, opened 2020-10-20
fixes #8978 - center content in image list view vertically

### LINKED ISSUE #8978: [Log: adding images] add resorting option and reintroduce image title prefix

fm-sys:
Screenshot_20201020_182600_cgeo geocaching


main/src/cgeo/geocaching/ui/recyclerview/ManagedListAdapter.java
```diff
@@ -168,8 +168,9 @@ public void clearView(@NonNull final RecyclerView recyclerView, @NonNull final R
                 //if necessary, pass change event to views with position change
                 if (currentDragItemStart >= 0) {
                     if (notifyOnPositionChange && currentDragItemStart != viewHolder.getAdapterPosition()) {
-                        notifyItemChanged(currentDragItemStart);
-                        notifyItemChanged(viewHolder.getAdapterPosition());
+                        for (int item = Math.min(currentDragItemStart, viewHolder.getAdapterPosition()); item <= Math.max(currentDragItemStart, viewHolder.getAdapterPosition()); item++) {
```
eddiemuc, line 171: Good catch, that was a bug indeed! But you may also use notifyItemRangeChange here so you don't have to use a loop

eddiemuc (approved):

main/src/cgeo/geocaching/ui/recyclerview/ManagedListAdapter.java
fm-sys: > But you may also use notifyItemRangeChange here so you don't have to use a loop

Ok. Thanks! I didn't know that...

PR #9238, opened 2020-10-20
fix #9233: format gpx xml on export

### LINKED ISSUE #9233: Pretty-print exported GPX
**Describe the bug:**
Exported GPX files are completely unformated and thus they are virtually unreadable. Please format these files with proper XML syntax driven newlines and indentation

**To Reproduce:**
1. Load a geocache
2. In the top right "three dots pull down", select "Export", then "GX"
3. Download the file to a system and open the file with a text editor, then you will see the mess

I'd be looking at text that is unparseable for the human eye and brain

Properly formatted XML

**Version of c:geo used:**
2020.10.10

Yes

eddiemuc:
fix #9233: format gpx xml on export

### DIFF
--- a/main/src/cgeo/geocaching/export/GpxSerializer.java
+++ b/main/src/cgeo/geocaching/export/GpxSerializer.java
@@ -78,6 +78,7 @@ public final class GpxSerializer {
 
         this.progressListener = progressListener;
         gpx.setOutput(writer);
+        gpx.setFeature("http://xmlpull.org/v1/doc/features.html#indent-output", true);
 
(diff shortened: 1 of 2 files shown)


fm-sys:
You need to also update the `testWriteEmptyGPX()` test

moving-bits:
Thanks @eddiemuc for finding this easy way to add output formatting.

I've just merged it and played with it a bit. The original use case of a formatted position history does not get solved with it yet, as `GpxSerializer` is used for cache lists only, so I added a new PR #9245 to add formatting output for position history and individual route exports as well.

PR #9240, opened 2020-10-20
swap ele and time tag order on position history GPX export (fix #9239)

### LINKED ISSUE #9239: change format of position history/track
I checked the [https://www.alltrails.com/de/converter](alltrails, ex gpsies) converter. Its reading our file. The gpx output has: linefeeds, 8 decimals, ele before time. (and adds 'on AllTrails' to our track name)

 `  <trk>
    <name>c:geo history trail on AllTrails</name>
    <trkseg>
    <trkpt lat="49.89012496" lon="8.761273776">
        <ele>215.00000</ele>
        <time>2020-10-20T16:13:39Z</time>
      </trkpt>`

reg. 1) I do not see how cutting off the mm, µm and nm of a measured value (not table value where you might make a 'identical' comparison), with relative accuracy of at max several dm) will do any harm, even if the value is later modified when transferred to another representation.
reg. 2) I agree, probably the prgram is not following the standard. It's outdated by 10 years but it was market leader then. Is successor BaseCamp following the standard? If everybody (?) is doing it that way, we could do it that way too (cf. putting in `<ele>0</ele>` to make OSM import (also not following the standard but the output of the Germin devices) happy #8519)

reg. 4) Minor issue for me
reg. 5) Alltrail was not touching the name there but just reproducing. In my example files it was not included. No need for a change.

_Originally posted by @geocachermgo in https://github.com/cgeo/cgeo/issues/9143#issuecomment-713142338_

moving-bits:


moving-bits:
retest this please

PR #9242, opened 2020-10-21
fix #6150: add map attribution

### LINKED ISSUE #6150: Add attribution for online tiles
Online tiles currently miss attribution. I guess we will fix that primarily for the mapsforge beta.

eddiemuc:
fix #6150: add map attribution

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -819,6 +819,10 @@
     <string name="map_source_osm_cyclosm">CyclOSM</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="map_source_osm_offline_combined">Combined (Offline)</string>
+    <string name="map_source_osm_offline_combined_attribution_pleasewait">Combined (%d maps, please wait...)</string>
+    <string name="map_source_osm_offline_combined_attribution_details">Combined (%d maps, %d attributions): click for details</string>
+    <string name="map_source_osm_offline_combined_attribution_single">Combined (%d maps): %s</string>
+    <string name="map_source_osm_offline_combined_attribution_dialog_title">Combined map attribution</string>
     <string name="init_sendToCgeo">Send to c:geo</string>
     <string name="settings_info_send2cgeo_title">Info on send2cgeo</string>
     <string name="init_sendToCgeo_name">Your device name</string>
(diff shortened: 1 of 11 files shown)


main/res/values/strings.xml
```diff
@@ -814,11 +814,13 @@
     <!-- map sources -->
     <string name="map_source_google_map">Google: Map</string>
     <string name="map_source_google_satellite">Google: Satellite</string>
+    <string name="map_source_osm_attribution_html"><![CDATA[© <a href="http://www.openstreetmap.org/" style="color: blue">OpenStreetMap</a> contributors]]></string>
```
Lineflyer, line 817: Do we need this as translatable string?

eddiemuc: I thought that maybe the link can lead to country-specific openstreetmap websites e.g. openstreetmap.de or openstreetmap.fr..

What do you think?

SammysHP:
For openstreetmap.org please use

```
<a href="https://www.openstreetmap.org/copyright">© OpenStreetMap contributors</a>
```

That's the easiest one. Other sources do not really tell us how to do a proper attribution. Maybe something like

```
© <a href="https://openstreetmap.de/">OpenStreetMap DE</a>, map data <a href="https://www.openstreetmap.org/copyright">OpenStreetMap contributors</a>

© <a href="https://www.cyclosm.org/">CyclOSM</a>, map data <a href="https://www.openstreetmap.org/copyright">OpenStreetMap contributors</a>
```

Black text should be enough IMHO.

main/res/values/strings.xml
SammysHP: See my comment below. The attribution is specific to a map source. In case of OpenStreetMap it must link to the copyright page.

eddiemuc: ok, will do it, see my reply there

eddiemuc:
OK, I will use these links, and I will put them into strings_not_translateable.xml (as requested by @Lineflyer 

One thing though: I really like it that the link is not displayed in black but in another color - otherwise no one sees that it is a link (which might also lead to accidental clicks on it). 
I put the color explicit, because if I don't put any color then the link displays in orange (I assume because of our theme) and this is really hard to read. Ok to stay with blue?

moving-bits:
> Ok to stay with blue?

yes

main/res/values/strings.xml
eddiemuc: all done

eddiemuc:
All done, please check screenshots in Issue #6150

PR #9245, opened 2020-10-22
add GPX formatting to position history / indiv. route exports

moving-bits:
#9238 add GPX output formatting for cache exports etc. only.
This PR adds it for exports of individual routes and position history.


eddiemuc:
:-)

PR #9249, opened 2020-10-22
use serverparam instead of dashboard for avatar url (fix #9172)

### LINKED ISSUE #9172: Adapt to serverparameters/params
Reminder to myself: Try to make use of serverparameters/params

Taken from https://github.com/cgeo/cgeo/pull/9166#issuecomment-705702665 by @Lineflyer:

> The AvatarUrl can also be loaded from https://www.geocaching.com/play/serverparameters/params which is probably the more efficient way...
> 
> Btw this page is already parsed from c:geo at other places, so maybe these requests can also be combined (pure guessing)?

Interesting, It also delivers other maybe useful information like the login state.

moving-bits:


moving-bits:
Also reactivates the avatarTest and removed the avatar-unrelated querying of #caches found (the current value is no longer available when using the serverparameters instead of the dashboard page).

Issue #9224, opened 2020-10-19
New Lint issues on CI: NonConstantResourceId


Lineflyer:
Not sure, if this is of any importance, but I just noticed a sudden jump in amount of coding issue by Lint on our CI:
![grafik](https://user-images.githubusercontent.com/949669/96506217-90756700-1257-11eb-9123-d2b9a16fb872.png)

It started with this job which reported 473 new issues:
https://ci.cgeo.org/job/cgeo%20continuous%20integration/3030/

Any relation to removal of eclipse parts or library updating? Just guessing...

UniQP:
Looks like the increase is due to some new checks, so it's not a regression.

bekuno:
We have now new 390 Issues of Type NonConstantResourceId.

Description:
Resource IDs will be non-final in Android Gradle Plugin version 5.0, avoid using them as annotation attributes

~~~
Avoid the usage of resource IDs where constant expressions are required. 
A future version of the Android Gradle Plugin will generate R classes with non-constant IDs 
in order to improve the performance of incremental compilation.
~~~

Sample: `@BindView(*R.id.license) protected TextView licenseLink;`

eddiemuc:
> Sample: `@BindView(*R.id.license) protected TextView licenseLink;`
> 

Ups. This is afaik the standard pattern in most of our activities to bind gui elements to class members using Butterknife lib. Should this be changed? Maybe even further: should Butterknife be removed m

bekuno:
> Maybe even further: should Butterknife be removed?

The answer is yes.
See https://github.com/JakeWharton/butterknife
~~~
Attention: This tool is now deprecated. Please switch to view binding. Existing versions will continue to work, obviously, but only critical bug fixes for integration with AGP will be considered. Feature development and general bug fixes have stopped.
~~~

The view binding is documented here:
https://developer.android.com/topic/libraries/view-binding

moving-bits:
We are also using those ID in switch statements to reference menu items, example from CacheDetailActivity:

```
0474 public boolean onContextItemSelected(final MenuItem item) {
0475     switch (item.getItemId()) {
0476         // waypoints
0477         case R.id.menu_waypoint_edit:
0478             if (selectedWaypoint != null) {
0479                 ensureSaved();
0480                 EditWaypointActivity.startActivityEditWaypoint(this, cache, selectedWaypoint.getId());
0481                 refreshOnResume = true;
0482             }
0483             return true;
[...]

Issue #9232, opened 2020-10-20
Butterknife should be replaced with ViewBindings


eddiemuc:
The Butterknife framework, which is used in c:geo for binding views to class members, is deprecated
See https://github.com/JakeWharton/butterknife

Code samples:
<pre>
@BindView(R.id.filters) protected LinearLayout filtersContainer; 
          ~~~~~~~~~~~~
@OptionsMenu(R.menu.filter_options)
             ~~~~~~~~~~~~~~~~~~~~~ </pre>

Attention: This tool is now deprecated. Please switch to view binding. 
Existing versions will continue to work, obviously, but only critical bug fixes for integration with AGP will be considered. 
Feature development and general bug fixes have stopped.

It should be replaced with view binding as documented here:
https://developer.android.com/topic/libraries/view-binding

(This issue was brought up here #9224)

moving-bits:
Thanks for creating the issue. IMHO we should wait with working on it until we've branched the next release, see #9226. It looks like we can do the migration step by step - otherwise a separate branch will be better probably.

moving-bits:
This has been finished meanwhile, so let's close this issue.

PR #9271, opened 2020-10-24
ignore NonConstantResourceId lint messages

bekuno:
related to #9224

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -152,6 +152,13 @@ android {
 
         // ExtraTranslation - old translation will be deleted by the next crowdin import
         disable 'ExtraTranslation'
+
+        // deactivate this rule only on CI
+        // see https://github.com/cgeo/cgeo/issues/9224
+        // Resource IDs will be non-final in Android Gradle Plugin version 5.0
+        if (isContinuousIntegrationServer()) {
+            disable 'NonConstantResourceId'
+        }
     }
 
     packagingOptions {


bekuno:
now 173 lint issues
https://ci.cgeo.org/job/cgeo%20pull%20request/3872/androidLintResult/

moving-bits:
If we disable this check, how can we track if we have migrated all occurrences (eg. by implementing #9232 and other means)?

bekuno:
> 
> 
> If we disable this check, how can we track if we have migrated all occurrences (eg. by implementing #9232 and other means)?

We have to disable this rule then and test if no more such message occure.
I add a big text to #9232.
And we have two issues to track the progress.

moving-bits:
Thanks for opening the extra issues.

Is there a way to disable this rule on our CI only? By this we would still get local notification (which helps during migration), without "polluting" the CI's results.

bekuno:
Skipping the check is now only active on CI.

@Lineflyer Is this OK from your side?

Lineflyer:
> @Lineflyer Is this OK from your side?

I do not often look at those test results (as I have no idea of the coding background). So whatever is considered as OK for the team is OK for me :)

moving-bits:
> Skipping the check is now only active on CI.

Thanks for updating this!

GitHub signals a branch conflict - can you update and squash the PR? Thanks.

PR #9272, opened 2020-10-24
fix #9266: prevent parsing waypoints to type 'own'

### LINKED ISSUE #9266: Waypoint type changes  if personal note is involved
Just had this problem (not tested if reroducible or the dependencies):
2020.10.23-NB on Android 11

- Open GC236A2
- Swipe to waypoint tab
- Add wayppoint with coordinates and Save (Default type: Referencepoint)
- Swipe to cache details
- Select to copy waypoint to personal note
- Swipe to waypoint tab
- Click on waypoint created before to edit it

Result:
- Waypoint type changes to Parking

eddiemuc:
fix #9266: prevent parsing waypoints to type 'own'
Waypoints of type "reference Point" with name "reference point" in it where accidentally parsed as wptype "own" because it uses the same translated name. This lead to such waypoints becoming "own" when parsing them repeatedly from PN. This leads to switch type to "Parking" in waypoint edit dialog because "own" is not an allowed type in that list.

Prevented this by not allowing "own" (and "Original") as parsed wptypes any more. All allowed wptypes in parsing have different translations now

### DIFF
--- a/main/src/cgeo/geocaching/enumerations/WaypointType.java
+++ b/main/src/cgeo/geocaching/enumerations/WaypointType.java
@@ -66,14 +66,20 @@ public enum WaypointType {
     }
 
     @NonNull
-    public static final List<WaypointType> ALL_TYPES_EXCEPT_OWN_AND_ORIGINAL = orderedWaypointTypes();
+    public static final List<WaypointType> ALL_TYPES = orderedWaypointTypes(false);
+    @NonNull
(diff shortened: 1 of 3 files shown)


eddiemuc:
I have problems running cgeo unit tests locally since last Gradle upgrade, will fix this PR as soon as I fix up my local IDE

eddiemuc:
Update: thought too short on this fix. The right way to do this is to give the different aspects on how to get waypoint type from text a clear priority, which I did now with latest change. Also, I made sure that if multiple (x) ae found, only the last one counts

PR #9276, opened 2020-10-25
lint changes/fixes

bekuno:
- configure lint for the crowdin process
  - add 'ExtraTranslation' ignore rule in build.gradle for runtime check
  - add 'ExtraTranslation' ignore rule in lint.xml for lint check
  - remove 'MissingQuantity' ignore check
  - remove 'ImpliedQuantity' ignore check
  - edit 'UnusedQuantity' ignore check to all files

- add some unused images (png/xml) to the ignore list

- add 'ObsoleteLintCustomCheck' ignore rule
  These warnings are only for external sources

- remove 'InvalidPackage' ignore rule for runtime check
  This is fixed in the meantime

- delete some ' UnusedResources' strings
  see https://ci.cgeo.org/job/cgeo%20pull%20request/3872/androidLintResult/type.-2046836529/package.-963969145/

@Lineflyer @moving-bits @eddiemuc @fm-sys 
Please check if I maybe did too much.

### DIFF
--- a/main/lint.xml
+++ b/main/lint.xml
@@ -10,7 +10,7 @@
     <!-- false positives for return values of RxJava subscribe() calls -->
     <issue id="CheckResult" severity="ignore" />
 
-    <!-- avoid outdated lint warnings typicly from external libraries -->
+    <!-- avoid outdated lint warnings typically from external libraries -->
     <issue id="ObsoleteLintCustomCheck" severity="ignore" />
 
     <issue id="DuplicateIds" severity="error" />
--- a/main/lint.xml
+++ b/main/lint.xml
@@ -10,6 +10,9 @@
     <!-- false positives for return values of RxJava subscribe() calls -->
     <issue id="CheckResult" severity="ignore" />
 
+    <!-- avoid outdated lint warnings typicly from external libraries -->
+    <issue id="ObsoleteLintCustomCheck" severity="ignore" />
+
     <issue id="DuplicateIds" severity="error" />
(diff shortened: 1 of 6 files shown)


fm-sys (approved):

main/lint.xml
```diff
@@ -10,6 +10,9 @@
     <!-- false positives for return values of RxJava subscribe() calls -->
     <issue id="CheckResult" severity="ignore" />
 
+    <!-- avoid outdated lint warnings typicly from external libraries -->
+    <issue id="ObsoleteLintCustomCheck" severity="ignore" />
+
```
moving-bits, line 15: Can we set the ignore flag for external libraries only?
Otherwise we would switch it off for our own code, too, which I'd like to avoid.

BTW: spelling is "typically"

moving-bits:
Generally LGTM, thanks for deep-diving into this. Especially the `ExtraTranslation` should help us get rid of some temporary pseudo issues after deleting strings.

main/lint.xml
bekuno: We do not use custom lint checks in c:geo, so there is no problem for the normal checks.
See https://ci.cgeo.org/view/Continous%20integration/job/cgeo%20continuous%20integration/3047/androidLintResult/type.-1692134117/

Will fix the typo :-)

Lineflyer (approved):
I can only judge the extra translation:
IMHO its fine to ignore these as Crowdni will take care sooner or later.

eddiemuc (approved):
Looks good to go

Issue #9270, opened 2020-10-24
[nightly]  OSM will not be displayed


geocachermgo:
I'm afraid, this fix was possibly a little bit too quick: in latest nightly the OSM will not be displayed, until you select the map tile entry again. This happens when using "Live map" from main screen as well as "Show list on map".

Moreover zoom level and position are no longer the same between calls of "Live map", instead it seems to use max zoom and a strange position ...

Sorry for not being more precise, but currently I'm out in the woods ...

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/9188#issuecomment-715911269_

geocachermgo:
I was about to report this too, but as a geocaching friend stated: Really hard is to get the STF.

geocachermgo:
I was about to report this too, but as a geocaching friend stated: Really hard is to get the STF.

geocachermgo:
Procedure to reproduce:
1 Chose offline map
2 Back to main page
3 Chose Live Map
Result: Blank screen instead of map:

![Screenshot_2020-10-24-15-42-31-930_cgeo geocaching](https://user-images.githubusercontent.com/43813636/97083362-b318ce80-160f-11eb-865c-c39edc5136c1.jpg)

eddiemuc:
There's also another change in this nightly related to map: adding map attribution. Can't tell now whether this is the cause of the problem. 
I tried to reproduce your problem on emulator without success. Do you have a log for analyzation?

MagpieFourtyTwo:
> There's also another change in this nightly related to map: adding map attribution. Can't tell now whether this is the cause of the problem.

Yep, remembered this one afterwards, too.
But what I don't get right now: why does it work when "Combined offline" is selected? Nevertheless, exactly this seems to vote for s/th related to the newly invented attribution ...

> I tried to reproduce your problem on emulator without success. Do you have a log for analyzation?

Sunday evening, earliest.

If you want to go on trying to reproduce: "Best" result (i. e. most probably failing) seems to be with lists, containing caches located on two different map tiles, with your position on one of these two, while "yours" is located on the active tile in map view (instead of "Combined offline").

geocachermgo:
Very reproducable for me, I created a logfile. Is there any non-public info in it? Otherwise I can send it to a developer (Eddie?) directly.

eddiemuc:
Got the log and have an idea. Hopefully I find the time tomorrow to dig into it. Right now I have to manage a sleepover...

[...]

PR #9278, opened 2020-10-25
related to #9270: fix memory leak in Routing

eddiemuc:
related to #9270: fix memory leak in Routing
The way passing of callback to Routing was handled, this lead to Routing keeping a reference to NewMap object even after destroy. Also, only the callback of last Activity calling "connect" was maintained. Both is fixed with this PR.


moving-bits:
@eddiemuc 
Thanks for catching those!

PR #9280, opened 2020-10-25
related to #9270: use own MapFiles for map attribution and close them

eddiemuc:
related to #9270: use own MapFiles for map attribution and close them


eddiemuc:
Ready for merge from my side

PR #9287, opened 2020-10-26
remove Android 11 warning (fix #9225)

### LINKED ISSUE #9225: Remove Android 11 incompatability warning
According to my comment https://github.com/cgeo/cgeo/issues/8457#issuecomment-709495213 there is no problem using c:geo with Android 11.
We should therefore remove the warning currently shown on main screen if c:geo is running on Android 11.

moving-bits:


main/src/cgeo/geocaching/network/StatusUpdater.java
```diff
@@ -81,7 +79,7 @@ static Status defaultStatus(final Status upToDate) {
             } catch (NumberFormatException e) {
                 // skip version check if no parseable number returned
             }
-            return Build.VERSION.SDK_INT >= 30 ? ANDROID11NOTICE_STATUS : Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP ? CLOSEOUT_STATUS : NO_STATUS;
+            return Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP ? CLOSEOUT_STATUS : NO_STATUS;
```
bekuno, line 82: SDK_INT is never < 21. Please change to 
`  return NO_STATUS;`
see https://ci.cgeo.org/view/Continous%20integration/job/cgeo%20continuous%20integration/3047/androidLintResult/type.398387310/

or add `@SuppressLint(" ObsoleteSdkInt")` if it should stay as template.

moving-bits: The latter, so I'll add a `@SuppressLint`

PR #9291, opened 2020-10-26
add splash logo (fix #8988)

### LINKED ISSUE #8988: Splash Screen while starting c:geo

moving-bits:
Alternative to #8993.

This PR implements its own `SplashActivity` which does nothing more than displaying the image and forwarding the intent to the original `MainActivity`.


fm-sys:
Tested, but didn't work for me. See https://github.com/cgeo/cgeo/pull/8993#issuecomment-718035795

moving-bits:
Adapted this PR a bit after having found slightly different examples - does work over here now :-)

I found no way for transparency or for setting the background according to our theme settings at this early stage of loading program, only predefined opaque colors, so I chose a middle-grey for background for now:
image

@fm-sys
Would be great if you could cross-check it, if you find the time.

fm-sys:
I found time to test it - and I was impressed. What a smooth startup! Now it's even more fun to open c:geo just to see the startup animation ;-)

Lineflyer:
Cool. I would rather use a (almost) black background (as black is the default theme in c:geo).

fm-sys:
What about the background color used at our home activity when transparency is deactivated?

moving-bits:
Updated it to the dark grey tone used on MainActivity when transparency is disabled:

image

moving-bits:
and thanks for cross-checking / testing!

PR #9296, opened 2020-10-27
Fix #9259 - Exclude translations from merging

### LINKED ISSUE #9259: Merge strategy for translations
This is more or less a reminder to myself to finally figure out the best merge strategy for Crowdin translations now that we have two branches (`release` and `master`) connected to Crowdin.

The duplication suppression on Crowdin works as intended (showing same string IDs only once and use it on both branches).
However as usually strings might be removed or added on `master` while `release` is stable, changes keep coming in for `release` as soon as strings existing on both branches are modified in translation.

At first I tried to merge translations to `release`first, then translations to `master` and afterwards merge `release` upstream to `master`. This resulted in merge conflicts most of the time. 

It might make more sense to reverse this approach:
- Merge translations to `master`
- Merge translations to `release`
- Merge branches

As `master` does always contain all commits contained on `release` (as those branches are merged regularly) IMHO (with my limited Git experience) it should work out without conflicts in most cases....

Need to try this once more to proof.

Lineflyer:
- Add `.gitconfig` to define a dummy merge strategy which alwasy returns true
- Add `.gitattributes` to apply this dedicated merge strategy for translated files
- Moves translations for Playstore description into common folder for easier wildcard matching in `.gitattributes`
- Change `crowdin.yml` to match the moved folders for translated Playstore description

**Important note**:
Contributors, who directly merge branches on the repository need to execute `git config --local include.path ../.gitconfig` once for their cloned local repository to allow git to evaluate and use the configuration in `.gitconfig`

Once merged I will add an explanation to https://github.com/cgeo/cgeo/wiki/Translation-%28L18N%29


fm-sys (approved):
I'm not really firm with Github, but from my understanding I would consider it as good...

Lineflyer:
Wiki has been updated:
https://github.com/cgeo/cgeo/wiki/Translation-(L18N)#conflict-handling-between-branches

Lineflyer:
@moving-bits 
Do you want to take a look before merging or shall we merge and "test" it during next branch merges after translations have been merged in.

moving-bits:
> I'm not really firm with Github, but from my understanding I would consider it as good...

Same here for me, so I would say: Let's merge and give it a try.

moving-bits (approved):

Issue #9288, opened 2020-10-26
Preserve zoom level for leaving/entering map


bekuno:
**Is your feature request related to a problem?**
This wish arose while testing the global filter.
Every time I went back to the map the map window was set to a very small window.

**Describe the solution you'd like**
When leaving the map then the current resolution/map window is saved and restored when opening the live map.
It is not necessary to store this in the database.

rsudev:
Do I understand you correctly, that you refer to the viewport shown?
If the viewport is no longer preserved across openings of the live map, this is a bug...

moving-bits:
Don't we have a minZoomLevel? Maybe this value needs to be decreased (or removed).

Lineflyer:
@bekuno 
Is it this same as described as a regression in #9293 

This behavior however came in only as a regression in todays nightly.
Before that it was working as you sugggested.

bekuno:
This is already in release 2020.10.10.

bekuno:
And on all maps.
Eg. GSat reset to 200m if greater than 200m
OSM online/offline to 100m

Lineflyer:
@bekuno You are right, how could I overlook such an issue.
That is a regression for sure, but no feature requests.

eddiemuc:
I created PR #9298 for this. However, I am really unsure whether I tackled this correctly (more details in PR). 
Is it possible that this never ever worked on the new map? (mapsforge 0.14 / v4)?

Lineflyer:
I checked older releases as I am still a bit puzzled not being stumbled upon this problem earlier:
- Open map
- Scroll to some random point and zoom out a bit
- Leave map
- Reenter map

| Version |  GMaps | OSM |
|--|--|--|
|2020.07.02| Fail | OK |
|2020.08.04 | Fail | OK |
|2020.10.04 | Fail | Fail |
|2020.10.10 | Fail | Fail |

Maybe it helps.

moving-bits:
The change in behavior for OSM has been triggered by the zoom level fix probably, which removed the adding/substracting 3 from the GM zoom level for OSM, leading to a lower min zoom level for OSM.

As for the bug itself: I think I have seen this behavior the first time after merging the Google Maps v2 patches we may have a look at that PR, if something got introduced triggering this. I'll try to dig into the commit history...

moving-bits:
[...]

PR #9298, opened 2020-10-27
related to #9288: preserve zoom level

eddiemuc:
related to #9288: preserve zoom level

This PR is supposed to make the NewMap preserve last zoom levels and restore them when map is reopened.
However, I really wonder whether I oversaw something in the code, because if I fixed this the right way this means that zoom preserving was never enabled for NewMap before... and I cannot really believe that.

@moving-bits : can you please take a close look at this one?

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/MfMapView.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/MfMapView.java
@@ -89,12 +89,12 @@ public class MfMapView extends MapView {
         getModel().mapViewPosition.setZoomLevel((byte) zoomLevel);
     }
 
-    public void zoomToViewport(final Viewport viewport) {
+    public void zoomToViewport(final Viewport viewport, final MapMode mapMode) {
 
         getModel().mapViewPosition.setCenter(new LatLong(viewport.getCenter().getLatitude(), viewport.getCenter().getLongitude()));
 
         if (viewport.bottomLeft.equals(viewport.topRight)) {
-            setMapZoomLevel(Settings.getMapZoom(MapMode.SINGLE));
+            setMapZoomLevel(Settings.getMapZoom(mapMode));
         } else {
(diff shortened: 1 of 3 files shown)


Lineflyer:
@eddiemuc 
What about GMaps. Obviously we seem to have a similar/the same problem there.

eddiemuc:
> @eddiemuc
> What about GMaps. Obviously we seem to have a similar/the same problem there.

Can't test GMaps (not possible on emulator). But if I found the correct issue, it would affect GMaps as well.
Simply put: as far as I could see, last zoom level is correctly stored on map view exit. But on map load, when previously stored zoom level is retrieved, the stored value is pushed upward to at least 16 (which is quite zoomed-in):

https://github.com/cgeo/cgeo/blob/8ae936acf84130945555b457381c00edc8bc52ee/main/src/cgeo/geocaching/settings/Settings.java#L66-L69

https://github.com/cgeo/cgeo/blob/8ae936acf84130945555b457381c00edc8bc52ee/main/src/cgeo/geocaching/settings/Settings.java#L923-L928

I changed that to:
`return  Math.max(0, Math.min(getInt(R.string.pref_lastmapzoom, 14), INITIAL_MAP_ZOOM_LIMIT));`

This was the main change. The rest is about storing different zoom levels for different kinds of "maps" (e.g. opening "live", opening from a cache list, opening for a single cache, ...)

moving-bits:
> Can't test GMaps (not possible on emulator).

You can use Google Maps with Android Studio / in the emulator, you just need to get your own set of developer keys and store them in the xml config. A how to for this is in our https://github.com/cgeo/cgeo/blob/master/README.md in the "API keys" section.

Lineflyer:
> This was the main change. 

Sounds logical. I assume GMaps needs a similar fix.

moving-bits (approved):
LGTM, thanks!

Issue #9292, opened 2020-10-27
OSM Map loading slower than it used to be


Lineflyer:
Taken from https://github.com/cgeo/cgeo/issues/9270#issuecomment-716890164 reported by @MagpieFourtyTwo as a possible regression after fixing that issue:

Map loading is way slower than ... a few days ago ;)
e. g.: with Show as list it takes up to 5 secs until the map is loaded completely (even with just BaWü tile selected),
which used to happen during a wink of the eye before

MagpieFourtyTwo:
@eddiemuc 
> is the map loading slower than before only when you select a single map file or also when you use "combined map"?

``Combined`` is not significantly faster. While it depends on how many map files are involved: If all caches to be displayed are on one, it *may* be a little bit faster, but if there are more than one map files involved, it's even slower. Although this may also be a matter of how big the displayed area is. Have to check more detailed, once I have more time.

SammysHP:
If map loading is the bottleneck, we should use the file that is already opened for rendering and in case of combined maps only when the popup is opened.

eddiemuc:
> If map loading is the bottleneck, we should use the file that is already opened for rendering and in case of combined maps only when the popup is opened.

Usage of same mapfile for map attribution and rendering was the cause for the map rendering problems solved last night. Can't really tell you why, but somehow underlying file got closed as it was time to render the map (no, I did not call close on MapFile after retrieving attribute info :-)). Might also be a problem with how we reuse MapSource instances in our code.

Currently for combined map the mapfiles are opened asynchronously to get attribution info out. For single maps it is currently synchronous. I can change this to asynchronous as well, but if this is the cause of the problem then opening of combined maps should currently work as fast as before, and according to @MagpieFourtyTwo it doesn't.

eddiemuc:
Created PR #9299. This PR makes calc to mal attribution asynchronous for single offline maps. In hope that this is the problems cause

MagpieFourtyTwo:
To consolidate my background I went through #6150 again, trying to find some entry points for further testing ...

Reading this: "to actually get the attribution information out of them, I have to open them" I thought it might possibly be a factor when having plenty of files, so I reduced my set from 17 to 3 map files - but to no avail. With regard to speed this reduction makes no difference at all. Even with just 1 map file it's as slow as with 17. This info just in case ...

[...]

PR #9299, opened 2020-10-27
related to #9292: make attribution calc asynchronous

eddiemuc:
related to #9292: make attribution calc asynchronous
This makes calc to mal attribution asynchronous for single offline maps. In hope that this is the problems cause

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -812,8 +812,9 @@
     <string name="map_source_osm_osmde">OpenStreetMap.de</string>
     <string name="map_source_osm_cyclosm">CyclOSM</string>
     <string name="map_source_osm_offline">Offline</string>
+    <string name="map_source_osm_offline_attribution_pleasewait">(Please wait…)</string>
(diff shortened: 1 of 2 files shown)


main/res/values/strings.xml
```diff
@@ -812,6 +812,7 @@
     <string name="map_source_osm_osmde">OpenStreetMap.de</string>
     <string name="map_source_osm_cyclosm">CyclOSM</string>
     <string name="map_source_osm_offline">Offline</string>
+    <string name="map_source_osm_offline_attribution_pleasewait">(please wait...)</string>
```
Lineflyer, line 815: You should use the ellipsis character instead of dots and start with a captial letter "(Please wait…)"

eddiemuc: done and done

eddiemuc:
@moving-bits although maybe not necessary to solve the underlying problem, would still be nice to integrate this. It also contains some inor code cleanups, and parsing map attribution asynchronously is still cleaner than doing it on the GUI thread.

PR #9304, opened 2020-10-27
lint changes/fixes

bekuno:
- categorie Correctness
  - type DefaultLocale:
    Implicitly using the default locale is a common source of bugs: Use `String.format(Locale, ...)` instead
    - CalculatorVariable.java
	- CompassActivity.java
	- Log.java
  - type SimpleDateFormat:
    To get local formatting use `getDateInstance()`, `getDateTimeInstance()`, or `getTimeInstance()`, or use `new SimpleDateFormat(String template, Locale locale)` with for example `Locale.US` for ASCII dates.
    - ContextLogger.java
	- Formatter.java
	- TrailHistoryExport.java

@eddiemuc @moving-bits 
Please check that the changes are set as intended in your PR´s.

### DIFF
--- a/main/src/cgeo/geocaching/CompassActivity.java
+++ b/main/src/cgeo/geocaching/CompassActivity.java
@@ -44,6 +44,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import java.util.Arrays;
+import java.util.Locale;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -408,7 +409,7 @@ public class CompassActivity extends AbstractActionBarActivity {
         while (direction >= 360f) {
             direction -= 360f;
         }
-        deviceHeading.setText(String.format("%3.1f°", direction));
+        deviceHeading.setText(String.format(Locale.getDefault(), "%3.1f°", direction));
 
         if (deviceOrientationMode.get() == DirectionData.DeviceOrientation.AUTO) {
             deviceOrientationMode.setTextDisplayMapper(d -> getString(R.string.device_orientation) + ": " + getString(dir.getDeviceOrientation().resId) + " (" + getString(DirectionData.DeviceOrientation.AUTO.resId) + ")");
@@ -419,7 +420,7 @@ public class CompassActivity extends AbstractActionBarActivity {
 
     /** formats a float to a decimal with length 4 and no places behind comma. Handles "-0" case. */
     private static String formatDecimalFloat(final float value) {
-        final String formattedValue = String.format("% 4.0f", value);
+        final String formattedValue = String.format(Locale.US, "% 4.0f", value);
         if (formattedValue.endsWith("-0")) {
             return "   0";
         }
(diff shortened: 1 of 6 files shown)


eddiemuc (approved):
Looks good to me. Thanks for taggling this.

main/src/cgeo/geocaching/export/TrailHistoryExport.java
```diff
@@ -113,7 +113,7 @@ protected File doInBackgroundInternal(final TrailHistoryElement[] trail) {
                 gpx.attribute("", "creator", "c:geo - http://www.cgeo.org/");
                 gpx.attribute(NS_XSI, "schemaLocation", NS_GPX + " " + GPX_SCHEMA);
 
-                    final SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
+                    final SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
```
moving-bits, line 116: I don't think we need a Locale for `SimpleDateFormat` as we give an explicit format string, but it should not harm either. (Valid for the other occurrences of `SimpleDateFormat` as well.)

main/src/cgeo/geocaching/CompassActivity.java
```diff
@@ -408,7 +409,7 @@ private void updateDeviceHeadingAndOrientation(final DirectionData dir) {
         while (direction >= 360f) {
             direction -= 360f;
         }
-        deviceHeading.setText(String.format("%3.1f°", direction));
+        deviceHeading.setText(String.format(Locale.US, "%3.1f°", direction));
```
moving-bits, line 412: Shouldn't we use the user's locale here (maybe default locale)?
Using Locale.US would format decimal numbers with a decimal point instead.

moving-bits (changes requested):

main/src/cgeo/geocaching/CompassActivity.java
eddiemuc: That would have been my original intention. But if language specific separators are preferred that's fine for me too.

bekuno: 1:1
And I would leave the decition to you ...
Ok, I will change it.

main/src/cgeo/geocaching/export/TrailHistoryExport.java
bekuno: See the lint message:
https://ci.cgeo.org/job/cgeo%20continuous%20integration/3057/androidLintResult/type.-11914057/file.-1822116914/

main/src/cgeo/geocaching/CompassActivity.java
moving-bits: Thanks. If I was looking at the right part of the dialog, it's localized in current release (probably due to implicit use of default locale).

bekuno: done

moving-bits (approved):

PR #9316, opened 2020-10-30
store UDC's from map to "user-defined caches" list

### LINKED ISSUE #8678: New UDC created from map should be saved to "User-defined caches" list
AFAIR new user defined caches should be stored on the dedicatd "Internal caches" list upon creation by long clicking on the map.
For me they are always stored on the default list.

BTW:
Shouldn't we also rename the list "Internal caches" to "User defined caches" to be consistent?

@moving-bits 
In case this is no bug it could be considered as a feature request. I meanwhile use the UDC quite a lot to remember locations or plan caches and I always need to search for the UDCs as I expect them on the "Internal caches" list.

fm-sys:
fix #8678

One question though: I found the following comment: 
```
// temporary workaround for on demand migration of the old "go to" history
// should be removed after some grace period (maybe summer 2020?)
```
Should I remove the migration code along with my changes, or should we leave it inside for a few more months?


moving-bits:
> Should I remove the migration code along with my changes, or should we leave it inside for a few more months?

I would leave it in for now - it's not that much code, and people are migrating pretty old database backups sometimes...

PR #9317, opened 2020-10-30
select target on geo intent (fix #9290)

### LINKED ISSUE #9290: Improve handling of incoming navigation intents
Inspired by #9289

On incoming navigation intens we could let the user decide how to handle the coordinates and let them choose between
- Save in "Go to"
- Create new UDC and save it as cache coordinate
- Add to an existing UDC and add an additional waypoint to it

This would e.g. further improve the handling of Adventure Labs in c:geo but also suitable for a lot of other use cases (e.g. save some topic bound POI into c:geo (e.g. some needed POI grouped by location).

moving-bits:
image

First entry is always "Create new cache", second always "Go to targets", followed by the other already existing UDC, sorted by last-changed-first.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1116,6 +1116,7 @@
 
 
     <string name="create_internal_cache">Create user-defined cache</string>
+    <string name="create_internal_cache_short">New cache</string>
     <string name="internal_cache_default_name">User defined cache #%1$d</string>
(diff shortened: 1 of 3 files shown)


bekuno:
https://ci.cgeo.org/job/cgeo%20pull%20request/3914/checkstyleResult/source.-6884541216426645898/#2395

PR #9321, opened 2020-10-31
show character counter starting from 3000 characters

### LINKED ISSUE #7390: Log posting with max text length fails
- Compose a log for a gc.com cache
- Enter as much text as possible as log text (until no further characters are accepted)
- Try to send the log online

Logging immediately fails with the corresponding c:geo failure window.
No long posted online.

Max characters allowed for log in c:geo seems not to match what gc.com supports (I assume).
As soon as I reduce the log text by four characters the logging works smooth.

##### Version of c:geo used:
2019.03.17-RC (current `release` branch)

Yes

##### System information:
Android 5.0.1, Samsung Galaxy S4

Seems **not** related to recent logging change #7388

fm-sys:
fix #7390

### DIFF
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -36,14 +36,19 @@ import cgeo.geocaching.utils.CalendarUtils;
 import cgeo.geocaching.utils.CollectionStream;
 import cgeo.geocaching.utils.ContextLogger;
 import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.ViewUtils;
 
 import android.R.string;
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -74,6 +80,7 @@
 public class LogCacheActivity extends AbstractLoggingActivity {
 
     private static final String SAVED_STATE_LOGENTRY = "cgeo.geocaching.saved_state_logentry";
+    private static final int LOG_MAX_LENGHT = 4000;
```
moving-bits, line 83: typo: LOG_MAX_LENGTH

fm-sys: Thanks for spotting it, fixed!

Issue #9282, opened 2020-10-25
Super-size routing distance display (optional)


moving-bits:
While being on a cache tour I frequently either set the next waypoint as routing target or open the popup for it. Both of it displays the distance to the selected waypoint.

When I'm using my bike, and the way gets a bit rougher, it's pretty hard to read the distance as the display gets shaken - so I probably need to stop, read the distance, and move on. "Proximity notification" or "talking compass" only help in some cases. (eg: You do not want to have your loudspeaker on in all cases and/or the surrounding may be too load for the text messages to be heard.)

So what I what like to see is a map overlay which displays the distance text ("120m") in super-size letters (like a quarter of the screen). This option should be switchable using the three dot menu in map mode, default to off.

Lineflyer:
Couldn't this be better triggered by tapping the distance shown on upper right when the direction/navigation line is active?
It would somehow be more consistent then on the popup or using some menu.

Issue #9256, opened 2020-10-22
Distance for Individual route not properly calculated 


MagpieFourtyTwo:
**Describe the bug:**
There is still something wrong with the distance shown for Individual route: Although the value seems to be correct while *adding* points to the route, it will not be decreased, when points are *removed* from the route.

**To Reproduce:**
Imagine there are three caches (1), (2), (3), with 200 m walking distance between them
1. go to map page (no matter which map product)
2. long click on cache (1) to start an Individual route
3. long click on cache (2) to add it to the route => route distance: [200 m]
4. long click on cache (3) to add it to the route => route distance: [400 m]
5. long click on cache (3) to remove it from route => distance remains the same: [400 m]
6. long click on cache (3) to again add it to the route => route distance: [600m]

Every time you now remove and add cache (3), another 200 m will be added to the distance.
BTW: After removing all points from the route via long clicks, the distance will still be shown,
nevertheless, the menu entry "Individual route" => "Clear individual route" will not be shown anymore,
so there seems to be no possibility, to clear the route (as it's not there anymore, anyway ;) ).
Just "seems", as leaving and re-entering map will correct the distance resp. remove the distance display.

**Version of c:geo used:**
2020.12.21-NB-13e8311

**Is the problem reproducible:**
Yes

**Screenshots:**
If scenario is not obvious enough, I can even provide a screen record.

**System information:**
```
Device: VOG-L29 (VOG-L29EEA, HUAWEI)
Android version: 10
Android build: VOG-L29 10.1.0.156(C431E19R2P5)
c:geo version: 2020.10.21-NB-13e8311
Google Play services: disabled - 20.39.15 (120400-335085812)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide caches: own/found archived
Hide waypoints: -
HW acceleration: enabled (default state)
System language: en_DE
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (74,0 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (74,0 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (74,0 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (307,5 MB) on system internal storage
[...]

PR #9324, opened 2020-10-31
refactor map distance display

moving-bits:
While working on #9282 I've stumbled upon the old, manual implementation of drawing distance infos in `CGeoMap`. Refactored this into using a couple of `TextView` elements, and merged layouts and drawing handling for all maps into one.

image

This is a prerequisite for #9282.

### DIFF
--- a/main/res/layout/map_distanceinfo.xml
+++ b/main/res/layout/map_distanceinfo.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content" >
+
+    <TextView
+        android:id="@+id/distance1"
+        android:layout_marginTop="-8dp"
+        android:layout_alignParentTop="true"
+        android:layout_alignParentRight="true"
+        android:layout_width="100dp"
+        style="@style/map_distanceinfo" />
+
+    <TextView
+        android:id="@+id/distance1info"
+        android:layout_marginTop="-8dp"
+        android:layout_toLeftOf="@id/distance1"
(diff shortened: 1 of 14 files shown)


Lineflyer:
Two questions for this:
1. I have to admit I do not understand what is changed here (reading your explanation....cannot read the code :( ). Any visual or functional change or "just" refactoring?
2. This made me think about another observation (slightly offtopic): While updating the manual I have seen, that on Google maps the individual route length is no shown on Google Maps (See screenshot on https://manual.cgeo.org/de/brouter#individuelle_route). Is this a bug or a missing setting? (Feel free to open a dedicated issue in case of bug).

moving-bits:
> 1. I have to admit I do not understand what is changed here (reading your explanation....cannot read the code :( ). Any visual or functional change or "just" refactoring?

So far it's just refactoring. `CGeoMap` did not make use of `TextView` overlays put on top of the map, but actually painted the distances onto the map's surface (for reasons unknown to me). So I changed this to using the same mechanism `NewMap` uses and, along the way, extracted the common code parts into separate files, reducing code duplication.

Changing it to a `TextView` is also a prerequisite for #9282, as only when having the display in a separate GUI item (`TextView` in this case) I can attach an `onClick` handler to it for switching settings.

> 2. This made me think about another observation (slightly offtopic): While updating the manual I have seen, that on Google maps the individual route length is no shown on Google Maps (See screenshot on [manual.cgeo.org/de/brouter#individuelle_route](https://manual.cgeo.org/de/brouter#individuelle_route)). Is this a bug or a missing setting? (Feel free to open a dedicated issue in case of bug).

For me it's shown here with 2020.10.29, just not updated on removal of an item - that's #9256, already fixed on `master`. (AFAICS the individual route's length is shown, but as you have no navigation target set, there's still only one distance being displayed.)

PR #9329, opened 2020-11-01
supersize distance view (fix #9282)

### LINKED ISSUE #9282: Super-size routing distance display (optional)

moving-bits:
Optionally display current distance to navigation target in a super-sized font at the maps' head:

image

Supersize display can be switched either by using a menu item (Map Settings => Supersize distance) or by tapping on the distance shown in the upper right corner.

### DIFF
--- a/main/res/drawable/icon_bcg.xml
+++ b/main/res/drawable/icon_bcg.xml
@@ -5,9 +5,9 @@
     <solid android:color="#80000000" />
 
     <corners
-        android:bottomLeftRadius="8dp"
-        android:bottomRightRadius="8dp"
-        android:topLeftRadius="8dp"
-        android:topRightRadius="8dp" />
+        android:bottomLeftRadius="2dp"
+        android:bottomRightRadius="2dp"
+        android:topLeftRadius="2dp"
+        android:topRightRadius="2dp" />
 
 </shape>
--- a/main/res/layout/map_distanceinfo.xml
+++ b/main/res/layout/map_distanceinfo.xml
@@ -44,4 +44,13 @@
         android:layout_alignParentRight="true"
         android:layout_width="100dp"
         style="@style/map_distanceinfo" />
+
+    <TextView
+        android:id="@+id/distanceSupersize"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_alignParentTop="true"
+        style="@style/map_distanceinfo_supersize" />
+
 </RelativeLayout>
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -75,6 +75,7 @@
     <string translatable="false" name="pref_excludeWpVisited">excludeWpVisited</string>
     <string translatable="false" name="pref_hide_track">hideTrack</string>
     <string translatable="false" name="pref_showCircles">showCircles</string>
+    <string translatable="false" name="pref_supersizeDistance">supersizeDistance</string>
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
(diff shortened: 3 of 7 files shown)


fm-sys:
Why does the supersize distance view have rounded corners on the left but not on the right side?  IMHO rounding off none of them is the way to go...

moving-bits:
Good catch - it's the styling I've reused from the small distance views.  I will change it.

Lineflyer:
@moving-bits 
I don't get the content of the screenshot in full extend:
There is a small distance (routed) showing 820m. Your supersize is showing 492m. Why is it different? 
In case you can supersize different of that values (routed, straight) shouldn't we include the icon in front to make this clear to the user?

My opinion would be to ommit another setting. Clicking on it to supersize and revert should do the job, don't you think?
I am thinking this way, because our menus have to many items already.

moving-bits:
We can show up to three distances currently, depending on your settings and data:
- distances for navigation target
  - straight distance (if configured or no routed distance available)
  - routed distance (if brouter available and straight distance within configured max distance)
- length of individual route

"Supersize" does show the first of the navigation target distances only (which is the straight distance, if no brouter is available, or if you have configured to show both straight and routet distance).

I can
- either add the symbol to the supersize window,
- keep all small windows (including their symbols, if any) additionally to the supersize window (that one without symbol)
- or close all small navigation target windows on opening the supersize window.

moving-bits:
> My opinion would be to ommit another setting. Clicking on it to supersize and revert should do the job, don't you think? I am thinking this way, because our menus have to many items already.

I'm undecided with this one. On the one hand I see the risk of overcrowding menus, on the other hand c:geo already has too many hidden functions which are pretty hard to find for the user. Hmm 🤔

Lineflyer:
> I'm undecided with this one. On the one hand I see the risk of overcrowding menus, on the other hand c:geo already has too many hidden functions which are pretty hard to find for the user. Hmm 🤔

You are right, but this is a pure visual on/off feature (a bit comparable to zoom the map or alike) and not a generic setting (which stays permanent). I would not include a setting option for a transient setting like this (and I also don't think we need it as permanent preference which would be the alternative).

moving-bits:
Updated the PR / removed the menu entry

Lineflyer:
@moving-bits 
Now that you removed the setting (Thanks!) my open question would be:
Does this anti-supersize if you click on it or how did you handle disabling?

Lineflyer:
Include on `master` before beta?

moving-bits:
> Does this anti-supersize if you click on it or how did you handle disabling?

Yes

> Include on `master` before beta?

Yes :-)

geocachermgo:
Note: with google map or satelit the supersize setting by click on the number is not changed until I move the map. With OSM or offline map the change is done immediatly.

moving-bits:
> Note: with google map or satelit the supersize setting by click on the number is not changed until I move the map. With OSM or offline map the change is done immediatly.

This is one of several inconsistencies between the two map types, since they have pretty different technical architectures internally. Hopefully most of this gets healed after merging GM and OSM into the (as-much-as-possible) same architecture as proposed by @rsudev.

PR #9333, opened 2020-11-01
 fix #9332: remove all restrictions for initial map zoom

### LINKED ISSUE #9332: [Nightly] Low map zoom is always reset to specific zoom level
**Describe the bug:**
A short while ago we had the problem that zoom levels >100 m (i. e. 200 m, 500 m, 1 km ...) were reset to 100 m in subsequent map calls, which has been corrected with #9293. But while this is working perfectly fine, it's now (or perhaps even then?) the other way round: zoom levels <100 m (i. e. 50 m, 20 m, 10 m ...) are reset to 100 m in subsequent OSM map calls ...

In GM it's nearly the same, only difference: in GM and GSat the magic limit is even 200 m.

**To Reproduce:**
1. open any map view (no matter if Live map or cache map)
2. zoom until the scale shows "50 m"
3. leave map
4. re-enter map

If OSM was selected, zoom scale will be reset to 100 m, with GM/GSat it will even be reset to 200 m

Zoom should be constant over subsequent calls. 

**Version of c:geo used:**
2020.11.01-NB-cd04617

Yes

eddiemuc:
fix #9332: remove all restrictions for initial map zoom, comment code for future coders taggling this area

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -64,11 +64,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 public class Settings {
 
     /**
-     * On opening a map, we limit the _initial_ zoom. The user can still zoom out afterwards.
-     */
-    private static final int INITIAL_MAP_ZOOM_LIMIT = 16;
-
-    /**
      * Separator char for preferences with multiple elements.
      */
     private static final char SEPARATOR_CHAR = ',';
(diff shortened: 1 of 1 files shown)


fm-sys (changes requested):
If we decide to remove the max zoom level completely, it doesn't make any sense to get the maximum of 0 and the previous zoom level. Or did I missed something? But the current code looks a bit over complicated...

eddiemuc:
> If we decide to remove the max zoom level completely, it doesn't make any sense to get the maximum of 0 and the previous zoom level. Or did I missed something? But the current code looks a bit over complicated... 

Ok I can simplify this, hopefully tomorrow

eddiemuc:
> > If we decide to remove the max zoom level completely, it doesn't make any sense to get the maximum of 0 and the previous zoom level. Or did I missed something? But the current code looks a bit over complicated...
> 
> Ok I can simplify this, hopefully tomorrow

Done

fm-sys (approved):
👍

PR #9337, opened 2020-11-01
extract own wp with empty coords from PCN (fix #9338)

### LINKED ISSUE #9234: Add wayoint's personal note data to exported GPX
Waypoints in cgeo support a free form text field, similar to the "Personal Cache Notes" at the geocache object level. Upon Export->GPX, the latter goes into the GPX file but the former does not.

Add the personal notes data that is associated with a waypoint in cgeo to the exported GPX.

**Additional context**
Let me use this scenario to open a broader discussion: What is the intent of providing the capability to add custom waypoints to a geocache object in cgeo? The waypoint has similar properties than those that come down from GC.com but their sole purpose seems to be limited to using it on cgeo itself. There is no way back to geocaching.com (or any other such platform I believe).

In my workflow, a tool lilke cgeo is merely a locally cached copy of data that is authoritatively stored at geocaching.com. The cached data makes sense in situations where I don't have mobile data, out in the field on a geoaching.trip. Usually, if I make any changes out in the field. e.g. changing corrected coordinates, adding or changing personal cache notes, I always reflect that back to geocaching.com because that is the one and only place where it authoritatively belongs to. Why is that so relevant with the "autoritative source"? Like many others I use multiple different tools to access and manipulate data from geoaching.com, e.g. GSAK, Locus Map, cgeo and the list may go on. They all have in common: They depend on one single place where they fetch their data from and they all go to geocaching.com to get it from there.  Whatever such a tool modifies or adds, it must send this data back to geoaching.com where all other tools will "refresh" their local copy of the data in regular time intervals or as needed and triggered by me. Each and every other workflow will undoubtedly lead to different data on different devices and now way to ever get it back in sync again.

Back to the original topic: If I add a new waypoint in cgeo, there doesn't seem to be any way to reflect that back to geoaching.com. Is that a limitation in geocaching.com? At least their API supports user created waypoints:

Create / Update / Delete User Wayoint

To close the loop: Since there is no way to use the data in the personal notes field of a waypoint in a programmatic way, "Export -> GPX" seems to be the only way to get to it but only once the export function adds it to the GPX.

### LINKED ISSUE #9338: Export waypoint with empty coords from PCN
For a custom waypoint that has been configured to solve multicaches with the **waypoint calculator** using the **Plain** option, I can use c:geo to successfully write to the Personal Notes and upload them to GC.com, for example:

```
{c:geo-start}
@Formula (F) (NO-COORD)
"N50 57.(A+E)(B+B)(E) W001 21.(D)(B+D)(A+B+B)
telephone number 
023806ABCDE
short walk
Checksum 30"
{c:geo-end}
```

Which is fantastic but when using c:geo on another device and downloading this cache, my expectation was that the same custom waypoint would be recreated ready for action in the field.

Unfortunately, the read process fails to populate this waypoint. I am guessing the reason is because the coordinates are not valid, that is, they are in the format of a valid formula format for the waypoint latitude and longitude but not one of the excepted valid coordinate formats:

```
Plain
N50 57.(A+E)(B+B)(E)
W001 21.(D)(B+D)(A+B+B)
A=|a+b| D=|a+b|
B=|a+b| E=|a+b|
Note ...
```

I see this style of waypoint calculations to be the most common usage, so it will be a shame if this cannot be automated to create this waypoint on load rather than going through copy/paste manual configurations, which is doable ... but a pain ...

If this was fixed, we could also do our planning in the cache page, updating the Personal Notes in the format template:

```
{c:geo-start}
@Title (F) (NO-COORD)
"Valid Lat/Long Formula for calculator
additional notes"
{c:geo-end}
```

Or in whatever format that works best for the WPC and WP authors.

Then, we load our GPX into c:geo and are good to go.

_Originally posted by @Billy-Wallace in https://github.com/cgeo/cgeo/issues/9234#issuecomment-713897472_

-> see #9211 for parsing formula

this issue is for extract user waypoint with empty coordinates from PCN (it will be already written to PCN, but no waypoint is created).
Advantage: no dummy-coordinates have to be added to the waypoint.

murggel:
This PR improves the extraction of waypoint from PCN for own waypoint with empty coords:
Own waypoint with empty coords will be extracted. 
If already a waypoint with same name and type exists, the new waypoint will not be added.
(fix #9338)
So user can add waypoint with formula in PCN and it will be extracted to user-note of waypoint:
e.g. 
@Final (F) (NO-COORD)
"N50 57.(A+E)(B+B)(E) W001 21.(D)(B+D)(A+B+B)"

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -1589,8 +1589,8 @@ public class Geocache implements IWaypoint {
         boolean changed = false;
         if (forceExtraction || !preventWaypointsFromNote) {
             for (final Waypoint parsedWaypoint : Waypoint.parseWaypoints(StringUtils.defaultString(text), namePrefix)) {
-                final Waypoint existingWaypoint = findWaypoint(parsedWaypoint.getPrefix(), parsedWaypoint.getCoords());
-                if (existingWaypoint == null) {
+                final Waypoint existingWaypoint = findWaypoint(parsedWaypoint);
+                if (null == existingWaypoint) {
                     //add as new waypoint
                     addOrChangeWaypoint(parsedWaypoint, updateDb);
                     changed = true;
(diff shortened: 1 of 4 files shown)


murggel:
Unfortunately I linked the commits to #9234 instead to #9338.
Probably somebody can help me to fix this?

rsudev:
You can amend the commits on your side, correcting the commit message and pushing (with --force) to the same branch, which will update them in this PR. That the branch name also contains the wrong issue is of less importance, as it will no longer be related to the commit once it is merged.
(Was that your question?)

murggel:
@rsudev thanks, I will try this.

moving-bits:
Or if it's about the second commit belonging to a different PR you can try the following:

For the new PR:
- gather the commit ID (SHA hash) from the second (= more recent) commit
- `git checkout master` to select new base
- `git checkout -b new_issue` to create a new branch based on that
- `git cherry-pick <commit-ID>` to select the changes from the commit you selected

and the do a regular `git push ...` on that branch.

For the old PR, to remove the extra commit:
- `git checkout <old-issue>`
- `git reset HEAD~1` to undo the last changes, they are now in an uncommitted state
- `git push --force` to push the new state (without the uncommitted files)

You may either `git rm ...` (remove) the uncommitted changes or `git stash` them away.

murggel:
@rsudev and @moving-bits Thanks for the hints. The amend and the push-force helped.

PR #9342, opened 2020-11-02
WIP Fixes for Sailfish OS

osma:
DO NOT MERGE

This PR disables the transparent background effect on MainActivity which causes problems on recent versions of Sailfish OS  on Xperia XA2 and (probably) Xperia 10 series devices (#8752).

The configuration screens are still freezing when you navigate between hierarchy levels, but that problem can be worked around by using a gesture - a swipe from the side of the screen and back - to almost put the app in the background but then cancel it by swiping back. This needs to be repeated many times but after configuring the app can be used normally.

### DIFF
--- a/main/res/values/themes.xml
+++ b/main/res/values/themes.xml
@@ -8,12 +8,6 @@
 
 
     <style name="cgeo_main.base" parent="@style/Theme.AppCompat">
-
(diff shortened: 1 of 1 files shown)


bekuno:
ok to test

bekuno:
@osma can this be closed now after merging #9517 ?

osma:
Sure, closing.

PR #9344, opened 2020-11-02
avoid crash when trying to log a PM cache as basic member

### LINKED ISSUE #9318: crash when logging a unloaded PM only cache as BM from search
**Describe the bug:**
Using a non- premium account, I enter in GC# of a premium cache and receive "error" message. But the option to log cache. Clicking "log" shuts down the app. 

**To Reproduce:**
1. From basic account, Go to search and type in GC# of premium cache. Search. 
2. Dialog box appears for Premium Member Only. 
3. Click LOG VISIT

App Crashes. 

A dialog box should appear to allow a basic membership to log a Premium Cache. 

**Version of c:geo used:**
2020.10.29

Yes

**Screenshots:**

Screenshot_20201030-202521_cgeo

**System information:**
```
--- System information ---
Device: SM-S767VL (j7topeltetfnvzw, samsung)
Android version: 9
Android build: PPR1.180610.011.S767VLUDS8BTI2
c:geo version: 2020.10.29
Google Play services: enabled - 20.39.15 (100300-335085812)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide caches: own/found 
Hide waypoints: original 
System language: en_US
System date format: M/d/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (1.0 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (1.0 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (1.0 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (4.1 MB) on system internal storage
Offline maps path: /storage/emulated/0/cgeo/maps
Map render theme path: 
Live map mode: true
Geocaching sites enabled:
Geocaching.com date format: MMM/dd/yyyy


**Additional context:**

### PR DESCRIPTION
**Pull request by fm-sys**
fixes #9318

Parse Type, Name, Geocode, Difficulty, Terrain and Size for PM only caches - needed for corectly displaying the logCacheActivity and for supporting log templates...

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -168,6 +168,13 @@ public final class GCConstants {
 
     static final String STRING_PREMIUMONLY_2 = "Sorry, the owner of this listing has made it viewable to Premium Members only.";
     static final String STRING_PREMIUMONLY_1 = "class=\"illustration lock-icon\"";
+    static final Pattern PATTERN_PREMIUMONLY_CACHETYPE = Pattern.compile("/app/ui-icons/sprites/cache-types\\.svg#icon-([^\"\\-]+)-?([^\"]+)?");
+    static final Pattern PATTERN_PREMIUMONLY_CACHENAME = Pattern.compile("<h1 class=\"heading-3\">(.+)</h1>");
+    static final Pattern PATTERN_PREMIUMONLY_GEOCODE = Pattern.compile("<li class=\"li__gccode\">([^<]+)");
+    static final Pattern PATTERN_PREMIUMONLY_DIFFICULTY = Pattern.compile("<span id=\"ctl00_ContentBody_lblDifficulty\"(?:.|\\s)*?<span>([0-5](?:[\\.,]5)?)</span>");
+    static final Pattern PATTERN_PREMIUMONLY_TERRAIN = Pattern.compile("<span id=\"ctl00_ContentBody_lblTerrain\"(?:.|\\s)*?<span>([0-5](?:[\\.,]5)?)</span>");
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by eddiemuc — APPROVED** [review]
This is really cool! Thanks for contribution

**Comment by eddiemuc** [comment]
As a (probably not intended but imho good) side effect, now c:geo can also parse PM caches when not logged in. This makes a test fail in CI however where a search without logged in expects an empty result (for GC2JVEH)

File: main/src/cgeo/geocaching/connector/gc/GCConstants.java [hunk 1]
```diff
@@ -168,6 +168,13 @@
 
     static final String STRING_PREMIUMONLY_2 = "Sorry, the owner of this listing has made it viewable to Premium Members only.";
     static final String STRING_PREMIUMONLY_1 = "class=\"illustration lock-icon\"";
+    static final Pattern PATTERN_PREMIUMONLY_CACHETYPE = Pattern.compile("/app/ui-icons/sprites/cache-types\\.svg#icon\\-([^\"\\-]+)-?([^\"]+)?");
+    static final Pattern PATTERN_PREMIUMONLY_CACHENAME = Pattern.compile("<h1 class=\"heading\\-3\">([^<]+)");
```
bekuno, line 172: This will not work when ^ or < are part of the cache name.
I tested with "<" - this is possible.
What about `<h1 class="heading-3">(.+)<\/h1>` ?
(plus extra \ of cause)

Interesting question: is or was a cache name eg. `head </h1> or <h1> not?` possible?
At least today the html tags were filtered out on my tests creating such cache name.

bekuno (approved):

main/src/cgeo/geocaching/connector/gc/GCConstants.java
fm-sys: You are totally right! Fixed.

PR #9351, opened 2020-11-05
related to #8457: let Mapsforge use Uris instead of Files

eddiemuc:
related to #8457: let Mapsforge use Uris instead of Files
Refactoring towards to SAF, no functional change: let MapsforgeMapProvider use Uris instead of Files


Lineflyer:
Could it be that after upgrading to a version containing this change the previous sele ted offline map is lost and the app comes up with GMaps instead (One time problem)?

eddiemuc:
Hm I tried to introduce the change in a way that this doesn't happen, but I can't exclude it. Would be a bug imho

Issue #7306, opened 2019-02-18
'Star' characters are displayed with odd colours/backgrounds


alan666notb:
I see some strange font issues with the latest beta release....

##### Detailed steps causing the problem:
- load GC6YWFT  [Church Micro 10381...Sanderstead - URC]
- scroll to bottom of cache page and observe the 'star' characters above and below the "For full information ....." text.

##### Actual behavior after performing these steps:
'Stars' are shown as red font on black background above the text and blue font on black background below the text.

##### Expected behavior after performing these steps:
On the previous release (and on website), all of the stars were red font on white background.

##### Version of c:geo used:
2019.02.17-RC

##### Is the problem reproducible for you?
Yes

alan666notb:
Screen shots ....

Old
![screenshot_old](https://user-images.githubusercontent.com/7459480/52959089-53274a80-338d-11e9-92cf-e18ea4d1d7d5.png)

New
![screenshot_new](https://user-images.githubusercontent.com/7459480/52958968-1d826180-338d-11e9-855d-77590d20f5c5.jpg)

Lineflyer:
This is IMHO caused by the new feature for contrast optimization to improve bad colored text on our black background.

I fear the implementation was not yet really tested with white background.

SammysHP:
It behaves exactly as intended. Fully saturated blue and red are "hard to read on white background", thus they get a dark background. I already mentioned these problems back then.

githubber2222:
How would the new look be easier to read? Maybe my eyes are different but the old one is clearly better.
Also, the text above the 'blue stars' is blue as well but no black background.

Lastly, the listing has the bottom row of stars in red, not blue.

@SammysHP your response looks like a lame excuse here, sorry. You intended that red turns blue? And visually same shade of blue above has NO black background?

SammysHP:
Oh, I haven't noticed that originally both rows had the same color.

Different parsing for visually the same color is possible because not all ways of changing the color are handled.

okainov:
Seems interesting (changing the foreground color), thia should not happen. may be some issue in  source cache formatting... I can try to take a look and debug a bit when I'm back from my trip

Upd: tried to open the cache, but it's premium-only, so I don't have access and cannot check it. If you find any other cache with such issue, please report.

alan666notb:
Here's a non-premium church-micro that shows the same problem: GC660KN

Could the whole 'contrast optimisation' thing not be turned off when using the light scheme?

[...]

PR #9359, opened 2020-11-08
use HtmlCompat.fromHtml instead of deprecated Html.fromHtml

bekuno:
Since API24 the plain Html.fromHtml(string) is deprecated in favor of Html.fromHtml(string, int).
There is a wrapper class HtmlCompat where the method switch is done according to the Build version of the device.
The implementation uses now only the int FROM_HTML_MODE_LEGACY (value 0).

Maybe we can optimize later the code using some of the new parameters to solve issue #7306.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -342,6 +342,9 @@ dependencies {
     // Support Library AppCompat
     implementation 'androidx.appcompat:appcompat:1.2.0'
 
+    // Support Library core (used for eg. HtmlCompat)
+    implementation 'androidx.core:core:1.3.2'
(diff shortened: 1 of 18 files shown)


Lineflyer:
@bekuno Do you think we should include this for the upcoming beta? As far as I understand it should be identical while running in legacy mode?

bekuno:
With this PR the code will only changed to a non deprecated form. There will be no functional change.
This can go to the next beta.

Lineflyer:
Lets try it out in nightlies then.

PR #9372, opened 2020-11-12
toggle supersize distances (fix #9367)

### LINKED ISSUE #9367: [Nightly] Extend Super-Sized routing distance display
Just checked out the new #9282 "Super-Size routing distance" feature - and really like it. Great idea!
Nevertheless was a bit disappointed, cause - contrary to the promising title - "SuSi" ;) unfortunately does not show the *routing* distance, but just the bee line's length (aka "straight distance").
So, how about extending it to display the *real* routing distance, too (as the title already pretended)?

As tapping one of the two (resp. up to three) different (small) distance display areas to activate SuSi is most probably way too fiddly, esp. in the initially described scenario, the "activation area" should be extend to all displayed distance value fields and then it possibly can just cycle through the distances, i. e. ...
1. tap on (small) bee line's length, routing distance or length of ind. route => SuSi straight distance
2. tap on SuSi straight distance => if routing is active, then show SuSi routing distance, else switch SuSi off
3. tap on SuSi routing distance => switch SuSi off

Length of individual route may safely be ignored here, as it does not change at all, thus has no influence on driving, anyway.

And, last but not least, to distinguish which distance SuSi is currently showing, she maybe even display the symbols from straight distance resp. routing distance.

moving-bits:
- make all (up to) three tiny distance fields clickable
- add toggle cycle: first distance supersized => second distance supersized (if available) => no distance supersized

### DIFF
--- a/main/src/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
+++ b/main/src/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
@@ -7,6 +7,8 @@ import cgeo.geocaching.settings.Settings;
 import android.view.View;
 import android.widget.TextView;
 
+import org.apache.commons.lang3.StringUtils;
+
 public class MapDistanceDrawerCommons {
     private static final String STRAIGHT_LINE_SYMBOL = Character.toString((char) 0x007C);
     private static final String WAVY_LINE_SYMBOL = Character.toString((char) 0x2307);
@@ -17,6 +19,7 @@ public class MapDistanceDrawerCommons {
     private final TextView distance2View;
     private final TextView routeDistanceView;
     private final TextView distanceSupersizeView;
+    private boolean bothViewsNeeded = false;
 
     public MapDistanceDrawerCommons(final View root) {
         distance1InfoView = root.findViewById(R.id.distance1info);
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
```diff
@@ -28,27 +31,34 @@ public MapDistanceDrawerCommons(final View root) {
...
+        routeDistanceView.setOnClickListener(v -> swap());
         distanceSupersizeView.setOnClickListener(v -> swap());
     }
 
     public void drawDistance(final boolean showBothDistances, final float distance, final float realDistance) {
-        final boolean bothViewsNeeded = showBothDistances && realDistance != 0.0f && distance != realDistance;
-        final boolean supersize = Settings.getSupersizeDistance();
-        final TextView d = supersize ? distanceSupersizeView : distance1View;
+        bothViewsNeeded = showBothDistances && realDistance != 0.0f && distance != realDistance;
+        final int supersize = Settings.getSupersizeDistance() % (bothViewsNeeded ? 3 : 2);
```
eddiemuc, line 42: I have a bit of a bad feeling with these "magic" supersize numbers (0, 1 and 2). Is it an option to use an enum instead.
Also, the code seems like it would be vulnerable if "bothViewsNeeded" is switched from true to false and vice versa in consective cars. I think this is actually not the case since a previously stored supersize value of 2 would be reduced to 0 in case bothViewsNeeded=false, which as far as I understand the code triggers the "correct" behaviour of not displaying a supersize. But there is some hidden magic in there which is not obvious imho.

moving-bits: Thanks for cross-checking. Those "magic numbers" are actually nothing more than a plain counter - first (1), second (2) or no (0) distance window being super-sized. "Second" only being possible if "bothViews" setting is activated and a calculated distance is available (resulting in `bothViewsNeeded == true`). And, exactly, fallback is intended as you've described. - Ok for you, or shall I change this to either using named constants or an enum class?

eddiemuc: ok for me, leave as is

eddiemuc (approved):

PR #9377, opened 2020-11-13
make cache share actions more consistent (fix #9339) 

### LINKED ISSUE #9339: Copy geocache link from geocache page
**Describe the bug:**
Copy a link for geocache if a geocache is open.
Since some update it is not possible any more to copy a link to the open geocache. If you select the cache on the map it is no problem to copy a link for this cache. But there open different dialogs.
It is possible to include this feature again to get a link to this geocache?

**To Reproduce:**
1. Open the geocache
2. Select 3 dots in right upper corner
3. Redirect >

Copy to clipboard is missing

There should be an entry with copy link to clipboard

**Version of c:geo used:**
2020.10.29

**System information:**
OnePlus 5
Android 10

cayacdev:
Issue #9339

### DIFF
--- a/main/res/menu/cache_options.xml
+++ b/main/res/menu/cache_options.xml
@@ -91,7 +91,6 @@
         android:id="@+id/menu_share"
         android:icon="@drawable/ic_menu_share"
         android:title="@string/cache_menu_share"
-        app:actionProviderClass="androidx.appcompat.widget.ShareActionProvider"
         app:showAsAction="ifRoom">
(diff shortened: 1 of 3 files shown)


rsudev:
ok to test

fm-sys (approved):
Tested it and it's working perfectly fine. Thanks for fixing this - very welcomed!

PS: don't forget to remove the draft flag when you're PR is ready... ;-)

PR #9384, opened 2020-11-14
auto-adjust super-sized distance font (fix #9380)

### LINKED ISSUE #9380: Supersized routed distance to big
If I supersize the routed distance the string is cut of (probably as the routed graphic in front consumes more space.
Not OK:

grafik

OK:
grafik

moving-bits:
- auto-resized font size in a range from 64sp to 120sp
- sets a smaller default font size for supersized distance display (72sp instead of 96sp) (fallback)


Lineflyer:
Also this one merged into master before merging branches. Will test in resulting beta.

PR #9385, opened 2020-11-14
enlarge distance info fields (fix #9381)

### LINKED ISSUE #9381: Routed distance icon no longer shown
Might be related to #9380 

Without using supersize I noticed, that the icon preceding the routed distance is replaced by ...

grafik

moving-bits:


Lineflyer:
I will merge this before merging branches and test with resulting beta later on.

PR #9389, opened 2020-11-15
bottom-align route distance in supersize mode

moving-bits:
follow-up to #9380, see https://github.com/cgeo/cgeo/issues/9380#issuecomment-727269485

bottom-align route distance info view if in supersize mode, to stay in white space.

Only applies if "show straight distance" is active, as otherwise the route distance is the only distance displayed beside the supersized one and can stay top-right in that case.

image

### DIFF
--- a/main/res/layout/map_distanceinfo.xml
+++ b/main/res/layout/map_distanceinfo.xml
@@ -47,6 +47,14 @@
         style="@style/map_distanceinfo" />
 
     <TextView
+        android:id="@+id/routeDistanceSupersize"
+        android:layout_marginTop="0dp"
+        android:layout_alignBottom="@id/distanceSupersize"
+        android:layout_alignParentRight="true"
+        android:layout_width="100dp"
+        style="@style/map_distanceinfo" />
+
+    <TextView
         android:id="@+id/distanceSupersize"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
--- a/main/src/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
+++ b/main/src/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
@@ -18,6 +18,7 @@ public class MapDistanceDrawerCommons {
     private final TextView distance2InfoView;
     private final TextView distance2View;
     private final TextView routeDistanceView;
+    private final TextView routeDistanceViewSupersize;
     private final TextView distanceSupersizeView;
     private boolean bothViewsNeeded = false;
 
@@ -27,6 +28,7 @@ public class MapDistanceDrawerCommons {
         distance2InfoView = root.findViewById(R.id.distance2info);
         distance2View = root.findViewById(R.id.distance2);
         routeDistanceView = root.findViewById(R.id.routeDistance);
+        routeDistanceViewSupersize = root.findViewById(R.id.routeDistanceSupersize);
         distanceSupersizeView = root.findViewById(R.id.distanceSupersize);
 
         distance1InfoView.setOnClickListener(v -> swap());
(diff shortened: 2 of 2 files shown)


MagpieFourtyTwo:
Hmm. Readability is better, of course, but still looks weird.
I would still suggest to leave the grey distance fields in the upper right corner
and to left align the supersized stuff instead.

This would even look more balanced ... in my eyes at least ... or what do you think?

SuSi

moving-bits:
Hm, I'm still a bit reluctant. But I understand the problem you're pointing to.

A centered text looks more balanced (to me, at least), especially depending on how much information you have on this screen else.

The problem arises mainly from having two little distance boxes (because you have "Show straight distance" activated and set an individual route), with the second box overlapping with the distance text (depending on the specific distance shown).

Having only one small distance field solves the overlapping (and makes the text shorter, as no symbols are needed in that case), and as you have the cache title present in the same line as the first small distance field in OSM, it looks very balanced over there.

Or imagine having "Show straight distance" disabled and no individual route - then it would look pretty odd to have the distance left-aligned.

Alas I know of no solution yet which fits all settings. Maybe
- always show the small distance boxes in the upper-right
- center the supersized text as default, but left-align if two small distance boxes are shown

?

MagpieFourtyTwo:
> - always show the small distance boxes in the upper-right
> - center the supersized text as default, but left-align if two small distance boxes are shown

Sounds good to me.

eddiemuc:
Suggestion:
* place the small distance boxes in a row (align them horizontally, not vertically), left-to-right
* Place supersize box under that row and center it

Lineflyer:
> place the small distance boxes in a row (align them horizontally, not vertically), left-to-right

That might not work as the first row might show the cache name if it is set as route target.

I initially suggested just to place the supersized distance below all other distances on the screen.
What about that? I dont see an essential need to have it on top of the screen.

eddiemuc:
> > place the small distance boxes in a row (align them horizontally, not vertically), left-to-right
> 
> That might not work as the first row might show the cache name if it is set as route target.

I took a look how Locus is doing it. If I "transfer" this behaviour to c:geo I would suggest to move down the (horizontally aligned) distance boxes in case there needs to be a cache name to display.

But I have no strong opinion about the gui in general, so I am fine with any other solution as well

Lineflyer:
I was thinking of the following principle, when I mean to append the supersize below everything else.

grafik


Its only a quick and dirty draft to show what I mean: Don't interfer with existing and do not change existing, but just place the supersize below.

fm-sys:
IMHO it looks kind of weird if SuSi isn't top-aligned. But what about moving all other stuff below it?

ZomboDroid 17112020132841

Lineflyer:
> But what about moving all other stuff below it?

This also looks quite clean IMHO

Whatever is easier to realize, but avoid the overlapping in any case.

MagpieFourtyTwo:
If "SuSi centered" is the way to go, I would prefer @fm-sys' example, too, as it appears to be way cleaner.

Nevertheless, there is a lot of "wasted space" left and right of it, so - even if it may sound a bit self-indulgent (which is most definitely not the case ;) ) - I still like @moving-bits' solution, i. e.

> - always show the small distance boxes in the upper-right
> - center the supersized text as default, but left-align if two small distance boxes are shown

based on [my example pic ](https://user-images.githubusercontent.com/9889015/99305977-cefd4200-2854-11eb-83dd-28d127064794.png)way better, as it makes better use of screen ressources (which is still an issue on a bunch of lady phones). While things are getting even worse, when add. using some overlay apps ... like e. g. the Unspeakable. ;)

> but avoid the overlapping in any case.

Totally agreed.

moving-bits:
Thanks for all the feedback and good ideas! I'll give it a try and see whats most feasible from both implementation and UX side.

moving-bits:
Ok, two alternatives so far:

Alternative a) Put the supersize window at the top, and place all other target and distance infos below:
imageimage

This examples also show that the alternative "put distance view at top right, and left align supersized text" probably won't work without overlapping text, depending on the length of the specific distance info text.

Alternative b) As today, but put the second small distance info below the supersize window:
imageimage

If you have disabled "Show straight distances", alternative b) would look like this:
imageimage

Yet another option could be to display the supersize window at the bottom and shrink the map accordingly, but I haven't found a way to achieve this yet. It should look something like this:

image

My personal favorite would be b)

MagpieFourtyTwo:
a)  -1
b) +1
c) -100 ;)

MagpieFourtyTwo:
b) would BTW even work with imperial units - although the dot of the "i" in "mi" will slightly scratch the grey distance area.
But it's in fact *really* slightly, so it should be tolerable.

Lineflyer:
a) My favorite
b) Can live with that, altthough I cannot understand why you would stack the overlays over another and not sure if nothing will overlap given different devices, fonts, etc.
c) No

eddiemuc:
+1 to @Lineflyer s opinion

eddiemuc:
And many thanks & kudos to @moving-bits for investing so much effort and moderation in finding a solution accepted by the whole community.

PR #9401, opened 2020-11-18
show second distance view below super size view

moving-bits:
- show second distance view below super size view
- move supersized text a bit down within its view to separate it a bit more from top text (target name, distance)


moving-bits:
show straight distance = false:
image

show straight distance = true:
image

Lineflyer:
@moving-bits Thanks for the finetuning.
This should target `release` though.

moving-bits:
> This should target `release` though.

It'd been based on `release` actually, just selected wrong target during PR creation... Have retargeted it now.

PR #9406, opened 2020-11-19
refactor map distance views

### LINKED ISSUE #9381: Routed distance icon no longer shown

moving-bits:
Refactor map distance views:
- target info (OSM only) and non-supersized distance get displayed without extra grey background when in supersize mode
- distances in small views are right-aligned
- distance type indicators are displayed in the same TextView as their distance (fixes #9381)
- lots of code and layout file cleanup

Some screenshots of different combinations:
image image
image image

### DIFF
--- a/main/res/layout/map_distanceinfo.xml
+++ b/main/res/layout/map_distanceinfo.xml
@@ -6,17 +6,22 @@
     xmlns:app="http://schemas.android.com/apk/res-auto">
 
     <TextView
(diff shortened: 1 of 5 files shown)


Lineflyer:
Pretty cool and integrats the new supersize even better
Lets try!

PR #9422, opened 2020-11-21
fix several AlertDialog themes

moving-bits:
Many of our dialogs use the default theme implicitly, not the c:geo theme. This PR fixes this for a couple of `AlertDialog`s by using `Dialogs.newBuilder()` instead of `new AlertDialog.Builder()`.

### DIFF
--- a/main/src/cgeo/geocaching/export/FieldNoteExport.java
+++ b/main/src/cgeo/geocaching/export/FieldNoteExport.java
@@ -11,6 +11,7 @@ import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.LocalStorage;
+import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.AsyncTaskWithProgress;
 import cgeo.geocaching.utils.Formatter;
 import cgeo.geocaching.utils.Log;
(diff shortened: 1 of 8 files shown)


fm-sys:
Have you checked if these dialogs will be displayed correctly if using light theme? Cause we had several issues with invisible text, etc in the past...

moving-bits:
> Have you checked if these dialogs will be displayed correctly if using light theme? Cause we had several issues with invisible text, etc in the past...

Not all of them, but a couple of. The change is about using a `ContextThemeWrapper()` with the appropriate `R.style.Dialog_Alert` or `R.style.Dialog_Alert_light` theme. I'm reusing a method `Dialogs.newBuilder`, which has already been in use for the `Dialogs` package for quite some time, and opened this for other dialogs based on `AlterBuilder`. (At least a couple of them, more to go...)

fm-sys:
As already expected, this PR had introduced new theming bugs for some of the dialogs. Haven't checked all of them yet... 

Screenshot_20201123_095856_cgeo geocaching

moving-bits:
This is truly broken, somehow: `Dialogs.newBuilder` is calling `ContextThemeWrapper` with `R.style.Dialog_Alert_light` if `Settings.isLightSkin` is `true`. And `R.style.Dialog_Alert_light` defines a black `text_color` - so why is the text color white, if the light skin is used in settings...? Strangely the "backup completion" dialog shows the correct colors, although it's using the same `Dialogs.newBuilder` internally...

For now, I'll do a revert for the specific dialog. But it would be really nice to redo styling from scratch, consistent, and based on current (compatibility) technology...

fm-sys:
> But it would be really nice to redo styling from scratch, consistent, and based on current (compatibility) technology...

I will try to look into it tomorrow, if I can get the theming for that dialog working... But you are right, that we should refactor this area totally as a long term solution

moving-bits:
Settings activity is using the `settings` / `settings_light` theme, which is even including `Dialog_Alert` resp. `Dialog_Alert_light` themes - which are the same ones used when calling `Dialogs.newBuilder`. So it *should* not make any difference (but obviously does). One problem may arise from `settings` and `Dialog_Alert` using different sets of parent themes - don't know how compatible this is across different Android versions.

IMHO we should start with a blank base theme in dark & light variant, based on a recent compatibility theme, and define all colors, standard margins, standard font sizes etc. in it. All other themes should be based on those two themes. Yet another large are where c:geo shows off its age :-(

Issue #8138, opened 2020-02-23
Proposal to access certain map settings


moving-bits:
Taken from https://github.com/cgeo/cgeo/issues/8112#issuecomment-590012624 by @MagpieFourtyTwo:

> Just an idea: What would you think about a (maybe semi transparent) icon list, where one can see the enabled resp disabled (e. g. crossed out) waypoint types?
> 
> Moreover I would prefer if the hide menu would not disappear on the very first tap, so one can enable/disable more than just one type in one go? Same applies to the Map Settings menu ... so one can set more than one option at a time - and a tap on the map, resp outside of the menu's area closes it ...?

moving-bits:
> Just an idea: What would you think about a (maybe semi transparent) icon list, where one can see the enabled resp disabled (e. g. crossed out) waypoint types?

Interesting idea. This could also be used for quick enabling / disabling.
OTOH map space is very precious. IMHO we should wait a bit for some feedback how frequently those features are used and then decide, whether to sacrifice some space for those toggles. Map bottom, in the center, might be a feasable place.

> Moreover I would prefer if the hide menu would not disappear on the very first tap, so one can enable/disable more than just one type in one go? Same applies to the Map Settings menu ... so one can set more than one option at a time - and a tap on the map, resp outside of the menu's area closes it ...?

Understood, and sounds reasonable to me. As this behavior would be distinct from the way menu handling works today, those "hide settings window" could be a separate popup dialog which opens after tapping the "hide" menu entry. (Only needed if we decide not to implement toggle buttons as described above.) Similar for the "Map settings".

fm-sys:
I decided continue this discussion rather here instead of #9281

@MagpieFourtyTwo had mentioned the idea of customisable action overlays like it is done in Locus. I really appreciate this idea and would like to add my two (OK, they have become three after all ;-) ) cents:

1. This action bar should be placed to the left and/or right of the screen as all other places are already used for displaying other information.
 
2. As a first step we can can just fill it with the most useful features and leave out the customization part and add it in a second later on IMHO.

3. Activating this overlay bar should be possible via the three dot menu, so it can be easily hidden if it's not needed right now and it takes too much screen space.

Ok, and for the hiding menu I agree that this should be a dialog rather a submenu...

moving-bits:
I've played around a bit to find a way for displaying a popup for the hide settings. This is one possible solution, using a `PopupMenu`:

[...]

PR #9455, opened 2020-11-28
map settings popup (rel. to #8138)

moving-bits:
A first implementation for a map bottom which opens a popup with a couple of map settings - hide options so far.
Implemented for both Google Maps and OSM. Button is placed at the map's bottom, horizontally centered.

"Hide" menu entry is still available for now and opens the same popup.

Several settings can be changed without having to re-open the popup. The changed settings will be applied on closing the popup with the "OK" entry.


fm-sys:
👍

Tested it and had two thoughts:
- can we still show the icons in the popup like it was in the overflow menu before? Cause icons are really useful for quickly finding the right menu without needing to read everything...
- It took a while to understand that I need to confirm the change with the ok button (which currently doesn't really look like a button btw). IMHO we can leave the ok button out completely and apply changes directly...

moving-bits:
Thanks for testing.

> can we still show the icons in the popup like it was in the overflow menu before? Cause icons are really useful for quickly finding the right menu without needing to read everything...

Haven't found a way to make them show up in a `PopupMenu` so far. It's not supported officially with a `PopupMenu`, and the workarounds from SO etc. I've tried did not work for me (or did not even compile).

The only way I see to get this back seems to me to use a `AlertDialog` based popup, which would have two more advantages over a `PopupMenu`:
- it could be placed screen-centered instead of above anchor element (button)
- it could have a "real" button for "OK" instead of just a plain menu entry

Downside is, you have to layout the `AlertDialog` from scratch. I've started doing so with a `RelativeLayout`, but the results so far were not promising... Maybe I need to create a `RecyclerView` instead for the entries.

> It took a while to understand that I need to confirm the change with the ok button (which currently doesn't really look like a button btw). IMHO we can leave the ok button out completely and apply changes directly...

The reason I choose a confirmation button is performance: After changing a setting, caches are invalidated and a redraw of the map is triggered. If you change more than one setting at the same time, the cycle is repeated several times.
I think user interface will be easier to understand if I manage to get a "real" button displayed, see above.

PR #9456, opened 2020-11-29
avoid crash when starting restore dialog from MainActivity

fm-sys:
#9439 had unfortunately introduced a bug, that the backup restore dialog can't be launched from within the MainActivity after a fresh installation without a crash... This problem will be solved with this PR.


bekuno:
Should this better target to release?

fm-sys:
This bug was introduced with #9439 which is only at master, so the PR should be fine as it is...

PR #9458, opened 2020-11-29
rel to #8457: introduce PublicLocalStorage framework, use SAF for offline maps and logfiles

eddiemuc:
rel to #8457: introduce PublicLocalStorage framework, use SAF for offline maps and logfiles


eddiemuc:
No longer needed

PR #9483, opened 2020-12-01
custom tabs optimization

### LINKED ISSUE #9419: Webview - No other browser option if Chrome installed
**Describe the bug:**
In the menu "open in browser" from cache only Chrome is offered.

**To Reproduce:**
1. Go to a cache page
2. Click on menu
3. Chose  "open in browser"

Only Chrome is offered

Chrome and Firefox should be offerend. This happens when I open (via long click) the menu of a single log with the menu item open in browser there. 

**Version of c:geo used:**
2020.11.21-NB-612de32

Yes 

Maybe connected to #6680  ?

fm-sys:
- fix #9419 (make custom tabs configurable)
- use custom tabs also for user context weblink actions

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -160,6 +160,7 @@
     <string translatable="false" name="pref_memberstatus">memberstatus</string>
     <string translatable="false" name="pref_coordinputformat">coordinputformat</string>
     <string translatable="false" name="pref_gccustomdate">gccustomdate</string>
+    <string translatable="false" name="pref_customtabs_as_browser">customtabs_as_browser</string>
     <string translatable="false" name="pref_googleplayservices">googleplayservices</string>
     <string translatable="false" name="pref_lowpowermode">lowpowerlocation</string>
     <string translatable="false" name="pref_lastdetailspage">lastdetailspage</string>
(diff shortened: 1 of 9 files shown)


Lineflyer:
Just a few questions for my technical background:
- Is this - technically speaking - really a tab inside of c:geo. I always thought it just a quick way to display a webpage in Chrome.
- In some thread we discussed about excluding c:geo itself as target for intents where also c:geo itself is a possible receiver. I guess this is a separate issue but I am just curious if we can use these current changes also to get rid of a possible loop (in case webview is not active): User selects "Open in browser" but c:geo is assigned as default for the target URL --> "Open in browser" will then just open the cache in c:geo once again

fm-sys:
> Is this - technically speaking - really a tab inside of c:geo.

Technically not. Chrome is launched externally with a set of customizable settings like the action bar color or extra options in the 3-dot-menu. But isn't it kind of internally from user perspective? 

> In some thread we discussed about excluding c:geo itself as target for intents where also c:geo itself is a possible receiver.

I already tried it but without success for now 

> guess this is a separate issue but I am just curious if we can use these current changes also to get rid of a possible loop 

Currently we launch the chooser showing all possible targets explicitly, so the default target will be ignored and the chooser is always shown. Hiding c:geo from that chooser list is just a minor UI flaw...

Issue #8323, opened 2020-05-10
Separate routing strategy for individual routes


moving-bits:
Taken from https://github.com/cgeo/cgeo/issues/8304#issuecomment-625419196 by @MagpieFourtyTwo:

Although I am aware of c:geo's KISS aspiration, but nevertheless: What's about two different routing strategies - one for standard routing and one for individual routes? And, to KISS (for "standard users"), perhaps handle them as one until user decides via option to have two, to be able to set them differently? This option my e. g. enable additional menu entries in "Routing" ...?

MagpieFourtyTwo:
As for me, e. g., standard settings would usually be "Routing = Car", "Individual Route = Walk|Bike" ...

MagpieFourtyTwo:
With the new Quick Settings popup in map, control of different Routing Strategies would be really easy and pretty intuitive:
We "just" need a second row of Routing Strategy symbols below the currently shown symbols for "normal" routing, plus a grippy headline like ``Routing for individual routes`` and that's it - at least for the GUI part.  ;)

BTW: Below these "Individual Routing Strategy symbols" the line ``[_] Automatically route to individual route (if exists)`` would fit in even better.  ;)

And in the row of indiv. route symbols there would perhaps even be some space to place a [Clear] button. Perhaps at the rightmost position, to separate it a bit from the strategy buttons ...?

fm-sys:
My two cents... 
1. IMHO two different routing strategies would probably confuse "standard users" 
2. Switching the routing strategy would be more time consuming for users like me, who always use the same routing strategy for both normal and individual route.

If this feature really necessary, it would need to be optional. But as Walking and driving at the same time is quite difficult anyway 🙃 and switching the routing type is done in nearly no time (thanks to the new map popup 👍) is don't see a real advantage...

Lineflyer:
I agree to @fm-sys

MagpieFourtyTwo:
Confusion: Discussed this earlier and already came to the conclusion, that it should be optional like ``[_] Use different routing strategy for standard and individual routes`` which defaults to false.

Advantage: This of course depends on personal workflow, but at least for ours there is a *big* one, as we seldomly do DriveIns, thus *always* have two different routes: The one to the parking lot (Car) and the one to the first coordinate (Walk or Bike). And if both routes use the same strategy, you always have to switch forth and back, to e. g. find an appropriate parking lot. Which gets even more important, if you do several Multies in one tour ...

[...]

PR #9490, opened 2020-12-03
Enhancements to map quick settings popup (rel. to #8138)

moving-bits:
A step towards #8138:
- add circles, direction line, compact icon, routing to quick settings
- highlight selected buttons with colorAccent
- remove old menu entries from map overflow menu

image

(more to come)


MagpieFourtyTwo:
> highlight selected buttons with colorAccent

Looks good!  :)

MagpieFourtyTwo:
Maybe the color of the routing strategy symbols can be a bit lighter, same color as the texts of "Use compact icons"?

moving-bits:
retest this please

moving-bits:
- fixed errors preventing the build
- switch all settings to positive "show ..." instead of mix of "hide" and "show"
- recalculating a route only gets triggered on routing mode change

image

MagpieFourtyTwo:
> retest this please

The contrast of esp. the routing buttons is still too low. For better visibility of esp. the selected symbol, the symbol itself should be plain white, like the texts of the compact symbols buttons. 

While the same applies to all text in the popup, being kind of grey, which is not consistent with all the other menus.

moving-bits:
@MagpieFourtyTwo 
Ah, you're right, have to work on the text contrast yet - this task got lost in the many comments of #8138. (Easier to have a separate issue for that... Not needed in this case, though, I'm already working on it.)

As for the buttons: I took the color from one of the other buttons. I would not use plain white in this case - too bright, and also not visible in light mode. (Most of our buttons are one-colored only, for both dark and light theme.) I'll try to make it a bit lighter, taking the shade from a different button.

MagpieFourtyTwo:
Fine! Thx!

MagpieFourtyTwo:
Just checked the new symbol and text color - looks good.

Perhaps it would even look better, if the dark contour line of the symbols would be darker / more visible, to increase recognizability of the currently selected symbol (border between white symbol and orange background looks a bit blurry, and the contours are a bit hard to see with the luminous background) ... but this may even be just a matter of taste ...

Nevertheless: Working with the Quick settings popup is a true joy!  :)

moving-bits:
I played a bit with the image settings. How about those:

imageimage

Lineflyer:
> I played a bit with the image settings. How about those:

Without having compared with current status it looks good IMHO.
BTW: Did you notice, that the accent color is not applied to the checkboxes when used in Google Maps?

fm-sys:
> BTW: Did you notice, that the accent color is not applied to the checkboxes when used in Google Maps?

It's part of a larger theming problem tracked in #7587

MagpieFourtyTwo:
Already wondered why they are blue in the screenshots. No, did not notice yet. But to be honest, I do not really like GM and use it for very high zoom levels only, cause it gives a better, i. e. not that cluttered overview (admittedly, zoom level definitions are still way better than in OSM), and GSat just sometimes, to get a better impression of the terrain, so this curiosity would have been a secret to me for a really long time ... ;)

As for the symbols: They still look a bit blurry, but they can be recognized way better, IMO. So I would be happy with them.

Thx for investigating!

moving-bits:
> > BTW: Did you notice, that the accent color is not applied to the checkboxes when used in Google Maps?
> 
> It's part of a larger theming problem tracked in #7587

I found a solution for applying `colorAccent` to our Google Maps, see #9515.

MagpieFourtyTwo:
Just wanted to note that the Quick Settings popup is a real joy to work with, esp. for changing the Routing Strategy.
Although I still catch myself occasionally, still heading towards the Hamburg menu to change it ... ;)
What I find remarkable: The number of clicks has remained the same, it still needs three clicks to re-route,
nevertheless it feels so much faster, easier and way more natural. Cool.
I guess it's because of a) shorter distances (esp. for the thumb), b) there's no need to reposition the phone anymore (esp with >6") or use the other hand, and finally you c) don't have to concentrate and read text to do so. It has become a real no-brainer.

BTW: Another wish, voiced a while ago, just crossed my mind: What would you think about different Routing Strategies for "normal" and individual routes? IIRC the idea has been dropped the other day, cause it was supposed to be too hard to understand and control for the average user ... but with the Quick settings menu it's as easy as one, two, three: Just another row of Routing Strategy symbols with a different header like ``Routing for Individual Routes`` and that's it ... at least seen from the operator's view ... ;)

If you agree, I would open an FR for it ...

moving-bits:
Thanks for your feedback! Happy to see if that's an UX improvement.

Regarding the separate routing strategy for individual routes: That's #8323 - not forgotten, not closed, only not yet implemented... :-)

MagpieFourtyTwo:
Ah, you're right - thanks for the reminder!  ;)

PR #9491, opened 2020-12-04
fix #7404: switch field note encoding to UTF167LE

### LINKED ISSUE #7404: Wrong field notes encoding

eddiemuc:
fix #7404: switch field note encoding to UTF167LE


Lineflyer:
Does this change only the encoding of field notes or other file outputs as well?

bekuno:
The method `writeFileUTF16` is only used here. So it is only a change for the field notes.

PR #9492, opened 2020-12-05
allow setting target from cache/waypoint popup for GMv2 (fix #7969)

### LINKED ISSUE #7969: Harmonize maps - Add target selector to Google Maps
New OSM map features an icon on the cache/WP popup to trigger setting this cache/WP as target (showing direct/routed line from position to cache). 

This is missing on Google Maps.

moving-bits:
Setting a navigation target from cache/waypoint popup in Google Maps, similar to what we have for OSM maps.
Working mostly, but having some rough edges still.

image

### DIFF
--- a/main/src/cgeo/geocaching/maps/AbstractMap.java
+++ b/main/src/cgeo/geocaching/maps/AbstractMap.java
@@ -1,9 +1,17 @@
 package cgeo.geocaching.maps;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.enumerations.LoadFlags;
+import cgeo.geocaching.location.Geopoint;
+import cgeo.geocaching.maps.interfaces.CachesOverlayItemImpl;
 import cgeo.geocaching.maps.interfaces.MapActivityImpl;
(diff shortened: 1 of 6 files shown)


moving-bits:
As I'm not that firm with the old map code, and I had to move around quite a bit: Can someone have a look at it? Maybe you, @rsudev, as you have been doing a lot with map code in the past?

One example:

In OSM code we have `onActivityResult` checking for `requestCode == REQUEST_CODE_TARGET_INFO` and `resultCode == RESULT_CODE_SET_TARGET`. After setting the target it does more things seemingly unrelated to setting a target, but still only executed in that code block:

```
final List<String> changedGeocodes = new ArrayList<>();
String geocode = popupGeocodes.poll();
while (geocode != null) {
    changedGeocodes.add(geocode);
    geocode = popupGeocodes.poll();
}
if (caches != null) {
    caches.invalidate(changedGeocodes);
}
```

Setting the target in Google Maps code does work without those lines, but I'm probably missing something.

Thanks!

moving-bits:
@rsudev 
Do you have time for a short review on this?

PR #9495, opened 2020-12-05
rel to #9488: report db problem to user

eddiemuc:
rel to #9488: report db problem to user

Changed c:geo code such that in case of fatal db problem user gets an error message asking him to report the problem by sending logfile. Code also makes sure that sending logfile is actually doable even when db is not accessible (this meant to safeguard some of db-using functions esp dbextension)

### DIFF
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -34,6 +34,7 @@ import cgeo.geocaching.ui.WeakReferenceHandler;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.BackupUtils;
+import cgeo.geocaching.utils.DebugUtils;
 import cgeo.geocaching.utils.Formatter;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.TextUtils;
@@ -278,6 +279,12 @@ public class MainActivity extends AbstractActionBarActivity {
         // don't call the super implementation with the layout argument, as that would set the wrong theme
         super.onCreate(savedInstanceState);
 
+        //check database
+        final String errorMsg = DataStore.initAndCheck(false);
+        if (errorMsg != null) {
+            DebugUtils.askUserToReportProblem(this, "Fatal DB error: " + errorMsg);
+        }
+
         // Disable the up navigation for this activity
         final ActionBar actionBar = getSupportActionBar();
         if (actionBar != null) {
@@ -508,6 +515,9 @@ public class MainActivity extends AbstractActionBarActivity {
                 return true;
             case R.id.menu_about:
                 showAbout(null);
+                return true;
+            case R.id.menu_report_problem:
+                DebugUtils.askUserToReportProblem(this, null);
                 return true;
             case R.id.menu_helpers:
                 startActivity(new Intent(this, UsefulAppsActivity.class));
--- a/main/res/menu/main_activity_options.xml
+++ b/main/res/menu/main_activity_options.xml
@@ -54,4 +54,11 @@
         app:showAsAction="ifRoom">
     </item>
 
+    <item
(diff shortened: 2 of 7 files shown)


eddiemuc:
image

main/src/cgeo/geocaching/MainActivity.java
```diff
@@ -278,6 +278,16 @@ public void onCreate(final Bundle savedInstanceState) {
...
 
+        //check database
+        final String errorMsg = DataStore.initAndCheck(false);
+        if (errorMsg != null) {
+           Dialogs.message(this, "Fatal DB error - please report",
+               "Fatal error while trying to open/create cache database. " +
+                   "c:geo cannot be used in this state.\n\n" +
+                   "Please support us in analyzing and fixing this problem by report it via Settings -> System -> Generate logfile. " +
+                   "\n\n[" + errorMsg + "]");
+        }
```
Lineflyer, line 289: We should make this translatable, shouldn't we?

fm-sys: IMHO we shouldn't as this is only a hopefully temporary debug feature...

Lineflyer: OK then.

moving-bits:
Thanks for adding this safeguard. 
How about adding a button "Generate Logfile" to make it easier for the user to send us an (extended?) logfile in this case?

eddiemuc:
> Thanks for adding this safeguard.
> How about adding a button "Generate Logfile" to make it easier for the user to send us an (extended?) logfile in this case?

Ok I did. Now its a full-blown solution for easy reporting of problems from user to c:geo. The code change is slightly bigger of course :-)

eddiemuc:
image

eddiemuc:
After clicking on "send by email":

image

main/src/cgeo/geocaching/MainActivity.java
eddiemuc: Actually, I made strings translateable now because it is now part of a bigger "user can report problem" solution

moving-bits:
> Ok I did. Now its a full-blown solution for easy reporting of problems from user to c:geo. The code change is slightly bigger of course :-)

Wow, thanks for adding it! Though I have to admit that I find it a bit overwhelming now... I'd imagined something like this:

---

Unexpected error occurred [dialog title]

Fatal db error... [error message]

Please help us fixing this error by sending debug info to us.

[Button "Don't send"] [Button "Send by email"]

---

with "Don't send" button just closing the dialog, and "Send by email" button creating the email as shown above (like it!), attaching logcat automatically. (Maybe extended logcat?)

I know that we have many more options to send debug info to us, but the variety offered probably confuses the "average user", whereas the "pro tester" does not need the hint at all.

What do you think?

eddiemuc:
Changing the displayed texts is easy, will do it. Hopefully later today

I was more referring to the different changes necessary underneath in Dialogs, ShareUtils, DebugUtils to actually be able to offer the seemingly simple "send via email" option :-)

moving-bits:
> Changing the displayed texts is easy, will do it. Hopefully later today

Thanks

> I was more referring to the different changes necessary underneath in Dialogs, ShareUtils, DebugUtils to actually be able to offer the seemingly simple "send via email" option :-)

Yeah, I've seen that - sometimes you have to move around quite a few things to get something seemingly easy working in our context... Thanks for tirelessly pursuing the goal to constantly improve c:geo!

eddiemuc:
OK, I changed the text. I additionally added a new MainActivity option "report a problem" with same text (but without the concrete db error) which now looks like this:

image


image

What do you think?

moving-bits:
> What do you think?

Sounds useful, thanks.

moving-bits:
@eddiemuc 
I merged `release` to `master` and needed to do some conflict resolution in `DebugUtils`, `ShareUtils` and `MainActivity` - please cross-check if everything's left in good shape...

eddiemuc:
@moving-bits looks good! Thanks for handling this

PR #9503, opened 2020-12-06
gk: always use CET timezone

### LINKED ISSUE #9496: Geokrety logging doesn't work when logging from timezone ahead of CET
**To Reproduce:**
1. Open some cache (I'm in SU caches)
2. Save the log with 'write note type"
3. Use Geokrety to set all as visited
5. Send the log

Cgeo will show "logging GK", but this will be really really quick and no log will be added to GK.

I first though it doesn't work just for "write note" logs, but today it didn't for for "found" logs as well...

**Version of c:geo used:**
You will find the c:geo version in c:geo Menu -> About c:geo

Yes / often

**Screenshots:**

**System information:**
```
Attach system information here if available (see c:geo Menu -> About c:geo -> Swipe right to System)
```

**Additional context:**

okainov:
This is just a blind coding, no IDE around :(

Fixes #9496

### DIFF
--- a/main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
+++ b/main/src/cgeo/geocaching/connector/trackable/GeokretyConnector.java
@@ -27,6 +27,7 @@ import java.util.Calendar;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
+import java.util.TimeZone;
 import java.util.regex.Pattern;
 
 import io.reactivex.rxjava3.core.Observable;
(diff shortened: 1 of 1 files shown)


okainov:
Tried APK from jenkins - looks fine, working now for me

moving-bits:
Shall we target this to `release` branch?

okainov:
@moving-bits amm, I have no idea :) Just in case if needed, created https://github.com/cgeo/cgeo/pull/9506 for release

Lineflyer:
@moving-bits Indeed we should use `release`. So lets close here in favor of #9506

PR #9517, opened 2020-12-08
Implement transparency through theme selection

rsudev:
Additionally attempt to switch it off by default on Sailfish OS

Solves the c:geo start problem on Sailfish OS (part of #8752 )

### DIFF
--- a/main/src/cgeo/geocaching/utils/SystemInformation.java
+++ b/main/src/cgeo/geocaching/utils/SystemInformation.java
@@ -87,7 +87,8 @@ public final class SystemInformation {
                 .append("\nOffline maps path: ").append(Settings.getMapFileDirectory())
                 .append("\nMap render theme path: ").append(Settings.getCustomRenderThemeFilePath())
                 .append("\nLive map mode: ").append(Settings.isLiveMap())
-                .append("\nGlobal filter: ").append(Settings.getCacheType().pattern);
+                .append("\nGlobal filter: ").append(Settings.getCacheType().pattern)
+                .append("\nSailfish OS detected: ").append(EnvironmentUtils.isSailfishOs());
         appendPermissions(context, body);
         appendConnectors(body);
         if (GCConnector.getInstance().isActive()) {
--- a/main/res/values/themes.xml
+++ b/main/res/values/themes.xml
@@ -9,11 +9,6 @@
 
     <style name="cgeo_main.base" parent="@style/Theme.AppCompat">
 
-        <!-- copy the style elements of the Wallpaper theme since AppCompat has no Wallpaper theme -->
-        <item name="android:windowBackground">@android:color/transparent</item>
-        <item name="android:colorBackgroundCacheHint">@null</item>
-        <item name="android:windowShowWallpaper">true</item>
-
         <item name="colorAccent">@color/colorAccent</item>
 
         <!-- system elements -->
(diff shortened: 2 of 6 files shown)


rsudev:
I wouldn't mind a review... @moving-bits or anyone else?

Lineflyer:
How about generally setting the default to "non-transparent"? I don't mind any on support from time to time even people ask for that, not knowing that this feature is already present.

rsudev:
That is IMO an entirely different topic, perhaps something for a discussion?

rsudev:
If no one objects I would merge it today, to have it in the nightlies not too late before the next release...

main/src/cgeo/geocaching/utils/SystemInformation.java
```diff
@@ -87,7 +87,8 @@ public static String getSystemInformation(@NonNull final Context context) {
                 .append("\nOffline maps path: ").append(Settings.getMapFileDirectory())
                 .append("\nMap render theme path: ").append(Settings.getCustomRenderThemeFilePath())
                 .append("\nLive map mode: ").append(Settings.isLiveMap())
-                .append("\nGlobal filter: ").append(Settings.getCacheType().pattern);
+                .append("\nGlobal filter: ").append(Settings.getCacheType().pattern)
+                .append("\nSailfish OS detected: ").append(EnvironmentUtils.isSailfishOs());
```
fm-sys, line 91: Should we really include this line always to system information? IMHO it would be enough to show "Sailfish OS detected!" only when it is really detected. As 99% will not have Sailfish anyway...

fm-sys (approved):
Looks good IMHO. 👍 

Only one little thing to keep system information a bit cleaner...

main/src/cgeo/geocaching/utils/SystemInformation.java
fm-sys: > even though they might not be relevant most of the time (e.g. the HW acceleration info or the global filter...)

A Sensor is something each Android Smartphone may or may not have. In contrast, a real Android smartphone can _never_ be Sailfish OS. 

From the point of view of support, it is probably even better in the case of Sailfish OS to put this "warning" right at the beginning of the system information as it might be the cause of misbehavior. But if the (for nearly all users useless) line is always present, this important information that it's a non Android OS might easily get lost.

rsudev: My example was not a sensor but some other (IMO) very tangential relevant information which also is always shown, as it was felt relevant at that time. Of course is it not expected that anything besides the android runtime running under Sailfish OS is detected as such, but may be in error. And likewise could an SFOS device not be detected and the supporter might not even be aware of that an entry in the system information is to be expected (in 2 years time).
But it is such a minor issue that I won't be arguing on it any further. Feel free to amend as you see fit.

fm-sys: As I don't have write access to the repo I can't amend your PR myself. Feel free to merge the PR as it is to have it out in the nightly, all other things can also be changed later on if we decide that it makes sense...

Lineflyer: Shall we append the sailfish info in same line with Android version?

bekuno: isSailfishOs() is searching for the string aosp, which is part of the product name, what is already part of the system information.
So I think we do not need this extra information.

moving-bits (approved):

PR #9526, opened 2020-12-11
make auto-target available in map quick settings (fix #9521)

### LINKED ISSUE #9521: Make "auto-target for individual route" available in map quick settings
Reminder to myself:

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/9281#issuecomment-741282747:_

Right now there is just one wish open: As I would consider "automatically route to individual route" as a really critical routing and map option, as it may influence the routing result even more than the routing style, I still would like to see an ``[_] Always route to individual route (if exists)`` or alike in the "Quick settings" popup of the map, as mentioned in in https://github.com/cgeo/cgeo/issues/8138#issuecomment-737571005 - but this may better be discussed over there, I suppose.

moving-bits:
image

@MagpieFourtyTwo
Currently the last section ("individual route" + checkbox) is only shown if either
- an individual route is defined and/or
- "auto target individual route" is activated

Or should we always show this part?

### DIFF
--- a/main/res/layout/map_settings_dialog.xml
+++ b/main/res/layout/map_settings_dialog.xml
@@ -58,8 +58,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginLeft="8dp"
-        android:layout_marginRight="8dp"
-        android:layout_marginBottom="8dp">
+        android:layout_marginRight="8dp">
         <TextView
             android:id="@+id/routing_title"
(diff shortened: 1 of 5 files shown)


MagpieFourtyTwo:
> Or should we always show this part?

Sounds good to me. This way it's there if it's important, but OTOH does not confuse the ones who do not know/use this feature anyway.

PR #9530, opened 2020-12-12
make backup/restore buttons look like buttons (fix #9486)

### LINKED ISSUE #9486: make "Start backup" a real button
There still was the suggestion of @alan666notb to make the ``Start backup`` text more prominent ...

While I would suggest the same for ``Start restore`` and ``Restore a different backup``, too,
cause for a not knowing user it's not really obvious that theses texts are in fact buttons ...
(to be honest, I stumbled upon this, too, when I used the new dialog for the first trime, 
although I already was *really* experienced in performing backups ... ;) ).

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/9104#issuecomment-737192373_

---

I started working on it, but it is not that simple how it might look like. First I have tried to use the "normal" c:geo button style, but just adding a working button to a preference screen is already a hard challenge. Up to now I haven't got it running at all...  :-/

So if someone else wants to dive into it, it would be very welcome!

moving-bits:
image

If we want to make other texts in preferences look like buttons as well, we only need to add the `android:layout="@layout/preference_button"` in the appropriate places of `main/res/xml/preferences.xml` now.

"title" is required and used as button label, "summary" is supported additionally (to be display below the button).
The only thing that might be irritating a bit at first is that not only the button is clickable, but the whole preference line (as is the case with all preferences).

### DIFF
--- a/main/res/layout/preference_button.xml
+++ b/main/res/layout/preference_button.xml
@@ -6,10 +6,9 @@
     android:orientation="vertical"
     android:descendantFocusability="blocksDescendants">
     <TextView android:id="@android:id/title"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        style="@style/button"
+        style="@style/button_full"
         android:layout_marginLeft="18dp"
+        android:layout_marginRight="18dp"
         android:layout_marginTop="10dp"
         android:textSize="16sp"
         android:clickable="false" />
--- a/main/res/layout/preference_button.xml
+++ b/main/res/layout/preference_button.xml
@@ -1,0 +1,22 @@
+<!-- Layout used for buttons in preferences -->
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="vertical"
+    android:descendantFocusability="blocksDescendants">
+    <TextView android:id="@android:id/title"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        style="@style/button"
+        android:layout_marginLeft="18dp"
+        android:layout_marginTop="10dp"
+        android:textSize="16sp"
+        android:clickable="false" />
+    <TextView android:id="@android:id/summary"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:paddingLeft="18dp"
+        android:layout_marginTop="10dp"
+        android:layout_marginBottom="10dp" />
+</LinearLayout>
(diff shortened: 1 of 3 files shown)


fm-sys:
Wow. Great! Only thing, I would use a button which fills the full width with centered text inside. IIRC there is already a style defined which is inherited from the button style you used...

moving-bits:
> I would use a button which fills the full width with centered text inside.

Would look like this:

image

- Pro: as the full line width is clickable anyway, it's better to fill it visually
- Con: mix of left-aligned text and horizontally centered text
- Neutral: summary text is still clickable

fm-sys:
👍 IMHO it looks way more clean. I vote for "full width"

moving-bits:
> 👍 IMHO it looks way more clean. I vote for "full width"

ok, then let's merge...

MagpieFourtyTwo:
Like the full line layout, too.
First was undecided, as the "short" one already looks really nice, but looking at the full line ones I got aware of that these in fact nicely match e. g. the WP action buttons. Good job!

But: What about the other settings entries, which are in fact buttons, too, at least seen from their action?
To leave them as they are could be considered inconsistent, but OTOH changing all of them to buttons may look overloaded and disharmonious ... 

Thus I would suggest a compromise and leave most of them as they are and change just the ones which have an immediate external effect, e. g. create a file or so ... like ``Generate logfile`` and ``Create memory dump``. This way I would say it would be consistent again.

What do you think?

moving-bits:
> What do you think?

I would stay with the settings` buttons for now. Can you prepare a list of which buttons to change? I'll adapt them, then. Thanks. (Starting with the "Generate logfile" and "Create memory dump" buttons now...)

MagpieFourtyTwo:
Agreed. Will have a closer look tonight.

MagpieFourtyTwo:
Ahem, not really sure how to understand the edit of your latest posting ...

Nevertheless, went through the settings and in fact found just one more, which should be changed, too, in order to be consistent again:

1. ``Offline Data`` => ``Delete orphaned files``
2. ``System`` => ``Generate logfile``
3. ``System`` => ``Create memory dump``

With regard to the ``Backup / Restore`` section: As IMO the buttons attract enough attention now, ``Create Backup`` and ``Restore Backup`` can possibly be united to one headline, i. e. the "Create Backup" headline may be changed to "Backup" or "Backups", and the "Restore Backup" headline may be removed completely?

BTW: Just stumbled upon a minor difference between ``Main screen`` => hamburger menu => ``Backup / restore`` and the other occurrences of this title, where the ``R`` is always a capital. Maybe the ``R`` in the hamburger menu should be uppercase, too?

moving-bits:
> Ahem, not really sure how to understand the edit of your latest posting ...

I had prepared #9533 to "buttonize" the three buttons under settings - system - debug (generate logfile, create memory dump, view settings).

> Nevertheless, went through the settings and in fact found just one more, which should be changed, too, in order to be consistent again:

Thanks for cross-checking. I'll update the "delete orphaned files" button.

Issue #9197, opened 2020-10-16
Option to reset gc.com account language to any language on app exit


SiridarHarkonnen:
What about adding an option into the app (i.e. System section) where there could be the language set up (and called on app exit). I have the site switched to English all the time after using the app. So the setting could have the languages option and once set i.e. to Czech, it would call the endpoint i.e. https://www.geocaching.com/play/culture/set?model.SelectedCultureCode=cs-CZ to set the language to Czech (or any other way you are using to switch the language to English) and than it would normally exit. How about that? It would be much appreciated from me.

geocachermgo:
As a remark: to use this the users must be quite disciplined to always end cgeo normally

SiridarHarkonnen:
Yes, it will be their responsibility. But still I guess a lot of them, including me, would welcome it.

Lineflyer:
Wouldn't it be a better method to try to parse in a language independent way? 
Maybe thats possible but was never evaluated.

moving-bits:
IMHO Changing a system-wide setting like language/locale should never be done by an end-user app like c:geo. In-app selection of which language resources to use would be way better, but I don't know if that's possible without unwanted side-effects.

moving-bits:
I found a way to change the in-app language. For a POC see #9532.

Lineflyer:
I think you misinterpreted this issue. This is IMHO about changing the website language. 

I think such feature would not be meaningful or controllable. We should instead try to make parsing language independent.

moving-bits:
> I think you misinterpreted this issue. This is IMHO about changing the website language.

You mean like "geocaching.com" parsing etc.? Then we should change the issue title accordingly.
(I was thinking of c:geo itself, as we had such requests earlier.)

geocachermgo:
My understanding: cgeo switches the language of the website to english to ease the parsing.
The request of the user is to switch it back to czech (or any other chosen language) when cgeo is ended normally.

moving-bits:
Ok, I see. I'll unlink the issues and change the title of this issue.

Lineflyer:
Please let us not try to find a solution like this.
Its likely that this will generate follow-up issue, e.g.:
- What is the definition of ending the app?
- What if the user has two devices and he ends cgeo on one of them
- There are certainly some combinations of starting cgeo where switching to english might not be triggered reliably.
- etc. etc.

So I would vote closing this and rather put the effort into checking and implementing language independent parsing.

[...]

Issue #7991, opened 2019-12-02
Remove "Force to english"


Lineflyer:
Since Android 8 the force function does not work anymore (at least on some devices, see #7053) and I consider this outdated anyway.

The background for implementing this was a lack of translated strings or wrongly translated strings, mkaing it hard to use the c:geo in the language defined by the system wide setting. Since we use Crowdin for a long time now, this problem is IMHO no longer existing in a relevant extent.

So I propose to remove this function completely.

Lomanic:
I'm not a fan of this change as I prefer to have cgeo in English with Android in my native language to follow development here and properly submit bug reports.

bekuno:
I am also in favour of solving the problem and thus preserving the language selection and extending it to a general language selection.

jonas-koeritz:
Android does support multiple display languages at once now. This makes this feature obsolote for the purpose of fixing missing translations, the system will automatically find the most preferred available translation of a string for you.
This is also what might have broken this feature...

For the purpose of forcing the app to english for submitting bug reports etc. this feature is still useful and should be working. I would vote in favor of fixing it.

jonas-koeritz:
> I am also in favour of solving the problem and thus preserving the language selection and extending it to a general language selection.

IMO general language selection is not a problem app developers should have to work on as the number of users using c:geo in a language that is not part of their preferred languages list or english should be insignificant.

bekuno:
> Android does support multiple display languages at once now. This makes this feature obsolote for the purpose of fixing missing translations, the system will automatically find the most preferred available translation of a string for you.

I do not really understand what you mean. 
The Android standard is that **apps** _have to_ use the **system** language.
All other configuration is not easy to manage because there is no standard language change mechanism for **apps**. 

> This is also what might have broken this feature...

This is broken because the old used mechanism is no longer supported in the recent Android versions. We had to implement newer methods.
See https://gunhansancar.com/change-language-programmatically-in-android/

I hope this works, my tests based on other web documentation were not successful.

jonas-koeritz:
I'm talking about the multiple locale support launched with Android 7. See https://developer.android.com/guide/topics/resources/multilingual-support for reference.
Users may select more than one locale, with one being their preffered and resources getting prioritized in the selected order.
![Screenshot_20200109-205801](https://user-images.githubusercontent.com/6485387/72100750-76e7dc00-3323-11ea-9d1b-d9e9c864f115.png)

[...]

PR #9532, opened 2020-12-13
Add option to user-select one of the integrated translations instead of auto-select by system-default

moving-bits:
Adds an option to set c:geo's language independently from system language setting. Switches language on
- main screen
- settings activity
- map activities

accordingly.

Addresses #7991 ~and #9197~.

For now it's a proof-of-concept only. Language can be set by entering an ISO code (eg.: "de" for German) in "settings" - "appearance" - needs to be made more user-friendly later on.

imageimage

Precedence is:
- "use English"
- user-defined locale (if "use English" is unchecked)
- system-default locale (if "use English" is unchecked and user-defined locale is empty)

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -1561,11 +1561,16 @@ public class Settings {
 
     /**
      * Return the locale that should be used to display information to the user.
+     * Precedence is useEnglish > userLocale > defaultLocale
      *
-     * @return either the system locale or an English one, depending on the settings
+     * @return either English locale, user-defined locale or system default locale, depending on the settings
      */
     public static Locale getApplicationLocale() {
-        return Settings.useEnglish() ? Locale.ENGLISH : Locale.getDefault();
+        if (Settings.useEnglish()) {
+            return Locale.ENGLISH;
+        }
+        final String userLocale = Settings.getString(R.string.pref_userLocale, "");
+        return StringUtils.isNotBlank(userLocale) ? new Locale(userLocale, "") : Locale.getDefault();
     }
 
     public static void setRoutingMode(@NonNull final RoutingMode mode) {
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -175,7 +175,7 @@ public class Settings {
             e.putInt(getKey(R.string.pref_version), prefsV0.getInt(getKey(R.string.pref_version), 0));
             e.putBoolean(getKey(R.string.pref_ratingwanted), prefsV0.getBoolean(getKey(R.string.pref_ratingwanted), true));
             e.putBoolean(getKey(R.string.pref_friendlogswanted), prefsV0.getBoolean(getKey(R.string.pref_friendlogswanted), true));
-            e.putBoolean(getKey(R.string.pref_useenglish), prefsV0.getBoolean(getKey(R.string.pref_useenglish), false));
+            e.putBoolean(getKey(R.string.old_pref_useenglish), prefsV0.getBoolean(getKey(R.string.old_pref_useenglish), false));
             e.putBoolean(getKey(R.string.pref_usecompass), prefsV0.getInt(getKey(R.string.pref_usecompass), 1) != 0);
             e.putBoolean(getKey(R.string.pref_trackautovisit), prefsV0.getBoolean(getKey(R.string.pref_trackautovisit), false));
             e.putBoolean(getKey(R.string.pref_sigautoinsert), prefsV0.getBoolean(getKey(R.string.pref_sigautoinsert), false));
@@ -680,11 +680,11 @@ public class Settings {
     }
 
     public static boolean useEnglish() {
-        return getBoolean(R.string.pref_useenglish, false);
+        return getBoolean(R.string.old_pref_useenglish, false);
     }
 
     public static void setUseEnglish(final boolean useEnglish) {
-        putBoolean(R.string.pref_useenglish, useEnglish);
+        putBoolean(R.string.old_pref_useenglish, useEnglish);
     }
 
     public static boolean isShowAddress() {
@@ -1561,15 +1561,13 @@ public class Settings {
 
     /**
      * Return the locale that should be used to display information to the user.
-     * Precedence is useEnglish > userLocale > defaultLocale
+     * Includes migration from the old "useEnglish" preference
+     * Precedence is userLocale > useEnglish > defaultLocale
      *
-     * @return either English locale, user-defined locale or system default locale, depending on the settings
+     * @return either user-defined locale or system default locale, depending on the settings
      */
     public static Locale getApplicationLocale() {
-        if (Settings.useEnglish()) {
-            return Locale.ENGLISH;
-        }
-        final String selectedLanguage = Settings.getString(R.string.pref_selected_language, "");
+        final String selectedLanguage = Settings.getString(R.string.pref_selected_language, Settings.getBoolean(R.string.old_pref_useenglish, false) ? "en" : "");
         return StringUtils.isNotBlank(selectedLanguage) ? new Locale(selectedLanguage, "") : Locale.getDefault();
     }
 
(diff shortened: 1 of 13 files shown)


MagpieFourtyTwo:
Great! Although I currently can just test how it will work with Android set to english and cgeo set to de ... will be pretty awkward, I guess..;)

moving-bits:
> Great! Although I currently can just test how it will work with Android set to english and cgeo set to de ... will be pretty awkward, I guess..;)

You can test with every language c:geo supports, so other alternatives would be "it" for Italian, "ko" for Korean, "ru" for Russian, and many more.

MagpieFourtyTwo:
Yep, this is obvious, but most regretfully my knowledges in ru, ko or traditional Chinese are kind of limited and most probably not quite sufficient to elaborately judge cgeo's translation results ... ;)

moving-bits:
:-)

But especially with languages using non-Latin characters it's easier to see if some parts are untouched by the change of app language.

moving-bits:
Updated this PR to use a language selector instead of an edit field.
The selector gets pre-filled with the languages c:geo has translations for (taken from our build config).

imageimage

fm-sys:
Isn't the force to English setting superfluous then? As English can slso be selected via the language selector. 
Only thing we might need is a migration script for this setting - or isn't it critical enough?

moving-bits:
Migration should be easy, I'll add it and remove the "Use English" preference as well.

MagpieFourtyTwo:
How about "translating" the two-letter country codes into the "real" language names (to show these in the list)?
Would make it easier for less computer-affin people ...

moving-bits:
> How about "translating" the two-letter country codes into the "real" language names (to show these in the list)?
> Would make it easier for less computer-affin people ...

already added with the last commit a few secs ago...

MagpieFourtyTwo:
Anticipatory Fulfillment. You're getting better and better! ;)

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -175,7 +175,7 @@ private static void migrateSettings() {
             e.putInt(getKey(R.string.pref_version), prefsV0.getInt(getKey(R.string.pref_version), 0));
             e.putBoolean(getKey(R.string.pref_ratingwanted), prefsV0.getBoolean(getKey(R.string.pref_ratingwanted), true));
             e.putBoolean(getKey(R.string.pref_friendlogswanted), prefsV0.getBoolean(getKey(R.string.pref_friendlogswanted), true));
-            e.putBoolean(getKey(R.string.pref_useenglish), prefsV0.getBoolean(getKey(R.string.pref_useenglish), false));
+            e.putBoolean(getKey(R.string.old_pref_useenglish), prefsV0.getBoolean(getKey(R.string.old_pref_useenglish), false));
```
fm-sys, line 178: As far as I have understood the code, this will not work as it will only be run if the settingsVersion is lower than 1. But on the other hand this feature isn't critical enough that it needs its own new settingsVersion...

moving-bits: This is some very old migration code, not directly related to the current PR.

The selected language is stored in a new string preference called "selectedLanguage" (whereas the old "useenglish" was a boolean preference). "Migration" is done in `Settings.java`:

```
public static String getUserLanguage() {
    return getString(R.string.pref_selected_language, Settings.getBoolean(R.string.old_pref_useenglish, false) ? "en" : "");
}
```

fm-sys: I always look at the code of others to learn something new, and this part had confused me a bit.Thanks for your explanation, now I got it and you're of course totally right!

moving-bits: One addition: The change in the lines above came from the fact that I renamed the string which holds the actual preference key "useenglish" from "pref_useenglish" to "old_pref_useenglish" to mark it as outdated.

MagpieFourtyTwo:
Well done.  BTW: Nice and *very* welcome side effect (I got ware of on first sight of the list):
One can now see at the tip of his finger, *how* many languages c:geo is used in.
Impressive.

MagpieFourtyTwo:
Current situation: Android set to "de", c:geo set to "en", which seemed to work pretty well over the last two days.

While caching this morning (thus have opened c:geo now and then to navigate to the [P], walk to the coords, look for position and stuff) I finally found and logged the cache, saved the log offline and happily went away ... while walking back to [P] s/th came to my mind which should be mentioned in the log, so I switched on and tried to reopen the log for edit.  But on short click on the text I just got the popup you get when clicking somebody else's log, which has of course no entry for "edit". Strange. Only way to edit it was, to again click on ``Log Visit``.

Then I recognized that the log was titled ``Dein gespeichertes Log`` - which did not really match my language setting, which was still set to ``en``, as I cross checked right after. After resetting it to ``de`` and afterwards to ``en`` again, the log was properly titled with ``Your saved log`` and even the edit worked again as usual.

Conclusion: There seem to be some scenarios, where the language switch does not properly resp. not *fully* work, as my log title was in Android's language (de), while e. g. the ``Details`` page and c:geo's main screen where still in c:geo's language (en).

Wild guess: The decision, if after tap on a log either the menu popup appears or edit will start, depends on a string compare of the log title ... and of course fails in cases of Babylonian language confusion?  ;)

moving-bits:
There may / will be more places where language change may not work for now.

Technical background:

It looks like you cannot change the language setting for the whole app any more in more recent Android versions, like we did in the past - that's why the "use English" settings does not work any more since quite some time.

What does work is configuring a different language for each so-called `Activity` (think of it as "screen/scope in c:geo"). Many of c:geo's activities are based on `AbstractActivity`, so with updating that I caught most of our screens / activities. I'd also included our two map activities and our `Settings` activity, as those are not based on `AbstractActivity`, but there will be more, and it looks like you've identified one.

Let's collect those findings here in this issue, so that we can add the language configuration to those activities as well.

MagpieFourtyTwo:
> There may / will be more places where language change may not work for now.

Hmm, probably I haven't fully understood, but my scenario didn't seem to be a "where" but more of a "when" ...
Cause usually it works in ``Logbook`` ... or does it make a difference if c:geo's opened from scratch vs. just wake up a sleeping instance?

MagpieFourtyTwo:
Right now it happened again: Took the phone, which was sleeping for about a quarter of an hour, switched it on and refreshed the listing of the cache whose log I just finished online on gc.com. Then swiped to ``Logbook`` to delete the saved offline log ... but there was the ``Dein gespeichertes Log`` again ... and just the popup menu instead of the possibility to edit ... then swiped around a bit ...

On page ``Logbook`` and ``Friends/Own Logs`` all texts are in German, apart from the bottom line with the page names. And all Menu entries are still in English. Page ``Description`` is English, while ``Details`` is a mixture: The topic titles in the upper half are in English, date format is obviously German. ``Auf dem Gerät gespeichert vor ein paar Minuten`` and the textual description of the attributes are in German, the rest below English again. ``Waypoint`` page is completely in English.

Then I closed c:geo, removed it from memory and started it from scratch - everything in English again. ``Your saved log`` is there again, and everything is as it used to be.

This time I will intentionally leave c:geo open and the ``Logbook`` active, put the phone on the desk and leave it there for a while, untouched ... let's see what will happen on next awakening ... ;)

MagpieFourtyTwo:
Okay ... even after an hour of sleep mode, c:geo is still completely in English.

Unexpected, but in fact matches a new insight, I just got: Quite the opposite of my initial assumption, the Babylonian effect must have happened even *before* I started to log the cache offline, i. e. in a scenario with normal use of c:geo, cause the log template has substituted the month's name with the German ``Dezember``.

eddiemuc:
@MagpieFourtyTwo 

> Wild guess: The decision, if after tap on a log either the menu popup appears or edit will start, depends on a string compare of the log title ... and of course fails in cases of Babylonian language confusion? ;)

Your guess is right unfortunately. I remember when I worked in this area and found it. I already thought that this may create funny effects in the future but forgot at the time to refactor it.

eddiemuc:
And a wild guess regarding pages with "mixed" translations: some translations are prefilled on page creation (from layout), but many are also filled programmatically. My guess would be that some of the programmatic cases are not using translation in a way that they pick up the cgeo language setting.

fm-sys:
Should we create an issue for this? Otherwise it might gets lost in this merged PR...

moving-bits:
> Should we create an issue for this? Otherwise it might gets lost in this merged PR...

I've created a separate issue and copied the comments related to it.

PR #9533, opened 2020-12-13
buttonize settings-systems debug buttons

moving-bits:
image


fm-sys:
Can these lines between the buttons also be avoided? As they look kind of weird at a first glance...

moving-bits:
> Can these lines between the buttons also be avoided? As they look kind of weird at a first glance...

They are built-in - don't know how to suppress them.

PR #9539, opened 2020-12-14
restart c:geo after language change automatically when leaving settings

fm-sys:
I first thought about restarting c:geo directly after changing the language setting, but then I decided against it. With this implementation c:geo will be restarted automatically when leaving the settings activity. Therefore, it's easier to change it back if you accidentally select the wrong one...

...and the same is done for the "bigSmileys" option too.

### DIFF
--- a/main/src/cgeo/geocaching/settings/SettingsActivity.java
+++ b/main/src/cgeo/geocaching/settings/SettingsActivity.java
@@ -861,6 +861,10 @@ public class SettingsActivity extends PreferenceActivity implements Preference.O
             final ListPreference listPreference = (ListPreference) preference;
             final int index = listPreference.findIndexOfValue(stringValue);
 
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/settings/SettingsActivity.java
```diff
@@ -861,6 +861,10 @@ public boolean onPreferenceChange(final Preference preference, final Object valu
             final ListPreference listPreference = (ListPreference) preference;
             final int index = listPreference.findIndexOfValue(stringValue);
 
+            if (isPreference(preference, R.string.pref_selected_language)) {
```
moving-bits, line 872: good catch, thanks!
Those lines had been there some time along the way, but got lost before I finalized the PR, due to switching between different implementation variants...

PR #9542, opened 2020-12-15
parse DNF

### LINKED ISSUE #3206: Parse DNF
With the new cache details page we would have the possibility to parse, that a user posted a DNF log for a certain cache.
This could be shown as a status in cache details ("Not found") and additionally be shown by a blue(?) smiley on the map.

See also #3129

fm-sys:
fixes #3206 - shows a blue emoji for caches which are logged as DNF


Lineflyer:
Nice.
Question: Shouldn't we also include parsing the DNF for caches which are parsed on live map? Or is it already included in this PR?

fm-sys:
> 
> 
> Nice.
> Question: Shouldn't we also include parsing the DNF for caches which are parsed on live map? Or is it already included in this PR?

Jup, the second one. Admittedly, I was a little lazy when writing the PR description. DNF will be parsed on live map as well as when downloading a cache completely...

Lomanic:
Thanks for taking the time to fix this long standing issue. While I don't use cgeo anymore nor geocache these days (for reasons you might guess), this is a real improvement and I hope to be able to use this feature soon when I will get back to the game.

PR #9547, opened 2020-12-16
hide all other menu icons when search is active

### LINKED ISSUE #9545: [Nightly] Layout search window too small at first
[minor]
**Describe the bug:**
After the first start of the new nightly that layout was inappropiate

There were less then 2 letters after GC visible in the search window. The list of the last search results was very narrow.

**Version of c:geo used:**
2020.12.16-NB-7e7fd92

Yes/No (was first like on the screenshot but switched back to normal, during writing of this report)

**Screenshots:**
Screenshot_2020-12-16-07-05-51-255_cgeo geocaching

fm-sys:
fix #9545

grafikgrafik


moving-bits:
Tests are failing due to GC3FJ5F being disabled, therefore let's merge...

PR #9559, opened 2020-12-20
UNKNOWN_CONNECTOR not used for finding geocode in description

### LINKED ISSUE #9336: Import gpx with uppercase in <desc>
**Describe the bug:**
Importing from .gpx file with 5 points imports only the last point.

**To Reproduce:**
1. Select list
2. Import from gpx file

If I import from file LAB_KAREL_GOTT.big.gpx, I can see only one point. If I change first word in tag desc from uppercase to camelcase, import works as expected - import 5 points.

I attach 2 files (renamed from .gpx to .txt to enable upload to github):
* LAB_KAREL_GOTT.big.gpx.txt - with uppercase in tag desc - 
[LAB_KAREL_GOTT.big.gpx.txt](https://github.com/cgeo/cgeo/files/5471865/LAB_KAREL_GOTT.big.gpx.txt)
* LAB_KAREL_GOTT.camel.gpx.txt - with camelcase in tag desc - 
[LAB_KAREL_GOTT.camel.gpx.txt](https://github.com/cgeo/cgeo/files/5471866/LAB_KAREL_GOTT.camel.gpx.txt)

As you can see, there is difference only this tag desc.

In gpx you can see that name of cache is set to GC8XYT7-X, but c:geo imports it with GC code "KAREL" (of course in upper case variant, in camelcase works as expected). Screenshot_20201101-230429_cgeo

It looks like a problem with xml parsing.

**Version of c:geo used:**
2020.10.29, but this problem is older.

Yes

murggel:
This PR fixes #9336, where for a labcache the geocode is wrongly found in the description. 

For finding a valid geocode the name, the description and the comment is used. 
This PR allows the unknown-connector (which allows any uppercase 5-letter-word as geocode) only for the name, but not for description and comment.

### DIFF
--- a/main/src/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/cgeo/geocaching/files/GPXParser.java
@@ -71,6 +71,7 @@ abstract class GPXParser extends FileParser {
     private static final Pattern PATTERN_GEOCODE = Pattern.compile("[0-9A-Z]{5,}");
     private static final Pattern PATTERN_GUID = Pattern.compile(".*" + Pattern.quote("guid=") + "([0-9a-z\\-]+)", Pattern.CASE_INSENSITIVE);
     private static final Pattern PATTERN_URL_GEOCODE = Pattern.compile(".*" + Pattern.quote("wp=") + "([A-Z][0-9A-Z]+)", Pattern.CASE_INSENSITIVE);
+
     /**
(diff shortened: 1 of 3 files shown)


moving-bits:
LGTM. Thanks @murggel for adding this fix!

PR #9564, opened 2020-12-21
a days work

eddiemuc:
related to #8457: a days work. Offline Map folder seems to work now in completeness


moving-bits:
@eddiemuc 
I'm really looking forward to seeing c:geo becoming SAF-capable - thanks for your hard work!

PR #9572, opened 2020-12-22
Update the Locus integration

bekuno:
- Update the Locus integration to the 0.9.x series (refactoring Locus)
- Update the CacheType's
- Update the deprecated storage access for sending huge data to Locus (>1000 Caches)
  The old implementation should not work with Android 10 and 11 (not tested)
  (see https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory() )
  Migrate to `context.getExternalFilesDir()`, therefore changed the data file location:
    old path:
    `/storage/emulated/0/Android/data/menion.android.locus.api/files/showIn.locus`
    new path:
    `/storage/emulated/0/Android/data/cgeo.geocaching/files/showIn.locus`

@Lineflyer: Please test using devices with Android 10 and 11
@eddiemuc: Please check if this fits in the SAF implementation

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -301,8 +301,7 @@ dependencies {
     implementation "com.squareup.leakcanary:plumber-android:$leakCanaryVersion"
 
     // Locus Maps integration
-    //noinspection GradleDependency
-    implementation "com.asamm:locus-api-android:0.3.17"
+    implementation "com.asamm:locus-api-android:0.9.21"
 
     // Mapsforge new version
     def mapsforgeVersion = '0.14.0'
--- a/main/src/cgeo/geocaching/apps/AbstractLocusApp.java
+++ b/main/src/cgeo/geocaching/apps/AbstractLocusApp.java
@@ -14,7 +14,6 @@ import android.Manifest;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.net.Uri;
-import android.os.Environment;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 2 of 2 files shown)


Lineflyer:
shall we merge this before branching the new beta of from `master` (include it in beta) or only after?
I prefer after merging as it allows more testing and I am already about to merge branches tonight or tomorrow.

bekuno:
This PR can wait.

eddiemuc:
@bekuno I was not aware that we have a Locus integration. What does it do?
SAF is a different approach to sharing files without usage of external storage.

bekuno:
You can at an single cache navigate with Locus. You can from a list of caches "show on map" - Locus or "Export to Locus".
Until 1000 caches they are sent each after, more caches as block using a file.

eddiemuc:
Ah thank you.
I assume that in this case the file is just created temporary for transfer and can be discharted later (no user or app wants to access this file after transfer is completed).

It might be that for this case you could even use the app-internal file storage (which is unaffected by SAF) since Locus gets implicit temporary read grant to the file anyway. But I haven't tried that out.

moving-bits:
@bekuno @eddiemuc 
I'm unclear about the status of this PR - anything still to be done, or is it ready to merge?

eddiemuc:
Regarding the changed file dir: as long as it works, it should be fine. In this case files are only created temporary, so I guess public access to them by user is not needed.

Regarding the other parts of the PR I can't tell.

bekuno:
It would be fine to have test results for Android 10 and 11 devices.
But the tests can also be done after merging.

moving-bits:
> It would be fine to have test results for Android 10 and 11 devices.
> But the tests can also be done after merging.

You mean merging for the next release (has been done already) or merging of the SAF branch?

bekuno:
This PR should be integrated in master.
It will integrazed into the SAF branch when @eddiemuc rebase them. But there is probably no need for a change.

bekuno:
Lets test it in the nightlies.

PR #9581, opened 2020-12-23
unify design of different export functionalities

fm-sys:
I'm on the go to unify the design of all different export functionalities, to have a consistent UX...

GPX export for example has IMHO two main problems:
- "Open share menu after GPX export" may not be clear for everyone, and you have to decide in advance on something that will only be relevant later on
- The toast "GPX exported to: ..." is way too quick, so it has happened to me several times that I had to repeat the export, because I had too little time to remember the file name

Therefore, it makes sense to use the same UI as we already have for backup and logfiles:

grafik

To avoid additional clicks, can we possibly avoid the following pop-up completely. Is there a real reason why you should not include the found status? We can either move the function to the settings or remove it completely. I am in favor of variant two, because in my opinion c:geo already has far too many settings options and tries to please everyone. (PS: [That](https://github.com/cgeo/cgeo/pull/5835) was the PR where it was added)

grafik

### DIFF
--- a/main/res/layout/gpx_export_fragment.xml
+++ b/main/res/layout/gpx_export_fragment.xml
@@ -17,9 +17,4 @@
         android:paddingBottom="20dp"
         android:paddingLeft="8dp"/>
 
-    <CheckBox
-        android:id="@+id/share"
-        style="@style/checkbox_full"
-        android:text="@string/init_share_after_export" />
-
 </LinearLayout>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -776,6 +776,7 @@
     <string name="init_create_memory_dump">Create memory dump</string>
     <string name="init_memory_dump">Memory dump</string>
     <string name="init_memory_dumped">Memory dumped to %s</string>
+    <string name="init_please_wait">Please wait…</string>
     <string name="init_hardware_acceleration_title">Hardware accelerated rendering</string>
     <string name="init_hardware_acceleration_note">Hardware acceleration renders graphical elements faster on the screen. However on some devices the Android operating system contains bugs and some text may appear blurred (notably bold characters). Disable hardware acceleration if this happens to you.</string>
     <string name="init_hardware_acceleration">Enable hardware acceleration</string>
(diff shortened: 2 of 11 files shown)


eddiemuc:
I opt for variant two too. I agree with @fm-sys : c:geo already has far too many settings

Lineflyer:
IMHO the exclusion of found state is a valid usecase as described in the referenced PR.

fm-sys:
> IMHO the exclusion of found state is a valid usecase as described in the referenced PR.

If we want to keep it, we shouldn't move it to the settings but keep it inside the dialog. However, means one click each time more until export starts...

fm-sys:
If we want to stay with the "Include found status" this PR should be ready.

This was my last GitHub commit for this year. Over Christmas, I will take some time off. 
I wish you all Merry Christmas and a happy and hopefully positive new year! Stay healthy!

Lineflyer:
> This was my last GitHub commit for this year. Over Christmas, I will take some time off.
> I wish you all Merry Christmas and a happy and hopefully positive new year! Stay healthy!

Thanks for your great contribution to c:geo and a very Merry Christmas to you!!!

moving-bits:
> I'm on the go to unify the design of all different export functionalities, to have a consistent UX...

Sounds good, thanks for looking into this!

> * "Open share menu after GPX export" as it may not be clear for everyone, and you have to decide in advance on something that will only be relevant later on
> 
> * The toast "GPX exported to: ..." is way too quick, so it has happened to me several times that I had to repeat the export, because I had too little time to remember the file name

Agreed with both.

> To avoid additional clicks, can we possibly avoid the following pop-up completely. Is there a real reason why you should not include the found status?  We can either move the function to the settings or remove it completely.

I would keep that option. Some might want to leave it "found" status for privacy reasons or for trying not to confuse the recipient's data.

fm-sys:
> I would keep that option. Some might want to leave it "found" status for privacy reasons or for trying not to confuse the recipient's data.

That's how it is done in this PR. The option is still there...

moving-bits:
> This was my last GitHub commit for this year. Over Christmas, I will take some time off.

Thanks for all your contributions throughout the year! 

Together with all the active contributors (code & testing etc.), 2020 has been a very productive year for c:geo!
- About 850 commits (translations not counted), with
- many bug fixes, 
- many new features, 
- and a few big steps forward to a technically more up-to-date c:geo!

For all of you: Have a nice holiday season, find some rest, and stay healthy!

Issue #9563, opened 2020-12-21
Splash screen flashing in background


Lineflyer:
Since some days I do have strange effects on startup of c:geo. Sadly there has also been a major version update (Andriod 10 - Android 11) also last days, so I am not sure whether the upgrade caused the problem or it was introduced by recent changes on `master`.

- 2020.12.20-NB
- Have setting active "Transparent background"
- Start c:geo

Two possible scenarios:
- Start up takes some time and the splash screen is shown normal, afterwards normal main screen with transparent background
- Start up is fast, but in that case the transparent background is replaced by an overdimensional c:geo logo for a second, afterwards normal main screen with transparent background

A video might explain it better than I can with words:
https://files.droescher.eu/public/PXL_20201221_164045435.mp4

@rsudev Any relation possible to the changes for Sailfish OS?

-

moving-bits:
Never seen that effect so far, neither in the emulator nor in my physical testing device (which is running Android 7, IIRC).

Maybe some splash screen transition effect? (Could be related to #9392, then.)
I've no transition actively set for the splash screen, so it may be some default depending on vendor and OS version.

Lineflyer:
@moving-bits #9584 provides no fix (looks exactly same as initially described in this issue).
Please decide if we shall keep it in anyway or revert?
@SammysHP Can you test regarding #9392 with newest nightly starting 2020.12.26-NB1?

moving-bits:
> @moving-bits #9584 provides no fix (looks exactly same as initially described in this issue).

This has been merged to `release` branch - did you test it from there (or merged `release` to `master` before testing)?

> Please decide if we shall keep it in anyway or revert?

We can keep it in there, it does not harm.

There's another thing I can try, depending on your reply. (Setting `<item name="android:windowActivityTransitions">false</item>` in the theme for the `SplashActivity`.)

Lineflyer:
Tested on nightly but change has of course been merged upstream before. So I should have tested correctly.

moving-bits:
> Tested on nightly but change has of course been merged upstream before. So I should have tested correctly.

Ok, thanks. I'll prepare a PR with the config info shown above - let's see if that helps.

Lineflyer:
No change with recent nightlies.
I do not consider it anyway important...just a glitch. Do not put to much effort in this and lets not try anything which might have negative effects on other devices. 
Maybe some day we will find it out or it gets implicitly solved once we modernize the frontend.

[...]

Issue #9392, opened 2020-11-15
Set activity transition for splash activity


SammysHP:
Currently (at least on my device) the splash activity exits with a sliding transition which looks weird. I'd prefer a simple fading transition or no transition at all.

MagpieFourtyTwo:
Nokia 3210? ;)

Over here I can be glad to get to see it at all. Most of the times it vanishes *so* fast that one could have the impression @Lineflyer described in #9386.

But, to be honest, I'm not really sad about it, as the quicker c:geo is "ready to serve", the better.

SammysHP:
I see it for about 0.5 seconds. So it appears and almost immediately slides down. Also I noticed that it is skipped if using old links in the launcher (which are associated with the main activity).

moving-bits:
> Currently (at least on my device) the splash activity exits with a sliding transition which looks weird. I'd prefer a simple fading transition or no transition at all.

Strange. Haven't set any transition effect manually in the code, and do not see that effect in the emulator (or on my physical mobile). Maybe device-, vendor- or Android version-specific?

Lineflyer:
@SammysHP 
Do recent nighlies still show this transition effect?
@moving-bits changed the transition while trying to debug #9563

SammysHP:
Seems to be fixed, the transition is gone. It just appears and disappears (although fading would look nicer).

The logo looks fuzzy and low resolution, but that's a different issue.

PR #9584, opened 2020-12-26
disable transitions for splashscreen

moving-bits:
c:geo did not have set any transitions for `SplashActivity` calling `MainActivity`, but there seem to some system-specific default transitions. This PR sets the activity options to null, which should disable any transition.

@Lineflyer @SammysHP 
Regarding #9563 and #9392: As I can reproduce neither of it in my test environment - can you cross-check with this PR on the devices where those issues occurred? Thanks.


Lineflyer:
I will merge and test in nightlies

Issue #9576, opened 2020-12-23
Wrong color accent for quick setting checkboxes on GMaps


Lineflyer:
Buttons are orange, checkboxes however are blue.

![Screenshot_20201223-122722_cgeo](https://user-images.githubusercontent.com/949669/102991915-818e8780-451a-11eb-99f1-46d82bc1c343.jpg)

Related to #9373 and #8783

MagpieFourtyTwo:
This is pretty strange - in even two ways:
1. the checkboxes over here look completely different, i. e. orange and not round but square, thus perfectly fine
2. we already had a similar issue, which has already been fixed by @moving-bits, IIRC, so are you perhaps using an old version?

MagpieFourtyTwo:
Ah, I'm currently using 2020.12.26-NB-a2423d5

Lineflyer:
I think it could be more related to your Android version...or you did not use GMaps.

MagpieFourtyTwo:
The boxes look the same, no matter if I use OSM or GMap or GSat.
Android is 10.1.0.171, Huawei.

moving-bits:
I can only guess here, as the dialog already activates a `ContextThemeWrapper` which sets the theme to our `DialogAlert` theme, which sets the `colorAccent` to the desired value.

But I've stumbled already earlier about the fact, that some themes work with `colorAccent` where others use `android:colorAccent` (or, to be more precise: checkboxes in some themes use it), so I'll prepare a test PR which sets `android:colorAccent` to `DialogAlert` theme - let's see if that helps.

Lineflyer:
@moving-bits  #9585 did not change anything related to this issue for me.
Please note, that besides the mentioned quick settings, e.g. also the map selection on GMaps has blue accents (and black background) on my device. I guess the complete styling needs to be equalized to OSM-Maps (see #9373) as some parents are siginificantly different.

Lineflyer:
This is how GMaps in c:geo looks for me (similar to settings).
Android 11, Samsung S20, 2020.12.26-NB1

![Screenshot_20201226-221915_cgeo](https://user-images.githubusercontent.com/949669/103159329-8710f980-47c8-11eb-8090-cb59d8da265d.jpg)
![Screenshot_20201226-221909_cgeo](https://user-images.githubusercontent.com/949669/103159330-88422680-47c8-11eb-98a2-3381b122de98.jpg)
[...]

PR #9585, opened 2020-12-26
use android:colorAccent in DialogAlert themes

moving-bits:
see https://github.com/cgeo/cgeo/issues/9576#issuecomment-751334802


Lineflyer:
I will merge and check results in nightlies.

PR #9597, opened 2020-12-27
Add OneTimeDialog for map quick settings (fix #9593)

### LINKED ISSUE #9593: Users missing map settings
Several mails on support mail from beta users 2020.12.22-RC:

- All the map settings are gone in the beta version?
- Why do I need to change hiding own caches now from the settings each time. It was better in last version where it was possible from the map menu!
- I can no longer change the routing between walk, drive. The menu is no longer where it was before.

Maybe it would be good to add a dismiss-able message box (didn't we introduce such generic function recently) when entering the map to explain, that all map customization has been moved to the quick settings and where they can be found?

moving-bits:
image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1285,6 +1285,7 @@
     <string name="map_gm_autorotation">Map auto rotation</string>
     <string name="map_gm_autorotation_disable">Disable map auto rotation?\n\nYou can enable it again using the Map rotation menu</string>
     <string name="quick_settings">Quick settings</string>
+    <string name="quick_settings_info">Some map settings (Map settings, hide, routing) have been moved to the new \"Quick settings\" popup.\nYou can open this popup using the icon in the map\'s bottom center.</string>
     <string name="map_show_in_map">Show in map</string>
 
     <!-- search -->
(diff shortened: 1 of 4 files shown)


fm-sys:
We might think about only showing the popup to users, who upgraded to the new c:geo version, but not for newcomers. 

@moving-bits what do you think? BTW, such differentiation has not yet been designed for one time messages so a small architecture adaption will be necessary...

fm-sys:
> We might think about only showing the popup to users, who upgraded to the new c:geo version, but not for newcomers.

I can/would like to create a PR for it (end of this week)

Lineflyer:
> I can/would like to create a PR for it (end of this week)

Sure, appreciated.

moving-bits:
For reference: @fm-sys 

Currently detection of "new user" works by checking `Settings.getLastChangelogChecksum() == 0` in `SplashActivity`, but this value is overwritten by `Settings.setLastChangelogChecksum(checksum)` in `MainActivity:checkChangedInstall` and thus no longer available in the map activities.

We need either find a different way to recognize a new installation (maybe some other settings not yet set) or forward this info to the map activities after a fresh install.

fm-sys:
I was thinking of a more generalized solution. For example detecting a fresh install and initialize all OneTimeMessages depending on their default show state (always, freshInstall, upgrade, never) which could be added as a third enum parameter...

moving-bits:
Sounds promising!

PR #9598, opened 2020-12-27
show toast only if individual route or track loaded (fix #9594)

### LINKED ISSUE #9594: Toast for route recalculation although no route
2020.12.22-RC (but also present on 2020.12.02 release):

- Fresh install and configure c:geo
- Open live map
- Enter quick settings
- Change routing to another mode (e.g. from walking to cycling)
- Close quick settings

Result:
Toast "Recalculating route in background" although there has never been set any destination or individual route

Expected:
Only show such toast if routing is in use

moving-bits:
Show "recalculating route" on map quick settings close only, if routing mode has changed and either individual route is configured and/or GPX track/route has been loaded.

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -846,7 +846,9 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
 
     private void routingModeChanged(final RoutingMode newValue) {
         Settings.setRoutingMode(newValue);
-        Toast.makeText(activity, R.string.brouter_recalculating, Toast.LENGTH_SHORT).show();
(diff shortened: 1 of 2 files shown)


Lineflyer:
> /or GPX track/route has been loaded.

Isn't a GPX track/route always containing the routing information in itself..meaning do we apply Brouter on it?

moving-bits:
> > /or GPX track/route has been loaded.
> 
> Isn't a GPX track/route always containing the routing information in itself..meaning do we apply Brouter on it?

Technically a track is supposed to contain detailed routing info, whereas a route contains separate waypoints for which a detailed route still has to be calculated (therefore BRouter to be applied to). They can be distinguished within the GPX file due to the tags used.

(IIRC I apply BRouter to track points as well, if they are farther away from each other than some specified distance - would need to cross-check, though.)

PR #9600, opened 2020-12-27
configure splash activity theme to not use activity transitions

moving-bits:
See #9563


moving-bits:
@Lineflyer 
I've merged this to `release` and `master` - can you cross-check if it does work now? Thanks.

PR #9601, opened 2020-12-27
make accuracy circle line configurable

moving-bits:
For consistency (and for being wished for by @MagpieFourtyTwo, IIRC) add color config for accuracy circles as well.

### DIFF
--- a/main/res/values/colors.xml
+++ b/main/res/values/colors.xml
@@ -35,5 +35,7 @@
     <color name="default_directioncolor">#80FF3333</color>
     <color name="default_circlecolor">#44BB0000</color>
     <color name="default_circlefillcolor">#66BB0000</color>
+    <color name="default_accuracycirclecolor">#66000000</color>
+    <color name="default_accuracycirclefillcolor">#08000000</color>
 
 </resources>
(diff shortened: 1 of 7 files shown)


MagpieFourtyTwo:
Great! Thy!

Nevertheless, seems as if it's not working properly yet, as changing the parameters for ``Accuracy circle fill color`` does not change anything at all (over here), while ``Accuracy circle line color`` primarily changes the circle fill color and opaqueness.

MagpieFourtyTwo:
Moreover: The same values for the settings of ``Accuracy circle line`` and ``Circle fill color`` (which in fact both do change the circle fill color) lead to different results. Is this expected?

moving-bits:
Hm, then it has never worked before - I simply replaced the hard-coded color constants with the values from preferences. Looks like I need to test this on a physical device, as in my emulated device I'll never get any accuracy circles... I'll try to look into it throughout the next days.

MagpieFourtyTwo:
Thx! But, no stress, as it's already usable right now, one just has to know.  ;)
And, yes, now that it can be made more visible, the proximity circle starts to make sense again.

MagpieFourtyTwo:
BTW: In Lockito you can set accuracy. And when setting up a route in Lockito, you will even get a proximity circle.

moving-bits:
Aaarg - copy & paste type while working on retrieving the colors in `MapLineUtils` - have prepared a fix for it...

MagpieFourtyTwo:
Thx!

PR #9604, opened 2020-12-27
Update downloaded maps (fix #9468)

### LINKED ISSUE #9468: Add option to update downloaded OpenStreetMaps
It would be great when the app has an option to update downloaded OpenStreetMaps and tells when a newer version of the map is avilable.

moving-bits:
Update checker for offline map files downloaded with the integrated MapDownloader

- store source URL and date info for downloaded map files
- on opening the "download map files" dialog check for
  - are there any files stored in `InstalledOfflineMaps`?
  - if yes, do those files still exist in local filesystem?
- if yes:
  - show "Check for map updates" button
  - on clicking: check for each of the above files, if there are updates available
  - if yes: offer them for download
  - if the user select one of those for download, the regular download process starts

New "Check for updates" button above maps list:
image

On pressing that button, c:geo checks the downloaded offline map files. Result can be either "No updates available" or a list of available updates:
imageimage

Also, this PR fixes some string resources of the map downloader (hard-coded strings now changed into a translatable resource).


moving-bits:
@eddiemuc 

The local map file location is stored as URI - will that be compatible with your upcoming SAF changes?

`InstalledOfflineMaps` probably needs to be respected during migration (see `localURI` => `string2`) .

For presenting the locally installed offline map files (those selected one by one by the user using the SAF dialogs) - do you need some place to store that info? If so, you could use `InstalledOfflineMaps` for this as well (maybe marking it with some flag for the upgrade checker to ignore those entries).

eddiemuc:
> @eddiemuc
> 
> The local map file location is stored as URI - will that be compatible with your upcoming SAF changes?
> 
> `InstalledOfflineMaps` probably needs to be respected during migration (see `localURI` => `string2`) .
> 
> For presenting the locally installed offline map files (those selected one by one by the user using the SAF dialogs) - do you need some place to store that info? If so, you could use `InstalledOfflineMaps` for this as well (maybe marking it with some flag for the upgrade checker to ignore those entries).

Hi @moving-bits , I would beg you not to store any absolute Uris to map sources anywhere since I finally got rid of them...

Simplifled speaking, with SAF there will be one absolute Uri stored for the Offline Maps to the path where it currently points to. This path can change anytime (whenever the user selects a new path in Settings). On map opening, this dir will be scanned (there is caching involved, don't worry) and the maps found in the dir can be used. The only other place where something is stored is the last displayed map source; here I only store the file name, not it's path.

Now what to do with your use case. I guess this depends on what should happen to already downloaded map files when the user changes the offline map dir (and thus does not see these maps any more). Should they still be watched for downloads? I would assume not. So maybe it is enough also for your use case to store map file names?

Note that with SAF you will be able to register to a notifier which informs you whenever the user changes a directory. This might be useful as well: in case you want to stop watching for updates of maps once they are no longer in the users dir, this might be the place to deregister watching.

Note also that I plan to offer a "copy"-function to the user whenever he changes its dir (he can then ioptionally copy or move the content of the old dir to the new one - I figured I can offer that since I have to write the very complicated copyAll function anyway for migration, and that way it will definitely be tested better).

What do you think?

eddiemuc:
@moving-bits come to think of it: it might imho make sense to store offline-map-related information not in the DB but in a file next to the actual offline map file. That way you don't have to keep track of the map file location at all. If you see the map file in the current offline map dir, you also see the info c:geo stored within it. File could be named e.g. "(mapfilename)-cgeo-info.txt" and can be a simple property file or something more complex if needed.

I think currently about also storing the log image files no longer as absolute Uris but only the filename inside the Image-dir. I think this makes things much easier. Combined with a copy/move-function on dir change this should do the trick imho.

moving-bits:
HI @eddiemuc,

thanks for the fast reply. Hm, I thought that having absolute Uris would make things easier for both SAF migration and the file access after that, as I have a full handle to the file involved.

From what I have read so far I'm still unsure what will be possible when using SAF and what won't. At one point in time I thought, that we probably won't be able to "read" a dir any longer, the reverse conclusion being that we need an individual "handle" (something like "tree:/...") for each and every file. That's where the "localUri" file comes to life.

But if I understand you correctly, having the file name (last path segment, e. g. "test.map") will be enough for file retrieval, as the current map dir can be prefixed to that filename. Then I can happily strip the localUri and replace it by a filename relative to the chosen map directory.

eddiemuc:
> thanks for the fast reply. Hm, I thought that having absolute Uris would make things easier for both SAF migration and the file access after that, as I have a full handle to the file involved.

It would be easier if we could exclude directory changes and what it means ti the logic of c:geo...

> From what I have read so far I'm still unsure what will be possible when using SAF and what won't. At one point in time I thought, that we probably won't be able to "read" a dir any longer, the reverse conclusion being that we need an individual "handle" (something like "tree:/...") for each and every file. That's where the "localUri" file comes to life.

With SAF we get from the user permission for a whole directory (using ACTION_OPEN_DOCUMENT_TREE). Although we won't know exactly where that dir is (well you can conclude most of it from Uri), we will have ful access inside this directory. We can create files, subdirs and subsubdirs... So I assume that what we should store are relative adresses within these dirs. In case of offline maps we have no further subdirs, so file names should do the trick I guess.

> But if I understand you correctly, having the file name (last path segment, e. g. "test.map") will be enough for file retrieval, as the current map dir can be prefixed to that filename. Then I can happily strip the localUri and replace it by a filename relative to the chosen map directory.

Yes, file name is enough. But as I said: we have to deal with the possibility that user changes map dir. Then of course we won't find previously accessible map files in the new dir. Thisnis why I had the idea with storing the info in a file inside offline map dir. Then the info will always be available when the map is.

moving-bits:
> @moving-bits come to think of it: it might imho make sense to store offline-map-related information not in the DB but in a file next to the actual offline map file.

I see the following pros / cons:
- pro: info gets moved with the map file, if the whole dir gets moved
- pro: self-copied map files can be made updateable by adding an manually created info file
- neutral: no path needs to be stored, only local filename (= same could be done with db)
- neutral: map and info files can differ, if someone starts moving around files directly (= same with db)
- con: db handling is easier, thanks to `DBExtension` - or do we have some key/value-pair file writing/reading class?

Such an info file could look like this:
```
remoteparsetype=1
remotepage=http://ftp-stud.hs-esslingen.de/pub/Mirrors/download.mapsforge.org/maps/v5/europe/germany/
remotefile=hessen.map
localfile=map43.map
displayname=Hessen
remotedate=2020-12-10
```

> With SAF we get from the user permission for a whole directory (using ACTION_OPEN_DOCUMENT_TREE). Although we won't know exactly where that dir is (well you can conclude most of it from Uri), we will have ful access inside this directory.

Does that include gettings files listed which c:geo hasn't created? In my (few) tests some months ago I could not make that work.

Use cases: Someone can move a map file into the map directory chosen and c:geo will recognize that file. - Or do we need sort of an "import function" to get such map files known to c:geo?
(Same applies to `log-properties.txt` where we know the filename and can infer the path, but the file is not created by c:geo.)

> Yes, file name is enough. But as I said: we have to deal with the possibility that user changes map dir. Then of course we won't find previously accessible map files in the new dir. Thisnis why I had the idea with storing the info in a file inside offline map dir. Then the info will always be available when the map is.

Ok, the user needs to grant us access to the new dir (with tree access) anyway for c:geo to function.
Now, if we are able to read directory content from the new dir, even if the files are not created by c:geo, then moving the base dir shouldn't be a problem, as we just have to read the new dir. Extra infos could be taken from either info files or from the database (as long as there are only relative paths stored). - But if we don't get access to those files, even if the user granted us access to the new base, then we do have a problem, as we could access neither the map files nor the info files.

eddiemuc:
Hi @moving-bits, I do not want to question the decision to store the info in the database too much. Both solutions will work, and if you already implemented the db-way then maybe its best to stay with it. Just think about whether this still works when user changes the map directory, that's all.

> * con: db handling is easier, thanks to `DBExtension` - or do we have some key/value-pair file writing/reading class?

One could use simply java.util.Properties for that which has load and store methods already build-in.

> > With SAF we get from the user permission for a whole directory (using ACTION_OPEN_DOCUMENT_TREE). Although we won't know exactly where that dir is (well you can conclude most of it from Uri), we will have ful access inside this directory.
> 
> Does that include gettings files listed which c:geo hasn't created? In my (few) tests some months ago I could not make that work.

As far as my tests went, this works. On an Android 11 (emulated) device and with target-sdk 30 set and "legacy-flag" removed I downloaded a map file directly from openandromaps using the device's chrome browser (other app than c:geo) and moved it into the maps folder which was permitted to c:geo using SAF/ACTIOn_OPEN_DOCUMENT_TREE). Afterwards this map could be read by c:geo.

If you have any evidence that this does not work, please send me the necessary test results. This would make large parts of the solution I am currently building obsolete...
Maybe in your tests you did not ask for READ_PERMISSION for the Document-Tree? (if you only ask for WRITE_PERMISSION then maybe you can't see the docs from other apps... but that's just a guesss).

moving-bits:
@eddiemuc 
Thanks for your comments.

> Hi @moving-bits, I do not want to question the decision to store the info in the database too much. Both solutions will work, and if you already implemented the db-way then maybe its best to stay with it. Just think about whether this still works when user changes the map directory, that's all.

That's fine, and actually the reason why I mentioned you in my initial posting, to get some feedback and to reflect how this will work with SAF.

> > * con: db handling is easier, thanks to `DBExtension` - or do we have some key/value-pair file writing/reading class?
> One could use simply `java.util.Properties` for that which has `load `and `store `methods already build-in.

Good to know, thanks for the pointer - I will have a look at it.

> As far as my tests went, this works. On an Android 11 (emulated) device and with target-sdk 30 set and "legacy-flag" removed I downloaded a map file directly from openandromaps using the device's chrome browser (other app than c:geo) and moved it into the maps folder which was permitted to c:geo using SAF/ACTIOn_OPEN_DOCUMENT_TREE). Afterwards this map could be read by c:geo.
> 
> If you have any evidence that this does not work, please send me the necessary test results. This would make large parts of the solution I am currently building obsolete...
> Maybe in your tests you did not ask for READ_PERMISSION for the Document-Tree? (if you only ask for WRITE_PERMISSION then maybe you can't see the docs from other apps... but that's just a guesss).

I will have a look at my (pretty old) test code for this. It's somewhere at the beginning of the `storage_framework` branch.

eddiemuc:
> I will have a look at my (pretty old) test code for this. It's somewhere at the beginning of the storage_framework branch.

Oh my, I am deeply sorry, I deleted this some time ago, squashed commits and force-pushed the new one. Unless you have a direct commit hash this is gone.

I had so many merge hells already rebasing storage_framework to master that at least I wanted to make my life easier by at least having to merge only one commit and not multiple ones.

I see if I can find the commit. Have to google how to do that in git.

Sorry again...

moving-bits:
> > I will have a look at my (pretty old) test code for this. It's somewhere at the beginning of the storage_framework branch.
> 
> Oh my, I am deeply sorry, I deleted this some time ago, squashed commits and force-pushed the new one. Unless you have a direct commit hash this is gone.

No problem. I'll look in my local git repo and/or look for my commit in the online git history.
But it has not been that much code back then, so probably your guess about having used only the "write" permission request might be true.

> I had so many merge hells already rebasing storage_framework to master that at least I wanted to make my life easier by at least having to merge only one commit and not multiple ones.

Very understandable. And I'm pretty happy that you deep-dived into the SAF things. I had read a stack of online docs about it at that time, but still wasn't able to get a grip on it. :-(

eddiemuc:
Thanks for your understanding.

I just learned about reflog (amazing what lies under the hood of git...) and was able to find a commit with your change. In it I see indeed that you acquired WRITE permission and PERSISTABLE permission but not READ permission:

image

(git commit hash is  585e15eea)

moving-bits:
Updated my PR to use `java.util.Properties` instead of a customized `DBExtension`

PR #9613, opened 2020-12-27
wizard update

moving-bits:
updates for installation wizard:
- geocaching platforms: put buttons for "gc.com" and "opencaching & more" on the same page
- integrated "advanced configuration" page into regular workflow; removed "services" button
- show configuration status on last page
- prev/next buttons: enlarged padding, added bottom margin

Configure geocaching platforms:
image

Integrated "advanced configuration" page:
image

Configuration status (with errors):
image

Configuration status (successful):
image


moving-bits:
Updated the wizard to include the following things:
- changelog is no longer shown on fresh installs
- configuration wizard is available from main menu (only for incomplete installations)
- removed about:starting page
- stay on platforms wizard page on returning from gc.com authorization (so that you can add other platforms, if needed)
- make platforms info on final wizard page easier to understand
- prefill (partial) gc.com authorization data (if exists)

moving-bits:
failing opencaching OKAPI client test is unrelated, needs to be investigated separately

Issue #9211, opened 2020-10-18
Waypoint Calculator Read/Write to Personal Notes


Billy-Wallace:
The **Waypoint Calculator** is a great new addition to **cgeo** and I will certainly use it for pre-planning field trips and for aiding me to enter values in the field with less chance of making errors, especially with complex formulas.

I do most of my pre-planning in **GSAK**. What would be a great addition to the WPC is the ability to save the data stored by the calculator to the **Personal Notes** and also to be able to read the WPC formatted data in the PNs and populate the calculator if the data fields are empty for that particular cache. In much the same way as we can now do with additional child waypoints.

This would allow me to create a macro in GSAK or even a spreadsheet (For batch processing) that would fill in as many values as possible and create the formatted WPC string ready to upload to the PNs. It would also offer the opportunity to work with other cachers and share the formatted WPC strings with each other when working towards large adventures.

The PN formatted string would need to be in a format that is easily parsed to populate the WPC fields.
{wpc-start}
@Coordinate Type
@Lat formula
@Long formula
@Main (capital) equation variables (1 to many) (delimited)
@Sub (lowercase) equation variables (0 to many) (delimited)
@Notes
{wpc-end}

I believe the PNs has a 2500 character limit, so there should be plenty of capacity to store personal notes + child waypoint formatted strings + WPC formatted strings.

_Keep up the good work cgeo team, it just keeps getting better and better, adding another fun element to Geocaching._

Billy-Wallace:
I have been using this feature fairly extensively now and finding that for most cases, the Plain option is used.

Preparation in the Cache page Personal Notes, something like:
N50 57.(A+E)(B+B)(E) W001 21.(D)(B+D)(A+B+B)
telephone number 
023806ABCDE
short walk
Checksum 30

Then in cgeo, copy the PNs into the notes of the Waypoint and you have everything needed for the calculation.

So the ability to read/write the above section of the PNs would be all that was required if the team implemented this feature.

murggel:
That would be a great feature for me as well, even, if the value of the variables are somehow exported too. I would use it in addition as archiv-functionaltiy in GSAK.
Or preparing there, calculate some values offline at home and the remaining out in the field.
One more point, but this would be a separate issue:
Some multi have only one set of variables for all stages. So it would be great, if the value of the variables are populated to all waypoints mit formulas, probably only on demand or as an option...

murggel:
I can have a look at it, because I thought about that already.

[...]

Issue #7913, opened 2019-10-23
Waypoint calculator: have a single field for latitude and longitude in simple mode


moving-bits:
Many listings with formulas for final coordinates have latitude and longitude part together, so it would be easier to copy both parts in one move into the waypoint calculator and not having to switch twice between listing and calculator. (This is meant to be in combination with automatically removing whitespace when copying formulas, see my separate feature request.)

eddiemuc:
See #9633

fm-sys:
Additional note: It's nearly impossible to "split" the coords automatically. Therefore we implemented a paste feature some time ago, which does the splitting if it's "easy enough" (not perfect but works in most cases)

PR #9633, opened 2020-12-30
Parse formula for waypoint in personal cache note (related to #9211)

murggel:
preparation and first implementation for parsing the PN for waypoints with formula as coordinates (related to #9211)

I used the GeopointParser as example for the FormulaParser and extracted some basics out of the CoordinateCalculateDialog to use it in the parser.

In first step only plain-formulas are parsed, I already tried some of the other formats, but with the buttons-representation there will probably some more refactoring necessary.

I am preparing a different PR for indicating, that a waypoint has a formula defined.

Supported syntax:
Like the "normal" waypoint-syntax, use (F-PLAIN) to indicate the plain-formula (like (NO-COORD) for empty coords), then the formula string with both coordinates and (optional) already known variables / equations separated with "|".
An existing formula will be exported in the same way, if a waypoint is stored in PN.

Example:
@Final (F) (F-PLAIN) N48 AB.(C*D/2) E 9 (C-D).(A+B) |A=a+b|a=5|

Only upper-case letters are allowed in the formula, for the variables list I don't have a regular expression at the moment. Probably I can add some later.

I also try to formulate some description for the manual.


murggel:
Can someone please have a look on it. I am really open for comments and improvements.

murggel:
I will shortly provide some examples... so someone can look over it and give some comments and improvements.

murggel:
Ok, I have to fix first the codacy-issues and I missed to delete some files

murggel:
Ok, with those 2 remaining issues from codacy I don't know how to handle. 
* I copied the already existing method "parseWaypointType" 1:1 to a new source-file -> NPath complexity to high
* class name FormulaParser does not match, but GeopointParser is also utility class and has similar name

murggel:
I am not sure, whether to leave class `CalculationUtils` in utils or move it to `calculator`

Lineflyer:
Sorry for my stupid question: What is PCN?

murggel:
> Sorry for my stupid question: What is PCN?

Ok, I thought that is a common abbreviation for PersonalCacheNote :-)

eddiemuc:
> Ok, with those 2 remaining issues from codacy I don't know how to handle.
> 
> * I copied the already existing method "parseWaypointType" 1:1 to a new source-file -> NPath complexity to high
> * class name FormulaParser does not match, but GeopointParser is also utility class and has similar name

I think Codacy only parses **changed** code parts. Maybe your findings are a result of changed rules in codacy (so they are applied to your new code but not to the old one)?

fm-sys:
> > Sorry for my stupid question: What is PCN?
> 
> Ok, I thought that is a common abbreviation for PersonalCacheNote :-)

We had used the abbreviation PN (PersonalNote) in some threads. While it is called PersonalCacheNote at GC, c:geo usually just uses the phase "PersonalNote"

murggel:
> > Ok, with those 2 remaining issues from codacy I don't know how to handle.
> > 
> > * I copied the already existing method "parseWaypointType" 1:1 to a new source-file -> NPath complexity to high
> > * class name FormulaParser does not match, but GeopointParser is also utility class and has similar name
> 
> I think Codacy only parses **changed** code parts. Maybe your findings are a result of changed rules in codacy (so they are applied to your new code but not to the old one)?

Thanks for the possible reason, that was also my thought. 
I will check, if I can refactor the class FormulaParser so it will be not a 'utility' class.
For the other function I don't have an idea yet.

eddiemuc:
Regarding the NPath complexitiy warning for method parseWaypointType():

I don't see how this method would be improvable wrt readability by splitting it up. Thus I would consider this a false-positive and would suggest to add followoing annotation:

`@SuppressWarnings("PMD.NPathComplexity") // method readability will not improve by splitting it up`

But I might be  biased because I wrote the original method :-)

Also I don't know how c:geo stands to the usage of "Suppress"-Annotations.  @moving-bits, @Lineflyer what is the convention here?

murggel:
> 
> 
> Regarding the NPath complexitiy warning for method parseWaypointType():
> 
> I don't see how this method would be improvable wrt readability by splitting it up. 

I also considered using some lambda-expression to extract the for-loops in a single method, but it would reduce the readability quite a lot :-(
So I would also prefer to somehow mark it as false positive. 

If it will help, I can add unit-tests for the 40 (oops) cases.

eddiemuc:
I thought about lambdas too, but it will only replace each for-loop-call with a lambda-call and present a few "how-to-transport-previously-calculated-intermediate-results" problems. I had a lambda version in place but decided that it was in fact less readable than the current version...

murggel:
> I thought about lambdas too, but it will only replace each for-loop-call with a lambda-call and present a few "how-to-transport-previously-calculated-intermediate-results" problems. I had a lambda version in place but decided that it was in fact less readable than the current version...


So we are waiting for the go for the suppress-message?
I found some for  unused-warnings, so there is already a SuppressWarning in the code, but no for those major warning like here.

eddiemuc:
Well, you have my personal go for the SuppressWarnings :-)

Maybe you can just put it there and if there are any other valid ideas people can suggest them during code review?

eddiemuc:
But please but a short comment behind the SuppressWarnings. Something like "splitting up this method will not increase it's readability" or so...

moving-bits:
@murggel 
Yes, go ahead. Sometimes it's necessary to suppress or ignore a warning.
If you know it's the way it should be and "ever" will be, you can suppress the warning (but probably leave a comment).
In case of those "npath complexity" warnings I would not suppress them, though, but keep them as a reminder. If someone ever touches that area again, he/she will be reminded, and maybe has, in the new context, an idea how to reduce complexity then.

eddiemuc:
> In case of those "npath complexity" warnings I would not suppress them, though, but keep them as a reminder. If someone ever touches that area again, he/she will be reminded, and maybe has, in the new context, an idea how to reduce complexity then.

@moving-bits the current case is about a Npath complexity warning, and without suppressing it @murggel can't get the code through Codacy. I would vote clearly for suppressing it (with a comment as of why next to the SuppressWarnings annotation) instead of "softening" the Codacy rules. (As discussed in another issue, don't have the id right now...)

moving-bits:
> @moving-bits the current case is about a Npath complexity warning, and without suppressing it @murggel can't get the code through Codacy.

I know, but we can still merge it, even if Codacy is flagged red. In such cases we do a manual check what the reason for the red flag is (in this case: NPath message), and if it's acceptable (as in this case) we continue anyway. (At least that's the way we have done it in the past. But seldom.)

moving-bits:
@murggel 
If you are ready to go with this PR, convert it from "Draft" to "PR" using the "Ready for review" button.

murggel:
> @murggel 
> If you are ready to go with this PR, convert it from "Draft" to "PR" using the "Ready for review" button.

I hope I will find some time in the next days before Homeschooling starts 😣
I need to tests with useless formulas, so no exception should be thrown and prepare some examples for demonstration. 

After that I will change to "ready for review"

eddiemuc:
> > @moving-bits the current case is about a Npath complexity warning, and without suppressing it @murggel can't get the code through Codacy.
> 
> I know, but we can still merge it, even if Codacy is flagged red. In such cases we do a manual check what the reason for the red flag is (in this case: NPath message), and if it's acceptable (as in this case) we continue anyway. (At least that's the way we have done it in the past. But seldom.)

I strongly disagree with this approach. It would mean that CI will continue to mark unrelated pull requests as red (don't know exactly how Codacy decides which code parts it checks). CI must be reliable for all PRs, it is either green or red, and red PRs shall not be merged. There might be hot fix emergency situations where this does not apply, but this isn't such a situation.

In fact, situations like this are the very reason why SuppressWarnings was invented: to mark false positives. By searching for this annotations one can find them later and reinvestigate if need be

Lineflyer:
> I strongly disagree with this approach. It would mean that CI will continue to mark unrelated pull requests as red (don't know exactly how Codacy decides which code parts it checks). CI must be reliable for all PRs, it is either green or red, and red PRs shall not be merged. There might be hot fix emergency situations where this does not apply, but this isn't such a situation.

I cannot judge the codewise quality and approach, but the opinion of @eddiemuc sounds reasonable. So indeed a dedicate annotation of such cases seems meaningful.

murggel:
> > I strongly disagree with this approach. It would mean that CI will continue to mark unrelated pull requests as red (don't know exactly how Codacy decides which code parts it checks). CI must be reliable for all PRs, it is either green or red, and red PRs shall not be merged. There might be hot fix emergency situations where this does not apply, but this isn't such a situation.
> 
> I cannot judge the codewise quality and approach, but the opinion of @eddiemuc sounds reasonable. So indeed a dedicate annotation of such cases seems meaningful.

@eddiemuc , @bekuno Is it possible to automatically create an overview where all those suppress warnings are listed?

eddiemuc:
> @eddiemuc , @bekuno Is it possible to automatically create an overview where all those suppress warnings are listed?

In Android Studio I guess the easiest way is to perform a "Find Usages" search on the annotation class "SuppressWarnings"

I don't know whether jenkins or Codacy can provide such a report.

moving-bits:
> I strongly disagree with this approach. It would mean that CI will continue to mark unrelated pull requests as red (don't know exactly how Codacy decides which code parts it checks). CI must be reliable for all PRs, it is either green or red, and red PRs shall not be merged. There might be hot fix emergency situations where this does not apply, but this isn't such a situation.
> 
> In fact, situations like this are the very reason why SuppressWarnings was invented: to mark false positives. By searching for this annotations one can find them later and reinvestigate if need be

I think we are not that much apart from each other: Static code analysis is important. It's good to have our CI flagging things which break our (self-chosen) rules. The difference in our approaches is how we want to be reminded of breaking things.

One approach is to not suppress a warning, but to overrule manually at merge time. By this it still gets counted by Codacy as a breaking part, and you will get an automatic reminder as soon as anyone touches that area again (at the latest). And you won't get automated reminders in between, so the build do not always stay red. This is important to make this handling useful.

The other approach is to explicitly mark a broken part as suppressed. By this Codacy won't complain during build, so all merges can pontentially be green. Sadly, Codacy will never ever warn about this part again. (Which may be wanted or not.) So you'll have to take care of that by yourself. (And have to remind yourself from time to time to actually do so.)

Which can easily be forgotten. That's why I proposed approach number one. But I can also happily live with approach number two. (Have been using it in some places myself already, where I found it appropriate.) We just have to keep in mind what approach we are using (and when), and that we need to check from time to time the existing code, if some of the suppress rules are still valid.

eddiemuc:
@moving-bits 
there might be a misunderstanding. I do not promote marking "broken" code parts with SuppressWarnings. Quite the opposite: code considered "broken" should never ever be merged into `master `or `release`. But the final decision whether code is broken or not is up to the developer and reviewer, not up to Codacy. Codacy is a very very useful tool, and I am glad we have it. But it is just a tool after all and might falsely flag code as "broken" when in fact it is not according to a developer's/reviewer's judgement.

For each code piece flagged by codacy, the developer (and latest the reviewer) has to make a final decision whether this is really broken code or not:
* In the vast majority of cases Codacy flags will indeed indicate broken code. Then this has to be fixed before merging.
* In some cases, however, the developer/reviewer may decide that codacy was wrong and the code is ok  despite the codacy flag (this is what I call a "false positive"). **Only** in this case SuppressWarnings shall be used, and this is why it is so important to put a comment next to it.

Under no circumstances shall code which is considered broken (by a developer/reviewer)  be marked with SuppressWarnings. Only "false positives" (code pieces which are wrongly flagged by codacy) shall be marked with this. As a consequence there is also no "follow-up-handling" necessary for code pieces marked with SuppressWarning, because they are considered ok.

murggel:
The CalculatorDialog uses CalcState to store the state in the waypoint. It is combined with the buttons of the dialog, so it was not possible to reuse it as it was to parse the CalcState from the PN.

I separated view and data from CalculatorDialog, so CalcState can be created now out of the scope from the dialog. I guess this can be improved, but as a first draft it works right now. 
So a formula can be parsed from the PN and stored in the waypoint.

What I found out today, the coordinates of the waypoint were not updated, if all variables are set. So the user first has to open calculator to get the formula evaluated. 
I am now working on that case, separating the evaluation of the formula from the dialog, but it will lead to more / different refactoring. So I am not sure, whether to put it in this broad PR or create a new one. 
Or probably when finished all task, split all up in smaller PRs (e.g. extraction of some method from waypoint-class in separate aspects (like parsing, note-combining).

@moving-bits @eddiemuc What do you think, as I guess one of you will be reviewer...?

I hope, you understand what I want to explain:-)

eddiemuc:
Hi @murggel I attemt to do a review today

> What I found out today, the coordinates of the waypoint were not updated, if all variables are set. So the user first has to open calculator to get the formula evaluated.

Is this something new (e.g. introduced by your change) or was this there also before your change? Would you consider this a bug or a yet-missing-feature? (sorry if this is a dumb question, I don't know exactly how the calculated coords work)

> I am now working on that case, separating the evaluation of the formula from the dialog, but it will lead to more / different refactoring. So I am not sure, whether to put it in this broad PR or create a new one.

Imho this depends. If you could easily incorporate this in your current change, then why not just do that. If this however 
makes your PR much more complicated or bigger then maybe it should be a separate issue + PR.

> Or probably when finished all task, split all up in smaller PRs (e.g. extraction of some method from waypoint-class in separate aspects (like parsing, note-combining).

Sounds like you want to do some refactoring. I personally prefer refactoring to go together with feature implementation, even if this enlarges the PR. But I know that many developers don't like too big PRs. Also they are hard to "maintain" during the phase of review (if this takes a longer time they have to be rebased to master, and if they are large/complicated then the rebase is compliated too...). So again it is a matter of judgement what is easier to maintain in the end imho.

main/src/cgeo/geocaching/calculator/ButtonData.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.io.Serializable;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Data used to capture the state of this particular button such that it can be restored again later
+ */
+public final class ButtonData implements Serializable, JSONAble {
```
eddiemuc, line 13: c:geo does not use Serializable put Parcelable. But also the later only if needed. I assume here it is probably not needed at all? (since you use json)

main/src/cgeo/geocaching/calculator/ButtonData.java
```diff
@@ -0,0 +1,111 @@
...
+                if ('A' <= buttonData.customChar && buttonData.customChar <= 'Z') {
+                    nextChar = buttonData.customChar;
+                    nextChar++;
+                }
+
+                return nextChar;
+            }
+        };
+
+        public abstract char getLabel(ButtonData buttonData);
```
eddiemuc, line 74: I think "getLabel" and "setAutoChar" could be instance methods of class "ButtonData" instead of moving them into the enum. Or am I missing something?

main/src/cgeo/geocaching/calculator/CalculationUtils.java
```diff
@@ -1,4 +1,4 @@
-package cgeo.geocaching.utils;
+package cgeo.geocaching.calculator;
```
eddiemuc, line 1: +1 for making a separate package out of this! :-)

main/src/cgeo/geocaching/calculator/CaseCheck.java
```diff
@@ -0,0 +1,25 @@
+package cgeo.geocaching.calculator;
+
+/**
+ * Class used for checking that a value is with in a given range.
+ * This is used to check for upper-case an lower-case letters.
+ */
+public final class CaseCheck {
```
eddiemuc, line 7: Could org.apache.commons.lang3.CharUtils be used instead?

If not: can this be moved as method to an existing Utils class? Maybe TextUtils?

main/src/cgeo/geocaching/models/WaypointParser.java
```diff
@@ -0,0 +1,454 @@
...
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Locale;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.jetbrains.annotations.NotNull;
+
+public class WaypointParser {
```
eddiemuc, line 27: Kudos for extraction of the parsing methods into an own class! :-)

main/src/cgeo/geocaching/models/WaypointUserNoteCombiner.java
```diff
@@ -0,0 +1,84 @@
...
+     * Split up given string into note and user note (separated with \n--\n).
+     * For userDefined waypoints only userNote is set.
+     *
+     * @param combinedNote note to split up
+     */
+    public void updateNoteAndUserNote(final String combinedNote) {
+        if (combinedNote != null) {
+            String fixedCombinedNote = combinedNote;
+            // \n was removed via validate
+            if (combinedNote.startsWith("--\n")) {
```
eddiemuc, line 51: Use the constant here instead of hardcoded --

main/src/cgeo/geocaching/utils/TextUtils.java
```diff
@@ -420,6 +428,29 @@ public static String replaceAll(final String text, final String startToken, fina
         return result;
     }
 
+    @NonNull
+    public static List<String> getAll(final String text, final char delimiter, final  Boolean ignoreEmpty) {
```
eddiemuc, line 432: I am not sure I understand what this method is doing. Is there a Unit test for it?

main/src/cgeo/geocaching/calculator/FormulaParser.java
```diff
@@ -0,0 +1,379 @@
...
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+/**
+ * Parse coordinates with formulas.
+ */
+public final class FormulaParser {
```
eddiemuc, line 24: An example formula in the class javadoc would be great

murggel:
> Hi @murggel I attemt to do a review today

Thanks for looking into it.

> 
> > What I found out today, the coordinates of the waypoint were not updated, if all variables are set. So the user first has to open calculator to get the formula evaluated.
> 
> Is this something new (e.g. introduced by your change) or was this there also before your change? Would you consider this a bug or a yet-missing-feature? (sorry if this is a dumb question, I don't know exactly how the calculated coords work)
> 
This is introduced by my changes for the new feature and only occurs on that. No bug on existing workflow. There all is fine.

main/src/cgeo/geocaching/calculator/FormulaParser.java
```diff
@@ -0,0 +1,379 @@
...
+         */
+        @Nullable
+        public abstract String parse(@NonNull List<String> groups);
+    }
+
+
+    /**
+     * Parser for MinDec format: X DD° MM.MMM'.
+     */
+    private static final class MinDecFormulaParser extends AbstractLatLonFormulaParser {
```
eddiemuc, line 193: Is MinDecFormulaParser the only subclass of AbstractLatLonFormulaParser? If yes, why do we need the subclass?

main/src/cgeo/geocaching/calculator/FormulaParser.java
```diff
@@ -0,0 +1,379 @@
...
+        private static final String STRING_LAT = "([NS]?)" + STRING_MINDEC;
+
+        //                                        (   1  )    (    2    )    (      3      )
+        private static final String STRING_LON = "([WEO]?)" + STRING_MINDEC;
+        private static final String STRING_SEPARATOR = "[^\\w'′\"″°." + WPC_DELIM + "]*";
+        private static final Pattern PATTERN_LAT = Pattern.compile(STRING_LAT, Pattern.CASE_INSENSITIVE);
+        private static final Pattern PATTERN_LON = Pattern.compile("\\b" + STRING_LON, Pattern.CASE_INSENSITIVE);
+        private static final Pattern PATTERN_LATLON = Pattern.compile(STRING_LAT + STRING_SEPARATOR + STRING_LON, Pattern.CASE_INSENSITIVE);
+
+        MinDecFormulaParser() {
```
eddiemuc, line 213: There should be coordinate parsers for many coordinates formats in some other class available (alas, they don't parse for variable values inside of course). Can't they be reused somehow (parametereized to accept also letters instead of numbers)?

main/src/cgeo/geocaching/calculator/FormulaParser.java
```diff
@@ -0,0 +1,379 @@
...
+         * Indicates which format is parsed
+         * @return coordInputFormat which can be parsed
+         */
+        public abstract Settings.CoordInputFormatEnum formulaFormat();
+    }
+
+    /**
+     * Abstract parser for coordinates that consist of two syntactic parts: latitude and longitude.
+     */
+    private abstract static class AbstractLatLonFormulaParser extends AbstractFormulaParser {
```
eddiemuc, line 102: Is AbstractLatLonFormulaParser the only subclass of AbstractFormularParser? If yes, why do we need the subclass?

main/src/cgeo/geocaching/calculator/FormulaParser.java
```diff
@@ -0,0 +1,379 @@
...
+        ResultWrapper(final String result, final int stringLength) {
+            this.result = result;
+            this.matcherLength = stringLength;
+        }
+    }
+
+    /**
+     * Abstract parser for coordinate formats.
+     */
+    private abstract static class AbstractFormulaParser {
```
eddiemuc, line 70: I think this abstract class could be replaced by an interface?

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /**
+     * Find if a variable exists in the supplied list with the given name
+     *
+     * @param name name to search for
+     * @param list list of variables
+     * @return first occurrence of the variable if it can found, 'null' otherwise
+     */
+    private static VariableData getVariable(final char name, final List<VariableData> list, final boolean remove) {
```
eddiemuc, line 134: This methods performance would profit from "list" being a map (char -> VariableData).  Would that be an adequate data structure for a collection of VariableData's?

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
```diff
@@ -0,0 +1,163 @@
...
+        return null;
+    }
+
+    /**
+     * Find if variable data exists in the supplied list with the given name
+     *
+     * @param name name to search for
+     * @return first occurrence of the data if it can found, 'null' otherwise
+     */
+    private static VariableData findAndRemoveData(final char name, final List<VariableData> list) {
```
eddiemuc, line 153: Isn't this the same as calling "getVariable" with "remove=true"?

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
```diff
@@ -0,0 +1,163 @@
...
+        final char[] sortedVariables = variableNames.toCharArray();
+        Arrays.sort(sortedVariables);
+
+        for (final char ch : sortedVariables) {
+            if (theCase.check(ch)) {
+                if (getVariable(ch, returnList, false) != null) {
+                    continue;
+                }
+
+                VariableData thisEquation = getVariable(ch, variables, true);
```
eddiemuc, line 45: This method modifies the content of parameter "variables", which should not be done. At least there should be a comment on the method telling which params will be modified.

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
```diff
@@ -0,0 +1,163 @@
...
+public final class CoordinatesCalculateUtils {
+
+    /** Flag values used to designate that no AutoChar has been set */
+    public static final char EMPTY_CHAR = '-';
+
+    private CoordinatesCalculateUtils() {
+        // Do not instantiate
+    }
+
+    public static List<VariableData> sortVariables(final List<VariableData> variables,
```
eddiemuc, line 30: I dont understand difference between "variables" and "varBank". Maybe a short comment would help?

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
```diff
@@ -0,0 +1,163 @@
...
+                                                   final List<VariableData> varBank,
+                                                   final String variableNames,
```
eddiemuc, line 32: If I understand the code correctly, then the value of "variableNames" could be extracted from list "variables". Is this correct? Then this parameter should be removed.

murggel:
> Imho this depends. If you could easily incorporate this in your current change, then why not just do that. If this however 
> makes your PR much more complicated or bigger then maybe it should be a separate issue + PR.
> 
> > Or probably when finished all task, split all up in smaller PRs (e.g. extraction of some method from waypoint-class in separate aspects (like parsing, note-combining).
> 
> Sounds like you want to do some refactoring. I personally prefer refactoring to go together with feature implementation, even if this enlarges the PR. But I know that many developers don't like too big PRs. Also they are hard to "maintain" during the phase of review (if this takes a longer time they have to be rebased to master, and if they are large/complicated then the rebase is compliated too...). So again it is a matter of judgement what is easier to maintain in the end imho.
> 

@eddiemuc 
Ok, having this in mind, I think at the moment it looks like it affects same files as in the current PR, so probably it would be then better to incorporate it in the current PR.

Now you have to decide, if you want to already start review or wait some more days...

tests/src/cgeo/geocaching/calculator/FormulaParserTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.junit.Test;
+import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
+import static org.assertj.core.api.Java6Assertions.assertThat;
+
+public class FormulaParserTest {
```
eddiemuc, line 13: Kudos for writing unit tests! :-)

eddiemuc (commented):
Few comments made, found nothing serious.

eddiemuc:
> @eddiemuc
> Ok, having this in mind, I think at the moment it looks like it affects same files as in the current PR, so probably it would be then better to incorporate it in the current PR.

I agree.
 
> Now you have to decide, if you want to already start review or wait some more days... it is still a draft only...

Whoops, already did one :-)

murggel:
> 
> > @eddiemuc
> > Ok, having this in mind, I think at the moment it looks like it affects same files as in the current PR, so probably it would be then better to incorporate it in the current PR.
> 
> I agree.
>  
> > Now you have to decide, if you want to already start review or wait some more days... it is still a draft only...
> 
> Whoops, already did one :-)

No problem, it's good for me to have early feedback :-)

main/src/cgeo/geocaching/calculator/ButtonData.java
murggel: I just moved the class out of the CalculatorDialog as whole.
But since there are different types of Buttons, (CustomInput, Blank, AutoChar and CustomChar) which leads to different behaviour, this is handled via the abstract methods getLabel and setAutoChar of the enum

main/src/cgeo/geocaching/calculator/CaseCheck.java
murggel: I just moved the class out of the CalculatorDialog as whole.

But I will have a look into the package if it fits the needs

eddiemuc:
@murggel when I reviewed the code I was not aware that many of the "new" code was just copied from other locations. So sorry. Of course you don't have to clean up this code if you don't want to (would be nice though)

murggel:
> @murggel when I reviewed the code I was not aware that many of the "new" code was just copied from other locations. So sorry. Of course you don't have to clean up this code if you don't want to (would be nice though)
> 
> 

If it is not too complicated I will give it a try

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Nice idea, didn't thougt about that. I will have a look into that...

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: No, there can / should be the others as well, but with your next comment, probably that will be done with reusing the other parser...

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Probably yes,  but with your previous comment, probably that will be done with reusing the other parser...

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Probably yes, but with your previous comment, probably that will be done with reusing the other parser...

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
murggel: Seems so.
I see 3 possibilities:
1. call getVariable(...false) in the method findAndRemoveData
2. replace the whole method findAndRemoveData with getVariable(...false)
3. leave both methods and remove the parameter remove from getVariable and adapt all calls of getVariable(....true)  with findAndRemoveData
4. combination out of all having 3 functions: 
  - findVariableData without boolean
  - findVariableDataAndRemove
  - findVariableData with boolean: the first two calls the third one

I am not sure, which one I prefer, but variant 1 is the less preferred

BTW isn't there any package in java which can handle that?

murggel: With your next comment, probably variant 3 or 4 will be best

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
murggel: I will rename the function and add description

main/src/cgeo/geocaching/utils/TextUtils.java
murggel: I remove it with String.split ;-)

main/src/cgeo/geocaching/models/WaypointUserNoteCombiner.java
murggel: Done

main/src/cgeo/geocaching/calculator/CaseCheck.java
murggel: Moved it as method to TextUtils

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Done

main/src/cgeo/geocaching/calculator/ButtonData.java
murggel: Done

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
murggel: Done

main/src/cgeo/geocaching/calculator/ButtonData.java
murggel: Hm, I am not sure, if it's needed for storing in DB? But I guess, there only the JSON-String is stored.

On testing on the emulator, I got some leaks-messages, which I have not seen before. I will switch to local master-branch, to check, whether they appear there as well or I have to take a closer look at this. Probably I will need your help then ...

eddiemuc: I can assure you that serialization is not needed for db store. 

Was this also in the classes before? (In case you moved them) I have the strong feeling it is not needed, but it also doesn't hurt.

murggel: It was already in the class before move.
I moved it then for VariableData as well.

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: I looked into it and remembered that I thought about it, but the other parser returns a Point /Double, whereas I need a string. So I just used it to copy and modify.
But I will think, if probably it can be reused somehow else, or just extend the classes with methods returning a string, which then can be converted to double ...

eddiemuc: If its too complicated don't do it. I assume the formulas only work for MicDec-formatted coordinates anyway, do they?

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
eddiemuc: I like 3 best :-)

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: You are probably right, for the other formats only letters will be accepted instead of digits, but no formula.
Formulas there only can appear in the variables.

main/src/cgeo/geocaching/models/WaypointParser.java
murggel: As reminder for me:
check, if there are more changes in parsing methods in master which has to be adapted to WaypointParser

bekuno:
retest this please

murggel:
> retest this please

I am still working on the review comments, therefore it is a draft ...

But nevertheless it can be tested

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
murggel: Done

main/src/cgeo/geocaching/calculator/CoordinatesCalculateUtils.java
murggel: Done

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Did not manage yet to extract the logic. 

I will keep the AbstractParser for the other coordinate-formats, if there will be also parsers for it later

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: changed to interface

main/src/cgeo/geocaching/calculator/FormulaParser.java
murggel: Did not manage yet to bring to a common base-class or logic...

If I have some more time, I will probably give it another try...

murggel:
@eddiemuc  I added now the calculation of the coordinates after parsing the formula.

murggel:
Parsing of the formula can be probably used for #7913.

Lineflyer:
retest this please

Job result seems inconclusive (No build running, but shown as pending)

murggel:
> retest this please
> 
> Job result seems inconclusive (No build running, but shown as pending)

Seems to be ok now.

Because it is still a draft, the pending probably means that state...

eddiemuc:
Build failed with

07:46:23  Execution failed for task ':main:installBasicDebug'.
07:46:23  > com.android.builder.testing.api.DeviceException: Could not create ADB Bridge. ADB location: /opt/android-sdk-linux/platform-tools/adb
07:46:23  

This seems to be a general problem in CI, not related to this specific PR

eddiemuc:
retest this please

fm-sys:
@murggel I did not follow the thread completely, but is the PR ready for merge?

murggel:
> @murggel I did not follow the thread completely, but is the PR ready for merge? 

No, not yet. But I hope I can finish it soon

murggel:
@eddiemuc I think th PR is in a state now to continue review, if you have time ;-) As far as I remember the commits starting with the one "evaluate formula after parsing" are the ones, which are not reviewed yet.
Sorry for having such a long break, but it was a busy time here.

I would squash the commits, once the review is finished, correct?

fm-sys:
> I would squash the commits, once the review is finished, correct?

You don't need to. This can also be done automatically with just a single click while merging...

eddiemuc (approved):
Hi @murggel, as far as I can see the code is very well structured and you changed everything as we discussed in the last review. I want to mention explicitely that you added many unit tests as well - which I support very much!

One quick question before I merge this: what exactly did you change since last review (I mean functionality-wise)?

murggel:
> Hi @murggel...
> One quick question before I merge this: what exactly did you change since last review (I mean functionality-wise)?
@eddiemuc I added
* evaluating the formula if variables are defined and set additional the calculated coordinates. 
* only upper-case letters are allowed for the formula (change in regExpr)
* a (owner-defined) waypoint is user-modified if it has a calcState (before it was only having coords or user note)


Missing points:
* Regular expression for valid characters in variables-expression
* text for the manual

PR #9637, opened 2020-12-31
merge old cache logs on importing GPX/PQ (fix #9524)

### LINKED ISSUE #9524: Merge save logs with logs from pocket query

When I update a cache via c:geo it downloads many more logs and I can get from my pocket query.  I like having this additional information when I'm out in the field.  However I have weekly pocket queries of a 1,000 caches in my area and unfortunately the pocket query only contains the last five logs.  When I update my offline list with my pocket query the previously saved logs are discarded leaving me with only five.

I'd like to have c:geo retain the previous logs and merge that with the information from the pocket query.  I realize this could leave gaps in the log history but that could be indicated somehow.

It is possible that there might be other data that is currently discarded that could be retained as well, favorite points come to mind...

moving-bits:
Solves #9524, but also merges old logs on importing from other GPX files or updating a cache using the connector.
Merging is done by per-log deleting old logs from same author, same date, same type before storing the new logs.

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCParser.java
@@ -420,7 +420,7 @@ public final class GCParser {
         final SearchResult result = new SearchResult(parsed.left);
         if (parsed.left == StatusCode.NO_ERROR) {
             result.addAndPutInCache(Collections.singletonList(parsed.right));
-            DataStore.saveLogs(parsed.right.getGeocode(), getLogs(parseUserToken(page), Logs.ALL).blockingIterable());
+            DataStore.saveLogs(parsed.right.getGeocode(), getLogs(parseUserToken(page), Logs.ALL).blockingIterable(), true);
         }
         return result;
     }
@@ -1815,7 +1815,7 @@ public final class GCParser {
                     mergeFriendsLogs(logEntries, specialLogEntries);
                     return logEntries;
                 }).cache();
-        mergedLogs.subscribe(logEntries -> DataStore.saveLogs(cache.getGeocode(), logEntries));
+        mergedLogs.subscribe(logEntries -> DataStore.saveLogs(cache.getGeocode(), logEntries, true));
         if (cache.isFound() || cache.isDNF()) {
             ownLogs.subscribe(logEntry -> {
                 if (logEntry.getType().isFoundLog() || (!cache.isFound() && cache.isDNF() && logEntry.getType() == LogType.DIDNT_FIND_IT)) {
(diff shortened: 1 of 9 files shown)


fm-sys:
Sounds promising! 👍

SammysHP:
But what about actually deleted logs? And logs with wrong type that was changed afterwards (probably less common)? This will cause inconsistencies with upstream data. Will caches be detected as "found" if there was a found log previously, but then deleted?

moving-bits:
> But what about actually deleted logs? And logs with wrong type that was changed afterwards (probably less common)? This will cause inconsistencies with upstream data. Will caches be detected as "found" if there was a found log previously, but then deleted?

There is indeed some trade-off, if you want to keep older logs without having to reload each cache after importing a GPX file. Personally, this would be acceptable for me in these cases, but probably not for everyone.

@SammysHP 
Do you have a proposal how to mitigate that problem without losing stored logs, and without having to completely reload them?
Or should I add a setting, so that the user can decide which way is preferred, and default that setting to today's behavior?

SammysHP:
> Do you have a proposal how to mitigate that problem without losing stored logs, and without having to completely reload them?
Or should I add a setting, so that the user can decide which way is preferred, and default that setting to today's behavior?

Not really. I think for imports from GPX it is OK to merge the logs this way, but when refreshing the cache by downloading the data from GC (or OC, where we get all logs AFAIK) we should replace all logs.

When parsing the data (from GPX), is the found status (also) determined from the logs or only from a separate flag?

moving-bits:
I have adapted the PR to merge logs only on importing from GPX/PQ.
All other actions which lead to storing logs behave the same as before this PR.

moving-bits:
> When parsing the data (from GPX), is the found status (also) determined from the logs or only from a separate flag?

To me it looks like it is parsed from waypoint status:

```
// waypoint.sym
waypoint.getChild(namespace, "sym").setEndTextElementListener(body -> {
    sym = body.toLowerCase(Locale.US);
    if (sym.contains("geocache") && sym.contains("found")) {
        cache.setFound(true);
    }
});
```

or from log in GPX file:
```
gcLog.setEndElementListener(() -> {
   final LogEntry log = logBuilder.build();
   if (log.getType() != LogType.UNKNOWN) {
       if (log.getType().isFoundLog() && StringUtils.isNotBlank(log.author)) {
           final IConnector connector = ConnectorFactory.getConnector(cache);
           if (connector instanceof ILogin && StringUtils.equals(log.author, ((ILogin) connector).getUserName())) {
               cache.setFound(true);
               cache.setVisitedDate(log.date);
           }
       }
       logs.add(log);
   }
});
```

(and other possible places)

SammysHP:
So you can't "unfind" a cache by importing a GPX with these changes?

moving-bits:
> So you can't "unfind" a cache by importing a GPX with these changes?

No change in this regard, to my understanding.

Haven't looked in the code in detail regarding this aspect, but my guess would be, that found state is set to false on creating the cache element in memory during import, and GPX import can set it to true from the imported data (but not from data stored in the log database tables).

PR #9638, opened 2020-12-31
Extends the OneTimeDialogs framework

fm-sys:
Extends the OneTimeDialogs framework so that dialogs can also be displayed only for new or updating users.

I have configured the existing OneTimeDialogs like this:
- explain offline found counter --> all users
- database overwrite notice --> of course also all users
- map quick settings --> only upgrading users

also see https://github.com/cgeo/cgeo/pull/9597#issuecomment-751507462

Sorry, but this PR will lead to one minor merge conflict, when merging it back to master. Unfortunately unavoidable...


moving-bits:
merge to release & master done

Issue #9589, opened 2020-12-26
Installation wizard


moving-bits:
While browsing through the Play Store review notifications I've stumbled upon a wide spread between recurring users (beta reviews) and other users (regular reviews). While most beta reviews are four or five star reviews, other reviews have a high occurence of one star reviews, frequently with remarks like
- no caches shown
- no search does work
- app does not work
- Did not let me see ANY geocaches only let me see the map
- ...

IMHO there is a problem with onboarding our new users, the learning curve for c:geo is too steep.

A couple of days ago, while working on a different issue, I had to install a new app "Lockito" for automatically looping over a self-created track. I had no idea how this app worked, but after installation a short wizard took me through the most relevant settings, and so I had this app up and running in no time.

That's when my old idea of a tiny installation wizard for c:geo resurfaced. Today I started working on it, and have come up with the following so far:
|![image](https://user-images.githubusercontent.com/3754370/103156498-65544a00-47a9-11eb-9618-134388b362fd.png)|![image](https://user-images.githubusercontent.com/3754370/103156501-6ab19480-47a9-11eb-9e38-99e84dd40000.png)|
|![image](https://user-images.githubusercontent.com/3754370/103156504-6eddb200-47a9-11eb-9f3e-b61393443bd1.png)|![image](https://user-images.githubusercontent.com/3754370/103156508-7309cf80-47a9-11eb-9d20-17730ac60e5c.png)|
|![image](https://user-images.githubusercontent.com/3754370/103156512-7735ed00-47a9-11eb-94a2-ac0cd9633d29.png)|![image](https://user-images.githubusercontent.com/3754370/103156515-7c933780-47a9-11eb-99d6-131706367871.png)|
|![image](https://user-images.githubusercontent.com/3754370/103156517-8157eb80-47a9-11eb-9a70-b655f6bd4243.png)|![image](https://user-images.githubusercontent.com/3754370/103156519-85840900-47a9-11eb-8130-bef2ab590a9b.png)|
|![image](https://user-images.githubusercontent.com/3754370/103156521-8a48bd00-47a9-11eb-9796-7684192242cf.png)||

As a "bonus" you can select "advanced" on the last screen for an easy access to some more advanced settings:
![image](https://user-images.githubusercontent.com/3754370/103156524-8fa60780-47a9-11eb-97db-c1cc2cb6dd60.png)

This, BTW, also makes our BRouter integration a bit more prominent, as has been mentioned (wished for) earlier in some other issue.

[...]

PR #9651, opened 2021-01-03
Related to #9589 - Wording proposals for installation wizard

Lineflyer:
Based on feedback of a "new user". 
Feedback and other proposals welcome (you may push to this branch).

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2029,27 +2029,27 @@
     <string name="next">Next</string>
     <string name="finish">Finish</string>
     <string name="wizard_welcome_title">Welcome to c:geo</string>
-    <string name="wizard_intro">It looks like this is the first time c:geo is used on this device. Shall I help you with configuration? It\'s just a few steps.</string>
-    <string name="wizard_intro2">Welcome back to the configuration wizard. It\'s just a few steps.</string>
+    <string name="wizard_intro">Just a few steps are needed to configure c:geo and get you going. Shall I help you with that?</string>
+    <string name="wizard_intro2">Welcome back to the configuration wizard. Your c:geo setup is not yet finished, shall I help you to complete it and get you going? It\'s just a few steps.</string>
     <string name="wizard_outro_ok">Thanks for configuring c:geo.\nHave fun!</string>
     <string name="wizard_outro_error">Incomplete configuration\nNot all parts of c:geo will work as expected!</string>
     <string name="wizard_permissions_title">Set app permissions</string>
-    <string name="wizard_permissions_intro">c:geo needs certain permissions to function correctly. I will explain them in the next steps.\nWithout these permissions you will not be able to use c:geo.</string>
-    <string name="wizard_platforms_title">Configure platforms</string>
-    <string name="wizard_platforms_intro">To download caches you need an account for at least one of the geocaching platforms supported by c:geo</string>
+    <string name="wizard_permissions_intro">c:geo needs certain permissions on your device to work correctly.\nI will explain more details in the next steps.</string>
+    <string name="wizard_platforms_title">Configure services</string>
+    <string name="wizard_platforms_intro">To find geocaches with c:geo you need an account for a geocaching service supported by c:geo.\nSelect the desired service below to either login with your existing account or create a new account.</string>
     <string name="wizard_platforms_gc">geocaching.com</string>
-    <string name="wizard_platforms_others">Opencaching &amp; more</string>
+    <string name="wizard_platforms_others">Opencaching &amp; others</string>
     <string name="wizard_welcome_advanced">Advanced configuration</string>
     <string name="wizard_advanced_offlinemaps_label">Offline maps</string>
-    <string name="wizard_advanced_offlinemaps_info">c:geo can download map files for offline usage</string>
+    <string name="wizard_advanced_offlinemaps_info">This page provides some advanced functions, which are optional to use.\n\nc:geo can download map files for offline usage.</string>
     <string name="wizard_advanced_brouter_label">Routing app</string>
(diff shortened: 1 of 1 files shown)


main/res/values/strings.xml
```diff
@@ -2029,27 +2029,27 @@
...
+    <string name="wizard_intro2">Welcome back to the configuration wizard. Your c:geo setup is not yet finished, shall I help you to complete it and get you going? It\'s just a few steps.</string
     <string name="wizard_outro_ok">Thanks for configuring c:geo.\nHave fun!</string>
     <string name="wizard_outro_error">Incomplete configuration\nNot all parts of c:geo will work as expected!</string>
     <string name="wizard_permissions_title">Set app permissions</string>
-    <string name="wizard_permissions_intro">c:geo needs certain permissions to function correctly. I will explain them in the next steps.\nWithout these permissions you will not be able to use c:geo.</string>
-    <string name="wizard_platforms_title">Configure platforms</string>
-    <string name="wizard_platforms_intro">To download caches you need an account for at least one of the geocaching platforms supported by c:geo</string>
+    <string name="wizard_permissions_intro">c:geo needs certain permissions on your device to work correctly.\nI will explain more details in the next steps.</string>
+    <string name="wizard_platforms_title">Configure services</string>
```
bekuno, line 2038: Hmm, we had the problem before that the dialog presents "services" as acronym for services, where you manage caches (geocaching.com, opencaching*. , ...) and services for additional information (rating, ...). To distinguish the use cases the term platform was used for the cache-provider. 
I think we have to use a clean wording so that new user can this distinguish.
So I would stay with platform, if this is correct for english native speeker, or at least use the term "geocaching service" all the time.

fm-sys: Maybe "Configure geocaching services" or "Configure geocaching platforms" to make the meaning a bit more clear...

fm-sys: I took a look at the installation wizard, and you are able to configure both, platforms as well as further services. So "Services" is the correct wording...

main/res/values/strings.xml
```diff
@@ -2029,27 +2029,27 @@
...
+    <string name="wizard_platforms_intro">To find geocaches with c:geo you need an account for a geocaching service supported by c:geo.\nSelect the desired service below to either login with your existing account or create a new account.</string>
     <string name="wizard_platforms_gc">geocaching.com</string>
-    <string name="wizard_platforms_others">Opencaching &amp; more</string>
+    <string name="wizard_platforms_others">Opencaching &amp; Others</string>
```
fm-sys, line 2041: Shouldn't "others" be lowercase?

main/res/values/strings.xml
moving-bits: I had intentionally chosen the wording "platforms", as the focus in this step should be to select a geocaching platform, like gc.com or oc.de, not a specialized service like gcvote.com

I had a trade-off to make here: gc.com is by far the most widely used platform (especially for geocaching beginners, which this wizard is mostly intended for), so I put in an extra button for it, linked directly to the specific settings page. The other platforms are much less common and more regional, so I did not want to put an extra button on the page for every other platform, but just linked to our general services settings page - which indeed lists both services and platforms. Maybe we need to split-up our settings pages accordingly into "platforms" and "other services"?

main/res/values/strings.xml
moving-bits: yes

main/res/values/strings.xml
```diff
@@ -2029,27 +2029,27 @@
...
     <string name="wizard_advanced_offlinemaps_label">Offline maps</string>
-    <string name="wizard_advanced_offlinemaps_info">c:geo can download map files for offline usage</string>
+    <string name="wizard_advanced_offlinemaps_info">This page provides some advanced functions, which are optional to use.\n\nc:geo can download map files for offline usage.</string>
     <string name="wizard_advanced_brouter_label">Routing app</string>
-    <string name="wizard_advanced_brouter_info">c:geo can navigate to targets and show routing info on your map by installing and configuring a companion app named \"BRouter Offline Navigation\"</string>
+    <string name="wizard_advanced_brouter_info">By installing and configuring a companion app named \"BRouter Offline Navigation\" c:geo can navigate to targets and show routing infos on your map.</string>
     <string name="wizard_advanced_restore_label">Restore</string>
-    <string name="wizard_advanced_restore_info">Restore settings / data from a previous installation</string>
+    <string name="wizard_advanced_restore_info">You can restore settings and/or geocache data from a previous installation.</string>
```
moving-bits, line 2048: "You can" may imply that there is stored data available. We should either check for existing backups (there's probably already a helper function for that) or adapt the text accordingly, e. g. "Try to restore ..."

moving-bits (changes requested):
Nicely done, thanks for cross-checking with a pure muggel ;-)

Only had a few comments, mainly about services vs. platforms.

main/res/values/strings.xml
fm-sys: IMHO a geocaching platform is also a service. In our settings we use that wording too.

Lineflyer: Thats why I also ended up with service here. If I look into c:geo Settings the first item is "Services".

main/res/values/strings.xml
Lineflyer: I will try to find a better wording not implying that a backup is available.

Lineflyer: Done

main/res/values/strings.xml
Lineflyer: Done

main/res/values/strings.xml
Lineflyer: No change done here.

moving-bits (approved):

PR #9666, opened 2021-01-04
option to use individual cache markers (fix #9523)

### LINKED ISSUE #9523: Emoji's as cache icons  for UDC's
I use UDC's for tagging all different types of places like my home, beautiful locations, supermarkets, etc.
For easy differentiation, it would be very useful if a custom emoji could be set as cache icon...

moving-bits:
Allows using unicode characters as cache markers on a per-cache base.
Fixes #9523 and gives a mitigation for other issues requesting to set individual cache icons without the hassle of handling user-selected image files, and without blowing c:geo's APK size.

- New menu entry in cache menu "Select cache icon"
- opens a popup where you can select an "icon" (= unicode character) from a given list (*)
- at the bottom of this popup a "recently used" list builds (max. 10 entries)
- you can reset the chosen icon to default in this popup
- the selected "icon" is used as cache marker on the map (and other places)
- on the map it uses the usual insets, and the "squared" white background (the same that is used for OC caches and UDC)
- "big smileys" option has precedence over the user-selected icon
- map cache markers with individual cache icons are using the marker cache (as the regular markers do)
- db migration adds a single field (int) to the cache table, and sets all existing caches to "default marker" (= 0)

For screenshots see https://github.com/cgeo/cgeo/issues/9523#issuecomment-753641788 and https://github.com/cgeo/cgeo/issues/9523#issuecomment-753669988.

(*) The list of characters supported currently is for demonstration purposes only, the list has to finalized yet.

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -125,7 +125,8 @@ public class DataStore {
         DBEXTENSION_PENDING_DOWNLOAD(1),
         DBEXTENSION_FOUNDNUM(2),
         DBEXTENSION_DOWNGRADEABLE_DBVERSION(3),
-        DBEXTENSION_ONE_TIME_DIALOGS(4);
+        DBEXTENSION_ONE_TIME_DIALOGS(4),
+        DBEXTENSION_EMOJILRU(5);
 
         public int id;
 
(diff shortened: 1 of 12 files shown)


main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -225,7 +227,8 @@
             88, // add timestamp to trail history
             89, // add altitude to trail history
             90, // add user guid to cg_caches and cg_logs
-            91  // add fields to cg_extension
+            91, // add fields to cg_extension
+            92  // add emoji id to ca_caches
```
bekuno, line 231: s/ca_caches/cg_caches/

moving-bits: done

PR #9667, opened 2021-01-04
set "on long tap" option default to true + add OneTimeMessage (fix #9615)

moving-bits:
- sets default for "on long tap" map option to true
- shows a OneTimeMessage on opening map, content depending on current setting (see screenshots)
- extends the OneTimeMessage framework to (optionally) support a "more information" URL (see screenshots)
- code deduplication: moves the existing map OneTimeMessage messages to `maps/MapUtils`

image.image

"More information" will take to our user manual, map settings section.


Lineflyer:
can't we display the one time message as soon as the user performs his first long click?

moving-bits:
> can't we display the one time message as soon as the user performs his first long click?

We probably could (haven't checked), but with showing this on opening the map I'd like to make some advertisement for individual route & UDC functions.

bekuno:
Personally I would make the text of the second version (long tab is disabled) with the same content and order as the first one.

Proposal:
- headline
- description (By long tapping ...)
- state (is enabled/is deactivated)
- settings (You can change the behavior ...)

moving-bits:
Good idea, I've unified the strings accordingly:

image

PR #9673, opened 2021-01-04
skip PM-only caches on storing

### LINKED ISSUE #9583: c:geo tries to download PM-only caches for BM
User report for beta version 2020.12.22-RC (and verified by me):

- Be logged in as basic member to gc.com
- Search for caches in Munich, Germany
- Load more results to have 40 caches in search result
- Select to store them to a new list (List name automatically set to Munich, Germany)
- Wait until download is finished
- Open the created list

Result:
The list has 40 caches and thus also contains premium only caches, which (up to my knowledge) have been skipped in earlier versions of c:geo while trying to save. Usually I had e.g. 40 caches as result but after downloading them only 40-premium caches in the list

Expected:
It makes no sense to store them as I cannot open them.
I do also see strange cache titles (e.g. for GC5G596) as presumably c:geo cannot resolve the unicode characters as it cannot open the PM cache.

Comment Lineflyer:
I could reproduce it exactly as described above.
Not sure since when this problem occurs, but I agree to the user, that in former versions the PM-only were contained in search results (which is OK) but skipped when storing.

fm-sys:
fixes #9583

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -17,6 +17,7 @@ import cgeo.geocaching.command.MakeListUniqueCommand;
 import cgeo.geocaching.command.MoveToListAndRemoveFromOthersCommand;
 import cgeo.geocaching.command.MoveToListCommand;
 import cgeo.geocaching.command.RenameListCommand;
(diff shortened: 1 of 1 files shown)


bekuno:
should target release?

fm-sys:
> 
> 
> should target release?

You are right, retarget done.

PR #9693, opened 2021-01-05
add set cache icon option to manage caches submenu (fix #9685)

### LINKED ISSUE #9685: Batch function to set cache icon for all caches in list
I have imported a Adventure Lab GPX to see adventure lab caches around my homezone.
At the moment those are imported with the default (grey ??) icons.

In order to better identify those on the map a custom icon would be useful but needs to be set manually for all 200 caches on my list.

The menu "Manage lists" could get an item "Set cache icons" which provides a batch function to set the icon for all caches on the list...same as already implemented for a single cache.

The reset function could however be a bit more tricky. In my case all caches have the same default icon...but that might be different for other scenarios and we should - in any case - provide some batch undo/reset in case someone uses this function accidentally.

moving-bits:


main/res/values/strings.xml
```diff
@@ -1833,6 +1833,7 @@
         <item quantity="many">Add this cache to favorites (%d remaining)</item>
         <item quantity="other">Add this cache to favorites (%d remaining)</item>
     </plurals>
+    <string name="command_set_cache_icons_progress">Setting cache icons</string>
```
Lineflyer, line 1836: Maybe: "Set custom cache icons"
I would also suggest this for the cache menu: "Set custom cache icon"

Lineflyer: On the other hand some translation (e.g. german=benutzerdefiniert) would then be rather long...mmh. 
Unsure..

bekuno: or "Set own cache icons", or ...personal... ?

moving-bits:
I've set the menu entry in Cache => Manage Caches to "Set cache icons" (plural) for now. We can adapt it later, when we find a good phrasing.

Also added an undo function.

PR #9697, opened 2021-01-06
edit map behavior strings

fm-sys:
We had used the `>` character also at other places in our strings to direct users to a settings sub menu, so it would be consistent to use them here as well. 

Example:
`"Settings" > "Logging"` in the offline counter one time message

While I'm quite sure that we should use the `>` instead of `-` I'm wondering if we also want to use quotation marks here...?


Lineflyer:
What about `→` as separator?
That what we use in FAQ.

Quotation marks are IMHO not necessarily needed.

moving-bits:
@fm-sys 
Have you seen the comment from @Lineflyer? Do you want to update your strings or stay with current version and merge this PR?

fm-sys:
Yes I had seen it but hadn't time to edit the PR so far... Will do it right now!

moving-bits:
No problem. Just wanted to send a signal that I haven't ignored your PR by not merging it :-)

fm-sys:
grafik

PR #9702, opened 2021-01-06
tiny adjustments to findWaypoint

moving-bits:
see https://github.com/cgeo/cgeo/pull/9678#discussion_r552693205 ff

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -1640,12 +1640,10 @@ public class Geocache implements IWaypoint {
         }
 
         //try to match name if prefix and coords are null
-        for (final Waypoint waypoint : waypoints) {
-            final String searchWpName = searchWp.getName();
-            if (!StringUtils.isBlank(searchWpName)) {
-                final String wpName = waypoint.getName();
-                final WaypointType wpType = waypoint.getWaypointType();
-                if (searchWpName.equals(wpName) && searchWp.getWaypointType().getL10n().equals(wpType.getL10n())) {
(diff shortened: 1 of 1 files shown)


murggel:
similar posted in (before I saw this PR) https://github.com/cgeo/cgeo/pull/9678#discussion_r552734502
Hm, I think there was a problem with waypoint-type OWN and WAYPOINT. Because they have same name and hence same waypoint-name created from that, a duplicate waypoint with same name could be created, so I checked the names. 
Checking the special waypoint types separately here?

moving-bits:
> Hm, I think there was a problem with waypoint-type OWN and WAYPOINT.

Do you have some test data I can check this against?

fm-sys:
> > Hm, I think there was a problem with waypoint-type OWN and WAYPOINT.
> 
> Do you have some test data I can check this against?

Shouldn't such special cases be also added as Unit Tests?

murggel:
> 
> 
> > > Hm, I think there was a problem with waypoint-type OWN and WAYPOINT.
> > 
> > 
> > Do you have some test data I can check this against?
> 
> Shouldn't such special cases be also added as Unit Tests?

you are right, I will check later (no kids needs maintenance :-) ), if there is some ... hm, build is red - test is red - probably this is one of those :-)

moving-bits:
Ah, now I see. Then let's stay with `getL10n()` comparison - the tests are green in that case.

murggel:
> Ah, now I see. Then let's stay with `getL10n()` comparison - the tests are green in that case.

Ok, thanks for looking into it and sorry for the confusion.

PR #9721, opened 2021-01-07
set 'vote' menu entry to greyed out, if connector inactive

moving-bits:
Related to https://github.com/cgeo/cgeo/discussions/9692:

Grey out the "vote" menu entry in cache details when either the GCVote connector is disabled or credentials are incomplete.
(It is already hidden when the current cache does not support GCVote.)


moving-bits:
retest this please

PR #9735, opened 2021-01-08
check glyph availability for custom icons

moving-bits:
- Android 6+: Automatically removes unsupported characters from emoji selector
- Android 5: Shows a OneTimeMessage about not being able to auto-filter unsupported characters
- Selected icons for icon groups which are supported even on Android 5

Android 5:
image.image

Android 6 vs. Android 10:
image.image


moving-bits:
Failing check is due to GeoKrety downtine, otherwise all tests are fine, so let's merge.

PR #9741, opened 2021-01-09
stay with connector-dependent marker for caches with custom icons

moving-bits:
see https://github.com/cgeo/cgeo/issues/9685#issuecomment-757126689


moving-bits:
Failing check is due to GeoKrety downtime, otherwise tests are fine, so let's merge.

PR #9746, opened 2021-01-09
never show configuration wizard as menu action

fm-sys:
It is not really intuitive to have two identical icons with different meanings. As an unambiguous "configuration wizard icon" does not exist, it's the best to always hide the icon...

Before it was like this:
grafik


Lineflyer:
oh...that probably means for the issue I opened regarding wizard vs. google backup, I did click on the wizard but not on settings.

PR #9754, opened 2021-01-10
add dependabot to c:geo

bekuno:
Configure dependabot for the project and the 3 subprojects.
Add ignore entries vor commons-io.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,15 +9,16 @@ updates:
     directory: "/cgeo-contacts"
     schedule:
       interval: daily
-  ignore:
-    # Ignore some updates to the apache 'commons-io' package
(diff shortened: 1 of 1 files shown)


fm-sys:
There is an official dependabot [config file validator](https://dependabot.com/docs/config-file/validator/) which currently fails for this configuration file. Don't know what's wrong through, maybe some indentations...?

bekuno:
> 
> 
> There is an official dependabot [config file validator](https://dependabot.com/docs/config-file/validator/) which currently fails for this configuration file. Don't know what's wrong through, maybe some indentations...?

Thanks for checking and pointing to the validator.
I missed a line, see the new commit.

bekuno:
File is ok in the validator, Build fails for GeoKreaty service, which is unreleted.
So lets merge and wait for the results.

Issue #6672, opened 2017-08-02
Sorting order lost after "update caches"


moving-bits:
##### Detailed steps causing the problem:
- show a list of caches
- sort by name
- chose "select mode" (Auswahlmodus)
- select a couple of caches
- chose "manage caches" => "update"

##### Actual behavior after performing these steps:

Selected caches get updated, but sort order is lost and cannot be restored as long as I'm in "select mode". 

Also my selection is lost, thus I cannot move the freshly updated caches to another list without having to reselect them first.

##### Expected behavior after performing these steps:

Sort order and selection should be kept.

##### Version of c:geo used:
2017.07.24

##### Is the problem reproducible for you?
Yes

##### System information:
```
--- System information ---
Device: SM-A510F (a5xeltexx, samsung)
Android version: 7.0
Android build: NRD90M.A510FXXU4CQG2
c:geo version: 2017.07.24
Google Play services: enabled - 11.3.02 (438-161239932)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: magnetometer & accelerometer
Hide own/found: true
Map strategy: auto
HW acceleration: enabled (default state)
System language: de_DE
System date format: dd.MM.yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (2,8 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (2,8 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (2,8 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (117,6 MB) on system internal storage
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
Geocaching.com date format: yyyy-MM-dd
Installed c:geo plugins:  none
--- End of system information ---
```

Lineflyer:
Might be related to #4507 , #5689

fm-sys:
Sorting order is fixed with #9774. I renamed the issue according to the remaining problem

Issue #376, opened 2011-09-05
Problems with quick selection in cache lists


cachapa:
I've been looking at the code for the cache lists (in the "nearby" and "stored" activities) trying to enable visual feedback when an item is selected, clicked or long-pressed.

There are a few problems in showing the list selectors, and most of them are related to the UI for selecting multiple caches.
To be clear, I think that selecting multiple caches is a nice feature, but I don't really like the way it's currently implemented. Particularly, the fact that one can fling any cache to the right to go into a kind of hybrid single-and-multiple-select mode.

My problems with this are as follows:
1. It's not obvious that the functionality is there;
2. Sometimes the user tries to tap an item, but flings it by mistake;
3. When one item has been flung, the list internally goes into the select-mode. In that mode, tapping a cache doesn't work anymore, even though visually the other caches still look clickable (this is made worse is the mode was started by mistake);
4. Due to the way the fling detection works, it's basically impossible to enable visual feedback for those list items;
5. There's a lot of code that goes into supporting this functionality which could be greatly simplified. I've also identified a few bugs related to that code which would be taken care of. Also, scrolling the list in multiple-select mode would get a significant performance boost.

My proposal is to do away with the hybrid selection mode, and keep the exclusive multiple selection mode which can be enabled from the menu.
I've been studying the code and I can do this myself, but I'd like your opinions on this before starting.

PR #9774, opened 2021-01-10
allow sorting also when in selection mode

### LINKED ISSUE #4507: Sort don't work while in selection mode
Steps to reproduce : 
- go into a list
- go into selection mode
- Sort by name

Sort has no effect on the list.
Filter seems to work anyway !

fm-sys:
fixes #4507 and partly #6672

It looks like this was explicitly introduced, but I can't find a reason for it. All my resorting tests work fine, also when in selection mode. Looking at the history, it is there since 2012. I have read through some issues linked in the old commit messages but it seems like as there was no real reason, so let's remove it...

### DIFF
--- a/main/src/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/cgeo/geocaching/ui/CacheListAdapter.java
@@ -303,7 +303,7 @@ public class CacheListAdapter extends ArrayAdapter<Geocache> implements SectionI
     }
 
     public void forceSort() {
-        if (CollectionUtils.isEmpty(list) || selectMode) {
+        if (CollectionUtils.isEmpty(list)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Lineflyer:
mmh..Maybe it was introduced to prevent resorting when sorted by distances and distance changes while selecting (e.g. close nearby caches or you are walking/driving)?

fm-sys:
> 
> 
> mmh..Maybe it was introduced to prevent resorting when sorted by distances and distance changes while selecting (e.g. close nearby caches or you are walking/driving)?

That's the issue where the bug/feature was introduced: #376 (4befdab254fcee09ecaeb08811c19ed9eb4d9249) Your guess might be true, however It could also be that it was only changed by mistake or is no longer relevant.

The fact that two caches swap positions during the selection is IMHO much more acceptable (or even wanted?) than if the sorting is simply ignored without any optical feedback.

bekuno:
retest this please

PR #9780, opened 2021-01-11
emoji-style presentation 

fm-sys:
Add `U+FE0F` (VARIATION SELECTOR-16) to all emojis to force emoji-style presentation 

Before:
grafik

Now:
grafik


moving-bits:
I like the results - was always a bit concerned why some characters stay white.

One request, though: I'm still heavily working on emojiselector / custom icons behind the scenes. Please postpone PR regarding emojis / custom icons / map markers a couple of days. Thanks.

bekuno:
retest this please

moving-bits:
@fm-sys 
Thanks for your patience and for adapting it to the current emojiselector implementation

PR #9782, opened 2021-01-11
add some more usefull emojis to selector iconset

fm-sys:


fm-sys:
Setting do not merge for now... (see https://github.com/cgeo/cgeo/pull/9780#issuecomment-758229969)

bekuno:
retest this please

moving-bits:
@fm-sys 
I've merged my emoji selector changes regarding the list markers, so you may go ahead with your two PR for custom icons. Please be aware that I have removed the circle emojis from the "symbols" group, as we have a dynamic replacement for them now. Also I have introduced a character space from the Unicode private use area to be used for the dynamic circles. There are constants provided regarding start and end of the used area - should be excluded from adding the variant selector.

PR #9792, opened 2021-01-13
correctly scroll in emoji selector

fm-sys:
Before that the LRU was unreachable if the emoji list was too long...

2021-01-13 15_34_43-cgeo – emojiselector xml  cgeo main  Android Studio


moving-bits:
@fm-sys 
Good catch, thanks!

Issue #9791, opened 2021-01-13
Parsing of "Plain format" strings in Waypoint Calculator


S-Bartfast:
Greetings all.

I have been much time away from this project but now finally with some time I hope to reacquaint myself with you all again. As the original designer of the Waypoint Calculator and finally with some time to spare I am hopeful that I may be able to contribute to things again. After some reading of the current issues it appears that many of the issues appear to be related to the parsing of the "Plain format" text in the calculator. Is this the case?

As it was, the original intent of the calculator was to enter coordinates in one of the regular formats (such as `DDD°MM.MMM'`) then substitute individual digits with equations (such as `N 51°24.AB5'`, with `A=a, B=a+2`). The "Plain format" was really only added as an option as an after thought more for completeness sake than anything, but I take it this "Plain" format is being used most frequently (such as `"N 51 24.a(a+2)5"`). Is this the case?

Assuming the "Plain" format is indeed being used frequently I take it there are usability issues regarding the parsing of the text. If so could we discuss the short comings of the current parser here and I will do my best to address them.

Thank you,
Slarti.

S-Bartfast:
Hmm, I seem to be having trouble working out how to navigate this service again: 😊 
How do I attach labels to this issue (ie. WP Calculator), and how do I assign it to myself?

S-Bartfast:
The first issue I've come across seems to be related to white space in the text: #7910

From memory this was done intentionally and uses the same rules for parsing as dose the regular "Plain format" parser for the non-calculator input. That is `"42 12.345"` would be parsed as `N 42­° 12.345'`, and `"42 12 345"` would be parsed as `N 42° 12' 345` and `"42 12.3 4 5"` would be parsed as an error.

From memory, the way the calculator works is that it substitutes all the equations for their respective values then attempts to parse the result as a valid coordinate format. Given that the result is required to have specific white space requirements in order to be parsed as a valid coordinate the formula entered into the field is required to have the same.

murggel:
@S-Bartfast 
I am a "normal" user of the app and a c++ developer who is new to java and app development. In the last time I used the calculator a lot - most plain format - and stumbled over some issues. So I took a look into it. Mainly to parse the personal note to extract formula for a waypoint.
See my attempts in #9633 / #9211.

Please have a look on that.
Still missing there to complete the workflow, but I am working on it, is to extract the evaluation of the formula without the Calculator-dialog.

I am stuck on extracting the other formats.
[...]

PR #9800, opened 2021-01-16
#9791 setting case in calculator

S-Bartfast:
Implementing filters for calculator input fields.

### DIFF
--- a/main/res/layout/coordinates_equations.xml
+++ b/main/res/layout/coordinates_equations.xml
@@ -20,11 +20,6 @@
         android:orientation="horizontal"
         android:padding="6dp" >
 
-        <RelativeLayout
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:padding="0dp">
-
             <androidx.gridlayout.widget.GridLayout
 
                 android:id="@+id/EquationTable"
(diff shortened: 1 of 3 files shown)


bekuno:
ok to test

moving-bits:
@bekuno 
Currently we have no Codacy checks (not only on this build) - can you check why, please?

bekuno:
@moving-bits 
codacy has currently a maintenance period, see https://status.codacy.com/
`In Progress - Database Maintenance  - Scheduled for Jan 16, 10:00 - Jan 16, 19:00 (CET)`

moving-bits:
@bekuno 

> codacy has currently a maintenance period, see [status.codacy.com](https://status.codacy.com/)
> `In Progress - Database Maintenance - Scheduled for Jan 16, 10:00 - Jan 16, 19:00 (CET)`

good to know. Thanks for researching it.

fm-sys (approved):
Great! I tested your code and everything works as expected. And using an `InputFilter` is codewise even better (I didn't know that something like this exists) then an `onTextChangeListener`.

Only thing which is totally unrelated to your PR:  Should we deactivate suggestions for the input fields? Thst can be done via `android:inputType="textNoSuggestions|textVisiblePassword"` according to SO for the best compatibility

S-Bartfast:
> using an `InputFilter` is codewise even better (I didn't know that something like this exists) then an `onTextChangeListener`.

Yeah, I wasn't aware of it either till I started trying to work it out.

Turned out there already was a TextChangeListener associated with the edit text (it was being used to validate and update the resultant coordinate field in real time). At first I tried using this listener but it had the problem that it would 'listen' for text to be entered, examine the entered text and modify it if necessary.

Although this worked it had the problem that the updating would actually set new text causing the listener to be invoked a second time. It was easy enough to ignore the second call but it was still a little messy. The filter mechanism however  works perfectly and seems to have been made for pretty much exactly this purpose: 👍

PR #9804, opened 2021-01-16
add calendar marker for future events with own will attend (fix #6517)

### LINKED ISSUE #6517: Make "Will attend" state more visible in geocache lists
Improvement proposal via support mail:

I am attending a large number of events.
In order to keep a better overview of my planned attendence it would be great if after logging "Will attend" this would be visualized in geocache lists (e.g. by a color marker or an icon).

moving-bits:
adds a little calendar icon top/left if
- cache is an event in the future,
- current user has logged an "will attend",
- but not "attended" yet

image

image


fm-sys:
Nice! But I'm wondering if we really should display the "7" always inside the calendar icon...?

moving-bits:
> Nice! But I'm wondering if we really should display the "7" always inside the calendar icon...?

Nothing special about the "7", just a number to indicate a calendar. I was looking for something that even works at a 12px x 12px resolution. Do you have a more prominent icon in mind?

fm-sys:
My first thought was that the actual day is displayed, which is not the case. But I'm also fine with leaving it like that. You're right, that the icon is so small that it isn't really important.

But what will happen if the "large found emojis" setting is activated? Will the calendar icon also be displayed large (which would be consistent)? And there displaying the correct day in the icon might makes sense (if it's not to much effort)...

Lineflyer:
Does this replace also the grey frowning smiley in case of an offline logged WA?

moving-bits:
> But what will happen if the "large found emojis" setting is activated? Will the calendar icon also be displayed large (which would be consistent)? And there displaying the correct day in the icon might makes sense (if it's not to much effort)...

"Found" (= attended) state will win over the "will attend" state, so only the smiley will be shown in that case. This marker is not a general-purpose replacement for all events, but only a personal reminder for upcoming events which you have logged a "will attend" for.

fm-sys:
> > But what will happen if the "large found emojis" setting is activated? Will the calendar icon also be displayed large (which would be consistent)? 
> 
> "Found" (= attended) state will win over the "will attend" state, so only the smiley will be shown in that case. This marker is not a general-purpose replacement for all events, but only a personal reminder for upcoming events which you have logged a "will attend" for.

Sorry, I think I expressed myself badly. I meant what happens if you haven't attended yet. Will the calendar icon will be displayed enlarged (which IMHO should be the case)?

moving-bits:
> Does this replace also the grey frowning smiley in case of an offline logged WA?

Not yet, offline log info seems to have precedence currently. Shall I change this?

bekuno:
retest this please

moving-bits:
Added "will attend" marker also for offline logs + double-size it (if configured)

PR #9805, opened 2021-01-16
show disabled/archived status on map also for waypoints (fix #5008)

### LINKED ISSUE #5008: Waypoints of disabled caches not shown as disabled on map
When a cache is disabled (grey backgnd in map) then the either

a) waypoints should have a grey background, too. Or
b) should not be displayed at all

What do you think? a) or b)?

Related to that: if you click on a waypoint of a disabled cache, then the little window with the waypoint short details & cache short details appears. Here the title of the cache should be striked thru but isn't (see screenshot).

K

screenshot_2015-06-05-13-00-04

moving-bits:
precedence is visited > archived > disabled > regular

image


bekuno:
retest this please

PR #9813, opened 2021-01-17
SAF string improvements

Lineflyer:
Some proposals for string improvements/corrections:

- Add articles at places, where they have been omitted while its still a sentence (not only a toast, which should be kept short)
- Try to phrase some operations more understandable for users
- Remove the "See logfile for details" as no normal user would know hoe or where.


eddiemuc:
Retest this please

eddiemuc (approved):
All fine of course. I just wonder why ci build fails

eddiemuc:
Seems like there is an xml error in the file:

`home/jenkins/slave/workspace/cgeo pull request/main/res/values/strings.xml:2125:3: Resource and asset merger: The element type "string" must be terminated by the matching end-tag "</string>".
22:05:55      org.xml.sax.SAXParseException; lineNumber: 2125; columnNumber: 3; The element type "string" must be terminated by the matching end-tag "</string>".
22:05:55      	at `

Lineflyer:
Upsi...will correct it.
Thanks CI

PR #9823, opened 2021-01-19
Guide user if brouter is not installed (fix #9819)

### LINKED ISSUE #9819: Deactive routing settings (or guide user) if brouter is not installed
Since we have the quick settings out on the market, it seems many users like it and use it.
Anyway since then also some users complaint came in like "I selected bike but it still shows the straight line. Why is that?". It turned out those users did not have brouter installed.

Indeed those settings are shown despite brouter being not installed on the device.
The easy solution would be to hide these settings completely in that case. The better solution would be to grey out all options except "straight line" and when the user clicks on another option to inform him, that he needs brouter to use that feature (ideally with link to the app and/or to the user guide).

fm-sys:
Guide the user to the playstore if brouter is not yet installed. For screenshots see https://github.com/cgeo/cgeo/issues/9819#issuecomment-762661620

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1287,6 +1287,7 @@
     <string name="map_as_list">Show as list</string>
     <string name="map_select_multiple_items">Select item</string>
     <string name="map_routing">Routing</string>
+    <string name="map_routing_brouter">Routing is only available if <a href="">BRouter</a> is installed</string>
     <string name="map_gm_autorotation">Map auto rotation</string>
     <string name="map_gm_autorotation_disable">Disable map auto rotation?\n\nYou can enable it again using the Map rotation menu</string>
     <string name="quick_settings">Quick settings</string>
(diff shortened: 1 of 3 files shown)


main/res/values/strings.xml
```diff
@@ -1287,6 +1287,7 @@
     <string name="map_as_list">Show as list</string>
     <string name="map_select_multiple_items">Select item</string>
     <string name="map_routing">Routing</string>
+    <string name="map_routing_brouter">Routing is only available when <a href="">brouter</a> is installed</string>
```
fm-sys, line 1290: I'm not that good at English, is "when" correct or should it rather be "if"?

Lineflyer: yes, IMHO it should be "if"

main/res/values/strings.xml
```diff
@@ -1287,6 +1287,7 @@
     <string name="map_as_list">Show as list</string>
     <string name="map_select_multiple_items">Select item</string>
     <string name="map_routing">Routing</string>
+    <string name="map_routing_brouter">Routing is only available if <a href="">brouter</a> is installed</string>
```
moving-bits, line 1290: LGTM, thanks for implementation. Just a tiny thing: The app is named "BRouter", or "BRouter Offline Navigation" (though I would stay with the short form) - can you adapt this, please?

fm-sys: Done!

PR #9826, opened 2021-01-19
show correct MapMarker in CachePopup

fm-sys:
Userdefined cache icons were ignored in the CachePopup, therefore swich to MapMarkerUtils for suporting user-defined cache icons and listmarkers in the cache icon...


fm-sys:
This will look like this:
grafik

We can also use `CacheListType.OFFLINE` if we want to hide the offline diskette. What do you think?

moving-bits:
> We can also use `CacheListType.OFFLINE` if we want to hide the offline diskette. What do you think?

As this is the cache popup shown when coming from the map we could use the same iconography as on the map, meaning: keep the offline diskette.

moving-bits:
> Userdefined cache icons were ignored in the CachePopup

Yes, this was the (last?) missing piece - thanks for completing it!

PR #9827, opened 2021-01-19
Remove some relics from the time of Eclipse

fm-sys:
There were many null analysis assertions in the code which are not necessary anymore when using Android Studio. Lint has complained that this condition must always be true and should be removed, which is done with this PR...


bekuno (approved):
LGTM

PR #9828, opened 2021-01-19
fixes incorrect compass rose when turning screen while in upright mode (fix #6829)

### LINKED ISSUE #6829: Compass rose incorrect when using "upside down" portrait mode
I did not even know, that this exists, but some phones seem to support a 180° display flip if you turn the phone upside down in portrait mode.
c:geo seems to follow this, but of course the compass rose bitmap is then wrong as it needs to be counting the degress counter-clockwise in this case.

Alternatively we should disable the usage of this upside down mode at all.

Device used (reported via support mail):
```
--- System information ---
Device: EVA-L09 (EVA-L09, HUAWEI)
Android version: 7.0
Android build: EVA-L09C432B391
c:geo version: 2017.10.23
Google Play services: enabled - 11.7.44 (448-173432861)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide own/found: false
Map strategy: auto
System language: de_DE
System date format: dd.MM.yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (17,1 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (17,1 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (17,1 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (204,0 KB) on system internal storage
Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
```

fm-sys:
Hopefully this fix doesn't introduce bugs on other devices, but at least for me everything works fine...

Edit: OK, I have thought again and by logic it should actually work for all devices!

### DIFF
--- a/main/src/cgeo/geocaching/CompassActivity.java
+++ b/main/src/cgeo/geocaching/CompassActivity.java
@@ -361,7 +361,7 @@ public class CompassActivity extends AbstractActionBarActivity {
 
                 updateDistanceInfo(geo);
 
-                updateNorthHeading(AngleUtils.getDirectionNow(dir.getDirection()));
+                updateNorthHeading(dir);
 
                 updateDeviceHeadingAndOrientation(dir);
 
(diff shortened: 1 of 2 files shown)


moving-bits:
Let's test it in the nightlies :-)
Thanks for providing a fix.

Issue #5407, opened 2016-01-20
Inconsistent list of saved lists


MisterAlea:
I can organize my saved caches in lists. There are two ways to select which list I want to see:
- Long tap the icon for lists in the c:geo home screen. Then a list can be selected from the list of lists. This opens the list itself.
- Short tap the icon for lists in the c:geo home screen. This opens the last used list. Now the list of lists can be opened from a dropdown on the top of the screen.

The two lists of lists are different. The first one does not contain the entries "history" or "new list" but the other one does. OK, I can open the history list from another icon on the home screen but I needed a discussion with rasch to recognize this. ;-)

My expectation is that the two lists of lists are consistent. 

Best Regards,

MisterAlea

PS: Thanks for improving and maintaining this great piece of software.

Lineflyer:
I agree.

Lineflyer:
setting for recheck. Assigning myself.

moving-bits:
As of today (release 2020-01-10) status is:
- on long press in homescreen you'll get all persisted lists + entries "new list" and "all"
- from within a list, using the drawer in the action bar, you'll get all persisted lists + "all" and "history"

On homescreen, the history is available using the "history" (clock) symbol in the action bar, and from within a list, creating a new list is done using "manage lists" => "new list".

Do we need to add the missing entries to the two lists?

MisterAlea:
> Do we need to add the missing entries to the two lists?

I would appreciate if you would do it.

moving-bits:
Done with #9830 and #9836, both lists do contain the same entries now.
But be aware that we added "history" entry to mainscreen, but removed "new list" from it, as this would break workflows for many people if added to the cache list's list spinner.

Lineflyer:
> "history" entry to mainscreen, but removed "new list" from it

Now I am confused....but maybe I need to look at the result in tomorrows nightly:
IMHO we had "Create new list" ever since when using the long press on lists on c:geo main screen. I suggested only to not add it to the dropdown selection in the top bar while already inside a list.

moving-bits:
The original issue here was to make both lists the same (long press on "store" on mainscreen and list spinner in cache list). Thus I added "history" to mainscreen and "new list" on cache list list spinner at first.

[...]

PR #9836, opened 2021-01-21
remove 'new list' from main screen 'stored' long press menu (rel. to #5407)

moving-bits:


Lineflyer:
Would this lead to "New list" being the first entry in the pulldown?
I think this will break many users workflow when switching back from some custom list to the default list IMHO.
Maybe better put it at the end here?

moving-bits:
Yes, will be the first entry. Will have to look up if this can be set otherwise.

The related issue was about making the mainscreen's "long press on stored" and the cachelist's spinner consistent (which would for me include "in the same order"). So put "create new" at the end for both?

Or maybe: Leave out "create new" in both lists? - I would find it sort of confusing to find this action item in the cachelist spinner anyway, and I - personally - have no need to have it on the mainscreen's long press, but don't know how other users use this regularly or not.

fm-sys:
> Or maybe: Leave out "create new" in both lists? - I would find it sort of confusing to find this action item in the cachelist spinner anyway, and I - personally - have no need to have it on the mainscreen's long press, but don't know how other users use this regularly or not.

Totally agree. A new list only needs to be created whenever you want to move/save caches. And in both cases, the function for creating a new list is available...

Lineflyer:
Also agree.

rsudev:
While having a 'create new' item in a selection is IMO not so uncommon I would also prefer/agree to leave it out in the above mentioned contexts.

moving-bits:
Adapted this PR accordingly to remove "new list" from stored's long press menu instead of adding it to the cache list's list selector.

PR #9843, opened 2021-01-23
add toasts for remove waypoints button (fix #9656)

### LINKED ISSUE #9656: Deactivate image buttons "extract waypoints to PN" and "delete extractred waypoints from PN" when they are not useful
**Describe the bug:**
Recent c:geo introduced several new features. 
One can be seen in the cache screen, below the description.

**To Reproduce:**
1. open a cache
2. go to description screen, scroll down
3. under "personal note" you will see 3 buttons. always.
btn-4

4. try to press the indicated button. **Nothing happens.**

From the icon I deduce it might have something to do with some waypoints. Being in this section would mean some waypoints of a different nature then the usual waypoints available on their dedicated screen. 
The trashcan icon in the upper right suggest to me that this button is meant to delete something. 
I presume there would be situations when under Personal note you could see some waypoints. And this button was meant to delete some/all of those waypoints. 

However, since there are no waypoints, there is nothing to delete. 
IMO this button should be shown only when there actually is content for it to act upon.

**Version of c:geo used:**
2020.12.03

Yes

moving-bits:
Adds toasts for the "remove waypoints" button. Now all four combinations are covered:
- synch waypoints: waypoints were added to PN / no storable waypoints found
- remove waypoints: removed waypoints / no waypoints found to remove

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -950,6 +950,8 @@
     <string name="cache_personal_note_edit">Edit</string>
     <string name="cache_personal_note_storewaypoints">Add/Update Waypoints</string>
     <string name="cache_personal_note_removewaypoints">Remove Waypoints</string>
+    <string name="cache_personal_note_removedwaypoints">Removed waypoints</string>
+    <string name="cache_personal_note_removewaypoints_nowaypoints">No waypoints found to remove</string>
     <string name="cache_personal_note_storewaypoints_success">Waypoints were added to personal note</string>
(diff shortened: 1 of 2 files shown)


Lineflyer:
@moving-bits 
The string `cache_personal_note_removedwaypoints` is meant as a confirmation, that waypoints have been removed, isn't it?
Shouldn't it be rather named "Waypoints removed" instead of "Removed waypoints" as the latter could be understand as a title where a list of waypoints follows. It also makes it difficult to translate without knowing the context.

moving-bits:
Yes, you're right. I've updated the string.

PR #9852, opened 2021-01-24
replace ButterKnife with viewbindings in about pages (rel. to #9232)

moving-bits:
This is a first shot at the viewbinding migration needed for the next Gradle version (and due to ButterKnife being discontinued). It changes ButterKnife `@BindView` to Android's viewbindings for our about pages.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -64,6 +64,10 @@ android {
         release
     }
 
+    buildFeatures {
+        viewBinding true
+    }
+
     buildTypes {
         debug {
             // debug build name contains git commit for better reproduction of bugs
--- a/main/src/cgeo/geocaching/AboutActivity.java
+++ b/main/src/cgeo/geocaching/AboutActivity.java
@@ -1,6 +1,11 @@
 package cgeo.geocaching;
 
 import cgeo.geocaching.activity.AbstractViewPagerActivity;
+import cgeo.geocaching.databinding.AboutChangesPageBinding;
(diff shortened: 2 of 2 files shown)


fm-sys:
Without diving deeper into it, the bindings look a bit complicated. Would it be an alternative initialize the class fields in the on create method via `findViewById`?

moving-bits:
> Without diving deeper into it, the bindings look a bit complicated. Would it be an alternative to initialize the class fields in the on create method via `findViewById`?

Actually it's much less boiler plate code as we had with `Butterknife`, where we had to write a `@BindView` for each field manually. Now it's just the `<GeneratedClassName>.inflate()` call instead of the regular inflate call, and then we can use the fields directly. The class name is derived from the layout file name, BTW, so the bindings for layout file `map_download_selector.xml` will lead to a generated view bindings class called `MapDownloadSelectorBinding`.

For details see https://developer.android.com/topic/libraries/view-binding#java

eddiemuc:
I agree with @moving-bits . If my understanding is correct, we don't need own class fields for each view any more with this approach because we have them all in the generated binding class. This would be a huge simplification imho.

moving-bits:
retest this please

fm-sys:
Ok, thanks for the clarification. I think I have got it now and totally agree with you. Quite cool 👍😁

moving-bits:
Let's wait with this PR and steps building upon it after the SAF merge => added "WIP" label

Issue #8409, opened 2020-06-06
Export/Import of waypoints fails or incomplete


Pentatops:
This are two bugs:
a) gpx export does not export waypoints without assigned coordinates
b) Gpc export/import does not transfer the "already visited" flag

To Reproduce:
1. create a new empty list of caches
2. load GC84G97, GC8GWH8 and GC84VDM from geocaching.com into the list
3. GC84G97: mark the waypoints "das Tier" and "Wegweiser" as visited
4. do a gpx export of the list (with options transfer and include status)
5. transfer the gpx file to another 2nd device (I used a cloud service)
6. import the gpx file into e new list
==> GC84G97 is missing the waypoints without coordinates, e.g. "Bild 1"
7. Go back to the first device
8. GC84G97: remove the flag "visited" from waypoint "das Tier" and set the flag "visited" at waypoint "Heermannshütte"
9. same export-import as above to the second device
==> the "visited" flags on the 2nd device remain unchanged neither has the flag of "Heermannshütte" be added nor the flag of "das Tier" be removed

Expected behavior/state after performing these steps:
- all waypoints should be transferred, independent if they have valid coordinates
- if, in the source, a waypoint has the flag "visited" set this setting should be transferred, too.

Version of c:geo used: 2020.04.25 on all devices

Is the problem reproducible:
Yes;

**System information:**
 I reproduced it on several different devices (Samsung, ASUS) using different Android versions (4.2; 5.0.1; 7.0; 10)

```
Attach system information of one of the 2nd devices:
Device: P00A (WW_P00A, asus)
Android version: 7.0
Android build: NRD90M.WW_P00A-V5.3.20-20180809
c:geo version: 2020.04.25
Google Play services: enabled - 20.18.17 (040400-311416286)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: absent
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: magnetometer & accelerometer
Hide caches: own/found
Hide waypoints: -
HW acceleration: enabled (default state)
System language: de_DE
System date format: dd.MM.yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (3,0 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (3,0 GB free) external non-removable
Geocache data: /storage/3934-3030/Android/data/cgeo.geocaching/files/GeocacheData (7,3 GB free) external removable
[...]

PR #9854, opened 2021-01-24
fix #9849: set modified coordinates for originalEmptyCoordinates on import

### LINKED ISSUE #9849: GPX import does not set coords of Owner defined WPs, even if user has set coords
**Describe the bug:**
Had a Multi this weekend, which has several WPs predefined by the CO.
These WPs have no coords by default - but I had filled some of them.
As we wanted to finish the Multi together, I exported to GPX and shared it with my buddies,
but with import of this GPX, they did not get the coords I had set in the predefined WPs.

Imported WPs did not get the coords, I have set before exporting the cache.

Even predefined WPs should get changed coord.

**Version of c:geo used:**
2021.01.24-NB-f03a863

Yes

murggel:
Waypoints with empty coordinates are getting exported with #8409. 
Unfortunately the cache-coordinates were set to null, if originalCoords were empty.

### DIFF
--- a/main/src/cgeo/geocaching/export/GpxSerializer.java
+++ b/main/src/cgeo/geocaching/export/GpxSerializer.java
@@ -200,7 +200,7 @@ public final class GpxSerializer {
             gpx.text("true");
             gpx.endTag(NS_CGEO, "userdefined");
         }
-        if (waypoint.getCoords() == null || waypoint.isOriginalCoordsEmpty()) {
+        if (waypoint.isOriginalCoordsEmpty()) {
             gpx.startTag(NS_CGEO, "originalCoordsEmpty");
             gpx.text("true");
             gpx.endTag(NS_CGEO, "originalCoordsEmpty");
--- a/main/src/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/cgeo/geocaching/files/GPXParser.java
@@ -310,11 +310,9 @@ abstract class GPXParser extends FileParser {
                         // there is no lookup code in gpx file
 
                         if (wptEmptyCoordinates) {
-                            waypoint.setCoords(null);
                             waypoint.setOriginalCoordsEmpty(true);
-                        } else {
-                            waypoint.setCoords(cache.getCoords());
                         }
+                        waypoint.setCoords(cache.getCoords());
 
                         waypoint.updateNoteAndUserNote(cache.getDescription());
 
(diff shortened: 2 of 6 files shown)


MagpieFourtyTwo:
Just for clarification: Coords of WPs (pre-defined as well as user-defined ones) will be overwritten unconditionally on import, right?

Cause otherwise it will a) not be possible to correct coords by importing a GPX and b) the imported cache will sometimes end up completely messed up, cause some coords will be like imported with the GPX (e. g. new ones that have been created by the import) and others will remain in an old version (e. g. the ones that already existed before importing the GPX) ... thus I would consider coords to *always* be updated on import.

murggel:
Good point, there is a merging logic to decide which dates are taken over. I will check for the cases of import from gpx and refresh from server ...

murggel:
The current bug only affects pre-defined waypoints without coordinates, for those the coordinates will be reset

MagpieFourtyTwo:
>   - data from gpx will be used to update existing coordinates in the waypoint (for user-defined and pre-defined waypoints)
>    (even if it is a pre-defined waypoint with defined coordinates by the CO and modified manually in the gpx, the data from gpx will be imported into the waypoint)
>  -  data from server will be used to update existing coordinates (for pre-defined waypoints). If coordinates were modified (through importing different coordinates from gpx or changed by owner on server), the data from server will be imported to the waypoint

Sounds good - with one exception:
Given a waypoint which already has valid coords in c:geo, this coords should never be overwritten by 0/0 coords,
no matter if they come via GPX or server - TMALSS: IMO valid coords should always win over 0/0.

murggel:
> Sounds good - with one exception:
> Given a waypoint which already has valid coords in c:geo, these coords should never be overwritten by 0/0 coords,
> no matter if they come via GPX or server - TMALSS: IMO valid coords should always win over 0/0.

0/0 will be interpreted as "no coords", hoping that there will never be a real cache at 0/0

bekuno:
The 0/0 Point is in the Gulf of Guinea, In the water. So it is hard to have it as valid coordinate.

MagpieFourtyTwo:
And even if - what's the problem with overwriting 0/0 by 0/0. ;)

murggel:
> And even if - what's the problem with overwriting 0/0 by 0/0. ;)

It will be overwritten by --/--

PR #9883, opened 2021-01-27
Introduce swipe to refresh gesture for cache details (fix #8888, fix #9880)

### LINKED ISSUE #8888: [UX] Pull-to-refresh gesture to refresh cache details
Drop-down menu in cache page is hard to use as it has many entries with only text labels and the action to refresh a cache is really common on a cache details page. 

Implement a "pull-to-refresh" gesture to refresh cache details. [It can be implemented with AndroidX](https://stackoverflow.com/a/22786827) that cgeo already uses. It would greatly improve UX, as this is one of the most (if not the most) used entries in the cache details menu, and this gesture is now common and expected (the official GC app implements it on the cache details view for example). Also it doesn't conflict with any existing gesture in the app. 

This should probably be made a separate dedicated feature request, but prepending icons for each actions of the cache details drop-down menu would greatly improve UX, this is something I think about for years now. 

**Additional context**
 If this is to be implemented, maybe we could add this pull-to-refresh gesture to search results lists as when you go back and forth between cache details and search results (or you go back to cgeo after a while), you can lose all the results not in local cache. There are maybe other views where this pattern is applicable.

### LINKED ISSUE #9880: Have "Refresh" available again via cache menu
After discussion and clarification with two beta users (2021.01.24-RC) both have somehow valid arguments to have "refresh" available in the cache menu, e.g.:

> If I read the logbook of a cache and notice, that the newest log entry is rather old I went to the menu and could refresh it right away. No I have to swipe back to the details page, refresh, wait and swipe to logbook again.

> I have activated the setting to open new caches with the last selected page which is most likely for me not the details page. So now I have either to deactivate this setting to have the possibility to easily refresh the data or I need to swipe forth or back (depending on where I start) to finally be able to refresh the information.

These are IMHO reasonable use cases (which might affect more users), so we should consider to have the refresh option back in the menu.

BTW:
Both users stated that save/delete are rather unlikely for them to be used from the menu. For those they are fine with using the icons. Their only concern was refreshing.

fm-sys:
As discussed in #8888 and #9880 this implements a pull-to-refresh gesture to refresh cache details. I've chosen a relative high threshold for pull down refresh - tests will show whether this distance is too large/small or just right. This is difficult to estimate on the emulator...

### DIFF
--- a/main/res/layout/cachedetail_activity.xml
+++ b/main/res/layout/cachedetail_activity.xml
@@ -2,21 +2,28 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="fill_parent"
-    android:layout_height="fill_parent"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
     android:background="?background_color"
     android:orientation="vertical"
-    tools:context=".CacheDetailActivity" >
+    tools:context=".CacheDetailActivity">
 
-    <androidx.viewpager.widget.ViewPager
-        android:id="@+id/viewpager"
-        android:layout_width="fill_parent"
+    <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
+        android:id="@+id/swipe_refresh"
+        android:layout_width="match_parent"
         android:layout_height="0dp"
-        android:layout_weight="1" />
+        android:layout_weight="1">
+
+            <androidx.viewpager.widget.ViewPager
+                android:id="@+id/viewpager"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent" />
+
+    </androidx.swiperefreshlayout.widget.SwipeRefreshLayout>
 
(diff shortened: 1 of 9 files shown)


fm-sys:
Codacy is failing. Anyway, I did not introduce new issues. They are just old ones, which come up again cause I edited these code parts. 

Or does anyone have an idea how to better design the functions (even if it is unrelated to the PR)?

moving-bits:
> Codacy is failing. Anyway, I did not introduce new issues. They are just old ones, which come up again cause I edited these code parts.
> 
> Or does anyone have an idea how to better design the functions (even if it is unrelated to the PR)?

On a quick glance splitting up that method into even more sub-methods, all of them doing nothing more than adding some data to the view, would not help improve readability, quite contrary. IMHO you could add a `@SuppressWarnings("PMD.NPathComplexity")` with appropriate comment at the top of the method.

fm-sys:
Only thing which we could do is extracting all the inner classes to their own files, as details activity is really way too long and it was hard to find the right places in the code. Although this does not help for `NPathComplexity`

fm-sys:
Extracting seems to be rather complex, so in case we should include it into release after all lets just add a suppress warning...

PR #9887, opened 2021-01-27
dependabot optimizing

bekuno:
try another version numbering for the ignore pattern


bekuno:
now its fine, dependabot closed the PR

Issue #9864, opened 2021-01-26
Use numbered variables for SAF strings


Lineflyer:
A contributor on Crowdin (in this case for Japanese) mentioned, that e.g. the string `contentstorage_err_create_failed` is hard to translate for him as his languages requires to use the placeholders/variables in a different sequence. This might apply also to other languages. 

Is it possible to give them a numbering (instead only `%s` -> `%1s` `%2s`)?

Edit:
Might apply for all strings which use more than one variable.

eddiemuc:
It is already possible to use numbered parameters (this is called "argument index" in Java). See here: https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html Format is e.g. like this: `%2$s` (references second parameter, format as string). Also `<` can be used to reference "next" parameter.

If no explicit argument indexes are given (like in my default translations), the formatter assumes that the variables are indexed in the order in which they are provided in the string. But other translations may use indexes as they please.

So what is requested should already be possible. Or am I misunderstanding something here?

Lineflyer:
I am not so deep into that numbered parameters, however when I do translations on Crowdin for other projects they are mostly already numbered in the source, which makes it easier to reuse them in my translated strings (without having to know how I could force them).

Example Source : "File %1$s is stored in folder %2$s"
Translation (weird example): "Im Ordner %2$s ist die Datei %1$s gespeichert"

bekuno:
Not using number values should be already a lint error.

To make it easy for you in strings.xml you can always use the numbered form, also when only one parameter is used.

bekuno:
> Might apply for all strings which use more than one variable.

change to: Must apply ...

fm-sys:
> > Might apply for all strings which use more than one variable.
> 
> change to: Must apply ...

On the other hand changing all strings would lead to high retranslation effort. But we can definitely introduce this rule for all new strings

bekuno:
> On the other hand changing all strings would lead to high retranslation effort. But we can definitely introduce this rule for all new strings

To clarify: A rule for more than 1 entry is already active. Here it is a error not to use positional parameters. So only some strings from the SAF PR should be affected.
Changing this for all strings is of cause no good idea. But it is a possibility for new/changed strings.

Lineflyer:
> A rule for more than 1 entry is already active. Here it is a error not to use positional parameters. So only some strings from the SAF PR should be affected.

But we have strings with more than one `%s` in it. Why did the lint check not fail in this case?

[...]

PR #9889, opened 2021-01-27
Fix #9885 - Use folder instead of directory

### LINKED ISSUE #9885: Unify wording: Folder or directory?
This is just a mnior one (alerted by some Crowdin users):
With SAF we recently introduced a couple of new strings which use (mostly) `folder` instead of `directory`. In the existing code before SAF we mostly used `directory` but IMHO also `folder` at one place.

I think we should use a unified wording here as there is no difference, or is there any?

From modern apps and user manuals I do much more often nowadays read `folder` (deutsch: Ordner), so I would prefer that.
Opinions?

Lineflyer:
Additionally one string shortening as requested in https://crowdin.com/translate/cgeo/162/en-pt#24300

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -455,7 +455,7 @@
     <string name="about_system_write_logcat_type">Create standard or extended logfile?\n\nStandard logfiles will be fine for most use cases.\nUse extended logfiles only when asked to do so.</string>
     <string name="about_system_write_logcat_type_standard">Standard</string>
     <string name="about_system_write_logcat_type_extended">Extended</string>
-    <string name="about_system_write_logcat_success">Logfile written to \"%1$s\" in \"%2$s\" dir</string>
+    <string name="about_system_write_logcat_success">Logfile \"%1$s\" written to folder \"%2$s\"</string>
     <string name="about_system_write_logcat_error">Error writing logfile</string>
     <string name="changelog_github">List of all changes</string>
     <string name="about_nutshellmanual">Online manual</string>
@@ -486,7 +486,7 @@
     <string name="settings_title_cachedetails">Cache Details</string>
     <string name="settings_summary_cachedetails">Customize the display of cache information.</string>
     <string name="settings_title_offlinedata">Offline Data</string>
-    <string name="settings_summary_offlinedata">Configure the data to be saved for offline use, the storage location and directories for GPX files.</string>
+    <string name="settings_summary_offlinedata">Configure the data to be saved for offline use, the storage location and folders for GPX files.</string>
     <string name="settings_title_logging">Logging</string>
     <string name="settings_summary_logging">Configure your signature and set logging options.</string>
     <string name="settings_title_map">Map</string>
@@ -498,12 +498,12 @@
     <string name="settings_title_map_behavior">Map behavior</string>
     <string name="settings_title_gpx">GPX</string>
 
-    <string name="settings_title_data_dir">Geocache data directory</string>
-    <string name="settings_title_data_dir_usage">Geocache data directory (%1$s used)</string>
+    <string name="settings_title_data_dir">Geocache data folder</string>
+    <string name="settings_title_data_dir_usage">Geocache data folder (%1$s used)</string>
     <string name="calculate_dataDir_title">Geocache data usage</string>
     <string name="calculate_dataDir">Calculating Geocache data disk usage</string>
     <string name="settings_data_dir_item">%1$s (%2$s free)</string>
-    <string name="confirm_data_dir_move_title">Move geocache data directory?</string>
+    <string name="confirm_data_dir_move_title">Move geocache data folder?</string>
     <string name="confirm_data_dir_move">Moving the geocache data might take a really long time, depending on your device and the amount of data.</string>
 
     <string name="settings_title_basicmembers">Options for Basic Members</string>
@@ -662,9 +662,9 @@
     <string name="init_backup_success">c:geo\'s database was successfully copied to:</string>
     <string name="init_backup_failed">Backup of c:geo\'s database failed.</string>
     <string name="init_backup_unnecessary">Database is empty, no backup necessary.</string>
-    <string name="init_backup_folder_exists_error">Couldn\'t create a new backup directory. Have you already made a backup in the last minute?</string>
+    <string name="init_backup_folder_exists_error">Unable to create a new backup folder. Did you already perform a backup in the last minute?</string>
     <string name="init_backup_no_backup_available">There is no backup file available.</string>
-    <string name="backup_confirm_overwrite">This will delete your oldest existing backup directory from %s. \nDo you want to continue?</string>
+    <string name="backup_confirm_overwrite">This will delete your oldest existing backup folder from %s. \nDo you want to continue?</string>
     <string name="restore_confirm_overwrite_database">WARNING: This will overwrite %s on your device!</string>
     <string name="restore_confirm_overwrite_database_and_settings">WARNING: This will overwrite %s and your current settings on your device!</string>
     <string name="restore_confirm_overwrite_settings">WARNING: This will overwrite your current settings on your device!</string>
@@ -691,15 +691,15 @@
     <string name="settings_restart">c:geo needs to be restarted now to bring loaded settings into effect</string>
 
     <string name="settings_info_offline_maps_title">Info on Offline Maps</string>
-    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can download maps from Mapsforge or even create your own maps from OSM data. You need to select the offline map directory first to get offline map support.</string>
+    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can obtain maps from different providers or even create your own maps (from OSM data). First you have to select the folder in which offline maps are to be stored.</string>
     <string name="settings_info_themes_title">Info on Map Themes</string>
     <string name="settings_info_themes">c:geo supports custom themes for offline maps. These can be used to change the color style of the map (e.g. to have a nightview map) or to highlight certain objects like cycle paths or height lines within the map.</string>
     <string name="init_mapsource_select">Select Map Source</string>
     <string name="init_local_file_system">Local File System</string>
-    <string name="init_base_directory_description">Base Directory for public local data</string>
-    <string name="init_map_directory_description">Directory with offline maps</string>
-    <string name="init_gpx_importexportdir">GPX Directory</string>
-    <string name="init_dataDir">Select data directory</string>
+    <string name="init_base_directory_description">Base folder for public local data</string>
+    <string name="init_map_directory_description">Folder with offline maps</string>
+    <string name="init_gpx_importexportdir">GPX Folder</string>
+    <string name="init_dataDir">Select data folder</string>
     <string name="init_dataDir_note">You may choose to store the additional data for geocaches (spoilers, log images, …) on your external storage (emulated or real external SD card depending on your device).</string>
     <string name="init_maptrail">Show history track</string>
     <string name="init_maptrail_length">History track max. length</string>
@@ -747,23 +747,23 @@
     <string name="init_debug_note">c:geo can generate a lot of debugging information. Activating this setting will affect the stability of c:geo and is only recommended if you are asked to send a log to the developers. Please keep debug logging disabled during normal usage of c:geo!</string>
     <string name="init_debug">Activate debug log</string>
     <string name="init_dbonsdcard_title">Database location</string>
-    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or in a directory on your user storage. Having it on the user storage may free some memory, but you may lose a bit of performance and c:geo will not work if the storage is not accessible or you remove the database file.</string>
+    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or into a folder on your user storage. We recommend to only activate this option if you need regular manual access to your database file.</string>
     <string name="init_dbonsdcard">On user storage</string>
     <string name="init_dbmove_dbmove">Moving Database</string>
     <string name="init_dbmove_running">Moving Database</string>
     <string name="init_dbmove_success">Successfully moved the database.</string>
     <string name="init_dbmove_failed">Failed to move database</string>
-    <string name="init_datadirmove_datadirmove">Moving Geocache data directory</string>
-    <string name="init_datadirmove_running">Moving Geocache data directory</string>
-    <string name="init_datadirmove_success">Successfully moved Geocache data directory.</string>
-    <string name="init_datadirmove_failed">Moving the Geocache data directory partly or completely failed. Please check old and new directories and move by hand if necessary.</string>
+    <string name="init_datadirmove_datadirmove">Moving Geocache data folder</string>
+    <string name="init_datadirmove_running">Moving Geocache data folder</string>
+    <string name="init_datadirmove_success">Successfully moved Geocache data folder.</string>
+    <string name="init_datadirmove_failed">Moving the Geocache data folder partly or completely failed. Please check old and new folders and move missing data by hand.</string>
     <string name="init_plain_logs">Plain Logs</string>
     <string name="init_summary_plain_logs">Display log entries without colors</string>
     <string name="init_use_native_ua">Android browser</string>
     <string name="init_summary_use_native_ua">Identify as Android browser. Solves login problems when using certain network providers.</string>
-    <string name="init_rendertheme_folder">Map Themes Directory</string>
+    <string name="init_rendertheme_folder">Map Themes Folder</string>
     <string name="init_maintenance">Maintenance</string>
-    <string name="init_maintenance_directories_note">c:geo stores images, log images and other files related to a cache in a separate directory. In some cases (like importing/exporting the database) this directory may contain outdated files, which can be deleted here.</string>
+    <string name="init_maintenance_directories_note">c:geo stores images and other files related to a cache in a dedicated folder. In some cases (e.g. when importing/exporting the database) this folder may contain obsolete files that can be deleted here.</string>
     <string name="init_maintenance_directories">Delete orphaned files</string>
     <string name="init_location">Geolocation</string>
     <string name="init_location_note">On devices equipped with Google Play Services, c:geo can automatically use a better geolocation provider. However, this prevents the use of an external BlueTooth GPS receiver.</string>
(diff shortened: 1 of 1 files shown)


main/res/values/strings.xml
```diff
@@ -747,23 +747,23 @@
...
     <string name="init_plain_logs">Plain Logs</string>
     <string name="init_summary_plain_logs">Display log entries without colors</string>
     <string name="init_use_native_ua">Android browser</string>
     <string name="init_summary_use_native_ua">Identify as Android browser. Solves login problems when using certain network providers.</string>
-    <string name="init_rendertheme_folder">Map Themes Directory</string>
+    <string name="init_rendertheme_folder">Map Themes Folder</string>
     <string name="init_maintenance">Maintenance</string>
-    <string name="init_maintenance_directories_note">c:geo stores images, log images and other files related to a cache in a separate directory. In some cases (like importing/exporting the database) this directory may contain outdated files, which can be deleted here.</string>
+    <string name="init_maintenance_directories_note">c:geo stores images, log images and other files related to a cache in a separate folder. In some cases (like importing/exporting the database) this folder may contain outdated files, which can be deleted here.</string>
```
bekuno, line 766: How about: `c:geo stores images and other files related to a cache in a separate folder. In some cases (e.g. when importing/exporting the database) this folder may contain obsolete files that can be deleted here.`

main/res/values/strings.xml
```diff
@@ -662,9 +662,9 @@
     <string name="init_backup_success">c:geo\'s database was successfully copied to:</string>
     <string name="init_backup_failed">Backup of c:geo\'s database failed.</string>
     <string name="init_backup_unnecessary">Database is empty, no backup necessary.</string>
-    <string name="init_backup_folder_exists_error">Couldn\'t create a new backup directory. Have you already made a backup in the last minute?</string>
+    <string name="init_backup_folder_exists_error">Couldn\'t create a new backup folder. Have you already made a backup in the last minute?</string>
```
bekuno, line 665: How about `No new backup folder could be created. Have you already created a backup within the last minute?`

main/res/values/strings.xml
```diff
@@ -691,15 +691,15 @@
     <string name="settings_restart">c:geo needs to be restarted now to bring loaded settings into effect</string>
 
     <string name="settings_info_offline_maps_title">Info on Offline Maps</string>
-    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can download maps from Mapsforge or even create your own maps from OSM data. You need to select the offline map directory first to get offline map support.</string>
+    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can download maps from Mapsforge or even create your own maps from OSM data. You need to select the offline map folder first to get offline map support.</string>
```
bekuno, line 694: Download from Mapsforge?
How about: `c:geo supports maps for offline use. You can obtain maps from different providers or even create your own maps (from OSM data). First you have to select the offline map folder in which the offline maps are to be stored.`

main/res/values/strings.xml
```diff
@@ -747,23 +747,23 @@
     <string name="init_debug_note">c:geo can generate a lot of debugging information. Activating this setting will affect the stability of c:geo and is only recommended if you are asked to send a log to the developers. Please keep debug logging disabled during normal usage of c:geo!</string>
     <string name="init_debug">Activate debug log</string>
     <string name="init_dbonsdcard_title">Database location</string>
-    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or in a directory on your user storage. Having it on the user storage may free some memory, but you may lose a bit of performance and c:geo will not work if the storage is not accessible or you remove the database file.</string>
+    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or in a folder on your user storage. Having it on the user storage may free some memory, but you may lose a bit of performance and c:geo will not work if the storage is not accessible or you remove the database file.</string>
```
bekuno, line 750: Maybe a better wording would be necessary here. I really do not understand the text.

bekuno (changes requested):
I suggested some extended changes for a better wording. I think it is better to optimize the selected texts complete instead to only exchange directory by folder and later do a second change with the complete text.

I would suggest also to unify the affected keywords within this PR for the same reason as before. 
@moving-bits @eddimuc What do you mean?

Lineflyer:
@bekuno 
Thanks for your feedback, will add some comments and lets wait for @moving-bits and @eddiemuc opinion.

main/res/values/strings.xml
```diff
@@ -2029,32 +2029,32 @@
...
+    <string name="downloadmap_info">Choose a map file for your region by tapping the \"Save\" icon.\n\nDownloading will run in the background. When downloading has completed, the downloaded file will be copied to c:geo\'s map folder and set as current map for c:geo.\n\nBe careful: Downloading a map can cause high network traffic!</string>
     <string name="downloadmap_choose">Choose map</string>
     <string name="downloadmap_filename">Downloading %s</string>
     <string name="downloadmanager_not_available">Map file cannot be downloaded, system download manager not available</string>
     <string name="download_started">Downloading started in background</string>
-    <string name="downloadmap_target_not_writable">Target \"%s\" is not writable - downloaded map file will be stored in system download directory and has to be moved manually to the map directory.</string>
+    <string name="downloadmap_target_not_writable">Target \"%s\" is not writable - downloaded map file will be stored in system download folder and has to be moved manually to the map folder.</string>
```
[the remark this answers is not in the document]
Lineflyer: LGTM.

main/res/values/strings.xml
```diff
@@ -2029,32 +2029,32 @@
...
     <string name="receivemapfile_error_filenotfound_exception">File not found error while receiving map file.</string>
     <string name="receivemapfile_kb_copied">%d kB copied</string>
     <string name="receivemapfile_cancelled">Map file transfer cancelled</string>
 
     <!-- Download map preference and menu entry -->
     <string name="downloadmap_title">Download offline map</string>
-    <string name="downloadmap_info">Choose a map file for your region by tapping the \"Save\" icon.\n\nDownloading will run in the background. When downloading has completed, the downloaded file will be copied to c:geo\'s map directory and set as current map for c:geo.\n\nBe careful: Downloading a map can cause high network traffic!</string>
+    <string name="downloadmap_info">Choose a map file for your region by tapping the \"Save\" icon.\n\nDownloading will run in the background. When downloading has completed, the downloaded file will be copied to c:geo\'s map folder and set as current map for c:geo.\n\nBe careful: Downloading a map can cause high network traffic!</string>
```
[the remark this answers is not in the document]
Lineflyer: LGTM

main/res/values/strings.xml
```diff
@@ -851,7 +851,7 @@
     <string name="contentstorage_selectfolder_dialog_choice_copy">Copy all content</string>
     <string name="contentstorage_selectfolder_dialog_choice_move">Move all content</string>
     <string name="contentstorage_selectfolder_dialog_copy_move_finished_title">Folder %s: Copy/Move done</string>
-    <string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html"><![CDATA[<p>The copy/move operation had the result: %s. %d file(s) and %d subdirectories were copied/moved.</p><p>Tap \'ok\' finalize the directory change.</p>]]></string>
+    <string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html"><![CDATA[<p>The copy/move operation had the result: %s. %d file(s) and %d subfolders were copied/moved.</p><p>Tap \'ok\' finalize the folder change.</p>]]></string>
```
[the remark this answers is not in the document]
Lineflyer: LGTM

main/res/values/strings.xml
Lineflyer: You are right, this strings is quite old.

main/res/values/strings.xml
Lineflyer: That was IMHO targeted for older Android versions, where system internal and (internal) user storage had different quota or one of them could even be an SD-card. Nowadays on more recent Android version it makes no difference most of the time, but there might be special setups out there on some devices.
Therefore the text was explicitly made rather unspecific.

bekuno: The key sounds "db-on-sdcard" - is with `user storage` the storage on a additional sdcard meant?

main/res/values/strings.xml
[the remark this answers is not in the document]
eddiemuc: Small remark: the text "ok to complete the change" may lead users to assume that something is happening aftetr clicking ok (even that the copy/move is somehow reverted if they don't click ok). But this is not the case. At the point this is displayed the copy/move is already done, no way to revert it.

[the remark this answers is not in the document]
fm-sys: grafik

> At the point this is displayed the copy/move is already done, no way to revert it.

Then there shouldn't be a cancel option at all, and we can leave out `Tap 'ok' finalize the folder change.` completely...

main/res/values/strings.xml
fm-sys: In android 1-3 there was no user storage directly on the device, just the private app-specific one. Later on, internal memory was added as an emulated SD card. So probably it has historical reasons... 

Nowadays you can choose between internal (emulated) memory and SD card (I stumbled across this setting a few days ago). The setting is useful, if you have a large db and prefer to use the SD card rather than the device memory

main/res/values/strings.xml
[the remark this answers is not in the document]
bekuno: Then the best would be to let the string unchanged here and make a complete PR with all changes.
IMHO we should also omit the special formatting strings (`<![CDATA[<p>`etc.).
A line brake after the first sentense would make it also more readable.
Also see my change suggestion.

eddiemuc:
Am I guessing correctly that you also fix #9864 with this PR?

main/res/values/strings.xml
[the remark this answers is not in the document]
eddiemuc: I added PR #9894 to remove the "cancel" option in above dialog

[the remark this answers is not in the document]
moving-bits: If we use a noun depending on an inserted number (here: "xx file(s)" and "xx folder(s)"), we need to add quantity strings and initialize the string accordingly. As we have two such nouns in this text, we probably need two separate strings.

main/res/values/strings.xml
moving-bits: How about `Unable to create the new backup folder. Have you ...`?

moving-bits (changes requested):
@Lineflyer 
Have added my feedback, either "thumbs up" on suggested translations or a text with proposed changes.

eddiemuc (approved):

main/res/values/strings.xml
```diff
@@ -455,7 +455,7 @@
     <string name="about_system_write_logcat_type">Create standard or extended logfile?\n\nStandard logfiles will be fine for most use cases.\nUse extended logfiles only when asked to do so.</string>
     <string name="about_system_write_logcat_type_standard">Standard</string>
     <string name="about_system_write_logcat_type_extended">Extended</string>
-    <string name="about_system_write_logcat_success">Logfile written to \"%1$s\" in \"%2$s\" dir</string>
+    <string name="about_system_write_logcat_success">Logfile written to \"%1$s\" in \"%2$s\" folder</string>
```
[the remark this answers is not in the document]
Lineflyer: done, comes in with next pr

main/res/values/strings.xml
Lineflyer: I simplified this option. AFAIK its nowadays only relevant if you want to have the possibility to access the DB by yourself.

main/res/values/strings.xml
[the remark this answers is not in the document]
Lineflyer: I left this untouched but only replaced dir by folder. The removal of [CDATA] should be done in a dedicate PR.

Lineflyer:
@eddiemuc 

> Am I guessing correctly that you also fix #9864 with this PR?

No, as I am not so sure about the mechanism behind someone of you should do this.

Also the removal of `[CDATA]` should be done seperately.

main/res/values/strings.xml
```diff
@@ -691,15 +691,15 @@
     <string name="settings_restart">c:geo needs to be restarted now to bring loaded settings into effect</string>
 
     <string name="settings_info_offline_maps_title">Info on Offline Maps</string>
-    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can download maps from Mapsforge or even create your own maps from OSM data. You need to select the offline map directory first to get offline map support.</string>
+    <string name="settings_info_offline_maps">c:geo supports maps for offline use. You can obtain maps from different providers or even create your own maps (from OSM data). First you have to select the offline map folder in which the offline maps are to be stored.</string>
```
bekuno, line 694: Double `offline maps`.
How about: ... First you have to select the folder in which the offline maps are to be stored.

main/res/values/strings.xml
```diff
@@ -851,7 +851,7 @@
     <string name="contentstorage_selectfolder_dialog_choice_copy">Copy all content</string>
     <string name="contentstorage_selectfolder_dialog_choice_move">Move all content</string>
     <string name="contentstorage_selectfolder_dialog_copy_move_finished_title">Folder %s: Copy/Move done</string>
-    <string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html"><![CDATA[<p>The copy/move operation had the result: %s. %d file(s) and %d subdirectories were copied/moved.</p><p>Tap \'ok\' finalize the directory change.</p>]]></string>
+    <string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html"><![CDATA[<p>The copy/move operation had the result: %s. %d file(s) and %d folders were copied/moved.</p><p>Tap \'ok\' finalize the folder change.</p>]]></string>
```
bekuno, line 854: Please use positional parameter.

main/res/values/strings.xml
```diff
@@ -747,23 +747,23 @@
     <string name="init_debug_note">c:geo can generate a lot of debugging information. Activating this setting will affect the stability of c:geo and is only recommended if you are asked to send a log to the developers. Please keep debug logging disabled during normal usage of c:geo!</string>
     <string name="init_debug">Activate debug log</string>
     <string name="init_dbonsdcard_title">Database location</string>
-    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or in a directory on your user storage. Having it on the user storage may free some memory, but you may lose a bit of performance and c:geo will not work if the storage is not accessible or you remove the database file.</string>
+    <string name="init_dbonsdcard_note">You may store the database of c:geo either system internally or into a folder on your user storage. We recommend to only activate this option if you need user access to your database file.</string>
```
moving-bits, line 750: IMHO the rationale for using user storage for the database is more something like "regular access" or "frequent access" to the database. If you only need it once in a while you are better of by doing a backup.

main/res/values/strings.xml
```diff
@@ -747,23 +747,23 @@
...
     <string name="init_dbmove_success">Successfully moved the database.</string>
     <string name="init_dbmove_failed">Failed to move database</string>
-    <string name="init_datadirmove_datadirmove">Moving Geocache data directory</string>
-    <string name="init_datadirmove_running">Moving Geocache data directory</string>
-    <string name="init_datadirmove_success">Successfully moved Geocache data directory.</string>
-    <string name="init_datadirmove_failed">Moving the Geocache data directory partly or completely failed. Please check old and new directories and move by hand if necessary.</string>
+    <string name="init_datadirmove_datadirmove">Moving Geocache data folder</string>
+    <string name="init_datadirmove_running">Moving Geocache data folder</string>
+    <string name="init_datadirmove_success">Successfully moved Geocache data folder.</string>
+    <string name="init_datadirmove_failed">Moving the Geocache data folder partly or completely failed. Please check old and new folders and move by hand if necessary.</string>
```
moving-bits, line 759: ... "and move missing files by hand" ...
or: "missing data" (to include both files and folders)

main/res/values/strings.xml
moving-bits: And we need quantity strings here, don't we?

moving-bits: And no reason for \<![CDATA[...]]>

moving-bits (changes requested):

main/res/values/strings.xml
Lineflyer: Done

main/res/values/strings.xml
Lineflyer: rephrased to "regular manual access"

main/res/values/strings.xml
Lineflyer: Done

main/res/values/strings.xml
Lineflyer: Someone else needs to handle this as I do not know, what needs to be done for these changes.

Lineflyer:
All remarks adressed now.
Only the quantity strings and [CDATA] replacement is something I don't know how it works.


Either you provide the sniplet to use or we handle it in a dedicated PR by one of you.

main/res/values/strings.xml
moving-bits: Let's start with removing the CDATA part and positional parameters:

`<string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html">The copy/move operation had the result: %1$s. Affected files: %2$d, affected folders: %3$d.\n\nTap \'ok\' finalize the folder change.</string>`

Maybe we can avoid quantity string (actually called `plurals`) completely here by using that text.

Or maybe:

`<string name="contentstorage_selectfolder_dialog_copy_move_finished_msg_html">Result: %1$s\nAffected files: %2$d\nAffected folders: %3$d\n\nTap \'ok\' finalize the folder change.</string>`

eddiemuc: Can we do this in a separate PR? Code changes are needed in conjunction with CDATA removal. I am willing to do this, but let's please first merge and finish this PR

eddiemuc: (reason: code currently expects HTML to be returned in this string, as the key suffix "_html" suggests)

eddiemuc:
I can do this in a dedicated PR if wanted. But only for saf stuff, not for map attribution (have no solution for you there). 
Also, be aware that I will change key names since suffix "_html" is no longer fitting

main/res/values/strings.xml
bekuno: I think we need a separate change (issue).
 - Text optimization - `finalize` sounds strange, as we can not abort anymore
 - Text optimization - no html formatting within translation string
 - keyword optimization - no `html` there
 - support for plurals
 
So I would only make the changes for the renaming to folder and make lint happy with positional parameters.

fm-sys: As we have removed the cancel option meanwhile in the corresponding dialog, we can IMHO leave out `\n\nTap \'ok\' finalize the folder change.` completely. In other dialogues we don't explain that you have to press OK either and especially it gets confusing if there are two different translations for OK in one language 😉

moving-bits:
> Only the quantity strings and [CDATA] replacement is something I don't know how it works.
> Either you provide the sniplet to use or we handle it in a dedicated PR by one of you.

For `contentstorage_selectfolder_dialog_copy_move_finished_msg_html` I've left a proposal in https://github.com/cgeo/cgeo/pull/9889#discussion_r567449204

main/res/values/strings.xml
eddiemuc: Don't bother too much over these particular lines. They need to change anyway soon in the context of saf ux redesign

Lineflyer:
Feel free to either merge or takeover into another PR or how you like :)
But as we have quite some PR outstanding dealing with strings, we should either merge soon or dismiss.

moving-bits:
> Feel free to either merge or takeover into another PR or how you like :)
> But as we have quite some PR outstanding dealing with strings, we should either merge soon or dismiss.

Let's merge it to get Crowdin running, and do open things separately.

Lineflyer:
FYI for german team members:
I used the opportunity to replace the (often asked on support term) "Karten-Thema" (for map theme) to "Kartendarstellung". Hope you are fine with that choice.

fm-sys:
> FYI for german team members:
> I used the opportunity to replace the (often asked on support) term "Karten-Thema" (for map theme) to "Kartendarstellung" on Crowdin transaltion. Hope you are fine with that choice.

Way better. "Kartendesign" would have been good as well...

PR #9894, opened 2021-01-29
rel to #9889: remove cancel option in dialog after copy move

eddiemuc:
rel to #9889: remove cancel option in dialog after copy move


moving-bits:
retest this please

PR #9906, opened 2021-01-30
Integrate SAF base folder query into wizard

moving-bits:
@eddiemuc 
Works in general, but I'm facing two issues, therefore setting "WIP" label for now:
- In `InstallWizardActivity.requestBaseFolder()` I've tried calling `selectPersistableFolder()` with a callback. But the callback got called only on selecting "cancel", not on "ok". (So I changed the code to work without callback, but it's sort of a hack.)
- This may be related to my second, more important issue: Even if I select a folder and grant access to it, my selection does not get persisted. In wizard's summary the entry is still shown as error, and in the `MainActivity` the base folder check will open.
- Strange thing: If I select the very same base folder coming from `MainActivity` it get's persisted. So what is the difference?

@eddiemuc
Do you have an idea what I'm missing here?

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -842,7 +842,7 @@
 
     <!-- Content Storage -->
     <string name="contentstorage_grantaccess_dialog_title">Grant access for data storage</string>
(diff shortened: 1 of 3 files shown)


moving-bits:
I think I've found it - need to integrate `ContentStorageActivityHelper.onActivityResult()`

moving-bits:
Requesting a SAF base folder is integrated as third part of permission requests and will look like this:

image

I calls the regular SAF folder selection dialog for now. As already discussed, that dialog should be streamlined a bit for the installation / first migration scenario to make it easier to understand.

eddiemuc:
Sorry I did not see your request until now. But you found the solution anyway.

Btw I integrated the content storage activity helper already in the base cgei activity class and it's inactivityresult method. Are you using another base class?

moving-bits:
> Sorry I did not see your request until now.

No problem

> Btw I integrated the content storage activity helper already in the base cgei activity class and it's inactivityresult method. Are you using another base class?

I'm using AppCompatActivity here. Need to check if it would work with AbstractActivity as well.

PR #9912, opened 2021-01-30
fix #9809: redesign format of uri display to users

### LINKED ISSUE #9809: Investigate showing the full path to base folder
While setting up the "Signal" messenger app I stumbled upon the fact, that it is somehow possible to show the absolute path to a folder selected via SAF.
I set up the backup directory in that app via SAF and it shows the complete path now in the settings. In my case: `Lars 512GB/AppBackup/Signal Messenger`. 

So it seems it is somehow possible for an app to know the absolute path (or did I overlook some requirement for that, which c:geo currently does not have)?
It would IMHO be very helpful for our users to see that as well instead of only the prefixed `...`.

eddiemuc:
fix #9809: redesign format of uri display to users

The Leanings from app Signal were incorporated in this one

### DIFF
--- a/main/src/cgeo/geocaching/utils/UriUtils.java
+++ b/main/src/cgeo/geocaching/utils/UriUtils.java
@@ -1,21 +1,38 @@
 package cgeo.geocaching.utils;
 
+import cgeo.geocaching.CgeoApplication;
+
+import android.content.Context;
 import android.content.UriPermission;
 import android.net.Uri;
+import android.os.Build;
+import android.os.storage.StorageManager;
+import android.os.storage.StorageVolume;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.content.ContextCompat;
 
 import java.io.File;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
+
 
 
 /**
  * Utility class prvoding helper methods when dealing with Uris
  */
 public final class UriUtils {
+
+    //constants for a unit test volume entry. VolumeId must be one which does NOT OCCUR in real life!
+    //(volumneIds in real life are typically of form xxxx-xxxx where x are hexadecimal digits e.g. 11F2-3A16)
+    public static final String UNITTEST_VOLUME_ID = "TEST-1111-2222";
+    public static final String UNITTEST_VOLUME_NAME = "TEST-SDCARD";
+
+    private static final Map<String, String> VOLUME_MAP = getVolumeMap();
 
     private UriUtils() {
         //no instance wanted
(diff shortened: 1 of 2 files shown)


eddiemuc:
See class UriUtilsTest (part of this PR) for many examples of how Uris will be formatted after this PR is in place

tests/src/cgeo/geocaching/utils/UriUtilsTest.java
```diff
@@ -12,27 +12,37 @@
...
+    //real-world Uri examples and their wanted display
+    //Each test case has entries "Description", "expected display Uri", "raw Uri
+    public static final String[][] URI_DISPLAY_TESTS = new String[][]{
+        new String[]{"Internal /cgeo", "/cgeo", "content://com.android.externalstorage.documents/tree/primary%3Acgeo"},
+        new String[]{"Internal /cgeo/logfiles", "/cgeo/logfiles", "content://com.android.externalstorage.documents/tree/primary%3Acgeo%2Flogfiles"},
+        new String[]{"Internal file in /cgeo/logfiles", "/cgeo/logfiles/log.txt", "content://com.android.externalstorage.documents/tree/primary%3Acgeo%2Flogfiles%2Flog.txt"},
+        new String[]{"SDCARD /cgeo with known volume id", "TEST-SDCARD/cgeo", "content://com.android.externalstorage.documents/tree/TEST-1111-2222%3Acgeo"},
+        new String[]{"SDCARD /Music/cgeo with unknown volume id", "TEST-UNKNOWNID/Music/cgeo", "content://com.android.externalstorage.documents/tree/TEST-UNKNOWNID%3AMusic%2Fcgeo"},
+        new String[]{"SDCARD /cgeo with unknown volume id", "TEST-UNKNOWNID/cgeo", "content://com.android.externalstorage.documents/tree/TEST-UNKNOWNID%3Acgeo"},
```
fm-sys, line 26: Should we really display "TEST-UNKNOWNID"? Wouldn't `...` be better in this case?

main/src/cgeo/geocaching/utils/UriUtils.java
```diff
@@ -25,27 +42,131 @@ private UriUtils() {
...
+//                break;
+//            }
+//        }
+//
+//        if (storageVolume == null) {
+//            return backupName;
+//        } else {
+//            return context.getString(R.string.StorageUtil__s_s, storageVolume.getDescription(context), backupName);
+//        }
+//    }
```
fm-sys, line 167: I guess this method is a unnecessary leftover... Am I right?

tests/src/cgeo/geocaching/utils/UriUtilsTest.java
eddiemuc: This is just like this in the test case. You may see that the same id appears also in the Uri. The test cases tests that whatever is in this place in the Uri (usually a UUID of a volume) is displayed there. And this only happens if c:geo cannot find the volumne name in the volume manager.

main/src/cgeo/geocaching/utils/UriUtils.java
eddiemuc: indeed. I will delete it

tests/src/cgeo/geocaching/utils/UriUtilsTest.java
fm-sys: Will it always be "TEST-UNKNOWNID" literally or is there another way to detect such cases? Then this scenario should be handled. However, I fear that it is not so easy...

eddiemuc: Again, this is for unit test only. In real life there will for sure not be a volume uuid in such a strange format. The real world uuids I saw had the Format xxxx-xxxx, where each x is a hexadecimal digit. Example: 1AD3-34FE. But of course we can't be sure what can happen out there in the real world.

Also, this is will only be shown to the user if Androids VolumeManager does not know the name for this volume. In the most common case such volume s name seems to be SDCARD.

fm-sys: Sorry, now I got it. Thanks for clarification...

Issue #9919, opened 2021-01-31
Support downloading map theme files


moving-bits:
Reminder to myself:

Extend MapDownloader to automatically offer to download "Elevate" theme file when downloading a map file from OpenAndroMaps source.

Also add update feature (if possible).

moving-bits:
Maybe we should wait for implementation of "zipped theme support" in Mapsforge to spare unpacking and cleanup steps.

eddiemuc:
> Maybe we should wait for implementation of "zipped theme support" in Mapsforge to spare unpacking and cleanup steps.

Good idea :-)

moving-bits:
BTW: Can't we temporarily base our implementation on a specific "nightly" build of Mapsforge (the first which includes the zipped theme support), and switch back to official release as soon as 0.16 is released?

eddiemuc:
> BTW: Can't we temporarily base our implementation on a specific "nightly" build of Mapsforge (the first which includes the zipped theme support), and switch back to official release as soon as 0.16 is released?

We integrate mapsforge as libraries via gradle. I assume the are downloaded from some gradle server (something like maven central). This can only grab library versions previously uploaded by mapsforge.

I assume to use some sort of "nightly" we would have to download and compile maosforge sources ourself which seems a bit much imho. 

I can ask mapsforge if they have some sort of nightly build and whether they are willing to make this publically available

eddiemuc:
Update: I posted a question to mapsforge

moving-bits:
> I assume to use some sort of "nightly" we would have to download and compile maosforge sources ourself which seems a bit much imho.

agreed

> I can ask mapsforge if they have some sort of nightly build and whether they are willing to make this publically available

Would be kind, thanks. Or maybe when they plan to release 0.16? I know their normal release cycle is more like "half a year", but maybe...

Lineflyer:
I always thought we are using our own fork https://github.com/cgeo/mapsforge in c:geo, or do we just use a tag to fetch from their repo while building c:geo.
Sorry for the unqualified question.

bekuno:
> I assume to use some sort of "nightly" we would have to download and compile mapsforge sources ourself which seems a bit much imho.

I do not see a nightly release.
But  - we can reference to a commit ID by using `jitpack.io` as additional ressource.
I will prepare a PR.

bekuno:
[...]

PR #9923, opened 2021-01-31
use dedicated commit for the mapsforge library

bekuno:
As needed by #9919:

To use a special commit instead a release it is possible using jitpack.org.
First you have to integrate jitpack as maven source.
On jitpack you must select your commit.
Go to https://jitpack.io/#mapsforge/mapsforge, select commit, an "Get it" from th commit list.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -130,6 +130,10 @@ subprojects {
         }
         // Android Support libraries are now in an online Maven repository
         google()
+        // integrate as last one - for versions by commit ID
+        maven {
+            url 'https://jitpack.io'
+        }
     }
 
     // disable warning of annotation processor about annotation types with no registered processor (like @NonNull)
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -307,12 +307,19 @@ dependencies {
     implementation "com.asamm:locus-api-android:0.9.23"
 
     // Mapsforge new version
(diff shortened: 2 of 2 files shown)


moving-bits:
@bekuno 
Thanks for preparing this configuration. Not sure, but don't we need the (currently) latest commit cdccd83a18, as according to the change log this is the commit that includes the ZIP render theme.

bekuno:
This is the commit after the needed one with the new release notes.

moving-bits:
@bekuno 
Need your help with this one:
I'm trying to access the `ZipRenderTheme` class, which is part of the mapsforge version included by this PR, but AS does not detect it. And if I select any other mapsforge class, AS will point me into the `mapsforge-map-0.15.0.jar`, although `build.gradle` clearly points to the specific commit. What do I have to do to be able to access the new classes introduced by that PR?

bekuno:
I tested the access of one of the mapsforge classes. It looks in AS like 
grafik
Then you should be able to access the needed class ....

moving-bits:
Thanks for cross-checking. I did it exactly this way, but AS neither did list any of the new classes like `ZipRenderTheme`, and only showed `mapsforge-map-0.15.0.jar` in the top left corner, although I have been on master branch. - But today our "special commit jar" is shown. Spooky :-) (Or maybe some caching problem.)

PR #9926, opened 2021-01-31
fix #9808: remove outdated basedir check dialog (replaced by wizard)

### LINKED ISSUE #9808: Integrate base folder selection into configuration wizard
This issue is a reminder, that after SAF has been merged to `master`, the selection of a base folder should probably be integrated into the "first startup" wizard.

eddiemuc:
fixes #9808
Removes outdated saf basedir check fialog on startup:
image

Because it has be replaced by wizard dialog:
image


moving-bits:
AFAICS we should keep this check, see https://github.com/cgeo/cgeo/issues/9808#issuecomment-770427892

Issue #9872, opened 2021-01-26
SAF migration: Better user experience


Lineflyer:
Taken from https://github.com/cgeo/cgeo/issues/9606#issuecomment-767478209 by @MagpieFourtyTwo

Anyway - I know, it's hard to explain what happens behind the scenes, but to me this text and the steps one has to take afterwards, are way to much to handle for Joe Average. In order to not getting the support flooded with questions about which directories to chose, how to get back maps and stuff and where will my exports go .., we need a clear list of steps over there, which tells the user how he can get to the situation he had before he was forced to grant access to some folders ...

---

I wonder a bit about your confusion here. AFAIR from my pre-tests of the SAF nightlies the usage is pretty straight forward:
- Nothing will ever be lost also if you select another than the existing `/cgeo` as base dir (because old data either remains or gets copied/moved)
- Offline map directory should IMHO be kept as it was before upgrading to SAF, I might be wrong though @eddiemuc 
- Upon exporting something from c:geo you will be informed where it was stored (same as before)

fm-sys:
> I wonder a bit about your confusion here. AFAIR from my pre-tests of the SAF nightlies the usage is pretty straight forward

I think the biggest mistake we make is that we overestimate our average users. The only thing they want is that everything works without problems. As has been said several times, the copy/move selection options are avoidable and not clear enough to non IT experts. Also the "explanatory text" is way too long. Users don't want to be told big why something is technically necessary, they just want to know what they need to do - with words as simple and compact as possible

MagpieFourtyTwo:
After closing the message window, I got exactly one more dialog to chose a folder and I selected ``internal\cgeo``. as has been recommended before. Main reason for my "bad feeling": I have had no clue at all, which settings have been effected by this choice. Main effect was, however, that there were no maps anymore in my Map selection drop down list - and this was because the above procedure most obviously changed my mapsVector folder from ``sdcard\Locus\mapsVector`` to ``internal\cgeo\maps``.

Have not said that something has been deleted, cause after switching the folder back the map list was there again, of course.

Nevertheless, kind of irritating, that in the Map Data config section the ``Directory with offline maps`` is shown as ``...\Locus\mapsVector (User-defined)``, no matter if I have chosen the folder from SD card or Internal Storage.

MagpieFourtyTwo:
[...]

PR #9940, opened 2021-02-01
convert file:/// to content:/// Uri so that it can be used as EXTRA_INITIAL_URI

fm-sys:
See https://github.com/cgeo/cgeo/issues/9872#issuecomment-770776886 as reference...


eddiemuc:
You know what? That's a pretty clever idea! Touche, and thanks!

eddiemuc (approved):

PR #9946, opened 2021-02-03
Offer to download theme on OpenAndroMaps map downloads (fix #9919)

### LINKED ISSUE #9919: Support downloading map theme files

moving-bits:
Last part for #9919: Offer downloading theme file "Elevate.zip" after downloading a map file from OpenAndroMaps (but only if "Elevate.zip" does not yet exist in theme folder).

image


Lineflyer:
For the normal menus (in contrast to the wizard) I would still prefer not using personification of c:geo ;)  (Would you like to download this as well?)

PR #9947, opened 2021-02-03
add multithreaded map rendering for OSM (fix #9895)

### LINKED ISSUE #9895: Multi-thread option for Mapsforge
Mapsforge support a multi-thread tile rendering option which we don't use yet AFAICS.

We should check whether we can use it and whether it speeds up tile rendering significantly in our use-case. If so we should integrate it.

See `MapWorkerPool` etc. in Mapsforge.

moving-bits:
For testing purposes only!

Set `NewMap` to 4 rendering threads (see #9895).
Fixed value, no user interface supplied, testing only.

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -821,6 +821,10 @@ public class Settings {
         return getInt(prefKeyId, getKeyInt(defaultValueKeyId));
     }
 
+    public static int getMapOsmThreads() {
+        return getBoolean(R.string.pref_map_osm_multithreaded, false) ? Math.max(1, getInt(R.string.pref_map_osm_threads, Math.min(Runtime.getRuntime().availableProcessors() + 1, 4))) : 1;
+    }
+
     public static int getCompactIconMode() {
         final String prefValue = getString(R.string.pref_compactIconMode, "");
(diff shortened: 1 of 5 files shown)


moving-bits:
Prepared a PR:
- option to activate multithreaded rendering (default = off; see Settings => Map => Other)
- use 1 thread (if not activated) / use number of cpus + 1 (min 1, max 4)
- value can be overridden by "map_osm_threads" preference value (has to be created manually)

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -821,6 +821,10 @@ public static int getMapLineValue(final int prefKeyId, final int defaultValueKey
         return getInt(prefKeyId, getKeyInt(defaultValueKeyId));
     }
 
+    public static int getMapOsmThreads() {
+        return getBoolean(R.string.pref_map_osm_multithreaded, false) ? Math.max(1, getInt(R.string.pref_map_osm_threads, Math.min(Runtime.getRuntime().availableProcessors() + 1, 4))) : 1;
```
fm-sys, line 825: Why are you using the hard-coded number 4 as limit for multithreading, and not always the available number of processors?

moving-bits: To limit resource usage, weighting additional memory use vs. incremental rendering performance gain.

PR #9955, opened 2021-02-05
saf: make copying asynchronous and with progress bar

eddiemuc:
saf: make copying asynchronous and with progress bar

Snapshot of progress bar (this will change rapidly)

image

Snapshot of copy conclusion dialog
image


eddiemuc:
Update: typo in strings.xml was fixed (it says now "Tap 'ok' **to** finalize the folder change' in above dialog)

PR #9956, opened 2021-02-05
integrate folder migration in wizard

moving-bits:
Integrates the first two available base folders into the installation & migration wizard. When in migration mode, the wizard dialogs will be like this:

image
image.image
image.image

The first screen is the opening screen and a short explanation what will follow, the next for screens will be repeated for every folder that needs migration.

This is currently implemented for "maps" folder and for "gpx" folder. (Base folder had already been implemented earlier.) Migration skips all folders no value has been set for.

As for the user interface: It skips the detailed intermediate dialog we currently have, and selects "move" method "under the hood".
For base folder we give an extra explanation for migrating user to select their old base folder in "/cgeo" as default. Do we need such a thing here as well?

@eddiemuc: Please crosscheck, if I'm calling the necessary steps with the right parameters for a default migration scenario, see my minimalistic method `ContentStorageActivityHelper.migratePersistableFolder()`.


eddiemuc:
From code review I conclude that this should work. Nicely done, also the migration check in the Settings.

Note that I changed in parallel the copy/move process quite a bit and also the dialog after copy/move. But I think this will fit together with your changes in the wizard. I just hope that there will be no merge conflict...

eddiemuc (approved):

moving-bits:
> From code review I conclude that this should work. Nicely done, also the migration check in the Settings.

Thanks for reviewing.

> Note that I changed in parallel the copy/move process quite a bit and also the dialog after copy/move. But I think this will fit together with your changes in the wizard. I just hope that there will be no merge conflict...

Seems like there is no merge conflict :-)

BTW: Am I right that BASE, OFFLINE_MAPS AND GPX are the only folders migrated to SAF already?

eddiemuc:
> BTW: Am I right that BASE, OFFLINE_MAPS AND GPX are the only folders migrated to SAF already?

If you refer to those folders whose location is selectable via our Settings, then you are right. MAP-THEMES and BACKUP are missing.

Also log images and spoilers still need migration. Other stuff is done as far as I know.

moving-bits:
Ok, thanks. Those can be easily integrated into the current logic of the wizard once they are implemented.

PR #9968, opened 2021-02-06
add textual info for routingmode in map quick settings popup

moving-bits:
image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1339,7 +1339,7 @@
     <string name="map_live_disable">Disable live</string>
     <string name="map_as_list">Show as list</string>
     <string name="map_select_multiple_items">Select item</string>
-    <string name="map_routing">Routing</string>
+    <string name="map_routing">Routing (Currently: %s)</string>
     <string name="map_routing_brouter">Routing is only available if <a href="">BRouter</a> is installed</string>
     <string name="map_gm_autorotation">Map auto rotation</string>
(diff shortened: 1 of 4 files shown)


fm-sys:
Is it really necessary? IMHO it's a bit confusing that the info uses the same style ad the headings...

moving-bits:
> Is it really necessary? IMHO it's a bit confusing that the info uses the same style ad the headings...

IMHO it's useful for those that do not understand the meaning of the buttons (especially "off" and "straight"), and by placing it in the "heading" it does not take up any additional space.

Maybe use "Routing (Straight)" or "Routing (Current: Straight)" instead of "Routing: Straight"?

fm-sys:
> Maybe use "Routing (Straight)" or "Routing (Current: Straight)" instead of "Routing: Straight"?

Jep, that would probably be better. Also we should IMHO (independent of this PR) try to find a better icon for "OFF"

baseline_near_me_disabled_black_18dpbaseline_link_off_black_18dpbaseline_explore_off_black_18dp

What about one of those? All three are material design icons...

moving-bits:
Adapted the headline accordingly in this PR and prepared #9971 for the new button.

bekuno:
A bit off topic - the spacing before the headlines is different. Before `use compact icons` we have nearly a line hight space, bot nothing before `Routing` and `Individual route`.
I would suggest to have a small space before every headline.

PR #9970, opened 2021-02-07
make compact icon buttons same width in map quick settings

moving-bits:
Currently the "compact icon mode" buttons in map quick settings popup look like this for OSM:

image

Google Maps do a better job here, although they share the popup's layout definition.

This PR changes the layout for compact icon mode buttons to look balanced for both maps:

image.image

### DIFF
--- a/main/res/layout/map_settings_dialog.xml
+++ b/main/res/layout/map_settings_dialog.xml
@@ -34,29 +34,33 @@
                 android:layout_height="wrap_content"
                 android:layout_marginLeft="6dip"
                 android:text="@string/map_dot_mode" />
-            <Button
-                android:id="@+id/compacticon_off"
-                style="@style/button_small"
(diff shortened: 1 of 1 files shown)


bekuno:
The buttons in https://github.com/cgeo/cgeo/pull/9971#issue-568935198 are a bit smaller.
Which version is correct?
(Personally I prefer the smaller version)

moving-bits:
> The buttons in [#9971 (comment)](https://github.com/cgeo/cgeo/pull/9971#issue-568935198) are a bit smaller.

This is what Google Maps auto-generated with the old settings. Alas, OSM did very differently with the same settings and ignored all the different settings I had tried for width, padding, margin etc.

So I removed all those layout settings but added a "layout_weight" with value 1 for each button in this line, and that's the result here, working consistently on both GM and OSM.

PR #9978, opened 2021-02-07
Detect special case source == target in folder migration dialog

fm-sys:
grafik

@eddiemuc There was a status code NOTHING_TO_COPY which was removed in b2e7130240a6e9230b5597c6dced17e164b16305? Why was it deleted, and can we re-add it? That would simplify the if statement and would provide the option to easily change the dialog title as well...

---
Unrelated to this PR but came me in mind while writing the PR description: Shouldn't folder paths end with "/"?

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -891,6 +891,7 @@
     <string name="folder_move_finished_title">Folder content was moved</string>
     <string name="folder_copy_finished_dialog_message">Content of folder %1$s was copied to %2$s.\n\n%3$s of %4$s and %5$s of %6$s were processed.</string>
     <string name="folder_move_finished_dialog_message">Content of folder %1$s was moved to %2$s.\n\n%3$s of %4$s and %5$s of %6$s were processed.</string>
+    <string name="folder_copy_move_not_necessary_dialog_message">Folder %1$s can be used without further migration.</string>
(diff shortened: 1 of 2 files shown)


eddiemuc:
> @eddiemuc There was a status code NOTHING_TO_COPY which was removed in b2e7130240a6e9230b5597c6dced17e164b16305? Why was it deleted, and can we re-add it? 

No specific reason for deletion, just wanted to simplify things so there is only one "ok" result and all other results are somehow an error. But you can readd it if you need it, no problem.


> Unrelated to this PR but came me in mind while writing the PR description: Shouldn't folder paths end with "/"?

Hm. Today's File-based paths don't end with / either and Uris usually don't end with / too. You really want to add this?

fm-sys:
> Hm. Today's File-based paths don't end with / either and Uris usually don't end with / too. You really want to add this?

Just asking... To be honest, I don't really care about it 😅

bekuno:
The title `"Folder content was moved"` is not correct, if nothing was moved.
But - why do we need the dialog? As we would try to `"keep it simple"` we can skip this dialog.
At the end maybe we can say - `"all folders where checked, nothing was changed"`.

moving-bits:
Do we still need this PR after integration of
- auto-detection of source==target
- and skipping any message if source==target

?

PR #9987, opened 2021-02-08
fix #9963: deactivate folder adjustment, fix uridisplay issue, fix pseudouri problem on windows

### LINKED ISSUE #9963: SAF: Map dir on extSD gets reset to default after device reboot
Using 2021.02.01-NB-bfd9fbf

- Samsung S20, Android 11
- Set offline map directory to a folder on my real external SD 
- Offline maps can be used normally
- End c:geo
- Reboot device (Remark: As I use Android encryption of my external SD it take approx. 20-30 seconds after reboot until it is mounted and available)
- Start c:geo
- Toast is shown, that offline folder is no longer accessible and reset to default

The reason for this problem is obvious: I started c:geo before SD was initialized completely by the system.

But still its really annoying to have to set the folder permission again in such cases.

Could we postpone this check until the folder content is really needed, i.e. start the map activity?

eddiemuc:
fix #9963: deactivate folder adjustment, fix uridisplay issue, fix pseudouri problem on windows

This PR effectively removes automatic adjustment of inaccessible public folder uris to default value. This should account for temporarily failing access to those uris.

### DIFF
--- a/main/src/cgeo/geocaching/downloader/MapDownloaderUtils.java
+++ b/main/src/cgeo/geocaching/downloader/MapDownloaderUtils.java
@@ -116,7 +116,7 @@ public class MapDownloaderUtils {
             @Override
             protected void execute() {
                 final PersistableFolder folder = PersistableFolder.OFFLINE_MAPS;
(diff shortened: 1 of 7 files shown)


Lineflyer:
retest this please

PR #9989, opened 2021-02-08
Fix #9988 - Mocked cache changed attributes

### LINKED ISSUE #9988: Testcase for GC3FJ5F needs to be adapted
I have recently set the attribute "Not suitable in winter" to my cache GC3FJ5F. This seems to break some unit test.

Sorry for that...but current weather made me aware, that the attribute was missing for that cache :)

Lineflyer:


Lineflyer:
retest this please

Lineflyer:
OK, I think I found it.

PR #9998, opened 2021-02-10
remove unused string

moving-bits:
fix lint error from last cgeo integration build


moving-bits:
HTML Source for trackable owner has changed, that's why the tests recently fail, see #9999.

PR #10006, opened 2021-02-13
fix #9609, #9806: scale offline log images on upload only

### LINKED ISSUE #9609: Migrate log image location and use SAF
Currently c:geo is using `Environment.getExternalStoragePublicDirectory` to store log images the user takes or selects while writing his geocache logs. 
Details about the current disk usage structure can be found here: https://github.com/cgeo/cgeo/wiki/Disk-usage-structure
We are talking about item <4> of the table in this wiki article.

Starting with target SDK=30 Android enforces to use the StorageAccessFramework (SAF) instead of plain file access permissions.
Therefore we should migrate from the current access mode into using the SAF.

The general approach has been discussed in #8457 already, which I try to summarize below (correct me, if I am wrong) for this item:
- `getExternalStoragePublicDirectory` is deprecated anyway, so it needs to be replaced sooner or later
- The majority voted to therefore remove this method of storing images and use the same location as for external user data (see #9606)

To be agreed/discussed:
- Is migration needed or could we simply ignore what is stored in `/pictures/cgeo`?
- Does it make sense to remove the deprecated method even before introducing SAF (two-stepped-approach) or better do it in the same step?

### LINKED ISSUE #9806: Cache log photos internally and scale/compress before sending only
Two users contacted us on support mail with this problem. Did not have time/opportunity to test it:
- Have picture stored in galery
- Post a log and add the picture with the legacy function (Menu - Add picture)
- Select "No scaling"
- Send log 
- Post again a log and add the same picture via quick button instead of the old function
- Send log

Result:
The first picture has a higher resolution than the second one. The second one seems to get scaled to 1024px on longer side.

Is this a bug or known limitation?
The examples posted by a user proofs that such limitation can results in pretty unusable pictures (in that users case he took a photo of a signpost with a lot of text, it was not possible to read it anymore).

eddiemuc:
This PR is a major change for storing offline log images and tries to fix #9609 and fix #9806.

Description will follow in a comment

### DIFF
--- a/main/res/layout/imagelist_item.xml
+++ b/main/res/layout/imagelist_item.xml
@@ -59,7 +59,7 @@
                 android:layout_gravity="left"
                 android:ellipsize="end"
                 android:focusable="false"
-                android:maxLines="2"
+                android:maxLines="3"
                 android:scrollHorizontally="true"
                 android:textColor="?text_color_headline"
                 android:textSize="14sp"
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -259,6 +259,7 @@
     <string translatable="false" name="pref_map_direction">map_direction</string>
     <string translatable="false" name="pref_map_routing">map_routing</string>
     <string translatable="false" name="pref_theme_menu">theme_menu</string>
+    <string translatable="false" name="pref_localstorage_version">localstorage_version</string>
 
     <!-- ids for persistable folder prefs -->
     <string translatable="false" name="pref_group_localfilesystem">group_localfilesystem</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -163,8 +163,9 @@
     <string name="log_problem_archive_text">This geocacher reported that this geocache should be archived. A community volunteer reviewer has been notified.</string>
 
     <string name="log_image_scale_option_noscaling">No scaling</string>
-    <string name="log_image_scale_option_entry">%1$d px</string>
-    <string name="log_image_info">%1$d x %2$d px, %3$d kB, %4$s</string>
+    <string name="log_image_scale_option_entry">%1$d x %2$d px</string>
+    <string name="log_image_scale_option_entry_noimage">%1$d px</string>
+    <string name="log_image_info">%1$d x %2$d px, %3$s\nScaled/Compressed:\n%4$d x %5$d px, ~%6$s</string>
     <string name="log_image_titleprefix">Image</string>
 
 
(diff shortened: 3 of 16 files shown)


eddiemuc:
Changes:

**User perspective**
Offline Log Image are now scaled on upload and not on save. This allows for change of wanted scale size also to an upper level after saving without loss of quality. It also results in more device space being used to store all images in original size until upload. Theoretically it also results in longer upload times because scaling is done there too, but this should not have a real impact because it is done real fast.

Displayed information per image in Offline Log has changed significantly:
* No file name is displayed any more, because the image is stored in private app space and thus not accessible by user outside c:geo anyway
* Image width/height and original file size is displayed as well as target scale size and estimated file size after scale and compression. Note that this is really a rough estimate (calculated by a very simple rule-of-thumb) and might not be the real value.

Looks like this:
image

**Device Storage**
Offline log images are stored in app private storage. More precisely: in GeoCacheData directory in their corresponding cache code dir, and there in a subdir "OfflineLogImages". Images are stored in original size and untampered with. When selecting images from device, a copy is made into that device. When images are made via camera from c:geo, the image is stored in public standard "Pictures" directory and ADDITIONALLY a copy is stored in above folder. The pic in the public folder will no longer be touched from thereon

image

** Database **
This PR introduces a new DB Version 94 which introduces a new column "targetScale" in "offlinelogimages" table. This allows for storing individual target scale sizes on a per-image-base

** Migration **

Existing log images in Pictures/cgeo are **copied** to their respective GeocacheData subfolders together with database upgrade to version 94. I decided to do a copy so user has still public access to his existing images in case he uses them in any other way up to now. Also if theres a problem in migration, he will at least not loose the images completely into private c:geo space.

fm-sys:
Very nice! 👍 

Just one question:
> Existing log images in Pictures/cgeo are copied to their respective GeocacheData subfolders together with database upgrade to version 94.

What will happen if the user has quite some files (let's say 100MB). Is copying done with a async task/progress dialog or could it result in ANR's?

eddiemuc:
> Very nice! 👍
> 
> Just one question:
> 
> > Existing log images in Pictures/cgeo are copied to their respective GeocacheData subfolders together with database upgrade to version 94.
> 
> What will happen if the user has quite some files (let's say 100MB). Is copying done with a async task/progress dialog or could it result in ANR's?

Right you are. I will change this.

eddiemuc:
@fm-sys 
I realized that there is not yet any form of versioning in LocalStorage to handle file migrations. So I added up a version upgrade routine in LocalStorage which now works quite similar to DataStore:
* There is an expected version maintained in a constant variable LOCALSTORAGE_VERSION (currenty set to 1)
* There is a new field in the Settings storing the current Version (defaults to 0 if not set)
* I created an asynchronous MigrationTask which handles the asynchronous display stuff. In the doInBackground routine of this Task file migration routines can place themselves and their migration tasks. Progress GUI will be handled by the framework, migrators can send messages to the progress bar.

Migration to version 1 now includes the copying of OfflineLogImages from legacy dir to GeocachData dirs. This goes so fast that I can't even see the progress bar. The following screenshot was done in debug mode (by pausing the thread):

image

Will push this to the branch any minute

PR #10011, opened 2021-02-14
Add db/file/settings version info to system information

eddiemuc:
Adds db/file/settings version info to system information

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -129,14 +129,22 @@ public class Settings {
         throw new InstantiationError();
     }
 
+    public static int getActualVersion() {
+        return getInt(R.string.pref_settingsversion, 0);
+    }
+
+    public static int getExpectedVersion() {
+        return 5;
+    }
+
     private static void migrateSettings() {
         //NO migration in NO_APP_MODE
         if (NO_APPLICATION_MODE) {
             return;
         }
 
-        final int latestPreferencesVersion = 5;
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -129,14 +129,22 @@ protected Settings() {
         throw new InstantiationError();
     }
 
+    public static int getActualVersion() {
+        return getInt(R.string.pref_settingsversion, 0);
+    }
+
+    public static int getExpectedVersion() {
+        return 4;
```
fm-sys, line 137: Shouldn't it be rather "5"? That's the current version...

eddiemuc: actually it is not. if you see the existing migration steps you will see that prop version is never set to 5, but only to 4

fm-sys: For me, it actually _is_ the case. Was there some kind of migration routine in the nightly which was removed afterwards? In any case, we should not go back with the version number...

Screenshot_20210214_120606_copy_536x430

eddiemuc: Ah now that you mention it: yes there was a version 5 spilling into the nightlies for a few days. Should never affect the real users. But as a service to the nightly users I will introduce a dummy version 5 with no migration work. I will push this any minute.

fm-sys: Thank you! That's probably the best, since we have a few dozen nightly users...

PR #10013, opened 2021-02-14
fix #9719: migrate spoiler image folder to SAF

### LINKED ISSUE #9719: SAF: migration of "Spoiler images"
This is an issue to deal with (possibly needed) SAF migration for "Spoiler image" folder

There is an external directory defined in LocalStorage namend "GeoCachePhotos" which is supposed to contain "spoiler images":

https://github.com/cgeo/cgeo/blob/ccb0385dd8b65b5fb124e418dbf4cf4499ad24e2/main/src/cgeo/geocaching/storage/LocalStorage.java#L344-L347

I do not have this dir on my installation. A quick code scan seems to indicate that this is only used for oc and su. 

image

I don't use SU and OC very seldom, have not come across these spoilers yet. What is this, and is it still needed? If yes, can it go to a private folder?

eddiemuc:
fix #9719: migrate spoiler image folder to SAF

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -270,6 +270,7 @@
     <string translatable="false" name="pref_persistablefolder_gpx">persistablefolder_gpx</string>
     <string translatable="false" name="pref_persistablefolder_backup">persistablefolder_backup</string>
     <string translatable="false" name="pref_persistablefolder_fieldnotes">persistablefolder_fieldnotes</string>
+    <string translatable="false" name="pref_persistablefolder_spoilerimages">persistablefolder_spoilerimages</string>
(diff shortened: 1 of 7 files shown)


moving-bits:
@eddiemuc 
Can I detect somehow whether migration has already taken place or not?
(So that I can integrate it into the migration wizard)

Lineflyer:
IMHO there is no need for specific migration as the spoiler folder is per definition below the base folder...at least I would strongly suggest to keep it like that and not have it as a selectable other folder.

moving-bits:
Ah, you're right, so I don't need any specific migration for this folder - it's enough if BASE folder is set/migrated.

PR #10014, opened 2021-02-14
fix #10009: make explicit when caches are deleted completey in cache lists

### LINKED ISSUE #10009: Question: Offline Log Deletion on Cache Deletion ok?
I noticed that when you delete a cache from c:geo which has a not-yet-send Offline Log, then this Offline Log is deleted (without any security question).

Is this wanted behaviour or a bug?

Reproducable as follows:
* opening any cache
* Create offline Log, write some log text
* Go back (this saves the log)
* Delete the cache by tapping the "delete" trash can icon
* -> offline log is gone

eddiemuc:
fix #10009: make explicit when caches are deleted completey in cache lists

If we are in a user-defined list, cache menu remove entries is unchanged:
image

If we are in a "special" list (All, Temporaray, History) then labels for remove-entries are different:
image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -371,6 +371,8 @@
     <string name="caches_nearby">Nearby</string>
     <string name="caches_remove_all">Remove all</string>
     <string name="caches_remove_selected">Remove selected</string>
+    <string name="caches_remove_all_completely">Delete all from device</string>
+    <string name="caches_remove_selected_completely">Delete selected from device</string>
     <string name="caches_delete_events">Delete past events</string>
     <string name="caches_upload_modifiedcoords">Upload modified coordinates</string>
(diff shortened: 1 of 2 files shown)


eddiemuc:
Note: logic remains unchanged, just the labeling was changed to reflect that logic is very different in the two cases

Lineflyer:
Are you sure it really works like that?

If I e.g. have a cache on the default list and(!) on another user defined list, and delete it from the default list it will IMHO still be stored in that other user defined list.

eddiemuc:
> Are you sure it really works like that?
> 
> If I e.g. have a cache on the default list and(!) on another user defined list, and delete it from the default list it will IMHO still be stored in that other user defined list. 

The "default" list in this sense is a "user-defined" list. "Special" Lists (where cache is really deleted) are just the three mentioned ones: All, Temporary and History

I can show you the exact code line where this is decided :-)

Lineflyer:
Sorry, I mixed up default with history list. You seem perfectly right :)

PR #10019, opened 2021-02-15
migrate backup to SAF

fm-sys:
My work is nearly done, although it was quite a lot to refactor. 

I took the opportunity to improve some code parts with what I learned about java in the last months (using `Consumer` for example - great stuff!)

I am currently stuck with having no access to the folder that is selected via the tree chooser intent. @eddiemuc Could you take a look at the code, if SAF is used correctly and why the tree chooser doesn't work? Is there something I need to do with the folder before it can be accessed?

Everything else seam to work fine with SAF...

### DIFF
--- a/main/src/cgeo/geocaching/utils/BackupUtils.java
+++ b/main/src/cgeo/geocaching/utils/BackupUtils.java
@@ -4,8 +4,12 @@ import cgeo.geocaching.MainActivity;
 import cgeo.geocaching.R;
 import cgeo.geocaching.settings.BackupSeekbarPreference;
 import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.storage.ContentStorage;
+import cgeo.geocaching.storage.ContentStorageActivityHelper;
 import cgeo.geocaching.storage.DataStore;
-import cgeo.geocaching.storage.LocalStorage;
+import cgeo.geocaching.storage.Folder;
+import cgeo.geocaching.storage.FolderUtils;
+import cgeo.geocaching.storage.PersistableFolder;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import static cgeo.geocaching.utils.SettingsUtils.SettingsType.TYPE_STRING;
@@ -17,9 +21,7 @@ import android.app.AlertDialog;
 import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.content.res.Resources;
-import android.database.sqlite.SQLiteDatabase;
-import android.database.sqlite.SQLiteException;
+import android.net.Uri;
 import android.util.Xml;
 import android.view.ContextThemeWrapper;
 import android.view.LayoutInflater;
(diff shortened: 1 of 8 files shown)


eddiemuc:
@fm-sys can you give me a pointer to the code part where the Problem is happening? And what happens exactly (do you get an error message or just a null value or something else)?

main/src/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -65,69 +70,69 @@ public BackupUtils(final Activity activityContext) {
     }
 
 
+    /* Public methods containing question dialogs, etc */
+
+    public void selectBackupDirIntent (final ContentStorageActivityHelper contentStorageHelper) {
+        contentStorageHelper.selectFolder(PersistableFolder.BACKUP.getUri(), this::restore);
```
fm-sys, line 76: Here I start the dirchooser

main/src/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -65,69 +70,69 @@ public BackupUtils(final Activity activityContext) {
...
-                    .setView(content)
-                    .setPositiveButton(activityContext.getString(android.R.string.yes), (alertDialog, id) -> {
-                        final boolean database = databaseCheckbox.isChecked();
-                        final boolean settings = settingsCheckbox.isChecked();
+        if (backupDir == null) {
+            return;
+        }
 
-                        alertDialog.dismiss();
+        if (!hasBackup(backupDir)) {
```
fm-sys, line 89: This one is called afterwards. `hasBackup` should return true, but it isn't...

`backupDir` looks normal at a first glance (correct URI)

main/src/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -380,99 +407,101 @@ private File createSettingsBackupInternal(final File backupDir, final Boolean fu
...
-     * Public methods for checking the backup availability
-     */
+    /* Methods for checking the backup availability */
+
+    public static boolean hasBackup(final Folder backupDir) {
+        // TODO:
+        //  When selecting a user defined restore directory, ContentStorage.get().list(backupDir) returns an empty List although files are present,
+        //  so getDatabaseFile()/getSettingsFile() returns null.
+        //     List<ContentStorage.FileInformation> a = ContentStorage.get().list(backupDir);
+        //  Why is it not working?
```
fm-sys, line 473: As described in the comment,  `ContentStorage.get().list(backupDir)` returns an empty List. Not `null` but size 0.

main/src/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -380,99 +407,101 @@ private File createSettingsBackupInternal(final File backupDir, final Boolean fu
...
-    public static boolean hasBackup(final File backupDir) {
         return getDatabaseFile(backupDir) != null || getSettingsFile(backupDir) != null;
     }
 
     @Nullable
-    public static File getDatabaseFile(final File backupDir) {
-        final File fileSourceFile = DataStore.getBackupFileInternal(backupDir, true);
-        return fileSourceFile.exists() && fileSourceFile.length() > 0 ? fileSourceFile : null;
+    private static ContentStorage.FileInformation getDatabaseFile(final Folder backupDir) {
+        return ContentStorage.get().getFileInfo(backupDir, DataStore.DB_FILE_NAME_BACKUP);
```
fm-sys, line 480: As `ContentStorage.get().getFileInfo` uses `list(Folder)` internally, the file is not found...

fm-sys (commented):
@eddiemuc I added some review comments to the relevant code parts. I hope that helps!

eddiemuc:
Funny, I just had the same problem whenever I select a folder from the "Downloads" document provider. When I select a folder from the "internal storage" provider it seems to work.
@fm-sys is it the same effect for you or does this happen also for "internal storage" prodivder folders?

fm-sys:
As far as I understood, it actually was the "internal storage" provider... 

```
W/cgeo: [main] Folder set to 'content://com.android.externalstorage.documents/tree/primary%3Acgeo%2Fbackup%2F2021-02-14%2017-07' 
```

Any idea how we could solve this problem? Might it be related that the folder is selected without the persistent flag?

fm-sys:
The other idea would be to use a "classic" file chooser instead. The more I think about it, the more I actually prefer this solution.

Then we wouldn't need the checkbox dialog anymore, and the user can select DB and/or settings backup directly in the file chooser chooser. (Note that this would only affect the "restore a different backup" functionality for sure)

Would that be a solution?

eddiemuc:
You mean the user would choose a single backup file instead of a directory? Could work I guess. 

There is a bug in ContentStorage for sure, but I haven't found it yet

fm-sys:
> You mean the user would choose a single backup file instead of a directory? 

Yes. Or two files at the same time, if the user wants to restore db & settings...

fm-sys:
Thanks, @eddiemuc! #10036 does the trick and folder selection is working now...

PR #10021, opened 2021-02-15
replace the last old profile link (fix #9150)

### LINKED ISSUE #9150: Retirement of old profile page on gc.com

fm-sys:
see https://github.com/cgeo/cgeo/issues/9150#issuecomment-779318569

IMHO it does not worth to release a new version just because of this small fix. Anyway, if we plan a bugfix-release it's targeted to release branch....

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -106,8 +106,8 @@ public final class GCConstants {
     static final Pattern PATTERN_TRACKABLE_ORIGIN = Pattern.compile("<span id=\"ctl00_ContentBody_BugDetails_BugOrigin\">([^<]+)<\\/span>[^<]*</dd>");
     /** Two groups ! */
     static final Pattern PATTERN_TRACKABLE_SPOTTEDCACHE = Pattern.compile("<a id=\"ctl00_ContentBody_BugDetails_BugLocation\" title=\"[^\"]*\" href=\"[^\"]*/seek/cache_details.aspx\\?guid=([a-z0-9\\-]+)\">In ([^<]+)</a>");
-    /** Two groups ! */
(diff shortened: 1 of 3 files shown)


fm-sys:
Is there any trackable which is permanently in the hands of a different user (not the owner), so I can write a unit test for it? As currently the regex is not covered by the unit tests...

moving-bits:
> Is there any trackable which is permanently in the hands of a different user (not the owner), so I can write a unit test for it? As currently the regex is not covered by the unit tests...

You can use one of mine, if you like: TB9NTJQ

fm-sys:
> You can use one of mine, if you like: TB9NTJQ

The problem is that the trackable needs to be permanently in the hands of one and the same user. No dropped logs, etc.
 
But as it's a really minor functionality, it's probably also OK if we find no trackable which is permanently matching these criterias. This is probably the reason why there were no unit test before...

moving-bits:
So you need a TB which is activated by user X, then gets grabbed by user Y and stays there?

fm-sys:
Jep. Or leave out the unit test 😉

bekuno:
> 
> 
> So you need a TB which is activated by user X, then gets grabbed by user Y and stays there?

E.g. a activated TB as a gift at a mega/giga event for supporter.
(I logged such a TB a long time ago, but the owner deleted then log)

Lineflyer:
Are we talking about "spotted by user"?
Isn't it sufficient that user y has once "discovered" the trackable of user x?

moving-bits:
@fm-sys 
I created such a TB: TB8WRVW

fm-sys:
> 
> 
> @fm-sys
> I created such a TB: TB8WRVW

Great, thank you! By chance, I have also discovered another trackable right now, which also meets the criteria. TB: TB4BPQK

It hasn't received any future logs since 2013, so it should be a rather good test case. @moving-bits So we can either think about a different special test case for you're trackable, or it may travel again ;-)

Issue #10030, opened 2021-02-16
SAF: Behavior of migration wizard if folder selection is aborted


Lineflyer:
I found time today to run another test of the SAF migration by upgrading from 2021.02.14 to todays nightly.

IMHO the wizard works really well and is understandable, but I stumbled over one scenario where I would like to get your feedback:

At some point of the different SAF folder selections aborted the selection process as I wanted to read again which folder I was asked to assign at this step. But in such case we takeover the legacy value (toast is shown) and the screen is gone (next wizard screen shown).

Two improvement proposals:
- If the selection is aborted we should IMHO return to the current step and offer something like "Skip for now" (to takeover the legacy value) or "Next" to try again to select a folder (Maybe in general its even better to name the default button "Select" instead of "Next" if it triggere the folder selection).
- Shouldn't we use different captions on the wizard pages instead of always the same "Set app permissions" so that you see already in this caption what is asked next (e.g. "Set base folder", "Set map path", "Set GPX path")? That was, what initially confused me and made me go back to doublecheck.

moving-bits:
> If the selection is aborted we should IMHO return to the current step and offer something like "Skip for now" (to takeover the legacy value) or "Next" to try again to select a folder (Maybe in general its even better to name the default button "Select" instead of "Next" if it triggere the folder selection).

I'll need to take a look at it what's possible / how to achieve this behavior.
If you skip, the path should not be set implicitly, IMHO, but left at "undetermined", which leads to using the legacy value for now, but keeps popping up the migration wizard on next restart.

> Shouldn't we use different captions on the wizard pages instead of always the same "Set app permissions" so that you see already in this caption what is asked next (e.g. "Set base folder", "Set map path", "Set GPX path")? That was, what initially confused me and made me go back to doublecheck.

Yes, had thought of that myself a couple of times, but always something different thwarted it :-(

moving-bits:
@eddiemuc 
I followed a cancelled folder selection (*) dialog back to `handleResultPersistableFolderSelection`, which will call the `callback` in this case. Is there any way I can detect a cancelled selection from the `PersistableFolder` that gets returned to the `callback`?
And if not, would it harm if I called the `callback` with a `null` value as marker for `selection cancelled`? (I would use this then in the installation wizard to adapt the wizard page accordingly instead of forwarding to the next wizard page).

[...]

PR #10031, opened 2021-02-16
use folder-specific titles in wizard (rel. to #10030)

moving-bits:


eddiemuc:
Hi @moving-bits , there is already a translated name for each persistable folder in strings.xml. maybe we can reuse thise instead of creating new keys just for the wizard?

moving-bits:
Ah, yes, you're right, thanks for reminding me. I'll combine a generic title with the folder-specific translation from `PersistableFolder`.

Issue #9607, opened 2020-12-27
Migrate and use SAF for offline map themes


Lineflyer:
This issue is similar to #9606 but for offline maps it might be a bit more complicated as we should not enforce this data to be in the same directory as the external user data (see  https://github.com/cgeo/cgeo/wiki/Disk-usage-structure , item <3>) and migration is more relevant (as c:geo kind of needs the offline map data to work as desired by the user).

Starting with target SDK=30 Android enforces to use the StorageAccessFramework (SAF) instead of plain file access permissions.
Therefore we should migrate from the current access mode into using the SAF.

The general approach has been discussed in #8457 already, which I try to summarize below (correct me, if I am wrong) for this item together with my ideas for the offline map case:
- On fresh installation we might use the SAF location requested in #9606 and store maps (downloaded by our own downloader) into `../maps/` below our base directory. This avoids another SAF request to the user for that default case
- We should however (for existing users and for users preferring to store them a different location than the "normal" external user data) offer the possiibility to set up a dedicated location for those via SAF.
- This could be implemented by a selection dialog in Settings - Maps, which offers 
  - Default = Use the base dir for external user data (corresponds to currently used `/cgeo/maps/`)
  - User defined = Let the user choose a base directory explicitly for his offline maps

To be agreed/discussed:
- How to deal with the migration case? Can we analyze the currently set path and if it is not the default we trigger the SAF request and help the user by prompting his current selection and ask him to confirm it via SAF?
- Do we want to keep another dedicated directory (yet another SAF setup) for map themes or at least combine maps and themes into one base directory?
- Can we use the opportunity also here to finally get rid of the ugly dir chooser?

-

eddiemuc:
Largely adressed in branch `storage_framework`, please use it as a base for contributions.
Aslo please give me (@eddiemuc) a short hint before you do since I currently `rebase ` this branch onto `master ` frequently using forced pushes.

eddiemuc:
Question by @geocachermgo here: https://github.com/cgeo/cgeo/issues/9606#issuecomment-751470052:
> Will this affect to read maps from the locus directory? 

Yes, as far as I can tell right now you will need to reselect the dir with locus maps in it one more time in maps settings after SAF migration

geocachermgo:
I have no problem with that, but was afraid that have to store 2 copies of the maps

eddiemuc:
@Lineflyer with regards to the open points in the description:

[...]

PR #10051, opened 2021-02-20
rel to #9607: migrate map theme folder to SAF

eddiemuc:
rel to #9607: synchronizeFolder method

Implements new Methods in FolderUtils to synchronize a public folder to an app internal (File) folder, complete with GUI support and Unit tests.
This is the base to implement map theme folder synchronization to an c:geo internal folder to speed up its usage.


fm-sys:
Is caching the files in internal storage really needed, when using the ziped theme provider?

eddiemuc:
@fm-sys sync is needed definitely to support legacy installations where users use the extracted zips (like today). Whether we can handle zips differently from this generic mechanism is something we can imho discuss in a second step. For now it is far easier to handle everything map theme related the same way. 
Also this was the way we decided in #9607 prior to implementation.

Specific to the ZIP handling: in my experience, when reading the ZIP from public storage you get a "maybe acceptable" performance. But it is still far worse than reading it from internal file app storage.

eddiemuc:
Since PR was not merged yesterday I will continue work on it now instead of starting fresh from master

eddiemuc:
This PR implements #9607 as described there.
* theme folder is synchronized on startup of c:geo, on settings map theme folder change and after downloading a map theme file
* together with this sync, the list of available xml themes is cached
* last recently used zip file is cached in memory. This includes all resources inside (e.g. images)

I did some tests with intentionally large zip files (I put a map file into it, just to make it very big). This results in bad performance on both c:geo startup and first map opening, because even reading zip entries out of a large zip file seems to take lots of time (even from file). With "normal" zip files (e.g. elevate.zip) there was no problem. Lets see whether this becomes an issue.

eddiemuc:
I aso identified that there should be a method in ZipXmlThemeResourceProvider to just scan for available xmlThemes (and not allocate memory). I will try to get this extension into mapsforge. In the meantime I wrote a helper method in `RenderThemeHelper`

eddiemuc:
See related PR for mapsforge here: https://github.com/mapsforge/mapsforge/pull/1190

main/src/cgeo/geocaching/settings/SettingsActivity.java
```diff
@@ -177,7 +151,7 @@ private void initPreferences() {
 
         for (final int k : new int[] {
                 R.string.pref_pass_vote, R.string.pref_signature,
-                R.string.pref_mapsource, R.string.pref_renderthemepath,
+                //R.string.pref_mapsource, R.string.pref_renderthemepath,
```
fm-sys, line 154: Might be a leftover...

main/src/cgeo/geocaching/storage/LocalStorage.java
```diff
@@ -181,6 +185,13 @@ public static File getOfflineLogImageDir(final String geocode) {
         return dir;
     }
 
+    @NonNull
+    public static File getMapThemeInternalDir() {
+        final File dir = new File(getGeocacheDataDirectory("shared"), MAP_THEME_INTERNAL_DIR_NAME);
```
fm-sys, line 190: IIRC there's a app cache directory (not accessible from outside) which could be used as well and AFAIK provides even slightly faster file access. It is made for such caching purposes and can be cleared by the system if storage space runs out (which would IMHO indeed makes sense and is fine). Don't know if it fulfills our needs though... 🤷‍♂️

`context.getCacheDir()` or something like that

main/src/cgeo/geocaching/settings/SettingsActivity.java
eddiemuc: It is indeed. I deleted the line and pushed the change.

main/src/cgeo/geocaching/storage/LocalStorage.java
eddiemuc: Hm I just read documentation of context.getCacheDir(). I would not really like to use it since it means to let the system take over control upon what's in it (in the sense that the system may delete content in it any time) and current implementation strongly relies on knowing exactly what is currently inside this dir.

Storage space should not be an issue imho:
* I think it can be expected that a user does not keep e.g. 2-3 different map theme bundles (either zip or extracted) on his device. 
* OpenAndroMap's `Elevate.zip` as an example has roughly 440KB
* -> imho its fair to assume that there should not be more data than e.g. 2-3MB in this dir. 
* For Offline Log images we recently decided to store raw images in the very same place- (see #9806), and even one such raw image takes more space than this (on my device about 3.5-4MB for a standard photo) 
* -> I think storage space argument can be neglected **or** we really have to rethink the "scale image upon upload" decision for Offline Log images

moving-bits:
Eagerly waiting for it... :-)
Just give me a short notice when this PR is finalized so I can merge it and complete my parts that rely on it.

eddiemuc:
It is finished from my side :-)

main/src/cgeo/geocaching/storage/LocalStorage.java
fm-sys: As I said, that was just such an idea that came to mind. You'll probably know best what's the way to go for this feature! 

Feel free to merge :-)

moving-bits:
@eddiemuc 
I'm trying to integrate map theme folder migration into the wizard right now. How can I detect whether that path has already been migrated (or not)? 

(For other `PersistableFolders` I could check like ` Settings.legacyFolderNeedsToBeMigrated(R.string.pref_persistablefolder_offlinemaps)`, but that does not work in this case, as `renderthemefile` seems to be used for both old & migrated path.)

eddiemuc:
@moving-bits 
I completely forgot legacy migration for map themes, many apologiers for this. I just created PR #10066 which should fix this.

There are two preferences currently used with very similar names:
* `renderthemepath `stores the legacy value for the map theme folder (and will no longer be used, but must be used for legacy value in migration)
* `renderthemefile `stores the currently selected map theme and will still be used. Currently the complete path to selected file is stored there, in the future only the map theme filename (without directory) will be stored (implementation can handle complete file paths though for legacy migration reasons)

While I worked at this I also added a new method "isLegacy()" to PersistableFolder. Instead of using the settings, you could also check the value of this method to decide whether a folder is in need of migration or not.
Sorry, I know this comes far too late...

moving-bits:
> @moving-bits
> I completely forgot legacy migration for map themes, many apologiers for this. I just created PR #10066 which should fix this.

No problem, and thanks for creating the PR. Looking at it, I got pretty close in #10064 - just should have used `renderthemepath` instead of `renderthemefile`. (BTW: Thanks for the explanations!)

> While I worked at this I also added a new method "isLegacy()" to PersistableFolder. Instead of using the settings, you could also check the value of this method to decide whether a folder is in need of migration or not.

Will also take a look at the `isLegacy()` method, but I guess this will be something different than `!folderNeedsMigration()`, as in the wizard I'm not only checking whether a folder is migrated yet or not, but also if the old setting has been used at all - if it's never been used so far, I don't do migration for it, to not bother the user with too many questions and paths at that point in time.

PR #10052, opened 2021-02-20
add toggle to pq list to display all/downloadable only

moving-bits:
Add a toggle (switch) to the main screen's pq list to toggle display between "all" and "downloadable only":

image.image

This makes live easier for people like me who have dozens of pq, but only very few of them active.

Toggle will be displayed on main screen's pq list only, not when coming from a cache list (as then the list is always filtered to "downloadable only"). The toggle's value gets remembered in the preferences. Default value is "show all" (as it is today's behavior).

### DIFF
--- a/main/res/layout/pq_actionbar.xml
+++ b/main/res/layout/pq_actionbar.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="horizontal" >
+
+    <TextView
+        android:id="@+id/title"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_alignParentLeft="true"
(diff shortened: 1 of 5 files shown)


fm-sys:
What about changing the label dynamically to make the meaning more clear? Depending on the setting, either "all" or "downloadable only"

moving-bits:
As space in the ActionBar is pretty limited I'd prefer to keep the text as short as possible, that's why I chose a one word label.

fm-sys:
You are right that space is limited. Anyway, I would use different texts for on and off to make the meaning more clear. What about "downloadable" or "online"?

Lineflyer:
How about "Recent" / "All"

moving-bits:
I'm still not happy with it. IMHO a label should not change for a switch. (The switch itself simply stands for "on" or "off", so its label should stay constant.)

A more precise description would be something like "Only show downloadable pocket queries" (with switch set to "on" = "only downloadable", "off" = "all"), but that would be way too long, so I chose "All" as a compromise (with "on" = "all", "off" = "only downloadable").

moving-bits:
Maybe "Show all" would help?

PR #10058, opened 2021-02-21
String improvements based on Crowdin feedback

Lineflyer:
Taken from issues our translators opened on Crowdin:

- Map quick settings: 
  - "Show..." is obsolete as 1. the meaning is clear by the checkboxes and 2. there is a caption "Show in map". Also smaller strings might fit better.
  - The mentioned "Show in map" should be "Show on map" (grammar) and a double quote at the end is missing
- Search bar hint says "Search for caches" but you can also search trackables. Replace by "Search caches/trackables"
- More understandable explanation of signature templates
-

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -571,7 +571,7 @@
     <string name="settings_service_active">Active</string>
     <string name="settings_navigation_disabled">This app seems not to be installed</string>
     <string name="init_signature">Signature</string>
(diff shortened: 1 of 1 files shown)


moving-bits (approved):
LGTM. 
One could rename the string var "map_show_in_map" to "map_show_on_map" as well, but that's done more easily in the IDE.

PR #10059, opened 2021-02-21
rename wizard 'skip' and add info/confirmation dialog (fix #9977)

### LINKED ISSUE #9977: SAF migration wizzard - what does the "Skip" Button mean?
The migration wizzard starts with the info screen. 
On the lower left corner we have the button `skip`.
This is not understandable, there is no info what the result of this butten will be.

moving-bits:
- renamed "skip" to "not now" on opening screen
- pressing the button will show an explanation + confirmation dialog with "back" and "ok" options

example (show for migration, but is similar in the first installation variant):
image.image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2147,6 +2147,8 @@
     <string name="wizard_status_location_permission">Location permission</string>
     <string name="wizard_status_basefolder">Base folder</string>
     <string name="wizard_status_platform">Geocaching services</string>
+    <string name="wizard_not_now">Not now</string>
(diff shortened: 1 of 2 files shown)


bekuno:
LGTM

Lineflyer:
Why should c:geo not be able to show your position on the map when skipping the migration wizard?

In case you apply the same dialog for migration and installation, still this will probably not happen as c:geo will still ask for location permission and stop when denied. Am I right?

moving-bits:
Currently the same text is used in install and in migration mode to keep things easy. How about stripping the "GPS" part?

"c:geo might not work as expected until configuration is completed. Depending on the missing configuration c:geo might not be able to read/store data like offline maps, geocaches, logfiles etc.\n\nWhenever you feel ready to finalize setup, either choose \"Configuration wizard\" in main menu or restart c:geo."

PR #10068, opened 2021-02-23
some polishing for map downloader

### LINKED ISSUE #7529: Support direct offline map download links
When recently discussing the current evolution of c:geo mapsforge support the maintainers of openandromaps suggested that c:geo should support a special url scheme to trigger download of map files from their site. Besides the orux and locus specific ones they do have 'generic' ones we could tie into. This sounds to me like an easier to implement first step for #6072

moving-bits:
some polishing for the map downloader:
- set downloaded Freizeitkarte (FZK) theme as current theme on successful download (last part to fix #7529)
- offer to download missing FZK theme after successful FZK map download
- make naming of maps' and themes' variables and classes consistent
- remember last selected type in map downloader


moving-bits:
retest this please

Issue #10070, opened 2021-02-24
[Nightly] Syc'ing themes folder to internal storage does not work


MagpieFourtyTwo:
**Describe the bug:**
After upgrading to the latest nightly this morning I got the wizard for changing the ``_themes`` folder from Legacy to SAF.
Did so, everything seemed to run smooth, the wizard sync'ed the folder to internal storage (as he reported in the popup),
but after opening the OSM live map, I do not see any themes at all anymore. And there should be plenty of ...
(about 16 or so, cause currently I experiment with different sizes for texts and symbols and stuff).

**To Reproduce:**
1. let the wizard switch your themes folder to SAF notation
2. open live map (default theme is shown)
3. try to switch to a differnet theme ...

**Actual behavior/state after performing these steps:**
Apart from ``default`` there are no themes at all.

**Expected behavior/state after performing these steps:**
My themes should be there as usual.

**Version of c:geo used:**
2021.02.24-NB-8f387a6

**Is the problem reproducible:**
Yes

**System information:**
```
--- System information ---
c:geo version: 2021.02.24-NB-8f387a6

Device:
-------
- Device type: VOG-L29 (VOG-L29EEA, HUAWEI)
- Android version: 10
- Android build: VOG-L29 10.1.0.150(C431E19R2P5)
- Sailfish OS detected: false
- Google Play services: disabled - 21.02.14 (120400-352619232)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Hide caches: own/found archived
- Hide waypoints: -
- Set language: en_DE (system default)
- System date format: dd/MM/y
- Debug mode active: no
- Live map mode: true
- Global filter: display all caches
- Last backup: 23 Feb, 22:49
- Routing mode: Car

Services:
-------
- Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / PREMIUM
   extremcaching.com: Logged in (Login OK)
- Geocaching.com date format: dd MMM yy
- BRouter connection available: true
- Installed c:geo plugins: contacts

Permissions & paths:
-------
- Fine location permission: granted
- Write external storage permission: granted
- System internal c:geo dir: /data/user/0/cgeo.geocaching (64,9 GB free) v1 internal isDir(6 entries)
[...]

PR #10078, opened 2021-02-27
rel to #10070: introduce user display name for map themes

eddiemuc:
rel to #10070: introduce user display name for map themes.

More details will follow in a comment shortly.

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
@@ -11,6 +11,7 @@ import cgeo.geocaching.storage.PersistableFolder;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.CollectionStream;
 import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.TextUtils;
 
 import android.app.Activity;
 import android.app.AlertDialog;
@@ -18,6 +19,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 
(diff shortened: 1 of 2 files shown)


eddiemuc:
This PR introduces "user-friendly" display of map themes in map theme selection list:
* XML Themes directly lying in map theme folder will be displayed by stripping away any subfolder and the .xml extension
* XML Thems from within ZIP files will be displayed same as above but with Zip file name + / prepended (both subfolder structures where the zip file lies and subfolder structures inside the Zip are stripped away).
* Result is ordered alphabetically.

To visualize this I created a very complex example map themes dir as follows:

image

This example contains:
* Xml themes in both the theme root dir as in a subfolder "subfolder"
* Zip theme packages in same folders
* A "special" zip package "mythemes.zip" in both root and subfolder. I created this special zip by downloading both Elevate theme from OpenAndroMaps and one Freizeitkarte Theme and putting them both into a single Zip file. E,g, inside this Zip file the following xml themes are located (including zip-internal subfolders):

```
mythemes\freizeitkarte-v5\freizeitkarte-v5.xml
mythemes\Elevate\Elevate.xml
mythemes\Elevate\Elements.xml
```

Given this setup, map theme selection dialog display is as follows:
image

To be able to debug situations like this, I let the "themeId" (which contains also all subpaths and extensions) be displayed as well **when in debug mode**. With above test setup this would look as follows:
image

image

moving-bits:
@eddiemuc 
After merging #10079 there is now a conflicting file in this PR - can you resolve & update this PR? Thanks.

eddiemuc:
@moving-bits done

BTW: same thing will happen again with either #10086 (if you merge this PR first) or this PR (if you merge #10086 first).

moving-bits:
Thanks for updating. Which way would be less hassle for you - start with merging this PR, and then #10086, or the other way around?

eddiemuc:
Good question. Maybe better to merge this one first

eddiemuc:
Btw: if you want I can also merge myself. Not so sure yet in which cases this is ok or not.

moving-bits:
> Good question. Maybe better to merge this one first

done

> Btw: if you want I can also merge myself. Not so sure yet in which cases this is ok or not.

We have no fixed rule regarding this. Personally I try to handle it like this: Trivial changes I merge myself, other changes I leave open for some time, if someone wants to take a look at it. And if I'm unsure with some more "critical" parts, I'll request a review.

MagpieFourtyTwo:
Looks perfectly fine. Just one hint regarding your test scenario; If you use it to just test the display of theme names, everything is fine, but if you additionally want to check if the themes really work in the map (e. g. while testing the folder sync), the ``subfolder\Elements.xml`` will most probably not, as OpenAndroMap theme .xml files need an ``ele_res`` folder besides it, otherwise they will not find their .svg files.

eddiemuc:
Yes this was just for testing the dialog :-) A little bit also for testing whether such situation is handled stable by c:geo (e.g. no unexpected crash).

MagpieFourtyTwo:
Is the sort already implemented?  If so, it does not work. If not, no problem, just want to be sure. ;)

eddiemuc:
@MagpieFourtyTwo 
It was implemented, then it fell out again due to a wrongly solved merge conflict (I do too many PRs in same file...). It should work now again.

PR #10085, opened 2021-02-27
brouter tile downloader

moving-bits:
First shot at an integrated tile downloader for BRouter. Still needs some polishing and extensive testing under different API constellations.

Prerequisites:
- BRouter needs to be installed
- Automatic tile downloading needs to be activated using Settings -> Maps -> Offline routing -> Automatic download
- Also path to the BRouter tile folder has to be set (something like `/brouter/segments4`)

Operations: (if prerequisites are met)
- If you try to calculate a route for which BRouter has no routing tile data yet, c:geo will show a box to confirm downloading the file.
- If you deny the download, you will not get asked for the same file until restarting c:geo
- If you accept, the download will happen in the background, and on successful receive, the routing tile data file will be stored in BRouter's `segments4` folder. On recalculating the route, the new data will be available.

image.image

(Nice side effect in settings: You can see now whether BRouter is installed or not, as options shown above will be greyed-out if BRouter is not installed.)

This seems to work nicely already, and downloading the tiles is pretty fast compared to using the original app, and non-blocking (except when doing the local copy).

~But still quite a few things are on the todo-list for this:~
More things done meanwhile:
- [x] Testing under Android 11
- [x] ... and under some pretty old API.
- [x] Downloaded tiles should overwrite existing tile data without extra confirmation.
- [x] Some strings still mentions "map download", some classes `OfflineMap` (should be `Download`) - needs to be generalized or individualized.
- [x] Some additional safety checks.
- [x] Update function.
- [x] Integrate in wizard (but show only when "automatic downloads" is activated and SAF path not yet set)

Ready to go from my side, removing the "WIP" label now.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -489,7 +489,7 @@
     <string name="settings_summary_map">Set map source, offline map settings and map details.</string>
     <string name="settings_title_map_data">Map Data</string>
     <string name="settings_title_offline_routing">Offline routing</string>
-    <string name="settings_summary_offline_routing">Offline routing is available if helper app \"BRouter\" is installed and configured.</string>
+    <string name="settings_summary_offline_routing">Offline routing is available if either helper app \"BRouter\" is installed and configured or internal routing is enabled.</string>
     <string name="settings_title_map_rotation">Map rotation (Google Maps)</string>
     <string name="settings_title_waypoints">Waypoints</string>
     <string name="settings_title_map_content">Map Content</string>
@@ -702,7 +702,6 @@
     <string name="init_base_directory_description">Base folder for public local data</string>
     <string name="init_map_directory_description">Folder with offline maps</string>
     <string name="init_brouter_directory_description">Folder with BRouter routing data</string>
-    <string name="init_brouter_directory_additionalinfo">(Default folder is something like \'%1$s\'. BRouter needs to be configured to use the same folder, see our online manual for details, especially when using Android 11 or higher.)</string>
     <string name="init_brouter_updateinterval_title">Update interval</string>
     <string name="init_brouter_updateinterval_description">Reminds you every x days for an update check</string>
     <string name="init_gpx_importexportdir">GPX Folder</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -605,7 +605,7 @@
     <string name="init_brouterThreshold_description">Maximum distance up to which the BRouter helper app (if installed) will calculate a route.</string>
     <string name="init_brouterShowBothDistances">Show straight distance</string>
     <string name="init_brouterAutoTileDownloads">Automatic download</string>
-    <string name="init_brouterAutoTileDownloads_description">Automatically offer to download missing routing tile data</string>
+    <string name="init_brouterAutoTileDownloads_description">Automatically offer to download missing routing tile data. BRouter data folder needs to be set in both c:geo and BRouter for this option to work.</string>
     <string name="brouter_recalculating">Recalculating route in background…</string>
     <string name="init_summary_brouterShowBothDistances">If routing is active: Show the straight distance in addition to the calculated route length.</string>
     <string name="init_useInternalRouting">Use internal routing</string>
@@ -702,7 +702,7 @@
     <string name="init_base_directory_description">Base folder for public local data</string>
     <string name="init_map_directory_description">Folder with offline maps</string>
     <string name="init_brouter_directory_description">Folder with BRouter routing data</string>
-    <string name="init_brouter_directory_additionalinfo">(Default folder is something like \'/brouter/segments4\')</string>
+    <string name="init_brouter_directory_additionalinfo">(Default folder is something like \'/brouter/segments4\'. BRouter needs to be configured to use the same folder, see \'storageconfiguration.txt\' in BRouter\'s \'segments4\' folder.)</string>
     <string name="init_gpx_importexportdir">GPX Folder</string>
     <string name="init_dataDir">Select data folder</string>
     <string name="init_dataDir_note">You may choose to store the additional data for geocaches (spoilers, log images, …) on your external storage (emulated or real external SD card depending on your device).</string>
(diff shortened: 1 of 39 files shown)


main/src/cgeo/geocaching/InstallWizardActivity.java
```diff
@@ -453,6 +457,18 @@ private void requestGpxfolder() {
         }
     }
 
+    private static boolean broutertilesFolderNeedsMigration() {
+        return Settings.isBrouterAutoTileDownloads() && PersistableFolder.BROUTER_TILES.isLegacy();
```
fm-sys, line 461: To be on the safeside, we should also check if BRouter is installed at all, shouldn't we?

main/res/values/strings.xml
```diff
@@ -703,6 +707,10 @@
     <string name="init_local_file_system">Local File System</string>
     <string name="init_base_directory_description">Base folder for public local data</string>
     <string name="init_map_directory_description">Folder with offline maps</string>
+    <string name="init_brouter_directory_description">Folder with BRouter routing data</string>
+    <string name="init_brouter_directory_additionalinfo">(Default folder is something like \'/brouter/segments4\'. BRouter needs to be configured to use the same folder, see \'storageconfiguration.txt\' in BRouter\'s \'segments4\' folder.)</string>
```
fm-sys, line 711: > Default folder is something like \'/brouter/segments4\'.

IMHO the most important point is, where the user can find this directory. `/Android/data/btools.routingapp/files/brouter/...`

And what about the `storageconfiguration.txt`? Why is it important or what does it mean? Yes, I found the file in the segments folder but I somehow don't understand it...

fm-sys (commented):
Great work, looking forward! 😁 I added two small comments...

main/res/values/strings.xml
moving-bits: This is strange: For some reason, BRouter used to store its data in `/brouter/` on some earlier device, and so I transferred this folder from device to device, never recognizing that on newer devices the folder has changed to `/Android/data/btools.routingapp/files/brouter/` :-(
I will change the default string accordingly.

On Android 11+ devices you cannot reach this folder any longer using SAF. You'll need to create a folder in public area to store the download data, and point both c:geo and BRouter to this folder. For that you'll have to edit the `storageconfig.txt` file in `segments4` folder. (But it's probably better to explain this in an FAQ entry instead of in the app.)

main/src/cgeo/geocaching/InstallWizardActivity.java
moving-bits: The option to activate auto downloading is greyed out when BRouter is not installed, so you can activate it only by importing a config file or via Settings => System => View Settings.

But anyway, better to add a test here.

main/res/values/strings.xml
Lineflyer: sounds like quite some effort and only for experts. Do we really want to implement such downloader which will not work unless you tweak external files manually? Or did I miss something...have to admit, that I did not follow the whole thread.

moving-bits: The downloader will work easily on systems up to Android 10 - you select the folder (same as you do with map folder etc.), and you are ready to go. 

More restrictive SAF handling in Android 11 makes things a bit more complicated, though. You'll have to tweak a folder setting in BRouter configuration then. Same thing that you had to do when you wanted to store the BRouter files on an SD card.

moving-bits: Anyway, as mentioned in the original issue, I'll try to contact the author of BRouter to see if he'll accept PR for BRouter. We could remove the download throttling in BRouter, and we could add a background downloading of tiles then, the same way as we do in c:geo now. My goal for this would be to trigger a download from c:geo in BRouter, and avoid the storage related settings in c:geo by this. But that's some dreams of the future...

moving-bits: Might be a misunderstanding - this PR uses SAF to access the routing data files, on both the receiving and the writing end. 

The "Android 11 restrictions" come from a different side: Starting with Android 11 you cannot access app-internal folders for a different app any longer, they are simply not listed any longer in the SAF selector. That's why in that case you need to configure BRouter to use a different folder, which is still reachable by other apps (using SAF). BRouter has a config file for doing exactly such a thing.

fm-sys:
I would also vote to not introducing new bigger features like this until we have started the SAF beta... This even allows us to contact the BRouter author and maybe patch the BRouter app before merging 😉

What do you think?

moving-bits:
> I would also vote to not introducing new bigger features like this until we have started the SAF beta... 

Agreed. That's why I have not merged this yet - waiting for SAF branching first.

> This even allows us to contact the BRouter author and maybe patch the BRouter app before merging 😉

Have already contacted him, am waiting for reply.

main/res/values/strings.xml
Lineflyer: Indeed, a misunderstanding :) I am relaxed now ;)

main/src/cgeo/geocaching/downloader/DownloaderUtils.java
```diff
@@ -0,0 +1,280 @@
...
+            temp = "default.map";
+        }
+        final String filename = temp;
+
+        final AlertDialog.Builder builder = Dialogs.newBuilder(activity);
+        builder.setTitle(title);
+        final View layout = View.inflate(activity, R.layout.mapdownloader_confirmation, null);
+        builder.setView(layout);
+        final TextView downloadInfo = layout.findViewById(R.id.download_info);
+        downloadInfo.setText(String.format(activity.getString(R.string.download_confirmation), StringUtils.isNotBlank(additionalInfo) ? additionalInfo + "\n\n" : "", filename, "\n\n" + activity.getString(R.string.download_warning) + (StringUtils.isNotBlank(sizeInfo) ? "\n\n" + sizeInfo : "")));
```
rubieHess, line 90: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

moving-bits: Thanks for pointing me to that info. Alas, it seems to not apply to Android apps and/or not in the context of dialogs: I've tried switching the string to using `%n`, but the placeholders get printed as literal `%n` in the dialog, so I reverted them to `\n`.

moving-bits:
Remaining CheckStyle issues are due to #10217 only, so let's merge here.

main/src/cgeo/geocaching/downloader/DownloaderUtils.java
rubieHess: noticed. thanks a lot!

Issue #10075, opened 2021-02-26
Unresponsive/Slow app when buffering themes on SD


Lineflyer:
After we do now have the "theme sync" I have to say, that c:geo is no longer nice to use on my Galaxy S20.
I currently have no clear idea why it is so slow compared to the old implementation.

I have stored the additional geocachedata on my SD, but first of all the SD should not be relevantly slower than the internal memory and additionally its no difference to before (I had the themes on SD also with the legacy implementation).

Have a look on the linked video: https://files.droescher.eu/public/Unresponsive_1.mp4

What I did:
- 2021.02.26-NB
- Migration was completed already
- Start c:geo
- Wait for ages until sync check complete
- Start live map, takes some seconds to start and some more seconds to show the map
- Tried to change between themes with same delay and unresponsiveness

I have to admit, that I have quite some files stored in my _themes directory. But I think there is some more generic problem here, as I would have expected it to be much faster in general.

I will do another test tonight or tomorrow using only 2-4 themes to have results for an average user.

Lineflyer:
```
--- System information ---
c:geo version: 2021.02.26-NB-a5a3de5

Device:
-------
- Device type: SM-G980F (x1seea, samsung)
- Android version: 11
- Android build: RP1A.200720.012.G980FXXU6DUB5
- Sailfish OS detected: false
- Google Play services: enabled - 21.02.14 (150408-352619232)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Hide caches: own/found
- Hide waypoints: -
- Set language: de_DE (system default)
- System date format: dd.MM.yy
- Debug mode active: no
- Live map mode: false
- Global filter: display all caches
- Last backup: 26. Feb., 13:49
- Routing mode: Fahrrad

Services:
-------
- Geocaching sites enabled:
   geocaching.com: Logged in (Anmeldung OK) / PREMIUM
   opencaching.de: Logged in (Anmeldung OK)
- Geocaching.com date format: d.M.yyyy
- BRouter connection available: true
- Installed c:geo plugins: contacts

Permissions & paths:
-------
- Fine location permission: granted
- Write external storage permission: granted
[...]

PR #10090, opened 2021-02-28
rel to 10075: load themes on startup, provide test mode for content themes

eddiemuc:
rel to #10075: load themes on startup, provide test mode for content themes

Also, while working on this I discovered a bnug in mapsforge ContentResolverResourceProvider: currently it will always root on the Tree root (not on the provided folder uri). Migitated this for now by copying class into c:geo and fixing error there, but needs to be fixed in mapsforge and then rerouted in c:geo as well.


Lineflyer:
While I did not review this PR codewise, let me merge it to allow testing in tomorrows nightly in regard of #10075.
If some adaption of fixing is needed according to others, please kindly open an issue for that.
We could also revert is necessary.

Issue #10095, opened 2021-03-01
Mass warnings: Cannot parse brouter output of length 29


Lineflyer:
While working on a support ticket (possibly unrelated to this report) I have seen mass warning with the following text in users logfiles while the system information even states `BRouter connection available: false`.

Not sure, whether it is a real problem.

Just in case:
Could it be related to the users problem, that clicking on map cache icons does not open the popup?

```
03-01 10:14:36.274  6977 11747 W cgeo    : [RxCachedThreadScheduler-1] cannot parse brouter output of length 29
03-01 10:14:36.274  6977 11747 W cgeo    : org.apache.harmony.xml.ExpatParser$ParseException: At line 1, column 0: syntax error
03-01 10:14:36.274  6977 11747 W cgeo    : 	at org.apache.harmony.xml.ExpatParser.parseFragment(ExpatParser.java:512)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at org.apache.harmony.xml.ExpatParser.parseDocument(ExpatParser.java:497)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at org.apache.harmony.xml.ExpatReader.parse(ExpatReader.java:317)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at org.apache.harmony.xml.ExpatReader.parse(ExpatReader.java:275)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at android.util.Xml.parse(Xml.java:57)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.maps.routing.Routing.parseGpxTrack(Routing.java:217)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.maps.routing.Routing.calculateRouting(Routing.java:210)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.maps.routing.Routing.getTrackNoCaching(Routing.java:181)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.models.Route.calculateNavigationRoute(Route.java:145)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.models.ManualRoute.toggleItemInternal(ManualRoute.java:133)
03-01 10:14:36.274  6977 11747 W cgeo    : 	at cgeo.geocaching.models.ManualRoute.loadRouteInternal(ManualRoute.java:99)
[...]

PR #10098, opened 2021-03-01
add BRouter installation state to system info (rel. to #10095)

moving-bits:


main/src/cgeo/geocaching/utils/SystemInformation.java
```diff
@@ -102,7 +103,7 @@ public static String getSystemInformation(@NonNull final Context context) {
         if (GCConnector.getInstance().isActive()) {
             body.append("\n- Geocaching.com date format: ").append(Settings.getGcCustomDate());
         }
-        body.append("\n- BRouter connection available: ").append(Routing.isAvailable());
+        body.append("\n- BRouter installed: ").append(ProcessUtils.isInstalled(context.getString(R.string.package_brouter))).append(" / connection available: ").append(Routing.isAvailable());
```
Lineflyer, line 106: Just to understand the difference;
What is the trigger, when `Routing.isAvailable` becomes true? The user actively uses it (by setting a navigation method other than straight) or something else?

Lineflyer (approved):

main/src/cgeo/geocaching/utils/SystemInformation.java
moving-bits: `Routing.isAvailable()` becomes true on connecting to the BRouter service, which gets triggered by opening the map (independent from routing mode).

PR #10102, opened 2021-03-02
fix #10101: avoid circular errors between Log and ContentStorage

### LINKED ISSUE #10101: c:geo not starting after restore of data
While testing other issues I ran into a problem after this scenario:

- 2021.03.01-NB
- Go to device settings - Apps - c:geo and delete all data for c:geo
- Start c:geo (comes up with wizard)
- Complete the wizard (permissions, base folder, login to geocaching.com, skip advanced wizard)
- c:geo works normally
- Go to c:geo menu - Settings - Restore and select to restore caches and settings (used a backup done with yesterdays nightly, settings did not include account data)
- Backup is restored and restart is triggered by dialog

Result:
- Grey screen is shown, after some time splash screen logo shows up
- App runs into ANR
- Consecutive tries to start c:geo end up in ANR

Not yer knowing the exact cause of this problem, at least the following came to my mind:
What happens to previously existing SAF persistent folders in such case? The settings are restored, but the permissions might be lost...or is this also covered by the backup?

eddiemuc:
fix #10101: avoid circular errors between Log and ContentStorag

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
@@ -464,7 +464,7 @@ public class RenderThemeHelper implements XmlRenderThemeMenuCallback, SharedPref
         public static MapThemeFolderSyncTask createAndExecute(final Folder source, final File target, final Consumer<FolderUtils.FolderProcessResult> callback) {
             final MapThemeFolderSyncTask task = new MapThemeFolderSyncTask(source, target, callback);
             task.execute();
-
+            Log.i("[MapThemeFolderSync] start synchronization " + source + " -> " + target);
             return task;
         }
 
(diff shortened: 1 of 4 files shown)


Lineflyer:
Shouldnt we split this in different PRs?

IMHO we do not necessarily need experimental logfile writing on master...but maybe others find it helpful?

eddiemuc:
@Lineflyer 
There will never ever be a log file written if there is not a log properties present and the necessary (and yet undocumented) property set. So there's no harm for it being on master, while on the other hand it will be available as a last resort. Also, there's no way getting this from experimental state to production state than testing it in the nightlies. It would probably have saved you a long waiting/testing time wrt the current issue if it had been in before. Can't say for sure of course as long as it's not available in master :-)

If you want a split nevertheless I can do it next time I am coding of course.

eddiemuc:
Removed added Logging functionality from this PR

eddiemuc:
retest this please

PR #10103, opened 2021-03-02
offer "open log in browser" also for trackables

fm-sys:
grafik

~The only flaw in my implementation is, that the title doesn't look very nice with this very long log id. But we can either leave it like this (hey, it's only about trackables ;-) ), remove the log id also for caches completely from the title or need to introduce some magic (-> e. g. only show id's shorter than 10 characters)~

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1139,7 +1139,7 @@
     <string name="cache_menu_challenge_checker">Search challenge checker</string>
     <string name="cache_menu_ignore">Ignore cache</string>
     <string name="cache_log_menu_decrypt">Decrypt/Encrypt</string>
-    <string name="cache_log_menu_title">Log %1$s of %2$s</string>
+    <string name="cache_log_menu_popup_title">Log of %s</string>
     <string name="ignore_confirm_title">Be careful</string>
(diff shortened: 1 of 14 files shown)


eddiemuc:
Iirc the log id in title was not so much wanted anyway, so I would suggest to remove it (maybe display log date instead?)

fm-sys:
> 
> 
> Iirc the log id in title was not so much wanted anyway, so I would suggest to remove it (maybe display log date instead?)

Done! I've decided to just remove the log id without adding the date instead. The reason is, that the date is already displayed outside the popup, so it would be an unnecessary and space-costing duplicate without any benefit...

fm-sys:
retest this please

fm-sys:
retest this please

fm-sys:
Don't know yet why the tests are failing. I will look into it, but no more today. Setting "do not merge" for now...

fm-sys:
OK, ready from my side. Should we include it to the SAF beta? It's IMHO in no way a critical change, but I'm also fine with waiting...

eddiemuc:
I would vote to include it. I don't see how it is related to SAF (so no disturbance to be feared) and it is a nice new feature

Lineflyer:
Nice addon, why not.

PR #10104, opened 2021-03-03
fix #9973: camera image lost on rotate, changed ActivityHelper design pattern to fix

### LINKED ISSUE #9973: Picture not attached to log when rotation involved
Found in 2021.02.06-NB, Samsung S20, Android 11:

- Hold phone in portrait mode
- Start logging a cache
- Write you log
- Typ icon to attach a new photo from Camera
- After camera has been opened, turn your phone to landscape mode (to make a nicer photo :) )
- Take the picture and confirm

Result:
You will return to c:geo logging view (now in landscape mode) but there is no photo attached

This happens as soon as the screen orientation of the logging view before and after invoking the camera is different.
Only if you return to the same screen orientation in the logging view, where you left it, it is working normal.

eddiemuc:
Fixes #9973 (camera image lost on device rotation), but to do that a deeper class desing issue had to be fixed. This fix is absolutely necessary because it may also produce strange problems in other areas of c:geo (which might not yet have been found), but it makes this a larger change than one would maybe expect when looking solely at the user.

Implementation detail: ImageActivityHelper was as of before not involved in Activity lifecycle and did not save/restore its state correctly. In case of rotation, the calling Activity (e.g. ImageListFragment or ImageSelectActivity) is internally destroyed and recreated, thus state got lost. This is the cause of the user-visible error.
Saving state of ImageActivityHelper was not straight-forward though because it contains callback lambdas which referred to calling activity which in current case were no longer there (also lambdas should not be serialized). Thus a larger design change was necessary.

### DIFF
--- a/main/src/cgeo/geocaching/ui/ImageActivityHelper.java
+++ b/main/src/cgeo/geocaching/ui/ImageActivityHelper.java
@@ -5,13 +5,16 @@ import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.models.Image;
 import cgeo.geocaching.utils.ImageUtils;
 import cgeo.geocaching.utils.Log;
-import cgeo.geocaching.utils.functions.Action1;
+import cgeo.geocaching.utils.functions.Action2;
 
 import android.app.Activity;
 import android.content.Intent;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.os.Bundle;
+import android.os.Parcel;
+import android.os.Parcelable;
 import android.provider.MediaStore;
 import android.view.View;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/ui/ImageActivityHelper.java
```diff
@@ -39,43 +41,89 @@
...
     private final Activity context;
+    private final Action2<Integer, List<Image>> callbackHandler;
 
-    private final int requestCodeCamera;
-    private final int requestCodeStorageSelect;
-    private final int requestCodeStorageSelectMulti;
+    private final Bundle runningIntents = new Bundle();
 
-    private final Map<Integer, IntentContextData> runningIntents = Collections.synchronizedMap(new HashMap<>());
+    //private final Map<Integer, IntentContextData> runningIntents = Collections.synchronizedMap(new HashMap<>());
```
fm-sys, line 55: Leftover?

fm-sys (approved):

main/src/cgeo/geocaching/ui/ImageActivityHelper.java
eddiemuc: Oh man, I am bad... It is good that you find these things! I remove the line and will force-push immediately.

eddiemuc:
Btw: many thanks to @fm-sys and @moving-bits for telling me about `git commit --amend` some days ago to use instead of `git rebase -i HEAD~2`. Makes my life a lot easier!

PR #10108, opened 2021-03-03
Fix error & Optimize MFMProviderTest

bekuno:
Move class to the correct place (fix package name error).

Change from using TestCase to junit.Test.
This is enough, the extended possibilities of TestCase are not used.


bekuno:
retest this please

PR #10111, opened 2021-03-05
adapt NoNumberIncrement to the new number log templates

fm-sys:


moving-bits:
@fm-sys 
I'm unsure what this PR does (but I'm not used to the log text placeholders), so can you help me? How would you describe this change / fix (?) in the changelog?

fm-sys:
Ok, the whole story:

Years ago there was introduced a Twitter plug-in, where you can define a log text which will be posted when a log is sent. Since the Twitter log templates are filled in only after the log entry has been sent, we do not count "+1" as normally. At the point, where I renamed the normal number template to "[ONLINENUM]" and introduced the new count logic for "[NUMBER]" I was not aware of this. So basically it could be called "minor bug fixes for the numbers template, when used by the Twitter plug-in"

moving-bits:
thanks, will add it to the changelog

PR #10118, opened 2021-03-06
add final coords to navigate-drive target selection (fix #10074)

### LINKED ISSUE #10074: Check on existing final coords before navigation is started
From support mail:

If I start navigation towards any cache (using Google Maps form c:geo navigation menu) c:geo asks me, whether I want to navigate to the cache coords itself or to the parking coordinates of the cache if existing. Thats great.

I do however miss this similar function on e.g. mysteries which have a final waypoint defined. This made me some times drive to the listing coordinates instead of the real final.

So I suggest the following:
If there is a final waypoint present and contain valid coordinates and there is **no** modified listing coordinate for the cache, c:geo should offer the very same selection providing the option to select the listing coordinates or the defined final waypoint (and also the parking coordinate if available) as navigation targets. Just reuse the same dialog and usage as for parking waypoints for this condition.

moving-bits:


main/res/values/strings.xml
```diff
@@ -1101,7 +1101,7 @@
     <string name="cache_dialog_favorite_remove_message">Removing cache from your favorites…</string>
     <string name="cache_menu_navigate">Navigate</string>
     <string name="cache_menu_navigation_drive">Navigation (Driving)</string>
-    <string name="cache_menu_navigation_drive_select_target">Cache or parking place?</string>
+    <string name="cache_menu_navigation_drive_select_target">Select target?</string>
```
moving-bits, line 1104: arg, I should have removed the question mark...

PR #10119, opened 2021-03-06
reset circle color in NewMap on color change in settings (fix #10114)

### LINKED ISSUE #10114: Circle color setup effective only after restart (OSM)
Reported by a user of 2021.02.14 and reproducible on 2021.02.27-NB

- Have circle color in default state
- Open OSM map
- Activate circles
- Go to Settings - Map and change circle color(s)
- Open map again

Circles stil shown in old color
After restart circles are shown in new color

(On GMaps it seems to work better but might require turning circles off and on again, not clearly reproducible)

moving-bits:


bekuno:
the bug is also on release.

moving-bits:
Haven't checked explicitly, but I think it has always been coded that way. It only turned out as a "bug" since we are able to change color for the cache circles. I could prepare a fix based to release, but IMHO this is only a minor glitch, for which a fix on master should be sufficient.

bekuno:
I can live with the work around. :-)
Then let it so.

PR #10120, opened 2021-03-06
respect showCircles setting on switch from NewMap to GM (fix #10115)

### LINKED ISSUE #10115: Active circles not shown on change of map type
2021.02.27-NB (Android 9):

- Have OSM active
- Activate circles in map quick settings
- Change to GMaps

Circles not shown although still active in map quick settings
Only after deactivating and activating the setting the circles are shown

moving-bits:


bekuno:
the bug is also on release.

moving-bits:
I can prepare this PR for release, but actually this is a pretty old bug - don't think that we need to fix this on release, IMHO.

bekuno:
> 
> 
> I can prepare this PR for release, but actually this is a pretty old bug - don't think that we need to fix this on release, IMHO.

I can live with the work around. :-)
Then let it so.

PR #10136, opened 2021-03-12
Request SAF permissions on settings backup restore (fix #10105)

### LINKED ISSUE #10105: Restore SAF permissions on settings backup restore
Follow-up of #10101:

Right now we assume that existing permissions for persistent folders are lost on app deinstallation / app data reset. This would mean, that if a user reinstalls c:geo and/or restores a settings backup, we will have the following situation:
The (restored) settings contain values for persistable folders which are no longer usable/connected/permitted.

We should first check if the assumption of lost permissions is correct (which is likely).
If the assumption is correct, we should find a suitable solution for such cases, which could IMHO be that we check while restoring settings, if the current URI are equal to the URI contained in the settings backup or not.
If they are the same its most likely, that there are still valid permissions behind, if not (means they are empty or different) we probably need to trigger the wizard or another clever way to let the user decide what to do.

moving-bits:
On settings restore: Check if any of the `PersistableFolder` settings in the to-be-restored file differs from current settings and request access grant for those.

For every `PersistableFolder` with a different setting a popup will show while restoring settings, asking the user to either keep the existing value or restoring the old value, and starting SAF grant process in the latter case:

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -694,6 +694,7 @@
     <string name="init_backup_settings_restore">Restore settings</string>
     <string name="settings_readingerror">Error reading settings</string>
     <string name="settings_restart">c:geo needs to be restarted now to bring loaded settings into effect</string>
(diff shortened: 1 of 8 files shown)


Lineflyer:
sounds good!

eddiemuc (approved):
Looks good

Issue #9081, opened 2020-09-27
Refactoring Filtering/Searching: Discussion 


eddiemuc:
This item is meant as a discussion item on how to refactor Filtering/Searching function in c:geo

@moving-bits was so nice as to create a new project for this topic: https://github.com/cgeo/cgeo/projects/16 

# Features

I extracted the following summary for wanted features from above project:
* Overall: 
  * regardless from where List is accesses, filtering behaviour and av. options should be the same (e.g. from stored lists, from map, from address search) #3346
  * filter options must be aware on what is possible depending on whether caches are transient, in db or only partly loaded (e.g. #5222)
* UI: better UX wanted, see e.g. Locus (#5877)
* Filtering: 
  * selectable value list for a filter shall only contain values which are present in list (e.g. country, see #8716; or #8273)
* New filters: 
  * caches which user favorited (#8377)  
  * caches found by a friend (=another user) #6511
  * by text (#5056)
  * negation filter (#2508)
* Performance / Stability:
  * Handling of large lists is not stable (slow or even crashing) e.g. #8343, #9015, #5523
  * filters shall run asynchronoulsy #5860
* Some bugs 
  * filtering for attributes does not work properly #8620
  * when coming from map then "hide own caches" can't be removed #8383
  * when opening cache from list, deleting it and then returning, cache is still shown #7540

# Technical solution

Today, filtering on list views are done completely in-memory, so even very large stored lists have to be loaded before filtering can be applied. Filtering is done using implementors of the `IFilter ` interface
  
As a starting point for discussion I propose the following:
* UI/UX:
  * A new filter dialog shall be implemented, designed nicely (maybe like proposed here: #5877)
  * This filter dialog is opened both when selecting in the Map "Hide options" (Verbergen) and when clicking "Filter" symbol in cache lists (any other place where filtering takes place?)
* Filter implementation:  
	* Each Filter implementation (`IFilter`) returns on which "type" of cache-data-availability it can be applied on. Types are "StoredInDB", "TransientPartlyLoaded" and "TransientFullyLoaded"
		* Using this, for each list (on map, as search result, as viewing stored lists) it can be filtered which filters may be applied to this list. In the common filter dialog, non-appliable filters are grayed out or sst similar
[...]

PR #10147, opened 2021-03-13
rel to #9081: introducing advanced filtering, sorting and limited database loading 

eddiemuc:
rel to #9081: first playing with filters

These are first steps towards #9081. It shall now be checked in!

Idea is a technical base framework where filters can be hierarchically constructed using AND, OR and NOT. They can also be saved and retrieved from (string) configurations.


eddiemuc:
This PR has now reached a stage where I would need feedback before I want to continue.

This PR, if merged, will add a new option "Cache Filter (Debug)" to the Cache Lists. This will only be added if "Debug Mode" is turned on:

image

If clicked, this will open the new "CacheFilterActivity" where a new filter can be configured. It looks like this:

image

It currently contains an odd collection of filters to add, not nearly complete but I think enough to make the concept testable:
image

**The GUI is FAR from being finished** (so for now please be gentle here...) but imho also enough the test the concept.

New filters can be added by clicking on "<click here to add filter>" Chosen filters can be removed (trash can symbol) or sorted ("sort-drag" symbol). Overal options are currently:
* Use OR instead of AND: self-explanatory I hope
* Invert filter: self-explanatory I hope
* Include Inconclusive: in some situations not all geocache fields are filled (e.g. when opening a list from map viewer, then personal notes are not loaded). If I have a filter on such a field in this situation, filter can't really decide whether the cache passes this filter or not. If this checkbox is set, then such caches ARE displayed in the result, if it is not set then they AREN't (try this out by using the "Personal Note" filter on both an Offline Stored list and on a list opened from map viewer for unstored caches)

If applied, filter is used on the cache list result e.g.:

image

eddiemuc:
I would like to put to discussion to merge this into master to get feedback from nightly users (those who are willing to turn on debug mode).

fm-sys:
Wow, I really like your work! Also the UX concept is pretty smart and IMHO easy to use (based on your screenshots, hadn't time to test it in real life). Only the design can/should be optimized, but as you wrote that's something for later on.

> those who are willing to turn on debug mode

Why should it dependent on debug mode? I would really like to see it for everyone in the nightlys. Simply call it "Filters (experimental)" and everything is fine. Nightly users want to test new alpha features (that's why they are nightly users ;-) ) so why are you hiding your great work?

Furthermore, I have not yet understood the advantage of resorting the filters. Sorry for the stupid question, but can you give me a hint?

eddiemuc:
> Wow, I really like your work! Also the UX concept is pretty smart and IMHO easy to use (based on your screenshots, hadn't time to test it in real life). Only the design can/should be optimized, but as you wrote that's something for later on.
> 

Thanks :-)

> Why should it dependent on debug mode? I would really like to see it for everyone in the nightlys. Simply call it "Filters (experimental)" and everything is fine. Nightly users want to test new alpha features (that's why they are nightly users ;-) ) so why are you hiding your great work?

Well this would be possible too. I remember though that this was not wanted in the past.

Among other things I didn't yet put resources into strings.xml for now because I expect much design changes and didn't want people to do unnecessary translation work already now.

> Furthermore, I have not yet understood the advantage of resorting the filters. Sorry for the stupid question, but can you give me a hint? 

I just included this for optics and if users prefer to sort to their liking. Otherwise they are stuck with the "add" order. And it was really easy to add because it is already contained in the base classes I use. 
Anyway, I am not insisting on it.

moving-bits:
> This PR has now reached a stage where I would need feedback before I want to continue.

Really like it! Thanks for your great work on it, will become a real improvement to filter handling in c:geo.

> This PR, if merged, will add a new option "Cache Filter (Debug)" to the Cache Lists. This will only be added if "Debug Mode" is turned on:

Why do you restrict it to debug mode? It has a "(Debug)" in it's name (could be "experimental"), so the nightlies' users should be warned ;-) And if you restrict it to those having debug mode activated, you'll probably won't get enough feedback in reasonable time.

Some quick ideas/feedback regarding the UI (I know that this is WIP yet, so this is mainly a collection of thought): (haven't had the time to install the APK yet, so this is from looking at the screenshots only)
- The displayed filter in the result set is very technical (the `-|AND(name:Hamburg:CONTAINS:false;type:MYSTERY:M...` line). Is there a way to make this line more understandable/readable for the end-user?
- Where are new filters added to in the list - at the top or at the end? As the action buttons are on top of the page, displaying newly added filters should be at the top, too, IMHO.
- "Clear filter" could be reduced to the trashcan symbol, "Add filter" to the "plus" symbol we use in other places for it. By this "Apply", "Delete", "Add" would fit into the same line to have more filter settings visible.

Anyway, looks really promising already! Looking forward to it...

moving-bits:
Played around with it a bit this morning - I have to say, I'm impressed!
It's easy to use, it's pretty fast (tested with app. 4000 caches in the emulator), you can combine conditions, it's mostly easy to handle. You get used to it pretty fast...

Some observations:
- If I create a type filter, select "Mystery" and then "inverse filter" I would have expected to see all caches which are not mysteries - instead I get an empty list (although app. 80% of the caches in the test list were not mysteries)
- If I add a filter, the new filter gets appended to the end of the list, which is ok. But then the action buttons should stay at the end as well to not have to scroll around for adding another filter. (Relevant if you have more than one or two filters, especially if you have some of the longer ones like type filter.) - Or append the newly created filter at the top of the list and keep the buttons there. - Or make the buttons fixed and scroll only the filter part.
- It's possible to add the same type of filter multiple times - intended?
- For personal notes a filter like "has" or "has not" would be nice
- Haven't been able to make the pattern filter option work. My fault, but maybe some short info about the expressions supported would help.

Please continue implementation... ;-)

And, yes, I would vote for integrating it into current nightlies to get more user feedback and to prioritize which filters to implement.

eddiemuc:
> Played around with it a bit this morning - I have to say, I'm impressed!
> It's easy to use, it's pretty fast (tested with app. 4000 caches in the emulator), you can combine conditions, it's mostly easy to handle. You get used to it pretty fast...

Thanks :-)

> Some observations:
> - If I create a type filter, select "Mystery" and then "inverse filter" I would have expected to see all caches which are not mysteries - instead I get an empty list (although app. 80% of the caches in the test list were not mysteries)

This must be a bug in my code. It should be exactly like you expect. I have to investigate this.

Does this happen for Offline Lists? In that case I would be interested in the generated SQL (is currently written to Log at warning level to ease my work...)

> - If I add a filter, the new filter gets appended to the end of the list, which is ok. But then the action buttons should stay at the end as well to not have to scroll around for adding another filter. (Relevant if you have more than one or two filters, especially if you have some of the longer ones like type filter.) - Or append the newly created filter at the top of the list and keep the buttons there. - Or make the buttons fixed and scroll only the filter part.

Good point, I will change it such that filters get added on top.

> - It's possible to add the same type of filter multiple times - intended?

Yes. That way you may create for example a filter which selects caches whose name starts with "Hamburg" and contain the word "blubb"

> - For personal notes a filter like "has" or "has not" would be nice

Good idea, easy to add

> - Haven't been able to make the pattern filter option work. My fault, but maybe some short info about the expressions supported would help.

Unfortunately this is all not documented yet. In short: You can use * and ? In the pattern as placeholders for multiple or single characters. Match is done against whole string, so you have to use *at beginning and end if you want to do a "contains"

eddiemuc:
Btw: "framework" is also work-in-progress. I plan to implement also a new setting to restrict the number of caches initially loaded/displayed in lists for offline Lists. Default maybe 1000. That way, if a user has e.g. >20000 caches in a list (e.g. the "all" list) it would not take 5-10 seconds to open it any more

This is the main reason why I introduced also SQL to the filtering framework (so people get those caches out of the 20000 which match their filter). However, I also need to include "sort" into it, so people restricting their list length as above will get the "top sorted" out of their 20000 caches.

moving-bits:
> > * If I create a type filter, select "Mystery" and then "inverse filter" I would have expected to see all caches which are not mysteries - instead I get an empty list (although app. 80% of the caches in the test list were not mysteries)
> 
> This must be a bug in my code. It should be exactly like you expect. I have to investigate this.
> 
> Does this happen for Offline Lists? In that case I would be interested in the generated SQL (is currently written to Log at warning level to ease my work...)

Yes, I had imported some stored GPX file and run the filters against it.
SQL in the log is this: `SELECT geocode FROM cg_caches t WHERE (geocode IN (SELECT geocode FROM cg_caches_lists WHERE list_id =939)) AND (NOT(((1=1))))`

> > * Haven't been able to make the pattern filter option work. My fault, but maybe some short info about the expressions supported would help.
> 
> Unfortunately this is all not documented yet. In short: You can use * and ? In the pattern as placeholders for multiple or single characters. Match is done against whole string, so you have to use *at beginning and end if you want to do a "contains"

I had tried something like this `[a-k]*.*` (so sort of "start with any letter between a and k") in the owner field. But I haven't tried much in this place, so no need to worry here yet.

But it leads to me the question which "type of patterns" is supported? I was thinking of regular expressions (my default assumption as a programmer), but `*` and `?` looks more like file system pattern. Which is totally fine, and probably much easier to handle for the end user, just hadn't expected it.

eddiemuc:
> Yes, I had imported some stored GPX file and run the filters against it.
> SQL in the log is this: `SELECT geocode FROM cg_caches t WHERE (geocode IN (SELECT geocode FROM cg_caches_lists WHERE list_id =939)) AND (NOT(((1=1))))`

Ok I see where my error lies... I currently use the "type" filter as a test field on whether the filtering behaves correctly also for filters not supplying any SQL. For those I add "1=1" as a "always-true" SQL expression.

But if negated then this will become "always false" which results in zero results. 

Not so easy to solve, I have to think about this
 
> I had tried something like this `[a-k]*.*` (so sort of "start with any letter between a and k") in the owner field. But I haven't tried much in this place, so no need to worry here yet.
> 
> But it leads to me the question which "type of patterns" is supported? I was thinking of regular expressions (my default assumption as a programmer), but `*` and `?` looks more like file system pattern. Which is totally fine, and probably much easier to handle for the end user, just hadn't expected it.

I trid to restrict to patterns which can also be expressed in SQL, as far as I know SQLite does not support regex

eddiemuc:
Have an idea wrt filtering problem: if inside a NOT expression, the "always true" has to be turned to "always false"...
But this means I have to do some redesign of the SqlBuilder - might take some time.

Unfortunately now I have to do my taxes instead of fixing this...

eddiemuc:
"NOT" on type filter should work now. Also, new filters are now added at top and filter is available also if not in debug mode

moving-bits:
> "NOT" on type filter should work now. Also, new filters are now added at top and filter is available also if not in debug mode

did a quick test, works like a charm. Thanks for the fast update!

jakedot:
would it be possible for this activity to eventually (maybe additionally to availability in lists and maps?) replace the global "cachetype" filter popup?

eddiemuc:
> would it be possible for this activity to eventually (maybe additionally to availability in lists and maps?) replace the global "cachetype" filter popup?
> 

Could be done. But my first goal is to get this "live" at all :-)

I am not using global filter myself, can anyone enlighten me what is the use case for this and whether this is really used?

eddiemuc:
Short heads-up:
image

image

image

Additionally I introduced a configurable load limit for DB queries. In the following screenshot this is set to "7" (just to show the feature, normally it would be set to e.g. 1000 or more):

image

moving-bits:
Having done a quick tour with today's `master` I can only bow my head - very promising! Thanks @eddiemuc for diving into those long-standing issues and for finding such a nice solution!

MagpieFourtyTwo:
Just took a glance ... playing around for a while ... but all that still comes to my mind is ... WOW.  Amazing. I'm flabbergasted.

...

Nevertheless, following Wilhelm Busch ;) there is of course already a new idea, borne while playing with it: Now that I have constructed a really complex filter (to BTW find exactly what I was looking for - REALLY great!) I would love to be able to save this painstakingly constructed filter in e. g. a personal filter set, Or to ex/import the SQL statement ...  or copy/paste it, to e. g. store it in a text file ... :)

Would be great if you would put this on the wish list ...  ;)

However, I'm *really* glad with the way this is looking right now. Great work! Congrats! My hat is off to you.

MagpieFourtyTwo:
``(Not) Found by <nickname>`` would be great, too - although it would of course just be able to filter for locally stored logs.
But as ``Friends/Own Logs`` usually contains the interesting stuff anyway, this would be perfectly fine, IMHO.  ;)

eddiemuc:
Hi @moving-bits and @MagpieFourtyTwo 

Thanks for the thanks :-)

To answer the questions:
* This is supposed to be the first step. The plan is to completely replace the existing filter, but since the new filter is not functionally equivalent yet (e.g. favorites, distance etc is still missing) I left the old filter for now.
* It is relatively easy now to add new filters to the set. More will come for sure, definitely Favorites, distance and some other things will be among them
* Saving/loading filters: had the same idea and everything is ready under the hud for it (specifically, a filter configuration can always be stored and parsed from a single string). My main Problem here was: had no idea yet how to integrate this into the filter GUI without disturbing the "casual filterer" too much. Any ideas?

In general, it would be great if wishes would be posted as new issues and assigned to the "Filter/Searching" project. I will work through those. Have to ask for a bit patience though

And also @MagpieFourtyTwo : under the hood, filtering and sorting is now SQL-based for offline Lists. Also I introduced an initial load limit (default to 1000, can be changed in settings). This should hopefully save the list load performance Problem you experienced in the past. Would be great if you find the time to check this. Note that I added no new indexes to cache table yet, we might so this later for e.g. most-filtered/sorted attributes if we run into performance problems.

MagpieFourtyTwo:
> And also @MagpieFourtyTwo : under the hood, filtering and sorting is now SQL-based for offline Lists. (...) 
> Would be great if you find the time to check this.

WFM - great work! Thx!

Issue #9914, opened 2021-01-30
Integrate BRouter tile downloading


moving-bits:
Downloading routing data using BRouter app is pretty slow, as the app limits the download rate. The server itself does not have this limitation.

Proposal:
If we get the ok from the author, we could use our integrated downloader for downloading and updating the routing tiles. We could even integrate an optional download of missing routing data, if a user tries to calculate a route for which we don't have the routing data yet.

@Lineflyer 
As you have been in contact with the author in the past: Could you ask him whether we are allowed to access his routing data server directly from within c:geo? Thanks.

moving-bits:
@eddiemuc

Thinking about the SAF side of this: Would it be possible at all?

I'm not sure where BRouter stores its routing data, whether this is an hidden/internal folder or public folder. I guess that an app-specific internal folder would be unreachable for us, right?

But if it's a public folder, we could gain access to it (by asking the user for an SAF read/write grant) and store routing tile data there, and BRouter would be able to see and to read those files, correct?

Lineflyer:
Server usage permission has been addressed to the Brouter author (Arndt).

Lineflyer:
BTW:
The author said in an earlier mail, that we could theroretically also include brouter into c:geo:

> > Mit c:geo nutzen wir derzeit das Routing in Verbindung mit der Brouter App. Alternativ
> wäre natürlich auch eine direkte Integration möglich. Ist die BRouter-Lib als OpenSource verfügbar?

> Ja schon, und wir haben extra die Lizenz von GPLv3 auf MIT geändert, damit das möglich ist.
Es ist eben nur keine sauber strukturierte Lib mit einer API, sondern wer das machen will muss sich vom HTTP-Server (also der Gegenstelle für BRouter-Web) oder der Mobile-App die API reverse engeneeren. So haben sie es bei Locus gemacht.

eddiemuc:
> @eddiemuc
> 
> Thinking about the SAF side of this: Would it be possible at all?
> 
> I'm not sure where BRouter stores its routing data, whether this is an hidden/internal folder or public folder. I guess that an app-specific internal folder would be unreachable for us, right?
> 
> But if it's a public folder, we could gain access to it (by asking the user for an SAF read/write grant) and store routing tile data there, and BRouter would be able to see and to read those files, correct?

Should be possible imho. Would be pretty similar to using shared offline map files I guess. Only Problem I could think of right now are performance issues in case files are too many (like with map theme resources).

[...]

PR #10156, opened 2021-03-13
Brouter integration [outdated, just for easy reference, temporarily]

moving-bits:
This is a first draft for integrating BRouter into c:geo - it's not intended for merging yet, only for internal testing - and also because I need some major help with SAF... (@eddiemuc, please see below)

(There's still quite some work to be done, e. g. regarding adjusting code style, removing unused code or code not necessary in the context of c:geo etc., but the main function - routing! - does work already. Please ignore any Codacy or CheckStyle errors for now... The point here is about SAF-migration - sounds familiar? :-))

For testing you need to do some preparations. (The configuration part will become much smoother later, without manual adjustments - just wanted to get things up and running ASAP first.) Here's how to start:
- create a new folder beneath c:geo base folder, called `routing`, and a subfolder within it, called `segments4`
- then you need the files from an existing BRouter installation from their private folder in '/Android/data/btools.routingapp/files/brouter/...`:
  - copy the files from their subfolder `profiles2` to the newly created folder `routing` (*.brf, lookups.dat)
  - copy the files from their subfolder `segments4` to the newly created folder `segments4` (*.rd5, storageconfig.txt)
- after this step you can (but don't have to) uninstall BRouter - c:geo won't need it any longer

Folder structure of c:geo will look similar to this after those steps:
grafik

@eddiemuc 
I could need your help with SAF migration: I've prepared `PersistableFolder` for `ROUTING_BASE` and `ROUTING_TILES` and tried switching the internally used brouter-related variables from `String`/`File` to `Folder`. Somewhere deep within that process I got stuck at trying to switch the inner workings of  `PhysicalFile` which seems to be some sort of a random access file with an associated cache. I have no idea how to do this in the context of SAF...

How to find the relevant files:
- all brouter classes are in a folder `brouter` within `main/src/cgeo/geocaching`
- entry point for the routing service is class `BRouterService` called from `maps/routing/Routing`
- the service creates a `BRouterWorker` which creates a `BRoutingEngine`
- the engine opens a `mapaccess/NodeCache` which tries to create a cache with `mapaccess/PhysicalFile`

(I've reverted the changes in brouter files regarding SAF, to keep it compilable and running for now.)

Can you give me hints on how to proceed with the file cache / routing tile file access?


eddiemuc:
I can hopefully look into this tomorrow

eddiemuc:
I only looked shortly into the PhysicalFile just now (on mobile screen), but my first thought is that this might not be possible with SAF since we never get a File handle out of it. Maybe it is possible to load the data into an in-memory array and do the necessary operations there instead. 

It seems to me that some parts of the code are dealing with checksum verification etc. Maybe we can even live without that if it is too hard to do it with SAF.

SammysHP:
Do we have a discussion issue for this? I always thought brouter is intended as a separate service app that handles routing requests from all other apps.

moving-bits:
@eddiemuc 
Thanks for having a first shot at it. I guess loading a whole rd5 file into memory won't be a viable option, as some segments are pretty huge (e. g. 150 MB). But I'm not that deep into the BRouter code yet as to decide whether this is needed at all.

I chose `BROUTER_TILES` to be a public path to be able to share the routing tile data between apps. Besides the BRouter app itself, at least Locus is also trying to integrate BRouter into their app. But public path means SAF very soon...

@Lineflyer 
Maybe we can ask menion how they tackle the SAF migration of BRouter?
Maybe they can share some ideas, or even disclose / donate some code?

An alternative could be to not share the tile data and simply store it in a c:geo private folder - which would be possible using the `File` interface. Probably the easiest solution, but no sharing of tile data then. (But sharing tile data only makes sense, if other apps are prepared to share the tile data as well, so they need SAF migration => sort of a circular reference...)
For my typical use of BRouter this would make no difference, as I'm using BRouter solely with c:geo. But even if I had e. g. Locus over here, using it's own tile data, this would mean app. 500 MB of duplicated tile data. I could live with that, having enough free "disk" space, but that might not be true for users with older devices.

I'll skip the routing tile data for now and continue with the config files, as having those accessible in a public folder would be a good thing anyway.

moving-bits:
> Do we have a discussion issue for this? I always thought brouter is intended as a separate service app that handles routing requests from all other apps.

We have an issue about auto-downloading routing data for BRouter (#9914). First try (implemented with #10085) handles this, but gets a bit complicated starting with Android 11, as the users have to adapt a BRouter-internal config file. Possible, but we are looking for an easier solution.

So I got into contact with the author of BRouter to ask whether we can enhace BRouter and integrate our downloader over there. He sort of refused due to very limited time on his side and fearing increased support takss for him, but offered us to integrate BRouter into c:geo - the same thing Locus is doing currently.

eddiemuc:
@moving-bits over night I had a thought:

 Mapsforge kind of has the same challenge, because I assume they need random access on  offline map files (don't think they read them into memory :-)), and they work with a FileInputStream (which you can get from SAF)

We might take a look in mapsforge code, maybe the solution is as simple as replacing `RandomAccessFile` with a `FileInputStream` or create necessary objects out of it.

eddiemuc:
@moving-bits looking into class "MapFile" of mapsforge, I think "PhysicalFile" can (and should) be rewritten to use "FileInputStream/FileChannel" and "ByteBuffer" instead of the "RandomAccessFile". Interface will look slightly different afterwards, but I believe anything needed can be achieved as well.

How do you want to proceed here?

moving-bits:
Ok, down to 713 Checkstyle / 800 Codacy warnings in the last build - coming from more than 19000 each after the first integration steps... 
More to be fixed next week.

As discussed in today's meeting:
As soon as I have a first version with not too many remaining Checkstyle/Codacy issues, I will create a new (temporary) branch `brouter` in c:geo's repo for the collaboration of @eddiemuc and me on the SAF update for this.

moving-bits:
Phew, down to 103 Codacy and 10 CheckStyle. I guess that's as far as it's feasible to fix for now. Let's first do the necessary adaptions to SAF etc., and then see which relevant errors are left to be fixed.

BTW: "Love" messages such as this: `The method 'encodeMicroCache(byte)' has an NPath complexity of 2147483647` (sic!) `, current threshold is 200`

eddiemuc:
> BTW: "Love" messages such as this: `The method 'encodeMicroCache(byte)' has an NPath complexity of 2147483647` (sic!) `, current threshold is 200`

Yeah, never understood exactly how this is calculated :-)

I would argue that for taking over code it might be ok to have some warning suppressings.

SammysHP:
It's just max int32 - probably the tool got confused by some recursion.

fm-sys:
What's the difference of `ROUTING_BASE` and `ROUTING_TILES`? As granting folder access via SAF is relatively much effort, I wonder if two different folders are really necessary...

moving-bits:
> What's the difference of `ROUTING_BASE` and `ROUTING_TILES`? As granting folder access via SAF is relatively much effort, I wonder if two different folders are really necessary...


Short answer:
- ROUTING_BASE is for config files; no additional grant needed, as this will be a subfolder `routing` of c:geo BASE folder
- ROUTING_TILES is for the tile data, which can be up to several GB (if you download them for the whole world). This is a folder potentially to be shared with other apps, e. g. Locus (or even BRouter if you like to), so this is to be included in the wizard.

Longer answer:
A regular BRouter installation uses several folders with different config files, and additionaly a tiles folder. On migrating this to c:geo I decided which config file is most probably needed by c:geo, and which isn't. In the end I stayed with the routing profile config files (`*.brf`), the tile data files (`*.rd5`) and one more file of (so far to me) unknown usage, which BRouter needs (`lookups.dat`). (Other files are already removed or are on my "can be removed" list.)

Tile data files vary widely in size, from a few MB to > 200 MB each, summing up to about 7 GB if you want to cover the whole world :-). Each tile covers an area of 5 degrees latitude/longitude each, so for covering Germany you need at least four tiles, summing up to app. 500 MB. This is why I think this data should be shared between apps, as we do it with the OSM offline map files. (At least Locus is also working on BRouter integration, although I don't know yet whether they also plan to share tile data, and you could even share this data with BRouter itself, which is the approach in #10085.)

Config files are pretty small, summing up to less than 100 kb, so I chose to put them into a predefined c:geo-specific subfolder, with no need for special SAF grants.

IF (and that's a big "IF", IMHO) we decide NOT to share any tile files, we can either skip SAF grant and put it into a subfolder of c:geo BASE, or we can even decide to store those files in c:geo app-internal folder only, which would mean that we could skip the whole SAF adaption, at the price of having that files in a fixed and limited storage compartment.

fm-sys:
Sharing the tile data definitely makes sense. I more wonder about the routing profile config files, which IMHO doesn't give any value to the user but pollutes the external storage. So either use the same public folder for both (definitely not my favorite) or store these config files internally which also simplifys SAF migration...

eddiemuc:
@fm-sys those would be new folders, and if they default to subfolders of BASE there is no migration and no user action necessary for them to work.

moving-bits:
> I more wonder about the routing profile config files, which IMHO doesn't give any value to the user 

I don't know how many users actually do configure their routing profiles, but on a quick glance I did not want to throw the possibility over board yet. But I'm open to discussion here. We can for sure hard-code / integrate those default profiles into c:geo if we do not see any use for the end users in being able to configure that.

That would leave the `lookups.dat`. I need to take a deeper look at the source code what this file is about, and whether this needs to be a physical file in the file system (which would require it to be stored somewhere), if it could be provided internally, or isn't needed at all. => Is on my todo list.

> or store these config files internally which also simplifys SAF migration...

No migration needed. If we need actual files, we need to provide defaults for it with c:geo anyway, so I would add them to c:geo's assets in this case, and write it to the above mentioned folders on first use.

fm-sys:
Yes, I know. Just want to keep the directory structure clean. As we have stored the themes in a subfolder of %maps%, could this be a possibility for routing data as well? As it's pretty much the same kind of data. Large files, needed for live map.

moving-bits:
Superseded by #10217

Issue #10155, opened 2021-03-13
Internationalization error in some plurals


Lineflyer:
Checking lint issues I found some occurrences of warning like this:

The quantity `'one'` matches more than one specific number in this locale (1, 21, 31, 41, 51, 61, 71, 81, 101, 1001, …), but the message did not include a formatting argument (such as `%d`). This is usually an internationalization error. See full issue explanation for more.

<p>Plural strings should generally include a `%s` or `%d` formatting argument. In locales like English, the `one` quantity only applies to a single value, 1, but that&apos;s not true everywhere. For example, in Slovene, the `one` quantity will apply to 1, 101, 201, 301, and so on. Similarly, there are locales where multiple values match the `zero` and `two` quantities. In these locales, it is usually an error to have a message which does not include a formatting argument (such as &apos;%d&apos;), since it will not be clear from the grammar what quantity the quantity string is describing.

In most cases it is caused by using an unnumbered value as `quantity=one` as it fits in English.
Example: The string `days_ago` uses 'Yesterday' for `quantity=one`. This will however make the translation of 'Yesterday' show up also if it was 101 days ago in some languages.

I will fix some of those which can be fixed purely by adjusting the source string.

eddiemuc:
For clarification: up to now I thought that the default strings.xml contains the english translation only. So I assumed that (also for plurals) we only have to take care about the rules as they apply in english language - and other translations may add the other cases as needed. (E.g. in chinese I learned that not even the "one" case is needed)

Is that not true? Becasue right now this leads (in english) to a lot of duplication.

okainov:
Russian is one of such beautiful languages :)

Lineflyer:
> Is that not true?

I also thought, that Crowdin would add needed other quantities depending on the language, but a while ago we learned, that we need to define all of them in our strings, otherwise those do not show up as possible quantity in specific languages. I might be wrong though.... @bekuno Didn't you add the missing quantities some months ago due to this reason?

Lineflyer:
Missing quantities have been discussed here:
#3599
#8456

The outcome was, that it *might* be a Crowdin bug, but it works (in most cases except french) when we add all quantities to the source.

Lineflyer:
See #8029 for lint warnings and lint configuration.

Lineflyer:
Done here.

PR #10157, opened 2021-03-13
Cleanup strings

Lineflyer:
- Work on #10155
- Remove strings, which are reported unused by lint ( @cgeo/team Please doublecheck, as I don't know how true these lint reports are)
- Replace unmatching variables (decimal instead of string)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -254,7 +254,6 @@
     <string name="warn_search_help_tb">Enter the code for a trackable. For example \"TB29QMZ\".</string>
     <string name="warn_log_text_fill">Please input some text for your log.</string>
     <string name="warn_load_images">c:geo failed to load images.</string>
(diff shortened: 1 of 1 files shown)


main/res/values/strings.xml
```diff
@@ -1839,7 +1831,7 @@
 
     <plurals name="days_ago">
         <item quantity="zero">%d days ago</item>
-        <item quantity="one">yesterday</item>
+        <item quantity="one">%d day ago</item>
```
fm-sys, line 1834: Why don't keep yesterday?

fm-sys: Ignore me, found the reason in https://github.com/cgeo/cgeo/issues/10155#issue-831002685

Lineflyer:
@eddiemuc or @bekuno 

Does this look OK for you?
It should tackle some lint issues about missing quantities (once translations come in) and unused resources.
But please check if the resources are really unused.

eddiemuc:
@Lineflyer strings in context of SAF-stuff look ok to remove

Issue #8029, opened 2019-12-26
Fix remaining lint issues and define baseline


moving-bits:
Due to several PR we recently brought the number of lint issues from about 300 down to app. 130. Now the somewhat trickier issues remain :-)

At least for some issue classes I'm unsure how to tackle them, therefore I open this issue with the (in terms of number) most relevant issues to discuss possible solutions. The following issues are listed in no particular order:

1) lint message: "Possible overdraw: Root element paints background `?background_color` with a theme that also paints a background (inferred theme is `@style/cgeo`)"
- source: several layout files
- code sample: `<LinearLayout ... android:background="?background_color" ...> ... </LinearLayout>`
- lint explanation: "If you set a background drawable on a root view, then you should use a custom theme where the theme background is null. Otherwise, the theme background will be painted first, only to have your custom background completely cover it; this is called "overdraw". NOTE: This detector relies on figuring out which layouts are associated with which activities based on scanning the Java code, and it's currently doing that using an inexact pattern matching algorithm. Therefore, it can incorrectly conclude which activity the layout is associated with and then wrongly complain that a background-theme is hidden. If you want your custom background on multiple pages, then you should consider making a custom theme with your custom background and just using that theme instead of a root element background. Of course it's possible that your custom drawable is translucent and you want it to be mixed with the background. However, you will get better performance if you pre-mix the background with your drawable and use that resulting image or color as a custom theme background instead."

---

2) lint message: "The resource `@attr/dropdownListPreferredItemHeight` is marked as private in androidx.appcompat:appcompat"
- source: several layout files
- code sample: `android:layout_height="?attr/dropdownListPreferredItemHeight"`
- lint explanation: "Private resources should not be referenced; the may not be present everywhere, and even where they are they may disappear without notice. To fix this, copy the resource into your own project instead."

---

3) lint message: "`@id/distance` can overlap `@id/heading` if @null, @null grow due to localized text expansion"
- source: several layout files
- lint explanation: "If relative layout has text or button items aligned to left and right sides they can overlap each other due to localized text expansion unless they have mutual constraints like `toEndOf`/`toStartOf`."

---

~4) lint message: "Do not concatenate text displayed with `setText`. Use resource string with placeholders."~
- ~source: many occurrences, but probably easy to fix (for most of them)~
[...]

PR #10161, opened 2021-03-14
Rel #8029, #10155 - Let lint abort on error

Lineflyer:
DO NOT MERGE

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -154,7 +154,8 @@ android {
 
     lintOptions {
         // generally we accept lint errors when building
-        abortOnError false
+        // setting to true for testing, related to #10155, #8029
+        abortOnError true
 
         // abort release builds in case of FATAL errors
         checkReleaseBuilds true


eddiemuc:
Hi @Lineflyer I am looking into the same thing right now. Problem is that we currently have about 80 Lint errors in code which would immediately lead to failing build if we apply this.

Lint however provides the option to give it a so-called "baseline" file which lists errors (or warnings) Lint should ignore for now. Providing such a baseline file together with turning on abortOnError 8as you did) should imho give us the wanted result.

I would prepare an own PR for this, if that's ok? (Sharing PRs usually involved git-magic which I am not so comfortable with...)

Lineflyer:
@eddiemuc 
Sure, go for it. I am just playing around here.
But the result is exactly as you predicted. Abort on error because of 84 errors.

So the baseline sounds like a reasonable idea, but the downside is, that it might make us ignore the existing 84 errors, which should also be analyzed if time permits (see #8029 ). We can discuss the basline idea in that issues.

I will drop this experiment here anyway.

PR #8129, opened 2020-02-16
complete cache/waypoint hiding info in system info

moving-bits:
The new (and not so new) hiding features should be reflected in system info:
- hide disabled
- hide archived
- hide original waypoints
- hide parking waypoints
- hide visited waypoints

image

Changes "Hide own/found" to "Hide caches" and lists up to three options "own/found", "disabled", "archived", or "-" if none set.
Adds "Hide waypoints" with up to three options "original", "parking", "visited", or "-" if none set.

PR #10162, opened 2021-03-14
rel to #8029: make build fail on lint error (defines baseline containing current errors)

eddiemuc:
rel to #8029: make build fail on lint error (defines baseline containing current errors)

This PR will switch build to fail if there are lint errors on it (it does not fail for warnings!)

Since current master has 84 lint errors and I can't fix them all right now (:-)), in order to not fail the build this PR also contains a baseline file which basically lets lint ignore those 84 errors when deciding whether to fail or not. But any new error from now on will fail the build.

### DIFF
--- a/main/lint-baseline.xml
+++ b/main/lint-baseline.xml
@@ -1,0 +1,1570 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="5" by="lint 4.1.2">
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="src/cgeo/geocaching/AboutActivity.java"
+            line="230"
+            column="17"/>
+    </issue>
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="src/cgeo/geocaching/address/AddressListActivity.java"
+            line="30"
+            column="17"/>
+    </issue>
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="src/cgeo/geocaching/CacheDetailActivity.java"
+            line="240"
+            column="17"/>
+    </issue>
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="src/cgeo/geocaching/CompassActivity.java"
+            line="70"
+            column="17"/>
+    </issue>
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="src/cgeo/geocaching/EditWaypointActivity.java"
+            line="200"
+            column="17"/>
+    </issue>
+
+    <issue
+        id="MissingSuperCall"
+        severity="Error"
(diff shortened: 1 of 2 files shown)


main/lint-baseline.xml
```diff
@@ -0,0 +1,1590 @@
...
+        severity="Error"
+        message="Overriding method should call `super.onCreate`"
+        category="Correctness"
+        priority="9"
+        summary="Missing Super Call"
+        explanation="Some methods, such as `View#onDetachedFromWindow`, require that you also call the super implementation as part of your method."
+        errorLine1="    public void onCreate(final Bundle savedInstanceState) {"
+        errorLine2="                ~~~~~~~~">
+        <location
+            file="C:\private\geocaching\github\cgeo\main\src\cgeo\geocaching\AboutActivity.java"
```
Lineflyer, line 15: Will it work with these absolute pathes?

eddiemuc: Hehe, didn't see that this is in the files. So you now know my local folder structure. I will probably replace the "root".

I just took the (locally created) lint xml report files and removed all "Warnings" from it. I think there's far more in this report file than really needed for the "lint-baseline" check, this line among them. But it is helpful I guess for those later trying to solve these issues to have this information.

Whether it will work with these paths: we will see it soon in this PR. If Jenkins can build it then it works (assuming that it does not have the very same file structure :-)).

eddiemuc: Jup, does not work, have to replace with relative paths. Done.

Lineflyer:
@eddiemuc 
Looks not so bad now:

Lint is only claiming about one error, which I indeed need to check in Chinese translation.

`3:03:25  Execution failed for task ':main:lintBasicDebug'.
13:03:25  > Lint found errors in the project; aborting build.
13:03:25    
13:03:25    Fix the issues identified by lint, or add the following to your build script to proceed with errors:
13:03:25    ...
13:03:25    android {
13:03:25        lintOptions {
13:03:25            abortOnError false
13:03:25        }
13:03:25    }
13:03:25    ...
13:03:25    Errors found:
13:03:25    
13:03:25    /home/jenkins/slave/workspace/cgeo pull request/main/res/values-zh-rTW/strings.xml:996: Error: Inconsistent formatting types for argument #1 in format string cache_rating_of_new ('%1$d'): Found both 'f' and 'd' (in values/strings.xml) [StringFormatMatches]
13:03:25        <string name="cache_rating_of_new">%2$d ? %1$d</string>
13:03:25                                                  ~~~~
13:03:25        /home/jenkins/slave/workspace/cgeo pull request/main/res/values/strings.xml:1181: Conflicting argument type here

Additionally the baseline seems to have one superflous error listed:

03:23  Lint found 1 error, 116 warnings (83 errors filtered by baseline lint-baseline.xml)
13:03:23  
13:03:23  1 errors/warnings were listed in the baseline file (/home/jenkins/slave/workspace/cgeo pull request/main/lint-baseline.xml) but not found in the project; perhaps they have been fixed?
13:03:23

Lineflyer:
Regarding

main/res/values-zh-rTW/strings.xml:996: Error: Inconsistent formatting types for argument #1 in format string cache_rating_of_new ('%1$d'): Found both 'f' and 'd' (in values/strings.xml) 

I fixed the string to matching variables on Crowdin. It was simply using wrong variables in the translation.
However I am wondering why the source string uses those different variable names after all:
` <string name="cache_rating_of_new">%2$d 的 %1$d</string>`

eddiemuc:
@Lineflyer for some reason lint does for one error not recognize the "omit" entry in lint baseline (hence one error is still reported and the same errors entry in lint-baselibe is claimed as "no longer found" in code.

I noticed that actually many lint errors are related to translations.

Lineflyer:
> I noticed that actually many lint errors are related to translations.

Errors or warnings?
For the error I have seen quite some of the quantity issues, which I am about to tackle with PR #10157 and later on Crowdin for each language.

eddiemuc:
I mean Errors. MAny of them are indeed related to quantity issues. May I should wait with the merge of this PR until #10157 is merged and adapt lint-baseline.xml according to it?

main/lint-baseline.xml
```diff
@@ -0,0 +1,1570 @@
...
+            column="97"/>
+        <location
+            file="res/values-ca/strings.xml"
+            line="1776"
+            column="5"/>
+    </issue>
+
+    <issue
+        id="StringFormatMatches"
+        severity="Error"
```
Lineflyer, line 434: From what I assume it means, that we should use `%1$d` placeholders instead of `%1$s` here as the the parameters `min` and `max` are integers. Correct?

That would be easy to solve and will solve the corresponding issue for each language.

main/lint-baseline.xml
```diff
@@ -0,0 +1,1570 @@
...
+            column="102"/>
+        <location
+            file="res/values/strings.xml"
+            line="1972"
+            column="5"/>
+    </issue>
+
+    <issue
+        id="ImpliedQuantity"
+        severity="Error"
```
Lineflyer, line 1374: These errors (for affected language) should be fixed implicitly after merging #10157 and incoming corresponding translations.

Lineflyer (commented):
As commented inline:
The errors regarding `StringFormatMatches` and `ImpliesQuantity` should be easy to solve and could be removed afterwards from the baseline.

main/res/values-zh-rTW/strings.xml
```diff
@@ -993,7 +993,7 @@
     <string name="cache_terrain">地形</string>
     <string name="cache_rating">評分</string>
     <string name="cache_own_rating">你的評分</string>
-    <string name="cache_rating_of_new">%2$d 的 %1$d</string>
+    <string name="cache_rating_of_new">%2$d 的 %1$.1f</string>
```
Lineflyer, line 996: Changed it in the same way on Crowdin, so that this change wont get overwritten by Crowdin again.
Should fix the corresponding error.

main/lint-baseline.xml
eddiemuc: Correct., using %1$d in cs translation should fix this

Lineflyer: Not only in cs but in the source, whereas the translations should be updated afterwards.
Please note, this error is present for each language.

I will tackle this and push to #10157

Lineflyer:
> I mean Errors. MAny of them are indeed related to quantity issues. May I should wait with the merge of this PR until #10157 is merged and adapt lint-baseline.xml according to it?

Yes, please. Give me some time to sort them out.
There are three main errors in strings:
- Quantity -> To be solved by #10157 and incoming translation
- String format invalid -> Incorrect chinese translation, I will remove them on Crowdin
- String format Matches -> Will add a fix to #10157 

Once #10157 is merged, I will adapt some languages on Crowdin, merge translations back.
Then we should have a better (smaller) baseline.

Lineflyer:
Sorry, merged it already by accident.
Anyway the baseline can be updated afterwards. I will comment in #8129 once everything is done on my side (Crowdin).

PR #10189, opened 2021-03-17
Treat missing ellipsis character as lint error

Lineflyer:
Make build fail (as we stop on lint error now) if strings or incoming translations use `...` instead of the designates ellipsis character. This can currently not be forbidden/mitigated on Crowdin while translating and is an often made error made by translators.

Thus not detecting it will lead to more lint warning in the project (and more ugly strings).


eddiemuc:
@Lineflyer the PR should target master, not release. In release lint will currently not fail the build on error.

eddiemuc:
@Lineflyer forget my previous comment, I just saw that this is already merged to release

Lineflyer:
@eddiemuc  Exactly.

PR #10201, opened 2021-03-18
keep routing mode settings enabled in quick settings (fix #10187)

### LINKED ISSUE #10187: Direction line active while it was deactivate before and can no longer be deactivated
From support mail for current beta 2021.03.16-RC1:

- No Brouter installed
- The user had direction line switched off in settings
- After upgrading to the beta version the direction line is shown
- Using map quick settings to deactivate it, is not possible. 
- Side fact: The sys info shows routing type "Walk" 

Something seems wrong in migration of the old setting to the new map quick setting.

I could reproduce at least the fact, that direction line cannot be deactivated anymore if Brouter is not installed.
The migration case needs to be debugged.

moving-bits:


fm-sys:
This would be a quick fix for #10187, However, with the small flaw that you can select routing profiles that are not available at all. Better would be to keep car, bike, running disabled. 

And one more thing which needs to be edited:
https://github.com/cgeo/cgeo/blob/9aa715f5a6b1069d29dc7d2166036cb43b8b1d22/main/src/cgeo/geocaching/maps/MapSettingsUtils.java#L88

That's probably also the reason why the settings was not migrated correctly.

fm-sys:
This would be a quick fix for #10187, However, with the small flaw that you can select routing profiles that are not available at all. Better would be to keep car, bike, running disabled. 

And one more thing which needs to be edited:
https://github.com/cgeo/cgeo/blob/9aa715f5a6b1069d29dc7d2166036cb43b8b1d22/main/src/cgeo/geocaching/maps/MapSettingsUtils.java#L88

That's probably also the reason why the settings was not migrated correctly.

Lineflyer:
> Better would be to keep car, bike, walking disabled.

As also commented in the issue:
Yes, that would be better.
You cant imagine how many users would contact support like "I selected Bike, but still get a straight line.".

moving-bits:
> And one more thing which needs to be edited:
> 
> https://github.com/cgeo/cgeo/blob/9aa715f5a6b1069d29dc7d2166036cb43b8b1d22/main/src/cgeo/geocaching/maps/MapSettingsUtils.java#L88
> 
> That's probably also the reason why the settings was not migrated correctly.

Not really, that would only be a display glitch (which should be fixed anyway), as this value did not get persisted. The map would still continue to use the actual stored value, WALK in the user's case.

> However, with the small flaw that you can select routing profiles that are not available at all. Better would be to keep car, bike, walking disabled.

Hm, maybe we need to tackle this in a different place, in `Settings.getRoutingMode()` => return STRAIGHT if BRouter is not available and set to a value other than OFF. In map quick settings we could show only OFF and STRAIGHT buttons then, hiding WALK / BIKE / CAR buttons.

fm-sys:
> In map quick settings we could show only OFF and STRAIGHT buttons then, hiding WALK / BIKE / CAR buttons.

IIRC we had the same discussion somewhere before, but decided to disable them instead of hiding (as kind of advertisement for BRouter)

> Not really, that would only be a display glitch (which should be fixed anyway), as this value did not get persisted. The map would still continue to use the actual stored value, WALK in the user's case.

It doesn't necessarily need to be persisted as at the Livemap, WALK without BRouter results in STRAIGHT. So if BRouter is temporarily unavailable for some reason the user setting will not be changed, but still WALK is shown in the popup as current routing mode. IMHO this isn't a glitch but designed that way.

moving-bits:
> > In map quick settings we could show only OFF and STRAIGHT buttons then, hiding WALK / BIKE / CAR buttons.
> 
> IIRC we had the same discussion somewhere before, but decided to disable them instead of hiding (as kind of advertisement for BRouter)

But it's hard to recognize whether they are disabled, as they are still shown in orange, but only do not react to tapping them - which may be pretty confusing. So we should either "grey out" the unusable buttons (which would mean extra button resource files, but ok) or remove the others.

Advertising for BRouter is there anyway, as we have added the nice info message below the buttons meanwhile.

fm-sys:
> 
> But it's hard to recognize whether they are disabled, as they are still shown in orange, but only do not react to tapping them - which may be pretty confusing. So we should either "grey out" the unusable buttons (which would mean extra button resource files, but ok) or remove the others.

Probably we should not spend too much time into this very small glitch as it will become obsolete with internal BRouter anyway... So for the time being, decide what's easiest for you! 😉

moving-bits:
Ok, I updated this PR. Map quick settings will look like this if BRouter is not installed:

image

(Actual setting for routing mode was BIKE in this case, but will get displayed as STRAIGHT, if BRouter is not installed.)

fm-sys:
Have you checked that RoutingMode.OFF is not also displayed as RoutingMode.STRAIGHT (it looks like that from the code)?

Lineflyer:
mmh..from looking at the picture:
People not knowing brouter will not understand this:
It says something about routing, says it unavailable but still I can select something but where is the real routing which is unavailable?
Isnt it possible to just disable the real routing options instead of hiding?

moving-bits:
What do we want / need to achieve?
The original issue was about not being able to disable the distance line, which is what this PR fixes.

> Isnt it possible to just disable the real routing options instead of hiding?

This is what we had before - show all buttons, but have them disabled. This approach has two defects, IMHO:
- You cannot enable / disable direction line (as the buttons are inactive)
- You have no visual indicator that the buttons are disabled.

The first part can be solved by activating the first two buttons (OFF / STRAIGHT).
If we want to show the three extra buttons, then the latter could be solved with "greying out" the WALK / BIKE / CAR buttons.

We could (IMHO: should) also adapt the info text to "More routing options will be available if BRouter is installed" (with the negative impact that this text becomes two lines.
Or change it to "More info on routing options", and link this to a popup explaing the rest, or link it to the FAQ.

OTOH: If we change the info text, there would be no need for greying-out the three extra buttons, we could simply hide them...

moving-bits:
> Have you checked that RoutingMode.OFF is not also displayed as RoutingMode.STRAIGHT (it looks like that from the code)?

You can toggle between the two states, and get displayed either "off" or "straight".
But I need to recheck for the initial value. May be that I forgot to adapt one line after rebasing this PR to current `release`.

Lineflyer:
> We could (IMHO: should) also adapt the info text to "More routing options will be available if BRouter is installed" (with the negative impact that this text becomes two lines.

IMHO the text would still be OK. Straight line is not really a routing. 
For me its also fine if all icons have the appearance they currently have, but the right most three icons are just not clickable. The text below explains why. Looks fine to me and advertises the routing feature as people see: "Oh, I can have a bike route...its not clickable...ah, I have to install Brouter".

Lineflyer:
Addon:
@moving-bits From reading your latest post I would interprete its technically not possible to have only some of the buttons available? Is that correct?
To better explain my posting above:
I was thinking it could be possible to show all the icons, but only have the left two clickable if BRouter is not installed.

fm-sys:
> From reading your latest post I would interprete its technically not possible to have only some of the buttons available? Is that correct?

No, that's possible too. I think we try to solve two issues with one PR. One thing is which buttons are enabled and the other thing how disabled buttons appear.

Btw, I found following idea at SO:
```
button.setAlpha(.5f);
button.setClickable(false);
```
or
```
getBackground().setColorFilter(Color.GRAY, PorterDuff.Mode.MULTIPLY)
    setClickable(false)
```
(see https://stackoverflow.com/a/27116948)

May I take over the PR? As I was it, who had originally implemented disabling the routing buttons if BRouter is not available...

moving-bits:
> @moving-bits From reading your latest post I would interprete its technically not possible to have only some of the buttons available? Is that correct?
> To better explain my posting above:
> I was thinking it could be possible to show all the icons, but only have the left two clickable if BRouter is not installed.

All of the following is possible:
1) Show only the two active buttons
2) Show all five buttons, but have only the active ones clickable
3) same as before, but visualize which buttons are inactive

Personally I would prefer 3 - 1 - 2 in this order, especially with @fm-sys having an elegant solution for greying out the inactive buttons.

moving-bits:
@fm-sys 

> May I take over the PR? As I was it, who had originally implemented disabling the routing buttons if BRouter is not available...

Sure, if you like.

Lineflyer:
My preference would be 3 - 2, in that order :)

fm-sys:
Variant 3 is done in #10210

moving-bits:
superseded by #10210

PR #10210, opened 2021-03-19
keep routing mode OFF/STRAIGHT always enabled in quick settings

### LINKED ISSUE #10187: Direction line active while it was deactivate before and can no longer be deactivated

fm-sys:
OK, here we are...

grafik

closes #10201
fixes #10187

~Sorry, it's targeted to master currently. I will change it this evening... (No more time right now)~


fm-sys:
retest this please

moving-bits (approved):

PR #10212, opened 2021-03-19
system info: move settings info to settings section

moving-bits:
Cosmetic issue only: Move this info

image

from end of "permissions & path" section to "program settings" section.


Lineflyer:
I have seen this new setting. Could you explain the meaning?

eddiemuc:
It shows you the migration version of the settings (comparable to the database version, but for the setting structure) and the number of (raw) entries in your settings

moving-bits:
retest this please

Issue #10176, opened 2021-03-15
BRouter integration: Migrate access of routing tile data to SAF


moving-bits:
Routing tile data (files with extension '.rd5') reside in the `segments4` subfolder. As they tend to be pretty large (up to 150 MB and more), and as there are many of them potentially, they should be sharable with other apps, e. g. Locus. Therefore accessing files via SAF is needed, and the current caching mechanism needs to be adapted. (See `PhysicalFile` etc.)

@eddiemuc 
This is the part we've talked about.

eddiemuc:
@moving-bits after BRouter stuff is now in master, can/should I start working on this issue?

moving-bits:
> @moving-bits after BRouter stuff is now in master, can/should I start working on this issue?

That would be great. Starting point is `brouter/BRouterService`.

eddiemuc:
Hi @moving-bits 
I thought before I dive into filtering too deep I should do the SAF migration of BRouter before that :-)

I took a look at the code, before I walt too much in the wrong direction I want to verify that my understanding of what needs to be done is correct:

There shall be two new public folders:
  * PersistableFolder.ROUTING_BASE (location defaults to BASE/routing). This folder shall contain:
       * multiple "rawTrackPath" files (named with suffix _rawtrack.dat"
       * multiple "profile" files (named e.g. fastbike.brf"
       * a file "lookups.dat" which you seem to have migrated to SAF already?
   * PersistableFolder.ROUTING_TILES (defaults to ROUTING_BASE/segment4). This folder contains multiple files with tile data inside (with suffix .rd5)

Is that about it or am I missing something?

I also noticed that class RouterEngine has a suspicious members "outfileBase" and "logfileBase", which however seems to be initialized to null always. Is this intended, can the be removed?

Lineflyer:
I think there is no need for two separate folders. Either all data could be in/below a single folder, or the non user modifiable data could be stored internally.

eddiemuc:
> I think there is no need for two separate folders. Either all data could be in/below a single folder, or the non user modifiable data could be stored internally.
> 
> 

Iirc the idea was to separate the data sharable between applications (the tile data, which is large) from the cgeo-specific data (the other data e.g. profiles). Just because there are two public folders does.not mean that both are settable by the user. We have many public folders not settable by the user e.g. logfiles or spoiler-image-dir

Lineflyer:
[...]

Issue #10175, opened 2021-03-15
BRouter integration: Migrate profile file access to SAF


moving-bits:
BRouter uses special routing profiles, provided in files ending in '.brf', stored in the new `routing` folder.
Access (and caching) for those files needs to be migrated to SAF.

PR #10217, opened 2021-03-20
BRouter integration

moving-bits:
(needed to restart from scratch. Somewhere during the Codacy/CheckStyle fixes in #10156 I've broken something, which prevents integrated BRouter from routing at all)

This is a basic integrated version of BRouter:
- import the original BRouter files
- packaged them into appropriate folders / packages
- some basic (and mostly automated) reformatting - both Codacy and CheckStyle will complain a lot here
- connected c:geo and internal BRouter

It's already a working version: c:geo is now connecting to this internal BRouter variant and doing actual routing (provided that ~`*.brf`, `lookups.dat` and~ `*.rd5` files are in the places expected by this internal BRouter version).
Edit: `*.brf` and `lookups.dat` get generated automatically meanwhile, so the only files you have to add manually are the routing tile data files `*.rd5`. This will be solved as soon as the tile downloader is integrated and activated.)

P.S.: Ok, this time I started with 1703 messages from Codacy and 2421 from CheckStyle, compared to app. 19000 each last time... :-)

With further commits I'll do either optimizations (removing unused BRouter code), solving some Codacy/CheckStyle complaints or implement more of the missing things from the project's todo-list.


moving-bits:
This PR is down from app. 19.000 Cadacy / CheckStyle messages each to 153 Codacy and 344 CheckStyle. For some of the open issues fixing them breaks the BRouter codec (have tried it...), other will be replaced by refactoring to SAF anyway, so we agreed upon merging this anyway, doing the last open steps for SAF migration, and continue trying to fix the remaining issues then (https://github.com/cgeo/cgeo/pull/10156#issuecomment-800845992).

Next steps after merging this PR will be #10175, #10176 and merging #10085.

fm-sys:
Tests are failing for all other PR's coursed by the new BRouter code (see #10270 for example). I don't know if future steps need to be done to get the light back to green...

moving-bits:
Yeah, have seen that (and merged #10270 anyway, because the only errors are coming from the BRouter integration). Don't know if we need to do immediate action (like creating some sort of "ignore" file), or if upcoming builds recognize that those are "old" messages not introduced by the new PR, and therefore not break the build because of it. Let's keep an eye on it.

eddiemuc:
I am confused. Why was the non-compliant BRouter code merged to master in the first place? I thought we agreed to make an own feature branch instead (like we did with storage_framework) until it is master-ready

moving-bits:
Hm, may be a misunderstanding on my side. I thought the parallel branch was due to SAF still being on `master` at that time, and with having a fresh `master` merging to `master` would be fine. Especially since testing those changes on `master` is way easier than on a separate branch (as we had experienced with the SAF branch).

eddiemuc:
@moving-bits it is in any case not a good idea imho to merge stuff to master which breaks the build. I am trying to fix this in #10273

fm-sys:
I think `master` branch is fine **as long as we really concentrate** on bringing BRouter into production state in a reasonable time. Otherwise it would indeed be very bad if we have not working code at `master` but no time/interest to continue development of this functionality...

eddiemuc:
@moving-bits , @fm-sys : as I said many times, in my opinion the master branch must **always** be green with regards to CI - except in very special emergency error situations (which this is not). It must never occur that PRs from unrelated features fail due to stuff carelessly merged there. If we start once to not care about green master any more it will degrade very quickly.

Experiments and red builds can be done on feature branches, this is what they are for. If really bad code must be used in master then suppressions must be used, but the build must stay green.

As an example, in the "windschatten" of the checkstyle problems I just found that there was already a new lint error introduced. This would have failed all successing builds too.

fm-sys:
> @moving-bits , @fm-sys : as I said many times, in my opinion the master branch must always be green with regards to CI

Totally agree. My comment above was more about if it's ok to keep the BRouter code in current state on `master` at all or if a revert is necessary. But I would guess that the work is doable until next feature release.

moving-bits:
> I think `master` branch is fine **as long as we really concentrate** on bringing BRouter into production state in a reasonable time. 

Agreed. I've brought Codacy/CheckStyle numbers for BRouter down to the current numbers, while checking routing functionality numerous times along the way to not break the actual code by applying coding style fixes. For the remaining issues I'm sort of stuck, but for sure very willing to help with them, if someone shows me how to tackle them.

> Otherwise it would indeed be very bad if we have not working code at `master` 

Currently it should be "Not working" in terms of coding style compliance only, not in terms of "routing functionality" or "breaking some other c:geo functionality".

> but no time/interest to continue development of this functionality...

Indeed. But if so, we wouldn't have integrated BRouter code at all in the first place.

PR #10221, opened 2021-03-20
Inform user that login problem prompt is clickable

Lineflyer:
I was wondering why quite some users still contact support and have no login information stored. Mostly they tell, that they see the warning on main screen but don't know how to login.
Maybe its just as simple as adding the help, that the message can be clicked (which will provide more info and guide to service settings).


bekuno:
Even the connector list is clickable. Didn´t realized that before!

Lineflyer:
> Even the connector list is clickable. Didn´t realized that before!

Indeed...such as at many places (e.g. also the satellite status)

moving-bits (approved):

PR #10222, opened 2021-03-20
fix #10209, fix #10200: gather system info async, add map info

### LINKED ISSUE #10200: Include active map in system information
For many issues its relevant for the support to know, whether the user uses Google or Mapsforge. Additionally it might sometimes helpful to also see how he uses Mapsforge (Which online tile provider / name and provider of offline map).
Currently this always requires some mail ping-pong.

So I suggest (if possible) to add two line to the section "Program settings", showing up like the following examples:
Active map type: Google Maps / Google Satellite / openstreetmap.org /openstreetmap,de / ... / Offline map
Offline map name: germany_oam / Provider: OpenAndroMaps 

For offline maps the map name should be as shown in map selector (file name without extension). If possible map provider would be nice (taken out of the attribution maybe?), but could also be omitted (as I do not know of any bug up to now depending on map file provider)

### LINKED ISSUE #10209: App slow to start AboutActivity
Taken from a comment of the beta version survey:

When trying to access "About c:geo " in main menu or clicking the small c:geo logo in the title of the main screen, the screen turns black and it takes some seconds until the "About" screen is shown.

---

I can sort of reproduce it: For me on my (pretty fast) Samsung S20 the screen does not turn black but there is a significant delay (at least on first try during a session).
From my feeling: Could it be related to the information gathering for the "System information" screen?. Could be, it takes some time to populate all the settings output?

eddiemuc:
fix #10209, fix #10200: gather system info async, add map info

### DIFF
--- a/main/src/cgeo/geocaching/AboutActivity.java
+++ b/main/src/cgeo/geocaching/AboutActivity.java
@@ -12,14 +12,17 @@ import cgeo.geocaching.ui.AnchorAwareLinkMovementMethod;
 import cgeo.geocaching.utils.ClipboardUtils;
 import cgeo.geocaching.utils.DebugUtils;
 import cgeo.geocaching.utils.LiUtils;
+import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.ShareUtils;
 import cgeo.geocaching.utils.SystemInformation;
 import cgeo.geocaching.utils.Version;
 
 import android.app.Activity;
+import android.content.Context;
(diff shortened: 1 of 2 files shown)


Lineflyer:
@eddiemuc 

New lint warning (don't know if it has relevance):

AboutActivity.java:319,           StaticFieldLeak, Priority: Normal
--
This field leaks a context object

eddiemuc:
@Lineflyer Good catch. Replaced the aboutactivity-context with the application context for gatherind sys information, this fixes the warning. I noticed however that there are 21 more...

Anyway, should be fixed for this PR now.

PR #10225, opened 2021-03-20
Move some strings to non translatable

Lineflyer:
Moving some strings into the non-translatable string file, as:
- Those does not contain translatable content (such as N° --.---..."
- Are own names and should not be translated (like Openstreetmap.org, Rmaps)
- Our app name should also not be fooled with on Crowdin

There have been occasions in the past where such strings have been translated and thus they are no longer correct (e.g. Rmaps to german Rkarten, Openstreetmap.de in spain to openstreetmap.es, etc.)

BTW:
This is my first PR which has been built locally on my machine before submitting it :)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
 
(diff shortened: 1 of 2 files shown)


Lineflyer:
The 339 new warnings should be solved later on by Crowdin.

PR #10231, opened 2021-03-21
allow TB special tracking code length 9

### LINKED ISSUE #10192: "Mars Perseverance Rover Trackable" does not come up in "Trackable" search
**Describe the bug:**
Mars Perseverance Rover Trackable does not come up when searching for a trackable with the trackable code [REDACTED]. But it works when you enter the trackable reference number TB5efxk.

See https://www.geocaching.com/track/details.aspx?tracker=TB5efxk

**To Reproduce:**
1. Open app
2. Go to "Search"
3. Scroll down to "Trackable"
4. Enter [REDACTED], hit enter
5. No result

All normal, the trackable was just not found.

I expect the https://www.geocaching.com/track/details.aspx?tracker=TB5efxk to come up for me to be logged.

**Version of c:geo used:**
2021.02.14

Yes

**System information:**
```
--- System information ---
Device: GM1913 (OnePlus7Pro_EEA, OnePlus)
Android version: 10
Android build: GM1913_21_210128
c:geo version: 2021.02.14
Google Play services: enabled - 21.06.13 (120400-358943053)
Low power mode: inactive
Compass capabilities: yes
Orientation sensor: present
Hide caches: -
Hide waypoints: -
System language: en_US / user-defined language: 
System date format: M/d/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (181.4 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (181.4 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (181.4 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (840.0 KB) on system internal storage
Last backup: never
Offline maps path: /storage/emulated/0/Maps
Map render theme path: /storage/emulated/0/Maps/MapThemes/Elevelo_Cycling.xml
Live map mode: true
Sailfish OS detected: false
Geocaching sites enabled:
```

bekuno:
fixes #10192

- edit pattern to additional allow 9 digit tracking code formatted for logging of TB5EFXK
- more tests for 5 to 10 digits length
- fix a possible nullable problem
- switch test class to simple junit4 tests

### DIFF
--- a/main/src/cgeo/geocaching/connector/trackable/TravelBugConnector.java
+++ b/main/src/cgeo/geocaching/connector/trackable/TravelBugConnector.java
@@ -26,7 +26,7 @@ public class TravelBugConnector extends AbstractTrackableConnector {
     /**
      * TB codes really start with TB1, there is no padding or minimum length
      */
-    private static final Pattern PATTERN_TB_CODE = Pattern.compile("(TB[0-9A-Z&&[^ILOSU]]+)|([0-9A-Z]{6})|([0-9]{5})", Pattern.CASE_INSENSITIVE);
+    private static final Pattern PATTERN_TB_CODE = Pattern.compile("(TB[0-9A-Z&&[^ILOSU]]+)|([0-9A-Z]{6})|([0-9]{3}[A-Z]{6})|([0-9]{5})", Pattern.CASE_INSENSITIVE);
 
     @Override
     public int getPreferenceActivity() {
@@ -35,6 +35,9 @@ public class TravelBugConnector extends AbstractTrackableConnector {
 
     @Override
     public boolean canHandleTrackable(@Nullable final String geocode) {
+        if (geocode == null) {
+            return false;
+        }
         return PATTERN_TB_CODE.matcher(geocode).matches();
     }
 
--- a/tests/src-android/cgeo/geocaching/connector/trackable/TravelBugConnectorTest.java
+++ b/tests/src-android/cgeo/geocaching/connector/trackable/TravelBugConnectorTest.java
@@ -2,20 +2,25 @@ package cgeo.geocaching.connector.trackable;
 
 import cgeo.geocaching.models.Trackable;
 
-import junit.framework.TestCase;
+import org.junit.Test;
 import static org.assertj.core.api.Java6Assertions.assertThat;
 
-public class TravelBugConnectorTest extends TestCase {
+public class TravelBugConnectorTest {
 
-    public static void testCanHandleTrackable() {
+    @Test
+    public void testCanHandleTrackable() {
         assertThat(getConnector().canHandleTrackable("TB1234")).isTrue();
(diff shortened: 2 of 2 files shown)


fm-sys:
Please don't merge! This will have unwanted side effects.

Screenshot_20210321_173334

If you currently search for a word with 6 characters c:geo will show a toast "trackingcode not found" instead of searching for caches with that name. This is a different issue for sure, but will dramatically increase if we adapt the regex to be less strict

bekuno:
This is certainly a different issue.

The global search uses different specialised searches, at least for cache names and for trackables.
And the trackable filter does not distinguish between tracking number and tracking code.
You can see from the results that the trackable search is only executed if no cache is found.
The problem is actually that the error message is misleading.

With my PR you can see that it uses a special filter criterion. 
Therefore, in addition, this should not be a big problem.

fm-sys:
> You can see from the results that the trackable search is only executed if no cache is found.
The problem is actually that the error message is misleading.

Trackable search is in fact executed before online cache name search. The listed caches arw only offline results. 

You can test the online search with "cache" and then press enter -> list of matching online caches will be listed 
Or "caches" and then press enter -> "trackingcode not found" 


But anyway, if all new trackingcodes will look like that, the change would for sure be OK for me. I more wonder about if this trackingcode was more like a one-time-special and if it's worth to break (although poorly working) existing functionality for something that brings no benefit... 

Is there any official statement whether the length of all trackingcodes has changed?

moving-bits:
Regarding tracking code with more than 6 digits: I wouldn't invest too much time here for now. I guess this has been a one-time marketing deal, and all other new TB will fit into the current scheme. Only when they are running out of digits they will add a seventh one. Or when there is another marketing show. :-)

bekuno:
> Regarding tracking code with more than 6 digits: I wouldn't invest too much time here for now.

We have two different things in discussion. 
At first support for this currently one time situation.
This is solved with this PR, and the filter is very special for this tracking code.
The other is the search mix in the global search for caches and trackable, and the search mix in the trackables search for tracking number and code. The could/should be discussed in an own PR.

bekuno:
> Is there any official statement whether the length of all trackingcodes has changed?

I don´t know. And I believe that this is currently the only one.
Therefore I created the special filter.

bekuno:
@Lineflyer 
What do you think?

Lineflyer:
I cannot judge the special filter as I don't "speak" RegEx.
The important thing for me would be, that any >6 characters are treated as search for other things than trackables in the quick search on the main screen except this special one.

bekuno:
The extra Filter is not the special  tracking code to not spoiler, but has the form 3 decimal plus 6 numerical symbols.

eddiemuc (approved):

PR #10233, opened 2021-03-21
update gradle to version 4.1.3

bekuno:
checked also the aapt2-proto reference, is further necessary

Not yet useful release notes in https://androidstudio.googleblog.com/2021/03/android-studio-413-available.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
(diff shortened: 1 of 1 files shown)


moving-bits:
really target this to release? I propose changing this to `master` to avoid having unknown obstacles on `release`

bekuno:
> 
> 
> really target this to release? I propose changing this to `master` to avoid having unknown obstacles on `release`

Yes, planned to target to release.
There are no known problems with this patch.
The gradle version 4.1.3 is the complement to AS 4.1.3, which we use for release and master.

PR #10251, opened 2021-03-24
fix 66 lint errors from baseline (MissingSuperCall, StringFormatMatch)

eddiemuc:
fix 66 lint errors from baseline (MissingSuperCall, StringFormatMatch)


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -238,7 +238,8 @@
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState, R.layout.cachedetail_activity);
```
rsudev, line 241: There were contexts, where this didn't work in the past and for some Android versions. Did you cross-check on light theme and older versions?

eddiemuc: No. But the oncreate method deleted was not even from Android system (it was c: geo self-made, despite the name oncreate suggesting otherwise) and I created the new method setthemeandcontentview to execute the new parts that were added in custom oncreate method. 

Cross-check can be done

eddiemuc: Hi @bekuno I tested this version on Android 5 (SDK21, Emulator) using both light and dark skin and clicking through all of the important dialogs. It seems to work.

Is there any specific situation where this did not work in the past that I should test additionally? Otherwise I would merge this.

eddiemuc: Sorry, I meant @rsudev before, not @bekuno ...

eddiemuc:
@Lineflyer if this PR is merged, there are 7 leftover errors in lint-baseline. All of them are "ImpliedQuantity" errors (one time for "zero", all other times for "one" quantity) and all of them claim that a parameter for the number is missing because the quantity is appliable to many numbers in the affected languages. Languages are lv and lt (I think this is Latvia/Lettland and Lithuania/Litauen).

I am at a loss how to fix those since I don't speak these languages. Would it be an option to simply delete these translations and let them be re-translated?


Example from `lint-baseline.xml`:

```
<issue
        id="ImpliedQuantity"
        severity="Error"
        message="The quantity `&apos;one&apos;` matches more than one specific number in this locale (1, 21, 31, 41, 51, 61, 71, 81, 101, 1001, …), but the message did not include a formatting argument (such as `%d`). This is usually an internationalization error. See full issue explanation for more."
        category="Correctness:Messages"
        priority="5"
        summary="Implied Quantities"
        explanation="Plural strings should generally include a `%s` or `%d` formatting argument. In locales like English, the `one` quantity only applies to a single value, 1, but that&apos;s not true everywhere. For example, in Slovene, the `one` quantity will apply to 1, 101, 201, 301, and so on. Similarly, there are locales where multiple values match the `zero` and `two` quantities.&#xA;&#xA;In these locales, it is usually an error to have a message which does not include a formatting argument (such as &apos;%d&apos;), since it will not be clear from the grammar what quantity the quantity string is describing."
        url="https://developer.android.com/guide/topics/resources/string-resource.html#Plurals"
        urls="https://developer.android.com/guide/topics/resources/string-resource.html#Plurals"
        errorLine1="        &lt;item quantity=&quot;one&quot;>Viena piezīme veiksmīgi augšupielādēta&lt;/item>"
        errorLine2="        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~">
        <location
            file="res/values-lv/strings.xml"
            line="1145"
            column="9"/>
    </issue> 
```

main/src/cgeo/geocaching/CacheDetailActivity.java
rsudev: As it is quite some time ago I do not remember the details any more. As the new implementation changes the call order for initializing theme and contentview with respect to the progress setting and the up action, those were the parts I would look into specifically.
Regarding the progress indicator, I am not up to date whether it is still in use at all (as it was phased out AFAIR by Google quite some versions ago). Previously the main usage context of this was the map, but that has been changed, if I am not mistaken.

rsudev:
Regarding the quantities, I would suggest to remove the translation on crowdin while leaving a related comment for the affected entities and target languages.

main/src/cgeo/geocaching/CacheDetailActivity.java
eddiemuc: Hi @rsudev , I checked progress bar in Android 5 SDK21 light mode, and there is indeed a problem: there is a black bar around it. However this bar is also present when using the master version, so I assume it is not introduced by the lint fixes

image

Is this what you meant?

Lineflyer:
@eddiemuc I will look into the baseline and fix those on Crowdin side either by deleting the translation (and let contributors retranslate) or fixing it directly if its clear to do.

main/src/cgeo/geocaching/CacheDetailActivity.java
rsudev: Now, that is the progress dialog (and those glitches are more likely have been introduced by recent theme changes rather than the change at hand).
In former Android version it was possible to have a progress indicator embedded in the ActionBar and that was used for the map update indication (and IIRC a few other places). But as I said, usage of that might be gone in the meantime.
If the up navigation (tapping the arrow in the upper-left corner) is still working as expected I would consider my concerns addressed.

eddiemuc: @rsudev ok the up navigation works. I will merge this PR

PR #10253, opened 2021-03-24
support changing found state for UDC (fix #10230)

### LINKED ISSUE #10230: Allow setting found state for UDC
Allow setting the found state of a UDC to either "found", "not found" (DNF) or default.

My use case is that I'm using UDC for lab caches (among other things). On having finished a lab cache I would like to see that on the map (or to hide the UDC the same way I hide other found caches). Currently I need to delete a UDC to achieve that, which also means losing the attached info (e. g. locations and answers to the stages).

Changing the found state could be put into the cache's "extras" submenu.

moving-bits:


bekuno (changes requested):
Thanks for the implementation.
The entry for "Set as found" is very hidden in the Advanced submenu.
I would vote for place this on the same position as "Log Visit" in the normal cache menu - at top of this menu.
And also on the other positions of "Log Visit", eg. in the cache overview screen.

moving-bits:
@bekuno 
I changed the menu entry position to top level (at same position as "log visit").
Additionally I adapted the code that this menu entry is also visible in the cache popup.

Works fine so far. Only drawback is that the cache popup's icon does not updated on state change.

Lineflyer:
But I assume its only shown for UDC?

moving-bits:
> But I assume its only shown for UDC?

Yes.

(Technically it's shown only for caches which belong to connectors supporting a found state change, but currently only the UDC connector supports this.)

Lineflyer (approved):

bekuno (approved):
Thanks for changing it.
Looks good.

bekuno (commented):
Now only the the status "found" is shown, "Not found" is hidden. 
Maybe lets show the status every time, so that users see easier the new possibility?

moving-bits:
What do you mean by
> Now only the the status "found" is shown, "Not found" is hidden.

?

The menu entry gets updated according to the respective status, so you can toggle state in both cache popup menu and cache detail page. The only thing that does not work is updating the cache icon in the cache popup. (It does work on cache details page, and after leaving the cache popup, the correct icon is shown on map as well.) Haven't found a way to change the icon dynamically in the popup yet.

bekuno:
Better with pictures:

No status = status "Not found"
Screenshot_20210327-203117_cgeo1

Status "Found"
Screenshot_20210327-203057_cgeo1

bekuno:
By the way, the status change is not visible in the cache popup.

moving-bits:
> By the way, the status change is not visible in the cache popup.

That's what I meant with

> The only thing that does not work is updating the cache icon in the cache popup.

Thanks for clarifying about the state. Reason for this is that we have two separate flags, a `found` flag and a separate `did not find` flag. Currently I'm setting or unsetting the `found` flag, but I'm not touching the `did not find flag`, as that is not persisted directly in the database.

fm-sys:
> `did not find flag`, as that is not persisted directly in the database.

Just for info: `did not find` is stored as "-1" (or "2", I'm not sure anymore) in the `found` column

moving-bits:
> > `did not find flag`, as that is not persisted directly in the database.
> 
> Just for info: `did not find` is stored as "-1" (or "2", I'm not sure anymore) in the `found` column

Ah, thanks! Typical case of not enough test data in my test database... :-)

So the question is: Shall I support "found" and "did not find" only or a three-state "found" / "default" / "did not find"?

fm-sys:
Possible would be dialog with these three options. Multiple context menu entries IMHO doesn't make sense

bekuno:
My thoughts before were not complete, I meant (originally) with "not found" the status "not yet found" == "not yet searched". After comparison with the behaviour of standard caches, the behaviour for UDC is analogous if one only takes "Found" as the sole permitted log variant.
On the other hand, the idea of (at least) 3-possibilities may be helpful. So if it doesn't mean too much implementation effort then gladly, otherwise fine with the current solution.

moving-bits:
Added setting DNF state, so that we have three options now: "Set as found", "Set as did not find" and "Reset found state" (showing only reasonable options). Currently for UDC only, but can be activated on a per-connector level.
BTW: I stayed with menu entries for easier handling (one click instead of two, and also less code).

PR #10257, opened 2021-03-25
lint: remove baseline since all errors in it are fixed

eddiemuc:
lint: remove baseline since all errors in it are fixed


Lineflyer:
Cool!

Issue #10202, opened 2021-03-18
App not starting after wizard completed


Lineflyer:
From support mail for beta version 2021.03.16-RC1:

Xiaomi Mi9, (probably) Android 10

User states:
- Updating from Google Play OK
- Wizard showed up on first start 
- As base folder he used the suggested (/cgeo) where he could see his existing data
- Setting other folders worked as well, he reports some hanging while selecting the map theme folder (so he went back, started the step again and could complete it without problem)
- He completed it succesfully (Have fun!... "Finish").
- c:geo started normally
- The user wanted to check whether there are map updates (using our downloader)
- Then he changed to his mail app to check an incoming mail
- Changing back to c:geo the c:geo went black instead
- Restarting c:geo = Black screen, nothing shown
- Same after reboot of device

Any clue from that or any data we might need (or can get) from the user?

Lineflyer:
Anything I can ask the user to do?
Otherwise I would recommend a fresh install to him. :(

bekuno:
Did he wait at least a half minute?
We have sometimes this startup-delay. But in this situation he should have the splash screen.  Hmm.

eddiemuc:
We have now the option of cgeo writing a separate logfile. User needs to place a log config file in logfiles folder. I can prepare such a config file if you're interested?

eddiemuc:
Here would be such a log config file.

[log-properties.txt](https://github.com/cgeo/cgeo/files/6172072/log-properties.txt)

The user should:
* place this file in its c:geo logfiles directory
* Start c:geo. This should result in a new file created in logfiles directory in format `cgeo-log-startup-(timestamp).txt`. This file should grow while c:geo is running
* Wait until black screen shows, wait some more, then kill app
* Send the created logfile to support

This would be the first usage of this new option in the wild, so I hope it works...

Lineflyer:
Sent to the user.

Lineflyer:
The first user gave feedback, that despite placing the log-properties.txt (as attached above) there was no logfile created.

Now we have a second user with a similar case:

---

When the beta was first installed and run, it led me though the wizard, and I attempted to answer the questions. At one point, it told me that I had picked a different directory, so I went back and tried again. On completion the app showed me my list of caches, and I selected the top one, but it didn't open it - the app appeared to hang.

[...]

PR #10262, opened 2021-03-26
rel to #10202: put safeguard: dont scan theme files larger than 5MB

eddiemuc:
rel to #10202: put safeguard: dont scan theme files larger than 5MB


moving-bits:
@eddiemuc 
Shouldn't this be targeted to `release` branch?

eddiemuc:
@moving-bits thanks for the catch, indeed it must be release

Issue #10266, opened 2021-03-27
Make settings_folder_changed  clearer


Lineflyer:
The current text is a bit difficult to understand in combination with a restore process.

Setting for folder "%1$s" will be restored to "%2$s".\n\nTap "%3$s" to keep current setting for this folder.\n\nTap "%4$s" to set the old folder. You will be asked to regrant c:geo access to that folder in that case.

How about using "restored folder" instead of "old folder", or even:

Folder "%1$s" will be changed to "%2$s" during restoration.\n\nTap "%3$s" to keep the existing folder or tap "%4$s" to set the restored folder instead. In the last case you will be asked to regrant c:geo access to that folder.

PR #10267, opened 2021-03-27
check missing file permission on restoring settings (fix #10215)

### LINKED ISSUE #10215: SAF: Check missing file permission on restoring settings
Reminder to myself:

When restoring settings check file permissions for persistable files against the current settings (currently loaded track only). If they differ, request grant for it.

Similar to what we have done with #10136 and #10213 for persistable folders.

moving-bits:
Similar to what we have fixed for folders already: Check settings for persistable files on restore and request re-grant if setting has changed.

image


Lineflyer:
Please see #10266 regarding a text proposal which also applies here.

fm-sys:
Do we really need to restore these single files? IMHO this complicates the restore process a lot. I would vote for just unloading these files if they aren't accessible anymore...

moving-bits:
IMHO we should keep it, as I would expect all settings to be restored from a backup.

Most users will not be affected anyway, as they don't have a trackfile set, so it does not hurt them. But if they have set one, they (or at least: me) would expect it to be restored. (And even those will only get asked, if they have changed their trackfile setting meanwhile.)

Additionally, this mechanism is generic, so if we happen to introduce more `PersistableUri` at some point in time, they will be included automatically in this backup/restore mechanism.

PR #10268, opened 2021-03-27
Fix #10266 - Improve settings_folder_changed

### LINKED ISSUE #10266: Make settings_folder_changed  clearer

Lineflyer:


moving-bits:
I rearranged the two options to keep the "positive" actions together, and rephrased it a bit for better readability:

`Folder \"%1$s\" will be changed to \"%2$s\" during restoration.\n\nTap \"%4$s\" to proceed. In that case you will be asked to regrant c:geo access to that folder.\n\nTap \"%3$s\" to keep the current folder setting instead.`

I will adapt the text in #10267 according to what we agree upon here.

Lineflyer:
Sounds good. Will adapt my PR and merge then.

Lineflyer:
ups...have to redo it

Issue #10224, opened 2021-03-20
How to make theme users aware of improvement by zip/sync?


Lineflyer:
Up to now we included improvement infos (use zipped theme, sync theme) at two places:
- In Settings - Map - Info on Map Themes
- In our FAQ

Both are not places where John Average will look at.
As you can see in the beta survey (and as I can see on support mail) some users notice this map start delay.
Some of those might contact us, but most will probably just think "It got worse" and do nothing.

I am no explicit fan of those one time messages (especially as it would be the third one on map) but maybe it would be just another case, where it might be helpful?!
It would just need to get shown under special conditions:
- Offline map used
- User defined theme active
- Sync is off
- Theme is not zipped

It could contain the text of the second section as in "Info on Map Themes" and a link to https://faq.cgeo.org/#theme-slow

BTW:
The link behind "More" in "Info on Map Themes" should rather target to that link as well instead of the user guide.

But maybe someone has a better proposal than using an one time message?!

moving-bits:
For this part

> BTW:
> The link behind "More" in "Info on Map Themes" should rather target to that link as well instead of the user guide.

see #10290

eddiemuc:
Ah I oversaw this one...

If someone suggests a text for the one time message I will implement it (with conditions as suggested by @Lineflyer )

Lineflyer:
Taken from `string name="settings_info_themes"` and updated. The string should be synced from its wording with the OTM:

```
Working with non-zipped map themes might decrease map viewer opening performance. You can zip or sync your themes to get faster performance and reduce the delay on starting the map.
Refer to our <a href="https://faq.cgeo.org#theme-slow">FAQ</a> for a detailed description.
```

eddiemuc:
PR #10305. I must say, this was really easy given the great OneTimeDialog-framework!

PR #10290, opened 2021-03-29
change 'more' link for map themes (rel. to #10224)

moving-bits:
related to #10224: Change link behind "more" on themes info to point to FAQ entry about syncing


Lineflyer:
Thanks!

PR #10299, opened 2021-03-31
fix #10295: make freizeitkarte theme work

### LINKED ISSUE #10295: Theme not working normal if unzipped
From support mail + Facebook:

- Use Freizeitkarte_DEU+SUED.map
- Use freitzeitkarte_V5 theme (in unzipped format)
- Open map around N50° 13,744' / E011° 15,864', you will see a large forest area (green background)
- Zoom closer, at some point the forest area (green background) vanishes
- The detailed problem seems to be, that at this zoom level the green area should be replaced by another marker (green area with tree icons)

Here comes the strange part:
If you use the same theme as *.zip the problem is gone and the theme works as it should AFAICS

While using the *.zip is a valid workaround, we should maybe try to find out the root cause, as this problem could also apply to other themes and/or to more of the theme contents been missing at some point.

eddiemuc:
fix #10295: make freizeitkarte theme work

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -309,7 +309,7 @@ dependencies {
     implementation "com.asamm:locus-api-android:0.9.28"
 
     // Mapsforge new version
-    def mapsforgeVersion = '9cdb3396fa'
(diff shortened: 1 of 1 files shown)


Lineflyer:
retest this please

PR #10304, opened 2021-04-01
fix #10176: migrate brouter to saf

### LINKED ISSUE #10176: BRouter integration: Migrate access of routing tile data to SAF

eddiemuc:
fix #10176: migrate brouter to saf

### DIFF
--- a/main/src/cgeo/geocaching/brouter/BRouterWorker.java
+++ b/main/src/cgeo/geocaching/brouter/BRouterWorker.java
@@ -14,7 +14,7 @@ import java.util.List;
 
 public class BRouterWorker {
     // public String baseDir;
-    public String segmentDir;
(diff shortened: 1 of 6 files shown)


eddiemuc:
This code has the small downside that it does not work. Have to investigate

bekuno:
retest this please

moving-bits:
superseded by #10404

PR #10305, opened 2021-04-01
fix #10224: introduce otm for slow theme usage

### LINKED ISSUE #10224: How to make theme users aware of improvement by zip/sync?

eddiemuc:
fix #10224: introduce otm for slow theme usage

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1339,6 +1339,8 @@
     <string name="init_summary_renderthemefolder_synctolocal">This option may improve map viewer opening performance at the cost of some device storage space.</string>
     <string name="init_renderthemefolder_synctolocal_dialog_title">Turn on theme folder synchronization</string>
     <string name="init_renderthemefolder_synctolocal_dialog_message">This will permanently synchronize the content of the Map Theme folder into an app-private folder for faster file access.\n\nYour map Theme Folder \'%1$s\' has currently %2$s, %3$s and a size of %4$s.\n\nDo you want to turn on folder synchronization?</string>
+    <string name="onetime_mapthemefixslow_title">Fix slow map theme rendering</string>
+    <string name="onetime_mapthemefixslow_message">Working with non-zipped map themes might decrease map viewer opening performance. You can zip or sync your themes to get faster performance and reduce the delay on starting the map.</string>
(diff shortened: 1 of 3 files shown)


eddiemuc:
Whenever a "slow map theme" is selected (non-zipped, non-sync) or map viewer is opened and such a theme is preselected, then following one-time-dialog appears:

image

The "More Information" button leads to the FAQ-page

fm-sys (approved):
LGTM 👍

Lineflyer:
@fm-sys For future: Feel free to merge if you reviews and/or tested a PR and it looks good to you. :)

fm-sys:
> @fm-sys For future: Feel free to merge if you reviews and/or tested a PR and it looks good to you. :) 

Ok, will do :-)

Just wanted to keep this PR open a little longer, cause I reviewed like 3 minutes after CI got green...

Lineflyer:
Ok, thats reasonable as well. Did not pay attention to the timeline in this case. 
For sure 1-2 days time depending on urgency might be wise, to give others a chance to review as well.

But anyway @cgeo/team:
Feel encouraged also to review merged PRs...we can still revert or fix issues for those.

eddiemuc:
Wrt merging PRs, after talking to @moving-bits some time ago I now follow the practice to merge self-insure non-critical small PRs myself after leaving 1-2 days review time - except of course something comes up during this time.

PR #10314, opened 2021-04-02
remove 'cancel' option from list selection (fix #9879)

### LINKED ISSUE #9879: Action items on list selection no longer in one line
Reported for 2021.01.24-RC and reproducible for me also in 2021.01.24-NB1

- Open map
- Tap unsaved cache
- Hit save icon in popup
- List selection dialog is shown

Result:
The "OK" / "Cancel" / "Last selected" have line breaks each which leads to a unnecessary scrollable view and not all items are visible and usable instantly.

grafik

Expected Result:
Have all items in one line

moving-bits:


moving-bits:
retest this please

Lineflyer:
Thanks @moving-bits 
More a "give-up" than a fix ;))

moving-bits:
> More a "give-up" than a fix ;))

Yes, I guess there is no real fix, as it seems we cannot influence the button part of the dialog.

Question is whether we want to implement a different approach like the one described in https://github.com/cgeo/cgeo/issues/9879#issuecomment-805218981?

PR #10323, opened 2021-04-04
SAF-Helper: introduce FileByteReader

eddiemuc:
SAF-Helper: introduce FileByteReader
This will help implementing #10176

### DIFF
--- a/main/src/cgeo/geocaching/storage/FileByteReader.java
+++ b/main/src/cgeo/geocaching/storage/FileByteReader.java
@@ -1,0 +1,61 @@
+package cgeo.geocaching.storage;
+
+import androidx.annotation.NonNull;
+
+import java.io.Closeable;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+
+import org.apache.commons.io.IOUtils;
+
+/**
+ * Helper to read files when only a {@link java.nio.channels.FileChannel} is available.
+ *
+ * Mimics the necessary parts of RandomAccessFile
+ */
+public class FileByteReader implements Closeable {
+
+    private final FileChannel fileChannel;
+
+    public FileByteReader(final FileChannel fileChannel) {
+        this.fileChannel = fileChannel;
+    }
+
+    public FileByteReader(final FileInputStream fis) {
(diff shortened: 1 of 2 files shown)


eddiemuc:
Retest this please

eddiemuc:
Retest this please

eddiemuc:
Retest this please

moving-bits:
@eddiemuc 
Thanks for providing `FileByteReader` class. I've finally reached a stage with BRouter integration where I felt prepared to tackle SAF migration of routing tile data.

I've tackled the calling chain and reached the grounds of `PhysicalFile`. Instead of a `File` I provide the `InputStream` returned by `ContentStorage.get().openForRead(PersistableFolder.ROUTING_TILES.getFolder(), filenameBase + ".rd5")`.

But now I'm stuck in the class constructor of `PhysicalFile`: I guess `FileByteReader` is meant to replace `RandomAccessFile`?

If I try to replace the old `ra = new RandomAccessFile(f, "r")` with `ra = new FileByteReader(is)` it does not work, as a `InputStream` is not accepted as parameter, it has to be a `FileInputStream`, but there is no method in `ContentStorage` returning a `FileInputStream`.

Changing `FileByteReader` to accept `InputStream` instead of `FileInputStream` does not help either, as `InputStream` does not provide a `getChannel()` method which `FileByteReader` seems to rely on heavily.

(Have not dived any deeper in replacing `RandomAccessFile` with `FileByteReader` - many more changes needed there, as there are quite a few methods being called from `RandomAccessFile`, but not supplied by `FileByteReader` - but haven't checked yet, whether they will still be needed.)

Can you give me some hint on how to continue?
I've put the current intermediate state (not yet compiling!) into a branch `brouter_saf_tilefiles` in my repo: https://github.com/moving-bits/cgeo/tree/brouter_saf_tilefiles

eddiemuc:
@moving-bits short answer is that the InputStream returned by scoped storage is in fact a FileInputStream IF it references a local file. So you can safely cast it. Same is done for map files btw (and I learned the trick from @pstorch)

But if you want I can tackle this issue further. I just forgot about it last couple of days.

moving-bits:
Ah, I see, didn't know that. Thanks for the pointer! I'll try to continue working on this the next days, let's see how far I will come :-)

PR #10329, opened 2021-04-04
skip checking folders on database-only restore (fix #10325)

### LINKED ISSUE #10325: Base folder checked while only restoring caches/waypoints
- Own developer build from current `release` branch
- Freshly install c:geo and complete the installation wizard
- Call Backup/Restore
- Select to restore but only "caches and waypoints"

Result:
During the restoration process c:geo asks to regrant access to the base folder

Expected result:
Two problems here:
1. The backup should contain the same base folder as currently set
2. I did not restore settings, so why is this shown after all as the process should only restore the database

moving-bits:


moving-bits:
retest this please

moving-bits:
retest this please

PR #10343, opened 2021-04-05
Geocache.findWaypoint() has unreachable code

bekuno:
The code `if (null != prefix) {` is every time true.
Therefore no check for matching coordinate or name is done.

The fix removes the useless `return null` statements and the useless null check.

@murggel: You created this code part. Do you have some test data? The I would try to create a test for this method.

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -1652,21 +1652,19 @@ public class Geocache implements IWaypoint {
         return changed;
     }
 
-    private Waypoint findWaypoint(final Waypoint searchWp) {
-        //try to match prefix
+    @Nullable
+    private Waypoint findWaypoint(@NonNull final Waypoint searchWp) {
+        // try to match prefix
         final String prefix = searchWp.getPrefix();
-        if (null != prefix) {
-            for (final Waypoint waypoint : waypoints) {
-                if (!StringUtils.isBlank(prefix) && !StringUtils.isBlank(waypoint.getPrefix()) && prefix.equals(waypoint.getPrefix())) {
-                    return waypoint;
-                }
+        for (final Waypoint waypoint : waypoints) {
+            if (!StringUtils.isBlank(prefix) && !StringUtils.isBlank(waypoint.getPrefix()) && prefix.equals(waypoint.getPrefix())) {
+                return waypoint;
             }
-            return null;
         }
 
-        //try to match coordinate
+        // try to match coordinate
         final Geopoint point = searchWp.getCoords();
-        if (null != point) {
+        if (point != null) {
             for (final Waypoint waypoint : waypoints) {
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

bekuno:
retest this please

murggel:
> 
> 
> The code `if (null != prefix) {` is every time true.
> Therefore no check for matching coordinate or name is done.
> 
> The fix removes the useless `return null` statements and the useless null check.
> 
> @murggel: You created this code part. Do you have some test data? The I would try to create a test for this method.


@bekuno 
I don't know, why I put the null check around it - should be  to avoid the then useless loop.
There should be already some UnitTest for parsing the waypoints from note - let me check, why the ones for looking for matching name passed ...

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1652,32 +1652,29 @@ public boolean addWaypointsFromText(@Nullable final String text, final boolean u
...
+    @Nullable
+    private Waypoint findWaypoint(@NonNull final Waypoint searchWp) {
+        // try to match prefix
         final String prefix = searchWp.getPrefix();
-        if (null != prefix) {
-            for (final Waypoint waypoint : waypoints) {
-                if (!StringUtils.isBlank(prefix) && !StringUtils.isBlank(waypoint.getPrefix()) && prefix.equals(waypoint.getPrefix())) {
-                    return waypoint;
-                }
+        for (final Waypoint waypoint : waypoints) {
```
murggel, line 1659: If you put 
 `if (!StringUtils.isBlank(prefix)) ` 
before the for-loop, you can avoid the then useless loop.

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -1652,32 +1652,29 @@ public boolean addWaypointsFromText(@Nullable final String text, final boolean u
...
+            if (!StringUtils.isBlank(prefix) && !StringUtils.isBlank(waypoint.getPrefix()) && prefix.equals(waypoint.getPrefix())) {
+                return waypoint;
             }
-            return null;
```
murggel, line 1664: keep the `return null`, because if a certain prefix is searched and no waypoint with that prefix is found, I assume a new one should be created.

murggel:
@bekuno 
The` return null;` for the geopoint-search is leading to a bug, because having a waypoint with name and coordinates in the personal note and a waypoint with same name already exist, a new waypoint is extracted instead of "reusing" the exising one.

I adapted locally the testcases (based on master), How can I share them? Shall I open a new PR?

bekuno:
@murggel You can create a complete PR. It should target to release.

murggel:
> 
> 
> @murggel You can create a complete PR. It should target to release.

PR #10346 created

Lineflyer:
Replaced by #10346

PR #10351, opened 2021-04-07
Set individual permissions titles in wizard

moving-bits:
Set individual titles for location and storage permissions in the wizard. (See https://github.com/cgeo/cgeo-guide/issues/25#issuecomment-814691333)

Existing strings are reused, so no need to wait for new translations here.

image.image


Lineflyer:
I would suggest not do this for `release` but for `master` only, as it will require new translations which will potentially not come in short termed for our planned release.

Lineflyer:
Sorry, I skipped reading your PR comments. Nevermind, well done :)

PR #10356, opened 2021-04-07
Improve issue and PR templates

Lineflyer:
## Description

- Reworked existing issue templates according to experience from practice
- Use comments as help for using the template (will be removed automatically in final issue)
- Use the new "allow blank issues" in `config.yml` instead of having an additional empty template
- Include a link to our support page as third option (instead of opening an issue on Github)
- New reworked approach for a PR template (as you can see here already)

## Related issues

None

## Additional context

Feel free to make proposals in this PR if you see more improvements / would like to include anything else

The issue selector will  then look like this (note the blank issue field below the box):
grafik


moving-bits (approved):
LGTM, thanks for the updates

PR #10357, opened 2021-04-07
add application/gpx to accepted MIME types (fix #10309)

### LINKED ISSUE #10309: c:geo not offered to open GPX with MIME type: application/gpx
**Describe the bug:**
Since some months I noticed, that c:geo is no longer offered if I try to open a GPX file I received by mail in my mail client.
Other GPX aware apps are offered though

**To Reproduce:**
1. Receive GPX exported from c:geo from another user
2. Open mail
3. Click the attached GPX

Offers Locus and another routing app 

c:geo should also be offered

**Version of c:geo used:**
2021.04.02-NB

Yes

**Screenshots:**
This is the intent triggered by my mail client, might be helpful to fix the problem I guess:
Screenshot_20210402-125016_Intent Intercept
Screenshot_20210402-125025_Intent Intercept

**System information:**
Samsung S20, Android 11

**Additional context:**
When I try to open a GPX using my file explorer, the intent has MIME type `application/octet-stream` and c:geo is offered as target app

moving-bits:
I've targeted this to `release`, but if we want to test it a bit, we can retarget it to `master` as well.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -373,6 +373,7 @@
                 <data android:mimeType="application/x-zip-compressed" />
                 <data android:mimeType="application/x-zip" />
                 <data android:mimeType="application/octet-stream" />
+                <data android:mimeType="application/gpx" />
                 <data android:pathPattern=".*\\.gpx" />
                 <data android:pathPattern=".*\\.zip" />
                 <data android:pathPattern=".*\\.ggz" />


fm-sys:
I think the fix is pretty much straight forward, so no unwanted side affects are to be expected...

Lineflyer:
I think it should be relatively safe to put this on `release`. The worst, which could happen is, that c:geo is offered for too many files, which I doubt.

PR #10370, opened 2021-04-11
Prepare release

moving-bits:
As agreed upon in today's meetup two things in preparation for the next release:

- disable OSM multi-threading by default
- display number of available processors in about - system page


Lineflyer:
Question (did not check):
Do we have the setting for "Mult thread" itself in the sysinfo? Else, it could be included and the amount of processors appended to it.

moving-bits:
> Do we have the setting for "Mult thread" itself in the sysinfo? Else, it could be included and the amount of processors appended to it.

I'll add it.

Issue #527, opened 2011-09-22
Default handler for mime type application/gpx+xml


Olen:
It would be great if c:geo was set as a handler for mime-type "application/gpx+xml", and would try to add these to stored caches automatically.

PR #10377, opened 2021-04-12
add application/gpx+xml to accepted MIME types

fm-sys:
Similar issue as in #10309 which happened to me in my file manager. Locus is listed but c:geo isn't. See also issue #527 as a reference, in which the MIME type was suggested/~added~ as well years before.

4 Size)


Lineflyer (approved):
Yes, maybe another such specialty of some file managers.

PR #10382, opened 2021-04-14
Optimize RouteSortActivity

fm-sys:
When using RouteSort, I always struggle about how saving is implemented. This PR will:
- protect RouteSortActivity from unwanted discard of changes
- optimize the user workflow by finishing RouteSortActivity after saving automatically 
(Why should someone click `save` if he is not finished yet?)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -273,6 +273,7 @@
     <string name="err_trackable_log_not_anonymous">Logging %1$s from c:geo cannot be done anonymously. Do you want to open the settings to fill in your credentials from %2$s?</string>
     <string name="err_trackable_no_preference_activity">Sorry, cannot find a suitable preference screen.</string>
     <string name="confirm_unsaved_changes_title">Unsaved changes</string>
+    <string name="confirm_discard_changes">Discard unsaved changes?</string>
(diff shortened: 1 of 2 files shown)


moving-bits:
I like the idea of combining saving and closing, thanks for bringing this up.

One idea though: On waypoint edit we have the "other way around" approach, meaning: There is no explicit save, but on closing the activity c:geo asks you, whether you want to save your changes (if any). Can we use the same approach here for consistency?

eddiemuc:
On offline log edit, on the other hand, we have autosave on closing the activity without asking the user. Seems like we have different approaches in different places of c:geo anyway.

fm-sys:
For waypoint edit, it is exactly the same as how I have done it here:

Screenshot_20210414_225241_cgeo geocaching_copy_270x585


First, I wanted to implement it exactly as suggested by @eddiemuc, but I think having a cancel option is not a bad idea either ;-)

But it is true that we have a cancel option im more than 10 places and handle it differently every time. Here a bit of unifying would be necessary...

moving-bits:
> For waypoint edit, it is exactly the same as how I have done it here:

Argh! You are right - sorry for the confusion.

Issue #10369, opened 2021-04-11
Avoid oversized avatar image


Lineflyer:
## Bug description 

Sometimes users fail to find the menu, where they can check their gc.com credentials and/or reauthorize although reading our FAQ and/or user guide.  Last days I had contact to a user who also failed to find it, so I asked for screenshots to guide him through. 
It turned out his avatar picture was blocking the relevant menu (see screenshot below).

Now, that I have seen this, it explains quite some support tickets, where people had problems to follow, what I instructed them to do.

### Steps to reproduce the problem

- Open c:geo Menu - Settings - Services
- Click on geocaching.com

### Actual result after these steps 

For some users the avatar image in that menu is just oversized blocking out the text next to it

### Expected result after these steps 

We should enforce a certain size of that image (in percentage of screen), so that it always looks as defined/desired

## c:geo version 

2021.02.14

## Reproducible 

No, I tried uploading several new avatar images on gc.com, but the scale in our menu was always OK.

## System information 

```
--- System information ---
Device: SM-G975U (beyond2qltesq, samsung)
Android version: 11
Android build: RP1A.200720.012.G975USQU5GUBH
c:geo version: 2021.02.14
Google Play services: disabled - 21.09.15 (150400-361652764)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide caches: -
Hide waypoints: -
HW acceleration: enabled (default state)
System language: en_US / user-defined language:
System date format: M/d/yy
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (33.5 GB free) internal
User storage c:geo dir: /storage/emulated/0/cgeo (33.4 GB free) external non-removable
Geocache data: /storage/emulated/0/Android/data/cgeo.geocaching/files/GeocacheData (33.4 GB free) external non-removable
Database: /data/user/0/cgeo.geocaching/databases/data (2.1 MB) on system internal storage
Last backup: never
GPX import path: /storage/emulated/0/cgeo/gpx
GPX export path: /storage/emulated/0/cgeo/gpx
Offline maps path: null
Map render theme path:
Live map mode: true
Global filter: display all caches
Sailfish OS detected: false
Fine location permission: granted
Write external storage permission: granted
Geocaching sites enabled:
   geocaching.com: Not logged in (Login failed)
[...]

PR #10390, opened 2021-04-15
restrict avatar size

moving-bits:
Tries to fix the oversized avatar image (#10369) by limiting their relative column width.

@Lineflyer 
Not sure if this works reliably, as I cannot reproduce the issue. Can you cross-check?

### DIFF
--- a/main/src/cgeo/geocaching/settings/CredentialsPreference.java
+++ b/main/src/cgeo/geocaching/settings/CredentialsPreference.java
@@ -120,6 +120,9 @@ public class CredentialsPreference extends AbstractClickablePreference {
         frame.setVisibility(View.VISIBLE);
         frame.addView(iconView);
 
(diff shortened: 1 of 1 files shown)


Lineflyer:
I would have to try how to enforce such oversized avatar. Will try on a BM account as time permits.

Lineflyer:
Could't you test it by replacing the stored image in c:geos private dirs?

moving-bits:
> Could't you test it by replacing the stored image in c:geos private dirs?

Thanks for the pointer, will try it. Need to lookup where the file is located.

moving-bits:
It seems to work fine:
I've updated the cached avatar image file by a different one, being twice as wide (but having same height), and it got displayed completely, without being distorted, and without pushing the text to the left.
So let's merge and give it a try...

PR #10395, opened 2021-04-17
avoid loop when using "open in browser" for trackables

fm-sys:
Avoid loop when using "open in browser" action, in case c:geo is set as default app for trackable links


fm-sys:
retest this please

PR #10408, opened 2021-04-18
replace icon-creation by ribbonizer by dedicated (ribbonized) images

### LINKED ISSUE #8896: ribbonizer does not work with adaptive launcher icons (from Android 8.0)

bekuno:
## Description

The icon which is to change is primary defined in the AndroidManifest.xml.
But also it is used in the about screen (and now in the configuration wizard).

To achive a dedicated icon for different build types we have two possibilities.
At first we can let substitute this (and this is also possible for other entries in AndroidManifest) by declaring a manifestPlaceholders definition in the gradle buildTypes definition, e.g. `manifestPlaceholders = [appIcon: "@mipmap/ic_launcher"]` and using it         `android:icon="${appIcon}"`.
This would be a little code change. But in this way we can only manipulate the AndroidManifest.xml.
The other way is the possibility to define an additional `res` directory for the build type(s). There we need only the changed files. The big advantage is that we can refer to the changed ressource in the whole project (what we need for the about screen).
The main code change is that we have to move the main ressource directory to his "normal" place (from Android point of view).

I decided to go the second way in the PR.

Additionals:
- The management for keys.xml was adopted.
- The res folder mipmap-anydpi-v26 has xml files without ribbon, therefore saved to rawimages.
  Maybe we can create adaptove icons later with ribbon (or another colored? difference).

For the ribbon I used different colors and strings to easier differentiate the build types. Used rc/beta=red, nightly=black, debug=blue. The legacy build type has no dedicated icon.
This is open to discussion.

## Related issues
fixes #8896
prereq for upgrade gradle plugin to version 7.0


bekuno:
Samples:
ic_launcher
ic_launcher
ic_launcher

bekuno:
retest this please

Lineflyer:
Nice :)
Thanks for finding a new solution.
As posted in the issue thread, it might be very helpful for average users to easily recognize whether they currently use a beta or nightly version.

Lineflyer:
BTW: 
Regarding the method you chose:
In fact we do only need to change the manifest, so I don't yet understand why you chose the "more extensive" change instead? Only to have the icon also in the about section and title screen? Maybe the launcher icon would be enough to avoid such folder structure change in the source.

bekuno:
currenty failing, because jenkins would need a change, see
`rm: cannot remove 'main/res/values/keys.xml': No such file or directory`

bekuno:
> 
> 
> BTW:
> Regarding the method you chose:
> In fact we do only need to change the manifest, so I don't yet understand why you chose the "more extensive" change instead? Only to have the icon also in the about section and title screen? Maybe the launcher icon would be enough to avoid such folder structure change in the source.

Exactly, to have the icon in the about section.

bekuno:
@Lineflyer would the crowdin.yml change enougt to stay in sync with crowdin or is also a change at crowdin necessary?

Lineflyer:
> @Lineflyer would the crowdin.yml change enougt to stay in sync with crowdin or is also a change at crowdin necessary?

Not sure, did you also change the location of the source (english) string file?
This would also need adaption then.

Shall we really do this major folder change for the benefit of the icon in the about section?

bekuno:
> 
> In fact we do only need to change the manifest, so I don't yet understand why you chose the "more extensive" change instead? 

An other advantage of changing the folder possition is that we then use the default res folder possition. So that it would be easier to configure flavor and buildType if necessary.

bekuno:
> Not sure, did you also change the location of the source (english) string file?

The strings folders are within the res folder.
I changed the configuration to
~~~
  - source: /main/src/main/res/values/strings.xml
    translation: /main/src/main/res/values-%two_letters_code%/%original_file_name%
~~~

bekuno:
retest this please

bekuno:
> currenty failing, because jenkins would need a change, see
> `rm: cannot remove 'main/res/values/keys.xml': No such file or directory`

I changed in the script `cgeo build and test for new commits and pull requests`
the last line to
`rm main/res/values/keys.xml || rm main/src/main/res/values/keys.xml` 
to support the old and new location.

fm-sys:
> Maybe the launcher icon would be enough to avoid such folder structure change in the source.

Maybe it's just me, but I would like to have a possibility to disable the ribbonizer totally 🙈

Ok, but if we want to really reimplement the ribbonizer:
- Why does the banner have to protrude beyond the icon edges, which make the actual cgeo icon much smaller?
- You need to adapt the translation merge strategy as well (see https://github.com/cgeo/cgeo/blob/master/.gitattributes)

bekuno:
> Maybe it's just me, but actually I would like to have a possibility to disable the ribbonizer totally see_no_evil

We need now (to make the next gradle plugin update) to remove the ribbonizer integration.
I replace it in the PR by dedicated icons for each buildType (see the icons above).

> Ok, but if we want to really reimplement the ribbonizer:
> 
>     * Why does the banner have to protrude beyond the icon edges, which make the actual cgeo icon much smaller?

The icons are created with the (old) ribbonizer.
We can replace them by nicer one, maybe also xml based.
 
>     * You need to adapt the translation merge strategy as well (see https://github.com/cgeo/cgeo/blob/master/.gitattributes)

Thanks, will have a look.

Lineflyer:
Any other opinions about the necessity to include the ribonized icon into the about section?
IMHO we could leave it with the launcher and spare the folder move (mainly because I am afraid of the side effects possible on Crowdin).

Lineflyer:
> > You need to adapt the translation merge strategy as well (see https://github.com/cgeo/cgeo/blob/master/.gitattributes)
>
> Thanks, will have a look.

Puh...not sure how this will work. 
Please note, that after merge it will in general be really difficult to merge changes from `release` to `master` if they touch the changed structure (as translations do).

moving-bits:
> Any other opinions about the necessity to include the ribonized icon into the about section?

Although it's nice to have different icons for launching the app (thanks for tackling this!), personally I do not have the additional need for different icons in the about screen. Especially not, if it comes at the price of yet another level in the folder hierarchy, which I already currently loathe as being too deep. (Although I have to admit that this is partly a matter of taste.)

If we want to show it in the about screen as well:
Couldn't we just set a flag during build time which gets evaluated at runtime, and if it points to a "nightly" build (etc.), display an additional badge on top of the c:geo icon? Should be possible by using a `RelativeLayout` and using image files for the badges only. (We do something similar on the mainscreen.)

Lineflyer:
@bekuno 
Can we convince you to apply the minimum solution only? I personally think and I also read from the posting of @moving-bits, that this should be sufficient and avoid the structural change.

bekuno:
I planned already to present the other solution. I did not had time until now, but will do it the next days.

bekuno:
>     * Why does the banner have to protrude beyond the icon edges, which make the actual cgeo icon much smaller?

The actual c:geo icon (for devices API 26++ = Android 8.0) is based on an adaptive icon https://github.com/cgeo/cgeo/blob/master/main/res/mipmap-anydpi-v26/ic_launcher.xml. In this implementation with the ribbon you see the old style png version.

For the new implementation I tryed to optimize the adaptive icon too. I played with modifying the foreground image part with adding the banner, but this is not fine (would created outside the visibile are on rounded icons). 
Therefore I will change the background ("map") color.

bekuno:
superseded by the implementation in #10468

Issue #10406, opened 2021-04-18
Significant higher ANR rate in 2021.04.15 (SAF) compared to 2021.02.14 (preSAF)


Lineflyer:
Statistics of today comparing those two releases (while still at 50% rollout rate):
![grafik](https://user-images.githubusercontent.com/949669/115157200-3af40880-a088-11eb-8fe7-0060ee49239d.png)

Will still continue to roll out but maybe someone could have a look into the Play Console if there is any specific finding, which we can optimize.

Lineflyer:
Looking at the scale it might still be not so dramatic:
The percentage raise is high, but the absolute values still  look rather low.

Lineflyer:
Right now two new users with c:geo hanging on the splash screen (running into ANR).

@eddiemuc Is there any evaluation/check of folder contents involved in the startup process. It should not...just asking. For one user I have seen that he set the same folder for maps and themes and its containing >=320 dirs and 8GB of data. I guess this should be irrelevant for starting the app...just want to make sure.

eddiemuc:
There should be no folder scan on startup. Also such folder scans, even if they are happening, should be limited to max. 5 subdirs to prevent overly long scan times.

eddiemuc:
@Lineflyer I am not so used to Google Play Console up to now. Is this the right place to look in ("Abstürze und ANRs", there's a long list of concree errors on the right which I don't want to copy here)?

![image](https://user-images.githubusercontent.com/6909759/115188538-5a6d4e80-a0e5-11eb-8041-02d29787cd96.png)

Lineflyer:
> I am not so used to Google Play Console up to now. Is this the right place to look in ("Abstürze und ANRs", there's a long list of concree errors on the right which I don't want to copy here)?

Yes, thats correct.
You can use the functions on top to e.g. filter on app version and filter on ANRs.

Lineflyer:
> There should be no folder scan on startup. Also such folder scans, even if they are happening, should be limited to max. 5 subdirs to prevent overly long scan times.

I received one users feedback, which is:

---

I noticed, that c:geo was extremely slow when dealing with the map settings menu. Therefore I renamed my existing map folder, created a new, empty map folder and assigned it in c:geo. Then I copied the maps over there and since then everything is working smoothly and c:geo starts up fast.

---

It will probably not easy to find the cause of this problem, however I am wondering why this affects the startup time after all.

This is the initial system information of the user (before he did the change described above):
```

--- System information ---
[...]

PR #10412, opened 2021-04-19
Issues/slow start theme 10406

eddiemuc:
related to #10406: aborts xml map theme scan after 2 seconds

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2153,6 +2153,7 @@
 
     <!-- Map Theme helper -->
     <string name="mapthemes_foldersync_finished_toast">%1$s folder synchronization finished after %2$s, %3$d of %4$s needed synchronization</string>
+    <string name="mapthemes_themescan_too_long_toast">Map theme scan took more that %1$d seconds, aborted after finding %2$d themes</string>
 
 
     <!-- default actions in dialogs -->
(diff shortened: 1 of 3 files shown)


Lineflyer:
Thanks @eddiemuc 

1. Any estimation how many files/dirs a typical device will return in those two seconds? It seems kind of a fast timeout?
2. Will themes not found in these two seconds be unavailable on the map or is there another scan when starting the map?

eddiemuc:
My assumption was that you wanted a quick fix to get out in production before weekend. Since no beta test is available it has to be a small change. This is why I introduced the time limit.

I can increase it easily e.g. to 5 or 10 seconds. What is the time limit triggering an ANR?

A better solution would of course be to put the scan to the background as @moving-bits suggests. But this is a bit more complicated to achieve (e.g. what happens if user opens map viewer before scan is finished - which can easily happen if it takes >5 seconds) and I may not have the time during the week to do this. Would You rather prefer to wait for a "better" fix then?

eddiemuc:
Time limit increased to 5 seconds

Lineflyer:
@moving-bits Are you OK with merging this as a first workaround? So we could push an update rather soon and have time afterwards to find a more suitable solution.

moving-bits:
> @moving-bits Are you OK with merging this as a first workaround? So we could push an update rather soon and have time afterwards to find a more suitable solution.

Sure. Haven't had the time to check the technical details of the PR, but the approach itself (limiting the time) sounds pretty reasonable as a workaround.

PR #10422, opened 2021-04-20
Optimize strings for our notification system

Lineflyer:
## Description

We do have a set of predefined strings we can use while pushing a notification to users. The advantage compared to a free text is, that those strings can be localized.
I think however the current wording is relatively limited, e.g. most of the time a change is not "breaking c:geo" completely, but only impacting some features.

Therefore I tried to phrase them more open, so that we can use them more often.

## Related issues

None


Lineflyer:
retest this please

moving-bits:
Thinking of that there may be more hotfixes to come the next days: Do you want to merge to `master` or to `release`?

Lineflyer:
Will redo for release after upcoming release.

Issue #10421, opened 2021-04-20
Cache type and owner parsing not OK (Website change)


Lineflyer:
## Bug description 

Users report problems, that cache type and partly other information is missing after loading cache details.
There was also a website change today, so it might be caused by that:
https://forums.geocaching.com/GC/index.php?/topic/361563-release-notes-website-links-now-point-to-new-search-results-april-20-2021/

### Steps to reproduce the problem

- Open cache details of an unsaved  cache

### Actual result after these steps 

- Cache type is shown as "??" 

### Expected result after these steps 

Cache type parsing working

## c:geo version 

2021.04.15

## Reproducible 

Yes

## System information 

not relevant

## Additional context 

Other problems might apply, e.g. owner has some problem
Taken from a users log:
```
4-20 20:00:10.998  6897  9615 W cgeo    : [Thread-17] owner not parsed correctly for GC97Z19
04-20 20:00:10.999  6897  9615 W cgeo    : [Thread-17] type not parsed correctly for GC97Z19
04-20 20:00:16.265  6897  4136 W DpmTcmClient: read failed bytes countRead: -1
04-20 20:00:17.523  6897  9615 W cgeo    : [Thread-17] owner not parsed correctly for GC97WH4
04-20 20:00:17.523  6897  9615 W cgeo    : [Thread-17] type not parsed correctly for GC97WH4
04-20 20:00:22.620  6897  9615 W cgeo    : [Thread-17] owner not parsed correctly for GC96ZMP
04-20 20:00:22.620  6897  9615 W cgeo    : [Thread-17] type not parsed correctly for GC96ZMP
```

Lineflyer:
A fix for this need to target `release`

Lineflyer:
Example scenario:
- Search for GC97Z19
- Cache type is shown as "??" instead of "Mystery"

The owner however looks normal to me.

Lineflyer:
Raised a user notification

moving-bits:
I'm working on it...

moving-bits:
Fixing the cache type seems to be pretty straightforward, but for the owner the cache details page only contains the guid now, and no longer the user id. I'll push a PR for the cache type, but let's keep this open as we need to find a solution for the user id as well.

moving-bits:
[...]

PR #10423, opened 2021-04-20
adapt cache type detection to website change (rel. to #10421)

moving-bits:
## Description
Adapts our code to the website change of cache detail page regarding the cache type. It's now referenced using the waypoint type instead of the type guid.

## Related issues
related to #10421, but not completely fixing it, as there is also a change regarding the cache owner.


Lineflyer:
Failing test seem to be caused by the remaining owner problem?!

moving-bits:
> Failing test seem to be caused by the remaining owner problem?!

The first two, yes. But the third is failing as cache pages without login do not provide the "nearby caches of this type" link which I used for type detection. I've updated the PR to use the cache symbol svg identifier which seems to use the waypoint id as well, and is also present when you're not logged in. Let's see if this works for the tests...

moving-bits:
There was another problem with the mocked data, as somehow the date format got interpreted differently, so that mocked data and actual data could not be made comparable any longer. I've adapted the date strings in the mocked data, and so the third test is also working again.

Now only the owner-related tests still fail (as to be expected).

Lineflyer:
Merging to test the fix.
Thanks @moving-bits

PR #10430, opened 2021-04-21
try to fix the parsing of owner id after website change

### LINKED ISSUE #10421: Cache type and owner parsing not OK (Website change)

fm-sys:
fixes #10421

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -45,7 +45,7 @@ public final class GCConstants {
     static final Pattern PATTERN_NAME = Pattern.compile("<span id=\"ctl00_ContentBody_CacheName\">(.*?)</span>");
     static final Pattern PATTERN_DIFFICULTY = Pattern.compile("<span id=\"ctl00_ContentBody_uxLegendScale\"[^>]*>[^<]*<img src=\"[^\"]*/images/stars/stars([0-9_]+)\\.gif\"");
     static final Pattern PATTERN_TERRAIN = Pattern.compile("<span id=\"ctl00_ContentBody_Localize[\\d]+\"[^>]*>[^<]*<img src=\"[^\"]*/images/stars/stars([0-9_]+)\\.gif\"");
-    static final Pattern PATTERN_OWNER_USERID = Pattern.compile("seek/nearest\\.aspx\\?u=(.*?)\">.{1,50}?seek/nearest\\.aspx\\?ul=\\1\">");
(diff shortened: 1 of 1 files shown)


Lineflyer:
@fm-sys The failing tests look like there is now a difference in the case the shown cache owner name differs from the user name, who owns the cache.
As you might know, the owner can define the shown user name while creating a cache so it might differ from the real user name. Could this be a problem for your PR?
You can use GC7C6K8 for testing. Owner is Lineflyer, shown owner is "Lineflyer & Velberlinle"

Test result on CI shows:
```
org.junit.ComparisonFailure: [Cache GC2JVEH: owner user id] expected:<"indianerjones[]"> but was:<"indianerjones[, der merlyn,reflektordetektor]">
```

Lineflyer:
Or the mocked data needs update but AFAIR @moving-bits already updated the mocked data.

moving-bits:
> but AFAIR @moving-bits already updated the mocked data.

I did so only regarding the date format, and manually so, no complete refresh.

fm-sys:
In fact the reason for the failing test is, that the mocked test data is outdated and therefore the new regex will fail and the user display name is probably used as fallback in this case? I will need to update the test data, the regex itself seems to behave just fine...

BTW, I can only try modify the test data manually, as I (currently) don't have a GC premium membership. The alternative would be to merge this PR as is and someone else needs to update the test data afterwards...

moving-bits:
I can look into it today or tomorrow, depending on available time.

PR #10433, opened 2021-04-21
rel to #10406: make map theme folder scan asynchronous

eddiemuc:
## Description
This PR, if merged, will make map theme folder scan asynchronous. This is to reduce ANRs as described in #10406 which are assumed to result from synchronus map theme folder scan on startup.
* An asynchronous folder scan will be started on application startup
* If such a scan has not finished when map viewer is first opened, then the same scan will be done synchronous (resulting in slow map viewer opening performance)
* As an additional measurement, scan will generally be restricted to a tree depth of at most 2 subdirs
* (The previously introduced scan time limit was removed instead)

This is a rather complex fix due to asynchronity of tasks. I would appreciate code review.

## Related issues
Related to #10406

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/RenderThemeHelper.java
@@ -82,7 +82,7 @@ public class RenderThemeHelper implements XmlRenderThemeMenuCallback, SharedPref
 
     private static final long FILESYNC_MAX_FILESIZE = 5 * 1024 * 1024; //5MB
 
-    private static final int THEMESCAN_MAX_SECONDS = 5;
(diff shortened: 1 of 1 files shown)


eddiemuc:
As with #10447:  I assume we don't have immediate plans for yet another release (except other website changes occur :-)), so I will wait 2-3 days to give people chance for a review- Unless we have a new "emergency web site fix" situation then, I will merge.

moving-bits:
I'm not that deep into those asynchronous things, therefore not a good candidate for a source review in this case. But from a user's perspective the results LGTM, no ANR and no strange results in the theme list.

eddiemuc:
@moving-bits thanks a lot for your review. I will merge later today

eddiemuc:
Ah on the other hand: why wait? :-)

PR #10436, opened 2021-04-21
adapt gc login to website change (fix #10434)

### LINKED ISSUE #10434: Login problems
Not sure whats going on, but it seems some new redirects are involved (which I never seen before).
Might also be a temporary server problem...as the website is slow in general but login with browser however works.

```
2021-04-21 21:02:56.762 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] POST https://www.geocaching.com/account/logout
2021-04-21 21:02:57.577 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] 403 [Forbidden] (815 ms) POST https://www.geocaching.com/account/logout (http/1.1)
2021-04-21 21:02:57.584 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/account/signin
2021-04-21 21:02:57.780 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] 200 (196 ms) GET https://www.geocaching.com/account/signin (http/1.1)
2021-04-21 21:02:57.911 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] POST https://www.geocaching.com/account/signin
2021-04-21 21:03:05.136 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] 200 (7225 ms) POST https://www.geocaching.com/account/signin (http/1.1) (=> https://www.geocaching.com/play/search)
2021-04-21 21:03:06.110 6468-6844/cgeo.geocaching I/cgeo: [network--5] Failed to log in Geocaching.com as Zubu for some unknown reason
2021-04-21 21:03:06.136 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/play/culture/set?model.SelectedCultureCode=en-US
2021-04-21 21:03:19.883 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] 200 (13747 ms) GET https://www.geocaching.com/play/culture/set?model.SelectedCultureCode=en-US (http/1.1) (=> https://www.geocaching.com/play/search)
2021-04-21 21:03:20.910 6468-6844/cgeo.geocaching I/cgeo: [network--5] changed language on geocaching.com to English
2021-04-21 21:03:20.952 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/account/signin
2021-04-21 21:03:32.383 6468-6845/cgeo.geocaching D/cgeo: [OkHttp] 200 (11431 ms) GET https://www.geocaching.com/account/signin (http/1.1) (=> https://www.geocaching.com/play/search)
2021-04-21 21:03:33.617 6468-6844/cgeo.geocaching W/cgeo: [network--5] GCLogin.login: failed to find request verification token
```

moving-bits:
## Description
adapt login for the GC connector to recent website change regarding login

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -80,7 +80,7 @@ public final class GCConstants {
     static final Pattern PATTERN_CACHES_FOUND = Pattern.compile(Pattern.quote("cache-count\">") + "\\s*([\\d,.]+)", Pattern.DOTALL);
 
     static final Pattern PATTERN_AVATAR_IMAGE_SERVERPARAMETERS = Pattern.compile("\"(https?:\\/\\/(img(?:cdn)?\\.geocaching\\.com|[^>\\\"]+\\.cloudfront\\.net)\\/avatar\\/[0-9a-f-]+\\.(" + IMAGE_FORMATS + "))\\\"");
-    static final Pattern PATTERN_LOGIN_NAME_LOGIN_PAGE = Pattern.compile("ctl00_ContentBody_lbUsername\">.*<strong>(.*)</strong>");
+    static final Pattern PATTERN_LOGIN_NAME_LOGIN_PAGE = Pattern.compile("<script type=\"text\\/javascript\">\\s*var dataLayer = \\[\\{\\s*\"username\": \"([^\"]*)\"");
 
     // Patterns for parsing trackables
 
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -78,6 +78,7 @@ public final class GCConstants {
     public static final Pattern PATTERN_LOGIN_NAME = Pattern.compile("class=\"user-name\"[^>]*>(.*?)</span>", Pattern.DOTALL);
     /** Use replaceAll("[,.]","") on the resulting string before converting to an int */
     static final Pattern PATTERN_CACHES_FOUND = Pattern.compile(Pattern.quote("cache-count\">") + "\\s*([\\d,.]+)", Pattern.DOTALL);
+    static final Pattern PATTERN_CACHES_FOUND_LOGIN_PAGE = Pattern.compile("\"findCount\": ([0-9]*)\\s*\\}];\\s*<\\/script>");
 
     static final Pattern PATTERN_AVATAR_IMAGE_SERVERPARAMETERS = Pattern.compile("\"(https?:\\/\\/(img(?:cdn)?\\.geocaching\\.com|[^>\\\"]+\\.cloudfront\\.net)\\/avatar\\/[0-9a-f-]+\\.(" + IMAGE_FORMATS + "))\\\"");
     static final Pattern PATTERN_LOGIN_NAME_LOGIN_PAGE = Pattern.compile("<script type=\"text\\/javascript\">\\s*var dataLayer = \\[\\{\\s*\"username\": \"([^\"]*)\"");
--- a/main/src/cgeo/geocaching/connector/gc/GCLogin.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCLogin.java
@@ -308,7 +308,13 @@ public class GCLogin extends AbstractLogin {
         setActualLoginStatus(TextUtils.matches(page, GCConstants.PATTERN_LOGIN_NAME_LOGIN_PAGE));
         if (isActualLoginStatus()) {
             setActualUserName(Settings.getUserName());
-            // number of caches found is not part of this page
+            int cachesCount = 0;
+            try {
+                cachesCount = Integer.parseInt(removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_LOGIN_PAGE, true, "0")));
+                setActualCachesFound(cachesCount);
+            } catch (final NumberFormatException e) {
+                Log.e("getLoginStatus: bad cache count", e);
+            }
             return true;
         }
 


main/src/cgeo/geocaching/connector/gc/GCConstants.java
```diff
@@ -80,7 +80,7 @@
     static final Pattern PATTERN_CACHES_FOUND = Pattern.compile(Pattern.quote("cache-count\">") + "\\s*([\\d,.]+)", Pattern.DOTALL);
 
     static final Pattern PATTERN_AVATAR_IMAGE_SERVERPARAMETERS = Pattern.compile("\"(https?:\\/\\/(img(?:cdn)?\\.geocaching\\.com|[^>\\\"]+\\.cloudfront\\.net)\\/avatar\\/[0-9a-f-]+\\.(" + IMAGE_FORMATS + "))\\\"");
-    static final Pattern PATTERN_LOGIN_NAME_LOGIN_PAGE = Pattern.compile("ctl00_ContentBody_lbUsername\">.*<strong>(.*)</strong>");
+    static final Pattern PATTERN_LOGIN_NAME_LOGIN_PAGE = Pattern.compile("<script type=\"text\\/javascript\">\\s*var dataLayer = \\[\\{\\s*\"username\": \"([^\"]*)\"");
```
UniQP, line 83: I wonder if the escaping of special characters (especially `"`) also changed and requires some more changes on our end.

moving-bits:
> The failing test `cgeo.geocaching.connector.gc.GCConstantsTest.testCacheCountOnline` looks suspicious...

I guess they changed the source for find counter as well, and that one we are checking only on subpages, not on the login page. I'm working on retrieving that from login page (see "dataLayer") as well...

main/src/cgeo/geocaching/connector/gc/GCConstants.java
moving-bits: Might very well be. Currently I'm taking all characters except quotation mark, regardless of any specific encoding. Need to investigate more deeply for special characters which would have to be encoded in Javascript. Do you have a regexp or utility function for decoding encoded chars in JS strings?

moving-bits: Additionally, the two regexpes are a bit fragile as they are position dependent within the script block - needs to be a bit more generalized. But that's a different task for another day...

bekuno:
Is it not possible to use https://www.geocaching.com/play/serverparameters/params,
which is parsed in `GCLogin.getServerParameters()`?

moving-bits:
> The failing test `cgeo.geocaching.connector.gc.GCConstantsTest.testCacheCountOnline` looks suspicious...

Fixed meanwhile. Now it's only the two owner-related tests failing.

main/src/cgeo/geocaching/connector/gc/GCConstants.java
UniQP: We could match the whole JSON string, then construct a `JSONObject` out of it and read the fields we need. Shouldn't be that complicated. For the JSON part, you can take a look at `main/src/cgeo/geocaching/models/CalcState.java`.

Lineflyer:
@moving-bits So we could merge this and the fix for owner related parsing from @fm-sys now or is there still work ongoing for one of those?

moving-bits:
> @moving-bits So we could merge this and the fix for owner related parsing from @fm-sys now or is there still work ongoing for one of those?

Regarding the login website change: Yes, the workaround is ready to merge. 
But it's a quick workaround only, for a more permanent solution we need to do some polishing, which we can do over the next days. (Possible drawback of the current workaround may be a somewhat "distorted" username (display issue only?) in the case the username consists of some special chars which have to be escaped in the Javascript source.)

As for the owner related parsing I cannot say, didn't have the time yet for a look at it.

fm-sys:
> As for the owner related parsing I cannot say, didn't have the time yet for a look at it.

This is ready as well, just the mocked test data needs to be updated though. Bus as this only affects our CI not the users, I think you can merge both if you want to do a small bugfix release 👍

baiti:
I was wondering .... is this a subtle attempt to "help" us moving to the API?

fm-sys:
> I was wondering .... is this a subtle attempt to "help" us moving to the API?

I had the same thought. But I think it's unrelated. Website changes have always existed and are the good right of Groundspeak. Finally, we are the ones who process their data so such situations are our fault. But you're right that the API would avoid such situations (although with many other downsides)

baiti:
> > I was wondering .... is this a subtle attempt to "help" us moving to the API?
> 
> I had the same thought. But I think it's unrelated. Website changes have always existed and are the good right of Groundspeak. Finally, we are the ones who process their data so such situations are our fault. But you're right that the API would avoid such situations (although with many other downsides)

Yes, they can change the UI at any time, however, they could and should have given us advanced notice such that we could test and integrate the changes without generating a disruption like the one we now had for two days. At least I would do this if I wanted to partner with some other entity.

Lineflyer:
> Yes, they can change the UI at any time, however, they could and should have given us advanced notice such that we could test and integrate the changes without generating a disruption like the one we now had for two days. At least I would do this if I wanted to partner with some other entity.

They can't judge what change might break our code. However we had some alignments in beforehand when it came to removal of features, etc. So they aligned in the past if they could easily predict it will make a problem. 
IMHO this is already a nice attitude, we cannot expect them to plan with c:geo while working on their website.

MagpieFourtyTwo:
> They can't judge what change might break our code

May or may not be. But although it might be true WRT c:geo, it most definitely does *not* apply to others, who do not just rely on header layout, but do even rewrite e. g. the header menus, like [GClh](https://github.com/2Abendsegler/GClh). And the guys over there were caught just as cold by the changes.

Of course GS can change their website as they like, as it is simply their property. Nevertheless, they still are just nothing but a service provider. And they are paid by their customers. Thus have to serve them and their needs. And still, a *lot* of these customers are just paying *them* because softwares like c:geo and GClh exist - and make a) caching more playable and b) gc.com more usable (or usable at all). At least me would not be a Cacher anymore, if esp. these programs would not exist.

Therefore, GS should be *urgently* interested in these programs doing well. Because *they* flush money into *their* pockets.

But do they really do ...?

BTW: Playing around in the productive system and changing things back and forth with no (recognizable) concept at all, is not really suitable to make life easier for programs relying on a structure, neither ...

My 2 cent (based on my experiences with gc.com): They did not tell in advance, cause they couldn't.
Cause they themselves did not even know what the results will probably be ...

main/src/cgeo/geocaching/connector/gc/GCConstants.java
moving-bits: @UniQP Can JSoup parse Javascript vars in the source, some code like this?
```
<script type="text/javascript">
//<![CDATA[
var a = true;
window['someName'] = {
  "var1": true,
  "var2": "some text",
  "var3": "",
  ...
}; window['someOtherName'] = 'someValue';//]]>
</script>
```
I would like to receive the contents of the "someName" var as key/value pairs.

UniQP: @moving-bits As JSoup is an HTML parser, I don't think it's possible (because it involves parsing JavaScript).

Issue #10440, opened 2021-04-22
Relogin triggered on many ocassions


Lineflyer:
While testing the current state of `release` for a potential short term release, I stumbled upon a general slowness, e.g. when opening a cache, doing a nearby search, adding a cache to watchlist, and many others.

Looking at the debug log I can clearly see, that all of the above mentioned actions trigger a rerun of the login process in the background. 
Example scenario:
- Turn debug logging on
- Start c:geo, wait until logged in
- Perform nearby search

Result:
- The debug log shows the GET command for the search and the corresponding reply
- Triggered by this reply (?) the login process is started by c:geo
- The nearby search is done again 
(Luckily there is no endless loop created)

It seems our "still logged in check" is broken and assumes not being logged in and therefore applying a login.

Log of above mentioned example:
```
2021-04-22 19:57:19.561 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] 200 (519 ms) GET https://www.geocaching.com/seek/nearest.aspx?lat=51.00205&lng=7.00535&tx=9a79e6ce-3344-409c-bbe9-496530baf758 (http/1.1)
2021-04-22 19:57:19.592 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/account/signin
2021-04-22 19:57:20.092 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] 200 (500 ms) GET https://www.geocaching.com/account/signin (http/1.1) (=> https://www.geocaching.com/play/search)
2021-04-22 19:57:20.132 20094-20158/cgeo.geocaching I/cgeo: [network--4] Already logged in Geocaching.com as Zubu (BASIC)
2021-04-22 19:57:20.148 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/play/culture/set?model.SelectedCultureCode=en-US
2021-04-22 19:57:21.047 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] 200 (899 ms) GET https://www.geocaching.com/play/culture/set?model.SelectedCultureCode=en-US (http/1.1) (=> https://www.geocaching.com/play/search)
2021-04-22 19:57:21.071 20094-20158/cgeo.geocaching I/cgeo: [network--4] changed language on geocaching.com to English
2021-04-22 19:57:21.099 20094-20491/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/account/signin
[...]

PR #10443, opened 2021-04-22
adapt logged-in check for non-login pages (rel. to #10440)

moving-bits:
Since recent website changes both login page and other cache pages have the same way the username is stored in the page's source, so detection can be merged into one, saving some extra roundtrips/checks.

Also language detection is adapted to that the page does not include a language selector any longer, but checks server parameters instead.


Lineflyer:
Ready for merge?

moving-bits:
> Ready for merge?

I've now removed the superfluous import, which should fix the one CheckStyle issue, so I guess you can merge.

Lineflyer:
OK, will merge and test whether #10440 is fixed.

PR #10450, opened 2021-04-23
adapt login regexp to yet another website change (rel. to #10440)

moving-bits:
Adapt to yet another cascade of changes on their login pages, changing the check from `dataLayer` to `headerSettings`. Current expression will now also work with either `window.varname` and `window['varname']` way of declaring, and with `username` being the last var or not.


Lineflyer:
Any overview whether the failing tests are real fails or false positives due to outdated mocked data?

moving-bits:
From a quick glance: Not sure about `cgeo.geocaching.CgeoApplicationTest.testSearchByGeocodeSpecialties`, but the other three use our outdated mocked files.

Lineflyer:
I will check later tonight and therefore merge now.
Thanks @moving-bits  Highly appreciated.

Will provide feedback asap.

Issue #10444, opened 2021-04-22
Reminder: Cleanup needed after hotfixes for recent website changes


moving-bits:
Reminder to myself:

We have implemented some hotfixes regarding the recent website change to get c:geo back running, but those partly consist of more or less fragile workarounds, which should be cleaned up:

For details see:
- #10436 detection of username and find counter: position within JS var and special chars in username
- #10439 fix remaining tests for mocked up files

(maybe more)

moving-bits:
@Lineflyer 

For testing the "special characters in username encoding": Can you ask a couple of users which have special characters in their username (umlauts, accents, Chinese characters, etc.?) if they can open a cache page, save its content and send it to us? 

(The one line in the `headerSettings` section containing the `"username": "xxx"` is probably sufficient, maybe with the full HTML header for correct encoding, if it differs from "utf-8".) You could forward it via PM to me. 

And maybe ask them what they get displayed as username in c:geo?

Thanks!

Lineflyer:
Up to now no one contacted us with any problems or wrongly shown user names. Will do however if it occurs.

I tried to rename my basic user (test user) but failed.
According to the website all names with special chars I tried are "unavailable"....whatever that means.

moving-bits:
Thanks for testing. AFAIK you cannot create user accounts with special characters any longer, even "-" or a space are no longer possible. But it has not always been that way, so for testing special characters it has to be some existing user with an older account.

But let's see if something surfaces in support, and keep this issue open for a month or so as a reminder for us.

moving-bits:
LGTM, thanks for cross-checking.

BTW: This reminds me of some challenges ("for each of the following characters find a cache which owner's name starts with that character"...)

moving-bits:
There seem to be no issues with special characters, otherwise we would have heard of it in the meantime. So let's close here for now (and reopen if needed).

PR #10455, opened 2021-04-24
update regular expression for caches count (fix #10453)

### LINKED ISSUE #10453: Find count shown as 0
Some users of 2021.04.24 report, that after "a while of usage" the find count, which was previously correct, drops to 0.
No more details yet, but three reports.

murggel:
## Description

Edit:
There are 2 different script-sections for the settings (see details below).
~~I added a second regExpr like the one for the username of the headerSettings-section.
Probably they can be merged to one to find *"findCount": 590,* or  *"findCount": 590*, but then less information is parsed.
With the extended pattern including the "headerSettings" it is probably more restrictive.~~
The login page has also the section `headerSettings` with the findCount contained (see regExpr for username). So now this section is used for the findCount as well.

@moving-bits  please check
~~or probably on detail-page the count should not be updated?~~

## Related issues

#10453

### Login-Page
```
<script type="text/javascript">
    var dataLayer = [{
  "username": "xxx",
  "accountId": 1234,
  "membershipLevel": "Premium",
  "signUpDate": "2000-01-01T00:00:00",
  "membershipExpirationDate": "2021-12-31T00:00:00",
  "findCount": 590
}];
</script>
```

### CacheDetail-Page
```
<script type="text/javascript">
//<![CDATA[
var isLoggedIn = true;
var lat=48.1111, lng=9.00000, guid='xxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx';
window['headerSettings'] = {
  "isAuthenticated": true,
  "paymentUrl": "https://payments.geocaching.com",
  "planUrl": "",
  "gameplayUrl": "",
  "mainElementId": "Content",
  "mapUrl": "",
  "locale": "en-US",
  "inlinePostfix": "--inline",
  "publicUrl": "https://www.geocaching.com",
  "hostname": "https://www.geocaching.com",
  "username": "xxx",
  "findCount": 590,
  "avatarUrl": "https://www.geocaching.com/images/default_avatar.png",
  "isBasic": false,
  "showRenew": false,
  "showDoNotSell": true,
  "userPublicGuid": "xxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
}; window['userRef'] = 'PRXXXX';//]]>
</script>
```

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCParser.java
@@ -1955,10 +1955,7 @@ public final class GCParser {
     public static int getCachesCount(final String page) {
         int cachesCount = -1;
         try {
-            String intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_LOGIN_PAGE, true, ""));
-            if (intStringToParse.isEmpty()) {
-                intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_HEADER, true, ""));
-            }
+            final String intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_LOGIN_PAGE, true, ""));
             if (!intStringToParse.isEmpty()) {
                 cachesCount = Integer.parseInt(intStringToParse);
             }
@@ -1967,7 +1964,6 @@ public final class GCParser {
         }
 
         return cachesCount;
-
     }
 
 
--- a/main/src/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCParser.java
@@ -1952,4 +1952,27 @@ public final class GCParser {
         return StringUtils.isNotEmpty(usernameOld) ? usernameOld : null;
     }
 
+    @Nullable
+    public static int getCachesCount(final String page) {
+        int cachesCount = -1;
+        try {
+            String intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_LOGIN_PAGE, true, ""));
+            if (intStringToParse.isEmpty()) {
+                intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_HEADER, true, ""));
+            }
+            if (!intStringToParse.isEmpty()) {
+                cachesCount = Integer.parseInt(intStringToParse);
+            }
+        } catch (final NumberFormatException e) {
+            Log.e("getLoginStatus: bad cache count", e);
+        }
+
+        return cachesCount;
+
+    }
+
+
+    private static String removeDotAndComma(final String str) {
+        return StringUtils.replaceChars(str, ".,", null);
+    }
 }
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1952,4 +1952,27 @@ public static String getUsername(final String page) {
         return StringUtils.isNotEmpty(usernameOld) ? usernameOld : null;
     }
 
+    @Nullable
```
bekuno, line 1955: @Nullable/@NonNull is for objects

really `cachesCount=-1` if it could not determined?
@Lineflyer What do you mean from support point of view?

main/src/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1952,4 +1952,27 @@ public static String getUsername(final String page) {
...
+    public static int getCachesCount(final String page) {
+        int cachesCount = -1;
```
bekuno, line 1957: really cachesCount=-1 if it could not determined?
@Lineflyer What do you mean from support point of view?

main/src/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1952,4 +1952,27 @@ public static String getUsername(final String page) {
...
+        try {
+            String intStringToParse = removeDotAndComma(TextUtils.getMatch(page, GCConstants.PATTERN_CACHES_FOUND_LOGIN_PAGE, true, ""));
```
bekuno, line 1959: Is there a chance to have a parameter to this method with the type of page (login or other) to directly call the right parser?

main/src/cgeo/geocaching/connector/gc/GCParser.java
murggel: Copy/paste mistake

main/src/cgeo/geocaching/connector/gc/GCParser.java
murggel: Somewhere it was mentioned, that -1 is not determined, where as 0 means no finds

main/src/cgeo/geocaching/connector/gc/GCParser.java
murggel: I will check that

murggel: seems that the login-page also has the headerSettings - I will have a closer look at this

murggel: Thanks for the hint. Sometimes you stick to one approach and don't see another one.

moving-bits: for the username I switched to `headerSettings` completely, as meanwhile (or maybe only on subpages) the `dataLayer` got extracted into calling an external script

murggel: ok, then switching copmpletly to headerSettings for findCount would be safe as well

main/src/cgeo/geocaching/connector/gc/GCParser.java
murggel: In MainActivity.fillView the FoundNumCounter is only updated, if getCachesFound>= 0, so no -1 shown to the user. FoundNumCounter stays then unchanged

fm-sys:
@murggel are you still working on this? Otherwise I think we could/should merge it rather soon. Thanks for fixing this!

murggel:
> @murggel are you still working on this? Otherwise I think we could/should merge it rather soon. Thanks for fixing this!
@fm-sys  I am done, if no more comments are on the regExp.
Just edited the PR-description.

moving-bits:
Let's merge it, need it as input for #10444

moving-bits:
Thanks @murggel for tackling this issue!

Issue #9899, opened 2021-01-29
Set marker for waypoint with calculated coordinates in the waypoint list


murggel:
I also thinking of some indication in the waypoint-list, to show the formula there 
for formula in plain format this can be done easily:
![grafik](https://user-images.githubusercontent.com/70113341/104110117-ba797b00-52d4-11eb-983f-7abe78f33663.png)

If the formula is the first part of the waypoint-note, the repetition of the formula can look strange.

If it is too long, the "calculated coordinates" can be omitted.
For the other format probably only the text "calculated coordinates" could be shown

_Originally posted by @murggel in https://github.com/cgeo/cgeo/issues/6833#issuecomment-757378884_

There was also the discussion to use a add-on-icon (SVG) for indication.

murggel:
@moving-bits pleas assign the issue to me...

I will play around a little bit

murggel:
Taken from #6833

> 
> 
> > As it is #7856 only seems to discuss issues with API < 21, which is a moot point now.
> 
> We increased the min API level to 21 along the way, so discussion changed. But he recommended to keep the compat lib due to some quirks in Android 5/6, see [#7856 (comment)](https://github.com/cgeo/cgeo/pull/7856#issuecomment-619408476).
> 
> > But regardless I'd recommend trying to get SVG icons working first and only revert to PNG (or whatever) if the SVG fails.
> 
> Highly appreciated. Might be a starting point for more migration towards SVG.

Shall we ask NotWood to continue with his work?

Or ... I already took a look at the PR (to be honest only to be able to play around with the calculater-svg) and found out that there is no more the need for the vector drawable compat library... only one additional definition in the gradle.build is necessary. The svg can then be imported via AndroidStudio and converted to vector-drawable-xml or even in libraries there are already vector-drawables available...
It seems, that there is a possibility to create the vector-drawable out of the svgs in the build-step, but there I am not the right person for... (this would only be necessary if we want to have the svg automatically converted to vector-drawable)
The provided XML from the PR can be still used

Now I am playing around with different calculator-svg to find .
I will create a separate issue for the image to show that a waypoint has calculated coordinates, I guess there will be more discussion, which does not fit to the initial topic of this issue.

murggel:
SVG for this case is working :-) 

I tried some motives and colors, but no one satisfied me completely so I need your feedback:
[...]

PR #10456, opened 2021-04-24
Indicate waypoint has calculated coordinates in waypoint list (rel #9899, #6833)

murggel:
## Description

Show text "Calculate coordinates" in the waypoint list for the coordinates of a waypoint, if it has calculated coordinates.
CalcState is only saved, if it has equations / variables. 
grafik

In the CoordinatesInputDialog the button "Calculate" has italic font, if already a formula is given (I can change this to bold / colored, ... ).
grafik

## Related issues

#6833 , #9899 

## Additional context

Because there is no way to delete the calcState of the waypoint, I added the logic, that a calcState is only set, if it has equations.
Probably it should be extended to "delete the calcState, if the coordinates get cleared by user". But I an not sure about this, so this could be an extra PR.

I am working on showing a icon in addition (like the "tick" for "visited" waypoints), but I am not yet satisfied with the look and feel of that icon, I am not sure, if it is visible enough on small screens (see https://github.com/cgeo/cgeo/issues/9899#issuecomment-782926595)

### DIFF
--- a/main/res/layout/waypoint_item.xml
+++ b/main/res/layout/waypoint_item.xml
@@ -84,6 +84,23 @@
         android:maxLines="1" />
 
     <TextView
+        android:id="@+id/calculate_coordinates"
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="left"
+        android:layout_marginLeft="12dp"
+        android:ellipsize="end"
+        android:focusable="false"
+        android:lines="1"
+        android:maxLines="1"
+        android:scrollHorizontally="true"
+        android:textColor="?text_color_headline"
+        android:textSize="14sp"
+        android:visibility="gone"
+        android:text="@string/waypoint_calculate_coordinates"
+        tools:visiblity="visible" />
+
+    <TextView
         android:id="@+id/note"
         android:autoLink="web|map"
         android:layout_width="fill_parent"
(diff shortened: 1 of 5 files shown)


Narrat:
I may miss something, but if the waypoint has calculated coordinates shouldn't the displayed text also be `Calculated coordinates`? `Calculate coordinates` sounds more like something that should be done. Or does tapping on the text bring up the calculator?

murggel:
> I may miss something, but if the waypoint has calculated coordinates shouldn't the displayed text also be `Calculated coordinates`? `Calculate coordinates` sounds more like something that should be done.

My thoughts as well, for the first implementation I just used the same text as on the button in the dialog, but sure, it can be changed...

> Or does tapping on the text bring up the calculator?

Good idea, will check that, if I can implement that...

moving-bits:
Thanks @murggel for diving into this issue. It's good to have a visual indicator for having a formula-supported waypoint.

> Show text "Calculate coordinates" in the waypoint list for the coordinates of a waypoint, if it has calculated coordinates.

This does not work for me if the coordinates are calculated and all variables are defined (so the calculated coordinates are fully determined). IMHO it should display the hint also in this situation.

> In the CoordinatesInputDialog the button "Calculate" has italic font, if already a formula is given (I can change this to bold / colored, ... ).

I like it the way it is - it gives you a good visual hint without being too obstrusive, so I would stay with italic font. You could even use "Calculated coordinates" here as well, would still fit and be consistent with the hint in the waypoint list.

moving-bits (approved):
Thanks @murggel for tackling this issue, and thanks for the last adaptions. LGTM

PR #10458, opened 2021-04-24
Optimize strings for our notification system

Lineflyer:
## Description

More general problem description for our notification system, to be able to use it more often.
Also making it clearer, that we are aware of the problem.

Additionally adding to the report problem prefix, that users should check FAQ and social media first.

## Related issues

None


Lineflyer:
Note:
This time I opened the PR towards release. I will NOT merge any translation PRs for the next days, so the new english texts are in but the currently localized texts are not deleted.
Only after next probable short term release and when more translations come in, I will merge translations again.

Issue #10454, opened 2021-04-24
Owner not parsed


Lineflyer:
Followup or reappearance of #10421 with 201.04.24:

- Open (an unknown/unsaved) caches, e.g. GC352Y3 (tested as Basic member)
- Click on owner name
- Select caches hidden or caches found

Result:
"No username given"

Log:

```
2021-04-24 12:05:14.349 28626-28659/cgeo.geocaching D/cgeo: [OkHttp] GET https://www.geocaching.com/seek/cache_details.aspx?decrypt=y&wp=GC352Y3&log=y&numlogs=0
2021-04-24 12:05:15.002 28626-28659/cgeo.geocaching D/cgeo: [OkHttp] 200 (652 ms) GET https://www.geocaching.com/seek/cache_details.aspx?decrypt=y&wp=GC352Y3&log=y&numlogs=0 (http/1.1) (=> https://www.geocaching.com/geocache/GC352Y3_die-alte-schule?decrypt=y&log=y&numlogs=0)
2021-04-24 12:05:15.115 28626-28751/cgeo.geocaching W/cgeo: [refresh--1] owner not parsed correctly for GC352Y3
[...]
```

MagpieFourtyTwo:
> - Select caches hidden or caches found

... is not even necessary - it's sufficient to check the window title, which just says ``About`` i. e. even the title is already missing the Owner's name.

Most regretfully I was not able to nail a scenario yet. Sometimes it works, sometimes it doesn't, even with the same Owner (name) and with caches, which have been handled in the absolute same way: This morning a bunch of caches have been published, which I opened one after the other via the notification mail link. And while the first one, https://coord.info/GC96WYF, had the owner problem, in the next one, https://coord.info/GC96WWV, the click on ``Owner`` => ``Caches hidden`` worked perfectly fine.

Afterwards I performed some more actions with the first one, but whatever I did, the Owner popup title remained empty and ``Caches hidden`` constantly failed. And it still fails, whatever I do (saving, deleting, opening from lists, open directly via GC code or via map ...), while the other still works like a charm.

I have another one, where the ``Owner`` => ``Caches hidden`` fails, just look at https://coord.info/GC99HF8 - but over there it may be just because of the ampersand ...?

MagpieFourtyTwo:
BTW: [GC352Y3](https://coord.info/GC352Y3) gives me an empty Owner, too.

Lineflyer:
Did you test with newest nightly?

[...]

PR #10459, opened 2021-04-24
adapt owner pattern to website change (rel. to #10454)

moving-bits:
Tries to fix the remaining problem with owner identification (#10454).

Our current pattern looks for a string terminated with `&a=0">`, whereas in the current page source there may be more parameters after the `&a=0`, so I removed the closing tag. 
All caches opened from me with the changed pattern allowed me to do an owner search (and I tried caches from areas abroad where I haven't been there yet, so I expect to not have any owner info stored for their owners yet), but I cannot prove that it will work. If someone has a reproducable scenario: please cross-check.


Lineflyer:
Will crosscheck after merge.

moving-bits:
> Will crosscheck after merge.

Thanks.

Ok, let's merge this fix + the updated mocked data to get closer to zero failing tests in the CI for the upcoming PR...

PR #10460, opened 2021-04-24
Temporarily insert german strings for notification system

Lineflyer:
## Description

I will only merge the Crowdin PR after some time for translators to catch up. Otherwise existing translations would be deleted because they are not untranslated after pushing the strings change for the source strings.

Therefore manually adding the german translations (to cover a huge user group) for our next short term release.

## Related issues

None


Lineflyer:
@moving-bits Hurray! You fixed the mocked caches...Now we are finally green again.

PR #10476, opened 2021-04-25
support setting found state for LC connector (fix #10475)

### LINKED ISSUE #10475: Allow lab caches to be set to found state

A lab cache should be allowed to be set to status "Found" manually. This will allow the filtering rules to apply to e.g. filter out saved lab caches marked as "found".

## Suggested solution

We could reuse what was implemented for UDC.

## Alternatives/workarounds

Only a personalized API could do that, but as suggested above we at least would have a local solution for saved lab caches.

moving-bits:
## Description
Support setting state to "found", "did not find" or removing those states for the LC connector (in the same way we have it for UDC)

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCConnector.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCConnector.java
@@ -78,6 +78,11 @@ public class LCConnector extends AbstractConnector implements ISearchByGeocode, 
     }
 
     @Override
+    public boolean supportsSettingFoundState() {
+        return true;
+    }
+
+    @Override
     public SearchResult searchByGeocode(@Nullable final String geocode, @Nullable final String guid, final DisposableHandler handler) {
         if (geocode == null) {
             return null;


baiti:
@moving-bits ... when we set the found state, wouldn't it make sense to record the date too? I would assume that the found date would then show up in the CacheDetailActivity?

moving-bits:
If I'm not mistaken (cannot look in the code currently) the found state is saved as a flag in the cache table, whereas the found date is retrieved indirectly by scanning the logs table for the respective log entry. So if we want to store a log date, we would need to either add a "logdate" column to the cache table (and fill it correctly in all circumstances), or add an artificial found log for the ALC in the log table.

fm-sys:
Recording the found date would require additional UI (date selection, etc.) and would make things more complicated without any real value IMHO. I would vote for keeping things simple...

baiti:
> Recording the found date would require additional UI (date selection, etc.) and would make things more complicated without any real value IMHO. I would vote for keeping things simple...

I am wholeheartedly a fan of the KISS (Keep It Simple Stupid) principle. In the given case I would just hard coded take `now()` as the date when any state was changed, found it or DNF it.

But I am ok with just not even thinking about it ;)

Issue #10477, opened 2021-04-25
Hide or disable LC Connector in settings if not accessible for user


moving-bits:
Not sure whether to call this a bug or a feature request:

## Bug description / Feature request
Currently the LC connector will be shown to all users of c:geo, and can be enabled/disabled by all users, although actually the connector is only usable for users having a premium gc.com account.

LC connector should either be shown only to users fulfilling that preconditions, or be greyed-out for all others.
Currently only the description text beneath the "Activate" checkbox is greyed-out.

Alternatively add some explanation to which users can access the connector.

Lineflyer:
Few words regarding defaults BTW:
I suggest to have it active by default for people who have activated the geocaching.com connector (and are premium members).
This will promote the feature (and therefore I suggested the default text #10472 to inform user, who don't want it, how to disable it).

moving-bits:
> Few words regarding defaults BTW:
> I suggest to have it active by default for people who have activated the geocaching.com connector (and are premium members).

AFAICS I can set it to a fixed value only, either `false` or `true`. Currently it's set to `false` as default, and wrt to your request I would change this to `true` now. That doesn't harm, because if the user is no premium member, no calls to the API will be done, regardless of this setting.

For the original issue a PR is already merged, giving a textual explanation on the prerequisites for this service.

Lineflyer:
@moving-bits Actually this is not fixed, isn't it?
As basic member I do see an activated labs connector after installation and authorization to gc.com. 
I can see a non greyed out menu, where I can activate/deactivate the connector but it has (of course) no effect. The only hint it the description text, which notes that it is only available to PM.
IMHO the connector should be visible and accesible but the activate switch be greyed out.

Lineflyer:
Addon after reading the original issue text:
Users don't read any text, they see a normal connector and its active but not working. Therefore giving an explanation below is IMHO no solution to this issue, but we should make it unclickable which will potentially make the user read the text below then.

Lineflyer:
I am sorry, that I am reopening this one, there is still a small glitch in it:

- Install newest nightly
- Rung wizard and authenticate as gc.com basic member
- Look into Settings - Services

Result:
You will see geocaching.com as Active and also "Geocaching.com Adventure Lab" as active
If you click into "Geocaching.com Adventure Lab" you will see a greyed out "Activate" function but still checked.

Expected result:
I am convinced, this will irritate many basic users.
The service should not be "ticked on" and shown as active in this scenario.

[...]

PR #10485, opened 2021-04-26
add info on LC preconditions + move LC close to GC in settings

moving-bits:
## Description
- adds some explanation about preconditions for LC availability (rel. to #10477)
- moves LC connector in settings directly beneath GC connector
- LC info text is never greyed-out, even if connector is not active

image.image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -535,7 +535,7 @@
     <string name="settings_gc_description">Authorize c:geo with geocaching.com to search for caches.</string>
     <string name="settings_ec_description">Authorize c:geo with extremcaching.com to search for caches and access/filter your found caches.</string>
     <string name="settings_su_description">Authorize c:geo with geocaching.su to search for caches.</string>
-    <string name="settings_lc_description">Include Adventure Lab caches on the map.</string>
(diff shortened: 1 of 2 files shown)


moving-bits:
> > LC info text is never greyed-out, even if connector is not active
> 
> Do we really want to advertise premium features or rather hide it completely for BM?

I've been undecided about this. Actually my first implementation was to disable the whole LC preference screen if GC connector is not activated or user is no premium member. Worked fine, the only problem was that if you had activated the LC connector while being authorized as a PM, then switched to a BM, then you could no longer access the LC page. But that is probably a developer/tester problem only...

So if you prefer making it dependent on PM status (probably in combination with GC connector being activated) I can reactivate my first implementation.

Lineflyer:
> Do we really want to advertise premium features or rather hide it completely for BM?

I would not call it advertisement (while it is somehow), but we also show comparable toast/messages when trying to access a PM cache as basic member. 
Therefore I think its OK like suggested by @moving-bits for me. Don't get exactly yet though, how it looks for a BM in this proposal?!

moving-bits:
> Don't get exactly yet though, how it looks for a BM in this proposal?!

Regarding this PR there will be no difference, as I do not check PM/BM status.
(The connector itself will be active on both cases, but returns empty results if in BM mode.)

PR #10468, opened 2021-04-25
replace icon-creation by ribbonizer by dedicated images (version 2)

bekuno:
## Description

This is the other solution for #8896 described in #10408.
Instead creation a resourse tree for the buildTypes in the first version we use now image resources with dedicated names.

We have two variants of the images. For old devises below API 26 = Android 8 we have images with ribbons as shown in https://github.com/cgeo/cgeo/pull/10408#issuecomment-822053784 , for newer we use the adaptive icon. Here the icon background color is changed according the buildType.

## Details
- remove ribbonizer code
- add static ribbonized icons with own names for buildTypes rc (beta), nightly and debug
- add colored backgrounds for adaptive icons for the buildTypes above
- define separate appIcons for the buildTypes above
- set the appIcons for the buildTypes above in the "About c:geo" Windows

## Related issues
fixes #8896
prereq for upgrade gradle plugin to version 7.0

### App Icons adaptive
debug:
cgeo_background_blue
nightly:
cgeo_background_black
beta:
cgeo_background_red

### About c:geo Activity Icons adaptive
debug:
cgeo_about_blue
nightly:
cgeo_about_black
beta:
cgeo_about_red

@Lineflyer Please have a look at older devices if the icons are fine there.

PR #10486, opened 2021-04-26
Store fonts and developer shirt design

Lineflyer:
## Description

Store the following data in our repository:
- Fonts I have found in an archive which are either very similar or - in fact- the font we use for our logo
- Some developer shirt design graphics (still using the old logo though)

## Related issues

PR #10468


bekuno:
Do we need the main/project/design/shirts/t-shirt - Kopie.svg ?

Do we use the ubuntu font(s) somewhere?

Lineflyer:
> Do we need the main/project/design/shirts/t-shirt - Kopie.svg ?

Looks to me like the old front icon for the shirt?!

> Do we use the ubuntu font(s) somewhere?

I received those back in the days from @SammysHP.
Not sure if we use them?

bekuno:
> > Do we need the main/project/design/shirts/t-shirt - Kopie.svg ?
> 
> Looks to me like the old front icon for the shirt?!

The name should be without " - kopie".

Lineflyer:
> The name should be without " - kopie".

Corrected.

moving-bits:
Should we wait for https://github.com/cgeo/cgeo/pull/10468#issuecomment-827321675 before merging this PR?

BTW: Would it be ok at all (wrt the license) to merge those fonts into a public repository (in contrast to just using them in our graphics, if granted)?
Or should be better store them in some team-internal repo?

Lineflyer:
> BTW: Would it be ok at all (wrt the license) to merge those fonts into a public repository (in contrast to just using them in our graphics, if granted)?

The license is about using not storing.
You are right however as the whole repo is declared under a certain license which then does not apply to all its contents. But isn't it the same for things we use under CC-BY... licenses in contrast to our Apache license?

moving-bits:
Maybe I've read too many of those "20 pages printed in 3 point font size" license agreements with their hair-splitting phrases... ;-)

From what I can decipher from the readme, the license is not very specific and only distinguishes between personal use ("use freely") and commercial use ("you may be charged for it"). So we are probably on the safe side.

moving-bits:
anything open here (e. g. about the Ubuntu fonts), or can we merge?

PR #10497, opened 2021-04-28
correct imperial conversion in map settings (fix #10341)

### LINKED ISSUE #10341: Max. routing distance wrong for imperial units
- 2021.03.28-RC
- Use imperial units (Settings - Appearance)
- Open Settings - Map

For map content - Calculate route (maximum distance) the range is shown from 0.00mi up to 0.62mi.
This looks like a wrong conversion.

murggel:
## Description

use correct  conversion factor for imperial length for the proximity notification and max routing distance.

For proximity use ft as unit (meter in metric system), for max routing distance use yd as unit (km in metric system)

## Related issues

#10341


moving-bits:
Thanks @murggel for tackling this issue!

Issue #10387, opened 2021-04-15
find a suitable geocode for ALC's


fm-sys:
>But let  me explain what my struggle was with the subject of "geocode" all the time. Here is one example. There are many more like this:
>
> ```
>main/src/cgeo/geocaching/CacheDetailActivity.java:1173
>
>addContextMenu(details.add(R.string.cache_geocode, cache.getGeocode()).right);
>```
>
> `addContextMenu` can't be just modified like this:
>
> `addContextMenu(details.add(R.string.cache_geocode, cache.getShortGeocode()).right)`
>
> as it needs the **full** geocode to do its work. Since it does not get a reference to the `cache` object, it cannot do a `cache.getShortGeocode()` at the place where it is needed to bind the string to the ui property.
>
> Of course I could have added another parameter to addContextMenu and pass the `cache.getShortGeocode()` in addition but that can't really be the solution or can it?

Ok, it really seems to be not that easy as I would have expected it. Need to think about it...

> I remember there was a reference code in the beginning, which the user could even define hisself.
E.g. for my ALC I used VelAdventure01 and the link to https://labs.geocaching.com/goto/VelAdventure01 is still working.

Just asking, is this identifier available in theory (even if it may not be usable)? Or the short link like `https://adventurelab.page.link/KSQT`? 

But probably it's better to continue in a dedicated issue...

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/10372#issuecomment-820609621_

baiti:
`https://adventurelab.page.link/KSQT`.....  the KSQT in this case, can only be used in the context of an URL to launch GC's app. It is absolutely useless for any other purpose.

There are two API calls that are relevant for us. The first just takes coordinates and comes back with a list of cache objects.

E.g.: `https://labs-api.geocaching.com/Api/Adventures/SearchV3?take=1&origin.longitude=48&origin.latitide=9`


The second call matters: `https://labs-api.geocaching.com/Api/Adventures/3b97bc3a-3581-42c2-a8f3-d9ababdae800`

I had started by putting the short thing above into `cache.geocode` and the second into `cache.uuid`.

And then I ran into the problems described here: https://github.com/cgeo/cgeo/pull/10372#issuecomment-820609621

The system expects: `cache.geocode = 3b97bc3a-3581-42c2-a8f3-d9ababdae800` as that value is needed everywhere where cache.geocode is referenced today and it is all good except that it is too long when it comes to rendeing it at the UI.

[...]

PR #10498, opened 2021-04-28
shortened ALC LC-codes issue_#10387

baiti:
## Description
Adventure Lab Caches use a UUID for identification in the place where other geocaches use a GC-code aka geocode. The UUID is required as an addressing component when fetching data from the geocaching.com servers. The problem with the UUID: It  looks ugly at the UI and it cannot be memorized by the human brain for instance on a tour when one cacher says to another: "Hey what's the GC code for that cache over there ....".

This PR addresses the UI aspects of ALC-UUIDs in the place where a geocode is shown for other caches. There are three relevant scenarios, pls refer to https://github.com/cgeo/cgeo/issues/10387 for a more detailed discussion of these three cases and how they are addressed by this PR. In summary: One case is addressed by a modification in the respective layout.xml file ellipsizing the string, another case is addressed by a new method on cache objects: getShortGeocode() which truncates the geocode to 8 characters.

We should have a discussion whether "8" is an appropriate number as regular geocodes will get longer over time. At this time they are 7 characters long so it will take a very ong time until 8 characters lead to shortened strings even for non-ALC caches.

## Related issues

https://github.com/cgeo/cgeo/issues/10387

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1171,7 +1171,7 @@ public class CacheDetailActivity extends AbstractViewPagerActivity<CacheDetailAc
 
             details.add(R.string.cache_type, cache.getType().getL10n());
             details.addSize(cache);
-            addContextMenu(details.add(R.string.cache_geocode, cache.getGeocode()).right);
+            addContextMenu(details.add(R.string.cache_geocode, cache.getShortGeocode()).right);
             details.addCacheState(cache);
 
             cacheDistanceView = details.addDistance(cache, cacheDistanceView);
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -51,6 +51,7 @@ import cgeo.geocaching.utils.LazyInitializedList;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MatcherWrapper;
 import cgeo.geocaching.utils.ShareUtils;
+import static cgeo.geocaching.utils.Formatter.generateShortGeocode;
 
 import android.app.Activity;
 import android.content.Context;
@@ -88,6 +89,7 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
+
 
 /**
  * Internal representation of a "cache"
@@ -641,7 +643,7 @@ public class Geocache implements IWaypoint {
 
     @NonNull
     public String getShortGeocode() {
-        return (geocode.length() <= 8) ? geocode : (geocode.substring(0, 8));
+       return generateShortGeocode(geocode);
     }
 
     /**
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -639,6 +639,11 @@ public class Geocache implements IWaypoint {
         return geocode;
     }
 
+    @NonNull
+    public String getShortGeocode() {
+        return (geocode.length() <= 8) ? geocode : (geocode.substring(0, 8));
+    }
+
     /**
      * @return displayed owner, might differ from the real owner
      */
--- a/main/src/cgeo/geocaching/AbstractDialogFragment.java
+++ b/main/src/cgeo/geocaching/AbstractDialogFragment.java
@@ -162,7 +162,7 @@ public abstract class AbstractDialogFragment extends DialogFragment implements C
         final String cacheSize = cache.showSize() ? " (" + cache.getSize().getL10n() + ")" : "";
         details.add(R.string.cache_type, cacheType + cacheSize);
 
-        details.add(R.string.cache_geocode, cache.getGeocode());
+        details.add(R.string.cache_geocode, cache.getShortGeocode());
         details.addCacheState(cache);
 
         cacheDistance = details.addDistance(cache, cacheDistance);
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -1025,7 +1025,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         }
         final Geocache cache = adapter.getItem(adapterInfo.position);
 
-        menu.setHeaderTitle(StringUtils.defaultIfBlank(cache.getName(), cache.getGeocode()));
+        menu.setHeaderTitle(StringUtils.defaultIfBlank(cache.getName(), cache.getShortGeocode()));
 
         contextMenuGeocode = cache.getGeocode();
 
--- a/main/src/cgeo/geocaching/NavigateAnyPointActivity.java
+++ b/main/src/cgeo/geocaching/NavigateAnyPointActivity.java
@@ -104,7 +104,7 @@ public class NavigateAnyPointActivity extends AbstractActionBarActivity {
 
                 if (null != item) {
                     title.setCompoundDrawablesWithIntrinsicBounds(item.getType().markerId, 0, 0, 0);
-                    detail.setText(item.getGeocode());
+                    detail.setText(item.getShortGeocode());
                 } else {
                     detail.setText(context.getString(R.string.create_internal_cache));
                 }
--- a/main/src/cgeo/geocaching/WaypointPopupFragment.java
+++ b/main/src/cgeo/geocaching/WaypointPopupFragment.java
@@ -79,7 +79,7 @@ public class WaypointPopupFragment extends AbstractDialogFragmentWithProximityNo
             if (StringUtils.isNotBlank(waypoint.getName())) {
                 setTitle(waypoint.getName());
             } else {
-                setTitle(waypoint.getGeocode());
+                setTitle(waypoint.getShortGeocode());
             }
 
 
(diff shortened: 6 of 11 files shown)


main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -639,6 +639,12 @@ public String getGeocode() {
         return geocode;
     }
 
+    @NonNull
+    public String getShortGeocode() {
+        final int maxLength = (geocode.length() < 8) ? geocode.length() : 8;
+        return geocode.substring(0, maxLength);
```
fm-sys, line 645: IMHO we should add the "…" character, if the Geocode got shortened. What do you think?

fm-sys: Something like...

```
return (geocode.length() <= 8) ? geocode : (geocode.substring(0, 8) + "…");
```

baiti: I would leave it as it is, but it could be done of course.

Lineflyer: In this special case for ALC its relatively irrelevant for the user to know, that the geocode is shortened. But as AFAICS this is a generic implementation we should IMHO include the ellipsis character.

baiti: I wouldn't mind adding it if that suits the overall style better. After all, the code sequence @fm-sys  provided is much cooler than what I came up with.

baiti: I am now using @fm-sys' statement,  however without the dots but those could now be added with a much simpler modification later.

baiti:
I have found additional places where the ugly string shows up. Some of them I have addressed with the second commit into this PR. The tough part is to make sure that only strings that are bound to UI elements get the shortened version. If I by accident assigned the shortened string to a variable that is processed further down in the code, for instance to reference a cache object, things would be failing.

Also, there are a few more instances where I have not yet found the places in the code which require the getShortGeocode():

- Click on the map in an area where many caches are brings up a "Select Item" popup. It still shows the ugly LC code.
- CacheDetails, when swiped over to "Images", still shows the ugly LC code.

If any of you could give me a pointer where these views are renderred, it would help. I keep searching for those in the meantime.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -2079,7 +2079,7 @@ protected void fillViewHolder(final View rowView, final WaypointViewHolder holde
                 noteView.setOnClickListener(new DecryptTextClickListener(noteView));
                 noteView.setVisibility(View.VISIBLE);
                 if (TextUtils.containsHtml(wpt.getNote())) {
-                    noteView.setText(HtmlCompat.fromHtml(wpt.getNote(), HtmlCompat.FROM_HTML_MODE_LEGACY, new SmileyImage(cache.getGeocode(), noteView), new UnknownTagsHandler()), TextView.BufferType.SPANNABLE);
+                    noteView.setText(HtmlCompat.fromHtml(wpt.getNote(), HtmlCompat.FROM_HTML_MODE_LEGACY, new SmileyImage(cache.getShortGeocode(), noteView), new UnknownTagsHandler()), TextView.BufferType.SPANNABLE);
```
baiti, line 2082: Not sure whether the change for the SmileyImage is required or helpful at all. Same for ImagesActivity.

rsudev:
>     * Click on the map in an area where many caches are brings up a "Select Item" popup. It still shows the ugly LC code.

This is rendered in NewMap.showSelection()

HTH

baiti:
> > ```
> > * Click on the map in an area where many caches are brings up a "Select Item" popup. It still shows the ugly LC code.
> > ```
> 
> This is rendered in NewMap.showSelection()
> 
> HTH

I found the place and the ugly code comes back from:

Line 1376: `final StringBuilder text = new StringBuilder(item.getItemCode());`

after that statement, a `Log.e(text.toString());` shows:

```
2021-04-29 12:19:13.952 5117-5117/cgeo.geocaching E/cgeo: [main] LC7DE324A6-6043-47BD-89C3-51AAE96E15DF
2021-04-29 12:19:13.963 5117-5117/cgeo.geocaching E/cgeo: [main] 008QDVF
2021-04-29 12:19:13.967 5117-5117/cgeo.geocaching E/cgeo: [main] 0020R9R
2021-04-29 12:19:13.969 5117-5117/cgeo.geocaching E/cgeo: [main] GC62B4D
2021-04-29 12:19:13.991 5117-5117/cgeo.geocaching E/cgeo: [main] LC7DE324A6-6043-47BD-89C3-51AAE96E15DF
2021-04-29 12:19:14.004 5117-5117/cgeo.geocaching E/cgeo: [main] 008QDVF
2021-04-29 12:19:14.009 5117-5117/cgeo.geocaching E/cgeo: [main] 0020R9R
2021-04-29 12:19:14.010 5117-5117/cgeo.geocaching E/cgeo: [main] GC62B4D
```

... and due to my lack of Java skills I haven't found a way to truncate what ends up in `text`. 
It aint so easy as to say `text = text.substring(0,8);` I don't even grasp yet what a StringBuilder is and why that is needed in the first place. Had I tried to come up with this myself I'd have written:

`final String text = item.getItemCode().substring(0,8);` ... Which for reasons unknown to me would certainly not have worked otherwise whoever wrote this initially, would have done so.

And actually doing a hardcoded truncation there I'd consider a hack. Ideally I'd want to do a `cache.getShortGeocode()` somewhere but haven't found the place where that could go either.

eddiemuc:
What about adding a new method `getItemCodeShortened()` to class `GeoitemRef`?

```
    @NonNull
    public String getShortItemCode() {
        return itemCode == null ? "" : (itemCode.length() > 8 ? itemCode.substring(0, 8) : itemCode);
    }
```

BTW: a StringBuilder is a helper object to "build" a string. Because String is an immutable class, it is supposed to be inefficient to build up larger strings using + because each time the + is applied a new String object is created kust to be thrown away immediately after. See https://docs.oracle.com/javase/tutorial/java/data/buffers.html 
To get a `String` out of a StringBuilder, just call `toString() `on it.

baiti:
> @NonNull
>     public String getShortItemCode() {
>         return itemCode == null ? "" : (itemCode.length() > 8 ? itemCode.substring(0, 8) : itemCode);
>     }

Yes! Much better than my hack. I have named it `getShortItemCode()` to be consistent with `getShortGeocode()`

Now we have a few places where we compare to hardcoded `8`, what is the right place to define a constant such that it will be inclued in all classes and can be referenced where needed and only one change would be required in case we decide to make it 9 (in 5 years) or 10 (in 15 years).

main/src/cgeo/geocaching/models/Geocache.java
baiti: > Something like...
> 
> ```
> @NonNull
> public String getShortGeocode() {
>     return (geocode.length() <= 8) ? geocode : (geocode.substring(0, 8) + "…");
> }
> ```
> 
> Or even better, exclude the "8" to a private static class variable

@fm-sys : The "8" is needed in more than one class now. What is the best place to define it as a constant such that it get's included everywhere and can be referenced where needed, and changed only once when needed?

fm-sys: I would place it in Geocache.java

But also the shortening should be done in static utility function (also in Geocache.java) so that if we want to change the design of short geocode only one place needs to be edited

baiti: I also need it in `Waypoint.java` and in `GeoitemRef.java`. I had hoped there could be a system wide constant, perhaps somewhere like: `main/res/values/numbers.xml`

fm-sys: As I wrote, you should **not** rawly use it in  `Waypoint.java` or `GeoitemRef.java`! Create a utility function instead...

```
private static final int SHORT_GEOCODE_MAX_LENGTH = 8;

...

@NonNull
public String getShortGeocode() {
    return generateShortGeocode(geocode);
}


@NonNull
public String generateShortGeocode(final String fullGeocode) {
    return (fullGeocode.length() <= 8) ? fullGeocode: (fullGeocode.substring(0, 8) + "…");
}
```

So that `generateShortGeocode` can be used also in `Waypoint.java` and `GeoitemRef.java`. 

The general rule should be: Avoid duplications. Ideally, when you want to change something in the future, you should only have to edit one single place, cause no one will remember such hidden code duplications.

moving-bits: agreed, but shouldn't `generateShortCode` then be some static helper method in `utils/Formatter`, e. g. `formatShortItemCode()`?

main/src/cgeo/geocaching/CacheDetailActivity.java
fm-sys: Are you sure that all code places where you changed `getGeocode` to `getShortGeocode` really don't need the correct geocode? Otherwise, this might break existing functionality which wouldn't be worth it...

main/src/cgeo/geocaching/models/Geocache.java
fm-sys: Yes, you are right, that's even better.  :)

main/src/cgeo/geocaching/CacheDetailActivity.java
baiti: I understand the risks involved and no, I am not sure, which is the reason I have already started to backout some of those changes. And I fully agree, it wouldn't be worth it.

baiti:
static utiliy function as suggested by @fm-sys and @moving-bits commited and pushed. The import statement looks suspicious to me but AS wanted it that way.

Also: A word on the risks involved when calling `getShortGeocode()` potentially in a wrong place. Since this function is a NOOP for all GC codes except ALC, the potential impact is limited to something related to ALCs not working. A pretty small surface to check I'd say. At least this was my thinking when I put those method invocations in.

From my perspective it is now not only "working" it also should meet the conding style and quality expectations the team might have.

There is one exception: I haven't figured out where the string is bound to the resource in DetailsDetailsActivity when swiped to "Images", see screenshot above. Personally, as a geocacher I couln't care less but if it can be fixed it would certainly even better. @moving-bits could you have a look and perhaps throw another pointer over the fence into my court?

Lineflyer:
IMHO its not really relevant for the picture title.

fm-sys (approved):
LGTM

fm-sys:
For the image, I think we could also use a generalized name for ALC pictures instead of the shortened geocode... But I guess, that has no real priority ;-)

PR #10500, opened 2021-04-28
switch LC connector active default to true (fix #10477)

### LINKED ISSUE #10477: Hide or disable LC Connector in settings if not accessible for user

moving-bits:
## Description
- sets default for LC connector to activated==true (fixes #10477)
- removes an unused LC method in Settings
- adds info about required restart to the preferences screen

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -492,7 +492,7 @@ public class Settings {
     }
 
     public static boolean isLCConnectorActive() {
-        return getBoolean(R.string.pref_connectorLCActive, false);
+        return getBoolean(R.string.pref_connectorLCActive, true);
     }
 
     public static void setLCConnectorActive(final boolean value) {
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -495,10 +495,6 @@ public class Settings {
         return getBoolean(R.string.pref_connectorLCActive, true);
     }
 
-    public static void setLCConnectorActive(final boolean value) {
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -492,11 +492,7 @@ public static boolean isECConnectorActive() {
     }
 
     public static boolean isLCConnectorActive() {
-        return getBoolean(R.string.pref_connectorLCActive, false);
-    }
-
-    public static void setLCConnectorActive(final boolean value) {
-        putBoolean(R.string.pref_connectorLCActive, value);
+        return getBoolean(R.string.pref_connectorLCActive, true);
     }
```
baiti, line 497: interesting. I see the set function greyed out in AS, but how is the connector activated if the user checks the "Activate" box then?

baiti (approved):

main/src/cgeo/geocaching/settings/Settings.java
moving-bits: If you are using preference screens only to change settings, you would need no setter methods, as those changes are handled by Android internally. Setter methods are needed only if you want to be able to change a setting from elsewhere in the program, may it be a menu entry or some dialog option (as in map quick settings popup).

PR #10507, opened 2021-04-29
move offline routing settings to navigation (fix #10502)

### LINKED ISSUE #10502: Move routing related map options to navigation settings
Seeing that the Settings - Map get bigger and bigger I just thought it might be logical anyway to move the new sub chapter "Offline Routing" from maps to Settings - Navigation?!

IMHO it fits in quite good into the topic "Navigation" and the navigation menu contains only a few items...would be better balanced.

moving-bits:
## Description
- moves the "offline routing" block from map settings to navigation settings
- renames "Navigation" to "Offline routing / Navigation" and adapts its description

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -510,8 +510,8 @@
     <string name="confirm_data_dir_move">Moving the geocache data might take a really long time, depending on your device and the amount of data.</string>
 
     <string name="settings_title_basicmembers">Options for Basic Members</string>
-    <string name="settings_title_navigation">Navigation</string>
-    <string name="settings_summary_navigation">Choose preferred navigation methods and select external navigation apps.</string>
+    <string name="settings_title_navigation">Routing / Navigation</string>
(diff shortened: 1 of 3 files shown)


main/res/values/strings.xml
```diff
@@ -510,8 +510,8 @@
     <string name="confirm_data_dir_move">Moving the geocache data might take a really long time, depending on your device and the amount of data.</string>
 
     <string name="settings_title_basicmembers">Options for Basic Members</string>
-    <string name="settings_title_navigation">Navigation</string>
-    <string name="settings_summary_navigation">Choose preferred navigation methods and select external navigation apps.</string>
+    <string name="settings_title_navigation">Offline routing / Navigation</string>
```
Lineflyer, line 513: I would rather suggest "Navigation / Routing" to 
1. keep the known one in front
2. keep it a bit shorter (offline is mentioned in the subtitle anyway)

Lineflyer: Compromise: "Routing / Navigation" ?

PR #10514, opened 2021-05-01
store cache to default list on found state change (if needed)

### LINKED ISSUE #10501: LC needs to be stored for settings as found
## Bug description 

LC needs to be stored. Otherwise  "set as found" is not displayed. 

Cache should be stored automatically in the default list and "set as found" should always be available

moving-bits:
## Description
- always show found state change menu entries in cache popup, even if cache is not yet saved
- on selecting such a menu entry, store cache to default list, if cache is not yet stored offline
- fixes #10501

### DIFF
--- a/main/src/cgeo/geocaching/AbstractDialogFragment.java
+++ b/main/src/cgeo/geocaching/AbstractDialogFragment.java
@@ -253,7 +253,7 @@ public abstract class AbstractDialogFragment extends DialogFragment implements C
 
     @Override
     public boolean onOptionsItemSelected(@NonNull final MenuItem item) {
-        if (CacheMenuHandler.onMenuItemSelected(item, this, cache, null)) {
+        if (CacheMenuHandler.onMenuItemSelected(item, this, cache, this::init)) {
             return true;
         }
         if (LoggingUI.onMenuItemSelected(item, getActivity(), cache, dialog -> init())) {
--- a/main/src/cgeo/geocaching/CacheMenuHandler.java
+++ b/main/src/cgeo/geocaching/CacheMenuHandler.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.apps.navi.NavigationSelectionActionProvider;
 import cgeo.geocaching.calendar.CalendarAdder;
 import cgeo.geocaching.connector.internal.InternalConnector;
 import cgeo.geocaching.enumerations.LoadFlags;
+import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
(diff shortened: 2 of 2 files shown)


baiti:
I just noticed that when in CacheDetailActivity on an ALC, the "refresh" option has been removed. I do not know yet how I can track when that change was made and by whom, nor do I understand, why this capability has been removed. Was there a design discussion that I have missed? All I know is that in the initial PR that functionality was available. There may be good reasons, however, it could also be a bug? could anybody shed some light into this? Also, if there is a reason to remove that capability, then it should be removed from list views too. Currently, if I store ALCs into a list and go to: "Manage Caches -> Refresh All", the ALCs will be refreshed, which is exactly what I had actually expected to happen.

Now directly on the topic at hand: Manually setting the found state was most likely introduced for user-defined caches, those with a ZZ geocode right? For some reason ALCs now also have these capabilities to manually set the found state. Unfortunately, if I refresh the adventure from the servers that state is lost (which might be the reason why the refresh capability has been removed from CacheDetailActivity?). Of course I can still click the trash can in CacheDetailView and the download the adventure again, same results, the manually assigned found state is gone.

moving-bits:
> I just noticed that when in CacheDetailActivity on an ALC, the "refresh" option has been removed. I do not know yet how I can track when that change was made and by whom, 

In Android Studio you can use right-click, "Git", "History" or "History for selection" in the file you want to check.

> Now directly on the topic at hand: Manually setting the found state was most likely introduced for user-defined caches, those with a ZZ geocode right? 

correct

> For some reason ALCs now also have these capabilities to manually set the found state. 

This was introduced to be able to hide found ALC for as long as we are using the restricted functionality.

> Unfortunately, if I refresh the adventure from the servers that state is lost 

This should not be the case. Maybe we need to check this in the refresh code in the connector for as long as we don't gather the actual state online. (And as soon as we switch to gathering the state online, we should remove the ability to set it manually.)

> Could it be that it was actually me who killed it by removing the `ILogin` capability from the connector? If that is true, what if a connector can actually do the refresh without rquiring a login context? 

I guess that at the time the ILogin interface has been designed that case did not exist.

> And also, if this method returns `false` here, why is it still possible to do a refresh when the ALC shows up in a list? I would assume, the `supportsRefresh()` is not checked there.

This sounds like an error to me. If you find the source for it, it would be nice if you could provide a fix for it. Or create an issue here for others to fix.

> PS: I have just removed the `&& getConnector() instanceof ILogin` from the code shown above and the refresh capability is back so at least my understanding of the code is improving ;)

👍 

> So how do I fix this now? Adding an ILogin capability to the connector and make it always return a good logged in state?

Or maybe introduce a separate capability and separate "login" capability from "is able to refresh" capability (e. g. something like "IRefreshCapability")?

PR #10518, opened 2021-05-01
add option to export individual route as track (fix #10516)

### LINKED ISSUE #10516: Export individual route as track (!)

I would like to be able to export individual routes that I created in cgeo as GPX tracks (!). This way, they can also be useful in other apps/devices which themselves do not having routing capabilities, and you get the exact same path on them as you planned in cgeo.

## Suggested solution

In the route export window, a new option "export as track" should be added.

moving-bits:
## Description
- adds option "Export as track" to individual route menu on map
- renames option "Export individual route" to "Export as route" in individual route menu on map
- default filename for export as track is prefixed with "track" instead of "route"
- exported track has one track segment per waypoint of the individual route

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -178,6 +178,11 @@
                 android:visible="false">
             </item>
             <item
+                android:id="@+id/menu_export_individual_route_as_track"
+                android:title="@string/map_export_individual_route_as_track"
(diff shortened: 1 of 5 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

Issue #10523, opened 2021-05-02
Integrate Pocket Query Creator and/or remove link as useful app


Lineflyer:
## What function are you missing

Some users on support mail used to use the pocket query creator app (in conjunction with c:geo). Due to the latest website changes, PQC is however no longer working (login issue).

## Suggested solution

As the app looks unmaintained on Github, maybe its possible to integrate such feature into c:geo instead.
Anyway also it should be removed from the useful app list, if it won't be fixed.

## Alternatives/workarounds

I opened an [issue](https://github.com/robneild/pocket-query-creator/issues/21) on the PQC repo to ask for fix and for an information about the license of this code.

## Additional context

https://github.com/robneild/pocket-query-creator

moving-bits:
I've created #10524 for removal from useful apps list

Lineflyer:
Lets give them a chance to react first.

bekuno:
The app is licensed with GPL v3, see https://github.com/robneild/pocket-query-creator/blob/master/LICENSE.

Perhaps @pstorch has other contact information (as he has also been developing the app since 2014).

pstorch:
Long time ago I also distributed a little bit to PQ Creator. I have an eMail-Adress of Rob. If you like I can try to reach him.

moving-bits:
@Lineflyer @pstorch 
Any news on this?

Lineflyer:
Not from my side. The issue I opened in their repo is unanswered. I guess this tool is no longer maintained. 

Seems we should remove and maybe file an feature request issue to take this functionality over to c:geo. ;)

moving-bits:
Ok, will merge the PR then.

bekuno:
> 
> 
> Long time ago I also distributed a little bit to PQ Creator. I have an eMail-Adress of Rob. If you like I can try to reach him.

@pstorch 
No answer from Rob so far. Please try to contact him.

Lineflyer:
There was an answer now from Rob in my issue: https://github.com/robneild/pocket-query-creator/issues/21#issuecomment-842071261

robneild:
I am working on a fix but I am pretty busy and may take a few weeks

Lineflyer:
@robneild 
As also just wrote in https://github.com/robneild/pocket-query-creator/issues/21 I think this problem is still open.
Any news from your side or maybe you are interested to port the functionality into c:geo instead if you have no plan to maintain PQC any longer?

moving-bits:
[...]

PR #10524, opened 2021-05-02
remove PQ creator from useful apps (rel. to #10523)

moving-bits:
## Description
- removes "Pocket Query Creator" from list of useful apps
- including logo and related string resources


moving-bits:
Let's wait a bit for owner's reaction whether PQC will be adapted to the recent website changes or not. Setting WIP label for now.

Issue #10527, opened 2021-05-03
supportsRefresh() for ALCs


baiti:
I just noticed that when in CacheDetailActivity on an ALC, the "refresh" option has been removed. I do not know yet how I can track when that change was made and by whom, nor do I understand, why this capability has been removed. Was there a design discussion that I have missed? All I know is that in the initial PR that functionality was available. There may be good reasons, however, it could also be a bug? could anybody shed some light into this? Also, if there is a reason to remove that capability, then it should be removed from list views too. Currently, if I store ALCs into a list and go to: "Manage Caches -> Refresh All", the ALCs will be refreshed, which is exactly what I had actually expected to happen.

Now directly on the topic at hand: Manually setting the found state was most likely introduced for user-defined caches, those with a ZZ geocode right? For some reason ALCs now also have these capabilities to manually set the found state. Unfortunately, if I refresh the adventure from the servers that state is lost (which might be the reason why the refresh capability has been removed from CacheDetailActivity?). Of course I can still click the trash can in CacheDetailView and the download the adventure again, same results, the manually assigned found state is gone.

Update: I have done some code scans and greps. It seems the refresh (or reload) capability comes in via `cache.supportsRefresh()`. and that looks like:

```
public boolean supportsRefresh() {
        return getConnector() instanceof ISearchByGeocode && getConnector() instanceof ILogin;
    }

```
Could it be that it was actually me who killed it by removing the `ILogin` capability from the connector? If that is true, what if a connector can actually do the refresh without rquiring a login context? And also, if this method returns `false` here, why is it still possible to do a refresh when the ALC  shows up in a list? I would assume, the `supportsRefresh()` is not checked there. Will continue to investigate...

PS: I have just removed the `&& getConnector() instanceof ILogin` from the code shown above and the refresh capability is back so at least my understanding of the code is improving ;)

So how do I fix this now? Adding an ILogin capability to the connector and make it always return a good logged in state?

_Originally posted by @baiti in https://github.com/cgeo/cgeo/pull/10514#issuecomment-831245028_

baiti:
I now have added a fake `ILogin` capability to the connector and it brings back the `cache.supportsRefresh()` capability, however, it causes side effects, for instance on the main screen where my login state at GC is shown, here is now an additional line for LCs, e.g.

`LC . (0) .` ... which is ugly and  not really helpful.

So I believe adding the ILogin capability is not the right thing to do, I rather question the logic:
```
public boolean supportsRefresh() {
[...]

PR #10528, opened 2021-05-03
supportsRefresh and manually assigned found state for ALC (#10527)

baiti:
## Description
The LC connector doesn't need a login state, consequently the ILogin capability has been removed from the connector. Now when in CacheDetailActivity, there is no refresh capability for ALCs anymore. There is a function in `...model/Geocache.java` that controls whether the refresh capability is bound to the UI. Its state depends on the && condition:
`return getConnector() instanceof ISearchByGeocode && getConnector() instanceof ILogin;` It just checks the existence of the capablity, not whether `isLoggedIn == true`. So what is it good for in the first place? What happens if I remove the check for the ILogin capability? This PR makes that change, pls review and let me know if I am missing a point here.

Once this is fixed, we will run into the situation that refreshing an ALC wipes out any manyally set found state as the `LCapi` does the following: `cache.setFound(false);` ... at this time I don't know what to do instead of setting it hardcoded to a specific value. In passive mode (which is what we're currently using), the API does not give us the true found state.

Do I need to check somehow if the method is called for an initial fetch or a refresh? In the first case I'd execute a `cache.SetFound()`, in the second case I'd  not touch that property at all?

## Related issues
#10527


main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -166,7 +166,9 @@ private static Geocache parseCache(final JsonNode response) {
             cache.setDifficulty((float) 1);
             cache.setTerrain((float) 1);
             cache.setSize(CacheSize.getById("virtual"));
-            cache.setFound(false);
+            if (DataStore.isThere("LC" + uuid, "", false)) {
+                cache.setFound(false);
+            }
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.CACHE));
```
baiti, line 172: not sure why the condition shouldn't read `if (!DataStore ....)`. I thought `isThere` returns `true`if the cache already is in datastore in which case we cont want to do the `setFound(false)`, however, the way it is coded now just "works" as intended.

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -199,7 +203,9 @@ private static Geocache parseCacheDetail(final JsonNode response) {
             cache.setDifficulty((float) 1);
             cache.setTerrain((float) 1);
             cache.setSize(CacheSize.getById("virtual"));
-            cache.setFound(false);
+            if (DataStore.isThere(geocode + uuid, "", false)) {
```
fm-sys, line 206: Don't `geocode + uuid` result in a string like "LC_uuid_uuid"? Seems like you are adding the `uuid` twice...

main/src/cgeo/geocaching/connector/lc/LCApi.java
fm-sys: I guess this is a reason of how `Geocache` objects get merged in memory vs. DB. Have you checked if leaving `cache.setFound()` out completely would work as well (which would be the cleaner way)?

baiti: Actually I was preparing for the unlikely case that we might run the API in active mode in which case we get the found state from the API, the code reads now:

```
final Boolean isComplete = response.get("IsComplete").asBoolean();
...
cache.setFound(isComplete);
```

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: Yes, that is without any doubts a bug introduced via careless copy / paste. Thx much.

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: pushed a new commit addressing your findings

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -187,10 +191,12 @@ private static Geocache parseCacheDetail(final JsonNode response) {
             final String[] segments = firebaseDynamicLink.split("/");
             final String id = segments[segments.length - 1];
             final String uuid = response.get("Id").asText();
+            final Boolean isComplete = response.get("IsComplete").asBoolean();
+            final String geocode = "LC" + uuid;
             final String ilink = response.get("KeyImageUrl").asText();
             final String desc = response.get("Description").asText();
             cache.setReliableLatLon(true);
-            cache.setGeocode("LC" + uuid);
+            cache.setGeocode(geocode);
```
baiti, line 199: I have tried for a long time to avoid referencing "LC" hardcoded here. There is a getname() in the connector but no matter how I tried to referene it via getname() the compile always failed with something like can't use static on a non static or so. This is clearly due to my lack of Java skills at the moment.

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -601,7 +600,7 @@ private IConnector getConnector() {
     }
 
     public boolean supportsRefresh() {
-        return getConnector() instanceof ISearchByGeocode && getConnector() instanceof ILogin;
+        return getConnector() instanceof ISearchByGeocode;
```
fm-sys, line 603: Oh, and `ISearchByGeocode` is also implemented for `InternalConnector` so we should probably add a hard-coded exception for it.

main/src/cgeo/geocaching/connector/lc/LCApi.java
fm-sys: _If_ we want to / can support active mode, the found state should always be applied and not only if cache is stored. But as long as we are in passive mode, `found` should be `null` (means no value set, by which the previous found state will always be automatically taken over). 

If you want to, you can add a small comment instead 
(something like: `// cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode`)

main/src/cgeo/geocaching/connector/lc/LCApi.java
fm-sys: You can create a private (or public?) class variable (similar to like there already is with `CACHE_URL` in LCConnector) which then can be used in `getNameAbbreviated()` as well as the LCApi...

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: Done

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: `getNameAbbreviated()` is already there in LCconnector. What I don't know is how to reference this from within LCApi. Whatever I tried did not compile.

main/src/cgeo/geocaching/models/Geocache.java
fm-sys: ` return getConnector() instanceof ISearchByGeocode && !InternalConnector.getInstance().equals(getConnector());` should be suitable (tested just now)

baiti: Ok, the change is trivial, but what does it do? What state is expressed by the right part of the `&&`? Perhaps the answer lies in knowing what the purpose of the InternalConnector is?

main/src/cgeo/geocaching/connector/lc/LCApi.java
fm-sys: Yes, it is only available if you have an instance of LCConnector. Therefore, you need to create a class variable in `LCConnector` like:

```
    @NonNull
    protected static final String GEOCODE_PREFIX = "LC";
```
(`protected` means, it is only available in the current package. In your case `cgeo.geocaching.connector.lc;`)

And then you can use it everywhere in the code:

e.g. `cache.setGeocode(LCConnector.GEOCODE_PREFIX + uuid);` or 

```
    @Override
    @NonNull
    public String getNameAbbreviated() {
        return GEOCODE_PREFIX;
    }
```


---

But I think this is definitely optional. Do it the way you like it, for me the current code would be fine too.

main/src/cgeo/geocaching/models/Geocache.java
fm-sys: `Internal Connector` is used for creating "user defined caches" inside c:geo via long-tab on map (a bit like waypoints, which belong to no cache and can be used for tagging nearly everything). Since local caches can't be refreshed, they should be excluded...

grafik

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: I am going to fix it. My experience is such that I will always go back to those code sequences I once understood and use those mechanisms then in other places. This is now such a code sequence.

main/src/cgeo/geocaching/models/Geocache.java
baiti: Ah ok, and they support a `ISearchByGeocode` but can't be refreshed which is why the refresh function must not be bound to the UI, right? So just leaving out the condition on the rights side of the `&&` would have broken it for UDC aka ZZ-caches.

fm-sys: Yes, nothing really bad would happen though, but we would have a useless button for UDC-caches (as `Internal Connector` was the only one, which is instance of `ISearchByGeocode `but not `ILogin`)

baiti: Related to this subject is a discussion I had started in the team channel, it is mainly directed at @Lineflyer but it appears to me that he doesn't get notified on new content in those channels. If @Lineflyer is reading this here: Pls go check the internal channel for a discussion on the topic at hand here. It is about showing actual "found state", a feature which will be high on demand as soon as the first ALC functionality in cgeo hits the road.

Lineflyer: My PC was broken, now the new one arrived and I am "alive" again. :)
Will check back the team chat as time permits.

moving-bits: Still cannot understand why we do try to infer the "connector is able to refresh cache date" from its "supports search by geocode" and "can login" capabilities. That may have made sense at the time where the connector framework has been implemented, but as we see now, that's no longer valid. So I would vote for a (IMHO) more clean solution by explicitly adding the capability

by either adding something like a "IRefreshable" capability to all connectors supporting a refresh.

or by adding a method like `canRefresh()` to the `AbstractConnector`, defaulting to `true`, and overwriting this method in all connectors that can't.

`Geocache.canRefresh()` needs to be adapted to either checking the connector for `instanceOf IRefreshable` or for `canRefresh() == true`.

(Whether `canRefresh()` has a fixed value per connector or even checks login state would be up to us.)

fm-sys: OTOH "supports search by geocode" means that cache details can be fetched via a geocode, which implies that the cacheobject get updated via internet, which always will be the case for all potentially following connectors except the `InternalConnector`. 

> `canRefresh()` to the `AbstractConnector`, defaulting to `true`

This would be fine as well, but an `IRefreshable` is IMHO definitely to much

fm-sys:
@baiti FYI, I changed the target branch of this PR to release: 

grafik

No further action needed, but if you want to rebase the PR anyway, you need to rebase it onto the `release` branch

baiti:
> @baiti FYI, I changed the target branch of this PR to release:
> 
> grafik
> 
> No further action needed, but if you want to rebase the PR anyway, you need to rebase it onto the `release` branch

Thx for pointing this out as I have never done this before. This would be a new step for me which I won't do at this time as I don't want to screw up the tight beta release schedule by making some uneducated git mess. I won't touch my local copy until the PR is merged, in which case I then will delete my local branch and do a git pull upstream master (assuming that it will then also be in master).

fm-sys:
retest this please

baiti:
After realizing that Geocache properties could just be left uninitialized, getting `null` assigned in the constructor, I am leaving `terrain` and `difficulty` uninitialized to as these properties do not exist for LC caches. A sniff test in the emulator showed that  this indeed works. I am not sure whether there are any filtering capabilities in cgeo based on difficulty and/or terrain and whether this change might break them.

baiti:
In an attempt to refresh my local environment, rebasing the branch on master, it picked up changes in `main/res/values/changelog_master.xml`. What is the right way to back this out?

Lineflyer:
> In an attempt to refresh my local environment, rebasing the branch on master, it picked up changes in `main/res/values/changelog_master.xml`. What is the right way to back this out?

Rebase on `release` should work, though did not try this myself yet.

> I was wondering: Did this happen because I was rebasing to master instead of release?

Yes, exactly, you should have rebased to `release` as this is now targeted to `release`.

baiti:
> > In an attempt to refresh my local environment, rebasing the branch on master, it picked up changes in `main/res/values/changelog_master.xml`. What is the right way to back this out?
> 
> Rebase on `release` should work, though did not try this myself yet.
> 
> > I was wondering: Did this happen because I was rebasing to master instead of release?
> 
> Yes, exactly, you should have rebased to `release` as this is now targeted to `release`.

I did this right after I edited the quoted post and it did not make any difference to the PR

Lineflyer:
> I did this right after I edited the quoted post and it did not make any difference to the PR

Oh, yes you are right. This is as git now considers the commit as part of your PR and thus it will stay.

But it will work very easy with this:
git rebase -i  <commit ID of current HEAD of `release`>

Then in the editor just delete the lines with the unwanted commit.

In your example:
git rebase -i 5012e85b3

It will bring up an editor with all you commits after the specified commit:
```
pick 02aa39b96 Initialize changelog on master
pick 52c04a3bd remove ILogin from supportsRefresh()
pick c23a43768 don't wipe out manually set found state
pick 923e5829f build local variable geocode only once
pick c6d7fd368 removed double add of uuuid
pick 61562a83c don't cache.setFound() while API passive
pick ed70c72b4 supportsRefresh(): additional expresseion &&-ded
pick bef60dd85 removed multiple use of hard coded "LC" prefix
pick edf9a496f isArchived added
pick 75c52c147 difficulty / terrain removed
```

You can now just remove the unwanted commit 02aa39b96 before saving the file.
Ideally you can also use it to squash you commit by replacing the `pick` by `squash` on all but one commits.

Lineflyer:
If you have problems, just report back, I can also do it from here on this PR.

baiti:
> If you have problems, just report back, I can also do it from here on this PR.

I will go for it first. It is all about learning that stuff and thx much for the previous post where you explained it all very nicely

Lineflyer:
> > If you have problems, just report back, I can also do it from here on this PR.
> 
> I will go for it first. It is all about learning that stuff and thx much for the previous post where you explained it all very nicely

I recommend leaving out the squash at first, but just use `git log` before the interactive rebase, do the rebase as advised (BTW: Instead of deleting the unwanted line you can also use the (new) command drop in front of that line), and take a look at `git log` afterwards. That way I learned as well in the past.

fm-sys:
There is an easy sequence which I have in my GitHub cheat sheet:

### Retarget a branch from master to release
```
git rebase --onto release master
```
(If the changes were already pushed, you need `git push --force-with-lease` afterwards)

baiti:
> There is an easy sequence which I have in my GitHub cheat sheet:
> 
> ### Retarget a branch from master to release
> ```
> git rebase --onto release master
> ```
> 
> (If the changes were already pushed, you need `git push --force-with-lease` afterwards)

```
xxxxi@xxxx:~/cgeo$ git rebase --onto release master
fatal: Does not point to a valid commit 'release'
```

fm-sys:
Can you try to use the explicit commit hashes instead? (Even though it has actually always worked with `'release'` for me)

baiti:
> Can you try to use the explicit commit hashes instead? (Even though it has actually always worked with `'release'` for me)

Here is what I did in the interest of rebasing:

```
git switch master 
git fetch upstream master 
git push origin
git switch cache_refresh 
git rebase upstream/master 
git push -f
```
After that I realized that there was one file too many in the PR and after a while it occured to me that I should have
rebased on top of release, which I then did:
```
git pull upstream release 
git switch cache_refresh 
git rebase upstream/release 
git push -f
```
The last 4 commands didn't make any changes to the PR. I am now wondering there I should have also pushed relase to origin?

fm-sys:
> xxxxi@xxxx:~/cgeo$ git rebase --onto release master
> fatal: Does not point to a valid commit 'release'

Try to use `git rebase --onto upstream/release upstream/master`. I guess that could work... (hopefully :))

baiti:
> > xxxxi@xxxx:~/cgeo$ git rebase --onto release master
> > fatal: Does not point to a valid commit 'release'
> 
> Try to use `git rebase --onto upstream/release upstream/master`. I guess that could work... (hopefully :))

```
xxxx@xxxx:~/cgeo$ git rebase --onto upstream/release upstream/master
Current branch cache_refresh is up to date.
xxxx@xxxx:~/cgeo$ git status
On branch cache_refresh
Your branch is up to date with 'origin/cache_refresh'.

nothing to commit, working tree clean
```

Lineflyer:
The rebasing onto another branch will IMHO not work any longer after you initially rebased to `master`, because the relevant commit is not contained in `release` and such will be regarded to be included as part of your PR. That's why git says, that your working tree is OK.

Did you try the interactive rebase as suggested above?
Thats the way to go in such a case.

Lineflyer:
I just checked out this PR and applied the rebase locally as described above...Works like a charm.

baiti:
> 
> Did you try the interactive rebase as suggested above?

No, not yet. I have done the command sequence as posted a few comments above and then I tried the two variants of what @fm-sys suggested.

I will now go ahead and apply your method.....

baiti:
> > Did you try the interactive rebase as suggested above?
> 
> No, not yet. I have done the command sequence as posted a few comments above and then I tried the two variants of what @fm-sys suggested.
> 
> I will now go ahead and apply your method.....

Worked! all golden again now.

For completeness: after I did a `git pull release`, would I need to do a `git switch release && git push origin` too?

baiti:
The latest commit fixes this: #10546

fm-sys:
> The rebasing onto another branch will IMHO not work any longer after you initially rebased to `master`, because the relevant commit is not contained in `release` and such will be regarded to be included as part of your PR.

The `--onto` does the trick. I'm using this sequence on a regular basis, if I create a PR and forgot that the fix needs to be targeted to release. But probably, it only works if the repo is in sync. So maybe an additional `git fetch --all` is needed. 

Anyway, the interactive rebase is a very powerful tool, and it also often helps me when I don't know the "perfect" command.

BTW, hooray @baiti! Seems like you just won the git challenge. I must admit that git really does not make it easy for us ;-)

Lineflyer:
> For completeness: after I did a git pull release, would I need to do a git switch release && git push origin too?

Its not necessary for this scenario with the PR.
I however do it all the time to update my fork to the current state of the upstream branches just for "housekeeping" and to have an always up-to-date base if I checkout a new branch to work on.

Lineflyer:
retest this please

moving-bits (approved):
LGTM

PR #10531, opened 2021-05-04
optimize LC extra description box in cachedetails

fm-sys:
#10499 has caused the side effect, that a duplicated separator line is displayed for all non-LC caches:

Screenshot_20210504_091338

This is fixed with this PR...


moving-bits:
Good catch, thanks!

PR #10535, opened 2021-05-04
make use of google translate in-app translation popup

fm-sys:
The Google Translate App supports such cool "in-app translation popups" nowadays (appearance not configurable and dependents on the device):

Screenshot_20210504_163744 / grafik  

---

Unfortunately (or maybe even fortunately to keep things simple ;-)) it doesn't allow configuring the target language, but since it remembers the last selected language(s) it is IMHO totally fine...

| before | after (if Translate App is installed) |
|-|-|
| grafik | grafik |

BTW, I also reordered the "share text" item, so the elements are in a more logical order.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -167,6 +167,7 @@
 
 
     <!-- translation -->
+    <string name="translate">Translate</string>
     <string name="translate_to_sys_lang">Translate to %s</string>
     <string name="translate_to_english">Translate to English</string>
     <string name="translate_length_warning">Translation may fail with a large amount of text.</string>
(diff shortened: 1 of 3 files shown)


Lineflyer:
IMHO we used to have that already in c:geo...or am I mixing things up here.

fm-sys:
The default android floating text selection toolbar also provides the "translate" action, which launches the same Google Translate popup. But for logs, we are opening the browser / launch the "full" translation app if available currently...

grafik

fm-sys:
retest this please

Lineflyer:
> The default android floating text selection toolbar also provides the "translate" action, which launches the same Google Translate popup. But for logs, we are opening the browser / launch the "full" translation app if available currently...

Ah, yes. That was where I have seen that. Nice addition for the logs.

PR #10536, opened 2021-05-04
Long tab power user shortcut in cache details popup

fm-sys:
Allow changing the assigned cache icon in cache details popup via long click on the marker in the popups menu bar.


Lineflyer:
You mean in the popup on the map?

fm-sys:
Yes, exactly. We have many such hidden shortcuts (and I am constantly discovering new ones). Currently you have to open the cache to change the icon, so this shortcut can dramatically optimize the work flow of power users (or at least for me :) ) 

Screenshot_20210506_001623_copy_239x269

Issue #118, opened 2011-08-08
Support adding caches to bookmark lists


gcey:
The geocaching.com web interface provides the function "Bookmark listing", which I am using e.g. to add a cache to my FTF list right after logging the FTF.
It should be possible to do this from c:geo (because in the evening at home I might forget to do this).

Lineflyer:
Given the recent discussion about the newly introduced ignore feature I think that supporting bookmark lists would be a good idea as this would on one hand also implictly add the support for viewing the ignore list and one the other hand implement a feature which a lot of users ask for.

So this feature would need a support for adding a cache to a bookmark list (like adding it to ignore list, which is already implemented) as well as viewing existing bookmark lists.

mapa6:
Is it possible implement to c.geo app "add cache to bookmark list" (e.g. FTF) by the same way like "add to watchlist" ?

PR #10539, opened 2021-05-05
parse and allow download of bookmark lists

### LINKED ISSUE #8513: Support importing bookmarks like pocket queries directly
You can view bookmark lists directly in the official app. For using them in c:geo you have to work around by using the website or making an intermediary PQ.

With the new bookmark lists on the website, there is an API which should be quite easy to use (the website uses it too).

**Additional context**
With the new map, the bookmark lists have become quite useful as you can easily filter stuff from the map and have it available for download instantly. Maybe we could use this to speed up map downloads too.

fm-sys:
Support importing bookmarks like pocket queries directly (fix #8513)

Currently, this is only one-way-support, means download of bookmark lists but no upload. For upload, we have another issue open (see #118)

grafik


fm-sys:
retest this please

Lineflyer:
Looks nice.

I am not sure at this moment:
Are bookmark lists a PM-only feature like pocket queries? If not, we possibly cannot squash those features in one UI....

fm-sys:
> Are bookmark lists a PM-only feature like pocket queries?

Yes, they are. Only the watch list is also available for BM's...

A shared UI definitely makes sense since both are almost identical and compatible with each other. (BTW: In the official Geocaching App Pocket Queries are even shown as normal bookmark list)

Lineflyer:
Yes, in that case I fully agree.

moving-bits:
Cool, looking forward to it!

PR #10551, opened 2021-05-07
handle case that no file size can be specified while importing zipped GPX files

fm-sys:
PR #10539 doesn't show the import progress while importing the GPX. 

After debugging, it seems that the file size is stored after the GPX file inside the ZIP. So to get the file size, the whole file needs to be streamed. Since nothing has to be done with the file content, this initial read-through is relatively quick and IMHO it definitely is worth it so that the progress can be displayed in the further processing. (Maybe 1-2 extra seconds, but in exchange the user knows in the following 20 seconds that the app has not got stuck but is still importing.) 

Furthermore, the entire file will be read only if the file size could not be found with the today's method, so it will not slow down existing functionality. The issue occurred to me when importing a bookmark list GPX-ZIP-file, but can occur with other (malformatted?!) ZIP files as well...

### DIFF
--- a/main/src/cgeo/geocaching/files/AbstractImportGpxZipThread.java
+++ b/main/src/cgeo/geocaching/files/AbstractImportGpxZipThread.java
@@ -8,6 +8,9 @@ import cgeo.geocaching.utils.TextUtils;
 
 import android.os.Handler;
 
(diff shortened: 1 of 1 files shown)


fm-sys:
retest this please

PR #10562, opened 2021-05-08
use dark title color in UDC popup

moving-bits:
## Description
With the recent addition of setting a cache icon directly in the map's "create UDC" popup, the title is no longer visible when using light theme. This PR defines an adapted style which uses black title color for this popup.

|before|after|
|---|---|
|image|image|

### DIFF
--- a/main/res/layout/dialog_title_button_button.xml
+++ b/main/res/layout/dialog_title_button_button.xml
@@ -6,7 +6,7 @@
 
     <TextView
         android:id="@+id/dialog_title_title"
-        style="@style/action_bar_title"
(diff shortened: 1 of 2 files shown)


fm-sys:
@moving-bits does this implementation also work in dark theme?

fm-sys:
Just tested it, and it seems like we have fixed a bug for 1% of our users but introduced a new one for all other :-/ 
I already have a PR in the pipeline… ;-)

moving-bits:
ah, thanks, was working on two other things, had not noticed this one

Issue #10557, opened 2021-05-07
Optimize nearby results for ALC


Lineflyer:
This issue is coming up on support mail quite regularly for the newest beta. AFAIR we have similar issue for interworking with other connectors having a different cache density (Need to look those up...only on mobile at this moment). But this one might be worked around for now by limiting the amount of results.

Example scenario:
- Have GC and LC active
- Search by address "Köln"
- Results contains approx. 50 ALC
- Results contains first 20 GC

The additional problem for users is, that the ALC cover a much wider area than the GC (show result on map to visualize).
While this is a more generic problem for this special LC case, how about just return a max. of 20 ALCs?

moving-bits:
> The additional problem for users is, that the ALC cover a much wider area than the GC (show result on map to visualize).

This reminds me of another "distance related" issue: #7279

baiti:
I have made some more or less arbitrary assumptions:
```
    protected static Collection<Geocache> searchByCenter(final Geopoint center) {
        final Parameters params = new Parameters("skip", "0");
        params.add("take", "100");
        params.add("radiusMeters", "10000");
```
Of course thes could be adjuted to be a good citizen among all the others.

Lineflyer:
If this is only used for the searches(?) I would suggest to "take"=20 as a quick "equalizer" to the gc.com results (where we also take 20). This should IMHO be done on `release` for the upcoming release.

The remaining problem would be the radius, where we should instead find a generic solution covering possible different cache densities from different connectors and is discussed in #4594 and #7279.
These issues might get relevant for more users now as they will use two connectors from now on, while they potentially only used a single connector (GC) up to now.

baiti:
> If this is only used for the searches(?) I would suggest to "take"=20 as a quick "equalizer" to the gc.com results (where we also take 20). This should IMHO be done on `release` for the upcoming release.

There are only two API services used:

- search (here we get back and array)
- fetch (here we  get details for exactly one object)

Now comes the not so easy part: the API we use only provides a `searchByCenter()`. There is no radius expectation communicated to this method so the radius is hard coded.

cgeo provides a `searchByBBox()` which operates on two diagonal points which are sufficient to describe a rectangular area. The dimensions of that rectangle are defined by the selected zoom level of the map.

Since the API in use doesn't support that, the center of the bounding box is calculated with trivial math:

```
latcenter = LatitudeMax()  + LatitudeMin())  / 2;
[...]

Issue #10569, opened 2021-05-08
ALC live map loading needs optimization


Lineflyer:
## Bug description 

The visible area of the live map is not completely populated with adventure lab caches, itseems not optimized to cover the whole viewport in an adaptive way (just guessing)

### Steps to reproduce the problem

- Set global cache type filter to ALC (for easier view of the problem)
- Open live map and zoom out some steps

### Actual result after these steps 

ALCs are only shown in some central part of the map.
If you move the map another cluster around the center of the map gets shown.
Sometimes even some ALC vanish when new ones are loaded after slightly moving the map

### Expected result after these steps 

The map should show all available ALC in the shown viewport.
(Remark: Some limits might apply on low zoom as our internal cache cache / buffer has a certain maximum AFAIK)
So lets say:
The result should be comparable to what we do for "normal" caches on the live map. If the user sees some "normal caches" in an area of the map he assumes, that this is all. But in fact some ALC might be missing.

## c:geo version 

2021.05.08-NB

## Reproducible 

Yes

## Additional context 

IMHO this should not be fixed in a hurry, but carefully compared to what we do with GC:
We want to avoid to many server requests but still assure that we show reliable information...thats sometimes difficult. 
@moving-bits 
You optimized quite a lot for the live map in the past. Maybe you know how to make ALC behave in a comparable way.

jakedot:
Isn't this behaviour deliberate? It's what happens for the other online GC caches too isn't it?

Lineflyer:
> Isn't this behaviour deliberate? It's what happens for the other online GC caches too isn't it?

If you zoom out to much it will also happen to GC and others due to the limit of 1000 caches in total on the viewport (AFAIR).

But the case here is different:
I guess this is connected to the "search by center" instead of "Search in Box" as described by @baiti in another issue.
If you search a center and provide a fixed radius the chance is high, that we do not fill the screen (or at least fill until the internal maximum is reached).
We should at least calculate from the current viewport an indivudual radius to cover this viewport.

baiti:
If a rectangular area needs to be filled, it doesn't help to call a method called `searchByCenter()`, however, that is the **only** search function the ALC API supports.

I have constructed an interface to map a `seaarchByBBox()` to a `searchByCenter()` but the `searchByBBox()` gets never called.

[...]

PR #10571, opened 2021-05-08
Optimize nearby results for ALC #10557

baiti:
see #10557 for the associated discussion. In a nutshell: LC connector returned 100 caches hardcoded while other connectors returned only 20 which caused an unbalanced result on the map (LCs dominated). Now I am adjusting to `take=20` to be on par.

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -79,7 +79,7 @@ final class LCApi {
             return Collections.emptyList();
         }
         final Parameters params = new Parameters("skip", "0");
-        params.add("take", "100");
+        params.add("take", "20");
         params.add("radiusMeters", "10000");
         params.add("origin.latitude", String.valueOf(center.getLatitude()));
         params.add("origin.longitude", String.valueOf(center.getLongitude()));


Lineflyer:
Wait a second...is it about nearby results or about map or both? Sorry for not knowing the code behind so detailed.

baiti:
> Wait a second...is it about nearby results or about map or both? Sorry for not knowing the code behind so detailed.

there is only one place in the code where the `take=xx` can be set. The only search function ever called is `searchByCenter()` ... AFAIK.

Lineflyer:
OK, then lets use this fix first to fix #10557. 
But now I begin to understand how it works, and why we have #10569 -> For that one we probably need a search by box then.

Lineflyer:
@baiti 
Just a small side note for keeping more overview.
You might want to use keywords like "Fix #1234" in the PRs title or text to invoke auto-close on merge.
https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

baiti:
> OK, then lets use this fix first to fix #10557.
> But now I begin to understand how it works, and why we have #10569 -> For that one we probably need a search by box then.

Yes, and there are 3 connectors that support a searchByBBox() ... GC is not one of them and LC can get at best the "inner circle of a rectangular'. At this time it returns at most 20 LCs within a radius of 10km, no matter what the viewport might be. Clearly not good. Which is why I had it set to 100. My rationale was that the API is so much faster than the scraper such that it doesnt consume too much time.

baiti:
> @baiti
> Just a small side note for keeping more overview.
> You might want to use keywords like "Fix #1234" in the PRs title or text to invoke auto-close on merge.
> https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

so the magic word is "Fix" ? The title is: "Optimize nearby results for ALC #10557". It should have been "Optimize nearby results for ALC Fix #10557" ?

Lineflyer:
Yes, or "Fix #10557: Optimize nearby results for ALC"
Or use the "Related issues" section in the PR template and write "Fixes #10557".

PR #10577, opened 2021-05-08
Fixes #10569 ALC live map loading needs optimization

### LINKED ISSUE #10569: ALC live map loading needs optimization

baiti:
searchByBBox() now implements a radius calculation that allows to fill a rectangular map area. See discussion: #10569

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -75,11 +75,11 @@ final class LCApi {
         final double loncenter = (lon1 + lon2) / 2;
         final Geopoint gp1 = new Geopoint(lat1, lon1);
         final Geopoint gp2 = new Geopoint(lat2, lon2);
-        final double radius = gp1.distanceTo(gp2) / 2;
+        final double radius = gp1.distanceTo(gp2) * 500; // we get diameter in km, need radius in m
         Log.d("_LC Radius: " + String.valueOf((int) radius));
         final Parameters params = new Parameters("skip", "0");
         params.add("take", "500");
-        params.add("radiusMeters", String.valueOf((int) radius * 1000));
+        params.add("radiusMeters", String.valueOf((int) radius);
         params.add("origin.latitude", String.valueOf(latcenter));
         params.add("origin.longitude", String.valueOf(loncenter));
         try {
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -79,7 +79,7 @@ final class LCApi {
         Log.d("_LC Radius: " + String.valueOf((int) radius));
         final Parameters params = new Parameters("skip", "0");
         params.add("take", "500");
-        params.add("radiusMeters", String.valueOf((int) radius);
+        params.add("radiusMeters", String.valueOf((int) radius));
         params.add("origin.latitude", String.valueOf(latcenter));
         params.add("origin.longitude", String.valueOf(loncenter));
         try {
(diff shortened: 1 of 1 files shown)


baiti:
ohh  nooo ... 165 files changed! git again.

baiti:
> ohh nooo ... 165 files changed! git again.

I guess I fixed it .... puuh. I was about to give that git challenge award back to @fm-sys

fm-sys:
Sorry! I had not thought of it, but we also have a method in the `Geopoint` class which does the calculation for us and makes things way easier:
`oneGeopoint.distanceTo(otherGeopoint)`

baiti:
> Sorry! I had not thought of it, but we also have a method in the `Geopoint` class which does the calculation for us and makes things way easier:
> `oneGeopoint.distanceTo(otherGeopoint)`

Oooook .... that really simplifies the code a lot. Will do tonight.

baiti:
@Lineflyer I'd consider it ready for a merge. At least from a coding perspective the diff should now be reasonably simple.
I have done a sniff test on the emulator and I believe it is "ok" now, however, it would certainly be valuable if you did a quick test too.

Lineflyer:
Sure, I will merge it tonight and merge upstream to test in nightlies before updating the beta version out there next days.

baiti:
Don't merge yet. The new "toDistance" function seems to get "km", I need "m" .... factor of 1000 needs to be applied.

baiti:
With commit https://github.com/cgeo/cgeo/pull/10577/commits/39a398378a2f7a08bf19b91ea1d642d5e9fc2e8e I just did, it should now finally be working.

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -67,10 +67,27 @@ protected static Geocache searchByGeocode(final String geocode) {
...
+        final double lon2 = viewport.getLongitudeMin();
+        final double latcenter = (lat1 + lat2) / 2;
+        final double loncenter = (lon1 + lon2) / 2;
+        final Geopoint gp1 = new Geopoint(lat1, lon1);
+        final Geopoint gp2 = new Geopoint(lat2, lon2);
+        final double radius = gp1.distanceTo(gp2) * 500; // we get diameter in km, need radius in m
+        Log.d("_LC Radius: " + String.valueOf((int) radius));
+        final Parameters params = new Parameters("skip", "0");
+        params.add("take", "500");
+        params.add("radiusMeters", String.valueOf((int) radius);
```
fm-sys, line 82: A ")" is missing here...

baiti: that happens if you're not at the android studio system and commit changes without doing a compile first.

I need to get my commandline gradle comand working. Perhaps that is the opportunity now to learn it. How do I trigger a build from the commandline?

Lineflyer:
Thanks, lets try it out :)

main/src/cgeo/geocaching/connector/lc/LCApi.java
fm-sys: I think a `./gradlew build` should do the trick. But cannot say as I don't use it myself...

PR #10585, opened 2021-05-10
Fix: Enable editing of imported waypoints with empty coordinates (fix #6915)

### LINKED ISSUE #6915: Handling of waypoint without coordinates in GPX files
Currently c:geo only exports waypoints with coordinates. Waypoint without coordinates (like #6883 or final waypoints read from the website) are skipped during export.
Import of waypoints without coordinates is possible by defining `lat` and `lon` as empty `""`. But this is not according to the GPX schema.

One idea was to use `lat=0` and `lon=0` which would be valid to the GPX schema but this is actually a valid coordinate. Defining this as NULL would be a "trick".

murggel:
## Description

Set `original coordinates empty` for imported gpx (e.g. GSAK, not exported via c:geo) for waypoint with 0/0
Otherwise coordinates of owner defined waypoints with empty coordinates could not be assigned

## Related issues

fix #6915

## Additional context

Only with the flag original coordinates empty set, the user can edit pre-defined waypoints with empty coordinates.
In some cases (e.g. gpx exported from GSAK) the flag is not set: the flag is only set, if the cgeo-extension is contained in the gpx.

### DIFF
--- a/main/src/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/cgeo/geocaching/files/GPXParser.java
@@ -311,10 +311,12 @@ abstract class GPXParser extends FileParser {
                         waypoint.setLookup("---");
                         // there is no lookup code in gpx file
 
-                        if (wptEmptyCoordinates) {
+                        waypoint.setCoords(cache.getCoords());
+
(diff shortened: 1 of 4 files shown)


murggel:
Ready for review

Issue #10583, opened 2021-05-09
Use short geocode on target name


Lineflyer:
## Bug description 

The target name display on map after you have set a cache as target (showing the navigation line towards it) is still using the long geocode of ALC caches. This leads to the problem, that the cache name is no longer visible.

### Steps to reproduce the problem

- Open live map
- Open popup for an ALC
- Click "Set as target" icon

### Actual result after these steps 

The top of the map shows e.g. "LC2131232-2312312-12312323-23123123..." but ommiting the name with follows after the geocode.

### Expected result after these steps 

Use the shortened geocode at this place (like we do on cache lists, etc)

## c:geo version 

2021.05.08-NB

## Reproducible 

Yes

## Screenshots 

![image](https://user-images.githubusercontent.com/949669/117587063-55b41d00-b11c-11eb-900c-f4345a75fb2d.png)

Lineflyer:
Fixed in #10592

PR #10592, opened 2021-05-11
short LC names and conditional D/T texts (Fixes 10583)

baiti:
Lots of description here: #10583

A word on "conditional D/T" texts: LC caches don't have D/T ratings. When the cache object is instantiated, the respective properties are left uninitialized. In various views this did the job and no D/T related strings were shown, however, in Map view when a target was set, `D 0.0 . T 0.0 .....` was shown. This PR addresses that bug in addition to showing short LC  code as reported in the referenced issue.


baiti:
There is another instance of "the ugly string" ™: It's in the loading popup where the spinner is shown. What do you think, is it worth while to go for this one too? We could just ignore it as it is a transient view only. Of course if we wanted to be perfect, this has to be addressed too ;)

Lineflyer:
>There is another instance of "the ugly string" ™: It's in the loading popup where the spinner is shown. What do you think, is it worth while to go for this one too? We could just ignore it as it is a transient view only. Of course if we wanted to be perfect, this has to be addressed too ;)

I think we can leave it like it is. No functional impact at that point.

PR #10594, opened 2021-05-11
use ellipsize in pocket query activity actionbar

fm-sys:
| before | after |
|-|-|
| Screenshot_20210511_182402_cgeo geocaching | grafik |


fm-sys:
retest this please

Issue #9285, opened 2020-10-25
Update or retire legacy version (API 16-20)


moving-bits:
Play Store is enforcing new policies regarding background access to location:
![image](https://user-images.githubusercontent.com/3754370/97119716-4cc8a480-1712-11eb-9814-7505681551cf.png)

see https://developer.android.com/distribute/play-policies

Although we're not using background access to location (AFAIK), Play Console still marks our app as violating the policies:
![image](https://user-images.githubusercontent.com/3754370/97119666-02dfbe80-1712-11eb-9b03-4314b5637e91.png)

TODO: Check if there is any action to be taken by us (applying for something, doing a code change or whatever)

@Lineflyer @mucek4

mucek4:
The only reason for c:geo that needs background permission is "talk to navigate". That's something that does not work at all on Huawei as when display is off app is paused. 
I am not sure if we change cgeo tu run in background (Icon in task bar) if location permission needs to be permanent or that counts as running.

Lineflyer:
I am wondering about this "background location permission" in general, as when I first time start c:geo the permission dialog offers:
- Allow during usage of the app
- Do not allow
- Do not allow and don't ask again

The first item sound like "foreground permission" and I have also seen other apps where one more option "Always allow" is shown on the permission dialog. 
Maybe its only a configuration issue (knowing that Talking compass will anyhow not work in background as also @mucek4 described, but thats a different story).

Lineflyer:
@moving-bits 
Can you provide a link of where you found the information shown on your second screenshot?

Lineflyer:
According to https://developer.android.com/training/location/permissions we do not explicitly request background location permission and we also target SDK29.
I guess the "problem" why we are shown as violating is, that we also have the legacy version targeting a lower SDK.
So IMHO we can just take the questionaire and confirm, that we do not intentionally use background location for that APK.

Lineflyer:
Addition:
Or simply drop legacy support before March 2021.

Lineflyer:
Offtopic (just on the way of renewing the age allowance check on Google Play):

I am unsure about "Yes" or "No" at this point  :) ...just kidding

![grafik](https://user-images.githubusercontent.com/949669/97304198-af798780-185b-11eb-8961-d44b54a2d019.png)

Lineflyer:
> @moving-bits
Can you provide a link of where you found the information shown on your second screenshot?

Nevermind, I found it meanwhile.
[...]

PR #10596, opened 2021-05-11
Attempt to fix cgeo-legacy

Lineflyer:
## Description

Without knowing what I do I just cherry-picked all commits, which I recognized to be related to the last batch of website changes and cherrypicked them onto `cgeo-legacy`. Some conflict resolution in `GCConstants.java` was needed, where I again only assumed the correct proceeding.

I did not yet check if it compiles and works at all. 
Review is welcome...and lets see what CI reports for that.

## Related issues

#9285


bekuno:
This was a great training for conflict resolving. But for now I could not reache a working state.

Beside the viewpagerindicator fix
`git cherry-pick fe553a4570949cfa61f91b0d79a52d34ed5e283e`
we need at first the undobar change 
`git cherry-pick 1a6aa42149f32dc176f6c5b1cbacd5e1289a7eab`, but also
some more `GCLogin` fixes, see https://github.com/cgeo/cgeo/commits/33bb92601edd839201b3b551526035df740495d9/main/src/cgeo/geocaching/connector/gc/GCLogin.java
Until now the result compiles, but the c:geo start crashes. I have to do it once more better. Don´t know, when it will be.

PR #10603, opened 2021-05-12
add filters for favorites, distance, enhance attribute filter, perform cleanup

eddiemuc:
* perform some cleanup in filter framework
* enhance attribute filter (allow include/exclude, clear button, tool tips)
* add favorite filter
* add distance filter


eddiemuc:
image

image

image

PR #10604, opened 2021-05-12
Better user experience for internal routing settings

### LINKED ISSUE #10599: Improve internal routing first use (routing profiles)
## Bug description 

If a "fresh user" first time activates internal routing he will see, that the routing profiles menus are empty.

- Open Settings - Routing
- Enable internal routing
- You will see Walk/BIke/Car profiles don't show active profiles
- Tap on one of them

It shows an empty menu
image

image

We could/should provide some default files already upon installation if possible or at least show an explanational text in the emtpy selections, that profiles will be downloaded together with first routing tiles. 

## c:geo version 

2021.05.11-RC

## Reproducible 

Yes

### LINKED ISSUE #10600: Routing profiles not set in menu after first tile downloads
## Bug description 

The routing profiles seem not correctly initialized after profiles have been received

- Activate internal routing and automatic download in Settings - Routing
- Go to the map and set a cache as target
- Automatic download is started (after confirmation)
- Wait for download and copy to complete
- Go into Settings - Routing

- Car profiles shows the actual selected profile, however Walk and Cycle show empty but if you enter the menu, they are set in fact.
- 
image

Used routing profiles should be shown in the settings menu

## c:geo version 

2021.05.11-RC

## Reproducible 

Yes

## Screenshots 

see above

## Additional context 

Related to #10599 (if profiles would be available before "first use" it could implicitly solve the situation

moving-bits:
## Description
- Check profile files on Setting's start as well (fixes #10599)
- Fix a bug related to filling the profile preferences' summary (fixes #10600)

### DIFF
--- a/main/src/cgeo/geocaching/settings/SettingsActivity.java
+++ b/main/src/cgeo/geocaching/settings/SettingsActivity.java
@@ -6,6 +6,7 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.apps.navi.NavigationAppFactory;
 import cgeo.geocaching.apps.navi.NavigationAppFactory.NavigationAppsEnum;
 import cgeo.geocaching.brouter.BRouterConstants;
+import cgeo.geocaching.brouter.util.DefaultFilesUtils;
 import cgeo.geocaching.connector.ConnectorFactory;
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

PR #10608, opened 2021-05-12
Alc optimizations

baiti:
as discussed.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -86,6 +86,7 @@
     <string translatable="false" name="pref_log_offline">log_offline</string>
     <string translatable="false" name="pref_logimages">logimages</string>
     <string translatable="false" name="pref_choose_list">choose_list</string>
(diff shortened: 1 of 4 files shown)


main/res/values/preference_keys.xml
```diff
@@ -253,6 +253,9 @@
     <string translatable="false" name="pref_su_tokenpublic">su_tokenpublic</string>
     <string translatable="false" name="pref_temp_su_token_secret">su-temp-token-secret</string>
     <string translatable="false" name="pref_temp_su_token_public">su-temp-token-public</string>
+    <string translatable="false" name="pref_cache_filter_config">cache_filter_config</string>
```
baiti, line 256: the `cache_filter_config` must be the result of an wrong resolution of a merge conflict. At least I have not intentionally added that line.

eddiemuc: @baiti I recently added this preference key to master for the new cache filter. It should not be on release, but it doesn't hurt there either. 

Maybe this was somehow left over innyour local copy when you switched between master and release?

PR #10611, opened 2021-05-13
alc optimizations (3rd edition)

baiti:
now based on a completely recreated release branch from master and synced to origin.

@Lineflyer ... you can now decide which one to merge. The other one can be closed then.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -276,6 +276,8 @@
     <string translatable="false" name="pref_map_routing">map_routing</string>
     <string translatable="false" name="pref_theme_menu">theme_menu</string>
     <string translatable="false" name="pref_localstorage_version">localstorage_version</string>
+    <string translatable="false" name="pref_alc_advanced">alc_advanced</string>
+    <string translatable="false" name="pref_alc_launcher">alc_launcher</string>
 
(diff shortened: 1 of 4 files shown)


Lineflyer:
No problem to have it in `release`.

baiti:
As a next iteration, I am scanning the code for how to disable the "Images" Screen on CacheDetailActivity (for ALCs).  What would be an elegant way to turn that view off without writing too much if ... then ... else logic checking for cacheType or something similar. Any pointers where to look and for what to look would be appreciated.

Lineflyer:
Why turning it off?

baiti:
> Why turning it off?

Becasue all it shows is the image that can already be seen on the "Description" Page. It is completely redundant.

Lineflyer:
No, its not 100% redundant:
1. You can use this page to open the image in an external app if you e.g. want a closer look on it
2. It could (potentially) also be used for stage images if this might come as well

baiti:
> No, its not 100% redundant:
> 
> 1. You can use this page to open the image in an external app if you e.g. want a closer look on it
> 2. It could (potentially) also be used for stage images if this might come as well

Ah ok I wasn't aware of this. So we leave everything just the way it is. Thx for pointing it out.

PR #10617, opened 2021-05-13
fix #10559: load/save filter capabilities, userdisplay text for filters

### LINKED ISSUE #10559: Add possibility to save/load filter configurations
Following Wilhelm Busch ;) there is of course already a new idea, borne while playing with it: Now that I have constructed a really complex filter (to BTW find exactly what I was looking for - REALLY great!) I would love to be able to save this painstakingly constructed filter in e. g. a personal filter set, Or to ex/import the SQL statement ...  or copy/paste it, to e. g. store it in a text file ... :)

Would be great if you would put this on the wish list ...  ;)

However, I'm *really* glad with the way this is looking right now. Great work! Congrats! My hat is off to you.

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/10147#issuecomment-834866609_

eddiemuc:
* adds capability to load and save filters
* for this, GUI also needed to change
* adds user-display-string for filters


eddiemuc:
image

image

image

image

Issue #10615, opened 2021-05-13
Receive map file extremely slow (for ZIP files)


gcey:
## Bug description 
The download of a 1 GB map file takes a few minutes, but then the "Receive map file" takes several hours and during this phase c:geo is very irresponsive.

## c:geo version 
2021.04.25

## Reproducible 
Yes

## Additional context 
Samsung Galaxy S9+, Android 10, map folder on SD card.

moving-bits:
Which map source are you using - OpenAndroMaps or one of the others?

(OpenAndroMaps is delivering the map as a zipped file, which reduces downloading time, but increases transfer time, whereas the other map sources use map files directly.)

gcey:
Zip file from Freizeitkarte.
But unzipping should if course not take hours.

eddiemuc:
Does this improve significantly if map folder is on internal storage instead of as card?

gcey:
I checked it now with a map folder in the internal storage and also there it is extremely slow (1MB per minute).

moving-bits:
@gcey 
Can you cross-check with a non-zipped map file of similar size, if it's that slow there as well?

(A file of 1 GB should be downloaded and transferred in less than 15 minutes normally, if there's enough space and memory available, fast internet access provided.)

gcey:
Now I tested it with "bayern.map" (422 MB) from Mapsforge.
The "Receive map file" took just a few seconds!

moving-bits:
Thanks for cross-checking. So indeed it seems to be related to zipped files.
(Or maybe to RAM consumption. How much free RAM does the device have approximately on which you experience that problem?)

gcey:
According to "Memory" in the "Developer Options" 3.4 GB of 5.9 GB are in use (while NOT in "Receive map file").

gcey:
Now (with c:geo 2021.05.20) it works perfectly.
Many thanks !!!

PR #10618, opened 2021-05-13
enhance user experience on receiving download (rel. to #10615)

moving-bits:
## Description
- tries to fix issue #10615
- fixes a full scan on opening the zip (will now stop after first found map file)
- uses `BufferedOutput`
- avoids flooding UI with progress messages (zip had been updating in chunks of mostly less than 1k!, thus very frequently)
- outputs the processed size in K/M/G etc.

## Additional context
I'm not sure if that really fixes the original issue, but in several tests c:geo now stays responsive (where we had ANR before) on starting to transfer a downloaded file, and while the system's download manager took more than 10 minutes for a 590 MB file to download, unzipping to app. 700 MB and transferring it to c:geo was done in about 50s in the emulator.

@gcey:
Do you have a chance to test this build when it's compiled, to check whether it fixes the problem you described in #10615?

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2081,6 +2081,7 @@
 
     <!-- ReceiveDownloadActivity -->
     <string name="receivedownload_intenttitle">Receive file</string>
+    <string name="receivedownload_receiving_file">Receiving file \"%s\"</string>
     <string name="receivedownload_success">File successfully stored as \"%s\"</string>
     <string name="receivedownload_error">Error while receiving file</string>
     <string name="receivedownload_error_io_exception">I/O error while receiving file. Is your folder \"%s\" writable?</string>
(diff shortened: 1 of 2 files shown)


gcey:
> Do you have a chance to test this build when it's compiled, to check whether it fixes the problem you described in #10615?

I think, I can only test it, when it is released officially.

moving-bits:
I just measured some times using the current code:

For above's file (~590 MB ZIP, ~700 MB unzipped) the app took about 8s for opening the ZIP file, producing an ANR while doing so, and then took a whole 8 1/2 minutes for unzipping/transferring the file.

Compared to this my test with this PR was much faster: No ANR and only 50s for the whole unzipping/transferring.

It's been just a single measurement, so it may not be representative, but it sounds promising..., so let's give it tray.

gcey:
Now (with c:geo 2021.05.20) it works perfectly.
Many thanks !!!

Issue #10621, opened 2021-05-14
Gc.com search for attributes not implementable: WebApi proxy v2: cgeo explicitely not authorized?


eddiemuc:
I found something fishy today when I tried to integrate V2 of gc.coms Web proxy Api.

The following things were all done while logged in as a premium member.

The following Uri is an example for a search of traditional caches (ct=2) with the V2 Proxy WebApi:
https://www.geocaching.com/api/proxy/web/search/v2?ct=2&take=500&skip=0&sort=distance&asc=true&app=cgeo 

When I issue it exactly like this I get an "unauthorized" message:
`{"statusCode":401,"statusMessage":"Unauthorized","errorMessage":"","uri":"http://websearch.groundspeak.biz/web/search/v2?ct=2&take=500&skip=0&sort=distance&asc=true&app=cgeo"}`

When I issue the SAME request with V1, it works:
https://www.geocaching.com/api/proxy/web/search?ct=2&take=500&skip=0&sort=distance&asc=true&app=cgeo 

And now the mean part: when I issue the same request (for V2) with an `app`-parameter-value **other** than `cgeo` or without the `app` parameter, it works too! E.g.:
https://www.geocaching.com/api/proxy/web/search/v2?ct=2&take=500&skip=0&sort=distance&asc=true&app=tumdedum
https://www.geocaching.com/api/proxy/web/search/v2?ct=2&take=500&skip=0&sort=distance&asc=true

Are they really explicitely blocking c:geo from using the V2 Proxy WebApi???

eddiemuc:
I saw my original statement wrt "other app" was a bit confusing. I meant "another value than `cgeo` for the parameter `app`". I edited issue description accordingly.

eddiemuc:
...and this is a problem because some of the filtering options of the online map seem only to be availablein v2, e.g. attribute filtering

baiti:
Just wondering: Is there any documentation around for this API?

Without any doubts, `app=cgeo` is explicitly kept out. But: why do you specify the app parameter at all? Don't you get all you need without specifying it?

eddiemuc:
> Just wondering: Is there any documentation around for this API?

Not that I know of. But it is very easy to reverse engineer it by looking at the requests GC.com sends when using different filter options on their website.

> Without any doubts, `app=cgeo` is explicitly kept out. But: why do you specify the app parameter at all? Don't you get all you need without specifying it?

[...]

PR #10622, opened 2021-05-14
search with filters against gc.com

eddiemuc:
First draft to query with filters against gc.com Proxy WebAPI. 

Looks promising, but since v2 is not usable, attribute filtering is not possible (see #10621 )


eddiemuc:
image

image

image

Issue #10579, opened 2021-05-09
Refactor settings to use PreferenceFragments instead of PreferenceActivity


moving-bits:
## Help needed
While deep-diving into the theming issue (#9373) yesterday and playing around with different strategies, I've stumbled upon our outdated preferences implementation. We are still using `PreferenceActivity` with several preference screens, which is deprecated for a long time. What is worse is that this is technically not compatible to the rest of our app, which is using AndroidX and `AppCompatActivity` - which means a lot of extra work and inconsistencies.

As I have zero experience with fragment handling I'm asking for help, so if someone is willing to migrate our preferences from `PreferenceActivity` to `AppCompatActivity` and `PreferenceFragments` (or even some full up-to-date AndroidX way), please go ahead, as this will help us moving forward towards a consistent theming approach.

fm-sys:
I've implemented a preference activity with AndroidX for two other Apps I wrote from scratch. I can try to, but I'm not sure if will have enough time to refactor this. Btw, it seams like multi screen preferences aren't really supported anymore which means we have to implement multiple activities/fragments...

fm-sys:
@rsudev you have started refactoring our preferences some time ago, haven't you? Is there any progress?

bekuno:
@SchoolGuy As you migrated the settings in WhereYouGo to Fragments (PR https://github.com/cgeo/WhereYouGo/pull/240), can you support or implement this also for c:geo?

SchoolGuy:
@bekuno Certainly yes. I am a bit busy catching up with things since I wrote my final exams on Wednesday but I guess I can come up with something in the next two weeks.

bekuno:
@SchoolGuy  I wish you that your final exams were very good. Feel free to help whenever you have some spare time :)

moving-bits:
Setting prio to "high" here, as using the old `PreferenceActivity` is not compatible to material design, not even to `AppCompatActivity`, and therefore requires lots of extra code and ugly workarounds to work at all, and even with those workarounds does not fit completely into our new material look...

I guess we should do this in several steps: Start with a single preference screen, migrate it to `PreferenceFragment`, and have it as a parallel menu entry besides our current settings. And then migrate screen by screen until the old preferences are gone.

[...]

PR #10630, opened 2021-05-15
add hidden filter

### LINKED ISSUE #10634: [New filter framework] Filter for "Hidden" date
A filter shall be implemented to search for "Hidden" date of caches.

eddiemuc:
Adds a "hidden" filter (including its integration in gc.com filter search) to new filter framework
fixes #10634

### DIFF
--- a/main/res/layout/date_range_selector_view.xml
+++ b/main/res/layout/date_range_selector_view.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/advancedfilter_generic_date"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:orientation="vertical">
+
+    <RelativeLayout
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:orientation="horizontal">
+
+        <TextView
+            android:id="@+id/date_from_label"
+            style="@style/text_label"
+            android:layout_width="100dp"
+            android:textAlignment="textEnd"
+            android:text="@string/cache_filter_datefilter_label_from"/>
+
+        <Button
+            android:id="@+id/date_from"
+            style="@style/button_full"
+            android:layout_centerVertical="true"
+            android:layout_toRightOf="@+id/date_from_label"
+            tools:text="5/3/2021" />
+    </RelativeLayout>
+
+    <RelativeLayout
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content">
+
+        <TextView
+            android:id="@+id/date_to_label"
+            style="@style/text_label"
+            android:layout_width="100dp"
+            android:textAlignment="textEnd"
+            android:layout_alignParentLeft="true"
+            android:text="@string/cache_filter_datefilter_label_to"/>
+
+        <Button
+            android:id="@+id/date_to"
+            style="@style/button_full"
+            android:layout_alignParentRight="true"
+            android:layout_toRightOf="@+id/date_to_label"
+            android:layout_centerVertical="true"
+            tools:text="5/3/2021" />
+
+    </RelativeLayout>
+
+
+
+</merge>
--- a/main/src/cgeo/geocaching/filters/gui/DateRangeFilterViewHolder.java
+++ b/main/src/cgeo/geocaching/filters/gui/DateRangeFilterViewHolder.java
@@ -1,0 +1,34 @@
+package cgeo.geocaching.filters.gui;
+
+import cgeo.geocaching.filters.core.DateRangeGeocacheFilter;
+import cgeo.geocaching.ui.DateRangeSelector;
+
+import android.view.View;
+
+
+public class DateRangeFilterViewHolder<F extends DateRangeGeocacheFilter> extends BaseFilterViewHolder<F> {
+
+
+    private DateRangeSelector dateRangeSelector;
+
+    @Override
+    public View createView() {
+
+        dateRangeSelector = new DateRangeSelector(getActivity());
+        return dateRangeSelector;
+    }
+
+    @Override
+    public void setViewFromFilter(final DateRangeGeocacheFilter filter) {
+        dateRangeSelector.setMinMaxDate(filter.getMinDate(), filter.getMaxDate());
+    }
+
+    @Override
+    public F createFilterFromView() {
+        final F filter = createFilter();
+        filter.setMinMaxDate(dateRangeSelector.getMinDate(), dateRangeSelector.getMaxDate());
+        return filter;
+    }
+
+
+}
--- a/main/src/cgeo/geocaching/ui/DateTimeEditor.java
+++ b/main/src/cgeo/geocaching/ui/DateTimeEditor.java
@@ -1,8 +1,11 @@
 package cgeo.geocaching.ui;
 
+import cgeo.geocaching.R;
 import cgeo.geocaching.utils.Formatter;
+import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.functions.Action1;
 
+import android.app.AlertDialog;
 import android.app.DatePickerDialog;
 import android.app.Dialog;
 import android.app.TimePickerDialog;
@@ -34,6 +37,9 @@ public class DateTimeEditor {
     private FragmentManager fragmentManager;
 
     private Calendar date;
+    private final boolean[] dateUnset = new boolean[]{ false };
+    private boolean allowUserToUnset = false;
+    private Action1<Date> changeListener = null;
 
     /**
      * Initializes the instance. Typically called in an activities 'onCreate' method.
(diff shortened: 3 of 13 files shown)


main/src/cgeo/geocaching/filters/gui/DateRangeFilterViewHolder.java
```diff
@@ -0,0 +1,54 @@
...
+//        rlp = new RelativeLayout.LayoutParams(ViewGroup.LayoutParams.WRAP_CONTENT, ViewGroup.LayoutParams.WRAP_CONTENT);
+//        rlp.addRule(RelativeLayout.ALIGN_PARENT_RIGHT);
+//        rlp.addRule(RelativeLayout.RIGHT_OF, R.id.label);
+//        rlp.setMargins(ViewUtils.dpToPixel(10), ViewUtils.dpToPixel(10), ViewUtils.dpToPixel(10), ViewUtils.dpToPixel(10));
+//        rl.addView(dateButton, rlp);
+//
+//        viewGroup.addView(rl);
+//        return dateButton;
+//    }
+
```
moving-bits, line 40: Can you remove the commented-out parts?

main/src/cgeo/geocaching/ui/DateTimeEditor.java
```diff
@@ -132,12 +171,20 @@ public Dialog onCreateDialog(final Bundle savedInstanceState) {
 
             // Create a new instance of DatePickerDialog and return it
             final DatePickerDialog dialog = new DatePickerDialog(getActivity(), this, year, month, day);
-            dialog.onDateChanged(dialog.getDatePicker(), year, month, day);
+            //dialog.onDateChanged(dialog.getDatePicker(), year, month, day);
```
moving-bits, line 174: Can you remove the commented line?

main/res/layout/date_range_selector_view.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
```
moving-bits, line 1: As I guess that you have already a couple of layout files for the new filtering system, and that there will be many more: May it be useful to put them all into a subfolder of `layouts` to reduce that endless list a bit and to group those filter-related files?

Lineflyer: We also recently discussed styles/themes in our dev-meeting. AFAIR @eddiemuc started with a new base style for the filtering and the idea could be anyway to establish a "second layout setup" (with proper parenting, inheritation, etc.) in parallel and whenever we touch an area we try to bring it onto this cleaner style hierarchy. 

Or is this a whole other topic and I misunderstood your comment?

moving-bits: Sadly I've missed that discussion completely due to my not being back in time...

So far I was only speaking of organization of layout files in the file system. Currently we have more than 100 layout files in a linear list, which is pretty annoying to handle when working with layouts, so some grouping may help here, especially if there will be lots of new files.

With regards to new base layout: I've also started with a new base layout, including some nice method for switching between dark and light mode, but was blocked by the fact, that our preferences are based on a different type of activities, not being compatible to `AppCompatActivity`, which would require us to maintain two sets of themes/styles. That's why I opened #10579.

main/src/cgeo/geocaching/filters/gui/DateRangeFilterViewHolder.java
eddiemuc: done

main/src/cgeo/geocaching/ui/DateTimeEditor.java
eddiemuc: done

main/res/layout/date_range_selector_view.xml
eddiemuc: @moving-bits 
> As I guess that you have already a couple of layout files for the new filtering system, and that there will be many more: May it be useful to put them all into a subfolder of `layouts` to reduce that endless list a bit and to group those filter-related files?

Correct, we should think about better structuring our layout files. But specifically this one (d`ate_range_selector_view.xml`) is **not** supposed to be cache-filter-specific. In an attempt to make gui components more reusable I created a custom view `DateRangeSelector `which may be used now anywhere one needs to make the user a date-range (currently of course this is only needed in cache filter). 

Up to now I try to use a name scheme for layout files, with the prefix specifying the place of usage (e.g. `cache_filter_`) and the postfix trying to specify the kind of usage (e.g. `_view`, `_activity`, `_fragment`). Consequently you will find all really cache_filter-specific layout files by their prefix.

I suggest to move the topic of reorganizing layout files to its own issue, it goes beyond the scope of cache filters. Would that be ok?

eddiemuc: > We also recently discussed styles/themes in our dev-meeting. AFAIR @eddiemuc started with a new base style for the filtering and the idea could be anyway to establish a "second layout setup" (with proper parenting, inheritation, etc.) in parallel and whenever we touch an area we try to bring it onto this cleaner style hierarchy.
> 
> Or is this a whole other topic and I misunderstood your comment?

@Lineflyer , @moving-bits 
This discussion was about incorporation https://material.io/ as the base of c:geo styles/themes to apply the design capabilities of this for a GUI rework. But I haven't done amything real on this up to now anyway, only thing I did so far is creating the new themes dark_material and light_material as copies of dark/light but inheriting from the material-themes instead of the appcompat themes. And the new activity CacheFilterActivity is the only one using these new themes so far. 

I can easily undo this again if wanted?

The only views really needing to be based on material thems currently are the new ContinuousRangeSlider and ItemRangeSlider; and those can also directly inherit from material themes (I use ContextThemeWrapper inside anyway). 

So should I remove this stuff for now? (but please in a separate issue)

moving-bits: > I suggest to move the topic of reorganizing layout files to its own issue, it goes beyond the scope of cache filters. Would that be ok?

Sure. I just got triggered by seeing your hard work on the new filters and the many layout-related things you do. I'll open a separate issue for organizing the layout files.

Lineflyer:
> I suggest to move the topic of reorganizing layout files to its own issue, it goes beyond the scope of cache filters. Would that be ok?

For sure!

main/res/layout/date_range_selector_view.xml
fm-sys: > In an attempt to make gui components more reusable I created a custom view DateRangeSelector which may be used now anywhere one needs to make the user a date-range (currently of course this is only needed in cache filter).

Something like this already exists as material design component: https://material.io/components/date-pickers

> But I haven't done amything real on this up to now anyway, only thing I did so far is creating the new themes dark_material and light_material as copies of dark/light but inheriting from the material-themes instead of the appcompat themes.

I don't know if this is a good idea. I have already created many PR as there were crashes caused by missing attribute definitions in different themes. Why can't we simply switch all appcompat themes to material (which are AFAIK also based on appcompat)?

moving-bits: (wrt to work on new base layout)

> So should I remove this stuff for now? (but please in a separate issue)

No, please go ahead! So far I've done a couple of things on resource organization wrt to dark mode / light mode - there is a nice new automatism for it, so that you don't have to apply dark theme / light theme anywhere, but set one config flag, and Android chooses the right resources for you. - But this mechanism does not work with the old preferences implementation we currently use, so that's one of the reasons why we should move forward to using `AppCompatActivity` and `PreferenceFragment` for it.

Another thing I've started working on along the way is placing most of the attribute values for themes/styles into the base theme, which allowed me to cut off a couple of in-between-themes, and also reduces the pain of whether a certain attribute is available in current theme or not.

But all of this can be merged at a later time, no worries.

eddiemuc: > > In an attempt to make gui components more reusable I created a custom view DateRangeSelector which may be used now anywhere one needs to make the user a date-range (currently of course this is only needed in cache filter).
> 
> Something like this already exists as material design component: https://material.io/components/date-pickers

Saws that and tried it first. It did not work for some reason (renders very strange, selection bubbles are off-place), also when using the "pure" Material Themes. I did not want to spend too muich time just now, and luckily for date/time select we had a rather good picker almost ready (only thing missing was the ability to also "deselect" a value which is neede in context of filter). 

Wrapping a View class around this gives us the ability to later replace the inner works with e.g. the Material-DatePicker without much fiddling elsewhere

eddiemuc: > > But I haven't done amything real on this up to now anyway, only thing I did so far is creating the new themes dark_material and light_material as copies of dark/light but inheriting from the material-themes instead of the appcompat themes.
> 
> I don't know if this is a good idea. I have already created many PR as there were crashes caused by missing attribute definitions in different themes. Why can't we simply switch all appcompat themes to material (which are AFAIK also based on appcompat)?

This should not happen here any more since the themes are based on Material-Themes, so all necessary attributes are there. Actually this is the way that Material.io wants you to work: make your own theme, but inherit from the Material ones as a base (actually our themes "dark" and "light" are doing the same but inherit from AppCompat - thanks to the work of @moving-bits. Material Themes are also based on AppCombat but have some extensions necessary for Material components as far as I understand).

eddiemuc:
Discussion is already a bit "off-topic", I think I can merge this PR now

main/res/layout/date_range_selector_view.xml
eddiemuc: > Another thing I've started working on along the way is placing most of the attribute values for themes/styles into the base theme, which allowed me to cut off a couple of in-between-themes, and also reduces the pain of whether a certain attribute is available in current theme or not.
> 
> But all of this can be merged at a later time, no worries.

Great! Thanks for your work. I did not look into the "auto-dark-mode" setting thing yet.

PR #10631, opened 2021-05-16
remove old c:geo logos

moving-bits:
## Description
We still had a couple of the old c:geo logos present in our resources (actionbar_cgeo.png, cgeo.png, cgeo_sendto.png, cgeo_logo.png), which this PR removes.

I've checked the sources and found no more usages. Also the build compiles fine after removing them and I was still able to create shortcuts etc. Still someone might cross-check if I've missed some indirect usage, therefore I target this to `master`.


main/lint.xml
```diff
@@ -46,8 +46,6 @@
 
     <!-- resources loaded by dynamic name lookup -->
     <issue id="UnusedResources">
-        <ignore path="res/*/cgeo_logo.png" />
-        <ignore path="res/*/cgeo_sendto.png" />
```
fm-sys, line 50: What does `resources loaded by dynamic name lookup` mean? Could it be, that they are used for our notification system? Just wondering why we have explicitly excluded them from lint...

moving-bits: Yes, our notification system may be one reason (as it is with the `attribute_maintenance.png`).
But I've done a text search on all files in `res` and `src` folder and could not find a reference to `cgeo_logo` and `cgeo_sendto` any longer, therefore removed them.

fm-sys: Maybe @Lineflyer can help us here? AFAIK the notification system icons aren't defined in-app, but the resource name is provided via the status message.

But you are right, that these old icons should be removed. Maybe we can just rename the "new_icon" resources so that they match the old names...?

Just guessing...

Lineflyer: The notification system can make use of all png graphics located in `master/main/res/drawable-*` and the name of the resource to use will be send within the reply from the status server to the client, which then uses it.

However we won't use the old logo there I think.

bekuno (changes requested):
Please move  `main/res/drawable-hdpi/cgeo_sendto.png` to `main/project/rawimages`. We do not have a original one of this file.
Maybe we should move later the outdated images in an `outdated` folder.

main/lint.xml
bekuno: > 
> 
> What does `resources loaded by dynamic name lookup` mean? Could it be, that they are used for our notification system? Just wondering why we have explicitly excluded them from lint...

This tries to describe `<issue id="UnusedResources">`.
Without this exclude you have messages for it in the CI. console. (CI dynamic looks for its using)

moving-bits:
> Please move `main/res/drawable-hdpi/cgeo_sendto.png` to `main/project/rawimages`. We do not have a original one of this file.
> Maybe we should move later the outdated images in an `outdated` folder.

Why should we keep those outdated files as active part in our repository?
If we ever got to need them again (which I doubt) git will have them for us.

So I vote for deleting them with this PR.

fm-sys:
> 
> 
> > Please move `main/res/drawable-hdpi/cgeo_sendto.png` to `main/project/rawimages`. We do not have a original one of this file.
> > Maybe we should move later the outdated images in an `outdated` folder.
> 
> Why should we keep those outdated files as active part in our repository?
> If we ever got to need them again (which I doubt) git will have them for us.
> 
> So I vote for deleting them with this PR.

We can move them to our wiki page [Project roots and history](https://github.com/cgeo/cgeo/wiki/Project-roots-and-history), so we have them (visibly) archived. But I would vote for moving them out of our active repo as well...

bekuno:
We have more outdated images in rawimages. So move the sendto image first in this folder and later externalize all outdated images to the wiki would be a good way.

moving-bits:
I'm still not convinced: I thought `rawimages` is for images in their source version (SVG source file, PNG used within c:geo), not for visibly storing outdated files. I would prefer to either remove them (git will still have them if we need them - just checkout an older commit) or at least to move them to a dedicated folder, e. g. `outdated` or something like that.

But anyway, I've moved the file to `rawimages` to get this little cleanup processed, and I'll open a different issue to remove all outdated files from the active part of the repo.

eddiemuc:
I am 100% with @moving-bits . Things no longer used by the app should not remain in the git repo.

I know of corner cases where a project has no other way of storing such things (but we have, e.g. wiki), then the usual workaround is to create a dedicated folder in the repo (something like "`archive`" or so). While this is still a bad solution (all developers plus CI plus everyone using the repo always have to transfer these outdated things on every `pull/fetch`) it is still preferable over having them in "productive" folders like `rawimages` (where an unknowing developer might even reference these outdated things in code by accident)

moving-bits:
Let's merge here, as the requested changes are implemented, and we have opened a separate issue for future handling of outdated image resources.

PR #10640, opened 2021-05-16
switch changelogs to markdown (fix #6586)

### LINKED ISSUE #6586: Can we use a normal changelog?
I was wondering, if we could provide a changelog.md file as proposed at http://keepachangelog.com. I generally like human readable change logs in open source projects, and ours is unfortunately not very human readable (due to XML formatting and being hidden in some folder).

What's your opinion? Should we try to maintain a purely markdown formatted changelog, and include it into the app resources via gradle build automatically? I'm sure the conversion itself can be done easily, however I'm not sure yet about our separation of changelog_release and changelog_master...

moving-bits:
## Description
While browsing old issues this morning I've stumbled upon #6586 - switch changelogs to markdown.

This has several advantages:
- Modifying changelogs is easer - no HTML needed, no XML, no escaping of umlauts and apostrophies
- We can link to the new `changelog_master.md` / `changelog_release.md` from our github pages, so that users can easily view current changes
- Managing the change info for new releases will be a matter of copy & paste only

(and maybe more which I have yet to explore, e. g. formatting capabilities)

Instead of the two changelog XML files in `values` we will now have two plain text markdown files in `assets` (see changed files in this PR).

For migration I propose to start with `changelog_master` and ignore the release changelog on `master` for now (or manually update it), and with next branching from `master` to `release` that branch will be switched as well.

Current changelogs would look like this:
image


Lineflyer:
Two notes from my side:
- Have fun with codacy :( Last time I was not able to solve the issues about indentation, maybe you will be successful. 
- While you are working in this area. Any chance to make the files localized so was can include them to Crowdin?

moving-bits:
> Have fun with codacy :( Last time I was not able to solve the issues about indentation, maybe you will be successful.

Well, just saw it - 46 "issues" (or pseudo-issues) - Codacy seems to be very picky. But maybe this can be configured somehow. I'll look into it...

> While you are working in this area. Any chance to make the files localized so was can include them to Crowdin?

Hm, how does Crowdin work with regard to complete files? E. g.:
- How do we manage translations for app description etc.?
- Does it recognize if a single line in a file got added/changed/removed and only request our translators to update that line, or will the whole file have to be translated again? 
- How can I recognize how much has been translated?

Lineflyer:
> How do we manage translations for app description etc.?

We can pass over the files we want. However we need to define new files for each languages, where translations will end up.
Not sure though, how we can handle that in the app but outside the resource area (which does the language selection for us).

> Does it recognize if a single line in a file got added/changed/removed and only request our translators to update that line, or will the whole file have to be translated again?

When handing over markdown or plain text you can/should set the flag `content_segmentation` for that file and Crowdin will split it into segments (using their own sometimes strange algorithm), which can each be translated. The output files will then contain mixed content until the whole file is translated.

> How can I recognize how much has been translated?

Only manually.

Lineflyer:
> Well, just saw it - 46 "issues" (or pseudo-issues) - Codacy seems to be very picky. But maybe this can be configured somehow. I'll look into it...

The "no trailing :" seems OK as an issue, the indentation and empty line things are IMHO rather irrelevant. I am sure it can be configured, but last time I just deactivated Codacy for the markdown files we use on the Codacy page. A graceful configuration of what checks should be done, would of course be better.

moving-bits:
> > How do we manage translations for app description etc.?
> 
> We can pass over the files we want. However we need to define new files for each languages, where translations will end up.
> Not sure though, how we can handle that in the app but outside the resource area (which does the language selection for us).

Ok, that seems feasible - creating new files is a one-time thing, and for handling that within the app there will be a way. Maybe adding the language selectors ("-de", "-en", "-zh-rTW") to the filename, and asking the system, which language selector is currently used. Need to test a bit for that.

> > Does it recognize if a single line in a file got added/changed/removed and only request our translators to update that line, or will the whole file have to be translated again?
> 
> When handing over markdown or plain text you can/should set the flag `content_segmentation` for that file and Crowdin will split it into segments (using their own sometimes strange algorithm), which can each be translated. The output files will then contain mixed content until the whole file is translated.
> 
> > How can I recognize how much has been translated?
> 
> Only manually.

Ok, if the first part works reliably (content segmentation, and returning full set of strings, mixed non-translated and translated), then the second part won't be needed. Maybe we can test both on a small subset of languages.

Lineflyer:
We can define the names and paths of the translated files with the placeholders listed here:
https://support.crowdin.com/vcs-integrations-configuring-online/#creating-a-configuration

Ideally something like `/main/assets/changelog/%two_letters_code%/`

fm-sys:
We should disable Codacy for it. That's how we have done it for all other markdown files.

For translation purposes, android AFAIK supports different assert folders (`/assets/`, `/assets-de/`, `/assets-pt/`, ...). Android will then automatically choose the file from the most suitable folder (as like `strings.xml` translations)

fm-sys:
> Ideally something like `/main/assets/changelog/%two_letters_code%/`

To make use of Androids automatic resource management:
`/main/assets-%two_letters_code%/changelog_master.md`
`/main/assets-%two_letters_code%/changelog_release.md`

Lineflyer:
> Ok, if the first part works reliably (content segmentation, and returning full set of strings, mixed non-translated and translated), then the second part won't be needed. Maybe we can test both on a small subset of languages.

Should work, see https://github.com/cgeo/cgeo/blob/master/main/project/playstore/description/translated/Latvian/app_description_long.txt as an example of not completely translated content.

Lineflyer:
> To make use of Androids automatic resource management:...

This is also working for things outside the `values`? Cool!

moving-bits:
Just played a bit with localization and got it to work from within the `raw` folders:

image

Lineflyer:
@moving-bits Did you see the suggestion where to place them from @fm-sys above?

I can take care of enabling on Crowdin later on. Just tell me, where the translated files need to end up.

main/build.gradle
```diff
@@ -408,6 +408,10 @@ dependencies {
 
     // Zxing barcode reader integration
     implementation 'com.google.zxing:android-integration:3.3.0'
+
+    // MarkDown view
+    def markwonVersion = '4.6.2';
```
fm-sys, line 413: Typo: markwon -> markdown

fm-sys: Sorry. Just realized that the library is called "markwon"

moving-bits: actually no, in this case, as the library calls itself "markwon"

moving-bits:
> @moving-bits Did you see the suggestion where to place them from @fm-sys above?

Yes, have seen that, but didn't work for me - which may be due to the fact that I had not switched my emulator to a language different than English at that time...

But anyway, putting them into the `raw`, `raw-de`, `raw-zh-rTW` folder hierarchy has the advantage, that I can address the resource in the source using `R.raw.changelog_master` etc. - there is no `R.asset.changelog_master` or similar.

> I can take care of enabling on Crowdin later on. Just tell me, where the translated files need to end up.

We have `values`, `values-de` etc. currently, and for the changelogs this will be a similar `raw`, `raw-de` etc.

Can we limit Crowdin to specific files, `changelog_master.md` and `changelog_release.md` in this case?
(As we have other files in the `raw` folder which should not get translated at all.)

Lineflyer:
> Can we limit Crowdin to specific files, `changelog_master.md` and `changelog_release.md` in this case?
> (As we have other files in the `raw` folder which should not get translated at all.)

Sure.

moving-bits:
And another question:
What needs to be provided as initial file for the other language? No file at all, an empty file, a file with the English texts?

With what language shall we start? `de` and `cs`? (Where do we have enough active translators, willing to try such an experiment?)

fm-sys:
>  What needs to be provided as initial file for the other language? No file at all, an empty file, a file with the English texts?

IIRC nothing has to be created manually inside the repo, Crowdin does this for us. We just have to adapt the configuration (which is independent of this PR).

Lineflyer:
You don't need to provide any translated files or directories for them. Should all be done by Crowdin.

> With what language shall we start? `de` and `cs`? (Where do we have enough active translators, willing to try such an experiment?)

IMHO its not possible to restrict the languages from Crowdin side. I think also in our code there is no need to restrict. If all goes well it will return english files for all locale not translated (yet).
If it all fails we can still revert on git :)

fm-sys:
> Yes, have seen that, but didn't work for me - which may be due to the fact that I had not switched my emulator to a language different than English at that time...
> 
> But anyway, putting them into the `raw`, `raw-de`, `raw-zh-rTW` folder hierarchy has the advantage, that I can address the resource in the source using `R.raw.changelog_master` etc. - there is no `R.asset.changelog_master` or similar.

Sorry. I had mixed up `raw` and `asset`. Nice to hear that it works! :)

moving-bits:
Ok, I've updated the PR to use `raw` as folder, and also tried to adapt the Crowdin config - @Lineflyer please cross-check.

Where can we config Codacy to ignore `.md` files in all folders `main/raw` and `main/raw-*`?

fm-sys:
> IMHO its not possible to restrict the languages from Crowdin side. I think also in our code there is no need to restrict. If all goes well it will return english files for all locale not translated (yet).

# Excluding Target Languages for Source Files

By default, the source files are available for translation into all target languages of the project. If some source files shouldn’t be translated into specific target languages, you can exclude them with the help of the parameter excluded_target_languages.

Configuration file example:
```
content_copy
"files": [
  {
    "source": "/resources/en/*.json",
    "translation": "/resources/%two_letters_code%/%original_file_name%",
    "excluded_target_languages": [
      "uk",
      "fr"
    ]
  }
]
```

https://support.crowdin.com/configuration-file/#excluding-target-languages-for-source-files

Lineflyer:
That is IMHO only possible if you use the "direct" Crowdin interface (meaning, no let Github handle it). 
Anyway lets start with all languages.

moving-bits:
Playing a bit with a configuration file for Codacy, trying to ignore all `.md` files in `raw` and `raw-*` folders. Let's see whether that works...

moving-bits:
Yeah, Codacy seems to be happy with its new config file ;-)

So from now on we can do Codacy config using the `.codacy.yml` file in our root folder.

crowdin.yml
```diff
@@ -13,3 +13,9 @@ files:
   - source: /main/project/playstore/description/_source/*.txt
     translation: /main/project/playstore/description/translated/%language%/%original_file_name%
     content_segmentation: 1
+  - source: /main/res/raw/changelog_master.xml
+    translation: /main/res/raw-%two_letters_code%/%original_file_name%
+    content_segmentation: 1
+  - source: /main/res/raw/changelog_release.xml
+    translation: /main/res/raw-%two_letters_code%/%original_file_name%
```
bekuno, line 20: They are now *.md instead of *.xml, isn´t it?

moving-bits: Ah, good catch, thanks. Will adapt the PR.

main/res/raw/changelog_master.md
```diff
@@ -0,0 +1,13 @@
+## Nightly build:
+
+### Map:
```
Lineflyer, line 3: But I would follow Codacys suggestion to remove the `:` behind headings

moving-bits: Ah, forgot that. Have updated PR accordingly...

Lineflyer (commented):
Looks good. Also Crowdin configuration looks fine.

moving-bits:
Merged now to `master`. Be aware:
- From now on all changes to `changelog_master` need to be made in the `raw/changelog_master.md` file in markdown syntax.
- Same applies for the release changelog on `master` branch, see `raw/changelog_release.md` file.
- Caution: To avoid merge conflicts on every merge from `release` to `master`, the file `values/changelog_release.xml` is still present. - But it does not get evaluated in this version any longer. On final migration to the md files after releasing the next version, the contents need to be synced with `raw/changelog_release.md`, and then `values/changelog_release.xml` can be deleted.

moving-bits:
BTW: With a link to [Changelog for nightly version](https://github.com/cgeo/cgeo/blob/master/main/res/raw/changelog_master.md) we can show our users the current changelog on `master` branch...

Lineflyer:
Crowdin looks good so far. It handles each line as string.
I translated and approved some german strings, lets see what comes back.

moving-bits:
Thanks for checking. Looking forward to it - having translated changelogs will make handling for many users much easier.

While we are at it, we can think about whether we still want to provide the latest changes only with each new version of c:geo, whether we want a complete ("endless") changelog, or something in between. 

With the current amount of changes we have in c:geo (very much appreciated, BTW!) I guess it's sometimes hard for the non-hardcore users to recognize and understand each change within the time frame to the next version, so maybe we should keep the changelog for two or three last feature releases (and all bugfix releases in between). Handling of this should be pretty easy with the new mechanism.

fm-sys:
Maybe we can also shorten the changelog. E.g. only list new large features, major changes and urgent fixes (but leave our all minor things). That would save much time for our translators and makes it more readable...

Lineflyer:
> Maybe we can also shorten the changelog. E.g. only list new large features, major changes and urgent fixes (but leave our all minor things). That would save much time for our translators and makes it more readable...

Yes, I do see two improvements possible:
- Stick to user understandable terminology (e.g. avoid abbreviations like SAF, UDC)
- Only list the effective change (You can now do XYZ / Tapping on XYZ does now work) but not technical details

There is still a button shown below the changelog "List of all changes" currently targets the release tags page on Github. This contains only a copy of whats in the changelog, which makes no sense.
It should then maybe retargeted to show the commit history of branch `release`?! WIll open a dedicated issue for that.

moving-bits:
I fully agree with the two "rules" mentioned by @Lineflyer (but think we follow them mostly already), and understand the wish to shorten release notes to make it easier for the users to see the most relevant changes "at a glance". (Also this would make live a bit easier for our translators.) Still I think that having a "full" documentation of the user-relevant changes is a good idea. A link to our commit history is way too technical for that, but a separate, non-translated, file could fulfill that, see https://github.com/cgeo/cgeo/issues/10662#issuecomment-842588257 for a proposal.

PR #10641, opened 2021-05-16
force show internal storage on SAF folder selection (fix #10639)

### LINKED ISSUE #10639: Android 5 users with SAF problems
## Bug description 

There are still some users on support mail (up to now all Android 5) which claim, that they are not able to select a folder as they cannot see any content in the dir selector presented to them.

## c:geo version 

2021.04.25

## Screenshots 

image
image

## Additional context 

I found this link for that topic:
https://stackoverflow.com/questions/41913210/action-open-document-tree-shows-empty-recent-list

It claims, that adding an extra to the intent could solve it:
```
Intent intent = new Intent("android.intent.action.OPEN_DOCUMENT_TREE");
intent.putExtra("android.content.extra.SHOW_ADVANCED", true);
startActivityForResult(intent, REQUEST_CODE);
```

fm-sys:
Although this seems kind of hacky, it actually works... ;-)

Furthermore, the usage of it should be uncritical. The worst thing which could happen is that the EXTRA is simply ignored is new/some android versions.

### DIFF
--- a/main/src/cgeo/geocaching/Intents.java
+++ b/main/src/cgeo/geocaching/Intents.java
@@ -94,6 +94,18 @@ public class Intents {
     private static final String PREFIX_INTERNAL = "cgeo.geocaching.intent.internal.";
     public static final String INTENT_CACHE_CHANGED = PREFIX_INTERNAL + "cache-changed";
 
+    /**
+     * To be used together with {@link android.content.Intent#ACTION_OPEN_DOCUMENT_TREE}
+     *
+     * The value is decide whether to show advance mode or not.
+     * If the value is true, the local/device storage root must be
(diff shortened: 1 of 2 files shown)


eddiemuc (approved):

Lineflyer:
> Although this seems kind of hacky, it actually works... ;-) 

How have you been able to test the difference?

fm-sys:
I checked "hide internal storage" > internal storage was hidden
then I compiled the my change > internal storage is visible for c:geo (although not for other apps)

So at least this simplifys the migration work flow a lot for average users. But there's unfortunately no guaranty that this helps for the mentioned android 5 device...

PR #10643, opened 2021-05-16
Theming cleanup & first step toward migration to MaterialComponents theming

moving-bits:
## Description
For easier discussion of several ideas around new theming I'll upload my current approach with day/night mode-related theming and resource management. Does work generally so far, but only for activities derived from `AppCompatActivity`. Which currently means, that it does not work for our preferences :-(


fm-sys:
Looks excellent so far! This will make things much easier, really looking forward :)

main/res/values/themes.xml
```diff
@@ -54,54 +50,51 @@
...
-       <item name="subtitleTextStyle">@style/TextAppearance.AppCompat.Widget.ActionMode.Subtitle</item>
-
-       <item name="android:buttonStyle">@style/button</item>
-       <item name="android:editTextStyle">@style/edittext</item>
-       <item name="android:windowContentOverlay">@null</item>
-       <item name="actionBarStyle">@style/cgeo.ActionBarStyle</item>
-
-        <item name="colorAccent">@color/colorAccent</item>
-
+    <style name="cgeo.base" parent="@style/Theme.AppCompat.DayNight.DarkActionBar">
```
fm-sys, line 53: Does it make an optical difference if you use the even more modern `Theme.MaterialComponents.DayNight` (AFAIK based on `AppCompat`) instead? Cause https://material.io/ provides some really nice additional features and would help c:geo to become more attractive in the future...

moving-bits: Haven't tried yet. Will it work even for old Android versions (API 20+)? I thought that is what `AppCompat` is meant for.

moving-bits:
rebased to current master and adapted the new `*_material` themes coming from the new filter implementation

main/res/values/themes.xml
fm-sys: Yes, `MaterialComponents` is in fact just a better styled variant of `AppCompat`. But it might also look weird in the c:geo UI. But it's worth to try...

eddiemuc: For some of the Material.io components, their themes are necessary since they add some new, requires attributes. 

For the case of apps wanting to stay with the layout base of appcompat, Material.io offers bridge themes though. Can we use one of those?

Those would be of form:
`Theme.MaterialComponents.DayNight.[...].Bridge`, e.g. in this example `Theme.MaterialComponents.DayNight.DarkActionBar.Bridge`

moving-bits: Haven't seen those themes "in action" yet, but I guess we can try to use an up-to-date look. I'm worrying about technical backward-compatibility - which of the Material themes will run correctly under Android 5? Need to read more about it yet...

My next step would then be to merge the approaches from @eddiemuc used for the new filter and what I have done so far in this PR, and try to get things into master for further collaborate development. Then we still have settings using the old approach, but we can migrate them later, as soon as we have migrated them to `PreferenceFragment`.

fm-sys: from https://material.io/develop/android/docs/getting-started:
> Doing an app-wide migration by changing your app theme to inherit from a Material Components theme is the recommended approach. However, be sure to test thoroughly afterwards, as components in existing layouts may change their looks and behavior.
---
>### Ensure you are using AppCompatActivity
>Using AppCompatActivity will ensure that all the components work correctly. If you are unable to extend from AppCompatActivity, update your activities to use AppCompatDelegate. This will enable the AppCompat versions of components to be inflated among other important things.
---

> I'm worrying about technical backward-compatibility - which of the Material themes will run correctly under Android 5?

The Material themes are using the whole AppCompat stuff. There are towsends of Android Apps already using googles material design themes out there, and many of them are even still supporting Android 4 or even lower... So backward-compatibility is definitely no problem...

---
>The following is the list of Material Components themes you can use to get the latest component styles and theme-level attributes.
>
> -    Theme.MaterialComponents
>-    Theme.MaterialComponents.NoActionBar
>-    Theme.MaterialComponents.Light
>-    Theme.MaterialComponents.Light.NoActionBar
>-    Theme.MaterialComponents.Light.DarkActionBar
>
>  [And one of these `DayNight` themes we should choose:]
>-    Theme.MaterialComponents.DayNight
>-    Theme.MaterialComponents.DayNight.NoActionBar
>-    Theme.MaterialComponents.DayNight.DarkActionBar

fm-sys:
Some strange effects I noticed while testing this PR:

Material (purple):
grafik

Both, material and Appcompat (orange header line in all activities):
grafik

Both, material and Appcompat as well (light background, although I'm using dark theme):
grafik


But: The theming still looks nearly the same.

main/res/values/themes.xml
moving-bits: Thanks @fm-sys and @eddiemuc for the additional insights. I'll continue with updating the base class to Material based theme.

moving-bits:
> Some strange effects I noticed while testing this PR:
> 
> Material (purple):
> [...]
> 
> Both, material and Appcompat (orange header line in all activities):
> [...]

probably the accentColor?

> Both, material and Appcompat as well (light background, although I'm using dark theme):
> [...]

Yeah, it's far from finished. So far I've mostly tried to ease resource loading and theme switching, haven't changed anything relevant to the theme itself.

> But: The theming still looks nearly the same.

Probably we can (and have to) remove much of the individual styling we use so far, to get material design shine through, e. g. we use our own backgrounds etc.

main/res/values/themes.xml
```diff
@@ -54,54 +50,51 @@
...
-        <item name="button_color_enabled">@color/button_enabled</item>
-        <item name="button_color_disabled">@color/button_disabled</item>
-   </style>
-
-    <!-- Identical to cgeo.base aside from different parent style and bug fixes-->
-    <style name="cgeo.base.light" parent="@style/Theme.AppCompat.Light.DarkActionBar">
+        <item name="text_color">@color/colorPrimary</item>
+        <item name="text_color_grey">@color/colorPrimaryDark</item>
+        <item name="text_color_headline">@color/colorHeadline</item>
+        <item name="text_color_hint">@color/colorPrimaryHint</item>
```
fm-sys, line 58: If using the DayNight theme, we should overwrite the default attributes like: 
```
<item name="colorPrimaryHint">@color/colorPrimaryHint</item>
```

Otherwise, this results in strange effects as purple buttons (with material design), as all "non-themed" use the default attributes `colorPrimary`, `colorPrimaryHint`, ...

Our "old" attributes like "?text_color" aren't needed anymore at all, as we can/should now use `@color/colorPrimary` or `?colorPrimary` directly in the layout files.

fm-sys:
> > (orange header line in all activities):
> > [...]
> 
> probably the accentColor?

It's our costum ActionBarStyle which was only be used in MainActivity before...

grafik

main/res/values/themes.xml
moving-bits: Hm, I'm having a hard time to understand the respective purposes of the different color names. https://material.io/blog/migrate-android-material-components provides a nice overview of the colors that come with material, see the two images about halfway down. If I'm interpreting this correctly, then `colorOnPrimary` would be the text color to display the text on a background colored with `colorPrimary`, and similar for `colorOnSecondary`/ `colorSecondary` - is that understanding correct? (That would be the opposite of what I'd used in my PR so far, but no harm done yet.)

fm-sys: I'm also confused by these names. I would vote for keeping the naming of the resources understandable (`@color/text_color` etc.) But overwrite the theme default attributes `colorOnPrimary` etc... (as soon as we managed to understand what they are used for ;-))

fm-sys: AFAIK, `colorPrimary` and `colorSecondary` are two different kinds of accent colors... (IMHO both orange in our case)

grafik

moving-bits: Maybe like this:
- colorPrimary as background for ActionBar, colorOnPrimary for header text
- colorSecondary as accent color
- colorBackground as background for content (cache lists etc.), colorOnBackground as what text_color is today (default text for primary content)
- for other colors we are currently using (e. g. hint color, dimmed text etc.) we can use our own variables, as there are no pendants in the default color scheme
- some other colors are unused currently (colorError, colorSurface and their respective `on` text colors; colorPrimaryVariant, colorSecondaryVariant, colorOnSecondary)

fm-sys: I can tell you that `colorPrimary` partly is used as an accent color as well. Here is an example of an simple material app I wrote myself:

(notice the two different accent colors - I used the raw material theme without any modifications) 

Screenshot_20210517_145801_com fmsys settlements_copy_270x585

moving-bits:
Ok, as a next step I removed nearly all (!) themes & color attributes and started with a fresh `<style name="cgeo" parent="Theme.MaterialComponents.DayNight"></style>`. After some tweaking in the code it's actually working already, although visually not yet fully functional. More to come...

moving-bits:
Removed all text sizes and all color individualization for a clean start meanwhile.

Got the popup working with transparent view of the underlying activity (had that one hidden by opaqueness before), but still cannot make the splash screen transparent, always get the light grey color (or other opaque variants, depending on different settings in the code).

As the button style seems to be slightly rounded, I adapted our "icon buttons" accordingly.

image

moving-bits:
To give you a visual impression of the current state here are some screenshots:

## light theme
|main screen|map|quick settings|cache popup|
|---|---|---|---|
|image|image|image|image|
|cache details|waypoints|cache list|list filter|
|image|image|image|image|


## dark theme
|main screen|map|quick settings|cache popup|
|---|---|---|---|
|image|image|image|image|
|cache details|waypoints|cache list|list filter|
|image|image|image|image|

This is using a bare minimum `themes.xml` file, and a `styles.xml` stripped off most of its colors and textSize infos.

I'm coming close to a point where I would propose to merge this into `master` to be able for collaborate testing and improving things.

fm-sys:
I'm not sure if I like the orange action bar in light mode. But OTOH the old dark action bar was not very suitable for light mode as well - probably just a matter of habit. At least it makes c:geo a bit more colorful...

Furthermore, the spacing between the checkboxes in the quick settings dialog seem to be a bit too large, thus consuming much space. 

I can't really estimate how much more work is still needed. On the one hand, a merge would make collaboration much easier, but on the other hand, I would like to see our new filtering system released first, before we start making even more such extensive changes.

eddiemuc:
> I would like to see our new filtering system released first, before we start making even more such extensive changes.

I would love to get the new simplified themes/styles into the system as soon as possible. Because I assume it would allow me to remove some hackish workarounds in the cache filter activity (e.g. them themes "dark_material" and "light_material").

Maybe we can set for now the "bigger" changes to existing values (like the mentioned action bar color) until this is discussed in detail?

moving-bits:
> Furthermore, the spacing between the checkboxes in the quick settings dialog seem to be a bit too large, thus consuming much space.

Maybe we can use the nice two-column layout @eddiemuc introduced today for the cache type filter. That would fit nicely for the purpose of the quick settings and at the same time save quite some space.

Arrangement could be:
|left column|right column|
|---|---|
|own/found|original|
|disabled|parking|
|archived|visited|
|GPX track/route|circles|

Alas I have no idea how to implement that in the quick map settings...

> I can't really estimate how much more work is still needed. 

Despite CI failing with an error I cannot reproduce locally (the attribute mentioned in the error message is no longer in use in the mentioned layout file) the current state does work fine over here. It needs some visual polishing, for sure, but haven't found any major glitches so far.

Since we have some way to go with the new filtering AFAIK, and since we have just released I guess we do have some time for testing and collaboration, so I guess merging today or in a couple of days should be ok.

moving-bits:
> I would love to get the new simplified themes/styles into the system as soon as possible. Because I assume it would allow me to remove some hackish workarounds in the cache filter activity (e.g. them themes "dark_material" and "light_material").

The new `themes.xml` has only a handful of themes:
- `cgeo` is the base theme for most of the app
- `cgeo.popup` is used by the cache and waypoint "popup" (to have a transparent view to the underlying activity; it's based on `cgeo`)
- `settings` and `settings.light` are temporary themes used for settings until we have migrated settings to `AppCompatActivity` and `PreferenceFragments`
- `SplashScreenTheme` is a minimal theme for the splash screen, based on a `NoActionBar` theme

Most of the color and background definitions are gone, and (except for settings) the drawables, which differ between light and dark mode, have been renamed and moved to `drawables` and `drawables-night` respectively. They can be addressed directly now using `R.drawable.xxx` instead using attributes.
The no longer used background drawables need to be removed as soon we have agreed upon that we don't need them any longer.

The two themes for settings are already based on `cgeo`, but some of the inherent functionality does not work with a non `AppCompatActivity` activity, so there are some workarounds for that.

Text colors have been either removed or based on the new material scheme (except for settings).

Popup actionbar coloring is still a bit hacky, does not seem to follow the theme, so it's hard-coded. Need to check dependencies in pursuit of the reason.

Text sizes have been removed mostly to fall back to the defaults for now. We should define constants later on for regular text, small text, heading text etc. to get a consistent sizing. (In our current app we have very creative regarding sizes...)

Probably more to it, which doesn't come to my mind now ;-)

eddiemuc:
> > Furthermore, the spacing between the checkboxes in the quick settings dialog seem to be a bit too large, thus consuming much space.
> 
> Maybe we can use the nice two-column layout @eddiemuc introduced today for the cache type filter. That would fit nicely for the purpose of the quick settings and at the same time save quite some space.

While this would only be fair (because I copied the layout for a single checkbox from the map quick settings :-)), I feel that there is an aspect you won't like with the two-column-approach either: if you switch orientation then checkboxes are again far apart from text:

image

@moving-bits if you're interested in adapting it, I started/collected a new helper class ViewUtils to help with dynamic view construction (which is needed a lot in filter context!), two-column-layout is using these helpers heavily, maybe this can be even more abstracted.

But before going too deep into this: I would really like a "common" checkbox layout throughout c:geo. Currently this layout is "icon", "text", "checkbox" (right-aligned). If this is not wanted we should change it once for everywhere.

fm-sys:
> Alas I have no idea how to implement that in the quick map settings...

Technically, the new bottom dialog is not much different than a normal alert dialog. Just set the content view and show it. No magic you have to be aware of...

moving-bits:
> While this would only be fair (because I copied the layout for a single checkbox from the map quick settings :-)), 

That's why the layout file looked so familiar ;-)

> I feel that there is an aspect you won't like with the two-column-approach either: if you switch orientation then checkboxes are again far apart from text:

Still better than in a one column layout...

> @moving-bits if you're interested in adapting it, I started/collected a new helper class ViewUtils to help with dynamic view construction

Thanks, had seen that, but need to look into it more thoroughly.

> But before going too deep into this: I would really like a "common" checkbox layout throughout c:geo. Currently this layout is "icon", "text", "checkbox" (right-aligned). If this is not wanted we should change it once for everywhere.

It may not be perfect, but that's the best variant we have come up so far.
(Putting the checkbox to the left, followed by icon and text looks a bit strange to me (with the icons being pretty dominant compared to the checkbox, due to their size), and putting the checkbox to the left, followed by text and then the icon would either look clumsy (flattering edge) or unrelated (if icons are right-aligned).)

moving-bits:
rebased to current master and fixed a few issues

Lineflyer:
My few cents regarding the screenshots above:
I have to see it in live to judge it. However I like the uniform style, more use of color accents and clearer dark/light consistency.
Not sure though if I still like our orange accent as it looks rather bright.
Also not sure if we should stick with the pure black or better use the opportunity to go to a dark dark grey instead.

I will try to build a dev build form this PR and take a look.

BTW:
Many recent apps adapt their theme according to device setting (for devices having the feature for "black mode" (or however it is called officially). Those apps typically offer to either follow system setting or fix to dark or light. That would be desireable for c:geo as well IMHO.

eddiemuc:
Would it be an option to let the user select his own accent color from a set of choices? E.g. orange or blue?

Lineflyer:
> Would it be an option to let the user select his own accent color from a set of choices? E.g. orange or blue?

I don't think it would be wise to do something like that. Its also like our identify somehow...

Regarding the background:
I just checked several system dialogs and Google apps with dark theme on my Google Pixel 3a (running Android 12 beta).
None of them use pure black but several shades of grey ;))

Going from black to some grey would also IMHO reduce the relative brightness of the color accent.

See some examples:
Device settings:
image

Google Mail first start wizard:
image

Pull down quick menu:
image


Google Photos search dialog:
image

Lineflyer:
I just checked out a build from this PR and I really appreciate your work. Looks already promising.
- As elaborated above I would favor a dark grey instead of the deep black to make it look smoother and reduce hard contrasts.
- The tap boxes (like "Add a waypoint") for sure need some optimization. Maybe either another text color or maybe it will also be OK if the background has is no longer such dark.

A remark (knowing its WIP):
At some places the font size has increased, which is good. However the font size in cache lists (at least on my Pixel) got way smaller.

Lineflyer:
One more idea (was just mentioned by a friend):
To avoid those large color accent buttons it could also be buttons using the color accent as border only (border inset with a bit gradient to fade out).

moving-bits:
Thanks for your feedback.

Regarding the accent color: I've set both `colorPrimary` and `colorSecondary` to our accentColor, including its variants `colorPrimaryVariant` and `colorSecondaryVariant` - mainly for testing to see where it pops up, but maybe that's too much orange... We can start varying that and see how it will look like then.

We can also reduce the brightness of our orange a bit to reduce the high contrast.

With regard to the "background black": I've seen both type of apps, those having a dark grey and those having a pure black. And even my system settings go to a pure black when set to night mode. I guess it depends on the individual developer's personal preference...

But let's give it a try. Next things I will start working on is changing the background color to some dark grey, and introducing text sizes (using parametrized values) for a consistent look & feel.

fm-sys:
> Those apps typically offer to either follow system setting or fix to dark or light. That would be desireable for c:geo as well IMHO.

Yes. That's easy to archive. I offered to implement this in some other thread. But I will open an explicit issue for it...

moving-bits:
I've played a bit with the colors, using a more greyish background in dark mode, and using variants for our accent color:

|original accent|less saturation|less value|
|---|---|---
|image|image|image|
|image|image|image|

Personally I'm tending to the "less saturation" variants.

fm-sys:
Looking at the screenshots the "less saturation" variant looks kind of faded. Interesting, I would have said anything except "less saturation".

I would keep the color accent as is in dark mode, but make it a bit darker for light mode (As there is not a single color which fits perfectly for light at the same time as dark):

- The current accent color is optimized for dark -> keep it

- light mode -> Too few contrast, make it a bit (but not much) darker

moving-bits:
rebased + a couple of text sizes included (not yet finished) + some bug fixes

Lineflyer:
@moving-bits 
My feedback:
- I think the "less value" example looks best judging by the screenshots for both white and black. IMHO that accent could then be used for both themes instead of tweaking one of them as suggested by @fm-sys 
- I think the grey for the dark theme is a bit to bright, could be darker to look less like a "day weather day sky" ;) Maybe swap the colors of background and title bar?

fm-sys:
I think the lass value looks kind of dirty and more like a light brown tone. I would like to keep the bright orange (IMHO it's suitable for both themes - as currently ;-)). BTW material.io explicitly suggested to use a bright and standing out color. 

But 3 people, 3 opinions 😅

Be aware that changing our color accent again would mean that we have to remake the c:geo icon and playstore images if we want to look consistent, which would be _very_ much work. Ok, maybe we have to redo the playstore images in the near future anyway caused by the new theming, but still bright colors are modern (currently). 

Another thing: The coloring also strongly influences how users perceive an app. 
Mat, faded, dark colors -> Old and stale 
Bright colors -> Modern and well maintained 

To make a long story short: I see no reason why we should change the rated accent color and I ask you very much to keep the current accent!

Lineflyer:
> But 3 people, 3 opinions 😅

Yes....its difficult. I could also favor to stay with the accent as it is. Then it would be the question (as posted some comments above) if we could style the buttons in a different way (e.g. text in accent color or only border in accent color, etc, etc.).

fm-sys:
> I could also favor to stay with the accent as it is.

Cool! 

> Then it would be the question (as posted some comments above) if we could style the buttons in a different way (e.g. text in accent color or only border in accent color, etc, etc.).

We even should (based on material). There are four different button groups:

grafik

1. **Text button** (low emphasis)
Text buttons are typically used for less important actions.

2. **Outlined Button** (medium emphasis)
Outlined buttons are used for more emphasis than text buttons due to the stroke.

3. **Contained button** (high emphasis)
Contained buttons have more emphasis, as they use a color fill and shadow.

4. **Toggle button**
Toggle buttons group a set of actions using layout and spacing. They’re used less often than other button types.

fm-sys:
Full reference: https://material.io/components/buttons

BTW, if we need to do some design decisions, we should always take a look at material.io as they have very informative and well researched guides...

Lineflyer:
Looks like button 1 and 2 are exactly what I thought of...not sure though how these look with our accent on dark background.

fm-sys:
@moving-bits

Implementation:

1. Text button ` style="@style/Widget.MaterialComponents.Button.TextButton"`
2. Outlined Button `style="?attr/materialButtonOutlinedStyle"`
3. Contained button `default/no specific style`

I think the "Text button" is hard to recognize as button if used alone (-> only use if there is a medium/high emphasis button besides). So default would IMHO be "Outlined Button". For special things like in the wizard "login to Geocaching" definitely high emphasis should be used.

Lineflyer:
Seems like "Tabs" is what we could use to replace this viewpagerindicator:
https://material.io/components/tabs#usage

moving-bits:
IMHO this can be merged now as a first (large) step towards material theming, therefore removing "draft" status now.

For a more detailed description what's included in this PR, see #10738, and for open issues and further discussion please refer to the new [project for short-term UI issues](https://github.com/cgeo/cgeo/projects/20)

Lineflyer:
Shall we really merge to `master` or maybe develop on a dedicate branch on our repo?

This would give way to the new filter as a next feature release and allow more time afterwards?
Just thinking..not sure.

moving-bits:
> Shall we really merge to `master` or maybe develop on a dedicate branch on our repo?
> 
> This would give way to the new filter as a next feature release and allow more time afterwards?
> Just thinking..not sure.

I would prefer to see it in `master` rather soon to avoid still having to keep branches aligned for a longer period of time. (I'm doing this already for a couple of weeks, and there have been plenty merge conflicts along the way.)
IMHO now is a good time - we've just released a couple of days ago, so no pressure to release from `master` within the next few weeks, so plenty of time for playing around, testing and refining theming as well.

Also see https://github.com/cgeo/cgeo/pull/10643#issuecomment-846431165

Lineflyer:
I can't judge the amount of remaining work. If you consider it doable in a reasonable time of maybe 1-2 months, than its really better to merge to `master`.
This will also get some drive into the bugfixing then as bugs are easier to spot.

eddiemuc:
I vote for merging the theme simplification to master soon. Can barely wait for it to be honest :-)

Don't wait for filter framework to be finished before doing it. In fact, I expect to go into at least one release with filter still marked as "experimental". Also, I would like to fix some "layout hacks" in filtering after new theme scheme is merged.

Many thanks to @moving-bits for taking this big step towards a better UI design!

moving-bits:
> I can't judge the amount of remaining work.

It depends what you regard as "remaining work" ;-)

PR #10643 gives you a fully functional version, there are only a handful of visual glitches, and not everything is already "state of the art material theming like". But there are now "show breaker" AFAICS.
Some of the open things are inconsistencies we have for years already.

Merging to `master` would enable us to continue collaborately, and also make live easier (a bit, at least) for some inner workings of the new filter system.

Lineflyer:
As all are in favor lets merge that and continue from there.
If time permits I will play around with the nightlies in the upcoming weeks and provide feedback.

PR #10645, opened 2021-05-16
Fix copy coordinate via long click on waypoint coordinates (fix #6711)

### LINKED ISSUE #6711: Copy WP coordinates in selectable format
On the main cache page, you can cycle through various formats by tapping on the coordinate 
 field.
A long press then gives you the option to copy the coordinates, in the format selected, to the clipboard.

On the Waypoints page, tapping on the coordinates cycles through the various formats but a long press does nothing.
Could the copy function be added there as well please?

murggel:
## Description

Fixes not working long click on waypoint coordinates in waypoint list
added the callback of the context-menu functions as for cache-coordinates

## Related issues

fix #6711


murggel:
Ready for review

PR #10646, opened 2021-05-16
fix #8620: filtering for attributes for oc caches

### LINKED ISSUE #8620: Filtering attributes for opencaching.de caches does not work
**Describe the bug:**
Filtering for caches stored from opencaching.de does not give the expected results in stored caches list.

**To Reproduce:**
1. Go to a list of stored geocaches where a cache with OC-only attribute exists
2. Filter by attribute "Only loggable at Opencaching"

No cache is listed. I know to have at least one cache with OC only in that list (for example [OC11EB4](https://opencaching.de/OC11EB4)).

At least that known cache should be listed in the list after applying the filter.

**Version of c:geo used:**
2020.07.02

Yes

**Screenshots:**

**System information:**
[system-information-G8441.txt](https://github.com/cgeo/cgeo/files/4909265/system-information-G8441.txt)

**Additional context:**
I understand that [attribute filtering can only be applied to offline caches](https://manual.cgeo.org/en/mainmenu/lists#filtering). I used a list of stored caches in this issue.

Also I have tried other attributes as well in the same manner ("Ticks", "Hilly area") with the same result.

eddiemuc:
fix #8620: fixes filtering for attributes for oc caches for NEW filter framework

### DIFF
--- a/main/src/cgeo/geocaching/enumerations/CacheAttribute.java
+++ b/main/src/cgeo/geocaching/enumerations/CacheAttribute.java
@@ -14,6 +14,7 @@ import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
 
+@SuppressWarnings("PMD.FieldDeclarationsShouldBeAtStartOfClass") // static maps need to be initialized later in enums
 public enum CacheAttribute {
     // read project/attributes/readme.txt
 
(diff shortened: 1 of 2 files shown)


eddiemuc:
retest this please

PR #10651, opened 2021-05-16
Fix wrong default value for proximity-preferences (fix #10598)

### LINKED ISSUE #10598: Wrong default for maximum distance for calculating route
## Bug description 

Not sure, whether this was done intentional, I did however just realize, that the default routing distance in Settings - Routing is set to 259km / 160,94mi.
I was triggered by a beta user reporting unresponsiveness of the app when setting a target, which is quite far away. It was caused by the routing engine needing a lot of time to render the route.

- Install c:geo
- Run wizard
- Open Settings - Routing

Calculate route shows 259km

I believe the default value was set to something like 20km (?) before.
Maybe we could set it to 50km as a compromise. Longer maximum lengths might however cause delays on older devices.

## c:geo version 

2021.05.11-RC

## Reproducible 

Yes (tested on nightly)

murggel:
## Description

Fix wrong default value for proximity-preferences:
avoid doubled conversion of the value, `onGetDefaultValue` should return the default value, which is in this case the integer metric value (default value in the preferences.xml is defined as metric distance) and not the internal integer progress value.

## Related issues

#10598

### DIFF
--- a/main/src/cgeo/geocaching/settings/SeekbarPreference.java
+++ b/main/src/cgeo/geocaching/settings/SeekbarPreference.java
@@ -117,7 +117,7 @@ public class SeekbarPreference extends Preference {
 
     @Override
     protected Object onGetDefaultValue(final TypedArray a, final int index) {
-        return valueToProgress(a.getInt(index, defaultValue));
+        return a.getInt(index, defaultValue);
     }
 
     @Override


murggel:
@moving-bits  please check, if my assumption is correct, that the returned value should be the "real" value and not the progress.

murggel:
Hm, can it be, that #10598 is a bug only in master?
I just realized, that the PR for removing the check of min/max is was only for master. So the bug should not appear in release. It was reported for beta-version 2021.05.11-RC

I am a little bit confused ...

murggel:
Can I change the target to master or do I have to create a new PR?
Probably wrong version-info in #10598 (see https://github.com/cgeo/cgeo/issues/10598#issuecomment-841874896)

murggel:
replaced with new PR targeting master

Issue #8793, opened 2020-08-12
[Nightly] Change "Personal Note" popup to a (replacing) window


MagpieFourtyTwo:
**Is your feature request related to a problem?**
Since #8773 the Personal Note popup grows vertically, if you put more text into it. Great stuff, But although the additional editing space catapults working with PNs to a completely new level, the popup behaves weird in some scenarios:

1. while editing e. g. the first line and scrolling down until the selected line leaves the window, the cursor position will change to the *now* first line in sight (after scrolling). This sometimes makes it hard (or at lest annoying) to find the original editing position (i. e. prior to scrolling). Even more annoying: If you did not recognize the change of cursor position, just start to type ... the mess is ready. Thus would be *way* better, if the cursor position would just stay where it has been set manually, prior to scrolling.
2. while editing e. g. the last line on the first page of a multipage text, and then scroll up a few lines, the cursor will move down and then will show up *between the letters of the keyboard.* Looks pretty buggy, at least awkward.
3. when (keyboard still active) scrolling down more than a page (to be exact: until the cursor described in 2. crosses the lower border of the keyboard), the title row will move upwards, over the upper border of the display.
4. sometimes it's not possible to scroll anymore, as after some (not reproducible point) every swype action immediately selects text.

All these issues do not happen (AFAIK) in e. g. the log window. So: Would it be possible (with a reasonable amount of effort) to switch the Personal Notes popup to a (replacing) window, like the one for logs is?

This way perhaps we can even get additional space for buttons like ``Store waypoints`` or ``Store lists`` or whatever ...

Lineflyer:
I also recognized the problem of the headline of the personal note going out of range while editing when a lot of text is involved.
Couldn't we keep the headline on screen and scroll the content?

MagpieFourtyTwo:
Another problem which we stumbled upon this weekend: If the text is about a page long or even longer, and you've started to edit somewhere in the upper half of the page, it's pretty tricky to reach the last lines, cause these will now be covered by the keyboard.
Only chance: Close keyboard and select text somewhere at the end, but as soon as you go one lint to far up, the line below will be hidden and unreachable again. Or you grab the cursor (if it already shows the copy/paste indicator) and draw it down ... but eventually this immediately selects all the text ... high risk of loosing your note ... and - as far as I would say - nothing to give it to John Average ...

So, as there a too many problems, I would still suggest to forget about the popup and instead implement a dedicated window.

fm-sys:
[...]

Issue #10659, opened 2021-05-17
[new Filter] Remove Map Viewer Filter from Map-Quick-Settings


eddiemuc:
> I would definitely prefer to have the new filter in the map (overflow) menu

Yep, like the overflow menu idea, too - and in fact this was the second place where I tried to find it.

> [Indicator for a "Filter active" in Map View]
> You could simply reuse the existing "filter active" hint which is shown on map currently if a global cache type filter is active

Thx for the link! I knew I had seen it somewhere, but could remember where.  ;)
Is there a special reason why this has not been used for the "classic" filters so far?

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/10575#issuecomment-842247169_

eddiemuc:
@Lineflyer,  @MagpieFourtyTwo: one remaining question: at what position in the overview menu would you like to have the filter entry placed?

![image](https://user-images.githubusercontent.com/6909759/118485078-3ae33900-b718-11eb-9f10-7d1a145cb88d.png)

Lineflyer:
Good question:
BTW: The screenshot you posted is a rather special one for a map of a single cache.

Anyway:
I would place it rather prominent so that is has chances (on devices with larger screen or higher density) to be shown as icon. I would suggest behind "Enable/Disable live" with the option "Show if room"!?

Lineflyer:
Addition:
But it makes no sense if the map is opened for a single cache. Shouldn't the filtering be hidden in that case (unless live mode is enabled in addition)?

eddiemuc:
> Addition:
> But it makes no sense if the map is opened for a single cache. Shouldn't the filtering be hidden in that case (unless live mode is enabled in addition)?

Well yes. Although I am personally not a friend of such ever-changing menues.

But if this is so important: shouldn't then also the filtering options in the quick settings vanish on single-cache-view?

Lineflyer:
> But if this is so important: shouldn't then also the filtering options in the quick settings vanish on single-cache-view?

You are right...these also only make limited sense.
But then for the sake of simplicity: Lets not do any hiding of the menu entries just because they will probably not be useful. Just forget my addition :)

fm-sys:
I would place it directly after the map source selector or even before that...

Lineflyer:
> I would place it directly after the map source selector or even before that...

Maybe better behind..quite some users changing their maps frequently (Google Maps overview, OSM deep dive for small paths).

[...]

Issue #3932, opened 2014-05-31
Replace popups


Bananeweizen:
Looking over the existing issues, I feel like we currently get into a lot of trouble with the cache and waypoint popups. We have to emulate an action bar there, we have to re-implement a lot of the logic of its menu, and still there are some bugs or inconsistencies (when compared with the real actionbar).

Instead of trying to fix all that in the popups, I suggest we instead remove them. By that, I do not mean to remove the _content_ of the popups, but instead just the popup around the content. Basically take the layout contained in the popup and show it as a _non modal_ box at the top or bottom of the map. Look at Neon, GeOrg or the GC app to see what I mean.

The action bar menus could then either be related to the map (if no cache is selected) or to the selected cache/waypoint. You have all seen this already with the contextual action bar, when selecting something in an app. 

And that has multiple benefits:
- popup code can be removed
- users can switch between caches _without_ needing to press "back" in between
- we can expand that concept more easily to show more than 1 cache, e.g. by just stacking 3 of those boxes onto each other.

Comments please?

moving-bits:
we have switched cache/waypoint popups to non-modal bottom sheets meanwhile, so let's close here

PR #10674, opened 2021-05-17
refactor quick settings popup to BottomSheetDialog (fix #10663)

### LINKED ISSUE #10663: Make Map settings pop-up a BottomSheetDialogFragment based on Material Design Guidelines
Signal has the new possibility to select the image size, which makes use of a so-called "bottom sheet". They are very intuitive and easy to use, as they appear at the bottom of the screen and are therefore easy to reach...

Screenshot_20210517_151911

Something like this I would really like to see for our quick settings pop-up too. (But it could be, that BottomSheets are only available as soon as we use material design theme)

Some links:
- The related commit in the Signal repo: [signalapp/Signal-Android@dd934e0](https://github.com/signalapp/Signal-Android/commit/dd934e0095197b30d760a4a12441381606f0d39f#diff-deda9be468ee329c8ea37e626318c07af1f148c546537c06f5b090cf5c19f3f5)
- https://medium.com/@kosta.palash/using-bottomsheetdialogfragment-with-material-design-guideline-f9814c39b9fc
- A very extensive description: https://material.io/components/sheets-bottom

fm-sys:
See #10663 for more information...

| half screen | full screen |
|-|-|
| grafik|grafik|

This is codewise quick and thirty yet, but IMHO the result looks very intuitive and modern...

Setting WIP for now, as I plan to add some ~helper classes~ generalized methods for it, so that it can be easily reused (e.g. for #8793 - might be interesting for @MagpieFourtyTwo ;-))


Lineflyer:
Nice in general...just I am struggeling with the total black background (also in the new filter) instead of some lightend black (or dark grey). But thats a more general topic when it comes to refactoring our styles/themes.

fm-sys:
> just I am struggeling with the total black background (also in the new filter) instead of some lightend black (or dark grey).

I don't like this background as well, but I wanted to keep things unified. It's the same background as used for lists, cache details, ........ nearly everywhere. 


BTW, what do you think: Could this also be used for a more modern cache details popup at our live map or even for advanced filtering? Just thinking...

Lineflyer:
> BTW, what do you think: Could this also be used for a more modern cache details popup at our live map or even for advanced filtering? Just thinking...

Yes, it could be an option for a map popup replacement...see #3932

Lineflyer:
> or even for  advanced filtering?

Such functions however IMHO deserve a dedicated fullscreen.

fm-sys:
> > or even for  advanced filtering?
> 
> Such functions however IMHO deserve a dedicated fullscreen.

It is also possible to launch the BottomSheetDialog directly as full screen. 
However, they feel much more integrated. E.g. you can close them via a swipe down gesture, which is a large benefit especially on larger screen sizes. 

But this discussion is totally off topic, let's continue it somewhere else as soon as we have some testing experience with BottomSheetDialogs for Quicksettings and PN...

moving-bits:
Thanks @fm-sys for exploring new UI/UX features, very welcomed. I've gave it a quick tour in the emulator, and some questions come to my mind:
- What is the advantage of having a BottomSheetDialog? From what I understand it may be easier to close on larger displays as you can use swipe down instead of having to use the arrow in the title bar. (But you could still use the back button.) OTOH you'll have more "work" for opening the dialog, as you still have to press the button, and then do another swipe to fully open the dialog (which some users may not recognize, and therefore missing some of the options).
- In the past we had some discussion (and even a minimal prototype) for displaying cache info as a fragment at the bottom of the window, supporting swipe up to expand on more info. Would that clash with this dialog?
- Since now the full width is used, the checkboxes and their "labels" are pretty far apart. We may need to rearrange them to show more clearly which belongs to what.
- Unrelated to the ButtomSheetDialog: I would put the new filter button into the dialog's actionbar, currently it's using pretty much space.
- BTW: Dialog title does not work well with light theme. IIRC we already have opened an issue for that.

image

eddiemuc:
> Unrelated to the ButtomSheetDialog: I would put the new filter button into the dialog's actionbar, currently it's using pretty much space.

For this specific part, please see #10659

fm-sys:
> * What is the advantage of having a BottomSheetDialog? From what I understand it may be easier to close on larger displays as you can use swipe down instead of having to use the arrow in the title bar. (But you could still use the back button.) 

Remember the many support mails "I can't close the quick settings pop-up". We meanwhile have the back button in the pop-up, but this is more a rather ugly workaround. And to be honest, I sometimes struggle with closing the popup as well. 

Good UX is often just a unconscious feeling. "it works" vs. "It's fun to use the intuitive ui" 


>OTOH you'll have more "work" for opening the dialog, as you still have to press the button, and then do another swipe to fully open the dialog (which some users may not recognize, and therefore missing some of the options).

Our current popup is that large, that you have to scroll on most devices anyway. But users would never expect, that they can scroll in a alert dialog, while they are already firm with the BottomSheetDialog (e.g. From Google maps) 

Furthermore, I explicitly implemented a "you can swipe up"-indicator / action button on the left (unfortunately only visible in dark theme currently), which is the way how it is recommended by material.io (and they have done much very good research) 

> 
> * In the past we had some discussion (and even a minimal prototype) for displaying cache info as a fragment at the bottom of the window, supporting swipe up to expand on more info. Would that clash with this dialog?

That wouldn't be a problem, as long as we don't want to show them both at the same time. ;-) 
> 
> * Since now the full width is used, the checkboxes and their "labels" are pretty far apart. We may need to rearrange them to show more clearly which belongs to what.

Could be. Maybe we can enlarge the outer margins 
> 
> * Unrelated to the ButtomSheetDialog: I would put the new filter button into the dialog's actionbar, currently it's using pretty much space.
> 

This is IIRC strongly dissuaded by material. But the map activity has much room as well...
> * BTW: Dialog title does not work well with light theme. IIRC we already have opened an issue for that.

I will fix this...

MagpieFourtyTwo:
> Our current popup is that large, that you have to scroll on most devices anyway.

Hmm ... "most devices" I usually see have screens with 6 inches and more ... thus leaving plenty of room ...

fm-sys:
> 
> 
> > Our current popup is that large, that you have to scroll on most devices anyway.
> 
> Hmm ... "most devices" I usually see have screens with 6 inches and more ... thus leaving plenty of room ...

That's true, but for these devices the "more work (click + swipe) for opening the dialog" argument probably isn't true as well ;-)

fm-sys:
grafik

I'm now using the action bar color which we have at other places as well...

fm-sys:
> OTOH you'll have more "work" for opening the dialog, as you still have to press the button, and then do another swipe to fully open the dialog (which some users may not recognize, and therefore missing some of the options).

It's possible to launch directly in fullscreen mode as well, so we can change the behavior later on if required (after some testing in the nightlys)

MagpieFourtyTwo:
Looks really nice, in general, like it - apart from the longer ways, the fingers have to go now, these increased significantly:
In order to e. g. change the routing strategy, it was just three taps with really short distances between them before.
But now, in order to close the "popup", you either have to go to the upper end of the screen or make a swipe down,
which must be long enough (and it must be *really* long), and has perhaps even to be repeated, if it was to short ...

Question: Would it be possible to decrease the minimum swipe distance needed to trigger the "close event" ...?

Lineflyer:
You can also just click outside like with the old dialog.

MagpieFourtyTwo:
Yes, I know - but there is just one "outside section" left (the one I meant with "the upper end of the screen"), instead of four.
And sp. the "bottom section", which was clearly preferred by the thumb, is gone.

fm-sys:
> Yes, I know - but there is just one "outside section" left (the one I meant with "the upper end of the screen"), instead of four.
> And esp. the "bottom section", which was clearly preferred by the thumb, is gone.

The problem was that on some devices (e.g. my device) all these four sides were nearly totally obscured by the dialog, so closing via an outside click was nearly impossible. 

But I have struggled with closing the bottom dialog as well (feels a bit like if the dialog wantsq  to stay and always comes back up) 

> Would it be possible to decrease the minimum swipe distance needed to trigger the "close event" ...?

Even though I don't know about it yet, I firmly believe that nothing is impossible... :)

Issue #10661, opened 2021-05-17
Optimum solution to combine/seperate/remove old and new filters and the hierarchy of them


Lineflyer:
I am still thinking about the optimum solution to combine/seperate/remove old and new filters and the hierarchy between them.

My current idea would be:

- Have an instance of the new filter in lists which is effective on lists only (already implemented)
- Have an instance of the new filter on map which is effective on map only (already implemented)
- Also replace the current global cache type filter by an instance of the new filter methods which then sets both of the above mentioned instances to the filtering defined here

So various use cases could be covered:
- Global filter is still in main menu but offers more options which are applied to the whole app (as users are used to even today)
- Alternatively users can setup a filter for the map only or for lists only by (instead of using the global filter). 

Would that be technically feasible and understandable/meaningful for users?

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/8383#issuecomment-842277457_

eddiemuc:
For my personal taste the whole distinction of global vs non-global filters is already too complicated. I would simply stay with one filter saved per usage type (always the last one used for this usage type), usage types being the Cachelisttypes (map viewer, offline lists, online search of different types, map-as-list, maybe pocket queries or such)

If we really need a global filter in addition, I suggest to leave the current cache type filter as-is (maybe add multi-select). Person I would leave it out.

Lineflyer:
> If we really need a global filter in addition, I suggest to leave the current cache type filter as-is (maybe add multi-select). Person I would leave it out.

From my support experience most users would like to have a base filter set for multiple cache types and (ideally) a limitation on D and/or T. If we could realize that in the global filter menu, that would be great for average users not interested in more complex filterings.

The question remaining would then be:
How would an active global filter interact with the filters on lower hierarchy (map/lists)?

Lineflyer:
Suggestion for content of a potential global filter (combined with my suggestion in https://github.com/cgeo/cgeo/issues/8383#issuecomment-842324171):

- Apply to: [ ] Map&Search [ ]  Offline Lists
- Cache type (Multi select)
- D-Rating (slider from/to)
- T-Rating (slider from/to)

eddiemuc:
> The question remaining would then be:
> How would an active global filter interact with the filters on lower hierarchy (map/lists)?

Only one option would be feasible: both filters are applied connected by AND.

[...]

PR #10691, opened 2021-05-20
fix part of #10661: basic vs advanced filter mode

eddiemuc:
fix part of #10661: basic vs advanced filter mode

### DIFF
--- a/main/res/layout/cache_filter_activity.xml
+++ b/main/res/layout/cache_filter_activity.xml
@@ -15,8 +15,36 @@
         android:orientation="vertical" >
 
     <RelativeLayout
-        android:layout_width="fill_parent"
+        android:layout_width="match_parent"
         android:layout_height="wrap_content">
+
+        <cgeo.geocaching.ui.ToggleButtonGroup
+            android:id="@+id/filter_basic_advanced"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_alignParentLeft="true"
+            android:layout_marginHorizontal="6dp"
+            android:layout_marginVertical="3dp"
+            android:layout_centerVertical="true"
+            tools:ignore="RelativeOverlap" />
+
+        <ImageButton
+            android:id="@+id/filter_additem"
+            style="@style/button_icon"
+            android:layout_alignParentRight="true"
+            android:layout_centerVertical="true"
+            android:src="@drawable/ic_menu_add"/>
+
+    </RelativeLayout>
+
+   <View
+        style="@style/separator_horizontal"
+        android:layout_marginTop="1dp" />
+
+     <RelativeLayout
+         android:id="@+id/filter_storage_options"
+         android:layout_width="fill_parent"
+         android:layout_height="wrap_content">
 
         <ImageButton
             android:id="@+id/filter_storage_manage"
@@ -48,6 +76,7 @@
     </RelativeLayout>
 
     <View
+        android:id="@+id/filter_storage_options_line"
         style="@style/separator_horizontal"
         android:layout_marginTop="1dp" />
 
@@ -59,20 +88,8 @@
         android:layout_marginLeft="6dp"
         android:layout_marginRight="6dp" />
 
-    <RelativeLayout
-        android:layout_width="fill_parent"
-        android:layout_height="wrap_content">
-
-        <ImageButton
-            android:id="@+id/filter_additem"
-            style="@style/button_icon"
-            android:layout_alignParentRight="true"
-            android:src="@drawable/ic_menu_add" />
-
-    </RelativeLayout>
-
-
     <View
+        android:id="@+id/filter_props_checkboxes_line"
         style="@style/separator_horizontal"
         android:layout_marginTop="1dp" />
 
--- a/main/res/layout/checkbox_item.xml
+++ b/main/res/layout/checkbox_item.xml
@@ -2,7 +2,7 @@
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:orientation="horizontal"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
+    android:layout_height="wrap_content"
     android:paddingTop="4dp"
     android:paddingBottom="4dp">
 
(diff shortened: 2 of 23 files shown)


eddiemuc:
Implements a "basic vs advanced" switch at top of filter configuration activity. 

Since "cache type" filter is currently far too long, the "basic" filter doesn't really look good. This will be fixed in a separate PR.

fm-sys:
Made a quick test in the emulator, but I come up with a total empty screen at basic, and it keeps remain (nearly) empty if switching to advanced. 

grafik
grafik

But then, if an add a new filter D/T/Type pops up as well. 





grafik


I think D/T/Type should only be taken oven if it actually "filters" (only a specific range/type(s) are selected)

---

Don't know if it's related, but I get following stacktrace if I press "Go!" while screen is empty:
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: cgeo.geocaching, PID: 24284
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean android.widget.CheckBox.isChecked()' on a null object reference
        at cgeo.geocaching.filters.gui.CheckboxFilterViewHolder.createFilterFromView(CheckboxFilterViewHolder.java:125)
        at cgeo.geocaching.filters.gui.FilterViewHolderCreator.createFrom(FilterViewHolderCreator.java:104)
        at cgeo.geocaching.filters.gui.GeocacheFilterActivity.getFilterFromView(GeocacheFilterActivity.java:296)
        at cgeo.geocaching.filters.gui.GeocacheFilterActivity.finishWithResult(GeocacheFilterActivity.java:282)
        at cgeo.geocaching.filters.gui.GeocacheFilterActivity.onOptionsItemSelected(GeocacheFilterActivity.java:207)
        at android.app.Activity.onMenuItemSelected(Activity.java:4137)
        at androidx.fragment.app.FragmentActivity.onMenuItemSelected(FragmentActivity.java:384)
        at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected(AppCompatActivity.java:228)
        at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
        at androidx.appcompat.app.AppCompatDelegateImpl.onMenuItemSelected(AppCompatDelegateImpl.java:1176)
        at androidx.appcompat.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:834)
        at androidx.appcompat.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:158)
        at androidx.appcompat.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:985)
        at androidx.appcompat.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:975)
        at androidx.appcompat.widget.ActionMenuView.invokeItem(ActionMenuView.java:623)
        at androidx.appcompat.view.menu.ActionMenuItemView.onClick(ActionMenuItemView.java:151)
        at android.view.View.performClick(View.java:7125)
        at android.view.View.performClickInternal(View.java:7102)
        at android.view.View.access$3500(View.java:801)
        at android.view.View$PerformClick.run(View.java:27336)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

eddiemuc:
Suggested view for basic filter:
image

What do you think?

fm-sys:
> What do you think?

Looks good 👍

Will this type filter layout also be used for advanced mode?

Lineflyer:
Looks like a perfect approach.
Lets IMHO use it and see whether we find any optimizations later on...but that looks already quite straight forward and easy to use.

eddiemuc:
> Will this type filter layout also be used for advanced mode?

Yes it will.

Ok then I will start fixing up the code base behind it...

fm-sys:
@Lineflyer fo you agree that this is a suitable replacement for the current "global filter" (means we can remove/replace it)?

moving-bits:
> What do you think?

Looks really good, like it 👍 

Only one slight change proposed: I would put the checkbox for Select all/none directly beside the label:

image

Otherwise it looks (at least to me) like having a dangling checkbox belonging to nothing.


Or put it in the same line as the headline, something like this - saves one line:

image

Or make it a slider (like we have in the PQ activity)

Maybe only label it "Select all" instead of "Select all/none"?

fm-sys:
> Or make it a slider (like we have in the PQ activity)

I would IMHO stay with a ckeck box. To be honest, I've never seen a slider in any UI which activates all ckeckboxes below. In this use case, always a checkbox is used. 

But I agree, that all/none should be moved directly beside the label.

Lineflyer:
Yes, I think we should go for that also to replace the current global filter, meaning to keep the menu item but fill it with this basic/advanced filter.

eddiemuc:
Thanks for the feedback. Two comments:
* Moving "select all" in the headline would clash with the "advanced" view where in this place there are the delete and the drag button. If this is really wanted then I would ask for a separate issue because this is a lot of work on its own. I would for now place it aligned to the "right column" (and change text to"Select all" of course)
* Replacing global filter with this is the plan, but there's a lot of work to be done behind the scenes before this can really be carried out. Keep in mind that new filtering is still marked as experimental for good reason, and global type filter is very widely spread in c:geo code. Much coding and testing ahead of us.

fm-sys:
> I would for now place it aligned to the "right column"

Totally fine 👍

> Replacing global filter with this is the plan, but there's a lot of work to be done behind the scenes before this can really be carried out. Keep in mind that new filtering is still marked as experimental for good reason, and global type filter is very widely spread in c:geo code. 

Take your time. We're just already looking forward ;-) 
And thanks for your really great work!

eddiemuc:
One more thing wrt cache types:
as you can currently see when you open the cache type filter on master, we have 23 (!) cache types currently used in c:geo (see https://github.com/cgeo/cgeo/blob/master/main/src/cgeo/geocaching/enumerations/CacheType.java). In order not to overflow the filter, we have to "group" them together somehow. Also, layout-wise it makes sense to have an even number (if types are presented in two columns).

gc.com is doing the same, and using their logic plus some of my own (esp for the non-gc.com-types) I now implemented the following "groups" and "grouping rules":

image

Grouping Rules:
* "Traditional" will also find GCHQ and PROJECT_APE (same logic as on gc.com)
* "EVENT" will also find MEGA, GIGA, COMMUN_CELEBRATION, GCHQ_CELEBRATION, GPS_EXHIBIT, BLOCK_PARTY (same logic as on gc.com **except** that gc.com also finds CITO in this setting!)
* **Unlike** gc.com I created an OWN search field for CITO to make the displayed number of cache types even
* "Virtual" will also find "LOCATIONLESS"
* "User-Defined" will also find "Unknown" and "All" (but I expect/hope that the latter is never actually assigned to a cache)

Please let me know if you want different "groups" or "group rules". Alas, this can also easily be changed/adapted at a later point in time.

eddiemuc:
(Edited last comment wrt cache type "all")

fm-sys:
"Userdef" should IMHO be "Userdefined" / "User-defined". That's only four/five characters more and the length which fits into the screen differs for every device anyway. How should "def" be translated into different languages? I guess in other languages it will be way longer anyway so I hope the displayed string just gets shorted by "..." if necessary...

PR #10692, opened 2021-05-20
fix #10684: make csah handle persisted uris correctly, switch trackutils to right action type

### LINKED ISSUE #10684: Track/route cannot be updated
## Bug description
A loaded track/route file cannot be replaced by a different track/route without manually unloading it before anymore.

## Steps to reproduce the problem
- Open OSM map
- Load a GPX track/route file => track/route is being displayed
- Load a different GPX track/route file (with confirming overwrite)

There is a display "File xy selected", but still the first track/route is shown

The newly selected track/route should be shown

## c:geo version 
Both current state of `release` and `master` branches, haven't tested against current release 2021.04.25

## Reproducible 
Yes

## Additional Context
If I'm not mistaken this is a consequence of making the SAF selectors state aware, see #10447.

eddiemuc:
fix #10684: make csah handle persisted uris correctly, switch trackutils to right action type

### DIFF
--- a/main/src/cgeo/geocaching/storage/ContentStorageActivityHelper.java
+++ b/main/src/cgeo/geocaching/storage/ContentStorageActivityHelper.java
@@ -364,7 +364,9 @@ public class ContentStorageActivityHelper {
             }
         }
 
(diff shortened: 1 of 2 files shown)


eddiemuc:
Set to "Draft" to prevent accidental merge until `release `is free again

PR #10665, opened 2021-05-17
Fix #10662 - Link to commit history on release branch

Lineflyer:
## Description

Changes the link for "See all changes" from githubs release page to the github commit history for branch `release`

## Related issues

Fix #10662

PR #10709, opened 2021-05-22
Link about:changelog to new full changelog (fix #10662)

### LINKED ISSUE #10662: Target "List all changes" to commit history
There is still a button shown below the changelog "List of all changes", which currently targets the release tags page on Github. This contains only a copy of whats in the changelog, which makes no sense.

It should maybe targeted to show the commit history of branch `release` instead.

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/10640#issuecomment-842416417_

moving-bits:
## Description
Link to our new full changelog from about:changelog page in app

Superseeds #10665


Lineflyer:
@moving-bits Seen my comment above?
Shouldn't we handle it same as for the "normal" changelogs (regarding branch wise handling) just without translation?
Now `release` users clicking on that link will see changes from `master`

moving-bits:
The idea behind this was to have a unified changelog covering all changes. Distinguishing between `release` and `master` would have to be done by following the headlines "next release" vs. "nightly".

It's not really comparable to what we show within our app, as we can display two files / text blocks in one view there, whereas the link goes to a single page, which could be either "master" or "release", if we want to distinguish. To include all changes from `release` in the info page for `master` as well, one would have to care for that manually (in addition to updating the page for `release`) - something I would like to avoid.

PR #10710, opened 2021-05-22
fix filtering issues with display, grouping and cache count

eddiemuc:
fix filtering issues with display, grouping and cache count


eddiemuc:
This is an after-fix of things merged earlier. I will therefore merge it quickly before the other stuff comes into the nightly.

PR #10714, opened 2021-05-22
fix #7540: always refresh offline lists in cachelistactivity.onresume

### LINKED ISSUE #7540: Filtered list not updated on change
From support mail, not yet verified:

- Set global filter to one cache type
- Open list of saved caches
- Open cache details of a cache on list
- Delete cache from that list
- Return to filtered list

- Deleted Cache is still shown on list until list is closed/reopened

- Deleted cache should no longer been shown

##### Version of c:geo used:
2019.04.16

Yes, but only if cache type filter is active

eddiemuc:
fix #7540: always refresh offline lists in cachelistactivity.onresume


moving-bits:
retest this please

PR #10723, opened 2021-05-23
fix #10715, fix #10717, fix #10716: various filter-related small issues

### LINKED ISSUE #10659: [new Filter] Remove Map Viewer Filter from Map-Quick-Settings

### LINKED ISSUE #10715: Filter: Basic/Advanced selection lost on rotation
## Bug description 

The selection whether to use basic or advanced filtering is lost on device rotation

- Open new filter
- Switch from basic to advanced
- Rotate device to landscape

Filter is back to basic mode

Filter should stay in advanced mode

## c:geo version 

2021.05.22-NB

## Reproducible 

Yes

### LINKED ISSUE #10716: Use switch style instead of box style to toggle basic/advanced
Just a cosmetic question:
Shouldn't we use this nice switch style (as used in the PQ view to enable "Show all")?

So it would be just a disabled switch labeled "Advanced" instead of two boxes with "Basic" and "Advanced"

### LINKED ISSUE #10717: Use textual button for adding filter element
Just another thought regarding the new filtering. I showed the new dialog to a geocaching friend. Besides the fact, that he really liked it, he had one struggle while first trying out the advanced mode:

He did not realize the (+) button shown after switching to advanced mode. He asked me, why the advanced mode is just the same as the basic mode, just he can select "OR instead AND", etc.

Therefore just thinking if adding new filter element should deserve more prominence, as e.g. a screen wide text button "Add filter element".

eddiemuc:
fix #10715, fix #10717, fix #10716: various filter-related small issues

Also revises the already closed #10659: force-adds filter icon to map action bar and removes its text

### DIFF
--- a/main/res/layout/cache_filter_activity.xml
+++ b/main/res/layout/cache_filter_activity.xml
@@ -18,22 +18,26 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content">
 
-        <cgeo.geocaching.ui.ToggleButtonGroup
+        <androidx.appcompat.widget.SwitchCompat
             android:id="@+id/filter_basic_advanced"
             android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
+            android:layout_height="38dp"
             android:layout_alignParentLeft="true"
-            android:layout_marginHorizontal="6dp"
-            android:layout_marginVertical="3dp"
+            android:layout_margin="6dp"
             android:layout_centerVertical="true"
-            tools:ignore="RelativeOverlap" />
+            android:text="@string/cache_filter_mode_advanced" />
 
-        <ImageButton
+        <TextView
             android:id="@+id/filter_additem"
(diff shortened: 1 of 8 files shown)


eddiemuc:
Basic:
image

Advanced:
image

Lineflyer:
Thats pretty much exactly what I imagined :)

PR #10725, opened 2021-05-23
add messageMarkdown for formattable message alerts

moving-bits:
## Description
While traveling along my themes and formatting journey I've stumbled upon the new "filter (experimental)" info dialog. While the message is pretty informative, it should get some formatting, IMHO, so I added a `messageMarkdown()` method which allows for just that:

image

(As this might be a not so frequently used method, I've introduced just the plain version, not all variants that exist for the other dialog methods.)


moving-bits:
retest this please

PR #10733, opened 2021-05-24
fix #10672: add grouping to filters

### LINKED ISSUE #10672: Wording and sorting improvement proposals for filter element names
It took me a while to find out what some filter elements are used for as they have rather short names which could IMHO be improved.

Proposal:
- Hidden --> Hidden Date (in order not to be confused with hidden by)
- Name --> Cache Name (in order not to be confused with Owner)
- Description --> Cache Description (Don't know why, but it sounds not intuitive to me without cache in it)
- Type --> Cache Type (Don't know why, but it sounds not intuitive to me without cache in it)

Furthermore I think the alphabetic sorting of them is not the most user friendly to quickly find a specific element (if you don't know its exact name).
Maybe it would be useful trying to group them, e.g. (Group names TBD):
- Basic 
  - Cache Type
  - Cache Name
  - Cache Description
  - Owner
  - Hidden Date
  - Status
- Detailed 
  - Size
  - Difficulty
  - Terrain
  - Favorites
  - Attributes
- User specific
  - Personal Note
  - Offline Log
  - Distance

eddiemuc:
fix #10672: add grouping to filters

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -497,6 +497,11 @@
     <string name="cache_filter_distance">Distance</string>
     <string name="cache_filter_hidden">Hidden Date</string>
 
+    <string name="cache_filtergroup_basic">Basic</string>
+    <string name="cache_filtergroup_detailed">Detailed</string>
+    <string name="cache_filtergroup_userspecific">User specific</string>
+
+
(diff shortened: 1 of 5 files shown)


eddiemuc:
Groups are assigned 1:1 as suggested in #10672 for now

PR #10748, opened 2021-05-25
Fix bug in extracting user note for waypoint with formula

murggel:
## Description

Fix bug in extracting user note for waypoint with formula from personal note.
Following linebreak was not considered correctly and the next waypoint was extracted as user note for the waypoint


murggel:
ready for review

PR #10765, opened 2021-05-26
set darker background for dark mode (fix #10752)

### LINKED ISSUE #10752: Material theming: background color way too light
IMHO the new / current background color is way too light. While I agree that a pure black is too dark, the color should be a dark gray instead of light gray (as it was before at some places).

| Gmail | c:geo | pure black |
|-|-|-|
| grafik | grafik | grafik |

I would suggest using the "GMail" color theme (-> swap/adapt action bar and background color inside c:geo)

moving-bits:
## Description
Set the background color #FF202020 as suggested in https://github.com/cgeo/cgeo/issues/10752#issuecomment-848974727

Result would be like this:
image.image.image

And in conjunction with #10764 it would look like this:
image.image

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -156,7 +156,7 @@
             android:name=".CachePopup"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/app_name"
-            android:theme="@style/cgeo.popup"
+            android:theme="@style/Theme.AppCompat.Transparent.NoActionBar"
             android:windowSoftInputMode="stateHidden" >
         </activity>
         <activity
@@ -727,7 +727,7 @@
             android:name="WaypointPopup"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/waypoint"
-            android:theme="@style/cgeo.popup"
+            android:theme="@style/Theme.AppCompat.Transparent.NoActionBar"
             android:windowSoftInputMode="stateHidden" >
         </activity>
         <activity
--- a/main/res/layout/cachelist_spinneritem.xml
+++ b/main/res/layout/cachelist_spinneritem.xml
@@ -16,6 +16,7 @@
         android:layout_width="match_parent"
         style="?attr/titleTextStyle"
         android:textSize="@dimen/textSize_toolbarPrimary"
+        android:textColor="@color/colorTextActionBar"
         android:layout_height="wrap_content"
         android:maxLines="1" />
 
@@ -26,6 +27,7 @@
         android:ellipsize="end"
         style="?attr/subtitleTextStyle"
         android:textSize="@dimen/textSize_toolbarSecondary"
+        android:textColor="@color/colorTextActionBar"
         tools:text="This is the subtitle"
         android:id="@android:id/text2"
         android:maxLines="1" />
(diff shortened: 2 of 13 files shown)


Lineflyer:
Thats looks already much more nice IMHO also in combination with the accent outlined buttons.

Lineflyer (approved):

fm-sys:
The Signal messenger uses `#121212` and in another app I saw `#141414`. I think `#FF202020` is still a bit to light compared to other apps which use dark mode. Would it be possible to compare these tree colors?

fm-sys:
OK, just a quick draw via MS Paint:
`#202020` vs. `#141414` vs. `#121212`
grafik

These are no original emulator pictures, so not be confused why the action bar icons look that ugly...

I am of course aware that these are only very slight nuances, about which no one ever thinks again later on. Nevertheless, I would prefer the middle one ;-)

Lineflyer:
Yup, #141414 looks good and its the middle one. Let's use that for the beginning I would say.
BTW: A lighter action bar looks good together with this IMHO...but that just another topic.

eddiemuc:
#141414 looks fine. #121212 would also be ok for me.

fm-sys:
`#303030` would IMHO be corresponding action bar color for `#141414`

Actionbar color can be adopted via: (there might be different ways though)

```
<style name="cgeo" parent="...">
      <item name="actionBarStyle">@style/ActionBarStyle</item>
      ...
</style>


<style name="ActionBarStyle" parent="Widget.MaterialComponents.Light.ActionBar.Solid">
    <item name="background">@color/mycolor</item>
</style>

```

Or change the primary color, but this will AFAIK affect other components as well...

moving-bits:
Will this parent

> <style name="ActionBarStyle" parent="Widget.MaterialComponents.Light.ActionBar.Solid">

work for both light and dark theme?

(I'm asking because it has a `.Light.` in its name, and I don't know how those themes work wrt to day/night mode. And we do want to stay with a single theme for both dark and light mode...)

fm-sys:
Not sure... But there are some other ActionBar themes which we can use as base... (don't know the exact difference between them)

grafik

moving-bits:
Set background color to #141414, ActionBar background color to #303030 and ActionBar title text color accordingly, using the same colors in light and dark theme. Works for regular themed activities, map, popup and settings:

|dark theme|light theme|
|---|---|
|image|image|
|image|image|
|image|image|
|image|image|

Seems ok to me except the cache popup, as suddenly the transparent background got lost, although I've changed nothing in that are. I'll dive into that, but wanted to share the screenshots with you anyway.

fm-sys:
Wow, looks really nice IMHO 🤩👍

moving-bits:
Ok, found a way to get transparency back for the map's popup:

image

PR is now ready to merge from my side, if CI is happy.

fm-sys:
Just for light theme, we might need a lighter action bar background as soon as we use SVG which are pure black for light mode and pure white for dark mode. But this is kind of unrelated and can be changed easily in the future...

moving-bits:
Haven't tried, but should be doable by moving the `colorBackgroundActionBar` to `colors_light.xml` / `colors_dark.xml` and adjusting colors as needed, at least that's the idea behind it. For now I'm using the same colors for both themes.

fm-sys:
Yes, all fine. Please merge as soon as CI is green. I feel like I can't survive another day with current background 😅

PR #10767, opened 2021-05-26
fix #10635: added filters for log entry, last found and log counts

### LINKED ISSUE #10635: [New filter framework] Filter for Logbook properties
A filter shall be implemented to filter for Logbook properties. Should include:
* Number of finds + DNFs
* Last found date
* Found by + Not found by (note: "not-found-by" is also supported by GC.com filter search!)
* log content text search

eddiemuc:
fix #10635: added filters for log entry, last found and log counts


eddiemuc:
image

PR #10770, opened 2021-05-27
use toggle button group in quicksettings

### LINKED ISSUE #10742: Material theming: Routing buttons in map quick settings are all-orange
## Bug description 
The routing mode buttons in map quick settings currently are all-orange, thus it's pretty hard to see which one is the active button.

That may change if we decide upon a different look for the buttons (see #10739), but we should keep this bug in mind and fix it (if still needed), when the look for out buttons has been decided upon.

fm-sys:
fix #10742

### DIFF
--- a/main/src/cgeo/geocaching/maps/MapSettingsUtils.java
+++ b/main/src/cgeo/geocaching/maps/MapSettingsUtils.java
@@ -12,7 +12,6 @@ import cgeo.geocaching.utils.IndividualRouteUtils;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.functions.Action1;
 
-import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.app.Dialog;
 import android.view.LayoutInflater;
@@ -28,6 +27,7 @@ import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 
 import java.util.ArrayList;
+import java.util.Objects;
 
 import com.google.android.material.button.MaterialButtonToggleGroup;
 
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/maps/MapSettingsUtils.java
```diff
@@ -158,6 +159,50 @@ public void setValue() {
         }
     }
 
+    private static class ToggleButtonWrapper<T> {
```
eddiemuc, line 162: @fm-sys in my understanding this would replace the ToggleButtonGroup, right? Can you externalize it so it can be used in other places as well (e.g. the filter activity...)? Or should I do that later myself?

fm-sys: First of all, it's a replacement for `ButtonController` (an internal class of `MapSettingsUtils`). But if it looks suitable for you, I can externalize it as well...

eddiemuc: I see now that it is not a view itself.

fm-sys:
grafik

fm-sys:
I removed the "(Current: xxx)" from the Routing mode title, as it was kind of inconsistent and I think the icons are quite clear anyway (we had different icons at the time we implemented the title text). If you tell me, that we definitely need to keep it, I can re-add it for sure...

moving-bits:
Leaving out the description for routing mode buttons is ok for me (as I do know their meaning ;-)), but I remember us having some discussion about the icons, which explicitly led to adding that text.

@Lineflyer 
What is your feedback from a support perspective?

Lineflyer:
> What is your feedback from a support perspective?

I think we can omit the description of current mode.
I think the buttons are very understandable.

Lineflyer:
BTW:
IMHO we could also remove/rephrase the text below regarding installation of BRouter as we now have the internal engine (which is the preferable way IMHO).

eddiemuc:
Can we merge this one rather quickly? Right now the buttons are almost unusable in master.

PR #10772, opened 2021-05-27
WIP - Material theming: icon buttons

### LINKED ISSUE #10769: Material theming: icon buttons
Followup of issue  #10739.

Text buttons are fine now, but the PR had broken our current icon button implementation...

material.io should explain the suggested icon button implementation somewhere...

fm-sys:
fix #10769

---
grafik

If we want to theme our icon buttons like buttons (as before), we need to switch from `IconButton` to `Button` as the `IconButton` class is not really supported by material design...

What do you think? Should we go this way, or is this getting too colorful?

### DIFF
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -3,10 +3,11 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="fill_parent"
     android:layout_height="fill_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:background="@color/colorBackgroundTransparent"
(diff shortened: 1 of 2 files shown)


Lineflyer:
Not sure....do we necessarily need to combine this with using the accent color?

BTW: How does your screenshot look on light theme?

fm-sys:
I think it makes sense to theme text and icon buttons in the same style. If that gets to colorful, I think we should remove the aaccent color from the text button as well...

fm-sys:
Superseded by https://github.com/cgeo/cgeo/pull/10937

PR #10784, opened 2021-05-28
fix #5738, fix #10287, fix #10735: add own, found and offline log hide to map quick settings

### LINKED ISSUE #5738: Feature: Exclude/show only own, Exclude found, Exclude disabled, Exclude archived
In settings, Cache details, currently there are two options: 
- Exclude Own **and** Found
- Exclude Disabled

My feature request is for separate exclusions, as follows:
- [ ] Exclude Found
- [x] Exclude disabled
- [x] Exclude Archived
- [ ] Exclude Own
- [ ] Show only Own

The last two to be mutually exclusive, if one is set.
If set "show only own", the above exclusions for disabled and archived to apply.

This is aimed to help cache owners and teams working with cache owners to better manage caches without the specific need to prepare special purpose GPX files for each situation described above.

Thank you.

### LINKED ISSUE #10287: New Map Quick Filter: Exclude Caches with Offline Found Logs

Caches with offline found logs are de facto already found ones. Only difference to the ones counted as "real" found is, that their (offline stored) logs are not yet uploaded to the server. But this difference does not matter for cacher's real life: An already found cache will not be visited again (in most of the cases ;) ), thus it should be possible to eliminated it from map and search lists, to be able to declutter them more easily (which sometimes is a real issue esp. on longer tours).

- A setting ``Exclude Caches with Offline Found Logs`` (similar to ``Cache Details`` => ``Exclude Own and Found``).
- A tick box ``Caches with Offline found logs`` in the map's ``Quick settings`` popup.
- These should hide caches with offline found logs in the same way as "real" finds from
  - map
  - search results
  - nearby list
  - ... forgotten s/th?

### LINKED ISSUE #10735: Map quick settings: individual and separate "show caches" option for owned and found geocaches
User feedback on Facebook Finnish geocacher group.

They would like to get a separate option for displaying  found and owned geocaches in the quick map settings, each independent of each other.

Currently both are together in one option.

eddiemuc:
fix #5738, fix #10287, fix #10735: add own, found and offline log hide to map quick settings

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -67,8 +67,10 @@
     <string translatable="false" name="pref_friendlogswanted">friendlogswanted</string>
     <string translatable="false" name="pref_opendetailslastpage">opendetailslastpage</string>
     <string translatable="false" name="pref_excludemine">excludemine</string>
+    <string translatable="false" name="pref_excludefound">excludefound</string>
     <string translatable="false" name="pref_excludedisabled">excludedisabled</string>
     <string translatable="false" name="pref_excludearchived">excludearchived</string>
+    <string translatable="false" name="pref_excludeofflinelog">excludeofflinelog</string>
     <string translatable="false" name="pref_global_wp_extraction_disable">globalWpExtractionDisable</string>
     <string translatable="false" name="pref_excludeWpOriginal">excludeWpOriginal</string>
     <string translatable="false" name="pref_excludeWpParking">excludeWpParking</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -727,8 +727,10 @@
     <string name="init_summary_select_language">Select in-app language (requires restart)\n</string>
     <string name="init_use_default_language">Use default language</string>
     <string name="init_category_exclude">Exclude caches</string>
-    <string name="init_exclude">Exclude Own and Found</string>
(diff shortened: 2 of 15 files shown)


eddiemuc:
image

This PR is dedicated to @magpiefourtytwo, @andrixnet, @jakedot, @Vike91, the Finnish user group and all others wishing for a separation of "own" and "found" hide option in map settings

Lineflyer:
Nice. Does does this interwork with the settings part (Settings/Cache-Details) where own/found are coupled right now?

jakedot:
@Lineflyer from looking at the committed code, yes.

eddiemuc:
> Nice. Does does this interwork with the settings part (Settings/Cache-Details) where own/found are coupled right now?

Added there as well of course:

image

(although once the new filtering system replaced the old one, those settings will vanish since they will only be valid in map settings and not anywhere else)

MagpieFourtyTwo:
With these (very welcome) additional entries (thx a lot!) the Quick settings menu got significantly longer ... and now it does not completely open up on first click anymore, so even on large screens an additional swipe up is needed now. Moreover it's kind of hard to reach the uppermost entries.

So, would it be possible do decrease the line spacing?  E. g. like shown below?

Left: State after click on Quick settings icon - Middle: Fully opened after swipe up - Right: Initial state with decreased line spacing

Quick settings

If this will come with layout overhaul anyway, just forget about my question.  ;)

PR #10786, opened 2021-05-28
cache detail list: increase textSize and spacing (fix #10759)

### LINKED ISSUE #10759: Material theming: Cache details - Unify text size
The cache details view still has rather small text. I guess it needs to be adapted to make use of the overall defined text size.

Maybe for the list in the first section it might be wise to use different text styles for label (e.g. name, type, size, etc.) and content.

Text color is another thing, but here we should wait until the background is fixed to see how it looks like.
grafik

moving-bits:
## Description
Increases text size in cache details list to `detailsPrimary` and increase the space between label and item:

image

### DIFF
--- a/main/res/layout/cache_information_item.xml
+++ b/main/res/layout/cache_information_item.xml
@@ -7,19 +7,20 @@
 
     <TextView
         android:id="@+id/name"
-        android:layout_width="80dip"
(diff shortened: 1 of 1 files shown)


Lineflyer:
Looks good IMHO.
Many users requested larger text back in the days.

PR #10787, opened 2021-05-28
Bottom sheet fixes

fm-sys:
This will change the initial expanding to 80% (instead of 50%). The quick settings popup will therefore now even fill more screen than the old alert dialog did. A big plus for intuitive and fast usability... Furthermore, many many other bugs are fixed.

grafik

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
+++ b/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
@@ -1017,6 +1017,14 @@ public final class Dialogs {
         return dialog;
     }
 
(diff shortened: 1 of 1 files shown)


moving-bits:
Does this also fix a weird transition I'm experiencing in the emulator?
- On opening the bottom sheet, it's immediately displayed at the preset size,
- then moves to the left, until not bottom sheet is visible,
- and then returns from the bottom, again to it's preset size, and stays there.

(Otherwise I can/will open a separate issue for this.)

fm-sys:
Haven't noticed this one, but I guess it's very likely that it is fixed as well... 

I will merge now, as there's nothing real which could be discussed in this PR. Feel free to retest the issue :)

fm-sys:
ups. wrong button 😅

PR #10799, opened 2021-05-29
implement about pages with ViewPager2 and material tabs

moving-bits:
## Description
Reimplements c:geo's about pages using ViewPager2 and Material Tabs instead of the old viewpager.
Can be a base to remove the old viewpager completely (after having migrated `CacheDetailsActivity` to ViewpPager2 / Material Tabs as well.

This is a first shot at migration. Looks fully functional to me, but could use more safeguards, and I'm not sure yet, whether this approach will work for `CacheDetailActivity` as well. Just wanted to start with the easier part as a MVP...

image.image.image
image.image

### DIFF
--- a/main/res/layout/avpactivity.xml
+++ b/main/res/layout/avpactivity.xml
@@ -1,0 +1,29 @@
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical">
+
+    <androidx.viewpager2.widget.ViewPager2
+        android:id="@+id/viewpager"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1" />
+
+    <View style="@style/separator_horizontal" />
+
+    <com.google.android.material.tabs.TabLayout
+        android:id="@+id/tab_layout"
+        app:tabTextColor="@color/colorTextActionBar"
+        app:tabSelectedTextColor="@color/colorAccent"
+        app:tabIndicatorColor="@color/colorAccent"
+        app:tabPaddingStart="36sp"
+        app:tabPaddingEnd="36sp"
+        app:tabMode="scrollable"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textSize="@dimen/textSize_detailsPrimary"
+        android:background="@color/colorBackgroundActionBar" />
+
+</LinearLayout>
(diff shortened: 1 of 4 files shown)


fm-sys:
Looks really nice 👍

BTW, is there a reason why the tabs/viewpager is at the bottom in c:geo?

Lineflyer:
> BTW, is there a reason why the tabs/viewpager is at the bottom in c:geo?

That was the initial implementation at the time we introduced the first viewpager. Before that, it was all on one loooong page.

Regarding the position:
- Keeping it at the bottom to keep a bit of the existing look and feel. (e.g. I know many users who in fact tap on the headings to change the tab rather than swiping.)
- Move it to the top to look like other apps and maybe easier to find and learn for new users..

Looking at the pictures above I think, that my second point is not really valid. The new implementation is much more visible and understandable than our existing viewpager, no matter if bottom or top.

So I would prefer to keep it at the bottom unless some Material/Google guideline say, that this is a bad idea as it blocks some other nice UI feature.

Lineflyer:
BTW: @moving-bits Please enlighten me what MVP means?

UniQP:
@Lineflyer [Model–view–presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter)

fm-sys:
Screenshot_20210529_215429_org mozilla firefox

There's in fact such a guideline. But I guess it's only together with bottom navigation at the same time. But even if we use a bottom navigation in thw future it will not be displayed in the cache details activity.

moving-bits:
> @Lineflyer [Model–view–presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter)

Actually in this case I meant [Minimum viable product](https://en.wikipedia.org/wiki/Minimum_viable_product) - what's the smallest useful implementation (in this case: in contrast to full migration of all viewpager activities directly in the first step).

moving-bits:
> There's in fact such a guideline. But I guess it's only together with bottom navigation at the same time. But even if we use a bottom navigation in thw future it will not be displayed in the cache details activity.

Yeah, two lines of navigation elements can be really confusing, especially when their connection is hard to grap.
But that's not the case for c:geo - we have only a single line of navigation elements, neatly placed close to where the thumb is ;-)

Lineflyer:
> But that's not the case for c:geo - we have only a single line of navigation elements, neatly placed close to where the thumb is ;-)

You are right with both points. IMHO its kind of consistent then to have navigation at the bottom also if we later use the bottom navigation on main level.

fm-sys:
> > @Lineflyer [Model–view–presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter)
> 
> Actually in this case I meant [Minimum viable product](https://en.wikipedia.org/wiki/Minimum_viable_product) - what's the smallest useful

Funny, at a first glance I thought it was "material view pager" 😅

PR #10794, opened 2021-05-28
implement theme mode setting "follow system" (fix #10721)

fm-sys:
grafik

...with dark as default...

PR #10801, opened 2021-05-29
move icon resource and color to light/night folders

moving-bits:
## Description
Instead of manual distinction using Settings.isLightMode() move color and icon to the appropriate files/folders (if possible)


fm-sys:
Quite some of them are already tackled in #10794. Could you postpone this PR as it will become a total merge hell for me otherwise?

fm-sys:
Ok, it's even nearly the same as done in #10794. The only thing which is missing in my PR is the `setBackgroundColor` line which I can easily add to my PR...

moving-bits:
@fm-sys 
ok, then let's extend your PR and close here

PR #10804, opened 2021-05-30
switch to AndroidStudio 4.2.1

### LINKED ISSUE #10548: AS 4.2.x
Before upgrading AS to 4.2 (and also the gradle plugin) we need to check if there are maybe problematic changes.
The release notes are at https://developer.android.com/studio/releases#4-2-0 for AS and https://developer.android.com/studio/releases/gradle-plugin#4-2-0 for the gradle plugin.
Especially we have to check the known issues at https://developer.android.com/studio/releases/#4.2-known-issues .

bekuno:
- switch to gradle plugin 4.2.1 (supersedes #10620 and #10537)
- use aapt2-proto:4.2.1 (supersedes #10619)
- use the parallel garbage collector (new in AS 4.2)

For the releaseNotes see
- AS: https://developer.android.com/studio/releases/#4-2-0
- AGP: https://developer.android.com/studio/releases/gradle-plugin#4-2-0

fix #10548
supersedes #10620 #10619 #10537

Remarks:
1. garbage collector:
  I made a test as suggested in https://developer.android.com/studio/build/profile-your-build#gradle-profile-option .
  You have to call the following commands with and without the `-XX:+UseParallelGC` part.
~~~
gradlew clean
gradlew --profile --offline --rerun-tasks assembleBasicDebug
~~~
  On my dev machine (dual core) it shows sometime a time reducing. But I think this machine is not representative.

2. v3 and v4 signing
   see https://developer.android.com/studio/releases/gradle-plugin#v3-v4-signing
   Android Gradle Plugin 4.2 now supports APK v3 and APK v4 signing formats. 
   This is not included in this PR .

3. new property `android.native.buildOutput`
   see https://developer.android.com/studio/releases/gradle-plugin#native-build-output
   This is not related to our java build. I saw no extra output during the tests. Therefore it is not used.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:4.1.3'
+        classpath 'com.android.tools.build:gradle:4.2.1'
 
         // un-mocking of portable Android classes
         classpath 'de.mobilej.unmock:UnMockPlugin:0.7.6'
@@ -105,9 +105,9 @@ allprojects {
                 // new versions of checkstyle use guava 27, which has troublesome dependencies: https://groups.google.com/forum/#!topic/guava-announce/Km82fZG68Sw
                 force 'com.google.guava:guava:26.0-jre'
 
-                // force this version because of conflict to our resolutionStrategy from lint-gradle 4.1.3 (referencing 4.1.0-alpha01-6193524)
+                // force this version because of conflict to our resolutionStrategy from lint-gradle 27.2.1 (referencing 4.1.0-alpha01-6193524)
                 // TODO check after gradle update (gradlew :main:lintBasicDebug)
-                force 'com.android.tools.build:aapt2-proto:4.1.3-6503028'
+                force 'com.android.tools.build:aapt2-proto:4.2.1-7147631'
             }
         }
     }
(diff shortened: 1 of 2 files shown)


moving-bits:
Is switching to AS 4.2 a prerequisite for it, or does it run in AS 4.1.x as well?

bekuno:
> 
> 
> Is switching to AS 4.2 a prerequisite for it, or does it run in AS 4.1.x as well?

Good question.
I upgraded AS to 4.2.1 first.
I don´t know, if the change in `gradle.properties` will result in a failure in AS 4.1.

Two points which struck me on AS 4.2.1:
- The font for the editor and terminal windows was changed
- The gradle commands list is defective 

grafik

eddiemuc:
lgtm, can this be merged?

bekuno:
> 
> 
> lgtm, can this be merged?

Yes from my side.

But I dont know if someone needs to stay with AS 4.1.x.

eddiemuc:
> > 
> > 
> > lgtm, can this be merged?
> 
> Yes from my side.
> 
> But I dont know if someone needs to stay with AS 4.1.x.
> 

Proposal: let's wait one more day, if no one complains until then we merge

moving-bits:
I'm still on 4.1, but will move to 4.2 then, so no objections from my side.

eddiemuc:
So let's switch to AS 4.2 :-)

bekuno:
>     * The gradle commands list is defective

This is a side effect of a new experimental setting (activated on default):
grafik

You can update the task lists by clicking on the context menu item "Relead Gradle Project":
grafik

PR #10807, opened 2021-05-30
set circular spinner color (fix #10793)

### LINKED ISSUE #10793: Progress spinner has wrong accent color
When refreshing a cache the progress spinner is using some greenish color instead of our orange color accent (tested on dark theme).
Seen in todays nightly 2021.05.28-NB

moving-bits:
## Description
Sets the color of our circular indeterminate spinner to `accentColor` (e. g.: open cache details when coming from map)


moving-bits:
will be superseeded by #10812, therefore closing here

PR #10826, opened 2021-06-02
adjust minHeight of all material checkboxes

moving-bits:
## Description
Found a way to apply the reduced line spacing to all material checkboxes, which will make it work not only in the map quick settings, but in other places, too, e. g. in the cache type filter:

|before|after|
|---|---|
|image|image|

The workaround (directly applying a new `minHeight` for the map quick settings item) can therefore be removed.

### DIFF
--- a/main/res/layout/map_settings_item.xml
+++ b/main/res/layout/map_settings_item.xml
@@ -26,7 +26,6 @@
         android:clickable="false"
         android:layout_width="30dp"
         android:layout_height="wrap_content"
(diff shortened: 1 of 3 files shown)


Lineflyer:
Without having seen it on a real device:
How about something in the middle between before and after?
But we can also just try and use it and decide later...

fm-sys:
Nice! The same should IMHO be applied for radio buttons as well...

PR #10841, opened 2021-06-04
fix #10817: apply two-column-layout to map-quick-settings

### LINKED ISSUE #10817: Make map quick settings for caches/waypoints a two columned table
With its recent additions and in combination with material theming and migration to a bottom sheet dialog, the map quick settings dialog got pretty long, too long for most screens to fit onto one page.

## Suggested solution
Put the "show" options for caches and waypoints into a two-column layout, like we have in the nice new "type" cache filter.

One possible order could be similar to this:

|first column|second column|
|---|---|
|__caches:__|own|
|found|w/offline log|
|disabled|archived|
|---|---|
|__waypoints:__|visited|
|original|parking|
|---|---|
|__Other:__|circles|

(With `---` indicating a thin horizontal divider line)

Below that table the other existing options (e. g. routing mode) will follow.

## Alternatives/workarounds
Other possible arrangements for above's options are described in #10809

eddiemuc:
fix #10817: apply two-column-layout to map-quick-settings


eddiemuc:
image

eddiemuc:
Damn! Although I discussed this up and down with you, I just now realize that I used white text color for the text instead of gray!

I assume you want me to change this back. But I have two questions (yet again... sorry...):

* Which color is the right one to use (I currently use `colorText `which is white)?
image

* The following is a screenshot of map quicksettings dialog in `master`, the lower part only. As you can see, here the text color of the headline "Individual route" is gray while the text after the checkbox is white. Is the checkbox text also supposed tp be gray?
* 
image

eddiemuc:
...changed the "subtitle" color to "colorTextheadline" (as the color name suggests already...), this gives me the following total picture:

image

moving-bits:
Looks really nice, thanks for your efforts!

Regarding the colors: 
I guess we need to define what colors we would like to have in our dialogs. (Despite the fact that we still need to find a way to easily apply them to customized dialogs...)

My opinion: As the background is a bit lighter (in dark mode) for dialogs as for activities, I would not use grey as the default color for text, the contrast would be too low - I would stay with white for it. Headings in grey would be ok, IMHO, as they are less frequent and a big bigger, and by choosing a (slightly) different color than the main text, it makes it a bit more distinguishable from main text.

TMALSS: Stay with white for main text (black in light mode), and grey for the headings.

main/res/layout/checkbox_item.xml
```diff
@@ -47,6 +50,7 @@
         android:id="@+id/item_checkbox"
         android:layout_width="30dp"
         android:layout_height="wrap_content"
+        android:minHeight="32dip"
```
moving-bits, line 53: Isn't that covered by the main theme defining a `checkboxStyle`, which defines a `minHeight`?

Lineflyer:
What happens if translations are longer? Line break or ellipsis?

eddiemuc:
> What happens if translations are longer? Line break or ellipsis?

Same as in filter Dialogs: ellipsis

main/res/layout/checkbox_item.xml
eddiemuc: > Isn't that covered by the main theme defining a `checkboxStyle`, which defines a `minHeight`?

You're right, that was an unwanted leftover from my "experiments". I removed it, thanks for the review!

PR #10848, opened 2021-06-04
use solid button in wizard for default action only (fix #10835)

### LINKED ISSUE #10835: Installation wizard still using wrong button type
The installation wizard still uses the "full orange" buttons.

`Master` as of 03.06.2021

moving-bits:
## Description
- uses solid button for default action only
- pages without a default action (because of having several actions of equal importance) will no longer have any solid button at all (e. g.: services page, advanced settings page)

image.image.image.image

### DIFF
--- a/main/res/layout/install_wizard.xml
+++ b/main/res/layout/install_wizard.xml
@@ -98,8 +98,13 @@
             android:layout_weight="1" />
         <Button
             android:id="@+id/wizard_next"
-            style="@style/InstallationWizardButtonFooter"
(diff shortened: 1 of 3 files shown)


fm-sys:
> image

"Geocaching.com" should get a fully orange styled button too.

moving-bits:
> "Geocaching.com" should IMHO get a fully orange styled button too.

I had chosen not to colorize it as to not promote any specific service explicitly.

Lineflyer:
Maybe both upper buttons to draw attention to the services instead of "Skip"?

PR #10849, opened 2021-06-04
Export formula to PN instead of coordinates for completed formula

murggel:
## Description

Exports the formula of a waypoint to the PN, even if all variables are defined and the coordinates could be calculated.
This would be to have the variables "archived" in the PN.

## Related issues

#9211 
#9633 

## Additional context

see comment of @Lineflyer  https://github.com/cgeo/cgeo/issues/9211#issuecomment-854134154


murggel:
retest this please

PR #10857, opened 2021-06-05
make slider color context-aware (fix #10852)

### LINKED ISSUE #10852: Fix slider layout (in filter dialog) after GUI redesign
After GUI redesign the labels on sliders for difficulty/terrain filters are not readible any more.

moving-bits:
## Description
Currently slider label uses a color retrieved from (static) application context, which returns black always.
Make this context-aware to retrieve a theme-specific color.

image.image


eddiemuc:
Thanks for the fix!

PR #10859, opened 2021-06-05
add attribution to material design icons (fix #10858)

### LINKED ISSUE #10858: Add license for materialdesignicons
Some of the svgs I used were taken from materialdesignicons.com, which is under the Apache 2.0 license.

We have to add it and mention, that some were modified. I am not sure, if we have to mention, which ones are modified or if a general hint is sufficient.

moving-bits:
## Description
Add attribution to material design icons in about:contributors' page

### DIFF
--- a/main/res/values/strings_not_translatable.xml
+++ b/main/res/values/strings_not_translatable.xml
@@ -145,6 +145,7 @@
     - [RRZE Icon set](https://github.com/RRZE-PP/rrze-icon-set) ([CC-BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/))\n
     - [Layers icon by Cole Bemis](https://iconfindr.com/1mNr3rl) ([CC-BY 3.0](https://creativecommons.org/licenses/by/3.0/))\n
     - [Google Material Design Icons](https://material.io/tools/icons/) ([Apache License version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html))\n
+    - [Material Design Icons](https://materialdesignicons.com/) ([Apache License version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html))\n
     - [DragSortListView library by Carl Bauer](https://github.com/bauerca/drag-sort-listview/) ([Apache License version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html))\n
     - [Geocoding courtesy of MapQuest](https://www.mapquest.com/)\n
     - [Geocoding data from OpenStreetMap](https://www.openstreetmap.org/)\n


fm-sys:
Please note that the icons are officially designed by material.io which is owned by Google. The website you mentioned is just a third-party website, kind of a wrapper for the icon collection. 

Correct attribution would be Material Design Icons (https://material.io/) Licensed under Apache-2.0 License.

moving-bits:
ok, if they are all from material.io originally, then we already have attribution for it, see the line above it.
My thought was, that there are also community-built icons not covered by material.io, but haven't checked it explicitly.

fm-sys:
Ok, then we should probably find a more suitable name then "Material Design Icons". Maybe "Community Material Icons" instead?

Cause the "material design" is the official Google project. So if we just use "material" it's clear that they do not directly belong to this design framework but are based on the same principles...

moving-bits:
Let's first ask @murggel whether the icons are all material.io based - then we don't need an addition attribution, or where else they do come from, and under what license (seems to be Apache 2.0, but let's be on the safe side).

murggel:
No, they are not all the official ones, some are from the "community"
I am just preparing an overview, which are original material.io, which are from community and which are modified...

moving-bits:
ok, then we need to look whether any of the community icons has a different license than Apache 2.0
Do we need to add a per-icon attribution? (Like we do for The Noun Project)

murggel:
I found following info respect to the license:
https://github.com/Templarian/MaterialDesign/blob/master/LICENSE
There it is mentioned, 
```
This icon collection is released as free, open source, and GPL friendly by the [Pictogrammers](http://pictogrammers.com/) icon group. You may use it for commercial projects, open source projects, or anything really.
Icons: Apache 2.0 (https://www.apache.org/licenses/LICENSE-2.0)
Some of the icons are redistributed under the Apache 2.0 license. All other icons are either redistributed under their respective licenses or are distributed under the Apache 2.0 license.
```

But I did not find, under which license e.g. the following icon is, which I used
grafik

So, if nothing is mentioned, does it mean, it is under the Apache 2.0 license?

moving-bits:
Yeah, that's probably the same as I have found: https://dev.materialdesignicons.com/license

So I guess: if no other license is mentioned specifically, Apache 2.0 license can be applied.

murggel:
> 
> 
> Yeah, that's probably the same as I have found: https://dev.materialdesignicons.com/license
> 
> So I guess: if no other license is mentioned specifically, Apache 2.0 license can be applied.

Ok, then I will go through all "communtiy"-icons and check, whether a special license is applied - I guess not, but who knows.

Thanks for cross-check!

PR #10862, opened 2021-06-05
revert global listheight, format alertdialog lists (fix #10861)

### LINKED ISSUE #10861: Overflow menu hight
Some of the recent changes had changed the overflow menu with as well, which is now vertically much to small compared to other apps...

Screenshot_20210605_163554

moving-bits:
## Description
- revert setting a global list line height
- add optional `styleRes` parameter to `Dialogs.newBuilder()`
- create a style `compactDialogs` based on `cgeo` and material alert dialogs with a reduced list line height and correct font colors and sizes
- and apply this style to the three "pick a list" dialogs

image.image.image

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
+++ b/main/src/cgeo/geocaching/ui/dialog/Dialogs.java
@@ -43,6 +43,7 @@ import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
+import androidx.annotation.StyleRes;
 import androidx.appcompat.app.AlertDialog;
 import androidx.core.content.res.ResourcesCompat;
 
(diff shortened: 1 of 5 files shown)


main/src/cgeo/geocaching/ui/dialog/Dialogs.java
```diff
@@ -1077,6 +1078,10 @@ public void onSlide(final @NonNull View bottomSheet, final float slideOffset) {
         return new MaterialAlertDialogBuilder(newContextThemeWrapper(context));
     }
 
+    public static AlertDialog.Builder newBuilder(final Context context, final @StyleRes int resId) {
```
eddiemuc, line 1081: I wonder if it makes sense to restrict the possibilities for possible dialog styles (and thus provide e.g. some enum as parameter instead of the "raw" resId)? My fear is that this encourages developers in the future to invent more and more "dialog styles"

eddiemuc (approved):

main/src/cgeo/geocaching/ui/dialog/Dialogs.java
moving-bits: Let's keep this in mind and see at the end of our current UI overhaul how many styles we need, and if we can stay (or come back to) a single `AlertDialog` style applied for all of our dialogs, or if we need this parameter for every dialog, with a specific value. (I'm thinking of all those user-defined dialogs.)

PR #10866, opened 2021-06-05
Optimize Icons

fm-sys:
- remove explicit alpha=1 definition from SVG

- use correct tint for three_dot_vertical

- use material.io compass icon
Screenshot_20210606_002034_org mozilla firefox

The official material design icon is much easier to recognize as compass than a "cross with a circle"...


moving-bits:
currently has a merge conflict

fm-sys:
Yes, noticed it. But actually it's still WIP anyway - I will set the according label...

moving-bits:
Oh, sorry - I was about to merge it already ;-)

murggel:
@fm-sys I have some optimiziations regarding some svgs as well (using absolute instead of relative paths)
Shall I share those with you or should I create an own PR?

fm-sys:
> Oh, sorry

No need to say sorry :) 



> @fm-sys I have some optimiziations regarding some svgs as well (using absolute instead of relative paths)
> Shall I share those with you or should I create an own PR?

Probably it's easier if you create a separate PR. Maybe after this one is finished, as I plan to modify some more icons to make them more understandable...?

murggel:
> > @fm-sys I have some optimiziations regarding some svgs as well (using absolute instead of relative paths)
> > Shall I share those with you or should I create an own PR?
> 
> Probably it's easier if you create a separate PR. Maybe after this one is finished, as I plan to modify some more icons to make them more understandable...? 

Ok, I also thought, some are not suitable, but I wanted to stick to the original for the first shot.
Thanks for your help...

fm-sys:
Ready from my side:

| new | Current/old icon |
|-|-|
| grafik | grafik |
| grafik | grafik |
| grafik | grafik|

fm-sys:
What do you think: Should we use the filled or outlined version?

grafik

fm-sys:
Filled would look like this: 
grafik

IMHO, this actually looks better than outlined...

murggel:
With the filled navigation, I would prefer the outlined version, the filled one looks so "bold". Do you also have a screenshot from light theme? 


BTW Do you want to use the same compass icon for the waypoint list?

fm-sys:
> Edit: You use the same compass icon for the waypoint list, correct?

Yes, never have understood why cgeo uses two different icons for one and the same action...

murggel:
@fm-sys  Does the icon for my_location works for you in light mode? In my emulator it is black... 
grafik
grafik

fm-sys:
Hm, that's strange. Need to look into it...

Lineflyer:
I agree to @murggel , that I the outlined versions are preferable.

murggel:
> Hm, that's strange. Need to look into it... 

The difference to the other ones - it is used in a selector. I dis not get that work. But probably it will work, if we remove the tint as a workaround, since the icon is only used in the action bar

fm-sys:
> But probably it will work, if we remove the tint as a workaround, since the icon is only used in the action bar

Yes, that would probably work also its not the preferred way. 


I guess the icon is inflated programmly and uses the content text color, not the one from the actionbar. Maybe I have more luck finding a solution... ;-)

fm-sys:
> I agree to @murggel , that I the outlined versions are preferable.

Note to myself:
Use outlined versions for "near me" icon as well

murggel:
> > I agree to @murggel , that I the outlined versions are preferable.
> 
> Note to myself:
> Use outlined versions for "near me" icon as well

For that one I am not sure, how the outlined one will look like 😊
For me it will be ok to gone with the outlined for the compass only and leave the near me filled.

fm-sys:
I think we need to be consistent. Currently, we mostly use filled icons in the actionbar. And IMHO, the filled compass icon is easier to recognize as compass than the outlined. 

So I would prefer if we go with the filled version for now. Often it's just a matter of habit. Let's use the filled version at least for some days and then decide if it's fine or if we can't get used to it at all. 

Anyway, if we really change the icon style to outlined, "near me" would look like this: 
Screenshot_20210608_120123_org mozilla firefox

fm-sys:
grafik

Found a way to get the icon coloring working in light theme...

fm-sys:
Let's merge. I guess this was not the last "optimize icons" PR, but let's test the current state in the nightlys to gather feedback...

murggel:
_Originally posted by @eddiemuc / @murggel in https://github.com/cgeo/cgeo/pull/10818#issuecomment-855421559
> Would it make sense to copy this list to a new wiki page?
>> IMHO yes, so new ones can easily be added.
We need probably a page to describe the modification / tint for the svg anyway


How can / shall we track the changes of the used icons? 
Add comment in the PR, which icon from which source is used or just update the wiki (and hope this will not be forget)?
Or wouldn't it be easier to add a comment regarding the origin of the svg in the svg itself?

moving-bits:
maybe it's the easiest to document it in the SVG itself, if there's such a possibility. 
Text could be something like "Based upon <name of pic> from material.io, <name of licence>".

murggel:
Ok, I will add a comment to the svgs, referencing the source of the svg ( file name and attribution name)

PR #10870, opened 2021-06-06
fix #10682: add list filter, generify long-checkbox-list-solution

### LINKED ISSUE #10682: [filtering] add "contained in list" filter
I have a list named "solved mysteries" in c:geo. When I go our caching, I would like to list all caches, that can be found immediately. These are "Tradis", ... and "solved mysteries". The advanced query would be "Type: ... OR List: ..."

Please add a filter "Contained in list" (checkbox based - similar to the currently existing type filter - multi selection would be an OR / selectable by the user)

BTW, while we are on the way (and I don't want to open an extra issue for it): What do you think of an option to choose if multiple selected attributes in the attribute filter are parsed as AND or OR?

## Alternatives/workarounds

Unfortunately, there is no one currently

eddiemuc:
fix #10682: add list filter, generify long-checkbox-list-solution


eddiemuc:
Freshly added filter:
image

Clicking on "Add more items" shows all lists:
image

After selecting and clicking 'ok' those lists are now shown:
image

If by using this mechanism all user lists are in display, then the "add more items" button goes away:
image

eddiemuc:
...and now also including the counts
image

fm-sys:
Cool! It's an OR when multiple lists are selected, right?

eddiemuc:
> Cool! It's an OR when multiple lists are selected, right? 

Yes it is :-)
(Same as in all filters in this style btw, e.g. cache type or cache size)

moving-bits:
I'm really looking forward to our next feature release, so many nice things coming. Thanks @cgeo/team and all our contributors for putting so much effort in it!

Issue #10868, opened 2021-06-06
[MapQuickSertings] Reduce font size or shorten strings


fm-sys:
My screen size is quite big, 6", but still all strings are truncated - some of them even unusable. We could either leave out "caches" and "waypoints" in the strings as that's contained in the title anyway, and/or reduce the font size... 

![Screenshot_20210606_084449_copy_539x545](https://user-images.githubusercontent.com/64581222/120915391-a8451280-c6a3-11eb-9d2d-5003e6d9524c.png)

moving-bits:
we should shorten the titles, e. g.
- "Gefundene Caches" => "gefunden"
- "Caches mit Offline-Log" => "Offline Log"
- "Park-Wegpunkte" => "Parken"
- "Kreise anzeigen" => "Kreise"

etc.

If that's still not enough, we could
- reduce icon size (they are too big in this view, IMHO)
- leave out the icons (are they really needed?)
- change font size for the labels from `textSize_detailsPrimary` to `textSize_detailsSecondary` (as last resort)

eddiemuc:
Icons are too large indeed...

Text values are already shortened in english version. I assumed that changing those will retrigger Crowdin translations, is this not the case?

![image](https://user-images.githubusercontent.com/6909759/120915898-ae88be00-c6a6-11eb-95ae-bd1005d7005d.png)

murggel:
I can try to use for those buttons a smaller icon size

fm-sys:
> I assumed that changing those will retrigger Crowdin translations, is this not the case?

Yes, retranslation will be triggered. So it might even be fixed in today's nightly, but I don't want to upgrade currently as there are major bugs in it. (#10867)

So part 1 is already solved, still we could reduce the icon size...

moving-bits:
> Text values are already shortened in english version. I assumed that changing those will retrigger Crowdin translations, is this not the case?

Yes, it should trigger Crowdin, but that may take some time until all translations are back.
If you need to change something in a breaking way, one should change the string id, so that all translations fall back to English.

eddiemuc:
> > Text values are already shortened in english version. I assumed that changing those will retrigger Crowdin translations, is this not the case?
> 
> Yes, it should trigger Crowdin, but that may take some time until all translations are back.
> If you need to change something in a breaking way, one should change the string id, so that all translations fall back to English.

[...]

PR #10872, opened 2021-06-06
use smaller icon-size in map-settings (rel #10868)

murggel:
## Description

use smaller icon size in map-settings
grafik

has also affects in filter-activity and even there for the cache-type-icons. I am not sure, if that is ok as well. Otherwise I have to find a different solution
grafik

## Related issues

#10868


eddiemuc:
I strongly believe that GUI elements should be similar across different activities and dialogs, and thus I think this is a wanted side-effect

Issue #9908, opened 2021-01-30
DNF state not exported in GPX


Lineflyer:
Not sure if its technically possible to solve:

2021.01.30-NB

- Have a list with two caches: One with found state (yellow smiley), one with state DNF (blue smiley)
- Export to GPX and activate checkbox "including found state"
- Import the GPX to another c:geo instance (same username)

Result:
- Caches imported. Found cache has yellow smiley but DNF cache has no smiley (only after manual refresh)

Expected:
- Export found state should include DNF as well

fm-sys:
That's indeed true. However I have never really worked with GPX files before, and I don't know, if there is an official parameter to store a DNF state. Does anyone have an idea? How do other geocaching tools do it?

moving-bits:
You can check at https://www.topografix.com/GPX/1/1/, that's the "official" spec for GPX files. But there are also a couple of extensions, e. g. GSAK or even c:geo itself.

Lineflyer:
Checked an export of a found cache and it contains a string in wpt/sym with content "Geocache Found". Not sure though whether its our own namespace extension.

murggel:
I checked the gpx-schema and it is an offical tag, but doesn't recommend values. So it depends on the programs which interact.

```
<sym>Scenic Area</sym>
Text of GPS symbol name. For interchange with other programs, use the exact spelling of the symbol on the GPS, if known.
<type>

<type>Hiking trail</type>
type (classification) of element. 
```
We use "Geocache" and "Geocache Found" on exporting GPX and specify the waypoint type in the type-tag.
At import we interpret sym- and type-tag to identify the waypoint type.

So we could introduce a sym-value "Geocache Not Found" to exchange with another cgeo-instance. But I don't know, how other programs will react.

Probably we can export a DNF-log-entry and act on importing on that log.

murggel:
@moving-bits you can assign this to me, if it is ok to use the GSAK-tag as I think this would be an easy and fast way to solve this, if it is mainly only relevant for transfering it from one c:geo-instance to another.

Lineflyer:
@murggel  Assignment done.

murggel:
Done with PR #10874.

fm-sys:
@murggel feel free to close issues yourself if you like. If possible with a short note why the issue was closed (e.g. duplicate, fixed by PR #xxxx, ...)

murggel:
[...]

PR #10874, opened 2021-06-06
Read and write GSAK-Extension for DNF and visitedDate

murggel:
## Description

Support DNF in exported GPX (use existing GSAK-extension-field)
Also export visited date

## Related issues

#9908


murggel:
@moving-bits where do I have to add the fix #xxx so that the issue will be closed automatically?  
Putting it in the PR-description does appearently not work.
Or does that only work for team-members, which would be fine for me...

moving-bits:
There are two ways:
- You can put it into the commit message, e. g. append a "(fix #12345)"
- or you can use the "linked issues" menu you can see at the right, and select the issue you want to auto-close on merging

Linked messages will be closed only when merged to `master` branches, so if you prepare a PR for `release`, the linked issue will not be closed directly, but still automatically as soon as `release` is merged back to `master` (which is something mostly @Lineflyer and I are doing on a regular basis).

fm-sys:
In the current PR description the "fix" is missing in front of the issue id. And closing words in the PR description are only working if the PR targets to master. 



>     * or you can use the "linked issues" menu you can see at the right, and select the issue you want to auto-close on merging

This is only available for team members, but putting it in the commit message does work best (also for `release`).

murggel:
Thanks guys, will try ot the next time

bekuno:
Additionally to the words of @fm-sys , a link in the referenced issue is then automatic created.

Issue #8716, opened 2020-07-30
Country filter should only offers countries contained in the list to be filtered


Lineflyer:
- Have one list with caches from many countries
- Have another list with caches of one country
- Open the list with caches of one country only
- Select Filter - Country

Result:
Filter options show all countries contained in all lists

Expected Result:
Offer only those countries in the filter which match the list content

Note:
I came across this as a user reported, that calling the filter keeps crashing the app, although he only has a single cache on the list where he tries. It turned out that he has a cache on another list, which causes a problem. (Still trying to find out which cache causes this to open a dedicated issue for that).

moving-bits:
Our current filter implementation is not suited for this, as the filters do not have any information regarding the current cache list and its contents. 

Related to #8273

eddiemuc:
Isnit ok to close this as duplicate of #8273 (which introduces a broader view on the topic)?

eddiemuc:
Before I dive into this deeper: do we really need a country filter? Seems like a rare edge case to me for people living/caching in areas along country borders

moving-bits:
Or people wanting to list "what caches have I visited in country xx". (Don't know, though, how frequent that use case is.)

eddiemuc:
See PR #10875 for an implementation and a discussion for this implementation

PR #10875, opened 2021-06-06
fix-attempt for #8716: add location filter as text filter with suggestions, fix oc location parsing, introduce material testlayout

eddiemuc:
fix-attempt for #8716: add location filter as text filter with suggestions, fix oc location parsing, introduce material testlayout

### DIFF
--- a/main/res/layout/cache_filter_generic_string.xml
+++ b/main/res/layout/cache_filter_generic_string.xml
@@ -11,14 +11,18 @@
         android:id="@+id/filter_options"
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:baselineAligned="false"
-        android:orientation="horizontal" >
+        android:baselineAligned="true"
+        android:orientation="horizontal">
 
         <Button
             android:id="@+id/select"
-            style="@style/button_full"
+            style="@style/button"
+            android:layout_marginHorizontal="10dp"
+            android:layout_height="wrap_content"
             android:layout_width="0dip"
+            android:layout_gravity="center_vertical"
             android:layout_weight="1"
+            android:gravity="center"
             tools:text="contains" />
 
         <RelativeLayout
@@ -26,6 +30,7 @@
             android:layout_weight="1"
             android:layout_height="wrap_content"
             android:baselineAligned="false"
+            android:layout_gravity="center_vertical"
             android:orientation="horizontal">
 
             <CheckBox
@@ -37,8 +42,8 @@
 
             <ImageView
                 android:id="@+id/item_info"
-                android:layout_width="36dp"
-                android:layout_height="36dp"
+                android:layout_width="18dp"
+                android:layout_height="18dp"
                 app:srcCompat="@drawable/settings_info"
                 android:layout_alignParentRight="true"
                 android:layout_centerVertical="true"/>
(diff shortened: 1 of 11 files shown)


eddiemuc:
This PR adds a location filter as a text filter (unlike today, where the location/country filter is a list filter). The reason for this is that the location is de facto stored as a text field at the cache, it is not localized etc in any way and most often it contains both country and region - and region should also be searchable:

image

In addition to pure-text-search I added autocompletion which is based on all available regions and countries in the database. The system is already well-known in c:geo from e.g. "Search by geocode" or "Search by owner": after typing two characters, a list opens with suggested countries and regions from database. In the following example I typed "er" (note that the database of my emulator is quite limited in terms of stored caches):

image

I think this is a valid compromise to maintaining a long list of countries and also allows users to define filters for countries/regions where e.g. no cache is yet in the database

eddiemuc:
Something else: I noticed during test, that from oc.de the location info (region and country) was not loaded from the server. Thus I added it, it is now loaded and filled in the DB. I hope this is ok, was there a specific reason for not including it yet?

Lineflyer:
> Something else: I noticed during test, that from oc.de the location info (region and country) was not loaded from the server. Thus I added it, it is now loaded and filled in the DB. I hope this is ok, was there a specific reason for not including it yet?

Not sure if all supported OC nodes are supporting this?

eddiemuc:
> > Something else: I noticed during test, that from oc.de the location info (region and country) was not loaded from the server. Thus I added it, it is now loaded and filled in the DB. I hope this is ok, was there a specific reason for not including it yet?
> 
> Not sure if all supported OC nodes are supporting this?

I added safeguard of course :-)

moving-bits:
currently has merge conflicts

eddiemuc:
Not any more.

btw, @moving-bits , I am not sure if I use EmojiUtils right. I wanted to draw markerIds in front of stored lists if one is assigned to it. E.g. like this ("Erste-Hilfe-Runde" has an earth emopji assigned as a list marker)
image

As far as I understand I have to create a EmojiUtil.EmojiPaint first. For this I copied 1:1 the code from class MapMarkerUtils, including e.g. using R.drawable.black_dot as a measure. I am not sure what all of this stuff is doing, can you take a look at it?

image

eddiemuc:
@moving-bits come to think of it, the emojis are a different topic, I will open a separate PR for this...

eddiemuc:
retest this please

moving-bits:
> As far as I understand I have to create a EmojiUtil.EmojiPaint first. For this I copied 1:1 the code from class MapMarkerUtils, including e.g. using R.drawable.black_dot as a measure. I am not sure what all of this stuff is doing, can you take a look at it?

Yes, will look at your PR. Just a bit of context: I need a Paint object to be able to draw on its surface, and I try to make the dimensions of this object the same as the type of image I want to replace by it. For the list markers on the map this is equivalent to the `black_dot` marker, for a cache symbol I take `marker_oc`, and for the buttons in the emojiutils popup I take `ic_menu_filter` as (size) template:

```
// calc sizes for markers
final Pair<Integer, Integer> markerDimensionsTemp = DisplayUtils.getDrawableDimensions(ResourcesCompat.getDrawable(context.getResources(), R.drawable.ic_menu_filter, null));
final int markerAvailable = (int) (markerDimensionsTemp.first * 0.5);
final Pair<Integer, Integer> markerDimensions = new Pair<>(markerAvailable, markerAvailable);
final EmojiPaint paint = new EmojiPaint(context.getResources(), markerDimensions, markerAvailable, 0, DisplayUtils.calculateMaxFontsize(35, 10, 150, markerAvailable));
```

Coming from the dimensions of the image I calculate the space available for the emoji itself, 50% height / 50% length in this case. Then I have to try different font sizes - which is the largest font size which makes the emoji drawn still fit into the calculated area. And to not have to try out every font size I give them some min/max values (initial: 35, min: 10, max: 150 in this case).

PR #10883, opened 2021-06-07
add emojis to stored-list-filter and create ImageParam class

eddiemuc:
* add emojis to stored-list-filter and create ImageParam class
* also introduces the class "ImageParam" to pass image parameters to methods. 
* Follow-up idea óf second point would be to create similar classes/builders "TextParam", "ActionParam" etc and use these to reduce the necessary overloads in utility classes like "Dialogs" or "ViewUtils" which are currently necessary e.g. to pass Texts as id OR String

image

@moving-bits I would appreciate a review by you because I extended "`EmojiUtils`" and could really use your opinion on this.

### DIFF
--- a/main/src/cgeo/geocaching/utils/EmojiUtils.java
+++ b/main/src/cgeo/geocaching/utils/EmojiUtils.java
@@ -1,5 +1,6 @@
 package cgeo.geocaching.utils;
 
+import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
 import cgeo.geocaching.databinding.DialogTitleButtonButtonBinding;
 import cgeo.geocaching.databinding.EmojiselectorBinding;
@@ -35,6 +36,9 @@ import androidx.appcompat.app.AlertDialog;
 import androidx.core.content.res.ResourcesCompat;
 import androidx.recyclerview.widget.GridLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
+
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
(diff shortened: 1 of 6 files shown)


eddiemuc:
@moving-bits can I merge this one? I want to build up on it.

main/src/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -416,6 +422,30 @@ public static BitmapDrawable getEmojiDrawable(final EmojiPaint paint, final int
...
+    @NonNull
+    public static BitmapDrawable getEmojiDrawable(final int wantedSize, final int emoji) {
+
+        EmojiPaint p;
+        synchronized (EMOJI_PAINT_CACHE_PER_SIZE) {
+            p = EMOJI_PAINT_CACHE_PER_SIZE.get(wantedSize);
+            if (p == null) {
+                final Resources res = CgeoApplication.getInstance().getApplicationContext().getResources();
+                final Pair<Integer, Integer> temp = new Pair<>(wantedSize, wantedSize);
+                final Pair<Integer, Integer> markerDimensions = new Pair<>((int) (temp.first * 1.2), (int) (temp.second * 1.2));
```
moving-bits, line 440: you don't need the intermediate step of creating `temp`, you can use `wantedSize * 1.2` directly in the second line

moving-bits:
> @moving-bits can I merge this one? I want to build up on it.

I gave it a (short) glance yesterday, but did not fully understand all aspects, and had not found the time to test locally. But anyway, we are still in an early dev phase (no urgent need to release from `master`), and the change is local, so I would vote for merging, and I'll try to give it a more thorough look throughout the next days.

eddiemuc:
Ok, thanks a lot for the (quick) review. I will correct your finding and merge afterwards

main/src/cgeo/geocaching/utils/EmojiUtils.java
eddiemuc: done

eddiemuc:
retest this please

eddiemuc:
failing test is unrelated, thus merging

Issue #10885, opened 2021-06-07
Add possibility to clear input in text edit field


murggel:
## What function are you missing

For some input fields, it would be nice to have an easy possibility to clear the (default) text - e.g. default coordinates in input for plain format in calculator (normally I want to paste the formula from clipboard), default name for new waypoints, ...

## Suggested solution

Would it be possible with the new material-theme to add a clear button?
![20210607_232628.jpg](https://user-images.githubusercontent.com/70113341/121090498-f879bd00-c7e8-11eb-8cc3-3ee5af5f99df.jpg)

fm-sys:
Alternative solution would be to select the whole text when the input fields gets clicked...

murggel:
My hope would be, introducing / using the textInputField #10763 comes probably with a configuration for a clear button.

https://material.io/components/text-fields/android#using-text-fields

But I did not step deeper into it...

eddiemuc:
Material.io do indeed come with a standard clear option, and I think it would be a great idea to use them in c:geo in many places

eddiemuc:
@murggel Material.io supports something like this:
![image](https://user-images.githubusercontent.com/6909759/121130460-b7a49700-c82e-11eb-8029-0888a7c1ea12.png)
(this is not an outlined text input field like I assume we want to use, but it shows the "clear" icon used by material.io).

Is this what you have in mind?

murggel:
> @murggel Material.io supports something like this:
> ![image](https://user-images.githubusercontent.com/6909759/121130460-b7a49700-c82e-11eb-8029-0888a7c1ea12.png)
> (this is not an outlined text input field like I assume we want to use, but it shows the "clear" icon used by material.io).
> 
> Is this what you have in mind?

Exactly. I was looking for something like that.

Do we have to change than all the fields or what do you mean with "this is not an outlined text input field like I assume we want to use"?

eddiemuc:
> Do we have to change than all the fields or what do you mean with "this is not an outlined text input field like I assume we want to use"?

As far as I understand, then in  #10763 we want to change all text input fields to material.io (which uses com.google.android.material.textfield.TextInputLayout and inner EditText views). So I fear we have to exchange all text input fields in c:geo in this issue. But maybe not in a single go but one after the other...

[...]

Issue #10763, opened 2021-05-26
Material Design: Make use of TextInputFields


fm-sys:
Make use of TextInputFields

https://material.io/components/text-fields/android

![grafik](https://user-images.githubusercontent.com/64581222/119679948-12a6b900-be41-11eb-89f3-2805806c9171.png)

This should IMHO be used for all TextInputs (search, coordinate input, ...), as the inputs are currently nearly invisible...

![grafik](https://user-images.githubusercontent.com/64581222/119680662-aed0c000-be41-11eb-85dd-bd53b182ea4a.png)

eddiemuc:
Note that in PR #10875 I introduced such a new textfield and created a first style "`textinput_dropdown`" for it. From screenshot above I concluded that you would like to use "Outlined-Boxes", which I did. Relevant resources:
* style.xml
* cache_filter_generic_string.xml

Note that I used a AutoCompleteTextView variant, but other variants can be added too of course. Also fine-tuning of the style might become necessary, but as a first draft I think it looks quite good!

Empty, not-yet-used textfield:
![image](https://user-images.githubusercontent.com/6909759/120932271-6644bc80-c6f5-11eb-90df-c3203dea6ff1.png)

After selecting and typing "my text":
![image](https://user-images.githubusercontent.com/6909759/120932301-807e9a80-c6f5-11eb-8b8d-24784151883a.png)

After focus leaves text field:
![image](https://user-images.githubusercontent.com/6909759/120932315-90967a00-c6f5-11eb-9902-cbfd38198183.png)

eddiemuc:
Ah, and I set the label "Search Text" on the text input field

Lineflyer:
Offtopic:
That reminds me, that we should use this styles of input fields in all other places as well. The current style with underline accent only looks kind of "broken". Do we have an issue for that already?

Lineflyer:
Forget my comment. In fact this is the issue I was looking for :D   I am kind of confused today.

moving-bits:
I'm trying to migrate the coordinates input dialog currently. First steps went smooth, but depending on input type I have the problem that layout padding inside the outline box is so large, that the numbers itself are barely visible (e. g.: see the cut-off "2" in the last three edit fields in the upper row). 

Any ideas on how to reduce it? No variant of padding* or margin* attributes have worked for me, neither the "Dense" variant.

[...]

PR #10887, opened 2021-06-08
use material TextInputFields for mainscreen searches (rel. to #10763)

moving-bits:
## Description
Use material TextInputFields with acticated "clear input" button for our searches on mainscreen:

image

Related to #10763 and #10885


fm-sys:
Looks good, like it! But shouldn't the TextInputFields use the same horizontal margins as the action buttons do?

PR #10890, opened 2021-06-08
use material TextInputLayout for edit waypoint page (rel. to #10763)

moving-bits:
## Description
- use material TextInputLayout for edit waypoint page 
- replace old viewbindings (colliding with Gradle plugin 5) with ViewBinding class

|default|multi-line edit|with bearing + distance|
|---|---|---|
|image|image|image|

(First two screenshots are from editing a given waypoint, third screenshot is from creating a user-defined waypoint)

## Additional context
- use style `textinput_dropdown` if you are embedding a single line `AutoCompleteTextView`
- use style `textinput_edittext` if you are embedding a single or multi line `EditText`
- use style `textinput_embedded` for the embedded `AutoCompleteTextView` or `EditText`

## Note
If you want to set visibility to an embedded `AutoCompleteTextView` or `EditText`, you will have to set the visibility of the surrounding `TextInputLayout`, otherwise it will not work.

### DIFF
--- a/main/res/layout/cache_filter_generic_string.xml
+++ b/main/res/layout/cache_filter_generic_string.xml
@@ -53,25 +53,13 @@
 
     <com.google.android.material.textfield.TextInputLayout
         android:id="@+id/searchtext"
-        android:hint="@string/cache_filter_stringfilter_searchtext_hint"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
         style="@style/textinput_dropdown"
-        app:endIconMode="clear_text"
         android:layout_below="@+id/filter_options"
         android:layout_alignParentLeft="true"
         android:layout_alignParentRight="true">
-
-    <AutoCompleteTextView
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:inputType="text"
-
-        />
-
+        <AutoCompleteTextView
+            style="@style/textinput_embedded"
+            android:hint="@string/cache_filter_stringfilter_searchtext_hint" />
     </com.google.android.material.textfield.TextInputLayout>
-
-
-
 
 </RelativeLayout>
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -198,23 +198,22 @@
         <item name="android:autoText">true</item>
         <item name="android:gravity">top|left</item>
         <item name="android:capitalize">none</item>
-    </style>
-
-    <style name="textinput_dropdown_full" parent="textinput_dropdown">
-        <item name="android:layout_width">fill_parent</item>
+        <item name="android:layout_width">match_parent</item>
         <item name="android:layout_height">wrap_content</item>
+        <item name="endIconMode">clear_text</item>
     </style>
 
-    <style name="textinput_edittext_full" parent="@style/Widget.MaterialComponents.TextInputLayout.OutlinedBox">
+    <style name="textinput_edittext" parent="@style/Widget.MaterialComponents.TextInputLayout.OutlinedBox">
         <item name="android:padding">6dip</item>
         <item name="android:autoText">true</item>
         <item name="android:gravity">top|left</item>
         <item name="android:capitalize">none</item>
-        <item name="android:layout_width">fill_parent</item>
+        <item name="android:layout_width">match_parent</item>
         <item name="android:layout_height">wrap_content</item>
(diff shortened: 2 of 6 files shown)


main/res/layout/cache_filter_generic_string.xml
```diff
@@ -54,8 +54,6 @@
     <com.google.android.material.textfield.TextInputLayout
         android:id="@+id/searchtext"
         android:hint="@string/cache_filter_stringfilter_searchtext_hint"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
         style="@style/textinput_dropdown"
         app:endIconMode="clear_text"
```
eddiemuc, line 58: Is it possible to add "app:endIconMode="clear_text" also to the "textinput_*" style?

main/res/layout/cache_filter_generic_string.xml
```diff
@@ -54,8 +54,6 @@
     <com.google.android.material.textfield.TextInputLayout
         android:id="@+id/searchtext"
         android:hint="@string/cache_filter_stringfilter_searchtext_hint"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
         style="@style/textinput_dropdown"
```
eddiemuc, line 57: Isn't it necessary to add the new style "textinput_embedded" also to the enclosed "Autocompletetextview" here?

main/res/values/styles.xml
```diff
@@ -198,14 +198,26 @@
         <item name="android:autoText">true</item>
```
eddiemuc, line 198: "textinput_dropdown" seems to be the same as "textinput_edittext_singleline" currently. I support the idea to keep the separate style for the future, but couldn't it just be left empty with parent="textinput_edittext_singleline" for now?

eddiemuc (approved):

eddiemuc:
@moving-bits thanks for the integration of generic styles for material.io text inputs! I vote for a quick merge so we can build upon it.

main/res/layout/cache_filter_generic_string.xml
moving-bits: Yes, it should - seems I have forgotten it - too many iterations yesterday... Will add it tonight.

main/res/layout/cache_filter_generic_string.xml
moving-bits: I have tried that, but the compiler complained about not knowin "app:endIconMode" attribute in `styles.xml`, so I keept it in the layout files.

main/res/values/styles.xml
moving-bits: Inheritance is a bit different: `Widget.MaterialComponents.TextInputLayout.OutlinedBox.ExposedDropdownMenu` is parent for `textinput_dropdown`, whereas `Widget.MaterialComponents.TextInputLayout.OutlinedBox` (without `ExposedDropdownMenu`) is parent for `textinput_edittext`. 
I had tried using `textinput_dropdown` for the `EditText` definitions as well first, but that did not work for me, I did not get the colored border, so I started searching for other style templates and ended up with that one.

main/res/layout/cache_filter_generic_string.xml
eddiemuc: I googled this quickly and found this: https://stackoverflow.com/questions/39336054/using-app-namespace-in-style
It seems that for "app" attributes in styles, you simply leave out the "app:" praefix

main/res/values/styles.xml
eddiemuc: You're right, I have overlooked that. Sorry, should have known...

moving-bits:
> @moving-bits thanks for the integration of generic styles for material.io text inputs! I vote for a quick merge so we can build upon it.

Thanks for taking time to review. I'll try to fix the remaining issues later today, and then we can merge.

main/res/layout/cache_filter_generic_string.xml
moving-bits: done

main/res/layout/cache_filter_generic_string.xml
moving-bits: Yeah, should have known that. Have used that many times in the past, but it did not come into my mind yesterday... :-(

Adjusted the styles accordingly.

PR #10894, opened 2021-06-09
introduce SimpleDialog and TextParam, replace Dialogs

eddiemuc:
This PR introduce SimpleDialog and TextParam and replaces (in large parts) the Dialogs class

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
@@ -1,0 +1,473 @@
+package cgeo.geocaching.ui.dialog;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.Keyboard;
+import cgeo.geocaching.ui.TextParam;
+import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.utils.functions.Action2;
+import cgeo.geocaching.utils.functions.Func2;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.text.Editable;
+import android.text.InputType;
+import android.text.TextWatcher;
+import android.view.Gravity;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.EditText;
+import android.widget.LinearLayout;
+import android.widget.ListAdapter;
+import android.widget.TextView;
+
+import androidx.annotation.LayoutRes;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
+import androidx.appcompat.app.AlertDialog;
+import androidx.core.util.Consumer;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import static java.lang.Boolean.TRUE;
+
+import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Builder-like class for simple dialogs based on {@link AlertDialog}.
+ * <p>
+ * If you want to show e.g. a simple text or confirmation dialog, use this class.
+ */
+public class SimpleDialog {
+
+    /**
+     * Convenience constant for button actions to do nothing (and close the dialog)
+     */
+    public static final DialogInterface.OnClickListener DO_NOTHING = (d, i) -> {
+    };
+
+    /**
+     * Define common button text sets
+     */
+    public enum ButtonTextSet {
+        OK_CANCEL(TextParam.id(android.R.string.ok), TextParam.id(android.R.string.cancel), null),
+        YES_NO(TextParam.id(R.string.yes), TextParam.id(R.string.no), null);
+
+        public final TextParam positive;
+        public final TextParam negative;
+        public final TextParam neutral;
+
+        ButtonTextSet(final TextParam positive, final TextParam negative, final TextParam neutral) {
+            this.positive = positive;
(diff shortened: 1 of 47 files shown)


eddiemuc:
This PR replaces large parts of class `Dialogs `with a new Builder-like class `SimpleDialog `to create simple dialogs (message, conform, input, select single/multiple items) in a unified way. This brings the following benefits:

* less code / removal of code duplication wrt Dialog creation throughout whole c:geo
* more unified way to create and maintain this kind of dialogs
* methods with less parameters (builder-like fashion),  code should be more understandable now 
* More content options for all dialogs, e.g. using emojis as images (through ImageParam) and  using html/markdown/linkify/accompanying-icon as text (through TextParam). This can also be used outside of `SimpleDialogs `(e.g. in `ViewUtils`)

PR is now ready for review (and hopefully later merge)

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -0,0 +1,446 @@
...
+            builder.setPositiveButton(getPositiveButton(), listener[0]);
+            if (listener.length > 1 && listener[1] != null) {
+                builder.setNegativeButton(getNegativeButton(), listener[1]);
+            }
+            if (listener.length > 2 && listener[2] != null) {
+                builder.setNeutralButton(getNeutralButton(), listener[2]);
+            }
+        }
+        final AlertDialog dialog = builder.create();
+        dialog.setOwnerActivity((Activity) getActivity());
```
fm-sys, line 229: What is the purpose of this? If we leave out "setOwningActivity" (which is only necessary for sharing autio channels ec
tc) we could  make this whole  class based on `Context` instead of `Activity`...

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -0,0 +1,446 @@
+package cgeo.geocaching.ui.dialog;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.Keyboard;
+import cgeo.geocaching.ui.TextParam;
+import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.utils.functions.Action1;
```
fm-sys, line 7: What's the difference between our self created `Action1` and `Consumer`?

fm-sys (approved):
While the first approach (textParam inside dialogs class) looked like making things even more complicated, I really like the current idea of a builder like dialog creation. Looking forward 👍

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: I thought about this too and in fact first implemented it this way (the unnecessary cast above is an artefact of this which I forgot to remove - I can do this). 

Then I realized that at least currently there is not a single place in c:geo using Dialogs with a context  which is not also an activity, so I can really not tell if this all would work when context is not an activity.

What would be a use case for a dialog which is NOT in the context of an activity?

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: There is no difference at all. I assume that `Action1` was created either before `androidx.core.util.Consumer` existed or for naming and usage consistency towards `Action2` (for which there is no equivalent in `androidx.core.util`)

The deeper underlying problem here is that the real function classes of JDK under `java.util.function` can only be used starting with SDK Level 24 while c:geo runs on SDK level 21. There would also be a `java.util.function.Consumer` class, as well as e.g. `BiConsumer `(replacement for `Action2`), `Function`/`BiFunction` (replacement for `Func1`/`Func2`), `Predicate`and many many more. I think for now it doesn't really matter whether we use Action1 or `androidx.core.util.Consumer` because once we switch to SDK 24, we should anyway convert all usages of both `androidx.core.util` and `cgeo.geocaching.utils.functions` to standard JDK functions in `java.util.function` anyway.

But if you insist I can change `Action1 `to `androidx.core.util.Consumer`. Should I?

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: I now introduced a new creation method `ofContext` which just takes a context (not an Activity). If at any time there will be a use case for it, it can be used.

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: Ah, what the heck, I changed `Action1` to `Consumer`...

eddiemuc:
While I worked on @fm-sys 's comments, IU also simplified the setting of individual button texts (no need for the caller to use `TextPara  `any more if he just uses StringRes ids).

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
fm-sys: Yes, an alert dialog needs to be called from an activity. But there might be places where we have just the context object of the activity... 🤷‍♂️ Not sure, but anyway `ofContext` sounds like a good idea 👍

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
fm-sys: > But if you insist I can change Action1 to androidx.core.util.Consumer. Should I?

Up to you, I was just wondering if there's any difference. Maybe we should change all usages to one of both methods to be consistent in our code...? But that's definitely unrelated ;-)

PR #10899, opened 2021-06-10
implement bottom navigation

fm-sys:
# Draft

|grafik|grafik|grafik|
|-|-|-|
|grafik|grafik|grafik|

Feel free to extensively test the PR so that we have some discussion stuff for our meeting ;-)

Known things that need optimization:
- layout hierarchy handling (e.g. when map gets opened from compass view)
- further main activity layout refactoring 
- don't start c:geo with "more" as basic selection 
- reset activity if already selected bottom navigation icon gets clicked another time
- working mode of "show as list" / "show on map" should be changed to share the full collection instead of only the filtered results 
- ...and probably very much more


Lineflyer:
Just a quick note of something which might be known:
I freshly installed the PR based debug version and it repeatedly informed me about the "Long tap on map" feature each time I selected live map although I activated "dont bother me" in that dialog on first appearance.

Lineflyer:
And a last quick feedback for today:
- Select Nearby
- Click on a cache from the result list

It seems the cache details view is already shown (empty) in the background while the cache is loaded.
You seem some lines and empty elements in the background behind the download progress dialog.

fm-sys:
> And a last quick feedback for today:
> 
>     * Select Nearby
> 
>     * Click on a cache from the result list
> 
> 
> It seems the cache details view is already shown (empty) in the background while the cache is loaded.
> You seem some lines and empty elements in the background behind the download progress dialog.

This looks indeed strange, but I can reproduce the issue with our current nightly as well. So it's independent of this PR... I will open a separate issue.

Lineflyer:
> Just a quick note of something which might be known:
> I freshly installed the PR based debug version and it repeatedly informed me about the "Long tap on map" feature each time I selected live map although I activated "dont bother me" in that dialog on first appearance.

This behavior was gone after next start.

Lineflyer:
Some unsorted thoughts:

- The bottom navigation is very intuitive and IMHO a very good solution, only...
- We loose some display space on each activity as the bottom navigation has to be shown. While I don't consider this bad on stored caches, search nearby its however a bit wasted space on the map. Not sure though if there is any way around that. Probably not or is there some mechanism to let it slide down or move it to a minimized mode (only text or only icon)?
- We loose the long press function on stored (previously used to quickly jump to certain list). 
- The quick start option (long press the launcher icon) is missing stored and search now. That is BTW the actual good solution to directly start c:geo on a certain activity, which I use very often.

fm-sys:
Thanks for your testing effort, very valuable!

> We loose the long press function on stored (previously used to quickly jump to certain list).

Maybe it's possible to implement the long click functionality for bottom navigation... 

> We loose some display space on each activity as the bottom navigation has to be shown. While I don't consider this bad on stored caches, search nearby its however a bit wasted space on the map. Not sure though if there is any way around that. Probably not or is there some mechanism to let it slide down or move it to a minimized mode (only text or only icon)?

There's indeed this slide down possibility for scroll activities. But I'm afraid this could confuse users. Is screen space really an issue these days?

> The quick start option (long press the launcher icon) is missing stored and search now. 

Need to look into it...

Lineflyer:
> There's indeed this slide down possibility for scroll activities. But I'm afraid this could confuse users. Is screen space really an issue these days?

You are right, maybe its just a matter of getting used to it. People initially seeing this implementation will probably not bother about that, and indeed the tendency nowadays goes towards big screen size respectively high resolution.

Regarding the previous thoughts about the startup behavior:
As written above I would prefer some kind of dashboard which presents the information currently contained in the bottom status (active connector, login state, find count, gps info) in a nice (graphical) way. Of course users might opt-in to start with the last active screen instead.
Is that meaningful and feasible?

Lineflyer:
Example from "another" app:
image

moving-bits:
Thanks @fm-sys for this PR, looks quite promising already! Some random things I've stumbled upon:
- c:geo should not start with the "more" page. Maybe better with some sort of dashboard, or directly the map, or make it user-selectable
- "more" page uses a pretty big font and huge line spacing. I can see just 8 menu entries, whereas on my physical phone 8 entries only cover half of the screen's height.
- We should rethink the order of the menu entries for "more" page - which ones should be visible directly on opening, for which ones is it ok to scrool.
- we still have the same stacking order as today, so you can move from map to list to detail to a different map etc. and all the way back through the stack, which is both hard to understand for the user sometimes and leads to some unexpected behavior on returning to an "old page" further down the stack showing old data.

More to come ;-)

fm-sys:
Thanks for your testing as well! Many of these things I already have in mind, but are just not yet implemented. Other things probably need further discussion. I will not react to every point for now, but rather collect them for our meeting this Sunday. Then we can go through all these things one by one, which is probably easier then discussing them all in one thread ;-)

Lineflyer:
One other thought from a fellow cacher, to whom I presented the bottom navigation:

He somehow misses the "c:geo look and feel" as we use some default icons instead of those icons he is used to from the old main screen.
Maybe we should think about reusing the old main screen icons but with a more "flat" design (meaning single colored or alike)...

fm-sys:
Just a small update:

grafik

fm-sys:
@Lineflyer as we have decided to not yet merge this PR, would it be possible to configure a second nightly job as we once had for the ALC full branch?

Or isn't it possible as the branch is part of my cgeo fork? If so, should we create a `bottom_navigation` branch in this repo where I can (force) push my commits to for easy testing?

Lineflyer:
If that makes life easiere for you, we can/should create a temporary `bottom_navigation`branch on this repo.
At the end the difference to using your own branch (in your fork) is, that its more visible to others of the team and we can use the CI to create a nightly from it.

Just give me a "Go" and I can create a branch from your PRs state and (if needed) configure the temporary nightly.

fm-sys:
Yes, please go for it! 

I'm currently using a debug build with LeakCanary enabled on my primary device which is quite annoying, so an extra nightly would be awesome ;-)


PS: Note that I will continue to force push to that branch, the same as if it would be at my own fork. So if anyone wants to push to that branch as well, please give me a short hint.

Lineflyer:
Branch `bottom_navigation` created with HEAD at the current state of this PR (unless I did it wrong).
Nightly will be build each night at 1:30am and is available from https://download.cgeo.org/cgeo-nightly-BN.apk

Version differentiator is `NBBN`  e.g. 2021.06.14-NBBN-<commit> 

Manual additional building on demand can be done on the CI: https://ci.cgeo.org/job/cgeo-nightly_BN/

Force push to `bottom_navigation` should be possible for core members as far as I understand the settings on GitHub. Please tell me, if it does not work for you.

fm-sys:
Yes, branch looks fine and force pushing does work for me as well. Thanks for configuring the CI job 👍

Issue #10711, opened 2021-05-22
Revise Global Cache filter: make it based on "Basic" new filter system


eddiemuc:
## What function are you missing
The current global filter does allow only to select one cache type. More global filtering would be good.

## Suggested solution
It should be possible to use a "Basic" set of the new filter framework as global filter. This includes cache type filter (multiple types possible), difficulty and terrain. The "global" filter, if set, will be connected to any other filter (context-dependant) using AND-logic.

## Alternatives/workarounds
Don't implement it

Lineflyer:
Basically I agree to that proposal.
From the users community I often heard the wish to be able to globally select
- more than one cache type at a time
- A range of D and/or T

From the existing logic users are already aware, that this global filter is applied to the whole app, so indeed the internal AND logic is the right way, as:
- Users who only use the global filter can just use it in the same way as before
- Users who previously used e.g. the old list filter can use the new filtering and the combination also behaves in the same way

Question:
Shall the active global filter be somehow indicated in the "lower level" filters?
Example:
A user applied a D-rating filter on global level. Later he opens a list and starts filtering on list level. Shall the filter dialog somehow indicate, that there is a global filter for D-rating active?
While being on list level the global filter is indicated below the action bar. Maybe its just enough to show the same indicator also in the filter dialog on list level.

eddiemuc:
In context of this item I would really like to discuss how the relationship should be in the future of a "global filter" towards other "context-specific" filters in the future.

As of now, implementation is as follows:
* for each "context", the last used filter is stored. This filter is restored as starting point whenever a user goes to this "context" again. "context" is currently defined by the c:geo "CacheListType", currently there are 11 such types defined (see list below)
* Additionally there is a "global" filter where one can specify one cache type (only).
* In each context, both the "context" filter and the global filter is applied in an AND fashion (both have to be fulfilled.

Now the question is how this should be in the future. I see three principail options:
* Reduce to one (global) filter. In each context, this filter is used, and whenever it is modified in any context the result becomes the new "global" filter. It is used everywhere. (This would be the approach that e.g. gc.com app uses today)
* Stay with "context filter" only, do not provide an additional global filter
* Keep both: apply the "context-specific" filter as well as an (optional) global filter in an AND fashion (this would be as close as possible to the c:geo system today)
Before implementing this I kind of need to have a common understanding which "strategy" should be supported in the future.

[...]

PR #10909, opened 2021-06-12
rel to #10711: store only one global filter setting, no contexts any more

eddiemuc:
related to #10711: store only one global filter setting, no contexts any more

This PR will remove context-specific filter storage completely. There will only be one filter set stored (always the last one used) across all contexts: map, offline cache lists and online search results (nearby search, owner, user, search-by-filter).

This PR will not yet replace the current additional global cache type filter, this remains in parallel for now


eddiemuc:
This makes especially sense in conjunction with @fm-sys 's approach for a new botton navigation in #10679

fm-sys:
> This PR will not yet replace the current additional global cache type filter, this remains in parallel for now

Thanks for not yet changing anything about the global cache type filter! That saves me quite some merge conflict hustle 😅

PR #10913, opened 2021-06-12
fix #8377, fix #10614: status filters for stored, favs, watchlist, premium and coords changed

### LINKED ISSUE #8377: FIlter for caches which I favorited myself
I want to see which caches on my found list I gave a favorite point to

There should be a filter in Filter/Favorites or Filter/With own information for this

None in c:geo

**Additional context**
From support mail

### LINKED ISSUE #10429: Option to hide unmodified (!) Mysterys from map view

I would like there to be a filter in the map view to hide all mystery caches that do NOT have modified coordinates. The reason being that unsolved mysteries are not located where their coordinates say anyway, so they clutter up the map unnecessarily.

Mysteries with modified coordinates have essentially been turned into traditionals, and the icon really shows where they are, so they should still be displayed on the map.

### LINKED ISSUE #10614: Filter to show only "unsaved" caches on the map
in the context of integrating Adventure Lab Caches I wanted a way to easily visualize new Adventures being published. I thought I would set the global filter to just ALCs, then go into Live Map, select a specific area and use "Store all unsaved" into some list. First time I do this, it would take everything on the map since nothing was saved before.

I would then come back later, perhaps next day and still filtering just for ALCs and say show me on the live map all objects that I haven't already stored. Just like "store all unsaved", just don't show me anything else on the map such that I get an idea or a "poor man's notification" of new stuff coming up in my area.

Does this perhaps already exist and I just haven't found it yet?

eddiemuc:
fix #8377, fix #10614, fix #10429: status filters for stored, favs, watchlist, premium and coords changed

This PR adds more status filter options to the already existing status filter.

Already existing:
* Found/NotFound
* Owned/Not Owned
* Active/Disabled/Archived

New:
* Stored/Not Stored
* Favorited/Not Favorited
* On Watchlist/Not on watchlist
* Premium/Basic
* Coords Changed / Coords not changed


eddiemuc:
The filters seem to work. I am however not quite happy with the styling myself. Reason is that the very big font size in toggle buttons forces me to use VERY short wording even on very wide screens:

image

Does anyone have a nice UX idea for this dilemma?

fm-sys:
I guess "c. original" means "without changed coordinates". But "c" is probably not easy to recognize as "coordinates" .

What about making a subtitle "coordinates" and below the options "all", "original/not changed" and "changed". Could be an idea for the other items as well... 

BTW, "is favorite" will hopefully only be displayed for PM's, correct?

eddiemuc:
> BTW, "is favorite" will hopefully only be displayed for PM's, correct? 

I don't think this would make sense:
* Afaik, favorites are also available on other services (e.g. "recommendations" on oc) where there's no concept of "premium"
* Filter might be applied to previously stored caches while logged in with another user

eddiemuc:
> What about making a subtitle "coordinates" and below the options "all", "original/not changed" and "changed". Could be an idea for the other items as well... 

That would work but take up even more space than already

eddiemuc:
Ok my current idea is to create a unique set of three buttons "Only", "Include" and "Exclude" and use it for each of the options. Then placing a label describing the option in front of it.

For example the row for "Found" would look like this:

`Found:.  Only|Include|Exclude`

Choosing "Only" would filter only found caches. Choosing "exclude" would filter only not found caches. Choosing "include" would not filter at all (both found and not found caches are in the result)

Would that be a feasible option?

And does anyone maybe have a better idea for the button wording or (even better) nice icons for icon buttons instead?

fm-sys:
Sorry to say that, but " Only|Include|Exclude" is IMHO even worse to understand :-/

eddiemuc:
Most recent screenshot:
image

eddiemuc:
I will change the button to All | Yes | No (and put labels in front of them) as discussed yesterday

eddiemuc:
I believe this PR is ready for review / merge now. Here is the new screenshot:
image

fm-sys:
Just another idea:

Exclude active, disabled and archived could be a (full width) multi selection toogle group as well:

    Active|Disabled|Archived

(the checked ones should be those  which are included, as it's otherwise a double negative which IMHO might be confusing)

eddiemuc:
> Just another idea:
> 
> Exclude active, disabled and archived could be a (full width) multi selection toogle group as well:
> 
> ```
> Active|Disabled|Archived
> ```
> 
> (the checked ones should be those which are included, as it's otherwise a double negative which IMHO might be confusing)

Technically no problem, but will the users realize that this is a multi-select (although it looks the same as the single-select-togglebars)?

fm-sys:
> Technically no problem, but will the users realize that this is a multi-select (although it looks the same as the single-select-togglebars)?

Probably yes, as we would preselect all three states by default. At least that would be less space consuming than the checkbox approach. 

Or we could use filter chips as well (like for cache size filter). As more as I think about it, I actually even prefer this option... 

But we can wait for further feedback for sure...

eddiemuc:
> Or we could use filter chips as well (like for cache size filter). As more as I think about it, I actually even prefer this option...

I like the idea very much! But would it be ok to do this in a separate PR and merge this one already?

main/res/values/strings.xml
```diff
@@ -554,10 +555,17 @@
...
+    <string name="cache_filter_status_select_no">No</string>
+
+    <string name="cache_filter_status_select_label_found">Found</string>
+    <string name="cache_filter_status_select_label_owned">Owned</string>
+    <string name="cache_filter_status_select_label_stored">Stored</string>
+    <string name="cache_filter_status_select_label_favorite">Favored</string>
+    <string name="cache_filter_status_select_label_watchlist">On Watchlist</string>
+    <string name="cache_filter_status_select_label_premium">Premium</string>
+    <string name="cache_filter_status_select_label_solved_mystery">Solved Mystery</string>
+    <string name="cache_filter_status_select_infotext_solved_mystery">This filter setting will only affect mystery caches.\n\n A mystery is considered \'solved\' if it has either changed coordinates (gc.com only) or a valid final waypoint filled with coordinates.</string>
```
moving-bits, line 568: we should write the domain name / service name as `geocaching.com` - `gc.com` is something different (although we use it frequently as abbreviation)

eddiemuc: Very true. I have removed the "like gc.com" phrase

eddiemuc:
retest this please

eddiemuc:
With chips it will look like this:
image

moving-bits:
failing test is unrelated, but there's a new lint error about an unused layout resource `dialog_textview.xml`

eddiemuc:
> failing test is unrelated, but there's a new lint error about an unused layout resource `dialog_textview.xml`

Don't really understand why this is coming up with this PR (it seems it was also unused before), but I deleted it

eddiemuc:
Failing tests are unrelated, let's merge

moving-bits:
> > failing test is unrelated, but there's a new lint error about an unused layout resource `dialog_textview.xml`
> 
> Don't really understand why this is coming up with this PR (it seems it was also unused before), but I deleted it

my bad - while working on some lint messages (see #10942 I recognized that this error is unrelated to your PR. Should have updated my comment, sorry.

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
```diff
@@ -39,18 +56,40 @@
...
-            (statusOwn == null || (cache.isOwner() == statusOwn)) &&
-            (statusFound == null || cache.isFound() == statusFound);
+            (statusOwned == null || (cache.isOwner() == statusOwned)) &&
+            (statusFound == null || cache.isFound() == statusFound) &&
+            (statusStored == null || cache.isOffline() == statusStored) &&
+            (statusFavorite == null || cache.isFavorite() == statusFavorite) &&
+            (statusWatchlist == null || cache.isOnWatchlist() == statusWatchlist) &&
+            (statusPremium == null || cache.isPremiumMembersOnly() == statusPremium) &&
+            (statusSolvedMystery == null || cache.getType() != CacheType.MYSTERY ||
+                (cache.hasUserModifiedCoords() ||
```
murggel, line 90: @eddiemuc there is already a function in Geocache.java to check that. See GeoCache.showUserModifiedCoords. That should give you the desired information?

eddiemuc: I am very sorry, I currently don't understand  your remark.

In above code, `cache` is a `Geocache` object, and I use its method `hasUserModifiedCoords()` for my check. I couldn't find a method `showUserModifiedCoords()` though. 

Method `hasUserModifiedCoords()` does not check for valid final coords, I use `Geocache.getFirstMatchingWaypoint()` for that (which I created as part of this PR).

How do you think this should be done instead?

murggel: Ah ok, my mistake, thought it was a function of the class Geocache, but it is a private one in MapMarkerUtils: showUserModifiedCoords

There is a method in Geocache.hasFinalDefined, then you don't have to lookup that ...

eddiemuc: Ah, you're right, did not see this one. Good catch, thank you. I will change this in one of my next PRs (since this one is already merged)

PR #10928, opened 2021-06-14
WIP - implement bottom navigation

fm-sys:
This draft PR is more or less a todo list for myself, what still needs to be done for the bottom navigation. We have decided to not yet include the bottom navigation into our first beta, so we have more time for fine-tuning the new navigation component.

Feel free to test the "bottom navigation nightly" from time to time, which is compiled out of this PR branch. As this special nightly build uses the same signing key as our "default nightly", it should be possible to switch between both versions back and forth: 
[🔗 Nightly APK](https://download.cgeo.org/cgeo-nightly-BN.apk)  [🔗 CI job](https://ci.cgeo.org/job/cgeo-nightly_BN/)

# Todo
- [ ] hierarchy handling/stacking order (e.g. when map gets opened from compass view)
- [ ] further improving layout of "more" page
- [ ] move initialization out of main activity
- [ ] don't start c:geo with "more" as basic selection (dashboard?!)
- [ ] update/status notification needs to be covered somewhere
- [ ] reset activity if already selected bottom navigation icon gets clicked another time
- [ ] long press function on stored (used to quickly jump to certain list)
- [ ] order of the menu entries for "more" page - which ones should be visible directly on opening, for which ones is it ok to scroll?

...and probably more to come...


moving-bits:
Tested it again today with regards to the status info (see https://github.com/cgeo/cgeo/discussions/10679#discussioncomment-923781 for a response to this) - looks pretty nice already. One proposal though: In light mode the bottom navigation is hard to separate from the window content. Maybe we can set the same background color here as we do in dark mode? (Similar to what is done automatically for the tabbed viewpager.)

|view|dark mode|light mode|
|---|---|---|
|tabbed view|image|image|
|bottom navigation|image|image|

fm-sys:
> Tested it again today with regards to the status info (see #10679 (reply in thread) for a response to this) - looks pretty nice already. 

How should I interpret this post? Is combining "more" and "connector status" ok for you, or would you still appreciate it, if things would be separated? 

IMHO it's not enough info for filling a whole separate activity though....

fm-sys:
>  In light mode the bottom navigation is hard to separate from the window content. Maybe we can set the same background color here as we do in dark mode?  

Yes, that's indeed inconsistent. Although I wonder if it would look more modern, if we would use a different actionbar color instead. Personally, I don't really like this middle-light gray... 

E.g. how Signal is doing it:

Screenshot_20210626_170214
Screenshot_20210626_170319

moving-bits:
> > Tested it again today with regards to the status info (see #10679 (reply in thread) for a response to this) - looks pretty nice already.
> 
> How should I interpret this post? Is combining "more" and "connector status" ok for you, or would you still appreciate it, if things would be separated?

Ah, my bad, two entangled infos in the same sentence. I meant "I did a quick test today, and it looks pretty nice already" and "reason for retest today was status info, my reply is in the other thread".

> IMHO it's not enough info for filling a whole separate activity though....

I'll copy this text to the other thread and reply over there.

moving-bits:
> Although I wonder if it would look more modern, if we would use a different actionbar color instead. Personally, I don't really like this middle-light gray...

I'm not fixed to the middle-grey, but I would avoid using pure black or pure white for actionbar / navigation bar - that's too colorless, IMHO, too much blending of c:geo into the system's infobar (is that its name?). Most apps I know use a more vivid color for those bars (sort of "brand color"), which is also neither black nor white. That's why I started with our accent color in the beginning, but I agree that may be bit too much orange :-)

fm-sys:
> 
> 
> > Although I wonder if it would look more modern, if we would use a different actionbar color instead. Personally, I don't really like this middle-light gray...
> 
> I'm not fixed to the middle-grey, but I would avoid using pure black or pure white for actionbar / navigation bar - that's too colorless, IMHO, too much blending of c:geo into the system's infobar (is that its name?). Most apps I know use a more vivid color for those bars (sort of "brand color"), which is also neither black nor white. That's why I started with our accent color in the beginning, but I agree that may be bit too much orange :-)

Let's continue this discussion in #11014. For now, I will use the same color for bottom navigation as for the action bar...

fm-sys:
An idea which just came to my mind after the meeting: 
Refactor the PR so that the `MainActivity` stays untouched and "more" is a completely different Activity. This would indeed allow us to merge the bottom navigation stuff with a hidden debug flag into master, and we could even release a new version with that hidden flag included, while the old functionality wouldn't be affected at all.

I guess, this wouldn't be too much work for me, and would help us to develop the bottom navigation over a longer time period at master, while still being able to release at any time... Would this be a way to go?

Time-wise, I would suggest doing so in about 4-6 weeks when I have more time for coding again, so if there should be any urgent regressions (which I wouldn't expect) I can fix them quickly.

moving-bits:
If activity separation is not too much work, this sounds like a viable way.

fm-sys:
Separating these activities should not be too much work. Luckily, the main activity is quite light. 

OTOH, reverting the main activity completely would be a step backward and would not result in a clean bottom navigation experience. But I already have some ideas at hand how such a dashboard could be implemented in a clean way (separated from the more list). This would result in a bottom navigation implementation which cannot be disabled completely, but might make even more sense and is not much work than reverting the main activity. 

My idea is to replace the "6 dots" with a bottom navigation view which has nothing selected as dashboard. This would give us more space to design a great looking dashboard ;-)


PS: Don't be confused by my random and continuously changing thoughts. This is more or less a talk to myself which helps me sorting my ideas. But I will try to provide screenshots of my ideas as soon as time permits.

Lineflyer:
> PS: Don't be confused by my random and continuously changing thoughts. This is more or less a talk to myself which helps me sorting my ideas. But I will try to provide screenshots of my ideas as soon as time permits.

All good, thats what we need and what makes a great result :)

fm-sys:
Just a small note: Please don't remove the dev status info (yet). It's quite helpful for working on the bottom navigation... (thanks @Lineflyer ;))

Lineflyer:
I thought already, that you might need it :)

fm-sys:
Splitting up the activity might take up to 5-6 weeks longer. Will not be able to code in the next time... 

Don't know if we want to remove the dev status message for the time being, OTOH it probably doesn't  hurt either..?

fm-sys:
Just found another app, which allows selecting the startup screen: OctoDroid.

Screenshot_20210722_225751_com gh4a_copy_270x585

We have that many settings in c:geo, so that one more setting should be acceptable as well. Many settings are IMHO totally useless, while this one would really give additional value to the user...

fm-sys:
Phew! At least I managed to rebase this branch onto the current master... We can discuss this Sunday how to continue with the bottom navigation.

fm-sys:
Some screenshots to keep you up to date. If c:geo cannot log in, a read warning will redirect the user to the "Dashboard":

Screenshot_20210915_070140_cgeo geocaching_copy_432x253
Screenshot_20210915_070055_cgeo geocaching_copy_431x374
Screenshot_20210915_070537

fm-sys:
Today, I turned my phone into landscape for the first time. While space is absolutely no problem in portrait mode, it's IMHO critical for landscape. 

Material.io suggests using a navigation rail instead for landscape/large-screen-devices:

Screenshot_20210917_162824_cgeo geocaching_copy_585x270
Screenshot_20210917_163719_org mozilla firefox

What do you think? Is this a way to go?

fm-sys:
BTW, these is even an abstract implementation of a navigation bar (`NavigationBarView`) that can be used to implement both components in parallel (Bottom Navigation and Navigation rail) sharing the same code base, which makes it extremely easy to maintain.

moving-bits:
> Material.io suggests using a navigation rail instead for landscape/large-screen-devices:

So this would put the bottom navigation to the left, gaining some space at the bottom, and the action bar would stay above the map? => LGTM.

fm-sys:
Here we go...

grafik

fm-sys:
Rebased onto current master...

PR #10932, opened 2021-06-14
improve svg

### LINKED ISSUE #10896: [SVG] Different icon for Backup / Restore

We have different icons for the backup / restore function.

In the action bar iof the main screen it is the **disc**-symbol:
grafik. 
Clicking on it, the backup / restore section of the settings will be opened.

There we have a **refresh**-symbol:
grafik

Those two symbols should be the same.

## Suggested solution

The "official" / material.io symbol for backup / restore is the following:
grafik

It looks similiar to the icon of the menu-entry "Personal Note". Probably we could find a better icon for that function (pencil)? But then it has the same icon as the menu-entry "Log Visit". But hey, how likely is it, that both are visible the same time in the action bar (it's the 3. and 6. menu-entry)?

## Alternatives/workarounds

grafik
But this is similar to the icon for "History". In landscape-mode both are visible at the same time (see screenshot above)
grafik

### LINKED ISSUE #10915: [SVG] Wrong color for log_image_icon and other optimizations
### Wrong color for log_img_icon

In dark mode, the log-image has black color on dark background instead of light color (GC7DY7J)
grafik

Recognized that while creating screenshot for #10769
Log of DL3BZZ (16.04.2021) has an attached image 

-> The tint-attribute in the svg is missing

### use only absolute path in SVG
to avoid possible bugs

### replace track.png with svg
created a svg similar to the existing one. Did not found yet a good alternative at material.io

murggel:
## Description

* add tint for actionBar (as workaround, until replaced with toolbar)
* add tint for log_img_icon 
* use svg for track for ic_menu_hidetrack
* use absolute coordinates in svg
* use backup-icon for restore

## Related issues

fix #10915 
fix #10896 

## Additional context

image
image


fm-sys (approved):
LGTM. If no one stops me, I will merge the PR this evening...

PR #10933, opened 2021-06-15
Fix wrong text color in about activity (text "carnero" was invisible)

fm-sys:
Before (string "canero" is black on black when using light theme):

grafik


moving-bits:
Thanks @fm-sys, I had missed (not seen) that string when fixing the color of the quote.

PR #10934, opened 2021-06-15
migrate individual route export dialog to material

moving-bits:
## Description
Migrates the "Export individual route as GPX" filename dialog to material `TextInputLayout`

## Question
Does work generally, but the dialog actually shrinks on entering the text input field, leading to info text below the input field being cut off. - Any ideas?

image.image


moving-bits:
failing test is related to website change & unrelated to this PR, so let's merge

PR #10937, opened 2021-06-15
Switch from ImageButton to Button

murggel:
## Description

Switching from ImageButton to Button, so the border of the icon-button is shown:
image
image

## Related issues

rel #10769 

## Additional context

I left the folder and map - element without border:
image

I left the "reorder"-button without border:
image

I still have problems with buttons, for which the icon ist set programatically:
image
image

### DIFF
--- a/main/src/cgeo/geocaching/settings/ViewSettingsActivity.java
+++ b/main/src/cgeo/geocaching/settings/ViewSettingsActivity.java
@@ -19,7 +19,6 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
 import android.widget.EditText;
-import android.widget.ImageButton;
 import android.widget.ListView;
 import android.widget.RadioButton;
 import android.widget.RadioGroup;
@@ -39,6 +38,7 @@ import java.util.Map;
 
 import javax.annotation.Nullable;
 
+import com.google.android.material.button.MaterialButton;
 import org.apache.commons.lang3.StringUtils;
 import org.xmlpull.v1.XmlPullParserException;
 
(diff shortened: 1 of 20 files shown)


main/src/cgeo/geocaching/settings/ViewSettingsActivity.java
```diff
@@ -118,8 +118,8 @@ public View getView(final int position, final View convertView, @NonNull final V
             View v = convertView;
             if (null == convertView) {
                 v = getLayoutInflater().inflate(R.layout.twotexts_twobuttons_item, parent, false);
-                ((ImageButton) v.findViewById(R.id.button_left)).setImageResource(R.drawable.ic_menu_edit);
-                ((ImageButton) v.findViewById(R.id.button_right)).setImageResource(R.drawable.ic_menu_delete);
+                ((Button) v.findViewById(R.id.button_left)).setCompoundDrawablesRelativeWithIntrinsicBounds(R.drawable.ic_menu_edit, 0, 0, 0);
```
fm-sys, line 121: Thanks for stepping into this issue! 

I read somewhere that there's a method `Button.setIconResource(R.drawable.ic_whatever)`. Don't know if it's part of the Button or MaterialButton class. However every button can simply be casted to MaterialButton inside the Java code... Maybe that solves the issue. 

If not there's a style `Widget.MaterialComponents.Button.OutlinedButton.Icon` from which our icon button style should be inherited from.

>In order to get the correct icon padding, it is recommended that you use a *.Icon style variant

murggel: Thanks, Casting to MaterialButton worked, but unfortunately not for the clear-icon in filter-view (AttributesFilterViewHolder) 😲
Inherit icon style from `Widget.MaterialComponents.Button.OutlinedButton.Icon` does not work either. 

Other ideas?

eddiemuc: > Thanks, Casting to MaterialButton worked, but unfortunately not for the clear-icon in filter-view (AttributesFilterViewHolder) 😲
> Inherit icon style from `Widget.MaterialComponents.Button.OutlinedButton.Icon` does not work either.
> 
> Other ideas?

I assume the difference between the button in AttributesFilterViewHolder and all others is that the button in `AttributesFilterViewHolder` is created via code, right?

I had styling problems with material.io before when creating view elements programatically. I solved them by creating simple layout files just containing one element. For example, I created layout `button_view.xml` which just contains a base-formatted `MaterialButton`. When I create the button by inflating this layout programatically, then later layouting seems to work correctly (and applies also other material styles correctly).

I assume `AttributesFilterViewHolder` simply needs to be changed to inflate the button in question out of this layout file or a similar one for Icon buttons.

murggel: > I assume the difference between the button in AttributesFilterViewHolder and all others is that the button in `AttributesFilterViewHolder` is created via code, right?

Right 

> I assume `AttributesFilterViewHolder` simply needs to be changed to inflate the button in question out of this layout file or a similar one for Icon buttons.

Thanks, I will try that

murggel: Great! Worked!
image

Thanks @eddiemuc

fm-sys: Looks good, thanks for your work 👍🎉

The PR is currently still set to draft, anything which needs to be done or is it ready?

eddiemuc:
The two failing tests are unrelated (one will be fixed if you rebase master), so they are no reason for not merging

murggel:
There will be merge-conflicts with PR #10938 in CacheDetailActivity
Once one PR is merged I will resolve them

murggel:
Will replace PR #10928

PR #10941, opened 2021-06-16
migrate SimpleDialog.input to material TextInputLayout

moving-bits:
## Description
migrate SimpleDialog.input to material TextInputLayout

image.image.image

The last one is not yet perfect, although it's using identical source code, but as settings are not based on `AppCompatActivity` I guess there's not much we can do about it.

@eddiemuc 
Can you cross-check if I haven't broken anything in `SimpleDialog.input`? Thanks.

### DIFF
--- a/main/res/layout/dialog_edittext.xml
+++ b/main/res/layout/dialog_edittext.xml
@@ -1,21 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<com.google.android.material.textfield.TextInputLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    android:orientation="vertical"
-    tools:context=".export.IndividualRouteExport" >
-
-    <com.google.android.material.textfield.TextInputLayout
-        android:id="@+id/input_frame"
-        style="@style/textinput_edittext_singleline"
-        app:endIconMode="clear_text">
-        <EditText
-            android:id="@+id/input"
-            style="@style/textinput_embedded"
-            android:inputType="text"
-            android:hint="" tools:ignore="LabelFor" /> <!-- hint needs to be set by caller -->
-    </com.google.android.material.textfield.TextInputLayout>
-
-</LinearLayout>
+    android:id="@+id/input_frame"
+    style="@style/textinput_edittext_singleline">
+    <EditText
+        android:id="@+id/input"
+        style="@style/textinput_embedded"
+        android:inputType="text"
+        android:hint="" tools:ignore="LabelFor" /> <!-- hint needs to be set by caller -->
+</com.google.android.material.textfield.TextInputLayout>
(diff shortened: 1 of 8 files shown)


eddiemuc:
Code is fine. Wrt label (ft in your example) I would propose to use longer unit names now that it is used as a label (e.g. "feet" "kilometer" instead of "ft" / "km").

There are, however, two new lint issues.

eddiemuc (approved):
Code is fine. Wrt label (ft in your example) I would propose to use longer unit names now that it is used as a label (e.g. "feet" "kilometer" instead of "ft" / "km").

There are, however, two new lint issues.

main/res/layout/dialog_edittext.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
eddiemuc, line 2: One more thing: I don't think you have to wrap the `TextInputLayout` in another layout here, you might have less problems with programmatic view creation if you just put the `TextInputLayout` here as top-level view (like e.g. done with the `MaterialButton` in `button_view.xml` or `button_icon_view.xml`). Also, such a view snippet is more flexible usabe by other parts of the code later.

fm-sys:
>> image




> Wrt label (ft in your example) I would propose to use longer unit names now that it is used as a label (e.g. "feet" "kilometer" instead of "ft" / "km").
> 


Actually, showing the suffix as hint text is kind of ugly in total. There's an official way to set a text prefix and suffix... See:
https://github.com/material-components/material-components-android/issues/1459#issuecomment-813377142
https://github.com/material-components/material-components-android/issues/1427#issuecomment-647103856

main/res/layout/dialog_edittext.xml
moving-bits: Ah, good idea, thanks! I'm so used to having either LinearLayout, FrameLayout or something similar as enclosing frame that I did not try TextInputLayout yet. Will do so, would make things easier.

moving-bits:
> Code is fine. Wrt label (ft in your example) I would propose to use longer unit names now that it is used as a label (e.g. "feet" "kilometer" instead of "ft" / "km").
> 
> There are, however, two new lint issues.

Thanks for taking the time to review. I'll look into the lint issues.

moving-bits:
> Actually, showing the suffix as hint text is kind of ugly in total. There's an official way to set a text prefix and suffix... See:
> [material-components/material-components-android#1459 (comment)](https://github.com/material-components/material-components-android/issues/1459#issuecomment-813377142)
> [material-components/material-components-android#1427 (comment)](https://github.com/material-components/material-components-android/issues/1427#issuecomment-647103856)

Ah, that's good to know, thanks for the links. I had searched for something like this, but couldn't find it on the material.io page for text fields. I'll look into the links provided and adapt the input method accordingly.

moving-bits:
Updated the PR to use `suffixText` for the units, removed the surrounding `FrameLayout`, and added some color definitions (which helps in legacy settings mode).

Works fine generally, except that if you use a suffix, the textbox changes its height?! A bit strange, but happens in legacy settings context only, does work in regular program context (although not yet used there).

image.image

moving-bits:

PR #10942, opened 2021-06-16
fix some lint issues

moving-bits:
## Description
Fixes a couple of "unused resource" lint issues


moving-bits:

PR #10943, opened 2021-06-17
some actionbar and theming improvements

fm-sys:
## Fix bad margins and ripple effects

grafik
grafik

grafik
grafik

Took way too long to find a working fix, but I was really annoyed by it ;-)


moving-bits:
failing tests are unrelated to this PR, so let's merge

PR #10944, opened 2021-06-17
make downloader tests accept a range instead of fixed value of entries

moving-bits:
## Description
Change downloader tests to accept a certain range of files when checking contents of remote servers (instead of expecting a specific fixed value) to not brake our CI if a single map gets added or removed from their service.


moving-bits:
failing test is unrelated and fixed meanwhile (with #10945)

Issue #10939, opened 2021-06-16
Geocaching.com: watchlist counter no longer valid (always returns 2)


eddiemuc:
It is my impression that the watchlist count for geocaching.com is no longer valid. It always seems to return 2. 
I assume this is the result of a website change because the watch count is no longer displayed behind the "Watch" toggle:

![image](https://user-images.githubusercontent.com/6909759/122215405-77d35480-ceab-11eb-992e-3f78fd12bb4a.png)

## Bug description 
When I open a geocaching.com cache stored on my device  with a watchlist count > 0 and I refresh this cache, the watchlist count returned is always 2.

Also, JUnit test `cgeo.geocaching.connector.gc.GCParserTest.testOnlineWatchCount` fails with `Expecting: <2> to be greater than <50>`

### Steps to reproduce the problem
* open c:geo
* open any previously stored geocaching.com cache where you know that the watchlist count is well bigger than 2 (Note: it seems that this is currently not verifyable on geocaching.com) and where a watchlist count well bigger than 2 is stored
* refresh the cache

### Actual result after these steps 
The watchlist count is 2

Note: if you start with a cache whose watchlist count is 0 or 1, the watchlist count will also be 2 after refresh

### Expected result after these steps 
The watchlist count reflects the real watchlist count

## c:geo version 
Current Release: 2021-05-20
current master: 2021-06-15

## Reproducible 
Yes, always

## Additional context 
Checked actual source of a geocaching.com html page. It seems the `2` is parsed by watchlist count regular expression because this regex scans for the term "watchlist.aspx" and returns the next expression within paremthesis behind it. It turns out, right now this is a javascript expression `(2)` some lines after `watchlist.aspx` .

## Additional TODO
Re-enable the GC watchcount test after fixing this issue. The test got disabled with PR #10945 to not distract with all builds failing due to the website being broken.

eddiemuc:
I have no fix for this currently because I can't find a way to parse the watchlist count from another place on the web page right now. Maybe someone else has an idea?
I hope we don't have to remove this feature...

eddiemuc:
See https://forums.geocaching.com/GC/index.php?/topic/362910-number-of-watchers-disappeared/&tab=comments#comment-5895198 
Maybe its a temporary bug

bekuno:
https://forums.geocaching.com/GC/index.php?/topic/362910-number-of-watchers-disappeared/&do=findComment&comment=5895429
[...]

PR #10945, opened 2021-06-17
temporarily disable GC watchcount test 

moving-bits:
## Description
temporarily disable GC watchcount test which fails due to website changes (see #10939)

(I'll add a reminder over there to re-enable this test after fixing the issue)


moving-bits:
failing test is unrelated and will be fixed as part of #10944

PR #10955, opened 2021-06-19
migrate enter log to TextInputLayout (including counter)

moving-bits:
## Description
Migrates the "enter log" dialog to `TextInputLayout` and replace our self-built character counter with material built-in

image.image

### DIFF
--- a/main/res/layout/logcache_activity.xml
+++ b/main/res/layout/logcache_activity.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView
+    xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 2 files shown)


fm-sys:
For multi-line text input fields we IMO shouldn't use the "clear all" button on the right. Imagine a user who accidentally hits this button and the whole 500 words log gets deleted... :-/

The button should only be there, where the loss wouldn't be too great ;-)

moving-bits:
I'm currently working on parts of the coordinates calculate dialog, which has a multi-line input field as well. I guess we should make this consistent across the input fields, so I would remove the "clear_text" from `textinput_edittext` style and apply it to `textinput_edittext_singleline` only.

Lineflyer:
> For multi-line text input fields we IMO shouldn't use the "clear all" button on the right. Imagine a user who accidentally hits this button and the whole 500 words log gets deleted... :-/

Agree!

Issue #10960, opened 2021-06-19
Image select activity: Refactor buttons


moving-bits:
## What function are you missing
In the "add image" activity (which is part of the logging activity) you get three buttons on editing an uploaded image: cancel / ok / delete (in this order).

![image](https://user-images.githubusercontent.com/3754370/122640123-fde6d980-d0fd-11eb-92b8-c0c19fd5341e.png)

IMHO this should be refactored, as
- text gets truncated (although they are relatively short)
- "ok" button should be the right-most button

## Suggested solution
I have no perfect solution at hand, but I guess confusion stems from the fact that we are trying to do two different things in one dialog:
- edit a description (which needs "cancel" and "ok")
- offer to delete an attached image

Maybe the solution would be to remove the "delete" button. You can still delete the image from the list of images.

eddiemuc:
The delete button was meant for users who want to take a more detailed look at an image before deleting it (in the list view you only see it very small).

But I am totally fine with removing it. It's not that important to have it here.

But on another note: maybe we should think about making the button text size smaller. I really think that three buttons with such short texts should fit next to each other on a normal screen size without truncating.

moving-bits:
Maybe we can put Cancel / Ok in one line, as regular buttons (maybe even without outline), and put the "delete" button into a separate line, full-width, with outline?

Regarding button font size: I'll play with it a bit, and will come up with a proposal.

moving-bits:
Alternatively, I would change the order to "cancel / delete / ok" or "delete / cancel / ok" (not sure which one's better).

murggel:
How about putting the "delete" to the 2 buttons on top of the activity?
![Screenshot_20210620-150027_cgeo.jpg](https://user-images.githubusercontent.com/70113341/122675150-6609ee00-d1d8-11eb-948d-33d59c3d3c77.jpg)

I guess there will be also problem with available space, but at least it does not disturb ok and cancel.

On leaving it at the bottom, I would vote for "delete", "cancel" "ok", because cancel and ok should always be at the same position

moving-bits:
Ok, let's start with re-ordering the buttons, "delete" first, then "cancel", then "ok".
If no-one has complained yet about truncated button labels, then we can leave it at three buttons in the same line, especially since we have reduced button label font size meanwhile.

PR #10966, opened 2021-06-19
reduce button font size

moving-bits:
## Description
As discussed in https://github.com/cgeo/cgeo/issues/10960#issuecomment-864392162: here's a proposal for a reduced button text font size.

|old font size|new font size|
|---|---|
|image|image|
|image|image|


Lineflyer:
The smaller font looks much smoother and no longer so intrusive.

eddiemuc:
I like it!

moving-bits:
ok, thanks for your feedback. Then let's merge.

PR #10968, opened 2021-06-20
migrate log template dialog (rel. to #10763)

moving-bits:
## Description
And now down into the more ugly parts of our preferences...
- Migrates the "log template" dialog to `TextInputLayout`. Needs some manual adjustments of colors.
- Adds some basic dialog styling for our old preferences.


moving-bits:
BTW: With this PR we should have covered every `EditText` and every `AutoCompleteTextView` (except the one in the coordinates input dialog we decided not to adapt due to space constraints), so that migration to `TextInputLayout` is completed and we can close #10763 after merging this PR. I'll add a dependency.

PR #10969, opened 2021-06-20
add colored border to coordinates input (for consistency)

moving-bits:
## Description
- related to #10763
- add a colored border around the coordinate input fields in coordinate calculator for consistency (both coordinates input and calculator input)
- there's no added functionality like `TextInputLayout` has, so no hint, no start/end icon etc.
- it has the same outer margins as `TextInputLayout`, but inner padding is made smaller deliberately, as so many input boxes are included which need space

image.image.image

### DIFF
--- a/main/res/drawable/textinputlayout_bcg_active.xml
+++ b/main/res/drawable/textinputlayout_bcg_active.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:left="6dp" android:right="6dp" android:bottom="6dp" android:top="6dp">
+        <shape>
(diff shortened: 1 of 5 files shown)


fm-sys:
Looking at the third picture, the "filled text view" looks much better than the current style for empty text inputs. Is there any way to apply the "filled" style for empty text inputs as well?

moving-bits:
retest this please

moving-bits:
> Looking at the third picture, the "filled text view" looks much better than the current style for empty text inputs. Is there any way to apply the "filled" style for empty text inputs as well?

The "filled style" is used in this special case only to show that those "input fields" are meant to be buttons. So it's a different use case here. Only on long-pressing such a "button" you get the actual input field, which is then formatted like the other input fields. (Short pressing toggles other actions on a per-field basis.)

PR #10970, opened 2021-06-20
align full button margin to TextInputLayout margin

moving-bits:
## Description
As requested in one of the issues related to `TextInputLayout`:
Our full-width buttons should use the same margins as our input fields, which is what this PR adjusts styles to.

image


moving-bits:
failing tests are unrelated, gc.com seems to have server problems - logging in via website fails as well currently

moving-bits:
retest this please

PR #10983, opened 2021-06-22
Fix #10939: Adapt watchlist pattern to website changes

### LINKED ISSUE #10939: Geocaching.com: watchlist counter no longer valid (always returns 2)

UniQP:


Lineflyer:
As the next `master` release cycle could take a bit longer, we should IMHO fix this on `release` in case we decide to publish a bugfix release before going into beta phase for current `master`.

UniQP:
@Lineflyer Retargeted to `release`.

Lineflyer:
@moving-bits Seeing that the map downloader test still fails on `release`. Did you change the test only on `master`? That would explain.

PR #10987, opened 2021-06-22
cleanup components section

fm-sys:
We had quite some leftovers in our components section. AFAIK we don't use them anymore, but can someone cross-check?

Especially, I'm not sure about these two ones:
     - [RRZE Icon set](https://github.com/RRZE-PP/rrze-icon-set) ([CC-BY-SA 3.0](https://creativecommons.org/licenses/by-sa/3.0/))
     - The [Tango! Desktop Project](http://tango.freedesktop.org/Tango_Desktop_Project) (Public Domain)

Furthermore, I removed the "from the Noun Project" which was behind every listed icon. This information was superfluous, since all elements are listed in "the Noun Project" subcategory anyway.

|before|now|
|-|-|
| grafik |grafik

### DIFF
--- a/main/res/values/strings_not_translatable.xml
+++ b/main/res/values/strings_not_translatable.xml
@@ -141,15 +141,11 @@
     - [Mapsforge](https://code.google.com/p/mapsforge/) (OSM-rendering)\n
     - [The Noun Project](https://thenounproject.com/) (basis for attribute icons):\n%1
     - [The Apache Commons Project](https://commons.apache.org/)\n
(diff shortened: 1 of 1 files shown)


moving-bits:
Have looked into this issue a week ago or so, but have not been able to identify all icons. For example the "blue info" icon may be coming from the "RRZE Icon set", but I cannot say for sure. I'm undecided as to this line, but appreciate the more concise look for the "The Noun Project" attributions.

fm-sys:
The "blue info" is way newer than the "RRZE Icon set" entry IIRC. But I can re-add the RRZE component if we want to be on the save side...

moving-bits:
let's merge here to get the optimizations into the release. If some icon source should indeed be missing, we can re-add it again easily.

PR #10994, opened 2021-06-22
fix size for emojis in emoji-select-dialog (fix #10948)

### LINKED ISSUE #10948: Emojis in emoji selector are way too small
As the title describes, the emojis in the emoji selector are displayed way too small...

grafik

murggel:
## Description

shows emojis in better size.
Use method to get drawable for given dp-size (as suggested from @eddiemuc in https://github.com/cgeo/cgeo/issues/10948#issuecomment-864382447), use only one calculation from desired dp-size to helper-class `EmojiPaint`.
Set suitable fontSize for the textView

image

Right margin still does not look very good... probably I tried margins and padding, but probably at the wrong widget / place?
 
## Related issues

fix #10948

## Additional context

The margin are not well defined and too many columns for tablet-landscape-mode. 
I am looking for that, but did not find a solution yet: Number of columns is calculated in respect of the screen width, but the dialog does not use the maximum possible width. So how did I get the correct dialog-width?

image

### DIFF
--- a/main/src/cgeo/geocaching/utils/EmojiUtils.java
+++ b/main/src/cgeo/geocaching/utils/EmojiUtils.java
@@ -7,6 +7,7 @@ import cgeo.geocaching.databinding.EmojiselectorBinding;
 import cgeo.geocaching.maps.CacheMarker;
 import cgeo.geocaching.storage.extension.EmojiLRU;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
+import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.functions.Action1;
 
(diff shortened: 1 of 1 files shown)


fm-sys:
According to SO, maybe `Dialog.getWindow().getDecorView().getHeight()` would help? (but probably this is only available as soon as the dialog gets displayed) 

Or can we just add margin to the emojis?

eddiemuc:
Hi @murggel , the PR looks fine so far and can be merged imho. I would have done it immediately except for your question wrt "number of column" calculation, because I have a solution suggestion and I don't know if you want to implement such a thing as part of this PR or not?

My strong suggestion would be to drop the approach of using a `GridLayoutManager` with calculated maxCol and use a `com.google.android.material.chip.ChipGroup` instead. This is a material.io layout class extending from `FlowLayout` (which unfortunately can't be used directly). It places children views from left to right until next view does not fit in line any more, then starts a new line.

I used this layout to implement the Attribute Filter which has almost same requirements as for creating a "grid" of clickable icons. The class to get a working example from would be `cgeo.geocaching.filters.gui.AttributesFilterHolder`, method `createView()`. As you can see in following screenshots (one in portrait, one in landscape) it works very good:

image


image

murggel:
thanks for the ideas, I would try the approach of @eddiemuc in a separate PR.
I am hiking this weekend, so I can start working on that not before next week.

eddiemuc:
Ok then I will.merge this PR

eddiemuc:
And thanks to @murggel for the contribution of course!

Issue #9088, opened 2020-09-28
Consider switching from PNG to SVG


okainov:
Having one scalable SVG icons should probably work much better than having dozens of PNG for all possible resolutions, also this would reduce size of the final APK. 

However this might require large efforts i.e. for the cases where icon source is gone / not available any more and it will make it harder to find the vector option.

**Additional context**
See also #6972

fm-sys:
There is already a PR for it, see #7856. But AFAIK some lower API versions do not have svg support. In addition I don't know how it will work with our light theme. And our apk size is quite small compared to other apps. 

To make a long story short: I don't see a real advantage in switching to SVG graphics...

okainov:
>There is already a PR for it, see #7856. 

Which feels forgotten/abandoned since half year already... But it's a good base probably.

> But AFAIK some lower API versions do not have svg support. 

From the linked PR:  "To work with API 16-20, the vector drawable compat library is used.". So it is not an issue.


In general, this issue was just to discuss the possibilities. Especially seeing some reported issues as well as reported icons inconsistency AND having a base PR for it already - feels like just a little piece of work left and it will make some improvements already.

moving-bits:
Generally I agree. Only that I do not see anybody around here having the knowledge necessary for creating/managing the SVG currently.
With #7856 seemingly abandoned by the original author, and no one in the team with knowledge + time for SVG, I would not want to stumble into the switch to SVG currently - although there would be a couple of advantages coming with the switch.

fm-sys:
> >But AFAIK some lower API versions do not have svg support.
>
>From the linked PR: "To work with API 16-20, the vector drawable compat library is used.". So it is not an issue. 

I'm not sure... Sounds different in https://github.com/cgeo/cgeo/pull/7856#issuecomment-619408476:
>>Also, if the PNG were needed only to support API level < 21, you may remove them as well.
>
> Let's leave the legacy support in, because it fixes vector issues on Lollipop and Marshmallow. The PR should be good-to-go for 21+

eddiemuc:
Update: I put a link to this issue into the Wiki page explaining how to create icons for c:geo (currently described procedures are for creating PNGs, but they consider how to create the "special" c:geo icons working in both dark and light mode): https://github.com/cgeo/cgeo/wiki/How-to-create-custom-icons

bekuno:
copy of @eddiemuc text in another issue. Here is the right place.

[...]

PR #10996, opened 2021-06-23
add origin as comment into svg (xml)

murggel:
## Description

for better documentation, where the svg was taken from, add a comment into the xml itself with the source-name and origin of the svg.
The table in https://github.com/cgeo/cgeo/pull/10818#issuecomment-855401938 is then deprecated and there is no need to track somewhere in the Wiki.

## Related issues

rel issue #9088
rel PR #10818

## Additional context

Probably for the modified there is more info necessary?

### DIFF
--- a/main/res/drawable/ic_action_layers.xml
+++ b/main/res/drawable/ic_action_layers.xml
@@ -1,3 +1,4 @@
+<!-- taken from layers-outline / materialdesignicons-google -->
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
     android:width="24dp"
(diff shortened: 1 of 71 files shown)


fm-sys:
> Probably for the modified there is more info necessary?

I don't think that we need to document it that detailed. It's IMHO totally enough to know where the icon was taken from...

murggel:
I added a section to the WIKI for adding svg (https://github.com/cgeo/cgeo/wiki/How-to-create-custom-icons#use-of-svg)

PR #11002, opened 2021-06-24
ALC API: send X-consumer-key (fixes issue_11001)

baiti:
The referenced issue addresses a lockdown that has been applied to the ALC public API. It is now required to send an "X-consumer-key" with each request. This PR prepares for sending that key, however, since we do not have a key at this time, merging this PR will not make a difference with respect to the provided functionality. Still, I think the change should be merged, otherwise I'd have to redo it later or stash it away locally (and manage it myself).

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -40,6 +40,7 @@ final class LCApi {
     private static final SynchronizedDateFormat DATE_FORMAT = new SynchronizedDateFormat("yyyy-MM-dd", Locale.getDefault());
     @NonNull
     private static final String API_HOST = "https://labs-api.geocaching.com/Api/Adventures/";
+    private static final String CONSUMER_KEY = "11111111-2222-3333-4444-5555555555555";
 
     private LCApi() {
         // utility class with static methods
@@ -50,8 +51,9 @@ final class LCApi {
         if (!Settings.isGCPremiumMember()) {
             return null;
         }
+        final Parameters headers = new Parameters("X-Consumer-Key", CONSUMER_KEY);
         try {
-            final Response response = apiRequest(geocode.substring(2)).blockingGet();
+            final Response response = apiRequest(geocode.substring(2), null, headers).blockingGet();
             return importCacheFromJSON(response);
         } catch (final Exception ignored) {
             return null;
@@ -76,12 +78,13 @@ final class LCApi {
         final double radius = gp1.distanceTo(gp2) * 500; // we get diameter in km, need radius in m
         Log.d("_LC Radius: " + String.valueOf((int) radius));
         final Parameters params = new Parameters("skip", "0");
+        final Parameters headers = new Parameters("X-Consumer-key", CONSUMER_KEY);
         params.add("take", "500");
         params.add("radiusMeters", String.valueOf((int) radius));
         params.add("origin.latitude", String.valueOf(latcenter));
         params.add("origin.longitude", String.valueOf(loncenter));
         try {
-            final Response response = apiRequest("SearchV3", params).blockingGet();
+            final Response response = apiRequest("SearchV3", params, headers).blockingGet();
             return importCachesFromJSON(response);
         } catch (final Exception ignored) {
             return Collections.emptyList();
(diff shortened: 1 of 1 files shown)


Lineflyer:
sounds meaningful, but should maybe target `release` for the case, that we receive such a key before we have `master` ready for rolling out. Chances are low...but still...

eddiemuc:
There's a small checkstyle finding in it (unused import statement)

fm-sys:
> sounds meaningful, but should maybe target `release` for the case, that we receive such a key...

If (and that's a big "if") we receive such a key, we would probably need to keep it private, similar to our GMaps API keys.

baiti:
> There's a small checkstyle finding in it (unused import statement)

Intreresting. The plugin in the IDE did not find it.

baiti:
> > sounds meaningful, but should maybe target `release` for the case, that we receive such a key...
> 
> If (and that's a big "if") we receive such a key, we would probably need to keep it private, similar to our GMaps API keys.

Fully agree to both parts of your statement. I figured moving out the string constant could be done later. I didn't want to loose the handling of request headers by the `Network` object. Pls review how I implemented that as I haven't seen it anywhere in the code being done so chances are it is not really clean coding what I did.

baiti:
> There's a small checkstyle finding in it (unused import statement)

Do you mean this one:

Screenshot from 2021-06-24 15-07-52

If yes, what is our policy, do we fix these things even if they are totally unrelated to the PR in question?

fm-sys:
> I figured moving out the string constant could be done later. 

Yes, for sure. We don't need to keep something private which does not exist at all ;-)

eddiemuc:
> > There's a small checkstyle finding in it (unused import statement)
> 
> Do you mean this one:
> 
> Screenshot from 2021-06-24 15-07-52
> 
> If yes, what is our policy, do we fix these things even if they are totally unrelated to the PR in question?

Hm I am puzzled. This should not be possible because whatever change causes this initially should have been blocked from merge as well.

Looking deeper into it I assume it was caused by a change I did in release which then maybe was merged into master without a checkstyle run.

Usual policy in this case is to fix these findings, even if you are not the source of it. At least if fix is as trivial as this one.

baiti:
> > > There's a small checkstyle finding in it (unused import statement)
> > 
> > 
> > Do you mean this one:
> > Screenshot from 2021-06-24 15-07-52
> > If yes, what is our policy, do we fix these things even if they are totally unrelated to the PR in question?
> 
> Hm I am puzzled. This should not be possible because whatever change causes this initially should have been blocked from merge as well.
> 
> Looking deeper into it I assume it was caused by a change I did in release which then maybe was merged into master without a checkstyle run.
> 
> Usual policy in this case is to fix these findings, even if you are not the source of it. At least if fix is as trivial as this one.

I was about to update GCParser.java and noticed that I must have a different version of that file in my local repo even though I refreshed it right before I made my changes to LCApi.java. Also, the import in question is actually required. If I remove it, it doesn't compile anymore. So as a consequence. I better don't touch it.

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -76,12 +78,13 @@ protected static Geocache searchByGeocode(final String geocode) {
         final double radius = gp1.distanceTo(gp2) * 500; // we get diameter in km, need radius in m
         Log.d("_LC Radius: " + String.valueOf((int) radius));
         final Parameters params = new Parameters("skip", "0");
+        final Parameters headers = new Parameters("X-Consumer-key", CONSUMER_KEY);
```
eddiemuc, line 81: Maybe put "X-Consumer-key" into a string constant

eddiemuc:
> > > sounds meaningful, but should maybe target `release` for the case, that we receive such a key...
> > 
> > 
> > If (and that's a big "if") we receive such a key, we would probably need to keep it private, similar to our GMaps API keys.
> 
> Fully agree to both parts of your statement. I figured moving out the string constant could be done later. I didn't want to loose the handling of request headers by the `Network` object. Pls review how I implemented that as I haven't seen it anywhere in the code being done so chances are it is not really clean coding what I did.

Overall change looks fine to me (I made one small comment)

eddiemuc:
> I was about to update GCParser.java and noticed that I must have a different version of that file in my local repo even though I refreshed it right before I made my changes to LCApi.java. Also, the import in question is actually required. If I remove it, it doesn't compile anymore. So as a consequence. I better don't touch it.

I suspect you need to do a rebase against `release` branch

baiti:
Will do a new one based on release branch

Lineflyer:
> Hm I am puzzled. This should not be possible because whatever change causes this initially should have been blocked from merge as well.
> Looking deeper into it I assume it was caused by a change I did in release which then maybe was merged into master without a checkstyle run.

Probably my bad, I was forced to perform a conflict resolution while merging `release` to master` and as `master` contained both imports I left them in while they might no longer be needed due to the watchlist change. I didn't know better.

PR #11003, opened 2021-06-24
ALC API: send X-consumer-key (based on release) (fixes issue_11001)

baiti:
The referenced issue addresses a lockdown that has been applied to the ALC public API. It is now required to send an "X-consumer-key" with each request. This PR prepares for sending that key, however, since we do not have a key at this time, merging this PR will not make a difference with respect to the provided functionality. Still, I think the change should be merged, otherwise I'd have to redo it later or stash it away locally (and manage it myself).


baiti:
@Lineflyer,  @fm-sys,
If we wanted to get that functionality back very quickly (within minutes), I could "provide" the necessary details, however we would need to put those into a file that doesn't go directly into the repo, just let me know if, and if yes, how we do it in detail.

eddiemuc:
failing test is unrelated, so let's merge here

PR #11005, opened 2021-06-24
Improve GPX log import

### LINKED ISSUE #9639: Store service-specific log id on PQ / GPX import
_Originally posted by @eddiemuc in https://github.com/cgeo/cgeo/issues/9524#issuecomment-752952048_

> In the database we have a field `service_log_id`, but that seems to be filled rarely. (Maybe only on direct download of caches via c:geo, not when using PQ/GPX files

Indeed service_log_id is currently filled only on direct download, but it should be filled for all downloads done after last release (where it was introduced). This is is service-specific and implemented for gc.com, oc and some other service. I added this field to enable direct links to log entries on website (can't remember issue id).

I guess it could be filled for GPX imports as well if the contain the service-specifc log ids

### LINKED ISSUE #10028: Duplicated GSAK logs on GPX import
A user contacted us, that he has a specific problems since some releases (probably starting with 2021.01.10):

- The user handles his caches with GSAK
- Then he exports the desired caches from GSAK into a GPX which he imports to c:geo. This usually also contains caches already stored (from the last import) to refresh them.
- The GSAK GPX contains a special log (as first log) which contains user specific information (selectable by the user)

In former versions this worked perfectly fine (first log is the GSAK special log).
Recently he has the problem, that this special log entry gets duplicated on each new import of a GPX with this special log.

The log part of the GPX looks like this (whole file available upon request):
```
 <groundspeak:logs>
    <groundspeak:log id="-3">
    <groundspeak:date>2021-02-15T08:00:00Z</groundspeak:date>
    <groundspeak:type>Write note</groundspeak:type>
    <groundspeak:finder id="0">GSAK</groundspeak:finder>
    <groundspeak:text encoded="False">Guilford County,  N
    </groundspeak:text>
   </groundspeak:log>
```

IMHO this is related to the recently introduced merge capability on GPX import but should/could be avoided.

fm-sys:
fix #9639 
fix #10028

- Store service-specific `log_id` on PQ / GPX import 
- avoid log entry duplicates while importing PQ / GPX

The second is done by don't search for a specific timestamp in the db anymore, but rather accept the whole time range of the log day while searching for duplicates.

It's currently targeted to master, as users seam not to be affected too much by this bug (~at least @Lineflyer hadn't reported any support requests~ there where really quite some requests at the start of this year which seem to  got lost). If we want to do another small bugfix release anyway, I can retarget the PR for sure...

### DIFF
--- a/main/src/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/cgeo/geocaching/files/GPXParser.java
@@ -5,6 +5,8 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.capability.ILogin;
+import cgeo.geocaching.connector.gc.GCConnector;
+import cgeo.geocaching.connector.gc.GCUtils;
 import cgeo.geocaching.connector.internal.InternalConnector;
 import cgeo.geocaching.connector.tc.TerraCachingLogType;
 import cgeo.geocaching.connector.tc.TerraCachingType;
@@ -619,8 +621,13 @@ abstract class GPXParser extends FileParser {
             try {
                 if (attrs.getIndex("id") > -1) {
                     logBuilder.setId(Integer.parseInt(attrs.getValue("id")));
+
+                    final IConnector connector = ConnectorFactory.getConnector(cache);
+                    if (connector instanceof GCConnector) {
+                        logBuilder.setServiceLogId(GCUtils.logIdToLogCode(logBuilder.getId()));
+                    }
                 }
-            } catch (final NumberFormatException ignored) {
+            } catch (final Exception ignored) {
                 // nothing
             }
         });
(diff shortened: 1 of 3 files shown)


fm-sys:
Oh, and will fix #10028 as well...

eddiemuc:
@fm-sys seems like you're the unlucky guy who got the warning about an unused import statement in GCParser innyour PR. Would it be ok for you to just remove it there (although your PR is unrelated to it of course). I feel an own PR for this is kind of too much for such a small thing, and I have no PR pending currently.

eddiemuc:
@fm-sys forget my comment, @moving-bits already fixed it in separate PR.

Just for your interest: this was introduced by a merge from release to master with merge conflict, unfortunately those seem not to run through checkstyle. Initial PR in release was done by me, merge was done by @Lineflyer 

Don't know if we need an enhanced process for this, I think it is ok to fix such small things as part of other Pars instead.

moving-bits:
rebasing this PR to current master will fix the CheckStyle issue

PR #11009, opened 2021-06-26
format waypoint type spinner (fix #10974)

### LINKED ISSUE #10974: WP type dropdown selector looks strange
No big issue, but to me the waypoint type selector when creating/editing a waypoint looks kind of narrow.
I mean the selector where you can choose the type of waypoint, e.g. "Reference Point", "Final", "Trailhead", etc.

Maybe we could give it the same room as the textfields above and below?

Screenshot_20210620-222432_cgeo

moving-bits:
## Description
- applies more room around waypoint type spinner
- apply dialog background to spinner popup to be distinguishable from background in dark mode

image.image.image

### DIFF
--- a/main/res/layout/editwaypoint_activity.xml
+++ b/main/res/layout/editwaypoint_activity.xml
@@ -77,7 +77,6 @@
             android:visibility="gone"
             android:layout_marginTop="16dp"
             android:layout_marginBottom="16dp"
-            android:popupBackground="@color/colorBackgroundDialog"
             tools:listitem="@android:layout/simple_spinner_item" />
 
         <com.google.android.material.textfield.TextInputLayout style="@style/textinput_dropdown">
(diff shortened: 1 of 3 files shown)


fm-sys:
Shouldn't we apply the spinners `popupBackground` to all places in c:geo (e.g. list selector, text input suggestions, ...) via our theme instead?

moving-bits:
Found a way to set the background via theme. Margins still need to be set manually, though.

eddiemuc:
Probably an own issue, but I would like it if we had a general style for spinners across c:geo, of course a nice one done with material.io. right now we have very different spinner/spinner-like controls across whole c:geo e.g. waypoint-type-select/image-scale-select vs. Log-type-select (in offline log)

moving-bits:
I have not found a material.io component for it, so we probably need to create our own component to achieve that.

Issue #9992, opened 2021-02-09
Tapping compass rose on waypoint tab triggers scrolling instead


andrcoding:
Since an update last summer/autumn, it has become difficult to access the map view directly by long pressing on the compass (icon on the right) of a waypoint (Opened cache-most left page - waypoints). Instead of pressing the compass, the scroll bar is often triggered. Before everything works fine. Moving the compass a little bit left or changing the touching areas would be helpful.

![cgeo](https://user-images.githubusercontent.com/78544594/107421389-2fc9ad00-6b1a-11eb-8173-7e72327fa106.png)

Edit: One of the system I have tested it: [system information.txt](https://github.com/cgeo/cgeo/files/5954264/system.information.txt)

fm-sys:
Can you please post which c:geo and Android version you use, or ever better post the system information, which you will find in c:geo Menu -> About c:geo -> Swipe right to System?

At least for me, the problem does not exist, so it's either meanwhile fixed or a device specific problem...

Narrat:
Just a FYI post, as I noticed that the requested information got edited into the first post :)

fm-sys:
Thanks for the hint. 

Seems like the cgeo version is up-to-date. So there might be a bug in our "fastscroll magic" which should normally gets deactivated after one second with no scrolling. 

@andrcoding does this happen to you always or only after performing a fast fling?

andrcoding:
From my experience, it happens not only after performing a fast fling. However, it seems that the "fastscroll magic" feature is not deactivated all the time after one second. Typically it is about two seconds. I will watch this behavior the next times of usage to make a clearer statement if "fastscroll magic" is deactivated while the problem occur. I will give an update.
In general, I have first noticed it after a major update but I am not sure which one it was. It could be version 2020.08.04.

@Narrat Thanks for informing. I thought there will be a notice.

andrcoding:
After some testing some feedback: It seems that the "fastscroll magic" feature is not always getting activated which is ok. In general, a reduction of the deactivation time will be good as it sometimes takes 2-3 seconds. In my opinion, 0.5 seconds should be totally enough. Thus, my problem is only related to time.

Lineflyer:
If this is configurable in the code we could indeed reduce it, but I suggest something around 1 second to allow users to view the scroll position before it disappears.

Lineflyer:
I stumbled over this issue while caching out in the field lately.
[...]

PR #11016, opened 2021-06-26
add extra padding for easier fastscroll in wpt list (rel. to #9992)

moving-bits:
## Description
This is a proposal for how to handle the conflict between fastscroll handler and content accessibility - for waypoint list in this case. I've added a small extra padding of `5dp`, see screens below.

If this works out fine, we can apply the same for the other fastscroll-enabled views.

|without extra padding|with added fastscroll padding|
|---|---|
|image|image|


eddiemuc:
looks better to me than before and is ok code-wise. Thus I will merge.

If padding is still not enough, let's open a new PR/issue for it

PR #11018, opened 2021-06-27
switch to SDK 30

eddiemuc:
This PR will switch c:geo to SDK 30 with all consequences (esp. no more usage of legacy storage access)

### DIFF
--- a/main/src/cgeo/geocaching/storage/ContentStorage.java
+++ b/main/src/cgeo/geocaching/storage/ContentStorage.java
@@ -84,6 +84,7 @@ public class ContentStorage {
             this.lastModified = lastModified;
         }
 
+        @NonNull
         @Override
         public String toString() {
             return
@@ -553,7 +554,7 @@ public class ContentStorage {
 
         for (Folder candidate : candidates) {
             //candidate is ok if it is either directly accessible or based on another public folder (which will become accessible later)
-            if (candidate != null && (candidate.getRootPersistableFolder() != null || ensureFolder(candidate, needsWrite))) {
+            if (candidate != null && ensureFolder(candidate, needsWrite, true)) {
                 return candidate;
             }
         }
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,9 @@ plugins {
 
     // use SpotBugs instead of FindBugs, see https://plugins.gradle.org/plugin/com.github.spotbugs
     id 'com.github.spotbugs' version '4.7.1'
+
+    // enables a gradle task 'tiTree' to print a task dependency graph, see https://gitlab.com/barfuin/gradle-taskinfo
+    id 'org.barfuin.gradle.taskinfo' version '1.0.5'
 }
 
 /*
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -5,7 +5,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 29
+    compileSdkVersion 30
 
     // signing is handled via private.properties
     signingConfigs {
@@ -20,7 +20,7 @@ android {
     defaultConfig {
         minSdkVersion 21
         //noinspection OldTargetApi
-        targetSdkVersion 29
+        targetSdkVersion 30
 
         // include only those language resources from libraries which we actively maintain ourselves in the translation project
         // not yet enough translations (~50%): "is","iw" (="he"),"zh"
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -53,10 +53,8 @@
         android:usesCleartextTraffic="true"
         android:supportsRtl="true"
         android:largeHeap="true"
-        android:requestLegacyExternalStorage="true"
+        android:preserveLegacyExternalStorage="true"
         >
-        <!-- TODO SAF: leave 'android:requestLegacyExternalStorage="true"' until done, remove completely when done -->
-        <!-- TODO SAF: put 'android:preserveLegacyExternalStorage="true" when using API30 -->
 
         <!-- Samsung Multi-Window support -->
         <uses-library
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -19,7 +19,7 @@ if (isContinuousIntegrationServer()) {
 }
 
 android {
-    compileSdkVersion 29 // TODO SAF: leave at 29 until done, change to 30 when done
+    compileSdkVersion 30
 
     compileOptions {
         // use the diamond operator and some other goodies in Android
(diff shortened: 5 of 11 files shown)


eddiemuc:
I will not merge this without approval some sort of second opinion/approval

eddiemuc:
CI build fails with following:
image

I am a bit lost right now. `android:preserveLegacyExternalStorage="true"` is a valid SDK 30 argument (but not present in SDK29), why is CI not recognizing this? I changed target SDK from 29 to 30 in `build.gradle`?

Also, build seems to run fine locally.

eddiemuc:
retest this please

fm-sys:
BTW, if using SAF, grading general storage access  permission is not necessary anymore in the wizard for API >= 29 (Android 10) 

In fact, that's  the whole idea of SAF that users don't have to grant this permission anymore...

UniQP:
Is SDK 30 installed on the agent?

eddiemuc:
Looks like I need to upgrade `cgeo-contacts` to SDK30 too in order to make our CI build work.

Is that ok, are any problems expected by this (I am not using this plugin myself)?

eddiemuc:
Also removing permission WRITE_EXTERNAL_STORAGE since it is no longer honored in Android 11 (see e.g. https://developer.android.com/about/versions/11/privacy/storage)

eddiemuc:
Funny, after removing the WRITE_EXTERNAL_STORAGE flag a test started to fail: cgeo.geocaching.brouter.expressions.EncodeDecodeTest.encodeDecodeTest 

It seems like this test hard-expects a file to be present when accessing it via SAF. However, since now the fallbacks to "file-access" will no longer work and no access permission is set up in our CI device, this assumption might no longer be valid.

I put the flag back in, let's see if this makes the test green again.

eddiemuc:
Ok, the removal of WRITE_EXTERNAL_STORAGE was not the culprit so I removed it again.

It seems that the test worked one time after the change (probably because it was able one last time to access via file due to the preserveLegacyExternalStorage flag), but now it is not any more.

I looked into the test code, and it is crystal clear that this test will not work in CI because it tries to access files using SAF - and the fallback option to use file-access will not work any more. Since the test is quite complex, I just deactivated for now.

@moving-bits may I ask you to take a look at the test and maybe rewrite it in a way such that it does not rely on SAF access any more?

eddiemuc:
Added some exception code to get more insights, root cause of test failure is this:

```
13:28:21  > Task :main:connectedBasicDebugAndroidTest
13:28:21  Starting 593 tests on android-28(AVD) - 11
13:29:03  
13:29:03  cgeo.geocaching.brouter.expressions.EncodeDecodeTest > encodeDecodeTest[android-28(AVD) - 11] [31mFAILED [0m
13:29:03  	java.lang.RuntimeException: Couldn't create file 'car-eco.brf' in '/storage/emulated/0/Documents/cgeo/routing[PERSISTABLE_FOLDER(BASE)#1:p-file:///storage/emulated/0/Documents::/cgeo/routing]'
13:29:03  	at cgeo.geocaching.brouter.util.DefaultFilesUtils.checkDefaultFiles(DefaultFilesUtils.java:58)
13:30:29  
13:30:29  > Task :main:connectedBasicDebugAndroidTest FAILED
13:30:29  
13:30:29  FAILURE: Build failed with an exception.
13:30:29  
13:30:29  Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
```

eddiemuc:
seems like usage of "Documents" folder has some surprises in it at Android 11...

eddiemuc:
My best guess: mime type problems. But this has to wait, on my way to the lake...

eddiemuc:
That's a new low!

In Android 11, it seems that even when using file access in some dirs, then FIles contained in dirs are NOT returned as part of dir.list() for these dirs. See e.g. dir.list compared to file.isFile in following log output

```
java.lang.RuntimeException: Could not create file /storage/emulated/0/Documents/cgeo/routing/car-eco.brf in dir /storage/emulated/0/Documents/cgeo/routing; dir.isDirectory()=true, dir.canRead=true, dir.canWrite=true file.isFile()=true, dir.content=segments4 *end*
at cgeo.geocaching.storage.FileContentAccessor.create(FileContentAccessor.java:43)
at cgeo.geocaching.storage.ContentStorage.create(ContentStorage.java:179)
at cgeo.geocaching.storage.ContentStorage.create(ContentStorage.java:161)
```

moving-bits:
> @moving-bits may I ask you to take a look at the test and maybe rewrite it in a way such that it does not rely on SAF access any more?

Haven't seen my dev machine the last days, but to give some context: On starting the routing engine c:geo checks whether the default *.brf files exist in the chosen routing folder, and tries to create them, if not - this should be the place where the error comes from. (And if no SAF rights are granted yet, which is probably the case on the CI, the default folder will be taken.)
Does this error happen on the CI only, or on a fresh emulator as well?

eddiemuc:
> Does this error happen on the CI only, or on a fresh emulator as well?

After try-and-error the whole day with scratch code sent to CI, I think the root cause is a very very strange behavior of using the Documents folder using File interface on an Android 11 device. At least at the emulator. I will remove this as possible default folder and hope this issue will be solved then.

eddiemuc:
Retest this please

fm-sys:
@eddiemuc have you seen my comment https://github.com/cgeo/cgeo/pull/11018#issuecomment-869133639 further up? Do you want to work on it in this PR or should we create an additional issue for it?

eddiemuc:
> @eddiemuc have you seen my comment https://github.com/cgeo/cgeo/pull/11018#issuecomment-869133639 further up? Do you want to work on it in this PR or should we create an additional issue for it? 

I think a separate issue would be better

eddiemuc:
IMHO: ready for review/merge

eddiemuc:
Switching to SDK 30 is a big change, and there's no real coming-back from it once it is out.

I will wait for feedback until tomorrow morning. If nothing comes, I will merge

Lineflyer:
I think we are as ready as we can be...from all what we know.

We can check the nightlies for some more days and then place it into beta as well.

main/src/cgeo/geocaching/storage/ContentStorage.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.jetbrains.annotations.NotNull;
```
moving-bits, line 31: Is there a reason why you chose "NotNull" instead of the androidx "NonNull"?

eddiemuc: Good catch. This was not intended (I followed the suggestion of AndroidStudio and didn't look close enough at the results). I corrected it

eddiemuc:
I just corrected comment of @moving-bits . As soon as build is through I will merge

Issue #11001, opened 2021-06-24
Adventure Lab caches not shown


Lineflyer:
## Bug description 

Users report, that lab caches are not longer loaded.

### Steps to reproduce the problem

- Have LC connector active
- View live map

### Actual result after these steps 

No ALC show up

### Expected result after these steps 

ALC should be shown

## c:geo version 

2021.05.20 , 2021.06.23 and nightly

## Reproducible 

Yes

Lineflyer:
See also:
https://github.com/mirsch/lab2gpx/issues/28

Lineflyer:
There was no upfront communication by Groundspeak (also not regarding a follow up access method), however as posted above (commit linkt) there seems to be an easy "workaround".

@baiti 

If we could include this "workaround" on `release` I could quickly replace the ongoing rollout of the bugfix release with a new one.

baiti:
I am investigating the situation. The referenced "See also"  mention a consumer key. We do not have a consumer key so I am not sure at this moment, what brought the back into the game.

Lineflyer:
See the commit link above (They just used something).

baiti:
> See the commit link above (They just used something).

That is a trivial fix then but what sense does THAT make?

baiti:
should we identify ourselves in the string pattern we use?

mirsch:
just to make things clear a bit:
I (developer of lab2gpx) recorded the network traffic of the official iOS app to see what has changed and am currently using the consumer key of the official iOS app. I don't currently know if the key depends on my device. That is why it is not in the source but is stored in a config file. The config is of course not checked into the repository.
You'll need a valid consumer key to make this work again.

Lineflyer:
Thanks @mirsch for clarification.

OK, then we might need to ask Groundspeak again and until then we will have no more ALC. 
We used the current method with their approval and they promised to contact us with next steps for using the partner API, but they did not and just block the current method. I don't think we should use any "other" consumer key.

baiti:
Update: By just sending the litereal: "THE-TOP-SECRET-CONSUMER-KEY" will not work.

baiti:
> just to make things clear a bit:
> I (developer of lab2gpx) recorded the network traffic of the official iOS app to see what has changed and am currently using the consumer key of the official iOS app. I don't currently know if the key depends on my device. That is why it is not in the source but is stored in a config file. The config is of course not checked into the repository.
> You'll need a valid consumer key to make this work again.

[...]

PR #11021, opened 2021-06-27
issue_11001 improved

baiti:
attached pls find an update that treats the consumer-key in a more intelligent way. The string constant in the earlier PR had no real meaning after all. This PR provides a way to actually make use of the key.

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -278,6 +278,7 @@
     <string translatable="false" name="pref_localstorage_version">localstorage_version</string>
     <string translatable="false" name="pref_alc_advanced">alc_advanced</string>
     <string translatable="false" name="pref_alc_launcher">alc_launcher</string>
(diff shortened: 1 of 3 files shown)


baiti:
@Lineflyer .... seems like none of you gets notified of new PRs if they aren't linked to a specific issue? Since #11001 is closed, I didn't know what to do. Should I open a new issue such that I could reference it via a tag in the PR?

What do you think with respect to the change being envisioned by this PR? Who wants to review?

Lineflyer:
I was notified...but are on vacation at this moment...more or less only reading some mails and opening new issue while out caching at the moment. Back in 1,5 weeks.

baiti:
> I was notified...but are on vacation at this moment...more or less only reading some mails and opening new issue while out caching at the moment. Back in 1,5 weeks.

ahhh ok. Enjoy!

moving-bits (approved):

moving-bits:

PR #11031, opened 2021-06-28
generify "ok" actionbars in c:geo

fm-sys:
IMHO, it's very important to use the same components for similar use cases. 

While we had the OK/Cancel action bar at many places like "edit PN", "edit waypoints" or "edit coordinates" it was not used at all places. 

This PR will change the action bar for "edit filter" and "edit rout order" to stay consistent in c:geo...

grafik
grafik


main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
```diff
@@ -184,27 +183,24 @@ protected void onRestoreInstanceState(@NonNull final Bundle savedInstanceState)
 
     @Override
     public boolean onCreateOptionsMenu(final Menu menu) {
-        super.onCreateOptionsMenu(menu);
```
eddiemuc, line 187: Call to "super" method should stay imho - unless there's a specific reason to omit it?

eddiemuc (approved):

main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
fm-sys: The super call is not really needed. The only purpose would be if we would have a standard menu defined for all activities and want to extend it. But otherwise, it has no advantages. That's why there's also no lint error like for other "missing super calls"... In other activities we often  don't have super calls for it as well. 

But if you insist I can add it again, as it doesn't hurt either ;-)

eddiemuc: No I am not insisting.

But unfortunately the PR has merge conflicts now

PR #11035, opened 2021-06-28
force singleline in map settings popup

fm-sys:
...to provide a way more cleaned up design experience (and because all other (sub-)titles in c:geo are single-line as well)

before|now
-|-
grafik|grafik

### DIFF
--- a/main/src/cgeo/geocaching/maps/MapSettingsUtils.java
+++ b/main/src/cgeo/geocaching/maps/MapSettingsUtils.java
@@ -70,18 +70,18 @@ public class MapSettingsUtils {
         final LinearLayout leftColumn = columns.get(0);
         final LinearLayout rightColumn = columns.get(1);
 
-        leftColumn.addView(ViewUtils.createTextItem(activity, R.style.map_quicksettings_subtitle, TextParam.id(R.string.map_show_caches_title)));
+        leftColumn.addView(ViewUtils.createSubtitleTextItem(activity, R.style.map_quicksettings_subtitle, TextParam.id(R.string.map_show_caches_title)));
         foundCb.addToViewGroup(activity, leftColumn);
         ownCb.addToViewGroup(activity, leftColumn);
         disabledCb.addToViewGroup(activity, leftColumn);
         archivedCb.addToViewGroup(activity, leftColumn);
         offlineLogCb.addToViewGroup(activity, leftColumn);
 
-        rightColumn.addView(ViewUtils.createTextItem(activity, R.style.map_quicksettings_subtitle, TextParam.id(R.string.map_show_waypoints_title)));
+        rightColumn.addView(ViewUtils.createSubtitleTextItem(activity, R.style.map_quicksettings_subtitle, TextParam.id(R.string.map_show_waypoints_title)));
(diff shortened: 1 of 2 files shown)


Lineflyer:
Not sure about this one.
Imagine longer strings in other languages (and in addition some languages where the important part only comes at the end of the string). I really appreciated, that we removed many places where we cut of strings and IMHO we should not introduce new ones.

Alternatively we should try a clean approach (e.g. by restricting the string length...if this is possible). But this might lead to non understandable abbreviations then.

fm-sys:
I totally understand your request. 

My idea was to make the UI more consistent as we have used single-line textViews for all other (sub-)titles in c:geo as well. 

I just checked the translations: "Afficher les points de passage :" (fr) is the longest one. 

And together with font size set to max:

grafik

To be honest, IMHO this looks _way uglier_ as if the string just get truncated.

In my opinion, the use case of these checkboxes should be very understandable, even if the string is partly invisible...


---
@ CI 
retest this please

Lineflyer:
To be honest I very much dislike apps which just truncate strings and force everything into a "static frame" instead of being responsive in design for such cases. 

Just taking it to the max: In french it would then probably read "Afficher les..." on both items left and right, which provides not the intended information at all. 

How about reducing the source to either only "Show" on both sides or "Show caches/waypoints" as single caption over the full length of the window.

fm-sys:
retest this please

eddiemuc:
Since discussion is ongoing, I would not merge yet?

Lineflyer:
Other opinions on that?

moving-bits:
I think whether an abbreviated string stays understandable depends on the language. In German, a truncated "Wegpunkte anzei..." is ok, especially since you see the complete "anzeigen" close to the left. In other languages this may not work so well.

Thus I would propose to stay with untruncated headings, and maybe ask the translators whether they are able to find a shorter translation.

bekuno:
I think also that a single line text would be better.
How about omitting the "show " part of the headlines, e.g. use the text "Caches", "Waypoints" etc.

Lineflyer:
We introduced "Show...." as the previous behavior was the other way around (i.e. which items to hide).

How about a caption above the curent over the whole width of the window like "Show the following map items". Then we can shorten the below headlines to "Caches" , "Waypoints", "Others".

moving-bits:
> How about a caption above the curent over the whole width of the window like "Show the following map items". Then we can shorten the below headlines to "Caches" , "Waypoints", "Others".

Sounds like a good solution to me, as it will shorten the strings for every language (at the cost of an additional line). Maybe those subheadings (Caches/Waypoints/Others) can be reduced in text size then, so instead of `textSize_headingSecondary` become a (to be created) `textSize_headingTertiary`.

> We introduced "Show...." as the previous behavior was the other way around (i.e. which items to hide).

... and it may become "hide" again in some time to stress that the map quick settings are an additional filter layer applied on top of the regular filter, which can additionally hide some elements, but cannot show additional elements which got excluded by the regular filter.

Lineflyer:
Setting "Do not merge" for now. Lets check a bit later how to proceed.

Lineflyer:
I suggest to close this for now as there was no clear majority for that change.
Lets wait and see user reports after the release is out.

PR #11037, opened 2021-06-28
new UDC icon

fm-sys:
The old icon had a white boarder around it, which was kind of ugly.

Screenshot_20210627_113713_cgeo geocaching

I just have cropped the old UDC icon, which was somehow the fastest way. To make it look well, we should either use a squared icon again, or change the map marker to our round marker version (I would prefer the second way)

grafik

grafik


fm-sys:
retest this please

Issue #10952, opened 2021-06-19
New filter: removing old cache list filter: what is left to do before this step?


eddiemuc:
This issue is about making the new filter system the primary and only filtering system for cache lists. This includes:
* Removing completely the old filter system and associated underlying classes (this is a lot!)
* Remove "experimental" mark from new system

Before we do this, I want to ask around what is still missing in the new filter system to go ahead and do this. Which filters must be added, which important issues must be addressed etc?

Lineflyer:
I did not 1:1 compare the old list filter and the new filter, but you for sure did this.
Anything which is explicitly not covered by the new filter, or anything still on your ToDo?

eddiemuc:
This was meant as a call to see what is really missing, not so much to find the leftover differences between old and new system. And if no one knows a specific filter, it might not be really needed.

But let's see what is still missing. If I did not miss anything it is:
* Found in last 30 days
* Present in different origins with different status (this one would be tricky to implement, I really want to be sure of it's value before that)
* Has vote
* Has trackables
* Has not been found yet
* Saved within/before last 8 hours/24 hours/7 days/30 days
* Has own vote
* Found but has no own vote
* Has changed coordinates (note: this is different from "solved mystery")
* Has user-defined waypoints

I am against implementing all these things just because they exist in the old filter. I would rather like to know which ones are really needed and focus efforts on these.

moving-bits:
I'm currently using this filter a lot:

> Saved within/before last 8 hours/24 hours/7 days/30 days

Reasoning behind that is to find data I need to check / update before going on a tour. (But it's only sort of a workaround, because if I myself change something in a cache (e. g.: user note), the "last save date" gets updated as well.)
But more important: If we are going to persist data we get from the live map (#10927), especially info about availability of a cache, my use case would be covered and I would not need that filter any longer.

moving-bits:
One strategy for those "more exotic" filters could be to leave them out intentionally in a first version and wait for user feedback if any of those filters are actually missed.

eddiemuc:
> One strategy for those "more exotic" filters could be to leave them out intentionally in a first version and wait for user feedback if any of those filters are actually missed.

That would be my personal favorite :-)

If this is considered to be too harsh by the majority, however, I got a second idea overnight: we could explicitly leave the "legacy" filter framework in but reduce it to those "exotic filters". And maybe add some note that this will be removed after one/two releases and users should contact us if they need some of them.

moving-bits:
[...]

PR #11043, opened 2021-06-30
rel to #10952: add things to status filter, add stored since filter

eddiemuc:
rel to #10952: add things to status filter, add stored since filter

Status filter can now filter for "has trackables", "has own vote" and "has offline log":

image

New filter "stored since" allows for filtering of storage date/timestamp:

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -501,6 +501,7 @@
     <string name="cache_filter_location">Location</string>
     <string name="cache_filter_stored_lists">Stored Lists</string>
     <string name="cache_filter_origin">Origin</string>
+    <string name="cache_filter_stored_since">Stored since</string>
 
     <string name="cache_filtergroup_basic">Basic</string>
     <string name="cache_filtergroup_details">Details</string>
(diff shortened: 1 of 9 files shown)


moving-bits:
clean look & very useful, thanks!

moving-bits:
One additional question: Does "favored" mean "I have given a favorite point to this cache"? The we should probably extend the text to something like "own favorite" to distinguish it from "has favorite points" (from whoever)

eddiemuc:
> One additional question: Does "favored" mean "I have given a favorite point to this cache"? The we should probably extend the text to something like "own favorite" to distinguish it from "has favorite points" (from whoever)

Done

moving-bits:
> Done

Thanks

murggel:
I like those emojis in the status filter.
So shouldn't the first three (found, own, stored) also use a colored emoji?
For "found" and "own" the emoji from map...

eddiemuc:
> I like those emojis in the status filter.
> So shouldn't the first three (found, own, stored) also use a colored emoji?
> For "found" and "own" the emoji from map...

Technically this is really easy to achieve of course. 

Right now those are the same icons which are used in the map quick settings. If it is ok to have different setting here then it should not be a problem.

PR #11058, opened 2021-07-02
improve loc switch touch target (another try)

fm-sys:
While testing the latest nightly, I noticed that my first fix in #11038 does not work properly... Therefore, another try. This time it seems fine in the emulator ;-)


fm-sys:
Failing tests are unrelated, let's merge...

PR #11062, opened 2021-07-02
optimize popup titlebar (fix #11022)

### LINKED ISSUE #11022: Optimize popup title bar
IMHO our map popup needs at least still dome finetuning.
In general the redesign is OK (except if we want to replace it by some other material design element e.g. a bottom sheet) but looking at this example we should at least thing about the title bar:
image

- How about omitting the cache name in the title bar...most of the time it does not fit anyway and its now already contained in the content below?
- The cache icon looks a bit narrow compared to the buttons on the right hand side. Does it have the same width/height already?
- What could we show instead of the title? Nothing or something like "Cache info" (if it fits in all languages)?
- It also looks strange, that the height of "More Details" (outline) is more narrow than the outline of the icon next to it.

Other thought:
Would a bottom sheet maybe suitable to replace it completely?
image

After clicking on a cache it could show its name only (one liner as in the example above) and the user can expand it to view the other details... Just thinking...no idea if its suitable and according to the meant usage of this element.

moving-bits:
## Description
- display cache geocode / waypoint "wp code" as popup title
- ~adjust "more details" button height to icon buttons~ (removed from this PR)

image

### DIFF
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -47,6 +47,7 @@
                     style="@style/button_icon"
                     app:icon="@drawable/ic_menu_hint"
                     android:layout_alignParentRight="true"
+                    android:layout_alignBaseline="@id/more_details"
                     android:layout_marginLeft="0dp"
                     android:visibility="gone"
                     tools:visibility="visible"/>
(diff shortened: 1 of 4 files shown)


fm-sys:
> adjust "more details" button height to icon buttons

Somehow, I disagree in this point. We should always try to make our general theming more consistent instead of editing the button height for special buttons, just because it might look a bit better...And IMHO this super-stretched button looks even more strange. I would say it's totally legit that text and icon buttons have different heights. 

If a change is needed at all: What about moving the icon to a different place, e.g. into the button group just below...

moving-bits:
I've opened #11065 to discuss the icon buttons (and the "more details" button in relation to it) and will update this PR to only change the titles.

moving-bits:
retest this please

PR #11064, opened 2021-07-03
replace viewpagerindicator by another one

moving-bits:
## Description
Update viewpager lib on `release` (by cherry-picking @bekuno's PR) to be able to compile `release` again (even if the old viewpager is no longer cached locally)


moving-bits:

Issue #7919, opened 2019-10-27
Sorting by 'date logged' doesn't work


jonbeton:
**Describe the bug:**
When I select 'sort by date logged' on a list of stored caches, sorting seems random.

**To Reproduce:**
1. Go to a list of stored caches (56 items in my case, 13 already found and logged online, 6 found and not yet logged online).
2. Click on 'Sort -> Date Logged'
3. Weird sorting order.

**Actual behavior/state after performing these steps:**
Items are sorted incorrectly, certainly not by 'date logged'.

**Expected behavior/state after performing these steps:**
Items are sorted in order of date logged.

**Version of c:geo used:**
2019.10.22-RC2 (I'm quite sure it's not a new issue)

**Is the problem reproducible:**
Yes

**Screenshots:**
Note: today we found 45 (first) - 42 - 51 - 50 - 49 - 48 (last). Immediately after finding a cache I always make a note 'OK' in c:geo, and then press 'back'. At home, I rely on the date logged when writing and publishing my log message.

First items in list, after sorting by 'Date Logged'
![image](https://user-images.githubusercontent.com/32091273/67638433-9b1b4b80-f8e4-11e9-982a-cbe7754f81d3.png)

Last items in list, after sorting by 'Date Logged'
![image](https://user-images.githubusercontent.com/32091273/67638439-a79fa400-f8e4-11e9-8435-5d8888e7c19e.png)

**System information:**
```
--- System information ---
Device: PRA-LX1 (PRA-LX1, HUAWEI)
Android version: 8.0.0
Android build: PRA-LX1 8.0.0.402(C432)
c:geo version: 2019.10.22-RC2
Google Play services: enabled - 19.5.30 (040408-275531062)
Low power mode: inactive
Compass capabilities: yes
Rotation vector sensor: present
Orientation sensor: present
Magnetometer & Accelerometer sensor: present
Direction sensor used: rotation vector
Hide own/found: false
HW acceleration: enabled (default state)
System language: en_GB
System date format: dd/MM/y
Debug mode active: no
System internal c:geo dir: /data/user/0/cgeo.geocaching (1.5 GB free) internal
User storage c:geo dir: /storage/81B3-1A13/cgeo (12.6 GB free) external removable
Geocache data: /storage/81B3-1A13/Android/data/cgeo.geocaching/files/GeocacheData (12.6 GB free) external removable
Database: /data/user/0/cgeo.geocaching/databases/data (13.5 MB) on system internal storage
[...]

PR #11067, opened 2021-07-03
clear visiteddate on clear offline log

eddiemuc:
clear visiteddate on clear offline log

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -599,6 +599,7 @@ public class Geocache implements IWaypoint {
     public void clearOfflineLog() {
         DataStore.clearLogOffline(geocode);
         setHasLogOffline(false);
+        DataStore.clearVisitDate(Collections.singleton(geocode));
         notifyChange();
     }
 


Lineflyer:
Is there any related issue or can you describe what problem this will solve?

eddiemuc:
As of now, if you create an offline log, save it but then clear it afterwards, the viditeddate will keep set at the date and time of the log entry. This will let the associated cache show up in history list and sorted According to the (already deleted) offline log entry date.

But I have to think this whole topic through again anyway, will set this PR to "draft/wip"

Lineflyer:
Same as me...just thinking if there is any reason for keeping the visitedate and also crawled older issues. But I did not find any reason for that. So I assume this is really a bug.

However, how will it behave in this use case:
- User performs offline log for some caches
- User exports field notes and logs online on the website
- User refreshes caches in c:geo to reflect found state
- User now deletes superfluous offline logs

Will the caches vanish from the history in that case? That would be unexpected. We should not delete a visiteddate if its logged online as found/DNF.

eddiemuc:
@Lineflyer exactly my thoughts.

My current though is that the root problem is the "double usage" of the viditeddate field. It is used to store two different things at the same time:
* The last visited date as available from the website/service. This date has only day-granularity (at least for GC.com)
* The offline log entry date of an existing offline log (if any is there). This is a real timestamp (millisecond-granularity)

In many cases the two things are the same, but in.the case you described they are different. Which then leads to the described problems. Other corner cases are possible

To solve this cleany I fear we have to introduce a second column in the cache table, fill it correctly and change the History-Filter as well as the "date logged" sorting Algo accordingly. A bigger change...

A more hacky solution would be to check on cache refresh whether a viditeddate is already on cache for same date but with timestamp granularity and not update it in those cases. Might be possible, but might get tricky in cache merge logic

eddiemuc:
Copied comments to #7919 for later reference. Will close this PR for now

Issue #11065, opened 2021-07-03
Optimize "icon buttons"


moving-bits:
## What function are you missing
While our new icon buttons looks fine in general, there are still two caveats (IMHO):
- they use more space than the old ones, although the old ones were already relatively large and easy to use (touch target)
- the image itself ("the inner part") is much smaller than with the old buttons, so they are harder to recognize, and there's much unused white space around it

This leads to a third thing, very visible in our cache popup:
The "more details" button does not have a fixed height identical to the icon buttons any longer, which makes this button way smaller, which leads to a strange look of both types are close by. The old look, having same height, was a very straight-forward, clean look, visually stable. But giving the "more details" button the same height as our new icon buttons makes it clumsy and way too big.

![image](https://user-images.githubusercontent.com/3754370/124348914-f7239080-dbec-11eb-9788-92c31f66bc1f.png)
(to the left: new style and two "old" buttons for comparison / to the right: old style)

## Suggested solution
- decrease the icon button size to the same dimensions as the old icon buttons had
- increase the size of the "inner image" to mostly fill the (decreased) dimensions
- adjust "more details" (and any other default button) to have (at least) the same height as the icon buttons

## Alternatives/workarounds
- make "more details" button full width
- find a different placement for the "hint" icon button, maybe by placing "refresh" and "delete" on the same line, and placing "edit list" and "hint" on the second

But that would solve only the third issue.

murggel:
I can tackle the size of the inner icon

murggel:
![image](https://user-images.githubusercontent.com/70113341/124351607-9f405600-dbfb-11eb-98ea-5b741e03b029.png)

I will prepare some more screenshots from other activities...

Lineflyer:
Thanks @murggel for working on this. Don't you think they are now "very massive" e.g. compared to the text size.
The relation between outer border and inner icon size is fine, but now they are rather huge and together with the new solid black look, those icons look somehow oversized for me.

What about rather keep the size of the graphic and have make the outer border smaller? Opinions?

murggel:
Yes, that was my first impression as well... I will play a little bit more with it...

murggel:
Probably using the outlined style instead 🤔

murggel:
outlined:
![image](https://user-images.githubusercontent.com/70113341/124360887-81d7b000-dc2c-11eb-82ee-f76806b0c2ed.png)

murggel:
a little bit smaller:

[...]

PR #11070, opened 2021-07-03
decrease button size, increase icon size

murggel:
## Description

* change button size and icon size:
button size: 42dp
icon size: 30dp
* adapt height of "Details"-button in cache-popup
* align routing-buttons in map-settings

## Related issues

rel #11065

## Additional context

screenshots:
https://github.com/cgeo/cgeo/issues/11065#issuecomment-873455080
https://github.com/cgeo/cgeo/issues/11065#issuecomment-873455127
https://github.com/cgeo/cgeo/issues/11065#issuecomment-873455684
https://github.com/cgeo/cgeo/issues/11065#issuecomment-873455309

### DIFF
--- a/main/res/values/dimens.xml
+++ b/main/res/values/dimens.xml
@@ -42,7 +42,10 @@
 
     <!-- buttons -->
     <dimen name="textSize_buttonsPrimary">@dimen/textSize_detailsSecondary</dimen>
-w
+
+    <!-- icons -->
+    <dimen name="icon_size">42dp</dimen>
+
     <!-- all other text -->
     <dimen name="textSize_detailsPrimary">16sp</dimen>
     <dimen name="textSize_detailsSecondary">14sp</dimen>
--- a/main/res/layout/cache_filter_list_item.xml
+++ b/main/res/layout/cache_filter_list_item.xml
@@ -46,8 +46,8 @@
 
         <ImageView
             android:id="@+id/filter_drag"
-            android:layout_width="40dp"
-            android:layout_height="40dp"
+            android:layout_width="@dimen/icon_size"
+            android:layout_height="@dimen/icon_size"
             android:layout_centerVertical="true"
             android:layout_marginLeft="6dp"
             android:layout_marginRight="6dp"
--- a/main/res/layout/imagelist_item.xml
+++ b/main/res/layout/imagelist_item.xml
@@ -82,8 +82,8 @@
 
         <ImageView
             android:id="@+id/image_delete"
-            android:layout_width="40dp"
-            android:layout_height="40dp"
+            android:layout_width="@dimen/icon_size"
+            android:layout_height="@dimen/icon_size"
             android:layout_centerVertical="true"
             android:layout_marginLeft="6dp"
             android:scaleType="centerInside"
@@ -94,8 +94,8 @@
 
         <ImageView
             android:id="@+id/image_drag"
-            android:layout_width="40dp"
-            android:layout_height="40dp"
+            android:layout_width="@dimen/icon_size"
+            android:layout_height="@dimen/icon_size"
             android:layout_centerVertical="true"
             android:layout_marginLeft="6dp"
             android:layout_marginRight="6dp"
--- a/main/res/layout/map_settings_dialog.xml
+++ b/main/res/layout/map_settings_dialog.xml
@@ -82,6 +82,7 @@
             <Button
                 android:id="@+id/routing_off"
                 style="@style/button_icon_accent"
+                android:layout_margin="0dp"
                 app:icon="@drawable/routing_off" />
 
             <Button
--- a/main/res/layout/popup.xml
+++ b/main/res/layout/popup.xml
@@ -37,9 +37,10 @@
                     android:id="@+id/more_details"
                     style="@style/button_small"
                     android:layout_width="fill_parent"
-                    android:layout_height="40dip"
-                    android:layout_toLeftOf="@id/offline_hint"
+                    android:layout_height="fill_parent"
                     android:layout_gravity="center_horizontal"
+                    android:layout_toLeftOf="@id/offline_hint"
+                    android:insetTop="0dp"
                     android:text="@string/popup_more" />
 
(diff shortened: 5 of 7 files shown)


murggel:
Have to check it for API 21

murggel:
ready for review

main/res/values/dimens.xml
```diff
@@ -42,7 +42,10 @@
 
     <!-- buttons -->
     <dimen name="textSize_buttonsPrimary">@dimen/textSize_detailsSecondary</dimen>
-w
+
+    <!-- icons -->
+    <dimen name="icon_size">42dp</dimen>
```
moving-bits, line 47: just a minor thing: the icon itself has a size of `30dp` (as defined in the style below), `42dp` is for the whole button, so maybe it should be called something like `iconButtonSize`, and the setting could be in the same group as `textSize_buttonsPrimary`.

moving-bits (approved):
LGTM, thanks @murggel for tackling the icon button issue!

eddiemuc (approved):
LGTM

Lineflyer:
Approved by two members. Merging.

main/res/values/dimens.xml
murggel: Will fix that in one of the next PRs...

fm-sys:
Sorry, I missed the discussion ccompletely but I'm convinced that the button size is now way too small! 

The larger button size was one of the main improvements which came along with the new material theme. And while using the latest nightly today while my phone was in an waterproof case I really had trouble to get the correct button pressed. Sometimes, it took five times until the button did react. While it was no problem at all with the previous versions... 

We simply cannot completly ignore usability, just because we want the icon buttons in the same size as text buttons. 

Screenshot_20210705_203427_org mozilla firefox

murggel:
I read something about sizes from 42dp and higher.
But of course we can change back to 48dp and e.g. 32dp for the icon.

It is a attribte in dimens.xml, so easy to change.

moving-bits:
Never had any problems with the buttons we have in our current release, and those are much smaller as those we have in current master. The "old new" buttons had been too large compared to the rest of the page, IMHO (but, agreed, that's also a matter of taste). I'll do a test on a physical device, just started charging an old mobile :-)

IIRC 48dp does not necessarily mean the button (or touch target) itself, but includes the margins. So a button with 42dp and 3dp margin on either side should be fine.

Lineflyer:
I also think the current size is OK: Larger than the "legacy" ones and easily clickable. The shouldn't be smaller though...so we used kind of the minimum. Not having measured it, I guess we are at about 7mm for the smallest?

murggel:
The size of the action-bar is 45dp, shall we use that size?

fm-sys:
> The size of the action-bar is 45dip, shall we use that size?

I guess, that would be a good compromise. With the "old new" we where at about 9mm and with the "new New" we are exactly at almost 7mm (have just messured). So 8mm (45dp) sounds like a good compromise ;-)

fm-sys:
> dip and dp is the same, correct?

Yes. At least AFAIK...

moving-bits:
To cite another issue's entry: We are supposed to [haggle](https://www.youtube.com/watch?v=-2iZjxSGca8)... ;-)

moving-bits:
> dip and dp is the same, correct?

Yes, `dip` seems to be the older name for it, current documentation uses `dp` (AFAICS).

This article has a nice overview: https://medium.com/analytics-vidhya/what-is-the-difference-between-px-dip-dp-and-sp-e4351fefa685

Lineflyer:
Interesting BTW: Learned from the article , that we should use "sp" to define font sizes to allow users to modify them as per their system font settings. Not sure if we do that at all places where it would be meaningful (e.g. cache description).

murggel:
> So a button with 42dp and 3dp margin on either side should be fine.

That's exactly the "new" button size 🤔

fm-sys:
> > So a button with 42dp and 3dp margin on either side should be fine.
> 
> That's exactly the "new" button size 🤔

Problem is, that the margin isn't part of the touch target... :-/

moving-bits:
> Interesting BTW: Learned from the article , that we should use "sp" to define font sizes to allow users to modify them as per their system font settings. Not sure if we do that at all places where it would be meaningful (e.g. cache description).

This is our font size definition since our UI overhaul, defined in `res/values/dimens.xml`:

```
<!-- font size dimensions ====================================== -->

    <!-- all toolbars except main screen -->
    <dimen name="textSize_toolbarSingle">20sp</dimen>
    <dimen name="textSize_toolbarPrimary">18sp</dimen>
    <dimen name="textSize_toolbarSecondary">16sp</dimen>

    <!-- headings (not toolbar) -->
    <dimen name="textSize_headingPrimary">20sp</dimen>
    <dimen name="textSize_headingSecondary">18sp</dimen>

    <!-- lists -->
    <dimen name="textSize_listsPrimary">20sp</dimen>
    <dimen name="textSize_listsSecondary">16sp</dimen>

    <!-- calculator labels between input buttons -->
    <dimen name="textSize_calculator">22sp</dimen>

    <!--- wizard -->
    <dimen name="textSize_headingWizard">26sp</dimen>
    <dimen name="textSize_detailsWizard">20sp</dimen>

    <!-- buttons -->
    <dimen name="textSize_buttonsPrimary">@dimen/textSize_detailsSecondary</dimen>

    <!-- all other text -->
    <dimen name="textSize_detailsPrimary">16sp</dimen>
    <dimen name="textSize_detailsSecondary">14sp</dimen>
```

fm-sys:
BTW, just discovered a "missing single-line parameter" bug which we should fix on the go...

grafik

Lineflyer:
@moving-bits Looks perfect....except...do we want button text to resize if the user selects larger fonts....have to try it out next days how this will look like

fm-sys:
I've just created PR #11107. If the beta is not yet compiling, this should be included as well. Especially because of some ugly quirks in the current build...

Screenshot_20210705_220727_cgeo geocaching_copy_389x222

PR #11071, opened 2021-07-03
adjust cache info hint color (rel. to #11069)

moving-bits:
## Description (1)
- regarding map cache popup's and cache details' info section
- distinguish colors of name and value sections

dark mode:
image.image

light mode:
image.image

## Description (2)
- adjusts colors on waypoints page
- removes unnecessary space above waypoints on waypoints page

image


fm-sys:
In light mode, the gray text color is quite hard to read. Maybe, we can do some fine tuning here?

moving-bits:
A couple of proposals:

image.image.image
(Coming from #ffaaaaaa as color, those pics are made with 9/8/7 instead of a)

or even a bit darker:
image.image
(this one uses #ff666666 as hint color)

fm-sys:
I like the `#ff666666` most

moving-bits:
> I like the `#ff666666` most

Wish granted ;-)
see PR #11085

PR #11077, opened 2021-07-03
top-align name/value pairs in cache/trackable infos

moving-bits:
## Description
Top-align name and info pairs instead of vertically centering them.
Does not make any difference, if info is short, but for longer info both columns will align neatly now (see "Name" and "Typ" entries in screenshots below).

|before|now|
|---|---|
|image|image|


Lineflyer (approved):
LGTM. Failing test is unrelated.

PR #11080, opened 2021-07-03
use same icons in filter and quick settings

### LINKED ISSUE #11060: Use same icons in filter and quick settings
> I like those emojis in the status filter.
> So shouldn't the first three (found, own, stored) also use a colored emoji?
> For "found" and "own" the emoji from map...

Technically this is really easy to achieve of course. 

Right now those are the same icons which are used in the map quick settings. If it is ok to have different setting here then it should not be a problem.

_Originally posted by @eddiemuc in https://github.com/cgeo/cgeo/pull/11043#issuecomment-873216523_

murggel:
## Description

* use icons from map for "Found", "Owned" and "Has Offline Log" in FilterView and QuickSettings
* Use different icons for "On watchlist" and "Own favorite"

image  image

API 21
image

## Related issues

fix #11060

## Additional context

Similar to GC
image

### DIFF
--- a/main/res/drawable/map_circle.xml
+++ b/main/res/drawable/map_circle.xml
@@ -4,7 +4,7 @@
     android:height="24dp"
     android:viewportHeight="24"
     android:viewportWidth="24"
-    android:tint="?android:textColorPrimary">
+    android:tint="@color/colorAccent">
   <path
       android:fillColor="#FF000000"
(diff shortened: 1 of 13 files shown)


Lineflyer:
Looks nice. 
Only CI has a warning about an unused import:


MapSettingsUtils.java:15, UnusedImportsCheck, Priorität: Normal
--
Unused import - cgeo.geocaching.utils.EmojiUtils

murggel:
Can someone fix that please, today I am at my computer only in the evening...

moving-bits:
> Can someone fix that please, today I am at my computer only in the evening...

I've removed the offending line. Let's see if this has worked...

moving-bits:

Issue #11044, opened 2021-06-30
Four digit favorite points truncated in list view


Lineflyer:
On my Samsung Galaxy S20 caches with more than 999 favorite points are truncated in cache lists.
Example: GC7VM56

We had this years ago and adapted the view accordingly. The recent changes seem to break it.

MagpieFourtyTwo:
Seen this, too, multiple times - but IIRC common sense of the mentioned discussion was that it's not really important if a cache has 1021, 1022 or 1120 favos: Once it has crossed 1k, it can be taken as granted that it will most probably be a good one anyway.  ;)  So it was finally decided that it's better to leave the column narrower to save space for the name.

Although I think it was said to present them as "1k+" or alike ...

However, we now have multi line name display, so space is not really an issue anymore ... thus: +1  ;)

P.S.: GC2586K may help as an example for an even wider favo count.  ;)

Lineflyer:
We made it 10k but show until 9999 upon users request after showing 1k only initially.

Lineflyer:
You can check five digit shorting here: https://coord.info/GC13Y2Y

MagpieFourtyTwo:
Yeah, remember, you're right. Thanks for clarifying.

MagpieFourtyTwo:
1k+ favos can be seen now, but in fact the display field is a little bit to narrow: The last digit can only be seen half (or even less):

![FavCount](https://user-images.githubusercontent.com/9889015/124369621-e9fab600-dc6d-11eb-8e27-a7895f33ad17.jpg)

While the screenshot looks even better over here than it is on the mobile (especially if it has waterfall edges) - then the last digit is almost impossible to recognize. With some digit combinations it even looks like a three digit fav count with some strange dots behind it.

Thus I would suggest to make the entire filed a bit wider ...

moving-bits:
> Thus I would suggest to make the entire filed a bit wider ...

see PR #11082

PR #11082, opened 2021-07-04
enlarge favo column just another bit (rel. to #11044)

moving-bits:
## Description
Give favo column another 5dp to avoid truncating a four-digit favo count, see https://github.com/cgeo/cgeo/issues/11044#issuecomment-873489753


moving-bits:
failing test is unrelated, but what's going on in our CI?

PR #11085, opened 2021-07-04
fine-tune cache hint color for light mode (rel. to #11069)

moving-bits:
## Description
Adjusts hint color used for cache info (popup / cache details) for light mode, see https://github.com/cgeo/cgeo/pull/11071#issuecomment-873540732


moving-bits:

PR #11088, opened 2021-07-04
adjust cacheslist spinner formatting (fix #10982)

### LINKED ISSUE #10982: Material Theming: List of Lists needs reformatting
## Bug description 
When being in a list and opening the drop down list to open another list, this drop down list is hard to read:
- no separator between list entries
- no color differentiation or border, thus hard to separate from the background (dark theme)
- fonts way too big

### Expected result
- List's background may perhaps be same grey as in the ``Pick lists`` dialog popup
- Fonts should be comparable to the ``Pick lists`` dialog popup
- Separator at least like in the lists of caches

moving-bits:
## Description
- reverts additional spacing between entries to original value
- uses hint color for "number of caches" line
- see https://github.com/cgeo/cgeo/issues/10982#issuecomment-873549971 for screenshots

### DIFF
--- a/main/res/layout/cachelist_spinneritem.xml
+++ b/main/res/layout/cachelist_spinneritem.xml
@@ -3,7 +3,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     style="?attr/spinnerDropDownItemStyle"
     android:layout_width="match_parent"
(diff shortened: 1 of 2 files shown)


moving-bits:
it's always the same test failing on CI currently, but unrelated to this PR, so let's merge

PR #11091, opened 2021-07-04
add queries section to AndroidManifest.xml (fix #11079)

### LINKED ISSUE #11079: Radar no longer available as navigation method
I have set up my default navigation tool (for click on the compass icon in a cache or a waypoint) to use "Radar" (using the app "GPS Status").

This worked well with the nightlies over the last week.
However after installing the latest nightly (may also be 1-2 days ago) it stopped working but the compass is opened instead.

Looking into Settings - Routing/Navigation - Default navigation the option "Radar" can no longer be selected. It is also no longer available when using the menu function "Navigate" on a cache.

No changes with respect to the installed app "GPS Status" or other device settings.
Maybe its related to SDK30?

moving-bits:
## Description
- Adds a "queries" section to `AndroidManifest.xml` to remain being able to query certain packages with the PackageManager even with target SDK 30
- I've integrated all packages declared in `strings_not_translateble.xml` package section as well as some more I found in the `apps/navi` folder
- Feel free to add additional package names necessary here, I'll update the PR then

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -929,4 +929,23 @@
         <receiver android:name=".settings.ShareBroadcastReceiver"/>
     </application>
 
+    <queries>
+        <!-- packages we want to see using the package manager need to be declared starting with targetSDK 30, and we need to use specific strings / cannot use @string/xxx notation -->
+        <package android:name="btools.routingapp" />
+        <package android:name="cgeo.contacts" />
+        <package android:name="com.eclipsim.gpsstatus2" />
+        <package android:name="com.google.android.apps.translate" />
+        <package android:name="com.google.android.street" />
+        <package android:name="com.google.zxing.client.android" />
+        <package android:name="com.groundspeak.react.adventures" />
+        <package android:name="com.silentlexx.gpslock" />
+        <package android:name="com.sygic.aura" />
+        <package android:name="com.sygic.aura_voucher" />
+        <package android:name="com.webmajstr.pebble_gc" />
+        <package android:name="de.wolffire.chirpwolf" />
+        <package android:name="googoo.android.btgps" />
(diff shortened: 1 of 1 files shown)


fm-sys:
There seem to be some more apps we are interacting with. At a first glance, I miss e.g. OruxMaps, OsmAnd, ... 

Furthermore, there are many more Locus versions, but AFAIK we don't need to add them all, as they are merged with the locus intigration library into  our manifest anyway... 

Screenshot_20210704_144504_cgeo geocaching_copy_432x1290

moving-bits:
> There seem to be some more apps we are interacting with. At a first glance, I miss e.g. OruxMaps, OsmAnd, ...

Do you have package names for them at hand?

> Furthermore, there are many more Locus versions, but AFAIK we don't need to add them all, as they are merged with the locus intigration library into our manifest anyway...

At least in the AndroidManifest.xml I'd seen while working on this PR, the only package name that got added automatically was Google Maps. IIRC the integration lib does search for several package names on trying to establish a connection, but I haven't seen adding anything to the manifest file.

fm-sys:
> IIRC the integration lib does search for several package names on trying to establish a connection, but I haven't seen adding anything to the manifest file.

Ok, could be. Haven't investigated further.



> > OruxMaps, OsmAnd, ...
> 
> Do you have package names for them at hand?

Don't know if these apps are queried at all but:

net.osmand
com.orux.oruxmapsDonate

BTW, you can easily get the id by looking at the playstore URL in the browser, e.g. https://play.google.com/store/apps/details?id=net.osmand&hl=de&gl=US

I think we should look through the whole navigation list, and check if we have forgot something...

eddiemuc:
Note that `queries` also supports specification of intents rather than package names. How are we contacting all those other apps? If it is via intent then this might be the better choice.

Issue #11101, opened 2021-07-04
Clarify merging effects for new changelogs


Lineflyer:
We just did the first merge of `master` to `release` while the changed changelog system is in place (localized changelogs).

As of today I am not sure how Crowdin reacts if I empty `changelog_master` and move the content to`changelog_release`. IMHO this could/would trigger new translations as it might just be considered a new file while the contained strings are already translated.
While discussing those changelos changes in recent past I came to the conclusion it would not be a problem, however now I might either have a block in my head or it does not work such easy.

Need to think about it...feel free to spend ideas or correct me

Right now I think we might need to go with one single changelog file which is not merged between branches by merge rule.
The only drawback would be, that in our nightlies the changes coming from `release` are not included.

BTW: @moving-bits Did you document somewhere how version_release.txt works?

Lineflyer:
My current plan is as follows:

Method 1: Merge the prepared changelog PR and see whats happening on Crowdin.
Good case: Strings in swapped files are recognized and the next Crowdin PR will just follow my PR for all languages.
Bad case: Strings will be untranslated again on Crowdin.

Method 2:
If method 1 does not work I will manually swap the changelogs for all languages via another PR.

moving-bits:
> Method 1: Merge the prepared changelog PR and see whats happening on Crowdin.

One could test this with a single string.

What does happen today when merging `strings.xml` from `master` to `release`. AFAIK no retranslation is triggered, so at least for XML files it seems to work. Not sure about multi-line .md files, though.

> Method 2:
> If method 1 does not work I will manually swap the changelogs for all languages via another PR.

AFAIK you cannot "implant" translations by this, can you?

Aside from that, the changelog needs to be adapted anyway, as we call the new filtering system "experimental" and note, that the old one is still available. Fixing this will trigger a couple of new translations...

Lineflyer:
> What does happen today when merging strings.xml from master to release. AFAIK no retranslation is triggered, so at least for XML files it seems to work. Not sure about multi-line .md files, though.

Right, the change of branch is not the problem and is handled fine by Crowdin. 

The problem is that we move the content of files here.

Lineflyer:
> AFAIK you cannot "implant" translations by this, can you?

As long as I do not Import anything from Crowdin we will after method 2 at least have all existing translations.

moving-bits:
[...]

PR #11102, opened 2021-07-04
Merge changelogs and burn in version code

Lineflyer:
This PR represents the normal process to merge changelogs after merging branches.

Related to #11101


Lineflyer:
I think the best solution for now is to perform the same process also for all translated files. This will offer us a ready to release translated changelog independently from what Crowdin might deliver afterwards.
Will do that in the evening, if there is no better idea.

@moving-bits Will then also try to fill the version.txt with the plain number and check.

moving-bits:
Just checked the source:
- if `changelog_master.md5` is empty, no block will be shown for it
- content of `version_release.txt` will be shown next (or "Next release", if empty), formatted as heading 2
- content of `changelog_release.md5` will be shown below that heading

So this PR LGTM.

moving-bits (approved):

Lineflyer:
OK.
Need to carefully look at Crowdin delivery after merging. Unfortunately it will then also contain the changes I made (remove "experimental", etc.).

Issue #11020, opened 2021-06-27
New Filter: remove legacy global filter support (legacy ISearchBy... capabilities, cachelisttypes, remove "global CacheType" from connectors)


eddiemuc:
With switching online capability search also to new filter, many legacy ISearchByxxx interfaces are not longer necessary and shall be removed:
* ISearchByKeyword
* ISearchByOwner
* ISearchByFinder
* ISearchByCenter
* ISearchByNextPage (TODO: needs to satay potentially for "SearchByFinder" and "SearchPocketQuery")

ISearchByFilter shall be extended by a method "getFilterCapabilities()" returning an enum Set informing about the filter capabilities supported by the connector (e.g. KEYWORD, OWNER, FINDER)

ISearchByViewport might be renamed to ISearchForMap to make clear that this is the method to use for map viewer search (which is the de facto usage today). This can be the only method still using the "map quick settings" filter also inside the connector (if connector supports this, e.g. currently only gc.com and oc)

By same call, number of Loaders for Cache List Activity  + Cache List Types can drastically be reduced (currently a 1:1 relationship to above search capabilities). Also, this will automatically build in the capability to re-issue online search when changing the filter afterwards in cache list activity

Overall, this should reduce code compexity in connectors as well as cache list activity and allow connectors to more flexible make their search capabilities available to c:geo - always according to their individual possibilities.

eddiemuc:
By same call, the number of cachelisttypes/loaders can be reduced.

Left TODOs:

- [ ] remove ISearchByInterfaces
- [ ] remove the interfaces, connector logic etc
- [ ] remove associated cachelisttypes
- [ ] find a way to transport parameters better for "online" capability

TODO llist: implemented connectors

 GCConnector
 OC/OCApi connectors
 SUConnector
 ECConnector
 LCConnector
TODO list: legacied ISearchByInterfaces

 ISearchByKeyword
 ISearchByCenter
 ISearchByOwner
 ISearchByFinder
TODO list: cachelisttypes:

COORDS
ADDRESS
NEAREST
OWNER
FINDER
KEYWORD
SEARCH_FILTER (remains...)

eddiemuc:
TODO: also add filter capability in cachelistactivity to re-filter server result locally for online searches, including the search criteria (not necessary right now since all connectors support all searches natively)

eddiemuc:
Fixed by PR #11918

PR #11115, opened 2021-07-06
fix #11103: remove "search by found-by" and "caches found"

### LINKED ISSUE #11103: New Filter Framework: remove "Found by" search?
While working on new filter framework I discovered a potential problem:

c:geo offers currently a lot of "Search by" functions in the search activity. Those were clearly originally copied from the old/legacy gc.com-search here: https://www.geocaching.com/seek/nearest.aspx

Among them is the search function "Found by" (Gefunden von):

image

Now the problem: 
* this kind of search is not supported by the new filter framework for geocaching.com
* It **can't** be supported by the new framework because the web proxy search interface used does not offer the possibility to search by "Found by" (it supports only search by "not found by")

Since gc.com is our "major service" I see two possible solutions out of this
* Remove "Found by" search function from c:geo. As said before this was clearly only introduced anyway because it was once available on (now legacy) gc.com search, and it is no longer available to the users via "normal" website
* Leave legacy search for this specific function. However, new filter framework can NOT be used in interconnection

I am strongly suggest the first solution (remove search function). What is your opinion?

eddiemuc:
fix #11103: 
This PR will remove the function "search by found-by":

In the following screenshot you can see that it can no longer be seen :-)
(was between "Keywords" search and "Hidden by" search)

image

### DIFF
--- a/main/res/layout/search_activity.xml
+++ b/main/res/layout/search_activity.xml
@@ -106,27 +106,6 @@
             <View style="@style/separator_horizontal" />
             <TextView
                 style="@style/separator_horizontal_headline"
-                android:text="@string/search_fbu" />
-        </RelativeLayout>
-
-        <com.google.android.material.textfield.TextInputLayout style="@style/textinput_dropdown">
-            <AutoCompleteTextView
-                android:id="@+id/finder"
-                style="@style/textinput_embedded"
-                android:hint="@string/search_fbu_prefill"
-                android:imeOptions="actionGo" />
-        </com.google.android.material.textfield.TextInputLayout>
-
-        <Button
-            android:id="@+id/search_finder"
-            style="@style/button_full"
-            android:text="@string/search_fbu_button" />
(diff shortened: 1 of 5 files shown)


eddiemuc:
This PR will only remove the function "on the surface". See #11020 for removal of the underlying functional code (will be done once it is definitely clear that it is no longer needed)

eddiemuc:
I will set this to draft for now. I completely forgot the "found"/"owned" searches from user context menu in log.

eddiemuc:
I am re-enabling this PR. It will now completey remove the functionality to search for found caches, also in user conterxt menu:

image

(In screenshot you can see that you don't see "Cached found" any more :-))

eddiemuc:
This should be merged only if we are absolutely sure that "search by found-by" and user "caches found" can be removed from c:geo

eddiemuc:
Off we go...

Issue #11133, opened 2021-07-07
Users claim found/disabled/archived caches are not filtered according to setting


Lineflyer:
I got various reports, that something is not working as expected by users. Still trying to find out more details from users.

To me it looks like the problem is, that the (formerly?) global hiding of found/disabled/archived is no longer working when performing various searches. Is it like that intentionally due to our rework of filters?

The main problem seems to be, that users always had set to hide found, disabled, archived and suddenly those caches show up and they don't know what to do or why they now have to apply (advanced) filters for it. 

Either we underestimated the change or there is a bug still in?

I did not find the time (due to vacation) to try the behavior myself... will do that later on.

eddiemuc:
Indeed the usage of these global settings was intentionally removed. They are now only used for map quick settings (and unfortunately for the "found by" search which can't be switched to new filtering system).

But I.forgot to remove the settings. Should I do that? Or should we have another discussion wrt global filter options?

Lineflyer:
I am afraid, that we should rethink it a little bit:
For the map we found a good way to keep those settings and use them almost in the same way. For the list they were also not used before so its not unexpected.
But we forgot that users expected those settings to be also used for searches (as they were until now).

At least we could make them available in the basic filter screen, not requiring users to deep dive into advanced mode.
And maybe we should migrate the legacy global settings into the search result filter (is there such a dedicated filter or do we now only have one filter set) on upgrade?

eddiemuc:
> I am afraid, that we should rethink it a little bit:
> For the map we found a good way to keep those settings and use them almost in the same way. For the list they were also not used before so its not unexpected.
> But we forgot that users expected those settings to be also used for searches (as they were until now).

Always open to rediscuss think. 
But we did really not forget about the searches, there were quite some discussions about this (for example oin #10951 but also in other places) and some users claiming to leave things as is for the searches (I think @magpiefourtytwo was among them).

Well, this is partly what betas are for: to rethink things :-) 

> At least we could make them available in the basic filter screen, not requiring users to deep dive into advanced mode.

The Status filter grew very big by now, many special options in it today, really include it in basic filter?

> And maybe we should migrate the legacy global settings into the search result filter (is there such a dedicated filter or do we now only have one filter set) on upgrade?

There is no dedicated filter for searches, just one global filter setting for everything. No context-specific filters any more.

I guess such a migration is possible. Could you describe its logic more precisely in an own issue? (you have to find a mapping from existing global settings into the new filters)

[...]

PR #11148, opened 2021-07-09
rel to #11133: add simplified status filter to basic filter

eddiemuc:
rel to #11133: add simplified status filter to basic filter

This PR adds a simplified version of the "Status" filter to the basic filter view. The basic filter view will then look as follows:

image
image

I included all elements which are also qvailable in map quick settings.

Downside is that the basic filter is now longer than a common phone screen


eddiemuc:
I am really unsure whether this feature request is wanted or not, thus hesitating to merge...

fm-sys:
I personally really like it, but let's keep it open for some days... (maybe until our dev meeting is done?)

Lineflyer:
I think/hope it will be part of our clarification on the meeting. 
I do personally also like it and IMHO it was a oversight anyway not to include it as "basic set".

PR #11152, opened 2021-07-09
override some colors in caches list (fix #11145)

### LINKED ISSUE #11145: Fav count has wrong color in dark mode
In current nightly the fav counter on cache lists in dark mode has bad contrast (especially if the box is colored due to GCVote ratings). I guess the color was changed from white to grey accidentally during the latest list font changes.

It should be white again.

White mode is OK= black.

moving-bits:
## Description
Override some colors (\*) in caches list for better contrast
(\*) distance, tb count, favo count

image.image


Lineflyer:
Looks much better IMHO.

PR #11168, opened 2021-07-10
convert leftover old edittext fields to material text layouts in distance and log entry filter

eddiemuc:
Convert leftover old edittext fields to material text layouts in distance and log entry filter:

image

### DIFF
--- a/main/res/layout/cache_filter_activity.xml
+++ b/main/res/layout/cache_filter_activity.xml
@@ -66,9 +66,11 @@
 
         <TextView
             android:id="@+id/filter_storage_name"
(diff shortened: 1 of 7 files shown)


eddiemuc:
@moving-bits  I saw that you chose to restore/duplicate the method `Dialogs.input` instead of using/enhancing the new methid `SimpleDialog.input`. Any particular reason for that?

For now I left the methods, but extracted the text field creation into `ViewUtils.createTextField` - it is usable in many places.

moving-bits:
No particular reason, just had not seen it at that time. Feel free to either merge it into `SimpleDialog.input` (or leave it as is for now).

PR #11181, opened 2021-07-11
make system info buttons non-scrolling (fix #11151)

### LINKED ISSUE #11151: System information has bad contrast
The system information has a gray text color, which looks like as if it is "currently disabled". I would suggest using white color instead like in all other pages of "about c:geo"...

moving-bits:
## Description
make the two buttons in about:system info fixed (non-scrolling)


fm-sys:
What about making the "create logfile" button also always visible? 

I would still keep the button at the bottom though, so a relative layout is probably needed.

moving-bits:
updated PR to make the "generate logfile" button fixed at bottom and always visible:

image

PR #11182, opened 2021-07-11
remove 'remove list' dialog (fix #11081)

### LINKED ISSUE #11081: Remove "Remove list" question dialog
## What function is annoying you ;)
In the past there already appeared a dialog if the last cache has been removed from a list like ``List is empty. Do you want to remove the current list?`` - which already became kind of annoying, esp. if you have some lists which get empty *on purpose* every single day, like e. g. ``Today`` (when all caches in today's ToDo list have been visited) or ``Offline Logs`` (when all visited caches are logged, refreshed, and then moved to ``All finds``). But this was "just" once a day, so I lived with it.

But things got worse, cause now this dialog additionally appears even on entering such a list via e. g. the ``Stored`` button in main screen or after a filter has been set ... or ... I don't know where else, as I have given up to count the number of clicks I have made just thjs evening to  just acknowledge this dialog with ``No``.  :/

An, no, I *never* *ever* want these lists to be deleted ...  ;)

So, can we *please* do s/th to dam this dialog spam flood? :) By either inventing a setting to disable this dialog or add a [Never] for certain lists which should never be deleted? Or even remove this dialog completely? Especially since one can certainly assume that a user who is able to create a list, is also able to delete it when he (really) no longer needs it.

TMALSS: IMO the current number of questions is near to paternalism and a bit exaggerated ...

## c:geo version
2021.07.04-NB-0168c2e

moving-bits:
## Description
Removes the "remove list" dialog which is not only shown after removing all caches from a list, but also when entering an empty list

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -975,17 +975,6 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         return true;
     }
 
-    private void checkIfEmptyAndRemoveAfterConfirm() {
-        final boolean isNonDefaultList = isConcreteList() && listId != StoredList.STANDARD_LIST_ID;
-        // Check local cacheList first, and Datastore only if needed (because of filtered lists)
-        // Checking is done in this order for performance reasons
-        if (isNonDefaultList && CollectionUtils.isEmpty(cacheList)
-                && DataStore.getAllStoredCachesCount(CacheType.ALL, listId) == 0) {
-            // ask user, if he wants to delete the now empty list
-            SimpleDialog.of(this).setTitle(R.string.list_dialog_remove_title).setMessage(R.string.list_dialog_remove_nowempty).setButtons(SimpleDialog.ButtonTextSet.YES_NO).confirm((dialog, whichButton) -> removeListInternal());
-        }
-    }
-
     private boolean cacheToShow() {
         if (search == null || CollectionUtils.isEmpty(cacheList)) {
             showToast(res.getString(R.string.warn_no_cache_coord));
@@ -2074,15 +2063,16 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         showProgress(false);
         hideLoading();
         invalidateOptionsMenuCompatible();
+        /* expand following block if necessary
         if (arg0 instanceof AbstractSearchLoader) {
             switch (((AbstractSearchLoader) arg0).getAfterLoadAction()) {
                 case CHECK_IF_EMPTY:
-                    checkIfEmptyAndRemoveAfterConfirm();
                     break;
                 case NO_ACTION:
                     break;
             }
         }
+        */
     }
 
     @Override


Lineflyer:
But this removes the dialog completely or do I interpret this wrong?
IMHO it should still be shown immediately after deleting all cache on it.

moving-bits:
See discussion in the related issue. Code for the dialog has not changed (before this PR), but suddenly the dialog pops up even on entering the list. So far nobody has found the reason for it (or nobody else has looked into it), so the proposal was to remove the dialog completely, as it's only a shortcut for a pretty easy task anyway.

moving-bits:
How do we proceed here?

Does anyone have an idea why the popups suddenly come even on entering a list, and not only on having deleted the last cache?

And / or can we live with this popup being removed completely?

Lineflyer:
@moving-bits 
You merged this?
Did you find a solution or is this feature now gone only due to the rare case of users keeping empty lists. Even then it might be helpful except for the case described in the issue (which is pretty rare IMHO).

moving-bits:
Yes, I merged it as no one came up with a different solution, and c:geo currently asks users on every (!) entering such a list, which is very annoying, IMHO.

I guess how annoying one finds this behavior depends on the personal use cases. I for example use empty lists on a regular basis, filling them only while planning or while being on a tour, and resetting them afterwards (but without wanting to recreate them every time).

eddiemuc:
I have the exact same use case as @moving-bits . For this use case the dialog is simply superfluous and unnerving.

There are obviously other usage patterns where lists shall be deleted when empty. I have two ideas for this:
* Define a list name prefix, e.g. "temp-". All lists with this prefix are automatically deleted by Cgeo as soon as they become empty
* Have a switch in settings telling the system to automatically delete all lists when they become empty
* The two above could be combined with a grace period (e.g. delete lists which are empty for more than 3 days)

fm-sys:
> * Define a list name prefix, e.g. "temp-". All lists with this prefix are automatically deleted by Cgeo as soon as they become empty
> 
> 
> * The two above could be combined with a grace period (e.g. delete lists which are empty for more than 3 days)

I think it's better to not introduce even more magic into cgeo.

About another setting: This would be possible for sure, but should we really make settings even more complex? 

For me, the dialog wasn't annoying, but I honestly don't miss it either.

moving-bits:
_If_ we want to re-introduce some magic here, then a more comfortable solution could be to introduce not a global switch but a per-list "do not ask" setting. (We have something similar for caches: "prevent waypoint extraction" or something like this.) Then we could exclude our "today", "found", "needs maintenance" (etc.) lists from being deleted.

Could be combined with 
- either bringing the dialog back (but fix it to show up only after having deleted/removed caches from active list)
- or introducing list-autodelete for empty lists.

Not sure if we want the latter, though, and for the former we still need to find the reason why this dialog suddenly pops up on every entering the list instead of only on deleting/removing caches.

fm-sys:
> * either bringing the dialog back (but fix it to show up only after having deleted/removed caches from active list)
> 
> * or introducing list-autodelete for empty lists.

I think we should never ever delete anything automatically without an explicit user interaction. But a "do not show again" checkbox sounds pretty cool 👍

Lineflyer:
> I think it's better to not introduce even more magic into cgeo.

Fully agree.

 > About another setting: This would be possible for sure, but should we really make settings even more complex?

We should really only provide new settings if a feature needs to be controlled as it blocks relevant usecases otherwise.
In this case it would only be working around a potential bug, so we should not add a setting.
 
> For me, the dialog wasn't annoying, but I honestly don't miss it either.

I am quite sure, many users will ask why its gone.
But after a while it will just be a forgotten convenience feature.

> I think we should never ever delete anything automatically without an explicit user interaction. But a "do not ask again" checkbox per list sounds pretty cool 👍

Deletion automatically is no option IMHO. I somehow also do not like "do not ask again" things for automating such things, because if I (at a later point of time) want to be asked, I need to reinstall/reset the app.

> (but fix it to show up only after having deleted/removed caches from active list)

That would IMHO still be the preferable way (incl. to accept the regression that it will be shown when entering empty lists until we have a fix).

Lineflyer:
May I take a guess here (Attention: Dangerous non-java-programming knowledge :) ):
You sometimes talk about  `onCreate` and `onResume`. My guess would be, that the check to show this dialog should only be contained in `onResume` coming from the menu to delete the caches in it but not in `onCreate` when entering the list. 
Or maybe its some trigger if the view gets " in focus".
I assume some of that code might have changed, therefore triggering the dialog.

PR #11193, opened 2021-07-12
Let status notification background use ActionBarColor

Lineflyer:
## Description

To me it would look more consistent if the status notifcation (new version, etc) would use the action bar color, because then
- on black theme it is no longer in same color as the backgroun
- on white theme its no longer a dark black

|Current master|New look in combination with PR#11192|
|---|---|
|Screenshot_1626089174|Screenshot_1626089650

|Current master|New look in combination with PR#11192|
|---|---|
|Screenshot_1626089187|Screenshot_1626089483

## Related issues

None

## Additional context

Please kindly doublecheck whether this `@drawable/helper_bcg.xml` is used elsewhere. I could not find any other spots yet.

### DIFF
--- a/main/res/drawable/helper_bcg.xml
+++ b/main/res/drawable/helper_bcg.xml
@@ -6,7 +6,7 @@
         android:width="2px"
         android:color="#99FFFFFF" />
 
-    <solid android:color="#FF191919" />
+    <solid android:color="@color/colorBackgroundActionBar" />
 
     <corners
         android:bottomLeftRadius="3dp"


fm-sys (approved):
The background drawable is also used for the `info_notloggedin` but it's more than fine to change that background for consistency as well...

Lineflyer:
> The background drawable is also used for the info_notloggedin but it's more than fine to change that background for consistency as well...

Just out of curiosity. I would have expected, that Android Studio can tell me where it is used, but I could not find this information. How did you check this out?

fm-sys:
This time I only have used github to find it out, as I'm not on my dev machine currently.

https://github.com/cgeo/cgeo/search?q=helper_bcg&type=

But Android studio does offer this information as well when you select the drawable/layout file in the directory explorer on the right and press ALT+F7 or right click and "show usages"

moving-bits (approved):
LGTM

PR #11198, opened 2021-07-12
Fix #11197 - Avoid cut of list spinner with large fonts

### LINKED ISSUE #11197: List spinner cut off with large fonts
Reported on support mail for 2021.07.10-RC.
User is user a bit enlarged system font setting.

image

Will work on that and try a fix

Lineflyer:
## Description

Found a way which seems to fix that issue. Not sure about side effects though due to my limited knowledge of this layout handling.
Therefore please review for unwanted effects.

## Related issues

#11197
## Additional context

|Current release with XL fonts |This PR with XL fonts|
|---|---|
|Screenshot_20210712-151045|Screenshot_20210712-151428

### DIFF
--- a/main/res/layout/cachelist_spinneritem.xml
+++ b/main/res/layout/cachelist_spinneritem.xml
@@ -3,7 +3,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     style="?attr/spinnerDropDownItemStyle"
     android:layout_width="match_parent"
-    android:layout_height="?attr/dropdownListPreferredItemHeight"
+    android:layout_height="wrap_content"
     android:layout_gravity="left|center_vertical"
     android:minHeight="?attr/dropdownListPreferredItemHeight"
     android:orientation="vertical"


fm-sys:
> Not sure about side effects though due to my limited knowledge of this layout handling.

I have never experienced a situation where some layout changes had really critical side affects. Worst thing which could happen is that we need to fine tune ita bit further. When starting with c:geo development I had absolutely no knowledge about android dev and theming either. 

My workflow is still even today Google/Stackoverflow > Try it out in the emulator > if it works create PR and merge. 

So nothing you need to be afraid ;) 

BTW, I really appreciate your help in these theming improvements.  Saves us much time and helps a lot 👍

moving-bits (approved):
LGTM with highest font size setting in emulator. Couldn't test, though, what happens if there's an even larger font setting somewhere, which would make the two lines in the actionbar not fitting any longer. Let's test this aspect in real life :-)

PR #11199, opened 2021-07-12
Avoid cut off favcount and tbcount in lists

Lineflyer:
## Description

I started looking through the layout while explicitly using the largest possible system font on my Google Pixel 3a.
This was one finding I tried to tackle here: The favorite count and also the inventory icon and count were cut off as they had a absolute height. I changed this to `wrap_content`. 
Furthermore added a 1dip margin below tbcount as otherwise it floats into the border of the favorite count.

Please review as these are my first steps into layout changes. :)

|Current release|This PR|
|---|---|
|Screenshot_20210712-153827|Screenshot_20210712-153229

## Related issues

None

### DIFF
--- a/main/res/layout/cacheslist_item.xml
+++ b/main/res/layout/cacheslist_item.xml
@@ -108,7 +108,7 @@
         <view
             android:id="@+id/direction"
             android:layout_width="78dip"
-            android:layout_height="28px"
+            android:layout_height="28dip"
             android:layout_centerHorizontal="true"
             android:layout_gravity="center"
             android:layout_marginTop="21dip"
--- a/main/res/layout/cacheslist_item.xml
+++ b/main/res/layout/cacheslist_item.xml
@@ -146,9 +146,10 @@
         <TextView
             android:id="@+id/inventory"
             android:layout_width="match_parent"
-            android:layout_height="22dip"
+            android:layout_height="wrap_content"
             android:layout_alignParentTop="true"
             android:layout_gravity="center_vertical|center_horizontal"
+            android:layout_marginBottom="1dip"
             app:drawableLeftCompat="@drawable/trackable_all"
             android:drawablePadding="-10sp"
             android:gravity="center"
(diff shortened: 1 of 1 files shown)


Lineflyer:
Note: Did not yet find out how to move away the direction arrow from the distance. Thats another PR as soon as I found it.

Lineflyer:
Fixed the direction arrow now as well by changing height from `px` to `dip`:
Screenshot_20210712-155050

main/res/layout/cacheslist_item.xml
```diff
@@ -108,7 +108,7 @@
         <view
             android:id="@+id/direction"
             android:layout_width="78dip"
-            android:layout_height="28px"
+            android:layout_height="28dip"
             android:layout_centerHorizontal="true"
             android:layout_gravity="center"
             android:layout_marginTop="21dip"
```
fm-sys, line 114: A bit unrelated to your PR but might be interesting for you:

There are two primary layout groups used in cgeo:
- Linear Layout
- Relative layout

In a linear layout all elements are added below each other. So no need to calculate the position manually, but child element can't overlap. 

In a relative layout you are way more flexible and can position elements everywhere you want relative to the view group size. There are parameters like `layout_alignParentBottom`, `layout_below`, `layout_marginTop` and much more. However RelativeLayouts are more komplex to write and take longer to inflate on the end device.

In this case, it seems to me the RelativeLayout is used to simulate a vertical Listview, which is IMHO neither nice to read nor good for performance reasons. If I would have written the code myself, I would have used a LinearLayout instead. If you have time and want to play around a bit more, you could try to refactor the "direction and distance" and "inventory and favorites" layout to a LinearLayout. But if you don't want to tackle this, it's totally fine as well as it had worked the last years with a RelativeLayout as well ;-)

moving-bits (approved):
LGTM

main/res/layout/cacheslist_item.xml
Lineflyer: Thanks for this explanation!!

Issue #11134, opened 2021-07-07
maps.me navigation not available


Lineflyer:
Reported by some beta users, could not test it yet.

With the last version I was able to navigate with "MAPS.ME". This does no longer work.

fm-sys:
Probably, we need to add maps me to the `query` section  in the manifest

Lineflyer:
More detailed user report:
maps.me is shown as navigation option, but when I click it there is a message box, that this app is not installed with a link to install it.
Clicking the link gets to a 404 page

AFAIR we use a kind of API plugin towards maps.me  (some code on our side). Maybe this code has problems with API 30.

The 404 is for sure a problem on their side as https://maps.me/get is linked in that box but not exisiting on their server.
But as the user has maps.me installed, the box should not show up at all.

Lineflyer:
+3 from support mail!

bekuno:
see https://github.com/cgeo/cgeo/blob/release/mapswithme-api/AndroidManifest.xml
package="com.mapwithme.maps.api"

Hope that this is the correct one.

moving-bits:
Thanks, will prepare a PR for it, so that we can try.

bekuno:
Or:   com.mapswithme.maps.pro 
from https://play.google.com/store/apps/details?id=com.mapswithme.maps.pro

bekuno:
By the way, I think that OSMAND was also supported?
See net.osmand - from https://play.google.com/store/apps/details?id=net.osmand

moving-bits:
will add it as well...

bekuno:
and there are more more apps at main/src/cgeo/geocaching/apps/navi - 

=> Oruxmap https://play.google.com/store/apps/details?id=com.orux.oruxmapsDonate

moving-bits:
more to come?

bekuno:
not for now.

PR #11210, opened 2021-07-12
add more package names for Locus app (fix #11200)

### LINKED ISSUE #11200: Show on map -> Locus no longer available
Current nightly 2021.07.12-NB:

- Open a cache list
- Select Menu -> Show on Map

Previously "Locus" was shown here, now its missing.
Maybe also another instance of a missing package name check allowance in our lists?

moving-bits:
## Description
Locus app uses several package names, depending on free/pro state and on distribution channel, see `LocusUtil` that comes with the integration files.
This PR adds the package names yet missing to `AndroidManifest.xml` to be able to query that package on Android 11 systems.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -946,6 +946,12 @@
         <package android:name="de.wolffire.chirpwolf" />
         <package android:name="googoo.android.btgps" />
         <package android:name="menion.android.locus" />
+        <package android:name="menion.android.locus.free.amazon" />
+        <package android:name="menion.android.locus.free.samsung" />
+        <package android:name="menion.android.locus.pro" />
+        <package android:name="menion.android.locus.pro.amazon" />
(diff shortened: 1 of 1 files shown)


bekuno (approved):
LGTM

Lineflyer:
Do we also need the same for maps.me (old name: mapswithme) ?
See #11134

moving-bits:
> Do we also need the same for maps.me (old name: mapswithme) ?
> See #11134

Sure. I just don't know their package names yet.

fm-sys:
You can always easily use the web version of playstore to easily find the package name (contained in the URL)

moving-bits:
> You can always easily use the web version of playstore to easily find the package name (contained in the URL)

Yes, I know, but have been too lazy (and too busy) to do this yet for those apps (and don't use them myself, so didn't know their package names yet).

PR #11211, opened 2021-07-12
show wp count after tab title (fix #11175)

### LINKED ISSUE #11175: Amount of waypoints not visible in tab title from description tab any more
## Bug description 

In the previous cgeo versions one could see the full title of the tab in the tab-bar in the bottom. Currently (probably due to bigger fonts?) only "waypoint" part is visible, but not an actual number of waypoints.

So in order to check how many are there, I really need to switch to waypoints tab, which is not convenient.

Open a cache in cgeo on "Details" tab
Check title of "waypoints" tab
Make sure amount is not visible

## System information 

```
Paste your system information here
```

## Additional context 

Screenshot_20210711-061435

moving-bits:
## Description
Show wp count after the tab title instead of in front of

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1260,15 +1260,7 @@
     <string name="cache_images">Images</string>
     <string name="cache_advanced">Advanced</string>
     <string name="cache_waypoints">Waypoints</string>
-    <plurals name="waypoints">
-        <item quantity="zero">%d waypoints</item>
-        <item quantity="one">%d waypoint</item>
-        <item quantity="two">%d waypoints</item>
-        <item quantity="few">%d waypoints</item>
-        <item quantity="many">%d waypoints</item>
-        <item quantity="other">%d waypoints</item>
-    </plurals>
-
+    <string name="waypoints_tabtitle" tools:ignore="PluralsCandidate">waypoints (%d)</string>
     <string name="cache_waypoints_add">Add Waypoint</string>
     <string name="cache_waypoints_add_currentLocation">Add current location</string>
     <string name="cache_waypoints_add_fromclipboard">Add from clipboard</string>
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -2422,7 +2422,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         // show number of waypoints directly in waypoint title
         if (pageId == Page.WAYPOINTS.id) {
             final int waypointCount = cache == null ? 0 : cache.getWaypoints().size();
-            return res.getQuantityString(R.plurals.waypoints, waypointCount, waypointCount);
+            return String.format(getString(R.string.waypoints_tabtitle), waypointCount);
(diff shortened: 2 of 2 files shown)


main/res/values/strings.xml
```diff
@@ -1261,12 +1261,12 @@
...
     <string name="cache_waypoints">Waypoints</string>
     <plurals name="waypoints">
-        <item quantity="zero">%d waypoints</item>
-        <item quantity="one">%d waypoint</item>
-        <item quantity="two">%d waypoints</item>
-        <item quantity="few">%d waypoints</item>
-        <item quantity="many">%d waypoints</item>
-        <item quantity="other">%d waypoints</item>
+        <item quantity="zero">waypoints (%d)</item>
+        <item quantity="one">waypoint (%d)</item>
```
fm-sys, line 1265: IMHO it should always be "Waypoints" as the number inside the "()" is no longer directly related to the name in front.

fm-sys: So no string array is necessary at all anymore, which also saves time for our translators ;)

moving-bits: Would be ok for English and German, but I don't know about other languages, whether there is sort of a "generic plural" as well.

Lineflyer: Also for german it could formally correct still be "Wegpunkt (1)" but "Wegpunkte (2)"...

Lineflyer (approved):

main/res/values/strings.xml
fm-sys: I think it's not about a "generic plural" but the fact, that the page can show multiple waypoints and not only a single one.

And the number now is totally separated from the generic page name. 

For "logs" we also do not only display "log" if there's only one log for the cache.

fm-sys:
> Also for german it could formally correct still be "Wegpunkt (1)" but "Wegpunkte (2)"...

Don't think so. Since number and name do no longer form a related word combination.

But do we really fight about a single "s"? 😅😂

main/res/values/strings.xml
moving-bits: I understand that the page can display several waypoints, and because of that one could use a generic "waypoints". But I don't know if there is such a "generic plural" in all languages. "generic" in the sense of "not bound to a specific number > 1". For English and German there is no difference if it's 2, 5, 100 of 1 million waypoints, whereas in other language the translation for "waypoints" has to change in those cases.

Lineflyer:
> > Also for german it could formally correct still be "Wegpunkt (1)" but "Wegpunkte (2)"...
> 
> Don't think so. Since number and name do no longer form a related word combination.
> 
> But do we really fight about a single "s"? 😅😂

No fight, just some "Korinthenkackerei". :)
The easy solution (no array) would be preferable I guess.

moving-bits:
I can change it to generic plural, without plurals array, and let our translators decide :-)
Also would add a lint exception to not mock about a string containing a number placeholder, but not plurals array.

moving-bits:
failing lint tests is a result from one string being removed - will be fixed automatically with next CrowdIn run.

PR #11212, opened 2021-07-12
add MapsWithMe package name to AndroidManifest.xml (fix #11134)

### LINKED ISSUE #11134: maps.me navigation not available

moving-bits:
## Description
Adds MapsWithMe package names to `AndroidManifest.xml` to be able to query those packages even when running on Android 11

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -939,6 +939,8 @@
         <package android:name="com.google.android.street" />
         <package android:name="com.google.zxing.client.android" />
         <package android:name="com.groundspeak.react.adventures" />
+        <package android:name="com.mapwithme.maps.app" />
+        <package android:name="com.mapswithme.maps.pro" />
         <package android:name="com.silentlexx.gpslock" />
         <package android:name="com.sygic.aura" />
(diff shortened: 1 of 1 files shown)


moving-bits:
fixed first mwm package name and added osmand

main/AndroidManifest.xml
```diff
@@ -947,6 +949,7 @@
         <package android:name="googoo.android.btgps" />
         <package android:name="menion.android.locus" />
         <package android:name="menion.android.whereyougo" />
+        <package android:name="net.osmand" />
```
Lineflyer, line 952: net.osmand.plus missing

moving-bits: hmpf.

added it :-)

Lineflyer: Does this support some placeholders or regex patterns? I guess not...

moving-bits: not that I know of

and no string placeholders as well, sadly - would be nice, as we have many of those package names already defined in `strings_not_translatable.xml`

PR #11220, opened 2021-07-13
make actionbar font size independent of device settings

fm-sys:
grafikgrafik

Furthermore, I added the "Android Touch" effect to the spinner (like we have at many other places like cache list). I'm unsure about the background color. Both would easily possible:

action bar color (currently done in this PR)|popup background color
-|-
grafik|grafik

### DIFF
--- a/main/res/layout/cachelist_spinner_actionbar.xml
+++ b/main/res/layout/cachelist_spinner_actionbar.xml
@@ -3,12 +3,10 @@
     xmlns:tools="http://schemas.android.com/tools"
     style="?attr/spinnerDropDownItemStyle"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
+    android:layout_height="match_parent"
     android:layout_gravity="left|center_vertical"
-    android:minHeight="?attr/dropdownListPreferredItemHeight"
     android:orientation="vertical"
-    android:background="@color/colorBackgroundActionBar"
-    tools:context=".CacheListSpinnerAdapter" >
+    tools:context=".CacheListSpinnerAdapter">
 
     <TextView
         android:id="@android:id/text1"
@@ -19,7 +17,8 @@
         android:textSize="@dimen/textSize_toolbarPrimary"
         android:textColor="@color/colorTextActionBar"
         android:layout_height="wrap_content"
-        android:maxLines="1" />
+        android:maxLines="1"
+        tools:ignore="SpUsage" />
 
 
     <TextView
@@ -31,6 +30,7 @@
         android:textColor="@color/colorTextHintActionBar"
         tools:text="This is the subtitle"
         android:id="@android:id/text2"
-        android:maxLines="1" />
+        android:maxLines="1"
+        tools:ignore="SpUsage" />
 
 </LinearLayout>
--- a/main/res/values/dimens.xml
+++ b/main/res/values/dimens.xml
@@ -21,9 +21,9 @@
     <!-- font size dimensions ====================================== -->
 
     <!-- all toolbars except main screen -->
-    <dimen name="textSize_toolbarSingle">20sp</dimen>
-    <dimen name="textSize_toolbarPrimary">18sp</dimen>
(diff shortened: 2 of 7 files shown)


Lineflyer:
I would stay with actionbar color as it "comes out of it" from users perspective.

fm-sys:
Ok, fine. That's the same thought I had as well...

moving-bits:
> I would stay with actionbar color as it "comes out of it" from users perspective.

same here.

One additional thing: At least in the left screenshot action bar color and regular background color are pretty close, which makes it hard to distinguish the popup from the rest. Maybe we should make both colors a bit more apart again...?

fm-sys:
> Maybe we should make both colors a bit more apart again...?

Had the same thought. Although I like the darker color very much it's maybe a bit too dark. But probably this would be something for a separate PR...

main/res/layout/cachelist_spinner_actionbar.xml
```diff
@@ -0,0 +1,36 @@
...
+    android:orientation="vertical"
+    tools:context=".CacheListSpinnerAdapter">
+
+    <TextView
+        android:id="@android:id/text1"
+        tools:text="This is the title"
+        android:ellipsize="end"
+        android:layout_width="match_parent"
+        style="?attr/titleTextStyle"
+        android:textSize="@dimen/textSize_toolbarSingle"
```
moving-bits, line 17: shouldn't that be `textSize_toolbarPrimary`?

main/res/values/dimens.xml
```diff
@@ -21,9 +21,9 @@
     <!-- font size dimensions ====================================== -->
 
     <!-- all toolbars except main screen -->
-    <dimen name="textSize_toolbarSingle">20sp</dimen>
-    <dimen name="textSize_toolbarPrimary">18sp</dimen>
-    <dimen name="textSize_toolbarSecondary">16sp</dimen>
+    <!-- (we explicitly use dp for toolbar to avoid cuts offs and follow material guidelines) -->
+    <dimen name="textSize_toolbarSingle">20dp</dimen>
+    <dimen name="textSize_toolbarSecondary">18dp</dimen>
```
moving-bits, line 26: Is there a reason why you remove the differentiation between a single-line actionbar text and a double-lined?

moving-bits:
This PR is about fixing a font size for the action bar, as some people complained their text got truncated when setting a larger font size.

Which leads me to a question we should discuss, IMHO: What do we value higher?
- A user's wish to be able to read at least a part of a text (expressed by increasing the font size system-wide)
- or a user's wish to see as much text as possible in the limited space of an action bar (even if some people may not be able to decipher it because its font size is too small for them).

I tend to the first opinion - if I increase font size to be able to read a text I know that something may get truncated, but I can read the beginning of the text at least - but I'm open to discussion.

main/res/values/dimens.xml
fm-sys: Reason is, that the the textSize_toolbarPrimary was only used for the Cache List but not for the live map, and I wanted to make the text size consistent between different activities. And I didn't found a way to difers between single and multi-line in the actionbar style...

moving-bits: Ok, understood. But then we should stay with `textSize_toolbarPrimary` and `textSize_toolbarSecondary` to have a consistent naming for text sizes.

fm-sys:
The action bar text use a quite big size anyway, so the size is not much smaller than the details text size even when using font size "extra large". I think a cutted off font size looks like a bug, and should always be avoided. BTW, (all/many?) other apps handle it in the same way, so we are definitely not the first with using a fixed font size for actionbar.

main/res/values/dimens.xml
fm-sys: Yes, you're right. will do so...

fm-sys: Done

fm-sys:
Let's just try it out. If users will complain that the font size is too small, we can still try to find a different solution...

PR #11225, opened 2021-07-13
add "+" icon to "add filter criteria"

fm-sys:
The button text is quite long, so the text might not be understandable, especially when using large font sizes:

grafik
grafik

When adding an icon in front, the meaning is much easier to understand in such cases:

grafik


moving-bits:
or do we need text at all? Why not use only a plus sign? (TBD)

fm-sys:
> 
> 
> or do we need text at all? Why not use only a plus sign? (TBD)

It was exactly that way some weeks ago, but there was the request to explicitly add a text to the button.

eddiemuc:
I like it.

PR #11228, opened 2021-07-13
don't show action bar in InstallationWizard

fm-sys:
Don't show Actionbar in InstallationWizard for a cleaner look.

before|now
-|-
grafik|grafik


moving-bits:
lint errors are unrelated, so let's merge

PR #11192, opened 2021-07-12
Fix #11014 - Action bar coloring

Lineflyer:
## Description

Proposal for action bar coloring, screenshots see https://github.com/cgeo/cgeo/issues/11014#issuecomment-878147799

I took:
- FF202020 for dark theme
- FF505050 for white theme

@moving-bits Please check if I understood the mechanic of `values-night` correctly to split up the color. At least it looks working on my emulator.

## Related issues

Fix #11014

PR #11233, opened 2021-07-14
make actionbar bg color a bit lighter in dark theme

moving-bits:
## Description
As discussed in https://github.com/cgeo/cgeo/pull/11220#issuecomment-879400901 we should make the actionbar color a bit lighter in dark theme to be able to differentiate between regular background and actionbar dropdown background color:

|before|after|
|---|---|
|image|image|

### DIFF
--- a/main/res/values-night/colors_dark.xml
+++ b/main/res/values-night/colors_dark.xml
@@ -15,7 +15,7 @@
     <color name="colorBackgroundTransparent">#44000000</color>
     <color name="colorBackgroundDialog">#FF424242</color>
     <color name="colorBackgroundSelected">#FF444444</color>
-    <color name="colorBackgroundActionBar">#FF202020</color>
+    <color name="colorBackgroundActionBar">#FF303030</color>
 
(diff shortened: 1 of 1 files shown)


fm-sys (approved):
Looks good 👍

Lineflyer:
In fact you now changed it back to how it was defined initially before my agreed change by you and @fm-sys  in  #11192.

IMHO its too light again now. Couldn't we use something in between please?

moving-bits:
> In fact you now changed it back to how it was defined initially before my agreed change by you and @fm-sys in #11192.

Oops :-)

> IMHO its too light again now. Couldn't we use something in between please?

How about a middle value as compromise?

image

(this is #FF2A2A2A)

Actionbar is a bit darker, but still (somewhat) distinguishable from regular background. Colors are pretty close, but it may be ok.

If that does not work either, we should discuss to
- either set spinner popup background color to dialog background color
- or change actionbar color to something completely different, non-black / non-grey

Issue #11139, opened 2021-07-07
material design SVG icons


ztNFny:
For compact icons map view the dot_....png files are used. Some of those only exist in drawable-mdpi which causes those icons to look quite bad.

Missing icons:
dot_found.png
dot_waypoint_flag.png
dot_waypoint_pkg.png
dot_waypoint_reference.png
dot_waypoint.png

moving-bits:
Can we change those dot markers to svg? (And would that fit into the mechanism which builds the map markers?)

ztNFny:
I'm using SVGs (for all the map icons) in my custom build, so that works. I'm attaching my icons, they are based on a certain platforms colors palette
[dots.zip](https://github.com/cgeo/cgeo/files/6779722/dots.zip)

moving-bits:
@ztNFny 
They look nice. May we include your icons in c:geo?

ztNFny:
Sure. It's 2 simple circles, nothing more.
Maybe you want to adjust the colors a tiny bit to differentiate from GC colors though? Its their exact colors right now. Not sure about the politics there.
I can also do a PR if you'd like.

moving-bits:
> Sure. It's 2 simple circles, nothing more.

great!

> Maybe you want to adjust the colors a tiny bit to differentiate from GC colors though? Its their exact colors right now. Not sure about the politics there.

we could use the colors from current dot pngs

> I can also do a PR if you'd like.

sure, thanks. Please base it on current `release` so that we can include it in our next beta.

ztNFny:
Will do, but probably after the weekend. Got some caches to find ;-)
I'll do with and without white circle, using the current colors. Then you can pick one. 
I also modded a lot of other icons, but most of them really based on GC ones. Would you guys be open to new marker icons for note, custom coords, found, ... based on material design icons? Well ... that'd probably turn into a whole redesign of map markers campaign, including flat design, then... Wouldn't want to start working on such a thing without some decision this is the way to go because design takes ages... 
Attaching a screenshot of my current mod with GC type icons and material icons for everything else.
![Screenshot_20210709-225607](https://user-images.githubusercontent.com/1258173/125135284-059c0b80-e109-11eb-90a0-53fc7157f520.jpg)

moving-bits:
This looks nice! I'll discuss that in our next dev meeting and come back to you.

Regarding the dot markers: I like the variant with the small white circle around it - makes them more visible especially on a crowded / very detailed map.

[...]

PR #11247, opened 2021-07-16
More type/waypoint icons (in preparation of vector icons)

ztNFny:
Add more icons for cache- and waypoint-types in preparation for the vector icons currently in development (#11139)
Today several types/waypoints share their icons, the in-work vector icons will change that. In preparation of that this is the code change to support more types already today.

For now the original PNGs are being cloned so that to the user nothing changes.


Lineflyer:
ok to test

PR #11252, opened 2021-07-17
fix #10951: merge map quick settings with global filter, create filter context, migrate global settings

### LINKED ISSUE #10951: New Filter: remove Global filters "exclude own/found, disabled. Archived",
Remove the global settings for filtering status (exclude own/found, disabled, archived) and make those a pure map viewer setting (accessible via mao quick settings)

This includes removing the corresponding settings in the Settings

Last chance to protest :-)

eddiemuc:
fix #10951: merge map quick settings with global filter, create filter context, migrate global settings

This PR, if merged, will sync the map quick settings related to cache filtering (OWN, FOUND, DISABLED, ARCHIVED, HAS_OFFLINE_LOG) no longer with the global exclude settings but with the currently active filter:
* states shown are extracted from current filter. (ALL or YES will lead to checked box, NO will lead to unchecked box)
* changes done in quick settings will be merged to current filter (Status filter, if not yet available, will be created; checked box will lead to ALL, unchecked box to "NO". 
* Important: it is not possible to set a status filter to "YES" in map quick settings (two states have to be mapped to three states)! In order to preserve YES status set in normal filter dialog, merge will only be done for really changed checkboxes (marked -> unmarked or unmarked->marked). This also means that an existing YES can only be converted to an ALL by first deselecting, then closing, then reopening, then selecting again a setting in map quick settings (I am pretty sure that an issue will fly in for that later...)
* If filter is "too complex" to merge status lossless (means: it is an AND or a NOT filter), then map quick settings cache filter is disabled with all checkboxes selected:

image

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -66,11 +66,6 @@
     <string translatable="false" name="pref_ratingwanted">ratingwanted</string>
     <string translatable="false" name="pref_friendlogswanted">friendlogswanted</string>
     <string translatable="false" name="pref_opendetailslastpage">opendetailslastpage</string>
-    <string translatable="false" name="pref_excludemine">excludemine</string>
-    <string translatable="false" name="pref_excludefound">excludefound</string>
-    <string translatable="false" name="pref_excludedisabled">excludedisabled</string>
-    <string translatable="false" name="pref_excludearchived">excludearchived</string>
-    <string translatable="false" name="pref_excludeofflinelog">excludeofflinelog</string>
     <string translatable="false" name="pref_global_wp_extraction_disable">globalWpExtractionDisable</string>
     <string translatable="false" name="pref_excludeWpOriginal">excludeWpOriginal</string>
     <string translatable="false" name="pref_excludeWpParking">excludeWpParking</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -506,6 +506,10 @@
     <string name="cache_filtergroup_details">Details</string>
     <string name="cache_filtergroup_userspecific">User specific</string>
 
+    <string name="cache_filter_contexttype_live_title">Live Filter</string>
+    <string name="cache_filter_contexttype_offline_title">Offline Filter</string>
+    <string name="cache_filter_contexttype_transient_title">Filter</string>
+
 
     <string name="cache_filter_include">Include</string>
     <string name="cache_filter_exclude">Exclude</string>
@@ -773,11 +777,6 @@
     <string name="init_select_language">Select language</string>
     <string name="init_summary_select_language">Select in-app language (requires restart)\n</string>
     <string name="init_use_default_language">Use default language</string>
-    <string name="init_category_exclude">Exclude caches</string>
-    <string name="init_exclude_own">Exclude Own</string>
-    <string name="init_summary_exclude_own">Exclude caches you own </string>
-    <string name="init_exclude_found">Exclude Found</string>
-    <string name="init_summary_exclude_found">Exclude caches you have found</string>
     <string name="init_showwaypoints">Show Waypoints</string>
     <string name="init_showwaypoint_description">If less than the given amount of caches are displayed on the map, their waypoints are shown additionally.</string>
     <string name="init_zoomincludingwaypoints">Zoom includes waypoints</string>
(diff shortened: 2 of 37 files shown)


eddiemuc:
Re-set to draft, I continue working on #10951 as part of this PR

Lineflyer:
Thanks for your efforts, sounds very good already.

> Important: it is not possible to set a status filter to "YES" in map quick settings (two states have to be mapped to three states)! In order to preserve YES status set in normal filter dialog, merge will only be done for really changed checkboxes (marked -> unmarked or unmarked->marked). This also means that an existing YES can only be converted to an ALL by first deselecting, then closing, then reopening, then selecting again a setting in map quick settings (I am pretty sure that an issue will fly in for that later...)

This sounds like I have to try it out in practice to understand it :) But I understand the underlying problem though.

> If filter is "too complex" to merge status lossless (means: it is an AND or a NOT filter), then map quick settings cache filter is disabled with all checkboxes selected:

The disabled state is good, maybe there should be a info toast when users still click on it, that says something like: "Setting cannot be changed as your current filter setting is not compatible. Remove the filter first to change these settings".

> Re-set to draft, I continue working on #10951 as part of this PR

You mean the offline vs. live filter?
That would be great.

Are you also plan to include the migration of the global settings inn that part?
IMHO it would be the most important thing: Migrate and remove the old stuff in our settings.

eddiemuc:
> Are you also plan to include the migration of the global settings inn that part?
> IMHO it would be the most important thing: Migrate and remove the old stuff in our settings.

Actually, I wanted to do this next, then I realized I have to include the "LIVE"/"OFFLINE" distinction first (because we said that we migrate to "Live" filter only (not to offline). So I have to do this first. Will do both as part of this PR

eddiemuc:
@Lineflyer 
The "merge" logic between "checkbox" map quicksettings and "three.state" ALL|YES|NO setting in status filter might become another discussion topic, so I try to explain it in the following using "FOUND" as an example (same applies to "OWN" and "HAS_OFFLINE_LOG").

* In map quick settings, only two states can be set: "show" = marked checkbox = show found caches, "hide" = unmarked checkbox = don't show found caches
* In status filter, THREE states can be set: ALL="show both found and non-found, YES="show only found", NO="show only non-found"
* When a user opens the map quick settings and a status filter already exists, , I map "ALL" and "YES" to "show" and "No" to "hide". Easy.
* More problematic is how to map quick settings back to a status filter when quick settings are closed, because I have to map "two" states to "three. a "show" in quick settings can mean "ALL" or "YES", but I have to make some decision here. I currently do it as follows:
    * when user has not changed the checkbox value, I do not change the status filter too (so ALL/YES are both preserved)
    * when user HAS changed the checkbox value, I map "show" to "ALL" and "hide" to "NO". 
* This means that using the map quick settings the user can NOT set "YES" value, and he can also not change the value from "YES" to "ALL" in one go
=> if this is a discussion topic, please open a separate issue for it and make a suggestion on how to handle it differently

Lineflyer:
> This means that using the map quick settings the user can NOT set "YES" value, and he can also not change the value from "YES" to "ALL" in one go
=> if this is a discussion topic, please open a separate issue for it and make a suggestion on how to handle it differently

That sounds totally logically to me:
AFAIU the quick settings checkboxes do in fact represent (compared to the filter):
Checked = All
Unchecked = No

All good here.
If someone objects, please feel free to convert the upper comment into a new issue. But IMHO thats the best method.

eddiemuc:
Ok, I added to this PR:
* introduction of Filter contexts as discussed (LIVE, OFFLINE and TRANSIENT)
* merging global exclude options into live filter

If I have not forgotten anything, this should conclude #10951. But this is a BIG change!

eddiemuc:
Contexts:

image

image

image

eddiemuc:
Warning if filter can't be changed from map quick settings:

image

fm-sys:
> Contexts:
> 
> image
> 
> image
> 
> image

Is "persistent" really understandable for users? I wouldn't know what is meant. What about "temporary"?

eddiemuc:
> Is "transient" really understandable for users? I wouldn't know what is meant. What about "temporary"? 

Hm I guess you are right. Too much used to IT slang I am :-)

What about just the word "Filter"? (Nothing else)

eddiemuc:
Changed "Transient Filter" to "Filter", will merge now

PR #11258, opened 2021-07-18
fix #11208: disallow returning of archived caches from oc

### LINKED ISSUE #11208: Archived caches - difference between connectors
I just now realized, that different services treat archived caches in a different way, which leads to a difference between older release and the current beta:

For gc.com all interfaces (map, search) do never return archived caches. Those are - very understandable - considered as no longer relevant but have more an archive character.

This seems to be different on OC (at least on OC.de). Here I suddenly see archived caches in my search results (but not on map).
With the old release this was not the case, although I did never use "Hide archived caches" option.

This seems to have changed as a side effect of the recent changes.
Is this intentional and does it make sense...I know they can be filtered anyway, so this is not the reason for this issue.
The reason is, that we now have different content of search results between different connectors.

Edit:
You could also say: If we let OC return archived caches then we should also do it on the map but not only in searches.
Or (my suggestion): Do not request archived caches in online searches...this would only encourage people to search archived caches (or less advanced users to search for them only later realizing, that they are archived).

eddiemuc:
fix #11208: disallow returning of archived caches from oc

### DIFF
--- a/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -345,9 +345,11 @@ final class OkapiClient {
                 if (!statusFilter.isExcludeDisabled()) {
                     value += "|Temporarily unavailable";
                 }
-                if (!statusFilter.isExcludeArchived()) {
-                    value += "|Archived";
-                }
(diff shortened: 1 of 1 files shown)


Lineflyer:
Thank you. I think this makes it more consistent after all.

PR #11259, opened 2021-07-18
fix #11246: add gcvote request to gc online search results

### LINKED ISSUE #11246: GCVote seems missing in our new search?
image

On the screenshots, it seems like the GCVote is missing on our new UI? Or was there a difference in the way the screenshots where created?

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/discussions/11069#discussioncomment-995496_

eddiemuc:
fix #11246: add gcvote request to gc online search results

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -148,7 +148,7 @@ public class GCMap {
             fillForBasicFilter(baseFilter, search);
         }
 
-        return GCWebAPI.searchCaches(search);
+        return GCWebAPI.searchCaches(search, true);
     }
 
     private static void fillForBasicFilter(@NonNull final BaseGeocacheFilter basicFilter, final GCWebAPI.WebApiSearch search) {
--- a/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
@@ -6,6 +6,7 @@ import cgeo.geocaching.enumerations.CacheAttribute;
 import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.StatusCode;
+import cgeo.geocaching.gcvote.GCVote;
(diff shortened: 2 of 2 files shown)


main/src/cgeo/geocaching/connector/gc/GCMap.java
```diff
@@ -148,7 +148,7 @@ public static SearchResult searchByFilter(@NonNull final GeocacheFilter filter,
             fillForBasicFilter(baseFilter, search);
         }
 
-        return GCWebAPI.searchCaches(search);
+        return GCWebAPI.searchCaches(search, true);
```
moving-bits, line 151: shouldn't including GCVotes depend on the system setting for GCVote connector being activated?

Lineflyer: Actually it was always like that:
In the GCVote settings the default is to load the ratings unless the user explicitly switches it off.
Autorizing to post votes is independant from that.

However that mentioned setting for loading votes should be respected.

moving-bits: Yes, that's what I meant. If someone switches GCVote setting of, no http(s) call to GCVote should be made.

eddiemuc: The switch in the settings is respected further down the call stack in the GCVote class itself

main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
```diff
@@ -811,6 +812,9 @@ static SearchResult searchCaches(final WebApiSearch search) {
         }
 
         result.addAndPutInCache(foundCaches);
+        if (includeGcVote) {
+            GCVote.loadRatings(foundCaches);
```
eddiemuc, line 816: `GCVote.loadRatings` checks for the Settings (whether to load GCVote or not).before issuing any requests to the server.

main/src/cgeo/geocaching/connector/gc/GCMap.java
moving-bits: ah, thanks for clarifying 👍

Issue #8733, opened 2020-08-02
Wrong saved date in popup on map


Lineflyer:
I noticed this before and we even discussed this briefly in some older issue. AFAIR it is a side effect of merging of data retrieved from the live map to caches stored on the device and was introduced while implementing the new live map parsing:

- Have a cache stored e.g. 20 days ago
- Open the live map and scroll to the position of this cache icon
- Click to open the popup

Result:
- The popup text might state (depending on ?): "Saved a few minutes ago"
- Open details of this cache from popup
- The details show "Saved on the device 20 days ago"

Expected result:
- Live map parsing should not alter the save date shown in popup and also show "Saved on the device 20 days ago"

Obviously the save date shown in the popup does not overwrite the existing date information (as later when you open the cache it is stil the old dat) but somehow a transient value is shown here.

Tested with 2020.07.02 and also current master

moving-bits:
Cache info of the stored cache is somewhat "hidden" by live map info - stored date is just one example for this. Hint / personal note is another example (which is the reason why with activated live map the "hint" button is not shown any longer even if there is hint/personal note data existing in the stored cache). And there may be more.

@jonas-koeritz 
You have been working in the live map area before. Do you have time to look into this?

jonas-koeritz:
This really might be a problem in my code. AFAIK there is only an "updated" field, not a separate field for the time the cache has been stored on the device. The updated field gets set by my code IIRC, this would be wrong in that case.

jonas-koeritz:
> This really might be a problem in my code. AFAIK there is only an "updated" field, not a separate field for the time the cache has been stored on the device. The updated field gets set by my code IIRC, this would be wrong in that case.

I've checked and it doesn't seem to be the case...I'm not setting the updated field :/

The Cache-Popup implementation is really convoluted and very abstract in my opinion. As it "only" serves a single purpose, maybe thes popup could be reworked to be more concise?

eddiemuc:
As part of #11265 I kind of stumbled into this older issue.

I looked into code, and I think that the root culprit here is the ambivalent usage of the "updated" field of a geoache, which contains the timestamp where a cache was last updated. But this field is used in two different ways/interpretations:
* On one hand, it is used to determine when a cache was last stored. That's the main use case
* On the other hand, this field is also updated to current timestamp when for a stored cache new information is loaded from the service (specifically: when cache data is loaded into the Live map). This "update" is however not stored back to the DB later (and also not the maybe new information loaded from service)

[...]

PR #11265, opened 2021-07-19
fix #11196: make storedsince range work for non-offline caches

### LINKED ISSUE #11196: filter "saved since" to '-' does not work
## Bug description 
The Filter "saved since" with value "-" has no result.

- Ensure in map that there are unsaved caches manually or by filter "Status" - "saved"
- open filter "saved since"
- set left and right value to '-"
- result: no caches in map are shown

as the hint says 'not saved' I expect to see not saved caches from the map

## c:geo version 
2021.07.10-RC

## Reproducible 
Yes

eddiemuc:
fix #11196: make storedsince range work for non-offline caches

also fixes the strange  "a few minutes ago" message on cache detail popup for caches stored much longer time before.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -2535,7 +2535,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         offlineRefresh.setOnClickListener(refreshCacheClickListener);
 
         if (cache.isOffline()) {
-            offlineText.setText(Formatter.formatStoredAgo(cache.getUpdated()));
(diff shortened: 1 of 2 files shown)


Lineflyer:
Fix #8733

eddiemuc:
After Thinking a bit, let me put this to draft for now. I want to make sure it is really valid to use detailedupdate instead of update in all cases before merge

eddiemuc:
I checked in code, and the connectors seem to set "detailedupdate" field every time a cache is refreshed. So I guess "detailedupdate" field is the more reliable field when it comes to estimate when a cache was last stored.

I will therefore merge this PR.

Issue #11257, opened 2021-07-17
Geocode input in LogTrackable Activity uses wrong style


fm-sys:
Geocode input in `LogTrackableActivity` doesn't use correct style:

![grafik](https://user-images.githubusercontent.com/64581222/126049453-691f8c1b-e21b-4bfb-871d-0e951e516c7e.png)
![grafik](https://user-images.githubusercontent.com/64581222/126049974-7f8873da-0d9c-4302-b102-e91acdad0953.png)

Lineflyer:
I tried to look into it, but it seems a bit to complex for me.
In fact the styling is applied in the corresponding XML, but I guess the class `AutoCompleteTextView` needs to be changed to also inherit/use that style?!

Besides that, the margin defined for the geocode text field should be removed to align it with the other text fields.

I am afraid, someone else needs to dive deeper into that. Still learning...

fm-sys:
In fact, it is working with other AutoCompleteTextViews (see e.g. `SearchActivity`). I also tried finding a quick solution before opening this issue, but I did not had luck either... Very strange!

moving-bits:
How do you reach this activity with "geocode" field visible?

fm-sys:
I have reached it via entering a GeoKrety GKxxxx code...

Lineflyer:
and setting log type to e.g. "Visited"

moving-bits:
Thanks for pointing me in the right direction. Problem solved, PR is created...

PR #11268, opened 2021-07-19
Fix #11267 - Wrong text field hint alignment

### LINKED ISSUE #11267: Text field hint not properly aligned
At some place I do see this for the text field description as soon as I fill in values.
image

It seems its caused by manually defined margins in affected views.
I will give it a try to fix it myself.

Lineflyer:
This also touches the broken geocode field in `logtrackable_activity.xml` as there is kind of the same problem. It does however not solve #11257


Lineflyer:
image

moving-bits (approved):
LGTM

BTW: We should get rid of more of those individual formattings and use styles instead...

Issue #11191, opened 2021-07-12
list limit - change max value in settings


bekuno:
The list limit has currently the maximum amount of 5000.
This should be a higher value, I would suggest to set it at least to 10000.

moving-bits:
Thinking of lists like "history" or "all" we should set the limit even higher...

MagpieFourtyTwo:
Over here "Unlimited" is working perfectly fine since the change to SQL-based filtering:
Not even a single crash (at least not reproducibly "long list related" ;) ) and stunningly fast (4 secs to display/filter "All").
And as I have about 20k caches stored offline, I would consider at least 20k (if not even more) as save ...
Thus a max value of even way more than 20k (50k?) would make sense, IMO
(e. g. to be able to test for the limits, if "Unlimited" gives problems somewhere).

eddiemuc:
@MagpieFourtyTwo I am glad it works so well. I am also a bit surprised because the change to SQL-based filtering did of course not reduce the amount of caches loaded into memory or the number of SQLs executed in background for it (and I always assumed this was the source of the slowness...). So I wonder where the slowness came from :-) Maybe from in-memory-sorting, but that is very far-fetched.

To the initial issue: the max value for "max limit" should not be measured against the max. list sizes our users use but against **the max. size of items on a list which the user would really read manually** (without adding an additional filter or sort because it is too much for him to read through). The whole feature should limit loading too much into memory and sparing the DB many unnecessary SQL statements. 

I think 10K would be ok, maybe even 20K, but for higher values the feature does not make sense any more anyway. If users really want to deal with unlimited 20K+ lists and load them into memory at a single tap, then they are allowed to do so (by deactivating max limit).

fm-sys:
> Maybe from in-memory-sorting, but that is very far-fetched.

Could likely be. Sorting is a quite expensive task. Depending on the sorting algorithm we use the sorting time can be much faster if the list is already presorted...

eddiemuc:
> > Maybe from in-memory-sorting, but that is very far-fetched.
> 
> Could likely be. Sorting is a quite expensive task. Depending on the sorting algorithm we use the sorting time can be much faster if the list is already presorted... 

Yes, but we're talking about 10+ seconds for lists of maybe 50K items... This is too much to be explained by sorting Algo imho. I believe Java SDK uses an adapted version of mergesort

MagpieFourtyTwo:
@eddiemuc
[...]

PR #11277, opened 2021-07-20
Set max limit for list to 50k for testing purpose (rel #11191)

murggel:
## Description

Set the load limit for lists from 10000 to 50000 for testing purpose. Should be probabaly adapted before going into production.

## Related issues

#rel 11191

### DIFF
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -921,7 +921,7 @@
         <cgeo.geocaching.settings.SeekbarPreference
             android:key="@string/pref_list_initial_load_limit"
             app:min="0"
-            app:max="5000"
+            app:max="50000"
             android:defaultValue="1000" />
 
         <PreferenceCategory android:title="@string/settings_title_gpx"
--- a/main/res/xml/preferences.xml
+++ b/main/res/xml/preferences.xml
@@ -921,8 +921,8 @@
         <cgeo.geocaching.settings.SeekbarPreference
             android:key="@string/pref_list_initial_load_limit"
             app:min="0"
-            app:max="50000"
-            android:defaultValue="1000" />
+            app:max="@integer/list_load_limit_max"
+            android:defaultValue="@integer/list_load_limit_default" />
 
         <PreferenceCategory android:title="@string/settings_title_gpx"
             android:layout="@layout/preference_category" >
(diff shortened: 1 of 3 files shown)


bekuno:
please target it to release.

moving-bits:
> please target it to release.

done

eddiemuc:
AFAIR we wanted to set the default value to 0 too (thus disabling list limitation by default). Is this the case?

Lineflyer:
> AFAIR we wanted to set the default value to 0 too (thus disabling list limitation by default). Is this the case?

Wasn't this done already? I remember it was done.

murggel:
> > AFAIR we wanted to set the default value to 0 too (thus disabling list limitation by default). Is this the case?
> 
> Wasn't this done already? I remember it was done.

Was not done in master nor release ... there it's still on 5000. Should I set it to 0 then?

main/res/xml/preferences.xml
```diff
@@ -921,7 +921,7 @@
         <cgeo.geocaching.settings.SeekbarPreference
             android:key="@string/pref_list_initial_load_limit"
             app:min="0"
-            app:max="5000"
+            app:max="50000"
             android:defaultValue="1000" />
```
fm-sys, line 925: Interesting though: In the settings the default is already set to 0.

https://github.com/cgeo/cgeo/blob/51162ae0f986506b6806d4abc068b5ed3d248e9a/main/src/cgeo/geocaching/settings/Settings.java#L1919

However, a number resource should be used for the default value, so that always the same value is used at both places...

murggel: Will add that

eddiemuc:
Looks good, thanks for the effort! Let's merge

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -1895,7 +1895,7 @@ public static void setCacheFilterConfig(final String config) {
     }
 
     public static int getListInitialLoadLimit() {
-        return getInt(R.string.pref_list_initial_load_limit, 0);
+        return getInt(R.string.pref_list_initial_load_limit, R.integer.list_load_limit_default);
```
fm-sys, line 1898: R.integer.* is just a unique ID.

It needs to be wrapped with `getKeyInt(R.integer.*)`

murggel: Ok.
So I create another PR because this one is already merged.

fm-sys: I just quickly created PR  #11289 for it myself, so that the issue doesn't get lost...

PR #11287, opened 2021-07-21
ALC no longer working (fix: #11001)

baiti:
fixes broken ALC functionality

### DIFF
--- a/main/templates/keys.xml
+++ b/main/templates/keys.xml
@@ -30,4 +30,7 @@
     <!-- Geocaching.su -->
     <string name="su_consumer_key" translatable="false">@su.consumer.key@</string>
     <string name="su_consumer_secret" translatable="false">@su.consumer.secret@</string>
+
+    <!-- Geocaching.com public ALC API -->
+    <string name="alc_consumer_key" translatable="false">@alc.consumer.key@</string>
 </resources>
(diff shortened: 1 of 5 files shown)


main/templates/keys.xml
```diff
@@ -30,4 +30,7 @@
     <!-- Geocaching.su -->
     <string name="su_consumer_key" translatable="false">@su.consumer.key@</string>
     <string name="su_consumer_secret" translatable="false">@su.consumer.secret@</string>
+
+    <!-- Geocaching.com public ALC API -->
+    <string name="alc_consumer_key" translatable="false">@alc.consumer.key@</string>
```
bekuno, line 35: As documented in https://github.com/cgeo/cgeo#api-keys-installation the key management has two parts - in the `main/template/keys.xml` you define the part for using the key in the code. And in `templates/private.properties` you define the point where the user should put his keys. 

The workflow described in the Readme works.

@baiti Please add the missing entry in `templates/private.properties` and maybe also a short text in the Readme.

baiti: > As documented in https://github.com/cgeo/cgeo#api-keys-installation the key management has two parts - in the `main/template/keys.xml` you define the part for using the key in the code. And in `templates/private.properties` you define the point where the user should put his keys.
> 
> The workflow described in the Readme works.

Yes, it works, but it confused me. `private.properties` we take from `./templates/` and `keys.xml` we update in `./main/templates/`. 
> 
> @baiti Please add the missing entry in `templates/private.properties` and maybe also a short text in the Readme.

first part is done and commited. Not sure about what to write into the READ.me.

moving-bits:
Setting the "do not merge" label for now, as we have to inject the key in our CI first

Lineflyer:
Preconditions for this PR are now met and a local test on my machine was succesful.
Merging now.

The CI still needs to be prepared for our productive/beta/nightly releases. I informed @mucek4 accordingly.

PR #11290, opened 2021-07-21
fix: wrap resource-id to integer for list-limit-default (rel #11191)

murggel:
## Description

wrap resource-id to integer for list-limit-default. 
Was 

## Related issues

rel #11191

### DIFF
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -1916,7 +1916,7 @@ public class Settings {
     }
 
     public static int getListInitialLoadLimit() {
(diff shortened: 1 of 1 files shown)


Lineflyer:
looks same as PR #11289

murggel:
> 
> 
> looks same as PR #11289

you are right, did not know, that @fm-sys was working on that too.

fm-sys:
Yes, Sorry! I was thinking whether I should create a follow-up issue for it so that it doesn't get lost, but decided that a quick PR would be easier. I was hoping that you will see my comment before starting with it as well... 

But as I've created the PR on my mobile, you're work wasn't completely useless as I guess that you have tested the change (unlike me 😅)

PR #11294, opened 2021-07-22
"remove list" optimizations

moving-bits:
## Description
- restore "remove list" dialog (reverts #11182)
- add option "do not ask me again" on a per-list basis
- add option to reset this flag for a list

|new dialog option|option to reset flag|option is disabled if flag not set|
|---|---|---|
|image|image|image|

### DIFF
--- a/main/res/menu/cache_list_options.xml
+++ b/main/res/menu/cache_list_options.xml
@@ -128,6 +128,12 @@
                 android:title="@string/caches_set_listmarker"
                 app:showAsAction="never|withText">
             </item>
+            <item
+                android:id="@+id/menu_set_askfordeletion"
+                android:title="@string/caches_askdeletion"
+                app:showAsAction="never"
+                android:visible="false">
+            </item>
         </menu>
     </item>
     <item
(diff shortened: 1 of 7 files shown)


fm-sys:
Can we change the checkbox string to "don't ask again for this list" to make clear this is a per-list setting?

moving-bits:
> Can we change the checkbox string to "don't ask again for this list" to make clear this is a per-list setting?

Sure. I had reused an existing string, but I agree a more specific string will be more precise.

bekuno:
If I set "do not ask again" and hit "yes" - how it will work later?
If I create a list with the same name (case insensitive?) - is then this setting in use for this list?

moving-bits:
> If I set "do not ask again" and hit "yes" - how it will work later?
> If I create a list with the same name (case insensitive?) - is then this setting in use for this list?

Setting is stored list-specific in the `cg_lists` table. If you remove a list (and not undo this), every info attached to the list is removed as well (name, list marker, "do not ask" setting etc.). 
If you create a new list (or recreate a list using the same name), the "do not ask me" setting defaults to `false` (or "0", to be more precise)

bekuno:
Ok. What about clarifying the string to "When 'No', do not ask me again for this list"?

fm-sys:
> Ok. What about clarifying the string to "When 'No', do not ask me again for this list"?

I think this would just make things harder to understand. We could rather just grey out the "yes" button if the checkbox is checked (like we already do it for one-time messages so this would only be consistent)

moving-bits:
I've added checkbox handling as described above:

image

PR #11296, opened 2021-07-22
add compact D/T filter (fix #11180)

### LINKED ISSUE #11180: compact D/T for Filters
To save some space in the filters I propose to compact this.

Original:
Screenshot_20210711-223641_cgeo-2

Proposal:
Screenshot_20210711-223641_cgeo optimized-2

murggel:
## Description

add new compact filter for difficulty and terrain.
image

Scale for range is only shown once.
image

I created a new GeocacheFilter and FilterViewHolder, which uses the existing ones for Difficulty and Terrain internally. 

## Related issues

fix #11180

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
@@ -1,0 +1,64 @@
+package cgeo.geocaching.filters.core;
+
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.storage.SqlBuilder;
+import cgeo.geocaching.utils.expressions.ExpressionConfig;
+
+public class DifficultyAndTerrainGeocacheFilter extends BaseGeocacheFilter {
+
+    public DifficultyGeocacheFilter difficultyGeocacheFilter = new DifficultyGeocacheFilter();
+    public TerrainGeocacheFilter terrainGeocacheFilter = new TerrainGeocacheFilter();
+
+    @Override
+    public Boolean filter(final Geocache cache) {
+        if (cache == null) {
+            return null;
+        }
+
+        final Boolean diffRange = difficultyGeocacheFilter.filter(cache);
+        final Boolean terrainRange = terrainGeocacheFilter.filter(cache);
+        if (diffRange == null) {
+            return terrainRange;
+        }
+        if (terrainRange == null) {
+            return diffRange;
+        }
+
+        return diffRange && terrainRange;
+    }
+
+    @Override
+    public boolean isFiltering() {
+        return difficultyGeocacheFilter.isFiltering() || terrainGeocacheFilter.isFiltering();
+    }
+
+    @Override
+    public void addToSql(final SqlBuilder sqlBuilder) {
+        difficultyGeocacheFilter.addToSql(sqlBuilder);
+        terrainGeocacheFilter.addToSql(sqlBuilder);
+    }
+
+    @Override
+    public void setConfig(final ExpressionConfig config) {
+        final ExpressionConfig diffConfig = new ExpressionConfig();
+        diffConfig.putDefaultList(config.get("difficulty"));
+        difficultyGeocacheFilter.setConfig(diffConfig);
+
+        final ExpressionConfig terrainConfig = new ExpressionConfig();
+        terrainConfig.putDefaultList(config.get("terrain"));
+        terrainGeocacheFilter.setConfig(terrainConfig);
+    }
+
+    @Override
+    public ExpressionConfig getConfig() {
+        final ExpressionConfig config = new ExpressionConfig();
+        config.put("difficulty", difficultyGeocacheFilter.getConfig().getDefaultList());
+        config.put("terrain", terrainGeocacheFilter.getConfig().getDefaultList());
+        return config;
+    }
+
+    @Override
+    protected String getUserDisplayableConfig() {
+        return difficultyGeocacheFilter.getUserDisplayableConfig() + " / " + terrainGeocacheFilter.getUserDisplayableConfig();
+    }
+}
--- a/main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
+++ b/main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
@@ -18,6 +18,7 @@ public enum GeocacheFilterType {
     PERSONAL_NOTE("note", R.string.cache_filter_personalnote, R.string.cache_filtergroup_userspecific, PersonalNoteGeocacheFilter::new),
     DIFFICULTY("difficulty", R.string.cache_filter_difficulty, R.string.cache_filtergroup_details,  DifficultyGeocacheFilter::new),
     TERRAIN("terrain", R.string.cache_filter_terrain, R.string.cache_filtergroup_details, TerrainGeocacheFilter::new),
+    DIFFICULTY_TERRAIN("difficulty_terrain", R.string.cache_filter_difficulty_terrain, R.string.cache_filtergroup_details, DifficultyAndTerrainGeocacheFilter::new),
     STATUS("status", R.string.cache_filter_status, R.string.cache_filtergroup_basic, StatusGeocacheFilter::new),
     ATTRIBUTES("attributes", R.string.cache_filter_attributes, R.string.cache_filtergroup_details, AttributesGeocacheFilter::new),
     OFFLINE_LOG("offlinelog", R.string.cache_filter_offlinelog, R.string.cache_filtergroup_userspecific, OfflineLogGeocacheFilter::new),
--- a/main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
+++ b/main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
@@ -1,0 +1,67 @@
+package cgeo.geocaching.filters.gui;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.filters.core.DifficultyAndTerrainGeocacheFilter;
+import cgeo.geocaching.filters.core.DifficultyGeocacheFilter;
+import cgeo.geocaching.filters.core.GeocacheFilterType;
+import cgeo.geocaching.filters.core.TerrainGeocacheFilter;
+import cgeo.geocaching.ui.TextParam;
+import cgeo.geocaching.ui.ViewUtils;
+
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+
+import androidx.annotation.NonNull;
+
+public class DifficultyAndTerrainFilterViewHolder extends BaseFilterViewHolder<DifficultyAndTerrainGeocacheFilter> {
+
+    private IFilterViewHolder diffView = null;
+    private IFilterViewHolder terrainView = null;
+
+    @Override
+    public View createView() {
+        final LinearLayout ll = new LinearLayout(getActivity());
+        ll.setOrientation(LinearLayout.VERTICAL);
+
+        diffView = createItemRangeSelectorLayout(GeocacheFilterType.DIFFICULTY, TextParam.text("D"), ll);
+        terrainView = createItemRangeSelectorLayout(GeocacheFilterType.TERRAIN, TextParam.text("T"), ll);
+        ((ItemRangeSelectorViewHolder) terrainView).removeScaleLegend();
+
+        return ll;
+    }
+
+    @Override
+    public void setViewFromFilter(@NonNull final DifficultyAndTerrainGeocacheFilter filter) {
+        if (diffView != null) {
+            diffView.setViewFromFilter(filter.difficultyGeocacheFilter);
+        }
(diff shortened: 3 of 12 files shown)


main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
```diff
@@ -0,0 +1,67 @@
...
+
+        diffView = createItemRangeSelectorLayout(GeocacheFilterType.DIFFICULTY, TextParam.text("D"), ll);
+        terrainView = createItemRangeSelectorLayout(GeocacheFilterType.TERRAIN, TextParam.text("T"), ll);
+        ((ItemRangeSelectorViewHolder) terrainView).removeScaleLegend();
+
+        return ll;
+    }
+
+    @Override
+    public void setViewFromFilter(@NonNull final DifficultyAndTerrainGeocacheFilter filter) {
```
murggel, line 35: DifficultyAndTerrainFilterViewHolder: 
warnings / remarks in AndroidStudio are shown for unchecked templates...
image

 I don't know how to get rid of it

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
```diff
@@ -0,0 +1,64 @@
...
+        terrainConfig.putDefaultList(config.get("terrain"));
+        terrainGeocacheFilter.setConfig(terrainConfig);
+    }
+
+    @Override
+    public ExpressionConfig getConfig() {
+        final ExpressionConfig config = new ExpressionConfig();
+        config.put("difficulty", difficultyGeocacheFilter.getConfig().getDefaultList());
+        config.put("terrain", terrainGeocacheFilter.getConfig().getDefaultList());
+        return config;
```
murggel, line 57: Don't know, if that is a good way to handle the config, probably there is a better / different way?

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
```diff
@@ -0,0 +1,64 @@
...
+        difficultyGeocacheFilter.addToSql(sqlBuilder);
+        terrainGeocacheFilter.addToSql(sqlBuilder);
+    }
+
+    @Override
+    public void setConfig(final ExpressionConfig config) {
+        final ExpressionConfig diffConfig = new ExpressionConfig();
+        diffConfig.putDefaultList(config.get("difficulty"));
+        difficultyGeocacheFilter.setConfig(diffConfig);
+
```
murggel, line 46: Don't know, if that is a good way to handle the config, probably there is a better / different way?

murggel:
ready for a first review - I added some comments, where I need feedback

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
```diff
@@ -0,0 +1,64 @@
+package cgeo.geocaching.filters.core;
+
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.storage.SqlBuilder;
+import cgeo.geocaching.utils.expressions.ExpressionConfig;
+
+public class DifficultyAndTerrainGeocacheFilter extends BaseGeocacheFilter {
```
eddiemuc, line 7: It is an interesting idea that your  new filter reuses the two existing ones. I rather thought about creating a new filter andthen delete the two other ones (as they are no longer necessary)

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
eddiemuc: Given that you reuse existing filters, this is a fine concept. Maybe a shortcut method can be added to "ExpressionConfig" to support this usage pattern - I never thought about the use case of a filter reusing other filters before :-))

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
eddiemuc: same here

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
```diff
@@ -0,0 +1,67 @@
...
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+
+import androidx.annotation.NonNull;
+
+public class DifficultyAndTerrainFilterViewHolder extends BaseFilterViewHolder<DifficultyAndTerrainGeocacheFilter> {
+
+    private IFilterViewHolder diffView = null;
+    private IFilterViewHolder terrainView = null;
```
eddiemuc, line 20: making these two variables of type "IFilterViewHolder<?>" should make the template warning below go away. I hope...

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
eddiemuc: See my comment in line 20. I hope this helps. But if nothing helps, you can also add a `@SuppressWarnings("rawtypes")` to the method

main/src/cgeo/geocaching/filters/gui/FilterViewHolderCreator.java
```diff
@@ -77,6 +77,9 @@ private FilterViewHolderCreator() {
             case TERRAIN:
```
eddiemuc, line 77: I think that "TERRAIN" and "DIFFICULTY" should be removed from the type list.

main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
```diff
@@ -18,6 +18,7 @@
     PERSONAL_NOTE("note", R.string.cache_filter_personalnote, R.string.cache_filtergroup_userspecific, PersonalNoteGeocacheFilter::new),
     DIFFICULTY("difficulty", R.string.cache_filter_difficulty, R.string.cache_filtergroup_details,  DifficultyGeocacheFilter::new),
     TERRAIN("terrain", R.string.cache_filter_terrain, R.string.cache_filtergroup_details, TerrainGeocacheFilter::new),
```
eddiemuc, line 20: I think that TERRAIN and DIFFICULTY should be removed from the type list

eddiemuc (approved):
Looks good already.

One thing that is left to do is to replace the "Base" filter configuration in "GeocacheFilterActivity" so it will use the new Difficulty-terrain-filter instead of the two "legacy" filters.

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
murggel: >  I never thought about the use case of a filter reusing other filters before :-))

It just came into my mind - and I thought just give it a try  :innocent: . If it does not work, I can still can create a independent one

murggel: > Maybe a shortcut method can be added to "ExpressionConfig" to support this usage pattern 

Hm, do you mean a static method to create a new config from an existing one? Passing  the key and storing in default-list resp. other way round?

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
murggel: It just came into my mind - and I thought just give it a try 😇 . If it does not work, I can still can create a independent one

main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
murggel: I think I need the types for creating the views.

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
murggel: Yes, but unfortunately it will then gives an error in setViewFromFilter
image

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
murggel: What does that warning mean? In what cases can that lead to an failure?

main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
murggel: Removed from selection list with commit 90a149e

main/src/cgeo/geocaching/filters/gui/FilterViewHolderCreator.java
murggel: see comment above

murggel:
@eddiemuc  Are there any test for filtering via the api yet? Should I try to add some? I recognized I missed some adaptions in `fillForBasicFilter` -> see commit 820e966

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
murggel: As remainder to myself: convert the string to a static member

eddiemuc:
> @eddiemuc  Are there any test for filtering via the api yet? Should I try to add some? I recognized I missed some adaptions in `fillForBasicFilter` -> see commit 820e966 They could be tested via integration-tests...

Sadly and shamefully there are no tests yet. This is a big fat point one todo list. There was simply too much "feature pressure" on the topic until now. 

But at least I created issues for this, and I **will** clean up these things (also removing large parts of code which are no longer used and refactoring ) once the release pressure is gone.

It is simply a very big topic, changing larger parts of the "inner structure" of Cgeo...

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
eddiemuc: Java allows for the usage of generic classes without specifying a class parameter (in this case it will assumed to be Object). This is called a "raw type". This is what you do with diffView and terrainView (you did not specify a class type using <>).

The method you call here however expects a parameter with a specified class type and uses this type. This can commonly lead to runtime exceptions because the raw class type may not be of the same class the method expects. But the compiler can no longer verify if this is the case or whether the code handles this correctly. This is why it issues this warning.

Im general, the usage of raw generic types should be avoided.

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
eddiemuc: What if you change the type of "diffView" to "IFilterViewHolder<DifficultyGeocacheFilter>"?

I know, generics are tricky...

main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
eddiemuc: Something like this:

in ExpressionConfig

```
/** creates a new config object, taking the content of one key of this config and putting it into the default list */
public void ExpressionsConfig getSubconfig(final String key) {
     final ExpressionConfig subConfig = new ExpressionConfig();
        subConfig.putDefaultList(this.get(key));
        return subConfig;
}
```

murggel: Good idea

main/src/cgeo/geocaching/filters/gui/DifficultyAndTerrainFilterViewHolder.java
murggel: Then I get there the warning of raw use ...

eddiemuc: I assume you want to get this done for tonights beta? Then maybe ignore the raw use warning for now, I can take a closer look once I have the code to look into (in release)

murggel: Yes, that was my thought as well, I still have some changes, but my computer is currently not really working, and we have an appointment now.
So probably the remaining stuff is just nice to have and I can clean it later.

So if you think the current state is ok, please feel free to merge, and I try in a few hours again...

If I need to convert from draft, probably that will work from my phone

murggel: Ah, I think removing the single filter is still local on my machine...

eddiemuc:
merge to include in beta

PR #11297, opened 2021-07-22
open filter-selection with long click (fix #11202)

### LINKED ISSUE #11202: Add long click to select filter in cacheListActivity and Map

Changing the current filter to an already stored is in the advanced filter-options. I would like to have an easy way to switch filter.

## Suggested solution

Long click on current filter text and filter-action in action-bar should open a list with the available filter to "quick"-select a different filter.

## Alternatives/workarounds

Move manage and save methods to basic filter view. As discussed we can postpone / leave that out and see, if long-click is sufficient.

murggel:
## Description

open filter-selection with long click on menu-item or filter-bar 
image

option to clear the current filter
image

## open points
* refresh of the map is currently (sometimes?) not working correct

## Related issues

fix #11202

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -554,7 +554,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
 
         initAdapter();
 
-        FilterUtils.connectFilterBar(this);
+        FilterUtils.initializeFilterBar(this, this);
         updateFilterBar();
 
         if (type.canSwitch) {
@@ -728,7 +728,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         });
 
         ListNavigationSelectionActionProvider.initialize(menu.findItem(R.id.menu_cache_list_app_provider), app -> app.invoke(CacheListAppUtils.filterCoords(cacheList), CacheListActivity.this, getFilteredSearch()));
-        FilterUtils.connectFilterMenu(this);
+        FilterUtils.initializeFilterMenu(this, this);
 
         return true;
     }
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -86,6 +86,7 @@ import cgeo.geocaching.utils.AngleUtils;
 import cgeo.geocaching.utils.CalendarUtils;
 import cgeo.geocaching.utils.DisposableHandler;
 import cgeo.geocaching.utils.EmojiUtils;
+import cgeo.geocaching.utils.FilterUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.ShareUtils;
@@ -103,7 +104,6 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Message;
-import android.text.TextUtils;
 import android.view.ContextMenu;
 import android.view.ContextMenu.ContextMenuInfo;
 import android.view.KeyEvent;
@@ -554,7 +554,8 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
 
         initAdapter();
 
-        prepareFilterBar();
+        FilterUtils.connectFilterBar(this);
+        updateFilterBar();
 
         if (type.canSwitch) {
             initActionBarSpinner();
@@ -663,7 +664,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         final LastPositionHelper lastPosition = new LastPositionHelper(this);
 
         applyAdapterFilter();
-        prepareFilterBar();
+        updateFilterBar();
 
         // resume location access
         PermissionHandler.executeIfLocationPermissionGranted(this, new RestartLocationPermissionGrantedCallback(PermissionRequestContext.CacheListActivity) {
@@ -727,6 +728,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         });
 
         ListNavigationSelectionActionProvider.initialize(menu.findItem(R.id.menu_cache_list_app_provider), app -> app.invoke(CacheListAppUtils.filterCoords(cacheList), CacheListActivity.this, getFilteredSearch()));
+        FilterUtils.connectFilterMenu(this);
 
         return true;
     }
@@ -763,7 +765,6 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         try {
             // toplevel menu items
             setEnabled(menu, R.id.menu_show_on_map, !isEmpty);
-            //setEnabled(menu, R.id.menu_filter_legacy, search != null && search.getCount() > 0); //remove for upcoming beta
             setVisibleEnabled(menu, R.id.menu_sort, !isHistory, !isEmpty);
             if (adapter.isSelectMode()) {
                 menu.findItem(R.id.menu_switch_select_mode).setTitle(res.getString(R.string.caches_select_mode_exit))
@@ -927,8 +928,6 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         } else if (menuItem == R.id.menu_invert_selection) {
             adapter.invertSelection();
             invalidateOptionsMenuCompatible();
-//        } else if (menuItem == R.id.menu_filter_legacy) { //remove for upcoming beta
-//            showLegacyFilterMenu(null);
         } else if (menuItem == R.id.menu_filter) {
             showFilterMenu(null);
         } else if (menuItem == R.id.menu_import_web) {
@@ -1039,7 +1038,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         if (view != null && Settings.getCacheType() != CacheType.ALL) {
             Dialogs.selectGlobalTypeFilter(this, cacheType -> {
                 refreshCurrentList();
-                prepareFilterBar();
+                updateFilterBar();
             });
         } else {
             FilterActivity.selectFilter(this);
@@ -1052,6 +1051,14 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
     @Override
     public void showFilterMenu(final View view) {
         GeocacheFilterActivity.selectFilter(this, currentCacheFilter, adapter.getFilteredList(), !resultIsOfflineAndLimited());
+    }
+
+    /**
+     * called from the filter bar view
+     */
+    @Override
+    public boolean showFilterList(final View view) {
+        return FilterUtils.openFilterList(this, currentCacheFilter);
     }
 
     private void setComparator(final CacheComparator comparator) {
@@ -1226,7 +1233,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
 
     private void setFilter() {
         applyAdapterFilter();
-        prepareFilterBar();
+        updateFilterBar();
         updateTitle();
         invalidateOptionsMenuCompatible();
     }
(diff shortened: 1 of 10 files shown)


main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -1347,12 +1352,26 @@ protected void onActivityResult(final int requestCode, final int resultCode, fin
...
         }
     }
 
+    @Override
+    public void refreshWithFilter(final GeocacheFilter filter) {
+        currentCacheFilter.set(filter);
+        setFilter();
+        refreshFilterForOnlineSearch();
+
+        refreshCurrentList(AfterLoadAction.CHECK_IF_EMPTY);
```
murggel, line 1365: without the following call of the methods, it did not work before, have to check with the new FilterContext...

main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -76,6 +76,7 @@
     private TextParam positiveButton = TextParam.id(android.R.string.ok);
     private TextParam negativeButton = TextParam.id(android.R.string.cancel);
     private TextParam neutralButton = TextParam.id(android.R.string.untitled);
+    private boolean neutralButtonNeedsSelection = true;
```
murggel, line 79: otherwise the neutral button was not shown / was enabled only, if an item was selected

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,105 @@
+package cgeo.geocaching.utils;
```
murggel, line 1: moved methods from MapUtils to FilterUtils.

murggel:
ready for a first review - I added some comments, where I need feedback

main/src/cgeo/geocaching/maps/google/v2/GoogleMapActivity.java
```diff
@@ -252,7 +253,18 @@ protected void onActivityResult(final int requestCode, final int resultCode, fin
...
+        FilterUtils.openFilterActivity(this, mapBase.getFilterContext(), mapBase.getCaches());
     }
 
+    @Override
+    public boolean showFilterList(final View view) {
+        return FilterUtils.openFilterList(this, mapBase.getFilterContext());
+    }
+
+    @Override
+    public void refreshWithFilter(final GeocacheFilter filter) {
```
murggel, line 265: refresh of the map is currently not working correct

main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -915,8 +916,6 @@ public boolean onOptionsItemSelected(final MenuItem item) {
         } else if (menuItem == R.id.menu_invert_selection) {
             adapter.invertSelection();
             invalidateOptionsMenuCompatible();
-//        } else if (menuItem == R.id.menu_filter_legacy) { //remove for upcoming beta
-//            showLegacyFilterMenu(null);
```
eddiemuc, line 919: Thanks for cleaning up my mess :-)

main/src/cgeo/geocaching/CacheListActivity.java
```diff
@@ -1199,7 +1206,7 @@ private Geocache getCacheFromAdapter(final AdapterContextMenuInfo adapterInfo) {
 
     private void setFilter() {
         applyAdapterFilter();
-        prepareFilterBar();
+        updateFilterBar();
```
eddiemuc, line 1209: I like it that you refactor things along the way!

main/src/cgeo/geocaching/maps/google/v2/GoogleMapActivity.java
```diff
@@ -252,7 +254,18 @@ protected void onActivityResult(final int requestCode, final int resultCode, fin
...
+        mapBase.getMapOptions().filterContext.set(filter);
+        MapUtils.filter(mapBase.getCaches(), mapBase.getMapOptions().filterContext);
+        mapBase.refreshMapData(true);
```
eddiemuc, line 269: Are you sure you want to pass "true" here? IIRC true means that the method assumes that circleDrawing has switched (something you can set in the map settings).

main/src/cgeo/geocaching/maps/google/v2/GoogleMapActivity.java
eddiemuc: Those cases are very tricky to find, I usually find them only by finding a reproducable case, then set a breakpoint in "MapUtils.filter" (or now "FilterUtils.filter") and see whether it is reached or not. It is hard to tell from looking at the code alone...

If you run in a dead end here, then I would need to get the stuff running on my local machine to dive in deep.

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public static boolean openFilterList(final Activity activity, final GeocacheFilterContext filterContext) {
+        final List<GeocacheFilter> filters = new ArrayList<>(GeocacheFilter.Storage.getStoredFilters());
+
+        if (filters.isEmpty()) {
+            return false;
+        } else {
+            final boolean isFilterActive = filterContext.get().isFiltering();
+            final FilteredActivity filteredActivity = (FilteredActivity) activity;
```
eddiemuc, line 44: This will generate a classcastexception in case activity is not a filteredactivity. Can you make the parameter of type "FilteredActivity"? (this is a minor thing)

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,107 @@
...
+            filterView.setVisibility(GONE);
+        } else {
+            final TextView filterTextView = activity.findViewById(R.id.filter_text);
+            filterTextView.setText(TextUtils.join(", ", filterNames));
+            filterView.setVisibility(View.VISIBLE);
+        }
+    }
+
+    public static void connectFilterBar(final Activity activity) {
+        final FilteredActivity filteredActivity = (FilteredActivity) activity;
```
eddiemuc, line 81: Same here (classcastexception)

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,107 @@
...
+        if (filterNames.isEmpty()) {
+            filterView.setVisibility(GONE);
+        } else {
+            final TextView filterTextView = activity.findViewById(R.id.filter_text);
+            filterTextView.setText(TextUtils.join(", ", filterNames));
+            filterView.setVisibility(View.VISIBLE);
+        }
+    }
+
+    public static void connectFilterBar(final Activity activity) {
```
eddiemuc, line 80: I was a bit confused by the method name. How about "initializeFilterBar" or "createFilterBar"?

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,107 @@
...
+        final FilteredActivity filteredActivity = (FilteredActivity) activity;
+        if (null != filteredActivity) {
+            final View filterView = activity.findViewById(R.id.filter_bar);
+            if (null != filterView) {
+                filterView.setOnLongClickListener(v -> filteredActivity.showFilterList(null));
+            }
+        }
+    }
+
+    public static void connectFilterMenu(final Activity activity) {
```
eddiemuc, line 90: same here, maybe "initializeFilterMenu" or "createFilterMenu"?

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,107 @@
...
+
+        new Handler().post(new Runnable() {
+            @Override
+            public void run() {
+                final FilteredActivity filteredActivity = (FilteredActivity) activity;
```
eddiemuc, line 95: same here (classcastexception)

eddiemuc (approved):
Looks very good already!

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: Unfortunately `FilteredActivity` is not an `Activity` and `SimpleDialog`needs an activity. Only thing I can probably do is to pass it as an additional parameter.

Or is there a possibility to check whether its of type of `FilteredActivity` ? I thought the cast would leed to a null-object...

main/src/cgeo/geocaching/maps/google/v2/GoogleMapActivity.java
murggel: Was not on purpose, I will change it

main/src/cgeo/geocaching/utils/FilterUtils.java
moving-bits: You can check with `if (activity instanceof FilteredActivity) { ... }`

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: renamed with commit 2f99647

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: renamed with commit 2f99647

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: done with commit abf0f09

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: done with commit abf0f09

main/src/cgeo/geocaching/utils/FilterUtils.java
murggel: done with commit abf0f09

main/src/cgeo/geocaching/utils/FilterUtils.java
eddiemuc: Just for the record: if you want to force a method parameter to extend two different classes/interfaces then you can do something like this (using generics):

```
    public static <T extends Activity & FilteredActivity> void doSomething(final T param) {
        //do stuff here...
    }
```

But the solution proposed by @moving-bits is totally fine too of course.

murggel: But I like your approach more 😉 I will change it

moving-bits: > Just for the record: if you want to force a method parameter to extend two different classes/interfaces then you can do something like this (using generics):

Ah, didn't know that. Way more elegant if the checks can be done at compile-time.

main/src/cgeo/geocaching/maps/google/v2/GoogleMapActivity.java
murggel: Looks good for me now, probably it was bad internet-connection

murggel:
Sorry, did a forced push with squashing some commits, so commits from earlier comments / review are gone...
Was not aware of that...

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,97 @@
...
+            filterView.setVisibility(View.GONE);
+        } else {
+            final TextView filterTextView = activity.findViewById(R.id.filter_text);
+            filterTextView.setText(TextUtils.join(", ", filterNames));
+            filterView.setVisibility(View.VISIBLE);
+        }
+    }
+
+    /** filterView must exist */
+    public static void initializeFilterBar(@NonNull final Activity activity, @NonNull final FilteredActivity filteredActivity) {
```
murggel, line 79: I did not get the CGeoMap to work with `<T extends Activity & FilteredActivity>`
so, for the other activities it looks strange, calling it with `initializeFilterBar(this, this)`

@eddiemuc probably you have a solution?

main/src/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -0,0 +1,97 @@
...
+        final View filterView = activity.findViewById(R.id.filter_bar);
+        filterView.setOnLongClickListener(v -> filteredActivity.showFilterList(null));
+    }
+
+    public static void initializeFilterMenu(@NonNull final Activity activity, @NonNull final FilteredActivity filteredActivity) {
```
murggel, line 84: I did not get the CGeoMap to work with `<T extends Activity & FilteredActivity>`
so, for the other activities it looks strange, calling it with `initializeFilterMenu(this, this)`

@eddiemuc probably you have a solution? 
Passing the view for menu_filter does not work (because of the Runnable handler), it will be null then on passing it to initializeFilterMenu

eddiemuc: Right, in CgeoMap the activity class and FilteredActivity class are probably different classes hm? Then the approach will not work of course. So better don't use it.

Apart from that I am not sure what the problem is?

murggel: No problem then. Was just not really sure about that.

murggel:
Ready for finishing review. The problem with "sometimes the map does not refresh" did not occur anymore, probably a internet connection effect

eddiemuc:
Looks good! Thanks @murggel for this contribution. Let's merge

eddiemuc:
Works like a charm!

murggel:
Thanks for your reviews and patience

Issue #11248, opened 2021-07-16
Map doesn't support vector drawables for markers


ztNFny:
Replace "marker.png" with a vector drawable -> crash on loading the map:

    java.lang.ClassCastException: android.graphics.drawable.VectorDrawable cannot be cast to android.graphics.drawable.BitmapDrawable
        at cgeo.geocaching.utils.CompactIconModeUtils.setCompactIconModeThreshold(CompactIconModeUtils.java:24)
        at cgeo.geocaching.maps.CGeoMap.initializeMap(CGeoMap.java:496)

@moving-bits any chance you could take a look at that? It's a blocker for me.

moving-bits:
What this functions tries to achieve is to calculate something like "how many drawables of the size of a "marker" does it take to fill up one fourth of the screen" (assumed they are rectangular, so considering height/width only).

How can we determine (programmatically) width and height of a svg in current context?

ztNFny:
Same problem in CompassMiniView

    public void onAttachedToWindow() {
        super.onAttachedToWindow();
        if (instances++ == 0) {
            final int drawable = isInEditMode() || !Settings.isLightSkin(getContext()) ? R.drawable.compass_arrow_mini_white : R.drawable.compass_arrow_mini_black;
            compassArrow = BitmapFactory.decodeResource(getResources(), drawable);
            compassArrowWidth = compassArrow.getWidth();
            compassArrowHeight = compassArrow.getWidth();
        }
    }

murggel:
> What this functions tries to achieve is to calculate something like "how many drawables of the size of a "marker" does it take to fill up one fourth of the screen" (assumed they are rectangular, so considering height/width only).
> 
> How can we determine (programmatically) width and height of a svg in current context?
> 
> --
> 
> Edit: Conversion to a bitmap is done only for calculation, not needed for drawing.

Is the calculation still needed when using svg? Or can we define a minimum size and calculate the size for the svg for the given amount of markers?

moving-bits:
I probably have solutions for both cases, PR will follow. Should work with both PNG and SVG images. Target to `master`?

PR #11307, opened 2021-07-23
make another calculation vectordrawable compatible (rel. to #11248)

moving-bits:
## Description
Since dot_black has been changed to a vector drawable the size calculation needs to be adapted


ztNFny:
Crash is fixed, thanks!

Do I get it right that the list marker get's the size of dot_black? So by adjusting that I can control the list marker size?

moving-bits:
> Do I get it right that the list marker get's the size of dot_black? So by adjusting that I can control the list marker size?

Exactly. Or instead of using a reference drawable element, you can calculate the desired size directly (e. g. by using one of the methods in `ViewUtils`) (since the size of a vector drawable will always be the same (in `dp`) as opposed to bitmap drawables, which may have different sizes depending on resoultion).

Issue #11276, opened 2021-07-20
Add [does not contain] to text filters 


RoNoLo:
## What function are you missing
I would like to exclude caches from the map by word/regex filter of the title and description.

As a father with two little kids I would like to exclude certain caches. Morst of them are T5, because they are high up in the trees and usually contain some words like "angel cache", "angeln" or similar in other languages. As long as geocaching.com has no category for these kind of caches (they are just all mushed into the T5 rating), I wish there were a way to still exclude them from my map of available caches. 

## Suggested solution
A text filter, which is applied to title and description, which can exclude caches from viewing them (like a postprocessing after retrieving them from the officially APIs). My use case would be a list of words or regular expressions, which remove all "tree fishing" (German: angel caches) from my view.

## Alternatives/workarounds
A filter of "terrain" ratings would serve _me_ probably evenly well, because everything with a rating 5 is out of the question for me and my children.

## Additional context
Not really, just that I wish c:geo would allow me to filter even more which caches I want to see and find. Another example could be caches were the description contains that you have to bring a "bottle" or "batteries" or similar.

eddiemuc:
Hi @RoNoLo, you might want to take a look at the new filter system which we plan to introduce in next release. You can already experience this system in our current beta version (information on how to become a beta tester can be found here: https://www.cgeo.org/development)

A draft version of the new filter manual can be found here: https://manual.cgeo.org/en/cachefilter (currently english only).

Would this help your use case?

![image](https://user-images.githubusercontent.com/6909759/126335249-b03dd273-4af4-4823-a5c6-c4d91b90e04e.png)

RoNoLo:
Thanks for your answer.

By looking at the screenshot I assume, that it's a collection of different filters which are connected via a pipe (the result will be piped to the next filter) until all filters are processed, thus every possible combination of Only Tradi, but less that T5 and also not containing the words "bottle", "angel" are inside the description are matched.

If that sums up what is in the planning I would be very pleased with this type of update. 

You could close this issue, if that is the case.

murggel:
@eddiemuc I think the selection "does not contain" is missing for those text-fields ...

RoNoLo:
Based on the screenshot I *assume* that the [CONTAINS] button can be switched to other operators, like "does not contains", "regex", ...

fm-sys:
Indeed, "does not contain" would be nice to have for consistency...

[...]

PR #11308, opened 2021-07-23
introduce spinner button (rel. to #11276)

moving-bits:
## Description
- Related to a discussion in https://github.com/cgeo/cgeo/issues/11276#issuecomment-885772840
- introduce a new outlined button style "button_full_spinner" based on "button_full"
- Switched "log cache" and "log trackable" type buttons to the new style

image.image

### DIFF
--- a/main/res/drawable/outline_arrow_drop_down_24.xml
+++ b/main/res/drawable/outline_arrow_drop_down_24.xml
@@ -1,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24"
+    android:tint="?attr/colorControlNormal">
+  <path
+      android:fillColor="@android:color/white"
+      android:pathData="M7,10l5,5 5,-5H7z"/>
+</vector>
--- a/main/res/layout/logcache_activity.xml
+++ b/main/res/layout/logcache_activity.xml
@@ -29,7 +29,7 @@
 
                 <Button
                     android:id="@+id/type"
-                    style="@style/button_full"
+                    style="@style/button_full_spinner"
                     android:layout_width="0dip"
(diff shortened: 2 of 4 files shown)


fm-sys:
Looks nice in general. Could we move the icon to the right border? That's how it is suggested by material.io... (although the screenshots are for text inputs) 

Screenshot_20210724_000310_org mozilla firefox

moving-bits:
I have created three variations for comparison:

grafik

- first one has "text centered", "icon right-aligned"
- second one has "text + icon centered"
- third one has "text left-aligned", "icon right-aligned"

My personal preferences would be:
- second one (as it is completely centered, as the button to the right is as well)
- first one (text is slightly off-centered, as it is centered in "button-width without icon")
- third one (would be best probably, if used as a single button)

fm-sys:
Thanks for the visualization. You are right, your initial choosen variant (in the new screenshots the second one) probably looks best.

Lineflyer:
I like the first one best...maybe because I am somehow used to, that a spinner is shown at the right end of a text field.

fm-sys:
> I like the first one best...maybe because I am somehow used to, that a spinner is shown at the right end of a text field.

Would also be OK for me. Whether 1 or 2 doesn't really matter to me. Both are IMHO fine. Btw, what happens if the font size is very large? Will the icon remain visible in both versions?

bekuno:
Maybe we can use the symbol U+25BC (9660) or U+25BE (9662)?

eddiemuc:
I like the first two approaches with a slight preference to option 2. 

Please don't use option 3

eddiemuc:
Although there's an ongoing discussion between option 1 and 2, I will now merge this PR (implementing option 2) so I can use the new style also on other places (and e.g. fix #11312). This is due to last beta being produced this evening, and I really want to get this fixed in next release. 

@Lineflyer if discussion reveals that option 1 should be used instead, we can make another PR. I really hope this is ok for everyone...

Issue #11204, opened 2021-07-12
Androidx.fragment.app.Fragment$InstantiationException


Lineflyer:
Crash report on Google play. 120 reports by 70 users for 2021.07.10-RC

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3792)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3968)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:85)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2307)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:246)
  at android.app.ActivityThread.main (ActivityThread.java:8512)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:602)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1130)
Caused by: androidx.fragment.app.Fragment$InstantiationException: 
  at androidx.fragment.app.Fragment.instantiate (Fragment.java:628)
  at androidx.fragment.app.FragmentContainer.instantiate (FragmentContainer.java:57)
  at androidx.fragment.app.FragmentManager$3.instantiate (FragmentManager.java:483)
  at androidx.fragment.app.FragmentStateManager.<init> (FragmentStateManager.java:85)
  at androidx.fragment.app.FragmentManager.restoreSaveState (FragmentManager.java:2728)
  at androidx.fragment.app.FragmentController.restoreSaveState (FragmentController.java:198)
  at androidx.fragment.app.FragmentActivity$2.onContextAvailable (FragmentActivity.java:149)
  at androidx.activity.contextaware.ContextAwareHelper.dispatchOnContextAvailable (ContextAwareHelper.java:99)
  at androidx.activity.ComponentActivity.onCreate (ComponentActivity.java:297)
  at androidx.fragment.app.FragmentActivity.onCreate (FragmentActivity.java:273)
  at cgeo.geocaching.activity.AbstractActivity.onCreate (AbstractActivity.java:128)
[...]

PR #11310, opened 2021-07-24
fix fragment instantiation exception in TabbedViewPager

moving-bits:
## Description
PR attempts to fix the fragment instantiation error described in #11204 by
- adding common fragment constructor,
- removing weakReferences to calling activity in fragments,
- supporting indirect fragment instantiation and `getArguments()` for fragments that need parameters,
- rebuilding `fragmentMap` dynamically on configuration changes
- ...

## How to test
For me the following steps worked reliably to force the error:
- open cache detail page
- switch to Android system's language selection screen and select a different language
- return to c:geo => configation change triggers the error described in the referenced issue

### DIFF
--- a/main/src/cgeo/geocaching/AboutActivity.java
+++ b/main/src/cgeo/geocaching/AboutActivity.java
@@ -152,8 +152,13 @@ public class AboutActivity extends TabbedViewPagerActivity {
         }
 
         @Override
+        public long getPageId() {
+            return Page.VERSION.id;
+        }
+
+        @Override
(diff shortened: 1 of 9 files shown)


fm-sys:
>     * switch to Android system's language selection screen and select a different language

BTW, same applies also for toggling the "system dark mode"

moving-bits:
Needs to be included in next beta as well, to see if the fix covers all cases for the related issue, which happens quite frequently in current betas - therefore merging this now.

PR #11320, opened 2021-07-25
rel to PR #11296: remove usage of raw types

eddiemuc:
rel to PR #11296: remove usage of raw types

@murggel: found a way to make the raw types "unraw". 

Maybe at a later point in time we may try to remove the Difficulty/Terrain-Filters as well as their FilterType completely, since they are no longer used excplicitely anywhere. But this is not necessary for beta I believe :-)


murggel:
> rel to PR #11296: remove usage of raw types
> 
> @murggel: found a way to make the raw types "unraw". 
> 
> Maybe at a later point in time we may try to remove the Difficulty/Terrain-Filters as well as their FilterType completely, since they are no longer used excplicitely anywhere. But this is not necessary for beta I believe :-)

I am not really sure to remove the single filter, the combined one uses only AND, so to be more flexible we have to add an option for OR or leave the 2 single ones selectable to the user  - but we will see from feedback

eddiemuc:
> I am not really sure to remove the single filter, the combined one uses only AND, so to be more flexible we have to add an option for OR or leave the 2 single ones selectable to the user - but we will see from feedback

Good point. Let's wait (for user feedback) and see

PR #11325, opened 2021-07-25
add formatted=false for search_filter_info_message

bekuno:
CI finds a error in the strings.xml:
/home/jenkins/slave/workspace/cgeo pull request/main/res/values/strings.xml:1583:4: Multiple substitutions specified in non-positional format of string resource string/search_filter_info_message. Did you mean to add the formatted="false" attribute?
(see eg. https://ci.cgeo.org/job/cgeo%20pull%20request/6546/console)

As this is a result of the html '%22' entity we can use the exclude attribute here.


eddiemuc:
Jupp, this should help

bekuno:
did not help :-(

bekuno:
> 
> 
> did not help :-(

It is ok.
The message for `main/res/values/strings.xml` is gone.
I hope that the next codacy sync will correct this for the other languages.

bekuno:
> I hope that the next codacy sync will correct this for the other languages.

Sync was successful.

PR #11337, opened 2021-07-27
change map rotation notice into OneTimeDialog (fix #11336)

### LINKED ISSUE #11336: add "do not show this message again" to gm map rotation notice
### Tell us your idea!

On using GM maps with "auto rotation enabled = automatic" you will receive a warning notice "to turn off map rotation go to ..." every now and then, which can get unnerving if you actually WANT auto rotation to be (and stay) enabled.

add a checkbox "do not ask me again" to this notice (and act accordingly)

### Is there an alternative?

_No response_

### Additional Information

from support / ticket #344376

moving-bits:
## Description
change map rotation notice into a OneTimeDialog (will be shown until "do not ask again" checkbox is set)

### DIFF
--- a/main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java
+++ b/main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java
@@ -23,9 +23,11 @@ import cgeo.geocaching.maps.interfaces.PositionAndHistory;
 import cgeo.geocaching.maps.mapsforge.AbstractMapsforgeMapSource;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.utils.Log;
 import static cgeo.geocaching.maps.google.v2.GoogleMapUtils.isGoogleMapsAvailable;
+import static cgeo.geocaching.storage.extension.OneTimeDialogs.DialogType.MAP_AUTOROTATION_DISABLE;
 
 import android.app.Activity;
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


main/src/cgeo/geocaching/maps/google/v2/GoogleMapView.java
```diff
@@ -146,7 +148,8 @@ public void onMapReady(final GoogleMap googleMap) {
             final float bearing = cameraPosition.bearing;
             if (canDisableAutoRotate && bearing == 0.0f && Settings.getMapRotation() == Settings.MAPROTATION_AUTO) {
                 canDisableAutoRotate = false;
-                SimpleDialog.of((Activity) getContext()).setTitle(R.string.map_gm_autorotation).setMessage(R.string.map_gm_autorotation_disable).confirm((dialog, which) -> {
+                final Context context = getContext();
+                Dialogs.advancedOneTimeMessage(context, OneTimeDialogs.DialogType.MAP_AUTOROTATION_DISABLE, context.getString(R.string.map_gm_autorotation), context.getString(R.string.map_gm_autorotation_disable), "", true, null, () -> {
```
fm-sys, line 152: Is there a reason why you haven't used the `Dialogs.basicOneTimeMessage()` method which just requires Context and DialogType (-> way simpler)?

fm-sys: Ok, just saw that basicOneTimeMessages aren't cancelable and have the "info icon". The icon would be fine IMHO and for the cancelable flag we should edit our basic OneTimeMessage instead. I don't remember a reason why these dialogs should not be able to cancel...

moving-bits: IIRC basicOneTimeMessage only displays a notice, but does not allow for starting some follow-up action on confirmation - which is something I need here.

fm-sys: True, have not seen this. Then it maybe would make sense to reference title and text via `MAP_AUTOROTATION_DISABLE.messageTitle` / ``MAP_AUTOROTATION_DISABLE.messageText`. 

Or we need to create some kind of "halfAdvanced" OTM. 

But I would also be fine with merging the current state.

moving-bits: I'll update the PR to using `MAP_AUTOROTATION_DISABLE.messageTitle` / `MAP_AUTOROTATION_DISABLE.messageText`

PR #11344, opened 2021-07-27
use Resource to get text-color for all API-version (fix #11342)

### LINKED ISSUE #11342: Black scale for D/T filter in dark mode for API 21
While checking for D/T filter config of #11331 I realized that the text color for the scale is dark in dark mode

API 21

murggel:
## Description

fixes dark slider-scale color in dark-mode for API < 23

use Resource to get text-color for all API-version

## Related issues

fix #11342


moving-bits (approved):
LGTM, and is way easier than the old solution!

PR #11349, opened 2021-07-28
make button outline more visible

moving-bits:
## Description
Our outline button uses a border stroke color which is barely visible, at least in dark mode. This PR increases opacity for stroke color a bit to make the outline more visible.

grafik.grafik

### DIFF
--- a/main/res/color/button_strokecolor_selector.xml
+++ b/main/res/color/button_strokecolor_selector.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- derived from mtl_btn_stroke_color_selector, but higher alpha value -->
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:color="?attr/colorPrimary" android:state_checked="true"/>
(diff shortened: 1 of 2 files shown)


Lineflyer:
From the screenshots I agree, that this helps to get more structure.

Did you only change it for dark theme or also for white theme.
If also white theme is changed, could you provide before/after for that as well.

moving-bits:
> Did you only change it for dark theme or also for white theme.
> If also white theme is changed, could you provide before/after for that as well.

Both themes are using the same declaration - sorry, forgot to add the pics for light theme as well. Here we are:

grafik.grafik

For light theme the old color is not so dramatic, but the new one fits as well, IMHO.

Lineflyer:
I tested in on my device and think its much better than before.
Additionally it harmonizes these button with the outline of text fields.

Will merge as there are no objections so far, we can still modify/revert if needed before releasing.

Lineflyer:
Damn, it was merged to `master`. Will try to cherry-pick to release.

Lineflyer:
Cherry-picked the commit to `release` as well.

PR #11359, opened 2021-07-29
get ALC True Found State

baiti:
This PR addresses getting the true "found it" state for adventure lab caches. This has been discussed before, however, with the way we interacted with the API initially, the true found state could not be acquired. With the new mechanisms, this is now possible to get that state.

In addition to adding this new function, the code in LCApi has been refactored to use more meaningful and more consistent names for variables and member functions.

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -71,9 +71,7 @@ final class LCApi {
             return null;
         }
         final Parameters headers = new Parameters(CONSUMER_HEADER, CONSUMER_KEY);
-        final String at = getAlcAccessToken();
-        Log.d("token " + at);
-        headers.add("authorization", "bearer " + at);
+        headers.add("authorization", "bearer " + getAlcAccessToken());
         try {
             final Response response = apiRequest(ALC_ADVENTURES_URI + geocode.substring(2), null, headers).blockingGet();
             return importCacheFromDetail(response);
@@ -230,7 +228,7 @@ final class LCApi {
             cache.setType(CacheType.ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
             cache.setArchived(response.get("IsArchived").asBoolean()); // we get that even in passive mode!
-            cache.setFound(response.get("IsComplete").asBoolean());    //as soon as we're using active mode
+            cache.setFound(response.get("IsComplete").asBoolean());    // active mode only
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.CACHE));
             return cache;
         } catch (final NullPointerException ex) {
@@ -258,7 +256,7 @@ final class LCApi {
             cache.setCoords(new Geopoint(location.get("Latitude").asText(), location.get("Longitude").asText()));
             cache.setType(CacheType.ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
-            cache.setFound(response.get("IsComplete").asBoolean());    //as soon as we're using active mode
+            cache.setFound(response.get("IsComplete").asBoolean()); // active mode only
             cache.setDisabled(false);
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             cache.setOwnerDisplayName(response.get("OwnerUsername").asText());
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.connector.lc;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.LoadFlags.SaveFlag;
@@ -17,6 +18,7 @@ import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.network.Network;
 import cgeo.geocaching.network.Parameters;
 import cgeo.geocaching.sensors.Sensors;
+import cgeo.geocaching.settings.Credentials;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.JsonUtils;
@@ -38,6 +40,7 @@ import java.util.Locale;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.reactivex.rxjava3.core.Single;
 import io.reactivex.rxjava3.functions.Function;
 import okhttp3.Response;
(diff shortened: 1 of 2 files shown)


baiti:
@fm-sys ... pls review from an Android / Java coding perspective.

baiti:
please note: One of the commits (strings.xml) is unrelated to the topic at hand. It is just, what I believe "better EnglisH" (at least US).

It is about comfort vs convenience. I believe, "convenience" is the better fit for what we are trying to say with that message but I certainly leave it to native speakers to make the final call.

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -62,9 +71,12 @@ protected static Geocache searchByGeocode(final String geocode) {
             return null;
         }
         final Parameters headers = new Parameters(CONSUMER_HEADER, CONSUMER_KEY);
+        final String at = getAlcAccessToken();
+        Log.d("token " + at);
```
fm-sys, line 75: Is the access token something which should not be shared? If so, it should not be logged in a production version, as log files might be uploaded to github, etc...

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -222,21 +230,19 @@ private static Geocache parseCache(final JsonNode response) {
             cache.setType(CacheType.ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
             cache.setArchived(response.get("IsArchived").asBoolean()); // we get that even in passive mode!
-            // cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode
+            cache.setFound(response.get("IsComplete").asBoolean());    //as soon as we're using active mode
```
fm-sys, line 233: //as soon as we're using active mode

Maybe change the comment to something more fitting for the current situation?

main/src/cgeo/geocaching/connector/lc/LCApi.java
```diff
@@ -252,17 +258,16 @@ private static Geocache parseCacheDetail(final JsonNode response) {
             cache.setCoords(new Geopoint(location.get("Latitude").asText(), location.get("Longitude").asText()));
             cache.setType(CacheType.ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
-            // cache.setArchived(response.get("IsArchived").asBoolean()); as soon as we're using active mode
-            // cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode
+            cache.setFound(response.get("IsComplete").asBoolean());    //as soon as we're using active mode
```
fm-sys, line 261: Here as well...

main/src/cgeo/geocaching/connector/lc/LCConnector.java
```diff
@@ -108,7 +108,7 @@ public String getExtraDescription() {
 
     @Override
     public boolean supportsSettingFoundState() {
-        return true;
+        return false;
```
fm-sys, line 111: As false is the default value anyway, you don't need to overwrite the method at all.

See:
https://github.com/cgeo/cgeo/blob/b571526c5ca19046802364637d9715006d2cf68d/main/src/cgeo/geocaching/connector/AbstractConnector.java#L118-L122

main/res/values/strings.xml
```diff
@@ -299,7 +299,7 @@
...
+    <string name="feature_info_offline_counter">You have some caches with an offline saved log, which might not yet be logged online. For your convenience the count of these logs is displayed as \"+ x\" behind your number of found caches in the footer at c:geo\'s main page. \n\nMoreover these \"x\" logs will be added to the \"NUMBER\" template you might use in your log signature. \n\nIf you don\'t like this counter extension, you can disable it under \"Settings\" > \"Logging\" and/or use the \"NUMBER (without offline logs)\" template instead. \n\nTo easily find remaining offline logs, go to the list \"History\" and filter for \"With personal data\" > \"With offline log\".</string>
```
fm-sys, line 302: With the bottom navigation and a redesigned dashboard, the dialog will probably be removed anyway, as the new style will be way more self-explaining. So I don't know if it makes sense to change anything here...?

signal-2021-07-21-195912

fm-sys (commented):
Wow, simply incredible! Really want to see this in production soon 😍

main/src/cgeo/geocaching/connector/lc/LCApi.java
baiti: the access_token is valid for 1 hour .... however, I'll remove the log statement.

main/src/cgeo/geocaching/connector/lc/LCConnector.java
baiti: can I leave it in in case we need to go back and set it to true?

eddiemuc: I would say yes :-)

Lineflyer:
First of all:
Thanks @baiti for your efforts.
I think this is reasonable and a great help for users.
Reading through the related communication I do not see blocking points as we still stick to the guidance provided  (i.e. present the coords of the ALC and refer to the ALC app and let users play there). That additional feature is a pure comfort feature from users perspective and quite reasonable.

My questions though:
1. As you say the token has a limited live time: Is this treated accordingly...meaning: Is the feature still working after that time without restart or alike?

2. I would suggest to leave out the string change for a very simple reason: If the string is changed all translations will fallback to English until new translations are provided by the crowd. Given the limited time until release and the minor importance of that change it might be not meaningful to go that way. You can still open that as separate PR towards `master` if desired.

baiti:
> First of all:
> Thanks @baiti for your efforts.
> I think this is reasonable and a great help for users.
> Reading through the related communication I do not see blocking points as we still stick to the guidance provided (i.e. present the coords of the ALC and refer to the ALC app and let users play there). That additional feature is a pure comfort feature from users perspective and quite reasonable.
> 
> My questions though:
> 
> 1. As you say the token has a limited live time: Is this treated accordingly...meaning: Is the feature still working after that time without restart or alike?

Of course, the method `getAlcAccessToken()` automatically refreshes the token once it's life time is over.

> 2. I would suggest to leave out the string change for a very simple reason: If the string is changed all translations will fallback to English until new translations are provided by the crowd. Given the limited time until release and the minor importance of that change it might be not meaningful to go that way. You can still open that as separate PR towards `master` if desired.

What is the easiest way to drop that commit? (This perhaps involves some git magic)

eddiemuc:
> What is the easiest way to drop that commit? (This perhaps involves some git magic)

I personally work locally with only one commit per feature, and whenever I update something I don't create a new commit but use "got commit --amend" to squash my changes into the previous commit.

If one uses this workflow, the easiest reversal in your case would have been to simply put in the old string again and do another "git commit --amend"

> Update: I believe I managed to drop the commit in question. I did an interactive rebase on HEAD~3 and dropped the commit in question, followed by a force push, was that the right thing to do?

Yes this works fine

> Update to the update: Would that interactive rebase with dropping that specific commit still have worked if I had squashed all commits into one first? Somehow I'd assume after squashing them they can't be addressed individually anymore?

You assume correctly. In that case you could have used "git commit --amend" though to modify the last commit on the branch (as I described above)

Lineflyer:
Any objections to merge this functionality?

fm-sys:
LGTM, feel free to merge! Should be included in the new beta IMHO...

Lineflyer:
I merged this to `release` and upstream to `master` plus created a new nightly.
Its somehow not working for me.

Any dependency for ALC already loaded before? I tried to refresh several of my found ALC but not found state appears.

baiti:
> I merged this to `release` and upstream to `master` plus created a new nightly.
> Its somehow not working for me.
> 
> Any dependency for ALC already loaded before? I tried to refresh several of my found ALC but not found state appears.

There shouldn't ne any dependencies. I'll check, stay tuned.

baiti:
Seems I can't debug via the emulator right now. ADB complains periodically:

```
2021-07-30 16:00:18,404 [ 258590]   WARN -            #com.android.ddmlib - Adb connection Error:EOF 
2021-07-30 16:00:18,404 [ 258590]   WARN -            #com.android.ddmlib - Unable to open connection to ADB server: java.io.IOException: Can't find adb server on port 5037, IPv4 attempt: Connection refused, IPv6 attempt: Connection refused 
2021-07-30 16:00:18,404 [ 258590]   WARN -            #com.android.ddmlib - Failed to start monitoring emulator-5554 
2021-07-30 16:00:18,404 [ 258590]   WARN -            #com.android.ddmlib - Cannot reach ADB server, attempting to reconnect
```

I have the emulator running and I can't launch an application in it. I am sure this is related to installing the update to Arctic Fox but I don't know how I could get back to the state before that update. Is that possible at all?

eddiemuc:
I can confirm that the "found" state is not retrieved (but the labs itself are, so that's a clear plus! :-)).

I tried to use the consumer key sent around together with my Geocaching.com credential on the swagger Api to get the token (using this URL: https://labs-api.geocaching.com/swagger/ui/index#!/Accounts/Accounts_Login)

But I got the response "invalid client id" (HTTP 401)

I think this is the root cause why "found" states are not working. Maybe out consumer key does not support login functionality somehow.

baiti:
> I can confirm that the "found" state is not retrieved (but the labs itself are, so that's a clear plus! :-)).
> 
> I tried to use the consumer key sent around together with my Geocaching.com credential on the swagger Api to get the token (using this URL: https://labs-api.geocaching.com/swagger/ui/index#!/Accounts/Accounts_Login)
> 
> But I got the response "invalid client id" (HTTP 401)
> 
> I think this is the root cause why "found" states are not working. Maybe out consumer key does not support login functionality somehow.

.... it supported that functionality until very very recently. Yesterday I did all the coding / testing and it definitely worked. More on the tem channel ....

Lineflyer:
As agreed with @baiti I just reverted this commit to allow a new beta release without running into trouble because of failing connections.

@baiti will repush this PR later or once we clarified the facts.

PR #11372, opened 2021-07-31
Search GUI optimization

### LINKED ISSUE #11340: Global search - set "Search with filter" at the first position
### Tell us your idea!

I think that the "Search with Filter" should go to the top of the list.
This is then the best way to use them.
The current position after "Hidden  by" and "Trackable" looks strange to me.

By the way, how about also move the "Geocode" search on the 2nd place after the Filter search?
(Maybe also other reorderering could be discussed)

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

### LINKED ISSUE #11354: Search GUI optimization
### Tell us your idea!

I stumbled at the search GUI over some naming/translation and GUI issues.

**1)** design
The individual sections are now hardly distinguishable.
The headline shown use a more visible style (bold?) and the line bolder than the box/button outline. 

**2)** Coordinates vs. Coordinate
The word "Coordinates" is in c:geo used on many places, not only in the search GUI.
The word Coordinates is the plural form of Coordinate. And mostly we mean a single coordinate (consisting of latitude and longitude). And we search in the search GUI by a single coordinate.
So lets use the single form here and check also other places.

**3)** Keywords vs. Keyword
The headline says "Keywords", the helper text and the button "Keyword".
The headline should use the single form (as the other headlines).
Can we really search for multiple keywords? (I did not tested it). If yes then use "Keyword(s)".

The German translation is also poor. 
Headline "Stichworte" (=plural), helper text and search button "Stichwörter" (=plural).
I propose to use "Stichwort" and if necessary "Stichworte".

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

bekuno:
## Description

-    Filter as first entry
-    Geocode as second entry
-    Latitude/Longitude as a two row button
-    Headline text bold, line in 3dip
-    "Geocode" instead of "geo code"
-    "Coordinate" instead of "Coordinates"
-    "Keyword" instead of "Keywords"

## Related issues

fixes #11340
fixes #11354

### DIFF
--- a/main/res/layout/search_activity.xml
+++ b/main/res/layout/search_activity.xml
@@ -17,6 +17,61 @@
             <View style="@style/separator_horizontal" />
             <TextView
                 style="@style/separator_horizontal_headline"
+                android:text="@string/search_filter" />
+        </RelativeLayout>
+
+        <RelativeLayout
+            android:layout_height="wrap_content"
+            android:layout_width="match_parent">
+
+            <Button
+                android:id="@+id/search_filter"
+                style="@style/button_full"
+                android:text="@string/search_filter_button"
+                android:layout_alignParentLeft="true"
+                android:layout_toLeftOf="@+id/search_filter_info"
+                android:layout_centerVertical="true"
+                android:layout_width="wrap_content"/>
+
+            <ImageView
+                android:id="@+id/search_filter_info"
+                android:layout_width="36dp"
+                android:layout_height="36dp"
+                app:srcCompat="@drawable/settings_info"
+                android:layout_marginLeft="0dp"
+                android:layout_marginRight="10dp"
+                android:layout_alignParentRight="true"
+                android:layout_centerVertical="true" />
+        </RelativeLayout>
+
+        <RelativeLayout style="@style/separator_horizontal_layout" >
+            <View style="@style/separator_horizontal" />
+            <TextView
+                style="@style/separator_horizontal_headline"
+                android:text="@string/search_geo" />
+        </RelativeLayout>
+
+        <com.google.android.material.textfield.TextInputLayout style="@style/textinput_autocompletetextview">
+            <AutoCompleteTextView
+                android:id="@+id/geocode"
+                style="@style/textinput_embedded"
+                android:hint="@string/search_geo"
+                android:imeOptions="actionGo"
+                android:inputType="textVisiblePassword"
+                android:text="GC"
+                tools:ignore="HardcodedText" />
+        </com.google.android.material.textfield.TextInputLayout>
+
+        <Button
+            android:id="@+id/display_geocode"
+            style="@style/button_full"
+            android:text="@string/search_geo_button" />
+        <!-- ** -->
+
+        <RelativeLayout style="@style/separator_horizontal_layout" >
+            <View style="@style/separator_horizontal" />
+            <TextView
+                style="@style/separator_horizontal_headline"
                 android:text="@string/search_coordinates" />
         </RelativeLayout>
 
@@ -61,30 +116,6 @@
             <View style="@style/separator_horizontal" />
             <TextView
                 style="@style/separator_horizontal_headline"
-                android:text="@string/search_geo" />
-        </RelativeLayout>
-
-        <com.google.android.material.textfield.TextInputLayout style="@style/textinput_autocompletetextview">
-            <AutoCompleteTextView
-                android:id="@+id/geocode"
-                style="@style/textinput_embedded"
-                android:hint="@string/search_geo"
-                android:imeOptions="actionGo"
-                android:inputType="textVisiblePassword"
-                android:text="GC"
-                tools:ignore="HardcodedText" />
-        </com.google.android.material.textfield.TextInputLayout>
-
-        <Button
-            android:id="@+id/display_geocode"
-            style="@style/button_full"
-            android:text="@string/search_geo_button" />
-        <!-- ** -->
-
-        <RelativeLayout style="@style/separator_horizontal_layout" >
-            <View style="@style/separator_horizontal" />
-            <TextView
-                style="@style/separator_horizontal_headline"
                 android:text="@string/search_kw" />
         </RelativeLayout>
 
@@ -122,38 +153,6 @@
             style="@style/button_full"
             android:text="@string/search_hbu_button" />
         <!-- ** -->
-
-        <RelativeLayout style="@style/separator_horizontal_layout" >
-            <View style="@style/separator_horizontal" />
-            <TextView
-                style="@style/separator_horizontal_headline"
-                android:text="@string/search_filter" />
-        </RelativeLayout>
-
-        <RelativeLayout
-            android:layout_height="wrap_content"
-            android:layout_width="match_parent">
-
-        <Button
-            android:id="@+id/search_filter"
-            style="@style/button_full"
-            android:text="@string/search_filter_button"
-            android:layout_alignParentLeft="true"
-            android:layout_toLeftOf="@+id/search_filter_info"
-            android:layout_centerVertical="true"
-            android:layout_width="wrap_content"/>
-
-            <ImageView
-                android:id="@+id/search_filter_info"
-                android:layout_width="36dp"
-                android:layout_height="36dp"
-                app:srcCompat="@drawable/settings_info"
-                android:layout_marginLeft="0dp"
-                android:layout_marginRight="10dp"
-                android:layout_alignParentRight="true"
-                android:layout_centerVertical="true" />
-
-        </RelativeLayout>
 
         <RelativeLayout style="@style/separator_horizontal_layout" >
             <View style="@style/separator_horizontal" />
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    >
+<resources xmlns:android="http://schemas.android.com/apk/res/android" >
 
     <!-- system definitions -->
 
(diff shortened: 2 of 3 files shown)


main/res/layout/search_activity.xml
```diff
@@ -17,68 +17,94 @@
...
         <com.google.android.material.textfield.TextInputLayout style="@style/textinput_autocompletetextview">
             <AutoCompleteTextView
-                android:id="@+id/address"
+                android:id="@+id/geocode"
                 style="@style/textinput_embedded"
-                android:hint="@string/search_address"
-                android:imeOptions="actionGo" />
+                android:hint="@string/search_geo"
+                android:imeOptions="actionGo"
+                android:inputType="textVisiblePassword"
```
Lineflyer, line 60: Knowing, that this inputType was in the code before, still the question: Why shall it use that inputType?

main/res/values/strings.xml
```diff
@@ -384,7 +385,7 @@
     <string name="caches_sort_favorites">Favorites</string>
     <string name="caches_sort_favorites_ratio">Favorites [%]</string>
     <string name="caches_sort_name">Name</string>
-    <string name="caches_sort_geocode">Geo Code</string>
+    <string name="caches_sort_geocode">Geocode</string>
```
Lineflyer, line 388: I believe "geo code" are two words in english?!

main/res/values/strings.xml
```diff
@@ -1362,7 +1363,7 @@
     <string name="cache_status_stored_morethantwentyfourhours">Before last twenty-four hours</string>
     <string name="cache_status_stored_morethansevendays">Before last seven days</string>
     <string name="cache_status_stored_morethanthirtydays">Before last thirty days</string>
-    <string name="cache_geocode">Geo code</string>
+    <string name="cache_geocode">Geocode</string>
```
Lineflyer, line 1366: see above

main/res/values/strings.xml
```diff
@@ -1568,22 +1569,22 @@
 
     <!-- search -->
     <string name="search_bar_hint">Search caches/trackables</string>
-    <string name="search_bar_desc">Caches (geo code, keyword), Trackables (TB code)</string>
-    <string name="search_coordinates">Coordinates</string>
-    <string name="search_coordinates_button">Search by coordinates</string>
+    <string name="search_bar_desc">Caches (geocode, keyword), Trackables (TB code)</string>
```
Lineflyer, line 1572: see above

main/res/values/strings.xml
```diff
@@ -1568,22 +1569,22 @@
...
+    <string name="search_coordinates">Coordinate</string>
```
Lineflyer, line 1573: Good catch. This is often done wrong.

main/res/values/strings.xml
bekuno: I found only this as single word, see https://en.m.wikipedia.org/wiki/Geocode

Maybe we should use "cache code" to distinguish to other geographical information?

Lineflyer: OK, then lets just use "geocode" as one word as you suggest. I wasn't aware of that.

main/res/layout/search_activity.xml
bekuno: Good question.
Originally (at least long before) it was declared as
~~~
android:inputType="textCapCharacters"
~~~
which sets all characters to uppercase.
With 7aff8e5de155af144a76663158365d02df1fb794 it was changed to 
~~~
android:inputType="textVisiblePassword"
~~~
Now after 42e195e20c8c0446887f6babe4334d6f0ccf907b we have the implementation using viewbindings and a programmatic change in capital letters
~~~
        binding.geocode.setFilters(new InputFilter[] { new InputFilter.AllCaps() });
        binding.trackable.setFilters(new InputFilter[] { new InputFilter.AllCaps() });
~~~

The result - the declaration in the xml is superfluous.

fm-sys:
Are you sure this should go to release? IMHO the changes are not that urgent and we need to find a good time to stop introducing even more stuff and finally go into production. OTOH, I don't know how much time is still needed to reach that state independent of this PR...

bekuno:
retest this please

Lineflyer:
> Are you sure this should go to release?

Good point. While its no big change, still its no bugfix and should therefore go to `master`.

bekuno:
> 
> 
> > Are you sure this should go to release?
> 
> Good point. While its no big change, still its no bugfix and should therefore go to `master`.

As it is no bugfix, it has some "new filter" related optimization and therefore should target to release.
The other are fixes for text errors and a change to "Latitude/Longitude as a two row button".

I can split this PR into 2 PRs (master and release), but I don't think it's worth the work.

main/res/layout/search_activity.xml
```diff
@@ -17,68 +17,93 @@
...
             <AutoCompleteTextView
-                android:id="@+id/address"
+                android:id="@+id/geocode"
                 style="@style/textinput_embedded"
-                android:hint="@string/search_address"
-                android:imeOptions="actionGo" />
+                android:hint="@string/search_geo"
+                android:imeOptions="actionGo"
+                android:text="GC"
+                tools:ignore="HardcodedText" />
```
fm-sys, line 61: You should to disable Autosuggestions for this text field

main/res/layout/search_activity.xml
```diff
@@ -167,8 +160,7 @@
                 android:id="@+id/trackable"
                 style="@style/textinput_embedded"
                 android:hint="@string/search_tb_hint"
-                android:imeOptions="actionGo"
-                android:inputType="textVisiblePassword" />
+                android:imeOptions="actionGo" />
```
fm-sys, line 163: Here as well.

I know textVisiblePassword was used for it before, and it is definitely still necessary. Even better use "textCapCharacters|textVisiblePassword". that will disable auto suggestions and start the keyboard in Capital Letters mode which is consistent as typed letters are converted to capital anyway.

fm-sys (changes requested):
Splitting up really doesn't make sense. So maybe `release` if it is necessary. But `textVisiblePassword` is still needed, otherwise you reintroduce bugs  which were already fixed!

main/res/layout/search_activity.xml
fm-sys: Alternatively, you could use `textNoSuggestions` instead of `textVisiblePassword`. But at least one of both.

bekuno: The `textCapCharacters` did not work on my device.
So I assume that the IME problems are not fixed by the keyboard software creators.
Therefore lets go back to `textVisiblePassword` and stay by the old implementation.

bekuno: See eg. discussion here: https://stackoverflow.com/questions/3789905/how-can-i-turn-off-suggestions-in-edittext .

bekuno:
The lines `android:inputType="textVisiblePassword"` are back.

I reverted squashing the two Coordinate-Buttons to one button.
This will target this to master later.

Additional a bug is fixed. The "Search by coordinate" sets Latitude/Longitude, but did not start search.

fm-sys (approved):
> Additional a bug is fixed. The "Search by coordinate" sets Latitude/Longitude, but did not start search.

AFAIK, this was the intended behavior. But I have no preference which one is better...

Lineflyer:
> As it is no bugfix, it has some "new filter" related optimization and therefore should target to release.
The other are fixes for text errors and a change to "Latitude/Longitude as a two row button".

The text fixes are also kind of a timely problem as these will trigger new translations after merge, which most probably needs several days for our contributors to complete.

> I can split this PR into 2 PRs (master and release), but I don't think it's worth the work.

If there is no bug explicitly fixed I would still prefer to have it on `master` **or** just wait with merging to `release`after we released our first version. That would allow a bit more time for translation and testing.
The alternative would be to merge now to `release` and wait some days for translation and testing....which would also be OK in general...but I guess we are all keen to see the release finally going out. I can live with all of that, except pushing it in and immediately releasing it.

Lineflyer (approved):
See comment.

bekuno:
I reduced the PR and removed the strings changes.
The string changes can go to master, as they are no new strings compared to the latest c:geo release.

Lineflyer:
Just for my understanding:
Besides the change of menu order in the xml files I do see a changed in `SearchActivity.java`. What is the background of this?

Someone with code insights shall decide if this is suitable for `release`.
The change of the menu item order is OK for me.

bekuno:
> 
> 
> Just for my understanding:
> Besides the change of menu order in the xml files I do see a changed in `SearchActivity.java`. What is the background of this?
> 
> Someone with code insights shall decide if this is suitable for `release`.
> The change of the menu item order is OK for me.

I had written it in short in the commit.
If you type the button "Search by coordinate" without an Latitude/Longitude, then Latitude/Longitude was filled with the current possition, but does not start search.
This is confusing, so with my change the search is started immediately.

fm-sys:
As written some comments before:


> AFAIK, this was the intended behavior. But I have no preference which one is better...

I have no problem with the old behavior. But I'm fine with the new new behavior as well. However, if we decide to directly start into search shouldn't we rather start the "nearby search" so that users know which coordinates where filled automatically?

Lineflyer:
> However, if we decide to directly start into search shouldn't we rather start the "nearby search" so that users know which coordinates where filled automatically?

Isn't it exactly the same anyway, because technically it should be the same request.

bekuno:
> 
> 
> > However, if we decide to directly start into search shouldn't we rather start the "nearby search" so that users know which coordinates where filled automatically?
> 
> Isn't it exactly the same anyway, because technically it should be the same request.

Correct.
The coordinates are filled at first in the corrdinate-boxes.
And then the search shows the coordinate in the search info line and the headline in the same was as it is with every given coordinate.

Screenshot_20210803-203347_cgeo

fm-sys:
> > However, if we decide to directly start into search shouldn't we rather start the "nearby search" so that users know which coordinates where filled automatically?
> 
> Isn't it exactly the same anyway, because technically it should be the same request.

There is one difference: the title bar either contains "nearly" or the coords... 

Screenshot_20210803_224826_cgeo geocaching


 Screenshot_20210803-203347_cgeo

Issue #11347, opened 2021-07-27
Advanced Filter - "Origin" - add all (active) entries automatic or/and add button "add all"


bekuno:
### Tell us your idea!

On most filter critera all possible entries are shown.
But for Origin we have an additional step to "Add more Items".

I think we can make the live easier here.
Possibilities:
1) show all origin sources here instead to have a selection dialog.
2) show initial all active sources
3) add a "Add all" button or list entry

My favorites is 1) followed by 2).

fm-sys:
Problem is, that the same code base is used for the "contained in list" filter as well (and users can have _many_ lists...)

bekuno:
> 
> 
> Problem is, that the same code base is used for the "contained in list" filter as well (and users can have _many_ lists...)

Why should that be a problem?
You hit "Add Filter Criteris": "Origin", annd then the list with all origin enties is shown.
(Maybe preselect with all active connectors)

Lineflyer:
> Why should that be a problem?

Well, I have >300 lists and would appreciate to not be flooded by all in this dialog (for contained in list filter). Its anyway dynamically filled with existing values from the filtered base AFAICS, which is just perfect.

bekuno:
I mean this point in the filter:
![Screenshot_20210727-230946_cgeo](https://user-images.githubusercontent.com/32555742/127228075-69d5a36e-395e-407c-aaa6-b7b4df858acc.jpg)

If I call "Add more items" then I only see connectors, no list names:

![Screenshot_20210727-231009_cgeo](https://user-images.githubusercontent.com/32555742/127228093-b5cc7921-f780-4b63-838a-0c8577161261.jpg)

What is the other use case?

fm-sys:
I think we should rather add a "select all" checkbox to the top of the "select items" dialog (similar to our type filter)

moving-bits:
@bekuno 
See the "stored lists" filter for the other use-case

murggel:
> I think we should rather add a "select all" checkbox to the top of the "select items" dialog (similar to our type filter)

And add a "select active ones" checkbox.

murggel:
> You hit "Add Filter Criteris": "Origin", annd then the list with all origin enties is shown.
> (Maybe preselect with all active connectors)

Would it be possible to have the active ones already added to the view, so user does not have to click the add button for the active ones. Question would be, should they be activated or deactivated per default

eddiemuc:
[...]

PR #11376, opened 2021-07-31
rel to #11347: add active services to origin filter automatically

eddiemuc:
rel to #11347: add active services to origin filter automatically

This PR will:
* add origins from active services to the list of selectable "origins" in origin filter automatically
* adds an "add all items" button to checkbox filters with "add more items" button (currently "origin" and "stored lists").

This PR will **not** completely solve #11347: a possibility to select all/none items in "add more items" dialog is not included. This feature deserves its own PR since it is not filter-specific.


eddiemuc:
image

PR #11380, opened 2021-07-31
LCApi - resolve some lint/SonarLint problem and optimizations

bekuno:
### funtional change
- check also for CONSUMER_KEY.isEmpty() to not call the LC endpoint

### lint and other fixes
- integrate DATE_FORMAT in the using method because of using Locale.getDefault()
- externalize some multi used strings
- optimize some log strings
- added/corrected some Nullable/NonNull declaration
- optimized importCacheFromJSON() method
- other minor changes

### DIFF
--- a/main/src/cgeo/geocaching/connector/lc/LCApi.java
+++ b/main/src/cgeo/geocaching/connector/lc/LCApi.java
@@ -45,12 +45,15 @@ import org.apache.commons.lang3.StringUtils;
 
 final class LCApi {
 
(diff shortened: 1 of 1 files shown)


baiti (approved):
Thanks much for jumping in and fixing my poorly written Java. Most of what you touched, I had put in via copy / paste from other sources, mainly the EC connector.

bekuno:
> 
> 
> Thanks much for jumping in and fixing my poorly written Java. Most of what you touched, I had put in via copy / paste from other sources, mainly the EC connector, except the functional change you made, that was an oversight of my own.

Your Java isn't so poor.
I have done much of the changes with help of AS and plugins - lint, SonarLint.

PR #11385, opened 2021-08-01
rel to #11204: make all fragment classes public (try to fix the problem by this)

eddiemuc:
rel to #11204: make all fragment classes public (try to fix the problem by this)


moving-bits:
Let's give it a try...

PR #11392, opened 2021-08-01
add 'set as start' to route sort

moving-bits:
## Description
I frequently find myself in a situation that I've planned a tour in a certain order some time ago, let's say in order "cache 1 - cache 10", but on starting the tour decide to start with "cache 3", continue to "cache 10", and do "cache 1" and "cache 2" only then. Reordering the tour is possible using the route sort activity, but can become quite cumbersome depending on route length and new entry point.

This PR adds a long tap action to the route sort activity, which will bring up a info message, and resorts the tour on confirmation.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2215,6 +2215,8 @@
         <item quantity="other">Route loaded with %d route points</item>
     </plurals>
     <string name="center_on_route">Center on route</string>
+    <string name="individual_route_set_as_start_title">Set as route start</string>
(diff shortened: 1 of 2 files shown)


fm-sys:
What about using a PopupMenu instead, like we have for cache/Waypoint/Logs list? That would be consistent and can easily be extended in the future... 

Screenshot_20210801_230108_cgeo geocaching_copy_249x248

moving-bits:
There should still be a confirmation dialog, IMHO, and as it's only one action so far, I've chosen not to use a popup dialog for now.

If we add more item-specfic actions later, we should use a popupmenu for sure.

Issue #11184, opened 2021-07-11
Cannot update offline map files


Lineflyer:
A beta users reports, that c:geo crashes when he follows the update notification for the offline map.

## c:geo version 

2021.07.10-RC

## System information 

```
## System information

c:geo version: 2021.07.10-RC

Device:
-------
- Device type: SM-G960F (starlteeea, samsung)
- Available processors: 8
- Android version: 10
- Android build: QP1A.190711.020.G960FXXUFFUE1
- Screen resolution: 1080x2076px (360x668dp)
- Sailfish OS detected: false
- Google Play services: enabled - 21.21.16 (120408-378233385)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Hide caches: foundarchived
- Hide waypoints: visited
- Set language: de
- System date format: dd.MM.yy
- Debug mode active: no
- Live map mode: false
- OSM multi-threading: true / threads: 4
- Global filter: display all caches
- Last backup: 11. Juli, 13:47
- Routing mode: Off
- Settings: v5, Count:158
- Map: Germany (Offline)
- Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:3735-3561:cgeo/Maps/germany.map
- Atts: Map data (c) OpenStreetMap contributors
- Theme: fzk-outdoor-soft-v5.zip:fzk-outdoor-soft-v5/fzk-outdoor-soft-v5.xml

Services:
-------
- Geocaching sites enabled:
geocaching.com: Logged in (Anmeldung OK) / PREMIUM
Geocaching.com Adventure Lab
- Geocaching.com date format: MM/dd/yyyy
- Routing: external / BRouter installed: true
- Installed c:geo plugins: contacts

Permissions & paths:
-------
- Fine location permission: granted
- Write external storage permission: granted
- System internal c:geo dir: /data/user/0/cgeo.geocaching (15,9 GB free) v2 internal isDir(8 entries)
- Legacy User storage c:geo dir: /storage/emulated/0/cgeo (15,9 GB free) v2 external non-removable isDir(7 entries)
[...]

PR #11393, opened 2021-08-01
set downloader destination to ExternalFilesDir (rel. to #11184)

moving-bits:
## Description
- related to #11184
- implements solution suggested in https://stackoverflow.com/questions/66050509/downloadmanager-throw-android-os-networkonmainthreadexception (changing download folder)

(works at least in API 30 emulator)


Lineflyer:
IMHO this needs more testing time....not sure if this really works on all devices out here.

What about we target this to `master` and first look into user reports with downloader problems before releasing it?

moving-bits:
sure, have rebased it to `master`

PR #11395, opened 2021-08-02
fix added waypoint not show immediately (fix #11201)

### LINKED ISSUE #11201: "Add current location" waypoints not shown
## Bug description 

Today I ran in a situation where the waypoint tab failed to update after adding waypoints. Found a way to reproduce

- Open a cache and swipe to waypoint tab (or use the Go to cache)
- Create a waypoint, enter some coordinates and save
- The waypoint is now shown in the list
- Tap on "Add current location" to create another waypoint 

The waypoint is added (total amount rising on bottom bar) but the newly created waypoint is not shown
You can add another one by "Add current location" which is also not shown, and so on.

The waypoints are shown if you leave and reenter the screen.

View needs to update to show the new waypoints immediately

## c:geo version 

2021.07.11-NB

## Reproducible 

Yes

moving-bits:
## Description
Makes viewpager fragment update on `notifyDatasetChanged` even if it's the currently active fragment


moving-bits:
retest this please

PR #11405, opened 2021-08-03
fix #11351: make lc connector detect premium after reconnect

### LINKED ISSUE #11351: Premium not detected in first session
### Describe your problem!

While testing several development builds and the current RC I noticed, that the detection as being a premium member does not seem to work immediately after first authorization / in first session.

Did not check whether its clearly reproducible, but I have seen this quite often (as I wanted to check ALC implementation and non was loaded).

### How to reproduce?

- Install c:geo
- Run wizard
- Authenticate to gc.com as PM from the wizard
- Complete the wizard

- I cannot see any ALC caches on the map
- Looking into setting I see the connector settings are greyed out
- Did not check system information though

PM status should immediately be detected/used

### Reproducible

Unclear

### c:geo Version

2021.07.28-RC

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #11351: make lc connector detect premium after reconnect

basic problem was that serverparameters were not reset in GCConnector after logout - so they stayed the same even after re-login. This incluided the premium state.

But this was not enough, some fiddling inside Settings for LC connector was also necessary to get this running...

@baiti : I would appreciate a review for this PR

### DIFF
--- a/main/src/cgeo/geocaching/settings/SettingsActivity.java
+++ b/main/src/cgeo/geocaching/settings/SettingsActivity.java
@@ -309,8 +309,22 @@ public class SettingsActivity extends PreferenceActivity implements Preference.O
 
     private void initLCServicePreference(final boolean gcConnectorActive) {
         final boolean isActiveGCPM = gcConnectorActive && Settings.isGCPremiumMember();
-        getPreference(R.string.preference_screen_lc).setSummary(getServiceSummary(Settings.isLCConnectorActive() && isActiveGCPM));
-        getPreference(R.string.pref_connectorLCActive).setEnabled(isActiveGCPM);
+        getPreference(R.string.preference_screen_lc).setSummary(getLcServiceSummary(Settings.isLCConnectorActive(), gcConnectorActive));
+        if (isActiveGCPM) {
+            getPreference(R.string.pref_connectorLCActive).setEnabled(true);
+        }
+    }
+
+    private String getLcServiceSummary(final boolean lcConnectorActive, final boolean gcConnectorActive) {
+        if (!lcConnectorActive) {
+            return StringUtils.EMPTY;
+        }
+
+        //lc service is set to active by user. Check whether it can actually be actived due to GC conditions
+        final int lcStatusTextId = gcConnectorActive && Settings.isGCPremiumMember() ?
+            R.string.settings_service_active : R.string.settings_service_active_unavailable;
+
+        return CgeoApplication.getInstance().getString(lcStatusTextId);
     }
 
     private void setWebsite(final int preferenceKey, final String urlOrHost) {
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/settings/SettingsActivity.java
```diff
@@ -309,8 +309,18 @@ private void initServicePreferences() {
...
-        getPreference(R.string.preference_screen_lc).setSummary(getServiceSummary(Settings.isLCConnectorActive() && isActiveGCPM));
-        getPreference(R.string.pref_connectorLCActive).setEnabled(isActiveGCPM);
+        getPreference(R.string.preference_screen_lc).setSummary(getLcServiceSummary(Settings.isLCConnectorActive(), gcConnectorActive));
+        if (isActiveGCPM) {
+            getPreference(R.string.pref_connectorLCActive).setEnabled(true);
+        }
+    }
+
+    private String getLcServiceSummary(final boolean lcConnectorActive, final boolean gcConnectorActive) {
+        return !lcConnectorActive ? StringUtils.EMPTY :
```
baiti, line 319: I find this statement overwhelmingly complex but this might just be the limitation of my brain built-in Java parser.

baiti (approved):
Except that comment wrt to complexity of the return statement the code looks good to me. I did, however, not test it as I currently do not know how to checkout your branch into my local repo and build an executable.

Lineflyer:
> I did, however, not test it as I currently do not know how to checkout your branch into my local repo and build an executable.

Quick lesson for that on command line:
git fetch upstream refs/pull/11405/head:nameforthebranch
get checkout nameforthebranch

Alternatively you can use Android Studios GIT integration: Menu - VCS - Git - View Pull Requests (thats needs authentication to github if not done before). 
Here you can look into the PRs and there should be a button "checkout"

main/src/cgeo/geocaching/settings/SettingsActivity.java
eddiemuc: Are you referring explicitly to the method returning the Lc connector state summary?

Basically it ads to the former two-way-state (active or not active) a third state "active/disabled" (in case LC connector is set to active but can't be used because the logged in user is not premium)

baiti: > Are you referring explicitly to the method returning the Lc connector state summary?
> 
> Basically it ads to the former two-way-state (active or not active) a third state "active/disabled" (in case LC connector is set to active but can't be used because the logged in user is not premium)

No, it is the multiline return statement in line 319 that I have a hard time to parse. It is the nested use of the ternary operator. In order to truely understand the logic behind it, I'd have to unravel it into if / then / else statements and using simple expressions / variables ... which I haven't done in the context of this review, instead, I just assume that it works the way you intend ;)

baiti:
> > I did, however, not test it as I currently do not know how to checkout your branch into my local repo and build an executable.
> 
> Quick lesson for that on command line (in fact checking out the PR rather than the original branch):
> git fetch upstream refs/pull/11405/head:nameforthebranch
> get checkout nameforthebranch
> 
> Alternatively you can use Android Studios GIT integration: Menu - VCS - Git - View Pull Requests (thats needs authentication to github if not done before).
> Here you can look into the PRs and there should be a button "checkout"

Thanks much for the lesson. Over the last months we have accumulated many such lessons in the various PRs of mine. Unfortrunately it is virtually impossible to search for these lessons. It really seems I have to bite the bullet and copy these things over into some private notebook... This is a project on its own.

main/src/cgeo/geocaching/settings/SettingsActivity.java
eddiemuc: @baiti That's a fair assessmemt and it clearly shows that my code was too complex. I cleaned it up, may you take a second look at the new version

PR #11407, opened 2021-08-03
use shorter key for D/T-config (fix #11362)

### LINKED ISSUE #11362: D/T filter setting - make it easier to read
> 
> 
> > Is there also a way to not double the D/T state in the system information -> Filter setting?
> 
> The info is not really doubled. The first difficulty represents the min value, the second one the max value. In current example, both min and max are "indefinite" which is represented by "-"
> 
> The representation could maybe be optimized. But I don't see a high priority here, this is just a technical view and not something the user will see (except in debug support situations).

Understandable.
Then let's put this beautification to prio low.

_Originally posted by @bekuno in https://github.com/cgeo/cgeo/issues/11361#issuecomment-889483120_

murggel:
## Description

use "d" and "t" as key instead of "difficulty" and "terrain" to shorten the config-output.

## Related issues

fix #11362 

## Additional context

if we want to use a shorter version, we should introduce it before releasing to avoid migrating effort.

Or we refactor the combined filter anyway and this here is not necessary

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/DifficultyAndTerrainGeocacheFilter.java
@@ -11,8 +11,8 @@ public class DifficultyAndTerrainGeocacheFilter extends BaseGeocacheFilter {
     public DifficultyGeocacheFilter difficultyGeocacheFilter;
     public TerrainGeocacheFilter terrainGeocacheFilter;
 
-    private static final String CONFIG_DIFFICULTY = "difficulty";
-    private static final String CONFIG_TERRAIN = "terrain";
+    private static final String CONFIG_KEY_DIFFICULTY = "d";
+    private static final String CONFIG_KEY_TERRAIN = "t";
 
 
     public DifficultyAndTerrainGeocacheFilter() {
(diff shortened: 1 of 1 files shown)


fm-sys:
If this config is not user displayed, we IMHO shouldn't spend any effort to optimize the configuration string. BTW, won't this change break stored filters for beta users?

While it's IMHO OK, if beta users need to setup things multiple times or stuff like that, still such breaking changes should be avoided if they are not absolutely necessary...

murggel:
> 
> If this config is not user displayed, we IMHO shouldn't spend any effort to optimize the configuration string.

This is only displayed to user in the cgeo-settings which is send to support.

> BTW, won't this change break stored filters for beta users?
> While it's IMHO OK, if beta users need to setup things multiple times or stuff like that, still such breaking changes should be avoided if they are not absolutely necessary...

That was also my concern, but then I would close the related issue #11362

fm-sys:
> That was also my concern, but then I would close the related issue #11362 

Would be fine for me.

I just added a comment how we could optimize the user visible side, so either we rename the issue or we move the suggestion into a new one...

murggel:
> 
> 
> > That was also my concern, but then I would close the related issue #11362
> 
> Would be fine for me.
> 
> I just added a comment how we could optimize the user visible side, so either we rename the issue or we move the suggestion into a new one...

Just saw your comment there, so I would rename the issue...

eddiemuc:
To  be honest, I would just merge this PR. Although changing config this way is prio low, this will probably haunt us worse when we want to implement new feature requests for the area. 

While right now we still have a "free shot" for these things (we agreed that no migration routines are necessary for beta users, see discussion for migration of the global exclude flags) we loose this free shot once the stuff is in production. I vote for merging this PR. The work is already done anyway.

eddiemuc (approved):

fm-sys:
> this will probably haunt us worse when we want to implement new feature requests for the area.

What do you mean? Isn't it codewise exactly the same if "difficulty" is stored instead of "D" in the config string? 

But if we want to merge it anyway:
What will happen if a user loads a saved "wrong configuration"? 

- Will this result in a crash? -> merging this PR into beta would be totally unacceptable 
- Will the old filter just be discarded? -> merge would be ok although annoying for beta users

murggel:
> > this will probably haunt us worse when we want to implement new feature requests for the area.
> 
> What do you mean? Isn't it codewise exactly the same if "difficulty" is stored instead of "D" in the config string? 
> 
> But if we want to merge it anyway:
> What will happen if a user loads a saved "wrong configuration"? 
> 
> - Will this result in a crash? -> merging this PR into beta would be totally unacceptable 
> - Will the old filter just be discarded? -> merge would be ok although annoying for beta users
> 
> 
> 

Didn't crashed for me while implementing, but I can do another test in emulator.

eddiemuc:
> > this will probably haunt us worse when we want to implement new feature requests for the area.
> 
> What do you mean? Isn't it codewise exactly the same if "difficulty" is stored instead of "D" in the config string?

the more options come in, the more complex and hard-to-debug the implementation gets. We have a free-shot to fix this now, won't have it later (when this is out for all users, all technical config changes/extensions/,.. need to be backed by migration routines). Let's not make an "Elefant" out of a "Mücke". I would merge this and be done with it.

> 
> But if we want to merge it anyway:
> What will happen if a user loads a saved "wrong configuration"?
> 
> * Will this result in a crash? -> merging this PR into beta would be totally unacceptable
> * Will the old filter just be discarded? -> merge would be ok although annoying for beta users#

The second thing - except if there are implementation flaws (where it would be good to discover it in beta anyway)

fm-sys:
> I would merge this and be done with it.

Feel free to do so if you think it makes sense...

PR #11418, opened 2021-08-06
add user time zone and reformat SystemInformation

fm-sys:
The program settings section was quite messed up, so I added a separate "Filters" section and moved all map related stuff directly above the map source section (Routing mode / Live map mode / OSM multi-threading).

Furthermore, "Timezone" is added.

|before |now |
-|-
Screenshot_20210806_154808_cgeo geocaching|grafik|

### DIFF
--- a/main/src/cgeo/geocaching/utils/CalendarUtils.java
+++ b/main/src/cgeo/geocaching/utils/CalendarUtils.java
@@ -162,4 +162,11 @@ public final class CalendarUtils {
 
         return new ImmutablePair<>(date.getTimeInMillis(), date.getTimeInMillis() + 86400000);
     }
+
+    /**
+     * Generate a time zone string for the users system time config
+     */
+    public static String getUserTimeZoneString() {
+        return new SimpleDateFormat("z", Locale.US).format(new Date());
+    }
 }
--- a/main/src/cgeo/geocaching/utils/SystemInformation.java
+++ b/main/src/cgeo/geocaching/utils/SystemInformation.java
@@ -94,18 +94,24 @@ public final class SystemInformation {
             .append("\n")
             .append("\nProgram settings:")
             .append("\n-------");
-            appendFilters(body);
+            appendSettings(body);
             body
-            .append("\n- Hide waypoints: ").append(hideWaypoints.isEmpty() ? "-" : hideWaypoints)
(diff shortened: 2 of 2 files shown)


main/src/cgeo/geocaching/utils/CalendarUtils.java
```diff
@@ -162,4 +164,13 @@ public static long parseDayMonthYearUS(final String date) {
...
         return new ImmutablePair<>(date.getTimeInMillis(), date.getTimeInMillis() + 86400000);
     }
+
+    /**
+     * Generate a time zone string for the users system time config
+     */
+    public static String getUserTimeZoneString() {
+        final TimeZone timeZone = TimeZone.getDefault();
+        final int timeZoneGMTOffset = timeZone.getRawOffset();
+        return timeZone.getDisplayName(false, TimeZone.SHORT) + " / " + (timeZoneGMTOffset < 0 ? "-" : "+") + TimeUnit.HOURS.convert(timeZoneGMTOffset, TimeUnit.MILLISECONDS);
```
bekuno, line 174: Please change this to output ".../UTC+x".
This is easier to understand later.

Your output says MEZ. Should't it be MESZ, or better non localized MEST?

fm-sys: Here we go. Complex requests result in a simple solution... Codewise much cleaner and it will print "GMT+02:00" in this example which is the US ISO standard.

bekuno: Hmm, now it is too simple .
Why not stay with both parts and use for the first one `getDisplayName(boolean daylight, int style, Locale locale)` with `daylight = true` and `locale = Locale.US` ?

fm-sys: > with `daylight = true`

This will then result in that daylight is always applied, independent of whether it is used in that time zone at all. 

So an additional check would be necessary, which however IMHO get much to complex. 

But why should we ever need both formats in the system information?

moving-bits:
@fm-sys 
I noticed that you've assigned this PR to yourself just about an hour ago. Anything upcoming here, or can I merge?

fm-sys:
> Anything upcoming here, or can I merge?

I had just batch-assigned all PR's to me I had created. But as long as it's not set to draft / WIP label assigned you can always feel free to merge my PR's.

PR #11421, opened 2021-08-06
detect special case that listing background image is the only spoiler

fm-sys:
Is there is only the background but no further spoilers, the previous text "spoiler" was misleading... Now, this case is handled specially:

grafik

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1748,6 +1748,13 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                     }
                     ImagesActivity.startActivity(activity, cache.getGeocode(), cache.getSpoilers());
                 });
+
+                // if there is only a listing background image without other additional pictures, change the text to better explain the content.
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1748,6 +1748,13 @@ public void setContent() {
                     }
                     ImagesActivity.startActivity(activity, cache.getGeocode(), cache.getSpoilers());
                 });
+
+                // if there is only a listing background image without other additional pictures, change the text to better explain the content.
+                if (cache.getSpoilers().size() == 1 && getString(R.string.cache_image_background).equals(cache.getSpoilers().get(0).title)) {
```
moving-bits, line 1753: not sure, but will we retrieve the cache page as a localized version, so that we can compare against a localized `R.string.cache_image_background`? Or will it always be an English string we have to look for?

fm-sys: we explicitly set this string resource (localized) as title when parsing the background. So as long as the user doesn't change the device setting, this check will work.

moving-bits: Ah, thanks, didn't know that - always learning something new while working on the code...

PR #11425, opened 2021-08-06
fix #11261: checkstyle: exclude check for parenthesis around logical and

### LINKED ISSUE #11261: Codacy: "Unnecessary parentheses around expression" warnings are too strict
Recently, Codacy starts complaining about "unnecessary parentheses" used in boolean expressions, Things like e.g. shown in following screenshot:

image

The underlying rule is `Checkstyle_UnnecessaryParentheses`:

image

I agree that this is in general a useful rules, and I also agree that in the given examples the parenthesis shall be removed.

But for boolean expressions like the above I find the parentheses extremely helpful for understanding the logic behind them - although they are maybe  not be strictly necessary because of operator precedence?

Would it be ok to "lighten" that rule?

And on a second note: if yes, how can I do that??? AFAICS Codacy uses our checkstyle rule file, but Jenkins does not complain about these things...

A third thing: I miss edit rights for Codacy, it always tells me I have to be "admin" to do something. Is there any way to change this?

eddiemuc:
fix #11261: checkstyle: exclude check for parenthesis around logical and

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -121,6 +121,8 @@
         <module name="SuppressWarningsHolder" />
         <module name="UnnecessaryParentheses">
             <property name="severity" value="info" />
+            <!-- put this check on all tokens except: LAND (Logical AND) -->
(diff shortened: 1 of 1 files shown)


eddiemuc:
Hm, this PR does not reduce the amount of false positives as much as I hoped for. There are also parenthesis around e.g. EQUAL/NOT_EQUAL-expressions or `instanceof` expressions or >=/<= expressions which frankly I find quite useful

To be honest, I really start to question this rule as a whole. What about just removing it from the config?

bekuno:
> To be honest, I really start to question this rule as a whole. What about just removing it from the config?

see #11261

bekuno:
DNM - will be replaced by #11465

moving-bits:
superseded by #11465, therefore closing here

PR #11427, opened 2021-08-07
remove reliableLatLon flag in Geocache model

fm-sys:
Remove unused code to simplify the migration to material design SVG icons (see #11139)

TODO: we should also remove the reliableLatLon  colum in our DB


fm-sys:
> TODO: we should also remove the reliableLatLon colum in our DB

While removing the reliableLatLon colum I noticed, that there are _many_ other unused columns. Or I reverted the change and annotated all unused columns, which should be removed all together later on (but not in this PR)

Issue #11411, opened 2021-08-04
Date picker returns wrong date for logging


Lineflyer:
### Describe your problem!

Two users independently reported the following bug for version 2021.07.30-RC:

When I select a logdate which is not today but some days in the past, after closing the date picker the shown date is one day earlier than selected.
It does seemingly never happen if the date is today but for all dates in the past.

---

I myself cannot reproduce it.
Maybe an issue depending on locale and/or timezones?

### How to reproduce?

- Start logging a cache
- Tap on the date to open the date picker
- Set it e.g. to 15th of July 2021
- Select OK to close the date picker

### Actual result after these steps?

Log date field in logging activity shows 14th of July 2021

### Expected result after these steps?

Log date field in logging activity shows 15th of July 2021

### Reproducible

Unclear

### c:geo Version

2021.07.30-RC

### System information

```shell
Example for one user:

## System information

c:geo version: 2021.07.30-RC

Device:
-------
- Device type: SM-G991U (o1qsqw, samsung)
- Available processors: 8
- Android version: 11
- Android build: RP1A.200720.012.G991USQS4AUFD
- Screen resolution: 1080x2176px (360x725dp)
- Sailfish OS detected: false
- Google Play services: enabled - 21.26.20 (150408-387585197)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Filters:
    LIVE: Status:Exclude Archived ([:inconclusive=false:advanced=false]status:exclude_archived)
    OFFLINE: None ([:inconclusive=false:advanced=false])
- Hide waypoints: original
- Set language: en_US (system default)
- System date format: M/d/yy
- Debug mode active: no
- Live map mode: false
- OSM multi-threading: true / threads: 4
- Last backup: Aug 2, 4:14 PM
- Routing mode: Walk
- Settings: v8, Count:132
- Map: Google: Satellite
  - Id: cgeo.geocaching.maps.google.v2.GoogleMapProvider$GoogleSatelliteSource
  - Atts: none
  - Theme: none

Services:
-------
- Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / PREMIUM
   Geocaching.com Adventure Lab
- Geocaching.com date format: MM/dd/yyyy
[...]

PR #11431, opened 2021-08-07
Fix date picker returns wrong date for logging

fm-sys:
Here we go, a fix for #11411


eddiemuc (approved):
Short, clean and to the point. I like it :-)

fm-sys:
> Short, clean and to the point. I like it :-)

Thanks :) 

I tested it for +2 and -8 time zone, but hopefully the fix will work in all scenarios... 😅

---

Failing test unrelated, therefore...

retest this please

PR #11432, opened 2021-08-07
Resort status filter and quick settings (fix #11343)

### LINKED ISSUE #11343: Quick settings - reorder "Show Caches:" and use same wording in QS and Filter
### Tell us your idea!

I suggest moving the entry for "Logged offline" to the 2nd position. (between "Found" and "Own") and "Disabled" last.
This allows for a more logical deactivation.

grafik

Additionally, use then the result of this discussion also for the filter.

The wording in the filter GUI is "Owned" and "Has Offline Log".
I propose to use the same wording in filter and QS.

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

fm-sys:
see https://github.com/cgeo/cgeo/issues/11343#issuecomment-888134489


fm-sys:
retest this please

PR #11434, opened 2021-08-07
Filter for "hide offline log" considers only "found" offline logs

### LINKED ISSUE #11430: Only hide found offline logs when using filter for hide offline logs
### Tell us your idea!

I am often using offline logs and I am often quite late with logging.
I like the idea to hide offline logs with the new filter rules.
Only one thing, I think could be changed:
The filter only should work when a found log is in the drafts list, not with DNF or Info. That leads to sometimes overlook a DNF when I get to the cache location a second time

I would suggest to set a more granulur filter rule or change this to caches marked as found only to be hidden.
Best would be to hide if any found log is in the draft list somewhere

### Is there an alternative?

Hide offline logged caches where the latest state is set to found.

### Additional Information

_No response_

murggel:
## Description

consider in StatusGeocacheFilter only offline-logs which are found-logs 
* adapt filter-method with LogType.isFoundLog 
* adapt where-clause in addToSql with logType in (LogType.getFoundLogIds)

## Related issues

fix #11430

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
@@ -3,16 +3,20 @@ package cgeo.geocaching.filters.core;
 import cgeo.geocaching.R;
 import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.enumerations.WaypointType;
+import cgeo.geocaching.log.LogType;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.SqlBuilder;
 import cgeo.geocaching.ui.ImageParam;
+import cgeo.geocaching.utils.CollectionStream;
 import cgeo.geocaching.utils.EmojiUtils;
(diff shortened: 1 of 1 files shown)


murggel:
With merging the PR there is no possibility to filter for empty offline-logs, which are not found ones.

Please correct me if I am wrong...

E.g. logging a "schneller offline log" with type DNF in the field and filling it later - how can I find those caches?

eddiemuc (approved):
Looks good codewise!

murggel:
I missed to check, whether the connector has to adapted to the filtering as well 

@eddiemuc No changes in connectors necessary?

eddiemuc:
> I missed to check, whether the connector has to adapted to the filtering as well 
> 
> @eddiemuc No changes in connectors necessary?

Nope, connectors can by their nature not know about offline logs

Lineflyer:
This PR was reverted by 6721f5198

fm-sys:
Wait! Indeed it was too early to merge this PR. But did we just made the same mistake again while revering the code without a unanimous decision?

Before we touch anything further we should collect a written down pro-contra-list...

PR #11449, opened 2021-08-09
Fix #11448 - Enlarge compass info and add space for direction

### LINKED ISSUE #11448: Font size for distance/direction in compass was reduced
### Describe your problem!

Beta users report, that on the compass its now harder to read the distance/direction.

Indeed a comparison of old (left) and new(right) on same emulators show it has reduced size.
image

Will try to solve this myself...need practice.

### How to reproduce?

- Open cache details
- Start Navigation - Compass

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021.08.08-RC

### System information

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

The `heading primary` text size leads to the compass direction/distance info be smaller compared to the old release. Therefore use explicit size for these text fields.
before_after

## Related issues

#11448

## Additional context

Additionally fixes the cut off direction info if it has three digits:
cut off orientation

### DIFF
--- a/main/res/layout-land/compass_activity.xml
+++ b/main/res/layout-land/compass_activity.xml
@@ -79,7 +79,7 @@
                  <TextView
                      android:id="@+id/device_heading"
                      style="@style/single_degree_style"
-                     android:width="36dip"
+                     android:width="42dip"
                      android:layout_toRightOf="@+id/device_heading_label"/>
 
                  <TextView
@@ -118,7 +118,8 @@
             android:layout_gravity="left"
             android:layout_marginLeft="3dip"
             android:layout_below="@+id/device_information"
-            android:text="@null" />
+            android:text="@null"
+            android:textSize="@dimen/textSize_compassTargeting" />
 
         <TextView
             android:id="@+id/distance"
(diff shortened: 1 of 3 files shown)


Lineflyer:
@moving-bits Can you crosscheck this PR please?

moving-bits (approved):
Technically it's ok, and one could argue choosing fixed text sizes are ok since only a single page is affected. OTOH it's four TextViews in two different layout files, so for easier consistency you could add a `textSize_detailsCompass` (or something like this) in `dimens.xml` and reference this value using `@dimen/textSize_detailsCompass`.

Lineflyer:
Fair enough and again learned something new.
PR has been adapted accordingly.

bekuno:
Doing the same for 'device_heading' would be perfect.

fm-sys:
> Doing the same for 'device_heading' would be perfect.

You mean the `42dip`? Can be done of course, but IMHO is not a must have...

bekuno:
To be consistent, then it is a must have, imho.

PR #11456, opened 2021-08-10
Fix #11453 - Add all permutations of pathPrefix for supported URLs

### LINKED ISSUE #11453: c:geo not offered as target app depending on URL capitalization
### Describe your problem!

Reported by a user on support mail:

The user coincidentally found, that depending how a coord.info link is formatted c:geo is offered or not offered as target app.

I could reproduce this issue (by sending such link to myself via Signal):
https://coord.info/GC9DBN6  -> offers c:geo
https://coord.info/gc9dbn6  -> offers c:geo
https://coord.info/Gc9dbn6 -> does not offer c:geo

### How to reproduce?

- Send such links with different capitalization to yourself by mail or messenger
- Click the link and check the offered apps

The third link does not offer c:geo as target app

c:geo should be offered as target app

### Reproducible

Yes

### c:geo Version

2021.08.08-RC

### System information

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

Added all the possible capitalizations to the URLs we support, just in case someone uses strange links (e.g. manually created).

## Related issues

#11453

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -454,6 +454,8 @@
                 <data android:host="www.coord.info" />
                 <data android:pathPrefix="/GC" />
                 <data android:pathPrefix="/gc" />
+                <data android:pathPrefix="/Gc" />
+                <data android:pathPrefix="/gC" />
             </intent-filter>
 
             <!-- geocaching.com -->
@@ -471,6 +473,8 @@
                 <data android:pathPrefix="/seek/cache_details.aspx" />
                 <data android:pathPrefix="/geocache/GC" />
                 <data android:pathPrefix="/geocache/gc" />
+                <data android:pathPrefix="/geocache/Gc" />
+                <data android:pathPrefix="/geocache/gC" />
             </intent-filter>
 
             <!-- opencaching.CZ -->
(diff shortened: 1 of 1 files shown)


Lineflyer:
Just in case you ask, whether there is no other possibility. AFAICS its not possible to apply advanced logic (besides `*` wildcards) and its by default case-sensitive.

moving-bits:
Just for the record: While this may work for the connector prefix, the geocode itself may have letters as well, and there are still plenty ways for the user to get the capitalization wrong there... But at least possible auto-capitalization by a keyboard should be covered now :-)

fm-sys (approved):
> Just for the record: While this may work for the connector prefix, the geocode itself may have letters as well, and there are still plenty ways for the user to get the capitalization wrong there... 

AFAIK this will work without problems, as we always convert geocodes to upper case as soon as we receive them as intent. The issue was just about the intent filter being too strict. 

Therefore, LGTM 👍

moving-bits:
Ah, you're right - we are just talking about prefixes here. Sorry, my bad.

moving-bits:
@Lineflyer 
Shall we include this in our next release (meaning: target it to `release`)?
It's a bugfix, and the fix itself should not harm anything.

Lineflyer:
I am a friend of using the save/pessimistic way. 
As this is not a serious bug and we cant say 1000% that is has no side effects we did not think of, I would rather keep it on `master`.

PR #11468, opened 2021-08-14
fix lc/al connector prefix (fix #10911)

### LINKED ISSUE #10911: Lab Cache Connector (LCConnector) uses same cache prefix as TerraCachingConnector
I realized that the chosen cache prefix code for Lab Caches "LC" is already used by another connector: the TerraCachingConnector (uses TC, CC and LC). Filtering for lab caches will thus also match all TerraCachine-Caches starting with LC (most likely not vice versa since TerraCaching uses a very restricted cache pattern: )

I don't really think this will be a huge problem in practice, but would it be better to choose another prefix for lab caches so they don't clash?

Fopr reference, this is the geocode pattern used by TerraCachingConnector: `(TC|CC|LC)[0-9A-Z]{1,4}`
(means: everything starting with TC, CC or LC with 1-4 letters/digits after)

_Originally posted by @eddiemuc in https://github.com/cgeo/cgeo/issues/10880#issuecomment-860040362_

moving-bits:
## Description
- renames LC prefix to AL prefix for all stored ALC caches (including all references)
- renames LC connector to AL (including package name / folder)

## Related issues
- fixes #10911

## Additional info
I've added a wiki page which documents the connectors c:geo supports and the geocode prefixes they use: https://github.com/cgeo/cgeo/wiki/Supported-geocaching-platforms

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -53,7 +53,7 @@
     <string translatable="false" name="pref_gc_avatar">gc_avatar</string>
     <string translatable="false" name="pref_gcvote_avatar">gcvote_avatar</string>
     <string translatable="false" name="pref_connectorECActive">connectorECActive</string>
-    <string translatable="false" name="pref_connectorLCActive">connectorLCActive</string>
+    <string translatable="false" name="pref_connectorALActive">connectorLCActive</string><!-- variable stays with "LC" for compatibility reasons -->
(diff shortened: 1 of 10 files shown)


eddiemuc:
LGTM. Let's hope we have no changes to do for this connector in release any more -  the class renaming would hit us on merging release back to master every time. Still this change is necessary, so let's merge.

PR #11471, opened 2021-08-15
remove old viewpagerindicator

moving-bits:
## Description
Since our switch to the new material tabbed viewpager2 the sources for the old viewpager are no longer needed

This PR removes the unnecessary source files as well as the proguard configuration for the viewpagerindicator


bekuno (approved):
Perfect.
This was also an point on my TODO list.
Thank you for shortening it.

PR #11472, opened 2021-08-15
remove duplicated 'activate' strings

moving-bits:
## Description
Nearly every connector has its own string "Activate" as label for settings. It's sufficient to have this string only once in our resources, therefore let's remove the duplicates.


moving-bits:
retest this please

PR #11482, opened 2021-08-16
Fix #11477 - Use more common english abbreviation for Traditional

### LINKED ISSUE #11477: Abbreviate "Traditional Caches" as "Trad" instead of "Tradi"
### Describe your problem!

From support mail:

Thanks for working so hard on the new version. It looks great, although I've noticed that you've abbreviated "Traditional" to "Tradi". This might be how it's done in Germany, but for the English version of the app, please can it be abbreviated to "Trad"? Tradi just sounds so wrong to English ears!

### How to reproduce?

Filter for "Cache Type"

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021.08.15

### System information

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

According to native speakers the abbreviation "Tradi"/"Tradis" is not used in English, but they commonly use "Trad"/"Trads".

## Related issues

#11477

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -44,7 +44,7 @@
     <string name="unknown">Unknown Type</string>
 
     <string name="all_types_short">All</string>
-    <string name="traditional_short">Tradi</string>
+    <string name="traditional_short">Trad</string>
     <string name="multi_short">Multi</string>
     <string name="mystery_short">Mystery</string>
     <string name="letterbox_short">Letterbox</string>


Surfoo:
Your PR has impacted all translations, it added this string but it was already translated.

Lineflyer:
That is a normal side effect. Anyway existing translations are still in by now until we merge in new translations from Crowdin. 
Even if we would have defined another string id, it would not be better.

PR #11496, opened 2021-08-18
fix #11488: make preimum caches available for basic members plus distance heuristic

### LINKED ISSUE #11488: Nearby search with filter "Premium = yes" gives empty list (at least for base members)
### Describe your problem!

With the introduction of the new filter and/or the new nearby search it is not possible to see PMO caches in the nearby list at least for base members.

Even for base members the information that a PMO cache is nearby might be interesting when caching together with premium members and to be able to log from there.

### How to reproduce?

Log in as base member.
Go to Nearby list, open the filter and select "Yes" for Premium caches.
List is empty even if PMO caches are around.

List is empty.

List should contain nearby PMO caches if Premium filter is not set to "No".

### Reproducible

Yes

### c:geo Version

2021.08.17-NB-240646f

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #11488: make preimum caches available for basic members plus distance heuristic

### DIFF
--- a/main/res/layout/cacheslist_item.xml
+++ b/main/res/layout/cacheslist_item.xml
@@ -90,13 +90,12 @@
         android:layout_marginTop="1dip"
         android:layout_gravity="right|center_vertical">
 
-        <view
+        <cgeo.geocaching.ui.DistanceView
             android:id="@+id/distance"
             android:layout_width="78dip"
             android:layout_height="wrap_content"
             android:layout_centerHorizontal="true"
             android:layout_gravity="center"
-            class="cgeo.geocaching.ui.DistanceView"
             android:ellipsize="end"
             android:gravity="center"
             android:lines="1"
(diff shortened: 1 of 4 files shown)


Lineflyer:
Question from my side (not having tested this PR yet):
Are you using again the approximate indication we used to have (so those caches will be shown with the `~` in front of the distance)?

eddiemuc:
> Question from my side (not having tested this PR yet):
> Are you using again the approximate indication we used to have (so those caches will be shown with the `~` in front of the distance)?

Yes. Was this disabled/discouraged before? (Because it was still in the code, I just fill it with my heuristic now instead of the parsed info from web site search result)

eddiemuc:
(unfortunately I am unable to provide screenshots currently)

Lineflyer:
> If the setting above IS checked, then the zoom level and shown area is adjusted such that all waypoints and the cache itself are visible (the cache is most likely not centered in this case)

It should be used indeed. My question was only to try to prevent you invented something new rather than reusing this method. Sounds perfect.

eddiemuc:
@Lineflyer your question made me unsure whether I can merge this PR function-wise. Can I merge or is a change needed?

Lineflyer:
All good. Merging for further testing.

PR #11509, opened 2021-08-22
Fix #11484 - Define explicit font size for user text input

### LINKED ISSUE #11484: Smaller font size for logging and entering coords
### Describe your problem!

A (visual-impaired) user contacted us on support mail with this text (shortened):

---

I have not good eyes, I was on more operation with them, so I have
big problems to read smaller texts.

And the new GUI design was changed in the way, that the recatngles
were added around all the input data and the font of the input texts
was significantly decreesed.

It is mainly hard for the input of coordinates or distance/azimuth,
where the numbers must be writtten exactly.

Please, should it be possible increese the input font, especially
for fieds with numerical inputs? There is enough space in the fields
in the horizontal direction, now unused.

[...]
On the other way, the rectangles around the input fields seems me
unnecessary, taking a lot space on the screen. But as it is a one
of the main changes in the last GUI, the revert back unfortunately
cannot be expected.

---

I made two screenshots to prove he is right:
Screenshot 2021-08-16 224041
Screenshot 2021-08-16 224124

### How to reproduce?

Use logging or coordinate input

Smaller fonts as before

Same font size?

### Reproducible

Yes

### c:geo Version

2021.08.15

### System information

_No response_

### Additional Information

I asked the user to try the general scle up option on his device settings.

Lineflyer:
## Description

Elements where user enter texts should IMHO deserve a larger font compared to pure reading as the user needs to validate the text while entering. At the same time the slight increase brings it close to the previous size (before UI rework).

- Define new dimensions for text input fields
- Increase main text size (primary) by 2sp (16 -> 18)
- Define secondary text size as before (16) for future use 
- Apply primary text input field font size to test input style

| Before | After  |  
|---|---|
|textinput_before| textinput_after|
|waypoint_fontsize_before| waypoint_fontsize_after|

## Related issues

Last part to fix #11484

### DIFF
--- a/main/res/values/dimens.xml
+++ b/main/res/values/dimens.xml
@@ -46,6 +46,10 @@
     <!-- compass -->
     <dimen name="textSize_compassTargeting">26sp</dimen>
 
+    <!-- text input fields -->
+    <dimen name="textSize_textinputPrimary">18sp</dimen>
+    <dimen name="textSize_textinputSecondary">16sp</dimen>
+
     <!-- all other text -->
     <dimen name="textSize_detailsPrimary">16sp</dimen>
     <dimen name="textSize_detailsSecondary">14sp</dimen>
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -272,6 +272,7 @@
         <item name="android:textColor">@color/colorText</item>
         <item name="colorOnSurface">@color/colorText</item>
         <item name="android:gravity">top|left</item>
+        <item name="android:textSize">@dimen/textSize_textinputPrimary</item>
     </style>
 
     <!-- edittext (TODO: should at some point be completely replaced by textinput) -->


Lineflyer:
Still a bit smaller than with pre-UI release, but IMHO a significant improvement and still inline with new design font sizes (all around 16sp to 18sp)

image

For comparison, this is the current release:
image

Lineflyer:
@moving-bits 
Could you doublecheck whether my change in `styles.xml` should be done within `textinput_embedded` or better in `textinput_edittext`.
AFAICS those are not always used both (e.g. for editing the personal note it still seems to use the default font size), am I right?

eddiemuc:
For what it's worth: I think you used the right place. `textinput_edittext` is used to style the textinput view **around** the edit text field (border, label, prefix/suffix text/icon) while `textinput_embedded` is used to style the "core" edittext field inside. And you want to style this core edittext field.

There are probably some legacy places in c:geo which are not yet switched to material.io Textinput, those places probably still use the legacy style `edittext`. I don't know if PN field is among those. For now, it might make sense to apply the new font size additionally to this legacy style to also cover those places.

But let's wait for @moving-bits s assessment.

moving-bits (approved):
LGTM, and @eddiemuc's explanation is correct as well 👍 

Only thing I'm not sure about is whether we really want to define a `textSize_textinputSecondary` style already with an actual use for it now. (And I'm unsure whether we ever want to define an input field with smaller font size anyway.)

BTW: Normally, a `textinput_embedded` should be enclosed by either `textinput_edittext`, `textinput_edittext_singleline` or `textinput_autocompletetextview`.

Lineflyer:
Well, I defined the `textSize_textinputSecondary` in case some text field really requires smaller fonts after changing it now. Current tests however show for now its OK.

I will merge and can remove that if needed later on.

PR #11515, opened 2021-08-22
fix input textsize for PN (rel. to #11484)

moving-bits:
## Description
PN edit field already used an explicit textSize definition. Changed this to the newly added `textSize_textinputPrimary` to make it consistent with other textinput fields.

## Related issues
#11484


Lineflyer (approved):

main/res/layout/fragment_edit_note.xml
```diff
@@ -20,7 +20,7 @@
             android:importantForAutofill="no"
             android:inputType="textMultiLine"
             android:minLines="8"
-            android:textSize="@dimen/textSize_detailsPrimary" />
+            android:textSize="@dimen/textSize_textinputPrimary" />
```
fm-sys, line 23: Wouldn't it be enough to just remove the definition completely and use the style value?

Lineflyer: If it is possible it would indeed avoid future searching (like I did before).

moving-bits: Updated the PR accordingly

moving-bits:
BTW: After updating the PR to use the `textinput_embedded` style, a slight "hack" is needed (set `android:minLines` to 99 instead of 8), otherwise the orange border does not fill the full available space:

|minLines=8 (as before)|minLines=99|
|---|---|
|image|image|

I guess instead of "99" you can use a different large-enough number, but "8" was definitively too small on my test device. Haven't seen any side-effects from using a "too-large" number yet.

PR #11523, opened 2021-08-23
Add "Has offline found log" to status filter & change quick settings mapping

### LINKED ISSUE #11430: Only hide found offline logs when using filter for hide offline logs

fm-sys:
fix #11430

---

grafik

grafik

grafik


fm-sys:
@eddiemuc Please review if I did everything right in my first _filter-framework-PR_. Thanks!

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -368,7 +368,7 @@ private static void migrateSettings() {
             legacyGlobalSettings.put(GeocacheFilter.QuickFilter.FOUND, !legacyExcludeFound);
             legacyGlobalSettings.put(GeocacheFilter.QuickFilter.DISABLED, !getBooleanDirect("excludedisabled", false));
             legacyGlobalSettings.put(GeocacheFilter.QuickFilter.ARCHIVED, !getBooleanDirect("excludearchived", false));
-            legacyGlobalSettings.put(GeocacheFilter.QuickFilter.HAS_OFFLINE_LOG, !getBooleanDirect("excludeofflinelog", false));
+            legacyGlobalSettings.put(GeocacheFilter.QuickFilter.HAS_OFFLINE_FOUND_LOG, !getBooleanDirect("excludeofflinelog", false));
```
eddiemuc, line 371: This is the migration routine for legacy c:geo versions before filtering was introduced. Your change would migrate the legacy global setting to filter for offline logs to the new "found offline log" instead of the old "offline log". Is this intended?

eddiemuc (approved):
LGTM except for one comment in settings migration. This comment is not critical.

main/src/cgeo/geocaching/settings/Settings.java
fm-sys: The new quick settings will no longer support `HAS_OFFLINE_LOG`. So we can either migrate the filter setting to `HAS_OFFLINE_FOUND_LOG` or drop it (it was only used in early beta). 

I can remove it, I don't have a real preference...

fm-sys: I removed it now. It's only about early beta users anyway... (they hopefully know how to open the quicksettings ;-))

PR #11525, opened 2021-08-23
adapt filter name of 'stored since' to be more clear

fm-sys:
Previously, the name was misleading. At the first, I thought it's the time stamp when I saved the cache the first time... But it's the 'last update' time stamp.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -512,7 +512,7 @@
     <string name="cache_filter_location">Location</string>
     <string name="cache_filter_stored_lists">Stored Lists</string>
     <string name="cache_filter_origin">Origin</string>
-    <string name="cache_filter_stored_since">Stored since</string>
+    <string name="cache_filter_stored_since">Last details update</string>
 
     <string name="cache_filtergroup_basic">Basic</string>
     <string name="cache_filtergroup_details">Details</string>


main/res/values/strings.xml
```diff
@@ -512,7 +512,7 @@
     <string name="cache_filter_location">Location</string>
     <string name="cache_filter_stored_lists">Stored Lists</string>
     <string name="cache_filter_origin">Origin</string>
-    <string name="cache_filter_stored_since">Stored since</string>
+    <string name="cache_filter_stored_since">Last details update</string>
```
Lineflyer, line 515: Maybe better something shorter like "Last updated"?

eddiemuc: There are two timestamps in c:geo per cache. One telling the last time when it's details were updated (typically via refresh) and a second one telling when anything was updated (set e.g. after the cache was loaded into map viewer from server - but as far as I remember not persisted in this case).

I remember discussions to update parts of a stored cache when it is loaded into the map viewer anyway (e.g. cache and found state). Once something like this is implemented it becomes very important to distinguish between "something but not all was updated" and "cache details were updated completely"

TMALSS: it might make sense to keep the "details" part in the filter/sort name.

PR #11534, opened 2021-08-25
remove transition animation when changing viewpager tabs (fix #11527)

### LINKED ISSUE #11527: Don't animate tab change when opening cache
### Describe your problem!

Three users reached out to us, with the following "problem":

The introduction of the material tabs instead of the old viewpager causes the cache infos always to start on the left most tab (wypoints) after opening and only then scrolls (animated) to the wanted tab.
Two problems with that:

- Two of the users have activated the option "Remember last page" in Settings - Cache-Details which leads to extensive/long lasting animations if e.g. gallery was the last opened tab.
- The third user does not use the "remember last page" but uses a rather old device (Android 6) and reports that even the auto scrolling from waypoints to details is stuttering and takes a lot of time. 

As this behavior occurs each time when opening a cache, I can assume it can be annoying to some users.
Is there a possibility to avoid this animation if a certain tab is put into focus on opening a cache or do we have to live with that?

### How to reproduce?

- Activate "Remember last page" in Settings - Cache-Details
- Open a cache and open the gallery tab
- Close the cache
- Open another cache

Animated scrolling until gallery tab is reached

Directly open with gallery tab 

respectively (if "Remember last page" is not active):

Directly open with details tab

### Reproducible

Yes

### c:geo Version

2021.08.15

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Disable the page transition for viewpager2 used in `CacheDetailActivity` and other places.

### DIFF
--- a/main/src/cgeo/geocaching/activity/TabbedViewPagerActivity.java
+++ b/main/src/cgeo/geocaching/activity/TabbedViewPagerActivity.java
@@ -74,7 +74,7 @@ public abstract class TabbedViewPagerActivity extends AbstractActionBarActivity 
             );
         }
 
-        new TabLayoutMediator(findViewById(R.id.tab_layout), viewPager, (tab, position) -> tab.setText(getTitle(getItemId(position)))).attach();
+        new TabLayoutMediator(findViewById(R.id.tab_layout), viewPager, true, false, (tab, position) -> tab.setText(getTitle(getItemId(position)))).attach();
     }
 
     private final ViewPager2.OnPageChangeCallback pageChangeCallback = new ViewPager2.OnPageChangeCallback() {


eddiemuc:
> BTW, just wondering: Why do we merge all UI improvements to release?

Imho we should only merge Bugfixes to release. In case of this PR I would argue that it is a critical Bugfix since many users complained about it.

moving-bits:
> BTW, just wondering: Why do we merge all UI improvements to release? [...]
> Not that it's bad in principle, but it seems to me that sometimes we develop too actively on the release branch

Generally agree with that statements, but in this case I regarded this a a bugfix, as we had several support tickets with users complaining about the delays.

And I guess that our last beta was somewhat special in this regard, as we started into a beta very early in the process to get a broader feedback, and continued development on `release`, applying a mix of fixes and minor changes along the way.

But you're right, we should keep in mind to restrict `release` to bugfixes again.

PR #11556, opened 2021-08-27
add reference + source for route invert order icon (rel. to #11389)

moving-bits:
## Description
see https://github.com/cgeo/cgeo/pull/11389#issuecomment-906777137
- adds reference to source file in svg for route invert order icon
- adds unmodified source file for this

### DIFF
--- a/main/res/drawable/ic_menu_swap_vert.xml
+++ b/main/res/drawable/ic_menu_swap_vert.xml
@@ -1,3 +1,4 @@
+<!-- taken from baseline_swap_vert_24 / material.io -->
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
     android:width="24dp"
     android:height="24dp"


moving-bits:
@murggel 
Following the discussion about SVG for the docs: Shall I remove the "SVG" (XML) from this PR then, keeping only the added comment?

murggel:
> @murggel 
> Following the discussion about SVG for the docs: Shall I remove the "SVG" (XML) from this PR then, keeping only the added comment?

If it shall be referenced in the manual,  you can keep it and I will add it to the DokuWiki, otherwise you can remove it...

moving-bits:
Hm, thinking about it, we seem to try to accomplish two things at the same time:
a) keep original / source file for a drawable
b) save a SVG/PNG version for documentation

This PR adds a comment in the Android drawable (`drawable/ic_menu_swap_vert.xml`) pointing to the source, which should be sufficient for b). And as this PR does not add a SVG but the original XML only, a) can be served better with the drawable actually used in c:geo (in combination with your script).

So if I'm not mistaken, adding the original drawable in XML format does not make sense here.

murggel:
> Hm, thinking about it, we seem to try to accomplish two things at the same time:
> a) keep original / source file for a drawable
> b) save a SVG/PNG version for documentation
> 
> This PR adds a comment in the Android drawable (`drawable/ic_menu_swap_vert.xml`) pointing to the source, which should be sufficient for b). And as this PR does not add a SVG but the original XML only, a) can be served better with the drawable actually used in c:geo (in combination with your script).
> 
> So if I'm not mistaken, adding the original drawable in XML format does not make sense here.

Ok, this is right, I thought you add the svg. You can remove the xml from raw images. We can get the svg then from web if necessary.

moving-bits:
ok, removed the drawable in `rawimages` from this PR, keeping only the added comment

murggel (approved):
Thanks

PR #11564, opened 2021-08-28
add 'does not contain' option

### LINKED ISSUE #11276: Add [does not contain] to text filters

fm-sys:
fix #11276

grafik

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/StringFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/StringFilter.java
@@ -22,6 +22,7 @@ public class StringFilter {
         IS_PRESENT(R.string.cache_filter_stringfilter_type_is_present),
         IS_NOT_PRESENT(R.string.cache_filter_stringfilter_type_is_not_present),
         CONTAINS(R.string.cache_filter_stringfilter_type_contains),
+        DOES_NOT_CONTAIN(R.string.cache_filter_stringfilter_type_does_not_contain),
         STARTS_WITH(R.string.cache_filter_stringfilter_type_starts_with),
         ENDS_WITH(R.string.cache_filter_stringfilter_type_ends_with),
         PATTERN(R.string.cache_filter_stringfilter_type_pattern);
@@ -108,6 +109,8 @@ public class StringFilter {
         switch (this.filterType) {
             case CONTAINS:
                 return matchGcValue.contains(matchTextValue);
+            case DOES_NOT_CONTAIN:
+                return !matchGcValue.contains(matchTextValue);
             case STARTS_WITH:
                 return matchGcValue.startsWith(matchTextValue);
             case ENDS_WITH:
@@ -173,6 +176,10 @@ public class StringFilter {
 
             switch (this.filterType) {
                 case CONTAINS:
+                    matchTextValue = "%" + matchTextValue + "%";
+                    break;
+                case DOES_NOT_CONTAIN:
+                    sb.append(" NOT");
                     matchTextValue = "%" + matchTextValue + "%";
                     break;
                 case STARTS_WITH:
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/filters/core/StringFilter.java
```diff
@@ -131,6 +134,9 @@ public void addToSql(final SqlBuilder sqlBuilder, final String columnExpression)
                 case IS_PRESENT:
                     sqlBuilder.addWhere(columnExpression + " IS NOT NULL AND " + columnExpression + " <> ''");
                     break;
+                case DOES_NOT_CONTAIN:
+                    sqlBuilder.addWhere("NOT " + getRawLikeSqlExpression(columnExpression));
```
eddiemuc, line 138: Might be more efficient to construct a "not like" SQL operator instead of inverting a "like" operator. But maybe it doesn't matter.

eddiemuc (approved):
Looks ok codewise, but CI fails with compile problem

main/src/cgeo/geocaching/filters/core/StringFilter.java
fm-sys: Could you show me how a "not like" SQL operator does look like?

I do very well understand the basic SQL concepts and example queries but unfortunately have never really worked with SQL actively...

eddiemuc: A like operator looks like this:

`Select * from gc_cache where name like '%beer%'`

A not like operator looks like this:

`Select * from gc_cache where name not like '%beer%'`

Your change will make the SQL like this:

`Select * from gc_cache where not name like '%beer%'`

It is probably a minor, neglectable thing

Lineflyer: > Select * from gc_cache where not name like '%bier%'

Wouldn't this select return all colums but not those where name is '%bier%'

Lineflyer:
retest this please

main/src/cgeo/geocaching/filters/core/StringFilter.java
eddiemuc: > > Select * from gc_cache where not name like '%bier%'
> 
> Wouldn't this select all colums but not those where name is '%bier%'

It would select all **rows** except the ones containing 'bier' in the name column - which would be the intended behavior of "does not contain".

Note that the actual SQL statement used by c:geo is far more complex (e.g. columns are selected explicitly, not using `*`. Also the function `lower()` is used to allow for case-insensitive select). The statement examples should just explain the `NOT LIKE` operator in contrast to a negated `LIKE` operator.

fm-sys: Ok, here we go. Thanks for the suggestion, looks IMHO also codewise much cleaner...

@eddiemuc Would you mind another quick cross-check?

eddiemuc: LGTM

PR #11565, opened 2021-08-28
Add automated branch merge workflow (only manual runs for now)

### LINKED ISSUE #11368: Automate branch merging
Our usual flow is to merge bugfix releases `release` and then back to `master`. Currently, this needs the explicit interaction by someone out of the core team, mostly @Lineflyer or @moving-bits.

However, automation is a create way to optimize our workflows and save time (for coding ;-)). 

Furthermore, it sometimes happens that important Bugfixes are only contained in `release` branch but are missing in the nightlys for some days. 

This could be improved by using Jenkins or Github Actions. Jenkins would probably need a crown job, while with Github Actions we can even set "push to `release`" as trigger directly, which is IMHO the preferred solution. 

[Github Actions](https://github.com/cgeo/cgeo/actions) is completely free for public repositories. 

If we decide to  do so, we should set up the flow with  Github Actions as follow:
- use `push to release branch` as trigger 
- apply our merge strategy "ours" 
- check if branch heads are equal, if so do nothing (maybe git is doing that automatically)  
- create a `merge branch 'release'` commit
- only use a regular push, no force  push (if there are merge conflicts a team member will still need to do the merge manually)

Please tell me if you like the idea or if there are any situations where `release` should explicitly not be merged to `master`.

fm-sys:
See discussion in #11368

### DIFF
--- a/.github/workflows/branch-merge.yml
+++ b/.github/workflows/branch-merge.yml
@@ -1,0 +1,33 @@
+name: Auto-merge branches 
+
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  # It was decided to don't run the workflow automatically for now
+  #schedule:
+  #  - cron: '0 0 * * *'
+
+env:
+  REPO: https://cgeo:${{secrets.GITHUB_TOKEN}}@github.com/cgeo/cgeo.git
+  RELEASE_BRANCH: release
+  MASTER_BRANCH: master
+
+jobs:
+  # This workflow contains a single job called "merge-branches"
+  merge-branches:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone and setup the repo
+        run: |
+          git clone ${{env.REPO}} -b ${{env.MASTER_BRANCH}} tmp
+          cd tmp
+          git config --local include.path ../.gitconfig
+          git config user.name cgeo branch-merge-bot
+          git config user.email 61464155+cgeo-ci-bot@users.noreply.github.com
+      - name: Merge branches and push 
+        run: | 
+          cd tmp    # path is reseted for every step, therefore we need to set it again 
+          git merge origin/${{env.RELEASE_BRANCH}} -m "merge branch '${{env.RELEASE_BRANCH}}' via workflow"
+          git push


.github/workflows/branch-merge.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  # It was decided to don't run the workflow automatically for now
+  #schedule:
+  #  - cron: '0 0 * * *'
+
+env:
+  REPO: https://cgeo:${{secrets.GITHUB_TOKEN}}@github.com/cgeo/cgeo.git
```
fm-sys, line 12: I'm not sure, if it needs to run on behalf of the organization (`https://cgeo:`) or an individual user (e.g. `https://cgeo-ci-bot:`). Probable, we simply need to try out which one works...

.github/workflows/branch-merge.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          git clone ${{env.REPO}} -b ${{env.MASTER_BRANCH}} tmp
+          cd tmp
+          git config --local include.path ../.gitconfig
+          git config user.name cgeo branch-merge-bot
+          git config user.email 61464155+cgeo-ci-bot@users.noreply.github.com
+      - name: Merge branches and push 
+        run: | 
+          cd tmp    # path is reseted for every step, therefore we need to set it again 
+          git merge origin/${{env.RELEASE_BRANCH}} -m "merge branch '${{env.RELEASE_BRANCH}}'"
```
Lineflyer, line 32: Maybe we could add an info into the message making clear, that the workflow has been used. 
Something like "Merge branch '${{env.RELEASE_BRANCH}}' via workflow"

fm-sys: Done

Lineflyer:
While I am not familiar with the workflow syntax I vote for merging this and trying it out.
As long as its manual it can only make our live easier IMHO.

Any objections?

Lineflyer (approved):

moving-bits:
There have been no objections, so let's try...

PR #11574, opened 2021-08-28
Support adding caches to bookmark lists (fix #118)

### LINKED ISSUE #118: Support adding caches to bookmark lists

fm-sys:
Only available for GC premium members - fix #118

grafik

grafik
grafik
grafik

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -410,6 +410,7 @@
     <string name="caches_remove_all_completely">Delete all from device</string>
     <string name="caches_remove_selected_completely">Delete selected from device</string>
     <string name="caches_delete_events">Delete past events</string>
+    <string name="caches_upload_bookmarklist">Upload to bookmark list</string>
     <string name="caches_upload_modifiedcoords">Upload modified coordinates</string>
     <string name="caches_upload_modifiedcoords_warning">Do you want to upload modified coordinates for all (selected) caches and overwrite the existing coordinates on the server? This cannot be undone.</string>
     <string name="caches_upload_allcoords">Upload coordinates for all caches</string>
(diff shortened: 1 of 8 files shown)


fm-sys:
TODO: support access from cache details activity

main/res/values/strings.xml
```diff
@@ -410,6 +410,7 @@
     <string name="caches_remove_all_completely">Delete all from device</string>
     <string name="caches_remove_selected_completely">Delete selected from device</string>
     <string name="caches_delete_events">Delete past events</string>
+    <string name="caches_upload_bookmarklist">Upload cache(s) to bookmark list</string>
```
Lineflyer, line 413: As you are inside the list menu, how about only "Upload to bookmark list" or (to make clear its about the website) "Add to website bookmark list"

Lineflyer:
Does it handle if the list contains caches not belonging to GC (UDC, other connectors)?

fm-sys:
> Does it handle if the list contains caches not belonging to GC (UDC, other connectors)?

Yes these caches will be ignored... See https://github.com/cgeo/cgeo/blob/75f49c3dc110f9437c54024bb48213d52981a5cf/main/src/cgeo/geocaching/connector/gc/GCParser.java#L1073-L1076

main/res/values/strings.xml
fm-sys: Done

main/res/values/strings.xml
```diff
@@ -1564,6 +1565,12 @@
     <string name="search_bookmark_list">Bookmark list</string>
     <string name="search_pocket_is_new">new</string>
     <string name="search_pocket_and_bookmark_loading">Loading a list of Pocket Queries and bookmark lists</string>
+    <string name="search_bookmark_loading">Loading bookmark lists</string>
+    <string name="search_bookmark_select">Select bookmark lists</string>
+    <string name="search_bookmark_new">Create new bookmark lists</string>
```
Lineflyer, line 1570: Should probably be singular "Create new bookmark list"

main/res/values/strings.xml
```diff
@@ -1564,6 +1565,12 @@
...
+    <string name="search_bookmark_adding_caches_success">All GC geocaches where added to the selected bookmark list</string>
```
Lineflyer, line 1571: "were" instead of "where" or even better "have been"

main/res/values/strings.xml
fm-sys: Thanks, good catch... Fixed.

main/res/values/strings.xml
fm-sys: You're totally right. That happens when you focus more on code architecture, than strings... Solved as well.

fm-sys:
Let's merge to master to have another cool feature for our upcoming beta. The code is completely decoupled, so at least no regressions are to be expected...

PR #11576, opened 2021-08-28
add delay option for updates check (fix #11548)

### LINKED ISSUE #11548: Offline data update check shouldn't prompt so often if canceled
### Describe your problem!

If the user has installed map/bbrouter data using c:geos built-in download feature after some time c:geo prompts the user to update that data on startup with a dialog "Do you want to check for updates of downloaded routing data files?"

That dialog has 2 buttons, OK to check and Cancel to not check. Select "Cancel", then restart c:geo and you'll be prompted again. The only way to not get asked is to accept the update check, otherwise you're prompted on every start.

Suggested solution: If the user doesn't want to check for updates ask again in X days, not right away.

### How to reproduce?

Install map/theme/bbrouter data via c:geo
Wait the update check time
Cancel the update check prompt
Kill c:geo
Start c:geo

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

-

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
- offer "later/ok" instead of "cancel/ok" for update check
- when selected "later", c:geo will prompt in three days instead of on next restart of c:geo

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2349,6 +2349,7 @@
     <string name="done">Done</string>
     <string name="finish">Finish</string>
     <string name="back">Back</string>
+    <string name="later">Later</string>
 
     <!-- ChipChoiceGroup -->
     <string name="chipchoicegroup_selectall">Select All</string>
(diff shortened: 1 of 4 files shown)


Lineflyer:
Good approach for now.

PR #11577, opened 2021-08-28
restrict AL connector to 100 results (fix #11538)

### LINKED ISSUE #11538: Way too many lab caches are loaded
Neither a real bug nor feature request:

Currently, 500 normal GC and 500 Lab caches are loaded. 

As there are much more normal caches than labs, this ration should be fine-tuned.

E.g. 500 GC / 100 ALC

grafik

moving-bits:
## Description
As described in the related issue, AL connector currently returns 500 hits on each request, which covers a much wider area than GC connector. To reduce server impact and make it more compatible with GC connector, restrict limit to 100 lab adventures.

### DIFF
--- a/main/src/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/cgeo/geocaching/connector/al/ALApi.java
@@ -92,7 +92,7 @@ final class ALApi {
         Log.d("_AL Radius: " + (int) radius);
         final Parameters params = new Parameters("skip", "0");
         final Parameters headers = new Parameters(CONSUMER_HEADER, CONSUMER_KEY);
-        params.add("take", "500");
+        params.add("take", "100");
         params.add("radiusMeters", String.valueOf((int) radius));
         params.add("origin.latitude", String.valueOf(latcenter));
         params.add("origin.longitude", String.valueOf(loncenter));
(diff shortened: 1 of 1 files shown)


baiti:
I was just wondering: Can this be addressed by using a hard coded constant? I always thought it depends on the density ratio:

non-ALC caches per area divided by ALC caches per area 

... this ratio could be adjusted by some sort of constant, e.g. "3" or maybe "4".

Using a hard coded "100" may work in a city like Berlin or Paris where it may yield a resaonably balanced mix but using "100" somewhere in New Mexico will most likely overpopulate the map with ALCs.

fm-sys:
> Using a hard coded "100" may work in a city like Berlin or Paris where it may yield a resaonably balanced mix but using "100" somewhere in New Mexico will most likely overpopulate the map with ALCs.

AL least it's less polluted as before (500) ;-)

Lineflyer:
IMHO the way of handling nearby caches depending on cache density (and with multiple connectors) need to be addressed anyway at some point in future.

You have to judge if a variable might help to easify a later implementation of some dynamic adjustment or if the hardcode does the trick for now.

fm-sys:
Let's merge to have it in our next beta. Although it's not perfect, it's at least much better than before...

PR #11587, opened 2021-08-29
Fix #11584 - Allow more space for log date by disabling ellipsis

### LINKED ISSUE #11584: Date in logbook cut off on larger fonts
### Describe your problem!

The date field on our logbook view is using ellipsis if the content does not fit.
Especially for locale which enforce a 4-digit year (like UK english on most devices) in combination with a small screen and/or larger font settings, users can end up with the year only showing `20...` which is pretty much uninformative.

image

### How to reproduce?

- Set locale to English - United Kingdom
- Depending on device used also increase font size a bit

Year cut off

Either the design should enforce the date to be shown (or maybe we can enforce a 2-digit year?!)

### Reproducible

Yes

### c:geo Version

2021.08.15

### System information

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

For some reason setting `marquee` does not work and I personally also think that marquee text is not so nice anyway.
To avoid ellipsis characters and get most possible room for the fields I suggest to simply disable ellipsizing at this place.

It should be a sufficient workaround for the problem described in #11584 without needing to increase the field size or similar.

## Related issues

#11584

## Additional context

If this later turns out not to be enough, there are AFAICS also possibilities to define a maximum font size scaling per text view.

### DIFF
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -503,7 +503,7 @@
         <item name="android:layout_width">wrap_content</item>
         <item name="android:layout_height">wrap_content</item>
         <item name="android:layout_gravity">right</item>
-        <item name="android:ellipsize">marquee</item>
+        <item name="android:ellipsize">none</item>
         <item name="android:gravity">left</item>
         <item name="android:lines">1</item>
         <item name="android:scrollHorizontally">true</item>


Lineflyer:
Tested this on some older and newer API levels with largest font and UK-english. Always fits.

Lineflyer:
Rebased onto `release` as it is a bugfix

PR #11589, opened 2021-08-29
fix #11572: correct filter text show logic, cleanup filter text logic

### LINKED ISSUE #11572: Filter status no longer shown in nearby search
### Describe your problem!

I have activated hiding of own, found and deactivated caches on the live map (live filter).
This is correctly shown in the filter status bar "Status: 3 Elemente".

When I use nearby search (where also the live filter is applied) it used to also show "Status: 3 Elemente" but since recently only shows "-".

See screenshots.
image

image

### How to reproduce?

- Open live map
- Use quick settings to filter out, found, own, deactivated
- Close live map
- Click on nearby search

Filter status bar shows `-` 
Filtering is working as expected

Filter status bar should show `Status: 3 Elemente`

### Reproducible

Yes

### c:geo Version

2021.08.28-NB1

### System information

```text
Samsung S20
Android 11
```

### Additional Information

_No response_

eddiemuc:
fix #11572: correct filter text show logic, cleanup filter text logic

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -1849,16 +1849,15 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
     }
 
     private void updateFilterBar() {
-        FilterUtils.updateFilterBar(this, getActiveFilterNames());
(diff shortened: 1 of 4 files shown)


fm-sys:
LGTM, let's merge...

PR #11595, opened 2021-08-29
don't show loading indicator on GMap (fix  #11585)

### LINKED ISSUE #11585: Stuck loading indicator on live map
On my emulator API 30 when running current `master` I have the following problem:
- Clicking a cache on live map shows the "Loading cache details" for split second before the popup is shown
- Note the "Loading cache details" window is still in background

image

- When closing the popup I got stuck here forever

image

Does not happen on my API 30 real device running todays nightly.

Emulator is running BM, real device is PM (if this makes any difference)

fm-sys:


fm-sys:
Kind of special to have a "red only" PR ;-)

grafik

Let's merge... (Nightly users will be grateful ;-))

PR #11620, opened 2021-09-03
fix watchlist (issue #11619)

### LINKED ISSUE #11619: Can't add/remove cache to watchlist
### Describe your problem!

I tap on the 'Add to watchlist' icon and get a "Changing your watchlist failed" popup

### How to reproduce?

1. open a cache
2. tap on the 'Add to watchlist' icon

get a "Changing your watchlist failed" popup

it should add the cache to your watchlist

### Reproducible

Yes

### c:geo Version

2021.08.,28

### System information

_No response_

### Additional Information

_No response_

ztNFny:
Update watchlist calls to match the update on GC.com

Fixes #11619

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCConstants.java
@@ -56,7 +56,7 @@ public final class GCConstants {
     static final Pattern PATTERN_IS_FAVORITE = Pattern.compile("<div id=\"pnlFavoriteCache\">"); // without 'class="hideMe"' inside the tag !
     static final Pattern PATTERN_FAVORITECOUNT = Pattern.compile("<span class=\"favorite-value\">\\D*([0-9]+?)\\D*</span>");
     static final Pattern PATTERN_COUNTLOGS = Pattern.compile("<span id=\"ctl00_ContentBody_lblFindCounts\"><p(.+?)</p></span>");
-    static final Pattern PATTERN_WATCHLIST_COUNT = Pattern.compile("data-watchcount=(\\d+)");
+    static final Pattern PATTERN_WATCHLIST_COUNT = Pattern.compile("data-watchcount=\"(\\d+)\"");
 
     /** Two groups ! */
     static final Pattern PATTERN_COUNTLOG = Pattern.compile("logtypes/([0-9]+)\\.[^>]+>\\s*([0-9,.]+)");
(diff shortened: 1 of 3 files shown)


fm-sys:
retest this please

(trigger CI - let's see if there are really more failing tests, or whether it was an one-time problem)

moving-bits:
@ztNFny 
Please use the keyword "fix #number_of_issue" in the PR's title to enable auto-closing on merge to `master`, e. g. for this case "fix #11619". Thanks.

bekuno:
> 
> 
> @ztNFny
> Please use the keyword "fix #number_of_issue" in the PR's title to enable auto-closing on merge to `master`, e. g. for this case "fix #11619". Thanks.

@moving-bits 
This PR used the keyword "Fixes #issue-nr" in the PR **description**. This is the official way to define the reference (see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue), and will auto-close the PR .
This form has three advanteges compared to only use the keyword in the PR title: at first you have the automatic linking on the PR "linked issue" list, second it is easy to navigate to the issue by clicking the PR number. And this second point is also useable after the PR is merged. Third you got a automatic created backlink on the referenced issue to the PR and an entry in the reference list there.

The title text with "fix #issue-nr" is only a auto-close marker, but of cause better to understand for reader than only "issue #issue-nr".

moving-bits:
But the references issue was NOT closed, even after merging `release` to `master` (which auto-closes the issues referenced by `fix #xxxx` in the PR title), that's why I closed the referenced issue manually.

bekuno:
> 
> 
> But the references issue was NOT closed, even after merging `release` to `master` (which auto-closes the issues referenced by `fix #xxxx` in the PR title), that's why I closed the referenced issue manually.

This is strange. I normally use the magic words only in the description, and this worked for these PR (normally).

moving-bits:
In the past I found only two methods were working reliably:
- Either adding "fix #xxxx" to the PR title (during commit, not sure if it's working if it is changed later in GitHub only)
- or adding a "linked issue" manually.

bekuno:
I will have a eye on my PR, maybe add the reference on both points.
By the way, on this PR at least the automatic linking to the Issue did work.
Additionally, if there is only the number in the title then I often miss the link in the description to be able to navigate to the issue.

fm-sys:
Problem is, that auto-close does only work when targeting `master`. So it needs to be contained in the commit message for PR's which are targeting to `release`, as only the commit message is merged back to `master` as well.

bekuno:
> 
> Problem is, that auto-close does only work when targeting `master`. So it needs to be contained in the commit message for PR's which are targeting to `release`, as only the commit message is merged back to `master` as well.

Of cause, this is the real reason! There is no "magic" with the PR title.
@moving-bits uses the Issue number in the commit and then in the PR as title, and I use the number in the description in the commit and PR. In both situation we have the Issue number in the commit and the auto close will work also when targeting the PR not to the default = master branch.

eddiemuc:
I also use (and recommend) the fix pattern in all three places: PR title, PR description and commit message. The commit message has the big advantage to have a great change documentation in git, even when the PRs are long gone/closed and forgotten.

PR #11623, opened 2021-09-03
remove disabling favo button (fix #11583)

### LINKED ISSUE #11583: Favorite button not enabled directly after posting found log
### Describe your problem!

From support mail and I was able to reproduce it:
When I log a cache and forget to assign a favorite point while logging, I can no longer add the favorite point after returning to the cache details.
I have to leave and reenter the cache to be able to assign a favorite point.

### How to reproduce?

- Open a cache
- Compose found log and send it (do not assign favorite point)
- When returned to the cache details, try to assign a favorite point using the corresponding `+` button

Button is greyed out, refreshing the cache or swiping from/to details does not help.
Closing and reopening the cache makes the button work as expected

Favorite button should be clickable without the need to close and reopen the cache

### Reproducible

Yes

### c:geo Version

2021.08.28

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Remove disabling the add/remove favo button - there's no need to, as the button's view gets already set to either `VISIBLE` or `GONE`.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1538,9 +1538,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             // Add/remove to Favorites is only possible if the cache has been found
             if (!cache.isFound()) {
-                binding.addToFavpoint.setEnabled(false);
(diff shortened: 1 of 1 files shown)


moving-bits:
failing test is unrelated (website change for watchlist)

Issue #11596, opened 2021-08-29
Refactor coordinate input dialog


Lineflyer:
### Describe your problem!

Seems my PR for font size optimization needs some fine tune.
On Facebook some people report cut off digits in coordinate input.

![image](https://user-images.githubusercontent.com/949669/131261403-14942cfe-1974-46b2-9c7e-d9d86d3555f8.png)

I asked for device type and font size settings to get a better impression.
Looks like it could be solved by not forcing rightbound.

### How to reproduce?

Coordinate input dialog

### Reproducible

Unclear

### c:geo Version

20210.08.28

Narrat:
Looking at the screenshot, I'm wondering why there are four chars reserved for the degrees and five and five for the minutes. Or does it only look like that?
Just visually I would expect that those input fields are as long as the chosen coord  (like DDD° MM.MMM')

Lineflyer:
As it is defined as a table it seems to shrink/expand those fields to fill the screen.
I am currently playing around with removing some margins and gravity options to get the most out of it. 

On the long run this dialog needs a fresh design...

Lineflyer:
@ztNFny 
As you are familar with style elements and formatting, do you have any idea for this.
Not necessarily completely redefining this view but at least finding out why there is so much prepending space on each element?

ztNFny:
I don't get why degree and minute fields have different sizes, according to the code they should be the same?

`

            <EditText
                android:id="@+id/EditTextLatDeg"
                style="@style/edittext_full"
                android:layout_marginLeft="2dp"
                android:layout_marginRight="2dp"
                android:gravity="right"
                android:inputType="number"
                android:textSize="@dimen/textSize_inputCoords"
                android:hint="@string/cc_hint_degrees"
                android:selectAllOnFocus="true" />

            <TextView
                android:id="@+id/LatSeparator1"
                android:text="°"
                android:textSize="@dimen/textSize_inputCoords"
                android:layout_marginRight="1dp"
                tools:ignore="HardcodedText" />

            <EditText
                android:id="@+id/EditTextLatMin"
                style="@style/edittext_full"
                android:layout_marginLeft="2dp"
                android:layout_marginRight="2dp"
                android:gravity="right"
                android:inputType="number"
                android:textSize="@dimen/textSize_inputCoords"
                android:hint="@string/cc_hint_minutes"
                android:selectAllOnFocus="true" />`

Lineflyer:
[...]

PR #11627, opened 2021-09-04
Work on #11596 - Set gravity left for seconds and fractions

Lineflyer:
## Description

As it can be read in #11596 the dialog might need to complete rework, I suggest a small workaround to make it a bit higher probability that the coords are readable also on very large fonts.

I set the gravity to `left` for seconds and fractions.
This makes at least the most common MinDec format look quite OK.

Before (Pixel 4 emulator, largest font):
image

After (Pixel 4 emulator, largest font):
image

This is how it looks with normal font sizes:
image

## Related issues

#11596


fm-sys:
IMHO this is a good solution, but shouldn't it be used for all text fields to stay consistent?

Lineflyer:
It somehow looked strange to me to make all leftbound.
And it makes sense, that the decimal fractions are now directly behind the `,` while the degree might stay apart from the reset.

moving-bits (approved):
LGTM, thanks for this workaround. Will merge this now to round-up our next bugfix release.

PR #11634, opened 2021-09-07
improve ALApi code

fm-sys:


baiti (approved):
LGTM .... besides ... I haven't even noticed that LC was renamed to AL in general

PR #11644, opened 2021-09-11
Fix #11638, Watchlist status not initialized/updated correctly (#11639) 

### LINKED ISSUE #11638: Watchlist status not initialized/updated correctly
### Describe your problem!

I recently submitted #11619 which was fixed in the latest release.
However it is still not working quite right - the displayed watchlist status is not accurate on refresh or initial load.

### How to reproduce?

1. In cgeo, load a cache that is **not** on your watchlist.
2. Do "Open in browser"
3. On the webpage, click "Watch" to add it to your watch list
4. return to cgeo and pull-down/release to refresh the cache details

- "The cache is not on your watchlist" is shown.
- Tapping on the "Add" icon results in a failure message
- **Note** the same issue arises if you load a cache that **is** already on your watchlist, but **isn't** currently stored in cgeo.

Watchlist status should be updated when cache details are initially loaded or refreshed.

### Reproducible

Yes

### c:geo Version

2021.09.08

### System information

_No response_

### Additional Information

_No response_

rsudev:
Apply fix from release on master


moving-bits:
Superseeded by merging `release` back to `master`

PR #11648, opened 2021-09-12
separate PQ list and bookmark list (fix #11458)

### LINKED ISSUE #11458: In "Lists / pocket queries" allow filtering by type
### Tell us your idea!

Since the addition of bookmark lists to the PQ download page the list can get very crowded.
When it was only PQs shown here there was (and still is, but now kinda useless) option to show only downloadable PQs vs. showing all of them.

It would be nice to change it to a filter that allows to:
- Show only downloadable PQs
- Show downloadable PQs + Lists
- Show Lists
- Show everything
(I think the other cases, like "show only PQs" are not really relevant)

Replace the existing switch with a menu for filtering

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Separates PQ list and bookmark list to avoid overcrowding.
"Show all" toggles between downloadable/all for PQ list, and between new/all for bookmark lists.


main/res/values/strings.xml
```diff
@@ -345,7 +345,8 @@
     <string name="menu_history">History</string>
     <string name="menu_filter">Filter</string>
     <string name="menu_scan_geo">Scan geocode</string>
-    <string name="menu_lists_pocket_queries">Lists / pocket queries</string>
+    <string name="menu_lists_pocket_queries">Pocket queries</string>
+    <string name="menu_lists_bookmarklists">Bookmarklists</string>
```
fm-sys, line 349: IMHO, there should be a space between "bookmark" and "lists" in English

fm-sys (commented):
Looks good in general. Maybe we should rename our `PocketQuery.java` class to something different as well (as it is used for bookmark lists as well as for PQs)

moving-bits:
> Maybe we should rename our `PocketQuery.java` class to something different as well (as it is used for bookmark lists as well as for PQs)

We could rename it to `Bookmarklist.java` or `Bookmarks.java` (or something similar), as pocket queries are sort of a subclass of bookmark lists.

main/res/values/strings.xml
moving-bits: Or maybe call it just `Bookmarks`?

fm-sys: I just have browsed a bit through their website and app, and it seems there's not a single place where they use the word "bookmark" anymore.

We still need "bookmark" to distinguish between our offline lists, but at least this means we cannot strip the word "lists" 

Screenshot_20210912_154235_com groundspeak geocaching intro_copy_432x1028

fm-sys:
> > Maybe we should rename our `PocketQuery.java` class to something different as well (as it is used for bookmark lists as well as for PQs)
> 
> We could rename it to `Bookmarklist.java` or `Bookmarks.java` (or something similar), as pocket queries are sort of a subclass of bookmark lists.

Or maybe `GCLists`?

fm-sys:
> > We could rename it to `Bookmarklist.java` or `Bookmarks.java` (or something similar), as pocket queries are sort of a subclass of bookmark lists.
> 
> Or maybe `GCLists`?

Or maybe more general `OnlineList`...

main/res/values/strings.xml
moving-bits: Ok, changed it to bookmark lists

moving-bits:
> > Or maybe `GCLists`?
> 
> Or maybe more general `OnlineList`...

I've refactored it to `GCList` as it is still GC-specific. Renamed the related layout files as well.

PR #11650, opened 2021-09-12
show info dialog in map when live mode is disabled (fix #7741)

### LINKED ISSUE #7741: Show a information toast/snackbar if live map is disabled
We still have a significant amount of users contacting our support who have the problem, that for them the live map is "not working".
After some conversation it turns out they accidentally disabled the live mode and due to unawareness thought the map stopped working as this disabled status is kept even after app restart.

AFAIR we discussed this somewhere here, but I cannot find it.

I suggest to make this setting transient, i.e. after restart of the app it should always be in live mode.
As we call the function "live map" on the mainscreen this is, what the users expect.

moving-bits:
## Description
Display a one time message on opening map when live mode is disabled.

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1518,6 +1518,8 @@
     <string name="map_language">Map language</string>
     <string name="map_live_enable">Enable live</string>
     <string name="map_live_disable">Disable live</string>
+    <string name="map_live_disabled">Live mode disabled</string>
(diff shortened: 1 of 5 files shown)


fm-sys:
Will this dialog only be shown when starting the top level map or also when going into the map via cache details?

moving-bits:
Good catch - currently it's shown always, but it should be restricted to `MapMode == LIVE` probably. I'll adapt the PR.

Issue #11615, opened 2021-09-02
[vectorIcons] adjust icon style for "Big icons for logged caches" ?


ztNFny:
Map markers need to be redesigned for "Big icons for logged caches" enabled (in MapMarkerUtils)

Expected look is TBD. Options:
- White map marker (with shape of connector) + smiley without border
- White map marker (with shape of connector) + smiley with border
- Smiley with border (in shape of connect)
- Smiley with border (always round)

Current style:
![image](https://user-images.githubusercontent.com/1258173/131861802-ec98f2c9-28f1-4c05-b3a7-aa63912c6e34.png)

ztNFny:
I'll implement option 1 for now, the other options require too many code changes. However some input from someone actually using that option would be welcome.

ztNFny:
> I'll implement option 1 for now, the other options require too many code changes. However some input from someone actually using that option would be welcome.

actually I won't, that code is in need of a refactoring. And maybe the current look is ok for those who prefer the big icons - after all you get exactly what it says :smile:

fm-sys:
![Screenshot_20210916_104242](https://user-images.githubusercontent.com/64581222/133580557-ed069ed6-d8f6-4f06-8fa6-d0b25cedf4b5.jpg)

Somehow, the cache type color is displayed even for found caches. This should IMHO not be the case for the "bigEmoji" style. Except for that, the style is perfectly fine 👍

ztNFny:
> Somehow, the cache type color is displayed even for found caches. This should IMHO not be the case for the "bigEmoji" style. Except for that, the style is perfectly fine

I've added a fix for not drawing the background in https://github.com/cgeo/cgeo/pull/11671
Question is, is the Hexagon with round log marker good or should it be a hexgon log marker?

fm-sys:
> Question is, is the Hexagon with round log marker good or should it be a hexgon log marker?

IMHO it's totally fine. Never have seen a face with corners ;-)

ztNFny:
> > Question is, is the Hexagon with round log marker good or should it be a hexgon log marker?
> 
> IMHO it's totally fine. Never have seen a face with corners ;-)

Spongebob would disagree 😉

ztNFny:
no ones is complaining, so let's close this for now

PR #11671, opened 2021-09-16
add found status in dotmarker cache (fix #11662) 

### LINKED ISSUE #11662: Incorrect compact icons (vector versions)
With current nightly I just planned my caching trips for next days and stumbled upon some (partly not deterministic) wrong compact icons:

- Some found caches shown cache color instead of smiley
- DNF caches show smiley
- Some unfound caches show smiley

I could not yet clearly reproduce any (pre)condition for that to happen.
But it happens often and you might notice if you switch between compact and normal mode.

Blind guess would be, that some icons "change places with others" regarding their compact marker, so it depends how many and what kind of caches and status is involved on the map.

I will post screenshots with examples next week (when back home)

2021.09.14-NB
Android 11
Samsung S20

ztNFny:


ztNFny:
Also addressing https://github.com/cgeo/cgeo/issues/11615#issuecomment-920710217

fm-sys:
LGTM 👍

PR #11676, opened 2021-09-17
Show latest offline logs as icons in cache details

### LINKED ISSUE #11646: Show the last few logs in map popup
### Tell us your idea!

When clicking a cache on the map the popup with summary information appears. This popup should get a new line showing icons for the latest logs available.
This would allow to get a quick glance at the "health" of the cache.

Similar features are implemented by GC little helper and Locus

Screenshot from GClh
image

### Usecase

As a cacher out in the field I select the next cache I plan to visit and want to see important information at a glance that will help me find the cache.
This includes D/T rating, size - and whether the cache was last found or not. If there's many DNFs I might not want to search for too long compared to a cache with recent founds. 

Potential problems:
- Information is outdated and thus the wrong conclusion is made - That however holds true for a lot of info when offline caching, the cache could be archived, moved, ... - Mitigation: Make it visible how old the information shown is.
- No logs available -> Don't show the section at all.

### Additional Information

Of course this can only use the latest information available, like for an offline cache this info might be outdated, but that's true for all the information, so not really an advantage.

You can assign this to me, I can take a look at this after vectorIcons is done.

ztNFny:
Adds the 8 latest log icons to cache details.

Code review (and improvements, branch is writeable) on the layout stuff would be welcome, I'm not sure whether LinearLayout is the best option for this, also the smileys size is set to a fixed int as I couldn't figure out the parents height.

## Related issues
Fixes #11646

## Additional context
image


Lineflyer:
The original issue was about the map popup not the cache details.

IMHO it fits more into the popup but in the cache details. If I already loaded/opened the cache its only a tap to see the last logs, but from the map popup it would be helpful...(but maybe only work for stored caches?).

I think its superflous information at this screen or at least not a big advantage.

ztNFny:
I added it to both screens

ztNFny:
> I think its superflous information at this screen or at least not a big advantage.

I like it here as you can at a glance see all the log types while in logs tab you have to scroll to see that. But yes, main use is in popup.
Maybe it's be better placed at the top of the logs tab than details.

Lineflyer:
If you added it to both (popup and details) you are probably right it should stay in both to be a bit consistent.

My previous post was focussed on the popup, but maybe you did it even more complete.

So functionwise I have to admit its nice. Lets try it out (if code review is OK). If needed we could fine tune later on.

What is the icon for a DNF? I would expect a blue frownie?!

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
```diff
@@ -227,4 +230,65 @@ public TextView addHiddenDate(@NonNull final Geocache cache) {
...
+
+    public void addLatestLogs(final Geocache cache) {
+        Context context = parentView.getContext();
+
+        final RelativeLayout layout = (RelativeLayout) activity.getLayoutInflater().inflate(R.layout.cache_information_item, null, false);
+        final TextView nameView = layout.findViewById(R.id.name);
+        nameView.setText(res.getString(R.string.cache_latest_logs));
+        final LinearLayout markers = layout.findViewById(R.id.linearlayout);
+
+        int smileySize = 48;
```
moving-bits, line 242: This will be a size in px, correct? That would lead to problems on screens with different densities.

Instead of using an absolute px value, would using one of the constants defined in DisplayUtils (and converting dp to px) help?

ztNFny: What I wanted to achieve is to make them as large as a line of text, not some explicit size. But stuff like parent.getHeight() didn't work and then I just went for a hardcoded size to get something.
If you have some idea how to properly grow them to the right height I'll use that (MATCH_PARENT works for height, but then width is wrong with both MATCH_PARENT and WRAP_CONTENT), otherwise I'll try to find some other constant that can be used.

ztNFny:
> What is the icon for a DNF? I would expect a blue frownie?!

I've implemented these icons/logtypes for now, all other logs are ignored (like published which imho isn't a useful info in that dialog. Will Attend I'd also not have added, but I had the icon ready anyway so I just added it ;-))
image

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
moving-bits: Can you use some of the scaling options for images, e. g. fit_center, and set height for surrounding LinearLayout to match_parent?

Or, if that does not work, you could use some "sp" value to take font scaling and different densities into account. 

Hard-coded px will only work for specific combinations of font scaling (see Android's system settings) and densities.

SammysHP:
Looks good, but it is not obvious which is the most recent log. Also the age in days of the latest log would be helpful.

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
```diff
@@ -227,4 +230,65 @@ public TextView addHiddenDate(@NonNull final Geocache cache) {
...
+        LinearLayout.LayoutParams lp = new LinearLayout.LayoutParams(smileySize, smileySize);
+        lp.setMargins(0, 0, 5, 0);
+
+        List<LogEntry> logs = cache.getLogs();
+        int i = 0;
+        while (i < logs.size() && markers.getChildCount() < 8) {
+            int marker = -1;
+            LogType lt = logs.get(i++).logType;
+            switch (lt) {
```
fm-sys, line 251: IMHO this switch Statement is quite ugly. Why not adding a new parameter to the `LogType` enum?

fm-sys (changes requested):

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
ztNFny: I couldn't figure out something with match_parent. For some reason I can't get Layout Inspector to work with c:geo, so trying to figure out what's going wrong here is a pain in the ...

I've now switched to using the text size as a base.

Lineflyer:
> but it is not obvious which is the most recent log

Mmh...do we need an indicator here. I guess most users would expect from newest (left) to oldest (right). I have no good idea how to visualize that without using much text.

> Also the age in days of the latest log

Agree, but also here, how to get it compact in that line? Maybe just place the number of days at the beginning in front of the icon row...difficult.

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
ztNFny: If the icon for a logtype was used in multiple places I'd fully agree, like this imho it's ok.

Do you think it'll be needed in additional places? Maybe we want to show the icon also in Logbook tab next to the status text?

fm-sys: Currently it's the only place, but we should always try to make things reusable. Usage in the Logbook tab sounds like a not too bad idea BTW. 

But there are more reasons, what if e.g. there is the need to introduce a new log type, etc? I would consider it easier to maintain if everything to mapped at a single place (the Logtype enum). Cause especially in such complex open source projects as cgeo, nobody can really keep every code part in mind and a spreaded switch statement is quite hard to keep bi sync... That's exactly the reason why we extensively use and should use enums at many places in cgeo.

ztNFny: ok, I'll move it to LogType.

#11684 opened for icons in logbook tab ;-)

ztNFny:
Removed the switch statement in favor of extending LogType. Also added a number of new log icons as an enabler for https://github.com/cgeo/cgeo/issues/11684

ztNFny:
> > but it is not obvious which is the most recent log
> 
> Mmh...do we need an indicator here. I guess most users would expect from newest (left) to oldest (right). I have no good idea how to visualize that without using much text.

Looking at other implementations of the feature, e.g. in GClh and Locus, there's also no indicator. LTR seems like the logical choice, at least for LTR languages. For RTL it probably should be reverted - but does c:geo even support those?

> > Also the age in days of the latest log
> 
> Agree, but also here, how to get it compact in that line? Maybe just place the number of days at the beginning in front of the icon row...difficult.

I don't see a compact way to get that info in. Even the text "(X days ago)" will use a lot of space. If we want that we should look into moving the smileys into a seperate section below the base info, similar to attributes.

But I think we're fine without this info: The important info imho is when the cache was last updated (which is already shown) as the most recent log can't be newer than that.
This "preview" will not completly replace looking at the Logbook anyway - it's more like "get the most important info at a glance", like "were there 5 DNFs recently, then I might not even visit it".
If the last found logs turns out to be 1 year old at that point - what does that really mean to you? You have to search harder as it might be covered by dirt, you need more info from previous logs, it might be missing, ... But for this you have to read the logs in detail.

fm-sys:
(Something seems to be wrong with our CI)

    18:43:19  org.gradle.api.UncheckedIOException: Could not add entry '/home/jenkins/.gradle/caches/transforms-3/3f816fadca748577347980fc62e853ef' to cache file-access.bin (/home/jenkins/.gradle/caches/journal-1/file-access.bin).

ztNFny:
> LGTM. Just one code style issue, codacy is claiming about...

strange, my local codacity didn't complain about that :-o Do I have to setup anything special?

fm-sys:
> 
> 
> > LGTM. Just one code style issue, codacy is claiming about...
> 
> strange, my local codacity didn't complain about that :-o Do I have to setup anything special?

There are quite some additional tests which are running on Codacy, e.g. `NPathComplexity` and such things. But I haven't found a way to locally set up these tests either (yet) ...

BTW, this time jenkins was successful 🎉

bekuno:
> > but it is not obvious which is the most recent log
> 
> Mmh...do we need an indicator here. I guess most users would expect from newest (left) to oldest (right). I have no good idea how to visualize that without using much text.

a little late for this PR, but ...
how about if the symbol if the newest log is slightly larger, e.g. 120 %?

bekuno:
> > strange, my local codacy didn't complain about that :-o Do I have to setup anything special?
> 
> There are quite some additional tests which are running on Codacy, e.g. `NPathComplexity` and such things. But I haven't found a way to locally set up these tests either (yet) ...

You need the "CheckStyle-IDEA" Plugin installed in AS.
The AS integration uses our c:geo configuration (defined by `.idea/checkstyle-idea.xml`, which references to `checkstyle.xml`), and the same file is used on codacy.

Then you can run Checkstyle checks from the Checkstyle Toolbar in AS. And you should activate the "Scan with Checkstyle" option in the Commit dialog. Also it is possible to call `gradlew checkstyle` in the AS terminal.
This call do we use also in our CI task for PR.

wolverine007:
Why only the last 8 icons?
There is plenty of space left to the rigth, so we might be able to state the last 10 as well.

Maybe we could add the dates for the last found and DNF or last two logs next to the information, when the caches has been stored:
Icons

ztNFny:
> Why only the last 8 icons?

I had to pick SOME number 😉

ztNFny:
That said, your screenshot with 10 smileys, it feels too busy to me.
I had 5 at the beginning, but that was too few for the info to convey, so it should be >5. 7 or 8 FEELS right to me somehow - no hard reason though

Lineflyer:
> I had 5 at the beginning, but that was too few for the info to convey, so it should be >5. 7 or 8 FEELS right to me somehow - no hard reason though

I think the current amount is quite OK. The use case is probably only to see whether there is a recent problem with that cache.
Additionally we have to think about smaller screens/lower res like this:

![image](https://user-images.githubusercontent.com/949669/134244680-7b955d34-a87c-4fc4-bf13-bdfbdc50f0e9.png

> Maybe we could add the dates for the last found and DNF or last two logs next to the information, when the caches has been stored:

Interesting idea!

wolverine007:
> 
> 
> That said, your screenshot with 10 smileys, it feels too busy to me.
> I had 5 at the beginning, but that was too few for the info to convey, so it should be >5. 7 or 8 FEELS right to me somehow - no hard reason though

Mainly driven by personal preferences - maybe combine it with an user chosen setting, so the user could chose between 5 and 10 or hide it in total (more fun while being already in the wild, searching for more than one hour and finding out that the last 5 logs were DNFs? :) :) :)

Lineflyer:
> maybe combine it with a user chosen setting

No, for sure not a setting for each and everything....we already have to much of them :(

ztNFny:
How about putting the info about last log date, last found date in a tooltip when clicking on the smiley row?
That way it wouldn't clutter the UI but still be easily accessible.

wolverine007:
> Interesting idea!
Thanks :)

Good point with the low res - I was already playing around with the English and German interface, while doing so I saw this cut-off in the German screen on the right:
English_German
 
So I suggest to name recent logs to Letzte Logs, which might fit without being dotted out.

And for Schwierigkeit no digit is missing, but three dots are smaller than eit,
so are we able to shift the full block to the right, so Schwierigkeit ist not displayed as Schwierigk... ?!

Lineflyer:
We can forbid the ellipsis (`...`) in general for this row to gain space. IMHO its better to show whats fitting instead of wasting three characters by three dots.

wolverine007:
> 
> 
> How about putting the info about last log date, last found date in a tooltip when clicking on the smiley row?
> That way it wouldn't clutter the UI but still be easily accessible.
Sounds awesome, with a pop-up you could display much more info at once

> We can forbid the ellipsis (...) in general for this row to gain space. IMHO its better to show whats fitting instead of wasting three characters by three dots.
Also here, sounds more than valid to me *thunbs up*

ztNFny:
"neuste logs" in german would fit better imho

Btw, i checked what others are doing (not that we have to follow that, but as a reference it's ok):
Locus: 5 log icons, no further info
GClh: 8 icons, showing the whole log as tooltip
Supported by that I would not add a further text like "last log on date" to the dialog, imho it's not important enough to take that valuable space - not all info can be available at once.

wolverine007:
> GClh: 8 icons, showing the whole log as tooltip
> Supported by that I would not add a further text like "last log on date" to the dialog

Yeah, but you should keep in mind, that the icons are quite small and hitting a single one might be impossible with my "sausage fingers", so the tooltip should be clickable via the full line and stating the complete info in once.
Or what do you think?

ztNFny:
> > GClh: 8 icons, showing the whole log as tooltip
> > Supported by that I would not add a further text like "last log on date" to the dialog
> 
> Yeah, but you should keep in mind, that the icons are quite small and hitting a single one might be impossible with my "sausage fingers", so the tooltip should be clickable via the full line and stating the complete info at once.
> Or what do you think?

My idea was to have a tooltip for the row,  not an individual log.
GClh is for desktop, implementing the same in cgeo doesn't make sense.

Question is, for the tooltip text, what should it contain? Imho the only relevant info is last logs date - for everything else it's probably too dependant on the situatiom and the user will look at the logbook tab anyway?!

Like, if the last log is a DNF, do you really care whether the found log before that was 1 day prior or 1 month? In that scenario i'd first look at the DNF users found count to see whether its a newbie, then read the DNF log and only then care about the previous found log.

Lineflyer:
> My idea was to have a tooltip for the row, not an individual log.

Might work...

> Question is, for the tooltip text, what should it contain?

wolverine007:
you could keep it simple with just the textual info
Last found: xx-xx-xxxx | Last DNF: xx-xx-xxxx
or
the way I would might prefer it:
(Icon) xx-xx-xxxx
(Icon) xx-xx-xxxx
(Icon) xx-xx-xxxx
...
..
.

ztNFny:
> you could keep it simple with just the textual info
> Last found: xx-xx-xxxx | Last DNF: xx-xx-xxxx
> or
> the way I would might prefer it:
> (Icon) xx-xx-xxxx
> (Icon) xx-xx-xxxx
> (Icon) xx-xx-xxxx
> ...
> ..
> .

Don't think icons are supported in that tooltip class.

Why do you care about the last dnf date (vs just the last log date)?

ztNFny:
Or: make the click on the row of smileys jump straight to the logbook tab. All info you want is available there and you save some swipes compared to opening the cache details, then swiping to the logs tab.

wolverine007:
> Why do you care about the last dnf date (vs just the last log date)?

Sometimes it is useful to see if these dates are close to each other or if they are not. For a mystery cache with a higher difficulty and low found count, this is more interesting than for a traditional with founds every day.

But you are right, much more info could be gathered inside the full log - so maybe a direct link is usefull as well.

Is the UI capable to do different actions for a click (=popup info box) and hold (=jump to logs)?

But maybe we are drifting away from the initial topic ;)

eddiemuc:
My 2 cents:
I don't like the idea of having yet another "secret" clickable action in the GUI, we have too many of those already. If the last log date should be displayed in the initial screen, then I vote for simply put in another line of information. If anyone wants more info, the log is just two swipes away.

fm-sys:
> If the last log date should be displayed in the initial screen, then I vote for simply put in another line of information. 

IMHO, we shouldn't add even more textual information to that page. cgeo is already today way too much text based, and the more text we add the more difficult it gets to find the information you really like to see.

So either use a tooltip or don't add it to the details page at all IMHO.

Lineflyer:
I would also be still in favor for a simple tooltip (no hidden long press) and that tooltip could just display the newest log date.
As said above, the rest is available in the logbook. The last finds icons + the last log date should be enough.

eddiemuc:
> I would also be still in favor for a simple tooltip (no hidden long press) and that tooltip could just display the newest log date.
> As said above, the rest is available in the logbook. The last finds icons + the last log date should be enough.

A tooltip by definition opens on long-click in Android (see https://developer.android.com/guide/topics/ui/tooltips)

ztNFny:
I can make the tooltip show by binding click to longclick

eddiemuc:
> I can make the tooltip show on click by binding click to longclick
> 
> myTextView.setOnClickListener{
> myTextView.performLongClick()
> }

My point was more going in the direction of Android UX conventions: as far as I understand an Android user would commonly expect tooltips to appear on long click because this is the way it is defined by Android itself (and most likely it is also done in other apps this way).

ztNFny:
We already bind click to longclick in other places, like the username in logs section.

eddiemuc:
> We already bind click to longclick in other places, like the username in logs section.

But there we don't display just a tooltip but start a real, self-defined action

eddiemuc:
> We already bind click to longclick in other places, like the username in logs section.

Ah and this is btw one of the examples for "hidden functions" in c:geo which I think we should avoid.

In fact, I learned about this specific feature (getting options by tapping user name) only after I saw it in code one day. In the years before (as a normal c: geo user) I never knew of this function.

ztNFny:
In the cache details we have several things responding to a single click:
Username -> open popup
Coordinates -> change format
Attributes -> switch between text/attributes

And some items, like name, GC code copy on long-click

Considering that I think having a tooltip "Last log date: XX" on click event fits the existing design - whether that existing design is discoverable is another topic imho ;-)


For log icons in logbook header (https://github.com/cgeo/cgeo/issues/11684#issuecomment-924178076) I currently also have implemented tooltips on click, maybe we can also discuss long vs short click there once I have the PR.

Issue #11659, opened 2021-09-13
Different final marker overlays


Lineflyer:
Quick note from vacation:

Newest NB with vector icons:
Some of my solved caches which are moved to final have a white flag on blue at lower right, some have a purple pin at that position.

Two questions arise:
1. Why is there a difference?
2. Why dont we use s small version of the final waypoint icon instead (red flag on white or inverted). Would IMHO be more consistent and understandable.

ztNFny:
Flag=corrected coordinates
Pin=has a final waypoint

Up until now these 2 completly different cases had the same icon which was simply wrong.

As for the color scheme: white background doesn't fit the scheme and has low visibility. Red background is too prominent. So i went for blie as most users will be accustomed from GC.

bekuno:
If we have final coordinates then we should show a final flag, but not a Pin.
Maybe we can use different final flag colors/backgrounds.

ztNFny:
A different flag can't be differentiated as the icon is too small.
That's why i went with the pin to indicate that you should look at the list of waypoints.

Using the same icon for 2 different cases and just differentiating by color is bad as it ignores accessibility.

Lineflyer:
Thanks @ztNFny for explanation. I might have missed the discussion due to my vacation.

My spontanous feedback would be, that the PIN overlay does not carry any valuable information (at least on the map). Because if a final waypoint is defined or not does not change the location of the shown icon. 
So if we now support to handle these cases different I would suggest to show no special indication at all for that case (final defined for that cache but icon is not on shown on that coords).

But let me think about it and use it some more days...still feedback welcome!

Regarding the color:
Do you have an example of how red flag on white or white flag on red would look like?

bekuno:
The final flag should have the same meaning on multi caches as on mysteries - the final pount is identified. 
If the cache coordinate is moved to another point then the situation is like a normal cache. Mayby change then the cache symbol color to green like the color for normal caches.

ztNFny:
> So if we now support to handle these cases different I would suggest to show no special indication at all for that case (final defined for that cache but icon is not on shown on that coords).

That's also my personal opinion, but there were several voices against removing the indicator in that other issue. Those users seem to only add a final waypoint instead of modifying the coordinates (which seems to be the more common use, at least all people i personally know do it that way).
So i kept an indicator but made sure it can be differentiated.

> Regarding the color:
> Do you have an example of how red flag on white or white flag on red would look like?

[...]

PR #11682, opened 2021-09-18
[vectorIcons] tweak "has final" marker

ztNFny:


fm-sys:
That's how it will look like:

grafik

I will not yet merge the PR, so that others have a chance to add comments.

@ztNFny: It's always helpful to provide pictures, especially if the PR is about graphic changes...

ztNFny:
This is based on the discussion in https://github.com/cgeo/cgeo/issues/11659
I don't think we need yet another place to have this discussion.

fm-sys:
> 
> 
> This is based on the discussion in #11659
> I don't think we need yet another place to have this discussion.

Ah, ok. Hadn't followed the discussion completely (over 40 comments 😯) and was not sure if there was an agreement. But it seems there was...

---

My proposal:
Let the red flag as known symbol for "has a final" (instead of the pin).
And use the blue flag for "is user modified".

_Originally posted by @bekuno in https://github.com/cgeo/cgeo/issues/11659#issuecomment-922370846_



Lets just use those two you pictured above in the way you suggested:
Blue : Corrected coords
Red: Has Final Waypoint (but no corrected coords)


_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/11659#issuecomment-922347778_

PR #11683, opened 2021-09-18
explicitly set textColor on Emojis as the default has some alpha

### LINKED ISSUE #11673: different opacities on emojis
### Describe your problem!

In the emoji selection dialog all emojis but the circles have partial opacity. Looking at the code I can see the implementation but I can't see any reason WHY this is done. The opacity settings were introduced by https://github.com/cgeo/cgeo/commit/ab5c5dd0dd712429016ea70286e2b7f83503d851

Imho opacity should be 1 for all emojis, being faded makes it harder to identify them as it lowers contrast.

Screenshot - all emojis but the circles are faded
signal-2021-09-16-122136_001

### How to reproduce?

-

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021.09.15

### System information

_No response_

### Additional Information

_No response_

ztNFny:
Display emojis in custom icon selector without transparency

## Related issues
Fixes #11673

### DIFF
--- a/main/res/layout/emojiselector_item.xml
+++ b/main/res/layout/emojiselector_item.xml
@@ -13,6 +13,7 @@
         android:layout_height="match_parent"
         android:layout_centerVertical="true"
         android:layout_centerHorizontal="true"
+        android:textColor="#ffffffff"
         android:gravity="center"/>
     <ImageView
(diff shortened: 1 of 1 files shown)


Lineflyer:
retest this please

PR #11687, opened 2021-09-18
Rename bookmark list string

Lineflyer:
## Description

The current string for adding a cache to the servers bookmark list it "Upload to bookmark list".
IMHO this sounds misleading as the cache is not uploaded but its simply, that the cache is added. So I suggest "Add to bookmark list", which is more understandable.

## Related issues

None


Lineflyer:
BTW: I also aligned all german strings to use "Lesezeichen" instead of "Bookmark". Currently it was mixed.

PR #11689, opened 2021-09-19
add icon for bookmark lists (fix #11685)

### LINKED ISSUE #11685: PQ and bookmark list menu use same icon
### Describe your problem!

On my tablet (where the action bar has more space) I do now see two of those "suitcase with magnifier" icons has the new bookmark list menu uses the same icon as the PQ list menu.

### How to reproduce?

Use portrait mode on a large screen device in the main menu

You will see two of those icons after each other

Have different icons 
Bookmark list could maybe something like a checklist with magnifier (or similar)

### Reproducible

Yes

### c:geo Version

2021.09.14-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Adds a separate icon for bookmark lists to be able to distinguish between PQ and bookmark lists

### DIFF
--- a/main/res/drawable/ic_menu_bookmarks.xml
+++ b/main/res/drawable/ic_menu_bookmarks.xml
@@ -1,0 +1,11 @@
+<!-- taken from material.io icons - bookmarks (outline) -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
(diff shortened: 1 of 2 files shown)


Lineflyer:
Release branch?

moving-bits:
Separating PQ and bookmarks is on `master` only, so I targeted the extra icon to `master` as well.

Lineflyer:
Ah ok.

PR #11691, opened 2021-09-19
Sort by 'Distance to target' for ADDRESS / COORDS searches

### LINKED ISSUE #11422: sort result of adress-search by distance to destination
### Tell us your idea!

For vacation I was looking for caches near our location. So I did a search for address. But the first caches which appears in the list were "far away" from the destination.

I guess it is because there are 400 cached loaded, sorted by distance to home / current location and not sorted by distance to destination location. 

Sort the seatch-result for address by distance to destination location

### Is there an alternative?

Add max distance to search for adress to restrict result to an specific area

Or

Prefill filter with coordinates so the distance range can be easily adapted - not sure if that makes sense and would be helpful

### Additional Information

_No response_

### LINKED ISSUE #11666: nearby search from cache returns caches around current position
### Describe your problem!

From support ticket (and verified in a quick test): Doing a nearby search from cache detail page returns caches around your current position instead of caches around the cache's position.	

### How to reproduce?

- open a cache which is not close to your current position
- open menu => extra => "Caches im Umkreis"	

c:geo will show you a list of caches around your current position (wrong value put into the filter)

Show caches around the cache's position

### Reproducible

Yes

### c:geo Version

2021.09.08

### System information

_No response_

### Additional Information

ticket 237974

fm-sys:
fix #11666 / fix #11422

### DIFF
--- a/main/src/cgeo/geocaching/sorting/SortActionProvider.java
+++ b/main/src/cgeo/geocaching/sorting/SortActionProvider.java
@@ -1,6 +1,7 @@
 package cgeo.geocaching.sorting;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.functions.Action1;
@@ -39,7 +40,7 @@ public class SortActionProvider extends ActionProvider implements OnMenuItemClic
 
     // Used to change menu Filter label
     private boolean isEventsOnly = false;
-    private boolean distanceToTargetAvailable = false;
+    private Geopoint targetCoords = null;
 
     private static final class ComparatorEntry {
         private final String name;
@@ -100,8 +101,8 @@ public class SortActionProvider extends ActionProvider implements OnMenuItemClic
 
     private void registerComparators() {
         registry.clear();
-        if (distanceToTargetAvailable) {
-            register(R.string.caches_sort_distance_target, TargetDistanceComparator.class, () -> TargetDistanceComparator.INSTANCE);
+        if (targetCoords != null) {
+            register(R.string.caches_sort_distance_target, TargetDistanceComparator.class, () -> new TargetDistanceComparator(targetCoords));
         }
         register(R.string.caches_sort_distance, DistanceComparator.class, () -> DistanceComparator.INSTANCE);
         if (isEventsOnly) {
@@ -176,7 +177,7 @@ public class SortActionProvider extends ActionProvider implements OnMenuItemClic
         this.isEventsOnly = isEventsOnly;
     }
 
-    public void setDistanceToTargetAvailable(final boolean distanceToTargetAvailable) {
-        this.distanceToTargetAvailable = distanceToTargetAvailable;
+    public void setTargetCoords(final Geopoint targetCoords) {
+        this.targetCoords = targetCoords;
     }
 }
--- a/main/src/cgeo/geocaching/sorting/SortActionProvider.java
+++ b/main/src/cgeo/geocaching/sorting/SortActionProvider.java
@@ -104,7 +104,7 @@ public class SortActionProvider extends ActionProvider implements OnMenuItemClic
         if (targetCoords != null) {
             register(R.string.caches_sort_distance_target, TargetDistanceComparator.class, () -> new TargetDistanceComparator(targetCoords));
         }
-        register(R.string.caches_sort_distance, DistanceComparator.class, () -> DistanceComparator.INSTANCE);
+        register(R.string.caches_sort_distance, GlobalGPSDistanceConparator.class, () -> GlobalGPSDistanceConparator.INSTANCE);
         if (isEventsOnly) {
             register(R.string.caches_sort_eventdate, EventDateComparator.class);
         } else {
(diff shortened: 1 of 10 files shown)


main/src/cgeo/geocaching/sorting/TargetDistanceComparator.java
```diff
@@ -0,0 +1,18 @@
...
+
+import cgeo.geocaching.location.Geopoint;
+
+/**
+ * sorts caches by distance to given target position
+ *
+ */
+public class TargetDistanceComparator extends AbstractDistanceComparator {
+
+    public static final TargetDistanceComparator DISTANCE_TO_TARGET = new TargetDistanceComparator();
```
eddiemuc, line 11: A construct using a static member is not necessary for target distance connector and should be removed. 

I assume this was copied from the approach for distance to current position. This is used to update the distance sort when the user moves and thus distance to caches may change.

For target distance however, the target coord and distance will remain constant, thus the target const can simply be passed in constructor. Usage of an ugly global var is not necessary.

main/src/cgeo/geocaching/sorting/SortActionProvider.java
```diff
@@ -99,6 +100,9 @@ private void register(@StringRes final int resourceId, final Class<? extends Cac
 
     private void registerComparators() {
         registry.clear();
+        if (distanceToTargetAvailable) {
+            register(R.string.caches_sort_distance_target, TargetDistanceComparator.class, () -> TargetDistanceComparator.DISTANCE_TO_TARGET);
```
eddiemuc, line 104: Don't use ugly global constant here, pass target coords in a constructor to `TargetDistanceComparator` instead.

The reason why this is different for distance to user is explained in my second comment.

main/src/cgeo/geocaching/sorting/TargetDistanceComparator.java
fm-sys: We however need an initialized instance here... I can rename both global vars to "INSTANCE", but this will not solve the general problem. Any idea?



grafik

fm-sys: I've just renamed both to "INSTANCE" for consistency. That's exactly how it is done for all other comparators as well. But I'm open for suggestions how to archive it differently...

main/src/cgeo/geocaching/sorting/SortActionProvider.java
```diff
@@ -171,4 +175,8 @@ public void setSelection(final CacheComparator selection) {
     public void setIsEventsOnly(final boolean isEventsOnly) {
         this.isEventsOnly = isEventsOnly;
     }
+
+    public void setDistanceToTargetAvailable(final boolean distanceToTargetAvailable) {
```
eddiemuc, line 179: Suggested solution: change this method as follows:
- Instead of passing a boolean, pass the actual target coordinate in this method and create a corresponding TargetDistanceComparator. 
- Store this as instance member instead of the boolean flag.
- replace checks for target availability with checks on whether the comparator is null
- use the comparator where needed

main/src/cgeo/geocaching/sorting/TargetDistanceComparator.java
eddiemuc: @fm-sys see my comment on another method for solution suggestion . Alternatively, create a new instance of the comparator here. For this to work, the target coord needs to get a member of the class (but I would personally prefer my other suggestion)

In all the cases where INSTANCE is used, the comparator is not parameterized, so usage of a constant instance is fine. TargetDistanceComparator however needs to be parameterized.

 "Distance to user" is special due to the ever-changing user-coord and thus necessity to resort (but on the other hand there can always be one user-coordinate globally at any point in time).

fm-sys:
@eddiemuc OK, hope I correctly understood your suggestion. Implemented in cb66ebd. 

Oh, and thanks for your review, feedback is always helpful and welcome! 👍

main/src/cgeo/geocaching/ui/CacheListAdapter.java
```diff
@@ -180,8 +183,11 @@ public CacheComparator getCacheComparator() {
         if (cacheComparator == null && series) {
             return NameComparator.INSTANCE;
         }
+        if (cacheComparator == null && targetCoords != null) {
+            return new TargetDistanceComparator(targetCoords);
+        }
         if (cacheComparator == null) {
-            return DistanceComparator.DISTANCE_TO_GLOBAL_GPS;
+            return DistanceComparator.INSTANCE;
```
eddiemuc, line 190: INSTANCE is the standard name used for singleton pattern, it is known to be used for classes where only one instance exists and is used.

 If class `DistanceComparator` is meant to be a Singleton class which is solely used for compare to global GPS then I suggest renaming it to e.g. `GlobalGPSDistanceConparator`

If class `DistanceComparator` should also be used for other cases (like it was implemented before) then I would strongly suggest leaving the constant name as it was - the class is not a Singleton in this case.

fm-sys: > If class `DistanceComparator` is meant to be a Singleton class which is solely used for compare to global GPS then I suggest renaming it to e.g. `GlobalGPSDistanceConparator`

Yes, it is designed as singleton class, as userGPS will be the only value which can change. For other use cases we have `AbstractDistanceComparator` and `TargetDistanceComparator`. The comparator provider is BTW designed that way that every comparator requires it's own class, so every comparator class needs to be a singleton. I will rename the class accordingly.

fm-sys: My first implementation indeed was to create simply a new instance of DistanceComparator (without new classes), but it was not working without deeper refactoring...

PR #11692, opened 2021-09-19
remove svgs from raw-images (rel cgeo/cgeo-guide#40)

### LINKED ISSUE #11562: Do we really need the rawimages/svg for Doku-Wiki?
### Tell us your idea!

If we use png for the Doku-Wiki, the question arised, if we still want to maintain 2 locations for the images (used vector drawables and rawimages) or if it is sufficient to create the png on demand.

It was planned to use the svg directly in the DokuWiki, but that was not successful - see cgeo/cgeo-guide#40

So I see following options:

Store "original" image as svg in rawimages/svg: 
Advantage: 
* png can be created "automatically" / script based
* svg already available if downloaded as svg and imported into AndroidStudio

Disadvantage:
* no easy handling for usages of icons available in AndroidStudio / Vector assets -> additional svg-file needed 
* have to keep them in sync

Store icon as png in rawimages/png:
Advantage:
* png directly available

Disadvantage:
* no easy handling for usages of icons available in AndroidStudio / Vector assets -> additional png-file needed 

Don't store original icon in rawimages:
Advantage:
* no need to keep them in sync

Disadvantage:
* if used in Doku-Wiki, the png has to be created somehow

To discuss with @Lineflyer

### Is there an alternative?

_No response_

### Additional Information

see discussion in #11555

murggel:
## Description

remove svgs from raw-images  to minimize synchronisation effort, as they are used in the manual and stored there.

## Related issues

cgeo/cgeo-guide#40
fix #11562

reverts #11403


Lineflyer:
Thanks!

Issue #11668, opened 2021-09-15
Waypoint calculator changes preset values


Lineflyer:
Quick note from my vacation.
Can einrich wich screenshots and better description later if needed:

- Open a cache
- Swipe to waypoint tab
- Create a new waypoint 
- Tap on cache coordinates (to take them over as base for the calculation as i mostly do)
- Tap on calculate coordinates
- The calculator opens with prefilled coords (as selected in last step)
- Tap all decimal digits to replace them by variables (DD MM.MMM mode) (outcome is e.g.:  N34 56.ABC E123 45.DEF)
- Enter some formulas for A-F
- End the calculator and save the waypoint (to continue later while doing the cache)
- Tap on the prepared waypoint again
- Tap on Calculated coordinates to open the calculator again (e.g. you now want to note down the value you just found)

Result:
The first digits have changed from the previous values (= cache coords) to your current location
(Example: You are currently at N11 11.111  now the calculator shows N11 11.ABC instead of the correct N34 56.ABC as entered before)

Can someone please try to reproduce...that bug extended my todays hike by several kilometers =8-< and I thought I am to stupid to use c:geo

ztNFny:
repoducable in todays nightly

ztNFny:
I think this is happening here: CoordinatesCalculateDialog.java

    public void onCreate(final Bundle savedInstanceState) {
...
        gp = getArguments().getParcelable(GEOPOINT_ARG);
        if (gp == null) {
            gp = Sensors.getInstance().currentGeo().getCoords();
        }

pet0etie:
you can assign this to me

pet0etie:
PR has been created : https://github.com/cgeo/cgeo/pull/11707

moving-bits:
While testing #11707 (thanks @pet0etie for providing the PR) I've stumbled upon a different situation that fails:
- Open a cache
- Swipe to waypoint tab
- Create a new waypoint
- Enter some coordinates (which are very different from the cache's coordinates to make the problem more visible)
- Tap on calculate coordinates

The calculator opens with prefilled with current coords instead of the ones I have entered in the step before.

Not sure if this is some kind of regression as well (in that case the PR provided would only work as a workaround for the specific use case described at the top of this issue), or if this has never worked so far at all.

PR #11707, opened 2021-09-21
issue 11668 : re-initialize calculated waypoint correctly

pet0etie:
## Description

corrections in CoordinatesCalculateDialog.java
on second visit we enter with empty gp ... so the-not-formula-part of the coords are initialised with those of your current location
re-initialisation is now done correctly and depending on the saved coordinate-format

## Related issues

issue #11668

## Additional context

checkstyle-test : no remarks => ok
translations : none => ok

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
@@ -319,61 +319,6 @@ public class CoordinatesCalculateDialog extends DialogFragment implements ClickC
         return ccd;
     }
 
-    private String replaceNonNumericByZero(final String str) {
-        return str.replaceAll("[^0-9]", "0");
-    }
-
-    private Geopoint loadGpFromSavedState() {
-        Geopoint gpret = null;
-        final List<ButtonData> buttons = savedState.buttons;
-
-        switch (savedState.format) {
-            case Min:
-                // N 00 01 ° 02 03 . 08 09 10
-                // E 11 12 13 ° 14 15 . 20 21 22
-                gpret = new Geopoint(
-                    "" + savedState.latHemisphere,
-                    replaceNonNumericByZero("" + buttons.get(0).inputVal + buttons.get(1).inputVal),
-                    replaceNonNumericByZero("" + buttons.get(2).inputVal + buttons.get(3).inputVal),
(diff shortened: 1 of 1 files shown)


Lineflyer:
Thanks @pet0etie for your contribution.
Without having an idea of the involved code part, I wonder why the fix is so "extensive" while the bug showed up suddenly after recent changes. But maybe I am wrong and it never worked?

Remark regarding syntax:
You might want to use the keyword "Fix #11668" (ideally in your commit message) to allow auto-closing of the issue on merge.
Also when referring to an issue in your text just prepend a "#" to generate an auto-link.

ztNFny:
I have to admit I didn't look into the root cause of the problem, but the code in this PR seems to be very "scripty". Like using "" to initialize a string instead of making sure you get a string and the explicit indexes should be in a loop instead.
Applying different solutions in different cases seems to also be prone to cause future issues. I think there must be a more elegant solution to this issue.

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
```diff
@@ -329,6 +329,9 @@ public void onCreate(final Bundle savedInstanceState) {
         gp = getArguments().getParcelable(GEOPOINT_ARG);
         if (gp == null) {
             gp = Sensors.getInstance().currentGeo().getCoords();
+            if (savedState != null) {
+                gp = new Geopoint(savedState.plainLat, savedState.plainLon);
+            }
```
bekuno, line 334: If savedState != null then you initialize gp twice.
Better e.g.:
~~~
 if (savedState != null) {
    gp = new Geopoint(savedState.plainLat, savedState.plainLon);
} else {
    gp = Sensors.getInstance().currentGeo().getCoords();
}
~~~

pet0etie:
after analyzing the code again everything starts with a null-gp when getting back into the CoordinateCalculateDialog again
my first thought was rather correct : i could "retrieve the original coordinates" from the savedState
but i used the "button-values" so it was indeed a hard time to get good coordinates from there
the current solution uses the savedState too, but uses the "plainLat" and "plainLon" 
so i ended up in only 3 lines of code ...

main/src/cgeo/geocaching/ui/dialog/CoordinatesCalculateDialog.java
pet0etie: suggestion was changed in last commit/push

PR #11717, opened 2021-09-22
Fix #11716 - Sync changelog_release.md from release branch

### LINKED ISSUE #11716: changelog_release is out of sync
@moving-bits 
Coincidently I have seen, that `main/res/raw/changelog_release.md` is out of sync.
Theoretically it should be filled on `release` and (due to merge on `master`) be identical on `master`.

But for some reason the content is totally different. 
I did not investigate this yet, but what might have happened?

Could that explain the change log "problem" you have seen after merging branches?

Lineflyer:
## Description

Make `changelog_release.md` identical to `release` branch as it seems to have be manually changed on `master` which should not be done.

@moving-bits 
What about the two items in there. Are those on master or do they need to be (re)added?
Edit: Forget it, they are contained already.

## Related issues

Fixes #11716

### DIFF
--- a/main/res/raw/changelog_release.md
+++ b/main/res/raw/changelog_release.md
@@ -1,5 +1,18 @@
+### Map
+- New: Added Mapy.cz as online map source
+- New: Individual route: Allow setting a new start and reversing the route
+- New: Multiple offline map downloads can now be triggered at once
+
 ### Cache details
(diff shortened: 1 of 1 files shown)


moving-bits (approved):
Thanks for restoring the changelog, hope that this will fix the recurring conflicts within current release

PR #11724, opened 2021-09-25
Fix #11709 - Make name column in cache details zoomable

### LINKED ISSUE #11709: Forbid ellipsis (`...`) for Difficulty and Recent Logs rows
Basing on the current discussion inside #11676, we identified two rows with ellipsis (...) being shown in the german translation of the cache details page.

So for the two rows difficulty/Schwierigkeit and recent logs/Neuste Logs this setting should be disabled to gain the space for showing the complete line:

134245286-0c5ef856-c7c4-4d21-816b-b093f32cd60c

~~~
We can forbid the ellipsis (`...`) in general for this row to gain space. IMHO its better to show whats fitting instead of wasting three characters by three dots.

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/11676#issuecomment-924375979_

Lineflyer:
## Description

Use `sp` for width of the name column to make it expand/shrink in same relation as the text. This avoids cut off single line names in that column.
I left the value same, it might need some fine tuning if e.g. some translations are much longer. However due to the original design I guess it will fit in most occasions now.

## Related issues

Fixes #11709

### DIFF
--- a/main/res/layout/cache_information_item.xml
+++ b/main/res/layout/cache_information_item.xml
@@ -7,7 +7,7 @@
 
     <TextView
         android:id="@+id/name"
-        android:layout_width="90dip"
+        android:layout_width="90sp"
         android:layout_height="wrap_content"
         android:layout_alignParentLeft="true"
         android:layout_alignParentTop="false"


moving-bits:
Nice approach, let's give it a try

wolverine007:
Awesome - thanks for handling this topic

PR #11735, opened 2021-09-26
cross-convert cache markers to solve tinting issue

### LINKED ISSUE #11616: [vectorIcons] Wrong icon style in "Set cache icon"
Advanced > Set cache icon, the dialog shows the cache type icon in the upper right to set it to default. That currently shows a white icon on white background.

Should show colored background + white icon (same as #11592)

Current:
image

To be:
image
colored background + icon, no white marker

### LINKED ISSUE #11715: [vectorIcons] Wrong cache icon background color in "Create user-defined cache" dialog
Cache icon background should be yellow instead of orange

image

This is caused by (MaterialButton) dialogButton.setIconTint(null); in InternalConnector.java:interactiveCreateCache()

Fix should be the same as #11616

fm-sys:
fix #11715
fix #11616

grafik
grafik

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -755,7 +755,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         } else if (menuItem == R.id.menu_tts_toggle) {
             SpeechService.toggleService(this, cache.getCoords());
         } else if (menuItem == R.id.menu_set_cache_icon) {
-            EmojiUtils.selectEmojiPopup(this, cache.getAssignedEmoji(), cache.getType().markerId, this::setCacheIcon);
+            EmojiUtils.selectEmojiPopup(this, cache.getAssignedEmoji(), cache, this::setCacheIcon);
         } else if (LoggingUI.onMenuItemSelected(item, this, cache, null)) {
             refreshOnResume = true;
         } else {
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -995,9 +995,9 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         } else if (menuItem == R.id.menu_upload_bookmarklist) {
             BookmarkUtils.askAndUploadCachesToBookmarkList(this, adapter.getCheckedOrAllCaches());
         } else if (menuItem == R.id.menu_set_listmarker) {
-            EmojiUtils.selectEmojiPopup(this, markerId, 0, this::setListMarker);
+            EmojiUtils.selectEmojiPopup(this, markerId, null, this::setListMarker);
         } else if (menuItem == R.id.menu_set_cache_icon) {
-            EmojiUtils.selectEmojiPopup(this, -1, 0, this::setCacheIcons);
+            EmojiUtils.selectEmojiPopup(this, -1, null, this::setCacheIcons);
         } else if (menuItem == R.id.menu_set_askfordeletion) {
             setPreventAskForDeletion(false);
         } else {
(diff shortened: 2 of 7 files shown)


moving-bits:
Ah, double-conversion does the trick 👍 - I'd stopped after a single conversion yesterday, but did not find a method accepting a `Bitmap`... \*blush\*

Anyway. One thing I've stumbled upon, not sure if it's intentionally: 

If you create a new UDC, the icon has the hexagonal form (which is fine, as UDC connector has a hexagonal shape), but if you start selecting a user-defined icon for this new cache, it will have a round form in the EmojiUtils dialog.

Same happens if you are coming from a cache list. If you select a user-defined symbol, the cache marker is always round, independent from connector type.

If you want to change this, you could change the `EmojiUtils.selectEmojiPopup`'s parameter from `@Nullable CacheType` to `@Nullable Geocache`, so that you have the associated connector available.

fm-sys:
> If you create a new UDC, the icon has the hexagonal form (which is fine, as UDC connector has a hexagonal shape), but if you start selecting a user-defined icon for this new cache, it will have a round form in the EmojiUtils dialog.

Yes, this is due to the fact that I currently only use the CacheType Marker - not the real geocache marker. I totally agree that the real cache marker would be more fitting, however, is there a way to get the cache marker without overlays (@ztNFny)?

moving-bits:
> Yes, this is due to the fact that I currently only use the CacheType Marker - not the real geocache marker. I totally agree that the real cache marker would be more fitting, however, is there a way to get the cache marker without overlays (@ztNFny)?

In my tries yesterday I created a new temporary geocache in `EmojiUtils` using something like:
```
final Geocache tempCache = new Geocache();
tempCache.setType = cache.getType();
tempCache.setGeocode = cache.getGeocode();
```

with `cache` being the changed parameter of `EmojiUtils.selectEmojiPopup`.
Then you'll get a similar cache marker, but without overlays.

ztNFny:
What exactly do you need? The base shape (white ... with outline) and the background shape both are defined by the connector. Background tint and foreground are defined by cache type.

I've implemented a helper function for the filter dialog to retrieve the background (circle) + tint + icon.

ztNFny:
Clarificatiom: The difference between that helper function and the normal map marker is the white base shape. I don't think we need that in dialogs like "set icon" or filters.

fm-sys:
> white base shape. I don't think we need that in dialogs like "set icon" or filters.

Well, it would be helpful to have it, as especially the blue mystery background is otherwise nearly invisible in dark mode...

ztNFny:
Right, i forgot about dark mode. In light mode it looks great without the background.
Should it also be with base marker in filter dialog?

fm-sys:
> Should it also be with base marker in filter dialog?

I think for the filters dialog it's fine as it is.

IMHO that's quite consistent:
White base shape -> always connector speciffic
without base shape -> general cachetype marker

ztNFny:
Ok, so what about the implementation then? I'd suggest a (new) helper function which takes connector and type as arguments and provides the marker. Can do that tomorrow, sinilar to cachetype marker

fm-sys:
> Ok, so what about the implementation then? I'd suggest a (new) helper function which takes connector and type as arguments and provides the marker. Can do that tomorrow, sinilar to cachetype marker

Don't know whether it's worth a separate icon creation algorithm including caching etc. Maybe it's also enough to add @moving-bits suggested temporary geocache method to MapMarkerUtils:

> In my tries yesterday I created a new temporary geocache in `EmojiUtils` using something like:
> 
> ```
> final Geocache tempCache = new Geocache();
> tempCache.setType = cache.getType();
> tempCache.setGeocode = cache.getGeocode();
> ```

ztNFny:
@fm-sys I extended the existing method, could you just cherry pick https://github.com/ztNFny/cgeo/commit/b829ce31c4b3d464ee618e1d3dea36c408db36d6 ?

fm-sys:
It will now display the connector specific sharp (normal cache marker but without overlays)

grafik

Issue #11520, opened 2021-08-23
Extend support for calculated variables in cache


eddiemuc:
Last days I thought about how we can add better support for calculated values into c:geo especially for mysteries and multis. We already have the waypoint calculator but it has a few drawbacks:
* it can only calculate waypoints from variables. However I often have to calculate other things as well esp numbers to open locks, checksums and such
* It does not allow for usage of e.g. formulas. For example I often need to calculate Buchstabenwortwerte (what's the English word for it?) or checksums and have to leave c:geo for it
* It is somewhat deeply buried into the waypoint details which makes it difficult to use.

Therefore I propose the introduction of a new cache tab "Variables" which allows for adding variables and put a formula on them. Here's a sketch on how this may look:

![IMG_20210823_133035__01.jpg](https://user-images.githubusercontent.com/6909759/130441186-ff781d32-f046-48ba-88f3-f881a6663876.jpg)

Additional ideas:
* Variables created here could be used in waypoint calculator of course
* A scan could be implemented trying to parse formulas/variables from listing and adding them here. Might be a heuristic approach where user can decide whether a result is correct or not before it is added. (A 100% correct scan is probably impossible to implement)
* Functions available for usage in formulas must be discussed. Whatever set is chosen it can easily be extended later.

Idea is also to make the entries editable directly (no "open" step necessary) to ease their usage in the field.

Some more details, probably not so important right now:
* No strong typing of variables. Current idea is to support text and int as values in the variables. If something looks like an int it is interpreted as such, otherwise as text. Is float support necessary?
* Variable names are Case-sensitive and start with a letter. Can be composed of letters and digits otherwise
* In a formula, every letter-block is interpreted as a constant text, except if a variable of same name exists (then it is interpreted as the variable of course). Optionally text can be enclosed by " or '

What do you think?

eddiemuc:
It seems the idea is supported (three thumbs-up), I may make this my next "big project" after cleanup of filters is done

eddiemuc:
Notes to self:
* In a later step, one could add support for "unknown varlables". E.g. "for unknown var E, iterate from 0 to 9 and create all possible waypoints as calculated by formula"...
* To be decided whether this is worth introduction of something like ANTLR or handleable with the expression parser for filters / calculated waypoints

eddiemuc:
Working on this feature I fear I thought too complicated. I will try to rethink and separate between functions and integer calculation explicitly. So a variable will be:
[...]

PR #11739, opened 2021-09-26
rel to #11520: variables: introducing CalculatorMap

eddiemuc:
rel to #11520:  variables: introducing CalculatorMap


bekuno:
retest this please

Issue #11737, opened 2021-09-26
Give workflow suitable permissions to push to master


fm-sys:
The workflow console looks perfect, seems like everything had worked as it should. However, the GitHub actions token has no suitable permission to push torwards master. It seems we need to adapt the repository config to make the workflow run successful... 

In the GitHub support community forum, it was suggested to create a new secret inside the repository settings. (…/settings/secrets). Then we can use it inside the workflow file:
```
${{ secrets.GITHUB_TOKEN_OURS }}
```
https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/23

```
Merge made by the 'recursive' strategy.
 main/src/cgeo/geocaching/utils/SystemInformation.java | 2 ++
 1 file changed, 2 insertions(+)
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more information.        
To https://github.com/cgeo/cgeo.git
 ! [remote rejected]     master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/cgeo/cgeo.git'
Error: Process completed with exit code 1.
```

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11733#issuecomment-927311590_

Lineflyer:
I can create a repository secret (as you described) in that menu.
What next...do you need this secret or does it simply have to be there?

fm-sys:
You don't need to share it with me. It's enough to add the secret in the settings and share the variable name of it.

Then we need to replace the Github token in the workflow script with the variable name of the newly created one. This ensures that the security critical token can be kept private while the script itself is public anyway.

https://github.com/cgeo/cgeo/blob/e116e23c2d3a04544f56bdde57212015e554b131/.github/workflows/branch-merge.yml#L12

(replace `GITHUB_TOKEN` with with the new secret)

Lineflyer:
I created one: ACTION_SECRET

fm-sys:
> 
> 
> I created one: ACTION_SECRET

@Lineflyer on behalf of which user the `ACTION_SECRET` was created? Probably we have to change `https://cgeo:${{secrets.ACTION_SECRET}}@github.com/...` to `https://THE_USER_NAME:${{secrets.ACTION_SECRET}}@github.com/...`

[...]

PR #11746, opened 2021-09-28
give workflow permission to push to master

fm-sys:
see #11737


Lineflyer:
As we currently have a commit to merge from `release` to `master` that would be the right time to test it out.

@fm-sys Go for it!

Issue #11721, opened 2021-09-23
Icon feedback - Earth cache, CITO


Lineflyer:
A few days ago I posted a sneak peak screenshot of the new icons to Facebook.

Now I tried to summarize the feedback and identify the top suggestions/critical feedback.
AFAICS it concentrates on three things:

1. A major percentage of users provided very positive feedback, only a few which liked the old multi-color icons more and think the new ones are to boring. But this is a clear minority in this non-representative survey anyway.
2. Earth cache icon: Many users provided feedback, that the earth cache icon is a) not easily recognizable and/or b) not easily distinguishable from virtual caches. Some users mentioned, that the yellow/orange core was helpful in the old icon.
3. Multi cache icon: Some users provided feedback, that they don't understand the icon respectively think its to far away from the old icon and/or the icon used in the website and in other apps.

@ztNFny 
What do you think regarding this feedback.
AFAIR you had multiple proposals for multi cache. Not sure about earth cache...

ztNFny:
Took a look at the Facebook comments and there's 3 kinds of comments:
Earthcache icon: 4 comment with a total of 16 upvotes
More / Different background colors: 2 comments with a total of 6 upvotes
Multi: 1 comment, no upvote

So let's consider EC and colors.

EC: I tried to keep the general scheme of having only white for the icon itself. If we want to stick with that we can only tune the icon. I looked into some icons for "geology", like a pickaxe or a rock, but nothing really says "earthcache" to me there. So an "earth" is probably best suited to convey "earthcache". Going with something like the original EC icon with the cutout simply doesn't work in monocolor either, you'll just get something that looks like a 3/4 circle.

Background colors They are currently grouped by whether/where the logbook is:
- Green = at listing -> Tradi
- Orange = multiple waypoints -> Multi
- Dark Blue = you have to do something to get the logbook -> Mystery, WIG, Letterbox
- Light Blue = no logbook but (usually) photo -> Virtual, EC, Webcam
- Red = Event
- Light green = special/rare stuff -> HQ, APE, Maze
- Yellow = user defined
Of those categories imho the only one that isn't clear is Letterbox - that could be assigned to orange or dark blue. With a green-field approach I'd have used orange as it's closer to a multi, but GC is using blue and I think it's best to be aligned with the biggest platform there.

But as the color is majorly criticized in the context of the EC icon maybe it's not really 2 issues but only one -> people are used to the EC icon being colorful and standing out. So we have to ask ourselves the 1 question: 

[...]

PR #11748, opened 2021-09-29
new earthcache icon

ztNFny:
related to but no solving #11721, so I'm not linking to that to avoid it being closed


Lineflyer:
Thanks @ztNFny.
Lets give it a try for the next beta with this icon and collect some feedback.
IMHO its more recognizable than the previous one.

PR #11760, opened 2021-10-02
allow HTML in navigation target description (fix #11730)

### LINKED ISSUE #11730: Navigation target select does not support html descriptions
### Describe your problem!

From support mail:
If the waypoint note is HTML the helper window, which is shown to select whether you want to navigate to the cache coordinates or parking coords does not properly support HTML.

In cases where the waypoint note is needed to select a desired navigation target, this can lead to problems.

### How to reproduce?

- Open GC75KBQ
- Select Navigate - Navigate (Drive)

The "Select target" window is shown and offers correctly the cache coordinates and the three available parking waypoints as possible target.
However in this specific case the owner assigned the same names for all parking waypoints (default at geocaching.com on creating them) but only the waypoint description contains the important differences.
These are however not shown but only `<html>...` is shown:
Screenshot_20210926-120730

The (first line of) the waypoint description should be shown / HTML should be supported

### Reproducible

Yes

### c:geo Version

2021.09.08

### System information

_No response_

### Additional Information

Support Ticket 572804

moving-bits:
## Description
Enable HTML in navigation target dialog for waypoint descriptions

image

### DIFF
--- a/main/res/layout/cacheslist_item_select.xml
+++ b/main/res/layout/cacheslist_item_select.xml
@@ -42,7 +42,6 @@
         android:layout_marginLeft="43dip"
         android:layout_marginTop="30dip"
         android:ellipsize="end"
-        android:lines="1"
         android:paddingBottom="5dip"
         android:paddingRight="3dip"
         android:scrollHorizontally="true"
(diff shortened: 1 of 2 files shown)


Lineflyer:
Your screenshot shows the follow up problem in this example, that the text is truncated and does still not provide useful information.
Maybe we should have the description in a new line below the codes?

But that can also be done in a dedicated issue.

moving-bits:
I've updated the PR to allow for multi-line info in target selector:

image

PR #11764, opened 2021-10-02
rel to #11520: add first version of variables/calculator tab

eddiemuc:
rel to #11520: add first version of variables/calculator tab

### DIFF
--- a/main/res/drawable/ic_menu_function.xml
+++ b/main/res/drawable/ic_menu_function.xml
@@ -1,0 +1,11 @@
+<!-- taken from ic_baseline_functions_24 / material.io -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24"
+    android:tint="?android:textColorPrimary">
+  <path
+      android:fillColor="@android:color/white"
+      android:pathData="M18,4H6v2l6.5,6L6,18v2h12v-3h-7l5,-5 -5,-5h7z"/>
+</vector>
--- a/main/res/layout/cachedetail_variables_page.xml
+++ b/main/res/layout/cachedetail_variables_page.xml
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/activity_viewroot"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical"
+    android:layout_marginTop="10dip"
+    android:nestedScrollingEnabled="true"
+    android:fastScrollEnabled="false"
+    android:scrollbarStyle="outsideOverlay"
+    tools:context=".VariablesViewPageFragment">
+
+    <LinearLayout
+        android:layout_width="fill_parent"
+        android:layout_height="wrap_content"
+        android:orientation="vertical" >
+
+        <TextView
+            android:id="@+id/variables_experimental_warning"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:textSize="@dimen/textSize_headingSecondary"
+            android:background="@drawable/mark_orange"
+            tools:text="Experimental Feature\n* point one\n* point two"
+            />
+
+
+        <LinearLayout
+            android:layout_width="fill_parent"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal">
+
+            <Button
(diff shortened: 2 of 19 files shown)


eddiemuc:
Early version, but may already be reviewed. 
image

eddiemuc:
Calculator formula syntax is described here: https://github.com/cgeo/cgeo/wiki/Calculator-Formula-Syntax

moving-bits:
Looks very promising, looking forward to it!

moving-bits:
> Calculator formula syntax is described here: [cgeo/cgeo/wiki/Calculator-Formula-Syntax](https://github.com/cgeo/cgeo/wiki/Calculator-Formula-Syntax)

Thanks for this extensive documentation. Some questions / remarks:
- Does the factorial operator have to be prepended? Mathematical way for factorial is after the number (`4!`).
- How are parameters in functions separated? You describe it as `;`, but the example is using `,`. (Either way would be ok, IMHO. Maybe even support both? Not sure how other countries/languages handle parameter lists.)
- Not sure about this one: Do variable names have to be case-sensitive? May be less error-prone if they are case-agnostic, but there may be listings which distinguish between upper-case and lower-case vars.
- Can we make variable names allow for alphanumeric charset, allowing var names like `x1` or `b3` - I know of caches using such variable names.

eddiemuc:
> > Calculator formula syntax is described here: [cgeo/cgeo/wiki/Calculator-Formula-Syntax](https://github.com/cgeo/cgeo/wiki/Calculator-Formula-Syntax)
> 
> Thanks for this extensive documentation. Some questions / remarks:
> 
> * Does the factorial operator have to be prepended? Mathematical way for factorial is after the number (`4!`).

You're right, this is an error in documentation (in implementation it has indeed to be behind the number). I fixed this.

> * How are parameters in functions separated? You describe it as `;`, but the example is using `,`. (Either way would be ok, IMHO. Maybe even support both? Not sure how other countries/languages handle parameter lists.)

Another erro in documentation, sorry. Currently it uses `;`. I tried to mimic syntax of Excel which also uses semicolon. I thought about comma too, but I actually wanted to avoid confusion with decimal separators in other languages... (as of now, only a point (`.`) is allowed as decimal separator)

> * Not sure about this one: Do variable names have to be case-sensitive? May be less error-prone if they are case-agnostic, but there may be listings which distinguish between upper-case and lower-case vars.

Yeah, I am not entirely sure about this myself... But as of now c:geo itself uses case-sensitive variable names in Calculated Coordinates (a is another var than A), so I thought I keep this. Can be discussed of course.

> * Can we make variable names allow for alphanumeric charset, allowing var names like `x1` or `b3` - I know of caches using such variable names.

I thought long and hard about this. The main problem with this are connected variable expressions. Meaning if you have a variable `A` with value `2`, then you want the expression `AA` to evaluate to `22` and the expression `AA(A+2)` to evaluate to `224`.
But if you allow variables with numbers, imagine you have an expression `A1A` and variables `A=2` and `A1=3`. Would this now evaluate to `212` or to `32`? Even if you find a decision for this specific case, we need some sort of common (and easy parseable!) rule for this. Even if you use $ (which I introduced to solve a similar problem with multi-character-varnames), the expression $A1A would be ambigious.
Also note that every rule we find must be such that the variables in an expression can be identified at parse time already (and not depending on whatever variables are defined). Otherwise parsing gets very very ugly...  

But yes, this (and also the other points) are very valid discussion topics. Maybe something for next dev call?

moving-bits:
Thanks for your detailed reply. Sounds all very logical to me. 

For the alphanumeric vars: As both ways exist "in the wild world", I guess, we simply have to decide which way to support, as they are mutually exclusive. (Or we could force spaces as separators between vars and constants. Not sure what's out there in real cache descriptions...)

For parameter lists let's stay with `;` then, at least I know of no numbering system using `;` as decimal separator. Supporting a locale specific separator can get ugly on importing/exporting data, so let's stay with a universal one.

eddiemuc:
@moving-bits over night I rethought about var names with digits. And I was wrong, using the $ syntax for variable names with more than one char, digits in var names are easily possible. Thus thanks a lot for the question, I will add that support (is on my to-do list now :-))

In my example above ( assume var `A=2`, var `A1=3`):
 * `A1A` would unambiguous evaluate to `212` (vars without $ are treated as single-char-varnames) 
* `$A1A` would result in "missing var value" error (it would search for a var with name `A1A` which is not defined)
* To use `A1` in a concat expression user has to write something like `$A1(A)` or `($A1)A` (would both evaluate to `32`)

eddiemuc:
> Or we could force spaces as separators between vars and constants. Not sure what's out there in real cache descriptions...

I most often find things like `N43 23.AB(B-1)` esp in coordinate patterns. This is where the concat stuff specifically aims at. Most of the time there's no space in-between

eddiemuc:
I implemented now a switch which will make the variable tab ONLY visible when debugmode is enabled. I also added a "experimental feature" description on top of the tab. I consider the PR mergeable with these changes:

image

image

PR #11768, opened 2021-10-02
fix icon style (fix #11732)

### LINKED ISSUE #11732: [vectorIcons] Wrong icon style in Cache/Waypoint selector
### Describe your problem!

Screenshot_20210926-120730

### How to reproduce?

- Open GC75KBQ
- Select Navigate - Navigate (Drive)

Wrong icon style in Cache/Waypoint selector

Show normal cache marker (like in cache lists)

### Reproducible

Yes

### c:geo Version

2021.09.08

### System information

_No response_

### Additional Information

_Extracted out of https://github.com/cgeo/cgeo/issues/11730_

fm-sys:
before|now
-|-
grafik|grafik

### DIFF
--- a/main/src/cgeo/geocaching/NavigateAnyPointActivity.java
+++ b/main/src/cgeo/geocaching/NavigateAnyPointActivity.java
@@ -2,6 +2,7 @@ package cgeo.geocaching;
 
 import cgeo.geocaching.activity.AbstractActionBarActivity;
 import cgeo.geocaching.connector.internal.InternalConnector;
+import cgeo.geocaching.enumerations.CacheListType;
 import cgeo.geocaching.enumerations.LoadFlags;
 import cgeo.geocaching.enumerations.WaypointType;
 import cgeo.geocaching.list.StoredList;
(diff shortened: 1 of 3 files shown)


fm-sys:
retest this please

moving-bits:
Looks good generally, but would it be possible to have the waypoint icons without the little "pin" at the bottom?

fm-sys:
> Looks good generally, but would it be possible to have the waypoint icons without the little "pin" at the bottom?

I'm unsure about this one, but however it has IMHO nothing to do with this PR. It's the same in `NewMap` in our current nightly:

So if we want to change this, it should IMHO go into an dedicated issue. 

Screenshot_20211003_092610_cgeo geocaching_copy_324x702

fm-sys:
BTW, the pin does IMHO even make sense. While the cache doesn't need the be at the listed coords, waypoints are always fixed (as long as they have coordinates). So this pin marker helps to quickly recognize whether it's a cache or waypoint...

moving-bits:
IMHO the pin makes sense on the map only, for pointing to a specific location on the map. I would leave it out in all other places. But I agree that this would fit for a separate issue.

Issue #11696, opened 2021-09-20
Unify map zoom controls 


fm-sys:
### Tell us your idea!

Currently we have two different zoom controls:

![Screenshot_20210920_113428_cgeo geocaching](https://user-images.githubusercontent.com/64581222/134013196-1c132886-3761-45af-80dd-5f4fd21771b5.png)
![Screenshot_20210920_113300_cgeo geocaching](https://user-images.githubusercontent.com/64581222/134013197-17ed2912-fb73-4617-9848-935ef05835fb.png)

### What solution would you suggest?

We could replace both with two FAB's (floating action buttons)

![Screenshot_20210920_154449_org mozilla firefox](https://user-images.githubusercontent.com/64581222/134013343-a882512e-a5f5-424c-9ab9-c99ece8f3f04.png)
_Image copied from  #8057_

ztNFny:
Of the 3 screenshots I like Google Maps' buttons design best. Transparency that allows visibility of what's below + clean design.

Lineflyer:
Indeed, the transparency is nice..not sure though if we can reuse the same buttons on the other maps.

moving-bits:
Maybe we can introduce our own buttons for both map types. GM's style is the best of the three shown here, IMHO - it's clear, not too much detail. Maybe we can combine it with our iconized button style for even more consistency.

If we switch to such a button arrangement, this could also help with placing our new quick buttons (see #11771), as it frees some space at the bottom. We could then place the new quick buttons at the bottom as well, for both map types, which BTW would also make the animations (bottom sheet / popup menu) more logical.

moving-bits:
So far I have been able to change the zoom controls for OSM to this:

|OSM (new)|GM|
|---|---|
|![image](https://user-images.githubusercontent.com/3754370/135752083-4d5813d5-99bc-489f-a0e6-fbcc121ccf00.png)|![image](https://user-images.githubusercontent.com/3754370/135752114-6ac72aa3-e6f5-44a2-ae14-bc454fa9580f.png)|

(My GM is broken, so I cannot show you transparency level for GM zoom controls for comparison.)

Still trying to figure out to move the zoom controls a bit up + left, to get them close to GM controls. Not sure if it's possible, though.

fm-sys:
> I have adapted the zoom controls for OSM to look more similar to those of GM

I would give them a bit more alpha, IMHO they look too transparent in the screenshot. But looks great already 👍

moving-bits:
[...]

PR #11771, opened 2021-10-03
introduce PopupMenu for individual route items (fix #11531)

### LINKED ISSUE #11531: Make sorting of Individual route easier
### Tell us your idea!

Currently, it is really annoying to dive deeply into sub menus for resorting the individual route.

Screenshot_20210825_171545_cgeo geocaching

I suggest making the individual route clickable, like it is done in locus. But instead of showing a big list of options, just show one directly clickable entry, "edit individual route". 

Screenshot_20210825_171445_menion

While we are on it, I think we can move the "delete individual route" into the sorting screen as well. (Similar to what we have in the filter framework)

Screenshot_20210825_171801

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
- Move individual route related menu items to a popup menu
- Move quick settings icon in map to a new button at the right, centered vertically
- Popup for individual route has its own button below the moved quick settings button

image.image


Lineflyer:
This takes much more space than the old gear icon. Not sure if I like it.

Is the routing icon shown all the time or only if a route exists? Please also think about, the probably only a minor percentage of users is using the route function.

moving-bits:
> This takes much more space than the old gear icon. Not sure if I like it.

I can adapt it, if requested. So far I've chosen to style it similar to our iconized buttons.

> Is the routing icon shown all the time or only if a route exists? Please also think about, the probably only a minor percentage of users is using the route function.

Both buttons are shown permanently. Not only to promote individual routing, but without the button being shown permanently, I would have to split some of the functions (e. g. "load individual route") between the overflow menu and the popup menu, which I find confusing.

moving-bits:
Regarding button size see this comparison: Popup buttons and the new quick settings buttons have the same size:

image

Lineflyer:
I wanted to test it...not sure if its related or just a problem with my AVD:

I can't get it working...seems I cant dismiss the "Long tap on map" message and the app is stuck.

image

Lineflyer:
> Not only to promote individual routing, 

There are many many users just using c:geo to find a cache icon on the map and go logging this cache. IMHO we might annoy such users with "shrinking" the map just to display a feature button they probably never use.

Also think about devices with smaller screens:
There have been already users claiming the map got smaller by the filter bar, which is now shown. It will get worse, once we have the bottom navigation. And now we start also shrinking on the vertical side.

Still a bit reluctant about this change..but maybe others can post their opinions.

Lineflyer:
> I wanted to test it...not sure if its related or just a problem with my AVD:

Forget about that. It seems to be a local problem on my AVD...its damn slow.

Lineflyer:
As a compromise (just reading #11696): If those button could have the dimensions of the GMaps zoom controls it would be much less intrusive.

ztNFny:
Why display these as overlay instead of normal toolbar buttons? I'm not a big fan of putting stuff over the map, the settings button was small enough to not get into the way, but if it's getting more I can see how it gets in the way of things.

moving-bits:
Toolbar is already too crowded, leaving not enough space for the text field to be of much use (depending on screen resolution and font size that is).
With #11776 I'll adapt OSM zoom controls to be more similar to GM zoom controls, which makes room at the bottom for a second button. Whether this will be displayed permanently or only when an individual route is active can still be discussed.

moving-bits:
Regarding the individual route: This is meant for planning your cache trip, and using navigation to lead you from cache to cache. 

Your workflow can be like this:
- before your tour:
  - do planning of a route at home, adding caches/waypoints by long-tapping on them
  - if necessary: sort individual route as required
- during your tour:
  - activate "set route start as target"
  - drive to the first cache/waypoint
  - long tap on that cache/waypoint to remove it from route
  - due to activated "set route start as target", the next cache/waypoint will be set as navigation target automatically
  - repeat last steps until your last cache/waypoint has been reached

ztNFny:
> Toolbar is already too crowded, leaving not enough space for the text field to be of much use (depending on screen resolution and font size that is).

I only get 2 toolbar icons in landscape here? There'd be a lot more space available to use. If the list title gets a lot smaller I wouldn't care at all, only the "X caches" should stay visible.
signal-2021-10-03-152207_001

> Regarding the individual route: This is meant for planning your cache trip, and using navigation to lead you from cache to cache.

I looked into it, but that seems overly complicated. What I do is create a list of the caches I want to visit, select the next one from the list, use "map" to draw a line to it, then just select the next cache and repeat. The time I'd have to invest in preparing the route is way more than I'd be able to save - and I'd have to 100% stick to a plan instead of just looking at the map for the next one. I see how that feature can be useful to some, but for me it just doesn't fit.

moving-bits:
Depending on how you've opened the map there will also be a "filter" icon in the toolbar. Also take into account portrait mode.

Anyway. I'll move the buttons to the bottom, where they will somewhat similar to scale etc. in height, thus not taking up more precious space.

ztNFny:
> Depending on how you've opened the map there will also be a "filter" icon in the toolbar. Also take into account portrait mode.

Nope, that appears only in landscape mode (where I also get the live map toggle). I never get filter in toolbar, it's always the first item in overflow menu.

moving-bits:
For me it looks like this, when coming from main screen => live map:

image

ztNFny:
> For me it looks like this, when coming from main screen => live map:

True, with live map the filter button is visible by default, with lists it isn't, wonder why. (I never ever use live map, I'm 100% offline.)

But even in your screenshot an additional button would fit in the toolbar. Or the map switch could be integrated into the map settings dialog.

fm-sys:
> As for routes, i haven't found any use for them so far, so an icon thats always over the map seems too intrusive there.

I agree that the button doesn't make much sense if there is no individual route set up. But if one is using that feature, the button will have a major advantage... So I would vote for showing it dynamically.

>image

Can we make the buttons appear in the same size / color style as the (new) zoom controls?

---

Also, we could refactor the `RouteSortActivity` to a bottomSheet for consistency. The bottomSheet can have a normal action bar where we can put all owerflow menu items in. As main content we will have the DragSortListview. A already have ideas about that, but that's another completely different topic... ;-)

eddiemuc:
> > For me it looks like this, when coming from main screen => live map:
> 
> True, with live map the filter button is visible by default, with lists it isn't, wonder why. (I never ever use live map, I'm 100% offline.)
> 
> But even in your screenshot an additional button would fit in the toolbar. Or the map switch could be integrated into the map settings dialog.

The logic to display the filter icon is very handcrafted/hacky, Google thinks that it should never be displayed. Due to this hackyness there may we'll be room for improvements :-)

moving-bits:
> So I would vote for showing it dynamically.

already working on it...

> Can we make the buttons appear in the same size / color style as the (new) zoom controls?

yes, as soon as #11776 is merged

fm-sys:
> > > For me it looks like this, when coming from main screen => live map:
> > 
> > 
> > True, with live map the filter button is visible by default, with lists it isn't, wonder why. (I never ever use live map, I'm 100% offline.)
> > But even in your screenshot an additional button would fit in the toolbar. Or the map switch could be integrated into the map settings dialog.
> 
> The logic to display the filter icon is very handcrafted/hacky, Google thinks that it should never be displayed. Due to this hackyness there may we'll be room for improvements :-)


It's a material design guideline, that action bar icons shouldn't take more than 50% of free space and should show 5 icons max. This is handled by android automatically. We BTW already use a hack to show the fourth icon (filter) as mentioned by @eddiemuc.

ztNFny:
> It's a material design guideline, that action bar icons shouldn't take more than 50% of free space and should show 5 icons max. This is handled by android automatically.

In my screenshot (https://github.com/cgeo/cgeo/pull/11771#issuecomment-932951556) that'd mean there's room for another 2 icons though?

moving-bits:
> > It's a material design guideline, that action bar icons shouldn't take more than 50% of free space and should show 5 icons max. This is handled by android automatically.
> 
> In my screenshot ([#11771 (comment)](https://github.com/cgeo/cgeo/pull/11771#issuecomment-932951556)) that'd mean there's room for another 2 icons though?

No, IIRC default is only 3 buttons. "5" is a typo, IIRC. Our fourth button is already hacked in.

moving-bits:
Ok, updated the PR to enable / disable the second button dynamically. Also reduced button size, but that will change again once #11776 is merged and I will adapt the icons here to the new OSM zoom controls.

image.image

Lineflyer:
That's IMHO a much better approach (show dynamically and align sizes between all buttons incl. zoom controls).

moving-bits:
> That's IMHO a much better approach (show dynamically and align sizes between all buttons incl. zoom controls).

Nah, button size will be aligned on map only - our standard buttons in map popup etc. are much larger. But I remember that the core team had some discussions back then when we introduced the material style icon buttons...

Anyway, I'll start adapting above's black buttons to the new OSM zoom control buttons, now that #11776 has been merged and styles and drawables are available on master.

Lineflyer:
> > That's IMHO a much better approach (show dynamically and align sizes between all buttons incl. zoom controls).
> 
> Nah, button size will be aligned on map only - our standard buttons in map popup etc. are much larger. But I remember that the core team had some discussions back then when we introduced the material style icon buttons...
> 

Oh, thats what I meant. Just for the map of course. If the buttons for quick settings and routing have the same size as one of the zoom buttons that is a good compromise and acceptable for the space its consuming.

moving-bits:
This is how it looks now:

image.image

Lineflyer:
Now it looks nicely embedded into the other controls.
Thanks @moving-bits for the willingness to rework it 👍🏼

moving-bits:
Always striving for the best solution :-) Thanks for the team's feedback!

Lineflyer:
BTW: You should probably obtain a personal GMaps key so your emulator also works with that ;)

eddiemuc:
> > > > For me it looks like this, when coming from main screen => live map:
> > > 
> > > True, with live map the filter button is visible by default, with lists it isn't, wonder why. (I never ever use live map, I'm 100% offline.)
> > > But even in your screenshot an additional button would fit in the toolbar. Or the map switch could be integrated into the map settings dialog.
> > 
> > The logic to display the filter icon is very handcrafted/hacky, Google thinks that it should never be displayed. Due to this hackyness there may we'll be room for improvements :-)
> 
> It's a material design guideline, that action bar icons shouldn't take more than 50% of free space and should show 5 icons max. This is handled by android automatically. We BTW already use a hack to show the fourth icon (filter) as mentioned by @eddiemuc.

Just in case if this turns into a new feature request:
* The "hacky" method to force-add more items to the action bar than wanted by Google/Android is `ViewUtils.extendMenuActionBarDisplayItemCount`
* The rules decoded by Google/Android itself can be found here: `androidx.appcompat.view.ActionBarPolicy#getMaxActionButtons`
* The hackis currently only triggered when opening the live map (regardless of whether it showns only stored caches or is really live). It is not triggered anywhere else, and it is also not triggered when the map is opened in any other way (e.g. by opening the map for a single cache or displaying a stored list on the map)

moving-bits:
> BTW: You should probably obtain a personal GMaps key so your emulator also works with that ;)

It did work until a couple of months ago it stopped working from one day to the next. Have been too lazy yet to find the cause for it...

PR #11776, opened 2021-10-03
adapt styling of OSM zoom controls (fix #11696)

### LINKED ISSUE #11696: Unify map zoom controls

moving-bits:
## Description
Adapt styling of OSM zoom controls to look more similar to GM zoom controls:
- use vertical arrangement (zoom in on top, zoom out below)
- use +/- symbols as zoom controls
- use semi-transparent white box as background
- disable auto-hide of zoom controls

image

### DIFF
--- a/main/res/drawable/zoomin.xml
+++ b/main/res/drawable/zoomin.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:drawable="@drawable/white_bcg"/>
+    <item
+        android:drawable="@drawable/zoomin_fg"
+        android:top="3dp"
+        android:bottom="3dp"
+        android:left="3dp"
+        android:right="3dp"
+        android:gravity="center"/>
+</layer-list>
--- a/main/res/drawable/zoomout.xml
+++ b/main/res/drawable/zoomout.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:drawable="@drawable/white_bcg"/>
+    <item
+        android:drawable="@drawable/zoomout_fg"
+        android:top="3dp"
+        android:bottom="3dp"
+        android:left="3dp"
+        android:right="3dp"
+        android:gravity="center"/>
+</layer-list>
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -137,6 +137,7 @@ import org.mapsforge.core.model.MapPosition;
 import org.mapsforge.core.util.Parameters;
 import org.mapsforge.map.android.graphics.AndroidGraphicFactory;
 import org.mapsforge.map.android.graphics.AndroidResourceBitmap;
+import org.mapsforge.map.android.input.MapZoomControls;
 import org.mapsforge.map.android.util.AndroidUtil;
 import org.mapsforge.map.layer.Layers;
 import org.mapsforge.map.layer.cache.TileCache;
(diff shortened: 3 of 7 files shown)


main/res/drawable/zoomin.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:drawable="@drawable/white_bcg"/>
+    <item
```
fm-sys, line 5: It's always a good idea to give visual touch feedback. Maybe adding a new layer `<item android:drawable="?android:attr/selectableItemBackground"/>` between background and icon?

GMap is doing that as well:
grafik

main/res/drawable/zoomout.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:drawable="@drawable/white_bcg"/>
+    <item
```
fm-sys, line 5: here as well...

main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
```diff
@@ -279,6 +280,15 @@ public void onCreate(final Bundle savedInstanceState) {
...
         mapView.setBuiltInZoomControls(true);
 
+        // style zoom controls
+        final MapZoomControls zoomControls = mapView.getMapZoomControls();
+        zoomControls.setZoomControlsOrientation(MapZoomControls.Orientation.VERTICAL_IN_OUT);
+        zoomControls.setZoomInResource(R.drawable.zoomin);
+        zoomControls.setZoomOutResource(R.drawable.zoomout);
+        final int padding = ViewUtils.dpToPixel(10.0f);
+        zoomControls.setPadding(padding, padding, padding, padding);
+        zoomControls.setAutoHide(false);
```
fm-sys, line 290: Should we really disable `AutoHide`? I know that's the way how GMap behaves, but is it really preferable?

main/res/drawable/zoomin.xml
moving-bits: Thanks for reviewing this PR.

I've tried adding the selectableItemBackground to the `zoomin.xml` in all three possible positions, but it does not seem to make any difference - no visual effects on hitting the "+" sign.

main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
moving-bits: To be discussed...

Pro: consistency to GM; also I find it kind of annoying having to hit the map twice to zoom. Since new zoom controls are smaller than the old one, hiding too much map space should not be a problem anymore.

Con: Some people may want to see every inch of map space (due to extremely small screens etc.)

main/res/drawable/zoomin.xml
fm-sys: Hm... At other places in c:geo `selectableItemBackground` works fine... 🤷‍♂️

fm-sys: Maybe it does only work for layout files, not for drawables.

There are other possibilities, see https://stackoverflow.com/questions/43284336/shaped-drawable-with-selectableitembackground-as-background

But maybe it's not really worth the effort...

moving-bits: Finally... Got to make it work, thanks for the link! `ripple` needed to be applied to the background XML, and the ripple color to be adjusted to `@android:color/black` instead of `?colorControlHighlight`.

main/res/drawable/zoomout.xml
moving-bits: As the ripple is applied to the background XML, it will work for both zoom buttons.

main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
moving-bits: Let's stay with `setAutoHide(false)` for now and wait for user feedback.

Lineflyer: I also would prefer auto hiding simply to avoid the problem, that I cannot click a cache below the zoom controls.

fm-sys: > that I cannot click a cache below the zoom controls.

You can simply move the map to reach the cache...

Lineflyer: or the other way around on GMaps: 
I cannot zoom but it opens the cache....but anyway you are right: Both kind of a luxury problem.

Issue #11591, opened 2021-08-29
[vectorIcons] Wrong icon style in search dropdown


ztNFny:
### Describe your problem!

all places that use icons need to build the icon from multiple layers (filter dialog, search dropdown, ...)
There'll probably some slight difference in the icon display depending on the place:
- Map = with pin
- list/search = with (circle) marker but without pin
- filter dialog = without marker (should be enough, needs to be tested)

![image](https://user-images.githubusercontent.com/1258173/131254169-f4d97e69-7f0e-4349-a2ac-b55e253aeaa8.png)

### How to reproduce?

-

### Reproducible

Yes

### c:geo Version

vectorIcons build

moving-bits:
Does someone know where the search results are created?

ztNFny:
> Does someone know where the search results are created?

SuggestionProvider:getLastOpenedCaches()

moving-bits:
Thanks. If a read the documentation correctly, the only thing we can return is a resource id (which is set to the cache type's marker id currently).

See https://developer.android.com/guide/topics/search/adding-custom-suggestions and https://stuff.mit.edu/afs/sipb/project/android/docs/reference/android/app/SearchManager.html#SUGGEST_COLUMN_ICON_1 for docs, and the code in `SearchSuggestionCursor.java`.

To me it looks like we need to create drawables with icon and colored background (round only) in a single xml file for each type.

This will probably also help with #11592 and in other, similar places, as we can reference a resource id then for a default marker, instead of having to build a layered drawable first.

ztNFny:
Isn't it possible to extend/replace that code to be able to use LayerDrawables?

If it isn't we'll have to generate 20 (types) * 3 (connector shapes) * 2 (with / without background) drawables - in addition to the 20 (icons) + 3 (backgrounds) + 3 (connectors) + 9 (colors) existing ones. Will be a PITA to maintain.

> To me it looks like we need to create drawables with icon and colored background (round only) 

Round only doesn't make sense to me. Why should a cache be square on map and in lists but round in history?

moving-bits:
The whole suggestion provider code is part of the android libraries, we are only filling some API with some pieces of data. - Haven't found a way to provide my own layout file, and even if, we would need to plug into into displaying each row, to add creating and displaying a layered drawable.

> Round only doesn't make sense to me. Why should a cache be square on map and in lists but round in history?

[...]

PR #11782, opened 2021-10-04
introduce generic geoItemSelector view creation utils

fm-sys:
Make UI more consistent and avoid duplicated code...

Pre-requirement for #11591

grafik


fm-sys:
Let's merge so we can continue in #11591

PR #11783, opened 2021-10-04
rel to #11520: refactorings, add first version for scan

eddiemuc:
rel to #11520: refactorings, add first version for scan


fm-sys:
A bit unrelated to the PR, but a general question. When thinking about the upcoming beta, it might make sense to couple the new versions tab to the build type instead of the debug mode switch. 

IMHO, this WIP feature shouldn't show up for normal users, even if they have debug enabled. OTOH, nightly users might already want to test it without activating debug. So in my opinion, this would be a very good solution to check the build type programmatically...

```
if (BuildConfig.BUILD_TYPE.equals("debug") || BuildConfig.BUILD_TYPE.equals("nightly")) {
    // enable
}

eddiemuc:
> When thinking about the upcoming beta, it might make sense to couple the new versions tab to the build type instead of the debug mode switch.

That's a great idea! I changed the PR accordingly.

PR #11786, opened 2021-10-04
introduce new changelog handling (fix #11101)

### LINKED ISSUE #11101: Clarify merging effects for new changelogs

moving-bits:
## Description
Introduces the necessary changes for our new changelog handling, as described in https://github.com/cgeo/cgeo/issues/11101#issuecomment-926912603

In short:
- anything before a feature release: put changes into `changelog_base.md` (valid for both `master` and `release` branches)
- after a feature release, follow-up changes on `release` go into `changelog_bugfix.md` on `release`
- set contents of `utils/BranchDetectionHelper.java` accordingly (branch / version number)
- `version_release.txt` is no longer used

(but check the docs for details)

As usual: Updating changelogs should be separate from the change itself to avoid (or at least reduce) merge conflicts.

Currently changelog handling is mostly done by @Lineflyer and @moving-bits, which is ok and is not affected by this change.

I'll create a wiki page to document new changelog handling.


moving-bits:
> @moving-bits I guess we can delete the test branches `changelogtest_*` by now?

yes & done

PR #11794, opened 2021-10-06
[vectorIcons] Use correct icon style in search dropdown

### LINKED ISSUE #11591: [vectorIcons] Wrong icon style in search dropdown

fm-sys:
fix #11591

grafik


ztNFny:
The text colors are broken with light theme

fm-sys:
> The text colors are broken with light theme.

Good catch!

> In light theme the dropdown is a dark grey and text should be white but is black (for title) and dark grey (for gc code)

As we use the same layout file at many other places (~6 places if I remember correctly), I guess we should rather change the background color to stay consistent.

That's how it looks at all other places:
grafik

ztNFny:
I don't see any reason for the dropdown being dark in light mode - it's just a list like many others, so it makes sense if it's light with dark text.

I have a minor comment regarding the layout, but will open an issue for that.

fm-sys:
> Does this PR thanks something with respect to search suggestions.
> 
> In the past it was a nice tool to crawl the offline stored caches...that should be kept.

The searchview suggestions and how the search view behaves does not change at all compared to previous versions of cgeo. It only changes the way how the results are styled and displayed.

To be precise: 
The caches doesn't need to be fully offline to appear in the search results. But they need to be contained in the DB. (e.g.  caches of the "history" list can have this state)

PR #11795, opened 2021-10-06
[Variables] allow selection of formula result value

fm-sys:
Today, I wanted to use the new Variables feature for the first time. It worked pretty well (thanks @eddiemuc!), but I was strongly disappointed, that there was no possibility to copy the result. It was a mystery with multiple lines of `rot13` text encryption, so it was not really possible to use the brain as clipboard ;-)

This PR will allow text selection. 

grafik


fm-sys:
> Hey, it's work in progress :-)

I know. And wanted to help making it great :-)

(Shouldn't sound negative or like an accusation in any way)

PR #11798, opened 2021-10-07
rel to #11520: add variable persistance and better scan

eddiemuc:
rel to #11520: add variable persistance and better scan


fm-sys:
Can't really judge, but shouldn't we split up the code for the beta first, before introducing a new DB version?

PR #11806, opened 2021-10-08
update contributor's list

moving-bits:
## Description
October 2021 update of contributors' list


eddiemuc:
I would vote for assigning the p flag also to @Lineflyer (in addition to @mucek4)

moving-bits:
done

PR #11825, opened 2021-10-10
fix lint: remove unused strings, fix static string usage

moving-bits:
## Description
- Removes three strings no longer used
- Fixes a static string usage in `AboutActivity`


eddiemuc:
PR #11837 might help

PR #11831, opened 2021-10-10
code cleanup: fixes some 'integer division in floating-point context'…

moving-bits:
## Description
fixes some 'integer division in floating-point context' issues


fm-sys:
Does this really increase code readability?

PR #11835, opened 2021-10-10
rel to #11227: add logging for startup times

eddiemuc:
rel to #11227: add logging for startup times


fm-sys:
Maybe @moving-bits can test the CI debug APK, so we do not need to merge it at all... The logging looks a bit too "verbose" to be included in a release version, isn't it?

eddiemuc:
Fine for me, let's put this on hold. I label the PR accordingly.
(But I really just inserted two lines, only the code intent has changed)

fm-sys:
> (But I really just inserted two lines, only the code intent has changed)

Ah, ok. Just had another look, and it's codewise totally file for sure. 

But it would have been a quite hard job to merge 133 lines of intent changes and 577 lines of code changes correctly without breaking anything... 😅

eddiemuc:
Resolved conflicts with master after bottom navigation is not integrated, and added a time measurement for database initialization as well.

PR #11837, opened 2021-10-10
fix #11825: make lint MissingDefaultResource informational

eddiemuc:
fix #11825: make lint MissingDefaultResource informational


moving-bits:
I'll give it try, thanks!

PR #11838, opened 2021-10-10
fix lint: remove unused strings, fix static string usage

moving-bits:
## Description
(recreated PR, as the first one got closed automatically)
- Removes three strings no longer used
- Fixes a static string usage in `AboutActivity`


moving-bits:
No problem in itself. I just wasn't able to reopen it - GitHub complained that I had already force-pushed (due to rebasing), so something in the steps in between blocked it. But anyway, no big deal: I created a new branch on top of the old one and pushed it to the server.

PR #11865, opened 2021-10-12
reduce size of waypoint markers (fix #11861)

### LINKED ISSUE #11861: New icons feedback: waypoint markers
### Tell us your idea!

Waypoints are combined of the service's general outline marker (circle / square / hexagon) and the waypoint specific logo inside. The outline's filling is pure white.

As the inner logo's size is the same as the cache symbols are, but the waypoint markers don't have a colored background, there's much empty space between logo and outline. The logos look a bit "lost".

Reduce the outline's size for waypoints to roughly the size the colored background for cache type markers have. This would both reduce the empty space and make waypoint markers more distinguishable form cache markers.

### Is there an alternative?

Either increase the waypoint logo size (not a good idea, IMHO, would look unbalanced), or leave it as is.

### Additional Information

Edit: This is with icons in default size (not in "compact icon mode")

moving-bits:
## Description
Reduces the size of waypoint markers to a) reduce white space in markers and b) make them better distinguishable from cache markers.

grafik

### DIFF
--- a/main/res/drawable/waypoint_marker.xml
+++ b/main/res/drawable/waypoint_marker.xml
@@ -1,0 +1,11 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="30dp"
+    android:height="30dp"
+    android:viewportWidth="36"
+    android:viewportHeight="36">
+  <path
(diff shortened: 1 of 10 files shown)


fm-sys:
Maybe a slightly bit more padding around it, so that the white circle outline has the same width as for caches? In the screenshots, it IMHO looks a bit too narrowed...

Issue #11817, opened 2021-10-09
Map Update uses 21-06 files


jakedot:
### Describe your problem!

When I get an update notification for map files, the update URL seems not to be the current / latest update at all:

remote.page=http://download.freozeitkarte-osm.de/android/2106

### How to reproduce?

- Get Update Notification for a Freizeitkarte Map
- Update
- Look up which URL was actually queried during update in /cgeo/maps/***.map-cgeo.txt.txt

### Actual result after these steps?

2106 Maps are downloaded

### Expected result after these steps?

latest maps are downloaded

### Reproducible

Yes

### c:geo Version

2021.10.08

bekuno:
Now the xml is updated to 2109

moving-bits:
But the links inside still point to 2106 folders, e. g. http://download.freizeitkarte-osm.de/android/2106/freizeitkarte_bayern.map.zip

bekuno:
Sorry, yes. I missread it on the second check.

jakedot:
What are we using as our basis for the downloads and update notifications? That repository XML?

Looks for me to be a problem on their end. We could workaround by replacing the "2106" with "latest" in the links, that seems to be a correct symlink to the current 2109 folder.

moving-bits:
Exactly, we are using the repository URL, as it gives us both, a list of which maps are offered by the server, and links for the downloads. We should not change that, and freizeitkarte-osm should fix the links on their side, IMHO.

jakedot:
Told them on twitter. https://twitter.com/uzuuru/status/1447109964275802118

bekuno:
Maybe we can add a plausibility check in the website test.

jakedot:
@moving-bits yes, those files did exist before the original update and still do. If you look in my original post, you'll notice that they are the reason I found the problem at all.

I think the problem with c:geo not noticing the fixed XML is that the "version string" and the update dates for the individual maps have not changed in the updated XML.

What I did to now get the latest maps was changing the date in the companion files to an earlier one, and then the update dialog is shown again and updates the maps correctly, afterwards the correct directory was listed in the remote.page entry.

jakedot:
> which is probaly work as designed.
This uses up space for the newly duplicated map though. I think the old one should be deleted if name changes on update.

moving-bits:
> Also, it looks to me as if the system download manager is only actively downloading when c:geo is the foreground process, download notification and progress stops quite quickly after switching to other apps...

[...]

PR #11876, opened 2021-10-13
use 'latest' subfolder for FZK downloads (rel. to #11817)

moving-bits:
## Description
Related to #11817: Use "latest" subfolder instead of "yymm" subfolder - helps avoiding folder naming errors on server side like that one that happened recently / always downloads the most recent version


moving-bits:
setting WIP for now until I have a fix for #11877 to be able to further test what's going on here

Lineflyer:
retest this please

moving-bits:
Although it does not fix the "does not recognize server-side changes immediately" part, it helps mitigating the consequences of such server-side errors, as it picks downloads from the "latest" folder instead of the "year-month" folder. Thus I vote for merging this PR and will remove the "WIP" flag.

PR #11882, opened 2021-10-13
fix crash when trying to open from AL App (fix #11878)

### LINKED ISSUE #11878: Crash or "hang" when trying to open from AL App
### Describe your problem!

We have three reports for current beta on support mail, that c:geo crashes or hangs when trying to invoke the navigation from the adventure lab app to generate a new target in c:geo.

### How to reproduce?

- Use Groundspeak Adventure Lab app
- Select a location of an adventure
- Invoke navigation and select to open it with c:geo

Crash or hang

Work normal

### Reproducible

Unclear

### c:geo Version

2021.10.11-RC

### System information

_No response_

### Additional Information

Not yet find the time to reproduce

fm-sys:
Get `NavigateAnyPointActivity` running again...

grafik


moving-bits:
Thanks for providing a fix. But shouldn't this be targeted to `release`? (The error already exists in current beta.)

PR #11884, opened 2021-10-13
Save waypoints to DB for ALC caches if alc_advanced is set

JakeDot:
For ALC, if the hidden property alc_advanced is set, the waypoints are also loaded. But they are not saved to DB, and this means that they are not displayed correctly / concurrently on the map, only the waypoints of the last opened ALC cache are visible at one time.

### DIFF
--- a/main/src/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/cgeo/geocaching/connector/al/ALApi.java
@@ -262,7 +262,7 @@ final class ALApi {
             cache.setDisabled(false);
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             cache.setOwnerDisplayName(response.get("OwnerUsername").asText());
-            cache.setWaypoints(parseWaypoints((ArrayNode) response.path("GeocacheSummaries")), false);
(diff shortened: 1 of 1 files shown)


bekuno:
add to whitelist

jakedot:
@bekuno looks to me as if the alc consumer key isn't included in the build, as I don't get any adventures loaded in the installed debug app...

moving-bits:
Our production keys will be inserted for production versions only, so only when building a new release or release candidate, but not for "common" CI builds.

baiti:
> @jakedot I understood this limitation, therefore I would like to merge this to enable you to test in the nightlies. I think this function can be included...does not harm as its only for "testing purposes".
> 
> Just the question to @moving-bits or @baiti if this change does what its intended for (as I have no idea about that parameter changing from false to true) as I am not familiar with the code.

Will review tomorrow. Need to lookup the function in question so see what true / false really does.

baiti (approved):
LGTM ! The parameter in question, the one changed from `false` to `true` reads:
`saveToDatabase` .. which is exactly what the code change tries to accomplish.

fm-sys:
Then let's merge and test it in the nightlys...

PR #11887, opened 2021-10-14
move status message to the top and make listview scrollable (fix #11853)

### LINKED ISSUE #11853: Notification message pushed out of screen if many connectors are active
### Describe your problem!

For sure a "low prio" problem:
If I activate enough connectors the notification message (e.g. "New release available") will no longer be shown on the dashboard

### How to reproduce?

- Enable 6 connectors (for my screen size)
- Raise a test notification or wait for a new nightly build

No notification shown as there is no more space on the screen.

Maybe we could place the notification on top and the connector list below (and make the connectors part scrollable).

Or we could try out to be more space efficient by putting the avatar left of the status instead above.

### Reproducible

Yes

### c:geo Version

2021.10.12-NB

### System information

_No response_

### Additional Information

_No response_

fm-sys:
fix #11853

- move status/"you are offline" message to the top
-  make connectors-listview scrollable, by explicitly setting the view bottom to `above="@id/location_area"`

grafik

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -7,68 +7,59 @@
     android:layout_height="fill_parent"
     tools:context=".MainActivity">
 
+    <androidx.fragment.app.FragmentContainerView
+        android:id="@+id/status"
+        android:name="cgeo.geocaching.StatusFragment"
+        android:layout_width="fill_parent"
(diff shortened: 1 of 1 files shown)


moving-bits:
Thanks for adapting the order of the elements, let's merge it.

PR #11891, opened 2021-10-14
unbreak waypoint markers

ztNFny:
Reverts cgeo/cgeo#11865


moving-bits:
> There was a downvote on the issue, there was a comment that the paddings are too small - it was neutral, not positive.

If you look at both the original issue and the PR, there were quite a few reactions at the time the PR got merged. One was neutral, and the feedback for it (increase padding) was respected in the final PR, the others were positive.

> The current look makes my eyes bleed, it's completly unbalanced.

I respect your opinion, although I don't share it, and I won't argue on terms of design. Hey, my design courses at the university are quite some time ago :-)

> I proposed scaling the whole icon - no response to that. I questioned why waypoints should be less important - no response either.

Give me some more time to response :-) I'm working on several issues in parallel here, and your feedback was just a few minutes old when you opened your PR. It's not primarily about the contents of the PR, it's also about collaboration culture. I'm fully open to discuss how to proceed here, and if the team decides it likes the original version better, than so be it. (I will have to find a way that works for me, then, without disturbing the majority's choice. Which would be totally fine for me.) What did upset me was to quickly open a "revert" PR without giving the team a chance to discuss. No personal objection involved.

ztNFny:
> The issue was open for app. a day before the PR got merged. Not much, agreed, but as you had been active during that timespan, actively commenting on other issues regarding the icons, my impression was that you had seen my original issue and at least not disagreed with it. Sorry, if I had misread that.

I had seen the issue and planner to take a look at the issue once i'm back from vacation and see what can be done there. I had not seen the comment with the screenshot or the PR. A @ mention would have helped that, i'd have objected there and then.

moving-bits:
> BTW: PR #11865 had re-enabled an already fixed issue. Waypoints are using the wrong shape. 

Ah, good catch - seems I had missed to update the AL connector to make it use the GC-shape as well, thus the default shape got called.

But anyway, let's merge here and continue the discussion along the original issue over there.

moving-bits:
> Someone needs to clean up conflicts to be able to merge this commit back into master please.

done

Issue #11904, opened 2021-10-15
Bottom navigation: How to structure the "more" page?


moving-bits:
### Tell us your idea!

Although I really like the new bottom navigation look in general, I'm still somewhat unsatisfied with the look of the "more" page. Currently it's more or less an unsorted bunch of links (no offense intended! I do know where all those links come from...). For a basic member, it looks like this:
![grafik](https://user-images.githubusercontent.com/3754370/137530656-aeebe4f3-85d7-41c8-b780-437a1a9ac4c9.png)

And a gc.com PM even has two more entries (bookmarks / pocket queries). Depending on screen length and font sizes, you'll even have to scroll...

I have no ready-at-hand idea how to do this, so I'd like to start a gathering of ideas how to enhance and structure this page.

### What solution would you suggest?

A starting point could be to transfer two or three easily recognizable and frequently used links to icons in the toolbar, e. g.
- Settings
- About c:geo

This should be easy to realize and cause no harm, but will reduce the list by two entries already.
(If we agree on implementing #11897, this would be the third icon on this toolbar.)

Next I thought of some grouping of the remaining ten entries, but alas they are nearly as much diverse as possible :-) - Have not found a helpful grouping for the entries yet.

Another idea could be to change the textual entries into a two-column layout of large graphical buttons (+ text), but this would solve neither the grouping problem nor the possibility of having to scroll.

Any ideas whether we can leave-out some of those entries? 
We had some discussion about the "scan geocode" feature lately, but haven't decide to remove it for now. 
"Useful apps" could become a separate tab on the "about c:geo" screen (as we have other useful links regarding c:geo over there already), which would simply transfer the problem to a different, but less-crowded, place.

Also we should discuss the order in which the entries are shown. IMHO the most important entries should stay at the top (as they mostly do, currently only "report problem" seems to me being too important for staying at the end of the list).

### Is there an alternative?

hopefully plenty :-)

fm-sys:
> no offense intended!

All fine. This is the place where I have worked the least time so far... I'm not yet convinced either ;-)

And you already have mentioned some IMHO really good ideas. 👍

Another idea, which I have seen in many other apps: Introducing some kind of priority-grouping... (high emphasis entries on the top, others on the bottom) 

![Screenshot_20211015_221912_com groundspeak geocaching intro_copy_270x585](https://user-images.githubusercontent.com/64581222/137548977-622bce33-6319-4f46-a2c3-c3468b72d0ee.jpg)

[...]

PR #11906, opened 2021-10-15
switch settings/about to buttons in MoreActivity (rel. to #11904)

moving-bits:
## Description
As discussed in #11904: Switch list entries "about c:geo" and "settings" to toolbar buttons, shown as icons (with text, if room):

grafik
grafik


moving-bits:
> I will continue accessing "about" via clicking on the c:geo icon anyway. But users still need to easily find it ;-)

I guess this will be a nice shortcut (for those who know) when you are on the login dashboard, but that could not replace some linking from the "more" page.

Regarding the "about c:geo" icon:
Is your concern about having "about c:geo" as icon in the toolbar at all, or is it about finding the right icon?

PR #11921, opened 2021-10-17
mutate all drawables that will be tinted

### LINKED ISSUE #11862: New icons feedback: Cache type markers in compact mode
### Describe your problem!

If you force your map to compact icon mode, some cache type markers will have a grey circle around them instead of a white. This seems to happen for mysteries and whereigos, but not for other cache types.

### How to reproduce?

- open map with has several caches of at least the two cache types mentioned above on it
- open quick settings and set "compact icon mode" to "on"

cache of types "mystery" and "whereigo" have a light grey circle around them, where all other caches have a white one

all caches having a white circle

### Reproducible

Unclear

### c:geo Version

2021.10.11-RC

### System information

```text
Using current beta 2021.10.11-RC
```

### Additional Information

I cross-tested it against a map from a different cache list, and there I have additional caches with a grey circle around them, but those additional caches are caches which are currently disabled (trads in that case)- at least I did not find any other. But other trads have a white circle.

And some even more confusing extra case: I have a multi, non-disabled, but showing a grey circle, and this multi has a lab adventure pin as logo?!

ztNFny:
Some drawables that later got tinted were not mutated

should fix #11862

### DIFF
--- a/main/src/cgeo/geocaching/utils/MapMarkerUtils.java
+++ b/main/src/cgeo/geocaching/utils/MapMarkerUtils.java
@@ -428,9 +428,9 @@ public final class MapMarkerUtils {
      */
     @NonNull
     private static LayerDrawable createWaypointDotMarker(final Resources res, final Waypoint waypoint) {
-        final Drawable dotMarker = DrawableCompat.wrap(ResourcesCompat.getDrawable(res, waypoint.getMapDotMarkerId(), null));
+        final Drawable dotMarker = DrawableCompat.wrap(ResourcesCompat.getDrawable(res, waypoint.getMapDotMarkerId(), null)).mutate();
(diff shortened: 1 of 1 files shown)


bekuno:
This PR should target to release.

Lineflyer:
> Are there nightlies built from release? I can't repoduce the issue myself, so it's have to go into a build for someone to test.

You can either take the APK from our CI.
Alternatively once merged into `release` we usually merge `release`back into `master` timely (meaning at least once a day after changes or immediately after larger changes). So it will be in nightly by then.

Lineflyer:
Can't review contentwise, but the change is small, thus merging to allow @ztNFny testing in next nightly.

ztNFny:
> Merged and pushed to `master`. @ztNFny A new nightly 2021.10.17-NB1 is just building (ETA: 10 min) in case you want to test drive something already tonight.

I can't test as I'm not able to repoduce the issue, my Android 11 devices all don't have those icon mixup issues.

PR #11934, opened 2021-10-20
rel to #11227: prevent UI suggestion loading, parallel login, multiple avatar downloads

eddiemuc:
rel to #11227: prevent UI suggestion loading, parallel login, multiple avatar downloads


eddiemuc:
Ah well, here goes nothing... I will merge this sometime this evening if no one stops me. I tested this as far as possible in my emulator. If stuff comes up I will fix them as it occurs, I guess this is what nightlies are for.

PR #11944, opened 2021-10-21
more custom filter icons

ztNFny:
New material design filter icons instead of using emojis
b


ztNFny:
> One suggestion for a change: The "Has own vote" icon is too near to "Owned" in color and symbol (star). Therefore my suggestion is the thumbs up icon for vote.

image

PR #11957, opened 2021-10-22
add offline found counter to bottom navigation (fix #11936)

### LINKED ISSUE #11936: Total amount of stored caches missing
### Describe your problem!

In the old main activity the lists icon used to show the total amount of stored caches. This was IMHO very helpful to see whether there is some leftover to clean up or also just to see the utilization.

For beginner users it could be helpful to see the figure once they stored caches for the first time or also just to verify that the caches they wanted to store have been stored.

If its technically possible with the bottom navigation I vote for not losing this feature.

### How to reproduce?

see above

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021.10.19-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Show number of caches stored offline as badge on bottom navigation's "list" entry:

image

Currently our accent color is used for the badge. 
If we want to make it less visually intrusive, we could use the bottom navigation's text colors instead.

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -16,8 +16,10 @@ import cgeo.geocaching.maps.DefaultMap;
 import cgeo.geocaching.network.HtmlImage;
 import cgeo.geocaching.network.Network;
 import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.utils.AndroidRxUtils;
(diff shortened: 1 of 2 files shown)


Lineflyer:
See: https://github.com/cgeo/cgeo/pull/11965#issuecomment-956567416
From the colors I would favor the "non alerting" color as its just an information but no action is required (unlike the red dot on login fail).

Lineflyer:
Regarding high numbers:
We should be able to display 4 digits though. If the user has >=10k caches, then for sure its enough to show something like ">10k".

Lineflyer:
retest this please

moving-bits:
According to the recent discussion in #11936 a different solution is preferred (show number in list selector), so let's close here.

PR #11965, opened 2021-10-23
add offline stored counter to dashboard (fix #11936)

### LINKED ISSUE #11936: Total amount of stored caches missing

moving-bits:
(Less intrusive alternative to #11957)

## Description
Add a "offline stored counter" to the dashboard, added below all connector infos:

image

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -71,6 +71,21 @@
         android:orientation="vertical">
 
         <TextView
+            android:id="@+id/offline_counter"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:ellipsize="end"
+            android:layout_gravity="center"
+            android:gravity="center_horizontal"
+            android:lines="1"
+            android:maxLines="1"
+            android:textIsSelectable="false"
+            android:textSize="@dimen/textSize_listsPrimary"
+            android:visibility="gone"
+            tools:text="1234 caches stored offline"/>
+        <View style="@style/separator_horizontal" />
+
+        <TextView
             android:id="@+id/nav_location"
             style="@style/location_current"
             android:textIsSelectable="false"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -457,6 +457,14 @@
         <item quantity="many">%d logged offline</item>
         <item quantity="other">%d logged offline</item>
     </plurals>
+    <plurals name="caches_stored_offline">
+        <item quantity="zero">%d caches stored offline</item>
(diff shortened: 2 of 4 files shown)


bekuno:
We had tickets with performance problems regarding lists.
Therefore if would be good to have the counter (if possible separated for each connector) in the system information.
This is a must have, if we have no user visible counter.

ztNFny:
> This is a must have, if we have no user visible counter.

There is a counter already though? The "all caches" list.

bekuno:
> 
> 
> > This is a must have, if we have no user visible counter.
> 
> There is a counter already though? The "all caches" list.

Not at the first place.
Screenshot_20211030-140305_cgeo
But from the list menu there is it:
Screenshot_20211030-140328_cgeo

Should be unified, imho.

ztNFny:
> Should be unified, imho.

I think that's because you can't save a cache to "History" or "All caches", those are virtual lists. They should probably not even show up in the pick a list menu then.

Lineflyer:
I think we should still have some kind of count somewhere.
From support mail I know many experienced users (who probably don't need it) but also a lot of "non teccy" users who even struggle to understand the concept of multiple lists. 
Therefore informing the user, that something changed after they did some save/delete action or just to make them aware they have something stored by just showing the totals is IMHO still a good idea and wI would not like to loose this feature with the bottom navigation.

With respect to the implementation I still struggle. I still somehow think, that using the bottom nav capability to show such counter should be used. This fits into the concept, that we e.g. show a red alert dot on the "Home" if there is attention required (e.g. login failed). 
So my vote finally goes to the other PR instead of this one.

Lineflyer:
retest this please

moving-bits:
According to the recent discussion in #11936 a different solution is preferred (show number in list selector), so let's close here.

SammysHP:
@ztNFny

> > Should be unified, imho.
> 
> I think that's because you can't save a cache to "History" or "All caches", those are virtual lists. They should probably not even show up in the pick a list menu then.

That dialog is also used as a list quick selection (long press on lists in bottom navigation).

PR #12006, opened 2021-10-27
always show waypoint marker with background

### LINKED ISSUE #11777: [VectorIcons] waypoint marker style in waypoint list
Sshould we use our standart waypoint marker for the waypoint list as well, or should we stay with the current design?

Screenshot_20211003_095433_copy_431x324

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11732#issuecomment-932883932_

### LINKED ISSUE #11778: [VectorIcons] remove pin from waypoint markers in non-map contexts
Looks good generally, but would it be possible to have the waypoint icons without the little "pin" at the bottom?

_Originally posted by @moving-bits in https://github.com/cgeo/cgeo/issues/11768#issuecomment-932862683_

Screenshot_20211003_092610_cgeo geocaching_copy_324x702

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11768#issuecomment-932880743_

IMHO the pin makes sense on the map only, for pointing to a specific location on the map. I would leave it out in all other places. But I agree that this would fit for a separate issue.

_Originally posted by @moving-bits in https://github.com/cgeo/cgeo/issues/11768#issuecomment-932901927_

ztNFny:
Whenever a waypoint image is shown display it as marker

fixes #11777
fixes #11778

### DIFF
--- a/main/res/layout/waypoint_item.xml
+++ b/main/res/layout/waypoint_item.xml
@@ -12,33 +12,28 @@
     android:paddingTop="9dp"
     tools:context=".CacheDetailActivity$WaypointsViewCreator$1" >
 
-    <RelativeLayout
+    <LinearLayout
         android:layout_width="fill_parent"
         android:layout_height="fill_parent"
         android:descendantFocusability="blocksDescendants"
         android:orientation="horizontal" >
 
         <ImageView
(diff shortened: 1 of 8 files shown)


eddiemuc:
Looks fine, let's merge

Issue #11991, opened 2021-10-26
Sorting: current choosen sort type and direction should be made visible to user in list view


eddiemuc:
### Tell us your idea!

This is a follow-up issue for #8981 where storing of sort settings per list was implemented. As part of this item, the sort selection GUI needed to be enhanced as well.

This issue is to discuss the sort "look & feel" and maybe find change requirements to how it is currently implemented.

One thing already mentioned in PR and partly issue above is to add a visual "sort indicator" in the overview list, but I haven't had an idea yet how to do that. (cite @fm-sys : "I would strongly suggest to have a visual feedback that a special sort order is applied (similar to what we have for filters). But this can go into a follow-up issue.")

### What solution would you suggest?

This is how the search submenu currently looks like: 

![image](https://user-images.githubusercontent.com/6909759/138602135-40a10096-8658-4ca1-b0e0-6221ebe5e5d9.png)

### Is there an alternative?

Do nothing

eddiemuc:
For comparison:

Option one (Second line to view filter) looks like this:

![image](https://user-images.githubusercontent.com/6909759/139487983-8cd6c51d-127f-4b23-a3a6-5dd77abf7a23.png)

Option three (same line as filter, sort is right-aligned) looks like this:

![image](https://user-images.githubusercontent.com/6909759/139456826-3416ba8f-d1cf-46a8-895e-4bfbe922f71b.png)

eddiemuc:
With PR #12015 I prepared everything for showing a sort GUI in either of the three discussed ways. I threw in some refactoring while I was at it.

We had some good experience in the past with "voting via emojicons", may I ask the team to vote for one of the following options?

* **Thumbs Up**: Second line to filter view (see screenshot "Option 1" in previous post)
* **Thumbs Down**: Same line as filter view, prepended to filter
* **Laughing**: Same line as filter view, right-aligned (see screenshot "Option 3" in previous post)
* **Party-head**: do nothing, don't show sort (reject this issue)
* **Eyes-looking-to-left**: I don't care, choose whatever you want

eddiemuc:
Voting seems to favor option 3. PR #12027 will implement this option and thus close this issue.

PR #12015, opened 2021-10-29
rel to #11991: add sort indicator to list gui

eddiemuc:
rel to #11991: add sort indicator to list gui

Adds a sort indicator as defined in #11991 in the last option

With an active filter + chosen sort, this looks somewhat like this:

image

If there's no filter and sort set to "Auto", then the bar is not visible at all (and takes no space). If there's no filter but a sort != auto, then it is shown same way in upper right corner


eddiemuc:
I changed the PR such that it will prepare the code and offer layout options but does not change anything. E.g. merging this PR will not actually change anything from the user's perspective. This should give the team time to discuss on how the sort option shall be displayed in #11991

PR #12016, opened 2021-10-29
tweak map quick settings icons

ztNFny:
after
image
image

before
before

disabled needed an update after the icon size changed on map
archived used to be just a red X, making it look out of place in that list of round icons, this should be a better representation
marker used the old overlay checkmark which isn't used anymore, using the standard marker now


eddiemuc:
Looks fine, let's merge. @ztNFny thanks for the continuous contributions!

PR #12026, opened 2021-10-31
new waypoint-style multi icon

ztNFny:
image

as discussed in https://github.com/cgeo/cgeo/discussions/11912


Lineflyer:
As discussed in #11912 lets take this for the next beta then and let the users decide in a survey.

Issue #12052, opened 2021-11-05
Improve filter view touch targets


fm-sys:
- clicking on the sort info should open the sort selection list

![Screenshot_20211103_064355_copy_300x150](https://user-images.githubusercontent.com/64581222/140471217-af7cd8ec-a8b0-46b4-989b-33f97cbed7f2.png)

- if only one of both infos is visible, the whole width should react as touch target. Not only the text...

![Screenshot_20211105_080017_copy_539x199](https://user-images.githubusercontent.com/64581222/140471502-e7a069a9-1cc8-49f3-9cd6-8413ffd84c38.png)

PR #12053, opened 2021-11-05
open sort selection list on sortbar click

fm-sys:
rel. to #12052

Screenshot_20211103_064355_copy_300x150


eddiemuc:
I was not aware that this is possible. Nice find :-)

PR #12056, opened 2021-11-06
position snackbar above bottom navigation (fix #12014)

### LINKED ISSUE #12014: Snackbar overlaps bottom navigation
### Describe your problem!

When you delete a cache from a list an undo snackbar is shown on top of the bottom navigation. This snackbar cannot be dismissed, even by the back key and you have to wait until it disappears to use to bottom navigation again.

### How to reproduce?

Remove cache from list.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021-10-29-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Use bottom navigation layout element as anchor view to position the snackbar above it.
If bottom navigation layout element is not found (null) placement will be as today.

image

### DIFF
--- a/main/src/cgeo/geocaching/command/AbstractCommand.java
+++ b/main/src/cgeo/geocaching/command/AbstractCommand.java
@@ -126,6 +126,7 @@ public abstract class AbstractCommand implements Command {
             if (StringUtils.isNotEmpty(resultMessage)) {
                 Snackbar.make(context.findViewById(android.R.id.content), resultMessage, UNDO_DURATION_MILLISEC)
                         .setAction(context.getString(R.string.undo), this)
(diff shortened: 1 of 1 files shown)


Lineflyer:
Is there any possibility, that the snackbar is triggered on a view where there is no bottom navigation? Respectively, is this handled gracefully in such case?

fm-sys:
`findViewById` will return `null` if the view does not exist, and `setAnchorView` allows passing null as parameter. So IMHO, this should be fine...

moving-bits:
> Is there any possibility, that the snackbar is triggered on a view where there is no bottom navigation? Respectively, is this handled gracefully in such case?

Yes, had tested the `null` case explicitly. The snackbar is being displayed in same position as today in that case (as written in the PR info).

PR #12064, opened 2021-11-07
tweak OSM zoom buttons

ztNFny:
Slightly tweak the look of OSM zoom button (size, position)

Before
signal-2021-11-07-155535

After
signal-2021-11-07-155209_001

Google Maps
signal-2021-11-07-155551


fm-sys (approved):
LGTM

Issue #11328, opened 2021-07-26
Long-click on filter button should allow restoring the last filter


ztNFny:
### Tell us your idea!

The new filter GUI it requires 3 clicks to unset a filter: Filter -> Trashcan -> Check.
All 3 buttons are in slightly different places, so it's easy to miss-click there (e.g. hitting the cancel button).

Once a filter is set the most common action is to unset the filter (as opposed to changing it), thus this action should be quick to perform.

### What solution would you suggest?

Implement a long-click action on the filter button to:
- If a filter is set unset it
- If no filter is set restore the previous filter

### Is there an alternative?

At least for clearing the current filter there could be a second toolbar button.

eddiemuc:
This request partly overlaps and partly collides with #11202

murggel:
> This request partly overlaps and partly collides with #11202 

Correct.

In #11202 there will be the option to delete / clear the current filter in the popup-menu, so one more click necessary as with a separate button.

Restoring the previous could be archived to have that one already selected when opening the popup, which means one more click for selecting a different one...

murggel:
Probably adding a button "Last filter" like "last selection" in list selection would be a possibility to restore the previous filter

fm-sys:
> Restoring the previous could be archived to have that one already selected when opening the popup, which means one more click for selecting a different one...

I don't think this would make sense. @ztNFny do you really need such a restore feature, or is it primarily about clearing the filter so #11202 is sufficient for you? If so, I think we can close here...

ztNFny:
> I don't think this would make sense. @ztNFny do you really need such a restore feature, or is it primarily about clearing the filter so #11202 is sufficient for you? If so, I think we can close here...

My main beef with the curren state is the number of clicks required to reset the filter.
I often (re)apply the filter, so an easy restore would be nice to have and with the long-click approach I initially suggested I thought that'd be a nice little feature to have in there. But it's not a must-have, so if you want to close it's ok.

ztNFny:
closing this in favor of the more clearly described #12075

PR #12070, opened 2021-11-08
don't clear openInAdvancedMode setting when deleting a filter

fm-sys:
Somewhere this issue was recently mentioned off-topic, but I don't find the place anymore...

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/GeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/GeocacheFilter.java
@@ -132,7 +132,12 @@ public class GeocacheFilter implements Cloneable {
 
     @NonNull
     public static GeocacheFilter createEmpty() {
-        return new GeocacheFilter(null, false, false, null);
+        return createEmpty(false);
+    }
+
+    @NonNull
+    public static GeocacheFilter createEmpty(final boolean openInAdvancedMode) {
+        return new GeocacheFilter(null, openInAdvancedMode, false, null);
     }
 
     @NonNull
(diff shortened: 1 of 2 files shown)


ztNFny:
I can't see that this makes any difference at all? Clearing the filter behaves the same with and without this.

I think you're refering to https://github.com/cgeo/cgeo/issues/11328

fm-sys:
> I think you're refering to #11328

Yes, exactly. Thank you!



> I can't see that this makes any difference at all?  

Have you really tested it with the "quick clear" button? For me, it had worked in the emulator...

grafik
 

https://github.com/cgeo/cgeo/issues/11328#issuecomment-950599277:
> Also that Clear is just even worse than the "Delete" function inside the filter dialog, it removes all the filter options and also sets the filter to simple instead of advanced mode.

fm-sys:
> had hoped it would fix the "removes all the filter options" part.

If you are referring to "use OR", "inconclusive" etc. it should be clarified first, if these settings are really expected to be persistent even after a complete clear... For "advanced" this is definitely the case (and already handled that way by the filters activity). But for the other settings...? 

Opinions?

eddiemuc:
Codewise it looks fine. Won't merge though because there seema to be a still-open discussion about this.

eddiemuc (approved):

ztNFny:
> If you are referring to "use OR", "inconclusive" etc. it should be clarified first

No. I'm refering to the actual filter settings.
The way I'm using the filters I always have Types and Status sections enabled, then (the 2 most frequently used scenarios) filter for "Found" or "Has offline log". To unset that filter I then have 2 options:
- press delete / clear -> next time I open the filter dialog I'll first have to add the section back -> many clicks
- go to the filter dialog, find the modified filter (e.g. "Found") and change it from "Yes" to "All" -> search for the filter, time consuming.

But I think this discussion should be in #11328, not in this PR which in itself is probably fine (but not solving #11328)

fm-sys:
The still ongoing discussion is unrelated to the bug, which was mentioned in that issue as off-topic. So let's merge here...

PR #11298, opened 2021-07-22
Refactor settings to AndroidX

SchoolGuy:
## Description

- Changes to AndroidX settings
- Uses MaterialUI for the Settings
- Removes the custom Preference Implementation and replaces the content with defaults

## Related issues

Fixes #10579 

Related Board with sub-issues: https://github.com/cgeo/cgeo/projects/21

## Additional context

- This will PR will just get big, there is no away around it. I know it is ugly but I won't cut it into multiple.
- Once I am done I will force push everything, until then don't give a damn about the Git history please.

## Plan for this PR

- [x] Splitup the one big XML into smaller ones
- [x] Create all fragments and replace the TODOs with their names
- [x] ~~Obsolete all custom Preference implementations with default ones~~
- [ ] Move the custom code from the main SettingsActivity into the fragments
- [x] Fix the styling of the settings
- [ ] Fix all bugs we can find

PR #12071, opened 2021-11-08
Log templates

ztNFny:
I have added 5 additional templates that work in the same way "Signature" does in defining them in preferences.
They do show up in the "insert" menu right below signature - but only if they are non-empty - from where you can insert them manually.

Showing the title is currently done in an ugly way, using one preference to store template name and text with a divider - I couldn't find any good way to have two associated preference keys (for title and text). If someone has an idea on that please feel free to pick up the code in PR (#12071). Due to this I've marked the PR as Draft for now.
(Also the number of templates is fixed now as it's using preferences, implementing title and text as some form of objects would probably allow making that dynamic)

image
image

### DIFF
--- a/main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
+++ b/main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
@@ -34,7 +34,11 @@ public abstract class AbstractLoggingActivity extends AbstractActionBarActivity 
 
         final SubMenu menuLog = menu.findItem(R.id.menu_templates).getSubMenu();
         for (final LogTemplate template : LogTemplateProvider.getTemplatesWithSignature()) {
-            menuLog.add(0, template.getItemId(), 0, template.getResourceId());
+            if (template.getResourceId() == 0) {
+                menuLog.add(0, template.getItemId(), 0, getString(R.string.init_log_template_prefix) + template.getName());
+            } else {
+                menuLog.add(0, template.getItemId(), 0, template.getResourceId());
+            }
         }
 
         final SubMenu menuSmilies = menu.findItem(R.id.menu_smilies).getSubMenu();
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -1958,4 +1958,18 @@ public class Settings {
         return sensitiveKeys;
     }
 
+    public static void putLogTemplate(final String preferenceBaseName, final String title, final String text) {
+        putStringDirect(preferenceBaseName + ".Title", title);
+        putStringDirect(preferenceBaseName, text);
+    }
+
+    public static ImmutablePair<String, String> getLogTemplate(final String preferenceBaseName) {
+        return new ImmutablePair<>(getStringDirect(preferenceBaseName + ".Title", ""), getStringDirect(preferenceBaseName, ""));
+    }
+
+    public static String getLogTemplatePreview(final String preferenceBaseName) {
+        final ImmutablePair<String, String> template = getLogTemplate(preferenceBaseName);
+        return template.getKey() + ": " + template.getValue();
+    }
+
 }
--- a/main/src/cgeo/geocaching/settings/TemplateTextPreference.java
+++ b/main/src/cgeo/geocaching/settings/TemplateTextPreference.java
@@ -1,5 +1,7 @@
 package cgeo.geocaching.settings;
 
+import static cgeo.geocaching.utils.LocalizationUtils.getString;
+
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.log.LogTemplateProvider;
@@ -21,6 +23,7 @@ import androidx.preference.PreferenceViewHolder;
 import java.util.List;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 
 public class TemplateTextPreference extends Preference {
 
@@ -30,6 +33,7 @@ public class TemplateTextPreference extends Preference {
     private static final String DEFAULT_VALUE = StringUtils.EMPTY;
     private SettingsActivity settingsActivity;
     private EditText editText;
+    private EditText editTitle;
     private String initialValue;
 
     public TemplateTextPreference(final Context context, final AttributeSet attrs) {
@@ -46,35 +50,60 @@ public class TemplateTextPreference extends Preference {
         final Preference pref = findPreferenceInHierarchy(getKey());
         if (pref != null) {
             pref.setOnPreferenceClickListener(preference -> {
-                editSignature();
+                editTemplate();
                 return false;
             });
         }
     }
 
-    private void editSignature() {
+    private void editTemplate() {
         settingsActivity = (SettingsActivity) this.getContext();
 
+        final boolean isSignature = this.getKey().equals(getString(R.string.pref_signature));
+
+        final ImmutablePair<String, String> template = Settings.getLogTemplate(this.getKey());
+
         final View v = LayoutInflater.from(getContext()).inflate(R.layout.template_preference_dialog, null);
+        editTitle = v.findViewById(R.id.title);
+        if (isSignature) {
+            editTitle.setText(R.string.init_signature);
+            editTitle.setEnabled(false);
+        } else {
+            editTitle.setText(template.getKey());
+        }
         editText = v.findViewById(R.id.edit);
-        editText.setText(getPersistedString(initialValue != null ? initialValue : StringUtils.EMPTY));
+        editText.setText(template.getValue());
         Dialogs.moveCursorToEnd(editText);
 
         final AlertDialog dialog = Dialogs.newBuilder(getContext())
             .setView(v)
-            .setPositiveButton(android.R.string.ok, (dialog1, which) -> {
-                final String text = editText.getText().toString();
-                persistString(text);
-                callChangeListener(text);
-            })
+            .setPositiveButton(android.R.string.ok, null)
             .setNeutralButton(R.string.init_signature_template_button, null)
             .show();
 
         // override onClick listener to prevent closing dialog on pressing the "default" button
+        dialog.getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener(v2 -> {
+            final String newTitle = editTitle.getText().toString();
+            final String newText = editText.getText().toString();
+            // check that for log templates both title and text are filled
+            if (!isSignature && StringUtils.isEmpty(newTitle) && !StringUtils.isEmpty(newText)) {
+                editTitle.setError(getString(R.string.init_log_template_missing_error));
+            } else if (!isSignature && !StringUtils.isEmpty(newTitle) && StringUtils.isEmpty(newText)) {
+                editText.setError(getString(R.string.init_log_template_missing_error));
+            } else {
+                if (isSignature) {
+                    persistString(newText);
+                } else {
+                    Settings.putLogTemplate(this.getKey(), newTitle, newText);
+                }
+                callChangeListener(newText);
+                dialog.dismiss();
+            }
+        });
         dialog.getButton(DialogInterface.BUTTON_NEUTRAL).setOnClickListener(v2 -> {
             final AlertDialog.Builder templateBuilder = Dialogs.newBuilder(TemplateTextPreference.this.getContext());
             templateBuilder.setTitle(R.string.init_signature_template_button);
-            final List<LogTemplate> templates = LogTemplateProvider.getTemplatesWithSignatureAndLogText();
+            final List<LogTemplate> templates = LogTemplateProvider.getTemplatesWithoutSignature();
             final String[] items = new String[templates.size()];
             for (int i = 0; i < templates.size(); i++) {
                 items[i] = settingsActivity.getString(templates.get(i).getResourceId());
--- a/main/res/layout/template_preference_dialog.xml
+++ b/main/res/layout/template_preference_dialog.xml
@@ -9,12 +9,21 @@
 
     <com.google.android.material.textfield.TextInputLayout style="@style/textinput_edittext">
         <EditText
+            android:id="@+id/title"
+            style="@style/textinput_embedded"
+            android:inputType="textEmailSubject"
+            android:autofillHints="@string/init_template_title"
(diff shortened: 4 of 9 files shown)


eddiemuc:
@ztNFny 
>  I couldn't find any good way to have two associated preference keys (for title and text)

In `Settings` class, instead of using methods `getString`/`putString`, use `getStringDirect`/`putStringDirect` in combination with `getKey`. An example on how this can be used is e.g. storing of filter strings per context or sort config per list type.

ztNFny:
> use getStringDirect/putStringDirect 

getStringDirect is just a wrapper around getString though?

`    private static String getStringDirect(final String prefKey, final String defaultValue) {
        return sharedPrefs == null ? defaultValue : sharedPrefs.getString(prefKey, defaultValue);
    }`

> in combination with getKey

That get's the name of the preference?!

What I need is storing 2 user-inputs, "Title" and "Text", the two fields that define a log template. I'm storing them as 1 preference with a delimiter now ("Title: Text"), would be better to store then as an object, e.g.

`{
Logtemplates: [
{ Title: "Test", Text: "Template 1"},
{ Title: "abc", Text: "Template 2"},
{ Title: "bla", Text: "Template 3"}
]
}`

eddiemuc:
@ztNFny you can go the object way if you like. We have Json support build-in, I believe there's even some JsonUtils class or something alike.

But I meant something simpler:
* Create a single "base" property key, e.g. R.string.pref_logTemplate
* Create an enum listing all templates allows (or even better: use arbitrary integer IDs, so you can add later as many as the user wants)
* Create the "store" method in Settings a bit like as follows:

```
public void putTemplate(final int id, final String title, final String template) {
  final String baseKey = getKey(R.string.pref_logTemplate);
  final String key = baseKey + "." + id;
  putStringDirect(key + ".title", title);
  putStringDirect(key + ".template", template);
}
```

* Do the reverse for getting the template's data

ztNFny:
> But I meant something simpler:

That's something I alredy tried in a similar way, it worked for consuming it but I failed in getting that into TemplateTextPreference.java and preferences.xml. When loading the template edit dialog in TemplateTextPreference.java the preference name isn't available (at least I couldn't find it), so I can't load the second preference.

eddiemuc:
> > But I meant something simpler:
> 
> That's something I alredy tried in a similar way, it worked for consuming it but I failed in getting that into TemplateTextPreference.java and preferences.xml. When loading the template edit dialog in TemplateTextPreference.java the preference name isn't available (at least I couldn't find it), so I can't load the second preference.

Not sure if I understand the problem. But on the "get" side the code of course has to be able to handle gracefully if the preference does not exist (yet) because nothing was stored there (yet).

Somethink like this should work (as with the former code snippet I just type this into the comment field, so no guarantee it compiles :-)):

//I use Pair here to return title + template text, but you can create an own value object instead of course
public Pair<String, String> getLogtemplate(final int id) {
final String baseKey = getKey(R.string.pref_logTemplate);
  final String key = baseKey + "." + id;
  final String title = getStringDirect(key + ".title");
  final String template = getStringDirect(key + ".template");
  return new Pair<>(title == null ? "" : title : template == null ? "" : template);
}

//this assumes that existing log templates are numerated with continuous ids starting with id = 0. Other solutions are possible (e.g. using an enum instead of an integer id), depends on how flexible you want this to be
public List<Pair<String, String>> getAllLogTemplates() {
  final List<Pair<String, String>> result = new ArrayList<>();
  int id = 0;
  while (getStringDirect(getKey(R.string.pref_logTemplate) + "." + id + ".template") != null) {
    result.add(getLogtemplate(id++);
  }
  return result;
}

ztNFny:
That's not what I meant.
"Using" the template with title + text ("pref_logTemplate1Title", "pref_logTemplate1Text") from LogTemplateProvider.java works fine.
Also non-existant values shouldn't be possible, the preferences are defined in preference_keys.xml (maybe they don't have to be?) to be used in preferences.xml

For the settings activity the available preferences are defined in "preferences.xml", from that one the settings dialog is created. In there only 1 preference name is defined, e.g. only the pref_logTemplate1Text.
Clicking the preference starts cgeo.geocaching.settings.TemplateTextPreference which then displays the dialog to modify the template text. I can use getKey() to retrieve the preference name, then apply a string search and replace to it to get the Title preference, but that seems very hackish to me.

I was hoping that there's "something" available that helps with such situations of handling multiple preference that belong to each other instead of having to implement either such a string-hack as described above or a custom Object-based solution. Haven't worked with android.preference.Preference much so far, so I just don't know whether that's available (but the whole Prefs handling in Android always seemed underwhelming to me whenever I touched it)

eddiemuc:
Ok understood.

To turn this around a bit, I think that c:geo uses preferences much more extensive as it is the idea behind them. Preferences normally are simple key-value-pairs and imho are not intended to store more complex objects (I don't know if there's any app with a more complex Settings "framework" as c:geo is). This is why I think the frameworks only support one-key-value display options as well. More complex business objects, I guess, should go rather to a database or similar.

So if we want to circum that we have to work our way around it ourselves. We have done so for quite some time now I think :-) (even the TemplateTextPreference is obviously a work from our own)

fm-sys:
Just a short side note:

Before spending too much work into these preferences things, please wait until  #11298 is finished. The whole preferences implementation will significantly change, so it probably makes sense to wait some more time...

ztNFny:
Too late, just found a not too ugly way ;-)

ztNFny:
A proper implementation would store the log templates as objects, with a GUI for adding, removing and re-arranging the templates. But I think this is good enough for now, max 5 templates and the insert dialog only shows non-empty ones. Stored as 3 preferences per template.

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -1931,4 +1936,17 @@ public static int getListInitialLoadLimit() {
         return sensitiveKeys;
     }
 
+    public static void putLogTemplate(final String preferenceBaseName, final String title, final String text) {
+        putStringDirect(preferenceBaseName + ".Title", title);
+        putStringDirect(preferenceBaseName + ".Text", text);
+        putStringDirect(preferenceBaseName, title + ": " + text);
+    }
+
+    public static String[] getLogTemplate(final String preferenceBaseName) {
```
eddiemuc, line 1945: Consider using Pair or ImmutablePair instead of an array

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -1931,4 +1936,17 @@ public static int getListInitialLoadLimit() {
         return sensitiveKeys;
     }
 
+    public static void putLogTemplate(final String preferenceBaseName, final String title, final String text) {
+        putStringDirect(preferenceBaseName + ".Title", title);
+        putStringDirect(preferenceBaseName + ".Text", text);
+        putStringDirect(preferenceBaseName, title + ": " + text);
```
eddiemuc, line 1942: Why is this third property needed?

main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
```diff
@@ -34,7 +34,11 @@ public boolean onCreateOptionsMenu(final Menu menu) {
 
         final SubMenu menuLog = menu.findItem(R.id.menu_templates).getSubMenu();
         for (final LogTemplate template : LogTemplateProvider.getTemplatesWithSignature()) {
-            menuLog.add(0, template.getItemId(), 0, template.getResourceId());
+            if (template.getResourceId() == 0) {
```
eddiemuc, line 37: Maybe make decision based on template name being nonnull (instead of resourceId). (I don't fully understand why the resourceId is needed as an attribute of the template anyway)

main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
```diff
@@ -34,7 +34,11 @@ public boolean onCreateOptionsMenu(final Menu menu) {
 
         final SubMenu menuLog = menu.findItem(R.id.menu_templates).getSubMenu();
         for (final LogTemplate template : LogTemplateProvider.getTemplatesWithSignature()) {
```
eddiemuc, line 36: Just an idea: what about sorting template order in menu by template name?

main/src/cgeo/geocaching/settings/TemplateTextPreference.java
```diff
@@ -80,9 +90,10 @@ private void insertSignatureTemplate(final LogTemplate template) {
     @Override
     protected void onDialogClosed(final boolean positiveResult) {
         if (positiveResult) {
-            final String text = editText.getText().toString();
-            persistString(text);
-            callChangeListener(text);
+            Settings.putLogTemplate(preferenceBaseName, title.getText().toString(), editText.getText().toString());
+            String displayValue = title.getText().toString() + ": " + editText.getText().toString();
+            persistString(displayValue);
```
eddiemuc, line 95: Wh is this still needed?

eddiemuc (approved):
I don't like very much the hardcoding of up to 5 values, but it will serve it's purpose. I have a few non-critical comments and will not merge yet in case you want to act on them.

main/src/cgeo/geocaching/settings/Settings.java
ztNFny: .Title and .Text are what's used for editing and inserting. "preferenceBaseName" is what's shown in the preferences dialog - that one can display only 1 value and that one must be a preference. So I can either display "just title", "just text" or this combined preference there.

image

main/src/cgeo/geocaching/settings/Settings.java
ztNFny: Thanks, will look into that.

main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
ztNFny: This code is handling not just the new template but the whole "insert" menu. "template" can be an "old" template like the signature or elements  like Date, Owner, ... - or the new templates. The old ones have a static name being defined as resource, only the 5 new templates have a dynamic name that's provided as string.
The check could either be  (template.getResourceId() == 0) or (!template.getName().isEmpty()), same result.

main/src/cgeo/geocaching/log/AbstractLoggingActivity.java
ztNFny: That'd remove the possibility for the user to decide about the order, I think the current is better than alphabetical order.

main/src/cgeo/geocaching/settings/TemplateTextPreference.java
ztNFny: refer to the answer about the third property - it's what's displayed in the prefs dialog. Those lines are needed for that dialog to update properly, without them it only updates when calling the Activity again.

eddiemuc:
@ztNFny all your responses are valid, shall I merge?

ztNFny:
@eddiemuc update pushed, should be good now

eddiemuc:
PR is still marked as "work in progress"

eddiemuc:
... and some checks are red now unfortunately (harmless stuff afaiks).

I have go to bed now but merge tomorrow morning if checks are green. Hope that's ok.

fm-sys:
> shall I merge?

Please not yet, at least not without the agreement of @SchoolGuy. Please think about his great work in #11298. He is helping us in a really incredible way with updating the preferences to androidx. This is quite a difficult (nearly impossible) task anyway, so we shouldn't make his life even harder.

We had agreed in one dev meeting (July/August?)  to postpone all code changes which are dealing with preferences and aren't absolutely necessary for the time being. And as far as I understood @SchoolGuy the migration is nearly ready...

ztNFny:
Ok, didn't know, let's wait in feedback there

SchoolGuy:
@ztNFny I got a checklist in the linked PR. Currently I am dealing with physical back problems which prevent me to fix them today but as of Saturday I should be able to progress the PR with focus. Thus please be patient. I promise I will help with merge conflicts I caused.

ztNFny:
@SchoolGuy no worries, this has time, i just wasn't aware of the freeze on settings. And maybe you even even have a better idea of storing it (or the new implementation offers one)

bekuno:
retest this please

(green test was a false positive result)

bekuno:
@mucek4 @Lineflyer 
I think our CI **cgeo-1** needs a restart, maybe also a clean up of the filesystem.

I restartet the latest PR tests - and all test on **cgeo-5** where fine.
Then here I started the restest before the cgeo-5 was ready again. So it runs on cgeo-1, and now the result is not ok.
The first problem is at
~~~
> Task :main:copyDefaultPreferencesToAndroid
...
Process 'command '/opt/android-sdk-linux/platform-tools/adb'' finished with non-zero exit value 1mkdir: '/data/data/cgeo.geocaching/shared_prefs/': File exists
~~~
Later the tests failed and did not finished:
~~~
Starting 594 tests on android-28(AVD) - 11
cgeo.geocaching.CgeoApplicationTest > testSearchByCoords[android-28(AVD) - 11] [31mFAILED [0m
	java.lang.AssertionError:
	Expecting actual not to be null
Daemon vm is shutting down... The daemon has exited normally or was terminated in response to a user interrupt.
~~~
The first idea was that the CI PR Build Time-out strategy was too hard, and maybe a problem with execution in parallel brings the problem.
But the change of this value did not really helped. Nevertheless, the 150% looked a little bit small for a criticel situation, so let's take the 200%.

bekuno:
retest this please

... hopefully on cgeo-5

eddiemuc:
@bekuno many thanks for diving into this! It is really hard to work with an unreliable CI

kumy:
retest this

Lineflyer:
@ztNFny 
As settings are now merged you could try to rebase this change to potentially include it then?

ztNFny:
rebased onto master with new preferences

ztNFny:
retest this

ztNFny:
is the CI down?

Lineflyer:
It doesnt react to the webhooks. I tried to debug but failed and opened an issue for that.

Lineflyer:
retest this please

Lineflyer:
retest this please

Lineflyer:
retest this please
(testing the webhook)

Lineflyer:
CI is working again now. Build for this PR is queued.

Lineflyer:
retest this please (checking executors)

Issue #10924, opened 2021-06-14
Display mysteries with changed coordinates with enlarged "corrected coordinates" marker


dcjkfgdjhd:
## What function are you missing

Unlike the GC.com website, mystery caches that are solved (i.e. coordinate was updated by user) are still displayed with the default ?-icon on the map. This means that you cannot easily distinguish them from unsolved mysteries.

## Suggested solution

Solved mysteries should have an entire large icon that shows the puzzle tile.

fm-sys:
(...connected to the big icon setting)

moving-bits:
@ztNFny 
While you are working on a new icon set it would be nice if you could provide an extra icon for this (using this icon would be opt-in, like the "big smiley" is)

ztNFny:
I don't understand this FR. Caches with corrected coordinates already have marker_usermodifiedcoords to distinguish them?!
If this is about dot-markers, I've introduced dot_marker_usermodifiedcoords in my branch.

dcjkfgdjhd:
>And how can you tell the actual type? Would you suggest to have a orange "corrected" for multi, a blue "corrected" for mystery and so on?

I guess different colored versions of the puzzle tile would be a cute addition. I suppose creating multiple versions of the same icon with different hues is not too much extra work, so I would appreciate it. 

But it would not be a priority for me. The important thing is that once a mystery has corrected coordinates, the way a cacher interacts with it is different: Now it is a place that it makes sense to actually go to. Before it was just a meaningless marker (due to fake coordinates) that a cacher that looks at the map will almost always just ignore. After solving it, the mystery has turned essentially into a traditional or a multi, thus a "visitable location". This transformation should be immediately visible on the map. 

> I actually disagree with the original statement opening this Issue: They have the marker_usermodifiedcoords and can thus be easily identified.

Sorry, but on a big map full of lots of mystery icons, the tiny "red flag attachment" to the icon is almost invisible. I have this problem almost every day in a city cluttered with mysteries. There is a reason why GC.com also does it on their map. When I see an icon like that, I know that this is essentially a traditional I can visit at any moment as part of some cache tour. Unlike the normal question mark icon, where this is not possible without additional preparation.

Lineflyer:
> (...connected to the big icon setting currently used for the found/DNF smiley)

I always thought its already part of that setting (which I rarely use). Indeed it could be part of it.

ztNFny:
[...]

PR #12083, opened 2021-11-12
Display caches with corrected coords with enlarged "blue flag" marker

fm-sys:
rel. to #10924

This will look like this:

grafik

@ztNFny are you ok with this implementation?

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -788,8 +788,8 @@
     <string name="init_summary_brouterShowBothDistances">If routing is active: Show the straight distance in addition to the calculated route length.</string>
     <string name="init_useInternalRouting">Use internal routing</string>
     <string name="init_summary_useInternalRouting">Use c:geo internal routing to be able to use automatic routing data downloading and other functions</string>
-    <string name="init_bigSmileysOnMap">Big icons for logged caches</string>
-    <string name="init_summary_bigSmileysOnMap">If enabled big emoticons will be shown for logged caches on the map instead of the cache type icon (requires restart).</string>
+    <string name="init_bigSmileysOnMap">Big icons</string>
+    <string name="init_summary_bigSmileysOnMap">If enabled, symbols like the \'found smiley\' or the \'corrected coordinates\' marker will be shown enlarged on the map instead of the cache type icon (requires restart).</string>
     <string name="init_global_wp_extraction_disable">Disable waypoint extraction</string>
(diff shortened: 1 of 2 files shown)


ztNFny (commented):
I suggest to use "else if" instead of "if" to connect the 2 cases for bottom-right, makes it easier to read the logic if each position has its own if-block.

ztNFny:
I'm wondering whether the "has corrected" info is really important enough to be displayed in this way.

And it doesn't fit in with the existing functionality of showing big icons for found caches. For a found cache you don't care about it anymore, it's just a smiley on the map, so far so good. But is the equivalent true for a cache with corrected coords, meaning is every cache with corrected coords the same? Could be a solved mystery or a started multi or ...

I'd say this would need a new setting. The existing is "Big icons for logged caches" which doesn't accurately reflect the new behaviour, so at least it'd need to be renamed. But imho it's a completly different functionality which shouldn't be tied into the existing one.

(disclaimer: I'm not using big icons mode)

fm-sys:
> I'm wondering whether the "has corrected" info is really important enough to be displayed in this way.

At least it was suggested in  #10924 and no one has disagreed. And it's the same as on GC.com ;-) 

> And it doesn't fit in with the existing functionality of showing big icons for found caches. For a found cache you don't care about it anymore, it's just a smiley on the map, so far so good. But is the equivalent true for a cache with corrected coords, meaning is every cache with corrected coords the same?

It's kind of a similar meaning. For found caches, you don't need to care about the cache type anymore because the cache is already found. 
For e.g. solved Mysteries or finished Multis, you don't need to care about the cache type as well, because the cache type turns into a normal Tradi as soon as it is solved.

> I'd say this would need a new setting

Those who are really using this setting where fine with connecting these two things into one and the same setting (based on the comments in  #10924). The bigSmiley thing is quite new to cgeo anyway, so it doesn't have a large user base until now. IMHO we shouldn't introduce new settings for such things until there are explicit requests which cannot be solved without a new one.

ztNFny:
Changing the preference name/description should be included with this change though.

fm-sys:
> Changing the preference name/description should be included with this change though.

1:0 for you. Will do so...

Lineflyer (approved):

fm-sys:
I've tried my best to write a short, but meaningful and understandable explanation...

grafik

ztNFny:
Lgtm

PR #12087, opened 2021-11-13
Display of multiple separate tracks in single GPX file (fix #10432)

### LINKED ISSUE #10432: GPX files containing multiple tracks are handled incorrectly
## Bug description 

If you load a GPX that contains multiple tracks, c:geo handles them as if they were a single track and glues them together with a connecting line. See screenshot.

### Expected result 

c:geo should simply draw multiple separate tracks.

moving-bits:
## Description
If a GPX file contains multiple tracks, display them separately instead of connecting them into one track:

|before|now|
|---|---|
|image|image|

(for testing see the GPX file provided in https://github.com/cgeo/cgeo/issues/10432#issuecomment-841788817)


moving-bits:
retest this please

moving-bits:
Let's retry, as the resulting APK seems to be non-parseable.

retest this please

PR #12088, opened 2021-11-13
fix #12062: calculator support chained floats

### LINKED ISSUE #12062: Formula syntax shall support variable chaining in float values
Current formula syntax supports variable chaining for integer values. E.g. expression `A1(B+1)` with `A=5` and `B=3` will evaluate to `514`.

However chaining is not supported for float numbers. E.g. expression `3.1A15` is invalid.

Float chaining shall also be supported. E.g. above example for `A=4` shall evaluate to `3.1415`

This is needed to support calculated waypoints in pure decimal degree lat/lon format

eddiemuc:
fix #12062: calculator support chained floats


eddiemuc:
retest this please

PR #12102, opened 2021-11-15
Owned double size

### LINKED ISSUE #12100: double-size icon for "owned"
### Tell us your idea!

With double-sized icons for found/corrected enabled also show a larger marker for "owned" caches

(From survey)

Related to  #12083

ztNFny:
implements a double-sized "owned" icon and fixes a marker display bug

fixes #12100


ztNFny:
Why didn't the auto-linking of the issue work?

Lineflyer:
> Why didn't the auto-linking of the issue work?

It worked, at least I see the reference.
It might be however, that it will not close the issue after merge as in the upstream merge to `master` Github will probably only consider the commit message which does not contain the trigger word/issue link. Lets see.

ztNFny:
retest this please

PR #12137, opened 2021-11-20
put DownloadManager.enqueue in background in update check

moving-bits:
## Description
As discussed in our last dev call: Put the call to `DownloadManager.enqueue` into background when called by the automatic update check, trying to avoid the `NetworkOnMainThreadException` which occur on some Samsung devices running Android 10.

Attempts to fix the error described #11184, but let's keep the issue open to see if the error is really resolved (as we cannot reproduce it here).

The change should be non-problematic, but I've targeted it to `master` anyway, to not force it into our upcoming release without having seen it in a couple of nightlies before.


moving-bits:
retest this please

eddiemuc:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

PR #12144, opened 2021-11-20
add filtering to view settings

moving-bits:
## Description
Adds filtering capability to "view settings":

image.image


fm-sys:
Good idea! But why not using a normal AndroidX SearchView component which will look not that much self crafted?

If the SearchView is not initialized with a list of items, it will not show a dropdown and you can add a onQueryChangedListener (like we also do in our MainActivity). Exactly what we need here, and even with less code...

moving-bits:
You are right, a consistent look & handling would be preferrable.

I had refrained from implementing `SearchView` component so far, as our implementation in `MainActivity` requires quite much code. But on a second analysis, most of it is coming from the `SearchSuggestionProvider`, which we don't need here :-)

So I gave it another try, and boiled the implementation down to the core needed for the settings' view. It's working nicely now for most parts, and looks much more consistent:

image

(tested it against API 21 and 29). 

But there's one thing left I haven't been able to get to work again:
- start entering a search term => list filters accordingly
- then tap the "enter edit mode" button

Search view and search result gets reset in this case (which has not been the case with the old implementation). The reason for this is that the `SearchView` component calls the `onQueryTextChange` listener with an empty string, effectively resetting the whole search on tapping the button. 

Any idea how to avoid this?

fm-sys:
> Search view and search result gets reset in this case (which has not been the case with the old implementation). The reason for this is that the SearchView component calls the onQueryTextChange listener with an empty string, effectively resetting the whole search on tapping the button.
>
>Any idea how to avoid this?

Currently, we call `invalidateOptionsMenu()` which resets the whole actionbar including the SearchView. Instead, we should only manipulate the `view_settings_edit` / `view_settings_add` items without recreating the entire menu.

Or you can use `searchView.setQuery(text, false)` to recreate the old SearchView state.

---

Another thing: search should be case-insensitive...

|||
-|-
grafik|grafik

moving-bits:
> Currently, we call `invalidateOptionsMenu()` which resets the whole actionbar including the SearchView. Instead, we should only manipulate the `view_settings_edit` / `view_settings_add` items without recreating the entire menu.

Good idea, changed it accordingly (although it seems to require me to either keep a reference to the `Menu` or to several `MenuItem` objects).

> Or you can use `searchView.setQuery(text, false)` to recreate the old SearchView state.

Had tried that first, updating my `currentSearch` in `onQueryTextChange`, but that didn't work, as `onQueryTextChange` at some point gets called with an empty string - probably triggered by `invalidateOptionsMenu()`.

> Another thing: search should be case-insensitive...

yep, this was missing partly, have completed it.

moving-bits:
This seems to be the "lucky PR" today, getting green builds twice in a row... 😆

PR #12148, opened 2021-11-21
fix #12145: move mapsforge svg to own dir, delete leftovers, patch mapsforge

### LINKED ISSUE #12145: Thousands of SVG cache files
### Describe your problem!

I've got 63,702 PNG files named `svg-\d+.png` in `/data/data/cgeo.geocaching/files`. They contain icons. Dates range from the present back to beginning of the year.

### How to reproduce?

…

_No response_

_No response_

### Reproducible

Unclear

### c:geo Version

Doesn't matter, various versions

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #12145: move mapsforge svg to own dir, delete leftovers, patch mapsforge

### DIFF
--- a/main/src/cgeo/geocaching/storage/LocalStorage.java
+++ b/main/src/cgeo/geocaching/storage/LocalStorage.java
@@ -46,6 +46,7 @@ public final class LocalStorage {
     private static final String GEOCACHE_DATA_DIR_NAME = "GeocacheData";
     private static final String OFFLINE_LOG_IMAGES_DIR_NAME = "OfflineLogImages";
     private static final String MAP_THEME_INTERNAL_DIR_NAME = "MapThemeData";
+    private static final String MAPSFORGE_SVG_CACHE_DIR_NAME = "mapsforge-svg-cache";
     private static final long LOW_DISKSPACE_THRESHOLD = 1024 * 1024 * 100; // 100 MB in bytes
 
     //Legacy directory names which should NO LONGER BE OF USE
@@ -60,7 +61,7 @@ public final class LocalStorage {
     private static File externalPrivateCgeoDirectory;
     private static File externalPublicCgeoDirectory;
 
-    private static final int LOCALSTORAGE_VERSION = 2;
+    private static final int LOCALSTORAGE_VERSION = 3;
 
 
     private LocalStorage() {
@@ -189,6 +190,13 @@ public final class LocalStorage {
     @NonNull
     public static File getMapThemeInternalSyncDir() {
         final File dir =  new File(getInternalCgeoDirectory(), MAP_THEME_INTERNAL_DIR_NAME);
+        dir.mkdirs();
+        return dir;
+    }
+
+    @NonNull
+    public static File getMapsforgeSvgCacheDir() {
+        final File dir =  new File(getInternalCgeoDirectory(), MAPSFORGE_SVG_CACHE_DIR_NAME);
         dir.mkdirs();
         return dir;
     }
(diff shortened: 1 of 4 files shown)


eddiemuc:
Update: mapsforge patch (in cgeo/mapsforge) is now included

eddiemuc:
This PR got lucky too - passed CI on first try :-)

eddiemuc:
Update: corresponding PR on mapsforge was already merged to their master. Very quick reaction over there! 

This means that on next mapsforge release we can safely switch to it (and abandoning current special build).

Lineflyer:
@eddiemuc 
Release is done, which means we can merge this to `release`, can't we?

eddiemuc:
Yes we can :-)

main/src/cgeo/geocaching/storage/LocalStorage.java
```diff
@@ -46,6 +46,7 @@
     private static final String GEOCACHE_DATA_DIR_NAME = "GeocacheData";
     private static final String OFFLINE_LOG_IMAGES_DIR_NAME = "OfflineLogImages";
     private static final String MAP_THEME_INTERNAL_DIR_NAME = "MapThemeData";
+    private static final String MAPSFORGE_SVG_CACHE_DIR_NAME = "mapsforge-svg-cache";
     private static final long LOW_DISKSPACE_THRESHOLD = 1024 * 1024 * 100; // 100 MB in bytes
```
Lineflyer, line 50: Sorry for asking to late:
Is this still inside some user-inaccessible area or does this target to our base dir or additional geocache data directory?
IMHO those are cache files and should stay internal (isn't there a specific cache dir assigned to each app by the system).

eddiemuc: Still inside the internal dir but not in the special android cache dir. Same like all the other dirs in the list.

The idea to move stuff to Android cache dir is worth exploring (there's more candidates than just the mapsforge SVG), but maybe worth its own issue

fm-sys: > The idea to move stuff to Android cache dir is worth exploring (there's more candidates than just the mapsforge SVG), but maybe worth its own issue


Yes, that would probably be a good idea!

Even though I have reinstalled cgeo recently, there's already 1.5GB content in the cgeo internal directory again.

Using the caching directory would have the benefit of having an easy way to delete cached files without touching important data like preferences or the DB.

Screenshot_20211122_073508_com android settings_copy_432x936

eddiemuc: I believe the vast majority of the internal space is consumed by downloaded cache data in dir `GeocacheData` (especially images from listings). The cache directory seems to contains just data from Google Maps, leakcanary and some debug stuff.

Lineflyer: @fm-sys 
AFAIK these figures include the "additional geocache data" located in user readable store but not only the internal data (like theme cache, database, etc.)

eddiemuc: @Lineflyer I checked on my device. I have around 5000 caches stored offline, and the total space consumed shown is about 1GB. This is around 200KB per cache. c:geo downloads and stores all images of a listing in the GeocacheData dir (together with some meta-info e.g. etag and last-modified), and this data alone should make up the larger part of the 200KB in average. If someone turns on saving of log images, the data size will surely become much more

Why not put this data in the cache dir?

fm-sys: > @fm-sys
AFAIK these figures include the "additional geocache data" located in user readable store but not only the internal data (like theme cache, database, etc.)

The external/public directory is more than 3GB on my device cause of quite some offline maps. So I believe this folder isn't included in these figures on my device...

> and this data alone should make up the larger part of the 200KB in average. If someone turns on saving of log images, the data size will surely become much more
>
>Why not put this data in the cache dir?

At least if the setting is enabled, pictures shouldn't be deleted automatically by the system. But in all other cases these files should IMHO go into the app-cache directory...

Lineflyer: Lets take the topic "Usage of system cache dir" into an own issue. Will open it.

PR #12174, opened 2021-11-26
make map/tile update info scrollable (fix #12165)

### LINKED ISSUE #12165: Routing/Map update info not scrollable
### Describe your problem!

Picture explains:

Screenshot_20211124-183834_cgeo

### How to reproduce?

Sure, this example shows a rare case (in fact I have a copy of worldwide routing files on my SD) but it might also apply to large fonts.

_No response_

if this view can be made scrollable by adding some parameter we should do so.
In fact. we should think about this for all such dialogs (also in future) to avoid any problems on large fonts (see other issues).

### Reproducible

Yes

### c:geo Version

2021.11.23-NB

### System information

```text
Android 11
Standard font size
```

### Additional Information

_No response_

moving-bits:
## Description
Make the `AlertDialog` window with routing/map update info scrollable

## Additional context
~@Lineflyer 
Can you cross-check (as time permits) - cannot test here, only have 20 or so tiles downloaded :-)~

Just did a retest with font size set to maximum - view is now scrollable as expected.

### DIFF
--- a/main/res/layout/downloader_confirmation.xml
+++ b/main/res/layout/downloader_confirmation.xml
@@ -1,28 +1,33 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:orientation="vertical" android:layout_width="match_parent"
-    android:layout_height="match_parent">
-
-    <TextView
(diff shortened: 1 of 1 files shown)


Lineflyer:
Merging then as I use the nightlies with the device with so much routing files on it.

PR #12183, opened 2021-11-27
rel to #11227: fix gui blockage on quick search

eddiemuc:
rel to #11227: fix gui blockage on quick search

### DIFF
--- a/main/src/cgeo/geocaching/search/SuggestionsAdapter.java
+++ b/main/src/cgeo/geocaching/search/SuggestionsAdapter.java
@@ -49,6 +49,8 @@ public class SuggestionsAdapter extends CursorAdapter {
 
     public void changeQuery(@NonNull final String searchTerm) {
         changeCursor(query(searchTerm));
+        //the following line would get new DB cursor asynchronous to GUI thread. Not used currently (see discussion in #11227)
+        //AndroidRxUtils.andThenOnUi(AndroidRxUtils.networkScheduler, () -> query(searchTerm), this::changeCursor);
     }
 
     private static Cursor query(@NonNull final String searchTerm) {
--- a/main/src/cgeo/geocaching/CgeoApplication.java
+++ b/main/src/cgeo/geocaching/CgeoApplication.java
@@ -42,7 +42,8 @@ public class CgeoApplication extends Application {
 
     @Override
     public void onCreate() {
-        try (ContextLogger ignore = new ContextLogger(Log.LogLevel.DEBUG, "CGeoApplication.onCreate")) {
+        Log.iForce("---------------- CGeoApplication: startup -------------");
+        try (ContextLogger ignore = new ContextLogger(true, "CGeoApplication.onCreate")) {
             super.onCreate();
 
             OOMDumpingUncaughtExceptionHandler.installUncaughtExceptionHandler();


Lineflyer:
This seems to be a correct fail:
https://ci.cgeo.org/job/cgeo-CI-PR-build/293/checkstyleResult/

main/src/cgeo/geocaching/search/SuggestionsAdapter.java
```diff
@@ -48,7 +50,8 @@ public View newView(final Context context, final Cursor cursor, final ViewGroup
     }
 
     public void changeQuery(@NonNull final String searchTerm) {
-        changeCursor(query(searchTerm));
+        //get new cursor asynchronous since this might take a while
+        AndroidRxUtils.andThenOnUi(AndroidRxUtils.networkScheduler, () -> query(searchTerm), this::changeCursor);
```
fm-sys, line 54: Although I have no idea, what the purpose of these different Schedulers are: Is `AndroidRxUtils.networkScheduler` really the most fitting one?

fm-sys: Maybe we could even create a new `DbScheduler` if we want to move all DB calls into the background...

eddiemuc: @fm-sys we might do this. But DB calls should already be done in the background in all places (as it is best practice to put all IO operations away from the main thread because they might be long-running). This is maybe not completely implemented in c:geo but should be... 

So this would not be just about creating a scheduler but to establish a concept for putting DB calls in the background everywhere across c:geo. This is definitely worth its own isssue.

Also I don't even understand the meaning of the existing schedulers and when to use which. So the first step to approach this would be to clarify their meaning and when they should be used.

eddiemuc:
> This seems to be a correct fail: https://ci.cgeo.org/job/cgeo-CI-PR-build/293/checkstyleResult/

Normal Codacy complaints :-)

ztNFny:
I think in its current state this does more bad than good. Starting c:geo and quickly (within what seems to be the db cleanup window) opening search that list is then never populated. If the list was populated after the DB is ready ok, but now it's never populated and I have to return to main screen and open search again.
I also had some cases where even after c:geo already running some time no search results would show, happening seemingly random about 5-10% of my attempts.

eddiemuc:
> I think in its current state this does more bad than good. Starting c:geo and quickly (within what seems to be the db cleanup window) opening search that list is then never populated. If the list was populated after the DB is ready ok, but now it's never populated and I have to return to main screen and open search again.
> I also had some cases where even after c:geo already running some time no search results would show, happening seemingly random about 5-10% of my attempts.

It is true that the reason for db blockage has to be further investigated. However the PR still makes sense:
- db calls (or more general: IO calls) should never be executed on main thread -> this is fixed by this PR
- suggestion list is updated as soon as db result is available. In the meantime the user can at least type and start a search. He only does not have the suggestion list available.

Also I tweaked logging a bit for future analysis, and I don't want to miss this

ztNFny:
> suggestion list is updated as soon as db result is available

That's not the case for me, the list never populates (without typing anything)

PR #12188, opened 2021-11-28
fix #11941: allow loading more caches for new gc search and oc

### LINKED ISSUE #11941: No possibility to load more caches on search results
### Describe your problem!

Not sure whether its a bug or a feature request.
Was reported by users on support mail and reproducible for me.

Since we introduced the "new search" there seems to be no possibility to load more than the 200 results.

### How to reproduce?

Users report the need to do that for e.g. following scenarios:
- Search for hidden by a user which has hidden more than 200 caches and your desired caches is not within the first 200
- Search for a keyword which returns more than 200 results and your expected result is not within the first 200

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2021.10.08

### System information

_No response_

### Additional Information

A bit related to #11489

eddiemuc:
fix #11941: allow loading more caches for new gc search and oc

This PR extends the "load more caches" functionality to the new GC search as well as OC search. Initially 200 caches are loaded per (active) connector, each "load more" will try to load 50 more caches from each connector

### DIFF
--- a/main/src/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/cgeo/geocaching/CacheListActivity.java
@@ -147,7 +147,6 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
     private static final int CACHE_LOADER_ID = 5; //arbitrary number, but must be fixed
     private static final String EXTRAS_NEXTPAGE  = "extras_nextpage";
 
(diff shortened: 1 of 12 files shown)


eddiemuc:
GC legacy search of course still returns 20 caches per "load more" since this is hard-bounded to web gui

kumy:
retest this

Issue #12146, opened 2021-11-21
Calculated Waypoints refactoring: can we remove non-plain calculator formats?


eddiemuc:
### Describe your problem!

I analyzed the calculated waypoint code in order to find a good way to introduce global variables (see  #11520). Current implementation however is very very complex and I feel myself unable to introduce the concept without doing a large refactoring in this area.

Large complexity parts are introduced by the "button per digit" logic of calculated formats other than "Plain". Here the user can only decide for each digit place whether it is a single-digit, a single-letter-var (no reusages of same var possible though) or a "blank". This can look then e.g. like this:

![image](https://user-images.githubusercontent.com/6909759/142756365-c7e920f3-fb52-4cae-a4cf-516e339afb05.png)

To give you a feeling about the complexity this creates in code: the input above is then saved to the database as a "state object" like the following (and there's lots  of code decicated to interpret, save and load this stuff):

```json
[...]

PR #12190, opened 2021-11-28
rel to #12146: introduce overflow character _ into calculator

eddiemuc:
rel to #12146: introduce overflow character _ into calculator

In order to replace the special "blank" symbol (used as an overflow/spillover placeholder) needed for button-bound calculated variable, we need to introduce same capacity to the new calculator engine. character _ is used for that


eddiemuc:
Idea may need rethinking

eddiemuc:
After experimenting with current cache calculator I think it works out.

eddiemuc:
...since the behavior of current calculator for waypoints with blank before a number is undefined anyway :-)

Issue #12085, opened 2021-11-13
Improve screen reader compatability


Lineflyer:
### Describe your problem!

A geocacher, who was trying to help a visual impaired geocacher to setup his device with c:geo, contacted us on support mail as he ran into various small problems in c:geos compatibility with screen readers.

I would like to work on this / check what can be improved and therefore opened this issue. 

After a first check it seems many things reported have implicitly be solved by recent design changes and introduction of bottom navigation. Therer are however still  some problems caused by some screen items in c:geo not having proper labels. 
I will read into this topics, test it and collect a list of ToDo here.

That could be something I could try to fix in our code as it mostly involves our XMLs (I guess).

### How to reproduce?

to be completed with a list of problems

### Reproducible

Yes

### c:geo Version

2021.10.08

Lineflyer:
From users mail:

Bugs found.
1. While C:Geo is just started, You are on the 6 button screen. 
On this screen, the 6 buttons are spoken and that is nice, but the "Nearby" button diverse from the others. 5 of them are spoken with the name first, while the picture of the buttons is a next selection that is spoken as "no label, knob". Only Nearby is spoken as "no label, knob" and then Nearby. (different order)

2. While in a cache, the kompas knob in the top is read as "no label, knob".

3. The attributes are only spoken as "no label" and the column number.

4. All the attributes do have the same display name.

5. If there is a icon on the next line this is spoken as "no label", but no matrix place.

6. The buttons for refresh, delete, add to watchlist and choose lists are only spoken as "knob"

---

Comment from my side:
I played around a bit and it seems there are some possibilities/standards:

- For buttons we can assign a property `android:text` and the screen reader will use that (in my tries it then reads `ContentOfString`+"Button"). There seems to be another possibility though, which is not in the XML itself but in the java code. In such cases the screen reader says "Button"+`ContentOfString`. (Need to look up an example and the corresponding code to show what I mean). Not sure, which of those is the preferred method.
- For ImageViews (like when tapping on a log image) we can add a property `android:contentDescription` which will be read then. 

Some other guidelines I found so far (needs to be checked whether we follow those):
- For EditTexts screen readers use the property `android:hint` 
- For images which are just for decoration a ` android:contentDescription=NULL` should be used, to make screen readers ignore it (alternatively the value `android:importantForAccessibility=no` can be used).

If anyone has experience with such things, feel free to provide some info and/or useful links. Otherwise I will start digging into it by reading some Android documentation.

PR #12193, opened 2021-11-28
attributes sorting in cache details and attribute filter (fix #12189)

### LINKED ISSUE #12189: Sort attributes in cache details and attribute filters
### Tell us your idea!

Reminder to myself

Sort attributes in cache details and attribute filters in the same way as it's done for attribute overview implemented with #12140 

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Sort attributes in cache details page and in attribute filter according to the attribute categories:

|cache details: icons|cache details: texts|attribute filter|
|---|---|---|
|image|image|image|

(also fixes one category assignment)

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1971,6 +1971,7 @@
     <string name="attribute_historic_no">No abandoned structure / No ruin</string>
     <string name="attribute_blind_people_yes">Handicap: Blind</string>
     <string name="attribute_blind_people_no">No handicap: Blind</string>
+    <string name="attribute_unknown">Unknown attribute</string>
 
     <!-- cache attribute categories -->
     <string name="cat_tools">Tools required</string>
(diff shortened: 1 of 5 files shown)


bekuno:
For `cache details: texts` - is it possibe to use different size or font style for category and attribute?

moving-bits:
> For `cache details: texts` - is it possibe to use different size or font style for category and attribute?

I've added underlining to the headings, see here:
image

(I had also tried bold, but this was a bit bulky, and did not clearly enough separate from the nex tline.)

> in text-view would it be possible to display "[icon] text"? helps learning the icons (though it's probably a separate feature request)

Would be a separate FR, as this would probably create the need for it's own list adapter etc.
But what one could try would be to add long press action / hints to the attribute icons, as it's done in the filter.

ztNFny:
> But what one could try would be to add long press action / hints to the attribute icons, as it's done in the filter.

I think that's not suited here, a short-click to show the text version is easier than holding to get the tooltip for only 1 attribute (and I think most users don't even know about holding for a tooltip)

moving-bits:
Too late :-)

But I think it does not harm either. I've added the tooltip to the `AttributesGridAdapter`, thus it is used for both cache details and attributes overview.

image.image

Lineflyer:
> But what one could try would be to add long press action / hints to the attribute icons, as it's done in the filter.

In this context please also try to check the screen reader compatibility.
While briefly looking into #12085 I have noted, that the attribute icons are currently not backed by some textual description, thus the screen reader is not able to read those.

moving-bits:
> In this context please also try to check the screen reader compatibility. While briefly looking into #12085 I have noted, that the attribute icons are currently not backed by some textual description, thus the screen reader is not able to read those.

added

PR #12200, opened 2021-11-29
show status error messages on refresh (fix #12194)

### LINKED ISSUE #12194: Failed "Refresh" does not give an error message, even if server request results in 404 aka "DNF"
### Describe your problem!

Currently, events are often canceled (for very well known reasons). And at least in BaWü the typical procedure is that the owner writes an announcement, explaining the reasons, and deactivates the listing, which of course is perfectly fine. But right after this the listing will additionally be retracted by the reviewer. Which IMO is kind of weird, cause this way even the cancellation info gets lost. And users who try to (re)open such a retracted listing just get a "DNF" - like here: [GC8MAEP](https://www.geocaching.com/geocache/GC8MAEP_abgesagt-weihnachtsbaumschmucken-wird-verschoben), without even knowing if they just mistyped the GC code ... or whatever ...

However, things get even worse when someone has previously saved such a listing offline - cause if you refresh an offline stored listing which has been retracted, c:geo just does nothing at all.  Although the server gives a 404, c:geo just completely ignores this and goes on as if everything is perfectly fine. The one and only place where you can see that s/th has gone wrong is the "Stored in device" text, which remains unchanged, no matter how often you (try to) refresh.  But if you don't check it, you will never ever see that this event has been cancelled. Or that this cache does not exist anymore.

### How to reproduce?

- save any cache offline which is prone to be retracted
- try to refresh it, after it has been retracted

- listing will not be updated
- nothing changes
- no error message will be displayed

An error message - or at least a toast message - should be displayed,
that this listing was not found on the server and that it has not been updated.
Perhaps even a hint may be displayed to enforce the user to check the listing by opening it in browser ...

### Reproducible

Yes

### c:geo Version

2021.11.28-NB-90aef0e

### System information

_No response_

### Additional Information

_No response_

fm-sys:
Show status error messages when refreshing a cache

grafik

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -2651,6 +2651,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         public void handleRegularMessage(final Message msg) {
             if (msg.what == UPDATE_LOAD_PROGRESS_DETAIL && msg.obj instanceof String) {
                 updateStatusMsg(R.string.cache_dialog_refresh_message, (String) msg.obj);
+            } else if (msg.what == UPDATE_SHOW_STATUS_TOAST && msg.obj instanceof String) {
+                showToast((String) msg.obj);
             } else {
                 notifyDataSetChanged(activityRef);
             }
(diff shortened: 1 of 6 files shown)


fm-sys:
In my quick test, I haven't detected any side effects. However, a review would be welcome...

fm-sys:
Let's test it in the nightlys...

PR #12201, opened 2021-11-29
make all layers used in layer drawables 36dp to properly center them

### LINKED ISSUE #12173: "archived" overlay has wrong size
Screenshot_20211123-215424__01

ztNFny:
fixes #12173

LayerDrawable is capable only of px offsets but with different screen densities float offsets would be needed to ensure all elements that make up a cache icon are properly centered.
To work around this limitation all drawables have been resized to 36dp so that no offset has to be applied and the position from the drawable itself is used.

Some places in code had to be updated to handle the new icon size.

At the same time I replaced colors in cache icons with @color references to make maintanance easier.

This should not introduce any visible changes other than fixing the slight icon offsets.


ztNFny:
it seems that auto-linking of issues is broken for PRs against release. I had the same happen recently
@moving-bits something you could look at?

Lineflyer:
The fixes keyword needs to be contained in the commit message to ensure auto-closing on merge.
Reason is, that the issue is auto-closed by Github only once a fix is merged onto the default branch (`master`). As this will only happen later (when merging `release` back into `master`) having it in the PR content or PR title does not do the job but only the commit message.

ztNFny:
> The fixes keyword needs to be contained in the commit message to ensure auto-closing on merge.

For auto-close that's clear. But I'm missing the auto-linking of the issue by mentioning it with "fixes #<issue>"

PR #12206, opened 2021-11-30
fix missing display of owner / favorite count (fixes #12205)

### LINKED ISSUE #12205: Favorite count and owner missing in cache details
### Describe your problem!

Starting with todays nightly 2021.11.30-NB the rows for favorite and owner are empty in cache details.

### How to reproduce?

Open a cache, look on the details tab

No favorite info, no owner info

Favorite and owner info should be shown

### Reproducible

Yes

### c:geo Version

2021.11.30-NB

### System information

_No response_

### Additional Information

_No response_

ztNFny:
"value" field was hidden for owner and favorite count

@moving-bits this was caused by #12191, could you please check whether the fix makes it work as you intended?

### DIFF
--- a/main/res/layout/cache_information_item.xml
+++ b/main/res/layout/cache_information_item.xml
@@ -37,7 +37,7 @@
             android:textSize="@dimen/textSize_detailsPrimary"
             android:textIsSelectable="false"
             android:textColor="@color/colorText"
-            android:visibility="gone"
+            android:visibility="visible"
             tools:text="value"/>
 
         <RatingBar


moving-bits (changes requested):
Thanks for diving into this.

> "value" field was hidden for owner and favorite count
> @moving-bits this was caused by #12191, could you please check whether the fix makes it work as you intended?

The field is hidden as default deliberately, as with adding the layout mediator (to be able to have wrapping within a single `LinearLayout`) some texts started to behave strange, not aligning to the left anymore, although all layout parameters pointed in that direction. After setting visibility for all fields to "gone", and only reenabling visibility for non-empty fields, everything worked as expected. 

Thus the fix for #12205 would be to `setVisibility(View.VISIBLE)` for the methods in `CacheDetailsCreator.java` involved in displaying favorite count and owner. (Or find the "magic parameter" for the view helper to avoid those side-effects, if such a parameter exists.)

ztNFny:
@moving-bits I'll retract this PR then, please submit a proper fix

moving-bits:
@ztNFny 
Do you want to dive into it, or shall I take over?
(Me being the one who introduced the bug in the first place.)

moving-bits:
> @moving-bits I'll retract this PR then, please submit a proper fix

ah, ok, will do

PR #12208, opened 2021-11-30
fix values not being displayed (fix #12205)

### LINKED ISSUE #12205: Favorite count and owner missing in cache details

moving-bits:
## Description
Fixes missing display of
- cache details: owner, favorite count
- trackable details: origin

(#12205)

Also fixes NPE on tapping some of the detail entries (geocode, coordinates etc.) (#12211)

### DIFF
--- a/main/res/layout/cache_information_item.xml
+++ b/main/res/layout/cache_information_item.xml
@@ -37,7 +37,6 @@
             android:textSize="@dimen/textSize_detailsPrimary"
             android:textIsSelectable="false"
             android:textColor="@color/colorText"
-            android:visibility="gone"
             tools:text="value"/>
 
         <RatingBar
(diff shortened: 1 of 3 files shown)


ztNFny:
Why this way around (set visibility to visible) and not the other (set it to gone)? Aren't there less cases where it's gone than visibile?

moving-bits:
> Why this way around (set visibility to visible) and not the other (set it to gone)? Aren't there less cases where it's gone than visibile?

Not sure about this, would have to check. But as I had started yesterday with making all fields `gone`, I simply finished it down the same road.

moving-bits:
Fixes both #12205 and #12211, and also inverses the logic from "disable all/enable on demand" to "disable on demand"

PR #12227, opened 2021-12-03
quick-load geocode from clipboard text

### LINKED ISSUE #12169: quick-load geocode from clipboard text
### Tell us your idea!

I often get messages like "I'm at cache GCxxxxxx, do you have a hint?" upon which I look up my cache notes in cgeo. The required steps to do that are:
1. Copy message
2. Paste message somewhere so i can select only the GC code (depends on the messenger)
3. In cgeo.main screen click search
4. Long-click the search field -> paste
5. On keyboard click the search icon

It'd be nice to have a more convenient way to load a cache whichs code is in clipboard.

Next to the search field/in main menu add a "load from clipboard" action that takes the current clipboard and, of it's more than just a geocode, extracts a geocode from it, then loads that cache. This would be similar to the existing "from clipboard" function in defining a waypoint.
image

### Is there an alternative?

_No response_

### Additional Information

Geocode recognition should look at the enabled connectors to see what a geocode could look like to avoid false recognitions. E.g. with only gc connector enabled OC123 shouldn't be recognized as geocode.

ztNFny:
## Description
Improves loading a geocode from clipboard by "recognizing" geocodes in text.
There's 3 parts to this commit:
1. fixes to the pattern(s) used for recognizing a GC-code in URLs
2. extension of the existing functionality in "Search" activity to auto-fill the geocode field with a geocode from clipboard. This now works not just for plain geocodes but also geocode-urls (e.g. https://coord.info/GC123) and geocodes contained in text ("Check out GC123"), see https://github.com/cgeo/cgeo/blob/3e311ec09ffc46158781fa5572733639495b1d3f/tests/src-android/cgeo/geocaching/connector/gc/GCConnectorTest.java#L115 for some more text examples
image
3. add a "paste" button to the quicksearch field in main activity. The button only shows up if clipboard contains a geocode (either pure, url or in text) and directly loads that geocode
image

## Related issues
fixes #12169

### DIFF
--- a/main/res/drawable/ic_menu_paste_search.xml
+++ b/main/res/drawable/ic_menu_paste_search.xml
@@ -1,0 +1,14 @@
+<!-- taken from https://fonts.google.com/icons?selected=Material%20Icons%3Acontent_paste_search%3A -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24"
+    android:tint="?android:textColorPrimary">
+  <path
+      android:pathData="M5,5h2v3h10V5h2v5h2V5c0,-1.1 -0.9,-2 -2,-2h-4.18C14.4,1.84 13.3,1 12,1S9.6,1.84 9.18,3H5C3.9,3 3,3.9 3,5v14c0,1.1 0.9,2 2,2h5v-2H5V5zM12,3c0.55,0 1,0.45 1,1s-0.45,1 -1,1s-1,-0.45 -1,-1S11.45,3 12,3z"
+      android:fillColor="#000000"/>
+  <path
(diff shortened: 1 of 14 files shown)


ztNFny:
Codacy complains about NPath complexity, but that was already bad before, nothing that could be solved as part of this PR.

moving-bits:
> Codacy complains about NPath complexity, but that was already bad before, nothing that could be solved as part of this PR.

Is there some meaningful way to reduce NPath complexity for those two methods?
If so, it would be nice if you could dive into it and improve the quality of our code base (as time permits). 
Otherwise I propose to add something like `@SuppressWarnings("PMD.NPathComplexity") // split up would not help readability` as we have done in other places (where it was applicable) in the past.

ztNFny:
I've very slightly decreased complexity in `onCreate` (it's way less complex than Codacy thinks somehow) and split `instantSearch` in 2

moving-bits:
Search tab crashes if you have a text in clipboard containing a TB code, e. g. "Any news on TB9GB8G?"

ztNFny:
> Search tab crashes if you have a text in clipboard containing a TB code, e. g. "Any news on TB9GB8G?"

fixed. Code was still using String.isEmpty() instead of StringUtils.isEmpty()

moving-bits (approved):
LGTM. Thanks for diving into this and for fixing the findings.

PR #12231, opened 2021-12-03
reenable SAF folder selection in settings (fix #12230)

moving-bits:
## Description
- enable SAF related preference entries to open the SAF folder selection
- store selected new folder
- and update UI info accordingly

@eddiemuc 
Can you cross-check if I did miss anything regarding SAF folder selection etc.? Thanks!

PR #12234, opened 2021-12-03
[refactor settings] cleanup and restore missing functionality

fm-sys:
Follow-up of #11298: cleanup and restore missing functionality

- directly update authorization preference to better reflect current login state
- introduce some generic helper methods
- remove outdated WIP comments
- handle map source changes correctly

### DIFF
--- a/main/src/cgeo/geocaching/settings/TokenPreference.java
+++ b/main/src/cgeo/geocaching/settings/TokenPreference.java
@@ -54,7 +54,6 @@ public class TokenPreference extends AbstractClickablePreference {
 
     @Override
     protected OnPreferenceClickListener getOnPreferenceClickListener(final SettingsActivity activity) {
-        //activity.setAuthTitle(tokenMapping.prefKeyId);
         return preference -> {
             if (tokenMapping.authActivity != null) {
                 final Intent authIntent = new Intent(preference.getContext(),
@@ -65,7 +64,6 @@ public class TokenPreference extends AbstractClickablePreference {
             }
             return false; // no shared preference has to be changed
         };
-
     }
 
     @Override
--- a/main/src/cgeo/geocaching/settings/fragments/PreferenceMapFragment.java
+++ b/main/src/cgeo/geocaching/settings/fragments/PreferenceMapFragment.java
@@ -4,7 +4,9 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.downloader.DownloadSelectorActivity;
 import cgeo.geocaching.maps.MapProviderFactory;
 import cgeo.geocaching.maps.interfaces.MapSource;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.settings.SettingsActivity;
+import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.ShareUtils;
 import static cgeo.geocaching.utils.SettingsUtils.initPublicFolders;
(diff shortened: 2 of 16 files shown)


moving-bits:
I'm already working on the first finding - ColorPickerPreference -, and the second finding (related to setting SAF folders) should be taken care of with #12231.

moving-bits:
Thanks for taking the time to collect those places. Alas, and no offense meant, this list does not really help - it's a mere collection of code lines that need some cleanup, it does not necessarily mean that there is some functionality missing (and even if so, it does not say which and what part of it), and it's hard to track progress on it (some of the findings are already outdated).

If you find the time it would be great if you could do a side-by-side comparison of old settings and new settings and check for missing functionality or bugs, settting by setting, and open separate issues for any findings, describing where something's missing and what the expected behavior would be. This would really help to tackle the last loose ends.

fm-sys:
> Thanks for taking the time to collect those places. Alas, and no offense meant, this list does not really help - it's a mere collection of code lines that need some cleanup, it does not necessarily mean that there is some functionality missing

That's 100% true. The main idea of this collection was to create a list of places where some cleanup is necessary - not to create a full overview of missing functionality. 

BTW, I just rebased the PR to better reflect the current state.

For missing functionality, quite some issues were created some days ago. But a 100% one-by-one comparison is nearly impossible...

moving-bits:
> For missing functionality, quite some issues were created some days ago.

Which was a great help for me and lead to most of them being fixed already, see https://github.com/cgeo/cgeo/projects/21, although there have grown some new yesterday :-)

main/src/cgeo/geocaching/settings/TokenPreference.java
```diff
@@ -54,6 +54,7 @@ public TokenPreference(final Context context, final AttributeSet attrs, final in
 
     @Override
     protected OnPreferenceClickListener getOnPreferenceClickListener(final SettingsActivity activity) {
+        //todo
         //activity.setAuthTitle(tokenMapping.prefKeyId);
```
fm-sys, line 58: R.string.settings_reauthorize / R.string.settings_authorize switch

More occurrences of `setAuthTitle` (see below) generalize?

main/src/cgeo/geocaching/settings/fragments/PreferenceMapFragment.java
```diff
@@ -56,6 +56,7 @@ private void initMapSourcePreference() {
         }
         prefMapSources.setEntries(entries);
         prefMapSources.setEntryValues(values);
+        // todo
```
fm-sys, line 59: Selecting combined as map source doesn't have a direct effect, related?

main/src/cgeo/geocaching/settings/fragments/PreferenceServiceGeocachingSuFragment.java
```diff
@@ -26,15 +26,15 @@ public void onCreatePreferences(final Bundle savedInstanceState, final String ro
             return true;
         });
 
-        // TODO
+        // TODO - why todo?
         final Preference auth = findPreference(getString(R.string.pref_fakekey_su_authorization));
         if (auth != null) {
             auth.setTitle(
                 getString(Settings.hasOAuthAuthorization(R.string.pref_su_tokenpublic, R.string.pref_su_tokensecret)
                     ? R.string.settings_reauthorize : R.string.settings_authorize));
```
fm-sys, line 34: Ah, probably this should be extracted to a generalized Utility method..

main/src/cgeo/geocaching/settings/fragments/PreferenceMapFragment.java
fm-sys: Old code:
```
-        if (isPreference(preference, R.string.pref_mapsource)) {
-            // reset the cached map source
-            MapSource mapSource;
-            try {
-                mapSource = MapProviderFactory.getMapSource(stringValue);
-            } catch (final NumberFormatException e) {
-                Log.e("SettingsActivity.onPreferenceChange: bad source id '" + stringValue + "'", e);
-                mapSource = null;
-            }
-            // If there is no corresponding map source (because some map sources were
-            // removed from the device since) then use the first one available.
-            if (mapSource == null) {
-                mapSource = MapProviderFactory.getAnyMapSource();
-                if (mapSource == null) {
-                    // There are no map source. There is little we can do here, except log an error and
-                    // return to avoid triggering a null pointer exception.
-                    Log.e("SettingsActivity.onPreferenceChange: no map source available");
-                    return true;
-                }
-            }
-            Settings.setMapSource(mapSource);
-            preference.setSummary(mapSource.getName());
-        }

fm-sys (commented):
I've browsed a bit through the code... And made some observations

main/src/cgeo/geocaching/settings/fragments/PreferenceMapFragment.java
fm-sys: Also show the current selection as summary...

fm-sys (commented):
I've done some more analysis. I will work on the findings as time permits...

fm-sys:
This PR should be ready to merge...

PR #12236, opened 2021-12-03
migrate ColorpickerPreference to AndroidX (fix #12177)

### LINKED ISSUE #12177: Refactor settings: Reenable color selectors
### Describe your problem!

This is more or less a reminder to ourselves, as we actively decided to disable the color selectors for now.

### How to reproduce?

- enter settings => map
- scroll to the end, tap on one of the colored fields

an (nearly) empty dialog window appears

a color selector window should appear, allowing to change the assigned preference value

### Reproducible

Yes

### c:geo Version

dev-build from current state of #11298

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Migrates the ColorpickerPreference to be compatible with AndroidX preferences

### DIFF
--- a/main/src/cgeo/geocaching/settings/ColorpickerPreference.java
+++ b/main/src/cgeo/geocaching/settings/ColorpickerPreference.java
@@ -1,32 +1,36 @@
 package cgeo.geocaching.settings;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.DisplayUtils;
 
 import android.content.Context;
+import android.content.DialogInterface;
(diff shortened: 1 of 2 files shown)


TacoTheDank:
Kinda wondering why a library isn't used instead? There's quite a few modern libraries that have ColorPickerPreferences.

moving-bits:
Yeah, the good old (and important) question of "make or buy" ;-)
I had done quite some research and some tests back then, before Colorpicker got implemented, but finally at that time it showed to be easier to implement that small functionality on our own than integration efforts would have been.

TacoTheDank:
Hmm, I see. Ok! 👍

Issue #12230, opened 2021-12-03
Refactor settings: Reenable selecting SAF folders


moving-bits:
### Describe your problem!

Currently, selecting new SAF folders is disabled, tapping on the respective preference entries are simply ignored.
And even if the SAF selector would be started, its result would not be processed.

Relevant for all SAF folders (base folder, map, map themes, routing tiles, ...)

### How to reproduce?

Go to settings => map and tap on "Folder with offline maps"

### Actual result after these steps?

nothing happens

### Expected result after these steps?

SAF folder selector should open, a selected folder should be processed accordingly, and the new setting's value be presented in the UI.

### Reproducible

Yes

### c:geo Version

current master

PR #12246, opened 2021-12-05
reenable setting base folder (rel. to #12230)

moving-bits:
## Description
Setting "base" folder was still missing in settings - this PR reenables it.

image


moving-bits:
Superseded by #12247

PR #12252, opened 2021-12-05
update OpenAndroMaps download URL for V5

ztNFny:
download URL for OpenAndroMaps changed


Lineflyer:
I noticed the same while checking their website.
Have installed Germany V5 map since a few days and its running smooth.

PR #12253, opened 2021-12-05
Restore old SeekbarPreference capabilities

moving-bits:
## Description
Restoring some old settings' code, upgrading it to AndroidX, to reenable our former `SeekbarPreference` capabilities:
- enable setting value by tapping on the value field right beside the slider
- enable logarithmic scaling (where applicable)
- enable value conversion (where applicable)
- enable displaying texts for min/max value in `BackupSeekbarPreference`

### DIFF
--- a/main/res/layout/preference_seekbar.xml
+++ b/main/res/layout/preference_seekbar.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingLeft="16dp"
+    android:paddingRight="16dp"
+    android:paddingTop="8dp"
+    android:orientation="vertical"
+    tools:context=".settings.SettingsActivity" >
+
+    <TextView
+        android:id="@+id/preference_seekbar_label_view"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="left"
+        android:visibility="gone"
+        android:gravity="left"/>
+
+    <LinearLayout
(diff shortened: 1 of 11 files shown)


fm-sys:
IMHO this looks like a bit too much code for "just adding a onClickListener to the text view" (and some other small features). Why can't we extend the androidx SeekbarPreference class and use their code and layout files instead of creating everything ourselves?

moving-bits:
If you have a look at the code, you'll notice that about 50% of changed lines are XML files, all of which (except the first) would still be needed if we could directly derive from AndroidX `SeekbarPreference`. The rest of the code is mostly doing the different conversions, which are needed anyway. But AFAIK we even cannot directly inherit from `SeekbarPreference` as some of the layout fields either do not exist (prefix) or are not exposed.
And most important (for me): Nearly all the code of this PR did already exist in c:geo, I only reactivated it and changed from `onBindDialogView` to `onBindViewHolder`, which saved me quite some time I would have needed for researching and implementing a new solution from scratch :-)

fm-sys:
The problem is that other built-in AndroidX features like "step values" are (currently) not possible with this PR which would be a step backwards as everything needs to be reinvented by ourselves...

So at least I would like to see the "step value support" :-)

Another idea when we go the way to implement everything ourselves:
Instead of using a separate `BackupSeekbarPreference` class maybe we can add two optional generalized fields `lowerEndLabel` and `upperEndLabel`. So this could also be used as well for other sliders, where we currently explain in the summary what "0" means.

moving-bits:
> The problem is that other built-in AndroidX features like "step values" are (currently) not possible with this PR which would be a step backwards as everything needs to be reinvented by ourselves...

We do use an AndroidX `SeekBar` component here, would need to look into it more deeply to check whether we can access the step values as well.

But this is a feature request, IMHO - we did not have step values in settings before, neither did we have lower and upper end labels. This PR is about getting settings migration feature-complete, everything else should be a separate issue.

Issue #10892, opened 2021-06-09
Autofill feature not working (Service login)


Lineflyer:
## Bug description 

I am using the "Keepass2Android" app to manage passwords from a keepass file on my android device. The tool provides an "auto complete" for text input fields based on auto-fill feature of Android to allow auto completion of username/password fields in other apps.

Since start of the redesign it does no longer work for filling the user/pass fields of the geocaching services.

### Steps to reproduce the problem

- Install https://play.google.com/store/apps/details?id=keepass2android.keepass2android&hl=de&gl=US
  (maybe not even needed to see the problem)
- Go to c:geo Menu - Settings - Services - geocaching.com - "Update authentication" (or any other geocaching service login)
- Tap on the user name field to invoke the system context menu, tap on Three-Dot and select "AutoFill"

### Actual result after these steps 

- A toast shows "Content cannot be autofilled"

### Expected result after these steps 

- Autofill should be allowed

## c:geo version 

2021.06.09-NB

## Screenshots 

![grafik](https://user-images.githubusercontent.com/949669/121314310-83e97000-c907-11eb-8c01-ea1b9d21fe28.png)
![grafik](https://user-images.githubusercontent.com/949669/121314340-8c41ab00-c907-11eb-9aa4-a11a1865476c.png)
![grafik](https://user-images.githubusercontent.com/949669/121314375-9368b900-c907-11eb-9f1d-3b64be162261.png)

Lineflyer:
This is how it looks for the Groundspeak app:
![grafik](https://user-images.githubusercontent.com/949669/121314731-f2c6c900-c907-11eb-88a1-9eae42174ba3.png)
![grafik](https://user-images.githubusercontent.com/949669/121314753-f9554080-c907-11eb-9446-304863891118.png)

If K2PA already stored the package name for the user/pass entry (after you once used it from that app) it will either directly offer the correct entry (shot 2) or allow to search it (shot 1):

Shot 1:![grafik](https://user-images.githubusercontent.com/949669/121314912-21dd3a80-c908-11eb-8cd2-b213cbb08961.png)

Shot2:
![grafik](https://user-images.githubusercontent.com/949669/121314967-2dc8fc80-c908-11eb-9515-986562e46443.png)

fm-sys:
Can reproduce the issue. Interesting though: Autofill does still work for EC, only GC seems to be affected

Lineflyer:
[...]

PR #12260, opened 2021-12-07
Try to fix autofill functionality

fm-sys:
@Lineflyer can you test, if...

a) #10892 is still reproducible?
b) the issue is fixed with this PR?


Lineflyer:
Its still reproducible with todays nightly. Can test but only in nightlies, thus merging as we can revert easily if its not helping.

PR #12265, opened 2021-12-07
Bump okhttp from 4.9.2 to 4.9.3 on release branch

bekuno:
from changelog:
- Fix: Don't fail HTTP/2 responses if they complete before a RST_STREAM is sent.

This should go to the release branch to have this bug fix in the next release

supersedes #12158


Lineflyer:
Sounds reasonable and we have some time to test upstream as well.

PR #12300, opened 2021-12-11
allow unlimited number of log templates (fix #12293)

### LINKED ISSUE #12293: enhance log templates
> Also the number of templates is fixed now as it's using preferences, 

That doesn't work well, IMHO. There will always be one who needs one more template, and others will call it unnecessary space consumption...

I can try to implement a dynamic solution if this is wanted.

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/6197#issuecomment-991354046_

fm-sys:
fix #12293

grafik

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -83,11 +83,9 @@
     <string translatable="false" name="pref_supersizeDistance">supersizeDistanceToggle</string>
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_signature">signature</string>
-    <string translatable="false" name="pref_logTemplate1">pref_logTemplate1</string>
-    <string translatable="false" name="pref_logTemplate2">pref_logTemplate2</string>
-    <string translatable="false" name="pref_logTemplate3">pref_logTemplate3</string>
-    <string translatable="false" name="pref_logTemplate4">pref_logTemplate4</string>
-    <string translatable="false" name="pref_logTemplate5">pref_logTemplate5</string>
+    <string translatable="false" name="pref_fakekey_category_logTemplate">fakekey_category_logTemplate</string>
+    <string translatable="false" name="pref_logTemplate_prefix">logTemplate</string>
+    <string translatable="false" name="pref_logTemplate_counter">logTemplate_counter</string>
     <string translatable="false" name="pref_sigautoinsert">sigautoinsert</string>
     <string translatable="false" name="pref_offlinelogs_homescreen">offlinelogs_homescreen</string>
     <string translatable="false" name="pref_trackautovisit">trackautovisit</string>
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -84,7 +84,7 @@
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_signature">signature</string>
     <string translatable="false" name="pref_fakekey_category_logTemplate">fakekey_category_logTemplate</string>
-    <string translatable="false" name="pref_logTemplates">pref_logTemplates</string>
+    <string translatable="false" name="pref_logTemplates">logTemplates</string>
     <string translatable="false" name="pref_sigautoinsert">sigautoinsert</string>
     <string translatable="false" name="pref_offlinelogs_homescreen">offlinelogs_homescreen</string>
     <string translatable="false" name="pref_trackautovisit">trackautovisit</string>
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -84,8 +84,7 @@
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_signature">signature</string>
     <string translatable="false" name="pref_fakekey_category_logTemplate">fakekey_category_logTemplate</string>
-    <string translatable="false" name="pref_logTemplate_prefix">logTemplate</string>
-    <string translatable="false" name="pref_logTemplate_counter">logTemplate_counter</string>
+    <string translatable="false" name="pref_logTemplates">pref_logTemplates</string>
     <string translatable="false" name="pref_sigautoinsert">sigautoinsert</string>
     <string translatable="false" name="pref_offlinelogs_homescreen">offlinelogs_homescreen</string>
     <string translatable="false" name="pref_trackautovisit">trackautovisit</string>
--- a/main/src/cgeo/geocaching/settings/Settings.java
+++ b/main/src/cgeo/geocaching/settings/Settings.java
@@ -163,6 +163,41 @@ public class Settings {
         }
     }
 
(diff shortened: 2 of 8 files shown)


main/res/values/preference_keys.xml
```diff
@@ -82,11 +82,8 @@
     <string translatable="false" name="pref_supersizeDistance">supersizeDistanceToggle</string>
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_signature">signature</string>
-    <string translatable="false" name="pref_logTemplate1">pref_logTemplate1</string>
-    <string translatable="false" name="pref_logTemplate2">pref_logTemplate2</string>
-    <string translatable="false" name="pref_logTemplate3">pref_logTemplate3</string>
-    <string translatable="false" name="pref_logTemplate4">pref_logTemplate4</string>
-    <string translatable="false" name="pref_logTemplate5">pref_logTemplate5</string>
+    <string translatable="false" name="pref_logTemplate_category">pref_category_logTemplate</string>
```
moving-bits, line 85: Can you remove the "pref_" in the string? "pref_" should only be part of the string's name, not of its content.

moving-bits (changes requested):
Nice extension, and running smooth, thanks!

One question, though: How can I remove an existing template? And what will happen, if I, for example, have created three templates, and remove number 1 or number 2 - will probably break the list templates logic.

fm-sys:
> How can I remove an existing template?

Currently this is not possible. But if we want to support this, I guess this could be done via a longClick action. Furthermore, we would need a separate counter prefKey containing the highest created LogTemplate number.

main/res/values/preference_keys.xml
fm-sys: Only for the category or also for the templates itself? For the LogTemplates I had kept the `pref_` for migration purposes, but this feature was only visible in the nightlys for just a couple of days anyway...

fm-sys: Whatever, I will simply change it as migrating will not work anyway when using the counter variable...

fm-sys:
Deleting is now possible via longClick

ztNFny:
I haven't had time to properly review this, but as a general point i don't think a dynamic list of tenplates should be stored as static preferences but instead it should either be in the sqlite db or something like a json array of templates stored in 1 preference key. That way there won't be any problems with deleting/adding/counting.

fm-sys:
> i don't think a dynamic list of tenplates should be stored as static preferences but instead it should either be in the sqlite db or something like a json array of templates stored in 1 preference key. 

Feel free to take the PR over if you like. But I won't have the time to do this myself.

fm-sys:
> That way there won't be any problems with deleting/adding/counting.

In the current implementation, there will be no problems with that as well...

ztNFny:
> Feel free to take the PR over if you like. But I won't have the time to do this myself.

I can take it if you want. My problem with the dynamic implementation was the front-end in Preferences. With that solved the backend is more my kinda thing.

fm-sys:
> > Feel free to take the PR over if you like. But I won't have the time to do this myself.
> 
> I can take it if you want. My problem with the dynamic implementation was the front-end in Preferences. With that solved the backend is more my kinda thing.

Cool, looking forward! How do we want to proceed? Should I merge, so you can start from master, or do you want to push commits directly to my branch?

ztNFny:
I'd rather not merge this into master or we'll have 3 types of template storage in a matter of a few days. I can just cherry-pick from your branch i think, then when it's done push to this PR?

moving-bits (approved):
Thanks for adding deletion and for fixing the pref key value. Adding & deleting works like a charm now.

moving-bits:
@ztNFny 
I guess this PR is ready to serve as a base for you to modify the way of storing the info. Will set "WIP" label here therefore.
If you decide not to switch the storage format, we should remove the "WIP" label here and merge this PR.

fm-sys:
@ztNFny can you estimate when you will work on this? As I would really like to see this PR finally merged...

ztNFny:
Not really. Maybe between Xmas and new years

ztNFny:
> @ztNFny can you estimate when you will work on this? As I would really like to see this PR finally merged...

If you want to merge then go ahead. If storage changes afterwards, well, so be it... As long as it doesn't go into a release but only nightly that's ok I guess

ztNFny:
@fm-sys finally got around to implementing the new storage. Should I push to your branch (would need access) or create a new PR?

ztNFny:
oops, I already had access, just unlocked the wrong SSH key. All pushed now. @fm-sys review welcome

main/res/values/preference_keys.xml
```diff
@@ -83,11 +83,8 @@
     <string translatable="false" name="pref_supersizeDistance">supersizeDistanceToggle</string>
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_signature">signature</string>
-    <string translatable="false" name="pref_logTemplate1">pref_logTemplate1</string>
-    <string translatable="false" name="pref_logTemplate2">pref_logTemplate2</string>
-    <string translatable="false" name="pref_logTemplate3">pref_logTemplate3</string>
-    <string translatable="false" name="pref_logTemplate4">pref_logTemplate4</string>
-    <string translatable="false" name="pref_logTemplate5">pref_logTemplate5</string>
+    <string translatable="false" name="pref_fakekey_category_logTemplate">fakekey_category_logTemplate</string>
+    <string translatable="false" name="pref_logTemplates">pref_logTemplates</string>
```
fm-sys, line 87: @moving-bits had suggested to remove the `pref_` in the key value... See a bit above https://github.com/cgeo/cgeo/pull/12300#discussion_r767251038

main/src/cgeo/geocaching/settings/Settings.java
```diff
@@ -1962,22 +2004,47 @@ public static int getListInitialLoadLimit() {
...
+        }
+        return null;
     }
 
-    public static String getLogTemplatePreview(final String preferenceBaseName) {
-        final ImmutablePair<String, String> template = getLogTemplate(preferenceBaseName);
-        return template.getKey() + ": " + template.getValue();
+    @Nullable
+    public static List<PrefLogTemplate> getLogTemplates() {
+        List<PrefLogTemplate> templates = null;
```
fm-sys, line 2038: Probably we should rather return an empty list instead of `null`, as other code places doesn't seem to check the null case correctly.

fm-sys (commented):
@ztNFny  I've added some suggestions and would like to ask you to check the Codacy warnings...

However, thanks for your work, looks good already 👍

ztNFny (approved):

main/res/values/preference_keys.xml
ztNFny: Most preference use the pref_-prefix so I just reused it. Dunno what makes sense here

fm-sys:
retest this please

fm-sys:
Let's test in real life...

PR #12318, opened 2021-12-13
rel to #11520: global calculated waypoint coordinates: first draft for single-char-dialog, minor improvements

eddiemuc:
rel to #11520: global calculated waypoint coordinates: first draft for single-char-dialog, minor improvements


eddiemuc:
I guess I will soon merge this, the code does not affect productively used features anyway (all features are and will remain being only visible in nightly and dev build)

PR #12321, opened 2021-12-13
D/T markers on cache icons

### LINKED ISSUE #12312: DT indicator on map markers
### Tell us your idea!

add an indicator to show the D / T rating for caches

image

Should be optional and disabled by default.
Only usable position is top-right, in that case the "stored" icon could move to top-center.

_No response_

### Is there an alternative?

_No response_

### Additional Information

as discussed in todays dev call

### LINKED ISSUE #12315: move "owned" marker to upper-left corner
### Tell us your idea!

The "owned" marker (star) is currently shown in upper-right corner - in the same place as "stored". It'd be better placed in upper-left - together with (offline) log icons.

A cache that you own (usually) can't be found by you, so grouping those makes sense.
If you adopt a found cache the owned information becomes more important than found - because unlike a found cache you now need to perform maintanance. Only case that isn't so clear-cut is an attended event, but also in that case owned "winning" vs found is ok.

_No response_

### Is there an alternative?

_No response_

### Additional Information

related to #12312

ztNFny:
Adds (optional) D/T markers to cache icons. Controlled by a setting in "Map" preferences, disabled by default, as discussed in yesterdays call.
Also moves the "owned" marker from upper-right to upper-left (fixes #12315)

image

fixes #12312


main/res/values/strings.xml
```diff
@@ -826,6 +826,8 @@
     <string name="init_summary_useInternalRouting">Use c:geo internal routing to be able to use automatic routing data downloading and other functions</string>
     <string name="init_bigSmileysOnMap">Big icons</string>
     <string name="init_summary_bigSmileysOnMap">If enabled, symbols like the \'found smiley\' or the \'corrected coordinates\' marker will be shown enlarged on the map instead of the cache type icon (requires restart).</string>
+    <string name="init_dtMarkerOnCacheIcon">D/T rating on cache icons</string>
+    <string name="init_summary_dtMarkerOnCacheIcon">Show an marker for the Difficulty and Terrain ratings on cache markers in upper-right corner ("stored" marker is moved to middle) (requires restart).</string>
```
fm-sys, line 830: Can we add a small sentence to the summary which half of the circle belongs to which of both values?

ztNFny: done

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -697,4 +705,41 @@ public static void resetLists() {
...
+    }
+
+    private static LayerDrawable createDTRatingMarker(final Resources res, final float difficulty, final float terrain) {
+        final Drawable background = DrawableCompat.wrap(ResourcesCompat.getDrawable(res, R.drawable.marker_rating_bg, null));
+        final InsetsBuilder insetsBuilder = new InsetsBuilder(res, background.getIntrinsicWidth(), background.getIntrinsicHeight());
+        insetsBuilder.withInset(new InsetBuilder(background));
+
+        final String packageName = CgeoApplication.getInstance().getPackageName();
+
+        Map<String, Integer> dt = new HashMap<>();
```
eddiemuc, line 728: Really necessary to create a map here with two entries? Might it be possible to extract the remainder of the method into a submethod and call it twice instead? This might increase readability

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -697,4 +705,41 @@ public static void resetLists() {
...
+        dt.put("d", (int) difficulty * 10);
+        dt.put("t", (int) terrain * 10);
+
+        for (Map.Entry<String, Integer> p: dt.entrySet()) {
+            if (p.getValue() >= 10) {
```
eddiemuc, line 733: I assume that the value is < 10 for caches without d/t rating (e.g. labs) correct? What happens in this case? Is an empty drawable created?

eddiemuc (approved):
Looks ok codewise. Made some non-critical comments. Thanks for the nice feature!

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
ztNFny: The drawable is created with 2 empty sections as I didn't want to create an empty "0" drawable.

ztNFny: Had to add them now though to be eliminate the map ...

ztNFny:
I replaced the HashMap for drawing the sections with a sub-function, had to add 4 new drawables for that though.
Also added some input sanitation in case some obscure service ever gets the idea of using ratings like 1,25

ztNFny:
Empty ratings look like this:
image

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -91,6 +94,8 @@ public static CacheMarker getCacheMarker(final Resources res, final Geocache cac
             .append(showPin(cacheListType))
             .append(showFloppyOverlay(cacheListType))
             .append(assignedMarkers)
+            .append(cache.getTerrain())
+            .append(cache.getDifficulty())
```
fm-sys, line 98: If we add more and more parameters to the HashMap, does caching makes sense at all? Would it make sense to add the DT value only to the HashMap, if the related setting is enabled?

fm-sys:
> Also added some input sanitation in case some obscure service ever gets the idea of using ratings like 1,25

E.g. Extremcaching allows D/T value between 1-8. Can this be handled without a crash (we don't need special drawables for it though)?

Lineflyer:
What if the circles would be filled from top to down? This would increase the visibility of lower D/T ratings as the color within the circle would then be more distant and not overlapping the cache type icon. I tried to mock it up on the upper icon in that screenshot:
 
image

ztNFny:
> E.g. Extremcaching allows D/T value between 1-8. Can this be handled without a crash (we don't need special drawables for it though)?

Yes, "r" will always be an int between 0 and 50 in steps of 5. A rating of 1,2 would result in 1, 8 in 5.

ztNFny:
> What if the circles would be filled from top to down? This would increase the visibility of lower D/T ratings as the color within the circle would then be more distant and not overlapping the cache type icon. I tried to mock it up on the upper icon in that screenshot:

I tried that, liked the bottom-filling better. Like a glass of water fills from the bottom, just feels more right to me.

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
ztNFny: Caching definitly makes sense. A "standard" situation of viewing a list, then that list as map, then previewing the cache, then opening it would result in 4 times drawing the marker instead of one. So even if the icon is unique it'll still be beneficial.

I'll make those 2 criteria optional though as you suggested.

main/src/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -697,4 +705,39 @@ public static void resetLists() {
...
+        insetsBuilder.withInset(new InsetBuilder(R.drawable.marker_rating_fg));
+        return buildLayerDrawable(insetsBuilder, 4, 0);
+    }
+
+    private static Drawable getDTRatingMarkerSection(final Resources res, final String packageName, final String ratingLetter, final float rating) {
+        // ensure that rating is an integer between 0 and 50 in steps of 5
+        final int r = Math.max(0, Math.min((int) (Math.round(rating * 2) / 2) * 10, 50));
+        final int tintColor = ResourcesCompat.getColor(res, res.getIdentifier("marker_rating_" + r, "color", packageName), null);
+        final Drawable segment = DrawableCompat.wrap(ResourcesCompat.getDrawable(res, res.getIdentifier("marker_rating_" + ratingLetter + "_" + r, "drawable", packageName), null));
+        DrawableCompat.setTint(segment, tintColor);
```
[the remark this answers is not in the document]
ztNFny: > Or don't we, because every segment can only be tinted in one specific color?

Yes, we don't :-)

eddiemuc:
> Empty ratings look like this:
> image
> 

What about showing no circle at all on caches without an rating? (Just an idea, and maybe it even makes the code easier... I would also be fine with leaving it as is)

ztNFny:
> What if the circles would be filled from top to down?

Btw, my first attempt at this was using 1 icon for d and t each, filling them like a clock:
image
(composed image with all layers visible)

While that works great for 1 rating there just isn't enough room for 2 markers.

ztNFny:
I've moved the center for the segments slightly from the middle to the left/right to account for the room lost to the divider bar. This increases the size relative size of 10 and 15 markers slightly.
image

PR #12330, opened 2021-12-14
Update mapswithe-api

### LINKED ISSUE #12175: In our project mapswithme-api the targetSdkVersion is outdated
### Describe your problem!

Google Play requires that apps target API level 29 or higher.

Defined is `targetSdkVersion 28`.
I believe that this is not directly relevant for the c:geo app in the playstore.
Also we must check if code changes are necessary when we switch to 30.

### How to reproduce?

see https://github.com/cgeo/cgeo/blob/acb63d2f6d66ab4ec11c388571f0985563a4e6ed/mapswithme-api/build.gradle#L10

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

current release and master

### System information

_No response_

### Additional Information

_No response_

bekuno:
- switch to target/compile SDK 30
- update build.gradle
- make lint and checkstyle happy

fixes #12175


Lineflyer:
Did you test that the interworking with mapswithme after that change?
Otherwise I would still want to merge but know, that I should try to test it.

bekuno:
Yes, I tested successful opening maps.me from cache detail and from cache lists.
It would be fine if you could also make some tests.

PR #12333, opened 2021-12-15
rel to #11520: calculated coordinates: error string formats, dialog navigation

eddiemuc:
rel to #11520: calculated coordinates: error string formats, dialog navigation


eddiemuc:
Looks somewhat like this now:

image

PR #12355, opened 2021-12-18
set android:exported="true" if this is required

### LINKED ISSUE #12354: AndroidManifest @ Android 12
### Tell us your idea!

Observed while testing for #12307 when set `targetSdkVersion 21`:

~~~
Error: android:exported needs to be explicitly specified for <activity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
~~~

This situation we have in main/AndroidManifest.xml
 - 11 x activity
 - 1 x receiver
 

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

bekuno:
fixes #12354

Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined.

We already set `android:exported="true"` if this is necessary.
Add this value also for the remaining intents.

(edited)

### DIFF
--- a/cgeo-contacts/AndroidManifest.xml
+++ b/cgeo-contacts/AndroidManifest.xml
@@ -19,7 +19,8 @@
         <activity
             android:name=".ContactsActivity"
             android:label="@string/app_name"
-            android:theme="@android:style/Theme.Translucent.NoTitleBar" >
+            android:theme="@android:style/Theme.Translucent.NoTitleBar"
+            android:exported="true" >
             <intent-filter>
                 <action android:name="cgeo.contacts.FIND" />
 
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -231,7 +231,7 @@
             android:label="@string/auth_twitter"
             android:launchMode="singleTask"
             android:windowSoftInputMode="stateHidden"
-            android:exported="false">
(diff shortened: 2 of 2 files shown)


fm-sys:
> Therefore I assume that all others should be set to `false`.

Like seen in the manifest all these activities have intent filters, means they are intended to be launched from outside cgeo. Therefore exported needs to be `true`

fm-sys (changes requested):
Change to `android:exported="true"` as it will otherwise break existing functionality...

fm-sys (approved):
LGTM! Thanks for working on it 👍

moving-bits:
I can try to look into this the next days. IIRC default value is neither "always true" nor "always false", but context-specific. And we should only export those activities which need to be called by other apps.

bekuno:
set DNM until it was verified.

moving-bits (approved):
According to the [docs](https://developer.android.com/guide/topics/manifest/activity-element) `false` is the default "when there are no intent filters" (`true` otherwise). All of the changed activities in this PR do have intents, so `true` would be their default currently, and thus the changes are fine. Thanks for diving into it!

Issue #12340, opened 2021-12-16
User stuck in version upgrade


Lineflyer:
### Describe your problem!

User reported, that after upgrading to 2021.12.13 he is stuck on startup showing this screen:

![image](https://user-images.githubusercontent.com/949669/146450735-a94cb883-ac2f-4b32-ab48-226d66618fcd.png)

He waited 1 hour, not completed until now.

### How to reproduce?

Unclear

### Reproducible

Unclear

### c:geo Version

2021.12.13

eddiemuc:
I assume a restart did not fix this?
Is this a sole case or are there more like this?

eddiemuc:
Checking the code, the observed behaviour would happen if the migration routine got stuck in deleting the old SVG cache directories. Even if those are extremely large, this should not take an hour, so I guess there's something wrong here.

Code part:

![image](https://user-images.githubusercontent.com/6909759/146514442-e4a4b751-103c-4fa4-ace6-9e33352f83ed.png)

@Lineflyer if the problem persists after restart, could you ask the user to place the following log properties in his "logfiles" directory, restart c:geo, waiting for some time (about 20 seconds stuck in above dialog should be nore than enough), then look for a created file with name "cgeo-log-support-(timestamp).txt" in the folder and send this to support?

[log-properties.txt](https://github.com/cgeo/cgeo/files/7733681/log-properties.txt)

Another question: does the user use external storage for c:geo internal data? And if yes, is the SD card inserted into the device?

Lineflyer:
> I assume a restart did not fix this? Is this a sole case or are there more like this?

Sole case until now.

> if the problem persists after restart, could you ask the user to place the following log properties in his "logfiles" directory, restart c:geo, waiting for some time (about 20 seconds stuck in above dialog should be nore than enough), then look for a created file with name "cgeo-log-support-(timestamp).txt" in the folder and send this to support?

Will do.

> Another question: does the user use external storage for c:geo internal data? And if yes, is the SD card inserted into the device?

Will ask :)

Lineflyer:
Request sent.
As I am on vacation starting today it might take some time to be able to check for respsonse.
In case @moving-bits wants to take a look its ticket 757227

Lineflyer:
Log returned by user


```
[...]

PR #12360, opened 2021-12-19
rel to #12340: log migration exceptions, let cgeo continue

eddiemuc:
rel to #12340: log migration exceptions, let cgeo continue


moving-bits:
Let's give it a try in the nightlies (and prepare a RC for the user being stuck in migration)

PR #12367, opened 2021-12-20
DO NOT MERGE, rel to #12340: emergency fix, skip migration

eddiemuc:
rel to #12340: emergency fix, skip migration

This PR is an emergency fix for #12340: it simply SKIPS the migration to local storage version 3 (this means that there will be leftover svg files on the user's device from mapsforge cache. They won't do any harm except they take away device space)

### DIFF
--- a/main/src/cgeo/geocaching/storage/LocalStorage.java
+++ b/main/src/cgeo/geocaching/storage/LocalStorage.java
@@ -481,8 +481,9 @@ public final class LocalStorage {
             //delete theme files from outdated internal theme file dir
             if (currentVersion < 3) {
                 setMigratedVersion(3, "Move Mapsforge SVG Cache Dir");
-                FileUtils.deleteFilesWithPrefix(getInternalCgeoDirectory(), "svg-");
-                FileUtils.deleteFilesWithPrefix(new File(getInternalCgeoDirectory(), "files"), "svg-");
+                //emergency fix for #12340: SKIP actual migration
+                //FileUtils.deleteFilesWithPrefix(getInternalCgeoDirectory(), "svg-");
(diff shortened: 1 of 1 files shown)


SammysHP:
This is labeled "DO NOT MERGE" and "WIP" and still got merged anyway?

fm-sys:
See https://github.com/cgeo/cgeo/issues/12340#issuecomment-999645695 f.

fm-sys:
However, it should IMHO be reverted as soon as the special RC was built...

Lineflyer:
But seemingly its on the wrong branch...

eddiemuc:
> However, it should IMHO be reverted as soon as the special RC was built...

Well we can revert it if you want. However, it will not do any harm anyway. It will bascially skip the deletion of old svg cache files from mapsforge (those were cached in the main folders, now they are cached in a different folder), and I think we can safely assume that those users affected by this have already upgraded anyway.

And leaving it in code might lead to future cases of error #12340 (I have still no idea what causes this and no one sending me back logs...)

Lineflyer:
@eddiemuc Was there another commit which provides more logging? I thought this was the logging PR...but its for skipping thar part. If there is already more logging we can first try this with the user.

Sorry if I missed that, still on vacation and on mobile only.

eddiemuc:
> @eddiemuc Was there another commit which provides more logging? I thought this was the logging PR...but its for skipping thar part. If there is already more logging we can first try this with the user.
> 
> Sorry if I missed that, still on vacation and on mobile only.

No worries. Injust merged, let's just do a normal release and be done with this for now. As said, the majority of users already migrated anyway and new installations of c:geo are not affected anyway.

Note that not deleting the leftover files will not create a worse situation than before anyway (but at least new files are created just one more additional time and not on every restart)

PR #12418, opened 2021-12-27
Add theme legend for Elevate theme and Freizeitkarte base theme

moving-bits:
## Description
This is a first implementation of a map theme legend. Depending on the selected theme, a map legend can be displayed, looking similar to this:

image.image

(All symbols used with permission from the theme authors.)

At first, a small dialog opens where you can select a category (open / close), which displays symbols for this category (e. g. "roads & paths"). Long-tapping on a symbol displays a hint.
This legend works completely offline. Tapping on the "info" button brings you to the more detailed online legend of the respective theme.

Implementation is incomplete yet, therefore setting "WIP" label.


fm-sys:
This looks like a classic use case for an "expandable list" which would provide a cleaner UI

1_RFc8DpwG5OVx3NA066F1_g

ztNFny:
should this be full-screen? I feel like having it as a popup is wasting valuable space for no good reason.

moving-bits:
> This looks like a classic use case for an "expandable list" which would provide a cleaner UI

I'll look into it if I can find some docs for it - parts of material.io components are nearly undocumented, others are not implemented at all for Android...

> should this be full-screen? I feel like having it as a popup is wasting valuable space for no good reason.

I guess both alternatives have their pros and cons. Doing it full-screen would have the potential to show more icons on the same page, doing it as a dialog has the benefit that you can still see parts of the map to have the symbols you are looking for still visible.

moving-bits:
> This looks like a classic use case for an "expandable list" which would provide a cleaner UI

Looks like there is no `<com.google.android.material.ExpandableListView>` for Android, so I continued trying a basic `ExpandableListView` instead to see how that works. Currently it's compilable, but the layout is completely broken.

moving-bits:
This is as far as I got:

- expansion indicators and dividers between categories are shown, but
- the content does not expand on tapping the open/close icon, and
- the size for the `ExpandableListView` stays within the marked box, although it should expand the dialog's height

image

I'll stop pursuing the `ExpandableListView` path for now until I (or someone else) finds some working example using material.io ExpandableListView.

Lineflyer:
> I guess both alternatives have their pros and cons.

IMHO it could be a full-screen, we already have way to many popups/dialogs which are "misused" for not only showing a status message but containing tons of information for which a full-screen is more adequate.

Maybe this also solves your expandable view problem?

moving-bits:
Current state is now:

image

More to come over the next days...

moving-bits:
Thanks to this nice method `Dialogs.bottomSheetDialogWithActionbar` converting the `AlertDialog` to a `BottomSheetDialog` was a matter of minutes:

image.image

ztNFny:
> Thanks to this nice method `Dialogs.bottomSheetDialogWithActionbar` converting the `AlertDialog` to a `BottomSheetDialog` was a matter of minutes:

That style could also be used in some other places, e.g. the "Cache Attributes" overview - or even at some future point replace the cache popup 😁

ztNFny:
Is the legend manually created for those 2 themes or will it work with any theme?

moving-bits:
> Is the legend manually created for those 2 themes or will it work with any theme?

Legend is theme-specific, as every theme is based on different markers set in the map files.
Currently it does work for Elevate/Elements theme (OAM), for Freizeitkarte base theme, and for Freizeitkarte outdoor themes.

On a generic base we could display the symbols provided in the theme files (POI) and display them in conjunction with their names. (But this would be an unsorted / ungrouped list then.)

PR #12420, opened 2021-12-27
add padding to compass rose in waypoint list (fix #9992)

### LINKED ISSUE #9992: Tapping compass rose on waypoint tab triggers scrolling instead

moving-bits:
## Description
- re-adds right padding to compass rose in waypoint list to avoid clashing with fastscroll bar
- additionally remove layout parameters not valid in a LinearLayout


ztNFny (changes requested):
margin wasn't lost but replaced by padding to increase the size of the clickable area, see https://github.com/cgeo/cgeo/issues/9992#issuecomment-1002107786

This shouldn't be merged

moving-bits:
> margin wasn't lost but replaced by padding to increase the size of the clickable area, see [#9992 (comment)](https://github.com/cgeo/cgeo/issues/9992#issuecomment-1002107786)
> 
> This shouldn't be merged

Fine for me as well (setting WIP label for now), but then the issue described in #9992 is in need for a different solution, as the current implementation seems to be incomplete. Maybe just the margin added to the icon needs to be increased another bit.

PR #12432, opened 2021-12-28
Fix #12431 - Increase detection of cache series to a max of 500 caches

### LINKED ISSUE #12431: Automatic sorting by name not working for larger series
### Describe your problem!

A user reported that for his trail/series of 81 caches the automatic sorting by name does not work.
I went to check the code and the detection is OK (the series of the user contain 4 common characters at beginning), but it seems we restrict this feature to a max of 50 caches:
https://github.com/cgeo/cgeo/blob/4028d8eb2faac13ec1e5b9dc5e9ddbf5b68aafd7/main/src/cgeo/geocaching/ui/CacheListAdapter.java#L668

I suggest to raise this to 100. 
Any objections or does anyone know why the limit is 50?

This looks so easy, that I can volunteer to adapt it, therefore assigning me.

### How to reproduce?

Store 80 caches with similar cache titles (e.g. Serie 1, Serie 2, etc.) 

Caches are sorted by default (distance)

Auto sorting (isSeriesList) should detect and sort by name.

### Reproducible

Yes

### c:geo Version

2021.12.24

### System information

_No response_

### Additional Information

Knowing that this is no longer so important as we have per list sorting preferences meanwhile, still I would like to fix it.

Lineflyer:
## Description

Increase the limit for detection whether a cache list contains a series of caches (`isSeriesList`) from 50 to 500 caches.

## Related issues

#12431

### DIFF
--- a/main/res/raw/changelog_base.md
+++ b/main/res/raw/changelog_base.md
@@ -11,6 +11,7 @@
 - New: Attributes overview (see Manage Caches => Attributes overview)
 - New: Add import from bookmark lists (GC premium only)
 - New: Invert sort-order on long click on sort bar
+- Change: Also perform automatic sorting by distance for lists containing cache series with more than 50 caches (up to 500)
 
 ### Cache details
 - New: Pass current cache coordinates to geochecker (if supported by geochecker)
--- a/main/src/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/cgeo/geocaching/ui/CacheListAdapter.java
@@ -665,7 +665,7 @@ public class CacheListAdapter extends ArrayAdapter<Geocache> implements SectionI
      */
     private void checkSeries() {
         boolean series = false;
-        if (list.size() < 3 || list.size() > 50) {
+        if (list.size() < 3 || list.size() > 500) {
             return;
         }
         final ArrayList<String> names = new ArrayList<>();


main/res/raw/changelog_base.md
```diff
@@ -11,6 +11,7 @@
 - New: Attributes overview (see Manage Caches => Attributes overview)
 - New: Add import from bookmark lists (GC premium only)
 - New: Invert sort-order on long click on sort bar
+- Change: Also perform automatic sorting by distance for lists containing cache series with more than 50 caches (up to 500)
```
moving-bits, line 14: FYI: Including the changelog message in the PR will lead to merge conflicts, if someone else creates a PR adding a change notice in the same line. Had that in the past frequently, and reverted procedure to adding the changelog messages later on.

Lineflyer: mmh...technically correct. But one could still remove it when it really happens to produce conflicts. IMHO its less likely than one might thing (same line, both PR opened).
Adding later one makes it more likely, that things are forgotten.

moving-bits: PR do not necessarily have to be open at the same time, it will also happen, when you haven't updated & rebased your local copy while working on your PR. It's sort of a "single point of failure", as changes in a PR are typically in different places, but the changelog will always be the same file. (We have worked like this a while ago, and it lead to quite a few merge problems back then.)

But anyway, for a single PR as this it won't hurt, and we can monitor how it will work if we follow this way more regularly.

PR #12437, opened 2021-12-29
replace cachelist attribute overview AlertDialog by BottomSheetDialog

moving-bits:
## Description
Replaces the cache list attribute overview `AlertDialog` by a `BottomSheetDialog`:

image

### DIFF
--- a/main/src/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/cgeo/geocaching/ui/CacheListAdapter.java
@@ -39,7 +39,6 @@ import android.widget.SectionIndexer;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
-import androidx.appcompat.app.AlertDialog;
 import androidx.core.content.ContextCompat;
 
 import java.lang.ref.WeakReference;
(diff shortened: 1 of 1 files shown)


moving-bits:
failing test is unrelated (triggered by outage of opencaching.us)

Lineflyer:
Just asking:
Is it the expected usage, that bottom sheets show up for something triggered from a menu?
Design wise I like it however :)

fm-sys:
> Just asking: Is it the expected usage, that bottom sheets show up for something triggered from a menu? Design wise I like it however :)

Not sure if there are any explicit recommendations for the usage of bottom sheets. However, I'm convinced that we should use them at way more places because they are simply awesome... Feels much more intuitive than an old and stale alert dialog! 😍

moving-bits:
> Just asking: Is it the expected usage, that bottom sheets show up for something triggered from a menu? Design wise I like it however :)

From https://material.io/components/sheets-bottom#modal-bottom-sheet I read that our usage is the intended one: _"Modal bottom sheets present a set of choices while blocking interaction with the rest of the screen. They are an alternative to inline menus and simple dialogs on mobile, providing additional room for content, iconography, and actions."_ So I guess we are fine here.

What we should think of is how to handle bottom sheet dialogs in landscape mode. One recommendation is: _"Don’t use full-width bottom sheets in large screen layouts. [...] A bottom sheet on mobile can also be swapped for a side sheet that shows supporting content on larger screens."_ But that's a separate issue.

Issue #11497, opened 2021-08-18
Scrolling triggers swipe to other tab


Lineflyer:
### Describe your problem!

Reported on support mail and video included with permission to publish:

https://files.droescher.eu/public/scroll_issue.mp4

Problem started in first beta and is still valid in production version 2021.08.15 for this user.
I cannot reproduce.

### How to reproduce?

See video

### Reproducible

Unclear

### c:geo Version

2021.08.15

### System information

```text
## System information

c:geo version: 2021.07.17-RC2

Device:
-------
- Device type: SM-G998B (p3sxeea, samsung)
- Available processors: 8
- Android version: 11
- Android build: RP1A.200720.012.G998BXXU3AUF6
- Screen resolution: 1080x2280px (360x760dp)
- Sailfish OS detected: false
- Google Play services: enabled - 21.24.18 (150400-383468479)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: orientation

Program settings:
-------
- Hide caches: -
- Hide waypoints: -
- Set language: de_DE (system default)
- System date format: dd.MM.yy
- Debug mode active: no
- Live map mode: true
- OSM multi-threading: true / threads: 4
- Global filter: display all caches
- Last backup: 13. Mai, 11:14
- Routing mode: Walk
- Settings: v5, Count:136
- Map: Google: Karte
- Id: cgeo.geocaching.maps.google.v2.GoogleMapProvider$GoogleMapSource
- Atts: none
- Theme: none

Services:
-------
- Geocaching sites enabled:
geocaching.com: Logged in (Anmeldung OK) / PREMIUM
Geocaching.com Adventure Lab
- Geocaching.com date format: dd/MM/yyyy
- Routing: external / BRouter installed: false
- Installed c:geo plugins: none

Permissions & paths:
-------
- Fine location permission: granted
- Write external storage permission: granted
- System internal c:geo dir: /data/user/0/cgeo.geocaching (166,0 GB free) v2 internal isDir(8 entries)
- Legacy User storage c:geo dir: /storage/emulated/0/cgeo (166,0 GB free) v2 external non-removable isDir(7 entries)
[...]

PR #12448, opened 2021-12-31
Fix #9992 - Reduce fast scroll bar active time

### LINKED ISSUE #9992: Tapping compass rose on waypoint tab triggers scrolling instead

Lineflyer:
## Description

- Reduce the time until fast scroll is disabled from 1000ms to 500ms to decrease the probability to interfere with nearby touch targets (see #9992).
- Could also help to improve #11497

## Related issues

#9992 
#11497

## Additional context

The alternative would be to disable the fastscroll for the waypoint tab, but this would mean there is no longer any possibility to quickly scroll through a long list (e.g. caches with hundreds of waypoints). Therefore I think this workaround is the way to go for now.

### DIFF
--- a/main/src/cgeo/geocaching/ui/FastScrollListener.java
+++ b/main/src/cgeo/geocaching/ui/FastScrollListener.java
@@ -11,7 +11,7 @@ import android.widget.ListView;
  */
 public class FastScrollListener implements AbsListView.OnScrollListener {
     private static final int MIN_COVERED_ENTRIES = 3;       // in number of entries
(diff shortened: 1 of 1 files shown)


moving-bits:

Issue #12378, opened 2021-12-23
[Global Calculated Coordinates/Waypoints] Implement export-from/import-to Personal Note


eddiemuc:
New global calculated coordinates/waypoints and variables will not yet be exported to Personal Note, and also not imported into waypoint list from it. This must be implemented.

eddiemuc:
Import/export of waypoints to/from PN needs to be changed for following reasons:
* Variables are now no longer part of a waypoint/calculated coordinate and must be imported/exported as a separate "entity"
* Current calculated coordinate format concentrates on waypoints with single-letter vars only, itmust be possible to export/import arbitrarily complex formula expressions with multi-char-variables

While I want to leave the ability to the user to also "free-flow" specifiy waypoints with calculated coordinates as well as variables in the PN (which are both then scanned automatically), the format must be a bit stricter for this specific use case - because it must be guaranteed that all formulas saved can also be parsed again (there's a separate issue for a calculated waypoint scanner in the listing, this scanner can be a bit more "heuristic")

Keeping all this in mind I suggest the following formats:

Calculated waypoints:
* Store coordinates in a format `{CC|<Latitude Formula>|<Longitude Formula}`
* Example: `{CC|N41°53.((A/(B/3))+D)/2|E12°28.(C*B)+(A*(B/2))+(A+(C*4)-1)}`
* Example as part of a waypoint: `@My Waypoint (W) {CC|N41°53.((A/(B/3))+D)/2|E12°28.(C*B)+(A*(B/2))+(A+(C*4)-1)} "this is a meaningful waypoint note"`
* Whether to use `CC` as the marker or something else can be discussed (e.g. `CALCULATEDCOORD` - but I would rather use something short so not too much PN space is taken and a user may be able to type this himself in reasonable time if he wants to specify sich coords manually)

Variables:
* Store variable in following format: `$<VarName>=<VarFormula>`
* Example: `$A=34*wordvalue('test')`
* To be discussed: if a cache has more than one variable (which is likely), they can either be stored in PN taking one line each or in on line separated by |. Example for separated by `|`: `$A=34*wordvalue('test') | $B=3 | $C=3+7*A-length(D)`

@murggel I realize that this makes much work obsolete which you did when implementing parsing of F-PLAIN formulas. But I fear the concept has changed to much that I can't take over much. Sorry for this.

For those worrying what happens with already stored coordinates (in the old format): there is a big task for migration set up (#12379), I plan to handle these things as part of this task.

[...]

Issue #12405, opened 2021-12-25
New WP calculator - Get rid of placeholder variables


Lineflyer:
While trying out the new waypoint calculator, which is combined with the cache variable tab, I had somehow a hard time to understand how its used.

@eddiemuc 
Can we talk through the following example here:

Typical multi cache:

- Stage 1: You will find the values for x and y at this stage.
- Stage 2: Go to N 51° 21.(x+1)y5 E006° 3x.650. You will find value z there.
- Final: Go to N 51° 21.z(y+2)x E006° 3x.(z+3)y3

With the old waypoint calculator I need to fill the calculator for stage 2 and final as follows:

- Waypoint for Stage 2:  N51° 21.AB5 E006° 3C.650  with A=x+1, B=y, C=x
- Waypoint for Final: N51° 21.ABC E006° 3D.EF3 with A=z, B=y+2, C=x, D=x, E=z+3, F=y

The drawback of these non global variables (lower case ones) are now, that I have to enter the values for x,y,z on each waypoint (where its required).

I would have expected, that the new approach stays the same with (in principle) the only difference, that the values of those lower case variables are now retrieved from the result/value in the variable tab. 
But either I don`t understand how to achieve this (which would put into question if an average user will understand the changed working mode) or its implemented in a different way (in which case I would ask why).

Lineflyer:
I made a user defined cache containing the example above to visualize the current mode of operation I use:

Stage 2 from above example:
![Screenshot_20211225-180002_cgeo](https://user-images.githubusercontent.com/949669/147389897-5c414575-21ad-446d-b0e9-5c971556f35a.jpg)

Final from above example:
![Screenshot_20211225-180111_cgeo](https://user-images.githubusercontent.com/949669/147389908-e9a14302-b45b-4601-84e3-b39313527707.jpg)

IMHO the "only" thing needed is to "remove" the right input side (lower case variables) and retrieve them from the variable tab instead, or not?

When trying the same in the new WP calculator I get much more than needed on the variable tab:
![Screenshot_20211225-180535_cgeo](https://user-images.githubusercontent.com/949669/147389987-d4870ee1-eeb4-474a-a390-626a69652149.jpg)
![Screenshot_20211225-180557_cgeo](https://user-images.githubusercontent.com/949669/147389989-b661a0a8-b087-4e95-a509-6fe0f520b3cb.jpg)

Lineflyer:
[...]

PR #12481, opened 2022-01-06
rel to #12378: change calculated config to something suitable for PN

eddiemuc:
rel to #12378: change calculated config to something suitable for PN


eddiemuc:
I need the `TextParser` contained in this PR for #12405 and will thus merge this

PR #12486, opened 2022-01-06
rel to #12405: first try to establish new guided calc coordinate input

eddiemuc:
rel to #12405: first try to establish new guided calc coordinate input


eddiemuc:
This PR still leaves some minor GUI issues in code, but I find it more important to get feedback on the usability than already polishing the GUI up

PR #12487, opened 2022-01-06
use the new WhereYouGo icon

ztNFny:
WhereYouGo will soon use the new icon (currently in beta), update the "launch" and "useful apps" icons to match it


moving-bits:
New WhereYouGo version has been published, so let's merge here now

Issue #12129, opened 2021-11-18
Allow multiple track files


moving-bits:
### Tell us your idea!

From support:

Extend c:geo lists by allowing to load a list-specific GPX track file, which gets displayed automatically on displaying that list. (Similar to list-specific icon.)

Use-case is planning a tour, creating a list for it and putting all related caches into the list. Then associate a track file to be displayed whenever this list is shown.
Extended use-case is planning a longer tour with multiple lists and a separate track file for each list.

### What solution would you suggest?

- extend list menu by "manage list" - "Set list track" to load a list-specific GPX track file
- extend list menu by "manage list" - "Remove track" to unload list-specific track file (displayed only if a list-specific track file is associated with this list)
- on pressing the "display this list on map" button the list-specific GPX track file gets set as currently selected GPX track file (replacing any currently set track file)
- optional: auto-activate track display on "display this list on map", if a track file is associated with this list

### Additional Information

ticket #154783

Lineflyer:
IMHO thats way to specific to couple it with lists.
I could more imagine a general pool of available tracks (so not loading/unloading a single file but adding several files to be shown/hidden on demand.

moving-bits:
If we want to pursue this way, I could image the map's "GPX-Track/Route" menu entry opening sort of a "track manager window" with the following options:

|Trackname|checkbox|icon|icon|
|---|---|---|---|
|Track 1|checkbox|icon|icon|
|Track 2|checkbox|icon|icon|
|Track 3|checkbox|icon|icon|

- The checkbox would be used for switching display of the specific track (on/off),
- the first icon would be a "center on this track" icon,
- the second icon would be used to unload the specific track.
- Either at the bottom or as part of the dialog's title there would be an "load track/route" icon.

Number of possible tracks would be "unlimited".

moving-bits:
@eddiemuc 
Currently SAF supports a single track file. Can you give me a pointer what would have to be done to support multiple track files at the same time?

eddiemuc:
> @eddiemuc 
> Currently SAF supports a single track file. Can you give me a pointer what would have to be done to support multiple track files at the same time?

Do you want to support a fixed set of files (each handpicked to be a track file by the user, c:geo does not know or need to know it's storage location) or rather a directory where all contained files are considered track files?

I think track file is the only case in cgeo today where we work with single files, all other use cases are "folder-based"

eddiemuc:
[...]

PR #12499, opened 2022-01-07
Unify tracks, routes, indiv. routes options dialog (rel. to #12129

moving-bits:
First step towards #12129: Unify handling of individual routes and tracks/routes from GPX into a `BottomSheetDialog`:

image.image

- new menu entry for maps: "Routes / Tracks" - opens the new unified dialog (same as the already existing button "tracks/routes" on the map's bottom does)
- menu entries "GPX-Track/Route" (with submenu), "Load individual route" and "Clear manual targets" are removed, as their functionality has been moved to the new dialog
- current handling is for the existing single track file with persisted URI - no adaption to multiple track files yet


fm-sys:
IMHO "set route start as target" should be removed from the "quick settings" as it is now contained in the new "routes" popup.

We could also think about moving the "routing mode" into the new popup as well. Would at least be more fitting... What do you think?

moving-bits:
> IMHO "set route start as target" should be removed from the "quick settings" as it is now contained in the new "routes" popup.

hm, need to think about it. Maybe it does not belong to the new popup, as it's not about loading/unloading/visibility of routes/tracks, but some navigation option?

> We could also think about moving the "routing mode" into the new popup as well. Would at least be more fitting... What do you think?

This should stay, IMHO, as it is relevant not only for individual routes, but for manually set targets as well.

Or we move all routing settings to the new popup, extending it from "routes/tracks" to "routing & tracks" (or something similar).
But then the quick access button should be visible all the time and not only when an individual route has been loaded.
=> Maybe a topic for next dev call?

ztNFny:
> This should stay, IMHO, as it is relevant not only for individual routes, but for manually set targets as well.

I don't use routes but I use BBrouting. Imho it's completely unrelated and such stay as is.

Issue #12384, opened 2021-12-23
[Global Calculated Waypoints] Allow multi-creation of waypoints based on calculated waypoint


eddiemuc:
https://github.com/siggel/coordinatejoker offers a nice possibility to test multiple input values and see the resulting waypoints. Would be nice to have this.

_Originally posted by @ztNFny in https://github.com/cgeo/cgeo/issues/12316#issuecomment-992511234_

eddiemuc:
I am thinking to allow for any existing calculated coordinate to specify a loop. On context menu of calculated waypoints add an option "Create multi waypoints" which opens a dialog allowing to specify:
* A variable used in the calculated coordinate of the waypoint (e.g. "A")
* A start value (e.g. 0), a stop value (e.g. 9) and a step value (e.g. 1)

Dislog would allow up to 2 of these "loops" to be specified.

On ok (and maybe after preview and confirmation esp when many waypoints would be created), c:geo would create a (normal) waypoint for each of the loop values. In the above example, 10 waypoints would be calculated and created usingthe values 0, 1, 2, ..., 9 for variable A.

Lineflyer:
Do you mean something llike suggested in #2137 ?
Would highly appreciate this, as its a quite common usecase (for me), that I have a missing solution due to an unavailable waypoint while playing a cache. This would allow getting an easy overview to find the potential correct solution e.g. looking at the map.

Lineflyer:
Taking over an idea given in #2137

> I have just seen a quite smart implementation of this in an android app:
You can actually enter WPs like that: N 51 20.[169]40 E 007 00.3[1479]30

ztNFny:
> Taking over an idea given in #2137
> 
> > I have just seen a quite smart implementation of this in an android app:
> > You can actually enter WPs like that: N 51 20.[169]40 E 007 00.3[1479]30

I like the approach with variables better.

 N 51 20.[169]40 E 007 00.[169]30
vs
 N 51 20.x40 E 007 00.x30 with x=169

One is 3 options, the other 3*3

ztNFny:
But maybe the global variables page could accept values like [0-9], [1,2,3] for a specific variable?

eddiemuc:
> Taking over an idea given in #2137
> 
> > I have just seen a quite smart implementation of this in an android app:
> > You can actually enter WPs like that: N 51 20.[169]40 E 007 00.3[1479]30

[...]

PR #12505, opened 2022-01-08
rel to #12384: first version for multi waypoint creation

eddiemuc:
rel to #12384: first version for multi waypoint creation

### DIFF
--- a/main/res/menu/waypoint_options.xml
+++ b/main/res/menu/waypoint_options.xml
@@ -47,5 +47,10 @@
         android:id="@+id/menu_waypoint_caches_around"
         android:title="@string/cache_menu_around">
     </item>
+    <item
+        android:id="@+id/menu_waypoint_generate"
+        android:title="Generate Waypoints"
+        android:visible="false">
+    </item>
 
 </menu>
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -48,6 +48,7 @@ import cgeo.geocaching.location.Units;
 import cgeo.geocaching.log.CacheLogsViewCreator;
 import cgeo.geocaching.log.LogCacheActivity;
 import cgeo.geocaching.log.LoggingUI;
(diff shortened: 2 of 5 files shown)


eddiemuc:
A first and not yet very comfortable possibiility to create multiple waypoints out of a calculated coordinate.

* Select a calculated coordinate with long click and choose new entry "Generate Waypoints":
image

* In the upcoming dialog, enter a range expression for the variable you want to iterate over, e.g.:
image

* In the upcoming list, select the generated coordinates which actually should become waypoints, e.g.:

image

=> waypoints are generated as selected:

image

image

eddiemuc:
When investing a little time you can create ratheer artistic waypoints constellations already:

image

image

image

image

moving-bits:
Very cool feature, love to see that in c:geo!

A couple of questions / ideas:
- is a syntax like "A=1,3,6" possible as well?
- are the generated waypoints preselected (dialog) - otherwise an "all" button would be nice
- shall we use a separate waypoint type for those generated waypoints to be able to use a different icon (for easier recognition on the map)
- the latter would also allow a function to remove all generated waypoints from the waypoint list which could come handy

eddiemuc:
I just relalized that the feature incomplete. It will not correctly work for indirect values. E.g. if there's a variable B=A+1 and I specif A=0-5 then B will not be updated accordingly. I probably should not merge this feature yet...

eddiemuc:
> - is a syntax like "A=1,3,6" possible as well?

Works already. You can also exclude single values by pretending ^

> - are the generated waypoints preselected (dialog) - otherwise an "all" button would be nice

Yes that would be handy :-) In plan to add this generically for the SimpleDialog.selectMulti function. But it is not there yet.

> - shall we use a separate waypoint type for those generated waypoints to be able to use a different icon (for easier recognition on the map)

Good idea

> - the latter would also allow a function to remove all generated waypoints from the waypoint list which could come handy

Also good idea. To prevent generating too many waypoints I was also thinking about building a map preview function to view the waypoints before even saving them.

eddiemuc:
This PR is superseeded by PR #12588, thus I will close/delete it

PR #12506, opened 2022-01-08
add visibility options to new route/track dialog

moving-bits:
- adds option to set/unset visibility of tracks and routes (new for individual routes)
- visibility values are not persisted yet
- also activates the "routes/tracks" quick access button when only a track is loaded

image


moving-bits:
Setting WIP label for now, will integrate this PR into #12507.

moving-bits:
Superseded by / included in #12507

PR #12520, opened 2022-01-14
SimpleDialog multiselect: add selectall option

eddiemuc:
This PR adds a "selectall" option to multiselect dialogs opened using SimpleDialog. For example:

image

### DIFF
--- a/main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
+++ b/main/src/cgeo/geocaching/ui/dialog/SimpleDialog.java
@@ -4,6 +4,7 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.activity.Keyboard;
 import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.functions.Action2;
 import cgeo.geocaching.utils.functions.Func1;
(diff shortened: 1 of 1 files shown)


Eldenroot:
I think it should be separated somehow... maybe a line after the "select all"?

fm-sys:
Not necessary IMHO. The purposed solution looks perfectly fine at least to me...

Lineflyer:
I also think the "select all" entry could look a little different (e.g. indented, cursive,..) to differentiate from the rest.

eddiemuc:
I propose to put the entry between `<>`. E.g. `<Select All (21)>`. This would be the same design like the "new list" entry is separated from existing list entries in the "save caches" dialogs.

Additionally we can make font type italic if this is wanted.

Putting a line under the entry would be very complicated technically (due to restrictions in AlertDialog which is used underneath) so I would not do this.

moving-bits:
Putting it into brackets sounds like a good idea, I would stay away from italic or other font attributes, though, to keep dialogs consistent.

eddiemuc:
Found a bug in impementation, do not merge. It is surprisingly complicated to create a working "select all checkboxes" programatically in an AlertDialog. It is actually programmed kind of stupidly

eddiemuc:
For comparison: this is what it looks like with <> and italic:

image

eddiemuc:
...and this is what it looks like with just <> but no italic:

image

I will commit the version without italic for now but leave this PR open in case there's more discussion

eddiemuc:
Bug is fixed too

Lineflyer (approved):
IMHO without italic and using <> is inline with other places, like for selecting lists to store a cache in, where we use `<New list>` as a selection item.

eddiemuc:
@Lineflyer is it ok to merge this one to master? (due to upcoming next release)

Lineflyer:
Yep, `master` please.

PR #12521, opened 2022-01-15
correct maps.me download

bekuno:
The old address does not more exist.
Instead there is a link on the home page to the map store.


bekuno:
see also https://github.com/mapsme/api-android/pull/35

Issue #12451, opened 2022-01-01
Next beta/feature release


Lineflyer:
It should be the time now to think about bringing current `master` (including bottom navigation) into a beta state in preparation of a new feature release (unless there is some critical regression or website change).

This also means we should stop adding new (untested) features at some point to be able update `release` to the head of `master` and have a more-or-less stable version. 

As the functions around waypoint calculation and variable tab are still developed in parallel to existing functions and (by default) hidden from end users its uncritical to continue work on this. Or shall this be completed first to be included unhidden in the next beta?

I suggest to check the open feature-PR whether to be included before or after branching and provide feedback here.
Same goes for issues: Do we have to-be-fixed regressions or problems in new features needing to be fixed before branching or before releasing a beta?

My first quick look (feel free to edit and append):
- [x] #12448 <-- Should be merged to mitigate #9992
- [x] #12441 <-- Include into next beta?
- [x] #12443 <-- Should be decided upon and fixed before beta
- [x] #12423 <-- Should be decided upon and fixed before beta
- [x] #12425 <-- Regression, which needs to be fixed before beta
- [x] #12409 <-- Deadline 04/2022
- [x] #12457 <-- Regression, which needs to be fixed before beta
- [x] #12558 
- [x] #12559
- [x] #12557
- [x] #12613
- [x] #11969 (remaining tasks)
- [x] #12654
- [x] #12667 Missing log count (which also causes missing fav percentage)
- [x] #12673 Log upload with pictures
- [x] #12659 Upcoming field note page change
- [x] #12452
- [x] #12689
- [x] #12688
- [x] #12701 (for release add temporary mitigation at least: add "clear manual targets" to menu)
- [x] #12704
- [x] #12710 <-- Regression, which needs to be fixed before release
- [x] #12573

eddiemuc:
Wrt waypoint calculation: I wouldn't recommend waiting for completion before next beta. There's still much work to do befor this can be presented to (beta) users and features might still break. Whole PN export/import and migration is still open, and I am not satisfied with solution ideas for the "guides input" yet at all.

eddiemuc:
Hm oraBe active in debug mode?

ztNFny:
I'd add https://github.com/orgs/cgeo/teams/team/discussions/52 to that list

moving-bits:
Added #12462 to fix #12457.

What about #12418? Should be non-critical as it adds a single menu entry (for OSM offline maps only), displaying an info dialog.

[...]

PR #12525, opened 2022-01-15
add setting for nearby search distance limit (fix #11489)

### LINKED ISSUE #11489: Make NearbySearch cache number configurable or introduce "load more"
> By the way. How big is the "box" (distance, number of caches) that is used for the search and is it configurable?

That is a relevant question, and recently asked by users.
Somehow users got used to the pagination (20 packs) as a convenient feature to just store those to have the closest, while now it requires some manual selection or filtering to achieve this (or store all 200 caches).
Maybe we could use a clever boxing or alike, will transfer in a new issue.

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/11488#issuecomment-900557792_

moving-bits:
## Description
Adds a setting to Settings => Services => Nearby search to define a limit up to which the services return caches on nearby search. A value of 0 means "unlimited" and is the default.
(Limit will get applied for nearby search only, neither for coord search or address search.)

### DIFF
--- a/main/src/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
+++ b/main/src/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
@@ -4,6 +4,7 @@ import cgeo.geocaching.filters.core.DistanceGeocacheFilter;
 import cgeo.geocaching.filters.core.GeocacheFilterType;
 import cgeo.geocaching.filters.core.IGeocacheFilter;
 import cgeo.geocaching.location.Geopoint;
+import cgeo.geocaching.settings.Settings;
 
 import android.app.Activity;
 
@@ -11,10 +12,12 @@ import androidx.annotation.NonNull;
 
 public class CoordsGeocacheListLoader extends LiveFilterGeocacheListLoader {
     @NonNull public final Geopoint coords;
+    private final boolean applyNearbySearchLimit;
 
-    public CoordsGeocacheListLoader(final Activity activity, @NonNull final Geopoint coords) {
+    public CoordsGeocacheListLoader(final Activity activity, @NonNull final Geopoint coords, final boolean applyNearbySearchLimit) {
         super(activity);
         this.coords = coords;
+        this.applyNearbySearchLimit = applyNearbySearchLimit;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


main/src/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
```diff
@@ -25,9 +26,12 @@ public GeocacheFilterType getFilterType() {
     @Override
     public IGeocacheFilter getAdditionalFilterParameter() {
         final DistanceGeocacheFilter distanceFilter = (DistanceGeocacheFilter) GeocacheFilterType.DISTANCE.create();
+        final int nearbySearchLimit = Settings.getNearbySearchLimit();
```
eddiemuc, line 29: The CoordsGeocacheListLoader is used for Nerby-Search, Coord-Search and Address-Search. My understanding was that the distance limitation should only apply for Nearby-Search. Thus this must be differentiated here.

fm-sys:
To make this feature useful, we should probably add an entry to the overflow menu as shortcut (only for "nearby"):

(There's much space left anyway, currently. Probably the only menu in cgeo, which has that less content ;-))

Screenshot_20220115_171307 (¼ Size)

main/src/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
Lineflyer: Lets overthink this a bit.
At least the problem of "overcoverage" in less cache densed areas also applies to coord/address-search...mmmh.

eddiemuc: > Lets overthink this a bit.
> At least the problem of "overcoverage" in less cache densed areas also applies to coord/address-search...mmmh.

I always thought that this is specific to "nearby" searches - since it is done in the field to find out what can immediately be reached. The other searches (by coordinate, address or keyword/name) seem more like something done during the planning phase of e.g. a cache tour. Here I would expect no distance limited results.

Also what then about the other searches (by keyword, by owner, found by)?

In any case, the current issue was specific to nearby search, so if this is really a discussion topic then a new issue should be created for it

moving-bits: I agree to those arguments

additionally see https://github.com/cgeo/cgeo/issues/12451#issuecomment-1013762860

bekuno: > Lets overthink this a bit. At least the problem of "overcoverage" in less cache densed areas also applies to coord/address-search...mmmh.

Yes, there is the same problem.
If I want to search for a address (or coord), then I want to see caches around this place.
As sample: search for "Spiekeroog" for a holiday trip. The map of the result is:

This is nearly unuseable, because I can not find the correct place. And a do not need caches in Hamburg.
A search by distance would be better here, then the destination will be placed nearly in the middle of the map.
A reduced amount of caches is fine, as loang as it is easy to reload more caches. (e.g. by switching to "live mode").

PR #12534, opened 2022-01-16
fix lint: unused drawables and values for cache ratings

moving-bits:
## Description
Fixes the 33 lint messages regarding unused resources/values related to cache ratings by creating a dummy enum referencing the drawables (which in itself use the color values).

Additionally fixes an older CheckStyle error about wrong import order.


Lineflyer:
Wouldn't it be better to just ignore the lint message by adding according rules to `lint.xml` ?

moving-bits:
> Wouldn't it be better to just ignore the lint message by adding according rules to `lint.xml` ?

Yes, already working on it, see #12541

PR #12540, opened 2022-01-16
Improve strings for caches with chirp attribute

Lineflyer:
## Description

A cache owner contacted us on support mail:
He owns and knows many caches, which use the "chirp attribute" but do not contain a chirp but some other transmitting related stuff like Wifi or NFC. As can be seen in many examples the attribute is also used for such kind of caches.

Users looking for such caches do contact him as they are confused by c:geo reporting, that CacheWolf is needed as it is a chirp cache, while it is not.

I suggest to give a short hint in the text used to link to CacheWolf, that it **might** be a chirp cache.

## Related issues

None

## Additional context

Support ticket 463619

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1423,8 +1423,8 @@
 
     <string name="cache_whereyougo_start">Open WhereYouGo to download and start this cartridge</string>
     <string name="cache_whereyougo_install">Install WhereYouGo player to download and start this cartridge</string>
-    <string name="cache_chirpwolf_start">Open Chirp Wolf to read the chirp data for this cache</string>
-    <string name="cache_chirpwolf_install">Install Chirp Wolf to be able to read chirp data</string>
+    <string name="cache_chirpwolf_start">This cache might use either a chirp transmitter or another kind of wireless beacon. For reading chirp data you can open Chirp Wolf.</string>
+    <string name="cache_chirpwolf_install">This cache might use either a chirp transmitter or another kind of wireless beacon. For reading chirp data you can install Chirp Wolf.</string>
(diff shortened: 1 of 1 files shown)


main/res/values/strings.xml
```diff
@@ -1423,8 +1423,8 @@
 
     <string name="cache_whereyougo_start">Open WhereYouGo to download and start this cartridge</string>
     <string name="cache_whereyougo_install">Install WhereYouGo player to download and start this cartridge</string>
-    <string name="cache_chirpwolf_start">Open Chirp Wolf to read the chirp data for this cache</string>
-    <string name="cache_chirpwolf_install">Install Chirp Wolf to be able to read chirp data</string>
+    <string name="cache_chirpwolf_start">This cache might use a chirp transmitter. Open Chirp Wolf to read the chirp data for this cache</string>
```
moving-bits, line 1426: New string is ok for me, IMHO.

If we want to stress possible alternatives a bit more, we could use something like: 

"_This cache might use a chirp transmitter or other wireless beacons._ Open Chirp Wolf to read the chirp data for this cache _(if available)_." and "_This cache might use a chirp transmitter or other wireless beacons._ Install Chirp Wolf to be able to read chirp data."

(Highlighting changes for visibility only, not to be included in the actual program.)

Lineflyer: I first was afraid to make it to long. 
But as this is only shown for those specific caches, I guess its fine. Will update the PR.

Lineflyer:
Lets merge to finalize conditions for merging branches.

moving-bits (commented):
LGTM

Issue #12533, opened 2022-01-15
Cache rating: lint claims unused resources


moving-bits:
### Describe your problem!

Recent builds claim a couple of lint messages about unused resources related to the cache rating feature.
There are 22 messages regarding drawables (`marker_rating_d_0.xml` etc.), and another 11 regarding colors (`marker_rating_0` etc.).

@ztNFny 
Can you cross-check whether any of those resources are actually used, remove unused resources and apply hints for lint to avoid any false positives?

### How to reproduce?

see lint report of any recent builds

### Actual result after these steps?

33 lint errors regarding "unused resources" related to the cache rating feature

### Expected result after these steps?

no false lint positives regarding cache rating feature

### Reproducible

Yes

### c:geo Version

current master

ztNFny:
All of them are used.
Dunno how to tell lint to ignore them.

moving-bits:
I don't have a solution at hand either, currently, but did some analyzing:

The drawables are loaded by concatenating their names dynamically, therefore neither compiler nor lint can automatically detect them being in use.
Colors are referenced by the related drawables only, thus because they are not detected as being in use, the colors are neither.

A mitigation for those lint messages would either have to include those 33 messages in our ignore list (not a nice solution), or to explain lint somehow that those resources are in use actually.

ztNFny:
does it make sense to add code just to make a "stupid" program happy?

moving-bits:
> does it make sense to add code just to make a "stupid" program happy?

Yes, IMHO, if it saves me time on checking lint status log repeatedly for the same messages. (Our current lint log is still too long.) The added code does not harm, and its purpose is documented. We do have such things in other places.

eddiemuc:
> does it make sense to add code just to make a "stupid" program happy?

IMHO it doesn't.

I checked into our existing `lint.xml` file, we already have a few ignore-entries for the UnusedResource warning:

![image](https://user-images.githubusercontent.com/6909759/149659851-d811508f-7e52-4ec2-91bd-8e5c42ceacd9.png)

I think the right way to go is to add these resources there as well. AFAI read on internet, regex may be used like this:

`<ignore regexp="ga_trackingId|google_crash_reporting_api_key" />`

moving-bits:
Ok, if there is a better declarative way, then I would prefer to go that way. Thanks for reminding me of that way. I'll dive into it.

PR #12541, opened 2022-01-16
adapt lint config for false positives on cache rating markers

moving-bits:
related to #12533

## Description
instead of adding code to reference the cache rating markers config lint to ignore those seemingly unused resources


moving-bits:
FYI: With these (and the other lint related fixes today) we are now down to 89 lint messages from more than 150 before: https://ci.cgeo.org/job/cgeo-CI_PR-build/androidLint/trendDetails/

PR #12550, opened 2022-01-17
fix #12511: implement calculated coordinate parsing rules according to issue

### LINKED ISSUE #12511: Calculated Coordinates: coordinates are not parsed as expected by Geocachers
Current Coordinate parser is pretty simple: it parses coordinates from left to right, each formula it finds is considered to result in degree / minute / second decimal value depending on position and follow-up charater (°, ' or ").

This results in some behavior unexpected to geocachers. This is especially true for the NDD MM.MMM format. For example: 

* `N50°00.A+B+C'` with `A=2, B=3, C = 5` results in `N50° 8.2'` (where 8.2 = 00.2 + 3 + 5) instead of `N50° 00.010'` (which would be expected by geocachers).
* Something like `N214°` parses correctly even though the degree value is totally out of range

On dev meeting on 9.1.22 we discussed the following measurements:

**1) Parse parts before and after decimal point as separate formulas**

Evaluate both parts to integers (round up if is is a float value, throw an error if this is not possible) then concatenate them

Doing this `N50°00.A+B+C'` with `A=2, B=3, C = 5` will then result in `N50°00.10'`

**2) add leading zeros to minute decimal part**

If a coordinate part is identified to be a minute value and the part after decimal point is less than 3 digits long then it is filled up with leading 0 filled up 

Doing this (combined with the first measurement) `N50°00.A+B+C'` with `A=2, B=3, C = 5` will then result in `N50°00.010'` (note the leading 0 before 10)

**3) Evaluate coordinate for correct value ranges**. 

For example latitude shall not be outside +/-90 degree.

This issue is used to track implementation of the above. There were other measurements discussed at the dev meeting. For discussion of those I open another issue #12512

eddiemuc:
fix #12511: implement calculated coordinate parsing rules according to issue

### DIFF
--- a/main/src/cgeo/geocaching/models/CalculatedCoordinate.java
+++ b/main/src/cgeo/geocaching/models/CalculatedCoordinate.java
@@ -5,10 +5,10 @@ import cgeo.geocaching.utils.TextParser;
 import cgeo.geocaching.utils.formulas.DegreeFormula;
 import cgeo.geocaching.utils.formulas.Value;
 import cgeo.geocaching.utils.functions.Func1;
+import static cgeo.geocaching.models.CalculatedCoordinateType.PLAIN;
 
 import android.os.Parcel;
 import android.os.Parcelable;
-import android.util.Pair;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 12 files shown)


eddiemuc:
Some example views:

image

image

Note however that this PR does NOT yet deal with the things written in issue #12579 (see latitude in following example):
image

eddiemuc:
Sorry, wrong example from #12579, here is the original one:

image

ztNFny:
> Sorry, wrong example from #12579, here is the original one:
> 
> image

As stated in #12572 IMHO that parsing is wrong, should be DMM instead of DMS

eddiemuc:
> As stated in #12572 IMHO that parsing is wrong, should be DMM instead of DMS

Understood. But I would really like to discuss and resolve this in a separate PR. This PR already is quite large and complex. Hope that's ok?

fm-sys:
> Hope that's ok?

IMHO yes. The DMM/DMS thing is totally off-topic when speaking about this PR, and we should always try to keep as small as possible...

I would vote for getting this PR into `master` rather soon, so that we can collect nightly user feedback.

Issue #12409, opened 2021-12-25
Fix security issue 


Lineflyer:
Details available (restricted) on https://github.com/orgs/cgeo/teams/team/discussions/52

I still wanted to open an issue here as there is a deadline set to fix it. 
Otherwise the app will be removed from Google Play.

Please use the linked thread for discussion, but this issue can be used to track the fix.

Lineflyer:
I uploaded a nightly version including the code from https://github.com/cgeo/cgeo/pull/12551 to Google Play.
The pre launch report does no longer show the issue (but only the other known one).

So we could potentially include it in next beta to 
1. get another test run
2. have it tested in beta

@eddiemuc 
Shall I cherry-pick it to release?

Lineflyer:
Lets close here.
Check of the beta version on Google Play did also no longer show this issue.

PR #12551, opened 2022-01-17
rel to #12409: strip permissions from searchactivity intent before giving it back

eddiemuc:
rel to #12409: strip permissions from searchactivity intent before giving it back

### DIFF
--- a/main/src/cgeo/geocaching/SearchActivity.java
+++ b/main/src/cgeo/geocaching/SearchActivity.java
@@ -30,6 +30,7 @@ import android.app.Activity;
 import android.app.SearchManager;
 import android.content.Intent;
 import android.content.res.Configuration;
+import android.os.Build;
 import android.os.Bundle;
 import android.text.Editable;
 import android.text.InputFilter;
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/SearchActivity.java
```diff
@@ -123,6 +133,17 @@ public final void onCreate(final Bundle savedInstanceState) {
         init();
     }
 
+    private static void removeFlags(final Intent intent, final int flags) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
Lineflyer, line 137: Isn't this IF statement doubled with the calling part above?

eddiemuc: Forgot to delete unused method, will remove it (codacy complains too...)

eddiemuc:
BTW: if this PR does NOT remove the warning, then I propose simply to not send the intent data back to the caller in the assumed problematic line

Lineflyer:
FYA:
I just merged branches!
If this shall be part of next beta you should please re-target it to `release` now!

eddiemuc:
> FYA: I just merged branches! If this shall be part of next beta you should please re-target it to `release` now!

done

Lineflyer:
Can someone provide me some details about how to potentially test if this has any side effects?
When is this code line / function (Searchactivity intent) used? What do I have to do?

eddiemuc:
> Can someone provide me some details about how to potentially test if this has any side effects? When is this code line / function (Searchactivity intent) used? What do I have to do?

Not exactly. Somehow you have to start a search from another app which is handled by c:geo but fails. 

In this special case it might be wise to first check whether the fix really solves the reported security issue and THEN test it afterwards - since the former seems to  be easier to achieve, and if the fix does not help with the security report then we have to abandon it anyway.

Lineflyer:
> Not exactly. Somehow you have to start a search from another app which is handled by c:geo but fails.

Do you mean "search" like e.g. a nearby search?

Could the problem itself be caused by the relatively new feature to trigger a nearby search from shared coordinates from another app?
E.g. Open Adventure Lab app, select a stage, use navigation icon --> Select "Search Caches Nearby with c:geo" as target

eddiemuc:
> Can someone provide me some details about how to potentially test if this has any side effects? When is this code line / function (Searchactivity intent) used? What do I have to do?



> > Not exactly. Somehow you have to start a search from another app which is handled by c:geo but fails.
> 
> Do you mean "search" like e.g. a nearby search?

No. As far as I can see from code, this search tries to find and extract a geocode from the URL passed with the intent, and then it searches for a cache with this geocode.

(There's a side path in code which also is doing a keyword search, but intent has to have rather c:geo-specific fields for this path to be chosen)

I have to admit that I don't know this code part very well.

As said before: I recommend first to try whether this code version will fix the security warning of Google before going into further testing.

Lineflyer:
> As said before: I recommend first to try whether this code version will fix the security warning of Google before going into further testing.

Then probably its better to merge to `master` only for now. Afterwards I can push nightlies to Googles test engine again to see if something changes.

OK?

eddiemuc:
Ok, will merge to master

PR #12558, opened 2022-01-18
Align MultiTrackFile GUI to the rest of cgeo

### LINKED ISSUE #12552: Improve toggle button UX
If discovered some inconsistencies with the new "multi trackfile"-popup, which should be fixed before/during our beta:

1. Show current vs. following  

In our map activity, we usually show the current state instead of the following action (e.g. "follow my position", "enable/disable sync")

Screenshot_20220117_071508_cgeo geocaching_copy_285x136

This should be done for "show/hide track" in the same way. 

Screenshot_20220117_071620_cgeo geocaching_copy_268x142

2. Tooltips 

Furthermore we already have a tooltip for "show/hide", but if we implement an icon-tooltip it should be done for all icon buttons in that dialog. 

3. Visual feedback 

And, last but not least (maybe the most important design guideline):
**Always give feedback to the user.** 

Especially as the buttons don't have a border around the icon, the button should have a ripple effect. We already do the same thing at other places... 

Screenshot_20220117_071820_com android gallery3d_copy_431x248

fm-sys:
## Description
- give visual feedback and show tooltip
- switch icon meaning for visibility button

## Related issues
(fix #12552)

### DIFF
--- a/main/res/layout/routes_tracks_item.xml
+++ b/main/res/layout/routes_tracks_item.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:paddingLeft="6dp"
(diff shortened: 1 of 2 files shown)


Lineflyer (approved):
LGTM from the ideas described in the corresponding issue.

moving-bits:
Please let decide on the issue described in https://github.com/cgeo/cgeo/issues/12552#issuecomment-1015864605 first before merging.

fm-sys:
> Please let decide on the issue described in [#12552 (comment)](https://github.com/cgeo/cgeo/issues/12552#issuecomment-1015864605) first before merging.

This is the only outstanding PR before we can release the beta. Can we simply merge it in the current form to have a consistent solution + some GUI improvements for the first beta?

Further optimization is possible for sure, but changing the "follow location" / "switch live" button design and all that is way too far away for the original issue. The "follow location" button isn't new so I don't see a reason why we should change the appearance exactly now?!

Lineflyer:
> This is the only outstanding PR before we can release the beta. Can we simply merge it in the current form to have a consistent solution + some GUI improvements for the first beta?

This proposal is OK for me. It allows us to push the beta before the weekend.
So I take the freedom to merge this but reopen the corresponding issue afterwards for final discussion.

PR #12559, opened 2022-01-18
[MultiTrackFiles] use better "center" icon

fm-sys:
IMHO, the meaning of the old center icon was not clear enough, so I suggest using a different one... 

|before|now|
-|-
grafik|grafik

### DIFF
--- a/main/res/drawable/center.xml
+++ b/main/res/drawable/center.xml
@@ -1,4 +1,4 @@
-<!-- material.op / adjust -->
+<!-- take from materialdesignicons.com / image-filter-center-focus -->
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
     android:width="24dp"
     android:height="24dp"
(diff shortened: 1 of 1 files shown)


Lineflyer (commented):
I visually like the proposed one more than the current one.
But how about reusing the center icon we have in the map action bar. That should make it very clear, as users know it.

fm-sys:
Added the label "Feadback required" for now...

moving-bits:
> I visually like the proposed one more than the current one. 

Ok for me as well

> But how about reusing the center icon we have in the map action bar. That should make it very clear, as users know it.

You mean the "follow my location" icon? I wouldn't use that, as it does something different: The "follow me" icon not only centers the map, but follows the user's GPS location, whereas the "center on track/route" just focusses the map on the center of the track/route, for a single time.
(Which reminds me that I need to check whether we need to deactivate the "follow me" setting on centering on a route/track. But I'm not at the dev machine currently.)

PR #12564, opened 2022-01-19
Fix some code style issues reported by android studio

fm-sys:
Fix some code style issues reported by android studio:
- Explicit type argument can be replaced with <>
- 'try' can use automatic resource management
- Anonymous type can be replaced with lambda
- Lambda can be replaced with method reference
- Statement lambda can be replaced with expression lambda
- correctly annotate NightMode variable


fm-sys:
retest this please

PR #12565, opened 2022-01-19
Allow in-line comments for variables 

### LINKED ISSUE #12496: Allow in-line comments for variables
### Tell us your idea!

Sometimes it is helpful to add a small hint to a variable, e.g. add some words explaining how the value was created (in case there was a mistake and you need to recreate it)

I would suggest allowing following syntax:

```
A+B #important 
```
Parsing should end, as soon the #-character is detected outside of a string. 

So `A+B #+C` should be evaluated as `A+B`

### Is there an alternative?

Instead of `#` we could also use `//` but I guess that would be more difficult to parse cause `/` is already a mathematic operator.

### Additional Information

_No response_

fm-sys:
Fix #12496

@eddiemuc a review would be appreciated ;-)

### DIFF
--- a/main/src/cgeo/geocaching/utils/formulas/Formula.java
+++ b/main/src/cgeo/geocaching/utils/formulas/Formula.java
@@ -408,6 +408,7 @@ public final class Formula {
             throw new FormulaException(EMPTY_FORMULA);
         }
         final FormulaNode x = parseExpression();
+        // line end or beginning of user comment
         if (!p.eof()) {
             throw new FormulaException(UNEXPECTED_TOKEN, "EOF");
         }
@@ -535,6 +536,9 @@ public final class Formula {
 
         if (p.eat('^')) {
             x = createNumeric("^", new FormulaNode[]{x, parseFactor()}, (nums, vars) -> Value.of(Math.pow(nums.getAsDouble(0), nums.getAsDouble(1))));
+        }
+        if (p.eat('#')) {
+            p.parseUntil(c -> '#' == c, false, null, true); // drop potential user comments
         }
 
         return x;
--- a/main/src/cgeo/geocaching/utils/TextParser.java
+++ b/main/src/cgeo/geocaching/utils/TextParser.java
@@ -88,7 +88,7 @@ public final class TextParser {
 
     /** Returns true if end of parseable expression was reached */
     public boolean eof() {
-        return ch == 0;
+        return ch == END_CHAR;
     }
 
     /** Marks current parse position */
--- a/tests/src/cgeo/geocaching/utils/formulas/FormulaTest.java
+++ b/tests/src/cgeo/geocaching/utils/formulas/FormulaTest.java
@@ -361,4 +361,21 @@ public class FormulaTest {
         assertThat(String.format(Locale.US, "%c%09.5f°", c, d)).isEqualTo("E009.13246°");
     }
 
+    @Test
+    public void userComments() {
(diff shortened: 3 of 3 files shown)


main/src/cgeo/geocaching/utils/formulas/Formula.java
```diff
@@ -408,7 +408,8 @@ protected void doCompile(final String rawExpression, final int startPos, final F
             throw new FormulaException(EMPTY_FORMULA);
         }
         final FormulaNode x = parseExpression();
-        if (!p.eof()) {
+        // line end or beginning of user comment
+        if (!p.eof() && !p.eat('#')) {
```
eddiemuc, line 412: I understand what you want to reach, but I believe if the code is applied like this then there might be some unwanted side effects in calculated coordinate parser (which uses concatennated Formula parsing instances) in case # is used there

E.g. something like this might actually get parsed: `N48 12#comment.345`.

But I haven't tried this out

eddiemuc: Formula class must set the Parser position to the next character AFTER the parsed formula and save in the expression the exact part of the string which was parsed from rawExpression (if stopChecker is used then the epression is not parsed until its end always, this is heavily used by DegreeFormula). 

I think your code will not put the parser position at the end of the parsed part of the rawExpression. It aborts when it finds #, leaving the parser position on the #. Parsing must continue until end of expression

Additionally I would allow also to close a comment again (e.g. by a second #) this would allow commenting also inside coordinate expressions without harming the parsing.

Finally, your solution only allows comments on the END of a parsed formula expression and not somewhere inbetween. For example `3.14 #this is pi# * R #this is the radius used # * 2` would not work

eddiemuc (changes requested):

eddiemuc:
I would implement this feature directly into TextParser class, e.g. adding a method setComments(final boolean allowComments). If set to true, then the "next" method, if it detects an #, should autoskip all chars until eof is reached or a second (comment-ending) # is found. What do you think?

eddiemuc:
Hold on I spoke too soon. Such a solution could not differentiate between # in string literals and those starting a comment. An all purpose solution is maybe not that easy

fm-sys:
I'm open for every solution, just tell me at which place I have to start. Probably you know the code best... 

From looking at the code, maybe the `parseConcatBlock()` method would be a good starting point, if we also want to allow comments inbetween?

eddiemuc:
Yes, probably concat block would be a good starting point, and comments could be handled a bit like string literals during parsing (except that of course their "content" is dismissed, and that reaching EOF while inside a comment must be handled as ok Situation while a previous concat block must still be handled).

But I would have to try it out myself, it's hard to pre-think how this would work best.

A good starting point might be to write unit tests to think about possible situations and how they should be treated by the implementation - and then start implementing this.

eddiemuc:
Looks really good already

eddiemuc:
PR looks fine for me. @fm-sys I don't know if you still want to work on this, but I think it is ready for merge

eddiemuc (approved):

fm-sys:
> PR looks fine for me. @fm-sys I don't know if you still want to work on this, but I think it is ready for merge

Thanks for your review! I've just added another unit test containing a comment after a closed parenthesis... 

Oh, and big thanks for the unit test idea! I'm using them far too less, but they really helped while quickly finding a working solution. I will try to use them more often from now on...

PR #12583, opened 2022-01-21
automatic backup reminder (fix #6315)

### LINKED ISSUE #6315: Automatic backup / Reminder for backuping database
I recently had (unplanned) to reinstall my smartphone from scratch. I tried to use the backup file of the database, but I noticed, it was quite old.

So I suggest to have a reminder every month (or maybe after a configurable time) to perform a new backup.

moving-bits:
(screenshots in this entry are outdated, the reminder is no longer being shown as a modal `AlertDialog`)

## Description
Adds an automatic backup reminder to c:geo which offers you to either create an automated backup now or postpone it a bit:

image

There's not much to configure: You can set the interval to 0 (effectively disabling automatic backup reminders) or to a certain amount of days when to be reminded next:

image

The default interval is "14 days", and you will be reminded for your first automatic backup 14 days after your first start of the new version (at least if you do not reset the preference value for `automaticBackupLastCheck` manually).

The maximum number of backups is set to 5, and "include account data" setting will be shared with manual backups.
Automatic backups are stored in a subfolder `auto` beneath the configured backup folder, so they won't interfere with manual backups. If the maximum number of automatic backups is reached, the oldest automatic backup will be deleted automatically.

This PR is targeted to `master`, but based on `release`, so that we could cherry-pick it to the upcoming release if requested.

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -892,7 +892,7 @@
     <string name="init_backup_history_delete_warning">You have saved more backups than your configured maximum value allows. Do you really want to remove all excess backups?</string>
     <string name="init_backup_automatic">Automatic backup reminder</string>
     <string name="init_backup_automatic_summary">Reminds you every x days to create an automatic backup (in addition to any manually created backups), 0=off</string>
-    <string name="init_backup_automatic_reminder">Shall I create an automatic backup for you now? This will only take a short while.</string>
+    <string name="init_backup_automatic_reminder">Do you want to create an automatic backup? This will only take a short while.</string>
     <string name="init_user_confirmation">I know what I\'m doing</string>
     <string name="init_backup_success">c:geo\'s database was successfully copied to:</string>
     <string name="init_backup_failed">Backup of c:geo\'s database failed.</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -890,6 +890,9 @@
     <string name="init_backup_history_length_min">Only recent</string>
     <string name="init_backup_history_length_max">Keep all</string>
     <string name="init_backup_history_delete_warning">You have saved more backups than your configured maximum value allows. Do you really want to remove all excess backups?</string>
+    <string name="init_backup_automatic">Automatic backup reminder</string>
+    <string name="init_backup_automatic_summary">Reminds you every x days to create an automatic backup (in addition to any manually created backups), 0=off</string>
+    <string name="init_backup_automatic_reminder">Shall I create an automatic backup for you now? This will only take a short while.</string>
     <string name="init_user_confirmation">I know what I\'m doing</string>
     <string name="init_backup_success">c:geo\'s database was successfully copied to:</string>
     <string name="init_backup_failed">Backup of c:geo\'s database failed.</string>
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -324,6 +324,10 @@ public class MainActivity extends AbstractBottomNavigationActivity {
 
             DownloaderUtils.checkForMapUpdates(this);
             cLog.add("mu");
+
+            // automated backup check
+            BackupUtils.checkForBackupReminder(this);
+            cLog.add("ab");
         }
 
         if (Log.isEnabled(Log.LogLevel.DEBUG)) {
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -63,10 +63,12 @@ import android.view.MenuItem;
 import android.view.View;
 import android.widget.ArrayAdapter;
 import android.widget.ImageView;
+import android.widget.RelativeLayout;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.widget.SearchView;
 import androidx.core.view.MenuCompat;
@@ -656,4 +658,20 @@ public class MainActivity extends AbstractBottomNavigationActivity {
             actionBar.setDisplayHomeAsUpEnabled(true);
         }
     }
+
+    // display action notifications, e. g. update or backup reminders
+    public void displayActionItem(final int layout, final @StringRes int info, final Runnable action) {
+        final RelativeLayout l = findViewById(layout);
+        if (l != null) {
+            l.setVisibility(View.VISIBLE);
+            updateHomeBadge(1);
+            ((TextView) l.findViewById(R.id.action_item_info)).setText(info);
+            l.setOnClickListener(v -> {
+                action.run();
+                l.setVisibility(View.GONE);
+                updateHomeBadge(-1);
+            });
+        }
+    }
+
 }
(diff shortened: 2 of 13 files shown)


fm-sys:
> This PR is targeted to `master`

This is totally fine. This is clearly a new feature IMHO, not a bugfix. And as soon as the beta is cut off no new features should be added anymore. Release small, release often... ;-)

Oh, and please don't add another Alert Dialog. They can be really annoying for users, and in fact this is one of the points I don't like at cgeo already today. Please don't make it even worse!

We have much space in our new dashboard, so why do we need to open a separate popup window for asking the user?

Please consider implementing it in a less disinstructive way. E.g. like this:

IMG_20220121_232628_copy_432x676

moving-bits:
> > This PR is targeted to `master`
> 
> This is totally fine. This is clearly a new feature IMHO, not a bugfix. And as soon as the beta is cut off no new features should be added anymore. Release small, release often... ;-)

I know. But with bottom navigation new to our users, it might be a way longer beta phase this time, so we might add a few extensions during that phase. Or might not. I just wanted to be prepared :-)

> Oh, and please don't add another Alert Dialog. [...] Please consider implementing it in a less disinstructive way. E.g. like this:

Interesting concept, Should be applied to the map and routing tiles update checks as well, then. I'll look into it.

main/src/cgeo/geocaching/MainActivity.java
```diff
@@ -652,4 +662,31 @@ public void updateSelectedBottomNavItemId() {
...
+                l.setVisibility(View.GONE);
+                updateHomeBadge(-1);
+            });
+        }
+    }
+
+    public void updateHomeBadge(final int delta) {
+        synchronized (notificationCounter) {
+            notificationCounter.set(notificationCounter.get() + delta);
+            final BottomNavigationView bottomNavigation = findViewById(R.id.activity_bottom_navigation);
```
fm-sys, line 683: The parent class `NavigationBarView` should be used, as we don't have a bottom navigation but a navigation rail in landscape mode.

Furthermore if you want to show a badge, this code should be part of `AbstractBottomNavigationActivity` so that it is displayed always if bottom navigation is visible.

moving-bits: Thanks for your remarks, had already been working on this (moving to `AbstractBottomNavigatinActivity` etc.), but am currently stuck with merging the badge with `onLoginIssue` - how many times is this called, and under what conditions?
I'll update this PR as a WIP to let you see the current state.

moving-bits:
Current state:

image

fm-sys:
> If any of these three reminders is active, the "home" icon in bottom navigation will display a red alert badge

Red IMHO should only be used for something really urgent, e.g. "login failed"

IMHO we either show a blue "info" badge, or we don't show any badge at all (I would actually prefer the later one). Or if we show a badge, then please be consistent and show it for cgeo status notifications as well...

fm-sys:
Oh, and please use consistent margins. Currently it looks, well, yes... ;-)

grafik

moving-bits:
> > If any of these three reminders is active, the "home" icon in bottom navigation will display a red alert badge
> 
> Red IMHO should only be used for something really urgent, e.g. "login failed"

No problem, can adapt this, but what about my original question: How can I trigger that notification? So far I have not been able to do so.

> IMHO we either show a blue "info" badge

I guess it does make sense to use blue (or `colorAccent`) for the non-critical notifications, and red for the critical ones.

> please be consistent and show it for cgeo status notifications as well...

Is on my todo list, but let's do one step after another

> Oh, and please use consistent margins. Currently it looks, well, yes... ;-)

is already implemented locally. Was a bit tricky, as both the layout include and the `TextView` already had the horizontal margins applied - but it got ignored, and I had to put a `FrameLayout` with margins around it.

fm-sys:
> > Red IMHO should only be used for something really urgent, e.g. "login failed"
> 
> No problem, can adapt this, but what about my original question: How can I trigger that notification? So far I have not been able to do so.

Some insides:
- when starting an AbstractBottomNavigationActivity, `registerLoginIssueHandler` will be called. This method checks the login state and stores the result in the static variable `loginSuccessful`.
- This method will be reevaluated as soon as the network state does change / when resuming to the activity.
- The result of `registerLoginIssueHandler` will call `onLoginIssue` which sets the badge.
- `onLoginIssue` is **overwritten** in MainActivity, to show the "login failed" notification **instead** of the badge. The reason is, that we want to direct the users focus to the "login failed" notification instead of the badge, which is shown at all other AbstractBottomNavigationActivities.

To implement the "info" badge, I suggest creating a static Boolean for each dashboard notice initialized with `null`. When starting an AbstractBottomNavigationActivity, check if the Boolean is still equal to `null` and if yes, execute your checks. After the checks are done, store `true` or `false` in the boolean for each check so that is doesn't need to be reevaluated all the time. Then just extend the `onLoginIssue` method to check the other booleans as well.

For showing the actual dashboard notice in MainActivity, overwrite the `onLoginIssue` (probably this method should be renamed then) and show the notifications. If you declare the booleans as protected, you can directly access them inside MainActivity.


Not sure if there are better ways, but that's at least the way how I would do it...

moving-bits:
Thanks for the additional insight. I've made some progress meanwhile:
- adjusted margins
- a blue badge is being shown on low prio notifications (currently: the three reminders for backup, map update, tile update), a red notification on login problems
- both notification states are cached, so when you switch between the bottom navigation activities, there are no new requests, and the current state is preserved

image.image

main/res/drawable/ok.xml
```diff
@@ -0,0 +1,11 @@
+<!-- taken from material.io icons / check -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
fm-sys, line 2: I'm quite sure we already have a checkmark icon inside cgeo... IIRC it's `ic_menu_done`

main/res/values/strings.xml
```diff
@@ -890,6 +890,9 @@
     <string name="init_backup_history_length_min">Only recent</string>
     <string name="init_backup_history_length_max">Keep all</string>
     <string name="init_backup_history_delete_warning">You have saved more backups than your configured maximum value allows. Do you really want to remove all excess backups?</string>
+    <string name="init_backup_automatic">Automatic backup reminder</string>
+    <string name="init_backup_automatic_summary">Reminds you every x days to create an automatic backup (in addition to any manually created backups), 0=off</string>
+    <string name="init_backup_automatic_reminder">Shall I create an automatic backup for you now? This will only take a short while.</string>
```
fm-sys, line 895: IMHO the "Shall I" Form should only be used in the installation wizard, but nowhere else...

Lineflyer: I would also stay with something more formal here.

main/res/drawable/ok.xml
moving-bits: ok, exchanged it for ic_menu_done, removed `ok.xml`

main/res/values/strings.xml
moving-bits: Aligned the text to the map/routing tile reminders: "Do you want to create an automatic backup? ..."

moving-bits:
image

Lineflyer:
> ...and "include account data" setting will be shared with manual backups.

What is the default for this one if the user never touched it?
Background: I would like to ensure, that this automatic backup does not backup personal data (as included in settings because of the login date) without explicit acknowledgement of the user (e.g. by prior setting the same for his manual backups).

moving-bits:
Default is `false`:
```
    public static boolean getBackupLoginData() {
        return getBoolean(R.string.pref_backup_logins, false);
    }
```

PR #12585, opened 2022-01-22
fix #12572: handle strange DMM formats

### LINKED ISSUE #12572: DMM coordinate with spaces incorrectly recognized as some kind of DMS
Create a waypoint with this the new formula calculator and this input
N 50° 00.1 2 3
E 7° 00.1 2 3

and enter the variable values

Expected: DMM Coords -> N50°00.123 E007°00.123
Actual: DMS Coords -> N50°00.1'23 E007°00.1'23

The input isn't a valid DMS notation, DMS must have a ' to seperate minutes and seconds but the input uses only a .
So the input can only be DMM

In addition: As DMM is way more common in in geocaching than DMS in doubt coords should be interpreted as DMM, only input that's clearly DMS (containing °, ', '') should be interpreted as DMS

Repoduced with a build based on https://github.com/cgeo/cgeo/commit/9eec3c1f1733924c19cee6eda21f5594ebd660a3 (16.1.22)

eddiemuc:
fix #12572: handle strange DMM formats

### DIFF
--- a/main/res/layout/waypoint_item.xml
+++ b/main/res/layout/waypoint_item.xml
@@ -111,6 +111,7 @@
                 android:textColor="@color/colorText_listsPrimary"
                 android:visibility="gone"
                 android:text="@string/waypoint_calculated_coordinates"
+                android:textStyle="italic"
                 tools:visiblity="visible" />
 
             <TextView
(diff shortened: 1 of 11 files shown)


eddiemuc:
image

PR #12586, opened 2022-01-22
show better display name for tracks

fm-sys:
grafik


moving-bits (approved):
Ah, that's nice! I was looking for something like this, but couldn't get rid of the prefix...

bekuno:
retest this please

PR #12588, opened 2022-01-22
rel to #12384: first version of waypoint generator from variable ranges

eddiemuc:
rel to #12384: first version of waypoint generator from variable ranges

First version for generation of waypoints from calculated coordinates

### DIFF
--- a/main/res/layout/coordinatescalculateglobal_dialog.xml
+++ b/main/res/layout/coordinatescalculateglobal_dialog.xml
@@ -173,6 +173,13 @@
                 android:text="Convert to plain coordinate" />
 
             <Button
+                android:id="@+id/generate_range_coordinates"
+                style="@style/button_full"
+                android:layout_width="fill_parent"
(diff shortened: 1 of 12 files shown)


eddiemuc:
Step 1: specify a **range** value for one (or more than one) variable using [], e.g. [0-9]. Values can be excluded using prepended ^, single values can be added via comma. For example [0-9, 11, ^3-6] would specify the values 0, 1, 2, 7, 8, 9, 11.

The first specified value is used per default in the GUI for all calculations. As of now only positive integers are allowed as values:

image


Step 2: use new button "generate range coordinates". Coords are calculated and listed in multiselect dialog:

image


Step 3: select coordinates where waypoints should be created and click ok to start creation:

image

eddiemuc:
retest this please

eddiemuc:
Added possibility to just show generated waypoints on map without storing:

image

image


image

Issue #12379, opened 2021-12-23
[Global Calculated Coordinates/Waypoints] Migrate local calculated coordinates/waypoints to global calculated coordinates/wayppoints


eddiemuc:
In order to replace the local calculated waypoints  with global calculated waypoints, existing waypoints must be migrated to the new format.

Different aspects are to be considered:
* Local waypoints must be migrated. Since different waypoints in same cache may use same variable names, a variable replacement strategy must be revised so each coordinate will kep its original value even after migration. 
* Previously exported calculated coordinates/waypoints to Personal Note must still be readable by c:geo (and then be transferred into global coordinate/waypoint)

eddiemuc:
Migration Test Data: "Degree Example 1":

![image](https://user-images.githubusercontent.com/6909759/151664026-a82aae46-7230-4f94-88ac-797d4c594327.png)

[...]

PR #12598, opened 2022-01-23
rel to #12379: introduce calculated waypoint migration preview

eddiemuc:
rel to #12379: introduce calculated waypoint migration preview


eddiemuc:
This PR, if merged, will introduce a "Calculated Waypoint Migration Preview" feature. It allows us to see how old calculated waypoints would be migrated to the new global cache variables without actually performing the migration. This allows for testing and discussion of the migration.

Feature is of course restricted to non-productive setups (development and nightly )

If one opens a waypoint list where there is at least one calculated waypoint (old method), a new button "Preview Waypoint Migration" appears:

image

Clicking it opens a simple Messagebox with information:
* Which waypoints would be due for migration
* Which global vars exist in the cache BEFORE migration
* How the coordinates of the migrated caches would be changed, and
* Which new global vars would be added by the migration

image

(Depending on how many calculated waypoints are in the cache, the text field can get quite long)

This PR is not primarily about the actually implemented algorithm for migration but should allow testing and discussion on this algorithm. As of now, following is implemented:
* All local variables in calculated coordinates, if there's already a var with same name and different formula on the cache, are appended with an index value to create a new unique var. E.g. `A` becomes `A2` (or if `A2` is also already in use then `A3` and so on)
* The var is replaced in formulas using this var using the syntax `($varname)` (this syntax will be recognized e.g. by the guided mode of the new variable calculator as a formula field due to the parenthesis).

Lineflyer:
Will read through and answer next days and meanwhile think about what would be expected migration from my perspective.

eddiemuc:
> Will read through and answer next days and meanwhile think about what would be expected migration from my perspective.

Let's move the overall discussion to #12379

eddiemuc:
Putting to wip, please don't merge

eddiemuc:
Process was changed: 

now whenever user opens a waypoint for editing containing legacy calculation coordinate data, a dialog like the following opens:
image

The user can choose to migrate the data as suggested, dismiss the old calculation data or postphone migration (cancel).

eddiemuc:
In the example above, there was already a variable a defined in the cache, but no variables A, F or G

PR #12603, opened 2022-01-23
remove spacing after attribute group titles (fix #12601)

### LINKED ISSUE #12601: Improve attribute grouping in cache details
Beta users feedback posted on our  Facebook page:

I'd consider removing empty line after group title so when there is only one attribute for group it would be more tightly grouped to the title instead of being title, empty, attribute, empty, title, empty, attribute... Maybe intend attributes a bit or add bullets in front of them?

moving-bits:
## Description
Replaces `<h5>`/`<h6>` formatting of attribute category titles by bold/underline to remove spacing after title.

Cache details page:
|before|after|
|---|---|
|image|image|

Cache list page:
|before|after|
|---|---|
|image|image|

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1281,7 +1281,10 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                         final String key = attr.getValue(enabled);
                         if (attributes.contains(key)) {
                             if (lastCategory != category) {
-                                attributesText.append("<h6>").append(category.getName(activity)).append("</h6>");
+                                if (lastCategory != null) {
+                                    attributesText.append("<br /><br />");
+                                }
+                                attributesText.append("<b><u>").append(category.getName(activity)).append("</u></b><br />");
                                 lastCategory = category;
                             } else {
                                 attributesText.append("<br />");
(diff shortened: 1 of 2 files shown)


ztNFny:
imho the new version is too condensed. And a heading-element seems like the logical choice for a attribute group heading

moving-bits:
> imho the new version is too condensed. And a heading-element seems like the logical choice for a attribute group heading

technically agreed. There was the explicit request to remove the huge whitespace below the heading, which comes when using `<h5>` or even `<h6>`, so I was looking for a different visualization of title vs. follow-up text, and came to using bold and underline. 

If there's a reasonable alternative to achieve a visible but smaller gap between heading and following paragraph (e. g.: a quarter of a line), we can try that instead.

ztNFny:
> If there's a reasonable alternative to achieve a visible but smaller gap between heading and following paragraph (e. g.: a quarter of a line), we can try that instead.

Styled element with margin-bottom set?

I don't have an issue with that whitespace, it's standard-styling for headlines to have space before the paragraph. To have less space would be ok, but to have none at all?

Ideal solution: Headline with less margin-bottom. If that's not possible I like the old better than the new. 

>  There was the explicit request to remove the huge whitespace below the heading

One users opinion shouldn't auto-trigger a change. Else I just explicitly request it changed back ;-)

bekuno:
> Ideal solution: Headline with less margin-bottom. If that's not possible I like the old better than the new.

I like the new style better than the old one. It is more compact and therefore more readable.
On the other hand, a headline styling would look better. Is it possible to use a css styling here?

PR #12623, opened 2022-01-28
material design shortcut icons

### LINKED ISSUE #11969: Align widget/shortcut icons with new material bottom nav icons
Edit note: Now this issue is about the widget / shortcut icons only

Simple question:
Do we want to stay with the current (kind of standard) icons on the bottom navigation?
On one hand it looks simple and modern, on the other hand c:geo looses a bit of its character without the custom icon and it looks just like any other app.
Don't get me wrong, I think the old icons cannot be reused but maybe some monochrome upgrades of the legacy icons?

The reason why this came to my mind is, that we still use the legacy icons for the widgets. So those should be replaced anyway.

ztNFny:
Provides material design shortcut icons to match bottom nav

image

fixes #11969

background color, shape and icon size are dictated by Material Design guidelines: **https://material.io/archive/guidelines/style/icons.html#icons-app-shortcut-icons**


fm-sys:
First of all, thanks for working on it!

That said, I think the current layout definitely needs improvement...

Launcher icons should always be "adaptive", which means the shape should follow the launcher default which differs between manufacturers.



Screenshot_20220130_004104_com huawei android launcher

Furthermore, we have a second way how users can create shortcuts. And this one looks totally broken currently...


Screenshot_20220130_004159_com huawei android launcher
Screenshot_20220130_004132_cgeo geocaching

ztNFny:
> Launcher icons should always be "adaptive", which means the shape should follow the launcher default which differs between manufacturers.

Material design guidelines say: "App shortcut icons are circular."

> Furthermore, we have a second way how users can create shortcuts. And this one looks totally broken currently... 

Oops, didn't know of that way. Strange though, it was using the same drawables as shortcut icons before and still is, wonder why it gives such a strange result now.

fm-sys:
> > Launcher icons should always be "adaptive", which means the shape should follow the launcher default which differs between manufacturers.
> 
> Material design guidelines say: "App shortcut icons are circular."

What was originally suggested and what has finally prevailed are two totally different things through.

As material.io design guides are written by Google, it's always a good idea to look how they handle it themselves:

Screenshot_20220130_132731_com huawei android launcher (1)

But even within individual apps it's inconsistent:

Screenshot_20220130_133007_com huawei android launcher

However, it seems to me that the adaptive version is way more popular.

ztNFny:
> As material.io design guides are written by Google, it's always a good idea to look how they handle it themselves:

imho Google is one of the worst examples to look at when it comes to following Material Design guidelines ;-) They ignore their own "rules" all the time. I usually look at the Material Design Awards winners for inspiration.

PR #12625, opened 2022-01-29
Fix #11664 - has_user_defined_waypoint filter added

### LINKED ISSUE #11664: "has user-defined waypoints" filter being missed
### Tell us your idea!

A user from support misses the filter "has user-defined waypoints" (ticket 779609)

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

Volthai7us:
## Related issues

Fix #11664


eddiemuc:
Ok to test

Volthai7us:
I added comments to some places, I can delete them and commit again?

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
```diff
@@ -75,6 +77,7 @@
     private Boolean statusHasOfflineLog = null;
     private Boolean statusHasOfflineFoundLog = null;
     private Boolean statusSolvedMystery = null;
+    private Boolean statusHasUserDefineWaypoint = null;//Volthai7us
```
eddiemuc: Please remove the comments with your GitHub name (//Voltahai7us) from the PR. We don't do this (would make code pretty messy over time :-)). `git blame` can be used anytime to find out who did what if need be.

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
```diff
@@ -224,6 +229,12 @@ public void setStatusSolvedMystery(final Boolean statusSolvedMystery) {
         this.statusSolvedMystery = statusSolvedMystery;
     }
 
+    public Boolean getStatusHasUserDefineWaypoint() { return statusHasUserDefineWaypoint; }//Volthai7us
```
eddiemuc: Code does not follow c:geo Codacy/checkstyle rules (see breaking Codacy run), those need to be fixed. 

I recommend you install the "checkstyle" plugin in your Android studio, run it using the c:geo configuration (checkstyle.xml) and fix all findings before pushing to server. This might save you many round-trips :-)

eddiemuc (dismissed):
Looks good apart from some "styling" stuff (see comments + red codacy check) which needs to be fixed.

fm-sys:
@eddiemuc do you have time to invent the corresponding SQL query? Cause if we don't do it directly now, we most probably do it never...

eddiemuc:
> @eddiemuc do you have time to invent the corresponding SQL query? Cause if we don't do it directly now, we most probably do it never...

If we feel that this is needed for a complete feature then we can't allow the PR to be merged before this is implemented as well. I am indifferent to this. Feature will work without, there's just a slight (in practice non-noteable) performance peak to it. I suggest to either create a separate issue for this or block this PR in review until we are satisfied. What do you think?

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
eddiemuc: ok now, thank you

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
eddiemuc: ok now, thank you

Volthai7us:
After I finished addToSql method implementation should I provide new PR or commit to this PR.

moving-bits:
> After I finished addToSql method implementation should I provide new PR or commit to this PR.

You can commit to this PR.

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
```diff
@@ -224,6 +229,12 @@ public void setStatusSolvedMystery(final Boolean statusSolvedMystery) {
         this.statusSolvedMystery = statusSolvedMystery;
     }
 
+    public Boolean getStatusHasUserDefineWaypoint() { return statusHasUserDefineWaypoint; }
```
SammysHP, line 232:
```suggestion
    public Boolean getStatusHasUserdefinedWaypoints() { return statusHasUserdefinedWaypoints; }
```

main/src/cgeo/geocaching/filters/core/StatusGeocacheFilter.java
```diff
@@ -108,6 +112,7 @@ public Boolean filter(final Geocache cache) {
             (statusHasOwnVote == null || (cache.getMyVote() > 0) == statusHasOwnVote) &&
             (statusHasOfflineLog == null || cache.hasLogOffline() == statusHasOfflineLog) &&
             (statusHasOfflineFoundLog == null || hasFoundOfflineLog(cache) == statusHasOfflineFoundLog) &&
+            (statusHasUserDefineWaypoint == null || (cache.hasUserdefinedWaypoints()) == statusHasUserDefineWaypoint) &&
```
SammysHP, line 115:
```suggestion
            (statusHasUserdefinedWaypoints == null || cache.hasUserdefinedWaypoints() == statusHasUserdefinedWaypoints) &&
```

SammysHP (changes requested):
I suggest to keep the naming consistent to `cache.hasUserdefinedWaypoints()`.

Volthai7us:
@SammysHP I will apply your suggest when I commit sql part.

Lineflyer:
FYI:
I took the freedom to reference the issue in this PR title (which allows auto-close once merged) and set the "Work in progress" label for the time being.

fm-sys:
> FYI: I took the freedom to reference the issue in this PR title (which allows auto-close once merged) and set the "Work in progress" label for the time being.

PR title will not work properly in all cases. The "fix #..." should be contained in the PR description or commit message. I added it to the PR description now :)

Volthai7us:
@moving-bits could you check again?

eddiemuc (approved):
Looks good to me. Thanks a lot for the contribution!

I won't merge since review of @moving-bits is requested and there's still a pending change request from @SammysHP

fm-sys (commented):
Seems like I've discovered two small copy-and-past mistakes. Maybe you can take a look… Apart from that it looks fine, thanks for the contribution 👍

fm-sys:
> @fm-sys Sorry I will fixed these problem in 2 hours

All fine, don't hurry :)

moving-bits (changes requested):
LGTM generally, thanks for your contribution!
Two things are left to do: As there have been other code changes meanwhile, targeting the same code parts, you'll have to do a so-called "rebase", resolving any conflicts along the way. (Lookup `git rebase upstream/master`, if you're unfamiliar with that.)
Also some variable naming inconsistencies are still present, see the remark from @SammysHP (use plurals form for `statusHasUserDefinedWaypoint` => `statusHasUserDefinedWaypoints` and `getStatusHasUserDefinedWaypoint` => `getStatusHasUserDefinedWaypoints`)

fm-sys:
> As there have been other code changes meanwhile, targeting the same code parts, you'll have to do a so-called "rebase", resolving any conflicts along the way.

As this can be quite confusing for git beginners, I  actually would recommend not to do so.

@moving-bits
While there are conflicts when selecting `rebase merge`, it however works just fine if you select `squash and merge`. The PR should be squashed while merging anyway, so no action required from @Volthai7us IMHO. 

Only the `Waypoint` -> `Waypoints` thing is still outstanding... 

Screenshot_20220210_064407_org mozilla firefox_copy_269x151

Volthai7us:
I hope it's okey now @fm-sys @moving-bits

moving-bits (approved):
> @moving-bits While there are conflicts when selecting `rebase and merge`, it however works just fine if you select `squash and merge`.

Ah, you're right, thanks for pointing this out.

> Only the `Waypoint` -> `Waypoints` thing is still outstanding...

Thanks @Volthai7us for fixing this. To my understanding all open review requests are solved by now. Have removed the "WIP" label therefore.

moving-bits:
Let's merge here now. Thanks @Volthai7us for your contribution.

PR #12630, opened 2022-01-29
use high-res version of avatar image 

fm-sys:
fix regression introduced by #12621

While using the latest nightly, I noticed that the avatar is displayed with a way worse resolution than before. Thus, this will change the code to use the high-res version of avatar image again...


Lineflyer:
Just wanted to report that very issue. Thanks for fixing!

PR #12632, opened 2022-01-29
fine-tune dashboard status messages (fix #12631)

### LINKED ISSUE #12631: Automated backup check - improvements
### Describe your problem!

I just saw the automated backup check for the first time.

So I just hit the "check" button, but nothing happened. I did wait some time, pressed it again - nothing! After closing and reopening the app, it still didn't work, and I haven't found any way to get it triggered. 

### How to reproduce?

- open "view settings"
- set `automaticBackupLastCheck` to a low number but higher than 0. (e.g. 1)
- close cgeo
- notification is displayed but doesn't work

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2022.01.29-NB-071456a

### System information

_No response_

### Additional Information

Furthermore, a visual bug: IMHO the action button should be vertically centered

Screenshot_20220129_132745_cgeo geocaching

moving-bits:
## Description
Fine-tuning for dashboard messages regarding backup & update reminder:
- remove action button => tapping on message activates action
- adapted messages accordingly
- fix: only one message could be shown at a time - now all (up to three) can be displayed

image

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -16,19 +16,21 @@
         android:layout_marginTop="16dp"
         tools:layout="@layout/status" />
 
-    <FrameLayout
+    <LinearLayout
         android:id="@+id/reminders"
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
         android:layout_marginHorizontal="16dip"
         android:layout_marginTop="16dp"
-        android:layout_below="@id/status">
+        android:layout_below="@id/status"
+        android:orientation="vertical">
 
         <include
             android:id="@+id/autobackup"
             layout="@layout/mainactivity_action_item"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:layout_marginBottom="8dp"
             android:visibility="gone" />
 
         <include
@@ -36,6 +38,7 @@
             layout="@layout/mainactivity_action_item"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:layout_marginBottom="8dp"
             android:visibility="gone" />
 
         <include
@@ -43,9 +46,10 @@
             layout="@layout/mainactivity_action_item"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:layout_marginBottom="8dp"
             android:visibility="gone" />
 
-    </FrameLayout>
+    </LinearLayout>
 
     <RelativeLayout
         android:id="@+id/info_notloggedin"
--- a/main/res/layout/mainactivity_action_item.xml
+++ b/main/res/layout/mainactivity_action_item.xml
@@ -7,7 +7,6 @@
     android:padding="8dp"
     android:background="@drawable/helper_bcg"
     android:gravity="center"
-    android:text="@string/init_backup_automatic_reminder"
     android:textColor="@color/text_icon"
     android:textIsSelectable="false"
     android:textSize="@dimen/textSize_detailsPrimary" />
--- a/main/res/layout/mainactivity_action_item.xml
+++ b/main/res/layout/mainactivity_action_item.xml
@@ -1,27 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/action_item_info"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:background="@drawable/helper_bcg">
-
-    <TextView
-        android:id="@+id/action_item_info"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_centerVertical="true"
-        android:layout_marginHorizontal="8dp"
-        android:layout_toLeftOf="@id/action_item_button"
-        android:gravity="center"
-        android:text="@string/init_backup_automatic_reminder"
-        android:textColor="@color/text_icon"
-        android:textIsSelectable="false"
-        android:textSize="@dimen/textSize_detailsPrimary" />
-
-    <Button
-        android:id="@+id/action_item_button"
-        style="@style/button_icon"
-        android:layout_alignParentRight="true"
-        app:icon="@drawable/ic_menu_done"/>
-
-</RelativeLayout>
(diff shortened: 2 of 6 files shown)


Lineflyer:
What I ask myself when seeing this messages though:
How can I now decline them to get rid of them if I do not want that action to be performed.

moving-bits:
> What I ask myself when seeing this messages though: How can I now decline them to get rid of them if I do not want that action to be performed.

With not having a dialog box any longer, you can only ignore them currently, or disable the functionality completely. But there's no way to postpone an action any more (no "later" action as we had in the dialog boxes).

If we want to have a "later" action, we could re-add a button, this time with a clock symbol or something like this.

Lineflyer:
How about something like "Do you want ..... Tap here to dismiss or perform this action" followed by a  yes/no dialog.

At least I would like to avoid not being able to get rid of it without performing the action.

moving-bits:
But wouldn't we have sort of a doubled dialog then? Sort of "Tap here to update" => "do you really want to update" => "yes I REALLY DO WANT TO UPDATE, I told you so" or "oops, no, I actually don't want to update"... :-)

Maybe a long-tap to dismiss temporarily?

Lineflyer:
> Maybe a long-tap to dismiss temporarily?

Good idea. IMHO that is more wise than buttons as long as the text describes it.

moving-bits:
Added ability to postpone dashboard actions by long-tap, extended messages accordingly:

image

Lineflyer:
Looks not that bad. I would not put the last part in parenthesis though.

Sorry for asking rather late:
If the user accepted "auto backup". Will further auto backups just be done or will this message be recurring? 
In that case the question would be: What is automatic about that?

I would prefer it in a way like "Do you want to do enable a regular automatic backup? Tap here to accept or long tap to decline." and from then on it will be done really automatically in background.

moving-bits:
> Looks not that bad. I would not put the last part in parenthesis though.

Ok, will adapt that.

> Sorry for asking rather late: If the user accepted "auto backup". Will further auto backups just be done or will this message be recurring? In that case the question would be: What is automatic about that?
> 
> I would prefer it in a way like "Do you want to enable a regular automatic backup? Tap here to accept or long tap to decline." and from then on it will be done really automatically in background.

We had a short discussion about that in the original issue: As we need to have exclusive access to the database, we need to inform the user anyway, and block any user actions until backup has been completed. But we did not want to force the user to be interrupted in the moment c:geo likes to perform an automatic update, so we decided to go with a message (we started with an `AlertDialog`, but switched to the current less-intrusive way later).

Switching on/off can be done in preferences, with default being "on".

fm-sys:
Sorry, but what is the purpose of postponing the backup? (Especially as normal notifications can't be postponed either) 

I can only imagine two use cases:
- I want to do the backup not now but some minutes later
- I want to do the backup never

If I want to do it in some minutes, the postpone feature will not help me as there is no way to get the dismissed notification reappearing. Instead, I can just ignore the notification until I'm ready. 

If I want to do no automated backups at all, it will be annoying if the notification is always reappearing. 

IMHO the only thing which would make sense is to link the long click action to the "backup settings" page where the user can configure this notification.

Lineflyer:
> IMHO the only thing which would make sense is to link the long click action to the "backup" screen in our settings where the user can configure this notification.

That would also be fine from my perspective. 
So the text would then be something like:
"This is a backup reminder. Tap here to backup your data now or long tap to configure."

Lineflyer:
The question would then be if we still need a separate backup storage path after all.

moving-bits:
I would leave it exactly as it is now, for several reasons:
- We were looking for a mostly-automated way to create backups, as most people seem to be unaware of this functionality, or are too lazy to use it. Putting it into the main menu did not help as much as wanted, so we want to make it more prominent. `AlterDialog` was too much for some, so we put it as a message on dashboard. Just linking from there to the regular backup mechanism is one step too much, IMHO, the action should be executed from the notification directly.
- We wanted to separate those semi-automatic backups from the regular ones, as "power users" may have created backups for specific reasons, and we don't want to overwrite those backups. And, again, asking "do you want to overwrite" might be too much for the average user in this case.
- Regarding postponing: For the "postpone by a few minutes" you simply endure the notification for that timespan. The more frequent scenario would be "I'm out on a caching tour today, so don't bother me today", IMHO - therefore a delay by a few days.
- The way it is implemented currently is very similar to the notifications for map updates and routing tile updates, so it's pretty consistent.

The only alternative I see is to enforce automatic backups, which goes into the direction of what @Lineflyer has suggested a bit further up:
- if automatic backups are enabled,
- check on c:geo start whether one is due,
- and do a full-automatic backup, if so, without any further question, blocking c:geo for that time.

This would delay the start by a few seconds in those cases, but would have minimal user-interaction impact otherwise, and would guarantee a regular backup (if not explicitly disabled by the user).

The one thing I'm unsure about this in that constellation is how big the impact will be for the user - how long will such a backup take, especially in the combination of relatively slow device and relatively large database, and would this impact be tolerable.

fm-sys:
> Regarding postponing: For the "postpone by a few minutes" you simply endure the notification for that timespan. The more frequent scenario would be "I'm out on a caching tour today, so don't bother me today", IMHO - therefore a delay by a few days.

A backup takes only a few seconds and doesn't even require internet connection. Most probably the use case would be "not now but this evening", but then the "postpone" feature would do more harm than good.

Again: Why should it be considered bothering at all? Normal notifications (e.g. "an update is available") can't be postponed either...

moving-bits:
> A backup takes only a few seconds and doesn't even require internet connection. 

Which leads us to a full-automatic solution then, as described above.

So if no one stops me, I will update the PR this evening, changing the auto backup behavior to:
- not display a notice on the dashboard,
- check need for auto-backup on startup, and perform it without further user interaction, if time has come
- (otherwise PR will be unchanged, e. g. configurability of auto-backup interval and dashboard notices for map/routing data update)

Lineflyer:
Lets start with that, then test and optimize if needed.

moving-bits:
> Lets start with that, then test and optimize if needed.

You mean "start with as it is now" or "start with full-automated" as described in my last post?

Lineflyer:
> You mean "start with as it is now" or "start with full-automated" as described in my last post?

Start with what you proposed (full-automatic).
There might however be some cases which need consideration, as e.g. what happens if backup cannot be done for whatever reason, first start behavior, etc. 
Thats what I meant with optimize if needed.

moving-bits:
updated PR accordingly

Issue #10744, opened 2021-05-24
Material theming: No transparent background for mainscreen


moving-bits:
## Bug description 
While drastically reducing the number of our themes, the layout options for having a transparent mainscreen background got kicked-out.

We can either remove that feature officially (need to remove the related setting, then), or need to implement the necessary theme or activity options to get transparency back. What do we want?

Lineflyer:
I think we can remove the transparent background. Its a relict of old times somehow. 
Even more our main menu will hopefully be reworked as well thus requiring a solid background anyway.

eddiemuc:
I assume the easier way would be to remove it. So I vote for doing it, although I use it myself

(I also remember being irritated a lot by this feature the first time I ever used c:geo, I don't know any other app which by default lets the system background shine through)

moving-bits:
> I assume the easier way would be to remove it. So I vote for removing it, although I use it myself
> 
> (I also remember being irritated a lot by this feature the first time I ever used c:geo, I don't know any other app which by default lets the system background shine through)

Same here. I'll remove the settings' options for transparency.

MagpieFourtyTwo:
To bed honest: I was kind of shocked, when I lost the beautiful background. Really liked to see my nice wallpaper in the app. Was *way* more beautiful than it is now ... boring grey ... really sad about it. :/

MagpieFourtyTwo:
Have to correct myself: It's not boring, but even mindboggling ugly. :{

Any plans for a background image? Or at least a friendly deep black?

MagpieFourtyTwo:
Just went a bit deeper and opened a cache. Sorry, guys, but this (dark) theme is absolutely unusable, esp  in the dark: text just grey on grey, the lack of contrast is a real pain, you really have to concentrate to read text, and in the very next moment, esp. on change to waypoint page, your eyes get burned by the extremely intrusive, stabbing, eye-burning orange of the oversized switches.

No, sorry, I really love c:geo, but right now it's a real pain to use this wonderful app. :(((

eddiemuc:
Thanks a lot for putting the new theme implementation in! I am pretty sure it will easify future changes of the GUI towards a more modern design in numerous ways (the drastic reduction/cleanup of themes/styles alone is a vast improvement!)

There might still be GUI glitches (I see for example many discussions wrt background colors :-)) but I am sure we can resolve them until next release.

eddiemuc:
Again, many thanks to @moving-bits for tackling this.

[...]

PR #12639, opened 2022-01-30
optionally use wallpaper as dashboard background (fix #11867)

### LINKED ISSUE #11867: Allow custom background on dashboard
### Tell us your idea!

To be discussed/agreed:

With the recent changes of main menu design (before bottom navigation dashboard was introduced) we removed the transparent background option.

While there is not any functional regression still some users asked on support mail and social media, whether this option or the option to assign a custom color or image will come back.

There could be a feature of loading a custom image and/or setting a custom color to the dashboard.

I am undecided, whether this is worth any effort (and not sure if someone is willing to implement) and it might bring new problems (e.g. font color contrast might become bad depending on color/picture), still I want to bring this into discussion, whether a majority here would like to see/have such option or not.

### Is there an alternative?

Keep it as is.

### Additional Information

_No response_

moving-bits:
## Description
Allow users to activate current system wallpaper as dashboard background

image.image

Default value is `false` = "do not use wallpaper as background"

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -154,6 +154,7 @@
     <string translatable="false" name="preference_screen_appearance">preference_screen_appearance</string>
     <!-- ============================================================================================================================================================================== -->
     <string translatable="false" name="pref_theme_setting">theme_setting</string>
+    <string translatable="false" name="pref_wallpaper">wallpaper</string>
     <string translatable="false" name="pref_showaddress">showaddress</string>
     <string translatable="false" name="pref_plainLogs">plainLogs</string>
     <string translatable="false" name="pref_selected_language">selectedLanguage</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -864,6 +864,8 @@
         <item>Light</item>
         <item>Dark</item>
     </string-array>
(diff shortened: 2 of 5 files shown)


Lineflyer:
Nice, didn't even know that there could be such a smart solution.

Lineflyer:
Question is, if we would need to outline fonts in such cases (to avoid e.g. white font on white background). But thats probably under responsibility of the user when choosing his wallpaper...

moving-bits:
> Nice, didn't even know that there could be such a smart solution.

This is thanks to @eddiemuc who had preserved that finding as a [comment](https://github.com/cgeo/cgeo/issues/10744#issuecomment-875609348)

moving-bits:
> Question is, if we would need to outline fonts in such cases (to avoid e.g. white font on white background). But thats probably under responsibility of the user when choosing his wallpaper...

One could discuss whether we want to set the activities background (dark grey in my case) to those text fields explicitly - would make no difference when no wallpaper is selected, and would give a colored background beneath the texts otherwise.

Lineflyer:
`release` or `master` to afterwards check/optimize the background issue?

fm-sys:
I guess `release` is fine. This setting isn't enabled by default anyway...

PR #12640, opened 2022-01-30
add adaptive shortcut icons for SDK26+

ztNFny:
supports adaptive icon shapes for shortcuts in SDK26+
fixes display of Widgets

follow-up on https://github.com/cgeo/cgeo/pull/12623#issuecomment-1025009003

Widget icons are always round as they must be generated from static resourcves

image


fm-sys (approved):
Looks much better IMHO, thanks for working on it 👍

AFAIK it is somehow possible to send `shortcuts` to the launcher as well, so we could think about replacing the old legacy `widgets` with the more modern adaptive `shortcuts` as well. But this should definitely be done in a separate PR anyway...

PR #12641, opened 2022-01-30
resume finished downloads (fix #12109)

### LINKED ISSUE #12109: [downloader] import not always triggered
### Describe your problem!

If c:geo is not running, it sometimes happens that the download doesn't get imported.

The download is complete, but the import process doesn't start - even if you restart cgeo

### How to reproduce?

It just happens sometimes... 

The file remains in the download directory but isn't processed by cgeo

c:geo should check on startup, if there are unprocessed downloads. If yes -> trigger the background import service 

### Reproducible

Unclear

### c:geo Version

2021.11.15

### System information

_No response_

### Additional Information

Screenshot_20211115_215221_cgeo geocaching

moving-bits:
## Description
Sometimes (map / routing data) downloads are finished, but somehow c:geo did not get triggered, as described in the related issue.

This PR adds a check to program start to query the system's `DownloadManager` for any finished, but not yet transmitted downloads, and ~asks the user, if they want to import them now~ directly triggers receiving such files, if any are pending.

~Currently it's realized as a dialog box, but will be changed to a dashboard notification, after PR #12632 has been merged. Therefore setting WIP label for now.~

(Additionally moves the `dumpDownloadmanagerInfos` method from `DebugUtils` to `DownloaderUtils` as it has a closer relationship.)

### DIFF
--- a/main/src/cgeo/geocaching/MainActivity.java
+++ b/main/src/cgeo/geocaching/MainActivity.java
@@ -330,6 +330,10 @@ public class MainActivity extends AbstractBottomNavigationActivity {
             // automated backup check
             BackupUtils.checkForBackupReminder(this);
             cLog.add("ab");
+
+            // check for finished, but unreceived downloads
+            DownloaderUtils.checkPendingDownloads(this);
+
         }
 
         if (Log.isEnabled(Log.LogLevel.DEBUG)) {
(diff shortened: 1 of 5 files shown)


fm-sys:
Thanks for working on it! 

> Currently it's realized as a dialog box



Importing will be done in the background anyway, so is there a reason why we have to ask users at all?

I really wonder why you like `AlertDialogs` that much... ;-)

moving-bits:
It's not about `AlertDialogs` :-) but about programs trying to think for the user and doing things "under the hood" without interacting with the user...
But in this case I guess you are right, as the actual download is already completed, and it's only about importing the file, so no unwanted network traffic involved here. I'll adapt the PR.

PR #12642, opened 2022-01-30
Log templates in offline log dialog

ztNFny:
add a menu item to the "Quick offline log" dialog to insert a Found log with text taken from template.
The menu item is shown only if any log template is defined.

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -140,7 +140,7 @@
     <string name="log_posting_image">Posting images…</string>
     <string name="log_posting_generic_trackable">Posting %1$s %2$d/%3$d</string>
     <string name="log_clear">Clear</string>
-    <string name="log_templates">Found it: Log template …</string>
+    <string name="log_templates">Found it: Log template…</string>
     <string name="log_post_not_possible">Download of data in progress…\nPlease try again in a few seconds or check your internet connection.</string>
     <string name="log_date_future_not_allowed">Future log dates are not allowed.</string>
     <string name="log_add">Add</string>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -140,6 +140,7 @@
     <string name="log_posting_image">Posting images…</string>
     <string name="log_posting_generic_trackable">Posting %1$s %2$d/%3$d</string>
     <string name="log_clear">Clear</string>
+    <string name="log_templates">Found it: Log template …</string>
     <string name="log_post_not_possible">Download of data in progress…\nPlease try again in a few seconds or check your internet connection.</string>
     <string name="log_date_future_not_allowed">Future log dates are not allowed.</string>
     <string name="log_add">Add</string>
--- a/main/src/cgeo/geocaching/log/LoggingUI.java
+++ b/main/src/cgeo/geocaching/log/LoggingUI.java
@@ -95,7 +95,7 @@ public final class LoggingUI extends AbstractUIFactory {
             list.add(new LogTypeEntry(null, SpecialLogType.CLEAR_LOG, false));
         }
         list.add(new LogTypeEntry(null, SpecialLogType.LOG_CACHE, false));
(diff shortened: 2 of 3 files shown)


main/res/values/strings.xml
```diff
@@ -140,6 +140,7 @@
     <string name="log_posting_image">Posting images…</string>
     <string name="log_posting_generic_trackable">Posting %1$s %2$d/%3$d</string>
     <string name="log_clear">Clear</string>
+    <string name="log_templates">Found it: Log template …</string>
```
moving-bits, line 143: I would either name it `Found it (log template)` or remove the space between "template" and "…"

main/src/cgeo/geocaching/log/LoggingUI.java
```diff
@@ -93,6 +95,9 @@ private static void showOfflineMenu(final Geocache cache, final Activity activit
             list.add(new LogTypeEntry(null, SpecialLogType.CLEAR_LOG, false));
         }
         list.add(new LogTypeEntry(null, SpecialLogType.LOG_CACHE, false));
+        if (Settings.getLogTemplates().size() > 0 && (logTypes.contains(LogType.FOUND_IT) || logTypes.contains(LogType.WEBCAM_PHOTO_TAKEN))) {
```
moving-bits, line 98: do we need to take `LogType.ATTENDED` into consideration as well?

main/src/cgeo/geocaching/log/LoggingUI.java
```diff
@@ -137,6 +146,30 @@ private static void showOfflineMenu(final Geocache cache, final Activity activit
...
+        builder.setTitle(R.string.log_templates);
+
+        final ArrayAdapter<Settings.PrefLogTemplate> adapter = new ArrayAdapter<>(activity, android.R.layout.select_dialog_item, templates);
+
+        builder.setAdapter(adapter, (dialog, item) -> {
+            final Settings.PrefLogTemplate logTemplate = adapter.getItem(item);
+            cache.logOffline(activity, new OfflineLogEntry.Builder<>()
+                .setLog(LogTemplateProvider.applyTemplates(logTemplate.getText(), new LogTemplateProvider.LogContext(cache, null, true)))
+                .setDate(Calendar.getInstance().getTimeInMillis())
+                .setLogType(cache.getPossibleLogTypes().contains(LogType.FOUND_IT) ? LogType.FOUND_IT : LogType.WEBCAM_PHOTO_TAKEN)
```
moving-bits, line 162: same as above (`LogType.ATTENDED`)

moving-bits (changes requested):

moving-bits:
I'm not that deep into that area, but seems to work fine code-wise. Have just a few minor annotations, see above.

main/src/cgeo/geocaching/log/LoggingUI.java
ztNFny: I decided against it, even thought about not having it for webcam but then included it as on a cache tour including a webcam you might log the webcam the same as the caches - unlikely for the event.

main/res/values/strings.xml
ztNFny: The "template" part is the most important of the line, so that shouldn't be in brackets. I'll remove the space before the ellipsis.

main/src/cgeo/geocaching/log/LoggingUI.java
moving-bits: Agreed, but you might define different default templates for different cache types. (This was actually requested recently on support, paired with "use it automatically", but that's a different story.)

ztNFny: There's no association between log types and templates. They are just text.

moving-bits: I know (which would rule out any automatism), still the user could create their "Found it (Webcam)", "Found it (Event)" and "Found it (everything else)" default log templates (in whatever combination).

But anyway, no blocker for me. We can merge it and see whatever user feedback may come in.

ztNFny:
addressed the feedback and rebased on latest master

PR #12653, opened 2022-01-31
Improvement for wallpaper setting string

Lineflyer:
## Description

- As we don't use the term dashboard I suggest to name it home screen here, as home is what the bottom nav button is called.
- Add "device" to make clear its the device wallpaper being used

## Related issues

#11867


moving-bits (approved):
LGTM

Issue #12573, opened 2022-01-20
Change text from "Live map" to "Map" in the bottom navigation


bekuno:
### Tell us your idea!

In the last dev meeting it was decided to change the label of the map icon in the bottom navigation from Live map to Map. 
This would reflect the behaviour of the universal map better than the name "live map".

Lineflyer:
All existing users would probably ask why its no longer a "live map"....just thinking from users perspective.
Besides that, our app description explicitly features live map as outstanding feature.

Anyway, I could live with both :)

andrcoding:
From my user perspective, labeling it just as a 'map' would not be confusing. In my eyes, it is only a live map when updates are activated. (I use it very often just as a map with gc stored offline because of bad internet connection outside) Here you could add a small symbol on the map at some border to indicate the 'live modus' (e.g. red dot). Otherwise, I recognize the live map only by the fact that the bar at the top of the map moves when I look at another map section.

wolverine007:
> Otherwise, I recognize the live map only by the fact that the bar at the top of the map moves when I look at another map section.
> 

By reading your post, something related to that bar came to my mind:

Right now the bar is a small orange bar moving from left to right, which is hard to recognize on a normal OSM offline map:

![IMG_20220130_185852.jpg](https://user-images.githubusercontent.com/4508228/151711453-0c04b2f9-5129-477c-86c5-df193c0cdb36.jpg)

Maybe it is possible to move this bar a little bit higher, so it is displayed inside the black status bar for better visibilty?

But just an idea.

Overall changing from "Live map" to "Map" is fine for me 👍🏼

fm-sys:
Currently, the map actionbar will show "Live-Map" or "Map" (depending on whether live is enabled)

IMHO this isn't really obvious so I would suggest to use "Live-Map" or "Offline-Map" (depending on whether live is enabled)

Reason: It already happened to me multiple times, that I had live disabled and was wondering, why no caches are loaded.

Then it's also more consistent if we show "Map" as general label in the bottom navigation.

moving-bits:
We should include this before releasing the next version. I've created PR #12715 for it.

PR #12657, opened 2022-02-01
use material progress indicator for better visibility on map

fm-sys:
Use Material progress bar for better visibility, as suggested in https://github.com/cgeo/cgeo/issues/12573#issuecomment-1025196125

It has a 2x stroke width compared to the old one and is directly attached to the top instead of being floating inside the map.

Before:

Now: 
grafik

### DIFF
--- a/main/res/layout/map_progressbar.xml
+++ b/main/res/layout/map_progressbar.xml
@@ -1,7 +1,6 @@
-<ProgressBar
+<com.google.android.material.progressindicator.LinearProgressIndicator
 	xmlns:android="http://schemas.android.com/apk/res/android"
 	android:id="@+id/map_progressbar"
-	style="@style/map_progressbar"
 	android:layout_width="match_parent"
 	android:layout_height="wrap_content"
 	android:indeterminateOnly="true"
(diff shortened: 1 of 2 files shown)


main/res/layout/map_progressbar.xml
```diff
@@ -1,7 +1,6 @@
-<ProgressBar
+<com.google.android.material.progressindicator.LinearProgressIndicator
 	xmlns:android="http://schemas.android.com/apk/res/android"
 	android:id="@+id/map_progressbar"
-	style="@style/map_progressbar"
 	android:layout_width="match_parent"
 	android:layout_height="wrap_content"
 	android:indeterminateOnly="true"
```
Lineflyer, line 6: Not sure about the difference but the code examples always show `android:indeterminate="true"` instead.

Lineflyer: Found it: https://developer.android.com/reference/android/widget/ProgressBar#attr_android:indeterminateOnly
Looks appropriate.

Lineflyer (approved):

PR #12662, opened 2022-02-02
centerCrop wallpaper on home screen (fix #12654)

### LINKED ISSUE #12654: Home screen background has wrong aspect ratio
### Describe your problem!

On my device (Samsung S20, Android 12) the new "take over the wallpaper as background" feature is working, but the image has a wrong aspect ration on the home screen.

Looks like the original device wallpaper image does not fit the aspect ration of my device screen. 
Therefore the device shows only part of it (keeping the aspect ratio) but on our home screen it looks "zoomed to fit" ignoring the aspect ratio.

Can provide screenshots later if needed.

### How to reproduce?

- Have some custom wallpaper (e.g. own photo) as device wallpaper
- Activate the "take over the wallpaper as background"  feature in c:geo

Wallpaper image is shown but with wrong aspect ratio

Should look like on the device background

### Reproducible

Yes

### c:geo Version

2022.01.31-NB

### System information

```text
Samnsung S20, Android 12
```

### Additional Information

_No response_

moving-bits:
## Description
Instead of using the wallpaper image as background drawable add an `ImageView` to the bottom navigation activity, so that we are able to use `scaleType=centerCrop`, which centers the image and crops everything that got too large. It's probably still not 100% the same for some launchers, but comes pretty close to what the default launcher does with the wallpaper.

|Resulting image is like this:|instead of this:|
|---|---|
|image|image|

### DIFF
--- a/main/res/layout-h390dp-land/activity_bottom_navigation.xml
+++ b/main/res/layout-h390dp-land/activity_bottom_navigation.xml
@@ -5,6 +5,12 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+    <ImageView
+        android:id="@+id/background"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:scaleType="centerCrop" />
+
     <FrameLayout
         android:id="@+id/activity_content"
         android:layout_width="wrap_content"
--- a/main/res/layout-land/activity_bottom_navigation.xml
+++ b/main/res/layout-land/activity_bottom_navigation.xml
@@ -5,6 +5,12 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+    <ImageView
+        android:id="@+id/background"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:scaleType="centerCrop" />
+
     <FrameLayout
         android:id="@+id/activity_content"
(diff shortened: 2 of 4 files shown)


fm-sys (changes requested):
Please be aware, there is a second layout file involved for landscape... 

But as the background image is only used for MainActivity anyway, we should IMHO not add that view to the general bottom navigation layout but the the MainActivity layout

moving-bits:
> Please be aware, there is a second layout file involved for landscape...

Ah, good to know, thanks for mentioning it. => I'll adapt the PR.

> But as the background image is only used for MainActivity anyway, we should IMHO not add that view to the general bottom navigation layout but the the MainActivity layout

I had it this way in the first version of this PR, but then the image will centered within the dashboard's dimensions only, not centered relative to the whole display, which would lead to a very visible difference to the wallpaper look, as the dashboard's center is more upward than the display's center.

But having the `ImageView` defined in the base layout should be no problem, as it gets filled in `MainActivity` only. When switching through the other buttom navigation activities, there will be no wallpaper visible.

moving-bits:
Adapted PR to support landscape layouts as well:

image

bekuno:
retest this please

fm-sys (approved):
Should probably be fine...

Issue #12673, opened 2022-02-04
Log with attached image returns upload failure although successful


moving-bits:
### Describe your problem!

On support we have two new tickets stating, that on logging with an attached picture c:geo returns an error, even on several retries. But when they check online with their browser, the log got sent.

Both tickets are about logs with attached images, don't know yet whether the error message also occurs if no image is attached (but I sort of doubt it, otherwise we would have tons of support tickets).

Logcat does not show any related errors for them.

I tried to reproduce this from within the emulator, and I'm not able to send a log at all, I always get an "unknown error". Logcat lists this:
```
2022-02-04 21:45:15.791 17728-17975/cgeo.geocaching E/cgeo: [AsyncTask #2] Error posting log
    java.lang.RuntimeException: java.io.IOException: request was not successful: Response{protocol=http/1.1, code=400, message=Bad Request, url=https://www.geocaching.com/api/proxy/web/v1/geocache/gc12345/GeocacheLog}
        at io.reactivex.rxjava3.internal.util.ExceptionHelper.wrapOrThrow(ExceptionHelper.java:46)
        at io.reactivex.rxjava3.internal.observers.BlockingMultiObserver.blockingGet(BlockingMultiObserver.java:94)
        at io.reactivex.rxjava3.core.Single.blockingGet(Single.java:3645)
        at cgeo.geocaching.connector.gc.GCWebAPI.postLog(GCWebAPI.java:926)
        at cgeo.geocaching.connector.gc.GCLoggingManager.postLog(GCLoggingManager.java:137)
        at cgeo.geocaching.log.LogCacheActivity$Poster.doInBackgroundInternal(LogCacheActivity.java:687)
        at cgeo.geocaching.log.LogCacheActivity$Poster.doInBackgroundInternal(LogCacheActivity.java:671)
        at cgeo.geocaching.utils.AbstractAsyncTaskWithProgress.doInBackground(AbstractAsyncTaskWithProgress.java:134)
        at android.os.AsyncTask$3.call(AsyncTask.java:378)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:289)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
[...]

PR #12679, opened 2022-02-05
rel to #12673: fix sending of normal logs (without images) for emulator API 29

eddiemuc:
rel to #12673: fix sending of normal logs (without images) for emulator API 29


moving-bits (approved):
Works like a charm. Thanks for diving into this!

PR #12683, opened 2022-02-05
Download caches as non blocking service

fm-sys:
First alpha version for #330

From some quick tests in the emulator, I can tell that it already works quite good.

The only bug / intended feature (?) I've discovered is, that refreshed caches will only reflect the update if the foreground activity gets restarted.

However, I've decided to start as "nightly only" feature for now and have added some deterrent warning ;-)

grafik

|download running|download finished|
-|-
grafik|grafik

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -921,6 +921,8 @@
 
         <service android:name=".downloader.ReceiveDownloadService" />
 
+        <service android:name=".service.CacheDownloaderService" />
+
         <!-- Receiver for custom tabs share intent -->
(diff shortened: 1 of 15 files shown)


fm-sys:
A general question:
Are you OK with introducing experimental alpha features to our master branch, as long as they are decoupled from existing features and can easily be disabled/removed?

fm-sys:
One upvote and no one did complain, so I guess we can merge here... 

I'm really looking forward how this will work on a real device. This might be one of the most wished features, so let's hope we can reach a state where it's ready for production :)

PR #12686, opened 2022-02-06
update internal BRouter engine to v1.6.3 (fix #12412)

### LINKED ISSUE #12412: Update Brouter sources
### Tell us your idea!

Seing that brouter was updated and looking at the [recent commit history](https://github.com/abrensch/brouter/commits/master) I guess we could/should also update our implementation accordingly.

@moving-bits 
Not sure how this could work..is there any versioning/dependency defined to check if updates are needed or was this pure manual work at that time?

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Update internal BRouter engine to v1.6.3 (the newest released version of BRouter)


moving-bits:
Removing draft state to get feedback from CI, but setting "WIP" label for now 

(I expect quite a few messages from CI, as coding styles are pretty different between those projects, and I'm sure I haven't caught all differences)

PR #12689, opened 2022-02-06
catch OutOfMemory exception on GM.drawHistory

moving-bits:
## Description
Some users have set their track history limit too high, so that they run into `OutOfMemory` exceptions in GM's `drawHistory`, as shown on Play Console stats (currently 260 occurrences for 27 users). This PR adds a try/catch block around it and logs the error to our log - better not draw a history than crash the map.


Lineflyer:
While catching such exception is meaningful IMHO, does it make sense to reduce the limit in general?

fm-sys:
It's device specific. Some low budget devices might crash, while it works like a charm on better devices. Furthermore, the length AFAIK can already be configured in the settings...

moving-bits:
Exactly as @fm-sys described: We have no hard-coded limit, user can set it to their choice. This PR adds catching the exception and puts an error message to the log, so that we can advice the user how to continue if they ask why their history line is no longer being displayed under certain conditions.

PR #12692, opened 2022-02-06
Show history on long-click of "search" (fix #12687)

### LINKED ISSUE #12687: Add "recently viewed caches" PopUp to Bottom-nav (longClick)
### Tell us your idea!

While doing a round of caches with for instance a bonus cache I am storing the gathered bonus information from the other caches directly inside the note section of the bonus cache.

Right now I have to scroll back to the bonus icon on the map, therefore I have to remember where the icon is stored, open the listing, go to desired cache section, click edit, etc pp

Afterwards I have to jump back to my current location, select the next target on the round, resulting in an endless loop of repeating the above.

My current idea was to add a long-press function to the Map (Live-Map) button which is showing for instance the last 5 visited caches, so you could easily jump into these caches detailed information.
Maybe it is also possible to remember which tab and area inside these historical caches was viewed before.

In general this function might improve the usage of the Map, when you for instance want to compare two different caches to decide which one will be your next target.

Maybe another solution is to add a new button inside the bottom navigation which is stating a list of all formerly visited caches (e.g. like the history shown in the search but much longer 100 entries maybe?) and adding here as well the above described long-press function to directly access the last 5 viewed caches. 

### Is there an alternative?

yes, but not user friendly, see above... 

### Additional Information

_No response_

fm-sys:
fix #12687

grafik

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1423,6 +1423,7 @@
     <string name="cache_share_field">Share</string>
     <string name="cache_time_full_hours">o\'clock</string>
     <string name="cache_listed_on">Listed on %s</string>
+    <string name="cache_recently_viewed">Recently viewed</string>
 
 
     <string name="cache_whereyougo_start">Open WhereYouGo to download and start this cartridge</string>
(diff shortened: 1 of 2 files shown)


SammysHP:
Do we have a better string for "select item", like "history" maybe?

ztNFny:
Could you change the title to something like "Recently viewed"?

fm-sys:
I've decided to go with "Recently viewed", as we already use the phrase "history" for our log history...

grafik

Issue #12626, opened 2022-01-29
Use separate waypoint type for generated waypoints


moving-bits:
### Tell us your idea!

A follow-up to #12384/#12588, as discussed briefly in https://github.com/cgeo/cgeo/pull/12505#issuecomment-1008110945:

Use a separate waypoint type for the generated waypoints, which would allow for both
- using a separate icon for those waypoints (to detect them more easily on the map),
- a function to "delete all generated waypoints" on the waypoints tab.

### What solution would you suggest?

- use a separate waypoint type for generated waypoints
- implement a new icon for this waypoint type
- add a function "delete all generated waypoints" to the waypoints tab

[Edit]
### TODO List
- [x] Create separate waypoint type with icon and compact mode icon - #12693
- [x] Add item to cache details to remove all generated waypoints - #12734
- [ ] Store reference to originating waypoint for all generated waypoints
- [ ] Add item to waypoint popup for generated waypoints "delete all from this series"
- [ ] Add item to waypoint popup for originating waypoints "delete all from this series"
- [ ] Add button "delete all generated for this waypoint" to calculate dialog above "generate range items"

moving-bits:
I could start working on the technical aspects of this issue, but this would only make sense if someone implements a new drawable.

@ztNFny 
Would you be willing to create a drawable for this new waypoint type?

ztNFny:
> Would you be willing to create a drawable for this new waypoint type?

Can do. How about the icon currently used to mark a wp as calculated? The 4 icons

eddiemuc:
The challenge with this issue is IMHO not so much the introduction of a new waypoint type (this is actually very easy) but to come up with a good UI concept to allow easier editing (or in this case: "mass-deletion") of selected waypoints (in this case: all wps of a specific type)

Lineflyer:
If such new type is introduced, would I still be able to use one of the other types for such a calculated waypoint?

Because I would not like to lose the possibility to define those as "Stage", "Trailhead", "Final" as required by the cache.

ztNFny:
What would be the concept for this new waypoint type? I sometimes use "coordinate joker" with Locus which adds possible locations to the map with their own icon - for that it'd be useful.
Today I did a multi with the new calculator, set up a waypoint of type final with the formula from listing. For that one I do not want to have some new waypoint type but it must be final - because that's what it is.

eddiemuc:
[...]

PR #12693, opened 2022-02-06
introduce new wapoint type for generated waypoints (rel. to #12626)

moving-bits:
## Description
Stores generated waypoints using a new waypoint type.
For screenshots see https://github.com/cgeo/cgeo/issues/12626#issuecomment-1030830136

Setting "WIP" for now, as some waypoint handling features should be added. For a proposal see https://github.com/cgeo/cgeo/issues/12626#issuecomment-1030825175

### DIFF
--- a/main/res/drawable/dot_waypoint_generated.xml
+++ b/main/res/drawable/dot_waypoint_generated.xml
@@ -1,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="12dp"
+    android:height="12dp"
+    android:viewportWidth="12"
+    android:viewportHeight="12">
+    <path
+        android:fillColor="#000000"
+        android:pathData="M3,3.5h6v1.5h-6z"/>
+    <path
+        android:fillColor="#000000"
+        android:pathData="M3,7h6v1.5h-6z"/>
+</vector>
--- a/main/res/drawable/waypoint_generated.xml
+++ b/main/res/drawable/waypoint_generated.xml
@@ -1,0 +1,22 @@
+<!-- based on material.io / calculate -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="36dp"
+    android:height="36dp"
+    android:viewportWidth="30"
+    android:viewportHeight="30">
+  <path
+      android:fillColor="#80000000"
+      android:pathData="M9.25,10.72h5v1.5h-5z"/>
+  <path
+      android:fillColor="#80000000"
+      android:pathData="M16,18.75h5v1.5h-5z"/>
+  <path
+      android:fillColor="#80000000"
+      android:pathData="M16,16.25h5v1.5h-5z"/>
+  <path
+      android:fillColor="#80000000"
+      android:pathData="M11,21l1.5,0l0,-2l2,0l0,-1.5l-2,0l0,-2l-1.5,0l0,2l-2,0l0,1.5l2,0z"/>
+  <path
+      android:fillColor="#80000000"
+      android:pathData="M17.09,13.95l1.41,-1.41l1.41,1.41l1.06,-1.06l-1.41,-1.42l1.41,-1.41l-1.06,-1.06l-1.41,1.41l-1.41,-1.41l-1.06,1.06l1.41,1.41l-1.41,1.42z"/>
+</vector>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -88,6 +88,7 @@
     <string name="wp_trailhead">Trailhead</string>
     <string name="wp_waypoint">Reference Point</string>
     <string name="wp_original">Original Coordinates</string>
+    <string name="wp_generated">Generated</string>
 
     <!-- waypoints short names for new waypoints -->
     <string name="wpnew_final">Final</string>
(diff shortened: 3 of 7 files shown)


ztNFny:
@moving-bits I accidentally committed my icon changes against your branch. if you mind them just revert, I have them in a branch of my own. Will test-drive them in my build for a few days and adjust them based on that first impression. If you have feedback to share please be welcome.

moving-bits:
> @moving-bits I accidentally committed my icon changes against your branch. if you mind them just revert, I have them in a branch of my own. 

No problem. As it looks that I need to fix some CI findings anyway, and as you still have your files in your own branch, I would simply update my PR and force-push.

> Will test-drive them in my build for a few days and adjust them based on that first impression. If you have feedback to share please be welcome.

Thanks for working on it. I'll provide screenshots to make it easier for the team to follow the path being pursued:

image.image

Some quick thought on those icons:
- pro: crisp & clear in both standard and compact mode
- con: I would associate the icon more with "has a formula" than with "has been generated"

I guess the latter is more relevant for the standard-sized icons. Maybe something like "[1-3]" would fit for those (using square brackets and a range of numbers)?

fm-sys:
> Maybe something like "[1-3]" would fit for those (using square brackets and a range of numbers)?

Maybe it would also be enough to just use square brackets instead of normal ones...? `[x]`

ztNFny:
> Maybe something like "[1-3]" would fit for those (using square brackets and a range of numbers)?

Too much content for the small space available. Height would be too small to recognize anything

moving-bits:
> > Maybe something like "[1-3]" would fit for those (using square brackets and a range of numbers)?
> 
> Maybe it would also be enough to just use square brackets instead of normal ones...? `[x]`

still think it's too close with the symbol for "has formula":

image

In this screenshot, only the last two waypoints do have a formula, whereas all waypoints above are generated (and have fixed coordinates, no formula). When using "[x]" as symbol for a waypoint this would look pretty confusing, IMHO, as the generated waypoints would have "[x]" as icon symbol, and the last two waypoints would have a "(x)" as symbol directly beneath the icon.

I would go for something more distinct.

moving-bits:
> > Maybe something like "[1-3]" would fit for those (using square brackets and a range of numbers)?
> 
> Too much content for the small space available. Height would be too small to recognize anything

Ok, understandable. I had feared something like this.

Hm, any other idea to make it more distinct from the other "(x)" symbol?

ztNFny:
But "has formula" and "generated" are closely related. You can only generate from a formula, so having that icon seems like an obvious choice to me.
Also the (x) icon indicates "variable", which is what the waypoint represents, the result of "testing" a variable.

The other idea I had was something like a joker in a card game but couldn't come up with a simple enough icon for that.

ztNFny:
I had tested a star like this, liked the "(x)" better, but I can show you other options on sunday
image

moving-bits:
Browsing material font, maybe some of those would fit?

image.image.image.image.image.image


Or maybe just a sequence of numbers, as here (but without the extra border):
image

Or maybe just the generic "waypoint" marker (two circles), but in a different color?
e. g.: 
image

ztNFny:
> Browsing material font, maybe some of those would fit?

None of them say "joker", "guess" or "variable" to me. The icon should have some association with the formula it comes from so you can identify it as such.

> Or maybe just the generic "waypoint" marker (two circles), but in a different color?

Just changing the color is bad UX and ignoring vision-impaired users (and also normal ones with bad visibility conditions, e.g. direct sunlight).

ztNFny:
The icon represents "the result of a formula with some variable(s) guessed". To me this offers the following keywords an icon could be selected from:
- result
- formula
- variable
- guess

moving-bits:
- "result" could be associated with the "=" sign, so maybe `=` or `[=]` might work
- or we could use "?" as the universal placeholder symbol, e. g. `[?]`, `1?7`, `.?.`, `•?•` or just plain `?`

fm-sys:
> [?]

I like this one most. At least it seem to be the most logical one... :)

ztNFny:
> > [?]
> 
> I like this one most. At least it seem to be the most logical one... :)

Similar to mystery though?

moving-bits:
> > > [?]
> > 
> > 
> > I like this one most. At least it seem to be the most logical one... :)
> 
> Similar to mystery though?

We would have a white background with colored symbol (inverted colors compared to mystery caches), and we have added brackets, so they should be distinguishable enough.

Alternatively we still have the first two proposals in https://github.com/cgeo/cgeo/issues/12626#issuecomment-1030677669

moving-bits:
> The new type should not appear when creating a waypoint, just noticed this in yesterday's build.

Meaning: The new type should never appear in the waypoint type selector, and waypoints of type "generated" should have no type selector?

moving-bits:
Hm, thinking about it, let me fine-tune this a bit:

The new type should never appear in the waypoint type selector, _if current waypoint type isn't "generated". Waypoints of type "generated" should have a waypoint type selector, to be able to convert such a waypoint to (e. g.) "Final"._

moving-bits:
Added variable display of type `GENERATED`:

|editing waypoint of type GENERATED|editing waypoint of different type|adding a new waypoint|
|---|---|---|
|image|image|image|

Issue #12456, opened 2022-01-01
high contrast Google maps theme


ztNFny:
### Tell us your idea!

Google maps supports custom themes: https://developers.google.com/maps/documentation/android-sdk/styling

We should provide a (optional) high-contrast theme to make using the map easier in common caching scenarios.

### Additional Information

While offline maps are certainly the better choice in some cases Google maps has to be used. Sadly Google had been decreasing the contrast of maps over the years - if you look at a small path in a wood that will be a thin white line on a light-grey background, almost impossible to see, especially when the sun is shining bright on your display.

![Screenshot_20220102-004549](https://user-images.githubusercontent.com/1258173/147862457-d08db43d-dcb2-42ea-b944-8743b7405558.jpg)
Spot the path

And just that situation is when i need Google maps - I'm in the woods, OSM doesn't have the path but Google does - but the path is hardly visible there.

ztNFny:
Preview of Google Maps styles in c:geo:

![image](https://user-images.githubusercontent.com/1258173/153056701-09e79be6-e738-4824-b2b9-df5408c73827.png)
![image](https://user-images.githubusercontent.com/1258173/153056843-5b732055-ad6e-42d4-b1db-9fe9db5840b2.png)

ztNFny:
I'm working on general style support first, will try to create a geocaching-optimized version of the default Google Maps layout later on then

fm-sys:
Already looks great 👍

PR #12705, opened 2022-02-08
Google Maps themes support

ztNFny:
implement theme support for Google maps with json themes
#12456 

- [x] general implemenation
- [x] final decision about themes to include
- [x] System mode theme switch (day/night)

image
image
image
image


ztNFny:
feedback on the implementation in GoogleMapView.java would be welcome.
I don't really like the static stuff, but as I have to call it from https://github.com/cgeo/cgeo/pull/12705/files#diff-a25a47bd36a33f4384c48602f15a543f9e344ab38781eb71485c8102a4330c60R828 I didn't find any good way without it.
Suggestions welcome! @eddiemuc @fm-sys

eddiemuc:
> feedback on the implementation in GoogleMapView.java would be welcome. I don't really like the static stuff, but as I have to call it from https://github.com/cgeo/cgeo/pull/12705/files#diff-a25a47bd36a33f4384c48602f15a543f9e344ab38781eb71485c8102a4330c60R828 I didn't find any good way without it. Suggestions welcome! @eddiemuc @fm-sys

Hi @ztNFny the link does not really work for me (at least it does not lead me to a specific line), but I assume you mean that you need to call the selectMapTheme method of class GoogleMapView from CgeoMap -> correct?

In that case, you can do the following to prevent the usage of static var:
* create the method "selectMapTheme" also in class `MapViewImpl`, make is abstract or leave it empty there. Make that method non-static of course. `GoogleMapView` extends class `MapViewImpl`, so it will implement this method then (if you make it non-static again).
* `CGeoMap` extends class `AbstractMap` which has access to the used `MapViewImpl` instance (stored in member `mapView`. So in `CGeoMap` you can then call the method using `this.mapView.selectMapTheme()`.

If you think that this class structure is unnecessarily complex I would definitely agree. I think this has historic reasons. One of these days we might want to clean this up.

ztNFny:
> In that case, you can do the following to prevent the usage of static var:

Thanks, that's exactly the advise I was looking for!

ztNFny:
non-static method and dynamic theme switching based on device theme added :-)

This should be ready to go code-wise. The themes to include I'd like to discuss during our next team meeting, leaving it in WIP until then.

@Lineflyer one item for sundays agenda :-)

PR #12709, opened 2022-02-09
display routes/tracks button if target is set (fix #12701)

### LINKED ISSUE #12701: Menu entry "Clear manual targets" is missing
### Describe your problem!

Ok, admittedly, it's not completely missing, but hard to find sometimes. ;)

### How to reproduce?

1. open any cache listing and route to the cache, e. g. by clicking the compass rose in the header
2. when in the cache's map, ``Enabled live``
3. now try to get rid of the route ...

As there is no ``Clear manual targets`` in the menu anymore (IIRC it was there a while ago),
*and* there is no bottom menu for routes (as there is no manual route right now),
you first have to click ``Routes / Tracks``, then the bottom route menu pops up,
and finally you will find the ``Clear manual targets`` button.

Either the bottom route menu button should be active as soon as there is *any* route active
*or* the ``Clear manual targets`` should be there as soon as there is a manual target active ...

### Reproducible

Yes

### c:geo Version

2022.02.07-NB-a93e0da

### System information

_No response_

### Additional Information

I would call myself an experienced user, but even me had a hard time to find a way to get rid of the manual target
and I would assume that "normal" users will just get frustrated ...

moving-bits:
## Description
Enables the "routes/tracks" quick button when a navigation target is set, which gives you a quick access to the "clear manual targets" option.
Also fixes a hard-coded label for the "Routes / Tracks" menu item.

## Additional context
The whole `mapOptions` and setting a target is pretty confusing to me. In the end I basically added something like 
```
final Geocache temp = getCurrentTargetCache();
if (temp != null) {
    lastNavTarget = temp.getCoords();
}
```
to both map implementations (under certain conditions).
A review / some testing would be welcomed to check whether this has unwanted side-effects.

### DIFF
--- a/main/res/menu/map_activity.xml
+++ b/main/res/menu/map_activity.xml
@@ -146,7 +146,7 @@
     </item>
     <item
         android:id="@+id/menu_routetrack"
-        android:title="Routes / Tracks" />
+        android:title="@string/routes_tracks_menu" />
     <item
         android:id="@+id/menu_check_routingdata"
         android:title="@string/check_tiles_menu" />
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -2258,6 +2258,7 @@
 
     <!-- route/track commons -->
     <string name="routes_tracks_dialog_title">Routes / tracks options</string>
+    <string name="routes_tracks_menu">Routes / Tracks</string>
     <string name="center_on_route_track">Center on route/track</string>
     <string name="unload_route_track">Unload route/track</string>
     <string name="route">Route</string>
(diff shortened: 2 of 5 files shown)


moving-bits:
@Lineflyer

I've compared this fix to the code we had in the last release, before unification of track / individual route settings. The `isTargetSet.call()` is equivalent to what we had before, and the `... lastNavTarget = temp.getCoords(); ...` fixes an already existing error that the menu entry was not shown under a couple of conditions (and both together fix the originating issue, #12701).

So my proposal would be
- to merge this PR,
- create a new beta from then-state of `release`, and 
- run that for a couple of days.

If everything runs smooth, we have a new release in time for the website changes, and if not, we can revert this PR, and create a release from that state.

Lineflyer:
Lets include it anyway.

PR #12713, opened 2022-02-10
Fix string: Auto backup is no longer a reminder

Lineflyer:
## Description

AFAIR the automatic backup feature does no longer work with reminders on the dashboard but is just performed.
If this is correct, then there is a left over: Wrong caption of the corresponding setting, which is fixed with this PR.

## Related issues

None


Lineflyer:
@moving-bits Is my assumption above correct, then please merge.

moving-bits (approved):
Assumption is correct, thank for catching this glitch

PR #12715, opened 2022-02-12
change map labels (fix #12573)

### LINKED ISSUE #12573: Change text from "Live map" to "Map" in the bottom navigation

moving-bits:
## Description
As described in #12573:
- change bottom navigation label to "Map" (instead of "Live map") (same with shortcuts)
- change map title from "Map" and "Live map" to "Offline map" and "Live map"

image


main/res/values/strings.xml
```diff
@@ -349,7 +349,7 @@
 
     <!-- main screen -->
     <string name="home_button">Home</string>
-    <string name="live_map_button">Live map</string>
+    <string name="live_map_button">Map</string>
```
fm-sys, line 352: Maybe it's worth to use the already existing resource `R.string.map_map` so that we don't have to wait for translations... Cause it would be bad if the bottom navigation isn't localized for our release.

moving-bits: adapted PR accordingly

moving-bits: Me neither, just tried it out and found it working. Helps with having a consisting naming scheme of string resources, `xxx_button` in this case for all home screen buttons, without having to duplicate all texts. Don't know what this means to Crowdin, though - we'll probably just have to try it out...

fm-sys: > Don't know what this means to Crowdin, though - we'll probably just have to try it out...

That might indeed be problematic. Probably, we have to mark the string as non-translatable or translators will wonder how to translate that cryptic string. 😅

moving-bits: Not sure, what _that_ would mean to c:geo-internal resource handling for translations... 
I propose to merge this PR, and check Crowdin timely, so that we would have the chance to resolve any problems that may arise. Crowdin has quite some language knowledge, so I wouldn't wonder if it would detect this string as being a reference, and not offer it for translation.

Lineflyer: I will merge and check Crowdin in some minutes.
AFAIR I can also set a string to hidden manually on Crowdin as an easy solution.

fm-sys:
> FYI: Crowdin is not clever enough for this

Ok, what does this mean for us? Do we want to change the string (on `master`) to an actual string or do we just keep it like this?

fm-sys:
> > FYI: Crowdin is not clever enough for this
> 
> Ok, what does this mean for us? Do we want to change the string (on `master`) to an actual string or do we just keep it like this?

Ah, it's already discussed in  #12725 😅

PR #12717, opened 2022-02-12
move routing mode settings to routes/tracks quick menu (fix #12716)

### LINKED ISSUE #12716: Consolidate routing options into "Route/Track" quick menu
### Tell us your idea!

Currently some route-related options are still in the "old" map quick settings menu, whereas most route-related options are in the route/track quick settings menu:

image.image

"Set route as target" is actually in both menus...

- Move routing mode options to the Route/Track quick menu
- Remove routing mode options and "set route as target" options from map quick settings menu

### Is there an alternative?

leave it as is

### Additional Information

_No response_

moving-bits:
## Description
Moves "routing mode" settings from map quick settings menu to routes/tracks quick settings menu, and removes the duplicated "set route start as target" from map quick settings (as they are already present in routes/tracks quick settings).

image.image


ztNFny:
Noooo 😟

We have 2 very different routing things in cgeo: "ad hoc" draw-a-line to the selwcted cache (this) and "planned" for a while tour. Imho they shouldn't be mixed.

Lineflyer:
> We have 2 very different routing things in cgeo: "ad hoc" draw-a-line to the selwcted cache (this) and "planned" for a while tour. Imho they shouldn't be mixed.

I also thought about that while reading the PR.

On one hand its always to avoid to change the location of existing feature controls, on the other hand it would be really consistent to have all routing stuff on the routing icon.
I was already wondering why the "route" button shows up on a single cache map with routing line to the cache active, while it contains no related option. This PR would make it fit.

I am a bit undecided right now...

ztNFny:
> I was already wondering why the "route" button shows up on a single cache map with routing line to the cache active,

Thats a quite recent change and in its current form a bug imo. Either that should be fixed or this PR merged (but im against the later)

moving-bits:
> > I was already wondering why the "route" button shows up on a single cache map with routing line to the cache active,
> 
> Thats a quite recent change and in its current form a bug imo.

That's for the "clear manual targets" as requested in #12701

moving-bits:
> We have 2 very different routing things in cgeo: "ad hoc" draw-a-line to the selwcted cache (this) and "planned" for a while tour. Imho they shouldn't be mixed.

Routing mode setting affects both, so there's no clear distinction by this. IMHO having all routing related options in the same quick button menu is more consistent.

moving-bits:
How to proceed here? 
Consolidate all routing options into the routing quick menu, or leave routing mode settings in map quick settings dialog? 

Let's use our voting system:
- use 👍 for merging this PR, consolidating all routing related options into the routing quick settings (right icon)
- use 👎 for not merging, keeping routing mode settings in the map quick settings (left icon)

ztNFny:
> use 👎 for not merging, keeping routing mode settings in the map quick settings (left icon)

That would should also mean "dont show the routing float button if the routing stuff isn't used", ie we have to find a solution for the "clear target" thing.

moving-bits:
> > use 👎 for not merging, keeping routing mode settings in the map quick settings (left icon)
> 
> That would should also mean "dont show the routing float button if the routing stuff isn't used", ie we have to find a solution for the "clear target" thing.

Not necessarily, but let's do this step by step, depending on the votes let's see what the consequences are/could be.

bekuno:
I would like to have a single place for routing.
But - in the screenshot above, I cannot find the distinction between automatic route and planned route.
How does the "Routing mode" there affect the both modi?
How can I easily deactivate the automatic route but set the routing mode for the automatic route?

moving-bits:
> in the screenshot above, I cannot find the distinction between automatic route and planned route. How does the "Routing mode" there affect the both modi?

Routing mode affects the calculated route, e. g. if you select "walk" you may get a route including small footpaths, whereas when you select "car" you will get a route only including ways suitable for cars. This applies to both routes from current location to a cache/waypoint (I guess that's what you are meaning by "automatic route") as well as planned routes.

> How can I easily deactivate the automatic route but set the routing mode for the automatic route?

Use the "clear manual targets" button from the route quick settings dialog.

moving-bits:
@cgeo/team: Somehow this issue seems to have slipped through. I would really like to get your feedback on this.

For those that have not voted yet, please do so in https://github.com/cgeo/cgeo/pull/12717#issuecomment-1046091513.

For those that have downvoted, I would like to get your feedback on why you downvoted - what is triggering your vote, and what should be changed in your opinion.

Thanks.

Lineflyer:
I mainly downvoted, because its a user-affecting change without functional need.
- Users will potentially not easily find the new spot 
- User guide needs change
- A high percentage of users might not use advanced routing but only the target line, thus they do not need the rest of this bottom sheet (despite that already the "clear manual targets" has been moved there

On the other hand I understand the background of cleaning up and having it at the right place. But do not underestimate, that things will get more complicate for users and not all users are such "tech safe".

ztNFny:
> what is triggering your vote, and what should be changed in your opinion.

I only use the target line, none of the "routing" functionality. In terms of UI I think it's a bad decision if it's moved to the new dialog I get a) an additional floating button on the map (that I think is quite bad in terms of UI, but 2 of them is worse than one) and b) a quite complex dialog (counting 10 buttons in there, before the addition of route mode) to access a simple setting.
And in general terms, imho the settings don't belong so clearly together: "We have 2 very different routing things in cgeo: "ad hoc" draw-a-line to the selected cache (this) and "planned" for a while tour. Imho they shouldn't be mixed." - Yes, when using the advanced routing stuff then route mode fits right in there, but when only using the draw-a-line mode it doesn't have any relation to the other routing settings in there.

moving-bits:
In today's dev meeting we found no clear preference when considering both above mentioned opinions and those from today's participants, so we decided to postpone this PR, and put it on the agenda for the next dev call. Therefore setting "WIP" label for now.

moving-bits:
Seems we have no clear opinion in favor of this PR as well as no real complaints about the current implementation, so let's close this PR.

PR #12719, opened 2022-02-12
fix #12718: dont cut off var tab buttons

### LINKED ISSUE #12718: [Variables] Buttons are cut off
### Describe your problem!

A picture tells more than thousand words :) 

Screenshot_20220212_101712_cgeo geocaching

### How to reproduce?

Use a small screen or large font size... 

Buttons are cut off. 

The buttons should be ordered dynamically. This can easily be done with a AndroidX `CoordinatorLayout`. 

### Reproducible

Yes

### c:geo Version

NB2022.02.12

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #12718: dont cut off var tab buttons

### DIFF
--- a/main/res/layout/cachedetail_variables_page.xml
+++ b/main/res/layout/cachedetail_variables_page.xml
@@ -26,15 +26,16 @@
             android:text="@string/variables_titletext"
             />
 
-        <LinearLayout
+        <com.google.android.material.chip.ChipGroup
             android:layout_width="fill_parent"
             android:layout_height="wrap_content"
-            android:orientation="horizontal">
+            android:orientation="horizontal"
+            android:layout_marginHorizontal="4sp">
 
             <Button
                 android:id="@+id/variables_add"
                 style="@style/button"
-                android:layout_margin="6dp"
+                android:layout_margin="3dp"
                 android:minWidth="2dp"
                 app:icon="@drawable/ic_menu_add" />
             <Button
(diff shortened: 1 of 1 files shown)


eddiemuc:
> Why not put everything in one `ChipGroup` instead of two? Would save some screen space when thinking about larger devices (e.g. tablets) 

I wanted to separate the "add" actions from the "other" actions. If you don't think this is meaningful (?) then one chipgroup would be enough indeed. What do you think?

fm-sys:
Hm... As all buttons are related to adding/deleting variables I guess this separation is IMHO not really necessary.

This is how it looks when I turn my mobile to landscape: I feel like the space could be used better ;-) 

Screenshot_20220212_151837_cgeo geocaching_copy_585x270

eddiemuc:
OK, PR was changed to one ChipGroup only. Looks now like this in portrait:

image

And like this in landscape:

image

fm-sys:
The margin looks quite large currently. If we reduce it a bit, all buttons will probably fit into one line. But this is device specific anyway...

moving-bits:
Funny that it uses different horizontal spacing between the buttons (see first line vs. second line in portrait mode). Any way to influence this?

eddiemuc:
> Funny that it uses different horizontal spacing between the buttons (see first line vs. second line in portrait mode). Any way to influence this?

Your're right, margin definition is different for the delete button than for all others. I will fix this

eddiemuc:
> The margin looks quite large currently. If we reduce it a bit, all buttons will probably fit into one line. But this is device specific anyway...

margin is set to 6dp right now. Changing it to 3dp gives us this:

image

Issue #12725, opened 2022-02-12
String live_map_button reference not working


Lineflyer:
Crowdin is not clever enough for this, so I set the string to be hidden from translation:
![image](https://user-images.githubusercontent.com/949669/153710007-4764feb4-d80c-49aa-a682-6d4a75ae616d.png)

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/12715#issuecomment-1037172779_

What I did not take into account:
Hiding this string from translation will remove it in the translated strings, which means it will fall back to the english string for all languages. 
You can see this in PR https://github.com/cgeo/cgeo/pull/12723

@moving-bits: Seems we need to refrain from using such linkage.

Lineflyer:
Best solution seems to be to implement to directly use `map_map` for `release` and clean it up on `master` again to use  `live_map_button` with new content `Map`.

fm-sys:
Please take a look at #12726 as quick-fix...

moving-bits:
> What I did not take into account: Hiding this string from translation will remove it in the translated strings, which means it will fall back to the english string for all languages.

Yes, it takes the English definition for `live_map_button`, recognizes it as a reference to `map_map`, and takes the _translated_ version of that resource. Isn't that exactly what is intended here? (Also see https://github.com/cgeo/cgeo/pull/12722#issuecomment-1037369655)

moving-bits:
> I would still like to avoid such references also to avoid manual steps on Crowdin.

Ok, understandable

> We could keep in the reference on release (as it works) and clean up on `master`. How about that?

Either way would be ok for me. If we have the time for outstanding translations on `release` (which we probably should calculate anyway, as there is a new string "Offline map"), then it does probably make sense to do it the same way on both branches. We could take #12726 and just remove renaming of shortcuts, for example.

PR #12726, opened 2022-02-12
avoid usage of string references

fm-sys:
rel. to #12725
- Use `Map` for bottom navigation
- Continue using the old `Live-Map` label for launcher shortcuts (is this the wanted behavior? IMHO "live" makes it sound more like c:geo, so I would prefer this on as launcher shortcut label. Last but not least to avoid unnecessary confusion for our users ;-))

@Lineflyer

### DIFF
--- a/main/res/menu/bottom_navigation_menu.xml
+++ b/main/res/menu/bottom_navigation_menu.xml
@@ -15,7 +15,7 @@
         android:id="@+id/page_map"
         android:enabled="true"
         android:icon="@drawable/ic_menu_mapmode"
-        android:title="@string/live_map_button" />
+        android:title="@string/map_map" />
     <item
         android:id="@+id/page_search"
         android:enabled="true"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -349,7 +349,6 @@
 
     <!-- main screen -->
     <string name="home_button">Home</string>
-    <string name="live_map_button">@string/map_map</string>
     <string name="caches_nearby_button">Nearby</string>
     <string name="advanced_search_button">Search</string>
     <string name="stored_caches_button">Stored</string>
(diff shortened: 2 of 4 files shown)


moving-bits:
Not sure if we really need this PR:

> rel. to #12725
> Use `Map` for bottom navigation

Seems to work as intended, see https://github.com/cgeo/cgeo/pull/12722#issuecomment-1037369655
So using a reference (in combination with excluding from Crowdin) seems to be exactly what we are looking for in such cases.

> Continue using the old `Live-Map` label for launcher shortcuts (is this the wanted behavior? IMHO "live" makes it sound more like c:geo, so I would prefer this on as launcher shortcut label. Last but not least to avoid unnecessary confusion for our users ;-))

I tend to disagree here: Why would we still call the shortcut "Live map", when it's calling a map (which can be either live or offline map), and when the bottom navigation label has been renamed to "Map" to reflect this. Still calling the shortcuts "Live map" seems inconsistent to me.
And regarding the argument "avoid unnecessary confusion for our users": It's a change, yes, but IMHO the old version ("Live map") was the confusing one, not the new one. Also the new one is consistent to the bottom navigation label, and therefore less confusing.

Lineflyer:
See: https://github.com/cgeo/cgeo/issues/12725#issuecomment-1037392908

Lineflyer:
>  tend to disagree here: Why would we still call the shortcut "Live map", when it's calling a map (which can be either live or offline map), and when the bottom navigation label has been renamed to "Map" to reflect this. Still calling the shortcuts "Live map" seems inconsistent to me.

I would also agree to use the same labels on bottom-nav as also on the corresponding launcher shortcuts.

moving-bits:
Then we should keep removal of the reference (as discussed in https://github.com/cgeo/cgeo/issues/12725#issuecomment-1037413083), but set the two values for the shortcuts to `map_map` instead of `map_live`.

Issue #12380, opened 2021-12-23
[Global Calculated Coordinates/Waypoints] Remove old code


eddiemuc:
After switching to new global waypoints is done, old local coordinate codebase can be removed (excluding parts which are needed e.g. to read old Personal Notes and to migrate).

This should not be done before the new implementation was out in the field for a while to prove it is stable.

Following classes/files are candidates for removal:
```
cgeo.geocaching.ui.EditButton
cgeo.geocaching.ui.CalculateButton
cgeo.geocaching.ui.CalculatorVariable
cgeo.geocaching.models.CalcState
cgeo.geocaching.ui.dialog.CoordinatesCalculateDialog
coordinatescalculate_dialog.xml (both layout variants)
coordinates_calculate.xml
coordinate_equation.xml
strings.xml: values marked with "legacy"
```

eddiemuc:
Took a look at this one today (removal of old Calculated Coordinate code). 

There is a large bunch of legacy code around which is still needed to parse the "old" format for calculated coordinates stored in personal note (the one with the "F-PLAIN"-marker created by @murggel  back in the days). If I remove this, then coordinates stored in personal note with this format can't be parsed again by c:geo. New calculated coordinates will be stored in a different format.

New format (used as of now): 
```
@Markierung 1 (W) {CC|N53 33.AAA|E009 59.BCD} "example_1"
$A=1 | $B=2 | $C=3 | $D=4
```
Old format (using `F-PLAIN`, no longer parseable if legacy code is removed): 
`@Markierung 1 (W) (F-PLAIN) N 53° 33.AAA E 009° 59.BCD' |A=1|B=2|C=3|D=4| "example_1"`

Question: (when) can we remove the parsing of F-PLAIN coordinates out or personal note?

eddiemuc:
Note for self: most classes of package calculator can be removed if F-PLAIN does not need to be supported:
![image](https://user-images.githubusercontent.com/6909759/170819854-d40c5f4f-58ee-43c6-b8f6-30a646eea25d.png)

Exceptions/Additions:
* CalculatedCoordinateMigrator needs to stay
* FormulaParser might need adaptions (is also used for "current" format parsing
* WaypointParser needs adaptions (might still use FormulaParser or something else)

murggel:
Probably we can make an announcement, that the migration of the F-Plain format will be removed in one of the next versions, so those who have still the format in some PN can do the migration before it will be removed.

eddiemuc:
[...]

PR #12742, opened 2022-02-14
fix #12382: enable global calculated coords / variables for all users

### LINKED ISSUE #12382: [Global Caclulated Waypoints/Variables]  make feature available for all users
### Tell us your idea!

The feature of global calculated waypoints/variables is currenty only available for developers and nightly users. This is due to its current development state. This issue is a collector issue for all things which have to be done to put the feature live for all users (beta + later production).

Following things are open for that:
- [x] #12381 
- [x] #12379 
- [x] #12378 

Apart from functionality I would expect that the GUI could use a better look than it currently has. I depend on suggestions here though, I am not very good at this. 

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

eddiemuc:
fix #12382: enable global calculated coords / variables for all users

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1505,9 +1505,8 @@
         <item>@string/waypoint_duplicate</item>
     </array>
     <string name="waypoint_cache_coordinates">Cache coordinates</string>
-    <string name="waypoint_calculate_coordinates">Calculate coordinates</string> <!-- Legacy: button text in coordinate input -->
-    <string name="waypoint_calculated_coordinates">Calculated coordinates</string> <!-- Legacy: line in waypoint description -->
-    <string name="waypoint_calculated_coordinates_global">Calculate coordinates (new)</string>
+    <string name="waypoint_calculated_coordinates">Calculated coordinates</string> <!-- Marks legacy calculated waypoints in waypoint list -->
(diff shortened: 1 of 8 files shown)


main/res/values/strings.xml
```diff
@@ -1507,7 +1507,7 @@
     <string name="waypoint_cache_coordinates">Cache coordinates</string>
     <string name="waypoint_calculate_coordinates">Calculate coordinates</string> <!-- Legacy: button text in coordinate input -->
     <string name="waypoint_calculated_coordinates">Calculated coordinates</string> <!-- Legacy: line in waypoint description -->
```
Lineflyer, line 1509: Is this one and the one above still needed?

eddiemuc: @Lineflyer there is a whole lot of code in c:geo which will no longer be needed after the switch. However, I want to keep it for now and see if things stay stable for at least one release. Then I will remove all that code. 

I did the same thing with filtering, and I find this to be a usefuil pattern. To not forget about the removal I created issue #12380 a while ago. I also collect there the things which need removal.

eddiemuc: That said, the two lines mentioned by you may create an "unused resource" warning. Have to check that.

eddiemuc: Removed one, kept the other (which is still used)

PR #12772, opened 2022-02-19
fix #12383, fix #12596: add ability to scan and copypaste calculated waypoints

### LINKED ISSUE #12383: [Calculated Waypoints] Implement Scan Function trying to find calculated waypoints in listing and auto-create Calculated Waypoints for them
c:geo should provide a scanner function in the wqypoint tab (third button "Scan for Wqypoints" under "Add Wayppoint" and "Add current location") which scans the cache listing for potential calculated waypoints and offers to create them. Quite similar to "+ Scan" function for variables, but for waypoints

### LINKED ISSUE #12596: WP Calc - Allow C&P for whole coord at once
### Tell us your idea!

Maybe its possible and I just don't know how:

In cache listings I often find things like `N 51° 15. (B) (A+B) (C-A+B) E 06° 5C. (B) (A-B) (C)` which are well prepared fomulas for the new parser but the usage flow to get them there is not always easy.

I would like to have the ability to copy & paste these in one batch to the WP calculator. 

One step further:
Ideally there could be some feature comparable to the "scan listing" which finds such strings and creates calculated waypoints for it.

### Is there an alternative?

At the moment you have to either copy latitude and logitude part after another, or copy the whole string and delete the non needed part in either field (lat/lon).

### Additional Information

_No response_

eddiemuc:
fix #12383, fix #12596: add ability to scan and copypaste calculated waypoints

### DIFF
--- a/main/res/layout/coordinatescalculateglobal_dialog.xml
+++ b/main/res/layout/coordinatescalculateglobal_dialog.xml
@@ -45,15 +45,10 @@
                 tools:listitem="@android:layout/simple_spinner_item" />
 
             <Button
-                android:id="@+id/cc_remove_spaces"
-                style="@style/button"
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:layout_marginLeft="10dp"
-                android:layout_marginRight="10dp"
-                android:layout_toRightOf="@id/cc_switch_guided"
+                android:id="@+id/cc_plain_tools"
+                style="@style/button_icon_accent"
+                android:layout_alignParentRight="true"
                 app:icon="@drawable/ic_menu_auto_fix_high"
-                android:text="@string/calccoord_remove_spaces"
                 android:layout_centerVertical="true"/>
 
             </RelativeLayout>
--- a/main/res/menu/cache_options.xml
+++ b/main/res/menu/cache_options.xml
@@ -110,6 +110,12 @@
                 android:visible="false">
             </item>
             <item
+                android:id="@+id/menu_scan_calculated_waypoints"
+                android:title="@string/cache_menu_scan_calculated_waypoints"
+                app:showAsAction="ifRoom"
+                android:visible="false">
+            </item>
+            <item
                 android:id="@+id/menu_toggleWaypointsFromNote"
                 android:title="@string/cache_menu_preventWaypointsFromNote"
                 app:showAsAction="ifRoom"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1366,6 +1366,7 @@
     <string name="menu_caches">Manage caches</string>
     <string name="cache_menu_event">Add to calendar</string>
     <string name="cache_menu_extract_waypoints">Extract waypoints</string>
+    <string name="cache_menu_scan_calculated_waypoints">Scan for waypoint formulas</string>
     <string name="cache_menu_set_found">Set as found</string>
     <string name="cache_menu_set_DNF">Set as did not find</string>
     <string name="cache_menu_reset_foundstate">Reset found state</string>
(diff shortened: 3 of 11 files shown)


eddiemuc:
Adds following option to scan for calculated waypoints:

image

image

image

Also adds the following copy&Paste option in calculated coordinate dialog:

image

image

Wrt algorithm to extract calculated waypoints from text/listing -> see #12383

bekuno:
I propose to rename it to "Scan for waypoints to be calculated", as in the moment of scanning they are not calculated.

eddiemuc:
> I propose to rename it to "Scan for waypoints to be calculated", as in the moment of scanning they are not calculated.

When the waypoint is created, c:geo will try to calculate it immediately (and store the resulting coordinate as the waypointÄs coord). This will work out only if all vars in the waypoint are already defined of course.

bekuno:
> > I propose to rename it to "Scan for waypoints to be calculated", as in the moment of scanning they are not calculated.
> 
> When the waypoint is created, c:geo will try to calculate it immediately (and store the resulting coordinate as the waypointÄs coord). This will work out only if all vars in the waypoint are already defined of course.

From user point of view to the listing the waypoint has to be calculated. It will be (maybe) calculated by c:geo after the scan.

ztNFny:
I agree that the wording should be changed. It's not a "calculated waypoint" in the listing but a formula.
How about "Scan for formulas" or "Scan for waypoint formulas"?

eddiemuc:
> I agree that the wording should be changed. It's not a "calculated waypoint" in the listing but a formula. How about "Scan for formulas" or "Scan for waypoint formulas"?

That would be wrong, it specifically scans for calculateable coordinates (which may or may not contain formulas), not for standalone formulas. Eg. it will find `N40 12.ABC E11 34.DEF`, but it will not find a standalone `A+B/2` (that would be found by "scan" function in variable page).

For now I changed to `Scan for waypoints to be calculated` as proposed by @bekuno. To be honest, I really don't have a strong opinion about the exact wording :-)

ztNFny:
>  That would be wrong, it specifically scans for calculateable coordinate

That's why I suggested "Scan for WAYPOINT formulas" though?!

"to be calculated" sounds awkward to me, like trying to avoid the word "formula" no matter what. If I wouldn't know what it does I'd have a hard time figuring it out. 
Also looking at the branch name "scan formula" seems like an obvious candidate.

As for the scan function in variables page, I actually don't see any usecase for that once we have this feature. Can't think of any cache using formulas for something that's not coordinates.

eddiemuc:
> > That would be wrong, it specifically scans for calculateable coordinate
> 
> That's why I suggested "Scan for WAYPOINT formulas" though?!

Ok, that's a fair point. How do you want to decide about this? As mentioned before, I don't really have a stroing opinion here. @bekuno would you approve (since the initial request came from you)?
 
> Also looking at the branch name "scan formula" seems like an obvious candidate.

Ah, my branch names are usually not a good indicator for these kind of things :-)

> As for the scan function in variables page, I actually don't see any usecase for that once we have this feature. Can't think of any cache using formulas for something that's not coordinates.

I had in mind things like e.g. in GC5JAQ6. Deep down in the listing you find (sorry, it's german): (peilt von hier...) `[X + Y + 7] Meter in Richtung [5*X + Y + 7] Grad`. Obviously this use case is discussable, but maybe in a separate issue?

bekuno:
>> That's why I suggested "Scan for WAYPOINT formulas" though?!

> Ok, that's a fair point. How do you want to decide about this? As mentioned before, I don't really have a stroing opinion here. @bekuno would you approve (since the initial request came from you)?

Approved, sounds good to me and is shorter.

eddiemuc:
> > > That's why I suggested "Scan for WAYPOINT formulas" though?!
> 
> > Ok, that's a fair point. How do you want to decide about this? As mentioned before, I don't really have a stroing opinion here. @bekuno would you approve (since the initial request came from you)?
> 
> Approved, sounds good to me and is shorter.

Ok, changed in PR

image

Issue #12769, opened 2022-02-17
Variables Input: show abbrevation for functions on function selection dialog


murggel:
### Tell us your idea!

While entering the formula for a variable, I need sometimes concatenated functions (e.g. checksum of lettervalue).
On selecting the second function the first one will be replaced.

### What solution would you suggest?

Insert the selected function at cursor position

eddiemuc:
The intention behind the formula insert function was meant for "novice" users who don't like to deal too much with formulas, and as a same time to quickly look up available functions. I always assumed that more advanced users (and only those would use more than one function in a var expression) would just type along.

I also provided abbreviations for most functionsnto ease typing e.g. wv, lv or bww can be used instead of lettervalue, and cs can be used instead of checksum

murggel:
> I also provided abbreviations for most functionsnto ease typing e.g. wv, lv or bww can be used instead of lettervalue, and cs can be used instead of checksum

Ok, that's fine.
Can we add those abbreviations behind the name of the function for a quick lookup then?

eddiemuc:
> > I also provided abbreviations for most functionsnto ease typing e.g. wv, lv or bww can be used instead of lettervalue, and cs can be used instead of checksum
> 
> Ok, that's fine. Can we add those abbreviations behind the name of the function for a quick lookup then?

Well there **IS** a big "Help" button on the page leading to complete documentation including available functions and usable abbrevations...

![image](https://user-images.githubusercontent.com/6909759/154804273-ead24ca1-bd18-41ed-8049-468c19cec576.png)

![image](https://user-images.githubusercontent.com/6909759/154804491-52d57fdc-9b3b-42b9-b379-2bb8f55cdd39.png)


Adding it to the list will make it less readable, would be look something like this:

![image](https://user-images.githubusercontent.com/6909759/154804435-641a7f79-42d7-4dce-8d46-b0d0f2e49f18.png)

Implementing this would be a one-liner, just change method VariableListView.getFunctionDisplayString to whatever should be displayed. You may use Markdown or HTML if you want (since this is a TextParam). Feel free to try out a few things.

E.g. like this:

![image](https://user-images.githubusercontent.com/6909759/154804500-51ee6179-8fff-4b93-9cc4-dfd8108bc8f0.png)

murggel:
For me it would only need the abbreviations, not the "long" word.
But probably it is really only learning the few abbreviations for lettervalue, checksum and vanity code...

eddiemuc:
[...]

PR #12781, opened 2022-02-20
rel to #12769: show short abbrevation of function in list

eddiemuc:
rel to #12769: show short abbrevation of function in list

### DIFF
--- a/main/src/cgeo/geocaching/ui/VariableListView.java
+++ b/main/src/cgeo/geocaching/ui/VariableListView.java
@@ -347,7 +347,14 @@ public class VariableListView extends LinearLayout {
         }
 
         private static TextParam getFunctionDisplayString(final FormulaFunction f) {
-            return TextParam.text(f.getUserDisplayableString());
+            //find the shortest abbrevation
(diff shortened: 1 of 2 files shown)


eddiemuc:
As wished in #12769, this PR implements showing the first shortest abbrevation for each function behind the translated function name in the list. Only function abbrevations with 3 or less chars are considered, if function does not have such an abbrevation, nothing is shown.

This looks like this:

image
image

bekuno:
This looks for user errorness, as only some entries are without a bracket entry.
I propose to present the shortest one for all points.
Maybe it is possible to create shorter commands for the missing points (e.g. Zeichenlänge => len, Runden => rd)?

eddiemuc:
> This looks for user errorness, as only some entries are without a bracket entry. I propose to present the shortest one for all points. Maybe it is possible to create shorter commands for the missing points (e.g. Zeichenlänge => len, Runden => rd)?

Changed as suggested, now looks like this:

image
image

bekuno (approved):

Issue #5782, opened 2016-06-11
New map: Map from list opens at current position


Lineflyer:
##### Detailed steps causing the problem:
- Open live map
- Activate to center own position
- Leave live map
- Open list of caches which are at another location
- Map the list

##### Actual behavior after performing these steps:
Map opens at own position (no cache shown as out of view)

##### Expected behavior after performing these steps:
Map should cover the coords of the listed caches

##### Version of c:geo used:
2016.06.10-NB1

##### Is the problem reproducible for you?
Yes

##### Other comments and remarks:
Not tested explicitly if this also happens with old map but I never before experienced the problem on old map

PR #12786, opened 2022-02-20
fix #12768: stay at position on map change from google to osm

### LINKED ISSUE #12768: Map jumps to current position
### Describe your problem!

Received three individual support mails describing the problem, that with 2022.02.13 the map jumps to the current position when changin the map source.
This reminds me of an older (fixed) issue and I cannot reproduce it.

### How to reproduce?

- Open GoogleSat map
- Scroll to a random position
- Change map source to OSM

Map changes to OSM but jumps back to current device position

Map changes to OSM but centered to same position as before

### Reproducible

No

### c:geo Version

2022.02.13

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #12768: stay at position on map change from google to osm

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -248,8 +248,12 @@ public class NewMap extends AbstractBottomNavigationActivity implements Observer
             individualRoute = savedInstanceState.getParcelable(BUNDLE_ROUTE);
             followMyLocation = mapOptions.mapState.followsMyLocation();
         } else {
+            if (mapOptions.mapState != null) {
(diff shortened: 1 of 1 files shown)


eddiemuc:
I tried to copy the behaviour used in CGeoMap (=Google Map) to NewMap (=OSM map), and it seems to do the trick

eddiemuc:
I would appreciate a short review before merging this to release

Lineflyer:
While I cannot really judge the change itself I used git blame to check up other changes in that area.
There was a change years ago introducing the line you now optimized:
https://github.com/cgeo/cgeo/commit/273e75e881f7c9c3c8c46096f3d2c3a61e3a9dd6

At that time it was a fix for the map opening at current position when mapping a list (see #5782). As long as it does not break this part, your change looks logical to me.

PR #12793, opened 2022-02-21
fix #12780: prevent keyboard fullscreen

### LINKED ISSUE #12780: [Variables] Opt-out from using fullscreen keyboard
Currently, the well known Android fullscreen keyboard UI is shown when editing a variable in landscape mode on a phone. 

However, it is possible to opt out using this input behavior via `android:imeOptions="flagNoFullscreen"`. Doing so would IMHO improve usability... 

Screenshot_20220219_225935_com whatsapp w4b

eddiemuc:
fix #12780: prevent keyboard fullscreen

### DIFF
--- a/main/res/layout/variable_list_item_advanced.xml
+++ b/main/res/layout/variable_list_item_advanced.xml
@@ -65,7 +65,6 @@
                         android:id="@+id/variable_formula_text"
                         android:inputType="text|textNoSuggestions"
                         tools:ignore="LabelFor" />
-
                 </com.google.android.material.textfield.TextInputLayout>
             </LinearLayout>
 
--- a/main/res/values/styles.xml
+++ b/main/res/values/styles.xml
@@ -269,6 +269,7 @@
         <item name="colorOnSurface">@color/colorText</item>
         <item name="android:gravity">top|left</item>
         <item name="android:textSize">@dimen/textSize_textinputPrimary</item>
+        <item name="android:imeOptions">flagNoFullscreen</item>>
     </style>
 
     <!-- edittext (TODO: should at some point be completely replaced by textinput) -->


fm-sys:
Hm, just thinking: Should we use it only for variables or all text inputs? Both would be ok for me...

eddiemuc:
Changed to all text inputs

Lineflyer:
That's a quite massive change if I interprete the consequences correctly, not sure if this fits all dialog designs.
Will take a look on emulator first (but we can for sure test it on master/in nightly if no one objects).

eddiemuc:
Well we could create a new setting for the user to decide whether he wants it everywhere or just in variable formulas...

Just kidding :-)

Should I merge?

Lineflyer:
IMHO its not usable at some places (e.g. multi-line text inputs like the personal note):
Without PR:
image

With this PR:
image

eddiemuc:
Ok won't merge

Lineflyer:
@eddiemuc For the variable list as originally requested it would maybe suitable, don`t you think?

fm-sys:
@Lineflyer  As I'm always looking for consistent solutions: What about disabling the fullscreen overlay for single line text inputs, but keep in for multi-line text fields? Would this be ok?

Lineflyer:
Did not really take a look of how it changes all the input we have and whether it only for multi-line has such negative impacts.
Furthermore: Is this limited to landscape or just on "space available"? I am not sure what else might be affected.

Generally speaking:
It might have a good reason, that fullscreen is the default behavior and this is an "opt-out" parameter. Not sure therefore if its wise to opt-out in general and patch it back in. It sounds more like "opt-out" for those dialogs which are unhandy to use in landscape. 
Indeed I had such problem with the (old) waypoint calculator guided mode in the past.

fm-sys:
> Furthermore: Is this limited to landscape

Yes, it's a landscape only feature.

> Not sure therefore if its wise to opt-out in general and patch it back in

We have two different styles for single/multi-line text layouts. So it would be a simple one liner to disable it only for one of both styles. No need for ugly "patching it back in". It's even way clearer than if we have to set it per layout... I can't imagine any side effects for single line textfields and even if - that's what nightlys are designed for ;-) 

To be honest, I never have spoken to a single android user who likes this fullscreen feature.

fm-sys:
Let's continue in #12780

PR #12801, opened 2022-02-22
Use c:cgeo user guide for formula syntax help

Lineflyer:
## Description

This changes the target for the help button to our user guide.

As a first step the English page is targeted.
Further improvement could be to dynamically create the link based on user language.
See: https://github.com/cgeo/cgeo-guide/issues/54

## Related issues

https://github.com/cgeo/cgeo-guide/issues/54

### DIFF
--- a/main/res/values/strings_not_translatable.xml
+++ b/main/res/values/strings_not_translatable.xml
@@ -306,5 +306,5 @@
     <string translatable="false" name="mailsubject_problem_report">c:geo problem report (%s)</string>
 
     <!-- Help Ressources -->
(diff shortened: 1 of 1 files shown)


eddiemuc:
There's a syntax error in the file also found by ci: missing closing string tag. Apart from that change is fine.

Generating language-specific urls might be added later when we actually have translation available

Lineflyer:
> There's a syntax error in the file also found by ci: missing closing string tag. Apart from that change is fine.

Ups...corrected.

I have also seen we use other links to the manual also exclusively targeting English. I guess we could leave it like that in our code as users can easily change the language after having opened the page.

PR #12802, opened 2022-02-22
prevent keyboard fullscreen in single line text fields

fm-sys:
As discussed in #12780: Prevent keyboard full-screen in single line text fields

### DIFF
--- a/main/res/layout/authorization_credentials_activity.xml
+++ b/main/res/layout/authorization_credentials_activity.xml
@@ -28,7 +28,7 @@
             <EditText
                 android:id="@+id/username"
                 android:inputType="textPersonName"
-                style="@style/textinput_embedded"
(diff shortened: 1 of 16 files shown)


Lineflyer:
Thanks @fm-sys, we will review this PR.
Be reminded to work for your exams instead...not meaning to be intrusive. just saying. Take your time for that step, c:geo will be there throughout.
Anyway I can understand coding is a valid means to relax from learning, just keep the balance ;)

fm-sys:
Thanks for the reminder :)

I'm already unsubscribed from the general notification emails and only watching a few special threads. Some small 15-minutes-evening-PRs might or might not come from time to time, but I will definitely stay away from larger projects...

Lineflyer:
As there were no negative comments, lets test this solution in nightlies.

PR #12817, opened 2022-03-01
support quick launch buttons on the home screen (fix #12790)

### LINKED ISSUE #12790: make "Nearby" button configurable
### Tell us your idea!

There are users, who will use "Go to" often, but "Nearby" seldom.
To support this we can make the "Nearby" button configurable.

Idea 1:
Let user decide, which function will have this button. 
This could be configuread in the same way as for the navigation button.

Idea 2:
Add long tap function to this button to select a second function.

### Is there an alternative?

Only add "Go to" as long tap function. 

### Additional Information

_No response_

moving-bits:
## Description
Add support for quick launch icons on homescreen.

- set of up to eight buttons, in a fixed order
- being displayed at the bottom of the home screen, centered horizontally, above location info
- configurable using settings => appearance => quick launch buttons
- button sections being displayed only if at least one quick launch setting is selected
- default configuration is "none selected"

|home screen|settings|select items|none selected|
|---|---|---|---|
|image|image|image|image|

### DIFF
--- a/main/res/layout/main_activity.xml
+++ b/main/res/layout/main_activity.xml
@@ -88,7 +88,7 @@
         android:id="@+id/connectorstatus_area"
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:layout_above="@id/location_area"
+        android:layout_above="@id/quicklaunchitems"
         android:layout_below="@id/info_notloggedin"
         android:layout_marginTop="32dp"
         android:layout_marginBottom="16dp"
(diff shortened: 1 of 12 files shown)


fm-sys:
Thanks for working on this feature, really appreciated 👍

I personally don't really like the outlined button appearance in this context. I had created a UI draft some months ago, see https://github.com/cgeo/cgeo/issues/11904#issuecomment-953153108. If interested, the code can still be found here: https://github.com/fm-sys/cgeo/commit/16a78a1855bce1a7eb28df12cd5237e9f577c740

grafik

Lineflyer:
My few cents/question:
As the buttons only have icons: Do these buttons provide a long tap tooltip? That would be appreciated IMHO as not all of them are easy to recognize.

Regarding the styling:
I like @fm-sys proposal as well as the initial proposal... The proposal of @fm-sys looks good on black theme, how would it look on white theme?

fm-sys:
> The proposal of @fm-sys looks good on black theme, how would it look on white theme?

Would probably be something like this (the idea is copied from Signal)

Screenshot_20220302_114701_org thoughtcrime securesms_copy_325x141

ztNFny:
It looks a bit to condensed to me, I'd suggest to add more margins to the buttons. @fm-sys' mockup already has that

In terms of code I don't think we should hardcode 8 buttons in main_activity.xml but instead add them dynamically to id/quicklaunchitems when creating the activity.

ztNFny:
I tweaked the style a bit:
image
image

I'm wondering whether these buttons - while the same size as other buttons in c:geo - are too large. On dashboard they are in direct comparison with bottom nav and toolbar which both use way smaller icons. I'd suggest aligning them with those - but then we lose alignment with other buttons (e.g. in cache details).

ztNFny:
In terms of code I don't think we should hardcode 8 buttons in main_activity.xml but instead add them dynamically to id/quicklaunchitems when creating the activity.

Also the current hard-coded selection is very limited, I don't think the feature should be introduced with such a limitation but should allow for more freedom already in V1 (e.g. selecting a custom list, deep settings links, ...)

moving-bits:
> In terms of code I don't think we should hardcode 8 buttons in main_activity.xml but instead add them dynamically to id/quicklaunchitems when creating the activity.

I had started with dynamic button creation, but that did not work well with outlined material buttons, so I switched to hard-coded buttons. If it works with the new style, than we should definitively switch to dynamic button creation, which would also allow to get rid of that ugly dynamic id retrieval.

moving-bits:
> In terms of code I don't think we should hardcode 8 buttons in main_activity.xml but instead add them dynamically to id/quicklaunchitems when creating the activity.

Finally made adding buttons dynamically working after some tweaking (as still some style parameters have to be set manually, even though they are part of the style definition).

> Also the current hard-coded selection is very limited, I don't think the feature should be introduced with such a limitation but should allow for more freedom already in V1 (e.g. selecting a custom list, deep settings links, ...)

Yes, current implementation is pretty basic, having eight standard actions. I could imagine parametrized actions, user-selectable icons and user-selectable order, but sadly I'm pretty time-constrained currently, so feel free to add commits to extend the feature set if you like (and have time available).

Lineflyer:
> > Also the current hard-coded selection is very limited, I don't think the feature should be introduced with such a limitation but should allow for more freedom already in V1 (e.g. selecting a custom list, deep settings links, ...)
> 
> Yes, current implementation is pretty basic, having eight standard actions. I could imagine parametrized actions, user-selectable icons and user-selectable order, but sadly I'm pretty time-constrained currently, so feel free to add commits to extend the feature set if you like (and have time available).

IMHO we could also do it step by step and start with these basics.
This has an advantage, that on one hand users asking us for shortcuts to "Go to" have a solution on the other hand we will probably get feedback an additional desired shortcuts and include this into the work on V2.

moving-bits:
Ok, then let's give it a try in the nightlies.

fm-sys:
Can we give these buttons a bit more bottom margin? Currently, they look really nailed... 

Screenshot_20220314_064026_cgeo geocaching_copy_270x585

PR #12820, opened 2022-03-01
fix #10621: enable Proxy WebAPI v2 and gc.com attribute search

### LINKED ISSUE #10621: Gc.com search for attributes not implementable: WebApi proxy v2: cgeo explicitely not authorized?

eddiemuc:
fix #10621: enable Proxy WebAPI v2 and gc.com attribute search

### DIFF
--- a/main/src/cgeo/geocaching/SearchResult.java
+++ b/main/src/cgeo/geocaching/SearchResult.java
@@ -37,8 +37,6 @@ import org.apache.commons.lang3.StringUtils;
 
 public class SearchResult implements Parcelable {
 
-    public static final String SEARCHSTATE_FINDER = "ss_finder";
-
     public static final String CON_LEFT_TO_FETCH = "con_lefttofetch";
     public static final String CON_URL = "con_url";
     public static final String CON_ERROR = "con_error";
--- a/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
@@ -163,6 +163,10 @@ class GCWebAPI {
             return this;
         }
 
+        public Set<CacheAttribute> getCacheAttributes() {
+            return this.cacheAttributes;
+        }
+
         /**
          * set to true to show ONLY own caches, false to HIDE own caches, null if both should be shown.
          * Works only for Premium members!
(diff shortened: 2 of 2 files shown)


eddiemuc:
This PR, if merged, will enable usage of gc.com Proxy WebAPI v2 instead of the currently used v1 (=non-versioned).

As a positive side effect it will now allow c:geo to do online searches for attributes on gc.com

Lineflyer:
Just thinking:
Does it make sense to merge `master` into `release` before merging this PR to `master`.

As both the market release and the nightlies are IMHO both performing good, we could cut an `release` including the variables (with beta phase) and have more time to test v2 API on `master` (and also we do not yet have an official reply, that it will keep working..still waiting for it).

eddiemuc:
If V2 is not working, we can switch back to V1 just by deleting three chars in code (the part `/v2` of the url)

Lineflyer:
> If V2 is not working, we can switch back to V1 just by deleting three chars in code (the part `/v2` of the url)

Saw that. But i thought a bit of stability and function check would be nice.
Who knows if it behaves the same as v1 in all details after all.

Lineflyer:
We received confirmation by Groundspeak, that the problem is fixed.

PR #12825, opened 2022-03-03
use gradle-versions-filter

bekuno:
Optimize the output of `gradlew dependencyUpdates` by applying default exclude rules (e.g. for RC or alpha versions).

So there are excluded now
 - com.android.tools.build:gradle [4.2.2 -> 7.3.0-alpha04]
 - com.squareup.okhttp3:mockwebserver [4.9.3 -> 4.10.0-RC1]
 - com.squareup.okhttp3:okhttp [4.9.3 -> 4.10.0-RC1]

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ plugins {
 
     // check for updates of gradle plugin dependencies
     id 'com.github.ben-manes.versions' version '0.42.0'
+    id 'se.ascp.gradle.gradle-versions-filter' version "0.1.10"
 
     // use SpotBugs instead of FindBugs, see https://plugins.gradle.org/plugin/com.github.spotbugs
     id 'com.github.spotbugs' version '4.7.6'


moving-bits:
@bekuno 

Not sure if it's related, but from the error message it somewhat looks like it:

cgeo nightly # 2645 failed with the following error message:

```
A problem occurred configuring root project 'cgeo'.
> Could not resolve all files for configuration ':classpath'.
   > Could not resolve se.ascp.gradle:gradle-versions-filter:0.1.10.
     Required by:
         project : > se.ascp.gradle.gradle-versions-filter:se.ascp.gradle.gradle-versions-filter.gradle.plugin:0.1.10
      > No matching variant of se.ascp.gradle:gradle-versions-filter:0.1.10 was found. The consumer was configured to find a runtime of a library compatible with Java 8, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '7.3.3' but:
          - Variant 'apiElements' capability se.ascp.gradle:gradle-versions-filter:0.1.10 declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares an API of a component compatible with Java 11 and the consumer needed a runtime of a component compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '7.3.3')
          - Variant 'runtimeElements' capability se.ascp.gradle:gradle-versions-filter:0.1.10 declares a runtime of a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares a component compatible with Java 11 and the consumer needed a component compatible with Java 8
              - Other compatible attribute:
                  - Doesn't say anything about org.gradle.plugin.api-version (required '7.3.3')
```

Anything that needs to be configured on one of our machines, or has it been a temporary error only?

eddiemuc:
Seems to be the JDK version used. Android Studio uses JDK 11 and it works. However, when I manually switch to a JDK 1.8 version in Android Studio, I get the very same error locally.

If I set this:

image

Then I get this:

image


I suspect that we need to install and use JDK 11 on the CI servers building nightlies and releases?

moving-bits:
> I suspect that we need to install and use JDK 11 on the CI servers building nightlies and releases?

Thanks for cross-checking. I'll try to reach @mucek4 for an upgrade...

PR #12828, opened 2022-03-04
extend map's 'long tap' setting to waypoints (fix #12821)

### LINKED ISSUE #12821: "long tap on map" & waypoints
### Tell us your idea!

I can set whether a long tap on the general map creates a new CACHE or not.
But when I look at the map of a cache, a long tap always wants to create a new WAYPOINT.

addtional option in "settings/Map/Map behavior" to set whether a waypoint should be created with a long click or not

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Make map's "long tap adds waypoint" function honor the map's "long tap enabled/disabled" setting.

### DIFF
--- a/main/res/xml/preferences_map.xml
+++ b/main/res/xml/preferences_map.xml
@@ -142,7 +142,7 @@
         android:title="@string/settings_title_map_behavior"
         app:iconSpaceReserved="false">
         <CheckBoxPreference
-            android:defaultValue="false"
+            android:defaultValue="true"
             android:key="@string/pref_longTapOnMapActivated"
             android:summary="@string/init_summary_longtap_map"
             android:title="@string/init_longtap_map"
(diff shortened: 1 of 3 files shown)


Lineflyer:
mmh..that means that it will deactivate the waypoint creation long tap for many users as the default for that now combined setting is deactivated.

Changing the default to active however will be intrusive for all live map users.

moving-bits:
That's true. But IIRC correctly the main reason for introducing that switch at all was that some users have problems with long taps activating something too early for them - which should apply to both UDC creation and waypoint creation. Therefore it sounds reasonable to me to extend the setting to both cases.

ztNFny:
Maybe change the default setting to true at the same time?

Lineflyer:
I cant tell how many people actually use the waypoint creation within a caches map context, however as this feature was implemented way earlier than the UDC long tap I guess its more likely known and used than the UDC creation.

So I think the best way out to solve these conflicting defaults is either to have a dedicated setting for waypoints (with default true) or just skip this change and keep it as it is. There was AFAIK never ever a support request or ticket here regarding accidental waypoint creation.

moving-bits:
In today's dev meeting (and with considering the latest user feedback in the related issue) we decided to
- have only one setting to handle "long tap on map", handling both long tap for UDC (general map) and for waypoints (cache map)
- and to make the default for this setting "enabled".

I will update this PR accordingly.

eddiemuc:
lgtm, thanks for tackling this

PR #12835, opened 2022-03-06
fix #12824: include waypoints in gpx for IndividualRouteExport

### LINKED ISSUE #12824: Include caches for individual route export
### Tell us your idea!

If you export an individual route as track, no waypoints are currently included. Therefore this track is pretty useless if you want to use if for your caching trip. 

Inside cgeo |Locus / Garmin Outdoor device -> no waypoints are visible 
-|-
Screenshot_20220302_213124_cgeo geocaching|Screenshot_20220302_213013_menion android locus pro computerBild

We should include the caches as waypoints when exporting the individual route as track/route. 

### Is there an alternative?

It was to be decided if we want to include a normal GPX waypoint containing only the name, or the full cache (which is in fact also just a GPX waypoint with more details) 

### Additional Information

_No response_

fm-sys:
Add caches as waypoint, when exporting an individual route as track/route. This makes the exported GPX useful in context of other outdoor apps:

Individual Route in c:geo|Garmin Cycle Computer |Locus
-|-|-
Screenshot_20220306_015704_cgeo geocaching_copy_540x1170|IMG_20220306_020301-01_copy_960x1280|Screenshot_20220306_020136_menion android locus pro computerBild_copy_540x1170

Note: This is the MVP approach currently. Only a waypoint with the geocode as name is added to the GPX, not the full cache...

fix #12824

### DIFF
--- a/main/src/cgeo/geocaching/export/IndividualRouteExport.java
+++ b/main/src/cgeo/geocaching/export/IndividualRouteExport.java
@@ -134,6 +134,10 @@ public class IndividualRouteExport {
                 XmlUtils.simpleText(gpx, NS_GPX, "time", timeInfo);
                 gpx.endTag(NS_GPX, "metadata");
 
+                for (RouteSegment loc : trail) {
+                    exportPoint(gpx, "wpt", loc.getItem().getPoint(), loc.getItem().getIdentifier());
+                }
+
                 gpx.startTag(NS_GPX, exportAsTrack ? "trk" : "rte");
                 XmlUtils.simpleText(gpx, NS_GPX, "name", "c:geo individual route " + timeInfo);
                 for (RouteSegment loc : trail) {


dcjkfgdjhd:
I noticed this change yesterday for the first time. 

It has caused me some problems on my bike tour yesterday, when planning a tour with caches, and then copying it to my Garmin Oregon 550 via an OTG cable:

- In order to shape my route, I will define many intermediate points while planning the route in c:geo. These points have no use in themselves; I just create them to help the routing algorithm pick the paths I want. In the current setup, they all end up as waypoints on my Garmin GPS, which is a lot of trash waypoints. Please add an option that allows the user to choose whether to include waypoints or not in the track export.
- Yesterday, I exported *both* the track and a *second* GPX file containing all caches along the way. My Garmin was *not* able to display the caches as caches. They only showed up as normal waypoints called "GC...". Probably the waypoints in the track export overrided the caches contained in the other GPX file. (I did some testing, but was unable to reproduce the behaviour. I suspect it has something to do with the order in which Garmin scans both GPX files?)

PR #12837, opened 2022-03-06
 fix #12770: apply stored hidden flag to route on load

### LINKED ISSUE #12770: Track visibility setting is not restored correctly
### Describe your problem!

When you import a GPX track but select to hide it on the map, it will become visible again if you reopen the map. 

### How to reproduce?

1. Open Livemap
2. Import track
3. Hide track
4. Close livemap
5. Open livemap 

Track is visible again 

Track should be hidden according to the last selection 

### Reproducible

Yes

### c:geo Version

2022.02.16-NB

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #12770: apply stored hidden flag to route on load


eddiemuc:
@moving-bits I would appreciate a review from you on this

moving-bits (approved):
LGTM, thanks for diving into this!

PR #12853, opened 2022-03-11
WIP make bugfix version number handling easier (fix #12852)

### LINKED ISSUE #12852: Avoid repetitive translation for changelog header
From Crowdin a user provided the idea, whether it would be possible to split the version number/name from the string `bugfix release`. This would allow a one time translation of the string and thus no more need for repetitive translation of the whole header short before a bugfix release occurs.

@moving-bits 
I could imagine, that we either skip keeping older changelogs in `changelog_bugfix.md` and only collect the changes for the next bugfix release. In such case the header could be constructed internally (like we do for feature releases).

Another possibility would maybe to use dedicated files for reach bugfix release (e.g. `changelog_20220310.md`) and then construct the header from the file name (which contains the version date) and just append the content. This could be done sequentially for all existing `changelog_*` files.

What do you think?

moving-bits:
## Description
Proposal for easier version number handling for bugfix releases - not to be merged yet.

@Lineflyer 
Please have a look at `BranchDetectionHelper.java` and `changelog_bugfix.md` - would this type of version number handling for bugfix releases be ok for you?
- `BranchDetectionHelper.java` contains a string array with version numbers for bugfix releases (will be empty´for feature release)
- `changelog_bugfix.md` contains `##` placeholders only at the places where version numbers will be inserted

Magic configuration line in `BranchDetectionHelper.java` would look like this for current release:
```
// should contain version names of active bugfix releases since last feature release, newest first
// leave curly brackets empty if not a bugfix release
public static final String[] BUGFIX_VERSION_NAME = new String[]{"2022.03.10", "2022.03.09", "2022.02.16"};
```

Results for current release:

image

Just a little thing to be discussed for fine-tuning:

Currently version numbers are inserted top-down, which ensures that the first placeholder always gets a version number (if at least one is configured in `BranchDetectionHelper.java`. Downside is, that we have to remember to add the placeholder `##` just before releasing, so two files need to be updated then.
Alternatively we could insert the version numbers bottom-up. This would enable us to always have `##` at the top of the `changelog_bugfix.md`, and reduces the need for files to be changed just before a bugfix release down to 1, but can lead to an empty version number at the changelog's start if we forgot to enter a version number in `BranchDetectionHelper.java` before releasing.

### DIFF
--- a/main/src/cgeo/geocaching/AboutActivity.java
+++ b/main/src/cgeo/geocaching/AboutActivity.java
@@ -19,6 +19,7 @@ import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.ShareUtils;
 import cgeo.geocaching.utils.SystemInformation;
 import cgeo.geocaching.utils.Version;
+import static cgeo.geocaching.utils.BranchDetectionHelper.BUGFIX_VERSION_NAME;
 
 import android.app.Activity;
 import android.content.Context;
@@ -37,12 +38,16 @@ import androidx.annotation.RawRes;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.ActionBar;
 import androidx.core.util.Consumer;
+import androidx.core.util.Pair;
 
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import io.noties.markwon.Markwon;
 import org.apache.commons.compress.utils.IOUtils;
@@ -51,6 +56,7 @@ import org.apache.commons.lang3.StringUtils;
 public class AboutActivity extends TabbedViewPagerActivity {
 
     private static final String EXTRA_ABOUT_STARTPAGE = "cgeo.geocaching.extra.about.startpage";
+    public static final Pattern VERSION_PLACEHOLDER_PATTERN = Pattern.compile("((^|[\\r\\n]+)[ \\t]*##[ \\t]*[\\r\\n]+)");
 
     private final GatherSystemInformationTask systemInformationTask = new GatherSystemInformationTask();
 
@@ -221,7 +227,7 @@ public class AboutActivity extends TabbedViewPagerActivity {
             final Markwon markwon = Markwon.create(activity);
 
             final String changelogBase = FileUtils.getChangelogMaster(activity).trim();
-            final String changelogBugfix = FileUtils.getChangelogRelease(activity).trim();
+            final String changelogBugfix = prepareChangelogBugfix(activity);
             if (BranchDetectionHelper.isProductionBuild()) {
                 // we are on release branch
                 if (StringUtils.isNotEmpty(changelogBugfix)) {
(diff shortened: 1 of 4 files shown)


Lineflyer:
> Currently version numbers are inserted top-down, which ensures that the first placeholder always gets a version number (if at least one is configured in BranchDetectionHelper.java. Downside is, that we have to remember to add the placeholder ## just before releasing, so two files need to be updated then.

Lineflyer:
> would this type of version number handling for bugfix releases be ok for you?

Looks simple and good.

>  Downside is, that we have to remember to add the placeholder ## just before releasing, so two files need to be updated then.

If this can already done as soon as the first change is added for a potential upcoming bugfix release, it would be fine. Then its also already contained in all translations right in time.
Or will this be a problem as the array does not yet have an entry then?

main/src/cgeo/geocaching/AboutActivity.java
```diff
@@ -239,6 +243,20 @@ public void setContent() {
...
     }
 
+    private static String prepareChangelogBugfix(final Activity activity) {
+        String changelog = FileUtils.getChangelogRelease(activity).trim();
+        int current = 0;
+        final int max = BUGFIX_VERSION_NAME.length;
+        final Pattern p = Pattern.compile("((^|[\\r\\n]+)##[\\r\\n]+)");
+        Matcher match = p.matcher(changelog);
+        while (match.find() && current < max) {
+            changelog = changelog.substring(0, match.start(0)) + "\r\n## " + BUGFIX_VERSION_NAME[current] + " Bugfix Release\r\n" + changelog.substring(match.end(0));
```
Lineflyer, line 253: Just a reminder, that the string `Bugfix Release` should be externalized for translation. But I think you are aware of that, just still WIP.

moving-bits:
> > would this type of version number handling for bugfix releases be ok for you?
> 
> Looks simple and good.
> 
> > Downside is, that we have to remember to add the placeholder ## just before releasing, so two files need to be updated then.
> 
> If this can already done as soon as the first change is added for a potential upcoming bugfix release, it would be fine. Then its also already contained in all translations right in time. Or will this be a problem as the array does not yet have an entry then?

Ok, then I will pursue this path further, trying to be compatible with both having "##" at the beginning or not, and if so, coping with enough version codes in the array or not.

Still hard-coded strings will be made translatable for sure, just kept this PR as a minimum-WIP to discuss the concept.

BTW: I could also automatically add "Feature Release" to the base changelog being displayed, as soon as a version number for this has been entered. (Will do so if no one objects.)

moving-bits:
Additional notes from today's dev call as a reminder to myself:
- maybe we can automatically add some info auto-generated into `build.gradle`?
- add a test which checks consistency of bugfix version array and number of placeholders in `changelog_bugfix.md`

main/src/cgeo/geocaching/AboutActivity.java
moving-bits: done

moving-bits:
Updated the PR:
- use "##" placeholders in `changelog_bugfix.md` instead of version strings
- place version strings for bugfixes in `BUGFIX_VERSION_NAME` variable in `BranchDetectionHelper.java`, _oldest_ version string first
- replacing placeholders is bottom-up
- if there are less version strings than placeholders, remaining placeholders get "Next version" as version name (as is the default during development of a bugfix release) => by this you can directly enter the "##" placeholder into the empty `changelog_bugfix.md` on `release` to not forget to enter this placeholder later on
- when creating a new `release` branch from `master` do empty `changelog_bugfix.md` (leave only the leading "##" in there), and empty the `BUGFIX_VERSION_NAME` variable between the two curly brackets.

Lineflyer:
sounds perfect to try!

Issue #12848, opened 2022-03-10
Map crashes at random


alan666notb:
### Describe your problem!

I tried using a track loaded from a GPX file for the first time.

The map has been crashing on me quite often, but not often enough to reproduce reliably.

Typically I have the map (and its route) on display, lock the phone and put it in my pocket.
A few minutes later I unlock the phone and the screen is blank/white for a few seconds before returning to the home screen.
Tapping 'map' gets everything running again.

### How to reproduce?

1. load a map and load a GPX track
2. lock phone
3. wait a bit
4. unlock phone

### Actual result after these steps?

Map crashes and goes back to home page

### Expected result after these steps?

Map should continue

### Reproducible

Unclear

### c:geo Version

2022.03.09

### System information

```
c:geo version: 2022.03.10

Device:
-------
- Device type: M2101K6G (sweet_eea, Redmi)
- Available processors: 8
- Android version: 12
- Android build: SKQ1.210908.001 test-keys
- Screen resolution: 1080x2177px (392x791dp)
- Pixel density: 2.75
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 22.06.15 (190400-428792003)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:206
- Set language: en_GB (system default)
- System date format: dd/MM/y
- Time zone: GMT
- Debug mode active: no
- Last backup: 7 Mar, 11:01
- Routing mode: Off
- Live map mode: false
- OSM multi-threading: true / threads: 4
- Map: Great_Britain_OAM_V5_Feb2022 (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:primary:OpenStreetMaps/Great_Britain_OAM_V5_Feb2022.map
  - Atts: Maps: (c)Christian Kernbeis & Michael Bechtold , Themes: (c)Tobias Kuehn, www.openandromaps.org 2012-2022
  - Theme: ElevateTheme_V5.zip:Elevate.xml

Filters:
-------
- Hide waypoints: -
- LIVE: Status:Exclude Archived ([:inconclusive=false:advanced=true]AND(attributes:inverse=false:attributesources=1;status:exclude_archived))
[...]

PR #12862, opened 2022-03-16
catch IllegalStateException in map lifecycle (rel. to #12848)

moving-bits:
## Description
Map throws an IllegalStateException on stopping the map (lifecycle method) if a layer is not assigned internally, leading to random crashes as described in #12848. This PR catches that exception, so that the map can be shut down normally.

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/caches/AbstractCachesOverlay.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/caches/AbstractCachesOverlay.java
@@ -296,7 +296,11 @@ public abstract class AbstractCachesOverlay {
         synchronized (this.bundleRef.get().getMapView()) {
             for (final GeoitemLayer layer : layerList) {
                 geoEntries.remove(new GeoEntry(layer.getItemCode(), overlayId));
-                layers.remove(layer);
+                try {
+                    layers.remove(layer);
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/maps/mapsforge/v6/caches/AbstractCachesOverlay.java
```diff
@@ -296,7 +296,11 @@ protected void clearLayers() {
         synchronized (this.bundleRef.get().getMapView()) {
             for (final GeoitemLayer layer : layerList) {
                 geoEntries.remove(new GeoEntry(layer.getItemCode(), overlayId));
-                layers.remove(layer);
+                try {
+                    layers.remove(layer);
+                } catch (IllegalStateException ignore) {
+                    // ignore IllegalStateException("layer is not assigned")
```
eddiemuc, line 302: Maybe add Log on debug level:

`Log.d("Ignored exception on layer removal", ignore); `

moving-bits: updated PR accordingly

eddiemuc: great, thanks!

Issue #12813, opened 2022-02-28
Offline map opens slower if activity history is long


Lineflyer:
### Describe your problem!

Today I had a day off and went caching spontaneously driving from cache to cache.
After a while I noticed, that when "mapping a cache" it took some seconds for the map to show up. It got worse and worse over the day.

Back home I tried again and realized, that I could use the back key of my device quite a lot of time iterating over several map activites until finally reaching the state "press again to exit".

Once I did this opening the cache map was fast again.

### How to reproduce?

 I tried again to open list, home, map, cache, back to list, and so on and could (kind of) reproduce, that opening the map for a cache or even the "normal" map got slower to start again.

### Actual result after these steps?

If you have such long activity history, open the details of a cache and then long-click the compass rose to open the map for the cache.
The activity itself opens fast but it takes 2-3 seconds until map and objects are loaded.

### Expected result after these steps?

There is no crash involved, everything worked fine throughout the day (else the activity stack would have been shorter ;) ), but I guess we should limit the activity history somehow and/or find out if several open maps block/slow down something and whether this is avoidable.

Additionally:
IMHO it makes no sense to keep a real long activity history. Do we have set a limit for that?

### Reproducible

Yes

### c:geo Version

2022.02.26-NB

### System information

```text
Android 12, Samsung S20
```

moving-bits:
Shall we limit the activity stack to, let's say, 10 or 15 entries?
I guess no one will like to tap back even further (or just remember what one has done 15 activities earlier :-))

Lineflyer:
AFAIR we discussed this some weeks ago and I thought we already have such limit. If we do not have, it makes IMHO sense to limit to a rather low number like 10 at max.

fm-sys:
Yes, I remember that we had discussed this but until now no one has implemented it... 

10 activities sounds like a good starting point, maybe even 5 would be enough? OTOH, who knows how our users are using the app. With 10 activities we should be on the safe side (hopefully).

Side note: I have absolutely no idea how such a limit could be implemented cleanly. But maybe someone else has a good idea :)

moving-bits:
I've found no way yet to count every activity opened within the app, but I can count activities opened using the bottom navigation icons, so I implemented a slightly different approach:
- count number of activities opened using the bottom navigation icons; also remember the highest count reached
- on pressing the back key for such an activity, check whether counter's current value is farther than MAX_DEPTH from maximum so far
[...]

PR #12869, opened 2022-03-20
limit bottom navigation backstack to 5 (rel. to #12813)

moving-bits:
## Description
Limits the backstack of activities started using the bottom navigation to 5. On pressing the back key, when reaching this limit a toast is shown "Please touch again to exit", and if done so within the next two seconds, c:geo finishes. (Similar to what we have today when reaching the opening activity, but triggered by reaching the "max 5 activities" limit.)

## Additional context
I set the limit to five for now for easier testing. Before releasing this, we should probably increase the limit to ten.

A review would be highly appreciated.

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -66,7 +66,6 @@ public abstract class AbstractBottomNavigationActivity extends AbstractActionBar
 
 
     private ActivityBottomNavigationBinding binding = null;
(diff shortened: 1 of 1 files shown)


fm-sys:
Please see my comment in https://github.com/cgeo/cgeo/issues/12813#issuecomment-1073337134. This should be discussed before merging...

fm-sys:
I guess this PR is superseded by #13000 so we can close here, right?

Lineflyer:
Right. 
BTW: IMHO the usage flow works pretty well with the current solution!

Issue #12783, opened 2022-02-20
Allow upload of own spoiler images


eddiemuc:
### Tell us your idea!

I was recently approached by a geocacher who would like to have the possibility to add own, private images to a geocache (e.g. spoiler images and such) which should be kept on the cache but of course not be made public (e.g. via log).

When I heard this I thought about the existing c:geo feature to upload spoiler images in the already existing folder `GeocachePhotos`.  With this in mind it would be very easy to allow the user from within c:geo to add and remove own images to this structure.

### What solution would you suggest?

* Add the well-known "add image from gallery" / "add image from camera" to the top of the image list of a cache to select new images to add
* Mark own "spoiler images" somehow (e.g. by adding "Own image" to the title) since with this feature their usage may become more common
* Put a delete button next to the own images for deletion

The following is a crudely MS-painted screenshot on how this could look like (the upper image is supposed to be an image from the cache, the lower one an own image)

![image](https://user-images.githubusercontent.com/6909759/154841051-8b80d532-4140-4f24-97b4-9faf72306dda.png)

### Is there an alternative?

No implementation

Lineflyer:
Interesting feature!
Just thinking if it might make more sense to develop something to extend the personal note by something like "personal files" to allow more broadly attaching anything you want (e.g. a PDF, a track). Adding pictures could be a first step in there and it could be extended lateron.

It was also from time to time asked by users to have such "cache based storage" for files (and even a linkage to a cloud storage like Dropbox, GDrive was asked for). IMHO we have some issue here about that generic topic.

moving-bits:
Such a generic "add personal file to geocache" would be very welcomed. I guess we do not differentiate between images and other file types, just present them as a list and let Android decide on which app to open on tapping the file. (As for images I mostly open the gallery app anyway to be able to zoom.)

eddiemuc:
Hm extending this to handling of user-specific arbitrary files makes the thing a little more complicated of course. A couple of questions coming to my mind:
* Where and how would this be displayed. Possibilities are e.g.
   * As a separate new list under personal notes on "Details" tab
   *  Integrated into "Image" tab (then images from cache, user-specific images and user-specific non-images are displayed in the same place) with clickable previews for all supported content type (already there for images). Probably "Image" is then not the right name for the tab any more
   * New tab e.g. "Personal storage" beside the existing "Image" tab, but visually similar to it (clickable previews for all currently stored content esp images, but of course with delete/add functions) 
[...]

PR #12875, opened 2022-03-24
rel to #12783: change layout to allow margins to left and right of categories

eddiemuc:
rel to #12783: change layout to allow margins to left and right of categories


eddiemuc:
image

image

Very small screen:

image

PR #12893, opened 2022-04-06
Show correct waypoint type for generated waypoints. 

### LINKED ISSUE #12892: Wrong waypoint type for generated waypoints when editing
### Describe your problem!

When editing a generated waypoint, the waypoint type of the spinner shows "parking"

### How to reproduce?

Add a waypoint with calculated coordinates, which has a variable with a range.
Create waypoints from that.

Screenshot_20220406-213748_cgeo

Edit one of the auto-generated waypoints.

In the activity, the type for the waypoint is shown as "parking"
Screenshot_20220406-213730_cgeo

In the activity, the type for the waypoint should be shown as "generated"

### Reproducible

Yes

### c:geo Version

2022.04.01-NB-511d6f2

### System information

_No response_

### Additional Information

_No response_

murggel:
## Description

Show correct waypoint type for generated waypoints. 
Use the underlaying adapter to retrieve possible waypoint-types

## Related issues

fix #12892

### DIFF
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -68,7 +68,6 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
 import org.androidannotations.annotations.EActivity;
 import org.androidannotations.annotations.Extra;
 import org.androidannotations.annotations.InstanceState;
-import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 
 @EActivity
--- a/main/src/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/cgeo/geocaching/EditWaypointActivity.java
@@ -68,6 +68,7 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
 import org.androidannotations.annotations.EActivity;
 import org.androidannotations.annotations.Extra;
 import org.androidannotations.annotations.InstanceState;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 
 @EActivity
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -333,29 +333,29 @@ public void onNothingSelected(final AdapterView<?> parent) {
...
-    private int getDefaultWaypointType() {
+    private int getDefaultWaypointTypePosition(final ArrayAdapter<WaypointType> wpAdapter) {
         // potentially restore saved instance state
         if (waypointTypeSelectorPosition >= 0) {
             return waypointTypeSelectorPosition;
         }
 
         // when editing, use the existing type
         if (waypoint != null) {
-            return POSSIBLE_WAYPOINT_TYPES.indexOf(waypoint.getWaypointType());
```
eddiemuc, line 348: Where else is this "POSSIBLE_WAYPOINT_TYPES" constant used? Can it be deleted now? If not: might this change cause conflicts with places where it is still in use?

main/src/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -333,29 +333,29 @@ public void onNothingSelected(final AdapterView<?> parent) {
...
+            return wpAdapter.getPosition(waypoint.getWaypointType());
```
eddiemuc, line 348: What should happen if (for whatever reason) the waypoint type is not found in the adapter? Returning -1 ok?

eddiemuc (commented):
Some questions

main/src/cgeo/geocaching/EditWaypointActivity.java
murggel: The array is used to initialize the spinner, depending on wheter it is a generated waypoint POSSIBLE_WAYPOINT_TYPES or POSSIBLE_WAYPOINT_TYPES_WITH_GENERATED is used.

main/src/cgeo/geocaching/EditWaypointActivity.java
murggel: Good question - will show what the old return-value was and was the new one is.

eddiemuc: ok, then it is finer

main/src/cgeo/geocaching/EditWaypointActivity.java
eddiemuc: ok

Issue #12581, opened 2022-01-21
Raise minSDK to 24


Lineflyer:
I propose to raise minSDK to API24 in reasonable time.

Reasoning is, that we have some dependencies currently stuck to older versions due to incompatibility to low APIs.
Examples:
-  #12577 
- https://github.com/cgeo/cgeo/blob/master/main/build.gradle#L293-L296
- More?

IMHO its more reasonable to drop support, rather thean implementing ugly workarounds and own methods instead.

Maybe we can finalize the bottom navigation beta phase and then cut off a new legacy version.

I will check installation numbers on API21-24 as time permits, so we know how many users could probably be affected.

fm-sys:
Legacy versions are always ugly, as we don't really have the necessary resources to maintain such a separate version over the long run.

Workarounds can be really ugly, but they are IMHO still better than excluding thousands of users of keep using cgeo.

From my experience, devices running Android 4 (<API 21) were really raw. But this isn't the case for Android 5-7. I have an Android 6 (API 23) phone as fallback device still in use, and an Android 5 (API 21) tablet flying around which also still works like a charm.

And I know _many_ people who use an Android 7 (API 24) phone even as primary device.

But before me make any final decisions, let's see which picture the Playstore statistics will show about our user group...

moving-bits:
According to the stats presented in current Android Studio, API usage generally has the following distribution:

![image](https://user-images.githubusercontent.com/3754370/150585044-d59da024-6367-48c6-8d22-5ad3c083b983.png)

(According to a related 9to5google article, this data seems to be from about beginning of November 2021)

Is there any Play Store statistics specific to c:geo?

eddiemuc:
Here is the best I could find. This one is specific for c:geo. Only active installations (means: once in last 30 days) are counted.

![image](https://user-images.githubusercontent.com/6909759/150587355-0551f10c-69bb-463b-9e0d-58ebbd77e65c.png)

Lineflyer:
Thanks @eddiemuc , here the same as table with percentages and sum of percentages for the discussed API ranges:

![image](https://user-images.githubusercontent.com/949669/150639011-3f1b6e1a-2e01-4f30-be8b-9c73f97fc0d6.png)

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
[...]

Issue #12471, opened 2022-01-04
Add text and emoji for log smileys


bekuno:
### Tell us your idea!

To make selecting the right smiley easier show also the description.

### What solution would you suggest?

Instead:
![Screenshot_20220104-214354_cgeo](https://user-images.githubusercontent.com/32555742/148121709-3907d2b3-76da-4a13-bd23-0ae41b749d99.jpg)

You will have:
![grafik](https://user-images.githubusercontent.com/32555742/148121137-0037303f-39f9-4386-9e24-e199c605124c.png)

Lineflyer:
Strings are not localized though...that would be a bit more effort then :(

eddiemuc:
What about simply showing the resulting smileys as emojis? if there are copyright concerns in using the original ones from gc.com then may uaing the closest similar emoji available on Android?

ztNFny:
how about showing an actual smiley instead of the code? Unicode emoji should work everywhere

ztNFny:
Quickly looked up some emojis that would fit - didn't find a match for all but spend only a few minutes:
😃😁😎😳😝😈😉🤡?🎱
☹️?😲😠?😪😘??❓

![image](https://user-images.githubusercontent.com/1258173/148124834-50f704c4-feac-4f0f-afeb-205f44635470.png)

bekuno:
> 

Here you are:
![Screenshot_20220104-230320_cgeo](https://user-images.githubusercontent.com/32555742/148129975-cf7ae560-f74b-4cc6-9fe6-c0be832ce101.jpg)

Eldenroot:
Cool, anyway use unicoded smileys would be better

bekuno:
> Edit: Added more. For 😊😒 I picked similar ones. For the "black eye" one I couldn't find any - and I have no idea what that smiley is even supposed to mean.

Do you mean 🎱 ?
This is the eight ball from the pool billard, which is in the middle of 15 numbered balls.

How do you would integrate the unicode images - as image or text? What is to do to render unicode text as image?

ztNFny:
> Do you mean 🎱 ? 

No, the one before that which is missing from my list. It's called "black eye"

ztNFny:
> What is to do to render unicode text as image?

Nothing, just put the unicode char.

bekuno:
> No, the one before that which is missing from my list. It's called "black eye"

German: "Veilchen" :-)
[...]

PR #12897, opened 2022-04-10
Use downloadable fonts instead of bundled fonts to retrieve missing e…

### LINKED ISSUE #12856: FYI: APK size increased significantly
Some days ago I noticed, that the cgeo APK recently got significantly bigger.

Screenshot_20220313_215014_com android documentsui

It's not necessarily a problem, but 10 MB -> 20 MB (+100%) looks like a quite huge step to me. It would at least be interesting, what caused this rapid growth.

bekuno:
…moji

fixes #12856

This should work, as all samples I found in the web are work in this way.
But the creation of the FontRequest (EmojiDownloadService Line 37) fails with 
`Method threw 'java.lang.NullPointerException' exception. Cannot evaluate androidx.core.provider.FontRequest.toString()`

Does anyone have a solution?


bekuno:
Maybe someone is able to test this on a API 21 (Android 5.0/Lollipop) device?

Lineflyer:
> Maybe someone is able to test this on a API 21 (Android 5.0/Lollipop) device?

I could test on an API 21 emulator if this helps. I might have a API 21 real device somewhere, but this would be more easy to test it in nightlies.

bekuno:
> > Maybe someone is able to test this on a API 21 (Android 5.0/Lollipop) device?
> 
> I could test on an API 21 emulator if this helps. I might have a API 21 real device somewhere, but this would be more easy to test it in nightlies.

We tested yesterday with @ztNFny and emulator. He could reproduce the problem.
The first API with all emoji is API 24.
So it would helpful to test on a real device lower API 24.

As this is probaly a rare used function and only a small percentage of user devices are involved, we could maybe stay without a emoji - the missing emoji is shown in https://github.com/cgeo/cgeo/issues/12471#issuecomment-1007813268

Lineflyer:
Without having it tested:

I think we should just accept the limitation to show no emojis / not all emojis on API <24.

What we should reach is, that the APK size is reduced again...IMHO much more important than full feature coverage on pretty old devices.

ztNFny:
It seems there's not really progress on this. Can we remove the bundled font please to get the size back down? Those old devices will just have to miss out on this feature.

fm-sys:
The question is whether we need it at all...? Especially as we had the discussion to increase the min API version recently (#12581)

bekuno:
We would need at least one other emoji for Blay Eye and Shy, as they are still missing until API 27.

ztNFny:
> We would need at least one other emoji for Blay Eye and Shy, as they are still missing until API 27.

imho we can ignore API <27 (just show the generic placeholder icon for those smileys). That's just 13% of devices and the loss in functionality is almost non-existent.

bekuno:
> > We would need at least one other emoji for Blay Eye and Shy, as they are still missing until API 27.
 
> imho we can ignore API <27 (just show the generic placeholder icon for those smileys). That's just 13% of devices and the loss in functionality is almost non-existe

Inclusive API 27 :-).
But these are only 2% more ...

ztNFny:
> Inclusive API 27 :-).
> But these are only 2% more ...

15% of devices, a feature that few people will use, breaking in a way that's nearly irrelevant (there's still the character and the text version) - I say we shouldn't take care about at all if it takes more than 15 minutes but instead just ignore those without any workarounds or if (API.LEVEL >...)

Lineflyer:
Replaced by https://github.com/cgeo/cgeo/pull/12935

PR #12921, opened 2022-04-24
multiple choice questions

baiti:
as discussed someplace else, pls review


ztNFny (approved):

moving-bits (approved):
LGTM, Thanks for diving into it so quickly!

PR #12922, opened 2022-04-24
catch some exceptions on JSON transformation

moving-bits:
## Description
catch some exceptions on JSON transformation


baiti (approved):
LGTM ...
and you fixed my `<ul> ... </ul> `placement also ;)

Issue #7817, opened 2019-09-01
Add "location" to log template menu


SammysHP:
In the logging activity you can insert various elements like date, time, number, owner name etc. Sometimes the log requires to post a location. I suggest to add another entry to add the current position, maybe with the help of the coordinates popup to modify them and optionally select a position on the map.

spacemonkey1101:
Hello There! 
I think I would be able to do this with some help.

fm-sys:
@spacemonkey1101  Very Wellcome!

You can use `Sensors.getInstance().currentGeo().getCoords()` `from cgeo.geocaching.sensors.Sensors` to get the current coordinates. For adding a new log template you can take a look at [#9064](https://github.com/cgeo/cgeo/pull/9064/files)

eddiemuc:
I hope users are turning off this feature actively when they log a mystery cache while still nearby it...

fm-sys:
Maybe we can hide this template in the signature settings and only show it in the drop down list while logging

fm-sys:
Otherwise, if we want to use the coordinate popup (which would be a really nice feature) it will not work with the log template architecture.

>and optionally select a position on the map.

Btw this is tracked here:  #7811

Lineflyer:
> I hope users are turning off this feature actively when they log a mystery cache while still nearby it...

I doubt that many users wil use such feature by default in their signature.
Furthermore its anyway users responsibility not to spoil in their logs. No need for us to block any function.

sophiastr:
i want to deal with this issue.
I am interested in contributing to the project you are developing.
Do you still need help?
I'm a beginner but I will try really hard.

eddiemuc:
Hi  @sophiastr , welcome to c:geo. We are always thankful for new contributors, so it is great that you want to get involved!

The first step would be to set up your local development environment. How to do this is described in the readme page (https://github.com/cgeo/cgeo#readme). If you plan to make a contribution, make sure to fork the cgeo repo, then clone your fork to your local machine (do not clone cgeo repo directly to your local machine). You should come as far as compiling and running cgeo from branch master as-is in an android phone emulator (those come with the IDE, Android Studio).

[...]

PR #12932, opened 2022-04-25
Feature/7817 add location to log template menu

nmrsmn:
## Description
Added the [LOCATION] placeholder to add to signatures, log templates and log entries.

## Related issues
#7817

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -815,6 +815,7 @@
     <string name="init_signature_template_log">Log text</string>
     <string name="init_signature_template_size">Size</string>
     <string name="init_signature_template_type">Type</string>
+    <string name="init_signature_template_location">Location</string>
     <string name="init_ratingwanted">GCvote Rating</string>
     <string name="init_summary_ratingwanted">Load cache rating from GCvote.com</string>
     <string name="init_friends_and_own_logs_wanted">Show Friends\'/Own</string>
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -324,18 +324,6 @@ public class LogCacheActivity extends AbstractLoggingActivity {
         super.onPause();
 
     }
-    /*
-    super.onResume();
-        // resume location access
-        PermissionHandler.executeIfLocationPermissionGranted(getActivity(),
-                new RestartLocationPermissionGrantedCallback(PermissionRequestContext.AbstractDialogFragment) {
-
-                    @Override
-                    public void executeAfter() {
-                        resumeDisposables.add(geoUpdate.start(GeoDirHandler.UPDATE_GEODATA));
-                    }
-                });
-     */
 
     private void setLogText(final String newText) {
         binding.log.setText(newText == null ? StringUtils.EMPTY : newText);
--- a/main/src/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/cgeo/geocaching/log/LogCacheActivity.java
@@ -22,6 +22,11 @@ import cgeo.geocaching.enumerations.StatusCode;
 import cgeo.geocaching.log.LogTemplateProvider.LogContext;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
+import cgeo.geocaching.permission.PermissionHandler;
+import cgeo.geocaching.permission.PermissionRequestContext;
+import cgeo.geocaching.permission.RestartLocationPermissionGrantedCallback;
+import cgeo.geocaching.sensors.GeoData;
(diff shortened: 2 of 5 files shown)


Lineflyer:
ok to test

nmrsmn:
> Usually c:geo holds a GPS lock only if needed (e.g. on map). Did you test how this behaves when the location is stale? E.g. open map to get location, then open the log dialog, move 100m -> what do you get?
> Or start c:geo, select cache from list view without going to map, how accurate is the location?

I’ve killed the app, started it back up, went to my stored caches and tried to log one of the stored caches. The location was correct and spot on for me. I’ve also tested it using the nearby function. 

I haven’t tested opening a log dialog and then move, but what do you expect to happen? Do you expect the log message to update when you move? If so did you consider people that are logging as they move to the next one? If they type slow or they edit their message, which takes some time, the location might change and isn’t the right location anymore.

ztNFny:
> I haven’t tested opening a log dialog and then move, but what do you expect to happen?

I'd expect it uses the coordinate that's "current" when pressing the button. And "current" in this case is probably "last known" - so it might be stale

>  Do you expect the log message to update when you move?

Of course once inserted nothing must change.

ztNFny:
btw, maybe it also makes sense to show such a dialog to be able to select a coordinate format? Currently it always returns it as DMM (using GeopointFormatter.Format.LAT_LON_DECMINUTE)

nmrsmn:
@ztNFny I'm not sure how to handle the formatting dialog in case you added the `[LOCATION]` placeholder to your signature for instance (or even when you've added it to a log template in settings), since those placeholders are being replaced upon starting a new log visit. 

A dialog is technically possible when selecting the location placeholder from within the `LogCacheActivity`, but at this moment all the `LogTemplate` implementations need to return a `String` used replacement that is being applied upon clicking that item, so at this moment a dialog isn't possible. I'm confident we can work something out, but is this still in the scope of this "beginner" issue then? Isn't this a good "first" step on which we can expand upon? 

Also, the `LogCacheActivity` isn't aware of any location updates like the `AbstractDialogFragment` is. I'll add that to make the location updates work and prevent the user from adding stale location.

ztNFny:
>  I'm confident we can work something out, but is this still in the scope of this "beginner" issue then?

I'm actually quite confused this was marked as a "good first issue". Imho it's quite a complex thing due to coordinate formats, figuring out what's the right location and possible delays.

> Also, the LogCacheActivity isn't aware of any location updates like the AbstractDialogFragment is.

I wonder what the battery impact of that is. But then again not too much time is spent on the log dialog probably ...

>  I'm not sure how to handle the formatting dialog in case you added the [LOCATION] placeholder to your signature for instance

Is that really the usecase though? Only use I see for including location in logs if the final location is off and you want to provide that info to others. It should probably *explicitly NOT* be available for inclusion in templates/signature - otherwise people will be posting the final locations of multis/mysterys.

nmrsmn:
> I wonder what the battery impact of that is. But then again not too much time is spent on the log dialog probably ...

I was wondering the same but I agree on your comment on not spending that much time on the log dialog, so I figured this was an OK solution..

> Is that really the usecase though? Only use I see for including location in logs if the final location is off and you want to provide that info to others. It should probably _explicitly NOT_ be available for inclusion in templates/signature - otherwise people will be posting the final locations of multis/mysterys.

Now I'm getting confused. My first thought is that it wouldn't be that hard to make the `[LOCATION]` replacement available only for when you're on the `LogCacheActivity` (maybe even restrict it to when there is no final location available like for multis/mysteres). But as of now I'm really unsure what I should implement, since all the stuff spoken about in the issue itself.. So maybe someone can shed some light on that matter, like @Lineflyer or @eddiemuc?

Lineflyer:
> I'm actually quite confused this was marked as a "good first issue". Imho it's quite a complex thing due to coordinate formats, figuring out what's the right location and possible delays.

Assuming that this template is only actively inserted while currently composing the log its IMHO sufficient to take what is delivered as current coordinate and insert is as DD MM.SSS as common geocaching format.

ztNFny:
> My first thought is that it wouldn't be that hard to make the `[LOCATION]` replacement available only for when you're on the `LogCacheActivity`

That can be achieved by adding it to LogTemplateProvider.getTemplatesWithSignature() instead of getTemplatesWithoutSignature()

nmrsmn:
Well in that case, if I understood correctly, I implemented the feature now as intended.

ztNFny:
How about also inserting the accuracy? Like "N 00 00.00 E 000 00.000 (±10 m)"
See https://github.com/cgeo/cgeo/pull/12934 for getting it

moving-bits:
Let's test this in the nightlies. Thanks @nmrsmn for your contribution, very welcomed!

PR #12935, opened 2022-04-25
remove emoji compat (bundled)

### LINKED ISSUE #12856: FYI: APK size increased significantly

bekuno:
fixes #12856

Removal of complete emoji compat Implementation.
Disadvantage: Up to API 27, two emoji are without a picture.
2022-04-25 22_17_05-cgeo – build gradle (_main) 2

This is a alternative implementation compared to #12897.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -415,11 +415,6 @@ dependencies {
     implementation "androidx.annotation:annotation:$annotationVersion"
     androidTestImplementation "androidx.annotation:annotation:$annotationVersion"
 
-    // emoji compat support
-    def emojiVersion = '1.1.0'
-    implementation "androidx.emoji:emoji:$emojiVersion"
-    implementation "androidx.emoji:emoji-bundled:$emojiVersion"
-
     // Testing Support Libraries
     androidTestImplementation 'androidx.test.ext:junit:1.1.3'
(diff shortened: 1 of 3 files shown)


ztNFny:
@Lineflyer should be mentioned in known issues in FAQ

Lineflyer:
@bekuno 
I think this is better and not increasing the APK size, the two missing emojis are acceptable IMHO.
Shall we merge to test in nightlies or is there any check to be done in advance?

@ztNFny 
Do you really think this is worth it? Up to now there are no visual emojis at all, I don't think many users will recognize the two missing ones.

ztNFny:
> @ztNFny
> Do you really think this is worth it? Up to now there are no visual emojis at all, I don't think many users will recognize the two missing ones.

Maybe let's wait if support gets questions about it?

bekuno:
> Shall we merge to test in nightlies or is there any check to be done in advance?

It is ready to merge.
A change of the problematic emoji can be done later if someone has an alternative.

PR #12942, opened 2022-04-30
add FileInformation cache to NodesCache (rel. to #12881)

moving-bits:
## Description
caches SAF file info in `NodesCache` for one minute to speed up calculation of route segments


eddiemuc:
I honestly don't think it is a good idea to mix in this function-specifix caching with the globally used class ContentStorage. Other usages might rely on this function to deliver most current info regardless of slowness.

I would rather propose a global (e.g. static) Cache for this information to be included with NodesCache.

moving-bits:
You're probably right, I'll move this to `NodesCache`

moving-bits:
removed the additional logging & changed caching to be `NodesCache`-local

PR #12948, opened 2022-05-01
fix #12912: add entry 'log in browser' to cache menu

### LINKED ISSUE #12912: Include link "Create new log in browser" (pre-solution for full support of multiple log profiles)
### Tell us your idea!

I often want to log caches I found for other, non-premium accounts too (those of my kids). I usually do this by logging in into my browser as the other account. Then I log for myself using c:geo, then I open the same cache in Browser using c:Geo's "open in browser" action and create a log there for the other account.

This workflow forces me to do some more clicks than I like, but it works ok-ish for non-premium caches. For premium caches, however, it doesn't work very well, because I have to do URL magic for the other, non-premium account to be able to log in browser.

The big solution for this would be to enable "log profiles" in c:geo - some sort of auto-logging feature for "secondary accounts" which could be enabled in the log entry activity. However, this issue is about something else.

My request would simply be to include a link "Create new Log in Browser" next to the already existing "Open in Browser" action. This action would open a new log entry screen in the Browser directly. This should also work if another (non-premium) account is currently logged in into the browser)

### Is there an alternative?

Full-blown log profile solution.
Or do nothing of course

### Additional Information

_No response_

eddiemuc:
fix #12912: add entry 'log in browser' to cache menu

### DIFF
--- a/main/res/menu/cache_options.xml
+++ b/main/res/menu/cache_options.xml
@@ -87,6 +87,12 @@
         app:showAsAction="ifRoom">
     </item>
     <item
+        android:id="@+id/menu_log_in_browser"
+        android:title="@string/cache_menu_log_in_browser"
+        android:visible="false"
+        app:showAsAction="ifRoom">
+    </item>
+    <item
         android:id="@+id/menu_edit_fieldnote"
         android:icon="@drawable/ic_menu_edit"
         android:title="@string/cache_personal_note"
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1392,6 +1392,7 @@
     <string name="cache_menu_vote">Vote</string>
     <string name="cache_menu_checker">Open Geochecker</string>
     <string name="cache_menu_challenge_checker">Search challenge checker</string>
+    <string name="cache_menu_log_in_browser">Log in browser</string>
     <string name="cache_menu_ignore">Ignore cache</string>
     <string name="cache_log_menu_decrypt">Decrypt/Encrypt</string>
     <string name="cache_log_menu_popup_title">Log of %s</string>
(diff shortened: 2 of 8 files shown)


ztNFny:
should this be in the log activity options menu instead maybe?

eddiemuc:
> should this be in the log activity options menu instead maybe?

That would be bad. It would mean that each time I want to call this c:geo would create a new dummy log which I then have to empty.

Entry should be in same place as "Log visit IMHO

ztNFny:
But isn't your usecase that you want to log from a second account (using the browser) in addition to the log from c:geo? So you always will have the c:geo log, which means you already are in that screen.

My concern is that the cache options menu is already very cluttered with lots of corner-case options and this is adding yet another one of those.

eddiemuc:
> But isn't your usecase that you want to log from a second account (using the browser) in addition to the log from c:geo? So you always will have the c:geo log, which means you already are in that screen.

I want to have my log first in the cache, then the secondary ones. My workflow is usually that I log first for myself (includes sending the log), then log for my kids. Thus at this point I am not anymore in the log activity. 

I could of course change my workflow: open log activity (for me), open browser immediately (for kid), then switch back to c:geo to log for myself, then switch back to already opened browser window to log for the kid. Well if you force me to do so I will. But does it really help?...

> 
> My concern is that the cache options menu is already very cluttered with lots of corner-case options and this is adding yet another one of those.

Full agreement. But the log menu is crowded as well (granted, not as much as the cache log, but still...). However, the first thing I would clean up there are all those actions which are also available in the cache pages (or could be accessed there much more meaningful), e.g. editing personal info or calling geochecker. This might be worth a separate issue though.

Issue #12947, opened 2022-05-01
Introduce Resize Factors for map font/lines/symbols


MagpieFourtyTwo:
### Tell us your idea!

In OSM map elements, esp. text is quite hard to read (i. e. really small) on high resolution displays, esp. if you use Elevate5 as map theme.  Therefore, I have written a script which increases fonts, lines and symbols by a configurable factor, which massively increases usability.

The other day I learned two things:
1. I am not the only one with this "problem". 
2. Locus has a setting for this, in fact, to increase all map elements by a configurable factor ...

... what gave me the idea that c:geo could perhaps offer this functionality, too.  ;)

### What solution would you suggest?

From my experience, it turned out to be best to have three factors:

1. Font Size
2. Line Width
3. Symbol Size

... as in fact I like the theme best, if fonts are increased to 160, lines to 120 and symbols to 130 percent,
while my teammate prefers F140/L120/S110 and a friend of ours even wanted to have F180/L120/S130.
Although my script can create all three in one go, life would be way easier, if c:geo would support to do this itself.  ;)

### Is there an alternative?

Just write yourself a script to increase sizes ...  ;)

### Additional Information

As an example here a comparison between original (in fact close to unreadable) vs. "my" F160, L120 S130:

![Current](https://user-images.githubusercontent.com/9889015/166129743-99b5491b-2d88-42ca-9475-952389318942.jpg) ![Font 160 Lines 120 Symbols 130](https://user-images.githubusercontent.com/9889015/166129750-7206c548-d96b-436b-b068-17b710c58b58.jpg)

eddiemuc:
@MagpieFourtyTwo in the meantime (until this is hopefully implemented) I would be interested in your script. Are you able to share it? What mechanism does it use to set these sizes?

MagpieFourtyTwo:
It once started as a quick'n dirty pure windows shell script, in fact a 5-liner, IIRC, but although it got a bit more elaborated (like e.  g. integrating some vbscript, cause cmd's "set" can just handle integer operations) over time (I use it for a couple of years now), and even a bit more user friendly (it's pretty straight forward, if you know what you're doing, and you can even generate a whole bunch of different settings combinations (F/L/S) in one go), but it's still no PhD script, as it sometimes still needs a bit of fine tuning (due to some performance hacks) and therefore some scripting knowledge, thus I would not give it to public in it's current state.

[...]

PR #12950, opened 2022-05-01
rel to #12947: add on-the-fly-replacement for xml attributes

eddiemuc:
rel to #12947: add on-the-fly-replacement for xml attributes


eddiemuc:
This PR adds a hidden setting to c:geo called `rendertheme_scale`. It can only be accessed through System Settings -> View/Edit Properties, no user-friendly input area available.

image

Expected value are three numeric factors, divided by `-` (in screenshot example above the factors are 3, 4 and 5). If given, the first value will be used to scale font size, the sceond to scale stroke-width and the third to scale symbol sizes.

Rendering example for value `3-4-5` (means: font size increased by factor 3, stroke-widths increased by factor 4, symbol sizes increased by factor 5)

image

eddiemuc:
This PR is no longer meaningful/needed

PR #12956, opened 2022-05-02
fix #12947: add setting to manipulate mapsforge scaling settings

### LINKED ISSUE #12947: Introduce Resize Factors for map font/lines/symbols

eddiemuc:
fix #12947: add setting to manipulate mapsforge scaling settings

### DIFF
--- a/.idea/checkstyle-idea.xml
+++ b/.idea/checkstyle-idea.xml
@@ -3,12 +3,12 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
+        <entry key="active-configuration-0" value="PROJECT_RELATIVE:C:/private/geocaching/github/cgeo/checkstyle.xml:cgeo" />
         <entry key="checkstyle-version" value="8.44" />
         <entry key="copy-libs" value="false" />
-        <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
-        <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
-        <entry key="location-2" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
+        <entry key="location-0" value="BUNDLED:(bundled):Sun Checks;All" />
+        <entry key="location-1" value="BUNDLED:(bundled):Google Checks;All" />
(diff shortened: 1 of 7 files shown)


eddiemuc:
This PR will add experimental  functionality for DEVELOPMENT and NIGHTLY only!

It will add the following setting in the map theme options dialog:

image

The setting allows to enter up to three float or percentage values separated by `:`. If given those values will be filled into the following mapsforge scaling options:
* User Scaling: a general scaling factor which up- or downscales the whole map (technically by manipulating the tile size in pixels). This will influence storke width, text sizes and symbol sizes alike
* Text Size scaling: a scaling factor for text size alone (will be applied on top of user scaling)
* Symbol size scaling: a scaling factor fort symbols aline (will be applied on top of user scaling)

For example, given a value of `100%:200%:500%`, the map will look something like this:

image

eddiemuc:
retest this please

eddiemuc:
PR is now ready for merge:

image

moving-bits:
> PR is now ready for merge:

Thanks for implementing this nice feature, looks good!

Just a tiny note: Do you mean "scale symbol sizes" or "scale symbols"?

eddiemuc:
> Just a tiny note: Do you mean "scale symbol sizes" or "scale symbols"?

Ah, I will correct this. Thanks for asking.

eddiemuc:
Update: allow scaling settings also for default theme, make percentage values selectable only every 10%

Issue #12958, opened 2022-05-02
lettervalue('ö')=0


geocachermgo:
### Describe your problem!

The letter value in the formulas for calcutated coordinates of umlauts are zero. Normally äüöß are converted to 27...30 or sometimes to ä=ae=1,5. Setting them to 0 is wrong.

### How to reproduce?

Create a variable in with lettervalue('äöüß')

### Actual result after these steps?

Value is 0

### Expected result after these steps?

Value should be 114

### Reproducible

Yes

### c:geo Version

2022.05.02-NB1-53b5e84

### Additional Information

At least a warning (red marked letters, if not converted ?) should appear.
é is also converted to 0 in c:geo
Gcwizard translates this to 5 in the selection "Französisch" and to 0 in the selection "Deutsch"

eddiemuc:
PR #12959 will handle german umlauts.

For the other two aspects of the issue:
* warning (making letters red) is very hard to achieve since it needs to happen in the input field (string not visible in output). never done before, is this really necessary?
* handling things like `é` in a generic way seems to be hard. I searched a bit for a method to get the "base letter" of a character (e.g. a method which gives me `e` when I pass `é` to it, `u` when I pass `û` to it and `C` when I pass `Ç` to it). Does anyone know about something like this?
   * If not: we could maintain an own list of these characters, is this wanted?

fm-sys:
> If not: we could maintain an own list of these characters, is this wanted?

The android keyboard might be helpful for this... At least [SwiftKey](https://play.google.com/store/apps/details?id=com.touchtype.swiftkey&hl=de&gl=US) shows all character variants which exist.

![Screenshot_20220503_174052_copy_431x423](https://user-images.githubusercontent.com/64581222/166487915-ff883cc5-baef-4d73-9e6c-3340e9e8b355.jpg)

Or this might be helpful as well:
https://stackoverflow.com/questions/3322152/is-there-a-way-to-get-rid-of-accents-and-convert-a-whole-string-to-regular-lette

moving-bits:
To be honest: I would not invest too much into extending character conversion. 99.9% of all lettervalues I've seen while out geocaching in different countries use A-Z = 1-26 (or maybe the other way around, but even that's rare). Umlauts are mostly ä=ae etc. Everything else get's owner-specific pretty fast, and we don't want to have to catch up with that, including complicating the UI by that.

So it's pretty useful that we have the most common variant implemented, but I would leave it at that.

[...]

PR #12959, opened 2022-05-03
rel to #12958: handle german umlauts in lettervalue

eddiemuc:
rel to #12958: handle german umlauts in lettervalue

This PR, if merged, will assign values 27-30 to german umlats äöüß in lettervalue

### DIFF
--- a/main/src/cgeo/geocaching/utils/formulas/FormulaUtils.java
+++ b/main/src/cgeo/geocaching/utils/formulas/FormulaUtils.java
@@ -7,13 +7,17 @@ import android.util.Pair;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * Holds implementations for functions in Formula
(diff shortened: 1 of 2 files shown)


ztNFny:
There's 2 common cases of handling these:
- ä = 27
- ä = ae = 6
Is it possible to provide a switch between these? To me it feels the ae case is more common.
GCWIzard has a dropdown where you can select different "styles" of conversion

eddiemuc:
Let's not forget that we talk about the lettervalue function used in variable expression and nothing else. Some time ago we did decide that those functions shall not replace fulll-blown GC-Helper-Apps like GCWizard but rather provide a basic set of functionality used most often. I took GCBuddy as reference, and here ä is always interpreted as 27 (not as ae).

My personal opinion: if someone wants text to be interpreted as `ae`, why doesn't he simply write ae instead of `ä`? 

image


If it seems necessary we can make lettervalue more adjustable. Several options are available for this:
* Option 1: create functions with distinct names to distinguish between different calculation methods e.g.:
    * lettervalue('ächz') interprets ä as 27
    * lettervalueSUbstituteUmlauts('ächz') interprets ä as 6
 * Option 2: put a second parameter to the function so user can select the calculation mode:
    * lettervalue('ächz', Mode.NORMAL) interprets ä as 27
    * lettervalue('ächz', Mode.SUBSTITUTE_UMLAUTS) interprets ä as 6

(OPtion 2 is probably more flexible but not easy because the expression language does not support enums yet)

Details need to be clarified: which options and combinations are supported etc etc. Probably worth its own issue and maybe nothing for release branch I would say?

ztNFny:
I took a closer look at the options GCWizard provides. The language options can't even agree on letters A-Z being 1-26 (polish, greek, russian) 😵
As we don't want to implement such a plethora of options I see only 2 realistic ways:
- Implement only the "base" version
- Implement whatever the devs want (seeing as most are german that's probably german then) but be clear that we don't want to offer a "full" solution

fm-sys:
http://www.buchstabenwortwert.de/ handles ß=30 BTW. However, I'm not sure if we should dive into such owner specific things at all...

moving-bits:
I would leave it with this base version currently implemented in this PR, see https://github.com/cgeo/cgeo/issues/12958#issuecomment-1116275047 for my 2cents on this... - no extra parameters to those functions, no alternative functions for lettervalue etc., please.

PR #12965, opened 2022-05-05
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/branch-merge.yml
+++ b/.github/workflows/branch-merge.yml
@@ -13,9 +13,14 @@ env:
   RELEASE_BRANCH: release
   MASTER_BRANCH: master
 
+permissions:
+  contents: read
+
 jobs:
   # This workflow contains a single job called "merge-branches"
   merge-branches:
+    permissions:
+      contents: write  # for Git to git push
     runs-on: ubuntu-latest
 
     steps:


fm-sys:
The workflow permissions are only relevant for the `GITHUB_TOKEN`, which we do not use in the workflow at all. We use a separate restricted `ACTION_SECRET` instead, so we could set the `GITHUB_TOKEN` permissions to `{}`. 

However, the security risk described in the link above is only relevant if unknown code is executed. This cannot happen as the workflow is never executed for unchecked PRs.

naveensrinivasan:
> The workflow permissions are only relevant for the `GITHUB_TOKEN`, which we do not use in the workflow at all. We use a separate restricted `ACTION_SECRET` instead, so we could set the `GITHUB_TOKEN` permissions to `{}`.
> 
> However, the security risk described in the link above is only relevant if unknown code is executed. This cannot happen as the workflow is never executed for unchecked PRs.

Ok. Sounds good. Thanks

PR #12971, opened 2022-05-07
rel to #12783: add endless loop detail view and jump on back

eddiemuc:
rel to #12783: add endless loop detail view and jump on back


eddiemuc:
Retest this please

eddiemuc:
Failing test is unrelated (relates to geolutions which seems to be unavailable currently), so let's merge

PR #12974, opened 2022-05-08
fix #12972: scale scale bar together with map user scale

### LINKED ISSUE #12972: Map scaling: Scale bar needs to scale as well
### Describe your problem!

When setting map scaling to 200%, leaving the other two scaling options unchanged, the whole map scales at factor 2, which is to be expected. Only the scale bar in the lower left corner does not change, therefore indicating twice the size:

|unscaled|map scaled to 200%|
|---|---|
|image|image|

(See the "Berliner Schloss" in the lower left part of the map)

### How to reproduce?

- open Mapsforge offline map
- set theme option to "scale map=200%", "scale text=100%", "scale symbols=100%"

according to the scale bar all streets, buildings etc. have doubled length/width

the scale bar needs to adjust its values according to map scale or needs to scale itself by the same factor

### Reproducible

Yes

### c:geo Version

current nightly

### System information

_No response_

### Additional Information

VTM scales the whole scale bar in this case:

image

(not sure if I like that one, though - scale bar gets quite clumsy...)

eddiemuc:
fix #12972: scale scale bar together with map user scale

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/NewMap.java
@@ -712,6 +712,9 @@ public class NewMap extends AbstractBottomNavigationActivity implements Observer
             layers.add(index, newLayer.getTileLayer());
             this.tileLayer = newLayer;
             this.renderThemeHelper.reapplyMapTheme(this.tileLayer, this.tileCache);
+            //trigger mapscalebar redraw - otherwise the shown distances will be wrong.
+            //See https://github.com/mapsforge/mapsforge/discussions/1313
+            this.mapView.getMapScaleBar().setDistanceUnitAdapter(this.mapView.getMapScaleBar().getDistanceUnitAdapter());
             this.tileLayer.onResume();
         } else {
             this.tileLayer = null;


eddiemuc:
Apparently Mapsforge Map Scale bar does not use the DisplayModel User Scale Factor . I worked around this by applying an own (delegating) DistanceUnitAdapter.

eddiemuc:
Update: found much easier way. MapScaleBar simply needs to be triggered to redraw itself

moving-bits:
Thanks for diving into this. Have not yet tested the new approach, but was able to get somewhat strange results with the first approach:

|without PR / 100%|without PR / 200%|
|---|---|
|image|image|

|with PR / 100%|with PR / 200%|
|---|---|
|image|image|

Mark the scale colorings (red=100m / blue=50m)

So without the PR the horizontal length is about 250m (which Google Map confirms) in 100%, but 500m in 200%.
Using the PR it shrinks from 250m down to 125m :-(

Will test with updated PR next.

eddiemuc:
Yes, it turned out that MapScaleBar (contrary to what I thought) DOES consider the user scale factor, but only after a map restart. So since I applied own scaling, after a map restart the scaling of map scale bar would be done two times - resulting in a differently wrong result of course.

With the changed approach this should not happen. Although mapsforge itself claims that a full map restart (actually he even recommends an application restart) is necessary after user scaling option is changed. This would however not fit into c:geo workflow at all. So let's hope it works with the fix.

eddiemuc:
Failing test in CI build is unrelated. @moving-bits should I merge?

moving-bits (approved):
Looks fine now / does work now as intended. Thanks for tackling this.

PR #12982, opened 2022-05-08
Localize quick launch button settings

Lineflyer:
## Description

Just noticed today, that the settings for the quick launch buttons are not localized but still hardcoded.
This PR fixes it.

## Related issues

None

## Additional context

Targeted to `release` to have a chance of fixing before we release current beta.


moving-bits (approved):
LGTM, thanks for translating it

PR #12983, opened 2022-05-08
Extend code formatting configuration

UniQP:
The goal of the PR is to provide a code formatting configuration that matches our current Java code style as close as possible.

To do so, it extends the .editorconfig by some configurations and fixes some code formatting problems with obvious solutions. I will open another PR where all Java files are formatted for reference.

PR #12984, opened 2022-05-08
Code formatting example

UniQP:
The PR formats all Java files with the code formatting configuration provided by #12983. It is intended as a reference and should not be merged.


UniQP:
Integrated in #12983

PR #12998, opened 2022-05-10
add option to not ROT13 the hint

ztNFny:
image


Narrat:
As the word clicked for me primarily is associated with a respective pointer device like a mouse, shouldn't it replaced with tapped (if that is the right form/word)? Which is more commonly used as a word for interacting with touch devices?

Lineflyer:
Yep, we agreed on "tap" as wording some time ago.

Regarding the feature:
Believe it or not, but sometimes we even get support tickets, that the hint "is broken". :)
I am fine with this feature as long as the default is enabled.

ztNFny:
Will change that tomorrow.

Should this also disable the "tap to en/decrypt" on hint or just change the default?

Lineflyer:
I guess it should disable the tap on the hint function if active.

ztNFny:
> 

Done

Lineflyer:
Any objections, else will merge tomorrow.

PR #13000, opened 2022-05-10
Clear activity stack when navigating via the bottom navigation (fix #12813)

### LINKED ISSUE #12813: Offline map opens slower if activity history is long

fm-sys:
As agreed in #12813:
- When navigating via the bottom navigation, the whole activity stack will be cleared
- When the current activity is the task root, pressing back will navigate to the dashboard

### DIFF
--- a/main/src/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/cgeo/geocaching/activity/AbstractActivity.java
@@ -25,7 +25,10 @@ import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.TranslationUtils;
 import cgeo.geocaching.utils.formulas.FormulaUtils;
 
+import android.content.BroadcastReceiver;
+import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
 import android.content.res.Resources;
 import android.os.Bundle;
 import android.util.AndroidRuntimeException;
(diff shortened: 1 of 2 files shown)


Lineflyer:
Lets test the behavior.

PR #13003, opened 2022-05-11
rel to #12783: make transitions local to image gallery activities, add image count in tab

eddiemuc:
rel to #12783: make transitions local to image gallery activities, add image count in tab


eddiemuc:
image

PR #13004, opened 2022-05-12
Support building nested filters

### LINKED ISSUE #11275: Add nested filters
What about opening a sub-filter screen, to allow building of nested filters via UI?

Screenshot_20210719_225138_copy_540x671

The user then can just add a "sub-filter" filter and then can do everything with it. Load a saved configuration, create and adopt the sub-filter, etc, etc...

I guess this would be the cleanest and feature richest solution, where most code parts AFAIK even could be reused without much effort.

There are very legit use cases for it: e.g. I want to display only traditionals OR caches which are contained in a specific list but what to hide all found ones. A filter which is currently not possible to build.

We could give users with the "sub-filter" filter the same possibilities as with the actionbar icon in maps viewer. Short click -> open filter screen | Long click -> stored filter selection.

Feature should go to `master` 

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/10581#issuecomment-882854827_ ff.

### LINKED ISSUE #12090: Filter setting: "Always show stored caches" even when they don't meet the filter settings
### Tell us your idea!

The new filter settings are great. However, when I voluntarily stored a cache on my device, that's because I *definitely* want to see it, no matter how the filter settings are currently set up.

In the filter settings, there should be a checkbox "Always show stored caches" which, when enabled, has the effect that stored caches are always visible, no matter if they meet the current filter criteria.

### Is there an alternative?

_No response_

### Additional Information

_No response_

fm-sys:
Advantage: Users can now combine `AND` and `OR` within one filter.

- Fix #11275
- Fixes #12090 implicitly as well

|Normal "toplevel" filter|Sub-filter|
-|-
grafik|grafik

This example will evaluate to `([:inconclusive=false:advanced=true]AND(OR(difficulty_terrain:d=-:d=3.0:t=-:t=3.0;status:found_yes);type:TRADITIONAL))`

There is no limit how deep a filter can be nested, so users can get really creative. ;-)
Example (build with the new UI): `([:inconclusive=false:advanced=true]AND(OR(NOT(AND(name::contains)))))`

Sidenote: This was rather easy to implement, as the backend luckily had already supported the parsing of nested filters. Only the frontend was missing until now.

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/GeocacheFilterContext.java
+++ b/main/src/cgeo/geocaching/filters/core/GeocacheFilterContext.java
@@ -22,6 +22,7 @@ public class GeocacheFilterContext implements Parcelable {
         FilterType(@StringRes final int titleId) {
             this.titleId = titleId;
         }
+
     }
 
     private FilterType type;
--- a/main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
+++ b/main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
@@ -32,7 +32,8 @@ public enum GeocacheFilterType {
     LOCATION("location", R.string.cache_filter_location, R.string.cache_filtergroup_details, LocationGeocacheFilter::new),
     STORED_LISTS("stored_list", R.string.cache_filter_stored_lists, R.string.cache_filtergroup_userspecific, StoredListGeocacheFilter::new),
     ORIGIN("origin", R.string.cache_filter_origin, R.string.cache_filtergroup_details, OriginGeocacheFilter::new),
-    STORED_SINCE("stored_since", R.string.cache_filter_stored_since, R.string.cache_filtergroup_userspecific, StoredSinceGeocacheFilter::new);
+    STORED_SINCE("stored_since", R.string.cache_filter_stored_since, R.string.cache_filtergroup_userspecific, StoredSinceGeocacheFilter::new),
+    LOGICAL_FILTER_GROUP(null, R.string.cache_filter_logical_filter_group, R.string.cache_filtergroup_special, AndGeocacheFilter::new);
 
 
     private final String typeId;
--- a/main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
+++ b/main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
@@ -14,6 +14,7 @@ import cgeo.geocaching.filters.core.LogicalGeocacheFilter;
 import cgeo.geocaching.filters.core.NotGeocacheFilter;
 import cgeo.geocaching.filters.core.OrGeocacheFilter;
 import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.ui.ImageParam;
 import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.TextSpinner;
 import cgeo.geocaching.ui.ViewUtils;
@@ -48,6 +49,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.jetbrains.annotations.NotNull;
 
 
@@ -58,6 +60,8 @@ public class GeocacheFilterActivity extends AbstractActionBarActivity {
 
     public static final int REQUEST_SELECT_FILTER = 456;
     public static final String EXTRA_FILTER_CONTEXT = "efc";
+    public static final String EXTRA_IS_NESTED = "extra_is_nested";
+    public static final String EXTRA_NESTED_FILTER_POSITION = "extra_nested_pos";
 
     private static final String STATE_CURRENT_FILTER = "state_current_filter";
     private static final String STATE_ADVANCED_VIEW = "state_advanced_view";
@@ -85,6 +89,7 @@ public class GeocacheFilterActivity extends AbstractActionBarActivity {
     private final TextSpinner<GeocacheFilterType> addFilter = new TextSpinner<>();
 
     private boolean processBasicAdvancedListener = true;
+    private boolean isNested = false;
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
@@ -95,8 +100,10 @@ public class GeocacheFilterActivity extends AbstractActionBarActivity {
         binding.filterPropsCheckboxes.removeAllViews();
         this.andOrFilterCheckbox = ViewUtils.addCheckboxItem(this, binding.filterPropsCheckboxes, TextParam.id(R.string.cache_filter_option_and_or), R.drawable.ic_menu_logic);
         this.inverseFilterCheckbox = ViewUtils.addCheckboxItem(this, binding.filterPropsCheckboxes, TextParam.id(R.string.cache_filter_option_inverse), R.drawable.ic_menu_invert);
-        this.includeInconclusiveFilterCheckbox = ViewUtils.addCheckboxItem(this, binding.filterPropsCheckboxes, TextParam.id(R.string.cache_filter_option_include_inconclusive), R.drawable.ic_menu_vague,
-                TextParam.id(R.string.cache_filter_option_include_inconclusive_info));
+
+        final ImmutablePair<View, CheckBox> includeInconclusiveFilterCheckboxItem = ViewUtils.createCheckboxItem(this, binding.filterPropsCheckboxes, TextParam.id(R.string.cache_filter_option_include_inconclusive), ImageParam.id(R.drawable.ic_menu_vague), TextParam.id(R.string.cache_filter_option_include_inconclusive_info));
+        binding.filterPropsCheckboxes.addView(includeInconclusiveFilterCheckboxItem.left);
+        this.includeInconclusiveFilterCheckbox = includeInconclusiveFilterCheckboxItem.right;
 
         filterListAdapter = new FilterListAdapter(binding.filterList);
         initializeFilterAdd();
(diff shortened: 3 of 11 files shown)


main/src/cgeo/geocaching/filters/core/GeocacheFilterContext.java
```diff
@@ -14,14 +14,19 @@ public class GeocacheFilterContext implements Parcelable {
     public enum FilterType {
         LIVE(R.string.cache_filter_contexttype_live_title),
         OFFLINE(R.string.cache_filter_contexttype_offline_title),
-        TRANSIENT(R.string.cache_filter_contexttype_transient_title);
+        TRANSIENT(R.string.cache_filter_contexttype_transient_title),
+        NESTED(R.string.cache_filter_contexttype_nestedfilter_title);
```
eddiemuc, line 18: You are abusing the "FilterType" to transport an information inside nested GeocacheFilterActivities (whether the activity was opened in context of a nested filter or a "main" filter). This is not good, FilterType serves another purpose which is to globally allocate different filter storages. Please remove "NESTED" and use an own dedicated extra-parameter instead (e.g. a boolean parameter "isNested" or something like that)

main/src/cgeo/geocaching/filters/gui/LogicalFilterViewHolder.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public LogicalGeocacheFilter createFilterFromView() {
+        return filter;
+    }
+
+    private void startSubFilterActivity() {
+        final GeocacheFilterContext filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.NESTED);
+        filterContext.set(GeocacheFilter.create(null, true, false, filter));
+        ((GeocacheFilterActivity) getActivity()).selectNestedFilter(filterContext, newContext -> setViewFromFilter((LogicalGeocacheFilter) newContext.get().getTree()));
```
eddiemuc, line 60: In my experience, working with lambdas attached to activities leads to uncercoverable results when the activity issuing the call is destroyed while in subactivity (lambdas can't be saved/restored unfortunately; also lambdas are connected to a concrete activity class instance, but this instance is gone of course after destroy/recreate). 

Does your code work also when you are in a nested filter view and then
* the "main" filter activity is destroyed? (I think this can be simulated somehow in developer settings that only one activity remains)
* you change configuration (e.g. by rotating your phone from portrait to landscape)
...and then go bac to main activity?

eddiemuc (changes requested):
See comments

main/src/cgeo/geocaching/filters/core/GeocacheFilterContext.java
fm-sys: Will do...

main/src/cgeo/geocaching/filters/gui/LogicalFilterViewHolder.java
fm-sys: Indeed, the lambda gets lost while rotating the device: `java.lang.NullPointerException: Attempt to invoke interface method 'void androidx.core.util.Consumer.accept(java.lang.Object)' on a null object reference`

Do you have any recommendation how something like this could be implemented in a better way?

eddiemuc: Unfortunately the only thing I found that works is: use a solution without any lambdas or listeners/callbacks. This makes the code ugly sometimes i am afraid (I had this problem massively with the ImageActivityHelper class).

In your case, I assume you strictly need to fall back to a communication via extra-parameters when opening a "nested filter" activity from the "main" activity, and via "onActivityResult" Intent data to transport something back from nested activity to "main" activity. In your case I assume it should be enough to pass the subfilter's filter config to the main activity, and this main activity can then build it in.

fm-sys: >In your case I assume it should be enough to pass the subfilter's filter config to the main activity, and this main activity can then build it in.

Had thought about it as well. However, what happens if we have multiple nested filters inside the global one? Probably it somehow has to keep track of the position where to insert the updated nested filter... I will try my best. :)

fm-sys: I think I've finally come up with a working solution. Wouldn't mind another review ;-)

.idea/checkstyle-idea.xml
```diff
@@ -3,11 +3,11 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration-0" value="PROJECT_RELATIVE:C:/private/geocaching/github/cgeo/checkstyle.xml:cgeo" />
+        <entry key="active-configuration-0" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle.xml:cgeo" />
```
eddiemuc, line 6: You accidentally checked in local changes to your checkstyle-config

main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
```diff
@@ -32,7 +32,8 @@ public enum GeocacheFilterType {
     LOCATION("location", R.string.cache_filter_location, R.string.cache_filtergroup_details, LocationGeocacheFilter::new),
     STORED_LISTS("stored_list", R.string.cache_filter_stored_lists, R.string.cache_filtergroup_userspecific, StoredListGeocacheFilter::new),
     ORIGIN("origin", R.string.cache_filter_origin, R.string.cache_filtergroup_details, OriginGeocacheFilter::new),
-    STORED_SINCE("stored_since", R.string.cache_filter_stored_since, R.string.cache_filtergroup_userspecific, StoredSinceGeocacheFilter::new);
+    STORED_SINCE("stored_since", R.string.cache_filter_stored_since, R.string.cache_filtergroup_userspecific, StoredSinceGeocacheFilter::new),
+    LOGICAL_FILTER_GROUP(null, R.string.cache_filter_logical_filter_group, R.string.cache_filtergroup_special, AndGeocacheFilter::new);
```
eddiemuc, line 36: You reuse the LogicalGeocacheFilter in a way I had not foreseen or expected. This filter was meant to encapsulate logical OR/AND or NOT groups, not to encapsulate nested filters....

If this works it works. But given that here you instanciate always AND-Filters (naturally, LogicalGeocacheFilter is abstract) I wonder if this setup works when a nested filter is an OR filter?

(I would have expected that you create a new Filter class, e.g. "NestedGeocacheFilter", instead of extending/reusing LogicalGeocacheFilter)

main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
```diff
@@ -383,6 +408,18 @@ public static void selectFilter(@NonNull final Activity context, final GeocacheF
         context.startActivityForResult(intent, REQUEST_SELECT_FILTER);
     }
 
+    public void selectNestedFilter(final LogicalFilterViewHolder holder) {
+        final GeocacheFilterContext nestedFilterContext = new GeocacheFilterContext(TRANSIENT);
+        nestedFilterContext.set(GeocacheFilter.create(null, true,
+                includeInconclusiveFilterCheckbox.isChecked(), holder.createFilterFromView()));
+
+        final Intent intent = new Intent(this, GeocacheFilterActivity.class);
+        intent.putExtra(EXTRA_FILTER_CONTEXT, nestedFilterContext);
```
eddiemuc, line 417: I believe you don't need a GeocacheFilterContext at all. You could simply pass the filter's config as a string:

`holder.createFilterFromView().toConfig()`

From such a config, you can always recreate a GeocacheFilter object using static method `GeocacheFilter.createFromConfig`

main/src/cgeo/geocaching/filters/core/GeocacheFilterType.java
fm-sys: Actually, I was surprised as well. But I would like to stay with this solution, so we don't have to reinvent the wheel. I was just trying to build a GUI which can correctly display logical expressions which were technically already possible.

A different possible syntax would be something like this: `AND(filter1;nested:"AND(...)")`? But I guess, it's preferable to not blow up the expression unnecessarily...

>But given that here you instanciate always AND-Filters

True, this is necessary because `LogicalGeocacheFilter` itself is abstract and cannot be instantiated. Luckily, the Supplier is only used when adding a new filter, not when reading the existing config. This happens, cause the `ExpressionParser` handles these logical groups specially.

>I wonder if this setup works when a nested filter is an OR filter?

Yes, when the user selects OR inside the nested filter activity, an OR instance gets returned as filter and will be attached to the `LogicalFilterViewHolder` which can hold all types of logical groups. So yes, it works for OR filters as well...

main/src/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
fm-sys: A filter context gets created anyway inside the `GeocacheFilterActivity`. So why not use it to pass the current filter...?

grafik

eddiemuc: Hm I guess you're right. Then just leave it as is

.idea/checkstyle-idea.xml
fm-sys: This is your private path to the repository, isn't it? I guess we should all add this file to our local ignore list (https://stackoverflow.com/a/23438899/13513639):
```
git update-index --assume-unchanged ".idea/checkstyle-idea.xml"
```


grafik

eddiemuc: whoops...

fm-sys:
Let's try it in the nightlys...

PR #13019, opened 2022-05-16
Support rendering of HTML tables

### LINKED ISSUE #5894: Better visualisation of HTML tables
Frequently we get questions if we can visualize tables in cache descriptions any better than we currently do.
On quite some caches those tables are used for e.g. calculations or collection of values while caching.

Will add an example later.

### LINKED ISSUE #12910: Identify and solve "table elements" in Cache listings
### Tell us your idea!

In some caches the cache description ends with the following text:

Diese Beschreibung enthält Tabellenelemente, die evtl. nur auf geocaching.com korrekt angezeigt werden.

Example: https://coord.info/GC3DGN4

I assume this refers to table markup usable in GC Listings which is not interpreted by c:geo?

Wouldn't it be possible to interpret this markup and thus render the cache listing correctly?

Interpret GC.com table markup and replace b corresponding HTML code

### Is there an alternative?

Leave everything as is

### Additional Information

_No response_

fm-sys:
Support rendering of HTML tables

fix #12910, fix #5894

### DIFF
--- a/main/src/cgeo/geocaching/ui/renderer/OfflineImageHandler.java
+++ b/main/src/cgeo/geocaching/ui/renderer/OfflineImageHandler.java
@@ -1,0 +1,32 @@
+package cgeo.geocaching.ui.renderer;
+
+import cgeo.geocaching.network.HtmlImage;
+
+import android.graphics.drawable.Drawable;
+import android.text.Spannable;
+import android.text.SpannableStringBuilder;
+import android.text.style.ImageSpan;
+
+import net.nightwhistler.htmlspanner.SpanStack;
+import net.nightwhistler.htmlspanner.TagNodeHandler;
+import org.htmlcleaner.TagNode;
+
+public class OfflineImageHandler extends TagNodeHandler {
+    private final HtmlImage imageGetter;
+
+    OfflineImageHandler(final String geocode) {
+        imageGetter = new HtmlImage(geocode, true, false, null, false);
+    }
+
+    @Override
+    public void handleTagNode(final TagNode node, final SpannableStringBuilder builder, final int start, final int end, final SpanStack spanStack) {
+        final String src = node.getAttributeByName("src");
+
+        builder.append("\uFFFC"); // gets spanned with the image
+
+        final Drawable drawable = imageGetter.getDrawable(src);
+        if (drawable != null) {
+            builder.setSpan(new ImageSpan(drawable), start, builder.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+        }
+    }
+}
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -455,6 +455,10 @@ dependencies {
     // Loupe for ImageGallery and zoomable ImageViews
     implementation 'io.github.igreenwood.loupe:loupe:1.2.2'
     //implementation 'io.github.igreenwood.loupe:extensions:1.0.1' //as of now not needed
+
+    // render HTML in listing (supports e.g. tables)
+    //implementation 'com.github.NightWhistler:HtmlSpanner:0.4'
+    implementation 'com.github.fm-sys:HtmlSpanner:ddf87e9'
 }
 
 // un-mocking of Android classes that don't depend on the Android device, but are portable Java only, like SparseArray
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -85,11 +85,11 @@ import cgeo.geocaching.ui.dialog.EditNoteDialog;
 import cgeo.geocaching.ui.dialog.EditNoteDialog.EditNoteDialogListener;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.ui.recyclerview.RecyclerViewProvider;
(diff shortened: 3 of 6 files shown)


main/src/cgeo/geocaching/ui/renderer/OfflineImageHandler.java
```diff
@@ -0,0 +1,31 @@
...
+
+import android.graphics.drawable.Drawable;
+import android.text.Spannable;
+import android.text.SpannableStringBuilder;
+import android.text.style.ImageSpan;
+
+import net.nightwhistler.htmlspanner.TagNodeHandler;
+import org.htmlcleaner.TagNode;
+
+public class OfflineImageHandler extends TagNodeHandler {
```
eddiemuc, line 13: Offline image handling weas not included in the solution before, or was it?

eddiemuc (approved):
Looks good to me, apart from Codacy comments,

main/src/cgeo/geocaching/ui/renderer/OfflineImageHandler.java
fm-sys: It was. Would have been quite bad if images were always re-downloaded when opening the listing.

Here (the `HtmlImage` part is inlined):
https://github.com/cgeo/cgeo/blob/b3b521ca57542ca560341f616ba55d5cf1460d5e/main/src/cgeo/geocaching/CacheDetailActivity.java#L1877

fm-sys:
Unfortunately, the lib doesn't seem to cover all rendering by default which was supported by the previous implementation. Luckily the lib can easily be extended to support those, but it would be helpful if we could collect a list of caches with special listings...

fm-sys:
While working with the library, I noticed there are quite some rendering bugs and missing styling handling which was working in the old implementation. These things can easily be fixed, however it's IMHO ugly to fix all those things inside the cgeo code. 

Would it be ok to create our own `HtmlSpanner` fork?

fm-sys:
> Take current solution, but if a table tag is detected then pass this part of the code to the new library?

The situation is a bit more complicated, cause other HTML tags need to be correctly rendered, even if they are placed inside the table. 

But the lib isn't that bad than it might sound like. In fact, the results are very promising. Many of those library bugs I mentioned are fixed meanwhile, however the current SNAPSHOT builds are not compiling and the last tagged release is from 2013.

I've meanwhile forked the library, fixed the gradle build and added support for two missing tags: https://github.com/fm-sys/HtmlSpanner

So IMHO that would be the way to go...

fm-sys:
retest this please

fm-sys:
retest this please

moving-bits:
retest this please

fm-sys:
Not sure whether it is possible to unit test this in a way which makes sense... Anyway, here as some good test cases which contain special rendering which might be helpful while reviewing this PR:

[GC1F03A](https://coord.info/GC1F03A)
[GC3DGN4](https://coord.info/GC3DGN4)
[GC9Q1HQ](https://coord.info/GC9Q1HQ)
[GC3FHVK](https://coord.info/GC3FHVK)
[GCYNH6](https://coord.info/GCYNH6)

And another thing: Probably it's better to track the HtmlSpanner fork on behalf of the [c:geo organization](https://github.com/cgeo) instead of my private fork. Can someone with the necessarily permissions (I think @Lineflyer ?) fork the [original HtmlSpanner repository](https://github.com/NightWhistler/HtmlSpanner)? I could then push the changes I made to that fork...

fm-sys:
Hi, just a quick reminder: How do we want to proceed with this PR?

moving-bits:
@Lineflyer (see https://github.com/cgeo/cgeo/pull/13019#issuecomment-1133880890), @fm-sys: Any news in this PR?

fm-sys:
> @fm-sys: Any news in this PR?

Haven't worked on it anymore for the last 3 months. Not sure whether it was already ready to be merged.

Furthermore, the question arises whether this is the way to go at all. We could use a full featured webview instead if this is wanted. Maybe even providing a possibility to switch between both modes. Just thinking...

Lineflyer:
> And another thing: Probably it's better to track the HtmlSpanner fork on behalf of the [c:geo organization](https://github.com/cgeo) instead of my private fork. Can someone with the necessarily permissions (I think @Lineflyer ?) fork the [original HtmlSpanner repository](https://github.com/NightWhistler/HtmlSpanner)? I could then push the changes I made to that fork...

Does that mean I should fork the mentioned repo as cgeo/HtmlSpanner ?

fm-sys:
> Does that mean I should fork the mentioned repo as cgeo/HtmlSpanner ?

Yes, that would be the way to go IMHO. At least if the path followed in this PR is considered good in general...

fm-sys:
Probably we should go a different way if we want to support HTML tables... Too many regressions are expected with this approach, just to add this very small additional feature. If I find time, I'll play around with some different ideas.

fm-sys:
Back in the days I created 4 PR's at the `HtmlSpanner` repo which are merged meanwhile... So maybe we don't need to use a fork anymore.

bekuno:
I would like to see better tables in c:geo.
@fm-sys Is there a chance (based on your work) ?

Lineflyer:
Closed as outdated as discussed in team meeting on October 8th 2023.

In general such approach could be a solution.

PR #13023, opened 2022-05-17
introduce a new setting key for the extended "long tap on map" and update strings

### LINKED ISSUE #12821: "long tap on map" & waypoints

ztNFny:
Introduce a new setting key for the now extended setting (https://github.com/moving-bits/cgeo/commit/5a2b8097dfec1544051f68c6e5936b503640ee9d) to avoid update-problems described in #12821

Also rename the strings to include "waypoints" which are now affected by this

fixes #12821

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1468,9 +1468,9 @@
     <string name="internal_goto_targets_title">\'Go to\' targets</string>
     <string name="internal_goto_targets_description">This cache stores your recent \"Go to\" targets. It is automatically created once you have used the \"Go to\" function in the main menu. The coordinates entered there will be stored into waypoints of this cache. If you do no longer need these coordinates you can safely delete this cache or any of its waypoints. Also you can move the cache to any other list of your choice.</string>
     <string name="init_longtap_map">Long tap on map</string>
-    <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache)</string>
-    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
-    <string name="onetime_info_longtap_enabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently enabled.\n\nYou can change this in Settings → Map → Map behavior</string>
+    <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache / waypoint)</string>
+    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes and user-defined caches and waypoints.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
+    <string name="onetime_info_longtap_enabled">By long tapping on map you can create individual routes and user-defined caches and waypoints.\n\nLong tap on map is currently enabled.\n\nYou can change this in Settings → Map → Map behavior</string>
     <string name="goto_targets_list">User-defined caches</string>
     <string name="clear_goto_history_title">Clear goto history</string>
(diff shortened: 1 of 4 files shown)


Lineflyer (approved):
Approving the general idea and the string change. Codewise needs a look from others.

Lineflyer:
@ztNFny Can you rebase this to `release` please?

main/res/values/strings.xml
```diff
@@ -1468,9 +1468,9 @@
...
     <string name="init_longtap_map">Long tap on map</string>
-    <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache)</string>
-    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
-    <string name="onetime_info_longtap_enabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently enabled.\n\nYou can change this in Settings → Map → Map behavior</string>
+    <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache / waypoint)</string>
+    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes and user-defined caches and waypoints.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
```
moving-bits, line 1472: Generally fine, only that I would avoid the repeated "and". Maybe change to "... create individual routes, user-defined caches and waypoints."? (Same in next line)

main/res/values/strings.xml
```diff
@@ -1468,9 +1468,9 @@
...
-    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
-    <string name="onetime_info_longtap_enabled">By long tapping on map you can create individual routes and user-defined caches.\n\nLong tap on map is currently enabled.\n\nYou can change this in Settings → Map → Map behavior</string>
+    <string name="init_summary_longtap_map">Activate long tap on map icon (create an individual route) or on free map space (create a user-defined cache / waypoint)</string>
+    <string name="onetime_info_longtap_disabled">By long tapping on map you can create individual routes, user-defined caches / waypoints.\n\nLong tap on map is currently disabled.\n\nYou can change this in Settings → Map → Map behavior</string>
+    <string name="onetime_info_longtap_enabled">By long tapping on map you can create individual routes and user-defined caches / waypoints.\n\nLong tap on map is currently enabled.\n\nYou can change this in Settings → Map → Map behavior</string>
```
moving-bits, line 1473: Thanks for tweaking the strings. But now it's mixed - "individual route, user-defined caches / waypoints" in the first line (1472), and "individual routes and user-defined caches / waypoints" in the second (1473). The first variant looks somewhat strange to me, so I would either use the second variant or "individual routes, user-defined caches and waypoints". Or maybe "individual routes and user-defined caches or waypoints"?

ztNFny: oops, changed it to "routes, caches and waypoints" at first, then forgot to change that one "," back to "and". Update pushed.

It's 2 features: "Add/remove to route" and "Add a point" (cache or waypoint, depending on where you are). that's why I phrased it with the "/"

Lineflyer:
@ztNFny Thanks for tackling this.

main/res/values/strings.xml
moving-bits: Understood. Thanks for tackling the preference issue and for adapting the strings.

PR #13035, opened 2022-05-21
support setting found state for UnknownConnector (fix #13029)

### LINKED ISSUE #13029: Change found state not only for "AL..." caches
### Tell us your idea!

Currently c:geo allows to set the found state ("Set as found") for adventure lab caches (more specifically for caches with an id starting with "AL").
Please provide this function for all caches.
Or at least for all imported caches or for all caches not connected to a service (like geocaching.com, where the found state can be set).

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Activates setting found state for UnknownConnector (which gets triggered for all caches not related to one of the other connectors, thus especially for caches imported via Lab2Gpx when using one of the "free" prefixes)

### DIFF
--- a/main/src/cgeo/geocaching/connector/unknown/UnknownConnector.java
+++ b/main/src/cgeo/geocaching/connector/unknown/UnknownConnector.java
@@ -56,4 +56,8 @@ public class UnknownConnector extends AbstractConnector {
         return null;
     }
 
+    @Override
+    public boolean supportsSettingFoundState() {
+        return true;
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

fm-sys:
The found state can be set for Labs as well as for UDC's, so it's totally consistent to allow it for all other "unknown" caches as well...

PR #13037, opened 2022-05-22
remove Geolutins (fix #12976)

### LINKED ISSUE #12976: Removing Geolutins?
### Tell us your idea!

Geolutins website seems to be down. See for yourself at http://www.geolutins.com/

Service does not seem to work anymore either, c:geo test is failing for it:

image
 

Maybe time to remove Geolutins from c:geo?

### Is there an alternative?

Leave it (hoping it will resurrect somehow) but comment out the failing test

### Additional Information

_No response_

moving-bits:
## Description
Remove support for Geolutins


fm-sys:
I would prefer to wait a bit longer before finally removing the service. The website is down for two weeks now, I would propose to wait two more weeks to give them some more time... (Maybe they are working on it and we just don't know about it? 🤷‍♂️) 

IMHO there is no reason to remove the connector as fast as possible.

moving-bits:
Ok for me as well, although from reading https://github.com/cgeo/cgeo/issues/12976#issuecomment-1133877166 I doubt the service will be revived - no answer in two consecutive years...?

bekuno:
The website is not down - the website is out of work.
At the current whois entry (e.g. https://whois.domaintools.com/geolutins.com) we see
- Created on 2004-04-28
- Expires on 2022-04-28

Currently the domain is parked by the registrar (godaddy).

The last useable snapshot on the wayback machine is https://web.archive.org/web/20210605060313/http://www.geolutins.com/index.php
There I did not find a email adress outside the geolutins.com domain.

I propose to target the PR to master and merge after the next release is live.

Lineflyer:
I appreciate the fact, that you want to be conservative in removing it, however the files touched with this removal are quite unlikely to later cause large conflicts in case we decide to revert.
At the same time this PR is nearly 100% safe from causing any functional regressions, therefore lets merge to `release`.

PR #13043, opened 2022-05-22
background downloader enhancements

fm-sys:
- rel. to #330

## show toast to inform users about what is happening

grafik

## show progress (e.g. "2/5") instead of current geocode

grafik


moving-bits:
Since background downloading of caches is already present an `release`, and the changes being small and useful: Would it make sense to target this PR to `release`?

fm-sys:
Although the code is present on `release` the feature is currently hidden via `BranchDetectionHelper` like all other "experimental" features. As this feature is not yet ready anyway, I guess targeting it to `master` makes more sense.

PR #13057, opened 2022-05-27
rel to #12783: redesign imagegallery to use gridlayout and simpler transitioning logic

eddiemuc:
rel to #12783: redesign imagegallery to use gridlayout and simpler transitioning logic


Lineflyer:
As we have seen, that problems in this hidden feature can still cause problems for users, my question would be: Do we need this on the release branch or would master be sufficient?

eddiemuc:
Retargeted to master

PR #13059, opened 2022-05-27
fix #13051: add possibility to check waypoint in geochecker

### LINKED ISSUE #13051: re-add menu-entry for geochecker
### Tell us your idea!

Please re-add the menu-entry for the geochecker.
Normally I need the geochecker for calculated coordinates - I copy the coordinates from the waypoint-list and open from there the checker.
If the cache has a checker, I could see it whether the menu-entry exists or not and I could use it directly from the menu.
Now I have to swipe to the description page, scroll and tap the icon - hoping that the cache has a checker and the icon exist - because I don't see it before.

re-add the menu-entry for the checker and keep the icon under the description.

### Is there an alternative?

_No response_

### Additional Information

menu-entry was removed in #12960

eddiemuc:
fix #13051: add possibility to check waypoint in geochecker

### DIFF
--- a/main/res/menu/waypoint_options.xml
+++ b/main/res/menu/waypoint_options.xml
@@ -47,5 +47,9 @@
         android:id="@+id/menu_waypoint_caches_around"
         android:title="@string/cache_menu_around">
     </item>
+    <item
+        android:id="@+id/menu_waypoint_open_geochecker"
+        android:title="@string/waypoint_open_geochecker">
+    </item>
 
 </menu>
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1558,6 +1558,7 @@
     <string name="waypoint_duplicate">Duplicate waypoint</string>
     <string name="waypoint_toclipboard">Copy waypoint to clipboard</string>
     <string name="waypoint_copy_of">Copy of</string>
+    <string name="waypoint_open_geochecker">Open Geochecker</string>
     <string name="waypointcalc_notes_prompt">Enter notes here…</string>
     <string name="waypoint_coordinate_formats_plain">Plain</string>
     <string name="from_clipboard">From clipboard</string>
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -136,6 +136,7 @@ import android.text.style.ForegroundColorSpan;
 import android.text.style.StyleSpan;
 import android.text.style.URLSpan;
 import android.text.util.Linkify;
+import android.util.Pair;
 import android.view.ContextMenu;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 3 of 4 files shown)


eddiemuc:
image

ztNFny:
looks good codewise. But I'd suggest to show the option only on final waypoints.

eddiemuc:
> I'd suggest to show the option only on final waypoints.

Why restrict it to a certain wp type? There are lazy people (like myself) out there who don't actually work with wp types at all (my "final" coordinates are most often normal reference waypoints), why force them to set the type just to be able to check the coord?

ztNFny:
To reduce clutter? The menu is already overcrowded as is with 10 items. Context-sensitive menus have been a best practice for years. (Cleanup of menus is another issue, but at least not adding more clutter should be taken care of in any issue - ie here)

eddiemuc:
> To reduce clutter? The menu is already overcrowded as is with 10 items.

I totally agree that the menu is cluttered. But to remove the geochecker function from waypoints with a type != final is IMHO still not a valid solution, because it assumes that geochecker only makes sense in context of a final waypoint. Examples where geochecker might make sense for non-final waypoints:
* lazy people (not using waypoint types)
* generated waypoints (trying to check some of them)

Cleaning up the context menus is a totally different and valid topic. Here are a few suggestions of what could be removed:

image

* "Edit" (Bearbeiten): is reached by short-click, why is it needed too in the context menu?
* "Visited" (Besucht): imho no needed as a shortcut at this place
* "Delete coordinates" (Koordinaten löschen) imho not needed as a shortcut at this place
* "Copy waypoint into clipboard" (Wegpunkt in die Zwischenablage kopieren): what is this used for?
* "Compass"/"Navigation" -> can be reached by clicking the compass icon, no need for separate context menu entry
* "Surrounding caches" (Caches im Umkreis): no need for a shortcut IMHO. coordinates can be copied, then normal search can be used
* Alternatively, "Compass/Navigation/SurroundingCaches" can be put into a new context menu  when tapping compass icon

moving-bits:
I agree that most of the entries you've listed could be removed without harm. "Copy waypoint into clipboard" should stay, though, as it is used (and was explicitly implemented) for copying waypoints between caches.

P. S.: I also agree that this menu entry should be available for all waypoint types.

ztNFny:
> "Copy waypoint into clipboard" should stay, though, as it is used (and was explicitly implemented) for copying waypoints between caches.

What's the usecase for that? Can't think of a single situation where that'd be needed (unless you're editing the wrong cache, but then we'd need a "copy" feature for lots of stuff). And can't that usecase be covered by putting the waypoints to note, then copying the note?

> lazy people (not using waypoint types)
> generated waypoints (trying to check some of them)

Generated is a valid usecase. Lazy not imho

SammysHP:
> - lazy people (not using waypoint types)
> - generated waypoints (trying to check some of them)

There might be a checker for stages of a multi cache. I think I've seen this once.

ztNFny:
> There might be a checker for stages of a multi cache. I think I've seen this once.

In this case this feature isn't for you anyway as it's intended for avoiding looking at the cache listing (from where you already have the checker link) to figure out which stage the checker is for - or even clicking the correct checker as c:geo will just pick a "random" checker from the listing.

eddiemuc:
Hm, discussion has lead a bit away from original intent of the PR (my fault, sorry for that). Since PR seems to be ok codewise I would merge it in a couple of hours (except if there are other concerns raised). We might further discuss (and adapt, if necessary) the display of the feature on different wp types later. Hope that's ok.

IMHO the more general problem of "overcrowded context menus" deserves its own issue + discussion space if we want to pursue it further.

PR #13061, opened 2022-05-28
rel to #12783: allow adding of non-images to gallery

eddiemuc:
rel to #12783: allow adding of non-images to gallery


eddiemuc:
This PR will allow Non-Image-Files to be added to image gallery:

image

eddiemuc:
It will also fix an error in image gallery: when adding/deleting images, count got lost which could lead to IndexOutOfBoundExceptions and strange GUI behaviour

eddiemuc:
image

eddiemuc:
CI has failing test cases which seems to be unrelated (see screenshot). I saw that @Lineflyer recently disabled the cache in question, am I assuming correct that this makes the test case fail?

image

fm-sys:
Does the large preview makes sense for non-images?

eddiemuc:
> Does the large preview makes sense for non-images? 

No, not much. This is why when you click on a non-image in the gallery view, the content is opened directly (not the details view). The details view is only reached when "swiping" through the detail images in detail view.

You might argue that we can leave them out also during swipe. But then other questions arise:
* Counts will differ in detail view and overview
* Insight info can't be displayed (although there is none yet except for the mime type)

Current solution is the easiest one to implement for now. We can discuss adjustment ideas, but imho this can be done better once the feature is available for usage. The whole thing is experimental anyway.

eddiemuc:
Will clena up code some more, thus marking PR as "in progress"

eddiemuc:
OK, ready now. 

I can't figure out how to make back-transitions work if CacheDetailActivity was destroyed inbetween. The setup with ViewPager2 and own Fragment tabs is too complex for me to see it through... However, if CacheDetailActivity still exists (which is the case normally) then back-transition works fine

PR #13063, opened 2022-05-29
tweak bearing indicator for unifiedmap

### LINKED ISSUE #13039: tweak map rotation marker
You can use this for now, i'll then tweak it compass

_Originally posted by @ztNFny in https://github.com/cgeo/cgeo/issues/12925#issuecomment-1129195169_

ztNFny:
closes #13039

### DIFF
--- a/main/res/drawable/bearing_indicator.xml
+++ b/main/res/drawable/bearing_indicator.xml
@@ -1,28 +1,15 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-    android:width="40dp"
-    android:height="40dp"
+    android:width="36dp"
+    android:height="36dp"
     android:viewportWidth="24"
     android:viewportHeight="24">
   <path
-      android:pathData="M12,12m-10,0a10,10 0,1 1,20 0a10,10 0,1 1,-20 0"
-      android:strokeLineJoin="round"
-      android:strokeWidth="1"
-      android:fillColor="#ffffff"
-      android:strokeColor="#ccb8b8b8"
-      android:fillAlpha="0.7529412"
-      android:strokeLineCap="round"/>
+      android:pathData="M12,12m-12,0a12,12 0,1 1,24 0a12,12 0,1 1,-24 0"
+      android:fillColor="@color/osm_zoomcontrolbackground" />
   <path
       android:pathData="M12.0005,3.5156 L8.9029,12H11.1997A0.8,0.8 0,0 1,12.0005 11.1992,0.8 0.8,0 0,1 12.8013,12h2.2969z"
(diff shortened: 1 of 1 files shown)


moving-bits (dismissed):
Thanks for diving into this. I do like the lighter color of the arrow's grey part compared to the original version. Comparing the new VTM compass rose to the compass rose on Google Maps, there are two major differences, though:
- compass rose on VTM is a bit larger now than the one on GM
- compass rose on GM has a grey border which makes it more distinguishable from the background, but it got removed from the VTM version

Would be nice if you could re-add the grey border and adjust the sizes to make the two more similar.

ztNFny:
None of our buttons (zoom, settings, ..) have borders, so this one is fully in line with those.

moving-bits:
So maybe we should use some... Have a look at the controls in Google Maps: Zoom controls use a slight shadow, and compass rose uses an actual border. Both make it way more readable (distinguishable from background). Thus I propose to not only add the border to VTM compass rose, but also add slight shadows to VTM zoom controls and quick settings buttons for both maps. (I'm talking about `UnifiedMap` here, but to be consistent same should apply to today's GMv2 and `NewMap`.)

ztNFny:
The Google maps view in c:geo is using an outdated look compared to what Google Maps currently uses. It's based on what material design from 3-4 years ago,, we shouldn't be something upon that now

ztNFny:
retest this please

moving-bits:
> The Google maps view in c:geo is using an outdated look compared to what Google Maps currently uses. It's based on what material design from 3-4 years ago,, we shouldn't be something upon that now

AFAIK it's the most recent version of GM library available (non beta), so I guess it's ok to use it. But I do not sting to that specific style, for me it's more about consistency - we should try to harmonize the map controls across the map implementations, if possible. Is there a way to tweak GM UI controls?

Regarding material style: What would be the applicable component for such a UI element? The closest I could find are floating action buttons (FAB), all of which are non-transparent and do have a shadow in the examples I've seen, making them much easier to recognize. 
In contrast to this, GM's compass rose is semi-transparent, but does have a clear border and a shadow, which makes it stand out more clearly from the background. Zoom control buttons do have a shadow as well. Our own map control buttons are semi-transparent as well, but do not use any specific border or shadow.
To complicate things even further, Google just released "Material 3" with a couple of further tweaks...

My proposal would be to decide on how our map controls should look like, and implement this consistently across all maps - GM, `NewMap` and `UnifiedMap`.

ztNFny:
I suggest moving the discussion about how map elements should look to a separate issue
This PRs icon is in line with the existing map elements as is.

bekuno:
retest this please

Issue #13080, opened 2022-06-03
Too many log type options for owned archived events in c:geo


Vike91:
### Describe your problem!

Tried to make an announcement on an archived event in c:geo. Failed due to unknown error.

Tried to log it on the website, only write note is available for archived events. 

### How to reproduce?

Go to own archived event in c:geo.
Make any log but will attend, attended or write note. In the example I used Announcement. 

### Actual result after these steps?

Logging failed due to unknown reason. 

### Expected result after these steps?

Announcement sent to the archived event. 

### Reproducible

Yes

### c:geo Version

2022.05.24-RC

Vike91:
Just to clarify, I think the c:geo logging screen has too many options for events that are both owned and archived. 

Announcement, maintenance performed, disabled or enabled logs don't work on the owned archived events.

PR #13081, opened 2022-06-03
show only logtypes that are available online

ztNFny:
we're currently using only the cache type, status and connector to determine the available log types.
This leads to situations where more log types are available in c:geo than on the platform, e.g. archived events (https://github.com/cgeo/cgeo/issues/13080) but also normal GC caches where c:geo shows 5 logtypes, including NM and NA - which on GC are only available as a "report problem" option.

This PR updates the list of available log types to match what's available online.


moving-bits:
Failing tests are unrelated (GC3FJ5F being disabled currently), so let's merge

PR #13086, opened 2022-06-04
Refresh single caches in background (rel. to #330)

fm-sys:
grafik


ztNFny:
Does this only work from the list menu or also when a cache is open? The later would be more interesting IMHO - the menu item is quite useless imho, using it takes longer than opening the cache and pull to refresh

fm-sys:
This PR will only change the cache list menu to the new background-downloader and only if cgeo is compiled as non-production build.

More should be done in a separate step, following the [best practice](https://deepsource.io/blog/git-best-practices/#1-make-clean-single-purpose-commits) to create commits/PRs which are as small as possible.

PR #13091, opened 2022-06-05
remove size-limit check while storing waypoints in PN

### LINKED ISSUE #10981: Save Waypoints: Remove rejection to store because of PN length
I often use the [Store Waypoints] function to get the coords of all my WPs into the PN. Over there I usually edit the stuff, to create my own list of stages (with add. info). And afterwards the {c:geo} section may even be gone completely.

But, when my PN is already close to 2500 chars (which happens pretty often as we love to do challenging Multis and Mysteries), the function denies to store the WPs because of PN length - which makes no sense at all, IMHO, cause a) c:geo's PN is not limited anyway, and b) the length will just be an issue on upload, which will get an appropriate error message, anyway.

## Suggested solution
Add the WP storage to PN, no matter how long the PN will get (perhaps the WPs will just be stored to share them via messenger ...?), there may be a toast when the limit of 2500 chars is reached, but please just write them to PN as they are.

Moreover: Consequently following the storage procedure's rejection, c:geo must even forbid PNs with more than 2500 chars - which would of course be complete nonsense. ;)

murggel:
## Description

Remove the size-limitation on export of waypoints to personal note.
Instead show toast on upload with character-count, which exceeds the size-limit (so user knows, how many characters he has to remove to avoid shortening on upload)

## Related issues

fix #10981

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1298,11 +1298,10 @@
     <string name="cache_personal_note_removedwaypoints">Waypoints removed</string>
     <string name="cache_personal_note_removewaypoints_nowaypoints">No waypoints found to remove</string>
     <string name="cache_personal_note_storewaypoints_success">Waypoints were added to personal note</string>
-    <string name="cache_personal_note_storewaypoints_limited_reduce">Waypoints were added to personal note (%d characters more than supported by Geocaching.com)</string>
     <string name="cache_personal_note_storewaypoints_nowaypoints">No storable waypoints found</string>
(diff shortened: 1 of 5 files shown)


murggel:
Failing tests are probably unrelated

murggel:
I added a message how many characters will be deleted on upload, so user know how many he has to reduce for uploading a complete note.

Should I add this also to the dialog on upload?
Or remove it from storing the waypoint and only on upload?

murggel:
will implement:  show count of truncated characters in message for upload

fm-sys:
retest this please

PR #13092, opened 2022-06-06
replace GC3FJ5F by GC40 in tests (fix #13082)

### LINKED ISSUE #13082: CI test fail due to GC3FJ5F being disabled
### Describe your problem!

GC3FJ5F is disabled but tests expect it to be enabled

@Lineflyer ping

### How to reproduce?

run 
[cgeo.geocaching.CgeoApplicationTest.testSearchByGeocodeBasis](https://ci.cgeo.org/job/cgeo-CI_PR-build/2032/testReport/junit/cgeo.geocaching/CgeoApplicationTest/testSearchByGeocodeBasis/)
[cgeo.geocaching.connector.gc.GCBaseTest.testSearchFromMap](https://ci.cgeo.org/job/cgeo-CI_PR-build/2032/testReport/junit/cgeo.geocaching.connector.gc/GCBaseTest/testSearchFromMap/)

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

*

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
replace GC3FJ5F by GC40 in tests


Lineflyer:
@moving-bits 
There is one assertion left, which needs to be corrected for the test to pass (see CI result).

moving-bits:
> @moving-bits There is one assertion left, which needs to be corrected for the test to pass (see CI result).

I had seen that, thanks. Failing part is comparing the cache owner display names between the mocked cache data and the actual cache data from loading the cache. 
For GC40 this string is "Pierre Cao, Speerpunt, adopted by Grokky" - in contrast to the owner id, which is "Grokky Grokson". For some yet unknown reason testing framework has the one string as owner display name for the mocked cache, and the other string for the loaded cache - which should not be the case. For GC3FJ5F both strings are "Lineflyer", so that comparison problem has not come up yet. 
Haven't found the time yet to investigate further where this string gets set to the false value. We should clarify this even if we decide not to exchange GC3FJ5F by GC40 in the tests.

fm-sys:
I've investigated the problem a bit further. In `GCBaseTest.testSearchFromMap()` we are still testing the MapDetailsAPI of GCs old map. The code for this is however only used in this test, but nowhere else.  In other words, we are testing code parts we aren't using at all... 😅

As this API does only return the owner id, not the display name, the test is failing for GC40. 

Solution: I would suggest dropping the test completely and remove all related code parts.

grafik

fm-sys:
The code was too deeply scattered into the code base, so I created a separate PR https://github.com/cgeo/cgeo/pull/13119. For fixing the tests on `release`, we should just comment out the problematic test `GCBaseTest.testSearchFromMap()` for now...

fm-sys:
Can we merge here to get the build green again?

moving-bits:
Thanks @fm-sys for your analysis and for providing #13119. Will merge here now to get builds back green.

PR #13095, opened 2022-06-06
add gc login status to system info (fix #13079)

### LINKED ISSUE #13079: Include last login failure reason in system information
### Tell us your idea!

Most of the "normal" support mails we receive are about various login issues. 
Currently the system information only includes the login state, which is typically something like "Logging in" (if the login is still in progress while the sys info was generated) or "Login failed" (without reason).

Also the attached logfile is not helpful as it does mostly not contain any helpful message (without debugging mode) to analyse login problems.

It would therefore be extremely helpful if information like this would be included directly within the system information:
- Last failed login time: timestamp
- Last failed login reason: reason, as used for user prompt (e.g. Password wrong, Communication error, Captach required)
- Last successful login time : timestamp

### Is there an alternative?

We could always log the login progress/result into the logfile, but there are arguments against:
- Clutters log
- Not all support mails include log but most include system information
- Harder to analyse as you need to search through the logfile
- The last failed and last success times might not be included in the logfile

### Additional Information

_No response_

moving-bits:
## Description
Store last login error (status & time) for gc connector as well as time for last successful login:

image

### DIFF
--- a/main/res/values/preference_keys.xml
+++ b/main/res/values/preference_keys.xml
@@ -413,6 +413,9 @@
     <string translatable="false" name="pref_gc_avatar">gc_avatar</string>
     <string translatable="false" name="pref_memberstatus">memberstatus</string>
     <string translatable="false" name="pref_gccustomdate">gccustomdate</string>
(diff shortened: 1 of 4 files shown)


fm-sys:
retest this please

PR #13102, opened 2022-06-07
Update mocked data for GC3FJ5F

Lineflyer:


Lineflyer:
It was decided to use GC40 instead. See #13092

PR #13103, opened 2022-06-07
fix failing tests due to listing changes in GC3FJ5F

ztNFny:
@Lineflyer several data of your cache are hardcoded in this class. for description it's comparing the beginning, for many other data the full string

### DIFF
--- a/tests/src-android/cgeo/geocaching/test/mock/GC3FJ5F.java
+++ b/tests/src-android/cgeo/geocaching/test/mock/GC3FJ5F.java
@@ -71,12 +71,12 @@ public class GC3FJ5F extends MockedCache {
 
     @Override
     public String getHint() {
-        return "[FINAL] Geteilter Baumstumpf, dann ganz unten (Spoiler-Bild)";
+        return "[FINAL] Unter umgestürztem Baumstumpf (siehe Spoilerbild)";
     }
 
     @Override
     public String getDescription() {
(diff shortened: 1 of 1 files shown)


Lineflyer:
Thanks @ztNFny for investigating... 
I was not so deep into those testing structure, but know I know.

I will take a look next days to combined updated mock data with your changes unless @moving-bits prefers to exchange it with GC40 as in his PR. @moving-bits Let me know.

moving-bits:
> I will take a look next days to combined updated mock data with your changes unless @moving-bits prefers to exchange it with GC40 as in his PR. @moving-bits Let me know.

No clear preference from my side. Someone suggested using GC40 (or some cache with infrequent disablings) instead of your cache since you mentioned regular maintenance several times a year, which would lead to tests failing during that time.

Lineflyer:
@moving-bits Lets take GC40. The only drawback is, that in case we wish to include certain parsing specialties into the test, we would need a cache we have control of. But AFAIK this is not the case at this moment.

So please continue here and I will close the other PR(s).

eddiemuc:
>  in case we wish to include certain parsing specialties...

IMHO test cases downloading real listings and comparing their values with expected ones should solely be used for connector and basic data download use cases - because those tests will always be subject to degration. More complex parsing logic should always be unit tested using non-connected, non-instrumented tests.

PR #13109, opened 2022-06-08
Enhance map long click functionality (fix #12361, fix #13104)

### LINKED ISSUE #12361: Pick up coordinates from map
### Tell us your idea!

Request from user at support:

It's cumbersome having to create a waypoint or user-defined cache on map just to get coordinates for a specific point.

Display a movable cross on the map (on demand) and display the cross's coordinates. Moving the cross by dragging it around updates the coordinates.

### Is there an alternative?

_No response_

### Additional Information

ticket number 685604

### LINKED ISSUE #13104: Allow adding map points to individual route directly
### Tell us your idea!

Often, to plan a route, you need to add intermediate points on the map.

Currently, as a workaround, I just create custom caches in the places I want, and then plan the route from custom cache to custom cache.

However creating custom caches involves lots of unnecessary taps.

I never ever use custom caches for anything else than the above described use.

So as an alternative to custom caches, when long pressing the map, it would be cool, if could just drop small map markers, like sticking a "needle" into a map. Creating another needle should just require a long tap and not open any new window. Just the needle appears.

And once you have dropped all needles, you can then plan a route from needle to needle.

### Is there an alternative?

_No response_

### Additional Information

_No response_

fm-sys:
This implementation brings several improvements:
- Create a UDC, even when a cache target is currently set
- Place WPs / UDCs more precisely
- Add a location to the individual route without having to create a WP / UDC first (#13104)
- Copy map coordinates (#12361)
- Make long-tab on map less disturbing (#12821)
- Fix inconsistency between GMap/Mapsfore when setting a cache as target while being at the live map

grafik

Currently only implemented for Mapsforge Map, therefore setting WIP...

### DIFF
--- a/main/res/drawable/map_pin.xml
+++ b/main/res/drawable/map_pin.xml
@@ -1,0 +1,25 @@
+<!-- Taken from https://de.wikipedia.org/wiki/Datei:Google_Maps_pin.svg -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="20"
+    android:viewportHeight="34.89234"
+    android:width="20dp"
+    android:height="34.89234dp">
+    <group
+        android:translateX="-814.596"
+        android:translateY="-274.3862">
+        <group
+            android:scaleX="1.185585"
+            android:scaleY="1.185585"
+            android:translateX="-151.1772"
+            android:translateY="-57.3976">
+            <path
+                android:pathData="M817.11249 282.97118c-1.25816 1.34277 -2.04623 3.29881 -2.01563 5.13867 0.0639 3.84476 1.79693 5.3002 4.56836 10.59179 0.99832 2.32851 2.04027 4.79237 3.03125 8.87305 0.13772 0.60193 0.27203 1.16104 0.33416 1.20948 0.0621 0.0485 0.19644 -0.51262 0.33416 -1.11455 0.99098 -4.08068 2.03293 -6.54258 3.03125 -8.87109 2.77143 -5.29159 4.50444 -6.74704 4.56836 -10.5918 0.0306 -1.83986 -0.75942 -3.79785 -2.01758 -5.14062 -1.43724 -1.53389 -3.60504 -2.66908 -5.91619 -2.71655 -2.31115 -0.0475 -4.4809 1.08773 -5.91814 2.62162z"
+                android:fillColor="#FF4646"
+                android:strokeColor="#D73534"
+                android:strokeWidth="1" />
+            <path
+                android:pathData="M826.0661 288.2528A3.0355 3.0355 0 0 1 819.9951 288.2528A3.0355 3.0355 0 0 1 826.0661 288.2528Z"
+                android:fillColor="#590000" />
+        </group>
+    </group>
+</vector>
--- a/main/res/menu/map_longclick.xml
+++ b/main/res/menu/map_longclick.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android">
+    <item
+        android:id="@+id/menu_add_waypoint"
+        android:title="@string/cache_waypoints_add">
+    </item>
+    <item
+        android:id="@+id/menu_udc"
+        android:title="@string/create_internal_cache">
+    </item>
+    <item
+        android:id="@+id/menu_add_to_route"
+        android:title="@string/context_map_add_to_route">
+    </item>
+    <item
+        android:id="@+id/menu_coords"
+        android:title="@string/context_map_get_coords">
+    </item>
+</menu>
--- a/main/res/menu/map_longclick.xml
+++ b/main/res/menu/map_longclick.xml
@@ -13,6 +13,10 @@
         android:title="@string/context_map_add_to_route">
     </item>
     <item
+        android:id="@+id/menu_add_to_route_start"
+        android:title="@string/context_map_add_to_route_start">
+    </item>
+    <item
         android:id="@+id/menu_coords"
         android:title="@string/context_map_get_coords">
     </item>
--- a/main/res/menu/map_longclick.xml
+++ b/main/res/menu/map_longclick.xml
@@ -18,6 +18,10 @@
     </item>
     <item
         android:id="@+id/menu_coords"
-        android:title="@string/context_map_get_coords">
+        android:title="@string/context_map_show_coords">
+    </item>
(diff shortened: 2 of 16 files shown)


moving-bits:
Really like the idea, thanks!

WRT to #12361 it would be nice to display the coordinates as well, so that you can give them to another geocacher, enter them in you car's navigation system etc.

Additionally, can you implement it for `UnifiedMap` as well? Would be great if we directly add new features to `UnifiedMap` as well to not loose functionality on some future switch and to avoid having to re-check all PR whether they include some stuff relevant for `UnifiedMap`.

fm-sys:
Here we go...

|||
-|-
grafik|grafik

moving-bits:
> > Additionally, can you implement it for UnifiedMap as well?
> 
> Does `UnifiedMap` already have a tap-handler interface at all? Otherwise it could be difficult to implement 😅

Not yet. It would be great if you could extend it as needed. Feel free to ask if you need any info on how things are related for `UnifiedMap`.

Until recently `UnifiedMap` did not much more than displaying a blank map, maybe tracks and routes. With recent changes it can be called from a cache detail page and displays a single cache, but no interaction yet. And no stored list of caches either.

> EDIT: For sure, I'm willing to port this feature to `UnifiedMap` - however, I think it is too early for this step. It would IMHO make much more sense to implement general geocache loading into the UnifiedMap first, before starting with such advanced stuff. I can open an issue as a reminder and assign it to myself.

Understandable, and ok for me as well. WRT to cache loading: I guess we should start discussing on how to handle different map types. I'd like to handle things a bit more explicitly as it is currently, therefore introduced `UnifiedMapType` with (potentially) more distinct types than today's map state. We need to collect a list of map types we want to support, and what our expectations are wrt to the map in relation to the different map types, e. g.
- setting zoom level
- setting position
- using filter
- setting title
- displaying or hiding waypoints for the caches
- ...

One more thing we should discuss is how to handle cache lists: Currently we take a list of caches that are part of a list, and call the map with a list of geocodes. The map then does not know any longer which list this is coming from. One thing being missed because of this is to be able to switch cache lists directly in map view. I see two possible alternatives for this:
- instead of passing a list of geocodes to the map, pass the info (map type) "cache list" and the list's id to the map, and let the map handle cache loading etc. - possible drawback: which filter should get applied to the map - map filter or last list filter?
- or pass a bunch of geocodes as today, but allow changing the list anyway, reloading the bunch of geocodes in this case.

I guess I'll open a discussion thread for those topics...

fm-sys:
> We could additionally display the coordinates in a second format: DDD° MM' SS.SSS", as this is a very common format for car navigation systems.

Maybe use our click-to-switch-coordinate-format behavior? DDD.DDDD° is BTW often used as well...

> And we could add a button to "start navigation" which would open our navigation tool selector

I would put it into the context menu, not the coordinate dialog. Right?

moving-bits:
sounds reasonable to me (both)

Would it be possible to integrate the coordinate display into the first popup as well (to avoid the extra tap), or would it look to crammed? E. g.:

N 52° 00.000
E 013° 00.000
\-----------------
Add waypoint
Create user-defined cache
Add point to route
Start navigation

fm-sys:
The context menu can only be styled with _very_ limited possibilities. One thing which would be possible is to set the coords as header text which would look quite good IMHO. 

A context menu with title would look like this:

Screenshot_20220608_135804_cgeo geocaching_copy_195x200


However, we cannot implement the coordinate format switching as the title doesn't has an OnClickListener. Do we really need to support multiple coordinate formats at all? The "my coordinates" textview in the MainActivity does only support one single format as well, and no one has ever asked...

moving-bits:
I was thinking of putting a single or multi-line entry for the coordinates into the menu, tappable for format switching, and adding a menu divider (which we have in our home screen's menu), if possible.

Using a header for this may lead to truncation (remember the cache name in cache popup on map?) as well as not being able (at least, not without some ugly hacks) to add an `OnClickListener` to it.

My use-case for having the DDD° MM' SS.SSS" format is getting coords to enter in car navigation (or to give them to somebody else for her/him to enter in their car navigation), but there may be other use-cases as well.

fm-sys:
> I was thinking of putting a single or multi-line entry for the coordinates into the menu

Never have seen that multi-line menu entries are possible... We could just add two menu entries. But this wouldn't look nice. 

> tappable for format switching

The menu will automatically close when clicking on it and I'm not aware of a possibility to suppress this behavior.

So if you really need multiple formats we will probably have to stay with the dialog.

fm-sys:
I just found out what android provides two totally different implementations for exactly the same UI component.

- Context Menu
- Popup Menu

Currently I'm using a Context Menu in this PR. However it seems, that the Popup menu can be utilized more flexible that a context menu which is bound to an specific activity. I will do some experiments...

Links for myself:
https://stackoverflow.com/questions/29726039/how-to-prevent-popup-menu-from-closing-on-checkbox-click/
https://stackoverflow.com/questions/13444594/how-can-i-change-the-position-of-where-my-popup-menu-pops-up

ztNFny:
I've added another menu item: "Add point to route (as first target)" to add a point at index 0 to the route.
Are you open to me pushing this to your branch? Otherwise I'll wait until you're done here.

I'll also take a look at an icon that fits better with our map icons.

fm-sys:
> Are you open to me pushing this to your branch? 

@ztNFny  Would be ok for me. 

Although I generally like to keep PR's as small as possible, it's IMHO fine in this case. Especially as I already added multiple new menu items anyway...

ztNFny:
@fm-sys pushed. If there's any problems with it please discard.

fm-sys:
Maybe we can use a shorter string? Problem is that menu entries can only be single line and will be truncated if necessary. 

 

E.g. `Add point to route start` or `Add point to beginning of route` instead of `Add point to route (as first target)`

Out of interest @ztNFny: 
Do you plan to add this functionality for caches/WPs as well (not only for empty locations)? This should however go into a separate PR and needs to be aligned with #13122.

ztNFny:
> Do you plan to add this functionality for caches/WPs as well (not only for empty locations)? 

Yes. But not sure about the UI part yet. It was easy enough to add here so I started with this. I don't want to make adding caches to routes harder, so need to investigate some possibilities here. Maybe have it in cache pop-up menu ...

fm-sys:
Yes, indeed difficult. Maybe a long click on "set as target"? However, this would be hard to find...

fm-sys:
Rebased.

BTW, I added a "navigate" menu entry as well:

grafik

PR #13115, opened 2022-06-09
Don't show unit test directory in nightly builds

fm-sys:
Currently, the unit test directory setting is displayed for developer and nightly builds. Show it only for developer builds instead.

Screenshot_20220610_000148_cgeo geocaching


fm-sys:
retest this please

PR #13116, opened 2022-06-10
Fix lint issues in cache info item

NotWoods:
## Description

- Replace RelativeLayout with LinearLayout for cleanup and perf, since the layout is simple
- Fix the lint error for ConstraintLayout Flow
- Fix the lint error for inflate params
- Use a custom class instead of ImmutablePair for clearer properties


eddiemuc:
The failing test in Jenkins is unrelated, but the Codacy issue needs to be fixed (non-final method parameters)

eddiemuc:
Oh, and many thanks for the contribution of course

PR #13119, opened 2022-06-10
Remove old GC map parser and other unused stuff

fm-sys:
Follow up to https://github.com/cgeo/cgeo/pull/13092#issuecomment-1152079783. 

While working on fixing the unit tests, I found nearly 1k lines of unused code deeply scattered in c:geo's code base. The code is a legacy of the old GC map implementation, which was replaced in 2019 by https://github.com/cgeo/cgeo/pull/7514 and many code parts were already removed in #11533.

I would like to target this change towards master, so that we have enough time to test it.

Adding "Do not merge" here as long as #13092 is WIP.


fm-sys:
Just resolved merge conflicts...

PR #13126, opened 2022-06-11
Annotation calls that should run on background thead

NotWoods:
## Description

Trying to catch blocking network calls on the main thread by adding `@WorkerThread` annotations to network code. This helps with readability and adds additional checks in Android Studio.

### DIFF
--- a/main/src/cgeo/geocaching/SearchResult.java
+++ b/main/src/cgeo/geocaching/SearchResult.java
@@ -19,6 +19,7 @@ import android.os.Parcelable;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.WorkerThread;
 import androidx.core.util.Consumer;
 
 import java.util.ArrayList;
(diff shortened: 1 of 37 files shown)


fm-sys:
Ideally, we should fix the network on main thread issue instead of not adding the annotation to the relevant methods (https://github.com/cgeo/cgeo/pull/13126/commits/0cb7107d6df3b8afc975985ac92041bd00772308). Not sure if this is possible without larger refactoring though...

NotWoods:
I'd prefer to leave that to a different PR since it requires a bigger refactoring

fm-sys:
> I'd prefer to leave that to a different PR since it requires a bigger refactoring

Fine for me. However, we should open a follow-up issue to keep track which methods are affected by "network on main-thread"

moving-bits:
> > I'd prefer to leave that to a different PR since it requires a bigger refactoring
> 
> Fine for me. However, we should open a follow-up issue to keep track which methods are affected by "network on main-thread"

I've created #13153 for this

Issue #13036, opened 2022-05-21
support targetSDK 31


moving-bits:
### Tell us your idea!

Play Store guidelines require `targetSDK` being set to 31 to be able to upload updates starting November 1st, 2022.

~(This would also enable us to merge some of the more recent changes to material libs waiting in our PR queue.)~ (this would be compileSDK)

For details see https://developer.android.com/google/play/requirements/target-sdk

### What solution would you suggest?

Migrate c:geo to `targetSDK` 31

### Is there an alternative?

Not really... Without setting `targetSDK` to 31 our app can continue in Play Store until November 1, 2023 until it gets delisted. Updates will be possible until November 2022 only, though, when not upgrading `targetSDK`.

### Additional Information

On a quick glance, there are several things that might hit us:

- Location: Users can request apps to retrieve only approximate location information. You must request the [ACCESS_COARSE_LOCATION](https://developer.android.com/reference/android/Manifest.permission#ACCESS_COARSE_LOCATION) permission any time you request [ACCESS_FINE_LOCATION](https://developer.android.com/reference/android/Manifest.permission#ACCESS_FINE_LOCATION).
- Intent filters: If your app contains [activities](https://developer.android.com/guide/components/activities/intro-activities), [services](https://developer.android.com/guide/components/services), or [broadcast receivers](https://developer.android.com/guide/components/broadcasts) that use [intent filters](https://developer.android.com/guide/components/intents-filters#Receiving), you must explicitly declare the [android:exported](https://developer.android.com/guide/topics/manifest/activity-element#exported) attribute for these components.
- Hibernation: Apps may be put into hibernation mode if they are not used over a period of time. In hibernation mode your app's runtime permissions and cache are reset, and you can't run jobs or alerts. You can check your [app’s hibernation status](https://developer.android.com/topic/performance/app-hibernation).
- [Pending intent mutability](https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability): You must specify the mutability of each PendingIntent object that your app creates.
- [Android App Links verification changes](https://developer.android.com/about/versions/12/behavior-changes-12#android-app-links-verification-changes): When using Android App Link verification, make sure that your intent filters include the BROWSABLE category and support the HTTPS scheme. (Relevant only when we start working on #13018)
[...]

PR #13136, opened 2022-06-11
raise compile SDK to 31

moving-bits:
## Description
- raise compile SDK to 31 to be able to upgrade some material libs to the most-recent versions
- leave target SDK at 30 for now to avoid behavioral changes (see #13036 for this)


moving-bits:
let's merge, as discussed in today's meetup

PR #13140, opened 2022-06-12
add resources from cache icon development

ztNFny:
add several resources created during development of cache icons. everything is in /project/ subfolder, so no impact on the application


ztNFny:
Codacy checks can be ignored, it's not actual code but just resources

Issue #12911, opened 2022-04-18
Allow selection of individual text inside log entries


eddiemuc:
### Tell us your idea!

C:geo allows selection of individual text inside Cache listings, but not inside log entries (only selection of whole text of a log entry). C:geo should allow selection of individual texts inside logs too. 

This would help me greatly e.g. to reselect a part of a log I just wrote to include in my next log (e.g. a common header for all logs of one day which I forgot to store in clipboard before posting or which was kicked out of clipboard by another selection in the meantime)

### What solution would you suggest?

C:geo should allow selection of individual texts inside logs too. 

### Is there an alternative?

Do nothing

ztNFny:
When writing your log in the menu there's "repeat last log" which might be an easier way to address your specific usecas

fm-sys:
> When writing your log in the menu there's "repeat last log" which might be an easier way to address your specific usecas


There are many other reasons why someone might want to select text of a log. I sometimes need this myself.

Currently I copy the whole log into the caches PN and can select individual parts there. But allowing individual text selection for logs would be great.

alan666notb:
Yes, I would like this as well.
And also I'd like to be able to search for (i.e. google) or translate the selected text.

PR #13150, opened 2022-06-14
fix #13141: make saved logs text non-selectable to fix double-click to edit

### LINKED ISSUE #13141: Offline Log: Open for edit needs two clicks (open context menu of "other" logs as well)
### Describe your problem!

Given an offline stored log (which should be uploaded later on).
If I want to edit it, I strangely have to click on the log text twice (select & edit?).

Same applies (as I got aware of right now) for the context menu of "other" logs.

### How to reproduce?

1. write an offline log
2. store it
3. try to edit it

You will need two clicks (where it was just one the other day - I guess before the "copy parts of log" stuff was invented?)

Edit window (resp. context menu for "other" logs) should open on first short click.

### Reproducible

Yes

### c:geo Version

2022.06.12-NB-ee97c24

### System information

_No response_

### Additional Information

_No response_

ztNFny:
Making log text selectable causes the first click to be consumed by the select-handler. Thus editing an offline log requires 2 clicks.
Doing setTextIsSelectable(false) on offline logs somehow doesn't work - no click handler is fired at all then, so instead setting setTextIsSelectable(true) on all but offline logs

### DIFF
--- a/main/res/layout/logs_item.xml
+++ b/main/res/layout/logs_item.xml
@@ -70,7 +70,6 @@
                 android:layout_height="wrap_content"
                 android:layout_gravity="left"
                 android:gravity="left"
-                android:textIsSelectable="true"
                 android:textSize="@dimen/textSize_detailsPrimary"
                 android:textColor="@color/colorText"/>
 
--- a/main/src/cgeo/geocaching/log/CacheLogsViewCreator.java
+++ b/main/src/cgeo/geocaching/log/CacheLogsViewCreator.java
@@ -177,6 +177,8 @@ public class CacheLogsViewCreator extends LogsViewCreator {
             holder.binding.author.setOnClickListener(new EditOfflineLogListener(getCache(), (CacheDetailActivity) getActivity()));
             holder.binding.logMark.setVisibility(View.VISIBLE);
             holder.binding.logMark.setImageResource(R.drawable.mark_orange);
+        } else {
+            holder.binding.log.setTextIsSelectable(true);
         }
     }
 


bekuno:
Retest this please

moving-bits (commented):
Technically it is working as expected: Offline logs can be edited with a single tap now, but texts from it can no longer be selected. For every other log type it works as before (context menu opening on second tap only, but texts can still be copied).

Not sure, though, whether the original feature request #12911 that resulted in the change leading to a second tap being necessary still is fulfilled after this change - @eddiemuc should comment on that.

An alternative may be described in the article linked in https://github.com/cgeo/cgeo/issues/13141#issuecomment-1156959876

eddiemuc:
> Not sure, though, whether the original feature request #12911 that resulted in the change leading to a second tap being necessary still is fulfilled after this change - @eddiemuc should comment on that.

I would see the original request as non-fulfilled after the change. Nevertheless the PR is very valid since original implementation caused other problems. A new technical solution has to be found.

moving-bits:
I've implemented the approach suggested in https://github.com/cgeo/cgeo/issues/13141#issuecomment-1156959876, see PR #13158 as an alternative to this PR. 

PR #13158
- will allow to open the edit mask for offline logs by tapping once on the log text, 
- open the context menu by tapping once on the log text for all other log types, 
- and keeps the ability to select texts in all log types.

ztNFny:
closed in favor of https://github.com/cgeo/cgeo/pull/13158

PR #13159, opened 2022-06-16
maps: put filter context into savedInstanceState (fix #13147)

### LINKED ISSUE #13147: map filter settings getting lost on screen rotation
### Describe your problem!

Ticket from support: Filter settings get lost on screen rotation of map

### How to reproduce?

- open search, search for a city, and let the caches be displayed on the map
- change some filter setting (e. g.: disable lab adventures)
- rotate screen

filter gets reset (in our example: lab adventures are being shown again)

filter should still be applied

### Reproducible

Yes

### c:geo Version

2022.06.06

### System information

_No response_

### Additional Information

ticket #312696

moving-bits:
## Description
Store filter context in savedInstanceState of `NewMap` and `CGeoMap` and restore it in `onCreate` to avoid losing filters on config changes for temporary filters, as described in #13147


eddiemuc (approved):
Lgtm, thanks for tackling

PR #13164, opened 2022-06-18
AL SearchV4 with support for hidden since parameter & relative date filter backend

ztNFny:
This combines 4 small changes to work towards a proper date filtering capability for ALs. I'm submitting it now to avoid the PR getting too big, even though GUI is still missing.

- support setting a relative date in Date filters
- set the "hidden" date on AL search results
- switch AL search to SearchV4
- support the hidden since X days parameter in AL search

### DIFF
--- a/main/src/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/cgeo/geocaching/connector/al/ALApi.java
@@ -232,6 +232,8 @@ final class ALApi {
             cache.setType(ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
             cache.setArchived(response.get("IsArchived").asBoolean()); // we get that even in passive mode!
+            cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
+            // cache.setFound(parseCompletionStatus(response.get("CompletionStatus").asInt()));
             // cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.CACHE));
             return cache;
@@ -346,5 +348,9 @@ final class ALApi {
         } catch (final ParseException e) {
             return new Date(0);
         }
+    }
+
+    private static boolean parseCompletionStatus(final int completionStatus) {
+        return (completionStatus == 0);
     }
 }
--- a/main/src/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/cgeo/geocaching/connector/al/ALApi.java
@@ -6,6 +6,7 @@ import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.LoadFlags.SaveFlag;
 import cgeo.geocaching.enumerations.WaypointType;
 import cgeo.geocaching.filters.core.BaseGeocacheFilter;
+import cgeo.geocaching.filters.core.DateRangeGeocacheFilter;
 import cgeo.geocaching.filters.core.DistanceGeocacheFilter;
 import cgeo.geocaching.filters.core.GeocacheFilter;
 import cgeo.geocaching.filters.core.OriginGeocacheFilter;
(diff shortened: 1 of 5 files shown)


main/src/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -98,28 +181,28 @@ protected static Collection<Geocache> searchByBBox(final Viewport viewport) {
...
     @NonNull
     protected static Collection<Geocache> searchByCenter(final Geopoint center) {
-        return searchByCenter(center, DEFAULT_RADIUS);
+        return searchByCenter(center, DEFAULT_RADIUS, null);
     }
 
     @NonNull
     @WorkerThread
-    private static Collection<Geocache> searchByCenter(final Geopoint center, final int distanceInMeters) {
+    private static Collection<Geocache> searchByCenter(final Geopoint center, final int distanceInMeters, final Integer daysSincePublish) {
```
eddiemuc, line 194: It somehow seems weird to me that a method called "searchByCenter" should take a parameter "daysSincePublish". Isn't it possible to somehow replace this function completely with calls to "searchByFilter"?

main/src/cgeo/geocaching/filters/core/DateRangeGeocacheFilter.java
```diff
@@ -34,6 +37,15 @@ public void setMinMaxDate(final Date min, final Date max) {
         this.dateFilter.setMinMaxDate(min, max);
     }
 
+    public void setRelativeMinMaxDays(final int daysBeforeToday, final int daysAfterToday) {
```
eddiemuc, line 40: The filter does not store whether the dates were set relative or absolute. This will lead to a strange situation: suppose today is the 12.6.22 and you set relative ranges to "previous 10 days until today". Then the filter would store the absolute min date 2.6.22 and the absolute may date 12.6.22. if you store this filter (e.g. as a named filter or current live filter or current offline filter) and open it e.g. 2 weeks later at 26.6.22, it would STILL filter for range 2.-12.6.22, although the user would expect it to filter for 16.-26.6. at this point in time.

I am afraid a flag and maybe the relative filter span needs to be stored and retrieved also from filter config to achieve the desired effect. Or am I missing something?

eddiemuc (commented):
One minor comment and one question about functionality (storage of "relative" filter info)

main/src/cgeo/geocaching/connector/al/ALApi.java
ztNFny: searchByCenter seems to be a common thing in the codebase, as opposed to searchByBBox. Didn't want to touch that.

main/src/cgeo/geocaching/filters/core/DateRangeGeocacheFilter.java
ztNFny: Hm, true... But as I see it to get the existing functionality of the DateRange filter combined with a relative range would require a helper filter that constructs a temporary DateRange filter (that's not added to the stored filter) whenever it's instantiated.

eddiemuc: My idea was: 
* the class DateFilter currently has two attributes: minDate and maxDate. This is enough to define absolute ranges
* One could add three more attributes: isRelative (boolean), minDaysToSubstract (int), maxDaysToAdd (int)
*  If flag "isRelative" is false then everything stays as is. If it is true, then all usages of "minDate" are replaces with "DateUtils.addDays(new Date(), -minDaysToSubstract), same for "maxDate"
* The configuration stored (done in "setConfig"/"getConfig") stores/retrieves the three new attributes as well

But I admit, this would be a little work to do...

moving-bits:
@eddiemuc 
Do you have time for an updated review, or shall we merge?

eddiemuc:
No time for real review. Looks ok on first glance. Please just merge

moving-bits:
ok, thanks for having a glance. will merge then

PR #13170, opened 2022-06-19
prevent long tap marker on OSM if map item got hit (fix #13168)

### LINKED ISSUE #13168: [UI] LongClick on cache behavior broken
### Describe your problem!

When performing a long click on a cache while using mapsforge map, a pin gets shown. However this pin should only appear when long-clicking on an empty location.

Screenshot_20220619_025235

### How to reproduce?

- Use Mapsforge
- long-click on a cache

=> ...an unexpected pin will appear. 

Same as it works for our Google map. No pin should appear. 

_No response_

### Reproducible

Yes

### c:geo Version

2022.06.19-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Prevent displaying a map marker on long tap if an existing map item got hit (OSM only).

### DIFF
--- a/main/src/cgeo/geocaching/maps/mapsforge/v6/TapHandler.java
+++ b/main/src/cgeo/geocaching/maps/mapsforge/v6/TapHandler.java
@@ -42,13 +42,16 @@ public class TapHandler {
         hitItems.clear();
     }
 
-    public void onLongPress(final Point tapXY) {
+    /** returns true if longTapContextMenu got triggered */
+    public boolean onLongPress(final Point tapXY) {
 
         final NewMap map = this.map.get();
 
         // show popup
         if (map != null && hitItems.isEmpty()) {
-            map.triggerLongTabContextMenu(tapXY);
+            map.triggerLongTapContextMenu(tapXY);
+            return true;
         }
+        return false;
     }
 }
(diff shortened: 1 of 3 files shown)


main/src/cgeo/geocaching/maps/mapsforge/v6/layers/TapHandlerLayer.java
```diff
@@ -50,7 +50,9 @@ public boolean onTap(final LatLong tapLatLong, final Point layerXY, final Point
     public boolean onLongPress(final LatLong tapLatLong, final Point layerXY, final Point tapXY) {
         longTabLatLong = tapLatLong;
         requestRedraw();
-        tapHandler.onLongPress(tapXY);
+        if (!tapHandler.onLongPress(tapXY)) {
+            longTabLatLong = null;
+        }
```
fm-sys, line 55: Works fine in the emulator. Although I think there should be a `requestRedraw()` after setting `longTabLatLong` to null. Or, why not inverting the if condition and setting the `longTapLatLong` only after the `onLongPress` had returned true?

main/src/cgeo/geocaching/maps/mapsforge/v6/TapHandler.java
```diff
@@ -42,13 +42,16 @@ public synchronized void finished() {
         hitItems.clear();
     }
 
-    public void onLongPress(final Point tapXY) {
+    /** returns true if map item was hit */
```
fm-sys, line 45: > returns true if map item was hit

Isn't it exactly the other way around? If no item was hit / if empty space was hit...

moving-bits: you're right, will change the comment

main/src/cgeo/geocaching/maps/mapsforge/v6/layers/TapHandlerLayer.java
moving-bits: I need to set `longTapLatLong` before calling `tapHandler.onLongPress()` as it depends on a non-null value for it, but moving the `requestRedraw()` after the `if` clause seems to be ok. Updated PR will follow.

PR #13171, opened 2022-06-19
move long tap popup slightly to the right to avoid overlap

moving-bits:
## Description
Move long-tap popup slightly to the right (by 10 dp) of the tapped location to avoid overlapping the marker:

|before|after|
|---|---|
|image|image|

(I had tried to hit the hospital's red cross sign with my tap. With this PR you'll be able to see the map marker's origin beside the popup again.)

### DIFF
--- a/main/src/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/cgeo/geocaching/maps/MapUtils.java
@@ -29,6 +29,7 @@ import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.SimplePopupMenu;
 import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.ClipboardUtils;
+import cgeo.geocaching.utils.DisplayUtils;
 import cgeo.geocaching.utils.FilterUtils;
 import static cgeo.geocaching.brouter.BRouterConstants.BROUTER_TILE_FILEEXTENSION;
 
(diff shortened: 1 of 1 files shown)


fm-sys:
Which API version are you using? This is how it looks like for me... (without this PR)  



Screenshot_20220619_131043_cgeo geocaching

Placing it below/above instead of besides makes IMHO more sense, as there might not be enough space. 

Like in this example:

Screenshot_20220619_131531_cgeo geocaching

For these cases I introduced the `padding` parameter, which however seems to take no effect in your case...

moving-bits:
> Which API version are you using? This is how it looks like for me... (without this PR)

API 31 Nexus 5X emulated device

> Placing it below/above instead of besides makes IMHO more sense, as there might not be enough space.

"not enough space" may happen on all sides, depending on where the user tapped. I guess that Android uses some mechanism to determine whether the menu can be displayed fully at the requested place, and changes position, if not. Not idea how to actively influence that mechanism, except by explcitly calculating a "good" position.

fm-sys:
> I guess that Android uses some mechanism to determine whether the menu can be displayed fully at the requested place, and changes position, if not

Androids placing magic works like this:
- If there is enough room in all directions it will be placed below the view. Left border of the view will match the left border of the popup.
- if there is not enough space below it will be placed above
- if there is not enough space on the right, the popup gets moved as much to the left as necessary.

For implementing a padding mechanism I added a transparent view with a specific height to the tap location. That way, we can add space below and above the tap location at the same time. However, it seems to not work on all android versions. I will have a look...

moving-bits:
Is offset given in pixels or in `dp`? If you need `dp`, then you'll probably have to do some calculation.

The following seems to work for me:

`.setPosition(new Point(tapX, tapY - offset), DisplayUtils.getPxFromDp(CgeoApplication.getInstance().getResources(), offset, 1.25f))`

image

fm-sys:
`Drawable.getIntrinsicHeight()` should actually return a pixel value. But maybe `getIntrinsicHeight` is wrongly used instead of the normal `getHeight` method?

moving-bits:
Hm, both `getIntrisicHeight` and `LayoutParams` expect pixels, should be fine.

For further testing I set the width of the anchor view to 100 and background color to red. Now it looks like this:

image

fm-sys:
Ok, the height looks correct. However, I really do not understand why the placing works fine on a Android 10 emulator but not on Android 12.

fm-sys:
Ah, I found the difference:
You have a filter set in your emulator enabled while I hadn't used a filter.

The XY coords we get from the touch event are relative while the popup menu coords are absolute.

An easy solution would be to transform the filter bar into an overlay above the map so that relative and absolute coords will always match.

PR #13182, opened 2022-06-22
Add ViewModels to cgeo

NotWoods:
## Description

This change replaces an async task with a viewmodel + livedata. This better mirrors modern Android practices, which recommend putting data inside a ViewModel. ViewModels survive configuration changes and the LiveData inside automatically cleans up when activities and fragments are stopped.

Extracting view data to a separate class makes it easier to replace the view implementation since logic and views are no longer tightly coupled. This would help with switching to Fragments/Jetpack Compose.

This is a fairly simple change to AboutActivity as a proof of concept for the rest of the app. Using LiveData already cleans up a lot of custom observer code from the activity. If you like it we can use this pattern elsewhere.

## Related issues

https://github.com/cgeo/cgeo/issues/9094#issuecomment-895090132

## Additional context

https://developer.android.com/topic/libraries/architecture/viewmodel


moving-bits:
@NotWoods 
Due to your other two PR there are now two conflicting files in this PR. Can you rebase/update once again, please?

moving-bits:
Thank for rebasing, and for this contribution!

PR #13184, opened 2022-06-22
Add lifecycle aware broadcast receiver

NotWoods:
## Description

Creates a helper class for listening to geocache refreshed broadcasts. The helper handles registering and unregistering a broadcast receiver.

## Additional context

https://developer.android.com/topic/libraries/architecture/lifecycle


fm-sys (approved):
LGTM, thanks for tackling these lifecycle things!

There are two unused imports in `AbstractMap`, could you fix those?

PR #13187, opened 2022-06-23
Actually use the new capabilities of search v2

### LINKED ISSUE #12415: Trackable indication no longer returned on search results
### Describe your problem!

I am quite sure we discussed this while implementing the new search in the past, I can however neither find a matching issue nor any final conclusion:

We do have some users on different support channels telling us, that they are unable to filter for caches containing trackables. Testing this out my final conclusion is, that (in contrast to the old search) the new search does either not contain this information in its results or we have a bug to map this into the results.

### How to reproduce?

- Perform search (e.g. nearby or take a location where you don't have any stored caches)
- Look at the result list (or try to filter it for "Has Trackable"

You will see, that in the result list no cache has a trackable indication (On the right above the favorite count) respectively the filter returns empty.
If you open some of the caches contained in the result you will sooner or later find a cache, that has trackables despite it was not shown in the result.

Show "has trackable" icon if the cache contains trackables.
If this information is not contained in the search, maybe there is some search option to trigger it?

### Reproducible

Yes

### c:geo Version

2021.12.24

### System information

_No response_

### Additional Information

_No response_

fm-sys:
- Parse trackable count (fix #12415)
- Parse cache region
- Parse attributes (this allows client side filtering of negative attributes, currently only positive attributes are supported)

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCMap.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCMap.java
@@ -228,7 +228,7 @@ public class GCMap {
             case NAME:
                 search.setKeywords(((NameGeocacheFilter) basicFilter).getStringFilter().getTextValue());
                 break;
-            case ATTRIBUTES: //TODO: does not work for v1!
+            case ATTRIBUTES:
                 search.addCacheAttributes(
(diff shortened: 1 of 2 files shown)


moving-bits:
Thanks for tackling those issues

PR #13188, opened 2022-06-24
fix long tap popup position (fix #13168)

### LINKED ISSUE #13168: [UI] LongClick on cache behavior broken

moving-bits:
## Description
Fixes map lon tap popup positioning problem described in #13171 and #13168 by changing filter bar into a map overlay, positioning the following map overlays relative to it.

This PR superseeds #13171.

### DIFF
--- a/main/res/layout/filter_sort_bar.xml
+++ b/main/res/layout/filter_sort_bar.xml
@@ -3,7 +3,9 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:orientation="vertical">
+    android:orientation="vertical"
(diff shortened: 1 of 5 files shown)


fm-sys (approved):
Can't test it currently, but I would say LGTM

PR #13192, opened 2022-06-25
fix typo in FileNameCreator (fix 13191)

### LINKED ISSUE #13191: Typo in String value for Prefix
Noticed the following where `prefix` seems to be incorrectly named as `praefix`.
Would we wanna refactor it to the correct name? Could raise a PR. 

https://github.com/cgeo/cgeo/blob/df7ee033cb43f9a08cad21f11001af574220c5d5/main/src/cgeo/geocaching/utils/FileNameCreator.java#L29

kvaithin:
Close #13191 

Fixes typo for prefix.


moving-bits:
ok to test

moving-bits:
Thanks @kvaithin for providing this clean-up, very welcomed!

PR #13195, opened 2022-06-25
avoid error log entries due to Toast.setGravity on API > 29

moving-bits:
## Description
API 30 changes `Toast` behavior:

Doing a `Toast.setGravity` is not only ignored visually starting API 30, but additionally leads to an error entry in the log file for every toast being decorated that way.

This PR skips `setGravity` for toasts when running on API 30+


fm-sys:
Do we have to set the gravity at all? AFAIK `Gravity.CENTER_HORIZONTAL | Gravity.BOTTOM` is the default anyway...

PR #13197, opened 2022-06-26
Use ViewModel for InstallWizardActivity

NotWoods:
## Description

Update the install wizard activity code to use ViewModel as its view controller, splitting logic away from the view code.

- Enums are now in their own file under a .wizard namespace
- Step logic is moved to the ViewModel
- The next outline button has been replaced with new logic to reuse the skip button


moving-bits:
@NotWoods 
To catch some of the Codacy issues upfront you may want to install `CheckStyle` plugin for Android Studio and do a run before the commit. It catches many of the things Codacy and lint are looking for, "missing final" and "wrong import order" being among them.

PR #13214, opened 2022-06-27
set dependabot limit to 30

### LINKED ISSUE #13193: dependabot not detecting dependency updates (or taking quite some time for it)
### Tell us your idea!

Discovered by chance that dependabot does not detect all lib updates or takes quite some time for it:
- At the very same time I was pushing a manual PR for updating the VTM dependency, dependabot opened a new PR regarding that VTM update, which had been released about a week ago (whereas it had detected the mapsforge update of the same base project nearly immediately)
- ok.http update is not detected yet, although nearly two weeks old

(There may be others, haven't checked)

Is there anything we can do about this, or do we simply have to live with it?

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

bekuno:
fixes #13193

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: daily
     # no limit :)
-    open-pull-requests-limit: 20
+    open-pull-requests-limit: 30
     ignore:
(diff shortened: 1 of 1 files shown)


moving-bits:
Reason for currently failing builds: see #13215

PR #13215, opened 2022-06-27
fix duplicate lifecycle.viewmodel class build error

moving-bits:
## Description
After updating some of the AndroidX classes suddenly kotlin version of lifecycle defaults to a different version, leading to class duplication. Fix it by explicitly stating the same version number for both, Java and Kotlin, version of the lib.


moving-bits:
It's still somewhat strange, as I had updated all five version numbers locally in my build.gradle before merging the five PR, and the build went without any problems locally. Only after merging the PR online, and updating locally to the new head of `master`, the build errors started locally as well... But luckily, SO was at our rescue easily.

PR #13217, opened 2022-06-27
force lifecycle-viewmodel-ktx version

bekuno:
force lifecycle-viewmodel-ktx version according to lifecycle-viewmodel version


moving-bits:
Duplicate of #13215

PR #13218, opened 2022-06-27
move CGeoMap BroadcastReceiver code to deferred map initialization (fix #13209)

### LINKED ISSUE #13209: maps.me navigation not working
>>> **See further down in this opening article for current state of workarounds.**

### Describe your problem!

Several user reports, that maps.me navigation is no longer working. It behaves as if maps.me is not installed (showing the installation link instead).

I can not reproduce on my device, but reporting users use a more recent maps.me version which is not available on Google Play for my account (yet).

Working maps.me version: 12.3.3-Google  (edit: fixed version number to 12.x)
Failing maps.me version:  13.0.71327 (and newer versions)

The API integration did not have any update since 2019, thus I suspect a problem on their side with more recent maps.me versions.

### How to reproduce?

- Have maps.me app installed
- Open a cache
- Select Navigate - Maps.me

It shows the message to install maps.me although its already installed:
image

(Edit: This is a message coming from the maps.me integration API code, not from c:geo itself. This messagebox is triggered by the integration code not being able to detect the installed maps.me app.)

Maps.me should open with target set to the cache coordinates.

### Reproducible

Yes

### c:geo Version

2022.06.06

### Additional Information

Support tickets 726566, 126520, 718492, 621874, 419584 (and others)

Edit: as of 2022-07-27 we have app. 50 support tickets regarding the integration, but still no reaction from maps.me

## Possible workarounds

Users of c:geo have provided us with individual workarounds working for them. Both require some technical knowledge and may not work in your specific setting. While both descriptions look reasonable, we have not tested them, so if trying to implement one of these in your environment you do so at your own risk. If you feel unsafe with any of the steps involved, please do not apply the workarounds.

### Workaround 1: Use older version of maps.me app

As versions 12.x have worked, and only versions 13.x and up contains the assumed bug in maps.me app, you can try to replace your current installation of maps.me with an older version, 12.x max.

- [Download version 12.x of maps.me app from apkmirror.com](https://www.apkmirror.com/?post_type=app_release&searchtype=apk&s=maps.me) - Make sure to select the 12.x version (still available as of 2022-07-30)
- Check the downloaded file for viruses etc. with a scanner of your choice (eg. [Virustotal](https://www.virustotal.com))
- Copy the downloaded apk file to your mobile and install it. You may have to temporarily allow installation of apk from unknown sources - make sure to disable this option after applying this workaround.

### Workaround 2: Import caches using kml:

Instead of directly opening a cache from within c:geo into maps.me, another user exported and converted bulk data, and imported it into maps.me:

- export caches as GPX
- convert GPX file to kml
- open this kml file with maps.me
- maps.me will create a bookmark file from it and display the caches

moving-bits:
## Description
The new BroadcastReceiver init code from #13184 had been placed in CGeoMap class initialization, but needs to be called after async map view initialization has been finished. Therefore moving it to `initializeMap()`.

### DIFF
--- a/main/src/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/cgeo/geocaching/maps/CGeoMap.java
@@ -418,19 +418,6 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
 
     public CGeoMap(@NonNull final MapActivityImpl activity) {
         super(activity);
-        // only add cache if it is currently visible
-        getActivity().getLifecycle().addObserver(new GeocacheRefreshedBroadcastReceiver(mapView.getContext()) {
-            @Override
-            protected void onReceive(final Context context, final String geocode) {
-                final Geocache cache = DataStore.loadCache(geocode, LoadFlags.LOAD_CACHE_OR_DB);
-
-                // only add cache if it is currently visible
-                if (caches.remove(cache)) {
-                    caches.add(cache);
-                    displayExecutor.execute(new DisplayRunnable(CGeoMap.this));
-                }
-            }
-        });
     }
 
     protected void countVisibleCaches() {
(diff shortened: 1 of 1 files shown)


moving-bits:
@NotWoods 
Can you cross-check, if your lifecycle observer changes introduced with #13184 still work after this change?

I needed to move the observer's init code for `CGeoMap` a bit, as its current placement lead to `CGeoMap` crashing, as the `mapView` get's initialized async and isn't available in the map contructor yet.

NotWoods:
Will double check later but this should be fine if it's called in onCreate

moving-bits:
Thanks for double-checking.

I've placed it even later than `onCreate` currently, in `initializeMap()`, being called when map initialization is finished;

```        
onCreate() {
   ...
   mapView.onMapReady(() -> initializeMap(trailHistory));
   ...
}
```

But maybe placing it directly in `onCreate` will work as well, haven't tried (and won't be able to check today).

NotWoods:
The deferred initialization will work but might miss messages from the broadcast receiver. `onCreate` should also work well, especially since it already references the activity.

moving-bits:
ok, let's give it a try - have prepared #13221 for it

PR #13234, opened 2022-07-02
set contacts compileSDK to 31

moving-bits:
## Description
Set compileSDK for cgeo-contacts to 31, same value as in the main cgeo project.
(Required to mitigate build errors on rebuild)


bekuno (approved):
LGTM

PR #13238, opened 2022-07-03
Make notifications less disturbing

fm-sys:
IIRC there were support tickets claiming that the background downloader makes too much sound. This makes notifications silent and less intrusive. (Proximity notifications are not affected)

Furthermore, fix a bug where the "download pending" icon was not shown correctly.


moving-bits:
Looks fine generally, thanks. One tiny idea: How about keeping the more noticable notification for map downloads? Those can take quite some time due to the large file sizes, so it might be nice to notice more prominently that such a download has finished. (Or instead of using the download type make it dependent on file size, maybe everything larger than 200 MB?)

PR #13242, opened 2022-07-04
Automatically stop speech service on destroy

NotWoods:
## Description
Using lifecycle observers to automatically check for `onDestroy`. This lets us make the start/stop functions private in SpeechService.


fm-sys:
Nice refactoring, saves quite some boilerplate code... 👍

PR #13243, opened 2022-07-04
Start/stop local broadcast receiver automatically using create/destroy

NotWoods:
## Description
More work decoupling logic from the activity methods. Helper class for local broadcast receivers that cleans itself up on activity destroy.

### DIFF
--- a/main/src/cgeo/geocaching/utils/LifecycleStartedBroadcastReceiver.java
+++ b/main/src/cgeo/geocaching/utils/LifecycleStartedBroadcastReceiver.java
@@ -1,0 +1,38 @@
+package cgeo.geocaching.utils;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.IntentFilter;
+
+import androidx.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


main/src/cgeo/geocaching/utils/LifecycleStartedBroadcastReceiver.java
```diff
@@ -0,0 +1,38 @@
...
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.IntentFilter;
+
+import androidx.annotation.NonNull;
+import androidx.lifecycle.DefaultLifecycleObserver;
+import androidx.lifecycle.LifecycleOwner;
+import androidx.localbroadcastmanager.content.LocalBroadcastManager;
+
+public abstract class LifecycleStartedBroadcastReceiver extends BroadcastReceiver implements DefaultLifecycleObserver {
```
fm-sys, line 12: IMHO the name `LifecycleStartedBroadcastReceiver` is kinda misleading. What about `Life cycleAwareBroadcastReceiver`?

PR #13249, opened 2022-07-07
Remove one-off methods from Geocache

NotWoods:
## Description

Trying to make Geocache more of a data container and less of a god class. This change removes some methods dedicated to sharing that were each only used in one location, and replaces the method `isPastEvent` with the equivalent CalendarUtils method.


fm-sys:
On the one hand I understand your purpose, on the other hand these utility methods might be hard to find especially as no one really knows the whole codebase of cgeo. From time to time we are discovering code duplications where different people had implemented one and the same functionality into different classes, cause they were not aware that there already is an utility method for it. So I'm a bit undecided how to proceed here...

NotWoods:
Geocache also handles so many tasks that it becomes pretty difficult to change anything around cache loading. Perhaps there can be a GeocacheUtils class that has a documentation link?

murggel:
I created PR #15046 which rebases this PR to the current master

eddiemuc:
Superseeded by #15046. Thanks to @murggel and @NotWoods for your contributions.

PR #13259, opened 2022-07-21
[map downloader] fix progress indicator getting lost on scroll

fm-sys:
Our download selector activity uses a recycler view, which causes the progress bar to vanish as soon as the view is scrolled out of the screen. With this PR, the progress bar gets reloaded, as soon as the view becomes visible again.


fm-sys:
retest this please 

_CI build was failing, but it's most probably unrelated to the PR. This comment will trigger a rebuild._

PR #13267, opened 2022-07-24
use radius from filter for OK API search

ztNFny:
use user-defined radius in requests to OK API search instead of hardcoded 200km

### DIFF
--- a/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -328,7 +328,7 @@ final class OkapiClient {
         final DistanceGeocacheFilter df = GeocacheFilter.findInChain(filters, DistanceGeocacheFilter.class);
         if (df != null) {
             searchCoords = df.getEffectiveCoordinate();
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/connector/oc/OkapiClient.java
```diff
@@ -848,6 +862,8 @@ private static void parseCoreCache(final ObjectNode response, @NonNull final Geo
         //set basic properties which are constant / not used for OC platforms
         cache.setPremiumMembersOnly(false);
         cache.setUserModifiedCoords(false);
+
+        DataStore.saveCache(cache, EnumSet.of(SaveFlag.CACHE));
```
fm-sys, line 866: Why should store the search results in the DB? IMHO, there shouldn't be any direct DB access from individual connectors...

ztNFny: Every connector has the DataStore.saveCache in its parse() method:
image

fm-sys: Sorry, never mind. Was just wondering why you added it... Thanks for the explanation!

PR #13286, opened 2022-07-28
Adjust info text for 'solved mystery' status filter (fix #13285)

### LINKED ISSUE #13285: Filtering for "solved mysteries only" seems to be complicated
### Describe your problem!

Coming from a user ticket on support I tried to set a list filter for "only show solved mysteries".
Looks straight forward by adding a status filter, leaving every option at its default, but set "solved mystery" to "yes".

But this does not work as expected - it does filter the list somehow, but you will see caches of all sorts, eg. traditional, letterbox, multi etc. - It seems to filter mysteries having no final coords, though. So despite the title "Solved mysteries" it's more of a "hide unsolved mysteries" filter.

You can achieve the original goal by either
- setting "solved mystery" to "no" and negate the filter
- or by setting "solved mystery" to "yes" and additionally set a cache type filter for type=mystery

Is this behavior intended? It seems a bit complicated to me. Or is there any shortcut I've missed?

### How to reproduce?

Have a list of caches, set a fresh filter in advanced mode, type=status, "solved mystery"=yes, leave all other at default

unsolved mysteries are excluded

display only solved mysteries, leave out any other cache type

### Reproducible

Yes

### c:geo Version

2022.06.06

### System information

_No response_

### Additional Information

_No response_

moving-bits:
Adjusts the info text for "solved mystery" status filter as described in #13285.
Also removes a leading space before the second sentence which lead to unwanted indentation.

@eddiemuc: Is this text ok for you?

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -627,7 +627,7 @@
     <string name="cache_filter_status_select_label_has_offline_found_log">Has Offline Found Log</string>
     <string name="cache_filter_status_select_label_solved_mystery">Solved Mystery</string>
     <string name="cache_filter_status_select_label_has_user_defined_waypoints">Has User Defined Waypoints</string>
-    <string name="cache_filter_status_select_infotext_solved_mystery">This filter setting will only affect mystery caches.\n\n A mystery is considered \'solved\' if it has either changed coordinates or a valid final waypoint filled with coordinates.</string>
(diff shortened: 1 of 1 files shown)


eddiemuc (approved):

moving-bits:
retest this please

moving-bits:
@mucek4:
It's the "Could not create ADB Bridge." error again. Can you restart the CI? Thanks.

fm-sys:
retest this please

PR #13303, opened 2022-08-03
alcIsLinear

baiti:
an attempt to handle the Linear vs Random for lab caches. Pls review. There must be some bug as I can't retrieve the property from the cache object, it always returns `false` even when upon setting it the Log() shows it was `true`. Once we have fixed this, I plan to add `RatingsAverage` and `RatingsTotalCount` according to the same pattern.

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -1172,6 +1172,9 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             }
 
             details.add(R.string.cache_type, cache.getType().getL10n());
+            if (cache.getType() == CacheType.ADVLAB) {
+                details.addAlcMode(cache);
+            }
             details.addSize(cache);
             activity.addContextMenu(details.add(R.string.cache_geocode, cache.getShortGeocode()).valueView);
             details.addCacheState(cache);
--- a/main/src/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/cgeo/geocaching/connector/al/ALApi.java
@@ -346,6 +346,8 @@ final class ALApi {
             cache.setCoords(new Geopoint(location.get(LATITUDE).asText(), location.get(LONGITUDE).asText()));
             cache.setType(ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
+            cache.setVotes(response.get("RatingsTotalCount").asInt());
+            cache.setRating(response.get("RatingsAverage").floatValue());
             cache.setArchived(response.get("IsArchived").asBoolean());
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             // cache.setFound(parseCompletionStatus(response.get("CompletionStatus").asInt())); as soon as we're using active mode
@@ -377,12 +379,21 @@ final class ALApi {
             cache.setCoords(new Geopoint(location.get(LATITUDE).asText(), location.get(LONGITUDE).asText()));
             cache.setType(ADVLAB);
             cache.setSize(CacheSize.getById("virtual"));
+            cache.setVotes(response.get("RatingsTotalCount").asInt());
+            cache.setRating(response.get("RatingsAverage").floatValue());
             // cache.setArchived(response.get("IsArchived").asBoolean()); as soon as we're using active mode
             // cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode
             cache.setDisabled(false);
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             cache.setOwnerDisplayName(response.get("OwnerUsername").asText());
             cache.setWaypoints(parseWaypoints((ArrayNode) response.path("GeocacheSummaries")), true);
+            final boolean isLinear = response.get("IsLinear").asBoolean();
+            if (isLinear) {
+                cache.setAlcMode(1);
+            } else {
+                cache.setAlcMode(0);
+            }
+            Log.d("_AL mode from JSON: IsLinear: " + cache.isLinearAlc());
             cache.setDetailedUpdatedNow();
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.DB));
             return cache;
@@ -403,7 +414,7 @@ final class ALApi {
         for (final JsonNode wptResponse : wptsJson) {
             stageCounter++;
             try {
-                final Waypoint wpt = new Waypoint(wptResponse.get(TITLE).asText(), WaypointType.PUZZLE, false);
+                final Waypoint wpt = new Waypoint("S" + String.valueOf(stageCounter) + ": " + wptResponse.get(TITLE).asText(), WaypointType.PUZZLE, false);
                 final JsonNode location = wptResponse.at(LOCATION);
                 final String ilink = wptResponse.get("KeyImageUrl").asText();
                 final String desc = wptResponse.get("Description").asText();
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -211,7 +211,8 @@ public class DataStore {
                     "cg_caches.watchlistCount," +  // 42
                     "cg_caches.preventWaypointsFromNote," +  // 43
                     "cg_caches.owner_guid," +  // 44
-                    "cg_caches.emoji";                       // 45
+                    "cg_caches.emoji," +       // 45
+                    "cg_caches.alcMode";       // 46
 
     /**
      * The list of fields needed for mapping.
@@ -228,7 +229,7 @@ public class DataStore {
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
-    private static final int dbVersion = 98;
+    private static final int dbVersion = 99;
     public static final int customListIdOffset = 10;
 
     /**
@@ -262,7 +263,8 @@ public class DataStore {
             95, // add table to store custom filters
             96, // add preventAskForDeletion to cg_lists
             97, // rename ALC caches' geocodes from "LC" prefix to "AL" prefix
-            98  // add table cg_variables to store cache variables
+            98, // add table cg_variables to store cache variables
+            99  // add alcMode to differentiate Linear vs Random
     ));
 
     @NonNull private static final String dbTableCaches = "cg_caches";
@@ -333,7 +335,8 @@ public class DataStore {
             + "watchlistCount INTEGER DEFAULT -1,"
             + "preventWaypointsFromNote INTEGER DEFAULT 0,"
             + "owner_guid TEXT NOT NULL DEFAULT '',"
-            + "emoji INTEGER DEFAULT 0"
+            + "emoji INTEGER DEFAULT 0,"
+            + "alcMode INTEGER DEFAULT 0"
             + "); ";
     private static final String dbCreateLists = ""
             + "CREATE TABLE IF NOT EXISTS " + dbTableLists + " ("
@@ -1672,6 +1675,15 @@ public class DataStore {
                         }
                     }
 
+                    // add alcMode to cg_caches
+                    if (oldVersion < 99) {
+                        try {
+                            createColumnIfNotExists(db, dbTableCaches, "alcMode INTEGER DEFAULT 0");
+                        } catch (final SQLException e) {
+                            onUpgradeError(e, 99);
+                        }
+                    }
+
                 }
 
                 //at the very end of onUpgrade: rewrite downgradeable versions in database
@@ -2243,6 +2255,7 @@ public class DataStore {
         values.put("preventWaypointsFromNote", cache.isPreventWaypointsFromNote() ? 1 : 0);
         values.put("owner_guid", cache.getOwnerGuid());
         values.put("emoji", cache.getAssignedEmoji());
+        values.put("alcMode", cache.getAlcMode());
 
         init();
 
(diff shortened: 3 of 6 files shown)


eddiemuc:
Without having looked into actual changes, I noticed that you added a Geocache property but did not change anything in DataStore. So most likely this new property will not be stored in database if the lab cache is saved offline - correct?

If you want to keep it a separate new prop, at least this has to be done. This will include a database change (new field).

In a first step, maybe it is easier to just include everything in already existing fields (e.g. description) and make it separate fields later?

baiti:
> Without having looked into actual changes, I noticed that you added a Geocache property but did not change anything in DataStore. So most likely this new property will not be stored in database if the lab cache is saved offline - correct?

Yes, correct.

> 
> If you want to keep it a separate new prop, at least this has to be done. This will include a database change (new field).
>

Would it be OK to add add a new field to the database? What implications would that have?
 
> In a first step, maybe it is easier to just include everything in already existing fields (e.g. description) and make it separate fields later?

If I just add it to a string, the property couldn't be programmatically queried later, right?

baiti:
on new properties in the db:

I could re-use (or abuse for that sake) an already existing field that has no meaning for LCs, e.g. `premiumMembersOnly`. Would that be ok? We'd only do something like this in order to not mess with the database schema.

eddiemuc:
> Would it be OK to add add a new field to the database? What implications would that have?

If there's need for it, of course the database can be changed/extended. All db stuff is encapsulated in the very big class DataStore, you will find all the create statements for tables as well as the SQL stuff for querying and updating the db there (rather basic, c:geo does not use any fancy persistence framework).

You would need to get familiar with this class if you want to extend the db. In addition to "db setup" for a new installation you would also need to provide "migration" code to update existing databases. You would effectively create a new database version .

> If I just add it to a string, the property couldn't be programmatically queried later, right?

Correct. But maybe as of now this is not yet needed.

> I could re-use (or abuse for that sake) an already existing field that has no meaning for LCs, e.g. premiumMembersOnly.

I would prefer not to start messing up the database field meanings like this. Once this catches on we will get into database hell  pretty quickly.

ztNFny:
I'm unable to do a review the next 2 weeks, no access to a dev env

main/src/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -383,8 +383,15 @@ private static Geocache parseCacheDetail(final JsonNode response) {
...
-            cache.setDetailedUpdatedNow();
+            final boolean isLinear = response.get("IsLinear").asBoolean();
+            if (isLinear) {
+                cache.setAlcMode(1);
+            } else {
+                cache.setAlcMode(0);
+            }
+            Log.e("_AL mode from JSON: IsLinear: " + cache.isLinearAlc());
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.DB));
+            cache.setDetailedUpdatedNow();
```
moving-bits, line 394: Not sure what impact this has, but why did you move this line _after_ `DataStore.saveCache(...)`?

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -2243,6 +2255,7 @@ public static boolean storeIntoDatabase(final Geocache cache) {
         values.put("preventWaypointsFromNote", cache.isPreventWaypointsFromNote() ? 1 : 0);
         values.put("owner_guid", cache.getOwnerGuid());
         values.put("emoji", cache.getAssignedEmoji());
+        values.put("alcMode", cache.getAlcMode());
```
moving-bits, line 2258: This line is fine, but you are missing to read this value on loading the cache, see method `createCacheFromDatabaseContent`, where you should add a line `cache.setAlcMode(cursor.getInt(46));` - this seems to be the culprit why the mode value gets lost on loading the cache from database.

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
```diff
@@ -182,6 +183,15 @@ public void addSize(final Geocache cache) {
         }
     }
 
+    public void addAlcMode(final Geocache cache) {
+        Log.e("_AL add mode to view: " + cache.isLinearAlc());
```
moving-bits, line 187: When you are finished with this PR, everything working, you should remove all those debugging `Log.e(...)` statements before finalizing this PR.

main/src/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1173,6 +1173,9 @@ public void setContent() {
 
             details.add(R.string.cache_type, cache.getType().getL10n());
             details.addSize(cache);
+            if (cache.getType() == CacheType.ADVLAB) {
```
moving-bits, line 1176: I would add the mode info above the size, to have it closer to the cache type (which I sort of regard the mode info being part of).

moving-bits (changes requested):
Looks fine to me generally, see a few remarks. I guess I may have found the culprit why mode state gets lost on loading the cache from db.

main/src/cgeo/geocaching/connector/al/ALApi.java
baiti: I felt if these lines are strictly executed synchronously that the term "...UpdatedNow" (past tense) is only really true **after** the `saveCache()` returns.

main/src/cgeo/geocaching/storage/DataStore.java
baiti: Thx much ... will go after this tomorrow

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
baiti: Ideally I'd put those into comments but I learned in a previous PR that this is considered bad practice. Would it be ok to leave them in as `Log.d()`

main/src/cgeo/geocaching/CacheDetailActivity.java
baiti: ok ... will do

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
moving-bits: `Log.d()` is fine as well, but `Log.e()` will break c:geo for users running c:geo in debug mode (settings => system => debug mode), as every `Log.e()` will terminate c:geo in this case.

main/src/cgeo/geocaching/connector/al/ALApi.java
moving-bits: Haven't checked, but doesn't `cache.setDetailedUpdateNow()` update some data fields which need to be persisted as well? Or are they in-memory fields only?

main/src/cgeo/geocaching/storage/DataStore.java
baiti: Done

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
baiti: > `Log.d()` is fine as well, but `Log.e()` will break c:geo for users running c:geo in debug mode (settings => system => debug mode), as every `Log.e()` will terminate c:geo in this case.

ahhh ... didn't know that. Thx for telling me.

main/src/cgeo/geocaching/CacheDetailActivity.java
baiti: Done

main/src/cgeo/geocaching/ui/CacheDetailsCreator.java
moving-bits: No problem. `Log.e()` is very convenient during development, as you can easily filter for that in LogCat window and "traffic" for `Log.e()` is way less than for `Log.d()`. So my usual way is to use `Log.e()` during development and either remove it or switch it to `Log.d()` or `Log.i()` before finalizing my PR, depending on whether I want to see those messages always (`Log.i()`) or only when the user has set c:geo into debugging mode (`Log.d()`).

main/src/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -377,12 +379,21 @@ private static Geocache parseCacheDetail(final JsonNode response) {
...
+            cache.setRating(response.get("RatingsAverage").floatValue());
             // cache.setArchived(response.get("IsArchived").asBoolean()); as soon as we're using active mode
             // cache.setFound(response.get("IsComplete").asBoolean()); as soon as we're using active mode
             cache.setDisabled(false);
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             cache.setOwnerDisplayName(response.get("OwnerUsername").asText());
             cache.setWaypoints(parseWaypoints((ArrayNode) response.path("GeocacheSummaries")), true);
+            final boolean isLinear = response.get("IsLinear").asBoolean();
+            if (isLinear) {
+                cache.setAlcMode(1);
```
fm-sys, line 392: What about using an `Enum` instead of magic numbers? The `Enum` could hold the user displayable strings as well, so no additional mapping would be necessary...

fm-sys (commented):
I've added a small suggestion how the code could be improved...

Furthermore: Do we need to add `alcMode` to `GeGeocache.gatherMissingFrom`?

moving-bits (approved):
LGTM, thanks for diving into this!

main/src/cgeo/geocaching/connector/al/ALApi.java
baiti: sure, could be done, however, I had actually envisioned to turn this into bitfield operations if, and only if more modes are needed. I was looking at: https://docs.oracle.com/javase/8/docs/api/java/util/BitSet.html

To instantiate: `BitSet bitField = new BitSet(); //can specify size`

To check if bit n is true `bitField.get(n);`
To set bit n `bitField.set(n, value); //can also use set(int) and clear(int) instead`

I had hoped, I could later, once needed and if needed at all cast the `alcMode` which is an integer now into a `BitSet` object. The main reason for this would be to avoid adding more columns to the database schema.

baiti:
> LGTM, thanks for diving into this!

Who's going to do the merging?

moving-bits:
> Who's going to do the merging?

Me :-) (or others from the team), just wanted to wait a bit for a possible reply of @fm-sys. I guess I'll give him a few hours, and merge after that (and we could do fine-tuning later on, if needed), and start preparing a new beta version after that.

moving-bits:
@baiti 
Besides using enums or not, have you seen the other comment from @fm-sys?

> Furthermore: Do we need to add `alcMode` to `GeGeocache.gatherMissingFrom`?

From a quick glance into the source I think handling `alcMode` should be added to `Geocache.gatherMissingFrom` as well.

main/src/cgeo/geocaching/connector/al/ALApi.java
fm-sys: Hm, don't really understand why we should use a BitSet.

I had thought of an enum with two parameters, `int id` and `@StringRes int userDisplayableName`. You can still store the id value as int inside the DB. However it's IMHO important that only one class needs to remember the actual mapping. 

We have similar code patterns at different places in cgeo, I will search for a meanful example...

baiti:
> I've added a small suggestion how the code could be improved...

Thx much. Yes, this could be done, I'd say let's decide and make an appropriate change if more "modes" come up for ALCs.

> 
> Furthermore: Do we need to add `alcMode` to `GeGeocache.gatherMissingFrom`?

That is a good question. I had looked at this and AFAICT there are only lazy initialized properties in there. Lazy initialized means the constructor inititialzes them to `null` and the `gatherMissingFrom` explictly goes after all of those.

`alcMode` is initiallized explicitly to `0` and overriden via the LC connector, if needed.

From all the tests I ran, "it worked" ... but I am still an absolute beginner with Java and my understanding of the cgeo code is still very limited.

main/src/cgeo/geocaching/connector/al/ALApi.java
moving-bits: I will merge now anyway to start preparing a beta version. Switch from "magic numbers" to "enum" could be added later without causing harm.

baiti:
> @baiti Besides using enums or not, have you seen the other comment from @fm-sys?
> 
> > Furthermore: Do we need to add `alcMode` to `GeGeocache.gatherMissingFrom`?
> 
> From a quick glance into the source I think handling `alcMode` should be added to `Geocache.gatherMissingFrom` as well.

How would I know if it needs to be added there? In other words, what would not work it it isn't added there?

eddiemuc:
> > @baiti Besides using enums or not, have you seen the other comment from @fm-sys?
> > 
> > > Furthermore: Do we need to add `alcMode` to `GeGeocache.gatherMissingFrom`?
> > 
> > From a quick glance into the source I think handling `alcMode` should be added to `Geocache.gatherMissingFrom` as well.
> 
> How would I know if it needs to be added there? In other words, what would not work it it isn't added there?

As far as I remember, `gatherMissingFrom` is used to merge two versions of the same cache when it is loaded from two different sources and those sourcea supply different parts of information and may have different ages.

The most prominent example is probably the Live Map if there's a cache displayed which is also stored offline in the database. Then we have two "sources" for its data:
* The one from the offline db: more complete, but probably outdated
* The one from the live call: most recent info, but not all fields filled (e.g. attributes or description etc)

`gatherMissingFrom` is used to "merge" the infos of one such Geocache object into the other, in the best possible way. There are surely some heuristics applied here and there.

Issue #13110, opened 2022-06-08
improve individual route rendering for coordinate-only-points


fm-sys:
### Tell us your idea!

When coordinate-only-points points are contained in the individual route, they cannot be removed easily.

Furthermore, the route seem to end in nowhere.

### What solution would you suggest?

These points should show a pin marker, so that we are able to remove these points again via long-click. 

![Screenshot_20220608_121243_cgeo geocaching_copy_324x702](https://user-images.githubusercontent.com/64581222/172592162-493cc0a2-2f3b-4cb8-82d3-fe58d473b3f6.jpg)

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/13104#issuecomment-1149736900_

moving-bits:
Maybe a follow-up issue of #13122

PR #13304, opened 2022-08-03
[NewMap] improve individual route rendering for coordinate-only-points

fm-sys:
For screenshots, see https://github.com/cgeo/cgeo/issues/13110#issuecomment-1204235834

@moving-bits as you've done quite some work in our map implementations recently + you're the inventor of "individual route": 

Do you think it's a good idea to squeeze coords-only routing points into the existing `GeoitemRef` model or should there be something own? Interestingly, our Google map doesn't use `GeoitemRef` at all, but is doing some really weird magic... So I'm a bit lost how to implement the long-click functionality there.

And are there any plans how taps will be handled in `UnifiedMap`? Ideally, there could be something way more flexible, allowing other map items like tracks to be clickable as well. We had discussed this some time ago in #11531

### DIFF
--- a/main/res/drawable/marker_routepoint.xml
+++ b/main/res/drawable/marker_routepoint.xml
@@ -1,0 +1,15 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="13dp"
+    android:height="13dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:pathData="M12,12m-12,0a12,12 0,1 1,24 0a12,12 0,1 1,-24 0"
+        android:fillColor="@color/dotBg_waypointOutline"/>
+    <path
+        android:pathData="M12,12m-10,0a10,10 0,1 1,20 0a10,10 0,1 1,-20 0"
+        android:fillColor="@color/just_black"/>
+    <path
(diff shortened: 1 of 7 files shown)


moving-bits:
It would be nice if `GeoitemRef` (or whatever else we use) knows what type of ref it is instead of relying on some internal or external "magic". (That's why `RouteItem` uses a `RouteItemType`.) But actually I'm not that deep into `GeoitemRef`, so I cannot give a valid answer to your question.

Regarding tap handler for `UnifiedMap`: I have not thought of it yet, and have not been able to work substantially on `UnifiedMap` over the last two months... If you have an idea on how to implement this, you are very welcomed to give it a try. `UnifiedMap` is our chance to make things a bit more flexible, while being ... unified. :-) 

If you have any questions on how things are implemented in `UnifiedMap` so far, feel free to ask. Or if you think, something should be implemented in a different way, go ahead. Let's do this as a shared code project - helps me to prevent my worst errors, and spreads knowledge of how things are implemented in `UnifiedMap` at the same time.

moving-bits:
> And are there any plans how taps will be handled in `UnifiedMap`?

`UnifiedMap` has tap handling implemented meanwhile, see `UnifiedMapActivity.onTap()` and `UnifiedMapActivity.handleTap()`.

Issue #12210, opened 2021-12-01
Hillshading for Mapsforge


ztNFny:
### Tell us your idea!

c:geo should offer hillshading on (offline) maps which is the coloring of the map to show elevation changes.

Locus and OruxMaps both do that on top of mapsforge offline vector maps, so it is available and working, but I'm having a hard time finding documentation on it.

![image](https://user-images.githubusercontent.com/1258173/144286497-6e3de5dc-edcd-4960-8053-5a65af77e57a.png)

### Additional Information

https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map/src/main/java/org/mapsforge/map/rendertheme/renderinstruction/Hillshading.java
https://docs.locusmap.eu/doku.php?id=manual:faq:how_to_add_map_shading#where_to_get_them
hgt-source: https://bit.ly/dtm-germany-1s

### Subtasks

- [x] Make it work
- [x] Integrate with SAF
- [ ] Add download manager (either like Routing or Maps)

moving-bits:
There are some samples for Android apps under https://github.com/mapsforge/mapsforge/tree/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android, and you can contact the devs via their mailing list mapsforge-dev@googlegroups.com

ztNFny:
I've started implementing this in https://github.com/ztNFny/cgeo/tree/issue/12210_hillshading

The major obstacle so far is SAF. Mapsforge requires a `java.io.File` object for the folder containing the dem files but as far as I understand with SAF that's not possible.
As a workaround to get the actual functionality working I've added `android.permission.MANAGE_EXTERNAL_STORAGE`, but obviously that can't stay.

@moving-bits looking at PRs it seems you've implemented most SAF stuff in c:geo. Is it true that getting a File isn't possible or could you give me a pointer as to how?
If it isn't possible I guess we'd have to re-implement all classes in `org.mapsforge.map.layer.hills` (in mapsforge upstream of course)

moving-bits:
> @moving-bits looking at PRs it seems you've implemented most SAF stuff in c:geo.

99% of SAF was implemented by @eddiemuc actually.

He also added some PR to mapsforge to help them become SAF-compatible for theming. Maybe this would be the right way as well for hill shading?

> Is it true that getting a File isn't possible or could you give me a pointer as to how?

[...]

PR #13334, opened 2022-08-14
add support for hillshading on OSM

ztNFny:
Implements support for hillshading in OSM maps.
UnifiedMap is not covered by this PR.

DEM files have to be obtained manually. Good sources I know are https://sonny.4lima.de or http://viewfinderpanoramas.org/dem3.html - this should be added to FAQ before a release.

fixes https://github.com/cgeo/cgeo/issues/12210


ztNFny:
There's 3 Codacy errors I have no idea what to do about:
```
main/src/cgeo/geocaching/maps/mapsforge/v6/layers/HillShadingLayer.java#L15
The utility class name 'HillShadingLayer' doesn't match '[A-Z][a-zA-Z0-9]+(Utils?|Helper|Constants)'

main/src/cgeo/geocaching/maps/mapsforge/v6/layers/HillShadingLayer.java#L15
Utility classes should not have a public or default constructor.
```

Why is it marked a "utility class"? Naming matches the other classes in that package.

```
main/src/cgeo/geocaching/maps/mapsforge/v6/layers/RendererLayer.java#L4
Unused import - org.mapsforge.map.android.graphics.AndroidGraphicFactory.
```

The import is used though, as "AndroidGraphicFactory.INSTANCE"

fm-sys:
> ```
> main/src/cgeo/geocaching/maps/mapsforge/v6/layers/HillShadingLayer.java#L15
> Utility classes should not have a public or default constructor.
> ```
> 
> 
>     
>   
> 
> Why is it marked a "utility class"?

It's marked as an utility class cause all methods inside inside are static. Therefore, no instances need to be created and the automatically generated default constructor is useless.

Please add something like this to your code:

```
private HillShadingLayer() {
    // Do not instantiate, utility class
} 
```

For the naming, IMHO `HillShadingHelper` would work as well (especially as it's technically not returning a layer instance anyway). But we can also leave it as is.

>     Unused import - org.mapsforge.map.android.graphics.AndroidGraphicFactory.
>
> The import is used though, as "AndroidGraphicFactory.INSTANCE"

I don't see where `AndroidGraphicFactory.INSTANCE` should be used...?

ztNFny:
> I don't see where `AndroidGraphicFactory.INSTANCE` should be used...?

new MemoryCachingHgtReaderTileSource(shadingFolder, new SimpleShadingAlgorithm(Settings.getMapShadingLinearity(), Settings.getMapShadingScale()), **AndroidGraphicFactory.INSTANCE**);

ztNFny:
> This line is in `HillShadingLayer` while codacy is complaining about an unused import in `RendererLayer`
> ;-)

ah fu... 

Adding the private constructor didn't get rid of the "The utility class name 'HillShadingLayer' doesn't match '[A-Z][a-zA-Z0-9]+(Utils?|Helper|Constants)'" error?!

eddiemuc:
> > This line is in `HillShadingLayer` while codacy is complaining about an unused import in `RendererLayer`
> > ;-)
> 
> ah fu...
> 
> Adding the private constructor didn't get rid of the "The utility class name 'HillShadingLayer' doesn't match '[A-Z][a-zA-Z0-9]+(Utils?|Helper|Constants)'" error?!

Rename it to e.g. `HillShadingLayerHelper` or `HillShadingLayerUtils` will solve the error.

main/res/values/preference_keys.xml
```diff
@@ -190,6 +190,7 @@
     <string translatable="false" name="pref_persistablefolder_offlinemaps">persistablefolder_offlinemaps</string>
     <string translatable="false" name="pref_fakekey_info_offline_mapthemes">fakekey_info_offline_mapthemes</string>
     <string translatable="false" name="pref_persistablefolder_offlinemapthemes">persistablefolder_offlinemapthemes</string>
+    <string translatable="false" name="pref_persistablefolder_offlinemapshading">pref_persistablefolder_offlinemapshading</string>
```
moving-bits, line 193: Please do not use "pref_" prefix for the preference variable value (only for preference variable name).

main/res/values/preference_keys.xml
```diff
@@ -396,6 +397,9 @@
         <item>@string/pref_value_theme_dark</item>
     </string-array>
 
+    <string translatable="false" name="pref_mapShadingLinearity">pref_mapShadingLinearity</string>
+    <string translatable="false" name="pref_mapShadingScale">pref_mapShadingScale</string>
+
```
moving-bits, line 403: Same as above (do not use "pref_" prefix for preference variable value)

main/res/values/preference_keys.xml
```diff
@@ -396,6 +397,9 @@
         <item>@string/pref_value_theme_dark</item>
     </string-array>
 
+    <string translatable="false" name="pref_mapShadingLinearity">pref_mapShadingLinearity</string>
```
moving-bits, line 400: Same as above (do not use "pref_" prefix for preference variable value)

main/res/xml/preferences_map.xml
```diff
@@ -70,6 +70,29 @@
             android:summary="%s"
             android:title="@string/settings_title_map_rotation"
             app:iconSpaceReserved="false" />
+        <Preference
```
moving-bits, line 73: Maybe shading could get its own preference category, putting some of the text in "init_mapshading_folder_text" into the category's description?

main/src/cgeo/geocaching/storage/PersistableFolder.java
```diff
@@ -40,6 +40,8 @@ public enum PersistableFolder {
      */
     //legacy setting: "renderthemepath", a pure file path is stored
     OFFLINE_MAP_THEMES(R.string.pref_persistablefolder_offlinemapthemes, R.string.persistablefolder_offline_maps_themes, Folder.fromPersistableFolder(OFFLINE_MAPS, "_themes")),
+    /** Offline Maps: optional folder for map shading files (dem) */
+    OFFLINE_MAP_SHADING(R.string.pref_persistablefolder_offlinemapshading, R.string.pref_persistablefolder_offlinemapshading, Folder.fromPersistableFolder(OFFLINE_MAPS, "_hgt")),
```
moving-bits, line 44: second parameter should be the preference's title, not it's preference key name - current implementation leads to a strange dialog content (title and text) when setting the folder

moving-bits (changes requested):
Thanks for diving into hillshading and SAF. Implementation seems to work generally, although I got some NPE in individual routes - not sure if it's related to this PR, though.
Some texts may need some fine-tuning for more clarity, but this can be done later.
Otherwise I have some minor findings, see review comments.

main/res/xml/preferences_map.xml
ztNFny: Not sure about this one. Map themes, which is kinda similar, doesn't have its own section either.

I'm also not sure whether pref_mapShadingLinearity and pref_mapShadingScale really need to be user-configurable settings. I've tried to find reasonable defaults for those.
Maybe we could get some feedback during beta whether the defaults are good enough and if they are remove those settings (from GUI at least).

ztNFny:
retest this please

main/res/xml/preferences_map.xml
moving-bits: Yes, could be a way to go.

ztNFny: Let's merge then - your Review is still blocking it somehow

fm-sys:
retest this please

moving-bits:
@mucek4 
It's the "Could not create ADB Bridge." error again, can you restart the executors, please?

main/res/xml/preferences_map.xml
moving-bits: > Let's merge then - your Review is still blocking it somehow

Open reviews or reviews requesting changes do not block the build process and do not lead to failing builds.
It's an error in our CI environment, sometimes the server fails with "Cannot create ADB bridge", so it cannot run the tests - which leads to a failed / red build.

Let's wait for @mucek4 to fix this, and retest the PR then.

moving-bits (approved):
LGTM, thanks for fixing the issues.

moving-bits:
retest this please

PR #13339, opened 2022-08-15
disable the "Send" button in log activity while it's not yet available

### LINKED ISSUE #13337: show loading indicator while retrieving information for logs
### Tell us your idea!

C:geo retrieves information about available log types and TBs in inventory when opening the log screen. During this time a log can't be sent and an error dialog is displayed.

Instead of having to guess when the losing is done and repeatedly pressing the send button to get that error dialog a "loading" indicator should be shown and the send button disabled during this time.

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

ztNFny:
disable the "Send" button in log activity while it's not yet available (loading TBs and log types), enable it once that is loaded.

partially fixes #13337 - there's no progress shown yet during the operation to indicate to the user that something is happening.
I noticed showProgress(true) in the code of that class which sounds quite fitting, but not sure what it's supposed to do. Legacy stuff, broken?

### DIFF
--- a/main/res/menu/abstract_logging_activity.xml
+++ b/main/res/menu/abstract_logging_activity.xml
@@ -4,10 +4,10 @@
 
     <item
         android:id="@+id/menu_send"
-        android:enabled="true"
+        android:enabled="false"
         android:icon="@drawable/ic_menu_send"
         android:title="@string/send"
-        app:showAsAction="ifRoom|withText"> <!-- enabled=true, we show a message if logging is not possible -->
+        app:showAsAction="ifRoom|withText">
     </item>
     <item
         android:id="@+id/menu_templates"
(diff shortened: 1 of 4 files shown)


fm-sys:
> I noticed showProgress(true) in the code of that class which sounds quite fitting, but not sure what it's supposed to do. Legacy stuff, broken?

Not sure. But even if it would work, it was probably designed as blocking progress identicator. However in this case, the loading indicator should not block the whole activity...

Adding a progress indicator is quite easy. Just add a `com.google.android.material.progressindicator.CircularProgressIndicator` view to the layout file and show/hide it at the same places where you disable/enable the "Send" button

moving-bits:
Let's give it a try in the nightlies...

PR #13357, opened 2022-08-27
more subsections in map settings

ztNFny:
Map settings has a lot of stuff, add 2 new subsections to structure it better
image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -732,6 +732,8 @@
     <string name="settings_title_map">Map</string>
     <string name="settings_summary_map">Set map source, offline map settings and map details.</string>
     <string name="settings_title_map_data">Map Data</string>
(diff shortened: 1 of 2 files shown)


moving-bits:
I like the idea of structuring the many options a bit more. "Offline Maps" sounds useful, for the second subsection I'm unsure whether we actually need to include "hillshading" into the heading - I guess I would stay with just "Offline Map Themes".

And, while we are at this, we should move the "Map Rotation" entry down to "Map Behavior" - it's not really related to theming.

ztNFny:
> And, while we are at this, we should move the "Map Rotation" entry down to "Map Behavior" - it's not really related to theming.

Already part of this PR 😉

moving-bits:
> Already part of this PR 😉

Oops, sorry - had only compared your screenshots with current installation on my mobile

Lineflyer (approved):

PR #13358, opened 2022-08-27
#13337 show loading indicator in log screens

ztNFny:
image

### DIFF
--- a/main/src/cgeo/geocaching/activity/ActivityMixin.java
+++ b/main/src/cgeo/geocaching/activity/ActivityMixin.java
@@ -11,7 +11,6 @@ import android.content.Intent;
 import android.os.Build;
 import android.view.ContextThemeWrapper;
 import android.view.Gravity;
-import android.view.View;
 import android.view.Window;
 import android.view.WindowManager;
 import android.widget.EditText;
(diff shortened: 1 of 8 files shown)


main/src/cgeo/geocaching/activity/ActivityMixin.java
```diff
@@ -62,8 +63,10 @@ public static void showProgress(final AppCompatActivity activity, final boolean
             return;
         }
 
-        activity.setSupportProgressBarIndeterminateVisibility(show);
-
+        View progressBar = activity.findViewById(R.id.progress_bar);
```
fm-sys, line 66: I would prefer to not create a utility method here. Showing/hiding a view is a simple one-liner so a utility method is too overwhelming IMHO.

fm-sys:
Oh, and could you remove the margin above the progress indicator to be consistent with the rest of cgeo? 

Screenshot_20220828_020036

main/src/cgeo/geocaching/activity/ActivityMixin.java
ztNFny: didn't create it but modified the existing (broken) utility method. What should happen to that one? Leaving it broken is no good, so either update or remove

PR #13379, opened 2022-09-01
add "DAYOFWEEK" as log variable

### LINKED ISSUE #13373: Have log placeholder for "day of the week"
### Tell us your idea!

Feature request from support mail:
Hi, thanks for making geocaching so much easier with this app. I was wondering if there is a possibility to add a new command in the signature logs. So far I love the fact the variables like date, time, owner name, gccode and gcname etc etc are implemented by the placeholders. I was wondering if there could be one more: day of the week.

So when I make a signature log like:
On this beautiful [DAYWEEK] we found this gorgeous location.

That it shows [DAYWEEK] as the according day: Monday, Tuesday, Wednesday etc.
Of course [DAYWEEK] is just an example, leaving it all up to you awesome people to find the most appropriate command.

### Is there an alternative?

_No response_

### Additional Information

_No response_

ztNFny:
fixes #13373

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -824,6 +824,7 @@
     <string name="init_signature_template_button">Insert Placeholder</string>
     <string name="init_signature_template_date">Date</string>
     <string name="init_signature_template_time">Time</string>
+    <string name="init_signature_template_day_of_week">Day of week</string>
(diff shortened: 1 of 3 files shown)


fm-sys (approved):
LGTM

PR #13409, opened 2022-09-10
delete orphaned UDC on c:geo restart (fix #13352)

### LINKED ISSUE #13352: User defined cache cannot be removed
### Describe your problem!

Click on the user defined cache on the map and click the "trashbin icon", the toast prompts that the user defined cache has been removed from the stored list. But the user defined cache hexagon icon is still on the map. The hexagon icon still exists even after restarting the app and rebooting the phone.

### How to reproduce?

1. Long tap on the map and create a user defined cache
2. Delete the cache with the trashbin icon
3. The cache icon still exists on the map

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2022.08.21

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Adds method to delete orphaned UDC and call it on every restart of c:geo (as it has been before PR #13225, though accidentally)

### DIFF
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -3747,14 +3747,14 @@ public class DataStore {
         databaseCleaned = true;
 
         try (ContextLogger ignore = new ContextLogger(true, "DataStore.cleanIfNeeded: cleans DB")) {
-            // check for UDC cleanup every time this method is called
-            deleteOrphanedUDC();
+            Schedulers.io().scheduleDirect(() -> {
+                // check for UDC cleanup every time this method is called
+                deleteOrphanedUDC();
 
-            // other cleanup will be done once a day at max
-            if (Settings.dbNeedsCleanup()) {
-                Settings.setDbCleanupLastCheck(false);
+                // other cleanup will be done once a day at max
+                if (Settings.dbNeedsCleanup()) {
+                    Settings.setDbCleanupLastCheck(false);
 
-                Schedulers.io().scheduleDirect(() -> {
                     Log.d("Database clean: started");
                     try {
                         final Set<String> geocodes = new HashSet<>();
@@ -3784,10 +3784,9 @@ public class DataStore {
                     } catch (final Exception e) {
                         Log.w("DataStore.clean", e);
                     }
-
                     Log.d("Database clean: finished");
-                });
-            }
+                }
+            });
         }
     }
 
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -3747,6 +3747,10 @@ public class DataStore {
         databaseCleaned = true;
 
         try (ContextLogger ignore = new ContextLogger(true, "DataStore.cleanIfNeeded: cleans DB")) {
+            // check for UDC cleanup every time this method is called
+            deleteOrphanedUDC();
+
+            // other cleanup will be done once a day at max
             if (Settings.dbNeedsCleanup()) {
                 Settings.setDbCleanupLastCheck(false);
 
@@ -3830,6 +3834,25 @@ public class DataStore {
             database.delete(dbTableExtension, "_type NOT IN (" + type + ")", null);
         }
         database.delete(dbTableExtension, "_type=" + DBEXTENSION_INVALID.id, null);
+    }
+
+    private static void deleteOrphanedUDC() {
+        final Set<String> orphanedUDC = new HashSet<>();
+        queryToColl(dbTableCaches,
+                new String[]{"geocode"},
+                "SUBSTR(geocode,1," + InternalConnector.PREFIX.length() + ") = '" + InternalConnector.PREFIX + "' AND geocode NOT IN (SELECT geocode FROM " + dbTableCachesLists + " WHERE SUBSTR(geocode,1," + InternalConnector.PREFIX.length() + ") = '" + InternalConnector.PREFIX + "')",
+                null,
+                null,
+                null,
+                orphanedUDC,
+                GET_STRING_0
+        );
+        final StringBuilder info = new StringBuilder();
+        for (String geocode : orphanedUDC) {
+            info.append(" ").append(geocode);
+        }
+        Log.i("delete orphaned UDC" + info);
+        removeCaches(orphanedUDC, LoadFlags.REMOVE_ALL);
     }
 
     /**


fm-sys (commented):
Looks good in general. However, we need to check whether the SQL query is performant enough in big DB's. Some users have stored 100.000 caches or even more... Which brings me to the point that `cleanIfNeeded` should IMHO not be executed on main thread how it is currently done.

main/src/cgeo/geocaching/storage/DataStore.java
```diff
@@ -3832,6 +3836,25 @@ private static void deleteOrphanedRecords() {
         database.delete(dbTableExtension, "_type=" + DBEXTENSION_INVALID.id, null);
     }
 
+    private static void deleteOrphanedUDC() {
+        final Set<String> orphanedUDC = new HashSet<>();
+        queryToColl(dbTableCaches,
+                new String[]{"geocode"},
+                "SUBSTR(geocode,1," + InternalConnector.PREFIX.length() + ") = '" + InternalConnector.PREFIX + "' AND geocode NOT IN (SELECT geocode FROM " + dbTableCachesLists + " WHERE SUBSTR(geocode,1," + InternalConnector.PREFIX.length() + ") = '" + InternalConnector.PREFIX + "')",
```
eddiemuc, line 3843: Performance of query might improve if in the 'geocode not in' subquery you remove the 'where substr...' part completely. That way SQLite might be able to use an index on dbtablecacheslists.geocode. 

Also, depending on indexes and cleverness of SQLite, usage of LIKE with wildcard % at end might be more performant than SUBSTR. But that's debatable.

moving-bits: My thinking behind that condition was that typically you have very low UDC to all cache types ratio. If you leave out that condition, the database will do a cross-product of all caches against all cache to list relations, whereas with that condition it will only cross the much lower number of UDC against the cache/list relations.

Example: Let's have 20,000 caches, 100 of them UDC, and 25.000 list/cache relations.
Without the condition, the database will match 20,000 caches with 25,000 list/cache-relations, using a cross-product of 500 million entries, whereas with that condition, it will cross only 100 against 25,000, resulting in only 2.5 million entries to be tested.

Regarding the LIKE vs. SUBSTR I'm unsure which one is better. Both should be able to use the index, depending on the query optimizer. LIKE is typically optimized for strings starting with non-wildcard, on the other hand it's performing case conversion. (Not sure if that's the case for SQLite as well.)

Do we have any reasonably sized database to test this in real life?

fm-sys: Using SUBSTR might require an additional functionbased index on the table. However, there is no clear answer which of both is more performant. People are undecided here...

https://stackoverflow.com/questions/7434668/performance-of-sql-comparison-using-substring-vs-like-with-wildcard

moving-bits: I did a "EXPLAIN QUERY PLAN SELECT ..." and (at least with SQLite on the PC) both queries return the same execution plan (and both use the available indexes):

```
1) SCAN cg_caches USING COVERING INDEX in_caches_geo
2) LIST SUBQUERY 1
3) SCAN cg_caches_lists USING COVERING INDEX in_lists_geo
```

Leaving out the `WHERE` condition in the subquery I get a shorter query plan (as expected):

```
1) SCAN cg_caches USING COVERING INDEX in_caches_geo
2) USING INDEX in_lists_geo FOR IN-OPERATOR
```

Still would be nice if we could get some large real life database to run those queries against it.

What about including this PR in upcoming beta version and wait for possible reactions and/or Play console data?

fm-sys: One of our "power users" is maybe willing to share their DB with cgeo support via mail. Maybe @MagpieFourtyTwo, @DaTechniker or @gitilo? We also understand if you don't want to do this...totally fine as well. 

In case we want to just merge it: In this case I would like to ask the users mentioned above to report back whether cgeo's startup time feels significantly slower in the near future.

eddiemuc: I am with @fm-sys that the PR should be merged to see how the change behaves in the field.

Wrt creating large databases: this should be very easy by writing a small test helper routine which creates a large number of (fake) Geocache objects using the methods of DataStore. I did this already e.g. for the unit tests for the new Filter framework, check e.g. class GeocacheFilterTestUtils. I created only single geocaches there (and then test whether they are correctly SQL-filtered), but by using a loop this could also be used for creating large DBS 
(But please make sure such methods are not executes as part of the unit tests when checking in :-))

moving-bits:
> Looks good in general. [...] Which brings me to the point that `cleanIfNeeded` should IMHO not be executed on main thread how it is currently done.

Updated PR so that `deleteOrphanedUDC` gets called in background thread (as the main cleanup was already before)

main/src/cgeo/geocaching/storage/DataStore.java
MagpieFourtyTwo: > power users

Although daily in use, my DB is not exceptionally big (~24 k entries), and as I seldom delete entries (just adding e. g. recommendations, solved mysteries etc. and even found ones will just be moved to a diff. list), I expect a vacuum to have a rather "moderate" effect at best. Nevertheless, I will of course check performance before/after and if there is a significant change, you can of course get the "before DB" for deeper analysis.

MagpieFourtyTwo: Just tried ``Delete orphaned files`` with latest Nightly (2022.09.19-NB-fb9d87e) - "Maintenance" took 2:15 minutes and vacuumed the DB from 631 to 599 MB (size of sqlite backup before and after).

Not that much, one would think ... nevertheless, it feels *way* more performant now. And even objectively measured it has gained a lot: After first start it always took a few seconds(!) until it was "ready for action" (i. e. until it accepted a new command like entering a list or open the map), and now it's instantly responsive. Was not even able yet to get this "milky" list or map icon (sort of app freeze) I always got before, when I tapped the lists or map icon. And even while changing between lists, those (like e. g. one with ~7.5k caches or history) which took this kind of commemorative second before, are now instantly displayed.

IOW: Great stuff, BIG success!  ;)

fm-sys:
Then let's test it in real life...

DaTechniker:
@fm-sys If you contact ,me via my GC profile I can provide you my database.
It is around 77000 caches, in a lot of folders with version 2022.09.04

moving-bits:
> @fm-sys If you contact ,me via my GC profile and send me your mailaddress I can provide you my database. It is around 77000 caches, in a lot of folders with version 2022.09.04

@DaTechniker 
Thanks for your offer. You could use our ticket system to upload a file (using https://cgeosupport.droescher.eu/_upload), and then send an e-mail to support@cgeo.org with the link provided to your after uploading the file. For testing purposes a backup of your cgeo.sqlite file would be sufficient, no other files & folder needed. Thanks.

main/src/cgeo/geocaching/storage/DataStore.java
moving-bits: Thanks @MagpieFourtyTwo for your feedback. What you have observed is the result of #13423, the addition to reindex and compress the database. 😃 
This PR here is about immediately removing deleted UDC from database. It got merged today, so you'll see it in the next nightly.

DaTechniker:
Will try to do this tomorrow evening. The database is around 1,54GB and I need to transfer it to a PC before uploading.

main/src/cgeo/geocaching/storage/DataStore.java
MagpieFourtyTwo: Yep, you're of course right - mixed issues up. Nevertheless, feels incredibly good. ;)

fm-sys:
> I uploaded the file and sent the data to [support@cgeo.org](mailto:support@cgeo.org)

Big thanks! Hopefully, this helps us investigating how we can improve the performance in case of large databases...

moving-bits:
@DaTechniker 
Not sure if you have received my replies through our support system. I'm still in need of a token to be able to open your file. Can you send it to our support system? Thanks.

DaTechniker:
@moving-bits 
No I did not get any message at all. 
I sent the password via support mail.

moving-bits:
Thanks @DaTechniker for the reply. Looks like we are having problems with our mail server for the ticket system, so that outbound emails are not being sent. Inbound does work, though, so I have received your message. Alas, the password (or what I can see from it) does not seem to work for me. I see two words for the password, the first starting with "F", the second ending with "s". I've tried both using two words as well as only the second, but 7-zip always gives me a "wrong password" :-(

DaTechniker:
@moving-bits 
Password for 7zip begins with < and ends with 5 and has 20 chars. I used AES256 format for compressing
I sent the mail today 13:41

moving-bits:
Thanks, I've an idea what's going on here. I'll try to reach you via a different channel now.

DaTechniker:
info sent a view minutes ago.

moving-bits:
Perfect, this time it has worked and I was able to extract the file. Thanks for your patience.

DaTechniker:
OK, no problem from my side.
I am glad someone supports the fine software and that I can help at least with this.

moving-bits:
Using the database provided by @DaTechniker (app. 78k caches, 1.5 GB size) deleting orphaned UDC took between 0.15 and 0.25 seconds on an unmaintained database, and after having done a reindex, it took 0.025 seconds. Looks fine to me so far.

Issue #11429, opened 2021-08-07
c:geo not reacting on large lists


DaTechniker:
### Describe your problem!

When opening long lists (a few 1000 caches) I always get the message c:geo not reacting
If I click wait, then the list opens

### How to reproduce?

Open list with many caches

### Actual result after these steps?

c:geo not reacting message of android

![Screenshot_20210807-083631_cgeo](https://user-images.githubusercontent.com/88573756/128592036-f7cd0257-5650-4fbe-8bea-a7fa4edfeb0c.jpg)

### Expected result after these steps?

No android message

### Reproducible

Yes

### c:geo Version

2021.07.30-RC

### System information

```shell
## System information

c:geo version: 2021.07.30-RC

Device:
-------
- Device type: SM-A705FN (a70qeea, samsung)
- Available processors: 8
- Android version: 11
- Android build: RP1A.200720.012.A705FNXXU5DUE2
- Screen resolution: 1080x2198px (411x837dp)
- Sailfish OS detected: false
- Google Play services: disabled - 21.26.21 (150400-387928701)
- HW acceleration: enabled (default state)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: orientation

Program settings:
-------
- Filters: 
    LIVE: Aktiv und Liste und Gespeichert und Typ eingeblendet* ([Aktiv und Liste und Gespeichert und Typ eingeblendet*:inconclusive=false:advanced=true]AND(type;stored_since:-:-;status:found_no:exclude_archived;stored_list))
    OFFLINE: Aktiv und Liste und Gespeichert und Typ eingeblendet* ([Aktiv und Liste und Gespeichert und Typ eingeblendet*:inconclusive=false:advanced=true]AND(type;stored_since:-:-;status;stored_list))
- Hide waypoints: visited
- Set language: de_AT (system default)
- System date format: dd.MM.yy
- Debug mode active: no
- Live map mode: false
- OSM multi-threading: false / threads: 1
- Last backup: 2. Apr., 18:50
- Routing mode: Walk
- Settings: v8, Count:202
- Map: Aut+ (FZK) (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:0000-0000:OfflineMaps/Aut+ (FZK).map
[...]

PR #13423, opened 2022-09-16
[Performance] possibility to vacuum and reindex db

fm-sys:
rel. to #11429

- add `VACUUM` and `REINDEX` to our manual maintenance routine
- add blocking progress dialog while maintenance is running

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1051,6 +1051,7 @@
     <string name="init_maintenance_directories_note">c:geo stores images and other files related to a cache in a dedicated folder. In some cases (e.g. when importing/exporting the database) this folder may contain obsolete files that can be deleted here.</string>
     <string name="init_maintenance_directories">Delete orphaned files</string>
     <string name="init_maintenance_start">Maintenance started</string>
+    <string name="init_maintenance_ongoing">This may take some time…</string>
     <string name="init_maintenance_finished">Maintenance finished</string>
     <string name="init_location">Geolocation</string>
     <string name="init_location_note">On devices equipped with Google Play Services, c:geo can automatically use a better geolocation provider. However, this prevents the use of an external BlueTooth GPS receiver.</string>
--- a/main/src/cgeo/geocaching/settings/fragments/PreferenceOfflinedataFragment.java
+++ b/main/src/cgeo/geocaching/settings/fragments/PreferenceOfflinedataFragment.java
@@ -28,12 +28,19 @@ public class PreferenceOfflinedataFragment extends BasePreferenceFragment {
         findPreference(getString(R.string.pref_fakekey_preference_maintenance_directories)).setOnPreferenceClickListener(preference -> {
             // disable the button, as the cleanup runs in background and should not be invoked a second time
             preference.setEnabled(false);
-            ActivityMixin.showShortToast(getActivity(), R.string.init_maintenance_start);
+
+            final ProgressDialog waitDialog = new ProgressDialog(getActivity());
+            waitDialog.setTitle(getString(R.string.init_maintenance_start));
+            waitDialog.setMessage(getString(R.string.init_maintenance_ongoing));
+            waitDialog.setCancelable(false);
+            waitDialog.show();
+
             AndroidRxUtils.andThenOnUi(Schedulers.io(), DataStore::removeObsoleteGeocacheDataDirectories, () -> {
                 final Activity activity = getActivity();
                 if (activity != null) {
                     ActivityMixin.showShortToast(activity, R.string.init_maintenance_finished);
                 }
+                waitDialog.dismiss();
             });
             return true;
         });
--- a/main/src/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/cgeo/geocaching/storage/DataStore.java
@@ -1924,6 +1924,16 @@ public class DataStore {
                 }
             });
         }
+
+        // Try to reclaim unused space and reindex all tables
+        try {
+            Log.d("Database clean: vacuuming the freed space");
+            database.execSQL("VACUUM");
+            Log.d("Database clean: recreate indices");
+            database.execSQL("REINDEX");
+        } catch (final Exception e) {
+            Log.w("DataStore.clean", e);
+        }
     }
 
     public static boolean isThere(final String geocode, final String guid, final boolean checkTime) {
@@ -3868,7 +3878,7 @@ public class DataStore {
                                 }
                                 if (!found) {
                                     // update prefix in database
-                                    ContentValues values = new ContentValues();
+                                    final ContentValues values = new ContentValues();
                                     values.put("prefix", newPrefix);
                                     database.update(dbTableWaypoints, values, "_id=?", new String[]{String.valueOf(cursor2.getInt(0))});
                                     usedPrefixes.add(newPrefix);


fm-sys:
retest this please 

_CI build was failing, it's the "Could not create ADB Bridge" error again. This comment will trigger a rebuild._

fm-sys:
retest this please

fm-sys:
@mucek4 could you restart the CI?

fm-sys:
retest this please

MagpieFourtyTwo:
As already mentioned in #13409: Works like a charm!

Although my database has "only" vacuumed from 631 to 599 MB (i. e. just 5 %, which was to be expected, as I rarely delete stuff), I nevertheless am allowed to enjoy a *significant* performance increase since ``Delete Orphaned Files`` has been performed (which BTW took 2:15 minutes). Seems as if reindexing did the main job. At least for me (resp. my DB).

However: Thx a lot! This gave an extra kick to using c:geo.  ;)

moving-bits:
Using a couple of different databases I got the following test results:

| item | database 1 | database 2 | database 3 | database 4 |
|---|--:|--:|--:|--:|
| #caches | 2.5k | 5.1k | 21.8k | 77.2k |
| file size | 48M | 154M | 348M | 1.619M |
| vacuum 1 | 17.4s | 194.5s | 126.3s | --- (*) |
| reindex 1 | 0.15s | 1.5s | 2.71s | 31.2s |
| vacuum 2 | 17.3s | 196.5s | 128.4s | --- (*) |
| reindex 2 | 0.15s | 1.6s | 2.8s | --- (**) |

(*) system crashes on `VACUUM` with this database
(**) only one test run for `REINDEX` with that database

Test scenario was:
- put unmodified database into a backup folder
- start c:geo
- restore from that folder
- do first test run (vacuum, reindex)
- do a fresh restore from that folder
- do second test run (vacuum, reindex)

Vacuum seems to be an extremely costly operation, although not scaling linear to file size or number of caches. 
For the super-large database provided by @DaTechniker I have not been able to finish a vacuum run on my emulated device, even though I had created a powerful one for theses tests (8 GB RAM in contrast to default 1.5 GB, 8 GB file storage in contrast to default 2 GB). It crashes with errors like this:
```
2022-09-24 19:47:25.545 9536-9571/cgeo.geocaching E/SQLiteLog: (7) failed to allocate 4360 bytes of memory
2022-09-24 19:47:25.545 9536-9571/cgeo.geocaching E/SQLiteLog: (7) failed to allocate 4360 bytes of memory
2022-09-24 19:47:25.545 9536-9571/cgeo.geocaching E/SQLiteLog: (7) statement aborts at 9: [INSERT INTO vacuum_db.'cg_logs' SELECT*FROM"main".'cg_logs'] out of memory
2022-09-24 19:47:25.586 9536-9571/cgeo.geocaching A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 9571 (RxCachedThreadS), pid 9536 (cgeo.geocaching)
2022-09-24 19:47:25.802 9536-9549/cgeo.geocaching E/libutils.threads: androidCreateRawThreadEtc failed (entry=0x7864bf7342e0, res=11, Success)
```

This implies that for the super large database a reindex never gets called. So I commented out the `VACUUM` to see if `REINDEX` would finish - which does in my single test run for this, taking app. 31s.

Proposal therefore would be to
- either reverse the order of `VACUUM` and `REINDEX`, (so that `REINDEX` always runs)
- put `VACUUM` and `REINDEX` in separate try/catch blocks (same)
- or ask the user which of those two operations they may want to run (so that the user can decide which operation to run).

SammysHP:
Could we just copy everything from the database into a second database, then swap and delete the files?

> or ask the user which of those two operations they may want to run (so that the user can decide which operation to run).

Most users wouldn't know what it does. ;)

fm-sys:
As vacuum doesn't seem to make a big difference (in terms of disc space and performance), we could IMHO think about dropping it completely (or move it into a separate maintenance action)

Reindex is quite fast and does improve performance a lot. We should IMHO run this automatically from time to time to be able to provide a smooth user experience.

MagpieFourtyTwo:
@fm-sys 
> As vacuum doesn't seem to make a big difference (in terms of disc space and performance),

Au contraire! ;)

Last week my DB gained about 2k caches (w/ complete update, i.e. ~15 logs each, plus plenty of images, since many Virtuals and ECs were included) - and most of them have now been deleted again.

And afterwards I performed ``Delete orphaned files`` ... in fact just out of curiosity, as I thought it would not make a big difference anyway, at least yet, with the three days grace period in mind, but: DB size before: 639 MB, after: 534 MB.
And apart from the saved 100 MB (which I would already call quite significant), c:geo has become faster again after that.

Funny enough, even map display (especially with ``Combined (Offline)`` and about 5 "involved" maps) seems to have become faster again ... really possible or just a false impression?

In any case, I would really love to see the possibility of vacuuming preserved.

moving-bits:
@MagpieFourtyTwo 
Which version of c:geo have you been using when you observed the described data?

MagpieFourtyTwo:
2022.10.09-NB-a0dd6a5, as the apk says, which I always save with the corresponding sqlite.
But right now I realize ... seems as if you have already removed vacuum ...?!
Ok ... kind of strange ... but interesting ...  ;)

Lineflyer:
>  afterwards I performed Delete orphaned files ... in fact just out of curiosity, as I thought it would not make a big difference anyway, at least yet, with the three days grace period in mind, but: DB size before: 639 MB, after: 534 MB.

Did I miss something.
Once you deleted those 2k caches, they are still in the grace period.
If you use "Delete orphaned files" those will be purged....that probably the size difference you noticed, isn't it?

MagpieFourtyTwo:
Wasn't sure if the caches, which are still "protected" by the grace period, would be immediately purged, too, but if so, yes, you're right.

At least "kind of".  ;)

Cause "problem" is, that DBs usually do *not* free up the space which was occupied by the deleted stuff.
Usually (and I guess, Android's sqlite does it the same way) they just mark this storage space as "overwritable".
Which is in fact, why a vacuum is necessary at all.

moving-bits:
> But right now I realize ... seems as if you have already removed vacuum ...?! Ok ... kind of strange ... but interesting ... ;)

Yes, that's why I asked... But interesting that the file size still shrinks.

> Cause "problem" is, that DBs usually do _not_ free up the space which was occupied by the deleted stuff. Usually (and I guess, Android's sqlite does it the same way) they just mark this storage space as "overwritable". Which is in fact, why a vacuum is necessary at all.

I can only guess here: Maybe SQLite is able to free unused space at the end of the physical file without doing a `VACUUM`, and need that only to free space which is buried somewhere in the middle of the file.

MagpieFourtyTwo:
> Maybe SQLite is able to free unused space at the end of the physical file without doing a VACUUM, and need that only to free space which is buried somewhere in the middle of the file.

Close, but not quite ... at least not for "normal" SQlite DBs ... and in order to verify this with cgeo.sqlite, I just manually VACUUMed my 534 MB backup file - and it only shrunk by just another 11 MB, which was way less than was to be expected, *if* the DB would not have done any "vacuuming" itself (I doubt that most of what has been deleted recently, all happened to be at the end of the DB file by accident).

Therefore my next step was to check the auto_vacuum pragma (which in fact determines how an SQlite DB handles free'd space) - and much to my surprise it is set to 1, i. e. "Full", which means (at least for "normal" SQlite DBs), the freelist pages are moved to the end of the DB and then the DB file gets truncated in order to give the free'd space back to the world ... *and this it does at every single transaction commit!*

Which is good in terms of file size, of course, but pretty bad in terms of performance (because add. data must be pushed around on every single commit). And *really* bad in terms of fragmentation - cause unlike the "real" VACUUM, auto_vacuum does *not* reorganize/defragment the DB. Which of course decreases performance even more, over time. Which in turn explains, why I experienced a significant performance boost after my very first ``[Delete Orphaned Files]`` which still contained a VACUUM. Cause a defragmented DB is of course faster than a heavily fragmented one ...

Bottom line, IMHO: Since c:geo's DB seems to be set to full auto_vacuum (which BTW has to be set during DB creation and AFAIK cannot be turned off again), it seems *even more important* to me, to perform a "real" VACUUM from time to time, in order to regain lost performance. Therefore, I would suggest to let VACUUM find it's way back into the ``[Delete Orphaned Files]`` function.

Moreover, I would suggest to think about changing the auto_vacuum pragma to 2 (= Incremental), which means auto_vacuum is only performed after a dedicated ``pragma incremental_vacuum`` statement (which maybe called e. g. once a day or so) - which may perhaps increase commit performance.

Disclaimer: All this of course w/o any guarantee, cause my knowledge is only based on the current standard implementation of SQlite - and although unlikely, it's nevertheless possible that Android's SQlite still does it's own stuff.  ;)

PR #13438, opened 2022-09-24
Add view for pending downloads

moving-bits:
## Description
Adds a view for pending downloads to the main menu. This view lists all pending downloads handled by the internal download manager (eg: maps, routing tiles) with current status info. Pending downloads can be cancelled.

image.image

In the long run this shall be extended with an automatic check for pending downloads at start of c:geo, offering this view optionally, if there are any stuck or broken downloads currently.
This will also replace our current debug view for pending downloads in the future.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -903,6 +903,7 @@
             android:label="@string/search_bookmark_list">
         </activity>
         <activity android:name=".downloader.DownloadConfirmationActivity" android:exported="false" />
+        <activity android:name=".downloader.PendingDownloadsActivity" android:exported="false" />
         <activity
             android:name="cgeo.geocaching.downloader.DownloadSelectorActivity"
             android:label="@string/download_title">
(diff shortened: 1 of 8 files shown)


moving-bits:
retest this please

moving-bits:
Failing tests for build are unrelated, but Codacy isn't giving me any result at all?!

fm-sys:
For some reasons my pending downloads list is incredible long but I don't know why. Probably downloads I mostly started a year ago or something like that. Is this intended?

Screenshot_20221010_205326_cgeo geocaching

moving-bits:
Probably downloads that failed, but never got cleared by c:geo.

That there are no detail information for most of the downloads seem to point out that the system's download manager doesn't even know any longer of this download, but only the c:geo-internal reference is shown. Shall we clear those stale downloads automatically?

Lineflyer:
> Shall we clear those stale downloads automatically?

At least we might probably consider to clear everything which came into this backlog before this new feature was introduced...

moving-bits:
> > Shall we clear those stale downloads automatically?
> 
> At least we might probably consider to clear everything which came into this backlog before this new feature was introduced...

I'll look into it. We can probably do so without migration by simply checking against the date value of the specific entry. If this is older than October 10th, 2022, and status of download is failed, then we can delete the entry silently.

PR #13439, opened 2022-09-25
REINDEX in regular intervals, remove VACUUM (rel. to #13423)

moving-bits:
- Call REINDEX every 90 days + when requesting maintenance manually
- Remove VACUUM


moving-bits:
Failing test is unrelated and completes successfully in local test environment

PR #13440, opened 2022-09-25
check pending downloads on init

moving-bits:
## Description
Checks on c:geo startup whether there are any pending downloads not currently running (ie: being paused or failed). Prompts the user to open the "pending downloads view" if such downloads have been found. Within that view the user can cancel downloads.

image.image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -725,6 +725,8 @@
     <string name="downloader_cancel_file">Cancel pending download of "%s" and remove downloaded parts?</string>
     <string name="downloader_cancelled_download">Cancelled download for "%s"</string>
     <string name="downloader_pending_info">There is at least one pausing or failed download. Do you want to open the list of pending downloads?</string>
+    <string name="downloader_retry_download">Retry download</string>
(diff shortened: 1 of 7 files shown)


fm-sys:
As a user I might want to restart the download instead of deleting it... Can we add this action as well?

moving-bits:
Sure. Only for failed downloads, or would you want to see this for paused downloads as well?
I don't know if there's a way to force a resume, so the action would be to cancel the existing download and create a new one.

fm-sys:
> Only for failed downloads, or would you want to see this for paused downloads as well?

Good question, not sure... In which cases a download will be paused?

moving-bits:
> > Only for failed downloads, or would you want to see this for paused downloads as well?
> 
> Good question, not sure... In which cases a download will be paused?

Missing network connectivity is the most prominent. For more details see here: https://github.com/cgeo/cgeo/blob/9fdd51c37421194cfabe152d6c9c9a09aaa63df2/main/src/cgeo/geocaching/downloader/PendingDownloadsActivity.java#L95-L125

I'd say let's add the "resume" button for failed downloads only, "cancel" for all types (running, paused, failed).

Issue #2724, opened 2013-05-13
GPX eport file naming


Lineflyer:
From support mail:

Currently the GPX export does only assing a file name with a time stamp.
If you use to export quite often it will not be easy to get an overview about the files.
I suggest to include the gccode or the list name into the export name:
- Export of a single cache -> Include its GC-code
- Export of multiple caches from a list -> Include the list name

SchietSchijf:
Hi,

I don't know what the proper way is for voting on a feature.
I would like to be able choose the filename for the export of a complete list.

I'm happy with the fact that exporting a single cache results in a file with mask 'GC code.GPX'.

For exporting lists it would be helpful to be able to define a default file mask.
mask options could be: 
- date time
- list name
- a generated number to avoid file name conflicts
- ...

In the most ideal case, the application settings would contain a section where I can specify the file mask and in the export dialog box I would be able to change the suggested result of the file name.

Example use case:
My wife saves geocaches that are part of a tour in a list while I'm off to work. She exports the list and shares it on our private cloud so that it is automatically downloaded to my smart phone.
We both have set the GPX import folder to the synchronized folder.

As the cloud has file versioning, the export file can be overwritten and having the name of the list in the file name makes it easy to identify which list I have to create before importing the GPX file.

moving-bits:
A simple and more generalized solution to this could be to ask the user for the gpx file name on save, displaying the default name. This is the way we currently do it for exporting an individual route.

jelleruben:
I like to see this to. So you can easy find your export files back. Now you must know which export which list is.

moving-bits:
The initial request is resolved meanwhile and part of current nightly version:
GPX files get named after the cache name (for a single cache: geocode + cache name) and after the current list (for exporting a list).

PR #13443, opened 2022-09-25
Add cache/list name to GPX file name (rel. to #2724)

moving-bits:
## Description
Adds cache / list name to filename on GPX export:

image.image

This is a first step towards #2724 (actually fulfilling the original request). Next step would be to make the filename editable (as we have for individual routes).


moving-bits:
Failing test is unrelated, so let's test this feature update in the nightlies.

PR #13451, opened 2022-09-29
Fix: CITO - Cash in Trash Out Event

Juggler98:
## Description

Fix: CITO - Cash in Trash Out Event

### DIFF
--- a/main/res/values-sk/strings.xml
+++ b/main/res/values-sk/strings.xml
@@ -24,7 +24,7 @@
     <string name="mega">Mega event</string>
     <string name="giga">Giga event</string>
     <string name="earth">EarthCache</string>
-    <string name="cito">CITO - Cash in Trash Out Event</string>
+    <string name="cito">CITO - Cache In Trash Out Event</string>
     <string name="webcam">Webcam Cache</string>
     <string name="virtual">Virtuálka</string>
     <string name="wherigo">Wherigo keška</string>


moving-bits:
Hi @Juggler98,

thanks for your will to contribute. Alas, there are two issues with it:
The cache type _is_ actually called "cache in...", not "cash in...", see https://www.geocaching.com/help/index.php?pg=kb.chapter&id=23&pgid=799, so no need for a fix here;
and even if, we would need to fix this on Crowdin, as all our translations are managed there.

Thanks anyway, but will close here.

Juggler98:
Hi, but actually there is need to fix the translations. Check the commit. I changed it from Cash in to Cache In. I was looking at this issue for years in the app and nobody fixed it.

fm-sys:
Hi @Juggler98, thanks for trying to contribute to this project. 

For sure the current string "Cash" is wrong and needs to be corrected. (probably @moving-bits has mixed up the before and after string) 

However, our translations are managed via crowdin.

Please submit your change here:
https://crowdin.com/project/cgeo

Juggler98:
@fm-sys thank you.

moving-bits:
@Juggler98 
Oops, my bad! Sorry for me mixing up old & new string (I should not do those things on my mobile), and thanks for fixing the string.

Issue #13427, opened 2022-09-19
Next beta / feature release


moving-bits:
As discussed in some of the recent issues, it seems to be a good time to prepare a new beta version. I'll start working on it:

- [x] empty changelog_bugfix.md on branch master / sync changelogs
- [x] fast-forward release branch to the head of master branch
- [x] Check/Import incoming Translations from Crowdin
- [x] check/insert correct version name in /main/src/cgeo/geocaching/utils/BranchDetectionHelper.java (using format YYYY.MM.DD-RC)
- [x] Build release on CI server
- [x] Perform smoke test with resulting cgeo-RC.apk before continuing
- [x] Upload and publish cgeo-RC.apk to Google Play
- [x] Trigger notification to status.cgeo.org via CI server once the release is available on Google Play (not just being processed, but being available for download - sometimes takes up to one day)

moving-bits:
From what I can see on support current beta seems to be running smooth, so I guess we are closing in on a new release. Let's collect issues here which should be resolved before releasing:

- [ ] #13408
- [ ] #13434
- [ ] #13448
- [ ] #13452
- [ ] #13456

Feel free to edit list list.

fm-sys:
https://github.com/cgeo/cgeo/issues/13448 is related to the new image gallery, which is only used in the nightlies / with special hidden settings key turned on. Should the fix be targeted to `release` anyway?

fm-sys:
> #13448 is related to the new image gallery, which is only used in the nightlies / with special hidden settings key turned on. Should the fix be targeted to `release` anyway?

I've just retargeted it to `release` now (#13455). Should be an uncritical change anyway...

Lineflyer:
As I did not have the time to keep up with all developments lately, please allow one question:
Does the currently running beta include the feature, that the language is no longer switched?

moving-bits:
Will start preparing release now:

- [ ] Check and merge open Crowdin translation pull requests for release
- [ ] Check/Insert correct version name in BranchDetectionHelper
- [ ] Build release on CI server
- [ ] Perform smoke test with resulting cgeo-release.apk before continuing
- [ ] Upload and publish cgeo-release.apk to Google Play
- [ ] Tag the release on Github
- [ ] Publish the release on Github
- [ ] Publish release on F-Droid
- [ ] Trigger notification to status.cgeo.org via CI server once the release is available on Google Play (not just being processed, but being available for download - sometimes takes up to one day)
[...]

PR #13455, opened 2022-10-01
use image cache for log/spoiler image activity

### LINKED ISSUE #13448: [new image gallery] Listing/log mage not available when trying from listing/log
From field, can provide more details later:

Nightly 2022.09.26

Have bunches of caches stored offline (including images)
Device is offline
When trying to open the listing or log images from their individual icons on listing/logbook tab cgeo fails to show them but shows the "unavaible" placeholder image.

All listing log pictures however show fine from the gallery tab, meaning: They are stored and should be available via both methods.

Side note:
If I put the device online, clicking the link in listing/log will show loading indication, downloads the image and shows it only then.

Almost reproducible for me, however on rare caches it works normal

fm-sys:
fix #13448

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -2314,7 +2314,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             if (activity.imageGallery == null) {
                 final ImageGalleryView imageGallery = binding.getRoot().findViewById(R.id.image_gallery);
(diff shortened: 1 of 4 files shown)


fm-sys:
> https://github.com/cgeo/cgeo/issues/13448 is related to the new image gallery, which is only used in the nightlies / with special hidden settings key turned on. Should the fix be targeted to `release` anyway?

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/13427#issuecomment-1264590294_

PR #13471, opened 2022-10-05
Make details text line in compass scalable (fix #13420)

### LINKED ISSUE #13420: Check if compass view can be optimized to be scaleable
For users with a relatively high text size setting (and/or a smaller display resolution) the compass view starts getting cut off in some elements, e.g.:
grafik

The example above is from a user with following details:
```
Device:
-------
- Device type: SM-M215F (m21nseea, samsung)
- Available processors: 8
- Android version: 12
- Android build: SP1A.210812.016.M215FXXU2CVCC
- Screen resolution: 1080x2131px (411x811dp)
- Pixel density: 2.625
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 22.26.15 (190400-461192076)
```

It should be checked whether we can
- make the text rows width scalable as well 
- scale the compass rose image itself to avoid the lower text line being pushed out of the screen

moving-bits:
## Description
Make details text line in compass views (portrait & landscape) scalable, so that it automatically adjusts its height according to font size setting:

||portrait|landscape|
|---|---|---|
|default font size|image|image|
|large font size|image|image|

### DIFF
--- a/main/res/layout-land/compass_activity.xml
+++ b/main/res/layout-land/compass_activity.xml
@@ -63,7 +63,7 @@
          <RelativeLayout
             android:id="@+id/device_information"
             android:layout_width="fill_parent"
-            android:layout_height="16dip"
+            android:layout_height="wrap_content"
             android:layout_alignParentBottom="false"
             android:layout_alignParentLeft="false"
             android:layout_alignParentTop="false"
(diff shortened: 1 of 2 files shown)


Lineflyer:
Failing tests unrelated.
Thanks @moving-bits

Issue #13449, opened 2022-09-28
Status filter: filter dnf


Aslang13:
### Tell us your idea!

A way to see and catch up only with
My DNF's or unclutter the map by not showing them

### What solution would you suggest?

Adding the option in the already existing status filter

### Is there an alternative?

Not that I know of 

### Additional Information

Notyet

jakedot:
It's a good idea because the existing "found" setting doesn't provide for filtering out or in DNFs. If I set it to no, I still get DNFs mixed in the rest of unfound caches.

Somehow I think that there already was a setting for this in the Status filter?

At least I have a saved preset called "Found or DNF" with two instances of the status filter in it, one set to "Found=Yes" and the other to "has offline log = yes".

eddiemuc:
Fixed with PR #13468

PR #13472, opened 2022-10-05
Do not use abbreviation DNF

Lineflyer:
## Description

In #13468 a new filter for DNF was introduced. I think however the label should not use the abbreviation but "Did not find" instead.

## Related issues

#13468 
#13449


moving-bits (approved):
LGTM

jakedot:
Looking forward to next nightly, thanks for merging those PRs!

PR #13474, opened 2022-10-05
Fix #13473 - Better strings for download retry

### LINKED ISSUE #13473: Non working variable in string for Crowdin
This string
https://github.com/cgeo/cgeo/blob/c2d0e2814c191f10e2f91f479652c8f8c55d7e0d/main/res/values/strings.xml#L729

seems to produce problems on Crowdin as the variable is not shown, thus not allowing to place it correctly when translating.

image

Maybe a whitespace is needed after the line break?

@moving-bits  
Introduced in your PR with commit https://github.com/cgeo/cgeo/commit/c2d0e2814c191f10e2f91f479652c8f8c55d7e0d

Lineflyer:
## Description

Adapt the string to accommodate the new feature, that files to retry download can now be selected. 

## Related issues

#13473

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -727,7 +727,7 @@
     <string name="downloader_cancelled_download">Cancelled download for "%s"</string>
     <string name="downloader_pending_info">There is at least one pausing or failed download. Do you want to open the list of pending downloads?</string>
     <string name="downloader_retry_download">Retry download</string>
-    <string name="downloader_retry_download_details">Do you want to retry the download of</string>
+    <string name="downloader_retry_download_details">Please select the files you want to download again:</string>
 
     <!-- Android system download manager status messages -->
     <string name="asdm_status_failed">Download has failed (and will not be retried)</string>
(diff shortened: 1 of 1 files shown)


main/res/values/strings.xml
```diff
@@ -727,7 +727,7 @@
     <string name="downloader_cancelled_download">Cancelled download for "%s"</string>
     <string name="downloader_pending_info">There is at least one pausing or failed download. Do you want to open the list of pending downloads?</string>
     <string name="downloader_retry_download">Retry download</string>
-    <string name="downloader_retry_download_details">Do you want to retry the download of</string>
+    <string name="downloader_retry_download_details">Please select the files you want to download again:</string>
```
moving-bits, line 730: There will always be only a single file, as this dialog pops up as soon as you hit the "retry" button in the view.

(This differs from other usages of the same dialog, eg. after performing an update check or routing data coverage check, and offering one or multiple files to download. In that case, a different info text is displayed, though.)

Lineflyer: OK. Then probably "Do you want to retry the download of the following file?"

Issue #7243, opened 2019-01-18
Support event times on geocaching.com


rsudev:
In the meantime support for event times was added by geocaching.com (https://forums.geocaching.com/GC/index.php?/topic/351014-release-notes-website-event-start-and-end-times-january-15-2019/, example see https://www.geocaching.com/geocache/GC82BJT)

The code looks quite parseable, probably necessary to check how time and date settings affect this.

```
<div class="minorCacheDetails Clear">
    <div id="mcd3">
                    Start time:
                    11:30 AM
    </div>
    <div id="mcd4">
                    End time:
                    12:00 PM
    </div>
</div>
```

gsstark:
Would this solve the time zone problem that makes the add to calendar command unusable presently?

bekuno:
I think this values are in local time.

Maybe we can use the info below the cache logs:
e.g.: (from https://www.geocaching.com/geocache/GC8D064) 
Current Time: <time datetime="2019-09-07T00:50:49Z">09/07/2019 00:50:49 Pacific Daylight Time (07:50 GMT)</time>

ztNFny:
Only place I can find for the timezone is implicit in the javascript defining the date - this seems to be always GMT:

start: new Date(2019, 09-1, 15, 05, 30),
end: new Date(2019, 09-1, 15, 06, 00)

ReggieDoodle:
Just looking through ‘good first issue’ tickets …

This appears to be partially done.  The request was to support event times from GC. This works.

So is the remaining work ‘just’ to resolve the time zone issue when adding an event to the calendar and several issues relating to this are mentioned.   Maybe close this ticket and create a new one for the specific problem and avoid feature creep?

moving-bits:
c:geo is not timezone-aware currently, all date/timestamps are stored "as is", without any info about which timezone they belong to. To change this would be a major rework, I guess, as it would need to take all interfaces to external services into account, which may or may not support timezones.

ReggieDoodle:
Ha ha.

I'll park it for now then but maybe someone should edit the ticket title ?

Further exploring found than if you set the time zone when creating the calendar entry (manually of course) it shows your local and their local time in the calendar so if your new API can resolve the time zone at the event location you will indeed be better than the rest 😉

moving-bits:
Let's close here, as the original issue is implemented. We should open dedicated issues for further improvements we would like to see implemented.

PR #13479, opened 2022-10-06
get event end time for correct calendar entries

ztNFny:
GC listings for some while now contain the events start and end time in a standard format. This allows for reliable determination of both these dates.

By knowing the end date we are now able to create calendar entries with the proper end time.

I chose to parse short description instead of the listing (where the same information is available in 2 more places) because that is already stored in DB - the other 2 places would require an additional DB entry with little additional value.

image

### DIFF
--- a/main/src/cgeo/geocaching/models/Geocache.java
+++ b/main/src/cgeo/geocaching/models/Geocache.java
@@ -190,8 +190,7 @@ public class Geocache implements IWaypoint {
     private Boolean hasLogOffline = null;
     private OfflineLogEntry offlineLog = null;
 
-    private Integer eventStartTimeInMinutes = null;
-    private Integer eventEndTimeInMinutes = null;
+    private EventTimesInMin eventTimesInMin = new EventTimesInMin();
 
     private static final Pattern NUMBER_PATTERN = Pattern.compile("\\d+");
 
@@ -400,7 +399,7 @@ public class Geocache implements IWaypoint {
             searchContext = other.searchContext;
         }
 
-        this.eventStartTimeInMinutes = null; // will be recalculated if/when necessary
+        this.eventTimesInMin.reset(); // will be recalculated if/when necessary
         return isEqualTo(other);
     }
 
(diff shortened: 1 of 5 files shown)


Lineflyer:
Implements #7243? Great!

main/src/cgeo/geocaching/models/Geocache.java
```diff
@@ -398,7 +400,7 @@ public boolean gatherMissingFrom(final Geocache other) {
             searchContext = other.searchContext;
         }
 
-        this.eventTimeMinutes = null; // will be recalculated if/when necessary
+        this.eventStartTimeInMinutes = null; // will be recalculated if/when necessary
```
fm-sys, line 403: Should we reset the end time as well (here and at all other places where eventStartTimeInMinutes gets cleared)?

ztNFny: end time is only used together with start time. So while it might happen that only start time is reset before accessing end time both start and end time will be recalculated. So not necessary to clean it up.

ztNFny:
retest this please

main/src/cgeo/geocaching/models/Geocache.java
fm-sys: This is totally true as of now, however keep in mind that the code base might change over time. Parameters IMHO shouldn't magically depend on each other, either reset both or create a very small data class holding both values.

ztNFny:
failing build is unrelated

moving-bits:
retest this please

moving-bits:
@ztNFny 

Build is failing with compile error:

```
> Task :main:compileBasicDebugAndroidTestJavaWithJavac
22:03:38  /home/jenkins/slave/workspace/cgeo-CI_PR-build/tests/src-android/cgeo/geocaching/utils/EventTimeParserTest.java:87: error: cannot find symbol
22:03:38          assertThat(cache.getEventTimeMinutes()).isEqualTo(14 * 60 + 20);
22:03:38                          ^
22:03:38    symbol:   method getEventTimeMinutes()
22:03:38    location: variable cache of type Geocache
```

ztNFny:
Ah, the classes we discovered on Sundays call to be not marked as sources. Of course Android studio would miss them in a refactor.
Will change that call tomorrow.

ztNFny:
retest this please

PR #13482, opened 2022-10-07
for calendar entries if event shortdesc is very short use normal description

ztNFny:
Creating a calendar entry for an event adds the URL and shortdesc as the entries description.
While for OC shortdesc usually contains some information for GC it has nothing but the date. With this PR if shortdesc doesn't have any meaningful content (defined by being <150 chars) description is used for the calendar entry instead.

Before:

After:


ztNFny:
retest this please

moving-bits:
Failing tests are unrelated, so let's merge

PR #13491, opened 2022-10-11
Display images in waypoints

baiti:


baiti:
Of the proposed (@moving-bits) sequence:

1. add a field for an image to the Waypoint model (and extend the db structure to keep it persisted) - this would hold the image's URI
2. on parsing the stage data separate the image URI, store it to the waypoint data
3. and download the image using the caching HTMLImage class
4. extend the waypoint edit layout to include an ImageView and display the cached image in it

(1) and (2) are in now. I am still working on (3) and (4).

baiti:
@moving-bits : I tried to work on (3) and (4) mentioned above but no success. Could you outline in more details how to accomplish this? I know what XML file needs to be changed but I don't know how to code that `<ImageView>` tag. haven't found any meaningful samples on stackoverflow and also nothing in the cgeo code I could use as a boiler plate.

On the side: If the geocache description contains image tags, do we also need to cache those? Currently for ALCs, I don't. It seems the HTML renderer fetches the image from the network automatically, however, most likely each and every time the activity is started.

moving-bits:
LGTM so far (but haven't checked if you got all relevant methods in `DataStorage`.

> @moving-bits : I tried to work on (3) and (4) mentioned above but no success. Could you outline in more details how to accomplish this? I know what XML file needs to be changed but I don't know how to code that `<ImageView>` tag. haven't found any meaningful samples on stackoverflow and also nothing in the cgeo code I could use as a boiler plate.

I had to lookup that for myself. My first approach was to retrieve the image by using the image URL, store the image and put it into its own `<ImageView>`. But during a bit of research for some more detailed info I found a different approach, which might be easier:

Taken from `models/Geocache.java` (around line 2000):

```
final HtmlImage imgGetter = new HtmlImage(cache.getGeocode(), false, true, forceRedownload);
// store images from description
if (StringUtils.isNotBlank(cache.getDescription())) {
    HtmlCompat.fromHtml(cache.getDescription(), HtmlCompat.FROM_HTML_MODE_LEGACY, imgGetter, null);
}
```

This might work for both waypoint description and cache description itself.
Need to lookup though, how displaying the content makes use of the cache, though. Will try to find out tomorrow.
Extracting and storing the image URL in a separate database field would not be needed then.

> On the side: If the geocache description contains image tags, do we also need to cache those? Currently for ALCs, I don't. It seems the HTML renderer fetches the image from the network automatically, however, most likely each and every time the activity is started.

Exactly. Which can lead to problems if you have a bad signal (and it consumes traffic unnecessarily). HtmlImage should be able to cure this.

Lineflyer:
Before this all gets to complicated:
We could just store the images of the stages into the cache gallery, just giving them the title of the respective stage.
IMHO its not too relevant to really see the image embedded into the stage and creating a new DB extension just for that....

moving-bits:
@baiti: I added some code to cache the image and display it in a separate view (so back to my original proposal), which has the benefit of displaying it in full size instead of tiny thumbnail.

baiti:
> @baiti: I added some code to cache the image and display it in a separate view (so back to my original proposal), which has the benefit of displaying it in full size instead of tiny thumbnail.

So does that mean we're all set and done? Do I backout the second commit from this PR and we just merge the first commit?

baiti:
@moving-bits ... I have reverted my commit wrt addint a database field. So this PR only holds the change referenced in the title. However, it is out of sync since you commited changes to the same file. I have no idea to sync my local environment such that this PR is also rebased. I'm afraid I will screw things up by messing with git.

fm-sys:
Additional DB and Geocache fields, just for this small nearly irrelevant feature sounds a bit too much IMHO. The second proposal from https://github.com/cgeo/cgeo/pull/13491#issuecomment-1275305681  (inline via HtmlImage to TextView) should be quite easy to implement.

baiti:
> Additional DB and Geocache fields, just for this small nearly irrelevant feature sounds a bit too much IMHO. The second proposal from [#13491 (comment)](https://github.com/cgeo/cgeo/pull/13491#issuecomment-1275305681) (inline via HtmlImage to TextView) should be quite easy to implement.

Yes, that is what I already concluded from @Lineflyer's remarks. However, in an attempt to take out the database changes from this PR I also took out two commits from @moving-bits because I did not realize that he had commited into my PR (didn't know that this was even possible). I hope these two commits can be recovered somehow.

ztNFny:
Nothing is ever lost with git, they are still accessible from the commit id:


@moving-bits pushed 2 commits.

528e650 fix storing image uri
56e4979 display (cached) image in separate view

moving-bits:
> So does that mean we're all set and done? Do I backout the second commit from this PR and we just merge the first commit?

@baiti: No, I built it on top of your PR, and the functionality of your commits is needed.

> Additional DB and Geocache fields, just for this small nearly irrelevant feature sounds a bit too much IMHO. 

@fm-sys: What is your concern with having those extensions?

> The second proposal from [#13491 (comment)](https://github.com/cgeo/cgeo/pull/13491#issuecomment-1275305681) (inline via HtmlImage to TextView) should be quite easy to implement.

I've tried that first, but ran into the very same problem @baiti already had, the non-scaling of the image within the text field, which makes the image useless. That's why I came back to my first proposal, which was easy to add to what @baiti had prepared already, and works fine.

> in an attempt to take out the database changes from this PR I also took out two commits from @moving-bits because I did not realize that he had commited into my PR (didn't know that this was even possible). I hope these two commits can be recovered somehow.

@baiti: No worries, git still has them (as @ztNFny already pointed out), and I have them in my local dev environment as well.
If ok for you I can try to force-push the branch when I'm back at my dev environment.

baiti:
> > So does that mean we're all set and done? Do I backout the second commit from this PR and we just merge the first commit?
> 
> @baiti: No, I built it on top of your PR, and the functionality of your commits is needed.

Yup, that I figured out but only once it was too late.

> I've tried that first, but ran into the very same problem @baiti already had, the non-scaling of the image within the text field, which makes the image useless. That's why I came back to my first proposal, which was easy to add to what @baiti had prepared already, and works fine.

cool

> 
> > in an attempt to take out the database changes from this PR I also took out two commits from @moving-bits because I did not realize that he had commited into my PR (didn't know that this was even possible). I hope these two commits can be recovered somehow.
> 
> @baiti: No worries, git still has them (as @ztNFny already pointed out), and I have them in my local dev environment as well. If ok for you I can try to force-push the branch when I'm back at my dev environment.

Of course! Your help is very much appreciated. I have seen the content of those two commmits via the web ui but on my local working copy I can't see them no matter what I do.

Lineflyer:
> >Additional DB and Geocache fields, just for this small nearly irrelevant feature sounds a bit too much IMHO.
>
> @fm-sys: What is your concern with having those extensions?

Not sure about the concerns of @fm-sys, but did you see my proposal in https://github.com/cgeo/cgeo/pull/13491#issuecomment-1276708810

Wouldn't that be kind of consistent and not requiring new database extensions "just for AL stage pictures"?

moving-bits:
> > > I hope these two commits can be recovered somehow.
> > 
> > @baiti: No worries, git still has them (as @ztNFny already pointed out), and I have them in my local dev environment as well. If ok for you I can try to force-push the branch when I'm back at my dev environment.
> 
> Of course! Your help is very much appreciated. I have seen the content of those two commmits via the web ui but on my local working copy I can't see them no matter what I do.

@baiti: I have just force-pushed those commits.

moving-bits:
> > > Additional DB and Geocache fields, just for this small nearly irrelevant feature sounds a bit too much IMHO.
> > 
> > 
> > @fm-sys: What is your concern with having those extensions?
> 
> Not sure about the concerns of @fm-sys, but did you see my proposal in [#13491 (comment)](https://github.com/cgeo/cgeo/pull/13491#issuecomment-1276708810)
> 
> Wouldn't that be kind of consistent and not requiring new database extensions "just for AL stage pictures"?

Putting the images into the gallery may be a next step, but IMHO most usefulness of those images comes from having them displayed directly with their respective waypoint.

If we don't put them into a separate `ImageView` (requiring a separate database field for storing the URL) but leave them in the `TextView`, we do have the scaling problem described above (very tiny image which is non-legible).

An alternative approach, if we would like to spare a database field at all costs, could be to parse for image url in the `TextView` at runtime, strip the url from the `TextView` and put it into the separate `ImageView`. A bit more costly at runtime, but with the advantage of being able to convert already stored listings.

fm-sys:
> If we don't put them into a separate ImageView (requiring a separate database field for storing the URL) but leave them in the TextView, we do have the scaling problem described above (very tiny image which is non-legible)

In my opinion this should just be a matter coding it in the right way. At the listing this works as well, I can give it a try for waypoints if this is wanted.

> @fm-sys: What is your concern with having those extensions?

It IMHO adds unnecessary complexity to the DB and Geocache object. This would be totally acceptable of course, however we have a much more elegant solution here... As written above, I will give it a try and if I fail we can still go the current way.

> An alternative approach, if we would like to spare a database field at all costs, could be to parse for image url in the TextView at runtime, strip the url from the TextView and put it into the separate ImageView

Hm. This sounds quite ugly as well... Wouldn't recommend going this path.

moving-bits:
> At the listing this works as well, I can give it a try for waypoints if this is wanted.

Would be great if you can find a way to display the images within `TextView` in a reasonable size. Currently they are inserted by a plain `` tag, which leaves them at a very tiny size. If it works within the `TextView`, we just have to apply the `HTMLImage` for waypoints.

I'm not pushing the currently implemented solution, although I think it's straightforward, but a more elegant solution would be even better.

> > An alternative approach, if we would like to spare a database field at all costs, could be to parse for image url in the TextView at runtime, strip the url from the TextView and put it into the separate ImageView
> 
> Hm. This sounds quite ugly as well... Wouldn't recommend going this path.

Not that I would like to implement that alternative. It's more like a hack, and I've mentioned it just for the sake of completeness, as it would avoid adding the extra methods and db field.

fm-sys:
--> PR #13511 implements the same functionality as simple one-liner...

PR #13502, opened 2022-10-13
add missing pending intent flags (rel. to #13036)

moving-bits:
## Description
Add missing flags required when targeting Android 12 (targetSDK 31)

@fm-sys 
As I've copied from your code ;-)
Can you cross-check whether using `FLAG_IMMUTABLE` is the right thing in those three places? Thanks.

### DIFF
--- a/main/src/cgeo/geocaching/apps/cachelist/MapsMeCacheListApp.java
+++ b/main/src/cgeo/geocaching/apps/cachelist/MapsMeCacheListApp.java
@@ -10,6 +10,7 @@ import android.app.Activity;
 import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
+import android.os.Build;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


fm-sys:
Actually I haven't really understood the difference either. But if everything still works, I would say it is good to go...

Documentation:
> Flag indicating that the created PendingIntent should be immutable. This means that the additional intent argument passed to the send methods to fill in unpopulated properties of this intent will be ignored. 
It is strongly recommended to use FLAG_IMMUTABLE when creating a PendingIntent. FLAG_MUTABLE should only be used when some functionality relies on modifying the underlying intent, e.g. any PendingIntent that needs to be used with inline reply or bubbles.

-> So `FLAG_IMMUTABLE` seems to be the recommended one for our use case.

moving-bits:
Ok, thanks. Then let's give it a try in the nightlies...

PR #13503, opened 2022-10-13
fix lint: unexpected text in resource file

moving-bits:
## Description
Removes unexpected text in resource file


Lineflyer:
du....Thats really strange ;)

moving-bits:
This can happen when using a certain feature of Android Studio:
If you tap the left shift key twice in short succession, a search window will open where you can enter fragments of a file name (or class name etc.), a very handy navigation feature.
If you try to open this popup while AS is still syncing, any entered text will end up in the active file. And if that does not lead to compile errors, this mistake may keep unnoticed for quite some time...

PR #13511, opened 2022-10-13
Display images in waypoints (v2)

fm-sys:
closes #13491

Was more or less a one-liner to get image rendering working... Furthermore, I did some cleanup in `ALApi`.

grafik|grafik
-|-


moving-bits:
Way more elegant than my approach, thanks for tackling this!

PR #13517, opened 2022-10-14
allow manual setting of range slider values (fix #12808)

### LINKED ISSUE #12808: Improve slider input for filtering
### Tell us your idea!

From a Facebook user:

In the advanced filter for favorites, it can be reeeeeeally tricky to get the filter set to something like "10+ favorite points" exactly, because of my big fingers. Can it be either less sensitive to change when releasing the button, or changed to be text fields for the favorite points/percentage limits?

And on a similar note, make the settings be displayed near the bar, because I'm almost never 100% sure that I got it set to 10 instead of 9 or 11 or 13 etc. And it immediately disappears when you let go.

see above

### Is there an alternative?

No

### Additional Information

_No response_

moving-bits:
## Description
Tapping on either thumb of a range slider allows manual setting of its value.
Ranges of given slider are honored. Out-of-bound values are reset to slider's min/max values.

image

Setting this to draft for now, as it always uses the given range, not the "display version of given range", which makes a difference for a percentage-based range, where internal range is 0.0 to 1.0, whereas displayed range is 0 to 100 (percent).

@eddiemuc : Would it be sufficient to apply the `labelMapper` function to the internal min/max value?

### DIFF
--- a/main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
+++ b/main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
@@ -2,23 +2,31 @@ package cgeo.geocaching.ui;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.databinding.ContinuousrangesliderViewBinding;
+import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.utils.functions.Func1;
 
 import android.content.Context;
+import android.text.InputType;
 import android.util.AttributeSet;
 import android.view.ContextThemeWrapper;
 import android.view.ViewGroup;
 import android.widget.LinearLayout;
+import android.widget.Toast;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.util.Consumer;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
+import com.google.android.material.slider.RangeSlider;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 
 /**
  * View displays and maintains a slider where user can select single or range of items from a sorted item list.
+ * Tapping on a thumb allows manual value input
  */
 public class ContinuousRangeSlider extends LinearLayout {
 
@@ -29,6 +37,7 @@ public class ContinuousRangeSlider extends LinearLayout {
 
     private Func1<Float, String> labelMapper;
     private int axisLabelCount = 10;
+    private int factor = 1;
 
     private ContinuousrangesliderViewBinding binding;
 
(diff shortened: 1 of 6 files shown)


eddiemuc:
@moving-bits : great idea to use the listeners in this way to detect touches on the thumbs!

Wrt to your question:
> Would it be sufficient to apply the labelMapper function to the internal min/max value?

That would give you only one direction, but it would not be sufficient to map the user input back to the internal value. Note also that the labelMapper is meant for pure display purposes today, not for really validating input. For example, in the favorite point slider an internal value of 1000.02 is mapped to the string ">1000", and in the percentage use case the labelMapper will return things like "80%" (so including the percentage sign).

I fear you might have to introduce yet another mapper which can map internal values to user input and the other way around too. That one could e.g. in the favorite point case map all inputs larger than 1000 (e.g. 1540) to the internal value 1000.02.

Staying at this example, if you want to allow users to enter values >1000 which are then really honored (e.g. if user enters 1540, and filter should then really filter out everything with more than 1540 favs), then things will get more complex. Because the filter function will only have the internal value at her disposal, not all the gui-stuff like labels.

main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
```diff
@@ -64,6 +73,70 @@ private void init() {
...
+                }
+            }
+
+            @Override
+            public void onStopTrackingTouch(final @NonNull RangeSlider slider) {
+                final List<Float> values = slider.getValues();
+                if (slider.getFocusedThumbIndex() == lastThumb) { // on touch end it's the focussed thumb, though!
+                    // still on the same thumb, and not having moved it out of grace area?
+                    if (lastThumb >= 0 && lastThumb < values.size() && Math.abs(lastValue - values.get(lastThumb)) < grace) {
+                        final String title = String.format(getContext().getString(R.string.number_input_title), String.format(Locale.getDefault(), "%.2f", slider.getValueFrom()), String.format(Locale.getDefault(), "%.2f", slider.getValueTo()));
```
eddiemuc, line 101: This should be sparable if you use the TextParam.id - variant in the SimpleDialog call later. This method allows you to pass a resource id (R.string.number_input_title) followed by parameter values.

main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
```diff
@@ -64,6 +73,70 @@ private void init() {
...
+                                        slider.setValues(newValue, values.get(0));
+                                    } else {
+                                        slider.setValues(values.get(0), newValue);
+                                    }
+                                }
+                            } catch (NumberFormatException e) {
+                                Toast.makeText(getContext(), R.string.number_input_err_format, Toast.LENGTH_SHORT).show();
+                            }
+                        };
+                        SimpleDialog.of((Activity) getContext()).setTitle(TextParam.text(title)).input(inputType, defaultValue, null, "", listener);
```
eddiemuc, line 135: You might want to use the SimpleDialog.ofContext method instead

moving-bits:
Thanks @eddiemuc for your reply & review. To be honest, I adapted the complete input dialog from the already existing (but old) `SeekbarPreference`. I will apply your feedback there as well. (Maybe `SeekbarPreference` can even be based on `ContinuousRangeSlider` at some point.)

Regarding the `LabelFormatter`: Yes, that thought came me overnight as well. Let's see if I can reuse more thoughts from `SeekbarPreference`.

main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
moving-bits: done

main/src/cgeo/geocaching/ui/ContinuousRangeSlider.java
moving-bits: done

moving-bits:
To keep things simple, I adapted `ContinuousRangeSlider` in the minimum possible way by adding a `displayFactor` variable to `setScale`, as all of our current usages of `ContinuousRangeSlider`
- have a linear factor of either 1 or 100
- are displaying (rounded) integer values only.

I added possibility to enter a sign character, though, to be able to enter negative values (if lower bound < 0).

If we want to create a full-blown replacement for the different `SeekbarPreference` variants, much more would need to be done (several mapping functions).

|favorite filter, percent variant|log counts filter, only positive numbers|
|---|---|
|image|image|

moving-bits:
retest this please

Lineflyer:
Why are there negative numbers and decimal digits in the caption above the text input?

moving-bits:
> Why are there negative numbers and decimal digits in the caption above the text input?

That's the allowed value range in that specific context.

fm-sys:
Wouldn't it be better to display the label mapping there, E.g. >1000 instead of 1000.2

Still not a perfect solution but maybe a bit better...?

moving-bits:
It should display only such characters that can be entered, IMHO. As our `ContinuousRangeSlider` currently supports only integer values, we could round it to the nearest integer. 
(If someone actually needs the extreme positions, e. g. "1000.2" to mark "> 1000", they can use the slider itself.)

fm-sys:
> we could round it to the nearest integer.

Sounds good as well 👍

moving-bits:
> > we could round it to the nearest integer.
> 
> Sounds good as well 👍

see #13531

Issue #13458, opened 2022-10-03
Build tests for trackables fail


moving-bits:
Since a couple of builds two tests constantly fail on our CI, probably due to our changed language handling:

`CgeoApplicationTest/testSearchTrackable/` and `CgeoApplicationTest/testSearchTrackableSpottedLogState/`

This issue is a reminder to us to investigate and fix this.

Issue #13228, opened 2022-06-30
Don't change website language on gc.com


Lineflyer:
### Tell us your idea!

We should consider to implement a language independent parsing of the website of geocaching.com.

AFAIR there have already been ideas in the past to base the parsing on language Independent elements and it would BTW also be a comfort for our users to be able to use the website in their native language.

### What solution would you suggest?

Change the parsing to base on elements not language dependent 

### Is there an alternative?

Make use of the official API (or base all whats possible on the webAPI?)

### Additional Information

@cgeo/team Please refer to c:geo team internal message board for further explanation

Link to #4555 , #167

Lineflyer:
@fm-sys 
I finally found some minutes to at least update to latest nightly and use that new switch not to change the language.
You (and others) reported, that there is a problem in such case in certain functions. Up to now I could however not see any problems.

Would you mind to summarize here, what parts/functions of c:geo might fail if the website is not in English according to your experience/tests?

fm-sys:
After merging cgeo/cgeo#13381 there will be only two language depended patterns left (we know of)

- Parsing of Travelbug Location
- Owner Name parsing for PM only caches viewed by basic members (we can drop this IMHO, as this information is never displayed anyway)

All other features are (already) working fine with different website languages.

IMHO these two regressions are so small that it would even be accessible acceptable to remove the language switch completely on master.

However, we should adapt our unit tests to test website parsing in at least two different languages... (maybe German and English?)

Lineflyer:
> Parsing of Travelbug Location

What exactly do you mean by that.
At this moment I do see (current master, w/o language switch) a problem for the TB status where it seems, that some random string of c:geo pops up instead:

![Screenshot_20220912-185715_cgeo](https://user-images.githubusercontent.com/949669/189713043-bfb2c664-16dd-438b-b956-4c7815bf3e6b.jpg)

Lineflyer:
After some days of using c:geo without language switch may I suggest (in order to speed up progress):

Lets negate the switch (= on upgrading to that version the switch to english will be off by default) and start a beta rather soon.
We know of some regressions (under clarification with "server side") which can be considered minor and beta users might provide valuable additional information about possible regressions we have overseen.

Other changes needed before beta:
- Clarify this stange TB parsing result in the screenshot above (better show nothing than such strange info)
- Remove the mentioned parsing of PM-only cache owner information for BM

fm-sys:
[...]

PR #13519, opened 2022-10-15
temporarily disable spotted user tb tests (rel. to #13458, #13228)

moving-bits:
## Description
Removing our website-language switching is incomplete as of now, causing two trackable-related tests to constantly fail.
This is not only annoying, but also distracting from actual build problems.

Therefore let's temporarily disable the parts of the two tests which lead to making the tests fail, to get green builds back.
We already do have an issue as reminder to put those tests back into active state: #13458.


moving-bits:
Good news: Build tests are back to green.

Bad news: There's an error in one of the translations, which leads to a new lint error, breaking the build:

```
res/values-lt/strings.xml:2718: Error: Replace "..." with ellipsis character (…, &#8230;) ? [TypographyEllipsis]
<string name="rtl_path_quality23">kokybė 2/3 (danga: žvyras, remontuotas, ...)</string>
```

I've fixed this on CrowdIn and started a resync.

PR #13530, opened 2022-10-16
Listing link and image handling

### LINKED ISSUE #9019: Improve handling of pics embedded in cache listing
Sometimes I stumble upon caches, where owners put the listing text into pictures or similar usage of images as main information source. 
Perfect example: GC7B8FC

Easy method to zoom these listing images

Using the example cache above the only solution I found was to swipe to image tab,scroll down completely and click on listing image to enlarge then in external app.

fm-sys:
- Use ShareUtils for listing link handling (⇾ will use Chrome WebView if the relevant setting was enabled by the user)
- Open ImageViewActivity when clicking on listing images (fix #9019) 
   * ImageViewActivity will only be used if the image isn't wrapped into an `<a href="..">...</a>` link tag
-  avoid image duplication in image gallery (fix cgeo#13088) 

### Example:
image_preview_compressed

### DIFF
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -162,6 +162,7 @@ import androidx.fragment.app.FragmentManager;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
@@ -1900,8 +1901,11 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                     @Override
                     public void onClick(final View textView) {
                         final String imageUrl = span.getSource();
+                        final Collection<Image> listingImages = cache.getNonStaticImages();
+                        CollectionUtils.filter(listingImages, i -> i.category == Image.ImageCategory.LISTING);
+
                         if (Settings.enableFeatureNewImageGallery()) {
-                            ImageViewActivity.openImageView(activity, cache.getGeocode(), Collections.singletonList(IterableUtils.find(cache.getImages(), i -> imageUrl.equals(i.getUrl()))), 0, null);
+                            ImageViewActivity.openImageView(activity, cache.getGeocode(), listingImages, IterableUtils.indexOf(listingImages, i -> imageUrl.equals(i.getUrl())), null);
                         }
                     }
 
--- a/main/src/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/cgeo/geocaching/CacheDetailActivity.java
@@ -126,10 +126,12 @@ import android.text.Spannable;
 import android.text.SpannableString;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
+import android.text.TextPaint;
 import android.text.method.LinkMovementMethod;
 import android.text.style.BackgroundColorSpan;
 import android.text.style.ClickableSpan;
 import android.text.style.ForegroundColorSpan;
+import android.text.style.ImageSpan;
 import android.text.style.StyleSpan;
 import android.text.style.URLSpan;
 import android.text.util.Linkify;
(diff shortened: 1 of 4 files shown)


moving-bits:
This is nice, thanks for diving into it!

One additional feature request - not sure if it's possible with reasonable effort, though: When you are in the image view that opens after tapping on a listing image, is there a way to get the other listing images by left-/right-swipe instead of having to close the image view and tapping on the next image?

Lineflyer:
Do we really want to show this "Click here for full image size" added to the listing (or was that only part of this listing coincidentally).

moving-bits:
> Do we really want to show this "Click here for full image size" added to the listing (or was that only part of this listing coincidentally).

It's actually part of the listing, see https://coord.info/GC3BG1K

fm-sys:
> When you are in the image view that opens after tapping on a listing image, is there a way to get the other listing images by left-/right-swipe instead of having to close the image view and tapping on the next image?

Yes, that should be possible without much effort. Not sure if the ordering will be correct though. I will have a look...

eddiemuc:
> > When you are in the image view that opens after tapping on a listing image, is there a way to get the other listing images by left-/right-swipe instead of having to close the image view and tapping on the next image?
> 
> Yes, that should be possible without much effort. Not sure if the ordering will be correct though. I will have a look... 

I want to kindly ask not to pursue this for the time being. I need to change then current behavior of the new gallery. Which btw is still not available to the Public. I think we should stop coding mainly for ourself and the nightly users but concentrate on what our real user base needs and sees.

I am beginning to think it was not good that I activated the new gallery for nightlies and internals. Maybe I should deactivate it again until it is stable?

fm-sys:
> Maybe I should deactivate it again until it is stable?

Why? The gallery works perfectly fine from my experience. I would even go that far to say that it is ready for a public release. Caches with thousands of images are really rare and I honestly haven't discovered a single one while being in the fields.

...although I agree that it would be better of course if thousands of images can be rendered as well. But that's rarely really needed. 

> I think we should stop coding mainly for ourself and the nightly users but concentrate on what our real user base needs and sees.

The functionality to view images in fullscreen was explicitly requested, and I really hope we can roll out the image gallery with all its side features soon.

> I want to kindly ask not to pursue this for the time being

And last but not least: How do I have to understand your comment? Is this PR ok to be merged or not?

Lineflyer:
I am undecided about the new image gallery knowing, that it has issues.

However @fm-sys is kind of right: I am now using it regularly since months and did not gave any issues in my caching use so far. So I guess the problematic caches are more rare and we should start using it with next beta.
The underlay can still be changed later on as long as we think the UI look is kind of stable/final we should put it into use.

eddiemuc:
> > Maybe I should deactivate it again until it is stable?
> 
> Why? The gallery works perfectly fine from my experience. I would even go that far to say that it is ready for a public release. Caches with thousands of images are really rare and I honestly haven't discovered a single one while being in the fields.

Try GC89AXV (on an average phone)

> And last but not least: How do I have to understand your comment? Is this PR in the current form ok to be merged or not?

My comment was refering solely to the new request to add left/right swiping possibility to the opened image, not to the rest of the PR. If this is not included in the PR then I am fine.

In the end it is a group decision whether to set the (in my eyes) flawy implementation. However, I expect that those in favor of releasing it will take over responsibility to maintain the feature in the field in case of user reports as well.

fm-sys:
>> Caches with thousands of images are really rare and I honestly haven't discovered a single one while being in the fields.
>
>Try GC89AXV (on an average phone)

Sure, these problems exist (and I was aware of these examples which were flying around in some threads). However, e.g. GC89AXV is over 400km away from my homezone and I would probably never have opened the cache except for testing purposes...

> My comment was refering solely to the new request to add left/right swiping possibility to the opened image, not to the rest of the PR. If this is not included in the PR then I am fine.

Meanwhile it is already included, however only enabled for nightly users. Ideally, I would like to ship all image gallery related features together as soon as we consider everything ready. But this can be discussed of course.

eddiemuc:
> > > Caches with thousands of images are really rare and I honestly haven't discovered a single one while being in the fields.
> > 
> > Try GC89AXV (on an average phone)
> 
> Sure, these problems exist (and I was aware of these examples which were flying around in some threads). However, e.g. GC89AXV is over 400km away from my homezone and I would probably never have opened the cache except for testing purposes...

That's actually what I mean when I say we have to stop developing just for ourself and look at our  user base: yes, by chance our dev and nightly usersmight not have such caches in their neighborhood. However, within the normal c:geo user group there WILL be such users. It is those users we develop the app for.

moving-bits:
Same here: I'm using it for quite some time, and it's working flawlessly for me. 
But I fully understand if you want to refactor necessary things first, before we make it the default, and even before we add more interconnections with it. Thus I withdraw my feature wish for the time being :-)

PR #13532, opened 2022-10-16
Revert "Use ViewModel for InstallWizardActivity (#13197)"

Lineflyer:
This reverts commit 60d5eb72297328a3f7d940cf081981572e2f039e.

## Description

This reverts 
```
60d5eb72297328a3f7d940cf081981572e2f039e is the first bad commit
commit 60d5eb72297328a3f7d940cf081981572e2f039e
Author: Tiger Oakes <contact@tigeroakes.com>
Date:   Thu Jun 30 13:10:33 2022 -0700

    Use ViewModel for InstallWizardActivity (#13197)
```

This commit is causing the installation wizard to "hang" on Android 6 devices.
I tested the revert and it fixes the problem.

NOTE!
I did only prepare this PR as assistance for @moving-bits currently preparing the release.
Ideally we should not revert but find the culprit of non working Android 6 within this commit. Therefore marking as DNM for now, but can for sure be used if immediate release is required.

My pure guess is, that there is some method/function contained which is only supported on API levels > Android 6.

@NotWoods 
As this was your commit, maybe you have an idea.

## Related issues

See also #13427


moving-bits:
@NotWoods 
Please see https://github.com/cgeo/cgeo/issues/13427#issuecomment-1280011087 and following for more details.

NotWoods:
I imagine something is getting out of sync, since some synchronous code was replaced with observers. I'm happy to revert and take another stab with additional tests in the future.

Issue #13138, opened 2022-06-12
New Image gallery uses too many resources


eddiemuc:
### Describe your problem!

I am not satisfied yet with the performance of the new image gallery and want to use this issue to track my work on this. Bad aspects are,:
* Gallery loads all images of cache (asynchronous), also those which can't be seen (only if one starts scrolling)
* transitions forth and esp back from detail view are unreliable. Especially if master activity was destroyed by system meanwhile.
* opening detail view will cause transactiontoolarge if  there are too many images in cache

### How to reproduce?

Open cache with many images. Example: GC89AXV

### Actual result after these steps?

See above. I consider this issue done if even on a low-performance device cache listings like GC89AXV can be opened without much delay, without crash and with the ability to scroll through all images in gallery view (both list and detail/zoom view)

### Expected result after these steps?

Steps to make this better:
* Go away from two-activity-solution, make detail view an overlay of list view in same activity instead
* use recyclerview, but make is scrollable inside viewpager2 fragment

This is not easy but doable

### Reproducible

Yes

### c:geo Version

All

eddiemuc:
Note to self: recreate Imagegallery completely using Fragments as described in GridToPager example

eddiemuc:
While working on this I had the impression that Html.fromHtml starts to break down when the description reaches a very large size. Example: GC89AXV (937000 chars). At least this happens in my emulator.

As a first step I was thinking about capping the description size at a certain, very high treshold, e.g. 100000 chars. What is the team thinking about this?

moving-bits:
(Another example would be GC6FJH3.)

Within a mobile app it's rarely useful to see such a long listing completely, and it really slows down the app, and hinders accessing the note. So limiting display to a default of even 5,000 or 10,000 characters should be enough in most cases.
A hint should be added, eg "display truncated to first 5,000 characters", with a link to display the full text (within the app), being loaded only on demand.

fm-sys:
> So limiting display to a default of even 5,000 or 10,000 characters should be enough in most cases.

10k characters can be quite short, e.g. for some longer Multis / mysteries or caches which contain Wikipedia abstracts. I would give at least 20k characters loaded by default. But I agree that more than that doesn't make much sense...

eddiemuc:
I found e.g. https://www.geocaching.com/geocache/gc8dvvq which has 38841 chars but is still ok-ish imho. So I would set the limit to 50k for now. hope that's ok. PR will follow

[...]

PR #13545, opened 2022-10-20
rel to #13138: make load of cache description asynchronous and limited

eddiemuc:
rel to #13138: make load of cache description asynchronous and limited

* Cache listings will now be rendered aynchronously. 
* Very long listings will now be shortened to first 100000 chars. A warning will be displayed that listing was cutted and option is given to re-render the full listing

image

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1362,6 +1362,9 @@
     <string name="cache_personal_note_upload_cancelled">Personal note upload cancelled</string>
     <string name="cache_description">Description</string>
     <string name="cache_description_table_note">Description contains table formatting which may need to be viewed at %s to be seen correctly.</string>
+    <string name="cache_description_rendering">Rendering description, please wait…</string>
+    <string name="cache_description_render_restricted_warning">Warning: Description is very large (%1$d chars). Only first %2$d%% were rendered.</string>
+    <string name="cache_description_render_fully">Render complete description</string>
     <string name="cache_area_description">Area Description</string>
     <string name="cache_cache_description">Cache Description</string>
     <string name="cache_virtual_description">Virtual Question</string>
--- a/main/res/layout/cachedetail_description_page.xml
+++ b/main/res/layout/cachedetail_description_page.xml
@@ -27,6 +27,14 @@
             tools:text="This is the cache description. It might be very long..."
             tools:visibility="visible" />
 
+        <Button
+            android:id="@+id/description_render_fully"
+            style="@style/button_full"
+            android:clickable="true"
+            android:focusable="true"
+            android:visibility="gone"
+            android:text="@string/cache_description_render_fully" />
+
         <RelativeLayout
             android:id="@+id/loading"
             android:layout_width="fill_parent"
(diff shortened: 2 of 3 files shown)


eddiemuc:
For this PR I had to dive deep into the implementation of how we render and display cache descriptions, because the "rendering" part (to be done in background thread) and the "display" part (which needs to be done in gui thread) were pretty entangled within each other. Various methods to "fix" certain render/display aspects and also error handling (e.g. unknown tags) had to be rearranged.

Because of this I would appreciate at least a short review to ensure I didn't forgot or accidentally delete/destroy some of the many pre- and post-processing steps for the description

fm-sys (commented):
At a quick glance, the code looks fine. However, haven't actually compiled and tested it...

Lineflyer:
I gave it some test in my emulator and could not see any difference to before...meaning it seems to work.

The problematic caches GC89AXV however still makes c:geo really unresponsive on my emulator. But this is probably due to the lot of images rather than lot of chars?!

```
2022-10-21 11:22:10.131 521-29342/? E/ActivityManager: ANR in cgeo.geocaching (cgeo.geocaching/.CacheDetailActivity)
    PID: 28260
    Reason: Input dispatching timed out (ac49929 cgeo.geocaching/cgeo.geocaching.CacheDetailActivity (server) is not responding. Waited 5001ms for FocusEvent(hasFocus=true))
    Parent: cgeo.geocaching/.CacheDetailActivity
    Load: 2.48 / 1.75 / 0.9
    ----- Output from /proc/pressure/memory -----
    some avg10=8.44 avg60=7.06 avg300=3.68 total=37039909
    full avg10=3.08 avg60=1.83 avg300=0.85 total=8002696
    ----- End output from /proc/pressure/memory -----
```

BTW:
Is your description of this PR correct?
You wrote the listing will be truncated if its above 100.000 chars...you probably mean less?

eddiemuc:
> I gave it some test in my emulator and could not see any difference to before...meaning it seems to work.

Thanks a lot for testing!

> The problematic caches GC89AXV however still makes c:geo really unresponsive on my emulator. But this is probably due to the lot of images rather than lot of chars?!

Correct. In my emulator, this will even bring it to a crash.
To be more specific: once this PR is merged, then the image gallery will be the ONLY thing left stopping c:geo to open GC89AXV gracefully... Obviously I will takkle this next.


> BTW: Is your description of this PR correct? You wrote the listing will be truncated if its above 100.000 chars...you probably mean less?

Actual implementation indeed truncates after 100000 chars. But you are right, we agreed on a value less than that. I suggested 50k = 50000 chars at the time (or 512000 if you want to calc in computer terms :-)). I will change it to 50000. It will not make much difference in practice though

eddiemuc:
Update: changed restrict limit to 50000 chars in PR

main/res/values/strings.xml
```diff
@@ -1362,6 +1362,9 @@
     <string name="cache_personal_note_upload_cancelled">Personal note upload cancelled</string>
     <string name="cache_description">Description</string>
     <string name="cache_description_table_note">Description contains table formatting which may need to be viewed at %s to be seen correctly.</string>
+    <string name="cache_description_rendering">Rendering Description, please wait…</string>
```
moving-bits, line 1365: "description" should start with a lower-case "d"

eddiemuc: done

PR #13546, opened 2022-10-20
add a preference category for hillshading, add explanatory text and link to wiki

ztNFny:
This ALSO addresses #13547 by replacing all android:text with android:summary in preference pages

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1004,7 +1004,7 @@
     <string name="settings_info_themes_title">Info on Map Themes</string>
     <string name="settings_info_themes">c:geo supports custom themes for offline maps. These can be used to change the color style of the map (e.g. to have a nightview map) or to highlight certain objects like cycle paths or height lines within the map.\n\nWorking with non-zipped map themes might decrease map viewer opening performance. In this case, the theme folder may optionally be synchronized to an app-private local folder for faster access (at the cost of some storage space).</string>
     <string name="settings_info_hillshading_title">Info on Hillshading</string>
-    <string name="settings_info_hillshading">Hillshading is a technique for creating topographic relief maps by adding a lighting effect to the map to indicate changes in elevation. Sloped will be darkened/lightened to generate a 3D-effect. The source files containing the elevation information have to be obtained manually - click this item for sources - and placed in the sources folder defined below.</string>
+    <string name="settings_info_hillshading">Hillshading is a technique for creating topographic relief maps by adding a lighting effect to the map to indicate changes in elevation. Slopes will be darkened/lightened to generate a 3D-effect.\nThe source files containing the elevation information have to be obtained manually - click this item for sources - and placed in the sources folder defined below.</string>
     <string name="init_mapsource_select">Select Map Source</string>
     <string name="init_local_file_system">Local File System</string>
     <string name="init_base_directory_description">Base folder for public local data</string>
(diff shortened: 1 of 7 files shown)


main/res/xml/preferences_map.xml
```diff
@@ -70,6 +70,15 @@
             android:summary="@string/init_summary_renderthemefolder_synctolocal"
             android:title="@string/init_renderthemefolder_synctolocal"
             app:iconSpaceReserved="false" />
+    </PreferenceCategory>
+    <PreferenceCategory
+        android:title="@string/settings_title_map_hillshading"
+        app:iconSpaceReserved="false">
+        <Preference
+            android:key="@string/pref_fakekey_info_offline_maphillshading"
+            android:text="@string/settings_info_hillshading"
```
moving-bits, line 79: use `android:summary` instead for the text to be displayed

moving-bits (changes requested):

main/res/xml/preferences_map.xml
ztNFny: I copied it from offline maps and themes info. Guess those are wrong as well then. Should i fix it in this PR?

moving-bits: I'm not sure. I've stumbled upon those as well after I'd commented on this PR, so I wonder whether there actually _should_ be an `android:text` attribute within preferences and whether we have only suppressed it somehow. Need to research this a bit first.

ztNFny: I opened https://github.com/cgeo/cgeo/issues/13547 for that when I noticed. Didn't have time to look into the cause yet

moving-bits: According to https://developer.android.com/develop/ui/views/components/settings/components-and-attributes there's no `text` attribute allowed / recognized within preferences. Maybe the old preferences lib tolerated it, but the AndroidX version does not seem to support it. (Also, during my very short research, I haven't found a single example using that attribute.)
So I guess the existing instances could be renamed to `android:summary`.

ztNFny: added a commit replacing all android:text with android:summary to this PR (3 dialogs)

ztNFny:
This PR now also addresses https://github.com/cgeo/cgeo/issues/13547 by replacing all android:text with android:summary in preference pages

main/res/values/strings.xml
```diff
@@ -1002,6 +1003,8 @@
...
     <string name="settings_info_themes_title">Info on Map Themes</string>
     <string name="settings_info_themes">c:geo supports custom themes for offline maps. These can be used to change the color style of the map (e.g. to have a nightview map) or to highlight certain objects like cycle paths or height lines within the map.\n\nWorking with non-zipped map themes might decrease map viewer opening performance. In this case, the theme folder may optionally be synchronized to an app-private local folder for faster access (at the cost of some storage space).</string>
+    <string name="settings_info_hillshading_title">Info on Hillshading</string>
+    <string name="settings_info_hillshading">Hillshading is a technique for creating topographic relief maps by adding a lighting effect to the map to indicate changes in elevation. Sloped will be darkened/lightened to generate a 3D-effect. The source files containing the elevation information have to be obtained manually - click this item for sources - and placed in the sources folder defined below.</string>
```
moving-bits, line 1007: Do you mean "sloped" or "slopes"?

moving-bits (approved):
LGTM, thanks for fixing the preferences' attributes as well. Just a tiny question about wording for the explanation for hillshading.

main/res/values/strings.xml
ztNFny: well spotted. fixed

PR #13549, opened 2022-10-20
Improve user-display-string for D/T Filter and other

### LINKED ISSUE #11409: D/T Filter and other: optimize user-display-string
We should optimize the front-end:
- `*-*` -> show nothing 
- `*-x` -> `<x`
- `x-*` -> `>x`
- `x-y` -> can stay like it is 

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11362#issuecomment-892225406_

murggel:
## Description

displayString-logic for min-max-filters

    *-* -> * (normally this will not be displayed)
    *-x -> <x
    x-* -> >x
    x-y -> x-y

## Related issues

fix #11409

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/OrGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/OrGeocacheFilter.java
@@ -40,6 +40,6 @@ public class OrGeocacheFilter extends LogicalGeocacheFilter {
 
     @Override
     public String getUserDisplayableType() {
-        return "^";
+        return " ^ ";
     }
 }
--- a/main/src/cgeo/geocaching/filters/core/OrGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/OrGeocacheFilter.java
@@ -40,6 +40,6 @@ public class OrGeocacheFilter extends LogicalGeocacheFilter {
 
     @Override
     public String getUserDisplayableType() {
-        return " ^ ";
+        return " OR ";
     }
 }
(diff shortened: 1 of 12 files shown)


main/src/cgeo/geocaching/filters/core/UserDisplayableStringUtils.java
```diff
@@ -0,0 +1,27 @@
+package cgeo.geocaching.filters.core;
+
+final class UserDisplayableStringUtils {
+
+    private UserDisplayableStringUtils() {
+        // Do not instantiate
+    }
+
+    static String getUserDisplayableConfig(final boolean minValueSet, final boolean maxValueSet, final String minValue, final String maxValue) {
```
eddiemuc, line 9: I believe you could leave out the parameters "minValueSet" and "maxValueSet" and instead make the parameters "minValue" and "maxValue" nullable. If "minValue" is null, then minValue is not set, same for maxValue. I believe this would make the code a bit easier, also on the caller side

eddiemuc (approved):
Really like that you separated out a separate UserDisplayableString Utils class and added Unit tests! I have one comment (see there), but it's not critical

main/src/cgeo/geocaching/filters/core/UserDisplayableStringUtils.java
murggel: good point

main/src/cgeo/geocaching/filters/core/OrGeocacheFilter.java
```diff
@@ -40,6 +40,6 @@ public void addToSql(final SqlBuilder sqlBuilder) {
 
     @Override
     public String getUserDisplayableType() {
-        return "^";
+        return " ^ ";
```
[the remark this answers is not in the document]
murggel: Since we have the unlocalized `NOT` for the NotFilter - I will go with the `OR`

PR #13550, opened 2022-10-21
Fix #13541 - Use better coloring for archived caches

### LINKED ISSUE #13541: Improve contrast for archived cache title
### Tell us your idea!

From support mail (summarized):
The archived caches show in C:Geo with a red title with strikethrough.
The problem is that this is hard to read when using Dark Theme, especially on a bright sunny day.

Could we either:
Harder solution: Use a more dynamic font colour - a lighter red when using dark theme, and a darker red when using light theme, or

Easier solution: On the cache details page, keep the title as red but change the "Name" field to the default colour (white for dark theme, black for light theme). 

### Is there an alternative?

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

- Apply different colors for name of archived caches for dark/light theme
- Remove coloring in cache details list (only show it in top bar of cache view and map popup)

This is draft and DNM, because its not applying the correct theme color for top bar and popup.
Its however working in list and in dropdown on home screen.

For better debugging I temporarily assigned 
GREEN = light theme
BLUE = dark theme

It looks like this now..dark theme is not applied everywhere:
| Dark | Light |
| --- | --- |
| image | image |
| image | image | 

## Related issues

#13541

### DIFF
--- a/main/res/raw/changelog_base.md
+++ b/main/res/raw/changelog_base.md
@@ -24,3 +24,4 @@ Due to upcoming restrictions in Play Store we have upgraded the targeted Android
 - New: Allow manual input of values in filters using sliders
 - New: Enable upload of modified coordinates for caches imported from GPX file, when cache has waypoint of type "ORIGINAL"
 - Change: Improve filter status line text
+- Change: User a better readable color for archived cache names in titles and remove coloring from cache details page
--- a/main/res/values-night/colors_dark.xml
+++ b/main/res/values-night/colors_dark.xml
@@ -6,6 +6,7 @@
     <color name="colorText">@color/just_white</color>
     <color name="colorTextDark">#AAFFFFFF</color>
     <color name="colorTextHint">#A0CCCCCC</color>
+    <color name="archived_cache_color">#3F51B5</color>
 
     <color name="colorTextHeadline">#88FFFFFF</color>
     <color name="colorTextHintActionBar">@color/colorTextHint</color>
--- a/main/res/values-night/colors_dark.xml
+++ b/main/res/values-night/colors_dark.xml
@@ -6,7 +6,7 @@
     <color name="colorText">@color/just_white</color>
     <color name="colorTextDark">#AAFFFFFF</color>
     <color name="colorTextHint">#A0CCCCCC</color>
-    <color name="archived_cache_color">#3F51B5</color>
+    <color name="archived_cache_color">#FFEA0000</color>
 
     <color name="colorTextHeadline">#88FFFFFF</color>
     <color name="colorTextHintActionBar">@color/colorTextHint</color>
--- a/main/res/values/colors.xml
+++ b/main/res/values/colors.xml
@@ -12,7 +12,6 @@
 
     <color name="just_white">#FFFFFFFF</color>
     <color name="just_black">#FF000000</color>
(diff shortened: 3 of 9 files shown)


Lineflyer:
Nope...its persistent. Even rebooted the emulator.

moving-bits (approved):
Hi @Lineflyer,

it got a bit complicated, so I added a new commit instead of just texting here. The trick was to make `TextUtils.coloredCacheText` context-aware, as a `Context` (which may be an `Activity`) holds the theming information. Before the change `coloredCacheText` used the application context, which does not know of the theme and uses the default color (as defined in `colors.xml` at that time) instead.
I also reverted your removal of calling `coloredCacheText` in `CacheDetailActivity` as it formats the cache title for other cache states as well and therefore needs to be kept in place.

Two other minor change are just for code cleanness:
- Colors which use different values in light/dark theme should be placed in `colors_light.xml` and `colors_dark.xml` respectively (instead of the general `colors.xml`). No technical necessity for this, just a c:geo convention to detect more easily that this color is themed.
- I added a `FF` opacity prefix to the color. Seems to be the default, but as we have all of our colors as eight-character values (including the opacity prefix) it's easier to compare to other colors.

I left the color values at blue/green for now for easier testing. Should be converted to the desired colors before merging.

Lineflyer:
Thanks for this lesson. 
I got a brief understanding of the problem, however unless you are familiar with the way Java works and programming in general, its really hard to understand the relations and dependencies and how this system works :)

Will give it a try for the colors, then.

Regarding the colored cache name in details list:
Are you referring to the strikethrough formatting for disabled or something else?
The idea why I removed it was, to have at least one place where the cache name is readable in normal colors and without special formatting to accomodate the needs of visual impaired, color blind or people using screenreaders (where strikthrough might also be a problem).

moving-bits:
> I got a brief understanding of the problem, however unless you are familiar with the way Java works and programming in general, its really hard to understand the relations and dependencies and how this system works :)

Very understandable. Although Android has quite a lot of documentation, it lacks description of concepts in several places, and theming is one of those. Documentation for this is mostly "in the code", code is very nested here, and introducing Jetpack libraries has complicated things even more, so a lot it by "trial and error". (Or I'm missing the one magical place where all of this is described in a comprehensible way...)

> Regarding the colored cache name in details list: Are you referring to the strikethrough formatting for disabled or something else? The idea why I removed it was, to have at least one place where the cache name is readable in normal colors and without special formatting to accomodate the needs of visual impaired, color blind or people using screenreaders (where strikthrough might also be a problem).

Yes, this was what I was referring to. I understand your point and suggest to remove colorization in the cache details page, as you will still have the colorized info in the titlebar there. But please add a comment in the code, that removal was done deliberately, and for the reason you've mentioned, otherwise it might be re-added in a couple of years or so to further consistency :-)

Lineflyer:
@moving-bits  Thanks for your help and explanation.

I added a coloring proposal for theme-aware archived caches now. IMHO the color on light theme is pretty OK, so I left it unchanged but only lightend up the dark theme to improve readability.

| Before | After |
| --- | --- |
| image | image |
| image | image |
|  |  |
| image | !image |
| image | image |
 
PR is ready to be merged after review.

moving-bits (approved):
LGTM, thanks for tackling this issue!

Issue #13551, opened 2022-10-21
Empty trackable page if GC-website is in english


murggel:
### Describe your problem!

When loading the trackable page (e.g. TB9XHDG) an empty page is shown - only the name in the header is shown.

### How to reproduce?

Switch language at geocaching.com to english

Open the trackable page for a trackable in c:geo

### Actual result after these steps?

GC-Language is english:
![grafik](https://user-images.githubusercontent.com/70113341/197299121-a0279d47-5105-4e78-bae4-e47f8cb22d66.png)

GC-Language is german:
![grafik](https://user-images.githubusercontent.com/70113341/197299538-f490bea5-778e-4495-bef6-76b011f4185a.png)

### Reproducible

Yes

### c:geo Version

2022.10.22

### System information

c:geo version: 2022.10.22-b8e4962 developer build

Device:
-------
- Device type: sdk_gphone_x86 (sdk_gphone_x86, google)
- Available processors: 2
- Android version: 11
- Android build: RSR1.201013.001
- Screen resolution: 1080x2088px (392x735dp)
- Pixel density: 2.75
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 22.41.13 (150700-480714934)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:94
- Set language: en
- System date format: M/d/yy
- Time zone: GMT
- Debug mode active: no
- Last backup: Oct 17, 9:44 PM
- Routing mode: Walk
- Live map mode: true
- OSM multi-threading: false / threads: 1
- Map: Germany (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:primary:Cgeo/maps/germany.map
  - Atts: Map data (c) OpenStreetMap contributors
  - Theme: none

Filters:
-------
- Hide waypoints: -
- LIVE: None ([:inconclusive=false:advanced=false])
- OFFLINE: None ([:inconclusive=false:advanced=true])

Stored Filters (#0):

Services:
-------
- Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / BASIC
- Geocaching.com date format: dd.MM.yyyy
- Geocaching.com website language: de-DE
[...]

PR #13554, opened 2022-10-22
rel to #13551: fix nullpointerexception in ImageUtils.forEachImageUrlInHtml

eddiemuc:
rel to #13551: fix nullpointerexception in ImageUtils.forEachImageUrlInHtml

### DIFF
--- a/main/res/values/strings.xml
+++ b/main/res/values/strings.xml
@@ -1363,7 +1363,7 @@
     <string name="cache_description">Description</string>
     <string name="cache_description_table_note">Description contains table formatting which may need to be viewed at %s to be seen correctly.</string>
     <string name="cache_description_rendering">Rendering description, please wait…</string>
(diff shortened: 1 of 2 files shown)


Lineflyer:
@eddiemuc Shall we put this also onto branch `release` as it was reported for the market version?

eddiemuc:
> @eddiemuc Shall we put this also onto branch `release` as it was reported for the market version?

The reported NPE could not have happened in release since I did not put the original change there. If the problem is in release, it must have another origin

Lineflyer:
At least @murggel reported the problem for the market version and now says its gone (for now)...still might be unassociated.

murggel:
> At least @murggel reported the problem for the market version and now says its gone (for now)...still might be unassociated.

I use the nightly Version- probably noted lnly the date but notvthe nightly-postfix to version...

Issue #13526, opened 2022-10-16
Height at mainscreen doesn't provide value 


fm-sys:
### Describe your problem!

Height above ground level needs improvement. 

![Screenshot_20221016_030851](https://user-images.githubusercontent.com/64581222/196013213-55c374cf-b6b6-4ab9-8dae-b89302f3b959.jpg)

### How to reproduce?

-

### Actual result after these steps?

It shows 35m instead of 110m as it should. This might be caused by a bad GPS signal. However it's IMHO better to show nothing instead of totally wrong info.

Is there a way to calculate the vertical accuracy?

Furthermore, it would be nice to add the "above ground level" abbreviation. `xxx m AGL` (region specific, e.g. `xxx m ü. NN` in Germany) 

### Expected result after these steps?

Show more accurate information and make clear what this number is about. 

### Reproducible

Unclear

### c:geo Version

2022.10.16

SammysHP:
Honesty, GPS height is mostly that unreliable. That's only 75 m off.

fm-sys:
Another idea: Users can meanwhile add SRTM files to cgeo which could be used to gather the actual height. If no SRTM data is available, I would suggest to not show the (seemly wrong) height at all.

moving-bits:
It's the same value as displayed in our compass, which is an average of the last five altitude readings.
Altitude readings are only included, if it is available at all (which it isn't in the emulator), and if it has a vertical accuracy.

One thing that comes into my mind: The value is inaccurate currently until it got five readings (as missing readings are considered as a height of zero). For the compass this is only a temporary problem for the first 2.5 seconds, as there is a new reading every app. half second, but I'm not sure about the main screen's value. This could/should be improved.

Additionally, we could display the average vertical accuracy of those five readings, similar to what we do for the position.

moving-bits:
Additionally we should mark the displayed value with some symbol, if it's based on less than five readings, e. g. prepend a "~" or append a "(?)" or some other symbol. Or we could suppress the value on mainscreen until we have five readings.

PR #13556, opened 2022-10-22
improve height display ( #13526)

moving-bits:
## Description
- Displays height on home screen only when full readings (at least 5) are available
- Prepends "~" in compass height view, if less than five readings have been used


moving-bits:
retest this please

PR #13562, opened 2022-10-25
targetSdk 31 part 2

bekuno:
Set targetSdkVersion 31 also for the other sub projects and the tests. 
Additionally, set minSdkVersion 21 for tests and fix typo.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -29,8 +29,8 @@ android {
 
     defaultConfig {
         minSdkVersion 21
-        targetSdk 31
         //noinspection OldTargetApi
+        targetSdkVersion 31
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)
 
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -20,7 +20,7 @@ android {
     defaultConfig {
         minSdkVersion 21
         //noinspection OldTargetApi
-        targetSdkVersion 30
+        targetSdkVersion 31
 
         // include only those language resources from libraries which we actively maintain ourselves in the translation project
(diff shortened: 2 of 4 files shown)


main/build.gradle
```diff
@@ -29,8 +29,8 @@ android {
 
     defaultConfig {
         minSdkVersion 21
-        targetSdk 31
```
moving-bits, line 32: FYI: That's what Android Studio changed `build.gradle` to when I changed the targetSdk in project settings. 
Any difference between the two?

bekuno: > FYI: That's what Android Studio changed `build.gradle` to when I changed the targetSdk in project settings. Any difference between the two?

Interesting, I did not got this change.
What I found is https://stackoverflow.com/questions/67250362/what-is-difference-between-compilesdk-and-compilesdkversion-in-android-studio-gr/67251145#67251145 .
But in the release notes for AGP is nothing written for it.

I tested also the update to AGP 7.* (in small steps and in one step) with the upgrade assistent, but no change in this was was suggested/made.
If we really want to use the short names, then we need first the update to AGP 7 as written in the above stackoverflow text, and then we should change all 3 *Version names.

moving-bits: That's weird, as I have not upgraded AGP (and I'm also still running AS 2021.1.1) - don't know why it has inserted the short form.

But anyway, I do not see any need to use the short form currently, so let's stay with the long form for now. Thanks for fixing it.

bekuno: > FYI: That's what Android Studio changed `build.gradle` to when I changed the targetSdk in project settings.
> Any difference between the two?

I could reproduce it in AS for all tree Versions names with the "Project Structure" dialog.

Issue #13538, opened 2022-10-18
Hillshading not working with Freizeitkarte themes


dcjkfgdjhd:
### Describe your problem!

I placed some relevant HGT files (from USGS.gov) in the correct folder and am using an offline map. But I see no hillshading.

Tried different slider settings for both options in the settings dialogue, no change.

If there is something else I need to do, then c:geo does not clearly communicate that to the user.

### How to reproduce?

See above.

### Reproducible

Yes

### c:geo Version

2022.10.17

ztNFny:
Does the map theme you're using support hillshading? c:geos "default" theme does.

dcjkfgdjhd:
Aha, that seems to be the issue. I am using the Freizeitkarte v5 Theme.

What needs to be done to make hillshading work with Freizeitkarte and who can do it?

ztNFny:
Please contact the author of that theme to request hillshading to be supported by it.

SammysHP:
> Please contact the author of that theme to request hillshading to be supported by it.

Why does it depend on the map theme? Isn't it a separate layer? What needs to be changed in the map theme?

ztNFny:
> > Please contact the author of that theme to request hillshading to be supported by it.
> 
> Why does it depend on the map theme? Isn't it a separate layer? What needs to be changed in the map theme?
>
> IMHO we should try to make it work with all themes. We can't expect dozens of authors changing something they aren't even aware of.

That's the way it is implemented in Mapsforge, it's not specific to cgeo. The theme has to define at which level the hillshading layer is placed.

[hillshading /] as defined in https://github.com/mapsforge/mapsforge/blob/master/resources/renderTheme.xsd#L274
So every theme author is in fact aware of this element (or should be if he read the documentation)

dcjkfgdjhd:
Where in Github can I find the XML file defining the cgeo default theme? 

Maybe I can use that as an example to make the necessary changes to the Freizeitkarte theme.

ztNFny:
> Maybe I can use that as an example to make the necessary changes to the Freizeitkarte theme.

add this line before `</rendertheme>` should do it:
`<hillshading zoom-min="9" zoom-max="17" />`

dcjkfgdjhd:
> add this line before `</rendertheme>` should do it:
> `<hillshading zoom-min="9" zoom-max="17" />`

Yes, that's it. Thanks a lot!

SammysHP:
[...]

PR #13565, opened 2022-10-27
add master toggle for Hillshading

ztNFny:


SammysHP:
I suggest to include this in the theme settings.

bekuno:
@SammysHP I think you mean the Theme Options menu, isnt it? I think that would be a good place.

ztNFny:
Theme options could be confusing to users though. They might already have 1 or more entries there from the theme.
In that case they'd have to enable 2 options in the same dialog to get any result, either option without the other won't do anything.
Not sure how to solve this from an UX standpoint...

moving-bits:
I would not put it into theme options, as that menu is generated dynamically by the options given in the render theme and thus "completely out of our control", even the preference keys are generated dynamically - I would not mix this.

Regarding the "two options problem": We do not know whether a theme will have a separate entry for it or not. So if we want to make this setting switchable via UI, we may always run into that problem.

eddiemuc:
> I would not put it into theme options, as that menu is generated dynamically by the options given in the render theme and thus "completely out of our control", even the preference keys are generated dynamically - I would not mix this.

We already have today the "map scaling" options in the theme options menu. Those are not part of the theme but separate settings. We put them there so one could set the scale on a per-theme-base. So we have already something not coming from the theme in the theme options.

> Regarding the "two options problem": We do not know whether a theme will have a separate entry for it or not. So if we want to make this setting switchable via UI, we may always run into that problem.

If such a toggle is really wanted and should be usable independent from the theme configuration, I suggest offering and coding such a feature directly into mapsforge, then use it from c:geo. We have done it before.

ztNFny:
> If such a toggle is really wanted and should be usable independent from the theme configuration, I suggest offering and coding such a feature directly into mapsforge, then use it from c:geo. We have done it before.

As our map implementation will be replaced by VTM "shortly" I dont think it's worth it. We should rather look into an offline hillshading implementation for that.

moving-bits:
@cgeo/team: How do we want to proceed here? 
I propose to merge it to `master` to include it in next beta.

SammysHP:
I vote for **no merge** because visibility of the hillshading layer can be controlled by the map theme:

1. The theme must explicitly include a definition for the hillshading. Without an explicit definition hillshading is disabled.
2. The hillshading layer has a category.
3. The category can be assigned to a layer.
4. The category can be switched on and off in the stylemenu.

Thus the visibility of the hillshading can be controlled either by the theme author (like any other element that is shown in the map) or by the user, if the theme author includes the option to hide it (like any other element that is shown in the map).

ztNFny:
Do all (popular) themes that support hillshading provide the stylemenu option? If that were the case a master switch wouldn't be needed.

We already have some options to overwrite map theme options, that's precedent FOR including the switch.

SammysHP:
> Do all (popular) themes that support hillshading provide the stylemenu option?

No, but they also don't include options for all kinds of POIs, borders, bicycle routes etc. If they include a hillshading definition and don't provide an option, one can expect that the author intended to always show the hillshading.

> We already have some options to overwrite map theme options, that's precedent FOR including the switch.

Which options for example?

PS: I've uploaded a modified Freizeitkarte theme in my previous post that supports enabling/disabling hillshading via the stylemenu.

ztNFny:
> Which options for example?

The scale factors.

SammysHP:
> The scale factors.

Which is a separate option in the Mapsforge library IIRC and not primarily controlled by `<rule />` definitions in the rendertheme.

devemux86:
> If they include a hillshading definition and don't provide an option, one can expect that the author intended to always show the hillshading.

Not necessarily, they just support hillshading as option.

Enable or not of hillshading can be managed by apps,
as even if the map style supports it, the apps may not.
(it requires different layers)

fm-sys:
Freizeitkarte will not provide such a toggle in their theme (see https://github.com/cgeo/cgeo/issues/13538#issuecomment-1312451595). And most other themes don't have such a toogle as well AFAIK...

main/res/xml/preferences_map.xml
```diff
@@ -79,6 +79,11 @@
             android:summary="@string/settings_info_hillshading"
             android:title="@string/settings_info_hillshading_title"
             app:iconSpaceReserved="false" />
+        <CheckBoxPreference
+            android:defaultValue="false"
+            android:key="@string/pref_maphillshading"
+            android:title="@string/settings_hillshading_enable"
+            app:iconSpaceReserved="false" />
         <Preference
```
moving-bits, line 87: You could make the other hillshading options depending on the master switch, so that they get greyed out if hillshading is disabled, see `android:dependency="@string/pref_maphillshading"`

moving-bits:
As decided in today's dev meeting we plan integrate this into our upcoming beta on `release`, therefore I changed target to `release`.

main/res/xml/preferences_map.xml
moving-bits: implemented with #13658

PR #13582, opened 2022-10-30
fix #13138: make image details open also in large cache lists

### LINKED ISSUE #13138: New Image gallery uses too many resources

eddiemuc:
fix #13138: make image details open also in large cache lists

### DIFF
--- a/main/res/layout/imageview_image.xml
+++ b/main/res/layout/imageview_image.xml
@@ -27,6 +27,27 @@
             tools:src="@drawable/cgeo_notification"/>
     </FrameLayout>
 
+    <com.google.android.material.progressindicator.CircularProgressIndicator
+        android:id="@+id/image_progress_bar"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:indeterminate="true"
+        android:layout_gravity="center"
+        android:layout_centerInParent="true"
+        android:foregroundGravity="center"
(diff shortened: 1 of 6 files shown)


eddiemuc:
Side notes:
- PR fixes also a problem that in some situations images can no longer be clicked/opened in Image Viewer
- PR also introduces "image not available" handling and "image loading" handling in Image viewer

PR #13588, opened 2022-11-01
remove local checkstyle error report in BackupUtils

eddiemuc:
remove local checkstyle error report in BackupUtils

This removes a checkstyle error which I always get when running it locally. For whatever reason it does not seem to occur on server side:

image


eddiemuc:
And does anyone now why this does not occur on server side?

eddiemuc:
I still don't understand why the checkstyle running on Jenkins did not fail the build due to the BackupUtils-problems above (in constrast to Codacy, it checks the whole codebase and not just changed parts afair).

Issue #13552, opened 2022-10-22
Finalize language switch removal


moving-bits:
Collection of tasks that need to be done to finalize removal of language switching:

- [ ] Rework trackable parsing (current owner, tb status) - after HQ have done their part (documentation: see emails to HQ)
- [ ] Reenable / rework tests: see #13458

PR #13589, opened 2022-11-01
language-independent 'tb spotted' parsing (rel. to #13552)

moving-bits:
## Description
- adapt tb location parsing to recent website changes (as agreed upon with HQ)
- re-enable tb location tests

## Test cases:
- tb 12345 (a tb currently in the hands of the owner)
- tb 12347 (a tb in the hands of a user other than owner)
- tb 12351 (a tb marked missing)
- tb TB6P14F (a tb currently in a cache)
- and one of my own tb which is currently in my own hand (so similar to first case, but owner being the user currently logged in)


moving-bits:
Failing tests are unrelated (GeoKrety), so PR should be fine to be merged

Issue #4394, opened 2014-09-30
Store trackables


Bananeweizen:
I've heard this request multiple times from different users now. Is there any good reason why we can't explicitly store trackables? E.g. potential use cases are:
- Looking up a trackable _before_ visiting the cache and then at the cache checking the target again (even without mobile data connection).
- Having a kind of inventory/bookmark list for interesting trackables, or for own trackables
- ...

Anyone with technical or usage related remarks?

kumy:
Related to #1883?

anondev-sudo:
Agree with this. Whe  I log geocaching on slow public transport wifi, it's annoying that I have to wait every time to load trackables in inventory from the server before I can submit log. It would be nice if that would be cached. It should fetch only the first time you try to log some cache during working session. When you don't drop TB to a cache while logging or you don't add new TB to your inventory via cgeo, then it shouldn't be fetched next time you log geocache and should use the cached information to make logging faster.

SammysHP:
> Whe I log geocaching on slow public transport wifi, it's annoying that I have to wait every time to load trackables in inventory from the server before I can submit log. It would be nice if that would be cached.

That isn't possible because c:geo needs to download the logging page to get various information. This issue is only about storing the description of a TB.

ztNFny:
Unless you have a lot of caches in inventory no additional web request is performed to load the TBs.

anondev-sudo:
Every time I log a cache it loads information to log it from server including those TB inventory

eddiemuc:
When logging, c:geo retrieves following information from website:
* log types allowed for this user for this geocache. We could probably cover this by syncing c:geo log ruleset to the ruleset used by gc.com
* reporting types allowed for this geocache. this is not used any more by gc.com connector
* available trackables eg inventory --> this is the core of the issue in my understanding
* available favorite points --> this needs to be handled as well

I agree that it would make sense to store the inventory on c:geo side instead of retrieving it every time again from server. "Refresh rules" for inventory can be discussed.
We need some solution for available fav points as well though. This could also be stored offline imho.

eddiemuc:
Update: I see now that available report problem types are not used ANYWHERE any more in c:geo. The related code pieces could be removed (refactoring). That's a different topic though

ztNFny:
We already have a solution in place to avoid those calls to the server, it's offline logging.
[...]

Issue #838, opened 2011-11-27
View users TB-inventory


Lineflyer:
Feature request:
I would be good to have an option to see the list of TBs which are currently in the hands of the user and open the detail page from there.
Use Case:
Sometimes I want to look up the goals of the TB i currently have with me. The only possibilty is to use the search for each TB which means I first have to lookup the  tracking code every time.
Also in the log view the TBs are ony listed but not linked to the TB-details.

On http://www.geocaching.com/my/default.aspx there is a list of the users inventory, which could be used to make a inventory list also in cgeo.
This could be a new menuitem in search and there could also be a link from logview to open the listed TBs detail page.

Issue #420, opened 2011-09-12
My caches/Trackables


triakcz:
hi,

could you create main menu item to access  list of my caches and traclables ? I think that it may be usable.

thanks Ondra

ldmpub:
+1

I really miss a "Show trackables in my inventory" with both TB and GK I currently have in my hands.

caigner:
As **owner** of serveral geocaches and trackables I would also love to see a "My own geocaches" and "My own trackables" button on the start page. There is room for those two buttons and it would make c:geo even more useful to owners.

Yes, there is a search page and a icon on the top right which lets me download a list of my own geocaches, but hey, it is all about usability. :-)

So: +1 here from me.

Thanks guys for the outstanding job you do! c:geo really rocks!

Kebap:
Here is an idea on how to implement the trackable inventory I described in #7081 

When you open a cache page to log a visit, you will see a section "inventory" at the bottom with links to all trackables in your hand (and more functionality relating to the logging). This list (sans logging functionality) can be copied as is and displayed directly when you click a (new) main menu item "inventory". This seems to be all. Maybe add some other existing menu options like sort, etc.

geocachermgo:
On an event I discussed with someone who is using cgeo +other apps (eg: geocaching app to easil access the Message Center.
He was missing a list to store trackables (mainly his own).

PR #13601, opened 2022-11-03
preserve last tb action per trackable (fix #13590)

### LINKED ISSUE #13590: Remember trackable actions per trackable
### Tell us your idea!

Currently the default action for trackables in your inventory is "No action". You can change in the settings that the default action should be "visited".

If you own some trackables that travel with you and some that don't, it's very tedious to select "no action" for stationary TBs and "visited" for travelling ones.

Application should remember what action has been chosen for every TB in your inventory and set it as the default for the next logs. 

The new TBs you add to your inventory should have the default action set, i.e. "no action" or "visited", depending on the settings.

### Is there an alternative?

Selecting appropriate action on every log.

### Additional Information

_No response_

moving-bits:
## Description
Remember last selected action per trackable


moving-bits:
retest this please

eddiemuc:
I might be wrong, but I feel a tendency to use DBExtension very much in last time. I thought this was meant to store small amounts of values not fitting nicely into our db table structure. Now I get the impression it is meant to "turn" our relational db approach into a NoSQL-for-the-poor (store everything in a key-value-fashion).

In this PR: isn't there a db table storing trackables where the last state could be stored in a new column? Instead of using a DBExtension?

eddiemuc:
> Under what conditions does a trackable gets added to the cg_trackable table?

I honestly don't know. I assumed it would contain the owned trackables, because afaik we don't have "offline trackables" in c:geo.

If this table is filled with something else, then please ignore my previous comment.

Lineflyer:
> If this table is filled with something else, then please ignore my previous comment.

I think from back in my head it contains the inventory of caches linked to the caches table....I might be wrong though as I never in depth took a look at our SQL-DB.

moving-bits:
Ok, I've done a bit of research: 

I cross-checked for what @Lineflyer has suggested. Indeed, on saving a cache page, the tb in the inventory of the cache gets stored to `cg_trackables` table. But although this tb data is stored already, tapping on a tb leads to an error when internet connection is switched off. I'll create a new issue for that.

Then I cross-checked our use-case here, whether tb data for the current user's inventory gets stored on logging a cache. This is _not_ the case, so I guess we need some mechanism to store the relevant data.

I see two options here:
- a) We can use the `cg_trackables` table for those tb as well.
  - Pro: No additional data structure needed
  - Con: Table structure consumes more space than `cg_extension` row; possibly additional data requests needed to fill the columns
- b) Or we use the `cg_extension` approach of this PR
  - Pro: Self-contained data structure with only the data needed for this use-case; no additional table; less space-consuming

Personally I would go with approach b, but I'm open for the other approach as well. In either case, I found out that we should adjust the index structure anyway, as
- table `cg_extension` does not have a key on the `_type` column, and
- table `cg_trackables` does not have a key on the `tbcode` column (which we would need for our use-case).

Depending on how we decide to continue here, I'll open an issue for updating the index structure as well.

eddiemuc:
Thanks for the detailed analysis!

 Since owned trackables are not stored in the db table yet, I would vote for b) personally. I assume the table is tied to a concrete cache (e.g. has a geocode field) which would make it unsuitable for storing owned trackables anyway

Lineflyer:
Could that be a foundation for #420 (better described in #838) ?

moving-bits:
> Could that be a foundation for #420 (better described in #838) ?

#838 as well as #4394 is about storing a user's tb inventory and/or a user's owned tb in our database and offer views for it. Indeed our current `cg_trackables` table might be a base for storing the relevant data (and `cg_extension` wouldn't be), but we would need to distinguish the different reasons why a trackable is stored there - cache's inventory, user's inventory, owned by a user -, and check the related queries. 

Getting the data, keeping it up to date, and adding the views for it is a somewhat larger task, so I'd say let's not wait for that.

PR #13610, opened 2022-11-06
fix several errors mentioned by prelaunch flight

moving-bits:
## Description
Before releasing a new version, Play Store does a prelaunch flight on several (virtual) devices, analyzing the resulting user experience. For c:geo there are quite a few results regarding accessibility (eg: missing labels) and other categories (eg: overlapping items, depending on translation). This PR intends to fix a couple of them.


moving-bits:
failing tests are unrelated / GeoKrety server having problems?

PR #13611, opened 2022-11-06
UnifiedMap (GM): Avoid crash on map end/restart

moving-bits:
## Description
Prevents a crash happening on closing and reopening UnifiedMap (GM)


moving-bits:
failing tests are unrelated / GeoKrety server having problems?

PR #13584, opened 2022-10-31
agp 7.3.1

bekuno:
Move AGP from 4.2.2 to 7.3.1, partially using the assistent.

Implement some migration steps ans resolve deprecation warnings
- Rewrite deprecated operators
- Migrate packagingOptions DSL properties to sub-blocks
- Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated

PR #13634, opened 2022-11-09
AGP 7.x

bekuno:
Move AGP from 4.2.2 to 7.1.3, partially using the assistent. Renaming `lintOptions` to `lint` as required from AGP.

The lint tool results in
~~~
res\values-fr\strings.xml:347: Error: For locale "fr" (French) the following quantities should also be defined: many [MissingQuantity]
    <plurals name="caches_more_caches_next_x">
~~~
As `plurals` is not supported for `french` on crowdin we have to disable it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:4.2.2'
+        classpath 'com.android.tools.build:gradle:7.1.3'
 
         // un-mocking of portable Android classes
         classpath 'com.github.bjoernq:unmockplugin:0.7.9'
--- a/build.gradle
+++ b/build.gradle
@@ -90,9 +90,12 @@ allprojects {
             resolutionStrategy {
                 componentSelection {
                     all { ComponentSelection selection ->
-                        if (['alpha', 'beta', 'rc', 'cr', 'm', 'pr'].any { qualifier -> selection.candidate.version.contains(qualifier)}) {
-                            selection.reject('Release candidate')
-                        }
+                        // deactivate this rule for now, as alpha software is needed as dependency for tests
+                        // from com.android.tools.utp:android-test-plugin-host-additional-test-output:30.3.1,
+                        // see https://github.com/cgeo/cgeo/pull/13634
+                        //if (['alpha', 'beta', 'rc', 'cr', 'm', 'pr'].any { qualifier -> selection.candidate.version.contains(qualifier)}) {
+                        //    selection.reject('Release candidate')
+                        //}
 
                         // see https://assertj.github.io/doc/#assertj-core-android
                         // AssertJ Core 3.x is compatible with Android API Level 26+, except for soft assertions and assumptions.
--- a/build.gradle
+++ b/build.gradle
@@ -113,10 +113,6 @@ allprojects {
                 }
                 // new versions of checkstyle use guava 27, which has troublesome dependencies: https://groups.google.com/forum/#!topic/guava-announce/Km82fZG68Sw
                 force 'com.google.guava:guava:26.0-jre'
-
-                // force this version because of conflict to our resolutionStrategy from lint-gradle
-                // TODO check necessity after every gradle update (gradlew :main:lintBasicDebug)
-                force 'com.android.tools.build:aapt2-proto:4.2.2-7147631'
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:7.1.3'
+        classpath 'com.android.tools.build:gradle:7.3.1'
 
         // un-mocking of portable Android classes
         classpath 'com.github.bjoernq:unmockplugin:0.7.9'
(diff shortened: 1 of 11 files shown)


bekuno:
Damn. All the time
~~~
23:08:02  FAILURE: Build failed with an exception.
23:08:02  
23:08:02  * What went wrong:
23:08:02  Execution failed for task ':main:connectedBasicDebugAndroidTest'.
23:08:02  > Could not resolve all files for configuration ':main:_internal-unified-test-platform-android-test-plugin-host-additional-test-output'.
23:08:02     > Could not find com.google.testing.platform:android-device-provider-local:0.0.8-alpha07.
23:08:02       Searched in the following locations:
23:08:02         - https://repo.maven.apache.org/maven2/com/google/testing/platform/android-device-provider-local/0.0.8-alpha07/android-device-provider-local-0.0.8-alpha07.pom
23:08:02         - https://jitpack.io/com/google/testing/platform/android-device-provider-local/0.0.8-alpha07/android-device-provider-local-0.0.8-alpha07.pom
23:08:02       Required by:
23:08:02           project :main > com.android.tools.utp:android-test-plugin-host-additional-test-output:30.1.3
~~~

But now I found the reason. Running
`gradlew :main:connectedBasicDebugAndroidTest -i `
(with the `-i` = `info`) parameter I got
~~~
> Task :main:connectedBasicDebugAndroidTest FAILED
Selection of com.google.testing.platform:android-device-provider-local:0.0.8-alpha07 rejected by component selection rule: Release candidate
~~~
Heureka - we have such a rule in our global `build.gradle`.

What now? 
As forcing to the latest non alpha version for the dependencies here did not work - see #13584,
I propose to deactivate (or remove?) this rejection rule.
Maybe we should instead add such a rule in the `dependabot.yml` file, as this is now our normal method for dependency checking.

build.gradle
```diff
@@ -90,9 +90,12 @@ allprojects {
...
                     all { ComponentSelection selection ->
-                        if (['alpha', 'beta', 'rc', 'cr', 'm', 'pr'].any { qualifier -> selection.candidate.version.contains(qualifier)}) {
-                            selection.reject('Release candidate')
-                        }
+                        // deactivate this rule for now, as alpha software is needed as dependency for tests
+                        // from com.android.tools.utp:android-test-plugin-host-additional-test-output:30.3.1,
+                        // see https://github.com/cgeo/cgeo/pull/13634
+                        //if (['alpha', 'beta', 'rc', 'cr', 'm', 'pr'].any { qualifier -> selection.candidate.version.contains(qualifier)}) {
+                        //    selection.reject('Release candidate')
+                        //}
```
Lineflyer, line 99: Finally you found the reason! Nice!

Lineflyer:
I will try out this PR configuration in my local AS.

main/lint.xml
```diff
@@ -26,6 +26,8 @@
     <issue id="MissingTranslation" severity="ignore" />
     <!-- old translation will be deleted by the next crowdin import -->
     <issue id="ExtraTranslation" severity="ignore" />
+    <!-- we will take what is given by crowdin - For locale "fr": "many" is missing currently -->
+    <issue id="MissingQuantity" severity="ignore" />
```
Lineflyer, line 30: Thats strange.
Crowdin follow the unicode plurals specification, which has no "many" for french: https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html

Is this configurable in Lint?

Lineflyer:
> I will try out this PR configuration in my local AS.

Cloned the PR and it upgraded and built without a problem.

main/lint.xml
bekuno: > Thats strange. Crowdin follow the unicode plurals specification, which has no "many" for french: https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html#fr

This is not the latest one (CLDR v37 vs. v38).
Please see at https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#fr

> Is this configurable in Lint?

No.
Only using the special lint rule. As we use crowdin for translation, this should not be problematic.

bekuno:
The PR is now ready to merge from my side.
The remaining new 14 lint warnings are no real new problems.
For the Android API relevant lint warnings I created dedicated issues.
Someone with deeper Androin (API) knowledge should resolve them.

main/lint.xml
bekuno: Edit: CLDR is currently at version 42.

bekuno:
> > I will try out this PR configuration in my local AS.
> 
> Cloned the PR and it upgraded and built without a problem.

Thats fine.
Then let's use it on master.

bekuno:
You will find the new lint warnings at https://ci.cgeo.org/job/cgeo-CI_commit-build/941/androidLintResult/new/

main/lint.xml
bekuno: Crowdin knows this issue, see https://community.crowdin.com/t/missing-many-plurals-for-french-language.

The state is:
~~~
We will update all languages where we missing new standards, like the “many” plural form. 
The appropriate epic task has been already created.

Still, it wouldn’t be released soon, because it requires additional investigation.

For example, some users already have French as 100% translated, and adding a new form means they 
will lose this progress. As a result, if their integration is based on the export option “export only 
fully translated file” it wouldn’t work until this new form becomes translated. For those who are fine 
with current forms, this change may quite uncomfortable.

Right now we’re discussing a solution that would work great for all our clients. 
You can sure that the languages will be definitely updated according to new standards.
~~~

Issue #13603, opened 2022-11-03
Filter for "Favorites = ?"


MagpieFourtyTwo:
### Tell us your idea!

When you import caches from e. g. a bookmark list, the list usually contains both, caches that were already saved before (and already contain the basic stuff) and caches which your c:geo has never seen since.

But although the latter can easily be identified, as they display a favorites count of "?", it is not possible to filter these out (at least I have not found one). Nevertheless, I would really like to be able to do so, in order to refresh only these (to get at least the basic info like spoiler pics, some logs, fav count and percentage ...) 

BTW: Refreshing the whole list (or "just" all caches with "Favorites = 0") is not an option at all, as updating several thousand or even hundreds of caches still takes ages ...

### What solution would you suggest?

Add a filter for "Favorites = ?"

### Additional Information

Just wondered about the displayed filter text ... which is BTW different, if you either use the sliders or enter 0 by keyboard:
- after entering a "0" as minimum value and a "0" as maximum value via keyboard,
   the display text is "Favorites: 0.0" - and there will only caches be displayed with fav count of 0 - perfectly fine.
- BUT: after *sliding* both sliders to "0" (quit the probably opened input box with [Cancel]),
  the display text is "Favorites; <0.0", which is a) kind of weird, as the scale has nothing less than 0,
  but b) even wrong, as not only caches with less than 0 favs (most probably those with "?") are displayed,
  but even the ones with fav count == 0 - thus the display text should be "Favorites: <=0.0."

murggel:
> - BUT: after *sliding* both sliders to "0" (quit the probably opened input box with [Cancel]),
>   the display text is "Favorites; <0.0", which is a) kind of weird, as the scale has nothing less than 0,
>   but b) even wrong, as not only caches with less than 0 favs (most probably those with "?") are displayed,
>   but even the ones with fav count == 0 - thus the display text should be "Favorites: <=0.0."
> 


I will have a look at this

PR #13660, opened 2022-11-13
rel #13603: correct filter display string for minimum / maximum values

murggel:
## Description

adapt filter display string: 
* use "less/greater than or equal to" instead of "less/greater than"
* use "equal" if range is limited to min or max value

"Favorites: < 2.0" => "Favorites: ≤ 2.0"
"Favorites: < 0.0" => "Favorites: ≤ 0.0" (not really correct, but otherwise the logic in the filter has to be changed)

## Related issues

rel #13603

### DIFF
--- a/main/src/cgeo/geocaching/filters/core/NumberRangeGeocacheFilter.java
+++ b/main/src/cgeo/geocaching/filters/core/NumberRangeGeocacheFilter.java
@@ -63,14 +63,9 @@ public abstract class NumberRangeGeocacheFilter<T extends Number & Comparable<T>
     public <V extends Number & Comparable<V>> void setMinMaxRange(@NonNull final V leftValue, @NonNull final V rightValue,
                                                                   @NonNull final V minValue, @NonNull final V maxValue,
                                                                   @NonNull final Func1<V, T> valueConverter) {
-        boolean foundMinUnlimited = leftValue.compareTo(minValue) <= 0;
-        boolean foundMaxUnlimited = rightValue.compareTo(maxValue) >= 0;
-        if ((leftValue.equals(rightValue))
-                || (leftValue.compareTo(maxValue) >= 0)
-                || (rightValue.compareTo(minValue) <= 0)) {
-            foundMinUnlimited = false;
-            foundMaxUnlimited = false;
-        }
+        final boolean foundMinUnlimited = leftValue.compareTo(minValue) < 0; // leftValue < minValue
+        final boolean foundMaxUnlimited = rightValue.compareTo(maxValue) > 0; // rightValue > maxValue
+
         setMinMaxRange(foundMinUnlimited ? null : valueConverter.call(leftValue),
                 foundMaxUnlimited ? null : valueConverter.call(rightValue));
     }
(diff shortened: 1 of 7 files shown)


Lineflyer:
Any objections for this one?
Looks helpful.

main/src/cgeo/geocaching/filters/core/NumberRangeGeocacheFilter.java
```diff
@@ -59,6 +60,22 @@ public void setMinMaxRange(final T min, final T max) {
         numberRangeFilter.setMinMaxRange(min, max);
     }
 
+    public <V extends Number & Comparable<V>> void setMinMaxRange(@NonNull final V leftValue, @NonNull final V rightValue,
+                                                                  @NonNull final V minValue, @NonNull final V maxValue,
+                                                                  @NonNull final Func1<V, T> valueConverter) {
+        boolean foundMinUnlimited = leftValue.compareTo(minValue) <= 0;
+        boolean foundMaxUnlimited = rightValue.compareTo(maxValue) >= 0;
+        if ((leftValue.equals(rightValue))
```
eddiemuc, line 68: I don't quite understand the meaning of this if statement. why would you set `foundmin/maxunlimited` to `false` under these circumstances, what is the intention for this? 
Also, this if statement would also set `foundmin/maxunlimited` to false if e.g. `leftValue` is equal to `rightValue`, but both are **outside** of the `minValue/maxvalue` range. Is this intended?

murggel: If left value equals right value, there should be `value` displayed, and no comparison symbol. 

The outside situation occurs in the situation, when the value has a converting function (e.g. favorite percentage). If the value is entered via the dialog, left and right are not equal, but both >= max:
grafik
In this case: without the enhanced if-statement `>= 100%` will be displayed - I thought `100%` will be better. Thats why this statement. Probably I will add some comments to it for better understanding

murggel: Probably it depends will, on how have the filters are implemented, whether to display 100% or >= 100%
I will check that...

eddiemuc: If I understand correctly then you don't want to change anything about the filter's logic but just the display string in the topm is that correct?

Because in this case all your code should be done in method `getUserDisplayString`. The `setMinMaxValue` method, where you placed parts of your change, will also set the inner state of the filter, and thus your changes may also affect the filters logic (not only it's display string)

murggel: Correct, I only wanted to change the display. 
Then I recognized the >= 100% issue and thought that could be enhanced.
But I wasn't really aware, that I changed the filter logic with my implementation  - so then we probably stay with the >= 100% and I will stay with changes only in `getUserDisplayString`

murggel: I removed the code which changed the minMaxRange for the filter, but kept the new function `setMinMaxRange` for the logic when to set the limit to `null`

eddiemuc: LGTM code-wise, thanks @murggel.
The failing tests seem unrelated to me, should this be handled before a merge somehow?

eddiemuc:
Retest this please

PR #13600, opened 2022-11-03
osm.paws.cz - new mapsforge maps source

osmpaws:
## Description
Add in-app download option for another (mapsforge) offline map source - osm.paws.cz

## Related issues
#13273 

## Additional context
Map and theme download is working, I have done some tests by myself with Android Studio. Whole new code is derived from existing. There is also source code for theme legend that is not done and is still almost original. However the option to show theme legend is not really in the menu for some reason. So this source code file can be removed or stay until I will finish it.

PR #13668, opened 2022-11-14
UnifiedMap: Add map legend for osm.paws.cz theme (rel. to #13600)

moving-bits:
## Description
Adds theme legend for the new osm.paws.cz map source / theme to UnifiedMap


Lineflyer:
retest this please

PR #13726, opened 2022-11-25
Use post request for uploading personal note to OC (okapi)

### LINKED ISSUE #13721: Personal note of 6000 char length fails to upload on opencaching.de
### Describe your problem!

Upload of a long personal note to opencaching.de is failing

see download of personal note #13651

### How to reproduce?

* Open e.g. OC16E8B in c:geo
* Enter a really long personal note (e.g. 6000 char)
* upload the note

Error appears: "Error uploading personal note"

I thought personal note up to a length of 20k is possible

### Reproducible

Yes

### c:geo Version

2022.11.22-NB

### System information

_No response_

### Additional Information

I tried to upload the downloaded long personal note from #13651 via postman and got the response 
`"414 Request-URI Too Long" 
The requested URL's length exceeds the capacity limit for this server.`

Note of 5700 char length ist working, 5800 length is not working...

The max personal length is 20k but it seems that it is not possible to upload a note of 20k via okapi

murggel:
## Description

On updating the personal note for OC with get-request, the url will be getting too long. 
Do a post-request with body instead.

## Related issues

fix #13721


murggel:
don't know why the test fails, succeeded on my local machine - have to check

Lineflyer:
Do we have any other OC test which requires a login?
Could be, that our CI either not have keys or no user configured?

murggel:
We have OC-test, but I don't know if they require a login...

murggel:
Seems that all existing Okapi-Client-test are only getting cache-data, which does only require a valid OKAPI key but you do not need to set credentials in the app (see `testOCSearchMustWorkWithoutOAuthAccessTokens`)

So can we add credentials to the tests or shall I delete the test?

Lineflyer:
I think it would be preferable to add credentials to be able to enhance our OC tests (thanks BTW for adding a test).
I do have however no idea how to achieve this on our CI and test system.

murggel:
Unfortunately I don't know either. We don't have that for gc.com either?

Until then I can probably somehow disable the test, so I can check-in the post-request?

bekuno:
I checked or CI (PR build, should in the same way the other builds).
In the initialization in the script `cgeo build and test for new commits and pull requests v2` the `main/res/values/keys.xml` is deleted. After that the gradle task `verifyCgeoKeys` will fill this file from the CI file `private.properties`.

This means we have to put OC.de credentials in this file.

@Lineflyer @mucek4 Can you request and integrate such a API key?

murggel:
Here is the site to request the API key

https://www.opencaching.us/okapi/signup.html

murggel:
Do I have then to login with a specific user in the test  - do we have a test-user?

bekuno:
> Do I have then to login with a specific user insidethe test - do we have a test-user?

Good question. You are right, we should have a login-user.

The integration is generated in the gradle task `copyDefaultPreferencesToAndroid`.
The CI has a `/srv/cgeo.geocaching_preferences.xml` as source.
 
This should then be patched.

bekuno:
> > Do I have then to login with a specific user insidethe test - do we have a test-user?
> 
> Good question. You are right, we should have a login-user.
> 
> The integration is generated in the gradle task `copyDefaultPreferencesToAndroid`. The CI has a `/srv/cgeo.geocaching_preferences.xml` as source.
> 
> This should then be patched.

I verified it with tests.
We need a dedicated OC.de test account and the `/srv/cgeo.geocaching_preferences.xml` has to be patched.
We need at least 
~~~
   <boolean name="connectorOCActive" value="true" />
    <string name="ocde_tokensecret">...</string>
    <string name="ocde_tokenpublic">...</string>
~~~

murggel:
@bekuno can I do something here?

bekuno:
> @bekuno can I do something here?

No. Someone with physical access to the CI has to set the OC.de configuration. I do not have this possibility.
@Lineflyer @mucek4 ?

bekuno:
@murggel The test is green now - what was changed?

murggel:
> @murggel The test is green now - what was changed?

Nothing intentionally... I will check when I am back to my machine

murggel:
I updated the PR to the latest master - probably some changes there

bekuno:
Found it - the test did not run, see https://ci.cgeo.org/job/cgeo-CI_PR-build/3888/testReport/cgeo.geocaching.connector.oc/OkapiClientTest/.
The `@Test` is missing.

murggel:
> > @bekuno can I do something here?
> 
> No. Someone with physical access to the CI has to set the OC.de configuration. I do not have this possibility.
> @Lineflyer @mucek4 ?
> 

Any news on this, otherwise I would deactivate the test, so the functionality can be merged

Issue #13730, opened 2022-11-26
Crash on 2022-11-24+ nightly build when downloading geocaches


Lomanic:
### Describe your problem!

Trying to load or refresh any GC.com geocache from the Internet with the 2022.11.25-NB-5942c61 nightly build results in a crash (after at least _Loading logs_ in the pop up).

This started after an upgrade to the 2022-11-24 nightly build.

### How to reproduce?

1. display a locally downloaded geocache
2. pull to refresh
3. have it crash after _Loading logs_ is displayed in the popup

This also occurs with a cache never downloaded and background refresh (batch downloading a list of geocaches).

### Actual result after these steps?

The app crashes after _Loading logs_ is displayed in the popup

### Expected result after these steps?

No crash, the geocache would be properly downloaded

### Reproducible

Yes

### c:geo Version

2022.11.25-NB-5942c61

### System information

> Copy and paste it from c:geo Menu -> About c:geo -> Swipe right to System

This triggers another crash, here's the information from the _Report as email_ menu

```text
c:geo version: 2022.11.25-NB-5942c61

Device:
-------
- Device type: Redmi Note 4 (lineage_mido, Xiaomi)
- Available processors: 8
- Android version: 7.1.2
- Android build: lineage_mido-userdebug 7.1.2 NJH47F 535058185f
- Screen resolution: 1080x1920px (400x687dp)
- Pixel density: 2.7
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 22.44.17 (040408-488410709)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:146
- Set language: en_CA (system default)
- System date format: y-MM-dd
- Time zone: GMT+01:00
- Debug mode active: no
- Last backup: Nov 5, 20:23
- Routing mode: Walk
- Live map mode: true
- OSM multi-threading: false / threads: 1
- Map: Google: Satellite
  - Id: cgeo.geocaching.maps.google.v2.GoogleMapProvider$GoogleSatelliteSource
  - Atts: none
  - Theme: none

Filters:
-------
- Hide waypoints: -
[...]

PR #13732, opened 2022-11-26
Revert "Bump commons-compress from 1.20 to 1.22"

moving-bits:
## Description
The crashes described in #13730/#13731 point to java.nio not supported. Dependency on this got introduced by an update of apache.commons.compress - which this PR therefore reverts.


bekuno:
@Lomanic Thanks for the detailed report, so that the crash reason was good findable
@moving-bits Thanks for the fast revert.
I add it additionally to the dependabot ignore list.

bekuno (approved):

moving-bits:
retest this please

Lomanic:
Thanks for the really quick fix, I can confirm there's no crashes anymore on Android 7.1.2 with c:geo 2022.11.27-NB-dd67eb0

PR #13733, opened 2022-11-26
Use Worker instead of ForegroundService for receiving downloads

moving-bits:
## Description
Using a ForegroundService gets more and more restricted with newer Android versions. Starting with Android 12, an app running in the background is no longer allowed to start a `ForegroundService` which blocks receiving downloads when c:geo was put into background meanwhile.
This PR mitigates this by using a `Worker` instead of a `ForegroundService` (as recommended by Google).

Optional things (that could be implemented as separate PR) are:
- Do we need to support import tasks running longer than ten minutes? Then we need to adapt to something like this: https://developer.android.com/topic/libraries/architecture/workmanager/advanced/long-running
- Do we want to use a `ForegroundService` (which might be a bit more responsive) when available (= app is in foreground state)? Then we need to be able to detect foreground state and create either a `Worker` or a `ForegroundService`. This PR is prepared for this switching, but does not implement the actual distinction.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -429,6 +429,9 @@ dependencies {
     implementation "androidx.lifecycle:lifecycle-common-java8:$lifecycleVersion"
     implementation "androidx.lifecycle:lifecycle-reactivestreams:$lifecycleVersion"
 
+    // workmanager (to replace foreground services for downloader)
+    implementation "androidx.work:work-runtime:2.7.1"
+
     // Support Library AppCompat
     implementation 'androidx.appcompat:appcompat:1.5.1'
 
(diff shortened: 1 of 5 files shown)


fm-sys:
Looks like a good alternative to foreground services. Though I'm not sure how it will perform on OEM devices (e.g. Huawei is quite restrictive when it comes to background tasks).

To me it sounds like as if [long-running workers](https://developer.android.com/topic/libraries/architecture/workmanager/advanced/long-running) use a foreground service internally, so this might be a good solution. Do you know if these workers can be started from the background in Android 12?

However, I think that almost all downloads should be doable it less then 10 minutes.

moving-bits:
Thanks for looking into it. Using a `Worker` is the new default approach for long running background tasks on Android, so hopefully all Android flavors follow those guidelines.

So far I'm using the regular `OneTimeWorker`, without special precautions for long running tasks. To my understanding even those map to a `ForegroundTask` on older systems internally. They are bound to a ten minute maximum on newer systems, but as this task is only about unzipping and copying an already downloaded file, this should be ok for most maps and devices. Only very slow devices in combination with large zipped map files (like OAM Germany map) may cause problems.

moving-bits:

PR #13744, opened 2022-12-02
Correctly detect archived caches in search results (fix #13619)

### LINKED ISSUE #13619: Archived geocaches incorrectly shown on map when filtering based on Cache Owner
### Describe your problem!

When map filter is configured to use "Cache Owner" condition to only show caches from a specific owner, archived caches are shown on the map even when map is configured to not show archived caches.  Additionally, those caches are not shown as archived, but rather shown as active.  Only after opening of such geocache (by clicking on the geocache icon on the map and selecting "More details", or clicking the "refresh" / "reload" button), which causes the geocache to be cached on the device, c:geo correctly determines the "archives" status and removes such cache from the map.  However, this does not solve the problem permanently, and it re-appears once cached information about the geocache is removed.  A possible workaround is to store such geocache in the device, so that its status is never forgotten.

### How to reproduce?

1. Switch to map and open filter settings.  Switch to the "Advanced" filter mode.
2. In the "Status" section of the filter configuration, ensure that "Archived" is not selected.
3. Use "Add filter criteria" to add "Cache Owner" condition.  Use "Contains" operator and enter cache owner name as "Search Text".  Make sure to specify owner with an archived cache.
4. Confirm filter configuration and review geocaches shown on map.

Archived caches are shown on map.  They are shown using the standard icon for the particular cache type, not greyed.  Only after opening them they disappear from the map.

Archived caches are not shown.

### Reproducible

Yes

### c:geo Version

2022.10.17

### System information

_No response_

### Additional Information

This may be related to the behaviour I'm seeing on the geocaching.com site map.  When looking at the map on the site with the default filter options, I do not see any archived caches.  However, after adding value for the "Hidden by" filter condition, archived caches are shown.  I wonder if cache "archived" status is included in the search results c:geo gets from the GC site and simply ignores it, assuming archived caches are never returned, or if that information is not part of the search results and hence loading of all geocache data is the only way to determine its archived status.

fm-sys:
Correctly detect archived caches in search results. 
Though, I'm still wondering why this didn't come up before...?

### DIFF
--- a/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
+++ b/main/src/cgeo/geocaching/connector/gc/GCWebAPI.java
@@ -918,6 +918,7 @@ class GCWebAPI {
 
                 c.setFavoritePoints(r.favoritePoints);
                 c.setDisabled(r.cacheStatus == 1);
+                c.setArchived(r.cacheStatus == 2);
                 if (r.owner != null) {
                     c.setOwnerDisplayName(r.owner.username);
                     c.setOwnerUserId(r.owner.username);


fm-sys:
retest this please

PR #13762, opened 2022-12-18
Add 'select next 100' option for cache lists

moving-bits:
## Description
Adds an "select next 100" option to cache list menu in select mode


moving-bits:
CAPTCHA_REQUIRED during tests, so

retest this please

moving-bits:
Reported error is unrelated.

@mucek4 
Can you have a look at the CI and try to fix the CAPTCHA_REQUIRED on the server?
Seems to be a case of https://manual.cgeo.org/en/loginproblems#login_failedcaptcha_required

PR #13766, opened 2022-12-19
Add parsing of flopp.net GPX files as tracks (fix #13661)

### LINKED ISSUE #13661: Use GPX file generated by flopp.net as route or track
A user on support mail used flopp.net to create a simple GPX with 20 waypoints (attached for reference). He would like to import it into c:geo either as route or maybe as a bunch of waypoints (=caches in c:geo).

It does however not work with any of our functions.
Is there a possibility we can process the file, or does the file need changes to be compatible?

Support ticket 540049

[monopoly_wilhelmshaven.zip](https://github.com/cgeo/cgeo/files/10001699/monopoly_wilhelmshaven.zip)

moving-bits:
## Description
Fix failing GPX import of flopp.net generated GPX files


moving-bits:
Failing tests are unrelated

Issue #13761, opened 2022-12-17
Running instrumented c:geo Tests in Android Studio / strange test & cgeo project structure


eddiemuc:
As of today, I realized that I am no longer able to run instrumented Tests from Android Studio:
* the instrumented test folder "src-android" is not recognized by AS as a test folder
* if I mark it accordingly by myself, it does not found necessary base classes to compile the tests

Running tests via gradle seems to work, but a pure command line output is really not useful in complex debug/development scenarios. Also I don't know how to run only single tests via command line.

While looking into this, two other things became clear:
* Our instrumented tests seems to be written using very very very old JUnit versions! For example they don't even use @Test annotation but still the convention of starting test method names with prefix "test"
* The whole folder structure of the project does not seem to fit to current Android app standards 

I can only speculate that the test execution problems have something to do with the strange and outdated cgeo project.

Folder structure should be something like this (Updated 19.12.22):
```
<root>
  app
    src
      main 
        java -> app sources (java)
        res --> ressources (for main)
      test   -> unit test sources
      androidTest  -> instrumented unit test sources
  other-module
    src
      main  --> module sources
      test
...
```

Folder structure currently is:
```
<root>
  main
    src --> app sources (only java)
    res --> ressources
  other-module
    src --> module sources
  tests
    src --> unit test sources
    src-android --> instrumented unit test sources
```

Questions
* Why do we have this strange structure?
* Would it be beneficial to "clean this up"?

ztNFny:
I opened https://github.com/cgeo/cgeo/issues/13494 for the same or at least a very similar topic recently. Got told in one of our calls that it's due to the way the CI pipeline is set up.
The way it's "supposed to be" is that src-android is only run by CI, not by gradle on a normal build as the tests in there connect against the services, using a user hardcoded in CI pipeline and potentially failing with other users.

So .. "legacy stuff"

eddiemuc:
Note for self (in case a "big move" is necessary in git):
https://vjeko.com/2020/11/24/understanding-renaming-moving-files-with-git/

eddiemuc:
> I opened https://github.com/cgeo/cgeo/issues/13494 for the same or at least a very similar topic recently. Got told in one of our calls that it's due to the way the CI pipeline is set up.
> The way it's "supposed to be" is that src-android is only run by CI, not by gradle on a normal build as the tests in there connect against the services, using a user hardcoded in CI pipeline and potentially failing with other users.
> 
> So .. "legacy stuff"

[...]

PR #13768, opened 2022-12-20
rel to #13761: refactor main source directory structure

eddiemuc:
rel to #13761: refactor main source directory structure

**DO NOT SQUASH THE COMMITS IN THIS PR ON MERGE!!!!! USE "REBASE AND MERGE"!!!**

PR #13769, opened 2022-12-21
Rel to #13761: Refactor/reorganize cgeo code structure 

eddiemuc:
rel to #13761:  Refactor/reorganize cgeo code structure

**DO NOT SQUASH THE COMMITS IN THIS PR ON MERGE!!!!! USE "REBASE AND MERGE"!!!**


eddiemuc:
Retest this please

eddiemuc:
I am puzzled. Although I changed the Jenkinsfile to new paths...

image

...those paths doesn't seem to be used in Jenkins:

image

If the Jenkinsfile is not used, why are we maintaining any at all? Shouldn't we then rather delete it from repo?

I can't resolve this on my side, no access to Jenkins. I will wait a few hours for reaction, and if none is given I will revert the name change of "main" to "app" (I can then do this in a later PR). @Lineflyer , @bekuno, @mucek4 ?

bekuno:
> I can't resolve this on my side, no access to Jenkins. I will wait a few hours for reaction, and if none is given I will revert the name change of "main" to "app" (I can then do this in a later PR). @Lineflyer , @bekuno, @mucek4 ?

Now you should have login rights to CI.

The jenkins script is not in use, and was not as long as I know. Don´t know why. Maybe it was not finished.
At CI are some dedicated scripts and direct scripts configured, and there are filters for results.

For now I would therefore say, that it is better to stay with the old name and do the renaming in a single step later.

eddiemuc:
> For now I would therefore say, that it is better to stay with the old name and do the renaming in a single step later.

Agree. I will then switch "back" to the other PR :-)

moving-bits:
Thanks @eddiemuc for diving into this issue. I'm in no way capable of doing an actual review here, as my experience with Android programming is coming from c:geo only. Nevertheless I did a quick test of this PR in my environment:
- Program compiles fine
- no problems detected during a short run of the program
- the following files remain in the source tree and need to be (re)moved manually:
  - main/cgeo.geocaching_preferences.xml
  - main/res/values/keys.xml
  - main/build

  Those files are not part of git, but we should leave a notice for the devs somewhere.
- the problem described in https://github.com/cgeo/cgeo/discussions/13606 is _not_ resolved
- "Project source files" view in AS is broken:
  image
  ("main" contains the left over files, should probably be removed; "app" is missing)
  "Project files" view does work, though, for both app's sources and tests, so it's maybe just a need to work a bit differently.
- tests do run, but 9 tests fail (although logged in as PM)

Another question springs into mind: If we implement this PR on `master`, will it co-exist with the different structure in `release`?

BTW: I did a short test locally, squashing the two commits into one (`git reset HEAD~1`, `git commit -a --amend` etc.), and retrieving history in AS (with right click on selected source, git, show history for selection) still did work.

eddiemuc:
@moving-bits thanks for your feedback, I will answer in PR #13768 and abandon this PR. Reason: CI seems to have larger problems with its scripts when renaming "main" to "app". Thus I leave folder at "main" and continue with the other PR.

PR #13774, opened 2022-12-22
rel to #13761: remove common source hack for contact addon

eddiemuc:
rel to #13761: remove common source hack for contact addon

This PR removes the hack of introducing a common source dir for interface IContacts between cgeo main and cgeo-contact-addon. 

If classes really need to be shared between modules, then a new separate gradle module shall be created. This was not done here, instead a hack was used to share a source dir outside of a module's structure. In the current case, however, instead of creating a shared module it was easier to just remove the single IContacts class, because it was only used to share a small number of String constants.


eddiemuc:
failing test is unrelated.

PR #13789, opened 2022-12-25
UnifiedMap: Update TrackLayer after adding track/route elements

moving-bits:
## Description
Fixes an update problem with route/track segments on UnifiedMap

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgePositionLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgePositionLayer.java
@@ -155,6 +155,7 @@ class MapsforgePositionLayer extends AbstractPositionLayer<GeoPoint> {
             final LineDrawable path = new LineDrawable(segment, isTrack ? trackStyle : routeStyle);
             trackLayer.add(path);
         });
+        trackLayer.update();
     }
 
 }


devemux86:
@moving-bits does it fix https://github.com/mapsforge/vtm/discussions/998 or the issue still exists?

moving-bits:
> @moving-bits does it fix [mapsforge/vtm#998](https://github.com/mapsforge/vtm/discussions/998) 

No, sadly, but from our discussion I took this as a prerequisite, so I want to add it to the code to not forget it later.
I'm still searching for workarounds to make it work reliably (as time permits).

VTM is still pretty fragile in our complex c:geo setting (eg: currently online maps do not work for me, although they have worked for quite some time already), maybe this is due to our specific implementation, maybe it's caused by using a lot of VTM features in one place, probably both. - Maybe I need to take a different approach to `UnifiedMap`, building a VTM-only implementation first, make sure it's stable, and only then extend it by GM map types and the abstraction layer necessary for this... 🤔

devemux86:
> I took this as a prerequisite

That's correct (see [VectorLayerActivity](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/VectorLayerActivity.java#L87)).

> Maybe I need to take a different approach to UnifiedMap, building a VTM-only implementation first, make sure it's stable, and only then extend it by GM map types and the abstraction layer necessary for this...

:+1: 

My guess would be if it's a thread issue? (especially since we can't see it in VTM samples)

As mentioned in our discussion, I have also seen something strange in Cruiser: https://github.com/mapsforge/vtm/discussions/998#discussioncomment-4493986.
When there is some free time, I would like to review VTM overlays and threads (if this is the cause).

Issue #11029, opened 2021-06-28
[API >= 29] don't ask for general storage access permission


fm-sys:
BTW, if using SAF, asking for grading general storage access permission is not necessary anymore in our wizard for API >= 29 (Android 10) 

In fact,  the whole idea of SAF is that users don't have to grant this permission anymore, but only need to grant permission to these folders the app really needs to access...

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11018#issuecomment-869133639_

fm-sys:
This should be done as soon as we have switched to target API 30...

Lineflyer:
But it can not only be removed for API >=29 but for all...or not?

fm-sys:
This whole SAF thing is relatively new, so it could be that older devices crash if they don't have this permission. In other apps I had seen that they check the API level programmly and I have veryfied myself that it works without the permission, but for sure we could check if it works without the permission on older devices as well... The storage access permission is a dynamic permission, which needs to be asked at runtime. And the app can decide, if at all or when it is the right time to do so.

Lineflyer:
Understood. Lets keep it as a safety net then for some time I would say.

BTW: The wizard without this permission should be tested again. At the moment SAF will always find/suggest the /cgeo/ folders with their default content as it is AFAIK created using the old permission before the SAF request is triggered.

Lineflyer:
> Understood. Lets keep it as a safety net then for some time I would say.

Addon after thinking about it again:
If such failures are still out there, users will be on legacy dirs and the wizard showing up on each start. Maybe we should really completely remove the legacy permission to enforce that such problems reach our support. Otherwise we might only delay the problem.

fm-sys:
Should be fixed by https://github.com/cgeo/cgeo/issues/13798, right?

PR #13802, opened 2022-12-30
fix #13798: allow usage of cgeo without location and storage permission

### LINKED ISSUE #13798: Allow usage of c:geo without permission to access fine GPS location
# Current Status

c:geo is currently implemented such that it can't be used when accessing fine GPS location permission is not granted. As long as it is not granted, the user will be caught in a loop of popup dialogs asking him/her to grant this permission.

# Problem statement

I question this design decision. I argue that c:geo can be used meaningfully also without location information, e.g. to search caches, store offline caches and use map viewer capabilities. 

Fro a technical perspective, this "asking" for permission spoils many parts of the code because asking for this permission is implemented in almost all activities.

Last but not least, the current implementation is an actual showstopper for GUI tests using Espresso. This is because the c:geo emulator instance on CI naturally does not have the fine location permission, and c:geo activities are implemented in such a way that they will hang during onResume method in an endless loop when asking for such permission over and over again. So on CI Espresso can never get Activities into a RESUME state - which is the prerequirement for GUI tests. See #13761

# Proposed solution

I propose NOT to make the permission mandatory. Instead, if user does not give fine location permission then the GPS sensor implementation should output a fake coordinate. I propose N 45° 17.460 W 122° 24.800 (I guess you know why :-))

c:geo can then reask fine location permission once on every c:geo startup. I think that's enough.

eddiemuc:
fix #13798: allow usage of cgeo without location and storage permission


eddiemuc:
This PR contains major changes to c:geo permission management:
* It removes the necessity for the user to grant either FINE_LOCATION permission nor WRITE_EXTERNAL_STORAGE permission , **both** is optional now. This makes a lot of sense since most parts of c:geo can also be used without these two permissions. There should not be any exceptions any more in these cases.
* If user does not grant LOCATION permission (or only coarse precision), an approbiate message is shown in Location Status on Main Activity. Clicking on this message leads the user to the Permission Grant dialog for LOCATION permission
* If user does not grant any LOCATION permission (not even coarse), then c:geo assumes the user is in Paris (at Notre Dame). This is not new, it was in the code all along and I kept it.
* If user does not grant WRITE_EXTERNAL_STORAGE permission, he/she is asked once on every c:geo startup to grant this permission. However, it is not enforced.  
* On the technical side: whole permission management was moved to new PermissionAction/PermissionContext class. Some old classes around permission management were removed, especially PermissionHandler

Some screenshots:

Status bar when NO Location permission is granted:

image


Status bar when only coarse location permission is granted:

image

Dialog flow when clicking on location status bar in this state:

image

Clicking "OK" opens the system's permission dialog, e.g.:

image

If anything is chosen resulting in something OTHER than c:geo getting FINE LOCATION access, then the following dialog opens:

image

In this last dialog, the three options are:
* Ask again: c:geo asks for location permission again
* Cancel: cancels the permisison request, everything stays as is
* Open Application Info: opens the system settings for c:geo, user may configure permissions there directly. This is especially important in cases where the Android system itself decides that the permission selection dialog should not be presented to the user (this happens sometimes, to my udnerstanding when user has requested it "too often". Happens a lot during testing :-))

Following is the corresponding dialog for Write storage access (dialog flow is the same as for location, just the explanation text and title is different):

image

Since permissions should be optional, I also build in a "Skip" button into the corresponding pages of the InstallWizard. The "Skip" button only appears after the user was asked once by the system to grant permissions (localization and storage). Example Storage:

image

Tapping "Next" leads to:

image

Tapping "Don't allow" leads to:

image

Tapping "Cancel" leads back to the wizard, but now there is a SKIP button available:

image

Lineflyer:
Interesting and indeed meaningful to handle permissions better in c:geo.
One question:
For what function would we need to "write external storage" permission? AFAIK we do use storage framework for storing reading data or the app based locations accessible without user permission. Any legacy part remaining here?

eddiemuc:
> For what function would we need to "write external storage" permission? AFAIK we do use storage framework for storing reading data or the app based locations accessible without user permission. Any legacy part remaining here?

I have no idea why c:geo would still need this permission. Current version of c:geo enforces it and claims that x:geo can't work without it, so I didn't want to remove it. But during several tests without this permission (e.g doing a backup, downloading a map, storing new own images. Accessing SD card) I didn't find an problems. However I just tested in an Android 12 emulator.

If this permission is indeed no longer needed, we can easily remove it. Anyone having an idea where this is needed?

@moving-bits in the code I found this permission being enforced in DownloadUtilities, was there any specific reason for this?

eddiemuc:
Update: did some tests on an SDK23 emulator w/o Storage permission, seems to work there as well.

Lineflyer:
> If this permission is indeed no longer needed, we can easily remove it. Anyone having an idea where this is needed?

The only reason I found is mentioned in #11029:
Users which still having pending migration to SAF....but IMHO thats no relevant case to consider anymore.

moving-bits:
This looks like a very reasonable and well thought out change, thanks for diving into it!

> @moving-bits in the code I found this permission being enforced in DownloadUtilities, was there any specific reason for this?

I wanted to make sure c:geo has access to the target folder before initiating a download.
(Not knowing enough about SAF handling on older devices I did sort of a copy&paste for that. Even better if we can get rid of the whole "access write permission" code, though - easier for code maintenance, and easier for setting up new users.)

eddiemuc:
If I don't receive an protest, I will remove in this PR the write_external_storage completely, then merge to master. We can then find out in nightlies if removal of this permission causes any problems

Lineflyer:
Go for it. I will try to test nightlies incl. installation on a few (also lower) API as time permits.

fm-sys:
> Open Application Info: opens the system settings for c:geo, user may configure permissions there directly. This is especially important in cases where the Android system itself decides that the permission selection dialog should not be presented to the user

Maybe we could detect this case automatically and directly open "app info"? Seems like a common implementation in many other apps...

main/src/main/java/cgeo/geocaching/InstallWizardActivity.java
```diff
@@ -175,14 +170,10 @@ private void updateDialog() {
...
-                title.setText(R.string.wizard_status_storage_permission);
-                text.setText(R.string.storage_permission_request_explanation);
-                setNavigation(this::gotoPrevious, 0, null, 0, this::requestStorage, 0);
-                break;
             case WIZARD_PERMISSIONS_LOCATION:
-                title.setText(R.string.wizard_status_location_permission);
-                text.setText(R.string.location_permission_request_explanation);
+                title.setText(PermissionContext.LOCATION.getExplanationTitle());
+                PermissionContext.LOCATION.getExplanation().applyTo(text);
+                //text.setText(R.string.location_permission_request_explanation);
```
fm-sys, line 176: A leftover?

eddiemuc:
> > Open Application Info: opens the system settings for c:geo, user may configure permissions there directly. This is especially important in cases where the Android system itself decides that the permission selection dialog should not be presented to the user
> 
> Maybe we could detect this case automatically and directly open "app info"? Seems like a common implementation in many other apps... 

If this is possible we could do it. But I wouldn't know how to detect this case. Do you know any documentation or open source example?

fm-sys:
You could measure the time until the permission gets denied. If you get the request result in let's say less than 100ms, it was most likely denied automatically by the OS.

eddiemuc:
> You could measure the time until the permission gets denied. If you get the request result in let's say less than 100ms, it was most likely denied automatically by the OS.

For my taste that would be a hackish solution. Are you sure that those "other apps" not simply use the "open application setting" solution always (instead of implementing a real dynamic permission handling like c:geo does)? For c:geo this is just a worst-case-fallback, it should (hopefully) not afect users in real-life. (in real-life most users will probably allow permissions to c:geo the first time they are asked for it)

PR #13806, opened 2023-01-01
rel to #13793: support display of geojson for mapsforge

### LINKED ISSUE #13793: Support Display of GeoJson data
Sometimes I have caches where it would be helpful to draw geometric forms on the map (lines, polygons, circles). An example for such a cache is the Heinzel Challenge around Munich (https://coord.info/GC2YYGC)

My proposal for this is to support the import and display of geojson files. 
* GeoJson is described here: https://geojson.org/
* An online GeoJson editor can be found e.g. here: https://geojson.io 
* I would also include support for basic formatting e.g. color, line width

I am not sure, but probably mapsforge and/or VTM already support GeoJson anyway?

GUI-wise I propose something like the integration of individual route files, although there should be the possibility to import multiple GeoJson files and turn their display on and off.

eddiemuc:
rel to #13793: support display of geojson for mapsforge

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -337,6 +337,9 @@ dependencies {
     // Jsoup HTML parsing
     implementation 'org.jsoup:jsoup:1.14.3'
 
+     // GeoJson parsing: https://github.com/cocoahero/android-geojson
+    implementation 'com.cocoahero.android:geojson:1.0.1@jar'
+
     // Junit only needed for local unit tests
     testImplementation 'junit:junit:4.13.2'
 
--- a/main/src/main/java/cgeo/geocaching/files/GPXTrackOrRouteImporter.java
+++ b/main/src/main/java/cgeo/geocaching/files/GPXTrackOrRouteImporter.java
@@ -1,9 +1,13 @@
 package cgeo.geocaching.files;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.location.GeoObject;
+import cgeo.geocaching.location.GeoObjectList;
+import cgeo.geocaching.location.IGeoDataProvider;
 import cgeo.geocaching.models.Route;
 import cgeo.geocaching.storage.ContentStorage;
 import cgeo.geocaching.utils.AndroidRxUtils;
+import cgeo.geocaching.utils.GeoJsonUtils;
 import cgeo.geocaching.utils.Log;
 
 import android.content.Context;
@@ -14,11 +18,14 @@ import androidx.annotation.NonNull;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 import org.apache.commons.io.IOUtils;
+import org.json.JSONException;
 
 public class GPXTrackOrRouteImporter {
 
@@ -29,8 +36,8 @@ public class GPXTrackOrRouteImporter {
         final AtomicBoolean success = new AtomicBoolean(false);
         AndroidRxUtils.andThenOnUi(Schedulers.io(), () -> {
             try {
-                final Route value = doInBackground(uri);
-                success.set(null != value && value.getNumSegments() > 0);
+                final IGeoDataProvider value = doInBackground(uri);
+                success.set(null != value && value.hasData());
                 if (success.get()) {
                     AndroidSchedulers.mainThread().createWorker().schedule(() -> {
                         try {
@@ -51,7 +58,7 @@ public class GPXTrackOrRouteImporter {
         });
     }
 
-    private static Route doInBackground(final Uri uri) {
+    private static IGeoDataProvider doInBackground(final Uri uri) {
         try {
             Route route = parse(new GPXTrackParser("http://www.topografix.com/GPX/1/1", "1.1"), uri);
             if (null == route) {
@@ -71,6 +78,9 @@ public class GPXTrackOrRouteImporter {
             }
             if (null != route) {
                 route.calculateNavigationRoute();
+            }
+            if (null == route) {
+                return parseAsGeoJson(uri);
             }
             return route;
         } catch (IOException e) {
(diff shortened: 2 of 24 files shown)


eddiemuc:
This PR as of now already enables geojson support: it simply displays all geojson files (identified by suffix `json`) which are stored in gpx folder. Only on mapsforge for now, not Google. Implementation was easy.

Integrating this into track file Dialog seems to be the hard part.

moving-bits:
Importing routes / tracks is done in `files/GPXTrackOrRouteImportet.java`, calling several different parsers. You could do the distinction depending on file suffix here and call a parser of your choice.

If GeoJSON objects are polygons only, you could even store it as a track/route (with multiple segments, if needed). But I guess there are other object types as well, e. g. circles etc.
But maybe the route object could be extended to not only contain arrays of points, but other objects as well. Maybe even base it on segments (objects) of GeoJSON objects?

eddiemuc:
> But maybe the route object could be extended to not only contain arrays of points, but other objects as well. 

Yes, I am trying to think about a good abstraction for this.

@moving-bits one question: I noticed that the custom mapsforge layer which renders the routes uses a cache inside where already calculated `Path` objects are reused. So e.g. Mercator transformations have to be calculated less often I assume. Instead a mapsforge `Polyline` layer could be used directly, but this one doesn't do an caching. Is this caching crucial e.g. due performance reasons?

moving-bits:
Performance may be one reason, but IIRC there was a problem with `Polyline` that it connects all segments to form a single line.
We allow to load track files containing several segments, and we want to display them as unconnected items. Manual calculation allowed me to differentiate between `moveTo()` and `lineTo()`, see `AbstractRouteLayer.translateRouteToPath()`).

(Similar applies to `UnifiedMap`: I had to use a `VectorLayer` for displaying tracks instead of a `PathLayer` for the very same reason. BTW: No extra caching applied here - but VTM is vector based anyway, not bitmap based.)

eddiemuc:
> > there was a problem with Polyline that it connects all segments to form a single line
> > I had to use a VectorLayer for displaying tracks instead of a PathLayer for the very same reason
> 
> 1 line overlay is expected to render 1 line.
> Can use many line overlays (Polyline, PathLayer) to render multiple lines (i.e. trkseg).
> 
> Or we could extend the current Mapsforge + VTM line layers.
> Pull requests are always welcome! :slightly_smiling_face: 

Mapsforge already has the `GroupLayer` for this. (I assume there is something similar in VTM?). I think this is enough to handle this use case.

devemux86:
> Mapsforge already has the GroupLayer for this. (I assume there is something similar in VTM?).

There is also in VTM, however I have never needed it.
(it is just a List)

It is easier to maintain lists of layers inside the app,
as each app has specific needs.

moving-bits:
From implementing track display in VTM I remember the following problem with the "one layer per segment" approach:
- I do not know the number of segments beforehand, it's a variable number of elements (and thus, layers, with this approach).
- The order of layers implicitly determines the z-order for display.
- `layers.set()` accepts an index number at which to set a layer, but the layer indices must not have "wholes" in it (so I cannot create a "large enough" gap to fill later with the layers required for my variable elements) - also a "large enough gap" approach is pretty risky, you never know which user data will be entered...

This means, that either the layers with the variable elements have to be at the top of the stack (making it most-prominent visually), or I have to remove and rebuild the layer stack on inserting an element potentially requiring multiple layers.

For those reasons I decided to put all tracks with all segments onto the same layer (which is pre-created layer on map startup at the index position for the required z-order), and make sure that I use a layer type which allows me to display unconnected polylines.

eddiemuc:
> This means, that either the layers with the variable elements have to be at the top of the stack (making it most-prominent visually)...

I don't understand. Isn't every segment of the route a "variable" element? And why is the z-order important in which they are displayed?

devemux86:
For an alternative way, see https://github.com/mapsforge/vtm/pull/101 and https://github.com/mapsforge/mapsforge/issues/1116.

- Layers.addGroup(group)
- Layers.addLayer(Layer, group)

moving-bits:
> > This means, that either the layers with the variable elements have to be at the top of the stack (making it most-prominent visually)...
> 
> I don't understand. Isn't every segment of the route a "variable" element? 

We have a variable number of tracks, and while most tracks contain only a single path, eeach track can contain a variable number of unconnected segments.
(A similar problem arises with c:geo's history line, which is only one line, but can contain unconnected segments as well.)

> And why is the z-order important in which they are displayed?

Top-most elements may hide elements on layers with lower z-index, so we decided to put caches top-most, waypoints somewhere below etc.

eddiemuc:
> > And why is the z-order important in which they are displayed?

> Top-most elements may hide elements on layers with lower z-index, so we decided to put caches top-most, waypoints somewhere below etc.

This should be solvable by putting elements with different z-priority in different Group Layers. Those would be permanent with a fixed z-order  but inside a group layer (where everything has same z-prio) one would be able to add/delete stuff w/o worrying about z-order. At least that's my understanding as of now. Am I missing something?

moving-bits:
> > > And why is the z-order important in which they are displayed?
> 
> > Top-most elements may hide elements on layers with lower z-index, so we decided to put caches top-most, waypoints somewhere below etc.
> 
> This should be solvable by putting elements with different z-priority in different Group Layers.

Ah, didn't know that. Thanks for the pointer, will dive into it...

eddiemuc:
> Ah, didn't know that. Thanks for the pointer, will dive into it...

Would.it be ok for you to wait on this? This would for sure clash with my ongoing geojson implementation. I can take a look at the z-order-thing too while I am at it.

(My idea is to treat both geojson and routes/tracks in the same "geo object" layer)

moving-bits:
> > Ah, didn't know that. Thanks for the pointer, will dive into it...
> 
> Would.it be ok for you to wait on this?

Sure, no problem. 

Z-order is fine currently, but I'm looking into https://github.com/mapsforge/vtm/discussions/998, and maybe using `GroupLayer` in combination with `PathLayer` instead of `VectorLayer` may help to solve that issue.

devemux86:
My suggestion would be to check the more powerful layer groups: https://github.com/cgeo/cgeo/pull/13806#issuecomment-1371298335

---

`GroupLayer` is just a `List`, can do a better implementation with your own layer lists.

moving-bits:
> My suggestion would be to check the more powerful layer groups: [#13806 (comment)](https://github.com/cgeo/cgeo/pull/13806#issuecomment-1371298335)
> 
> `GroupLayer` is just a `List`, can do a better implementation with your own layer lists.

Did a very short test for tracks using layer groups, and was no longer able to reproduce the issue described in https://github.com/mapsforge/vtm/discussions/998.

Will now wait for this PR being finalized before restructuring `UnifiedMap` to use layer groups for VTM.

eddiemuc:
> Will now wait for this PR being finalized before restructuring UnifiedMap to use layer groups for VTM.

If you're just modifying Unified map then go ahead (don't wait for this PR).

devemux86:
> layer groups for VTM

Note: the layer groups are also available in Mapsforge.

moving-bits:
> > Will now wait for this PR being finalized before restructuring UnifiedMap to use layer groups for VTM.
> 
> If you're just modifying Unified map then go ahead (don't wait for this PR).

done with #13833

eddiemuc:
This PR, if merged, will enable display of GeoJSON objects for Mapsforge maps only! (Google and UnifiedMap are not yet supported). Under the hood it already abstracts out geoobject-list and route/track into the base interface called IGeoDataProvider

PR #13807, opened 2023-01-02
fix #13761: repair waypoint activity tests

### LINKED ISSUE #13761: Running instrumented c:geo Tests in Android Studio / strange test & cgeo project structure

eddiemuc:
fix #13761: repair waypoint activity tests


eddiemuc:
This PR consolidates all existing still meaningful Espresso Tests for Waypoint Activity in one test class WaypointActivityTest and removes the leftover outdated (and quite complex) Espresso class hierarchy. It concludes issue #13761

PR #13812, opened 2023-01-03
differentiate size "not chosen" and "other" for GC.com

### LINKED ISSUE #13795: Filtering of cache size "not chosen" versus "other"
### Describe your problem!

Reported by a user on support mail:

The filter in the map shows caches with the size "not chosen" even if I only filter on "other".

Many challenges does not equal "not chosen" and "other", so that could cause problems if I would find one of the wrong kind.

### How to reproduce?

For example https://coord.info/GC25KQD, a "not chosen" that still shows up with filter with only "other".
That same cache shows up if I only filter on "not chosen" too.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2022.12.21

### System information

_No response_

### Additional Information

Support ticket 484323

ztNFny:
fixes #13795

not sure whether this should be merged. See https://github.com/cgeo/cgeo/issues/13795#issuecomment-1369584410

### DIFF
--- a/main/src/main/java/cgeo/geocaching/enumerations/CacheSize.java
+++ b/main/src/main/java/cgeo/geocaching/enumerations/CacheSize.java
@@ -24,9 +24,9 @@ public enum CacheSize {
     REGULAR("Regular", 3, R.string.cache_size_regular, "regular", new int[]{3}, "M"),
     LARGE("Large", 4, R.string.cache_size_large, "large", new int[]{4}, "L"),
     VERY_LARGE("Very large", 5, R.string.cache_size_very_large, "xlarge", null, "XL"), // used by OC only
-    NOT_CHOSEN("Not chosen", 6, R.string.cache_size_notchosen, "", null, "-"),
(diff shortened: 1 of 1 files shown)


ztNFny:
retest this please

ztNFny:
@moving-bits can we merge this?

moving-bits:
Did a quick test to understand what this PR does. I guess this is meant for filtering?
If I select "not chosen" I would expect earth caches to be listed (from reading the comment in the source), but those will show up when selecting "other" (like events).

ztNFny:
> If I select "not chosen" I would expect earth caches to be listed (from reading the comment in the source), but those will show up when selecting "other" (like events).

You'll get ECs published until 2012:

> "Not chosen" (size 1) can no longer be selected at GC - it was available for all caches and the only value for EC, virtual, Webcam until 2012. Since then those types are using "Other" (size 6).

moving-bits:
ok, I see. I can understand the issue posted in the referenced issue, OTOH it might break the workflow for other users used to filter for "other" only, now no longer getting all EC etc.

Anyway, let's try this in the nightlies and see what may come up as user feedback.

PR #13818, opened 2023-01-04
declare vtm natives as runtimeonly to solve android studio issues

eddiemuc:
Declare vtm natives as runtimeonly, trying to solve android studio issues

I started to experience issues with my local Android Studio similar to what @moving-bits had describes some weeks ago (one by one classes started to "disappear" from the knowledge of Android Studio, were marked red etc).

I had the impression that this is caused by VTM native libraries being part of compile path. Removing them from this path resolved the issue for me. This is what this PR is doing (->moving it to "runtimeOnly"). VTM seems to work with this change, at least in the emulator.

@moving-bits does this also solve your problems? If yes, maybe we should merge it?

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -376,10 +376,10 @@ dependencies {
     implementation "org.locationtech.jts:jts-core:1.16.1"
 
     implementation "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion"
-    implementation "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-armeabi-v7a"
-    implementation "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-arm64-v8a"
-    implementation "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-x86"
-    implementation "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-x86_64"
+    runtimeOnly "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-armeabi-v7a"
+    runtimeOnly "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-arm64-v8a"
+    runtimeOnly "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-x86"
+    runtimeOnly "$mapsforgeVTMSource:vtm-android:$mapsforgeVTMVersion:natives-x86_64"
 
     implementation "$mapsforgeVTMSource:vtm-http:$mapsforgeVTMVersion"
 


eddiemuc:
@devemux86 I would be interested in your opinion as well. Shouldn't it be enough to include the native VTM libs as "runtimeOnly" dependency?

devemux86:
(IDEA seems to work properly)

See https://github.com/mapsforge/vtm/discussions/959 for another suggestion, does it help?

eddiemuc:
> (IDEA seems to work properly)
> 
> See https://github.com/mapsforge/vtm/discussions/959 for another suggestion, does it help?

Ah this might work too. Have to try it out. Thanks!

devemux86:
I made a test with `runtimeOnly` on Cruiser and it seems to work properly.

moving-bits:
Thanks @eddiemuc and @devemux86 for diving further into this issue. I did a (very short) test during a break... and it seems to work for the issue described in https://github.com/cgeo/cgeo/discussions/13606 (tested for c:geo project only, so far) 👍 👍 👍 

I applied the changes in `build.gradle`, did a `clean project`, then "invalidate caches and restart", and after having reloaded the external libs, all red markings are gone (meaning: all imports could be resolved).

So far I have tested only for a couple of files I had problems with in the past, but as the problem affected different and changing files, I'll keep an eye on it if it resurfaces.

Later this evening I will also check for my new dev laptop as well. Very nice machine, and c:geo is fine over there, but I have problems compiling vtm - AS complains about failing resolution for `android.content.Context` import in dozens of files and stops the build. (So not only failing resolution, and build working otherwise, as in the issue linked above, but completely breaking build. Might be a totally different issue (maybe laptop not yet fully initialized for AS?). I'd started integrating vtm "from scratch" over there, but breaking vtm build currently knocks me out.)

devemux86:
> I have problems compiling vtm - AS complains about failing resolution for android.content.Context import in dozens of files and stops the build

Please see VTM (+ Mapsforge) main page:
_"Building VTM / Mapsforge: set ANDROID_HOME environment variable with the Android SDK installation directory"_

We maintain the VTM / Mapsforge Android modules as Java libraries -> jar (not Android libraries -> aar).
There are no Android resources in these modules and we avoid the aar and their possible complications.
This means that the android.jar path must be declared manually (unless there is an automatic solution).

eddiemuc:
@moving-bits great to hear! I will await further feedback from you for your other machine. If positive then I suggest we merge this PR.

If we find problems later, then we can try the "native libs first" solution too.

moving-bits:
> Please see VTM (+ Mapsforge) main page: _"Building VTM / Mapsforge: set ANDROID_HOME environment variable with the Android SDK installation directory"_

Aaaarrgghhh! I had missed this on setting up my new laptop. Thanks for pointing me to this. 👍 
Now it compiles fine, and `Context` import is no longer marked in red. 👍 

@devemux86 
Maybe gradle should throw a warning when the environment variable is unset?
I'll prepare a PR for it, if ok for you. 
(All four Android-related `build.gradle` files?)

devemux86:
>  I'll prepare a PR for it, if ok for you.

Sure! Should be 3 build.gradle on each library?

moving-bits:
> > I'll prepare a PR for it, if ok for you.
> 
> Sure! Should be 3 build.gradle on each library?

I had looked for references to `ANDROID_HOME` in `build.gradle` files and found four files: vtm-android, vtm-android-example, vtm-android-gdx and vtm-android-mvt

devemux86:
> I had looked for references to ANDROID_HOME in build.gradle files and found four files: vtm-android, vtm-android-example, vtm-android-gdx and vtm-android-mvt

I had looked for android.jar in libraries, bur sure if there more references.

moving-bits:
> > I had looked for references to ANDROID_HOME in build.gradle files and found four files: vtm-android, vtm-android-example, vtm-android-gdx and vtm-android-mvt
> 
> I had looked for android.jar in libraries, bur sure if there are more references.

see https://github.com/mapsforge/vtm/pull/1001

devemux86:
> see https://github.com/mapsforge/vtm/pull/1001

Is it working properly? (with `if` at the end)

Will you also send a pull request in Mapsforge?

moving-bits:
> > see [mapsforge/vtm#1001](https://github.com/mapsforge/vtm/pull/1001)
> 
> Is it working properly? (with `if` at the end)

I'd done a quick test against an undefined env var ("ANDROID_HOME2") and build instantly failed. On changing back to the (meanwhile defined) var "ANDROID_HOME" build proceeded as usual.

> Will you also send a pull request in Mapsforge?

yes, will do

PR #13831, opened 2023-01-07
Update bug_report.yml with current version number

bekuno:
Set the current c:geo version number as templete.

Maybe this should be integrated as a step in the release workflow.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -76,7 +76,7 @@ body:
     attributes:
       label: c:geo Version
       description: What version of c:geo are you using? You will find it in c:geo Menu -> About c:geo.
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

bekuno:
retest this please

bekuno:
retest this please

SammysHP:
I would go the other way: Replace it with a YYYY.MM.DD placeholder. Having it always up-to-date gives us false information if the reporter hasn't answered the question.

bekuno:
@Lineflyer What do you think?

Lineflyer (changes requested):
Sorry, did not notice this PR.
I agree to @SammysHP to use a generic placeholder to avoid misinterpretation and at the same time avoid the necessity to regularly update the template.

Lineflyer (approved):

PR #13867, opened 2023-01-14
fix #13838: allow var sync also from waypoint notes, enhance sync algorithm

### LINKED ISSUE #13838: Changing variable values in personal cache note does not update "variables" tab
### Describe your problem!

When you write things like "A=5" in the personal cache note, and then go to the "variables" tab, then c:geo automatically created a new variable called "A" and gives it the value "5".

However, if you then edit the personal cache note to, say, "A=7" then the variable in the "variables" tab keeps the old value "5".

This can lead to false calculations of the next stage, final, etc. and these errors are very hard to spot (you have to compare every variable in the PCN with the variables in the "variables" tab one by one). 

### How to reproduce?

See above.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2022.12.21

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #13838: allow var sync also from waypoint notes, enhance sync algorithm

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/connector/gc/GCParserTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/connector/gc/GCParserTest.java
@@ -208,7 +208,7 @@ public class GCParserTest {
     private void assertWaypointsFromNote(final Geocache cache, final Geopoint[] expected, final String note) {
         cache.setPersonalNote(note);
         cache.setWaypoints(new ArrayList<>(), false);
-        cache.addWaypointsFromNote();
+        cache.addCacheArtefactsFromNotes();
         final List<Waypoint> waypoints = cache.getWaypoints();
         assertThat(waypoints).hasSize(expected.length);
         for (int i = 0; i < expected.length; i++) {
@@ -250,7 +250,7 @@ public class GCParserTest {
         cache.setWaypoints(new ArrayList<>(), false);
         cache.setPersonalNote("\"Parking area at PARKING=N 50° 40.666E 006° 58.222\n" + "My calculated final coordinates: FINAL=N 50° 40.777E 006° 58.111\n" + "Get some ice cream at N 50° 40.555E 006° 58.000\"");
 
-        cache.addWaypointsFromNote();
+        cache.addCacheArtefactsFromNotes();
         final List<Waypoint> waypoints = cache.getWaypoints();
 
         assertThat(waypoints).hasSize(3);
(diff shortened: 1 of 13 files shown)


eddiemuc:
PR will:
1. sync variables and waypoints also from waypoint user notes (upon waypoint saving)

image

2. enhance algorithm: if last variable value in note was identical with value in var list and was changed, then var value is overwritten

3. if variable values in notes are out-of-sync with var values in notes (cache or waypoints) then a new button appears under cache notes allowing user to overwrite values in var list with those from note

image

image

moving-bits:
Thanks for adding this nice feature. 

I played with it a bit (works nicely), but would like to see an additional feature:
If you have a variable in a waypoint user note, which is out of sync, and you close the user note, it would be great to get a notification (popup?) for this to be able to get vars back in sync.

- Added a waypoint to a trad. cache
- Opened this waypoint from map, entered A=1, B=3 into separate lines of user note => Variables got created, fine.
- Opened waypoint from map again, changed A to 5 => Variable in variables list got changed, fine.
- Changed value of A in variables view to 7, opened waypoint from map again, changed A to 6 => Variable in variables list did not change (ok), but you do not get a notification about vars being out of sync.

I understand that you do not want to change the user note text (IIRC), but on receiving a notification the user could decide between change value in cache-global variables list, change value in user note manually, or ignore the difference.

eddiemuc:
Thanks for the feedback

> ...but would like to see an additional feature: If you have a variable in a waypoint user note, which is out of sync, and you close the user note, it would be great to get a notification (popup?) for this to be able to get vars back in sync.
> ...
> I understand that you do not want to change the user note text (IIRC), but on receiving a notification the user could decide between change value in cache-global variables list, change value in user note manually, or ignore the difference.

Potential problem I see here: when the user decides NOT to sync, then he/she will get this popup over and over again whenever the waypoint is edited. This could be a real pain. 

On the other hand, if we decide that note values should **always** override the var list then we don't need a popup at all.

I personally would not change the user note text, but this is a group decision not my personal one.

Would it be ok to handle this in a separate, new issue (not in this PR)?

moving-bits:
> Potential problem I see here: when the user decides NOT to sync, then he/she will get this popup over and over again whenever the waypoint is edited. This could be a real pain.

understandably. So maybe just add an "out of sync" button (like you have done for cache note), if vars are out of sync. Less intrusive, and user can decide how to react. But I would place the button above the input field in this case, as it may be hidden below the keyboard otherwise.

> Would it be ok to handle this in a separate, new issue (not in this PR)?

sure

eddiemuc:
> So maybe just add an "out of sync" button (like you have done for cache note), if vars are out of sync. Less intrusive, and user can decide how to react.

Thought about this myself. But there's a big difference in waypoint edit view when compared to cache personal note: the screen is the same in which the actual text-typing for the user note is happening. So unless we implement a "live check" function continuously checkng whether the just-typed-text would (if saved) create an asyncronity in var list (and this is quite complex), then the button would not appear "in-time" but only after the "edit waypoint" view is closed/saved and then reopened. Also the other way around: if there's an asynchronity and I correct it by changing the text, then button would stay unless i closed/save and reopen the waypoint edit view. This seemed too complex to me for the problem at hand, so I stayed with the button at cache user note.

moving-bits:
I transferred the discussion about user note "out of sync"-button into a separate issue, see #13878, so that we can merge here without burying this discussion.

PR #13877, opened 2023-01-15
Add route calculation for prepending points (fix #13849)

### LINKED ISSUE #13849: "Prepend to route" does not calculate new segment
Notes from today's dev call:

When adding a new waypoint to an existing route using "prepend to route", a straight line is drawn between the existing route and the new element. Instead, this new segment should be calculated as well (which it will be after closing and re-opening the map.)

"Append to route" does work as expected, though.

moving-bits:
## Description
Until now only the last segment got recalculated on adding a segment. This PR adapts this behavior if new segment got prepended.


moving-bits:
failing tests are unrelated, but...

retest this please

moving-bits:
retest this please

moving-bits:
Failing tests are unrelated, so let's merge

PR #13879, opened 2023-01-15
fix #13793: add geojson support for google maps

### LINKED ISSUE #13793: Support Display of GeoJson data

eddiemuc:
fix #13793: add geojson support for google maps

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/GeoObject.java
+++ b/main/src/main/java/cgeo/geocaching/location/GeoObject.java
@@ -1,10 +1,10 @@
 package cgeo.geocaching.location;
 
-import android.graphics.Color;
 import android.os.Parcel;
 import android.os.Parcelable;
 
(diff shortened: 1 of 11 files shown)


eddiemuc:
Outdated PR, can be deleted

Issue #13796, opened 2022-12-28
Make membership state more transparent


Lineflyer:
### Tell us your idea!

We had a user on support mail, who was missing the pocket query menus in c:geo.
It turned out, that c:geo failed to recognize him as premium member. This might occur on login failure (e.g. missing network coverage) or directly after obtaining a premium membership but before next login with c:geo.

While the fix (= restart c:geo or relogin) is easy, the user made a proposal for a potential improvement in such situation:

### What solution would you suggest?

User suggested:
If a basic member tries to use the pocket queries function, you should generate a message that says "premium members only feature".  Also, you should display the membership level on the login (home) screen.

### Additional Information

Support ticket 289067

fm-sys:
> If a basic member tries to use the pocket queries function, you should generate a message that says "premium members only feature". 

Currently we hide PM only feature totally to do not advertise those. This can be discussed of course...

Lineflyer:
IMHO this is sufficiently solved by showing the membership state on the home screen.

PR #13880, opened 2023-01-15
Display GC member state on home screen (rel. to #13796)

moving-bits:
## Description
Displays either "(Basic)" or "(Premium)" after connector status info on home screen (for GCConnector only):

image

Not sure if we should make those strings translatable or not. How does GC handle this?


bekuno:
I checked

bekuno:
I checked on the website by changing the language - all time the state text was "Premium Member".
So I vote for not translating.

Lineflyer:
Well, I think it should be translated of course. 
Beside other languages (where the wording geocachers common use might not be the English ones like in Germany) also the geocaching website uses proper german:

For PM:
image

And for BM:
image

fm-sys:
IMHO the member state isn't related to the login result.

So I would vote for following syntax:
```
GC · Login OK · Basic Member
```

Or even put it in the next line?

moving-bits:
I've switched from `Login OK (Basic)` to `Login OK · Basic`, let's see how it's works in real life

PR #13911, opened 2023-01-29
rel to #13793: refactor google maps drawing object handling

eddiemuc:
rel to #13793: refactor google maps drawing object handling

This PR reworks almost completely the way in which c:geo renders drwaings onto Google Map. Intention is to fix previous bugs and performance problems due to incorrect concurrency handling in the asynchronous google object drawer:
* use a **single** command queue for add/remove commands (instead of two different ones) to avoid problems when add/remove is used for same objects repeatedly (which unfortunately happens often in current implementation)
* optimize number of objects removed + drawn anew by making difference checks already on add/remove commands
* allow more effective replacement of objects
* document code
* add unit testing

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleCacheOverlayItem.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleCacheOverlayItem.java
@@ -72,9 +72,9 @@ public class GoogleCacheOverlayItem implements CachesOverlayItemImpl, MapObjectO
                     .radius(GoogleCachesList.CIRCLE_RADIUS)
                     .zIndex(GoogleCachesList.ZINDEX_CIRCLE);
 
-            return new MapObjectOptions[]{MapObjectOptions.from(marker), MapObjectOptions.from(circle)};
+            return new MapObjectOptions[]{MapObjectOptions.marker(marker), MapObjectOptions.circle(circle)};
         } else {
-            return new MapObjectOptions[]{MapObjectOptions.from(marker)};
+            return new MapObjectOptions[]{MapObjectOptions.marker(marker)};
         }
     }
 
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleCachesList.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleCachesList.java
@@ -46,15 +46,15 @@ public class GoogleCachesList {
         }
         if (this.options == null) {
             this.options = options;
-            mapObjects.requestAdd(this.options);
+            mapObjects.add(this.options);
         } else {
             final Collection<MapObjectOptions> toRemove = diff(this.options, options);
             final Collection<MapObjectOptions> toAdd = toRemove.size() == this.options.size() ? options : diff(options, this.options);
 //            Log.i("From original " + this.options.size()  + " items will be " + toAdd.size() + " added and " + toRemove.size() + " removed to match new count " + options.size());
             this.options = options;
 
-            mapObjects.requestRemove(toRemove);
-            mapObjects.requestAdd(toAdd);
+            mapObjects.remove(toRemove);
+            mapObjects.add(toAdd);
         }
     }
 
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapObjects.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapObjects.java
@@ -1,59 +1,23 @@
 package cgeo.geocaching.maps.google.v2;
 
-import java.util.ArrayList;
-import java.util.Collection;
-
 import com.google.android.gms.maps.GoogleMap;
-import com.google.android.gms.maps.model.CircleOptions;
-import com.google.android.gms.maps.model.MarkerOptions;
-import com.google.android.gms.maps.model.PolygonOptions;
-import com.google.android.gms.maps.model.PolylineOptions;
 
 /**
- * class to wrap GoogleMapObjectsQueue, able to draw individually map objects and to remove all previously
- * drawn
+ * class to wrap GoogleMapObjectsQueue
  */
 public class GoogleMapObjects {
 
     private final GoogleMapObjectsQueue queue;
-    /**
-     * list of object options to be drawn to google map
-     */
-    private final Collection<MapObjectOptions> objects = new ArrayList<>();
 
     public GoogleMapObjects(final GoogleMap googleMap) {
         queue = new GoogleMapObjectsQueue(googleMap);
     }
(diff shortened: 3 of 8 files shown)


Lineflyer:
Would you like to see this in the next beta or shall we merge this only after branches have been merged (to allow more nightly testing first)?

eddiemuc:
For this one I would prefer a nightly test phase before going to beta. It is a pretty big design change

eddiemuc:
I added a "do not merge" Label for now. Merge only after beta is done


Note to self: when two caches of same type etc have same coordinates -> will they be displayed as two separate markers?

moving-bits:
Would be nice to have something similar for `UnifiedMap`... We do have some sort of object (marker) cache there already, but no queueing, no removal of elements etc. - Maybe it could be even generalized for all of our map types?

eddiemuc:
> Would be nice to have something similar for `UnifiedMap`... We do have some sort of object (marker) cache there already, but no queueing, no removal of elements etc. - Maybe it could be even generalized for all of our map types?

As of now it seems UnifiedMap uses the same c:geo-google-abstractions as "old" implementatio (including queueing mechanism which I refactored)

eddiemuc:
...or I overlooked something in code

moving-bits:
> As of now it seems UnifiedMap uses the same c:geo-google-abstractions as "old" implementatio (including queueing mechanism which I refactored)

It doesn't use `GoogleMapObjectsQueue` so far (and probably shouldn't, as it's GMv2). I'm looking for a queueing and marker caching mechanism (abstraction) which covers both UnifiedMap Google and VTM implementations. (Maybe it can even be abstracted to cover all four map types?)

eddiemuc:
> It doesn't use `GoogleMapObjectsQueue` so far (and probably shouldn't, as it's GMv2). 

Looking into class `cgeo.geocaching.unifiedmap.googlemaps.GoogleMapsPositionLayer`, I see that it uses class `cgeo.geocaching.maps.google.v2.GoogleMapObjects` to display things. This class in turn relies on `GoogleMapObjectsQueue`.

I concluded from this that UnifiedMap uses same mechanisms as "old" map for Google. Is this wrong?

eddiemuc:
> I'm looking for a queueing and marker caching mechanism (abstraction) which covers both UnifiedMap Google and VTM implementations. (Maybe it can even be abstracted to cover all four map types?)

One step after the other :-)

moving-bits:
> > It doesn't use `GoogleMapObjectsQueue` so far (and probably shouldn't, as it's GMv2).
> 
> Looking into class `cgeo.geocaching.unifiedmap.googlemaps.GoogleMapsPositionLayer`, I see that it uses class `cgeo.geocaching.maps.google.v2.GoogleMapObjects` to display things. This class in turn relies on `GoogleMapObjectsQueue`.
> 
> I concluded from this that UnifiedMap uses same mechanisms as "old" map for Google. Is this wrong?

Ah, partly a misunderstanding (on my side): `UnifiedMap` uses `GoogleMapObjectsQueue` for drawing lines and circles (sort of a left-over, not migrated yet), but not for displaying caches. 
I was thinking of `GeoItemCache` in `UnifiedMap` and whether it could be fused with the work you are doing here for GMv2.

eddiemuc:
> I'm looking for a queueing and marker caching mechanism (abstraction) which covers both UnifiedMap Google and VTM implementations. (Maybe it can even be abstracted to cover all four map types?)

@moving-bits I thought about this some more and I would be read to give this a try (when time permits). I would generalize the queue mechanism to be usable for all 4 map types and also think about an abstraction for the objects we paint onto different maps (maybe I can use GeoObject as a base). Ok for you?

moving-bits:
> @moving-bits I thought about this some more and I would be read to give this a try (when time permits). I would generalize the queue mechanism to be usable for all 4 map types and also think about an abstraction for the objects we paint onto different maps (maybe I can use GeoObject as a base). Ok for you?

Yes, for sure, this would be a huge step forward for `UnifiedMap`! Thanks for considering it.

eddiemuc:
@moving-bits my idea is currently to build some sort of map-implementation-agnostic "MapObjectGroup" where one can add/remove map-implementation-agnostic "MapObjects". This should be able to handle all types of geoobjects we currently use (markers, circles, polylines, polygons) and also provide click-listeners as well as move/Zoom-Listeners

Then for each map implementation (google V2/v3, mapsforge, VTM) an interface must be implemented for the actual MapObjectgroup to use to delegate the actual drawing/listening stuff to the concrete map implementation

In a first step I would probably try this out as a separate new layer of unifiedmap.

Is there already an issue I could relate my work to?

fm-sys:
@eddiemuc this would be such a great improvement... Awesome!

BTW, it would be great if this interface could provide click listeners for all types of map overlays. We had some feature requests in the past to make e.g. routes (=> polylines) clickable which where highly appreciated but postponed cause nobody had an idea how to implement it.

devemux86:
I have added touch events to path layers in both Mapsforge and VTM.
e.g. can tap and long press the calculated routes in Cruiser

See the Mapsforge [OverlayMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/OverlayMapViewer.java) and VTM [PathOverlayActivity](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/PathOverlayActivity.java) examples.

eddiemuc:
> I have added touch events to path layers in both Mapsforge and VTM.
> e.g. can tap and long press the calculated routes in Cruiser

Vtm/mapsforge supports this very well. Challenge will be to find an abstraction which works well for both VTM/mf and google

eddiemuc:
> We had some feature requests in the past to make e.g. routes (=> polylines) clickable

What should happen when you click e.g. a route?

moving-bits:
`UnifiedMapActivity` has a very basic abstraction layer:

- its method `onTap()` gets called with coordinates and `isLongTap`
- `UnifiedMapActivity` calls its current map view provider to retrieve a list of affected items. If it's more than one, you get a popup selection menu.
- on selecting an item, `handleTap()` is called, which does the distinction between the different objects

By using this method I was able to provide a standardized popup selection for both VTM and Google Maps (a feature not yet implemented in today's GMv2 map).

moving-bits:
> > We had some feature requests in the past to make e.g. routes (=> polylines) clickable
> 
> What should happen when you click e.g. a route?

Let's open a separate issues for this to collect ideas (I'll create one). Abstraction layer can be implemented first, and this use-case can then be added later when ideas are settled.

eddiemuc:
Outdated PR, superseeded by new GeoItemLayer, can be deleted

PR #13928, opened 2023-02-01
Refresh caches in route (fix #13382)

### LINKED ISSUE #13382: update caches on route
### Tell us your idea!

Add a "refresh caches" item to a route. Similar to a list, just got the caves added to a route.

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Adds a "refresh" button for individual route to route/track menu which triggers a forced reload of the route's caches in the background.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/RouteTrackUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/RouteTrackUtils.java
@@ -11,6 +11,9 @@ import cgeo.geocaching.location.Viewport;
 import cgeo.geocaching.maps.routing.RouteSortActivity;
 import cgeo.geocaching.models.IndividualRoute;
 import cgeo.geocaching.models.Route;
+import cgeo.geocaching.models.RouteItem;
+import cgeo.geocaching.models.RouteSegment;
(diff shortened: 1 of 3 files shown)


fm-sys:
Wouldn't it be better to offer a general "show individual route as list" feature, which would be way more powerful?

moving-bits:
> Wouldn't it be better to offer a general "show individual route as list" feature (=> CacheListActivity), which would be way more powerful?

What should be shown in such a list? A route can consist of geocaches, waypoints and "coordinate only points", whereas a list can only show geocaches (AFAIK). So we would either have to extend our lists to be able to show the other entry types as well (loooks like a bigger task to me) or we would prefill this list with geocaches only (something the user probably would not expect).

So I propose to start with this PR, and if someone wants to extend list functionality we can replace it with that then.

eddiemuc:
I think both features would be valuable.

* This PR offers a quick way to refresh all caches participating in a route. Makes sense and should be merged imho
* A second feature could be to open all caches **participating in a route** (cache header or waypoint) in a list to do other stuff with them. Would be a new issue/feature-request imho

PR #13942, opened 2023-02-02
update jenkinsfile

bekuno:
for the changed `main` structure.

This file is used only in the CI job `cgeo-CI_pipeline-experimental`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,33 +13,33 @@ node('has-emulator') {
         try {
             // group steps by stages for better monitoring in the Jenkins pipeline stage view
             stage('Checkout from GitHub') {
-            
+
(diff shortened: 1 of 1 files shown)


bekuno:
@Lineflyer Con you create such a experimental job or how can I do it?

bekuno:
> @Lineflyer Con you create such a experimental job or how can I do it?

Or merge this trivial PR without extra testing it?

Lineflyer:
I guess you can just merge it. 
I only created the experimental jobs to check how it works and whether we can change from the old job to the new pipelines. But I did not continue to work on that, so I just left it there for future  use.

Lineflyer (approved):

PR #13952, opened 2023-02-04
Add Cruiser as navigation app

moving-bits:
## Description
Rudimentary addition of Cruiser as navigation app

Do not merge yet - build is intended for testing by @devemux86.


devemux86:
Thanks!

PR #13974, opened 2023-02-11
Route optimization (experimental feature)

moving-bits:
## Description
~WIP~ Text outdated, see updates below...

This is a basic implementation of a route optimization routine. Not much of a UI, so far, only a button to start it, and results are written to log only. PR implements both hill climbing algorithm and simulated annealing algorithm, as they have different characteristics depending on input set.

Tested it with route sets of 5, 12, 48 and 68 points. For the two larger sets time to calculate the initial distance matrix takes pretty long (four minutes in the emulated device, whereas the two search runs take less than 50s combined in this case).

Some questions:
- ~Is BRouter able to do multiple route calculations in parallel / is it thread safe?~ => yes, except creating `rawtrack` files (which are not relevant here)
- ~How to distribute the N*(N-1) route calculations across the available threads (number depending on available CPU)?~ => implemented
- ~How to continue with the searches only after all distances have been calculated?~ => implemented
- ~How to give user feedback (x% distances calculated? found route of length x etc.) from background thread to the UI? (And should a modal popup be used for it, or should a popup only be presented after having finished the calculations)?~ => implemented

### DIFF
--- a/main/src/main/java/cgeo/geocaching/brouter/core/ProfileCache.java
+++ b/main/src/main/java/cgeo/geocaching/brouter/core/ProfileCache.java
@@ -16,7 +16,7 @@ import cgeo.geocaching.storage.PersistableFolder;
 public final class ProfileCache {
 
     private static long lastLookupTimestamp = 0;
-    private static ProfileCache[] apc = new ProfileCache[1];
+    private static ProfileCache[] apc = new ProfileCache[10];
     private static final boolean debug = Boolean.getBoolean("debugProfileCache");
     private BExpressionContextWay expctxWay;
     private BExpressionContextNode expctxNode;
--- a/main/src/main/java/cgeo/geocaching/maps/routing/RouteOptimizationHelper.java
+++ b/main/src/main/java/cgeo/geocaching/maps/routing/RouteOptimizationHelper.java
@@ -2,10 +2,19 @@ package cgeo.geocaching.maps.routing;
 
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.models.RouteItem;
+import cgeo.geocaching.ui.dialog.SimpleProgressDialog;
+import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.Log;
 
 import android.content.Context;
+import android.view.View;
+import android.widget.Button;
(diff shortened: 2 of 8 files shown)


moving-bits:
Did a first try on parallelizing calculation of distance matrix, but do not yet get the expected results:
- Some sporadic EOF errors - probably two threads trying to read the same input file
- And time gain is only minimal, about 15%, not depending whether I'm using 3, 7 or 16 threads (tested on two different images with 4 or 8 CPU) - maybe the emulator cannot really make use of more virtual (and physical) CPU cores? Need to test against a physical device...

moving-bits:
Some progress here:
- put calculations in parallel background threads
- a progress dialog shows progress and current results
- redo and cancel options

|Generating distance matrix|Calculating route optimizations|Finished route optimization|
|---|---|---|
|image|image|image|

Some things yet to do:
- "use" option is not implemented yet
- dialog formatting
- some generalizations in `SimpleProgressDialog`
- replace hard-coded strings with resources

moving-bits:
To test this feature:
- create a new setting named `experimental_tsp` and set it to true
- got to map and create an individual route
- open route edit and press the "wizard button" at the top

fm-sys:
Great to see this feature!

One question: Geocachers often have a fixed start and end point for their tour (the tour might or might not be a roundtrip). How will the route optimization behave in these cases?

grafik

moving-bits:
> One question: Geocachers often have a fixed start and end point for their tour (the tour might or might not be a roundtrip). How will the route optimization behave in these cases?

I'm always calculating an actual round-trip, starting with the first point of the given route, including all further stops in optimized order, and returning to the first point. There's no need to add starting point as end point manually.

fm-sys:
Sounds good and will probably fit for most of all cases. Nice solution without the need to add additional UI complexity 👍

SammysHP:
> There's no need to add starting point as end point manually.

Except for this situation: a planned trip from point A to B with the plan to visit all downloaded caches in between.

moving-bits:
> > There's no need to add starting point as end point manually.
> 
> Except for this situation: a planned trip from point A to B with the plan to visit all downloaded caches in between.

That would need some adaption to the algorithm, so let's merge here without for now, and create a separate issue for this use case.

PR #13980, opened 2023-02-15
unifiedmap: create unified geoitemlayer

eddiemuc:
unifiedmap: create unified geoitemlayer

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/Viewport.java
+++ b/main/src/main/java/cgeo/geocaching/location/Viewport.java
@@ -90,11 +90,14 @@ public final class Viewport {
      */
     public boolean contains(@NonNull final ICoordinates point) {
         final Geopoint coords = point.getCoords();
-        return coords != null
-                && coords.getLongitudeE6() >= bottomLeft.getLongitudeE6()
-                && coords.getLongitudeE6() <= topRight.getLongitudeE6()
-                && coords.getLatitudeE6() >= bottomLeft.getLatitudeE6()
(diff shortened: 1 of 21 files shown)


eddiemuc:
This PR contains a first version of a map-independent geoitem layer to place all sorts of items upon a map in a map-provider-agnostic way. The PR also contains map-provider-implementations for all map-libaries currently used by c:geo (mapsforge c6, vtm and google v2)

It also contains a "test layer" which draws some test items onto the map for testing purposes (e.g. if map-independent alignment and placement works). It can be activated by enabling the hidden new boolean setting `feature_unified_geoitem_layer`. The test layer is drawn around coordinate `N40 00.000 E10 00.000`

eddiemuc:
image
image
image

fm-sys:
Great, really nice work! 

Next step would be to not only generalize drawing but also responding to item click events. Any plans about it yet?

eddiemuc:
> Great, really nice work!
> 
> Next step would be to not only generalize drawing but also responding to item click events. Any plans about it yet?

Yes, but today I run into a possible dead-end. I wanted to calculate myself from the geoitems whether it was clicked and when now. See screenshot (the cyan rectangles are the "click zones" calculated for the markers):
image

But look what happens if I rotate the Google Map...
image

And I can't imagine what happens when I use 3D mode :-)

moving-bits:
Great work, thanks for putting all this effort into it! Looking forward to seeing this being used for cache & waypoint markers "in real life"...

PR #13987, opened 2023-02-17
fix #13982: update application level locale on user locale change in settings

### LINKED ISSUE #13982: List of extended filters does not get translated if c:geo language is set to non-default
Stumbled upon a strange thing while answering support tickets and having switched c:geo's language from "default system" to "en":

The dialog box that opens when you tap on "add filter criteria" does not get translated, it still shows them in the system's default language. Only when I switch the phone's default language to eg. "English", the dialog gets translated to English. (And it stays English, even if I then configure a different language within c:geo.)

So somehow this dialog does not seem to respect the language setting configured within c:geo, but always resolves to the system's language.

Version: Release 2023.02.12

eddiemuc:
fix #13982: update application level locale on user locale change in settings

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CgeoApplication.java
+++ b/main/src/main/java/cgeo/geocaching/CgeoApplication.java
@@ -42,7 +42,7 @@ public class CgeoApplication extends Application {
         instance = application;
     }
 
-    public static Application getInstance() {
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

bekuno:
retest this please

Issue #14011, opened 2023-02-21
Cannot build c:geo


fm-sys:
I cannot build c:geo any longer, does anyone have an idea on how to fix this?

![grafik](https://user-images.githubusercontent.com/64581222/220316357-243c5692-c223-43a6-bae3-c99cf383df36.png)

eddiemuc:
The mentioned download link in your error message https://jitpack.io/com/github/k3b/k3b-geoHelper/v1.1.11/k3b-geoHelper-v1.1.11.jar works for me. Maybe a temporary internet access problem on your developer machine or on server side?

fm-sys:
Indeed, now it works again. Thanks for looking into it! Internet was working fine for me at that time, so it must have been a temporary server-side problem. 

Let's close for now...

drinkcat:
For future reference, just in case...

I had a similar issue, but on version k3b-geohelper 1.1.12 as that's what master pulls now.

Ended up adding this to `gradle.properties` (based on this [stackoverflow answer](https://stackoverflow.com/questions/65957683/gradle-sporadically-times-out-fetching-jar-from-jitpack)):
```
systemProp.org.gradle.internal.http.connectionTimeout=180000
systemProp.org.gradle.internal.http.socketTimeout=180000
```
and the sync step finally succeeded.

PR #14012, opened 2023-02-21
align margins and optimize out of sync dialog 

### LINKED ISSUE #13883: Inconsistent margins
### Describe your problem!

Some buttons in cgeo are not properly aligned and use inconsistent margins. 

### How to reproduce?

Open cache and create out of sync variable

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2023.01.16

### System information

_No response_

### Additional Information

Screenshot_20230116_093419_cgeo geocaching_edit_339809508401273
Screenshot_20230116_093331_cgeo geocaching_edit_339677364949209

fm-sys:
fix #13883

setting as draft, as I couldn't test my PR cause of #14011

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.ui.dialog;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.Keyboard;
+import cgeo.geocaching.databinding.DialogCustomTitleContentBinding;
 import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.utils.LocalizationUtils;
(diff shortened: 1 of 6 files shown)


fm-sys:
grafik

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -218,6 +219,22 @@ private void applyCommons(final AlertDialog.Builder builder) {
         }
     }
 
+    private void applyCommonsForSelectionDialogs(final AlertDialog.Builder builder) {
+        // By default message is not supported if a listview is shown, therefore we need to build the dialog ourselves and set it as header
+        if (message != null || title != null) {
```
eddiemuc, line 224: Wouldn't it be enough to apply the custom header if message != null? Setting a custom title (without message) is also supported for selection dialog w/o custom header.

fm-sys: Yes in principle, although this would slightly increase complexity cause we would need to set the title in two different ways depending on whether message != null.

Do you have a preference?

eddiemuc: I would call applyCommons in case message is null.

Anyway, current implementation is fine too.

fm-sys: Ok, will change it...

PR #14022, opened 2023-02-23
support setting line width per track (fix #14019)

### LINKED ISSUE #14019: Support changing line width for individual tracks
### Tell us your idea!

It would be great if the "select track color" dialog would support settings an individual line width. 

Add an additional slider

### Is there an alternative?

There are no alternatives to customize the line width on a per track basis. One could only change the width for all tracks currently. 

### Additional Information

_No response_

fm-sys:
fix #14019

The "select track color" dialog will show an additional slider for customizing the track width.

grafik

Example:

grafik


fm-sys:
Use line width slider also in settings as requested in https://github.com/cgeo/cgeo/issues/14019:

grafik

PR #14023, opened 2023-02-24
fix #13999: handle image urls in listing without protocol

### LINKED ISSUE #13999: Image parsing in cache listing fails for incomplete url
### Describe your problem!

By tapping on an image in a cache listing our new image viewer should open.

We have a user reporting that he gets an "c:geo can't find any suitable application" toast when tapping on the checker image in specific cache GC777HK.

### How to reproduce?

- open GC777HK in c:geo's cache detail view
- tap on the checker image for project-gc

error toast "c:geo can't find any suitable application"

Reason: image url is incomplete, it reads `src="//maxcdn.project-gc.com/Images/Checker/27171"` and is missing either `http:` or `https:`.
Still it would be nice if we could detect such shortcomings and add the missing protocol, if needed.

image should open in built-in image viewer

### Reproducible

Yes

### c:geo Version

2023.02.12

### System information

_No response_

### Additional Information

ticket 607468

eddiemuc:
fix #13999: handle image urls in listing without protocol


eddiemuc:
@fm-sys as part of this PR I changed a bit the logic for assigning clickableSpans to images. It would be great if you could have a look at this.

fm-sys (approved):
> ```
>  //don't make images clickable which are surrounded by a clickable span, this would suppress the "original" click
>            //(most prominent example: <a href> links with an  tag inside, e.g. for challenge checkers)
> ```

I thought that I had tested that the original click would win in these cases. But maybe that wasn't true for all edge cases... Anyway, LGTM!

PR #14041, opened 2023-03-01
agp 7.4.2

bekuno:
AGP 7.4.2 is in sync with AS 2022.1.1 Patch 2

For AS see https://developer.android.com/studio/releases?utm_source=android-studio#android-studio-electric-eel-|-2022.1.1-patch-2-february-2023 
For AGP see https://developer.android.com/studio/releases/gradle-plugin#android-gradle-plugin-7.4.2-february-2023

Both are bug fixes.
At `master` we are using already AGP 7.4.1, so this is only a bugfix for c:geo.

As suggested in https://github.com/cgeo/cgeo/pull/14037#issuecomment-1449992144 targeted to release.

The `versions`-Update is already in use at master.

replaces #14037

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:7.3.1'
+        classpath 'com.android.tools.build:gradle:7.4.2'
 
         // un-mocking of portable Android classes
         classpath 'com.github.bjoernq:unmockplugin:0.7.9'
@@ -23,7 +23,7 @@ plugins {
     id 'idea'
 
     // check for updates of gradle plugin dependencies
-    id 'com.github.ben-manes.versions' version '0.44.0'
+    id 'com.github.ben-manes.versions' version '0.45.0'
     id 'se.ascp.gradle.gradle-versions-filter' version "0.1.16"
 
     // use SpotBugs instead of FindBugs, see https://plugins.gradle.org/plugin/com.github.spotbugs


moving-bits:
Is there a way to set different AGP versions in `build.gradle` depending on Android Studio version?

Background for this question: I needed to downgrade AS recently to AS 2021.3.x to "fix" some 100% CPU blockage in AS 2022.1, but needed to downgrade AGP to 7.3.1 as well, as this is the last version compatible with AS 2021.3 - now I need to make sure not to check-in this file... Is there a way to program an automatic switch (or upper limit) into `build.gradle` depending on detected AS version?

bekuno:
> Is there a way to set different AGP versions in `build.gradle` depending on Android Studio version?
> 
> Background for this question: I needed to downgrade AS recently to AS 2021.3.x to "fix" some 100% CPU blockage in AS 2022.1, but needed to downgrade AGP to 7.3.1 as well, as this is the last version compatible with AS 2021.3 - now I need to make sure not to check-in this file... Is there a way to program an automatic switch (or upper limit) into `build.gradle` depending on detected AS version?

I don't think this is possible because the AGP versions lead to different behaviour.
Therefore, I would downgrade AGP instead.

Maybe we should be conservative in general and stick with the AGP-compatible, second to last AS version?

moving-bits:
> Therefore, I would downgrade AGP instead.

Sounds reasonable. Should be done on master, then, as `release` currently still uses AGP 7.3.1

> Maybe we should be conservative in general and stick with the AGP-compatible, second to last AS version?

I propose to generally stay with a reasonable version of AGP being supported by both current (release) version of AS and previous release version. This would be AGP 7.3.1 at max currently, being supported by both AS 2022.1 "Electric Eel" and AS 2021.3 "Dolphin".

moving-bits:
> > > Maybe we should be conservative in general and stick with the AGP-compatible, second to last AS version?
> > 
> > 
> > I propose to generally stay with a reasonable version of AGP being supported by both current (release) version of AS and previous release version. This would be AGP 7.3.1 at max currently, being supported by both AS 2022.1 "Electric Eel" and AS 2021.3 "Dolphin".
> 
> Exact. I will create a PR.

Thanks.

We should also take CheckStyle into consideration, as the most recent CheckStyle versions are not compatible with AS 2021.3.1, see https://plugins.jetbrains.com/plugin/1065-checkstyle-idea/versions.

Most recent AS version in the "Dolphin" series is 2021.3.1 P1, which means we are limited to CheckStyle 5.71.1

bekuno:
replaced by #14092

Issue #13315, opened 2022-08-06
Add fullscreen mode to c:geo


moving-bits:
### Tell us your idea!

From Support:

A user is using c:geo mostly in landscape mode, but there's only a small fraction of the map visible due to toolbar, side menu and bottom navigation.

### What solution would you suggest?

Offer an optional "full screen" mode, which auto-hides toolbar / menu / bottom navigation etc., and displays them on demand (eg: a special button or swipe gesture in a specific place/direction)

(Which elements to auto-hide is TBD)

### Is there an alternative?

leave it as is

### Additional Information

ticket 987083

fm-sys:
This was discussed in #11841, however it was considered not intuitive enough...

fm-sys:
Maybe it could however be implemented as an optional feature (off by default)

fm-sys:
> 
> 
> > > Maybe it could auto-hide on map view?
> > 
> > 
> > This would definitely be possible, and other apps are doing this as well. Although, I'm not sure if this will rather confuse users...
> 
> Seeing that Google Maps as THE most popular map app is doing it it should be ok. Others like Locus also do auto-hide btw.

Coming back to the auto-hide functionality. This could look like this: 

The animation needs to be improved for sure, but is this wanted in general?

![Animation](https://user-images.githubusercontent.com/64581222/140982701-4b0bf46e-15fe-4a89-81b8-0478057f27dd.gif)

Development started in https://github.com/fm-sys/cgeo/commit/8b7467bcbba1269779c0d63e0aedf7be913989a5

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/11841#issuecomment-964419212_

moving-bits:
A few thoughts from my side:
- I would not use auto-hiding, as the "bumping screen" is somewhat disturbing, missing menu navigation may be counter-intuitive, and functionality is not really needed in portrait mode on today's long screens.
- But what about "hiding on demand"? (as well as unhiding) - Button could be enabled (default=off) via settings, and be displayed above the zoom buttons (to not disturb too much in landscape mode)
- Filter bar should also be hidden
- Can the animation be made less bumpy? (Eg by simply extending the visible map, and by moving the map controls to their new place.)

Buntbart:
I would also like to see a button above the "+" button. It should have a fullscreen icon, something like: https://lite.qwant.com/?q=icon+fullscreen&t=images

moving-bits:
[...]

PR #14044, opened 2023-03-03
Hide map's action bar on demand

moving-bits:
## Description
To gain more vertical space on the map (esp. in landscape mode) you can hide the action bar on demand by short-tapping on a free space of the map. Unhiding is done the same way.

For screenshots see https://github.com/cgeo/cgeo/issues/13315#issuecomment-1450862440.


moving-bits:
retest this please

moving-bits:

PR #14046, opened 2023-03-03
Fix possible NPE + suppress lint about string concat

moving-bits:
## Description
Suppress lint message complaining about string concat in a place where it doesn't make sense + prevent a possible NPE in the same place.


moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

PR #14047, opened 2023-03-03
Set compileSDK to 33

moving-bits:
## Description
Set `compileSDK` to 33 to be able to upgrade some dependencies.

### DIFF
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -6,8 +6,8 @@ apply plugin: 'com.android.application'
 
 android {
     namespace 'cgeo.contacts'
+    compileSdk 33
 
-    compileSdk 32
 
     // signing is handled via private.properties
     signingConfigs {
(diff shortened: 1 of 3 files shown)


bekuno:
Have you checked whether changes are required according to https://developer.android.com/about/versions/13/behavior-changes-13?

moving-bits:
> Have you checked whether changes are required according to [developer.android.com/about/versions/13/behavior-changes-13](https://developer.android.com/about/versions/13/behavior-changes-13)?

Wouldn't those become relevant only on setting `targetSDK` to 33?

bekuno:
> > Have you checked whether changes are required according to [developer.android.com/about/versions/13/behavior-changes-13](https://developer.android.com/about/versions/13/behavior-changes-13)?
> 
> Wouldn't those become relevant only on setting `targetSDK` to 33?

You are right.
I started at https://developer.android.com/about/versions/13/setup-sdk after searching for compileSdk 13.
And then misinterpreted target vs compile.

Then lets switch.

Issue #13097, opened 2022-06-07
Handle namespace conflicts for trackables


Lineflyer:
### Describe your problem!

A user contacted us, as he got reports, that his "car trackable sticker" cannot be discovered using c:geo.

The problem here is, that his tracking code starts with `GK` and is thus interpreted by c:geo as the public code of a geokrety trackable (which coincidentally exists) and the geokret is opened instead.

### How to reproduce?

- Open search activity
- Enter a valid Groundspeak tracking code starting with GK into the trackable search

### Actual result after these steps?

The Geokret, whose public code is matching the Groundspeak trackable code is opened instead

### Expected result after these steps?

Short term solution(?):
This problem does also occur if the geokret service is NOT activated. In such cases there should not be any communication with geokret servers but rather only the active connectors should be used, leading in the correct trackable to be returned.

Generic solution:
Ideally requests should go out to all active trackable connectors in parallel and (in case of multiple matches) a selection be offered to the users, which trackable he wants to continue with.

Linked to #4822 (seems to describe the cause of this problem already).

### Reproducible

Yes

### c:geo Version

2022.06.06

ztNFny:
The problem here is that all geocode-searches expect to return exactly 1 result - unlike string-searches. Adressing this would require rewriting the general search-logic to 1) get all results for a geocode, 2) if count == 1 show it, elseif count > 1 show a list of results (like for string searches)

Inside the trackable-search we first check whether the code provided is a geocode in any of the connectors (so anything starting with TB, GK, GC, OC, ...), only if that fails we check for a trackable-code.

The only realistic path I see for this (that has an okish effort-value ratio) is in src/cgeo/geocaching/connector/ConnectorFactory.java to initialize only enabled connectors.

Lineflyer:
Over time more users contacted us with problems based on this issue:

There seem to be more and more tracking codes of Geocaching.com trackables existing out there which start with GK.
As GeoKrety is active in c:geo by default, once users try to look up the trackable via its tracking code the GeoKrety will show up instead and the user has no ability to load the GC trackable with c:geo....or at least I don`t know how (except he knows the public code TB....).

moving-bits:
`ConnectorFactory` has a list of all connectors:

```
private static final Collection<TrackableConnector> TRACKABLE_CONNECTORS = Collections.unmodifiableCollection(Arrays.<TrackableConnector>asList(
    new GeokretyConnector(),
[...]

PR #14056, opened 2023-03-05
TB namespace conflict workaround for disabled GK connector (rel. to #13097)

moving-bits:
## Description
Workaround for TB namespace conflicts for TB starting with "GK":
Leave out GeoKrety TB connector of list of TB connectors, if GeoKrety connector is not activated. In this case no GeoKrety TB will be searched for online.
Be aware, though, that Geokrety TB already stored locally in the database will still be retrieved by the TB search.

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/connector/ConnectorFactoryTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/connector/ConnectorFactoryTest.java
@@ -108,6 +108,7 @@ public class ConnectorFactoryTest {
 
     @Test
     public void testGetTrackableFromURL() throws Exception {
+        ConnectorFactory.updateTBConnectorsList(true); // make sure GK connector is included
         assertThat(ConnectorFactory.getTrackableFromURL("https://www.geokrety.org/konkret.php?id=30970")).isEqualTo("GK78FA");
         assertThat(ConnectorFactory.getTrackableFromURL("https://www.geokrety.org/konkret.php?id=30970")).isEqualTo("GK78FA");
         assertThat(ConnectorFactory.getTrackableFromURL("https://geokrety.org/konkret.php?id=30970")).isEqualTo("GK78FA");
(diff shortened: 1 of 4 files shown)


moving-bits:
GK-related tests failing, which is to be expected when GK connector is not active.

@eddiemuc 
How can I activate GK connector for tests? (Would need to be done before `TRACKABLE_CONNECTORS` static variable gets set.)
Or can I check in code whether we are running tests currently, and always include the connector in this case?

eddiemuc:
> How can I activate GK connector for tests? (Would need to be done before `TRACKABLE_CONNECTORS` static variable gets set.)
> Or can I check in code whether we are running tests currently, and always include the connector in this case?

In such situations I usually try to abstract out an initialization routine (or a constructor) where I can parameterize initialization. In your case e.g. using a boolean parameter to decide whether to initialize GK connector or not and call this in production with the "GK-activates" flag and for tests with a fixed value.

For more concrete advise I am afraid I would have to look deeper into this.

moving-bits:
Thanks for your reply. In this case it's a static variable `TRACKABLE_CONNECTORS` in `connector/ConnectorFactory`, so not being called explicitly... 🤔. I was hoping for some "setting test prerequisites" routine where I can activate GeoKrety connector for tests, before variable `TRACKABLE_CONNECTORS` is set.

eddiemuc:
But there has to be some variability to this? Because iiuc then the trackable must also be activateable/deactivateable at runtime when the user activates/deactivates the GK connector? (Or does this require a restart?)

moving-bits:
Currently it requires a restart. But I guess one could make it a non-static variable instead, and pass a parameter to initialization routine whether to force all connectors or only active ones. Performance punishment of not using a static variable shouldn't be too harsh, and this would also save users the restart on changing the GeoKrety activity status (although this is done infrequently, probably). I'll pursue this approach further, thanks for your input.

moving-bits:
Workaround is in the nightlies for quite some time, and IMHO it's not worth the effort to refactor TB handling to detect namespace conflicts (and present selection dialogs, if so) in case GK connector is activated, so I propose to close this issue.

Issue #14053, opened 2023-03-04
Disappearing waypoints on OSM map


Lineflyer:
### Describe your problem!

Reported by users of 2023.02.12 and 2023.02.25 and now also reproducible for me:
When mapping a list (not sure of other contexts) some of the caches waypoints might vanish from the map when opening and closing the popup for a cache.

### How to reproduce?

- Waypoint limit for map is set to 30 caches
- Using Openandromaps offline map file
- I have a list with 13 caches of a trail
- On one of the caches I added 3 parking waypoints
- I map the list
- I repeatedly tap on that cache and close the popup

### Actual result after these steps?

Look at the three parking waypoints in north, west, south of the shown map disappear and reappear seemingly randomly:

https://user-images.githubusercontent.com/949669/222897653-fe631b75-5962-43b2-bfa5-8a93e04f741d.mp4

### Expected result after these steps?

All waypoints should be shown all the time

### Reproducible

Yes

### c:geo Version

2023.03.04-NB

### System information

```text
Android 13, Samsung S20
```

Lineflyer:
Related to #13675 ?

moving-bits:
IIRC we had some trouble in the past with waypoints of same geocache not having distinct prefixes, which leads to random effects on OSM. Do you have access to your database and can check in table `cg_waypoints`, column `prefix` for your three newly created parking waypoints whether they have the same or distinct prefix values?

Lineflyer:
You are right:
All three created waypoints have no prefix:
![image](https://user-images.githubusercontent.com/949669/222916414-53086c39-41bf-462d-b517-9a8593f98af0.png)

AFAIR I created them by mapping the cache and long tap on the map.

moving-bits:
IIRC `NewMap` uses Geocode + prefix to distinguish waypoints from each other. If multiple waypoint have the same signature, the (randomly) last one will win. As the map gets updated pretty frequently currently, this changes a lot over time.
We can either try to make sure, that every waypoint gets its distinct signature, or it may be solved implicitly with the the great work going on on geoitem layer (#13980), as soon as it starts being used for waypoints as well.

eddiemuc:
> We can either try to make sure, that every waypoint gets its distinct signature, or it may be solved implicitly with the the great work going on on geoitem layer (#13980), as soon as it starts being used for waypoints as well

Even the new geoitemlayer will require us to have a distinct key for each item placed on the map. There's no way around it I am afraid...

[...]

PR #14057, opened 2023-03-05
Assign prefix on adding waypoints (rel. to #14053)

moving-bits:
## Description
Make `EditWaypointActivity` assign new waypoints a (generated) prefix.

## Detailed info
- Editing a waypoint will already assign a prefix if none exists and waypoint is detected as new.
- This will fail, though, for adding new waypoints, as they are currently not being recognized as new.
- This is due to the following code in `onCreate` (`waypointId = extras.getInt(Intents.EXTRA_WAYPOINT_ID);`), which assigns "0", if the value for waypoint id is missing in the intent. "0" is regarded a valid waypoint id, though, therefore this (newly to be created) waypoint will not be regarded as new, therefore the prefix check skipped.

Solution: Add the "new waypoint" marker (= -1) to calling intent on adding a new waypoint.


moving-bits:

Issue #13993, opened 2023-02-18
Multiple lines to navigation target


moving-bits:
### Describe your problem!

We have two users on support which experience multiple navigation lines being drawn to their current navigation target.

### How to reproduce?

One user gives the following description:
- Upgraded to latest system & now get multiple lines to cache when navigating via map. Seems to establish a line from the point where map nav was initiated and another from current location. Once yesterday I saw 4-5 lines from different points on my travel toward the cache.
- Here are pics I was able to capture today about the multiple vector lines.  I get two vector lines when switching from compass to map navigation. If I then switch to enable the live map, it generates multiple vector lines.

### Actual result after these steps?

![image](https://user-images.githubusercontent.com/3754370/219899608-937326d8-babf-451c-bcac-bbd65b7a8c05.png)

### Reproducible

Unclear

### c:geo Version

2023.02.12

### System information

```text
## System information

c:geo version: 2023.02.12

Device:
-------
- Device type: SM-S901U (r0qsqw, samsung)
- Available processors: 8
- Android version: 13
- Android build: TP1A.220624.014.S901USQS2BWA2
- Screen resolution: 1080x2115px (360x705dp)
- Pixel density: 3.0
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 23.04.13 (190408-505809224)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:113
- Set language: en_US (system default)
- System date format: M/d/yy
- Time zone: CST
- Debug mode active: no
- Last backup: never
- Routing mode: Walk
- Live map mode: true
- OSM multi-threading: false / threads: 1
- Map: Google: Satellite
- Id: cgeo.geocaching.maps.google.v2.GoogleMapProvider$GoogleSatelliteSource
- Atts: none
- Theme: none

Filters:
-------
- Hide waypoints: -
- LIVE: Cache Type: Adv Lab ([:inconclusive=false:advanced=false]AND(type:ADVLAB;difficulty_terrain:d=-:d=-:t=-:t=-;status))
- OFFLINE: None ([:inconclusive=false:advanced=false]AND(type;difficulty_terrain:d=-:d=-:t=-:t=-;status))

Stored Filters (#0):

Services:
-------
[...]

PR #14070, opened 2023-03-08
rel to #13993: mitigate google maps problems with drawing multiple lines

eddiemuc:
rel to #13993: mitigate google maps problems with drawing multiple lines

This PR does not completely fix the problem (there are still "leftovers" on the map when many many objects are drawn/removed in quick succession), but it should mitigate it for "normal" navigation usages.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapObjectsQueue.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapObjectsQueue.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.maps.google.v2;
 
 import android.os.Handler;
 import android.os.Looper;
+import android.util.Pair;
 
 import java.util.Collection;
 import java.util.HashMap;
@@ -24,8 +25,8 @@ public class GoogleMapObjectsQueue {
 
     private boolean repaintRequested = false;
 
-    private final ConcurrentLinkedQueue<MapObjectOptions> requestedToAdd = new ConcurrentLinkedQueue<>();
-    private final ConcurrentLinkedQueue<MapObjectOptions> requestedToRemove = new ConcurrentLinkedQueue<>();
+    private final ConcurrentLinkedQueue<Pair<MapObjectOptions, Boolean>> requested = new ConcurrentLinkedQueue<>();
+   // private final ConcurrentLinkedQueue<MapObjectOptions> requestedToRemove = new ConcurrentLinkedQueue<>();
 
     private final RepaintRunner repaintRunner = new RepaintRunner();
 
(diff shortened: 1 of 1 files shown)


eddiemuc:
retest this please

moving-bits:
retest this please

eddiemuc:
It seems CI is unreliable again.... Really hard to work with. I assume failing test is unrelated.

But since this PR goes to release and changes a more fundamental aspect of underlying GMap logic, I would really appreciate a review.

moving-bits:
> It seems CI is unreliable again.... Really hard to work with. I assume failing test is unrelated.

Yes, failing test is unrelated to this PR. Tests are failing due to @Lineflyer having published a new cache, so that the cache tested against in the failing test no longer gets listed on the first page of the test results :-) I'll adapt the test accordingly.

main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapObjectsQueue.java
```diff
@@ -94,19 +99,25 @@ private class RepaintRunner implements Runnable {
...
-            MapObjectOptions options;
-            while ((options = requestedToRemove.poll()) != null) {
-                final Object obj = drawObjects.get(options);
-                if (obj != null) {
-                    removeDrawnObject(obj);
-                    drawnBy.remove(obj);
+            Pair<MapObjectOptions, Boolean> command;
+            while ((command = requested.poll()) != null) {
+                final MapObjectOptions options = command.first;
+                if (command.second) {
```
moving-bits, line 107: It looks to me that this changes the program logic: If I understand the changelog correctly, the old code prioritizes removals over additions - if an object is in both `requestToAdd` and `requestToRemove`, it will be removed, and the entry in `requestToAdd` will be deleted. (For this, `removeRequested` will be processed first.) In contrast to this the new code will process the entries strictly in the order in which they were put into the queue, which could lead to adding and directly removing an object, or adding it several times etc. - Is this change intentionally?

eddiemuc: @moving-bits your observations are 100% correct, and this change is indeed intentionally. 

My observation was that the old algorithm seems to fail when the same list of objects is removed and added in fast succession. Somehow in these situations, several "removes" got lost. I suspected that they were executed earlier than the actual adding in these cases, leading to leftover objects. 

In some test runs my changes seemed to better the situation. 

But you got me thinking with "...and the entry in `requestToAdd` will be deleted. ". I did not consider this in my thinking why the original algo failed. I am not so sure now anymore whether my "fix" is actually the right way to go.

eddiemuc:
Outdated

Issue #13278, opened 2022-07-27
Make cache list additional info configurable


moving-bits:
### Tell us your idea!

Currently we display GC code, D/T rating, Cache size and Premium state as additional info in cache lists.
I would like to see other info here as well, namely year/month of publishing date (for Jasmer challenge), but depending on my current geocaching goals it may be a different info.
On the other hand, I don't care for Premium/Basic state of a cache, so some space could be saved.

### What solution would you suggest?

Make items for additional info line configurable, defaulting to today's items.

Configuration could be a simple multi-item select list (as we have for home screen's additional buttons), to select of a predefined list of items. For a start we could support:
- gc code, d/t rating, Premium/Basic state (those are today's items)
- year/month of cache publishing (for Jasmer challenges)
- (tbc)

### Is there an alternative?

leave it as is

ztNFny:
the current information is manually grouped to fit its designated lines - first line gc code, D / T, second line lists. If this was completly user-configurable how would grouping work?
If e.g. "lists" would be before D / T, gc code then due to the different length of list names gc code would be all over the place.

moving-bits:
Fixed with #14073

Lineflyer:
Although this is already closed:
I would like to see the "last logs smiley row" in there :)

moving-bits:
> Although this is already closed: I would like to see the "last logs smiley row" in there :)

No worries, it's meant to be extensible. 
I'll create a separate issue for it so this request does not get lost.

PR #14073, opened 2023-03-10
Make cache list info items configurable

moving-bits:
## Description
- Make cache list info items configurable
- Add migration for quicklaunch items and showListsInCacheList

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -1182,7 +1182,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         registerForContextMenu(listView);
 
         adapter = new CacheListAdapter(this, adapter == null ? new ArrayList<>() : adapter.getList(), type, sortContext);
-        adapter.setStoredLists(Settings.showListsInCacheList() ? StoredList.UserInterface.getMenuLists(true, PseudoList.NEW_LIST.id) : null);
+        adapter.setStoredLists(StoredList.UserInterface.getMenuLists(true, PseudoList.NEW_LIST.id));
         applyAdapterFilter();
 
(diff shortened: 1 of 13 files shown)


moving-bits:
retest this please

Lineflyer:
Whats this about?
Is there any related issue explaining this feature?

moving-bits:
Sorry, forgot to link to: This is #13278

ztNFny:
Why is this preference on "offline data"? It should be in appearance.
I had to look up this PR and check the code to even find the setting - and that's with knowing it was added

moving-bits:
I've put it there because the old setting for displaying lists was there. Also the list load limit is there (should be moved as well, if we decide to put this setting in a different place).

ztNFny:
All the old settings refer to offline lists only - at least according to their texts. If stuff like load limit also applies to searches its text needs to be changed

Lineflyer:
I had the same feeling at first when looking for this new feature, but then the "loading limit for offline lists" should move with it.

ztNFny:
So in Appearance a new sub-section "Cache Lists"?

fm-sys:
Appearance can probably be renamed to "general". Appearance is a bit misleading as map appearance for example is _not_ there

moving-bits:
I'll prepare this for the nightlies, so we can test it

moving-bits:
What about the "ask which list to store caches in" entry?

Lineflyer:
> Appearance can probably be renamed to "general". Appearance is a bit misleading as map appearance for example is _not_ there

"General" is a bit to generic ;) ......not knowing a better name though.

ztNFny:
Can we rename "Jasmer" to "Place month" or something similar? Even knowing what the jasmer challenge is I didn't know what to expect from the item, had to add it and see it in action.

PR #14076, opened 2023-03-11
add possibility to clear recently viewed caches

fm-sys:
Add possibility to clear recently viewed caches and handle special case that no caches were recently viewed.

grafik

grafik

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -122,6 +122,12 @@ public abstract class AbstractBottomNavigationActivity extends AbstractActionBar
 
     private boolean onSearchLongClicked() {
         final ArrayList<Geocache> lastCaches = new ArrayList<>(DataStore.getLastOpenedCaches());
+
+        if (lastCaches.isEmpty()) {
+            showToast(R.string.cache_recently_viewed_empty);
+            return true;
+        }
+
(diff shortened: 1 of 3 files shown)


main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
```diff
@@ -129,14 +130,18 @@ public View getView(final int position, final View convertView, @NonNull final V
...
-        Dialogs.newBuilder(this)
+        final AlertDialog.Builder builder = Dialogs.newBuilder(this)
                 .setTitle(R.string.cache_recently_viewed)
                 .setAdapter(adapter, (dialog, which) -> CacheDetailActivity.startActivity(this, lastCaches.get(which).getGeocode()))
                 .setPositiveButton(R.string.map_as_list, (d, w) -> {
                     CacheListActivity.startActivityLastViewed(this, new SearchResult(lastCaches));
                     ActivityMixin.overrideTransitionToFade(this);
                 })
-                .show();
+                .setNegativeButton(R.string.cache_clear_recently_viewed, (d, w) -> Settings.clearRecentlyViewedHistory());
```
moving-bits, line 140: When there are no recently view caches, "clear" and "show as list" do not really make sense. Maybe just "ok" to close the dialog?

fm-sys: Fair enough. When thinking about it, a dialog isn't necessary at all in this case. Will change it to a toast notification...

fm-sys:
grafik

Issue #7569, opened 2019-04-28
Map zoom buttons z-order not ok


Lineflyer:
##### Detailed steps causing the problem:
- Open live map (new Google map)
- Move the map so that some caches are shown underneath the +/- zoom buttons
- Typ on + or - button to zoom in or out

##### Actual behavior after performing these steps:
- Cache popup of cache shown "below" the button might open

##### Expected behavior after performing these steps:
- Zoom buttons should have piority

##### Version of c:geo used:
2019.04.28-NBGM-ccdf59f

##### Is the problem reproducible for you?
Yes, most of the time at least

##### System information:
Android 5.0.2

MagpieFourtyTwo:
Hmm, works for me (using latest nightly).

Nevertheless, concerning +/- I had another effect yesterday:  Tapped (-), which then stayed(!) selected, i. e. the color did not change back to white, and the map zoomed out more and more ... really strange.  De-selecting by e. g. tapping (-) again or tapping on (+) did not work, zooming in via pinch neither (just stopped zooming as long as I left pinching fingers on the display, but as soon as they lift off, zooming out went on), had to leave the map page to stop it.

Lineflyer:
This is not a fault on master/nightly but specific for the new google map implementation which is on branch GoogleMapsV2.

rsudev:
It sounds like an effect that I have seen occasionally on all of our map implementations over the years. As we don't really 'control' the zoom buttons and when/how they react to taps (they are built-in to the map) I guess we cannot do much here.

MagpieFourtyTwo:
> This is not a fault on master/nightly but specific for the new google map implementation which is on branch GoogleMapsV2.

Yep - already thought so on re-reading my answer ... ;)

moving-bits:
Zoom control are always displayed in front of map markers for me, but a tap on them sometimes get routed to the zoom control only, and other times to the zoom control AND the map marker.

So far I have not found any documentation on setting a z-order for the controls or prevent the click/tap event cascading.

PR #14079, opened 2023-03-11
Resolve false triggerings of action bar hide (rel. to #14066)

### LINKED ISSUE #14066: Early feedback about action bar hiding
### Describe your problem!

Google map:
- [x] zoom by double tap triggers action bar hiding 
- [x] zoom using zoom controls triggers action bar hiding 
- [x] compass is hidden behind actionbar 

(this does not happen with Mapsforge maps)

General:
- [x] the current animation is quite "bumpy". Such an animation IMHO must look really smooth or users will not understand what happened on the screen.
- Having tested it a bit, it indeed feels like a great improvement for small landscape layouts, however I personally won't like to see it on tablets or portrait layouts. UPDATE: I'll get used to it. Since the above bugs are fixed, I can live with this feature :)

### How to reproduce?

-

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

NB2023.03.06

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Partly resolve #14066 by trying to remove false triggerings of action bar hide on double tapping or tapping zoom controls. Implemented for GMv2 only to test this solution.

@fm-sys 
Can you give it a try in your environment? Thanks.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapView.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapView.java
@@ -83,6 +83,7 @@ public class GoogleMapView extends MapView implements MapViewImpl<GoogleCacheOve
     private final ScaleDrawer scaleDrawer = new ScaleDrawer();
     private DistanceDrawer distanceDrawer;
 
+    private WeakReference<AbstractBottomNavigationActivity> activityRef;
     private WeakReference<PositionAndHistory> positionAndHistoryRef;
     private View root = null;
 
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapView.java
```diff
@@ -425,24 +433,13 @@ public boolean onSingleTapUp(final MotionEvent e) {
                         final Point waypointPoint = googleMap.getProjection().toScreenLocation(new LatLng(closest.getCoord().getCoords().getLatitude(), closest.getCoord().getCoords().getLongitude()));
                         if (insideCachePointDrawable(p, waypointPoint, closest.getMarker(0).getDrawable())) {
                             onCacheTapListener.onCacheTap(closest.getCoord());
```
fm-sys, line 435: Works fine! Maybe move the cache tap listener as well to fix #7569?

moving-bits: Thanks for reviewing.

I had tried to move `cacheTapListener` to `mapOnClickListener` as well, but that will trigger the enormous delay when tapping on a cache marker, which (according to the comment in the source) probably was the reason to write the `GestureListener` in the first place. Thus I reverted to stay with triggering action bar hiding/unhiding.

PR #14080, opened 2023-03-11
Mitigate hiding of compass in GMv2 (rel. to 14066)

moving-bits:
## Description
PR tries to mitigate hiding of compass in GMv2 due to the new action bar hiding function, see #14066.

Solution is not perfect yet, as compass is hidden at first, but gets positioned correctly after first hide / unhide action bar. (Found no way yet around this limitation.)


moving-bits:
retest this please

fm-sys:
Seems to work in principle, although there's probably a bit too much space now:

grafik

fm-sys:
Thought about it a bit. The position is probable fine if we want to avoid moving the compass with every filter bar change... I've implemented an alternative way in #14082 which uses view animator

moving-bits:
Superseded by #14082

PR #14089, opened 2023-03-12
Preference to toggle actionbar hiding on map

ztNFny:


moving-bits:
Thanks for diving into this, you beat me by a minute... Some suggestions:
- I would name the pref title "Short tap on map" and move the preference close to the "Long tap on map" preference (so one or two items up).
- In the preference description you could go a bit more into detail: "Short tap on free map space to ..."
- Currently you disable the feature in `FilterUtils` which is ok, but still leaves the map click listener etc. in place. You could instead move the check one level up in the hierarchy to spare some extra calls and resources.

ztNFny:
> * Currently you disable the feature in `FilterUtils` which is ok, but still leaves the map click listener etc. in place. You could instead move the check one level up in the hierarchy to spare some extra calls and resources.

It looks to me like only GoogleMapView.java is adding an event handler for this, all other places are using existing handlers. Also that'd require the check to be in 5 places instead of 1. Imho this is the correct place.

> * I would name the pref title "Short tap on map"

Would be confusing imho. You're tapping the map all the time - if anything it'd be "Tap on empty map area" - but that's very undescriptive compared to the current title which says exactly what the setting does.
And "short tap" doesn't exist imho, it's just tap.

PR #14096, opened 2023-03-13
visualize coords-only-points of individual route

### LINKED ISSUE #13110: improve individual route rendering for coordinate-only-points

fm-sys:
Points which do not belong to a specific cache will show a small dot. Long-clicking it will remove the routing point.

Implemented for NewMap and CgeoMap only. I didn't want to touch unified map yet, as long as GeoItemLayer is still under development... 

fix #13110
close #13304

grafik

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/downloader/DownloaderTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/downloader/DownloaderTest.java
@@ -66,9 +66,9 @@ public class DownloaderTest {
         // check one named entry
         final Download d = findByName(list, "Portugal");
         assertThat(d).isNotNull();
-        final String sizeInfoString = d.getSizeInfo(); // 297M
+        final String sizeInfoString = d.getSizeInfo(); // 261M as of 2023-03-15
         final int sizeInfoInt = Integer.parseInt(sizeInfoString.substring(0, sizeInfoString.length() - 1));
-        assertThat(sizeInfoInt).isBetween(275, 350);
+        assertThat(sizeInfoInt).isBetween(250, 350);
     }
 
     @Test
--- a/main/src/main/java/cgeo/geocaching/files/GPXTrackOrRouteImporter.java
+++ b/main/src/main/java/cgeo/geocaching/files/GPXTrackOrRouteImporter.java
@@ -2,9 +2,9 @@ package cgeo.geocaching.files;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.location.GeoObjectList;
-import cgeo.geocaching.location.IGeoDataProvider;
 import cgeo.geocaching.models.Route;
 import cgeo.geocaching.models.geoitem.GeoPrimitive;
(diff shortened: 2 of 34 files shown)


eddiemuc:
Info: generic GeoItemLayer is finished (since yesterday). It is meant to be used in both unified map and "old" mapsforge/GMv2 map, allowing for a step-by-step migration of layers towards unified map. Would be good to restrain from including new features for the time being (until concrete layers are migrated).

fm-sys:
> until concrete layers are migrated

@eddiemuc 
Out of interest: Do you plan to migrate these existing layers?

>  It is meant to be used in both unified map and "old" mapsforge/GMv2 map

One layer which is used in all three maps? I fear it would probably be easier to only refactor layers inside unified map and hope that we can retire the "old" mapsforge/GMv2 soon. If we don't have to care about the old maps while building unified map, we can hopefully avoid doing our worst design patterns again. Is it still worth to put so much effort into these legacy map implementations?

eddiemuc:
> > until concrete layers are migrated
> 
> @eddiemuc 
> Out of interest: Do you plan to migrate these existing layers?

That's the plan. As time permits of course...

> >  It is meant to be used in both unified map and "old" mapsforge/GMv2 map
> 
> One layer which is used in all three maps? I fear it would probably be easier to only refactor layers inside unified map and hope that we can retire the "old" mapsforge/GMv2 soon. 

On the contrary. One time we have to move the logic of the layers towards GeoItemLayer anyway. From that point on forward, we can add new features and they will be available for all maps (old and new) immediately. The alternative to this approach would be to stop all map development until unified map is ready for production. I feel that this is not the preferred option (proven by e.g. this PR :-))

ztNFny:
I'd suggest to adjust the style of the dot to match our other icons for visual consistency. Meaning e.g. a thinner black border than they have now.
Do you want me to have a go at it?

fm-sys:
> I'd suggest to adjust the style of the dot to match our other icons for visual consistency. Meaning e.g. a thinner black border than they have now. Do you want me to have a go at it?

Sure, feel free and go ahead 👍

ztNFny:
I have pushed my icon to your branch. I reduced the icon size to 20px but intentionally kept the vector-drawable itself at 24dp to make it a bit easier to hit.

Design goals:
- be a secondary element, i.e. it's smaller than caches, even in dot mode
- be different enough from a cache to not confuse it with one
- feel like a native c:geo element on map

fm-sys:
Really nice, thanks for looking into it!

fm-sys:
(Resolved merge conflicts)

moving-bits:
retest this please

PR #14099, opened 2023-03-14
Fix #14094 - Improve strings for cache list info items

### LINKED ISSUE #14094: Improve strings for list info item configuration
### Describe your problem!

To be platform neutral we should not use "GC Code" but "geo code" as we do in cache details.

Sorry for only writing an issue, but dev environment is not with me to fix it instantly.

Maybe we can (in general?) also prefer to reuse existing strings for such new features to make it easier to keep consistent wording and avoid additional translation effort at the same time. I guess most of the items of the list items config is already existing somewhere (e.g. in cache details view).

### How to reproduce?

Check string cacheListInfoItem_GCCode

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2023.03.12-NB

### System information

_No response_

### Additional Information

_No response_

Lineflyer:
## Description

Proposal for improved strings for this new feature

## Related issues

#14094

### DIFF
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -959,23 +959,23 @@
     <string name="init_list_initial_load_limit">List Load limit</string>
     <string name="init_summary_list_initial_load_limit">Maximum number of caches to load from database initially when displaying stored lists. Choose 0 for \'unlimited\'</string>
     <string name="init_summary_listdisplay">Show membership of caches to additional lists in a cache list</string>
-    <string name="init_title_cacheListInfo1">Show infos in cache list (Line 1)</string>
-    <string name="init_title_cacheListInfo2">Show infos in cache list (Line 2)</string>
-    <string name="init_summary_cacheListInfo">Select one or more items to be displayed for each cache in cache list</string>
+    <string name="init_title_cacheListInfo1">Configure list info lines (Line 1)</string>
+    <string name="init_title_cacheListInfo2">Configure list info lines (Line 2)</string>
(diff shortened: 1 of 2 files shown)


main/src/main/res/values/strings.xml
```diff
@@ -959,23 +959,23 @@
...
+    <string name="init_title_cacheListInfo2">Configure list info lines (Line 2)</string>
+    <string name="init_summary_cacheListInfo">Select the additional information to be shown for each cache in cache lists</string>
     <string name="init_livelist">Show Direction</string>
     <string name="init_summary_livelist">Show an arrow with compass direction for caches in a list</string>
     <string name="init_quicklaunchitems">Quick launch buttons</string>
     <string name="init_summary_quicklaunchitems">Select one or more items as quick launch buttons on home screen</string>
 
     <!-- names for CacheListInfoItem -->
-    <string name="cacheListInfoItem_GCCode">GC Code</string>
+    <string name="cacheListInfoItem_GCCode">Geocode</string>
```
bekuno, line 971: I propose to also change the string name here.

Lineflyer: mmh..strange. What am I doing wrong. I used the context option "Find usage":
image

Lineflyer: found it. Commit incoming

moving-bits:
FYI: To rename a variable (string or other) in AS, you can mark the variable name, press Shift-F6, and enter the new name. AS will then automatically rename all occurrences of the var (taking variable scope into account).

bekuno:
Should this also work for enum values?

moving-bits (approved):

PR #14102, opened 2023-03-15
Two row-layout for CheckableItemSelectActivity (fix #14087)

### LINKED ISSUE #14087: Different layout for CheckableItemSelectActivity
### Describe your problem!

When opening the cache list info configuration for the first time, nothing is preselected. When leaving the activity again, the current state will be stored which leads to information being missing in CacheListActivity. 

### Reproducible

Yes

### c:geo Version

2023.03.12

### Additional Information

Two more things (unrelated to the main issue, sorry that I was to lazy to open separate issues):
- There could be a bit more margin at the left
- instead of having an "add" button for every element, it's IMHO more common to have two sections "selected items" and "available items" where all elements can be moved between those two sections.

moving-bits:
## Description
Use drag & drop to select entries (instead of using buttons) as suggested in #14087

### DIFF
--- a/main/src/main/res/layout/checkable_item_select_activity.xml
+++ b/main/src/main/res/layout/checkable_item_select_activity.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.core.widget.NestedScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="vertical">
+
+        <include android:id="@+id/headerSelected" layout="@layout/info_header" />
+
+        <androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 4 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

main/src/main/res/layout/checkable_item_select_activity.xml
```diff
@@ -0,0 +1,41 @@
...
+
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/recyclerView"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent" />
+
+        </LinearLayout>
+</androidx.core.widget.NestedScrollView>
+
+    <!--<androidx.core.widget.NestedScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
fm-sys, line 22: Nice! Commented out code should be removed though...

moving-bits: Sure, an is already locally. Currently I'm waiting for the CI tobe usable again (captcha problem on both executord), @mucek4 is already working on it.

PR #14141, opened 2023-03-25
Use d/t marker resources by id when possible

moving-bits:
## Description
Make d/t marker resources identifiable by id rather than accessing them by name only. Also fixes a couple of "unused res" lint messages


moving-bits:
retest this please

PR #14143, opened 2023-03-26
Fix lint: Remove unused strings

moving-bits:
## Description
Remove several strings no longer in use (old gallery, updated preferences etc.)


moving-bits:
retest this please

PR #14149, opened 2023-03-26
Display log summary on cache overview (fix #13274)

### LINKED ISSUE #13274: Display log summary on overview page
### Tell us your idea!

User request on support: Display the log summary (number of founds etc.) also in the cache summary on the cache detail page.

image

### Is there an alternative?

Leave it as is

### Additional Information

ticket #625615

moving-bits:
## Description
Display log summary on cache overview. 
Also a prerequisite for including such an overview in cache list info.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
+++ b/main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
@@ -18,13 +18,8 @@ import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
-import androidx.appcompat.widget.TooltipCompat;
-
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
 
 import org.apache.commons.text.StringEscapeUtils;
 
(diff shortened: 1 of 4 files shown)


ztNFny:
a strong AGAINST-vote from me. We should NEVER duplicate info imho. This would be the first piece of info shown in multiple tabs.

moving-bits:
Yes, it's a duplication, but cache overview page is the first page you'll get when tapping on a cache in the map. (BTW one might then discuss whether the "recent cache smileys" is sort of a duplication as well... - but it's still a useful shortcut.)

It's been a user request some eight months ago, with one upvote and no downvotes until today. But we can discuss this for sure.

ztNFny:
> (BTW one might then discuss whether the "recent cache smileys" is sort of a duplication as well... - but it's still a useful shortcut.

well, it kinda is - but then again to see the same info in logs you usually have to scroll several pages. So the condensing into a single line adds considerable value.
The total number of logs is not something that's usually relevant imho, unlike the recent 2-3 logs.

Description would be - again imho - many many times more relevant than total logs, but for sure we don't want to duplicate that into details tab.

fm-sys:
As you probably know, I'm also one of the guys continuously advocating to decrease UI complexity instead of increasing in further... So I'm with @ztNFny

moving-bits:
I'm totally fine with the vote, so let's close this PR. I only wished that feedback for the original issue would have come a bit earlier, before starting implementation.

PR #14151, opened 2023-03-26
fix adding caches to GC lists / creating GC lists

ztNFny:
the existing calls to https://www.geocaching.com/api/proxy/web/v1/lists now require a "X-Verification-Token" header, the value of which can be obtained from cache pages: input[name="__RequestVerificationToken"]

There's also a new API for managing lists - https://www.geocaching.com/plan/api/lists/ - but this PR doesn't switch to that as it's not required (yet?)

Targeted against release as this should be in the next bugfix


ztNFny:
retest this please

ztNFny:
@moving-bits any idea what's going on here? There are many tests failing that aren't remotely effected by this PR. Problem with the build machine again?

moving-bits:
retest this please

moving-bits:
> @moving-bits any idea what's going on here? There are many tests failing that aren't remotely effected by this PR. Problem with the build machine again?

It seems that we have problems with one of the two executors, have seen this over the last weeks as well. Both executors have been restarted yesterday, but this problem seems to persist.

moving-bits:
Build is green meanwhile, so let's give it a try in the nightlies

Issue #14109, opened 2023-03-17
Crash when opening caches from map


geocachermgo:
### Describe your problem!

Since some days the nightlys are quite unstable.
If I open caches from the map view after the second or third when I click on "Mehr Details" c:geo chrashes and jumps back to the main page

### How to reproduce?

Go to map view
Shift to area with no stored caches
xxx
Click on cache symbol
Click on more details
Use back button to go back to map
Goto xxx

### Actual result after these steps?

c:geo crashes and restarts at main page

### Expected result after these steps?

No crash, but show details of cache

### Reproducible

Yes

### c:geo Version

2023.03.17-NB-f73cf2a

### System information

```text
## System information

c:geo version: 2023.03.17-NB-f73cf2a

Device:
-------
- Device type: MI 8 (dipper, Xiaomi)
- Available processors: 8
- Android version: 10
- Android build: QKQ1.190828.002 test-keys
- Screen resolution: 1080x2029px (392x737dp)
- Pixel density: 2.75
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: disabled - 23.08.55 (100400-514581917)
- Memory:  Available:1,1 GB, Total:5,5 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:224
- Set language: de_DE (system default)
- System date format: dd.MM.yy
- Time zone: GMT+01:00
- Debug mode active: no
- Last backup: 12. März, 08:34
- Routing mode: Bike
- Live map mode: true
- OSM multi-threading: true / threads: 4
- Map: Google: Satellit
  - Id: cgeo.geocaching.maps.google.v2.GoogleMapProvider$GoogleSatelliteSource
  - Atts: none
  - Theme: none

Filters:
-------
- Hide waypoints: -
- LIVE: Status: Eigener=Nein ([:inconclusive=false:advanced=true]AND(status:owned_no;size;difficulty_terrain:d=-:d=-:t=-:t=-;type))
- OFFLINE: Keine ([:inconclusive=false:advanced=true])

Stored Filters (#3):
- gefunden: [gefunden:inconclusive=false:advanced=false]AND(status))
[...]

PR #14157, opened 2023-04-01
rel to #14109: log content and size of individual route on saving

eddiemuc:
rel to #14109: log content and size of individual route on saving

This PR logs (on debug level) infos about content and size of `IndividualRoute` before it is saved into transaction when leaving NewMap (Mapsforge V6). This is to further analyze #14109 because this object seems to become very large (around 2MB)


moving-bits:
No longer required after #14165?

eddiemuc:
> No longer required after #14165?

Indeed no longer required

Issue #14113, opened 2023-03-19
Design pattern idea for unified map 


fm-sys:
In our last dev meeting, I had briefly discussed with @moving-bits how we could decouple parts of our unified map implementation in a better way than it is done in CgeoMap / NewMap. I had promised to think about it a bit, and that's what I came up with:

At almost all places in cgeo we currently trigger UI updates by actively calling a method after data has changed.
However, AndroidX provides many smart features to write event driven code (observer pattern), which would IMHO enable us to write code which is cleaner and easier to understand.

In short, I recommend to write a data holder class based on `ViewModel`, in which all fileds are wrapped into a `MutableLiveData`. 

Advantages:
- Moving all data into a ViewModels will allow us to cleanly decouple parts of unified map, even if they need to share data between each other 
- ViewModels are lifecycle aware, so the data will survive layout changes automatically 
- ViewModels will unregister listeners automatically when activity is shut down, which avoids memory leaks 
- MutableLiveData is a wrapper class which provides an observer based interface for listening to data changes. 

Ivve found a rather good article describing the background of using LiveData: (sample code is written in kotlin, but it works with Java in the very same way)
https://medium.com/mobile-app-development-publication/understanding-live-data-made-simple-a820fcd7b4d0

Very detailed Java tutorial:
https://betterprogramming.pub/create-an-app-that-uses-livedata-and-viewmodel-in-java-f8086ca94229

Side note: I haven't used view models myself before, but I just tried it out in one of my private projects and it was a pleasure to work with ViewModels and MutableLiveData.

eddiemuc:
Does this mean that current development on unifiedmap should be stopped until this is sorted out?

fm-sys:
> Does this mean that current development on unifiedmap should be stopped until this is sorted out?

Development doesn't need to stop, but at least thats the reason why I was wondering how you want to build one layer which works across all 3 map implementations (https://github.com/cgeo/cgeo/pull/14096#issuecomment-1465744333). I would suggest to use the GeoItemLayer only for unified map, so that we are free in terms of code design and don't have to fear that we'll break our current maps. Or we could backport the design pattern to our old maps, which however seems to me like quite some unnecessary additional effort...

If you are interested and we want to go with it, I could provide a MVP implementation of my idea based on the current state of unified map (as time permits of course).

eddiemuc:
> thats the reason why I was wondering how you want to build one layer which works across all 3 map implementations.

[...]

Issue #14027, opened 2023-02-25
Add route optimization for linear routes


moving-bits:
_Originally posted by @SammysHP in https://github.com/cgeo/cgeo/issues/13974#issuecomment-1445053272_

> > There's no need to add starting point as end point manually.

> Except for this situation: a planned trip from point A to B with the plan to visit all downloaded caches in between.

---

Additional info:

Currently our optimization algorithm plans a route _beginning and ending_ with the first point of the current individual route, trying to optimize the path for the points visited in between, visiting all points of the individual route along this way.

Planning for a linear route would have to find an optimized route _beginning with the first point of the current individual route_ and _ending with the last point of the current individual route_, visiting all points in between in an optimized order.

ztNFny:
I tested this for the first time today and while ordering points works in general there's one major item missing: there's no fixed end-point and thus even after adding my target to route to make it a round-trip after optimizing the route that point is no longer last, making it a linear route.
I suggest to just set the last point in the route as fixed.

ztNFny:
> Currently our optimization algorithm plans a route beginning and ending with the first point of the current individual route

That doesn't seem to be the case. It just ends the route wherever it'll end up shortest.

moving-bits:
> > Currently our optimization algorithm plans a route beginning and ending with the first point of the current individual route
> 
> That doesn't seem to be the case. It just ends the route wherever it'll end up shortest.

It does not add the starting point as last point to the visible route, but when calculating a route length it uses the starting point also as final point, making it a round trip. This is why this issue here exists, as there are currently no linear routes being calculated.

ztNFny:
> It does not add the starting point as last point to the visible route, but when calculating a route length it uses the starting point also as final point, making it a round trip

That's definitly not what I'm seeing here (master). The route ends with the cache farthest away from my start point, this route doesn't make any sense as a round-trip.

ztNFny:
![image](https://user-images.githubusercontent.com/1258173/230737715-e855dd2b-582b-4415-bbcb-826400369051.png)
1 is route start, 2 is the last cache after auto-sorting
red is my hand-drawn shortest round-trip-route

moving-bits:
Thanks for the detailed illustration. Question now is: Is it a programming error or an artifact of both algorithms only being approximations and no full evaluations?

Route initialization is done like this:
```
for (int i = 0; i < routeSize; i++) {
    best[i] = i;
}
best[routeSize] = 0;
```

[...]

Issue #13166, opened 2022-06-18
Fatal exception caused by java.lang.IllegalStateException


ztNFny:
### Describe your problem!

Latest master (https://github.com/cgeo/cgeo/tree/fa5e0e9374d737b48705b780c3ebcb431e58994f) crashs on startup
Crash is caused by the changes in https://github.com/cgeo/cgeo/pull/13142

### How to reproduce?

```
2022-06-18 20:18:57.551 17841-17841/cgeo.geocaching E/AndroidRuntime: FATAL EXCEPTION: main
    Process: cgeo.geocaching, PID: 17841
    java.lang.RuntimeException: Unable to start activity ComponentInfo{cgeo.geocaching/cgeo.geocaching.MainActivity}: android.view.InflateException: Binary XML file line #17 in cgeo.geocaching:layout/main_activity: Binary XML file line #17 in cgeo.geocaching:layout/main_activity: Error inflating class androidx.fragment.app.FragmentContainerView
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3449)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
     Caused by: android.view.InflateException: Binary XML file line #17 in cgeo.geocaching:layout/main_activity: Binary XML file line #17 in cgeo.geocaching:layout/main_activity: Error inflating class androidx.fragment.app.FragmentContainerView
     Caused by: android.view.InflateException: Binary XML file line #17 in cgeo.geocaching:layout/main_activity: Error inflating class androidx.fragment.app.FragmentContainerView
[...]

PR #14159, opened 2023-04-02
rel. to #14113 - Showcase: new architectural design for unified map

fm-sys:
@eddiemuc @moving-bits 

As discussed in #14113, here is the promised example.

I started decoupling different layers related to current user's position. I would be really interested to get some feedback what you think about this concept.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/AbstractPositionLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/AbstractPositionLayer.java
@@ -1,37 +1,25 @@
 package cgeo.geocaching.unifiedmap;
 
-import cgeo.geocaching.CgeoApplication;
-import cgeo.geocaching.R;
 import cgeo.geocaching.location.Geopoint;
-import cgeo.geocaching.maps.PositionHistory;
 import cgeo.geocaching.maps.routing.Routing;
 import cgeo.geocaching.models.IndividualRoute;
 import cgeo.geocaching.models.Route;
-import cgeo.geocaching.models.TrailHistoryElement;
 import cgeo.geocaching.models.geoitem.IGeoItemSupplier;
 import cgeo.geocaching.settings.Settings;
-import cgeo.geocaching.utils.AngleUtils;
-import cgeo.geocaching.utils.ImageUtils;
-import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapLineUtils;
 import cgeo.geocaching.utils.functions.Action1;
 import cgeo.geocaching.utils.functions.Action3;
 import cgeo.geocaching.utils.functions.Func1;
 import cgeo.geocaching.utils.functions.Func2;
-import static cgeo.geocaching.settings.Settings.MAPROTATION_AUTO;
-import static cgeo.geocaching.settings.Settings.MAPROTATION_MANUAL;
 
-import android.graphics.Bitmap;
 import android.location.Location;
 import android.view.View;
 
 import androidx.annotation.Nullable;
-import androidx.core.content.res.ResourcesCompat;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
-import java.util.Objects;
 
 import org.oscim.core.GeoPoint;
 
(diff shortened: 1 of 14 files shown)


eddiemuc:
Let's see if I understood the use case correctly:
* In the heart of your change there's the new UnifiedMapViewModel class whose aim it is to provide all (unified-)map-related data using the mutuallivedata-pattern
* As a first example you implemented provisioning of the Location/Heading data which is used in many places inside the map
* You also used this first example of mutual live data to provide new layer versions to paint using this info:
  * a new PositionLayer which paints positionMarker and accuracy circle
  * a new PositionHistoryLayer which paints the (optional) history trail available in c:geo
  
Correct?

My short feedback to your showcase: I like the pattern very much! I would be all in favor to use it for UnifiedMap.

Two things I noticed (both not relevant for the general decision to use this new pattern):
* In PositionLayer you created two instances of GeoItemLayer, each storing exactly one geoobject (one for positionmarker, one for accuracy circle). While I assume that this works, GeoItemLayer was not optimized for holding only one single element. Each GeoItemLayer will create own thread logic with own thread time scheduling to update map objects, and this introduces some overhead which would not be justified if such a layer would be created for single items only. It would be better to group several items into one GeoItemLayer (in this case: use same layer instead of creating a second one). GeoItemLayer instances can even be shared between "logical" layers as long as keys are not duplicated between them.
* PostionHistoryLayer contains business logic: the storing of trail history. Maybe we can find a way to put this logic also into the View Model

Great work! All in favor!

fm-sys:
Thanks for the nice feedback :) 

> Let's see if I understood the use case correctly

Yes, exactly. You summarized it pretty well 👍 (sorry that I was too lazy to document it myself)

> In PositionLayer you created two instances of GeoItemLayer, each storing exactly one geoobject

If I understand it correctly, the z-order can currently be set on a per-layer basis only. (or did I missed something?) But yes, this can and should probably be changed...

> PostionHistoryLayer contains business logic

Good catch! Just didn't had an idea how to do it better... Will give it a try.

moving-bits:
Can't say much about the Android frameworks involved here (don't know them yet in detail), but unifying code across different map types is what `UnifiedMap` is all about, and the nice new GeoItemLayer gives the measures to do that in this case. Also using methods provided by Android framework for maintaining state etc. instead of doing that on our own, using lots of "boilerplate code", seems to be very reasonable. So very much appreciated what you have started in your showcase!

Just we should test the results thoroughly, as we have introduced new Android mechanisms in other places in the past, and suddenly all kind of errors pop up - very likely not coming from the new code, but from "glueing" new architectural principles and pretty old legacy code in those places (eg. database IllegalStateException, SQLiteDatabaseLockedException in different places).

fm-sys:
> Just we should test the results thoroughly, as we have introduced new Android mechanisms in other places in the past, and suddenly all kind of errors pop up - very likely not coming from the new code, but from "glueing" new architectural principles and pretty old legacy code in those places (eg. database IllegalStateException, SQLiteDatabaseLockedException in different places).

You are probably talking about problems like e.g. #13166. These things are not necessarily related to glueing new architectural principles and pretty old code in general, but are probably just accidental mistakes which of course will break the code. This can happen easily, however isn't related to new architecture components but can happen at any time when doing refactoring. 

What I want to say: 
I do not believe that my code is without errors (on the contrary), but all problems which might arise should be solvable without ugly hacks. 

But for sure, testing it thoroughly is important ;-)

eddiemuc:
>> In PositionLayer you created two instances of GeoItemLayer, each storing exactly one geoobject

> If I understand it correctly, the z-order can currently be set on a per-layer basis only. (or did I missed something?) But yes, this can and should probably be changed...

Correct. I thought this was ok since this is how layers are used today in the "old" map implementation.
If this is a too-hard restriction, then setting a z-Level can be moved towards GeoItem/GeoPrimitive. Implementation for GoogleMaps is super-easy (Google Maps sets the z-level per item anyway), Mapsforge/VTM I would have to look into but should be doable.

fm-sys:
> > PostionHistoryLayer contains business logic
> 
> Good catch! Just didn't had an idea how to do it better... Will give it a try.

Ok, I came up with a solution. I moved the `PositionHistory` object into the view model (wrapping it with `MutableLiveData` as well). If it's set to `null` the position history gets hidden, otherwise it will contain the current `PositionHistory` object. The layer itself now only contains drawing, everything else is handled at other places.

fm-sys:
How do we want to proceed further? 

For sure, there is still very much to do... Refactor drawing of caches, track, route, distance view which are the things already implemented to unified map. And then all the rest, which is still missing.

However, from my side this PR should be ready. So I would suggest merging it, and collaborate further on `master` branch.

moving-bits:
Let's get this into the nightlies for further collaboration. Thanks @fm-sys for diving into this!

fm-sys:
Just for info (and to avoid unnecessary work):

I'm currently working on a refactored tracks layer based on geoitemlayer and this new design concept

moving-bits:
> I'm currently working on a refactored tracks layer based on geoitemlayer and this new design concept

Great, thanks! I'll then wait with implementing a "isRoundtrip" flag for individual routes (as proposed in https://github.com/cgeo/cgeo/issues/14027#issuecomment-1501084361) until new track layer is in to avoid unnecessary implementation work.

fm-sys:
> I'll then wait with implementing a "isRoundtrip" flag for individual routes (as proposed in https://github.com/cgeo/cgeo/issues/14027#issuecomment-1501084361) until new track layer is in to avoid unnecessary implementation work.

Up to now, I've only touched track handling, not yet the individual route. But even when refactoring the individual route as well, it will most likely not affect each other... So you can work on "linear route optimization" if you like :)

Issue #14158, opened 2023-04-01
Remove/rework the [Render Complete Description] button


MagpieFourtyTwo:
### Tell us your idea!

Stumbled upon the ``Render Complete Description`` button for the first time today. And, TBH, it was annoying to the max.  :/

Cause every single time this cache has been closed (no matter if by me, to either quit the app or just switch to a different cache forth and back, or by Android's "background process cleaning task"), I had to click this button over and over again. And every single time I had to do so (mind you: after scrolling all the way down, to first get to the button), the page's scroll position went back up to the top and I had to scroll all the way down *again.*

Which I did not like in even two respects at once, because firstly I had double the scrolling effort on every single open, and secondly I always had to search the previous position line by line to get to where I was before .... extremely annoying.

Still, I would understand if this should help to work around any download problems (to display at least something as soon as possible e. g. if the download takes too long due to weak signal, bad server connections, 404s or really huge listings), but why shouldn't a description be displayed completely on the first go if it's data is already stored on the device anyway ...?

Three questions therefore arise for me:
- What was the intention of this button? Was it to save rendering time (which IMNSHO should not be an issue at all with modern CPUs) or to get down to the PN a little bit faster (which would be even faster via ``Menu`` => ``Personal note`` anyway) or what was the reason ...?
- Why do I have to press this button every single time, again and again?
- Why do I first have to scroll down, to get to this button, if afterwards I have to scroll down again, anyway?

### What solution would you suggest?

As for me, I see three possibilities:
1. remove this button completely (cause it has no obvious benefit, but just annoyances (pls. correct me, if I missed s/th))
2. if consensus will still be, to keep it: pls. avoid superfluous scrolling, pls. ...
    - remember the scroll position, to keep the window position where it i, or,
       if this is not recommendable, perhaps because of potentially incomplete renderings above this position,
    - move the button to the top, i. e. *before* the description or ...
    - add a "fast scrolling" bar to the description
4. if the button has a good reason, remember the [Render Complete Description] click automatically,
    for at least the current session or ...
5. if user should still have the (manual) choice, ...
   - add a tick ``[_] always show all`` or ...
   - change the button to a toggle (i. e. either ``Render all immediately`` or ``Render after click``)

### Is there an alternative?

Live with the annoyances.  ;)

fm-sys:
Rendering a long description can make cgeo unresponsive. That's why we have implemented this render limit.

However, I agree that the scroll position should be preserved.

[...]

PR #14160, opened 2023-04-02
rel. to #14158 - preserve scroll position

fm-sys:
Preserve scroll position when using "render complete description" button

https://user-images.githubusercontent.com/64581222/229358204-c01caef8-3529-4e27-a3f4-a3a5154546d8.mp4

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1697,7 +1697,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             binding.getRoot().setVisibility(View.VISIBLE);
 
             // load description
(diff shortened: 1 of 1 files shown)


moving-bits:
Nice, thanks for diving into it. Can you provide something similar for "edit waypoint"? (recent user request on support, need to look up yet whether we already have a ticket for it)

Lineflyer:
@moving-bits You sure its about waypoints? I thought it was about leaving the description tab towards waypoint tab and returning to description (which does not store scroll position).
That would mean: Generally store scroll position of description while the user swipes through other tabs of the same cache until he closes the cache.

fm-sys:
This is the problematic code part:

grafik

Instead of reloading everything, this could be made more context aware...

I'll move this discussion into a separate issue

PR #14169, opened 2023-04-08
fix import of public bookmark lists

ztNFny:
Bookmark lists are now downloaded as GPX instead of ZIP, breaking our list import. This is the fix for newly introduced functionality on master, not yet in release. It's required in addition to #14168


ztNFny:
retest this please

PR #14170, opened 2023-04-08
route linear, keeping first and last waypoint and only sorting those …

ztNFny:
Use linear instead of round-trip routing when using route optimization. The current round-trip scenario is broken per https://github.com/cgeo/cgeo/issues/14027#issuecomment-1500953310


ztNFny:
retest this please

moving-bits:
Round-trip calculation is fixed by https://github.com/cgeo/cgeo/issues/14027#issuecomment-1500969719 / #14171 meanwhile, thus I would prefer to continue with #14027 (implementing linear route calculation as addition) instead of changing the current behavior, thus setting WIP for now. I'll prepare a PR for #14027.

PR #14171, opened 2023-04-08
fix round-trip route optimization

ztNFny:
fixes error in round-trip route optimization that cause it to be linear instead of round-trip


moving-bits:
failing tests are unrelated

PR #14172, opened 2023-04-08
Add linear mode for route optimization

moving-bits:
## Description
Allow user to choose between "round trip" and "linear" mode for route optimization


moving-bits:
retest this pleae

moving-bits:
Let's close this PR, see discussion and proposal in https://github.com/cgeo/cgeo/issues/14027#issuecomment-1501084361

PR #14174, opened 2023-04-09
Avoid race conditions on reordering route (fix #14135)

### LINKED ISSUE #14135: Editing route sometimes leads to invalid lines
### Describe your problem!

Stumbled upon this during my last cache tours: Editing an individual route can lead to invalid lines being drawn on the map. Closing and reopening the map fixes it. Happend with Mapsforce `NewMap`, have not yet tested against other map types.

Have observed this in both current release 2023.02.25 and current state of `master`, but do not remember having seen this in older releases of c:geo.

### How to reproduce?

Does not happen always, but some relatively safe way to reproduce it is:
- create an individual route with a couple of entries (four seems to be enough)
- edit individual route and move one point from close to one end to close to the other (eg: fourth point of a four point route to second position) and save changes

- sometimes the new route shows invalid lines
- closing and opening the map fixes it (as well as opening and closing route edit does)

Example:
- created a route with four points in order 1-2-3-4
- moved the last point to second position (creating 1-4-2-3 in original numbering)
- but it looks like the route is 1-3-2-4 or 1-4-3-2 in original numbering

|before|after edit|after closing/opening map|
|---|---|---|
|image|image|image
|see (a)|see (b)|see (c)|

a) numbers indicate cache order within route
b) see the extra line between (original) points 2 and 3
c) numbers indicate cache order within route; extra line is gone

correctly update map lines for route

### Reproducible

Yes

### c:geo Version

2023.02.25

### System information

_No response_

### Additional Information

Update: Seems to be OSM-specific, was not able to reproduce this behavior on GMv2 map

moving-bits:
## Description
Update route on screen only after having fully reloaded it, to avoid race conditions as shown in related issue

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/IndividualRoute.java
+++ b/main/src/main/java/cgeo/geocaching/models/IndividualRoute.java
@@ -67,7 +67,7 @@ public class IndividualRoute extends Route implements Parcelable {
     }
 
     public void reloadRoute(final UpdateIndividualRoute updateRoute) {
-        clearRouteInternal(updateRoute, false);
+        clearRouteInternal(null, false);
         AndroidRxUtils.andThenOnUi(Schedulers.io(), this::loadRouteInternal, () -> updateRoute(updateRoute));
     }
 


moving-bits:
Codacy is stuck since a couple of hours, or at least its results are not reported back here - not sure what's going on... 🤔

fm-sys:
PR was analyzed correctly, however results were not reported. I've just logged in to Crowdin and selected to reanalize the PR. Now everything is green...

PR #14177, opened 2023-04-09
migrate tracks layer for unified map

fm-sys:
This migrates only the track part, not yet routes...


fm-sys:
Notes to myself: 
- Track is currently lost on changing tile source
- Check whether removal of tracks does work as expected

PR #14178, opened 2023-04-09
Avoid truncation of first smiley row on line breaks (fix #14163)

### LINKED ISSUE #14163: [Nightly] "Last logs smiley row" gets truncated, if the list of smileys hits EOL
### Describe your problem!

If the row of smileys does not fit completely into the current line, some smileys get lost.

### How to reproduce?

1. Go to ``Settings`` => ``Appearance`` => ``Configure list info lines``
2. Set ``Active Elements`` in a way that ``Recent Logs`` will most probably be the last item of the line,
    *but* it should not always entirely fit in (so "normally" you would get a CR/LF somewhere inside of the smiley row).
    On my display over here, for example: ``Difficulty``, ``Terrain``, ``Size``, ``Recent logs`` ... (afterwards whatever you like)
3. Open any list containing offline stored caches with at least 8 logs (current number of smileys to be displayed)
4. Check the row of smileys ...

- When the smiley row does not fit into it's configured line completely, this line will show no smileys at all,
   and the second line just shows *some* smileys.
- The few ones that will be shown, are the ones from the *end* of the "original" smiley row
   (i. e. if there are just two smileys shown, these are smiley # 7 and # 8 out of 8)
- Seems as if the *more* free space is in the "initial" line (where the first smileys should have gone to),
   the *less* smileys are shown in the second line.

The "row of smileys" should ...
- *either* be wrapped correctly at the end of the line ...
- *or* go completely to the next line (which I BTW would call the better solution)

### Reproducible

Yes

### c:geo Version

2023.04.03-NB-22c32ee

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Mitigation for truncated smileys on first line when a smiley line wraps
(see https://github.com/cgeo/cgeo/issues/14163#issuecomment-1501177328 for details)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/Formatter.java
+++ b/main/src/main/java/cgeo/geocaching/utils/Formatter.java
@@ -14,9 +14,12 @@ import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.extension.PocketQueryHistory;
 
 import android.content.Context;
+import android.os.Build;
+import android.text.Spannable;
 import android.text.SpannableString;
 import android.text.SpannableStringBuilder;
 import android.text.format.DateUtils;
+import android.text.style.ForegroundColorSpan;
 import android.text.style.ImageSpan;
 import static android.text.format.DateUtils.MINUTE_IN_MILLIS;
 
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

PR #14181, opened 2023-04-10
Enable coreLibraryDesugaring to provide access to modern java APIs on old API levels

fm-sys:
> `coreLibraryDesugaringEnabled` is a feature in Android's build system that enables developers to use modern language features and APIs on older Android versions. Some benefits of using this feature include the ability to write more expressive, modular, and maintainable code, as well as leveraging the latest language features available in Java. However, there are also potential downsides to using `coreLibraryDesugaringEnabled`, including performance overhead, increased APK size, compatibility issues, and debugging issues. Developers should consider these pros and cons before deciding to use this feature in their Android projects.

___

Some cool Java features like `Streams` are only natively supported starting with API 24 while we are still targeting API 21. However, google provides a desugaring compatibility library which we could use. However, we need to weigh the potential benefits against the potential drawbacks and consider whether it is the right choice for us.

More details:
https://developer.android.com/studio/write/java8-support

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -294,7 +294,7 @@ android {
 
 dependencies {
     // desugaring of Java 8 APIs
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.2.2'
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
 
     // Apache Commons
     implementation 'org.apache.commons:commons-collections4:4.4'
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -24,7 +24,10 @@ android {
 
 
     compileOptions {
-        // use the diamond operator and some other goodies in Android
+        // https://developer.android.com/studio/write/java8-support
+        // enable Java 8 APIs like Streams on older Android versions
+        coreLibraryDesugaringEnabled true
+        // use the diamond operator, lambdas and some other Java 8 language features in Android
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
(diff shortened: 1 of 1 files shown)


main/build.gradle
```diff
@@ -290,6 +293,9 @@ android {
 }
 
 dependencies {
+    // desugaring of Java 8 APIs
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.2.2'
```
fm-sys, line 297: Version 1.2.2 is outdated meanwhile... Should be updated if we agree on it

bekuno: @fm-sys Please update the PR to the latest lib.
Then we can compare the apk size.

bekuno:
lets check the PR with the current outdated lib.

retest this please

moving-bits:
build with 1.2.2 is 26.23 MB ([cgeo-debug.apk](https://ci.cgeo.org/job/cgeo-CI_PR-build/6438/artifact/main/build/outputs/apk/basic/debug/cgeo-debug.apk))

main/build.gradle
moving-bits: done for testing purposes

mortenbruhn:
This seems like a good trade-off.

Having only [recently](https://github.com/cgeo/cgeo/pull/14777) entered development on c:geo, I found it incredibly frustrating to not have access to stuff that I've been accustomed to having when doing Java development for at least 10 years now. There are even [features that were introduced in Java API 1.7](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#group-java.lang.String-) which are not in Android API 21.

The size increase of the APK is miniscule if the stats that @moving-bits are accurate. 

If there are also no significant performance penalties, I think it is well worth including this -- perhaps even a better option than bumping the minimal Android API of c:geo, since supporting old APIs longer will allow people to use their phones longer to geocache, which in turn positively impacts the environment and global warming, as well as making the pastime we all love more accessible to the masses.

eddiemuc:
I see a lot of thumbs up for this and no thumbs down yet.

Let's enable this in master to try it out. I suggest that developers use new features only sparsely for now so we have the chance to roll back in case we discover unwanted side effects.

PR #14199, opened 2023-04-15
Catch NPE on drawing position marker in NewMap

moving-bits:
## Description
Catch NPE on drawing position marker, when marker is not available
(This is currently the most prominent crash on Play statistics for release 2023.03.25 (in terms of "user affected"))

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
+++ b/main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
@@ -77,19 +77,20 @@ public class PositionLayer extends Layer {
             arrowNative = ImageUtils.convertToBitmap(ResourcesCompat.getDrawable(CgeoApplication.getInstance().getResources(), R.drawable.my_location_chevron, null));
             rotateArrow();
         }
-        if (arrow != null) {
+        final Bitmap localArrow = arrow;
+        if (localArrow != null) {
             final int left = centerX - widthArrowHalf;
             final int top = centerY - heightArrowHalf;
-            final int right = left + this.arrow.getWidth();
-            final int bottom = top + this.arrow.getHeight();
+            final int right = left + localArrow.getWidth();
+            final int bottom = top + localArrow.getHeight();
             final Rectangle bitmapRectangle = new Rectangle(left, top, right, bottom);
             final Rectangle canvasRectangle = new Rectangle(0, 0, canvas.getWidth(), canvas.getHeight());
             if (!canvasRectangle.intersects(bitmapRectangle)) {
                 return;
             }
-            canvas.drawBitmap(arrow, left, top);
+            canvas.drawBitmap(localArrow, left, top);
         } else {
-            Log.d("PositionLayer.draw: arrow=null, arrowNative=" + arrowNative);
+            Log.d("PositionLayer.draw: localArrow=null, arrowNative=" + arrowNative);
         }
     }
 


eddiemuc:
There is a check whether arrow is null. Why did you remove this check?

And more important: why is there a NPE when arrow is checked for null? Apparently it is not the arrow object causing this?

moving-bits:
> There is a check whether arrow is null. Why did you remove this check?

This would be included in try try/catch block as well.

> And more important: why is there a NPE when arrow is checked for null? Apparently it is not the arrow object causing this?

Play stats lists errors like this:
```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'int android.graphics.Bitmap.getWidth()' on a null object reference
  at org.mapsforge.map.android.graphics.AndroidBitmap.getWidth (AndroidBitmap.java:114)
  at cgeo.geocaching.maps.mapsforge.v6.layers.PositionLayer.draw (PositionLayer.java:83)
  at org.mapsforge.map.layer.LayerManager.doWork (LayerManager.java:95)
  at org.mapsforge.map.util.PausableThread.run (PausableThread.java:149)
```

The source line is within the `if (arrow != null)` block, eg: `final int right = left + this.arrow.getWidth();`, so `arrow` seems to be null, although it's not null a few lines before, so I guess it's a race condition. I could either add a locking mechanism (causing some overhead and possible deadlock situations) or simply catch and ignore the error (as the position marker is not that important and gets redrawn frequently), which is what I chose here.

main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
```diff
@@ -77,7 +77,7 @@ public void draw(final BoundingBox boundingBox, final byte zoomLevel, final Canv
             arrowNative = ImageUtils.convertToBitmap(ResourcesCompat.getDrawable(CgeoApplication.getInstance().getResources(), R.drawable.my_location_chevron, null));
             rotateArrow();
         }
-        if (arrow != null) {
```
eddiemuc, line 80: I suggest an alternative approach:
- before line 89, store arrow in a local variable (e.g. `final Bitmap lArrow = this.arrow;`
- then use this local variable instead (e.g. line 80 would read `if (lArrow != null) {`

That way, race conditions would not be able to set lArrow to null in-between it's usage.

eddiemuc:
Understood. I suggested an alternative approach as review comment

main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
moving-bits: Thanks for the proposal. How would this approach compare resource & performance-wise?

eddiemuc: > Thanks for the proposal. How would this approach compare resource & performance-wise?

I don't think resource/performance is really an impact in this case either way, since the majority of CPU in this class is probably used by recalculating rotation of the image over and over again. Compared to this, everything we do here is probably neglectable.

But if we go into detail analysis, I would assume that the approach with local var is more performant. It doesn't create any new objects (just a pointer) and the arrow bitmap might be hold into memory a few milliseconds longer. On the other hand you save the creation and catching of an exception, and creation of exceptions is expensive in terms of performance. Which is ok since they are exceptions and should not be used for normal control flow of application logic. 

The other aspect is that you should normally not catch runtime exceptions like NPE anyway. If necessary to fix urgent production issues that's ok imho, but it should only ever be a last resort. Because if code generates NPEs then something is not right with it.

moving-bits: Thanks for the detailed explanation. Just for my understanding (and sorry if this might be a dumb question):

It looks to me that several instances of `draw()` get called in parallel, thus leading to race conditions if one instance has done the `if (arrow != null)` check already, but not yet the actual drawing, and before that the other instance overwrites/deletes `arrow`, thus the first instance runs into the NPE when trying to do their calculation or the actual drawing.

Now, if the first instance creates a pointer to `arrow`, then the other instance modifies/deletes `arrow`, what will the pointer of the first instance point to? Will it create a local copy ("copy on modify") before the other instance modifies the object, or will it still point to the same memory space the other instance is modifying?

eddiemuc: All variables in Java store references ("pointers") with the exception of primitive types. In our case this would mean that
* Start point is that `this.arrow` points to a valid `Bitmap` containing the arrow image
* One thread calls `draw()`, the reference to this bitmap is stored in local `lArrow` variable
* Another thread somehow sets `this.arrow` to `null`. Which means `this.arrow` doesn't point to the bitmap any more. If this would have been the last reference to the bitmap, then Garbage Collector would kick in and delete the bitmap eventually. But in our case we would still have our local variable `lArrow` pointing to it so GC leaves it in memory.
* Now our first thread is doing the drawing. The method is finished at some point which makes our local variable `lArrow` invalid. If now this was the last reference to the actual bitmap, then at this point GC would delete it from memory

So there will not be any "implicit" copying of the actual bitmap, it is just "hold" into memory a few more millisceonds

Having said that: I just looked into code, and it really really puzzles me how this.arrow can ever become null again after it was non-null for the first time. Because the only place where something is assigned to arrow is line 106, and there another object (not null) is assigned to it...:

`arrow = AndroidGraphicFactory.convertToBitmap(tmpArrow);`

moving-bits: Thanks for the detailed explanation. I'm always learning something new...
Have adapted the PR accordingly.

Issue #14161, opened 2023-04-02
Map not centered


fm-sys:
BTW: I just noticed, that the position marker does not even get centered on the map!
It's in fact centered relative to the entire display, "wasting" a lot of space while heading north.
In fact over here the position marker ends up about 7 mm (resp. ~14 %) too high - which I would consider a *real* issue.

IOW: Even in the one and only scenario, where hiding the nav bar would give add. 9 mm of map space (i. e. driving north, exactly), we would achieve nearly the same map info gain by just correcting the centered position to the "map center".
This, BTW, is the same in landscape mode, where it of cause is even worse, as there are just about 18 mm of free space between centered map position and header bar, anyway.

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/14066#issuecomment-1482954746_

fm-sys:
A picture tells more than a thousand words:

![grafik](https://user-images.githubusercontent.com/64581222/229361426-9238de38-d4a6-4b15-93eb-7c9166436baa.png)

MagpieFourtyTwo:
Same applies to portrait mode, of course.

fm-sys:
> Same applies to portrait mode, of course.

Sure, thanks for mentioning it explicitly. I did chose landscape to make the problem even more visible ;-)

Lineflyer:
Indeed...I never explicitly noticed. Maybe the bug was introduced some year(s) ago when we renovated all the action bar stuff.
For north bound map it should be centered, for rotating maps we could think about a 2/3 approach to allow more map visible in moving direction.

moving-bits:
> Maybe the bug was introduced some year(s) ago when we renovated all the action bar stuff. 

It's a side-effect on the new "tap hides action bar" functionality. To avoid the map "hopping around" (resizing, recentering, and therefore changing position) the map needs to be sized to the maximum size required. Action bar etc. will lay on top of parts of the map until when in use and not hidden.

Therefore we will calculate a new map position which will fit the default (action bar shown) in both portrait and landscape modes.

MagpieFourtyTwo:
>Therefore we will calculate a new map position which will fit the default (action bar shown) in both portrait and landscape modes.

IMHO, one can live just fine with this solution. Although it may not be perfect, it's already way better than the current implementation, cause after comparing the two "worst cases" (landscape mode, non rotating map), as is ...
- driving north with the "old" centered position (left cross in @fm-sys' screenshot) vs.
[...]

PR #14200, opened 2023-04-15
Keep old mapsize when 'tap to hide actionbar' is disabled

moving-bits:
## Description
Implements the approach proposed in https://github.com/cgeo/cgeo/issues/14161#issuecomment-1508089543:
Do not move the map view beneath the action bar, when "hide on tap" is disabled.

Additionally extracts the map hiding methods into their on class (instead of being buried in `FilterUtils`) and uncouples it from `FilterUtils.initializeFilterBar()`.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/HideActionBarUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/HideActionBarUtils.java
@@ -1,0 +1,73 @@
+package cgeo.geocaching.utils;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.AbstractBottomNavigationActivity;
+import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.ui.ViewUtils;
+
+import android.app.Activity;
+import android.view.View;
+
+import androidx.annotation.LayoutRes;
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.ActionBar;
+
+public class HideActionBarUtils {
+
+    private HideActionBarUtils() {
+        // utility class
+    }
+
+    /** use this instead of AbstractBootmNavigationActivity.setContentView for being able to use a action bar toogle */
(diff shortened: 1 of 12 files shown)


fm-sys:
I was too late to comment on the original PR:

The method `AbstractBottomNavigationActivity.setStableLayout()` doesn't fit at its current place, as it would be potentially relevant for activities without BottomNavigation as well. Please move it to `ActivityMixin`.

main/src/main/java/cgeo/geocaching/utils/HideActionBarUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class HideActionBarUtils {
+
+    private HideActionBarUtils() {
+        // utility class
+    }
+
+    public static void showActionBarSpacer(@NonNull final Activity activity, final boolean showSpacer) {
+        activity.findViewById(R.id.hideActionBarSpacer).setVisibility(showSpacer ? View.VISIBLE : View.GONE);
+        if (!Settings.getMapActionbarAutohide()) {
```
fm-sys, line 22: Wouldn't it be easier to just not call `setStableLayout` if MapActionbarAutohide is disabled, instead of adding a spacer to all views?

fm-sys: And even if the spacer is still necessary, it would need to be placed at a different location

Android 7 - Navigation rail behind actionbar:

grafik

moving-bits: > Wouldn't it be easier to just not call `setStableLayout` if MapActionbarAutohide is disabled, instead of adding a spacer to all views?

Good idea, I'd tested getting rid of `setStableLayout()` at all before, but that didn't work. Removing it for "tap to hide actionbar disabled" does work, though. PR is adapted.

> And even if the spacer is still necessary, it would need to be placed at a different location
> Android 7 - Navigation rail behind actionbar:

Does happen without the spacer as well. Will try to adapt this PR or create a separate for it later.

Issue #13176, opened 2022-06-19
Add "prepend to route" option for geocaches


ztNFny:
#13109 added the option to add a point either to the end or *beginning* of an existing route.
The same should be available for caches.

===========
> Do you plan to add this functionality for caches/WPs as well (not only for empty locations)? 

Yes. But not sure about the UI part yet. It was easy enough to add here so I started with this. I don't want to make adding caches to routes harder, so need to investigate some possibilities here. Maybe have it in cache pop-up menu ...

_Originally posted by @ztNFny in https://github.com/cgeo/cgeo/issues/13109#issuecomment-1154256074_

Ideas:
- long-click on "set as target" button in cache popup - Pro: Quick to access - Con: hard to discover
- menu item in cache popup - Pro: Easy to discover - Con: Many clicks, menu is already quite crowded
- show a popup on long-pressing a cache - Pro: Easiest to discover - Con: Complicates adding caches to a route (so much that imho this is a no-go)

ztNFny:
Ideas:
- long-click on "set as target" button in cache popup - Pro: Quick to access - Con: hard to discover
- menu item in cache popup - Pro: Easy to discover - Con: Many clicks, menu is already quite crowded
- show a popup on long-pressing a cache - Pro: Easiest to discover - Con: Complicates adding caches to a route (so much that imho this is a no-go)

moving-bits:
As discussed in today's dev meeting: Long tap should open a context menu, as this is consistent with adding "free points" and also prevents users from accidentally adding caches to an individual route (had several issues on support about this).

ztNFny:
> As discussed in today's dev meeting: Long tap should open a context menu, as this is consistent with adding "free points" and also prevents users from accidentally adding caches to an individual route (had several issues on support about this).

Would be good if the old behavior (add cacje to route on long-tap) can be restored with a setting as it's a lot faster than the new proposed one (1 vs 3 taps)

moving-bits:
Why would it be three taps? First (long) tap for opening the popup, second tap to select "append to route"?

Regarding a setting: I guess I would need to see it in practice to be able to say if the new solution would be too annoying for appending to a route or not.

Lineflyer:
I would prefer consistency of map behavior over faster (but somehow unexpected) action. And we should please avoid more settings whenever possible.

Lineflyer:
Just an example of support cases, which come in quite frequently on our mail and on the user groups:

[...]

PR #14221, opened 2023-04-23
NewMap: Add long-tap context menu for caches/waypoints (rel. to #13176)

moving-bits:
## Description
Adds a "long tap" context menu for caches/waypoints on map, providing (up to) the following items:
- append to route
- prepend to route
- remove from route
- remove from route end
- remove from route start

This allows to add a cache/waypoint multiple times to a route, specifically as both start and end point to create a round-trip.
Long tap on a cache/waypoint will now always trigger this context menu, no longer auto-toggle route item (as discussed in related issue among others.)

Implemented for `NewMap` only currently, to gather some feedback.

|Screenshot 1|Screenshot 2|
|---|---|
|image|image|

Screenshot 1: Cache is not yet part of the route, so allow to append or prepend to route.
Screenshot 2: Cache is part of the route (at the end), so allow prepend or remove.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
@@ -585,6 +585,17 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
         overlayPositionAndScale.repaintRequired();
     }
 
+    public void handleCacheWaypointLongTap(final IWaypoint item, final int tapX, final int tapY) {
+        final RouteItem routeItem = new RouteItem(item);
+        if (MapUtils.isPartOfRoute(routeItem, individualRoute)) {
+            MapUtils.createCacheWaypointLongClickPopupMenu(activity, routeItem, tapX, tapY, individualRoute, overlayPositionAndScale, this::updateRouteTrackButtonVisibility)
+//                    .setOnDismissListener(menu -> tapHandlerLayer.resetLongTapLatLong())
+                    .show();
+        } else {
+            toggleRouteItem(item);
+        }
+    }
+
     public void toggleRouteItem(final Geopoint geopoint) {
         if (individualRoute == null) {
             individualRoute = new IndividualRoute(this::setNavigationTargetFromIndividualRoute);
--- a/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
@@ -18,6 +18,7 @@ import cgeo.geocaching.models.Download;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.IndividualRoute;
 import cgeo.geocaching.models.RouteItem;
+import cgeo.geocaching.models.RouteSegment;
 import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.settings.Settings;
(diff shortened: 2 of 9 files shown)


moving-bits:
retest this please

fm-sys:
Nice, although it requires one more click. Imagine adding 50 caches of a power trail to individual route. 50 vs 100 clicks can make a huge difference. 

Maybe we can show the popup only if the cache is already part of the route? 

|_Selection options_|
|-|
Remove from {start/pos x/end} (show multiple times for each time it is part of the route) 
Add to start
Add to end

moving-bits:
> Maybe we can show the popup only if the cache is already part of the route?

I'm a bit undecided about this one. For sake of consistency one could argue that it should be always a long-tap (also for non cache/waypoint-items of an individual route). This would also prevent users from creating an individual route unknowingly (we have this on support regularly).

OTOH it sounds reasonable and intuitive enough to give it a try.

@cgeo/team: More opinions?

> Imagine adding 50 caches of a power trail to individual route

For this I could image a cache list function: "Add all (selected) caches to individual route" and then do a route optimization.

fm-sys:
I guess it wouldn't make much difference in terms of complexity and would at least be consistent. Although I agree it's probably a rare case...

moving-bits:
Added item removal from specific position in route:

|(1)|(2)|(3)|
|---|---|---|
|image|image|image|

Created a star-shaped route, always returning to the center, and finally removed the center from start of route to better explain different scenarios:
1. Tapping on center point, it lists "Prepend to route" and the several positions this point is part of the route.
2. This "extension" is at position 2 of the route and can be appended or prepended to it (or removed from that position).
3. The starting point can be either removed (from start) or appended to the route.

Order of possible menu items is:
- Append to route (as default action)
- Prepend to route
- Remove from start
- Remove from position xxx
- Remove from end

moving-bits:
@fm-sys @eddiemuc 
I could need some help here: I'm trying to implement the new long tap functions for individual routes for UnifiedMap. The function itself is working, but the route line is not updated on the map. How can I trigger the `IndividualRouteLayer` to update its contents?

(If this is an explicit method to call, I would need to wrap it in a method implementing the interface `IndividualRoute.UpdateIndividualRoute`, and pass this as parameter to the call of `MapUtils.createCacheWaypointLongClickPopupMenu()` in `UnifiedMapActivity.handleTap()`.)

fm-sys:
> Order of possible menu items is:
> 
>     Append to route (as default action)
>     Prepend to route
>     Remove from start
>     Remove from position xxx
>     Remove from end
>

One could argue whether "remove" should be listed above add elements, so it is nearer to reach after long click. 

> @fm-sys @eddiemuc
> I could need some help here: I'm trying to implement the new long tap functions for individual routes for UnifiedMap. The function itself is working, but the route line is not updated on the map. How can I trigger the `IndividualRouteLayer` to update its contents?

No need to call the layer directly, but you need to inform the view model about the changed data. This can be done e.g. via `UnifiedMapViewModel.getIndividualRoute().notifyDataChanged()`. The ViewModel can be accessed from everywhere via `new ViewModelProvider(activity).get(UnifiedMapViewModel.class);`, or some classes hold the view model instance as private field.

moving-bits:
Thanks @fm-sys for the pointer, it's working now.
Have swapped "add" and "remove" positions as well.

moving-bits:
retest this please

PR #14225, opened 2023-04-24
Add 'refresh and keep list' to cache download options (fix #14190)

### LINKED ISSUE #14190: Refresh All Caches Shown Inside "Show As List"
### Tell us your idea!

A few days ago I planned to go into the wild, upfront I opened the target area to see which caches are there.

Cause I am keeping a lot of my extended Homezone offline inside cgeo, I selected the show as list entry for the caches being displayed on the live map to update them, cause the local data was really outdated for that area.

But there was no "refresh all caches" entry inside the manage caches menu available. Only the option to store all shown caches to a new list has been provided. 

Offer for the "Show As List"-list entries also the refresh all caches option. 

### Is there an alternative?

Store all caches again into a temporary list

### Additional Information

_No response_

moving-bits:
## Description
Add a "refresh caches, but keep their assigned lists" option to cache download options available in map and "show as list":

image.image


fm-sys:
Just a small idea, maybe name it "refresh but don't change lists" or  "refresh without add to list" to make it more clear...

moving-bits:
changed it a bit:

image

PR #14231, opened 2023-04-29
Fix lint: remove two ununsed icons

moving-bits:
## Description
Remove two unused icons (probably coming from early testings for bottom navigation)


moving-bits:
retest this please

PR #14237, opened 2023-04-30
Extract LogCacheActivity.Poster task

moving-bits:
## Description
Extract LogCache.Poster task to prevent leaks


moving-bits:
retest this please

PR #14262, opened 2023-05-13
Add more number emojis (separate category)

moving-bits:
## Description
We have users on support asking for more choices to mark cache sequences with symbols. This PR adds circled and parenthesized numbers as addition to the existing plain numbers, and puts all numbers into a separate category:

image


ztNFny:
As those emojis are (usually) white they will be invisibile on the cache marker.
Xiaomi devices have that issue with the current number emjois that are non-white in pretty much every one but their fonts.

fm-sys:
The newly added emojis are monochrome simbols, means they are colored in text colors. If the current text color doesn't fit, we can easily change it.

BTW, would it make sense to give all emojis a while background if they are used as list marker?

ztNFny:
> BTW, would it make sense to give all emojis a while background if they are used as list marker?

That will end up looking very ugly. You'd have to put a white square behind everything.
Users are able to select an icon that's visible to them.

moving-bits:
> As those emojis are (usually) white they will be invisibile on the cache marker. Xiaomi devices have that issue with the current number emjois that are non-white in pretty much every one but their fonts.

This is how it looks like in the emulator:

image

Intended use is for cache icons, less for cache list markers - they would probably too small anyway for that use case.

moving-bits:
retest this please

PR #14270, opened 2023-05-17
Support for cache recommendation for opencaching based services

### LINKED ISSUE #3879: Support favorites on oc.de
I do not know whether this is possible via the API:
Currently OC-caches show the amount of favorties but the user is unsable to set the cache as favorite while logging or afterwards.

mzylowski:
## Description
Hello, I noticed that there is no possibility to log a cache with recommendation (aka GC favorite) on opencaching.pl site. I decided to implement this small feature. I'm not advanced in Java/android developmen, so reach me if I did something wrong.

## Related issues
Fixes #3879

## Additional context
Change bases on taking `rcmds_left` field from user query from OKAPI: https://opencaching.pl/okapi/services/users/user.html

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
@@ -357,7 +357,7 @@ public class LogCacheActivity extends AbstractLoggingActivity {
             final int favoritePoints = ((ILoggingWithFavorites) loggingManager).getFavoritePoints();
             if (favoritePoints > 0) {
                 binding.favoriteCheck.setVisibility(View.VISIBLE);
-                binding.favoriteCheck.setText(res.getQuantityString(R.plurals.fav_points_remaining, favoritePoints, favoritePoints));
+                binding.favoriteCheck.setText(res.getQuantityString(((ILoggingWithFavorites) loggingManager).getFavoriteCheckboxText(), favoritePoints, favoritePoints));
             }
         } else {
             binding.favoriteCheck.setVisibility(View.GONE);
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -2411,6 +2411,14 @@
     <string name="command_rename_list_result">Renamed list</string>
     <string name="command_delete_list_result">Deleted list</string>
     <string name="command_unique_list_result">Deleted caches from other lists</string>
+    <plurals name="fav_points_remaining_oc">
+        <item quantity="zero">Recommend this cache (%d remaining)</item>
+        <item quantity="one">Recommend this cache (%d remaining)</item>
+        <item quantity="two">Recommend this cache (%d remaining)</item>
+        <item quantity="few">Recommend this cache (%d remaining)</item>
+        <item quantity="many">Recommend this cache (%d remaining)</item>
+        <item quantity="other">Recommend this cache (%d remaining)</item>
+    </plurals>
     <plurals name="fav_points_remaining">
         <item quantity="zero">Add this cache to favorites (%d remaining)</item>
         <item quantity="one">Add this cache to favorites (%d remaining)</item>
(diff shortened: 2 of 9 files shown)


moving-bits:
ok to test

bekuno:
Retest this please

main/src/main/res/values/strings.xml
```diff
@@ -2411,6 +2411,14 @@
     <string name="command_rename_list_result">Renamed list</string>
     <string name="command_delete_list_result">Deleted list</string>
     <string name="command_unique_list_result">Deleted caches from other lists</string>
+    <plurals name="fav_points_remaining_oc">
+        <item quantity="zero">I want to recommend this cache (%d remaining)</item>
```
Lineflyer, line 2415: Maybe a better/shorter wording would be `Recommend this cache (%d remaining)`. This is also similar to the favorite wording.

Lineflyer:
@mzylowski 
Thanks for your contribution!
You may want to check the warnings from the cgeo CI default build result and correc them.

Not sure about the Codacy status...it looks somehow broken. I will try to retrigger it on Codacy..maybe some kind of permission problem.

main/src/main/java/cgeo/geocaching/connector/oc/OCUserInfo.java
```diff
@@ -0,0 +1,19 @@
+package cgeo.geocaching.connector.oc;
+
+import cgeo.geocaching.connector.UserInfo;
+
+import androidx.annotation.NonNull;
+
+public class OCUserInfo extends UserInfo {
```
eddiemuc, line 7: No need for an OC-specific extension imho. I would recommend to extend UserInfo with an (optional) field `remainingFavoritePoints ` instead (with default -1 if user or platform has no fav points)

main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -357,7 +358,11 @@ private void initializeFavoriteCheck() {
             final int favoritePoints = ((ILoggingWithFavorites) loggingManager).getFavoritePoints();
             if (favoritePoints > 0) {
                 binding.favoriteCheck.setVisibility(View.VISIBLE);
-                binding.favoriteCheck.setText(res.getQuantityString(R.plurals.fav_points_remaining, favoritePoints, favoritePoints));
+                if (connector instanceof OCApiConnector) {
```
eddiemuc, line 361: Please don't put any  logic here which depends on the used connector. The whole idea of the connector concept is to keep the differences of the connected platforms out of the common activities. 

If I understand correctly you want to differentiate the wording "favorite" (gc) from "recommendation" (oc). If you really feel the need to do so please abstract it out towards the IFavoriteCapability interface (e.g. add a method which gives the correct resId for the word to use in display).  

Personally I feel we should have the same wording in c:geo across platforms/connectors , means we should use "Favorite" for all platforms, but that's debatable of course (hence the suggestion to abstract it out to IFavoriteCapability interface). What is imho not debatable however is that we use the same wording in code everywhere - and here we use "Favorite" (eg. IFavoriteCapability and other classes) and not "recommendation".

eddiemuc (commented):
Thanks a lot for your contribution! I left a few comments. Codacy/lint needs to be checked as well of course.

main/src/main/java/cgeo/geocaching/connector/oc/OCUserInfo.java
mzylowski: I fixed that by adding additional constructor in UserInfo class

main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
mzylowski: I prepared a new method in ILoggingWithFavorites interface (to be consistent with already implemented ``` ((ILoggingWithFavorites) loggingManager).getFavoritePoints();```.

From technical point of view I fully understand your concerns about unified wording. For a geocacher who caching only opencaching.pl I wonder 'favorite' can be a bit missleading, especially for fresh players.
I prepared a change where I added oc-only wording but with suggestion of @Lineflyer. Of course if cgeo lead programmers have strong option about this kind of oddities I can change my PR to use 'favorite' text also for OC.

eddiemuc:
Lgtm, thanks a lot for your contribution!

Issue #13690, opened 2022-11-19
Find better handling of login attempts with wrong credentials


Lineflyer:
### Tell us your idea!

As also discussed in our November dev meeting:

I was wondering, why there are so many users contacting us because of the "Captcha required" problem. Previously this mostly occurred if some anonymizer services (like TOR) or public WiFi were involved and thus geocaching.com enforcing the Captcha due to the used IP-address.

However since some time the Captcha is also triggered by consecutive login attempts with wrong password (tested that).
This implies, that c:geo itself can trigger the Captcha enforcement but consecutively trying to login with wrong credentials.

### What solution would you suggest?

If a login attempt fails with the specific reason of wrong username/password, c:geo should refrain from making any more login attempts and inform the user proactively to check his password and login manually via Start - Menu - Services - geoacaching.com.
The second part could be achieved by providing a more specific login failed warning on the home screen as we do today.

Implementation might however be quite difficult, as we AFAIK have many trigger points in our code triggering a login if we detect during an network operation, that the login might no longer be valid. Not sure about that though.

### Is there an alternative?

We already inform the user about a login failure using a home screen message.  We could/should at least add the error reason there to make users aware, that action is needed.

PR #14277, opened 2023-05-20
Print GC login error message on home screen (rel. to #13690)

moving-bits:
## Description
In case of login error at GC print the error message we get from the server to the home screen:

image


moving-bits:
retest this please

Issue #14290, opened 2023-05-24
load GC PQ preview from map instead of nearest.aspx


ztNFny:
### Tell us your idea!

For previewing PQs we're currently using https://www.geocaching.com/seek/nearest.aspx?pq= which is one of those old pages with lots of HTML, also limited to 20 results/page.

### What solution would you suggest?

There's also a "preview on map" option, loading to https://www.geocaching.com/map/default.aspx?pq= which loads the actual data as geojson from https://tiles01.geocaching.com/map.pq?pq=..&st=..
This is a "nice" data format (json) and seems to deliver up to 1000 items.

moving-bits:
Hasn't this been fixed with #14296? Then we could close here...

bekuno:
@ztNFny You did not use the autoclose function in your PR. For details see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue.

PR #14296, opened 2023-05-27
load PQ preview from map instead of legacy search results (#14290)

ztNFny:
For previewing PQs we're currently using https://www.geocaching.com/seek/nearest.aspx?pq= which is one of those old pages with lots of HTML, also limited to 20 results/page.
There's also a "preview on map" option, loading to https://www.geocaching.com/map/default.aspx?pq= which loads the actual data as geojson from https://tiles01.geocaching.com/map.pq?pq=..&st=..
This is a "nice" data format (json) and seems to deliver up to 1000 items.

Cache information that can be obtained this way:
- name
- geocode
- type
- archived
- disabled
- coordinates
- found
- userModifiedCoordinates

Missing info compared to old search page:
- d/t rating
- placed date
- size
- owner name
- favorite count

I think the advantage of getting more than 20 caches more than makes up the "loss" in information about these caches, especially as that info can easily be retrieved by storing the caches.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -1826,7 +1826,8 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         final Intent cachesIntent = new Intent(context, CacheListActivity.class);
         Intents.putListType(cachesIntent, cacheListType);
         cachesIntent.putExtra(Intents.EXTRA_NAME, pocketQuery.getName());
-        cachesIntent.putExtra(Intents.EXTRA_POCKET_GUID, pocketQuery.getGuid());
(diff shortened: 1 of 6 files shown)


ztNFny:
retest this please

fm-sys:
> Missing info compared to old search page:
> 
>     * d/t rating
> 
>     * placed date
> 
>     * size
> 
>     * owner name
> 
>     * favorite count

favorite count would be interesting to have, everything else isn't displayed anyway. But I agree that the advantages outweigh the disadvantages.

ztNFny:
Retest this please

moving-bits:
Error is

```
16:41:44  /home/jenkins/slave/workspace/cgeo-CI_PR-build/main/src/main/java/cgeo/geocaching/loaders/PocketGeocacheListLoader.java:28: error: exception JsonProcessingException is never thrown in body of corresponding try statement> Task :main:compileBasicDebugJavaWithJavac
16:41:44  
16:41:44              } catch (JsonProcessingException e) {
```

PR #14298, opened 2023-05-28
use fragments in unified map (ALPHA STATE!)

fm-sys:
rel. to #14113

Took much longer than I would have expected... But finally I have a first version ready.

Core changes:
- GM and VTM are now wrapped into fragments
    - they have access to the parent activity (no need for holding activity references)
    - native lifecycle handling (fragments receive lyfecycle events like onResume, onPause, onDestroy automatically)
- Separate map specific layout from other map activity UI elements
    - filterbar, track/route, target view, etc are not map implementation specific and are therefore handled in `UnifiedMapActivity`
    - custom zoom buttons / custom compass rose are VTM specific and are therefore completely handled inside the VTM fragment
- When changing map provider, not the whole activity layout gets replaced but only the mapview inside the fragment container
- Map instances are no longer stored as static singleton pattern, which avoids memory leaks

Disclaimer: Refactoring all this was quite a complex task. I not yet found time to re-enable all features which were already implemented. Therefore, some code parts are currently commented out. Furthermore, I'm quite sure I introduced some bugs...

TODO list (no guarantee that it's complete):

- [x] Re-enable Reset FollowMyLocation on map drag
- [ ] Re-enable Tap handling
- [ ] Re-enable actionbar hiding
- [ ] Re-enable LoadInBackgroundHandler / MapsforgeGeoitemLayer
- [x] Re-enable Show/Hide Progressbar
- [ ] Re-enable distance view 
- [ ] Refactor MapsforgePositionLayer to use geoitemlayer


fm-sys:
retest this please 

_CI build was failing, but it's most probably unrelated to the PR. This comment will trigger a rebuild._

fm-sys:
Lets merge as discussed in our dev meeting. I'll move the todo list to https://github.com/cgeo/cgeo/issues/14113

PR #14312, opened 2023-06-07
Avoid NPE in PositionLayer.draw (fix #14295)

### LINKED ISSUE #14295: NPE in Mapsforge PositionLayer.draw
Reported by Play Store for current release 2023.05.15:

```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'int android.graphics.Bitmap.getWidth()' on a null object reference
  at org.mapsforge.map.android.graphics.AndroidBitmap.getWidth (AndroidBitmap.java:114)
  at cgeo.geocaching.maps.mapsforge.v6.layers.PositionLayer.draw (PositionLayer.java:84)
  at org.mapsforge.map.layer.LayerManager.doWork (LayerManager.java:95)
  at org.mapsforge.map.util.PausableThread.run (PausableThread.java:149)
```

moving-bits:
## Description
Mapsforge's AndroidBitmap can be non-null, but the referenced bitmap being destroyed already, still leading to an NPE as mentioned in the referenced issue.
This PR adds a check for `isDestroyed()`.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
+++ b/main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/layers/PositionLayer.java
@@ -78,7 +78,7 @@ public class PositionLayer extends Layer {
             rotateArrow();
         }
         final Bitmap localArrow = arrow;
-        if (localArrow != null) {
+        if (localArrow != null && !localArrow.isDestroyed()) {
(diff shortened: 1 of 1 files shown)


eddiemuc:
The question here is rather: why is the bitmap destroyed when we use it? I fear there's some higher level logic error in our code.

moving-bits:
retest this please

PR #14317, opened 2023-06-10
Make startscreen selectable (fix #14284)

### LINKED ISSUE #14284: Allow selecting startup tab (e.g. "map" instead of "start" tab)
### Tell us your idea!

When launching c:geo, the app always opens the start tab. I don't really like this 

(a) for privacy reasons because it means anyone sitting next to me on the train can see my GC.com username 
(b) this startup tab has no really useful functions, so displaying this first is unnecessarily inconvenient

An option to select any of the tabs as the startup default would be great. I would choose "map", other people may for example choose "nearby" or "search", depending on their use cases.

### Is there an alternative?

In the long run, one might consider getting rid of the not-very-useful "start" tab completely.

Or there might be a user option, which tabs to display at all (I never really use "nearby" either, so I would probably hide it, if that option existed)

### Additional Information

_No response_

moving-bits:
## Description
Let the user decide which page of bottom navigation they want to start with. Default is "HOME".

See settings => appearance => Default screen:

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/SplashActivity.java
+++ b/main/src/main/java/cgeo/geocaching/SplashActivity.java
@@ -31,7 +31,7 @@ public class SplashActivity extends AppCompatActivity {
                 intent.putExtra(InstallWizardActivity.BUNDLE_MODE, firstInstall ? InstallWizardActivity.WizardMode.WIZARDMODE_DEFAULT.id : InstallWizardActivity.WizardMode.WIZARDMODE_MIGRATION.id);
             } else {
                 // otherwise regular startup
-                intent = new Intent(this, MainActivity.class);
+                intent = Settings.getStartscreenIntent(this);
                 intent.putExtras(getIntent());
             }
             cLog.add("fi");
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -22,6 +22,7 @@ import cgeo.geocaching.ui.GeoItemSelectorUtils;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.utils.AndroidRxUtils;
 
+import android.app.Activity;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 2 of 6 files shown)


fm-sys:
Doesn't our main activity still contain some startup routines like DB optimization, map update, check for not yet imported map downloads, etc? (can't look it up currently) 

If yes, these should probably be moved into splash screen activity.

moving-bits:
Shall we move all those initialization routines to the `SplashActivity`, or shall the routing to a different opening screen then be moved to `MainActivity` after the initializations have finished?

SammysHP:
Interesting… My launcher directly starts the map activity. I hope this doesn't skip any maintenance tasks.

moving-bits:
> Interesting… My launcher directly starts the map activity. I hope this doesn't skip any maintenance tasks.

Yes, it currently does skip a lot of initialization/maintenance tasks, therefore @fm-sys's comment and my further enquiry.

moving-bits:
@fm-sys 

> Shall we move all those initialization routines to the `SplashActivity`, or shall the routing to a different opening screen then be moved to `MainActivity` after the initializations have finished?

Additional question: How does it work when staring c:geo via shortcuts on the mobile's home screen - will the init routines ever be called in this case? (Maybe we need to open a separate issue, if the answer is "no".)

fm-sys:
> @fm-sys
> 
> > Shall we move all those initialization routines to the `SplashActivity`, or shall the routing to a different opening screen then be moved to `MainActivity` after the initializations have finished?

IMHO going from SplashActivity to MainActivity and finally to the target activity is quite messy... So I would say it should go into SplashActivity

Another idea would be to execute it as a background task decoupled from any activity. Not sure if this would work in our case though.

> 
> Additional question: How does it work when staring c:geo via shortcuts on the mobile's home screen - will the init routines ever be called in this case? (Maybe we need to open a separate issue, if the answer is "no".)

The answer indeed is no. But it gets executed as soon as you will press "back" in the map (or a different) activity and are therefore directed to the HOME dashboard.

moving-bits:
One could also decouple it into a separate init method, which gets called by every startable activity, secured by a flag to make sure it gets run only once (like we have it for parts of MainActivity initialization). I guess at least some parts need to run before any other tasks can be executed, but need to cross-check this.

I'll open a separate issue for this.

Issue #13481, opened 2022-10-06
Non working found state information for ALC


Lineflyer:
### Describe your problem!

I open this as a bug as in first implementation this feature used to work:

For an unknown reason the ALC-API stopped returning the found state for adventure lab caches during the first implementation and test phase. Therefore the starting implementation for ALC in cgeo was released without that feature.

As this is however a major issue for users and was working initially we should clarify the reason and enable that feature if possible.
I will address this to the API owner.

@baiti 
I could need something like a technical two-liner describing how it initially worked and what has changed without prior notice (e.g. requests and responses OK/NOK).

### How to reproduce?

View ALC caches on map or perform coordinate/address search
 

### Actual result after these steps?

The ALC returned might contain ALC already found by the user but not marked as found in c:geo.

### Expected result after these steps?

Found ALC should be marked as found in search results/map

### Reproducible

Yes

### c:geo Version

2022.09.04

baiti:
a two-liner won't be sufficient, however let me try to be as brief as possible:

The API we're using is referred to as "the public API" (as opposed to "the partner API"). The public API has a variety of endpoints, some of which can be called without any authorization, some require a prior authorization. Authorized requests require a `consumer_key` which identifies the application (e.g. cgeo) making the request. The `consumer_key` has properties attached to it, which define what functionality is allowed to be used and what is not allowed to be used. A very important function is to use the `login` method of the API. By doing so, the subsequent requests are not only made with a `consumer_key` (which identifies the tool), but also with an `access_token` that identifies the individual users If we want to get the true found state of a lab cache's stages, we must tell the other side for which user we're making the request, which is only possible after we have successfully obtained an `access_token` by using the `login` method with a capable `consumer_key`.

The `consumer_key` that was assigned to cgeo initially supported the `login` method so we could acquire an `access_token` per individual user in order to obtain the "true found state". During the initial implementation and debug phase, even in the first release we made, we were able to fetch the true found state but after a short period of time, this functionality ceased to work. We could no longer use the `login` method so a specific capability must have been removed from the properties of our `consumer_key`.

I believe we have discussed that with the party who provided us the key but the capability was never enabled again. Rationale being that the public API is an unsupported API and we should use the partner API instead, which initially didn't support anything for Lab caches but learned how to do so at about just the time we had the discussions.

fm-sys:
[...]

PR #14318, opened 2023-06-10
#13481 AL found state

baiti:
## Description
This is an attempt to finally get the "true found state" for Adventure Lab Caches. The `SearchV4` method, when called with the user's `publicGuid` is supposed to return that state.

My tests on the emulator showed that indeed the true found state is reported once the `publicGuid` is provided to the `SearchV4` API method.

Please note: When explicitly fetching a Lab Cache by its ID, we still use the method [/api/Adventures/{id}](https://labs-api.geocaching.com/swagger/ui/index#!/Adventures/Adventures_Get) which does **not** return the true found state, however once the lab is on the map, it got there as a result of using the SearchV4 API, which now provides that state and it seems that state is cached when subsequently loading the cache detailed view. Please also note that this method: [/api/Adventures/{id}](https://labs-api.geocaching.com/swagger/ui/index#!/Adventures/Adventures_Get) is now deprecated, so we might want to look for an appropriate replacement.

## Related issues
#13481

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.connector.al;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.connector.gc.GCLogin;
 import cgeo.geocaching.enumerations.CacheSize;
 import cgeo.geocaching.enumerations.LoadFlags.SaveFlag;
 import cgeo.geocaching.enumerations.WaypointType;
@@ -201,6 +202,7 @@ final class ALApi {
         query.setTake(100);
         query.setRadiusInMeters(distanceInMeters);
         query.setRecentlyPublishedDays(daysSincePublish);
+        query.setCallingUserPublicGuid(GCLogin.getInstance().getPublicGuid());
         try {
             final Response response = apiPostRequest("SearchV4", headers, query, false).blockingGet();
             return importCachesFromJSON(response);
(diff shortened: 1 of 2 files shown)


ztNFny:
Will check tomorrow

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -351,6 +353,7 @@ private static Geocache parseCache(final JsonNode response) {
             cache.setArchived(response.get("IsArchived").asBoolean());
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
             // cache.setFound(parseCompletionStatus(response.get("CompletionStatus").asInt())); as soon as we're using active mode
+            cache.setFound(response.get("IsComplete").asBoolean());
```
fm-sys, line 356: Maybe you could remove the outdated comment about found state above.

fm-sys (commented):
Have you tested it with our API key as well? Excellent solution if it works! Probably one of the most requested features...

(Maybe I'll finally re-enable LCs again inside cgeo :))

Let's wait for @ztNFny's review and afterwards let's merge. Can't wait to test it in the nightlys!

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
baiti: Yes, actually wanted to but just forgot after all the excitment ;)

baiti:
> Have you tested it with our API key as well? Excellent solution if it works! Probably one of the most requested features...

I never used any other key so it must be ours for sure.

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
baiti: done

ztNFny (approved):

ztNFny:
looks good. Interesting that they chose to implement it like this, this way you can search any users finds - but there's still no way (with labs API) to search a users hides. I don't think this behaviour is intended, but as long as it works...

With this and "	"CompletionStatuses": [0]" in the query we could even implement online search by found status (0 = found, 1 = started, 2 = not found)

PR #14319, opened 2023-06-10
Add 'set as target' to map's long tap popup (fix #13539)

### LINKED ISSUE #13539: Add "Go to here" option to the long-press menu in the map
### Tell us your idea!

When long pressing the map, a popup menu appears with many useful options.

But one option is missing there: Selecting that point as the current destination for routing.
(Not as in "create a multi point route" but as in "a route from my current location to that point".)

Add this option to the popup menu.

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Adds "set as target" option to map's long tap menu:

image.image


moving-bits:
notes to myself: 
- needs to enable the track/route button
- and within this popup, "clear manual targets" button should be active

moving-bits:
retest this please

moving-bits:
retest this please

PR #14322, opened 2023-06-11
show the CacheWaypointLongClickPopupMenu always

ztNFny:
Currently on long-tapping a cache that's already part of a route a menu to add it again or remove it from the route is shown, but not on a cache that's not part of a route. That makes it impossible to add a cache to the start of a route (which was iirc the whole point of the original change).

This PR always shows the menu and introduces a setting to revert to the old menu-less way of toggling by long-click.

Alternative:
If it was possible to have a "long-tap" without menu and a "longer-tap" with menu that'd be the best solution imho.

To discuss:

- Imho long-tapping a cache that's already part of a route you'll (pretty much) always want to remove it, we shouldn't show the menu there.
- Connect logging a cache with route: Remove cache from route on saving/sending a log.


ztNFny:
@moving-bits let's discuss this in todays call

ztNFny:
PR updated to match our discussions in todays call.

moving-bits (approved):
LGTM, thanks for implementing

PR #14326, opened 2023-06-11
Add PMD integration for c:geo

### LINKED ISSUE #13990: NPath complexity threshold
We should seriously discuss whether the 200 threshold should be raised. It does not even nearly reflect the complexitiy of all the main classes of c:geo, and it is unrealistic to think that we will ever have time and passion to achieve this in the whole codebase.

_Originally posted by @eddiemuc in https://github.com/cgeo/cgeo/issues/13981#issuecomment-1434170545_

bekuno:
fixes #13990

Codacy checks Java code with checkstyle and PMD rules. (https://app.codacy.com/gh/cgeo/cgeo/patterns/list) It has 62 of 271 rules enabled for Java.

The NPathComplexity is a PMD rule. Therefore we need a PMD configuration in c:geo.

The PR adds integration with the app for c:geo (main) and the cgeo-contacts addin to be able to use a Gradle task. 
The version number is aligned to codacy.
And it adds a configuration file pmd.xml in the root folder. This file activates tree groups of tests (codestyle, design, errorprone). Tests with state error where excluded in the configuration. It would be useful later to activate some of the excluded tests and correct them. Maybe the way around to activate only a subset of dedicated rules is cleaner (see https://docs.pmd-code.org/latest/pmd_rules_java_bestpractices.html). The NPathComplexity is configured separately so that it only reports problems from the value 2000. This value is open for discussion.

If you want to use the Intellij PMD plugin, you have to set the path to the configuration field in the settings manually. It is not possible to have a pre-configured idea configuration like we have for checkstyle, as it uses a common configuration file misc.xml. One problem with the PMD plugin is that it is not possible to specify at the project level. Therefore, the sub-project directory for mapsme is excluded in the configuration file.

The gradle tasks are strange. They use the configuration from build.gradle and the configuration file, but ignore the exclude directives.

Let's see after creating the PR what results Codacy will produce.


moving-bits:
retest this please

bekuno:
retest this please

PR #14327, opened 2023-06-11
UnifiedMap - remove unnecessary getter methods from viewmodel

fm-sys:
All livedata objects are final, so it's save to mark them as public instead of providing getter methods. This makes the viewmodel clearer and makes clear that accessing the fields will never have side effects.

Edit: I know it's a common Java practice to encapsulate fields, even if they are final. I agree with that practice in general, however not in this specific case as the fields of the view model get accessed only by a very limited amount of classes. It's pretty much the same as placing the fields directly inside the activity class, no reason for getters here.


moving-bits:
retest this please

PR #14331, opened 2023-06-12
Do not display gc.com error while logging in

moving-bits:
## Description
As discussed in yesterday's team meeting: Do not display gc.com error while logging in


moving-bits:
retest this please

PR #14333, opened 2023-06-12
Decouple init/maintenance tasks from MainActivity (fix #14321)

### LINKED ISSUE #14321: Decouple init tasks and make sure they are run before any other tasks
Coming from the discussion in #14317:

c:geo's initialization and maintenance tasks are part of `MainActivity`, but there are "side-entries" to run c:geo, and #14317 will add another, leading to situations where our init & maintenance tasks are not run at all (if the user never opens the home activity) or only at a later time.

Change this behavior so those tasks are run before any other tasks, but make sure they are not executed every time an activity is started.

One solution could be to extract all init and maintenance tasks which are required for c:geo (and not specific to a certain activity) into a separate method that gets called by all startable activities, and secure this by a flag to make sure it's run only once.

"Startable activities" are (at least) those from bottom navigation:
- home screen
- cache lists ("stored")
- map
- search
- nearby

(feel free to extend this list if required)

Do other intents to be secured in this way as well? E. g. importing a GPX file, importing a downloaded map / routing tile etc.?

moving-bits:
## Description
Extract non-`MainActivity`-specific init/maintenance tasks to `AbstractBottomNavigationActivity` and call them in `onCreate` of related activities


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -212,6 +212,7 @@ public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         HideActionBarUtils.setContentView(this, R.layout.unifiedmap_activity, true);
+        runInitAndMaintenance(savedInstanceState);
```
fm-sys, line 215: Wouldn't it be possible to place this line in `AbstractBottomNavigationActivity#onCreate` so we don't need to call it everywhere separately?

moving-bits: Thinking behind that was that the calling activity should be initialized before calling init tasks, specifically setContentView.

fm-sys: Then `onStart` maybe?



https://developer.android.com/guide/components/activities/activity-lifecycle

moving-bits: Good idea. I needed to split up the routine, though, as creating backup tools requires a `savedInstanceState`, which is available in `onCreate`, but not in `onStart`. PR is updated.

PR #14342, opened 2023-06-16
add a relative date filter looking into the future to filter events by date

### LINKED ISSUE #14330: Improve hidden date filter (for event caches)
### Tell us your idea!

On support a user wants to use the filtered search to search for events taking place today or in near future.
For this purpose we uses the "filtered search"  , adds cache type = Event and a with "relative to today"-option filter for hidden date.

Unfortunately the hidden date filter does only allow "<3 days in the past" for the hidden date and no future hidden date.

For the use case of this users it would be necessary to search for hidden=Today and for hidden dates in the future using the "relative to today" option.

### Is there an alternative?

The alternative would be to use fixed dates, but as the user would like to store and reuse the filter it would only work (without the need of manual correction) if he uses the "relative to today" option.

### Additional Information

Support ticket 991014

ztNFny:
adds a new "Event Date" filter which is actually just a "Hidden Date" filter but with relative range reaching into the future, not the past

image

fixes #14330

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
@@ -313,7 +313,7 @@ public class GCConnector extends AbstractConnector implements ISearchByGeocode, 
                 GeocacheFilterType.NAME, GeocacheFilterType.OWNER,
                 GeocacheFilterType.TYPE, GeocacheFilterType.SIZE,
                 GeocacheFilterType.DIFFICULTY, GeocacheFilterType.TERRAIN, GeocacheFilterType.DIFFICULTY_TERRAIN,
-                GeocacheFilterType.FAVORITES, GeocacheFilterType.STATUS, GeocacheFilterType.HIDDEN, GeocacheFilterType.LOG_ENTRY);
+                GeocacheFilterType.FAVORITES, GeocacheFilterType.STATUS, GeocacheFilterType.HIDDEN, GeocacheFilterType.EVENT_DATE, GeocacheFilterType.LOG_ENTRY);
     }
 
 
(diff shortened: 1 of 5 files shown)


ztNFny:
Retest this please

bekuno:
This filter should also be possible for other platforms. Eg. OCDE: https://www.opencaching.de/viewcache.php?cacheid=198533

ztNFny:
> This filter should also be possible for other platforms. Eg. OCDE: https://www.opencaching.de/viewcache.php?cacheid=198533

It is working though? Didn't text with online search, but as I just duplicated the HIDDEN-date filter it might be that either OC doesn't support that or we simply never implemented it.

bekuno:
Maybe the second.
I did not follow the implementation deatails for the hidden date feature. I only saw that you only modified the GCConnector class.

ztNFny:
> Maybe the second. I did not follow the implementation deatails for the hidden date feature. I only saw that you only modified the GCConnector class.

Search by hidden date is not supported by OKAPI: https://www.opencaching.de/okapi/services/caches/search/all.html

eddiemuc:
> > Maybe the second. I did not follow the implementation deatails for the hidden date feature. I only saw that you only modified the GCConnector class.
> 
> Search by hidden date is not supported by OKAPI: https://www.opencaching.de/okapi/services/caches/search/all.html

True, but when filtering offline caches it should work.

See https://github.com/cgeo/cgeo/wiki/%22Search-by-Filter%22-support-by-c%3Ageo-Connectors for current state of supported online filters per connector

ztNFny:
> True, but when filtering offline caches it should work.

it does?! Offline filtering doesn't require any connector specific implementation.

eddiemuc:
> > True, but when filtering offline caches it should work.
> 
> it does?! Offline filtering doesn't require any connector specific implementation.

Correct.

But filtering for hidden date worked already for gc before. I assume that event date = hidden date field-wise, correct? So I would have expected that no gc-specific connector work is necessary either to enable event date filtering.

ztNFny:
> But filtering for hidden date worked already for gc before. I assume that event date = hidden date field-wise, correct? So I would have expected that no gc-specific connector work is necessary either to enable event date filtering.

As it's a new filter (basically an alias of HIDDEN) that new filter had to be added.
image

moving-bits:
@eddiemuc 
Ok to merge then?

eddiemuc:
> @eddiemuc 
> Ok to merge then?

Ok from my side

PR #14320, opened 2023-06-10
Support target selector for more navi apps (fix #13540)

moving-bits:
## Description
Google Maps (Driving) navigation app supports selection of target waypoint for a cache, if multiple waypoints exist.

This PR extends this feature to all navi apps derived from `AbstractPointNavigationApp`:
- AbstractRadarApp (inherited by RadarApp, PebbleApp)
- CompassApp
- CruiserNavigationApp
- GoogleNavigationApp (all variants)
- InternalMap
- OruxMapsApp
- OsmAndApp
- OtherMapsApp (all variants)

PR #14345, opened 2023-06-18
target selection only for driving navigation

### LINKED ISSUE #14344: "map" navigation shows popup with parking coordinates
In todays nightly pressing the compass icon which in my config is mapped to "map" (default navigation) opens a popup showing the cache and its parking coordinates instead of just drawing the line to the cache.

Showing where a cache is is THE main function of cgeo and this popup, requiring an additional selection - each time the button is pressed, which can be "very often" if you switch between listing and map - adds useless overhead to this.
If i want to draw a line to the parking coordinates i can go to the waypoints and select parking there.

ztNFny:
fixes #14344 by offering target selection only for those navigation apps that are explictily targeted towards driving

Apps with explicit "Driving" navigation mode in cgeo:
- Google Navigation (Driving)
- Sygic (Driving)

Other navigation apps not implementing this right now:
- Maps.me
- OruxMaps
- OsmAnd
- Locus

If we want to implement it for those other apps I suggest adding a second profile for them to explicitly set driving mode, then implement it for that.


main/src/main/java/cgeo/geocaching/apps/navi/AbstractPointNavigationApp.java
```diff
@@ -36,24 +36,30 @@ protected AbstractPointNavigationApp(@NonNull final String name, @Nullable final
...
+        assert coords != null; // asserted by caller
+        navigate(context, coords);
+    }
+
+    @Override
+    public void navigateWithTargetSelector(@NonNull final Context context, @NonNull final Geocache cache) {
         final ArrayList<IWaypoint> targets = new ArrayList<>();
         targets.add(cache);
         for (final Waypoint waypoint : cache.getWaypoints()) {
-            final Geopoint coords = waypoint.getCoords();
```
moving-bits, line 50: By changing the next two lines you also revert #14339 (which is not related to the reverted #14320)

main/src/main/java/cgeo/geocaching/apps/navi/AbstractPointNavigationApp.java
```diff
@@ -36,24 +36,30 @@ protected AbstractPointNavigationApp(@NonNull final String name, @Nullable final
     }
 
     @Override
-    public void navigateWithoutSelector(@NonNull final Context context, @NonNull final Geocache cache) {
```
moving-bits, line 39: By reversing the logic from `navigateWithoutSelector` back to `navigate` navigation you'll end up in a loop if you select the geocache (not one of its waypoints) as navigation target.

moving-bits (changes requested):

main/src/main/java/cgeo/geocaching/apps/navi/AbstractPointNavigationApp.java
ztNFny: re-implemented

ztNFny:
retest this please

moving-bits (approved):
LGTM, thanks for fixing the issues

PR #14346, opened 2023-06-18
DT marker: differentiate between missing and not supported

### LINKED ISSUE #14343: DT marker: differentiate between missing info and not available info
For geocaches DT info is available but might not be loaded = missing
For labs DT simply does not exist = not available

This should be differtiated in Ui

ztNFny:
fixes #14343

for D/T cache marker differentiate between "not supported by connector" -> "-" and "information missing" -> "?"

Afaik AL is the only connector not supporting DT, so this change effects only those as of right now

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/AbstractConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/AbstractConnector.java
@@ -292,6 +292,11 @@ public abstract class AbstractConnector implements IConnector {
     }
 
     @Override
+    public boolean supportsDifficultyTerrain() {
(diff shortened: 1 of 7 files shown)


fm-sys:
> Afaik AL is the only connector not supporting DT, so this change effects only those as of right now

Same applies for UDC's AFAIK

ztNFny:
> > Afaik AL is the only connector not supporting DT, so this change effects only those as of right now
> 
> Same applies for UDC's AFAIK

good catch, added

fm-sys:
retest this please

PR #14348, opened 2023-06-18
fix map marker caching

### LINKED ISSUE #14347: wrong icon from cache for 1/1 caches
caused by https://github.com/cgeo/cgeo/issues/14346

Caches with D/T 1/1 have the same icon as caches not support D/T.

This is caused by a quirk in icon cache:

float negative = -1.0f;
float positive = 1.0f;
final int hashNegative = new HashCodeBuilder().append(negative).append(negative).toHashCode();
final int hashPositive = new HashCodeBuilder().append(positive).append(positive).toHashCode();

Results in both having the same value?!?!
hashNegative = 1828739817;
hashPositive = 1828739817;

Any ideas what's causing this and how to best work around it?

Offending line: https://github.com/cgeo/cgeo/blob/e64063285923f47e505b87ad8d226fe82ba8ce56/main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java#L688

ztNFny:
fixes #14347

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
@@ -685,7 +685,7 @@ public final class MapMarkerUtils {
     }
 
     public static Drawable getDTRatingMarker(final Resources res, final float difficulty, final float terrain) {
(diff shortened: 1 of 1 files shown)


fm-sys:
Retest this please

PR #14354, opened 2023-06-18
Unified map: fix tiles not placed correctly in VTM (fix #13593)

### LINKED ISSUE #13593: Unified map: tiles not placed correctly in VTM
tiles not placed correctly in VTM

Cgeo version: 2022.11.01

Screenshot_20221101_222635_cgeo geocaching

fm-sys:
Unified map: fix tiles not placed correctly in VTM (fix #13593)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
@@ -192,6 +192,10 @@ public class MapsforgeThemeHelper implements XmlRenderThemeMenuCallback {
         if (mTheme != null) {
             mTheme.dispose();
         }
+        // reset scale to avoid tile scaling issues (see #13593)
+        CanvasAdapter.userScale = 1;
+        CanvasAdapter.textScale = 1;
+        CanvasAdapter.symbolScale = 1;
     }
 
     private ThemeFile createThemeFor(@NonNull final ThemeData theme) throws IOException {


main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
```diff
@@ -192,6 +192,10 @@ protected void disposeTheme() {
         if (mTheme != null) {
             mTheme.dispose();
         }
+        // reset scale to avoid tile scaling issues (see #13593)
```
moving-bits, line 195: This will disable all scaling effects for VTM-based maps effectively. Shouldn't we hide the related settings in theme options then as well (for the time being), until this is fixed in VTM?

fm-sys: > This will disable all scaling effects for VTM-based maps effectively.  

This is just called while disposing the theme. As soon as the theme is loaded, `applyScales(...)` will take care of setting the scales again.

Have you tested if scaling really doesn't work? If yes, there would be a bug inside the implementation as it's at least supposed to work correctly...

moving-bits: This is how it looks currently:

|Scaling options|NewMap|VTM|
|---|---|---|
|image|image|image|

moving-bits: This is how it looks on VTM using current master, without this PR:

Not exactly useful, but it DOES do some scaling :-)

image

(Differences to NewMap screenshot may be due to different map position and/or different theme settings besides scaling options, have not cross-checked.)

fm-sys: Ok, indeed this doesn't look right. If we doesn't find a different solution I would just vote for dropping the scaling options... (still better than wrongly placed tiles)

PR #14359, opened 2023-06-21
Replace progress dialogs with inline views in cache listing

### LINKED ISSUE #14355: Upload personal note in background
### Tell us your idea!

Personal Note upload currently is a foreground activity with a blocking dialog. This blocking seems unnecessary here

Turn it into a background activity and show a notification on error.

### Is there an alternative?

_No response_

### Additional Information

_No response_

### LINKED ISSUE #14358: Replace progress dialogs with inline views in cache listing
### Tell us your idea!

Several actions in cache listing show a progress dialog, e.g. Loading a cache, refresh, delete, store, (un)watch.
This dialog is blocking, preventing the user from further interactions with the app. In many cases there's no need for it to be blocking - e.g. while adding a cache to the watchlist I don't see why I shouldn't be able to read the description.

The important info is the completion message (shown as a toast anyway), not the progress message shown in the dialog, so that dialog could be removed without losing value.

Replace the progress dialog with inline progress indicators.
E.g. refresh, store, watch could use LinearProgressIndicator at the top of the cache listing.

Cache loading is the only one a bit more complex. We could either do a LinearProgressIndicator and add info as it's loaded (would be great for slow connections as data becomes accessible to the user as it becomes available) or replace it with an inline view (or leave that one as is)

### Is there an alternative?

_No response_

### Additional Information
extension of #14355

ztNFny:
Replace most dialogs when performing network actions from CacheDetails view with progress bar. Only one not replaced is the initial cache loading.
fixes #14358
fixes #14355
see these issues for more reasoning

PR also includes #14356 due to some rewrites, will rebase once that one is merged

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -80,7 +80,6 @@ import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.TrackableListAdapter;
 import cgeo.geocaching.ui.UserClickListener;
 import cgeo.geocaching.ui.ViewUtils;
-import cgeo.geocaching.ui.WeakReferenceHandler;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.EditNoteDialog;
 import cgeo.geocaching.ui.dialog.EditNoteDialog.EditNoteDialogListener;
@@ -113,7 +112,6 @@ import static cgeo.geocaching.apps.cache.WhereYouGoApp.isWhereYouGoInstalled;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
-import android.app.ProgressDialog;
 import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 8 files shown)


ztNFny:
retest this please

moving-bits:
Nice refactoring, makes users' life easier, thanks for diving into it!

FYI: PR #14356 has been merged

ztNFny:
Further improved this to show
LinearProgressIndicator on refreshing the cache (visible in all tabs as the action can be triggered from all tabs)
image

Circle progress indicator inside the button for store/watch/favorite
image

ztNFny:
retest this please

ztNFny:
I think I've covered every action in CacheDetailsActivity now. The only one not modified is the actual cache load, but that one is special as showing a cache with only half the info available would pose some challenges and require several UI adjustments.

LogCacheActivity should probably use background progress shown in CacheDetailsActivity as well, but let's leave that out here to avoid this PR getting huge.

moving-bits (dismissed):
I'm not that deep into that handler stuff, but from a quick smoke test most of it seems to work as intended. Only thing I've stumbled upon is a missing "favorites" icon when a cache is not one of your favorite-marked caches:

image

(If a cache is already one of your favorite ones, then the icon is being displayed.)

ztNFny:
fixed the missing icon - an old bug actually that only became visibile now. Both android:visibility and tool:visibility were set.

Also fixed an accidental revert along the way.

ztNFny:
"reset coordinates" is still a dialog, putting it back to WIP

ztNFny:
I think that's all the progressdialogs - please let me know if I've missed one

moving-bits (changes requested):
> fixed the missing icon - an old bug actually that only became visibile now. Both android:visibility and tool:visibility were set.

Does not seem to be fixed, at least not in the emulator. If you go to the list of stored caches and open a cache that is not on your watchlist, the "heart" icon to spend a favorite is missing. Still true after reloading the cache or when using "search by gc code".

If you open a cache that is part of your favorite list, then the icon is displayed.

ztNFny:
> > fixed the missing icon - an old bug actually that only became visibile now. Both android:visibility and tool:visibility were set.
> 
> Does not seem to be fixed, at least not in the emulator. If you go to the list of stored caches and open a cache that is not on your watchlist, the "heart" icon to spend a favorite is missing. Still true after reloading the cache or when using "search by gc code".
> 
> If you open a cache that is part of your favorite list, then the icon is displayed.

Works for me. Are you sure you tested with a found cache? You can't award FP to unfound caches. Maybe we shouldn't show the whole favorites-section in that case - but that'd be a seperate issue/PR.

moving-bits (approved):
> Works for me. Are you sure you tested with a found cache? You can't award FP to unfound caches. Maybe we shouldn't show the whole favorites-section in that case - but that'd be a seperate issue/PR.

You are right, works for a found cache. Thanks for diving into the blocking request issue.

PR #14367, opened 2023-06-25
UnifiedMap: Fix initial load of individual route (fix #14363)

### LINKED ISSUE #14363: UnifiedMap: Route not calculated on initial start of map
If you open unified map and already have a defined individual route, the waypoints of the route will be connected with straight lines only, not routed lines.

This changes when your action triggers a recalculation of the route, eg by entering "route edit mode" and confirming the "changes".

Adding caches/waypoints to the route makes their connection a routed line.
Also if you remove one route point "in the middle", their two connection points will then be connected by a routed line as well.

Tested with VTM only so far, with today's state of `master` branch.

moving-bits:
## Description
On activity init BRouter is not yet ready and returns empty routes, which get converted to straight connections. 
The callback for Routing.connect() therefore needs to reload the route, not only refresh its display.

Whether we want to leave in the `viewModel.reloadIndividualRoute();` line after the `Routing.connect()` line is a matter of taste: Shall the uncalculated route be displayed already, or do we wait for a calculated route?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -282,7 +282,7 @@ public class UnifiedMapActivity extends AbstractBottomNavigationActivity {
 
         // FilterUtils.initializeFilterBar(this, this);
 
-        Routing.connect(ROUTING_SERVICE_KEY, () -> viewModel.individualRoute.notifyDataChanged(), this);
+        Routing.connect(ROUTING_SERVICE_KEY, () -> viewModel.reloadIndividualRoute(), this);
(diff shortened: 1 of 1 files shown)


Lineflyer:
> Whether we want to leave in the viewModel.reloadIndividualRoute(); line after the Routing.connect() line is a matter of taste: Shall the uncalculated route be displayed already, or do we wait for a calculated route?

As long as "wait for" does not mean, that there is a delay for the user in showing the map it is indeed only a matter of taste. 
As calculating longer routes on slow devices might take some time, I personally think its better to draw the straight line first as it is some kind of "work in progress" indication. If no line shows up for several seconds I would otherwise think, that I did something wrong.

moving-bits:
Ok, as it may take a couple of seconds for longer routes, I'll adapt the PR to draw the straight line version on map startup.

PR #14378, opened 2023-06-27
update button icons in image gallery

ztNFny:
update buttons in image gallery to a) use current material design look, b) use filled (instead of outline) for "add file"

Old:
image

New:
image

### DIFF
--- a/main/src/main/res/drawable/ic_menu_camera.xml
+++ b/main/src/main/res/drawable/ic_menu_camera.xml
@@ -6,6 +6,6 @@
     android:viewportHeight="24"
     android:tint="?android:textColorPrimary">
   <path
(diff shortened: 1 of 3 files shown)


moving-bits:
Thanks for updating / modernizing the icons.

Is there some common understanding in material design where to put the "+" sign to?
(IMHO it would look more consistent if they were all in the same corner.)

ztNFny:
Apparently not. Those are standard icons.

moving-bits:
> Apparently not. Those are standard icons.

ok, I see. Although it's a bit weird, IMHO: I just scanned material icons font a bit, and they have those "add" / "plus" insets in at least five (!) different positions... 🤔

ztNFny:
They even have different + symbols...

PR #14383, opened 2023-07-02
Increase max value in FavoritesFilter (fix #11355)

### LINKED ISSUE #11355: Increase max val for Logs count filter
While I would increase the max val for Logs count, too - if possible in one go. Cause 1000 is a little bit low ...

_Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/11345#issuecomment-888982801_

moving-bits:
## Description
Increase max. value in FavoritesFilter to 10000 (if not in percentage mode)


fm-sys:
Considered as not necessary by the original author (see #11355)

moving-bits:
then let's close here

PR #14384, opened 2023-07-02
Replace ContextMenu by TextSpinner for TB log type (fix #11271)

### LINKED ISSUE #11271: Make type selectors consistent
Currently the trackable log activity displays a button labelled with current log type. If you press this button, a popup appears, in which you can change the log type.

image

This is different from what we use in waypoint edit or in cache log edit for type selection.

## Suggested solution
Make it consistent across type selections, at least between the two log type selections.

Either use a type spinner instead of the button. E. g. like we have in waypoint edit.

Or at least make it consistent with cache log type selector, which uses radio buttons.

|waypoint type|cache log type|
|---|---|
|image|image|

moving-bits:
## Description
For log type selection when logging caches we switched to `TextSpinner` component a while ago (instead of using a `ContextMenu`). 
This PR does the switch for trackable log types.

@eddiemuc 
Did I miss anything while applying `TextSpinner` here?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogTrackableActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogTrackableActivity.java
@@ -23,6 +23,7 @@ import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.settings.SettingsActivity;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.ui.DateTimeEditor;
+import cgeo.geocaching.ui.TextSpinner;
 import cgeo.geocaching.ui.dialog.CoordinatesInputDialog;
 import cgeo.geocaching.ui.dialog.CoordinatesInputDialog.CoordinateUpdate;
 import cgeo.geocaching.ui.dialog.Dialogs;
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/log/LogTrackableActivity.java
```diff
@@ -258,35 +259,10 @@ public void onConfigurationChanged(@NonNull final Configuration newConfig) {
...
-    }
-
     private void init() {
-        registerForContextMenu(binding.type);
-        binding.type.setOnClickListener(this::openContextMenu);
+        logType.setTextView(binding.type).setDisplayMapper(LogTypeTrackable::getLabel);
+        logType.setValues(possibleLogTypesTrackable);
+        logType.setChangeListener(this::setType);
 
         setType(typeSelected);
```
eddiemuc, line 267: Method "settype" should now be deletable. Instead the Textspinner's "set" value should be used. 

A good check in code is that the TextView "binding.type" should not be used anywhere in code except when assigned to the TextSpinner using it's "setTextView" method.

eddiemuc (commented):
See comment

main/src/main/java/cgeo/geocaching/log/LogTrackableActivity.java
moving-bits: Thanks for cross-checking. Method `setType` does a couple of things related to the selected value, so I guess it needs to stay. But I replaced a call to `binding.type.setText()` (now `logType.setTextView(binding.type)` is the only place left where `binding.type` is being used) and fixed a missing update to the spinner's content.

PR #14388, opened 2023-07-04
Extend scaling options

moving-bits:
## Description
Extend map scaling options (experimental)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -2253,7 +2253,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             // title
             holder.binding.name.setText(StringUtils.isNotBlank(wpt.getName()) ? StringEscapeUtils.unescapeHtml4(wpt.getName()) : coordinates != null ? coordinates.toString() : getString(R.string.waypoint));
-            holder.binding.textIcon.setImageDrawable(MapMarkerUtils.getWaypointMarker(activity.res, wpt, false).getDrawable());
+            holder.binding.textIcon.setImageDrawable(MapMarkerUtils.getWaypointMarker(activity.res, wpt, false, Settings.getIconScaleEverywhere()).getDrawable());
 
             // visited
             /* @todo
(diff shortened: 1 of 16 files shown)


moving-bits:
retest this please

ztNFny:
Interesting. D/T markers are broken though 
image

ztNFny:
> Interesting. D/T markers are broken though image

missing ", scalingFactorCacheIcons" in line 736:
insetsBuilder.withInset(new InsetBuilder(R.drawable.marker_rating_fg, scalingFactorCacheIcons));

What do you have in mind for the settings UI? A slider 50-200?

ztNFny:
Oops, those icons are not just used on map but everywhere:
image
image

moving-bits:
> Interesting. D/T markers are broken though
> missing ", scalingFactorCacheIcons" in line 736: insetsBuilder.withInset(new InsetBuilder(R.drawable.marker_rating_fg, scalingFactorCacheIcons));

thanks, I'll look into this

> What do you have in mind for the settings UI? A slider 50-200?

For now I started with a "hidden feature" to do some testing, but when we are going public with this, it'll be a slider probably, and 50-200 would be a valid range.

> Oops, those icons are not just used on map but everywhere:

Yeah. Initially the use-case I was thinking of was to support visually impaired people - font sizes are scaling, cache/waypoint icons are not. But your post made me rethink this. I see at least two different use-cases, and they need to be handled differently:
- enlarge icons to help visually impaired people => scaling of icons should happen everywhere in the app
- shrink icons to "see more of the map", but still have the details of the regular icons => scaling should happen on the map only

This could be solved by a switch whether to apply the scaling "everywhere" or "on map only.

I'll try to dive into this over the next days.

ztNFny:
Visually impaired will need more than larger icons. I think that's already solved by the global Android density settings. And our map is pretty unusable for visually impaired anyway, nothing that larger icons could fix. 
So I think it's down to the second usecase.

ztNFny (commented):
may I suggest to call "forMap" parameter "scaled" or "withScaling" or something along the lines?

You're not scaling dot icons, is that by intention?

moving-bits:
> may I suggest to call "forMap" parameter "scaled" or "withScaling" or something along the lines?

changed it to `applyScaling`

> You're not scaling dot icons, is that by intention?

yes (so far) - the dot icons are already pretty small, scaling them further down could make them unreadable, IMHO. But let's see how it works out in daily use

PR #14401, opened 2023-07-07
Fix #14365 - Route distance view not updated

### LINKED ISSUE #14365: UnifiedMap: Route distance view not updated on route recalculation
### Description:
Looking at the screenshots in #14364 (see below), (and seemingly confirmed in a quick smoke test) distance view for individual route does not get updated on recalculation of a route when using `UnifiedMap`.

It does get updated, though, if you append/prepend or remove a route item to/from the individual route.
(But only for the affected segment, not for the whole route, so adding and removing the same point reverts to 550m, taken the example described below, instead of returning to 759m after removing the additional point - which is even stranger...)

### Screenshots copied from #14364:
The two screenshots in the first column (taken from `NewMap` and `CGeoMap` respectively) show a distance of 759m, which is the calculated distance for the three route items, whereas the three screenshots from `UnifiedMap` still show 550m, which is the straight line distance. (Remember, that `UnifiedMap` currently does not calculate route on startup, see #14363, and only calculates and updates the route on going into route edit and saving the (un-)changed route. This updates the route display, but not the route distance view for the individual route.)

|NewMap|VTM before|VTM after|
|---|---|---|
|image|image|image|

|CGeoMap|Google Map Unified|
|---|---|
|image|image|

(taken on a different device, though, compared to Mapsforge/VTM screenshots)

fm-sys:


fm-sys:
@moving-bits this PR is a nice example to see how `livedata` can be used and what's the benefit of it. As all data is available via the view model, you can just observe whatever you need without complicated refactoring.

PR #14405, opened 2023-07-09
Hide elements of bottom navigation

moving-bits:
## Description
Small preview to demonstrate how hiding certain elements of bottom navigation could be implemented (as has been requested in the past for at least "home screen" and "nearby search").

The actual bottom navigation items are still available and reachable, just no longer being displayed in bottom navigation.
If you have a c:geo shortcut to eg "nearby", you can still call this from your mobile's home screen. But you won't see it in bottom navigation (and cannot accidentally hit the "nearby" search entry).

Currently, it's implemented for "nearby" only:

image

Before we also implement this for the other elements as well, we should make sure, that all elements are available as c:geo shortcuts, otherwise hiding the home screen would effectively lock you out from changing settings.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -196,6 +196,7 @@ public abstract class AbstractBottomNavigationActivity extends AbstractActionBar
     @Override
     protected void onResume() {
         super.onResume();
+        showHideBottomNavItems();
         updateSelectedBottomNavItemId();
         startLoginIssueHandler();
         registerReceiver(connectivityChangeReceiver, new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION));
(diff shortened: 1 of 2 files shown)


ztNFny:
> Before we also implement this for the other elements as well, we should make sure, that all elements are available as c:geo shortcuts

imho Shortcuts (I'm assuming you mean launcher shortcuts) are not something we can rely on here. This is a quite unknown "power-user" feature, not even supported in all launchers. If a user manages to hide "Home" chances are quite high he has no idea about launcher shortcuts to re-access the setting.
As Home is the only place to access cgeos settings I would vote for not being able to hide "Home" - which would probably mean this feature becomes a "hide nearby-tab" as that's the only tab that's not required for normal use.

moving-bits:
> imho Shortcuts (I'm assuming you mean launcher shortcuts) are not something we can rely on here. This is a quite unknown "power-user" feature

agreed. I consider hiding a bottom navigation entry as a "power-user feature" as well, so why not make it config-item only (no UI, just document it in our manual as well as how to create launcher shortcuts)?

> As Home is the only place to access cgeos settings I would vote for not being able to hide "Home" - which would probably mean this feature becomes a "hide nearby-tab" as that's the only tab that's not required for normal use.

There had been explicit requests to hide (or better: get rid of) home screen as well back then when we implemented bottom navigation, which such a setting could help with. Creating a launcher shortcut (if available) could give those users access to items like settings etc. in case of need.

But I could live with not being able to hide "home screen" as well.

ztNFny:
Can we add settings to bottom-nav? I use that way more often than nearby. Optional, not by default.

Then we could enforce home or settings.

moving-bits:
Do you mean "settings" instead of "home"? (optional) - This would still leave out other functionality of home screen, if we do not support a "home screen" launcher shortcut. And if there's no UI for the settings, you would not be able to revert this setting.

Or do you mean "settings" instead of "nearby". This could work, indeed. (I'm using a quick launch button on my home screen for this.)

One could even make the last position of the bottom navigation configurable with every quick launch item we have available for home screen. (Have not checked icon compatibility yet.) Additional entry "empty" would be required, though, to not use a fifth item in bottom navigation.

fm-sys:
> One could even make the last position of the bottom navigation configurable with every quick launch item we have available for home screen.

I would be in favor of this!

Hide elements of bottom navigation would IMHO just be another useless feature (sorry to be so hard), but being able to replace an item with something you actually need brings great additional benefit. Like it!

moving-bits:
Superseded by #14419

PR #14406, opened 2023-07-09
Toggle proximity notifications from quick settings (fix #13745)

### LINKED ISSUE #13745: Enable Proximity Sounds via Map Quick Settings
### Tell us your idea!

Right now it is a long journey to enable the Proximity Sound feature via the settings menu and find the correct spot inside the map settings section.

It might be easier, when this feature could be enabled via the map quick settings menu:

IMG_20221203_155541

Add an additional button inside the map quick settings menu, to enable or disable the proximity sound feature.

In general the option to define your own sound might be cool as well.

### Is there an alternative?

See above, do it again and again via the long settings menu. 

### Additional Information

_No response_

moving-bits:
## Description
Adds a toggle to map quick settings, which enables/disables proximity notifications.
Toggle is shown only, when at least one of the proximity notifications (general/specific) has been activated.
This toggle does not change the proximity notification settings but rather acts as a "master toggle".

image

Edit: Changed UX meanwhile, see below


ztNFny:
Look should follow the other icons (grey/orange)

I'm not sure about position and non-text, makes it hard to understand what the button does.

moving-bits:
> Look should follow the other icons (grey/orange)

disabled state is grey, tint needs yet to be set for active state

> I'm not sure about position and non-text, makes it hard to understand what the button does.

long tap on icon will provide a detailed info for this

moving-bits:
> I'm not sure about position

A different approach could be to add a checkbox with "enable proximity notifications" at the end of the popup, after a new sub heading "proximity notifications". 

Or we merge the existing checkbox "Set route to start" and the new one under a renamed sub heading of "Other" (or something similar).

moving-bits:
Updated PR to use a checkbox instead of toggle buttons:

image

Last section will only be visible if an individual route exists and/or at least one of the proximity notifications is enabled in settings. Only relevant options are shown.

PR #14414, opened 2023-07-14
Update Mapsforge & VTM to v0.19.0

moving-bits:
## Description
Updates both Mapsforge and VTM to v0.19
(Removes usage of specific commit id of Mapsforge to support hillshading on SAF, as this is part of v0.19 meanwhile)


fm-sys:
Let's test it in the nightlys...

PR #14417, opened 2023-07-15
Add message center quicklaunch item

moving-bits:
## Description
As discussed in our last team meeting: Make message center available on home screen even when unread messages notification is disabled.

This PR adds a quicklaunch item which the user can activate. Tapping on it opens message center in the browser.

image


ztNFny:
Does this get a (red) notification bubble when there are unread messages?

moving-bits:
> Does this get a (red) notification bubble when there are unread messages?

No, this this just a startup link. Blue dot goes to "home" icon in bottom navigation, same as for other update notifications. Plus message box on home screen.

PR #14419, opened 2023-07-15
Make last bottom navigation item customizable

moving-bits:
## Description
Makes the last bottom navigation item customizable:
- default (= nearby search)
- empty
- any of the quick launch items (same as on home screen)

Configuration is in settings => appearance => Custom Menu Entry

This is an alternative to #14405.

Example: Last item replaced by "pocket queries":
image

Configuration dialog: (updated)
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/MainActivity.java
+++ b/main/src/main/java/cgeo/geocaching/MainActivity.java
@@ -30,7 +30,6 @@ import cgeo.geocaching.sensors.GnssStatusProvider.Status;
 import cgeo.geocaching.sensors.LocationDataProvider;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.settings.SettingsActivity;
-import cgeo.geocaching.settings.ViewSettingsActivity;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.extension.FoundNumCounter;
 import cgeo.geocaching.storage.extension.PendingDownload;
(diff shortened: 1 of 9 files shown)


bekuno:
I'm with fm-sys in https://github.com/cgeo/cgeo/pull/14405#issuecomment-1636705022.
Don't implement"empty", maybe only show this menu with long click of the bottom menu entry.

moving-bits:
> I'm with fm-sys in [#14405 (comment)](https://github.com/cgeo/cgeo/pull/14405#issuecomment-1636705022). Don't implement"empty", maybe only show this menu with long click of the bottom menu entry.

Reasoning behind the "empty" feature is a very bad user experience I had during my last cache trip: I was going by bike, but had forgotten the adapter to fix my smartphone on the bike, so I kept holding it in my hands. During the trip I frequently hit the "nearby" button accidentally, by this leaving the map I was looking at. This happened at least 20 time during the trip, which was extremely annoying. My guess is, that this was due to the buttons taking up the full width of the screen when five buttons are being displayed. Being able to hide one of those buttons temporarily would have solved the issue easily.

Hiding the fifth button will look like this:
image

So it's not a blank space at the fifth position, but the other four buttons get evenly distributed, which looks totally fine, IMHO.

ztNFny:
I'd like to have an actual "empty" item. While biking I very frequently manage to hit the bottom right button (but somehow never any of the other), for that it'd be nice to have a "do nothing" section in there.

moving-bits:
> I'd like to have an actual "empty" item. While biking I very frequently manage to hit the bottom right button (but somehow never any of the other), for that it'd be nice to have a "do nothing" section in there.

You mean like a "placeholder"? So the first four buttons would stay where they are now, and a fifth place would stay empty?

ztNFny:
> > I'd like to have an actual "empty" item. While biking I very frequently manage to hit the bottom right button (but somehow never any of the other), for that it'd be nice to have a "do nothing" section in there.
> 
> You mean like a "placeholder"? So the first four buttons would stay where they are now, and a fifth place would stay empty?

Yes. As an additional option to "hide" of course.

PR #14432, opened 2023-07-20
Remove the temporary string in strings.xml

bekuno:
Remove the temporary string in strings.xml, as the error is solved in the meantime, the issue https://github.com/material-components/material-components-android/issues/1346 is fixed.


bekuno:
retest this please

Issue #14371, opened 2023-06-25
"Insert last log" discarded when adding image first


Lineflyer:
### Describe your problem!

Described by a user on support mail, not verified

### How to reproduce?

- Compose and send a log for cache 1
- Start logging cache 2
- Insert a picture to be attached to the log
- Use "Insert last log text" from the menu to get the log text of cache 1

### Actual result after these steps?

- Nothing pasted (empty)

### Expected result after these steps?

- Log text from cache 1 is pasted

This does only work if the user does add the image only **after** using "insert last log text"

### Reproducible

Unclear

### c:geo Version

2023.06.18

### Additional Information

Support ticket 235587

murggel:
Can't reproduce, but we could save the last log only if it is not empty...

murggel:
@Lineflyer shouldn't happen anymore with the fix in PR #14434

PR #14434, opened 2023-07-20
save only non-empty-log-text to be used as "repeat last log" (rel #14371)

murggel:
## Description

Currently even an empty log text is saved for future use and therefor overrides the last "useful" log-text

This PR saves only non-empty-log-text for using it in "repeat last log".
Saving an empty log is still possible.

## Related issues

rel #14371 ("Insert last log" discarded when adding image first)


fm-sys:
How will one still be able to delete an unsent log?

murggel:
> How will one still be able to delete an unsent log? 

This is not for (not) saving a log, but  saving the last log entry-text for pasting it in as "repeat last log".

Will enhance the description

fm-sys:
Ah, OK. This sounds like a nice enhancement 👍

PR #14436, opened 2023-07-21
First stab at OrganicMaps integration

moving-bits:
## Description
This is a first stab at OrganicMaps integration. Their API is somewhat incomplete as of now, compared to MWM, so I filled some of the voids. For now, this is a navigation integration only.

### Some technical info
Integration into c:geo is done similar to MWM integration, by copying the "lib" section of OrganicMap's `android-api` project, and then adapting a few files to avoid name clashes with MWM API (namely names of Download dialog layout resource and a string).

`OrganicMapsAPI.java` adds some methods yet missing in the OrganicMaps API.


moving-bits:
OrganicMaps uses a different code style than we have established for c:geo, therefore we receive quite a lot of Codacy warnings.

- Shall we ignore those warnings (and configure our CI to do so), as those files are 3rd party, and we may update them from time to time, 
- or shall I adapt the coding style, as they are copied files anyway (which might updating them in the future a bit harder)?

Any opinions?

bekuno:
The checkstyle integtration is defined in the build.gradle only for c:geo.
I asume that codacy will therefore take the own defaults.

I would as first ignore the warnings, maybe you are able to define a dummy checkstyle.xml to skip the warnings.
(I can't do it at least the next 2 weeks)

moving-bits:
I took a look at how we handle this for the old MWM API - here it's just a lint config. I'll see what happens if I copy this to OM API...

moving-bits:
> I took a look at how we handle this for the old MWM API - here it's just a lint config. I'll see what happens if I copy this to OM API...

Ok, Codacy doesn't care about lint config, so I guess it needs to be configured on the Codacy server-side (path exclusion). I don't have the necessary admin rights for this.

@Lineflyer 
Can you add the path exclusion to Codacy?

moving-bits:
retest this please

PR #14440, opened 2023-07-22
Fix test for temporarily disabled GC40

moving-bits:
## Description
Fix test for temporarily disabled GC40


moving-bits:
Codacy does not print any error or warning, and build tests are back to green, so let's merge.

PR #14462, opened 2023-07-30
Implement search center indicator for Unified Map

### LINKED ISSUE #7783: Have marker for search center
From support mail (translated):
It is a nice feature, that I can now directly open the map from the address search result, however sometimes it can be helpful to get a marker on the map, which shows the coordinates of the search result / search center to help with orientation on the map.

moving-bits:
## Description
Implements search center indicator for UnifiedMap.
In conjunction with #14449 this fixes #7783.

(BTW: First time for me working with the new UM layers and UM viewmodel - quite nice to work with!)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
@@ -38,6 +38,8 @@ public class UnifiedMapViewModel extends ViewModel implements IndividualRoute.Up
 
     public final HashMapLiveData<String, IWaypoint> geoItems = new HashMapLiveData<>();
     public final MutableLiveData<Geopoint> longTapCoords = new MutableLiveData<>();
+    public final MutableLiveData<Geopoint> coordsIndicator = new MutableLiveData<>(); // null if coords indicator should be hidden
+
     /**
      * LiveData wrapping the PositionHistory object or null if PositionHistory should be hidden
      */
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
@@ -5,7 +5,9 @@ public class LayerHelper {
     // constants for layer ordering (higher = more visible)
     // layer index numbers must be consecutive, starting with 1 for ZINDEX_BASEMAP
 
-    public static final int ZINDEX_POSITION = 13;
+    public static final int ZINDEX_POSITION = 14;
+
+    public static final int ZINDEX_SEARCHCENTER = 13;
 
     public static final int ZINDEX_TRACK_ROUTE = 12;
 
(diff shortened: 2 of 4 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
```diff
@@ -53,6 +56,11 @@ public void setCurrentPositionAndHeading(final Location location, final float he
         positionAndHeading.setValue(new Pair<>(location, heading));
     }
 
+    /** null if coords indicator should be hidden */
+    public void setCoordsIndicator(@Nullable final Geopoint geopoint) {
```
fm-sys, line 60: This setter method might not be necessary here. You could just call `viewmodel.coordsIndicator.setValue(geopoint);` from inside UnifiedMapActivity... At least that's the way how I did it at most other places. Explicit setter methods inside the viewmodel should IMHO just be used if there is more processing involved but not for simple pass through.

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
```diff
@@ -20,6 +20,8 @@
 import androidx.lifecycle.MutableLiveData;
 import androidx.lifecycle.ViewModel;
 
+import javax.annotation.Nullable;
```
fm-sys, line 23: IIRC we use `androidx.annotation.Nullable` normally. Not sure what's the difference thought... (but it'll get obsolete anyway if we remove the setter again)

fm-sys (commented):
Thank you very much for trying it out, LGTM!

I've added one suggestion though...

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
moving-bits: ah, that's even easier, thanks for pointing this out. I'm still in the process of getting used to the new ViewModel for UnifiedMap...

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
moving-bits: Yeah, don't know why I've select this one from the suggestions (or whether it got added automatically). Anyway, is removed meanwhile.

moving-bits:
> I've added one suggestion though...

Thanks for cross-checking. Updated my PR accordingly.

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapViewModel.java
fm-sys: > ah, that's even easier, thanks for pointing this out. I'm still in the process of getting used to the new ViewModel for UnifiedMap...

Really appreciated! Don't hesitate to ask if any questions should arise :)

PR #14467, opened 2023-07-31
Fix crash on screen rotation when using custom button (fix #14465)

### LINKED ISSUE #14465: Crash in landscape
With Saturdays master (https://github.com/cgeo/cgeo/commit/4e4a22db776afaf370d6cec6abf1c64641ddb4dd) i get crashes whenever cgeo is started in landscape or when rotating the device and doing any UI interaction.
Screen size in landscape is 1706x1080 - might be relevant as i can't reproduce on another device.
In portrait everything is fine.

FATAL EXCEPTION: main
Process: cgeo.geocaching, PID: 17939
java.lang.RuntimeException: Unable to resume activity {cgeo.geocaching/cgeo.geocaching.MainActivity}: java.lang.ClassCastException: com.google.android.material.navigationrail.NavigationRailView cannot be cast to com.google.android.material.bottomnavigation.BottomNavigationView
	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4935)
	at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:4974)
	at android.app.servertransaction.ResumeActivityItem.execute(ResumeActivityItem.java:60)
	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
	at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:176)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2345)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:233)
	at android.os.Looper.loop(Looper.java:344)
	at android.app.ActivityThread.main(ActivityThread.java:8212)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:584)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1034)
Caused by: java.lang.ClassCastException: com.google.android.material.navigationrail.NavigationRailView cannot be cast to com.google.android.material.bottomnavigation.BottomNavigationView
	at cgeo.geocaching.activity.AbstractBottomNavigationActivity.setCustomBNitem(AbstractBottomNavigationActivity.java:252)
	at cgeo.geocaching.activity.AbstractBottomNavigationActivity.onResume(AbstractBottomNavigationActivity.java:206)
	at cgeo.geocaching.MainActivity.onResume(MainActivity.java:443)
	at android.app.Instrumentation.callActivityOnResume(Instrumentation.java:1491)
	at android.app.Activity.performResume(Activity.java:8280)
	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:4925)
	... 13 more

moving-bits:
## Description
Adds missing check for menu component used for landscape mode

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
@@ -63,6 +63,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import com.google.android.material.badge.BadgeDrawable;
 import com.google.android.material.bottomnavigation.BottomNavigationView;
 import com.google.android.material.navigation.NavigationBarView;
+import com.google.android.material.navigationrail.NavigationRailView;
 
 
 public abstract class AbstractBottomNavigationActivity extends AbstractActionBarActivity implements NavigationBarView.OnItemSelectedListener {
@@ -249,7 +250,10 @@ public abstract class AbstractBottomNavigationActivity extends AbstractActionBar
     }
 
     private void setCustomBNitem() {
-        final MenuItem menu = ((BottomNavigationView) binding.activityBottomNavigation).getMenu().findItem(MENU_NEARBY);
+        final MenuItem menu = (binding.activityBottomNavigation instanceof  NavigationRailView
+                ? ((NavigationRailView) binding.activityBottomNavigation).getMenu().findItem(MENU_NEARBY) // landscpae mode
+                : ((BottomNavigationView) binding.activityBottomNavigation).getMenu().findItem(MENU_NEARBY)); // portrait mode
+
         menu.setVisible(true);
         menu.setEnabled(true);
         final int item = Settings.getCustomBNitem();


main/src/main/java/cgeo/geocaching/activity/AbstractBottomNavigationActivity.java
```diff
@@ -249,7 +250,10 @@ public void updateSelectedBottomNavItemId() {
     }
 
     private void setCustomBNitem() {
-        final MenuItem menu = ((BottomNavigationView) binding.activityBottomNavigation).getMenu().findItem(MENU_NEARBY);
+        final MenuItem menu = (binding.activityBottomNavigation instanceof  NavigationRailView
```
fm-sys, line 253: Probably you can just use [NavigationBarView](https://developer.android.com/reference/com/google/android/material/navigation/NavigationBarView) which is the superclass of both

moving-bits: A typical case of bad class naming. And of "I should have RTFM" ;-) Thanks for the pointer. Fixed with #14469.

moving-bits: > Maybe we should also rename our layout resource id from `activityBottomNavigation` to something which sounds a bit more like NavigationBarView?

Ok, but let's do this in a separate PR. 

Proposal would be:
- name the layout files `activity_navigation.xml` instead of `activity_bottom_navigation.xml`,
- and name the navigation component part within those files `navigationBarView` (or maybe with underlines) instead of `activity_bottom_navigation`

moving-bits: See #14484 for renaming of files / view id

PR #14469, opened 2023-08-01
Simplify component identification for screen rotation (rel. to #14467)

moving-bits:
## Description
User superclass `NavigationBarView` instead of the two involved sublasses.


moving-bits:
retest this please

PR #14483, opened 2023-08-04
Add OpenTopoMap as map provider for Mapsforge and VTM (fix #9470)

### LINKED ISSUE #9470: Add OpenTopoMap
It would be great when the app also supports OpenTopoMap
[https://wiki.openstreetmap.org/wiki/OpenTopoMap](https://wiki.openstreetmap.org/wiki/OpenTopoMap)

moving-bits:
## Description
Adds OpenTopoMap as map provider for both `NewMap` and `UnifiedMap`, including attribution etc.


moving-bits:
retest this please

PR #14485, opened 2023-08-04
WIP take map bearing into account

moving-bits:
## Description
Tries to take map bearing into account for heading indicator on UnifiedMap VTM

Can test against emulator only, currently, and here positionAndHeading observer never gets an update. Need to check against physical device, therefore setting WIP for now.


fm-sys:
From my experience, simulating a route playback instead of setting fixed coords does work well even in the emulator.

moving-bits:
> From my experience, simulating a route playback instead of setting fixed coords does work well even in the emulator.

I was using Lockito to do such a simulation, and map center was moving well, only the heading indicator did not move at all.

fm-sys:
Maybe Lockito doesn't simulate heading changes. At least I know that the indicator was rotating into traveling direction when using the android studio feature.

PR #14526, opened 2023-08-19
fix crash on adding a cache to favorites, then very quickly returning to map

ztNFny:
Crash condition introduced by https://github.com/cgeo/cgeo/pull/14359


moving-bits:
Can you target this to `release`, please? #14359 is part of recently released version 2023.08.16

ztNFny:
rebased onto release

Issue #14499, opened 2023-08-10
targetSdk 33


bekuno:
### Tell us your idea!

From Google: https://developer.android.com/google/play/requirements/target-sdk
~~~
Starting on August 31, 2023:

   - New apps must target Android 13 (API level 33) or higher; except for [Wear OS](https://developer.android.com/wear) apps, which must target a version between Android 11 (API level 30) and Android 13 (API level 33), inclusive.
   - App updates must target Android 13 or higher and adjust for behavioral changes in Android 13; except for Wear OS apps, which must target Android 11.
~~~

We are currently at targetSdk 31.

### What solution would you suggest?

Move to targetSdk 33 after the next release.

### Is there an alternative?

No, from August 31, 2023

### Additional Information

https://developer.android.com/google/play/requirements/target-sdk#pre13
https://developer.android.com/about/versions/13/behavior-changes-13

moving-bits:
From a quick scan at least this section seems to be relevant to us: https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions

moving-bits:
PR #14527 will set `targetSDK` to 33. Let's see in the nightlies whether additional adaptions are required.

Lineflyer:
I just visited Google Play Console first since longer time.
Seems they are enforcing this requirement, means we need next release either to be SDK33 targeted or need to release prior end of August.

I am posting this, as I remember to have read some threads on here with a (kind of major) regression regarding long tap on map and route planning. So we should thinkn about whether a short term bug fix before end of August makes sense to get rid of that.

Sorry I I missed some other thread dealing with this release scenarios...to less time to catch up at this moment.

moving-bits:
I propose to (and can prepare a) release before the weekend. We have reverted the Google renderer update which caused the long tap on map markers problem, and have some minor other fixed on `release` as well.

We have upgraded `master` to targetSDK 33 a couple of days ago and no problems reported for this so far, so we have a possible update in the queue, but it needs some more "time in the wild" before we should release that.

Lineflyer:
Sounds good as some tickets from users keep incoming which look related to this long tap problem.

moving-bits:
As discussed in our last dev meeting I did some tests regarding `targetSDK` 33:
- Create a fresh API 33-based emulator
- Install c:geo (no backup)
- Upload some files (images and other docs) using Android Studio's "device explorer"
[...]

PR #14527, opened 2023-08-19
Set targetSDK to 33

moving-bits:
## Description
As per Play Store requirements set `targetSDK` to 33. (Apps not supporting `targetSDK` 33 will be blocked from updating on Play Store starting end of August, see #14499.)

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -32,7 +32,7 @@ android {
     defaultConfig {
         minSdk 21
         //noinspection OldTargetApi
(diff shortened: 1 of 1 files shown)


bekuno:
Looks good in the build log.

moving-bits:
A quick test in the emulator was positive as well, I was able to grant rights on setting up c:geo, and to select arbitrary files of "image" gallery document selection. So let's give it a try in the nightlies.

PR #14532, opened 2023-08-20
Add c:geo version info to logfile (fix #14530)

### LINKED ISSUE #14530: add BuildConfig.VERSION_NAME in the initial application log
### Tell us your idea!

Currently the BuildConfig.VERSION_NAME is not contained in the c:geo logfile.
This could be very helpful when checking the logs.

I propose to add them within the init block (`CGeoApplication.onCreate`).

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Add c:geo version info to logfile (fix #14530)
(Using `Log.e()` to make sure it's included in a logfile and not filtered out by log level)


moving-bits:
retest this please

Issue #14375, opened 2023-06-26
database errors reported by Play Store


moving-bits:
This is a collection of different database errors Play Store reports for c:geo. All of them are reported for current 2023.06.18 release of c:geo, most (or probably all) of them also for earlier versions. 

I tried to dig into some of the messages, and several articles on StackOverflow suggest the cause to be parallel access to the database _initialization routines_ / _multiple DBHelper objects_ as well as multiple closing operations (whereby the latter may be a cause of the former as well).

### java.lang.IllegalStateException
in `cgeo.geocaching.storage.DataStore$DBExtension.load`

```
Exception java.lang.IllegalStateException: Cannot perform this operation because the connection pool has been closed.
  at android.database.sqlite.SQLiteConnectionPool.throwIfClosedLocked (SQLiteConnectionPool.java:1393)
  at android.database.sqlite.SQLiteConnectionPool.waitForConnection (SQLiteConnectionPool.java:967)
  at android.database.sqlite.SQLiteConnectionPool.acquireConnection (SQLiteConnectionPool.java:615)
  at android.database.sqlite.SQLiteSession.acquireConnection (SQLiteSession.java:916)
  at android.database.sqlite.SQLiteSession.executeForCursorWindow (SQLiteSession.java:836)
  at android.database.sqlite.SQLiteQuery.fillWindow (SQLiteQuery.java:62)
  at android.database.sqlite.SQLiteCursor.fillWindow (SQLiteCursor.java:157)
  at android.database.sqlite.SQLiteCursor.getCount (SQLiteCursor.java:144)
  at android.database.AbstractCursor.moveToPosition (AbstractCursor.java:238)
  at android.database.AbstractCursor.moveToNext (AbstractCursor.java:287)
  at cgeo.geocaching.storage.DataStore$DBExtension.load (DataStore.java:628)
  at cgeo.geocaching.storage.DataStore$DBExtension.load (DataStore.java:617)
  at cgeo.geocaching.storage.extension.FoundNumCounter.load (FoundNumCounter.java:30)
  at cgeo.geocaching.storage.extension.FoundNumCounter.updateFoundNumAndUserName (FoundNumCounter.java:43)
  at cgeo.geocaching.storage.extension.FoundNumCounter.getAndUpdateFoundNum (FoundNumCounter.java:67)
  at cgeo.geocaching.connector.oc.OCApiLiveConnector.login (OCApiLiveConnector.java:147)
[...]

PR #14537, opened 2023-08-20
WIP Move backup/restore to BackupActivity

moving-bits:
## Description
- Requesting a backup (either manually or automatically) or a restore just sets an internal flag in settings and informs the user about necessary restart.
- `SplashActivity` checks for this flag and calls `BackupActivity` if the flag is set, otherwise calls regular start activity

## Related issues
See discussion in #14375

## Additional context
This is a draft version still unfinished and having some UX issues, posting it here to get some early feedback.

## Notes to myself
- "Continue" button should only be accessible when cancelling current action, or when current action has finished. Requires an extension for restore actions, as it does not use a callback yet.
- ~On requesting a backup/restore, ask user before restarting c:geo (or at least give some user feedback instead of just a blank screen)~ _(implemented)_
- actions missing yet: changing number of backups; moving database

### DIFF
--- a/main/src/main/java/cgeo/geocaching/SplashActivity.java
+++ b/main/src/main/java/cgeo/geocaching/SplashActivity.java
@@ -1,5 +1,6 @@
 package cgeo.geocaching;
 
+import cgeo.geocaching.backup.BackupActivity;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.ContentStorageActivityHelper;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
@@ -8,6 +9,7 @@ import cgeo.geocaching.utils.FileUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.TextUtils;
+import static cgeo.geocaching.settings.Settings.EXCLUSIVEDBACTION.EDBA_NONE;
 
 import android.content.Intent;
 import android.os.Bundle;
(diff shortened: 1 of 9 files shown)


main/src/main/java/cgeo/geocaching/SplashActivity.java
```diff
@@ -29,6 +30,9 @@ public void onCreate(final Bundle savedInstanceState) {
                 // new install, base folder missing or folder migration needed => run installation wizard
                 intent = new Intent(this, InstallWizardActivity.class);
                 intent.putExtra(InstallWizardActivity.BUNDLE_MODE, firstInstall ? InstallWizardActivity.WizardMode.WIZARDMODE_DEFAULT.id : InstallWizardActivity.WizardMode.WIZARDMODE_MIGRATION.id);
+            } else if (Settings.getBackupRestoreRequested() > 0) {
+                // if any pending backup requests are found: start backup
+                intent = new Intent(this, BackupActivity.class);
```
eddiemuc, line 35: Wouldn't it be enpough ti synchronously execute backup/restore here? Is a new activity really needed?

main/src/main/java/cgeo/geocaching/backup/BackupActivity.java
```diff
@@ -0,0 +1,67 @@
...
+        final Intent intent = Settings.getStartscreenIntent(this);
+        intent.putExtras(getIntent());
+
+
+        try (ContextLogger cLog = new ContextLogger(Log.LogLevel.DEBUG, "BackupActivity.onCreate")) {
+            backupUtils = new BackupUtils(this, savedInstanceState == null ? null : savedInstanceState.getBundle(STATE_BACKUPUTILS));
+
+            if ((Settings.getBackupRestoreRequested() & Settings.BACKUP_AUTO) > 0) {
+                Settings.clearBackupRestoreRequested(Settings.BACKUP_AUTO);
+                backupUtils.backup(() -> Settings.setAutomaticBackupLastCheck(false), true);
```
eddiemuc, line 38: Careful. "Backup" also backs up settings. So if executed here and other flags are set (e.g. "RESTORE"), then those would also be backed up. And when this backup then is applied, this flag would also be restored.

Better first check all flags, then clear all flags, then execute depending on the flags.

main/src/main/java/cgeo/geocaching/SplashActivity.java
moving-bits: Possibly yes, but I wanted to keep backup details clear from `SplashActivity` (which is not much more than some switch on startup), and also `SplashActivity` uses a very reduced and special theme, whereas `BackupActivity` uses the regular c:geo theme.

main/src/main/java/cgeo/geocaching/backup/BackupActivity.java
moving-bits: Good catch, will adapt accordingly.

Thanks for taking the time for a review.

moving-bits:
Let's try the lock/unlock approach first, setting WIP label for now

moving-bits:
The solution implemented by @eddiemuc seems to work fine, so let's close this PR with the alternative approach

PR #14548, opened 2023-08-26
fix #14516: allow edit of log image caption default praefix

### LINKED ISSUE #14516: Default image name - as setting under logging
### Tell us your idea!

Currently each image added to a log entry is named (captioned) "Image", with multiple images numbered sequentially from 2 upwards. 

It would be useful to be able to set the default naming rule for images.

I suggest a setting group under "Logging" as follows: 
checkbox "Default image caption"
- unchecked = use filename
- checked = use custom text defined by user, including the use of placeholders (same as log text)

### Is there an alternative?

No at this time.

### Additional Information

_No response_

eddiemuc:
fix #14516: allow edit of log image caption default praefix


moving-bits (commented):
Thanks for diving into this. I have two proposals:

- Can you make the input field a Material Style one? (Colored box around it etc., as in other places)?
- I would name the settings' label something shorter like "Default Caption for Log Images" or so

eddiemuc:
> Thanks for diving into this. I have two proposals:
> 
> * Can you make the input field a Material Style one? (Colored box around it etc., as in other places)?
> * I would name the settings' label something shorter like "Default Caption for Log Images" or so

done and done

Issue #13648, opened 2022-11-13
Emoticons in cache descriptions are extremely small.


eliannelavoie:
### Describe your problem!

Old school emoticons (not emojis) in cache descriptions are much smaller than the text.

![Screenshot_20221112-203858](https://user-images.githubusercontent.com/13744678/201501492-5d5d86e8-200d-4455-ba33-40bdd5971a1f.png)

### How to reproduce?

Open the cache with Geocode GC62QE5.

Go to the description tab.

### Reproducible

Yes

### c:geo Version

2022.10.17

### System information

```text
## System information

c:geo version: 2022.10.17

Device:
-------
- Device type: Pixel 6 Pro (raven, google)
- Available processors: 8
- Android version: 13
- Android build: TP1A.221005.002
- Screen resolution: 1440x2899px (380x765dp)
- Pixel density: 3.40875
- System font scale: 0.9 / used scale: 0.9
- Sailfish OS detected: false
- Google Play services: enabled - 22.43.12 (190400-483592595)

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:82
- Set language: en_CA (system default)
- System date format: y-MM-dd
- Time zone: EST
- Debug mode active: no
- Last backup: never
- Routing mode: Walk
- Live map mode: true
- OSM multi-threading: false / threads: 1
- Map: Quebec (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:primary:Geocaching/maps/quebec.map
  - Atts: Map data (c) OpenStreetMap contributors
  - Theme: none

Filters:
-------
- Hide waypoints: parking 
- LIVE: None ([:inconclusive=false:advanced=false])
- OFFLINE: None ([:inconclusive=false:advanced=false])

Stored Filters (#0):

Services:
-------
- Geocaching sites enabled:
   geocaching.com: Logged in (Login OK) / BASIC
- Geocaching.com date format: M/d/yyyy
- Geocaching.com website language: en-US
- Last successful login on geocaching.com: 2022-11-12 18-33
- Routing: internal / BRouter installed: false
- Installed c:geo plugins:  none

Permissions & paths:
-------
- Fine location permission: granted
- Write external storage permission: granted
[...]

PR #14549, opened 2023-08-26
fix #14521: ensure minimum image size for display

### LINKED ISSUE #14521: Listing Images too small
### Describe your problem!

C:geo's internal Browser is rendering Listing images too small. I think this issue is similar to #13648

### How to reproduce?

Please have a look at the screenshots of https://coord.info/GC15Z0Q I made using Chrome on one hand, and C:geo on the other hand. Chrome's images are looking ok, cause they are about double the size compared to the tiny, hardly recognizable images within c:geo.

shot chrome
shot cgeo

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2023.08.16

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #14521: ensure minimum image size for display

Fix is to apply a minimum size onto all images displayed in listing. Currently hardcoded at 20dp (width + height). Looks now like this:

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/network/HtmlImage.java
+++ b/main/src/main/java/cgeo/geocaching/network/HtmlImage.java
@@ -224,7 +224,7 @@ public class HtmlImage implements Html.ImageGetter {
         if (FileUtils.isFileUrl(url)) {
             return Observable.defer(() -> {
                 final ImmutableTriple<Bitmap, Metadata, Boolean> data = loadCachedImage(FileUtils.urlToFile(url), true);
-                return data != null && data.left != null ? Observable.just(ImmutablePair.of(ImageUtils.scaleBitmapToFitDisplay(data.left), data.middle)) :
+                return data != null && data.left != null ? Observable.just(ImmutablePair.of(ImageUtils.scaleBitmapToDisplay(data.left), data.middle)) :
                         Observable.just(IMAGE_ERROR_DATA);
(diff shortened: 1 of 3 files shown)


eddiemuc:
Note: fix should be tested a while on master before pushing to public - hence I pushed onto master (not release)

eddiemuc:
Additional note: cache GC62QE5 from issue #13648 looks like this:

image

fm-sys:
I guess we should use `sp` instead of `dp`. Have you tested different android font size settings?

ztNFny:
I know of several caches using 1*1px "invisible" images, applying a fixed size would probably break the display.
Can't we scale them relative to text?

eddiemuc:
> I know of several caches using 1*1px "invisible" images, applying a fixed size would probably break the display. Can't we scale them relative to text?

The too-small-images also break the display of course in other ways. 
Problem is that I don't have the text size at the point where the image is scaled. We are talking full-fledged HTML, text could be any size around the image. 20dp seemed to be a good compromise for min image since this is about the size of "normal" text. 

A special exception for the 1x1px case could be build in I guess. We will probably never catch all HTML display variants with our TextView-based HTML solution anyway.

eddiemuc:
Update:
* I applied special handling for 1x1 images -> they will be left untouched
* I used 20sp instead of 20dp, so this should scale with text on other font settings

Issue #14547, opened 2023-08-26
Visualize waypoints and caches belonging together (online/offline)


eddiemuc:
### Tell us your idea!

Base issue is the same as written in #14502: If we do several multi-caches in the same area all waypoints have the same icon (puzzle piece). This way it is not possible to differentiate the waypoints for the different caches. Example (from #14502):

![waypoints](https://github.com/cgeo/cgeo/assets/85786015/624e4467-977c-46ce-915e-c12159ec998f)

The solution proposed and implemented in #14502 is good but works only for caches formerly stored offline and with custom cache symbols assigned. I want a solution which also works for pure online caches (e.g. on live map) and without having to change the cache symbol.

### What solution would you suggest?

Introduce a new setting in "Map Options" named "Display group marker for caches with waypoints". If this setting is chosen then on the map:
* For each cache with waypoints a hash number between 0-15 is calculated from coordinates and geocode e.g. like this: `(ASCII-Code-of-last-char-of-geocode + latituteAsE6 + longituteAsE6) % 16 `  (or something similar)
* Out of a fixed set of 16 symbols one is chosen for the cache
* This symbol is displayed both on the right side of the cache symbol and the belonging waypoints

That way it would be easy to see which waypoints belong to which cache. This would also work for online caches.

Using 16 symbols is just exemplary, we could agree on another (larger or smaller) set

### Is there an alternative?

Not provide the feature

eddiemuc:
Update: symbol is just transiently chosen for map display. Choice won't be persisted on the cache.

eddiemuc:
Would look something like this:
![image](https://github.com/cgeo/cgeo/assets/6909759/9877f083-8464-4042-9f2f-03474cb32b02)

eddiemuc:
Draft PR: #14553

fm-sys:
Hm, this looks IMHO kinda ugly. Why not changing the white waypoint background into different colors?

ztNFny:
In the can 2 weeks ago we agreed to use a marker in the top left to indicate this, something I'm working on in #14502
I find it disappointing that the shared results of that call are now being ignored - why do i even bother spending lots of time finetuning UI components just to have someone ignore that decision sand implement a solution that clashes with existing functionality and goes against that group-decision?!

eddiemuc:
[...]

PR #14553, opened 2023-08-26
rel to #14547: auto-group-function for waypoints to cache on map using markers

eddiemuc:
rel to #14547: draft for auto-group-markers

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/Geocache.java
+++ b/main/src/main/java/cgeo/geocaching/models/Geocache.java
@@ -1616,6 +1616,15 @@ public class Geocache implements IWaypoint {
         return false;
     }
 
+    public boolean hasWaypointsWithCoordinates() {
+        for (Waypoint waypoint : waypoints) {
+            if (waypoint.getCoords() != null) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     public boolean hasGeneratedWaypoints() {
         for (Waypoint waypoint : waypoints) {
             if (waypoint.getWaypointType() == WaypointType.GENERATED) {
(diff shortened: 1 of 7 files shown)


ztNFny:
Mixing this with list markers is a bad idea imho.

eddiemuc:
> Mixing this with list markers is a bad idea imho.

It will only intermix if you have two or more list markers for one cache (because the cache is in two or more lists with marker). Then it get's strange anyway because it is undefined which list markers are the two "selected one's" for display, and which one is left or right. 

I personally use at max one list marker per cache only. In this case there's no mixup

ztNFny:
Well, i use more... 
It'll be quite confusing to have icons with different meaning in the same place.
Also this feature should be optional if it's added at all.

eddiemuc:
> this feature should be optional if it's added at all.

Full agreement. This PR is definitely not merge-ready

eddiemuc:
For what it's worth: this PR would now be ready for merge.

It introduces a new setting:
image

Setting is disabled by default.

If enabled then automarkers are generated on waypoints and cache they belong to on map view (only):
image
image

Marker is chosen via hashcode from an emoji set consisting of all fully opaque circles and square/rectangle symbols:
image
image
image

However, this topic seems to be somewhat contraversal, so I'll leave it at that for now

ztNFny:
See https://github.com/cgeo/cgeo/issues/14547#issuecomment-1694612295 for why i think this PR is a no-go in terms of UX

eddiemuc:
Team consent seems to move in direction of "cache star lines", so I am closing this PR

PR #14554, opened 2023-08-26
rel to #14375: radical approach to move database operations to splash activity

eddiemuc:
rel to #14375: radical approach to move database operations to splash activity

This PR moves all operations in need of a closed DB towards SplashActivity /BackupUtils and allows them only on c:geo startup. This is a rather radical change. Actions are: 
* Auto-Backup
* manual backup
* restore from latest folder
* restore from chosen folder
* moving data from internal to sd card or back

If merged, then all buttons in settings related to such action will only set a flag internally and trigger the following dialog:
image

Then, on next startup, the following dialog will appear:
image

Only here, using this dialog, can the actions actually being executed.

Note that dialogs and GUI are far from finished. Also this change is rather radical and thus should be discussed in the team. For those reasons I mark the PR as draft/WIP

### DIFF
--- a/main/src/main/java/cgeo/geocaching/InstallWizardActivity.java
+++ b/main/src/main/java/cgeo/geocaching/InstallWizardActivity.java
@@ -67,7 +67,6 @@ public class InstallWizardActivity extends AppCompatActivity {
     private WizardStep step = WizardStep.WIZARD_START;
     private boolean forceSkipButton = false;
     private ContentStorageActivityHelper contentStorageActivityHelper = null;
-    private BackupUtils backupUtils;
 
     private static final int REQUEST_CODE_WIZARD_GC = 0x7167;
 
@@ -99,7 +98,6 @@ public class InstallWizardActivity extends AppCompatActivity {
         supportRequestWindowFeature(Window.FEATURE_NO_TITLE);
         setTheme(R.style.NoActionbarTheme);
 
-        backupUtils = new BackupUtils(this, savedInstanceState == null ? null : savedInstanceState.getBundle(BUNDLE_BACKUPUTILS));
         if (savedInstanceState != null) {
             step = WizardStep.values()[savedInstanceState.getInt(BUNDLE_STEP)];
             mode = WizardMode.values()[savedInstanceState.getInt(BUNDLE_MODE)];
@@ -236,9 +234,9 @@ public class InstallWizardActivity extends AppCompatActivity {
                     setButtonToDone();
                     DataStore.resetNewlyCreatedDatabase();
                     if (BackupUtils.hasBackup(BackupUtils.newestBackupFolder(false))) {
-                        backupUtils.restore(BackupUtils.newestBackupFolder(false));
+                        BackupUtils.startAction(this, BackupUtils.StartupAction.RESTORE);
                     } else {
-                        backupUtils.selectBackupDirIntent();
+                        BackupUtils.startAction(this, BackupUtils.StartupAction.RESTORE_OTHER_FOLDER);
                     }
                 }, button3Info, R.string.wizard_advanced_restore_info);
                 break;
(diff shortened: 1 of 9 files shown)


fm-sys:
- we can trigger the clean restart automatically
- cgeo can know which option the user wants to perform based on the previous action. We don't need the separate dialog.

eddiemuc:
> * cgeo can know which option the user wants to perform based on the previous action. We don't need the separate dialog.

Idea would of course be to replace the current many-options in the setting with a single one.

fm-sys:
> > * cgeo can know which option the user wants to perform based on the previous action. We don't need the separate dialog.
> 
> Idea would of course be to replace the current many-options in the setting with a single one.

Although one could argue why we have an own sub-settings page if it only have one option. I would vote to leave it there UI vice. But as sad, this is just UI ;-)

eddiemuc:
I wanted I can leave the GUI and follow @fm-sys s suggestion. Let me think about how to introduce this in a nice way

eddiemuc:
Changed to following workflow:
* When action is requested in settings, then this is registered and user is asked for manual restart:
image

* Upon next restart, user is asked whether the action should really be executed:
image

* Upon "ok" action is executed

I tried automatic restart using ProcessUtils.restartApplication() but somehow my freshly set setting "startupaction" gets lost in the process. Also, restart behaves a bit strange (shows MainActtivity first, then a little bit after the SplashActivity). Can't quite figure this out.

I have to pause/end my approach here due to time restrictions at least for next couple of days. Thus maybe it is better to follow @moving-bits approach.

moving-bits:
I've stumbled across the same problems...

> I tried automatic restart using ProcessUtils.restartApplication() but somehow my freshly set setting "startupaction" gets lost in the process.

which is why I introduced a 1s delay after changing the setting to give the system some time to store the value before exiting. Don't know if there is a more "official" way to register if a setting got persisted.

Maybe JakeWharton's "ProcessPhoenix" has a more viable approach for process restart, have not been able to test yet. (But see their issue 47 and the proposed, but not yet merged, patch, to make it run on recent Android systems.)

> Also, restart behaves a bit strange (shows MainActtivity first, then a little bit after the SplashActivity). Can't quite figure this out.

Same here, don't have a solution for that yet.

eddiemuc:
Probably we require a solution w/o the need of a programmatic app restart. Seems to be an ugly topic

eddiemuc:
A completely different approach comes to my mind:

We could introduces a "read/write lock" concept for all database access:
* we introduce a "multiple-read/single.write lock" for the database object. Whenever an operation on the db is performed, the corresponding thread aquires a "read" lock for the database object from it. The lock is hold until the db operation is finished
* A "database changing" operation (means: one requiring closing and reopening the db) aquires a "write" lock.

That way we could safely do "database changing" operations w/o having to move them to app start (like we do now).

However, I assume this would mean quite some changes to DataStore because all routines accessing the db would be affected

eddiemuc:
...and introducing a lock also introduces the danger of "deadlock"...

fm-sys:
> also somehow upon such restart my preference values are not saved.

I don't know if you have already tried it, but setting preferences is normally done async, however there is a method to persist the preferences synchronously

(I'm on vacation as well so I can only tell what comes to my mind, no dev machine though)

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/issues/14375#issuecomment-1694664730_

eddiemuc:
I will abandon this PR in favor of either @moving-bits s approach (which seems to be more elegant on the UX side) or my other approach in #14557 

@moving-bits @fm-sys @cgeo/team after looking into the "restart app" topic, I stopped believing that the approach to move the db stuff to startup is the right way to go. The "restart app" is clearly not something encouraged/clean in the Android world. I would prefer going the "read-write-lock" way. But if the team thinks different that's also fine for me.

PR #14557, opened 2023-08-27
rel to #14375: lock all database access and differentiate between access and changing using read-write-lock

eddiemuc:
rel to #14375: lock all database access and differentiate between access and changing using read-write-lock

A different approach to solve #14375: introduce a read-write-lock-mechanism onto DataStore:
* each access to the database requires a read-lock
* each "database-changing" access which requires a "close" (e.g. backup restore move) requires a write-lock

That way in theory all dangling cursors and insert/update/delete operations should be finished when the "write" lock is aquired and close/reopen should be save.

Would that be the better way to solve #14375? Of course this requires discipline when introducing new database interaction - it must always aquire a lock.

@moving-bits

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarActivity.java
@@ -456,7 +456,7 @@ public abstract class AbstractNavigationBarActivity extends AbstractActionBarAct
         try (ContextLogger cLog = new ContextLogger(Log.LogLevel.DEBUG, "AbstractBottomNavigationActivity.runInitAndMaintenance")) {
 
             //check database
-            final String errorMsg = DataStore.initAndCheck(false);
+            final String errorMsg = DataStore.initAndCheck();
             if (errorMsg != null) {
                 DebugUtils.askUserToReportProblem(this, "Fatal DB error: " + errorMsg);
             }
--- a/main/src/main/java/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DataStore.java
@@ -80,6 +80,7 @@ import android.net.Uri;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
+import androidx.core.util.Supplier;
 
 import java.io.File;
 import java.io.FilenameFilter;
@@ -103,6 +104,7 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
@@ -233,6 +235,7 @@ public class DataStore {
      */
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
+    private static final ReentrantReadWriteLock databaseLock = new ReentrantReadWriteLock();
     private static final int dbVersion = 100;
     public static final int customListIdOffset = 10;
 
@@ -552,9 +555,10 @@ public class DataStore {
     }
 
     public static int getActualDBVersion() {
-        init();
-        return database == null ? -1 : database.getVersion();
-
+        return withAccessLock(() -> {
+            init();
+            return database == null ? -1 : database.getVersion();
+        });
     }
 
     public static boolean versionsAreCompatible(final SQLiteDatabase databaseToCheck, final int oldVersion, final int newVersion) {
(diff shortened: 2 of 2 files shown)


eddiemuc:
Since in the diff the change looks so huge, a quick exmplanation what was done in DataStore:
* Introduced a ReentrantReadWriteLock "databaseLock" as a "guard" for the "database" object
* Introduced methods "withChangeLock" taking a Runnable or Supplier. Those are then executed in a "write lock" context
* Likewise, introduced methods "withAccessLock" to execute code in a "read lock" context
* Surrounded all database-changing functions (=all with "closeDb" need: backup, restore, move) with "withChangeLock"
* Surrounded all database-acccessing methods (read, insert, update, delete) with "withAccessLock". This accounts for the majority of the changes

In theory this should work.

moving-bits:
Looks good to me on a quick glance. Some questions for better understanding the solution:
- the app can have multiple access locks at the same time, but only one change lock (and access locks and change locks are mutually exclusive)
- is there some timeout involved when trying to acquire a lock that is currently not available? How will the app react if it does not get the lock? (Will the "connection pool closed" errors be replaced by "cannot acquire lock" errors?)
- we will have to surround every non-exclusive db access (reading/writing data, but not moving the database etc.) with `withAccessLock(() -> { ... })` or `return withAccessLock(() -> { ... })`?
- nesting such `withAccessLock`-frames will not cause harm?

Still I'm undecided which solution to prefer.
- This PR seems to be a well-thought and clean technical solution. Needs some discipline when introducing new DB interaction, but that's fine.
- The "do all exclusive DB stuff at startup" variant OTOH is temporarily blocking the user from doing other stuff in c:geo during that time (most of which would be requiring interaction with database anyway, which is not available during that action), so it may be easier to handle from both coding and UX point of views.

I don't have a personal preference for either solution yet.

eddiemuc:
Hi @moving-bits 

A read-write-lock is a very generic concept of computer science e.g. explained here: https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock. To cite from there: "An RW lock allows [concurrent](https://en.wikipedia.org/wiki/Concurrency_control) access for read-only operations, whereas write operations require exclusive access.". 

> * the app can have multiple access locks at the same time, but only one change lock (and access locks and change locks are mutually exclusive)

Correct. "access lock" are "read locks", and "change lock" is a "write lock" in that sense.

> * is there some timeout involved when trying to acquire a lock that is currently not available? How will the app react if it does not get the lock? (Will the "connection pool closed" errors be replaced by "cannot acquire lock" errors?)

If a thread wants to aquire a lock then the thread blocks until the lock becomes available. Means:
* if another thread holds a write lock, then all threads wanting to aquire a read or write lock will be blocked until the lock is released
* If some threads hold a read lock and another thread wants a read lock then it is granted immediately though

There's no timeout, threads will wait "forever" until the lock is free. And yes, if the lock is never free'd the waiting threads will freeze.

I used the Java "ReentrantReadWriteLock" in "fair mode". This class is documented here: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html 

This class has a so-called "fair-mode" where the policy of which thread gets locks in which order is determined. We can use this if wanted. Currently I use the default "unfair mode".

> * we will have to surround every non-exclusive db access (reading/writing data, but not moving the database etc.) with `withAccessLock(() -> { ... })` or `return withAccessLock(() -> { ... })`?

Correct

> * nesting such `withAccessLock`-frames will not cause harm?

No because this is a so-called "Reentrant"-lock. Means it can handle nested locks/unlocks from same thread. However, each "lock" must be paired with a call to "unlock". E.g. if by recursion the same thread requires a lock 3 times, then there must also be 3 calls to unlock before the thread releases the lock.

> * The "do all exclusive DB stuff at startup" variant OTOH is temporarily blocking the user from doing other stuff in c:geo during that time (most of which would be requiring interaction with database anyway, which is not available during that action), so it may be easier to handle from both coding and UX point of views.

Correct, but only database interaction is blocked. Which is valid because the database is moved/backed-up/restored in that phase, so database interaction SHOULD be blocked.

eddiemuc:
@moving-bits @cgeo/team I suggest merging this PR and observe in nightlys whether:
- the approach really fixes the database "closed" problem, and
- does not create thread/app freezes (it should not but one can never be 100% sure with locks/synchronizes)

In case the second one becomes a problem, the locking can easily be deactivated in the central withAccessLock/withchangelock functions.

Ok to merge?

fm-sys:
> I suggest merging this PR and observe in nightlys whether: 
> * the approach really fixes the database "closed" problem

We will only know this as soon as it's out in a beta, as we don't get Google crash reports for nightlys. Anyway, this shouldn't be a blocker for merging, even the other way around ;-)

Another thought going in a different direction: Could it theoretically be that this crash does happen while c:geo gets killed by the OS while a DB request is running?

moving-bits:
> Ok to merge?

I'd vote for merge. Let's test it a couple of days in the nightlies whether something more obvious comes up, then prepare a beta version (because only those logs show up in Play Console).

eddiemuc:
So be it

Issue #14514, opened 2023-08-16
UI efficiency improvement - edit log image


andrixnet:
### Tell us your idea!

Currently the UI for adding a log image is reasonable, with the necessary fields and works ok. 

Logging multiple finds and adding images to each of them I discovered the UI makes it somewhat cumbersome. 
My use case was simple: load an already existing image (photos taken earlier), add a caption, end.
The UI requires me to either scroll or close the keyboard to finalize this form, because the "OK" button is hidden by the keyboard. 

Keyboard open, covering "OK" button.
![Screenshot_20230814-160705_cgeo](https://github.com/cgeo/cgeo/assets/3382777/339ba99d-b473-43ab-9fff-dbea7cff5b39)

Keyboard closed, "OK" button visible.
![Screenshot_20230814-160711_cgeo](https://github.com/cgeo/cgeo/assets/3382777/0e777004-a774-4b5e-aca6-13dd136e2c95)

### What solution would you suggest?

Since the most often use case is only edit the caption, it would be more efficient to have access to the "OK" button above the area of the screen where the soft keyboard appears. 

My suggestion would be above the "existing" and "new" buttons, with the logic that most action buttons in c:geo are on the top part of the screen.

### Additional Information

User desiring to edit description still has scroll or "next" button on keyboard.
User desiring to change image size still has scroll. 
But these controls I believe are much less often accessed.

ztNFny:
The 3 buttons could be moved to a toolbar, this would match the UI we have in log screen

fm-sys:
> The 3 buttons could be moved to a toolbar, this would match the UI we have in log screen

...waypoints, personal note and many other places. Sounds consistent - like it!

eddiemuc:
I agree with the request. I believe that the "edit (log) image" dialog could use a complete UX makeover.

Considering the new image edit functions requested in #14515, the "edit image" dialog needs to provide the following UX elements:

* Central Element: the image itself
* Additional edit attributes:
  * caption
  * description
  * Resolution selection
* Actions on image
   * select another image from camera (today a text button with label "new")
   * select  another image from storage (today a text button with label "existing")  
   * rotate left / right / 180 degree
   * flip horizontal / vertical
   * crop (thats optional, but I would find it nice...)
* Actions closing the dialog:
  * OK: apply changes to (log) image
  * Cancel: revoke all changes and close dialog
  * Delete: delete the image (in log image list) and close dialog

[...]

PR #14561, opened 2023-08-28
rel to #14514: redesign GUI for image edit

eddiemuc:
rel to #14514: redesign GUI for image edit

Following the layout discussed in the issue, when merging this PR layout will look like this:
* "delete" was removed completely, "ok/cancel" was moved into action bar using symbols
* title was changed from cache title to constant "edit image"
* order of input elements was changed to the order suggested in issue

image

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -732,7 +732,7 @@
             android:windowSoftInputMode="stateHidden" >
         </activity>
         <activity
-            android:name=".ImageSelectActivity"
+            android:name=".ImageEditActivity"
             android:label="@string/log_image" >
         </activity>
 
--- a/main/src/main/java/cgeo/geocaching/ImageEditActivity.java
+++ b/main/src/main/java/cgeo/geocaching/ImageEditActivity.java
@@ -33,13 +33,14 @@ public class ImageEditActivity extends AbstractActionBarActivity {
 
     private static final int RC_EDIT_IMAGE_EXTERNAL = 50;
 
+    private static final int ANIMATION_DURATION_IN_MS = 200;
+
     private static final String SAVED_STATE_IMAGE = "cgeo.geocaching.saved_state_image";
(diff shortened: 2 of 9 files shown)


eddiemuc:
* now also contains animation for image rotate/flip

fm-sys:
Some suggestions (note that this is just my personal opinion):

- I'm not sure if we are consistent with it in cgeo, but it should be "Edit image" for English language (only first word capitalized)
- didn't we want to remove the "add from camera/galary" buttons?
- place the "flip/rotate/edit" button bar horizontally centered
- add some padding below and above the scale selection so that it doesn't look too narrow

eddiemuc:
> * I'm not sure if we are consistent with it in cgeo, but it should be "Edit image" for English language (only first word capitalized)
> * place the "flip/rotate/edit" button bar horizontally centered
> * add some padding below and above the scale selection so that it doesn't look too narrow

I implemented those (and I also found a bug that image background was not transparent, which is fixed now). Dialog now looks like this:

image

eddiemuc:
> * didn't we want to remove the "add from camera/galary" buttons?

Did we? I might have overlooked that. But why should they get removed? (in the context of "edit image" those buttons have of course not an "add" semantic but a "replace" semantic)

I can remove them if wanted though

fm-sys:
These buttons are a legacy of the time were users could only have one image per cache. IMHO they don't make any sense anymore, especially as you had also removed the delete button which might have had a little more reason to still exist

fm-sys:
### Off topic - My somewhat radical option about UI design:

Every additional button hurts, as it costs time for the user understanding that he doesn't need it.

For sure this really doesn't fit to c:geo in many terms, but I can recommend the book "Insanely Simple: The Obsession that Drives Apple's Success"

I'm often frustrated that we don't manage to build such clean UI in c:geo, but I've learned to come over it. Still trying to minimalize retundancy whenever it is possible ;)

eddiemuc:
,Buttons are removed:

image

eddiemuc:
Ok to merge?

fm-sys:
Let's test and finetune afterwards if necessary...

PR #14564, opened 2023-08-29
optimize lint configuration

bekuno:
- change paths to changed c:geo "main" structure
- add ignore filter for changelog files
- compact filter for rating files
- add an IconDuplicates ignore entry

@moving-bits Adding the routing_dummy.brf file to c:geo source was intentional, wasn't it?
@ztNFny Using the same attributes_kids* file was  intentional, wasn't it?

### DIFF
--- a/main/lint.xml
+++ b/main/lint.xml
@@ -22,6 +22,11 @@
 
     <issue id="IconDensities" severity="ignore" />
 
+    <issue id="IconDuplicates" severity="ignore">
+        <!-- planned duplicate -->
+        <ignore path="src/main/res/drawable-xhdpi/attribute_kids*.png" />
+    </issue>
+
     <!-- unfinished localizations are okay since we always have English as fallback -->
     <issue id="MissingTranslation" severity="ignore" />
     <!-- old translation will be deleted by the next crowdin import -->
--- a/main/lint.xml
+++ b/main/lint.xml
@@ -56,9 +56,6 @@
         <!-- changelog files are referenced depending of release phase -->
         <ignore path="src/main/res/raw/changelog_*.md" />
 
-        <!-- part of the internal brouter -->
-        <ignore path="src/main/res/raw/routing_dummy.brf" />
-
         <!-- translations no longer used -->
         <ignore path="src/main/res/values-*/strings.xml" />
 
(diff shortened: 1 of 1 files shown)


ztNFny:
> @ztNFny Using the same attributes_kids* file was intentional, wasn't it?

I didn't really touch the icons, I just generated them from the existing sources in higher quality. Dunno about the history of those.

> * compact filter for rating files

I had that before but lint didn't like it.

bekuno:
> > @ztNFny Using the same attributes_kids* file was intentional, wasn't it?
> 
> I didn't really touch the icons, I just generated them from the existing sources in higher quality. Dunno about the history of those.

Ok. I now remember back when I updated the Cache attribute icons.
The KIDS* have different ACODE (for OC) with the same mappint to GC ID, but different use in OCDE and OCPL.
(see also https://wiki.opencaching.eu/index.php?title=Cache_attributes#cite_ref-:0_7-0)
So the same icon is fine.

> > * compact filter for rating files
> 
> I had that before but lint didn't like it.

Now the wildcard filter is working :-)

moving-bits:
> @moving-bits Adding the routing_dummy.brf file to c:geo source was intentional, wasn't it?

Yes, it's part of BRouter's new "elevation only" mode introduced with v1.7.x

bekuno:
> > @moving-bits Adding the routing_dummy.brf file to c:geo source was intentional, wasn't it?
> 
> Yes, it's part of BRouter's new "elevation only" mode introduced with v1.7.x

But shouldn't then this file be referenced somewhere?
It should probally defined in `cgeo.geocaching.brouter.util.DefaultFilesUtils.DefaultFiles`, I would say.

bekuno:
> > > @moving-bits Adding the routing_dummy.brf file to c:geo source was intentional, wasn't it?
> > 
> > 
> > Yes, it's part of BRouter's new "elevation only" mode introduced with v1.7.x
> 
> But shouldn't then this file be referenced somewhere? It should probally defined in `cgeo.geocaching.brouter.util.DefaultFilesUtils.DefaultFiles`, I would say.

@moving-bits 
I removed now the entry for remove routing_dummy.brf from UnusedResources ignore list.
So you can check this later on.

PR #14567, opened 2023-08-30
Navigate up when using back key

fm-sys:
Navigate up when using back key and show different icon for "(one level up)" folder.

From time to time, I was struggling with the fact that back key finishes the download activity instead of navigating up. This annoyed my every time so much that I finally fixed this thing.

As I introduced a special `isBackDir` flag anyway, I also gave it a new dedicated icon on the fly:

grafik

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/downloader/DownloaderTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/downloader/DownloaderTest.java
@@ -30,7 +30,7 @@ public class DownloaderTest {
     private static int count(final List<Download> list, final boolean isDir) {
         int i = 0;
         for (Download d : list) {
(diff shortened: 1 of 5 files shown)


moving-bits:
Back key works fine now, thanks for diving into this.

How shall we handle the "back arrow" in the action bar? Currently it still finishes the activity and returns to the map. (At least for me in the emulator.) - Shouldn't this move one folder up as well? (Similar to handling in settings.)

fm-sys:
Can easily be done of course. Just wasn't sure if this is wanted...

moving-bits:
> Can easily be done of course. Just wasn't sure if this is wanted...

I'd vote for "yes", as we have similar behavior in settings (arrow icon moves one level up).

PR #14576, opened 2023-09-03
Distinguish between default and extended config mode

moving-bits:
## Description
Show fewer settings in default config mode, as proposed in [internal discussion](https://github.com/orgs/cgeo/discussions/78).
To toggle extended settings, see switch "extended settings" in system's config.

image.image

Some things are not yet working fully:
- ~some descriptive texts are missing yet~
- ~"Offline settings" cannot be reached currently~
- ~toggling basic/extended settings requires restart of current settings screen~
- ~search index has not been tested with yet~
- toggling basic/extended settings closes system settings

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/PreferenceTextAlwaysShow.java
+++ b/main/src/main/java/cgeo/geocaching/settings/PreferenceTextAlwaysShow.java
@@ -1,0 +1,34 @@
+package cgeo.geocaching.settings;
+
+import android.content.Context;
+import android.util.AttributeSet;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.preference.Preference;
+import androidx.preference.PreferenceViewHolder;
+
+/**
+ * Behaves exactly the same as Preference, only needed to have a separate class to compare against
+ */
+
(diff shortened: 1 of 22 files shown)


ztNFny:
2 suggestions:
- instead of "expert mode" being a regular setting could it be shown in settings dialog toolbar as a button? That'd allow quick toggling
- instead of the hardcoded list of settings could that be an attribute in the settings.xml files for each setting?

moving-bits:
> * instead of "expert mode" being a regular setting could it be shown in settings dialog toolbar as a button? That'd allow quick toggling

Had thought about this as well in the beginning, but decided against it, as it would interfere with search in settings. And "show extended settings" is a "set once and forget" setting for most people.

> * instead of the hardcoded list of settings could that be an attribute in the settings.xml files for each setting?

Had tried this as well, but that would require to gather the attribute value on preference instantiation, and that would require every setting we use to be derived from some c:geo-specific base class (which catches instantiation and reads the specific attribute value). At least I do not know of a different way to get the attribute value from a preference.
OTOH it's just a couple of basic preferences left, so maybe a hardcoded list is ok?

moving-bits:
retest this please

Issue #14573, opened 2023-09-02
Make "system default" theme the default 


fm-sys:
>> I'll have to add that we still use "dark" instead of "system default" as default. This can be changed of course...
>> 
>> But if we want to change the default theme, then maybe timely align it with other design changes (#14570) so that users only get confused once ;-) 
>
> Did not remember that, but you are right. Anyway we should not play around with that default IMHO.

It has historic reasons why c:geo uses dark mode as default. Back in the days the option "system default" didn't exist at all and light mode just had looked totally ugly.

Nowadays, users who appreciate dark themes have most probably turned their whole phone dark, while there is no reason to force users to a dark theme while they actually don't like it.

My vote clearly goes to automatic theme as default. Any opinions?

(extracted the discussion from #14569)

Lineflyer:
AFAIR we discussed that in the past as we decided to implement the automatic settings and decided to still stay with black as default as it is kind of a branding since ever in c:geo.

fm-sys:
I also thought that there was some kind agreement about it, but couldn't find anything related. To me it seems as if we just continued to use dark as the default at the time, without any discussion. (#10794) 

That's why I opened this issue. Apart from branding and habit, there is actually no real reason why c:geo should be dark IMHO. But I'm interested how others think about it...

moving-bits:
I propose to use "automatic" for new installations. This won't affect existing users, even not when migrating to a new device (as long as they restore their settings from backup).

(Even though I'm using dark mode myself, it looks somewhat outdated compared to more recent apps. We could think of introducing _a few_ more colors into c:geo, chosen from a harmonized color palette, to make it look a bit more up to date visually... But that's a different discussion.)

moving-bits:
> That's why I opened this issue. Apart from branding and habit, there is actually no real reason why c:geo should be dark IMHO. But I'm interested how others think about it...

There are OLED displays out there consuming less power when using dark colors, that's _one of_ the reasons why some people are using dark mode for their system / for their apps.

fm-sys:
> There are OLED displays out there consuming less power when using dark colors, that's one of the reasons why some people are using dark mode for their system / for their apps.

Totally agree! But these users have system-wide dark mode enabled anyway ;-)

>  looks somewhat outdated compared to more recent apps. We could think of introducing a few more colors into c:geo, chosen from a harmonized color palette, to make it look a bit more up to date visually... But that's a different discussion.

Let's discuss this in context of #14570

Issue #14570, opened 2023-09-01
Migrate to material3


fm-sys:
### Tell us your idea!

Material Design v3 (aka Material You) is published for almost a year now, so I guess most early issues are solved meanwhile. Furthermore it becomes more and more common and gets adopted by other really popular apps. 

### What solution would you suggest?

Switch to m3

### Is there an alternative?

Let c:geo look outdated again ;-) 

### Additional Information

![01092023081132](https://github.com/cgeo/cgeo/assets/64581222/3d69cacc-7646-4275-9855-6006274bac0a)

eddiemuc:
What does this mean implementation-wise? Is it "just" switching icons and a library (to use new GUI components), or is there more work to do?

fm-sys:
The icons even doesn't have changed significantly. For sure, from time to time some icons get some small tweaks, but nothing we really have to care about.

In theory it's just changing the style parent from `MaterialComponents` to `Material3`, but probably it'll be a bit more work.

https://material.io/blog/migrating-material-3

Wer could implement `DynamicColors` as well (Material **You**, coloring based on launcher wallpaper colors), but that would be much more work.

moving-bits:
I've done a quick test with replacing the base themes (when available). Some first findings (TBC):

![image](https://github.com/cgeo/cgeo/assets/3754370/a1dc9bc2-ba01-418e-9723-127144ed00f5).![image](https://github.com/cgeo/cgeo/assets/3754370/e0986019-6ff2-4fc5-aea9-20681f202309).![image](https://github.com/cgeo/cgeo/assets/3754370/ffb50812-7351-4005-93c6-557136af13bd).![image](https://github.com/cgeo/cgeo/assets/3754370/658ea193-d8f7-418a-b7dd-0318467461b6)

fm-sys:
> (Even though I'm using dark mode myself, it looks somewhat outdated compared to more recent apps. We could think of introducing a few more colors into c:geo, chosen from a harmonized color palette, to make it look a bit more up to date visually... But that's a different discussion.)

_Originally posted by @moving-bits in https://github.com/cgeo/cgeo/issues/14573#issuecomment-1704068109_
           
Material 3 is designed to look quite nice if implemented right. I would like to take a closer look at this as time permits... Maybe in the middle of next week

moving-bits:
> I would like to take a closer look at the coloring as time permits... Maybe in the middle of next week

Highly appreciated. Maybe you can push directly to PR #14574?

[...]

PR #14580, opened 2023-09-06
Material3 migration

fm-sys:
closes #14574, closes https://github.com/cgeo/cgeo/issues/14573
rel. to #14570

See #14570 for screenshots (they are not 100% up to date though). 

I would recommend merging to master rather quickly, as it can easily conflict with other UI changes cause of color resource rename, etc. Furthermore, it's nice to get nightly feedback rather quickly...

Just for reference: I also changed the default theme to "follow system", new users and everyone who didn't switch the preference at least once will be affected (=> everyone who hasn't even noticed that there are different themes). Reason for that decision: For the first time, c:geo is mostly optimized for light theme. Of course, I checked the dark variant as well, but all optimization and fine-tuning was done while looking at the light version, so it now looks really acceptable compared to older light theme versions. Sorry for not waiting for a final agreement in #14573, still I think it's the right thing to do. System-wide dark theme was introduced with Android 10, and since then it is recommended to follow the system theme as default option. (https://developer.android.com/develop/ui/views/theming/darktheme#change-themes)

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -417,7 +417,7 @@ dependencies {
     configurations.all*.exclude module: "play-services-measurement"
 
     // Support Library Design
-    implementation 'com.google.android.material:material:1.6.1'
+    implementation 'com.google.android.material:material:1.9.0'
 
     // ProcessPhoenix, reliable process restart
     // implementation 'com.jakewharton:process-phoenix:2.0.0'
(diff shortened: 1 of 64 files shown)


moving-bits:
Some small quirk I've stumbled upon:

Cache list shows an extra line at the bottom. The space between those two lines is clickable, but without any action.

image

This line is visible only when just a few caches are part of the list, so that the screen is not filled completely.

moving-bits:
Let's give it a try in the nightlies. Thanks for your efforts.

PR #14587, opened 2023-09-08
fix #14579: add difficulty terrain matrix filter

### LINKED ISSUE #14579: Filter for missing T/D in 81er Matrix
### Tell us your idea!

It would be very nice to have a filter which shows all caches which help to fill the 81er matrix one more time. 

I could be a checkbox in the T/D filter, like on the Website.

### Is there an alternative?

_No response_

### Additional Information

The website have this filter too, so may that can be used?

eddiemuc:
fix #14579: add difficulty terrain matrix filter

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/capability/IDifficultyTerrainMatrixNeededCapability.java
+++ b/main/src/main/java/cgeo/geocaching/connector/capability/IDifficultyTerrainMatrixNeededCapability.java
@@ -1,0 +1,16 @@
+package cgeo.geocaching.connector.capability;
+
+import cgeo.geocaching.connector.IConnector;
+
+import androidx.annotation.NonNull;
+
+import java.util.Collection;
+
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+/** capability to provide the needed difficulty-terrain-combinations to fill the 81-er-matrix */
+public interface IDifficultyTerrainMatrixNeededCapability extends IConnector {
+
+    @NonNull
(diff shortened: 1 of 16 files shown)


eddiemuc:
retest this please

ztNFny:
Shouldn't this replace the existing D/T filter? I don't see a point in having 2 filter GUIs for the same criteria.

eddiemuc:
> Shouldn't this replace the existing D/T filter? I don't see a point in having 2 filter GUIs for the same criteria.

I would definitely leave the old filter GUI for the "simple" filter view - the matrix is far too complex for common users IMHO. We could however think about removing the simple view from the "advanced" mode or present a combined view (which would then get more complex of course). Maybe something to discuss in team meeting?

ztNFny:
Is it more complex though? From looking at the screenshot I expect selecting "T4, D*" is one click:
image
That's actually less clicks than the old sliders where you had to set both min and max.

Showing T4-T5 would be 3 clicks now, more clicks, but not "complex" imho. I feel like the matrix-view makes it easier to get a WYSIWYG feeling, making it less complex.

Let's take it for the meeting.

eddiemuc:
There's another problem: unlike the "simple" D/T filter, the "matrix" filter is not supported by most geocaching servers for online filtering. AFAIK just gc.com supports the online search for D-t-combinations; okapi for example just allows separate (range) search for difficulty and terrain.

eddiemuc:
Current screenshots after updates with feedback (screenshot design has now with material3 look due to parallel changes, but this is unrelated to this PR):

image

eddiemuc:
As discussed: remove "online" gc.com option for now and shorten button text

PR #14588, opened 2023-09-08
Use fragment backstack on navigateUp in settings (fix #14552)

### LINKED ISSUE #14552: Bottom navigation done wrong - problems with custom bottom navigation items
### Describe your problem!

I like the idea of having the last bottom navigation icon customizable, however it is done completely wrong. 

### How to reproduce?

- set the last item e.g. to "settings" 

When clicking on the last item the bottom navigation disappears. But even worse, there is no way to get back. When using the back button cgeo gets closed completely. 

- Bottom navigation should always stay visible after using it

- c:geo should guide to the dashboard first before exiting 

### Reproducible

Yes

### c:geo Version

Since July 2023 (nightly) 

### System information

_No response_

### Additional Information

Introduced in  #14419

moving-bits:
## Description
Use fragment backstack when navigating up in `SettingsActivity` and fall back to default `navigateUp` behavior, if no fragment left on backstack

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/SettingsActivity.java
+++ b/main/src/main/java/cgeo/geocaching/settings/SettingsActivity.java
@@ -141,6 +141,14 @@ public class SettingsActivity extends CustomMenuEntryActivity implements Prefere
         setResult(NO_RESTART_NEEDED);
     }
 
+    @Override
+    public boolean onOptionsItemSelected(final MenuItem item) {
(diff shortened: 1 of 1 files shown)


fm-sys:
Even this PR doesn't fix the problem when using the system back key...

fm-sys:
https://github.com/cgeo/cgeo/assets/64581222/59bfdab9-684c-4a9a-b36c-2a96dcdf073f

moving-bits:
> Even this PR doesn't fix the problem when using the system back key...

seems I need to do the same "trick" for `onBackPressed()` - PR is coming

Issue #7911, opened 2019-10-23
BetterCacher implementation


BetterCacher:
_We want to suggest to implement BetterCacher in c:geo._

**What is BetterCacher?**
At BetterCacher you will find special geocaches. We'd like to show you which caches are WHY great. We think that a geocache that has been placed with much love and work deserved a few words about it instead of just clicking on a number or a star. 
BetterCacher is free for everyone, is a community project and exist since 2009.
You can find out more on our website: [bettercacher.org ](https://bettercacher.org)

**How can the implantation look like?**
An implantation could take place under the attributes. If the geocache is listed on BetterCacher, an attribute could be displayed with the BetterCacher logo ([link to the picture](https://bettercacher.org/blatt.gif); feel free to use it). By clicking on the attribute, a short explanation about BetterCacher and the link to the description (forum) could be displayed. The forum describes what exactly makes this cache so special.
It may also be possible to implement a filter to specifically search for BetterCacher listed caches.

**How can data be retrieved?**
BetterCacher provides an API: [api.bettercacher.org ](https://api.bettercacher.org)
Anyone can use it for free under CC BY-NC-SA.
There you can find out if a caches is listed or not. If a cache is listed, some data are outputted.

**Why should BetterCacher be implemented in c:geo?**
- With BetterCacher geocachers are able to tell if and why the cache is special. You can specifically search for great geocaches.
- With the implementation, access to BetterCacher would also be possible via app (in addition to the website).

We are available for further questions at info@bettercacher.org.

okainov:
What about Premium? I just checked the map on BetterCacher and out of 6 caches I see around Munich 3 are in fact Premium-only. Do you provide some kind of filters there?

BetterCacher:
Thanks for your reply. None of our users has so far requested an indication whether the caches are available to basic members. I fail to see why this would be an issue regarding the integration of the bettercacher flag in c:geo, because your app already provides that information, doesn't it?

bekuno:
As a general implementation idea
- add a service entry for bettercacher in the c:geo settings (like GCVote)
- when opening a cache in the detail view, load the status from the bettercacher api if the service is active
- if found, add a link to the Bettercacher cache forum page, otherwise a link to the website where it could be added
- possibly the cache text from the bettercacher forum can be displayed in c:geo as another tab
- Saving the forum ID when saving the cache locally

I don't see that at the moment it is possible to 
- Creating a bettercacher entry directly from c:geo (new entry or comment)
[...]

PR #14591, opened 2023-09-09
rel to #7911: change to new cgeo specific fulldata API

eddiemuc:
rel to #7911: change to new cgeo specific fulldata API

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/bettercacher/BetterCacherAPI.java
+++ b/main/src/main/java/cgeo/geocaching/connector/bettercacher/BetterCacherAPI.java
@@ -9,7 +9,6 @@ import cgeo.geocaching.models.Category;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Tier;
 import cgeo.geocaching.network.HttpRequest;
-import cgeo.geocaching.network.HttpResponse;
 import cgeo.geocaching.utils.CollectionStream;
 import cgeo.geocaching.utils.CommonUtils;
 import cgeo.geocaching.utils.ContextLogger;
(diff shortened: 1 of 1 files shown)


eddiemuc:
Note: this PR should be merged rather quickly. Until then bettercacher.org integration uses the "full API" to gather missing caches on map search (including e.g. logs).

main/src/main/java/cgeo/geocaching/connector/bettercacher/BetterCacherAPI.java
```diff
@@ -292,9 +237,9 @@ public Geocache createGeocache() {
             cache.setCoords(GeopointParser.parse(lat + " " + lng, null));
             // "type": "unknown",
             cache.setType(getFromJson(CACHE_TYPE_MAP, type, CacheType.UNKNOWN));
-            // "difficulty": "4_5",
+            // "difficulty": "4.5",
```
Lineflyer, line 240: Just my few cents:
Is this a left over or for documentation purpose. For the latter it should probably be explained, what it shall tell.

main/src/main/java/cgeo/geocaching/connector/bettercacher/BetterCacherAPI.java
```diff
@@ -292,9 +237,9 @@ public Geocache createGeocache() {
...
             cache.setDifficulty(getNumberFromJson(difficulty, 0));
-            // "terrain": "1_5",
+            // "terrain": "1.5",
```
Lineflyer, line 242: see above

main/src/main/java/cgeo/geocaching/connector/bettercacher/BetterCacherAPI.java
Lineflyer: Ah OK :)

PR #14595, opened 2023-09-10
Issues/fix last saved log 14371

### LINKED ISSUE #14371: "Insert last log" discarded when adding image first

eddiemuc:
fix #14371: save last log only on leaving log activity (not on save log)


eddiemuc:
Git fail

Issue #14603, opened 2023-09-13
Cannot access offline data settings


Lineflyer:
### Describe your problem!

With current nightly 2023.09.13 I am not able to access the settings for "offline data".

### How to reproduce?

- c:geo Home screen
- Menu
- Settings

### Actual result after these steps?

The item "offline data" is greyed out, tapping doesn't do anything

### Expected result after these steps?

Normal access to the settings behind "offline data"

### Reproducible

Yes

### c:geo Version

2023.09.13-NB

Lineflyer:
Is this a result of the new "extended settings" option?
If yes, I suggest we should either default to extended settings=on for existing installations or at least if the user has used those settings before. Its highly unexpected and looks like a loss of functionality or a bug.

moving-bits:
> Is this a result of the new "extended settings" option?

yes

> If yes, I suggest we should either default to extended settings=on for existing installations 

that I would like to avoid as I guess many (most?) users are sort of overwhelmed by the multitude of options c:geo offers. The "pro" user can easily activate extended settings once and c:geo will remember their choice.

> or at least if the user has used those settings before.

How to detect that? (Which setting to use as indicator?)

> Its highly unexpected and looks like a loss of functionality or a bug.

IMHO the more stringent alternative would be to hide the offline data settings at all, same as we do with all other extended settings when extended mode is disabled.

fm-sys:
IMHO the more stringent alternative would be to hide the offline data settings at all, same as we do with all other extended settings when extended mode is disabled.

Would be more consistent!

---



A general thing: I don't like the reduced settings in the current form at all. The main problem is, that it's not intuitive enough to enable "advanced mode". IMHO it's a must have to have this switch on top of the main settings screen.

Lineflyer:
I perfectly understand the background of the change to create a basic set of settings.
I am somehow only struggling with the consequence, that also expert users will suddenly miss some options (and see a disabled main category....but that's secondary).
 
They will need to manually find the enabled of advanced settings down in the settings sub menu.
Most apps I know offers such switch quite prominently. How about having a big switch "Basic/Advanced" on the setting menu titlebar instead of burying it deep down below?

Sorry, if this was already discussed (did not follow the initial discussion but trying to get more involved again now).

fm-sys:
Proposal / Implementation:

[...]

PR #14606, opened 2023-09-13
make extended settings more prominent

fm-sys:
rel to #14603

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceSystemFragment.java
+++ b/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceSystemFragment.java
@@ -31,11 +31,6 @@ public class PreferenceSystemFragment extends BasePreferenceFragment {
 
         findPreference(getString(R.string.pref_persistablefolder_testdir)).setVisible(BranchDetectionHelper.isDeveloperBuild());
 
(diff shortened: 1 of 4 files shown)


Lineflyer:
I am really impressed by the speed you provided this PR. :)
Please take a look into the discussion in #14606 before we implement this improvement though.

fm-sys:
Well, I had coded this as visual demo anyway, and it felt wasteful to just throw it away. If we come up with an even better solution, I'm fine with it as well :)

moving-bits (approved):
Nice and clean solution, without the side-effects of using the action bar here. Thanks for diving into this.

moving-bits:
Only thing that comes to my mind: Do we really need a "Settings" subheading (category) - the whole page is about settings. It looks a bit lost, IMHO.

fm-sys:
> Do we really need a "Settings" subheading (category)

I added it to have a clean separation between the "filter" and the real preferences, and to make the filter toogle even more prominent. IMHO it looks way worse if it's just part of it. But that's just my personal taste...

moving-bits:
How about this?

image

fm-sys:
I don't have a strong opinion about this. Anyone else?

eddiemuc:
> I don't have a strong opinion about this. Anyone else? 

Me neither

PR #14609, opened 2023-09-14
Hide offline data settings in basic config mode

moving-bits:
## Description
Hide offline data settings in basic config mode instead of only disabling them (to be consistent with other extended settings)


moving-bits:
failing test unrelated, GeoKrety doing their scheduled update?

retest this please

PR #14617, opened 2023-09-16
rel to #7911: fix geocaching merge problem with tier and categories

eddiemuc:
rel to #7911: fix geocaching merge problem with tier and categories


eddiemuc:
retest this please

eddiemuc:
Is Jenkins having hickups again?

moving-bits:
it seems to happen when two builds are running concurrently. One will be ok, the second gets those typical 11 or 12 failing tests

Issue #14612, opened 2023-09-15
custom connector icon broken


ztNFny:
![image](https://github.com/cgeo/cgeo/assets/1258173/690b1e2b-c2c2-4d2f-90b8-c5ffd824b1de)

moving-bits:
`background_other` is non-square, but scaling assumed drawables to be squares. I have a fix available.

ztNFny:
I broke the drawable recently, ill post a new version..but need the pc to do that

moving-bits:
Ok, then I'll push the fix for non-square drawables as non-closing PR, so that you can close this issue with your updated drawable.

Issue #14621, opened 2023-09-16
change how LayerDrawable scaling works


ztNFny:
### Tell us your idea!

Especially cache markers (but also some other places like attributes) make heavy use of LayerDrawables constructed on runtime.
If scaling of such a drawable is required then every single layer has to be scaled independently. This scaling only accepts integer pixel sizes, so if you have a 13px round source image and scale it to 50% the resulting 6 or 7 px image will be off-center.

### What solution would you suggest?

We should have a way to scale the whole drawable, after it's built. This should also allow scaling interim drawables, e.g. to allow building a marker from 3 full-size layers, then scaling the whole marker to its target size and adding it in one of the corners of the map marker.
This would allow for cleaner code and more flexible use of drawables = less need to duplicate drawables with slightly different configs.

So far the only way I found to achive something like this is to convert to BitmapDrawable inbetween which seems kinda ugly. Also I didn't look into performance impact of this.

### Is there an alternative?

Do nothing.

PR #14622, opened 2023-09-16
fix markers broken by14598

ztNFny:
https://github.com/cgeo/cgeo/pull/14598 broke some cache markers when scaling (which happens pretty much always) due to several quirks in the LayerDrawable implementation I happened to forget about :-(
This is a better implementation that changes nothing a normal user should notice, just working around all those quirks by avoiding scaling whenever possible.

fixes https://github.com/cgeo/cgeo/issues/14612


moving-bits:
Not sure if it's related, but small cache type markers seem to be off-centered for emojis (see waypoint at the top):

image

And when looking at the small "offline found log" marker for the cache at the bottom it looks like the small marker is cut-off slightly at the left.

Anything we can do about it?

ztNFny:
> Not sure if it's related, but small cache type markers seem to be off-centered for emojis (see waypoint at the top):

that's related to https://github.com/cgeo/cgeo/issues/14621
The position can only be set in full px which leads to elements being slightly off. That's why all graphics used in the cache markers are full-size (36*36px), to ensure they all scale exactly the same.
As emojis aren't full size they might be off-center - depending on the exact resultion, DPI  of your device and scaling setting in cgeo.

> And when looking at the small "offline found log" marker for the cache at the bottom it looks like the small marker is cut-off slightly at the left.

can't reproduce, it's all there on all my devices
image

moving-bits:
> > Not sure if it's related, but small cache type markers seem to be off-centered for emojis (see waypoint at the top):
> 
> that's related to #14621 The position and size can only be set in full px which leads to elements being slightly off. [...]

ok, thanks for cross-checking

> > And when looking at the small "offline found log" marker for the cache at the bottom it looks like the small marker is cut-off slightly at the left.
> 
> can't reproduce, it's all there on all my devices

ok, maybe it's device-specific and due to specific combination of resolution and screen size
It's just a screenshot I created by using the emulator's function. Emulated device is a Pixel 3 running on API 33 with 1080x2160 resolution.

PR #14625, opened 2023-09-16
Mark '-' strings as non-translatable

moving-bits:
## Description
"-" as a string does not need to be translated, IMHO, therefore marking it as non-translatable

### DIFF
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -104,7 +104,7 @@
     <string name="cache_cat_bc_hiking">Hiking</string>
     <string name="cache_cat_bc_other">Other</string>
 
-    <string name="cache_cat_desc_bc_unknown">-</string>
+    <string name="cache_cat_desc_bc_unknown" translatable="false">-</string>
     <string name="cache_cat_desc_bc_nature">Surrounded by special nature</string>
     <string name="cache_cat_desc_bc_location">Shows you a great location</string>
     <string name="cache_cat_desc_bc_recommended">Recommended by the BetterCacher team</string>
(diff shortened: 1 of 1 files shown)


fm-sys:
All not translate strings should IMHO be moved to the dedicated "strings not translatable" file

moving-bits:
> All not translate strings should IMHO be moved to the dedicated "strings not translatable" file

I wanted to keep their context. But I can move them to the dedicated file (or move them as constants back to their source file instead of loading a const from a resource).

BTW: Why do we keep non-translatable strings in a separate file? Was it a technical restriction back then, or is there a non-technical reason for this?
Crowdin is able to handle `translatable="false"`. And in situations like the one here we break the context of the strings, which makes it harder to see what belongs together. (Currently `strings.xml` is a huge CLOB, IMHO, mostly lacking any recognizable order and naming convention. But that's a different story.)

eddiemuc:
> > All not translate strings should IMHO be moved to the dedicated \"strings not translatable\" file

> I wanted to keep their context. But I can move them to the dedicated file (or move them as constants back to their source file instead of loading a const from a resource).

Please don't move strings to dedicated file just because the are marked as "non-translateable". It is essential that we keep the context, as you wrote. Non-translateable is marked via the flag, not via a dedicated file.

BTW: if something like "-" is used somewhere and shall be non-translateable then IMHO there's no need to put it into Strings anyway. A constant in code would be sufficient

moving-bits:
> BTW: if something like "-" is used somewhere and shall be non-translateable then IMHO there's no need to put it into Strings anyway. A constant in code would be sufficient

In this specific case (Category/Tier) those strings are referenced as resource ID in an enum. I merged the two occurrences, moved it to the non-translatable file, and gave it a generic name. See #14628.

fm-sys:
Maybe crowdin was not able to handle the non-translatable flag back then... 

I got a similar review on one of my early PRs, so I did it here as well. But that doesn't mean that this convention has to stay the way it currently is.

PR #14629, opened 2023-09-17
Add UI for icon scaling options (rel. to #14388)

moving-bits:
## Description
Adds settings' UI for icon scaling options as discussed in https://github.com/cgeo/cgeo/pull/14388#issuecomment-1622076437

image

Two caveats:
- I needed to rename the preference key, as seekbar preferences require an `int` preference, but for some reason I had created the scaling factors as `long` values. This will reset icon scaling factors once on upgrade.
- ~Currently a restart of c:geo is required to bring changed settings into effect.~

~I've tried resetting the marker cache + updating the scaling factor variables in `MapMarkerUtils` on changing the scaling settings' values, but on opening the map the next time within the same session, no markers got drawn and the map crashed eventually, without any useful info in Logcat. Any ideas?~

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/SeekbarPreference.java
+++ b/main/src/main/java/cgeo/geocaching/settings/SeekbarPreference.java
@@ -227,6 +227,7 @@ public class SeekbarPreference extends Preference {
         // get views
         final SeekBar seekBar = (SeekBar) holder.findViewById(R.id.preference_seekbar);
         valueView = (TextView) holder.findViewById(R.id.preference_seekbar_value_view);
+        holder.setDividerAllowedAbove(false);
 
         // init seekbar
         seekBar.setMax(maxProgress);
@@ -280,4 +281,12 @@ public class SeekbarPreference extends Preference {
             });
         });
     }
+
+    /** apply mapping to change notifications */
+    @Override
+    public boolean callChangeListener(final Object newValue) {
+        final OnPreferenceChangeListener opcl = getOnPreferenceChangeListener();
+        return opcl == null || opcl.onPreferenceChange(this, progressToValue((int) newValue));
+    }
+
 }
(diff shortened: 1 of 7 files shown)


moving-bits:
> markers are cached in overlaysCache in MapMarkerUtils, you'd have to clear that cache

thanks. I had tried clearing overlayCache and emojiPaintMap + setting scalingFactorCacheIcons /scalingFactorWpIcons to their new values, but on opening the map the next time, it stayed with the cache loading indicator, did not paint any icons/markers, and eventually crashed (without any c:geo related error in the logs).

ztNFny:
we have the same issue (require restart) for large smileys btw. Never tried making it work without restart

ztNFny:
> I could actually live with the restart - changing the scaling options is probably nothing you would do on a daily basis - but I would like to understand why clearing the overlayCache isn't enough in this case. I guess I must have missed something in my code.

it works for me - I did a quick test on top of my current dev branch. I added the reset call on "offline log" because that's what I was working on and it gets me real-time results
https://github.com/ztNFny/cgeo/commit/e006b3c39d9f6fba6c4f97cc3299a4065a25f186

after logging a cache it's redrawn with its new size:
image

after going back to list and to map again all markers are the new size

moving-bits:
tried this on a different emulator, with the same result. But finally got some logging info:
`OOM allocating Bitmap with dimensions 14250 x 14250`
Image dimensions are way too large.
On assigning the new value manually in `MapMarkerUtils` I need to divide by 100.
Oops.
I'll adapt this PR accordingly.

ztNFny:
public synchronized static void resetCache() {
        overlaysCache.clear();
        emojiPaintMap.clear();
        scalingFactorCacheIcons = Settings.getLong(R.string.pref_mapCacheScale, 100) / 100.0f;
        scalingFactorWpIcons = Settings.getLong(R.string.pref_mapWpScale, 100) / 100.0f;
    }


should probably be an init method to make sure init and reset do the same thing

moving-bits:
updated PR:
- scaling gets applied on opening the map, no restart of c:geo required anymore
- `resetAllCaches()` is called during class init
- found and fixed a value mapping error in `SeekBarPreference` in conjunction with `onChangePreferenceListener`

Issue #14502, opened 2023-08-12
Waypoint Icon of Multi Cache


flat2let:
### Tell us your idea!

If we do several multi-caches in the same area all waypoints have the same icon (puzzle piece). This way it is not possible to differentiate the waypoints for the different caches. 

### What solution would you suggest?

I would like to be able to give the waypoint icons (puzzle pieces)different colors to see which ones belong to a certain cache.

### Is there an alternative?

alternatively it could be different waypoint icons, other than puzzle pieces,

### Additional Information

![waypoints](https://github.com/cgeo/cgeo/assets/85786015/624e4467-977c-46ce-915e-c12159ec998f)

flat2let:
Thanks so much for all the work you do. Geocaching would not be great without c:geo

Lineflyer:
IMHO a valid request also just recently discussed on some social media channels.

How about when changing the cache icon itself would trigger the waypoints to have markers (e.g. in lower right) with icon chosen for the cache itself?

Lineflyer:
Another similar feature request but different use case / realisation (feel free to split into another issue):

When going for a multi cache with multiple given waypoints in close vicinity to each other (which have to be visited in sequence) its sometimes cumbersome to find the correct next waypoint from the bunch on the map.

For this I could imagine a on-demand feature to change waypoint icons to numbers using the sequence provided by the owner (via its prefix). This could either change the icon itself or also add a marker icon with the number to it.

eddiemuc:
How about allowing the user to connect waypoints of a cache with lines in a chooseable color in the order of their appearance in the listing? E.g. like in the image below (unlike this quickly-made-up example, lines should be BEHIND icons imho):

![image](https://github.com/cgeo/cgeo/assets/6909759/ee8ec205-55bf-47d8-8fe8-ebdabb35222d)

flat2let:
I think the suggestion of numbering the waypoints would go well together with my idea. The lines are a good idea as well. 
I would like to be able to either color the puzzle pieces or the background. They could be connected with lines and the numbers could be added as icons. But I am happy either way whatever is the easiest to do. As long as you can differentiate.

moving-bits:
> How about allowing the user to connect waypoints of a cache with lines in a chooseable color in the order of their appearance in the listing? E.g. like in the image below (unlike this quickly-made-up example, lines should be BEHIND icons imho):

One could extend the "individual route" feature for this to allow for
[...]

PR #14632, opened 2023-09-19
[14502] for linear labs use numbers as waypoint marker, for all WPs show type icon as overlay

ztNFny:
implements the remaining items from #14502 per https://github.com/cgeo/cgeo/issues/14502#issuecomment-1676448955:

- for linear ALCs use emoji-numbers instead of puzzle icon. As those only cover 0-9 stage numbers > 10 stage are handled by substracting 10 from the stage count -> stage 10 is displayed as 0
- for all waypoints show the associated cache type as marker in top-left (if no custom cache icon (emoji) is set for the cache)

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
@@ -58,8 +58,8 @@ public class EmojiUtils {
     private static final int CUSTOM_SET_SIZE = CUSTOM_SET_SIZE_PER_OPACITY * OPACITY_VALUES;
 
     // internal consts for plain numbers
-    private static final int NUMBER_START = 0x30;
-    private static final int NUMBER_END = 0x39;
+    public static final int NUMBER_START = 0x30;
+    public static final int NUMBER_END = 0x39;
 
     // Unicode custom glyph area needed/supported by this class
     private static final int CUSTOM_ICONS_START = 0xe000;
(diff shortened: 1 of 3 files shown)


moving-bits:
Thanks for diving into this. Two follow-ups:
- Will the numbering of stages be beginning with 0 or with 1? While using 0 is quite common in IT, it isn't in real life, so I would prefer starting numbering waypoints with 1. (But maybe it's already the case, haven't checked the source.)
- There are emojis for 10+ as well, so in the (relatively infrequent) case of lab adventures with 10 stages, you could use those emojis. (Haven't seen labs with more than 10 stages yet, though.)

ztNFny:
> * Will the numbering of stages be beginning with 0 or with 1?

1-based, only stage 10 is displayed as "wrong" 0. 

> * There are emojis for 10+ as well,

only in the "number with circle"-style though. Using "number" for 1-9, then "number with circle" for 10+ seems kinda ugly and confusing. And I think in general the "number" emojis are nicer, especially as the "number with circle" ones are really small:
image


As there are never (until now) more than 10 stages and 10 stages don't happen so frequently (only mega/giga and sponsored labs) showing 10 as 0 is the better solution here imho as it produces a nicer result for 95% of labs.

moving-bits:
agreed. Also I haven't seen a ten-stage _linear_  ALC yet in real life.

ztNFny:
> Also I haven't seen a ten-stage _linear_ ALC yet in real life.

https://labs.geocaching.com/goto/dd8d0a80-92f8-401a-ae8d-2b644d10b3dd

moving-bits:
Interesting. All ten-stage ALC I've seen so far were non-linear. Linearity may even make sense in this case, whereas otherwise it's mostly just a plain PITA, IMHO :-)

fm-sys:
Maybe just display ➕ for stages >= 10?

ztNFny:
> Maybe just display ➕ for stages >= 10?

imho "last digit" is better than "+":

- if there actually are more than 10 stages then 11=1, 12=2, ... which still guides the user through the order as opposed to +,+,+,..
- for a 10 stage lab both "0" and "+" will need to use to think for a moment what it means. Then he'll probably check the waypoints list, see that the icon means stage 10 and accept that. So equal in functional regard. But 0 looks more like the other icons, so why introduce a break?

Issue #14582, opened 2023-09-07
OC Found Date Not Rendered


baiti:
### Describe your problem!

For GC caches, we show whether cache was found by the user and in brackets, the date it was found:

`Status: Found (8/31/23)`

For OC caches we just say

`Status: Found`

I have done some investigations and I believe I know why that is: It is because the OKAPI service we use to fetch the cache object gives a boolean `is_found` but the `date_found` is not provided.

I have worked a solution for this but I am not (yet) deep enough in the code to fix it myself, however, I plan to do it so I am assigning it to myself unless somebody else wants to jump right at it.

### How to reproduce?

Just open load an OC cache you already found.

### Reproducible

Yes

### c:geo Version

2023.08.24

moving-bits:
Where do we get the OC's found state from? Don't we get the found date from there as well? If so, we need to persist it in our database, if we want to display it without having to retrieve it every time we display cache's details.

Found data is stored by setting the `found` flag to `1` and storing date/time in the `visiteddate` column, so one might investigate further, why this is not filled for OC-based caches (value seems to be `0` always for those).

eddiemuc:
> Where do we get the OC's found state from? Don't we get the found date from there as well? If so, we need to persist it in our database, if we want to display it without having to retrieve it every time we display cache's details.

As @baiti already wrote in the issuea initial description (and U double-checked this): the OKAPI does indeed only return an isFound-Flat but not a visited-date.

I assume for GC.com we either scrape the date from the web page or we use the "friends logs" service (I haven't checked implementation). Both options are not available for oc afaik

moving-bits:
Ok, then a different approach could be to make an extra call to the API if found state is returned as `true` and query `user_logs_only` (https://www.opencaching.de/okapi/services/caches/geocache.html), then check for the date. (Do this on updating full cache only.) Requires a [Level-3 authentification](https://www.opencaching.de/okapi/introduction.html), though, - not sure what we are doing in our OKAPI-based connectors.
This should resolve both the "not all logs downloaded" and "don't know the username to compare against" issues.

baiti:
[...]

PR #14634, opened 2023-09-19
Issue 14582 (visitedDate not shown on UI)

baiti:
See description of the issue for more details

## Description

Extract visitedDate from the logs of the current user if the logType is either 2 (found it) or 10 (attended)

## Related issues

#14582

## Additional context

I am not sure  whether `getUserName()` gets me the uername from Opencaching.de or Geocaching.com. In my tests it worked, which may be due to tha fact that I am using the same username on OC and GC. Could somebody check if I am using the right approach for getting the username?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -55,7 +55,6 @@ import cgeo.geocaching.utils.JsonUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.SynchronizedDateFormat;
 import static cgeo.geocaching.connector.capability.ILogin.UNKNOWN_FINDS;
-import static cgeo.geocaching.settings.Settings.getUserName;
 
 import android.annotation.SuppressLint;
 import android.net.Uri;
@@ -198,6 +197,7 @@ final class OkapiClient {
     private static final String SERVICE_CACHE_ADDITIONAL_CURRENT_FIELDS = "gc_code|attribution_note|willattends|short_description";
     private static final String SERVICE_CACHE_ADDITIONAL_L3_FIELDS = "my_notes";
     private static final String SERVICE_CACHE_ADDITIONAL_CURRENT_L3_FIELDS = "";
+    private static final String SERVICE_CACHE_FOUND_DATE_FIELDS = "code|name|is_found|latest_logs";
 
     private static final String METHOD_SEARCH_ALL = "services/caches/search/all";
     private static final String METHOD_SEARCH_BBOX = "services/caches/search/bbox";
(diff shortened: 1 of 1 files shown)


fm-sys:
`Settings.getUserName()` can't know which connector you are requesting so it's most likely GC. 

OT:
IMHO we should deprecate the getUserName() method totally as it doesn't fit to the connector based concept of cgeo.

baiti:
Slightly OT ... but still somewhat related: I spent some time looking at the code and it seems we don't have a concept of managing a `userid` based on the connector. `Settings.getUserName()` seems to give me the GC user. What if somebody wants to exlusively use OC and not GC? What if someone maintains different usernames on GC and OC?

I think this is a bigger problem than just the "visitedDate" and further, I do think it should be resolved one way or another.

In the OC case: there is an OKAPI service called "userinfo". This returns the username, among other things. But it doesn't make any sense to call that service each and every time I want to parse a geocache's log for the `visitedDate`. It must be possible, to store that variable somewhere I can later access it without running a costly OKAPI method.

I have tried to find such a way but my Java skills are way way way not sufficient to accomplish such a task. I repeatedly ran into  compile errors because things were "static" where they shouldn't be static and the othe way around. The error messages I got just confused me more, the more I tried.

moving-bits:
> Slightly OT ... but still somewhat related:

I've created a separate issue for this for further discussion.

main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
```diff
@@ -809,9 +810,19 @@ private static Geocache parseCache(final ObjectNode response) {
             cache.setLogPasswordRequired(response.get(CACHE_REQ_PASSWORD).asBoolean());
 
             cache.setDetailedUpdatedNow();
+
+            final List<LogEntry> logs = parseLogs((ArrayNode) response.path(CACHE_LATEST_LOGS), cache.getGeocode());
+            final String me = getUserName();
```
moving-bits, line 815: This won't work in cases where OKAPI username is different from GC username (or GC isn't used at all).

eddiemuc:
Additional to the username discussion, the solution will only work if the current user's found log is among the latest logs on the cache. This is because we only retrieve the last x logs for a cache of course. (Don't know how many, I assume something around 40 or 50?)

moving-bits:
Instead of using this workaround for guessing a user's found log we may stick to the approach used by c:geo for presenting a "found date" (although that may have drawbacks as well, but that's a different story), and investigate why that mechanism does not get triggered for OC-based caches. See https://github.com/cgeo/cgeo/issues/14582#issuecomment-1732253748

baiti:
> Additional to the username discussion, the solution will only work if the current user's found log is among the latest logs on the cache. This is because we only retrieve the last x logs for a cache of course. (Don't know how many, I assume something around 40 or 50?)

As far as I remember the "fields" say to get them all. Otherwise I wouldn't have coded it that way.
PS: I was a bit surprised about the decision to ge them all but the rationale most likely was that there aren't many logs on any of thise OC caches anyway.

baiti:
I gave it another try. It worked on my emulater. Perhaps the

`params.add("user_logs_only", "true");`

should be replaced with some sort of a constant? Ideas how it should be named then?

main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
baiti: pls re-review

fm-sys:
Just came to my mind again: We already store the user name for the FoundNumCounter on a per-connector base...

moving-bits (approved):
LGTM, thanks for adapting the PR

PR #14635, opened 2023-09-19
New GCLog API: Basic Cache logging + images + trackable status

eddiemuc:
First implementation for new GC.com Log API.

To activate usage of new API by c:geo, a new hidden boolean setting `feature_new_gc_log_api` must be added and set to value `true`. 

With this activated the following should work via new logging API:
* Simple logging of geocaches, e.g. log date, type, text
* Adding of images to a log (including name + description)
* Automatic trackable status setting to "visited"/"dropped" according to setting in c:geo

The following will NOT work for the new setting:
* "add to favorites" setting in log will be ignored (-> as of now it is not clear whether this can be handled by new API)
* Logging of Trackables seems to be using the new API too on website, but was not yet touched in c:geo

The implementation is to be considered a first try. Alpha testing is heavily needed, preferably with Log leve set to debug for easy analysis in case of problems.

Reference to internal discussion: https://github.com/orgs/cgeo/discussions/77

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
@@ -191,6 +191,9 @@ public class GCConnector extends AbstractConnector implements ISearchByGeocode, 
     @NonNull
     @Override
     public String getCacheCreateNewLogUrl(@NonNull final Geocache cache) {
+        if (Settings.enableFeatureNewGCLogApi()) {
+            return GC_BASE_URL + "live/geocache/" + cache.getGeocode() + "/log";
+        }
         return GC_BASE_URL + "play/geocache/" + cache.getGeocode() + "/log";
     }
 
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
@@ -68,6 +68,7 @@ public final class GCConstants {
     static final Pattern PATTERN_FAVORITECOUNT = Pattern.compile("<span class=\"favorite-value\">\\D*([0-9]+?)\\D*</span>");
     static final Pattern PATTERN_COUNTLOGS = Pattern.compile("<span id=\"ctl00_ContentBody_lblFindCounts\"><ul(.+?)</ul></span>");
     static final Pattern PATTERN_WATCHLIST_COUNT = Pattern.compile("data-watchcount=\"(\\d+)\"");
+    static final Pattern PATTERN_CSRF_TOKEN = Pattern.compile("\"csrfToken\":\"([^\"]+)\"");
 
     /**
      * Two groups !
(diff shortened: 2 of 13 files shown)


bekuno:
This should imho target to the `release` branch, so that we can fast include it in a new release.

moving-bits:
> > This should imho target to the `release` branch, so that we can fast include it in a new release.
> 
> This change is (or will be) too big to go directly to release/beta. It needs to be tested throughly 8n nightlies first.

It could be merged to `release` (and therefore be tested in the nightlies for a couple of days) and we can create a beta version then.
But this should happen only after current beta is released (which will probably be done end of this week).

Alternatively you could base it on `release`, but target it to `master` first. Then we would have the chance to cherry-pick to `release` if required.

eddiemuc:
Side note to this topic: I am very restricted timewise in the next couple of weeks. If the API topic is time-critical then please don't wait up for me to proceed here and take over. My PR may or may not serve as a base for other implementations, but it is marked as "Draft" for a reason :-).

eddiemuc:
I edited title and description of this PR to reflect its current state. It is IMHO ready for merge into master,. Next step has to be alpha testing of what has been implemented. Not everything works for now (see description).

In this state I really don't want to see this code merged to `release`. If someone feels otherwise, feel free to rebase and direct it towards release.

eddiemuc:
I also noted that there is no official issue for this topic as of now

fm-sys:
"Add to Favorite" would be nice... Maybe as a temporarily workaround, call the dedicated fav API after logging until we find out how to do it better?

moving-bits:
Thanks for diving into it! Let's give it a try in the nightlies.
Any findings should be reported in the linked discussion for now.

eddiemuc:
> "Add to Favorite" would be nice... Maybe as a temporarily workaround, call the dedicated fav API after logging until we find out how to do it better? 

Would be the next step, I agree.

The current architecture of this is a bit strange anyway, because the pattern of using the "normal" addtofavorite routine behind the curtain (and not rely on platform-specific features whilst logging) could be used in general for all connectors. I wonder why it was decided in the past to make such routines dependent on connector-log-capabilities anyway?

fm-sys:
> I wonder why it was decided in the past to make such routines dependent on connector-log-capabilities anyway?

There is a simple reason for it: Geocachers leave fav points to acknowledge the work of the owner and to say "thank you!". The owner receives an email with the log and the hint similar to "this user gave a favorite point".

When using the "normal" addtofavorite routine owners don't get this notification, which takes away one of the main reasons to use this feature at all.

eddiemuc:
> > I wonder why it was decided in the past to make such routines dependent on connector-log-capabilities anyway?
> 
> There is a simple reason for it: Geocachers leave fav points to acknowledge the work of the owner and to say "thank you!". The owner receives an email with the log and the hint similar to "this user gave a favorite point".
> 
> When using the "normal" addtofavorite routine owners don't get this notification, which takes away one of the main reasons to use this feature at all.

That actually makes a lot of sense! Thanks for the enlightment.

PR #14639, opened 2023-09-20
on sending or storing a log remove the cache from current route

### LINKED ISSUE #13089: remove from route on (offline) log
### Tell us your idea!

When a cache is on the active route remove it when saving/sending alog. Should be an option

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

ztNFny:
Add a (disabled by default) option to remove caches from current route on saving an offline log or sending a log. Implemented for all 3 maps.

As the IndividualRoute is not a singleton and the log dialog is "very far away" from the maps I didn't see any way to notify the maps about the external change of the route, so the only option to implement this was to reload the IndividualRoute on every resume.

fixes #13089


moving-bits:
For changing caches we use some broadcasting mechanism, see `GeocacheChangedBroadcastReceiver`, maybe something similar could be used for this use case. Or maybe switch `IndividualRoute` to a singleton?

Removing an entry should be done on database level (create static helper in `DataStore.java`), then reloading the route, instead of loading, removing in memory, storing and reloading. (Would be easier if `IndividualRoute` would be a singleton.)

Edge case: Any item may be part of an individual route multiple times. On creating the log we should probably remove the first occurrence only, as the next occurrences (if existent) might still be relevant for routing (otherwise they wouldn't have been included multiple times in the first place).

eddiemuc:
@fm-sys introduced a nice architectural pattern for unified map to do model changes in a central place. Could this be reused here?

(In any case In wouldn't bother to try any technical refactorings related to the "old" map architecture but concentrate such efforts solely on the new unified map path)

ztNFny:
added a static helper method in DataStore

A broadcast mechanism would be great, but I agree with @eddiemuc it's probably only worth doing that for new map. However I very likely won't be able to look into this "optimization" in the next weeks.

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -3560,6 +3560,21 @@ public static void saveIndividualRoute(final List<RouteItem> routeItems) {
         });
     }
 
+    public static void removeFirstMatchingIdFromIndividualRoute(final String id) {
+        withAccessLock(() -> {
+            init();
+            database.beginTransaction();
+            try {
+                database.delete(dbTableRoute, "precedence = (SELECT precedence FROM " + dbTableRoute + " WHERE id = ? OR id LIKE ? ORDER BY precedence LIMIT 1)", new String[] { id, id + "-%" });
```
moving-bits, line 3568: On removing a cache from an IndividualRoute this would also remove all waypoints of this cache from the IndividualRoute (resp. the first one of them), as their generated id start with the same prefix (geocode). To get the identifier for the element to be removed have a look at `RouteItem.buildIdentifier()`

main/src/main/java/cgeo/geocaching/models/IndividualRoute.java
```diff
@@ -29,6 +29,10 @@ private enum ToggleItemState {
     private boolean loadingRoute = false;
     private SetTarget setTarget = null;
 
+    public IndividualRoute() {
+        super(true);
```
moving-bits, line 33: still needed? (wasn't there before)

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -850,6 +850,9 @@ public void onPause() {
 
     @Override
     protected void onResume() {
+        if (Settings.removeFromRouteOnLog()) {
+            viewModel.reloadIndividualRoute();
```
moving-bits, line 854: how would this get triggered? (Where does the actual removal of RouteItem happen?)

Due to using the viewModel architecture you don't need an actual reload from database in `onResume`, you can simply remove the selected item from the `IndividualRoute` in memory, that should trigger the update on the map automatically. (But you still need to persist the removal, so calling `removeFirstMatchingIdFromIndividualRoute()` is still required.)

moving-bits (changes requested):

moving-bits (approved):
Haven't tested, but from looking at the code LGTM.

PR #14641, opened 2023-09-21
Show permission update info on API33+ w/o notification permission

moving-bits:
## Description
Shows a OneTimeDialog informing about required POST_NOTIFICATIONS permission and optionally starts configuration wizard.
Happens on API33+ systems only, and only if notification permission is missing.


moving-bits:
Let's test this in the nightlies

fm-sys:
Why do we need a OneTimeDialog in between...? 🤔

moving-bits:
> Why do we need a OneTimeDialog in between...? 🤔

I've chosen this way for two reasons:
- c:geo should not run into installation wizard on a simple update, but should explain first what has changed, and it should give the user the choice whether to run through it now or later
- and at the same time the OTD stores the info whether we have asked the user already or not

PR #14645, opened 2023-09-23
Remove AnrdoidBeam (fix #14644)

### LINKED ISSUE #14644: Remove AndroidBeam (?)
AndroidBeam seems to be removed from Android starting with API 34 (see [this article](https://www.xda-developers.com/android-beam-permanent-removal-android-14/) and findings in #14619), though I was not able to find some "official" confirmation.

Maybe it's being removed only partially, but c:geo source will no longer compile when setting `targetSDK` to 34, as method `nfcAdapter.setNdefPushMessageCallback` is no longer available then. I do not know of a viable replacement method.

Inferring from current requirement cycles by Google, setting `targetSDK` to 34 will be a prerequisite for uploading any further updates of c:geo to Play Store starting summer of 2024, so I guess we will need to remove support for AndroidBeam completely from c:geo until then (except we find some workaround or the missing method will be restored).

Supporting nearby share, as proposed in #5097, might be an alternative.

moving-bits:
## Description
Removes discontinued AndroidBeam from c:geo


moving-bits:
Note to myself: can `android.permission.NFC` from `AndroidManifest` be removed as well?

moving-bits:
as discussed in today's dev meeting: let's merge to `master`

PR #14652, opened 2023-09-24
targetSdk 33 for subprojects

bekuno:
Set targetSdk 33 for all other projects to stay in line with `main`.

Follow up changes should not be necessary, imho.

### DIFF
--- a/cgeo-contacts/build.gradle
+++ b/cgeo-contacts/build.gradle
@@ -22,7 +22,7 @@ android {
     defaultConfig {
         minSdk 21
         //noinspection OldTargetApi
-        targetSdk 31
+        targetSdk 33
 
         // include only those language resources from libraries which we actively maintain ourselves in the translation project
         // not yet enough translations (~50%): "is","iw" (="he"),"zh"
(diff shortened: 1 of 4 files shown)


bekuno:
Strange result in CI, especially since the `main` part was not changed:
~~~
22:18:56  > Task :main:lintAnalyzeBasicDebug
22:18:56  e: /home/jenkins/.gradle/caches/transforms-3/95927e43ecba0de0d69c1e13e5561e0d/transformed/viewpager2-1.1.0-beta02/jars/classes.jar!/META-INF/viewpager2_release.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected 
~~~

retest this please

bekuno:
And additionally - why does it work without an `implementation` line?

Will try with it.

bekuno:
> And additionally - why does it work without an `implementation` line?
> 
> Will try with it.

No success.

In the meantime, this error is also raised on other CI builds.
So it is maybe a gradle cache problem.
I added `./gradlew cleanBuildCache` in the PR build.
Let´s test it.

retest this please

bekuno:
retest this please

bekuno:
The task `cleanBuildCache` is not more a gradle task.
And the task `clean` is already called when starting the CI PR job.
No success so far.

bekuno:
Its fine now.

Issue #14602, opened 2023-09-12
Adding a filter-possibility into the list selection


murggel:
### Tell us your idea!

Adding a better filter into the list selection might also help...

_Originally posted by @wolverine007 in https://github.com/cgeo/cgeo/issues/14463#issuecomment-1672793299_

Having a bunch of lists, it is not easy to select the correct ones. 

### What solution would you suggest?

Having a "search/filter"-option / field in the select-dialog, the stored lists can be limited to the desired ones.

Something like this (but also for multiselection...)

![59cqH](https://github.com/cgeo/cgeo/assets/70113341/15bd4891-8919-4c84-ab18-542327953e88)

ztNFny:
like https://github.com/abumoallim/Android-Multi-Select-Dialog ?

bekuno:
> like https://github.com/abumoallim/Android-Multi-Select-Dialog ?

No development since 5 years :-(

ztNFny:
The screenshots in OP are waaay older ;-) My question was whether this is what @murggel is asking for with this request.

Also - some stuff is just finished, no further dev needed. The implementation looks very trivial, could use some updates to the latest libs, but codewise there isn't much.

murggel:
> The screenshots in OP are waaay older ;-) My question was whether this is what @murggel is asking for with this request.
> 
> Also - some stuff is just finished, no further dev needed. The implementation looks very trivial, could use some updates to the latest libs, but codewise there isn't much.

The  screenshot was just a quick example for the functionality, of course not for look&feel...

murggel:
And of course should this be possible for the option of single select a list ;-)

moving-bits:
We should not introduce yet another software dependency for something trivial like a list filter. (But I guess linking to the project was just meant to show an example, not necessarily to include this library.)

If we add a search filter to our list selector, it should be added as a feature to our default dialog components (if feasible), IMHO, so that it adheres to c:geo look&feel, and so that we can reuse it.

moving-bits:
For multi-select lists see PR #14675.

For single select lists (specifically the grouped ones) I have additional questions:
- How to handle group headings?
  - Should filtering also apply to group headings? (If so, then see follow-up question below)
  - Or shall all group headings always be visible?
  - Or just those which have matching entries?
  - Or shall group heading become invisible when filtering?
- If a group heading matches to the filter string, shall all entries of this group be displayed?

[...]

PR #14653, opened 2023-09-25
Replace 'select list' dialogs by SimpleDialog (rel. to 14602)

moving-bits:
## Description
As a first step towards #14602 replace self-built list selection dialogs with calls to `SimpleDialog`


moving-bits:
@eddiemuc 
Can you cross-check, please, that I haven't broken something relevant in `SimpleDialog`? Thanks.

eddiemuc:
A bit hard to inspect on a mobile, but seems ok to me

moving-bits:
> A bit hard to inspect on a mobile, but seems ok to me

thanks for cross-checking. Let's test it in the nightlies

PR #14663, opened 2023-09-29
First draft for filtering on UnifiedMap

moving-bits:
## Description
Started another try on getting filtering for `UnifiedMap` implemented. Filter bar get displayed now, and filter pops up on hitting the "filter" icon, but otherwise I'm still stuck:

- If I'm not mistaken, all geo items are stored in `viewmodel.geoItems` (caches, waypoints etc.). Is this similar to `caches` in `CGeoMap` (`LeastRecentlyUsedSet<Geocache>`) / `NewMap` (`CachesBundle`)?
- how to apply filtering to `geoItems`? 
- and how to update map after having updated `geoItems`?
- how to restore the original list of items (eg: `SearchResult`) when a more restrictive filter settings gets unchecked?

@eddiemuc 
Can you shed some light on it?


eddiemuc:
I try to take a look at it during weekend

eddiemuc:
It looks to me as if you set up the GeocacheFilterContext correctly already, including saving/restoring it within the activity lifecycle.  This context object, if set up correctly, gives you access to the "right" filter to apply depending of the context the map was opened (e.g. LIVE context when map is in live mode vs OFFLINE context when map is opened from a stored list).

What I can't find in the code currently is that the filter is actually executed. To execute the filter on a bunch of geocaches (or waypoints), there are two fhelper methods in class `MapUtils` (one for caches, one for waypoints).

The old map implementation did not share a viewmodel state with others. Basically it has two separate "cache loaders" running in parallel: one for online caches (via connector) and one for stored caches (via DataStore). Both the connectors and the DataStore also understand filtering and will pre-filter their result as best they can (connector via whatever the API provides, DataStore via SQL). The result however is then post-filtered again using the `MapUtils` methods mentioned above.

With the shared viewmodel, the pattern might have to be changed a bit:
* Waypoints and Geocache objects seems to be stored in the same list (`viewmodel.geoItems`). Filters can however only be applied to geocaches or Collections of geocaches. If waypoints and geocaches have to remain in same list for other reasons, then we need to extend the filter methods to be able to handle both in same collection. This is possible, I can offer to provide a code change for this. Should I do that?
* Filtering changes the content of the filtered collection! This is not a problem in the old implementation because those are "throw-away-collections" anyway (they are then used for display and not shared with others. Is this a problem with the shared viewmodel? @fm-sys ?

eddiemuc:
Overall: is there an actual need to share the current list of displayed caches/waypoints in the shared viewmodel? WHo else is using this list, and for what purpose?

fm-sys:
There isn't any reason why geocaches are stored like this in unified map currently. Feel free to throw it away or refactore it in a way it makes sense. 

As of now, the geocaches are part of the viewmodel to be able to render them + decouple loading and display code base. That being said, the current implementation is probably pretty inefficient and at least needs some performance optimization.

eddiemuc:
> ... the current implementation is probably pretty inefficient and at least needs some performance optimization.

Current waypoint filtering mechanism is inefficient and should be refactored. That can be done rather locally.

(Re)loading caches from online connectors / database on filter change is necessary IMHO. Especially if filter is "widened" (means: may return more results as before). It is far from trivial to tell from a filter change whether this could be the case - would be a rather complex optimization with limited value. I would try to optimize other things before that.

Reason for online rerun  is that state may have changed there also by other means (e.g. a user added a new cache/log/... meanwhile). 

Database access could in general greatly be improved by using a "real" DB abstraction layer (Android Room or such) which probably includes caching across more complex database queries as well. But that's a (big!) topic on its own I suppose.

fm-sys:
Just for clarification: I actually didn't had the big picture in mind while writing my short comment. Just wanted to say that there is no need to stick with what's currently there at unified map...

moving-bits:
Thanks @eddiemuc @fm-sys for your comments so far. 
I've made some progress meanwhile, activating `MapUtils.filter()` in `LoadInBackgroundHandler`, so it *should* (and probably does) filter the loaded caches/waypoints, but it does not get reflected on the map. I guess `UnifiedMapActivity.refreshMapData()` needs to trigger an update of the drawn objects. But how to do this? @fm-sys?

fm-sys:
Hi, I had originally promissed to take a look at the veiwmoddel / how to display the filtered results correctly. However it seems more complicated than I would have expected it. 

> GeoItemLayer unfortunately does not yet handle updates of whole groups of objects. Sorry for wrong information. Could be build in however

@eddiemuc do you think it'll make sense to do this? OTOH it doesn't sound very efficient to render bitmaps for _all_ elements everytime just so that GeoItemLayer ignores already existing elements again. 
IMHO the diff comparison should ideally be done somewhere inside the viewmoddel, not GeoItemLayer. I've already started with this back then, but doesn't really seem to work yet. 

What do you think is the way to go?

fm-sys:
@moving-bits finally I found time to work on this as promised. Filtering caches as well as changing filters and reload does work now. 

I've used the logic from `CgeoMap` (using a `LeastRecentlyUsedSet` inside viewmodel) and compare the new cache set with the old map refresh to check which caches are new/changed/removed inside `GeoItemsLayer`.

There are still some to-dos, e.g. update the filter bar if necessary, some cleanup and maybe some more things. Do you want to take over from here?

moving-bits:
Hi @fm-sys, thanks for the update, great news! I'll try to look into it more in-depth over the next days and will gladly take over from here.

moving-bits:
@fm-sys 
Found a small glitch, maybe you have a quick idea at hand how to fix it:
- if I'm on the live map and add a filter to exclude cache type "traditional", they get removed from the map => ok
- if I open the filter again, include "traditional" again (and maybe remove "mysteries"), the traditional caches won't be added to the map until I move the map, zoom out/in etc., whereas "mysteries" will be removed immediately

Additionally, I found that opening a filtered map, caches won't be drawn until I move and/or zoom the map, even when the loading indicator signals loading has finished. Long-tap (for opening a popup) seems to work as well.

How can I trigger updating the map automatically in this case?

(All observations in emulator only, having started UnifiedMap from home screen icon.)

fm-sys:
I thought I had already fixed this glitch by using

```
if (loadInBackgroundHandler != null) {
    loadInBackgroundHandler.onDestroy();
}
loadInBackgroundHandler = new LoadInBackgroundHandler(this);

```

LoadInBackgroundHandler remembers if caches in this area are already loaded, so we must recreate it or introduce a method resetting the cached viewpoint (the second is probably the cleaner approach)

moving-bits:
I've digged a bit deeper into this:
- `viewModel.caches.observe()` in `GeoItemsLayer` gets called both on removing and adding a cache type in the filter
- `caches` variable seems to have reasonable sizes (and have the same size before removing a certain cache type and after adding the same type later on)
- `GeoItemLayer.put()` gets called for the missing elements on re-adding the cache type
- BUT: those items only reappear on the map on moving the map / zooming / long-tapping

@eddiemuc 
How can I trigger a map update when all "puts" to the `GeoItemLayer` are done?

eddiemuc:
> @eddiemuc How can I trigger a map update when all "puts" to the `GeoItemLayer` are done?

This should not be necessary, update of the map should happen on a per-put-basis (athough asynchronously in a background thread). Maybe there's a bug in the geoitemlayer, I would have to look into this. Are there any errors/exceptions in the log?

eddiemuc:
Edit: if a map implenentationrequires an explicit "update" to draw a new object, thenthis update has to be triggered in the map-specific Layer implementations, e.g. MapsforgeV6GeoItemLayer, MapsforgeVtmGeoItemLayer and/orGoogleV2GeoItemLayer, and there in the add method.

For which map type does this happen? MapsforgeV6, VTM or Google? Or all of them?

moving-bits:
> For which map type does this happen? MapsforgeV6, VTM or Google? Or all of them?

I'm currently into getting filtering in VTM to work, haven't tried the other maps.

Updating does happen, but only if I move the map around, zoom in or out, or long-tap somewhere on the map.

Same for initial loading of the map. This seems to be a problem for current release 2023.09.26 already on opening UnifiedMap.

fm-sys:
> > For which map type does this happen? MapsforgeV6, VTM or Google? Or all of them?
> 
> I'm currently into getting filtering in VTM to work, haven't tried the other maps.
> 

I think to remember that it's working properly on google-unifiedmap. That's the map implementation I used while developing the new cache update mechanism for unified map. Can you verify it is working there?

moving-bits:
You are right:
- UnifiedMap Google does work
- UnifiedMap OSM Online does not work
- UnifiedMap OSM Offline does not work

Test case is:
- open UnifiedMap without a filter
- set filter to exclude cache type traditional => traditional caches get removed
- set filter to include all cache types => map unchanged, traditional caches missing
- tap on map, move map or zoom => traditional caches appear

fm-sys:
=> there must be a bug in `MapsforgeVtmGeoItemLayer` 

(I'll also start debugging in parallel)

eddiemuc:
> => there must be a bug in `MapsforgeVtmGeoItemLayer` 
> 
> (I'll also start debugging in parallel) 

I probably just forgot to put in a map update command.

fm-sys:
> > => there must be a bug in `MapsforgeVtmGeoItemLayer`
> > (I'll also start debugging in parallel)
> 
> I probably just forgot to put in a map update command.

The update command is there...

grafik

...but interestingly, no matter what I tried I couldn't reproduce the issue. It takes some short time until the caches are loaded, but then the caches are rendered correctly for me without the need to move the map. (Android 14 Emulator)


[Screen_recording_20231022_005325.webm](https://github.com/cgeo/cgeo/assets/64581222/0691ad7a-5835-408f-be7c-5dc80d115df6)

moving-bits:
> ...but interestingly, no matter what I tried I couldn't reproduce the issue. 

That's fascinating: I fired up a different emulator (Pixel 3a API 30), and here indeed it does work, whereas my Pixel 3a API 33 is failing reproducible. So maybe it's specific to that emulator. Let's put this issue aside for the moment, but keep an eye on it if it fails again in a different setting. Thanks to all who investigated further into this!

I guess we can then merge this PR now and continue with further tasks that need this base (quick settings menu, displaying & filtering waypoints etc.), removing Draft status therefore.

@fm-sys 
BTW: How do you create your screen recordings?

fm-sys:
> BTW: How do you create your screen recordings?

I have a button for it in the toolbar, or if it should not be there this feature can be accessed via the extended controls
|||
-|-
grafik|grafik

fm-sys:
Just noticed another glitch, that filter bar will not update when using the "quick filter change"

fm-sys:
> I guess we can then merge this PR now and continue with further tasks that need this base (quick settings menu, displaying & filtering waypoints etc.)

+1 for merging

Which of those topics you will work on first? I might have some spare time today, so I would take a look at one of the other things...

moving-bits:
> Which of those topics you will work on first? I might have some spare time today, so I would take a look at one of the other things...

I'm a bit time-constrained today, so you may choose what to start with, if time permits.
Just found one more thing missing: Identifying clicked items, see `UnifiedMapActivity.onTouch()`. Needs some glue code between `GeoItemLayer` and the new `UnifiedMapViewModel.caches` - maybe that's a good starting point?

fm-sys:
> That's fascinating: I fired up a different emulator (Pixel 3a API 30), and here indeed it does work, whereas my Pixel 3a API 33 is failing reproducible. So maybe it's specific to that emulator.

It's getting even more complicated. Today, the same emulator which was working yesterday shows this weird behavior without changing anything compared to yesterday. We should definitely debug this, although I don't know how to start.

PR #14667, opened 2023-09-30
GeocacheFilter configurations: change format to json

eddiemuc:
GeocacheFilter configurations: change format to json


eddiemuc:
This PR will use json configurations for filters, but can still deal with old config format. When merged and no problems occur, next step is to migrate stored filters to json format. In a third step, legacy code can be removed

PR #14675, opened 2023-10-01
Add filtering to SimpleDialog.selectMultiple (rel. to 14602)

moving-bits:
## Description
Adds a filter box to `SimpleDialog.selectMultiple()`:

image.image

- Filter will only be displayed when there are 10 entries or more initially.
- Entry for "Select all" reflects the number of visible elements and only gets applied to visible elements.
- Entry for "Select all" will not be filtered out.
- Filtered-out but still checked elements are part of the result.
- Uses a custom layout for the `AlertDialog` to be able to inject the filter box. Therefore requires a custom adapter as well (as it's already done in `SimpleDialog.selectSingle()`).
- Checkbox state of "Select all" can become invalid currently when used in combination with filtering.
- Tested only with "Select lists for cache" for now.

Setting draft state for now due to the last two items.

@eddiemuc 
Can you cross-check if I haven't broken some of the use-cases for `SimpleDialog.selectMultiple()`? Thanks.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
@@ -7,6 +7,7 @@ import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.TextUtils;
+import cgeo.geocaching.utils.functions.Action1;
 import cgeo.geocaching.utils.functions.Action2;
 import cgeo.geocaching.utils.functions.Func1;
 import cgeo.geocaching.utils.functions.Func2;
@@ -22,6 +23,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
+import android.widget.CheckedTextView;
 import android.widget.EditText;
 import android.widget.ListAdapter;
 import android.widget.ListView;
@@ -408,10 +410,7 @@ public class SimpleDialog {
         final CharSequence[] itemTexts = new CharSequence[addSelectAll ? items.size() + 1 : items.size()];
         final boolean[] itemSelects = new boolean[addSelectAll ? items.size() + 1 : items.size()];
         final Set<T> result = new HashSet<>();
-        if (addSelectAll) {
-            itemTexts[0] = TextParam.text("<" + LocalizationUtils.getString(R.string.chipchoicegroup_selectall) + " (" + items.size() + ")>")
-                    .setMarkdown(true).getText(null);
-        }
+        final ArrayList<Integer> filteredPositions = new ArrayList<>(itemTexts.length);
         int idx = offset;
         for (T item : items) {
             final TextParam tp = displayMapper.call(item, idx);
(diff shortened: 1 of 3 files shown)


main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -474,26 +463,83 @@ public final <T> void selectMultiple(@NonNull final List<T> items, @NonNull fina
...
+                    ctv.setChecked(newStateIsChecked);
+                    itemSelects[originalPos] = ctv.isChecked();
+                    checkOkButton(dialog, nonEmptyResultRequired, result.size());
+                });
+                return v;
+            }
+        };
+        lv.setAdapter(adapter);
+
+        ((EditText) dv.findViewById(R.id.filter)).addTextChangedListener(new TextWatcher() {
```
eddiemuc, line 508: Consider using `ViewUtils.createSimpleWatcher()` here

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -474,26 +463,83 @@ public final <T> void selectMultiple(@NonNull final List<T> items, @NonNull fina
 
         final AlertDialog dialog = builder.create();
         dialog.show();
-        checkOkButton(dialog, nonEmptyResultRequired, result.size());
+
```
eddiemuc, line 466: The way you added the field it will be:
* displayed in ALL multiple choice dialogs. I would consider making this dynamic (e.g. only show if there are more than 5 entries)
* not be displayable in single-choice dialogs. There might be such dialogs with many choices where such a filter also makes sense (e.g. filter selection in advanced filter dialog)

eddiemuc (commented):
See comments. Thanks for diving into this

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
moving-bits: Thanks for the pointer. I'll extend this to support `onTextChanged()` as well.

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
moving-bits: Thanks for reviewing the PR.

> The way you added the field it will be:
> * displayed in ALL multiple choice dialogs. I would consider making this dynamic (e.g. only show if there are more than 5 entries)

Good hint, I'll add a check.

> * not be displayable in single-choice dialogs. There might be such dialogs with many choices where such a filter also makes sense (e.g. filter selection in advanced filter dialog)

Done intentionally for now. I wanted to make it work for multi selections first and get some feedback. Single selection dialogs have some different inner workings, will work on adaption for it next.

moving-bits:
retest this please

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: I strongl Believe your use case would also work with current implementation (w/o the need to extend to context changed). I use it for example to react on any typing change on formula text field on variables page

moving-bits: ah, didn't know that, had only worked with `onTextChanged()` before. I've adapted the PR to work with `afterTextChanged()`.

Issue #14685, opened 2023-10-05
Support "new" maintenance log types


Lineflyer:
### Tell us your idea!

The GC website will soon rework the maintenance log types.

Seemingly the possibility to write a found log with additionally selecting "Report problem" will be dropped and they return to, what was implemented some years ago (i.e. dedicated log types for NM and NA).

The only difference is the naming of the NM and NA log types.

### What solution would you suggest?

I suggest to follow that implementation and also drop the "Report problem" with predefined log templates.

That would look like this:
![grafik](https://github.com/cgeo/cgeo/assets/949669/1a93cf38-4ce1-4c5a-a59e-3ab99a84dc61)

### Is there an alternative?

I think the "Report problem" function will stop working with the new log page anyway?!
So I guess we have to change anyway.

Lineflyer:
BTW:
That brings back the follow up feature request (should be somewhere in our issues) to support more than one offline log (e.g. "Found it" + "Needs maintenance". 
At the moment you can use the offline "Found it" including "Report problem", which will then no longer be possible.

Lineflyer:
Side note:
The log types are already all listed and available in c:geo.
So we would just need to do two things:
1. Rename the NM, NA log types to follow the new website wording
2. Drop the "Report problem" field and logging flow from c:geo

eddiemuc:
> BTW: That brings back the follow up feature request (should be somewhere in our issues) to support more than one offline log (e.g. "Found it" + "Needs maintenance". At the moment you can use the offline "Found it" including "Report problem", which will then no longer be possible.

If the main reason for this feature is to allow for e.g. a "found" log and a "maintenance" log to exist at the same time, then my suggestion is to just leave the "report problem" dropdown in c:geo log screen and then to send two logs in the background on "send log" (one with the actual chosen type and a second one with a type depending on the selected value in "report problem"). 

Allowing c:geo to keep multiple offline logs for same cache just for this use case would be overdesigned imho - this would be a larger change.

SammysHP:
The "report problem" function is also required for OpenCaching.

Lineflyer:
Good arguments. Lets keep what we have.

bitlux:
Why is the renaming proposed in https://github.com/cgeo/cgeo/issues/14685#issuecomment-1748351254 not planned? The log types were renamed in 2023: https://www.geocaching.com/blog/2023/11/renaming-of-maintenance-log-types/.

[...]

PR #14687, opened 2023-10-05
Work on #14685 - Align log strings to website

Lineflyer:
## Description
Align log type strings to those used on the Groundspeak website

## Related issues
#14685


Lineflyer:
On second thought I agree...lets drop it.

PR #14690, opened 2023-10-05
Fix missing distance info in certain setting (fix #14688)

### LINKED ISSUE #14688: [Nightly] Map -> Navigation:  Straight Distance field missing (if there's no Route Distance to be displayed)
### Describe your problem!

Maybe s/th went wrong while implementing the new elevation field on map screen, cause since I read about the new ``Show Elevation`` option (and activated it out of curiosity), the ``Straight distance`` field, which was always visible before, is missing if there is no ``Routing distance`` to be shown, resp. appears and disappears, simultaneously with the ``Routing distance`` field.

BTW (and before opening a dedicated FR for it): Would it be a big deal to show the current speed, too, along with the distances and the elevation?

### How to reproduce?

1. in ``Settings`` => ``Routing / Navigation`` activate ``Show straight distance``
2. in map screen select any cache and route to it
   Note: easiest way to reproduce the problem is to take a cache that is just out of reach by car, because then the routing algorithm tries to recalculate every second or so, which makes the problem more visible
3. set routing to ``Direct line`` and notice that the "meant to be always on" ``Straight distance`` field is missing
5. set routing to ``Car`` and notice that ``Straight`` and ``Routing distance`` are flickering simultaneously

``Straight distance`` is missing, if there is no ``Routing distance`` to be displayed.

``Straight distance`` should always be shown.

### Reproducible

Yes

### c:geo Version

2023.10.05-NB-c419f8a

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Reenables display of (straight) distance info when "show straight distance info" is enabled in settings and routing mode is set to STRAIGHT

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapDistanceDrawerCommons.java
@@ -10,7 +10,7 @@ package cgeo.geocaching.maps;
  * - distances1/distances2 placeholder will be filled with straight distance, routed distance,
  *   individual route length and elevation info (depending on settings and current data)
  * - By tapping on any of the distance fields either straight or routed distance can be supersized.
- *   It gets removed from the distance view containers and displayed in a larger font:
+ *   It gets removed from the distance view containers and displayed in a larger font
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

PR #14691, opened 2023-10-05
Fix dialog formatting issues for multi select (rel. to #14602)

moving-bits:
## Description
Fix font size of new multi selection dialog. Item spacing is a bit too dense compared to the old dialog, so setting WIP for now.


moving-bits:
Let's try it in the nightlies

PR #14703, opened 2023-10-08
Add two user-defined routing profiles (fix #14700)

### LINKED ISSUE #14700: Routing profiles: Add "Boat" resp. "Canoe" and "MTB"
### Tell us your idea!

In [BRouter's routing profile downloads](https://brouter.de/brouter/profiles2/) I just stumbled upon "river.brf" - and was immediately caught by the arousing possibilities, because that's exactly what we've always missed when going for boat caches so far.

BTW: Perhaps [Canua](https://play.google.com/store/apps/details?id=com.touchingcode.canua&hl=en&gl=US) may be added to [c:geo's useful apps section](https://manual.cgeo.org/en/mainmenu/utilityprograms) cause even in the free version it's extremely helpful for planning boat caches.

However: Some "dry" couch tests showed good results - thus it would be great to have a corresponding button in the map quick settings (and of course profile settings, to choose which profile to use - default should be "river.brf", of course).

And perhaps a second choice for biking may be implemented in the same go, as we often switch between road bike (for commuting) and MTB (mtb.brf - for the fun) - but being forced to always change the profile via the route settings is a bit annoying. 

Icons could be like e. g. in [Komoot](https://www.komoot.de/plan/@49.0000000,9.2149000,9.000z?sport=mtb&constitution=4) resp. a minimalistic canoe like [this one](https://www.flaticon.com/de/kostenloses-icon/ruderboot_3124366) on flaticon.com.

Routing Icons Boat

Additional routing profile choices for "Boat" resp. "Canoe" and "MTB".

### Is there an alternative?

Switching profiles in routing settings.

### Additional Information

_No response_

moving-bits:
## Description
Adds two user-defined routing profile for easier toggling between multiple user-selected routing profiles. See discussion in #14700.

image

Names of selected user-defined profiles will be shown below toggle button row.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapSettingsUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapSettingsUtils.java
@@ -1,6 +1,7 @@
 package cgeo.geocaching.maps;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.brouter.BRouterConstants;
 import cgeo.geocaching.databinding.MapSettingsDialogBinding;
 import cgeo.geocaching.enumerations.WaypointType;
 import cgeo.geocaching.filters.core.GeocacheFilter;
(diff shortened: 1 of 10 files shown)


MagpieFourtyTwo:
Maybe I'm wrong, but wouldn't it be better to hide these two profile buttons from the Quick Settings menu, if internal routing is *not* activated ...?

moving-bits:
Should be the case since latest nightly

MagpieFourtyTwo:
Yep - working like a charm - thanks a lot!

PR #14704, opened 2023-10-08
gclogapi: enable for beta

eddiemuc:
This PR, if merged, enables new gc log APi for beta


main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -706,7 +705,7 @@ public static boolean enableFeatureNewGCLogApi() {
         if (contains(R.string.pref_feature_new_gc_log_api)) {
```
moving-bits, line 705: Shouldn't we remove this method and the related preference key completely? Or do you want to leave that in temporarily as safety guard?

eddiemuc: I want to have a way to disable the new log API in case major bugs come up (so users have a way to quickly disable new API).

Old code will be removed when API proves to be stable in production for a while.

moving-bits: ok, understood, that's what I meant with "safety guard"

PR #14711, opened 2023-10-10
Fix #14710 - Do not use autofill for geo and tracking code

### LINKED ISSUE #14710: Trackable search field is tagged as a password field?
### Describe your problem!

All the other search fields act as normal input fields but the "Trackable identification" field brings up Google password manager offering to use saved passwords in it.
I often have to tap away the overlay that appears, when I go to search for a trackable.
Now, I'm not quite sure if this is specific to certain version of Android but it surely is a case of the search fields somehow being tagged differently because for no other fields it ever offers to use a saved password.
Screenshot_20231010-120431

### How to reproduce?

It doesn't appear every time but if I close c:geo, then force stop it and clear the cache, next time I will open it and go to search for the trackable, it pops up again.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2023.09.26

### System information

```text
Device: Pixel 4A (5G)
Android: 13, build TQ3A.230901.001
```

### Additional Information

_No response_

Lineflyer:
## Description

Edit:
Solution is now to disable autofill hoping that this will disable password suggestions, despite the text fields still being declared as visible passwords.

## Related issues

fixes #14710

### DIFF
--- a/main/src/main/res/layout/search_activity.xml
+++ b/main/src/main/res/layout/search_activity.xml
@@ -59,9 +59,10 @@
             <AutoCompleteTextView
                 android:id="@+id/geocode"
                 style="@style/textinput_embedded_singleline_go"
-                android:inputType="textVisiblePassword"
                 android:text="GC"
-                tools:ignore="HardcodedText" />
+                tools:ignore="HardcodedText"
+                android:inputType="textVisiblePassword"
+                android:importantForAutofill="no" />
         </com.google.android.material.textfield.TextInputLayout>
 
         <Button
@@ -181,7 +182,8 @@
             <AutoCompleteTextView
                 android:id="@+id/trackable"
                 style="@style/textinput_embedded_singleline_go"
-                android:inputType="textVisiblePassword" />
+                android:inputType="textVisiblePassword"
+                android:importantForAutofill="no" />
         </com.google.android.material.textfield.TextInputLayout>
 
         <Button


fm-sys:
`textNoSuggestions` sounds good! Maybe it didn't exist back then...?

Lineflyer:
but why not suggest formerly entered codes?

bekuno:
For the history of this entry see #7149

main/src/main/res/layout/search_activity.xml
```diff
@@ -180,8 +179,7 @@
             android:hint="@string/search_tb_hint" android:labelFor="@id/trackable">
             <AutoCompleteTextView
                 android:id="@+id/trackable"
-                style="@style/textinput_embedded_singleline_go"
-                android:inputType="textVisiblePassword" />
+                style="@style/textinput_embedded_singleline_go" />
```
moving-bits, line 182: I read the linked issue and played around a bit with Android's textInput settings. In the end I came up with the following:

```
<AutoCompleteTextView
    android:id="@+id/trackable"
    style="@style/textinput_embedded_singleline_go"
    android:digits="ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
    android:inputType="textCapCharacters" />
```

This will show an alphanumeric keypad, allowing direct input of letters (automatically upper-casing them) and number input on long-press.

GC does not use all letters, but as I'm not sure which letters are allowed for the other connectors, I left in the full set A-Z.

Lineflyer:
Cool, will adapt this PR accordingly and try it out before pushing.

Lineflyer:
According to [this older thread](https://stackoverflow.com/questions/58180576/disable-google-password-manager-list-out-saved-passwords-in-android) it could work to still user `textVisiblePassword` but add an `android:importantForAutofill="no"`` as well.

It seems this flag not only disables suggestions but also the storing:
image

main/src/main/res/layout/search_activity.xml
ztNFny: That restores the state before https://github.com/cgeo/cgeo/issues/7148 though? The point of textVisiblePassword is to force the number row, not just by long-press

ztNFny:
editText.setInputType(InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS)
could also work

Lineflyer:
I cannot reproduce the issue on same API version and same preconditions the issue author described.

Trying to keep the convenient input type (with visible numbers) I suggest to try with `android:importantForAutofill="no"` only and request feedback from issue author either by using nighlies or next beta/release.
@moving-bits Your opinion?

eddiemuc:
For what it's worth: the way I solve this is to use GBoard and set the option to always display digits on qwertz/y keyboards.

Lineflyer:
Ready for merge

Lineflyer:
> editText.setInputType(InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS) could also work

Yes, but it did not have the advantage of a number row being shown but shows the "normal" keyboard (at least on the different emulators I tried).

main/src/main/res/layout/search_activity.xml
Lineflyer: Found another solution. Number row is in again.

moving-bits (approved):

fm-sys:
Using `textVisiblePassword` is a really bad workaround!

In fact the behavior is highly dependent on the keyboard you are using.

One my device (using SwiftKey keyboard) e.g. the clipboard feature is missing when using `textVisiblePassword`. Which is IMHO way worse than a missing number row. BTW, you can choose whether to always show the number row at least for some keyboard apps. IMHO it's not up to c:geo to "optimize" the keyboard in such a hacky way. 

Normally | with textVisiblePassword
-|-
Screenshot_20231011_230033_cgeo geocaching|Screenshot_20231011_230050_cgeo geocaching

PR #14715, opened 2023-10-11
hide the whole "unified map" preferences category if not enabled instead of only 1 setting

ztNFny:
PreferenceCategory pref_fakekey_unifiedmap has 2 entries but only one of them was hidden depending on R.string.pref_showUnifiedMap


moving-bits:
Thanks for looking into this.
IMHO this fix could go to `release`, as we have this bug already on `release`. (I tried to rebase it to `release`, but that did not work, so I rebased it back to `master`.)

ztNFny:
retargeted

Issue #14649, opened 2023-09-24
On map don't show final-waypoint if coords match cache-coords


ztNFny:
Waypoint icon shown behind cache leads to strange display as the waypoint has cache type as marker. The waypoint marker has always been drawn, but as it sits behind the cache marker it's usually not visible.
![image](https://github.com/cgeo/cgeo/assets/1258173/d61c5c62-bdeb-4c81-83a3-6d9dd83ce796)

To repoduce:
1. Create a waypoint
2. set waypoint-coords as cache coords
3. view cache on map

Suggested fix:
Option a) don't draw waypoint marker if waypoint coords == cache coords
Option b) draw waypoint marker but do not draw cachetype marker if waypoint coords == cache coords

Lineflyer:
I also noticed that after we added the small overlays to the waypoint. It was like that ever since, but only now its disturbing as the overlay is visible. +1 for hiding them.

ztNFny:
Fix seems incomplete, still too many WPs visible

PR #14716, opened 2023-10-11
don't draw waypoints without note at cache coords

ztNFny:
Better fix for https://github.com/cgeo/cgeo/issues/14649 (Waypoint icon shown behind cache icon) by not drawing the waypoint icon on map if waypoint coords are the same as cache coords.

Old fix covered only Google Maps, this also avoids some corner cases.

UnifiedMap doesn't support drawing Waypoints yet but a similar implementation will be required there when it does.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
@@ -1407,9 +1407,14 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
                 final boolean forceCompactIconMode = CompactIconModeUtils.forceCompactIconMode(cachesCnt);
                 if (mapOptions.mapMode == MapMode.SINGLE || cachesCnt < Settings.getWayPointsThreshold()) {
                     for (final Waypoint waypoint : waypointsToDisplay) {
-                        if (waypoint != null && waypoint.getCoords() != null && waypoint.getCoords().isValid() && !(waypoint.getCoords() == waypoint.getParentGeocache().getCoords() && waypoint.getNote().isEmpty())) {
-                            itemsToDisplay.add(getWaypointItem(waypoint, forceCompactIconMode));
+                        if (waypoint == null || waypoint.getCoords() == null || !waypoint.getCoords().isValid()) {
+                            continue;
                         }
+                        final Geocache cache = waypoint.getParentGeocache();
+                        if (waypoint.getNote().isEmpty() && cache != null && cache.getCoords() != null && cache.getCoords().isValid() && waypoint.getCoords().equals(cache.getCoords())) {
+                            continue;
+                        }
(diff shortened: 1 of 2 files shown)


ztNFny:
retest this please

moving-bits:
I understand the reasoning behind this PR, trying to avoid a dangling cache type icon marker behind the original cache icon, and would follow this logic for waypoints of type "original coordinates". But isn't there a conflict of goals with stages of a multi cache? 

Frequently stage 1 has the same coordinates as the cache itself, and on creating a route you will probably add all (known) stages to your route. The route will be drawn, even without the waypoint marker being visible, but the user cannot easily mark this waypoint as visited anymore or remove the waypoint from the route. Both those tasks have mitigations, but it will be somewhat confusing for the user (and probably cumbersome). (And there may be other scenarios.)

(For lab adventures and some multi stages it may work if not only user-defined waypoint notes are checked but listing waypoint notes as well, haven't checked in the code yet.)

Mitigation for this scenario could be to not draw the cache type marker for such waypoints (which would be inconsistent with other waypoints, though). Any other ideas?

ztNFny:
The waypoint is drawn if the description is non-empty, which is the case for all labs and most multis.
During implementation I was thinking about "parking" WPs - but then I decided that those don't really matter - if there is one at the cache coords it's the same whether you navigate to WP or cache and on map you can't see it anyway.

ztNFny:
> Frequently stage 1 has the same coordinates as the cache itself, and on creating a route you will probably add all (known) stages to your route.

If you can't see S1 you can't add it to the route though? So the route would consist of listing (now even faster to add as you don't get the popup for selecting S1 or listing!), S2, S3, ...

moving-bits:
> The waypoint is drawn if the description is non-empty, which is the case for all labs and most multis.

That's good, and probably will cover most use cases.

> > Frequently stage 1 has the same coordinates as the cache itself, and on creating a route you will probably add all (known) stages to your route.
> 
> If you can't see S1 you can't add it to the route though?

You can add stages (waypoints) also from waypoint list ("append all to individual route").

And marking this stage/waypoint as visited will then require you to open the cache, move to waypoint tab, find the right waypoint, and mark it as visited from there. Or by opening the waypoint via route item list.

But I guess there will be no automatism covering all use cases. Not sure which way we should go for this (admittedly) edge-case.

ztNFny:
> You can add stages (waypoints) also from waypoint list ("append all to individual route").

Oh, that seems new?

> And marking this stage/waypoint as visited will then require you to open the cache, move to waypoint tab, find the right waypoint, and mark it as visited from there

If it's not on the map no need to mark it visited [;)] To remove from route you could use the routing menu.
I think this combination of "append all waypoints to route" and a S1 WP without description at cache coords is really unlikely to happen. And for that very rare case there are workarounds. Compared to the current state it's certainly the way smaller problem.

PR #14718, opened 2023-10-11
Regrant Uri permissions on restore

moving-bits:
## Description
When restoring a backup which includes persisted URI, the user needs to regrant access to those files.

We had this for persisted track file and removed functionality when we switched to supporting multiple track files with a different storage method. Now we have persisted URI for proximity notification audio files and need to re-enable this mechanism.


moving-bits:
retest this please

Issue #14672, opened 2023-10-01
Lab adventures losing found state on zoom


moving-bits:
User report on support:

"I am very happy with the new feature that lab caches automatically become visible on the map as found.  but unfortunately that doesn't work very well for me.  When I zoom in or out on the map, the color of the smileys suddenly changes back to purple.  If I close the app and start it again, they are yellow again until I zoom in or out again"

"It happens with all foud labs ! It happens with stored and unstored labs ! And always when i start cgeo i see them as found, but when i zoom in or out on the map the labs become visible as not found yet. But when i set them as found by manual option the stay visible as found."

### Extract from system information

c:geo version: 2023.08.24

Device:
-------
- Device type: ONEPLUS A3003 (OnePlus3, OnePlus)
- Android version: 9
- Android build: ONEPLUS A3003_28_191104
- Google Play services: enabled - 23.35.15 (100400-565551795)
- Memory: Available:1,7 GB, Total:5,6 GB, Threshold: 216,0 MB, low:false

Program settings:
-------
- Settings: v9, Count:211
- Set language: nl_NL (system default)
- System date format: dd-MM-yy
- Map: Google: Satellite

Filters:
-------
- Hide waypoints: -
- LIVE: Status: Uitgeschakeld uitsluiten, Gearchiveerden uitsluiten ([:inconclusive=false:advanced=false]AND(type;difficulty_terrain:d=-:d=-:t=-:t=-;status:exclude_disabled:exclude_archived))
- OFFLINE: Geen ([:inconclusive=false:advanced=false]AND(type;difficulty_terrain:d=-:d=-:t=-:t=-;status))

Stored Filters (#0):

Services:
-------
- Geocaching sites enabled:
- geocaching.com: Logged in (Ingelogd) / PREMIUM
- Geocaching.com Adventure Lab
- opencaching.de: Logged in (Ingelogd)
- opencaching.nl: Logged in (Ingelogd)
- Geocaching.com date format: dd/MMM/yyyy
- Geocaching.com website language: nl-NL
- Last login error on geocaching.com: communication error (2023-07-02 12-46)
- Last successful login on geocaching.com: 2023-09-21 19-38
- Routing: internal / BRouter installed: false
- Installed c:geo plugins: contacts
[...]

### additional information
ticket 581385

baiti:
[...]

PR #14729, opened 2023-10-14
ALC Loosing Found State (issue #14672)

baiti:
## Description

An attempt to address the issue of `searchByGeocode()` not giving us the true found state of a lab cache. The envisioned solution is to do a subsequent `search()`, which actually seems to be superfluous since we actually got the searched object, however, we got it without the found state. The `search()` function gives us the found state so we merge that into the object we wanted to fetch in the first place. This fix is a circumvention for a missing functionality in the api we're using.

## Related issues

issue #14672

## Additional context

I have not been able to verify whether this fix actually resolves the referenced issue. Reason is that I don't have a userid which I can use. It only works with a premium account, however, with my premium account I don't do lab caches ;)

P.S.: I have removed two methods which were flagged as being "not used". If those should remain in the code, I will hve to update the PR accordingly. Pls let me know.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -62,7 +62,6 @@ final class ALApi {
     private static final String LATITUDE = "Latitude";
     private static final String TITLE = "Title";
     private static final String MULTICHOICEOPTIONS = "MultiChoiceOptions";
-
     private static final int DEFAULT_RADIUS = 10 * 1000; // 10km
 
     private ALApi() {
@@ -148,6 +147,13 @@ final class ALApi {
         }
     }
 
+    // To understand the logic of this function some details about the API is in order.
+    // The API method being used does return the detailed properties of the
+    // object in question, however it does not return the true found state of the object so
+    // we have to do an additional search, a search which will give us much less details about
+    // the object but does indeed give us the true found state of the object. Once we got
+    // that information, we merge it into the object we wanted to lookup initially.
+
     @Nullable
     @WorkerThread
     protected static Geocache searchByGeocode(final String geocode) {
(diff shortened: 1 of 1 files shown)


Lineflyer:
If I am reading the code right, this change is done not only for the found state of lab caches but checking the found state of all results. 
IMHO we should give this change more time to test in nithtlies, which would mean to only merge it once we branched of for the next beta. Only my few cents without deep code knowledge...

baiti:
> If I am reading the code right, this change is done not only for the found state of lab caches but checking the found state of all results. IMHO we should give this change more time to test in nithtlies, which would mean to only merge it once we branched of for the next beta. Only my few cents without deep code knowledge...

It does a search based on a `Geopoint` within a radius of 1 meter. In most cases the search will return a Collection with just one object, the one we're interested in. Even if it returned multiple objects (within that 1 meter radius) I am comparing the `gecocode` aka `alc-id` for a match and only on a match I am extracting the found state.

Also, the `search()` function I am using is exclusively scoped at adventure lab caches, it does not target any other geocache objects managed by other connectors. The function can never return any other object that an adventure lab cache.

Of course I am ok with postponing a merge for as long as you think it is appropriate. I am even more ok with not throwing it into the upcoming beta.

moving-bits (approved):
LGTM from looking at the code (haven't tested actual execution - currently have no caches in reach for me that fit the criteria).

moving-bits:
WRT merging this PR: As we are planning to start a new beta phase today, which is typically two weeks, I see two choices:
- we could merge it today and include it directly in upcoming beta - change is restricted to certain cache type and radius only, LGTM, and would be easy to remove from beta, if need should be.
- or we give it a couple of days in the nightlies first (hoping that some people text explicitly this area), and cherry-pick it into beta version, then preparing a second beta.

Currently I tend to option one.

ztNFny:
I will try to test this tonight, can't promise anything tho

moving-bits:
Let's try it in the nightlies

PR #14747, opened 2023-10-20
open downloader activity when clicking on system downloader notification

fm-sys:
open downloader activity when clicking on system downloader notification

note that this does not yet open the correct subfolder as this would be much more complex. Still way better than nothing ;-)

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -929,6 +929,9 @@
             <intent-filter>
                 <action android:name="android.intent.action.DOWNLOAD_COMPLETE" />
             </intent-filter>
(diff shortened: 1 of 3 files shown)


moving-bits:
What is the goal of linking the notification? Just to have something available, or to add more downloads, or something else?
One could use the "pending downloads" view instead, it will give you detailed info on the download currently running (as well as the failed ones). Keep in mind that downloads may not only originate from map downloader, but from routing tiles etc. as well.

fm-sys:
Yes, it's just to have something available. Pending downloads is probably more meaningful, so I second your idea.

PR #14752, opened 2023-10-21
Use icons for distance and elevation info (fix #14744)

### LINKED ISSUE #14744: Elevation info needs a better icon
### Describe your problem!

I talked to some beta users and even myself was fooled by the elevation info on the map.
As it shows "a ramp" users might first thing, that it shows an elevation difference (e.g. if a route is active in parallel).
This is emphasized by the fact, that its shown in the same screen area, which normally only shows distances to target, routing length, etc.

At least we should IMHO try to find a better icon, ideally we could also find a better place to show that info on the map screen (e.g. upper left instead upper right) to visualize its not related to routing and distance information.

Ideas for icons....did not check usage conditions:
image
Source: https://www.flaticon.com/de/kostenloses-icon/elevation_9620066

### How to reproduce?

see above

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

Nightly/Beta

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Replaces Unicode characters by icons for distance and elevation symbols in map view.
Uses selected routing type-specific icons instead of generic "wavy line" symbol, if applicable.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/TextParam.java
+++ b/main/src/main/java/cgeo/geocaching/ui/TextParam.java
@@ -11,6 +11,7 @@ import android.text.TextUtils;
 import android.text.method.LinkMovementMethod;
 import android.widget.TextView;
 
+import androidx.annotation.ColorInt;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
@@ -45,6 +46,7 @@ public class TextParam {
 
     private ImageParam image;
     private int imageSizeInDp = -1;
+    @ColorInt private int imageTintColor = 1;
 
 
     /**
@@ -118,6 +120,14 @@ public class TextParam {
         return this;
     }
 
+    /**
+     * set tint color for compound image
+     */
+    public TextParam setImageTint(@ColorInt final int imageTintColor) {
+        this.imageTintColor = imageTintColor;
+        return this;
+    }
+
     private TextParam(@StringRes final int textId, final CharSequence text, final TextParam[] concatTexts, final Object... params) {
         this.textId = textId;
         this.text = text;
@@ -138,7 +148,7 @@ public class TextParam {
      * Applies the current settings of this TextParam to a textview.
      * Parameter forceNoMovement allows to force not setting a movement method even if other params suggest. This is important
      * if TextParam is used in a context where resulting TextView needs to remain clickable by itself
-     * * Sets text returned by {@link #getText(Co ntext)}
+     * * Sets text returned by {@link #getText(Context)}
      * * Calls {@link #adjust(TextView, boolean)} on the textview
      */
     public void applyTo(@Nullable final TextView view, final boolean forceNoMovement) {
@@ -227,6 +237,13 @@ public class TextParam {
                 view.setCompoundDrawables(imageDrawable, null, null, null);
             }
 
+            // set image tint (if given)
+            if (imageTintColor != 1) {
+                final Drawable[] d = view.getCompoundDrawables();
+                if (d.length > 0) {
+                    d[0].setTint(imageTintColor);
+                }
+            }
             //Add margin between image and text (support various screen densities)
             view.setCompoundDrawablePadding(ViewUtils.dpToPixel(10));
         }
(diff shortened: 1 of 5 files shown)


main/src/main/java/cgeo/geocaching/ui/TextParam.java
```diff
@@ -118,6 +119,14 @@ public TextParam setImage(final ImageParam image, final int imageSizeInDp) {
         return this;
     }
 
+    /**
+     * set tint color for compound image
+     */
+    public TextParam setImageTint(final int imageTintColor) {
```
eddiemuc, line 125: I assume this is a color int? Then you should annotate the parameter with `@ColorInt`

eddiemuc (approved):

main/src/main/java/cgeo/geocaching/ui/TextParam.java
moving-bits: thx, have added the annotation

ztNFny:
@moving-bits I noticed this is being overhauled anyway - could I suggest to adjust the icon size while at it? They are currently too large in relation both to the neighboring text as well as toolbar icons. If you want I can take care of tuning the sizes (started on that yesterday evening but didn't really have time) if you want - just let me know once it's in a state where that makes sense.

moving-bits:
Sure, some adjustments are needed here - icons in distance views are a bit too large in the right column, and way too small in supersize view (although one could discuss whether a large icon is really needed for supersize view):

image

I'm reusing the icons used in map quick settings for routing mode resp. the route/track popup icon, only elevation icon is new (and defaults in terms of sizes to what comes from Google fonts).

To my understanding adjusting the icon sizes would have to be done in `UnifiedTargetAndDistancesHandler.syncView()` by adding the `imageSizeInDp` parameter to the `setImage` call.

moving-bits:
"Whole route length" icon is the same as route/track popup icon at the map's bottom (intentionally). 
It's a default Google font icon when you search for "route", but we may choose/create a different one.

I propose to keep the elevation icon for now until we have decided what elevation info we want to have, eg:
- { current | tap } position elevation, 
- { target | individual } route { accumulated incline | remaining accumulated include | height diff between start and end | height diff between current position and end | ... }
- ...

wrt to alignment of small distance views: 
- There are two view containers in `map_distanceinfo.xml`, which get filled dynamically. One container is for the upper views (used for first info, if supersize is activated / for all infos, if no supersize) and one for the views below the supersize view.
- Currently their alignment is kept individually to maximize the space available for target name to the left, and map to the bottom.
- Alignment probably needs to be done programmatically (compare and set view sizes).

ztNFny:
Alignment and space usage could be optimized by not having the big distance take the whole width but in 1 line with the two smaller distances:
image

moving-bits:
Yes, this would things make much easier technically, as we would need only one container for the small distance views and no manual alignment. But I wanted to maximize the space available for the supersize view. If you are on a bumpy road with your bike, font size matters :-)

ztNFny:
There's plenty of width available on all my devices - but granted, I don't have "very narrow" or ooold small screens in use.
We could omit the unit in large view and make it always 4 digits max? You have some idea whether you are 100 m or 100 km away from the cache, so that shouldn't cause too much confusion. Or as a compromise make the unit smaller - or worded differently: Emphasize the number! ;-)

moving-bits:
> There's plenty of width available on all my devices - but granted, I don't have "very narrow" or ooold small screens in use. 

On a 6.1" screen it's easy to hit the width limit (even without right column) if you increase the font size to one of the larger levels.

> We could omit the unit in large view and make it always 4 digits max? 
> You have some idea whether you are 100 m or 100 km away from the cache, so that shouldn't cause too much confusion. Or as a compromise make the unit smaller - or worded differently: Emphasize the number! ;-)

Number of digits is already limited, see `Units.getDistanceFromKilometers()`.
I would prefer to leave the unit in. Smaller would be ok for me. (But that's just my personal preference.)

One could leave out the icon in supersize view, at least if both distances (straight & routed) are shown, as the other one in small distance view would have a symbol and thus makes it deterministic.

Not sure if I like to have the supersized text off-centered, though, would need to test it in real life. Might look a bit strange if no or only one of the small distance views are being displayed.

ztNFny:
If no small views are shown big one can use full width

moving-bits:
My most prominent use case is displaying route length info as small view top right, and distance to current target in supersize mode. Currently, the small view is right beside the target cache's name, and supersize box is below, so still very compact, and text is centered.(*)

image.image

_(*) At least it has been centered before, until we switched from Unicode-based symbol to separate icon view. Still pretty close, but should be fixed, IMHO._

Issue #12927, opened 2022-04-24
Unified map: add support for proximity notifications


moving-bits:
### Missing functionality

Add support for proximity notifications

(Requires cache loading and handling having been implemented first)

PR #14754, opened 2023-10-22
UnifiedMap: Enable map quick settings

moving-bits:
## Description
Enables the map quick settings menu. Most of its functions do work already, including changing routing mode.

Functions yet missing are:
- [x] circles
- [ ] proximity notifications are not yet available in UnifiedMap (see #12927 as reminder)
- [x] switching compact icon mode does not yet trigger drawing small icons

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/layers/GeoItemsLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/layers/GeoItemsLayer.java
@@ -1,6 +1,5 @@
 package cgeo.geocaching.unifiedmap.layers;
 
-import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.maps.CacheMarker;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Waypoint;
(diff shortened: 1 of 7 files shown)


fm-sys:
I can help here as well if you like...

moving-bits:
> I can help here as well if you like...

sure, very welcomed 👍

main/src/main/java/cgeo/geocaching/unifiedmap/layers/GeoItemsLayer.java
```diff
@@ -32,16 +34,22 @@ public GeoItemsLayer(final AppCompatActivity activity, final GeoItemLayer<String
 
             final Map<String, Integer> currentlyDisplayedGeocaches = new HashMap<>();
 
+            final boolean forceCompactIconMode = CompactIconModeUtils.forceCompactIconMode(caches.size()); // todo: only use caches in viewport and update while moving the map
```
moving-bits, line 37: We need to count caches in viewport anyway, to be able to update the cache counter in toolbar, so maybe this can be put into the viewmodel cache observer, updating GeoItemLayer and holding a counter available for UnifiedMapActivity for triggering toolbar update after filtering?

fm-sys: We IMHO need to observe map movements directly, as the cache list is only refreshed if necessary. This is however not yet designed in our map specific implementations.

fm-sys:
Done from my side. I didn't work on proximity notifications as I neither have ever used them nor I consider them important for now.

moving-bits:
Nice! Thanks for working on it. PN has a separate issue, so we can track it easily.

Issue #14763, opened 2023-10-26
Distance to cache info missing on map


Lineflyer:
### Describe your problem!

Frequently reported for new release 2023.10.25:

Users claim, that after updating the distance to a cache is missing when opening the map for a cache, while it was there before the update.

One user adds, that after switching routing mode to "straight" and then back, the distance gets displayed then.

### How to reproduce?

TBD
Several support tickets to be analyzed

### Reproducible

Unclear

### c:geo Version

2023.10.25

### Additional Information

Tickets 118609 (Google Satellite), 720867 (OSM offline), 563794 (OSM offline)

moving-bits:
should be resolved with release 2023.11.01, otherwise feel free to reopen

Issue #14765, opened 2023-10-26
Distance info flickering on map


Lineflyer:
### Describe your problem!

Reported by several users on social media and support:

The distance info on map is flickering/flashing

One user reports further detail:
- distance info tends to not shop up while standing still (but not always), and flickering while moving slowly (walking)
- distance info is visible constantly while moving fast (car)

### How to reproduce?

TBD from support tickets

### Reproducible

Unclear

### c:geo Version

2023.10.25

### Additional Information

Ticket 717381 provides a video recording. Distance info is constantly switching on an off for reason unknown yet.
Further tickets: 972832, 179403

moving-bits:
As we have not yet been able to reproduce this issue in our test environment: 
Is some of the users affected by this issue able to test a manually installed version of c:geo to check whether that version fixes this issue?

Download is available from PR #14766 and can be downloaded here: https://ci.cgeo.org/job/cgeo-CI_PR-build/6361/artifact/main/build/outputs/apk/basic/debug/cgeo-debug.apk (or see the linked PR, check in "Show all checks" section, follow the "Details" link in "default" line and install the "cgeo-debug.apk" (which is the same link as above)).

fm-sys:
Cannot reproduce either. Maybe limited to a specific android version?

GeoPineseeker:
I can produce this on Pixel3 (CalyxOS AOSP 13 with latest updates) and Pixel6 (CalyxOS AOSP 13 with latest updates), sent a support ticket for Pixel6 with details, for which the id is 103602.

GeoPineseeker:
Debug version (https://ci.cgeo.org/job/cgeo-CI_PR-build/6361/artifact/main/build/outputs/apk/basic/debug/cgeo-debug.apk) does fix the issue on pixel 3 (CalyxOS AOSP 13 with the latest updates).

GeoPineseeker:
After some further testing, it seems that if I select from the gear icon routing mode as straight line (dog bone icon) then the distance is constantly on. But then it seems that I'm unable to click the distance to make it use larger font. Also with the debug version it was flashing "walking" and "straightline" icons constantly.

moving-bits:
Thanks for your testing efforts. Do you have "Show straight distance" enabled in settings => routing / navigation?

GeoPineseeker:
I cannot find a setting "Show straight distance" anywhere under settings -> routing / navigation.

moving-bits:
[...]

PR #14766, opened 2023-10-27
Restructure distance data collection

moving-bits:
## Description
Restructures the way distance data is collected and interpreted back to the same way it was before #14668, but keeping the technical refactoring etc.

Attempts to fix issue #14763 and maybe #14765.

I was not able to reproduce any of those two issues, so I cannot check the effectiveness of this PR.
Thus: Can anyone out there having one or both of those issues check the build from this PR?
(Otherwise we might put it into the nightlies, update related issues, and wait a bit for feedback.)


moving-bits:
retest this please

moving-bits:
Let's test this in the nightly version

Issue #14731, opened 2023-10-15
Next feature release 2023.10.xx


moving-bits:
I'll start preparing a new beta version tonight. Let's use this thread to collect any issues that arrive during beta phase that need to be fixed before final release. Feel free to edit this initial post for adding such issues:

- [ ] ?? #14736
- [x] #14681
- [x] #14729 (cherry-pick from `master`)
- [x] #14744
- [ ] ?? #14745
- [x] #14760

moving-bits:
I'll start preparing a new beta version now:

- [ ] Empty `changelog_bugfix.md` on branch `master`
- [ ] Wait for a Crowdin cycle or trigger sync manually for branch `master`
- [ ] Merge the related Crowdin translation pull requests for branch `master`
- [ ] Delete outdated translations branch for `release`
- [ ] Fast forward `release` branch to HEAD of `master`
- [ ] Check/insert correct version name in `BranchDetectionHelper.java` on branch `release`
- [ ] Build release on CI server
- [ ] Perform smoke test with resulting ```cgeo-RC.apk``` before continuing
- [ ] Upload and publish ```cgeo-RC.apk``` to Google Play on beta channel
- [ ] Trigger notification to status.cgeo.org via CI server
- [ ] Post release info to Facebook 
- [ ] Post release info to Twitter

bravewolfcz:
I don't know how to edit the opening post, so I'll write a new one.
The Show elevation info for current position on map option will start chopping the map when enabled. After turning off the elevation option, the map moves smoothly again.
Pixel 6 Pro Android 14
C:geo 2023.10.15-RC2

Lineflyer:
Added #14681 above

moving-bits:
I propose to include #14729 in beta version as well (cherry-pick from `master`). It's available in master since a couple of days, we have no complaints about this fix at all (though, admittedly, it's a specific use-case), review was fine, and it's limited to AL connector anyway.

moving-bits:
I'll start preparing a new beta release now:

- [ ] Wait for a Crowdin cycle or trigger sync manually for branch `release`
- [ ] Merge the related Crowdin translation pull requests for branch `release`
- [ ] Cherry pick #14729 to `release`
- [ ] Check/insert correct version name in `BranchDetectionHelper.java` on branch `release`
- [ ] Build release on CI server
- [ ] Perform smoke test with resulting ```cgeo-RC.apk``` before continuing
[...]

PR #14767, opened 2023-10-27
Revert distance info refactoring and elevation info addition

moving-bits:
## Description
This is an alternative approach to #14766, which reverts the distance views refactoring and all things based on it (new icons, elevation info), as discussed in https://github.com/cgeo/cgeo/issues/14731#issuecomment-1782526599


moving-bits:
Implemented fix seems to work fine, no need for a revert anymore, so let's close this PR

Issue #14768, opened 2023-10-27
Elevation chart for route / individual route


moving-bits:
### Tell us your idea!

It would be nice to see the change of elevation along a route, be it a route to currently selected target or an individual route.

### What solution would you suggest?

Add an option to display such a chart, vertical axis would be elevation, horizontal axis distance, with left-most point start point of route, right-most point end point of route.

Example:
![image](https://www.topo-gps.com/android/en/manual/html/_images/route-profile3.png)
_(Source: https://www.topo-gps.com)_

Displaying this info should be on-demand, not permanent.
This could be either a split view as shown above or a separate view.
(If we go for a split view we could use a draggable marker on the elevation chart to highlight the corresponding point on the map.)

### Is there an alternative?

don't add such a function

### Additional Information

BRouter already gives us elevation info together with routing data, we just don't store it yet.

fm-sys:
+1 but please don't integrate it in our old map. This could be shown when clicking on a track line in unified map.

moving-bits:
I'm looking for a lightweight lib for drawing such point charts, but that seems to be not that easy... There are some powerful libs, but adding quite a bit weight to the app, and only few others. https://github.com/Madrapps/plot seems to be such a lightweight lib, mainly consisting of four small files. But it's using Kotlin & Composables (as most of the other libs do as well). While neither Kotlin nor Composables should be a problem in itself, I have not been able yet to get it to run in context of c:geo. I can add a Composable view to an XML layout, but I don't know how to use `setContent` for this view, meaning: how to address the Composable function for it. I've tried staying with a Kotlin file for the method defining the Composable content, but our build run seems to ignore it. Any ideas?

devemux86:
- https://github.com/PhilJay/MPAndroidChart (used by OsmAnd?)
- https://github.com/jjoe64/GraphView

moving-bits:
Thanks, additionally I had YCharts: https://github.com/codeandtheory/YCharts

Both GraphView and MPAndroidChart seem to be wide-spread, but both are no unmaintained for four or five of years already, sadly, having > 100 open pull requests. Additionally. https://github.com/jjoe64/GraphView/issues/691 made me sceptical - don't want to blow up our apk size by four meg just for a simple chart.

Plot had no updates for two years as well, but is small enough to fully-integrate it in our code, as it basically consists of just four small files. It seems lightweight, supporting only line charts, therefore I started with that, but I may eventually look into the others as well.

[...]

PR #14776, opened 2023-10-29
[unified map] fix caches not filtered correctly if filter is complex

fm-sys:


fm-sys:
This was a bug which caused caches to show up, if they couldn't be filtered by the connector but only on client side... Visible e.g. in https://github.com/cgeo/cgeo/issues/14768#issuecomment-1784139717

PR #14778, opened 2023-10-29
Simple Dialog Advanced version

eddiemuc:
Simple Dialog Advanced version


eddiemuc:
This PR will redesign and simplify SimpleDialog usage as well as unify single and multiple choice lists. It is a very large PR because several new model classes were created (getter/setter logic takes a lot of lines...) and existing usages of SimpleDialog needed to be adapted.

The new structure is meant to allow for easy inclusion of features such as filtering/grouping into all places in c:geo, it is also meant to enforce a similar look & feel across all Simple-Dialog usages. I am pretty sure that fine-tuning will be necessary GUI-wise, but I believe this PR is ready for merge.

Some visual examples:

List Selection with list marker display:
image

Filtering also in single-select:
image

Grouping can be expanded/reduced:
image

Icons in Log Type selection:
image

Item Actions now integrated (was separate before):
image

PR #14779, opened 2023-10-29
UnifiedMap: Elevation chart for individual route (rel. to 14768)

moving-bits:
## Description
First draft of elevation chart, implemented for `UnifiedMap`.

I switched to AppDevNext/AndroidChart, as discussed in [#14768](https://github.com/cgeo/cgeo/issues/14768#issuecomment-1784198635).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -26,6 +26,7 @@ import cgeo.geocaching.maps.routing.Routing;
 import cgeo.geocaching.maps.routing.RoutingMode;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.RouteItem;
+import cgeo.geocaching.models.RouteOrRouteItem;
 import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.sensors.GeoData;
 import cgeo.geocaching.sensors.GeoDirHandler;
@@ -829,7 +830,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
     public void onTap(final int latitudeE6, final int longitudeE6, final int x, final int y, final boolean isLongTap) {
 
         // lookup elements touched by this
-        final LinkedList<RouteItem> result = new LinkedList<>();
+        final LinkedList<RouteOrRouteItem> result = new LinkedList<>();
 
         for (String key : clickableItemsLayer.getTouched(Geopoint.forE6(latitudeE6, longitudeE6))) {
 
@@ -838,7 +839,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
 
                 final Geocache temp = DataStore.loadCache(geocode, LoadFlags.LOAD_CACHE_OR_DB);
                 if (temp != null && temp.getCoords() != null) {
-                    result.add(new RouteItem(temp));
+                    result.add(new RouteOrRouteItem(new RouteItem(temp)));
                 }
             }
 
@@ -847,7 +848,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
 
                 for (RouteItem item : viewModel.individualRoute.getValue().getRouteItems()) {
                     if (identifier.equals(item.getIdentifier())) {
-                        result.add(item);
+                        result.add(new RouteOrRouteItem(item));
                         break;
                     }
                 }
@@ -858,15 +859,14 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
 
                 for (Waypoint waypoint : viewModel.waypoints.getValue()) {
                     if (fullGpxId.equals(waypoint.getFullGpxId())) {
-                        result.add(new RouteItem(waypoint));
+                        result.add(new RouteOrRouteItem(new RouteItem(waypoint)));
                         break;
                     }
                 }
             }
 
             if (key.startsWith(IndividualRouteLayer.KEY_INDIVIDUAL_ROUTE) && !isLongTap) {
-                toggleElevationChart(this, viewModel.individualRoute.getValue(), nonClickableItemsLayer, getString(R.string.individual_route));
-                return;
+                result.add(new RouteOrRouteItem(viewModel.individualRoute.getValue()));
             }
 
         }
@@ -887,14 +887,14 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
             handleTap(result.get(0), isLongTap, x, y);
         } else {
             try {
-                final ArrayList<RouteItem> sorted = new ArrayList<>(result);
-                Collections.sort(sorted, RouteItem.NAME_COMPARATOR);
+                final ArrayList<RouteOrRouteItem> sorted = new ArrayList<>(result);
+                Collections.sort(sorted, RouteOrRouteItem.NAME_COMPARATOR);
 
-                final ArrayAdapter<RouteItem> adapter = new ArrayAdapter<RouteItem>(this, R.layout.cacheslist_item_select, sorted) {
+                final ArrayAdapter<RouteOrRouteItem> adapter = new ArrayAdapter<RouteOrRouteItem>(this, R.layout.cacheslist_item_select, sorted) {
                     @NonNull
                     @Override
                     public View getView(final int position, final View convertView, @NonNull final ViewGroup parent) {
-                        return GeoItemSelectorUtils.createRouteItemView(UnifiedMapActivity.this, getItem(position),
+                        return GeoItemSelectorUtils.createRouteOrRouteItemView(UnifiedMapActivity.this, getItem(position),
                                 GeoItemSelectorUtils.getOrCreateView(UnifiedMapActivity.this, convertView, parent));
                     }
                 };
@@ -916,27 +916,30 @@ public class UnifiedMapActivity extends AbstractNavigationBarActivity implements
 
     }
 
-    private void handleTap(final RouteItem item, final boolean isLongTap, final int tapX, final int tapY) {
+    private void handleTap(final RouteOrRouteItem item, final boolean isLongTap, final int tapX, final int tapY) {
+        final RouteItem routeItem = item.isRouteItem() ? item.getRouteItem() : null;
         if (isLongTap) {
             // toggle route item
-            if (Settings.isLongTapOnMapActivated()) {
+            if (routeItem != null && Settings.isLongTapOnMapActivated()) {
                 if (Settings.isShowRouteMenu()) {
-                    MapUtils.createCacheWaypointLongClickPopupMenu(this, item, tapX, tapY, viewModel.individualRoute.getValue(), viewModel, null)
+                    MapUtils.createCacheWaypointLongClickPopupMenu(this, routeItem, tapX, tapY, viewModel.individualRoute.getValue(), viewModel, null)
 //                            .setOnDismissListener(menu -> tapHandlerLayer.resetLongTapLatLong())
                             .show();
                 } else {
-                    viewModel.toggleRouteItem(this, item);
+                    viewModel.toggleRouteItem(this, routeItem);
                 }
             }
-        } else {
+        } else if (routeItem != null) {
             // open popup for element
-            if (item.getType() == RouteItem.RouteItemType.GEOCACHE) {
+            if (routeItem.getType() == RouteItem.RouteItemType.GEOCACHE) {
                 // @todo: do we need a DataStore.loadCache() before?
-                CachePopup.startActivityAllowTarget(this, item.getGeocode());
-            } else if (item.getType() == RouteItem.RouteItemType.WAYPOINT && item.getWaypointId() != 0) {
+                CachePopup.startActivityAllowTarget(this, routeItem.getGeocode());
+            } else if (routeItem.getType() == RouteItem.RouteItemType.WAYPOINT && routeItem.getWaypointId() != 0) {
                 // @todo: do we need a DataStore.loadWaypoint() before?
-                WaypointPopup.startActivityAllowTarget(this, item.getWaypointId(), item.getGeocode());
+                WaypointPopup.startActivityAllowTarget(this, routeItem.getWaypointId(), routeItem.getGeocode());
             }
+        } else {
+            toggleElevationChart(this, item.getRoute(), nonClickableItemsLayer, getString(R.string.individual_route));
         }
     }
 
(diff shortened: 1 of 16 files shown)


moving-bits:
retest this please

moving-bits:
what's going on here on the CI?

```
[...]
23:18:44  /home/jenkins/slave/workspace/cgeo-CI_PR-build/main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java:21: error: cannot access LineChart
23:18:44  import com.github.mikephil.charting.charts.LineChart;
23:18:44                                            ^
23:18:44    bad class file: /home/jenkins/.gradle/caches/transforms-3/241bf7b3dbdc53264fb46631469fd427/transformed/AndroidChart-3.1.0.17-api.jar(/com/github/mikephil/charting/charts/LineChart.class)
23:18:44      class file has wrong version 61.0, should be 55.0
23:18:44      Please remove or make sure it appears in the correct subdirectory of the classpath.
23:18:44  /home/jenkins/slave/workspace/cgeo-CI_PR-build/main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java:22: error: cannot access Legend
[...]
```

(builds fine in my local dev environment)

eddiemuc:
Looks like our CI uses Java 11 (max supported class version 55.0) whereas the library was compiled and needs Java 17 (support for class version 61.0).

See https://docs.oracle.com/javase/specs/jvms/se20/html/jvms-4.html#jvms-4.1-200-B.2

eddiemuc:
Looks like our CI uses Java 11 (max supported class version 55.0) whereas the library was compiled and needs Java 17 (support for class version 61.0).

See https://docs.oracle.com/javase/specs/jvms/se20/html/jvms-4.html#jvms-4.1-200-B.2

Also I thought we can only use libraries compiled against max. Java 8, so that they can run on older Android versions down to 21. But I am not really sure in this.

moving-bits:
Thanks for checking. I'll do a cross-check against API 21 emulator, but it looks like we will have to update our CI sooner or later...

bekuno:
Lets have a look here: https://developer.android.com/build/jdks#toolchain

bekuno:
Please check, if AndroidChart [3.1.0.16](https://github.com/AppDevNext/AndroidChart/releases/tag/3.1.0.16) works.
In version 3.1.17 they changed the kotlin version from 1.8.x to 1.9.0.

Would be very interesting, as I did not find something relevants at https://kotlinlang.org/docs/whatsnew19.html.

moving-bits:
I've made a downgrade to 3.1.0.16, let's see what our CI thinks of it

moving-bits:
@bekuno 
This time the build is green. Not sure if we have different executor configurations, and this time just were lucky to get the one supporting Java 17, or if downgrading to 3.1.0.16 was the actual key to success in this case.

bekuno:
> @bekuno This time the build is green. Not sure if we have different executor configurations, and this time just were lucky to get the one supporting Java 17, or if downgrading to 3.1.0.16 was the actual key to success in this case.

Should be the second case (fingers crossed).
Looks similar to the compileSdk 34 error such as for the Material Lib.

moving-bits:
@eddiemuc 
I could use a pointer from you regarding GeoItemLayer:
I'm trying to connect taps on elevation chart with drawing a circle on the map.

When I'm using a `GeoPrimitive.createCircle(...)` it does work, but I don't know how to scale the circle's radius according to map zoom, but when I'm using a drawable resource instead (using `GeoPrimitive.createMarker(...)`), the marker does not show up on the map, though - seems that something is still missing in my code - but what?

See `ElevationChartUtils` around line 136.

Thanks.

eddiemuc:
> @eddiemuc I could use a pointer from you regarding GeoItemLayer: I'm trying to connect taps on elevation chart with drawing a circle on the map.
> 
> When I'm using a `GeoPrimitive.createCircle(...)` it does work, but I don't know how to scale the circle's radius according to map zoom, but when I'm using a drawable resource instead (using `GeoPrimitive.createMarker(...)`), the marker does not show up on the map, though - seems that something is still missing in my code - but what?
> 
> See `ElevationChartUtils` around line 136.
> 
> Thanks.

You are the first one really using this layer except for my own test layer GeoItemTestLayer stuff. So it is close to certain that there are still hidden bugs in the geoitemlayer-implementation.

Like all other primitives, the  circle produced by `GeoPrimitive.createCircle(...)` uses real-world measurements. Its radius is thus in meters/kilometers, not pixels/dps.  It is fully intentional that it "grows"/"shrinks" when zooming in or out. Anyway, I believe your use case is a marker use case anyway.

Your marker code looks fine to me at first glance. I would have to take a deeper look. My first question would be: is the bitmap passed to the marker really valid? (I assume that the circle marker you used is a vector graphic, and I believe I have not tried to convert that to a bitmap yet). But there might also be a bug in the geoitemlayer cide.

fm-sys:
Whole current state of unified map is completely based on GeoItemLayer, and I used markers a lot there. (also vector drawables)

main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java
```diff
@@ -0,0 +1,153 @@
...
+                                center = point;
+                                break;
+                            }
+                            lastPoint = point;
+                        }
+                    }
+                    // update marker if position found
+                    if (center != null) {
+// icon marker isn't drawn
+                        final GeoItem marker = GeoPrimitive.createMarker(center, GeoIcon.builder().setBitmap(ImageUtils.convertToBitmap(ResourcesCompat.getDrawable(CgeoApplication.getInstance().getResources(), R.drawable.circle, null))).build());
```
fm-sys, line 136: You may want to add `.buildUpon().setZLevel(LayerHelper.ZINDEX_XYZ).build()` to specify a ZLevel. Don't know whether this makes a difference. At least I used it nearly everywhere.

You can also take a look at `CoordsIndicatorLayer` which does pretty much the same as you need here.

moving-bits:
Thanks @eddiemuc & @fm-sys for looking into this.

> You are the first one really using this layer except for my own test layer GeoItemTestLayer stuff.

That's where I copied the code from :-)

> Like all other primitives, the circle produced by `GeoPrimitive.createCircle(...)` uses real-world measurements. Its radius is thus in meters/kilometers, not pixels/dps. It is fully intentional that it "grows"/"shrinks" when zooming in or out. Anyway, I believe your use case is a marker use case anyway.

Yes, think so. I only switched to circles remporarly when my marker didn't show up on the map, to cross-check whether it's some error in the logic or a display problem.

> My first question would be: is the bitmap passed to the marker really valid?

The drawable does exist (is a vector drawable), and I had tried two different one to be on the safe side. I have not checked the result from creating the bitmap yet, but at least it does not throw a NPE.

> You may want to add `.buildUpon().setZLevel(LayerHelper.ZINDEX_XYZ).build()` to specify a ZLevel. Don't know whether this makes a difference. At least I used it nearly everywhere.

Might very well be, I'll test it.

> You can also take a look at `CoordsIndicatorLayer` which does pretty much the same as you need here.

Ah, that's sort of an example I was looking for, thanks. Will try to dive into it over the weekend.

moving-bits:
> On Gmaps, it was working fine without the need of any adoptions. On VTM the marker only appears as soon as you move the map. (similar emulator problem as before? Probably we should investigate this...)

Maybe it (either geo object or map) needs some sort of invalidation notice?

> Furthermore: may I suggest adding a closing button to the panel?

Sure, good idea. 
But where to place it? Not all tracks will end with lower elevation than they started... 🤷  
Maybe a gesture or right-click action? Or add some extra space on top?

fm-sys:
> Sure, good idea.
> But where to place it? Not all tracks will end with lower elevation than they started... 🤷
> Maybe a gesture or right-click action? Or add some extra space on top?

As I would vote to generalize this to gpx imported tracks as well, we could show the track name on top. Then we would have space for the "X".
 But I think it's not even a problem to let it overlap with the chart. Just make the button semi transparent as all buttons on our map and everything is fine IMHO.

fm-sys:
As it works on all other map types, it's probably a bug in the `MapsforgeVtmGeoItemLayer` implementation. I've opened #14787 to track this problem, as this problem is not limited to this PR. 

@eddiemuc  I tried to investigate, but do not really understand the inner workings. Maybe you could dive into it?

eddiemuc:
I will look into it

main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java
moving-bits: I've added ZLayer anyway to keep the marker on top

moving-bits:
> Furthermore: may I suggest adding a closing button to the panel?

I've added a close button at top left-position, similar to bottom sheet dialogs. Title is added as well.
Instead of toggling the chart using a menu item you can now tap on the individual route in the map.

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -862,6 +864,11 @@ public void onTap(final int latitudeE6, final int longitudeE6, final int x, fina
                 }
             }
 
+            if (key.startsWith(IndividualRouteLayer.KEY_INDIVIDUAL_ROUTE) && !isLongTap) {
+                toggleElevationChart(this, viewModel.individualRoute.getValue(), nonClickableItemsLayer, getString(R.string.individual_route));
+                return;
+            }
```
fm-sys, line 870: This will cause that caches aren't clickable anymore if the individual route is below them. 
- So either change the `LinkedList<RouteItem> result` to some super type which can hold more than only route items. This would have the advantage that tracks would appear in the "clicked items popup".
- Or explicitly search for key `IndividualRouteLayer.KEY_INDIVIDUAL_ROUTE` in case `result.size() == 0`

fm-sys:
grafik

Although this is probably the most compact variant, it's IMHO better to have a consistent style. Therefore, I would vote to mimic the style of map quick settings and track / route popup.

main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    public static void toggleElevationChart(@NonNull final AppCompatActivity activity, final Route route, final GeoItemLayer<String> geoItemLayer, final String title) {
+        final View chartBlock = activity.findViewById(R.id.elevation_block);
+        final LineChart chart = activity.findViewById(R.id.elevation_chart);
+        if (chart == null) {
+            return;
+        }
+        if (chartBlock.getVisibility() == View.VISIBLE || route == null) {
+            chartBlock.setVisibility(View.GONE);
```
fm-sys, line 55: Just discovered another bug, you should also remove the `ELEVATIONCHART_MARKER` when hiding the elevation chart

fm-sys:
And: you need to observe individual route changes and redraw the chart + remove marker (always/if not on route anymore).

Sorry for cluttering this thread. I found the bugs at different times and always thought I found everything now.

moving-bits:
> Although this is probably the most compact variant, it's IMHO better to have a consistent style. Therefore, I would vote to mimic the style of map quick settings and track / route popup.

Is there some Material element that looks like a bottom sheet dialog but is non-modal? Or do we have to create this on our own? (Shouldn't be too hard, but no need to duplicate code if such a thing already exists.)

> Sorry for cluttering this thread. I found the bugs at different times and always thought I found everything now.

No worries. Thanks for taking the time to test this new feature!

main/src/main/java/cgeo/geocaching/unifiedmap/layers/ElevationChartUtils.java
moving-bits: done

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
moving-bits: done (by creating a holder class `RouteOrRouteItem`)

fm-sys:
(Test is failing, cause I think you forgot to add the `RouteOrRouteItem` class to git.)

> Is there some Material element that looks like a bottom sheet dialog but is non-modal? Or do we have to create this on our own? (Shouldn't be too hard, but no need to duplicate code if such a thing already exists.)

Yes, there are non-modal bottom sheets, e.g. most prominently used in Google Maps. But this is only about the swipe away logic, not the design layout. Need to look up how to use them though.

Actually, our bottom sheet dialogs are self designed as well. You can take a look at `Dialogs.bottomSheetDialogWithActionbar(...)` as reference

moving-bits:
Switched to `Toolbar` for the elevation chart:

### portrait mode:
image

### landscape mode:
Interesting things happen here. Per default, the elevation chart's toolbar is hidden behind the map's toolbar, and only becomes visible on tapping on the map:
image
image
I could probably make the toolbar become visible all the time even in landscape mode, but then it would reside below the map's toolbar - and I'm unsure about the "nested toolbars" look then.

fm-sys:
> Interesting things happen here. Per default, the elevation chart's toolbar is hidden behind the map's toolbar, and only becomes visible on tapping on the map

I'm almost sure this is a side effect of the hide toolbar feature. Could it be that the Toolbar is rendered below the action bar?

fm-sys:
Next step could be to make all controls of the track / route popup available as well. 

This would be great in terms of usability, especially if it also would be available for gpx tracks. Just click on a track and have all controls directly available. 

Sorry I'm requesting way too much. I'm totally fine with moving this into an own feature request :) 

Screenshot_20231104_100146_cgeo geocaching

moving-bits:
> > Interesting things happen here. Per default, the elevation chart's toolbar is hidden behind the map's toolbar, and only becomes visible on tapping on the map
> 
> I'm almost sure this is a side effect of the hide toolbar feature. Could it be that the Toolbar is rendered below the action bar?

yes & yes. You can see the rendered toolbar in the second landscape screenshot

moving-bits:
> Next step could be to make all controls of the track / route popup available as well.

Let's put that in a separate issue. This one's already getting too large...

Also we should discuss how to access those controls (I suggest long tap), and which of the controls to show.

moving-bits:
> And: you need to observe individual route changes and redraw the chart + remove marker (always/if not on route anymore).

Done

main/src/main/res/layout-land/unifiedmap_activity.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
```
bekuno, line 4: Please have a look at the CI lint result, eg. at https://ci.cgeo.org/job/cgeo-CI_commit-build/1657/androidLintResult/new/.
Also the `xmlns:tools` here is unnecessary.

moving-bits: all three fixed with PR #14817

PR #14781, opened 2023-10-30
make "toggle talking" to a real toggle-button

### LINKED ISSUE #9127: Make "toggle talking" to a real toggle-button
Not really. Just consistency in UI.

The menu-entry "toggle talking" should be a real toggle button like "Prevent waypoint from note". There the text is changing due to the current setting.
So the two states could be
- Start talking / Start speech
- Stop talking / Stop speech

Implementation like the other toggle-button. New icon is necessary (could be easily created from original icon and "main_nearby_disabled.png"

**Additional context**
grafik

grafik

murggel:
## Description

Change "toggle talking" to a "real" toggle-button with 2 states:
"Start talking" 
"Stop talking" 

represented by following icons
image

## Related issues

fix #9127

## Additional context

Current text: "Toggle talking", therefore I used "Start talking" and "Stop talking", the balloon on changing speech-service states "Started speech" and "Stopped speech". But can be discussed...


murggel:
Current implementation doubles `setMenuToggleSpeech(final MenuItem menuItem, final boolean speechActive)`

Where can I put it to avoid duplicated code? Didn't find a nice place. Didn't want to create a new "utility"-class?

eddiemuc:
> Current implementation doubles `setMenuToggleSpeech(final MenuItem menuItem, final boolean speechActive)`
> 
> Where can I put it to avoid duplicated code? Didn't find a nice place. Didn't want to create a new "utility"-class?

I suggest `ActivityMixin` (which seems to be kind of an Utila class for Activity-related util methods).

 In theory you could also put it into `AbstractActivity`. However, since I am really no friend of putting util methods into base classes of inheritance hierarchies, my vote goes against this.

murggel:
> > Current implementation doubles `setMenuToggleSpeech(final MenuItem menuItem, final boolean speechActive)`
> > 
> > Where can I put it to avoid duplicated code? Didn't find a nice place. Didn't want to create a new "utility"-class?
> 
> I suggest `ActivityMixin` (which seems to be kind of an Utila class for Activity-related util methods).
> 
>  In theory you could also put it into `AbstractActivity`. However, since I am really no friend of putting util methods into base classes of inheritance hierarchies, my vote goes against this.
> 
> 


Thanks for the hint.
While moving, an idea for a generic toggle-helper class come to my mind, which encapsulate the toggle-logic. As static member I made the quadruple of icon and text for toggle-speech. So it can be used for other toggle-items as well.

@eddiemuc What do you think about it?

eddiemuc:
Like it!

main/src/main/java/cgeo/geocaching/ui/ToggleItem.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Encapsulates the mechanism to toggle text and image.
+ *
+ * Supports this for menu-item.
+ *
+ * Class is supposed to be used in Activity to toggle ui-elements.
+ */
+public class ToggleItem {
+    public static ToggleItem followMyLocation = new ToggleItem(R.drawable.ic_menu_mylocation, R.drawable.ic_menu_mylocation_off, -1, -1);
+    public static ToggleItem liveMode = new ToggleItem(R.drawable.ic_menu_sync_enabled, R.drawable.ic_menu_sync_disabled, R.string.map_live_disable, R.string.map_live_enable);
```
eddiemuc, line 20: I suggest creating ToggleItem instances not here centrally but in the activity classes where they are used.

Or is the same instance of ToggleItem used in multiple places and ToggleItem is kind of a "ToggleItemType" class? In that case your choice would be correct imho. I suggest then to convert the class to an enum.

murggel: > I suggest creating ToggleItem instances not here centrally but in the activity classes where they are used.
> 
> Or is the same instance of ToggleItem used in multiple places and ToggleItem is kind of a "ToggleItemType" class? In that case your choice would be correct imho. I suggest then to convert the class to an enum.

The latter was my intention, so the user of the class does not have to think about which string and icon to use.
I will convert it to an enum.

PR #14794, opened 2023-11-04
UnifiedMap: Enable tracks for elevation chart

moving-bits:
## Description
Enable tracks for elevation chart:
- make tracks clickable
- add prefix to their key to make them identifiable
- allow elevation chart to get completely new data
- show "no elevation data available" in the chart if no elevation data is available

@fm-sys 
One thing is missing, though: How can I observe changes to the track? That may happen if someone changes the routing mode.
(`UnifiedMapViewModel.setTrack` is called in this case, but how to trigger an observer for this, delivering the key of the related track?)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -25,6 +25,7 @@ import cgeo.geocaching.maps.RouteTrackUtils;
 import cgeo.geocaching.maps.routing.Routing;
 import cgeo.geocaching.maps.routing.RoutingMode;
 import cgeo.geocaching.models.Geocache;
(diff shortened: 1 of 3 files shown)


fm-sys:
There is a `trackUpdater` inside the view model which can be observed. This code snipped should work:

```
viewModel.trackUpdater.observe(activity, event -> {
    String key = event.peek();
    ... // check if track key matches the currently displayed track, if yes refresh
```

Off Topic: Note that it is by far not yet decided whether this is the optimal solution at the end for this specific scenario. I'm still playing around with different solutions for observer based code...

moving-bits:
> There is a `trackUpdater` inside the view model which can be observed. This code snipped should work:

Thanks for this pointer. I had seen the `trackUpdater` but did not recognize how to extract the relevant key. Works like a charm now.

PR #14797, opened 2023-11-05
[cache/WP popup] show as bottom sheet dialog

fm-sys:
**Feedback required!** What do you think about showing the cache and WP popup as bottom sheet?

Popup will close when clicking outside. I'ts (currently) not like on Google Maps where you can still move the map.

|Cache|WP|
-|-
grafik|grafik


ztNFny:
showing hint currently resizes the popup downwards - would in this implementation the resize happen upwards?

moving-bits:
> **Feedback required!** What do you think about showing the cache and WP popup as bottom sheet?

why not? looks ok, and gives us the chance for some space saving. But having it non-modal would be essential, IMHO.

For making the bottom sheet a bit more compact, one cut remove the "More Details" button and place it as small button in the toolbar, same with "Edit" button for waypoints.

ztNFny:
> one cut remove the \"More Details\" button and place it as small button in the toolbar

That's the most important button in that dialog, if anything it should be bigger (no not really, but please don't make it any harder to open the CacheDetails). CachePopup is just an intermediate step in accessing Details.

moving-bits:
> Sugestions:
> * "Open cache details" button should always be in the same place (right now it place differs between cache, cache with hint shown, waypoint)

One could place it either at the bottom (if we want to stay with a full-width button) or place it as first button into the toolbar (which I would prefer).

> * The refresh/list info/buttons use a lot of vertical space while leaving most of that space empty. They could at least be combined into 1 line (same goes for the "watch" button in CacheDetails activity)

Generally agreed, but we would need to take care of the actual list info, especially if more than one list is assigned.

> * For waypoint dialog maybe we can leave out "most" of the cache info? Waypoints are most often selected while playing a cache, no point in seeing info like rating, ... in that case?

Agreed. Probably name & type (+ maybe size) should be enough.

> * For waypoint dialog I'd switch cache and waypoint info. You're most likely interested in the WP, so the "Edit" button should be near where all the buttons usually are -> bottom

If we reduce cache info to name, type, size, only a single info line remains, which could be combined with the waypoint's info lines. Then the "edit" button would move close to the bottom automatically. Would look likes this (if we want to keep the full-width button):

image

(Just think of having quite a bit of map space regained.)

ztNFny:
> place it as first button into the toolbar (which I would prefer).

again, horrible idea imho - see https://github.com/cgeo/cgeo/pull/14797#issuecomment-1793841170

> > * The refresh/list info/buttons use a lot of vertical space while leaving most of that space empty. They could at least be combined into 1 line (same goes for the "watch" button in CacheDetails activity)
> 
> Generally agreed, but we would need to take care of the actual list info, especially if more than one list is assigned.

right now those 2 buttons take up the height of 6 lines of text, so even with only 1 button there'd still be 3 lines. More than enough for a cache assigned to maany lists

moving-bits:
I've added to this PR to make the bottom sheets non-modal, so that the map in background can still be used. Swiping down the bottom sheet closes it. Tapping on a different cache/waypoint replaces the view. We should discuss whether to add the small "close" symbol we have in other bottom sheets as well.

image

Implemented for all three map types. `CachePopup` / `WaypointPopup` are no longer needed, opening the details view is handled by `MapUtils.showCacheDetails()` and `MapUtils.showWaypointDetails()` respectively.

fm-sys:
I already had forgotten this PR... Looks awesome, thanks for your work!

Some things through:
- toolbar actions + overflow menu should be readded
- clicking outside the bottom sheet should close it (if no map item was hit)
- clicking back should close bottom sheet instead of map activity if one is open

Then we don't need a separate close button IMHO.

moving-bits:
Some quick fixes to waypoint details view:

- leave out most of cache details info, add a compressed first line with cache info (name, type, GC code)
- move "edit" button to bottom (swap with "More details" button, which is renamed to "Cache")

image

moving-bits:
> * toolbar actions + overflow menu should be readded

agreed

> * clicking outside the bottom sheet should close it (if no map item was hit)

I'm undecided about this: Being able to move the map around while still seeing details view is useful, IMHO.

> * clicking back should close bottom sheet instead of map activity if one is open

agreed

I'll give those two a try.

fm-sys:
>> clicking outside the bottom sheet should close it (if no map item was hit)
>
> I'm undecided about this: Being able to move the map around while still seeing details view is useful, IMHO.

Totally agree. But we should differentiate the touch events between click and move. (Google Maps has the same behavior BTW)

I can take care of that one if you like.

---

BTW, when working on shared branches, please don't use force pushes as this may accidentally overwrite changes or break the git history. It's better to flood the branch with commits and squash on merge instead.

fm-sys:
Also we should show it as side sheet in landscape mode

|||
-|-
grafik|Screenshot_20231119_123544_com google android apps maps

moving-bits:
> > > clicking outside the bottom sheet should close it (if no map item was hit)
> > 
> > I'm undecided about this: Being able to move the map around while still seeing details view is useful, IMHO.
> 
> Totally agree. But we should differentiate the touch events between click and move. (Google Maps has the same behavior BTW)
> 
> I can take care of that one if you like.

Sure, please go ahead. It should be triggered on "short tap on empty map space" only then. You may want to call the newly introduced `MapUtils.removeDetailsFragment()` for this.

fm-sys:
> > > > clicking outside the bottom sheet should close it (if no map item was hit)
> > > 
> > > 
> > > I'm undecided about this: Being able to move the map around while still seeing details view is useful, IMHO.
> > 
> > 
> > Totally agree. But we should differentiate the touch events between click and move. (Google Maps has the same behavior BTW)
> > I can take care of that one if you like.
> 
> Sure, please go ahead. It should be triggered on "short tap on empty map space" only then. You may want to call the newly introduced `MapUtils.removeDetailsFragment()` for this.

Thanks! 

long tap on empty map space is already reserved for other actions, so yes...it'll be short tap only :)

Lineflyer:
Just my few cents:
I totally like the idea and the screenshots.
I would have had the same comments as others here:
- Tap outside should close 
- Landscape should have a side-by-side approach

I am undecided about the "lost" functionality represented by the icons before (store, refresh, etc.). 
Is that the function behind the "edit" button?

Anyway we can still optimize the solution once a first approach has been merged and tested in the nightlies.

moving-bits:
> * clicking back should close bottom sheet instead of map activity if one is open

back key is handled properly now

> * toolbar actions + overflow menu should be readded

where did those get injected in `CachePopup` / `WaypointPopup`? Couldn't find them on doing a quick scan.

> Also we should show it as side sheet in landscape mode

Yes, but I'd propose to show the details view on the right, map on the left.

> BTW, when working on shared branches, please don't use force pushes as this may accidentally overwrite changes or break the git history. It's better to flood the branch with commits and squash on merge instead.

Sure. I was just fixing some Codacy errors, and thinking that I'm the only one working on that branch I wanted to keep the history clean. But I'll stay away from push-forcing this branch from now on.

Let's decide in today's meeting whether we want to merge this PR today for easier collaboration without squashing the large change into one big commit.

moving-bits:
> I am undecided about the "lost" functionality represented by the icons before (store, refresh, etc.). Is that the function behind the "edit" button?

For waypoints it's something like "start navigation", "toggle talk", "set as found", all of which would be available after hitting the "Cache" button. But at least for some of them, direct accessibility would be helpful.

They got "lost" for now by switching from `WaypointPopup` (which is an activity that opens the `WaypointPopupFragment`) to directly showing that fragment. I'm currently searching for a way to restore the context menu.

Cache popup will have the same problem, as it got changed in the same way, leaving out the `CachePopup` in between.
Layout of the `CachePopupFragment` did not get optimized in any way yet. Some ideas are in the discussion above, but none yet implemented. I started with the easy part ;-)

fm-sys:
> Yes, but I'd propose to show the details view on the right, map on the left.

fine for me

> where did those get injected in CachePopup / WaypointPopup? Couldn't find them on doing a quick scan.

The popup toolbar was set as activity actionbar view and then was filled "the classic way" via onCreateOptionsMenu etc. This need to be refactored when using the fragment approach...

> Sure. I was just fixing some Codacy errors, and thinking that I'm the only one working on that branch I wanted to keep the history clean. But I'll stay away from push-forcing this branch from now on.

I was indeed not working on that branch, so everything is fine. I got massive merge conflicts while pulling your changes though, but just did a `git reset --hard` locally to fix this.

fm-sys:
This is the relevant actionbar code:

https://github.com/cgeo/cgeo/blob/9222ecdc8ba2a2617796a00ccc8ea7e017675ec0/main/src/main/java/cgeo/geocaching/AbstractDialogFragment.java#L180-L204

fm-sys:
Suggestion for making the bottom sheet more compact without feature loss:

(this would make the "more details" button harder to reach, so a swipe up gesture as replacement would be good.)

|||
-|-
grafik|grafik

ztNFny:
I've been thinking for quite a while that 2 rows for D/T seems quite the waste ... Maybe that could be adressedy at least in popup?

fm-sys:
> I've been thinking for quite a while that 2 rows for D/T seems quite the waste ... Maybe that could be adressedy at least in popup?

While I like the space expensive star layout in the cache details, a `D 1.5 • T 1.5` would be sufficient for cache details...

Personally I wouldn't add the additional DT marker, as it's already shown in the cache icon anyway (if enabled).

moving-bits:
> Suggestion for making the bottom sheet more compact without feature loss:
> 
> (this would make the "more details" button harder to reach, so a swipe up gesture as replacement would be good.)

I like the idea of combining some of the buttons and separating them from the details text ("stored on this device..." and "list x, y, z". But I would prefer to stay with c:geo's button layout instead of the chips-like.

We can have up to five buttons:
- More details (should be the most prominent)
- hint
- refresh
- edit lists
- delete or store

Proposal:
- First row: "More details" - "Hint" (only if needed) - "Refresh" (only if stored already)
- Second row: Details info for age & lists - "Edit lists" - "Store" or "Delete"

"More details" would be a textual button, the other icons.

The text for "Stored on this device xx ago" could be shortened to "xx ago" or "stored xx ago" to fit into one line, lists would go below this info.

fm-sys:
The reason why I'm tending to use chips is that they are more typically expected to be scrollable, which would allow us to place them all into a single row.

I would order them:
- more details
- store / edit list
- delete
- refresh
- hint

Making "more details" a text button and the rest icon only should take an amount of space which should even fit on most devices, so scrolling will rarely be needed - and even if - will only affect IMHO rather unimportant items.

I think we can discuss these things this evening.

moving-bits:
Landscape mode is implemented:
image

Toolbar menus are back as well, but not all menu entries are working yet (eg: "sharing") - need to investigate further.

moving-bits:
Playing around a bit with layout options. Here's another proposal for discussion this evening:

|before|after|
|---|---|
|image|image|

Move refresh / store / delete buttons a bit up into the upper block, right aligned as the other icon buttons are. "More details" and "hint" stay where they are today, and "List edit" moves one place up + to the right. Textual info below ("stored ago" and lists) get more space.

Or one can swap the last two blocks, moving "More details" and "hint" buttons to the bottom to have a fixed space for them. "Stored ago", list info and "List edit" button would move one up, then. Could look like this:

|before|after|
|---|---|
|image|image|

> OT: Not saying that googles UI always has to be good, but they brought scrolling to a whole new level ;-)

Yeah, I know. And I would really like to avoid additional horizontal scrolling just for some buttons that should be quite accessible. Otherwise we could just leave them out, as they are reachable using the "More details" button anyway... (no, don't want that)

ztNFny:
I like the chip-sized buttons, but if they could look like any other cgeo button that'd be better.
That and remove 1 line by combining DT should give us enough space?

fm-sys:
While @moving-bits was playing with the layout, I played with transitions a bit...

We might even want to leave out the "more details" button completely:

https://github.com/cgeo/cgeo/assets/64581222/fa2c01ce-1cb9-4a3f-b427-0d42bce8e50a

moving-bits:
Proposal for compressing some of the detail info given in cache "popup":

Combine D/T and FP info into a single line:

image

ztNFny:
> Proposal for compressing some of the detail info given in cache "popup":
> 
> Combine D/T and FP info into a single line:
> 
> image
> 

I wouldn't put DT and FP together, it's too unrelated IMHO. If we want to reduce further we could combine DT and size, this is all info you check to determine where a cache could be hidden.

That said, I've been thinking about introducing a size icon (in addition to the text) for a while - an icon is faster to recognize than a text. Just didn't get around to it yet. Maybe that's an idea you could discuss in todays call as well?

fm-sys:
Update: After pushing several fixes, this PR should now work without mayor bugs. Therefore, let's merge as discussed in the team meeting and continue with layout optimization on master...

fm-sys:
@moving-bits the action bars do appear indeed below each other also I was not able to reproduce yesterday. But only if action bar hiding is disabled.

Screenshot_20231120_083347_cgeo geocaching

ztNFny:
Where do we collect feedback regarding this?

Used it the first time today (with today's master branch) and what i didn't like visually is the "jumping" of the dialog when performing actions that resize the dialog, like adding an offline log or saving. A transition animation would probably solve that.

After marking a caches as found the map is not updated, this is with #14855 included (unless it's not in master?!)

Issue #14806, opened 2023-11-05
Cache download stuck in "Render view" - Bettercacher missing response


Lineflyer:
### Describe your problem!

While preparing a cache trip I downloaded several caches by using the map, opening the popup and tapping on store.
After doing so for a while, I suddenly ran into the problem, that the download progress got stuck in "Render view" (last step of downloading). It only vanishes after 30-45 seconds.

I did a short look into the debug log (only briefly) and it seems stuck, as there is no reply to the request to bettercacher.org. 
It looks like the site is either overloaded or there is some throttling or Anti-ddos involved.

Whatever might be the reason:
The non responsibility of a service should not block the cache from being downloaded.

I set the label "Unverified" as I might be wrong...did only a short look into the debug log.
However since I turned off bettercacher.org in c:geo the problem is gone.

### How to reproduce?

- Enable bettercacher.org service
- Download several caches (in shorter time) until you note the download is slow/stuck

### Actual result after these steps?

Stuck in cache download until timeout

### Expected result after these steps?

Bettercacher requests should run "in background" and not block UI or degrade cache download performance

### Reproducible

Unclear

### c:geo Version

2023.11.03-NB

### Additional Information

The reason for the unresponsiveness should be clarified with the owners of bettercacher.org. I would expect a proper HTML return code instead of "playing dead".

Lineflyer:
Here an example log (filtered logcat for string "bettercacher", where you can see some requests with late response, some with no response. This time the log is derived from refreshing a list of caches.

```
2023-11-05 11:22:45.306 18484-7292  cgeo         cgeo.geocaching  D  [refresh--2] [CtxLog]BetterCacherAPI.callAPI, uri=https://api.bettercacher.org/cgeo/data?gccodes=GC7P70G, fullData=false:START {BetterCacherAPI.callAPI:320/BetterCacherAPI.lambda$getCaches$1:308/BetterCacherAPI.$r8$lambda$OAZiphYId8SI9aI0NLSGuruHUI0:0/BetterCacherAPI$$ExternalSyntheticLambda0.call:0/CommonUtils.executeOnPartitions:84/BetterCacherAPI.getCaches:303/BetterCacherConnector.amendCaches:79/AmendmentUtils.amendCachesInternal:47}
[...]

PR #14807, opened 2023-11-05
rel to #14806: introduce timeout for http calls to bettercacher.org

eddiemuc:
rel to #14806: introduce timeout for http calls to bettercacher.org


eddiemuc:
Updates:
* add fix for failing test cgeo.geocaching.downloader.DownloaderTest.testOpenAndroMaps (checked map has increased in size on server site)
* add fix for potential unclosed/leaking http connections

moving-bits:
Let's merge this, not alone to get CI tests running again (thanks for fixing them along the way)

PR #14777, opened 2023-10-29
Short form bookmark url support

mortenbruhn:
## Description
- Added support for using c:geo to open coord.info short urls for bookmarks

## Related issues

None known

## Additional context
It has not (yet) been spun up, not on virtual or a phone as APK.

Comments and suggestions welcome.

PR #14818, opened 2023-11-06
Refactored ImportBookmarkLinks using newly available tools from Java 1.8

mortenbruhn:
## Description
- Refactored `ImportBookmarkLinks` using some of the tools made available by #14181
- Refactored `ImportBookmarkLinksTest` to use `Parameterized`
- Moved `UrlToIdParser` and implementing classes to separate files to allow for re-use elsewhere

## Related issues
- #14777
- #14181

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/ImportBookmarkLinks.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/ImportBookmarkLinks.java
@@ -19,18 +19,17 @@ import java.util.Optional;
 
 public class ImportBookmarkLinks extends AppCompatActivity {
 
-    ImportBookmarkLinks(@NonNull UrlToIdParser urlToIdParser) {
+    public static final String BOOKMARK_LIST_API_PREFIX =
+            "https://www.geocaching.com/plan/api/gpx/list/";
+    private final UrlToIdParser intentUrlParser;
+
+    ImportBookmarkLinks(@NonNull final UrlToIdParser urlToIdParser) {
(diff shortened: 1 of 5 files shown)


bekuno:
Ok to test

eddiemuc (commented):
Code is fine. It is quite extensive for the problem at hand, but I know from own experience that Java is just very noisy :-) 

I hesitate from merging for now because I am not sure whether we will leave in desugaring for good. In case we have to remove it again, then this PR has to be reverted - probably manually. 
@mortenbruhn would you be willing to do such a revert in this case?

mortenbruhn:
> Code is fine. It is quite extensive for the problem at hand, but I know from own experience that Java is just very noisy :-)
> 
> I hesitate from merging for now because I am not sure whether we will leave in desugaring for good. In case we have to remove it again, then this PR has to be reverted - probably manually. @mortenbruhn would you be willing to do such a revert in this case?

AFAIK, desugaring is only necessary in this case if the Android API version in `minSdk` remains at level 21. If we upgrade to a more recent one, the need for desugaring goes away.

TL;DR: Yes, I'll take my chances :)

eddiemuc:
> AFAIK, desugaring is only necessary in this case if the Android API version in `minSdk` remains at level 21. If we upgrade to a more recent one, the need for desugaring goes away.

That's true. This has been discussed before, and the team wishes to keep support for SDK 21 alive. So unfortunately upgrade is not an option.

> TL;DR: Yes, I'll take my chances :)

All right, let's take our chances together :) Thanks a lot for your contribution!

PR #14826, opened 2023-11-11
fix #14825: parse gc personal note as html with special linebreak handling

### LINKED ISSUE #14825: Personal Note, downloaded from gc.com is missing CR/LF
### Describe your problem!

When downloading a PN from gc.com, it does not contain any CR/LF anymore, but shows ``&#10;`` instead.
Moreover e. g. ``>`` will appear as ``&gt;`` in c:geo's PN ...

First thought they switched the PN to Unicode ... but umlauts still come normal.  Or is there some XML in between? Nevertheless, I would guess this happens because of a website change anyway, cause we've cross checked it with two different Nightlies (2023.10.14-NB-04446c4 and 2023.11.10-NB-0e450a) and both show this strange behavior - and I'd be really surprised if I missed this misbehavior for more than a month.  ;)

### How to reproduce?

Just download resp. refresh any cache who has a serverside PN containing CR/LFs.

PN text looks scrambled.

Text should be displayed "human readable".  ;)

### Reproducible

Yes

### c:geo Version

2023.11.10-NB-0e450a

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #14825: parse gc personal note as html with special linebreak handling

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -33,6 +33,7 @@ import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.CollectionStream;
 import cgeo.geocaching.utils.DisposableHandler;
+import cgeo.geocaching.utils.HtmlUtils;
 import cgeo.geocaching.utils.JsonUtils;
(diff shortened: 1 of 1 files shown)


eddiemuc:
Retest this please

eddiemuc:
Failing tests are unrelated and belong to a different problem described here: #14827

ztNFny:
Already got several users asking me about this bug, we should try to release this asap.

PR #14839, opened 2023-11-15
UnifiedMap: Activate OneTimeMessages

moving-bits:
## Description
Enables one-time messages in UnifiedMap


fm-sys:
I vote against these messages at unified map. They where added in times where caches got added to individual route accidentally easily. This is not the case anymore, as we meanwhile show a popup on longclicking an cache.

moving-bits:
I have no strong opinion about those one time messages. We should keep them consistent across all maps, though.

But just for clarity: There are at least two map related OTM: long tap info & live mode info, maybe more (haven't checked).

fm-sys:
True. Didn't had the second one at mind. I would vote for remove 'long tap info' on all maps and keep 'live mode info'

Lineflyer:
> I would vote for remove 'long tap info' on all maps and keep 'live mode info'

Agree!

moving-bits:
I'll prepare a PR for removing the "long tap" OTM

PR #14841, opened 2023-11-15
Context menu for routes/tracks on map and elevation chart

moving-bits:
## Description
- shows popup menu when doing a long-tap on individual route, supporting edit, show/hide, center for now
- adds an overflow menu to elevation chart with the same entries

currently implemented for individual route only, partly prepared for tracks (but no items yet)

image.image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/RouteTrackUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/RouteTrackUtils.java
@@ -7,6 +7,7 @@ import cgeo.geocaching.files.GPXIndividualRouteImporter;
 import cgeo.geocaching.files.GPXTrackOrRouteImporter;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.Viewport;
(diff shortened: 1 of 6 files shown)


moving-bits:
Updated PR to
- include more options for individual routes
- also present a context menu for tracks
- show icons on elevation chart context menu

image.image

moving-bits:
retest this please

fm-sys:
Just noticed while testing that it is possible to show and hide the popup once, but seemingly not a second time.

moving-bits:
> You need to use unified map to access this feature. We wanted to give us a reason why we should finally finish the unified map migration :)

To be able to use UnifiedMap, go to settings => system => view settings, enable edit mode. Then
- create an entry named `showUnifiedMap` and set its value to `true`, if you want to get a quick launch button for UnifiedMap on home screen, but want to stay with old maps otherwise,
- create an entry named `useUnifiedMap` and set its value to `true`, if you want to switch maps, using UnifiedMap in all contexts

Lineflyer:
Ah, OK.
I enabled `showUnifiedMap` for now to test it.

PR #14849, opened 2023-11-18
UnifiedMap: Enable adding waypoints to current target (fix #14799)

### LINKED ISSUE #14799: Cannot create WP on map, even if a target cache is set
The "Add waypoint" option is missing on unified map

|actual|expected|
-|-
grafik|grafik

moving-bits:
## Description
Enable adding waypoints to current target cache on UnifiedMap


fm-sys:
Thanks!

PR #14850, opened 2023-11-19
New GC Log API: remove legacy code

eddiemuc:
New GC Log API: remove legacy code

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/trackable/TravelBugLoggingManager.java
+++ b/main/src/main/java/cgeo/geocaching/connector/trackable/TravelBugLoggingManager.java
@@ -15,7 +15,6 @@ import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.extension.LastTrackableAction;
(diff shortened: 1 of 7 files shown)


main/src/main/java/cgeo/geocaching/connector/trackable/TravelBugLoggingManager.java
```diff
@@ -71,30 +69,29 @@ public LogResult postLog(final Geocache cache, final TrackableLog trackableLog,
...
+//                    trackableLog.action,
+//                    date.get(Calendar.YEAR),
+//                    date.get(Calendar.MONTH) + 1,
+//                    date.get(Calendar.DATE),
+//                    log);
+//        } catch (final Exception e) {
+//            Log.e("TrackableLoggingManager.postLog", e);
+//        }
+//
+//        return new LogResult(StatusCode.LOG_POST_ERROR, "");
```
fm-sys, line 94: Was it intentional that this code is commented out?

eddiemuc: No, should also be removed. I will take care of it, thanks for the hint

eddiemuc:
There's no way back, thus merging this

Issue #14642, opened 2023-09-22
Default TB log type to "visited" on directly logging a TB


moving-bits:
Reported on support:

When using search => TB => log TB, default logging action seems to be random for yet unknown TB. For some it seems to be "grab", for others "discover". Request is to set the default to "visited" when logging a TB from the TB-specific page (resp. last action, when TB is already known).

Ticket numbers: 345502, 838122, 791319, 655197

Lineflyer:
Could be a migration dependant issue.

gcey:
IMO ist should be "discover", when logging a TB from the TB-specific page.
"Visit" only makes sense on cache logs.
"Grab" is a dangerous default, because it cannot be undone.

moving-bits:
It's pretty nested code, but to my understanding it looks like it's doing this on loading a trackable from a search and starting to log it:
- retrieve trackable
- read list of available logging actions - the parsing involved here differs depending on an internal configuration setting
  - in my newly created emulator, the setting was unset, so it always used the new method. That one returned "NOTE" for a trackable which is my own, and "RETRIEVED_IT" + "DISCOVERED_IT" for a TB not yet logged by me => looks ok to me
  - if I force c:geo to use the old method, it returns an empty list as possible logging actions. This leads to "RETRIEVED_ID", "GRABBED", "NOTE" and "DISCOVERED_IT" as possible actions => ???
- c:geo then reads the last TB logging action you have used
  - if that is yet unset, RETRIEVED_IT is used, otherwise its value is checked against the available logging actions for this TB
  - if it is available for this TB, it gets set
  - if it's not available for this TB, c:geo uses the first entry in the list returned above

The only explanation I can think of currently is, that the affected users somehow have a value for the internal setting, so that c:geo uses the old method of determining the available logging actions, and that this old method no longer works.

@eddiemuc 
It looks you had introduced this switch (`Settings.enableFeatureNewGCLogApi()`) and usage of it in `TravelBugLoggingManager.loadInBackground()` - can you cross-check whether this switch can indeed be involved in giving unexpected results and whether this switch is still required at all?

eddiemuc:
> eddiemuc It looks you had introduced this switch (Settings.enableFeatureNewGCLogApi()) and usage of it in TravelBugLoggingManager.loadInBackground() - can you cross-check whether this switch can indeed be involved in giving unexpected results and whether this switch is still required at all?

[...]

PR #14851, opened 2023-11-19
rel to #14642: if wanted default type not possible, select best default

eddiemuc:
rel to #14642: if wanted default type not possible, select best default


moving-bits:
+1
Thanks for diving into this

PR #14855, opened 2023-11-20
Notify map after logging cache from popup (fix #9046)

### LINKED ISSUE #9046: Found cache not marked found after logging from popup
This might be a duplicate as this is no new bug IMHO, but I could not find a similar issue although I think it is already known.

Scenario:
- Open live map
- Click on cache to open popup
- Click on popup "three-dot" -> Log visit
- Compose found log and send log to server
- Return to live map

Result:
- Cache is still shown (as unfound)
- Only after leaving/ starting the live map the cache is correctly hidden

Expected result:
- Cache status should be updated on map (and maybe even filtered out it "hide found" is active) directly after logging

moving-bits:


ztNFny:
Is this only for new map or for the old ones as well?
 I recently needed something similar for removing caches from route on logging and had to implement it with a general reload on resume.

moving-bits:
> Is this only for new map or for the old ones as well?

This is for all maps, only recently added to UnifiedMap actually.

> I recently needed something similar for removing caches from route on logging and had to implement it with a general reload on resume.

I don't think something like this is implemented for routes yet (but haven't checked).

ztNFny:
> > Is this only for new map or for the old ones as well?
> 
> This is for all maps, only recently added to UnifiedMap actually.

Wasnt working for me, but i found the issue, see #14870

PR #14873, opened 2023-11-23
Attempt to fix #14868 (multiple cache popups in backstack)

### LINKED ISSUE #14868: [BottomSheet CachePopup] viewing multiple caches in popup adds them to the back stack
With the new BottomSheet CachePopup when you view multiple caches as popup, then swipe up to view the details, all those viewed caches are in the back stack and you need to go back maaany times to get back to map.

To reproduce:
1. On (old offline) map view cache 1 in CachePopup
2. On (old offline) map view cache 2 in CachePopup
3. Swipe up to get to CacheDetails
4. Back
5. 
Expected: Map Activity visible
Actual: Cache 1 opened in CacheDetails

moving-bits:
## Description
Clear backstack on opening cache details activity from map cache/waypoint popup.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -943,15 +943,26 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         NavigationAppFactory.showNavigationMenu(this, cache, null, null, true, true);
     }
 
-    public static void startActivity(final Context context, final String geocode, final boolean forceWaypointsPage) {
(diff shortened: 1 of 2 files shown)


fm-sys:
I think we should fix the root cause instead of messing around with the backstack.

`addBottomSheetCallback` adds a new callback every time, and as soon as the sheet gets expanded, all callbacks which were ever added get executed. Quickfix would be to use `setBottomSheetCallback` which removes all previous callbacks (but this method got deprecated)

grafik

fm-sys:
See PR #14875 for an alternative (IMHO better) approach

PR #14888, opened 2023-11-25
[unified map] add temporary unified map indicator

fm-sys:
For testing and support purposes, it would be helpful to see if someone is currently using the unified map. We should IMHO remove this marker again before a public beta.


fm-sys:
grafik

fm-sys:
The indicator (including the drawable) will be removed again anyway in a few weeks before we roll it out to all users. IMHO it's not worth spending time in fixing this.

PR #14892, opened 2023-11-25
SimpleDialog: add generic view capacity for select-dialogs

eddiemuc:
SimpleDialog: add generic view capacity for select-dialogs


eddiemuc:
This allows for using SimpleDialog for dialogs with more complex item lists (with their own layout), e.g. like the cache selection list when clicking on the map:
image

eddiemuc:
retest this please

fm-sys:
> This allows for using SimpleDialog for dialogs with more complex item lists (with their own layout), e.g. like the cache selection list when clicking on the map: image

I'm not sure if I like the design specially for this dialog. To me this looks completely overkill here, too many features no one will ever need and use.

fm-sys:
IMHO all these extra features should be disabled per default and can be opted in on a per-dialogs basis at places where it really makes sense.

Good UI only provides features a user really needs, as every additional element takes time to understand what it does, even if it's just the time for recognizing that you won't need it.

I know that c:geo is not really an example for clean UI design in general, but please always try to make it a little bit better instead of worse.

eddiemuc:
> IMHO all these extra features should be disabled per default and can be opted in on a per-dialogs basis at places where it really makes sense.

That's the way it is designed.

Should I disable grouping, filtering or both?

It should be noted that grouping is only done if there are actually entries in the list from different groups. Also filtering is only activated if the list has more than a configurable number of elements.

eddiemuc:
Update: since I assume the GUI discussion may need a longer discussion, I changed the PR to a now pure technical cleanup. Which means enabling SimpleDialog also for more complex item views as e.g. used in cache selection list.

I removed the usage of grouping for good (this needs to be set eplicitely per dialog). 

Note however that the default SimpleDialog uses a text filter as soon as there are 5 or more items in the list (this value can be customized of course). This means that it will also appear here if 5 or more cache/waypoint/route items are up for selection. 

Example for two items:
image

Example for 14 items:
image

Filtering by geocode in the second example:
image

Filtering by name in the second example:
image

fm-sys:
If  filtering is only shown for 5 or more items I think this is fine for now IMHO. (We may change this limit based on beta feedback.) 

BTW, I would call the text field "search" instead of "filter" as this better describes what this text field does from user perspective. While it's technically correct that it _filters_ items based on text match, users will use this feature to _search_ for one specific item.

eddiemuc:
>  I would call the text field "search" instead of "filter" as this better describes what this text field does from user perspective. While it's technically correct that it _filters_ items based on text match, users will use this feature to _search_ for one specific item.

Ok, I will change this

> If filtering is only shown for 5 or more items I think this is fine for now IMHO. (We may change this limit based on beta feedback.)

I agree. After the change above (renaming "Filter"  to "Search") I will merge therefore. I will react on any related GUI discussion outcomes coming up in the nightlies/beta.

ztNFny:
> This means that it will also appear here if 5 or more cache/waypoint/route items are up for selection.

This just landed in my builds. To me 5 feels to early to show the field, I'd suggest a number a lot higher (15?).
Reason: Even with 10  caches it's faster to scan the list than to open keyboard and type some filter criteria.

eddiemuc:
> > This means that it will also appear here if 5 or more cache/waypoint/route items are up for selection.
> 
> This just landed in my builds. To me 5 feels to early to show the field, I'd suggest a number a lot higher (15?).
> Reason: Even with 10  caches it's faster to scan the list than to open keyboard and type some filter criteria.

I am totally open for this. 15 feels a bit much though, how about 10?

Code change would be trivial. I used this chance to try for the first time changing a code file directly from GitHub app which resulted in PR #14919

ztNFny:
> I am totally open for this. 15 feels a bit much though, how about 10?

I'm unsure about the exact value myself. The threshold should be "the number of items it takes longer to read than it takes to open keyboard + type 2-3 chars". 10 feels okish, even though I personally would go even higher, but hard to tell.

eddiemuc:
What about never showing the search textview initiall, but instead  placing a filter symbol in the title bar?
If tapped, the search textview would be toggled off and on.

image

image

ztNFny:
> What about never showing the search textview initiall, but instead placing a filter symbol in the title bar?

ok for me. You have to tap the top of the dialog anyway to focus the search field, so that additional click wouldn't "cost" much

fm-sys:
Again, I would use a search instead of filter icon here ;-)

You may even want to consider using a material search bar view component:
https://github.com/material-components/material-components-android/blob/master/docs/components/Search.md

moving-bits:
> What about never showing the search textview initiall, but instead placing a filter symbol in the title bar? If tapped, the search textview would be toggled off and on.

Good idea... and if the keyboard is activated automatically when enabling the search field (and hidden when disabling search field), then it's only a single tap to start filtering/searching, like today, but space-saving.
Would vote for search icon as well here instead of filter icon (due to reasons mentioned by @fm-sys).

eddiemuc:
I moved the discussion to a new issue so it doesn't get lost

PR #14893, opened 2023-11-25
Open BRouter download if external routing is configured

moving-bits:
## Description
When using the "check for missing download data" menu entry, open BRouter download activity, if external routing is configured.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
@@ -35,14 +35,20 @@ import cgeo.geocaching.ui.dialog.SimplePopupMenu;
 import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.ClipboardUtils;
 import cgeo.geocaching.utils.FilterUtils;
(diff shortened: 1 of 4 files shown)


main/src/main/java/cgeo/geocaching/maps/MapUtils.java
```diff
@@ -146,38 +148,46 @@ public static Spanned getColoredValue(final String value) {
...
-                DownloaderUtils.triggerDownloads(activity, R.string.downloadtile_title, R.string.check_tiles_missing, missingDownloads, null);
-            }
-        });
+            });
+        } else {
+            // if external routing configured: try to open BRouter downloader
+            final String bRouterPackage = activity.getString(R.string.package_brouter);
+            final Intent intent = new Intent();
+            intent.setComponent(new ComponentName(bRouterPackage, bRouterPackage + ".BInstallerActivity"));
+            activity.startActivity(intent);
```
fm-sys, line 189: We probably need to guard this call to be save against activity not found exceptions, do we?

moving-bits: Added try/catch + hide menu entry, if external routing is configured, but BRouter app not installed

PR #14900, opened 2023-11-26
fix #14827: GC.com: switch foundby-search to web api, remove legacy search result parsing for nearest.aspx

### LINKED ISSUE #14827: GC.Com nearest.aspx side announced for retirement: "Found By" Search redesign needed for gc.com
### Describe your problem!

Update 14.11.23: apparently the nearest.aspx side was dropped by accident by gc.com, this will be reverted. However, retirement is announced for a later release, so c:geo needs to prepare for this

"Found By" search does not  work any more for gc.com when nearest.aspx is retired. In this case,  only results on other platforms (e.g. oc.de) will be found. 

Under the hood, c:geo uses the following Web page/URL to implement "found by" search: https://www.geocaching.com/seek/nearest.aspx?ul=Moun10Bike

If this URL is down (e.g. HTTP500) then feature doesn't work.
image

Side note: this also leads to failure of some of our tests on CI:
image

Side note to side note: the failing test seem to indicate that "search by owner" and "search by coordinate" is also not working, This is however not true since the methods tested in these tests is no longer in use for productive "search by owner/coordinate". Those two tests should probably get deactivated

### How to reproduce?

Onyl reproducible if nearest.aspx is down

* Log into c:geo
* Start a "Found by" search for a gc.com user where you know he/she has finds

No caches are found for that user on gc.com

Caches should be found for this user on gc.com

### Reproducible

Yes

### c:geo Version

first discovered in 2023.11.01, but not relevant

### System information

```text
n/a
```

### Additional Information

_No response_

eddiemuc:
fix #14827: GC.com: switch foundby-search to web api, remove legacy search result parsing for nearest.aspx

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/CgeoApplicationTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/CgeoApplicationTest.java
@@ -4,10 +4,16 @@ import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.connector.gc.GCConstants;
 import cgeo.geocaching.connector.gc.GCLogin;
+import cgeo.geocaching.connector.gc.GCMap;
 import cgeo.geocaching.connector.gc.GCMemberState;
 import cgeo.geocaching.connector.gc.GCParser;
 import cgeo.geocaching.enumerations.LoadFlags;
 import cgeo.geocaching.enumerations.StatusCode;
+import cgeo.geocaching.filters.core.DistanceGeocacheFilter;
+import cgeo.geocaching.filters.core.GeocacheFilter;
+import cgeo.geocaching.filters.core.GeocacheFilterType;
+import cgeo.geocaching.filters.core.LogEntryGeocacheFilter;
+import cgeo.geocaching.filters.core.OwnerGeocacheFilter;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.Viewport;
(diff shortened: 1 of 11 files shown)


eddiemuc:
This PR switches foundby-search to gcwebapi. By this, it removes the last usage of the "old" parsing of search results of the nearest.aspx page. Therefore, with this PR a lot of now-legacy parsing code is also removed

fm-sys:
Is PQ preview already using a new API as well?

Thanks for doing these cleanups, really appreciated!

eddiemuc:
> Is PQ preview already using a new API as well?

PQ Preview seems to use the Map-API under https://tiles01.geocaching.com/map.pq. I don't know that one very well. This seems to have been refactored in May 2023 by @ztNFny

ztNFny:
As the new search returns max 1000 results afaik what sorting does this use? Is it center-based or most recent or does it follow c:geos list sorting?
"Found by" is a special case where center-based might not be the prefered option but most recent.

eddiemuc:
> As the new search returns max 1000 results afaik what sorting does this use? Is it center-based or most recent or does it follow c:geos list sorting? "Found by" is a special case where center-based might not be the prefered option but most recent.

Interesting, haven't thought about that.

I made some quick tests. It seems when searching on gc.com with one "foundBy", then the "lastFoundDate" of the response is filled with the found date of the "foundBy" user (not the actual last found date of the cache), and the result can be sorted by this using parameter sort=lastFound

Would make a lot of sense to include that in c:geo. However, may I kindly ask to move this into its own isse due to its complexity?

PR #14911, opened 2023-11-28
enable unified map for nightly users (can be disabled via settings)

fm-sys:
IMHO it's time for our nightly users to try out unified map

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
+++ b/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
@@ -8,6 +8,7 @@ import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.settings.SettingsActivity;
 import cgeo.geocaching.unifiedmap.tileproviders.AbstractTileProvider;
 import cgeo.geocaching.unifiedmap.tileproviders.TileProviderFactory;
+import cgeo.geocaching.utils.BranchDetectionHelper;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.ShareUtils;
 import static cgeo.geocaching.utils.SettingsUtils.initPublicFolders;
@@ -16,6 +17,7 @@ import static cgeo.geocaching.utils.SettingsUtils.setPrefClick;
 import android.content.Intent;
 import android.os.Bundle;
 
+import androidx.preference.CheckBoxPreference;
 import androidx.preference.ListPreference;
 import androidx.preference.MultiSelectListPreference;
 import androidx.preference.Preference;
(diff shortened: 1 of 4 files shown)


main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
```diff
@@ -33,11 +35,14 @@ public void onCreatePreferences(final Bundle savedInstanceState, final String ro
 
         initMapSourcePreference();
 
-        final boolean showUnifiedMap = Settings.getBoolean(R.string.pref_showUnifiedMap, false);
+        final boolean showUnifiedMap = Settings.getBoolean(R.string.pref_showUnifiedMap, false) || !BranchDetectionHelper.isProductionBuild();
```
eddiemuc, line 38: I would advise to just call `Settings.useUnifiedMap` here (instead of repeating the logic expression used there). This would keep the "decision logic" whether to use unified map or not in one place.

main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
```diff
@@ -73,8 +78,11 @@ public void onResume() {
         // display checkbox pref for unified map, if showUnifiedMap is enabled
         final Preference p = findPreference(activity.getString(R.string.pref_fakekey_unifiedmap));
         if (p != null) {
-            p.setVisible(Settings.getBoolean(R.string.pref_showUnifiedMap, false));
+            p.setVisible(Settings.getBoolean(R.string.pref_showUnifiedMap, false) || !BranchDetectionHelper.isProductionBuild());
```
eddiemuc, line 81: Same as above. I would advise usage of `Settings.useUnifiedMap` here

main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
fm-sys: This line is for deciding whether the unified map section is visible inside the settings.

If we make it depend on `useUnifiedMap` a nightly user could disable unified map easily but without the possibility to re-enable it.

main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapSourcesFragment.java
fm-sys: Same as above

fm-sys:
More and more bugs are found inside unified map, so I'm unsure if we should merge this. 

On the one hand, we would get much more feedback if we enable unified map for nightly users, on the other hand we would intentionally break working features for them... Any opinions?

fm-sys:
As discussed in our dev meeting, unified map is ready for nightly users. I'll merge tomorrow evening if no one stops me...

fm-sys:
retest this please

PR #14918, opened 2023-11-28
update map downloader icons

ztNFny:
replace
image
with

image


fm-sys:
Thanks for taking care of this... Looks really nice!

PR #14947, opened 2023-12-02
Rearrange buttons in cache bottom sheet

moving-bits:
## Description
As discussed in our last dev meeting here's a small proposal for rearranging the buttons in cache bottom sheet. It's for discussion only for now, therefore not yet implemented for cache details activity (where it probably should look the same as here).

Changes are:
- Move "More details" button (with optional hint button) to the end
- Move "Edit lists", "Refresh", "Delete" and "Store" buttons into a single horizontal row. (As it can be either "Store" or "Delete", a maximum of three buttons is displayed here)
- Text for "stored ago" and "lists" go to the left of those buttons
- Text for "stored ago" is a bit shortened and without line break
- On enabling hint this will be displayed between the row with the small buttons and the large "More details" button, divided by a separator, so that the "More details" buttons always stays in a fixed place at the bottom of this view

Looks like this:

|unstored|stored 1|stored 2|
|---|---|---|
|image|image|image|

A small problem remains, as can be seen in "stored 2": The layouts don't reflow on hiding buttons, so the `LinearLayout` around the buttons will always consume the space for four buttons, even if only up to three are being displayed, leading to text to the left wrapping, although there would be enough space.
How can I let Android dynamically adjust sizes?

(I had started with a `RelativeLayout` first, but then only one button got displayed at all, even though I had `toLeftOf` relationships defined.)

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/utils/FormatterTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/utils/FormatterTest.java
@@ -108,16 +108,16 @@ public class FormatterTest  {
         if (!StringUtils.equals(Locale.getDefault().getLanguage(), Locale.ENGLISH.getLanguage())) {
             return;
         }
-        assertThat(Formatter.formatStoredAgo(0)).isEqualTo("Stored in device\n");
-        assertFormatStoredAgo(DateUtils.MINUTE_IN_MILLIS * 10, "Stored in device\na few minutes ago");
-        assertFormatStoredAgo(DateUtils.MINUTE_IN_MILLIS * 20, "Stored in device\nabout 20 minutes ago");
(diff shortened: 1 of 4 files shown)


fm-sys:
Looks pretty nice IMHO, thanks for working it out!

> A small problem remains

You could try using Visibility.Gone instead of Visibility.Invisible when hiding the buttons. Not sure if this is already the case. Let me know whether it works...

moving-bits:
> You could try using Visibility.Gone instead of Visibility.Invisible when hiding the buttons. Not sure if this is already the case. Let me know whether it works...

This is a `View.GONE` already, but the enclosing `LinearLayout` seems to keep its size even when the content gets smaller. Don't know how to trigger a relayout on this (and the neighboring column with holds the texts).

moving-bits:
retest this please

PR #14949, opened 2023-12-02
use different app id for debug builds

fm-sys:
From time to time I need to debug some code on my primary device which is a real hassle currently. This PR makes our manifest app id independent (use relative names and remove unnecessary usage of parent activity flag) and adds a applicationIdSuffix to developer builds. This allows c:geo dev builds to be installed in parallel to the main version.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -46,7 +46,7 @@
     <application
         android:name=".CgeoApplication"
         android:allowBackup="true"
-        android:backupAgent="cgeo.geocaching.backup.CentralBackupAgent"
+        android:backupAgent=".backup.CentralBackupAgent"
         android:icon="${appIcon}"
         android:roundIcon="${appIconRound}"
         android:label="@string/app_name"
@@ -199,7 +199,7 @@
                 android:value="cgeo.geocaching.MainActivity" />
         </activity>
         <activity
-            android:name="cgeo.geocaching.helper.UsefulAppsActivity"
+            android:name=".helper.UsefulAppsActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/helpers"
             android:parentActivityName="cgeo.geocaching.MainActivity"
@@ -209,7 +209,7 @@
                 android:value="cgeo.geocaching.MainActivity" />
         </activity>
         <activity
-            android:name="cgeo.geocaching.helper.QueryMapFile"
+            android:name=".helper.QueryMapFile"
             android:exported="true">
         </activity>
         <activity
@@ -235,7 +235,7 @@
             </intent-filter>
         </activity>
         <activity
-            android:name="cgeo.geocaching.address.AddressListActivity"
+            android:name=".address.AddressListActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/search_address_result"
             android:parentActivityName="cgeo.geocaching.SearchActivity"
@@ -411,11 +411,11 @@
             android:label="@string/map_map" >
         </activity>
         <activity
-            android:name="cgeo.geocaching.log.LogCacheActivity"
+            android:name=".log.LogCacheActivity"
             android:label="@string/log_new_log" >
         </activity>
         <activity
-            android:name="cgeo.geocaching.log.LogTrackableActivity"
+            android:name=".log.LogTrackableActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/trackable_touch"
             android:exported="true">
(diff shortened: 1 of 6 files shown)


fm-sys:
(I'm quite curious if our CI can handle this without adaptions)

bekuno:
Retest this please

fm-sys:
I think the instrumented tests are not able to access the app cause of the changed version number. I'll look into it.

fm-sys:
Note to myself: At some places we still use the hardcoded app ID. This needs to be fixed.

fm-sys:
While working on this, I noticed somete things which might be considered as bug. 

Our app uses a file provider which has to be unique on the whole device, otherwise the app can't be installed. We use `org.cgeo.fileprovider` although our package id is `cgeo.geocaching`, why? However, I'm not sure about side effects when changing this.

To make debug builds installable in parallel, I changed the fileprovider to `org.cgeo.debugfileprovider` for debug builds. However, as CI preferences injection did no longer work I wanted to change the app id back to normal while the build script was still using the `debugfileprovider`. This did fail however, as the file provider was already blocked. => Installed APK's don't get uninstalled after the CI run is complete. Is this the wanted behavior?

```
12:10:50  > Task :main:installBasicDebug
12:10:50  Installing APK 'cgeo-debug.apk' on 'android-30(AVD) - 11' for :main:basic-debug
12:10:51  Unable to install /home/jenkins/slave/workspace/cgeo-CI_PR-build/main/build/outputs/apk/basic/debug/cgeo-debug.apk
12:10:51  com.android.ddmlib.InstallException: INSTALL_FAILED_CONFLICTING_PROVIDER: Scanning Failed.: Can't install because provider name org.cgeo.debugfileprovider (in package cgeo.geocaching) is already used by cgeo.geocaching.debug
```
Quick fix was to just change the fileprovider name again.

@bekuno As you are probably the person which is most actively working on the gradle scripts, I would apprechiate a review.

main/AndroidManifest.xml
```diff
@@ -192,24 +192,16 @@
             android:name=".AboutActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:label="@string/about"
-            android:parentActivityName="cgeo.geocaching.MainActivity"
             android:windowSoftInputMode="stateHidden" >
-            <meta-data
-                android:name="android.support.PARENT_ACTIVITY"
-                android:value="cgeo.geocaching.MainActivity" />
```
bekuno, line 199: Removing the `android:parentActivityName` should go to a separate issue/PR, as it is not necessary for the `app id` PR, imho.

fm-sys:
Just learned how to split up a commit into multiple ones :)

IMHO we can merge, with the possibility to revert individual parts if any problems should arise, which I do not really suspect.

moving-bits:

moving-bits:
When doing my first app compile & run that included this PR, I had some strange effects, eg. BRouter was not able to connect, GM showing blank map + crashing in GeoItemLayer on map change etc.

What helped in my case was:
- Clear caches and restart AS
- delete all installations of c:geo from emulator
- add exception to your Google Maps key (if you have restricted it by app id)
- and do a fresh build & install then

eddiemuc:
Thanks @moving-bits for your comment about this PR. I was going crazy because c:geo started switching around translations of log types in logcacheactivity to strange and seemingly random names. Following your steps restored normal behaviour.

Lineflyer:
> When doing my first app compile & run that included this PR, I had some strange effects, eg. BRouter was not able to connect, GM showing blank map + crashing in GeoItemLayer on map change etc.
> 
> What helped in my case was:
> 
> * Clear caches and restart AS
> * delete all installations of c:geo from emulator
> * add exception to your Google Maps key (if you have restricted it by app id)
> * and do a fresh build & install then

I do also have such problems now.
I have tried everything except a fresh instance of the emulator (which I would like to avoid as I customized it for my needs).

moving-bits:
did you delete all instances of c:geo and c:geo tests in your emulator (2 or 3 icons), then invalidate caches in AS, then rebuild c:geo?

Lineflyer:
Will try with the invalidate caches and report back.

fm-sys:
Using `file > invalidate caches...` was enough for me. Note that you may need to back up and restore your cgeo settings, as c:geo will be installed as a new app.

(Also, you might want to uninstall the outdated app if everything works.)

Lineflyer:
I needed to remove the cgeo installation on all my (many) emulators. Afterwards I had to invalidate caches and restart, run the build on all emulators once and reboot all of them (after!) installation of the new package.
Without reboot it won't start the app triggered from Android Studio...some error like "splashactivity could not be started".

Now everything looks good.

moving-bits:
> (Also, you might want to uninstall the outdated app if everything works. But they can actually be installed in parallel - that was the intent of this PR 😉)

Keeping the old app did not work for me, in none of my emulators. Always needed to uninstall all existing builds, then invalidate caches in AS, then rebuild and reinstall.

PR #14962, opened 2023-12-04
Show refresh button only for caches already stored

moving-bits:
## Description
While working on cache popup / cache details "offlinebox" layout I've stumbled upon the fact, that the "refresh" button is always being displayed when a connectors supports it. IMHO this does not make sense for caches not yet stored offline - "store" button is sufficient in this case. - This PR removes the "refresh" button when the cache is not yet stored offline.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -2751,7 +2751,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             offlineEdit.setOnLongClickListener(moveCacheListener);
         }
 
-        offlineRefresh.setVisibility(cache.supportsRefresh() ? View.VISIBLE : View.GONE);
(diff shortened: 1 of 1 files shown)


fm-sys:
Not sure, but hadn't we even discussed this in the Dev meeting?

IMHO there are valid reasons for having a refresh button in these cases: E.g. I'm looking for geocaches on the map and want to check the GC vote ratings or last logs. As these infos are only part of the full cache download, it's quite convenient having a possibility to fetch them on demand without needing to store the cache.

moving-bits:
> Not sure, but hadn't we even discussed this in the Dev meeting?

No, we didn't. That's why it's a separate PR.

> IMHO there are valid reasons for having a refresh button in these cases: E.g. I'm looking for geocaches on the map and want to check the GC vote ratings or last logs. As these infos are only part of the full cache download, it's quite convenient having a possibility to fetch them on demand without needing to store the cache.

Ah, ok, had not thought of such a use case (never have used c:geo this way). It's like storing a cache, but without assigning a list to it. 

Ok, then let's close this PR.

PR #14968, opened 2023-12-06
UnifiedMap: Add scale bar to GM (fix #14963)

### LINKED ISSUE #14963: Map scale bar missing on Google Unified
Map scale bar is missing on Google Map/Satelite in Unified Map

moving-bits:


fm-sys:
Wow, that's interesting! Wouldn't have expected that. Especially as it looks pretty native at our old gmap...

PR #14971, opened 2023-12-08
introduce flat marker mode (fix #14479)

### LINKED ISSUE #14479: UnifiedMap: Heading indiciator needs to take map rotation into account
Found a bug while playing a bit with UnifiedMap during my vacation:

If you rotate OSM map, heading indicator will point in direction of travel and adjust accordingly, but only as long as your map is heading northwards. If you rotate your map, heading indicator will still point into the same direction, although it should adjust according to the rotated map.

(Switching to UM / Google Maps made c:geo crash, so I could not check whether it's related to OSM only.)

Version is release 2023.06.18

fm-sys:
I tried my best to extend geoItemLayer to our requirements, and implement this styling flag in VTM and gmapV2. It's not necessary to implement it for old mapsforge, as that map cannot be rotated at all, so this flag is useless.

@eddiemuc I would aprechiate a short review. thx!

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/geoitem/GeoIcon.java
+++ b/main/src/main/java/cgeo/geocaching/models/geoitem/GeoIcon.java
@@ -22,6 +22,7 @@ public class GeoIcon implements Parcelable {
     private final float xAnchor;
     private final float yAnchor;
     private final float rotation;
+    private final boolean flat;
 
     //lazy initialized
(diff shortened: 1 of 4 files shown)


eddiemuc:
Retest this please

eddiemuc (approved):
Looks very good to me. Thanks for stepping in!

PR #14973, opened 2023-12-08
WIP UnifiedMap: Auto rotate map (if enabled)

moving-bits:
## Description
Seems to work for UnifiedMap GM, but UnifiedMap VTM gives strange results (as if map rotation is being ignored, will have to investigate further).

Includes fixes for `UnifiedMapViewModel.setCurrentPositionAndHeading()` and `NavigationTargetLayer.repaintHelper()`


fm-sys:
I didn't know you were working on it as well currently. I just wanted to start over, but noticed that it would be good to give location handling inside unified map some general refactoring.

This PR should probably be rebased onto #14974. I can help you with this if you like.

moving-bits:
No problem, I can rebaes / redo after you have merged #14974. Setting WIP label here for now.

fm-sys:
As it's just a few lines of code, I already re-implemented it locally, however with using the real heading instead of using a calculated value which can never be that accurate. (A video is following shortly)

Please let me know which path we want to continue.

moving-bits:
superseded by either #14975 or #14976, so let's close here

PR #14974, opened 2023-12-08
[Unified Map] extract LocUpdater into a separate class + more

fm-sys:
### refactoring:

extract LocUpdater into a separate class and do some cleanup related to location handling


fm-sys:
retest this please

PR #14975, opened 2023-12-08
Alternative approach for implementing autorotation mode (fix #14481)

### LINKED ISSUE #14481: UnifiedMap: OSM map auto rotation not working
manual map rotation does work, but auto rotation does not, and menu entry for map rotation is not yet implemented

fm-sys:
Alternative approach for implementing autorotation mode. See comments in #14973 and #14481

closes #14976, #14481

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
@@ -4,8 +4,8 @@ import cgeo.geocaching.maps.PositionHistory;
 import cgeo.geocaching.sensors.GeoData;
 import cgeo.geocaching.sensors.GeoDirHandler;
 import cgeo.geocaching.sensors.LocationDataProvider;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.utils.AngleUtils;
-import cgeo.geocaching.utils.Log;
 
 import android.location.Location;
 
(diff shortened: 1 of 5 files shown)


moving-bits:
Something's broken with map rotation here. I'll try to find a way to reproduce it:

Part 1)
- open VTM osm.org
- set map rotation to manual and rotate the map to be non-north
- switch map rotation mode to off

Map rotates back to north orientation, but compass rose is still at whatever degree it has been before.
Only on tapping on the compass it vanishes.

Part 2)
- switch to Google Maps

Compass rose is being displayed even though map is northbound => should not be the case

Part 3)
- heading indicator is spinning wildly
- if I set map rotation to AUTO, then the map starts spinning wildly as well

(Part 3 may be caused by an artifact of the emulator, though)

https://github.com/cgeo/cgeo/assets/3754370/d0a5da5a-75cf-419c-9b15-678dfddaa6c2
(Interestingly, the emulator's integrated screen recorder wasn't able to record this, maybe frames changing too frequently, so I did an actual screen-recording)

Tested with Pixel 3a API 33 emulator.

fm-sys:
1 and 2 should be easy to fix, will do.

3 might be related to the emulator orientation. Could you try to play around with the orientation sensor values which can be controlled via the emulator extended settings?

fm-sys:
1 and 2 are fixed, furthermore I could reproduce 3 which only happens if the device orientation is set to 0°/0°/0° thus resulting in undefined behavior.

Our compass activity also does funny things in this case, so it's a more general bug not specially related to this PR.

fm-sys:
Checks are green, so let's try it out how it behaves in real life...

PR #14977, opened 2023-12-08
Fix NPE in getElevation()

moving-bits:


main/src/main/java/cgeo/geocaching/maps/routing/Routing.java
```diff
@@ -219,7 +219,7 @@ public void characters(final char[] ch, final int start, final int length) throw
                     }
                 }
             });
-        } catch (SAXException ignore) {
+        } catch (SAXException | NullPointerException ignore) {
```
eddiemuc, line 222: It is never a good idea to catch and ignore an NPE. What is the underlying root cause for the NPE? Would be better to fix that one instead.

moving-bits: You're right, PR will follow (found the cause meanwhile).
NPE source line was in `Xml.parse(...)`, and I was in the midst of working in a different problem, so I took the shortcut...

PR #14979, opened 2023-12-08
Use LifecycleAwareBroadcast to notify map on offine log (fix #14880)

### LINKED ISSUE #14880: [BottomSheet CachePopup] "Remove from route on log" doesn't work
the new BottomSheet CachePopup breaks the option "Remove from route on log" (which relies on OnResume to be called on Map activity)

moving-bits:
Use LifecycleAwareBroadcastReceiver to notify map on offine log (fix #14880)


moving-bits:
retest this please

Issue #13573, opened 2022-10-30
UnifiedMap: List selection


moving-bits:
Add list selection to ActionBar, if map is displaying a list.
(`type == UMTT_SearchResult && fromList <> 0`)

(Note: When coming from a (filtered) cache list, it will display the filtered result, as it is today. When changing to a different list, the regular map filter gets applied. When you select the list you were coming from, it will also be filtered by the map filter, no longer by the list filter you had been using in cache list.)

moving-bits:
Fixed with #14980 and #15205

PR #14980, opened 2023-12-08
UnifiedMap: Allow cache list selection

moving-bits:
## Description
First draft of list selection, when UnifiedMap got started with a list, see #13573

image

One drawback, though: Due to the additional space for the cache list spinner, the map title gets shortened quite a bit:

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarActivity.java
@@ -146,6 +146,17 @@ public abstract class AbstractNavigationBarActivity extends AbstractActionBarAct
         return true;
     }
 
+    private boolean onMapLongClicked() {
+        if (!Settings.useUnifiedMap()) {
+            return false;
+        }
+        new StoredList.UserInterface(this).promptForListSelection(R.string.list_title, selectedListId -> {
+            DefaultMap.startActivityList(this, selectedListId);
+            ActivityMixin.overrideTransitionToFade(this);
+        }, false, PseudoList.NEW_LIST.id);
+        return true;
+    }
+
     private boolean onSearchLongClicked() {
         final ArrayList<Geocache> lastCaches = new ArrayList<>(DataStore.getLastOpenedCaches());
 
@@ -254,6 +265,7 @@ public abstract class AbstractNavigationBarActivity extends AbstractActionBarAct
     private void setCustomBNitem() {
         // remove existing long tap listeners before changing menu
         findViewById(MENU_LIST).setOnLongClickListener(null);
+        findViewById(MENU_MAP).setOnLongClickListener(null);
         findViewById(MENU_SEARCH).setOnLongClickListener(null);
 
         final MenuItem menu = ((NavigationBarView) binding.activityNavigationBar).getMenu().findItem(MENU_CUSTOM);
@@ -277,6 +289,7 @@ public abstract class AbstractNavigationBarActivity extends AbstractActionBarAct
 
         // set long click event listeners
         findViewById(MENU_LIST).setOnLongClickListener(view -> onListsLongClicked());
+        findViewById(MENU_MAP).setOnLongClickListener(view -> onMapLongClicked());
         findViewById(MENU_SEARCH).setOnLongClickListener(view -> onSearchLongClicked());
     }
 
--- a/main/src/main/java/cgeo/geocaching/apps/cachelist/InternalCacheListMap.java
+++ b/main/src/main/java/cgeo/geocaching/apps/cachelist/InternalCacheListMap.java
@@ -3,7 +3,6 @@ package cgeo.geocaching.apps.cachelist;
 import cgeo.geocaching.R;
 import cgeo.geocaching.SearchResult;
 import cgeo.geocaching.apps.AbstractApp;
-import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.maps.DefaultMap;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
@@ -35,6 +34,6 @@ class InternalCacheListMap extends AbstractApp implements CacheListApp {
 
     @Override
     public void invoke(@NonNull final List<Geocache> caches, @NonNull final Activity activity, @NonNull final SearchResult search) {
-        DefaultMap.startActivitySearch(activity, cls != null ? cls : Settings.getMapProvider().getMapClass(), search, null, StoredList.TEMPORARY_LIST.id);
+        DefaultMap.startActivitySearch(activity, cls != null ? cls : Settings.getMapProvider().getMapClass(), search, null);
     }
 }
--- a/main/src/main/java/cgeo/geocaching/maps/DefaultMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/DefaultMap.java
@@ -3,6 +3,7 @@ package cgeo.geocaching.maps;
 import cgeo.geocaching.SearchResult;
 import cgeo.geocaching.enumerations.WaypointType;
 import cgeo.geocaching.filters.core.GeocacheFilterContext;
+import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.settings.Settings;
@@ -107,21 +108,18 @@ public final class DefaultMap {
         }
     }
 
-    public static void startActivitySearch(final Activity fromActivity, final Class<?> cls, final SearchResult search, final String title, final int fromList) {
+    public static void startActivitySearch(final Activity fromActivity, final Class<?> cls, final SearchResult search, final String title) {
         if (Settings.useUnifiedMap()) {
-            Log.e("Launching UnifiedMap in searchResult mode (1) (item count: " + search.getGeocodes().size() + ", title='" + title + "', fromList=" + fromList + ")");
-            new UnifiedMapType(search, title, fromList).launchMap(fromActivity);
+            Log.e("Launching UnifiedMap in searchResult mode (item count: " + search.getGeocodes().size() + ", title='" + title + "')");
+            new UnifiedMapType(search, title).launchMap(fromActivity);
         } else {
-            new MapOptions(search, title, fromList).startIntent(fromActivity, cls);
+            new MapOptions(search, title, StoredList.TEMPORARY_LIST.id).startIntent(fromActivity, cls);
         }
     }
 
     public static void startActivitySearch(final Activity fromActivity, final SearchResult search, final String title, final int fromList) {
         if (Settings.useUnifiedMap()) {
-            Log.e("Launching UnifiedMap in searchResult mode (2) (item count: " + search.getGeocodes().size() + ", title='" + title + "', fromList=" + fromList + ")");
-            final UnifiedMapType mapType = new UnifiedMapType(search, title, fromList);
-            mapType.filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.TRANSIENT);
-            mapType.launchMap(fromActivity);
+            startActivityList(fromActivity, fromList);
         } else {
             final MapOptions mo = new MapOptions(search, title, fromList);
             mo.filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.TRANSIENT);
@@ -129,4 +127,12 @@ public final class DefaultMap {
         }
     }
 
+    public static void startActivityList(final Activity fromActivity, final int fromList) {
+        if (Settings.useUnifiedMap()) { // only supported for UnifiedMap
+            Log.e("Launching UnifiedMap in list mode, fromList=" + fromList + ")");
+            final UnifiedMapType mapType = new UnifiedMapType(fromList);
+            mapType.filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.OFFLINE);
+            mapType.launchMap(fromActivity);
+        }
+    }
 }
--- a/main/src/main/java/cgeo/geocaching/ui/TwoLineSpinnerAdapter.java
+++ b/main/src/main/java/cgeo/geocaching/ui/TwoLineSpinnerAdapter.java
@@ -1,0 +1,86 @@
+package cgeo.geocaching.ui;
+
+import cgeo.geocaching.R;
+
+import android.content.Context;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ArrayAdapter;
+import android.widget.TextView;
+
+import androidx.annotation.LayoutRes;
+import androidx.annotation.NonNull;
+
+import java.lang.ref.WeakReference;
+import java.util.List;
+import java.util.Objects;
+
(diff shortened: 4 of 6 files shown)


fm-sys:
> One drawback, though: Due to the additional space for the cache list spinner, the map title gets shortened quite a bit 

As this is no longer a sub-activity of cache list, I think we could show the bottom navigation and remove the back key in the action bar instead. This would give us some more space there, with the side effects that map view height gets a bit smaller.

Furthermore, we should think about improving the entry point:

- Is it a special scenario of stored, just a different mode to view lists? Should we then remember if the user had previously viewed cache lists in 'list' or 'map' mode?

- Or is it considered as an advanced map mode? How to access it? In this case we could show the list selector always on the live map, so that users can directly switch from live to list map.

fm-sys:
We should discuss in the dev meeting if we want to include such a feature directly from day one, or wait until the "normal" unified map runs stable.

BTW, really like it! Thanks you so much for working on these things...

moving-bits:
I like the idea of switching to "regular map mode" (= showing bottom navigation, no back indicator) on switching to a different list. This would help with visualizing the difference - map from a filtered list of caches at first (cache list filter applied), then map of list (map filter applied).

It could be combined with introducing a long tap on bottom navigation's map symbol: select a list, then display contents of this list on the map. (Similar to long tap on search button leading to displaying the cache list)

moving-bits:
As discussed in yesterday's dev meeting: Let's discuss a bit further how deep this change may go. Setting WIP label for now.

moving-bits:
Proposal for cache list handling:
- mode A) (when coming from a cache list)
  - Show list name in heading, no selection indicator, but "back" indicator
  - use cache list filter (respects pre-filtering)
  - tapping on list name opens popup menu for list selection. If a list is selected, it changes to mode B
  - no bottom navigation shown

- mode B) (after a list has been selected)
  - Show list name in heading, no "back" indicator, but selection indicator
  - show all caches of selected list, with live map filter applied
  - tapping on list name opens popup (as above)
  - bottom navigation is shown

- long-tap on "map" in bottom navigation
  - allows list selection (as long tap on "search" does)
  - selecting a list enters map in mode B for the selected list

eddiemuc:
Counterproposal:

Change the map view as follows:
* As proposed, include a dropdown box in the action bar. Let this dropdown contain a first entry "Live Map", followed by a second entry "all stored caches", followed by a list of all stored lists.
* Tapping on this list lets the user select any of the entries
* Whenever the user selects "Live Map", map filter switches to the live filter and live caches are shown
* Whenever the user selects one of the stored lists, map filter switches to the "Stored" filter and only the caches of the respective list are shown
* No back indicator ever; always show bottom navigation. Map is always opened as main activity, never as sub-activity
* Long-tap on "map" in bottom navigation can be removed. It can now be done via the action bar
* Separate "live map" action can be removed. It can now be done via action bar

When user is in "stored" tab and taps "show on map" -> simply switch bottom navigation to map and select respective cache list

Where I am undecided is when user wants to map a single cache (coming from cache details view). This is because cache details view already is a sub-activity and I assume the user wants to be able to go back from the map to the cache details view in this case. Thus I propose for this case:
* Open map view as sub-activity in this case (with back arrow, without bottom navigation, showing cache name as title)
* BUT still make the action bar a dropdown. when it is selected, the first option shown is the cache title, followed by above options ("Live map" as well as stored lists). 
* When user selects any entry but the first one, this first entry is removed and activity switches to "main activity" (remove back arrow, show bottom navigation).

fm-sys:
Totally agree with @eddiemuc, IMHO really consistent concept! 
Just wouldn't show the dropdown in single cache sub-activity mode. In this case I would favor to keep a (maybe renamed) "enable live" menu item as today.

moving-bits:
Ah, I remember we had started a discussion into that direction during our last dev call. Thanks for the reminder and the concise description.

Generally agree to the proposal, with a few remarks / questions:
- with today's implementation you can map a single cache and then activate "live mode", which gives you the one specific cache and dynamically loads all still active caches. This is not identical to "live map", but pretty close. Not sure if the difference is relevant for anybody, just wanted to mention it for the sake of completeness
- displaying a list when coming from cache list activity: which filter shall be applied? Today you can filter in CacheListActivity, then map the filtered subset. What would be applied in the proposal above? And would the filter context be changed or reset when switching to a different list?
- mapping a single cache: I agree that a sub activity should be used in this case, to be able to return to the cache; I'm undecided, though, if I would display the dropdown indicator in this case, as back indicator + dropdown indicator consume much of the available space, leaving not much of the dropdown content in collapsed state
- regarding the "long tap on map icon": this is a function not available today; proposal to add this was just meant as similarity to today's "long tap on search icon", which gives you a list chooser as well, and enters cache list on selection. (Though one could argue whether putting this on the "stored icon" would be a better choice for this, then put the "last selected" long tap action on "search icon" instead.) Anyway, I have no strong opinion about those long tap actions.

Lineflyer:
> Totally agree with @eddiemuc, IMHO really consistent concept!

So do I.
Its pretty much how it works today (with respect to when the bottom bar is shown/not shown) but unifies it with the ability to switch lists and enable live.

But we should not use the possiblity to "just" disable live mode while on the map. 
Use Case for me: Start with normal live map, disable live mode (to avoid spamming the map with caches) and pan to another region, enable live mode

> Just wouldn't show the dropdown in single cache sub-activity mode. In this case I would favor to keep a (maybe renamed) "enable live" menu item as today.

Here I tend to disagree.
Once the user got familiar with the concept of being able to switch his view with the action bar, the proposal of @eddiemuc would IMHO be a perfect fit.

Lineflyer:
> * with today's implementation you can map a single cache and then activate "live mode", which gives you the one specific cache and dynamically loads all still active caches. This is not identical to "live map", but pretty close. Not sure if the difference is relevant for anybody, just wanted to mention it for the sake of completeness

AFIACS this is currently already technically identical to the normal live map: Once I "enable live" in a single cache map the app "jumps" to the normal map activity (incl. bottom bar to be shown). I assumes this means, we break out of the sub-activity once live is enabled....isn't it?

> * displaying a list when coming from cache list activity: which filter shall be applied? Today you can filter in CacheListActivity, then map the filtered subset. What would be applied in the proposal above? And would the filter context be changed or reset when switching to a different list?

What you propose above is already how we do it, isn't it?
A mapped list IMHO currently also uses the "stored" filter as long as you do not set it in live mode. The logic of when which filter (live/stored) is used, was elaborated in a longer discussion and has proven to be logical for our users...at least I never got any complain on support with the current implementation. 

> * mapping a single cache: I agree that a sub activity should be used in this case, to be able to return to the cache; I'm undecided, though, if I would display the dropdown indicator in this case, as back indicator + dropdown indicator consume much of the available space, leaving not much of the dropdown content in collapsed state

Would need to try it out but as written above I would slightly prefer a consistent UI (= always dropdown).

> * regarding the "long tap on map icon": this is a function not available today; proposal to add this was just meant as similarity to today's "long tap on search icon", which gives you a list chooser as well, and enters cache list on selection. (Though one could argue whether putting this on the "stored icon" would be a better choice for this, then put the "last selected" long tap action on "search icon" instead.) Anyway, I have no strong opinion about those long tap actions.

I like the idea of "long tap on map icon" as a shortcut to map a list. Would IMHO still be helpful despite the dropdown (same as it is for lists).
Also it would be consistent: Long tap on list = Select list to show as list  / Long tap on map = Select list to show as map

moving-bits:
> > * with today's implementation you can map a single cache and then activate "live mode", which gives you the one specific cache and dynamically loads all still active caches. This is not identical to "live map", but pretty close. Not sure if the difference is relevant for anybody, just wanted to mention it for the sake of completeness
> 
> AFIACS this is currently already technically identical to the normal live map: Once I "enable live" in a single cache map the app "jumps" to the normal map activity (incl. bottom bar to be shown). I assumes this means, we break out of the sub-activity once live is enabled....isn't it?

Ah, you're right, seems already to be the case. Somehow I thought it was different with UnifiedMap.

> > * displaying a list when coming from cache list activity: which filter shall be applied? Today you can filter in CacheListActivity, then map the filtered subset. What would be applied in the proposal above? And would the filter context be changed or reset when switching to a different list?
> 
> What you propose above is already how we do it, isn't it? A mapped list IMHO currently also uses the "stored" filter as long as you do not set it in live mode. The logic of when which filter (live/stored) is used, was elaborated in a longer discussion and has proven to be logical for our users...at least I never got any complain on support with the current implementation.

I would vote for this. From reading the proposal above I was unsure whether the filter context of the cache list would still be applied, that's why I asked.

> > * mapping a single cache: I agree that a sub activity should be used in this case, to be able to return to the cache; I'm undecided, though, if I would display the dropdown indicator in this case, as back indicator + dropdown indicator consume much of the available space, leaving not much of the dropdown content in collapsed state
> 
> Would need to try it out but as written above I would slightly prefer a consistent UI (= always dropdown).

Have a look at the second screenshot in the opening post. This is on a emulated Pixel 3a (IIRC). When displaying three or four menu icons + overflow menu indicator + back indicator + dropdown indicator, not much space is left for the dropdown itself. I'm undecided here between consistency and readability.

> > * regarding the "long tap on map icon": this is a function not available today; proposal to add this was just meant as similarity to today's "long tap on search icon", which gives you a list chooser as well, and enters cache list on selection. (Though one could argue whether putting this on the "stored icon" would be a better choice for this, then put the "last selected" long tap action on "search icon" instead.) Anyway, I have no strong opinion about those long tap actions.
> 
> I like the idea of "long tap on map icon" as a shortcut to map a list. Would IMHO still be helpful despite the dropdown (same as it is for lists). Also it would be consistent: Long tap on list = Select list to show as list / Long tap on map = Select list to show as map Short tap on list or map = Be able to switch lists in dropdown for both

I second this. Just note that this would be a change compared to today, where the "select list to display as list" is a long-tap action on the "search" icon, whereas a long-tap on "stored/list" icon gives you the "recent" list. But IMHO the plus an consistency would be worth this change.

eddiemuc:
> But we should not use the possiblity to "just" disable live mode while on the map. 
> Use Case for me: Start with normal live map, disable live mode (to avoid spamming the map with caches) and pan to another region, enable live mode

Phew, that's a hard one. It seems to me this is a not originally intended but valid use case possible due to the current way one can activate/deactivate live mode.

At first I thought we could introduce another drop-down Menü point named "Plain map only". When chosen no cache is displayed any more (I remember that such an option was requested some time ago). But this would differ from today in that also already loaded caches would vanish from the map - and I assume those should stay visible?

Could this become a newly more explicit overflow action named "Activate/deactivate cache download" (or something similar)? 

Also a functionality question: should disabling of this just stop downloading caches from the Internet or also reading and showing caches from the database?

eddiemuc:
> Have a look at the second screenshot in the opening post. This is on a emulated Pixel 3a (IIRC). When displaying three or four menu icons + overflow menu indicator + back indicator + dropdown indicator, not much space is left for the dropdown itself. I'm undecided here between consistency and readability.

True. Thus a more radical suggestion: remove the actions from the bar and place them onto the map ("go to gps position", layer selection, filter icon). A bit like e.g. locus map is doing it. 

We may save some space on the map by merging the "quick settings" and the "route/track settings" into one dialog?

moving-bits:
> True. Thus a more radical suggestion: remove the actions from the bar and place them onto the map ("go to gps position", layer selection, filter icon). A bit like e.g. locus map is doing it.
> 
> We may save some space on the map by merging the "quick settings" and the "route/track settings" into one dialog?

IIRC we had some discussion back then about why we should not merge those two. One of the reasons being "I need only the quick settings", another "the dialog will get pretty long & crowded".

But aside from that: 

I think we cannot move the quick settings further to the left, at least not without moving the scale bar to a different place, as depending on resolution we will need the full space left of this button. This is valid even for Google Maps, though OSM additionally have the map credentials.

That would leave us with the space to the right of quick settings (and probably route/track settings), and to the left of the zoom buttons. Those could be moved up a bit to get some more space, but I'm not sure if that gives us enough space (at least not on all devices out there).

What could be alternatives?

One could use a FAB as anchor for a small popup for those buttons, but that would defeat the reason why we want to keep them out of the overflow menu, at least partially (they still may be easier to hit in one-handed mode, compared to overflow menu).

Or we can place the buttons in a vertical line, starting top right below the distance views, in a line with zoom buttons, below the distance views?
Would make the map view smaller, though, but that might be acceptable with having autorotation mode?

Or we move up scale bar and map credentials, and get nearly the whole bottom row for such buttons.
Though that might look a bit strange with another navigation element, bottom navigation, directly below it.

How does Locus (or other apps) handle such a situation?

moving-bits:
One idea regarding the "live" mode:

- keep the "enable live" / "disable live" menu entry, don't add a "Live map" entry to the title spinner
- entries of spinner would be "All caches", "List 1", ..., "List n"

Thus enabling live could be combined with any of the spinner items (as well as with single cache mode), and live mode could be disabled temporarily.

eddiemuc:
> One idea regarding the "live" mode:
> 
> * keep the "enable live" / "disable live" menu entry, don't add a "Live map" entry to the title spinner
> * entries of spinner would be "All caches", "List 1", ..., "List n"
> 
> Thus enabling live could be combined with any of the spinner items (as well as with single cache mode), and live mode could be disabled temporarily.

In this scenario, assume that live mode is currently enabled. What is then the meaning of the selection in the drop down list? What happens if e.g. the user switches from "List 1" to "List 3" or from "List 3" to "All caches" or from "All caches" to "List 3"? 

If enabling of "live mode" de facto cancels out the meaning of the dropdown selection, then it shouldn't be its own control. At least it should be somehow placed "more prominently" than the dropdown control since it seems to be "superior".

moving-bits:
> In this scenario, assume that live mode is currently enabled. What is then the meaning of the selection in the drop down list? What happens if e.g. the user switches from "List 1" to "List 3" or from "List 3" to "All caches" or from "All caches" to "List 3"?

It would adjust the viewport (location & zoom level) to show the contents of the selected list, then adds caches from live connectors. Similar to what "map a single cache, then activate live mode" does today.

> If enabling of "live mode" de facto cancels out the meaning of the dropdown selection, 

It doesn't cancel it out, at least not completely, as displaying a list's contents has two parts:
- show a specific set of caches (this part would be overridden by extending the list of caches by those being returned from live connectors)
- move the map to a certain location and zoom in a way, that list contents are visible

Typical use-case for this:
- prepare a list of cache you want to visit on a tour
- on the day you want to make this tour, go to map, select live mode, and add new caches in that area

moving-bits:
update: rebased; added map type `UMTT_List`; list spinner always shown; long-tap action for bn MAP icon; no back indicator in list mode
yet to do: fill first spinner item with current values for non-list modes (+ probably more)

moving-bits:
Current implementation attempts to come as close as possible to @eddiemuc's proposal, but respects "long tap on map icon" and "temporarily enable/disable live mode" requests:
- actionbar has a dropdown including all list names (including "all caches" and "history"); first entry in dropdown is for current list type
- Tapping on this list lets the user select any of the entries
- Whenever the user selects one of the lists, map filter switches to the "Stored" (Offline) filter and only the caches of the respective list are shown
- No back indicator ever; always show bottom navigation. Map is always opened as main activity, never as sub-activity
  - exception to this: if a single cache/waypoint gets displayed, then
    - start map as sub-activity, don't show bottom navigation
    - if user selects a list from dropdown, then switch to "map as main activity"
- enable/disable live mode stays "as is"
- When user is in "stored" tab and taps "show on map" -> simply switch bottom navigation to map and select respective cache list
- long tap on bottom navigation "map" entry opens a list selection; selecting a list opens map with contents of this list

PR #14984, opened 2023-12-10
Don't show 'swipe to open' in landscape format (fix #14982)

### LINKED ISSUE #14982: Bottom sheet slide up not working in landscape mode
### Describe your problem!

When using current nightly on my tablet the bottom sheet shows up at the right corner in full height.
It also shows "Swip up to open cache details" but this does not work.

### How to reproduce?

See above.
Can provide screenshot if needed tomorrow.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2023.12.08-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
Cache/waypoint side sheet (landscape format):
- Don't show 'swipe to open'
- Make free space below side sheet touchable (to allow map movements on tablets)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
@@ -357,17 +357,20 @@ public class MapUtils {
 
     private static void configureDetailsFragment(final Fragment fragment, final AppCompatActivity activity, final Runnable onUpSwipeAction) {
 
-        final FragmentTransaction ft = activity.getSupportFragmentManager().beginTransaction();
-        final SwipeToOpenFragment swipeToOpenFragment = new SwipeToOpenFragment();
-        ft.replace(R.id.detailsfragment, swipeToOpenFragment, TAG_SWIPE_FRAGMENT);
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/maps/MapUtils.java
```diff
@@ -424,8 +426,6 @@ public void onSlide(@NonNull final View sheet, final float slideOffset) {
             };
 
             b.addCallback(callback);
-            swipeToOpenFragment.setOnStopCallback(() -> b.removeCallback(callback));
```
fm-sys, line 427: It's important to remove the callback as soon as the popup closes. See #14868. If it cannot be done via the swipeToOpenFragment, we need some other callback.

moving-bits: I've updated the PR to remove the callback on closing the popup

fm-sys (approved):
I took a closer look and realized that you only removed the callback for the side-sheet which is uncritical as it doesn't have the up-swipe behavior anyway thus wouldn't cause the side effects described in https://github.com/cgeo/cgeo/issues/14868. 

Still it's probably cleaner to remove it, so thanks for finding a solution!

PR #14993, opened 2023-12-11
rel to #14547: show waypoint connection via star lines

eddiemuc:
rel to #14547: show waypoint connection via star lines

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapStarUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapStarUtils.java
@@ -1,0 +1,68 @@
+package cgeo.geocaching.maps;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.models.Waypoint;
+import cgeo.geocaching.models.geoitem.GeoGroup;
+import cgeo.geocaching.models.geoitem.GeoItem;
+import cgeo.geocaching.models.geoitem.GeoPrimitive;
+import cgeo.geocaching.models.geoitem.GeoStyle;
+import cgeo.geocaching.ui.dialog.SimplePopupMenu;
+import cgeo.geocaching.utils.LocalizationUtils;
+import cgeo.geocaching.utils.functions.Action1;
+
+import android.graphics.Color;
+
+import androidx.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility methods to draw Geocache "Stars" onto a map.
+ * A "star" means that all waypoints of a cache are connected with the caches' header coordinate by lines.
+ * This visualization helps to quickly locate which waypoints/caches belong together.
+ */
+public final class MapStarUtils {
+
+    private MapStarUtils() {
+        //no instance
+    }
+
+    /** Adds entries to add/remove star to a cache's context menu, if necessary */
+    public static void addMenuIfNecessary(final SimplePopupMenu menu, final Geocache cache, final boolean isStarDrawnOnMap, final Action1<Boolean> setStarDrawnOnMap) {
+        if (!canHaveStar(cache)) {
+            return;
+        }
+        if (isStarDrawnOnMap) {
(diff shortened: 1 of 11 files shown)


eddiemuc:
An alternative approach to PR #14553: this PR lets user select a new option "show lines to waypoints" on long-tap:
image

For selected caches, the header coord is then connected by straight lines to all waypoints:

image

Following restrictions apply as of now:
* Feature is only available for unified map
* All lines are now red. This may lead to confusions in dense cache areas. A color scheme might be meaningful
* As of now, selection is only stored in viewmodel (not persisted). So it will be void when the map is exited and reentered
* since feature requires waypoints to be available, the feature is de facto only available for stored geocaches

fm-sys:
> As of now, selection is only stored in viewmodel (not persisted). So it will be void when the map is exited and reentered

This may be considered as intended behavior. When setting a cache as target, it also gets voided on leaving the map. This helps to declutter the map and avoids manual cleanup work, so I would vote to leave it transient.

moving-bits:
I like the idea, thanks for diving into it. Let me propose two little changes:
- I would make the lines a bit thinner, maybe two third of the current value
- Can you use some auto-coloring to get different colors for different caches?

eddiemuc:
> * I would make the lines a bit thinner, maybe two third of the current value

Sure, easy

> * Can you use some auto-coloring to get different colors for different caches?

I tried that, but I couldn't find a good set of colors to choose from. Most colors are hardly visible in certain colored backgrounds, e.g. "green" in the woods or "yellow" in cities. Black, Brown and Blue would maybe work, but we would need a set of e.g. 20-30 colors. And I don't know how to find so many. Does anyone know how Locus is solving this problem?

eddiemuc:
I reduced the line width. For the coloring I see no easy solution.

If the team wants to proceed into the direction of providing "cache stars" I would merge this PR. I feel that for the coloring discussion some personal experience is necessary before proceeding. Any opinions?

moving-bits:
> I reduced the line width. 

Thanks

> For the coloring I see no easy solution.

Do we really need a set of 20-30 colors? I don't think that a user would request waypoint connections for so many caches. (I guess it would look pretty crowded in such a case, even if they use different colors.) Maybe 5 colors or so would be fine?

eddiemuc:
> > For the coloring I see no easy solution.
> 
> Do we really need a set of 20-30 colors? I don't think that a user would request waypoint connections for so many caches. (I guess it would look pretty crowded in such a case, even if they use different colors.) Maybe 5 colors or so would be fine?

Question is: when a user activates a "star" for a cache: which color (from a set of e.g. 5 colors) is then chosen? 
* If we choose based on a hash of e.g. cache coordinates+geocode, then chances are that colors of nearby caches are the same if the set is small.
* If we choose based on the order of turning on stars for caches, then how to handle a case where we should choose colors for a group of caches/stars to be activated at once (e.g. "show me star lines for all caches currently visible")? Get's complex, and I don't like complex solutions :-)

moving-bits:
Valid questions :-)
The will probably be no automatism that works in all cases. Simplest solution therefore could be to let the user select the color, and only allow activation per cache, not per group.

bekuno:
> Do we really need a set of 20-30 colors? I don't think that a user would request waypoint connections for so many caches. (I guess it would look pretty crowded in such a case, even if they use different colors.) Maybe 5 colors or so would be fine?

I would go to max 20 colors, which are predefined.
Maybe from https://sashamaps.net/docs/resources/20-colors/ ?
 
> Question is: when a user activates a "star" for a cache: which color (from a set of e.g. 5 colors) is then chosen?

My idea for the strategy:
Use a defined sequence of colours.
If there are more caches to colour than available colours, then start again with the first colour. At this moment we should check if the star base of the previous cache with the same colour keeps a minimum distance (80m?). If this is not the case, then skip this colour.
If no colour is usable, then "bad luck", in which case there will be a collision.
Please pay no attention to any colour backgrounds on the map. The colour list may need to be optimised.

I would use a dotted or dashed line for better differentiation from most other lines.

Issue #2012, opened 2012-09-01
Option to delete and/or edit own logs


jeancaffou:
Sometimes when typing, the sneaky keyboard hides and you press "send log" accidentially.

Lineflyer:
This is the same as the ability to edit logs in general which we declined for now. Refer to mentioned issues.

Lomanic:
I couldn't find anywhere in the mentioned issues why this was declined. This is elementary and lacking since forever, we shouldn't have to open a browser, go to the cache listing, find our log, click the "Upload image/edit log" link, modify the text and send the form (and then, reloading the cache data locally as described in https://github.com/cgeo/cgeo/issues/8888#issuecomment-681129322). Mistakes happen in logs in cgeo and it's not unusual to want to fix them.

fm-sys:
I agree, this should be reopened. There is no real reason, why this shouldn't be integrated in c:geo if someone would be interested in implementing this sometime...

eddiemuc:
I would tackle this issue, but only after #2553 / #8822 is finalized, if that's all right

eddiemuc:
#8822 is merged (and first bugs already coming in...) so time to think about this one.

I suggest to approach this as follows and would be eager to have some feedback:

**General idea**:
- place "edit"/"delete" icon buttons near those logbook entries which user is allowed to edit
- for edit: Reuse (offline) log edit dialog to "load in" existing logs, then edit and resend them

**MVP**:
- start with geocaching.com
- keep "one-offline-log-per-cache" policy for now: if log should be edited while for same cache an offline log already exists, existing offline log is deleted (after questioning user of course)
- add some sort of hint/label to offline log dialog when editing existing entry. Allow conversion from "editing existing log entry" to new log entry

**Extensions/final solution**:
- add feature also for opencaching
- extend c:geo GUI such that multiple offline log entries can be stored per cache ("edit" as well as "new")
- allow refresh of currently edited log entry

That would be the rough idea. What do you think?

Lomanic:
> place "edit"/"delete" icon buttons near those logbook entries which user is allowed to edit

I think that would clutter the UI if it's always displayed. I would have seen this implemented using the menu displayed on long press on a log entry (where one can copy/share or translate log's text currently).

Lineflyer:
We could even discuss to swap current long and short click.

Currently short click performs ROT13. Its IMHO rarely useful and could either go into long press or into a common menu behind short click together with edit/delete.

[...]

PR #14994, opened 2023-12-11
rel to #2012: enhance gc.com log api to allow edit/delete for logs/images

eddiemuc:
rel to #2012: enhance gc.com log api to allow edit/delete for logs/images

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/connector/gc/GCLogAPITest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/connector/gc/GCLogAPITest.java
@@ -1,0 +1,258 @@
+package cgeo.geocaching.connector.gc;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.connector.ImageResult;
+import cgeo.geocaching.connector.LogResult;
+import cgeo.geocaching.connector.StatusResult;
+import cgeo.geocaching.connector.trackable.TrackableBrand;
+import cgeo.geocaching.enumerations.StatusCode;
+import cgeo.geocaching.log.LogEntry;
+import cgeo.geocaching.log.LogType;
+import cgeo.geocaching.log.LogTypeTrackable;
+import cgeo.geocaching.log.TrackableLog;
+import cgeo.geocaching.models.Image;
+import cgeo.geocaching.models.Trackable;
+import cgeo.geocaching.test.NotForIntegrationTests;
+import cgeo.geocaching.utils.ImageUtils;
+import cgeo.geocaching.utils.Log;
+import cgeo.geocaching.utils.UriUtils;
+
+import android.net.Uri;
+
+import java.io.File;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.function.Supplier;
+
+import org.junit.Test;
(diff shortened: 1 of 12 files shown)


eddiemuc:
This PR includes a roundtrip log/logimage creation/editing/deletion test. Let's see whether this works on CI as well...

eddiemuc:
As claimed before: this PR introduces methods to add, edit and delete cache logs, cache log images and trackable logs (only create/delete for now) against gc.com using the new log API.

It also introduces tests for this:
* one test which creates a test cache log, modifies it in several ways (including image upload) then deletes it afterwards
* one test which creates a trackable log and deletes it afterwards

This has some side-effects:
* Test takes rather long (about 30-40 seconds local)
* Since the test creates real logs against one of my own caches and trackables (and deletes them afterwards) I will receive emails from gc.com, including the user used for the test. Is this o with the team?

eddiemuc:
Update:test creates only logs of type "Note" of course (no "Found"/"DNF" logs or something like it)

fm-sys:
You may want to consider creating an unpublished cache (not yet submitted to review) from a fake GC account or use a cache which is already archived.

I don't have a problem with it, even the other way around: I think it's a good thing that the log API is finally unit tested. However, I'm not sure if GC is fine with that. It would be bad if @mucek4's personal GC account gets blocked cause of suspicious behavior.

eddiemuc:
> You may want to consider creating an unpublished cache (not yet submitted to review) from a fake GC account or use a cache which is already archived.

I don't own any archived caches. This would also not solve the problem with trackables.

As for unpublishes caches: I discovered recently that it is no longer allowed to publish write notes for unpublishes caches. Also, no one except the "owner" could see and log to such a cache.

> I don't have a problem with it, even the other way around: I think it's a good thing that the log API is finally unit tested. However, I'm not sure if GC is fine with that. It would be bad if @mucek4's personal GC account gets blocked cause of suspicious behavior.

Good thought, haven't considered that. Wouldn't it be better in general to use a fake account for CI?

fm-sys:
However, you may want to consider our `NotForIntegrationTests` annotation. Then the test will only get executed once a day (watchdog tests) but not for PR builds. This would avoid mail spamming and speeds up PR builds.

eddiemuc:
> However, another thought: You may want to consider our `NotForIntegrationTests` annotation. Then the test will only get executed once a day (watchdog tests) but not for PR builds. This would avoid mail spamming and speeds up PR builds.
> 

Watchdog test seems to execute only the tests in `WatchdogTest.java`. There is one other test marked with `@NotForIntegrationTests` which is not executed by the watchdog.

We have another job `cgeo nightly` which seems not to execute tests at all. Not sure what is the meaning of this.

eddiemuc:
> > Wouldn't it be better in general to use a fake account for CI?
> 
> Probably yes... Although it would cost us some extra money for an additional premium account.

40€ a year seems neglectable

eddiemuc:
Update:  added `NotForIntegrationTests` annotation. 
This de facto stops CI from executing the tests and thus makes them useless.
However, it allows ne to merge the PR and continue with implementing log editing

eddiemuc:
Retest this please

PR #14998, opened 2023-12-11
Swap 'append/prepend to route' menu entries (fix #14991)

### LINKED ISSUE #14991: Optimize cache longclick map context menu
Screenshot_20231211_082634_cgeo geocaching

The "append to route" option is more commonly used, therefore we should swap the context menu entries to reduce the distance between longclick and menu item.

moving-bits:


moving-bits:
retest this please

Issue #11872, opened 2021-10-13
[vectorIcons] missing translucency for visited waypoints


bekuno:
### Describe your problem!

With the new icon implementation a visited waypoint has a "checked" marker. 
But the WP is not more translucent. This was a fine feature to see also more details on the map.

### How to reproduce?

Take or create a waypoint on map.
Set the waypoint to visited.

### Expected result after these steps?

The waypoint icon is translucent.

### Reproducible

Yes

### c:geo Version

2021.10.11-RC

ztNFny:
Help me understand in which situation is that relevant? To me it seems that the option in map quick setting takes good care, at least for the usecases i have in mind:
when I play a multi and also do some other caches around it I use the map with waypoints visible and visited waypoints hidden. 
If i explitily want to see where i've been I show the visited waypoints - in that case I want them visible, not half-visible.
How are you using this?

> Not mandatory if technically problematic though.

We'd have to implement something to draw the whole layer drawable, then convert that to a bitmap, then set that as translucent - I think.

bekuno:
> 
> 
> We can add it to the survey - transparent vs checked.

Good idea.

fm-sys:
This feature is now removed for over 2 years and no one ever claimed it again, so I would vote to close here. Feel free to reopen if you disagree.

PR #15000, opened 2023-12-11
Mark visited waypoints with semi-transparent bg (optional)

moving-bits:
as discussed in yesterday's dev meeting
marked as WIP

image

Setting is configurable in settings => map content & behavior => waypoints

Edit: to be discussed, whether the little green check mark should still be displayed, if semi-transparency is configured

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -1340,6 +1340,10 @@ public class Settings {
         return getInt(R.string.pref_showwaypointsthreshold, getKeyInt(R.integer.waypoint_threshold_default));
     }
 
+    public static boolean getVisitedWaypointsSemiTransparent() {
+        return getBoolean(R.string.pref_visitedWaypointsSemiTransparent, false);
+    }
+
     /**
      * The threshold for brouter routing (max. distance)
      */
--- a/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapContentBehaviorFragment.java
+++ b/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceMapContentBehaviorFragment.java
@@ -38,6 +38,10 @@ public class PreferenceMapContentBehaviorFragment extends BasePreferenceFragment
             MapMarkerUtils.clearCachedItems();
             return true;
         });
+        findPreference(getString(R.string.pref_visitedWaypointsSemiTransparent)).setOnPreferenceChangeListener((preference, newValue) -> {
+            MapMarkerUtils.clearCachedItems();
+            return true;
+        });
     }
 
     public void updateNotificationAudioInfo() {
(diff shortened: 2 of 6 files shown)


fm-sys:
TBH, the difference is almost invisible at least on the screenshot, the feature doesn't provide additional value in its current state IMHO. The foreground should be semi transparent as well, I would say... (disclaimer: I just want to mention that I neither need or want to use this feature. At the end, the ones who want to use it need to decide on it)

Lineflyer:
What is the use case behind it or who is the requestor?
Is it worth another setting? 
AFAIR we dropped the transparent mode in favor of the check mark and additionally added a hide function for visited WPs. Sounds like transparency is superflous.

moving-bits:
This is #11872. Some people (including me, I have to add) find it useful to still see visited waypoints while being on tour (so cannot hide it), but want to make them less intrusive visually, which semi-transparency was ideal for.

I had implemented it already a while ago, but never formed it into PR, so we decided in our dev meeting to create a WIP from it. As I could not find the old approach anymore, I restarted from scratch. Can be fine-tuned visually for sure (eg: making the whole icon semi-transparent, not only the background, etc.)

eddiemuc:
> What is the use case behind it or who is the requestor? Is it worth another setting? AFAIR we dropped the transparent mode in favor of the check mark and additionally added a hide function for visited WPs. Sounds like transparency is superflous.

We decided to proceed like this in our last team call:

image

(I would use the feature if available, but I have no strong opinion about it)

fm-sys:
Really like you new draft, _way better_ than the first one IMHO. 

I think we could still show the checkmark, but no strong option about it.

bekuno:
I like this result.

I would not hide the "checkmark", especially if this semi-transparent visualisation is used for all waypoints of found caches, i.e. regardless of whether the waypoint has the status "found".

moving-bits:
Thanks for your feedbacks.

> I think we could still show the checkmark, but no strong option about it.

My reasoning was that this would be the same info ("found waypoint") given in two ways, and as every marker clutters the map, I've left it out.

> I would not hide the "checkmark", especially if this semi-transparent visualisation is used for all waypoints of found caches, i.e. regardless of whether the waypoint has the status "found".

This visualization is used for "waypoints marked as visited" only, not (yet) for unmarked waypoints of found caches.

Shall this PR be extended to make waypoints of found caches semi-transparent as well?

bekuno:
> Shall this PR be extended to make waypoints of found caches semi-transparent as well?

+1

fm-sys:
> Shall this PR be extended to make waypoints of found caches semi-transparent as well?
> 
>

Especially if this is done, we definitely need to show the checkmark.

There's nothing wrong with showing it in two ways, it just makes it easier to recognize.

Lineflyer:
Do you really think it needs an setting?
I would favor to decide for one of the solutions (transparency or checkmark) and avoid another (of already so many) setting.

moving-bits:
Have updated PR:
- "visited marker" is shown again
- waypoints of caches marked as found get semi-transparency as well
- slightly increased alpha value (increased opaqueness)

moving-bits:

FH-Inway:
@moving-bits Just discovered this feature (I am an irregular user) and wanted to let you know that I like it!
Thanks for implementing it.

PR #15012, opened 2023-12-13
Map zoom controls

fm-sys:
Use self-written zoom control for all map types and implement continuous zooming when keep touching zoom control


fm-sys:
retest this please

Issue #15003, opened 2023-12-12
[Unified Map] [Nightly] Some observations ...


MagpieFourtyTwo:
### Describe your problem!

Honestly, I really wanted to keep ``Use unified map`` enabled, but right now it's just unusable for me.

Issues I observed (as simple enumeration for now, cause I don't know, what is already known - can create dedicated issues once it's sorted out, which ones are really new resp. not yet in progress anyway):

- [x] 1. Routing to a cache does not work with OSM, if routing is started from within the cache's detail page
   i. e.: select OSM as map source (no matter if special tile or ``Combined (offline)``),
   then open any cache's detail page (NOT via map popup!) and navigate to it, using the navigation symbol in the header
   => over here map just freezes in zoom level "10.000 km"
   Strange side effect: Even though OSM freezes in a completely white area, tapping on map opens a list of waypoints ...
   As soon as I switch map to GMap, the zoom level changes to "4000 km" (which is still way to big), but the route will be displayed
   Switching back to OSM will freeze it again
   Selecting a cache from the OSM map and navigate to it via the navigation symbol in the cache's popup works - sometimes.

- [x] 2. same applies when opening a list and then ``Display on map``
  Edit as of 29.12.23: Map now opens with the listed caches in view,
  but opening behavior is strange (for more info see comment [1872385893](https://github.com/cgeo/cgeo/issues/15003#issuecomment-1872385893))

- [x] 3. line width and opaqueness of navigation route line differs from the previous (non unified) implementation as well as from representation in GMap: First, in bigger zoom levels (e. g. at 5 km) the routing line is way to wide, and strangely gets *narrower* the more I zoom in, until it finally reaches road width at zoom level of 10 m. Seems as if the width scale factor should be just the other way round; moreover it shows dots (in GMyp) resp. slices (in OSM) at it's edges, which make it hard to follow the route.

- [x] 4. history track line is *way* narrower than it was before.

- [x] 5. in GMap long tap on [+] resp. [-] opens the map context menu
   in OSM this long tap used to continue zooming in or out - but now it just does nothing until I release the button

- [x] 6. ``Combined (offline)`` went to the bottom of the list - can this please go up again, where it was before?  ;)

[...]

PR #15013, opened 2023-12-13
UnifiedMap: Fix single cache zoom (rel. to #15003)

moving-bits:
Fixes mapping a single cache (without waypoints) or waypoint for UnifiedMap VTM.

(Does NOT fix mapping a list or a cache having waypoints, so NOT fixing zooming to bounds which are an area, not a point)


fm-sys:
retest this please

moving-bits:
Retest this please

PR #15020, opened 2023-12-14
rel to #2012: new async task tooling for activity background worker tasks

eddiemuc:
rel to #2012: new async task tooling for activity background worker tasks


eddiemuc:
got a little longer than expected, but most of it is boilderplate code

eddiemuc:
Need to merge to continue work on #2012

Issue #14948, opened 2023-12-02
[unified vtm] fail to render if viewport is too large


fm-sys:
Steps to reproduce:

- create a list with a cache in USA and Australia
- show map as list

=> map stays grey and can't be moved zoomed

![Screenshot_20231202_231014_cgeo geocaching](https://github.com/cgeo/cgeo/assets/64581222/00e67766-6d08-4173-a8dd-b5b0f984db36)

moving-bits:
it may be that the zoom level calculated by `UnifiedMapActivity` for showing all caches in the same view is not supported by current map. Can you compare its value against `getZoomMin()` / `getZoomMax`?

moving-bits:
> it may be that the zoom level calculated by `UnifiedMapActivity` for showing all caches in the same view is not supported by current map. Can you compare its value against `getZoomMin()` / `getZoomMax`?

Forget this part - UnifiedMap currently ignores the values for min/max 🤷‍♂️, and during my tests I've seen pure nonsense in the map files being reported for supported min/max values (so ignoring those limits may actually be a useful mitigation). 🤔

moving-bits:
Just tried to find a small code example for reproducing this bug, therefore modified the original VTM example `MapsforgeActivity` by adding the following code snippet around line 244 (before the `catch` block):

```
Bitmap bitmapPoi = new AndroidBitmap(BitmapFactory.decodeResource(getResources(), R.drawable.marker_poi));
MarkerSymbol symbol = new MarkerSymbol(bitmapPoi, MarkerSymbol.HotspotPlace.CENTER, false);
ItemizedLayer mMarkerLayer = new ItemizedLayer(mMap, symbol);
mMap.layers().add(mMarkerLayer);
List<GeoPoint> pts = new ArrayList<>();
pts.add(new GeoPoint(51.487, 7.0425));
pts.add(new GeoPoint(47.649, -122.348));
pts.add(new GeoPoint(-33.858, 151.215));

List<MarkerInterface> markers = new ArrayList<>();
for (GeoPoint gp : pts) {
    markers.add(new MarkerItem(gp.toString(), "", gp));
}
mMarkerLayer.addItems(markers);
mMap.animator().animateTo(new BoundingBox(pts));
```

This draws a small dot in Seattle, Sidney and Essen (home of the GC7G1GA event).
[...]

PR #15026, opened 2023-12-15
rel to #14948: unified map: put another animation before bound animation to avoid missing map width/height

eddiemuc:
rel to #14948: unified map: put another animation before bound animation to avoid missing map width/height


fm-sys:
Looks good! Is this PR ready to be merged?

PR #15036, opened 2023-12-16
Make DB downgrade work on freshly created database

eddiemuc:
Make DB downgrade work on freshly created database


moving-bits:
> Make DB downgrade work on freshly created database

This is an elegant solution, thanks for stepping in! (Have been AFK this weekend)

PR #15046, opened 2023-12-17
Cleanup of GeoCache class (rebase PR 13249)

murggel:
## Description

Rebases PR #13249 to the current master
Use CalendarUtils.isPastEvent instead of Geocache.isPastEvent -> remove Geocache.isPastEvent
Use sharing functions directly from ShareUtils instead of wrapper-methods of Geocache... -> Remove wrapper-methods from Geocache

## Additional context

I will create a follow PR to remove the Connector-related methods to a ConnectorUtility-class


eddiemuc:
LGTM, thanks @murggel for diving into this. Also thanks to @NotWoods for the original PR #13249 (and sorry it took us so long to apply it...)

PR #15051, opened 2023-12-19
rel to #2012: allow log deletion (experimental, behind feature toggle)

eddiemuc:
rel to #2012: allow log deletion (experimental, behind feature toggle)


eddiemuc:
Retest this please

PR #15080, opened 2023-12-23
UnifiedMap: Reduce CPU consumption of map (fix #15073)

### LINKED ISSUE #15073: UnifiedMap: High CPU consumption with new autoration implementation
Since a couple of days I noticed a very high CPU consumption for the emulator when entering UnifiedMap. Today I tested a few commits to narrow it down, and it seems to be the new algorithm for autorotation (PR #14975) causing it:

- UnifiedMap VTM CPU usage with build 518a88ae36db (first one with PR #14975): 89% (totalling to 100% CPU with other processes)
- UnifiedMap VTM CPU usage with build a982ce5535da (last one before PR #14975): 11%
- Home screen with the latter build: < 5% CPU usage

Those readings were done with autorotation mode set to "manual", so it's not the map rotation itself, but something else. Maybe location (heading) gets updated too frequently?

@fm-sys 
Can you have a look at it? Thanks.

moving-bits:
## Description
UnifiedMap: Reduce CPU consumption of map by reducing location update frequency from 'update immediately' to 2-5 per second
(twice per second as default, as in our old maps, and five times per second as a compromise for autorotation mode. Could probably reduced further - need to try in real life)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
@@ -19,6 +19,7 @@ public class LocUpdater extends GeoDirHandler {
 
     // minimum time in milliseconds between position overlay updates
     private static final long MIN_UPDATE_INTERVAL = 500;
+    private static final long MIN_UPDATE_INTERVAL_AUTOROTATION = 200;
     private long timeLastPositionOverlayCalculation = 0;
     // minimum change of heading in grad for position overlay update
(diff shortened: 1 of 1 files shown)


fm-sys:
Please give me time to look into it...

moving-bits:
sorry, did already merge, as UnifiedMap in emulator was driving me crazy...
But we can adjust later if necessary.

fm-sys:
See PR https://github.com/cgeo/cgeo/pull/15083 as follow up

PR #15082, opened 2023-12-23
rel to #14642: repair trackable logtype default selection

eddiemuc:
rel to #14642: repair trackable logtype default selection


moving-bits:
Let's try in the nightlies. Thanks for providing the fix!

PR #15083, opened 2023-12-23
[unified map] increase autorotation to 10 FPS

fm-sys:
Tested PR #15080 on emulator and real device. I think the fix was a bit too optimistic ;-)

Rotating the map only after 200 + x ms (<5 FPS) caused autorotation to look really laggy. 

We trigger location updates via `postValue`, which means they only get processed as soon as the UI thread goes idle. So posting too many updates would simply cause them to be skipped. 

Still, we should avoid unnecessary CPU consumption, so I gave it 10 FPS as compromise. (Actually, it's still even less as 10 FPS, as location updates only get delivered every 50-200ms measured after the 100ms delay is over.)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LocUpdater.java
@@ -19,7 +19,7 @@ public class LocUpdater extends GeoDirHandler {
 
     // minimum time in milliseconds between position overlay updates
     private static final long MIN_UPDATE_INTERVAL = 500;
-    private static final long MIN_UPDATE_INTERVAL_AUTOROTATION = 200;
+    private static final long MIN_UPDATE_INTERVAL_AUTOROTATION = 100;
     private long timeLastPositionOverlayCalculation = 0;
     // minimum change of heading in grad for position overlay update
     private static final float MIN_HEADING_DELTA = 15f;


moving-bits:
I guess we have two different use-cases here: One is to have a just in time and super-precise orientation, which might make use of a ten-times per second update, whereas during navigation on a bike, even today's twice per second was more than enough for me (in combination with today's smooth animation during rotation), and being "out in the wild" I value power-saving much more than having a precise north-orientation every tenth of a second. Maybe we need a setting for it, I regret to say.

fm-sys:
Or, what I would favor, switch between these modes automatically: 

If speed is more than 5 km/h for over 5 seconds (values can be adopted) -> Average orientation and only update it way less frequently, maybe even calculate heading out of GPS position. Heading changes would need to be animated to avoid map looking laggy.

Standing still -> User wants a precise and instant feedback. Set new headding instantly without animating it, but with high FPS rate. Our compass doesn't update only twice a second either ;)

eddiemuc:
+1 for a setting. Could be a slider telling how frequently the position shall be updated, or a Boolean switching between "high" and "low" frequency.

I, for one, would use a low frequency update all the time. Also when standing still. I value my battery 🔋 too much :-) Also I don't like such shakiness. High frequency would render the feature useless for me. But other users may see it differently. Hence the setting.

fm-sys:
I think if we go with a setting it should be a boolean. Users shouldn't mind about internal workings. Setting an update FPS rate via slider would definitely be too much IMHO.

Thinking about the "two use-cases" argument:
Should it really be added as a deeply hidden setting or rather as a fourth map rotation mode so that users can use both map rotation modes in map based on their current needs. Would be something like... 

Map rotation ->
- off
- manual
- auto (driving) 
- auto (compass)

moving-bits:
Pro fourth rotation mode setting:
- easy access for the user, easy to switch between the modes (need to think about the name, though, IMHO, "auto (compass)" vs. "auto (driving)" does not really explain what's the difference from a user's perspective)

Pro separate boolean setting (something like "use high precision in autorotation mode for low speeds")
- it's a "set and forget" option, does not clutter the UI
- you don't have to change all occurrences of autorotation in the source for the second variant, only need to adapt the one place per map where the min time difference is calculated

moving-bits:
See PR #15111 for a proposal that implements
- separate autorotation mode "Autorotation (precise)", speed dependent, update every 80ms
- in contrast to "Autorotation (low power)", reduced to 500ms (as other map rotation modes)

bekuno:
retest this please

bekuno:
@fm-sys This PR is now on conflicting state.

fm-sys:
I think this PR is superseded by https://github.com/cgeo/cgeo/pull/15111. Thanks @moving-bits for taking care of it.

Issue #15084, opened 2023-12-23
Too much semi-transparency


bekuno:
### Describe your problem!

While testing the semi-transparent WP from PR #15000 ( I love it) I discovered strange behaviours.

First, some of the found tradi caches are normal shown, but some are semi-transparent.
![grafik](https://github.com/cgeo/cgeo/assets/32555742/b2795446-02bd-43a0-b477-16ddda475fa0)

I did not found a reason for it.

Second, the smileys on the caches recent logs are every time semi-transparent.
![grafik](https://github.com/cgeo/cgeo/assets/32555742/a09e9c0e-0856-4b40-a779-725391b3a105)

### How to reproduce?

see above

### Reproducible

Yes

### c:geo Version

2023.12.23-NB

moving-bits:
I'll look into it

PR #15086, opened 2023-12-23
Mark waypointMarker as mutated

moving-bits:
## Description
Attempts to fix the artefacts described in #15084, but still shows some of them (therefore marked as draft).
Even though the created waypoint marker gets marked as mutated (which marks all elements of the `LayerDrawable` as mutated) still somehow the marker displayed on the map switches to a semi-transparent one.

Details / test case:
- Have "semi transparency for visited waypoints" enabled
- Display a map with visited and unvisited waypoints
- Even if a waypoint of an unvisited waypoint may be ok in the beginning, it switches to being marked as visited (with semi-transparency) after some time, if you open and close this marker. It looks like once there has been a single waypoint marker with semi-transparency, all waypoint markers being drawn after that one get semi-transparency as well, even though the code does NOT add it per se & even though the `LayerDrawables` are marked as mutated.

What else is missing?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
@@ -252,9 +252,14 @@ public final class MapMarkerUtils {
                     .append(cache.isArchived())
                     .append(cache.isLinearAlc() ? waypoint.getPrefix() : false)
                     .append(cache.getAssignedEmoji())
-                    .append(cache.getType());
+                    .append(cache.getType())
+                    .append(cache.isFound());
         }
         final int hashcode = hcb.toHashCode();
+
+        if (cache != null && StringUtils.equals(cache.getName(), "Der Hausmeister vom Palast der Republik")) {
+            Log.e("wpt hash=" + hashcode + ", v=" + waypoint.isVisited() + ", cf=" + (cache != null ? cache.isFound() : "---") + " wpt=" + waypoint.getName() + (cache != null ? " (" + cache.getName() + ")" : ""));
+        }
 
         synchronized (overlaysCache) {
             CacheMarker marker = overlaysCache.get(hashcode);
@@ -329,7 +334,9 @@ public final class MapMarkerUtils {
             addListMarkers(res, insetsBuilder, getAssignedMarkers(cache), false, applyScaling);
         }
         final LayerDrawable ld = buildLayerDrawable(insetsBuilder, 8, 8);
+        ld.mutate(); // @todo: somehow this does NOT mark all included parts as mutated
         if ((waypoint.isVisited() || (cache != null && cache.isFound())) && Settings.getVisitedWaypointsSemiTransparent()) {
+            Log.e("wpt v=" + waypoint.isVisited() + ", cf=" + (cache != null ? cache.isFound() : "---") + " wpt=" + waypoint.getName() + (cache != null ? " (" + cache.getName() + ")" : ""));
             ld.setAlpha(144);
         }
         return ld;


main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
```diff
@@ -252,10 +253,15 @@ public static CacheMarker getWaypointMarker(final Resources res, final Waypoint
                     .append(cache.isArchived())
                     .append(cache.isLinearAlc() ? waypoint.getPrefix() : false)
                     .append(cache.getAssignedEmoji())
-                    .append(cache.getType());
+                    .append(cache.getType())
+                    .append(cache.isFound());
         }
         final int hashcode = hcb.toHashCode();
 
+        if (cache != null && StringUtils.equals(cache.getName(), "Der Hausmeister vom Palast der Republik")) {
```
eddiemuc, line 261: :-)

moving-bits: I was mapping a part of Berlin (obviously :-)) and this virtual cache was a fine test case, its stages being shown as unfound first, but due to a found multi with stages nearby later turning into found stages :-(

eddiemuc: Ah. I thought it might be a comment related to bug fixing work :-)

Merry Christmas to everyone. 🎄🎁🦌⛄🎅

eddiemuc:
If I understand the PR correctly you want to change existing markers (make them semi-transparent) in case of a setting. 

I would advise not to do that. If I remember correctly the logic of mapmarkerutils is that caches instances are treated immutable.

Instead, I would use the setting itself as another hashcode-property. So if the user changes the setting then new marker instances will be created and stored automatically.

eddiemuc:
The same pattern is used for the setting `showDTMarkers` (for cache markers). See the hashcode-building part in method `getCacheMarker`

moving-bits:
Further up, where `createWaypointMarker()` is being called, I've added the relevant conditions to the hashcode builder (`waypoint.isVisited()` and `cache.isFound()`), but that did not help. Somehow all waypoint markers get the transparency applied to from some point in time on, though I thought calling `mutate()` would prevent exactly this. I'm calling `mutate()` on the resulting `LayerDrawable`, which in turn calls it on every layer.

eddiemuc:
Please don't use mutate. It will be applied to a cached marker instance and might have weird side effects on other parts of code (which assume the instance will never change).

Please additionally include Settings.showwaypointstransparent in the hashcode. Otherwise changing the setting will have strange effects.

Out of my head: maybe we have a hashcode clash. But that seems unlikely. I would have to look deeper into code

moving-bits:
> Please don't use mutate. It will be applied to a cached marker instance and might have weird side effects on other parts of code (which assume the instance will never change).

we use it already in `MapMarkerUtils` for applying the background tint, IIRC

Doesn't `mutate()` simply tell the Android graphics system to not reuse the internally cached resource instance (which may have some tinting applied etc.) for the new marker, but create a new one? It should not change an already created marker. (At least that's my understanding, but I may be wrong here, as I'm not that deep into internal Android graphics' workings)

> Please additionally include Settings.showwaypointstransparent in the hashcode. Otherwise changing the setting will have strange effects.

On changing the setting I reset the cache (as we do for some of the other settings having impact on the marker icons).

> Out of my head: maybe we have a hashcode clash. But that seems unlikely. I would have to look deeper into code

that's why I put the hash code in the logs, but at least in the cases I have cross-checked, this did not seem to be the case

main/src/main/java/cgeo/geocaching/utils/MapMarkerUtils.java
moving-bits: > Merry Christmas to everyone. 🎄🎁🦌⛄🎅

Thanks & same to all of you @cgeo/team. Thanks for another fantastic year for c:geo, for all the dedication, thoughts and efforts you have put into the best geocaching app for Android :-)

fm-sys: Thanks for all the work everyone has spent on improving cgeo during the last year... Merry Christmas! 🎅🎄🌟🎁

eddiemuc:
> > Please don't use mutate. It will be applied to a cached marker instance and might have weird side effects on other parts of code (which assume the instance will never change).
> 
> we use it already in `MapMarkerUtils` for applying the background tint, IIRC
> 
> Doesn't `mutate()` simply tell the Android graphics system to not reuse the internally cached resource instance (which may have some tinting applied etc.) for the new marker, but create a new one? It should not change an already created marker. (At least that's my understanding, but I may be wrong here, as I'm not that deep into internal Android graphics' workings)

Hm I might have gotten confused here. As long as mutate is only used during creation of a marker (which is then never changed again while cached) it should be ok.

> > Please additionally include Settings.showwaypointstransparent in the hashcode. Otherwise changing the setting will have strange effects.
> 
> On changing the setting I reset the cache (as we do for some of the other settings having impact on the marker icons).

Ok, then we use two different cache-ensuring mechanisms intermixed. Not good IMHO. Personally I would prefer solely including settings in hashcode, that way cache consistency ensurance is encapsulated in MapMarkerUtils and doesn't rely on other classes clearing the cache at the right point.

Are we sure that the cache is cleared at all relevant places?

moving-bits:
Have not been able to reproduce the artifacts after cold booting the emulator, so maybe this PR works as intended...

moving-bits:
retest this please

moving-bits:
> > On changing the setting I reset the cache (as we do for some of the other settings having impact on the marker icons).
> 
> Ok, then we use two different cache-ensuring mechanisms intermixed. Not good IMHO. Personally I would prefer solely including settings in hashcode, that way cache consistency ensurance is encapsulated in MapMarkerUtils and doesn't rely on other classes clearing the cache at the right point.
> 
> Are we sure that the cache is cleared at all relevant places?

Currently we use this cache-clearing mechanism in `PreferenceMapContentBehaviorFragment` for changes on those preferences:
- d/t marker on cache icon
- big smileys
- semi transparency on visited waypoints

For some reason I don't know it's also called in `CGeoMap.onPause()`, but that call got introduced more than ten years ago (and neither NewMap nor UnifiedMap seem to require it).

Additionally we clear all icon caches (including emojis) on changing the scaling preference (called in `PreferenceAppearanceFragment`) and on program startup.

moving-bits:
Let's try it in the nightlies. It shouldn't break anything, but can help mitigating the linked issue.

PR #15088, opened 2023-12-25
Sheet info: Remove old layout listeners before setting new one

moving-bits:
## Description
Sheet info: Remove old layout listeners before setting new one
Also moves all sheetInfo methods to common map fragment class


moving-bits:
retest this please

moving-bits:
CI fails on this PR, showing outdated source code lines that are not part of this PR. Let's close here therefore an redo.

PR #15093, opened 2023-12-25
Sheet info: Remove old layout listeners before setting new one

moving-bits:
## Description
Sheet info: Remove old layout listeners before setting new one
Also moves all sheetInfo methods to common map fragment class


moving-bits:
ok, Git hickup clarified: some other change on master introduced lines no longer working with this PR, though this PR in itself was fine (and compiling fine locally) - only failed on automatic rebase to `master`. Will rebase this PR to `master` locally, fix the side-effect from the other PR, and force-push again.

Issue #14881, opened 2023-11-25
[BottomSheet CachePopup] improve swipe-up-view


ztNFny:
When swiping up the screen is covered by a generic "Swipe top open details view" most of the way up. We should use this space better.
![image](https://github.com/cgeo/cgeo/assets/1258173/eaa3e3e8-1c16-4720-a69a-8b73a6e25adf)

fm-sys:
You are suggesting to use this space better, but without the info what to show? I would consider this feature request incomplete. "Please change it to something else, don't know what" 

Keeping the cache popup view unchanged will cause the transition to look bumpy as the popup differs from the actual cache details activity layout-wise.

Furthermore, this generic view is normally visible only for a fraction of a second while swiping up and indicates what will happen if you release the bottom sheetat the current position.

ztNFny:
> You are suggesting to use this space better, but without the info what to show?

I intentionally left that out. Keeping CachePopup would be the minimal requirement, but maybe we find some better use for that space? Like a CachePopupExpanded with more data.
I feel that BottomSheet is in such an early state of development that there's room for brainstorming.

> Keeping the cache popup view unchanged will cause the transition to look bumpy as the popup differs from the actual cache details activity layout-wise.

I disagree, keeping CachePopup would be a lot better. Starting CacheDetailsActivity first shows an empty view with Loading dialog before showing the actual cache data. So going from "anything" to CacheDetailsActivity redraws the screen twice: Empty activity with dialog, filled CacheDetailsActivity.

As of now we have 4 transitions with only 1 of them being smooth (the one between the 2 swipe views): CachePopup, Swipe with Up arrow, Swipe with Checkmark, CacheDetails empty, CacheDetails ready

> this generic view is normally visible only for a fraction of a second while swiping up

Unfortunately that's not true. Once fully swiped up that view stays visible for approx half a second before CacheDetailActivity is started. So it's very visible. (That delay should be reduced as far as possible as it makes c:geo slower)

ztNFny:
How about as a "very fast" first step to display that "Swipe to .." stuff below the existing CachePopup info?  Would probably be little work (didn't look into the implementation) and imho a lot better than what we have, then we can still improve upon it later.

ztNFny:
> How about as a "very fast" first step to display that "Swipe to .." stuff below the existing CachePopup info? Would probably be little work (didn't look into the implementation) and imho a lot better than what we have, then we can still improve upon it later.

[...]

Issue #14874, opened 2023-11-23
Waypoint popup: Cannot swipe up if description is too long


moving-bits:
New waypoint details view in maps:

If the waypoint description for a waypoint is so long that the bottom sheet is already fully expanded, you cannot swipe up to open the cache details view. Not sure if there is a mitigation for this.

fm-sys:
No idea how to fix this easily. But at least as long as we keep the additional "open" button in, it shouldn't be a problem.

ztNFny:
Maybe we could limit the height of the bottom sheet to max x% (60-70 maybe?)? A fullscreen BottomSheet would probably be confusing to users as well, thinking they are in CacheDetails instead.

moving-bits:
No idea about the technical details, but cache details page supports a swipe-down even if the page is completely filled, so I guess there must be a way to do a swipe-up even on full waypoint details pages, at least if you have scrolled down to the bottom of the sheet.

> Maybe we could limit the height of the bottom sheet to max x% (60-70 maybe?)? 

This would not help with swiping up, since you still have to distinguish between "I want to see more of the text, expand the window" and "I want to swipe up". (Except if you want to keep the bottom sheet size at a maximum of x%, which would limit visibility of waypoint description.)

But with regard to this:

> A fullscreen BottomSheet would probably be confusing to users as well, thinking they are in CacheDetails instead.

such a limit on initially opening the bottom sheet would still be useful.

fm-sys:
A fix for both problems (which would technically even be really easy to implement) would be to limit the description textview to a specific number of `maxLines`

moving-bits:
> A fix for both problems (which would technically even be really easy to implement) would be to limit the description textview to a specific number of `maxLines`

but how to determine a reasonable number for `maxLines` across different screen sizes & font sizes?

fm-sys:
Probably it should be possible to calculate this via screen height and font size (sp-to-px convertion)

Or we just choose something which fits on most normal devices.

moving-bits:
Seems to have become better since introducing the "swipe handle", and it's a minor thing anyway, with having the "details" button, so let's close here.

PR #15097, opened 2023-12-25
sheet info: use bottom swipe info instead of fullscreen transition

moving-bits:
## Description
Cache / waypoint sheet:
Replaces the fullscreen swipe-up gesture and transition with a smaller info box below the rest of the content:

image

(visible on swipe-up only)

Info sheet stays visible while cache details load


fm-sys:
What's the benefit of it? Tested the PR and - sorry to say - this feels completely buggy.

We may discuss showing the popup content while loading, but the popup definitely needs to be full height in extended state. 

It just feels totally unexpected if an activity suddenly pops up, although I haven't swiped it to that state.

Apps which have similar up-swipeable components (Google Maps, Musik Apps like Spotify), in all these cases the transition will go all the way to the top. And there are reasons for it, it just looks completely ugly otherwise.

Until now, no reasonable argument was shown why the current visually really clean implementation should be changed to the worse, and not a single nightly user has complained about it. (There was the request to show the hint text instead of cache details activity, but that's a completely different feature request which won't get solved by this PR in any way)

moving-bits:
Hey, that's what PR are for: To have a look at a proposal and to discuss it. (Though probably a bit less emotional, I would prefer.) What one person feels is visually fine, others may think of "can be improved" or even "needs further improvement". We should discuss such remarks and proposals, but without judging them, IMHO. BTW: Google apps are not always the best examples for visually consistent and easy to use apps, some of their apps and websites are quite the contrary, to be honest.

But back to this topic: In the related issue #14881 a couple of things are named which lead me into investigating that. For me it was:
- long swipe way
- content of popup not visible during swiping, and especially during loading the activity (which may take something between a half and a couple of seconds, depending on current state of device)
- using a text below the info sheets content was already proposed in that issue as a possible alternative
- (one could even discuss not using the swipe-up at all, as its behavior is incomplete/inconsistent currently - but that's not what I'm aiming at)

There are more things which are not working currently with the swipe-up, that should be fixed (though they are out of the scope of this PR):
- if you have a long description text, you can scroll this text to its end, but you cannot scroll back, as this gesture will be regarded as swipe down (#14874)
- in some constellations (again: long text) swipe up cannot be used at all, it simply does not work

Don't get me wrong: I don't say that this PR is the ultimate solution (it surely isn't), but let us discuss what can & what should be improved in current implementation.

Lineflyer:
- IMHO the long swipe way in the current implementation is OK for me. This helps to not accidentally open the details (which might be a time consuming action while its loaded) while e.g. cleaning the display in the rain or simply having thick fingers ;)
- What is IMHO missing in the current implementation is any hint, that there actually is a swipe up gesture available, thus my suggestion in #14881 to replace the "more details" button by something else. The simplest possibility would be to add "Up arrows" to the button text (but that's probably not nicely looking).

With respect to this PR, which I checked out on my emulator:
I have to agree to @fm-sys, that it now looks like something is "blocking the UI"/hanging while swiping.

moving-bits:
> What is IMHO missing in the current implementation is any hint, that there actually is a swipe up gesture available, thus my suggestion in [[BottomSheet CachePopup] improve swipe-up-view #14881](https://github.com/cgeo/cgeo/issues/14881) to replace the "more details" button by something else. The simplest possibility would be to add "Up arrows" to the button text (but that's probably not nicely looking).

Would the regular "swipe up" indicator suffice for this? Like in map quick settings:

image

fm-sys:
> Though probably a bit less emotional, I would prefer. 

Honestly I felt bad writing this and if definitely wasn't meant as offense. UI design is an emotional topic. In fact both versions are pretty much the same with or without this PR. However it _feels_ completely different. I tried to express my opinion and write more than just "don't like it" so it need to be emotional, maybe too emotional though.

fm-sys:
> - long swipe way

I don't know if you are using the nightlys or dev builds on a real device (which I can recommend ;)), but I never had the problem that the swipe gesture was interpreted wrong. I can only speak for myself though. This is mostly due to the fact that not only swipe distance but also the velocity is taken into consideration for interpreting the swipe.

moving-bits:
> > Though probably a bit less emotional, I would prefer.
> 
> Honestly I felt bad writing this and if definitely wasn't meant as offense. UI design is an emotional topic. In fact both versions are pretty much the same with or without this PR. However it _feels_ completely different. I tried to express my opinion and write more than just "don't like it" so it got emotional, maybe too emotional though.

All good for me, now, I've calmed down a bit myself meanwhile :-)
Let's close this PR and continue discussion about possible improvements in #14881.

moving-bits:
> > * long swipe way
> 
> I don't know if you are using the nightlys or dev builds on a real device 

So far I've been on emulator only. The only testing device I can install my own builds on (on a regular basis) is a pretty old S5 neo running on Android 6, with limited device space left. While trying to activate the dev builds there I've stumbled across #15101.

Meanwhile I'm able to install c:geo on that device in parallel to the release version, but it's very laggy and crashes frequently (which I guess is due to the age of the system, not caused by c:geo itself, as it runs fine on a more sophisticated device). So it's somewhat hard to distinguish between a "real error" and something caused by running on a low-end device (from today's perspective).

Issue #15101, opened 2023-12-26
Parallel installation of c:geo failing to show map & settings


moving-bits:
Today I tried installing a dev version of c:geo parallel to the latest release version using AS.

Installation itself worked (although AS started the wrong version after that, meaning: the release version), but the dev installation was only partially usable (no map, no settings). From the logfile gathered I can see two exceptions:

### exception 1: binding BRouter service

```
12-26 09:15:16.909 13134 13134 W cgeo : [main] UncaughtException
12-26 09:15:16.909 13134 13134 W cgeo : java.lang.RuntimeException: Unable to start activity ComponentInfo{cgeo.geocaching.developer/cgeo.geocaching.AboutActivity}: java.lang.SecurityException: Not allowed to bind to service Intent { cmp=cgeo.geocaching/.brouter.InternalRoutingService }
12-26 09:15:16.909 13134 13134 W cgeo : at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3255)
12-26 09:15:16.909 13134 13134 W cgeo : at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3351)
12-26 09:15:16.909 13134 13134 W cgeo : at android.app.ActivityThread.access$1100(ActivityThread.java:222)
12-26 09:15:16.909 13134 13134 W cgeo : at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1796)
12-26 09:15:16.909 13134 13134 W cgeo : at android.os.Handler.dispatchMessage(Handler.java:102)
12-26 09:15:16.909 13134 13134 W cgeo : at android.os.Looper.loop(Looper.java:158)
[...]
```
(and a similar one from UnifiedMap)

I remember having the same exception in the emulator before I uninstalled the previous dev version (with cgeo.geocaching as package name).

### exception 2: accessing clipboard
```
12-26 09:16:27.659 13433 13433 W System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'int android.content.ClipData.getItemCount()' on a null object reference
12-26 09:16:27.659 13433 13433 W System.err: at android.os.Parcel.readException(Parcel.java:1626)
12-26 09:16:27.659 13433 13433 W System.err: at android.os.Parcel.readException(Parcel.java:1573)
12-26 09:16:27.659 13433 13433 W System.err: at android.content.IClipboard$Stub$Proxy.getPrimaryClip(IClipboard.java:197)
[...]

PR #15107, opened 2023-12-26
Catch errors/exceptions on reading clipboard

moving-bits:
## Description
Fixes the other part of #15101


main/src/main/java/cgeo/geocaching/utils/ClipboardUtils.java
```diff
@@ -37,10 +37,18 @@ public static void copyToClipboard(@NonNull final CharSequence text) {
...
-            return text != null ? text.toString() : null;
+        if (clipboard != null) {
+            try {
+                final ClipData clip = clipboard.getPrimaryClip();
+                if (clip != null && clip.getItemCount() > 0) {
+                    final CharSequence text = clip.getItemAt(0).getText();
+                    return text != null ? text.toString() : null;
+                }
+            } catch (NullPointerException e) {
+                // may happen if clipboard service throws exception while reading clipboard contents
```
fm-sys, line 48: Can't we avoid the NPE here without using try/catch?

moving-bits: There seem to be two possible causes for errors:

The "clipboard != null" check is added to catch if `getSystemService()` returns null, similar check as in `copyToClipboard()` a few lines further up.

The NPE does not seem to get triggered directly by our code. From looking at the log (see linked issue) it seems to be a "bit further down the pipe" during unparceling, that's why I added the try/catch. But maybe there's a way to check this upfront?

eddiemuc: `ClipboardManager` seems to have a method `hasPrimaryClip`. Maybe checking this one first prevents the NPE? But this is just a guess, we would need to look into the (Android) source code (which is easy in AS)

moving-bits: This is the part being called in ClipboardManager:

```
public @Nullable ClipData getPrimaryClip() {
    try {
        return mService.getPrimaryClip(mContext.getOpPackageName(), mContext.getUserId());
    } catch (RemoteException e) {
        throw e.rethrowFromSystemServer();
    }
}
```

Thanks for the pointer to `hasPrimaryClip`. I can change the PR to use this instead of the try/catch. 
I cannot say what's the actual cause for the error I found in the logs. Maybe it's device-specific (configuration, old Android version etc.) and not worth doing a try/catch for a rare occurrence.

moving-bits: Updated PR and its title accordingly

PR #15111, opened 2023-12-28
UnifiedMap: Support precise & low poper map autorotation modes

moving-bits:
## Description
Add support for "Autorotation (precise)" and "Autorotation (low power)" modes 
- "precise" = updates every 80ms / min heading diff = 1° (downgrades to default mode, if speed > 5 km/h or no speed readings)
- "low power" = updates every 500ms / min heading diff = 10°

Implemented for UnifiedMap only, other maps use "low power" mode.

@fm-sys / @eddiemuc 
I've tried to implement the different suggestions, but please cross-check if your needs are covered.


eddiemuc:
looks ok for me codewise.

moving-bits:
retest this please

fm-sys:
Looks good 👍

Didn't had time to test on a emulator, but looks fine codewise and we can fine tune afterwards if necessary.

PR #15117, opened 2023-12-29
fix codacy issues

eddiemuc:
fix codacy issues


eddiemuc:
I overlooked that Codacy did not run on the PR I just merged, thus I didn't saw these issues before.

PR #15150, opened 2024-01-03
UnifiedMap: Show waypoint if set as target

### LINKED ISSUE #15137: waypoint not shown on map
### Describe your problem!

waypoint not shown on map

### How to reproduce?

Open waypoint list of a cache
Click on compass, right of waypoint
Click on map in top line

Only symbol for cache is shown

symbol for cache and waypoints are shown

### Reproducible

Yes

### c:geo Version

2024.01.02-NB-eeec2a4

### System information

```text
Routing is shown to the waypoint, so the focus on the wayloint is kept (see #15131)
```

### Additional Information

_No response_

moving-bits:
## Description
Some fixes to UnifiedMap "set as target" mode:
- if waypoint is set as target: only display this waypoint, but do so without checking "display waypoint limit"
- if cache is set as target: only display its waypoints, not those from other caches in given viewport
- also attempts to fix #15137

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -338,26 +338,28 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                         mapFragment.setCenter(Settings.getUMMapCenter());
                     }
                     break;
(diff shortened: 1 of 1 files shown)


bekuno:
Btw., Codacy has fixed the error 500 problem for pull requests.
You can see again the PR at https://app.codacy.com/gh/cgeo/cgeo/pull-requests/open

PR #15152, opened 2024-01-03
Downloader: Optionally keep temporary files for further processing

moving-bits:
Optionally keep temporary downloaded files, but only for map sources that support this (currently OAM maps only).
Setting for this is in settings => map sources => keep temporary files (visible in extended settings mode only).


moving-bits:
retest this please

Issue #13312, opened 2022-08-06
Keyword search vs. BM?


moving-bits:
### Describe your problem!

User on support:
- search by keyword "Rävafallet", results are similar to a plain nearby search, found caches do not contain the keyword
- user is a basic member
- no caches stored offline that contain that keyword

on further testing he stored some caches containing that keyword
- now while entering the keyword he gets related caches as autocomplete suggestion
- but the search itself still delivers only results unrelated to the given keyword

Which raises two questions:
- Does search by keyword only work for PM (and if so, shouldn't we disable it or add a hint, if the user is a basic member)?
- Will/should caches stored locally be included in the search result?

### How to reproduce?

- login to c:geo as a basic member
- do a keyword search, eg by "Rävafallet"

### Actual result after these steps?

search results look like a plain "nearby" search has been performed, found caches have no specific relation to the keyword given

### Expected result after these steps?

if technically possible: search result should deliver caches containing the given keyword,

otherwise "search by keyword" should be disabled or a specific hint given, if "search by keyword" is not available to basic members

### Reproducible

Yes

### c:geo Version

2022.07.31

### System information

```text
ticket 218005
```

tobihein:
I also have the same issue with version 2023.12.21 (installed from f-droid-store).

I am also a basic member

eddiemuc:
It is quite likely that keyword search on GC.com does not work for BM because it is also not offered on the website for BM.

But we can't just remove the keyword search for BM because on other geocaching platforms (eg oc) keyword search will of course work.

tobihein:
Thanks for the explanation. I am pretty sure that this worked in the past (years ago).

In that case a small hint in the app would be nice and avoid confusion. But I am not a power-cacher so I don't know if it makes sense in all cases. 

Is it possible to check programmatically if it is a geocaching basic account?

moving-bits:
> It is quite likely that keyword search on GC.com does not work for BM because it is also not offered on the website for BM.
> 
> But we can't just remove the keyword search for users logged in as BM at GC.com. Because on other geocaching platforms (eg oc) keyword search will of course work.

A mitigation could be to check if the user has activated only geocaching.com connector + is a basic user, and if so, disable keyword search. Not 100%, but would be fine for a larger group of users.

> Is it possible to check programmatically if it is a geocaching basic account?

Yes, we do have that information.

SammysHP:
Does it work for offline stored caches, though?

eddiemuc:
[...]

PR #15162, opened 2024-01-04
Disable keyword search for gc.com basic members

moving-bits:
## Description
Mitigation for #13312 for gc.com-only users:
Disables keyword search for a user, if he/she is a gc.com basic member and only active connectors are gc.com and/or lab adventures.


moving-bits:
retest this please

Issue #12600, opened 2022-01-23
Support variables for projections


Lineflyer:
### Tell us your idea!

Now its gets totally wild ;)

In my todays cache GCKGVY a projection is needed to get from last stage to the final:
`Den Cache findest du: (I) (D) (A) Meter in (E) (H) (A) Grad`

So this feature request is to also support variables in bearing/distance for projected coordinates as a waypoint.
For sure nothing important and not to be considered mandatory for the first go live.

### What solution would you suggest?

Not sure about how it could be implemented UI wise. Probably it would need (besides the coords theirselves) additional fields for bearing and distance which support variables/formulas. 
The projection then needs to be calculated into the resulting coords as soon as the variables have valid values.

### Is there an alternative?

No.
For that particular cache I created another waypoint (Final) with same coordinate formula as for the stage upfront (stage 7) and pasted the text for the projection as a note. Still I needed to manually look up the variable values and fill them in to create the correct final waypoint.

Volthai7us:
Hi, I want to work on this issue but I need some guidance. Could you help to figure out where in the code I should go and make changes? This is my first time contributing to an open source project.

eddiemuc:
I don't consider this request to be wild at all. It would be a perfect use case for the new global variables.

That said, I don't think we should implement this right away but first think about how we want to handle Waypoint projections in general. 

As of now, waypoint projection works in a one-time-calculation way: user enters angle+distance into a waypoint, the waypoint is calculated anew and the projected calculation is saved. The projection parameters however are NOT saved.

Introducing variables into this scenario could be as simple as allowing to enter a formula into the angle+distance field. This is easy to implement. However, this would mean that calculation of angle+distance is only done one time from the vars (when they are entered) and projection is immediately calculated. The angle/distance formulas are not saved, and if the values of the vars change at a later time, the projection would not be recalculated.

I would argue that, before introducing vars into this, we should change the way we handle projections: a projection should be specified by entering a reference waypoint, an angle and a distance. Then this waypoint would recalculate itself every time the reference waypoint or angle/distance is changed. Then introducing variabes in angle/distance also makes sense.

However, this would be a bigger task, so I would like to put this up for discussion first.

eddiemuc:
> Hi, I want to work on this issue but I need some guidance. Could you help to figure out where in the code I should go and make changes? This is my first time contributing to an open source project.

[...]

PR #15165, opened 2024-01-05
rel to #12600: allow usage of variables in waypoint projection

eddiemuc:
rel to #12600: allow usage of variables in waypoint projection

### DIFF
--- a/main/project/rawimages/arrow_north_east.svg
+++ b/main/project/rawimages/arrow_north_east.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 -960 960 960" width="24"><path d="m216-160-56-56 464-464H360v-80h400v400h-80v-264L216-160Z"/></svg>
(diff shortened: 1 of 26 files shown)


eddiemuc:
This PR, if merged, will enable waypoint projections using formulas with cache variables. It contains various other refactorings as well which were "lying on the way" :-)

Since this is a first draft especially GUI-wise, the function is hidden behind a feature property called `feature_calculated_projections`. To enable, one has to manually create this property as boolean and set it to `true`.

When opening a waypoint, the familiar fields "Distance" and "Bearing in °" will be gone. Instead a new spinner will be available containing projection modes:

image
image

Choosing a projection will make necessary fields visible. Those are formula fields. A new "target coordinate" view becomes visible as well:
image

Entering values and/or formulas will immediately calculate projections (if all inputs are valid):
image
image

eddiemuc:
In waypoint list, info is shown like this (this is an example for a waypoint with both calculated coordinate AND projection):

image

eddiemuc:
retest this please

eddiemuc:
retest this please

eddiemuc:
This PR does not change anything towards the user, thus merging here. I will create a second PR to enable new logic

PR #15175, opened 2024-01-05
fix #15173: restrict formula scan regex to make evil regex situation more unlikely

### LINKED ISSUE #15173: Formula parsing on variables tab leads to ANR
User report on support and reproducible for me with current `master`:

- open cache GCAHQ78
- swipe to variables tab
- select "scan listing"

Result: ANR

Hang seems to be occurring in FormulaUtils.scanText's call to `m.find()`:

image

variable `start` gets values 0, 468, 1095, 1106, 1146 and 1162 before the hang starts.

In contrast to this, cache details page's "waypoints => scan for waypoint formulas" works instantly and gives you two results.

eddiemuc:
fix #15173: restrict formula scan regex to make evil regex situation more unlikely


eddiemuc:
Since this is a fix for release, I would appreciate a short code review before merge.

eddiemuc:
No feedback here. I guess I just merge and we all hope for the best.

PR #15176, opened 2024-01-06
refactor builder pattern LogEntry and OfflineLogEntry: make non-generic

eddiemuc:
refactor builder pattern LogEntry and OfflineLogEntry: make non-generic


eddiemuc:
Implements a better inheritance builder pattern for LogEnrty and OfflineLogEntry, results in usage of non-generic builder classes.
See https://stackoverflow.com/questions/21086417/builder-pattern-and-inheritance for details

eddiemuc:
This is a pure refactoring PR with no behavioral change. thus let's merge

PR #15177, opened 2024-01-06
fix #12600: enable waypoint projection with variables

### LINKED ISSUE #12600: Support variables for projections

eddiemuc:
fix #12600: enable waypoint projection with variables

Details about this feature can be found here: https://github.com/cgeo/cgeo/pull/15165#issuecomment-1879226154

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/activity/WaypointActivityTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/activity/WaypointActivityTest.java
@@ -4,6 +4,7 @@ import cgeo.geocaching.EditWaypointActivity;
 import cgeo.geocaching.Intents;
 import cgeo.geocaching.R;
 import cgeo.geocaching.enumerations.CacheType;
+import cgeo.geocaching.enumerations.ProjectionType;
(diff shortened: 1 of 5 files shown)


eddiemuc:
Merging this would change GUI and behaviour of waypoint projection. I received no feedback for this yet, so I hesitate with merging.

PR #15182, opened 2024-01-07
improve swipe up indicator color scheme

ztNFny:
Old

New


fm-sys:
This had annoyed me as well, but I didn't found the time to report it yet. The new version is much better IMHO. 

Big thanks for providing the PR!

PR #15190, opened 2024-01-12
UnifiedMap: Continue trail history even in hidden state (fix #15185)

### LINKED ISSUE #15185: [info] position record interrupted
### Describe your problem!

Between 21.12. and 7.1. the position recording did not work. Unfortunately I cannot say to which cgeo nightly version that correlates since I don't have an history of my updates. Normally I have the recording not visible on map, I only switch it on occasionally.
It looks like it started recording again since I activated "show position recording on map" and  I checked the settings and toggeled the position-recording-button off and on again

### How to reproduce?

Since recording is now working, not reproduceable.

No position is shown on map and in the gpx export file for the mentioned time interval.

Position recording shoud work if switched on in settings.

### Reproducible

No

### c:geo Version

2024.01.07-NB1-9bd14f4

### System information

_No response_

### Additional Information

Just for info, issue can be closed.

moving-bits:


fm-sys:
I thought not tracking the position if trail history is disabled is the intended behavior, also at our old map...?

moving-bits:
Admittedly, the behavior of showing or not showing the history track is inconsistent on old map, but it does continue to record the track. At least when:
- opening the map in "show track" mode
- then hide it
- change position multiple times
- then show track

It will then display the last movements, including those from the time where displaying the track was disabled.

Looks like recording the track depends on whether displaying it was enabled on starting the map or not. If so, this behavior would not make sense, IMHO. (But I have not verified this hypothesis in code yet.)

PR #15199, opened 2024-01-14
[cache popup] fix gap while swiping up

fm-sys:
fix gap while swiping up in cache popup.

That's how it was before the PR:

Screenshot_20240114_233826_cgeo geocaching


moving-bits (approved):
looks way better & is a cleaner implementation - thanks for fixing!

PR #15208, opened 2024-01-17
integrate external contacts addon into c:geo (fix #13777)

### LINKED ISSUE #13777: Integrate cgeo-contacts into main cgeo app
Since I recently had certain insights into the cgeo project structure, I wonder why cgeo-contacts is still a separate app/plugin instead of being integrated into the main app.

I assume that cgeo-contacts was created as a separate app in ancient times when Android didn't have Runtime Permissions (previous to SDK23/Marshmallow) and the user had to give general permission to an app for reading contacts. Is this correct?

Although I am well aware that cgeo still supports SDK21, I feel the time is ripe to re-integrate cgeo-contacts into cgeo and use runtime permissions for contacts access, It is time for SDK21/22 users to upgrade if they still want to use this feature.

Also, I feel that cgeo-contacts is in danger of being misused. This is because its intends could be used by any app on the user's phone, not just cgeo. Thus other apps may exploit cgeo-contacts to search in user's contacts. This is probably not what the user wants.  (At least, if I understood code correctly, the contact data is not returned to the calling app, so probably the risk is small. Still it gives me a chill from data privacy perspective...)

fm-sys:
see discussion in #13777, can be merged if we decide for variant 1.


moving-bits:
Will the rights for reading contacts be requested on installing c:geo, or only checked when doing some action requiring those access rights (and only then being requested if missing). We should strive for the latter, IMHO.

bekuno:
> Will the rights for reading contacts be requested on installing c:geo, or only checked when doing some action requiring those access rights (and only then being requested if missing). We should strive for the latter, IMHO.

For API 21 and 22, the rigth is requested at the time of installation.
From API 23, it is (and must implemented) at runtime when necessary.

See https://source.android.com/docs/core/permissions/runtime_perms?hl=en

bekuno:
@fm-sys Do we already request the permission in the code? I do not see it in your PR.

fm-sys:
> Will the rights for reading contacts be requested on installing c:geo, or only checked when doing some action requiring those access rights (and only then being requested if missing). We should strive for the latter, IMHO.

The permission is requested on using the "open contact card" action, not already in the installation wizard. So users will understand why they need to grant the permission, if they find that action.

The whole implementation including  requesting the permission was done by @eddiemuc in https://github.com/cgeo/cgeo/pull/13800, I only enabled the code which was already in cgeo for more than a year. 

(BTW, I think "search in contacts" would better describe the action instead of "open contact card", but that's a different topic)

fm-sys:
https://github.com/cgeo/cgeo/assets/64581222/9efbce19-a22b-42d5-927a-a0b7b29d1307

eddiemuc:
> The whole implementation including requesting the permission was done by @eddiemuc in #13800, I only enabled the code which was already in cgeo for more than a year.

Exactly. Since then I am patiently waiting for c:geo to raise minSDK to 23. But I am more than fine to integrate now :-) Thanks @fm-sys for retriggering the discussion.

moving-bits:
Any objections to merge this?

(Requests new permissions, see https://github.com/cgeo/cgeo/pull/15208#issuecomment-1896993439, but allows us to retire the contacts plugin.)

Will merge this tomorrow evening if no one contradicts.

moving-bits:
no objections, so let's merge

PR #15213, opened 2024-01-20
Fix Hylly theme download error 404 (fix #11793)

### LINKED ISSUE #11793: 404 error on Hylly map file download
Just a heads up, the internal downloader is returning a 404 error on the Hylly map file.

_Originally posted by @Vike91 in https://github.com/cgeo/cgeo/issues/10511#issuecomment-935100404_

Side note by @fm-sys:
I can reproduce the issue

moving-bits:
## Description
When map downloaded initiates a map download, it also checks for required theme file. If that cannot be found, it initiates that download as well.
For Hylly, it used a fixed url prefix for the theme file, but their website now (?) placed the files into the same subfolder as the map, so let's use that one.


moving-bits:
retest this please

Issue #14706, opened 2023-10-09
Re-installation: restore backup process not straight forward and incomplete


MagpieFourtyTwo:
### Describe your problem!

Just downgraded to yesterday's nightly (due to #14705) and had some issues with the configuration assistant in the restore process:

- after starting c:geo for the "first" time, I first had to grant access permission for the base folder (which appears to be mandatory, even if I just want to restore a backup, in order to at least get access to the backup files)
- but then, to my very surprise, I was asked by the assistant to enter my gc.com credentials (which should come with the restore, anyway, thus there should be no need to ask for, IMHO).
- in the next step I was asked for offline map configuration - and that's the screen where I found a possibility to restore a possibly existing backup for the first time (one step too late, I would say).
- So I restored it, and then ... I was confronted with the very same offline map configuration screen as I got before, although the restore seemed to be successful (although there was no success confirmation, but there were no errors neither), and again I had the options to either click ``OFFLINE MAPS`` or ``RESTORE`` - why that? I of course had the possibility to just ``SKIP``, but it's all but clear what the consequences of this ``SKIP`` will be ...

In fact, pressing ``SKIP`` and ``FINISH`` (plus granting access to photos and media) eventually did the job and c:geo started in my restored environment, but this behavior is kind of weird ... 

### How to reproduce?

Re-install c:geo and try to restore a previously performed backup (using the configuration assistant).

### Actual result after these steps?

Backup was restored, but ...

- IMHO question to restore a backup came one step too late (should be prior to asking for service credentials)
- there was no restore success message (restart just happened after another ``SKIP`` and ``FINISH`` click)
- after restore Map and theme folders were reset to default (but had to be on external SD card)
- after next start, the previously loaded polygon file could not be displayed anymore
- routing profile ``Bike`` was set to a wrong file (while ``Car`` profile file was correct - both live in the same folder on ext. SD card)
- folder for routing data was reset to default (but had to be in BRouter's segment4 folder on ext. SD card)
- GPX folder was reset to default (but had to be "GPX" in "download" folder in internal memory)
- c:geo did not ask for permission to access the needed folders on external SD card (that's perhaps why the corresponding settings have been reset to default?)

### Expected result after these steps?

[...]

PR #15217, opened 2024-01-20
Include active trackfiles in backup (rel. to #14706)

moving-bits:
## Description
Includes active track files into database backup/restore.
"active" meaning: those, which are selected for display on the map, disregarding their enabled/disabled state.
(Contents of configured GPX folder are not backed up by this feature.)

Backup/restore is performed when database gets backed up or restored, no separate selection for track files provided. (This is intentionally, as "active" track files are referenced in the database, and this data should be kept in sync.)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
@@ -14,6 +14,7 @@ import cgeo.geocaching.storage.ContentStorageActivityHelper;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.Folder;
 import cgeo.geocaching.storage.FolderUtils;
+import cgeo.geocaching.storage.LocalStorage;
 import cgeo.geocaching.storage.PersistableFolder;
 import cgeo.geocaching.storage.PersistableUri;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
@@ -55,6 +56,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.reactivex.rxjava3.schedulers.Schedulers;
 import org.apache.commons.collections4.CollectionUtils;
@@ -73,6 +75,7 @@ public class BackupUtils {
     private static final String ATTRIBUTE_VALUE = "value";
     private static final String TAG_MAP = "map";
     private static final String SETTINGS_FILENAME = "cgeo-settings.xml";
+    private static final String TRACKS_SUBFOLDER = "tracks";
 
     private static final int MAX_AUTO_BACKUPS = 4;  // most recent + x more
     private static final String AUTO_BACKUP_FOLDER = "auto"; // subfolder of PersistableFolder.BACKUP
@@ -300,7 +303,9 @@ public class BackupUtils {
         };
 
         if (database) {
-            restoreDatabaseInternal(backupDir, consumer);
+            AndroidRxUtils.andThenOnUi(Schedulers.io(),
+                () -> FolderUtils.get().copyAll(Folder.fromFolder(backupDir, TRACKS_SUBFOLDER), Folder.fromFile(LocalStorage.getTrackfilesDir()), false),
+                () -> restoreDatabaseInternal(backupDir, consumer));
         } else {
             consumer.accept("");
         }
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -300,7 +305,19 @@ public void restoreInternal(final Activity activityContext, final Folder backupD
         };
 
         if (database) {
-            restoreDatabaseInternal(backupDir, consumer);
+            final Folder backupTracksDir = Folder.fromFolder(backupDir, TRACKS_SUBFOLDER);
+            final List<ContentStorage.FileInformation> tracks = ContentStorage.get().list(backupTracksDir);
+            AndroidRxUtils.andThenOnUi(Schedulers.io(), () -> {
+                    final AtomicInteger tfCopied = new AtomicInteger(0);
+                    for (ContentStorage.FileInformation fi : tracks) {
+                        final Uri uri = ContentStorage.get().copy(fi.uri, Folder.fromFile(LocalStorage.getTrackfilesDir()), FileNameCreator.forName(fi.name), false);
```
eddiemuc, line 313: In class FolderUtils you will find methods to copy multiple files/folders at once

eddiemuc:
What about following scenario:
* User performs backup -> tracks are copied to backup dir
* User uninstalls c:geo -> internal dir is cleared, but his public "track" folder is left untouched (track files are still in it)
* User installs c:geo and restores backup -> backup track files are copied into the public "track" folder and hence doubled-up

Wouldn't it be easier just to check the public track folder (and re-copy to internal app dir) if track file is not found in app-internal dir?

main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
moving-bits: Thanks for the pointer, I'll look into it.

moving-bits:
> What about following scenario:
> * User performs backup -> tracks are copied to backup dir
> * User uninstalls c:geo -> internal dir is cleared, but his public "track" folder is left untouched (track files are still in it)
> * User installs c:geo and restores backup -> backup track files are copied into the public "track" folder and hence doubled-up
>    Or is that a false assesment?
> 
> Wouldn't it be easier just to check the public track folder (and re-copy to internal app dir) if track file is not found in app-internal dir?

Track files, which a user select for display on the map, will be copied into an app-internal folder (to make them accessible using `File`). You can delete the track's source, and displaying the track will still work. Also, there's no longer a connection between the source file and the app-internal copy, c:geo would not know whether the source is still available. Thus I think it should be included in the backup. (Typically it's just a handful of files, if at all, and those files having just a few kb in size, so it does not inflate the backup that much.)

main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
moving-bits: updated PR accordingly in both backup and restore methods

eddiemuc:
> > What about following scenario:
> > 
> > * User performs backup -> tracks are copied to backup dir
> > * User uninstalls c:geo -> internal dir is cleared, but his public "track" folder is left untouched (track files are still in it)
> > * User installs c:geo and restores backup -> backup track files are copied into the public "track" folder and hence doubled-up
> >   Or is that a false assesment?

Did you check that no "double-up" of files happens in the public folder in the above scenario? Or (if it does) is this an accepted side-effect?

moving-bits:
> Did you check that no "double-up" of files happens in the public folder in the above scenario? Or (if it does) is this an accepted side-effect?

backup/restore act with backup folder and app-internal cache folder only, public GPX folder is not involved. Backed-up track files go into `tracks` subfolder of current backup folder, thus no need to check for duplicates here.

There may be duplicates between public GPX folder and app-internal track files cache, but IMHO it's not worth comparing each file in the GPX folder with app-internal track files copy. And if we would leave out those files from backup if they are present in GPX folder (at the time of the backup), backup state would be inconsistent, if you move the backup folder to a different device without copying the GPX folder at the same time.

eddiemuc:
> There may be duplicates between public GPX folder and app-internal track files cache, but IMHO it's not worth comparing each file in the GPX folder with app-internal track files copy. And if we would leave out those files from backup if they are present in GPX folder (at the time of the backup), backup state would be inconsistent, if you move the backup folder to a different device without copying the GPX folder at the same time.

Hm. I don't like it too much, but I won't oppose it.

PR #15224, opened 2024-01-22
finalize #13777: remove cgeo-contacts from codebase

eddiemuc:
finalize #13777: remove cgeo-contacts from codebase

This PR, if merged, will remove the cgeo-contacts plugin completely from the c:geo codebase


eddiemuc:
Once this PR is merged it will naturally no longer be possible to build new releases of cgeo-contacts for Google Play Store etc. At least not from master. Once current master goes beta/prod, we would have to remove the plugin/app there

PR #15249, opened 2024-01-27
fix #15236: re-enable star on filter name for unsaved changes

### LINKED ISSUE #15236: [Nightly] Displayed filter names lost their star
### Describe your problem!

Named filters used to have a star after they were altered - this star has somehow gone lost.
What has already led to some astonishment on several occasions, because the filter seemed to give incorrect results.

### How to reproduce?

1. create a filter
2. save this filter with a (hopefully mnemonic) name
3. open map or any list
4. select the saved/named filter
5. the filter name will be shown in the header row
6. alter the filter
7. have a look at the filter name's end ...

8. the filter name display is unchanged,
   notice that there is no star at the end of it's name, which shows that the filter has been modified
   and does no longer match the saved one 

8. as soon as any criteria of the filter changes, the star should be displayed

### Reproducible

Yes

### c:geo Version

2024.01.25-NB-48ffb91

### System information

_No response_

### Additional Information

_No response_

eddiemuc:
fix #15236: re-enable start on filter name for unsaved changes

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/config/JsonConfigurationUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/config/JsonConfigurationUtils.java
@@ -70,8 +70,8 @@ public final class JsonConfigurationUtils {
         if (jsonConfig1 == null || jsonConfig2 == null) {
             return false;
         }
-        final JsonNode node1 = jsonConfig1.getJsonConfig();
-        final JsonNode node2 = jsonConfig2.getJsonConfig();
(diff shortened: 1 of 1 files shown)


MagpieFourtyTwo:
Very good! Has already caused a lot of confusion. :p Happy to see it fixed. Thx!

eddiemuc:
While fixing this bug I realized that the handling of unsaved filter changes is not handled very consistent in general (this is not new but was in the code from the beginning). One of these days this has to be cleaned up...

PR #15267, opened 2024-02-01
supportSettingFoundState - ssue #14672

baiti:
Back while we weren't aware of methodically retrieving the true found state of a LC in the context of a logged in user, we had controls to manually set the found state to either `found it` or `did not find`.  A few months ago we found appropriate  API methods to determine that state and consequently the explicit controls via the UI are no longer necessary.

This PR implements a change to turn those controls off.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
@@ -108,7 +108,7 @@ public class ALConnector extends AbstractConnector implements ISearchByGeocode, 
 
     @Override
     public boolean supportsSettingFoundState() {
-        return false;
+        return Settings.isALCfoundStateManual();
     }
 
     @Override
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
@@ -108,7 +108,7 @@ public class ALConnector extends AbstractConnector implements ISearchByGeocode, 
 
     @Override
     public boolean supportsSettingFoundState() {
-        return true;
+        return false;
     }
 
     @Override
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -921,6 +921,8 @@
     <string name="settings_ec_description">Authorize c:geo with extremcaching.com to search for caches and access/filter your found caches.</string>
     <string name="settings_su_description">Authorize c:geo with geocaching.su to search for caches.</string>
     <string name="settings_lc_description">Include Adventure Lab caches on the map.\n\n(Available to \"Geocaching.com\" premium members only. Requires an activated \"Geocaching.com\" service.)\n\nChanging this setting requires a restart.</string>
+    <string name="settings_lc_found_state_title">Found State Handling</string>
+    <string name="settings_lc_found_state_handling">If checked, do not automatically obtain the true found state for Adventure Lab caches, instead, provide controls for the user to manually set the found state</string>
     <string name="lc_default_description">This is a geocaching.com Adventure Lab Cache. Adventures can only be played with the Groundspeak Adventure Lab App. Tap on the Adventure Lab icon on the \"Details\" page to open/install the Adventure Lab App for this adventure from there.\nIf you don\'t want to see Adenture Lab caches on your map and search, you can deactivate using the \"settings\" menu - Services - Geocaching.com Adventure Lab</string>
     <string name="init_summary_oc_de">Load caches from opencaching.de</string>
     <string name="init_oc_de_description">Authorize c:geo with opencaching.de to search for caches and access/filter your found caches.</string>
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -921,7 +921,7 @@
     <string name="settings_ec_description">Authorize c:geo with extremcaching.com to search for caches and access/filter your found caches.</string>
     <string name="settings_su_description">Authorize c:geo with geocaching.su to search for caches.</string>
     <string name="settings_lc_description">Include Adventure Lab caches on the map.\n\n(Available to \"Geocaching.com\" premium members only. Requires an activated \"Geocaching.com\" service.)\n\nChanging this setting requires a restart.</string>
-    <string name="settings_lc_found_state_title">Found State Handling</string>
+    <string name="settings_lc_found_state_title">Manual found state</string>
     <string name="settings_lc_found_state_handling">If checked, do not automatically obtain the true found state for Adventure Lab caches, instead, provide controls for the user to manually set the found state</string>
     <string name="lc_default_description">This is a geocaching.com Adventure Lab Cache. Adventures can only be played with the Groundspeak Adventure Lab App. Tap on the Adventure Lab icon on the \"Details\" page to open/install the Adventure Lab App for this adventure from there.\nIf you don\'t want to see Adenture Lab caches on your map and search, you can deactivate using the \"settings\" menu - Services - Geocaching.com Adventure Lab</string>
     <string name="init_summary_oc_de">Load caches from opencaching.de</string>
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -357,7 +357,9 @@ final class ALApi {
             cache.setRating(response.get("RatingsAverage").floatValue());
             cache.setArchived(response.get("IsArchived").asBoolean());
             cache.setHidden(parseDate(response.get("PublishedUtc").asText()));
-            cache.setFound(response.get("IsComplete").asBoolean());
+            if (!Settings.isALCfoundStateManual()) {
+                cache.setFound(response.get("IsComplete").asBoolean());
+            }
             DataStore.saveCache(cache, EnumSet.of(SaveFlag.CACHE));
             return cache;
         } catch (final NullPointerException e) {
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -164,10 +164,12 @@ final class ALApi {
         try {
             final Response response = apiRequest(geocode.substring(2), null, headers).blockingGet();
             final Geocache gc = importCacheFromJSON(response);
-            final Collection<Geocache> matchedLabCaches = search(gc.getCoords(), 1, null);
-            for (Geocache matchedLabCache : matchedLabCaches) {
-                if (matchedLabCache.getGeocode().equals(geocode)) {
-                    gc.setFound(matchedLabCache.isFound());
+            if (!Settings.isALCfoundStateManual()) {
+                final Collection<Geocache> matchedLabCaches = search(gc.getCoords(), 1, null);
+                for (Geocache matchedLabCache : matchedLabCaches) {
+                    if (matchedLabCache.getGeocode().equals(geocode)) {
+                        gc.setFound(matchedLabCache.isFound());
+                    }
                 }
             }
             return gc;
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -703,6 +703,10 @@ public class Settings {
         return getBoolean(R.string.pref_alc_advanced, false);
     }
 
+    public static boolean isALCfoundStateManual() {
+        return getBoolean(R.string.pref_foundstate_al, false);
+    }
+
     public static boolean enableFeatureUnifiedGeoItemLayer() {
         return getBoolean(R.string.pref_feature_unified_geoitem_layer, false);
     }
--- a/main/src/main/res/values/preference_keys.xml
+++ b/main/src/main/res/values/preference_keys.xml
@@ -70,6 +70,8 @@
     <string translatable="false" name="preference_screen_al">preference_screen_al</string>
     <string translatable="false" name="pref_connectorALActive">connectorLCActive</string><!-- variable stays with "LC" for compatibility reasons -->
     <string translatable="false" name="pref_fakekey_al_website">fakekey_al_website</string>
+    <string translatable="false" name="pref_foundstate_al">foundstate_al</string>
+
 
     <!-- settings for opencaching.de screen -->
     <string translatable="false" name="preference_screen_ocde">preference_screen_ocde</string>
--- a/main/src/main/res/xml/preferences_services_geocaching_com_adventure_lab.xml
+++ b/main/src/main/res/xml/preferences_services_geocaching_com_adventure_lab.xml
@@ -16,6 +16,14 @@
         <Preference
             app:summary="@string/settings_lc_description"
             app:iconSpaceReserved="false" />
+        <CheckBoxPreference
+            android:defaultValue="false"
+            android:key="@string/pref_foundstate_al"
+            android:title="@string/settings_lc_found_state_title"
+            app:iconSpaceReserved="false" />
+        <Preference
+            app:summary="@string/settings_lc_found_state_handling"
+            app:iconSpaceReserved="false" />
     </PreferenceCategory>
 
     <PreferenceCategory


main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
```diff
@@ -108,7 +108,7 @@ public String getExtraDescription() {
 
     @Override
     public boolean supportsSettingFoundState() {
-        return true;
+        return false;
```
fm-sys, line 111: Thanks for working on it!

You may want to remove the method completely, as the default implementation in abstract connector also just returns false.

baiti: I thought about it but decided to leave it in, in case at some point our API methods being used don't give us the required properties anymore. In that case we could just toggle the switch.

If you want me to take it out, I'll do of course.

fm-sys: No clear preference from my side, both is fine for me. I'll wait with merging though, to give others time to comment...

baiti: thought about it some more ... I think taking it out is ok as it can be brought back, if needed, within minutes.

baiti: At first I thought, how does she do this  as it is not possible in cgeo to use different accounts for GC and LC but I believe I understand the use  case now. Let me see whether I got it right. You use user X for GC and LC enabled with cgeo. You use a user Y with GC's labs app to actually "play" the labs and their found state is correctly reflected in GCs app, however for your user Y. Now when you work with cgeo in the context of user X, you want to manually set the LCs you played for user Y with the app to a `found it`. Right?

I know of others who do the same. Primary reason being to not "pollute" their stats on GC with the massive amounts of "points" that accumulate by playing labs. Frequently they use basic accounts for playing the labs as there are almost no restrictions for basic users when playing labs with GC's app.

Having said all this, there is no need to change the existing functionality in cgeo. We could, if we wanted to, make this variable a configurable option, perhaps somewhere in advanced settings? On the other hand we're having discussions that there are already way too many setting switches ... so adding even more is perhaps not such a good idea either.

murggel: > At first I thought, how does she do this  as it is not possible in cgeo to use different accounts for GC and LC but I believe I understand the use  case now. Let me see whether I got it right. You use user X for GC and LC enabled with cgeo. You use a user Y with GC's labs app to actually "play" the labs and their found state is correctly reflected in GCs app, however for your user Y. Now when you work with cgeo in the context of user X, you want to manually set the LCs you played for user Y with the app to a `found it`. Right?
> 
> I know of others who do the same. Primary reason being to not "pollute" their stats on GC with the massive amounts of "points" that accumulate by playing labs. Frequently they use basic accounts for playing the labs as there are almost no restrictions for basic users when playing labs with GC's app.
> 

You got it right, that's exactly how and why I do this

murggel: > Having said all this, there is no need to change the existing functionality in cgeo. We could, if we wanted to, make this variable a configurable option, perhaps somewhere in advanced settings? On the other hand we're having discussions that there are already way too many setting switches ... so adding even more is perhaps not such a good idea either.

I will be fine with that setting, but would prefer the possibility to use a different user for the labs in c:geo - but I guess that would probably not be so easy

fm-sys: I'm one of the guys trying to stop every new setting. ;) 

However I'm primarily against settings which could be avoided by just finding good defaults and build better UIs. In this case it definitely makes sense to make it configurable, and the LC connector specific settings page is still pretty clean so we could add it there.

Screenshot_20240202_104214_cgeo geocaching_copy_270x585

murggel: So the option would be something like "I use a different gc-account" or " don't fetch found state from gc-account" or "..."
I am not sure which wording we should use

fm-sys: > would prefer the possibility to use a different user for the labs in c:geo

I'm not sure, can basic members access the API methods we use at all? And what would GC do if they recognize that we seem to let basic members use lab methods (which we currently don't) even if it's just the secondary account of a PM?

murggel: > > would prefer the possibility to use a different user for the labs in c:geo
> 
> I'm not sure, can basic members access the API methods we use at all? And what would GC do if they recognize that we seem to let basic members use lab methods (which we currently don't) even if it's just the secondary account of a PM? 

Good point - as I mentioned before - I can live with the setting then

moving-bits: Technically wise, removing the method from `ALConnector` would be the way to go, as @fm-sys has mentioned above.

With regard to @murggel's use-case, I propose something different, though:
- Open a separate issue for optionally using `ALConnector` with a separate user,
- and wait with removing the option to manually set found state for `ALConnector` until that feature request is implemented.

baiti: > > would prefer the possibility to use a different user for the labs in c:geo
> 
> I'm not sure, can basic members access the API methods we use at all? And what would GC do if they recognize that we seem to let basic members use lab methods (which we currently don't) even if it's just the secondary account of a PM?

To all I know, technically the APIs we're consuming also work for basic accounts exactly the way they do for premium.

The LC related API methods in the partner API, aka GC Live API are scoped at premium accounts only. Since we aren't using those we have technically no restrictions. Just like GC's Lab App, which also does not restrict basic accounts by any means. With one exception: There seem to be some recent functions in the GC Lab App which also require premium but to all I know these are more or less irrelevant.

Scoping our implementation at premium account is purely to be good citizen and to mimic the behaviour an "API partner app" would be forced to.

baiti: > Technically wise, removing the method from `ALConnector` would be the way to go, as @fm-sys has mentioned above.
> 
> With regard to @murggel's use-case, I propose something different, though:
> 
> * Open a separate issue for optionally using `ALConnector` with a separate user,
> * and wait with removing the option to manually set found state for `ALConnector` until that feature request is implemented.

I think her use case wouldn't work anymore as her "user Y" (as I called it), would certainly be a basic user.  In order to make this useful, we would have to remove some voluntary restrictions from our implementation. Which pretty much relates to the quesiton @fm-sys asked further up in this thread.

baiti: Let me add some clarification with respect to what "a user" is in the context of the AL connector: When it comes to comsuming the labs api, we are not in a logged in state, we do not have any `access_token` for a specific user. Neither today nor in the forseeable future. Most of the AL functionality in cgeo is completely independet from a GC account, except **one** subtle association and that is the one we need in order to determine the true found state of an AL. In order to retrieve that all we need to know is the `public GUID` but no authorization at all. I would consider this a "side effect" of that API method, perhaps, in the light of CCPA an even unwanted side effect, a side effect which may go away the more we talk about it the more likely.

So if we provided a way for a user to manually enter the public GUID of **any** user we would get the found state for that user. Let's link that back to the "user Y" in the example further above, the use case we discussed would be working.

Also, and to be complete: Our `consumer_key` does not support to "login" a user, at least not with the now deprecated `login` method, perhaps it would work via OAuth2, I haven't tested that but I wouldn't be surprised if that didn't work either as we are not supposed to really have a logged in context, in which, if we had it, we could actually "solve" those labs by submitting correct answers.

With all that no clarified, @moving-bits, would it make sense to extend the AL connector to have some sort of a login to a specific account? We could sort of mimic a login to that user's GC account but only in order to retrieve that `public-guid` and nothing else.

moving-bits: Thanks @baiti for the additional details. We could then implement something like this:
- checkbox for AL connector "use GC account data" (default = true)
- if unchecked: show credentials form for username/password
  - if data in this fields get changed: login with those credentials, and on successful login store guid for that "AL user" in our settings
- when doing requests against the API, either use guid for "GC user" or from "AL user", depending on setting
- don't refresh AL guid on every request, but only when credentials are changed (or when switching "use GC account data" to false)
- I'm not sure if we can reuse the existing GC login methods (but we should try to) - we need to check for possible side-effects

Lineflyer: My few cents (I hope I did not miss that part of the discussion):
Do we need to consider basic members who import lab caches via GPX? Those users should still be able to set the found state locally. Am I right?

baiti: > Thanks @baiti for the additional details. We could then implement something like this:
> 
> * checkbox for AL connector "use GC account data" (default = true)
> * if unchecked: show credentials form for username/password
>   
>   * if data in this fields get changed: login with those credentials, and on successful login store guid for that "AL user" in our settings
> * when doing requests against the API, either use guid for "GC user" or from "AL user", depending on setting
> * don't refresh AL guid on every request, but only when credentials are changed (or when switching "use GC account data" to false)
> * I'm not sure if we can reuse the existing GC login methods (but we should try to) - we need to check for possible side-effects

Yes, that would be a solution that support the scenario we're discussing here. However, assuming you all understand the scenario @murggel described, the account logging into the AL connector would very much likely be a basic account. Would that render sufficient functionality? @murggel, could you do a quick test: Logout from your premium account, loginto you "user Y" account, check whether you get the necessary functionality for user Y's AL caches. I would assume it would be sufficient for you then if you can see the dots on the map and the true found state on them. You wouldn't get much more unless we changed some of our policies which we most likely wouldn't (e.g. limiting functionality to premium accounts even though we could technically do the same for basic what we do for premium).

murggel: I will do a test this evening with my phone where I have them stored and with another one where they are not yet stored

murggel: > I will do a test this evening with my phone where I have them stored and with another one where they are not yet stored

@baiti But since a premium account for ALC is necessary - I will only see the already stored information  - correct?

baiti: I am not exactly sure how much you can see with a basic account.

baiti: I would assert: if you log into your "user Y" and you don't see any AL with it, the your use case is broken and you most likely would want to continue to use your current workflow with manually setting your found state? If that is true, it would also be true for most other cachers interested in the "user Y" approach, in essence, nobody would ever log into two accounts, e.g. user X (premium) for regular GC caches and user Y (basic) for LC objects.

murggel: But then the option "don't fetch the real found state" would make sense for my workflow

murggel: I just tried it with a fresh installation  - as expected - I don't see any lab caches with my basic account
[]()

main/src/main/res/values/strings.xml
```diff
@@ -921,6 +921,8 @@
     <string name="settings_ec_description">Authorize c:geo with extremcaching.com to search for caches and access/filter your found caches.</string>
     <string name="settings_su_description">Authorize c:geo with geocaching.su to search for caches.</string>
     <string name="settings_lc_description">Include Adventure Lab caches on the map.\n\n(Available to \"Geocaching.com\" premium members only. Requires an activated \"Geocaching.com\" service.)\n\nChanging this setting requires a restart.</string>
+    <string name="settings_lc_found_state_title">Found State Handling</string>
```
fm-sys, line 924:
```suggestion
    <string name="settings_lc_found_state_title">Manual found state</string>
```

Our preference names are typically describing the "if enabled"-functionality and only the first word is capitalized.

@baiti Despite from this, is the PR ready to merge?

fm-sys: (BTW, just noticed: pretty cool reviewing feature to make code suggestions which can directly be committed with a single click via github.com UI)

baiti: I appreciate your suggestions and will integrate those shortly, however, taking the classic approach with `git commit`, `git push`.

Wrt readiness: I'd consider it ready but let's wait till @murggel had a chance to do a test.

murggel: Unfortunately I can't test it right now, because I am waiting for @lineflyer to get the consumer key - probably he has overseen my request...
I don't know, if some one of you can give it to me.

moving-bits: @murggel 
have a look at your inbox

fm-sys: You could have done a `git pull` with the PR branch checked out, or use the equivalent button in your IDE if you prefer (sometimes also called "update project" instead of "pull")

murggel: I am currently testing ... will give feedback probably tomorrow evening

murggel:
I am currently testing ... will give feedback probably tomorrow evening

murggel:
@baiti My first observation: still toggling found-state while moving the map.
I did some debugging and found, that in method `ALAPi::parseCache` the found state is also retrieved from the server request and set in Geocache.
So there you have to do the wrap with  `if (!Settings.isALCfoundStateManual())` as well.

If you want, I can push my commit to your PR.

murggel:
With the above mentioned change everything works as expected.

BUT that means:
* with manual handling the "found" found state of the premium account is not fetched (which is on purpose). In my use-case that will be no problem, because I have them already stored with the correct found state, and that state will be kept even on refresh
* without manual handling the former manual set found state gets lost on refreshing the lab. If we decide to deactivate the manual handling per default, this could a problem for those user, who are not aware of having the state set manually and are refreshing the lab (on purpose or because it is in a list / map range)

baiti:
> @baiti My first observation: still toggling found-state while moving the map. I did some debugging and found, that in method `ALAPi::parseCache` the found state is also retrieved from the server request and set in Geocache. So there you have to do the wrap with `if (!Settings.isALCfoundStateManual())` as well.
> 
> If you want, I can push my commit to your PR.

Yes, please do and thanks much

murggel:
> > @baiti My first observation: still toggling found-state while moving the map. I did some debugging and found, that in method `ALAPi::parseCache` the found state is also retrieved from the server request and set in Geocache. So there you have to do the wrap with `if (!Settings.isALCfoundStateManual())` as well.
> > If you want, I can push my commit to your PR.
> 
> Yes, please do and thanks much

done

murggel:
the failing test [cgeo.geocaching.CgeoApplicationTest.testSearchTrackableSpottedLogState](https://ci.cgeo.org/job/cgeo-CI_PR-build/7706/testReport/junit/cgeo.geocaching/CgeoApplicationTest/testSearchTrackableSpottedLogState/) is not related to my changes 
@moving-bits / @fm-sys   seems, that the cacheGuid has changed

image

moving-bits:
@murggel 
Would you say "go" for squash & merge this PR?

murggel:
> @murggel 
> Would you say "go" for squash & merge this PR?

Probably one point to discuss, but this can be done outside of this PR as I mentioned in my above comment:

If we decide to deactivate the manual handling per default, this could a problem for those user, who are not aware of having the state set manually and are refreshing the lab (on purpose or because it is in a list / map range). I am not sure, if we can provide a good solution here (keep found state on "merging" the cache (but only for labs?

fm-sys:
> who are not aware of having the state set manually and are refreshing the lab

Isn't this the case already today?

murggel:
> > who are not aware of having the state set manually and are refreshing the lab
> 
> Isn't this the case already today? 

Ups - You are correct, I think I missed that 🙈 didn't checked current implementation 😅
That was probably my assumption  / wish

murggel:
> @murggel 
> Would you say "go" for squash & merge this PR?

Yes, since my concern affects current implementation as well - see comment above

moving-bits:
Ok, then let's test this in the nightlies

Issue #15100, opened 2023-12-26
UnifiedMap: Black boxes and other artifacts


Lineflyer:
### Describe your problem!

This strange behavior is like 90% reproducible on my device:
Once I select "Enable live" on the map of a single cache some map elements get black and stay like that.

Additionally you might notice in the video below, that the zoom "jumps" once I enable live.

### How to reproduce?

- Open cache details
- Long tap compass rose to open cache map
- Select "Enable live" from menu

### Actual result after these steps?

Video:
https://github.com/cgeo/cgeo/assets/949669/ba2eaaee-0216-433a-96d5-b41078e1f37d

### Expected result after these steps?

Map should stay at same zoom and all map elements should stay visible

### Reproducible

Unclear

### c:geo Version

2023.12.26-NB

### System information

```text
Android 13
Germany OAM v5
Elevate theme
Map design: City
```

devemux86:
There were reports for multiple maps (not supported) or not properly destroy / re-initialize map (e.g. in fragments):
- https://github.com/mapsforge/vtm/issues/281

moving-bits:
Also reported on support, ticket 498670

moving-bits:
probably related to #15108

moving-bits:
> There were reports for multiple maps (not supported) or not properly destroy / re-initialize map (e.g. in fragments):
> 
>     * [Map renders with black street labels, stops rendering, using two map instances on Android mapsforge/vtm#281](https://github.com/mapsforge/vtm/issues/281)

@devemux86 
We use a single map fragment for VTM only.

But maybe we are missing to call some lifecycle method?
- We call mMapView.xxx for xxx = { `onResume()`, `onPause()` and `onDestroy()` }
- On switching map providers we call `mMap.clearMap()`

devemux86:
@moving-bits 

If it helps, there is a fragment sample in vtm-android-example.
Look for the FragmentActivity and [MapFragment](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/MapFragment.java) classes.

moving-bits:
Thanks for the pointer. I've tried to align the two example files to our implementation, and they look mostly similar (fragment calling the lifecycle methods as described above). One significant difference is, that we recreate the fragment in `onCreate()`, but even if I let it recreate only if there has been no (VTM) fragment before, behavior does not seem to change.

I have been able to reproduce the black boxes using the following way (in the emulator):
[...]

PR #15268, opened 2024-02-02
Test basic VTM (similar to MapFragment example)

moving-bits:
Do not merge / PR just for reference, see issue #15100

Mimics VTM's example "MapFragment"
(Requires "Berlin (OAM).map" file to be present in configured map folder.)

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -390,7 +390,7 @@ dependencies {
 
     // Mapsforge VTM Snapshot from official master branch (via https://jitpack.io/#mapsforge/vtm)
     String mapsforgeVTMSource = 'com.github.mapsforge.vtm'
-    String mapsforgeVTMVersion = '7879086e56' // LineBucket: transparent lines (9d68bdb) + dropDistance (31646b3)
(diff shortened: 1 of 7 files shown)


ztNFny:
With this one I get strange non-black artifacts:
image

latest Germany OAM

moving-bits:
thanks for cross-checking. I had this kind of artifacts as well when using the VTM.DEFAULT theme - may or may not be related to the cross-hatch issue (which is part of #15100 and got solved meanwhile by the change in #15310).

But no black labels for street names etc., no blinking when using this version?

ztNFny:
> But no black labels for street names etc., no blinking when using this version?

No, but I fear that doesn't mean much as with "normal" builds I get black when enabling live which isn't possible here.

moving-bits:
Ok, I've updated this PR to include the "transparent line" fixes for VTM and checks for germany (oam).map as well as Elements.zip theme for easier testing. I don't get black boxes anymore with @murggel's test case (https://github.com/cgeo/cgeo/issues/15100#issuecomment-1939449817)

moving-bits:
with getting positive result from #15411 so far, we can close this WIP

PR #15273, opened 2024-02-02
Harmonize map scale slider range

### LINKED ISSUE #14965: Reduce map scale slider range
Currently it's possible to scale the map from 10% to 500%.

I've played a bit with these values and think that 50% to 250% would be more than enough. Reducing this range would have the advantage that it gets easier to select a specific value.

Edit: just checked, and cache markers can be scaled from 50% to 200%. Maybe we should align these ranges, and 50-200% should also be enough for map scaling.

fm-sys:
Reduce map scale slider range to what we allow for cache markers. (fix #14965)


moving-bits:
Maybe target to `release` to have it included in our next beta?

PR #15286, opened 2024-02-04
fix #15276: show scrollbar on long item selection lists

### LINKED ISSUE #15276: Suport scrollbar in list selection dialog
User request on support:

This user has > 100 lists defined. When using list selection, only a couple of them are shown (depending on screen size). The user knows, that he can swipe in this list.

Still, the user is missing the scrollbar, as it would be an indicator for him where in this list of list name he is - eg. "first third", "middle" etc.

Thus the user is requesting to add a visible scrollbar to the list selection dialog.
(This might also be a good visual indicator for other users that do not know they can swipe in such a long list.)

ticket 575947

eddiemuc:
fix #15276: show scrollbar on long item selection lists


moving-bits:
Shall we target this to `release` to have it included in next beta version?

eddiemuc:
Closing this PR in favor of #15293 (which is identical but with target `release`)

Issue #15296, opened 2024-02-05
Crash on opening cache with waypoint with calculated projected coordinates


murggel:
### Describe your problem!

When I try to open a cache with a waypoint, which has calculated projected coordinates, c:geo crashes, if the variable has no value (?).
The waypoint was exported to the personal note.

### How to reproduce?

Create waypoint with calculated projection, don't enter a value for the variable / delete the value of the variable.
Export the waypoint to personal note.
Sometimes it crashes already here.
Close the cache.
Open the cache

Crash

### Reproducible

Yes

### c:geo Version

2024.02.05-NB-e45a320

### Additional Information

[logcat-2024-02-05-21-17-49-2.txt](https://github.com/cgeo/cgeo/files/14171604/logcat-2024-02-05-21-17-49-2.txt)

eddiemuc:
Changes proposed in PR https://github.com/cgeo/cgeo/pull/15299 might solve this.

However, this was created from a mobile device and I have no way of testing it. So please don't merge.

eddiemuc:
Tested PR #15299, it fixes the issue

PR #15299, opened 2024-02-06
rel to #15296: might fix crash on projection formulas with unfilled vars

eddiemuc:
rel to #15296: might fix crash on projection formulas with unfilled vars


eddiemuc:
Retest this please

eddiemuc:
Retest this please

PR #15301, opened 2024-02-08
show images linked in "personal note" in Images tab

ztNFny:
Recognizes image URLs in personal note and add them to images tab, treating them the same as Listing / Log images.

image

- Image URLs are recognized using a regex.
  - Because of this the URL must contain .(jpeg|jpe|jpg|png|webp|gif|svg)
  - Performance shouldn't be an issue as notes are limited to 2500 chars on GC, longer notes could appear in c:geo but are unlikely
- I added a new category "Personal Note" to images tab
  - not sure whether that is needed or they should go into the existing "Own", but the images in that category are editable from the images tab, so they are different in behaviour.
  - The category is placed below Own, before Listing, as I think those images are more likely to be needed than listing images

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/Geocache.java
+++ b/main/src/main/java/cgeo/geocaching/models/Geocache.java
@@ -2231,6 +2231,7 @@ public class Geocache implements IWaypoint {
     @NonNull
     public Collection<Image> getImages() {
         final List<Image> result = new LinkedList<>(getSpoilers());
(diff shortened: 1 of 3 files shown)


moving-bits:
Would you like to have this included in our next beta version? Then it should be targeted against `release` as we are in beta version cycle already. (Next beta probably to be due in a couple of days.)

ztNFny:
Master is fine. PC is off for the next few days, thus i can't rebase anyway.

Issue #15029, opened 2023-12-15
[Unified Map][VTM] "Edges" of routes/polylines produce darker regions, also when zoomed out


eddiemuc:
(Originally discussed in #15003)

Corresponding issue in VTM: https://github.com/mapsforge/vtm/issues/1082 (with example code in PR https://github.com/mapsforge/vtm/pull/1081)

It seems that drawing polylines (e.g. Routes) with VTM with a certain thickness and a  semi-transparent color leads to "darker edges" when those line segments are overlapping on polyline points. This effect is also visible when zooming out.

To visualize the issue I deliberately selected a very broad route width and compared Unified GMap with Unified VTM:

GMap:
![image](https://github.com/cgeo/cgeo/assets/6909759/7271a3c9-7469-461f-92bf-ec445131470d)

VTM:
![image](https://github.com/cgeo/cgeo/assets/6909759/81c3cee8-fc0c-40f1-8824-69b732f68ce0)

eddiemuc:
A mean example for my reference. Polyline of:
[N 53° 32.004' · E 010° 04.190', N 53° 32.514' · E 010° 05.445', N 53° 32.247' · E 010° 05.461', N 53° 32.463' · E 010° 05.568', N 53° 32.203' · E 010° 05.650', N 53° 32.436' · E 010° 05.019']

![image](https://github.com/cgeo/cgeo/assets/6909759/f5537f41-cd8a-468e-b9b6-37940243047c)

![image](https://github.com/cgeo/cgeo/assets/6909759/f2379963-a9bc-4333-a905-b159fe4bfe5b)

eddiemuc:
@devemux86 I modified the vtm-android-example to include an example of this. You will find the PR here: https://github.com/mapsforge/vtm/pull/1081 

It also contains some explanations and screenshots.

This might be a very dumb config error on our side, but I don't know how to proceed here. I would be glad for some pointers...

eddiemuc:
@eddiemuc please see my answer in the PR.

_Originally posted by @devemux86 in https://github.com/cgeo/cgeo/issues/15003#issuecomment-1857774454_

eddiemuc:
@MagpieFourtyTwo / team: it looks like this is a bigger issue and not resolved easily. Is this a showstopper for UnifiedMap/VTM for now?

eddiemuc:
[...]

PR #15310, opened 2024-02-10
rel to #15100: deactivate VTM transparent line rendering to solve strange artefacts

eddiemuc:
rel to #15100: deactivate VTM transparent line rendering to solve strange artefacts


moving-bits:
Let's put this in the nightlies to cross-check for the other types of artifacts (black boxes instead of labels and/or icons).

Reopen #15029

ztNFny:
Unfortunately this doesn't solve it either. I get less black blocks but now also some flickering ones.

moving-bits:
> Unfortunately this doesn't solve it either. I get less black blocks but now also some flickering one

I've added this observation to the tracking issue (#15100) for documentation purposes

Issue #15247, opened 2024-01-26
New cache pop up is crashing back to map


wolverine007:
### Describe your problem!

The new cache pop up box is crashing when being opened via swipe up.
Cgeo is shortly displaying the details and then going back to map. 

### How to reproduce?

open any cache by selecting it on the map

### Actual result after these steps?

Cache details are displayed shortly, then app is going back to map

### Expected result after these steps?

Show cache details like being done by clicking the according button. 

### Reproducible

Yes

### c:geo Version

2024.01.24-RC

### System information

```text
## System information

c:geo version: 2024.01.24-RC

Device:
-------
- Device type: 23049PCD8G (marble_eea, POCO)
- Available processors: 8
- Android version: 13
- Android build: TKQ1.221114.001 test-keys
- Screen resolution: 1080x2176px (392x791dp)
- Pixel density: 2.75
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 24.02.13 (190400-599553831)
- Memory:  Available:1,6 GB, Total:7,0 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: active
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: orientation

Program settings:
-------
- Settings: v9, Count:281
- Set language: de
- System date format: dd.MM.yy
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: never
- Last auto backup: 24. Jan., 09:32
- Routing mode: Bike
- Live map mode: true
- Use unified map: false
- OSM multi-threading: true / threads: 4
- Map: Deu+nord (FZK) (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMapSource:primary:cgeo/maps/Deu+nord (FZK).map
  - Atts: (c) Map: FZK project (free for private use); Map data: OpenStreetMap contributors; Contour data: U.S. Geological Survey and J. de Ferranti
  - Theme: fzk-outdoor-soft-v5.zip:fzk-outdoor-soft-v5/fzk-outdoor-soft-v5.xml

Filters:
-------
- Hide waypoints: visited
[...]

PR #15311, opened 2024-02-10
fix sheetRemoveFragment crash #15247

ztNFny:
Check FragmentManager state before removing the Fragment AND use commitNow() to force the commit to happen instantly. With commit() it's enqueued and when it finally happens the FragmentManager might already be saved.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
@@ -154,13 +154,20 @@ public abstract class AbstractNavigationBarMapActivity extends AbstractNavigatio
         }
         clearSheetInfo();
         final FragmentManager fm = getSupportFragmentManager();
+        final FragmentTransaction ft = fm.beginTransaction();
         final Fragment f1 = fm.findFragmentByTag(TAG_MAPDETAILS_FRAGMENT);
         if (f1 != null) {
-            fm.beginTransaction().remove(f1).commit(); // #15247: consider commitAllowingStateLoss() instead of commit() ?
+            ft.remove(f1);
+            if (!fm.isStateSaved()) {
+                ft.commitNow();
+            }
         }
         final Fragment f2 = fm.findFragmentByTag(TAG_SWIPE_FRAGMENT);
         if (f2 != null) {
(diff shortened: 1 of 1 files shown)


moving-bits (commented):
Are multiple commits to the same `FragmentTransaction` actually ok?

[Docs](https://developer.android.com/guide/fragments/transactions) seem to state otherwise: "The final call on each `FragmentTransaction` must commit the transaction. The commit() call signals to the `FragmentManager` that all operations have been added to the transaction."

ztNFny:
> Are multiple commits to the same `FragmentTransaction` actually ok?
> 
> [Docs](https://developer.android.com/guide/fragments/transactions) seem to state otherwise: "The final call on each `FragmentTransaction` must commit the transaction. The commit() call signals to the `FragmentManager` that all operations have been added to the transaction."

Your right, seems there's only 1 commit per transaction. The second check prevents any expectation from being raised.
But probably the key to solving the issue is commitNow() instead of commit(). And if i see it right commitNowAllowingStateLoss() would probably be the ideal solution as that one does the check internally.

eddiemuc:
Fragment commit concept seems kind of broken when you need 4 different kinds of commit to make it work...

If I understand the solution correctly we use the commit with the "lowest implications available" (synchronous and not checking for state loss) and that makes the solution work somehow. Ok for me to go this way, but in some way our usage of fragments seems to be flawed. 

As @ztNFny mentioned already: why is the sheet-removal-method post-delayed anyway?

eddiemuc (approved):

ztNFny:
> As @ztNFny mentioned already: why is the sheet-removal-method post-delayed anyway?

Without the delay pop-up is removed first, "empty" map is shown for a moment, only the CacheDetailsActivity starts.

I think the "proper" way would be merging Popup amd Details into 1 activity, but that's a big rewrite.

moving-bits:
This fix needs to go to `release`. I'll try to retarget this PR using GitHub.

moving-bits:
Ok, retargeting does not work - too many differences. @ztNFny: Do you want to create a new PR using `release` as base, or shall I step in?

ztNFny:
Please do, i won't be able for a few days.

moving-bits:
> Please do, i won't be able for a few days.

done with #15314

moving-bits:
superseded by #15314, so let's close here

PR #15314, opened 2024-02-11
Try to fix sheetRemoveFragment crash (rel. to #15247)

moving-bits:
## Description
Tries to fix the crash on info sheet removal by force-commiting and allowing state-loss.
Related to (and hopefully fixes) issue #15247.

Closes #15311 (for a short discussion see over there)


moving-bits:
Let's try this in the nightlies

PR #15316, opened 2024-02-11
rel to #15029: set vtm transparent attribute on a per-line basis

eddiemuc:
rel to #15029: set vtm transparent attribute on a per-line basis

This PR upgrades VTM to latest version on VTM master, in order to use new "per-line" transparency flag (see VTM  https://github.com/mapsforge/vtm/pull/1100)


moving-bits:
Works like a charm, thanks @devemux86 and @eddiemuc for the fix!

PR #15323, opened 2024-02-12
Add "contained in individual route" filter (fix #11532)

### LINKED ISSUE #11532: Individual Route filtering
### Tell us your idea!

Sometimes it would be helpful if the individual route could be listed in our cache list activity.

Add a new filter "Contained in Individual Route" with the options "yes" or "no". 

### Is there an alternative?

1. Add it to the status filter instead (But it is way too long already today, see #11288)
2. Add a "show as list" feature to individual route
3. Add a new "special list" like "All" or "History"

### Additional Information

_No response_

fm-sys:
Implements a "contained in individual route" filter, like suggested in #11532.

I've abstracted a generic `BooleanGeocacheFilter`, so that we can potentially split up our way too long status filter in a next step.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterType.java
+++ b/main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterType.java
@@ -23,6 +23,7 @@ public enum GeocacheFilterType {
     DIFFICULTY_TERRAIN_MATRIX("difficulty_terrain_matrix", R.string.cache_filter_difficulty_terrain_matrix, R.string.cache_filtergroup_details, DifficultyTerrainMatrixGeocacheFilter::new),
     RATING("rating", R.string.cache_filter_rating, R.string.cache_filtergroup_details, RatingGeocacheFilter::new),
     STATUS("status", R.string.cache_filter_status, R.string.cache_filtergroup_basic, StatusGeocacheFilter::new),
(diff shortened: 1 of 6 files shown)


main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterType.java
```diff
@@ -23,6 +23,7 @@ public enum GeocacheFilterType {
     DIFFICULTY_TERRAIN_MATRIX("difficulty_terrain_matrix", R.string.cache_filter_difficulty_terrain_matrix, R.string.cache_filtergroup_details, DifficultyTerrainMatrixGeocacheFilter::new),
     RATING("rating", R.string.cache_filter_rating, R.string.cache_filtergroup_details, RatingGeocacheFilter::new),
     STATUS("status", R.string.cache_filter_status, R.string.cache_filtergroup_basic, StatusGeocacheFilter::new),
+    IndividualRoute("individualroute", R.string.cache_filter_individualroute, R.string.cache_filtergroup_userspecific, IndividualRouteGeocacheFilter::new),
```
moving-bits, line 26: all the other filter names are written in caps

fm-sys: You are right, interesting that codacy didn't complain. I thought there was a "Enum names must be uppercase" check somewhen...

PR #15326, opened 2024-02-12
Add 'go to' deprecation notice

moving-bits:
## Description
As discussed in yesterday's dev meeting: Add a deprecation notice for "go to" functionality.


fm-sys:
We won't delete the goto cache itself, right? So maybe it would be better to only show the notice when calling via the deprecated link...?

moving-bits:
I had tried that first, showing the dialog in `QuickLaunchItem`. But when called via `AbstractNavigationBarActivity` it opens the dialog and immediately crashed, noting a "activity has leaked a WindowDecor" (or something like this), though the given activity is still valid (eg: `MainActivity`, still running).

moving-bits:
Alternatively we could use a one-time message to not annoy those who use ZZ0 regularly (but with the risk that some people may not read the notice and just click it away - but we can't take care of every situation probably)

moving-bits:
Rebased + changed message to a OneTimeMessage to be less-intrusive

moving-bits:
failing test is unrelated, PR #15379 not yet aplied

moving-bits:
retest this please

PR #15329, opened 2024-02-13
UnifiedMap VTM: Don't reset tilesource on enabling live mode

moving-bits:
## Description
Workaround related to #15100 ("black boxes" in @murggel's use case)

@ztNFny @murggel 
Can you please retest your "black boxes" use-cases with this build?
It's up to current `release` branch with those few changes seen below, and for me the black boxes do not appear in the described use-case any longer.


moving-bits:
Let's try it in the nightlies

PR #15335, opened 2024-02-15
fix #15333: geojson display on unifiedmap

### LINKED ISSUE #15333: Geojson display different between NewMap and UnifiedMap
Geojsons are drawn very different on UnifiedMap. Is this by intention? I find the "old" display style to be better suited than the old. But it's not really a bug, more of a question - if it's as designed.
@moving-bits 

[de.geojson.txt](https://github.com/cgeo/cgeo/files/14285709/de.geojson.txt)

ztNFny:
Align display of geojson between Google Maps and OSM.

#15333 mentions 2 issues:

detail-level of the polygons:
I addressed this by setting dropDistance() to 1 instead of 2

display-style of polygons:

- strokeWidth() was divided by 2, I removed that division. The removed division leads to track lines being twice as wide now though (see screenshot below), seems they don't get the same width set as before?
- fill alpha was applied twice. fillAlpha() applies it from the fill color, but that already includes the alpha (caluclated in GeoJsonUtils.colorFromJson()), so we get alpha = alpha * alpha.
Either fillColor() needs to get only the rgb component of the argb value or - what I did in the current state of the PR or set fillColor() to the value without alpha:
.fillAlpha(Color.aToFloat(fillColor))
.fillColor(Color.setA(fillColor, 255))

Not sure which of those is prefered.
 

After:
image
image

Before:
image
Google Maps (unchanged):
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
@@ -5,7 +5,6 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.GeopointConverter;
 import cgeo.geocaching.models.geoitem.GeoIcon;
-import cgeo.geocaching.models.geoitem.GeoItem;
 import cgeo.geocaching.models.geoitem.GeoPrimitive;
 import cgeo.geocaching.models.geoitem.GeoStyle;
 import cgeo.geocaching.models.geoitem.ToScreenProjector;
@@ -123,7 +122,7 @@ public class MapsforgeVtmGeoItemLayer implements IProviderGeoItemLayer<Pair<Draw
     public Pair<Drawable, MarkerInterface> add(final GeoPrimitive item) {
 
         final int fillColor = GeoStyle.getFillColor(item.getStyle());
-        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle()) / (item.getType() == GeoItem.GeoType.POLYGON ? 1.5f : 2f);
+        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle()) / 1.5f;
         final Style style = Style.builder()
                 .strokeWidth(ViewUtils.dpToPixelFloat(rawStrokeWidth))
                 .strokeColor(GeoStyle.getStrokeColor(item.getStyle()))
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.GeopointConverter;
 import cgeo.geocaching.models.geoitem.GeoIcon;
+import cgeo.geocaching.models.geoitem.GeoItem;
 import cgeo.geocaching.models.geoitem.GeoPrimitive;
 import cgeo.geocaching.models.geoitem.GeoStyle;
 import cgeo.geocaching.models.geoitem.ToScreenProjector;
@@ -24,7 +25,6 @@ import java.util.concurrent.locks.ReentrantLock;
 
 import org.oscim.android.canvas.AndroidBitmap;
 import org.oscim.backend.canvas.Bitmap;
-import org.oscim.backend.canvas.Color;
 import org.oscim.backend.canvas.Paint;
 import org.oscim.core.GeoPoint;
 import org.oscim.core.Point;
@@ -123,12 +123,11 @@ public class MapsforgeVtmGeoItemLayer implements IProviderGeoItemLayer<Pair<Draw
     public Pair<Drawable, MarkerInterface> add(final GeoPrimitive item) {
 
         final int fillColor = GeoStyle.getFillColor(item.getStyle());
-        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle());
+        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle()) / (item.getType() == GeoItem.GeoType.POLYGON ? 1.5f : 2f);
         final Style style = Style.builder()
                 .strokeWidth(ViewUtils.dpToPixelFloat(rawStrokeWidth))
                 .strokeColor(GeoStyle.getStrokeColor(item.getStyle()))
                 .fillAlpha(1f) // GeoJsonUtils.colorFromJson() already calculates the color using fill and fill-opacity, don't apply it again
-                //.fillAlpha(Color.aToFloat(fillColor))
                 .fillColor(fillColor)
                 .transparent(true) ////See #15029. Following parameter prevents rendering of "darker edges" for overlapping semi-transparent route parts
                 .dropDistance(ViewUtils.dpToPixelFloat(1)) //see #15029. This setting stops rendering route parts at some point when zooming out
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
@@ -125,12 +125,13 @@ public class MapsforgeVtmGeoItemLayer implements IProviderGeoItemLayer<Pair<Draw
         final int fillColor = GeoStyle.getFillColor(item.getStyle());
         final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle());
         final Style style = Style.builder()
-                .strokeWidth(ViewUtils.dpToPixelFloat(rawStrokeWidth) / 2f)
+                .strokeWidth(ViewUtils.dpToPixelFloat(rawStrokeWidth))
                 .strokeColor(GeoStyle.getStrokeColor(item.getStyle()))
-                .fillAlpha(Color.aToFloat(fillColor))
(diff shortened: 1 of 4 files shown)


eddiemuc (approved):

ztNFny:
@eddiemuc I just added a modification to the OP, you probably didn't see that one, sorry:
"The removed division leads to track lines being twice as wide now though (see screenshot below), seems they don't get the same width set as before?"

ztNFny:
> * strokeWidth() was divided by 2, I removed that division. The removed division leads to track lines being twice as wide now though (see screenshot below), seems they don't get the same width set as before?

I added an (ugly?) workaround for that. Doing it upstreams seems the better way but that'd be changes in many places with potentially much impact.

main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
```diff
@@ -123,14 +123,14 @@ private int getZLevel(final GeoPrimitive item) {
     public Pair<Drawable, MarkerInterface> add(final GeoPrimitive item) {
 
         final int fillColor = GeoStyle.getFillColor(item.getStyle());
-        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle());
+        final float rawStrokeWidth = GeoStyle.getStrokeWidth(item.getStyle()) / (item.getType() == GeoItem.GeoType.POLYGON ? 1 : 2);
```
eddiemuc, line 126: This one I really don't like. I don't see this fix so urgently needed that we should apply a hack. We should take the time to investigate what is happening further up in the code and apply a real fix.

fm-sys: Our fist goal should be that geoitems are rendered consistent for all map types. If we need a factor to get consistent rendering it's totally fine. But 1 distance unit should be the same width for all map implementations and geoitem types.

ztNFny: >  If we need a factor to get consistent rendering it's totally fine. But 1 distance unit should be rendered as the same width for all map implementations and geoitem types.

Consistency is exactly what this PR implements with the conditional factor.

Screenshots show a styled geojson (Duisburg), a unstyled geojson (Cologne) and a route on unified/old OSM/Google Maps.
There's still a slight different between Unified Google/OSM, but that's due to dropDistance() only being applied to OSM (should it be applied to Google as well?)

Screenshot_20240215_201502
Screenshot_20240215_202355
Screenshot_20240215_201644
Screenshot_20240215_201510

ztNFny:
Pushed another commit to adjust the default fill of polygons to match old map:
Screenshot_20240215_205535
Screenshot_20240215_201502

main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeVtmGeoItemLayer.java
eddiemuc: Yes but it is a hack to implement consistency exactly in the reported bug case - and will most likely cause inconsistentcy in other cases (needs to be throughly tested).

The Geo layer is designed such that any width in a GeoStyle should be strictly interpreted as dp. That means:
* any code using the Geo layer needs to make sure that given widths are in do
* An implementation of a Geo layer for a specific map provider (VTM, Google, mapsforge) needs to make sure to convert given values such that it is drawn in the wanted dp width

An implementation using different factors for different shapes (poly line vs polygon) can't possibly achieve this consistency.

I know that VTM is more tricky than plain Google API, but let's please find consistent solutions.

ztNFny: > and will most likely cause inconsistentcy in other cases (needs to be throughly tested).

Do you have anything in mind not yet covered by the screenshots above? I can check it then.
Before this PR Geojson rendering is inconsistent VTM Google vs VTM OSM vs old maps. I'm trying to make them look the same everywhere and this is the best fit for all the cases i could think off.

Applying a random factor is certainly not nice (why i called it "ugly" in my original comment), but i couldn't find any better place to fix it to get the same result across all maps. Seems the pipeline to get things into Google vs OSM is still quite different, even in VTM.

eddiemuc: I would suggest to do the following:

Draw a polyline next to a polygon using the same line width. You might use the `GeoItemTestLayer `class to do so (I created it for the purpose to try out things like that). Then view the result across Google and VTM and check:
* is the line width rendered the same for both objects?
* is the line width rendered the same for both map types?
* does line width behave as expected when zooming in or out?

If you find that different factors for different types (polygon, polyline) need to be used in VTM then your solution is the correct fix. Because the different geoitemlayer-implementations are then consistent to each other.

If not then the fault lies on the input side, although I could not explain why line width would differ across map types in this case.

ztNFny: > Draw a polyline next to a polygon using the same line width. You might use the `GeoItemTestLayer `class to do so (I created it for the purpose to try out things like that). Then view the result across Google and VTM and check:

Did that - and you're right, polygon and polyline width are different now (with the different factors). However I realized that all elements but the polygon are to thin on OSM vs Google:
Screenshot_20240216_091312
Screenshot_20240216_091344

So I changed the factor from 2f to 1.5f globally and am now getting a consistent rendering OSM vs Google vs Old:
Screenshot_20240216_091944
Screenshot_20240216_091912

BUT .. the navigation/history lines are now wrong: 75 (OSM) vs 63 (Google). And this is what I compared against initially (why I introduced the second factor). So I guess that's the real issue that needs to be fixed in regards to the factor.

I'll update this PR with the change of the factor from 2f to 1.5f, leaving the wide OSM navigation line as the todo-item (because I don't have the time to look into that today).

ztNFny:
As this PR has grown beyond its original scope I'll summarize its content up to now:

- OSM: Don't apply alpha to polygon fill twice. Reason: Alignment Google vs OSM
- OSM: Change dropDistance() to 1 to make polygons (a bit) more accurate when zooming out. Reason: Alignment Google vs OSM
- OSM: Change line width division factor from 2 to 1.5. Reason: Alignment Google vs OSM vs Old Map
- Change default polygon fill alpha. Reason: Alignment Unified vs Old

Todo:
- c:geos lines (history, navigation) widths are different OSM vs Google vs old map
OSM unified: 74
Google unified: 62
OSM old: 42
Google old: 42

moving-bits:
IMHO all fixes for UnifiedMap should go to `release` currently to make the best version of UnifiedMap available for next beta / next release

eddiemuc:
> As this PR has grown beyond its original scope I'll summarize its content up to now:
> 
> * OSM: Don't apply alpha to polygon fill twice. Reason: Alignment Google vs OSM
> * OSM: Change dropDistance() to 1 to make polygons (a bit) more accurate when zooming out. Reason: Alignment Google vs OSM
> * OSM: Change line width division factor from 2 to 1.5. Reason: Alignment Google vs OSM vs Old Map
> * Change default polygon fill alpha. Reason: Alignment Unified vs Old
> 
> Todo:
> 
> * c:geos lines (history, navigation) widths are different OSM vs Google vs old map
>   -- OSM unified: 74
>   -- Google unified: 62
>   -- OSM old: 42
>   -- Google old: 42
>   (To test this I recommend setting the line width to a very large value in Appearance Settings as any factor applied will become more apparent then)

I moved the TODOs into their own issue: #15343

eddiemuc:
> IMHO all fixes for UnifiedMap should go to `release` currently to make the best version of UnifiedMap available for next beta / next release

@ztNFny will you be able to rebase to release, or should I create a (new) PR with same changes for release?

ztNFny:
I can do it tonight. With the todos in a new issue thus is ready to merge

fm-sys:
@eddiemuc if you cannot wait you can easily checkout the branch and run a `git rebase --onto upstream/release upstream/master`. (I recommend to put this command on everyone's Git cheat sheet)

The you can retarget the PR via the Github UI. Just did the second step, but didn't had a PC at hand to actually do the rebase.

fm-sys:
(The commit history now looks broken, as it shows all commits  which are not yet part of release branch. Running the git command above will finalize the rebase.)

eddiemuc:
> @eddiemuc if you cannot wait you can easily checkout the branch and run a `git rebase --onto upstream/release upstream/master`. (I recommend to put this command on everyone's Git cheat sheet)
> 
> The you can retarget the PR via the Github UI. Just did the second step, but didn't had a PC at hand to actually do the rebase.

No need to rush. And I am always happy if I can stay away from git mumbo jumbo :-)

eddiemuc:
> (The commit history now looks broken, as it shows all commits which are not yet part of release branch. Running the git command above will finalize the rebase.)

I changed back to `master` for now to removes the "brokenness" :-) 

(Sersiously I don't understand why GitHub does not do the rebase udner the hood when you change target branch)

fm-sys:
> Sersiously I don't understand why GitHub does not do the rebase udner the hood when you change target branch

Well, this is intentional. Would you really want that Github removes commits based on some magic without your agreement? And rebasing is quite easy, really nothing more than pasting the cammand and hit enter :-)

eddiemuc:
> > Sersiously I don't understand why GitHub does not do the rebase udner the hood when you change target branch
> 
> Well, this is intentional. Would you really want that Github removes commits based on some magic without your agreement? And rebasing is quite easy, really nothing more than pasting the command from above (or something suitable for the use case) and hit enter :-) 

Could be made optional on branch change. Just rebasing commits of PRs onto new branch. If I understand you correctly this would be very easy and foolproof. Anyway, I am not the product manager for GitHub :-)

ztNFny:
Rebase done. Interestingly I couldn't rebase, there were many conflicts git couldn't resolve. Had to create a new branch based on release and cherry-pick those commits.

PR #15337, opened 2024-02-15
reorder layers (fixes #15334)

### LINKED ISSUE #15334: Track layer displayed on top of cache layer with OSM map source
On OSM maps (both online / offline) the track layer is displayed on top of the cache icon while with Google Maps and the old map implementations it's drawn between map and cache icon layers.

OSM:
image
Google Maps
image

Version: Todays master

ztNFny:
reordered some UnifiedMap layers

This affects only OSM for some reason, not sure how Google Maps layers are layered??

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
@@ -10,13 +10,13 @@ public class LayerHelper {
 
     public static final int ZINDEX_SEARCHCENTER = 13;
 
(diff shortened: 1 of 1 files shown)


moving-bits (approved):
LGTM, thanks for fixing

moving-bits (changes requested):
but it needs to go to `release`, as we should include this fix in the next beta

ztNFny:
> but it needs to go to `release`, as we should include this fix in the next beta

Done

moving-bits:
retest this please

moving-bits (approved):

PR #15361, opened 2024-02-20
current navigation method icon shouldn't be larger than text

ztNFny:
The icon for the currently selected navigation method in large view is overly large, adjust it to text height

Before
Screenshot_20240220_104218

After
Screenshot_20240220_104611


moving-bits:
retest this please

Issue #14759, opened 2023-10-25
(only pedestrian) routing lost - how to recover or is it a bug?


itsbrody:
### Describe your problem!

I'm often walking for caches, that means I use the pedestrian routing profile and this worked very well.
As preparation for a "far from home trip" I downloaded map and routing information to do the same abroad (cache walks in the city). But when I tried to route "pedestrian" in the city this did not work. Also updating the routing information did not help. I thought that all routing capability has lost, although all settings seem ok - it also does not differ between cgeo internal and brouter routing.

When now back in my home location I accidently switched to "bike routing" and surprisingly that works - the same as car routing. Only "pedestrian routing" is still lost and I don't know, how to recover or reset it.

### How to reproduce?

select a cache and switch in fast settings between the different routing capabilities.

none, bike, car are working
pedestrian does same as none.

### Actual result after these steps?

pedistrian  shows a straight line between my position and the selected cache. Bike, Car a routed line to the cache

### Expected result after these steps?

also ped. routing should show a "route".

### Reproducible

Yes

### c:geo Version

2023.09.26

### System information

```text
d.externalstorage.documents/tree/primary%3Acgeo/document/primary%3Acgeo%2Ffield-notes, Av:true, files:81, dirs:0, totalFileSize:110,8 KB, free space: 4,3 GB, files on device: 3194880)
  - SPOILER_IMAGES: /cgeo/GeocachePhotos (Default)[/cgeo/GeocachePhotos[PERSISTABLE_FOLDER(BASE)#1:p-content://com.android.externalstorage.documents/tree/primary%3Acgeo::/GeocachePhotos]] (Uri: content://com.android.externalstorage.documents/tree/primary%3Acgeo/document/primary%3Acgeo%2FGeocachePhotos, Av:true, files:0, dirs:0, totalFileSize:0 B, free space: 4,3 GB, files on device: 3194880)
  - ROUTING_BASE: /cgeo/routing (Default)[/cgeo/routing[PERSISTABLE_FOLDER(BASE)#1:p-content://com.android.externalstorage.documents/tree/primary%3Acgeo::/routing]] (Uri: content://com.android.externalstorage.documents/tree/primary%3Acgeo/document/primary%3Acgeo%2Frouting, Av:true, files:13, dirs:1, totalFileSize:644,0 MB, free space: 4,3 GB, files on device: 3194880)
[...]

Issue #15264, opened 2024-01-31
Opening a cache from link in trackable logbook does not work


Lineflyer:
### Describe your problem!

Reported on support mail for our beta version 2024.01.24-RC and reproducible for me on `master`:
Once the user taps on a geocache name shown in the logbook of a trackable it leads to a "cache not found" toast

### How to reproduce?

- Open GC7540G (or directly open any trackable you know and continue from there (skip next two steps then)
- Swipe to trackable tab
- Open any trackable (which has some logs with reference to caches)
- Tap on a geocache name

### Actual result after these steps?

Toast "The requested cache could not be found"

### Expected result after these steps?

Cache details should be loaded/opened

### Reproducible

Yes

### c:geo Version

2024.01.24-RC

moving-bits:
This is actually a bug on their website:

Every cache in the logbook links to https://www.geocaching.com/geocache/GC0 (which links to a cache named "GC0", which is non-existent).

(I've dropped them a note.)

moving-bits:
they have fixed that bug meanwhile, so let's close here

Lineflyer:
Really...tried yesterday and just now and still had the same issue.

moving-bits:
...okay... 
GC had reported back as "fixed" a few weeks ago upon request from us, and I haven't got any report on this on support since weeks.

But:

Just checked the website with one of my TB => links seem to be fine now
Cross-checked with today's nightly in c:geo => error message 😠 

So it may be a different cause now, need to investigate. (But strange that we did not hear anything about it on support yet.)

moving-bits:
from a short debug session:
seems that in `TrackableLogsViewCreator.fillCountOrLocation()` we do have a gc code in the guid field, and a blank gccode field, which leads to c:geo trying to open a cache by guid, but with a gc code as guid.

@eddiemuc 
IIRC you have adapted c:geo to the trackable website changes a few months ago, can you have a look at it?

eddiemuc:
I'll look into it as time permits

eddiemuc:
Fixed with PR #15475

PR #15366, opened 2024-02-22
Internal routing: Don't throw exception on error while reading rawtrack

moving-bits:
## Description
see discussion in https://github.com/cgeo/cgeo/issues/14759#issuecomment-1958864817


moving-bits:

PR #15368, opened 2024-02-22
UnifiedMap: Reload caches/waypoints in onResume (fix #15341)

### LINKED ISSUE #15341: UnifiedMap: New user-defined cache does not show up immediately
Field note:

When creating a new UDC on UnifiedMap, it will not show up immediately, but only after closing and reopening the map.

moving-bits:
## Description
`NewMap` does a complete reload of all layers, including `StoredOverlay` in `onResume()`. `UnifiedMap`doesn't do this, but on the downside needs to be notified of new UDC.


moving-bits:

fm-sys:
> NewMap does a complete reload of all layers, including StoredOverlay in onResume(). UnifiedMapdoesn't do this, 

Maybe we should add this instead? I think it's nearly impossible to catch all cases...

moving-bits:
> > NewMap does a complete reload of all layers, including StoredOverlay in onResume(). UnifiedMapdoesn't do this,
> 
> Maybe we should add this instead? I think it's nearly impossible to catch all cases...

changed this PR and its title accordingly

PR #15370, opened 2024-02-23
Keep closed elevation chart from popping up again (fix #15047)

### LINKED ISSUE #15047: Route elevation chart keeps popping up on return to c:geo
### Describe your problem!

This one is not clearly reproducible but kept happening all day today while caching (but not yesterday in same scenario).
After leaving c:geo map and coming back the elevation chart for my individual route shows up without tapping the route.  

I repeated the scenario described below for all caches on my trail and it happend in approx. 70% of cases.

### How to reproduce?

- Had a list of approx. 20 caches
- Mapped the list
- Created an individual route along this cache trail
- Using function "Remove from route after cache is logged"
- Using function "Set destination to route start"
- Approaching cache while on map
- Opening bottom sheet and invoke external navigation (in my case: Radar)
- Radar app is opened, c:geo in background
- Press back in radar app to close it
- Returning to c:geo map

Bottom sheet is dismissed (see other issue) but immediately after return to c:geo the route elevation chart appears and I need to manually close it.

No unexpected appearance of elevation chart

### Reproducible

No

### c:geo Version

2023.12.18-NB

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Let UnifiedMap remember that elevation chart got closed + check this state before automatically reopen chart for individual route


moving-bits:

PR #15375, opened 2024-02-24
fix #15194, fix #15363: autogroup list by name part, add list icon to list spinner

### LINKED ISSUE #15194: Support "subfolders" / sublists
### Tell us your idea!

A user on support asks for support of subfolders (sublists), which they want to use for easier organization of their tours. Basically this is meant for a better visual representation, not for changes in filtering.

Example:
- List A
  - List A1
  - List A2
- List B
- List C

List A1/A2 would become visible when tapping on List A.
Filtering for List A would only list caches which are assigned to List A, not necessarily those being stored in List A1 and/or List A2 as well (except they are stored also in List A).

- support a hierarchical relationship between lists
- list selection / list spinner: visual representation for a list having sublists could be a small "+" before or after the list name. Tapping on a "+" symbol would display the sub-lists (with some indentation) and "+" changes to a "-" symbol. Tapping on that "-" symbol would remove all sub-lists from display
- when filtering for caches of a certain list or when displaying a cache list for a certain list would ONLY include caches stored in chosen list, not those stored in any of its sublists (so no changes to filtering involved)
- we would need a clever way to manage the hierarchical relationship (indent / outdent; lists within same hierarchy level would still be sorted alphabetically)
- maybe more...

### Is there an alternative?

leave it as is, do not support sublists

### Additional Information

ticket 192556

### LINKED ISSUE #15363: Add Default/Custom Symbol To Each List-View
### Tell us your idea!

At the moment the custom symbol per list is only visible inside the dialog "Add cache to list":

If you are selecting the stored list section, and opening the dropdown view there, no symbols are displayed anymore:

As well the symbols are missing inside the move cache view:

Add the already assigned custom symbols to all views, were the user is interacting with lists.

### Is there an alternative?

No.

### Additional Information

_No response_

eddiemuc:
fix #15194, fix #15363: autogroup list by name part, add list icon to list spinner

This PR tries to resolve the mentioned feature request by auto-grouping all stored list names after the name part before the first found `-` character. This may e.g. look like this:

image

With groups deflated:
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -804,6 +804,8 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         markerId = newListMarker;
         MapMarkerUtils.resetLists();
         adapter.notifyDataSetChanged();
+        refreshSpinnerAdapter();
+        refreshCurrentList();
     }
 
     private void setPreventAskForDeletion(final boolean prevent) {
--- a/main/src/main/java/cgeo/geocaching/CacheListSpinnerAdapter.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListSpinnerAdapter.java
@@ -1,6 +1,8 @@
 package cgeo.geocaching;
 
 import cgeo.geocaching.list.AbstractList;
+import cgeo.geocaching.list.StoredList;
+import cgeo.geocaching.ui.TextParam;
 
 import android.view.LayoutInflater;
 import android.view.View;
@@ -55,7 +57,7 @@ class CacheListSpinnerAdapter extends ArrayAdapter<AbstractList> {
         }
 
         final AbstractList list = getItem(position);
-        holder.title.setText(list.getTitle());
+        TextParam.text(list.getTitle()).setImage(StoredList.UserInterface.getImageForList(list)).applyTo(holder.title);
         if (list.getNumberOfCaches() >= 0) {
             holder.subtitle.setVisibility(View.VISIBLE);
             holder.subtitle.setText(cacheListActivity.getCacheListSubtitle(list));
--- a/main/src/main/java/cgeo/geocaching/list/StoredList.java
+++ b/main/src/main/java/cgeo/geocaching/list/StoredList.java
@@ -116,14 +116,8 @@ public final class StoredList extends AbstractList {
                     .setSelectAction(TextParam.id(R.string.cache_list_select_last), () -> lastSelectedListSet)
                     .setChoiceMode(SimpleItemListModel.ChoiceMode.MULTI_CHECKBOX)
                     .setItems(lists)
-                    .setDisplayMapper((item) -> TextParam.text(item.getTitleAndCount()))
-                    .setDisplayIconMapper((item) -> {
-                        if (item instanceof StoredList && ((StoredList) item).markerId > 0) {
-                            return ImageParam.emoji(((StoredList) item).markerId, 30);
-                        }
-                        return ImageParam.id(R.drawable.ic_menu_list);
-                    })
                     .setSelectedItems(selectedListSet);
+            setListDisplay(model);
 
             SimpleDialog.of(activityRef.get()).setTitle(TextParam.id(titleId))
                 .setNegativeButton(null)
@@ -148,8 +142,8 @@ public final class StoredList extends AbstractList {
             final SimpleDialog.ItemSelectModel<AbstractList> model = new SimpleDialog.ItemSelectModel<>();
             model
                 .setItems(lists)
-                .setDisplayMapper((item) -> TextParam.text(item.getTitleAndCount()))
                 .setChoiceMode(SimpleItemListModel.ChoiceMode.SINGLE_PLAIN);
+            setListDisplay(model);
 
             SimpleDialog.of(activityRef.get()).setTitle(titleId).selectSingle(model, item -> {
                         if (item == PseudoList.NEW_LIST) {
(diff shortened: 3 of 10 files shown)


eddiemuc:
Also fixes #15363:
image

eddiemuc:
failing test is unrelated

moving-bits:
> This PR tries to resolve the mentioned feature request by auto-grouping all stored list names after the name part before the first found `-` character. This may e.g. look like this:

interesting approach, like it!

How does it react to an entry having a certain prefix as the only entry - will it still create a group for it? Or in other words: May you get single-item groups? (My personal preference would be to avoid single-item groups and just to display this item instead.)

eddiemuc:
> How does it react to an entry having a certain prefix as the only entry - will it still create a group for it? Or in other words: May you get single-item groups? (My personal preference would be to avoid single-item groups and just to display this item instead.)

Good point. As of now it would create single-item-groups in this case, but I can fix that

moving-bits:
> Fixed now:

great, thanks for adding this!

eddiemuc:
fainling test is unrelated

eddiemuc:
Retest this please

ztNFny:
Suggestion: Group icons.
As the groups are auto-generated maybe use the first lists icon?

ztNFny:
> by auto-grouping all stored list names after the name part before the first found `-` character.

"-" is not a suitable seperation char imho. I now have several groups like "Nordrhein" (for Nordrhein-Westfalen), "Sachsen" (for Sachsen-Anhalt), ...
Maybe : instead?

eddiemuc:
I propose to change the PR from using - to using : and move any further discussion to the original issue (instead of this PR)

ztNFny:
Is it by intention that there's no grouping in the list shown from toolbar (spinner?)?

eddiemuc:
> Is it by intention that there's no grouping in the list shown from toolbar (spinner?)?

Is that possible?

ztNFny:
Can you sort History and All caches list at the bottom of the list again please, below groups.

ztNFny:
In general sorting needs additional work imho. Order (in list selector) is currently 1. "stored", 2. ungrouped lists, 3.special lists (all, history), 4. Groups.
I'd suggest it to ne 1. "Stored", 2. Grouped and ungrouped lists, just sort them alphabetically, 4 special lists.

moving-bits:
while I like that [sorting proposal](https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963056726), we would then need to find a way to distinguish between lists that belong to a group and those that do not. Currently, all lists don't have an indentation. If we intermix grouped and ungrouped lists, an ungrouped list following a group would be not identifiable as such.

ztNFny:
> while I like that [sorting proposal](https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963056726), we would then need to find a way to distinguish between lists that belong to a group and those that do not. Currently, all lists don't have an indentation. If we intermix grouped and ungrouped lists, an ungrouped list following a group would be not identifiable as such.

Indentation would be nice, but I read from the comment in https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963052138 that not everything is possible everywhere. 
As the full name is shown for all lists the name is the identifiable criteria or whether it's grouped already.

eddiemuc:
> > while I like that [sorting proposal](https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963056726), we would then need to find a way to distinguish between lists that belong to a group and those that do not. Currently, all lists don't have an indentation. If we intermix grouped and ungrouped lists, an ungrouped list following a group would be not identifiable as such.
> 
> Indentation would be nice, but I read from the comment in https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963052138 that not everything is possible everywhere. 
> As the full name is shown for all lists the name is the identifiable criteria or whether it's grouped already.

Indentation would be doable with some extension to the simple dialog grouping model.

https://github.com/cgeo/cgeo/pull/15375#issuecomment-1963052138 refers to usage of grouping in spinner dropdowns. Also it is a question, not a statement. That one could be solved by changing it from a spinner to a simple dialog, if this is wanted?

Wouldn't have thought that this topic got so big, but I like it :-)

eddiemuc:
> In general sorting needs additional work imho. Order (in list selector) is currently 1. "stored", 2. ungrouped lists, 3.special lists (all, history), 4. Groups.
> I'd suggest it to ne 1. "Stored", 2. Grouped and ungrouped lists, just sort them alphabetically, 4 special lists.
> 

Possible, would need some rework on grouping model in simple dialog

eddiemuc:
> With changing it to simply dialog, then the filtering would also be possible, correct? Which is currently not the case in the spinner pull down.

Correct

eddiemuc:
> The spinner already uses an array adapter with custom views as well. So no need to switch to a dialog for enabling grouping.

Does `ArrayAdapter` support grouping (and probably text filtering)?

fm-sys:
You can (IIRC) use any adapter, e.g. the one you already use in simple dialog.

eddiemuc:
> You can (IIRC) use any adapter, e.g. the one you already use in simple dialog.

Nope, that one is an adapter for `RecyclerView`. AFAIK you can't use those on `Spinner` or `ListView`.
Also it is designed to interact closely with `SimpleItemListView` (where it is also defined) and not meant for external usage. Some things I wouldn't even know how to implement in a `Spinner`'s DropDown-List e.g. how to tell it to interpret taps on group headers not as item selection but as "extend/compress" signal or where to place a text search field.

ztNFny:
> That one could be solved by changing it from a spinner to a simple dialog, if this is wanted?

As the dialog offers features that spinner can't offer (search, grouping?) imho using the dialog would be a gain. In fact already today I prefer to long-tap on "lists" in bottom nav over the spinner as I prefer that dialog.

moving-bits:
let's merge it to get it into the nightlies
any extensions wrt to icons can be done in a separate PR then

Issue #15359, opened 2024-02-19
Invalid drawable added to LayerDrawable! Drawable already belongs to another owner but does not expose a constant state.


moving-bits:
Found several times in a logfile on support:

```
02-19 19:59:04.599 25927 25927 W LayerDrawable: Invalid drawable added to LayerDrawable! Drawable already belongs to another owner but does not expose a constant state.
02-19 19:59:04.599 25927 25927 W LayerDrawable: java.lang.RuntimeException
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.graphics.drawable.LayerDrawable$ChildDrawable.<init>(LayerDrawable.java:1857)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.graphics.drawable.LayerDrawable$LayerState.<init>(LayerDrawable.java:1977)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.graphics.drawable.LayerDrawable.createConstantState(LayerDrawable.java:174)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.graphics.drawable.LayerDrawable.mutate(LayerDrawable.java:1782)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.widget.ImageView.applyImageTint(ImageView.java:739)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at android.widget.ImageView.setImageTintList(ImageView.java:661)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at androidx.core.widget.ImageViewCompat$Api21Impl.setImageTintList(ImageViewCompat.java:125)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at androidx.core.widget.ImageViewCompat.setImageTintList(ImageViewCompat.java:53)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at cgeo.geocaching.ui.GeoItemSelectorUtils.setViewValues(GeoItemSelectorUtils.java:167)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at cgeo.geocaching.ui.GeoItemSelectorUtils.createGeocacheItemView(GeoItemSelectorUtils.java:57)
02-19 19:59:04.599 25927 25927 W LayerDrawable: 	at cgeo.geocaching.ui.GeoItemSelectorUtils.createRouteItemView(GeoItemSelectorUtils.java:133)
[...]

PR #15376, opened 2024-02-24
rel to #15359: try to make drawable mutatable

eddiemuc:
rel to #15359: try to make drawable mutatable


eddiemuc:
This PR won't work. I will close it

PR #15379, opened 2024-02-25
Fix TB spotted test

moving-bits:
## Description
Looks like GC has fixed the "TB spotted" links on the TB page (*), so add GC code to the now failing test.

(*) not sure if only on some machines or already system-wide, haven't received a reply from GC yet, thus keeping the related issue open for now.


moving-bits:
Let's merge it to get default build back green instead of hiding the actual PR-related errors.

PR #15381, opened 2024-02-25
Fix lint: Remove two c:geo contacts resources no longer used

moving-bits:
Removes two resources formerly used by c:geo integration for contacts plugin


moving-bits:
Retest this please

PR #15382, opened 2024-02-25
Use only constants for db table names outside DataStore.java

moving-bits:
## Description
Outside its originating defining file (`DataStore.java`) refer to database table names only by means of predefined constants


eddiemuc:
While this is a perfectly valid change (no objections into merging it), it raises the question whether the same needs to be done for column names as well. (the fact that only table names were defined as constants and not column names was the reason I did not use these constants in the first place).

moving-bits:
Valid point, I'll add them as well

PR #15385, opened 2024-02-25
respect "Show routing menu" setting

ztNFny:
The setting "Show routing menu" (enabled by default) allows users to disable the long-tap menu on map, thus adding/removing caches to the route quickly. On unified map that setting is ignored for all caches that have a waypoint (= for most caches) due to https://github.com/cgeo/cgeo/pull/14993
This PR makes unified map respect the users choice in all cases.


moving-bits:
retest this please

PR #15390, opened 2024-02-27
Add URL to JakeDot in contributors

jakedot:
See title


fm-sys:
Nice URL ;)

bekuno:
This adress is fine as IDN.

But I propose to not use it in our international project, and instead stay to only use adresses with ASCII symbols (ACE).
This would be `xn--4ca.at` here.
Or use directly the forwarded adress (*@geocaching.com) or the github address.

jakedot:
I will change the domain to jakedot.net or if one of the team could do it for me, I'm on vacation right now.

Thanks for your consideration.

eddiemuc:
> I will change the domain to jakedot.net or if one of the team could do it for me, I'm on vacation right now.
> 
> My plan was to redirect ã.at to my GitHub profile.
> 
> Thanks for your consideration.

Done via GitHub app on my mobile.

By mistake I directly committed to master instead of a new branch + PR Sorry for that. But I figure the change is so small that I don't need to revert?

PR #15406, opened 2024-03-02
add "Edit Personal Note" button to CachePopup

ztNFny:
add a "Edit personal note" button to CachePopup for quick access to editing the note.
As all the "storing the note and handling updates" stuff is in CacheDetailActivity, not in EditNoteDialog, an intermediate CacheDetailActivity is started and finished.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -263,12 +263,14 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         String name = null;
         String guid = null;
         boolean forceWaypointsPage = false;
+        boolean forceEditPersonalNote = false;
 
         if (extras != null) {
             geocode = extras.getString(Intents.EXTRA_GEOCODE);
             name = extras.getString(Intents.EXTRA_NAME);
             guid = extras.getString(Intents.EXTRA_GUID);
             forceWaypointsPage = extras.getBoolean(EXTRA_FORCE_WAYPOINTSPAGE);
+            forceEditPersonalNote = extras.getBoolean(EXTRA_EDIT_PERSONALNOTE);
         }
 
         // When clicking a cache in MapsWithMe, we get back a PendingIntent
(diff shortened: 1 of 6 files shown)


Lineflyer:
Only some remarks:
- Isn't that the spot, where the "hint" button appears if a cache hint is present? Is this handled in some way?
- AFAIR the "hint" button is also shown if a PN is present and on button-press it is then shown on the popup. Shouldn't this "edit note" button then be shown next to that personal note?

ztNFny:
> * Isn't that the spot, where the "hint" button appears if a cache hint is present? Is this handled in some way?

There are several buttons that might be shown next to details: WIG, Lab, Hint - and now this one

> * AFAIR the "hint" button is also shown if a PN is present and on button-press it is then shown on the popup. Shouldn't this "edit note" button then be shown next to that personal note?

I have some problems with that: 
- Already todays display of note in hint is kinda unintuitive, it'd be even more strange to have to tap "show HINT" first to then be able to tap "edit NOTE"
- Hint/Note layout would end up quite strange
- 2 taps instead of 1, the second of which would have to find its target first - not very user friendly for an operation probably used frequently.

ztNFny:
> When is this button supposed to show up?

Only condition is connector.canAddPersonalNote(cache), which for GC translates to a Premium account. Maybe your c:geo didn't properly sign in and thus premium features are not enabled?

moving-bits:
> > When is this button supposed to show up?
> 
> Only condition is connector.canAddPersonalNote(cache), which for GC translates to a Premium account. Maybe your c:geo didn't properly sign in and thus premium features are not enabled?

ah, good point - I was probably using the wrong emulator. Thanks

fm-sys:
Basic members can add personal cache notes inside cgeo as well. Should we really hide the button for them?

ztNFny:
> Stumbled upon a small glitch in dark mode: (wrong icon tinting)

Good catch, fixed

> Basic members can add personal cache notes inside cgeo as well. Should we really hide the button for them?

I got that from some existing edit note function, didn't think too much about it. But you're right, that restriction doesn't make much sense - removed.

PR #15408, opened 2024-03-02
use BitmapCompat for scaling log images

ztNFny:
use BitmapCompat instead of Bitmap for scaling (log) images as it provides much higher quality images (that are also smaller)

Before:
temporary_image_

After:
temporary_image


moving-bits:
retest this please

PR #15409, opened 2024-03-03
 UnifiedMap VTM: make map scales global (fix #15059)

### LINKED ISSUE #15059: UnifiedMap: Applying text/icon scaling behaves erratic
_Based upon discussion in #15035:_

> > > ...except that text/symbol scale factors seems not to be applied yet. But that could be a yet-missing feature of c:geo's VTM implementation
> > 
> > 
> > They are working but you need to restart cgeo to apply them. There actually should be a third one, however it was causing problems. You can read more about it here: #13593
> 
> Restart did not do anything. I applied a text scale of 450% but text sizes are still pretty "normal". (using openandromaps with velocity theme). I didn't try any other maps/themes. Probably there is a bug hidden here?

Did a few more tests and was able to get scaling with all themes installed in this emulator (Default, Elevate/Elements, FZK and Velocity/Voluntary), although the behavior was NOT consistent: I had to restart the map sometimes, and/or open and close theme options menu (without changing anything in between).

Additionally changes to text size / icon size settings in theme options do not necessarily get applied immediately on closing the theme options menu.

Setting "Prio: Low" as I regard this a minor issue - probably one would not change this option that frequently, but set it once, and then forget about it.

moving-bits:
## Description
- re-adds userScale for VTM, and makes text scale / symbol scale global (instead of theme-specific)
- settings are in settings => appearance => unified map
- changing any of those three settings forces an app restart on closing the settings activity
- setting gets applied in CGeoApplication

### DIFF
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -2781,6 +2781,7 @@
     <string name="mapserver_hylly_themes_info">Detailed map theme, built for \"Hylly\" maps</string>
     <string name="brouter_info">Offline routing service</string>
     <string name="maptheme_elevate_categoryinfo">Relevant for: [A] Areas, [P] POI, [R] Routes, [W] Ways</string>
+    <string name="maptheme_vtm_scaling_summary">Scaling options for offline maps (require restart)</string>
     <string name="maptheme_scale_map_title">Map Scale Factor</string>
     <string name="maptheme_scale_map_summary">Scales map; 100% = no scaling</string>
     <string name="maptheme_scale_text_title">Text Scale factor</string>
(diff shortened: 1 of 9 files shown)


main/src/main/res/values/strings.xml
```diff
@@ -2783,6 +2783,7 @@
     <string name="maptheme_elevate_categoryinfo">Relevant for: [A] Areas, [P] POI, [R] Routes, [W] Ways</string>
     <string name="maptheme_scale_map_title">Map Scale Factor</string>
     <string name="maptheme_scale_map_summary">Scales map; 100% = no scaling</string>
+    <string name="maptheme_scale_map_summary_vtm">Scale map (OpenStreetMap offline maps only), 100% = no scaling (requires restart)</string>
```
fm-sys, line 2786: I think we can shorten it a bit

```suggestion
    <string name="maptheme_scale_map_summary_vtm">Scale offline maps (requires restart)</string>
```

moving-bits:
I've updated the PR to make all three VTM scaling options global, see settings => appearance => Unified Map

image

(The headings with their individual capitalizations are taken from today's theme settings, we should fix that sometime.)

main/src/main/res/values/strings.xml
moving-bits: (different approach taken meanwhile, therefore marking as resolved here)

PR #15411, opened 2024-03-03
UnifiedMap: recycle map fragment on calling sub activity (rel. to #15100)

### LINKED ISSUE #15100: UnifiedMap: Black boxes and other artifacts

moving-bits:
## Description
yet another stab at #15100: (updated description)
- destroy map fragment in `onPause` and recreate on returning to calling map
- alternatively (configurable by setting feature key, see further down): destroy map fragment only if a sub-map is requested, and restart activity after returning to calling map
- so far no differentiation between GoogleMaps and VTM, though only the latter seems to need this cycle. can be added later, if this PR actually fixes #15100

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -707,6 +707,14 @@ public class Settings {
         return getBoolean(R.string.pref_feature_unified_geoitem_layer, false);
     }
 
+    public static boolean isFeatureEnabledDefaultFalse(@StringRes final int featureKeyId) {
+        return getBoolean(featureKeyId, false);
+    }
+
+    public static boolean isFeatureEnabledDefaultTrue(@StringRes final int featureKeyId) {
+        return getBoolean(featureKeyId, true);
+    }
+
     public static String getALCLauncher() {
         return getString(R.string.pref_alc_launcher, "");
     }
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -105,6 +105,7 @@ import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.widget.PopupMenu;
 import androidx.lifecycle.ViewModelProvider;
 
+import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -152,10 +153,13 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
     private boolean overridePositionAndZoom = false; // to preserve those on config changes in favour to mapType defaults
 
     private Spinner mapSpinner = null;
+    private UnifiedMapState lastMapStateFromOnPause = null;
+    private static WeakReference<UnifiedMapActivity> unifiedMapActivity = new WeakReference<>(null);
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        acquireUnifiedMap(this);
 
         HideActionBarUtils.setContentView(this, R.layout.unifiedmap_activity, true);
 
@@ -506,6 +510,9 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
     }
 
     private void updateCacheCountSubtitle() {
+        if (mapFragment == null) {
+            return;
+        }
         final int cacheCount = mapFragment.getViewport().count(viewModel.caches.getValue().getAsList());
         String subtitle = res.getQuantityString(R.plurals.cache_counts, cacheCount, cacheCount);
 
@@ -642,6 +649,9 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
     }
 
     private void initFollowMyLocation(final boolean followMyLocation) {
+        if (mapFragment == null) {
+            return;
+        }
         Settings.setFollowMyLocation(followMyLocation);
         ToggleItemType.FOLLOW_MY_LOCATION.toggleMenuItem(followMyLocationItem, followMyLocation);
 
@@ -1144,7 +1154,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
         super.onSaveInstanceState(outState);
         outState.putBundle(STATE_ROUTETRACKUTILS, routeTrackUtils.getState());
 
-        outState.putParcelable(BUNDLE_MAPSTATE, getCurrentMapState());
+        outState.putParcelable(BUNDLE_MAPSTATE, lastMapStateFromOnPause);
         outState.putParcelable(BUNDLE_MAPTYPE, mapType);
         if (mapType.filterContext != null) {
             outState.putParcelable(BUNDLE_FILTERCONTEXT, mapType.filterContext);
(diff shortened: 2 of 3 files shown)


eddiemuc (commented):
Won't lie to you: I think this is an incerdibly ugly solution which undoubtly will cause us much pain in the future.

Does it at least solve the observed VTM problems?

moving-bits:
Thanks, @eddiemuc, for the honest words. (Totally fine for me, BTW, that's why I asked for a review from you.)

I've just been on a short tour with this build, using c:geo mostly for navigation (individual route, auto-rotating map) and cache details, and it worked flawlessly. You don't get any performance penalty from fragment destroying as long as you don't open a second map. But there are remaining issues, see https://github.com/cgeo/cgeo/issues/15100#issuecomment-1975136202.

Regarding the "ugly": Which part of this PR do you mean - the "destroy fragment" part or the "on demand" part, or both?

If destroying the fragment won't help we will probably have to recreate the activity instead, this will be even more costly, so an "on demand" variant would be good to have.

eddiemuc:
> Regarding the "ugly": Which part of this PR do you mean?

The general idea to keep a (weak) reference to a unifiedmapactivity in cgeoapplication

Anyway, if it works it works. Can't argue with that :-)

moving-bits:
> > Regarding the "ugly": Which part of this PR do you mean?
> 
> The general idea to keep a (weak) reference to a unifiedmapactivity in cgeoapplication

I need a reference for the delayed callback to destroy the map fragment. I chose a `WeakReference` to not disturb the garbage collector. If the original activity got destroyed for some reason (and the map fragment with it), the reference returns null.

> Anyway, if it works it works. Can't argue with that :-)

Sadly, it doesn't work reliably (though it did on my short tour), I have been able to provoke more black boxes and other distortions on my test device. It's so frustrating...

main/src/main/java/cgeo/geocaching/CgeoApplication.java
```diff
@@ -166,4 +169,13 @@ private static String getApplicationHashkey(final Context context) {
         return sb.toString();
     }
 
+    public static void acquireUnifiedMap(final UnifiedMapActivity newUnifiedMapActivity) {
```
eddiemuc, line 172: Still don't like the idea...

But if you really want to store a global weak reference to a UnifiedMapActivity then you might as well do it directly in the UnifiedMapActivity class. No need to do that in CgeoApplication.

BTW: what happens when for some reason two UnifiedMapActivities exist at the same time? (this is perfectly allowed by Android lifecycle)

moving-bits:
Changed the PR to use `activity.recreate()` in `onResume()`. No artifacts so far, keep fingers crossed...
(Will work on optimizations later.)

main/src/main/java/cgeo/geocaching/CgeoApplication.java
moving-bits: What can we use instead to notify the "old" activity on demand? It should return to the caller (`UnifiedMapActivity` via means of `CGeoApplication` in this case) after it has reached the old activity (or if the old activity does no longer exist), so that the new MapView is only created after the old one got destroyed.

Having two instances of UnifiedMapActivity at the same time within the same application context is the thing we need to protect against, to my understanding, as then VTM would use the same global vars and state info, which mixes things up.
(Not sure what would happen if we start a UnifiedMap, then go back to the mobile and start another UnifiedMap using a direct link. Would this still be the same application context - then the implemented protection should help - , or if not, then VTM should not run into those problems.)

ztNFny:
@moving-bits as I can pretty reliably reproduce the black boxes, should I test this PR?

moving-bits:
> @moving-bits as I can pretty reliably reproduce the black boxes, should I test this PR?

as time permits. Yes, that would be helpful, thanks.

I've tested it on my physical test device, where I could reproduce @murggel's scenario reliably, and was no longer able to reproduce it, even after a stacking of at least 5 pairs (map => cache details => map => cache details etc.), and turning and zooming the map wildly.

main/src/main/java/cgeo/geocaching/CgeoApplication.java
fm-sys: IMHO the "on demand" part is the problem. 

Have you tested if always recreating the fragment really significantly degrese performance? As doing so would significantly streamline the code. 

(We just discussed in university that one of the reasons why architecture gets hard to maintain is caused by developers trying to optimize performance too early :) )

eddiemuc: > (`UnifiedMapActivity `via means of `CGeoApplication` in this case)

Again: as long as its only UnifiedMapActivity handling VTM, the code from CgeoApplication should be moved there. No need to put it into CgeoApplication. You just need a place to store a "global" weak reference, this might as well be a static variable in class `UnifiedMapActivity`.

fm-sys:
> IMHO the "on demand" part is the problem.

Just read your comment in the other thread about the problems coming with it.

Maybe it would still be worth investing these synchronization problems instead, as they might happen with the "on demand" approach as well.

> Again: as long as its only UnifiedMapActivity handling VTM, the code from CgeoApplication should be moved there.

Totally agree. You may even create a new class exclusively fulfilling the purpose of storing the reference.

main/src/main/java/cgeo/geocaching/CgeoApplication.java
moving-bits: done

and added a visual comparison of both methods, see below

moving-bits:
|video a|video b|
|---|---|
|[video a.webm](https://github.com/cgeo/cgeo/assets/3754370/7708f011-6c43-46fd-b9e2-02ce8895788b)|[video b.webm](https://github.com/cgeo/cgeo/assets/3754370/73a088de-9416-4487-ba59-d43092fb4614)|

I've made short videos of both approaches:
- video a shows the variant with immediate fragment destruction (variant a), video b shows the delayed variant (variant b)
- When calling cache details from map, then navigate for this cache, this opens a second map. In this case, both variants behave the same when returning from cache details - map will be recreated. (Only the point in time when the first fragment gets destroyed, is different)
- In all other cases, eg: calling cache details, then returning to the map, the variants will behave different: Variant a will still destroy the map (as it will do on any `onPause` event), whereas in variant b this is not necessary, as no second map got created.
- The videos show this second case, opening cache details from map, then returning to the map. In variant a you will notice the map going black on starting cache details, this is the fragment removal, and another, longer, black screen on returning from cache details, when the map is recreated. Variant b has smooth transitions, as the map fragment does not get destroyed. The duration of "blackness" depends on the hardware speed.

If you want to test on your own hardware, you can use the first commit for variant a, or both commits for variant b.

moving-bits:
> @moving-bits as I can pretty reliably reproduce the black boxes, should I test this PR?

@ztNFny 
Did you find the time to cross-check this PR with your test cases?

ztNFny:
> > @moving-bits as I can pretty reliably reproduce the black boxes, should I test this PR?
> 
> @ztNFny 
> Did you find the time to cross-check this PR with your test cases?

It looked good, but performance is pretty bad. Not really suitable for normal use imho.
Right now there are too many merge conflicts.

moving-bits:
> It looked good, but performance is pretty bad. Not really suitable for normal use imho. Right now there are too many merge conflicts.

Thanks for cross-checking. I guess you tried the default variant, which always destroys the fragment?

If so, you could add a setting named `feature_delayedMapFragment`, type `boolean`, set to `true`. It will then destroy and recreate the map only if a sub-map is called, but not if you only go to cache details, then return to the map. This is way faster.

ztNFny:
I used that.

moving-bits:
Opening a cache details page and returning to the map should have no performance penalty at all when using with the mentioned key enabled. 

If starting a sub map (eg.: "navigate to this cache"), then map fragment gets destroyed and recreated, which is a (relatively) costly operation. But as far as we know so far, this seems to be unavoidable with VTM. At least we have found no other way yet, and even the author recommend map view recreation.

moving-bits:
Another question: While testing with the key being enabled: Did you encounter any problems, eg. broken maps, crashes etc.? Would you say defaulting the key to `true` would be ok? Or better stay with default `false`?

ztNFny:
> Opening a cache details page and returning to the map should have no performance penalty at all when using with the mentioned key enabled.

Confirmed, I was talking about the switching maps scenario. Will test it in the wild once more tomorrow. Just rebased your commit onto latest master to be able to test: https://github.com/ztNFny/cgeo/tree/vtmFragmentRecreationRebase

Btw, is this already known? Annoys me when very much about UnifiedMap. When switching from OSM Offline map to Google Satellite (the only 2 views I really use, OSM for overview, Satellite for when I need to see what the situation is (like is the cache supposed to be in a tree)) then Satellite view jumps to N0 E0 (desert) for about half a second or so which make me lose all orientation i just had gained on the OSM map.
I can create a proper issue Monday if not yet reported.

ztNFny:
I used the "delayed recycling" (from my above branch) in the field today and user experience was ok. My previous tests were aimed at specifically trying to reproduce the black boxes, in that situation the performance hit was annoying. But in "real" usage I didn't encounter it that much, so I think this work-around should be acceptable for normal use.

moving-bits:
Thanks, @ztNFny, for the extensive testing. I'm glad that we may have found a working solution, finally.

moving-bits:
Today's team meeting:
- set flag to `true` initially
- merge PR to `release`
- and test for a couple of days in the nightlies

PR #15415, opened 2024-03-03
UnifiedMap: Hillshading

ztNFny:
Compiled changes in this PR:
- Hillshading for UnifiedMaps VTM with online source (Mapilion) 
  - requires (manual) pref "rapidapiKey" to be set
  - personal API key can be obtained for free but with limited quota at https://rapidapi.com
  - a second source (maptoolkit) is implemented but not used as of now
- Hillshading for UnifiedMaps Mapsforge with offline files
- Hillshading UI split into master setting in preferences (where downloading the files is explained) and map options toggle
- Adopt new hillshading algorithm for offline maps provided by mapsforge (@Sublimis)
- Offline hillshading file downloader


fm-sys:
I think geoitemlayer is unfortunately just not designed to hold a BitmapTileLayer (yet). 

Is this correct @eddiemuc?

eddiemuc:
> I think geoitemlayer is unfortunately just not designed to hold a BitmapTileLayer (yet).
> 
> Is this correct @eddiemuc?

@fm-sys is correct. GeoItemLayer is designed to handle placements of markers, geoobjects and text labels onto a map. It was never meant to deal with Tile Layers. That would be a totally different concept not fitting into its current design

ztNFny:
> That would be a totally different concept not fitting into its current design

What would be your suggestion to implement this then? Should a new intermediate layer be created? Or is the current approach already the right direction? (It works just fine, at least for OSM, I just thought it doesn't match the intended design - but maybe it does?)

eddiemuc:
> > That would be a totally different concept not fitting into its current design
> 
> What would be your suggestion to implement this then? Should a new intermediate layer be created? Or is the current approach already the right direction? (It works just fine, at least for OSM, I just thought it doesn't match the intended design - but maybe it does?)

I never thought about a concept unifying bitmap tile layers - because I always thought this is a VTM/mapsforge specific thing (can you overlay bitmap tiles onto a Google map?). 

I am also not much into hillshading to be honest But as far as I remember it is a ton of data. And the geo item layer holds everything in memory. This is totally fine for the stuff we do with it (cache markers with reusable bitmaps from a marker cache, polyline paths, occasional polygon, ...), but I assume translating hillshading data into many bitmap objects (or polylines/polygons?) is huge in data size? 

My assumption would that for bitmap layers you somehow "point" the implementation (VTM or mapsforge) to the data to display for different regions and zoom levels, then the implementation just does the heavy lifting. Does that approach not work for hillshading?

eddiemuc:
If I understand this PR correctly then it is using the described approach: "pointing" the implementation ton the bitmap source depending on position and zoom level (by means of a URL with placeholders for this data) and delegating the heavy lifting to the implementation (e.g. downloading the bitmap data, displaying correct bitmaps according to zoom level and position, maybe caching in files in-between etc).

eddiemuc:
Update: I briefly looked into Google Map API, it turns out it supports Tile Overlays. If I understand correctly, one has to provide an implementation of `com.google.android.gms.maps.model.TileProvider` for custom map overlays. This `TileProvider` is then called by GoogleMaps to get the overlay bitmaps for specific coordinates and zoom levels. Details can be found here:

* https://developers.google.com/android/reference/com/google/android/gms/maps/model/TileProvider
* https://developers.google.com/android/reference/com/google/android/gms/maps/model/TileOverlay

Implementing those might be the tricky part. Once that is done however, adding it to a google map seems to be a one-liner e.g.:

```
GoogleMap map = ...
TileProvider tileProvider = ...
map.addTileOverlay(new TileOverlayOptions().tileProvider(tileProvider));
```

Wrt unification: I assume that if we provide a generic class returning bitmap data for a given position and zoom level, such a class would be easily usable across map implementations (Google, VTM, mapsforge), provided that all of those implementations have support for bitmap overlays. So let's continue unification discussion once such an implementation is available for hillshading.

ztNFny:
@moving-bits you recently added support for custom map urls. We could do something similar to allow the usage of online hillshading sources.

The only hillshading online sources I could find are freemium, for your own use you can generate an API key, but once you reach X requests it's a paid offering which is why we can't include it in cgeo. But if users can enter their own URL / API-key we could go with the offline version.

This PR already implements a preference "rapidapiKeyMaptoolkit" to hold the key, just not any UI.

moving-bits:
> @moving-bits you recently added support for custom map urls. We could do something similar to allow the usage of online hillshading sources.

I'm not sure if this is meant as a request into my direction? If so, I would adapt the existing PR to
- use a single generic "Hillshading Tileprovider" (instead of the two specific ones currently in the PR)
- add some input fields in Unified Map's settings to be able to enter
  - URI to tileserver
  - tile mapping scheme
  - api key
- use those values to configure the tile provider

@ztNFny 
Is it that what you had in mind? Shall I start implementing in that direction?

ztNFny:
Well, I first wanted to ask whether we want to expand those custom sources direction at all. But it seems like your response is yes ;-)

In terms of UI I'd suggest something similar as we have for signatures, where the user can add any number of custom sources. Each source should have URL, Name and a "is overlay" option - the later would then make it appear not as map source but as an map overlay source - where with the current state of this PR the hillshading one is displayed.

I don't understand why you want 3 pieces of info: "URI to tileserver, tile mapping scheme, api key"
That's all just the URL to me? https://maptoolkit.p.rapidapi.com/tiles/{Z}/{X}/{Y}/hillshading.png?rapidapi-key=ABC

If you agree with this approach I could take a look at it next week, am on a business trip then and should have time in the evenings ;-)

moving-bits:
> Well, I first wanted to ask whether we want to expand those custom sources direction at all. But it seems like your response is yes ;-)

TBH: Personally, I would not use it, for the following reasons:
- Mapsforge / VTM is for me about having _offline_ maps, being able to geocache without the need of online data connection. Online hillshading source would not fit into this scheme.
- I'm rarely using hillshading anyway - only when I'm in an area where this is relevant (eg: Alps), if at all. And for those, I can switch to "old" Mapsforge currently.
- With the recent additions to Mapsforge library (rotation, fractional zoom), the need for VTM declines, and maybe we will even remove it in favor of Mapsforge in UnifiedMap (@fm-sys has started work on it). Not yet decided finally, but in sum of those three points I'm not sure if I would put efforts into a multi-source interface. But that's only _my_ point of view on this.

Having said that, if you want to dive into this feature, please keep in mind, that for selectable items (meaning: tile providers selectable by the user), a hash value of the tile provider (class name) is used to remember the last selected item. Thus if you use a single class, instantiated multiple times for different user-defined map sources, we will need to find a different mechanism.

This should not be a problem for multiple hillshading providers, as they should be selected in settings anyway (using some other mechanism). But on the other hand: Why would one have the need for multiple hillshading providers?

---

My preferred solution would be: Use the identical mechanism for hillshading across (OSM) maps. This mechanism should be offline-capable (after downloading initial data like hgt files). Probably stay with the mechanism used today by "old" Mapsforge. 

So a (IMHO more sustainable) approach would be to either implement such a hillshading support for VTM, or skip it for VTM and wait for "new" Mapsforge becoming part of UnifiedMap.

ztNFny:
> So a (IMHO more sustainable) approach would be to either implement such a hillshading support for VTM

Well of course that would be the ideal solution. I tried implementing it for a few days, but got nowhere, so I went with the solution in this PR to at least enable something.

But you're right, if VTM is not the way to go then we shouldn't invest too much time here. So I propose to leave this PR as is for the time being and review it in half a year to check the current VTM state.

bekuno:
retest this please

bekuno:
There are checkstyle warnings:
https://ci.cgeo.org/job/cgeo-CI_PR-build/8924/checkstyleResult/

ztNFny:
rebased

moving-bits:
FYI: With Mapsforge having been added to UnifiedMap recently (see #15984), you would now be able to implement a pure-offline hillshading layer like the "old" `NewMap` uses.

If we decide to go with Mapsforge instead of VTM in UnifiedMap, I would rather have an offline hillshading function than an online version.

Note: There's also some new hillshading algo in current Mapsforge (not yet released as regular release, but already available in c:geo), see discussion in https://github.com/mapsforge/mapsforge/discussions/1512. Haven't played with it yet, though.

devemux86:
@Sublimis contributed new hillshading algorithms and improvements.

The new API can be seen in [HillshadingMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/HillshadingMapViewer.java).

Hillshading magnitude can be overridden via `RenderThemeHandler.HILLSHADING_MAGNITUDE` with values [0..255].

ztNFny:
> I would rather have an offline hillshading function than an online version.

While in general I agree with that statement keep in mind that offline hillshading has 1 big disadvantage as of now: You have to manually take care of getting the files. Until we solve that the features reach will always be very limited.

> @Sublimis contributed new hillshading algorithms and improvements.

looks great, will look into adding that

ztNFny:
> Note: There's also some new hillshading algo in current Mapsforge (not yet released as regular release, but already available in c:geo), see discussion in [mapsforge/mapsforge#1512](https://github.com/mapsforge/mapsforge/discussions/1512). Haven't played with it yet, though.

on latest master I don't have the new classes available, could you please recheck? I also tried to change the commit hash in build.gradle myself, but no repo is offering it. Shouldn't it be coming in via jitpack.io? Tried f3350ef and 40bbecf

moving-bits:
> on latest master I don't have the new classes available, could you please recheck? I also tried to change the commit hash in build.gradle myself, but no repo is offering it. Shouldn't it be coming in via jitpack.io? Tried f3350ef and 40bbecf

Current `master` uses build `01e17c1` (which is Mapsforge's master as of 09/18). This should already include all or most changes with regard to the new Hillshading algorithm.

If you change this value to a newer commit hash from Mapsforge, it takes Jitpack some time to compile that on their side. Just try another build a few minutes later, then Jitpack should have it ready.


> > I would rather have an offline hillshading function than an online version.
> 
> While in general I agree with that statement keep in mind that offline hillshading has 1 big disadvantage as of now: You have to manually take care of getting the files. Until we solve that the features reach will always be very limited.

I would be willing to implement either auto-download of hgt tiles (like we have for routing tile data) or adding it to the download manager - whatever would be easier to implement. Just had not started yet as it would not help for UnifiedMap VTM. But as it looks like, direction goes to UnifiedMap Mapsforge currently (see also discussion in Cruise 5 beta).

ztNFny:
we've had hillshading for old map for a long time ;-) Problem is not implementing the downloader but having a source to download from.

> Current `master` uses build `01e17c1` (which is Mapsforge's master as of 09/18). This should already include all or most changes with regard to the new Hillshading algorithm.

unfortunately not. But I guess I'll just have to wait for jitpack - never used that one.

devemux86:
> offline hillshading has 1 big disadvantage as of now: You have to manually take care of getting the files. Until we solve that the features reach will always be very limited.

I will check if we can host the hillshading data in [Mapsforge server](https://download.mapsforge.org).
We are mostly interested for the 3" data that cover the whole world.

What about the 1" data? They cover only Europe and North America.
And 1" data take much more time to process, leading to slow map rendering.

ztNFny:
> I will check if we can host the hillshading data in [Mapsforge server](https://download.mapsforge.org). We are mostly interested for the 3" data that cover the whole world.
> 
> What about the 1" data? They cover only Europe and North America. And 1" data take much more time to process, leading to slow map rendering.

When I looked into this I didn't notice much of a difference between 1" and 3", at least not for the normal geocaching usage. As 1" is significantly larger I always just went with 3".

For Europe I'm using https://sonny.4lima.de/ as source, for ROW http://viewfinderpanoramas.org/dem3.html

devemux86:
> But as it looks like, direction goes to UnifiedMap Mapsforge currently (see also discussion in Cruise 5 beta).

**Cruiser 5** will only have Mapsforge as map engine (not OpenGL).

Map rotation implementation was the last requirement.
Users would also like fractional zoom instead of the current integer zoom steps.

I implemented an optional fractional zoom in Mapsforge via `Parameters.FRACTIONAL_ZOOM`,
but it needs more work and thought on how we want it to work.

eddiemuc:
> > Current `master` uses build `01e17c1` (which is Mapsforge's master as of 09/18). This should already include all or most changes with regard to the new Hillshading algorithm.
> 
> unfortunately not. But I guess I'll just have to wait for jitpack - never used that one.

You can actively trigger builds also directoy on jitpack server by clicking "get it". Check https://jitpack.io/#mapsforge/mapsforge to see what is already available.

image

I assume you can use the newest commit f3350ef261 for the hillshading use case

ztNFny:
> You can actively trigger builds also directoy on jitpack server by clicking "get it". Check https://jitpack.io/#mapsforge/mapsforge to see what is already available.

It's working now. I checked that page, it was already green when I first tried to pull it, but apparently it still needed some more time.

ztNFny:
UnfiedMaps Mapsforge with HS :-)

image

ztNFny:
should be ready to merge now

Changes:
- Hillshading for VTM with online source
-- requires (manual) pref "rapidapiKey" to be set
-- personal API key can be obtained for free but with limited quota at https://rapidapi.com
- Hillshading for UnifiedMaps Mapsforge
- Hillshading UI split into master setting in preferences (where downloading the files is explained) and map options toggle

moving-bits:
Thanks for the implementation.

Where can I find the map options toggle for hillshading in UnifiedMap?
Hillshading is enabled in Settings => Map Sources, hgt file is downloaded.
I've looked in Overflow menu => Theme Options as well as in Map sources, but can't see such an entry in this build.

ztNFny:
> Where can I find the map options toggle for hillshading in UnifiedMap?

image

If you're using VTM then you rapidapiKey set, for UnifiedMaps Mapsforge only the master toggle needs to be enabled.

moving-bits:
> If you're using VTM then you need rapidapiKey set, for UnifiedMaps Mapsforge only the master toggle needs to be enabled.

Ok, I was confused by not seeing this additional menu entry even though I had enabled hillshading in settings. By adding a (dummy) key, the menu entry comes up. And when switching to a different part of Germany (with some more visible elevation changes), hillshading also works for Unified Map Mapsforge now. Thanks.

ztNFny:
> > If you're using VTM then you need rapidapiKey set, for UnifiedMaps Mapsforge only the master toggle needs to be enabled.
> 
> Ok, I was confused by not seeing this additional menu entry even though I had enabled hillshading in settings. 

Old and new mapsforge use the master switch.
VTM Online map uses the rapidapiKey hidden pref + master switch. Depending on VTMs future that rapidapiKey should be un-hidden, but for now i don't want to invest additional effort.

ztNFny:
Discover an issue: with Unified mapsforge mapping a list the list is not focused. Guess HS breaks the viewpoint calculation.

devemux86:
Mapsforge server + fast mirror now also host 3" hillshading data for the whole world.

https://download.mapsforge.org

ztNFny:
automatic downloader integrated :-)

let's give this a try in nigthlies

moving-bits:
Thanks for your in-depth work on hillshading support. Let's try it in the nightlies.

devemux86:
For your information:

- https://github.com/mapsforge/mapsforge/pull/1548

ztNFny:
@moving-bits maybe you.could.pull latest mapsforge master to get this?

ztNFny:
If anything I'd suggest to offer it as a preference

moving-bits:
I've made some tests with `StandardClasyHillShading` and `HiResStandardClasyHillShading` after having downloaded 3" hgt files using c:geo's integrated downloader. All tests with UnifiedMap Mapsforge:

1=Without hillshading
2=`StandardClasyHillShading`
3=`HiResStandardClasyHillShading`
|1|2|3
|---|---|---|
|Screenshot_1728077970|Screenshot_1728077857|Screenshot_1728077947|
|Screenshot_1728077162|Screenshot_1728077219|Screenshot_1728077187|
|Screenshot_1728077650|Screenshot_1728077277|Screenshot_1728077309

At least in the two emulators I tested with there seems to be no relevant speed difference between `StandardClasyHillShading` and `HiResStandardClasyHillShading`

devemux86:
I also made tests using Cruiser and Mapsforge samples on desktop:
`StandardClasyHillShading` seems to render the map tiles faster than `HiResStandardClasyHillShading`.

Perhaps @Sublimis can comment if there is indeed any performance difference.

Sublimis:
Hi!

On multiprocessor systems, systems with "enough" CPU cores, the execution time (time until results are produced) of two algorithms is roughly the same due to efficient parallelization. The more CPU cores and threads the better.

However, it is quite certain that `HiResStandardClasyHillShading` uses more energy (CPU cycles) than `StandardClasyHillShading`, since it does more calculations. It calculates 4x more pixels, and also does more calculations per pixel.

In order to measure as objectively as possible how much more energy the HiRes algorithm uses, multithreading should be disabled (the reading and computing thread count set to zero), and then the execution times should be compared.

Our provisional benchmark results in such a single-threaded setup show that the HiRes algorithm is about 3x slower, so we can say that it uses about 3x more energy (by activating more CPU cores in multi-threaded setup).

We can try to do "more scientifical" benchmarking results soon, if this would be useful?

Sublimis:
It should also be noted that virtually all mapping solutions have some sort of caching, so the fact that `HiResStandardClassyHillShading` uses 3x more energy for its calculations probably wouldn't matter that much in the real world.

Sublimis:
@devemux86 

Yes, and this is why we made the multi-threaded hill shading conservative in terms of thread count (so it doesn't set the thread count to 4 or 8 by default as one might expect, but 1 or 2).

PR #15417, opened 2024-03-04
change collapsed group icon to right-arrow instead of up-arrow

ztNFny:
change the "collapsed" icon in simple dialog groups from up-arrow to right-arrow
new:
image
old:
image


fm-sys:
As it's uncritical, maybe even target `release`?

ztNFny:
It's uncritical, so master is fine :-)

fm-sys:
OK, you choice :-)

PR #15416, opened 2024-03-03
rel to #15194: use stores list dialog for dropdowns

eddiemuc:
rel to #15194: use stores list dialog for dropdowns

PR #15428, opened 2024-03-08
fix crash opening UnifiedMap with non-Google map

ztNFny:
https://github.com/cgeo/cgeo/pull/15416 introduced a crash opening Unified Map with non-google maps. This PR doesn't "solve" the problem but at least the crash by catching it - it's a band-aid until the issue can be properly fixed


ztNFny:
See #15430

Issue #15425, opened 2024-03-08
Add a draggable scale to maps


Futuray-pgm:
### Tell us your idea!

I think it will be realy useful to made the scale draggable around the map. 

### Additional Information

I know that it can be really hard to implement.

MagpieFourtyTwo:
Maybe I'm missing something, but what exactly would be the useful part?

brivars:
> Maybe I'm missing something, but what exactly would be the useful part?

I'm guessing, to estimate distances / do rough measurements. I have sometimes found myself trying to pan the map around in a way that the scale indicatir would roughly sit so I could guesstimate a distance between two caches.

However I'm not sure if a "draggable scale" would be the best solution. 
What would be better would be to add some sort of a ruler function, as found on Organic Maps, Google Maps, and other mapping apps.

moving-bits:
To measure the distance between two caches, it's easier to put them into a so-called "individual route" (by long-tapping on them). c:geo will then display the distance between them (or, to be more specific: the length of the route), either as bee-line or routed for walking, bicycle or car, depending on your routing settings.

Futuray-pgm:
I wanted that a scale like the one that already exist bottom of each map.

And it will be easier to measure between two points without to create two personalised caches.

eddiemuc:
I imagine something like this:
- on map long-tap, add a new option to context Menü: "start measurement" (wording tbd)
- toast appear: "tap on target for measurement". Tapped start point will be marked b e.g. a cross symbol
- next tap on map will NOT trigger the normal action but instead draw a line from start tap to second tapped point with length written onto it (e.g. 530m or 3,2km)
- measurement line can be tapped to be deleted. Tapping will open a confirm dialog asking for line to be deleted 
- nothing of the above will be persisted in any way. Leaving map and reentering will remove measurements

Futuray-pgm:
Yes!
And when clicking the measurement, a context menu appears with this options:
- "`change color`"
- "`informations`"
- "`save measurement`" (into the current list)
- "`save all measurements`"
- "`delete measurement`"
- "`delete all measurements`"

fm-sys:
IMHO cgeo should focus on geocaching. While measurements might be a small nice additional feature, it should be nothing more than that. Focus on simplicity!

- on map long-tap, select start measurement (that paint is used as start)
- user can do the measurement by moving the map around (center is used as target point)
- pressing back button key removes it again (no confirmation necessary)

[...]

PR #15429, opened 2024-03-08
change position of "Elevation" menu item

ztNFny:
https://github.com/cgeo/cgeo/pull/15127 added an "Elevation" item to map long-tap menu that's "not perfect" imho:

The "last" (just like "first") menu-item-position is special because it's the one menu item that can be used without further looking at the screen - you just know to hit the menu at its bottom. For all other menu items you have to carefully aim your tap. And now this special position is used by an item that never needs to be clicked (because there's no click action), thus making it harder to tap "Navigate"

Solution is easy: Move the menu item, my proposal would be below "copy coordinates", which is what this PR does.
image

My second issue (and the reason for the "Feedback required" tag) is that there are now 2 related/similar menu items: Copy coords & Elevation.
They are both giving you information about the point you tapped on the map, taken from the same source. Height is just another (usually hidden) dimension of coordinates.
So they could get combined into 1 line. The question is a) whether that's wanted and b) in which way.

My initial thought was to show both coords and elevation, but that results in a very long string - ugly:
image

Maybe just have 1 menu item which shows the Elevation (if available) and on click the "Copy coordinates" dialog? Menu item text would be "Copy coords" with elevation not available, "Elevation X m" with elevation available - but that has the disadvantage of inconsistency, depending on the selected map, also not great.

### DIFF
--- a/main/src/main/res/menu/map_longclick.xml
+++ b/main/src/main/res/menu/map_longclick.xml
@@ -21,16 +21,16 @@
         android:title="@string/context_map_show_coords">
     </item>
     <item
+        android:id="@+id/menu_elevation"
+        android:title=""
+        android:visible="false">
+    </item>
+    <item
         android:id="@+id/menu_target"
         android:title="@string/cache_menu_target">
     </item>
     <item
         android:id="@+id/menu_navigate"
         android:title="@string/cache_menu_navigate">
-    </item>
-    <item
-        android:id="@+id/menu_elevation"
-        android:title=""
-        android:visible="false">
     </item>
 </menu>


moving-bits:
Moving the elevation info I'm fine with, but I don't have a good solution for the second issue yet. I see a couple of alternatives:
- have it as a separate menu item (= what we have today) => easily accessible, but uses space and is non-clickable
- combine with coordinates
  - show coordinates and elevation directly => makes the popup menu pretty wide & ugly (as shown in screenshot above)
  - just display elevation, and open coordinate info on tap => takes up less space, but who will know that you will find coords behind elevation info?
  - only show coord menu item (same we have today), and display elevation in the popup that follows => might be ok, but is an extra tap for elevation info
  - only show coord menu item, but add elevation info in brackets behind the menu title: "Show coordinates (36,5m)" => easy to access, but may be confusing at first; also the info, that this elevation is taken from map info, not from GPS, is lost then

Personally, I would just move up the elevation info (as is done in this PR), and keep the info in the popup directly.
If we want to get rid of the extra menu item, then probably the variant "put elevation info into coordinates popup" would be my preference.

ztNFny:
My idea: 1 menu item "coordinate information" that opens a dialog with:
- coords (clickable to switch format)
- elevation
- distance from current location (addressing part of the usecase of https://github.com/cgeo/cgeo/issues/15425)
- a "copy coordinates" button

moving-bits:
Setting WIP label for this PR, as the discussion favors a different solution, as proposed in https://github.com/cgeo/cgeo/pull/15429#issuecomment-2002128498 by @ztNFny

Lineflyer:
> My idea: 1 menu item "coordinate information" that opens a dialog with: [...]

Sounds like a good idea

PR #15430, opened 2024-03-08
fix NPE in #15416

eddiemuc:
fix NPE in #15416


eddiemuc:
Given severity of the problem, I will quick-merge this PR as soon as it is green

ztNFny:
I can verify it tomorrow, not before evening though.

PR #15432, opened 2024-03-09
re-add all caches to map before applying a new filter (fixes #15399)

### LINKED ISSUE #15399: UnifiedMap: Can't clear/change filter from long-tap menu
### Describe your problem!

Clearing / changing filters from maps long-tap filter menu don't work.
When selecting a different filter this new filter seems to be applied on top of the old one.
Seems the "clear" is never applied.

Using the filter dialog to clear the filter works fine.

### How to reproduce?

1. Map a list (didn't test with live)
2. apply a filter to the map, e.g. a "DNF"
3. Long-tap on the filter icon in toolbar, select "clear" to remove the filter

no change, still only DNF caches visibile

map shows all caches again

### Reproducible

Yes

### c:geo Version

2024.02.28

### System information

_No response_

### Additional Information

_No response_

ztNFny:
When selecting a different filter this new filter seems to be applied on top of the old one. This PR fixes that by reloading all caches, making applying a new and clearing a filter work as expected.


eddiemuc:
Thanks for the fix!

PR #15437, opened 2024-03-10
Show navigation hint

moving-bits:
## Description
I'm experimenting a bit with navigation hint data which can be obtained from BRouter. With this PR, it would be displayed as symbol + text, eg. "keep straight (symbol) in 99.6m"

image

If we pursue this path, next step would be to integrate this info into the "supersize cycle", so that the info can be shown in large on request. Ultimately, an optional voice output can be imagined as well.


bekuno:
Retest this please

eddiemuc:
So routing and nvaigation **do** finally merge into one thing?

moving-bits:
- updated to current state of `master` (including removal of old maps)
- added supersize view for navigation hint

image

open issues:
- [x] supersize requires "show straight distance" to be activated to show turn by turn info
- [x] turn by turn info sometimes gets lost, probably after some location update

moving-bits:
retest this please

PR #15442, opened 2024-03-10
hide "log in browser" for non-premium caches

ztNFny:
see https://github.com/cgeo/cgeo/discussions/15434#discussioncomment-8737148

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheMenuHandler.java
+++ b/main/src/main/java/cgeo/geocaching/CacheMenuHandler.java
@@ -139,7 +139,7 @@ public final class CacheMenuHandler extends AbstractUIFactory {
             NavigationSelectionActionProvider.initialize(menu.findItem(R.id.menu_navigate), cache);
         }
         menu.findItem(R.id.menu_log_visit).setVisible(cache.supportsLogging() && !Settings.getLogOffline());
-        menu.findItem(R.id.menu_log_in_browser).setVisible(cache.getCreateNewLogUrl() != null);
+        menu.findItem(R.id.menu_log_in_browser).setVisible(cache.isPremiumMembersOnly() && cache.getCreateNewLogUrl() != null);
         menu.findItem(R.id.menu_log_visit_offline).setVisible(cache.supportsLogging() && Settings.getLogOffline());
         menu.findItem(R.id.menu_set_found).setVisible(cache.supportsSettingFoundState() && !cache.isFound());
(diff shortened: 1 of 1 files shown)


eddiemuc:
I would suggest renaming the menu entry if this PR is applied e.g. "Log for basic user logged into browser" (maybe someone can think of something shorter with same meaning...)

ztNFny:
Maybe it could also be moved from CacheDetails to Log dialog?

ztNFny:
I can think of 2 situations when this menu item will be used:
- Basic member who has got the cache via GPX: In that case logging from inside c:geo is not possible - so trying to send the log should (offer to) open the log page in browser. More user friendly, no more link required.
- Premium user who wants to log the cache also with a basic account (e.g. kids): In that case the Premium user is logging the cache already, so the menu item could be moved into the (what to me feels more appropiate) Log dialog instead of CacheDetails.

ztNFny:
> I would suggest renaming the menu entry if this PR is applied e.g. "Log for basic user logged into browser" (maybe someone can think of something shorter with same meaning...)

How about "get basic-member log link"?

Lineflyer:
Whats wrong with "Log cache in browser" if its anyway just displayed in case a BM wants to log a PM-only cache.

eddiemuc:
> Whats wrong with "Log cache in browser" if its anyway just displayed in case a BM wants to log a PM-only cache?

My thought was that users will probably not understand why this option is displayed only for premium caches.

Anyway, probably I am overthinking this and no one will notice. So I am fine with leaving it be.

ztNFny:
> Whats wrong with "Log cache in browser" if its anyway just displayed in case a BM wants to log a PM-only cache?

It describes what the button does, but not why that button even exists. To most users it's completely useless. See https://github.com/cgeo/cgeo/discussions/15434#discussioncomment-8737148 where it has to be explained to me

murggel:
I just stumpled upon this PR and I don't want to start a new discussion , but for me this feature is not only relevant for premium only caches but also for basic caches. With that function I can log from my device also for my children with a different account and no own device the caches.
So if that would only be for premium caches possible, I would be sad...

eddiemuc:
+1 for being sad if log in browser is not available any more in basic caches. Makes my  log-for-kids- workflow so much easier if I don't have to differentiate between basic and premium.

eddiemuc:
Closing this very old PR for now. Feel free to reopen if necessary.

Issue #15440, opened 2024-03-10
crash applying a filter to a stored list on map


ztNFny:
1. Open a list
2. Map the list
3. Apply a filter 

FATAL EXCEPTION: main
Process: cgeo.geocaching, PID: 7785
java.util.ConcurrentModificationException
	at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:760)
	at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:782)
	at cgeo.geocaching.filters.core.GeocacheFilter.filterList(GeocacheFilter.java:182)
	at cgeo.geocaching.maps.MapUtils.filter(MapUtils.java:105)
	at cgeo.geocaching.unifiedmap.UnifiedMapActivity.refreshMapData(UnifiedMapActivity.java:921)
	at cgeo.geocaching.unifiedmap.UnifiedMapActivity.refreshWithFilter(UnifiedMapActivity.java:913)
	at cgeo.geocaching.utils.FilterUtils$$ExternalSyntheticLambda3.accept(Unknown Source:4)
	at cgeo.geocaching.ui.dialog.SimpleDialog.lambda$selectSingle$4(SimpleDialog.java:472)
	at cgeo.geocaching.ui.dialog.SimpleDialog$$ExternalSyntheticLambda4.accept(Unknown Source:4)
	at cgeo.geocaching.ui.dialog.SimpleDialog.lambda$selectItems$7$cgeo-geocaching-ui-dialog-SimpleDialog(SimpleDialog.java:573)
	at cgeo.geocaching.ui.dialog.SimpleDialog$$ExternalSyntheticLambda8.call(Unknown Source:13)
	at cgeo.geocaching.ui.SimpleItemListModel.triggerChange(SimpleItemListModel.java:408)
	at cgeo.geocaching.ui.SimpleItemListModel.setSelectedItems(SimpleItemListModel.java:380)
	at cgeo.geocaching.ui.SimpleItemListView$ItemListAdapter.handleClick(SimpleItemListView.java:257)
	at cgeo.geocaching.ui.SimpleItemListView$ItemListAdapter.lambda$onBindViewHolder$2$cgeo-geocaching-ui-SimpleItemListView$ItemListAdapter(SimpleItemListView.java:268)
	at cgeo.geocaching.ui.SimpleItemListView$ItemListAdapter$$ExternalSyntheticLambda5.onClick(Unknown Source:4)

[...]

PR #15443, opened 2024-03-10
rel to #15440: rollback #15432 to fix concurrent-mod-exception

eddiemuc:
rel to #15440: rollback #15432 to fix concurrent-mod-exception


ztNFny:
Should be targeted to release.

Also i think #15444 should fix this better, to be discussed in the issue

eddiemuc:
> Should be targeted to release.
> 
> Also i think #15444 should fix this better, to be discussed in the issue

Ok, closing here

Issue #15386, opened 2024-02-25
Live map not showing caches


moving-bits:
Don't know for a good title for this, so let's try to describe a bit more in details what seems to happen:

We had several reports throughout the last weeks, that live map stayed (seemingly) empty for the users, not displaying any caches. For some of them it's only lab adventures showing up, but no other cache types.

I'm using beta version currently, and have observed this myself, though not consistently. Sometimes zooming in/out far enough seems to help, but not always.

I tried to debug a bit into it today, by adding several log statements in the connectors' `searchByViewport()` methods:
- GC search returns 200 hits, AL search 100 hits, `ConnectorFactory.searchByViewport()` correctly reports 300 hits in that case. Sometimes, GC search returns 0 hits, but AL still 100.
- I've checked the returned coordinates for the non-labs, they are within the given viewport. But they are not being displayed in the current viewport.
- sometimes, if you move the map around a bit, you'll see some non-labs at one of the corners/edges of the viewport (slightly outside visible viewport)
- searching further, I found a hard-coded limitation to a maximum of 200 caches for non-labs - is that intentionally? Didn't we have 1000 in earlier times? Especially with larger viewports, nearly none of the caches in that area will be loaded by this, and if they happen to be a bit more on the outside, it looks like an empty result. 
- (Keep in mind, that we search for "current viewport * factor 3" for default and we do cache this search result, so any follow-up search within this area will return the old search result, even if you have zoomed in a bit and moved the map.)

A way to reproduce this would be like this:
- go to an area with "enough" caches
- zoom out far enough to see a somewhat larger area (like a few square km)
- activate live mode => caches are being loaded, but most likely only labs are being displayed in the visible area
- scroll around a bit - nothing but lab adventures
- if you zoom in a few levels, you can scroll around quite a bit, but are still within the cached are for this search result, so you won't get any caches other than labs, except when you leave the originally visible area far enough to come to the edges of the search viewport

This would at least be an explanation, and if this train of thought is correct:
Can't we at least set the limit back to 1000 if searching on live map? (May be different when searching from within lists.)

moving-bits:
related tickets (among others):
375998, 552503, 690870

(both for RC and for current release 2024.01.06)

ztNFny:
From https://github.com/cgeo/cgeo/issues/15261:

> My best guess from looking at the behaviour and some breakpoints is that loading the 200 new caches invalidates the previously loaded live-map-caches, thereby removing them and only showing the new ones.

[...]

PR #15445, opened 2024-03-10
rel to #15386: set viewport search limit to 500 for gc.com

eddiemuc:
rel to #15386: set viewport search limit to 500 for gc.com


moving-bits:
retest this please

moving-bits:
should go to `release`, as the referenced issue is happening with current release 2024.01.06 already

PR #15446, opened 2024-03-10
rel to #15386: set viewport search limit to 500 for gc.com

eddiemuc:
rel to #15386: set viewport search limit to 500 for gc.com


moving-bits:
thanks for rebasing

Issue #9283, opened 2020-10-25
Resource IDs will be non-final in Android Gradle Plugin version 5.0, avoid using them in switch case statements


bekuno:
This issue is intended to document the progress towards eliminating this code construct.

See #9224 and #9271 (and the sister issue #9232)

moving-bits:
This issue should be fixed with the three steps above, can't find any "case R.id" anymore in the code, therefore closing.

PR #15447, opened 2024-03-10
Update AGP to 8.0.2

bekuno:
Update to the next AGP Version 8.0.x to the highest patch level.

This needs at least gradle 8.0, so switched to 8.0.2.

Further changes where done with the AS "AGP Upgrade Assistant".

- Update default R8 processing mode The default for the R8 mode in AGP is now full mode, rather than the previous compatibility mode. If your project requires compatibility mode, it needs to be set explicitly in the gradle.properties file.

  Note: No problems seen so far on local tests.

- Enable buildConfig build feature The default value for buildFeatures.buildConfig is changing, meaning that the Android Gradle Plugin will no longer generate BuildConfig classes by default. This processor adds a directive to preserve the previous behavior of generating BuildConfig classes for all modules; if this project does not use BuildConfig, you can remove the android.defaults.buildfeatures.buildconfig property from the project's gradle.properties file after this upgrade.

  Note: We use the automatic BuildConfig class cgeo.geocaching.BuildConfig. The assistent "Migrate BuildConfig to Gradle Build Files" reconfigures from gradle.properties to build.gradle.

- Preserve transitive R classes R classes in this project are transitive, pulling in information from their dependencies. The default behaviour in the Android Gradle Plugin is changing; this processor inserts a property in gradle.properties to preserve the current default. You can migrate your project to non-transitive R classes now or later using the "Migrate to Non-Transitive R Classes" action under "Refactor".

  Note: No changes in c:geo using the "Migrate to Non-Transitive R Classes" beside of setting android.nonTransitiveRClass=true.
  Edit: This parameter is the default, so not set.

- Preserve constant R class values R classes in applications and tests previously used constant values that can be inlined by the java compiler. The default behaviour in the Android Gradle plugin is changing to speed up builds and allow more precise shrinking; this processor inserts a property in gradle.properties to preserve the current default.

  Note: For switching to android.nonFinalResIds=true see https://stackoverflow.com/questions/69120822/resource-ids-will-be-non-final-by-default-in-android-gradle-plugin-version-8-0 I have not checked, if we have such constructs.
  Edit: This was already cleaned up, so we can use the default parameter value = not set.

- Enable aidl explicitly where needed The default value for buildFeatures.aidl has changed, and some modules in this project appear to be using AIDL. Build files will be modified to explicitly enable AIDL in modules with AIDL sources.

  Note: As the assistent enables this feature, than it should be ok.


bekuno:
> I dimly remember that we changed to non-final resource IDs some time ago (removing many switch statements, heavy work done by @moving-bits ). Shouldn't we be able to turn this on?

You are completely right. 
This was Issue #9283 3 1/2 years ago.

bekuno:
DNM - this PR needs #15449 to be fixed.

bekuno:
replaced by #16694

PR #15454, opened 2024-03-13
UnifiedMap: Apply list filter already on loading caches (fix #15414)

### LINKED ISSUE #15414: UnifiedMap: Mapping filtered list loads all caches
There's a design flaw in Unified map that makes it *significantly* less performant than old maps. It seems that UnifiedMap always loads all the caches of a list and only then applies a filter.

To reproduce:
- Load a list with many caches (20k in my case)
- Apply a filter to the list that leaves only a few caches (~50)
- Map the list

Result:
- Old maps: Map is displayed pretty much instantly
- Unified map: Before https://github.com/cgeo/cgeo/pull/15395 there was a ANR, now the map just sits there for ages, c:geo is unresponsive. After a very long time caches will show up on the map but any further actions have very bad performance.

moving-bits:
## Description
One idea for speeding up loading cache lists: 
Apply filter already on loading caches (in contrast to: loading ALL caches first, then filter some of them out)

@ztNFny 
Can your cross-check if that speeds up your use-case? (as time permits)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -421,7 +421,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                 // load list of caches belonging to list and scale map to see them all
                 final AtomicReference<Viewport> viewport3 = new AtomicReference<>();
                 AndroidRxUtils.andThenOnUi(Schedulers.io(), () -> {
-                    final SearchResult searchResult = DataStore.getBatchOfStoredCaches(null, mapType.fromList);
+                    final SearchResult searchResult = DataStore.getBatchOfStoredCaches(null, mapType.fromList, mapType.filterContext.get(), null, false, -1);
                     viewport3.set(DataStore.getBounds(searchResult.getGeocodes(), Settings.getZoomIncludingWaypoints()));
                     addSearchResultByGeocaches(searchResult);
                     if (viewport3.get() != null) {


ztNFny:
Loading a big list (20k) with a small filter (100) applied is now fast (about twice as long as loading in list view).
Loading the big list with a big filter (19k of 20k) is still very slow, 1 minute as opposed to 4 seconds in list view)

eddiemuc:
> Loading the big list with a big filter (19k of 20k) is still very slow, 1 minute as opposed to 4 seconds in list view)

I have two guesses what could be causing this:
- list view has a built-in "offline load" limit. Means: it will not load all caches into memory but only the first x ones (configured in appearance settings: list load limit). Feature was not liked by many when introduced, so you have probably set it to "unlimited". If not it may explain why list loading is faster than map loading
- unified map geo item layer does not produce all "add icon to map" requests at once but places them in chunks onto the GUI thread. This is meant to keep the GUI responsive even on large map changes. But in your case it might also visibly slow down the display phase?

ztNFny:
list load limit is unlimited

> This is meant to keep the GUI responsive even on large map changes. 

That doesn't work at all. UI is 90% unresponsive during the loading period.
Of course mapping 20k caches is not a normal scenario, I ran it specifically to test the performance of this. And the old maps weren't exactly fast either. But if it's something we can improve on imho it's worth looking into.

Is loading large lists not slow for you?

eddiemuc:
> list load limit is unlimited
> 
> > This is meant to keep the GUI responsive even on large map changes. 
> 
> That doesn't work at all. UI is 90% unresponsive during the loading period.

I tried it out. The GUI freezes way before mapmia even started. At this point the icon-load-approach is not yet active, so this problem is caused by something else

> Is loading large lists not slow for you?

It is. Moreover c:geo is crashing on it

ztNFny:
With regard to this PR everything is fine, it can be merged. Let's track the "loading a large list is slow" in a seperate issue.

Issue #15478, opened 2024-03-17
How to sort "List of lists" where the current Cache is stored in


MagpieFourtyTwo:
> your remark about "Lists in groups" draw my attention: How is "show lists of current cache at top of list" supposed to work with grouped lists? Would be a deal breaker for me if these lists would remain hidden in their folders ... I would prefer to see these lists at the top, too ... with their entire "coloned name" ...

Yes, that's worth discussing. In the PR I merged this morning it is implemented such that groups with a selection in them are sorted atop, and inside any group selected items are sorted atop. If anything else is wished let's open a new issue for discussion.

_Originally posted by @eddiemuc in https://github.com/cgeo/cgeo/issues/15464#issuecomment-2002452556_

MagpieFourtyTwo:
> groups with a selection in them are sorted atop, and inside any group selected items are sorted atop

Good thing as first approach, but in fact a bit impractical: The reason (or at least mine) for grouping lists in folders, is to get a better overview. Which is in fact only (or especially then) needed, if you have a bunch of lists. Moreover a folder of lists most probably contains more than just a few lists (otherwise grouping would not make very much of a difference).

Now imagine folders whose listings are longer than the popup height and a cache, which is stored in several lists, which are organized in different folders (e. g. grouped by T-Rating-Ranges, Quality Levels, Regions, plan for next trips, or stuff which is needed for some special challenges). And now have a look at this cache's "List of lists": Even with the first (open) folder in view you will not be able to see the second list where he is stored in. Not even talked about the list, which is in the third or fourth folder ... got the picture?

This will again end up in scrolling and/or closing and opening folders - which will then be the same amount of or even more hassle as it was without any folders at all. Plus the (at least once on every search) effort to scroll down to the next folder resp. "non foldered list entry", just to find out if there are no more lists where he is stored in ...

TMALSS: With this strategy you *always* have to scroll (even if the cache is in just one list), instead of just getting all lists at a glance.

This I would highly prefer if the lists, where a cache is stored in, would *always* be listed at the top of the list. With the colon in their name. Otherwise IMO the "top of the list" feature makes no sense. And the grouping of lists even less ...

MagpieFourtyTwo:
> groups with a selection in them are sorted atop, and inside any group selected items are sorted atop

[...]

PR #15479, opened 2024-03-17
rel to #15478: place selected lists on top ungrouped

eddiemuc:
rel to #15478: place selected lists on top ungrouped


ztNFny:
for downvote-reason see https://github.com/cgeo/cgeo/issues/15478#issuecomment-2002554918

eddiemuc:
There seems to be a majority for this solution so I am merging now. This shall not mean the end of the current discussion, let's continue it in #15478 and find the best solution as a team.

PR #15480, opened 2024-03-17
rel to #15462: try to reduce database accesses for list loading

eddiemuc:
rel to #15462: try to reduce database accesses for list loading

A first step to work on performance issues on loading of large lists into unified map: load all caches from DB at once (not cache-by-cache).

@moving-bits could you take a look at this PR?


fm-sys:
LGTM I would say

moving-bits (approved):
LGTM, thanks for diving into it

PR #15504, opened 2024-03-23
Replace MapQuest geocoder by OSM Nominatum via c:geo proxy geocoder

moving-bits:
## Description
- Replace MapQuest geocoder by OSM Nominatum via c:geo proxy geocoder
- Enable `OsmNominatumGeocoder` for home screen coords reverse geocoding as fallback for integrated Android geocoder

(@Lineflyer: I'm using `api.c-geo.de` for now, `api.cgeo.org` runs into ssl configuration error curently)


Lineflyer:
yes, we can keep using api.c-geo.de for now.

moving-bits:
@Lineflyer / @mucek4 

Seems that our CI is currently unable to reach `api.c-geo.de`, see error message below (extract from log):

```
Caused by: java.net.ConnectException: failed to connect to api.c-geo.de/2001:1640:5::3:d5 (port 443) from /fec0::dce2:31ac:789b:ac71 (port 38950) after 30000ms: isConnected failed: ENETUNREACH (Network is unreachable)
at libcore.io.IoBridge.isConnected(IoBridge.java:287)
```

Test runs fine when started locally in the emulator.

Lineflyer:
No restriction on my side. Maybe a firewalling on CI side?

bekuno:
Maybe a CI firewall configuration?
@kumy @samueltardieu @mucek4

kumy:
I never heard about an egress firewall on the CI infra.
/cc @mucek4

kumy:
May you try to force to use ipv4 (is it something you can force from the java code?) I suspect an ipv6 peering issue there... or the receiving service dropping our packets...

kumy:
`fec0::dce2:31ac:789b:ac71` is a link ip. Maybe ipv6 is not really functional from inside the ci runners 🤷

kumy:
Using the Jenkins "Script Console", I confirm that runners `cgeo-4` `cgeo5` cannot connect to `api.c-geo.de` (ipv4 neither ipv6)
- no ping
- no http[s]

While `master` can ping/https but **only when using ipv6**

Lineflyer:
Would it be of any help to limit `api.c-geo.de` to IPv4 only?

moving-bits:
I have disabled the test for now (as it is currently, without this PR), that's why the build is back to green. IMHO this PR is ready to merge, so that we have an alternative for the non-working Mapquest geocoder.

We should discuss, though, how to proceed with the test:
- Probably #15529 will help here with the technical side, as it is a different server environment.
- We should decide about the "where" - unit test or watchdog test, what is more appropriate?

Issue #15505, opened 2024-03-23
Users missing the stars for D/T on the map sheet


Lineflyer:
I discussed with a bunch of users, which are a bit disappointed regarding the change from popup to bottom sheet.
While at first they claimed a smaller font (which turned out not to be the case) later it turned out, what they were really missing is the better readability of D/T represented by stars.

Some users now instead use the D/T-icons on the map, but its still not the same for them.

I have to somewhat agree, that it easier/faster to recognize D- and T-values if they are presented graphically. 
Not sure if we want and can find a solution which is more graphical for D/T but still compact for the bottom sheet.

moving-bits:
How about putting it in the same line, eg: `D 5.0 ***** (separator) T 2.0 **`
This would keep it compact, and if the screen is too small, the textual "T 2.0" should still be visible. (to be tested)

ztNFny:
> How about putting it in the same line, eg: `D 5.0 ***** (separator) T 2.0 **` This would keep it compact, and if the screen is too small, the textual "T 2.0" should still be visible. (to be tested)

how about "D 5.0 / T 2.0 - ***** / **"? That way it's easier to handle overflow and it also ensures that texts are always in the same position (where the user expects them to be).

Or maybe show the marker in line with the rating instead of the stars?

My personal vote would be for not re-adding the stars, I also removed them in DetailsActivity in my builds - but I have my own builds, so I don't have any strong opinion on this ;-)

bekuno:
Instrad to use stars I propose to use a colored background in the colors already used for the cache D/T info, eg.
![grafik](https://github.com/cgeo/cgeo/assets/32555742/cb288ddd-f5e3-46a7-97ec-9ee5ffcea477)
Perhaps we could use only part of it instead of the full colouring, analogous to the circle representation. But then we have to be aware of the theme colour.

Lineflyer:
> Instead to use stars I propose to use a colored background in the colors already used for the cache D/T info, eg.

Sounds like a good space saving ideas and would allow to visualize D/T at first sight.
Only the styling needs to be well done to be readable on white/black theme (as you said).

ztNFny:
A quick stab at adding the circle marker to popup:
![image](https://github.com/cgeo/cgeo/assets/1258173/156650a8-05bd-4c8e-af9c-0d200960a352)

MagpieFourtyTwo:
[...]

PR #15518, opened 2024-03-25
DT indicator in CachePopup

ztNFny:
A quick stab at adding the circle marker to popup to fix https://github.com/cgeo/cgeo/issues/15505
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/CacheDetailsCreator.java
+++ b/main/src/main/java/cgeo/geocaching/ui/CacheDetailsCreator.java
@@ -256,7 +256,7 @@ public final class CacheDetailsCreator {
         final LinearLayout linearLayout = layout.findViewById(R.id.linearlayout);
 
         final ImageView catIcon = new ImageView(context);
-        catIcon.setBackground(createDTRatingMarker(res, cache.getDifficulty(), cache.getTerrain(), false));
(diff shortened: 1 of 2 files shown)


eddiemuc:
I like it.
Maybe a tiny bit larger?

ztNFny:
> I like it.
> Maybe a tiny bit larger?

Not easily done. The scaling is hardcoded in MapMarkerUtils.

I can do it afterwards if this PR is well received, otherwise too much work to invest.

SammysHP:
Also `D` and `T` are kind of redundant with the labels. Not sure what lead to the removal of the stars, but I think the numbers plus a graphical representation like the stars or a bar graph would be better.

Of course this PR is still an improvement!

moving-bits:
while I see the resemblance with the markers on the map, I think the visual identifier is too small here, so not that much gain over the pure textual representation. IMHO the proposal of @bekuno given in https://github.com/cgeo/cgeo/issues/15505#issuecomment-2016784459 would be easier to grap quickly.

ztNFny:
as none of the alternatives proposed gained any track I went ahead and adjusted the size of the D/T marker in the popup to match that of the log icons.
PR is ready to merge then.

image

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
Retest this please

PR #15529, opened 2024-03-28
Unit tests on GitHub

kumy:
## Description
Create workflows to run the Unit tests from GitHub Workflows
- watchdog: scheduled every 4 hours
- avr caching: schedule once a day.
  - That should maintain pre-built AVR images in github cache
- unit tests
  - Configured to run on each PR
  - on push to branch master
  - code is tested in parallel in a matrix of Android API level 26 and 29
  - CI could be run manually but it need to be merged first
  - When run manually, the JDK version can be overridden
  - Full run takes approximately 6min
  - I already created all the secrets on GitHub side

Sample of [a successful run](https://github.com/cgeo/cgeo/actions/runs/8467215395)

Feel free to ask questions :)

## Related issues
:shrug:

## Additional context
- API 30 has been disabled in the build matrix as some tests are failing
- Built APK is attached to the [build details](https://github.com/cgeo/cgeo/actions/runs/8467143195)
image
- test results are attached to the [build annotation](https://github.com/cgeo/cgeo/actions/runs/8467143195)
image
- Builscan are [published to graddle.com](https://gradle.com/s/qsk3yzpfs27cs)
image
- test results in PR job status
image

## Notes
I didn't had much spec so I did how I feel it.
First time trying the PR feature, I hope it will work!

### DIFF
--- a/.github/workflows/watchdog.yml
+++ b/.github/workflows/watchdog.yml
@@ -3,7 +3,7 @@ name: Run watchdog tests
 on:
   push:
   schedule:
-    - cron:  '0 */4 * * *'
+    - cron: '0 */4 * * *'
   workflow_dispatch:
     inputs:
       JDK_VERSION:
@@ -28,6 +28,13 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ matrix.branch }}
+
+      # https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/
+      - name: Enable KVM group perms
+        run: |
+            echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+            sudo udevadm control --reload-rules
+            sudo udevadm trigger --name-match=kvm
 
       - name: Setup cgeo preferences and keystore
         uses: ./.github/actions/cgeo-preferences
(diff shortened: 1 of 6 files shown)


Lineflyer:
Thanks @kumy for your effort. That looks really promising.
I could have never dived into this so deep due to my limited knowledge.

Questions:
- Do Github Workflows have any limitation in terms of how often they can be run per day/hour/etc.?
- If so, is there any other (paid?) plan which could be helpful in case we run into those limitations?
- It looks like it covers all CI activities we need on daily basis besides building production releases (which is totally fine to keep limited to our CI server). Correct?
- Any technical or other issues which would prevent to merge and start using it in parallel to our CI until we might decide to disable the hooks towards the CI (in case we prefer the Github Workflow)?

What does the @cgeo/team  think about this approach for a slow migration to Github Workflows to either enhance or even replace the CI-server?
I would personally prefer the Github Workflows as we don't need to care about servicing the server and the applications, which at this moment is not covered very good by our resources.

eddiemuc:
Love it!

kumy:
1 and 2: you ll find your answers at https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits. 20 concurrent jobs is the most important limit.

3. It can also be possible, we need to store the signing key on GH secrets. The resulting artifact (apk) can be published to the GH tag/release assets. WF can be restricted to some users by using the "deployments" capabilities. We can do that later...

4.  I don t anticipate any issue.

Just to let you know, it s also possible to host our own GH runners, but that would not solve our Human resources issues...

kumy:
> start a trial phase on master?

Just to be 100% clear
It's not only "push to `master`", but also **any PR** (whatever the target branch is) + the manual run
image

Lineflyer:
@kumy 
Another specific question just coming to my mind:
Any chance you know, whether adding a new commit to an existing PR will re-trigger the test build and whether this can be surpressed by any means (e.g. a keyword or flag or a minimum time in between subsequent commits)?

Reason:
While in general it is meaningful to re-test if a PR was changed, we have an issue with our Crowdin integration which delivers a separate commit into the translation PR for each language. This means it could happen, that on creating a new PR with several outstanding language translations our workflows will queue 20-30 jobs and be busy for a long time.

Ideally we could add some kind of: Test PR once and run next test for this PR earliest after xx minutes/hours?

Lineflyer:
> Just to be 100% clear
> It's not only "push to `master`", but also **any PR** (whatever the target branch is) + the manual run

Does this work for `release` while the config files are only merged to `master` in this first step? Otherwise we would need to merge this PR to `release` and merge `release` upstream to `master`.

.github/workflows/watchdog.yml
```diff
@@ -0,0 +1,121 @@
...
+env:
+  JDK_VERSION: ${{ inputs.JDK_VERSION || '11' }}
+  API_LEVEL: 26
+
+jobs:
+  watchdog-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        branch: [ "main", "release" ]
```
Lineflyer, line 22: Should be `master` instead of `main`

kumy: true! good catch

Lineflyer:
> If a human push a new commit but he don't want to trigger the build, then add [skip ci] in the commit message.

Well, AFAIR I configured Crowdin to actually include [skip CI] already. It never worked to make our CI-server to skip testing. But maybe it works with the workflow, else your idea mentioned further below could be an alternative.

kumy:
> But maybe it works with the workflow

I have good hope :wink:
see https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs for the doc

Lineflyer:
Thanks @kumy again for preparing this.
During our monthly dev call we just decided to merge this and let it run in parallel to CI for a (yet undefined) period of testing and gaining more experience.

bekuno:
Hello @kumy 
The step 
(https://github.com/cgeo/cgeo/blob/381da0463147947fb5c3a329380c13aae25c6212/.github/actions/cgeo-preferences/action.yml#L181)
 does not work, see eg. https://github.com/cgeo/cgeo/actions/runs/8680942454/job/23802624507?pr=15580.
The system runs `base64 -d <<< > key.keystore`, so the variable was not parsed correctly.

What do we need to configure?

kumy:
@bekuno 😔 this PR has been created by dependabot. IIRC, has access to a special set of secrets which are distinct from normal GH Workflows.

We should define the secrets for it, what do you prefer, should we use all the same secrets values 🫣 or create distinct ones?

I also suppose PR from forked repositories may also don't have access to the secrets too 🤔

PR #15541, opened 2024-03-31
Move elevation info to position indicator (fix #14756)

### LINKED ISSUE #14756: Position of elevation info on map
> If we really don't want to have it close to the other numbers in top right position, an alternative approach could be to put it as default info at the bottom of the long press popup menu.

Its an option, the user actively needs to turn on, isnt it?
Then I dont see a problem showing it next to the position icon then. The user is interested in it and he can also easily move the map if something is covered below. 

The top right is IMHO not at all the right position for an information not related to a route. 

Shall we take this to a new issue?

_Originally posted by @Lineflyer in https://github.com/cgeo/cgeo/issues/14744#issuecomment-1774210148_

moving-bits:
## Description
as discusses in the related issue

example screenshot:
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
@@ -28,6 +28,7 @@ import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.PersistableFolder;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
 import cgeo.geocaching.ui.CoordinatesFormatSwitcher;
+import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.SimplePopupMenu;
 import cgeo.geocaching.utils.AndroidRxUtils;
@@ -42,9 +43,13 @@ import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.ComponentName;
 import android.content.Intent;
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.Paint;
 import android.graphics.Point;
 import android.text.Html;
 import android.text.Spanned;
+import android.text.TextPaint;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.TextView;
@@ -72,6 +77,9 @@ public class MapUtils {
     private MapUtils() {
         // should not be instantiated
     }
+
+    private static TextPaint elevationTextPaint = null;
+    private static Paint elevationPaint = null;
 
     // filter waypoints from owned caches or certain wp types if requested.
     public static void filter(final Set<Waypoint> waypoints, final GeocacheFilterContext filterContext) {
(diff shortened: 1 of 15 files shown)


Lineflyer:
Much better IMHO

main/src/main/java/cgeo/geocaching/maps/MapUtils.java
```diff
@@ -351,4 +359,38 @@ private static void addMenuHelper(final SimplePopupMenu menu, final int uniqueId
...
+        }
+
+        if (elevationPaint == null) {
+            elevationPaint = new Paint();
+            elevationPaint.setColor(res.getColor(R.color.osm_zoomcontrolbackground));
+            elevationPaint.setStrokeCap(Paint.Cap.ROUND);
+            elevationPaint.setStrokeWidth(textSizeInPx);
+        }
+
+        final Bitmap bm = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888);
```
eddiemuc, line 384: Maybe use `ImageUtils.createBitmapForText(final String text, final float textSizeInDp, @Nullable final Typeface typeface, @ColorInt final int textColor, @ColorInt final int fillColor)` instead?

main/src/main/java/cgeo/geocaching/unifiedmap/layers/PositionLayer.java
```diff
@@ -50,6 +53,16 @@ public PositionLayer(final AppCompatActivity activity, final GeoItemLayer<String
                                     .setBitmap(markerPosition).build())
                             .buildUpon().setZLevel(LayerHelper.ZINDEX_POSITION).build());
 
+            if (locationWrapper.location.hasAltitude() && Settings.showElevation()) {
+                layer.put(KEY_ELEVATION,
+                        GeoPrimitive.createMarker(new Geopoint(locationWrapper.location), GeoIcon.builder()
+                                        .setRotation(0f)
+                                        .setFlat(true)
+                                        .setBitmap(MapUtils.getElevationBitmap(activity.getResources(), markerPosition.getHeight(), locationWrapper.location.getAltitude()))
```
eddiemuc, line 61: Maybe use `GeoIcon.Builder.setText` instead? (Or use `TextBitmapProvider` directly to have more control)

eddiemuc (commented):
See comments

main/src/main/java/cgeo/geocaching/unifiedmap/layers/PositionLayer.java
moving-bits: thanks for the two pointers, will look into it

moving-bits: Ok, I got some results when using `GeoIcon.builder().setBitmap(new GeoIcon.TextBitmapProvider(...)`, but that leaves two questions open:
- Currently the elevation info gets centered beneath the position marker. How can I position the text "south" of the elevation marker? (In my bitmap variant I create a blank bitmap part the size of the marker as a spacer.)
- How can I create round edges for the background around the elevation info?

|current|instead of|
|---|---|
|image | image|

eddiemuc: For position I believe there's a method setYShift() or something like this on the builder

Round corners are not supported currently I am afraid. This must be added to the ImageUtils method and then also to the Text bitmap provider first.

If this is too complex, maybe leave I
PR as was before and I can refactor some other time. It's no biggie

eddiemuc: Update: the method I was looking for above is called `setYAnchor()`, but it is currently not suited for your use case (it is only implemented to move the image anchor within the image bounds), So also here adaption would be needed. As you already suggested: best merge this PR as is onto release, we can refactor later (on master).

PR #15556, opened 2024-04-05
Some fixes for TB parsing

moving-bits:
## Description
TB parsing fixes (mitigating the recent website changes)

- added aria attribute in trackable icon / type parsing (optional, since not used in all places)
- new HTML structure for log entities

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/CgeoApplicationTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/CgeoApplicationTest.java
@@ -143,7 +143,7 @@ public class CgeoApplicationTest {
         assertThat(tb.getOwner()).isEqualTo("Die Eybacher");
         assertThat(tb.getOwnerGuid()).isEqualTo("5888ea6c-413b-4b60-959d-d3d729ad642b");
 
-        // Following data can potentially change over time. However, is's very unlikely as the trackable is lost since years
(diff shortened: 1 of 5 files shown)


moving-bits:
All TB-related tests are back to green, thus removing "WIP" label / draft state from this PR

bekuno (changes requested):
Can you add the htm-sources for the test caches (TB2J1VZ, TB4BPQK) please?

moving-bits:
> Can you add the htm-sources for the test caches (TB2J1VZ, TB4BPQK) please?

why? Those TB are loaded online on each run, and results are compared against that. Nothing changed in that part of the tests.

This is different to the TB touched by this PR, those are compared against an offline copy.

bekuno:
> > Can you add the htm-sources for the test caches (TB2J1VZ, TB4BPQK) please?
> 
> why? Those TB are loaded online on each run, and results are compared against that. Nothing changed in that part of the tests.
> 
> This is different to the TB touched by this PR, those are compared against an offline copy.

The offline copy of this two caches would be interesting to compare them in a situation like today, so we can see what has changed.

moving-bits:
> The offline copy of this two caches would be interesting to compare them in a situation like today, so we can see what has changed.

ok, have added them as part of #15573

Issue #15562, opened 2024-04-06
Website changes: failing tests


bekuno:
Currently these two tests are failing.
(I don't saw a Issue for that)

~~~
13:10:39  cgeo.geocaching.connector.gc.GCConstantsTest > testCacheCountOnline[android-30(AVD) - 11] [31mFAILED [0m
13:10:39  	java.lang.AssertionError:
13:10:39  	Expecting:
~~~

~~~
13:11:54  cgeo.geocaching.log.LogTemplateProviderTest > testNoNumberIncrement[android-30(AVD) - 11] [31mFAILED [0m
13:11:54  	java.lang.NumberFormatException: For input string: ""
13:11:54  	at java.lang.Integer.parseInt(Integer.java:627)
~~~

moving-bits:
cache counter is fixed with #15569

Lineflyer:
Is that the reason the current nightly (without this fix) shows a find count which is some days behind?

MagpieFourtyTwo:
Cache counter wrong again. "Missing" 4 caches, so maybe it's been wrong for three days already, without me realizing it.

MagpieFourtyTwo:
Forgot to mention the version: Just updated to today's nightly (2024.04.13-NB-4ae1565) but counter is still 4 caches behind.

moving-bits:
> Cache counter wrong again. "Missing" 4 caches, so maybe it's been wrong for three days already, without me realizing it.

Yeah, cache counter source has changed again during the last few days. PR #15584 adjusts accordingly and should be part of next nightly and next release.

MagpieFourtyTwo:
Yep, have afterwards seen your comment over there. Nevertheless, thx for the info!

moving-bits:
Cache counter seems to be finally fixed with #15584, build is back to green since #15594, therefore closing here for now

PR #15569, opened 2024-04-07
Fix found counter detection (rel. to #15562)

moving-bits:
## Description
make found counter detection a bit more flexible to positional changes in JS array


eddiemuc:
Good change, thank you. Would make sense to apply same robustness to GC login at some point in the future

PR #15571, opened 2024-04-07
Improve selected position info

moving-bits:
## Description
Implements the ideas described in https://github.com/cgeo/cgeo/pull/15429#issuecomment-2002128498:
- Combines map long-tap menu entries for "Show coordinates" and "Elevation (map)" into a single "Selected position" entry
- Tapping on it opens a dialog with following options:
  - Display coordinates (format switchable)
  - Display elevation from map
  - Display distance to current position (straight line)

image

Closes #15429

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapUtils.java
@@ -22,6 +22,7 @@ import cgeo.geocaching.models.IndividualRoute;
 import cgeo.geocaching.models.RouteItem;
 import cgeo.geocaching.models.RouteSegment;
 import cgeo.geocaching.models.Waypoint;
+import cgeo.geocaching.sensors.LocationDataProvider;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.ContentStorage;
 import cgeo.geocaching.storage.DataStore;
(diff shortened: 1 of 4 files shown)


moving-bits:
failing tests are unrelated

ztNFny:
Thanks for taking care of this, I just didn't have the time for this yet.
One thing I had in mind was adding icons to the dialog (maybe instead of the text, not sure, have to try it) - I can take care of that after this PR.

ztNFny:
Text alignment is broken. It's right-aligned to the widest text, not the.dialog

moving-bits:
This was intentional, as having text/values unaligned looked pretty ugly, and having them aligned to the dialog's borders made them being very far apart from each other, looking unrelated.

(But this could be changed, if preferred, or even won't happen at all if we switch to icons instead of text. The last two lines could then be put into a single line.)

moving-bits:
@ztNFny 
Maybe something like this?

image

(is a hand-crafted bitmap for now)

maybe skip the little separator dot between elevation and distance?

ztNFny:
> This was intentional, as having text/values unaligned looked pretty ugly, and having them aligned to the dialog's borders made them being very far apart from each other, looking unrelated.

ok, I'll try to come up with something

PR #15573, opened 2024-04-07
Fix update_trackables script + update downloaded tb files

moving-bits:
## Description
- adjusts `update-trackables` script to changed source folder structure and adds some progress info
- updates existing tb file copies
- adds tb file copies for tb2j1vz and for tb4bpqk as requested in https://github.com/cgeo/cgeo/pull/15556#pullrequestreview-1984697976


bekuno (approved):
LGTM

PR #15584, opened 2024-04-10
Findcount adaption (simplified version)

moving-bits:
## Description
"findCount" source seems to have changed again (or there are variations depending on login state)


ztNFny:
it'd seem a simpler version would work just as well:
\"findCount\":\\s*([0-9]+)

moving-bits:
sure, but I wanted to have some sort of "safeguard" (context) around to avoid false positives

moving-bits:
> I just realized public static final Pattern PATTERN_LOGIN_NAME_CACHE_COUNT = Pattern.compile(""publicGuid":"[^\"]+","referenceCode":"[^\"]+","id":[0-9]+,"username":"([^\"]+)","dateCreated":"[0-9:T-]+","findCount":([0-9]+),"); already contains the find count. So no need for a second expression - or this one should be adjusted as well?

good catch! But this one does work on the login page only, but not on other pages. `getCachesCount` will be called in at least two situations: login and loading a cache page from the server. Provided source for `findCount` is different (eg: "username" and "findCount" in different order!), so we either have to use two separate patterns, or we will find some common abstraction. I'll investigate further... (but not today)

If we want to have a working findCount for the next nightly already: I have updated the PR to implement @bekuno's request.

ztNFny:
> loading a cache page from the server

On a normal cache page neither of those expressions is a match for me?
I have this in source: image

I'd really suggest to go with a much simpler expression of \"findCount\":\s*([0-9]+)[,\s]+

moving-bits:
I have adapted the pattern to the simplified version suggested by @ztNFny

moving-bits:
failing tests are unrelated, so lets merge

PR #15594, opened 2024-04-13
Add missing tb discovered/retrieved block to test data

moving-bits:


moving-bits:
Build is back to green :-)

PR #15603, opened 2024-04-14
some cleanups

bekuno:
- remove unused import
  - string "context_map_show_coords" is unused
  - style "cgeo.fullScreenDialog2" is unused
  - let "expand_collapsed" unused in code - @ztNFny ?
  - let "ic_menu_note" unused in code - @ztNFny ?


ztNFny:
expand_collapsed was used in an earlier version of SimpleDialog, can be removed, can be easily added again if needed

ic_menu_note I'd propose to leave in, it's unused but the complimentary icon to ic_menu_note_edit, not unlikely we could need that again.

bekuno:
> expand_collapsed was used in an earlier version of SimpleDialog, can be removed, can be easily added again if needed
> 
> ic_menu_note I'd propose to leave in, it's unused but the complimentary icon to ic_menu_note_edit, not unlikely we could need that again.

Ok, done.

Issue #14950, opened 2023-12-03
No PM detection on first login after version upgrade


Lineflyer:
### Describe your problem!

Since a few days to weeks I noticed, that each time I start up c:geo after installing a new nightly, it shows my membership state as "Basic" on the start screen.
After killing and restarting c:geo it correctly detects me as "Premium".

### How to reproduce?

- Be logged in as premium member (additionally I also have OC.de activated...not sure if this can lead to a different behavior)
- Download new nightly version via in app notification
- Start installation of new nightly via tapping the APK in download manager
- After upgrade has finished tap on "Open"

### Actual result after these steps?

c:geo is starting up and shows "Login OK * Basic".
Also all PM feature (like upload personal note) are missing.

### Expected result after these steps?

c:geo should show "Login OK * Premium" and PM functions should be available

### Reproducible

Yes

### c:geo Version

2023.12.03-NB

### System information

```text
Tested on Android 13
```

alan666notb:
Ah - interesting, I've noticed this as well and assumed it was just a 'glitch'.
This is on latest release & beta versions.

Lineflyer:
Still happening and the main reason for support tickets after a version upgrade: "I installed the new version and now I can no longer use the premium features".

As a relogin is required the problem stays several days for users except if they force a new login.

moving-bits:
To test this behavior: How can I force that situation in the emulator?
Removing credentials leads to a relogin, but here PM-detection seems to work for me.

Lineflyer:
Maybe a fresh install on emulator?

moving-bits:
> Maybe a fresh install on emulator?

Have tried this, but this triggers the initial authentication process, which seems to be working.

I have no idea currently what's different (in terms of logging in) after a version upgrade.

fm-sys:
We have a method inside the code which checks if the version was upgraded and calls the relevant actions. Let it always return true and you should be fine for testing. (just a guess)

Lineflyer:
> I have no idea currently what's different (in terms of logging in) after a version upgrade.

Just a blind guess:
Maybe some cached data (cookies?) no longer available?

PR #15604, opened 2024-04-14
reset login info on encountering race condition

moving-bits:
as discussed in today's dev call: reset internal login data structure when encountering a race condition while logging in
hopefully mitigates #14950


bekuno:
Retest rhis please

bekuno:
Retest this please

Issue #15611, opened 2024-04-15
[UnifiedMap] Setting "hide visited waypoints" does not work for showing single cache on map


murggel:
### Describe your problem!

When showing a single cache on the map, the option to hide / show visited waypoints does not work.

### How to reproduce?

Open Cache in detail view
Set some waypoint to visited
Navigate to the cache in map
Open settings-popup-dialog
deactivate option to show visited waypoints

### Actual result after these steps?

All waypoints are shown

### Expected result after these steps?

Only not visited waypoints should be shown

### Reproducible

Yes

### c:geo Version

2024.04.04 Nightly

### Additional Information

Somehow related to #15614

murggel:
PR #15659 will hopefully fix this issue

PR #15612, opened 2024-04-15
Fix filtering waypoints for current cache in map-view

murggel:
## Description

On mapping a single cache filter the waypoints in `reloadCachesAndWaypoints`

## Related issues

related #15611

## Additional context

does not fix the refresh problem on changing the setting in the settings-popup-dialog


murggel:
retest this please

Issue #15608, opened 2024-04-15
Further steps for CI functions with GH actions


bekuno:
After merging PR #15529 the configuration needs some tuning.

- [x] configure secrets for normal workflow
- [x] configure secrets for workflows from dependabot (if necessary)
- [x] Fix the bug "emulator won't stop" ReactiveCircus/android-emulator-runner#385
- [x] clarify, if workflow(s) should not run from cloned repositories - and if yes, configure it
- [ ] create documentation to configure secrets for cloned repositories

(Feel free to extend the list)

bekuno:
For failing standard test see https://github.com/cgeo/cgeo/actions/runs/8693327234/job/23839868312?pr=15607

Lineflyer:
AFAICS the secrets are already stored in our repo key/secrests storage.
If this problem only comes up for builds on forks (looks a bit like that) we should disable this feature. We do not need that and it blocks resources.

bekuno:
> AFAICS the secrets are already stored in our repo key/secrests storage. If this problem only comes up for builds on forks (looks a bit like that) we should disable this feature. We do not need that and it blocks resources.

The failing test above links to a PR **in the c:geo repo** for master against the source in a cloned repo (typically for a normal PR).

Lineflyer:
mmh...thats indeed a pretty normal case.
I would have thought, that is uses the secret of the target repo rather than requiring the secrets to be located in the source/fork repo.

fm-sys:
Might be a security feature. Otherwise developers might introduce malicious code to retrieve the secrets... Still problematic for our workflow.

fm-sys:
The relevant part out of it:

> workflows from forks do not have access to sensitive data such as secrets

kumy:
Yes this is our issue, a security issue.
@bekuno From our mails, you wanted to disable such jobs for PR from external repositories, am I right?

I did a small [POC](https://github.com/geokrety/just-a-test/blob/main/.github/workflows/android.yml#L8) to test the expression required in such case
- main repo: https://github.com/geokrety/just-a-test
- forked repo: https://github.com/kumy/just-a-test

It allows:

- all jobs to execute on [PR](https://github.com/geokrety/just-a-test/pull/1) from the same repo
- [All jobs on push](https://github.com/geokrety/just-a-test/actions/runs/8711014311)
- [PR from forks repo](https://github.com/geokrety/just-a-test/pull/2) can skip same "secured steps"
[...]

PR #15625, opened 2024-04-16
Use pull_request_target to allow secrets access

Lineflyer:
See: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,7 @@ name: Unit tests
 on:
   push:
     branches: [ "master", "androidci" ]
-  pull_request:
+  pull_request_target:
   workflow_dispatch:
     inputs:
       JDK_VERSION:


Lineflyer:
@kumy 

I assume this is the right way to overcome the secret sharing problem for forks.

The reason its not working yet for this very PR is possibly this restriction (mentioned in the above link). Thus I assume it will work on the next PR after this change is merged to our repo:

> pull_request_target runs in the context of the target repository of the PR, rather than in the merge commit. This means the standard checkout action uses the target repository to prevent accidental usage of the user supplied code.

I think we can go this way without to much risk of disclosing our secrets as I also just configured our actions to only run if triggered by a organisation member. Any outside collaborator PR should need pre-approval by the team AFAIU the setting:

image

Lineflyer:
> pull_request_target runs in the context of the target repository of the PR, rather than in the merge commit. This means the standard checkout action uses the target repository to prevent accidental usage of the user supplied code.

Rethinking this:
This probably means, that the code change contained in a PR coming from a fork repository will never be tested, but this is our usecase.
If this is the case, we could only do the following:
- Use pull_request_target to trigger a approval
- When approved the worklof should run in the source repository context (merge commit) if I understand the linked article correctly

kumy:
@Lineflyer no, please don't do that :pray: 
I read somewhere (need to find my source) that `pull_request_target` action is the implementation before the security issue was discovered. So you'll be rolling back the security fix.

See https://github.com/cgeo/cgeo/issues/15608#issuecomment-2059773013 for another way of preventing failures.

> configured our actions to only run if triggered by a organisation member. Any outside collaborator PR should need pre-approval by the team

:+1: really nice

Lineflyer:
@kumy  Oh, that happened at the same time. I will read through that first and ocmment over there.

Issue #15540, opened 2024-03-30
"add more lists" in filter > stored lists is unstyled


ztNFny:
List selection dialog should use the new grouped style with list icons as well, instead it's a plain list.

PR #15627, opened 2024-04-16
style list selection dialog in "Stored Lists" filter #15540

ztNFny:
before:
image

after:
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/gui/CheckboxFilterViewHolder.java
+++ b/main/src/main/java/cgeo/geocaching/filters/gui/CheckboxFilterViewHolder.java
@@ -2,7 +2,6 @@ package cgeo.geocaching.filters.gui;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.filters.core.IGeocacheFilter;
-import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.ui.ImageParam;
 import cgeo.geocaching.ui.TextParam;
@@ -20,7 +19,6 @@ import androidx.annotation.NonNull;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -32,9 +30,9 @@ import org.jetbrains.annotations.Nullable;
 
 public class CheckboxFilterViewHolder<T, F extends IGeocacheFilter> extends BaseFilterViewHolder<F> {
 
-    private final ValueGroupFilterAccessor<T, F> filterAccessor;
+    final ValueGroupFilterAccessor<T, F> filterAccessor;
 
-    private final Set<T> visibleValues = new HashSet<>();
+    final Set<T> visibleValues = new HashSet<>();
     private final Map<T, ImmutablePair<View, CheckBox>> valueCheckboxes = new HashMap<>();
 
     private ImmutablePair<View, CheckBox> selectAllNoneCheckbox;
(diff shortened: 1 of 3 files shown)


ztNFny:
> See my comment in code.
> 
> In general I believe this only changes the "add select item" dialog and not the view of selectable items. Correct?

Correct. The other view probably doesn't contain to many items so i thought no need to change that.

moving-bits:
Setting WIP label for now to wait for requested sub-classing

moving-bits:
Thanks for implementing subclassing. Can you have a look at the Codacy issues, please? (as time permits)

moving-bits:
> > Thanks for implementing subclassing. Can you have a look at the Codacy issues, please? (as time permits)
> 
> It's not implemented, can't make it work, see comment

Ah, sorry, then I had misinterpreted the commit. Which comment are you referring to?

ztNFny:
https://github.com/cgeo/cgeo/pull/15627#pullrequestreview-2015384835

moving-bits:
this is strange: When I click on your link, only a small popup opens, but it does not move to any comment:

image

I'm not sure if this is a reply to eddiemuc's review comment further up - for me it does not show any inline reply to it, only your reply to this other question (https://github.com/cgeo/cgeo/pull/15627#issuecomment-2060386200)

ztNFny:
line 79: ((CheckBox) getValueCheckbox((T) value).right).setChecked(true);

any idea why i get java.lang.ClassCastException: java.util.HashSet cannot be cast to cgeo.geocaching.list.StoredList here on submitting the dialog? The same code works fine in the non-sub-class.

eddiemuc:
> line 79: ((CheckBox) getValueCheckbox((T) value).right).setChecked(true);
> 
> any idea why i get java.lang.ClassCastException: java.util.HashSet cannot be cast to cgeo.geocaching.list.StoredList here on submitting the dialog? The same code works fine in the non-sub-class.

The expression in line 79 is the body part of a lambda passed to the multiprompt function of StoredList.UserInterface. looking at that function, "value" is a set of integers (the selected item indexes I assume), so trying to cast it to StoredList (T is StoredList in this context) naturally results in cce. I can't imagine that this worked before.

main/src/main/java/cgeo/geocaching/filters/gui/StoredListCheckboxFilterViewHolder.java
```diff
@@ -0,0 +1,85 @@
...
+public class StoredListCheckboxFilterViewHolder<T, F extends IGeocacheFilter> extends CheckboxFilterViewHolder {
+    public StoredListCheckboxFilterViewHolder(ValueGroupFilterAccessor filterAccessor) {
+        super(filterAccessor);
+    }
+
+    public StoredListCheckboxFilterViewHolder(ValueGroupFilterAccessor filterAccessor, int colCount, Set alwaysVisibleItems) {
+        super(filterAccessor, colCount, alwaysVisibleItems);
+    }
+
+    public View createView() {
```
eddiemuc, line 29: It should not be necessary to copy complete code of this method into subclass just to change a few lines. I recommend to abstract the superclass.better by introducing a new method e.g. to get the function to place on the button "add items", then just implement this method in the subclass. Currently the subclass contains must duplicated code, this should be avoided.

ztNFny:
> . I can't imagine that this worked before.

Well, it does...

eddiemuc:
(my GitHub seems to be a bit messed-up right now. I just got an email notification about a reply by @ztNFny but can't find it here. Also, like @moving-bits , I fail to find the earlier code comment reply of  @ztNFny. I am currently restricted to GitHub Mobile app, maybe there's some flaws in there)

@ztNFny would you be able to pass a version of the non-subclassed version with same working line for anayzation (only if you're interested)? I can't explain why this would work, but I may overlooking something.

ztNFny:
I'm confused about those comments too. Both are still visible on the website for me but not in the app. Should probably use only "normal" comments for now, seems there's something broken.

For the issue: I was 100% sure I had it working with "that integer", but now that I reapplied that commit to my dev branch it crashs. Very strange ... Anyway, I looked into that and got a working version now. Will now look into "proper" subclassing.

ztNFny:
> ready to go

oops, forgot to remove the WIP label

moving-bits:
> I'm confused about those comments too. Both are still visible on the website for me but not in the app.

maybe you started a "review", but did not finish it? GitHub web seems to remember those comments internally, but they are not visible to others until you finish the review. - Alternatively use the "add single comment" button to immediately store the comment.

PR #15638, opened 2024-04-18
fix deleting own logs on GC

### LINKED ISSUE #15587: Error when deleting owner log
Hello I tried to use delete option delete option in cgeo to delete my own owner log (note) but it didn't work. I had to use web version to do it and there it worked. I think that the request url/parameters are different when owners want to delete logs and there also shouldn't be field reason shown in c:geo in this case.

ztNFny:
fixes #15587

for own logs NO body must be provided, for other logs a body must be provided


eddiemuc (approved):
Thanks!

Issue #15620, opened 2024-04-16
Same route with both bike and car profiles


andrixnet:
### Describe your problem!

Navigating towards a more distant cache, c:geo gave me erroneous results for routing. 
The destination cache was 27km away. Profile was car. 
The generated route was definitely not a car route because near the start it crossed the street and sidewalks in places where there was bike lane and definitely not passable by car. 

### How to reproduce?

Either internal or Brouter app.
Car profile.
Having bike lanes in en-route that follows paths not usable by cars.
Likely dependent on geography.
Navigate to a more distant cache.

### Actual result after these steps?

I obtained a very complicated route that is not usable by a car in places.
I obtained the same route with both car and bike profiles using internal routing.
I obtained the same route using BRouter app set to car-fast profile. 
Routing map data files up to date.

![Screenshot_20240414-151217_cgeo](https://github.com/cgeo/cgeo/assets/3382777/6b483222-6726-449b-85ea-6dbeae759af1)
![Screenshot_20240414-151314_cgeo](https://github.com/cgeo/cgeo/assets/3382777/e7ff01e6-8de6-4c5e-97f7-f82c01c93ffb)

### Expected result after these steps?

First and foremost a car usable route using car profile.
Having bike lanes in the area, use them with bike profile but not car.

### Reproducible

Yes

### c:geo Version

2024.04.13

### System information

```text
## System information

c:geo version: 2024.04.13

Device:- Device type: SM-G965F (star2ltexx, samsung)
- Available processors: 8
- Android version: 9
- Android build: PPR1.180610.011.G965FXXS7CSK4
- Screen resolution: 1440x2792px (411x773dp)
- Pixel density: 3.8500001
- System font scale: 1.1 / used scale: 1.1
- Sailfish OS detected: false
- Google Play services: enabled - 24.12.17 (100400-623887440)
- Memory: Available:2.2 GB, Total:5.5 GB, Threshold: 216.0 MB, low:false

Sensor and location:- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:- Settings: v9, Count:220
- Set language: en_GB (system default)
- System date format: dd/MM/y
- Time zone: GMT+03:00
- Debug mode active: no
[...]

PR #15647, opened 2024-04-19
Append configured routing modes to system information

moving-bits:
related to #15620/#15621 and others


moving-bits:
failing test is unrelated

PR #15651, opened 2024-04-20
rebrand gradle enterprise to develocity

bekuno:
Develocity is the new company name for gradle.
The renaming here is part of the rebranding.


bekuno:
retest this please

bekuno:
DNM.
I found the error source description at https://docs.gradle.com/develocity/gradle-plugin/current/#publishing_on_demand:
~~~
Interoperability with Gradle 8.4 and older

If the build does not apply the plugin, Gradle will automatically apply the most recent version available at the time that version of Gradle was released.

Gradle versions older than 8.4 do not recognize the Develocity plugin as the replacement for deprecated Gradle Enterprise and Build Scan plugins. When --scan argument is used, Gradle attempts to apply its version of the plugin which may conflict with the Develocity plugin explicitly applied to your project. If the plugin applied by Gradle wins, the new functionality provided by the Develocity plugin will not be available.

This case will be highlighted using the following message:

WARNING: Deprecated Gradle Enterprise plugin [version: 3.16.2] was applied alongside Develocity plugin.
If your build uses the latest Develocity functionality, it may fail or not run as expected since the deprecated plugin may not support requested features.
The simultaneous application of Develocity and Gradle Enterprise plugins will cause the build to fail in a future release of the Develocity plugin.

To resolve the issue, upgrade to a newer Gradle version or configure on-demand publication using one of the options described in [Publishing based on criteria](https://docs.gradle.com/develocity/gradle-plugin/current/#publishing_based_on_criteria) section.
~~~

bekuno:
Lets do it fresh later.

PR #15653, opened 2024-04-21
Refine username detection during login

moving-bits:
## Description
- do not terminate detection on JS-escaped double quotes in username
- check for and unescape JS-escaped Unicode characters
- along the way: shortens username detection regexp to look for username only, as findcount is detected otherwise meanwhile

(switching to reading `serverParameters` seems to be non-feasable here, as it is not yet available during login)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
@@ -95,11 +95,11 @@ public final class GCConstants {
 
     //Try to find pattern as follows:
     //"publicGuid":"abc.def","referenceCode":"PRxyz","id":1234567,"username":"user","dateCreated":"2012-01-21T20:51:14","findCount":15123,
-    public static final Pattern PATTERN_LOGIN_NAME_CACHE_COUNT = Pattern.compile("\"publicGuid\":\"[^\"]+\",\"referenceCode\":\"[^\"]+\",\"id\":[0-9]+,\"username\":\"([^\"]+)\",\"dateCreated\":\"[0-9:T-]+\",\"findCount\":([0-9]+),");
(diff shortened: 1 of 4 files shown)


main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
```diff
@@ -95,11 +95,11 @@ public final class GCConstants {
 
     //Try to find pattern as follows:
     //"publicGuid":"abc.def","referenceCode":"PRxyz","id":1234567,"username":"user","dateCreated":"2012-01-21T20:51:14","findCount":15123,
-    public static final Pattern PATTERN_LOGIN_NAME_CACHE_COUNT = Pattern.compile("\"publicGuid\":\"[^\"]+\",\"referenceCode\":\"[^\"]+\",\"id\":[0-9]+,\"username\":\"([^\"]+)\",\"dateCreated\":\"[0-9:T-]+\",\"findCount\":([0-9]+),");
+    public static final Pattern PATTERN_LOGIN_NAME1 = Pattern.compile(",\"referenceCode\":\"[^\"]+\",\"id\":[0-9]+,\"username\":\\s*\"(([^\\\\\\\"]*(\\\\[a-z\\\"])*)*)\\\",");
     public static final Pattern PATTERN_FINDCOUNT = Pattern.compile("\"findCount\":\\s*([0-9]+)[,\\s]");
```
bekuno, line 99: I have an idea to recognise that we are in the right place (block), but without the need for an ordered position (reference to another entry).
The block starts with `window['chromeSettings'] = {`, so we could use something like
`window\['chromeSettings'\].*"username":\s*\"(([^\\\\\"]*(\\[a-z\"])*)*)\"`
Maybe additional escapes are necessary.

Issue #15648, opened 2024-04-19
Cleanup csrf calls in logging


ztNFny:
#15637 left the old csrf retrieval method in code in case GC reverts the change as they recently do quite often. 
When the site is stable the old calls should be removed. Maybe in 2 weeks?

moving-bits:
is this fixed with #15658 or is something remaining to be done here?

PR #15658, opened 2024-04-22
fix TB logging, remove legacy code, only fetch TB log page if required

ztNFny:
fix retrieving the CSRF token for some TB logs

make getting the TB html page conditional as it's only required for RETRIEVE logs

all log pages seem stable in their use of the new CSRF call now, so removing the legacy code, fixing https://github.com/cgeo/cgeo/issues/15648


main/src/androidTest/java/cgeo/geocaching/connector/gc/GCLogAPITest.java
```diff
@@ -31,7 +31,7 @@
 
 public class GCLogAPITest {
 
-    private static final String TEST_GEOCODE = "GC8H4YX"; // Am Hart
+    private static final String TEST_GEOCODE = "GC7E3PE"; // Am Hart
```
bekuno, line 34: There is no cache GC7E3P.

bekuno (changes requested):

ztNFny:
retest this please

bekuno:
retest this please

bekuno (approved):
Let's check it with the nightly watchdog.

PR #15659, opened 2024-04-22
Refresh map data after opening / closing settings dialog

### LINKED ISSUE #15611: [UnifiedMap] Setting "hide visited waypoints" does not work for showing single cache on map

### LINKED ISSUE #15614: [UnifiedMap] Changing options in map-settings-dialog does sometimes not trigger a refresh
### Describe your problem!

When mapping a list in unified map and changing the filter / settings via settings-popup-dialog, the caches and waypoints are not correctly shown

### How to reproduce?

Have a list with found caches with visited waypoints and caches with e.g. parking
Map the list
Change setting in popup-dialog: hide found caches and hide visited waypoints
-> the found cache is not shown, but still the waypoints
Change setting back to "show found caches"
-> the found cache is not shown again

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2024.04.04 Nightly

### System information

_No response_

### Additional Information

Somehow related to #15611 

#15576 will not fix this issue here

### LINKED ISSUE #15661: Quick filter not working properly on mapped lists
### Describe your problem!

Reported on support mail and reproducible for me with nightly build 2024.04.24-NB

When changing visible cache types via quick filter on a mapped list only the first change seems to be applied.

### How to reproduce?

- Have a list with at least one found cache, one offline logged cache, one archived caches and one unfound cache
- Map the list
- Open quick settings and e.g. disable archived caches (not to be shown)
- The archived caches is filtered out correctly
- Open quick settings and enable archived caches again (to be shown)

(Same sequence applies also to other filters (found, offline logged))

Archived caches still hidden

Archived caches should be shown again corresponding to the quick filter setting

### Reproducible

Yes

### c:geo Version

2024.04.18

### System information

_No response_

### Additional Information

Problem is specific to mapped lists and does not happen on a live map as fas as I tested

Support ticket 300052

murggel:
## Description

With my limited understanding of the BackgroundHandler and loading caches - but with debugging and trying:
If no `loadInBackgroundHandler` is set, then the caches / waypoints have to be reloaded via `reloadCachesAndWaypoints` to adapt to the possible changes made in the settings-popup.
Reloading and recreating the loadInBackgroundHandler is only done, if the relevant settings have changed.
Therefore I enhanced the `SettingsCheckboxModel` with the possibility to get that information.

## Related issues

fix #15611
fix #15614
fix #15661

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/AbstractMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/AbstractMap.java
@@ -123,7 +123,7 @@ public abstract class AbstractMap {
         //
     }
 
-    public abstract void refreshMapData(boolean circlesSwitched);
+    public abstract void refreshMapData(boolean circlesSwitched, final boolean filterChanged);
 
     public boolean isTargetSet() {
         return /* StringUtils.isNotBlank(targetGeocode) && */ null != lastNavTarget;
--- a/main/src/main/java/cgeo/geocaching/maps/AbstractMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/AbstractMap.java
@@ -123,7 +123,7 @@ public abstract class AbstractMap {
         //
     }
 
-    public abstract void refreshMapData(boolean circlesSwitched, final boolean filterChanged);
+    public abstract void refreshMapData(boolean circlesSwitched, boolean filterChanged);
 
     public boolean isTargetSet() {
         return /* StringUtils.isNotBlank(targetGeocode) && */ null != lastNavTarget;
--- a/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/CGeoMap.java
@@ -790,7 +790,7 @@ public class CGeoMap extends AbstractMap implements ViewFactory, OnCacheTapListe
             if (mapOptions.isLiveEnabled) {
                 mapOptions.isStoredEnabled = true;
                 mapOptions.filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.LIVE);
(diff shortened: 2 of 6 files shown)


murggel:
@eddiemuc / @moving-bits please check if my assumption and fix is correct

ztNFny:
Could you trigger the reload only if a relevant setting has changed? Changing e.g. navigation mode shouldn't trigger the reload.

murggel:
> Could you trigger the reload only if a relevant setting has changed? Changing e.g. navigation mode shouldn't trigger the reload.

Thought about that already, but the only changed-info is about the circles. 
For live map the LoadInbackgroundHandler.load is also called, if nothing changed, but there is some mechanism with useLastSearch, but even this is false, if the pop-up is closed without a change.

I will try, if I can find an easy way to transport/ transfer the information, if a setting is changed, which needs to update the map-data...

murggel:
Probably I can check the resulting filter....

murggel:
@eddiemuc  / @moving-bits can someone has a look on it please...

moving-bits:
Let's give it a try in the nightlies. Thanks for diving into it!

murggel:
I updated to the current nightly and I get crashes of unified map, not sure if they are related to this PR, did not have time yet to check it

PR #15690, opened 2024-05-04
fix #15689: enable finding archived owner caches again (introducing parameter sa=1 in gc.com websearch)

### LINKED ISSUE #15689: GC.com owner search: archived caches no longer found
[Ticket #226966]
, cgeo 2024.04.25

Searching for an owner's caches does not return archived caches any more.

Inspection reveals that this is due to a new parameter in the web search (sa=1) which apparently has to be set for such searches to include archived caches as well.

eddiemuc:
fix #15689: enable finding archived owner caches again (introducing parameter sa=1 in gc.com websearch)


jakedot:
LGTM

moving-bits:
retest this please

eddiemuc:
This PR also removes LoadCachesHandler from usage in CacheListActivity since the usage of this handler interfered badly with cache list reloading on filter change. The PR does not yet remove the Handler itself so we're able to restore the old logic quickly in case something unforeseen happens.

eddiemuc:
retest this please

PR #15693, opened 2024-05-05
UnifiedMap VTM: Add markers in batches (rel. to #15462)

moving-bits:
## Description
For reference purposes only - this is a technology study, not an actual PR, therefore setting "Do not merge" label.

Brings down marker displaying times on VTM from 140s to app. 10s (9k markers) in my (emulated) test environment.

For details see https://github.com/cgeo/cgeo/issues/15462#issuecomment-2094820119


moving-bits:
superseded by #15694

PR #15694, opened 2024-05-05
rel to #15462: add batch-processing of marker add/remove for VTM

eddiemuc:
rel to #15462: add batch-processing of marker add/remove for VTM

This PR creates its own version of the VTM MarkerLayer to allow adding/removing items w/o triggering a costly "populate" with each item action. "populate" is called once per batch only.

Please review carefully.


moving-bits (approved):
Clever yet simple approach! LGTM, thanks for diving into it.

PR #15718, opened 2024-05-10
Enable install wizard credential check for OC connectors

### LINKED ISSUE #15702: First use experience - Wizard shows error if only OC account is configured
### Describe your problem!

I introduced c:geo to a new user today, which does only want to use Opencaching.
While going through the configuration wizard all works well, but the summary at the end shows "Geocaching Server: FEHLER".

It seems to be a requirement for this check to pass, that a geocaching.com account is configured.

### How to reproduce?

- Install c:geo
- Run installation wizard
- Do not configure a geocaching.com account but only an opencaching account
- Complete wizard

Wizard summary screen shows "Geocaching Server: FEHLER"

Wizard should not report this error if at least one of the supported geocaching services has been successfully configured

### Reproducible

Yes

### c:geo Version

2024.05.24

### System information

```text
Android 13 device
```

### Additional Information

_No response_

moving-bits:
(fix #15702)

## Description
Adds credentials check for OC based connectors


moving-bits:
Retest this please

PR #15719, opened 2024-05-11
Export individual route as track with single segment (fix #15684)

### LINKED ISSUE #15684: Strange double track segment in GPX export
### Describe your problem!

When exporting a GPX track from c:geo, the code is strange, for example:

```
<trk>
    <name>c:geo individual route 2024-05-02T03:16:57Z</name>
    <trkseg>
      <trkpt lat="47.50146" lon="11.137798">
        <name>COORDS! N47 30,088 E11 08,268</name>
      </trkpt>
    </trkseg>
    <trkseg>
      <trkpt lat="47.501479" lon="11.137816" />
      <trkpt lat="47.501514" lon="11.137737" />
      <trkpt lat="47.501573" lon="11.137696" />
      <trkpt lat="47.501622" lon="11.137606" />
      <trkpt lat="47.501724" lon="11.137559" />
```

I don't understand why there is this strange first track segment with only one point.

**One big issue with this is, that Garmin Connect cannot import such GPX files.
Either it only imports the first, quasi empty track segment. Or it bugs completely and does not import at all.**

I experimentally removed this first track segment, and now Garmin Connect can import. 

_(I just got my first Garmin GPS watch, so I am not yet sure if this is the only reason why c:geo exports fail in Connect, but in this one example, the extra track segment definitely was the cause, and it worked after removing that.)_

### How to reproduce?

See above.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2024-04-25

### System information

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Use only a single track segment when exporting an individual route as track

### DIFF
--- a/main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
+++ b/main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
@@ -88,10 +88,12 @@ public class IndividualRouteExportTask extends AsyncTaskWithProgress<RouteSegmen
 
             gpx.startTag(NS_GPX, exportAsTrack ? "trk" : "rte");
             XmlUtils.simpleText(gpx, NS_GPX, "name", "c:geo individual route " + timeInfo);
+            if (exportAsTrack) {
+                gpx.startTag(null, "trkseg");
+            }
             for (RouteSegment loc : trail) {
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

PR #15766, opened 2024-06-08
ticket #926180: remove separate log length check

eddiemuc:
ticket #926180: remove separate log length check

c:geo contained a separate log length check in addition to the "log text counter" in the text field. This length check counted line break as two characters. This seems to be a legacy thing, as of now gc.com counts line breaks only as one character. Thus this PR removes the separate check.

This solves a problem reported in the mentioned support ticket with log lengths close to 5000 characters


eddiemuc:
Reference: https://www.geoclub.de/forum/t/c-geo-aktuelle-bugs.67870/page-44#post-1385743

PR #15768, opened 2024-06-08
Google Map: add 'Details' theme

eddiemuc:
Google Map: add 'Details' theme


eddiemuc:
Support ticket #15768 [816040](https://cgeosupport.droescher.eu/scp/tickets.php?id=7969)

It is lined out there that on Google Maps detailled POIs like restaurants are not (resp no longer?) shown. Example screenshot (area around cache GCTD8F):
image

This PR adds a new Google Maps Theme "Details" which, if activated, explicitely sets all google maps features to visible:
image

Same area as above with this theme enabled::
image

moving-bits:
Thanks for looking into this.

Wouldn't that feature be relevant for all existing themes?
So instead of creating a separate theme for it, how about making it a theme option which can be applied to all GM themes?

eddiemuc:
> Thanks for looking into this.
> 
> Wouldn't that feature be relevant for all existing themes?
> So instead of creating a separate theme for it, how about making it a theme option which can be applied to all GM themes?

Not so easy. I don't know how the existing themes were created tbh, but they partly also set poi-related Google Maps properties. I have no idea how this interrelates. I would suggest to make a more fine-tuned and better-designed feature only on `master`. And this might need more discussion too.

PR #15782, opened 2024-06-11
fix #15618: Don't overwrite sortContext 

### LINKED ISSUE #15618: Search results for a location no longer sorted by distance to target location
### Discussed in https://github.com/cgeo/cgeo/discussions/15616

<div type='discussions-op-text'>

<sup>Originally posted by **Lineflyer** April 16, 2024</sup>
We got a ticket on support, claiming that sorting a location search result by distance to that location does **no longer** work but the results are shown with distance to your current position.
</div>

<br>

We indeed had that feature (PR #11691) so this is clearly a regression and should be fixed.

murggel:
## Description

Use correct sortContext for search with target-coords:
The created sortContext (with target-coords) in CacheListAdapter.onCreate was overwritten by CacheListAdapter.initAdapter, hence the target-coords got lost.
No need to create the sortContext in initAdapter.

## Related issues

fix #15618

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -1146,8 +1146,6 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
     }
 
     private void initAdapter() {
-
-        sortContext = GeocacheSortContext.getFor(type, "" + listId);
         final ListView listView = getListView();
         registerForContextMenu(listView);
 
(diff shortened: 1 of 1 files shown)


murggel:
Probably to discuss and I can implement that in this PR as well or create a separate PR:

What distance-value does the user expects? The one to the searched location or to the current location?
I would prefer the distance to the target coords, because that is the interesting value for me, not the distance from my current location and it would make the sorting distance "visible".

eddiemuc (approved):

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

PR #15787, opened 2024-06-14
Move cache store/refresh from popup into background

moving-bits:
- download caches in background when triggering a store / refresh from cache popup
- disable the refresh / store buttons when starting the download
- update popup contents on having successfully received the cache data


moving-bits:
retest this please

ztNFny:
This breaks the list selection dialog from popup for already stored caches.
To repoduce:
1. Store a cache
2. Open Popup for the cache
3. Open list selection dialog
-> No lists selected

PR #15789, opened 2024-06-15
fix #15770: add google map options and hybrid map

### LINKED ISSUE #15770: Handling of Google Maps Styling Options
### Tell us your idea!

While working on a ticket and creating PR #15768 I realized that Google Maps offers a wide range of very detailled styling options, including visibility and coloring for all sorts of elements on the map. Details can be found eg via Google's Map Style editor at https://mapstyle.withgoogle.com/:

image
image

There are also some predefined themes available, which howeber seems to be just preconfigured using the same advanced option set.

As of today c:geo offers for GM a bunch of "Themes" which are technically just complete configuration sets out of the vast options available. Those are Default, NIght, Retro and High Contrast. Probably those were copied at the time from the then-available preconfigured themes? PR #15768 adds a new theme "Details" to this set.

It might be worth discussing whether we want to transport more detailled GM config options into c:geo, as we already do for OSM maps in Theme options.

One suggestion would be to allow "theme options" for GM as we do for OSM to turn features of the map "on" or "off". OPtions would be e.g. "Attractions", "Business" etc. In contrast to OSM we would have to offer a three-way-select (visibility "on", "Off" or "inherit from theme"). Also we would have to respect the hierarchy of GM (e.g. an option "Point of interest" sets visibility for all POIs while "Business" may override this with another visibility setting for Business-POIs)

We could also allow to set different colors, fills and such for different elements, but this would go too far IMHO. Up to discussion of course

### Is there an alternative?

Not to do anything

### Additional Information

_No response_

eddiemuc:
rel to #15770: add google map options and hybrid map

* corrects a bug: Google Map option "Satellite" does now use "Hybrid" again (like in old map)
* adds Google Map options to show "traffic" as well as "(3d) buildings", both defaulting to false

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -562,11 +562,16 @@ public class Settings {
     }
 
     public static void putBoolean(final int prefKeyId, final boolean value) {
+        putBooleanDirect(getKey(prefKeyId), value);
+    }
+
+    public static void putBooleanDirect(final String key, final boolean value) {
+
         if (sharedPrefs == null) {
             return;
         }
         final SharedPreferences.Editor edit = sharedPrefs.edit();
-        edit.putBoolean(getKey(prefKeyId), value);
+        edit.putBoolean(key, value);
         edit.apply();
     }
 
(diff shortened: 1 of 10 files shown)


eddiemuc:
3d building example
image

traffic example:
image

hybrid example:
image

new map options:
image

eddiemuc:
Update: enhanced the map theme options to show individual groups of features as discussed in #15770 :
image

The feature groups are those used by Google itself

eddiemuc:
Readjusted
image

PR #15794, opened 2024-06-16
Search for coordinates: Show direction and distance to target and not to current position

murggel:
## Description

Sorts the result of the search for target-coordinates according to the target-coordinates and not according to the current coordinates.

## Related issues

rel #15618

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/CacheListAdapter.java
+++ b/main/src/main/java/cgeo/geocaching/ui/CacheListAdapter.java
@@ -436,7 +436,12 @@ public class CacheListAdapter extends ArrayAdapter<Geocache> implements SectionI
 
         distances.add(holder.binding.distance);
         holder.binding.distance.setCacheData(cache.getCoords(), cache.getDistance());
-        holder.binding.distance.update(coords);
+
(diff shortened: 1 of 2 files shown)


murggel:
Feel free to discuss:
What distance-value does the user expects? The one to the searched location or to the current location?
I would prefer the distance to the target coords, because that is the interesting value for me, not the distance from my current location and it would make the sorting distance "visible".

murggel:
I tried it with italic - but the difference is not really visible

fm-sys:
I think the main problem is that the tiny compass is showing the direction from the current position (at least it looks like that) while the distance is based on the target coords. Maybe hiding the compass in these cases?

murggel:
No - the compass shows also the direction from the target coords to the specific cache.

I will provide two screenshots, one with current implementation and one with the new one...

murggel:
|current implementation|new implementation|
|---|---|
|||

murggel:
Just found out, that mapping the list, the target coordinates are not shown as cross...
But this is a different topic...

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
retest this please

PR #15795, opened 2024-06-16
add "corrected coordinates" filter

ztNFny:
Add a "corrected coordinates" filter in addition to the existing "Solved Mystery" filter which only ever applies to Type=Mystery or the "Has user defined waypoints" filter which only applies if there's a custom waypoint as well - which doesn't exist if the coords are coorected on gc.com or the custom waypoint is deleted.
Without this new filter it's impossible to filter a list to show only e.g. "Type=Wherigo AND Corrected Coordinates"


ztNFny:
Retest this please

Issue #15799, opened 2024-06-17
Display more prominent if location permission is missing 


TrollBlox:
### Tell us your idea!

My phone automatically removed location permissions when I took a break from geocaching this winter. When I opened the app again, oblivious to the fact it was unable to access my location it was acting strange putting my location nowhere near where I was. The app didn't seem to realize it wasn't able to access my location.

### What solution would you suggest?

The app should check if location permission is enabled, and request it again, giving the option to decline as well.

### Is there an alternative?

I had to manually give location permission in the settings after I realized the issue. I would consider myself a techy person, so someone with less experience could have issues.

eddiemuc:
When you don't have location permission you should see an orange warning on the home screen in the bottom (where also your current location is shown)
 Tapping on this should bring you to a dialog to grant location permission again.

Is this not working on your installation? If yes, this would be a bug.

TrollBlox:
you're right; I manually removed the permission and I now see the warning. it's very hard to see though.

eddiemuc:
> it's very hard to see though.

That's fair, could be more prominent. My suggestion:
- on home screen put a warning message in a box at the top. Tappable to assign permission, long-tap to dismiss (same behavior as eg reminder to check for updates map/route data)
- on the map display position arrow in gray color (same as eg Google Maps when location is unavailable)

I will reopen as feature request if that's ok?

TrollBlox:
I like these ideas!

SammysHP:
IIRC we decided against a more prominent warning because there are people using c:geo without granting the location permission. There were some issues about this topic in the past.

eddiemuc:
> IIRC we decided against a more prominent warning because there are people using c:geo without granting the location permission. There were some issues about this topic in the past.

That is true, but I think the suggestions are not too prominent (a message that can be removed by long-tap and a different color of the position arrow)

moving-bits:
A more prominent warning about missing location permission has been implemented with #15804, so let's close here.

PR #15804, opened 2024-06-22
show notification for missing location permission (rel. to #15799)

moving-bits:
Adds a short notification on main screen if location permission has not been granted currently. (This is similar to "check for map updates" notification, but without "blue dot" badge.)
Short tap on that notification opens the permission request, long-tap dismisses the notification, as suggested in https://github.com/cgeo/cgeo/issues/15799#issuecomment-2174993229,


moving-bits:
retest this please

moving-bits:
retest this please

PR #15808, opened 2024-06-23
fix #15748: allow favorite point change on log editing

### LINKED ISSUE #15748: Editing cache logs does not take care of existing favorite points
### Describe your problem!

When editing existing found logs, where a favorite point was assigned, this fact is not reflected in the edit view.

Found by a user reported on support mail and reproducible for me.

### How to reproduce?

- Open a found cache which you awarded a favorite point
- Tap on your found log and select to edit your log

The edit view does not contain the checkbox "Add to favorite" at all.
When sending the edited log the favorite point gets lost (meaning: No longer awarded to the cache)

The checkbox "Add to favorites" should be shown in the edit view and reflect the current state of whether or not the user assigned a favorite point to this cache.
Changes to the checkbox state should be send with the edited log (keep, add or remove favorite point depending on previous state).

### Reproducible

Yes

### c:geo Version

2024.05.28-NB

### System information

_No response_

### Additional Information

Ticket 340758

eddiemuc:
fix #15748: allow favorite point change on log editing

I had to change the wording of the favorite point checkbox in the log dialog so it matches both the "create log" and the "edit log" situation:

Before: Add this cache to favorites (%d remaining)
After: Make this cache a favorite (%d remaining)


eddiemuc:
@Lineflyer I used GC3A0E7 for some test logs, hope you don't mind. They are all deleted again

PR #15810, opened 2024-06-23
fix #15696: introduce search limit for coord/address searches, same as for nearby searches

### LINKED ISSUE #15696: Limit search radius for address search / coordinate search against gc.com (to allow for meaningful sorting by other criterias e.g. favorite points)
Ticket #520453
c:geo version 2024.04.25

### Current situation

When performing an address search or a coordinate search against gc.com, c:geo by default delivers the nearest 200 finds to the user. However, this is only as long as the user sorts the results by distance (the default setting).

If user changes sort to e.g. sort by favorite points, she will NOT get the same results as before sorted by favorite points but a different set of caches (sorted by favorite points). This is because c:geo sends hard-coded to search within a 16km radius of the given coordinate. So user will get 200 results within that radius around the coordinate with highest favorite points

(Note to the @cgeo/team : I finally found out what the search parameter `rad=16000` means - it is the search raduis of course :-)) 

### Wanted behaviour

It would be good if the user could adjust the search radius according to her own needs. The same is possible on the gc.com website. I suggest to add radius fields to the "coordinate" and "address" search in our "online search" activity for this, prefilled with the value "16".

### Alternative

User could add an advanced "distance" filter on top of her search and restrict distance with this. However, this only works for coordinates and is quite cumbersome.

eddiemuc:
fix #15696: introduce search limit for coord/address searches, same as for nearby searches

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
+++ b/main/src/main/java/cgeo/geocaching/loaders/CoordsGeocacheListLoader.java
@@ -28,12 +28,10 @@ public class CoordsGeocacheListLoader extends LiveFilterGeocacheListLoader {
 
     @Override
     public IGeocacheFilter getAdditionalFilterParameter() {
-        final DistanceGeocacheFilter distanceFilter = (DistanceGeocacheFilter) GeocacheFilterType.DISTANCE.create();
-        if (applyNearbySearchLimit) {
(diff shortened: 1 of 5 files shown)


fm-sys:
Don't you think using the same limit for both would be enough?

eddiemuc:
> Don't you think using the same limit for both would be enough? 

Then it would have been that way before. But it wasn't :-)

Issue #14601, opened 2023-09-12
show notification when GC live map is throttled (HTTP 429)


capoaira:
Issue from geoclub: https://www.geoclub.de/forum/t/caches-werden-nicht-geladen.82695/

When you move the map too often (about 10 times per minute) the caches doesn't load. After some time the caches load again. 
Here is a video of the behaviour: https://1drv.ms/v/s!Am4rTBGtAHyblp0PUzWb0lRh6Gqdkg 
(After 30 s and after 70 s in the video)

This is because of restrictions by GS. You can see the same behaviour on the search map on the website. 

It could be a solution to inform the user with a message about this restriction.

capoaira:
We jad the same problem with the GClh, maybe our solution can help:
https://github.com/2Abendsegler/GClh/pull/2124

ztNFny:
GC has been throwing plenty of HTTP 429 errors for me lately - I think they're actively trying to ruin their site :-(
We should probably show a toast in case of HTTP 429, something along the lines of "Server responded with "Too many requests", please try again in a moment"

eddiemuc:
Question: does this happen only when map is moved too quickly/frequently to new areas (resulting in attempts to download new caches) or does it also happen when moving the map e..g. forth and back between same two areas? Reason I am asking: the later should imho **not** result in frequent server calls because c:geo should temporarily cache already downloaded caches.

capoaira:
This happens only if requests are necessary. When I move the map for and back, the caches stay on the map. But when I do that inaccurate, it load caches (orange loading indicator at the top) and it happened again.

Lineflyer:
I think the suggestion of @ztNFny to handle the 429 is the thing we should do. That would also give a reasonable feedback to users but not modify timestamps or alike.

PR #15843, opened 2024-07-06
add logging and toast for HTTP 429 (Too many request)

ztNFny:
various requests may produce a HTTP 429 Too many requests, e.g. live map on GC.com (https://github.com/cgeo/cgeo/issues/14601)

This PR adds logging for that case and also a Toast to inform the user about the throttling.
image

Not sure where in the code to put this kind of error handling - it's a universal thing (that's where I put it for now) but maybe it should be handled case-specific? So setting feedback required for now.

Also hardcoded strings for now ;-)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/network/HttpRequest.java
+++ b/main/src/main/java/cgeo/geocaching/network/HttpRequest.java
@@ -1,5 +1,7 @@
 package cgeo.geocaching.network;
 
+import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.utils.JsonUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.RxOkHttpUtils;
@@ -102,6 +104,10 @@ public class HttpRequest {
     public <T> Single<T> requestJson(final Class<T> clazz) {
         headers("Accept", "application/json, text/javascript, */*; q=0.01");
         return requestInternal(r -> {
+            if (r.getStatusCode() == 429) {
+                Log.w("Request throttled: " + this.getRequestUrl());
+                ActivityMixin.showToast(CgeoApplication.getInstance(), "Please slow down, too many requests to " + HttpUrl.parse(getRequestUrl()).host());
+            }
             T result;
             final String bodyString = r.getBodyString();
             r.close();
--- a/main/src/main/java/cgeo/geocaching/network/HttpRequest.java
+++ b/main/src/main/java/cgeo/geocaching/network/HttpRequest.java
@@ -1,6 +1,5 @@
 package cgeo.geocaching.network;
 
-import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.utils.JsonUtils;
 import cgeo.geocaching.utils.Log;
@@ -104,10 +103,6 @@ public class HttpRequest {
     public <T> Single<T> requestJson(final Class<T> clazz) {
         headers("Accept", "application/json, text/javascript, */*; q=0.01");
         return requestInternal(r -> {
-            if (r.getStatusCode() == 429) {
-                Log.w("Request throttled: " + this.getRequestUrl());
-                ActivityMixin.showToast(CgeoApplication.getInstance(), "Please slow down, too many requests to " + HttpUrl.parse(getRequestUrl()).host());
-            }
             T result;
             final String bodyString = r.getBodyString();
             r.close();
@@ -150,6 +145,10 @@ public class HttpRequest {
             final T mappedResponse = mapper.apply(response);
             if (mappedResponse instanceof HttpResponse && response != mappedResponse) {
                 ((HttpResponse) mappedResponse).setFromHttpResponse(response);
+            }
+            if (response.getStatusCode() == 429) {
+                Log.w("Request throttled: " + this.getRequestUrl());
+                ActivityMixin.showApplicationToast("Please slow down, too many requests to " + HttpUrl.parse(getRequestUrl()).host() + "/" + HttpUrl.parse(getRequestUrl()).encodedPath());
             }
             return mappedResponse;
         });


ztNFny:
retest this please

main/src/main/java/cgeo/geocaching/network/HttpRequest.java
```diff
@@ -102,6 +104,10 @@ public Single<HttpResponse> request() {
     public <T> Single<T> requestJson(final Class<T> clazz) {
         headers("Accept", "application/json, text/javascript, */*; q=0.01");
         return requestInternal(r -> {
+            if (r.getStatusCode() == 429) {
```
eddiemuc, line 107: I suggest to move this code part to line 150 (into method `requestInternal`) instead of here. Then it will be used in all cases

main/src/main/java/cgeo/geocaching/network/HttpRequest.java
```diff
@@ -102,6 +104,10 @@ public Single<HttpResponse> request() {
...
+                Log.w("Request throttled: " + this.getRequestUrl());
+                ActivityMixin.showToast(CgeoApplication.getInstance(), "Please slow down, too many requests to " + HttpUrl.parse(getRequestUrl()).host());
```
eddiemuc, line 109: I believe there's a method "showapplicationtoast" or something like this.

ztNFny:
I suggest to merge this as is to be give visibility to those issues and enable working on solutions.

ztNFny:
restest this please

MagpieFourtyTwo:
> I used this build not for about 2h and got like ~100 of those toasts already, doing normal activities. Seems like we have a lot of places where we hit some limits :-o

Is this perhaps because of IIRC @eddiemuc's increase of the number of caches fetched in one server call for live map?
AFAIR it was bumped from x00 to 1000 or so, to get rid of the problem with caches loaded being out of center of the current view port ...?

eddiemuc:
> Is this perhaps because of IIRC @eddiemuc's increase of the number of caches fetched in one server call for live map?\nAFAIR it was bumped from x00 to 1000 or so....

I recently increased the size of the c:geo internal cache for displayed caches on the maps viewport. This should not influence the number of server requests though. (Well never say never in software development of course :-))

We can try to decrease the number of caches returned per call. I don't know if GC.com triggers HTTP 429 solely based on number of requests or also on size of response though.

ztNFny:
Let's wait for this to be merged and provide some data before we take blind actions 😉

eddiemuc:
One additional insight: I did a short "test" just now: I copied one typical cache-retrieving Url in my browser's and hit the refresh button every 4-5 seconds or so (mainly because gc.com takes a real long time to respond...). Even in this "manual" setup I hit a HTTP 429 after 10-15 calls. Next call returned a result with HTTP 200 again though.

So it seems we deal with rather "short-lived" but frequent periods of HTTP 429's.

MagpieFourtyTwo:
> (mainly because gc.com takes a real long time to respond...)

I observed the same performance problems yesterday evening.  Additionally I got a lot of "Server unreachable" and "SSL errors", too.

A traceroute showed that there seemed (or even seems, have not checked again today) to be a problem in the main EU<->US provider's infrastructure (packages were sent over a really weird route and add. even took a nap for a while).

So, is it perhaps possible that there are time-outs happening in c:geo, resulting in more retries than usual?

ztNFny:
There's massive connection problems in the last 24h.
But what i was seeing was well before that. Anyway, no point in speculation, let's merge this and then we'll see where and when those issues arise 👍

SammysHP:
It's a bit late now, but shouldn't the toast be shown only when debug mode is enabled? Could be annoying (even for nightly users) when it happens frequently.

Lineflyer:
Feedback from RC-user:

Just tried the newest beta 2024.07.14 and found a pretty annoying popup on my way: "slow down. too many requests to geocaching.com". That popup stays in screen (for 30 seconds or so?) even when I terminate c:geo and go to any other app. 

I can not reproduce, but happened when I had Réunion (Fr) as a whole on screen.

ztNFny:
Even a long toast should be only 3.5s, nowhere near 30s.

ztNFny:
>  newest beta 2024.07.14 

This shouldn't be in release branch but only in master? We need to fix the reason for those "too many request" responses and this PR is only the enabler for that by making them visible in the first place.

fm-sys:
> Even a long toast should be only 3.5s, nowhere near 30s.

Multiple toasts can get queued and displayed one after each other. 

> It's a bit late now, but shouldn't the toast be shown only when debug mode is enabled?  

I agree that those toasts should only be shown if debug mode is turned on. At least it's definitely nothing for a beta.

eddiemuc:
> >  newest beta 2024.07.14 
> 
> This shouldn't be in release branch but only in master? 

I assume that master was merged to release as we do for new releases (in this case a beta). There should never be something in master which is not allowed to go to end users (beta or production) as well. Experiments can be shielded by eg putting them in debug mode only, setting up a secret property for enablement or restricting them to certain environments (eg dev only)

eddiemuc:
I created #15882 for the issue in beta so I can reference users from support to it

MagpieFourtyTwo:
Got the message too, of course, multiple times, but most astonishingly even immediately after starting c:geo in the morning, on first manual map movement at zoom level of 500 meters with just 10 caches currently displayed.  Moreover with a zoom level of 2 km (i. e. a radius of about 5 km) there are just about 260 caches, thus with the very first server request there should have been loaded *all* caches of the whole region.  So why did I get the toast?

BTW: If I just start c:geo without any manual interaction (apart from activating the ``Stay with me`` button), and just driving along, I did not get a single toast at all.

Seems to me as if every single manual(!) map movement creates a new server request, no matter if a) the area displayed on map has already been covered by a previous server request (i. e. how far the manual movement went) nor b) which zoom level is currently chosen - in fact, how likely it is that a new request may perhaps bring new caches.

While writing this I remembered a strategy implemented in lots of other map apps:  If there is a manual scroll, just let the user decide if he wants a new server request, i. e. give a clickable message which reads like "Search in this area".  Which would most probably already "eliminate" most of the currently appearing toasts ...?

MagpieFourtyTwo:
Moreover a "manual request" would "solve" a kind of annoying (but nevertheless of course home brewn) "problem" I often have, when I just want to scroll to a different region, which is several manual map movements away, to look at caches in this distant area - while the caches "on the movement way" are not only of no interest anyway, but even hinder my "unobstructed map view" (cause they clutter the cities whose names I in fact need to first find the target location).  Additionally there are of course a lot of server requests which not only superfluous but even disturbing.

And, yes. I know, the "right" way would be to just deactivate Live, then scroll to target and activate Live again, but, TBH, who really does so?  Esp. if you're a ... non-experienced user.  \*shrug*

SammysHP:
> TBH, who really does so?

\*raises hand\*

But even the geocaching.com map shows a "search here" button after a while (which I find annoying).

MagpieFourtyTwo:
> *raises hand*

Gummipunkt++  ;)

And even geocaching.com's map shows "slow down" messages, too, for a few days now, which I have never ever seen before.

MagpieFourtyTwo:
Ah, BTW:  Have played a bit more to force "slow down" messages in c:geo,
and it seems that you get way more (and very reproducible) messages, if you do ``pinch2zoom & move`` in one go.
While it doesn't matter, how much you zoom or how far you move, even the smallest movement of just a pixel *plus* a minimal zoom gesture gives at least after the third action (within a few seconds) to to a toast.  Sometimes even three toasts for three such actions.  It's just that it takes a second or two until the toasts appear.

And it appears to be way "harder" to get a slow down toast, if you just move around a bit (with e. g. 500 m zoom with not that much speed resp. frequency).

MagpieFourtyTwo:
Ah, forgotten: In the viewport where I tested the above scenario, there were only two caches on display.
And the surrounding area has not that much caches, neither. 
Which in fact seems to falsify my guess in [comment 2218740095](https://github.com/cgeo/cgeo/pull/15843#issuecomment-2218740095):

> Is this perhaps because of IIRC @eddiemuc's increase of the number of caches fetched in one server call

... i. e. the slow down message appears to be independent from the sheer number of caches fetched,
but for real on the server request count.

Fun fact:  Just scrolled around a bit with 20 km zoom (i. e. maximum number of fetched caches possible, I would say) and did not even get a single toast at all.  And even with zooming in and out while moving it took about 10 actions until the first toast appeared ... am I the only one who does not understand a bit anymore ...?!? 🤪

ztNFny:
Fyi, you can check logcat for the exact failing requests, might give you a better idea to the cause.

PR #15868, opened 2024-07-13
Add debug function: Database inspection

moving-bits:
## Description
Both while working on some dev issue as well as while working on some support ticket I have encountered the need to read some specific database content in the past. While there is some (cumbersome and read-only) database inspection integrated in Android Studio, there is no way for the non-technical savy user to retrieve such info.

This PR adds a light-weight database inspection view to settings => system => debug section.

- table selector
- display all colums of selected table
- pagination (10 / 100)
- edit fields (non PK-fields)
- search all colums
- sort by column ASC/DESC

## Additional context

|Table selection|View content|
|---|---|
|Screenshot_1720866296|Screenshot_1720866279|


moving-bits:
What do those errors mean?

- Could not load custom lint check jar file /home/jenkins/.gradle/caches/transforms-3/0c517e535c0404d8810aa3a3e9fa6c1e/transformed/activity-1.8.0/jars/lint.jar
- java.lang.UnsupportedClassVersionError: androidx/activity/lint/ActivityIssueRegistry has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0

(and then some lint messages are listed, which are regarded as warnings only in our configuration, which apparently could not been loaded for some reason)

Is this some "Java 17" issue?

bekuno:
> What do those errors mean?
> 
> ...
> 
> Is this some "Java 17" issue?

I think yes.
I assume that your library projects are build with Java 17. Correct?

moving-bits:
> I think yes. I assume that your library projects are build with Java 17. Correct?

The first build was "openjdk17" indeed, but after I had seen the error message, I've updated `jitpack.yml` to use "openjdk11", which should be compatible to our current CI. (I also updated this PR to use the new commit hash.)

eddiemuc:
> > I think yes. I assume that your library projects are build with Java 17. Correct?
> 
> The first build was "openjdk17" indeed, but after I had seen the error message, I've updated `jitpack.yml` to use "openjdk11", which should be compatible to our current CI. (I also updated this PR to use the new commit hash.)

The class reported to be compiled with Java 17 is androidx/activity/lint/ActivityIssueRegistry. Surely not a class you created yourself I assume. More likely a dependency you included.

BTW: my impression is that this database viewer is far too oversized to be included in c:geo just for debug purposes

bekuno:
retest this please

moving-bits:
> The class reported to be compiled with Java 17 is androidx/activity/lint/ActivityIssueRegistry. Surely not a class you created yourself I assume. More likely a dependency you included.

for `dbinspection` lib I started a new project in Android Studio with the dependencies it suggested. Among others, that's material 1.12.0 and appcompat 1.6.1 (originally this was 1.7.0, but that gave me some grieves during my experiments with Gradle & Jitpack). For testing purposes I could downgrade those libs to use the identical versions c:geo uses currently (and remove some of the unused default libs). => will give it a try

> BTW: my impression is that this database viewer is far too oversized to be included in c:geo just for debug purposes

The resulting `.aar` file is a mere 14 kB in my local build.

bekuno:
Looks good now in build https://ci.cgeo.org/job/cgeo-CI_PR-build/8951/.
Please use no newer dependencies in your library than c:geo.

In the builds before, it claims some lint errors:
~~~
23:34:45                    Lint found 4 errors, 40 warnings. First failure:
23:34:45    
23:34:45    src/main/java/cgeo/geocaching/ui/CheckableItemSelectActivity.java:137: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
23:34:45        public void onBackPressed() {
23:34:45                    ~~~~~~~~~~~~~
23:34:45    
23:34:45    The full lint text report is located at:
23:34:45      /home/jenkins/slave/workspace/cgeo-CI_PR-build/main/build/intermediates/lint_intermediate_text_report/basicDebug/lint-results-basicDebug.txt
~~~
I checked this file directly. It reports:
~~~
src/main/java/cgeo/geocaching/ui/CheckableItemSelectActivity.java:137: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
    public void onBackPressed() {
                ~~~~~~~~~~~~~
src/main/java/cgeo/geocaching/EditWaypointActivity.java:313: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
    public void onBackPressed() {
                ~~~~~~~~~~~~~
src/main/java/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java:376: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
    public void onBackPressed() {
                ~~~~~~~~~~~~~
src/main/java/cgeo/geocaching/maps/routing/RouteSortActivity.java:216: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
    public void onBackPressed() {
                ~~~~~~~~~~~~~
...
~~~
But this errors are gone in the last build.
The problems exist. So, we have probably the CI java error as a side effect of a new lint check.

Btw., your PR adds a new lint wearning: "Very long vector path" for [ic_database.xml](https://ci.cgeo.org/job/cgeo-CI_PR-build/8951/androidLintResult/new/source.-7461286956291889425/#10).

main/build.gradle
```diff
@@ -337,6 +337,10 @@ dependencies {
     // GeographicLib
     implementation 'net.sf.geographiclib:GeographicLib-Java:2.0'
 
+    // DB inspection
+    implementation 'com.github.moving-bits:datatables:-SNAPSHOT'
+    implementation 'com.github.moving-bits:dbinspection:b6c15506ee'
```
bekuno, line 342: General question to @cgeo/team.
Is it really necessary to have separate external projects here? Wouldn't it be better to have them in c:geo (directly or as dedicated repos)?

eddiemuc: If we want to go this path I vote for direct code integration . Except if database inspector is maintained and published as it's own library to a public place like maven central, then of course we can just link it.

moving-bits: That's what this library is intended to be (as it's use case goes beyond c:geo). It's a public repo and the lib is published to Jitpack already.

moving-bits:
> I checked this file directly. It reports:
> 
> ```
> src/main/java/cgeo/geocaching/ui/CheckableItemSelectActivity.java:137: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
>     public void onBackPressed() {
>                 ~~~~~~~~~~~~~
> src/main/java/cgeo/geocaching/EditWaypointActivity.java:313: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
>     public void onBackPressed() {
>                 ~~~~~~~~~~~~~
> src/main/java/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java:376: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
>     public void onBackPressed() {
>                 ~~~~~~~~~~~~~
> src/main/java/cgeo/geocaching/maps/routing/RouteSortActivity.java:216: Error: Overriding method should call super.onBackPressed [MissingSuperCall]
>     public void onBackPressed() {
>                 ~~~~~~~~~~~~~
> ...

I guess those are messages we have configured as "warnings" in c:geo, but default to "error".

> Btw., your PR adds a new lint warning: "Very long vector path" for ic_database.xml

That icon has been taken from fonts.google.com => icons => database :shrug:, I've only adjusted the color + added a source notice.

fm-sys:
+1 for keeping it separated. c:geos codebase is already too large. All things which are not explicitly related to geocaching and thus can be extracted easily should not clutter cgeos code base IMHO.

bekuno:
> > Btw., your PR adds a new lint warning: "Very long vector path" for ic_database.xml
> 
> That icon has been taken from fonts.google.com => icons => database 🤷, I've only adjusted the color + added a source notice.

Sure. We had this also on other vector drawables.
Please annotate it with `tools:ignore="VectorPath"`. (as we have in other files)

moving-bits:
> Please annotate it with `tools:ignore="VectorPath"`. (as we have in other files)

Fair enough. I have updated the PR accordingly.

PR #15870, opened 2024-07-13
UnifiedMap: Highlight selected cache/waypoint (fix #15801)

### LINKED ISSUE #15801: Indicate selected cache on map
### Tell us your idea!

When a cache is selected in the map view, it should indicate which one is selected (when the bottom panel is open). It can get confusing if a lot of caches are loaded on the map. The official Geocaching app has this, where the circle icon switches to a pin.

It could grow larger, change the color, or get an outline. Basically anything would work.

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
This is a technical demo for highlighting the active cache on the map (so far UnifiedMap only).
As soon as you tap on a cache (which opens the cache info sheet), a "corona" is being displayed around the cache icon. (Same could be done for waypoint icons and for compact icon variants.)

From looking at `marker.xml` and `marker_pin.xml` as well as `MapMarkerUtils.createCacheMarker` I would have expected a different result, though. Both drawables have 36dp width and height, and their viewports are 36/36 as well, so I would have expected, that `marker_pin.xml` is sort of an overlay to `marker.xml` (which would fit at being treated as inset).

Thus: aligning another drawable at "bottom center" should create an exact overlay, if that drawable has the same size.

But it does not, it aligns to the `marker.xml`'s visual bottom border, above the pins visible top.

Having said that, the result would still be working as a highlighting indicator for a selected cache, though:

|GC.com cache highlighted|UDC cache hightlighted|
|---|---|
|Screenshot_1720901625|Screenshot_1720903345|

(Although one would probably want to adjust the highlighter's size and choose a different color, as the current one is too close to the UDF's background color.)

Marking this PR as draft.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/RouteItem.java
+++ b/main/src/main/java/cgeo/geocaching/models/RouteItem.java
@@ -166,6 +166,14 @@ public class RouteItem implements Parcelable {
         return DataStore.loadCache(cacheGeocode, LoadFlags.LOAD_CACHE_OR_DB);
     }
 
+    @Nullable
+    public Waypoint getWaypoint() {
+        if (waypointId != 0) {
+            return null;
+        }
+        return DataStore.loadWaypoint(waypointId);
+    }
+
     @NonNull
     public String getShortGeocode() {
         return generateShortGeocode(cacheGeocode);
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LayerHelper.java
@@ -5,12 +5,14 @@ public class LayerHelper {
     // constants for layer ordering (higher = more visible)
     // layer index numbers must be consecutive, starting with 1 for ZINDEX_BASEMAP
 
+    public static final int ZINDEX_CACHE_WAYPOINT_HIGHLIGHTER_GEOITEM = 101;
+    public static final int ZINDEX_CACHE_WAYPOINT_HIGHLIGHTER_MARKER = 100;
+
     public static final int ZINDEX_ELEVATIONCHARTMARKERPOSITION = 16;
     public static final int ZINDEX_POSITION = 15;
     public static final int ZINDEX_POSITION_ELEVATION = 14;
     public static final int ZINDEX_SEARCHCENTER = 13;
-    // some OSM-only layers
-    public static final int ZINDEX_SCALEBAR = 12;
+    public static final int ZINDEX_SCALEBAR = 12;   // OSM only
     public static final int ZINDEX_GEOCACHE = 11;
     public static final int ZINDEX_WAYPOINT = 10;
     public static final int ZINDEX_COORD_POINT = 9;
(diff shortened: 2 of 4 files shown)


ztNFny:
I'll try to take a look at positioning and design next weekend..design-wise i'm nit convinced...

moving-bits:
> I'll try to take a look at positioning and design next weekend.

Thanks

> design-wise i'm not convinced...

Agreed. Maybe we find a solution to the positioning.

moving-bits:
While crunching a bit on it a slightly different idea came to my mind:
As this is solely about highlighting a single cache (or waypoint): Why not just add a colored circle on top of it?

|just colored circle|accented background|
|---|---|
|Screenshot_1721572171|Screenshot_1721572154|

Here I used our accent color. This may be a bit too unobstrusive, depending on cache type (multi? user-defined cache?), so one might want to choose a more distinctive color instead.

ztNFny:
The issue was with your SVG, its size was wrong, here's a quick fix for that:
```
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="36dp"
    android:height="36dp"
    android:viewportWidth="36"
    android:viewportHeight="36">
    <path
        android:pathData="M2,18a16,16 0,1 0,32 0a16,16 0,1 0,-32 0z"
        android:strokeWidth=".75"
        android:fillColor="@color/colorAccent"
        android:strokeColor="@color/colorAccent"/>
</vector>

```

image

However I'm still no fan of that. For one imho this aura effect looks strange. And then in this screenshot a specific problem (actually only half-related) can be seen: the selected cache is half-hidden behind others. If we choose to add some highlighting i think we should also try to bring the cache forward.

ztNFny:
> Why not just add a colored circle on top of it?

"just colored circle" looks good to me. Buuut: The line will go trough all the elements which are part of the cache marker, e.g. disk, D/T, list icons, ...

ztNFny:
played a bit more with different background circle sizes and while it's better than nothing it's far from good imho, best is

```
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="36dp"
    android:height="36dp"
    android:viewportWidth="36"
    android:viewportHeight="36">
    <path
        android:pathData="M0,18a18,18 0,1 0,36 0a18,18 0,1 0,-36 0z"
        android:fillColor="@color/colorAccent"/>
</vector>
```

Enlarging the whole marker (125%) would be an option, but I guess also that one would require redrawing the marker itself.

moving-bits:
Thanks for deep-diving into this. IMHO the last screenshot is fine - it's actually just meant to be a visual indicator which cache is currently selected, so every highlighting is better than before. I'm sure there will be many edge cases (like overlapping cache markers etc.), but I'm not sure if we will be able to avoid all of them with a reasonable invest.

I'll update my PR with the xml from https://github.com/cgeo/cgeo/pull/15870#issuecomment-2241718152

ztNFny:
Set a filter in a way that the cache is removed from map after found, log cache from popup - cache is removed but the circle isn't.

fm-sys:
> image


Can we show cache and colored circle in front of all other caches? (should be pretty easy by setting the z-layer, maybe also draw the same cache icon a second time on top so that we don't need to refresh the normal geocache layer?)

moving-bits:
> Set a filter in a way that the cache is removed from map after found, log cache from popup - cache is removed but the circle isn't.

hm, highlighting background gets removed on closing the info sheet currently. - Potentially removing it on applying a filter will need some rewrite. I'll try to look into it over the next days.

> Can we show cache and colored circle in front of all other caches? (should be pretty easy by setting the z-layer, maybe also draw the same cache icon a second time together with the circle so that we don't need to refresh the normal geocache layer?)

How can I retrieve the fully-built marker for a certain geocache (to be able to put it again on the map, but on a higher z-level)?

fm-sys:
This should do the trick:
```
MapMarkerUtils.getCacheMarker(activity.getResources(), cache, null, true)
```

See also https://github.com/cgeo/cgeo/blob/master/main/src/main/java/cgeo/geocaching/unifiedmap/layers/GeoItemsLayer.java

ztNFny:
Maybe just remove the highlight on any resume of the map? Would run more often then required but what's the harm in that ...

moving-bits:
Updated the PR to
- highlight also waypoints
- display selected item and its colored background on top of all other elements (highest z-level)

moving-bits:
retest this please

PR #15886, opened 2024-07-20
Show http 429 warning in debug mode only

moving-bits:
## Description
To get our beta user up and running again, switch the "error 429 / slow down" toast to debug mode for now.
If no one objects I'll start building a new beta version including this PR in two hours or so.

I'll try to come up with something less blocking in a separate PR.


moving-bits:
(I thought the DNS problem in the runner was fixed? :thinking: Let's try it again.)

retest this please

PR #15889, opened 2024-07-20
WIP show http 429 warning overlay

moving-bits:
## Description
Proposal for a new warning when http 429 errors occur. The text is being displayed as a text overlay for one second. If a new one occurs while one is being displayed already, the duration gets reset to one second (but not added on top).

Screenshot_1721507577

This is a WIP, having been implemented for UnifiedMap only so far. If feedback is "go" I will add it to GMv2 and old Mapsforge as well.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
@@ -5,15 +5,22 @@ import cgeo.geocaching.CachePopupFragment;
 import cgeo.geocaching.R;
 import cgeo.geocaching.SwipeToOpenFragment;
 import cgeo.geocaching.WaypointPopupFragment;
+import cgeo.geocaching.network.HttpRequest;
 import cgeo.geocaching.unifiedmap.UnifiedMapViewModel;
+import cgeo.geocaching.utils.LifecycleAwareBroadcastReceiver;
 import cgeo.geocaching.utils.functions.Action1;
 
 import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
 import android.content.res.Resources;
+import android.os.Handler;
+import android.os.Looper;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewTreeObserver;
 import android.widget.FrameLayout;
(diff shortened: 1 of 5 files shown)


eddiemuc:
I vote against this PR. 

429 is clearly reproducible. We don't need to invest any more time in displaying it "nicely" to the end users. We have everything on our hands to analyze the topic and find solutions within c:geo such that it triggers not too many requests.

eddiemuc:
I thought about this overnight and have an alternative suggestion now:

We have methods already in ViewUtils to show toasts. These methods handle eg i8n with parameters, logging and context handling (activity vs application).

We could add an optional parameter here to show a specific toast (text) AT MOST every x seconds. Then for our 427 toast we set this parameter to eg 10 seconds

moving-bits:
> 429 is clearly reproducible. We don't need to invest any more time in displaying it "nicely" to the end users. 

I'd like to contradict here. The original PR was about informing the user if there are any server-side restrictions that lead to live map being incomplete, which is a useful information. Just the way it was presented to the end user was too intrusive, due to too many notifications being displayed. This should be fixed, IMHO.

In which way it gets fixed (overlay, as shown above, or a toast with max frequency) I have no strong opinion about.

> We have everything on our hands to analyze the topic and find solutions within c:geo such that it triggers not too many requests.

+1 for analyzing and reducing / removing unnecessary calls to the server

ztNFny:
a few fixes to improve 429s have been implemented but of course they still occur, e.g. when panning the map quickly. We should think about how to make them visibile to the user to let them know there's a problem and live map isn't showing the "truth"

moving-bits:
I've rebased this PR to current `master` and shortened the error message to only show the host name, but no longer the path (which is irrelevant to the end user, IMHO). Logging still contains the full path.

image


I'll prefer an overlay window like in this PR to a toast, as a toast is limited to a single line only and will cut off the relevant parts of the information.

ztNFny:
It seems the default state for the icon is wrong. Opening the map it's there and it only disappears once i run into a 429

moving-bits:
> It seems the default state for the icon is wrong. Opening the map it's there and it only disappears once i run into a 429

Thanks, fixed with #16419

PR #15891, opened 2024-07-21
Enable hint rotation independantly from setting (fix #15887)

### LINKED ISSUE #15887: Hint cannot be rotated if "encode hint" setting is unset
User report on support: (ticket 592519)

Cache GC9G3NA has an unencrypted hint (cache owner seems to have done the ROT13 already).
Hint gets displayed encoded in c:geo, and user cannot tap on the hint to decode the hint.

---

I can reproduce the behavior if I disable the "encode hint" option in settings. Hint gets displayed encoded in this case, and I taps on the hint get ignored.

This brings up two questions:
- Why does c:geo encode the hint at all even with this option not being set?
- Any why does it offer rotation only if that option is enabled?

(Not sure whether this would be a bug report or a feature request - probably both.)

moving-bits:
## Description
Allow hint text rotation independantly from "show hint encoded" setting.


moving-bits:
retest this please

moving-bits:
retest this please

PR #15898, opened 2024-07-23
Use MSL altitude if available (fix #14424)

### LINKED ISSUE #14424: Display elevation above Mean Sea Level instead of above WGS84 Ellipsoid
### Describe your problem!

Right now c:geo displays the elevation of the current location within the Home-screen and Compass screen **_above WGS84 Ellipsoid (=Ellipsoidal Height)_**. (at least on my phone: Xiaomi Redmi 10 Pro, Android 13)
This is the raw elevation a GPS-receiver determines. Bus this elevation is not used in everydays life, like elevations of mountain peaks for example. Their elevations are so called **_"Elevation above Mean Sea Level"_** which both differ depending on the location on Earth. In Austria about 45 m. So if the peak of a mountain has 1000m above mean sea level, c:geo displays: 1045m by mistake.

Please correct the elevation value to be displayes as _"m above mean sea level"_ . There are correction files available in the Internet, google for "Geoid".

See: https://support.virtual-surveyor.com/en/support/solutions/articles/1000261349-the-difference-between-ellipsoidal-geoid-and-orthometric-elevations-

### Reproducible

Yes

### c:geo Version

2023-07-16

### System information

Xiaomi Redmi 10 Pro, Android 13

moving-bits:
## Description
On Android 14+ systems use the new feature to retrieve MSL-based altitude (see [docs](https://developer.android.com/reference/android/location/Location#getMslAltitudeMeters())), if such data is available for current location.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GooglePositionAndHistory.java
@@ -19,6 +19,7 @@ import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.unifiedmap.geoitemlayer.GeoItemTestLayer;
 import cgeo.geocaching.unifiedmap.geoitemlayer.GoogleV2GeoItemLayer;
 import cgeo.geocaching.utils.AngleUtils;
+import cgeo.geocaching.utils.GeoHeightUtils;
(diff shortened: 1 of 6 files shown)


moving-bits:
retest this please

moving-bits:
Retest this please

moving-bits:
retest this please

PR #15907, opened 2024-07-25
Adapt MapsforgeDownloader test to new map structure

moving-bits:
## Description
Mapsforge seems to have two more countries in Europe with state-specific sub maps, therefore need to adapt the Mapsforge downloader test


devemux86:
We now provide maps for regions of more large countries:
Spain, Netherlands, Brazil, China, India, Indonesia, Japan, etc.

devemux86:
We now provide maps and poi for regions of more large countries:
Spain, Netherlands, Brazil, China, India, Indonesia, Japan, etc.

PR #15915, opened 2024-07-27
Fix some accessibility issues

moving-bits:
## Description
On publishing a new release Play Store does some prelaunch tests. One major part (at least for us) is about accessibility issues. Sadly I found no way to export this report, so I start a PR here to fix them one-by-one. This is a first start, more to come.


bekuno:
Please have a look at the CI lint results.

The https://ci.cgeo.org/job/cgeo-CI_PR-build/9085/androidLintResult/new/ shows 2 new entries.

One of them is `The resource `R.string.google_maps_style_details` appears to be unused`.
The string is in your strings.xml, but not the line in GoogleMapsThemeHelper.java, if I see it correct.
(The souece was https://github.com/cgeo/cgeo/commit/589b25c3653b7d44f52ac2496cf75dbb5f067253)

moving-bits:
Removed the strings that are no longer in use

bekuno:
retest this please

ztNFny:
This PR needs to be partially reverted.
It added "android:hint" to many buttons - I guess what it wanted to do is add "android:tooltipText". "hint" is intended for text input fields as the "Hint text to display when the text is empty." On a button hint never makes any sense.

Those "hint"s are a visible element of the buttons so now most buttons look like this:
image


@moving-bits as I don't know what the Play Store Console reported could you please share why you added hint?

ztNFny:
https://github.com/cgeo/cgeo/pull/15944 converts those hints into tooltipText - assuming that was the intended change

PR #15919, opened 2024-07-27
Consider null in VariableList.toMap correctly (fix #15589)

### LINKED ISSUE #15589: Variables for projection get prefilled with NULL
Quick note from out and about:

Today I prepared a multi and added a calculated projection like E*5+1 meters and F+10 degrees.

Afterwards I synced the waypoints to my personal note.

When looking at the variable tab afterwards, that E and F (previously empty) now contain the value `null`

Current nightly

murggel:
## Description

in method VariableList.toMap consider null-value as empty string

## Related issues

fix #15589


moving-bits:
retest this please

bekuno:
retest this please

PR #15941, opened 2024-08-02
allow up to 200 waypoints to be calculated

ztNFny:
calculating coordinates supports placeholders (e,g, [:0-9]) in formulas to then draw multiple waypoints. This functionality is currently limited to 20 waypoints  - if the formula+variables would result in 21 or more waypoints the option to draw the waypoints is disabled.
In real life use i've often encountered the situation where this 20 limit is not enough, e.g. when I'm missing 2 out of 6 bonus numbers when it'll be 100 waypoints. Even in this situation just drawing all 100 of them will usually point to a reasonable spot on the map that enables finding the cache.

This PR raises the limit of 20 (arbitrary number I guess) to 200 (another arbitrary number). Could also be 1000 to support 3 vars missing ...

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/formulas/IntegerRange.java
+++ b/main/src/main/java/cgeo/geocaching/utils/formulas/IntegerRange.java
@@ -25,7 +25,7 @@ public class IntegerRange {
     }
 
     public static IntegerRange createFromConfig(final String config) {
-        final List<Integer> parsed = parseConfig(config, 20);
+        final List<Integer> parsed = parseConfig(config, 200);
         return parsed == null ? null : new IntegerRange(ArrayUtils.toPrimitive(parsed.toArray(new Integer[0])));
     }
 


eddiemuc:
The limit of values is per parenthesis. If You want to produce more values/coordinatea you can already do something like this:



This would produce 1000 waypoints.

I actually like it that the user has to type a bit more when she wants to create such many waypoints. Should not be too easy th cheat :-)

I limited the number because this is at the same the the max array size stored in-memory per parenthesis in the formula. If not limited, memory consumption will quickly go up a lot while typing a formula. I guess 200 would still work though

eddiemuc:
@ztNFny I have an idea how to make IntegerRange work for large amounts of values w/o taking too much memory. Then we can make ranges limitless.

 I try to create a PR for it this weekend. Would that be ok?

ztNFny:
@eddiemuc feel free to close

moving-bits:
superseded by #15946, so let's close here

PR #15943, opened 2024-08-02
improve map viewport comparison for live map

### LINKED ISSUE #15942: Reduce number of live map searches on GC (HTTP 429, too many requests)
Using Live Map currently produces a lot of requests, resulting in HTTP 429 (too many requests) errors on GC.com (and I assume for the other services the request count is just as high, just that they don't throw errors).
We need to find a way to reduce the number of requests sent by live map, so let's collect some ideas first:

- Manual request: Instead of searching on move show a "Search here" button that the user has to click (suggested in https://github.com/cgeo/cgeo/pull/15843#issuecomment-2236223801)
  - Pro: easy to implement, might also help with some usecases like scrollign to another location
  - Con: Big change in UX, breaks some workflows (e.g. live map set on your location while on the move)
- Minimum distance: Only search if the distance to old location is greater X
  - Pro: Most likely not even noticable by user
  - Con: Implementation requires comparing old and new viewport (and *ideally* not just viewport but visible caches to avoid reloading caches that were loaded 2 map moves ago) which could become quite complex.
- Scroll threshold: Similar to minimum distance, but more "simple". Addressing the fact that right now a 1px move is enough to trigger the update (https://github.com/cgeo/cgeo/pull/15843#issuecomment-2236815890)
  - Pro: No UX impact, easier implementation than proper "minimum distance"
  - Con: not as good ;-)

Any other ideas?

ztNFny:
For UnifiedMap use a similar logic as for old map to compare whether the current viewport requires a new request to server for LiveMap results. Previous implementation was triggered with *any* zoom change.

partially fixes #15942

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/LoadInBackgroundHandler.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/LoadInBackgroundHandler.java
@@ -65,8 +65,8 @@ class LoadInBackgroundHandler {
                 final int currentZoom = map.getCurrentZoom();
 
                 // check if map moved or zoomed
-                final boolean moved = previousViewport == null || currentZoom != previousZoom || mapMoved(previousViewport, currentViewport);
-                if (moved) {
+                final boolean useLastSearchResult = null != lastSearchResult && null != previousViewport && (previousViewport.includes(currentViewport) || !mapMoved(previousViewport, currentViewport));
+                if (!useLastSearchResult) {
                     load(currentViewport);
                     previousZoom = currentZoom;
                     previousViewport = currentViewport;


fm-sys:
I haven't tested it, but there may be some problems with that implementation (blind guesses):

- while zooming out `includes` will never return true, thus it may still trigger a lot of requests

- when being at zoom level "world" caches by far won't get loaded for the whole viewport. While zooming in, `includes` is always true so no requests will happen at all?

I would suggest to compare the viewport size instead and reload when the relative zoom difference exceeds some bounds

moving-bits:
> I haven't tested it, but there may be some problems with that implementation (blind guesses):

me neither yet, I'm not at my dev machine currently, but from the "old" caching mechanism for Mapsforge / GMv2 I'd say...

> * while zooming out `includes` will never return true, thus it may still trigger a lot of requests

true, as it cannot be guaranteed that there aren't more caches outside the currently stored viewport.
But at least, the viewport that gets saved is the outer boundary of all (GC) caches that got returned for the last stored query.
(Downside of this is: If you have an area with only one or two caches returned by the query, the stored viewport may be very small. But the same logic may be positive, if some caches are missing due to a size limit having been reached, see next point.)

> * when being at zoom level "world" caches by far won't get loaded for the whole viewport. While zooming in, `includes` is always true so no requests will happen at all?

There is a size limit. If the last query returned at least x caches, it will still fire a new request to take care of "white areas" where it hasn't retrieved any caches for due to limit being reached. x should be less, but close to the maximum a query for caches will return for GC.

> I would suggest to compare the viewport size instead and reload when the relative zoom difference exceeds some bounds

This sounds quite arbitrary to me (you may miss quite some caches in high-density areas). Or I don't understand the proposal. :)

moving-bits:
Decision from today's dev meeting: Let's merge and recheck if we still get so many 429 errors

I'd like to add: If test is successful, put it into `release` as well

PR #15944, opened 2024-08-02
hint should be tooltipText on buttons

ztNFny:
https://github.com/cgeo/cgeo/pull/15915 added "hint" to Button / ImageView elements instead of the intended tooltipText which should only be used for TextView as placeholder text

### DIFF
--- a/main/src/main/res/layout-land/map_settings_button.xml
+++ b/main/src/main/res/layout-land/map_settings_button.xml
@@ -16,7 +16,7 @@
             android:id="@+id/map_settings_popup"
             android:src="@drawable/map_quick_settings"
             style="@style/map_quicksettingsbutton"
-            android:hint="@string/quick_settings"
+            android:tooltipText="@string/quick_settings"
             android:contentDescription="@string/quick_settings"/>
     </LinearLayout>
     <LinearLayout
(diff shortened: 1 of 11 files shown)


moving-bits:
Thanks for fixing this. Play Store linked to this page: https://support.google.com/accessibility/android/answer/7158690#zippy=%2Cview. I interpreted this as "use `contentDescription` for ImageViews, `hint` for everything else", but probably I misinterpreted that for our buttons.

ztNFny:
Question is, do we use the tooltipText or remove it completely?

fm-sys:
We already use tool tips at other places in cgeo. They are quite useful IMHO, not just for accessibility reasons.

Screenshot_20240802-232044

moving-bits:
Having tooltips is definitely useful, they should be added where they are missing currently. I'm not sure, though, if they help with accessibility - to my understanding `contentDescription` needs to be added as well.

ztNFny:
Retest this please

fm-sys:
AFAIK tooltips are used by screen readers as well...

bekuno:
Primary we should use `contentDescription` (https://developer.android.com/reference/android/view/View#attr_android:contentDescription). Additionally we can use `tooltipText` (https://developer.android.com/reference/android/view/View#attr_android:tooltipText) for a short tool tip (for API 26+).

ztNFny:
let's merge this as it fixes the bug for button display. Other accessibility improvements may be added later independent of this PR

fm-sys:
Honestly, how many blind people use c:geo at API versions <26? 

I believe adding tooltips is (more than) enough. Everything else would be over the top IMHO.

 But as pointed out by @ztNFny, we can discuss this later on...

fm-sys:
@ztNFny could you re-target this PR to release please?

ztNFny:
> @ztNFny could you re-target this PR to release please?

done

PR #15945, opened 2024-08-03
fix #15935: open list selection with previous selected list in view

### LINKED ISSUE #15935: Stored Lists: tapping on list selection opens list selection on top (not where the currently selected list is located)
Request from a [Support Ticket #324796](https://cgeosupport.droescher.eu/scp/tickets.php?id=8313)
c:geo version 2024.07.26-RC

A user on support asks for the new list selection dialog for Stored lists (opened when tapping on currently shown list in header) opens already scrolled to the list currently selected in "Stored List" Activity. Currently, it always opens at the top of the list - which requires lots of scrolling if one just wants to move to the "next list".

Example: in the following screenshot, a list with name "white" is opened. When user taps on this list name, the list selection dialog opens, but the list "white" is not visible (one needs to scroll down to see it).

image
 

Original request:

I use Lists a lot in c:geo and have created a lot of Lists.
When being in a List of Caches, and want to navigate to another List, I click on the downward pointing arrow at the top, next to the List name.

Before the latest Release, that would have opened a drop-down-list of all Lists.
Which is indeed the expected behavior of clicking on such a downward pointing arrow.
More, that drop-down-list of all Lists opened positioned at the level of the current List.
That made navigation to adjacent Lists very easy.
Also, the width of the drop-down-list was automatically resized to the length of the lists names, if needed to the full screen width. 
Perfect !

Since the latest Release, clicking on the downward pointing arrow at the top now opens a 'Pick a list' window.
That window always opens at the top of the list of Lists, instead of at the level of the current List.
To navigate to a List adjacent to the one I am in, I have to scroll down the full list.
Clumsy.
Also, the width of that object is not resized.

It seems that in the current Release, the list of List used when being in a List, has the same behavior now as the List of lists used when being in a Cache and storing it in Lists.
It shouldn't, as these two usecases are very different.

I  would like to ask to go back to the previous drop-down-list of all Lists.

If this is not possible, I  would like to ask to open the 'Pick a list' window positioned at the level of the current List.

eddiemuc:
fix #15935: open list selection with previous selected list in view

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -554,7 +554,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         new StoredList.UserInterface(this).promptForListSelection(R.string.gpx_import_select_list_title, listId -> {
             new GPXImporter(CacheListActivity.this, listId, importGpxAttachementFinishedHandler).importGPX();
             switchListById(listId);
-        }, true, 0, listNameMemento);
+        }, true, Collections.singleton(0), -1, listNameMemento);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


eddiemuc:
image

PR #15949, opened 2024-08-04
fix #15857: allow selection of own images on cache log

### LINKED ISSUE #15857: Allow selection of images stored on cache when writing a log entry
### Tell us your idea!

Feature Request is from another user

Sometimes, when you store own images on a cache, you want to use one or several of those images also on a log file you write for that very same cache. As of now, c:geo does not allow for an easy selection of those images (one would have to navigate manually down the very non-intuitive directory structure of the GeocachePhotos folder in her file system).

It would be nice to have an easy way to select such images in a log file

I suggest to add a third "image select" button (next to "select image" and "take photo") which will open the image selection file dialog in the GeocachePhotos-subfolder of that cache. Button should only be available if there are any photos stored for the cache.

### Is there an alternative?

Do nothing

### Additional Information

_No response_

eddiemuc:
fix #15857: allow selection of own images on cache log


eddiemuc:
image
image

fm-sys:
The icon isn't really good IMHO. Not sure if we can get better, but we should at least add a tooltip text

eddiemuc:
> The icon isn't really good IMHO. Not sure if we can get better, but we should at least add a tooltip text

I assume we should add tooltip to all three buttons in this case.

Any suggestion for a better icon? ("Star" seems to be the symbol for "owned" inside c:geo)

Lineflyer:
> Any suggestion for a better icon? ("Star" seems to be the symbol for "owned" inside c:geo)

Maybe a picture icon with a star in its corner to symbolize"own" +"image".

eddiemuc:
tooltips were added

eddiemuc:
> > Any suggestion for a better icon? ("Star" seems to be the symbol for "owned" inside c:geo)
> 
> Maybe a picture icon with a star in its corner to symbolize"own" +"image".

Maybe this one?
image

eddiemuc:
image

Issue #15929, opened 2024-07-28
Optionally Load images in their original dimensions


ppete2:
### Tell us your idea!

Sometimes it's important for solving Caches to see Listing images in their original dimensions - and not their downsized version. Right now c:geo just displays the 640 px width version of images in the "Images"-tab. Please implement the optional displaying of at least Listing images in their original dimensions - maybe also Log images, but this is less important.

An example: Open Cache https://coord.info/GC1CT3N which contains an image named "Testbild ..."

Using the URL which is also used by c:geo https://img.geocaching.com/cache/large/409a5f5f-abe2-4a31-aadc-b0eec2157e79.jpg we get a downsized image version of 640x427px where the text on the board in the image is not readable any more.
Using the URL of the original sized image https://img.geocaching.com/cache/409a5f5f-abe2-4a31-aadc-b0eec2157e79.jpg we get the original-sized image (=1024x683 px) , where we could read the text.

eddiemuc:
Current logic is:
* load images from the links at the "bottom" of the page first ("spoiler images")
* then also load images from image links in URL IF they are not also present in list at listing bottom. This is checked by using the image filename only (not looking at whole URL). 

Assumption behind this logic was that if an image is present in both the listing AND in the bottom of the listing then the image in the bottom is of higher quality (hence the "large" in the URL).

For the listing in question it seems to be the other way around: the "large" image is of lower quality than the "non-large". This is strange. Is this an error in the listing or is this a common thing?

ppete2:
The _"large"_ within the URL is somewhat missleading, but has been introduced because there are even smaller version of images (e.g. for Gallerys). The format of image-URLs on the bottom of the Listing ("spoiler images") as well as log images has always be this way:

1a) **Spoiler** images, **reduced** to 640px: https://img.geocaching.com/cache/large/409a5f5f-abe2-4a31-aadc-b0eec2157e79.jpg

1b) **Spoiler** images, **original** dimension:  https://img.geocaching.com/cache/409a5f5f-abe2-4a31-aadc-b0eec2157e79.jpg

2a) **Log** images, **reduced** to 640px: https://s3.amazonaws.com/gs-geo-images/6f8c2c23-ce48-4f9f-8978-a314cbd6120d_l.jpg

2b) **Log** images, **original** dimension: https://img.geocaching.com/cache/log/6f8c2c23-ce48-4f9f-8978-a314cbd6120d.jpg

Lineflyer:
[...]

PR #15963, opened 2024-08-09
Dynamic image size

ztNFny:
Since 2016 there's actually some code to load original instead of "/display/", but only for that 1 out of 4 generated sizes. And it applies only to listing background and spoiler images (but not to inline images which are more commonly used today).

PR adds code to handle all 5 sizes, applies it to all listing images, always requests the original one and fixes some smaller issues.

Open question:

- (How) do we let the user decide about the prefered size (New setting?)?
- What's the default size?
- Is it the same for all types of images (listing, listing background, TB, log)?

Open todos:

- Apply the same to logs, TBs. 
- Rewrite the listing html code before loading it to apply the different image size there as well - right now it only applies to images tab, resulting in the image (maybe) getting loaded twice.

Discussion in https://github.com/cgeo/cgeo/issues/15929 please

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/ImageUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/ImageUtils.java
@@ -3,6 +3,7 @@ package cgeo.geocaching.utils;
 import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
 import cgeo.geocaching.models.Image;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.ContentStorage;
 import cgeo.geocaching.storage.Folder;
 import cgeo.geocaching.storage.LocalStorage;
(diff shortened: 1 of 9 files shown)


main/src/main/java/cgeo/geocaching/utils/ImageUtils.java
```diff
@@ -2,6 +2,7 @@
 
 import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
+import cgeo.geocaching.connector.gc.GCParser;
```
eddiemuc, line 5: We should not create a dependency from a general purpose class to a service-specific connector.

Imho the image url conversation logic should be moved to a method inside imageutils, but commented there to be for GC
 That would reflect both that the logic is for for GC only but also reflect that we decided nevertheless to apply it to all images (regardless the source, eg oc or others). 

The last aspect is of utter importance because in the future for all changes we have to make sure that they don't negatively affect images of other service sources. This can't be reached if logic is in GC connector (a developer then has to assume that changes would only affect GC service and nothing else)

eddiemuc (changes requested):

ztNFny:
This currently breaks loading Images in some caches, need to investigate, e.g. GC8BXXN

ztNFny:
This currently breaks loading Images in some caches, need to investigate, e.g. GC8BXXN

ztNFny:
Rewrote the code heavily to apply basically everywhere and be a bit cleaner.
All that's left is adding a preference - and I wonder what would be a good place for that one. Maybe in the GC connectors settings? Default I'd like to suggest Original

moving-bits:
> All that's left is adding a preference - and I wonder what would be a good place for that one. Maybe in the GC connectors settings? Default I'd like to suggest Original

If I understand this PR correctly (haven't fully followed the linked discussion), it will apply to images for GC caches only - correct? Then the GC connector settings would be a good place.

(Follow-up question: Do lab adventures use some similar scaling variants? Then applying it to images loaded for labs would be a nice addition. Setting could still stay in GC connector settings, as those are the base for labs anyway.)

ztNFny:
Preference added in Geocaching.com section:
image

Should be good to go now.
Possible enhancement for future: Option to load "small" by default, then later on load "large" on demand from image dialog. This would require a lot of changes though, like even storing which size was downloaded and UI changes in image viewer.

PR #15967, opened 2024-08-11
Apply team's decision to stick with pencil as log icon (closes #15709)

### LINKED ISSUE #15709: Undo New Log-Symbol in Current Nightly
### Tell us your idea!

The former pencil has been replaced with a smiley in one of the last nightly releases.

Why?

Not all logs you will provide for a cache will result in a Found It = Smiley.

Placing a note or something else is not being reflected by a smiley, while an universal pencil is relfecting any log type. 

Switch smiley back to former pencil

### Is there an alternative?

no, the smiley is just irritating

### Additional Information

_No response_

fm-sys:
see #15709


moving-bits:
failing test is unrelated, build runs fine when compiled locally, so let's merge

PR #15968, opened 2024-08-11
beautify gradle task  copyDefaultPreferencesToAndroid

bekuno:


bekuno:
We get now new errors - for all used CI executors.
Sample:
~~~
23:45:48  cgeo.geocaching.ExampleEspressoTest > simple[android-30(AVD) - 11] [31mFAILED [0m
23:45:48  	androidx.test.espresso.base.RootViewPicker$RootViewWithoutFocusException: Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus. Root:
23:45:48  	Root{application-window-token=android.view.ViewRootImpl$W@d7e00b5, window-token=android.view.ViewRootImpl$W@d7e00b5, has-window-focus=false, layout-params-type=1, layout-params-string={(0,0)(fillxfill) sim={forwardNavigation} ty=BASE_APPLICATION wanim=0x10302fe
~~~
@fm-sys Maybe this is a side effect from using `debugBuildAppId` now on CI?
@Lineflyer @kumy @mucek4 Besides the preferences file - is there something additionally to configure (like on a new (virtual) device?

kumy:
As far as I remind there are only 4 files to configure and nothing related to Android or Java, only cgeo conf. It's documented in the readme there https://github.com/cgeo/cgeo-executor

bekuno:
I got the configuration assistent on a new AVD on my AS installation, but I imported a own prefenrence file before.
So, it was unclear for me, if probally something related to the data structure could be necessary. (e.g. create data folder)

bekuno:
retest this please

PR #15981, opened 2024-08-16
Prevent stacking of navigation popups (fix #15937)

### LINKED ISSUE #15937: Map: Cache Quick View: tapping and holding "navigate" and swiping up will open dozens of identical popups
Support Ticket #679403
c:geo problem report (2024.07.26-RC)

Bug reported by user and reproducible for me.

* Open a cache on map in the quick view (half screen)
* press and hold the "navigate" button. 
* When the navigate popup show, keep pressing and swipe up. 

--> You will end up with tens of popups asking HOW to navigate. (on top of each other)
image

image
(it is a bit hard to see, but what we have here is many many identical popups on top of each other)

Desired behaviour: popup should only be opened once.

moving-bits:
## Description
Long-tapping on navigation button in cache infosheet leads to stacking of navigation app selection, as the menu items fires multiple times.
This PR prevents this by disabling the menu item (and the dialog re-enabling it on dismissal).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -958,7 +958,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
      * referenced from XML view
      */
     public void showNavigationMenu(@SuppressWarnings("unused") final View view) {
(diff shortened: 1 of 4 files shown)


moving-bits:
failing tests seem to be unrelated, let's test this PR in the nightlies

PR #15984, opened 2024-08-18
Add UnifiedMap support for Mapsforge (non-VTM)

moving-bits:
## Description
Optionally offer Mapsforge bitmap-based maps for UnifiedMap.

Implementation is based on the first commit of @fm-sys in many areas, and using @eddiemuc's implementation of `GeoItemLayer` for Mapsforge.

Supports both offline and online maps from Mapsforge (except "combined" and "user-defined"). Rotation by gestures does work already, general labels stay horizontally.

To activate Mapsforge in UnifiedMap, create a new setting named `useMapsforgeInUnifiedMap`of type `boolean` and set it to `true`.

When activated, it adds all supported (and not deactivated) map sources to the map selection menu, adding a "(MF)" suffix to the name.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -914,6 +914,7 @@
         </activity>
         <activity android:name=".maps.mapsforge.v6.RenderThemeSettings" />
         <activity android:name=".unifiedmap.mapsforgevtm.MapsforgeThemeSettings" />
+        <activity android:name=".unifiedmap.mapsforge.MapsforgeThemeSettings" />
 
         <!-- fileprovider to be able to share files -->
         <provider
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -366,12 +366,12 @@ dependencies {
     // -- Mapsforge / Mapsforge --
 
     // Mapsforge official version
-    String mapsforgeSource = 'org.mapsforge'
-    String mapsforgeVersion = '0.21.0'
+    // String mapsforgeSource = 'org.mapsforge'
+    // String mapsforgeVersion = '0.21.0'
 
     // Mapsforge Snapshot from official master branch (via https://jitpack.io/#mapsforge/mapsforge)
-    //String mapsforgeSource = 'com.github.mapsforge.mapsforge'
-    //String mapsforgeVersion = 'fbc75a4' // master as of 18.07.22
+    String mapsforgeSource = 'com.github.mapsforge.mapsforge'
+    String mapsforgeVersion = 'd866ea2' // master as of 2024-08-15, including rotation & fractional zoom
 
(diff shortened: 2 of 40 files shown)


eddiemuc:
> some interactions with map (tapping on cache, route etc.)

Recognition of tapping on thoae things is handled by the GeoItemLayer

moving-bits:
> > some interactions with map (tapping on cache, route etc.)
> 
> Recognition of tapping on thoae things is handled by the GeoItemLayer

Thanks, taphandler layer for Mapsforge was missing in my implementation. I'll update my PR.

moving-bits:
failing tests are unrelated, removing draft status

moving-bits:
as discussed in today's dev meeting: let's merge to master to get broader testing

reminder: you need to set a config key to see anything different

eddiemuc:
@moving-bits my last implementation of the unifiedmap geoitemlayer for mapsforge was rather crude (talking about cgeo.geocaching.unifiedmap.geoitemlayer.MapsforgeV6GeoItemLayer) because at the time this was just a proof-of-concept for the item layer concept to me. (I thought we were going VTM at the time, I assume all of us did).

Many things were done to the VTM layer since. Is this implementation still working ok with the now-added mapsforge support to unifiedmap? If would actually surprise me if everything just works out of the box...

moving-bits:
Haven't tested in-depth or even done a cache tour with it yet, but quick test is ok. But that's just cache/waypoint icons and individual route, no tracks, no GeoJSON etc.

eddiemuc:
> Haven't tested in-depth or even done a cache tour with it yet, but quick test is ok. But that's just cache/waypoint icons and individual route, no tracks, no GeoJSON etc.

Ok for now. Let's just say I expect that things will come up :-)

ztNFny:
should we make this available in GUI instead of just a hidden option? If VTM dies then we should get nightly feedback asap i guess

fm-sys:
> > Haven't tested in-depth or even done a cache tour with it yet, but quick test is ok. But that's just cache/waypoint icons and individual route, no tracks, no GeoJSON etc.
> 
> Ok for now. Let's just say I expect that things will come up :-)

@eddiemuc you mentioned that some features are not implemented yet for Mapsforge. Probably we should track them somewhere.

moving-bits:
> should we make this available in GUI instead of just a hidden option? If VTM dies then we should get nightly feedback asap i guess

What do you have in mind - a simple setting in the "Map sources - Unified Map" section to activate Mapsforge in addition to VTM?

Or do you plan to swap precedence - let Mapsforge be the new default for UnifiedMap, and use a (maybe visible) setting to use VTM additionally?

ztNFny:
I'd first go with an option next to the unified toggle.

PR #15986, opened 2024-08-19
fix #15439: add copy-intent to chooser on text share

### LINKED ISSUE #15439: Copy to clipboard not possible
### Describe your problem!

In the main page of a cache copy to clipboard not possible. Eg  coodinates, name, gc code,...

### How to reproduce?

Open cache
Long click on gc-code

menu for text share appears (list of other apps that use text as an input), 3 pages at my phone.
IMG_20240310_153120

possibility to copy text into clipboard

### Reproducible

Yes

### c:geo Version

2024.03.10-NB-7871581

### Additional Information

I checked the options if the "long click" behaviour can be changed, but found nothing. As a workaround it's possible to chose a text editor and copy it into the clipboard from there.

_No response_

eddiemuc:
fix #15439: add copy-intent to chooser on text sharem using the mechanism suggested by @fm-sys 

It will solve the issue, but it will also look "duplicated" on system which provide a clipboard-share on their own. Examples:

This is an Android 33 system. It provides it's own "copy" action. The c:geo "copy" intent is for some reason shown using the c:geo application icon instead of the icon provided by c:geo:
image

This is an Android 28 system. It provides now a "copy" intent (this one is from c:geo) and a "copy to clipboard" intent (this one is from the system):
image

I put it up to discussion how to proceed here


moving-bits:
> I put it up to discussion how to proceed here

This is a helpful extension for the affected users, but the duplication does not look nice and may confuse many users... I vote for using a setting to activate it. That way it would not harm most users, and those who are in need of such an entry can activate it.

(If we want to avoid a (visible) setting, we could make this setting hidden, so users on support claiming the missing "copy" function could be given the instruction on how to create the key.)

fm-sys:
IIRC it was mentioned that all tickets come from Xiaomi users. Although it's not a nice solution, I would vote to use an opt-in approach per manufacturer starting with a list just containing Xiaomi and extend it as soon as we should get any support tickets from another manufacturers device as well. Seeing that "copy" is a default android feature at least since API 28, we shouldn't annoy the big mayority of users IMHO.

main/src/main/java/cgeo/geocaching/utils/ShareUtils.java
```diff
@@ -78,10 +79,18 @@ private static void shareAsEmail(final Context context, final String subject, fi
 
     private static void shareInternal(final Context context, final Intent intent, @StringRes final int titleResourceId) {
```
fm-sys, line 80: `shareInternal` gets used for many different types of share intents, including e.g. file share. Maybe make it a text share only feature?

eddiemuc:
> IIRC it was mentioned that all tickets come from Xiaomi users. Although it's not a nice solution, I would vote to use an opt-in approach per manufacturer starting with a list just containing Xiaomi and extend it as soon as we should get any support tickets from another manufacturers device as well. Seeing that "copy" is a default android feature at least since API 28, we shouldn't annoy the big mayority of users IMHO. 

I strongly vote against any hidden magic decision logic whether to show this or not. Setting would be the right approach IMHO

MagpieFourtyTwo:
> IIRC it was mentioned that all tickets come from Xiaomi users.

Honor has the same problem.  And since Huawei is ... hrm ... "kind of very similar", they most likely have it, too,

main/src/main/java/cgeo/geocaching/utils/ShareUtils.java
eddiemuc: Ok, restricted to text input

eddiemuc:
Build in a setting to enable the copy intent. Defaults to false.

image

PR #15995, opened 2024-08-22
Use bigDecimal for converting double to string in formula

murggel:
## Description

Using `BigDecimal` for more precision for decimal values in evaluation of value to string in formula.

## Additional context

I had a mystery where I had to calculate the checksum of a high pow-value (which is a double value)
With the current implementation not all necessary digits are shown (the scientific-notation is used).
Switching to `BigDecimal` all digits are shown

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/formulas/Value.java
+++ b/main/src/main/java/cgeo/geocaching/utils/formulas/Value.java
@@ -3,6 +3,7 @@ package cgeo.geocaching.utils.formulas;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
@@ -62,7 +63,8 @@ public class Value {
             if (raw == null) {
                 asString = "";
             } else if (raw instanceof Number && !(raw instanceof Integer) && !(raw instanceof BigInteger)) {
-                asString = DOUBLE_TO_STRING_FORMAT.format(((Number) raw).doubleValue());
+                final BigDecimal bigDecimalValue = new BigDecimal(((Number) raw).doubleValue());
+                asString = DOUBLE_TO_STRING_FORMAT.format(bigDecimalValue);
             } else {
                 asString = raw instanceof String ? (String) raw : raw.toString();
             }
--- a/main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
+++ b/main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
@@ -61,6 +61,13 @@ public class FormulaTest {
     }
 
     @Test
+    public void bigDecimals() {
+        final int doublePowCs = 847; // manually calculated checksum of pow(99^99)
+        assertThat(Formula.evaluate("cs(99^99)").getAsDouble()).isEqualTo(doublePowCs);
+        assertThat(Formula.evaluate("99^99").getAsDouble()).isEqualTo(Math.pow(99, 99));
+    }
+
+    @Test
     public void complex() {
         assertThat(eval("-2.5 + 3 * (4-1) + 3^3")).isEqualTo(33.5d);
     }
--- a/main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
+++ b/main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
@@ -62,11 +62,14 @@ public class FormulaTest {
 
     @Test
     public void bigDecimals() {
-        final int doublePowCs = 847; // manually calculated checksum of pow(99^99)
-        assertThat(Formula.evaluate("cs(99^99)").getAsDouble()).isEqualTo(doublePowCs);
+        final double doubleResult = 3.697296376497268E197;
         assertThat(Formula.evaluate("99^99").getAsDouble()).isEqualTo(Math.pow(99, 99));
-        assertThat(String.valueOf(Formula.evaluate("99^99").getAsString())).isEqualTo("369729637649726802192985226395427290145296428445515959701359650120802601667133273280053721002700400354392780458116125965728631706472588849812738072765460822138161108630185181415759762204338929270784");
+        assertThat(Formula.evaluate("99^99").getAsDouble()).isEqualTo(doubleResult);
         assertThat(String.valueOf(Formula.evaluate("99^99").getAsDouble())).isEqualTo("3.697296376497268E197");
+
+        final int doublePowCs = 847; // manually calculated checksum of pow(99^99)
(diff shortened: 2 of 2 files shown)


murggel:
I don't know, how to avoid / supress the new codacy warning (constructor with double is available for BigDecimal)

bekuno:
retest this please

main/src/main/java/cgeo/geocaching/utils/formulas/Value.java
```diff
@@ -62,7 +63,8 @@ public String getAsString() {
             if (raw == null) {
                 asString = "";
             } else if (raw instanceof Number && !(raw instanceof Integer) && !(raw instanceof BigInteger)) {
-                asString = DOUBLE_TO_STRING_FORMAT.format(((Number) raw).doubleValue());
+                final BigDecimal bigDecimalValue = new BigDecimal(((Number) raw).doubleValue());
```
eddiemuc, line 66: It seems odd to create a BigDecimal from a double and expect a more precise outcome from it. (I guess this is also what the codacy warning is about). 

I don't doubt the need of using BigDecimal instead of double, but I feel this should be done from a more fundamental base (using BigDecimal during checksum calculation)

murggel: > It seems odd to create a BigDecimal from a double and expect a more precise outcome from it. (I guess this is also what the codacy warning is about). 
> 

But that is indeed the case for the transformation to a string value, cause the double creates a scientific string with exponential form, whereas the BigDecimal creates the "long" form

murggel: > I don't doubt the need of using BigDecimal instead of double, but I feel this should be done from a more fundamental base (using BigDecimal during checksum calculation)


But it's not only for checksum...

murggel: > > It seems odd to create a BigDecimal from a double and expect a more precise outcome from it. (I guess this is also what the codacy warning is about). 
> > 
> 
> But that is indeed the case for the transformation to a string value, cause the double creates a scientific string with exponential form, whereas the BigDecimal creates the "long" form
> 
> 

I will enhance the test case to show this behavior

main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
```diff
@@ -60,6 +60,15 @@ public void bigIntegers() {
         assertThat(Formula.evaluate("cs(" + longMaxBeyond + ")").getAsInt()).isEqualTo(longMaxCs + 1);
     }
 
+    @Test
+    public void bigDecimals() {
+        final int doublePowCs = 847; // manually calculated checksum of pow(99^99)
+        assertThat(Formula.evaluate("cs(99^99)").getAsDouble()).isEqualTo(doublePowCs);
+        assertThat(Formula.evaluate("99^99").getAsDouble()).isEqualTo(Math.pow(99, 99));
+        assertThat(String.valueOf(Formula.evaluate("99^99").getAsString())).isEqualTo("369729637649726802192985226395427290145296428445515959701359650120802601667133273280053721002700400354392780458116125965728631706472588849812738072765460822138161108630185181415759762204338929270784");
+        assertThat(String.valueOf(Formula.evaluate("99^99").getAsDouble())).isEqualTo("3.697296376497268E197");
```
eddiemuc, line 69: Assert should be for double value, not for its string representation

eddiemuc (commented):
I like the added unit tests. Thanks for that.

There is clearly something fishy going on here. Checksum calculation should not use double values at all but integer values instead, and I dimly remember introducing BigInteger some time ago. I am currently not sure whether this PR fixes the root error or just a symptom of it. I would need to take a closer look into code

murggel:
I think the problem is not the checksum-function here, but the string representation of the double value.
Maybe the culprit lies in that the pow-function returns a double-value and not an integer, so it will be transformed to a string to calculate the checksum.
I can check that ...

But even if I don't want to calculate the checksum but just want to show the double value, it only gives the scientific notation.

So I thought the change in getAsString would fix all cases

murggel:
I think the problem is not the checksum-function here, but the string representation of the double value.
Maybe the culprit lies in that the pow-function returns a double-value and not an integer, so it will be transformed to a string to calculate the checksum.
I can check that ...

But even if I don't want to calculate the checksum but just want to show the double value, it only gives the scientific notation.

So I thought the change in getAsString would fix all cases

murggel:
> ... I dimly remember introducing BigInteger some time ago. I am currently not sure whether this PR fixes the root error or just a symptom of it. I would need to take a closer look into code

I guess introducing BigDecimal would be the better way to go...
I will give that approach a try

main/src/test/java/cgeo/geocaching/utils/formulas/FormulaTest.java
murggel: Thats in line 67...

murggel: Even not sure, if it is the same on all machines..🤔

eddiemuc:
I had a little time and I think I found the culprit. It is this code part in class `Formula`:
image

Here, the "power"-function is actrually calculated, and it uses `double Math.pow(double, double)`. In this calculation, precision already gets lost for such high numbers.

This part of code could be change to use `BigInteger.pow` or `BigDecimal.pow` instead (for the later, class `Value` has to be enabled to deal with `BigDecimal` too). But those variants of pow can only use integer as exponent. Is that accceptable? Something like this is then no longer possible:
image

@murggel class Value can handle BigInteger in a generic way in the sense that toString of Value will always return the full precision stringified integer value. There is however no "getAsBigInteger" method in it. Up to now this was not needed.

Code reference:
https://github.com/cgeo/cgeo/blob/4e5081ef44d55ab22ce0ecbd1c82ff2d1d333eb1/main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java#L634

eddiemuc:
Thinking more about this, it gets worse: imagine an expression like this

checksum(99^99+1)

99^99 already produces a long overflow, so power must be calculated using BigInteger in order not to loose digits (and thus give a wrong checksum). But then 1 is added, so plus operator also has to use BigInteger. So every basic math operation has to use BigInteger/BigDecimal in case any operation results in a long/double overflow. 

That would become very expensive computation-wise. I am leaning towards leaving everything as is and accept overflow cases. Should be seldom in Real-Life anyway.

murggel:
You are probably right.
I thought BigDecimal would only have an effect for the string representation without exponent. If thet would have been the case, then all other operations with double would have had benefit of that.
But indeed, the pow-function here gives a wrong result (then, the test case unfortunately deals with wrong expectations 😥) 

So I agree then with you to keep it as it is.

PR #16001, opened 2024-08-26
Feature/issue 15756 projection variables

### LINKED ISSUE #15756: Waypoint projection does not instantiate variables
### Describe your problem!

Using variables in waypoint projection (see #12600) does not create variables.

### How to reproduce?

1. Create a new waypoint
2. Set starting coordinates
3. Select "projection"
4. Enter variables in heading/distance
5. Save waypoint

No variables were created in the variables page.

Variables should be created in the variables page like it's done for the regular waypoint calculator. Additionally it would be nice to have the same quick access to variables as in the waypoint calculator.

### Reproducible

Yes

### c:geo Version

2024.05.11-NB-c926353

### System information

_No response_

### Additional Information

_No response_

murggel:
## Description

This PR creates variables used in projection and shows them in the EditWaypointActivity.

I looked around in CoordinatesCalculateGlobalDialog and tried to used common logic.

## Related issues

fix #15756

## Additional context

grafik

### DIFF
--- a/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
@@ -38,7 +38,6 @@ import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.formulas.Formula;
-import cgeo.geocaching.utils.formulas.FormulaUtils;
 import cgeo.geocaching.utils.formulas.VariableList;
 import cgeo.geocaching.utils.html.UnknownTagsHandler;
 import static cgeo.geocaching.models.Waypoint.getDefaultWaypointName;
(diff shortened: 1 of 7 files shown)


murggel:
I will fix the codacy-issues first

main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -391,11 +397,29 @@ public static ImmutableTriple<FormulaEditText, FormulaEditText, TextSpinner<Dist
         }
     }
 
+    private Set<String> getNeededVariablesForProjection() {
+        final Set<String> projectionVars = new HashSet<>();
+        final ProjectionType pt = projectionType.get();
+        if (pt == ProjectionType.BEARING) {
+            FormulaUtils.addNeededVariables(projectionVars, binding.projectionBearingDistance.getFormulaText());
```
eddiemuc, line 404: `FormulaUtils.addNeededVariables` will unnecessarily recompile the formula while it is already available in `FormulaEditText`. Suggestion: add a method `getNeededVars` to class `FormulaEditText` and use this method instead.

eddiemuc (approved):
lgtm. comment is not critical. Thanks for your work!

main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
murggel: done

main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -391,11 +396,28 @@ public static ImmutableTriple<FormulaEditText, FormulaEditText, TextSpinner<Dist
         }
     }
 
+    private Set<String> getNeededVariablesForProjection() {
+        final Set<String> projectionVars = new HashSet<>();
+
+        final ImmutableTriple<FormulaEditText, FormulaEditText, TextSpinner<DistanceUnit>> fields = getFieldsForProjectionType();
+        if (fields.left == null) {
```
eddiemuc, line 403: I assume the check has to be then other way around? (`... !=null`)
As is the code will produce an NPE in next line if `fields.left` is null

murggel: Ups

murggel: And the vars will not be added :-o

main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
```diff
@@ -391,11 +396,28 @@ public static ImmutableTriple<FormulaEditText, FormulaEditText, TextSpinner<Dist
...
+            fields.left.addNeededVariables(projectionVars);
+        }
+        if (fields.middle == null) {
```
murggel, line 406:
```suggestion
        if (fields.middle != null) {
```

main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
murggel, line 403:
```suggestion
        if (fields.left != null) {
```

eddiemuc:
Lgtm, I will merge. Thanks for your work!

PR #16012, opened 2024-08-29
Formulas: add numeric overflow protection

eddiemuc:
This PR adds numeric overflow protection to Formula evaluations (the necessity of this was triggered by some recent work of @murggel). It also contains some Formula-internal refactorings.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/CalculatedCoordinate.java
+++ b/main/src/main/java/cgeo/geocaching/models/CalculatedCoordinate.java
@@ -4,7 +4,6 @@ import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.utils.TextParser;
 import cgeo.geocaching.utils.formulas.DegreeFormula;
 import cgeo.geocaching.utils.formulas.Value;
-import cgeo.geocaching.utils.functions.Func1;
 import static cgeo.geocaching.models.CalculatedCoordinateType.PLAIN;
 
 import android.os.Parcel;
@@ -16,6 +15,7 @@ import androidx.annotation.Nullable;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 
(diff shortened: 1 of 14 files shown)


murggel:
Improving the pow-method is still in progress? The result for B is not correct. It should end with 9 (my calculations were 369729637649726772657187905628805440595668764281741102430259972423552570455277523421410650010128232727940978889548326540119429996769494359451621570193644014418071060667659301384999779999159200499899)

eddiemuc:
> Improving the pow-method is still in progress? The result for B is not correct. It should end with 9 (my calculations were 369729637649726772657187905628805440595668764281741102430259972423552570455277523421410650010128232727940978889548326540119429996769494359451621570193644014418071060667659301384999779999159200499899)

No. B explicitely calculates with double values (in contrast to A) and thus accepts the reduced precision. double can't hold a precise value.

eddiemuc:
> > > Improving the pow-method is still in progress? The result for B is not correct. It should end with 9 (my calculations were 369729637649726772657187905628805440595668764281741102430259972423552570455277523421410650010128232727940978889548326540119429996769494359451621570193644014418071060667659301384999779999159200499899)
> > 
> > 
> > No. B explicitely calculates with double values (in contrast to A) and thus accepts the reduced precision. double can't hold a precise value.
> 
> But for the user it will show a wrong result. So this should be reflected somehow - then probably show the exponential notation and not all digits

If this is a problem, then I suggest to restrict the range for double calculations such that 4 digits after the decimal point can still be stored (and consider higher double valus as overflow)

murggel:
> > > > Improving the pow-method is still in progress? The result for B is not correct. It should end with 9 (my calculations were 369729637649726772657187905628805440595668764281741102430259972423552570455277523421410650010128232727940978889548326540119429996769494359451621570193644014418071060667659301384999779999159200499899)
> > > 
> > > 
> > > No. B explicitely calculates with double values (in contrast to A) and thus accepts the reduced precision. double can't hold a precise value.
> > 
> > But for the user it will show a wrong result. So this should be reflected somehow - then probably show the exponential notation and not all digits
> 
> If this is a problem, then I suggest to restrict the range for double calculations such that 4 digits after the decimal point can still be stored (and consider higher double valus as overflow)

Consider higher double value as overflow would help here, I guess 4 digits after the decimal point are too few

eddiemuc:
> > > > No. B explicitely calculates with double values (in contrast to A) and thus accepts the reduced precision. double can't hold a precise value.
> > > 
> > > But for the user it will show a wrong result. So this should be reflected somehow - then probably show the exponential notation and not all digits
> > 
> > If this is a problem, then I suggest to restrict the range for double calculations such that 4 digits after the decimal point can still be stored (and consider higher double valus as overflow)
> 
> Consider higher double value as overflow would help here, I guess 4 digits after the decimal point are too few, but don't have a clear idea of a good value
> 

A double in Java uses 52 bits for its mantisse. 2^52 is 4.503.599.627.370.496. if we want to keep 4 digits after decimal point, this would mean that we have to consider all doubles over/under +/- 450.359.962.737 as overflow already...

eddiemuc:
Not as easy as I thought. Suddenly not all integers are also doubles (because range of long is now bigger than double...). I need to rethink this. Setting PR to "do not merge"

eddiemuc:
I think I will give itna try to change formula calculation completely from using double/long towards BigDecimal/BigInteger. Formulas are usually not that complex and I believe that performance impact will be acceptable.

I will use this PR as a base for the change

eddiemuc:
Replaced by PR https://github.com/cgeo/cgeo/pull/16047

PR #16016, opened 2024-08-29
fix #16013: allow manual login on gc.com (eg on captcha problem)

### LINKED ISSUE #16013: Geocaching.com: CAPTCHA problem seems to get bigger and bigger
On support, we have multiple people each day with the annoying CAPTCHA-problem when trying to use geocaching.com. It seems that the number of users unable to resolve this with our known standard methods (see below) is increasing.

I open this issue for discussion:
* Are there any ideas on how to ease the pain on this one? (is it maybe somehow possible to build a workaround such that users can log into geocaching.com with their web browser, then copy out some cookies and paste it into c:geo? Just blind-shooting ideas...)
* Are there any other tips to overcome the issue which are NOT yet part of the text below?

Thanks for feedback!

Current support text for CAPTCHA problem:
your login fails in c:geo as the geocaching website for security reasons requires you to solve a [CAPTCHA](https://en.wikipedia.org/wiki/CAPTCHA) before login is possible again. We cannot influence on c:geo side under which conditions Groundspeak decides to trigger such a CAPTCHA, however we have collected much data on how to prevent it.

As far as we know the CAPTCHA can be triggered by the following situations:

You several times tried to login with a wrong password
You tried to login from a suspicious IP address (e.g. a public WiFi, from a foreign country, using VPN or anonymizer services)
The mail address stored with your profile at geocaching.com is not verified or no longer valid
To overcome this problem please proceed as follows:

Use your phone browser to go to geocaching.com, login onto the website and once solve the shown CAPTCHA.
Afterwards go to c:geo Start - Menu - Settings - Services - geocaching.com and use "Update authorization" to doublecheck your username and password and try to login again
If the problem still remains, please check that the following before trying again as described above:

Check on the Geocaching website [settings page](https://www.geocaching.com/account/settings/emailpreferences), that your mail address is valid and verified
Make sure you are not using any VPN services (such as NordVPN, TOR, etc.) on your device
Try to use another internet connection (e.g. mobile network instead of Wifi)
Another trick that has been known to help is triggering a "captcha removal" via the official geocaching.com app (which you can find in Google Play Store):

If you have already installed the official app: try to log into this app. After that, update your authorization in c:geo (switch to c:geo, go to settings => services => geocaching.com and long-tap on "update or remove authorization", then re-enter your credentials)
If you have not yet installed the official app: install it, log into itm then update your authorization in c:geo (switch to c:geo, go to settings => services => geocaching.com and long-tap on "update or remove authorization", then re-enter your credentials). If this way is successful, you may uninstall the official app again at your own preference
Further help on login problems can also be found in our [user guide](https://manual.cgeo.org/en/loginproblems). There is a special section dedicated to the captcha problem further down this page.

If nothing helps, feel free to contact us again.

eddiemuc:
fix #16013: allow manual login on gc.com (eg on captcha problem)


eddiemuc:
I think I finally found a way to handle the CAPTCHA problem: using a WebView it is possible to allow users to sign in on the webpage itself (through c:geo), then we can extract the cookies from this webpage and use it for c:geo authentification. This seems to work fine.

The way it is implemented right now: if the gc login fails (I simulated this using a wrong username, I don't know how to provoke a captcha) then a button "Manual login" appears:
image

Tapping it opens a fullscreen dialog whose main part is a webview pointing to the gc.com login page:
image

User shall log in (and probably complete a captcha while doing so). He ends up with something like this:
image

User shall then click ''ok'. It is then when c:geo extracts the cookies from the webview and tries to authenticate itself with it and tries to re-login:
image

..which then completes successfully:
image

eddiemuc:
Updates: through my excessive tests I produced the captcha error (apparently my behavior was suspicious :-)). I can thus verify that this PR gives user the ability to overcome the captcha problem.

moving-bits:
retest this please

fm-sys:
Great! 

We could even fill the credentials automatically via Java Script. We may decide to don't purpose that path as non technical users may feel uncomfortable that we know their login data and use them "without their permission". 

But at least we should be able to detect automatically if the login is completed.

eddiemuc:
> Great!
> 
> We could even fill the credentials automatically via Java Script. We may decide to don't purpose that path as non technical users may feel uncomfortable that we know their login data and use them "without their permission".
> 
> But at least we should be able to detect automatically if the login is completed.
> 
> If a site gets loaded in the webview and URL == https://www.geocaching.com/play/search then login is successful.

This would require diving deeper into Android WebView which I want to avoid on `release` branch. IMHO this should be saved for a possible extension of this now-quick-fix on the `master` branch. 

If you feel different, please modify the PR accordingly. Please test the scenario throughly before merging to `release` branch.

fm-sys:
Totally agree that this can be done in a follow up PR at master branch :)

moving-bits:
As those Captchas are a very frequent and annoying thing (for both our uses and for us on support) I'd like to see this in our next release, therefore I merge here now and set up a new beta version. (We need a new beta anyway due to Play Store's way of checking foreground service type declarations.)

PR #16039, opened 2024-09-05
Repair GC Manual login

eddiemuc:
Repair GC Manual login: show soft keyboard on textfield click


moving-bits:
Not directly related to the details of this PR, but to the login problem in general:

What happens if someone cannot do the first login at all, if the user gets a captcha error while authorizing against geocaching.com for the first time? Can the manual login scheme be extended to support this scenario as well?

PR #16074, opened 2024-09-14
Wherigo: add basic audio support

eddiemuc:
Wherigo: add basic audio support


eddiemuc:
This PR should conclude base implementations of all openwig interfaces I could find

PR #16075, opened 2024-09-14
Show testDir preference in developer builds only (fix #16071)

### LINKED ISSUE #16071: Collapsing empty space on system settings page
### Describe your problem!

Opening the "System" section of the settings shows an empty space that quickly collapses right after the page is shown. 
The purpose of this is unclear and confusing. 

Is there something wrong with system settings? 
Is there something gone from there (as a consequence of user actinos or other)
:confused

### How to reproduce?

1. go to settings
2. tap on system

Above "Geolocation" there is an empty space that takes just about a second to collapse. 
It was quite difficult to capture it in a screenshot: 
Screenshot_20240904-094300_cgeo

I see no purpose for this behaviour

### Reproducible

Yes

### c:geo Version

2024.09.01

### System information

_No response_

### Additional Information

_No response_

moving-bits:
Toggles testDir preference behavior:
Instead of having it included by default and removing it in non-developer builds, do not have it included by default, but add it dynamically in developer builds (to avoid the effect of vanishing blank space as described in #16071).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceSystemFragment.java
+++ b/main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceSystemFragment.java
@@ -16,6 +16,9 @@ import static cgeo.geocaching.utils.SettingsUtils.setPrefClick;
 import android.content.Intent;
 import android.os.Bundle;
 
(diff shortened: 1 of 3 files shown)


moving-bits:
retest this please

PR #16097, opened 2024-09-20
[UnifiedMap] enable live for adress search

murggel:
## Description

Enable live-option for showing map after search for address / coordinate 
(restore behavior of old map)

Consider UnifiedMapTypeType.UMTT_TargetCoords as "Live-Map"-able

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -157,12 +157,12 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
     private boolean overridePositionAndZoom = false; // to preserve those on config changes in favour to mapType defaults
 
     private final CacheListActionBarChooser listChooser = new CacheListActionBarChooser(this, () -> getSupportActionBar(), newListId -> {
-            final Optional<AbstractList> lNew = StoredList.UserInterface.getMenuLists(false, PseudoList.NEW_LIST.id).stream().filter(l2 -> l2.id == newListId).findFirst();
-            if (lNew.isPresent()) {
-                new UnifiedMapType(newListId).launchMap(this);
-                finish();
-            }
-        });
+        final Optional<AbstractList> lNew = StoredList.UserInterface.getMenuLists(false, PseudoList.NEW_LIST.id).stream().filter(l2 -> l2.id == newListId).findFirst();
+        if (lNew.isPresent()) {
+            new UnifiedMapType(newListId).launchMap(this);
+            finish();
+        }
+    });
 
     private UnifiedMapState lastMapStateFromOnPause = null;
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -194,15 +194,15 @@ public void onCreate(final Bundle savedInstanceState) {
             if (mapType != null) {
                 viewModel.followMyLocation.setValue(mapType.followMyLocation);
             } else {
-                viewModel.followMyLocation.setValue(Boolean.TRUE.equals(viewModel.followMyLocation.getValue()) && mapType.type == UMTT_PlainMap);
+                viewModel.followMyLocation.setValue(Boolean.TRUE.equals(viewModel.followMyLocation.getValue()));
```
moving-bits, line 197: why is "follow my location" involved in this?

Is the change in behavior (no longer checking map's type) intended?
If it's just about a NPE on checking `mapType`, then this should be checked separately.

murggel: Yes, it's just to avoid a NPE.
But you are right, probably it should be considered in the if-case.

I will create a  separate PR vor the NPE and revert this commit

moving-bits: Fixing the NPE is fine, and I wouldn't mind having it included in this PR. I just stumbled about the workflow change - not checking for `mapType.type == UMTT_PlainMap` anymore.
If it's just about the NPE, you could insert a `&& mapType != null` before the `&& mapType.type == UMTT_PlainMap` (assumed that the NPE came from checking `mapType.type`).

murggel: But that is the if around - so we are in the else-case "type == null"...
So there would be the NPE

But probably the check for the type belongs to the if-case "type != null"

I will check and test that separately

moving-bits: Ok, this is coming from old code from `NewMap`:

```
if (savedInstanceState != null) {
...
} else {
    if (mapOptions.mapState != null) {
        followMyLocation = mapOptions.mapState.followsMyLocation();
    } else {
        followMyLocation = followMyLocation && mapOptions.mapMode == MapMode.LIVE;
    }
...
}
```

There has been a differentiation between `mapOptions` and `mapState` back then. Not sure what this means for `UnifiedMap`, though, would need to dig in deeper.

murggel: I provide a separate PR with simplification - we can discuss it there

murggel: see PR #16102

moving-bits:
Sorry, but we have new merge conflicts after having merged #16102

PR #16102, opened 2024-09-21
Avoid NPE in unified map activity

murggel:
## Description

Simplify old code-segments from `NewMap` to avoid access to mapType.followMyLocation  where mapType is null

see https://github.com/cgeo/cgeo/pull/16097#discussion_r1769621736


moving-bits (approved):
LGTM, let's give it a try in the nightlies. Thanks for refactoring.

PR #16110, opened 2024-09-22
Add 'delete caches immediately' action (fix #15402)

### LINKED ISSUE #15402: FR: Add menu entry "Remove caches from device" (for offline lists)
### Tell us your idea!

From time to time users ask me how to get *reliably* rid of a bunch of caches (in the current list resp. filter).
Then I usually answer that ``Manage caches`` => ``Remove from list`` is what they are looking for.
Which in most cases is the moment when the first question marks start to appear on their faces ...
"But I don't just want to delete from *list,* I want to delete from *device*, completely ...?!"
It then usually takes some time until they understand that this is the same ...
... to immediately get even bigger question marks again, once I add that this does (of course) only work, if all these caches in question are in no other list(s) than just the current one ...
IOW: If a cache is in just the current list, it will be deleted from device, while those stored (resp. referenced) in multiple lists, live on in the other list(s) ...

And exactly that's the second when most average users are completely out.  ;)

Nevertheless, there sometimes are a few who even recover from this shock and, despite everything, ask again how this can be done reliably - and to those I then explain *my* procedure to completely delete caches from the device:
1. set a filter or select the caches in question
2. ``Manage caches`` => ``Move all`` (to an empty resp. a new list)
3. switch to this list
4. ``Manage list`` => ``Make list unique`` 
5. ``Manage caches`` => ``Remove all from list``

Maybe there is an easier way to completely and reliably delete caches from the device, but apart from the fact that even I (who can probably be considered a more experienced user) don't know how, it's honestly quite hard to argue (especially to less experienced users) why there isn't an easier resp. more intuitive way to do it.

However, none of these, not even the ones who understood, were really satisfied with my answers ... which I can fully understand. ;)

Add an menu entry like ``Remove all from device`` (resp. ``Remove selected from device (n)``) to the ``Manage caches`` menu list.
And if a cache is referenced in other lists, too, don't care, just entirely delete from the device.  For real.  ;)
And WRT the ``Undo`` function - if it's hard to recover all the other list references, I would just restore it to the current list.

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Adds a "Delete caches immediately" action to `CacheListActivity` => Manage Caches menu.

After a confirmed dialog this will remove the selected caches from database immediately (no 72 hours grace period) as well as the cache data folder in the local filesystem that holds log pictures etc. downloaded for those caches.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -678,6 +678,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             setEnabled(menu, R.id.menu_add_to_route, !isEmpty);
             setMenuItemLabel(menu, R.id.menu_add_to_route, R.string.caches_append_to_route_selected, R.string.caches_append_to_route_all);
             setVisibleEnabled(menu, R.id.menu_drop_caches, isHistory || containsStoredCaches(), !isEmpty);
+            setVisibleEnabled(menu, R.id.menu_delete_caches_immediately, containsStoredCaches(), !isEmpty);
             setVisibleEnabled(menu, R.id.menu_delete_events, isConcrete, !isEmpty && containsPastEvents());
             setVisibleEnabled(menu, R.id.menu_clear_offline_logs, isHistory || isOffline, !isEmpty && containsOfflineLogs());
             setVisibleEnabled(menu, R.id.menu_remove_from_history, isHistory, !isEmpty);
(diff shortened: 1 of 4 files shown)


SammysHP:
I wonder if this is really necessary because there are many ways caches are added to the database (just open them once), and also several methods to remove them. I think the original goal was to remove a cache from all lists so that it can be removed by the next garbage collection (which can always be triggered manually).

What I'd like to see is a "remove from all lists" function with undo bar.

Lineflyer:
I also don't see a real necessity for such function. Let the housekeeping do its work.
Additionally its just yet another new menu item in our already large menu.

Lineflyer:
BTW:
The usecase described in #15402 had a different demand than deleting caches permanently. Its about removing some filtered caches more easy...isnt it?

moving-bits:
Setting "do not merge" for now due to ongoing discussion in related issue

PR #16116, opened 2024-09-27
hillshading downloader

ztNFny:


main/src/main/java/cgeo/geocaching/downloader/HillshadingTileDownloader.java
```diff
@@ -0,0 +1,118 @@
+package cgeo.geocaching.downloader;
+
+import cgeo.geocaching.CgeoApplication;
+import cgeo.geocaching.R;
+import cgeo.geocaching.brouter.BRouterConstants;
+import cgeo.geocaching.brouter.mapaccess.PhysicalFile;
```
moving-bits, line 6: hillshading downloader should not have any reference to brouter classes

main/src/main/java/cgeo/geocaching/downloader/HillshadingTileDownloader.java
```diff
@@ -0,0 +1,118 @@
...
+            analyzePage(Uri.parse(url), list, page);
+        }
+        for (Download download : list) {
+            tiles.put(download.getName(), download);
+        }
+        return tiles;
+    }
+
+    @Override
+    protected boolean verifiedBeforeCopying(final String filename, final Uri file) {
```
moving-bits, line 111: This code is BRouter-specific. Probably we do not have a verification routine for downloaded hillshading tiles, so you can omit this method completely.

ztNFny:
will integrate this into https://github.com/cgeo/cgeo/pull/15415

PR #16120, opened 2024-09-28
Generalize some strings used in routing/hillshading downloads

moving-bits:
hillshading download check used some strings still mentioning "routing". This PR generalizes those string to make it working for both routing data and hillshading data


ztNFny:
thanks for catching those. You get blind after so many hours of testing stuff ...

PR #16123, opened 2024-09-28
WherigoPlayer: add things list, add caching for media and audio, add first icons and styling

eddiemuc:
WherigoPlayer: add things list, add caching for media and audio, add first icons and styling


ztNFny:
How about these for the icons? Can adjust them and add as xml if you want.

ztNFny:
How about these for the icons? Can adjust them and add as xml if you want.

eddiemuc:
Sure, I like them.

eddiemuc:
@ztNFny I will merge this PR for now, but there will be many more. Would be glad to replace icons with your new ones then

ztNFny:
> @ztNFny I will merge this PR for now, but there will be many more. Would be glad to replace icons with your new ones then

Please ping me when the UI is more or less stable. There are many buttons right now that could use a better icon, but i guess the UI won't stay as it is?

eddiemuc:
> > @ztNFny I will merge this PR for now, but there will be many more. Would be glad to replace icons with your new ones then
> 
> Please ping me when the UI is more or less stable. There are many buttons right now that could use a better icon, but i guess the UI won't stay as it is?

That's correct. Will be a while until gui stable.

PR #16125, opened 2024-09-28
fix hillshading detection in legacy map

ztNFny:
Fix broken detection for old map whether the provider supports hillshading https://github.com/cgeo/cgeo/pull/15415

(it was using the UnifiedMap TileSource previously so it worked in tests but it didn't actually work)


moving-bits:
retest this please

Issue #9967, opened 2021-02-06
Merge "map theme" and "map style" into one dialog


moving-bits:
When using OpenStreetMap I'm always struggling with "map theme" and "theme options" - where to change what etc., and probably having to open two dialogs when switching between themes.

Why not merge them into one dialog with a spinner at the top, where you can select the map theme (e. g.: "default", "Elements", "Elevate"), and below that the theme-specific options.

What do you think?

moving-bits:
Not sure if it's worth the effort, so let's close my own FR

moving-bits:
Discussion from today's dev meeting:
- reopen this issue
- move other map-relevant appearance settings into this dialog/activity (including hillshading)
- put theme-independent settings at the top, then theme selection, then theme-specific settings

This would be an alternative approach to #15847. Let's create a mockup and discuss the alternatives.

eddiemuc:
Addendums:
* Also merge "Map language" into this new dialog
* Same dialog should be useable for both GMap and OSM-based maps (this is additionally difficult of course because GMap has a very different concept for "themes" than OSM)

PR #16127, opened 2024-09-28
15847 reorg map menu

ztNFny:
Reorganize map-related settings under one tree:
image

@moving-bits there's 1 thing about this bit that's not working - the icons in menus are always white, tinting them to foreground color doesn't work. Any idea why that is?
image

depends on https://github.com/cgeo/cgeo/pull/16125
and will probably have conflicts with https://github.com/cgeo/cgeo/pull/16122 and https://github.com/cgeo/cgeo/pull/16126


moving-bits:
Not sure if I like this change in its current state. Have a look at my current map source menu:

image

Now I will have to scroll to get to one of the other actions...

But even without that (admittedly untypically) long list of map sources, this PR adds additional taps to change the theme or to change theme settings. At least those two should go one menu level up, which makes it four new entries in "Map sources" menu. Those four should be positioned at the menu's top, if we pursue this change, to make them accessible more easily.

ztNFny:
who changes map themes frequently? And even more so for theme settings... I think I've touched that like twice over many years - and that probably was to test something. Imho having them in a submenu doesn't hurt at all.

What your screenshot shows is unaffected by this PR fwiw.

moving-bits:
> who changes map themes frequently? And even more so for theme settings... I think I've touched that like twice over many years - and that probably was to test something. Imho having them in a submenu doesn't hurt at all.

different here: I do change theme settings way more frequently, e. g. changing subtheme from "city" to "cycling" etc.

> What your screenshot shows is unaffected by this PR fwiw.

what the screenshot does not show is that the new menu entries appear at the bottom, which, in case of a long map list, requires additional scrolling

ztNFny:
> what the screenshot does not show is that the new menu entries appear at the bottom, which, in case of a long map list, requires additional scrolling

Your list already requires scrolling to change maps - which going by the list of maps you frequently have to do. So you're already scrolling all the time as is.

moving-bits:
> @moving-bits there's 1 thing about this bit that's not working - the icons in menus are always white, tinting them to foreground color doesn't work. Any idea why that is?

use `app:iconTint` instead of `android:iconTint` - many parts of Material libs override the original attributes, you'll have to use the "app" prefix then to set the value

ztNFny:
We could put the appearance submenu at the top, before the list of maps?

ztNFny:
> We could put the appearance submenu at the top, before the list of maps?

I've moved it up in the rebase. Personally I feel it belongs at the bottom, but top is ok for me.

moving-bits:
> I've moved it up in the rebase. Personally I feel it belongs at the bottom, but top is ok for me if that's a consensus

Though it's still not my favorite change I guess I can get used to it (with "appearance" submenu being at the top). And I like the icons :-)

ztNFny:
And we could move the "delete offline data" back into that menu 😉

eddiemuc:
Hillshading is far too prominent IMHO. Should be moved to theme options (same level as things like 3d buildings or traffic)

ztNFny:
> Hillshading is far too prominent IMHO. Should be moved to theme options (same level as things like 3d buildings or traffic)

It's related to map appearance, why should it not be in there? It's not a theme option.

(For me HS is the only one of those settings i toggle frequently, unlike map theme or even theme options that i never change - find a good one once and stick with it forever)

ztNFny:
> Hillshading is far too prominent IMHO. 

If you don't enable the master toggle - guessing by your comment that you don't want to use it ever - the option in menu won't even show at all.

moving-bits:
> Hillshading is far too prominent IMHO. Should be moved to theme options (same level as things like 3d buildings or traffic)
> > It's related to map appearance, why should it not be in there? It's not a theme option - those are defined by the theme, not by cgeo.

For hillshading itself I would second this - it's just another map decoration feature like showing contour lines, map background etc. From a user's perspective the source of this data (defined in theme vs. coming from separate data) is not relevant, IMHO.

BUT: This menu entry here is not about hillshading in general - this is handled by our "master switch" buried in settings, so even less prominent than being in some theme options menu. This entry is VTM-only and enables an online data source, generating traffic along the way. 

Therefore following suggestion:
- rename the menu entry so that the "online character" becomes clear, e. g. "enable online hillshading layer"
- put it at the end of the online datasources section
- make it hideable (like the other data sources in UnifiedMap)

(Reason for the last item is: A user may want to use hillshading, but always stay offline. Thus hillshading with Mapsforge is fine, hillshading with VTM is not. But last item is less important compared to the other two, as you have the master switch as last resort.)

> (For me HS is the only one of those settings i toggle frequently, unlike map theme or even theme options that i never change - find a good one once and stick with it forever)

Different users, different use cases: I'm using HS rarely, only when I'm close to the Alps, which I may be once a year or so, whereas in contrast to this I'm using the other toggles quite regularly, depending on my means of transportation, the region I'm in etc.

ztNFny:
> BUT: This menu entry here is not about hillshading in general

No, it's also for offline hillshading

eddiemuc:
With the given explanations I can accept the prominent place for hillshading, provided it is ok to place also two options at the same place which I myself toggle most frequently. Those are "3D buildings" and "traffic" (for Google maps). I can provide a hidden setting too to turn this on or off.

ztNFny:
Traffic for Gmaps makes sense to me
3D buildings - is this really something that's toggled frequently? If so, ok

moving-bits:
I don't think that we'll find a common understanding which options are important enough (frequently used enough) to be placed so prominently - it will just be too different for each of us.

IMHO all those options (including hillshading) are just decoration/theming options and therefore belong into the "Theme Options" menu.

And to make this more accessible, this belongs to the main menu itself (so one level up). 

The other entries ("Select map theme" and "Map language") may be "hidden" in the sub menu or be placed at the end of the main menu level.

With this, "Map sources" menu could look like:

- Theme Options (or maybe call it "Map Appearance")
- Online map sources...
- Offline map sources...
- Select map theme
- Map language (if available for current map)
- Manage offline data

Or, if you want to group related entries, put Theme Options / Map Appearance further down below map sources:

- Online map sources...
- Offline map sources...
- Theme Options (or maybe call it "Map Appearance")
- Select map theme
- Map language (if available for current map)
- Manage offline data

eddiemuc:
> IMHO all those options (including hillshading) are just decoration/theming options and therefore belong into the "Theme Options" menu.

I understand your argument. Would be ok for me to

> With this, "Map sources" menu could look like:
> ... 
> Or, if you want to group related entries, put Theme Options / Map Appearance further down below map sources:
> ...

Or simply leave the theme related options in the main/overflow menu, similar to today?

- Theme Options (or maybe call it "Map Appearance")
- Select map theme
- Map language (if available for current map)

eddiemuc:
To make things more complex :-) an additional idea:

We could put most needed settings additionally into quick settings. Eg 4 more checkboxes for hillshading, 3d buildings and traffic. To make room we could eg remove the "compact icon" selection and remove the "other settings" headline (since the only other setting is also routing-related)

eddiemuc:
> > We could put most needed settings additionally into quick settings.
> 
> but that would be just a variant of above's discussion... - What are the "most needed" settings?

That's true but maybe we need to have that discussion. After all, the whole "quick settings" is nothing else than a collection of what we assumed (in the past) to be the "most needed" settings. So if we feel that new things become "most needed" they should be put there. And if we feel that things in there are not "most needed" they need to be removed.

If we say we don't want to make a decision what is "most needed" then consequently we have to remove the "quick settings". I assume nobody wants this (I don't :-))

ztNFny:
My idea was also to have the most important settings in appearance menu and then everything else 1 level deeper - it's just that we disagree about importance of settings. 

I do not like the idea of mixing theme options with cgeo controlled ones. Theme options are completly different for every theme, so that will be very confusing to users.

Also i don't think we should split map appearance settings in completly different places, like map menu + quick settings, but instead have them in 1 place with a structure that fits most users. And btw, this is the whole reason for thus redesign - currently map settings are spread over multiple places.

ztNFny:
> With this, "Map sources" menu could look like:
> 
> * Theme Options (or maybe call it "Map Appearance")
> * Online map sources...
> * Offline map sources...
> * Select map theme

So I asked around at 2 events to get some more opinions on that. The message from that - and that's a subjective one of course - is that "map theme" is more frequently used than those detailed options in "map options" - actually like 70% were not even aware of the later existing!

Therefore I don't think that "Theme options" should be a top-level entry, especially not the first of them.

I guess there are some power-users that love "their favorite" menu item, but that's a situation we'll encounter whenever we do anything and I don't see any other solution than fully customizable menus to make everyone happy. 
So I'd actually propose to leave the PR as is - "Appearance with Map theme and options" as the top entry. Or let's close this PR and leave everything in the messy state it's in ...

moving-bits:
If it's just this or that, either merge this PR "as is" or leave everything "as is", I would vote for the latter.

But to be more detailed:
- Having a discussion about what is a good structure for our map related items is a good thing.
- I like the more visual approach by the icons you have in this PR.
- I have no strong opinion about whether the map appearance settings (like theme options) are in one or the other entry points (Map Sources menu vs. overflow menu), one could get used to both variants.
- I'm not convinced that "hillshading toggle in map sources" is the right place - too prominent, IMHO.
- But what triggers me most is having "Theme Options" (or whatever we may call this item in the future" one menu level deeper. For me this would be way more inconvenient to change settings (and it would not help with other users getting to know those settings, either).

As our users have quite different use cases, we may not find a single set of "most used" settings. Ideally we can at least find a basic set that should be quickly accessible, but then differ on what comes next. "Hillshading" is a good example - for me this would be a setting I'm going to toggle maybe once a year, if at all, but it seems to have a bigger importance for your use cases (which is totally fine). So maybe we need a different mechanism for this, maybe some kind of user-defined "favorites". I'll open a separate discussion thread for this to have it uncoupled from a PR.

ztNFny:
> I'm not convinced that \"hillshading toggle in map sources\" is the right place - too prominent, IMHO.

Please review this PR - because HS is not in map sources but in map appearance, the same level as theme options. Having a discussion based on false assumptions makes that discussion very straining and pointless.

> If it's just this or that, either merge this PR \"as is\" or leave everything \"as is\", I would vote for the latter.

I find these discussions very hard to have as they always end in a "my favourite option must be right there at the top and i don't care whether i'm the only person in the world that's using it". As that can't be solved I don't see any way to continue here.
So I'll be happy with a nice menu in my own builds and everyone else can have an ugly one, fine with me...

ztNFny:
> But what triggers me most is having \"Theme Options\" (or whatever we may call this item in the future\" one menu level deeper. For me this would be way more inconvenient 

It's 1 additional tap every now and then. How often do you even change them?.please count the taps you do in-between in normal use of cgeo, like opening caches, scrolling, typing, ... I'd guess that tap is 1 in a million.

moving-bits:
From today's dev meeting: See #9967 for a different approach + discussion. Setting "WIP" for now.

ztNFny:
> From today's dev meeting: See #9967 for a different approach + discussion. Setting "WIP" for now.

seeing that in this PR the main argument against it was "accessing xxx will take 1 tap more today" I really wonder about that alternative solution which will require at least 2 taps more, in places spread over the screen ...

chrisdebian:
Tried to build and test this against current master (2026-08-05) but the patch no longer applies. Both files it touches, `main/src/main/java/cgeo/geocaching/maps/MapProviderFactory.java` and `main/src/main/java/cgeo/geocaching/maps/mapsforge/v6/NewMap.java`, were removed entirely by commit `5d67b264` ("Remove legacy maps") when the legacy map implementation was dropped in favor of UnifiedMap.

More importantly, the feature this PR proposes, grouping the offline-map menu entries into one submenu, was already implemented independently and merged via #16378 ("group offline-data related entries in 1 submenu"), then further refined by the fix for #18117 ("Make offline data mngt more accessible"). Current master already has a `menu_manage_offline_data` submenu doing this.

This PR looks fully superseded at this point, probably safe to close.

moving-bits:
Thanks for cross-checking, let's close here.

Issue #15775, opened 2024-06-09
Toolbar for rotation mode change


moving-bits:
Idea from today's dev meeting:

- Show compass rose permanently on the map (instead of enabling / disabling it)
- Tap on compass rose opens a small toolbar temporarily with options to
  - north-orientate the map (as today's tap action)
  - rotation modes (3 or 4, depending on map)
  - (and maybe a "fix mode" option as requested in some other issue)
- Tapping on any of those actions (or somewhere else on the map) will close the toolbar.

moving-bits:
_taken from https://github.com/cgeo/cgeo/issues/15847#issuecomment-2227273344 by @ztNFny and @moving-bits_

> > "Map rotation" needs to be way more prominent, IMHO - I see myself switching it frequently. (But that's already part of a different issue.)
> 
> Again interesting how different people are in using this stuff. I never ever used that, in any maps app. North must be up, nowhere else.

Interesting, we need to keep that in mind for #15775. Probably we should have a setting like "use map rotation". If disabled, there's no compass rose, no rotation gestures, no menu entries for rotation mode, whereas when enabled, we could talk about the stuff in #15575.

BTW: My use case for using auto-rotation is when I'm cycling with using c:geo as navigation app. Then I like to see my upcoming route without having to constantly rotate the map in my brain. Whereas when I'm standing still, or searching at some location, I like to have a non-rotating map.

PR #16137, opened 2024-09-29
UnifiedMap: Reposition "follow my location" and compass rose

### LINKED ISSUE #15774: Move "follow me" button on map from toolbar to map
Idea from today's dev meeting:

Move the "follow me" icon from the toolbar to the map area, below or beside the compass rose.

Seems to be a more logical position, and gives more space in the toolbar for the "live mode" button (which seems to be forced to the overflow menu since the list chooser changes).

moving-bits:
## Description
Tries to fix #15774 and #15775:

- position compass rose below distance windows, right-aligned; position varies with more detailed distance infos
- "follow my location" button moved from toolbar to below compass rose
- long tap on compass rose opens rotation mode quick settings
- pulls up rotation indicator handling from map fragments to `UnifiedMapActivity`
- some code cleanup

Implemented for UnifiedMap (VTM + Mapsforge + GM), CGeoMap / NewMap are unchanged.

image.image


ztNFny:
I'd suggest to not "clue" those buttons together but have them as 2, like map / route settings. Also maybe move them a bit up so vertical and horizontal distance are equal?

like
image

moving-bits:
I will look into it. (That's just some margins defined in the layout file + using a different background (without divider) for compass rose button.)

moving-bits:
Have adjusted a few settings:
- right margin of "rotation mode" / "follow my location" buttons is now identical to zoom buttons
- top margin and right margin of the two new buttons is now identical
- vertical distance between the two new buttons is now the same as horizontal distance between quick settings buttons

image

ztNFny:
Nice!
Only the "position" icon looks a bit too large, compared to the others. But of you want I can also take a look at that one later.

ztNFny:
At a quick glance that looks a lot better.

moving-bits:
As this is done just by scaling down a 38dp image without or with nearly no padding, a correctly dimensioned xml drawable would be the way to go, IMHO. So if you find the time, an update would be very welcomed.

The files are `map_followmylocation_btn.xml` and `map_followmylocation_off_btn.xml` and the scaling down is done in `map_bearingindicator.xml` layout file:

```
<ImageView
    android:id="@+id/map_followmylocation_btn"
    android:tooltipText="@string/menu_centerposition"
    android:layout_width="24dp"
    android:layout_height="24dp"
    android:layout_centerInParent="true"
    android:contentDescription="@string/menu_centerposition"
    style="@style/map_quicksettingsbutton" />
```

ztNFny:
I'll look at it, go ahead and merge. Something irks me about that compass as well, can't tell what exactly...

moving-bits:
failing tests are unrelated (connection error to GC), so let's merge here for further fine-tuning of the icons

PR #16163, opened 2024-10-04
Mapsforge: Use new HiResStandardClasyHillShading (rel. to #15415)

moving-bits:
## Description
Updates Mapsforge to latest build to include new hi resolution hillshading algorithm + parallelization while calculating hillshading data

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -371,7 +371,7 @@ dependencies {
 
     // Mapsforge Snapshot from official master branch (via https://jitpack.io/#mapsforge/mapsforge)
     String mapsforgeSource = 'com.github.mapsforge.mapsforge'
-    String mapsforgeVersion = 'f3350ef261' // master as of 2024-09-18, including rotation & fractional zoom
+    String mapsforgeVersion = '72a8a01' // master as of 2024-10-04, including rotation & fractional zoom + new hillshading algorithms
 
     // Mapsforge Snapshot from cgeo's GitHub repository (via https://jitpack.io/#cgeo/mapsforge)
     // String mapsforgeSource = 'com.github.cgeo.mapsforge'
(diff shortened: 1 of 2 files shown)


moving-bits:
#16202 should be merged first, then this PR updated to no longer include MF update - therefore setting "do not merge" label for now

moving-bits:
Rebased after having merged the Mapsforge update, thus removing the "do not merge" label

ztNFny:
I think we should eval performance impact first - and i guess then make it a setting if there's impact

ztNFny:
I've run some tests, comparing the different maps and HS engines. I manually stopped the time using a stopwatch, so in all fairness you'd probably need to substract a bit on each time - but as this is not a scientific study I think the results are good enough.

For all tests I've discarded the first run, assuming that one is caching the HS files into memory. Then I've ran the tests 10 time, discarded the top and bottom values and calculated the average of the remaining 8.

I've ran the tests with 2 map zoom levels, a region of 20*35km and 80*150km

@devemux86 Think you'll be interested in this

Unified MF
- No HS: 0,9 / 1,5
- HS Standard: 3,65 / 6,68
- HS HighRes: 3,7 / 23,75

Legacy MF
- No HS: 0,9 / 1,4
- HS Standard: 3,7 / 5,87 
- HS HighRes: 3,6 / 22,03

Summary:
- HS Standard takes about 4* as long to load as no HS
- HS HighRes for small areas is equally fast as HS Standard, but for large areas it takes *considerable* more time

Conclusion:
- HS HighRes load times are unacceptable for larger map areas
- Comparing the rendering of HS Standard vs HighRes even on a large PC screen differences are only noticable on small (zoomed in) zoom levels
- As an option we could use Standard on high and HighRes on small zoom levels. However that'd trigger a recalculation and thus a flickering on zooming, so if anything that should be an option.
- Suggestion: Let's stick with Standard

moving-bits:
Thanks for the extensive testing

> Suggestion: Let's stick with Standard

Ok for me, therefore closing here.

Issue #16164, opened 2024-10-05
Wherigo: NPE in cartridge


eddiemuc:
For the sake of testing I tried a different cartridge in the emulator: GC3XXBF ("Die Hard im Ruhrpott"). 

I started it, then moved current location a few hundred meters away from "the bomb", got the message "too far", moved back a bit, and then c:geo displayed "you have hit a bug", but seemingly continued to work. Actually it had lost its inventory, though (which makes continuing this specific cartridge impossible).

Here's the logcat:
```
10-05 00:03:02.663 10957 10957 I cgeo    : [main] WORKERTASK{wherigo-download}:post FINISHED{wherigo-download}
10-05 00:03:35.847 10957 10957 W cgeo    : [main] WHERIGOGAME: Creating engine...
10-05 00:03:35.860 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Creating state...
10-05 00:03:35.865 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Building javafunc map...
10-05 00:03:35.865 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Loading stdlib...
10-05 00:03:35.874 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: calling...
10-05 00:03:35.874 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Registering WIG libs...
10-05 00:03:35.876 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Building event queue...
10-05 00:03:35.876 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Loading gwc...
10-05 00:03:35.876 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: pre-setting properties...
10-05 00:03:35.876 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: loading code...
10-05 00:03:35.877 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: parsing...
10-05 00:03:35.890 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: calling...
10-05 00:03:35.998 10957 11831 W cgeo    : [Thread-25] WHERIGOGAME: Starting game...
10-05 00:03:35.998 10957 11831 I cgeo    : [Thread-25] WHERIGOGAME: pos: null
[...]

PR #16169, opened 2024-10-05
rel to #16164: add System.err and .out to standard logcat exerpt

eddiemuc:
rel to #16164: add System.err and .out to standard logcat exerpt

Handling #16164 I realized that we need to include also System.out and System.err output in the standard logfiles generated by c:geo. Otherwise it won't be possible to analyze and fix Wherigo-related errors


eddiemuc:
Retest this please

PR #16184, opened 2024-10-06
fix hillshading toggle for Unified Mapsforge

ztNFny:
hillshading toggle for Unified Mapsforge was still tied to pref_rapidapiKey which should be VTM only.
Fix to have the identical behaviour on all 3 maps as originally intended


main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/UserDefinedMapsforgeVTMOnlineSource.java
```diff
@@ -29,6 +29,7 @@ public class UserDefinedMapsforgeVTMOnlineSource extends AbstractMapsforgeVTMOnl
             }
             setTilePath(tilePath);
         }
+        supportsHillshading = true;
```
moving-bits, line 32: some fine-tuning: some of those "supportsHilldhading = true;" are not needed anymore as the super class already sets this value

Issue #16151, opened 2024-10-01
UM: Compass Rose - change South color and hide it if Map Rotation is off


MagpieFourtyTwo:
### Tell us your idea!

In fact I have even three things:  ;)
1. the compass rose is hard to read for people with color vision deficiency (i. e. to distinguish red North from green South),
    thus I would propose to make it more contrasting, e. g. fill the South end white or silver, with e. g. a green outline.
2. the compass rose does not really make sense if map rotation is disabled anyway,
    thus I would propose to hide it when rotation is off (which can even be understood as indicator of rotation on/off)
3. to make things easier WRT compass rose visible or not, I would swap positions of compass rose and follow me

I BTW would have put the two icons to the top left of the map, cause it would then not go up and down just because of changing number of distance fields shown ... but I missed this thread where this was discussed.  :/

Lineflyer:
Do you really mean the c:geo compass  or are you refering to the compass icon on map or something else.

MagpieFourtyTwo:
Sorry for being imprecise, I mean the small compass icon in the top right edge of the map, just above the moved follow me switch.

eddiemuc:
> 1. the compass rose is hard to read for people with color vision deficiency (i. e. to distinguish red North from green South),

Agreed. I propose to do same as Google Maps. Basically the have a much lighter silvery south arrow, mich like @MagpieFourtyTwo suggested

> 2. the compass rose does not really make sense if map rotation is disabled anyway,
>     thus I would propose to hide it when rotation is off

Agree. I just tried it, and it seems this is already the case (I tried VTM on offline map). Also when clicking on the compass rose the map goes back to "north orientation" and vanishes (also the right behavior IMHO. And almost similar to gmap :-)).

> 3. to make things easier WRT compass rose visible or not, I would swap positions of compass rose and follow me

"Follow me" is in action bar. Really put compass rose in action bar?

> I BTW would have put the two icons to the top left of the map...

On VTM Compass rose is already on top left. I am confused....

bekuno:
See here: https://github.com/cgeo/cgeo/pull/16137.

MagpieFourtyTwo:
@eddiemuc Sorry for the confusion, I'm of course talking about the latest nightlies and there it's like bekuno already referenced.

MagpieFourtyTwo:
[...]

PR #16205, opened 2024-10-12
UnifiedMap: Increase touch size for followMyLocation

moving-bits:
Also handle taps on image's background

Related to discussion in https://github.com/cgeo/cgeo/issues/16151#issuecomment-2400913210


MagpieFourtyTwo:
No change (resp. effect reverted) in 2024.10.14-NB-710b154 - the area where a click leads to a change of FollowMe is still (or again) restricted to the very center of the button.

If you e. g. hit the button at the border, although the button flickers, FollowMe does not change. You (still) must hit it perfect center to make a change.

moving-bits:
You're right, that change got lost somehow during subsequent merges. I'll create a fix for it.

PR #16207, opened 2024-10-12
Move compass rose to left, followMyLocation to zoom controls

moving-bits:
## Description
Proposal to fix the remaining issues described in #16151:
- Restore old compass rose form + position (left-aligned, directly below target/supersize info)
- Move follow my location button to above zoom control, but with some spacing

|1|2|3|
|---|---|---|
|image|image|image|


moving-bits:
failing test is unrelated (retrieve OC cache)

SammysHP:
I prefer this version to the previous one.

One additional note: I noticed that the zoom controls have a different transparency compared to the buttons at the bottom.

MagpieFourtyTwo:
Me too. It's way more functional plus it looks really good. Good work - thx!

moving-bits:
From today's dev meeting: Let's merge here, and implement new position of "follow my location" button in old map types as well (for consistency)

MagpieFourtyTwo:
Don't know if this was already changed and got lost during the position change, but the area where the FollowMe button must be hit, is still (or again) way too small (2024.10.14-NB-710b154).

For example: If you don't hit it perfect center but at the outer borders, the button flickers, but FollowMe does not change. 

Please increase the "switching area".

PR #16211, opened 2024-10-12
Wherigo: add one-time-message

eddiemuc:
Wherigo: add one-time-message

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -197,7 +197,7 @@
         <activity
             android:name=".wherigo.WherigoActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
-            android:label="Wherigo Player"
+            android:label="@string/wherigo_player"
             android:windowSoftInputMode="stateHidden" >
         </activity>
         <activity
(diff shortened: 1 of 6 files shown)


moving-bits:
Can those messages contain HTML links or Markup?
Then we could add a link that links directly to the specific setting (using `cgeo-setting://prefkey` syntax, IIRC)

eddiemuc:
> Can those messages contain HTML links or Markup? Then we could add a link that links directly to the specific setting (using `cgeo-setting://prefkey` syntax, IIRC)

As far as I can see OneTimemessage allows adding one link to description. I quickly looked into this but couldn't see an easy way to link to the Appearance page or any of the two relevant settings (custom menu entry or quick link item). @moving-bits maybe something to add at a later point in time?

moving-bits:
c:geo allows to deep-link settings: For quick launch buttons this is `<a href="cgeo-setting://quicklaunchitems_sorted">...</a>`, and for custom menu button, it's `<a href="cgeo-setting://custombnitem">...</a>`. Haven't tested yet, though, whether it works with "advanced settings" disabled.

eddiemuc:
> c:geo allows to deep-link settings: For quick launch buttons this is `<a href="cgeo-setting://quicklaunchitems_sorted">...</a>`, and for custom menu button, it's `<a href="cgeo-setting://custombnitem">...</a>`. Haven't tested yet, though, whether it works with "advanced settings" disabled.

I tried it. It links you to the "Appearance" settings page all right and scrolled to the right place (Quick Launch Items). When advanvced settings are disabled it still opens "Appearance" settings for you, but naturally you can't see the setting...

If those settings become so important in context of Wherigo, maybe we should make them "non-advanced" ones?

PR #16213, opened 2024-10-13
Wherigo: internationalize / localize texts

eddiemuc:
Wherigo: internationalize / localize texts


eddiemuc:
No change of logic, just translation. Thus merging directly.

PR #16222, opened 2024-10-16
WIG UI

ztNFny:
various changes to the WIG player UI, mainly icon changes

@eddiemuc I also added a "load" button next to save for storing/load a savegame. It's non-functional for now though - didn't bind it.
Please review and drop some commits if you don't like them

image


eddiemuc:
@ztNFny thanks a lot for diving into this! I really like the icons and are in favor of merging.

Just one thing about the load button: the reason I didn't place one there is that in order to load a game one has to choose a cartridge first (at least that was my thinking). So I integrated loading a game into the cartridge details page which can be accessed with the leftmost button. 

What would you expect to happen when user taps the "load" button?

ztNFny:
@eddiemuc I had some more thoughts about the WIG UI, putting them here for discussion (as you're probably the only one looking into it at the moment anyway):

When opening the WIG view without a catridge loaded the view could instead of an empty view show a list of all the offline cartridges. Also the icons that are only relevant when a WIG is loaded should be hidden.
image

Media is displayed with full width, yet most of the images in WIGs are very small (iirc there's a max size in WIGs), causing the images to get blown up and get blurry. So a) it doesn't look nice because it's blurry, b) the image is huge in comparison to the text - and depending on the image dimensions you might even have to scroll a long way to get to the text.
I'd suggest to add a max width and height to the media view.
image

ztNFny:
> Just one thing about the load button: the reason I didn't place one there is that in order to load a game one has to choose a cartridge first (at least that was my thinking). So I integrated loading a game into the cartridge details page which can be accessed with the leftmost button.
> 
> What would you expect to happen when user taps the "load" button?

That whole row of icons is only relevant when the user has already opened a cartridge. The "load" button should show the 2nd step of the "Play"-dialog - a list of the savegames available.
Usecase: I'm playing a WIG, save regular, then screw something up and want to go back to the last good state - exactly what savegames are for ;-) Right now I need to close and reopen the cartridge which seems cumbersome. 

(btw - maybe we could also find a better solution for save slots? There are now 4 hardcoded slots, something I guess taken over from Wheryougo, but it doesn't make too much sense)

eddiemuc:
> > What would you expect to happen when user taps the "load" button?
> 
> The "load" button should show the 2nd step of the "Play"-dialog - a list of the savegames available.

Understood. Makes a lot of sense
 
> (btw - maybe we could also find a better solution for save slots? There are now 4 hardcoded slots, something I guess taken over from Wheryougo, but it doesn't make too much sense)

Yes that's a takeover of WhereYouGo (just like the "big media" display). I am open for other solutions.

I will merge this PR now, thanks a lot for the work!

ztNFny:
btw, feel free to merge and revert the save button if you think it doesn't make sense (at this point in time)

eddiemuc (approved):

ztNFny:
> Yes that's a takeover of WhereYouGo (just like the "big media" display). I am open for other solutions.

i'd suggest a list of existings save games (to be able to overwrite them) + a single "new savegame" item. Then the user could have as many savegames as they want (instead of a hard upper limit) and not be confused by having to select between 4 identical empty slots for the first one.

Issue #16212, opened 2024-10-12
Transparency of map buttons


moving-bits:
One additional note: I noticed that the zoom controls have a different transparency compared to the buttons at the bottom.

_Originally posted by @SammysHP in https://github.com/cgeo/cgeo/issues/16207#issuecomment-2408566691_

ztNFny:
The resource files are using the same color definitions though, but in different ways.

Quick settings is ImageButton with @style/map_quicksettingsbutton which uses @drawable/white_bcg which uses @color/osm_zoomcontrolbackground
Follow locationis ImageButton with @style/map_quicksettingsbutton which uses @drawable/white_bcg which uses @color/osm_zoomcontrolbackground
Zoom button is a path with fillColor @color/osm_zoomcontrolbackground

Seeing as Quick settings and location are defined in the same way, how are they different??!

SammysHP:
Maybe differences in a parent container?

Btw, a few more nitpicks:

- Location button uses slightly different foreground color than the zoom controls
- location button misses the rounded corners at the top

![Screenshot_2024-10-15-06-17-07-452_cgeo geocaching](https://github.com/user-attachments/assets/c93da3cf-b3be-47d6-9a7a-59ffcb7fda7e)

ztNFny:
> * location button misses the rounded corners at the top

This is because after the latest change there's now 2 ImageButtons on top of each other - I guess a workaround to increase the hit area - I think we should look for a better solution

PR #16225, opened 2024-10-16
finetune map button layout

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/16212

https://github.com/cgeo/cgeo/issues/16212

### DIFF
--- a/main/src/main/res/layout/map_zoom_control.xml
+++ b/main/src/main/res/layout/map_zoom_control.xml
@@ -7,7 +7,7 @@
         android:id="@+id/container_followmylocation"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:layout_above="@id/container_zoomin"
+        android:layout_above="@id/container_zoom"
         android:layout_alignParentRight="true"
         android:layout_marginRight="13dp"
         android:layout_marginBottom="6dp">
@@ -20,21 +20,8 @@
     </RelativeLayout>
 
     <LinearLayout
-        android:id="@+id/container_zoomin"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_alignParentRight="true"
-        android:layout_above="@id/container_zoomout"
-        android:layout_centerHorizontal="true"
-        android:layout_marginRight="13dp">
-        <ImageButton
-            android:id="@+id/map_zoomin"
-            android:src="@drawable/map_zoomin"
-            android:contentDescription="@string/zoom_in"
-            style="@style/map_zoombutton" />
-    </LinearLayout>
-    <LinearLayout
-        android:id="@+id/container_zoomout"
+        android:id="@+id/container_zoom"
+        android:orientation="vertical"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_alignParentRight="true"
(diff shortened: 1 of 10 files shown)


ztNFny:
retest this please

main/src/main/res/layout/map_zoom_control.xml
```diff
@@ -7,52 +7,42 @@
...
             android:id="@+id/map_followmylocation_btn"
-            android:tooltipText="@string/menu_centerposition"
-            android:layout_width="24dp"
-            android:layout_height="24dp"
-            android:layout_centerInParent="true"
-            android:contentDescription="@string/menu_centerposition"
-            style="@style/map_quicksettingsbutton" />
+            android:src="@drawable/map_quick_settings"
+            style="@style/map_quicksettingsbutton"
+            android:tooltipText="@string/quick_settings"
```
moving-bits, line 18: tooltipText and contentDescription should point to "follow my position", currently a long-tap on this icon would show a "quick settings" tooltip

main/src/main/res/values/styles.xml
```diff
@@ -531,6 +531,16 @@
...
     </style>
+    <style name="map_zoombutton">
+        <item name="android:layout_width">38dp</item>
+        <item name="android:layout_height">38dp</item>
+        <item name="android:gravity">bottom</item>
+        <item name="android:insetTop">0dp</item>
+        <item name="android:insetBottom">0dp</item>
+        <item name="android:padding">8dp</item>
+        <item name="android:background">@drawable/mark_transparent</item>
+        <!--<item name="android:tint">@color/osm_zoomcontrol</item>-->
```
moving-bits, line 542: can be removed (as well as in style above)

moving-bits (changes requested):
Thanks for diving into this, looks much cleaner now!
Have only two small nitpicks, see comments. Thanks!

moving-bits:
One additional question:

Without this PR, the map overlay buttons (zoom controls, quick settings, follow my location etc.) have different levels of transparency/opaqueness:

image

As you can see in the screenshot, zoom buttons are a bit more opaque than map quick settings and route/track quick settings.

With this PR, all buttons will have the higher transparency of those buttons (= less opaqueness), which makes it harder to read the button depending on what beneath them (like the motorway in those screenshots):

image

Can you adjust all map overlay buttons to have the same transparency, based on the more opaque version of the zoom controls in the first screenshot? Thanks.

ztNFny:
The different opaqueness levels were a bug 😉

moving-bits:
> The different opaqueness levels were a bug 😉 Zoom buttons had 2 half-transparent layers on top of each other

Yeah, I know, and it's good that it will be fixed by this PR. Still, the resulting opaqueness of this double-layering was better than the slight opaqueness of the other buttons. So how can we increase the opaqueness without double-layering of images?

SammysHP:
I prefer this version. The buttons can be seen sufficiently and I am still able to guess what's behind them.

ztNFny:
I've adjusted opacity to be between the 2 old ones, don't think anyone would notice the change unless in a direct comparison

ztNFny (commented):
I've adjusted opacity to be between the 2 old ones, don't think anyone would notice the change unless in a direct comparison

PR #16189, opened 2024-10-07
Fetch cache-data for unsaved caches without merging data from cache cache (fix #16175)

murggel:
## Description

Fetch cache-data for unsaved caches without merging from already downloaded cache-data, so the newest version is used
For stored caches merging / gathering missing data from existing cache is performed, so no data will be lost (preserve current behaviour)

## Related issues

fix #16175 

## Additional context

PR #16226 is prerequisite for thia PR to fix failing test (Fix the date-format in GC40)

PR #16226, opened 2024-10-16
Fix the date-format in the html-page of GC40

murggel:
## Description

Fix the date-format in the html-page of `GC40`

With the subsequent commits I tried to reproduce / understand the behaviour of the failing test `testParseCacheFromTextWithMockedData` in the case, it was executed solely. So it seemed, that the culprit is that previous tests were executed and the cache was then already in the database.

## Related issues

prerequisite for PR #16189


bekuno (approved):
Ok, lets format it like the other test data.

PR #16230, opened 2024-10-17
manually add event date + start/end time to listing text

### LINKED ISSUE #16229: GC website change: Event time not in listing
GC events no longer include the event date and time at the top of the listing. The info is still available below the cache name.

This means that nowhere in c:geo the start and end times of an event are visible.
This also breaks event date recognition for adding events to the calendar.

ztNFny:
GC removed the event date + start / end times from listings. Because of this it's no longer possible to see the event times anywhere in c:geo and functions relying on these dates, like creating calendar entries, are failing.

With this PR the event times are manually prefixed to the events description, making it appear and function exactly like before the change.

fixes #16229

Targeted against release as this is a breaking change.


eddiemuc (approved):
Lgtm. Will only work for English, but afair we switch GC.com locale to English anyway?

Thanks for the quick fix, will merge due to criticality.

ztNFny:
> Lgtm. Will only work for English, but afair we switch GC.com locale to English anyway?
> 
> Thanks for the quick fix, will merge due to criticality.

It's not English only, the AM/PM is optional.

fm-sys:
> afair we switch GC.com locale to English anyway?

Not anymore since around 2 years or so ;)

Issue #16214, opened 2024-10-13
UnifiedMap: Make Mapsforge default, VTM "on demand"


moving-bits:
As discussed in today's dev meeting:

- swap defaults: make Mapsforge the default for UnifiedMap and VTM only visible "on demand"
- check for missing Mapsforge map types (eg: combined map?)

fm-sys:
Features (potentially) missing in Mapsforge: 7ffd55603277dde31d800a877e5f063d56f693a3 6b43cbbfa02d4e35400f31b720b5367dcebb29c2 acbe6d610e78aac721163699e1f1638e6946e623 2343b705fd7b79e3793c2534b5f33accaa15bc05

ztNFny:
During my tests of Unified Mapsforge I've experienced it to be very slow in comparison to both old Mapsforge and VTM. In fact in emulator it never successfully loaded Germany (OAM) whereas the other 2 implementations have no issue with that.

MagpieFourtyTwo:
With Map Source set to Mapsforge, "Combined (offline)" got lost from the map selection drop down in map view this morning, most probably after updating to latest Nightly - why that?

In System Setting "Map Sources" => "Hide Map Sources" it was still present (when Source is set to Mapsforge), but (of course) not ticked - but after a serving restart of c:geo it disappeared over there, too.

If I change Map Source to VTM, everything is fine, it appears in the drop down, and in the list of "Hide Map Sources" (although the latter requires a restart of c:geo, otherwise it will not show up in hideable sources).

Has "Combined (offline)" been removed on purpose for Mapsforge or just accidentally? (I hope it's the later, cause this would be a real showstopper for us)

ztNFny:
> Has "Combined (offline)" been removed on purpose for Mapsforge or just accidentally? (I hope it's the later, cause this would be a real showstopper for us)

https://github.com/cgeo/cgeo/pull/16235

MagpieFourtyTwo:
Aaah - thx for the hint!

PR #16233, opened 2024-10-18
UnifiedMap: Make Mapsforge default implementation (rel. to #16214)

moving-bits:
## Description
VTM is still available, use switch in Settings => Map Sources to enable display of both VTM and Mapsforge in UnifiedMap

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -112,6 +112,10 @@ public class Settings {
     public static final int CUSTOMBNITEM_NONE = -1;
     public static final int CUSTOMBNITEM_NEARBY = 0;
 
+    public static final int UNIFIEDMAP_VARIANT_MAPSFORGE = 1;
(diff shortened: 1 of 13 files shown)


ztNFny:
Would you consider changing the "show both" option to a "show other"? Or maybe we should just add a global map type switch with options Legacy, Unified Mapsforge, Unified VTM (and maybe add a note to the later that it's experimental)?

ztNFny:
Oh, I was thinking about a 4 way toggle then - to include also the option of the legacy map instead of having a unified On/Off + a unified source selection.
But actually I'm not really sure on the state of "old" map vs unified, considering that I (as well as most devs I gues) have been running Unified for so long. I think it's time to make Unified the default - but I guess that's a seperate discussion, maybe for next call. So sorry for bringing it up off-topic here ...

moving-bits:
I left that out intentionally, as we have settings depending on "Use UnifiedMap yes/no" - at least the built-in automatic toggling of dependent settings requires boolean values AFAIK.

PR #16235, opened 2024-10-18
UnifiedMap Mapsforge: Support multi-offline map (rel. to #16214)

moving-bits:


moving-bits:
I'm facing a couple of crashes when using combined maps on UnifiedMap Mapsforge. Needs further testing, therefore setting WIP for now.

moving-bits:
Have found the cause for the crashes and updated the PR accordingly, therefore removing the "WIP" label.

PR #16243, opened 2024-10-19
align compass design to other map elements

### LINKED ISSUE #16151: UM: Compass Rose - change South color and hide it if Map Rotation is off

ztNFny:
adjust the design of compass rose on map to be in line with the other map overlays (quick settings, zoom, follow location).
It's now using the same background color, transparency, border style (none) and size (8% larger)

Also I've made the south arrow lighter (999 -> bbb) to give it more contrast vs north which fixes #16151

Old:
image
New:
image


ztNFny:
retest this please

moving-bits:
I fear the same issues as described for the original PR (#13063): Removing the border makes it harder to distinguish the compass rose from the rest of the map, depending on what's currently beneath it.

But as our map buttons have a bit more opacity meanwhile, let's give it a try.
And having a consistent visual appearance is definitely a good thing! Thanks for diving into it.

PR #16251, opened 2024-10-20
Fix zoom controls for old Mapsforge map (rel. to #16225)

moving-bits:
## Description
Fixes separator between zoom controls of old Mapsforge map:

|before|after|
|---|---|
|image|image|


moving-bits:
CI seems to be broken, "Could not create ADB Bridge."

Can someone with access to the service restart the executor - maybe @bekuno @mucek4?

bekuno:
I can't.
I wrote now a mail to @mucek4 .

moving-bits:
retest this please

PR #16253, opened 2024-10-20
fix #16220: Wherigo: add option to delete savefiles on save and load

### LINKED ISSUE #16220: Wherigo Player: The possibility to select what to delete
### Tell us your idea!

At the moment, in Wherigo Player, when you select the delete button, there is only the option to delete the cartridge (the cartridge name does not appear) along with the existing savegames.

It would be good to have the option to choose what to delete: everything, only the savegames, or a specific savegame.

If feasible, an example would be to use checkboxes to select what should be deleted.

If I recall correctly the autosave always replace the previous autosave with the new one, so it would be 1 check box to the autosave if present + 4 checkboxes for the saves if they exist, 1 for the cartridge, and 1 for all.

Screenshot_2024-10-15-22-12-31-915_cgeo geocaching

### Is there an alternative?

Keep as it is

### Additional Information

_No response_

eddiemuc:
fix #16220: Wherigo: add option to delete savefiles on save and load


bekuno:
Retest this please

PR #16256, opened 2024-10-21
Remove filled final from corrected coordinates filter

### LINKED ISSUE #16247: Add a filter for "original waypoint" (don't consider filled final waypoint)
### Tell us your idea!
I use "set as corrected coordinates" for those mysteries, where I already verified the final position, otherwise I am not quite sure, if the position of the final waypoint is correct.
With current implementation of the filter I cannot distinguish those mysteries.

I would like to have a filter "has original coordinate" for mystery, which does not consider any filled final waypoint for mysteries.

### Is there an alternative?
No

The filter "corrected coordinates" counts mysteries also, if they have a valid final waypoint set (without setting it as "corrected coordinates").

### Additional Information

related to #11940

murggel:
## Description

Remove check for filled final waypoint from filter `corrected coordinates`
Introduce method `createOriginalWaypoint` to avoid duplicate code

## Related issues

fix #16247

### DIFF
--- a/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
@@ -758,10 +758,7 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
                         cache.resetUserModifiedCoords(waypoint);
                     } else {
                         if (!cache.hasUserModifiedCoords()) {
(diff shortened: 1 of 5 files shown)


eddiemuc (approved):

eddiemuc:
Lgtm. Also contains some refactorings which I like. Thanks for contributing!

PR #16257, opened 2024-10-21
Update finalDefined for calculated waypoint

murggel:
## Description

Update the member `finalDefined` after recalculation of waypoint
Store it in Database

## Related issues

rel #16247

### DIFF
--- a/main/src/main/java/cgeo/geocaching/models/Geocache.java
+++ b/main/src/main/java/cgeo/geocaching/models/Geocache.java
@@ -1355,13 +1355,7 @@ public class Geocache implements IWaypoint {
             }
         }
         resetFinalDefined();
-
-        boolean savedToDatabase = false;
-        if (saveToDatabase) {
-            DataStore.saveFinalDefinedStatus(this);
-            savedToDatabase = DataStore.saveWaypoints(this);
-        }
-        return savedToDatabase;
+        return saveToDatabase && DataStore.saveWaypoints(this);
     }
 
     /**
@@ -1594,13 +1588,7 @@ public class Geocache implements IWaypoint {
             // when waypoint was edited, finalDefined may have changed
             resetFinalDefined();
         }
-
-        boolean savedToDatabase = false;
-        if (saveToDatabase) {
-            DataStore.saveFinalDefinedStatus(this);
-            savedToDatabase = DataStore.saveWaypoint(waypoint.getId(), geocode, waypoint);
-        }
-        return savedToDatabase;
+        return saveToDatabase && DataStore.saveWaypoint(waypoint.getId(), geocode, waypoint);
     }
 
     private void assignUniquePrefix(final Waypoint waypoint) {
@@ -1883,7 +1871,6 @@ public class Geocache implements IWaypoint {
         }
         if (hasCalculatedWp) {
             resetFinalDefined();
-            DataStore.saveFinalDefinedStatus(this);
             DataStore.saveWaypoints(this);
         }
         return hasCalculatedWp;
--- a/main/src/main/java/cgeo/geocaching/models/Geocache.java
+++ b/main/src/main/java/cgeo/geocaching/models/Geocache.java
@@ -1355,7 +1355,13 @@ public class Geocache implements IWaypoint {
             }
         }
         resetFinalDefined();
-        return saveToDatabase && DataStore.saveWaypoints(this);
+
+        boolean savedToDatabase = false;
+        if (saveToDatabase) {
+            DataStore.saveFinalDefinedStatus(this);
+            savedToDatabase = DataStore.saveWaypoints(this);
+        }
+        return savedToDatabase;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


murggel:
I am undecided whether to save the flag to the database in directly inside the method `resetFinalDefined` (and rename it then to resetAndSaveFinalDefined)...

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -2339,6 +2339,31 @@ public static void saveUserModifiedCoords(final Geocache cache) {
...
 
+    public static boolean saveFinalDefinedStatus(final Geocache cache) {
+        cache.addStorageLocation(StorageLocation.DATABASE);
+        cacheCache.putCacheInCache(cache);
+        Log.d("Updating finalDefined of " + cache + " in DB");
+
+        final ContentValues values = new ContentValues();
+        values.put("finalDefined", cache.hasFinalDefined() ? 1 : 0);
+
+        init();
```
eddiemuc, line 2350: Surround all database access with a call to `withAccessLock` (see how other methods in Data store are doing it)

main/src/main/java/cgeo/geocaching/models/Geocache.java
```diff
@@ -1859,10 +1868,26 @@ public void changeVariables(final Map<String, String> newVars) {
...
+
+    public boolean recalculateWaypoints(final CacheVariableList variableList) {
+        boolean hasCalculatedWp = false;
+        for (Waypoint wp : getWaypoints()) {
+            hasCalculatedWp |= wp.recalculateVariableDependentValues(variableList);
+        }
+        if (hasCalculatedWp) {
+            resetFinalDefined();
+            DataStore.saveFinalDefinedStatus(this);
+            DataStore.saveWaypoints(this);
```
eddiemuc, line 1887: Maybe `saveFinalDefined` logic should be included in `saveWaypoints` method?

main/src/main/java/cgeo/geocaching/models/Geocache.java
```diff
@@ -1591,7 +1594,13 @@ public boolean addOrChangeWaypoint(final Waypoint waypoint, final boolean saveTo
             // when waypoint was edited, finalDefined may have changed
             resetFinalDefined();
         }
-        return saveToDatabase && DataStore.saveWaypoint(waypoint.getId(), geocode, waypoint);
+
+        boolean savedToDatabase = false;
+        if (saveToDatabase) {
+            DataStore.saveFinalDefinedStatus(this);
+            savedToDatabase = DataStore.saveWaypoint(waypoint.getId(), geocode, waypoint);
```
eddiemuc, line 1601: Maybe `savefinaldefinedstatus` should be called from inside `savewaypoint`?

eddiemuc (changes requested):

eddiemuc:
One mandatory remark (use withccesslock in Data store), the others are optional. Thanks for the contribution!

eddiemuc:
> I am undecided whether to save the flag to the database in directly inside the method `resetFinalDefined` (and rename it then to resetAndSaveFinalDefined)...

I think both would be ok

main/src/main/java/cgeo/geocaching/models/Geocache.java
murggel: Would be possible, but then I need to pass the cache / finalDefined...
If that would be fine...

eddiemuc: Isn't the check for finaldefined necessary for every "save waypoint" call anyway? Because every change of any waypoint ma lead to a change of this flag

eddiemuc: Eg if someone changes the type of a waypoint from "final" to "puzzle" or vice versa

murggel: I think, this is covered by addOrChangrWaypoint
But I would like to have it together in one method to avoid discrepancy

main/src/main/java/cgeo/geocaching/storage/DataStore.java
murggel: Merged the new function into saveWaypoint

main/src/main/java/cgeo/geocaching/models/Geocache.java
murggel: Merged the new function into saveWaypoint

main/src/main/java/cgeo/geocaching/models/Geocache.java
murggel: Merged the new function into saveWaypoint

murggel:
> > I am undecided whether to save the flag to the database in directly inside the method `resetFinalDefined` (and rename it then to resetAndSaveFinalDefined)...
> 
> I think both would be ok
Merged the new function into saveWaypoint

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -5315,8 +5324,13 @@ public static SearchResult getBatchOfStoredCaches(final Geopoint coords, final i
     }
 
     public static boolean saveWaypoint(final int id, final String geocode, final Waypoint waypoint) {
+        // parent could be lazy loaded, so not inside access lock
```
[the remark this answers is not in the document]
eddiemuc: I assume the getparentcache method uses a cache load method from datastore which is acquiring a lock first (in this case a read lock), so it is correct to do this outside the accesslock

eddiemuc (approved):
Lgtm. Thanks for diving into this

PR #16259, opened 2024-10-22
Clear filter should be possible, if no filter is defined

murggel:
## Description

If a filter is currently active, but no filter is saved (only "temporary" filter), the long click on the filter-icon does nothing. At least I should have the possibility to clear the current active filter.

first proposal

second proposal

## Additional context

I reused the current dialog-box and just kept the "clear" button, that is, why it looks probably weird.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
@@ -1,6 +1,7 @@
 package cgeo.geocaching.utils;
 
 import cgeo.geocaching.R;
+import cgeo.geocaching.activity.AbstractActivity;
 import cgeo.geocaching.activity.FilteredActivity;
 import cgeo.geocaching.filters.core.GeocacheFilter;
 import cgeo.geocaching.filters.core.GeocacheFilterContext;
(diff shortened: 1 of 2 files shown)


fm-sys:
Thanks for the PR, really useful! 

One remark though: I believe it would be better to create a separate dialog in code for this edge case. A user which never had stored any filters would probably be pretty confused about the useless search icon.

murggel:
> Thanks for the PR, really useful! 
> 
> One remark though: I believe it would be better to create a separate dialog in code for this edge case. A user which never had stored any filters would probably be pretty confused about the useless search icon. 

Reasonable, I will check, if we can hide the search icon, if no items are selectable...

fm-sys:
> will check, if we can hide the search icon, if no items are selectable

IMHO there is no benefit in making things more complicated than they need to be. Just create an independent dialog for this case and we should be fine :)

murggel:
I added a toast, if no filters are stored yet.
Should the toast also be shown, if a filter is active and the new dialog appears or just when nothing happens?

fm-sys:
Do we need that toast? Maybe just set the "No filter are stored yet." message as dialog content indicating that saved filters would appear there.

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -38,29 +39,40 @@ public static void openFilterActivity(final Activity activity, final GeocacheFil
...
         final List<GeocacheFilter> filters = new ArrayList<>(GeocacheFilter.Storage.getStoredFilters());
+        final boolean isFilterActive = filterContext.get().isFiltering();
 
         if (filters.isEmpty()) {
+            filteredActivity.showToast(R.string.cache_filter_storage_load_delete_nofilter_message);
+
+            if (isFilterActive) {
+                SimpleDialog.of(filteredActivity).setTitle(R.string.cache_filter_storage_clear_title)
+                        .setPositiveButton(null)
+                        .setNeutralButton(TextParam.id(R.string.cache_filter_storage_clear_button))
```
moving-bits, line 52: Is there a specific reason why you chose the neutral button? To my understanding a single button should be right-aligned (normally the "OK" button, but that can have a different label). Would read something like this:
```
SimpleDialog.of(filteredActivity).setTitle(R.string.cache_filter_storage_clear_title)
        .setPositiveButton(TextParam.id(R.string.cache_filter_storage_clear_button))
        .show(() -> filteredActivity.refreshWithFilter(GeocacheFilter.createEmpty(filterContext.get().isOpenInAdvancedMode())));
```

murggel: This was on purpose to have it on the same location as in the other regular dialog to avoid switching places for same function

murggel:
> Do we need that toast? Maybe just set the "No filter are stored yet." message as dialog content indicating that saved filters would appear there.

I wanted to avoid a separat click

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
moving-bits: Which other regular dialog do you mean - selecting a filter? (Although it looks similar, this would be something different - select a value from a list of one or more entries, without using an action button, versus selecting "Clear" as action button.)

murggel: The dialog on long click on the filter button to select the filter from the list

moving-bits:
> > Do we need that toast? Maybe just set the "No filter are stored yet." message as dialog content indicating that saved filters would appear there.
> 
> I wanted to avoid a separat click
> 
> Or do you mean as content for the dialog with the clear Button for the filter?

I've probably stumbled about the same: Long-tap on filter with no stored filters opens a dialog and displays a toast at the same time, which is pretty confusing, IMHO. I would put the "no filters stored yet" as message into the dialog (which at the same time would make the dialog look less empty).

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
murggel: But it is the same click, where I expect the other dialog...

murggel:
Ok I will provide a screenshot tomorrow

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
moving-bits: Ok, thanks - I'd seen only the variant without the "clear" button:

image

(Seems to depend on having a filter stored, but using a different one.)

IMHO the label should be right-aligned (using "positive" button) in both cases for consistency with material design. Although I agree that we have some aligning issue in the dialog - indentation of dialog heading is different from content, but that's a separate issue.

murggel: > Ok, thanks - I'd seen only the variant without the "clear" button:
> 

The clear button appears, if you have a filter active

PR #16260, opened 2024-10-23
Add separator for VTM-offline-maps

murggel:
## Description

Consider `AbstractMapsforgeVTMOfflineTileProvider` and `MapsforgeVTMMultiOfflineTileProvider` also for separated offline-maps

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/TileProviderFactory.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/TileProviderFactory.java
@@ -57,7 +57,10 @@ public class TileProviderFactory {
             }
             if (!hide) {
                 final int id = tileProvider.getNumericalId();
-                parentMenu.add(tileProvider instanceof AbstractMapsforgeOfflineTileProvider ? R.id.menu_group_map_sources_offline : R.id.menu_group_map_sources_online, id, i, tileProvider.getTileProviderName()).setCheckable(true).setChecked(id == currentTileProvider);
+                final boolean isOfflineMap = (tileProvider instanceof AbstractMapsforgeOfflineTileProvider)
+                        || (tileProvider instanceof AbstractMapsforgeVTMOfflineTileProvider)
+                        || (tileProvider instanceof MapsforgeVTMMultiOfflineTileProvider);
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/TileProviderFactory.java
```diff
@@ -57,7 +57,10 @@ public static void addMapviewMenuItems(final Activity activity, final PopupMenu
             }
             if (!hide) {
                 final int id = tileProvider.getNumericalId();
-                parentMenu.add(tileProvider instanceof AbstractMapsforgeOfflineTileProvider ? R.id.menu_group_map_sources_offline : R.id.menu_group_map_sources_online, id, i, tileProvider.getTileProviderName()).setCheckable(true).setChecked(id == currentTileProvider);
+                final boolean isOfflineMap = (tileProvider instanceof AbstractMapsforgeOfflineTileProvider)
+                        || (tileProvider instanceof AbstractMapsforgeVTMOfflineTileProvider)
+                        || (tileProvider instanceof MapsforgeVTMMultiOfflineTileProvider);
```
moving-bits, line 62: the last part (`tileprovider instanceof MapsforgeVTMMultiOfflineTileProvider`) is not required, as a `MapsforgeVTMMultiOfflineTileProvider` is a subclass of `MapsforgeVTMOfflineTileProvider`

moving-bits (changes requested):
Thanks for diving into this issue, good idea!

PR generally LGTM, just one small finding, see comment.

main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/TileProviderFactory.java
murggel: Removed

murggel:
Is it on purpose, that the `OpenStreetMap` has a different name `OSM `in unifiedMap?

moving-bits:
> Is it on purpose, that the `OpenStreetMap` has a different name `OSM `in unifiedMap?

That's probably just been a shortcut back then when I started implementing UnifiedMap. Should better be the same as in the "old" maps, and probably preferably the long form to get the connection to the related web services.

murggel:
> > Is it on purpose, that the `OpenStreetMap` has a different name `OSM `in unifiedMap?
> 
> That's probably just been a shortcut back then when I started implementing UnifiedMap. Should better be the same as in the "old" maps, and probably preferably the long form to get the connection to the related web services.


I will provide a PR for that...

moving-bits (approved):
LGTM, thanks for the contribution

PR #16263, opened 2024-10-25
Use resource-string for map-names

murggel:
## Description

Use resource-string for map-names in unified map

## Related issues

https://github.com/cgeo/cgeo/pull/16260#issuecomment-2433385979


moving-bits (approved):
LGTM, thanks for fixing!

Only thing that could be discussed is to mark the two hillshading-related strings as translatable (to translate "hillshading"), but this would be optional, IMHO.

PR #16274, opened 2024-10-27
fix #16238: add wherigo actionbar to most important activities when wherigo is running

### LINKED ISSUE #16238: Wherigo: add "quick access actionbar" to all activities while wherigo is in play
This is an idea which came up in some discussion about Wherigo Player which I want to keep track on.

Idea is to display an "action bar" below the real action bar on all c;:geo activities in case a Wherigo is currently playing. Placement shall be like the "filter bar" in map view/list view, UX view to be discussed. Tapping on this "wherigo bar" would bring the player to the Wherigo activity and/or open any pending dialogs.

eddiemuc:
fix #16238: add wherigo actionbar to most important activities when wherigo is running

image

image

### DIFF
--- a/main/src/main/res/layout/activity_navigationbar.xml
+++ b/main/src/main/res/layout/activity_navigationbar.xml
@@ -5,6 +5,12 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
+    <cgeo.geocaching.wherigo.WherigoInfoBarView
+        android:id="@+id/wherigo_bar_view"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_alignParentTop="true"/>
+
     <ImageView
         android:id="@+id/background"
         android:layout_width="match_parent"
@@ -15,7 +21,7 @@
         android:id="@+id/activity_content"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:layout_alignParentTop="true"
+        android:layout_below="@id/wherigo_bar_view"
         android:layout_above="@id/activity_navigationBar" />
 
     <com.google.android.material.bottomnavigation.BottomNavigationView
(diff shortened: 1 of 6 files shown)


fm-sys:
I'm wondering whether this implementation should better go into `AbstractNavigationBarActivity.java` / `activity_navigationbar.xml
` instead of the map/list specific layout files?

eddiemuc:
> I'm wondering whether this implementation should better go into `AbstractNavigationBarActivity.java` / `activity_navigationbar.xml
> ` instead of the map/list specific layout files? 

Can be argued. I decided the current way for following reasons:
- as less mixing of wherigo code and "non wherigo" code as possible (team wanted a clean-as-possible solution)
- less complex (a view component can take care of its updates internally while this has to be triggered externally in navigation bar)
- is it ensured 100% that now and in the foreseeable future the set of activities with nav bar and wherigo bar will be the same?

I basically followed the "guide" on how filter-sort-bar was implemented, I just encapsulated the update algo by using a view instead of a layout.

How to proceed here?

eddiemuc:
Updated the list above (basically about the question whether wherigo bar should be shown below or above filter bar if both are present)

fm-sys:
I really like the idea to implement it as view component. This and adding it to `activity_navigationbar.xml` shouldn't conflict each other.

> Update: If we tie wherigo bar to nav bar, then it will be displayed on top of the filter bar in case both is shown (in map and list view). Is this wanted?

I would consider this acceptable or even preferred. The filter bar is tightly coupled with the activity content, so it's better to show it below (thus closer to the activity content). That may be a personal preference of course.

> is it ensured 100% that now and in the foreseeable future the set of activities with nav bar and wherigo bar will be the same?

At least that would be consistent. If implemented as view component it could be added to all bottom navigation activities + a potential white-list of other meanful activities

eddiemuc:
@fm-sys comments are incorporated, CI is green. Let's merge here

main/src/main/res/layout/activity_navigationbar.xml
fm-sys, line 1: @eddiemuc just noticed while reviewing another PR that you added the whereigo gar to only one of 3 layout file variants. Probably this wasn't intended.

eddiemuc: Indeed...

Issue #16285, opened 2024-10-28
custom date format


ztNFny:
c:geo uses the system language to determine the date format, which for english is DD/MM/YYYY. This to me personally is confusing because I'm used that if I see x/x/xxxx it's MM/DD/YYYY and I have found myself scrolling through the logs to find a day >12 to determine whether it's DD/MM or MM/YY.
Most apps use DD MMM (YYYY) format so I've encountered this issue only in c:geo so far.

Request:
A setting to manually set the (short) date format (from a predefined list) or to set the date locale in general.

Issue #16249, opened 2024-10-20
Log view: long date format overflows available space


eddiemuc:
Ticket #778348 https://cgeosupport.droescher.eu/scp/tickets.php?id=8678
c:geo version: 2024.09.18

A user on support has problem with date display in log view: the year is displayed with 4 digits (format dd/mm/yyyy) which leads to truncation on his screen (he sees only the "20" part of a year eg "2023").

His system delivers format string "dd/MM/y" which explains this behavior. However, currently there is no chance to "overwrite this inside or outside of c:geo.

My suggestion is to always use an abbreviated year for display in log view, even if the system returns a "4-digit-year" as short date format.

Original user request:
is there a setting I can change so that everyone's logs for a cache appear with date format dd/mm/yy instead of dd/mm/yyyy please? On my screen, all I see is dd/mm/20 and the rest of the year is missing. 

- c:geo version: 2024.09.18
- Set language: en_GB (system default)
- System date format: dd/MM/y

SammysHP:
I'm not sure the date format is to blame here. The layout just doesn't use the screen space efficiently. What about making the log text almost full-width (with a bit of margin on the left side) and give name, date and type its own full-width header?

eddiemuc:
> I'm not sure the date format is to blame here. The layout just doesn't use the screen space efficiently. What about making the log text almost full-width (with a bit of margin on the left side) and give name, date and type its own full-width header?

That could be done, but it would not be what the user requested. He wants an abbrevated year format

SammysHP:
> but it would not be what the user requested

Sure, but we are dealing with an [XY problem](https://en.wikipedia.org/wiki/XY_problem) here.

eddiemuc:
> > but it would not be what the user requested
> 
> Sure, but we are dealing with an [XY problem](https://en.wikipedia.org/wiki/XY_problem) here.

I can't look the user in the head of course. But if I were the user I would want an abbrevated date instead of displaying a superfluous "20" in front of every year in every log date. This is totally independent from any layout discussions of the log view

SammysHP:
> I can't look the user in the head of course.

Fortunately that is not necessary.

> But if I were the user I would want

It does not matter what the user wants. They might have a proposal for how to achieve their goal, but that is neither a description of the problem nor the only way to fix it. There might be an alternative.

> an abbrevated date instead of displaying a superfluous "20" in front of every year in every log date.

[...]

PR #16286, opened 2024-10-28
make short date format configurable

ztNFny:
added a new "Localization" section in Settings where I grouped the new option with the old unit and language:
image

available configurable date formats:
image

in use:
image

fixes https://github.com/cgeo/cgeo/issues/16285
fixes https://github.com/cgeo/cgeo/issues/16249

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -2522,4 +2522,8 @@ public class Settings {
         newDate.setTimeInMillis(getLong(R.string.pref_last_used_date, newDate.getTimeInMillis()));
         return newDate;
     }
+
+    public static String getShortDateFormat() {
+        return getString(R.string.pref_short_date_format, "");
+    }
 }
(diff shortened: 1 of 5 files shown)


ztNFny:
Settings item is in wrong position, will fix later, WIP for now

fm-sys:
> Also display texts should show examples

Maybe use the current day as example format? (this wouldn't really work at days like 02.02. though)

ztNFny:
> > Also display texts should show examples
> 
> Maybe use the current day as example format? (this wouldn't really work at days like 02.02. though) 

That'd require the display texts to be generated which is probably overkill. 
I'd just go do m for something like 17.2.2024 to make it quite clear in any date format

ztNFny:
retest this please

moving-bits:
I like the approach to make the date format configurable + grouping of localization preferences. Not sure, though, if we have covered all relevant data formats.

Alternative approaches could be:
- automatically retrieve a list of available date formats and create the list dynamically - covers all available date formats, but can have lots of duplicates (if not filtered further)
- make the formatting arguments available directly - most flexible, but probably too technical for most users

But let's give it a try in the nightlies and see if more date formatting wishes come up.

ztNFny:
> * make the formatting arguments available directly - most flexible, but probably too technical for most users

this technically is already available - you can go into the settings editor and enter your custom format

PR #16290, opened 2024-10-30
more refinements to map buttons, incl. touch effect

ztNFny:
fixes #16255

and some refactoring :-)

PR #16296, opened 2024-10-31
add shadows to map overlay buttons

ztNFny:
As requested by @moving-bits a [long time ago](https://github.com/cgeo/cgeo/pull/13063#issuecomment-1146778743) this PR now adds a shadow effect to all map buttons (zoom, quick settings, location, compass).

Small but unavoidable side effect: The buttons get a bit darker. This is because the buttons have white background with a slight transparency and the shadow is drawn behind the whole button - so also behind the transparent part.
As the buttons are already #FFF there's no way to make them whiter to offset the shadow. Reducing the transparency would make them whiter - but also less transparent, which they already are only slightly.

Alternative: We remove the transparency completly. (Transparency levels between old and none don't work, even old has only little transparency, any tuning very quickly arrives at no transparency)

(this PR relies on/includes https://github.com/cgeo/cgeo/pull/16290)

| Before | After |
| :---:   | :---: |
| Screenshot_20241031_190942 | Screenshot_20241031_201203 |

| darker transparent buttons = After | Non-transparent buttons | 
| :---: | :---: |
| Screenshot_20241031_201203 | Screenshot_20241031_201311 |

### DIFF
--- a/main/src/main/res/drawable/bearing_indicator.xml
+++ b/main/src/main/res/drawable/bearing_indicator.xml
@@ -1,16 +1,12 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-    android:width="40.8dp"
-    android:height="40.8dp"
-    android:viewportWidth="24"
-    android:viewportHeight="24">
+    android:width="37.4dp"
+    android:height="37.4dp"
+    android:viewportWidth="22"
+    android:viewportHeight="22">
   <path
-      android:pathData="M12,12m-11,0a11,11 0,1 1,22 0a11,11 0,1 1,-22 0"
-      android:fillColor="@color/osm_zoomcontrolbackground"
-      android:strokeLineCap="round"/>
+      android:pathData="M11,2.516 L7.903,11h2.297A0.8,0.8 0,0 1,11 10.199,0.8 0.8,0 0,1 11.801,11h2.297z"
+      android:fillColor="#ffd32f2f"/>
   <path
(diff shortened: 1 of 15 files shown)


moving-bits:
like it, makes the buttons more readable. Thanks for diving into this!

ztNFny:
> like it, makes the buttons more readable. Thanks for diving into this!

which version? Transparent vs non-transparent?

moving-bits:
> > like it, makes the buttons more readable. Thanks for diving into this!
> 
> which version? Transparent vs non-transparent?

IMHO the non-transparent version is even more readable, as the buttons stand out a bit more

ztNFny:
after using both versions for half a day each the transparent one feels more natural, cgeo-ish to me. Could be because I'm used to it being slightly transparent. So I vote in favor of transparent, but could live with both version. A 70-30 vote I guess.

ztNFny:
there are some more issues with the non-transparent versions, like the scale on Google Maps being covered by the buttons. So I suggest we go with the transparent version.
If further tweaks are required in future we can still adjust.

moving-bits:
Although I would prefer the non-transparent version, the version with transparency would be ok for me as well.

eddiemuc:
Does this PR currently implement the "transparent" version (which from discussion seems to be the preference)? In this case I suggest to merge this PR.

ztNFny:
> Does this PR currently implement the "transparent" version (which from discussion seems to be the preference)? In this case I suggest to merge this PR. 

Yes

PR #16299, opened 2024-10-31
move needs maintenance attribute to own category

### LINKED ISSUE #16297: "firstaid_yes" attribute should be "Needs maintanance"
The "firstaid_yes" (id 42) attribute on GC is grouped into CacheAttributeCategory.CAT_SURROUNDINGS, probably to indicate something like "there's a first aid kit nearby" - which maybe the attribute, judging by it's name, was once intended to be.
However the attribute is used to indicate "Needs maintanance", being added after such a log.

Group assignment and design of the icon should be changed, either to black background or a whole new category

ztNFny:
move the "firstaid" attribute which is only used on GC for "needs maintenance" to its own category:
image
image

fixes #16297


eddiemuc:
lgtm, thanks for the fix

PR #16240, opened 2024-10-19
adjust event time recognition

ztNFny:
adjusts the regexp to work for all languages and adds a additional try-catch to avoid issues on furture website changes

PR #16302, opened 2024-11-01
12 hour date formats are stupid ...

ztNFny:
https://github.com/cgeo/cgeo/pull/16240 had some flaws that this PR now adresses:
- event times of 0:xx / 12:xx had a 12 hour offset (seriously, who invented this stupid 12h time format?!?! doesn't make any sense at all)
- GC didn't migrate all old events (???) and some events now had a duplicate time
- format hours as 2-digits (9:30 -> 09:30)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -386,11 +386,11 @@ public final class GCParser {
                     sDesc.append("<b>")
                             .append(new SimpleDateFormat("dd MMMM yyyy", Locale.getDefault()).format(cache.getHiddenDate()))
                             .append(", ")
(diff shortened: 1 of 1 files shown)


moving-bits:
We got a ticket on support probably pointing into the same direction: (ticket 131152)

_"I see the latest c:geo version now shows event times again.  However, I think there's still a small issue - on e.g. GCAZCV6 it shows as being 13 hours long, from 11am till midnight, rather than the scheduled 1 hour from 11am to 12pm (noon)."_

ztNFny:
> I'll now go look into the suggestions from [#16240 (comment)](https://github.com/cgeo/cgeo/pull/16240#issuecomment-2425214938) if they can make life easier

Should have done that from the beginning, too bad I didn't see it when I started the implementation (and then didn't have time to look into the comment).
Updated the code to use that object which makes it a looooot easier. No more regex magic required

moving-bits:
Thanks for updating the implementation

ztNFny:
@moving-bits is this already in a released build? I just realized some issues with it that I need to look into :-(

ztNFny:
> > @moving-bits is this already in a released build? I just realized some issues with it that I need to look into :-(
> 
> yes, it's part of current release 2024.11.09

Damn. But it seems few people are looking at events outside their home timezone 😉

PR #16308, opened 2024-11-01
fix #16219: wherigo: add cache context via cache detail selection and save file

### LINKED ISSUE #16219: Wherigo Player: add an optional Geocache context
The current Wherigo Player in c:geo is "Cartrdige-centric". It allows one to download cartrdiges, play a cartridge, saving a game's state and loading a previously stored state.

When used from within c:geo, Cartridges are most often associated with a Geocache. 

1. In most cases this is a 1:1 relation (A Wherigo Geocache is linked to exactly one Wherigo Cartridge)
2. There are also Cartridges linked to many geocaches eg the famous [Reverse Wherigo Cartridge](https://www.wherigo.com/cartridge/details.aspx?CGUID=dcdcd2ff-c171-4487-93bc-678f6d03ac4f) and cartridges linked to NO geocache. 
3. The other way around, there are definitely many Geocaches w/o a cartrdiges and probably some with multiple cartridges associated too (don't know an example for the later one though)

Question is whether it is useful to establish a "link" between geocaches and cartridges in c:geo in any way. Possible use cases:

- a user may want to "see" what geocache he is after while playing the Wherigo (for case 1 above)
- a user may want to store/load geocache-specific savefiles for a cartridge (for case 2 above)

Second question would be how this can be achieved. Ideas:
- c:geo could store in a local database all associations between cartridges (identified by cguid) and geocaches (identified by geocode) that it is aware of
- whenever a geocache listing is opened and parsed, such a connection could be made and stored
- when a cartridge is opened via a cache detail page, then the cartridge by default is "played" in the context of this cache
- on starting to play a cartridge (either new game or load game), a "cache context" could be prefilled as best as possible, but changed by the user
- save files could be context-specific (eg filename contains geocache code)

Thoughts?

eddiemuc:
fix #16219: wherigo: add cache context via cache detail selection and save file

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1601,7 +1601,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                 CacheUtils.setWherigoLink(activity, cache, binding.sendToWhereyougo);
             }
             binding.playInCgeo.setVisibility(isEnabled && Settings.enableFeatureWherigo() ? View.VISIBLE : View.GONE);
-            binding.playInCgeo.setOnClickListener(v -> WherigoActivity.startForGuid(activity, WhereYouGoApp.getWhereIGoGuid(cache), true));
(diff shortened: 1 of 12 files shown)


eddiemuc:
image

image

image

Issue #16314, opened 2024-11-03
NPE on importing a GPX file


moving-bits:
similar to #16310, but different symptoms:

Ticket 343578

c:geo fails to import a GPX file. Log reveals the following error:

```
11-02 18:09:32.818 10355 10922 E cgeo    : [Thread-6] Importing caches failed - unknown error: 
11-02 18:09:32.818 10355 10922 E cgeo    : java.lang.NullPointerException: Attempt to invoke virtual method 'boolean cgeo.geocaching.models.Geocache.recalculateWaypoints(cgeo.geocaching.models.CacheVariableList)' on a null object reference
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.CacheVariableList.recalculateWaypoints(CacheVariableList.java:106)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.CacheVariableList.saveState(CacheVariableList.java:73)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.Geocache.changeVariables(Geocache.java:1884)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.Geocache.addCacheArtefactsFromText(Geocache.java:1859)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.Geocache.addCacheArtefactsFromNotes(Geocache.java:1817)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.models.Geocache.addCacheArtefactsFromNotes(Geocache.java:1809)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.storage.DataStore.lambda$saveCaches$20(DataStore.java:2403)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.storage.DataStore.$r8$lambda$x6se3Wos1Bdhqcul_16qtR7gcQw(DataStore.java:0)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.storage.DataStore$$ExternalSyntheticLambda26.run(R8$$SyntheticClass:0)
11-02 18:09:32.818 10355 10922 E cgeo    : 	at cgeo.geocaching.storage.DataStore.withAccessLock(DataStore.java:1023)
[...]

PR #16317, opened 2024-11-03
rel to #16314: cachevariablelist: fix npe on nonavailable cache

eddiemuc:
rel to #16314: cachevariablelist: fix npe on nonavailable cache


eddiemuc:
retest this please

PR #16335, opened 2024-11-10
wig gui

ztNFny:
- separate drawable for internal WIG player with correct size

image

- separate box on CacheDetailsActivity with its own explanatory text
image

- remove the accent style from WIG buttons - this is not used anywhere in c:geo and so shouldn't be used for WIGs either
before
image

after
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1276,7 +1276,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             binding.removeFromWatchlist.setOnClickListener(new RemoveFromWatchlistClickListener());
             updateWatchlistBox(activity);
 
-            // WhereYouGo, ChirpWolf, Adventure Lab
+            // internal WIG player, WhereYouGo, ChirpWolf, Adventure Lab
+            updateWherIGoBox(activity);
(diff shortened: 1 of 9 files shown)


eddiemuc (commented):
LGTM.
Only finding (more a question) is the casing of Wherigo. It is written like "WherIGo" in this PR. Where is this coming from? On official page and anywhere else I looked Wherigo is written w/o intermal casing simply as "Wherigo".

Is this supposed to be a c:geo-style brading of Wherigo? (like eg Geeoh Go calls its Player "WheriGO" and GCDroid calls it "WheriDroid"? If yes, shouldn't we use something more related to c:geo like e.g. Wheri:geo?

ztNFny:
I actually thought that's the official way of writing it, capitalizing the words. But after some google it seems it isn't ...

Do we want to call the c:geo internal player by some name? But maybe that's another discussion not to be had in this PR ... I'll just change it to lowercase for now

eddiemuc:
> I actually thought that's the official way of writing it, capitalizing the words. But after some google it seems it isn't ...
> 
> Do we want to call the c:geo internal player by some name? But maybe that's another discussion not to be had in this PR ... I'll just change it to lowercase for now

Ok, then I wait with merge.

eddiemuc:
Lgtm, thanks for the great work especially with the icons!

Issue #16338, opened 2024-11-10
WIG map shows caches (by default)


ztNFny:
Mapping a WIG, no matter whether it's the "global" map button or an individual waypoint, show's a map with all offline stored caches (no matter whether live mode is enabled or not).
I'd expect the map to show only the WIGs zones.

eddiemuc:
Current implementation just opens the map in it's "current mode" (no parameter istransported to the map) with the viewport set to either a single zone (if a zone is selected) or zoomed out to display all zones (if "global" is selected).

The request made here would require a new "wherigo map mode" or similar. 

I am open to this discussion but personally don't have a use for a map view which is "wherigo only". I require however a way to turn the map back to "live" quickly if this feature should be implemented. (current "live" button is enough for that).

ztNFny:
I'm not asking for a new mode, just the mode that's already existing for e.g. viewing a single cache. Mapping a WIG or a WIG zone is identical to mapping a cache or a single waypoint of it from a usecase point of view - but the implementation is different - thus it's a bug.

eddiemuc:
Currenty the (unified) map has following types (sorry, it is not called "mode"). Which one would be the one to use?

![image](https://github.com/user-attachments/assets/adee2964-81f4-4783-b890-ad2ee44bb519)

ztNFny:
Found it - GeocacheFilterContext.FilterType

        LIVE(R.string.cache_filter_contexttype_live_title),
        OFFLINE(R.string.cache_filter_contexttype_offline_title),
        TRANSIENT(R.string.cache_filter_contexttype_transient_title);

ztNFny:
Only selected cache shown:
            final UnifiedMapType mapType = new UnifiedMapType(geocode);
            mapType.filterContext = new GeocacheFilterContext(GeocacheFilterContext.FilterType.TRANSIENT);
            mapType.launchMap(fromActivity);

All caches shown:
            new UnifiedMapType(geocode).launchMap(fromActivity);

ztNFny:
PR coming in :-)

eddiemuc:
> Found it - GeocacheFilterContext.FilterType
> 
> ```
>     LIVE(R.string.cache_filter_contexttype_live_title),
>     OFFLINE(R.string.cache_filter_contexttype_offline_title),
>     TRANSIENT(R.string.cache_filter_contexttype_transient_title);
> ```
> 
> Live = Offline + Live Offline = Offline only Transient = none at least that's how I read it

[...]

PR #16339, opened 2024-11-10
don't enable live map for WIG mapping

ztNFny:
if map filter context is explicitly set to GeocacheFilterContext.FilterType.TRANSIENT don't enable Live mode, no matter what map type is displayed

@moving-bits there's a lot of "magic" happening in determining when to show which caches (only selected, stored, live) that's quite intransparent even after debugging this code. Thus I'm asking for your review as most of this seems to come from you.
Maybe this magic could also be replaced by a more transparent implementation, but I fear that'd be a big rewrite. Or maybe it's not so magic after all and I'm just missing something??

fixes https://github.com/cgeo/cgeo/issues/16338

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/DefaultMap.java
+++ b/main/src/main/java/cgeo/geocaching/maps/DefaultMap.java
@@ -146,4 +146,12 @@ public final class DefaultMap {
             mapType.launchMap(fromActivity);
         }
     }
+
+    public static void startActivityWherigoMap(final Activity fromActivity, final Viewport viewport, final String mapTitle) {
+        if (Settings.useUnifiedMap()) { // only supported for UnifiedMap
+            Log.e("Launching UnifiedMap in viewport mode, viewport=" + viewport + ")");
+            final UnifiedMapType mapType = viewport == null ? new UnifiedMapType() : new UnifiedMapType(viewport, mapTitle);
+            mapType.launchMap(fromActivity);
+        }
+    }
 }
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapType.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapType.java
@@ -55,6 +55,13 @@ public class UnifiedMapType implements Parcelable {
         this.viewport = viewport;
     }
 
+    /** open map and scale to a given viewport */
+    public UnifiedMapType(final Viewport viewport, final String title) {
+        type = UnifiedMapTypeType.UMTT_Viewport;
+        this.viewport = viewport;
+        this.title = title;
+    }
+
     /** set geocode as target */
     public UnifiedMapType(final String geocode) {
         type = UnifiedMapTypeType.UMTT_TargetGeocode;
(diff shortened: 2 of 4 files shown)


ztNFny:
marking as don't merge as this is probably not a good approach based on comments - hoping for a better implementation :-)

ztNFny:
let's try this in another way. As UMTT_Viewport is not used outside of WIG it shouldn't have any additional impact

moving-bits (changes requested):
> let's try this in another way. As UMTT_Viewport is not used outside of WIG it shouldn't have any additional impact

This seems to work, technically, for displaying the zone without any other cache, similar to navigating to a single cache, but still gives an inconsistent UX:
- on mapping a single cache with "navigate to", the map's title changes to the name of this cache,
- whereas on "show this zone on map", the map title stays at "offline", though not displaying any offline caches.

Thus, if pursuing this way, we should use the zone name as title (see `UnifiedMapActivity.calculateTitle()`), and we should rename the `UnifiedMapTypeType` from `UMTT_Viewport` to something WIG-related.

ztNFny:
> Thus, if pursuing this way, we should use the zone name as title (see `UnifiedMapActivity.calculateTitle()`), and we should rename the `UnifiedMapTypeType` from `UMTT_Viewport` to something WIG-related.

While that Feature Request is unrelated to this PR I nevertheless added it to get this PR passed.

main/src/main/java/cgeo/geocaching/maps/DefaultMap.java
```diff
@@ -146,4 +146,12 @@ public static void startActivityViewport(final Activity fromActivity, final View
             mapType.launchMap(fromActivity);
         }
     }
+
+    public static void startActivityWherigoMap(final Activity fromActivity, final Viewport viewport, final String mapTitle) {
+        if (Settings.useUnifiedMap()) { // only supported for UnifiedMap
+            Log.e("Launching UnifiedMap in viewport mode, viewport=" + viewport + ")");
```
eddiemuc, line 152: Use `Log.iForce()` for info logs which should appear in the logfile anyway.

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapType.java
```diff
@@ -55,6 +55,13 @@ public UnifiedMapType(final Viewport viewport) {
         this.viewport = viewport;
     }
 
+    /** open map and scale to a given viewport */
+    public UnifiedMapType(final Viewport viewport, final String title) {
```
eddiemuc, line 59: With this new constructor in place, I assume that the constructor taking only a Viewport (not a name) is no longer used. Thus it should be removed.

eddiemuc (approved):
LGTM. I have two minor comments only

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapType.java
eddiemuc: I introduced it for Wherigos... it was not there before. But as said, it is minor. No harm in leaving it in

PR #16345, opened 2024-11-15
Allow changing coordinates for UDC

moving-bits:
## Description
Allow changing coordinates for UDC by using a menu entry "set coordinates", which opens the coordinate input dialog:

|menu|dialog|calculate|
|---|---|---|
|image|image|image|

@eddiemuc 
When using calculated coordinates, I need a `calcStateString`, which would normally be stored in a waypoint. Do we have a place for it on cache level already, and how would I retrieve the new calc string, if a user changes the formula?

Setting WIP label for now

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -53,6 +53,7 @@ import cgeo.geocaching.log.LogCacheActivity;
 import cgeo.geocaching.log.LoggingUI;
 import cgeo.geocaching.models.CacheArtefactParser;
 import cgeo.geocaching.models.CalculatedCoordinate;
+import cgeo.geocaching.models.CoordinateInputData;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
 import cgeo.geocaching.models.Trackable;
@@ -81,6 +82,7 @@ import cgeo.geocaching.ui.ToggleItemType;
 import cgeo.geocaching.ui.TrackableListAdapter;
 import cgeo.geocaching.ui.UserClickListener;
 import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.ui.dialog.CoordinatesInputDialog;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.EditNoteDialog;
 import cgeo.geocaching.ui.dialog.EditNoteDialog.EditNoteDialogListener;
@@ -101,6 +103,7 @@ import cgeo.geocaching.utils.ImageUtils;
 import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
+import cgeo.geocaching.utils.MenuUtils;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.ProgressBarDisposableHandler;
 import cgeo.geocaching.utils.ProgressButtonDisposableHandler;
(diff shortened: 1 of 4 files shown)


eddiemuc:
> @eddiemuc When using calculated coordinates, I need a `calcStateString`, which would normally be stored in a waypoint. Do we have a place for it on cache level already, and how would I retrieve the new calc string, if a user changes the formula?

The `calcStateString` is the stored configuration of the underlying `CalculatedCoordinate` instance which is used by the Waypoint to do the actual coordinate calculation. Such a config has the format `{CC|<latPatern>|<lonPattern>|<type-char>}` where the "pattern" parts refer to the actual formula text entered by the user in the lat-field and the lon-field. type-char is needed only to store the edit mode the user was using. (either plain or one of the different "coordinate entering helpers")

A `CalculatedCoordinate` can be retrieved from such a string using CalculatedCoordinate.createFromConfig(config). From a given `CalculatedCoordinate`, the confir can be obtained using `getConfig()`.  The config string is used to store cc state (eg in Activity lifecycle), in database and in personal notes (for storing and parsing waypoints).

As of now, calculated coordinates are only used for waypoints. If you want to allow editing calculated coordinates on a cache level then this is new territory, but possible due to encapsulation of all the logic in `CalculatedCoordinate`:
* For persistance you would at least need a new database field on cache level storing the state above. 
* You need to ensure that every time a variable value is changed that the geocache coordinate is recalculated, updated across c:geo and stored in database. 
* Note that the current calculated coordinate (for waypoints) is persisted in database in addition to the formula and must be updated in case needed. This is due to location-based filters relying on filtering on a database level for performance reasons. Same is needed for geocache of course. 
* If you use calculated coordinates, you have to handle at all relevant places the case that a coordinate is (termporary) not valid because calculation is invalid (variable does not exist, dividing by zero, ...) or results in illegal coordinate

There might be more, which I forgot right now. If this is too much on one go, I recommend first allowing UDCs to have only static user-entered coordinates and add variable-usage in a later step.

eddiemuc:
My comment above on calculated coordinates was updated

ztNFny:
how about just creating a waypoint with the option "set as correct coords" set by default? That way the change in logic would be minimal, all the existing UI could be kept and the same goal achieved. And if you don't want to keep the waypoint it could be deleted again in the same transaction.

ztNFny:
Here's a quick and dirty implementation on how I'd do it:
https://github.com/ztNFny/cgeo/commit/661d067ff726d3a1eba3bc732d03ac4ff007ef83

moving-bits:
Thanks, @eddiemuc and @ztNFny, for your feedbacks.

> [...] There might be more, which I forgot right now. If this is too much on one go, I recommend first allowing UDCs to have only static user-entered coordinates and add variable-usage in a later step.

Thanks for the detailed description of the inner workings and the things that we would have to keep in mind if we wanted to pursue the "use variables for cache coordinates" thing. I tend to skip that part - the original issue was about setting specific coordinates only, and we should probably avoid unnecessary complications.

> Here's a quick and dirty implementation on how I'd do it: [ztNFny@dac3258](https://github.com/ztNFny/cgeo/commit/dac325882dab48aa7bdda50e5213853d46fbdecd)

That's an interesting approach, thanks. It gives the additional possibility of using a projection + using variables, but it leaves an (otherwise) unnecessary waypoint. (This could be healed by storing the related info attached to the cache itself, but that would require additional handling, see above.) And it also adds the "coordinates changed" flag, which may be confusing.

I'll play around with it a bit more...

ztNFny:
It shouldn't leave an extra waypoint, I'm deleting that one. But as it's an untested inplementation maybe i missed something - it's just a demonstrator and far from refined.

moving-bits:
Updated the PR to skip "calculated waypoint" support. It now has "my coordinates" and "cache coordinates" (= reset) options, re-using the infrastructure c:geo uses for waypoints.

image

moving-bits:
retest this please

ztNFny:
> Updated the PR to skip "calculated waypoint" support. It now has "my coordinates" and "cache coordinates" (= reset) options, re-using the infrastructure c:geo uses for waypoints.
> 
> image
> 

What's the point of "cache coordinates"? As i understand it the only purpose of that dialog os to change the caches listing coords - so "cache coords" wouldn't do anything?

moving-bits:
> What's the point of "cache coordinates"? As i understand it the only purpose of that dialog os to change the caches listing coords - so "cache coords" wouldn't do anything?

It's sort of a "reset" button - if you have changed the coordinates, you can push that button to reset them to the cache's coordinates. (Or you can close the dialog without saving, and reopen.)

PR #16354, opened 2024-11-16
UnifiedMap Mapsforge: restrict zoom to min/max zoom (fix #16352)

### LINKED ISSUE #16352: WIG zone display zooms in too far
This is probably not WIG-related but an error in map (provider) configuration:

If you let c:geo's internal WIG player display a zone ("show on map" for zone), it will zoom in the map as much as possible - which leads to a blank map background for Mapsforge online maps, which do not support that zoom level.

image.image

Seems to affect UnifiedMap Mapsforge with online maps only, but neither UnifiedMap Mapsforge with offline maps nor UnifiedMap VTM.

I'll look into it.

moving-bits:
## Description
Restricts UnifiedMap Mapsforge's zoom to zoom min/max given by tile provider, even for `zoomToBounds()` call (which is called in UMTT_Viewport mode used by WIG player). This is relevant especially for online tile providers, which do not provide map tiles for higher zoom levels.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/AbstractMapFragment.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/AbstractMapFragment.java
@@ -154,7 +154,7 @@ public abstract class AbstractMapFragment extends Fragment {
     }
 
     public int getZoomMax() {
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

PR #16355, opened 2024-11-16
fix #16320: fix wherigo audio quirks, add basic sound controls

### LINKED ISSUE #16320: Wherigo: fix quirks in handling of audio
Wherigo player has some quirks when handling audio:

- Example "Uma Aventura no Carrical" (GC5ARN2, cguid: c1a4fd95-07d7-4667-94c1-d2e6eead9cae): this cartridge, when opened, should play an audio file endlessly (in a loop)
   - in c:geo Wherigo player the playback is stopped suddenly after a while. Why is this?
   - when wherigo player is ended, playback of sound file should be ended too. 
   - if wherigo plays an audio file, there should be the option to stop it, maybe also restart it
- Example  "2016 hin und 2017 zurueck" (cguid: 2677e80d-7dc0-4731-a4ac-2555de932573): this cartridge, when opened, shall play an audio file one time (not in a loop) which is a little longer. But after a while most of the times the audio stops playing

eddiemuc:
fix #16320: fix wherigo audio quirks, add basic sound controls


eddiemuc:
image

eddiemuc:
image

Issue #16341, opened 2024-11-11
Buttons with and without accent color


eddiemuc:
While working on Wherigo Player and triggered by a rightful comment of @ztNFny I wondered what the " design rule" is in c:geo for buttons and usage of ascent color. It seems that:
* pure text buttons are in ascent color
* pure icon buttons are NOT in ascent color
* buttons with both icon and text are in ascent color

I concluded the above "rules" from the way I found it in c:geo as of now. There are however also exceptions to some of these rules nowadays. Question is:
* should those be the button ascenting rules? If not, how should they be changed? Or are there no such rules?
* if we have rules, should we enforce them to existing screens?

Example for icon buttons (NOT in ascent color, taken from cache details):
![image](https://github.com/user-attachments/assets/16b04775-8b75-4176-8da9-1a31482ab8f4)

Example for text buttons (in ascent color, taken from waypoint tab):
![image](https://github.com/user-attachments/assets/b5a249b5-7b8e-4df8-b93c-d2a2eb89f7c2)

Example for icon + text buttons (in ascent color, taken from waypoint details):
![image](https://github.com/user-attachments/assets/778cb2e3-c1a7-4502-98ca-63999f81dfe5)
![image](https://github.com/user-attachments/assets/291e3caf-0aa2-4ea9-a114-5b63dc41f384)

And then there's the variable tab with all-ascent buttons. BUT according to the rule at least the first button (the one with the "plus"-icon only) would have to be turned into a non-ascent button wouldn't it? That would look strange though
![image](https://github.com/user-attachments/assets/cf32b5dd-3712-4f2c-847f-8b6387a02484)

And here is the Wherigo player with both icon buttons and a text button (taken from wherigo activity whilst a dialog was paused):
![image](https://github.com/user-attachments/assets/fa436eb9-62e2-4002-a63e-4da5bb462e5e)

Finally, I just saw that the icons in the Wherigo buttons are now no longer in ascent color but only the texts are (taken from cartidge detail dialog):
![image](https://github.com/user-attachments/assets/5d75c3eb-b5e9-484c-b125-bc2a7812b9ee)
(this looks kinda nice. Should all buttons with text and icons be changed into this format?)

ztNFny:
I like the white icon + orange text style.
Having different icon color for "icon only" vs "icon with text" as "proposed" in the sentence about variable tab seems to be a really confusing approach.

eddiemuc:
[...]

PR #16359, opened 2024-11-17
Variables tab icons

ztNFny:
changed the icons in variables tab in a way that imho they can work without texts. It's 2 commits, 1. to change the icons, 2. to remove the texts, so in case you oppose to removing the texts the first could still be used standalone.

related to https://github.com/cgeo/cgeo/issues/16341

image


eddiemuc:
> ... in case you oppose to removing the texts...

In case the "you" relates to me (?): In like your approach a lot!

There's just this one button adding the "next free variable" (the "simple plus" button in your PR) where I would prefer if the variable to be added would be somehow made visible to the user. But that's maybe a detail.

I would merge this PR, I keep it open for a while to let others share their thoughts though.

ztNFny:
> In case the "you" relates to me (?): In like your approach a lot!

yeah, i had actually typed that text in the issue and then realized it should go into a comment :-D

> There's just this one button adding the "next free variable" (the "simple plus" button in your PR) where I would prefer if the variable to be added would be somehow made visible to the user. But that's maybe a detail.

on that button I wanted to keep the text initially, then I realized there's no matching button style existing right now and decided it's not worth the effort to create it ;-) But of course you may disagree with that and re-add the text - feel free!

ztNFny:
@eddiemuc if you want to re-add the text, should I drop this part of the change?
image

eddiemuc:
> @eddiemuc if you want to re-add the text, should I drop this part of the change?

That would be great, thanks!

eddiemuc (approved):
LGTM, thanks for diving into this

Issue #16367, opened 2024-11-18
Event time display follows users timezone, not event


ztNFny:
After https://github.com/cgeo/cgeo/pull/16302 the event time display in c:geo now follows the time zone the user has set in https://www.geocaching.com/account/settings/preferences while previously it was in the events local timezone.

Both versions have their pros and cons:
- Event timezone: I guess that's what users expect to see when viewing a listing. Also it's what GC shows at the top of the listing and what was previously added in short description field - the users timezone setting in preferences is just ignored.
- Users timezone: Actually correct in relation to the user, but maybe unexpected. Calendar entries for an event in another timezone will be correct

(Option 3 would of course be to do both, but that'd require additional DB fields for the times)

I vote for changing it back to event timezone - opinions?

PS: users timezone is of course not available in the json ...

ztNFny:
any feedback on the current solution? would be nice to discover potential issues before another release this time ;-)

moving-bits:
The fix has been released as 2024.11.22, and I have asked two users on support if the problem they had reported has been resolved, which they confirmed. So I guess we can close here for now.
Thanks for adapting the fix.

PR #16368, opened 2024-11-18
Revert "use eventCacheData JS object as event time source"

ztNFny:
revert back to using the HTML as source for event date as that one is in events timezone

see https://github.com/cgeo/cgeo/issues/16367 for discussion


ztNFny:
To clarify: This doesn't fully revert https://github.com/cgeo/cgeo/pull/16302 but only the second commit of it, which switched from HTML to JSON as data source. The original fixes contained in that PR are still in:
- event times of 0:xx / 12:xx had a 12 hour offset (seriously, who invented this stupid 12h time format?!?! doesn't make any sense at all)
- GC didn't migrate all old events (???) and some events now had a duplicate time
- format hours as 2-digits (9:30 -> 09:30)

eddiemuc:
Doesn't this need to go to release branch?

While I am in favor of this PR, the situation is yet another reminder that c:geo has to become truly timezone-aware. In an ideal world (...) we could offer the user both showing expected (event-local) event time and correct calendar entries :-)

ztNFny:
> Doesn't this need to go to release branch?

done

Issue #16357, opened 2024-11-17
Offline Map management issues


moving-bits:
Ticket 899787 on support:
- ~Bug: User has four offline maps configured, but only two are shown in "home screen" => "Manage offline data" => "Delete offline data"~ => fixed with PR #16404
- ~Feature request: User requests to have the "delete offline data" available in map context~ => fixed with PR #16369

ztNFny:
> * Feature request: User requests to have the "delete offline data" available in map context

you could pull the "Manage offline data" from https://github.com/cgeo/cgeo/pull/16127

moving-bits:
WRT to the first part, not all maps being offered for deletion: This may be caused by maps installed manually, as we check for the downloader's companion file. I've asked the ticket issuer for a cross-check.

PR #16369, opened 2024-11-18
Allow deleting offline data from map context (rel. to #16357)

moving-bits:
## Description
Adds "Delete offline data" to map sources menus, which triggers the offline data selection dialog (same as can be reached via home screen => manage offline data => delete offline data)


fm-sys:
I'm not sure whether that yet another feature clutters the map menu more then it's worth it...

fm-sys:
Ah, ok. Fine. I thought it was about the "delete orphaned files". Sorry for the misunderstanding on my side.

PR #16378, opened 2024-11-23
group offline-data related entries in 1 submenu

ztNFny:
image
image

before:
image
image


main/src/main/res/values/strings.xml
```diff
@@ -1980,6 +1980,7 @@
     <string name="onetime_mapthemefixslow_message">Working with non-zipped map themes might decrease map viewer opening performance. You can zip or sync your themes to get faster performance and reduce the delay on starting the map.</string>
     <string name="map_theme_select">Select map theme</string>
     <string name="map_theme_options">Theme Options</string>
+    <string name="map_manage_offline_data">Manage offline data</string>
```
moving-bits, line 1983: This string is already available as `menu_manage_offline_data`, which can be reused as it's for the identical action

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -856,16 +855,6 @@ public boolean onOptionsItemSelected(@NonNull final MenuItem item) {
             mapFragment.selectThemeOptions(this);
         } else if (id == R.id.menu_routetrack) {
             routeTrackUtils.showPopup(viewModel.individualRoute.getValue(), viewModel::setTarget);
-        } else if (id == R.id.menu_select_mapview) {
```
moving-bits, line 859: Have you cross-checked whether creating the menu entries statically in `onPrepareOptionsMenu` recognizes map source changes (eg. deleting one of the offline maps)?

moving-bits (changes requested):
Thanks for cleaning up the menus. Please check the two comments added to the source view.

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
ztNFny: Oh, so that's the idea behind that strange code (which btw exists only in Unified, not in legacy map). For adding new maps this wasn't a problem as the map is added from another activity, but with the re-added Delete it can indeed cause problems.
However I'm wondering whether this relatively rare case is worth the hassle of the working currently implemented (again only for Unified), recreating the menu contents on the fly...

moving-bits (approved):
Thanks for the update

PR #16386, opened 2024-11-24
Wherigo: display warning when mapping zones and unified map is disabled

eddiemuc:
Wherigo: display warning when mapping zones and unified map is disabled

### DIFF
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -3013,6 +3013,8 @@
     <string name="wherigo_otm_shortcuts_title">Wherigo Player Shortcuts</string>
     <string name="wherigo_otm_shortcuts_message">If you want to use c:geo Wherigo Player regularly, consider creating a quick launch button or custom menu item for it under Settings -> Appearance</string>
 
+    <string name="wherigo_map_supported_for_unified_map_only">Displaying of Wherigo locations on map is not supported for your map settings. Please switch to unified map under Settings -> Map Sources -> %s</string>
+
     <!-- default actions in dialogs -->
     <string name="previous">Previous</string>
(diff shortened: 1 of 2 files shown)


main/src/main/res/values/strings.xml
```diff
@@ -3013,6 +3013,8 @@
     <string name="wherigo_otm_shortcuts_title">Wherigo Player Shortcuts</string>
     <string name="wherigo_otm_shortcuts_message">If you want to use c:geo Wherigo Player regularly, consider creating a quick launch button or custom menu item for it under Settings -> Appearance</string>
 
+    <string name="wherigo_map_supported_for_unified_map_only">Displaying of Wherigo locations on map is not supported for your map settings. Please switch to unified map under Settings -> Map Sources -> Unified Map (beta)</string>
```
moving-bits, line 3016: This triggers a question (not directly related to this PR): Shall we remove "beta" status from UnifiedMap in current nightly version?

eddiemuc: I have no opinion about that. I changed the PR such that this question can be answered separately (and, if need be, changed in a separate PR)

ztNFny:
How about showing WIG only if UnifiedMap is enabled? Like to change the feature switch to return the map type

eddiemuc:
> How about showing WIG only if UnifiedMap is enabled? Like to change the feature switch to return the map type

WIG and unified map are unrelated in all but the display of mapping zones. It is perfectly possible tomolay a wherigo w/o map usage. This I don't want to connect the two in this way.

On the contrary, maybe we get a few more users to switch to unified map with current approach

eddiemuc:
BTW there's no feature switch any more for wherigo. It is enabled for all users.

Issue #16391, opened 2024-11-25
Wrong time of events


Bogmen:
### Describe your problem!

Die Startzeit in Events wird falsch angezeigt.
Ein Event, das um 12:00 Uhr starten soll, wird mit 00:00 Uhr angezeigt.
Ein Beispiel:
https://coord.info/GCB0PBW

### How to reproduce?

Ständig

### Actual result after these steps?

./.

### Expected result after these steps?

./.

### Reproducible

Yes

### c:geo Version

2024.11.22

### System information

```text
## System information

c:geo version: 2024.11.22

Device:
-------
- Device type: SM-N960F (crownltexx, samsung)
- Available processors: 8
- Android version: 10
- Android build: QP1A.190711.020.N960FXXU9FVH1
- Screen resolution: 1440x2792px (411x773dp)
- Pixel density: 4.5499997
- System font scale: 1.3 / used scale: 1.3
- Sailfish OS detected: false
- Google Play services: enabled - 24.46.31 (100400-699197446)
- Memory:  Available:2,6 GB, Total:7,5 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:193
- Set language: de_DE (system default)
- System date format: dd.MM.yy
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: 10. Feb. 2023, 23:23
- Last auto backup: 25. Nov., 14:21
- Routing mode: Straight (shortest.brf / trekking.brf / car-eco.brf / (keine) / (keine))
- Live map mode: false
- Use unified map: false
- OSM multi-threading: true / threads: 4
- Map: Kombiniert (Offline)
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMultiMapSource
[...]

PR #16393, opened 2024-11-26
fix event time yet again

ztNFny:


ztNFny:
I don't have access to my IDE (and won't for a few days). Can someone please test this with the event from https://github.com/cgeo/cgeo/issues/16391 and GC website set to languages EN, ES, DE? And also an event at 00:00 with the 3 languages and one with some "normal" time?

moving-bits (changes requested):
Thanks for diving into this!

I did a short test with GCB0PWB and time zone set to "UTC+00:00 Dublin, ..." as is given in the related issue:

With web page language set to English, the PR seems to work for this cache, but it breaks when language is set to Spanish or German. 00:00-13:00 is being shown as event time in those cases.

ztNFny:
back in front of the PC - I hope I found all the potential problems now ...

moving-bits (changes requested):
I've tested with three different types of events (starting at 00:00, 12:00 and 15:30) in three languages (EN, ES, DE) and different time zones - found no problems so far.

Just a small Checkstyle warning left: https://ci.cgeo.org/job/cgeo-CI_PR-build/10347/checkstyleResult/new/

ztNFny:
> I've tested with three different types of events (starting at 00:00, 12:00 and 15:30) in three languages (EN, ES, DE) and different time zones - found no problems so far.
> 
> Just a small Checkstyle warning left: https://ci.cgeo.org/job/cgeo-CI_PR-build/10347/checkstyleResult/new/

fixed

PR #16403, opened 2024-11-30
GPX/XML Parsing: add tools for mixed StAX (Pull) and DOM-like parsing in Android

eddiemuc:
GPX/XML Parsing: add tools for mixed StAX (Pull) and DOM-like parsing in Android

As a prerequisite for a redesign of the GPX parsing, this PR adds XML parsing utils for mixed StaX (Pull) and DOM-like parsing. Unfortunately the later has to be hand-implemented because there's no natively supported function to combine PullParser and DOM directly.


moving-bits:
@eddiemuc 
How critical are those changes - is it safe to include them in a beta version tomorrow or should we wait a few days?

eddiemuc:
Totally uncritical. Code is not used yet anywhere in c:geo

PR #16405, opened 2024-12-01
GPX refactor preparation: introduce new model class NamedGeoCoordinate to handle wpts which are neither Geocache nor Waypoint

eddiemuc:
GPX refactor preparation: introduce new model class NamedGeoCoordinate to handle wpts which are neither Geocache nor Waypoint


eddiemuc:
Reading GPX data we basically have three things to read: GEocache, Waypoint and "basic coordinates". Realizing that the later is also used across c:geo in other places (eg as Route/Track points), I propose to make a new first-class-citizen out of it. This PR does the necessary refactoring: 

* rename `IWaypoint` to `INamedGeoCoordinate`
* rename `ICoordinates` to `ICoordinate` and `CoordinatesType` to `CoordinateType` (the wrong usage of plural always annoyed me here anyway :-))
* move "`getId()`" and "`getWaypointType()`" from `IWaypoint`/`INamedGeoCoordinate` to `Waypoint` (long overdue because both fields were never used for `Geocache` anyway)
* introduce "`getElevation()`" to `ICoordinate`/`ICoordinates` (long overdue if you ask me)
* Provide `NamedGeoCoordinate` as direct implementation of `INamedGeoCoordinate`, introduce respective `CoordinateType` member

If this refactoring goes through, then in the future `NamedGeoCoordinate` might also be used in `RouteItem` and `Route` classes.

This refactoring is touching some cores, so I leave time for review.

eddiemuc:
@moving-bits would be good to have this PR in next beta. No functional change but some refactorings which I would like to have in next release. Otherwise release and master will divert for a long time

moving-bits (approved):

eddiemuc:
Thanks!

Issue #16373, opened 2024-11-20
[UnifiedMap][Nightly] Crash on Mapsforge offline map after reactivate c:geo / map


murggel:
### Describe your problem!

I have some crashes on reactivating map-view after a longer period having c:geo running in the background (leaving map-view in background)

### How to reproduce?

Show some caches in map-view in UnifiedMap Mapsforge
Switch to another app for a longer time
Switch back to c:geo, map view is open, after a very short time c:geo crashes before an action can be finished (panning, opening a cache from map-view)

### Actual result after these steps?

crash

### Expected result after these steps?

no crash

### Reproducible

Yes

### c:geo Version

2024.11.18-NB-79c4a30

### Additional Information

I will send the logcat to support.

What I saw in there, which can help for finding the culprit:
`Bitmap  : Error, cannot access an invalid/free'd bitmap here!`
`Abort message: 'Error, cannot access an invalid/free'd bitmap here!'`

murggel:
PR #16406 for catching the exception, but I am not sure, if this is the right approach

eddiemuc:
Although your report was "earlier" I have the impression that analysis continues more in #16430. Thus closing here (please reopen if you see a different angle here).

PR #16406, opened 2024-12-01
Catch IllegalArgumentException in MapsforgeFragment

murggel:
## Description

Catch IllegalArgumentException while adding / removing observer

## Related issues

rel #16373

## Additional context

Does not happen for me on the emulator, only on my device


murggel:
Catching the exception, but I am not sure, if this is the right approach

PR #16431, opened 2024-12-11
rel to #16430: move addObserver from onstart to onresume to match removeObserver in onPause

eddiemuc:
rel to #16430: move addObserver from onstart to onresume to match removeObserver in onPause

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforge/MapsforgeFragment.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforge/MapsforgeFragment.java
@@ -167,14 +167,13 @@ public class MapsforgeFragment extends AbstractMapFragment implements Observer {
             doReapplyTheme = false;
         }
         mMapView.getLayerManager().getLayers().add(new MapEventsReceiver());
-        mMapView.getModel().mapViewPosition.addObserver(this);
     }
 
     @Override
     public void onResume() {
         super.onResume();
         // mMapView.onResume();
-
+        mMapView.getModel().mapViewPosition.addObserver(this);
     }
 
     @Override


eddiemuc:
I believe that the exception in https://github.com/cgeo/cgeo/issues/16430#issuecomment-2533969389 is caused by the mapsforgeFragment's state moving from PAUSED back to RESUME and then again to PAUSE. In the last change, the observer would be removed already (by previous call to PAUSE) and not been added again in the meantime (because it was done n onStarted instead of onResume()), leading to observed exception.

I cannot reproduce the error case though, so this is a bit guessing.

@moving-bits can you please double-check the change?

eddiemuc:
retest this please

eddiemuc:
failng build is unrelated

murggel:
I can reproduce a crash on the map after resume as well, but I am not sure,  if it is the same error-message.
But I can test with your PR tomorrow .

murggel:
> I can reproduce a crash on the map after resume as well, but I am not sure,  if it is the same error-message.
> But I can test with your PR tomorrow .

With this PR, the map was often frozen after switching back to cgeo map state.

So unfortunately not really an improvement.

eddiemuc:
> > I can reproduce a crash on the map after resume as well, but I am not sure,  if it is the same error-message.
> > But I can test with your PR tomorrow .
> 
> With this PR, the map was often frozen after switching back to cgeo map state.
> 
> So unfortunately not really an improvement.
> 

Then it might be something more tricky to solve

eddiemuc:
> And just catching the exception as a quick and dirty fix?

Sure we can try that. Don't know what happens then though

eddiemuc:
I can reproduce the problem reliably so far on my physical phone with following steps:
- open live map
- make sure that mode is unified map with mapsforge adapter
- make sure that an offline map is selected 
- change the app (in my case In am opening Whatsapp on my phone)
- switch back to c:geo

--> cgeo app freezes, I have to restart 

Unified map vtm seems to work fine in same scenario. I would suggest this as a workaround for users on support for now.

bekuno:
retest this please

PR #16434, opened 2024-12-15
Try to avoid deadlock on layer removal (rel. to #16430)

moving-bits:
## Description
See https://github.com/cgeo/cgeo/issues/16430#issuecomment-2543550292


moving-bits:
retest this please

ztNFny:
This breaks layer order on MF
image

ztNFny:
> This breaks layer order on MF image

actually it doesn't - the order was broken before already . I'll open a ticket

PR #16440, opened 2024-12-20
fix hillshading and routing download button visibility

ztNFny:
download routing/hillshading buttons were visible when they shouldn't be

change the logic for activating hillshading buttons from 1 master toggle in preferences to "any hillshading files are downloaded"


moving-bits:
Is this something that should go into the next release?
If so, then this PR should be targeted against `release`, which is where current beta version is being built from.

ztNFny:
Don't care, master is fine

PR #16451, opened 2024-12-24
UnifiedMap Mapsforge: Use LabelLayer instead of ThreadedLabelLayer

moving-bits:
## Description
As proposed in https://github.com/cgeo/cgeo/issues/16430#issuecomment-2560807345: uses default `LabelLayer` instead of `ThreadedLabelLayer`


eddiemuc:
Build is failing due to unnecessary import statement

PR #16464, opened 2024-12-31
UnifiedMap MF: Update lib to include HS + performance improvements

moving-bits:
see https://github.com/cgeo/cgeo/issues/16438#issuecomment-2566283723

Hillshading: https://github.com/mapsforge/mapsforge/pull/1605, https://github.com/mapsforge/mapsforge/pull/1597
Map fixes & improvements: https://github.com/mapsforge/mapsforge/pull/1599, https://github.com/mapsforge/mapsforge/pull/1586 (and others)

PR #16465, opened 2024-12-31
UnifiedMap MF: Update lib to include HS + performance improvements

moving-bits:
alternative to #16464, targeted to `master`

closes #16464


moving-bits:
superseded by #16467

PR #16466, opened 2024-12-31
UnifiedMap Mapsforge: Try to avoid 'invalid/free'd bitmap' error

moving-bits:
see https://github.com/cgeo/cgeo/issues/16430#issuecomment-2566522437


ztNFny:
since applying this I sometimes get a "blank" map view when switching between lists and map, blank meaning no map, just overlays

devemux86:
Reminder, there is already an issue:
- https://github.com/mapsforge/mapsforge/issues/1595.

devemux86:
We've never seen anything like it, but we never recreate the map view or activity.

Map is only "destroyed" once when exiting the app.

Issue #16473, opened 2025-01-01
“History Track Max Length” and “Far Distance” sliders are entangled


ella-TAS:
### Describe your problem!

The “History Track Max Length” and “Far Distance” sliders in the (extended) Map Content & Behavior settings menu seem to be interconnected. Setting one also sets the other.

### How to reproduce?

1. Enter the (extended) Map Content & Behavior settings menu
2. Set the “Far Distance” slider

### Actual result after these steps?

The “History track max. length” slider is set to the same value and max value (1000 instead of 60000) as the Far Distance slider.

### Expected result after these steps?

The sliders do not interfere with eachother.

### Reproducible

Yes

### c:geo Version

2024.12.30-RC

### System information

```text
## System information

c:geo version: 2024.12.30-RC

Datetime: January 1, 14:01

Device:
-------
- Device type: SM-G973F (beyond1lteeea, samsung)
- Available processors: 8
- Android version: 12
- Android build: SP1A.210812.016.G973FXXSGHWC1
- Screen resolution: 1080x2042px (411x777dp)
- Pixel density: 2.8875
- System font scale: 1.1 / used scale: 1.1
- Sailfish OS detected: false
- Google Play services: enabled - 24.47.38 (190400-705893532)
- Memory:  Available:3.7 GB, Total:7.2 GB, Threshold: 390.0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:99
- Set language: en_US (system default)
- System date format: M/d/yy
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: never
- Last auto backup: never
- Routing mode: Walk (shortest.brf / trekking.brf / car-eco.brf / (none) / (none))
- Live map mode: true
- Use unified map: false
- OSM multi-threading: false / threads: 1
- Map: OpenStreetMap.org
  - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OsmMapSource
  - Attrs: © OpenStreetMap DE, map data OpenStreetMap contributors
  - Theme: none

Filters:
-------
[...]

PR #16474, opened 2025-01-01
don't reuse same seekbar in different holders

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/16473

not sure this is the best way to solve it?


moving-bits (approved):
seems to work as expected now, so let's give it a try

I propose to target this to `release`, though, as the error happens in beta version as well.

ztNFny:
rebased

moving-bits:
this seems to be the "waypoint migration" PR now... 🤷‍♂️

ztNFny:
fixed. that's what happens when you reuse the same local branch for different things :-(

moving-bits:
Thanks for fixing

PR #16483, opened 2025-01-01
remove non-functional filter from lists activity

ztNFny:
the filter in bookmark lists activity didn't do anything, the preferences weren't used anywhere, so removing it


moving-bits:
Is this about the "show all" filter in action bar?

That one should be kept:

It filters between "all" and "only new" lists, meaning: Lists that have either not being downloaded yet or have changed after the last download. You will see a small (new) beside the date info below the list's name.

ztNFny:
> Is this about the "show all" filter in action bar?
> 
> That one should be kept:
> 
> It filters between "all" and "only new" lists, meaning: Lists that have either not being downloaded yet or have changed after the last download. You will see a small (new) beside the date info below the list's name.

huh, I actually couldn't figure out what that one does, even by looking at the code. I suggest to at least change the label to something more explicit than "show all"

Issue #16459, opened 2024-12-28
[Unified] Waypoints of stored Labs are not made visible visible until lab cache is tapped, "visited" waypoints are not filtered out


eddiemuc:
c:geo-Version: 2024.12.17-RC

Tested with Lab "Zwischen Einkaufstempel und Architektur" in Hamburg near N 53° 33.053' E 009° 59.636'.
This problem might also occur with other cache types, I haven't tested this.
Problems happens on Unified Mapsforge, VTM and Google.

Precondition: 
- unified map is selected
- map position is currently placed somewhere such that an offline-stored Lab Cache and it's waypoints are inside the viewport. 
- Some of the waypoints are marked as "visited", some are not
- map is not currently opened (eg we might be in "stored list" section)

Error reproduction:
- Open the map --> Lab is shown (ok) but not its waypoints (Error):
![image](https://github.com/user-attachments/assets/8abacfbf-16a9-4bd8-a50c-c0fd597bfa8b)

- open the popup for the lab cache:
![image](https://github.com/user-attachments/assets/d152e444-d4ed-4242-9a25-bdeae1147a30)

- close popup again --> waypoints of lab are now shown...
![image](https://github.com/user-attachments/assets/105e4362-4acd-408d-abaa-40ccae243ad4)

- ...but already visited waypoints are ALSO shown allthough they should be filtered out:

![image](https://github.com/user-attachments/assets/5cbd8b01-dcef-4e1d-8772-6724645fe6c0)

eddiemuc:
@moving-bits I debugged into this and found that when initially opening the map, the only place where waypoints are loaded is the method `UnifiedMapActivity.loadWaypoints()`. This method however is called only once with an illegal (just-a-dot) viewport, resulting in zero waypoints being loaded. 

I am not sure how the cache/waypoint loading/filtering architecture of unified map is supposed to behave here.

eddiemuc:
Can't extract the systematics of this out of the code. If I would need to tackle this then I would need to redesign a lot.

bekuno:
Maybe the same reason as in #16461?

eddiemuc:
> Maybe the same reason as in #16461?

There seem to be some things intertweened here. I will take a look at #16461 too.

While testing around I noticed that each change on a cache (eg marking a waypoint as visited) causes a retriggering of the online search in live mode. This is not ideal

moving-bits:
[...]

PR #16484, opened 2025-01-01
rel to #16459: redesign live map update, many small refactorings

eddiemuc:
rel to #16459: redesign live map update, many small refactorings

While working on #16459 I stumbled from one thing into another. The result is a pretty big redesign and needs to be evaluated in nightlys before it can go to beta.


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -264,6 +274,11 @@ public void onCreate(final Bundle savedInstanceState) {
 
         viewModel.mapCenter.observe(this, center -> refreshListChooser());
```
moving-bits, line 275: If we add an observer to the whole viewport, do we still need to observe map center?

moving-bits:
Thanks for diving into this rather complex issue. Haven't cross-checked everything - PR is too large for me so early in the new year :-)

Shall we merge this to `master` for further testing and getting feedback, or are you still working on some things here?

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
eddiemuc: That's a very good point, and I don't think we need both. Also, the way that mapCenter is set/updated is much better than my approach. Thanks for the review, I will rework this part of the PR

eddiemuc:
> Thanks for diving into this rather complex issue. Haven't cross-checked everything - PR is too large for me so early in the new year :-)
> 
> Shall we merge this to `master` for further testing and getting feedback, or are you still working on some things here?

Yeah I got a little carried away... 
I will redesign a bit based on your comment, please don't merge yet

main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
eddiemuc: ok, fixed

eddiemuc:
Should be ready for merge now.

Note: this PR also includes a new state system of the Live Geocache loader. This state is indicated using the "load progress bar" for live map as follows:
* no load running and last run was ok: nothing is shown (no chang here)
* load running: progress indicator in orange (no change here)
* load requested but not triggered (short waiting period was introduced): progress indicator in blue
* load resulted in error: solid red progress indicator
* load resulted in overflow: solid gray progress indicator

This is meant as a discussion starting point, I would like to get feedback on how this "feels" in the wild.

ztNFny:
What is "overflow" in this context?

eddiemuc:
> What is "overflow" in this context?

Sorry, I need to explain stuff better...
"Overflow" means if the result from an online query is bigger than the limit we set (or to be more precise for technical reasons: to be exactly what we set as a limit). For example, when querying data from geocaching.com for a certain viewport and with filter, we limit results to 500. If we get 500 results, then it is likely that there are more caches than 500 in the requested area and that we are missing some (thus not displaying everything there is to the user). This is an "overflow".

--> I am open for better ideas on how to call this :-)

eddiemuc:
> Haven't looked at your implementation yet, just reading the above comment: for the case "error" and "partial result/overflow" I'd suggest a warning icon on map as already implemented for the 429 situation that the user can click to get more info. A red/grey progress bar is probably too hidden and without background knowledge not decipherable.

I have renamed "overflow" to "partial result" in code. Thanks for the suggestions.

I agree that the color scheme above is most likely not the end of GUI design-. However, keep in mind that today the user does not know at all whether an error or a partial result situation has occured. This is a first idea to give some visual indication to try this out. We may later decide whether to remove it again or how to visualize it. 

If the GUI discussion is a blocker for merge, then I will remove these changes again.

ztNFny:
> If the GUI discussion is a blocker for merge, then I will remove these changes again.

not a blocker at all, was just throwing around ideas.

I actually cherry-picked it now and gave it a test drive. In terms of GUI I'd suggest the following:
- load requested but not triggered: orange progress indicator in determinate mode (I assume this is a fixed delay of x seconds? didn't spot the code delaying it yet)
- load running: orange progress indicator in indeterminate mode
- load resulted in error: like 429 warning, red error icon
- load resulted in overflow: like 429 warning, half-filled circle icon

then the 429 icon becomes a "all map loading errors/warnings that the user needs to be informed about"-icon

eddiemuc:
> > If the GUI discussion is a blocker for merge, then I will remove these changes again.
> 
> not a blocker at all, was just throwing around ideas.
> 
> I actually cherry-picked it now and gave it a test drive. In terms of GUI I'd suggest the following:
> 
> * load requested but not triggered: orange progress indicator in determinate mode (I assume this is a fixed delay of x seconds? didn't spot the code delaying it yet)
> * load running: orange progress indicator in indeterminate mode
> * load resulted in error: like 429 warning, red error icon
> * load resulted in overflow: like 429 warning, half-filled circle icon
> 
> then the 429 icon becomes a "all map loading errors/warnings that the user needs to be informed about"-icon

I will remove GUI changes from PR and will then merge. GUI can be discussed separately then (if it is wanted at all)

ztNFny:
> I will remove GUI changes from PR and will then merge. GUI can be discussed separately then (if it is wanted at all)

Why? Please don't. Some UI is better than none and we can still tweak it later - will be much harder to re-invent it without a solid starting point.

eddiemuc:
> > I will remove GUI changes from PR and will then merge. GUI can be discussed separately then (if it is wanted at all)
> 
> Why? Please don't. Some UI is better than none and we can still tweak it later - will be much harder to re-invent it without a solid starting point.

"some UI" is already there, designed to be of little implementation effort and to improve later. 
Your suggestions are totally fair but will create an amount of work which would be better suited in a separate PR. If they are a condition for merge, it is better to remove UI changes for now.

ztNFny:
When mapping a list there seems to be bug in relation to isLiveMap(). Not 100% sure on the steps to reproduce yet, cache loading seems to depend on something else, but at least the first 2 steps are reproducable:
1. Open map, enable live
2. Open a list, map it
-> Settings.isLiveMap() = true -- but it's an offline list and live is disabled
3. Enable live mode -> no caches get loaded
4. Disable live mode
5. Enable live mode -> now caches get loaded

eddiemuc:
> When mapping a list there seems to be bug in relation to isLiveMap(). Not 100% sure on the steps to reproduce yet, cache loading seems to depend on something else, but at least the first 2 steps are reproducable:
> 
> 1. Open map, enable live
> 2. Open a list, map it
>    -> Settings.isLiveMap() = true -- but it's an offline list and live is disabled
> 3. Enable live mode -> no caches get loaded
> 4. Disable live mode
> 5. Enable live mode -> now caches get loaded

Couldn't reproduce with the PR. Is this not happening if you are on the master branch?

ztNFny:
> Is this not happening if you are on the master branch?

You're right, it's already like this on master. But a customization of mine relies on that, so it wasn't always like that...

Quick test option in https://github.com/cgeo/cgeo/commit/606a22701f2f8d9ba07b6c453504639bb50bfa08 - hold the live mode button to get the current isLiveMap() value. I'd have expected that it returns false when mapping a list - or is that a misunderstanding of how the map works?

eddiemuc:
> Quick test option in [606a227](https://github.com/cgeo/cgeo/commit/606a22701f2f8d9ba07b6c453504639bb50bfa08) - hold the live mode button to get the current isLiveMap() value. I'd have expected that it returns false when mapping a list - or is that a misunderstanding of how the map works?

I am not entirely sure, but I believe for the map to really "go live" (means: check for caches online)  both `Settings.isLiveMap()` must be true and the current `mapType's enableLiveMap()` method must also return true. The later is only the case for the "plain map" and the  targetcoords map. But not for `UMTT_List` which is used when offline lists are displayed.
@moving-bits is this interpretation correct?

moving-bits:
"live mode" on UnifiedMap is controlled by the value of `viewModel.transientIsLiveEnabled`:

On opening a map, this gets _initialized_ to the values delivered by `Settings.isLiveMap() && mapType.enableLiveMap()`, the latter meaning `type == UnifiedMapTypeType.UMTT_PlainMap || type == UnifiedMapTypeType.UMTT_TargetCoords`.

After that, the user can enable / disable live mode by tapping the "live mode" button in the action bar, which changes the value of `viewModel.transientIsLiveEnabled`, but also remembers the last setting by storing it to `Settings.isLiveMap`.

Issue #16382, opened 2024-11-24
Recognize (german) google maps coordinates format


geocachermgo:
### Tell us your idea!

With long click on Google maps it possible to copy the coordinates. The format is
45,4800641, 9,2105307
It's not possible to paste this clipboard content into a cgeo waypoint.

### What solution would you suggest?

The format #,#####, #,####
should be treated like #.#####, #.####

### Is there an alternative?

currently I paste into the waypoint note, replace , by . copy that and then it's recognized as coordinates.

geocachermgo:
How can I check if it's included in the nightly?

SammysHP:
@geocachermgo It isn't, otherwise this issue would be closed.

ztNFny:
> @ztNFny could you explain what was the reason for you to gave up with https://github.com/cgeo/cgeo/pull/16387? Just for documentation purposes, if someone else wants to take a look... 

It didn't work and i didn't understand why. The coordinates parser is black magic

ztNFny:
> @ztNFny  The regular expression for the longitude is not correct ... the backslashes before the square brackets are wrong...
> Shall I add a commit with the according changes?

I'm not working on this in any way.

murggel:
> > @ztNFny  The regular expression for the longitude is not correct ... the backslashes before the square brackets are wrong...
> > Shall I add a commit with the according changes?
> 
> I'm not working on this in any way.

ok, then I make a new PR...

murggel:
ok, the problem is not the reg-expression, but that the "comma" is used as separator and as "dot", so it is difficult to distinguish between the doubled usage / interpretation

murggel:
> ok, the problem is not the reg-expression, but that the "comma" is used as separator and as "dot", so it is difficult to distinguish between the doubled usage / interpretation

It is not so easy to enhance the `GeopointParser` for that problem without breaking existing or creating unwanted behaviour, but probably we could add some special logic in `hasClipboardCoordinates`.

moving-bits:
See #13840 for similar request (Russian locale in that request)

geocachermgo:
At https://github.com/cgeo/cgeo/blob/45ea14ba4a37227e5ba249a904320fcbdf1a2f2d/main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesInputDialog.java#L311

I found:
new Geopoint(StringUtils.defaultString(ClipboardUtils.getText()));

where are the patterns?

murggel:
[...]

PR #16486, opened 2025-01-02
add parser for DD,DDDDDDD format

ztNFny:
add parser for DD,DDDDDDD format used by e.g. Google Maps in German/Russian

fixes https://github.com/cgeo/cgeo/issues/16382

This will require some good testing though, the coordinate format seems to be prone to false positives - hence marking as Feedback required.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/GeopointParser.java
+++ b/main/src/main/java/cgeo/geocaching/location/GeopointParser.java
@@ -417,7 +417,7 @@ public class GeopointParser {
 
         //                                        (       1       )
         private static final String STRING_LON = "(-?(\\d++),(\\d++))\\b°?";
-        private static final String STRING_SEPARATOR = "[^\\w'′\"″°.=-]*";
+        private static final String STRING_SEPARATOR = ", ";
         private static final Pattern PATTERN_LAT = Pattern.compile(STRING_LAT, Pattern.CASE_INSENSITIVE);
         private static final Pattern PATTERN_LON = Pattern.compile(STRING_LON, Pattern.CASE_INSENSITIVE);
         private static final Pattern PATTERN_LATLON = Pattern.compile(STRING_LAT + STRING_SEPARATOR + STRING_LON, Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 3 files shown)


murggel:
I guess for the copy-format you need a test for "47,648883, 122,348067" - the problem is the comma between the latitude and longitude

ztNFny:
> I guess for the copy-format you need a test for "47,648883, 122,348067" - the problem is the comma between the latitude and longitude

It works just fine for "47,648883, 122,348067", the problem wasn't the comma between the values but parsing 47,64 as double instead of 47.64

murggel:
> > I guess for the copy-format you need a test for "47,648883, 122,348067" - the problem is the comma between the latitude and longitude
> 
> It works just fine for "47,648883, 122,348067", the problem wasn't the comma between the values but parsing 47,64 as double instead of 47.64

But your test works with current implementation as well (there is a function, which swaps dot and comma and tries with both variants)
So the current not working difference is the comma inbetween.

At least that were my findings...

ztNFny:
I didn't spent much time on the test class so far (less than 2 min tbh). But I've verified with manual tests that clipboard now recognises those coords from German google maps.

Due to the comma-dot-swapping functions there are 2 ways to solve this:
1. Take care of the Lat-Lon-seperator, let the swapping take care of the decimal separators - your suggestion
2. Parse decimal comma - this PR

ztNFny:
With some tweaks I'm down to 1 false positive: "6, 12, 16, 29" gets internally converted to "6,12, 16,29" by getParseInputs() which is then recognized as coordinates

murggel:
As discussed in the todays dev-meeting, I tried to extend the parser to consider at least a minimum amount of digits after the decimal-comma.  And with some minor modifications to your regex I got all tests working.

@ztNFny Shall I create a separate PR or commit the changes to this PR? Reverting is possible at every time though.

ztNFny:
Feel free to push them here

murggel:
> Feel free to push them here

I just pushed my changes

murggel:
Considering only 5 decimal places for DegDecParser  has probably to many side effects for already existing notes

murggel:
Failing tests does not seem related to this PR

eddiemuc:
Can/should both #16486 and #16491 be merged or just one of them (which one?) or none?

murggel:
> Can/should both #16486 and #16491 be merged or just one of them (which one?) or none?

If #16486 will be nerged, #16491 will be obsolete and no need to merge 

With the last changes (at least 5 decimal places) the new parser in #16486 does not have to many unwanted matches and I am fine with  using it.

PR #16489, opened 2025-01-02
fix MainActivity dropdown background color

ztNFny:
background color of MainActivity dropdown was wrong (layout is reused in search and main activity which require slightly different colors)


ztNFny:
retest this please

PR #16496, opened 2025-01-02
fix #16494: mark edited logs as friend log

### LINKED ISSUE #16494: Edited Log disappears from "friends" section
[Ticket #893601](https://cgeosupport.droescher.eu/scp/tickets.php?id=8934)
version: 2024.12.01

When a user creates a log entry, then edits this log entry (w/o refreshing the cache) then the log entry is not visible any more in the "friends" section. I can reproduce this problem.

Steps to reproduce
* Create and send a new log entry for a cache -> log entry shows up correctly in both "common" and "friends log sections 
* DO NOT REFRESH THE CACHE!
* open the created entry for edit, edit it and resend it -> BUG: log entry has now disappeared from "friends" section (but not from "common" section)
* when refreshing the cache, the log entry appears correctly again also in the "friends" section

eddiemuc:
fix #16494: mark edited logs as friend log


jakedot:
lgtm

PR #16498, opened 2025-01-03
rel #15939: add inventory count filter

### LINKED ISSUE #15939: Trackable text search
### Tell us your idea!

It would be great to have a filter that allows searching for specific text in the Name or Description of trackables stored in the cache. In particular, I would use it to search for TB hotels containing geocoins, as they often contain the words "coin" or "geocoin" in their names.

It could look similar to the "Cache Name" filter but look for caches that contain trackables with a specific text in their Name, Goal, or Details.

### Is there an alternative?

I am not aware of alternatives except manually checking all caches with trackables.

### Additional Information

_No response_

murggel:
## Description

Add filter for inventory count

## Related issues

rel #15939


eddiemuc (approved):
LGTM, thanks for the contribution!

ztNFny:
is the term "inventory" clear enough? I wouldn't know what it's supposed to mean (but I'm not really into trackables). Maybe "trackable count" would be a better description?

murggel:
> is the term "inventory" clear enough? I wouldn't know what it's supposed to mean (but I'm not really into trackables). Maybe "trackable count" would be a better description?

You are  correct - the status filter is named "has trackables" - probably we should rename the new one to "trackables count"

PR #16501, opened 2025-01-03
Fix failing tests: Consider new filters in test

murggel:
## Description

The 2 new filters "LIST_ID" and "VIEWPORT" (https://github.com/cgeo/cgeo/commit/857a61bfc44aeef83cdd6e395dff780b021cf961) has to be considered in getFilledInstance.

## Additional context

The intention of the assert is, that all filters are considered in the generic filled test (at least, if not considered, then on purpose)


eddiemuc (approved):
Oh that was me...

Lgtm, thanks for the contribution

Issue #16504, opened 2025-01-04
waypoints disappearing in single cache view


ztNFny:
@eddiemuc just noticed another issue likely caused by your PR: open a cache with parking waypoints, map it - parking wps disappear after about a second.
If you're not able to reproduce easily i'll do a proper report later once home.

_Originally posted by @ztNFny in https://github.com/cgeo/cgeo/issues/14931#issuecomment-2571262854_

Single cache, just map it, parking wps disappear after 1s. No further steps like enabling live map required.

https://github.com/user-attachments/assets/d66fe652-c154-4918-ac46-76a84b101db9
Quick video

ztNFny:
Testing in emulator I can NOT reproduce this - but on my real device it's fully reproducable. Any idea what could influence this?

eddiemuc:
> Testing in emulator I can NOT reproduce this - but on my real device it's fully reproducable. Any idea what could influence this?

Some questions:
- would the cache in question be filtered out by your offline or your live filter (if you weren't in single cache view)?
- the cache from your video is archived in reality but not in your local copy. This might be related

ztNFny:
No filter and happens with all caches.

eddiemuc:
Do the waypoints reappear if you move the map a bit and then wait for like 1-2 seconds?

ztNFny:
> Do the waypoints reappear if you move the map a bit and then wait for like 1-2 seconds?

No

I've now rebased my build onto master and am still able to repro. Importing the settings from my phone into the emulator makes the issue reproducable - now it's just a few hundret settings to figure out which is causing this...

ztNFny:
Found the setting, was easier than expected:
Settings -> Map Content & Behaviour -> Show waypoints = 0

This setting until now affected only multi-cache views, not mapping a single cache.

ztNFny:
similar issue exists when opening a cache and mapping a single waypoint of it

for the first issue https://github.com/cgeo/cgeo/blob/f39d375311d6dca15352610d22d723de718d61a7/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java#L590 needs to take into account the type of map - for mapping a single geocache it needs to be true

ztNFny:
The issue of disappearing waypoints when mapping them is indepentent of waypoint_threshold_default settings. To reproduce:
1. Open a cache with waypoints
2. Click the compass near a waypoint to map it
3. after 1 second the waypoint disappears from map 
[...]

PR #16508, opened 2025-01-04
Live map status visuals

ztNFny:
Improve live map status visuals:
* replace grey / blue bar with status indicator below compass
* merge status indicator with 429 warning
* REQUESTED state uses determinate progress bar

image
image
image
image
image

PR includes https://github.com/cgeo/cgeo/pull/16503

PR #16509, opened 2025-01-04
fix disappearing waypoints in single-cache-mode

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/16504

@eddiemuc  feel free to replace it with a different solution in case you don't like this one


eddiemuc:
Good fix! I will take it over in my other PR so we avoid merge conflicts. I hope thats ok

ztNFny:
if you want you can also cherry-pick the visual stuff (https://github.com/cgeo/cgeo/pull/16508) into your PR - it's only marked as WIP due to the dependency on yours

Issue #16463, opened 2024-12-31
"cacti nearby" attribute not supported


ztNFny:
A legacy attribute "cacti nearby" that was only ever used 4 times with 1 surviving cache https://coord.info/GC684 exists. This attribute is not supported by c:geo.

As noone cared for many years and there's only 1 such cache out there I'm wondering whether it's worth adding it?

moving-bits:
> As noone cared for many years and there's only 1 such cache out there I'm wondering whether it's worth adding it?

probably not :man_shrugging:

jakedot:
I'm for adding it for completeness' sake.

PR #16513, opened 2025-01-04
only show "no"-version of attributes where available

ztNFny:
Only a few attributes on GC are available in Yes/No versions and other connectors don't support No at all.
Update the list of attributes to reflect availability and ddjust the attribute filter to display the "no" versions only for those attributes where one is available.

Also add the cacti attribute, fixes https://github.com/cgeo/cgeo/issues/16463


ztNFny:
retest this please

PR #16522, opened 2025-01-05
UnifiedMap refactoring: put mapType into ViewModel

eddiemuc:
UnifiedMap refactoring: put mapType into ViewModel


eddiemuc:
@moving-bits : I am refactoring UnifiedMapActivity and have some more complex questions. I figure you might know the answers :-)

**Calculation of `compatibilityMapMode`** 

Looking at method `setMapModeFromMapType`, the `compatibilityMapMode` solely depends on `mapType.type`. Can it be made into a property of the enum? 
(Note: the current PR is already doing this)

**Storing `mapType `and `mapType.filterContext`**

In `UnifiedMapActivity`, `mapType `and `mapType.filterContext` are stored (`onSaveInstanceState`) and restored (`onCreate`) in two separate states right now (`BUNDLE_MAPTYPE`  and `BUNDLE_FILTERCONTEXT` ). Why is it not enough to store only the mapType (which includes of course also its field `mapType.filterContext`)?
(Note: the current PR is already doing this)

**Meaning of parameter `setDefaultCenterAndZoom` of method `reloadCachesAndWaypoints` and member variable `overridePositionAndZoom`**

I tried to figure out how the above things work together and have the following impression:
* both `setDefaultCenterAndZoom` and `overridePositionAndZoom` are solely important for how method `reloadCachesAndWaypoints` behave
* they are uses to differentiate three cases:
   * Initial call when activity is created (`setDefaultCenterAndZoom` is true in this case)
   * Initial call when activity is resumed after is was paused (eg after a screen rotation; `overridePositionAndZoom` is true in this case
   * All other cases where the map simply needs to be refreshed (eg after a cache was updated; both `setDefaultCenterAndZoom` and `overridePositionAndZoom` are false in this case)
   * There is never a call to the method where both  `setDefaultCenterAndZoom` and `overridePositionAndZoom` are true

Is my assesment correct? Where is it wrong? (It might be possible to replace the members with an enum)

**Meaning of `UnifiedMapState`**

There is a member var `mapState` of type  `UnifiedMapState`.
* The member `transientIsLiveEnabled` of type `UnifiedMapState` doesn't seem to be used anywhere. Can it be removed?
* Type `UnifiedMapState` holds the map center. This seems somehow superfluous when also the current Map Viewport is available. Is there a reason to store the Map Center separately in an own type?
* `UnifiedMapState` seems to hold only position parameters. I wonder how it got its name?

I am thinking about creating a type `UnifiedMapPosition` holding the Map Viewport (and thus implicitely the maps center) and the Maps current zoom level. Then one could replace current variable viewport/viewportIdle of UnifiedMapModel with an instance of  `UnifiedMapPosition` and refresh it accordingly. Is there anything I am overseeing here?

moving-bits:
> @moving-bits : I am refactoring UnifiedMapActivity and have some more complex questions. I figure you might know the answers :-)

I'll give it a try...

> **Calculation of `compatibilityMapMode`**
> 
> Looking at method `setMapModeFromMapType`, the `compatibilityMapMode` solely depends on `mapType.type`. Can it be made into a property of the enum? (Note: the current PR is already indirectly doing this because mapType is moved to `UnifiedMapModel`)

compatibility mode is used solely to retrieve the correct "last zoom level", as those methods expect an old map's map mode, not a UnifiedMap's mapType.

> **Storing `mapType `and `mapType.filterContext`**
> 
> In `UnifiedMapActivity`, `mapType `and `mapType.filterContext` are stored (`onSaveInstanceState`) and restored (`onCreate`) in two separate states right now (`BUNDLE_MAPTYPE` and `BUNDLE_FILTERCONTEXT` ). Why is it not enough to store only the mapType (which includes of course also its field `mapType.filterContext`)? (Note: the current PR is already doing this)

This is probably just copied from the old map, which also stores `filterContext`.
mapType alone is not sufficient in all cases, though, to determine the filter context, see the constructors in `UnifiedMapType.java`.

> **Meaning of parameter `setDefaultCenterAndZoom` of method `reloadCachesAndWaypoints` and member variable `overridePositionAndZoom`**
> 
> I tried to figure out how the above things work together and have the following impression:
> 
>     * both `setDefaultCenterAndZoom` and `overridePositionAndZoom` are solely important for how method `reloadCachesAndWaypoints` behave

Yes, I needed a way to fine-tune subsequent calls to `reloadCachesAndWaypoints`. See also the comment: `to preserve those on config changes in favour to mapType defaults`

> * they are uses to differentiate three cases:
>       
>    * Initial call when activity is created (`setDefaultCenterAndZoom` is true in this case)
>    * Initial call when activity is resumed after is was paused (eg after a screen rotation; `overridePositionAndZoom` is true in this case
>    * All other cases where the map simply needs to be refreshed (eg after a cache was updated; both `setDefaultCenterAndZoom` and `overridePositionAndZoom` are false in this case)
>    * There is never a call to the method where both  `setDefaultCenterAndZoom` and `overridePositionAndZoom` are true
> 
> Is my assesment correct? Where is it wrong? (It might be possible to replace the members with an enum)

Yes, AFAICS

> **Meaning of `UnifiedMapState`**
> 
> There is a member var `mapState` of type `UnifiedMapState`.
> 
> * The member `transientIsLiveEnabled` of type `UnifiedMapState` doesn't seem to be used anywhere. Can it be removed?

looks so

> * Type `UnifiedMapState` holds the map center. This seems somehow superfluous when also the current Map Viewport is available. Is there a reason to store the Map Center separately in an own type?

Back when it was introduced it used to hold map center + zoom level (similar to what `NewMap.prepareMapState()` does), but this may be changed.

> * `UnifiedMapState` seems to hold only position parameters. I wonder how it got its name?

This is derived from `NewMap.prepareMapState()`, although UnifiedMap uses live data for some of those fields.

> I am thinking about creating a type `UnifiedMapPosition` holding the Map Viewport (and thus implicitely the maps center) and the Maps current zoom level. Then one could replace current variable viewport/viewportIdle of UnifiedMapModel with an instance of `UnifiedMapPosition` and refresh it accordingly. Is there anything I am overseeing here?

LGTM on a quick glance

eddiemuc:
PR is ready now imho

PR #16524, opened 2025-01-05
Show lastUsedFilterName in filter-dialog

### LINKED ISSUE #16512: Make "filter" button in map view toggle between last used filter on/off
### Tell us your idea!

Probably most people will only toggle between their default filter (e.g. hiding founds) and no filter. 

Currently this is a little inconvenient because it requires quite a few clicks.

Make "filter" button in map view toggle between **last used** filter on/off.

To get all filter options, use long press on the button.

### Is there an alternative?

_No response_

### Additional Information

_No response_

murggel:
## Description

Show the last used filter in filter-selection-dialog, if no filter is active.
So it will be easier, to disable temporarily the current filter and enable the one before.

## Possible implementations
Fixed name is currently the "active" variant

## Related issues

fix #16512

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
+++ b/main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
@@ -27,6 +27,7 @@ public class GeocacheFilterContext implements Parcelable {
 
     private FilterType type;
     private GeocacheFilter filter;
+    private GeocacheFilter previousFilter = null;
 
     public GeocacheFilterContext(final FilterType type) {
         setType(type);
@@ -56,11 +57,23 @@ public class GeocacheFilterContext implements Parcelable {
     }
 
     public void set(final GeocacheFilter filter) {
+        if (filter != null) {
+            final GeocacheFilter currentFilter = get();
+            final boolean sameFilter = currentFilter.filtersSame(filter) && currentFilter.getName().equals(filter.getName());
+            if (currentFilter.isFiltering() && !sameFilter) {
+                previousFilter = currentFilter;
+            }
+        }
+
         if (type == FilterType.TRANSIENT) {
             this.filter = filter == null ? GeocacheFilter.createEmpty() : filter;
-        } else {
+        } else if (filter != null) {
             Settings.setCacheFilterConfig(type.name(), filter.toConfig());
         }
+    }
+
+    public GeocacheFilter getPreviousFilter() {
+        return previousFilter;
     }
 
     @NonNull
@@ -76,10 +89,14 @@ public class GeocacheFilterContext implements Parcelable {
         public GeocacheFilterContext createFromParcel(final Parcel in) {
             final FilterType type = (FilterType) in.readSerializable();
             final String filterConfig = in.readString();
+            final String previousFilterConfig = in.readString();
 
             final GeocacheFilterContext ctx = new GeocacheFilterContext(type);
             if (filterConfig != null) {
                 ctx.set(GeocacheFilter.createFromConfig(filterConfig));
+            }
+            if (previousFilterConfig != null) {
+                ctx.previousFilter = GeocacheFilter.createFromConfig(previousFilterConfig);
             }
             return ctx;
         }
@@ -99,5 +116,6 @@ public class GeocacheFilterContext implements Parcelable {
     public void writeToParcel(final Parcel dest, final int flags) {
         dest.writeSerializable(type);
         dest.writeString(filter == null ? null : filter.toConfig());
+        dest.writeString(previousFilter == null ? null : previousFilter.toConfig());
     }
 }
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
@@ -510,7 +510,7 @@ public class SimpleDialog {
         if (model.selectSetSupplier != null) {
             this.setNeutralButton(model.selectSetActionText);
         }
-        if (!selectionConfirmedViaButton) {
+        if (!selectionConfirmedViaButton && buttonClickAction == null) {
             //remove "negative/positive" buttons
             setPositiveButton(null);
             setNegativeButton(null);
--- a/main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
@@ -13,9 +13,11 @@ import cgeo.geocaching.ui.ViewUtils;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 
 import android.app.Activity;
+import android.content.DialogInterface;
 import android.graphics.Typeface;
 import android.os.Handler;
 import android.text.style.StyleSpan;
+import android.util.Pair;
 import android.view.View;
 import android.widget.TextView;
 
@@ -42,6 +44,17 @@ public class FilterUtils {
                 filteredList, true);
     }
 
+    @NonNull
+    private static Pair<TextParam, GeocacheFilter> getLastFilterActionForContext(@NonNull final GeocacheFilterContext filterContext) {
+        final GeocacheFilter previousFilter = filterContext.getPreviousFilter();
+        final boolean hasValidLastFilter = null != previousFilter && previousFilter.isFiltering();
+
+        if (hasValidLastFilter) {
+            return new Pair(TextParam.id(R.string.caches_filter_select_last), previousFilter);
+        }
+        return new Pair(null, null);
+    }
+
     public static <T extends AbstractActivity & FilteredActivity> boolean openFilterList(final T filteredActivity, final GeocacheFilterContext filterContext) {
         final List<GeocacheFilter> filters = new ArrayList<>(GeocacheFilter.Storage.getStoredFilters());
         final boolean isFilterActive = filterContext.get().isFiltering();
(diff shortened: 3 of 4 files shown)


ztNFny:
What if an unstored filter was previously used?

murggel:
> What if an unstored filter was previously used?

Then no "last selection" is possible


image

ztNFny:
Looking at the original feature request I'd guess that it's not so much about named filters but - well, unsaved ones (whatever to call them). Restoring a named filter is already very easy, just select it from the list, for that this PR is only a tiny improvement.

murggel:
> Looking at the original feature request I'd guess that it's not so much about named filters but - well, unsaved ones (whatever to call them). Restoring a named filter is already very easy, just select it from the list, for that this PR is only a tiny improvement.

Yes, a tiny improvement, but with a bunch of filters a easy one.

But you are right, I can check, whether I can restore a unsaved previous filter (so not only the filter-name has to be stored, but the complete filter).

Question would be, how to name it in the dialog then - but then probably only "last selection" without a name...

murggel:
Next question would be 
Only toggle if no filter is active or show always the option to switch to the last active filter?

murggel:
> Question would be, how to name it in the dialog then - but then probably only "last selection" without a name...

Probably "Last filter" would be better here then, because it doesn't have to be a "selection"

ztNFny:
You could store a key value pair of name and filter. If name is not null show the name.

ztNFny:
In terms of ui, is "last filter" correct? Sounds a bit strange to me, "previous filter" feels more correct - but I'm not native English. Chatgpt - yes yes ... - thinks previous: 

The term **"previous filter"** is more correct and appropriate in this context. Here's why:

### 1. **"Previous" Implies Chronological Order**
   - The word **"previous"** specifically refers to something that came immediately before the current state or action. In your case, the "previous filter" clearly communicates that it is the filter the user had selected just before the current one.
   - Example: If the user applied "Filter A" and then switched to "Filter B," the "previous filter" would be "Filter A."

### 2. **"Last" Can Be Ambiguous**
   - The word **"last"** often implies something final or ultimate. In some contexts, it might be interpreted as the most recent filter, but it could also be taken to mean the final filter in a sequence or the last filter the user will ever apply.
   - Example of ambiguity:
     - "Last filter" could mean the one used just before the current one (correct in this case).
     - But it could also mean the filter applied at the very end of a process or session (less relevant here).

### 3. **Consistency with User Interface Terminology**
   - In user interface design, "previous" is widely used to refer to the state, action, or selection immediately preceding the current one. For example:
     - "Previous page" in pagination refers to the page before the current one.
     - "Previous option" refers to the choice made before the current one.
   - Using "previous filter" aligns with this well-understood convention, reducing potential confusion.

### Conclusion:
Use **"previous filter"** to refer to the filter selected immediately before the current one. It is clear, precise, and aligns with standard terminology in software user interfaces.

murggel:
> You could store a key value pair of name and filter. If name is not null show the name.

I store the filter, which may have a name - then I can show the name - will check for that (but not yet any more ;-) )

murggel:
> ### Conclusion:
> 
> Use **"previous filter"** to refer to the filter selected immediately before the current one. It is clear, precise, and aligns with standard terminology in software user interfaces.

So the same should be valid for the last list-selection, correct?


Shall we rename it then?

ztNFny:
Probably. But as I'm not a native speaker I'm not 100% certain about it. Let's discuss?

murggel:
> Probably. But as I'm not a native speaker I'm not 100% certain about it. Let's discuss?

I am not native as well... but as more as I think about it, "previous selection" sounds misleading to me, I would associate the previous list-selection of that specific cache for me....
So I would leave it as it is and use only "previous filter" - that sounds reasonable for me..

eddiemuc:
> > You could store a key value pair of name and filter. If name is not null show the name.
> 
> I store the filter, which may have a name - then I can show the name - will check for that (but not yet any more ;-) )

I suggest to add this to the `geocachefiltercontext`. Have a second method there to get the last/previous filter and store the last/previous filter there on `set` if it differs from the current one (use `GeocacheFilter.filtersSame` to compare)

Should the saving of the last/previous filter survive a map restart or even a c:geo restart? Then it must be stored in settings

eddiemuc:
> I suggest to add this to the `geocachefiltercontext`. Have a second method there to get the last/previous filter 

@murggel I saw you did just that already :-) Sorry, I should have studied your PR before my previous post

murggel:
> Should the saving of the last/previous filter survive a map restart or even a c:geo restart? Then it must be stored in settings
> 

I would not store it in settings yet, for me temporary is sufficient. 

If needed we can add that later

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -40,6 +41,21 @@ public static void openFilterActivity(final Activity activity, final GeocacheFil
                 filteredList, true);
     }
 
+    @Nullable
+    private static Pair<Pair<TextParam, TextParam>, GeocacheFilter> getNeutralActionForContext(final GeocacheFilterContext filterContext) {
```
eddiemuc, line 45: To transfer three parameters use class  `ImmutableTriple` instead of nested `Pairs`

Context-wise: if I understand code correctly, then "last/previous filter" choice is only given when no filter is currently selected. Is this correct? Why not offering this option always?

eddiemuc (commented):
Thanks for your contributing, I have some comments/questions though

main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
```diff
@@ -56,13 +57,18 @@ public GeocacheFilter get() {
     }
 
     public void set(final GeocacheFilter filter) {
+        lastUsedFilter = get();
```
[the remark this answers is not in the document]
murggel: > Only do this if new filter differs from current one. At least do `if(get() != filter)`.
> 

Good point...

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
murggel: > To transfer three parameters use class  `ImmutableTriple` instead of nested `Pairs`


Good point, didn't had that in mind


> 
> Context-wise: if I understand code correctly, then "last/previous filter" choice is only given when no filter is currently selected. Is this correct? Why not offering this option always?
> 
> 
See comment 
https://github.com/cgeo/cgeo/pull/16524#issuecomment-2571794976

main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
```diff
@@ -56,13 +57,18 @@ public GeocacheFilter get() {
...
         if (type == FilterType.TRANSIENT) {
             this.filter = filter == null ? GeocacheFilter.createEmpty() : filter;
         } else {
             Settings.setCacheFilterConfig(type.name(), filter.toConfig());
         }
     }
 
+    public GeocacheFilter getLastUsedFilter() {
```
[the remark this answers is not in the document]
murggel: Renamed it to `getPreviousFilter`

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
murggel: Used `ImmutableTriple`

main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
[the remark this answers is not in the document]
murggel: Has some weird behaviour, if I use `GeocacheFilter.filtersSame` which I don't understand yet - so for a first version only do `if(get() != filter)`.

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
murggel: I will have a look into showing it always

murggel: To get it work, I had to avoid that the negative button is removed automatically...

main/src/main/java/cgeo/geocaching/filters/core/GeocacheFilterContext.java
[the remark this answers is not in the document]
eddiemuc: There was a bug in GeocacheFilter.filtersSame, I fixed it meanwhile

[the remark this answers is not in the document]
murggel: Ok, I will try again

main/src/main/java/cgeo/geocaching/utils/FilterUtils.java
```diff
@@ -55,27 +68,33 @@ public static <T extends AbstractActivity & FilteredActivity> boolean openFilter
...
         }
+
+        final SimpleDialog.ItemSelectModel<GeocacheFilter> model = getGroupedFilterList(filters);
+
+        final Pair<TextParam, GeocacheFilter> lastFilterButton = getLastFilterActionForContext(filterContext);
+        SimpleDialog.of(filteredActivity)
+                .setTitle(isFilterActive ? TextParam.id(R.string.cache_filter_storage_select_clear_title) : TextParam.id(R.string.cache_filter_storage_select_title))
+                .setPositiveButton(null)
+                .setNeutralButton(isFilterActive ? TextParam.id(R.string.cache_filter_storage_clear_button) : null)
+                .setNegativeButton(lastFilterButton.first)
```
murggel, line 88: I used the negative button for `Previous filter`, but to get it work, that both, negative and neutral buttons are shown (but no positive button), I had To get it work, I had to check for `buttonClickAction` in SimpleDialog.

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -510,7 +510,7 @@ private <T> void selectItems(final ItemSelectModel<T> options, final Consumer<Se
         if (model.selectSetSupplier != null) {
             this.setNeutralButton(model.selectSetActionText);
         }
-        if (!selectionConfirmedViaButton) {
+        if (!selectionConfirmedViaButton && buttonClickAction == null) {
```
murggel, line 513: I used the negative button for `Previous filter`, but to get it work, that both, negative and neutral buttons are shown (but no positive button), I had To get it work, I had to check for `buttonClickAction` in SimpleDialog.

murggel:
@eddiemuc ready for further review

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
eddiemuc: I guess that's ok because such a case is pretty rare.
In fact does it really make sense to have a dialog with negative button but no positive one?

murggel: I used the negative button as a second neutral one, not as the real intention of positive / negative meaingn

Issue #15992, opened 2024-08-20
UDCs should get globally-unique IDs (to help sharing UDCs via GPX-Export/Import)


MagpieFourtyTwo:
### Describe your problem!

Occasionally we exchange UserDefinedCaches (UDC) via GPX to communicate a meeting point or the address of a cachemate ... which is way more comfortable than sending coordinates only, cause ``export/share/import`` is way easier than ``copy coords/open messenger/paste/send/copy/create UDC/name it/paste coords`` ... you got the story, I guess.  ;)  Apart from that all the members in the group have the same name and everybody knows what we're talking about.  Plus you can add parking lots, trailheads, remarks in personal notes etc. pp ...

However, this method always seemed to work perfectly fine - till yesterday, when something really strange happened:
My mate sent me the GPX of an UDC, containing name, coords, access point etc. pp. of some friends we want to go to.
And I imported the GPX ... but at the end the imported UDC was 300 km south of the region where I expected it to be. 😲

I then tried to reconstruct what had happened and realized that I have had a UDC with the same ID like the UDC she had sent me - and since IDs of UDCs are not unique, fate took its course ...

- On import of the GPX c:geo simply "renamed" the existing UDC - without notification!
  (Edit: Even worse: Just realized that I even deleted this "previously existing" UDC months ago!)
- Moreover all already existing waypoints with matching names got the imported coords
- if an already existing matching WP had a note, it remained unchanged (although it makes no sense anymore, with new context and coords), except if the imported one had a note, too - then the note of the exiting WP got overwritten
- all previously existing waypoints, which had no "same name fellow" in the GPX remained unchanged
- at least happy to see that an existing PN remains untouched (while a PN in the GPX gets imported, if the previously existing UDC did not have a PN yet)

TMALSS:  c:geo *added* all new information, and "updated" (i. e. overwrote) already existing stuff, leaving the rest untouched.
Which I think is not only rude but even pretty pointless, cause this results in a complete mess of mixed old and new data,

But the best is still to come:  Although c:geo changed the name of the previously existing UDC, *it did not change it's coordinates*, so at the end we had a new name with old coords!

Which is, compared to the previous observations, at least inconsequent - and makes the import completely incomprehensible and enigmatic, plus leaving the poor c:geo user in complete unawareness of the fatal things that have just happened to their data.

This should be fixed.  Cause confusion is just as inevitable as annoyance if user's data is corrupted by an import.

### How to reproduce?

You can simulate the above if you ...

[...]

PR #16528, opened 2025-01-06
generate random IDs for UDCs

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/15992 by generating a random ID for UDCs in range ZZ[0-9A-Z]{4}. 1,8 million possibilities should be random enough 😉

also some cleanup that was supposed to be done in 2020 :-)
And a bugfix for getting the default name and description that was previously broken (due to the provided string being "" instead of null)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
@@ -46,7 +46,7 @@ public class InternalConnector extends AbstractConnector implements ISearchByGeo
 
     // prefix - must not contain regexp characters
     public static final String PREFIX = "ZZ";
-    private static final String GEOCODE_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    private static final String GEOCODE_CHARS = "0123456789ABCDEFGHJKLMNPQRTUVWXYZ";
     private static final int GEOCODE_LENGTH = 4;
 
     // predefined id (without prefix) and geocode (with prefix) for certain special caches:
--- a/main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
@@ -34,6 +34,7 @@ import androidx.annotation.Nullable;
 
 import java.util.EnumSet;
 import java.util.HashSet;
+import java.util.Random;
 import java.util.Set;
 import java.util.regex.Pattern;
 
@@ -45,10 +46,11 @@ public class InternalConnector extends AbstractConnector implements ISearchByGeo
 
     // prefix - must not contain regexp characters
     public static final String PREFIX = "ZZ";
+    private static final String GEOCODE_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    private static final int GEOCODE_LENGTH = 4;
 
     // predefined id (without prefix) and geocode (with prefix) for certain special caches:
-    public static final int ID_HISTORY_CACHE = 0;
-    public static final String GEOCODE_HISTORY_CACHE = geocodeFromId(ID_HISTORY_CACHE);
+    public static final String GEOCODE_HISTORY_CACHE = "ZZ0";
 
     // store into UDC list
     public static final int UDC_LIST = -1;
@@ -211,15 +213,14 @@ public class InternalConnector extends AbstractConnector implements ISearchByGeo
      * creates a new cache with the given id, if it does not exist yet
      *
      * @param context       context in which this function gets called
-     * @param id            internal (numeric) id of the cache
+     * @param geocode       internal id of the cache
      * @param name          cache's name (or null for default name)
      * @param description   cache's description (or null for default description)
      * @param assignedEmoji cache's assigned emoji (or 0 for default cache type icon)
      * @param geopoint      cache's current location (or null if none)
      * @param listId        cache list's id, may be NEW_LIST
      */
-    protected static void assertCacheExists(final Context context, final long id, @Nullable final String name, @Nullable final String description, final int assignedEmoji, @Nullable final Geopoint geopoint, final int listId) {
-        final String geocode = geocodeFromId(id);
+    protected static void assertCacheExists(final Context context, final String geocode, @Nullable final String name, @Nullable final String description, final int assignedEmoji, @Nullable final Geopoint geopoint, final int listId) {
         if (DataStore.loadCache(geocode, LoadFlags.LOAD_CACHE_OR_DB) == null) {
 
             int newListId = listId;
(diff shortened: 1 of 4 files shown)


main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
```diff
@@ -45,10 +46,11 @@ public class InternalConnector extends AbstractConnector implements ISearchByGeo
 
     // prefix - must not contain regexp characters
     public static final String PREFIX = "ZZ";
+    private static final String GEOCODE_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
eddiemuc, line 49: Suggest to leave out I, O and S (to avoid confusion with 1, 0 and 5)

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
```diff
@@ -45,10 +46,11 @@ public class InternalConnector extends AbstractConnector implements ISearchByGeo
...
+    private static final int GEOCODE_LENGTH = 4;
```
eddiemuc, line 50: Suggest to use 6

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
```diff
@@ -211,15 +213,14 @@ protected static int getUdcListId(final Context context) {
...
+     * @param geocode       internal id of the cache
      * @param name          cache's name (or null for default name)
      * @param description   cache's description (or null for default description)
      * @param assignedEmoji cache's assigned emoji (or 0 for default cache type icon)
      * @param geopoint      cache's current location (or null if none)
      * @param listId        cache list's id, may be NEW_LIST
      */
-    protected static void assertCacheExists(final Context context, final long id, @Nullable final String name, @Nullable final String description, final int assignedEmoji, @Nullable final Geopoint geopoint, final int listId) {
-        final String geocode = geocodeFromId(id);
+    protected static void assertCacheExists(final Context context, final String geocode, @Nullable final String name, @Nullable final String description, final int assignedEmoji, @Nullable final Geopoint geopoint, final int listId) {
```
eddiemuc, line 223: I believe method can be made private (this is unrelated to the issue but would be nice anyway :-))

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
```diff
@@ -275,9 +270,22 @@ public static void assertHistoryCacheExists(final Context context) {
      * @return geocode      geocode of the newly created cache
      */
     public static String createCache(final Context context, @Nullable final String name, @Nullable final String description, final int assignedEmoji, @Nullable final Geopoint geopoint, final int listId) {
-        final long newId = DataStore.getNextAvailableInternalCacheId();
```
eddiemuc, line 278: I believe DataStore.getNexytAvailableInternalCacheId() can be deleted too.

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
```diff
@@ -275,9 +270,22 @@ public static void assertHistoryCacheExists(final Context context) {
...
-        assertCacheExists(context, newId, name, description, assignedEmoji, geopoint, listId);
-        return geocodeFromId(newId);
```
eddiemuc, line 280: I believe method geocodeFromId() can be deleted too. This will also lead to a unit test in need to be removed

eddiemuc (commented):
Some small comments, also Codacy has an error. 
Thanks for the contribution

ztNFny:
The reason I went with ZZ[0-9A-Z]{4} is that this is the pattern currently supported. We can change that to ZZ(Base31){6} of course, but it'll break downwards compatibility. Do we want that?

eddiemuc:
> The reason I went with ZZ[0-9A-Z]{4} is that this is the pattern currently supported. We can change that to ZZ(Base31){6} of course, but it'll break downwards compatibility. Do we want that?

What do you mean with "currently supported"?

Are you referring to this line?
https://github.com/cgeo/cgeo/blob/9612d63941e9ae277620d254a5cc64d8e5f82d01/main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java#L57

Wouldnt it be enough to exchange the 4 with a 6 (resulting in `Pattern.compile("(" + PREFIX + ")[0-9A-Z]{1,6}", Pattern.CASE_INSENSITIVE)`)

Funnily enough, that pattern already "supports" letters in the ZZ-geocode while today we only generate numerics :-)

eddiemuc:
Addendum:
* my suggestion was to use some sort of "BASE34" (I don't know why gc.com leaves out U and L), but BASE31 is ok too of course
* I believe that it is of not much harm if the connector has a "slightly too wide" range of geocodes it accepts. The most important thing in case of Internal connector would be to check for prefix ZZ

ztNFny:
By staying in ZZ[0-9A-Z]{4} a GPX exported from "after this change" can be imported into cgeo "before this change".
Expanding it beyond that, e.g. by increasing the length to 6, breaks that compatibility. An old version of cgeo would import ZZ123456 as unknown cache type, not as UDC.

As 1,8 million possibilities is most likely good enough to avoid any conflicts (the usecase here is manually shared gpx files) i didn't see any reason to break that compatibility.

eddiemuc:
> By staying in ZZ[0-9A-Z]{4} a GPX exported from "after this change" can be imported into cgeo "before this change". Expanding it beyond that, e.g. by increasing the length to 6, breaks that compatibility. An old version of cgeo would import ZZ123456 as unknown cache type, not as UDC.

Understood, haven't thought of that. 
Then I suggest to generate 4-digit-codes, BUT increase compatibility of InternalConnector to accept longer codes (e.g. maybe up to 10 digits). This will allow us to generate longer codes some day in the future w/o breaking compatibility with the then-older versions of c:geo.

ztNFny:
did the recommended cleanup and increased the length of supported codes to ZZ[0-9A-Z]{1,10}. Generated codes are still [0-9A-Z]{4}

ztNFny:
retest this please

ztNFny:
retest this please

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
eddiemuc: This one?

ztNFny: staying with {4} I don't want to reduce the available range. And also those codes are usually not read/written, so don't see the risk of confusing those characters. (for UDCs the geocode could probably be completly hidden from the user)

ztNFny:
retest this please

main/src/main/java/cgeo/geocaching/connector/internal/InternalConnector.java
ztNFny: it's a reduction of possible values from 1,7 to 1,2 million (base 31 would be 0,9 million), statistically this means that for ~1300 geocodes generated the likelyhood of 1 conflicting is 50%. Of course this number is not relevant locally (becase there we can check for dups but only on exchanging GPX files
(36^5 would be ~9000, 31^5 6300)

eddiemuc:
LGTM, let's merge here. Thanks for the contribution!

Issue #16499, opened 2025-01-03
Live Map: "Partial" status indicator stuck / improve live map algorithm


eddiemuc:
PR #16484 introduced a status system for map live updates. By this I means: online calls to the cache servers to  fetch geocache data for current viewport and filter. 

As with this PR status is visualized with different "status bars" colors. Below a screenshotof this status bar in the RUNNING mode:
![image](https://github.com/user-attachments/assets/117dd3c6-2429-4146-a32a-452c334c0cc9)

As with above PR, status and visualization are as follows: 
* STOPPED_OK: no request is currently waiting, no request is running, everything is fine. Nothing is shown (no change to before)
* RUNNING: a request is currently processed. Shown by indetermined progress indicator in orange (no change to before)
*  REQUESTED: a request was triggered but is not yet processed (intentional delay to avoid unwanted online requests). Shown by indetermined progress indicator in blue
*  STOPPED_ERROR: previous request resulted in an error. Shown by red status bar
*  STOPPED_PARTIAL_RESULT: previous request executed ok but only parts of the caches in viewport/filter were returned. Shown by gray status bar.

This issue is meant for discussion on how to proceed here GUI-wise

eddiemuc:
_Originally posted by @ztNFny in https://github.com/cgeo/cgeo/issues/16484#issuecomment-2568186794_

I actually cherry-picked it now and gave it a test drive. In terms of GUI I'd suggest the following:
- load requested but not triggered: orange progress indicator in determinate mode (I assume this is a fixed delay of x seconds? didn't spot the code delaying it yet)
- load running: orange progress indicator in indeterminate mode
- load resulted in error: like 429 warning, red error icon
- load resulted in overflow: like 429 warning, half-filled circle icon

then the 429 icon becomes a "all map loading errors/warnings that the user needs to be informed about"-icon

ztNFny:
I'll try to implement my suggestion in the next days.
@eddiemuc could you quickly point me to where to find the "load requested" timer is? Couldn't spot that one yet

eddiemuc:
> I'll try to implement my suggestion in the next days. @eddiemuc could you quickly point me to where to find the "load requested" timer is? Couldn't spot that one yet

The delay in ms is defined in constant `LiveMapGeocacheLoader.PROCESS_DELAY` which is currently set to 2000ms. However, since the thread only checks every 250ms for updates, the delay will be randomly between 2000 and 2500ms .

ztNFny:
[...]

PR #16535, opened 2025-01-07
rel to #16499: refactor live map algorithm part 1

eddiemuc:
rel to #16499: refactor live map algorithm part 1

This PR should clean up the structures.


ztNFny:
is PROCESS_DELAY = 3000 by intention? We had reduced it to 1000 before, 3s delay feels quite slow.

eddiemuc:
> is PROCESS_DELAY = 3000 by intention? We had reduced it to 1000 before, 3s delay feels quite slow.

We can change it again

Issue #16527, opened 2025-01-06
Mapping Offline Lists: disable the auto-"Zoom-to-Bounds"


eddiemuc:
[Ticket #830503](https://cgeosupport.droescher.eu/scp/tickets.php?id=8946)
cgeo 2024.12.30-RC

When a stored list is opened in map, then the maps viewport zooms/moves to the bounds of the list (to fit all of the list into the map viewport).

A user on support requests the feature to disable this behaviour. Either as a setting in "Settings -> Map Content and Behaviour" or in map quick settings

ztNFny:
The zoom is also something that regulary annoys me. However moving the viewport is quite a practical feature and I think removing that would cause more problems than good. Imagine you're mapping a list in Spain, then next thing one in Germany. Without the current logic you'd first need to zoom out to figure out where you even are, then pan to the caches, then zoom in again - major pain.

Maybe something like this could work:
- On leaving a map view remember the current viewport and zoom level
- On mapping a list calculate viewport
- If "remembered viewport" is inside "list viewport" use "remembered viewport", else use "list viewport" (the current behaviour)

thavorianer:
I am the user who asked this.

I didn't mean this when displaying the map from a list, but only when you are on the map and then switch between the lists at the top. You only really want to see the saved caches in this area.

You could switch this off using the cogwheel at the bottom.

eddiemuc:
We could have an action button next to each list in the drop-down. Visually a bit like in named filter selection dialog (with a different icon of course, not a trash can :-))
![Screenshot_2025-01-09-06-45-58-53_a916ceb0c8a82619faeb76fe6b8ae997.jpg](https://github.com/user-attachments/assets/ddad616f-5ac4-45eb-a4c4-36b4d24ea7f9)

When this button is tapped the list could be opened w/I zoom. When the list is tapped it is opened with zoom (like today)

thavorianer:
Possible

![Screenshot_20250109_090212_cgeo](https://github.com/user-attachments/assets/74b71f4b-5aa7-4419-a3ff-856d5e8134d1)

Then i guess the font must be a bit smaller?
Or the total number is a button to view all, the name is to show in the current zoom.

ztNFny:
I don't like the button idea because of the visual consequences it has for the dropdown:
- each line will be a lot higher than it is right now (compare filter selection dialogs on map vs FilterActivity)
- also horizontall it'd be taking a lot of space from list titles.

[...]

PR #16558, opened 2025-01-12
"intelligent" restore map position on mapping a list

ztNFny:
when opening map from a list or search compare previous map center with viewport required to show the whole list. If it's inside use old viewport, else go to whole list viewport.
This helps in situations where you switch between list and map views so you don't always have to re-zoom the map.

partially addresses https://github.com/cgeo/cgeo/issues/16527
it doesn't add an "fit all" buttons to GUI

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -440,7 +440,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                 }, () -> {
                     if (viewport3.get() != null) {
                         if (setDefaultCenterAndZoom) {
-                            mapFragment.zoomToBounds(viewport3.get());
+                            restoreOrSetViewport(viewport3.get());
                         }
                         refreshMapData(true);
                     }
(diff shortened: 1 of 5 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -440,7 +440,12 @@ private void reloadCachesAndWaypoints() {
                 }, () -> {
                     if (viewport3.get() != null) {
                         if (setDefaultCenterAndZoom) {
-                            mapFragment.zoomToBounds(viewport3.get());
+                            final Geopoint storedMapCenter = Settings.getUMMapCenter();
```
eddiemuc, line 443: Shouldn't you apply this logic only if the new Setting `pref_autozoom_consider_lastcenter` is set to true?

ztNFny: ooops, what a stupid mistake ... Adding a setting and then not using it :-(

eddiemuc (approved):
Lgtm

dcjkfgdjhd:
I was asked to comment on whether this problem is now fixed.

- It is fixed when you use the buttons at the bottom of the screen to switch between list and map view.
- It is not fixed when in map view you click on the list name (top left) and select a different list. It is thus not possible to quickly check out another list for the same area.
- It is also not fixed for when in list view you use the map button in the top toolbar. This does reset zoom still.

PR #16563, opened 2025-01-12
fix #16553: seekbar not settable if not initialized beforehand

### LINKED ISSUE #16553: not able to enter a value or set the slider to a certain position for address search limit
> Thanks for the hint. But this does not work. I am not able to enter a value or set the slider to a certain position.

This should work. If not we might have another bug here?

We have.

eddiemuc:
fix #16553: seekbar not settable if not initialized beforehand


moving-bits:
Thanks for diving into this. Have you cross-checked your solution against #16473?

eddiemuc:
> Thanks for diving into this. Have you cross-checked your solution against #16473?

I have only cross-checked the two distance-seekbars (nearby and Coord/address) against each other. They are not entangled.

eddiemuc:
Do not merge this PR

PR #16568, opened 2025-01-15
Show offline caches mode

ztNFny:
Introduce a long-tap on "Live" button to enter from any offline map (e.g. list map, single cache map) a "database map", showing all caches stored offline in c:geo, no matter which list they are stored.

This "database map" mode can today only be accessed by entering "main map" (live map), disabling live mode, leaving and returning to main map.


ztNFny:
They have been substantial changes to the map loader after this PR was created, not surprised it's broken now

moving-bits:
See #16753 for an alternative approach (not yet fully tested)

moving-bits:
superseded by #16753

PR #16581, opened 2025-01-18
more menu icons

ztNFny:
More menu icons - and a cleanup of our menus

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/ToggleItemType.java
+++ b/main/src/main/java/cgeo/geocaching/ui/ToggleItemType.java
@@ -21,7 +21,7 @@ public enum ToggleItemType {
     LIVE_MODE(R.drawable.ic_menu_sync_enabled, R.drawable.ic_menu_sync_disabled, R.string.map_live_disable, R.string.map_live_enable),
     SELECT_MODE(R.drawable.ic_menu_select_end, R.drawable.ic_menu_select_start,
             R.string.caches_select_mode_exit, R.string.caches_select_mode),
-    TOGGLE_SPEECH(R.drawable.ic_menu_text_to_speech_on, R.drawable.ic_menu_text_to_speech_off,
+    TOGGLE_SPEECH(R.drawable.ic_menu_text_to_speech_off, R.drawable.ic_menu_text_to_speech_on,
             R.string.cache_menu_speechDeactivate, R.string.cache_menu_speechActivate),
     WAYPOINTS_FROM_NOTE(-1, -1, R.string.cache_menu_allowWaypointExtraction, R.string.cache_menu_preventWaypointsFromNote);
 
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -718,6 +718,9 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             menu.findItem(R.id.menu_advanced).setVisible(cache.getCoords() != null);
         }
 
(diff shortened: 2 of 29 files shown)


main/src/main/java/cgeo/geocaching/ui/ToggleItemType.java
```diff
@@ -21,7 +21,7 @@ public enum ToggleItemType {
     LIVE_MODE(R.drawable.ic_menu_sync_enabled, R.drawable.ic_menu_sync_disabled, R.string.map_live_disable, R.string.map_live_enable),
     SELECT_MODE(R.drawable.ic_menu_select_end, R.drawable.ic_menu_select_start,
             R.string.caches_select_mode_exit, R.string.caches_select_mode),
-    TOGGLE_SPEECH(R.drawable.ic_menu_text_to_speech_on, R.drawable.ic_menu_text_to_speech_off,
+    TOGGLE_SPEECH(R.drawable.ic_menu_text_to_speech_off, R.drawable.ic_menu_text_to_speech_on,
```
moving-bits, line 24: Did you swap those icons intentionally?

In current release they are representing the current state "talk is on/off", like the "live mode" icon does on map. Swapping those icons would move from "representing state" to "representing action when being pressed", like we have in other places.

ztNFny: Yes, intentional. They didn't match the text before. Which was never noticed as the icon wasn't shown.

moving-bits: The icon is shown if you rotate your mobile device to landscape mode. The icon "disabled loudspeaker" has a hint "start talking" currently, and the icon with the active loudspeaker has a hint "stop talking" - which is correct, IMHO. (Tested with current release 2025.01.18)

ztNFny: > "disabled loudspeaker" has a hint "start talking" currenly

That's not a correct use. The title/hint is using an action text, not a status one.

Correct is:
A) action text + matching action icon, e.g. disabled speaker + "stop talking"
B) status text + status icon: disabled speaker + "talking disabled"

ztNFny: Btw and off-topic, I've never ever seen anyone use that option, we should probably move it further down the list (and prevent it from being in landscape toolbar)

ztNFny: @moving-bits if you see the need to discuss this further I can split off that part of the PR so that the rest can be merged

moving-bits: > > "disabled loudspeaker" has a hint "start talking" currenly
> 
> That's not a correct use. The title/hint is using an action text, not a status one.
> 
> Correct is: A) action text + matching action icon, e.g. disabled speaker + "stop talking" B) status text + status icon: disabled speaker + "talking disabled"

agreed. To keep the icon in sync with map's "live mode" I propose to implement option B, so adapting the hints instead of swapping the icons.

ztNFny: This will require new strings i guess? For translators to notice it?

ztNFny: Could you merge here ans then we do the icon reverse + new strings in another PR?

moving-bits: ok, I'll prepare a PR for swapping icons + adapting hints

PR #16585, opened 2025-01-19
fix missing local save when editing PN from popup

ztNFny:
Make sure that a cache is stored when editing a cache from map popup


ztNFny:
@moving-bits can we get this merged? It's causing data loss and thus is targeted against release.

Issue #16000, opened 2024-08-26
HIerarchical lists: rename list name prefixes


eddiemuc:
From support[Ticket #839865](https://cgeosupport.droescher.eu/scp/tickets.php?id=8437)

As Beta user of c:geo, I am using the new feature that allows to build a hierarchical structure of lists.

I have a structure like this :

Prefix:Suffix A
Prefix:Suffix B
Prefix:Suffix C
...

I would like to have a function that allows me to rename the "Prefix" part of all lists with this prefix in one go.

PR #16594, opened 2025-01-25
Add parent list selector to create/rename list dialogs

moving-bits:
## Description
Adds a parent list dropdown selector to create list / rename list dialog, if such lists exist:

|Create list 1|Create list 2|Create list 3|
|---|---|---|
|Screenshot_1737808497|Screenshot_1737805983|Screenshot_1737805999

|Rename list with parent lists|Create list without parent lists|Rename lists without parent lists|
|---|---|---|
|Screenshot_1737807890|Screenshot_1737808474|Screenshot_1737807841|

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -621,6 +621,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             MenuUtils.setVisible(menu, R.id.menu_lists, isOffline);
             MenuUtils.setVisible(menu, R.id.menu_drop_list, isNonDefaultList);
             MenuUtils.setVisible(menu, R.id.menu_rename_list, isNonDefaultList);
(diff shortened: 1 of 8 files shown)


moving-bits:
Second commit adds a menu entry "Rename list prefix" to the "manage lists" submenu to rename a list prefix (fixes #16000):

|before|rename|after|
|---|---|---|
|Screenshot_1737841314|Screenshot_1737842385|Screenshot_1737841286|

This function can be used for renaming a list prefix as well as for removing or adding a hierarchy level.

A confirmation message will be shown before updating the database. If the "to" string contains no list level marker anymore, a different confirmation message will be shown to notify the user, that all matching lists will be converted to top-level list.

Menu entry "Rename list prefix" will only be shown if there is at least one list having a ":" list marker in its name.

murggel:
Do we have a chance to adapt the filter if a renamed list is used there?

moving-bits:
> Do we have a chance to adapt the filter if a renamed list is used there?

The "stored list" filter stores the list's id, not its name, so it's unaffected by a rename.

Any other filters that would be affected?

moving-bits:
Let's merge here and open dedicated issues for filters if there are any specific filters affected.

PR #16597, opened 2025-01-26
[rel #16382] Consider not only 2 digits minimum for deg comma parser

murggel:
## Description

Fix the regexpr to consider only 1 digit for longitude value before comma (in `getParseInputs` /  `removeSpaceAfterSeparators`)

The other parsers were able to deal with the not removed space, because this was supported in their own regexpressions.
Since I didn't want to change to much for the existing parsers, I added a new method for the special format

## Related issues

#16382


ztNFny:
Deg is always 1, 2 or 3 chars, maybe the regexp could be adjusted to \d{1,3},... to reduce fals e positive matches?

murggel:
> Deg is always 1, 2 or 3 chars, maybe the regexp could be adjusted to \d{1,3},... to reduce fals e positive matches?

Done - and consider only comma / no point, since it is only for comma/comma -format

SammysHP:
What about negative latitude?

murggel:
> What about negative latitude?

Should be covered, but I can add it explicitely to the front of the regexp as optional

PR #16598, opened 2025-01-26
Remove support for extremcaching.com

moving-bits:
to be discussed, therefore setting "WIP" label for now


bekuno:
How does the change deal with existing EC caches?
For my understanding they should be stay as is in the database and on the map (_offline_ functions). Only all _online_ functions should be removed.
What will happen when exporting/importing such caches via gpx?

moving-bits:
> How does the change deal with existing EC caches?

Currently this PR is "radical", removing all functions related to EC.

Have not tested it (I don't have any EC data), but it should be like this to my understanding:
- Existing GPX with EC caches can still be imported, but there would be no longer any special handling for it.
- EC caches in database can still be used, but will show up as "unknown connector".

moving-bits:
In our last dev meeting we decided to switch the connector to an offline connector. This got implemented with #16667, so let's close here.

Issue #16609, opened 2025-01-31
Addding "problem report" while editing a log won't show up on the server


moving-bits:
User report on support (ticket 529943):

When adding a "report a problem" only while editing an existing log, this error report won't get posted.

c:geo release 2025.01.18

ztNFny:
"report problem" isn't supposed to show up in edit mode. There's a flaw in the check logic, PR coming up.

moving-bits:
Fixed with #16610 (hiding the non-applicable "report problem" checkbox on editing an existing log)

PR #16610, opened 2025-01-31
hide "report problem" when editing a log

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/16609

code had && instead of ||. Changed the condition to be more readable


ztNFny:
rebased to release

PR #16614, opened 2025-02-01
UnifiedMap: Set displayName for downloaded offline maps (fix #16586)

### LINKED ISSUE #16586: Ability to rename map sources
### Tell us your idea!

Thank you very much for the recent map updates, these are great!

Now that it is possible to define a user defined tile server as a source, I would love to also give this source a custom name to be displayed in the map layer dropdown.

Even better would be if it would also be possible to give *every* map source a user defined name.

See above.

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Allow configuring displayname for offline map sources downloaded with c:geo:

|map selection|folder data|companion file|
|---|---|---|
|image|image|image|

Only `displayname` property in companion file needs to be changed, everything else is untouched. Updating such a map source does work, `displayname` gets preserved. Map sources are sorted by `displayname` within their respective categories.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/downloader/CompanionFileUtils.java
+++ b/main/src/main/java/cgeo/geocaching/downloader/CompanionFileUtils.java
@@ -2,12 +2,15 @@ package cgeo.geocaching.downloader;
 
 import cgeo.geocaching.models.Download;
 import cgeo.geocaching.storage.ContentStorage;
+import cgeo.geocaching.storage.Folder;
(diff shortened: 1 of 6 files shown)


ztNFny:
This broke the "(VTM)" suffix for double map sources

moving-bits:
> This broke the "(VTM)" suffix for double map sources

PR #16671 will fix this

Issue #16436, opened 2024-12-17
Add variable proximity circle radius 


LayanJethwa:
### Tell us your idea!

There is currently an option to turn on a proximity circle around all caches with a 0.1 mile radius, which is helpful for cache placement, but I would also like if this could have an option for a 2 mile radius to show the distance between a mystery cache final and the dummy coordinates.

### What solution would you suggest?

There should be an option to change the radius of these circles in the settings.

### Is there an alternative?

I have to go into Geocaching toolbox and copy paste the coordinates in of all the caches I want circles for, and it is tricky to do on my phone.

jakedot:
This would be very useful for Lab caches.
They have variable zone sizes., which is returned in API calls, which I am using over at my [Lab2GPX clone](https://jakedot.net/l2g) to add it to the cache description.

I have the code for that (in PHP) at my repository at [jakedot/labs2gpx](https://github.com/jakedot/lab2gpx) of anyone is interested.

ztNFny:
> This would be very useful for Lab caches.
> They have variable zone sizes., which is returned in API calls, which I am using over at my [Lab2GPX clone](https://jakedot.net/l2g) to add it to the cache description.
> 
> I have the code for that (in PHP) at my repository at [jakedot/labs2gpx](https://github.com/jakedot/lab2gpx) if anyone is interested.
> 
> Maybe we could add a special color for Lab Stages?

That's actually a way more complex request as it adds 
- many different sizes at once
- per waypoint
- stored in db
- imported from gpx (which had no standard field for this info) 

For labs I'd suggest geojson instead.

moving-bits:
IMHO "variable proximity circle" and "geofence circle for lab stages" should be treated separately.

Geofence circle for lab stages could be calculated and displayed automatically,
- if we store the geofence radius for lab stages
- and add a map quick settings to enable / disable display of such circles.
- Color definition may be configurable or fixed (maybe use the same color as integrated Wherigo uses for its zones?)

Exporting into / reading from GPX would be a separate thing, but this may be beyond the initial request in https://github.com/cgeo/cgeo/issues/16436#issuecomment-2571450399

moving-bits:
See #16620 for an implementation proposal.

PR #16618, opened 2025-02-02
UnifiedMap: Show geofences for lab stages (fix #16591)

### LINKED ISSUE #16591: Fetch and display zone size for Adventure Labs
Lab Caches have variable zone sizes., which is returned in API calls, which I am using over at my [Lab2GPX clone](https://jakedot.net/l2g) to add it to the cache description.

I have the code for that (in PHP) at my repository at [jakedot/labs2gpx](https://github.com/jakedot/lab2gpx) if anyone is interested.

Maybe we could add a special color for Lab Stages? 

 _Originally posted by @JakeDot in [#16436](https://github.com/cgeo/cgeo/issues/16436#issuecomment-2571450399)_

moving-bits:
## Description
Added option to show geofences for lab adventure stages:

image.image

(A solution for #16436 could be built upon this feature by allowing users to set "geofence" manually for waypoints.)

Edit: Separate option to enable/disable geofence circles has been removed meanwhile. Geofence circles will be switched by using the "circles" checkbox.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -411,6 +411,7 @@ final class ALApi {
 
                 wpt.setGeocode(geocode);
                 wpt.setPrefix(String.valueOf(stageCounter));
+                wpt.setGeofence((float) wptResponse.get("GeofencingRadius").asDouble());
 
                 final StringBuilder note = new StringBuilder("<img src=\"" + ilink + "\"></img><p><p>" + desc);
                 if (Settings.isALCAdvanced()) {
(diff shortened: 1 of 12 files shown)


ztNFny:
nice! 3 remarks:
- Do we really need a new option in UI for this? imho it could be integrated with the existing "Circles" setting
- color should be a setting instead of hardcoded
- default color should be R.color.cacheType_lab

moving-bits:
> * Do we really need a new option in UI for this? imho it could be integrated with the existing "Circles" setting

I was thinking of potential collisions, especially in combination with the planned extension mentioned above (#16436). But you are right, those should be infrequent, as lab stages don't have a proximity circle.

> * color should be a setting instead of hardcoded
> * default color should be R.color.cacheType_lab

Here I wanted to keep things simple... But an option could be added, if necessary (actually two options, one for border + one for fill, if we want to have it similar to distance circles).

ztNFny:
> But an option could be added, if necessary

All of those line elements are configurable. Whether that's really needed ... but for consistency we shouldn't start with this one being the only exception

> (actually two options, one for border + one for fill, if we want to have it similar to distance circles).

2 options sounds like overkill, I'd have assumed that those circles use just 1 color with the fill being a transparent version of the border

moving-bits:
Updated the PR:

image.image

Geofence fill color uses geofence line color with alpha value restricted to a maximum of 64.

Shall we use the same algorithm for proximity circle, removing one option?

ztNFny:
> Shall we use the same algorithm for proximity circle, removing one option?

2 options ;-) Accuracy as well. I'd be fine with it.

Buuut ... looking at those old defaults we have:
- Accuracy: Circle 102, Fill 8 -> 8%
- Distance: Circle 68, Fill 44 -> 64%

ztNFny:
Default for all circle lines is alpha 102, so the default for labs should probably be #667b1fa2 instead of #ff7b1fa2 (sorry for my previous "wrong" suggestion)

moving-bits:
Default values would look like this (with alpha restriction for fill color to max=48):

image

(still using lab color as geofence color in this screenshot)

JakeDot:
LGTM

PR #16634, opened 2025-02-08
fix #16617: unified live map: don't call server after partial/error result when too close to previous position

### LINKED ISSUE #16617: [Nightly] Infinite download requests
### Describe your problem!

When map is zoomed out until view port is too big to get all caches, we get the "Incomplete" notice, which is fine. 
But after a second or so the icon vanishes and the next download request starts (as indicated by the floating bar at the top of the map).
Then we again get the "Incomplete" notice. 
And after a second or so the next download request starts ...
This way it goes on forever ...

### How to reproduce?

1. Open map
2. Zoom out until you get the "Incompletet" icon
3. Wait ...

After a second the icon goes away and the next download starts.

There should not be another download started until I move a significant distance (of let's say at least the width of my current view port).

### Reproducible

Yes

### c:geo Version

2025.02.02-NB-b0cff9c

### System information

```text

```

### Additional Information

This happened to me while sitting in my living room, thus may be triggered by an imprecise signal. Nevertheless, in the map the position indicator did not move.

If helpful, I can even provide a video later (right now .mp4 does not upload for whatever reason).

eddiemuc:
fix #16617: unified live map: don't call server after partial/error result when too close to previous position

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/Viewport.java
+++ b/main/src/main/java/cgeo/geocaching/location/Viewport.java
@@ -4,6 +4,9 @@ import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.ICoordinate;
 import cgeo.geocaching.models.Waypoint;
+import cgeo.geocaching.models.geoitem.GeoItem;
+import cgeo.geocaching.models.geoitem.GeoPrimitive;
+import cgeo.geocaching.models.geoitem.GeoStyle;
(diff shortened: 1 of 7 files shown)


eddiemuc:
The fix works, but it will have a strange effect on the showing of is partial state (it will vanish and not return). I have to rework the state system.

If the bug is too annoying we can merge here and I will do this in a separate PR. Or we keep this PR open and I work it directly in. What should it be?

(You probably noticed that I have not too much time for coding lately unfortunately)

eddiemuc:
New statushandling now added

PR #16640, opened 2025-02-08
Remove WhereYouGo from 'useful apps' list

moving-bits:
## Description
Since having an integrated Wherigo player, a reference to WhereYouGo in "useful apps" list is no longer required


JakeDot:
relevant: https://github.com/cgeo/cgeo.github.io/issues/340

PR #16644, opened 2025-02-09
Attempt to fix issue #16608

moving-bits:
## Description
One of the recent changes in file handling removed the `<data scheme="file" />` attribute to allow for `scheme="content"` calls additionally. According to the docs, if none scheme is given, both `file` and `content` should be used, but maybe this is different when being opened from apps.

To test this, this PR configures separate blocks for both schemes.

@Lineflyer / @ztNFny: Do you have a chance to test whether this change helps in your use-cases?

Setting "do not merge" label for now.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -304,7 +304,50 @@
 
                 <data android:mimeType="*/*" />
                 <data android:host="*" />
-                <!-- <data android:scheme="xxx" /> neither "file" nor "content" means: both -->
+                <data android:scheme="file" />
+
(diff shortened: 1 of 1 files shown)


moving-bits:
> with this merged c:geo isn't available at all as file handler from Signal (from saved local gpx / zip it still is)

:man_shrugging: :thinking: 

but thanks for cross-checking (I don't have signal installed on a device where I can test custom builds, so I can't check for myself)

moving-bits:
this approach seemingly does not work, thus closing here. But maybe #16658 will work?

PR #16652, opened 2025-02-09
Update Mapsforge to v0.23

moving-bits:


moving-bits:
needs further code adjustments for upgrading Mapsforge, so let's close here

PR #16657, opened 2025-02-12
Support Wherigofoundation (WIP)

moving-bits:
## Description
As we got a another request on support to support wherigofoundation.com (see ticket 570714), I started a simple extension to our code (detecting the server, extracting the guid etc.). 

But it's not that simple, sadly, as wherigofoundation needs its own set of credentials + logging in on the page + selecting the right variant for cartridge download.

Due to time constraints I stopped here for now. So if someone wants to step in, feel free to add the missing parts.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -218,6 +218,22 @@
                 <data android:pathPrefix="/cartridge/downloads.aspx" />
             </intent-filter>
 
+            <!-- wherigofoundation.com -->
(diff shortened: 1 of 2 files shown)


JakeDot:
I think it's time for a c:geo Plugin for Wherigo Utils. If we want to become a platform.

moving-bits:
Closing here, see discussion in https://github.com/cgeo/cgeo/issues/16911

PR #16659, opened 2025-02-12
fix #16022: ReplaceSearchResult instead of add (UnifiedMap)

### LINKED ISSUE #12737: Update list from map popup doesn't update map view
### Describe your problem!

When the map shows caches of a specific list and you move a cache to another list from the map popup, the map won't be updated automatically.

### How to reproduce?

1. Open list
2. Show caches on map
3. Open popup for a cache
4. Tap on "edit lists" and move the cache to another list
5. Close the popup

Cache is still shown on the map

Cache should not be visible on the map as it belongs to another list now.

### Reproducible

Yes

### c:geo Version

2022.02.13

### System information

_No response_

### Additional Information

_No response_

### LINKED ISSUE #16022: Remove cache from map when deleting from list (show list on map)
### Tell us your idea!

When managing cache lists under "Stored" one has the option of showing the entire list of caches on the map. 
See map button on top.

While browsing the map, if I tap on a cache and tap remove from list, the cache icon on the map gets updated (ie the floppy subicon is removed), but it remains visible. 

This is incorrect within the logic that this map view is supposed to show only caches from list X.

My recommendation is to remove the cache icon (and associated waypoints) from this map view, keeping in with the logic behind the purpose of this view.

Showing the cache under the rules of the general map should be unaffected.

### Is there an alternative?

go back to list view, then tap on show list on map again.

### Additional Information

Showing list member caches using this button: 
Screenshot_20240828-183917_cgeo

murggel:
## Description

If caches get removed from list, don't show them in list-map-view, so remove caches before adding new ones

Hopefully that does not have any performance impact

## Related issues

fix #16022
fix #12737

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -443,7 +443,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                 AndroidRxUtils.andThenOnUi(Schedulers.io(), () -> {
                     final SearchResult searchResult = DataStore.getBatchOfStoredCaches(null, mapType.fromList, mapType.filterContext.get(), null, false, -1);
                     viewport3.set(DataStore.getBounds(searchResult.getGeocodes(), Settings.getZoomIncludingWaypoints()));
-                    addSearchResultByGeocaches(searchResult);
+                    replaceSearchResultByGeocaches(searchResult);
                 }, () -> {
                     if (viewport3.get() != null) {
                         if (setDefaultCenterAndZoom) {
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -646,13 +646,15 @@ public static void refreshWaypoints(final UnifiedMapViewModel viewModel) {
     }
 
 
-    public void addSearchResultByGeocaches(final SearchResult searchResult) {
-        Log.d("add " + searchResult.getGeocodes());
+    public void replaceSearchResultByGeocaches(final SearchResult searchResult) {
+        Log.d("replace " + searchResult.getGeocodes());
+        viewModel.caches.write(true, caches -> { // use post to make it background capable
+            caches.clear();
```
murggel, line 652: I use caches.clear here. Set::clear didn't work for me - no idea why, I will check again...

PR #16666, opened 2025-02-16
[WIP] Add Digital Root function

ianfhunter:
## Description

Adds a new function for users when driving coordinates 

## Related issues

N/A

## Additional context

Not very familiar with java/android so have made the theoretical changes but not run then through testing just yet


eddiemuc:
HI @ianfhunter I believe what you try to achieve is already available with function "`checksum`" (or "ics" for short):

image

e.g. using `ics(998)` would result in 8 (whereas `cs(998)` results in 26)

We can however add "digitalroot", "droot" and "dr" as additional function names for that function. Is this what you seek for?

eddiemuc:
ok to test

ianfhunter:
Ah, it may be, I can't follow exactly what the iterative checksum is doing, but the digital root would also produce 8

https://en.wikipedia.org/wiki/Digital_root

Just for clarity, it does this by the following iterative summation: 
9 + 9 + 8 = 26 (> 1 digit so continue)
2 + 6 = 8 (1 digit so end) 

If this is how `ics` works, I think adding those function names as aliases is enough.

SammysHP:
I think we are talking about an iterative sum of digits ("iterierte Quersumme" in German). That's what `ics` does. But I'm not sure if it is worth to add even more aliases.

ianfhunter:
I don't mind what happens as I know now what I can use, but it seems to me that Digital Root is the correct name in English for this

PR #16667, opened 2025-02-16
Make extremcaching.com an offline connector (fix #16570)

### LINKED ISSUE #16570: Extremcaching.com sundown?
The site extremcaching.com is down since a couple of days.

The WHOIS status shows:
- Expires On: 2025-01-22
- Status: clientDeleteProhibited, clientRenewProhibited, clientTransferProhibited, clientUpdateProhibited

With the status `clientRenewProhibited` it looks like a renewal of this page is not planned.

I have not yet received any feedback from an EC contact.

moving-bits:
## Description
As discussed in our last dev meeting:
Remove all online capabilities for extremcaching.com connector, but leave offline capabilities unchanged


moving-bits:
@bekuno 
Can you cross-check with your offline data for EC whether this works as expected with this PR?
(EC caches still being labelled as EC, but no online access to their server)
Thanks.

moving-bits:
Let's give it a try in the nightlies, as we should publish a bugfix release rather soon due to the seekbar crashes polluting the logs.

PR #16671, opened 2025-02-17
Fix missing (VTM) indicator when displayname is set (rel. to #16614)

moving-bits:
## Description
See https://github.com/cgeo/cgeo/pull/16614#issuecomment-2662286190


ztNFny:
Thx

Issue #15874, opened 2024-07-14
Inline listing translation


ztNFny:
I'm currently traveling abroad and with each listing I open I do a "select all, translate" to translate the text in Google translate.
Would be great to have that already built-in into c:geo. Recognizing the text is in a language that needs translation is probably hard to implement, but maybe it'd at least be possible to replace the original texts with the translated versions in their native places, ideally even keeping the layout.

How it'd work: Item in overflow menu "translate listing" which passes the texts to a translator service (Google Translate??), then uses the returned text to replace the original listing text.

Resources: 
https://developers.google.com/ml-kit/language/translation

wolverine007:
+1 for this feature

in May the next Giga in Prague will take place. Many listings in CZ are in local language only.

So might be cool to translate every listing automatically to your desired target language. 

Instead of identifying if a listing needs a translation, all listings should be translated automatically when the user activates this feature.

eddiemuc:
+1 for this feature too.

But we should be aware that translations may not have same quality as using "Google translate" online, since it is done on the device and using English as intermediary language if translating between two non-engliah languages. Also there's yet another file resource to manage wrt download and storage (translation packages per language)

PR #16673, opened 2025-02-20
inline listing translation

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/15874

Todos:
- [x] log translation
- [x] separate UI from translation logic
- [x] UI for download / deletion
- [x] try to keep listing formatting

image
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -255,7 +255,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     private HtmlStyle descriptionStyle = HtmlStyle.DEFAULT;
 
-    final public MutableLiveData<OfflineTranslateUtils.Language> translationSourceLanguage = new MutableLiveData<>();
+    final private MutableLiveData<OfflineTranslateUtils.Language> translationSourceLanguage = new MutableLiveData<>();
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
@@ -1855,7 +1855,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         OfflineTranslateUtils.LanguagePackDownloadHandler languagePackDownloadHandler = null;
 
         private void initializeListingTranslator() {
-            if (Settings.getTranslationTargetLanguage().isEmpty()) {
+            if (!Settings.getTranslationTargetLanguage().isValid()) {
                 binding.descriptionTranslate.setVisibility(View.GONE);
                 return;
             }
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -104,6 +104,7 @@ import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.MenuUtils;
+import cgeo.geocaching.utils.MlKitTranslateUtil;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.ProgressBarDisposableHandler;
 import cgeo.geocaching.utils.ProgressButtonDisposableHandler;
@@ -114,7 +115,6 @@ import cgeo.geocaching.utils.functions.Action1;
 import cgeo.geocaching.utils.html.HtmlStyle;
 import cgeo.geocaching.utils.html.HtmlUtils;
 import cgeo.geocaching.utils.html.UnknownTagsHandler;
-import cgeo.geocaching.utils.MlKitTranslateUtil;
 import cgeo.geocaching.wherigo.WherigoActivity;
 import cgeo.geocaching.wherigo.WherigoUtils;
 import cgeo.geocaching.wherigo.WherigoViewUtils;
@@ -256,11 +256,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     private HtmlStyle descriptionStyle = HtmlStyle.DEFAULT;
 
-    public void setListingTranslationLanguage(MlKitTranslateUtil.Language listingTranslationLanguage) {
-        this.listingTranslationLanguage.setValue(listingTranslationLanguage);
-    }
-
-    public MutableLiveData<MlKitTranslateUtil.Language> listingTranslationLanguage = new MutableLiveData<>();
+    final public MutableLiveData<MlKitTranslateUtil.Language> listingTranslationLanguage = new MutableLiveData<>();
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
@@ -1880,7 +1876,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                 }
             });
             binding.descriptionTranslateButton.setOnClickListener(v -> translateListing());
-            binding.descriptionTranslateNote.setOnClickListener(v -> MlKitTranslateUtil.showLanguageSelection(getActivity(), cda::setListingTranslationLanguage));
+            binding.descriptionTranslateNote.setOnClickListener(v -> MlKitTranslateUtil.showLanguageSelection(getActivity(), cda.listingTranslationLanguage::setValue));
             binding.descriptionTranslate.setVisibility(View.VISIBLE);
 
             // identify listing language
@@ -1908,7 +1904,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                 return;
             }
 
-            MlKitTranslateUtil.Language sourceLng = cda.listingTranslationLanguage.getValue();
+            final MlKitTranslateUtil.Language sourceLng = cda.listingTranslationLanguage.getValue();
 
             MlKitTranslateUtil.translateText(cda, sourceLng,
                 unsupportedLng -> {
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -104,7 +104,7 @@ import cgeo.geocaching.utils.LocalizationUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.MenuUtils;
-import cgeo.geocaching.utils.MlKitTranslateUtil;
+import cgeo.geocaching.utils.OfflineTranslateUtils;
 import cgeo.geocaching.utils.ProcessUtils;
 import cgeo.geocaching.utils.ProgressBarDisposableHandler;
 import cgeo.geocaching.utils.ProgressButtonDisposableHandler;
@@ -178,7 +178,6 @@ import java.util.Map;
 import java.util.Set;
 
 import com.google.android.material.button.MaterialButton;
-import com.google.mlkit.nl.languageid.LanguageIdentification;
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.disposables.CompositeDisposable;
 import io.reactivex.rxjava3.functions.Function;
@@ -256,7 +255,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     private HtmlStyle descriptionStyle = HtmlStyle.DEFAULT;
 
-    final public MutableLiveData<MlKitTranslateUtil.Language> listingTranslationLanguage = new MutableLiveData<>();
+    final public MutableLiveData<OfflineTranslateUtils.Language> translationSourceLanguage = new MutableLiveData<>();
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
@@ -1853,7 +1852,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
         }
 
-        MlKitTranslateUtil.LanguagePackDownloadHandler languagePackDownloadHandler = null;
+        OfflineTranslateUtils.LanguagePackDownloadHandler languagePackDownloadHandler = null;
 
         private void initializeListingTranslator() {
             if (Settings.getTranslationTargetLanguage().isEmpty()) {
@@ -1864,10 +1863,10 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             final CacheDetailActivity cda = (CacheDetailActivity) getActivity();
 
             // add observer to listing language
-            cda.listingTranslationLanguage.observe(this, lng -> {
+            cda.translationSourceLanguage.observe(this, lng -> {
                 if (null == lng || Settings.getLanguagesToNotTranslate().contains(lng.getCode())) {
                     binding.descriptionTranslate.setVisibility(View.GONE);
-                } else if ("und".equals(lng.getCode())) {
+                } else if (OfflineTranslateUtils.LANGUAGE_UNKNOWN.equals(lng.getCode())) {
                     binding.descriptionTranslateButton.setEnabled(false);
                     binding.descriptionTranslateNote.setText(R.string.translator_language_unknown);
                 } else {
@@ -1876,19 +1875,16 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                 }
             });
             binding.descriptionTranslateButton.setOnClickListener(v -> translateListing());
-            binding.descriptionTranslateNote.setOnClickListener(v -> MlKitTranslateUtil.showLanguageSelection(getActivity(), cda.listingTranslationLanguage::setValue));
+            binding.descriptionTranslateNote.setOnClickListener(v -> OfflineTranslateUtils.showLanguageSelection(cda, cda.translationSourceLanguage::setValue));
             binding.descriptionTranslate.setVisibility(View.VISIBLE);
 
             // identify listing language
-            LanguageIdentification.getClient().identifyLanguage(binding.description.getText().toString())
-                    .addOnSuccessListener(lngCode -> {
-                        cda.listingTranslationLanguage.setValue(new MlKitTranslateUtil.Language(lngCode));
-                    })
-                    .addOnFailureListener(e -> {
+            OfflineTranslateUtils.detectLanguage(binding.description.getText().toString(),
+                    cda.translationSourceLanguage::setValue,
+                    error -> {
                         binding.descriptionTranslateButton.setEnabled(false);
-                        binding.descriptionTranslateNote.setText(e.getMessage());
+                        binding.descriptionTranslateNote.setText(error);
                     });
-
         }
 
         private void translateListing() {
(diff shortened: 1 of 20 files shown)


main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1857,114 @@ public void setContent() {
 
         }
 
+        private void identifyTranslationLanguage() {
```
eddiemuc, line 1860: Please move the translation and related logic (eg download of language packages including related dialogs) to its own new class (maybe a new Utils class). That way we can easily reuse this logic in other places in the future eg waypoint texts, logs, trackable listings and whatever will come to our mind in the future. The utility class would ideally offer only very simple methods like "getTranslateLanguage" (null if none), "translate" (with parameters of type String and Context; the later in case dialogs are needed).

At some point (not necessarily with first PR) we need to allow users to manage/delete downloaded translation packages. Logic to access/list and delete those can then also be encapsulated in said Utils class

ztNFny: See the todo-list in PR description 😉

ztNFny:
Huge refactoring of the code done, @eddiemuc now a review would be appreciated :-)

Still missing is UI for deleting the downloaded language models and log translation, but I'd put those in a separate PR as I'm not sure when I'll get to it.

ztNFny:
restest this please

ztNFny:
some remarks before they come up in review:
- model download size is not reported
- model download progress is not available
- translation is always XX->en<-YY, so translating from french to german is actually 2 translations

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1853,101 @@ public void setContent() {
...
+                        cda.listingTranslationLanguage.setValue(new MlKitTranslateUtil.Language(lngCode));
+                    })
+                    .addOnFailureListener(e -> {
+                        binding.descriptionTranslateButton.setEnabled(false);
+                        binding.descriptionTranslateNote.setText(e.getMessage());
+                    });
+
+        }
+
+        private void translateListing() {
```
eddiemuc, line 1894: I believe this method as a whole could be moved to the utils class. 
Eg a method "translate(Context context, CharSequence sourceText, Language sourceLanguage, BiConsumer<Language, String> translatedTextConsumer>) which transparent to the caller also handles all the "download stuff" logic (asking user if he wants to download etc etc)

eddiemuc (changes requested):
Thanks for this change, I can't wait to have it in my c:geo installation :)

I have some change requests though

main/src/main/java/cgeo/geocaching/utils/MlKitTranslateUtil.java
```diff
@@ -0,0 +1,191 @@
...
+    private static List<Language> getSupportedLanguages() {
+        final List<Language> languages = new ArrayList<>();
+        final List<String> languageIds = TranslateLanguage.getAllLanguages();
+        for (String languageId : languageIds) {
+            languages.add(new Language(TranslateLanguage.fromLanguageTag(languageId)));
+        }
+        return languages;
+    }
+
+    public static CharSequence[] getSupportedLanguageDisplaynames() {
```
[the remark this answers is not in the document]
ztNFny: let's discuss this in 1 comment only :-)

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1853,101 @@ public void setContent() {
...
+                return;
+            }
+
+            final CacheDetailActivity cda = (CacheDetailActivity) getActivity();
+
+            // add observer to listing language
+            cda.listingTranslationLanguage.observe(this, lng -> {
+                if (null == lng || Settings.getLanguagesToNotTranslate().contains(lng.getCode())) {
+                    binding.descriptionTranslate.setVisibility(View.GONE);
+                } else if ("und".equals(lng.getCode())) {
```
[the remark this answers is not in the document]
ztNFny: "und" = undefined, will add a constant

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
ztNFny: Everything that's left in this class is very specific to the UI of the specific activity. Out of 55 lines only 2 are calls to some translation stuff - and those 2 are the wrappers.

> which transparent to the caller also handles all the "download stuff" logic (asking user if he wants to download etc etc)

all of that is already in the utility method

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1853,101 @@ public void setContent() {
 
         }
 
+        MlKitTranslateUtil.LanguagePackDownloadHandler languagePackDownloadHandler = null;
+
+        private void initializeListingTranslator() {
+            if (Settings.getTranslationTargetLanguage().isEmpty()) {
```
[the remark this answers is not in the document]
ztNFny: Same as the comment for translateListing() - everything in this method is very specific to the activity it's called from, it's all UI.
> returns an enum with values "OFF", "SOURCE_LANGUAGE_NOT_SUPORTED", "LANGUAGE_PACKAGES_NOT_DOWNLOADED" and "AVAILABLE"

that's exactly what "cda.listingTranslationLanguage.observe()" is - a return of 3 possible values and then updating the UI according to it. I don't see how a single line of this could be moved to a utility class without making that utility class specific to CacheDetailActivity - which is not the point of a utility class.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1853,101 @@ public void setContent() {
...
+                    binding.descriptionTranslateButton.setEnabled(true);
+                    binding.descriptionTranslateNote.setText(getResources().getString(R.string.translator_language_detected, lng.toString()));
+                }
+            });
+            binding.descriptionTranslateButton.setOnClickListener(v -> translateListing());
+            binding.descriptionTranslateNote.setOnClickListener(v -> MlKitTranslateUtil.showLanguageSelection(getActivity(), cda.listingTranslationLanguage::setValue));
+            binding.descriptionTranslate.setVisibility(View.VISIBLE);
+
+            // identify listing language
+            LanguageIdentification.getClient().identifyLanguage(binding.description.getText().toString())
```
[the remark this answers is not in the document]
ztNFny: well, can do it, but the resulting code will be exactly the same size...

Utility:
            MlKitTranslateUtil.detectLanguage(binding.description.getText().toString(),
cda.listingTranslationLanguage::setValue, 
error -> {
                binding.descriptionTranslateButton.setEnabled(false);
                binding.descriptionTranslateNote.setText(error);
            });

No utility:
            LanguageIdentification.getClient().identifyLanguage(binding.description.getText().toString())
                    .addOnSuccessListener(lngCode -> {
                        cda.listingTranslationLanguage.setValue(new MlKitTranslateUtil.Language(lngCode));
                    })
                    .addOnFailureListener(e -> {
                        binding.descriptionTranslateButton.setEnabled(false);
                        binding.descriptionTranslateNote.setText(e.getMessage());
                    });

ztNFny:
log translation and model deletion implemented :-)

main/src/main/java/cgeo/geocaching/settings/fragments/PreferenceCachedetailsFragment.java
```diff
@@ -1,13 +1,26 @@
...
+import androidx.preference.ListPreference;
+import androidx.preference.MultiSelectListPreference;
+
 public class PreferenceCachedetailsFragment extends BasePreferenceFragment {
     @Override
     public void onCreatePreferences(final Bundle savedInstanceState, final String rootKey) {
         initPreferences(R.xml.preferences_cachedetails, rootKey);
+
+        final ListPreference translateTargetLng = findPreference(getString(R.string.pref_translation_language));
+        translateTargetLng.setEntries(MlKitTranslateUtil.getSupportedLanguageDisplaynames());
```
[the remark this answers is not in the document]
eddiemuc: IMHO Interfaces should define High-Level concepts for multiple users and not usage-speciric constructs. In this specific example this means to me to offer Interfaces using collections rather than arrays. 

As for utilities for conversions of collections into arrays and vice versa: I believe that the JDK already offers a great deal of utilities here (one of them being streams which you used), but if you feel you want an additional utility function to convert lists to arrays I won't mind. I suggest to put it into CommonUtils if you want to proceed this.

main/src/main/java/cgeo/geocaching/utils/MlKitTranslateUtil.java
```diff
@@ -0,0 +1,191 @@
...
+
+import com.google.mlkit.common.model.DownloadConditions;
+import com.google.mlkit.common.model.RemoteModelManager;
+import com.google.mlkit.nl.translate.TranslateLanguage;
+import com.google.mlkit.nl.translate.TranslateRemoteModel;
+import com.google.mlkit.nl.translate.Translation;
+import com.google.mlkit.nl.translate.Translator;
+import com.google.mlkit.nl.translate.TranslatorOptions;
+
+public class MlKitTranslateUtil {
```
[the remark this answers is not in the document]
eddiemuc: If the interface is specific to the underlying implementation then the interface is probably not optimal and should be changed. That's the point of an interface, it should abstract from implementation.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -253,6 +256,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     private HtmlStyle descriptionStyle = HtmlStyle.DEFAULT;
 
+    final public MutableLiveData<MlKitTranslateUtil.Language> listingTranslationLanguage = new MutableLiveData<>();
```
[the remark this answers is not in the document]
eddiemuc: I won't argue further wrt usage of mutablelivedata. Usage example for the "traditional" pattern can be found through the code, usually by searching for usage of methods starting with "refresh" (I am not on my computer right now).

I disagree however in not saving the user state. It is bad enough that we don't restore state correctly in other situations, those have to be considered as bugs. No reason to introduce bugs in new features.

eddiemuc:
Wrt reusage capability: in suggest to also use the new translation options for trackable listing in this one PR. If everything related to translations (including eg dialogs for errors like "language not supported" or download handling) is indeed encapsulated in Utils class, then this should be super easy and with only very few new code.

ztNFny:
> Wrt reusage capability: in suggest to also use the new translation options for trackable listing in this one PR. If everything related to translations (including eg dialogs for errors like "language not supported" or download handling) is indeed encapsulated in Utils class, then this should be super easy and with only very few new code.

Will look into it, should be a quick change once i get the chance to start the IDE again.
I always forget about trackables even existing 😁 I even removed all the TB related UI from my builds.

ztNFny:
I require some feedback on the language detection: https://developers.google.com/ml-kit/language/identification/android
There's 2 ways to include this:
- Bundled: App size increases ~900KB
- Unbundled: App size increases ~200KB, the model is then downloaded on first use

I'd go for bundled as I don't like apps just downloading stuff

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -0,0 +1,368 @@
...
+                            RemoteModelManager.getInstance()
+                                    .download(new TranslateRemoteModel.Builder(lng.getCode()).build(), new DownloadConditions.Builder().build())
+                                    .addOnSuccessListener(s -> Toast.makeText(context, context.getString(R.string.translator_model_download_success, lng.toString()), Toast.LENGTH_SHORT).show())
+                                    .addOnFailureListener(e -> Toast.makeText(context, context.getString(R.string.translator_model_download_error, lng.toString(), e.getMessage()), Toast.LENGTH_LONG).show());
+                        }
+                    });
+        });
+    }
+
+    public static void initializeListingTranslatorInTabbedViewPagerActivity(final TabbedViewPagerActivity cda, final LinearLayout translationBox, final String translateText, final Runnable callback) {
```
eddiemuc, line 217: Parameter cda seems to be used only to retrieve the Status (and one time to do a translation, but for this LocalizationUtils can be used which doesn't need a `Context` or `Activity`). Why not replacing it with a parameter of type Status? Or maybe make this whole method a part of the "Status" class?

Suggest to rename "translateText" to "languageDetectionText"? (I figured from code that this is what it is for?)

In order to abstract this method from usage in GUIS not being a "TranslateBox" model: what about passing a TextView (for note) and a Button (for click action) parameter instead. Make both nullable.

IMHO the callback here should already get the fully initialized Translator object (resp a Lambda for it, see later on this). This also means that this method should already handle all necessary checks and download if necessary (including progress dialog and such).

main/src/main/java/cgeo/geocaching/activity/TabbedViewPagerActivity.java
```diff
@@ -37,6 +38,7 @@ public abstract class TabbedViewPagerActivity extends AbstractActionBarActivity
     private long[] orderedPages;
     private ViewPager2 viewPager = null;
     private Action1<Long> onPageChangeListener = null;
+    public OfflineTranslateUtils.Status translationStatus = new OfflineTranslateUtils.Status();
```
eddiemuc, line 41: I assume this could be "protected". I would actually put it directly in using Activity, but it is ok to have it here.

Still unsolved: values of this object need to be preserved on Activity recreation (using save bundle) eg on screen orientation change.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -0,0 +1,368 @@
...
+            return code.hashCode();
+        }
+
+        @Override
+        public int compareTo(@NonNull final Language lng) {
+            return this.toString().compareTo(lng.toString());
+        }
+    }
+
+    public static class Status {
```
eddiemuc, line 301: You use Status as member type in an activity and thus have to save and recreate it there eg on screen orientation change. 

You might get in trouble here because of objects like MutableLiveData, TranslationProgressHandler and a Consumer because those are not serializable/parcelable. Better move those things to another class which can be recreated from a "status" object on activity (re)start

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1848,46 @@ public void setContent() {
 
         }
 
+        private void translateListing() {
+            final CacheDetailActivity cda = (CacheDetailActivity) getActivity();
+
+            if (cda.translationStatus.isTranslated()) {
```
eddiemuc, line 1854: I don't understand this if statement.

If I understand the code of "Status" correctly, then translateListing() is supposed to be called (via callback) whenever a new translation is needed (eg because source or target language changed or necessary download is done etc etc) and can be performed.
* Why is then logic here to check and maybe correct translation statusÜ
* Why is the method left when translation is done but at the same time "translation not done" marked?
* What is the meaning of the strange logic for "hint" here?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1847,6 +1848,46 @@ public void setContent() {
...
+                } else {
+                    binding.hint.setText(cache.getHint());
+                }
+                return;
+            }
+
+            final OfflineTranslateUtils.Language sourceLng = cda.translationStatus.getSourceLanguage();
+            cda.translationStatus.startTranslation(2, cda, cda.findViewById(R.id.description_translate_button));
+
+            OfflineTranslateUtils.getTranslator(cda, sourceLng,
```
eddiemuc, line 1868: IMHO, the callback from Status should already pass the fully initialized Translator object. I will comment more on this on the Status object.

eddiemuc (changes requested):
Code is still not stable against recreation of Activities (eg screen orientation change). That would from my side be the only change I deem mandatory.

All other comments are only recommendations from my side. We have very different coding styles :-) but c:geo is pretty colorful in this regard anyway.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
ztNFny: This is the "if already translated restore the original, untranslated listing".

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
ztNFny: I understand your point about restoring it. I disagree with the priority of it. 
Also I don't know how to do it and will not have the time to figure that out soonish. So if that is a requirement I'll have to abandon this PR for the next weeks.
Anyone is free to contribute that part of course.

eddiemuc: That's fair. Using the "Status" approach you have chosen I too don't see an obvious way to accomplish this. In other places I created "Activity helper"s to accomplish this (eg Image-Selektion and SAF), but this would mean a major refactoring.

I will leave decision on how to proceed to the dev team.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
ztNFny: see my comments in the previous round of review - this is the GUI stuff specific to Cache/Trackable actvitity, it contains nothing but GUI, no way to make it more abstract as such.

main/src/main/java/cgeo/geocaching/activity/TabbedViewPagerActivity.java
ztNFny: no, needs to be public for the lambdas

ztNFny:
all but the storing state are addressed. I'd propose to merge like this and then maybe someone else has the chance to take a look at storing that - I need to take care of t-shirts now ;-)

eddiemuc:
> all but the storing state are addressed. I'd propose to merge like this and then maybe someone else has the chance to take a look at storing that - I need to take care of t-shirts now ;-)

There are also still test failures and codacy warnings

moving-bits:
Nice feature, love it. Some suggestions I have:
- As translation does not preserve HTML fully (losing links and images along the way), I propose to store the translated text into a separate database field and offer the user an easy way to switch between those two. Maybe a new tab for the translated text?
- The translation box text ("Translate from ...") needs to be updated according to state:
  - If the target language gets downloaded, text reads "downloading ...", which is fine. But the text does not get updated on having finished the download.
  - If a cache listing got translated, the text still reads "Translate from ...". The box should either be removed on having finished the translation, or should read something like "Text got machine translated from ...", or could be used to toggle between original listing text and translated text.

ztNFny:
You can already toggle it.
And why store it in DB if it's an ofine translation? That's useless!

moving-bits:
> You can already toggle it.

Ok, seems that by tapping the translation button again the original text is being displayed (with links and image being intact). Then no offline storage is necessary, but the description text for the button needs improvement definitely, to inform the user about toggling. 

> And why store it in DB if it's an ofine translation? That's useless!

Since we seem to have a working toggle, it's no longer required.

moving-bits:
retest this please

moving-bits:
As discussed in yesterday's dev meeting: Let's merge here and fine-tune step by step, if necessary.

Issue #16613, opened 2025-02-01
Labs being filtered out by D/T filter even with active "Show caches without D/T"


moving-bits:
User report on support about labs not being shown. User has an active D/T filter, which is the cause. But during retesting I stumbled upon that even with an activated "Show caches without D/T" labs are not being displayed. Anything else missing?

c:geo release 2025.01.18
ticket 718417

ztNFny:
Works for me

moving-bits:
Here's a more detailed description of how to reproduce this issue:
- c:geo release 2025.01.18
- open map
- open filter
- my default filter is an extended filter and has cache type (all selected) and status filter ("found" and "found offline" set to "no", everything else set to "all")
- add a filter for "difficulty and terrain", change d or t range, activate "show caches without d/t"
- and return to map

The crucial step seems to be change the d/t range away from the default "1.0 - 5.0" range. With default range, labs are still being displayed, with a different range they aren't anymore.

ReggieDoodle:
Hi. I new on here but just being investigating another filtering issue so happy to look into this if you want?
Dave

moving-bits:
Sure, every contribution is very welcomed

ReggieDoodle:
@moving-bits 

Slight issue - when I run cgeo in the emulator from AS I don't see any labs on the map or in searched lists (no filters applied)

In the services section I have the option activated.   

![Image](https://github.com/user-attachments/assets/34554347-808f-4844-91d1-2086ae4b6ac5)

When setting up AS I edited keys.xml as instructed (README.md). I notice the last line is an ALC_CONSUMER_KEY, which is blank for me - should that be populated ?

`

![Image](https://github.com/user-attachments/assets/9b69618c-7a30-40a9-b10a-3ec017e97723)

`

ReggieDoodle:
Seems Labs have a fixed DT value of -1.0 and the current code is looking for a value of zero when then option is ticked.

ztNFny:
> When setting up AS I edited keys.xml as instructed (README.md). I notice the last line is an ALC_CONSUMER_KEY, which is blank for me - should that be populated ?

For ALC unfortunately there's no official process to request keys (unlike OC, Google Maps, ...) so it's not documented as it doesn't exist. The key we have for the official c:geo builds we can't share.
There are unofficial ways to get hold of a key though - after all all apps implementing that need to communicate their keys to the server ;-)

ReggieDoodle:
[...]

PR #16680, opened 2025-02-24
Add special case for lab caches that have a DT of minus 1

ReggieDoodle:
## Description

Fix issue 16613 so that lab caches are shown when the box is ticked

## Related issues

bug #16613 

## Additional context

Resubmitted as first request was against master and it is meant for the release branch

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/core/NumberRangeFilter.java
+++ b/main/src/main/java/cgeo/geocaching/filters/core/NumberRangeFilter.java
@@ -35,8 +35,13 @@ public class NumberRangeFilter<T extends Number & Comparable<T>> {
     }
 
     public boolean isInRange(final T value) {
-        if (includeSpecialNumber != null && specialNumber != null && isEqualValue(value, specialNumber)) {
-            return includeSpecialNumber;
+        if (includeSpecialNumber != null && specialNumber != null) {
+            if (isEqualValue(value, specialNumber)) {
+                return includeSpecialNumber;
+            }
+            if (specialNumber.compareTo(value) > 0) { // Special case for lab caches
+                return includeSpecialNumber;
+            }
         }
 
         if (minRangeValue != null && minRangeValue.compareTo(value) > 0) {


bekuno:
ok to test

moving-bits:
Thanks @ReggieDoodle for your contribution!

Seems to run fine on the map now, but I cannot make it work in cache lists.

@eddiemuc  / @murggel 
Is there some different "magic" involved for `CacheListActivity` filtering?

My test case:
- open a list which has lab adventures in it
- set or add a D/T filter with at least one range not starting at 0, and tick the "Include caches without D/T" box

Cache list should still show lab adventures now, but does not over here.

eddiemuc:
Further down the class there's a method `asdRangeToSql`. It must build an SQL statement matching the filter algorithm. This is used when loading caches from database. I assume that the method needs to be adapted as well.

ReggieDoodle:
@moving-bits - @eddiemuc 

it worked for me on the map and search->filter->search by filter 

however, I’m not familiar enough with the internals yet to know when it calls sql and when it invokes the GC web search so maybe I need to revisit…

Tks

main/src/main/java/cgeo/geocaching/filters/core/NumberRangeFilter.java
```diff
@@ -35,8 +35,13 @@ public NumberRangeFilter(final Func1<String, T> numberParser, final Func1<Float,
     }
 
     public boolean isInRange(final T value) {
-        if (includeSpecialNumber != null && specialNumber != null && isEqualValue(value, specialNumber)) {
-            return includeSpecialNumber;
+        if (includeSpecialNumber != null && specialNumber != null) {
+            if (isEqualValue(value, specialNumber)) {
+                return includeSpecialNumber;
+            }
+            if (specialNumber.compareTo(value) > 0) { // Special case for lab caches
```
eddiemuc, line 42: This change de facto converts the special number logic from an "equals" logic to a "equal or less than logic".

you need to adjust method "addRangeToSQL" accordingly so generated SQL is also changed from "equals" to "equal or less then" (around code lines 165/165)

Also you're changing a multiple-purpose here for a special case behavior. Was it throughly tested that this new logic also behaves correctly in all other usages of this filter helper?

eddiemuc (changes requested):

eddiemuc:
>  I’m not familiar enough with the internals yet to know when it calls sql and when it invokes the GC web search so maybe I need to revisit…

Simplified each filter class filters for a specific criteria (eg Cache Name or Cache Type). Filters are typically parameterized (eg the cache type filter takes the cache types to filter for). A filter provides following mechanisms for filtering:
- instance check (method "filter"): given an instance of the Geocache class it tells you whether the instance satisfies the filter or not
- sql (method "addToSal"): provides a snippet for a SQL WHERE statement which can be used to retrieve exactly those caches from the database which satisfies the filter
- online filtering (not in filter class but in connectors): provides the necessary means to ask an (online) connector for exactly those caches satisfying this filter 

"Instance check" and "SQL" are implemented in each filter class. The "online filtering" is highly specific for each connector and thus implemented there. Connectors typically only support a small subset of all filters available in c:geo

The mechaniam above are used in different scenarios:
- database retrieval (eg when loading lists into memory): in that case the generated SQL is applied for DB retrieval. The retrieved caches are then additionally filtered using instance check
- online retrieval via connector (eg live map or live search): in that case the connector applies filters as best as possible when issuing the online request. The result is then additionally filtered using instance check 
- single cache filtering using instance check in many places in code

ReggieDoodle:
@eddiemuc 

Thanks for the insight.  The OP was to allow the user to load caches in a certain DT AND all lab caches.  A strange scenario and the only common area called for both data types I could find was inside the range filter as they are both loaded differently?

Normal caches are loaded through SQL or the web search but labs are loaded by a special call to the ‘not so public’ labs API using what they call the V4 search command. This has no filtering and does not use SQL so is it necessary to proliferate this specific test elsewhere ?

To me, the ‘special number’ is the problem here - it’s hard coded to zero to allow the original use case of including caches without a DT value (not sure how you get one of them in the system anyway?) through the filters.   Maybe that approach needs a rethink?

tks
Dave

eddiemuc:
Hi @ReggieDoodle 

> Normal caches are loaded through SQL or the web search but labs are loaded by a special call to the ‘not so public’ labs API using what they call the V4 search command. This has no filtering and does not use SQL

Lab caches are handled the same way as every other cache in c:geo. 

The Lab API is encapsulated in a connector (the ALConnector) and this connector is used for online searches like every other connector. Once a Lab cache is downloaded it may be stored into the c:geo database as part of a list. When this list is then loaded from the database, then also the contained lab caches are queries via SQL.

ReggieDoodle:
Ok, thanks for that info.

So back to the underlying problem - the special  number.   When can a cache have zero D or T ratings ?

ReggieDoodle:
@eddiemuc 

I'm just revisiting this after working on another project 

You say I'd be changing multi purpose code for a special case but isn't the whole concept of the use the specialNumber a special case in itself to handle a situation where caches can exist in the system without a DT value - **when does that happen please** ?

Another option might be rather than load labs and assign them DT of minus one why not assign them zero then the filters don't need to be changed.   **Is the a specific reason they are minus one ?**

I see this is done in the main Geocache class thus 

```
public float getDifficulty() {
        return getConnector().supportsDifficultyTerrain() ? difficulty : -1;
 }

```
where supportsDifficultyTerrain() is set false in the ALConnector but also by the InternalConnector which looks to be used a lot


Tks
Dave

ReggieDoodle:
Thanks @eddiemuc -

I still do not see where data gets in to the system with zero but if you are happy with the change 'to equals or less than' I will implement it in the sql p, remove the comment and check all other uses - shame I still can't get unit tests to run locally...

Think the second option though probably a proper fix is maybe too much for me to take on until I have a clearer understanding of the system.

Tks 
Dave

murggel:
How about having a list of special numbers instead of a "range"?
So other usages of the special number will not be affected.

eddiemuc:
> I still do not see where data gets in to the system with zero ...

I guess zero "leaks" into the system in all places where a Geocache is created and difficulty/terrain is NOT explicitely set on it. In that case it will default to zero, and those values will be stored.

I am not too happy with this either... there is actually lots of stuff which would benefit from some refactoring. But who has the time? :-)

eddiemuc:
@ReggieDoodle 

To do a database migration, you have tp create a new "database version". You need to follow these steps (which are all done in file` DataStore.java`):
* increase the vale of field "dbVersion" by one (that field is defined in line 259 and currently has the value 105. You need to change it to 106)
* add "106" to the field DBVERSIONS_DOWNWARD_COMPATIBLE which starts at line 279. This tells c:geo that the database upgrade from version 105 to 106 is "downward compatible"
* Now the major part: navigate to the END of Method DataStore.DBHelper.onUpgrade (end is around line 1921). You will notice in this very long method a lot of statements like "if (oldVersion < 105)", and then some database migration logic inside this "if" statement (either data changes or database structure changes like adding a DB field). Create a similar if statement "if (oldVersion y 106)" and place the data migration logic inside. In you case this would be two SQL UPDATE statements: "UPDATE cg_caches SET difficulty = 0 WHERE difficulty <= 0" and  "UPDATE cg_caches SET terrain = 0 WHERE terrain <= 0"

ReggieDoodle:
Sounds ok - will take a look tmrw

Dave

ReggieDoodle:
Work will be done under ticket #16712

ztNFny:
> The more I look at this the more I think my second suggestion of changing the labs DT to zero make sense rather than have a double check on the special number. 

There are today 2 different ratings for a reason: 0 = unknown, -1 = unsupported.
We can't just eliminate one of the 2 without loss in functionality. See https://github.com/cgeo/cgeo/issues/16613#issuecomment-2643731845

ztNFny:
>> Another option might be rather than load labs and assign them DT of minus one why not assign them zero then the filters don't need to be changed. Is the a specific reason they are minus one ?
> No reason, and you are right. Don't know why sometimes 0 is used and sometimes -1

There is a reason - and I guess the discussion took a wrong turn based on this comment.

Let's first change the values to constants to avoid this in the future:
public final static int RATING_UNKNOWN = 0;
public final static int RATING_NOT_SUPPORTED = -1;

ReggieDoodle:
Just an observation.  The listed position of an adlab set loads into cgeo with a D/T rating of 1/1 so why don't the stages ?

eddiemuc:
> > > Another option might be rather than load labs and assign them DT of minus one why not assign them zero then the filters don't need to be changed. Is the a specific reason they are minus one ?
> 
> > No reason, and you are right. Don't know why sometimes 0 is used and sometimes -1
> 
> There is a reason - and I guess the discussion took a wrong turn based on this comment.
> 
> Let's first change the values to constants to avoid this in the future: public final static int RATING_UNKNOWN = 0; public final static int RATING_NOT_SUPPORTED = -1;
> 
> Unknown is on partial data availability, e.g. when doing a live mapping of a PQ. Those caches have a DT rating but we don't know it. Labs and UDCs however do not support that at all. This is different. In terms of icons one is displayed as "?", the other as "-".

Being PQ probably the only place where a D/T rating could be unknown (for "normal" partial live update it is known) I took a look at PQ preview on the web page. I found that D/T is apparently included there, see following screenshot of the table header:
image

 Wouldn't it be an option to "fix" PQ loading to include the D/T and thus eliminating the reason to differentiate between NOT_SUPPORTED and UNKNOWN for D/T? 
(again, I am not per se against such a differentiation, but I wonder why D/T should be such a special case about it. If we need this differentiation why isn't it down anywhere else?)

ztNFny:
That table isn't used for preview of PQ though, it's a map call and (at least at the time it was implementated) that didn't include DT. That table doesn't include any position info so it's be quite useless as a preview source.

If we manage to not have any "unknowns" would be great of course. Not sure whether there are other sources for unknowns though.

ReggieDoodle:
Guys

I had a play this morning and introduced the concept of a second specialNumber in the DT filters  (that is the only place the NumberRangeFilter is used)

It worked on the map and a live filter - labs were displayed when the box was ticked.  I didn't touch the SQL for time being but that's an easy change

To do it for real I'd propose renaming specialNumber and my new second one to reflect the constants discussed above ( RATING_UKNOWN & RATING_UNSUPORTED) but let me know if you believe is a viable approach

Tks

ztNFny:
The differentiation between -1 and 0 seems to be used only in MapMarkerUtils so far. I could prepare a PR to move the differentation between the values there, returning only 0 to all consumers of getDifficulity()

ReggieDoodle:
> The differentiation between -1 and 0 seems to be used only in MapMarkerUtils so far. I could prepare a PR to move the differentation between the values there, returning only 0 to all consumers of getDifficulity()

So would labs then display zero for DT?

ztNFny:
> The differentiation between -1 and 0 seems to be used only in MapMarkerUtils so far. I could prepare a PR to move the differentation between the values there, returning only 0 to all consumers of getDifficulity()

https://github.com/cgeo/cgeo/pull/16715

eddiemuc:
> > The differentiation between -1 and 0 seems to be used only in MapMarkerUtils so far. I could prepare a PR to move the differentation between the values there, returning only 0 to all consumers of getDifficulity()
> 
> see https://github.com/cgeo/cgeo/pull/16715 for an alternative approach

PR #16715 does not fix issue #16613 (which is about filtering). It does, however, opens the way for the solution suggested by @ReggieDoodle :
- make sure that no code assigns -1 an more ever to difficulty or terrain field, only 0 (this might already be the case)
- make sure the magic number used in number range filter is zero (this might already be the case)
- implement database migration as discussed further above 

I will merge PR #16715 in a minute

eddiemuc:
@ReggieDoodle if you're still willing to do this (wouldn't blame you if this forth and back frustrates you... very sore for that), then please do it against current master because this is where @ztNFny aimed PR #16715 to

ztNFny:
> > > The differentiation between -1 and 0 seems to be used only in MapMarkerUtils so far. I could prepare a PR to move the differentation between the values there, returning only 0 to all consumers of getDifficulity()
> > 
> > see https://github.com/cgeo/cgeo/pull/16715 for an alternative approach
> 
> PR #16715 does not fix issue #16613 (which is about filtering). It does, however, opens the way for the solution suggested by @ReggieDoodle :
> - make sure that no code assigns -1 an more ever to difficulty or terrain field, only 0 (this might already be the case)
> - make sure the magic number used in number range filter is zero (this might already be the case)
> - implement database migration as discussed further above 
> 
> I will merge PR #16715 in a minute 

Uhm, why do you think it doesn't fix the issue? It removes the place where -1 is returned as rating, thus returning 0 and thus making the filter work.

eddiemuc:
> Uhm, why do you think it doesn't fix the issue? It removes the place where -1 is returned as rating, thus returning 0 and thus making the filter work.

There will still be offline stored caches in DB where the value is -1. Current SQL filter won't return those if it checks for zero only. This is what Dave's suggested change was all about

ReggieDoodle:
I can do the database migration if you think its necessary. Its definitely
an edge case now. Can't image many users have a stored list including ALs
that they want to filter on DT but guess it ties off the loopholea

ReggieDoodle:
> There will still be offline stored caches in DB where the value is -1. Current SQL filter won't return those if it checks for zero only. This is what Dave's suggested change was all about

Ok, guys - just done a test with current master and can confirm lab caches are shown for live filters (on map and via a live search) now when the box is ticked so that fixes #16613 for live data - **thank you**

For an offline list that was created prior to pulling master the filter behaves as before so that would benefit from the db update however for a newly created list if I use this filter (with box ticked) 

image

I do **not** see labs and I would expect to do so 

But if I use this one (ie filter in all DT values) and DO NOT tick the box I see labs, which is not what I would not expect if they now have DT of zero

image


I will look at the SQL in a debugger ....

ReggieDoodle:
I did some more tests this morning

For a live filter the change made by @ztNFny fixes the issue 

But there is definitely an issue with stored lists and the SQL builder in this situation  

Using a new stored list on master which contained saved labs and other caches I get the following (only showing the pertinent part of the select statement as its huge)

1. Full D & T range with special ticked - does shows labs which is correct 
AND (((t.difficulty = 0.0) OR ((1=1))) AND ((t.terrain = 0.0) OR ((1=1)))) 

2. Selected D range no special ticked - does not show labs which is correct
AND (((t.difficulty >= 2.0)) AND (1=1)) 

However

3. Selected D range with special ticked - does not show labs but **it should do**
AND (((t.difficulty = 0.0) OR ((t.difficulty >= 2.0))) AND ((t.terrain = 0.0) OR ((1=1))))

4. Selected D & T range with special ticked - does not show labs but **it should do** 
AND (((t.difficulty = 0.0) OR ((t.difficulty >= 2.0) AND (t.difficulty <= 4.0))) AND ((t.terrain = 0.0) OR ((t.terrain >= 2.0) AND (t.terrain <= 4.0)))) 

Item 4 was the original use case in the ticket - when the range is not full (1-5) the labs aren't shown when they should - but the reporter didn't say they were using a stored list. This really is an edge case now


Thoughts please - is it worth fixing and doing a DB update path for this situation?

moving-bits:
> Thoughts please - is it worth fixing and doing a DB update path for this situation?

Examples 3 and 4 do fail currently as old lab adventures are stored with D/T -1 in the database, so the conditions "difficulty = 0.0" resp. "terrain = 0.0" won't get triggered. 

I would definitely vote for updating those. Doing a database update is pretty lightweight and downward compatible due to our framework. If you don't want to dive into this, just drop me a note, I'll implement it then.

ReggieDoodle:
Happy to do it.  Need the experience 😉

However, my test was against a saved list where labs already had zero D and
T so I need to look at why the swl isn't returning as expected.  I pasted
into sqlite and it worked fine .....

ReggieDoodle:
Al good now - I have made a DB update path and tested the filters on a stored list and all ok.

This can be closed once PR#16742 is merged into master

PR #16683, opened 2025-02-26
Follow my location in elevation chart (fix #15763)

### LINKED ISSUE #15763: Display current location on elevation graph, plus some metrics
### Tell us your idea!

c:geo now has the ability to display elevation graphs for routes (and soon for tracks too, see  #15683).

With this cool new feature, it would now make sense to also see how far along such a graph one has already come, e.g. like this:

grafik

All this would require would be to project the current position onto the displayed track/route (probably this only makes sense when we are not more then, say, 100m away from the track/route).

It would also allow visualising the remaining route/track distance. There are certainly multiple ways to do this; one nice way could be to use the labels of the x-axis as in my mockup drawing.

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Elevation charts "follows current location", marking the closest nearby point on the current track and adapting the x-axis labels to it. Maximum allowed distance between current location and point on track is 100m. Minimum distance between two updates is 50m.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/Units.java
+++ b/main/src/main/java/cgeo/geocaching/location/Units.java
@@ -60,9 +60,9 @@ public class Units {
 
         final ImmutablePair<Double, String> scaled = scaleDistance(distanceKilometers);
         final String formatString;
-        if (scaled.left >= 100) {
+        if (Math.abs(scaled.left) >= 100) {
             formatString = "%.0f %s";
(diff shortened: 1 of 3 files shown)


ztNFny:
I'd suggest to change the crosshair color to avoid creating a mess of lines - R.color.colorTextHint seems reasonable

moving-bits:
Good idea, done - looks like this now:

image

PR #16698, opened 2025-03-01
fix #16557: add preventAskForListDeletion back to database query

### LINKED ISSUE #16557: Dialog to remove empty list reappears despite the user setting "Do not ask me again"
Bug reported by [Ticket #947418](https://cgeosupport.droescher.eu/scp/tickets.php?id=8963), confirmed by me
c:geo problem report (2025.01.09-NB-90d440f)

c:geo has the feature of asking if a list should be removed if it is empty. In this dialog, user can select "do not ask me again"; in that case user should never be asked again. 

image

In some situations however, the dialog reapperas again for the same list. One scenario I could reproduce is the following:
* Create a new list and assign one cache to it 
* Go to list, look at it
* Open the one-assigned cache, remove it from the list, then go back to list view using "back" button
* --> c:geo asks whether now-empty-list should be removed --> correct behaviour
* Check "do not ask me again" and tap "No" --> empty list stays
* Switch to another app (eg email, your favorite messenger or whatever), then switch back to c:geo
* --> c:geo asks AGAIN whether empty list should be removed --> ERROR

eddiemuc:
fix #16557: add preventAskForListDeletion back to database query

### DIFF
--- a/main/src/main/java/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DataStore.java
@@ -4499,7 +4499,9 @@ public class DataStore {
             }
 
             try {
-                final String query = "SELECT l._id AS _id, l.title AS title, l.emoji AS emoji, COUNT(c.geocode) AS count" +
+                final String query = "SELECT l._id AS _id, l.title AS title, l.emoji AS emoji," +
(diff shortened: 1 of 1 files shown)


papenmap:
I don't know if this case is closed now, but in my latest version the issue is still occurring. 
Screenshot_20250309_083900_cgeo

PR #16701, opened 2025-03-01
fix overflow menu icon color #16619

### LINKED ISSUE #16619: dark icons in dark theme in overflow menu
Light theme "sometimes" shows dark icons in toolbar (instead of white):
Image

I haven't yet figured out steps to reproduce, but after some time of using c:geo this starts happening.

I thought it'd be a bug in tinting of menu icons I introduced in MenuUtils:tintToolbarAndOverflowIcons(), but it's also happening in activities where that method isn't called at all, e.g. in MainActivity

ztNFny:
fixes #16619

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/MenuUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/MenuUtils.java
@@ -4,11 +4,14 @@ import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
 
 import android.annotation.SuppressLint;
+import android.content.Context;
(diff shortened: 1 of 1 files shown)


ztNFny:
restest this please

papenmap:
I don't know if this means that the case is closed now, but in my latest version (2025.03.02-NB-9a454de) the menu icons are still dark on dark background. 
Screenshot_20250309_082208_cgeo

moving-bits:
The fix got applied after the date the nightly version you are using got built, so it's not included there yet. We have a technical problem currently that prevents creating new nightly versions - we are working on it.

Issue #16635, opened 2025-02-08
java.lang.IllegalStateException on setting view in SeekbarPreference


moving-bits:
For current release 2025.01.18 this is accountable for about 50% of our crashes:

```
Exception java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.
  at android.view.ViewGroup.addViewInner (ViewGroup.java:5477)
  at android.view.ViewGroup.addView (ViewGroup.java:5296)
  at android.view.ViewGroup.addView (ViewGroup.java:5236)
  at android.view.ViewGroup.addView (ViewGroup.java:5208)
  at cgeo.geocaching.settings.SeekbarPreference.setExtraView (SeekbarPreference.java:192)
  at cgeo.geocaching.settings.SeekbarPreference.onBindViewHolder (SeekbarPreference.java:186)
  at androidx.preference.PreferenceGroupAdapter.onBindViewHolder (PreferenceGroupAdapter.java:422)
  at androidx.preference.PreferenceGroupAdapter.onBindViewHolder (PreferenceGroupAdapter.java:49)
  at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder (RecyclerView.java:7747)
  at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder (RecyclerView.java:7847)
  at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline (RecyclerView.java:6646)
  at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline (RecyclerView.java:6917)
  at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition (RecyclerView.java:6757)
  at androidx.recyclerview.widget.RecyclerView$Recycler.getViewForPosition (RecyclerView.java:6753)
  at androidx.recyclerview.widget.LinearLayoutManager$LayoutState.next (LinearLayoutManager.java:2362)
  at androidx.recyclerview.widget.LinearLayoutManager.layoutChunk (LinearLayoutManager.java:1662)
  at androidx.recyclerview.widget.LinearLayoutManager.fill (LinearLayoutManager.java:1622)
  at androidx.recyclerview.widget.LinearLayoutManager.onLayoutChildren (LinearLayoutManager.java:687)
  at androidx.recyclerview.widget.RecyclerView.dispatchLayoutStep2 (RecyclerView.java:4645)
[...]

PR #16707, opened 2025-03-03
update mapsforge and VTM to 0.24.0

ztNFny:
updates both mapsforge and VTM

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -367,7 +367,7 @@ dependencies {
 
     // Mapsforge official version
     String mapsforgeSource = 'org.mapsforge'
-    String mapsforgeVersion = '0.23.0'
+    String mapsforgeVersion = '0.24.0'
 
     // Mapsforge Snapshot from official master branch (via https://jitpack.io/#mapsforge/mapsforge)
     // String mapsforgeSource = 'com.github.mapsforge.mapsforge'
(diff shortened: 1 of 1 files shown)


moving-bits:
Please target this to release branch, which is currently using the faulty v0.23 VTM

ztNFny:
rebased

moving-bits:
retest this please

moving-bits:
Thanks @ztNFny for rebasing. The last build threw errors similar to what I got yesterday when trying to upgrade to a snapshot version. Let's see what the rebuild says, whether it's a temporary hiccup, or if it's not compatible with our release branch. (If the latter, we might need to downgrade VTM on release to the version we've used before 0.23)

moving-bits:
Ok, I did a retest on a second machine locally: 

c:geo compiles fine with this PR when being applied to `master`, but fails to do so when targeted to `release`. Some comments on StackOverflow hint to AGP version (which we have upgraded on `master` but not yet on `release`, IIRC). I would not change this on `release` now, though.

Thus we should revert this PR to be targeted against `master` (sorry, @ztNFny, for the extra work involved), and I'll create a PR for `release` to downgrade VTM to a pre-0.23-version (which does not include the tesselation error and seemed to be running fine before upgrading to v0.23).

ztNFny:
Can you do it? Wine have time to rebass this week

moving-bits:
> Can you do it? Wine have time to rebass this week

sure, will do

bekuno:
The problem should be fixed with AGP 8.0. (see eg. #14738)
We can cherry-pick Commit 35c42c9.

But we have now Issue #16709.
So we have to wait until it is fixed, because also the release job is running on CI node master.

moving-bits:
> The problem should be fixed with AGP 8.0. (see eg. #14738) We can cherry-pick Commit [35c42c9](https://github.com/cgeo/cgeo/commit/35c42c96eb1be5ffbbd3944d38a154a474771f6f).

Yes, I guess so from StackOverflow comments. But as I would like to release a bugfix release rather soon (due to the significant amount of crashes because of #16635), I'm not sure whether I want to switch to a new AGP on `release` already.

moving-bits:
superseeded by #16711

PR #16715, opened 2025-03-05
handle cache connectors not supporting rating in MapMarkerUtils only

### LINKED ISSUE #16613: Labs being filtered out by D/T filter even with active "Show caches without D/T"

ztNFny:
Alternative approach to fixing #16613
Changes Geocache class to not differentiate between connectors supporting D/T ratings but always return the default rating (0) and move the differentation into MapMarkerUtils

fixes #16613


eddiemuc:
Does not fix #16613 but prepares codebase for such a fix

Issue #16193, opened 2024-10-09
User note in ALC waypoint is deleted


tinkus-gc:
### Describe your problem!

It happens to me quite often that the user notes in the waypoints of an ALC are deleted. I still couldn't find out the situation when this happens. But it's annoying because I store some information there which is then lost. I often have a "backup" in the personal notes but for ALCs these cannot be uploaded to geocaching.com and are sometimes also lost if I'm not very careful.

### How to reproduce?

cannot reproduce

### Expected result after these steps?

My notes shouldn't be deleted!

### Reproducible

No

### c:geo Version

2024.09.18

### System information

```text
## System information

c:geo version: 2024.09.18

Device:
-------
- Device type: Pixel 8a (akita, google)
- Available processors: 9
- Android version: 14
- Android build: AP2A.240905.003.A1
- Screen resolution: 1080x2153px (411x820dp)
- Pixel density: 2.625
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 24.38.34 (190400-681941525)
- Memory:  Available:1.7 GB, Total:7.4 GB, Threshold: 216.0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:334
- Set language: en
- System date format: M/d/yy
- Time zone: GMT+02:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: Oct 4, 12:39 PM
- Last auto backup: Oct 9, 12:12 PM
- Routing mode: Straight (shortest.brf / trekking.brf / car-eco.brf / car-fast.brf / fastbike.brf)
- Live map mode: false
- Use unified map: false
- OSM multi-threading: true / threads: 4
- Map: Combined (Offline)
 - Id: cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider$OfflineMultiMapSource
[...]

PR #16716, opened 2025-03-06
LC: Preserve user note on refresh

baiti:
There are no related issues, however, while I was investigating: #16193 I noticed that the stage number in LCs is now shown twice. This is just a change to remove the now unnecessary stage number which was inserted into the stage text at a time where the "white dots" with those numbers weren't there yet.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -404,7 +404,7 @@ final class ALApi {
         for (final JsonNode wptResponse : wptsJson) {
             stageCounter++;
             try {
-                final Waypoint wpt = new Waypoint("S" + stageCounter + ": " + wptResponse.get(TITLE).asText(), WaypointType.PUZZLE, false);
+                final Waypoint wpt = new Waypoint(wptResponse.get(TITLE).asText(), WaypointType.PUZZLE, false);
                 final JsonNode location = wptResponse.at(LOCATION);
                 final String ilink = wptResponse.get("KeyImageUrl").asText();
                 final String desc = wptResponse.get("Description").asText();
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -4,6 +4,7 @@ import cgeo.geocaching.R;
 import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.gc.GCLogin;
 import cgeo.geocaching.enumerations.CacheSize;
+import cgeo.geocaching.enumerations.LoadFlags;
 import cgeo.geocaching.enumerations.LoadFlags.SaveFlag;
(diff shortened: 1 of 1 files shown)


baiti:
In my second commit I am trying to address #16193. @moving-bits: I can't test this while on the road. At least it compiles. Could you review the changes and perhaps test them briefly?

Here is how:

- Load any LC, perhaps one that wasn't loaded before. 
- Put some text into personal notes
- Refresh the LC, check whether the text is still there. It should, at least if my understanding of the code is correct.

baiti:
Update: I have now indeed managed to actually run that code on the emulator. Steps took forever on the 8GB Laptop but finally worked. I have verified that the personal note now does not get wiped out on a cache reload / refresh.

Whether that ulltimatey solves the referenced issue I can't say.

baiti:
And another update: I had to backout the change that removed "Sx: stagename" because at least in the code I had on the emulator, these nice circle icons with the stage number in it are not rendered. I have no idea why. In fact I was surprised to see it rendered on my production phone. Why would that have a function that isn't in "master"?

Well anyway, we can resolve that later, so the PR is now only to address the missing personal notes.

ztNFny:
> these nice circle icons with the stage number in it are not rendered.

some stupid phone vendors chose to put a font that has those emoji as white instead of dark ... and then they're invisible
I should probably look into replacing them with vectors after all ...

baiti:
> > these nice circle icons with the stage number in it are not rendered.
> 
> some stupid phone vendors chose to put a font that has those emoji as white instead of dark ... and then they're invisible I should probably look into replacing them with vectors after all ...

Is that the reason? My real phone (wihich renders the icons you see on the screenshot) is a Pixel 7, the one in the emulator (which does not show the icons) is a PIxel 3), both are from Google so I have a hard time to believe its the vendor causing the issue. In the emulator I see this:

image

So this is not a font problem. The code I run in the emulator renders something different than what my real phone's current cgeo renders (2025.01.18)

As long as we don't understand the details of this, I cannot remove the "S1" from the text string.

ztNFny:
That is a non-linear lab. Stage numbers are only shown for linear ones

baiti:
> That is a non-linear lab. Stage numbers are only shown for linear ones

Ahh that explains it! Thanks. However, as you can see, I render stage numbers anyway, just sequencing them in the order the owner editet them into the lab builder. Would you say I shouldn't do so? Or would it make sense if you sequenced them too? Unconditional? Linear as well as non-linear?

ztNFny:
Didn't we have A,B,C, ... for unordered and 1,2,3 for linear? (At least that's what I have in my tool ;-))

ztNFny:
> > these nice circle icons with the stage number in it are not rendered.
> 
> some stupid phone vendors chose to put a font that has those emoji as white instead of dark ... and then they're invisible I should probably look into replacing them with vectors after all ...

https://github.com/cgeo/cgeo/pull/16719

moving-bits (approved):
From a quick test: LGTM, thanks for fixing this issue

PR #16737, opened 2025-03-09
Revert updates for AGP & Mapsforge to be Java 11-compatbile for testing purposes

moving-bits:
## Description
As discussed in today's dev meeting: 
Temporarily revert AGP & MF updates to become Java 11-compatible again to test next nightly with Java 11.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         // these dependencies are used by gradle plugins, not by our projects
 
         // Android gradle plugin
-        classpath 'com.android.tools.build:gradle:8.0.2'
+        classpath 'com.android.tools.build:gradle:7.4.2'
 
         // un-mocking of portable Android classes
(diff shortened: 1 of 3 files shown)


moving-bits:
(Have reset master executor to Java 11 already)

murggel:
@moving-bits  No crash with reverted AGP for Caches with photos and logs...
So we need to check for new versions of in that context used libraries?

moving-bits:
I've set master ececutor back to Java 17 (without updating AGP or MF) and started a new nightly build. The build is green and is available in the default nightly location.

murggel:
Current nightly is working - no crashes with photos and logging

moving-bits:
> Current nightly is working - no crashes with photos and logging

thx for cross-checking

That leaves the AGP update as probable cause for the observed issues.

PR #16742, opened 2025-03-10
Add database update path for lab cache DT rating

### LINKED ISSUE #16613: Labs being filtered out by D/T filter even with active "Show caches without D/T"

ReggieDoodle:
## Description

Add database upgrade path so set adlabs DT rating to zero

## Related issues

PR #16680 
PR #16715 
Fix #16613

## Additional context

Tested on current master with a stored list made before #16715 was pushed and a list created today

This fixes the original issue reported in #16613


eddiemuc (approved):
Lgtm, thanks for the contribution

eddiemuc:
Ok to test

PR #16765, opened 2025-03-16
Enable move-to / copy-to for all lists (fix #15733)

### LINKED ISSUE #15733: Add "move multiple caches to another list" function to list opened by "show as list"
### Describe your problem!

When selecting multiple caches in the list view, I was unable to find any way to move them to another list.

It seems that in the current build, only single caches can be moved one by one. Or it is so well hidden that despite me actively searching the feature I could not find it.

### How to reproduce?

See above.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2024.04.25

### System information

_No response_

### Additional Information

_No response_

murggel:
## Description

Adds "move multiple caches to another list" function to list opened by "show as list" 

Behavior will be:
Move: remove from existing list-assignment and add only to the new one
Copy: keep existing list assignment and add to the new list

The entries are only shown, if there are stored caches in the list.
If no stored caches are in the list, the entries for storing the caches offline should be used instead.
In case of mixed caches (stored and unstored caches), the unstored caches will be downloaded and stored in the selected list.

## Related issues

fix #15733

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -582,8 +582,12 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             if (!isOffline && !isHistory) {
                 menu.findItem(R.id.menu_refresh_stored).setTitle(R.string.caches_store_offline);
             }
-            MenuUtils.setVisibleEnabled(menu, R.id.menu_move_to_list, isHistory || isOffline, !isEmpty);
-            MenuUtils.setVisibleEnabled(menu, R.id.menu_copy_to_list, isHistory || isOffline, !isEmpty);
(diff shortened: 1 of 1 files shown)


moving-bits:
I propose to rename the menu entries to make clear, that they will handle stored caches only.
(Or extend the feature, that unstored caches will be downloaded and stored in the selected list.)

murggel:
> I propose to rename the menu entries to make clear, that they will handle stored caches only. (Or extend the feature, that unstored caches will be downloaded and stored in the selected list.)

Ok, then I missed that in the description - unstored caches will be downloaded and stored in the selected list

moving-bits:
> Ok, then I missed that in the description - unstored caches will be downloaded and stored in the selected list
> 
> => updated the description

Thanks

PR #16767, opened 2025-03-16
Show internal connector as active in filter and add display name (fix #14832)

### LINKED ISSUE #14832: Source filter always hides UDC
### Describe your problem!

When selecting a filter for cache source/origin (meaning the geocaching platform) there is no option for user defined caches, thus those are always hidden by the filter.
Even if you select "Show all" they are still hidden.

### How to reproduce?

- Open live map (might be the same on lists)
- Activate advanced filter
- Add filter for cache source
- Select "Show all"

User defined caches not shown on map

Either the filter should contain a source = "User defined caches" or at least if "Show all" is selected it should include UDC

### Reproducible

Yes

### c:geo Version

2023.11.01

### System information

_No response_

### Additional Information

Support ticket 207298

murggel:
## Description

* Show internal connector per default for selectable connectors in filter. `isActive()` method returns true as @eddiemuc suggested.
* Add displayName to IConnector to make connector-name translatable as @fm-sys suggested

## Related issues

fix #14832

## Additional context

I use already existing translation for "User defined caches" (goto_targets_list), but I did not reuse the key, I introduced a new key to be independent.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/AbstractConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/AbstractConnector.java
@@ -423,5 +423,9 @@ public abstract class AbstractConnector implements IConnector {
         return getName().hashCode();
     }
 
-
+    @Override
+    @NonNull
+    public String getDisplayName() {
+        return getName();
+    }
 }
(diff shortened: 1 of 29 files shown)


murggel:
Is there any other place where to use the display name?

Issue #16721, opened 2025-03-07
Print meaningful error message on error 403 while logging


moving-bits:
User request on support (ticket 947769)

On retrieving an error 403 while sending a log, print a meaningful message like "currently not logged in, check authorization" or something similar, instead of just printing the technical details.

Edit: The text is currently something like this:

Posting log failed, 
Reason: It seems that your log was not posted. Please check it on the cache originating website. 
The website responded with following message: GCLoggingManager.createLog: java.lang.RuntimeException: java.io.lOException: request was not successful: Responsefprotocol=http/1.:1, code=403, message=Forbidden, url=https://www.geocaching.com/ /account/oauth/token]

(have not checked, though, if the text depends on user's language setting on the website. One would need to check c:geo source where we print the error code 403 and print a more meaningful message if the function returned that code)

ReggieDoodle:
I will take a look at this next…

ReggieDoodle:
Anyone know how to recreate this specific error?
 

All three me mentioned above could be addressed but unfortunately are trapped in different places in the logging process

moving-bits:
I cannot say how to reliably reproduce the error, but here is some more info from the related support ticket:

- user states that he gets logged out frequently
- he doesn't notice it immediately, but on posting a cache log when he gets the error shown above
- status protocol shows "communication error" as last login error (same day, but still may be unrelated)

ReggieDoodle:
It seems there are several routes that can take the user to the logging failed dialog and all are probably caused by either being logged out or having a poor internet connection so the simplest approach is probably to show a generic message shown in the dialog 

![Image](https://github.com/user-attachments/assets/a610f579-571e-47d3-98e5-082b7ab0076f)

If it is possible to add an expander button to the dialog that could show the actual detail for them that want to know ?

ReggieDoodle:
Code is here https://github.com/ReggieDoodle/cgeo/tree/Fix%2316721

moving-bits:
> I also need someone to give a suitable German translation of the message as well as any other commonly used languages or is it just a case of hitting Google translate?

We use CrowdIn for translations, thus we need to define texts in English only. Users will translate them on CrowdIn, and we can download the translated files later.

ReggieDoodle:
Ok, will make a pull request for you to review

Tks

ReggieDoodle:
Pull request still waiting to be reviewed please 

Tks

ReggieDoodle:
[...]

PR #16769, opened 2025-03-17
Simplify error message when posting log fails

ReggieDoodle:
## Description

Simplifies the message shown to the user when posting a log fails 

## Related issues

Bug #16721 

## Additional context

Tested locally by having false login credentials prior to submitting log

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
@@ -566,7 +566,7 @@ public class LogCacheActivity extends AbstractLoggingActivity implements LoaderM
         } else if (!LogCacheActivity.this.isFinishing()) {
             SimpleDialog.of(LogCacheActivity.this)
                     .setTitle(R.string.info_log_post_failed)
-                    .setMessage(TextParam.id(R.string.info_log_post_failed_reason, statusResult.getErrorString()).setMovement(true))
+                    .setMessage(TextParam.id(R.string.info_log_post_failed_simple_reason))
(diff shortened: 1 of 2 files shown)


bekuno:
ok to test

main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
```diff
@@ -566,7 +566,7 @@ private void onPostExecuteInternal(final StatusResult statusResult) {
         } else if (!LogCacheActivity.this.isFinishing()) {
             SimpleDialog.of(LogCacheActivity.this)
                     .setTitle(R.string.info_log_post_failed)
-                    .setMessage(TextParam.id(R.string.info_log_post_failed_reason, statusResult.getErrorString()).setMovement(true))
```
moving-bits, line 569: Is this the only place where `R.string.info_log_post_failed_reason` gets used?
If so, please remove it from `strings.xml`. If not, please cross-check whether the text needs to be swapped in the other places as well.

ReggieDoodle: Tks

The only other use is in LogViewCreator line 162 when deleting a log fails

Setting false login parameters and trying to delete an existing log pops us a similar verbose 401 message so I think replacing this as per the other usage would be acceptable 

However, as I had added and deleted a lot of logs during testing I did find that I had a cache saved in my offline list in a stale state and when I attempted to delete a log that wasn't actually there anymore (as I had already deleted it on the GC site directly) my new message popped up (which in this scenario was misleading). The phone, in the same scenario showed a verbose 403 message (forbidden). Rare edge case but repeatable so I have added a second new message to cover this situation.

ReggieDoodle:
Tks

Issue #16777, opened 2025-03-22
File type detection issues


moving-bits:
We got a couple of new issues regarding opening files in c:geo lately, which are most likely being caused by the recent changes we have introduced in that area.

Those are related issues AFAICS:
- from Play Store stats for release 2025.03.06: `SecurityException` in `FileTypeDetector.getFileType (FileTypeDetector.java:35)` (> 300 events)
- from Play Store stats for release 2025.03.06: NPE in `FileTypeDetector.getFileType (FileTypeDetector.java:35)` (> 300 events)
- from support (ticket 971441): c:geo opens for mp3 files / cannot play mp3 files anymore
- from support (ticket 377074): opening GPX file from WhatsApp leads to crash
- from support: several tickets stating problems with sharing data (but that may be unrelated)
- from GitHub: #16608

We did not have those issues before the recent changes for local file handling:
#16593, #16600, #16611

Those issues are related to the way an app can declare that it can handle certain file types:
- Filtering by file extension in `AndroidManifest.xml` does not work reliably
- You can filter by MIME type, but most apps don't set a correct MIME type on exporting files
- Very different ways Android apps handle sharing files
- and some side-effects of SAF

Filtering for file extension got some improvement lately with `pathSuffix` and `pathAdvancedPattern`, which may or may not help here (haven't tested it yet), but they are only available starting with API 31.

Therefore my proposal would be:
- Side effects of the changes mentioned above are causing too much harm, IMHO, therefore I would revert them. This would cause local "gwc" and "map" files no longer being detected by c:geo.
- I would then experiment with the new `pathSuffix` option whether I can re-enable detecting local "gwc" / "map" files on API 31+ systems at least (which would be Android 12 or newer).

Any objections?

moving-bits:
PR #16781 implements this approach, currently based on `master`

SammysHP:
We got more requests to make it possible to open local gwc files.

eddiemuc:
> We got more requests to make it possible to open local gwc files.

Workaround for now: copy to wherigo folder. Cartridge can then be opened from within c:geo Wherigo Player

moving-bits:
`SecurityException` still happening for 2025.08.31 as reported by Play Store stats (264 occurrences so far):

```
Exception java.lang.RuntimeException:
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:4280)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:4467)
[...]

PR #16781, opened 2025-03-22
Add API 31+ filtering for local file access

moving-bits:
## Description
- Reverts the changes in `AndroidManifest.xml` that came with #16593, #16600 and #16611 to get rid of the side-effects mentioned in #16777
- adds API 31+ compatible replacements for opening local gpx, gwc and map files

@Lineflyer @ztNFny 
Can you cross-check, if this version works for you and also avoids the problems you have described in #16608?

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -312,6 +312,19 @@
                 <data android:pathPattern=".*\\..*\\..*\\..*\\.map" />
                 <data android:pathPattern=".*\\..*\\..*\\..*\\..*\\.map" />
             </intent-filter>
+            <intent-filter> <!-- Android API 31+ version with better name filtering -->
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <category android:name="android.intent.category.ALTERNATIVE" />
+                <category android:name="android.intent.category.SELECTED_ALTERNATIVE" />
+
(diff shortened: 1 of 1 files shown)


ztNFny:
With this i've got 2 entries to open files and i can't twll any difference between them. Cgeo and cgeo local file.handler

moving-bits:
@ztNFny 
Thanks for cross-checking. This lead me to the idea of adding some more info to the intent receivers we declare in `AndroidManifest.xml`. I have name them as follows:
- GPX receiver (`CacheListActivity`): c:geo GPX import
- Map file receiver (`HandleLocalFilesActivity`): c:geo local map handler
- Wherigo file receiver (`HandleLocalFilesActivity`): c:geo local Wherigo cartridge handler

(All labels to be discussed)

You will still see multiple intent receivers in some cases, as `CacheListActivity` matches quite frequently due to their MIME type definitions, but at least you'll have a chance to make a "knowledgeable selection". Downside seems to be, that the default selection does not have their label being displayed, it just shows "Open with c:geo", but the variants have their label printed:

Example when tapping on a local Wherigo cartridge:

image

(Cannot test with opening from WhatsApp, Signal etc., though, as I don't have that in my test environment. c:geo is programmend to listen for GPX files only in that case - here the MIME type definitions should get triggered.)

moving-bits:
This issue is getting more reports on support & Play Store feedback, so let's merge here. I'll also prepare a version for `release` branch next to get a fix (resp. a revert) into a next bugfix version.

PR #16782, opened 2025-03-22
Prevent downloading log images if flag not set (fix #16778)

### LINKED ISSUE #16778: c:geo downloading log pictures despite having set "store log pictures" to off
Report from user on support (ticket 553165):

Since recently c:geo seems to store log pictures offline despite the user having set the "store log pictures" to off. Test case for him was updating 6,000 caches stored offline, and c:geo suddenly taking 50 GB of memory.

(related to #16699)

moving-bits:
## Description
Prevent immediate downloading of log images on cache update if flag in settings is not set (fix #16778)

(The logs itself will still be analyzed and links to log images be stored, so if you open the image gallery for such a cache, log images start getting downloaded, even if the flag is unset. TBD if this is an intended behavior.)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -1631,7 +1631,7 @@ public final class GCParser {
     }
 
     private static void addImagesFromGallery(@NonNull final Geocache cache, final DisposableHandler handler) {
-        if (StringUtils.isBlank(cache.getGuid())) {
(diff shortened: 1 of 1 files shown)


Jsydog222:
> ## Description
> Prevent immediate downloading of log images on cache update if flag in settings is not set (fix #16778)
> 
> (The logs itself will still be analyzed and links to log images be stored, so if you open the image gallery for such a cache, log images start getting downloaded, even if the flag is unset. TBD if this is an intended behavior.)
>

PR #16790, opened 2025-03-28
Collapse long log entries per default (fix #16789)

### LINKED ISSUE #16789: Automatically collapse long logs in log view
### Tell us your idea!

Hi, this is my first issue report! I searched for similar requests but didn’t find any, so I hope this is not a duplicate.

Currently, when viewing logs in c:geo, very long logs take up a lot of space, making it harder to quickly browse through them. Many logs contain lengthy, often redundant messages, such as generic "Thank you" notes in multiple languages. This clutters the log view, making it difficult to find relevant information.

- Logs longer than X characters (e.g., 300-500?) are initially collapsed.
- A “Show more” button appears to expand the full log.
- Option in settings to enable/disable log collapsing.
- Optional suggestion: Allow users to define the threshold (number of characters) at which logs are collapsed. This would let players customize the feature based on their preferences and device screen size.

### Is there an alternative?

_No response_

### Additional Information

_No response_

moving-bits:
## Description
Per default show log entries with a maximum of 5 lines. If a log entry is longer, a vertical ellipsis symbol is being displayed. Long-tapping on the log entry will expand the view.

|collapsed state|expanded state|
|---|---|
|image|image|

Edit: If a log item is collapsible, a menu item will be added to the log entry's context menu to either collapse or expand the log item:

|non collapsible|collapsed|expanded|
|---|---|---|
|image|image|image|

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogsViewCreator.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogsViewCreator.java
@@ -172,6 +172,11 @@ public abstract class LogsViewCreator extends TabbedViewPagerFragment<LogsPageBi
 
             }
 
+            // expand/collapse
+            if (holder.binding.log.isCollapsible()) {
+                ctxMenu.addItem(holder.binding.log.isCollapsed() ? R.string.menu_expand_log : R.string.menu_collapse_log, holder.binding.log.isCollapsed() ? R.drawable.expand_less : R.drawable.expand_more, it -> holder.binding.log.setCollapse(!holder.binding.log.isCollapsed()));
+            }
+
             //Copy to clipboard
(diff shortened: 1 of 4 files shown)


ztNFny:
Imho accessing the full log is too hard. Should be short tap on the ellipsis. And probably no "collapse" is required as reopening the cache will collapse all of them again.

ztNFny:
and 5 lines is really *very* little. If it's a hardcoded value it should be around 25 imho, a default for a setting could be 15-20. I'm mostly annoyed about logs with >100 lines.

moving-bits:
I've changed "long tap" to "short tap", so a short tap in collapsed state will expand the log, all subsequent short taps will open the context menu. First short tap on a log that did not get collapsed at all will open the context menu immediately.

With regard to log length: "25" looks a bit lenghty to me, I've set the limit to 15 for now. Let's see how it works in real life.
I did not implement a separate setting for this - not necessary, IMHO.

Issue #16728, opened 2025-03-09
Offline logging crashes


moving-bits:
User report on support (ticket 479688) for current nightly 2025.03.09-NB:

c:geo crashes on offline logging.

In the logfile provided I can see the error for #16727 + this one (not sure if it's related):

```
03-09 15:41:17.412 23499 24199 W cgeo    : [RxComputationThreadPool-8] HtmlImage.loadImageFromStorage
03-09 15:41:17.412 23499 24199 W cgeo    : java.lang.RuntimeException: java.lang.InstantiationException: java.lang.Class<com.drew.metadata.exif.ExifIFD0Directory> cannot be instantiated
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.metadata.tiff.DirectoryTiffHandler.pushDirectory(SourceFile:0)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.metadata.exif.ExifTiffHandler.setTiffMarker(Unknown Source:44)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.tiff.TiffReader.processTiff(Unknown Source:25)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.metadata.exif.ExifReader.extract(SourceFile:0)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.metadata.exif.ExifReader.extract(SourceFile:0)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.metadata.exif.ExifReader.readJpegSegments(Unknown Source:28)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.jpeg.JpegMetadataReader.processJpegSegmentData(Unknown Source:40)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.jpeg.JpegMetadataReader.process(Unknown Source:58)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.jpeg.JpegMetadataReader.readMetadata(SourceFile:0)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.jpeg.JpegMetadataReader.readMetadata(SourceFile:0)
03-09 15:41:17.412 23499 24199 W cgeo    : 	at com.drew.imaging.ImageMetadataReader.readMetadata(SourceFile:0)
[...]

Issue #16727, opened 2025-03-09
Crash in settings


moving-bits:
User report (ticket 479688) on support:

Current nightly (2025.03.09-NB) crashes for logging settings:

```
03-09 15:41:45.975 23499 23499 E AndroidRuntime: FATAL EXCEPTION: main
03-09 15:41:45.975 23499 23499 E AndroidRuntime: Process: cgeo.geocaching, PID: 23499
03-09 15:41:45.975 23499 23499 E AndroidRuntime: java.lang.IllegalArgumentException: Internal error: TypeReference constructed without actual type information
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at com.fasterxml.jackson.core.type.TypeReference.<init>(Unknown Source:31)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at cgeo.geocaching.settings.Settings$1.<init>(Unknown Source:0)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at cgeo.geocaching.settings.Settings.getLogTemplates(Unknown Source:13)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at cgeo.geocaching.log.LoggingUI.showOfflineMenu(Unknown Source:105)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at cgeo.geocaching.log.LoggingUI.onMenuItemSelected(Unknown Source:18)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at cgeo.geocaching.CacheDetailActivity.onOptionsItemSelected(Unknown Source:415)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at android.app.Activity.onMenuItemSelected(Activity.java:4719)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at androidx.activity.ComponentActivity.onMenuItemSelected(Unknown Source:0)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at androidx.fragment.app.FragmentActivity.onMenuItemSelected(Unknown Source:0)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected(Unknown Source:0)
03-09 15:41:45.975 23499 23499 E AndroidRuntime: 	at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(Unknown Source:2)
[...]

PR #16819, opened 2025-04-06
Update to AGP 8.2 + Gradle 8.2

bekuno:
New attempt to update the gradle infrastructure.

- AGP 8.2.2
- Gradle 8.2.1 
  Version 8.2 is the minimum version of Gradle compatible with AGP 8.2.2.
- Update default R8 processing mode 
  see https://r8.googlesource.com/r8/+/refs/heads/master/compatibility-faq.md#r8-full-mode => no compatibility mode necessary
- Enable buildConfig build feature 
  see https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#default-changes
- `Refactor` > `Migrate BuildConfig to Gradle Build Files`
- Preserve transitive R classes 
  see https://developer.android.com/build/optimize-your-build#use-non-transitive-r-classes
- `Refactor` > `Migrate to Non-Transitive R Classes`
- Preserve constant R class values 
  see https://developer.android.com/build/optimize-your-build#use-non-constant-r-classes
- `Refactor` > `Migrate ResValues to Gradle Build Files`
- Enable aidl explicitly where needed 
  see https://developer.android.com/develop/background-work/services/aidl
- Migrate testCoverageEnabled to enableUnitTestCoverage and enableAndroidTestCoverage
  It will be removed in AGP 9.0.0.
- Downgrade Jackson to 2.12.x 
  Compatibility of version 2.13 is only with MinSdk 24

I tested the scenarios from #16727 and #16728 on VM with API 21 and 34 and had no exceptions.


bekuno:
@moving-bits 
There are now some lint error, see https://ci.cgeo.org/job/cgeo-CI_PR-build/11543/consoleFull.

~~~
16:18:25  > Task :main:lintBasicDebug FAILED
16:18:25  Lint found 26 errors, 81 warnings. First failure:
16:18:25  
16:18:25  src/main/java/cgeo/geocaching/downloader/PendingDownloadsActivity.java:198: Error: Unexpected implicit cast to MaterialButton: layout tag was Button [WrongViewCast]
16:18:25              buttonResume = itemView.findViewById(R.id.button_left);
~~~

The description in AS is "Unexpected implicit cast to MaterialButton: layout tag was Button"

Any idea?
Maybe a result of "Refactor > Migrate to Non-Transitive R Classes"?

moving-bits:
I'll look into it when I'm back on my computer

murggel:
@bekuno do you want me to check something special? Or was the review-request by accident?

bekuno:
> @bekuno do you want me to check something special? Or was the review-request by accident?

@murggel Yes, please test - see [#16728](https://github.com/cgeo/cgeo/issues/16728#issuecomment-2709148952) .

moving-bits:
> 16:18:25  src/main/java/cgeo/geocaching/downloader/PendingDownloadsActivity.java:198: Error: Unexpected implicit cast to MaterialButton: layout tag was Button [WrongViewCast]

fixed with #16821

bekuno:
@moving-bits 
I greped the lint error file from CI: [lint_errors](https://github.com/user-attachments/files/19623343/lint_errors_16819.txt)
Please have a look at the other 24 errors.

murggel (commented):
I tested the scenarios from #16727 and #16728 on VM with API 33 and 34 and had also no exceptions

moving-bits:
> @moving-bits I greped the lint error file from CI: [lint_errors](https://github.com/user-attachments/files/19623343/lint_errors_16819.txt) Please have a look at the other 24 errors.

PR #16827 fixes a couple of the lint warnings, but I have no solution for the reported error (it's 24 instances of the same error): 

Problem seems to be, that we have 24 generic `startActivity()` calls with action = `VIEW`, that technically match our (non-exported) `SettingsActivity`. The error goes away if I remove action = `VIEW` from `SettingsActivity` (which would render internal direct linking to settings non-functional) or if I set `exported="true"` for `SettingsActivity` (with possible side-effects yet unknown to me).

I have not found a way to declare those 24 calls to never use c:geo itself as target (at least not in a way that let lint recognize that the reason for its complaint no longer exists), so only mitigation I see currently is to ignore this lint error for the time being. (BTW, it seems to be a new one - haven't seen this lint message before.)

bekuno:
This lint check is new in the higher AGP (or Gradle ?) version.
Now I started in AS the local lint (.\gradlew lint) and got the same errors.

Explanation for the `UnsafeImplicitIntentLaunch` lint error:
This intent matches a non-exported component within the same app. In many cases, the app developer could instead use an explicit Intent to send messages to their internal components, ensuring that the messages are safely delivered without exposure to malicious apps on the device. Using such implicit intents will result in a crash in an upcoming version of Android.

I found also this page:
https://developer.android.com/privacy-and-security/risks/implicit-intent-hijacking

bekuno:
retest this please

bekuno:
@moving-bits 
If you don't have a solution yet, thenm I would ignore the errors for now and create a separate issue.

PR #16828, opened 2025-04-07
Improve cache list menu (rel #16195)

murggel:
## Description

New entry "Remove from other lists" in "Manage caches"-menu  works for all caches in list and for selection (menu"Manage list" is now disabled for select-mode)
Existing entry  "Show Attribute Overview" in "Manage caches"-menu works now on selected caches

## Related issues

rel #16195 

## Additional context

Some more GUI-improvements

Avoid unnecessary calls:
* Get upfront necessary options
* Get list of selected or all caches only once
* Count checked caches only if select-mode is active
* Handover checkedCount

Improve Labels:
* No count for delete past events
* No count for clear offline logs
* No count for upload modified coordinates
* Add count for export-entries in select-mode

Improve Visibility:
* Disable "Manage list"-menu  in select-mode
* Disable "Import"-menu in select-mode
* Hide "Clear offline log", if no caches with offline logs are in the list / selected


murggel:
Ready for review

PR #16830, opened 2025-04-08
config_gpx_export_details

baiti:
I have made an attempt to make the data elements that get included in a gpx export configurable. By default, the logs get included and also travel bugs. I wouldn't know what tool importing the GPX would be interested in the logs, nor the travel bugs. And if such a tool exists, well, then configure the new options accordingly.

Now there is one problem: I am on the road again with my very old and slow laptop, I have a hard time even compiling the code and running the emulator takes forever. Having said this: It seems my if-conditions don't evaluate correctly as the logs and travel bugs are always included even if I uncheck the checkboxes in the UI.

@moving-bits : Could you have a quick look at it?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/export/GpxExport.java
+++ b/main/src/main/java/cgeo/geocaching/export/GpxExport.java
@@ -84,6 +84,12 @@ public class GpxExport extends AbstractExport {
         final CheckBox includeFoundStatus = layout.findViewById(R.id.include_found_status);
         includeFoundStatus.setChecked(Settings.getIncludeFoundStatus());
 
+        final CheckBox includeLogs = layout.findViewById(R.id.include_logs);
+        includeLogs.setChecked(Settings.getIncludeLogs());
+
+        final CheckBox includeTravelBugs = layout.findViewById(R.id.include_travelbugs);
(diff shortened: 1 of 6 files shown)


moving-bits (approved):
From having a quick look at the code (not having tested it): LGTM

moving-bits:
> I wouldn't know what tool importing the GPX would be interested in the logs, nor the travel bugs.

Maybe not some other tool, but if a user wants to transfer cache data from one device to another, they can also use GPX export & import (including logs etc.)

baiti:
> From having a quick look at the code (not having tested it): LGTM

Pls do a quick test. I believe it is not working and includes the logs even if the checkbox is unchecked. I have not been able to figure out where the bug is but there must be something missing in the chain of events.

moving-bits:
Found the culprit:
While exporting you check for state of the flags in settings, but you don't update settings with the values of the dialog.

PR #16844 fixes this.

baiti:
> Found the culprit: While exporting you check for state of the flags in settings, but you don't update settings with the values of the dialog.
> 
> PR #16844 fixes this.

Ahh I see. How could I have missed that. Thx much for fixing it.

Issue #16852, opened 2025-04-15
Nightly: Crash on logging / editing offline log (Jackson/IllegalArgumentException)


moving-bits:
We got three similar reports on support pointing to a problem related to either logging or editing an existing offline log.

Whether this error is related to new new AGP/Gradle combination or to the Jackson downgrade I cannot say yet.

Error message is like this:
```
04-15 11:49:24.657 16085 16085 E AndroidRuntime: FATAL EXCEPTION: main
04-15 11:49:24.657 16085 16085 E AndroidRuntime: Process: cgeo.geocaching, PID: 16085
04-15 11:49:24.657 16085 16085 E AndroidRuntime: java.lang.IllegalArgumentException: Internal error: TypeReference constructed without actual type information
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at com.fasterxml.jackson.core.type.TypeReference.<init>(Unknown Source:31)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at cgeo.geocaching.settings.Settings$1.<init>(Unknown Source:0)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at cgeo.geocaching.settings.Settings.getLogTemplates(Unknown Source:12)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at cgeo.geocaching.log.LogTemplateProvider.getTemplatesWithSignature(Unknown Source:17)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at cgeo.geocaching.log.AbstractLoggingActivity.onCreateOptionsMenu(Unknown Source:19)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at cgeo.geocaching.log.LogCacheActivity.onCreateOptionsMenu(Unknown Source:0)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at android.app.Activity.onCreatePanelMenu(Activity.java:4787)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at androidx.activity.ComponentActivity.onCreatePanelMenu(Unknown Source:2)
04-15 11:49:24.657 16085 16085 E AndroidRuntime: 	at androidx.fragment.app.FragmentActivity.onCreatePanelMenu(Unknown Source:2)
[...]

PR #16855, opened 2025-04-16
add jackson proguard rules

bekuno:
Lets test if the rules help against the error #16852.


bekuno:
I have no errors on a physical device with Android 9 with the PR result apk.
Let's test in the nightlies.

PR #16861, opened 2025-04-18
rel to #16779: wherigo zone: add compass and copy coordinate

eddiemuc:
rel to #16779: wherigo zone: add compass and copy coordinate

Adds to Wherigo Player:
* ability to navigate to a zone using compass
* ability to copy zone center coordinates to clipboard
* long-tap on a zone on map is no longer recognized. This allows users to do other stuff in map zone area available on long-tap eg create user-defined cache
* set zone center as target when opening map (to get routing and distance info for it)


eddiemuc:
Retest this please

PR #16771, opened 2025-03-18
offline hillshading with VTM

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/16760, one of the last steps in feature parity.
also removes the online hillshading soruces

should probably use a release version of VTM but none is available so far, so do we wait? not sure how it was done in the past.

Issue #16703, opened 2025-03-01
[Nightly] Map “Combined offline” no longer shows many forests


MagpieFourtyTwo:
### Describe your problem!

If Live Map is set to "Combined offline" most of the woods in our region are not shown anymore.
If I switch to "Baden-Wuerttemberg_oam.osm.map", the woods are there again:

Map Combined:
![Image](https://github.com/user-attachments/assets/010a474e-93b4-4a2c-a5ef-71660ace8696) ![Image](https://github.com/user-attachments/assets/b3aa68ed-81b9-4608-b8e1-dbc17bb1fac9)

Map BaWue:
![Image](https://github.com/user-attachments/assets/54a9536f-2f38-44d1-8a67-b37f1635f1cd) ![Image](https://github.com/user-attachments/assets/6274418a-e4ee-4c27-af11-d5a2a059951d)

### How to reproduce?

1. open live map
2. go to any area with woods, covered by an OSM map
3. set map to a "dedicated" OSM offline map
4. switch map to "Combined offline"

### Actual result after these steps?

Some woods, which where there with the "dedicated" map, will disappear.

### Expected result after these steps?

Woods should be shown no matter if "Combined offline" or the dedicated map is displayed.

### Reproducible

Yes

### c:geo Version

2025.03.01-NB-3f16d2c

### System information

```text

```

### Additional Information

There are no overlapping maps for the region BaWue, esp. Germany South is not installed.

MagpieFourtyTwo:
Just seen that VTM has been updated - I'm currently using UM variant VTM.
If I switch UM variant to "Mapsforge", everything is fine.
Plus I get hillshading, as I just realized, so I will most probably stay with Mapsforge ...  ;)
Or are there reasons to stay with VTM ...?

eddiemuc:
It is my current belief that VTM will no longer be pursued by us

MagpieFourtyTwo:
Appears to be a good idea - while playing around with it this evening I additionally got completely white squares plus these very well known red triangles.  :/

devemux86:
> Plus I get hillshading

Hill shading will be available in VTM soon, I'm currently finalizing it:

- https://github.com/mapsforge/vtm/issues/1180

ztNFny:
> Appears to be a good idea - while playing around with it this evening I additionally got completely white squares plus these very well known red triangles.  :/

That's fixed in VTM master

ztNFny:
[...]

PR #16864, opened 2025-04-19
Offline hillshading for VTM

moving-bits:
## Description
- Updates VTM to v0.25, jts-core to 1.20 (as VTM uses it itself)
- Updates Mapsforge to v0.25
- Integrates the hillshading changes from #16771


moving-bits:
Everything seems to be fine - except that it does not show any hillshading at all when selecting VTM. Hillshading on Mapsforge does work. I've cross-checked the VTM hillshading example - seems to be pretty much what we have here.

I have no idea currently why it's not working, therefore setting WIP for now.

moving-bits:
Ok, strange observation: If I move the map around a bit, hillshading appears eventually, but leaving out some squares sometimes:

image.image.image

(This looks somehow similar to what happens with the forests in #16703, although I don't see any "flickering of squares" here)

@devemux86: Any idea what might be causing this?

devemux86:
@moving-bits 

Is it one or multiple maps? It's best to check and test the VTM [HillshadingActivity](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/HillshadingActivity.java) example.
Changing map layers requires to reload the map, e.g. call `map.clearMap()`.

VTM uses a separate hillshading layer with its own cache.
This is now also available in Mapsforge with separate cache (see https://github.com/mapsforge/mapsforge/pull/1672).

About the forests, I asked which map files to test, but there was no answer.

devemux86:
@moving-bits 

Is it one or multiple maps? It's best to check and test the VTM [HillshadingActivity](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/HillshadingActivity.java) example.
Changing map layers requires to reload the map, e.g. call `map.clearMap()`.

VTM uses a separate hillshading layer with its own cache.
This is now also available in Mapsforge with separate cache, see [HillshadingMapViewer](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/HillshadingMapViewer.java) example.

About the forests, I asked which map files to test, but there was no answer.

moving-bits:
> Is it one or multiple maps?

In my tests it has always been a single map only (here: Andorra or Germany maps from OpenAndroMaps with Elevate theme).

> It's best to check and test the VTM [HillshadingActivity](https://github.com/mapsforge/vtm/blob/master/vtm-android-example/src/org/oscim/android/test/HillshadingActivity.java) example.

Will do, but in the end it will have to work in our complex environment, and I need to figure out which part in our environment may be leading to this behavior.

> Changing map layers requires to reload the map, e.g. call `map.clearMap()`.

We create the fragment, add basemap, labels and building layers, then scalebar, then hillshading.

I have tried with adding a `clearMap()` after adding the hillshading layer, but that did not help. The part of the map that is visible initially, does not get hillshading applied to in that specific zoom level. 
- If I pan the map, hillshading will be displayed, but only on the outsides of the initial viewport.
- If I zoom out, hillshading will be displayed full-screen.
- If I zoom in, no hillshading will be displayed (only when additionally panning to a place which is outside the initial viewport).

Do I have to clear the hillshading cache additionally, or something like this?

> About the forests, I asked which map files to test, but there was no answer.

I'm having the "forests" problem as well. For me it occurs with both Germany map and Netherlands map, both from OpenAndroMaps with Elevate theme.

devemux86:
@moving-bits 

- About the forests, I answered in its issue:
https://github.com/cgeo/cgeo/issues/16703#issuecomment-2817148222

- About the hillshading, do you see this even with the small Andorra map?
If VTM samples and Cruiser work, then it could be something in the implementation.

Generally better to avoid fragments, especially with OpenGL maps, there have been problems in the past.
Using activities like in [vtm-android-example](https://github.com/mapsforge/vtm/tree/master/vtm-android-example) samples always seems to have the best results.

Issue #4222, opened 2014-08-26
Support for *.sqlitedb offline maps


heldersf:
Hi there, thanks for the enourmous support and improvements on this app.

Is it possible to add the habillity to support *.sqlitedb maps as otther apps do?
I am also using neongeo and cachesense apps  because they have the possibility to work with *.sqlitedb maps and also *.map .
c:geo supports mapsforge maps, which have the *.map format. those apps besides that also support *.sqlitedb  (created by rmaps app, by locus and by Mobile_Atlas_Creator_for_pc among others). the advantage is that those maps can be created with sattelite imagging. 
I know that c:geo also has the hability to download sattelite maps with "static maps" feature, but maybe is desirable to have the whole area in a unique map.

regards,

moving-bits:
Quite some time since this issue has been opened - not sure if this issue is still relevant. But just to let you know: c:geo does support loading maps in `.mbtiles` format meanwhile, to be used as background maps (supported since current nightly version, will be part of the next feature release).

moving-bits:
c:geo can only support what the libraries used for displaying the maps support, in our case Google Maps, Mapsforge and VTM. If `.sqlitedb` means MBTiles format, we do support it meanwhile, but if it means something different, we won't be able to support it. Therefore closing here.

PR #16869, opened 2025-04-20
UnifiedMap VTM: Add background map capability

moving-bits:
## Description
Support background maps in `.mbtiles` format for VTM (display, download, update, settings).

|Default "world.map"|background map|
|---|---|
|image|image|

Using regular "world.map" is fast, but has nearly no detail info. Using country-specific maps instead requires much device space + is very slow in low zoom levels. An alternative is using bitmap based overview maps like those provided by [OpenAndroMaps](https://www.openandromaps.org/downloads/ubersichts-karten). Those maps are provided in `.mbtiles` format, which is supported by VTM.

|enable/disable|start downloader|downloader|
|---|---|---|
|image|image|image|

Users can decide which zoom levels to support with background maps by selecting the appropriate download. (Supporting more zoom levels cost more device space.)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/downloader/DownloadSelectorActivity.java
+++ b/main/src/main/java/cgeo/geocaching/downloader/DownloadSelectorActivity.java
@@ -16,6 +16,7 @@ import cgeo.geocaching.utils.ShareUtils;
 import android.annotation.SuppressLint;
 import android.app.DownloadManager;
 import android.content.Context;
+import android.content.Intent;
 import android.database.Cursor;
 import android.database.CursorIndexOutOfBoundsException;
 import android.os.Bundle;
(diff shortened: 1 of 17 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

moving-bits:
failing test (testGetOCCache) is unrelated and runs fine locally

SammysHP:
Does this also fix #4222?

moving-bits:
> Does this also fix #4222?

Might be, if the TO actually wants `.mbtiles` maps. There seem to be some differences between sqliteDB and MBTiles map sources, at least OpenAndroMaps generates separate files for them: https://www.openandromaps.org/downloads/ubersichts-karten

Additionally: This PR implements MBTiles support for a specific use case - displaying them as background maps. There is no individual selection currently, all found maps in MBTiles format will be displayed together, if that option is enabled. Wouldn't be too hard to implement, probably, but some more UI would be needed.

ztNFny:
The background map is rendered even for regions where the map is available on low zoom levels. Imho "real" map should win

moving-bits:
> The background map is rendered even for regions where the map is available on low zoom levels. Imho "real" map should win

For UnifiedMap Mapsforge this is what it did for me (tested with the OAM Berlin.map).

Not sure how to influence it, though, as it's just another layer. Do you know if VTM and Mapsforge maps have a way of setting a priority for such "base map layers"?

Issue #16867, opened 2025-04-20
Nightly: Crash on importing pocket queries


moving-bits:
Reported by user on support (ticket 987541):

---

Trying to import pocket queries. PQ list shows up with correct dates. Cgeo crashes violently upon importing. 

1: create (or select) a "list" in cgeo
2:floppy disc - "reading file" - kaboom (picture) 

But

 3 dots - import - pocket queries - pq's are loaded - select the Magnifying glass - PQ IS loaded and all caches in the PQ _ARE_ shown on screen. I can load the caches, but not from the PQ; it's the same download as selecting "refresh" from the long press on a cache name. (Pic 2 shows PQ name and number of caches in that PQ and the first few caches in that pq) 

Selecting the magnifying glass - 3 dots - manage caches - move all actually allows me to store the caches on the pq in a list in Cgeo.

---

Provided logfile has many entries like these:

```
4-19 22:19:45.262 13671 13930 E AndroidRuntime: FATAL EXCEPTION: Thread-3
04-19 22:19:45.262 13671 13930 E AndroidRuntime: Process: cgeo.geocaching, PID: 13671
04-19 22:19:45.262 13671 13930 E AndroidRuntime: java.lang.ExceptionInInitializerError
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at org.apache.commons.compress.archivers.zip.ZipArchiveEntry.setExtra(SourceFile:0)
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at org.apache.commons.compress.archivers.zip.ZipArchiveInputStream.getNextZipEntry(Unknown Source:318)
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at cgeo.geocaching.files.AbstractImportGpxZipThread.parseZipFileGpxContent(Unknown Source:16)
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at cgeo.geocaching.files.AbstractImportGpxZipThread.doImport(Unknown Source:8)
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at cgeo.geocaching.files.AbstractImportGpxThread.doImport(SourceFile:0)
04-19 22:19:45.262 13671 13930 E AndroidRuntime: 	at cgeo.geocaching.files.AbstractImportThread.run(Unknown Source:15)
[...]

PR #16871, opened 2025-04-21
Disable R8 full mode optimizations (test)

moving-bits:
## Description
Test PR to check whether this avoids effects like #16727 or #16867

If this works, we can use this as a fall back and quick fix, and try to investigate further into detailed ProGuard rules.


moving-bits:
FYI: superseded by #16872, which seems to be working as well

PR #16877, opened 2025-04-21
fix #16875: wherigo, allow target selection for thing action

### LINKED ISSUE #16875: Wherigo: feature missing to use inventory items with usage targets
### Tell us your idea!

Current Wherigo Player implementation misses the ability to use inventory items with usage targets. This aspect has to be added.

Equivalent in WhereYouGo is coded in classes ListActionsActivity and ListTargetsActivity. 
Starting point is ListActionsActivity.callAction when z.hasParameter() is true and z.getActor() == thing:

Image

An example where this leads to problems is GC9TB11 (tutorial) where you can't use the "Taler" to give to the "Waffenhändler" (when you klick on "benutzen mit..." then a list of possible targets should appear which does not currently

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

eddiemuc:
fix #16875: wherigo, allow target selection for thing action

### DIFF
--- a/main/src/main/java/cgeo/geocaching/wherigo/WherigoDialogManager.java
+++ b/main/src/main/java/cgeo/geocaching/wherigo/WherigoDialogManager.java
@@ -209,6 +209,9 @@ public class WherigoDialogManager {
     }
 
     private void clearInternal() {
+        if (currentDialogControl != null) {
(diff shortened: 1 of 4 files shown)


eddiemuc:
Jobs seem to be failing in setup phase, so I assume fails are unrelated to PR

Issue #16866, opened 2025-04-19
Nightly: Wherigo Player Error when select Play Cartridge


rouroni1:
### Describe your problem!

Open Wherigo Player, select a Cartridge, press Play.

Cartridges wont start. 

### How to reproduce?

Open Wherigo player.
Open a Wherigo Cartridge (tryed  all my cartridges, and wadmeisters reverse)
Select play

### Actual result after these steps?

Cartridges wont start and there's a new message on Icon Bug

### Expected result after these steps?

Cartridge start

### Reproducible

Yes

### c:geo Version

2025.04.20-NB-9364ddc

### System information

```text
## System information

c:geo version: 2025.04.20-NB-9364ddc

Datetime: 20 de abril às 00:15

Device:
-------
- Device type: 21121210G (ingres_eea, POCO)
- Available processors: 8
- Android version: 14
- Android build: UKQ1.230917.001 release-keys
- Screen resolution: 1080x2186px (393x795dp)
- Pixel density: 2.75
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 25.13.33 (190400-745790146)
- Memory:  Available:3,3 GB, Total:10,9 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v9, Count:220
- Set language: pt_PT (system default)
- System date format: dd/MM/yy
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: 3/10/2024, 23:05
- Last auto backup: 16/04, 23:20
- Routing mode: Walk (shortest.brf / trekking.brf / car-eco.brf / (none) / (none))
- Map mode: UnifiedMap / live / OSM multi-threading: off
- Map: Spain-Portugal (OAM)
  - Id: cgeo.geocaching.unifiedmap.tileproviders.AbstractMapsforgeOfflineTileProvider:primary:CGEO/maps/Spain-Portugal (OAM).map
  - Attrs: 
  - Theme: Elevate.zip:Elevate.xml

Filters:
-------
[...]

PR #16887, opened 2025-04-24
Integrate OpenWIG codebase into c:geo

eddiemuc:
Integrate OpenWIG codebase into c:geo


eddiemuc:
Retest this please

eddiemuc:
Retest this please

SammysHP:
What's the reason for this, was there a discussion about it? How does it affect other apps using the library or was it always a custom fork intended just for c:geo? I don't think there are many users of the library, but we shouldn't break it intentionally.

eddiemuc:
> What's the reason for this, was there a discussion about it? How does it affect other apps using the library or was it always a custom fork intended just for c:geo? I don't think there are many users of the library, but we shouldn't break it intentionally.

See https://github.com/cgeo/openWIG/pull/21#issuecomment-2821765908

Intention of course is to extend openWIG code for c:geo (especially for debugging) w/o having to worry that this breaks other usages of the library. Especially the later is challenging and hinders further enhancements for openWIG usage in c:geo.

Copying code to c:geo doesn't mean that the original openWIG repo has to be deleted of course. It can stay, be used by other projects and be maintained by the same group of devs who is maintaining it today.

That's my opinion only. Can be discussed of course.

eddiemuc:
The alternative to integration of code would be to make OoenWIG Library maintainable again to handle bugs in c:geo (see eg https://github.com/cgeo/cgeo/issues/16866#issuecomment-2818751179). But thats a major piece of work on its own:
- build structure needs to be cre6ans maintenaned (GitHub actions or Jenkins)
- release publish structure is missing (relies currently on jitpack solely; no publishing to eg maven rep or Google)
- currently NO TEST AUTOMATION AT ALL which is a no-go for further maintenance (for my last attempt to do something there I had to build releases after each change and test with c:geo manually! This is not working for me)
- test infrastructure must be set up such that it will detect problems for ALL users of the library (not just c:geo)
- regular maintenance of used library versions__
- regular maintenance of build system eg Gradle and Java version compatibity

We are talking serious work here. Who would be willing to go into the maintenance roles for openWIG as a standalone library?

eddiemuc:
I am leaving this PR open on purpose until Sunday in case there's a need to discuss this further.

bekuno:
The openWIG repository was cloned into the cgeo organization in conjunction with the simultaneous cloning of the WhereYouGo repository. Nobody but me and Lineflyer had pushed code there. We had a few contributions, all related to WhereYouGo.
Maybe someone is using the code, but we had no such feedback.
After integrating the WhereYouGo functionality into c:geo, it is a ‘natural’ step to also integrate the basic parts from the openWIG repository. By the way, the LUA part in openWIG was already a copy from the unmaintained original [kahlua ](https://github.com/krka/kahlua2) repository.

bekuno:
There was development on a cloned repository for kahlua: https://github.com/yanmingsohu/kahlua2/tree/master/core/src/se/krka/kahlua/
Maybe worth to have a look?
Additional idea: Instead of integration to kahlua code in the cgeo namespace we could integrate this as subproject, like [organicmaps-api](https://github.com/cgeo/cgeo/tree/master/organicmaps-api). Then we have not to deal with different namespaces comparing both trees. And maybe it can be easier to change to another LUA implementation.

eddiemuc:
> There was development on a cloned repository for kahlua: https://github.com/yanmingsohu/kahlua2/tree/master/core/src/se/krka/kahlua/
> Maybe worth to have a look?
> Additional idea: Instead of integration to kahlua code in the cgeo namespace we could integrate this as subproject, like [organicmaps-api](https://github.com/cgeo/cgeo/tree/master/organicmaps-api). Then we have not to deal with different namespaces comparing both trees. And maybe it can be easier to change to another LUA implementation. 

The idea behind the code copy is to start modifying it for c:geo purposes. Don't see why keeping original namespaces will help here. We didn't do it for BRouter.

Meanwhile we have bug #16866 out there in the field. My concern is solely trying to solve it quickly. If not by modifying an openWIG copy, what is the best way to achieve this?

bekuno:
> The idea behind the code copy is to start modifying it for c:geo purposes. Don't see why keeping original namespaces will help here. We didn't do it for BRouter.
 
The idea for a sub project within the c:geo codebase is to keep the original namespace. 
So it would easier to compare changes between different implementations/changes directly. 
Btw., the same question I ask me when integration brouter.

The openWIG library has the 2 parts - openwig and kahlua. The first part is directly for WIG and should be directly integrated in c:geo.
But the LUA runtime was in development in the referenced repository. Probably there are already bugs fixed.

> Meanwhile we have bug #16866 out there in the field. My concern is solely trying to solve it quickly. If not by modifying an openWIG copy, what is the best way to achieve this?

I argument completely for the integration of the openWIG library in c:geo to easier find and fix errors.
My question is only, would it make sense to use a subproject configuration for kahlua.

bekuno:
> My question is only, would it make sense to use a subproject configuration for kahlua.

Or even easier, no subproject but integrate besides the cgeo.geocaching code tree with its original name.
(This can also be relocated after merging this PR, if the team wants to go this way.)

SammysHP:
Why can't we keep it in a separate repository in our organization namespace? Personally I don't like multiple (more or less independent) projects in the same repository.¹ We could either update our fork of OpenWIG or even create another fork of it.

¹ I also often use submodules, but understand the difficulties they introduce.

eddiemuc:
> Why can't we keep it in a separate repository in our organization namespace? Personally I don't like multiple (more or less independent) projects in the same repository.¹ We could either update our fork of OpenWIG or even create another fork of it.

Our OpenWIG Repo is the ",original" one these days. It is not a fork. It also claims that it is solely for purpose of Where you go development. See https://github.com/cgeo/openWIG.

Now that WhereYouGo is more or less gone (?) I feels that openWIG becomes part of c:geo too.

All this doesn't mean one can't keep it as separate library or submodule. There just needs to be someone to set this up and maintain it. Who would that be? There was no one doing this for many many years. That's a strong argument against a separate repository.

eddiemuc:
> The openWIG library has the 2 parts - openwig and kahlua. The first part is directly for WIG and should be directly integrated in c:geo.
> But the LUA runtime was in development in the referenced repository. Probably there are already bugs fixed.

I googled it.
The original repo of Kahlua is unmaintained: https://code.google.com/archive/p/kahlua/. So there won't be updates anymore.

Another implementation kahlua2 is referenced. This project was also not updated an more for 12 years; the parts we are using were not updated for around 15 years. For all practical purposes both the original as well as the follow-up project are unmaintained. Reference: https://github.com/krka/kahlua2/tree/master/core/src/se/krka/kahlua/vm. Switching to kahlua2 would be a piece of work on its own which can be done separately from the issue at hand. (And currently I don't see the need tbh) Keeping package structures from Kahlua would not help form migration to kahlua2 though.

> > Meanwhile we have bug #16866 out there in the field. My concern is solely trying to solve it quickly.
> 
> I argument completely for the integration of the openWIG library in c:geo to easier find and fix errors.
> My question is only, would it make sense to use a subproject configuration for kahlua.

It is the discussion itself that takes the time and hinders us from moving forward. As long as there's discussion, no work gets done :-)

eddiemuc:
Update: In looked a little closer into kahlua2. Although package names suggest similarities to Kahlua, code is not compatible and openWIG would not work or compile against it. Examples from https://github.com/krka/kahlua2/tree/master/core%2Fsrc%2Fse%2Fkrka%2Fkahlua%2Fvm:
- interface `LuaTable` is now called `KahluaTable` 
- `KahluaTable` expects methods to be implemented which are not present in OpenWIG's `LuaTableImpl` class. Among other things, a `KahluaTableIterator` needs to be implemented 

I assume there are more incompatibility examples.

bekuno:
> The original repo of Kahlua is unmaintained: https://code.google.com/archive/p/kahlua/. So there won't be updates anymore.

The google code was completely closed some years ago.
All projects were transfered to github and are accessible in the google code archive. 

> Another implementation kahlua2 is referenced. This project was also not updated an more for 12 years; the parts we are using were not updated for around 15 years. For all practical purposes both the original as well as the follow-up project are unmaintained. Reference: https://github.com/krka/kahlua2/tree/master/core/src/se/krka/kahlua/vm. 

This is the migrated project from google code.

There was development on a cloned github repository: https://github.com/yanmingsohu/kahlua2/tree/master/core/src/se/krka/kahlua/

I proposed only to check if the updates there are bugfixes which are usefull for c:geo.

> Meanwhile we have bug https://github.com/cgeo/cgeo/issues/16866 out there in the field. My concern is solely trying to solve it quickly.

Maybe fixed there?

eddiemuc:
> There was development on a cloned github repository: https://github.com/yanmingsohu/kahlua2/tree/master/core/src/se/krka/kahlua/
> 
> I proposed only to check if the updates there are bugfixes which are usefull for c:geo.
> 
> > Meanwhile we have bug https://github.com/cgeo/cgeo/issues/16866 out there in the field. My concern is solely trying to solve it quickly.
> 
> Maybe fixed there?

Quick check: https://github.com/yanmingsohu/kahlua2/tree/master/core/src/se/krka/kahlua/ Code state is also incompatible to openWIG code.

fm-sys:
I would personally vote to keep the separate repository, but treat it as if it's a private library just for c:geo purposes.
If someone else is using https://github.com/cgeo/openWIG they will need to either don't upgrade the library or have to deal with the fact that there are (quite some?) breaking changes.

eddiemuc:
> What can be done here that can't be done in https://github.com/cgeo/openWIG? 

Sure, that's the master repo for openWIG afaik. Will you fix the problem there?

moving-bits:
I don't understand why we should NOT integrate the code into c:geo. I mean, technically I do understand the wish to keep things separate, but I don't think it applies to the situation at hand in a way that a separate library is needed:

- We don't maintain the original code, just a c:geo-specific copy of otherwise unmaintained code
- this copy has four forks - two of them from ourself, two others from outside, which have not been updated since years => thus I don't think we need to take care of external users of "our" lib
- integrating the code would make things way easier to handle. Currently testing, fixing or enhancing things is quite tedious - you need to create a new version, make jitpack recognize it, then update c:geo to recognize & use this new version, just to test a single code change. That's a lot of "wasted" work on a library only used by us, IMHO.
- It would be less tedious, if the lib had some useful testing routines, as we could use an in-lib test cycle. This could be added, for sure, but who would be willing to do that? I guess we all rather want to spend our coding efforts on enhancing c:geo than on such a lib.

Thus I vote for integrating the openWIG code into c:geo codebase. 
Whether it's into the regular code tree (separate folder) or as submodule like we have for MWM and Organic Maps APIs, I have no strong preference.

SammysHP:
> - integrating the code would make things way easier to handle. Currently testing, fixing or enhancing things is quite tedious - you need to create a new version, make jitpack recognize it, then update c:geo to recognize & use this new version, just to test a single code change. That's a lot of "wasted" work on a library only used by us, IMHO.

That's definitely a benefit I can agree with!

I'm not in the position to tell you what to do, so please continue with the merge if it helps the development. It just feels like pulling in more and more code from other repositories might make it harder to maintain c:geo over time. But maybe it's the opposite and might speed up development if everything is in one place. :man_shrugging:

eddiemuc:
> I would personally vote to keep the separate repository, but treat it as if it's a private library just for c:geo purposes.
> If someone else is using https://github.com/cgeo/openWIG they will need to either don't upgrade the library or have to deal with the fact that there are (quite some?) breaking changes. 

This idea got a lot of votes so I want to understand it. IMHO it would give us all the downsides of a separate library (own test routines, own build to maintain, own release process to maintain so it can be used in c:geo) but none of the upsides (usable also outside of c:geo).

What is the big benefit of this approach that I fail to see?

eddiemuc:
> Thus I vote for integrating the openWIG code into c:geo codebase.

+1 :-)

> Whether it's into the regular code tree (separate folder) or as submodule like we have for MWM and Organic Maps APIs, I have no strong preference.

The reason for integration into c:geo package structure is that I (we?) plan to modify both the openWIG and the Kahlua code to help c:geo-specific debugging and fix c:geo specific errors. Keeping original packaging would simply be misleading because code will be different from original. My exam5for this approach would be Brouter where we did the same eg for SAF integration.

Why are we keeping copies of MWM and OrganicMap I don't know. If we don't plan to modify the code in a c:geo specific way then indeed we could leave it in a separate repo. Comes with the downside of no Bugfixes or maintenance for these parts of c:geo of course.

ztNFny:
I don't really have an opinion. It feels more correct to separate it as a lib, even if that lib is only used by cgeo. Keeping it separate has overhead though, build pipeline and stuff - not wanting to invest the effort to set that up and maintain it would be a valid excuse to go with the second best way.

PR #16892, opened 2025-04-26
Add "application language" to selection of translation language

murggel:
## Description

Add "application language" to selection of translation language
image

#### Disabled
image

#### Application language
image

#### Language
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1890,7 +1890,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             OfflineTranslateUtils.getTranslator(cda, cda.translationStatus, sourceLng,
                 unsupportedLng -> {
                     cda.translationStatus.abortTranslation();
-                    binding.descriptionTranslateNote.setText(getResources().getString(R.string.translator_language_unsupported, sourceLng));
+                    binding.descriptionTranslateNote.setText(getResources().getString(R.string.translator_language_unsupported, unsupportedLng));
                 }, modelDownloading -> binding.descriptionTranslateNote.setText(R.string.translator_model_download_notification),
     translator -> {
                     if (null == translator) {
--- a/main/src/main/java/cgeo/geocaching/TrackableActivity.java
+++ b/main/src/main/java/cgeo/geocaching/TrackableActivity.java
@@ -645,7 +645,7 @@ public class TrackableActivity extends TabbedViewPagerActivity {
             OfflineTranslateUtils.getTranslator(cda, cda.translationStatus, sourceLng,
                     unsupportedLng -> {
                         cda.translationStatus.abortTranslation();
-                        binding.descriptionTranslateNote.setText(getResources().getString(R.string.translator_language_unsupported, sourceLng));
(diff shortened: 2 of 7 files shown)


murggel:
Retest this please

murggel:
Retest this please

murggel:
Seems that there is a problem creating adb bridge?

murggel:
> What happens if the application language is not supported for translation? Translation will then fail (in what way?) and the user won't know why.
> 
> I don't understand why this value is needed, the user needs to select the target language once, whether they select app language or one specific one is not much difference. 

Using the application language is the default behavior for the current implementation, if the translation language is not selected explicitly. Then the dialog for downloading appears. So I just added this option to the selection...

Current application language (danish):
image

No translation language is selected (default on fresh installation):
image

Dialog to confirm downloading danish:
image

murggel:
> What happens if the application language is not supported for translation? Translation will then fail (in what way?) and the user won't know why.
> 
> I don't understand why this value is needed, the user needs to select the target language once, whether they select app language or one specific one is not much difference.

Ok, I will check, what happens, if application language is not in the list of the supported languages...

ztNFny:
Sorry, didn't realize that system language was already the default. In that case displaying it in UI makes total sense of course!

I know that i implemented the "translation available" check in a way to check that the selected language is valid, so that one would return false in case the app language is unsupported. So only the settings UI would be misleading as it the suggests "app language" even if that one is not available. 
How about showing it as "app language: <language>" and disable the menu item of the language is not supported?

murggel:
> Sorry, didn't realize that system language was already the default. In that case displaying it in UI makes total sense of course!
> 
> I know that i implemented the "translation available" check in a way to check that the selected language is valid, so that one would return false in case the app language is unsupported. So only the settings UI would be misleading as it the suggests "app language" even if that one is not available. How about showing it as "app language: " and disable the menu item of the language is not supported?

Good idea - will try - and then use "disabled" as default?

ztNFny:
I'd keep app language as default to make the feature easier to discover.

murggel:
> I'd keep app language as default to make the feature easier to discover.

Ah ok, then the message will be shown, that the language is not supported

ztNFny:
> I guess, I have to address the case, that the target language is not supported...

Unsupported target language is caught here:
```
    private static void getTranslator(final String lng, final Consumer<Translator> consumer) {
        final String targetLng = TranslateLanguage.fromLanguageTag(Settings.getTranslationTargetLanguage().getCode());
        if (null == targetLng) {
            return;
        }
```

This is in CacheDetailActivity handled here: 
                   ```
 if (null == translator) {
                        binding.descriptionTranslateNote.setText(R.string.translator_translation_initerror);
                        return;
                    }
```

So I think the required change would be for the new option in settings to indicate whether the app language is supported.

OR ... as an alternative: How about actually not using app language but initializing the target language value with app language but making it a static value after that? (to clarify: When you first start c:geo and your system language is german the the target language preference is set to "de" and from then on it's just that value, the same as if it had been manually selected). The only question then is what should be the fallback if the language is unsupported - how about initializing it as "en" then?

murggel:
I currently implemented it as gathering the real language code "on demand", so only at translation time it would be detected as unsupported and a message would be shown to the user.

But I like the idea to already show it in the selection as unsupported...
Will have a look after the weekend  - will be on the road...

murggel:
Show current application language
grafik

Show to the user, if application language is not supported for translation and fallback to "undeleteable" language
grafik

PR #16904, opened 2025-05-02
Catch links to c:geo settings in Markdown (fix #16838)

### LINKED ISSUE #16838: Lint: Error:UnsafeImplicitIntentLaunch, Warning:UnsafeIntentLaunch
With gradle 8.1, new lint checks regarding intent launch were integrated.
From gradle 8.2, the UnsafeImplicitIntentLaunch check is active and an error.

With moving c:geo to gradle 8.2 (in #16819), lint is running into 24 errors and 2 warnings regarding intent launch.

From the lint result:

>   Explanation for issues of type "UnsafeImplicitIntentLaunch":
   This intent matches a non-exported component within the same app. In many
   cases, the app developer could instead use an explicit Intent to send
   messages to their internal components, ensuring that the messages are
   safely delivered without exposure to malicious apps on the device. Using
   such implicit intents will result in a crash in an upcoming version of
   Android.

>   Explanation for issues of type "UnsafeIntentLaunch":
   Intent that potentially could come from an untrusted source should not be
   launched from an unprotected component without first being sanitized. See
   this support FAQ for details:
   https://support.google.com/faqs/answer/9267555

For both points exist quick fixes in AS. I'm unsure of the implications.

The `error` is now moved to `warning` in #16819 in `build.gradle` and should be removed after the issue is fixed.

moving-bits:
## Description
Filter links starting with `cgeo-setting://` in markdown texts (like our changelog) and use an explicit intent for it, which avoids the overlapping implicit intent definitions in `AndroidManifest.xml` leading to the lint errors described in #16838.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -281,12 +281,6 @@
                 <action android:name="android.intent.action.APPLICATION_PREFERENCES" />
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="@string/settings_scheme" />
-            </intent-filter>
         </activity>
         <activity android:name=".settings.ViewSettingsActivity" />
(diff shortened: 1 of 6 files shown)


bekuno:
retest this please

PR #16907, opened 2025-05-03
fix #16822: wherigo, add possibility to contact cartridge author via wherigo.com

### LINKED ISSUE #16822: Proposal to change Wherigo error message
On receiving an error message from a Wherigo cartridge, we currently print something like this:

```
Wherigo Problembericht Enginefehler: you hit a bug! please report at [openwig.googlecode.com](http://openwig.googlecode.com/) and i'll fix it for you! java.lang.RuntimeException: Tried to call nil stack trace: at @E:/webtempfiles/3641b205-3406-4cb4-bf03-100c6bfb6e67/_cartridge-new.enc:3314 at @E:/webtempfiles/3641b205-3406-4cb4-bf03-100c6bfb6e67/_cartridge-new.enc:2049 at @E:/webtempfiles/3641b205-3406-4cb4-bf03-100c6bfb6e67/_cartridge-new.enc:2046
```

This leads to users hitting the "send error to c:geo support" button, even though we can't really help them on support in 99.9% of all cases.

I propose to
- explain the user that they have hit an internal bug in the cartridge which only the cartridge author can fix,
- and maybe redirect them to the wherigo.com page for this cartridge, where the author of the cartridge is listed

on hitting errors that come from the Wherigo-internal engine.

eddiemuc:
fix #16822: wherigo, add possibility to contact cartridge author via wherigo.com

Allows user to navigate to cartridge detail page on wherigo.com in both the cardtridge info dialog and the error dialog. The later is meant so the user may contact the cartridge author (via wherigo.com) if he/she wants to report an error to him/her.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/apps/cache/WhereYouGoApp.java
+++ b/main/src/main/java/cgeo/geocaching/apps/cache/WhereYouGoApp.java
@@ -30,7 +30,7 @@ public class WhereYouGoApp extends AbstractGeneralApp {
     public void navigate(@NonNull final Context context, @NonNull final Geocache cache) {
         final List<String> guids = WherigoUtils.getWherigoGuids(cache);
         if (!guids.isEmpty()) {
-            context.startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(WherigoUtils.getWherigoUrl(guids.get(0)))));
+            context.startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(WherigoUtils.getWherigoDownloadUrl(guids.get(0)))));
         }
     }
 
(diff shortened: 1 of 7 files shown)


bekuno:
retest this please

eddiemuc:
Failing tests are unrelated

JakeDot:
maybe a dumb question but why would somebody use wherigo.com as a messenger, when they can do it on gc.com?

JakeDot:
I think there should be a toggle option (or other) in settings which Site the user wants to use...

eddiemuc:
> maybe a dumb question but why would somebody use wherigo.com as a messenger, when they can do it on gc.com?

- cartridge author is not necessarily the owner of the cache who references it (eg Reverse Wherigo by Waldmeister)
- although it is very common, not all wherigo cartridges have an associated cache. Wherigos can be played independent from GC.com

eddiemuc:
> I think there should be a toggle option (or other) in settings which Site the user wants to use...

As of now we only support wherigo.com

JakeDot:
I will implement a sample WIG platform and integrate it in my repo. https://ã.at/jakedot.cgeo

JakeDot:
> cartridge author is not necessarily the owner of the cache who references it (eg Reverse Wherigo by Waldmeister)

Could we offer a dialog or similar to contact either of these two on gc.com or is that Overkill?

PR #16908, opened 2025-05-03
UnifiedMap: Fix rotation popup

moving-bits:
## Description
Changing rotation mode via popup did not work, as the map fragment's `setMapRotation` method got called instead of the activity's `setMapRotation`.


bekuno:
retest this please

moving-bits:
failing tests are unrelated

PR #16910, opened 2025-05-03
WIP: Make offline translation preserve HTML

moving-bits:
## Description
This PR tries to preserve HTML during offline translation of a cache description.

|Original|Translated|
|---|---|
|image|image| 

Partly works already, images are preserved, as well as most formatting, but only _some_ links:

|Original|Translated|
|---|---|
|image|image|

Also, depending on the specific source text, sometimes paragraphs are lost:

|Original|Translated|
|---|---|
|image|image|

This may be some difference between `JSoup.parse()` and Android's `HTML.from()`, but I haven't cross-checked yet.

## Technical info
The old way of translating a cache listing was to get the listing text's `Spannable.getText()`, translate this string, and display the result. Due to calling `getText()` all HTML formatting gets removed.

The approach of this PR is to translate the text directly after loading it from the database, then apply `HTML.from()` and all the "magic" for loading the images etc. Google's ML translation misses some HTML attributes, though, which lead to "HTML reminders" being displayed in the translation (especially some style attributes), thus I added another step: First parse the string loaded from the database with `JSoup.parse()`, then translate all pure-text blocks (and leave the other unchanged), then apply c:geo's "magic" for images etc., then display the result.

Haven't tested yet with renderings other than "original".

Setting WIP for now.

Any input on how to fix the open issues would be highly welcomed.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
@@ -20,9 +20,9 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -35,6 +35,9 @@ import com.google.mlkit.nl.translate.TranslateRemoteModel;
 import com.google.mlkit.nl.translate.Translation;
 import com.google.mlkit.nl.translate.Translator;
 import com.google.mlkit.nl.translate.TranslatorOptions;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.TextNode;
 
 public class OfflineTranslateUtils {
 
@@ -127,18 +130,21 @@ public class OfflineTranslateUtils {
     }
 
     public static void translateParagraph(final Translator translator, final OfflineTranslateUtils.Status status, final String text, final Consumer<String> consumer, final Consumer<Exception> errorConsumer) {
-        final List<String> origText = Arrays.asList(text.split("\n"));
-        final List<String> translatedText = new ArrayList<>();
-        for (String p : origText) {
-            translator.translate(p)
+        final Document document = Jsoup.parseBodyFragment(text);
+        final List<TextNode> elements = document.children().select("*").textNodes();
+        final AtomicInteger remaining = new AtomicInteger(elements.size());
+        for (TextNode textNode : elements) {
+            translator.translate(textNode.text())
                     .addOnSuccessListener(translation -> {
-                        translatedText.add(translation);
-                        if (origText.size() == translatedText.size()) {
-                            consumer.accept(String.join("\n", translatedText));
+                        textNode.text(translation);
+                        // check if all done
+                        if (remaining.decrementAndGet() == 0) {
+                            consumer.accept(document.body().html());
                             status.updateProgress();
                         }
                     })
                     .addOnFailureListener(e -> {
+                        Log.e("err: " + e.getMessage());
                         status.abortTranslation();
                         errorConsumer.accept(e);
                     });
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -179,6 +179,7 @@ import java.util.Set;
 import java.util.function.Consumer;
 
 import com.google.android.material.button.MaterialButton;
+import com.google.mlkit.nl.translate.Translator;
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.disposables.CompositeDisposable;
 import io.reactivex.rxjava3.functions.Function;
@@ -1748,7 +1749,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             binding.getRoot().setVisibility(View.VISIBLE);
 
             // load description
-            reloadDescription(this.getActivity(), cache, true, 0, activity.descriptionStyle);
+            reloadDescription(this.getActivity(), cache, true, 0, activity.descriptionStyle, null, null, null);
 
             //check for geochecker
             final String checkerUrl = CheckerUtils.getCheckerUrl(cache);
@@ -1874,7 +1875,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             if (cda.translationStatus.isTranslated()) {
                 cda.translationStatus.setNotTranslated();
-                reloadDescription(cda, cache, true, 0, cda.descriptionStyle);
+                reloadDescription(cda, cache, true, 0, cda.descriptionStyle, null, null, null);
                 if (TextUtils.containsHtml(cache.getHint())) {
                     binding.hint.setText(HtmlCompat.fromHtml(cache.getHint(), HtmlCompat.FROM_HTML_MODE_LEGACY, new HtmlImage(cache.getGeocode(), false, false, false), null), TextView.BufferType.SPANNABLE);
                 } else {
@@ -1901,10 +1902,13 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                         binding.descriptionTranslateNote.setText(getResources().getText(R.string.translator_translation_error, error.getMessage()));
                         binding.descriptionTranslateButton.setEnabled(false);
                     };
+                    reloadDescription(cda, cache, true, 0, cda.descriptionStyle, translator, cda.translationStatus, errorConsumer);
+        /*
                     OfflineTranslateUtils.translateParagraph(translator, cda.translationStatus, binding.description.getText().toString(), translatedText -> {
                         displayDescription(getActivity(), cache, translatedText, binding.description);
                         binding.descriptionTranslateNote.setText(String.format(getString(R.string.translator_translation_success), sourceLng));
                     }, errorConsumer);
+        */
                     OfflineTranslateUtils.translateParagraph(translator, cda.translationStatus, binding.hint.getText().toString(), binding.hint::setText, errorConsumer);
                 });
         }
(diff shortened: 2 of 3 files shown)


main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -127,21 +132,38 @@ public static void getTranslator(final Activity activity, final Language sourceL
...
+        final Document document = Jsoup.parse(text);
+        final Elements elements = document.children().select("*");
+        final AtomicInteger remaining = new AtomicInteger(elements.size());
+        for (Element element : elements) {
+            final String current = element.ownText();
+            if (StringUtils.isBlank(current)) {
+                remaining.decrementAndGet();
+            } else {
+                // translate
+                final String[] origText = current.split("\n");
```
eddiemuc, line 144: When translating HTML text snippets it shouldn't be necessary to preserve line breaks. I think it should be enough to translate `current` in one go.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -127,21 +132,38 @@ public static void getTranslator(final Activity activity, final Language sourceL
...
-                    })
-                    .addOnFailureListener(e -> {
-                        status.abortTranslation();
-                        errorConsumer.accept(e);
-                    });
+        final Document document = Jsoup.parse(text);
+        final Elements elements = document.children().select("*");
+        final AtomicInteger remaining = new AtomicInteger(elements.size());
+        for (Element element : elements) {
+            final String current = element.ownText();
```
eddiemuc, line 139: You extract text using `ownText()`...

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -127,21 +132,38 @@ public static void getTranslator(final Activity activity, final Language sourceL
...
+                // translate
+                final String[] origText = current.split("\n");
+                final AtomicInteger remaining2 = new AtomicInteger(origText.length);
+                for (int i = 0, max = origText.length; i < max; i++) {
+                    final int j = i;
+                    translator.translate(origText[i])
+                        .addOnSuccessListener(translation -> {
+                            origText[j] = translation;
+                            if (remaining2.decrementAndGet() == 0) {
+                                element.text(String.join("\n", origText));
```
eddiemuc, line 152: ...but you set it again using `text()`. Could that build up a difference in resulting html?

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
```diff
@@ -127,21 +132,38 @@ public static void getTranslator(final Activity activity, final Language sourceL
...
-                    .addOnFailureListener(e -> {
-                        status.abortTranslation();
-                        errorConsumer.accept(e);
-                    });
+        final Document document = Jsoup.parse(text);
+        final Elements elements = document.children().select("*");
+        final AtomicInteger remaining = new AtomicInteger(elements.size());
+        for (Element element : elements) {
+            final String current = element.ownText();
+            if (StringUtils.isBlank(current)) {
```
eddiemuc, line 140: Ignoring blanks is maybe causing missing paragraphs in result. Maybe add a `\n` for each blank? (Just guessing here)

eddiemuc:
> Perhaps there's too much magic going on in JSoup. How about just parsing for tags (starts with `<`, ends with `>`) and translate everything between two tags?
> 
> To test whether it is JSoup changing the HTML or something else, maybe just apply `JSoup.parse` to description, then simply convert back to HTML from resulting `Document` (w/o translating/modifying any texts) using `.body().html()`, then compare how result is displayed.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
moving-bits: Thanks, @eddiemuc, for taking the time to review.

The `split()` code is from the original parsing. As `ownText()` seems to normalize the text, removing newlines along the way, it's indeed no longer necessary. In JSoup docs I found a method `wholeOwnText()` which seems to preserve such whitespace, which might solve the problem with the vanishing paragraph breaks - I'll play with it (and either find a solution for the paragraphs or remove the special newline handling).

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
moving-bits: `text()` would include the text of all siblings, which to my understanding would break formatting.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
moving-bits: There is no `setOwnText()` or something like that - but as `text()` belongs to a single element, it just sets that's (own) text.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
moving-bits: I'll dive into it to see where the line breaks get lost, either in the first branch here, or whether `wholeOwnText()` in the second branch will help.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
eddiemuc: > The `split()` code is from the original parsing. 

I believe the original parsing needed this because it parsed text already converted by `Html.fromHtml()`. In the result of this, `<p>` and `<br>` was already converted into \n

However, your approach parses HTML text nodes in which there should no `\n` being present or they shouldn't have any effect on the display anyway.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
eddiemuc: Using text would indeed break formatting.

But using ownText could also lead to unwanted results. Eg for the example in JSoup documentation:



Using `ownText` and `text` as you do right now would lead to `<p>Hello now!<p>` (removing the bold `there`), wouldn't it? (Can't try, not on my machine)

moving-bits:
> Perhaps there's too much magic going on in JSoup. How about just parsing for tags (starts with `<`, ends with `>`) and translate everything between two tags?

My hope is (was), that JSoup (or similar parsers) are better in handling non-well-formed HTML.
 
> To test whether it is JSoup changing the HTML or something else, maybe just apply `JSoup.parse` to description, then simply convert back to HTML from resulting `Document` (w/o translating/modifying any texts) using `.body().html()`, then compare how result is displayed.

I've looked into the HTML source. It's a complicated bundle of intertwined `<table>`, `<span>` and `<br>` elements (typical for generated HTML), and, indeed, JSoup strips some `<br>` here when doing a simple `JSoup.parse(html).body().html()`. I'll have a look if I can configure something here, otherwise there's probably not much that I can do.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
eddiemuc: > In JSoup docs I found a method wholeOwnText() which seems to preserve such whitespace...

I don't believe this solves the problem. White spaces/line breaks in HTML **should** be ignored; keeping them would be a bug.

moving-bits: After adapting + comparing with the original HTML: Using `wholeOwnText()` does not make a difference here, as the line breaks in that specific case come from a couple of `<br>` between even more `<span>`, and JSoup somehow strips the `<br>` :-(

I have remove handling of newlines for now.

moving-bits: > > In JSoup docs I found a method wholeOwnText() which seems to preserve such whitespace...
> 
> I don't believe this solves the problem. White spaces/line breaks in HTML **should** be ignored; keeping them would be a bug.

My assumption was, that `HTML.from()` somehow translates two consecutive newlines into a paragraph, but that assumption was wrong. And, having looked in the source for the specific cache, the "paragraphs" were done quite differently.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
eddiemuc: Come to think of it, keeping formatting in translation is far from trivial anyway. Because order of words and their meaning can of course change during translation. Eg if sentence "Hello there now!" would be translated to "Hallo jetzt da draußen!", then how should we know that now the last two words must be bold instead of the middle one?

moving-bits:
Found a way that looks much better: Instead of retrieving all elements (and deciding on "empty" elements or not whether to translate them), I let JSoup give me just the text nodes and only translate them. Looks way better now:

image.image

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
moving-bits: Switching to translating just text nodes instead of elements resolved the `text()` / `ownText()` issue, but comes at the cost of having to translate incomplete sentences sometimes (which I had though my original approach already would do). Let's see how it works in real life.

moving-bits:
"Old checks" are green, so let's give it a try in the nightlies.

main/src/main/java/cgeo/geocaching/utils/OfflineTranslateUtils.java
eddiemuc: > Switching to translating just text nodes instead of elements resolved the `text()` / `ownText()` issue, but comes at the cost of having to translate incomplete sentences sometimes (which I had though my original approach already would do). Let's see how it works in real life.

Tbh I would personally prefer translating whole sentences (and loosing formatting) over keeping format (and translating sentence parts only). This is up to debate of course.

What about neglecting basic formats like bold/italic/underline in favor of whole sentences, but keeping separate trandlations for list elements and in table cells?

PR #16920, opened 2025-05-07
Use new coordinate entry dialog in all places except waypoint page

ReggieDoodle:
## Description
Extends the earlier work to utilise the new dialog in these additional places

- User define cache 'Set Coordinates' option
- Search card 'Coordinates'
- Logging a Geokrety TB

## Related issues

PR16799 
PR16889

## Additional context
Also included these improvements

- Constructor is no longer public as recommended by @eddiemuc 
- The 'My coordinates' button now shows accuracy as per the old dialog

I have tested the above use cases and replicated the same functionality as before except in these cases

1. User defined cache does not have a button to use the 'cache coordinates' - seem pointless when the purpose is to change the existing coordinates 
2. GK TB does not have the 'calculate coordinates' button. The rational here is you are dropping a TB in or visiting a cache so you must presumably already have the coordinates and do not need to compute them.

The waypoint entry page is NOT included in this change as that is more complex due to it being coupled to the calculate coordinate global dialog.  That will be done later

Note:
I replicated the usage of GeoDIrHandler from the old dialog (which was Fragment based) to show the positional accuracy by calling 

_geoUpdate.start(GeoDirHandler.UPDATE_GEODATA)_ 

when the dialog was shown but couldn't find away to stop it - the fragment dialog disposes of it when onPause is called.

Do I need to kill it also or does it get stopped automatically when _dialog.dismiss()_ is called

Tks
Dave

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -82,7 +82,6 @@ import cgeo.geocaching.ui.ToggleItemType;
 import cgeo.geocaching.ui.TrackableListAdapter;
 import cgeo.geocaching.ui.UserClickListener;
 import cgeo.geocaching.ui.ViewUtils;
-import cgeo.geocaching.ui.dialog.CoordinatesInputDialog;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.EditNoteDialog;
 import cgeo.geocaching.ui.dialog.EditNoteDialog.EditNoteDialogListener;
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -845,9 +845,9 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         });
     }
 
-    private void setCoordinates(Activity activity) {
+    private void setCoordinates(final Activity activity) {
         ensureSaved();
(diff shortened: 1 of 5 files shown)


eddiemuc:
> I replicated the usage of GeoDIrHandler from the old dialog (which was Fragment based) to show the positional accuracy by calling geoUpdate.start(GeoDirHandler.UPDATE_GEODATA) when the dialog was shown but couldn't find away to stop it

@ReggieDoodle the call to `geoUpdate.start()` returns an object of type `Disposable`. You stop the handler by calling method `dispose()` on this object.

ReggieDoodle:
> > I replicated the usage of GeoDIrHandler from the old dialog (which was Fragment based) to show the positional accuracy by calling geoUpdate.start(GeoDirHandler.UPDATE_GEODATA) when the dialog was shown but couldn't find away to stop it
> 
> @ReggieDoodle the call to `geoUpdate.start()` returns an object of type `Disposable`. You stop the handler by calling method `dispose()` on this object.

Thanks

bekuno:
ok to test

eddiemuc:
There seem to be some lint errors and a check style error in the build:

https://ci.cgeo.org/job/cgeo-CI_PR-build/11912/

eddiemuc:
> > There seem to be some lint errors and a check style error in the build:
> > 
> > https://ci.cgeo.org/job/cgeo-CI_PR-build/11912/
> 
> Yeah, put my hand up to the check style one but I think the lint ones are WhereIGo?

Right. Sorry. Those should self-correct once new translations are done via crowdin

ReggieDoodle:
Retest this please

moving-bits:
Thanks, @ReggieDoodle, for your contribution, highly appreciated!

PR #16924, opened 2025-05-08
Fix #13186: Correct file naming counter during file name creation

### LINKED ISSUE #13186: File names use wrong counter
Generated file names get a number suffix that is incremented each time a file name is generated.

https://github.com/cgeo/cgeo/blob/df7ee033cb43f9a08cad21f11001af574220c5d5/main/src/cgeo/geocaching/utils/FileNameCreator.java#L57

This counter starts with 2, because the AtomicInteger is initialized with 1, but `addAndGet()` increments the value before it is returned.

`getAndIncrement()` can be used to simplify the call and fix this issue.

On the other hand it would be even better if the counter is only incremented when a file with the same name already exists.

kennygchen:
## Description
Fixes the initialization of the file name counter in `FileNameCreator.java`.
Previously, the counter was initialized with `1` and used `addAndGet(1)`, which caused the first generated file name to have a suffix of `2`.

## Related issues
Fixes #13186

## Additional context
- Updated `AtomicInteger` initialization to start at `0`

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/FileNameCreator.java
+++ b/main/src/main/java/cgeo/geocaching/utils/FileNameCreator.java
@@ -23,7 +23,7 @@ public class FileNameCreator {
     public static final FileNameCreator TRACKFILE = new FileNameCreator("track", "gpx");
     public static final FileNameCreator OFFLINE_LOG_IMAGE = new FileNameCreator("cgeo-image-%s", "jpg");
 
-    private final AtomicInteger fileNameCounter = new AtomicInteger(1);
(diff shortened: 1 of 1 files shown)


eddiemuc:
Ok to test

eddiemuc:
@kennygchen you (accidentally?) included changes to `checkstyle-idea.xml` into your commit, can you please remove those? Thanks

kennygchen:
It is removed. Sorry about that

eddiemuc:
> It is removed. Sorry about that

No problem. Happens to me all the time :-)

eddiemuc:
Fix lgtm. Thanks for diving into this!

PR #16930, opened 2025-05-10
Preserve time-stamp of own log on refreshing the cache (fix #9653)

### LINKED ISSUE #9653: keep log time at reload if date is unchanged
When I reload a cache (example: Haldenstelle 1) that I have found, my log time is set to 00:00

Since only the date but not the time is given by the reload, the time should be not altered, if the date of the log stays the same.

**Additional context**
IMG_20210103_143338

### LINKED ISSUE #16898: Editing a log set it's time to mdnight (00:00:00)
### Describe your problem!

When editing a log, the time get reset from the usual T12:00:00 to T00:00:00.
This will put the caches in the wrong order in the (classic) overview page.

### How to reproduce?

With C:GEO version 2025.04.02 Log a cache. Log another cache. Edit the log of the first cache.

The second logged cache will appear to be logged first on geocaching.com/my
When retrieving the cache via the api the time is set to T00:00:00

The first logged cache will stay as first logged cache on geocaching.com/my
When retrieving the cache via the api the time is not changed (usually T12:00:00)

### Reproducible

Yes

### c:geo Version

2025.04.02

### System information

```text

```

### Additional Information

_No response_

murggel:
## Description

Keep time of time-stamp of stored own logs, if online log has same date.

Also for offline log, if the log-text is empty (to support keeping time-stamp for logging via field-notes)

## Related issues

fix #9653
fix #16898

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -1720,12 +1720,13 @@ public final class GCParser {
      */
     private static void mergeFriendsLogs(final List<LogEntry> mergedLogs, final Iterable<LogEntry> logsToMerge) {
         for (final LogEntry log : logsToMerge) {
-            if (mergedLogs.contains(log)) {
-                final LogEntry friendLog = mergedLogs.get(mergedLogs.indexOf(log));
-                final LogEntry updatedFriendLog = friendLog.buildUpon().setFriend(true).build();
-                mergedLogs.set(mergedLogs.indexOf(log), updatedFriendLog);
-            } else {
+            final int logIndex = mergedLogs.indexOf(log);
+            if (logIndex < 0) {
                 mergedLogs.add(log);
+            } else {
+                final LogEntry friendLog = mergedLogs.get(logIndex);
+                final LogEntry updatedFriendLog = friendLog.buildUpon().setFriend(true).build();
+                mergedLogs.set(logIndex, updatedFriendLog);
             }
         }
     }
@@ -1736,15 +1737,15 @@ public final class GCParser {
             logTimesToMergeMap.put(logToMerge.serviceLogId, logToMerge);
         }
 
-        for (final LogEntry mergedLog : mergedLogTimes) {
+        for (int i = 0; i < mergedLogTimes.size(); i++) {
+            final LogEntry mergedLog = mergedLogTimes.get(i);
             final LogEntry logToMerge = logTimesToMergeMap.get(mergedLog.serviceLogId);
             if (logToMerge != null) {
                 final Date dateTimeLogTime = new Date(mergedLog.date);
                 final Date logTime = new Date(logToMerge.date);
                 if (!logTime.equals(dateTimeLogTime) && DateUtils.isSameDay(dateTimeLogTime, logTime)) {
                     final LogEntry updatedTimeLog = mergedLog.buildUpon().setDate(logToMerge.date).build();
-                    final int logIndex = mergedLogTimes.indexOf(mergedLog);
-                    mergedLogTimes.set(logIndex, updatedTimeLog);
+                    mergedLogTimes.set(i, updatedTimeLog);
                 }
             }
         }
@@ -1755,12 +1756,11 @@ public final class GCParser {
             return;
         }
 
-        // TODO iterate via index to avoid indexOf
-        for (final LogEntry mergedLog : mergedLogTimes) {
+        for (int i = 0; i < mergedLogTimes.size(); i++) {
+            final LogEntry mergedLog = mergedLogTimes.get(i);
             if (logToMerge.isMatchingLog(mergedLog)) {
                 final LogEntry updatedTimeLog = mergedLog.buildUpon().setDate(logToMerge.date).build();
-                final int logIndex = mergedLogTimes.indexOf(mergedLog);
-                mergedLogTimes.set(logIndex, updatedTimeLog);
+                mergedLogTimes.set(i, updatedTimeLog);
 
                 break;
             }
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -18,6 +18,7 @@ import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
 import cgeo.geocaching.log.LogTypeTrackable;
+import cgeo.geocaching.log.OfflineLogEntry;
 import cgeo.geocaching.models.GCList;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Image;
@@ -76,8 +77,10 @@ import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
 import okhttp3.Response;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.time.DateUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.text.StringEscapeUtils;
 import org.jsoup.Jsoup;
@@ -1593,25 +1596,8 @@ public final class GCParser {
             return;
         }
 
-        DisposableHandler.sendLoadProgressDetail(handler, R.string.cache_dialog_loading_details_status_logs);
-        final String userToken = parseUserToken(page);
-        final Observable<LogEntry> logs = getLogs(userToken, Logs.ALL);
-        final Observable<LogEntry> ownLogs = getLogs(userToken, Logs.OWN).cache();
-        final Observable<LogEntry> specialLogs = Settings.isFriendLogsWanted() ?
-                Observable.merge(getLogs(userToken, Logs.FRIENDS), ownLogs) : Observable.empty();
-        final Single<List<LogEntry>> mergedLogs = Single.zip(logs.toList(), specialLogs.toList(),
-                (logEntries, specialLogEntries) -> {
-                    mergeFriendsLogs(logEntries, specialLogEntries);
-                    return logEntries;
-                }).cache();
-        mergedLogs.subscribe(logEntries -> DataStore.saveLogs(cache.getGeocode(), logEntries, true));
-        if (cache.isFound() || cache.isDNF()) {
-            ownLogs.subscribe(logEntry -> {
-                if (logEntry.logType.isFoundLog() || (!cache.isFound() && cache.isDNF() && logEntry.logType == LogType.DIDNT_FIND_IT)) {
-                    cache.setVisitedDate(logEntry.date);
-                }
-            });
-        }
+        // merge log-entries (friend-logs and log-times)
+        mergeAndStoreLogEntries(cache, page, handler);
 
         //add gallery images if wanted
         addImagesFromGallery(cache, handler);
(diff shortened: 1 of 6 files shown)


main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1597,21 +1599,45 @@ private static void getExtraOnlineInfo(@NonNull final Geocache cache, final Stri
...
+                getLogs(userToken, Logs.FRIENDS).cache() : Observable.empty();
+
+        final List<LogEntry> ownLogsFromDb;
+        if (!ownLogs.isEmpty().blockingGet()) {
+            ownLogsFromDb = DataStore.loadLogsOfAuthor(cache.getGeocode(), GCConnector.getInstance().getUserName(), true);
+        } else {
+            ownLogsFromDb = Collections.emptyList();
+        }
+
+        final Single<List<LogEntry>> ownTimeLogs = Single.zip(logs.toList(), ownLogs.toList(),
```
murggel, line 1612: Please help here with correct implementation.
Since I am not familiar with the Observable in Java, I didn't manage to change the time of the ownLogEntry, so the following methods are aware of the change.

This implementation with Single.zip and unused first argument works, but is not good...

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1597,21 +1599,45 @@ private static void getExtraOnlineInfo(@NonNull final Geocache cache, final Stri
...
-        final Observable<LogEntry> specialLogs = Settings.isFriendLogsWanted() ?
-                Observable.merge(getLogs(userToken, Logs.FRIENDS), ownLogs) : Observable.empty();
-        final Single<List<LogEntry>> mergedLogs = Single.zip(logs.toList(), specialLogs.toList(),
-                (logEntries, specialLogEntries) -> {
+        final Observable<LogEntry> friendLogs = Settings.isFriendLogsWanted() ?
+                getLogs(userToken, Logs.FRIENDS).cache() : Observable.empty();
+
+        final List<LogEntry> ownLogsFromDb;
+        if (!ownLogs.isEmpty().blockingGet()) {
+            ownLogsFromDb = DataStore.loadLogsOfAuthor(cache.getGeocode(), GCConnector.getInstance().getUserName(), true);
```
murggel, line 1607: Note to myself: Check implementation with deactivated friend-option

murggel:
Note to myself: check for adding unittests / integrationtests

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: @eddiemuc / @moving-bits 
I searches the last days for a solution to avoid the unnecessary parameter, but didn't find any.
Can one of you please help me?

murggel: Now I tried with making the log-date editable and using blockingGet(), that works

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Works
* applyCommonsToOwnLog is called on own logs, where the friend-flag is set
* if for what ever reason, the friend-flag is not set in the db, we first read the entries with active flag, and if there are no such entries, we read the entries without active flag

main/src/main/java/cgeo/geocaching/log/LogEntry.java
```diff
@@ -57,7 +57,7 @@ public class LogEntry implements Parcelable {
     /** The log message */
     @NonNull public final String log;
     /** The log date */
-    public final long date;
+    public long date;
```
murggel, line 60: I made this member editable, because the objects are observable and creating a new one instead of changing an existing one wouldn't work

murggel:
Ready for a first review

main/src/main/java/cgeo/geocaching/log/LogUtils.java
```diff
@@ -495,4 +498,28 @@ private static ImmutableTriple<ImageResult, List<Image>, Runnable> updateImages(
     private static boolean imagePropertiesEqual(final Image oldImg, final Image newImg, final String newTitle) {
         return Objects.equals(oldImg.title, newTitle) && Objects.equals(oldImg.getDescription(), newImg.getDescription());
     }
+
+    /**
+     * Checks, if the logs are representing the same log-entry.
+     * Log-text can be empty (to deal with field-notes)
+     */
+    public static boolean isMatchingLog(final LogEntry log1, final LogEntry log2) {
```
eddiemuc, line 506: Move method to class `LogEntry`?

main/src/main/java/cgeo/geocaching/log/LogEntry.java
eddiemuc: This breaks `LogEntry` from being immutable although it still "looks like" it is. 

Setting date should normally work with

`logEntry.buildUpon().setDate().build()`

(This creates a new `LogEntry` instance of course)

That doesn't work in your case?

(I am not against making Log entry mutable, but then it should be done for whole class which is a bigger task)

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -18,6 +18,8 @@
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
 import cgeo.geocaching.log.LogTypeTrackable;
+import cgeo.geocaching.log.LogUtils;
```
eddiemuc, line 21: You placed the logic into GCParser. Wouldn't that mean that it won't work for eg OC caches? Or are those not affected by the bug?

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1597,22 +1601,45 @@ private static void getExtraOnlineInfo(@NonNull final Geocache cache, final Stri
...
         final Observable<LogEntry> logs = getLogs(userToken, Logs.ALL);
         final Observable<LogEntry> ownLogs = getLogs(userToken, Logs.OWN).cache();
-        final Observable<LogEntry> specialLogs = Settings.isFriendLogsWanted() ?
-                Observable.merge(getLogs(userToken, Logs.FRIENDS), ownLogs) : Observable.empty();
-        final Single<List<LogEntry>> mergedLogs = Single.zip(logs.toList(), specialLogs.toList(),
-                (logEntries, specialLogEntries) -> {
-                    mergeFriendsLogs(logEntries, specialLogEntries);
-                    return logEntries;
-                }).cache();
-        mergedLogs.subscribe(logEntries -> DataStore.saveLogs(cache.getGeocode(), logEntries, true));
```
eddiemuc, line 1607: This is unrelated to this PR, but why the hell is a Parser class manipulating the database???

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1597,22 +1601,45 @@ private static void getExtraOnlineInfo(@NonNull final Geocache cache, final Stri
...
+        final Observable<LogEntry> friendLogs = Settings.isFriendLogsWanted() ?
```
eddiemuc, line 1604: Also unrelated to this PR, but we should really remove all the boilerplate rxutils constructs here. The main blocking thing (loading the page from Internet) is done blocking anyway, all the data parsing happens in no-time in contrast to that. And the whole observable stuff makes the code almost impossible to read.

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1703,6 +1727,34 @@ private static void mergeFriendsLogs(final List<LogEntry> mergedLogs, final Iter
         }
     }
 
+    private static void mergeLogTimes(final List<LogEntry> mergedLogTimes, final Iterable<LogEntry> logTimesToMerge) {
+        for (final LogEntry logToMerge : logTimesToMerge) {
```
eddiemuc, line 1731: Quadratic complexity.

Maybe use a lookup hashtable by logId to prevent this?

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Because it is manipulating e.g. the visited date of the cache...

Maybe we need to rework the workflow, so updating the visited date is done outside

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Good idea, will  check on that

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Good question, will check on this

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Since I am not familiar with that stuff, I couldn't do it

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
eddiemuc: Exactly, update should happen outside of the parser.
But as said, this is unrelated to your PR

main/src/main/java/cgeo/geocaching/log/LogUtils.java
eddiemuc: It's no biggie. Is done like this in other places as well....

But for the books: OOP and DDD is about encapsulating data and related logic together....

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: But makes it probably easier for me

main/src/main/java/cgeo/geocaching/log/LogUtils.java
murggel: I guess, I will move the whole logic of updating the found-date to a n own class, then this logic of finding a matching cache will be moved there, because it is kind of a special logic for this case. .

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Seems that from OC we receive already the correct time, so need to merge the time stamp

murggel: Don't know if other connectors support the "friend", which is also considered here

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Since `getLogs` is only used inside GCParser for `getExtraOnlineInfo` I think I can change it to a "simple" list. If it is not observable anymore, then a new LogEntry should be possible to handle. Will check on that.

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Done

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: @eddiemuc  Hm - getLogs does a http request, which I guess should stay blocking... 
So I am not sure, how to change that. Any help appreciated

main/src/main/java/cgeo/geocaching/log/LogUtils.java
murggel: Moved to LogEntry for now

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: I tried to use toList().blockingGet -> that seems to work

murggel: Maybe I add an marker whether the logs have changed and need to be stored, to avoid unneccesary db-access

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
murggel: Maybe I can move the method from GCParser to GCConnector, but this doesn't make it much better...

murggel:
squashed commits

murggel:
Damn time-zones - methods `sameDay` does not work correctly with our stored times for those around midnight local time (at least from germany in emulator) 

Have to check on my real device

murggel:
> Damn time-zones - methods `sameDay` does not work correctly with our stored times for those around midnight local time (at least from germany in emulator)
> 
> Have to check on my real device

On my device it seems all fine

@eddiemuc so ready for next review...

moving-bits:
Anything left to be done here, or can it be squashed & merged?

murggel:
> Anything left to be done here, or can it be squashed & merged?

@eddiemuc what do you think?

eddiemuc:
> > Anything left to be done here, or can it be squashed & merged?
> 
> @eddiemuc what do you think?

Let's merge and test in nightlies

PR #16934, opened 2025-05-10
Fix short date formatting (fix #16925)

### LINKED ISSUE #16925: c:geo shows wrong log date in logbook view
In the logs of https://coord.info/GCT4W8, the date of Geokasher's and KEMBORG's logs is shown incorrectly as 2025-12-31 (an impossible date because it is in the future):

Here is the logbook on geocaching.com:

moving-bits:
## Description
Date formatter "Y" uses "week years" in contrast to "y" using "years" (which is what we need here)
(https://docs.oracle.com/javase/8/docs/api/java/util/GregorianCalendar.html#week_year)

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -274,7 +274,7 @@ public class Settings {
     }
 
     public static int getExpectedVersion() {
-        return 9;
+        return 10;
     }
 
     private static void migrateSettings() {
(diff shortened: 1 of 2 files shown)


ztNFny:
oops, my bad. Why does every stupid language define their own incompatible date formatters ... :-(
Considering the string is stored in settings a migration routine would probably be required?

moving-bits:
Settings' migration is added, "old" checks are green, so let's merge

PR #16889, opened 2025-04-26
New Coordinate Entry Dialog - fix layout issues from initial review

ReggieDoodle:
## Description
Updates to layout recommended by @murggel in code review

1. Border / padding improved
2. Added degree & minute signs were appropriate
3. Added hints for all fields as per original dialog

## Related issues
Fix 11335 & PR 16799

PR #16799, opened 2025-04-03
Code review for Fix#11335

ReggieDoodle:
This is my first exposure to the UI in cgeo - well any android app actually - so I would like some feedback please. I’m not expecting it to be ‘production ready’.

Ticket #11335 outlines some issues with the current coordinate entry dialog and this is my attempt to recreate the same functionality in an alert based dialog. 

The design code (xml) is quite verbose atm but that was how it grew as I learned what various options did.  It can be styled and compressed no doubt.

The dialog logic is similar to the existing dialogs - some code was lifted straight from it - this is labelled as such. The main difference is that the existing one uses binding, though not in the sense I am used to it (WPF view models) so I haven’t recreated that approach.

My current approach returns a string representing latitude and longitude via the callback - this maybe should just return the validated geopoint to save parsing it twice - but it ended up like that as the origanal UI just took a string anyway so it mimicked that.

The UI in the distance filter itself has a new button to launch the dialog, as per usages elsewhere but the slider representing distance hasn’t been touched. It could so with some improvement, as per my comment on the ticket

Thanks
Dave

PR #16964, opened 2025-05-21
Call new coordinate dialog from waypoint page

ReggieDoodle:
This PR is the third stage of implementing the new coordinate entry dialog 

## Description
The waypoint coordinate entry can call the Calculate coordinate dialog which is fragment based so this manages that action.  

There is still work to do here but I wanted someone to check how I have implemented the call to the (fragment) calculator dialog from the (alert) coordinate dialog before I go any further.  At the moment the coordinates are passed back using the interface from the old dialog (which ultimately will be removed from the system) so that also needs to be addressed too - see TODO statements in the code- so I need some advice there also.

[Update : I've now fixed the interface issue in another branch]

## Related issues
PR #16799
PR #16889
PR #16920

## Additional context
As I not a regular user of the calculator option I created a discussion ticket (https://github.com/cgeo/cgeo/discussions/16957) to ask for advice on how to test it after my changes but as yet no one has commented. The tool is documented here https://manual.cgeo.org/en/waypointcalculator but I think that is out of date


bekuno:
ok to test

ReggieDoodle:
Add buttons to the dialog for clearing the coordinates and using the cache's coordinates - these are only active for waypoints - not the other usages of the dialog

murggel:
I am just playing around a little bit with the new coordinate dialog in edit waypoint activity.
Most workflows are working fine, only some needs improvement:


it seems, that the opening the dialog always uses "my coordinates" to prefill the coordinates, and the current coordinates of the waypoint are ignored

murggel:
Opening "calulate coordinate" with prefilled coordinates opens the "calculate coordinate"-dialog with empty fields, instead of prefilled with the coordinates from the first dialog

murggel:
Returning from "calculate coordinate"-dialog with "convert to plain coordinate":
only buttons "my coordinates" and "from clipboard" is shown, "cache coordinates", "calculate coordinate" and "clear coordinates" are missing

ReggieDoodle:
@murggel - thank you for testing.

it’s going to be a few days before I can look at any of these as we have some issues at home atm but please carry on….

Dave

murggel:
The "Copy from clipboard" is always visible, even if there is no valid string in the clipboard (happens also in the places, where the new dialog is already used). It should only be visible, if paste does make sense - if the content can be interpreted as Geopoint.
Since this is a small change inside the dialog, I can create a PR for that , if you want.

ReggieDoodle:
Been offline for a while following a house move - just catching up....

ReggieDoodle:
> The "Copy from clipboard" is always visible, even if there is no valid string in the clipboard (happens also in the places, where the new dialog is already used). It should only be visible, if paste does make sense - if the content can be interpreted as Geopoint. Since this is a small change inside the dialog, I can create a PR for that , if you want.

Fixed

moving-bits:
cgeo.geocaching.downloader.DownloaderTest.testOpenAndroMapsThemes is failing currently (see #17144) - can be ignored for now

ReggieDoodle:
> Opening "calculate coordinate" with prefilled coordinates opens the "calculate coordinate"-dialog with empty fields, instead of prefilled with the coordinates from the first dialog

Only occurs when source is Plain coords - seems to be okay for formatted coords

ReggieDoodle:
> Returning from "calculate coordinate"-dialog with "convert to plain coordinate": only buttons "my coordinates" and "from clipboard" is shown, "cache coordinates", "calculate coordinate" and "clear coordinates" are missing

Hi @murggel 
I'm back at the keyboard after a house move - well a keyboard - seem to have lost the original one ! ! 

Anyway, looking through your comments and I'm not sure what the use case is here ?  What is the workflow for calculating a coordinate, then converting it plain coords then changing that result with the cache coords, another calculation or clearing them?

All seems rather pointless ?   Shouldn't it just accept the result and let you save it (or cancel) without any buttons shown at all?

Dave

murggel:
> > Returning from "calculate coordinate"-dialog with "convert to plain coordinate": only buttons "my coordinates" and "from clipboard" is shown, "cache coordinates", "calculate coordinate" and "clear coordinates" are missing
> 
> Hi @murggel 
> I'm back at the keyboard after a house move - well a keyboard - seem to have lost the original one ! ! 

Uh, I know it is hard to loose a familiar tool 😪

> 
>  What is the workflow for calculating a coordinate, then converting it plain coords then changing that result with the cache coords, another calculation or clearing them?
> 
> All seems rather pointless ?   Shouldn't it just accept the result and let you save it (or cancel) without any buttons shown at all?
> 
> Dave
> 

I just compared to the original dialog / workflow, so I guess skipping the whole dialog after coming back from converting to plain coordinates would be fine for me. No additional confirmation necessary, the calculation is gone and not recoverable in that step, the final confirmation would be in the edit waypoint dialog.

But when the dialog is displayed, it should look as if it has just been opened. So showing all possible buttons.

ReggieDoodle:
Hi again @murggel - I've pushed some changes to fix showing the correct coordinates when the calculator opens and closes plus a mod to not show all the unnecessary option buttons when plain coords are returned by the calculator.

I've also check the clipboard now as you suggested

Tks 
Dave

murggel:
Hi Dave, I hope, I will find some time next week to look into your changes...

ReggieDoodle:
Ok, thank you

murggel:
Found some time to look into it..

Looks nice.

murggel:
What I am still not sure, is the outline ... it needs space and if we go with it, I would prefer to have it then in the dialog for the calculated coordinates then as well. But that can be done in a separate PR.

murggel:
> Opening "calulate coordinate" with prefilled coordinates opens the "calculate coordinate"-dialog with empty fields, instead of prefilled with the coordinates from the first dialog
>  

>  I've pushed some changes to fix showing the correct coordinates when the calculator opens
> 


Still happens to me

murggel:
>  I've pushed some changes to fix showing the correct coordinates when the calculator opens and closes plus a mod to not show all the unnecessary option buttons when plain coords are returned by the calculator.
> 

Is there a need to show this dialog after returning from converting to plain coords, when there are no buttons? I would prefer to skip that and show directly the "Edit waypoint" activity
image

ReggieDoodle:
> > Opening "calulate coordinate" with prefilled coordinates opens the "calculate coordinate"-dialog with empty fields, instead of prefilled with the coordinates from the first dialog
> >  
> 
> > I've pushed some changes to fix showing the correct coordinates when the calculator opens
> 
> Still happens to me

Ok, will take another look

ReggieDoodle:
> > I've pushed some changes to fix showing the correct coordinates when the calculator opens and closes plus a mod to not show all the unnecessary option buttons when plain coords are returned by the calculator.
> 
> Is there a need to show this dialog after returning from converting to plain coords, when there are no buttons? I would prefer to skip that and show directly the "Edit waypoint" activity 

Yes, you are right.   I will try skip that

ReggieDoodle:
> > > Opening "calulate coordinate" with prefilled coordinates opens the "calculate coordinate"-dialog with empty fields, instead of prefilled with the coordinates from the first dialog
> > >
> > 
> > 
> > > I've pushed some changes to fix showing the correct coordinates when the calculator opens
> > 
> > 
> > Still happens to me
> 
> Ok, will take another look

@murggel  This works ok for me, unless I not understanding the problem?

image

murggel:
> @murggel This works ok for me, unless I not understanding the problem?

My workflow is as follows:

| Step | Image |
| --- | --- |
|Add new waypoint |image |
| Open coordinate-dialog | image | 
| Open calculated coordinates dialog (probably change here to cache coordinates as starting values for calculation) | image |
| Empty coordinates | image |

ReggieDoodle:
> > @murggel This works ok for me, unless I not understanding the problem?
> 
> My workflow is as follows:
> 
Ah, ok - I had not tried it with plain coords and without guided mode - will have a look where its dropping them...

(so many options -  do people really use all these - does it need to be this complex ???)

murggel:
> > > @murggel This works ok for me, unless I not understanding the problem?
> > 
> > My workflow is as follows:
> > 
> Ah, ok - I had not tried it with plain coords and without guided mode - will have a look where its dropping them...
> 
> (so many options -  do people really use all these - does it need to be this complex ???)

I never use the guided mode 😅

ReggieDoodle:
Hi @murggel 
I've fixed the issue you reported where the WP coords were not passed into the calculator page when using plain format

Think this should be good to go now ?
Dave

eddiemuc:
I am not very deep in this topic. I suggest to merge and collect feedback on nightlies. @murggel ok for you?

ReggieDoodle:
There might be one or two minor issues that come up but I believe the main functionality works as expected.

Once in master I need to remove the old dialog and it's code and rename the new one.

ReggieDoodle:
There might be one or two minor issues that come up but I believe the main functionality works as expected.

Once in master I need to remove the old dialog and it's code and rename the new one.

murggel:
> I am not very deep in this topic. I suggest to merge and collect feedback on nightlies. @murggel ok for you?

@eddiemuc OK for me to test workflow and UI in nightlies.

murggel:
> There might be one or two minor issues that come up but I believe the main functionality works as expected.
> 
> 

I guess mainly try to skip the dialog after changing to plain coordinates.

This can be done in a separate PR...

ReggieDoodle:
Yes, I had a quick try to fix that last night but because one dialog calls
another I couldn't find an easy way back to the originating waypoint page
whilst keeping the data in the required state. It's an edge case. As you
say, it can be looked at later

Issue #12196, opened 2021-11-29
Support found date for ALC/UDC


Lineflyer:
### Tell us your idea!

From support mail:
When setting a lab cache to "Found" it seems not to record the date, when I set it to found. This would be helpful in general and specifically as I export them to GSAK for my record keeping.

### What solution would you suggest?

Save the current date as found date when the user sets an adventure lab cache  manually as "Found".
(Should then also apply to UDC as we also have the found-feature for those)

### Is there an alternative?

Don't implement it. Rare use case?

Lineflyer:
@ztNFny Why thumbs down. Would it interfere with other features or create an inconsistency or alike?

fm-sys:
IMHO we should rather try to switch to the partner API (only for Lab Caches). Then we don't need these ugly workarounds anymore... Or was there a specific reason why we shouldn't do this?

Lineflyer:
I am in favor of it (and if it is just to show to GS that we are willing to work on it if conditions are OK).  Would you @fm-sys be willing to help @baiti with that? I think its too much of a job for a single person.

ztNFny:
Lab API doesn't allow fetching completed labs unfortunately. You can only get completed lab stages (grouped by lab, but you don't know how many stages there are unless you fetch that from another call) or search in an area and get the completion status of the results.

Lineflyer:
@baiti 
Although this is off-topic for this issue. Maybe we should discuss whether we want to implement as @ztNFny offered help as well?
Topic for our next dev call?

baiti:
> Lab API doesn't allow fetching completed labs unfortunately. You can only get completed lab stages (grouped by lab, but you don't know how many stages there are unless you fetch that from another call) or search in an area and get the completion status of the results. I guess we should first implement the authenticated search. As I know the labs API quite well I can help if it's wanted.

Quite a lot can be accomplished via the Lab API. However, our official (and my be only temporary) API key extremely limits the methods we can call.

You weren't with us at the time when our key wasn't "restricted" to what it is now. Initially I already had implemented to retrieve the true found state for a user, then all of a sudden the respective method no longer worked.

baiti:
> @baiti Although this is off-topic for this issue. Maybe we should discuss whether we want to implement as @ztNFny offered help as well? Topic for our next dev call?

Good idea. Looking forward to it.

ztNFny:
I'm not talking sbout the current cgeo API key, what I mentioned are limitations of the API itself.
[...]

PR #16968, opened 2025-05-22
#12196 prepare for foundDate

baiti:
This PR is in preparation for providing not only a foundState for lab caches but also a foundDate. I have refactored the code in general as it had a lot or repeated statements so the PR improves the overall logic I believe. At this time the foundDate code is tied to a DEBUG switch which I have set to false so no undesired behaviour should surface.

I think it would be worthwhile to have it in the code base for quality improvement and easier maintenance in the future. I have done tests on the emulator and have not found any problems.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -55,9 +55,13 @@ import org.apache.commons.lang3.StringUtils;
 final class ALApi {
 
     @NonNull
-    private static final String API_HOST = "https://labs-api.geocaching.com/Api/Adventures/";
+    private static final boolean DEBUG = false;
+    private static final String API_HOST = "https://labs-api.geocaching.com/Api/";
+    private static final String ADVENTURES = "Adventures/";
+    private static final String JOURNAL = "GeocacheLogs";
     private static final String CONSUMER_HEADER = "X-Consumer-Key";
     private static final String CONSUMER_KEY = LocalizationUtils.getString(R.string.alc_consumer_key);
+    private static final String AUTHORIZATION = "Bearer XXX"; // access_token must be acquired somehow
 
     private static final String LOCATION = "/Location";
     private static final String LONGITUDE = "Longitude";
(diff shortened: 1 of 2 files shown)


baiti:
@moving-bits: I was wondering what might prevent a merge? Even though it isn't directly fixing an issue, I believe it is a refactoring that improves the code structure and quality.

moving-bits:
Hi @baiti, sorry for not reacting earlier to this PR, I guess we were waiting if something for "set found date" will follow-up and be included in this PR. (BTW: If that's not available through API, but we know the lab's found state, we could use the locally available found dates of its stages and use the youngest of them as found date for the lab, in case of found lab=true.)

To get this PR included into our next beta or release version, please retarget it to `release` branch, as we have already branched off some weeks ago in preparation for the last beta. Thanks for your contribution.

baiti:
> Hi @baiti, sorry for not reacting earlier to this PR, I guess we were waiting if something for "set found date" will follow-up and be included in this PR. (BTW: If that's not available through API, but we know the lab's found state, we could use the locally available found dates of its stages and use the youngest of them as found date for the lab, in case of found lab=true.)

Are the found dates of the stages really locally available? Using the youngest was my plan but I have not been able to find those dates. I know those dates can be obtained via the API but not with our `consumer_key`.

@moving-bits : WRT retargeting to release: I believe I do not have the authority to do this on the GitHub UI.

moving-bits:
> Are the found dates of the stages really locally available?

Not in the database, but I thought you can get them via some API call (condition would be: found lab=true, but still false in db, then do extra call for stages), but...

>I know those dates can be obtained via the API but not with our `consumer_key`.

ok, then this idea is "dead".

> @moving-bits : WRT retargeting to release: I believe I do not have the authority to do this on the GitHub UI.

I can try to do this via Github UI, but this will fail in many cases, so it may happen that this PR is broken afterwards. You would then need to checkout `release` locally, create a new PR branch, and cherry-pick you local version of this PR, then either force-push it to this server branch, or as a new PR.

(BTW: Can you see the "edit" button beside the PR's title? This is the place where you can also change the PR's target - but with the mentioned limitations.)

moving-bits:
@baiti 
I tried switching target of this PR to `release`, but as expected, it then tried to pull in 15 or so other commits, so I reverted to `master` and created a new PR instead.

For this, I created a new branch in my local git environment, based on `release` and then cherry-picked your changes into that new branch. This I did with `git cherry-pick <your commit id>`. The commit id I took from the latest commit on your PR, see some way further up on this page here. From there on, it's just like any other PR.

List of commands:
```
git checkout release
git checkout -b issue_16968
git cherry-pick 467cae5210fcb9fb0fcc99296ff53d366bb280fd
git push -u origin issue_16968
```

Resulting PR is #17046.

baiti:
I am closing this one as it is outdated by now.

PR #16976, opened 2025-05-24
long-tap on search suggestion to prefill it as search term

### LINKED ISSUE #16970: Make search fields remember their text even on failed searches
User request on support (ticket 775305) for 2025.05.16-RC:

"In addition, the new search page is horrible. Please go back to where we have those nice blank spaces. 

In the old search, if there was a problem and you needed to retype it, the information you were searching was still there when you return to the search page. On the new arrangement, you lose everything and have to completely retype. 

This makes search mostly unusable."

ztNFny:
adds a long-tap action on any item in search suggestions/history to use its text as search term
fixes #16970


bekuno:
Please target this to release.

ztNFny:
rebased

PR #16979, opened 2025-05-25
Fix track detection for tracks/route menu

moving-bits:
## Description
This PR fixes an error reported by a user on support (477952).

The user has loaded some tracks on the map. Opening the track/routes bottom sheet listed the tracks, but with the wrong action buttons. Being shown were the action buttons for an individual route, not for a track, and tapping on one of the buttons did not work. - This was caused by wrongly detecting tracks as individual routes.


bekuno:
Retest this please

PR #16981, opened 2025-05-25
keep current caches waypoints visible when enabling live mode

### LINKED ISSUE #16837: Enabling live mode from a cache map hides its waypoints
1. Map a cache with waypoints
2. Enable live mode
-> waypoints are gone, only caches are shown

Vtm and MF

ztNFny:
fixes #16837

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -649,27 +649,30 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
         final GeocacheFilter filter = viewModel.mapType.filterContext.get();
         final Viewport viewport = viewModel.viewport.getValue();
 
-        //should waypoints be displayed at all?
-        final boolean waypointsAreVisible = viewModel.mapType.isSingleCacheView() || (viewModel.caches.readWithResult(viewport::count) < Settings.getWayPointsThreshold());
-        if (!waypointsAreVisible) {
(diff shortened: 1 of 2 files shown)


moving-bits (changes requested):
I had given this a try some time ago, starting with "map type = target_geocode", but failed with the live loaders' complexity. I guess, the other way around, extending "plain mode" by displaying the target (if given), is valid as well.

Two things I've stumbled upon:

Do you want to handle the "set waypoint as target" case as well? You seem to, as you assign `waypointId` in `getPlainMapWithTarget()` (would not be needed for just setting a cache as target), but it does not fully work as expected:

Test case "set waypoint as target":
- Cache having two waypoints, one close by, one far away
- Map set to show 100 waypoints
- Open cache, set far away waypoint as target
- Enable live mode
- Scroll to the other waypoint => it does not show up, even if less than 100 caches are being displayed. It only shows up when showing at least the info sheet for the related cache.
- Same seems to be true for any other cache that has waypoints.

Method `UnifiedMapType.isSingleCacheView()` is no longer used.

moving-bits:
> targeting a single WP shows only that WP, no other WPs of the cache.

That part works as expected.

> Then enabling live adds caches per normal live cache rules and keeps what was visible before - the sole WP.

But it does not show the other caches' waypoints, even if configured to do so (waypoint limit > 0).

ztNFny:
> But it does not show the other caches' waypoints, even if configured to do so (waypoint limit > 0).

That's the case I didn't test as I never want to see other caches WPs ;-) Fixed now.

ztNFny:
failing test is unrelated

Issue #15613, opened 2024-04-15
Problem opening map file - eading magic byte has failed


bekuno:
### Describe your problem!

Logfiles shows the following error:
~~~
04-14 20:44:09.257 30688 30769 E cgeo    : [RxCachedThreadScheduler-1] Problem opening map file 'content://com.android.externalstorage.documents/tree/6363-3661%3AAndroid%2Fmaps%2Fvector/document/6363-3661%3AAndroid%2Fmaps%2Fvector%2FFreizeitkarte_ita%2Bnorth.map.zip.map'
04-14 20:44:09.257 30688 30769 E cgeo    : org.mapsforge.map.reader.header.MapFileException: reading magic byte has failed
04-14 20:44:09.257 30688 30769 E cgeo    : 	at org.mapsforge.map.reader.MapFile.<init>(MapFile.java:293)
04-14 20:44:09.257 30688 30769 E cgeo    : 	at cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider.createMapFile(MapsforgeMapProvider.java:346)
04-14 20:44:09.257 30688 30769 E cgeo    : 	at cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider.readAttributionFromMapFileIfValid(MapsforgeMapProvider.java:316)
04-14 20:44:09.257 30688 30769 E cgeo    : 	at cgeo.geocaching.maps.mapsforge.MapsforgeMapProvider.getAttributionIfValidFor(MapsforgeMapProvider.java:298)
~~~

The download/copy was not successful. You see it on the name `Freizeitkarte_ita_north.map.zip.map` (doubled `map`).
The file size is 0 byte, so of cause no magic byte could not read.

 

### How to reproduce?

Don´t know, when the process was interupted.
You can easy create a 0Byte file with name ending `.map` for testing.

### Actual result after these steps?

The file is skipped, but only after trying to read them. 

### Expected result after these steps?

If the file size is null, the file should be skipped or better deleted, if possible.

### Reproducible

Yes

### c:geo Version

Nigthly from master

PR #16989, opened 2025-05-28
Add 0 byte file size check to map/theme downloaders (fix #16988)

### LINKED ISSUE #16988: Add safeguard to downloader: Do not offer 0 byte files for download
As the Elevate_Winter.zip theme file download is currently broken (0 byte file on server), a follow-up problem surfaced: The changed file still appeared as an update (which is technically correct due to newer file timestamp), so updating that file erases the existing one with a broken one. c:geo downloaders should not offer 0 byte files as updates.

moving-bits:


moving-bits:
Failing test `DownloaderTest.testOpenAndroMapsThemes` is to be expected, as the provided download is currently broken (0 byte file)

moving-bits:
> Is the Issue in #15613 also conered with this PR?

No, that's not covered. Only server-side problems (0 byte files) are "handled" by not listing such files in the downloader. 
If a download fails and leaves a 0 byte file locally for some reason (should not happen anymore, as failed downloads should delete incomplete files), this will still lead to errors as described in the linked issue.

bekuno:
retest this please

PR #16991, opened 2025-05-28
UnifiedMap: Fix compact icon mode auto switching (fix #16966)

### LINKED ISSUE #16966: Auto compact icon mode not switching back to large icons
Reported by user on support & reproducible for me with current beta:

- set map to "compact icon mode = auto"
- open map on a place showing only a few caches => large icons being shown
- zoom out until map switches to compact icon mode => works
- zoom in until map should switch back to large icons => it does not work

moving-bits:


moving-bits:
retest this please

PR #16935, opened 2025-05-11
Fix #16625: Move upload entries and update labels

kennygchen:
## Description
- Fixes a UI inconsistency by moving "Upload modified coordinates" and "Upload coordinates for all caches" into the "Export" submenu.
- Renames the submenu to "Export / Upload" to better reflect its contents.
- Updates string labels of existing export actions to begin with "Export" for clarity and consistency.
- Adds icons to the new upload entries to match the visual style of other items.

## Related issues
Fixes #16625

## Additional context
image
image

PR #16994, opened 2025-05-29
Fix strings

moving-bits:
## Description
Renaming menu items in #16935 had the side-effect of changing dialog titles etc. as well, as some strings are used in several places. This PR untangles those strings.


bekuno:
Retest this please

Issue #431, opened 2011-09-13
Missing direction-arrow in detail-view


SammysHP:
Today I missed the little direction-arrow at the distance-line in the detail-view.

moving-bits:
Fixed with #17017

PR #17017, opened 2025-06-07
#431: Add CompassMiniView in cache detail view

kennygchen:
## Description
- Replace the static compass icon in the action bar of `CacheDetailActivity` with a dynamic, rotating `CompassMiniView` with the same behavior seen in the cache list view. The click and long click actions on the compass icon still work, triggering default and secondary navigation as before.

- Add a user setting: "Show direction in cache detail view"
  The user can toggle whether the dynamic compass arrow appears in the detail view. If disabled, the original compass icon will appear instead. It is enabled by default.

## Related issues
#431 

## Additional context
AISelect_20250607_113400_cgeo
Screenshot_20250607_113528_cgeo (1)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -72,6 +72,7 @@ import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.extension.OneTimeDialogs;
 import cgeo.geocaching.ui.AnchorAwareLinkMovementMethod;
 import cgeo.geocaching.ui.CacheDetailsCreator;
+import cgeo.geocaching.ui.CompassMiniView;
 import cgeo.geocaching.ui.DecryptTextClickListener;
 import cgeo.geocaching.ui.FastScrollListener;
 import cgeo.geocaching.ui.ImageGalleryView;
@@ -88,6 +89,7 @@ import cgeo.geocaching.ui.dialog.NewCoordinateInputDialog;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.ui.recyclerview.RecyclerViewProvider;
 import cgeo.geocaching.utils.AndroidRxUtils;
+import cgeo.geocaching.utils.AngleUtils;
 import cgeo.geocaching.utils.CacheUtils;
 import cgeo.geocaching.utils.CalendarUtils;
 import cgeo.geocaching.utils.CheckerUtils;
@@ -223,6 +225,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
     private GeoDirHandler locationUpdater;
 
     private MenuItem menuItemToggleWaypointsFromNote = null;
+
+    private CompassMiniView compassMiniView;
 
     /**
      * If another activity is called and can modify the data of this activity, we refresh it on resume.
@@ -449,7 +453,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             start = initial;
         }
         if (start) {
-            geoDataDisposable.add(locationUpdater.start(GeoDirHandler.UPDATE_GEODATA));
+            geoDataDisposable.add(locationUpdater.start(GeoDirHandler.UPDATE_GEODIR));
         }
     }
 
@@ -717,6 +721,9 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
         if (cache != null) {
             // top level menu items
+            if (getUseLiveCompassInNavigationAction() && cache.getCoords() != null) {
+                initCompassMiniView();
+            }
 
             final MenuItem ttsMenuItem = menu.findItem(R.id.menu_tts_toggle);
             ttsMenuItem.setVisible(!cache.isGotoHistoryUDC());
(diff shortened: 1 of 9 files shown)


bekuno:
ok to test

bekuno:
I'm not a fan of this dynamically changing eye-catcher.

ztNFny:
I'm not sure what I think about this. Considering that the icon might or might not be compass it's kinda strange. It also feels quite distracting.
If we merge then I'd vote for default=disabled

SammysHP:
Very creative way of doing it and I like that you can see it from every page, for example while reading logs or the description.

How does it work with the power savings mode? IIRC it disables location services if you aren't on the details page, but I could be wrong.

kennygchen:
I totally understand the concerns about the compass being a distraction but I think that it could be useful for some people.
That's why I added a setting so users have the choice.
Making it disabled by default might be a good idea.

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -1047,6 +1048,10 @@ public static boolean isLiveList() {
         return getBoolean(R.string.pref_livelist, true);
     }
 
+    public static boolean isLiveDetail() {
```
eddiemuc: NOt a good name. People will think of "live map" which is something completely different".

How about something like "useLiveCompassInNavigationAction"?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -224,6 +226,11 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     private MenuItem menuItemToggleWaypointsFromNote = null;
 
+    private CompassMiniView compassMiniView;
+    private float azimuth = 0;
```
eddiemuc: Why is azimuth needed here as member variable?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -224,6 +226,11 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
...
+
+    private final boolean isLiveDetail = Settings.isLiveDetail();
```
eddiemuc: Don't cache settings in member variables. Users might change it while the activity is still alive. 
If you need to access it often, add a method "getXYYZ"  and use it.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -448,7 +455,9 @@ private void startOrStopGeoDataListener(final boolean initial) {
         } else {
             start = initial;
         }
-        if (start) {
+        if (start && isLiveDetail) {
+            geoDataDisposable.add(locationUpdater.start(GeoDirHandler.UPDATE_GEODATA | GeoDirHandler.UPDATE_DIRECTION));
+        } else {
```
eddiemuc: This is wrong. The ELSE path would execute for start=false which is not intended

Instead, use just GeoDirHandler.UPDATE_GEODIR

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -883,7 +904,44 @@ public void updateGeoData(final GeoData geo) {
...
+                if (activity.compassMiniView != null && activity.compassMiniView == activity.findViewById(R.id.compass_action)) {
+                    activity.setActualCoordinates(geo.getCoords());
+                } else if (activity.cache != null && activity.cache.getCoords() != null) {
+                    activity.initCompassMiniView();
+                }
+            }
+        }
+
+        @Override
+        public void updateDirection(final float direction) {
```
eddiemuc: Remove this method. Instead above, replace method `updateGeoData `with method `public void updateGeoDir(@NonNull final GeoData newGeo, final float newDirection) ` The new one can handle direction and geoposition changes in one place instead of two.

main/src/main/java/cgeo/geocaching/ui/CompassMiniView.java
```diff
@@ -88,7 +88,7 @@ public void setTargetCoords(final Geopoint point) {
         targetCoords = point;
     }
 
-    void updateAzimuth(final float azimuth) {
+    public void updateAzimuth(final float azimuth) {
```
eddiemuc, line 91: You might want to take this opportunity to rewrite CompoassMiniView in a self-contained way: maybe this class should by itself register and unregister from geoposition and direction updates and not rely on being called for updates by the outside world?

main/src/main/java/cgeo/geocaching/ui/NavigationActionProvider.java
```diff
@@ -43,9 +45,14 @@ public View onCreateActionView() {
         if (navigationSource != null) {
 
             final LayoutInflater layoutInflater = LayoutInflater.from(context);
-            view = layoutInflater.inflate(R.layout.navigation_action, null);
-
-            final View navItem = view.findViewById(R.id.default_navigation_action);
+            View navItem = null;
+            if (Settings.isLiveDetail() && navigationSource instanceof CacheDetailActivity) {
```
eddiemuc: Why restrict to cachedetailsactivity?

eddiemuc (changes requested):
Thanks for the contribution. I found one bug and have some other comments though

main/src/main/java/cgeo/geocaching/ui/NavigationActionProvider.java
kennygchen: The reason I restrict it to CacheDetailActivity is because in the map, the action bar already has the same icon, which I think could confuse the user. And I think that the compass doesn’t add much value there since the user is already viewing the map and can see their orientation directly.

Screenshot_20250608_235046_cgeo (1)

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -314,6 +314,7 @@ private static void migrateSettings() {
             e.putBoolean(getKey(R.string.pref_maptrail), prefsV0.getInt(getKey(R.string.pref_maptrail), 1) != 0);
             e.putInt(getKey(R.string.pref_lastmapzoom), prefsV0.getInt(getKey(R.string.pref_lastmapzoom), 14));
             e.putBoolean(getKey(R.string.pref_livelist), prefsV0.getInt(getKey(R.string.pref_livelist), 1) != 0);
+            e.putBoolean(getKey(R.string.pref_live_compass_in_navigation_action), prefsV0.getBoolean(getKey(R.string.pref_live_compass_in_navigation_action), false));
```
moving-bits: no need for migration here, as this is a new setting

moving-bits:
Let's try it in the nightlies.

Thanks @kennygchen for your contribution, very welcomed!

PR #17018, opened 2025-06-07
Add some missing OC cache types

### LINKED ISSUE #16992: Opencaching unknown and owncache different icons
### Tell us your idea!

At least from OCRO (https://www.opencaching.ro) I have discovered the following situation: 
- cache_type=quiz/puzzle (site) = cache_type=mystery (c:geo)
This is ok, it is a reasonable equivalent with GC types.
The icon is the same as for GC mystery type (blue BG and question mark)

- cache_type=unknown (site) = cache_type=unknown (c:geo)
This is a catch-all type for all caches that do not specifically fit the existing defined types.
The icon is grey BG with 3 points horizontally. 
Although IMO this might not be the best representation, at the moment I can't think of a better suggestion.

- cache_type=owncache (site) = cache_type=unknown (c:geo)
The icon is grey BG with 3 points horizontally. 
First of all I request that this be implemented as specific cache_type. 
Second, I'd like to suggest a different visual representation. I can create a SVG, but I can't seem to find any of the other types as SVG in the code repository to start from as a template. **Please help.**

Moving cache and podcast cache I am still investigating. 

_No response_

### Is there an alternative?

_No response_

### Additional Information

_No response_

bekuno:
Define the open Caches at OCPL as follow:

~~~
        'Moving'   =>  CacheType.LOCATIONLESS
        'Other'    =>  CacheType.UNKNOWN
        'Podcast'  =>  CacheType.VIRTUAL
        'Own'      =>  CacheType.LOCATIONLESS
~~~
fixes #16992
see https://github.com/opencaching/opencaching-pl/issues/2520

@andrixnet OK?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -1193,6 +1193,18 @@ final class OkapiClient {
         if ("Webcam".equalsIgnoreCase(cacheType)) {
             return CacheType.WEBCAM;
         }
+        if ("Moving".equalsIgnoreCase(cacheType)) {
+            return CacheType.LOCATIONLESS;
+        }
+        if ("Other".equalsIgnoreCase(cacheType)) {
+            return CacheType.UNKNOWN;
+        }
+        if ("Podcast".equalsIgnoreCase(cacheType)) {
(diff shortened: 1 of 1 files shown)


bekuno:
Sample:
Web:
grafik
c:geo:
grafik

PR #17019, opened 2025-06-07
fix #13868:  Show only empty variables

### LINKED ISSUE #13868: Option to show only empty variables in variable view
### Tell us your idea!

Sometimes it would be helpful to see only the variables for which no value has been entered yet.
Otherwise, with long variable lists, you have to scroll a lot and often overlook one that has not yet been assigned.

Option to show only  empty variables in variable view (like the option to hide visited waypoints in waypoint view)

  

### Is there an alternative?

leave it as it is

### Additional Information

_No response_

murggel:
## Description

Option to hide completed variables in variable-tab

Improve time of sorting the variables

## Related issues

fix #13868

## Improving the header 
=> new PR #17229

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/VariableListView.java
+++ b/main/src/main/java/cgeo/geocaching/ui/VariableListView.java
@@ -407,14 +407,12 @@ public class VariableListView extends LinearLayout {
         }
 
         public void sortVariables(final Comparator<String> comparator) {
-            if (comparator == null) {
+            if (comparator == null || variables == null) {
                 return;
             }
-            sortItems((v1, v2) -> comparator.compare(v1.getVar(), v2.getVar()));
-            if (variables != null) {
-                variables.sortVariables(comparator);
-            }
-            callCallback();
+
+            // don't reinitialize from original list, hence, only sortVariables and not sortItems
+            variables.sortVariables(comparator);
         }
 
         public void selectVariableName(final String oldName, final Action2<String, String> callback) {
--- a/main/src/main/java/cgeo/geocaching/ui/VariableListView.java
+++ b/main/src/main/java/cgeo/geocaching/ui/VariableListView.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.ui;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.databinding.VariableListViewBinding;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.ui.recyclerview.ManagedListAdapter;
 import cgeo.geocaching.utils.Log;
@@ -248,11 +249,10 @@ public class VariableListView extends LinearLayout {
                     return;
                 }
                 //disable filter
-                filterEnabled = false;
-                this.visibleVariables.clear();
-                this.setFilter(d -> true, true);
-
+                setVisibleVariables();
+                return;
             }
+
             //this is a costly operation
             //Thus check whether new and old set contains same elements (e.g. no change)
             if (filterEnabled && newVisibleVariables.size() == this.visibleVariables.size() && this.visibleVariables.containsAll(newVisibleVariables)) {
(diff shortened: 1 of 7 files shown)


murggel:
@ztNFny Do you have any other ideas, how to improve the header of the variables page, to make room for the checkbox?

I tried to change the 2 buttons for creating new variables into full buttons like in the waypoint page, but rejected that, because it would need even more space. So I came to the above variants

ztNFny:
you could check the bookmarks/PQ dialog where we have a toggle in toolbar. but that probably won't work here.
making the add variable buttons full-width seems counter-productive, i wouldn't do that.
Re wording, what's a "completed" variable? Maybe "filled in" or "valued" variables?

murggel:
> Re wording, what's a "completed" variable? Maybe "filled in" or "valued" variables?

Those variables are hidden, which have a dedicated value or whose formula / result could already be calculated.
Only those variables are shown,  which result is not determined yet.

So "completed" means "fully calculated'.
"Filled" can be interpreted as "formula entered, but not all variables have a value"

ztNFny:
if it has a formula and is not requiring any user input I'd consider it "to be hidden"

murggel:
> if it has a formula and is not requiring any user input I'd consider it "to be hidden"

I would not hide those, which are not completely calculated, to see what effect has the input of a specific variable

murggel:
So probably a two way or three way option as in the status filter - or is that too much?






Or than as option which to show...

moving-bits:
I'd stay with current interpretation ("fully calculated y/n") and wording, and not add additional UI complexity for this specific case.

main/src/main/java/cgeo/geocaching/ui/VariableListView.java
```diff
@@ -398,11 +410,9 @@ public void sortVariables(final Comparator<String> comparator) {
             if (comparator == null) {
                 return;
             }
-            sortItems((v1, v2) -> comparator.compare(v1.getVar(), v2.getVar()));
```
murggel, line 401: `sortItems` will reset the items to the original list, which we don't want.

main/src/main/java/cgeo/geocaching/ui/VariableListView.java
```diff
@@ -398,11 +410,9 @@ public void sortVariables(final Comparator<String> comparator) {
...
             if (variables != null) {
                 variables.sortVariables(comparator);
             }
-            callCallback();
```
murggel, line 405: `callCallback` is not necessary, because the variables didn't change

main/src/main/java/cgeo/geocaching/utils/formulas/VariableList.java
```diff
@@ -177,7 +177,6 @@ public int removeVariable(@NonNull final String var) {
 
     public void sortVariables(final Comparator<String> comp) {
         Collections.sort(variableList, comp);
-        wasModified = true;
```
murggel, line 180: Don't set `wasModified`, because the variables didn't change

murggel:
retest this please

murggel:
With current implementation a completed variable stays visible at the moment it gets completed. You have to trigger a refresh, whether on changing the tab or clicking on tidy up to update the view state. Probably also, when adding / removing a variable.

I am not sure, whether to hide the variable, if you exit the edit-field.

murggel:
I split the improvement of the header-layout to a new PR #17229

ztNFny:
> I split the improvement of the header-layout to a new PR #17229

How about using the same UI as https://github.com/cgeo/cgeo/issues/17185#issuecomment-3124248129

murggel:
> > I split the improvement of the header-layout to a new PR #17229
> 
> How about using the same UI as https://github.com/cgeo/cgeo/issues/17185#issuecomment-3124248129

Didn't see that, have to look into it or do you mean something like this (see my comment above? Probably in a first version only one option

> 
> 

But then we should also change it in the waypoint tab

murggel:
I guess it takes a little more space than the checkbox

image image



Waypoint tab
See https://github.com/cgeo/cgeo/pull/17229#issuecomment-3146387612

image image

murggel:
Will change it to "Completed" / "Visited" only

eddiemuc:
Is this PR finished/ can be merged?

murggel:
> Is this PR finished/ can be merged?

branch updated to current master - ready from my side...

PR #17020, opened 2025-06-07
Fix  #17007:  Move "recently used" lists in list selection dialog to the top

### LINKED ISSUE #17007: Highlight "recently used" lists in list selection dialog in some way
User request on support (ticket 970010):

When using the list selection dialog you have the option to use the "recently used" selection. Problem is that you have no visual indication of which lists those were. User asks for some visual indicator or other form of support.

Ideas mentioned are:
- (~label the button with list name~) => won't work for long list names or multiple lists
- have some preview window listing the new list names on hitting the "recently used" button, asking for confirmation before overwriting
- show some visual indicator beside the list names that are part of "recently used"

Any ideas?

murggel:
## Description

Moving the lists from the "last selection" to the "top group", while current lists remain unchecked as well in the "top group" to easily "recheck" them.
(Un)checking list-entries does not move the lists to the / from the "top group"

Open "Pick the lists" 

Select "Last selection": add checked last selected list (keep original selection unchecked at top)

Change selection does not move 

Select "Last selection" again (no more moving)

## Related issues

fix #17007

### DIFF
--- a/main/src/main/java/cgeo/geocaching/list/StoredList.java
+++ b/main/src/main/java/cgeo/geocaching/list/StoredList.java
@@ -41,6 +41,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import com.google.android.material.textfield.TextInputEditText;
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

PR #17021, opened 2025-06-08
Add a photo viewer in "edit image" under log pages

### LINKED ISSUE #16018: Show photo viewer in "edit image" under log page
### Tell us your idea!

Currently when accessing the "Edit image" dialog from the log page, the user can perform various actions regarding the image itself, add caption/description text. However, tap on the image yields nothing. 

Screenshot_20240825-195725_cgeo

Given the fact that choosing an image from the library will always leave you dealing with thumbnails, one will have difficulties discering differences between similar images based only on thumbnails. Since the chooser only works with thumbnails, I propose that on the "Edit image" dialog, tap on the image itself to open it in the internal image viewer the same way as tapping on an image in the logs or images tabs when viewing a cache does.
That way the user can make better judgements about the chosen image and can do this without a laborious process of switching to another application (ie image gallery + viewer), looking up the image and viewing it larger then a thumbnail.

### Is there an alternative?

_No response_

### Additional Information

_No response_

wsw-stack:
## Description

Added a new button at the bottom of the edit page:
image
And then they'll land on the image details page (a newly added layout):
image

## Related issues

fix #16018 

## Additional context

I've setup a basic new layout for showing the image details, is it a better idea to reuse the existing `imageview_activity.xml`? Or adding a new layout for this feature is better?

I'm a beginner in Android development & PR contribution! Thanks for understanding and the valuable feedback!

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -300,6 +300,8 @@ android {
 }
 
 dependencies {
+    implementation 'androidx.activity:activity:1.10.1'
+
     // desugaring of Java 8 APIs
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5'
 
--- a/main/src/main/java/cgeo/geocaching/ImageDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/ImageDetailActivity.java
@@ -1,0 +1,147 @@
+package cgeo.geocaching;
+
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.os.Bundle;
+import android.view.GestureDetector;
+import android.view.MenuItem;
+import android.view.MotionEvent;
+import android.view.ViewGroup;
+import android.widget.ImageView;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+import com.igreenwood.loupe.Loupe;
+
+public class ImageDetailActivity extends AppCompatActivity {
+
+    private ImageView imageView;
+    private ViewGroup container;
+    private Loupe loupe;
+
+    @Override
+    protected void onCreate(@Nullable final Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        // Inflate the layout
+        setContentView(R.layout.imagedetail_activity);
+
+        // Find views
+        container = findViewById(R.id.container);
+        imageView = findViewById(R.id.image_detail_view);
+
+        // Attach Loupe to enable pinch-zoom, pan, double-tap and pull-to-dismiss
+        loupe = new Loupe(imageView, container);
+
+        // Optional: add pull-to-dismiss listener
+        loupe.setOnViewTranslateListener(new Loupe.OnViewTranslateListener() {
+            @Override
+            public void onStart(@NonNull final ImageView view) {
+                // gesture start
+            }
+
+            @Override
+            public void onViewTranslate(@NonNull final ImageView view, final float amount) {
+                // amount: drag ratio
+            }
+
+            @Override
+            public void onRestore(@NonNull final ImageView view) {
+                // restore if threshold not reached
+            }
+
+            @Override
+            public void onDismiss(@NonNull final ImageView view) {
+                // dismiss activity when pulled beyond threshold
+                finish();
+            }
+        });
+
+        // Optional: detect single taps by layering a GestureDetector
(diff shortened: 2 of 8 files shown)


SammysHP:
This PR includes another, unrelated change.

What happens when you tap on the image? If this is currently unused, I suggest to open the image when tapping on it and remove the button.

Also please don't change unrelated files.

You implemented your own image viewer activity, but we should already have something like that for the log image viewer.

wsw-stack:
> This PR includes another, unrelated change.
> 
> What happens when you tap on the image? If this is currently unused, I suggest to open the image when tapping on it and remove the button.
> 
> Also please don't change unrelated files.
> 
> You implemented your own image viewer activity, but we should already have something like that for the log image viewer.

I just added a button at the bottom, is that better or making the image clickable (and showing image details) is better?
Thanks for your feedback! I'll take that advice and try to change that.

main/build.gradle
```diff
@@ -300,6 +300,8 @@ android {
 }
 
 dependencies {
+    implementation 'androidx.activity:activity:1.10.1'
```
eddiemuc, line 303: This dependency should not be needed for the task at hand, please remove

main/src/main/java/cgeo/geocaching/ImageDetailActivity.java
```diff
@@ -0,0 +1,147 @@
...
+import androidx.appcompat.app.AppCompatActivity;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+import com.igreenwood.loupe.Loupe;
+
+public class ImageDetailActivity extends AppCompatActivity {
```
eddiemuc, line 23: Don't create a new activity.

* If you want to jump to a new activity for image display, then use  `cgeo.geocaching.ImageViewActivity`. tryx use method `openImageView` from it to open new activity
* If instead you want to make the image zoomable inside the Edit-Activity, take a look at method `cgeo.geocaching.utils.ImageUtils.createZoomableImageView`

main/src/main/java/cgeo/geocaching/ImageEditActivity.java
```diff
@@ -140,6 +141,21 @@ public void onCreate(final Bundle savedInstanceState) {
         }
 
         loadImage();
+
+        // check the image details
+        binding.logImgViewBtn.setOnClickListener(view -> {
+            if (image == null) {
+                Toast.makeText(view.getContext(),
```
eddiemuc, line 148: I assume you might silently ignore the (unlikely) situation of a null-image. (if you want you may write a log message). But IF you insist on a toast, then it's text must be internationalized and not hardcoded.

main/src/main/java/cgeo/geocaching/ImageEditActivity.java
```diff
@@ -140,6 +141,21 @@ public void onCreate(final Bundle savedInstanceState) {
         }
 
         loadImage();
+
+        // check the image details
+        binding.logImgViewBtn.setOnClickListener(view -> {
```
eddiemuc, line 146: Why have an extra button and not open the image viewer when clicking on the image itself?

main/src/main/java/cgeo/geocaching/ImageEditActivity.java
```diff
@@ -140,6 +141,21 @@ public void onCreate(final Bundle savedInstanceState) {
...
+            if (image == null) {
+                Toast.makeText(view.getContext(),
+                                "Image not available",
+                                Toast.LENGTH_SHORT)
+                        .show();
+            } else {
+                final Intent imageDetailIntent = new Intent(getApplicationContext(), ImageDetailActivity.class);
```
eddiemuc, line 153: Use `ImageViewAcvtivity` instead

main/src/test/java/cgeo/geocaching/location/GeopointConverterTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+public class GeopointConverterTest {
```
eddiemuc, line 13: I believe this is part of another PR? It was probably included here by accident. Please remove from this PR

eddiemuc (commented):
Thanks for the contribution. I have some comments however

bekuno (changes requested):
Please review changes on `AndroidManifest.xml`.
Is there really something to change?

moving-bits:
> Please review changes on `AndroidManifest.xml`. Is there really something to change?

Most likely not, at least not involving such a large reformatting. Setting WIP label for now.

PR #17010, opened 2025-06-04
Fix #16818: Rearrange Appearance and CacheDetail and add settings for log line limit

murggel:
## Description

Add settings to configure maximum number of lines to be shown for a log

## Related issues

Fix #16818

## Additional context

image

Issue #16818, opened 2025-04-06
Expand logs by default / all at one


SammysHP:
Kind of a follow-up to #16789.

When I'm reading logs, usually I'd like to find something, like a hint. Now that logs are collapsed, I have to open every single log, skim over it, move on to the next one. All and all, collapsing the logs doesn't help my workflow and I'd like to disable it.

moving-bits:
Which variant would you prefer - a setting to enable/disable collapsing by default, or a button in log view to "expand all logs"?

ztNFny:
Before adding another UI element can we please have some concept for where to put it, how often it's used, ... We're yet again complicating the UI again - which the original feature already did.

Lomanic:
To be honest, the UI for expanding longer _comments_ in NewPipe, another OSS app I use, is better and more standard in my opinion, it's just a single tap anywhere on a long comment to expand/collapse it. cgeo already does something when a comment is tapped, which is opening a menu. I think cgeo should align with this more intuitive UI, and the popup menu should be opened with a long press, and the collapse/expand action be triggered by a single tap.

This has bugged me since the long press on logs, which made a menu appear on the top of the screen, has been changed to that quick tap pop up menu, see https://github.com/cgeo/cgeo/issues/9151#issuecomment-707066551.

murggel:
I like the idea for the setting for MAX_LINES (A value of 0 could then perhaps mean Disable log collapsing) and I would work on that...

Question would be, in which category to put this setting (Cache Details / Logging)?

murggel:
"Plain Logs" sound good  maybe I create a new section "Logs" in "Appearance" (and move "Show Friends'/Own" to the new section...)

moving-bits:
Settings => Cache Details would be a good place, IMHO, and I like the idea of setting MAX_LINES value (with 0=don't collapse by default).

Maybe Settings => Appearance => Plain Logs should be moved to Settings => Cache Details as well, and those three together form a new sub category "Logs".

murggel:
> Settings => Cache Details would be a good place, IMHO, and I like the idea of setting MAX_LINES value (with 0=don't collapse by default).
> 
> Maybe Settings => Appearance => Plain Logs should be moved to Settings => Cache Details as well, and those three together form a new sub category "Logs".

I will provide a screenshot for that in the next days...

I will introduce sub categories - these don't exist currently in cache-detail-settings

murggel:
Cache-Details Settings with new introduced headings:

[...]

PR #17031, opened 2025-06-09
Use log-line-limit-default

murggel:
Use log_line_limit_default instead of list_load_limit_default

related to PR #17010 / related to issue #16818


bekuno (approved):
I had only an eye on CI.
I think you know what is right here.

Issue #17026, opened 2025-06-09
Stuck in crash after certain GPX track has been loaded


Lineflyer:
Using 2025.06.09-NB I prepared for a bike tour.

After loading the GPX track it was not shown on the map.
Trying to remove the file c:geo crashes when trying to open the "Routing/Tracks" menu on map, thus I am stuck with this problem.

(Was only to be solved by removing the entry in `cg_extension`-table manually)

Problematic file:
https://cloud.droescher.eu/index.php/s/BGLYZ5W9tdkTrCE

Log:
```
2025-06-09 12:16:14.843  4477-4477  Andro...time cgeo.geocaching  E  FATAL EXCEPTION: main
                                                                     Process: cgeo.geocaching, PID: 4477
                                                                     java.lang.NullPointerException: Attempt to invoke interface method 'boolean cgeo.geocaching.models.geoitem.IGeoItemSupplier.isHidden()' on a null object reference
                                                                     	at cgeo.geocaching.maps.RouteTrackUtils.configureContextMenu(SourceFile:168)
                                                                     	at cgeo.geocaching.maps.RouteTrackUtils.$r8$lambda$64DwF8IwwTAZeEsbbYyNL8E01BU(SourceFile:334)
                                                                     	at cgeo.geocaching.maps.RouteTrackUtils$$ExternalSyntheticLambda13.call(SourceFile:0)
                                                                     	at cgeo.geocaching.maps.Tracks.traverse(SourceFile:92)
                                                                     	at cgeo.geocaching.maps.RouteTrackUtils.updateDialogTracks(SourceFile:330)
                                                                     	at cgeo.geocaching.maps.RouteTrackUtils.showPopup(SourceFile:132)
                                                                     	at cgeo.geocaching.unifiedmap.UnifiedMapActivity.$r8$lambda$KK2_7TMnUtZhvzm9M8qmFN-Q2uI(SourceFile:361)
                                                                     	at cgeo.geocaching.unifiedmap.UnifiedMapActivity$$ExternalSyntheticLambda23.onClick(SourceFile:0)
                                                                     	at android.view.View.performClick(View.java:8464)
                                                                     	at android.view.View.performClickInternal(View.java:8441)
                                                                     	at android.view.View.-$$Nest$mperformClickInternal(Unknown Source:0)
                                                                     	at android.view.View$PerformClick.run(View.java:32966)
                                                                     	at android.os.Handler.handleCallback(Handler.java:959)
[...]

PR #17032, opened 2025-06-09
Make parsing errors more visible in log files (rel. to #17026)

moving-bits:
Log errors / exceptions during GPX parsing as errors instead of warnings to make them available in standard logfiles with default log level.


bekuno:
Retest this please

moving-bits:
Failing Test is unrelated

PR #17049, opened 2025-06-14
Set targetSDK to 35

moving-bits:
## Description
As discussed in our last dev meeting: Let's try targetSDK 35 early in the current dev cycle, as some breaking changes are to be expected from reading the docs.

~One is already very visible directly after start: The new edge-to-edge mode hides UI elements under the topbar (eg: in settings), therefore setting WIP label for now.~ (that seems to be fixed as of Jul 5th)

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -33,7 +33,7 @@ android {
 
     defaultConfig {
         minSdk 21
-        targetSdk 34
+        targetSdk 35
 
         versionName versionNameFromDate()
         versionCode versionCodeFromDate(0)
--- a/main/src/main/res/values-v35/themes.xml
+++ b/main/src/main/res/values-v35/themes.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="cgeo_base" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
+        <item name="android:fitsSystemWindows">true</item>
+        <item name="android:enforceNavigationBarContrast">true</item>
+        <item name="android:enforceStatusBarContrast">true</item>
+    </style>
+
+</resources>
--- a/main/src/main/res/values/themes.xml
+++ b/main/src/main/res/values/themes.xml
@@ -2,8 +2,8 @@
 <resources>
 
     <!-- theme for (nearly) all activities -->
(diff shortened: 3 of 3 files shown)


moving-bits:
For details on relevant changes have a look at the following docs:
- Android Studio: Tools => Android SDK Upgrade Assistant => Select upgrade path API 34 to 35
- Changes in behavior: https://developer.android.com/about/versions/15/behavior-changes-15

Some things will become visible only when using a device running Android 15.

Edge-to-edge enforcement is quite picky! Here's a collection of some docs for it:
- https://medium.com/androiddevelopers/insets-handling-tips-for-android-15s-edge-to-edge-enforcement-872774e8839b
- https://developer.android.com/develop/ui/views/layout/edge-to-edge#java

There is a setting to temporarily disable edge-to-edge enforcement, but that is supposed to not stay for long, so it could only be a last resolve if we don't have the final adaptions in place before end of August 2025.

moving-bits:
As discussed in today's dev meeting: Let's get this into the nightlies to get feedback from using it on real devices

SammysHP:
While I understand the benefits of early real-life feedback, I wonder if it makes sense to revert this change temporarily, now that we know some open issues. The layout changes make it hard to use the nightly, so there's a higher chance testers revert back to the release and other issues stay hidden for longer time. We could then restore this commit together with the fixes for the problems discovered so far.

moving-bits:
This time it's not that early feedback actually - Play Store requirements enforce usage of targetAPI 35 by end of August, so we are pretty late already... We could revert the change, sure, but my fear would be that no one will be working on those issues then anytime soon. I would prefer if we could join forces and try to fix both
- a) the issues upcoming in current release (we receive quite a few tickets on support currently, will start creating dev tickets soon) and 
- b) the issues resulting from the targetSDK change.

SammysHP:
These are valid reasons. I thought there's more time until the Play Store enforces the new version.

moving-bits:
I think I have read something about the possibility to extend this period until November under certain circumstances, but I can't find the option to apply to that for c:geo.

Since a few year, Play Store API enforcement is always end of August for the API level from the year before, so as Android 16 got published this year (= API 36), Play Store enforces API level 35 by end of August this year.

Issue #11023, opened 2021-06-27
CI/Jenkins: make logcat output accessible for Test runs


eddiemuc:
Today, when a build is executed on Jenkins and a test fails, we only get the exception reported which caused the test failure.

Most often it is really hard to analyze the root cause of a test failure using this exception. I often end up bringing "dummy" exception throws into the code and let it rerun on CI just to analyze such a situation. I would like to avoid this.

It would be extremely useful to have also the logcat output available which was produced during the failing test run.

As a starting point, it would be enough to simply have a capture of whole logcat output (debug level) available for whole test run. Assumption is that this is easier to produce than a logcat output per test.

kumy:
@eddiemuc I've added logcat capture to the GitHub workflow. The logcat will be attached to test results artifacts [and in case of test failure it will post the logcat as a PR comment - but this feature is not enabled]

like in https://github.com/cgeo/cgeo/actions/runs/15758528656

![Image](https://github.com/user-attachments/assets/c906edab-77c1-4b17-b11e-311efa9d3034)

PR #17068, opened 2025-06-19
Ci watchdog

kumy:
## Description

- Capture logcat output #11023
  - logfile will be attached to builds as artifacts
  - in case of test failure, the logcat should be posted on the PR (feature disabled, feel free to able it or remove it as needed)
- Use a mutex service to lock the build instead of relying on branches
  - Should help implementing the lock on Jenkins side
  - A further step could be to surround individual (or block) of Test Cases that are known to need the lock - this way we can remove the lock mechanism on the whole tests side which should speedup our time to test

## Related issues

- #11023
- #15608


kumy:
please retest this

kumy:
please retest this

Issue #17070, opened 2025-06-19
Rename History menu item in German


JakeDot:
The menu button for the cache History is named inconsistently with the opened List view. (Fundhistorie vs. Verlauf).

Maybe we could change the string?

PR #17072, opened 2025-06-19
cgeo/cgeo#17070 Fix German Translation Fundhistorie / Verlauf

JakeDot:
https://github.com/cgeo/cgeo/issues/17070


moving-bits:
Thanks for your contribution, but this is not the right place to rename strings. We are using Crowdin for all of our translations, and any changes you do here, will be overwritten on next sync. Please apply your changes to https://crowdin.com/translate/cgeo/all/en-de

JakeDot:
What

PR #17074, opened 2025-06-20
CI fixes

kumy:
## Description

- ci: Check websites status just before running tests
- ci: Unlock only from jobs that locked it
- ci: Fix timeout value
- ci: let watchdog tests run independently

## Related issues

#15608


kumy:
please retest this

PR #17090, opened 2025-06-26
Fix icon for custom waypoints of linear labs (fix #17089)

### LINKED ISSUE #17089: Additional WPs for sequential ALCs have wrong icon
### Describe your problem!

I noticed today while adding a custom parking waypoint to an sequential adventure lab cache, that it will show as icon "1" instead of icon "P".

I guess the code generating the 1-5 waypoint enumeration is not expecting any other waypoints to be existing?!

### How to reproduce?

- Save adventure lab cache with play type=sequential 
- Map the adventure lab cache
- Tap on map and select to create a new waypoint
- Choose waypoint type "Parking"
- Store the waypoint

The parking waypoint uses the icon "1"

The parking waypoint should use the icon "P"

### Reproducible

Yes

### c:geo Version

2025.06.24-NB

### System information

```text
Samsung S24, Android 15
```

### Additional Information

Might be the same for other waypoint types (not tested)

moving-bits:


moving-bits:
Failing test is unrelated, see PR #17092 for a (non-complete yet) attempt to fix that

PR #17094, opened 2025-06-28
Allow elevation chart to be resized

moving-bits:
## Description
Allows enlarging the elevation chart's height by tapping on its title to get more details "on a view". Toggle state gets remembered.

image


moving-bits:
retest this please

moving-bits:
retest this please

PR #17108, opened 2025-06-30
Country Filters (wip)

JakeDot:

### DIFF
--- a/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
+++ b/main/src/cgeo/geocaching/filter/CountryFilterFactory.java
@@ -16,32 +16,22 @@ import java.util.Map;
 import java.util.Objects;
 
 public class CountryFilterFactory implements IFilterFactory {
-
-
-    static class CountryFilter extends AbstractFilter {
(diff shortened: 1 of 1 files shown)


main/src/cgeo/geocaching/filter/CountryFilterFactory.java
```diff
@@ -10,38 +10,28 @@
 import androidx.annotation.NonNull;
 
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 public class CountryFilterFactory implements IFilterFactory {
```
eddiemuc, line 18: @JakeDot note that the filter system has changed completely since this class was created. As of now the framework it is based on (eg IFilterFactory) des not exist any more.

JakeDot:
I thought so but wasn't sure. Thx

Gesendet von Proton Mail für Android

-------- Ursprüngliche Nachricht --------
Am 01.07.25 07:59 schrieb Eddie :

PR #17133, opened 2025-07-13
Delete orphaned trackingfiles on clean up and restore backup

### LINKED ISSUE #17125: [Nightly] Creating Backup creates a backup of trackfiles in the backup-dir
### Describe your problem!

When creating the backup, about 400 track files are created in a sub-folder “tracks” - where do they come from? I don't see where the original files are located to delete them (probably in an internal directory). So it is unnecessary disk space.
Deleting the "Positionsverlauf" does not change anything

The last thing to be backed up is the database and settings, it would be nice if this happened first, then I could cancel copying the tracks...

Now I can guess why c:geo has been hanging a lot lately.

### How to reproduce?

Creating a backup
open backup-directory
open sub-dir "tracks"

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

[Nigthly] 2025.07.11-NB-7f9e227

### System information

```text
Send via email
```

### Additional Information

_No response_

murggel:
## Description

Delete tracking files before restore (if backup has tracking files) - since they are not so long in the backup, only delete them, if there are some in the backup
Delete orphaned tracking files on database-clean up

## Related issues

fix #17125


murggel:
@eddiemuc  I am not sure, if I used the correct methods to remove the files...

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -4324,6 +4327,29 @@ private static void deleteOrphanedUDC() {
         removeCaches(orphanedUDC, LoadFlags.REMOVE_ALL);
     }
 
+    private static void deleteOrphanedTrackfiles() {
+        // current used trackfiles
+        final ArrayList<Trackfiles> currentTrackFiles = Trackfiles.getTrackfiles();
+        // all trackfiles
+        final List<ImmutablePair<ContentStorage.FileInformation, String>> trackFiles = FolderUtils.get().getAllFiles(Folder.fromFile(LocalStorage.getTrackfilesDir()));
```
eddiemuc, line 4334: You are using the ContentStorage framework here. Although this will work, you may make your life more complex than necessary. Because the trackfile copies are stored in app internal storage, they will always be referencable using simple `File` operations instead.

E.g. here: 
`final FIle[] trackFiles = LocalStorage.getTrackFilesDir().listFiles()`

And then later to delete one of them:
`trackFiles[i].delete()`

(I am assuming that all track files are stored in same dir and that there is no subdirectory structure beneath `LocalStorage.getTrackFilesDir()`

main/src/main/java/cgeo/geocaching/utils/BackupUtils.java
```diff
@@ -302,8 +302,16 @@ public void restoreInternal(final Activity activityContext, final Folder backupD
 
         if (database) {
             AndroidRxUtils.andThenOnUi(Schedulers.io(),
-                () -> FolderUtils.get().copyAll(Folder.fromFolder(backupDir, TRACKS_SUBFOLDER), Folder.fromFile(LocalStorage.getTrackfilesDir()), false),
```
eddiemuc, line 305: I see now here where you probably got the usage of ContentStorage framework from :-)
Having seen this, you may probably leave also the above as is

murggel: Then I would leave it as it is...

murggel:
@eddiemuc ready for merge from my side.

PR #17139, opened 2025-07-16
Logo fixes

ztNFny:
fix splashscreen, use a higher res version of the logo foreground, replace some low-res raster versions


ztNFny:
@moving-bits i don't get your 2nd commit - the dialog wasn't broken? In fact it was better than before. What is your fix supposed to fix?

moving-bits:
> @moving-bits i don't get your 2nd commit - the dialog wasn't broken? In fact it was better than before. What is your fix supposed to fix?

The logo on the installation wizard screen was way too small, and it was sort of left-aligned, which looked very odd (at least over here in the emulator) - no longer looked like a "welcome to your new app" screen.

ztNFny:
It's centered and large here ...

moving-bits:
> It's centered and large here ...

This is strange: I deleted my PR branch and reloaded it from server, now it is fine. As I'm using a script here, I'm 100% safe that it has been done in an identical way. Maybe some caching problem... Anyway, let's merge here now to get the release rolling.

PR #17148, opened 2025-07-19
Skip migration of UDC (fix #17147)

### LINKED ISSUE #17147: UDC migration fails for users having many UDC
After analysis of ticket 201565 on support it seems that we have a bug in our migration routine for UDC geocode randomization:

The current select statement for migration is like this:

```
final String query = "select geocode from cg_caches where geocode >= 'ZZ1000' and geocode < 'ZZ2000'";
```

which attempts to select "unmigrated" UDC based on the assumption, that no user has ZZ codes > 2000.

For the ticket's user this is not true, as the user has app. 9000 UDC, numbers going up to somewhere like `ZZ11500`.

Also, the comparison above does not work for users having geocodes > `ZZ10000`, as the statement above treats the geocodes as a string. Here, a `Z10000 < ZZ2000` is true. For the selection to work the statement would need to be switched to compare numbers, eg. something like this: `final String query = "SELECT geocode FROM cg_caches WHERE geocode LIKE 'ZZ%' AND CAST(substr(geocode,3,10) AS INTEGER) >= 1000 AND CAST(SUBSTR(geocode,3,10) AS INTEGER) <= 2000";`

But to my understanding this would not help, either, as there seems to be no collision detection in `InternalConnector.generateRandomId()` or the migration routine itself.

Therefore I propose to remove the migration routine for now (leaving a blank `105` migration block) and reopen the related issue.

If no one objects I'll create a hotfix with this tomorrow afternoon.

(There is also a Play Store error related to this, `SQLiteConstraintException` leading to an app crash in `DataStore.initAndCheck()`)

Edit: User is back working with c:geo due to a hand-crafted db update (manually added db update 104, then set db version to 106, effectively skipping the 105 step.

moving-bits:
## Description
See #17147

I'll create a hotfix from this if no one objects until tomorrow (Sunday) afternoon.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DataStore.java
@@ -1902,23 +1902,7 @@ public class DataStore {
                     }
 
                     // Migrate UDC geocodes from ZZ1000-based numbers to random ones
-                    if (oldVersion < 105) {
-                        try {
-                            final String query = "select geocode from cg_caches where geocode >= 'ZZ1000' and geocode < 'ZZ2000'";
-                            try (Cursor cursor = db.rawQuery(query, null)) {
-                                while (cursor.moveToNext()) {
-                                    final String geocode = cursor.getString(0);
-                                    final String newGeocode = InternalConnector.PREFIX + InternalConnector.generateRandomId();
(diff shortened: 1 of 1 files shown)


moving-bits:
Failing tests are unrelated

PR #17149, opened 2025-07-20
fix #17142: add foss flavor and exclude mlkit offline translation from it

### LINKED ISSUE #17142: latest release added MLKit (proprietary), blowing the APK size beyond 40 MB
We had to roll back the latest release, for the above stated reasons:

* per-app size limit at IzzyOnDroid is 30 MB
* proprietary components are tolerated at best, but this one now blew their number over the "red line"

Updates are hence disabled here for now. Is there a chance for a "libre" or at least "free-er" build flavor we could pick instead? MLKit is also considered to track, so this would have meant the 4th anti-feature here…

eddiemuc:
fix #17142: add foss flavor and exclude mlkit offline translation from it


eddiemuc:
failing tests are unrelated

eddiemuc:
Failing tests are unrelated.

I will test this locally to ensure I didn't accidentally screw with existing implementation. This might result in some more small changes. Apart from that this PR is ready for review

PR #17153, opened 2025-07-20
catch fatal crashes (fix #10811)

### LINKED ISSUE #10811: Catch fatal exceptions in c:geo with our "report problems" feature
We already have an exception handler, which however still let the app crash.l This could be avoided and our nice "report a problem" dialog could be shown, which would make tracking for fatal crashes way easier...

https://github.com/cgeo/cgeo/blob/9abb5b7c5e90eabbc3a6f87da54370ad81f86f70/main/src/cgeo/geocaching/utils/OOMDumpingUncaughtExceptionHandler.java#L1-L49

fm-sys:
fix #10811

Image


SammysHP:
Awesome image!

fm-sys:
> Awesome image!

Thanks 😊

fm-sys:
Adjusted layout so that "report problem" button keeps visible for longer error messages as well.

moving-bits:
Will errors that get caught by this new feature, still be reported by Play Store statistics? This is essential, IMHO, as
- they will be collected automatically (we do not depend on users actually sending a bug report)
- they get sorted / classified / grouped automatically, so we can see their relevance (how many occurences etc.)

Without those errors being reported to Play Store stats we loose a valuable source for getting to know problems (crashes / ANRs) as early as possible. Otherwise we would depend on users actually sending us the reports - and either don't get the actual significance or get flooded with unsorted reports. (Already today we cannot cope with the amount of incoming support tickets in a timely manner.)

Best would be if we get both, individual report by the user, so that we can react to the user and help him/her individually, and the Play Store stats.

fm-sys:
> Will errors that get caught by this new feature, still be reported by Play Store statistics?

Good catch! Indeed, it doesn't according to [Stack Overflow](https://stackoverflow.com/questions/12536520/does-setting-thread-setdefaultuncaughtexceptionhandler-prevent-googles-error-re), at least not in the default implementation.

While I guess it is really helpful if users can write what they did before the crash happened, and nightly users can even copy the stacktrace directly and paste it on GitHub.

I hope I can find a solution which can utilize both methods. Otherwise, it may be a possibility to only call it if c:geo was [not installed from the Play Store](https://stackoverflow.com/questions/39799618/is-there-a-way-to-check-where-from-app-is-installed/39799997) thus not receiving crash reports anyway.

fm-sys:
Updated the implementation to always call the default exception handler additionally, so play console crash statistics should always be reported. 

Depending on the OS this may prevent the "Oops! c:geo crashed!" screen to be shown, however in my tests the screen is correctly shown on the first crash and from the third crash onwards, only during the second crash in a row the custom handler gets killed as well by the OS.

fm-sys:
Let's give it a try, I hope this will work

PR #17159, opened 2025-07-21
modify import icon

fm-sys:
The arrow in the import icon was pointing downwards, which I personally believe is pretty confusing. Rotate it by 180°.

Screenshot_20250721-144436

grafik

### DIFF
--- a/main/src/main/res/drawable/ic_menu_import.xml
+++ b/main/src/main/res/drawable/ic_menu_import.xml
@@ -5,6 +5,6 @@
     android:viewportHeight="24"
     android:tint="?android:textColorPrimary">
     <path
-        android:pathData="M6,2C4.89,2 4,2.9 4,4V20A2,2 0 0,0 6,22H18A2,2 0 0,0 20,20V8L14,2M13,3.5L18.5,9H13M10.05,11.22L12.88,14.05L15,11.93V19H7.93L10.05,16.88L7.22,14.05"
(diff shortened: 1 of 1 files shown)


ztNFny:
I understand your concern about the old icon. However the proposed one seems off-balance and in terms of iconography is off. 
Looking at the google font lib there's 2 options but tbh I don't find them 100% convincing either.
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:upload_file:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=load&icon.size=24&icon.color=%23e8eaed
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:folder_open:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=folder+open&icon.size=24&icon.color=%23e8eaed

fm-sys:
Do you have a better idea for an icon? The icon is also used together with some others in the context menu, which is why I tried to keep the look consist (screenshot with old icon):

Screenshot_20250728-141250

fm-sys:
moved the arrow a bit more into the edge:


grafik
grafik

ztNFny:
Unfortunately this PR was merged very quickly 🙁 The new icon is "syntactically" wrong - the action is "import FROM file" but it's showing a "INTO file" icon.

> Do you have a better idea for an icon?

Well I posted 2 suggestions in the post above yours ...

PR #17161, opened 2025-07-21
Unified map UX improvements

fm-sys:
Some quality of live UX improvements :)

- make small map lines actually clickable
- don't overlap long click popup with long tab location indicator (was only happening if actionbar hiding is disabled)


fm-sys:
retest this please

Issue #17150, opened 2025-07-20
Use new logo everywhere


fm-sys:
@ztNFny thanks for the great new icon! I honestly didn't notice when it changed on my phone, but looking back I never want to see the old one again :)

So let's update it everywhere:

- [ ] Playstore app icon
- [ ] Playstore dev account icon
- [ ] Github organization icon
- [ ] Gravatar for cgeo@cgeo.org (and others?)
- [ ] Crowdin logo
- [ ] Facebook
- [ ] Twitter
- [ ] ...something else?

fm-sys:
(I don't have write access to any of them, so please help where you can :))

fm-sys:
As quick reference, here are the new logo files again:

<img width="192" height="192" alt="Image" src="https://github.com/user-attachments/assets/842b87ab-11fe-4d82-9b77-02c7dd8eb9f9" />
<img width="192" height="192" alt="Image" src="https://github.com/user-attachments/assets/c60be90b-4e97-418e-85ef-3fcb842cf585" />

SammysHP:
Updated the GitHub organization icon. Do we still use ~~Twitter~~ X?

edit: Just tried to log in on X, but I don't have access to the mail address for 2FA.

fm-sys:
We haven't used Twitter/X since Feb 2023 anymore. Just wanted to mention it here, as we still have "Twitter" links in our app as well as on our homepage. 

Side note: I would be in favor of deleting our c:geo twitter/X account and remove all legacy links to it.

Lineflyer:
I uploaded the new playstore app icon, its being digested by Google right now and should be live together with the new release.
The Google Play dev account icon can probably only be changed by the account owner @mucek4

Lineflyer:
Crowdin updated as well.

Lineflyer:
Added some items above.

Lineflyer:
> Side note: I would be in favor of deleting our c:geo twitter/X account and remove all legacy links to it.

Agree

fm-sys:
I could recreate our Playstore graphics, let's discuss this further over here:  #11915

PR #17162, opened 2025-07-21
remove twitter references from c:geo

fm-sys:
See https://github.com/cgeo/cgeo/issues/17150#issuecomment-3094470307. If someone wants to keep "twitter", maybe also just for archiving their original name, please raise your hand ;)


moving-bits:
Failing tests are unrelated

PR #17166, opened 2025-07-22
remove self-painted overlay from shortcuts (fix #12649)

### LINKED ISSUE #12649: Replace cgeo widgets with adaptive shortcuts
AFAIK it is somehow possible to send `shortcuts` to the launcher as well, so we could think about replacing the old legacy `widgets` with the more modern adaptive `shortcuts`. 

_Originally posted by @fm-sys in https://github.com/cgeo/cgeo/pull/12640#pullrequestreview-867169164_

As soon as time permits I will add a sample code how this could be done. IIRC I have something like this already used for some of my other projects...

fm-sys:
Don't know if this is just a me-problem, but our c:geo shortcuts did look really ugly, at least on my device:
grafik

Remove our custom painted c:geo logo from the shortcut icon, and let the launcher badge the icon instead (some launchers still will not, but that's IMHO not our problem):
grafik


ztNFny:
@fm-sys should this go into release?

fm-sys:
It's not a regression, but it's that way for years now. I think it can stay on master...

PR #17176, opened 2025-07-23
Image Helper: add wrong mimeType to user message

eddiemuc:
Image Helper: add wrong mimeType to user message

This would have helped me with a recent ticket if it has already been in code (ticket ref: https://cgeosupport.droescher.eu/scp/tickets.php?id=9590)


eddiemuc:
Failing test is unrelated

Issue #17184, opened 2025-07-25
load TBs from log page instead of separate network request


ztNFny:
The full list of available TBs is already available in the log page but we make a separate network request (GCWebAPI.getTrackableInventory()) to load it.
This was probably implemented this way for the old logging page but is not required today.

SammysHP:
The logging page only shows the first 20 trackables. Additional trackables have to be loaded via the "load more" button (which might call the web API). Can you verify if your suggestion works with this scenario?

ztNFny:
> The logging page only shows the first 20 trackables. Additional trackables have to be loaded via the "load more" button (which might call the web API). Can you verify if your suggestion works with this scenario?

No idea where to get more than 20 TBs ... But if you have them you can check the logs page source code.

Issue #753, opened 2011-11-04
Maximum of 20 TB shown in cache-details


Lineflyer:
If I open a cache with more than 20 TB (Yes I know, this is rare) the cache-details and the TB-page will only show 20 of them.
Example: At the moment GC193BP  has 24 TBs, but in c:geo I can only count 20.

PR #17193, opened 2025-07-27
get available TBs from log page instead of separate web call

ztNFny:
fixes https://github.com/cgeo/cgeo/issues/17184

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
@@ -176,6 +176,7 @@ public final class GCConstants {
     static final Pattern PATTERN_LINEBREAK = Pattern.compile("<(br|p)[^>]*>");
     // new logpage logtype pattern:         logSettings.logTypes.push({"Value":46,"Description":"Owner maintenance","IsRealtimeOnly":false});
     static final Pattern PATTERN_TYPE4 = Pattern.compile("\"logTypes\":\\[([^]]+)]");
+    static final Pattern PATTERN_LOGPAGE_TRACKABLES = Pattern.compile("\"trackables\":\\[([^]]+)]");
     static final Pattern PATTERN_MAINTENANCE = Pattern.compile("<span id=\"ctl00_ContentBody_LogBookPanel1_lbConfirm\"[^>]*>([^<]*<font[^>]*>)?([^<]+)(</font>[^<]*)?</span>", Pattern.CASE_INSENSITIVE);
     static final Pattern PATTERN_VIEWSTATEFIELDCOUNT = Pattern.compile("id=\"__VIEWSTATEFIELDCOUNT\"[^(value)]+value=\"(\\d+)\"[^>]+>", Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
     static final Pattern PATTERN_VIEWSTATES = Pattern.compile("id=\"__VIEWSTATE(\\d*)\"[^(value)]+value=\"([^\"]+)\"[^>]+>", Pattern.CASE_INSENSITIVE | Pattern.MULTILINE);
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
@@ -176,7 +176,8 @@ public final class GCConstants {
     static final Pattern PATTERN_LINEBREAK = Pattern.compile("<(br|p)[^>]*>");
     // new logpage logtype pattern:         logSettings.logTypes.push({"Value":46,"Description":"Owner maintenance","IsRealtimeOnly":false});
     static final Pattern PATTERN_TYPE4 = Pattern.compile("\"logTypes\":\\[([^]]+)]");
-    static final Pattern PATTERN_LOGPAGE_TRACKABLES = Pattern.compile("\"trackables\":\\[([^]]+)]");
+    static final Pattern PATTERN_TOTAL_TRACKABLES = Pattern.compile("\"totalTrackables\":\"([\\d]+)\"");
+    static final Pattern PATTERN_LOGPAGE_TRACKABLES = Pattern.compile("\"trackables\":\\[(.+?)\\]\\},\"__N_SSP\"");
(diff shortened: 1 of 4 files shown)


ztNFny:
should be tested with more than 20 TBs, see https://github.com/cgeo/cgeo/issues/17184#issuecomment-3127652834
Does someone have that?

main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
```diff
@@ -176,6 +176,7 @@ public final class GCConstants {
     static final Pattern PATTERN_LINEBREAK = Pattern.compile("<(br|p)[^>]*>");
     // new logpage logtype pattern:         logSettings.logTypes.push({"Value":46,"Description":"Owner maintenance","IsRealtimeOnly":false});
     static final Pattern PATTERN_TYPE4 = Pattern.compile("\"logTypes\":\\[([^]]+)]");
+    static final Pattern PATTERN_LOGPAGE_TRACKABLES = Pattern.compile("\"trackables\":\\[([^]]+)]");
```
eddiemuc: This pattern shall parse a complex json inside the web page, correct?

Wouldn't it fail if eg a trackable name contains a `]`?

Should be tested at least.
(Can't test it, I'm in the subway...)

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1558,21 +1557,21 @@ public static List<LogTypeTrackable> parseLogTypesTrackables(final String page)
...
-            return CollectionStream.of(availableTypes)
-                    .filter(a -> a.value > 0)
-                    .map(a -> LogType.getById(a.value))
-                    .filter(t -> t != LogType.UPDATE_COORDINATES)
-                    .toList();
+        final AvailableLogType[] availableTypes = parseLogTypes(page);
+        if (availableTypes == null) {
+            return Collections.emptyList();
         }
+        return CollectionStream.of(availableTypes)
```
eddiemuc: While you're editing this part of code anyway, you might want to replace usagea of `CollectionStream` with normal Java Streaming (`Arrays.asList(availableTypes).stream()`).  `CollectionStream` is from a time when we couldn't use normal Streaming in c:geo code. It is deprecated these days (I have to mark it accordingly one of these days...)

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1583,7 +1582,40 @@ private static AvailableLogType[] parseLogTypes(final String page) {
...
         } catch (final Exception e) {
             Log.e("Error parsing log types from [" + match + "]", e);
-            return new AvailableLogType[0];
+            return null;
+        }
+    }
+
+    @NonNull
+    static List<Trackable> parseTrackables(final String page) {
+        final GCWebAPI.TrackableInventoryEntry[] trackableInventoryItems = parseTrackablesJson(page);
```
eddiemuc, line 1602: Class `TrackableInventoryEntry` should probably moved to `GCParser` if it is no longer used in `GCWebAPI`?

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1583,7 +1582,40 @@ private static AvailableLogType[] parseLogTypes(final String page) {
...
+            final Trackable trackable = new Trackable();
+            trackable.setGeocode(entry.referenceCode);
+            trackable.setTrackingcode(entry.trackingNumber);
+            trackable.setName(entry.name);
+            trackable.forceSetBrand(TrackableBrand.TRAVELBUG);
+            return trackable;
+        }).toList();
+    }
+
+    private static GCWebAPI.TrackableInventoryEntry[] parseTrackablesJson(final String page) {
```
eddiemuc, line 1616: As already pointed out by @SammysHP , I fear this code will only return the trackables visible for the user on the first load of the log page (probably first 20?). It doesn't consider trackable inventory paging (or whatever else gc.com is using on their log page when not all user trackables fit on first page). 

Should be tested at least.

main/src/main/java/cgeo/geocaching/connector/gc/GCWebAPI.java
```diff
@@ -890,24 +890,6 @@ private static void tryGuessMissingDistances(final List<Geocache> caches, final
...
-     * https://www.geocaching.com/api/proxy/trackables?inCollection=false&skip=0&take=50
-     */
-    @NonNull
-    @WorkerThread
-    static List<TrackableInventoryEntry> getTrackableInventory() {
-        final List<TrackableInventoryEntry> trackableInventoryEntries = new ArrayList<>();
-        int skip = 0;
-        TrackableInventoryEntry[] entries;
-        do {
-            entries = apiProxyReq().uri("/trackables?inCollection=false&inInventory=true&take=" + MAX_TAKE + "&skip=" + skip).requestJson(TrackableInventoryEntry[].class).blockingGet();
```
eddiemuc: The old code ensures all trackables are loaded using `take` and `skip` parameter in a loop. Don't know if this is necessary and which mechanisms are available for that when using the web page.

main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -1583,7 +1582,40 @@ private static AvailableLogType[] parseLogTypes(final String page) {
...
+        if (StringUtils.isBlank(page)) {
+            return null;
+        }
+
+        final String match = TextUtils.getMatch(page, GCConstants.PATTERN_LOGPAGE_TRACKABLES, null);
```
eddiemuc, line 1621: My heart dies a little each time we introduce another HTML page parsing part into c:geo and class `GCParser` 😀

eddiemuc (dismissed):

eddiemuc:
Before merging, handling of many trackables (eg 200+) as well as trackables using special characters (eg `]` in their name) should be tested.

eddiemuc:
> Before merging, handling of many trackables (eg 200+) as well as trackables using special characters (eg `]` in their name) should be tested.

Ia this resolved? Can we merge the PR?

Lineflyer:
@ztNFny Does this fix #753 ?

ztNFny:
> @ztNFny Does this fix #753 ?

that was fixed years ago ;-)

PR #17229, opened 2025-08-01
Improve header in variable and waypoint-tab

murggel:
## Description

Improve header of waypoint and variables tab
* Fix the header of waypoint and variables tab, so it stays always on top
* Add variable name to button of next variable
* Shorten info-text in variable tab
* Using chip-layout

| Waypoints | Variables |
| --- | --- |
| image | image |

## Related issues

Split up PR #17019

## Additional context
Not sure yet about the waypoint tab, because the header is quite large (in some cases there is a third button ("Add from clipboard").

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -2281,9 +2281,9 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                     }
                 });
 
-            binding.hideVisitedWaypoints.setChecked(Settings.getHideVisitedWaypoints());
-            binding.hideVisitedWaypoints.setOnCheckedChangeListener((buttonView, isChecked) -> {
-                    Settings.setHideVisitedWaypoints(isChecked);
+            binding.chipVisitedWaypoints.setChecked(!Settings.getHideVisitedWaypoints());
+            binding.chipVisitedWaypoints.setOnCheckedChangeListener((buttonView, isChecked) -> {
+                Settings.setHideVisitedWaypoints(!isChecked);
 
                     final List<Waypoint> sortedWaypoints2 = createSortedWaypointList(cache);
                     Collections.sort(sortedWaypoints2, cache.getWaypointComparator());
--- a/main/src/main/res/layout/cachedetail_waypoints_page.xml
+++ b/main/src/main/res/layout/cachedetail_waypoints_page.xml
@@ -42,11 +42,12 @@
             android:text="@string/cache_waypoints_add_fromclipboard"
             android:visibility="gone" />
 
-        <CheckBox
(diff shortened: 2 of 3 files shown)


murggel:
Not sure yet about the waypoint tab, because the header is quite large (in some cases there is a third button ("Add from clipboard").

murggel:
Using chip-layout
image image

murggel:
How about adding some waypoint-types to hide? Only visible if available... 

image

Don't know how about the other types...
image

But it needs again more room...
Or add a filter-view... But that would be definitly a separate PR

ztNFny:
> Not sure yet about the waypoint tab, because the header is quite large (in some cases there is a third button ("Add from clipboard").

I'm wondering whether we could get rid of all but the standard "add wp" button. Both current location and clipboard are available in the dialog. It'd be 1 additional click for what's probably a small subset of cases.
For current location imho it's advisable to open the dialog anyway - in waypoints tab c:geo is not actively polling the location, so error rate is ~20m while after opening the dialog it starts polling.

> Using chip-layout

Nice! I'll go ahead and "steal" that for the logs tab :-)

> How about adding some waypoint-types to hide? Only visible if available...

I wouldn't go for showing all the types but how about
- visited
- extracted from personal note (needs a better title and not sure that info is even available to poll)
- generated (from range-formula)

ztNFny:
> Nice! I'll go ahead and "steal" that for the logs tab :-)

as soon as you merge it ;-)

murggel:
> > Not sure yet about the waypoint tab, because the header is quite large (in some cases there is a third button ("Add from clipboard").
> 
> I'm wondering whether we could get rid of all but the standard "add wp" button. Both current location and clipboard are available in the dialog. It'd be 1 additional click for what's probably a small subset of cases. For current location imho it's advisable to open the dialog anyway - in waypoints tab c:geo is not actively polling the location, so error rate is ~20m while after opening the dialog it starts polling.
> 
> > Using chip-layout
> 
> Nice! I'll go ahead and "steal" that for the logs tab :-)
> 
> > How about adding some waypoint-types to hide? Only visible if available...
> 
> I wouldn't go for showing all the types but how about
> 
>     * visited
> 
>     * extracted from personal note (needs a better title and not sure that info is even available to poll)
> 
>     * generated (from range-formula)

I would then leave this PR as it is and create a new issue, where we can discuss this -> see #17252

murggel:
Ready for review

murggel:
> > Nice! I'll go ahead and "steal" that for the logs tab :-)
> 
> as soon as you merge it ;-)

I am not able to merge it...

eddiemuc:
Lgtm

ztNFny:
@murggel seems the "visited" chip is always shown. Imho it should only be shown if there's at least 1 WP marked as visited.

PR #17231, opened 2025-08-02
UnifiedMap, GeoItemLayer: add better logging to analyze problems with nonclickable items

eddiemuc:
UnifiedMap, GeoItemLayer: add better logging to analyze problems with nonclickable items

This PR improves logging inside UnifiedMapActivity and GeoItemLayer. It is designed to analyze the problem with non-clickable items which @murggel was experiencing. It is at the same time designed in a way that it can stay in the code permanently (and might be useful to analyze future, not-yet-known problems).


eddiemuc:
"default" and "Codacy" are green

PR #17233, opened 2025-08-02
fix #17232: Wherigo, display no symbols on push dialog buttons when not default ok button

eddiemuc:
fix #17232: Wherigo, display no symbols on push dialog buttons when not default ok button


eddiemuc:
Codacy s green. checkstyle/lint errors were introduced by previous PRs and are unrelated to this one -> should be fixed in own PR

ztNFny:
why is the logic different for button1 and button2?
afaict the goal is to have icons for the "undefined" dialogs that show ok/cancel and not for custom ones where the buttons would be "text1/text2". But seeing the code now for the undefined dialog only ok will have an icon?

eddiemuc:
> why is the logic different for button1 and button2?
> afaict the goal is to have icons for the "undefined" dialogs that show ok/cancel and not for custom ones where the buttons would be "text1/text2". But seeing the code now for the undefined dialog only ok will have an icon?

Openwig states to show a second button ONLY if manual text is provided for button2. Thus my initial thought of a default OK/cancel logic was wrong.

PR #17235, opened 2025-08-02
fix #17210: fix handling of load-more-caches-footers for offline lists

### LINKED ISSUE #17210: "load more" does not respect offline filters
When using offline filters, a button at the bottom of the list suggests there are more caches available, but it just uses all caches of the list as reference. Tapping on it does nothing.

Image

eddiemuc:
fix #17210: fix handling of load-more-caches-footers for offline lists


eddiemuc:
Codacy "green", failing tests at "default" is unrelated

PR #17236, opened 2025-08-02
fix lint errors

eddiemuc:
fix lint errors


eddiemuc:
Codacy + default green

PR #17237, opened 2025-08-02
fix #17230: Fix condition for deleting logs

### LINKED ISSUE #17230: Distinguish "Delete log" actions
Currently there are two different log context menu options called "Delete log". One is to delete and offline log saved in c:geo, and the other is to delete someone's log as a Cache Owner action. For most times, they don't appear on the menu together.

However, if you have an offline log saved on your own cache, you can see these together, and you can't tell which one is which.

I suggest calling the one I have circled red "Delete offline log".
Additionally the one that is circled blue should not be there at all when pressing the context menu on an offline log, as it is an online action.

Image
Image
Image

murggel:
## Description

Fix OR-condition in `GCConnector::canDeleteLog`: add missing brackets for correct OR-condition

## Related issues

fix #17230

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConnector.java
@@ -220,15 +220,15 @@ public class GCConnector extends AbstractConnector implements ISearchByGeocode, 
 
     @Override
     public boolean canEditLog(final Geocache cache, final LogEntry logEntry) {
-        //needs to be log author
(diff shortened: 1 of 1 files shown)


murggel:
@eddiemuc couldn't really test it - Maybe I try to create a test-case...

eddiemuc:
@murggel I believe you found the culprit and fixed it. Thanks for undoing my bugs :-)
Will merge now

murggel:
> Thanks for undoing my bugs :-)

If it's that easy 😇

Issue #17169, opened 2025-07-22
Adding geocaches to bookmark often fails w/o explanation


eddiemuc:
Taken from support: https://cgeosupport.droescher.eu/scp/tickets.php?id=9265 
Reproduced by me with 2025.07.17

When using "add selected (caches) to bookmark list", the behavior is very shaky.
* In first step c:geo retrieves and shows a list of existing bookmark lists. Thus usualy works fine
* After choosing a list, then very often the process afterwards fails for no obvious reasons with error message "Fehler beim Hinzufügen von Caches zur Lesezeichenliste"

I could reproduce the problem easily while having good internet connection and being logged in as PREMIUM several times. I found following exception in log:

```
07-22 17:33:24.620 25288 27733 E cgeo    : [OkHttp] GCParser.uploadPersonalNote - cannot upload caches to GC.com bookmark list
07-22 17:33:24.620 25288 27733 E cgeo    : java.io.IOException: unsuccessful response: Response{protocol=http/1.1, code=401, message=Unauthorized, url=https://www.geocaching.com/api/proxy/web/v1/lists/BM7D93C/geocaches}
07-22 17:33:24.620 25288 27733 E cgeo    : 	at cgeo.geocaching.network.Network.$r8$lambda$wUQJhvGOfGQvK1C_-6clmCSzy0k(SourceFile:705)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at cgeo.geocaching.network.Network$$ExternalSyntheticLambda5.apply(SourceFile:0)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at io.reactivex.rxjava3.internal.operators.single.SingleFlatMap$SingleFlatMapCallback.onSuccess(SourceFile:77)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at io.reactivex.rxjava3.internal.operators.single.SingleCreate$Emitter.onSuccess(SourceFile:68)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at cgeo.geocaching.utils.RxOkHttpUtils$1.onResponse(SourceFile:48)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at okhttp3.internal.connection.RealCall$AsyncCall.run(SourceFile:519)
07-22 17:33:24.620 25288 27733 E cgeo    : 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1156)
[...]

PR #17244, opened 2025-08-03
warn on adding too many caches to list

ztNFny:
- add cache count to bookmark lists selection
- add warning when trying to add more caches to a list than it can hold

fixes https://github.com/cgeo/cgeo/issues/17169

Image

Image


eddiemuc:
Failing tests are unrelated

PR #17245, opened 2025-08-03
change short description pattern

### LINKED ISSUE #17224: GC.com: short description parsing error
Support ticket https://cgeosupport.droescher.eu/scp/tickets.php?id=9642
c:geo Version 2025.07.22

When opening geocache https://www.geocaching.com/geocache/GC4Z19C_1-was-ware-wenn there's a parsing error in the cache's short description:

Image

Image

This error is caused by the usage of `<span>` tags by the author in the cache description - and because GCParser uses <span> as the end delimiter of the short description:

Image

Image

ztNFny:
extend the pattern to avoid false matches to fix https://github.com/cgeo/cgeo/issues/17224

However this makes it also more fragile to changes on geocaching.com.

Considering that ShortDescription hasn't been available for new caches for a long time now I'm not sure whether this is worth it ... 60/40 against merging this from me.


eddiemuc:
Retest this please

Lineflyer:
Seeing, that no more complaints came similar to #17224 how about a "Wont fix" for now and closing here?
If you say, that its a more fragile regex, not fixing looks reasonable to me.

PR #17247, opened 2025-08-03
use button instead of checkbox for giving favorites in log dialog

ztNFny:
image
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
@@ -304,7 +304,7 @@ public class LogCacheActivity extends AbstractLoggingActivity implements LoaderM
             binding.favoriteCheck.setChecked(offlineLogEntry.favorite);
             //If fav check is set then ALWAYS make the checkbox visible. See https://github.com/cgeo/cgeo/issues/13309#issuecomment-1702026609
             if (offlineLogEntry.favorite) {
-                binding.favoriteCheck.setVisibility(View.VISIBLE);
+                binding.favoriteBox.setVisibility(View.VISIBLE);
             }
             binding.logPassword.setText(offlineLogEntry.password);
             inventoryAdapter.putActions(offlineLogEntry.inventoryActions);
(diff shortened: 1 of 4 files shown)


eddiemuc:
This seems unrelated to an issue (?). Setting to "feedback required" for now

WeinWalker:
Looks nice, does not add functionality. The checkbox is fine with me

eddiemuc:
Closing here?

ztNFny:
> Closing here?

Why? Merge.

The only comment was from a non-contributor. And of course a UI change doesn't add functionality, that's the whole point of UI.

eddiemuc:
@cgeo/team I will give a grace period of another day for comments. I will merge afterwards and expect then no discussions about this change any more

bekuno:
I'm unsure.
We already use the crossed-out heart symbol on the cache main page to mark the cache as a non-favourite.
It would make more sense to use this symbol here as well to indicate non-favourite status.

ztNFny:
> I'm unsure.
> We already use the crossed-out heart symbol on the cache main page to mark the cache as a non-favourite.
> It would make more sense to use this symbol here as well to indicate non-favourite status.

And also then keep the active state when enabled?

SammysHP:
Without direct comparison I don't see which variant is the active state. Even now I'm not completely sure if the orange or white version adds it to favorites.

ztNFny:
> Without direct comparison I don't see which variant is the active state. Even now I'm not completely sure if the orange or white version adds it to favorites.

interesting - to me there wasn't any question that the accent color which is used to indicate all "active" stuff is meaning active/enabled.

I've modified the "inactive" icon to the "remove" drawable now:
image
image

SammysHP:
> interesting - to me there wasn't any question that the accent color which is used to indicate all "active" stuff is meaning active/enabled.

The problem is, that a button indicates an action when you press it, but it doesn't communicate a state that well. That's why we have checkboxes and toogle switches.

WeinWalker:
I have never noticed any color changes of the Like Button on the details page... 

Thx for all your work

eddiemuc:
The "mark as favorite" button now has kind of the "opposite" logic of buttons in other area (apart from color). 
- Here, when one does NOT want to give a favorite point a crossed-out heart is displayed (and tapping button ADDs a favorite point to it)
- eg for watchlist, when crossed-out glasses are displayed this means the cache IS on the watchlist (and tapping button REMOVES it from the list)

ztNFny:
Because it's an indicator - that's why i didn't want to put the crossed-out heart!

Having "accent color" = active is a pattern used in many places of cgeo and noone had problems identifying it as such there.

I thought this simple change would greatly improve the discoverability and accessibility of the "mark as fav" feature, but let's just close here then ... 🙁

SammysHP:
> I thought this simple change would greatly improve the discoverability and **accessibility** of the "mark as fav" feature

Using colors doesn't improve accessibility. What if you're color blind or use high contrast black and white?

ztNFny:
> > I thought this simple change would greatly improve the discoverability and **accessibility** of the "mark as fav" feature
> 
> Using colors doesn't improve accessibility. What if you're color blind or use high contrast black and white?

It's using a standard ui element so it's even screen reader friendly. And any high contrast setting will highlight the difference between the 2 states even more. 

Using colored indicators is a common theme in both app and web development for ... basically forever:






But I've closed this PR now, so no need for you to express your disliking of it any further.

SammysHP:
Please don't get me wrong, this isn't about me disliking a specific change. I just wanted to highlight a possible usability issue with that change.

Btw both of your examples show an outline vs solid icon, compared to two solid icons with different color.

Issue #17164, opened 2025-07-21
Loading / updating cache data takes a long time


moving-bits:
We have a few tickets on support claiming long loading times for cache data with the new version 2025.07.17. I have seen that for myself for the better part over the last weeks, but can't say, though, whether it's related to the new version.

It may be some server problems (although one of the users said, there are no mentions of this in the gc forums), or it could be some other issue causing this.

Anyone experience long loading times (or mostly short loading times)?

SammysHP:
The website is also slow for me from time to time. Don't we log the time for the requests in the (extended) log?

bekuno:
I had sometimes ANR messages after a fresh start of c:geo.
After waiting some (more) minutes it works normal.
It was not reproducible.

eddiemuc:
No noticable problems for me. Sure server is sometimes slow but that's nothing new :-)

ztNFny:
geocaching.com has been slow, sometimes very very very slow, for the better part of 2-3 weeks - including frequent API outtages. I have error rates of about 15% in some scripts.

alan666notb:
I find refreshing  or loading new caches is now dreadfully slow, 20-25 seconds each at the moment.
It's been going on a while now, can't pinpoint when it got really bad.

murggel:
@moving-bits  are the changes from PR #17136 and PR #17080 (sync timestamp of online log) included in that version? Maybe this causes the performance-problems?

SammysHP:
Just compared c:geo with the browser and there's indeed a huge difference in loading time (2-3s in the browser vs. 20s in c:geo).

moving-bits:
> [@moving-bits](https://github.com/moving-bits) are the changes from PR [#17136](https://github.com/cgeo/cgeo/pull/17136) and PR [#17080](https://github.com/cgeo/cgeo/pull/17080) (sync timestamp of online log) included in that version? Maybe this causes the performance-problems?

Both PR mentioned have been merged to `master`, thus they are not part of the release version 2025.07.17

moving-bits:
> Just compared c:geo with the browser and there's indeed a huge difference in loading time (2-3s in the browser vs. 20s in c:geo).

is this from the same machine?
does it make any difference if you change c:geo's user agent for testing purposes?

ztNFny:
> Just compared c:geo with the browser and there's indeed a huge difference in loading time (2-3s in the browser vs. 20s in c:geo).

[...]

PR #17249, opened 2025-08-03
rel to #17164: log HTTP calls which take a long time

eddiemuc:
rel to #17164: log HTTP calls which take a long time (as of not that's 5 seconds)


eddiemuc:
Codacy and default are green

ztNFny:
probably no longer needed? We never had that slow requests. But anyway, more logging also doesn't hurt.

eddiemuc:
> probably no longer needed? We never had that slow requests. But anyway, more logging also doesn't hurt.

Doesn't hurt I would argue

PR #17260, opened 2025-08-04
optimize GC login flow for performance

ztNFny:
optimize the GC login flow:
Calling https://www.geocaching.com/account/signin with a logged-in user redirects to https://www.geocaching.com/play/search which is quite large (and not really used for anything)
Afterwards we call https://www.geocaching.com/play/serverparameters/params to load info about the user, language, ...

This PR changes the signin-call to redirect to https://www.geocaching.com/play/serverparameters/params which a) saves 1 HTTP call, b) prevents us from receiving the large /play/search response

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCConstants.java
@@ -103,6 +103,7 @@ public final class GCConstants {
 
     // Info box top-right
     public static final Pattern PATTERN_LOGIN_NAME2 = Pattern.compile("window(?>\\.|\\[')(?:headerSettings|chromeSettings)(?>'\\])?\\s*=\\s*\\{[\\S\\s]*\"username\":\\s*\"([^\"]*)\",?[\\S\\s]*\\}");
+    public static final Pattern PATTERN_LOGIN_NAME3 = Pattern.compile("user:info\":\\s*\\{\\s*\"username\":\\s*\"(([^\\\\\\\"]*(\\\\[a-z\\\"])*)*)\\\",");
     /**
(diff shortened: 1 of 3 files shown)


eddiemuc:
Code lgtm.
Failing test could be related because class was touched. Will redo tests

eddiemuc:
Retest this please

eddiemuc:
Test `cgeo.geocaching.connector.gc.GCConstantsTest.testCacheCountOnline(GCConstantsTest.java:52)` seems to fail permanently with this PR. @ztNFny could you please take a look?

ztNFny:
https://www.geocaching.com/play/serverparameters/params doesn't have info about the find count, so that one is not updated during login anymore. So we need to add another request again to update that - question is whether there's a "small"page that has the info.

eddiemuc:
> https://www.geocaching.com/play/serverparameters/params doesn't have info about the find count, so that one is not updated during login anymore. So we need to add another request again to update that - question is whether there's a "small"page that has the info.

Would it work to forward to any other page with slightly less content than the start page but still with the header containing found into? Eg https://www.geocaching.com/plan/lists?

ztNFny:
"any other page" will work. The smaller the page the more benefit. serverparams is very small + loaded anyway, so that'd have been the ideal fit...
Let's try to identify a page with minimal content that still works - ideally some json.

ztNFny:
Found a good one, need to do some code refactoring -> WIP

ztNFny:
now fetching https://www.geocaching.com/account/settings/homelocation as login redirect - this is another page we're anyway fetching during the login process, so performance result is as good as the previous.

eddiemuc:
Lgtm. Thanks for the contribution!

PR #17263, opened 2025-08-04
filter for own/friends/owner logs tab

ztNFny:
Allows filtering the "Own/Friends/Owner logs" tab by log type

image


main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
```diff
@@ -89,13 +90,58 @@ protected void addHeaderView() {
...
+            if (!binding.chipOwn.isChecked()) {
+                logs.removeAll(ownLogs);
+            }
+            if (!binding.chipFriends.isChecked()) {
+                logs.removeAll(friendsLogs);
+            }
+            if (!binding.chipOwner.isChecked()) {
+                logs.removeAll(ownerLogs);
+            }
+            logsAdapter.clear();
```
eddiemuc: If Understand the approach correctly then code would +in case of filtering) first add all logs to adapter (in LogsViewCreator), then delete them all here again, then re-add the filtered logs? 

Is it possible to improve this by moving the filter logic to LogsViewCreator (but only using it from CacheLogsVirwCreator)?

eddiemuc:
PR is ok for me, but waiting for review of @murggel as requested

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
ztNFny: The logic is specific to CacheLogsViewCreator, it shouldn't be in the abstract class. That one doesn't know about the cache and can't do the comparisons required.

eddiemuc: Filter comparison logic could easily be passed to superclass eg via a filter `Predicate<LogEntry>`.

ztNFny: Not easily as the context isn't available there. And what would be the goal of doing it? It'll be like 20 additional lines of code to have a specific function in a generic place.
If we implement the same functionality for the TB log view all the actual logic will be different, the only commonality would be building the result list - a 1-liner.

eddiemuc: I don't believe this would result in more code lines. More important, it would safe unnecessary double-adding/removing of logs to the adapter.

Looking closer ad the code: isn't it possible to write a predicate to filter LogEntries based on the LogEntry-data alone?
- LogEntry has a friend flag to identify whether it is a friend log or not
- "owner" and "own" are already identified via LogEntry data alone
WOuldn't it be possible to get rid of most of the list-adding/removing-logic doing that?

In theory you could also consider usage of https://github.com/cgeo/cgeo/blob/master/main/src/main/java/cgeo/geocaching/ui/recyclerview/ManagedListAdapter.java which has the filter function already build in (see method `setFilter()` of the class). Unfortunatly, this adapter is for `RecyclerView`, not for `ListView`

ztNFny: I've moved the filtering into getLogs() to avoid updating the adapter. But it simply doesn't belong in LogsViewCreator - that one doesn't handle retrieving the log entries in any way, it just draws them.

> LogEntry has a friend flag to identify whether it is a friend log or not

"friend logs" are not just friend logs but "friends + own + owner", that's why "own + owner" might need to be removed from getFriendsLogs() result. The alternative would be storing 4 kinds of logs in DB (any, own, owner, friends) instead of the current 2, but that's a very big change with little benefit.

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
```diff
@@ -131,21 +162,21 @@ private void addLogCountsHeader() {
...
     }
 
     private void addEmptyLogsHeader() {
         if (countview2 != null) {
-            binding.getRoot().removeHeaderView(countview2);
+            binding.logsItems.removeHeaderView(countview2);
             countview2 = null;
         }
 
         if (getLogs().isEmpty()) {
```
murggel, line 175: I found one glitch - if all filters are disabled, the text "The logbook of this cache is empty." is shown as well as the count per log-type
image

I guess, since the filtering is now in getLogs, it returns 0 if all filters are deactivated

murggel:
I like the filtering!

But I would prefer, if the filter buttons are beneath the header with the count.
image

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
ztNFny: That's the correct behaviour imho. All filters unchecked = don't show me anything.
Text might be adjusted to say "you've filtered out everything"

murggel: I agree for not showing any logs.

But is it correct, that both texts are shown?
I would prefer to see only the counts of log types and not the text "The logbook of this cache is empty."

ztNFny: > But is it correct, that both texts are shown?

Well, at that moment the logbook is empty, so ... Not showing any text would probably look strange I guess? No hard feelings about it though.

ztNFny:
> I like the filtering!
> 
> But I would prefer, if the filter buttons are beneath the header with the count. image

Then the filter-buttons will also scroll. Is that intended?

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
murggel: If you filter for e.g. "friend logs" you don't adjust the count of the log-types as well, so I would always show the count of the "unfiltered" logs in that tab. I don't need a text that tells me, no one's are shown - my opinion 

But I guess some users need a hint, that the logs a filtered out (see problems with filter im map)

ztNFny: It is doing just that already? The count is never adjusted.
I think I misunderstand your point - could you please rephrase?

murggel: Probably I misunderstood your point...

I thought you didn't want to show the log count

murggel: So I would go then with an additional text like your proposal "All logs are filtered out" (adapted) - which to be honest I missed / overread 😬

ztNFny: My intention is:
- logcount stays static, not adjusted by filter
- list view shows either the logs or the "is empty" text

hope that's clear now

In addition to that we could change the "is empty" text if it's empty due to a filter - but not sure whether it's really feasible to do, need to check.

ztNFny:
> > I like the filtering!
> > But I would prefer, if the filter buttons are beneath the header with the count. image
> 
> Then the filter-buttons will also scroll. Is that intended?

Placing the filters above the counter makes me want to update the counter on filtering ...

ztNFny:
empty view:
image

updating count:
image
image

murggel:
> But I would prefer, if the filter buttons are beneath the header with the count.
> image

If the count does consider the filter, then showing the count beneath the filter would be ok for me...

Bur probably put them both outside of the scrollable area?

ztNFny:
> > But I would prefer, if the filter buttons are beneath the header with the count.
> > image
> 
> If the count does consider the filter, then showing the count beneath the filter would be ok for me...
> 
> Bur probably put them both outside of the scrollable area?

They are both inside the scrollable area now ... See latest screenshots ;-)

murggel:
I would prefer, that the filter buttons stay on top of the page and don't disappear while scrolling....

ztNFny:
> I would prefer, that the filter buttons stay on top of the page and don't disappear while scrolling....

the reason i've now made them scroll is that toggling a filter re-inits the view and jumps again to the top of the list. So using them while scrolled makes little sense.
(Also they're taking up permanent space then)

murggel:
> > I would prefer, that the filter buttons stay on top of the page and don't disappear while scrolling....
> 
> the reason i've now made them scroll is that toggling a filter re-inits the view and jumps again to the top of the list. So using them while scrolled makes little sense.
> (Also they're taking up permanent space then)

Then I probably put the filter and buttons in waypoint tab also back inside the scrollable area. With the changing to chip-layout I also put them outside...

ztNFny:
Hm, just checked waypoints and i like only the actual list scrolling there. Damn ... we should be consistent there. 

Then i guess I'll change it to fixed again here 😁

murggel:
> Hm, just checked waypoints and i like only the actual list scrolling there. Damn ... we should be consistent there. 
> 
> Then i guess I'll change it to fixed again here 😁
> 

Even if it takes so much space with the buttons?
I am really unsure there...

bekuno:
> > Hm, just checked waypoints and i like only the actual list scrolling there. Damn ... we should be consistent there.
> > Then i guess I'll change it to fixed again here 😁
> 
> Even if it takes so much space with the buttons? I am really unsure there...

I would love it if the buttons were outside. That would be much more practical.

ztNFny:
For waypoints i frequently have the need to check something in the list, then add a new WP which requires scrolling up again. 
e.g. find a WP to copy it's coordinates, then create a new one based on it with projection

That's why there i find it useful to have them fixed. 

Space however is a good reason against it, at least on smaller devices ...

I was contemplating whether we could have just 1 button in that dialog - "my location" is available from the "add wp"dialog - that'd again save some space.

eddiemuc:
> ... toggling a filter re-inits the view and jumps again to the top of the list. So using them while scrolled makes little sense.

We have this challenge in other places (eg stored cache list filtering or reloading a caches description). There we solve it by storing the Y-position of the scrollview before updating and reapplying it after updating the list.

murggel:
> For waypoints .... Space however is a good reason against it, at least on smaller devices ...
> 
> I was contemplating whether we could have just 1 button in that dialog - "my location" is available from the "add wp"dialog - that'd again save some space.

That is something, what @ReggieDoodle also suggests somewhere (in his PR with the new Coordinates dialog ?)

murggel:
> > ... toggling a filter re-inits the view and jumps again to the top of the list. So using them while scrolled makes little sense.
> 
> We have this challenge in other places (eg stored cache list filtering or reloading a caches description). There we solve it by storing the Y-position of the scrollview before updating and reapplying it after updating the list.
> 
> 

I can have a look at it, if that is feasible here too

ztNFny:
> > > ... toggling a filter re-inits the view and jumps again to the top of the list. So using them while scrolled makes little sense.
> > 
> > We have this challenge in other places (eg stored cache list filtering or reloading a caches description). There we solve it by storing the Y-position of the scrollview before updating and reapplying it after updating the list.
> > 
> > 
> 
> I can have a look at it, if that is feasible here too

It isn't applicable here. By changing the filters the list changes completely. If you jump to the previous y-position you'll be looking at a different log (as what previously was log #3 might now be log #1)

ztNFny:
made the filters "stuck" on top of the page again - if we decide to "unstick" them again it's a simple commit to revert :-)

This should be ready to go now.

murggel:
> made the filters "stuck" on top of the page again - if we decide to "unstick" them again it's a simple commit to revert :-)
> 
> This should be ready to go now.

Am I petty, if the count could also stuck on top (also for the page with all logs)?

Otherwise it looks fine to me :-)

ztNFny:
> > made the filters "stuck" on top of the page again - if we decide to "unstick" them again it's a simple commit to revert :-)
> > 
> > This should be ready to go now.
> 
> Am I petty, if the count could also stuck on top (also for the page with all logs)?

Wouldn't want to do it as part of this PR , let's discuss in an separate issue.

outside of this PR: imho there's no value having this sticky on the screen.

murggel:
> > > made the filters "stuck" on top of the page again - if we decide to "unstick" them again it's a simple commit to revert :-)
> > > 
> > > This should be ready to go now.
> > 
> > Am I petty, if the count could also stuck on top (also for the page with all logs)?
> 
> Wouldn't want to do it as part of this PR , let's discuss in an separate issue.
> 
> outside of this PR: imho there's no value having this sticky on the screen.
> 

Agree

murggel:
> > For waypoints .... Space however is a good reason against it, at least on smaller devices ...
> > I was contemplating whether we could have just 1 button in that dialog - "my location" is available from the "add wp"dialog - that'd again save some space.
> 
> That is something, what @ReggieDoodle also suggests somewhere (in his PR with the new Coordinates dialog ?)

Ah, it was @ztNFny  suggesting it in  https://github.com/cgeo/cgeo/issues/17229#issuecomment-3148103357

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
```diff
@@ -26,6 +26,8 @@
 import androidx.annotation.NonNull;
 import androidx.appcompat.widget.TooltipCompat;
 
+import com.google.android.material.chip.Chip;
```
bekuno: This is the Checkstyle error:
~~~
           <property name="groups" value="cgeo,android,androidx,java,javax,*" />

~~~
 So, the `com` is wrong between `androidx` and `java`.

ztNFny: strange, i didn't touch any of the import settings in intellij, so why would it be wrong? Anyway, pushed again now, let's hope this fixes it.

moving-bits: IntelliJ/AS has the habit of auto-inserting new imports in places different to what our CeckStyle rules expect - not sure if that could be synced - @bekuno?

eddiemuc: Would be great if AS could pick this up automatically from checkstyle config somehow. Is there a way?

I had to set c:geo's import rules manually in AS config. (c:geo rules are strangely different from any other project I work in)

ztNFny: I don't think there's any checkstyle connect it can do. I've configured the order manuelly in AS in Settings, Editor, Code Style, Java, Imports, Import Layout

ztNFny:
failing test (CgeoApplicationTest.testSearchByCoords) is unrelated

main/src/main/java/cgeo/geocaching/log/CacheLogsViewCreator.java
bekuno: > Would be great if AS could pick this up automatically from checkstyle config somehow. Is there a way?

We have in the project the file `<cgeo>\.idea\checkstyle-idea.xml`.
The default value is set to "cgeo-checks".
If you don't change this and add to your personally git ignore file, then you have this active.
The AS checkstyle should then use this configuration.

bekuno: > c:geo rules are strangely different from any other project I work in

This rules are historical.
Of cause we can change it, if this makes sense. 
I fear that this will create many changes. (Probably only for changes files in PR)

eddiemuc: > > Would be great if AS could pick this up automatically from checkstyle config somehow. Is there a way?
> 
> We have in the project the file `<cgeo>\.idea\checkstyle-idea.xml`.
> The default value is set to "cgeo-checks".
> If you don't change this and add to your personally git ignore file, then you have this active.
> The AS checkstyle should then use this configuration.

I was referring to the AS import optimize setting (Settings, Editor, Code Style, Java, Imports, Import Layout). The ones @ztNFny referred to above

bekuno: > > > Would be great if AS could pick this up automatically from checkstyle config somehow. Is there a way?
> > 
> > 
> > We have in the project the file `<cgeo>\.idea\checkstyle-idea.xml`.
> > The default value is set to "cgeo-checks".
> > If you don't change this and add to your personally git ignore file, then you have this active.
> > The AS checkstyle should then use this configuration.
> 
> I was referring to the AS import optimize setting (Settings, Editor, Code Style, Java, Imports, Import Layout). The ones @ztNFny referred to above

We discussed it in the meeting today. See Issue #17287

eddiemuc:
Time to test this in nightlies

PR #17264, opened 2025-08-05
support disabling items in SimpleItemList

ztNFny:
image

follow-up to https://github.com/cgeo/cgeo/issues/17169

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
+++ b/main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
@@ -57,6 +57,7 @@ public class SimpleItemListModel<T> {
     private String filterTerm = null;
 
     private final Set<T> selectedItems = new HashSet<>();
+    private final Set<T> disabledItems = new HashSet<>();
 
     private final List<Consumer<ChangeType>> changeListeners = new ArrayList<>();
 
@@ -472,6 +473,23 @@ public class SimpleItemListModel<T> {
      *  collection will always have 0 or 1 element only. */
     public Set<T> getSelectedItems() {
         return selectedItems;
+    }
+
+    /** Sets the currently selected items */
+    public SimpleItemListModel<T> setDisabledItems(final Iterable<T> disabled) {
+        if (disabled != null) {
+            disabledItems.clear();
+            for (T sel : disabled) {
+                disabledItems.add(sel);
+            }
+        }
+        return this;
+    }
+
+    /** Gets currently selected items. If ChoiceMode is a SINGLE mode then the returned
+     *  collection will always have 0 or 1 element only. */
+    public Set<T> getDisabledItems() {
+        return disabledItems;
     }
 
     /** Adds a model change listener */
--- a/main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
+++ b/main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
@@ -10,11 +10,13 @@ import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.functions.Func5;
 
 import android.content.Context;
+import android.graphics.Paint;
 import android.util.AttributeSet;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.LinearLayout;
+import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -270,6 +272,20 @@ public class SimpleItemListView extends LinearLayout {
         @Override
         public void onBindViewHolder(@NonNull final ItemListViewHolder holder, final int position) {
             holder.fillData(getItem(position));
+
+            if (model.getDisabledItems().contains(getItem(position).value)) {
+                for (int i = 0; i < holder.binding.itemViewAnchor.getChildCount(); i++) {
+                    final View child = holder.binding.itemViewAnchor.getChildAt(i);
+                    if (child instanceof TextView) {
+                        final TextView tv = (TextView) child;
+                        tv.setPaintFlags(tv.getPaintFlags() | Paint.STRIKE_THRU_TEXT_FLAG);
+                        tv.setTextColor(getResources().getColor(R.color.colorTextHint));
+                    }
+                }
+                holder.binding.itemCheckbox.setEnabled(false);
+                holder.binding.itemRadiobutton.setEnabled(false);
+                return;
+            }
 
             holder.binding.getRoot().setOnClickListener(v -> handleClick(holder.getBindingAdapterPosition()));
             holder.binding.itemChecker.setOnClickListener(v -> handleClick(holder.getBindingAdapterPosition()));
--- a/main/src/main/java/cgeo/geocaching/connector/gc/BookmarkUtils.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/BookmarkUtils.java
@@ -17,6 +17,7 @@ import android.content.Context;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class BookmarkUtils {
     private static final String NEW_LIST_GUID = "FakeList";
@@ -57,6 +58,7 @@ public class BookmarkUtils {
             model
                 .setItems(lists)
                 .setDisplayMapper((l) -> TextParam.text(l.getName() + (NEW_LIST_GUID.equals(l.getGuid()) ? "" : " (" + l.getCaches() + ")")))
+                .setDisabledItems(lists.stream().filter(l -> l.getCaches() >= 1000).collect(Collectors.toSet()))
                 .setChoiceMode(SimpleItemListModel.ChoiceMode.SINGLE_PLAIN);
 
             SimpleDialog.ofContext(context).setTitle(R.string.search_bookmark_select)
--- a/main/src/main/java/cgeo/geocaching/connector/gc/BookmarkUtils.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/BookmarkUtils.java
@@ -56,7 +56,7 @@ public class BookmarkUtils {
             final SimpleDialog.ItemSelectModel<GCList> model = new SimpleDialog.ItemSelectModel<>();
             model
                 .setItems(lists)
-                .setDisplayMapper((l) -> TextParam.text(l.getName()))
+                .setDisplayMapper((l) -> TextParam.text(l.getName() + (NEW_LIST_GUID.equals(l.getGuid()) ? "" : " (" + l.getCaches() + ")")))
                 .setChoiceMode(SimpleItemListModel.ChoiceMode.SINGLE_PLAIN);
 
             SimpleDialog.ofContext(context).setTitle(R.string.search_bookmark_select)


ztNFny:
The warning is already in https://github.com/cgeo/cgeo/pull/17244

Hiding those "invalid" ones will probably get a lot of "why is it missing" questions. The list is there but not a valid target, so making it "gone" seems like bad UX.

eddiemuc:
I think you can solve this w/o adding generic "disabled" support (which ianbtw hard to achieve, your implementation would only do it for simole textview lists).

- design the display mapper such that it paints the nonvalid items in a disabled fashion to your liking
- when a disabled entry is clicked then reopen the whole dialog
(As an alternative to the second, there might even be the possibility to register a custom click handler to use which doesn't auto-close the dialog. I would need to look this up in code though.)

ztNFny:
Disabled items are already unclickable. The only actual gap I see is the "select all", should br easy to solve that one.

eddiemuc:
... And generalizing "disabled look" for virws other than pure text.

What about grouping? Should group headers also look disabled when their children are all disabled?

Imho its far easier to adapt the displaymapper and clicklistenrr for this specific dialog. Or are there other use cases for this which justify a generalization?

ztNFny:
Where is it used without text? Will look into a possible visualization based on that. 

A group should be disabled if it's set to disabled - but we can revisit that once there's a usecase for it and decide based upon an actual scenario.

What's the con of having a general implementation? The only missing point you mentioned will be easily fixed, so why have a special case when it can be covered genetically.

eddiemuc:
> Where is it used without text? Will look into a possible visualization based on that.

For example for cache selection eg from click on map. They are displayed in their own list item view (with header row, details row and the cache icon)

> A group should be disabled if it's set to disabled - but we can revisit that once there's a usecase for it and decide based upon an actual scenario.

valid

> What's the con of having a general implementation? The only missing point you mentioned will be easily fixed, so why have a special case when it can be covered genetically.

Nothing against a generalization. It's just a **lot** of work and testing to get all cases out. Just for your use case alone it seems oversized. I probably also don't think of all cases right now, it would be part of the work to carefully analytzing SimpleDialog as a whole to see what can go bunkers there. (and of course SimpleDialog itself for sure has its own share of hidden bugs + ba stuff in it that we haven't just not found yet)

Just looked at code. To handle  clicks without dialog closing you can use `ItemSelectModel.setButtonClickedListener` to register a free-style handler for clicks and simply NOT pass a handler when calling `selectSingle` (pass a null value here). The handler passed with`ItemSelectModel.setButtonClickedListener` will NOT close the dialog. In the handler you simply do nothing when a disabled item was hit and do your logic (plus closing the dialog) when a non-disabled item was hit.

I only realize that there IS a bug in SimpleDialog in that  the `buttonclicklistener ` is not yet handled in direct select mode (indeed SimpleDialog has its share of bugs). I fixed that in PR #17265 just now. That's a good fix to have anyway, even if you decide to extend SimpleDialog for generic diasbled handling.

ztNFny:
> I believe your change would not work for a multi-select scenario with "select all" option

done

> For example for cache selection eg from click on map.

image

> To handle clicks without dialog closing 

considering the item is disabled i just didn't attach any click-handler on it at all

> indeed SimpleDialog has its share of bugs

just realized that "select all" also selects the "create new list" option - which imho should be treated as a special option and not be included in the selection

eddiemuc:
Need more time for this one. Want to try handling of groups as well as dialogs with checkboxes and radiobuttons. Maybe it would be a good idea to have some "test dialogs" to quickly check features.

Quick glance: it seems "select visible" is not handled?

eddiemuc:
Test todos for disabled items as a generalised feature (as a reminder for myself, feel free to enhance):
- [x] verify "select visible" / "select all"
- [x] handling of group headers (tbd: possibly neglect for now, but maybe its simple to add?)
- [x] handling of multiselect + radio-select (+ single select, but that one is already tested)
- [ ] multiselect: handling pre-selecting values which are disabled
- [x] action button on disabled items
- [x] more complex item layouts (probably not necessary since already tested, see screenshot above)
- [ ] ... probably there's more

Questions: 
* "disabled" feature is the first one introducing a hard-coded view style into SimpleDialog. Is this wise? What if in some contexts "disabled" status of items should be displayed in a different way? Maybe abstract out a method to set a list item to "disabled" and just have a default implementation?
* current "default" disabled style has similarities to non-disabled content in other parts of c:geo; specifically displaying disabled caches in cache list is visually similar to(strike-though text), but those items are not disabled in the sense of SimpleDialog (they are and should be still tappable). Maybe find another more generic "disabled" style?

ztNFny:
> Quick glance: it seems "select visible" is not handled?

done

> handling of group headers (tbd: possibly neglect for now, but maybe its simple to add?)

imho nothing required there - a group can't be disabled

> action button on disabled items

it doesn't do anything due to there not being any listener attached

> current "default" disabled style has similarities to non-disabled content in other parts of c:geo

I've changed the style to alpha(0.5f) and strikethrough. It is not very different to the style of "disabled cache" of course as there aren#t many ways to express a disabled state - naturally.

main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
```diff
@@ -553,4 +567,18 @@ public int getSpanSize(final int pos) {
            });
         }
     }
+
+    private List<TextView> getAllTextViews(final View view) {
+        final List<TextView> textViews = new ArrayList<>();
+        if (view instanceof TextView) {
```
eddiemuc: Maybe use `ViewUtils.walkViewTree` here to safe some code lines?

eddiemuc:
I tried out your changes by extending the dialog examples (see attached patch file, I can commit this to your branch if you like?). You reach the test dialogs via Settings -> System -> GUI Test Screen (at very bottom).
[add-example.patch](https://github.com/user-attachments/files/22135632/add-example.patch)

I saw two strange effects:

PREselected items will be selected, regardless of them being disabled:

image

Also, for some reason, some random not-disabled items are displayed disabled. Although when I click on them they are "un-disabled":

image

ztNFny:
> PREselected items will be selected, regardless of them being disabled:

the to-be behaviour here depends a lot on the specific usecase I guess. I can think of several usecases each for both possible behaviours (allowing and not allowing pre-selection) right away. E.g. for the "add to bookmarks" scenario it could show that the item is already on a bookmark list but that list is full and new items can't be added.
I'd propose to leave some gap here until a real usecase is encountered. Or we introduce a "allowDisabledToBePreselected(boolean)" from the start - question is whether this cleaness in design will ever be needed/used - like do we need to cover 100% of possible usecases or 100% of those we actually use.

> Also, for some reason, some random not-disabled items are displayed disabled. Although when I click on them they are "un-disabled":

will check that

eddiemuc:
One more thing I found: when "disabled item" selection is changed after dialog is displayed programatically (eg as a reaction of a button click), then the dialog is not updated.

> the to-be behaviour here depends a lot on the specific usecase I guess. I can think of several usecases each for both possible behaviours (allowing and not allowing pre-selection) right away. E.g. for the "add to bookmarks" scenario it could show that the item is already on a bookmark list but that list is full and new items can't be added.
> I'd propose to leave some gap here until a real usecase is encountered. 

I agree with your assessment, but not with your solution proposal. If we want to build this specific to the bookmark scenario, then there would be no need to add a generic disabled-item-handling into the SimpleDialog. It would then much easier to implement the behavior for BookmarkUtils only.  That would be my counterproposal.
If it becomes part of the SimpleDialog class, then I insist that the solution is implemented in a generic way.

ztNFny:
> If it becomes part of the SimpleDialog class, then I insist that the solution is implemented in a generic way.

Then it'd say it's ready - it's implemented in a generic way: Disabled items can be selected by code. That's generic and it's not wrong. There might be other usecases, but those would be special then.

eddiemuc:
I disagree. It would be ready if the feature is combinable with the other features of SImpleDialog, which it is currently not (see my findings).

ztNFny:
> I disagree. It would be ready if the feature is combinable with the other features of SImpleDialog, which it is currently not (see my findings).

as stated I'll look into your second point about "wrong" items being selected.
For the "marking disabled items as selected" it's imho a perfectly valid behaviour. If you want to define it differently ... feel free. The current behaviour is generic and working.

eddiemuc:
> > I disagree. It would be ready if the feature is combinable with the other features of SImpleDialog, which it is currently not (see my findings).
> 
> as stated I'll look into your second point about "wrong" items being selected. 

ok, then we are aligned

> For the "marking disabled items as selected" it's imho a perfectly valid behaviour. If you want to define it differently ... feel free. The current behaviour is generic and working.

ok that is indeed discussable.

Dont forget this finding though:  when "disabled item" selection is changed after dialog is displayed programatically (eg as a reaction of a button click), then the dialog is not updated.
(this should be easily fixable anyway, I assume you just have to throw a notification in the setter)

ztNFny:
There's a general wonkyness of that test dialog (not sure whether it's that test dialog or the underlying SimpleItem dialog implementation) on scrolling, items get "intermingled":
image
image

Those are not related to disabled-support, it also happnes before any of my commits:
image

eddiemuc:
> There's a general wonkyness of that test dialog (not sure whether it's that test dialog or the underlying SimpleItem dialog implementation) on scrolling, items get "intermingled"...

Thanks my man, you fond a bug indeed.

The underlying problem is that in RecyclerView Item-Views are reused. And TextParam did not properly "reset" already used TextViews. Thus reusing TextViews which contained images before would still show those images.  I created PR #17408 to fix it.

The same it probably the cause of the "disabled"-items appearing in the list BTW. To solve this you probably need to "remove" your "diabled"-formatting from items every time an item is NOT disabled (probably the majority :-))

ztNFny:
should be fixed. some icon issue still exists, see #17408

main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
```diff
@@ -271,6 +274,33 @@ private void handleClick(final int adapterPos) {
         public void onBindViewHolder(@NonNull final ItemListViewHolder holder, final int position) {
             holder.fillData(getItem(position));
 
+            if (!model.getDisabledItems().isEmpty()) {
```
eddiemuc, line 277: This might probably not work when a list first has disabled items, then later all disabled items are removed (programatically). Not entirely sure myself though.

main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
```diff
@@ -475,6 +476,23 @@ public Set<T> getSelectedItems() {
...
     }
 
+    /** Sets the currently selected items */
+    public SimpleItemListModel<T> setDisabledItems(final Iterable<T> disabled) {
+        if (disabled != null) {
+            disabledItems.clear();
+            for (T sel : disabled) {
+                disabledItems.add(sel);
+            }
+        }
```
eddiemuc, line 486: I am almost sure you need to fire a Change Event here, so list is re-rendered if disabled items are changed after  dialog is already shown (eg programatically after button tap or something like this). "triggerChange(ChangeType.COMPLETE)" should do it (will re-render the whole table, but I assume disabled is not used often enough for special handling).

You can try this out eg by calling "setDisabledItems" AFTER "selectSingle/Multiple" was called

main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
```diff
@@ -271,6 +274,33 @@ private void handleClick(final int adapterPos) {
...
+                final boolean isEnabled = !model.getDisabledItems().contains(getItem(position).value);
+
+                ViewUtils.walkViewTree(holder.binding.itemViewAnchor, vi -> {
```
eddiemuc, line 280: The logic to make a view look "disabled/enabled" could probably be extracted into a Utility method (eg into ViewUtils) so it can be used by other places as well

main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
ztNFny: dialog.selectMultiple(model, item -> display(ctx, "Selected", "Selected: " + item));
                model.setDisabledItems(disabledItems);

gives the correct result

eddiemuc: Try this:
`ViewUtils.runOnUIThread(true, () -> model.setDisabledItems(disabledItems));`

main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
eddiemuc: Did you test this?

main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
ztNFny: Also gives the correct result

main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
ztNFny: Yes

ztNFny:
@eddiemuc can this be merged?

eddiemuc:
> @eddiemuc can this be merged?

Sure. Sorry for the wait

Issue #17261, opened 2025-08-04
Rename "bearing" to "projection" in the context of waypoint projection


SammysHP:
We got a (legitimate) [complaint](https://www.geoclub.de/forum/t/c-geo-aktuelle-bugs.67870/post-1412467) about the German translation "Peilung" for a waypoint projection. Taking a bearing is the process of finding the direction of a target, while projection describes finding a point based on direction and distance.

The wrong word is used throughout the code, for example in the translations and even here:

https://github.com/cgeo/cgeo/blob/db90120e35e5eccbfd3ec96a368e5c502ba4b934/main/src/main/java/cgeo/geocaching/enumerations/ProjectionType.java#L15-L20

SammysHP:
On the other hand, you could also understand it as "a projection based on a (bearing|offset)", which means the current wording could be valid as well…

toolfish:
A bearing is a bearing and a projection is a projection.

SammysHP:
@toolfish So what do you suggest? Please submit a pull request if possible.

ztNFny:
at least in german listings "Peilung" feels like the more common term. If the listing says to do a "Peilung" then people will know what to do, if it says to do a "Wegpunktprojektion" they probably won't. So while it technically might not be correct I feel it's the more user-friendly way.

ztNFny:
> at least in german listings "Peilung" feels like the more common term. If the listing says to do a "Peilung" then people will know what to do, if it says to do a "Wegpunktprojektion" they probably won't. So while it technically might not be correct I feel it's the more user-friendly way.

just searched all the listings of caches stored in c:geo:
37 results for "Projektion", but a quick check tells me most use it in another meadning, eg GC3JXJP
340 results for "Peilung"
297 results for "peile"

toolfish:
Yes but still wrong. Better teach the users to use correct wording. "Peilung" is the opposite of "Projektion".

ztNFny:
Per dictionary it is wrong, yes. But per real world usage it's the term that people actually use and understand.
Duden made "gleiche" and "selbe" equal because real world use didn't align with the dictionary decision.

Imho the goal of c:geo should be to enable the user to do something, not to teach them about Duden. By renaming it to a term that most users don't know it'd effectively make this function unavailable for them.

SammysHP:
Considering the usage of the word and the complexity of changing it, let's close this issue.

PR #17268, opened 2025-08-06
Update ProjectionType.java

toolfish:
changed "bearing" to "projection" since a bearing is not a projection (which is done here)

## Description
See https://www.geoclub.de/forum/t/c-geo-aktuelle-bugs.67870/post-1412467


ztNFny (changes requested):
This PR is broken in multiple ways:
- Changing the ENUM name without changing its usages
- Referenced to undefined strings.

To change the user display text only the string would need to be changed, the internals shouldn't be touched.

As for the intent of the change itself there's still an ongoing discussion in #17261

eddiemuc:
I doubt that this PR compiles (due to reasons mentioned by @ztNFny ). Also discussion seems unconcluded. Will close here for now.

PR #17274, opened 2025-08-07
add an option to not change the size of images loaded from GC

ztNFny:
GC has 5 sizes for each image and we currently have a logic for the user to select to load a specific one of those (actually not for thumb as that one is tiny and of no use). However there's currently no option to NOT rewrite the request but instead load whatever the listing owner used when creating the listing.

Also setting that new "unchanged" size as default as that's what you get with any other app.

### DIFF
--- a/main/src/main/res/values/preference_keys.xml
+++ b/main/src/main/res/values/preference_keys.xml
@@ -490,11 +490,13 @@
     </string-array>
 
     <!-- Geocaching.com => image size -->
+    <string translatable="false" name="pref_value_gc_imagesize_unchanged">UNCHANGED</string>
     <string translatable="false" name="pref_value_gc_imagesize_original">ORIGINAL</string>
     <string translatable="false" name="pref_value_gc_imagesize_large">LARGE</string>
     <string translatable="false" name="pref_value_gc_imagesize_display">DISPLAY</string>
     <string translatable="false" name="pref_value_gc_imagesize_small">SMALL</string>
     <string-array name="pref_gc_imagesize_values">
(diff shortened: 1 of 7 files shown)


main/src/main/res/values/preference_keys.xml
```diff
@@ -490,11 +490,13 @@
     </string-array>
 
     <!-- Geocaching.com => image size -->
+    <string translatable="false" name="pref_value_gc_imagesize_unchanged">UNCHANGED</string>
```
eddiemuc, line 493: I don't quite unterstand what those values are for?

eddiemuc:
Took a brief look at the code. To my understanding the problem arises because the original implementation chose to hold translations for different enum values in an array and thus tied enum value positioning to translated erro positioning. This leads exactly to the problem you are experiencing right now.

I suggest to change l18n pattern towards the enum translation pattern used all over the rest of c:geo eg CacheType, CacheSize and so on. This pattern parameterizes each enum value with a stringId and provides a method getL10n or getUserDisplayableString etc which provides the translated value. The translated values are NOT part of an array in strings.xml but are separate values.

In this concrete situation unfortunately this leads to having to retranslate the values a second time. But you're safe for the future if there's ever the need to change the valuelist again.

Example CacheSize:

Class CacheSize.java

image

image
(would suggest to use LocalizationUtils here, but thats another thing)

strings.xml:

image

ztNFny:
reworked to an enum with StringRes pattern

eddiemuc (approved):
Lgtm

Issue #17157, opened 2025-07-20
targetSDK 35: Graphical glitches in newer nightlies


Lineflyer:
I updated from a rather old nightly (some weeks old) to 2025.07.19-NB-2a580a3 and noticed some differences and glitches:

- Some screens do now have bad formatting. Need to check in details which are affected, but foremost I noticed it in the variable edit screen (when entering data into the variable values). The text fields are cropped and smaller as the text. 
- Some texts in screens seems to have changed their font sizes, e.g. the tab names at the bottom of the screen on cache detail view got larger than before. 
- Some paddings changed, e.g. the distance between icon and text at the main page bottom bar.

I assume this might be due to some recent generic change (e.g. target API??).

I did not follow all issues and latest development actions, thus I don't know if this is already worked on.
Please tell me, whether I shall open dedicated issues for all findings mentioned above.

moving-bits:
thanks, for now I just put a "targetSDK 35" in front of the title - hopefully there won't be so many glitches that we need a separate label for it...

MagpieFourtyTwo:
Thanks for creating the issue - wanted to do so for a few days already (mentioned it in #17135), but my constitution did not let me.
However, here's what I've observed:

- seemingly all input fields have no padding anymore (pretty strange for text inputs and even worse on edits, cause it can be hard to distinguish letters from the border lines)
- fields which should get a value are often overlapped by others
- some buttons look pretty strange (text not centered (example see pic in #17135)
- esp. the waypoint note edit field, the edit log text window and the input windows of signature templates do not obey the keyboard overlay anymore, the written text disappears behind the keyboard and is nearly impossible to edit (if it has >2 lines), cause there's no possibility to scroll the underlying window upwards.

Emulators seem to still show an "intact" representation, and some other "real" devices do as well, thus some screenshots to illustrate.

Waypoint Note with keyboard active and without it:
![Image](https://github.com/user-attachments/assets/093f2115-f307-4b4c-a534-171307a726c5)   ![Image](https://github.com/user-attachments/assets/b0be940f-77a4-4440-b869-975af858674f)

Waypoint edit page with various layout glitches:
![Image](https://github.com/user-attachments/assets/d5e2f9a1-b776-4b64-83ae-200019a29f04)   ![Image](https://github.com/user-attachments/assets/faabf519-c26b-434a-8481-cac6fbd547e1)

[...]

Issue #17151, opened 2025-07-20
TargetSDK 35: Map action bar padding glitch


SammysHP:
### Describe your problem!

After updating to today's nightly, the action bar on the map has a large padding. Entering full screen mode doesn't fix the problem, the additional padding stays visible.

### How to reproduce?

Open map (tested with unified Google and VTM on- and offline).

### Reproducible

Yes

### c:geo Version

2025.07.20-NB-2c98699

### System information

```text

```

### Additional Information

Was working with 2025.07.13

SammysHP:
I also noticed that the bottom bar is flickering when switching between activities. Not sure if this is something new.

SammysHP:
![Image](https://github.com/user-attachments/assets/b735f5b1-9763-4ebb-b2e2-0ce81cb70d73)

moving-bits:
Does your system run Android 15?
Do you have "fullscreen on tap" enabled for the map? ("Toolbar beim Tippen ausblenden")

(c:geo is now targeting Android 15, which comes with a couple of changes, one of it the enforcement of full-screen mode. Probably our opt-out is incomplete.)

SammysHP:
> Does your system run Android 15?

Yes.

> Do you have "fullscreen on tap" enabled for the map? ("Toolbar beim Tippen ausblenden")

Also affirmative. Disabling "fullscreen on tap" fixed the problem.

SammysHP:
Looking at the screenshot above, it also looks like there's less padding at the bottom of the screen.

In addition, the map has now a partially hidden "unified map" tag in the corner. Never seen it before, although I was using unified map all the time.

![Image](https://github.com/user-attachments/assets/775614ee-07de-4c91-9f82-171f910e6435)

The view pager indicator in the cache details is also missing padding.

![Image](https://github.com/user-attachments/assets/e1a3e949-d2c4-4d53-b0a0-82f8f010bda8)

Should I rename this issue to collect targetSdk=35 issues or should I open individual issues?

ztNFny:
seems that sdk 35 removes all paddings? this is popping up all accross the app

JakeDot:
> Does your system run Android 15? Do you have "fullscreen on tap" enabled for the map? ("Toolbar beim Tippen ausblenden")

I didn't have this problem on Google Pixel 7 with android 16, it only started when using Samsung S22 Android 15. But I think this coincides with the sdk35 change, however I'm not sure about the setting... It sure works now that I disabled it.

[...]

PR #17310, opened 2025-08-17
Related to #17157 - Workaround for UI glitch API35

Lineflyer:
## Description
Found coincidentally, that the feature edge-to-edge enforcement might cause UI problems on Android 15 for some apps.
The indicated workaround was to opt out of this feature on Android 15.

This is most probably not a fix and AFAICS this will not work for Android 16 to opt out.
This means the underlying issue needs to be resolved to make c:geo compatible with this edge-to-edge feature

I did quickly run it on an emulator and the UI glitches in text field paddings are gone.
Could not check the map top padding as this problem was not visible in the emulator even before.

## Related issues
#17157 and maybe #17151

## Additional context
Did not record the precise source for this workaround, however this might be a starting point for further investigation:
https://developer.android.com/about/versions/15/behavior-changes-15?hl=de#edge-to-edge

### DIFF
--- a/main/src/main/res/values-v35/themes.xml
+++ b/main/src/main/res/values-v35/themes.xml
@@ -2,9 +2,7 @@
 <resources>
 
     <style name="cgeo_base" parent="Theme.MaterialComponents.DayNight.DarkActionBar">
-        <item name="android:fitsSystemWindows">true</item>
-        <item name="android:enforceNavigationBarContrast">true</item>
-        <item name="android:enforceStatusBarContrast">true</item>
+        <item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>
     </style>
 
 </resources>


ztNFny:
This workaround is strongly discouraged from. I'd rather not add this workaround as then there's no more motivation at all to fix this - until A16 comes out.

We could always return to this workaround in case time runs out for the Playstore window though as a last resort.

Lineflyer:
Fair enough, at least we do now know, that the "edge-to-edge" is the culprit which needs to be fixed.

ztNFny:
With the deadline getting closer and closer I think we'll have to merge this workaround.
I just spent half a day on this without getting anywhere - can't understand what Google was thinking there. Unless you're using Material 3 this is a very bad mess.

eddiemuc:
There are lint warnings for this PR.

Lineflyer:
> Please remove the other 3 properties, they aren't needed with android:windowOptOutEdgeToEdgeEnforcement

Removed other properties (they were in there already before)

SammysHP:
Are you sure it's safe to remove them? Maybe they are still necessary for older targets.

ztNFny:
> Are you sure it's safe to remove them? Maybe they are still necessary for older targets.

They are in the v35 file, so only applied to target 35.
windowOptOutEdgeToEdgeEnforcement is actually overwriting those values which is why setting them is useless, see https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement

moving-bits:
> > Please remove the other 3 properties, they aren't needed with android:windowOptOutEdgeToEdgeEnforcement
> 
> Removed other properties (they were in there already before)

I had introduced them when setting `targetSDK` to 35 to mitigate some effects on the map, see #17049 (which also links to some further documentation). When opting out of "edge to edge enforcement" for the moment, those attributes can be removed.

But we need to investigate further (maybe even implement edge to edge?) before we can set `targetSDK` to 36 (aka "BAKLAVA"), as then the opt-out flag won't work anymore.

moving-bits:
Deadline is closing in, so let's merge here to get nightlies working fine on Android 15 again (+ be able to create new feature releases in the future)

PR #17312, opened 2025-08-17
fix #17311: Show waypoints for mapType which enableLiveMap, not only for visible caches.

### LINKED ISSUE #17311: [UnifiedMap] Waypoints in offline map are sometimes not shown, if cache is outside the viewport
### Describe your problem!

I have a long multi with multiple stages. 
When zoomed in with other caches on the map, the waypoints of the long multi are sometimes not shown.

### How to reproduce?

I have a long multi with multiple stages. 
1. Open the cache in the map
2. Zoom in, so only some waypoints are visible, but not the cache itself.
3. Switch online mode on and off -> other caches and the waypoints are shown
4. Close map
5. Open map again -> only the other caches are shown, but not the waypoints any more

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2025-08-15-NB-ab65091

### System information

```text

```

### Additional Information

_No response_

murggel:
## Description

Show waypoints for mapType which enableLiveMap, not only for visible caches.
So waypoints in the visible area are shown, even if the cache-coordinates are outside of the considered viewport

## Related issues

fix #17311

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -663,7 +663,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
         final Set<Waypoint> waypoints;
 
         //show all waypoints be displayed or just the ones from visible caches?
-        final boolean showAll = TRUE.equals(viewModel.transientIsLiveEnabled.getValue());
+        final boolean showAll = viewModel.mapType.enableLiveMap();
         if (showAll) {
             waypoints = DataStore.loadWaypoints(viewport);
         } else {
@@ -675,6 +675,7 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                 return wpSet;
             });
         }
+
         //filter waypoints
         MapUtils.filter(waypoints, filter);
         viewModel.waypoints.write(wps -> {


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -680,7 +680,7 @@ public static void refreshWaypoints(final UnifiedMapViewModel viewModel) {
 
         if (viewModel.caches.readWithResult(viewport::count) < Settings.getWayPointsThreshold()) {
             //show all waypoints be displayed or just the ones from visible caches?
-            if (TRUE.equals(viewModel.transientIsLiveEnabled.getValue())) {
```
murggel: I am unsure, whether to keep the old condition or not. 

As far as I understood the code, transientIsLiveEnabled can only be true for mapTypes with enableLiveMap == TRUE.
So that would not be necesssary, but to avoid any risk....

eddiemuc: Your understanding is correct, you can leave out the old condition with your new one.

eddiemuc (approved):
LGTM. Thanks for fixing one of my bugs :-)

eddiemuc:
Can you rebase this to release?

murggel:
> Can you rebase this to release?

Done

PR #17290, opened 2025-08-10
Improve buttons and bevaiour for adding a waypoint

murggel:
## Description

* Use small buttons
* Add name and accuracy to toast for adding current location
* Scroll to added waypoint
* Hide "visited"-chip, if no waypoints are visited

## Related issues

rel #17252

## Additional context

476389063-51c95692-f278-43d3-9222-1a80e23c2003

476389134-894a709d-df09-4138-81d3-bc66e81967df

PR #17313, opened 2025-08-17
Same padding in variables, waypoint and logs-view

murggel:
## Description

Adds consistent padding for variable, waypoint and logs-tab

## Related issues

https://github.com/cgeo/cgeo/pull/17290#issuecomment-3192881612

| Original | Padding for whole layout | Padding for button-groups |
| --- | --- | --- |
| grafik | grafik | grafik |
| grafik | grafik | grafik |
| grafik | grafik | grafik

### DIFF
--- a/main/src/main/res/layout/cachedetail_imagegallery_page.xml
+++ b/main/src/main/res/layout/cachedetail_imagegallery_page.xml
@@ -4,7 +4,6 @@
     android:id="@+id/image_gallery"
     android:layout_width="fill_parent"
     android:layout_height="fill_parent"
-    android:padding="4dip"
     tools:context=".CacheDetailActivity$ImageGalleryCreator">
 
 </cgeo.geocaching.ui.ImageGalleryView>
--- a/main/src/main/res/layout/cachedetail_variables_page.xml
+++ b/main/src/main/res/layout/cachedetail_variables_page.xml
@@ -18,7 +18,6 @@
     <LinearLayout
         android:layout_width="fill_parent"
         android:layout_height="wrap_content"
-        android:layout_marginHorizontal="4sp"
         android:orientation="horizontal">
 
         <Button
@@ -63,6 +62,7 @@
         android:id="@+id/chip_completed_variables"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
+        android:layout_marginHorizontal="7sp"
         style="@style/chip_choice"
         android:text="@string/cache_variables_chip_completed_variables"
         tools:context=".VariablesViewPageFragment" />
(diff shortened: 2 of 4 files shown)


murggel:
> I would propose to go for padding/margin on the buttons instead of a padding on the whole page. As you can see from the last screenshot the proposed solution moves the scrollbar to the left as well.
> 
> Variables page had: android:layout_marginHorizontal="4sp" I'd propose to use that for the buttons on waypoints tab - and probably also apply it to the ChipGroup (if it supports it)

Agree.
For the logs I added a padding anyway, for my liking they were too close to the side

murggel:
@ztNFny I pushed the changes for using padding in LinearLayout of buttons and chip group and not for the whole page.
Please re-review

ztNFny:
> @ztNFny I pushed the changes for using padding in LinearLayout of buttons and chip group and not for the whole page. Please re-review

cherry-picked it and will give it a try on real device 👍🏽

ztNFny:
It seems listing is the view with smallest paddings now 😁

There's one thing (in several views) that looks strange now, the lines:

murggel:
> It seems listing is the view with smallest paddings now 😁
> 

Shall I adapt it as well?

murggel:
> There's one thing (in several views) that looks strange now, the lines: 
> 
> 

I noticed that too, but the same is visble in the original screenshots.

Since the "header" stays always visible, you can interpret the separators hierarchically...

AFAIR it was not so easy to change the shorter ones - so I have to shorten the "header" line, if all should have the same length...

murggel:
@ztNFny maybe we wait until Edge2Edge is done and see how it looks with those changes

eddiemuc:
> @ztNFny maybe we wait until Edge2Edge is done and see how it looks with those changes

Edge2edge will not affect the placement of items toward each other inside a viewa content. Its just about things like navigating bary action bar and its placement beneath system bars, system controls and stuff like this

murggel:
> > It seems listing is the view with smallest paddings now 😁
> 
> Shall I adapt it as well?

I will try to use the smaller padding in the other tabs as well... 

Edit: pushed changes with smaller paddings

ztNFny:
> > There's one thing (in several views) that looks strange now, the lines: 
> > 
> > 
> 
> I noticed that too, but the same is visble in the original screenshots.
> 
> Since the "header" stays always visible, you can interpret the separators hierarchically...
> 
> AFAIR it was not so easy to change the shorter ones - so I have to shorten the "header" line, if all should have the same length...

Ok then, i just looked at the current state closely and that's the thing that stood out.

eddiemuc:
PR lgtm. can it be merged?

murggel:
> Ok then, i just looked at the current state closely and that's the thing that stood out.

@ztNFny so try it in the nightly and probably adapt later?

murggel:
> PR lgtm. can it be merged?

@eddiemuc Let me squash it first and adapt the commit-description

Edit: done

ztNFny:
> > Ok then, i just looked at the current state closely and that's the thing that stood out.
> 
> @ztNFny so try it in the nightly and probably adapt later?

go for it! :-)

Issue #15916, opened 2024-07-27
GCode should be case sensitive


Uli001:
### Describe your problem!

The codes for a adventur and adventure stations are case sensitive. Here's an example:

https://adventurelab.page.link/hY7A
https://adventurelab.page.link/HY7a 

This can lead to problems when importing or exporting. 

### How to reproduce?

Importing a GPX from lab2gpx into GSAK with case sensitive GCCode works. I can import the same GPX into a list in CGeo. Then I go out to cache and write down clues in my notes. After exporting from CGeo, all GCCodes are in capital letters. GSAK recognizes this as a new cache/station.

### Actual result after these steps?

all GCCodes are in capital letters

### Expected result after these steps?

The original GCCode as the GPX should be retained case sensitive.

### Reproducible

Yes

### c:geo Version

2024.07.20

Billy-Wallace:
I can verify that this is an issue. Adventure Lab codes should **Not** be converted to uppercase. There are many examples of Lab stages with the same corecode characters the differ in case (xxxx in LBxxxx01 and ALxxxx).
This gives Groundspeak lots of room for expansion (26 small letters + 26 capital letters + 10 digits)^4 characters)

Additionally, the Codes in GSAK are stored and matched with case-sensitivity. 
When Lab stages are exported from c:geo, they no longer match the same Lab stages in GSAK, leading to incorrect SQL related query results and new waypoints being created if they are imported into GSAK (or any database that correctly stores case-sensitive codes).

JakeDot:
Was this implemented meanwhile? I don't remember recent problems in that area?

moving-bits:
no changes here yet, otherwise this issue would have been closed

Uli001:
Are you planning to do something here? I have just been on a tour in England and am now again faced with the problem of checking 350 LABs so that the databases in GSAK are not destroyed. The databases in CGeo are also affected. A GPX with 25 LABs imported but only 15 arrived in the CGeo. 10 of the LAbs are available with large and small letters. The code is therefore identical for CGeo, which leads to overwriting 10 LABs.

Uli001:
I'll send you a file. Included are 3 LABs or 25 stations.

Download from LAB2GPX
[LCorV701] A Liverpool Georgian Walk : S1 Peter Kavanagh's
...
Import into GSAK 
[LCorV701] A Liverpool Georgian Walk : S1 Peter Kavanagh's
...
Import into CGeo
[...]

Issue #3139, opened 2013-08-17
Search by geocode is case sensitive


Lineflyer:
I am not sure whether this is a regression or was like that all the time??!

2013.08.17-NB

Steps:
- Search for a GC-code which is already stored on your device but use lower case characters

Result:
- Cache is downloaded from network instead directly opened from list
- Status "Saved for offline" is correctly shown, but however the network call is unnecessary

I think the search entry should just be converted to uppercase.

PR #17323, opened 2025-08-22
Consider case sensitive for some connectors (to support imported geocodes)

murggel:
## Description

To consider case sensitive for imported geocodes, convert geocodes to uppercase only for some connectors.

Conditions:
Search should still work and return all possible caches.

## Related issues

rel #15916 
rel #3139

## Additional context

#3139 should still work and load the correct cache.
Unfortunately the search is converted to uppercase... so it works for GC and OC, but currently not for AL


murggel:
Probably this PR and the associated issues are obsolete. All Labs are having now the guid as code, since the firebase-link is no more available.

Lab2gpx already returns the guid (but in a format of `ALhttps://labs.geocaching.com/goto/guid`) - see PR #17388
I am not sure, how GSAK will handle it - see https://gsak.net/board/index.php?showtopic=37091&st=0&#entry285301

murggel:
> Will try to adapt to the new lab2gpx-version see PR [#17388](https://github.com/cgeo/cgeo/pull/17388)

Seems that lab2gpx works fine for c:geo with the latest version.

> How about GSAK I am currently not sure, how this will / can be handled...
-> seems they have a [solution](https://gsak.net/board/index.php?showtopic=35689&st=40&#entry285272)

PR #17329, opened 2025-08-24
LAB url

ztNFny:
Lab API has changed and is no longer returning a https://adventurelab.page.link/ link but instead using https://labs.geocaching.com/goto/

From the API (not the lab one) doc: "The firebase dynamic link returned for adventures is obsolete. Deeplink is now being returned and should be used instead."

This PR changes the AL parser to use the DeepLink property as source (it's value is currently identical to FirebaseDynamicLink, so no real change, but preparing for future).
It also changes the CACHE_URL to https://labs.geocaching.com/goto/

This will (probably) break existing, offline-stored caches! The identifier for the old links was a different one than for the new ones. I can't verify this as I don't have a single lab stored.
Only chance I see to avoid this breaking would be checking whether the cacheid is a GUID (new) or not (old) and select the correct base URL. If someone wants to add that please feel free ;-)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
@@ -329,8 +329,8 @@ final class ALApi {
         try {
             final Geocache cache = new Geocache();
             final JsonNode location = response.at(LOCATION);
-            final String firebaseDynamicLink = response.get("FirebaseDynamicLink").asText();
-            final String[] segments = firebaseDynamicLink.split("/");
+            final String deepLink = response.get("DeepLink").asText();
+            final String[] segments = deepLink.split("/");
             final String geocode = ALConnector.GEOCODE_PREFIX + response.get("Id").asText();
             cache.setGeocode(geocode);
             cache.setCacheId(segments[segments.length - 1]);
@@ -364,8 +364,8 @@ final class ALApi {
         try {
             final Geocache cache = new Geocache();
             final JsonNode location = response.at(LOCATION);
-            final String firebaseDynamicLink = response.get("FirebaseDynamicLink").asText();
-            final String[] segments = firebaseDynamicLink.split("/");
+            final String deepLink = response.get("DeepLink").asText();
+            final String[] segments = deepLink.split("/");
             final String geocode = ALConnector.GEOCODE_PREFIX + response.get("Id").asText();
             final String ilink = response.get("KeyImageUrl").asText();
             final String desc = response.get("Description").asText();
--- a/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/al/ALConnector.java
@@ -32,7 +32,7 @@ import org.jetbrains.annotations.NotNull;
 public class ALConnector extends AbstractConnector implements ISearchByGeocode, ISearchByFilter, ISearchByViewPort {
 
     @NonNull
-    private static final String CACHE_URL = "https://adventurelab.page.link/";
+    private static final String CACHE_URL = "https://labs.geocaching.com/goto/";
 
     @NonNull
     protected static final String GEOCODE_PREFIX = "AL";


murggel:
> This will (probably) break existing, offline-stored caches! The identifier for the old links was a different one than for the new ones. I can't verify this as I don't have a single lab stored. Only chance I see to avoid this breaking would be checking whether the cacheid is a GUID (new) or not (old) and select the correct base URL. If someone wants to add that please feel free ;-)

I can try to add the check .... Shall I push to your PR or wait until merged and create a new PR?

ztNFny:
@murggel go ahead.

I just thought of something... If for old caches we already stored the GUID it's probably better to use that one together with the /goto/ link instead of returning either old or new link depending on stuff.
So always return /goto/ + getGuid().

Could you please verify this on an old stored one? My dev machine is already powered down.

murggel:
> @murggel go ahead.
> 
> I just thought of something... If for old caches we already stored the GUID it's probably better to use that one together with the /goto/ link instead of returning either old or new link depending on stuff. So always return /goto/ + getGuid().
> 
> Could you please verify this on an old stored one? My dev machine is already powered down.

We don't store the "guid" in the database for ALs, so `getGuid` unfortunately returns an empty string...
 
I just tried it with stored and unstored ones - unfortunately there are labs for which the API does not return a GUID for the deep link e.g. https://labs.geocaching.com/goto/ALSVRTT
so checking for GUID and return deprecated or new URL does not work

So I guess we have to go / stay with breaking already stored once - or we just store the GUID (= cacheid) now and can distinguish with that...

I will try, if that works...

murggel:
> So I guess we have to go / stay with breaking already stored once - or we just store the GUID (= cacheid) now and can distinguish with that...
> 
> I will try, if that works...

Seems to work - so I have to check, where the GUID of the cache is used, so saving the id as guid does not break something else ...

ztNFny:
Goto links are actually NOT using the guid but another guid ... So i guess this PR is as good as it gets ...

murggel:
> Goto links are actually NOT using the guid but another guid ... So i guess this PR is as good as it gets ...

The field "guid" is empty for AdventureLabs - so we can store the "other guid" there or the "real guid" of the lab / doesn't matter, if we just use that field to distinguish, whether it is an old stored lab or a new stored one...

murggel:
> > Goto links are actually NOT using the guid but another guid ... So i guess this PR is as good as it gets ...
> 
> The field "guid" is empty for AdventureLabs - so we can store the "other guid" there or the "real guid" of the lab / doesn't matter, if we just use that field to distinguish, whether it is an old stored lab or a new stored one...

So probably storing the "real guid" makes more sense. But then it is kind of an abuse for checking which url to use?

ztNFny:
Id and goto both use guids but they are not related. Can't use that info unfortunately. We just have to use the new goto url from now on.

murggel:
I think you missunderstood me - I don't want to use the ID as guid inside the link - I want to store the ID as guid-member of the cache. For old caches we didn't store any guid, so we can distinguish whether to use the old url for labcaches without a stored guid and the new goto for labcaches with stored guid (but using the guid from the deep link for the url)

see https://github.com/cgeo/cgeo/commit/9a6f312abbd933456ab6612530f1f1bac9899e0c
(I created a own branch for showing the changes I have in mind - I can push the last commit to your branch)

murggel:
> > > Goto links are actually NOT using the guid but another guid ... So i guess this PR is as good as it gets ...
> > 
> > 
> > The field "guid" is empty for AdventureLabs - so we can store the "other guid" there or the "real guid" of the lab / doesn't matter, if we just use that field to distinguish, whether it is an old stored lab or a new stored one...
> 
> So probably storing the "real guid" makes more sense. 

And then we have the guid as real member and not hidden in the geocode (makes it probably also easier in future)

eddiemuc:
I suggest to merge here and continue further discussions and changes in another issue/PR

ztNFny:
> I think you missunderstood me - I don't want to use the ID as guid inside the link - I want to store the ID as guid-member of the cache. For old caches we didn't store any guid, so we can distinguish whether to use the old url for labcaches without a stored guid and the new goto for labcaches with stored guid (but using the guid from the deep link for the url)
> 
> see [9a6f312](https://github.com/cgeo/cgeo/commit/9a6f312abbd933456ab6612530f1f1bac9899e0c) (I created a own branch for showing the changes I have in mind - I can push the last commit to your branch)

So you basically want to use "isPresent(guid)" as a marker whether to use old or new URL? While I understand the approach this "abuse" seems "not nice"... Also wouldn't it cause problems with labs being stored right now? Those are be stored with the goto URL info about according to the isPresent(guid) logic would use the page.link URL.

I think it's not possible to solve this without "abusing" something - but wouldn't it maybe be better to use the stored-date then? That'd allow us to use the new URL from the point in the API change, avoiding the gap-period - and it's not more or less ugly than the guid approach.
The change was implemented on 19.8.2025 (not sure on the time)

ztNFny:
I just found this info: " The service will shut down on August 25, 2025."
https://firebase.google.com/docs/dynamic-links

So it seems there's no point in implementing a switch - the service will be unavailable as of today.

ztNFny:
I just found this info: " The service will shut down on August 25, 2025."
https://firebase.google.com/docs/dynamic-links

So it seems there's no point in implementing a switch - the service will be unavailable as of today.

@murggel

murggel:
> I just found this info: " The service will shut down on August 25, 2025."
> https://firebase.google.com/docs/dynamic-links
> 
> So it seems there's no point in implementing a switch - the service will be unavailable as of today.
> 
> @murggel 

Ok, so no fallback possible for already stored labcaches... maybe we need an announcement / topic in FAQ for that, so labcaches have to be refreshed... and probably problems with imported from GSAK / lab2gpx?

baiti:
One of the properties of the `FirebaseDynamicLink` was that if the target application wasn't installed on the user's smartphone, Playstore was automatically launched, pointing to the app in question, offering an immediate install.

Does `DeepLink` also provide these capabilities? I haven't tested it and I don't consider this property really important, however, it would be worth while to find out what happens if such a link is referenced and the app isn't installed. Will that throw some error? Should we somehow programmatically check whether tha Lab App is installed and if it is not, explicitly reference the app via Playstore? Sort of emulating the behaviour of `FirebaseDynamicLink` then?

PR #17355, opened 2025-08-30
Adapt c:geo to edge2edge enforcement of target API 35

moving-bits:
## Description
First steps in adapting c:geo to the new API 35 edge to edge enforcement. It mimics today's behavior, not making any actual use of edge to edge yet.

Current status (list probably to be extended):
- [x] main screen
- [x] pending downloads
- [x] map (standalone)
  - [x] general appearance
  - [x] filter bar, sort bar
  - [x] distance info
  - [x] filter edit activity
- [x] map (from cache)
  - [x] navigation bar overlap
  - [x] cache title overlap
- [x] settings screens
- [x] cache details: all sub screens

Especially for map view, edge to edge mode might be useful, see the following sample screen:

image

(~Some elements are missing yet (bug), eg. current target (for cache map etc.), distance infos and (maybe) filter bar.~ => implemented meanwhile, screenshot is outdated.)

This is a screenshot with "hide action bar on tap" active, normally the action bar would be shown as well. But IMHO the immersive effect is quite useful on map view, therefore I'd like to propose some further steps here - will create a separate issue for it for further discussion. (Edit: See #17356)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -163,6 +163,7 @@ import androidx.annotation.WorkerThread;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.widget.TooltipCompat;
 import androidx.core.text.HtmlCompat;
+import androidx.core.view.WindowInsetsCompat;
 import androidx.fragment.app.FragmentManager;
 
 import java.lang.ref.WeakReference;
@@ -263,6 +264,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
+        calculateInsets |= WindowInsetsCompat.Type.navigationBars(); // no bottom navigation, need to take navigation bar into account for insets calculation
         super.onCreate(savedInstanceState);
         setThemeAndContentView(R.layout.tabbed_viewpager_activity_refreshable);
 
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -163,7 +163,6 @@ import androidx.annotation.WorkerThread;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.widget.TooltipCompat;
 import androidx.core.text.HtmlCompat;
-import androidx.core.view.WindowInsetsCompat;
 import androidx.fragment.app.FragmentManager;
 
 import java.lang.ref.WeakReference;
@@ -264,7 +263,6 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
-        calculateInsets |= WindowInsetsCompat.Type.navigationBars(); // no bottom navigation, need to take navigation bar into account for insets calculation
         super.onCreate(savedInstanceState);
         setThemeAndContentView(R.layout.tabbed_viewpager_activity_refreshable);
 
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.activity;
 
 import cgeo.geocaching.R;
 import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.utils.Log;
 
 import android.os.Bundle;
 import android.view.View;
@@ -43,6 +44,9 @@ public class AbstractActionBarActivity extends AbstractActivity {
                 if (isEdgeToEdge) {
                     // make room for action bar on top + whatever other insets are requested with calculateInsets variable
                     final ViewGroup activityContent = v.findViewById(R.id.activity_content);
+                    if (activityContent == null) {
+                        Log.e("edge2edge: activityContent not found in " + this);
+                    }
                     if (activityContent != null && !skipActionBarInsetCalculation) {
                         final float actionBarHeight = getResources().getDimension(R.dimen.actionbar_height);
                         final Insets innerPadding = insets.getInsets(calculateInsets);
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
@@ -1,20 +1,57 @@
 package cgeo.geocaching.activity;
 
-import android.os.Bundle;
+import cgeo.geocaching.R;
 
+import android.os.Bundle;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.Window;
+
+import androidx.annotation.NonNull;
 import androidx.appcompat.app.ActionBar;
+import androidx.core.graphics.Insets;
+import androidx.core.view.OnApplyWindowInsetsListener;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.WindowCompat;
+import androidx.core.view.WindowInsetsCompat;
+import androidx.core.view.WindowInsetsControllerCompat;
 
 /**
  * Classes actually having an ActionBar (as opposed to the Dialog activities)
  */
 public class AbstractActionBarActivity extends AbstractActivity {
 
+    protected int calculateInsets = WindowInsetsCompat.Type.statusBars() | WindowInsetsCompat.Type.displayCutout();
+    boolean skipActionBarInsetCalculation = false; // for edge to edge configuration, see onCreate()
+
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         initUpAction();
-    }
 
+        // edge2edge configuration
+        final Window currentWindow = getWindow();
+        final WindowInsetsControllerCompat windowInsetsController = WindowCompat.getInsetsController(currentWindow, currentWindow.getDecorView());
+        windowInsetsController.setSystemBarsBehavior(WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE);
+        ViewCompat.setOnApplyWindowInsetsListener(currentWindow.getDecorView(), new OnApplyWindowInsetsListener() {
+            @Override
+            public @NonNull WindowInsetsCompat onApplyWindowInsets(final @NonNull View v, final @NonNull WindowInsetsCompat insets) {
+                // only apply padding calculations when edge2edge is currently active
+                final Insets systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars());
+                final boolean isEdgeToEdge = systemBars.top == systemBars.bottom; // seems to be no official API for this
+                if (isEdgeToEdge) {
+                    // make room for action bar on top + whatever other insets are requested with calculateInsets variable
+                    final ViewGroup activityContent = v.findViewById(R.id.activity_content);
+                    if (activityContent != null && !skipActionBarInsetCalculation) {
+                        final float actionBarHeight = getResources().getDimension(R.dimen.actionbar_height);
+                        final Insets innerPadding = insets.getInsets(calculateInsets);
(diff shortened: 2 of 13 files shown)


main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+        // edge2edge configuration
+        final Window currentWindow = getWindow();
+        final WindowInsetsControllerCompat windowInsetsController = WindowCompat.getInsetsController(currentWindow, currentWindow.getDecorView());
+        windowInsetsController.setSystemBarsBehavior(WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE);
+        ViewCompat.setOnApplyWindowInsetsListener(currentWindow.getDecorView(), new OnApplyWindowInsetsListener() {
+            @Override
+            public @NonNull WindowInsetsCompat onApplyWindowInsets(final @NonNull View v, final @NonNull WindowInsetsCompat insets) {
+                // only apply padding calculations when edge2edge is currently active
+                final Insets systemBars = insets.getInsets(WindowInsetsCompat.Type.systemBars());
+                final boolean isEdgeToEdge = systemBars.top == systemBars.bottom; // seems to be no official API for this
```
eddiemuc: If I understand this correctly, you want to differentiate between edge-to-edge being active or not. I guess you want to make this differentiation based on the devices SDK version?
--> why not instead enabling edge-to-edge for ALL Android version (down to SDK21). Hence we don't have to support two types of layout in the future

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
         initUpAction();
-    }
 
+        // edge2edge configuration
+        final Window currentWindow = getWindow();
+        final WindowInsetsControllerCompat windowInsetsController = WindowCompat.getInsetsController(currentWindow, currentWindow.getDecorView());
+        windowInsetsController.setSystemBarsBehavior(WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE);
+        ViewCompat.setOnApplyWindowInsetsListener(currentWindow.getDecorView(), new OnApplyWindowInsetsListener() {
+            @Override
+            public @NonNull WindowInsetsCompat onApplyWindowInsets(final @NonNull View v, final @NonNull WindowInsetsCompat insets) {
```
eddiemuc: I believe the "main" work should be done on AbtractActivity so Activites NOT having an ActionBar are also covered.

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         initUpAction();
-    }
 
+        // edge2edge configuration
+        final Window currentWindow = getWindow();
+        final WindowInsetsControllerCompat windowInsetsController = WindowCompat.getInsetsController(currentWindow, currentWindow.getDecorView());
```
eddiemuc: I miss a call to WindowCompat.enableEdgeToEdge() (note: seems to require core lib 1.17)

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+import androidx.core.view.WindowCompat;
+import androidx.core.view.WindowInsetsCompat;
+import androidx.core.view.WindowInsetsControllerCompat;
 
 /**
  * Classes actually having an ActionBar (as opposed to the Dialog activities)
  */
 public class AbstractActionBarActivity extends AbstractActivity {
 
+    protected int calculateInsets = WindowInsetsCompat.Type.statusBars() | WindowInsetsCompat.Type.displayCutout();
```
eddiemuc: ime() probably missing?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+    boolean skipActionBarInsetCalculation = false; // for edge to edge configuration, see onCreate()
+
     @Override
     public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         initUpAction();
-    }
 
+        // edge2edge configuration
+        final Window currentWindow = getWindow();
```
eddiemuc: Some of this logic might be needed in other places as well (e.g. for dialogs). It might make sense to move it to a utils class e.g. ViewUitls?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+import androidx.core.view.WindowInsetsCompat;
+import androidx.core.view.WindowInsetsControllerCompat;
 
 /**
  * Classes actually having an ActionBar (as opposed to the Dialog activities)
  */
 public class AbstractActionBarActivity extends AbstractActivity {
 
+    protected int calculateInsets = WindowInsetsCompat.Type.statusBars() | WindowInsetsCompat.Type.displayCutout();
+    boolean skipActionBarInsetCalculation = false; // for edge to edge configuration, see onCreate()
```
eddiemuc: I perrsonally don't like the usage of attributes to transport config info from subclass to superclass. This requires the subclass to know when it needs to set the attribute (before first usage). Maybe using methods instead?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -263,6 +264,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
+        calculateInsets |= WindowInsetsCompat.Type.navigationBars(); // no bottom navigation, need to take navigation bar into account for insets calculation
```
eddiemuc: WIndowsInsetsCompat.Type.systemBars() should alread include navigation bars

eddiemuc:
@moving-bits appreciate your work on this. My biggest comment would be: I vote for making edge-to-edge-mode active for ALL Android versions (not just for >= 15/16), and to focus first on enabling edge-to-edge (so we can get loose of the "suppress edge-to-edge-flag) w/o making changes to current design.

Do you want support on this? If yes, how can I best help out here?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -1,20 +1,62 @@
...
+                    if (activityContent == null) {
+                        Log.e("edge2edge: activityContent not found in " + this);
+                    }
+                    if (activityContent != null) {
+                        final float actionBarHeight = skipActionBarInsetCalculation ? 0f : ViewUtils.dpToPixelFloat(10f) + getResources().getDimension(R.dimen.actionbar_height);
+                        final Insets innerPadding = insets.getInsets(calculateInsets);
+                        activityContent.setPadding(innerPadding.left, (int) (innerPadding.top + actionBarHeight), innerPadding.right, innerPadding.bottom);
+                    }
+                }
+                return insets;
```
eddiemuc: Probably use WindowInsetsCompat CONSUMED?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: My first goal was to mimic current behavior on targetSDK35 without using the opt-out flag, therefore I started that way. I've switched to calling `enableEdgeToEdge()` (needing to update core lib to 1.17 and compileSDK to 36 for this), so it's common ground now for older API levels as well.

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: done

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: done, upgraded to core 1.17 and compileSDK 36 for this

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: added

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: What do you mean? It's now part of AbstractActivity, thus pretty low level in our activity hierarchy. (But if it should ever be needed outside an activity, we can outsource it to a helper method.)

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: changed to `configureEdge2Edge()` method call. Still needs to be called before `onCreate()`, though.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
moving-bits: default contains `statusBars()`, not `systemBars()`. 

I cannot include `navigationBars()` per default, as this would add extra space at the bottom for all activities that have a bottom navigation.

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
moving-bits: That did not work for me, it removed all insets for statusBar, navigationBar etc.

moving-bits:
Thanks, @eddiemuc, for taking the time for this detailed review. I've tried to incorporate your remarks in an updated commit. If no further things arise from review, I suggest to merge to `master` for further testing - it should look & behave like today's c:geo, even with targetSDK 35 and without opt-out flag -, and then work on further enhancements (actually making use of edge to edge mode) in separate PR. For a discussion on that, see issues #17356 and #17375.

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
eddiemuc: I had in mind eg the full-screen dialigs used in Wherigo Player. They also beed edge-to-edge enabled but are not activity-based. But we can do that in a later step/PR

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
eddiemuc: Strange. I thought this would be the right way to handle this from what I read.

eddiemuc:
Fine to merge and continue to work from master for me. There may be some strange effects for nightly users in the meantime however.

eddiemuc:
@moving-bits shoudl we merge this?

moving-bits:

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
eddiemuc: @moving-bits you are (of course :-)) correct, adding "navigationBars()" indeed adds a strange extra space at the bottom. I don't understand why this is happening though. Did you find out what is actually causing this?

moving-bits: Haven't checked yet, but my guess would be the structure of the layout file: I'm adding the extra spacing to a layout element with id "activity_content", and probably the layout file with bottom navigation has the bottom navigation element placed below that one. And as material 2 bottom navigation seems to take the navigation bar into account automatically, I did not investigate further.

eddiemuc: That's probably right. 

A different concept might be needed here. I assume the activity neeeds to know if top/bottom/left/right padding is handled "elsewhere" (eg by the bottomnavigationbar or actionbar) or whether it needs to apply padding to the specified side. I have to think about this

eddiemuc: I believe I have solved this in https://github.com/cgeo/cgeo/pull/17380

PR #17365, opened 2025-08-31
add icon to "Create new list" pseudolist

ztNFny:
- dedicated icon for "Create new list" entry when storing a cache
- 
image

- change the "import" icon from outline to solid to match the style of other icons
- 
image


moving-bits:
> * change the "import" icon from outline to solid to match the style of other icons

thanks for fixing my bug

Issue #17364, opened 2025-08-31
"Calculate coordinate" always opening in "guided mode"


ztNFny:
When adding a new waypoint the "Calculate coordinate" dialog always opens with "Guided mode" enabled.

This changed quite recently, I assume with the new Coordinates dialog. However I can't find any recent code change relating to the setChecked() of that spinner... I'm sure this was different before as I never use guided mode and before recently didn't have to disable it always but was starting in non-guided mode (a global setting?)

ztNFny:
@ReggieDoodle could you please look into this?

ReggieDoodle:
Ok, will do.

ReggieDoodle:
Just checked and pre my changes mine always opened in guided mode.

Prior to @murggel picking up an issue with none guided mode I had never used it so I assumed guided was the default ?

So which is it to be?

ztNFny:
> Just checked and pre my changes mine always opened in guided mode.
> 
> Prior to [@murggel](https://github.com/murggel) picking up an issue with none guided mode I had never used it so I assumed guided was the default ?
> 
> So which is it to be?

How did you check? I just reverted to a build from end of June and it's defaulting to non-guided there (was assuming that it's a global setting but seems not)

ztNFny:
The change is that previously it used to be newType = PLAIN and now it's DEGREE_MINUTE
This is set in CoordinateInputDialog:283

ReggieDoodle:
> The change is that previously it used to be newType = PLAIN and now it's DEGREE_MINUTE This is set in CoordinateInputDialog:283

That sounds like an easy change (or another issue?) but there must be a global setting cos I never saw non guided mode unless I've specifically changed it to test the UI

So, moving on.....

1. Is it a global setting - should it be if it isn't
2. If I manipulate a WP and switch modes should it reopen in that mode for the next edited or new WP
3. What should the default be 

TBH, I did ask at the start of the dev work for some guidance on how to test the calculator (as I never use it myself) but no one stepped up

ztNFny:
> TBH, I did ask at the start of the dev work for some guidance on how to test the calculator (as I never use it myself) but no one stepped up

Unfortunately there's no tests for 99% of the UI. You have to manually go through stuff.

> I never saw non guided mode unless I've specifically changed it to test the UI

strange - for me it was vv ...

I suggest to add a global setting, seems to be the easiest way out. Depending on some other setting or viewstate just seems strange. Default probably guided=true then.

[...]

PR #17371, opened 2025-08-31
remember coordinate input non-guided mode selection

ztNFny:
Remember if the user disables "guided mode" for formula input

fixes https://github.com/cgeo/cgeo/issues/17364

@ReggieDoodle would appreciate a review as it's your code

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
@@ -14,6 +14,7 @@ import cgeo.geocaching.models.CoordinateInputData;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.sensors.LocationDataProvider;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.ui.CalculatedCoordinateInputGuideView;
 import cgeo.geocaching.ui.SimpleItemListModel;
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
```diff
@@ -297,11 +299,8 @@ public View onCreateView(@NonNull final LayoutInflater inflater, final ViewGroup
     // splitting up that method would not help improve readability
     @SuppressWarnings({"PMD.NPathComplexity", "PMD.ExcessiveMethodLength"})
     private void refreshType(final CalculatedCoordinateType newType, final boolean initialLoad) {
-        if (!initialLoad && calcCoord.getType() == newType) {
```
eddiemuc, line 300: Why does the fail-fast needs removal?

eddiemuc (commented):
Looks ok to me, although its hard to tell from code whether this will behave as expected in all situations. I trust this was tested throughly with all sorts of guided modes and coordinate formats as well?

main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
ztNFny: Switching between the modes doesn't necessarily change newType, so with the fail-fast it would be failing too often. I removed it after several rounds of testings of different scenarios as it turned out to be the solution with least complexity - and also I didn't really find scenarios where the fail-fast would be triggered "normally"

eddiemuc:
Well, lets try in the nightlies :-)

ReggieDoodle:
Thanks for sorting this out guys.

Issue #17375, opened 2025-09-01
Upgrade to SDK35: Edge-to-Edge


eddiemuc:
I figure Edge-to-Edge (enforced with SDK35) deserves its own issue.

Links collected for this so far:
* Official Android doc: https://developer.android.com/develop/ui/views/layout/edge-to-edge 
* Helpful video: https://www.youtube.com/watch?v=gIUe2OOzohw
* Blog: https://medium.com/androiddevelopers/insets-handling-tips-for-android-15s-edge-to-edge-enforcement-872774e8839b

Special links helping with edge-to-edge for the old-fashoined ActionBar used by c:geo:
* Help on extending background color of actionbar (toolbar) under system bar: https://dev.to/lejdi/handling-android-toolbars-in-android-15-1od
* Help on getting the action bar view: https://stackoverflow.com/questions/20023483/how-to-get-actionbar-view

PRs by @moving-bits : 
* Already merged to allow usage of PlayStore: https://github.com/cgeo/cgeo/pull/17361
* Working on the "big" solution: https://github.com/cgeo/cgeo/pull/17355

The following was originally posted by @moving-bits in [#16275](https://github.com/cgeo/cgeo/issues/16275#issuecomment-2972669730)

Excerpt from SDK upgrade assistant info regarding edge-to-edge mode:

If your app is not already edge-to-edge, you are most likely impacted. In addition to the scenarios for apps that are already edge-to-edge, you should consider the following:
1. If your app uses Material 3 Components (androidx.compose.material3) in compose, such as TopAppBar, BottomAppBar, and NavigationBar, these components are likely not impacted because they automatically handle insets.
1. If your app is using Material 2 Components (androidx.compose.material) in Compose, these components don't automatically handle insets. However, you can get access to the insets and apply them manually. In androidx.compose.material 1.6.0 and later, use the windowInsets parameter to apply the insets manually for BottomAppBar, TopAppBar, BottomNavigation, and NavigationRail. Likewise, use the contentWindowInsets parameter for Scaffold.
1. If your app uses Views and Material Components ( com.google.android.material), most views-based Material Components such as BottomNavigationView, BottomAppBar, NavigationRailView, or NavigationView, handle insets and require no additional work. However, you need to add android:fitsSystemWindows="true" if using AppBarLayout.
1. For custom composables, apply the insets manually as padding. If your content is within a Scaffold, you can consume insets using the Scaffold padding values. Otherwise, apply padding using one of the WindowInsets.
[...]

PR #17380, opened 2025-09-03
rel to #17375: Edge 2 Edge for Wherigo, fix actionbar, fix imageview

eddiemuc:
rel to #17375: Edge 2 Edge for Wherigo, fix actionbar, fix imageview


eddiemuc:
This will not be the end of it....

eddiemuc:
+ fix for the "gap" problem over bottomnavigationbar (+ on side of navigation bar in landscape)

eddiemuc:
Build is green (at least the important parts of it :-)).
I will merge this PR rather quickly if no one objects. I saw some artefacts in the current master which this PR (hopefully) fixes, and I am pretty sure we are not done yet with edge-2-edge-discussions anyway. So I want to get this into nightlies as quickly as possible,

main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
```diff
@@ -48,7 +47,7 @@ public abstract class AbstractNavigationBarMapActivity extends AbstractNavigatio
 
     @Override
     public void onCreate(final Bundle savedInstanceState) {
-        configureEdge2Edge(0, Settings.getMapActionbarAutohide());
+        setFixedActionBar(false);
         super.onCreate(savedInstanceState);
```
moving-bits, line 51: Will a fixed value of false work in both map modes - "hide on tap" enabled yes/no?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -29,4 +41,75 @@ public void setTitle(final CharSequence title) {
...
+            return false;
+        }
+        return actionBar.isShowing();
+    }
+
+    @Override
+    @NonNull
+    protected Insets calculateInsetsForActivityContent(@NonNull final WindowInsetsCompat windowInsets, @NonNull final Insets def) {
+        final Insets insets = super.calculateInsetsForActivityContent(windowInsets, def);
+        this.actionBarTranslation = Math.min(insets.top, ViewUtils.dpToPixel(50));
```
moving-bits: There are a few places with hardcoded `ViewUtils.dpToPixel(50))` - intentionally?

main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
eddiemuc: It will lead in both cases ("hide on tap" enabled and disabled) to the map being drawn also under the ActionBar. Although in case of "hide on tap" disabled, this part of the map will never be revealed. Is this problematic from your perspective?

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
eddiemuc: That's a lapsus, I can make a constant out of it. The value of 50 is experimental though.
(I expect that more work will be needed on edge-2-edge anyway)

main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
moving-bits: One thing that comes into my mind: Old implementation with "hide on tap" activated had/has the problem, that mapping a list may hide caches under a then-visible actionbar. With "hide on tap" disabled, a smaller map was drawn, therefore cache distribution could be calculated accordingly.
If now both modes have map parts under the actionbar, we may need to adjust zoom level calculation when mapping a list.

eddiemuc: > One thing that comes into my mind: Old implementation with "hide on tap" activated had/has the problem, that mapping a list may hide caches under a then-visible actionbar. With "hide on tap" disabled, a smaller map was drawn, therefore cache distribution could be calculated accordingly. If now both modes have map parts under the actionbar, we may need to adjust zoom level calculation when mapping a list.

True, the "hiding" problem will now occur in both cases ("hide on tap" enabled or disabled). For "hide on tap" disabled we can switch again to old solution (by setting "setFixedActionBar" to true in this case), but the problem would remain for "hide on tap" like in old solution. There's no easy fix for this I guess, otherwise it would have been implemented with "old" solution already?

PR #17388, opened 2025-09-03
Support new Lab2gpx format

murggel:
## Description

For Adventure Labs Lab2gpx already returns the guid (but in a format of `ALhttps://labs.geocaching.com/goto/guid`), since the firebase-link is no more available.

This PR adapt the recognization for the geocode.
Lab2gpx has an option to convert the code to uppercase which will lead to `ALHTTPS://LABS.GEOCACHING.COM/GOTO/GUID`


murggel:
with current lab2gpx it seems to work without any changes at our side

PR #17393, opened 2025-09-04
fix #17275: offer google translation for cache/trackable listing, hint, waypoints and wherigo

### LINKED ISSUE #17275: Google translate option for listing
> Would it be possible to have another button that automatically opens the description in Google Translate? The current method of highlighting the text and then clicking translate in Google Translate works for me and gives a good translation, and the Google app allows you to download languages for fully offline translation as well. The only problem with this approach is that it can be quite time-consuming to highlight the whole thing, especially for a long description. 

 _Originally posted by @LayanJethwa in [#17213](https://github.com/cgeo/cgeo/issues/17213#issuecomment-3164931899)_

eddiemuc:
fix #17275: offer google translation for cache/trackable listing, hint, waypoints and wherigo

This PR uses the approach already implemented for logs and extends it towards the mentioned other c:geo artefacts


eddiemuc:
codacy doesn't seem to run on this PR

eddiemuc:
added support for DeepL. I don't think we should focus so much on Google Translate only

PR #17400, opened 2025-09-04
fix #17273: Add menu-entry to set waypoints as visited  

### LINKED ISSUE #17273: Automatically mark waypoints as visited
### Tell us your idea!

When the coordinates of a cache are corrected, there should be an option to automatically mark all the waypoints of the cache as visited, so that they don't interfere with other icons on the map.

There should be a menu item in menu map content and behaviour, where you can select a checkbox to mark waypoints of caches with corrected coordinates as visited.

### Is there an alternative?

Manually marking them as visited, but this is time-consuming especially if you have downloaded the GPX from the website.

### Additional Information

_No response_

murggel:
## Description

Add menu-entry to set waypoints as visited 

A dialog will appear to select the waypoint-types, which should be set to visited.

image

Image

## Related issues

fix #17273

## Additional context

To discuss, whether the (checked) waypoint-types should be remembered and proposed as pre-selection or add a button "Last selection" (like in list-view), so for some point in time, only a confirmation would be necessary

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -613,7 +613,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             }, result -> {
                 if (result) {
                     if (waypointPos > 0) {
-                        selectedWaypoint = WaypointsViewCreator.createWaypointList(cache).get(waypointPos - 1); // set new list position to avoid jumping to the top
+                        // set new list position to avoid jumping to the top
+                        selectedWaypoint = WaypointsViewCreator.createWaypointList(cache, false).get(waypointPos - 1);
                     }
                     notifyDataSetChanged();
                     GeocacheChangedBroadcastReceiver.sendBroadcast(CacheDetailActivity.this, cache.getGeocode());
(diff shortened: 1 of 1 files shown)


murggel:
To discuss, whether the (checked) waypoint-types should be remembered and proposed as pre-selection or add a button "Last selection" (like in list-view), so for some point in time, only a confirmation would be necessary

murggel:
Ready for review.

I will try to implement remembering the last selection in a different PR.

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -612,7 +614,7 @@ public boolean onContextItemSelected(final MenuItem item) {
             }, result -> {
                 if (result) {
                     if (waypointPos > 0) {
-                        selectedWaypoint = WaypointsViewCreator.createSortedWaypointList(cache).get(waypointPos - 1); // set new list position to avoid jumping to the top
+                        selectedWaypoint = WaypointsViewCreator.createWaypointList(cache).get(waypointPos - 1); // set new list position to avoid jumping to the top
```
eddiemuc: What's this about? Sorting having strange effects here?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1195,6 +1200,54 @@ private void dropGeneratedWaypoints() {
...
+    private List<Waypoint> getSortedWaypointList(final Geocache cache) {
+        if (null != cache && cache.hasWaypoints()) {
+            final List<Waypoint> waypoints = cache.getWaypoints();
+            Collections.sort(waypoints, cache.getWaypointComparator());
+            return waypoints;
+        }
+        return Collections.emptyList();
+    }
+
+    private Set<WaypointType> getWaypointTypes(final Iterable<Waypoint> waypoints) {
```
eddiemuc: Maybe make a static helper method in class "Waypoints"?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1195,6 +1200,54 @@ private void dropGeneratedWaypoints() {
         }
     }
 
+    private List<Waypoint> getSortedWaypointList(final Geocache cache) {
```
eddiemuc: Good call to make this a method. I am surprised it was not before.
But maybe move it to class Geocache or as helper method to class Waypoint?

eddiemuc (approved):
LGTM. Waiting with merge due to minor comments

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: That was kind of strange - the method-name was "createSortedWaypointList" but no sort was performed (see separat commit "Rename"). I just renamed the method to "createWaypointList", created a new method "createSortedWaypointList" woth sorting and changed the related calls

murggel: But maybe sorting would be the correct way here, otherwise it could be a little jumping when deleting a waypoint - will check that

eddiemuc:
Merge?

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: Maybe a better implementation would be to get the position before deleting the waypoint...

murggel: But that would be a separate PR - still I am not sure, if that is really better. 
So far from my testing, sorting at that place does it make not really better, even depending if the option "Prevent waypoints from note" is checked

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: Done - moved to class Geocache

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: Done - moved to class Waypoint

murggel:
@eddiemuc Ready for further review

murggel:
> Merge?

I changed the method WaypointsViewCreator.createWaypointList - just have a quick look on the changes...

eddiemuc:
lgtm

PR #17408, opened 2025-09-05
SimpleDialog/TextParam: make sure to reset TextViews properly if they are reused eg in Recyclerview

eddiemuc:
SimpleDialog/TextParam: make sure to reset TextViews properly if they are reused eg in Recyclerview

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/ImageParam.java
+++ b/main/src/main/java/cgeo/geocaching/ui/ImageParam.java
@@ -9,6 +9,7 @@ import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.util.Pair;
+import android.view.View;
 import android.widget.ImageView;
 
 import androidx.annotation.DrawableRes;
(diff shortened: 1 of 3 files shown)


ztNFny:
There's still some strange behaviour when selecting items:
[Screen_recording_20250905_194839.webm](https://github.com/user-attachments/assets/92c707c5-7f48-43f4-8d8e-206b2010c608)

eddiemuc:
> There's still some strange behaviour when selecting items:
> [Screen_recording_20250905_194839.webm](https://github.com/user-attachments/assets/92c707c5-7f48-43f4-8d8e-206b2010c608)

You are referring to the changing icons (from red swuare to red flag and cice versa at the other item)?

Selecting of the one item in the movie was done by you or did that also happen by itself?

ztNFny:
Yes, the icons changing "randomly" on selecting an item.

eddiemuc:
@ztNFny thanks a lot! Was a bug in ImageParam wrt sizing of emoji drawables. It was always sized to the view in our case. and in case the view was transparnet before the size was 0. And that lead to an emoji consisting only of a red dot. And this red dot was then expanded acrosss the forced size of 30x30dp. Hence the red square.

I included the fix in this PR too

PR #17412, opened 2025-09-06
rel to #17375: move legacy compassrose movement from HideActionBarUtils to GoogleMapView

eddiemuc:
rel to #17375: move legacy compassrose movement from HideActionBarUtils to GoogleMapView

While working on #17375 and trying to refactor showing/hiding ActionBar fpr edge-2-edge, I noticed method `HideActionBarUtils.adaptLayoutForActionBarHelper`. Apparently this method is used to adjust position of compassrose in case of actionbar showing/hiding.

I also noticed however that this method is only effective in used for legacy GoogleMapView class. It was still called from unifiedmap but with no effect (because compassrose field passed was null). 

This PR moves said method to legacy GoogleMapView in order to refactor unifiedmap further. @moving-bits I would appreciate a review before merge. Are my thoughts correct?

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapView.java
+++ b/main/src/main/java/cgeo/geocaching/maps/google/v2/GoogleMapView.java
@@ -45,6 +45,7 @@ import android.view.MotionEvent;
 import android.view.View;
 
 import androidx.annotation.NonNull;
+import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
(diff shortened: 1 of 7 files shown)


moving-bits (approved):
> Apparently this method is used to adjust position of compassrose in case of actionbar showing/hiding.

yes (nowadays)

> I also noticed however that this method is only effective in used for legacy GoogleMapView class. It was still called from unifiedmap but with no effect (because compassrose field passed was null).

Yes. This method used to be not for compass only, but also for some more fields, but that got changed over time by putting more things into a layout file dependency instead of manually moving things around.

Your changes seem to work for me, so I would say "go". Thanks for cleaning up - this will make removal of legacy maps easier in the future.

PR #17425, opened 2025-09-09
fix #11763: allow multi-selection for pocketqueries / bookmarklists

### LINKED ISSUE #11763: Allow selecting several PQ for import
### Tell us your idea!

From support mail:

I am using a lot of pocket queries, which I usually import into the same list in c:geo.
Currently I have to import each PQ individually, so it would be time saving if I could select several of them and import them into a list as a batch.

Allow multiple selection in the PQ list and import all of those into a selected list

### Is there an alternative?

Import one by one, which is time consuming.
Combining PQ on sever side not possible as they already use the maximum amount of caches or cover different areas or search criteria.

### Additional Information

_No response_

murggel:
## Description

Allow multi-selection for pocketqueries / bookmarklists
On long click start / stop select mode.

### Open points
* A toast is shown, if no entry is selected - I didn't manage to disable the button
* What to do, if not all selected pocketqueries  are downloadable? 
Download only downloadable (and show maybe a toast) or none?
* Progressbar is shown per downloaded pocket query, didn't manage to show it in one progressbar yet

## Related issues

fix #11763

## Additional context
Bookmarklists are only downloadable, preview is currently not available

### Pocket Queries

grafik 
grafik

### Bookmark lists
grafik 
grafik


ztNFny:
> What to do, if not all selected pocketqueries are downloadable?

I'd propose to disable those non-downloadable items. The checkbox has an enabled/selectable flag.

murggel:
> > What to do, if not all selected pocketqueries are downloadable?
> 
> I'd propose to disable those non-downloadable items. The checkbox has an enabled/selectable flag.

But they should be selectable for preview...

So I could hide the download-button, if a non-downloadable item is selected or just don't download it und bring a message / toast which contains the (not) downloaded pqs

ztNFny:
i pushed a fix for disabling the download button

murggel:
> i pushed a fix for disabling the download button

Thanks, will try that

murggel:
> > i pushed a fix for disabling the download button
> 
> Thanks, will try that

Disabling works, but the download button is slightly misaligned (and I assume, the icon-color is not the correct one - with old light theme it was still black instead of white)...

grafik


so still needs adaption

ztNFny:
setEnabled(false) doesnt work, the onclick handler is still registered. You need setClickable(false).
Your commit broke it 😉

murggel:
> setEnabled(false) doesnt work, the onclick handler is still registered. You need setClickable(false). Your commit broke it 😉

For me it was the other way round... I had a breakpoint in the clickhandler...

murggel:
Setting the iconTint programmatically works for me, but don't know, if that is the only option

ztNFny:
> > setEnabled(false) doesnt work, the onclick handler is still registered. You need setClickable(false). Your commit broke it 😉
> 
> For me it was the other way round... I did had a breakpoint in the clickhandler...

I tested it last night and have another place in code with the same behaviour. setEnabled() doesnt imapct onclick handlers manually set.

murggel:
> > > setEnabled(false) doesnt work, the onclick handler is still registered. You need setClickable(false). Your commit broke it 😉
> > 
> > 
> > For me it was the other way round... I did had a breakpoint in the clickhandler...
> 
> I tested it last night and have another place in code with the same behaviour. setEnabled() doesnt imapct onclick handlers manually set.

Probably depending on android version? Then I probably need to set both...

murggel:
> Setting the iconTint programmatically works for me, but don't know, if that is the only option

Weard - with a fixed color version in the xml works, but for me not with @color/...

ztNFny:
In uiutils i implemented a method to color icons depending on whether they are in toolbar or overflow menu, use that one. Its used e.g. in cachedetails

murggel:
> > Setting the iconTint programmatically works for me, but don't know, if that is the only option
> 
> Weard - with a fixed color version in the xml works, but for me not with @color/...

For me it works with `app:iconTint` in the xml

murggel:
> The download button is slightly misaligned (and I assume, the icon-color is not the correct one - with old light theme it was still black instead of white)...
> grafik
> 
> so still needs adaption

Fixed with not using action_bar_icon (I guess the problem is the weight) and setting background

|all downloadable| One of the selected is not downloadable|
| --- | --- |
|image |image|

murggel:
Failing tests should be unrelated

murggel:
Ready for review

main/src/main/java/cgeo/geocaching/connector/gc/AbstractListAdapter.java
```diff
@@ -30,6 +42,92 @@ protected static final class ViewHolder extends AbstractRecyclerViewHolder {
         }
     }
 
+
+    private static class TouchListener implements View.OnClickListener, View.OnTouchListener {
```
eddiemuc, line 46: it seems odd to have something as "generic" as this deeply buried in the gc connector part of c:geo. But then, the same goes with having an "AbstractListActivity" and an "AbstractListAdapter" specific for GC connector. So that's not something to be discussed as part of this PR.

eddiemuc:
An overall review revealed no problems to me. Let's try in the nightlies and adjust if necessary

main/src/main/java/cgeo/geocaching/connector/gc/AbstractListAdapter.java
murggel: The "AbstractListActivity" and an "AbstractListAdapter" is specific to "OnlineLists", but I didn't manage to rename them... (we have 2 "AbstractListActivity")

murggel: I can try to get rid of the duplicate TouchListener and make a more generic one in a separate PR...

PR #17430, opened 2025-09-10
fix #17291: Consider current timezone for sorting dates in sql

### LINKED ISSUE #17291: Wrong position of own log inside the logs-tab
### Describe your problem!

After logging from c:geo and refreshing the cache later, the own log always appears as last entry of the day

Keeping the logged time is probably the cause.

### Possible solution
Sorting per logging ID for logs with same log-day

murggel:
## Description

Consider current timezone for sorting dates in sql

## Related issues

fix #17291


murggel:
I need to test it for different timezones though

murggel:
There is probably another problem if the time zone is changed after logging in. I'm not entirely sure how the date is stored in the database (whether the time zone is taken into account or as UTC), I still need to check that...

Edit: the time is stored in UTC (milliseconds). 

So Ii guess, the problem is, that for the online logs, we don't get a dedicated time (only day) which then will probably lead to a different day than the current log-time.

murggel:
Maybe its sufficient to sort after service-log-id? Unfortunately not all logs have a service-log-id?

ztNFny:
still not working ... :-(

murggel:
> still not working ... :-(

Can you give me a gpx of the cache with your log?

murggel:
Maybe we have to consider the online date and change only the time of the online day on merging the log-date

ztNFny:
my own log is stored with date:
Wed Sep 10 2025 14:31:39 GMT+0200
whereas all logs after it are stored with "no time":
Wed Sep 10 2025 00:00:00 GMT+0200

Sorting by date alone will thus never work (at least for GC, didn't check other connectors). 

I think the way to go would be:
1. get datetime from db
2. convert to date
3. within date sort by service_log_id

murggel:
> my own log is stored with date: Wed Sep 10 2025 14:31:39 GMT+0200 whereas all logs after it are stored with "no time": Wed Sep 10 2025 00:00:00 GMT+0200
> 
> Sorting by date alone will thus never work (at least for GC, didn't check other connectors).
> 
> I think the way to go would be:
> 
>     1. get datetime from db
> 
>     2. convert to date
> 
>     3. within date sort by service_log_id

ok, then sorting already in sql won't work and we have to do the sorting in our code... I wanted to avoid that...

murggel:
Will try that, could that become a performance-thing?

ztNFny:
sql can do it:
SELECT * FROM cg_logs ORDER BY DATE(date, 'unixepoch'), service_log_id DESC;

murggel:
> sql can do it: SELECT * FROM cg_logs ORDER BY DATE(date, 'unixepoch'), service_log_id DESC;

No, that was done before, I changed it first to 
ORDER BY Date((date/1000, 'unixepoch') DESC, service_log_id DESC
to get rid of the time and using only year, month and day... but that still failed for GMT-Offset, so I used

ORDER BY Date((date+" + offsetMillis + ")/1000, 'unixepoch') DESC, service_log_id DESC
with offsetMillis the Offset for GMT

But still seems not to work...

ztNFny:
/1000 is not getting rid of the time though?! converting to DATE() does that

murggel:
> /1000 is not getting rid of the time though?! converting to DATE() does that

Correct, but it is converting the milliseconds into seconds

murggel:
Latest changes (force push) should work now

murggel:
Ready for review and merge

Issue #17405, opened 2025-09-05
Google MLKit Offline Translation: Attribution required


eddiemuc:
I stumbled across this page: https://developers.google.com/ml-kit/language/translation/translation-terms

"Attribution: An app that uses the ML Kit on-device Translation API must comply with the Google Cloud Translation API [attribution requirements](https://cloud.google.com/translate/attribution), as far as they apply to applications. These requirements cover guidelines on how your app must handle layout, Google attribution, and branding."

This includes to my understanding naming of services, buttons and such as well as markingmarking translation results to be coming from google (eg "translated by Google" with their logo). More details can be found here: https://cloud.google.com/translate/attribution

IMHO this is a must if we want to continue using MLKit Translations in c:geo. Feedback?

eddiemuc:
Tbh I don't see us having zhe option not to do anything. I see two options:
- Option 1: adding Google attribution
- Option 3: remove MLKit Translation completely

ztNFny:
Looking into what actually needs to be done. If it's too complex - albeit not great quality - very accessible and fully offline translation ... with the complexities caused by offline translate (size, dependencies, now this) and as much as the time spent hurts - i'd be in line with removing it.

moving-bits:
> Tbh I don't see us having zhe option not to do anything. I see two options:
> 
> * Option 1: adding Google attribution
> * Option 3: remove MLKit Translation completely

I'm wondering what "option 2" might be? 🤣

Anyway, with the new integration of Google Translate and DeepL (thanks!) I would prefer option 3 to reduce complexity.

Lineflyer:
Same as for the discussion (some years ago) about implementing turn-by-turn navigation into c:geo, I think that we should rely on specialized apps for such purposes.
Meaning: Better support handing over texts to be translated to several other apps (like Google Translate, maybe more) than to implement and maintain such features ourselves.

So Option 3 for me although I like the convenience of the current feature, but the price is high in terms of app size, attribution, etc.

Lineflyer:
Another disadvantage in our current implementation:

<img width="1130" height="359" alt="Image" src="https://github.com/user-attachments/assets/9a41f84a-a9fe-4669-bcdf-2aa9b3f7bdba" />

ztNFny:
> Another disadvantage in our current implementation:
> 
[...]

PR #17431, opened 2025-09-10
MLKit attribution

ztNFny:
add attribution per https://developers.google.com/ml-kit/language/translation/translation-terms by displaying a logo on translated text and changing some translation strings to point to the source
fixes https://github.com/cgeo/cgeo/issues/17405

image
image
image
image


eddiemuc:
@ztNFny I believe TrackableActivity is not part of the PR. Should I merge it anyway or do you want to add it to this PR?

eddiemuc:
Read through google docs I feel that two things may be missing:

From website:
**Google Translate service**: Use only the term "Google Translate" to refer to the service made available through these APIs and at URLs such as http://translate.google.com. Do not modify this word mark, for example, through hyphenation, combination abbreviation, or acronym such as: Google-Translate, GT, GoogTrans, GTrans, etc.

-> text in context menu is currently "google translate", must be "google translate service"
-> this probably also applies to "external translation" btw (in release branch this is only used for loga)

From website:
**Translation buttons, Mobile apps**
The button or user action that triggers a translation should be clearly labeled 'Translate with Google.'

-> we use "translate to French with google", and the text is not in the button but left of it. Maybe reformulate to "translate with google to French"? Wrt button I hope they don't mind putting the text next to it (not on it)?

ztNFny:
> -> text in context menu is currently "google translate", must be "google translate service"

Where are you getting the "service" from? It's saying to use the term "Google Translate".

> we use "translate to French with google", and the text is not in the button but left of it. Maybe reformulate to "translate with google to French"? Wrt button I hope they don't mind putting the text next to it (not on it)?

As a tooltip is less visible I decided to put it in the text. We can also put it in the tooltip.
I think the current placement of the language should be fine, the important part is that it's clear that the translation is provided by Google.

ztNFny:
closing was me pressing the wrong button

ztNFny:
> @ztNFny I believe TrackableActivity is not part of the PR. Should I merge it anyway or do you want to add it to this PR?

TrackableActivity accepts the logo as spannable, only listing text in CacheDetailsActivity needs special handling.

PR #17203, opened 2025-07-28
change (light) theme color

fm-sys:
During the event, some of us discussed that we don't really like the current action bar color in light theme. I personally would go with the tinted version. What do you think?

| current | pure white | slightly tinted (currently used in this PR) |
|---------|------------|--------|
| Screenshot_20250728-145027 | Screenshot_20250728-160159 | Screenshot_20250728-160324 |
| Screenshot_20250728-161138 | Screenshot_20250728-160929 | Screenshot_20250728-160900 |

PR #17434, opened 2025-09-12
Use dark text on light systembars (rel #17203)

murggel:
## Description

Due to the change to lighter action-bar color and using dark text color, the text on the systembar is hardly visible

This PR changes the setting windowLightStatusBar and windowLightNavigationBar accordingly

## Related issues

rel #17203


murggel:
Qzick and dirty solution that works, but will looks for better solutions

murggel:
Because we now have a light action bar with dark text in light mode, couldn't we use directly
`<style name="cgeo" parent="Theme.MaterialComponents.DayNight">`?
But I didn't get it working that way

murggel:
I would vote for merging this in the nightly as soon as possible for better visibility of the system bars, so we have some more time for PR #17458

PR #17433, opened 2025-09-12
AL API store stage images

JakeDot:
Store images from Adventure Lab waypoints to cache listing images page.
Added new Image category "Waypoints"

fixes #17065

PR #17442, opened 2025-09-14
AL API store stage images

JakeDot:
See https://github.com/cgeo/cgeo/pull/17433


JakeDot:
ok to test

moving-bits:
@ztNFny @baiti 
As you are much deeper into the ALC stuff - can you review this PR (as time permits)?

baiti:
> @ztNFny @baiti As you are much deeper into the ALC stuff - can you review this PR (as time permits)?

I have briefly looked at the commits. From all I can see parsing and filling the "cache" object hasn't changed. The changes are about treating images in the AL objects in the same way as in the GC objects and it is definitelty good to have this consistency here. I have not tested the branch though.

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -403,61 +444,70 @@ private static Geocache parseCacheDetail(final JsonNode response) {
...
                 wpt.setGeofence((float) wptResponse.get("GeofencingRadius").asDouble());
 
-                final StringBuilder note = new StringBuilder("<img src=\"" + ilink + "\"></img><p><p>" + desc);
+                final Image spoilerImage = getImage(ilink, wptName, desc);
+
+                wptImages.add(spoilerImage);
+
+                handler = new ImageHandler(cache, wpt, ilink, spoilerImage);
+
+                downloader.waitForEndCompletable(handler);
```
eddiemuc: What is the need for this? Will this introduce some sort of blocking?

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -70,28 +75,82 @@ private ALApi() {
...
+    private static Image getImage(final String ilink, final String wptName, final String desc) {
+        return new Image.Builder()
+                .setUrl(ilink)
+                .setTitle(wptName)
+                .setDescription(desc)
+                .setCategory(Image.ImageCategory.WAYPOINT)
+                .build();
+    }
+
+    static class ImageHandler extends DisposableHandler {
```
eddiemuc: Why do we need a handler class here?

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -70,28 +75,82 @@ private ALApi() {
...
+
+        @NonNull
+        private static String imageString(final String localUri) {
+            return "<img src=\"" + localUri + "\"></img><p><p>";
+        }
+    }
+
     @JsonIgnoreProperties(ignoreUnknown = true)
-    public static final class ALSearchV4Query {
+    static class ALSearchV4Query {
```
eddiemuc: Why the need to change all the modifiers?

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -403,61 +444,70 @@ private static Geocache parseCacheDetail(final JsonNode response) {
...
                 } else {
                     wpt.setOriginalCoordsEmpty(true);
                 }
-                if (result == null) {
-                    result = new ArrayList<>();
-                }
-
-                result.add(wpt);
-            } catch (final NullPointerException e) {
+            } catch (final Exception e) {
```
eddiemuc, line 374: Minor: maybe only catch RuntimeException?

main/src/main/java/cgeo/geocaching/models/Waypoint.java
```diff
@@ -64,6 +64,8 @@ public class Waypoint implements INamedGeoCoordinate {
     private Geopoint preprojectedCoords = null;
     @Nullable
     private Float geofence; // radius in meters
+    @Nullable
```
eddiemuc: Why not private?

main/src/main/java/cgeo/geocaching/models/Waypoint.java
```diff
@@ -324,6 +325,15 @@ public void setGeofence(@Nullable final Float geofence) {
         this.geofence = geofence;
     }
 
+    @Nullable
```
eddiemuc, line 334: Main:
with this PR, a new image attribute is introduced to the central "Waypoint" model class. 
But I don't see this handled in all the cases necessary eg:
* its content seems not to be stored in database for offline caches
* it seems not to be handled in display eg in waypoint details
--> how is this adressed?

eddiemuc (changes requested):
"Request changes" because of seemingly missing handling of new "image" attribute of "Waypoint" in database storage and waypoint display. 
I also don't understand the necessity of a handler in ALApi.

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
JakeDot: I wanted to simplify the class structure, private isn't needed in a local class that's not used outside of the source file.

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
JakeDot: Doesn't matter much if it's rethrown anyway but you may be right, if you aren't interested in checked exceptions in the log.

JakeDot:
retest this please

main/src/main/java/cgeo/geocaching/models/Waypoint.java
JakeDot: Added saving and restoring the field from DB

moving-bits:
@JakeDot 
I cannot comment on the actual content of this PR (see review by @eddiemuc for this), but I ask you to redo this PR technically - it cannot be merged in current state, as it contains several unrelated commits.

Please start with a fresh branch from current state of master, then cherry-pick your own commits only, leaving out the in-between-commits from other users. See git documentation for how to cherry-pick.

As a general note: Typically you do not need to update a PR with commits from other people at all, GitHub will do this so-called "rebase" during merge for you automatically, as long as there is no merge conflict.

Ad hoc I would say there are only two scenarios where a local rebase is useful for a PR: 
- Either if there are conflicting changes on the branch you want to merge your PR into (which would prevent your PR being mergable), 
- or if you need some more recent changes from that branch to continue your work on your PR.

So if you feel you need to update your PR to the current head of the branch you are working on (like `master`), please do so only by _rebasing_ your branch (not by _merging_ with other branches). A typical line of git commands for this would be:

```
git fetch upstream
git checkout master
git pull --rebase upstream master
```
Those commands update your local copy of branch `master` to the current state of c:geo's `master` branch.

Then, in a second step, bring your PR's working branch to the current state of `master` branch:

```
git checkout your-branch-name
git rebase upstream/master
```

If you have conflicting changes, you will need to fix them locally before this rebase can be completed.

If you have rebased your local branch, coming from a state which you have pushed to GitHub already, you will need to use the "--force" option on next push command, as you have changed the base which your PR is based upon:

``` 
git checkout your-branch-name
git push --force
```

(For the sake of completeness: If your PR should be targeted against `release` instead of `master`, replase `master` with `release` in the command lines above. But: c:geo typically uses PR targeted against `release` branch for urgent bug fixes only.)

eddiemuc:
Before more time is invested into this PR:

this PR introduces the concept of a waypoint image to c:geo: a waypoint may have an image associated with it. (Just one, multiple images are not supported). By this construct, the image can then be made available for offline usage (like cache listing images today). This construct is then used to store the optional image in an Adventure Labs waypoint.

is the @cgeo/team fine with this approach?

murggel:
> Before more time is invested into this PR:
> 
> this PR introduces the concept of a waypoint image to c:geo: a waypoint may have an image associated with it. (Just one, multiple images are not supported). By this construct, the image can then be made available for offline usage (like cache listing images today). This construct is then used to store the optional image in an Adventure Labs waypoint.
> 
> is the @cgeo/team fine with this approach?

I like the idea, but why restrict it to Adventure Labs? That would be helpful for Multis as well to associate own images to a waypoint.  
So it should be possible to select own images and add them to a waypoint like for a cache.

How would the images be shown to the user? In the cache-image-tab ot will there be a own (image-)tab per waypoint (because a waypoint does not have its own tab currently)

But maybe this can be done in a second step / next PR

baiti:
This is meant to provide some moral support for @JakeDot:

I have been following the trajectory of this PR over more than two months now, initially as a reviewer since I implemented a large part of the `ALApi`. Over time and after many many commits I realized that it is going down the same path as quite a few of my earlier PRs where I ended up hoplessly in a mess and more than once it required assistance from the real Pro's to get me back out of the mess again. More than once it meant doing exactly what @moving-bits wrote here: https://github.com/cgeo/cgeo/pull/17442#issuecomment-3523960566 and I still consider this the ultimate recipe for myself, should I work on another PR again. His workflow, printed and laminated and now resting under my pillow always only seconds away from my dev system.

Now to the content: I have seen quite a few changes where my elementary use of Java has been rewritten much more elegantly so reviewing this kind of a PR helps me to improve not only my Java skills but also my understanding of the overall concepts and codebase of cgeo, however, speaking of which, the latter is still not good enough such that I could give meaningful suggestions how these images should be handled in the end. In general what I tried to do even from the very beginning is to treat these LCs the same way as regular geocaches. With limited success though, just think of the `geocode` for one ugly example. So perhaps the most important point: Can it be made to work in the same way as it works for any other geocache waypoint object? Ideally this could mean to use common implementation for GC and LC (and others).

I am looking forward to see this PR being merged (once redone). I have a feature branch of my own sitting on my dev system and I will need @moving-bits' recipe to see whether it still can be rebased on top of an upodated master. I have my doubts but I will try. The topic I am working on is one of the two high on the list of our users:

- **true found state** (which is solved I believe)
- **foundDate**, or **completion date** which is not solved up to this day as it requires information from the servers that are very hard to get with only the `consumer_key` that was provided to us.

eddiemuc:
> > Before more time is invested into this PR:
> > 
> > this PR introduces the concept of a waypoint image to c:geo: a waypoint may have an image associated with it. (Just one, multiple images are not supported). By this construct, the image can then be made available for offline usage (like cache listing images today). This construct is then used to store the optional image in an Adventure Labs waypoint.
> > 
> > is the @cgeo/team fine with this approach?
> 
> I like the idea, but why restrict it to Adventure Labs? That would be helpful for Multis as well to associate own images to a waypoint.  
> So it should be possible to select own images and add them to a waypoint like for a cache.
> 
> How would the images be shown to the user? In the cache-image-tab ot will there be a own (image-)tab per waypoint (because a waypoint does not have its own tab currently)
> 
> But maybe this can be done in a second step / next PR

I didn't mean own images. I was only talking about images from the server (in case of adventure labs from the adventure lab server). Associating own (local) images to waypoints would be a different topic.

JakeDot:
> I am looking forward to see this PR being merged (once redone). I have a feature branch of my own sitting on my dev system and I will need @moving-bits' recipe to see whether it still can be rebased on top of an upodated master. I have my doubts but I will try. The topic I am working on is one of the two high on the list of our users:
> 
>     * **true found state** (which is solved I believe)
> 
>     * **foundDate**, or **completion date** which is not solved up to this day as it requires information from the servers that are very hard to get with only the `consumer_key` that was provided to us.

Could you push this branch to your GitHub, I'm very interested

JakeDot:
> I like the idea, but why restrict it to Adventure Labs? That would be helpful for Multis as well to associate own images to a waypoint.

I doubt there are many multis (or other caches) that have an image (or multiple) associated with them in a listing in the wild, because there is no option to do so apart from pasting a link into the waypoint's description field.


> So it should be possible to select own images and add them to a waypoint like for a cache.

 This would probably be good to implement in a seperate PR.

JakeDot:
is there someone who can tell me what the correct import order is here (import static cgeo.geocaching.models.Image.ImageCategory.WAYPOINT;) since the lint will not tell me?

eddiemuc:
> is there someone who can tell me what the correct import order is here (import static cgeo.geocaching.models.Image.ImageCategory.WAYPOINT;) since the lint will not tell me?

https://github.com/cgeo/cgeo/blob/master/.editorconfig#L17

murggel:
> > is there someone who can tell me what the correct import order is here (import static cgeo.geocaching.models.Image.ImageCategory.WAYPOINT;) since the lint will not tell me?
> 
> https://github.com/cgeo/cgeo/blob/master/.editorconfig#L17
> 

See also the wiki-page
https://github.com/cgeo/cgeo/wiki/Coding-Conventions

JakeDot:
@cgeo/team does anybody have tome to review?

JakeDot:
@eddiemuc I changed it up a bit, removing unnecessary stuff.

main/src/main/java/cgeo/geocaching/models/Waypoint.java
```diff
@@ -182,16 +186,20 @@ public static void mergeWayPoints(@NonNull final List<Waypoint> newPoints, @Null
...
+                        newWaypoint.merge(oldWaypoint);
+                    }
+                } else if (oldWaypoint.isUserDefined() || forceMerge) {
                     newPoints.add(oldWaypoint);
-                } else {
-                    newWaypoint.merge(oldWaypoint);
                 }
-            } else if (oldWaypoint.isUserDefined() || forceMerge) {
-                newPoints.add(oldWaypoint);
+            } catch (NullPointerException e) {
```
eddiemuc, line 200: Please remove. Instead make sure (and test throughly) that the code doesn't throw NPEs

main/src/main/java/cgeo/geocaching/models/Waypoint.java
```diff
@@ -323,6 +332,23 @@ public void setGeofence(@Nullable final Float geofence) {
         this.geofence = geofence;
     }
 
+    @Nullable
+    public Image buildImage() {
```
eddiemuc, line 335: suggestion: rename to "getImage()"

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -245,7 +245,7 @@ public static String getNameFor(final int id) {
     /**
      * The list of fields needed for mapping.
      */
-    private static final String[] WAYPOINT_COLUMNS = {"_id", "geocode", "updated", "type", "prefix", "lookup", "name", "latitude", "longitude", "note", "own", "visited", "user_note", "org_coords_empty", "calc_state", "projection_type", "projection_unit", "projection_formula_1", "projection_formula_2", "preprojected_latitude", "preprojected_longitude", "geofence"};
+    private static final String[] WAYPOINT_COLUMNS = {"_id", "geocode", "updated", "type", "prefix", "lookup", "name", "latitude", "longitude", "note", "own", "visited", "user_note", "org_coords_empty", "calc_state", "projection_type", "projection_unit", "projection_formula_1", "projection_formula_2", "preprojected_latitude", "preprojected_longitude", "geofence", "image"};
```
eddiemuc, line 248: DataStore misses much logic:
* is the new column created for new databases and added for existing databases?
* where is this new column used? Can't see any get and set method for it
* What is in the new image-column? an int? a text? what is it used for (it should be used to hold a reference to an offline stored image I assume)

main/src/main/res/layout/waypoint_item.xml
```diff
@@ -147,6 +148,13 @@
                 tools:visiblity="visible"
                 tools:ignore="HardcodedText"/>
 
+            <ImageView
+                android:id="@+id/waypoint_item_image"
```
eddiemuc, line 152: This new image field is never used/filled it seems?

main/src/main/java/cgeo/geocaching/models/Waypoint.java
```diff
@@ -323,6 +332,23 @@ public void setGeofence(@Nullable final Float geofence) {
...
+        if (StringUtils.isBlank(note)) {
+            return null;
+        }
+        // Extract first image from note HTML
+        final List<Image> images = cgeo.geocaching.utils.ImageUtils.getImagesFromHtml(
```
eddiemuc, line 340: It seems the image is not retrieved via new storage column in database but parsed again from the note. There's not much use in the whole change in this case. I thought the image should become a new explicit member of a waypoint

main/src/main/java/cgeo/geocaching/connector/al/ALApi.java
```diff
@@ -70,84 +68,7 @@ private ALApi() {
         // utility class with static methods
     }
 
-    @JsonIgnoreProperties(ignoreUnknown = true)
```
eddiemuc, line 73: Changes in this file seem unrelated to the issue, are they?

eddiemuc (changes requested):
Have many questions

main/src/main/java/cgeo/geocaching/storage/DataStore.java
JakeDot: I tried to eliminate the new column, but it seems copilot didn't clean everything up correctly. Saving the URL of stage images to DB isn't very useful, that's why I decided not to go with this approach anymore, as images should still be cached without this field.

main/src/main/res/layout/waypoint_item.xml
JakeDot: Correct, see above in DataStore

main/src/main/java/cgeo/geocaching/storage/DataStore.java
eddiemuc: You can also edit the PR manually :-)
In it's current state the PR can't be used

PR #17449, opened 2025-09-14
dynamic actionbar coloring

ztNFny:
dynamic colored action bar
image
image
image
image
image

Issue #17403, opened 2025-09-05
ActionBar: replace with Toolbar?


eddiemuc:
While working on edge-2-edge I noticed that our way of producing an actionbar is way outdated. Consequently, our code to eg doing edge-2-edge for it or hiding it is very "interesting" too.

While I have given up on switching to Kotlin and thus Compose, we could maybe instead at least switch to a Toolbar approach. This seems to be the second-chioce-recommendation of Google (first choice being Compose): https://developer.android.com/develop/ui/views/components/appbar

What is the team's feedback on this? 
Also is there a recommendation for a Toolbar to use if we want to make said switch?

moving-bits:
Material 2 seems to provide an app bar: https://m2.material.io/components/app-bars-top/android

(And, BTW, introducing Kotline & Compose gradually is still a valid approach, IMHO, just quite some refactoring + learning curve)

moving-bits:
Description of how to migrate `AppCompat.ActionBar` to `AndroidX.ToolBar` as app bar: https://developer.android.com/develop/ui/views/components/appbar/setting-up#java

PR #17458, opened 2025-09-16
Improved coloring style for action bar

murggel:
## Description

* Dynamic colored action bar (as option)
  - CacheDetailActivity
  - LogCacheActivity
  - CompassActivity
  - ImageGalleryActivity
  - Popupfragments
* Transparent Action bar in map view
* changes for light mode
  -  correct colored systembars
  - light tabbar in light mode
  - light action bar in light mode would also be possible (WIP)
* wherigo is not completely tested yet 

Setting "colored theme" in "appearance" (advanced settings), 

## Related issues

rel #17203

## Additional context

enhancement to / replacement of PR #17449

candidates for separate PR to merge prior into master:
* correct colored systembars
* Transparent Action bar in map view

### DIFF
--- a/main/src/main/java/cgeo/geocaching/SearchActivity.java
+++ b/main/src/main/java/cgeo/geocaching/SearchActivity.java
@@ -17,6 +17,7 @@ import cgeo.geocaching.filters.gui.GeocacheFilterActivity;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.search.GeocacheAutoCompleteAdapter;
 import cgeo.geocaching.search.SearchAutoCompleteAdapter;
+import cgeo.geocaching.search.SearchUtils;
 import cgeo.geocaching.sensors.LocationDataProvider;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
@@ -613,11 +614,15 @@ public class SearchActivity extends AbstractNavigationBarActivity {
         });
         searchButtonItem = menu.findItem(R.id.menu_gosearch_icon);
         searchView = (AutoCompleteTextView) searchViewItem.getActionView();
-        searchView.setBackground(AppCompatResources.getDrawable(this, R.drawable.mark_transparent));
-        // configure keyboard
-        searchView.setInputType(InputType.TYPE_CLASS_TEXT);
-        searchView.setImeOptions(EditorInfo.IME_ACTION_SEARCH);
-        searchView.addTextChangedListener(ViewUtils.createSimpleWatcher((s) -> updateSuggestions()));
+        if (null != searchView) {
+            searchView.setBackground(AppCompatResources.getDrawable(this, R.drawable.mark_transparent));
+            SearchUtils.setSearchViewColor(searchView);
+
+            // configure keyboard
+            searchView.setInputType(InputType.TYPE_CLASS_TEXT);
+            searchView.setImeOptions(EditorInfo.IME_ACTION_SEARCH);
+            searchView.addTextChangedListener(ViewUtils.createSimpleWatcher((s) -> updateSuggestions()));
+        }
         return true;
     }
 }
--- a/main/src/main/java/cgeo/geocaching/search/SearchUtils.java
+++ b/main/src/main/java/cgeo/geocaching/search/SearchUtils.java
@@ -105,10 +105,14 @@ public class SearchUtils {
     public static void setSearchViewColor(final SearchView searchView) {
         if (searchView != null) {
             final AutoCompleteTextView searchAutoComplete = searchView.findViewById(androidx.appcompat.R.id.search_src_text);
-            if (searchAutoComplete != null) {
-                searchAutoComplete.setTextColor(searchAutoComplete.getContext().getResources().getColor(R.color.colorTextActionBar));
-                searchAutoComplete.setHintTextColor(searchAutoComplete.getContext().getResources().getColor(R.color.colorTextActionBar));
-            }
+            setSearchViewColor(searchAutoComplete);
+        }
+    }
+
+    public static void setSearchViewColor(final AutoCompleteTextView searchAutoCompleteView) {
+        if (searchAutoCompleteView != null) {
+            searchAutoCompleteView.setTextColor(searchAutoCompleteView.getContext().getResources().getColor(R.color.colorTextActionBar));
+            searchAutoCompleteView.setHintTextColor(searchAutoCompleteView.getContext().getResources().getColor(R.color.colorTextActionBar));
         }
     }
 }


main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1009,6 +1010,10 @@ private void notifyDataSetChanged() {
 
         cache = search.getFirstCacheFromResult(LoadFlags.LOAD_ALL_DB_ONLY);
 
+        if (cache != null) {
+            getActionBarView().setBackgroundColor(ColorUtils.getActionBarColor(getResources().getColor(cache.getType().typeColor)));
```
eddiemuc: Careful when using getActionBarView(), please do a null check in it before usage.

The method is based on a hack and was introduced by me only very recently to solve problems with edge-2-edge and hiding actionbar on tap for map view. Can't say it is well-tested yet esp for all the older SDKs and devices we support. We shouldn't base mandatory functionality on it (yet).

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
```diff
@@ -108,7 +114,9 @@ private void applyTranslation() {
         if (actionBar != null) {
             actionBar.setTranslationY(-actionBarSystemBarOverlapHeight);
             actionBar.setPadding(0, actionBarSystemBarOverlapHeight, 0, 0);
-            actionBar.setBackgroundResource(R.color.colorBackgroundActionBar);
+
+            // don't set background here, as this would override possible action bar themes
```
eddiemuc: Tbh I can't even tell anymore why I set the background here. Probably just a forgotten leftover from debugging edge-2-edge. Would be wrong here anyway (background and Y-translation are two very different things). So I guess you can safely delete that line here w/o need to further commenting it

eddiemuc (approved):
Lgtm apart from minor comments

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: Then I maybe will look for a different solution

eddiemuc: If you find sth different I would be interested for the edge-2-edge problem as well. :-) unfortunately we currently use very outdated way for actionbars. See #17403

For this PR I am just asking for a null-check. If for some installations the view is null, then I guess it is ok to have cachetype-coloring not working for them.

murggel:
Transparent action bar in map-mode:
image image

murggel:
Colored light mode
image  image  image

image  image  image  

image

murggel:
Colored dark mode
image   image  image
image  image 
 image

murggel:
Wrong text color
Fixed: use correct color for TextView `android:id="@+id/info_notloggedin_text"`
image -> grafik

murggel:
Transparency is too less (or even missing?)
Fixed: Check transparency value for the distance-layout and use a different one
image -> grafik

murggel:
TODO: Maybe a slightly lighter color for the dark mode
image   image  image

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
ztNFny: without it coloring of status bar was broken

ztNFny:
> Transparent action bar in map-mode: image 

navbar icon color needs to be explictly set here

ztNFny:
> Colored dark mode image

The slider should probably be colored as well

main/src/main/java/cgeo/geocaching/utils/ColorUtils.java
```diff
@@ -52,4 +52,18 @@ public static String colorToString(@ColorInt final int color) {
     public static int colorFromResource(@ColorRes final int colorRes) {
         return CgeoApplication.getInstance().getResources().getColor(colorRes);
     }
+
+    public static int getActionBarColor(@ColorInt final int colorInt) {
```
ztNFny: perfectly fine for testing, but let's afterwards replace it with manually selected colors. All but orange seems ok to me so far

murggel:
Configure colored action bar:

grafik

murggel:
Fixed: Color the slider 
image -> grafik

main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
murggel: Ok, thanks for the hint, will take care of that...

murggel: Unfortunately I can't set explicitly the color for the status bar, only the "contrast", so I changed it via `setAppearanceLightStatusBars`.
Also in dark mode the status bar has also been colored not in the action bar color, but (I guess) in the background color.
Only necessary if the action bar is hidden.

ztNFny:
@cgeo/team is there an agreement to go with the colored actionbar (at least as an option) as shown in the screenshots above? I'd start selecting the colors then - but this is fine-tuning that I only want to do if the approach isn't abandonded again.

murggel:
> I'd start selecting the colors then - but this is fine-tuning that I only want to do if the approach isn't abandonded again.

Are you also looking for the neutral color or do we stay with the old (light or dark) grey?

I can to add also a (intern) setting in the meantime for light/dark action bar in light mode

ztNFny:
> Are you also looking for the neutral color or do we stay with the old (light or dark) grey?

I'll try - but that one is really hard.

> I can to add also a (intern) setting in the meantime for light/dark action bar in light mode

Could you add a "colorful actionbar - light mode" setting for testing? Would like to get some feedback on it. (I like dark better, but let's see ...)
Instead of
Math.max(0f, hsl[2] - 0.15f);
make it
Math.max(0f, hsl[2] + 0.20f);
in ColorUtils:getActionBarColor().

murggel:
> Could you add a "colorful actionbar - light mode" setting for testing? Would like to get some feedback on it. (I like dark better, but let's see ...)
> Instead of
> Math.max(0f, hsl[2] - 0.15f);
> make it
> Math.max(0f, hsl[2] + 0.20f);
> in ColorUtils:getActionBarColor().

How about making the value itself configurable?
And a second one for the dark theme, so you can use an independent value.

The settings for the values of course only for the testing time

ztNFny:
-15 works quite well imho, no need for further "unclean" parameterization there.

I was playing with a light version and that one would be around +20 - question is whether light would be wanted/liked.
Also we shouldn't have too many options but only those really worth it - it's additional effort for all (light action bar would require dark icons for example)

main/src/main/java/cgeo/geocaching/enumerations/CacheType.java
```diff
@@ -194,4 +197,22 @@ public boolean applyDistanceRule() {
...
+    public static int getActionBarColor(final Context context, final CacheType cacheType, final boolean isLightSkin) {
+        final int actionbarColor;
+        if (cacheType != null) {
+            // convert to HSL
+            final int colorInt = context.getResources().getColor(cacheType.typeColor);
+            final float[] hsl = ColorUtils.getHslValues(colorInt);
+
+            // darker color by 15%
+            final float offSet = isLightSkin ? -0.15f : -0.10f;
+            hsl[2] = Math.min(1f, hsl[2] + offSet);
```
ztNFny: this should be Math.max(0f, ...) - you're substracting after all ;-)

main/src/main/java/cgeo/geocaching/utils/ColorUtils.java
```diff
@@ -63,6 +63,11 @@ private static float[] getHslValues(@ColorInt final int colorInt) {
         return hsl;
     }
 
+    public static int getColorFromHslValues(final float[] hslValues) {
+        return androidx.core.graphics.ColorUtils.HSLToColor(hslValues);
+
+    }
+
     public static boolean isBrightnessDark(@ColorInt final int colorInt) {
```
ztNFny, line 71: this is probably problematic. yellow/orange with L<0.5f will still be bright.
But with the color values currently being discussed this method is probably not needed after all ... If it is let's look for a better way to do this.

main/src/main/java/cgeo/geocaching/enumerations/CacheType.java
murggel: > this should be Math.max(0f, ...) - you're substracting after all ;-)

I have that in my local already 😊

ztNFny:
The "pull-up-bar" on cache popup has some transparency on it - not sure whether that's an edge2edge bug or because of this PR

murggel:
> The "pull-up-bar" on cache popup has some transparency on it - not sure whether that's an edge2edge bug or because of this PR

That is probably related to the transparent action bar in map-mode

Will change that to a different implementation 

Edit: but in a separate PR, so reverting the corresponding commit here...

main/src/main/java/cgeo/geocaching/utils/ColorUtils.java
murggel: Or we can adapt the value or ...
Yeah, we will wait for the colors and decide then

murggel: Currently it is only used on the defined default-actionbar-color to set the appearance of the systembars... So all used colors for colored actionbar should be able to use the same text-color
Maybe I remove this method than and implement only the "special" case

murggel:
@ztNFny any new feedback here?
Which values shall we use?

ztNFny:
> @ztNFny any new feedback here?
> Which values shall we use?

No time currently

moving-bits:
Shall we squash & merge here to nightlies to get user feedback? We can still fine-tune colors along the further way.

murggel:
Coloring for WhereIGo is still missing

ztNFny:
Please go ahead without my custom colors..i'd just like to add them before merging to release.

murggel:
Shall the default be the colored action bar or current color?

ztNFny:
> Shall the default be the colored action bar or current color?

Imho: colored
General direction should be fine, tuning can still be done. But i simply don't have the capacity to look into it.

murggel:
> Coloring for WhereIGo is still missing 

Maybe in a different PR since we can think about different colors for the different states of the Wherigo.
I tried to color the actinbar / titlebar in the Wherigo-blue, but didn't like it.

image

image

murggel:
> > Shall the default be the colored action bar or current color?
> 
> Imho: colored General direction should be fine, tuning can still be done. But i simply don't have the capacity to look into it.

Then let me check tomorrow the current state and I will give feedback, if it can be merged to give it a try in the nightly...

I will then remove the labels...

murggel:
One glitch, but don't know if that was the same with the old dark actionbar...

In light mode the search "preview" text is hardly visible 




I thought I had a commit adressing that problem... maybe I put that in a separate branch - have to check when I am back home

But the search text itself is visible

murggel:
> One glitch, but don't know if that was the same with the old dark actionbar...
> 
> In light mode the search "preview" text is hardly visible
> 
> I thought I had a commit adressing that problem... maybe I put that in a separate branch - have to check when I am back home
> 

Fixed, I missed the search-cards...

moving-bits:
Retest this please

murggel:
I think, we can merge and get feedback from nightly users...

moving-bits:
Failing tests are unrelated, so let's get feedback from nightly users

murggel:
There is a setting "colored theme" in "appearance" (advanced settings), which is enabled by default and be turned off for "old" look and feel

Lomanic:
FYI in the nightly the changelog is wrong (and it took some time to find this PR): `New: colored toolbar in cache details pages` appears in the `General` section instead of the `Cache details` one

moving-bits:
> FYI in the nightly the changelog is wrong (and it took some time to find this PR): `New: colored toolbar in cache details pages` appears in the `General` section instead of the `Cache details` one

Yeah, this is a bit tricky, as "cache details" can refer to both the program part "Cache Details" and the places where details of a cache are being displayed. The new functionality is visible on both the "Cache Details" program part as well as on the map (info sheet that appears when tapping on a cache), therefore we placed it in the "general" section. But this choice can be discussed, of course.

PR #17469, opened 2025-09-19
fix #15412: avoid doubled route items

### LINKED ISSUE #15412: Route items doubled after adding one cache and adjusting position
### Describe your problem!

Today I prepared a hiking trip with 15 caches by creating an individual route along those caches and the parking coordinates.
The final route was 8,17km long, which is as expected:
image

Then I noticed I forgot to add one cache (marked in screenshot with red dot) and did the following:
- Long tap 
- Append to route
- Tap on route icon
- Edit route
- Move the new cache to the matching position

Afterwards the modified route was shown correct (i.e. got some hundred meters longer than before) but suddenly it doubled its length after a few seconds.
image

Going into route editing again I do see all items listed twice now. So basically there was another iteration of the same route appended at the end of the existing route. While trying to reproduce to make a screenshot I was unable to see this duplication of route items in the edit mode but I can still reliably reproduce doubling of the length using the above mentioned steps.

Closing and reopening the map solves the problem permanently.

### How to reproduce?

see above

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

2024.03.03-NB

### System information

_No response_

### Additional Information

_No response_

murggel:
## Description

Avoid clearing / updating route, if it is already loading / updating from database
(but I don't know yet, why / how that happens, has something to do with AndroidRxUtils.andThenOnUi in IndividualRoute::reloadRoute

## Related issues

fix #15412

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -1450,12 +1450,14 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
             recreate(); // restart with a fresh MapView
         }
 
-        if (Settings.removeFromRouteOnLog()) {
-            viewModel.reloadIndividualRoute();
-        }
         super.onResume();
         reloadCachesAndWaypoints();
         MapUtils.updateFilterBar(this, viewModel.mapType.filterContext);
+        
+        if (Settings.removeFromRouteOnLog()) {
+            viewModel.reloadIndividualRoute();
+        }
+
         if (tileProvider != null) {
             tileProvider.onResume();
         }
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -1450,12 +1450,14 @@ protected void onResume() {
             recreate(); // restart with a fresh MapView
         }
 
-        if (Settings.removeFromRouteOnLog()) {
-            viewModel.reloadIndividualRoute();
-        }
         super.onResume();
         reloadCachesAndWaypoints();
         MapUtils.updateFilterBar(this, viewModel.mapType.filterContext);
+
```
murggel, line 1456: I moved the reload of the individual route after reloading caches and waypoints. 
I didn't notice any influence of the behavior, but from my understanding it would make more sense. Maybe if the caches coordinates / status would have changed in the meantime, but don't know, when that can happen...

moving-bits:
Let's give it a try in the nightlies. Thanks for tackling that issue!

Issue #17322, opened 2025-08-22
Render "empty" map space transparent


JakeDot:
> I would like to ask if this behaviour of the background maps is considered a bug - I would have thought the background maps should be displayed everywhere a normal map has no map data, i.e. the "unused" space in normal maps should be rendered transparent instead of grey.
> 
> <img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/5e412c87-5940-4ebe-b2e6-9f9d622ea6c6" /> 

 _Originally posted by @JakeDot in [#17102](https://github.com/cgeo/cgeo/issues/17102#issuecomment-3109588661)_

Additionnally, if you have multiple offline maps (FZK in my case) the empty space is bleeding into other maps when using the combined offline map source.

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/6b8e5ae8-ae92-404a-b886-439080981d45" />

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/dc0a1bab-917e-421d-a024-fa54455f694a" />

moving-bits:
Not sure whether this is a theme issue, a map data thing or a library issue.

@JakeDot: Which library are you using in the screenshots - VTM or Mapsforge (bitmap)

@devemux86: Can you shed some light on the map data vs. theme vs. library question - what influences non-transparent coloring of free map space?

JakeDot:
It's VTM only.

-------- Ursprüngliche Nachricht --------
Am 22.08.25 22:21 schrieb moving-bits :

devemux86:
You can try some tests to see what happens:
- See if Mapsforge is different than VTM

- Try other map providers, like Mapsforge or OpenAndroMaps

- Change the `map-background` color to transparent in render themes (at the top)

We have not seen any problem with multiple neighboring map files using Mapsforge or VTM in Cruiser.
e.g. Switzerland + Austria

devemux86:
To combine vector and bitmap maps,
you probably need to modify the render theme and make the sea and land transparent.

---

Mapsforge and probably all maps are created connected with the themes with this [process](https://github.com/mapsforge/mapsforge/blob/master/docs/MapCreation.md).

Maps are rectangles, sea rectangle at bottom, land on top and then the map layers.
In above image there is the Austria map with land (and its color) over the whole map.

There is no distinction between data inside a country and empty area outside the country.
All are considered as land (filled or not) and have been colored the same way.

[...]

PR #17480, opened 2025-09-21
Test drive VTM transparent map background (rel. to #17322)

moving-bits:
## Description
Renders background of VTM maps as transparent (empty tags and key=natural + value in SEA, NOSEA) to avoid the blank rectangles around vector maps when being displayed on top of a background map.

This feature only gets applied when two conditions are met:
- background maps are enabled
- you have set a preference key named `vtmBackgroundTransparent` of type `boolean` to `true`

|before|after|
|---|---|
|Image|image|

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
@@ -31,6 +31,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
```diff
@@ -138,6 +141,19 @@ protected void reapplyMapTheme(final Map map, final AbstractTileProvider tilePro
...
                 mTheme = map.setTheme(xmlRenderTheme);
+
+                if (Settings.getMapBackgroundMapLayer() && Settings.isFeatureEnabledDefaultFalse(R.string.pref_vtmBackgroundTransparent)) {
+                    map.setTheme(ThemeLoader.load(xmlRenderTheme, new ThemeCallbackAdapter() {
+                        @Override
+                        public int getColor(final String[] keys, final String[] values, final int color) {
+                            final List<String> k = Arrays.asList(keys);
+                            final List<String> v = Arrays.asList(values);
+//                            Log.e("keys=" + Arrays.asList(keys) + " -> " + Arrays.asList(values));
+                            return (((k.isEmpty() && v.isEmpty())) || (k.contains("natural") && (v.contains("sea") || v.contains("nosea")))) ? 0x000000ff : color;
```
JakeDot: Should this be 0x000000ff or 0x00000000 instead?

eddiemuc:
ready to merge?

main/src/main/java/cgeo/geocaching/unifiedmap/mapsforgevtm/MapsforgeThemeHelper.java
moving-bits: updated to 0x0

moving-bits:
> ready to merge?

Have added preference handling and done some code cleanup. Ready to merge if build is green.

moving-bits:
failing tests are unrelated

PR #17490, opened 2025-09-21
initial fastlane structures from IzzyOnDroid

IzzySoft:
this PR provides you with a "fastlane starter package" – to give it into your hands to define how your app is presented, and make it easy to keep descriptions & graphics in sync with your development. Some notes to sum them up in a place easy to find for you:

* the [IzzyOnDroid Fastlane Documentation](https://gitlab.com/IzzyOnDroid/repo/-/wikis/Fastlane) can guide you with maintaining the structures here
* take special care for limits, e.g. per-release changelogs can have max 500 chars, `full_description.txt` max 4,000 chars, graphics must have specific aspect ratios etc.
* you can have "bigger graphics" here; the IoD updater takes care to resize/optimize them as needed
* for the `full_description.txt` here I've used HTML compressed into a single line, to prevent the fdroid software converting each line break to a `<br>`. The tags used for this are supported by F-Droid.org as well, and to my knowledge even by PlayStore.
* whenever you add a new "type" of metadata that has not been there before (e.g. the first changelog, a featureGraphic if it wasn't yet present), please let us know so we can check and enable it.
* Details on all this in the linked documentation.
* Once merged, your changes will be synced whenever a new release is pulled in here – to make sure it stays up-to-date with the app itself.
* Should you have questions, be welcome to ask :wink:

And now: Enjoy!


eddiemuc:
ok to test

bekuno:
retest this please

IzzySoft:
May I ask what the status is here? Would be great to have it merged, so we can move on.

For clarification: the content of this PR in no way conflicts with your code, it's entirely separate – so there's no danger of side-effects. As the original comment states, it's just providing app metadata (description & graphics) in a standardized directory structure, outside the code tree.

IzzySoft:
@eddiemuc @bekuno any chance?

IzzySoft:
May I kindly ask what's holding this back, @eddiemuc @bekuno @moving-bits? I doubt the failing unit tests have to do with this PR, which just ships metadata (app description & graphics) outside the code path.

moving-bits:
Sorry, @IzzySoft, for coming back to you so late - we are currently pretty short on spare time.

We had a short discussion in today's dev meeting about this PR, and came up with some questions:
- Do we understand it correctly, that this would lay the foundation for c:geo app description in your system, and that every update to these files would be reflected automatically in your system (after some sync time)? 
- Will this work for IzzyOnDroid only, or will F-Droid use this description as well? (We are not listed at F-Droid directly, but have our own F-Droid repo, AFAIK).
- Is the base folder for the fastlane structure fix or can we move it to some other place initially?
- Release-specific descriptions are optional to my understanding, correct? I'm not sure if we will be able to maintain this constantly. - Can an existing changelog file (in Markdown format) be used as release description (by linking to a specific file, tagged with a release-specific tag)?
- In case we would provide release descriptions, how will this be linked to a specific build? (We set tags in our repo to mark releases. All releases are built from `release` branch, not from `master`.)

IzzySoft:
Thanks for responding, @moving-bits! In order of your bullet points:

* yes, and yes (it is synchronized whenever a new release is being pulled in)
* yes, it applies to both (incl. the "sync on new release")
* pretty much fix if you ask F-Droid (see the linked documentation¹); flexible for IzzyOnDroid (as long as you let us know when you moved it, and where to – we can configure the "entry point" flexibly)
* I assume by "release specific" you mean the per-release changelogs. They are optional – so no prob if you skip one, or two, or … If there is none, we'll just not show any. And no, Markdown only works for the full description (and then also only for IzzyOnDroid²); changelogs must be plain text (also no HTML tags).
* the branch used for a release is the one the release tag points to, right? We usually clone from the tag (and only fall back to `HEAD` in case cloning from the tag fails).

----

¹ to make it easier for you to find, [here's the corresponding part](https://izzyondroid.org/docs/general/Fastlane/#locations):

> For F-Droid.org, there are 3 possible entry-points the process looks for Fastlane, always located in the root of the app's repository:
>
> * `/fastlane/metadata/android/<locale>/` (which is what this PR uses)
> * `/metadata/<locale>/` (which is F-Droid's internal representation)
> * `/src/<buildFlavor>/fastlane/metadata/android/<locale>/`

----

² IzzyOnDroid supports full (standard) Markdown, but also full HTML (with some elements sanitized, think e.g. of `<script>` and `onClick=`) and plain text. F-Droid only supports the latter two – and also replaces each line break with `<br>`, which is why we recommend "HTML all in one line". There's a "middle way" I call "Markdown Lite", working for all places (including Playstore). Again, quoting for you so you don't have to search for it:

> […] Make it at least close to Markdown, e.g. separate paragraphs by an empty line in between, so it can be parsed as such. Basic Markdown is not that hard to achieve! You can even use simple HTML formatting (like `<b>` or `<i>`) here, even `<a href` (though Play Store would remove the tags for these). This is a very good compromise if you want to use Fastlane with multiple targets as it works for F-Droid, IzzyOnDroid and Play Store – all 3 of them will sanitize, of course. This is what I call **Markdown Lite**.

moving-bits:
Thanks @IzzySoft for the fast reply and the useful pointers. I'd say: Let's start with this PR (thanks for preparing it!) and see if we can fine-tune it later on.

IzzySoft:
Thanks @moving-bits :hugs: All set at our end then (and an initial refresh was run to ensure it works). Any future changes you might make, would be synchronized automatically now whenever a new release is being pulled in.

Issue #17492, opened 2025-09-23
Tapping on Geocache in map triggers "Geocache data missing"


eddiemuc:
We have reports on support from users tapping on a Geocache in the map and getting the message "Geocache data missing" instead of the cache opening.

https://cgeosupport.droescher.eu/scp/tickets.php?id=9861
https://cgeosupport.droescher.eu/scp/tickets.php?id=9847

This message was introduced by myself. Some might remember the problem of eg @murggel where she claimed that tapping on a geocache on the map didn't lead to any action. Assuming this ro be a problem of the geoitem layer I dived into it. But thr layer worked fine, instead I found that indeed geocache data is sometimes missing for caches on the map. Don't know the reason, probably a caching problem? 

I introduced the dialog to make thr problem visible. Apparently this was successful. However, the cause of missing data is not yet resolved.

moving-bits:
Might this be related to some other reports on support claiming that no location info is shown for them sometimes in cache popup / cache details?

One example for this is ticket [660237](https://cgeosupport.droescher.eu/scp/tickets.php?id=9868), but there have been more in the recent months.

murggel:
Had it today with a AdvLab near Paris while I am in Germany. 
On first tap all seems OK, then zooming and moving around, and then tapping on the cache leads to "cache data missing"

MagpieFourtyTwo:
Unfortunately, I don't have a solution, not even a reproducible scenario, but at least a workaround:
When a cache icon gives this message, just do ``Show as list`` and go back to the map - now the icon behaves as expected.

moving-bits:
When this happens to you, do the scenarios have anything in common? Eg:
- live mode enabled?
- cache in question is stored locally?
- certain actions done before (like having zoomed out so far, that > 2000 caches have been displayed in total since opening the map, though not necessarily in the same viewport)
- ...

MagpieFourtyTwo:
From memory:

> live mode enabled?

Yes.

> cache in question is stored locally?

No (I wouldn't bet my life on it, but at the moment I can't remember any offline cache at all that would have led to this message)

> certain actions done before (like having zoomed out so far, that > 2000 caches have been displayed in total since opening the map, though not necessarily in the same viewport)

Most definitely not - last time this happened to me I had just opened the map and only two caches were displayed, neither of which had details. However, it is rather rare that all of them have no details, usually, it is one or two out of a dozen - and typically without any excessive map actions beforehand.

moving-bits:
[...]

PR #17520, opened 2025-10-07
rel to #17492: various measurements to prevent geocache missing data on selecting cache on map

eddiemuc:
rel to #17492: various measurements to prevent geocache missing data on selecting cache on map


eddiemuc:
codacy is green, failing test in default is unrelated (BRouter-related)

Issue #17488, opened 2025-09-21
Many ANR happening on c:geo startup


moving-bits:
Reported by Play Store stats for 2025.08.31 with more than 1000 (!) occurrences so far:

- SourceFile - cgeo.geocaching.storage.DocumentContentAccessor.queryDir (error 1, 890 occurrences)
- SourceFile - cgeo.geocaching.storage.DocumentContentAccessor.delete (error 2, 139 occurrences)
- ANR triggered by thread waiting for a binder transaction
- Different vendors (Redmi, Samsung, Google etc.), Android 14-16 only (so far)

Log of main thread (with comments):
```
#00  pc 0x00000000000c99a8  /apex/com.android.runtime/lib64/bionic/libc.so (__ioctl+8)
#01  pc 0x000000000006b0bc  /apex/com.android.runtime/lib64/bionic/libc.so (ioctl+156)
#02  pc 0x0000000000076dd4  /system/lib64/libbinder.so (android::IPCThreadState::transact+1220)
#03  pc 0x0000000000063ce8  /system/lib64/libbinder.so (android::BpBinder::transact+168)
#04  pc 0x00000000001af238  /system/lib64/libandroid_runtime.so (android_os_BinderProxy_transact+152)
at android.os.BinderProxy.transactNative (Native method)

Dieser Binder-Aufruf dauert möglicherweise zu lange, sodass der Hauptthread in den Wartezustand wechselt und der ANR-Fehler ausgelöst wird. [Weitere Informationen](https://developer.android.com/topic/performance/anrs/find-unresponsive-thread#slow-binder)

      at android.os.BinderProxy.transact (BinderProxy.java:682)
      at android.content.ContentProviderProxy.query (ContentProviderNative.java:499)
      at android.content.ContentResolver.query (ContentResolver.java:1229)
      at android.content.ContentResolver.query (ContentResolver.java:1161)
      at android.content.ContentResolver.query (ContentResolver.java:1117)
      at cgeo.geocaching.storage.DocumentContentAccessor.queryDir (SourceFile:406)
      at cgeo.geocaching.storage.DocumentContentAccessor.findCreateSubdirectory (SourceFile:372)
      at cgeo.geocaching.storage.DocumentContentAccessor.getSubdirUri (SourceFile:355)
      at cgeo.geocaching.storage.DocumentContentAccessor.getFolderUri (SourceFile:321)
[...]

PR #17521, opened 2025-10-07
rel to #17488: Log configuration on startup: avoid blocking main thread for file access

eddiemuc:
rel to #17488: Log configuration on startup: avoid blocking main thread for file access


eddiemuc:
codacy green, failing test on "default" is unrelated

Issue #17485, opened 2025-09-21
NPE in PersistableFolder.registerForIndirectChange


moving-bits:
Reported by Play Store stats for 2025.08.31 (77 occurrences so far):

```
Exception java.lang.ExceptionInInitializerError:
  at cgeo.geocaching.utils.Log.<clinit> (SourceFile:84)
  at cgeo.geocaching.CgeoApplication.onCreate (SourceFile:128)
  at android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1266)
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:7619)
  at android.app.ActivityThread.-$$Nest$mhandleBindApplication (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2400)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8762)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:604)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1067)
Caused by java.lang.NullPointerException:
  at cgeo.geocaching.storage.PersistableFolder.registerForIndirectChange (SourceFile:100)
  at cgeo.geocaching.storage.PersistableFolder.<init> (SourceFile:92)
  at cgeo.geocaching.storage.PersistableFolder.<clinit> (SourceFile:38)
```

eddiemuc:
This results from new persistable folder BACKGROUND_MAPS. It occurs when `CgeoApplication.getInstance().getApplicationContext().getExternalMediaDirs()[0]` returns `null`

<img width="1800" height="91" alt="Image" src="https://github.com/user-attachments/assets/2f6e9307-7a89-4cd6-8384-8b041213d629" />

The usage of `CgeoApplication.getInstance().getApplicationContext().getExternalMediaDirs()` is deprecated. I am not 100% sure what is the intended use case. I will issue a PR to do the null check (and probably clean up this default folder stuff a bit). However could someone advise what the intention was here wrt background maps?

eddiemuc:
[...]

PR #17523, opened 2025-10-07
rel to #17485: persistablefolder-prevent-npe-on-null-defaultfolder

eddiemuc:
rel to #17485: persistablefolder-prevent-npe-on-null-defaultfolder


eddiemuc:
codacy green, failing test on "default" is unrelated (BRouter)

PR #17529, opened 2025-10-08
Wherigo Player: provide location accuracy/precision

eddiemuc:
Wherigo Player: provide location accuracy/precision

Via support I was informed that some cartridges need the location accuracy/precision to function. This PR lets the player provide this value.


eddiemuc:
codacy and default green, merging here

PR #17533, opened 2025-10-08
fix #17500: show coord warning in waypoint-list

### LINKED ISSUE #17500: Visualization of formula overflow
### Tell us your idea!

It happened to me a few times now that I didn't notice the formula overflow while editing the variables and looking at the waypoints.
Therefore, I would like the warning to be displayed also in the waypoint list and waypoint editing dialog.

Image

Image

Image

### Is there an alternative?

_No response_

### Additional Information

_No response_

murggel:
## Description

Show warning-icon, if the coordinates of the waypoint have the warning.

Image

Image

## Related issues

fix #17500


eddiemuc:
Lgtm

Issue #10595, opened 2021-05-11
Allow previewing bookmark lists


fm-sys:
Bookmark lists can be imported as GPX since #10539

But they also can be fetched via `https://www.geocaching.com/api/proxy/web/v1/lists/BMxxxxx/geocaches?skip=0&take=100` which would allow to offer the "preview" action for bookmark lists as well...

![grafik](https://user-images.githubusercontent.com/64581222/117193657-6cd4d100-ade3-11eb-856b-247dcc0c4b3e.png)

fm-sys:
This might be even easier, as soon as we refactored the gc.com search as discussed in https://github.com/cgeo/cgeo/issues/10027#issuecomment-835285341

jakedot:
good thing you opened this issue, I'd been noticing this as well yesterday. Additional remark: the BMs have a "Bookmark List" label, but the PQs don't - maybe we should add a label for them too? Old users would recognize them, but what about new ones?

Lineflyer:
@jakedot I don't understand your sentence. You mean PM not PQ, do you? But still then, what do you mean with "Bookmark list" label?

As far as I understood @fm-sys the bookmark list reading will be limited to PM, as BM only have their watchlist but no individual bookmark lists.

moving-bits:
The entries resulting from a bookmark list have a tiny label "Bookmark list", see the last two entries in the screenshot above.
In contrast to this, the entries resulting from PQ don't have a "Pocket query" label or something like this.

Lineflyer:
Ah....I totally misunderstood as I did not connect it to the screenshot in my mind :D

murggel:
Since the separation of bookmark lists and pocket queries in #11458, the label is not necessary any more, hence I would remove it.

murggel:
If PR #17425 is merged, I would start working on this issue - I already have a spike...

eddiemuc:
> If PR [#17425](https://github.com/cgeo/cgeo/pull/17425) is merged, I would start working on this issue - I already have a spike...

That PR is marked as WIP + Feedback required, really ready to merge?

murggel:
> > If PR [#17425](https://github.com/cgeo/cgeo/pull/17425) is merged, I would start working on this issue - I already have a spike...
> 
> That PR is marked as WIP + Feedback required, really ready to merge?

Not yet, but as soon as I have tested the latest changes, I will remove the labels and mark as ready for review...

PR #17534, opened 2025-10-08
rel #10595: Remove label "bookmark list"

murggel:
## Description

Remove label "bookmark list", since it is now in a separate activity
image

## Related issues

rel #10595


eddiemuc:
Lgtm

murggel:
I missed to check, whether the string is used somewhere else or it can be deleted 

Will check ...

Edit - is still in use

Issue #17545, opened 2025-10-12
[Nightly] Found glitches with new color scheme


murggel:
Collection for issues with new color scheme:

- [x] PR #17546 - Wrong color for subtitle in list view after search (correct color in "normal" list view)
<img width="250" alt="Image" src="https://github.com/user-attachments/assets/cbfc3854-b7b4-4287-bbe2-2192e3aed0e3" /> <img width="250" alt="Image" src="https://github.com/user-attachments/assets/7073483e-85c6-4aeb-939d-de3f8a82abba" />
- [x] PR #17553 - Remove transparency for action bar in map-mode
- [x] PR #17552 - Light status bar in map-mode
- [x] PR #17557 - Correct color for nightly-message (color for icon is still wrong)

MagpieFourtyTwo:
Coloring sounds like a good and fancy idea, and on first sight I really liked it, but after working with it for a few days now I would really love to go back, cause especially with dark design, it's much too intense - even in daylight. And at night ... it's a real pain.  Just check yourself: Switch off lights, get used to the dark and then open the listing of a Multi ... you'll be blinded for seconds. 🙈  And this every single time you're going to edit your PN ... sorry, but that's no fun at all. And my major "glitch".  🤷🏼‍♂️

Apart from that: Esp. with archived caches the header text is hard to read sometimes - even the screenshot image has problems to distinguish the colors 😉

![Image](https://github.com/user-attachments/assets/4cf46ba1-14bf-429d-9db6-33250d6c5531)

murggel:
> Coloring sounds like a good and fancy idea, and on first sight I really liked it, but after working with it for a few days now I would really love to go back, cause especially with dark design, it's much too intense - even in daylight. And at night ... it's a real pain. 
> 

Thanks for the feedback - we are still working on the colors...


There is a setting "colored theme" in "appearance" (advanced settings), which is enabled by default and be turned off for "old" look and feel

![Screenshot_20251011_140320_cgeo.jpg](https://github.com/user-attachments/assets/bdc7a122-7212-4887-b6cd-b8dd05e98320)

MagpieFourtyTwo:
Aaah - great! Thanks!  🤪

bekuno:
Many color problems on map view in light mode (shown with active routing):
- System headline: white text on white background
- c:geo head - 4 different bachground color areas and different forground colors - maybe semi-transparent on top of each other
- c:geo foot: white on white instead default (black/orange) on light grey
[...]

PR #17546, opened 2025-10-12
rel #17545: adapt subtitle-color

murggel:
## Description

Fix glitches of new color scheme:
* Adapt subtitle-color

Image Image

## Related issues

rel #17545


main/src/main/java/cgeo/geocaching/utils/ActionBarUtils.java
murggel, line 1: Introduced helper-methods in ActionBarUtils to set subtitle and title of action bar.

PR #17557, opened 2025-10-20
rel #17545: fix status message color

murggel:
## Description

Use `colorTextMainInfo` as status message color.

IconTint is currently not set, because it does not work for multi-color icons (e.g. used first-aid-icon)

## Related issues

rel #17545


bekuno:
retest this please

Issue #17565, opened 2025-10-26
Missing permissions


bekuno:
With the AGP 8.13.0 in #17561, we now get explicit `SecurityException`s: `MissingPermission`.

Details from a separate lint check:
~~~
src\main\java\cgeo\geocaching\service\AbstractForegroundIntentService.java:58: Error: Call requires permission which may be rejected by user: code should explicitly check to see if permission is available (with checkPermission) or explicitly handle a potential SecurityException [MissingPermission]
        notificationManager.notify(getForegroundNotificationId(), notification.build());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src\main\java\cgeo\geocaching\service\CacheDownloaderService.java:266: Error: Call requires permission which may be rejected by user: code should explicitly check to see if permission is available (with checkPermission) or explicitly handle a potential SecurityException [MissingPermission]
        notificationManager.notify(Settings.getUniqueNotificationId(), Notifications.createTextContentNotification(
        ^
src\main\java\cgeo\geocaching\sensors\GeoDataProvider.java:43: Error: Call requires permission which may be rejected by user: code should explicitly check to see if permission is available (with checkPermission) or explicitly handle a potential SecurityException [MissingPermission]
                geoManager.requestLocationUpdates(LocationManager.GPS_PROVIDER, 0, 0, gpsListener);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src\main\java\cgeo\geocaching\sensors\GeoDataProvider.java:48: Error: Call requires permission which may be rejected by user: code should explicitly check to see if permission is available (with checkPermission) or explicitly handle a potential SecurityException [MissingPermission]
                geoManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, 0, 0, networkListener);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #17566, opened 2025-10-26
rel to #17565: handle missing notification permission / SecurityException: try sending toast instead

eddiemuc:
rel to #17565: handle missing notification permission / SecurityException: try sending toast instead


eddiemuc:
Default + codacy green

Issue #17143, opened 2025-07-18
Next release after 2025.07.17


moving-bits:
(NB: This issue is meant for @cgeo/team only as internal working space)

Now that the new feature release 2025.07.17 has been packaged, let's collect things that need to be done next:

- [x] #17142
- [x] #17181

moving-bits:
Starting to work on a bugfix release now. It will include two fixes:
- #17145
- #17148


- [x] Check/insert correct version name in `BranchDetectionHelper.java`
- [x] Wait for a Crowdin cycle or trigger sync manually
- [x] Merge the related Crowdin translation pull requests for branch `release`
- [x] Build release on CI server
- [x] Perform smoke test with resulting ```cgeo-release.apk``` before continuing
- [x] Upload and publish ```cgeo-release.apk``` to Google Play
- [x] Tag the release on Github in format `market_yyyymmdd`
- [x] Publish the release on Github including upload of ```cgeo-release.apk``` to Github
- [ ] Publish release on F-Droid => not working currently due to #17142
- [ ] Trigger notification to status.cgeo.org via CI server
- [ ] Post release info to Facebook 
- [ ] Post release info to Twitter/X

ztNFny:
- Fix for TB inventory - urgent: https://github.com/cgeo/cgeo/pull/17183
- Fix for Install wizard icon size - low risk: https://github.com/cgeo/cgeo/pull/17178

fm-sys:
For install wizard icon size we have two PRs open. Main difference is the icon shape:
- https://github.com/cgeo/cgeo/pull/17165
- https://github.com/cgeo/cgeo/pull/17178

No preference from my side. Which one should we use?

ztNFny:
> For install wizard icon size we have two PRs open. Main difference is the icon shape:
> - https://github.com/cgeo/cgeo/pull/17165
> - https://github.com/cgeo/cgeo/pull/17178
> 
> No preference from my side. Which one should we use? 

Oh sorry, didn't see yours!
I'd go with mine as it's already against release + includes another minor fix.
Icon shape i prefer square - otherwise wouldn't have done it this way 😁

fm-sys:
Ok, then let's go with yours

Lineflyer:
Preparing a release as discussed on event chat:
- [x] Check/insert correct version name in `BranchDetectionHelper.java`
- [x] Wait for a Crowdin cycle or trigger sync manually
[...]

PR #17567, opened 2025-10-29
Update VTM on release

moving-bits:
## Description
Update VTM on `release` to the same version as on `master` due to unrecoverable build errors when using version 25.0 - even on local builds I'm no longer able to use v25.0 (after deleting my local gradle cache) - even though this has worked in the past by re-triggering module build on jitpack.io

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -385,7 +385,7 @@ dependencies {
     // -- Mapsforge / Mapsforge --
 
     // Mapsforge official version
-    String mapsforgeSource = 'org.mapsforge'
+    String mapsforgeSource = 'com.github.mapsforge.mapsforge'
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

moving-bits:
See https://github.com/cgeo/cgeo/issues/17143#issuecomment-3463002329

bekuno:
The problem is on another place line 388.
You have to replace it by the content of line 392.
Please compare with the VTM declaration.

moving-bits:
> The problem is on another place line 388. You have to replace it by the content of line 392. Please compare with the VTM declaration.

Thanks, have updated accordingly, and now PR build is green. Let's merge and give the bugfix release another try.

PR #17569, opened 2025-10-30
Set Mapsforge dependencies to @jar

moving-bits:
## Description
As suggested in https://github.com/cgeo/cgeo/issues/17143#issuecomment-3466474380, add `@jar` to the Mapsforge dependencies to check whether Jitpack will deliver the files for the used versions again in that case


moving-bits:
Basic build test is green, so let's merge it to `release` and see if we can create a release build from it.

PR #17575, opened 2025-11-02
fix #10595: Preview of Bookmarklists

### LINKED ISSUE #10595: Allow previewing bookmark lists

murggel:
## Description

Add possibility of preview for single bookmarklists.
For multiple selected lists it does not work currently, because the preview does not show all caches on the list and the "load more" does work only for single lists and not for multi-selection.

image
image

"Load more"

image image

## Related issues

fix #10595


murggel:
Ready for review

PR #17561, opened 2025-10-25
Update AGP/Gradle

bekuno:
- Gradle 8.13
- AGP 8.13.0
- modify all build.gradle for fixing deprecated syntax

References:
- https://developer.android.com/studio/releases#android_gradle_plugin_and_android_studio_compatibility
- https://developer.android.com/studio/releases#api-level-support
- https://developer.android.com/build/releases/gradle-plugin#updating-gradle
- https://developer.android.com/build/releases/gradle-plugin#android_gradle_plugin_and_android_studio_compatibility
- (https://developer.android.com/build/kotlin-support)

PR #17577, opened 2025-11-07
set minimal AGP version for kotlin 2.2

bekuno:
see https://developer.android.com/build/kotlin-support

related to #17561


bekuno:
The lint errors already exist with this AGP version.
Let's close in Let's close in favor of #17561.

PR #17587, opened 2025-11-15
Remove legacy maps

moving-bits:
## Description
Prepares removal of legacy maps.

This will be the last step in our switch to UnifiedMap after implementation, deprecation and setting UnifiedMap active (#17584).

Not to be merged yet, as some fine-tuning & clean-up may be necessary, therefore setting WIP label.


moving-bits:
finally, as discussed in today's dev meeting: let's merge 🚀

PR #17599, opened 2025-11-23
fix #17598: preserve personal note whitespaces on refreshing/downloading cache data

### LINKED ISSUE #17598: Personal Note not properly compared (gets doubled) on refresh
### Describe your problem!

On refresh of an offline store cache the PN's server version will often be added to the local version, even if there are no (visible) difference at all.

### How to reproduce?

1. store any cache offline and add a PN (which should not only contain letters, but blanks and CR/LFs as well)
2. upload PN to server (it does not matter if you first save and then upload it or if you use the upload button already shown in the edit window) 
3. refresh the offline stored cache listing

In c:geo the PN will (often) be doubled now.

PN should only be there once (if there are no differences, of course).

### Reproducible

Yes

### c:geo Version

2025.11.23-NB-7cd5348

### System information

```text

```

### Additional Information

Sometimes it's pretty hard to reproduce, but I at least managed to nail two scenarios in which the download procedure most definitely changes the content which is stored on the server: 

First, if you enter this text as PN:

```
blubb   blubb

blubb

```

... after upload and refresh, you will end up with this: 

```
blubb   blubb

blubb
---
blubb blubb

blubb
```

Please note that in the appended server version (i. e  the second one) there's just one blank between the first two blubb's, plus that the last CR/LF is missing.

Now open the cache in browser and check the text in the online PN: You will see that there are still three blanks - thus while downloading two blanks get lost somehow.

Second: Once a PN has been doubled, it appears to be impossible to get rid of the version doubling. You can delete the appended one and upload your version as often as you like, after a refresh the server's version will always be appended.

The only possibility to get rid of the doubling is to delete the local (first) version (i. e. from the beginning up to and including the '---' line), thus keep the downloaded version, and then save and uploadn it - now a refresh will result in just one version of the PN.

Following this I would guess that at any point of store locally/upload/store on server/download/compare versions a different trim strategy is used, so that either blanks and/or CR/LFs are treated resp. trimmed differently.

eddiemuc:
fix #17598: preserve personal note whitespaces on refreshing/downloading cache data

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCLogin.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCLogin.java
@@ -571,9 +571,18 @@ public class GCLogin extends AbstractLogin {
     @Nullable
     @WorkerThread
     String getRequestLogged(@NonNull final String uri, @Nullable final Parameters params) {
+        return getRequestLogged(uri, params, null);
+    }
+
+    /**
+     * GET HTTP request. Do the request a second time if the user is not logged in
+     */
+    @Nullable
+    @WorkerThread
+    String getRequestLogged(@NonNull final String uri, @Nullable final Parameters params, @Nullable final Boolean removeWhitespace) {
         try {
             final Response response = Network.getRequest(uri, params).blockingGet();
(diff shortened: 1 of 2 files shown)


eddiemuc:
Retest this please

PR #17601, opened 2025-11-23
 Testing of colored action bar

murggel:
Dev-branch for testing purpose of colored action bar

* Color can be configured via saturation and lightness of the HSL-value for a color.
* Height of colored progress bar: 0 -> complete actionbar
* Option to not color the status-bar (not really recommended for light mode)

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/SeekbarUI.java
+++ b/main/src/main/java/cgeo/geocaching/ui/SeekbarUI.java
@@ -136,6 +136,9 @@ public class SeekbarUI extends LinearLayout {
         valueView.setOnClickListener(v2 -> {
             final String currentValue = valueToShownValue(progressToValue(seekBar.getProgress()));
             int inputType = InputType.TYPE_CLASS_NUMBER;
+            if (getMinValue() < 0) {
+                inputType |= InputType.TYPE_NUMBER_FLAG_SIGNED;
+            }
             if (getHasDecimals()) {
                 inputType |= InputType.TYPE_NUMBER_FLAG_DECIMAL;
             }


moving-bits:
It's more realistic to view the different settings on an actual device or at least at a larger size, but here are some screenshots for comparison anyway:

### Default values
saturation: -2, brightness: -5
Saturation-2

### Saturation -20
saturation: -20, brightness: -5
Saturation-20

### Brightness -20
saturation: -2, brightness: -20
brightness-20

(I've also tried values of -30 for saturation and/or brightness, but that's too much, IMHO)

### Opinions
(feel free to add your own, not limited to the screenshots above)
- @moving-bits: My favorite values for dark mode are
  - color action bar
  - set saturation to -20, keep brightness at default (-5)
- @murggel : My current setting for dark mode is
  - don't color action bar (but I would be fine with the other option as well)
  - keep saturation (-2), set brightness to -15

moving-bits:
any other opinions?

murggel:
> any other opinions?

I know @anders71 likes the variant with the progress bar

How about @MagpieFourtyTwo

MagpieFourtyTwo:
In dark theme (the only one I use): If at all, then only Progress bar colored, but not the entire header region - that would simply be too much for a dark theme. If one has chosen a dark theme he will most probably not be bothered with colors.

moving-bits:
I asked a user on support, that uses nightly versions in dark mode, about their opinion on colored action bar:
- it's ok the way it is currently in nightly versions
- colored action bar is preferred over colored line
- user would like to have some expert settings to configure saturation/brightness
- but the app's functionality itself is much more important than this coloring

(ticket 130129)

MagpieFourtyTwo:
For me, TBH, I like it all black - which is best for the eyes, esp. in the dark.

The only colored thingy I would like to see a little more “prominent” is the “update indicator”.
If it would be possible to make it stand out more clearly (e.g., by doubling its width), that would be great.
But the rest is fine as it is.

murggel:
> For me, TBH, I like it all black - which is best for the eyes, esp. in the dark.
> 

The colored version would be an option for the dark mode, so for you nothing changes.

murggel:
See new PR  #17754 for the implementation

PR #17616, opened 2025-12-13
Limit gc trackable visits to 100 per log

moving-bits:
## Description
Skips sending a cache log when number of GC trackable set to "visit" exceeds 100 (see https://github.com/orgs/cgeo/discussions/105)

Can someone from the team please cross-check whether this is the right way to implement this?


eddiemuc (approved):
Lgtm

moving-bits:
thx for cross-checking

PR #17630, opened 2025-12-17
fix #17609: Wherigo, trigger autosave from c:geo side on save-worthy actions

### LINKED ISSUE #17609: Wherigo Player Auto-Safe every x minutes
### Tell us your idea!

I was in the field recently playing a Wherigo. Unfortunately the system ended c:geo whilst I walked from one Zone to tne next, and also unfortunately the cartridge in question didn't do any autosafe. So my game was lost and I had to start over. 

I suggest introducing an "autosave" feature on c:geo side which works something like this:
* whilst a game is running, it checks every couple of minutes (eg 10 minutes) if something safe-worthy has occured since the last save (auto or manual)
* If yes, c:geo triggers an auto-safe by itself

Safe-worthy events include:
* a new zone, task, thing, ... became visible
* a dialog or input dialog was displayed

### Is there an alternative?

Leaving everything as is means that the user has to rely either in the cartridge owner to place auto-save-points or on him/herself to regularly save the game manually.

### Additional Information

_No response_

eddiemuc:
fix #17609: Wherigo, trigger autosave from c:geo side on save-worthy actions

Save-worthy is:
* a change of visibility for a zone, task, inventory or item
* a finished push dialog or input dialog

Auto-safe files are only created if at least 30 seconds no safe was done and (in case of push/input dialog) action was at least 10 seconds ago. Both restrictions should prevent double-safes in case the cartridge saves automatically or the player saves manually.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/GeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/GeoItemLayer.java
@@ -7,7 +7,6 @@ import cgeo.geocaching.models.geoitem.GeoItem;
 import cgeo.geocaching.models.geoitem.GeoPrimitive;
 import cgeo.geocaching.models.geoitem.ToScreenProjector;
 import cgeo.geocaching.utils.AsynchronousMapWrapper;
-import cgeo.geocaching.utils.CommonUtils;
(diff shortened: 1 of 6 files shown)


eddiemuc:
Failing tests seem unrelated

JakeDot (commented):
LGTM

PR #17641, opened 2025-12-20
Add NoMap tileprovider (fix #17436)

### LINKED ISSUE #17436: Add "empty" map
Quite the opposite of what we have right now: an entry in the map selection, that renders no map at all (just the base map, if available, but that can be disabled already). Sometimes this makes it easier to get an overview of a list with scattered caches or if you want to focus only on the caches without being distracted by a map.

This is related to #17322 because right now the map bounding box hides the base map. But in addition, not rendering a map makes the map faster to use, especially at low zoom levels. And all other points mentioned earlier stay valid of course.

moving-bits:
## Description
Adds a "no map" entry to map selection, which shows a blank background instead of a detailed map, for easier orientation in crowded areas.

image

### DIFF
--- a/main/src/main/res/values/strings.xml
+++ b/main/src/main/res/values/strings.xml
@@ -1416,7 +1416,7 @@
     <string name="map_source_google_terrain">Google: Terrain</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="map_source_osm_offline_combined">Combined (Offline)</string>
-    <string name="map_source_nomap">(no map)</string>
+    <string name="map_source_nomap">No map</string>
     <string name="map_source_attribution_dialog_title">Map Attributions</string>
     <string name="init_sendToCgeo">Send to c:geo</string>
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

main/src/main/res/values/strings.xml
```diff
@@ -1416,6 +1416,7 @@
     <string name="map_source_google_terrain">Google: Terrain</string>
     <string name="map_source_osm_offline">Offline</string>
     <string name="map_source_osm_offline_combined">Combined (Offline)</string>
+    <string name="map_source_nomap">(no map)</string>
```
SammysHP: I suggest to simply call it "No map" or what was the intention to put in in parentheses?

moving-bits: > what was the intention to put in in parentheses?

Just to visualize that it's no regular map, but I have no strong preference here. And it's separated by a divider anyway.

PR #17656, opened 2025-12-22
Java 17, minSDK 26

JakeDot:
Update subprojects, make Java 17 default

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -27,8 +27,8 @@ android {
         // enable Java 8 APIs like Streams on older Android versions
         coreLibraryDesugaringEnabled true
         // use the diamond operator, lambdas and some other Java 8 language features in Android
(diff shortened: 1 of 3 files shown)


moving-bits:
Do you have some reliable source that shows which API level is minimum requirement for full Java 17 support?

API level 26 (Android 8) got released in 2017, but Java 17 seems to have been released only in 2021, so it's very likely that Android 8 either does not support any Java 17 features at all, or at least leaves out some of them. Except if there is some desugaring library that covers the gap.

Same is true for Java 11 (the LTS version before 17), that got released in 2018, so still after Android 8 release.

JakeDot:
I don't. I just remember a Android SDK page saying those were compatible with each other, must have been wrong then.

I guess this PR is moot then.

Issue #17563, opened 2025-10-25
Issues inside Quick Settings menu


wolverine007:
### Describe your problem!

When I set my language to English inside c:geo the Quick Settings menu is still stating Schnelleinstellungen in the header (even after killing c:geo inside the task manager - in all other areas the language switch takes place on the fly):
![Image](https://github.com/user-attachments/assets/61d3eac5-6052-499c-9f8b-96e9a3def703)

![Image](https://github.com/user-attachments/assets/03db3fa3-0631-49f1-b9b9-e3127f5ffa7e)

![Image](https://github.com/user-attachments/assets/0d0b9995-b74e-4b39-ac1f-44871979571f)

![Image](https://github.com/user-attachments/assets/c153bc62-59d0-43ee-8122-3cb5f7f86d41)

### How to reproduce?

Set language to English, Norwegian, etc pp - Schnelleinstellungen stays in the header.

### Actual result after these steps?

Language changes everywhere apart from the quick settings menu header.

### Expected result after these steps?

Adjust the header of the quick settings pop up to the defined language.

### Reproducible

Yes

### c:geo Version

2025-10-01

### System information

```text

```

moving-bits:
Just a short feedback without having looked into the code yet:
- The first issue clearly is a bug. Possible cause is the strings for window title might be using an application context (manual language switching in c:geo probably not being applied to), whereas the other strings might be using the current view's context (including manual translation switch)
- The second issue is something controlled by Android and will behave differently with different screen sizes, font sizes etc. - nothing we can influence AFAIK

murggel:
For the first issue see PR #17662

eddiemuc:
> * The first issue clearly is a bug. Possible cause is the strings for window title might be using an application context (manual language switching in c:geo probably not being applied to), whereas the other strings might be using the current view's context (including manual translation switch)

While investigating this for `LocalizationUtils`, I find that the application context language is indeed switched on user language change:

https://github.com/cgeo/cgeo/blob/e792608a9b6dc6df6f67fc964a1944485533fe02/main/src/main/java/cgeo/geocaching/CgeoApplication.java#L191-L199

So this is probably not the source of the problem

murggel:
> While investigating this for `LocalizationUtils`, I find that the application context language is indeed switched on user language change:
> 
[...]

PR #17662, opened 2025-12-26
Use Resource for title of "QuickSettings"

murggel:
## Description

Use getString instead of pure string-ID to make sure, the activity-context is used

## Related issues

rel #17563

## Additional context

Since I can't reproduce it but had it sporadic as well, I hope, this can fix it.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/maps/MapSettingsUtils.java
+++ b/main/src/main/java/cgeo/geocaching/maps/MapSettingsUtils.java
@@ -151,7 +151,7 @@ public class MapSettingsUtils {
             dialogView.mapSettingsProximitynotificationMastertoggle.setChecked(Settings.isProximityNotificationMasterToggleOn());
         }
 
-        final Dialog dialog = Dialogs.bottomSheetDialogWithActionbar(activity, dialogView.getRoot(), activity.getString(R.string.quick_settings));
+        final Dialog dialog = Dialogs.bottomSheetDialogWithActionbar(activity, dialogView.getRoot(), R.string.quick_settings);
         dialog.setOnDismissListener(d -> {
             boolean filterChanged = false;
             final boolean circleChanged = circlesCb.valueChanged;
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/Dialogs.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/Dialogs.java
@@ -258,7 +258,7 @@ public final class Dialogs {
      * create a bottom sheet dialog with action bar
      */
     public static BottomSheetDialog bottomSheetDialogWithActionbar(final Context context, final View contentView, final @StringRes int titleResId) {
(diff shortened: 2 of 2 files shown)


main/src/main/java/cgeo/geocaching/maps/MapSettingsUtils.java
```diff
@@ -151,7 +151,7 @@ public static void showSettingsPopup(final Activity activity, @Nullable final In
             dialogView.mapSettingsProximitynotificationMastertoggle.setChecked(Settings.isProximityNotificationMasterToggleOn());
         }
 
-        final Dialog dialog = Dialogs.bottomSheetDialogWithActionbar(activity, dialogView.getRoot(), R.string.quick_settings);
```
moving-bits, line 154: We could fix `Dialogs.bottomSheetDialogWithActionbar` to use `context.getString()` to fix potential other places where this error occurs.

murggel: I will check the calls and change it accordingly

murggel: Changed `Dialogs.bottomSheetDialogWithActionbar` to use `context.getString()` instead of `LocalizationUtils.getString`.

There are more places where `context.getString()` instead of `LocalizationUtils.getString` could be used - shall I fix them as well or wait first, if this fixes really the behavior

eddiemuc: Wouldn't it make more sense to change `LocalizationUtils` such that it respects manual language setting in c:geo?

If all else fails, `CgeoApplication` provides a method to get the currently active activity, if any:

    `public Activity getCurrentForegroundActivity()`

eddiemuc: Alternatively here is how you could get a localized context from a given context (say, the application context inside LocalizatioUtils :-))

```
// Create a new Configuration object with the specified locale
Configuration configuration = new Configuration(resources.getConfiguration());
configuration.setLocale(locale);

// Create a new Context with the updated configuration
Context localizedContext = context.createConfigurationContext(configuration);
```

murggel: To be honest, I didn't want to change to much before it's not verified, that this solution solves the problem. 

But then I would probably try the variant with the new localized context....

murggel:
@eddiemuc I guess this PR will be superseeded by PR #17703, correct?

Or should I remove usage LocalizationUtils for those places, where a context is already given?

eddiemuc:
> @eddiemuc I guess this PR will be superseeded by PR #17703, correct?
> 
> Or should I remove usage LocalizationUtils for those places, where a context is already given?

I would prefer to use LocalizationUtils everywhere, but this js up for discussion of course

murggel:
see PR #17761

PR #17665, opened 2025-12-27
Sharing githelper script

eddiemuc:
Sharing githelper script

Copilot can only work directly in the main remote repository. Also it creates many commits during its work. This is different from my usual workflow via my own fork and always amending the same commit.

To deal with this I created (via copilot :-)) two bash helper functions for myself:
* git_checkout_local: to check out a remote branch locally 
* git_squash: to squash commits in current branch into one
While both can be achieved with git alone too, I find it comfortable to have some checks and failsafes included in the script. 

This PR contains the script. Is there any demand of sharing this via cgeo repo?

### DIFF
--- a/main/project/scripts/githelpers.sh
+++ b/main/project/scripts/githelpers.sh
@@ -1,0 +1,260 @@
+#!/usr/bin/env bash
+# Git workflow helper functions for local repositories.
+# Source this file in your shell (e.g., `. ./githelpers.sh`).
+
+# Fail with a message to stderr
+_git_fail() { echo "ERROR: $*" >&2; return 1; }
+
+# Check if current directory is inside a git work tree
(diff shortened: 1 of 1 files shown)


bekuno:
I propose to add the `scripts` directory inside `/main/project/`.
(Additionally, we can move the folder `/main/project/` direct into the root folder.)

eddiemuc:
> I propose to add the `scripts` directory inside `/main/project/`. (Additionally, we can move the folder `/main/project/` direct into the root folder.)

Ok, I moved scripts folder to /main/project. I did not move project folder to root because I don't know what side effects this might have (theres so much stuff in that folder)

eddiemuc:
default and codacy is green, lets merge here

PR #17669, opened 2025-12-27
copilot: modify instructions to not separate static and non-static imports

eddiemuc:
copilot: modify instructions to not separate static and non-static imports


eddiemuc:
fast-merge

PR #17670, opened 2025-12-27
rel #17562: Improve input dialog for range slider

murggel:
## Description

Shorten title and add the range-value to message inside the dialog

## Related issues

rel #17562

## Description

Shorten title and add the range-value to message inside the dialog

## Related issues

rel #17562

## Additional context

| old dialog | new dialog |
| --- | --- |
|   |  |

PR #17672, opened 2025-12-27
fix #17562: Add label to slider

### LINKED ISSUE #17562: Tweak Favorites Filter
### Tell us your idea!

At the moment the favorites filter function is set with a slider, which starts with 0 and goes up to >1000.
As you move the slider you could see the current value, if you release your finger it might happen that the slider jumps to another value, which is not notable by the user, cause the values are not shown. 
For sure you could click on each end-point to enter the numbers in a pop-up, but even than you could not see the values currently defined:
Image

Adding two fields stating the current value above oder below the slider might help a lot.

Add two fields stating the currently set values for lower and upper value.

### Is there an alternative?

Click the slider end to see the value inside a pop up:

Image

### Additional Information

_No response_

murggel:
## Description

Adds label with the current selected values next to the slider. If value is equal min / max, the label is not shown

## Related issues

fix #17562


murggel:
This PR will have a merge conflict with PR #17670, but I wanted to have it separate

murggel:
> This PR will have a merge conflict with PR #17670, but I wanted to have it separate

Merge-conflicts are resolved now

eddiemuc:
lgtm

PR #17685, opened 2025-12-29
Copilot: make gradle commands run offline only

eddiemuc:
Copilot: make gradle commands run offline only

(this will be fast-merged)


bekuno:
Why do you want to have all gradle commands in offline mode?
Sometimes you need download new dependencies.
See eg. https://www.baeldung.com/gradle-offline-mode

eddiemuc:
Deleted previous comment in favor of new issue: https://github.com/cgeo/cgeo/issues/17688

PR #17691, opened 2025-12-29
Lower max distance for history track (fix #17690)

### LINKED ISSUE #17690: Geofence history track
### Tell us your idea!

Would it be possible to geofence the history track so that it doesn't log my position when I'm in certain locations, at home for example ?

Right now, since I use c:geo both on the road and at home, all history track lines converge to my house and it makes for a messy result - see screenshot.

The ability to prevent recording of history within a radius of x meters would be useful.

Image

Geofence (in meters) around a certain location to prevent the recording of history track points.

### Is there an alternative?

None I can think of

### Additional Information

_No response_

moving-bits:
## Description
- Lowers the maximum distance up to which two points of a history track get drawn connected from 10,000 m to 500 m (default)
- Adds a config option to set this value in the range 10 .. 20,000 m
- Puts the three history track-related config options into their own category

image

This is for testing purposes currently, default value of 500 m needs to be cross-checked "in real life". Also imperial units need to be taken into consideration in the UI.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -1143,6 +1143,10 @@ public class Settings {
         return getInt(R.string.pref_maptrail_length, getKeyInt(R.integer.historytrack_length_default));
     }
 
+    public static int getMaximumMapTrailDistance() {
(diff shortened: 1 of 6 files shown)


moving-bits:
failing tests are unrelated

moving-bits:
Let's try it in the nightlies

PR #17713, opened 2025-12-31
Add delete all pending downloads button (fix #17174)

### LINKED ISSUE #17174: Failed Downloads stay indefinitely
### Describe your problem!

When downloading many files like for instance routing or hillshading data, it can easily happen that c:geo crashes at some point or the downloads fail in some other way. If this happens, the user is forced to delete all entries in the Pending downloads activity by hand.

### How to reproduce?

Open map, zoom out to maximum level allowed by the current map, select "Manage offline data", "Check for missing hillshading data" for example, force close c:geo, open "Pending downloads" again.

Endless list with thousands of entries, which takes several seconds to load with the possibility of c:geo crashing, without any functionality but to delete the downloads from the cg_extension table individually.

There should be a button to remove all failed downloads at once and clean up any possible partial results.

### Reproducible

Yes

### c:geo Version

current nightly

### System information

```text

```

### Additional Information

_No response_

JakeDot:
Since it can happen that very many downloads stall/fail and it is a pain to remove them all manually, I added a button for removing all of them at once.

Fixes #17174

### DIFF
--- a/main/src/main/res/layout/pending_downloads_activity.xml
+++ b/main/src/main/res/layout/pending_downloads_activity.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/activity_content"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical">
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/delete_all_button"
+        style="@style/button_icon_with_text"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="8dp"
+        android:text="@string/downloader_delete_all_pending"
+        app:icon="@drawable/ic_menu_delete"
+        app:iconGravity="start" />
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/list"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1"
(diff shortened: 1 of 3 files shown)


main/src/main/res/layout/pending_downloads_activity.xml
```diff
@@ -0,0 +1,31 @@
...
+        android:id="@+id/delete_all_button"
+        style="@style/button_icon_with_text"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="8dp"
+        android:text="@string/downloader_delete_all_pending"
+        app:icon="@drawable/ic_menu_delete"
+        app:iconGravity="start" />
+
+    <androidx.recyclerview.widget.RecyclerView
```
moving-bits: you could use `<include layout="@layout/generic_recyclerview" />' to include the `RecyclerView` part instead of having to duplicate the layout code

JakeDot: changed in commit [1c14c71](https://github.com/cgeo/cgeo/pull/17713/commits/1c14c719ff77c74b15b165e4b92b9f714a765ef0)

JakeDot: @moving-bits

I reverted that change because it results in a build error:

src/main/res/layout/pending_downloads_activity.xml:19: Error: Duplicate id @+id/activity_content, defined or included multiple times in layout/pending_downloads_activity.xml: [layout/pending_downloads_activity.xml defines @+id/activity_content, layout/pending_downloads_activity.xml => layout/generic_recyclerview.xml defines @+id/activity_content] [DuplicateIncludedIds]

moving-bits: Ah, you are right, `generic_recyclerview` defines a full layout (including `activity_content`), not just the view itself. - Sorry for the confusion.

JakeDot:
retest this please

Issue #17714, opened 2026-01-01
c:geo sending invalid bounding box to GC web request


moving-bits:
We got an urgent request from GC to fix our calls to their web API:

> Noticed that cgeo is causing a lot of 400 errors in the web search api. They are making search requests with a bad latitude, which causes the service to return a 400 validation error.
> ...
> here's what cgeo needs to do instead of what they are currently doing, in order to fix this:
> ```Bounding box lat/long pairs separated by a ',' where the first pair is upper left and the second lower right. Example: box=47.7341,-122.2365,47.4956,-122.4361```

An example for such a bad request would be 
`https://www.geocaching.com/api/proxy/web/search/v2?box=108.643292,-172.582875,-61.752832,-53.490463&rad=1`
(I guess this is an internal sample call, as the example Bryan sent to us used a different domain and c:geo is using a different value for the `rad` parameter, but anyway, we should look into this.)

This would be `connectors/gc/GCWebAPI.java`, `execute()` method.

I propose to implement a quick-fix, which ignores such bad calls, logging them + returning an empty `SearchResult`, and then leave this issue open to analyze our code where such bad values may come from.

As I'm planning a new bugfix release anyway to be published before the weekend, I would like to include the quick fix.

moving-bits:
PR #17715 adds a safeguard as quick-fix

moving-bits:
fixed with #17715

bekuno:
Reopen, as in https://github.com/cgeo/cgeo/pull/17715 was only a safeguard added.

moving-bits:
Removing "Prio: High" label, as the main problem (server problems due to invalid coordinates) is fixed.

Remaining problem is about why c:geo calculates such invalid coordinates at all.

PR #17715, opened 2026-01-01
Add coordinate range safeguard to web API search (rel. to #17714)

moving-bits:


eddiemuc:
I propose as quick fix to reduce overflowing Viewports to their max/min-allowed value instead of aborting the whole request.

Logging makes sense though, so we have a chance to find the source of this.

moving-bits:
> I propose as quick fix to reduce overflowing Viewports to their max/min-allowed value instead of aborting the whole request.

To my understanding that would not help to fix the coordinates for the user (in the sense that we afterwards have the coordinates the user wanted in the first place). We would then retrieve data for different, but still wrong coordinates. And as this fix is about reducing unnecessary calls to the server, I would rather go with no call at all.

If we know the reason for the miscalculation, we might (or might not) be able to do something about it, eg. performing a normalization (aka "modulo calculation") or something else.

eddiemuc:
Fine for me to go forward this way

moving-bits:
Let's get this into nightly version then. I can prepare a new bugfix release tomorrow.

main/src/main/java/cgeo/geocaching/location/Viewport.java
```diff
@@ -428,6 +428,17 @@ public boolean isJustADot() {
         return bottomLeft.equals(topRight);
     }
 
+    public boolean hasInvalidRange() {
+        return topRight.getLatitudeE6() > 90000000
+                | topRight.getLatitudeE6() < -90000000
```
JakeDot: Why are you not chaining with short-circuited boolean operator || ?

PR #17717, opened 2026-01-01
fix #17368: use System-Settings as default theme for new installation

### LINKED ISSUE #17368: set theme color in the Configuration wizard to System Default
### Tell us your idea!

Currently the initial theme color in the Configuration Wizard is set to black.
This is strange if you use a light system theme. 
Better set it to System default or present a theme selection dialog.

The initial theme color should be System Default.

### Is there an alternative?

Create a Theme selector in the Configuration Wizard.
But I am against it because it makes the Configuration Wizard unnecessarily complex.

### Additional Information

_No response_

murggel:
## Description

Use System-Settings as default theme for new installation.
Keep old logic for migration scenario

## Related issues

fix #17368

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -309,7 +309,6 @@ public class Settings {
             e.putInt(getKey(R.string.pref_lastmapzoom), prefsV0.getInt(getKey(R.string.pref_lastmapzoom), 14));
             e.putBoolean(getKey(R.string.pref_livelist), prefsV0.getInt(getKey(R.string.pref_livelist), 1) != 0);
             e.putBoolean(getKey(R.string.pref_units_imperial), prefsV0.getInt(getKey(R.string.pref_units_imperial), 1) != 1);
-            e.putBoolean(getKey(R.string.old_pref_skin), prefsV0.getInt(getKey(R.string.old_pref_skin), 0) != 0);
             e.putInt(getKey(R.string.pref_lastusedlist), prefsV0.getInt(getKey(R.string.pref_lastusedlist), StoredList.STANDARD_LIST_ID));
             e.putInt(getKey(R.string.pref_version), prefsV0.getInt(getKey(R.string.pref_version), 0));
             e.putBoolean(getKey(R.string.pref_friendlogswanted), prefsV0.getBoolean(getKey(R.string.pref_friendlogswanted), true));
@@ -530,10 +529,6 @@ public class Settings {
 
     public static String getString(final int prefKeyId, final String defaultValue) {
         return getStringDirect(getKey(prefKeyId), defaultValue);
-    }
-
-    public static boolean hasKey(final int prefKeyId) {
-        return hasKeyDirect(getKey(prefKeyId));
     }
 
     private static boolean hasKeyDirect(final String prefKey) {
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -1408,8 +1412,16 @@ public static void setAppTheme(final DarkModeSetting setting) {
     }
 
     private static DarkModeSetting getAppTheme(final @NonNull Context context) {
```
moving-bits, line 1409: Would it be feasible to separate migration of old settings and setting default values here?

Settings migration should be done in `Settings.migrateSettings()`, and setting a default value normally is not more than simply setting the default parameter to the method used internally for reading a preference key.

murggel: I will have a look at it

murggel: @moving-bits can you please check, if my implementation is correct, I am not familiar with migration implementation

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -515,6 +516,20 @@ private static void migrateSettings() {
                 Log.e("Migrated map mode to UnifiedMap: " + tileprovider);
             }
 
+            if (hasKeyDirect(getKey((R.string.old_pref_skin)))) {
```
eddiemuc: I believe this code will not be executed for current installations because you don't increase the settings version number (currently 11, must be 12).
Settings migration has a versioning system very similar to the one on datastore, but much less used.

Personally I try to avoid settings migration whenever possible. When you introduce a new setting, I find it easiest to just return the "migrated" value from the respective getter method in Settings whenever the setting does not yet exist.

murggel: > I believe this code will not be executed for current installations because you don't increase the settings version number (currently 11, must be 12).
> Settings migration has a versioning system very similar to the one on datastore, but much less used.

The new setting already exits in current version, so no need for migration.

> Personally I try to avoid settings migration whenever possible. When you introduce a new setting, I find it easiest to just return the "migrated" value from the respective getter method in Settings whenever the setting does not yet exist.
> 

That was my first approach, but @moving-bits mentioned to use migration instead.
Personally I like the migration idea to not clutter up with old settings, butvI would be good with both approaches

eddiemuc: > > I believe this code will not be executed for current installations because you don't increase the settings version number (currently 11, must be 12).
> > Settings migration has a versioning system very similar to the one on datastore, but much less used.
> 
> The new setting already exits in current version, so no need for migration.

Still this code won't be executed I think. Because at the very beginning of the migrateSettings() method there's a check on what version the settings have (currently 11). And if the settings have the most current version then the whole migrate-method is not executed

murggel: > > > I believe this code will not be executed for current installations because you don't increase the settings version number (currently 11, must be 12).
> > > Settings migration has a versioning system very similar to the one on datastore, but much less used.
> > 
> > The new setting already exits in current version, so no need for migration.
> 
> Still this code won't be executed I think. Because at the very beginning of the migrateSettings() method there's a check on what version the settings have (currently 11). And if the settings have the most current version then the whole migrate-method is not executed 

But that's fine, it doesn't need to be executed on current version. 

But I have to check, when the "new" setting was introduced and move it to that version

murggel: Oh, OK, if someone never changed the setting, then probably it needs to be executed

murggel: Oh, OK, if someone never changed the setting, then probably it needs to be executed

eddiemuc: migrateSettings is specifically designed to migrate existing AND new installations.

murggel: Can you please check the first commit only, that is the "direct" implementation. 
And I will check with @moving-bits whether to use migration or not

eddiemuc: > Can you please check the first commit only, that is the "direct" implementation. 
> And I will check with @moving-bits whether to use migration or not

I would prefer the original solution (first commit only). It seems less complex overall.

eddiemuc: > Can you please check the first commit only, that is the "direct" implementation. 
> And I will check with @moving-bits whether to use migration or not

I would prefer the original solution (first commit only). It seems less complex overall.

murggel: @moving-bits will you be also ok with the "old" implementation? Then I would revert the last commit

moving-bits: Yes, I would be fine with it.

Alternatively you could leave out migration of old light/dark preference completely - this setting is gone since four years or so, and it's a non-critical setting (= easy to fix for the user).

murggel:
Retest this please

murggel:
I am not sure, if the failing tests are related to the change - will check asap

moving-bits:
At least the "WaypointActivityTest.checkPermissions" error I got as well in my recent PR, and "CgeoApplicationTest.testSearchByGeocodeBasis" is one of the more "flaky" test.

murggel:
I reverted the last commit, so ready for review

PR #17718, opened 2026-01-01
Fix #17370: skip input dialog after converting to simple coords

### LINKED ISSUE #17370: waypoint calculator: Missing buttons after conversion to simple coordinates
1. create waypoint
2. add formula
3. save waypoint
4. re-open waypoint edit dialog: opens in formula mode
5. "convert to simple mode"
-> dialog is missing the 4 buttons

Image

after re-opening they're back:

Image

murggel:
## Description

This PR dismiss the input dialog after converting calculated waypoint into simple coordinates and returns direct to the edit-waypoint-activity.

## Related issues

Fix #17370

## Workflow
| Create waypoint with calculated coordinates | Convert to simple coordinates|Returns automatically (without any confirmation) to edit-waypoint (User can confirm or cancel the coordinates / waypoint ) | 
| --- | --- | --- |
|image |image |image |

### to discuss
* should there be a confirmation dialog? If no, the formula would be lost, if the waypoint with formula wasn't be saved before...

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
+++ b/main/src/main/java/cgeo/geocaching/ui/dialog/CoordinatesCalculateGlobalDialog.java
@@ -200,7 +200,19 @@ public class CoordinatesCalculateGlobalDialog extends DialogFragment {
         binding.convertToPlain.setOnClickListener(v -> {
             // When the callback is hit it will clear the calculator state associated with the waypoint
             final CoordinateInputData cid = createFromDialog();
-            callback.onDialogClosed(cid.getGeopoint());
+            final Geopoint plainGp = cid.getGeopoint();
+            if (plainGp == null) {
+                SimpleDialog.of(this.getActivity()).setTitle(R.string.calccoord_convert_to_plain)
+                        .setMessage(TextParam.id(R.string.calccoord_convert_to_plain_error))
+                        .setPositiveButton(TextParam.id(R.string.button_continue))
+                        .setNegativeButton(TextParam.id(R.string.cancel))
+                        .confirm(() -> {
+                            callback.onDialogClosed(null);
+                            dismiss();
+                        });
+                return;
+            }
+            callback.onDialogClosed(plainGp);
             dismiss();
         });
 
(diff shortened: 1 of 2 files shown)


bekuno:
Retest this please

moving-bits:
Not sure if this is intended behavior:
- Open a cache (GC96684 in my case)
- Add a waypoint, select "calculated coordinates" and edit the north coordinate to "N50 40.abc", store waypoint
- Reopen waypoint, enter calculated coordinate edit mode, and change to simple coordinate, store waypoint
- Reopen waypoint: coordinates are empty now
- Tap on coordinates: your current location will be preset

This is different, if all variables of your calculated coordinates are available and lead to valid coordinates, then the calculated coordinates are used for this waypoint as new simple coordinates.

Thus I propose to add a warning in case there are unresolved variables.

murggel:
> Not sure if this is intended behavior:
> - Open a cache (GC96684 in my case)
> - Add a waypoint, select "calculated coordinates" and edit the north coordinate to "N50 40.abc", store waypoint
> - Reopen waypoint, enter calculated coordinate edit mode, and change to simple coordinate, store waypoint
> - Reopen waypoint: coordinates are empty now
> - Tap on coordinates: your current location will be preset
> 
> This is different, if all variables of your calculated coordinates are available and lead to valid coordinates, then the calculated coordinates are used for this waypoint as new simple coordinates.
> 
> Thus I propose to add a warning in case there are unresolved variables.


Edit:
Not intended  - in current nighty the current location is used in that case...
Maybe a warning is fine, that plain coordinates are not possible and a dialog-box whether to continue or if current or cache-coordinates should be used? Or keep it empty to re-enter...

murggel:
Please improve the text: 
Converting empty or invalid coordinates to plain coordinates will lead to empty coordinates.

Probably something like:
The coordinates are empty or invalid. Continuing will result in empty coordinates.


grafik grafik

moving-bits:
Maybe something like "this will remove the formula and reset the waypoint to empty coordinates" (or: "... to current location", if that is the actual result)

murggel:
> Maybe something like "this will remove the formula and reset the waypoint to empty coordinates" (or: "... to current location", if that is the actual result)

It's easier reseting to empty coordinates.

Removing the formula is always the case, so I would skip that...

murggel:
The message now:

"The formula does not evaluate to valid coordinates. 
Converting to plain coordinates will reset the waypoint to empty coordinates.

Cancel / Continue"

murggel:
@moving-bits  I would propose to merge

murggel:
Retest this please

PR #17719, opened 2026-01-01
rel to #17714: only send valid viewport boxes to gc web api

eddiemuc:
rel to #17714: only send valid viewport boxes to gc web api

Alternative approach to PR #17715


moving-bits:
Superseded by #17714 (see also https://github.com/cgeo/cgeo/pull/17715#issuecomment-3704014998)

PR #17753, opened 2026-01-11
develocity Publishing buildScan only on-demand

bekuno:
Publishing every build is the default state.
As we have access to our build logs, this is not necessary. 
You can use `./gradlew --scan` id you want the build scan.


bekuno:
retest this please

PR #17760, opened 2026-01-12
(Temporarily) remove 'Hide ActionBar on tap' in map

moving-bits:
## Description
As discussed in yesterday's dev meeting: Temporarily removing the "hide ActionBar on tap" functionality from map might make migration to a clean `Toolbar` / `MaterialToolbar` and (hopefully) resolving some edge2edge issues on the map easier.


moving-bits:
May be unnecessary, see PR #17775, therefore setting WIP label for now

moving-bits:
As discussed in today's dev meeting: Our approach is to remove `SupportActionBar` completely, hiding/unhiding toolbar should no longer be a problem then, therefore closing here.

PR #17767, opened 2026-01-14
fix #17747: Enhance Specials filter with popup selection for individu…

### LINKED ISSUE #17747: Feature Request: Improve “Special” Cache Types in Filters
### Tell us your idea!

While reviewing the filters page, and before reading existing threads on the topic, I found myself wondering what the “Special” and “Others” filters actually represent. This is not explicit to users, and even as a regular geocacher, I had difficulty interpreting them.

Does “Special” mean special events? How can users know which cache types are included there and which ones are placed under “Others”?

Important related issues in this discussion:

- https://github.com/cgeo/cgeo/issues/16376
- https://github.com/cgeo/cgeo/issues/11923

I am a software tester and QA, and I fully understand that too much information should not be placed on a single page, as this can negatively impact UI clarity and usability. Geocaching has a large number of cache types, and it is not practical to display all of them in a flat menu.

However, as of 2026, CCE, Block Party, and Giga-Events no longer exist, while Mega-Events and the occasional GPS Adventures Maze still do. Many geocachers would appreciate having these cache types filtered independently or, at the very least, handled in a clearer and more intuitive way.

Create an expandable filter group for Events, which would include:

- All event cache types
- GPS Adventures Maze (since it uses Attend logs, even though geocaching.com does not group it as an event in its own filters)

This expandable group could be placed either:

- On the main filters page, or
- On the advanced filters page

Behavior example:

- If the Events filter is disabled → the map shows no events
- If enabled (collapsed) → the map shows all events
- If expanded → the geocacher can selectively enable or disable specific event types

Pros

- Clearer and more intuitive filter structure
- Better discoverability of event-related cache types
- Reduced confusion around “Special” and “Others” categories
- Improved usability without cluttering the main UI

Cons

- Slightly increased complexity in filter logic
- Additional UI elements to maintain

### Is there an alternative?

_No response_

### Additional Information

_No response_

JakeDot:
…al event types

- Add custom click handler to COMMUN_CELEBRATION checkbox that opens multi-select dialog
- Implement partial special event type selection (MEGA_EVENT, GIGA_EVENT, etc.)
- Update TypeGeocacheFilter to support filtering individual special event types
- Use EnumSet for efficient enum collection handling
- Maintain accessibility by setting OnClickListener directly on checkbox
- Add unit tests for partial selection scenarios

The Specials checkbox now opens a dialog allowing users to select individual special event types instead of all-or-nothing. The filter correctly handles partial selections alongside other cache type filters.

Fix #17747

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/core/TypeGeocacheFilter.java
+++ b/main/src/main/java/cgeo/geocaching/filters/core/TypeGeocacheFilter.java
@@ -6,7 +6,17 @@ import cgeo.geocaching.enumerations.CacheType;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.utils.JsonUtils;
 
-import static cgeo.geocaching.enumerations.CacheType.*;
+import static cgeo.geocaching.enumerations.CacheType.BLOCK_PARTY;
+import static cgeo.geocaching.enumerations.CacheType.COMMUN_CELEBRATION;
+import static cgeo.geocaching.enumerations.CacheType.GCHQ;
+import static cgeo.geocaching.enumerations.CacheType.GCHQ_CELEBRATION;
+import static cgeo.geocaching.enumerations.CacheType.GIGA_EVENT;
+import static cgeo.geocaching.enumerations.CacheType.GPS_EXHIBIT;
+import static cgeo.geocaching.enumerations.CacheType.LOCATIONLESS;
+import static cgeo.geocaching.enumerations.CacheType.MEGA_EVENT;
+import static cgeo.geocaching.enumerations.CacheType.PROJECT_APE;
+import static cgeo.geocaching.enumerations.CacheType.UNKNOWN;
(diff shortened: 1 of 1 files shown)


JakeDot:
@bekuno do you know why adb seems to be unavailable in ci?

bekuno:
> @bekuno do you know why adb seems to be unavailable in ci?

Probably a memory recource problem on the Jenkins server.
We need @mucek4 to rstart it.

moving-bits:
I have contacted him...

JakeDot:
@moving-bits @eddiemuc is one of you into this stuff and can review?

eddiemuc:
> @moving-bits @eddiemuc is one of you into this stuff and can review?

The feature suggested in the original issue would have impact towards the GUI. Given that the topic of how to handle "filter by type" GUI-wise was subject to many discussions in the past, this issue is not read for implementation imho. So any PR for it might be in vain. So is review effort.

Codewise I don't understand why this feature should cause so many new codelines.

PR #17775, opened 2026-01-16
Migrate ActionBar to MaterialToolbar

### LINKED ISSUE #17403: ActionBar: replace with Toolbar?

moving-bits:
## Description
Migrates c:geo to using `MaterialToolbar` instead of AppCompat's `SupportActionBar`

This PR is built on top of #17711 as of today, fixing a couple of it's quirks and addressing some more loose ends.

Fixes #17403

## Known issues
(not all of them are necessarily related to this PR)
- [x] Settings: Extra padding at the top after having changed one item which recreates Activity (like setting theme)
- [x] Settings: Non-active radio button indicators are missing in `ListPreference` (eg: theme or language selection)
- [x] Submenus: Titles have wrong color in light mode (Map / CacheListActivity / CacheDetailsActivity)
- [ ] Coloring of Toolbar does not color system bar area
- [x] Viewpager activities have an extra padding between toolbar and content
- [x] Crash on viewing logbook image
- [x] Some extra spacing at the top in landscape mode
- [x] CacheListActivity has extra space at the bottom
- [x] filter bar in map is placed at the top instead of below toolbar
- [x] `MainActivity`: Update notifications are placed behind toolbar

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -119,6 +119,7 @@ import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
+import androidx.core.graphics.Insets;
 import androidx.core.util.Consumer;
 import androidx.loader.app.LoaderManager;
 import androidx.loader.content.Loader;
@@ -440,6 +441,12 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
                 }
             }
         });
+    }
+
+    @Override
+    @NonNull
+    protected Insets calculateInsetsForActivityContent(@NonNull final Insets def) {
+        return calculateInsetsWithToolbarInPortrait(def);
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


moving-bits:
Ready for review, all issues known up to now are fixed (except coloring of system bar, but that could/should be handled separately).

moving-bits:
retest this please

moving-bits:
Failing tests are unrelated (some OC stuff)

moving-bits:
Retest this please

murggel:
@moving-bits  PR #17879 makes the text of status bar readable, coloring of status bar is still open

moving-bits:
> @moving-bits PR #17879 makes the text of status bar readable

Thanks for fixing!

moving-bits:
> coloring of status bar is still open

should be fixed with #17891

PR #17777, opened 2026-01-17
Some more cleanup after raising minAPI to 26

moving-bits:
## Description
After raising min API to 26, we no longer need some mitigations for insets creation, which leads to a smaller signature of `InsetsBuilder` as well.


bekuno:
retest this please

bekuno:
retest this please

PR #17778, opened 2026-01-17
Fix some lint errors

moving-bits:
## Description
- remove unused preference key
- remove unused drawable
- use `RequireAPI` instead of `TargetAPI` + add formal check


bekuno:
retest this please

main/src/main/java/cgeo/geocaching/permission/PermissionContext.java
```diff
@@ -24,7 +23,7 @@ public enum PermissionContext {
     LOCATION(new String[]{Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION}, R.string.permission_location_explanation, R.string.permission_location_explanation_title),
     SEARCH_USER_IN_CONTACTS(new String[]{Manifest.permission.READ_CONTACTS}, R.string.permission_contacts_read_explanation, R.string.permission_contacts_read_explanation_title),
     LEGACY_WRITE_EXTERNAL_STORAGE(new String[]{Manifest.permission.WRITE_EXTERNAL_STORAGE}, R.string.permission_legacy_write_external_storage_explanation, R.string.permission_legacy_write_external_storage_explanation_title),
-    @TargetApi(33) // permission will only be requested on API 33+ devices (see InstallWizardActivity)
+    @androidx.annotation.RequiresApi(33) // permission will only be requested on API 33+ devices (see InstallWizardActivity)
```
bekuno: Please use a import statement

bekuno (changes requested):
There is another `@TargetApi(33)`statement in `PermissionContext`.
Although it is not marked by lint as a problem, I assume we can also switch to `@RequiresApi(33)`, can't we?

main/src/main/java/cgeo/geocaching/permission/PermissionContext.java
moving-bits: done

moving-bits:
> There is another `@TargetApi(33)`statement in `PermissionContext`.

Where? AFAICS notifications permission was the only one.

bekuno:
> > There is another `@TargetApi(33)`statement in `PermissionContext`.
> 
> Where? AFAICS notifications permission was the only one.

https://github.com/cgeo/cgeo/blob/536e4c72ac4101843b100d57003424526a420415/main/src/main/java/cgeo/geocaching/permission/PermissionContext.java#L27

bekuno:
retest this please

bekuno:
Now there are 2 lint errors in the CI run (https://ci.cgeo.org/job/cgeo-CI_PR-build/14181/console):
~~~
src/main/java/cgeo/geocaching/InstallWizardActivity.java:74: Error: Field requires API level 33 (current min is 26): NOTIFICATIONS [NewApi]
    private final PermissionAction<Void> askNotificationsPermissionAction = PermissionAction.register(this, PermissionContext.NOTIFICATIONS, b -> gotoNext());
                                                                                                                              ~~~~~~~~~~~~~
src/main/java/cgeo/geocaching/InstallWizardActivity.java:420: Error: Field requires API level 33 (current min is 26): NOTIFICATIONS [NewApi]
        return PermissionContext.NOTIFICATIONS.hasAllPermissions();
                                 ~~~~~~~~~~~~~

   Explanation for issues of type "NewApi":
   This check scans through all the Android API calls in the application and
   warns about any calls that are not available on all versions targeted by
   this application (according to its minimum SDK attribute in the manifest).

   If you really want to use this API and don't need to support older devices
   just set the minSdkVersion in your build.gradle or AndroidManifest.xml
   files.

   If your code is deliberately accessing newer APIs, and you have ensured
   (e.g. with conditional execution) that this code will only ever be called
   on a supported platform, then you can annotate your class or method with
   the @TargetApi annotation specifying the local minimum SDK to apply, such
   as @TargetApi(11), such that this check considers 11 rather than your
   manifest file's minimum SDK as the required API level.

   If you are deliberately setting android: attributes in style definitions,
   make sure you place this in a values-vNN folder in order to avoid running
   into runtime conflicts on certain devices where manufacturers have added
   custom attributes whose ids conflict with the new ones on later platforms.

   Similarly, you can use tools:targetApi="11" in an XML file to indicate that
   the element will only be inflated in an adequate context.
~~~

main/src/main/java/cgeo/geocaching/InstallWizardActivity.java
```diff
@@ -39,7 +40,7 @@ public class InstallWizardActivity extends AbstractActivity {
     private static final String BUNDLE_CSAH = "csah";
     private static final String BUNDLE_BACKUPUTILS = "backuputils";
 
-    private static final boolean DO_LEGACY_WRITE_STORAGE = android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.Q; // < SDK29
+    private static final boolean DO_LEGACY_WRITE_STORAGE = Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.Q; // < SDK29
```
bekuno: Please also shorten `android.os.Build.` to `Build.` on the secend place.

PR #17780, opened 2026-01-18
Move background maps path to public folder with fallback

JakeDot:
## Description

Makes background maps loadable from public maps directory additional to old folder in app internal directory.

See jakedot/cgeo#152


JakeDot:
This approach doesn't work because the mbtiles library only accepts file based folders and doesn't work with SAF.

moving-bits:
> This approach doesn't work because the mbtiles library only accepts file based folders and doesn't work with SAF.

That's because mbtiles files are basically SQLite database files, and SQLite database API does not support SAF, it only works with `File` access. That's why I had to use a public app-specific folder which is accessible using `File` API for reading background maps.

An alternative to using that folder would have been to read the mbtiles file from an arbitrary location using SAF file picker, then copy it to an app-internal folder (which can be accessed using `File` API), but then you would end up with two copies of the file, one of which would not be directly accessible for the user.

PR #17790, opened 2026-01-25
Remove triangular UnifiedMap marker image

moving-bits:
## Description
Removes the orange triangular "UnifiedMap" marker image, which we had used in the past to distinguish screenshots for deprecated maps from screenshots for UnifiedMap. This is no longer needed, as we have removed the old map implementations meanwhile.


moving-bits:
retest this please

moving-bits:
failing tests are unrelated

Issue #17797, opened 2026-01-30
Importing a route GPX is possible, but breaks caches already saved.


JakeDot:
### Describe your problem!

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/eb9f1dc7-453a-4fc6-9686-a63dfa394b10" />

Routes can be exported and contain geocache data.

When such a route is imported instead, as a GPX file, the cache data is overwritten with the GPX data, which doesn't seem to include all the details.

### How to reproduce?

1. Export current route to GPX
2. Import GPX (to any List)
3. Mangled data

### Reproducible

Yes

### c:geo Version

2026.01.20-faf18ef developer build

### System information

```text

```

### Additional Information

@copilot can you find out what data is exported from the current route to GPX, focus on the title field of the geocache, which seems to be populated with the caches GC-Code?

ReggieDoodle:
Sorry to hijack your ticket but how does the integration with copilot work?
Is there a guide? Can anyone use it?
Tks
Dave

JakeDot:
@ReggieDoodle 

If I'm informed correctly team members can use it in the cgeo/cgeo repo, anyone else (if they paid for it I guess) can use it in their forks.

I've been using it via the web interface

<img width="1755" height="343" alt="Image" src="https://github.com/user-attachments/assets/09e7f5ea-05b2-4848-8773-08e32b8505e9" />

and via copilot plugin in Android Studio

<img width="1917" height="996" alt="Image" src="https://github.com/user-attachments/assets/ac0726c9-6d4d-4975-906a-c8a84db28d63" />

eddiemuc:
> Sorry to hijack your ticket but how does the integration with copilot work? Is there a guide? Can anyone use it? Tks Dave

Hi Dave,

For usage of GitHub Copilot there are good guides by GitHub itself eg https://docs.github.com/en/copilot/get-started/quickstart. Basically, you need to have the right license with github and then can start in github gui or in the IDE of your choice (with github copilot integration) right away. For c:geo you usually need to work in your own c:geo fork though.

There's no c:geo-specific guide for copilot (to my knowledge). But I guess c:geo's handling of copilot can be summarized like this:
c:geo is **prepared for copilot** usage, but each **contributor remains responsible for everything created in his/her name** (with or without copilot or other tools).

By **prepared for copilot** I mean:
[...]

PR #17798, opened 2026-01-30
Improve route GPX handling

JakeDot:
**rel to #17797: Fix 
cache name overwrite and improve GPX export** - **rel to 
#17797: Ensure getName() never returns null**

 

j
## Description
<!-- Provide a summary of the cont

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/models/GeocacheTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/models/GeocacheTest.java
@@ -583,6 +583,28 @@ public class GeocacheTest {
     }
 
     @Test
+    public void testMergeNameWhenGeocodeUsedAsName() {
+        // Test for issue #17797: When importing route GPX, cache name should not be overwritten with geocode
+        final Geocache stored = new Geocache();
+        stored.setGeocode("GC12345");
+        stored.setName("My Awesome Cache");
+        stored.setDetailed(true);
+        stored.setCoords(new Geopoint(40.0, 8.0));
+        saveFreshCacheToDB(stored);
+
+        // Simulate a cache imported from a route GPX where the name field contains the geocode
+        final Geocache imported = new Geocache();
+        imported.setGeocode("GC12345");
+        imported.setName("GC12345"); // Name is set to geocode (the problem)
+        imported.setCoords(new Geopoint(40.0, 8.0));
+
+        imported.gatherMissingFrom(stored);
+
+        // The name should be taken from the stored cache, not kept as the geocode
+        assertThat(imported.getName()).as("merged name should not be geocode").isEqualTo("My Awesome Cache");
+    }
+
+    @Test
     public void testMergeLocalUserModifiedCoordsNotServerSideModified() {
         final Geocache stored = new Geocache();
         stored.setGeocode("GC12345");
--- a/main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
+++ b/main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
@@ -3,13 +3,17 @@ package cgeo.geocaching.export;
 import cgeo.geocaching.R;
 import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.location.Geopoint;
+import cgeo.geocaching.models.Geocache;
+import cgeo.geocaching.models.RouteItem;
 import cgeo.geocaching.models.RouteSegment;
+import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.storage.ContentStorage;
 import cgeo.geocaching.storage.PersistableFolder;
 import cgeo.geocaching.utils.AsyncTaskWithProgress;
 import cgeo.geocaching.utils.CalendarUtils;
 import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.ShareUtils;
+import cgeo.geocaching.utils.SynchronizedDateFormat;
 import cgeo.geocaching.utils.UriUtils;
 import cgeo.geocaching.utils.xml.XmlUtils;
 import cgeo.org.kxml2.io.KXmlSerializer;
@@ -17,6 +21,7 @@ import cgeo.org.kxml2.io.KXmlSerializer;
 import android.app.Activity;
 import android.net.Uri;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import java.io.BufferedWriter;
@@ -26,8 +31,11 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Date;
+import java.util.Locale;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.xmlpull.v1.XmlSerializer;
 
 public class IndividualRouteExportTask extends AsyncTaskWithProgress<RouteSegment, Uri> {
@@ -40,6 +48,16 @@ public class IndividualRouteExportTask extends AsyncTaskWithProgress<RouteSegmen
 
     private static final String PREFIX_XSI = "xsi";
     private static final String NS_XSI = "http://www.w3.org/2001/XMLSchema-instance";
+
+    private static final String PREFIX_GROUNDSPEAK = "groundspeak";
+    private static final String NS_GROUNDSPEAK = "http://www.groundspeak.com/cache/1/0/1";
+    private static final String GROUNDSPEAK_SCHEMA = NS_GROUNDSPEAK + "/cache.xsd";
+
+    private static final String PREFIX_GSAK = "gsak";
+    private static final String NS_GSAK = "http://www.gsak.net/xmlv1/6";
+    private static final String GSAK_SCHEMA = NS_GSAK + "/gsak.xsd";
+
+    private static final SynchronizedDateFormat dateFormatZ = new SynchronizedDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
 
     IndividualRouteExportTask(final Activity activity, final String filename, final boolean exportAsTrack) {
         super(activity, activity.getString(R.string.export_individualroute_title));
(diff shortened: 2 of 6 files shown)


SammysHP:
Can you share a before/after comparison of the generated GPX file? I think this breaks waypoint references and handling in other apps. Historically, caches and waypoints use the geocode as their name and other apps expect this format.

JakeDot:
retest this please

JakeDot:
> Can you share a before/after comparison of the generated GPX file? I think this breaks waypoint references and handling in other apps.

# Sample Route GPX Export

This is a sample GPX export from a route containing three items:
1. **A geocache** (GC1TEST - "Sample Traditional Cache")
2. **A waypoint** of that cache (Parking waypoint)
3. **A simple coordinate** point

## Full GPX Output

```xml
<?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
<gpx version="1.1" creator="c:geo - http://www.cgeo.org/" 
     xmlns="http://www.topografix.com/GPX/1/1" 
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
     xmlns:groundspeak="http://www.groundspeak.com/cache/1/0/1" 
     xmlns:gsak="http://www.gsak.net/xmlv1/6" 
     xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd http://www.groundspeak.com/cache/1/0/1 http://www.groundspeak.com/cache/1/0/1/cache.xsd http://www.gsak.net/xmlv1/6 http://www.gsak.net/xmlv1/6/gsak.xsd">
  
  <metadata>
    <name>c:geo individual route</name>
    <time>2026-02-01T08:20:00Z</time>
  </metadata>

  <!-- Waypoint 1: Geocache with full metadata -->
  <wpt lat="48.859683" lon="9.1874">
    <name>GC1TEST</name>
    <desc>Sample Traditional Cache</desc>
    <groundspeak:cache id="123456" available="True" archived="False">
      <groundspeak:name>Sample Traditional Cache</groundspeak:name>
      <groundspeak:placed_by>CacheOwner</groundspeak:placed_by>
      <groundspeak:type>Traditional Cache</groundspeak:type>
      <groundspeak:container>regular</groundspeak:container>
      <groundspeak:difficulty>2.5</groundspeak:difficulty>
      <groundspeak:terrain>3</groundspeak:terrain>
    </groundspeak:cache>
    <gsak:wptExtension>
      <gsak:Watch>true</gsak:Watch>
      <gsak:IsPremium>false</gsak:IsPremium>
      <gsak:FavPoints>15</gsak:FavPoints>
      <gsak:GcNote>Great cache with nice view!</gsak:GcNote>
      <gsak:UserFound>2021-01-01T00:00:00Z</gsak:UserFound>
    </gsak:wptExtension>
  </wpt>

  <!-- Waypoint 2: Cache waypoint with parent reference -->
  <wpt lat="48.86" lon="9.1875">
    <name>GC1TEST-01</name>
    <desc>Parking</desc>
    <gsak:wptExtension>
      <gsak:Parent>GC1TEST</gsak:Parent>
    </gsak:wptExtension>
  </wpt>

  <!-- Waypoint 3: Simple coordinate (no extensions) -->
  <wpt lat="48.8605" lon="9.188">
    <name>COORDS! N 48° 51.630 E 009° 11.280</name>
    <!-- No extensions for simple coordinates -->
  </wpt>

  <!-- Route definition -->
  <rte>
    <name>c:geo individual route 2026-02-01T08:20:00Z</name>
    <rtept lat="48.859683" lon="9.1874">
      <name>GC1TEST</name>
    </rtept>
    <rtept lat="48.86" lon="9.1875">
      <name>GC1TEST-01</name>
    </rtept>
    <rtept lat="48.8605" lon="9.188">
      <name>COORDS! N 48° 51.630 E 009° 11.280</name>
    </rtept>
  </rte>

</gpx>
```

## Breakdown by Item Type

### 1. Geocache Waypoint (GC1TEST)

The geocache includes:
- **Basic waypoint data**: coordinates, name, description
- **Groundspeak cache extension** with:
  - Cache ID, availability, and archived status
  - Cache name, owner, type, and container size
  - Difficulty and terrain ratings
- **GSAK extension** with:
  - Watchlist status (`Watch`)
  - Premium membership requirement (`IsPremium`)
  - Favorite points count (`FavPoints`)
  - Personal note (`GcNote`)
  - Found date (`UserFound`)

### 2. Cache Waypoint (Parking)

The cache waypoint includes:
- **Basic waypoint data**: coordinates, name (geocode + prefix), description
- **GSAK extension** with:
  - Parent cache reference (`Parent: GC1TEST`)
  - Would also include `Child_ByGSAK: true` if it was user-defined

### 3. Simple Coordinate Point

The coordinate point includes:
- **Only basic waypoint data**: coordinates and formatted name
- **No extensions** - keeps the GPX clean for simple points

## Key Differences from Original Implementation

Before this enhancement, all waypoints would have looked like waypoint #3 - just coordinates and a name. Now:

- **Geocaches** get full metadata that can be used by GSAK and other tools
- **Cache waypoints** maintain their relationship to parent caches
- **User-defined waypoints** are properly flagged
- **Simple coordinates** remain clean and simple

This allows exported routes to be much more useful when imported into other geocaching tools!

SammysHP:
Shouldn't the waypoint be named `011TEST`? I'm not sure if that has changed in recent years, but in the past the waypoint geocode was constructed by removing the platform prefix and replacing it with the two-character waypoint ID. (This can cause conflicts with other platforms and I don't know it it is still used anywhere when reading GPX files.)

JakeDot:
Maybe the route export doesn't know about that. I will fix that.

SammysHP:
That's why I asked for a before/after comparison. I wasn't sure about the previous implementation. Maybe that was by design or it was something that should be tracked in a separate issue.

JakeDot:
@SammysHP fixed

eddiemuc:
It is not really clear to me what this PR is doing. Closing for nbow. Please reopen if necessary

JakeDot:
@eddiemuc 

This PR fixes the problem that arises when you export a route and then import it with the general GPX import instead of the route import function:

Geocache names disappear and are replaced by the GC-Code in c:geo's DB because the import treats the <name> tag in the GPX as the cache name in the absence of groundpeak: and gsak: namespace tags.

The PR adds those namespaces into the exported GPX, which also improves interoperability with other GPX reading tools.

Please reopen.

main/src/androidTest/java/cgeo/geocaching/models/GeocacheTest.java
```diff
@@ -582,6 +582,28 @@ public void testMergeDownloadedStored() {
         assertThat(download.getHint()).as("merged hint").isEmpty();
     }
 
+    @Test
```
eddiemuc, line 585: Why is this test in intrumented test folder (not in unit test folder)?

main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
```diff
@@ -41,6 +49,16 @@ public class IndividualRouteExportTask extends AsyncTaskWithProgress<RouteSegmen
     private static final String PREFIX_XSI = "xsi";
     private static final String NS_XSI = "http://www.w3.org/2001/XMLSchema-instance";
 
+    private static final String PREFIX_GROUNDSPEAK = "groundspeak";
```
eddiemuc, line 52: In this file there seems to be much duplication of stuff available in GpxSerializer

main/src/main/java/cgeo/geocaching/export/IndividualRouteExportTask.java
```diff
@@ -134,6 +155,113 @@ private void exportPoint(final XmlSerializer gpx, final String tag, final Geopoi
         gpx.endTag(null, tag);
     }
 
+    private void exportWaypoint(final XmlSerializer gpx, @NonNull final RouteItem item) throws IOException {
```
eddiemuc, line 158: seems to be duplicated from GpxSerializer?

main/src/test/java/cgeo/geocaching/export/IndividualRouteExportTaskTest.java
```diff
@@ -0,0 +1,17 @@
+package cgeo.geocaching.export;
+
+import org.junit.Test;
+import static org.assertj.core.api.Java6Assertions.assertThat;
+
+/**
+ * Unit tests for IndividualRouteExportTask to verify namespace declarations.
+ */
+public class IndividualRouteExportTaskTest {
```
eddiemuc, line 9: File is too much for just one one-line-test

main/src/test/java/cgeo/geocaching/models/GeocacheUnitTest.java
```diff
@@ -0,0 +1,84 @@
+package cgeo.geocaching.models;
+
+import cgeo.geocaching.enumerations.CacheType;
+import cgeo.geocaching.location.Geopoint;
+
+import org.junit.Test;
+import static org.assertj.core.api.Java6Assertions.assertThat;
+
+public class GeocacheUnitTest {
```
eddiemuc, line 9: I like the idea of having (non-instrumented) unit-tests for Geocache!

eddiemuc (commented):
It seems your approach duplicates many things from GpxSerializer and is specific only to when you export geocache data from an individual route. This seems not an ideal solution to me. It would be better to unify export via GpxSerializer and IndividualRoute.

Note: I don't think this can be done by copilot

eddiemuc:
PR reopened, but see my comments

SammysHP:
> This PR fixes the problem that arises when you export a route and then import it with the general GPX import instead of the route import function

Wouldn't it make sense to fix the import and make the merging algorithm more robust against such issues?

moving-bits:
Is this PR about exporting routes (and importing it into tools other than c:geo, therefore supplying them with more meta data), or is it about conflicts when importing a c:geo-created GPX for an individual route as a regular GPX file?

If the latter, just the import should be fixed, without adding all the meta data.

If the former, then
- metadata should be added by using the existing methods instead of duplicating them (see @eddiemuc's comments)
- importing such a file as an individual route file should be tested to not break anything on it's main use case. You should tests all kinds of different waypoints and cache types (GC caches, OC caches, Lab adventures, single waypoints, plain coordinates etc.).

Setting "WIP" label for the time being.

moving-bits:
Closing due to lack of feedback

PR #17807, opened 2026-02-01
Create FOSS stub for ElevationChart

JakeDot:
This PR enables FOSS builds by providing a stub implementation of ElevationChart that doesn't depend on MPAndroidChart library.

## Problem
The `ElevationChart` class depends on the MPAndroidChart library (`com.github.mikephil.charting`), which is not available in FOSS (F-Droid) builds. This causes compilation failures when building the FOSS variant.

## Solution
- **Moved to nofoss variant**: Relocated the full `ElevationChart.java` implementation from `main/src/main/` to `main/src/nofoss/`
  - The nofoss variant includes Google Play Services and other non-FOSS dependencies
- **Created FOSS stub**: Added a minimal stub implementation in `main/src/foss/` with the same API surface but no-op methods
  - All methods are empty stubs that do nothing
  - Maintains API compatibility so calling code doesn't need changes

## Impact
- ✅ FOSS builds can now compile successfully
- ✅ No functionality loss for non-FOSS builds (they still get the full implementation)
- ✅ No code changes needed in `UnifiedMapActivity` or other consumers
- ⚠️ FOSS builds will not have elevation chart functionality (expected trade-off)

## Files Changed
- **Moved**: `main/src/main/.../ElevationChart.java` → `main/src/nofoss/.../ElevationChart.java`
- **Created**: `main/src/foss/.../ElevationChart.java` (stub)

## Testing
Tested compilation of FOSS variant - passes Java compilation phase.


moving-bits:
FOSS/light version of c:geo is working perfectly fine with ElevationChart and it's using a FOSS-compatible license, so I don't see a reason why we would want to remove it.

PR #17812, opened 2026-02-02
Implement multi-connector search support

murggel:
## Summary 

Support connector-specific filter tree splitting for multi-connector searches with different owner/login names.

## Problem

When searching across multiple geocaching connectors, each connector may have a different login name for the user. Filter trees containing owner/origin filters need to be evaluated per-connector, as a filter like "owned by me" must resolve to the correct login name for each connector. Previously, the filter tree was applied uniformly, which led to incorrect results when different connectors use different account names.

### `OwnerGeocacheListLoader`

A second constructor is added that accepts a **list of connectors** instead of a single username. In this mode, `getAdditionalFilterParameter()` builds a combined `OR` filter: for each connector in the list, it creates an `AND(OwnerFilter[username], OriginFilter[connector])` branch using the connector's own login name (via `ILogin.getUserName()`). This allows each connector to be queried with its specific owner name. The original single-username constructor remains for backward compatibility.

### `GeocacheFilter`

The `GeocacheFilter` class is extended with logic to split and adapt the filter tree per connector:

- **`getAndChainIfPossibleForConnector(IConnector)`**: New public entry point that returns the AND-chain of base filters relevant for a specific connector.
- **`keepFilterForConnector()`**: Recursively strips or adapts connector-specific filter nodes (especially `OriginGeocacheFilter`) from the filter tree. Redundant or inapplicable branches are pruned.
- **`keepLogicalGeocacheFilterForConnector()`**: Handles logical combinators (AND, OR, NOT) — invalidates branches where origin constraints exclude the given connector.
- **`isFilterValidForConnector()`**: Checks whether an AND filter is compatible with the given connector, considering nested `OriginGeocacheFilter` and `NOT(OriginGeocacheFilter)` nodes.
- **`processLogicalFilterForConnector()`**: Generic helper that rebuilds logical filter nodes with only the children remaining after connector-specific pruning.

## Impact

- No breaking changes to existing filter behavior
- Improves correctness of connector-specific searches (e.g. owner name searches on opencaching vs. geocaching.com)
- The `OR(AND(owner, origin), ...)` structure built by `OwnerGeocacheListLoader` is correctly split per connector by `GeocacheFilter`, so each connector's search backend receives only its relevant filter branch
- Filter tree splitting is purely in-memory; no storage or serialization changes

**_implemented and summary-creation with support of copilot_**

### DIFF
--- a/main/src/main/java/cgeo/geocaching/SearchActivity.java
+++ b/main/src/main/java/cgeo/geocaching/SearchActivity.java
@@ -7,6 +7,7 @@ import cgeo.geocaching.address.AddressListActivity;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.IConnector;
 import cgeo.geocaching.connector.al.ALConnector;
+import cgeo.geocaching.connector.capability.ILogin;
 import cgeo.geocaching.connector.capability.ISearchByGeocode;
 import cgeo.geocaching.connector.gc.GCConnector;
 import cgeo.geocaching.connector.internal.InternalConnector;
@@ -59,6 +60,8 @@ import androidx.appcompat.content.res.AppCompatResources;
 import androidx.cardview.widget.CardView;
 import androidx.core.graphics.Insets;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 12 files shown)


murggel:
For testing I have to check, if I can fake the owners of caches for different connectors (e.g. import via GPX)

main/src/main/java/cgeo/geocaching/SearchActivity.java
```diff
@@ -471,12 +474,38 @@ private void init() {
...
 
         addSearchCard(R.string.search_own_caches, R.drawable.ic_menu_owned)
-                .addOnClickListener(() -> findByOwnerFn(Settings.getUserName()));
+                .addOnClickListener(this::searchOwnCaches);
 
         addSearchCard(R.string.caches_history, R.drawable.ic_menu_recent_history)
                 .addOnClickListener(() -> startActivity(CacheListActivity.getHistoryIntent(this)));
     }
 
+    private void searchOwnCaches() {
```
JakeDot: Can you group this new function with the findByOwnerFn overloads?

JakeDot:
I implemented this a little differently in jakedot#200.

murggel:
> I implemented this a little differently in jakedot#200.

You are right, my implementation is slightly wrong

murggel:
First draft of implementing multi-connector-search-support:
For each connector keep the relevant branch of the search-tree.

I stick to the current approach for the AND-chain.

murggel:
Maybe replaced by PR #18002

Issue #17540, opened 2025-10-12
Calculator: Add function sum("A";"Z")


murggel:
### Tell us your idea!

For various calculations, it is necessary to add up consecutive variables. 

### What solution would you suggest?

A "sum" function would therefore be helpful, where you can specify the start and end values

sum("A";"Z")

eddiemuc:
This can only be implemented if you pass the values as string eg:

`sum("A";"Z")`

Otherwise the variables A and Z would be evaluated to their value, and only their value is then passed to the function.

murggel:
> This can only be implemented if you pass the values as string eg:
> 
> `sum("A";"Z")`
> 
> Otherwise the variables A and Z would be evaluated to their value, and only their value is then passed to the function.

Would be fine for me as well

murggel:
I will try that with Copilot. Did already tried a prompt locally, will enhance that and run it here

murggel:
I closed PR #17771  in favor of a more clean approach -> see PR  #17827

PR #17814, opened 2026-02-03
rel to #17540 and PR #17771: Allow to define functions with own variable dependencies

eddiemuc:
rel to #17540 and PR #17771: Allow to define functions with own variable dependencies

@murggel  this is an extension of the FormulaFunction features to allow defining own, additional variable dependencies baased on constant (literal) parameter values. It contains an example-defined function "TESTSUM" which is a very poor summing-implementation but demonstrates how to use the features. The example can be found in class FormulaFunctions

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
+++ b/main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
@@ -17,6 +17,7 @@ import android.text.style.ForegroundColorSpan;
 import android.util.Pair;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.core.util.Supplier;
 
 import java.math.MathContext;
@@ -111,6 +112,7 @@ public final class Formula {
         private FormulaNode[] children;
 
         public final Set<String> neededVars;
+        public final boolean hasRanges;
 
         FormulaNode(final String id, final FormulaNode[] children,
                     final Func3<ValueList, Function<String, Value>, Integer, Value> function) {
@@ -126,15 +128,16 @@ public final class Formula {
         FormulaNode(final String id, final FormulaNode[] children,
                     final Func3<ValueList, Function<String, Value>, Integer, Value> function,
                     final Func4<ValueList, Function<String, Value>, Integer, Set<Integer>, CharSequence> functionToErrorString,
-                    final Consumer<Set<String>> explicitelyNeeded) {
+                    final Consumer<Set<String>> explicitlyNeeded) {
 
             this.id = id;
             this.children = children == null ? FORMULA_NODE_EMPTY_ARRAY : children;
             this.function = function;
             this.functionToErrorString = functionToErrorString;
-            this.neededVars = Collections.unmodifiableSet(calculateNeededVars(explicitelyNeeded, children));
+            this.neededVars = Collections.unmodifiableSet(calculateNeededVars(explicitlyNeeded, children));
+            this.hasRanges = hasRanges(this);
 
-            if (this.neededVars.isEmpty() && !hasRanges(this)) {
+            if (isConstant()) {
                 //this means that function is constant!
                 this.function = createConstantFunction();
                 final CharSequence csResult = evalToCharSequenceInternal(y -> null, -1).getAsCharSequence();
@@ -145,6 +148,19 @@ public final class Formula {
 
         public String getId() {
             return this.id;
+        }
+
+        public boolean isConstant() {
+            return this.neededVars.isEmpty() && !this.hasRanges;
+        }
+
+        @Nullable
+        public Value getConstantValue() {
+            if (!isConstant()) {
+                return null;
+            }
+            //if this is a constant function, then the following call will work
+            return function.call(null, null, -1);
         }
 
         public Set<String> getNeededVars() {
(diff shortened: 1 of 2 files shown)


murggel:
@eddiemuc thanks, will check that out

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
```diff
@@ -136,10 +137,14 @@ private static class FormulaNode {
 
             if (this.neededVars.isEmpty() && !hasRanges(this)) {
                 //this means that function is constant!
-                this.function = createConstantFunction();
+                final Pair<Value, Func3<ValueList, Function<String, Value>, Integer, Value>> constantPair = createConstantFunction();
```
eddiemuc: Note to self: this is far too complex. Member `constantValue` is not needed at all. Moreover, by introducing it I skipped the possibility that constant evaluation raises an exception. Remove this crap again.

All we need are two methods in FormulaNode:
- `isConstant()`: returns true if getNeededVars() returns an empty set AND hasRanges() returns false
- getConstantValue(): returns null if `isConstant()` is false. Otherwise calls `function` with empty var set and returns result (or raises exception)

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
```diff
@@ -858,11 +863,22 @@ private FormulaNode parseFunction(final String functionName) {
         }
 
         final FormulaFunction formulaFunction = Objects.requireNonNull(FormulaFunction.findByName(functionName));
+        final FormulaNode[] paramNodes = params.toArray(new FormulaNode[0]);
+        final List<Value> constantValues = new ArrayList<>();
+        for (FormulaNode paramNode : paramNodes) {
+            if (paramNode.constantValue != null) {
```
eddiemuc: Note to self:
Remove this, non-constant parameters should be added as null values, so the function can know at compile time how many parameters were given and on which position those were located.

eddiemuc (commented):
I didn't do it right the first time...

eddiemuc:
...and now I can't ask copilot for help because this is a cross-repo-PR. This is really a downside! Hope GitHub fixes this soon

murggel:
> I didn't do it right the first time...

Ok, then I will look first for a different approach to parse and expand ranges.
My first one (from copilot) seems also way too complex

eddiemuc:
> > I didn't do it right the first time...
> 
> Ok, then I will look first for a different approach to parse and expand ranges.
> My first one (from copilot) seems also way too complex 

I think I will cleanup and include this extension to formula framework anyway because it is a nice feature enhancement to write functions.

eddiemuc:
Fixed bad code and commented out example code (I left it in so murggel can take a look).
Ready for review imho

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
eddiemuc: done

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
eddiemuc: done

eddiemuc:
@murggel can I merge this PR or will this mess with your work on sum and refactoring?

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
```diff
@@ -858,11 +874,20 @@ private FormulaNode parseFunction(final String functionName) {
         }
 
         final FormulaFunction formulaFunction = Objects.requireNonNull(FormulaFunction.findByName(functionName));
+        final FormulaNode[] paramNodes = params.toArray(new FormulaNode[0]);
+        final List<Value> constantParamValues = new ArrayList<>();
+        for (FormulaNode paramNode : paramNodes) {
+            constantParamValues.add(paramNode.getConstantValue());
```
murggel, line 880: You need to check, whether paramNode.isConstant, otherwise you will add a nullptr.

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
```diff
@@ -858,11 +874,20 @@ private FormulaNode parseFunction(final String functionName) {
...
+        }
+
+        final Set<String> explicitelyNeededVars = formulaFunction.getExplicitelyNeededVars(constantParamValues);
```
murggel: If you want you can fix the typo ... `explicitly`

murggel:
> @murggel can I merge this PR or will this mess with your work on sum and refactoring?

This will not mess with my work - i just tested it and found a bug -> see my review comment

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
eddiemuc: That's intentional. It gives the implementing function the chance to check at compile time:
- how man parameters were passed
- which ones were constant and which ones were not

main/src/main/java/cgeo/geocaching/utils/formulas/Formula.java
eddiemuc: Yes

eddiemuc: done

murggel: The typo is in all related names...

eddiemuc: corrected there as well

murggel:
@eddiemuc thanks, will continue my work on this base -> PR  #17827

Looks promising

PR #17834, opened 2026-02-08
fix #17812: Implement multi-connector owner search support with ConnectorUsernameMap (extends #17812)

JakeDot:
## Description

Refactors connector-username mapping to use a dedicated type with proper nullability annotations. Addresses review feedback from PR jakedot#197 to replace the magic string "default" with explicit `null` values and introduce a specialized HashMap class.

Addresses JakeDot's review comments on PR https://github.com/cgeo/cgeo/pull/17812. The original implementation had a critical filtering bug where OR(Owner1, Owner2) would incorrectly match any owner on any connector. For example, user "JakeDot" on geocaching.com and "JakeDotNet" on opencaching would also match JakeDotNet's caches from geocaching.com.

### Filter Logic Fix

    Changed from OR(Owner="JakeDot", Owner="JakeDotNet")
    To OR(AND(Connector=GC, Owner="JakeDot"), AND(Connector=OC, Owner="JakeDotNet"))
    Special case: null key matches owner on any connector (backward compatibility)

**Key Changes:**
- Created `ConnectorUsernameMap` class extending `HashMap<@Nullable String, @NonNull String>` where `null` keys represent "any connector"
- Replaced all `Collections.singletonMap("default", username)` usages with `null`-keyed `ConnectorUsernameMap` instances
- Updated method signatures in `CacheListActivity.startActivityOwner()`, `OwnerGeocacheListLoader` constructor, and `SearchActivity` methods
- Added `Intents.EXTRA_CONNECTOR_USERNAMES` constant for Intent serialization
- Implemented proper deserialization with type safety and error handling in `CacheListActivity`
- Simplified filter logic in `OwnerGeocacheListLoader` to check `platformName == null` directly instead of comparing to "default" string

## Related issues

rel #17812

## Additional context

The changes maintain the same functional behavior while improving type safety and code clarity. A `null` key in the map represents a username that should match any connector, while a non-null key represents a specific connector name.

<details>

<summary>Original prompt</summary>

> Based on PR https://github.com/JakeDot/cgeo/pull/197: change the Connector-Username Map type to a new class extending HashMap<@Nullable String, @NonNull String>

</details>

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/ConnectorUsernameMap.java
+++ b/main/src/main/java/cgeo/geocaching/connector/ConnectorUsernameMap.java
@@ -1,0 +1,26 @@
+package cgeo.geocaching.connector;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.HashMap;
+
+/**
+ * A specialized HashMap for mapping connector names to usernames.
+ * Keys can be null (representing "any connector") or non-null connector names.
+ * Values are non-null usernames.
+ * 
+ * This class implements Serializable to support passing the map via Android Intent extras.
+ */
+public class ConnectorUsernameMap extends HashMap<@Nullable String, @NonNull String> {
+    
+    private static final long serialVersionUID = 1L;
+    
+    public ConnectorUsernameMap() {
+        super();
+    }
+    
+    public ConnectorUsernameMap(final int initialCapacity) {
+        super(initialCapacity);
+    }
+}
--- a/main/src/main/java/cgeo/geocaching/loaders/OwnerGeocacheListLoader.java
+++ b/main/src/main/java/cgeo/geocaching/loaders/OwnerGeocacheListLoader.java
@@ -1,21 +1,29 @@
 package cgeo.geocaching.loaders;
 
+import cgeo.geocaching.connector.ConnectorUsernameMap;
+import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.filters.core.AndGeocacheFilter;
 import cgeo.geocaching.filters.core.GeocacheFilterType;
 import cgeo.geocaching.filters.core.IGeocacheFilter;
+import cgeo.geocaching.filters.core.OrGeocacheFilter;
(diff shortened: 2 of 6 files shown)


main/src/main/java/cgeo/geocaching/loaders/OwnerGeocacheListLoader.java
```diff
@@ -25,8 +33,43 @@ public GeocacheFilterType getFilterType() {
 
     @Override
     public IGeocacheFilter getAdditionalFilterParameter() {
-        final OwnerGeocacheFilter ownerFilter = GeocacheFilterType.OWNER.create();
-        ownerFilter.getStringFilter().setTextValue(username);
-        return ownerFilter;
+        final OrGeocacheFilter aggregator = new OrGeocacheFilter();
```
eddiemuc, line 36: Don't use Or-Filter. They don't work well with Online-Fiters

main/src/main/java/cgeo/geocaching/connector/ConnectorUsernameMap.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.HashMap;
+
+/**
+ * A specialized HashMap for mapping connector names to usernames.
+ * Keys can be null (representing "any connector") or non-null connector names.
+ * Values are non-null usernames.
+ * 
+ * This class implements Serializable to support passing the map via Android Intent extras.
+ */
+public class ConnectorUsernameMap extends HashMap<@Nullable String, @NonNull String> {
```
eddiemuc, line 15: Why is a specialized Map class needed?

eddiemuc (changes requested):
This PR uses a wrong approach. It tries to fiddle in the approach specifically for online list searches and creates Or-Filters for it. I am not sure if this even works.

Worse: the approach won't work in other live places (eg map) this way.

Better approach would be to include the owner-map into the Owner-filter itself

eddiemuc:
I think this PR approaches a wrong path and suggest its deletion. @JakeDot unless you can explain exactly how your PR is supposed to solve the problem at hand

PR #17839, opened 2026-02-09
Allow simple lists to become parent lists (fix #17679)

### LINKED ISSUE #17679: Lists are not able to be placed under existing list
### Describe your problem!

When saving a PQ to a new list, there is no option to use any of the pre-existing lists as a parent, only (none) and (create new list).

Also on renaming lists.

Image

Image

### How to reproduce?

Create new / rename list, try to select parent.

_No response_

_No response_

### Reproducible

Yes

### c:geo Version

nightly

### System information

```text

```

### Additional Information

_No response_

moving-bits:
## Description
Add non-parent lists to list creation dialog to enable easy creation of nested lists from existing non-parent lists.

First group is list of existing parent lists, marked by ending colon, sorted alphabetically.
Second group is list of non-parent lists, without ending colon, sorted alphabetically.

## Example
There is an existing non-parent list named "Berlin".
Let's add the cache to a newly created list "Moabit", which is a sub-list of "Berlin":

|Open list assignment|Create new list "Moabit"|
|---|---|
|image|image|

|New list is assigned|"Berlin" is now a parent list"|
|---|---|
|image|image|

### DIFF
--- a/main/src/main/java/cgeo/geocaching/list/StoredList.java
+++ b/main/src/main/java/cgeo/geocaching/list/StoredList.java
@@ -92,7 +92,7 @@ public final class StoredList extends AbstractList {
         private final WeakReference<Activity> activityRef;
         private final Resources res;
 
-        private static final String GROUP_SEPARATOR = ":";
+        public static final String GROUP_SEPARATOR = ":";
 
         public UserInterface(@NonNull final Activity activity) {
             this.activityRef = new WeakReference<>(activity);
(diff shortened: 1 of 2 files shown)


moving-bits:
retest this please

bekuno:
retest this please

moving-bits:
rebased + resolved merge conflicts

moving-bits:
failing tests (`testAvatar` and `testGetOCCache`) are unrelated, so let's merge

ztNFny:
This PR in its current form causes multiple issues, see https://github.com/cgeo/cgeo/issues/17679

While fixing those specific bugs I was really wondering whether this PR makes sense or should be reverted in its entirety.:
- Users working with groups expect the existing groups to be selectable - now they also get a list of all first-level lists
- If we have both "groups" and "lists" in the list then to work-around the previous point the groups are listed before lists, but that breaks alphabetic ordering, making the resulting list quite confusing if you don't know why it's the way it is
- For creating new hiearachies there's already a "create new parent list" option

imho the benefit added by this PR is far outweighted by the negative UX impact it has.

I'll nevertheless continue on the PR to fix the bugs - but I'd like to start the discussion about reverting this.

PR #17847, opened 2026-02-13
Remove WhereYouGo helper method QueryMapFile

moving-bits:
## Description
Removes the old helper method for WhereYouGo
(which allowed to read the non-SAF path to a Mapsforge 3 compatible map file from c:geo by WhereYouGo)


moving-bits:
retest this please

PR #17827, opened 2026-02-06
Fix #17540: Add range-formulas

murggel:
## Description

This PR adds some range-based formulas to the function-selection, for which a range can be specified
* add
* multiply
* min
* max
* avg
* count

## Related issues

fix #17540

## Additional context

To discuss, if all of them go into the section "Complex numeric" or e.g. count in "Simple numeric"

grafik

<br>
grafik

grafik

grafik

PR #17851, opened 2026-02-14
Enhance add range formula

murggel:
## Description

* Fix type
* Add new group "Aggregate Numeric"
* Use ErrorValue.isError

## Related issues

rel #17540
rel PR #17827

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/formulas/FormulaError.java
+++ b/main/src/main/java/cgeo/geocaching/utils/formulas/FormulaError.java
@@ -32,7 +32,7 @@ final class FormulaError {
             super(errorString);
         }
 
-        public static ErrorValue create(final CharSequence cs) {
+        public static ErrorValue of(final CharSequence cs) {
             return new ErrorValue(cs);
         }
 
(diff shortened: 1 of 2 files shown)


bekuno:
retest this please

main/src/main/java/cgeo/geocaching/utils/formulas/FormulaError.java
```diff
@@ -32,7 +32,7 @@ protected ErrorValue(@NonNull final CharSequence errorString) {
             super(errorString);
         }
 
-        public static ErrorValue of(final CharSequence cs) {
+        public static ErrorValue create(final CharSequence cs) {
```
eddiemuc: This is very nitty.

I know this is a big debate in coding. However I prefer "of" over "create" as name of static object creators because:
- it is shorter 
- it hides the fact how the returnes object is produced. For example, using "of" I could later introduce a caching mechanism or something else without changing the method name

But again, this is nitty and no reason not to merge.

murggel: Ok, I wasn't aware of that common usage  - will change it back then

murggel: Done

PR #17874, opened 2026-02-19
Add GC Live API integration with OAuth2 authentication

baiti:
Route all geocaching.com operations through the official Live API (api.groundspeak.com/v1) instead of HTML scraping, with a toggle in Settings to switch between API and scraping. Includes OAuth2 + PKCE authorization flow, batch cache retrieval, parallel log/spoiler fetching, and finds count display in cache details.

This is a PoC, do not merge.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -121,6 +121,15 @@
             android:windowSoftInputMode="adjustNothing" >
             <!-- 'adjustNothing' fixes the issue described in #11610 -->
             <!-- don't add configChanges="orientation|screenSize" as we use different base layouts depending on the device orientation -->
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data
+                    android:scheme="${gcLiveRedirectScheme}"
+                    android:host="oauth2.callback"
+                    android:path="/callback" />
+            </intent-filter>
         </activity>
         <activity android:name=".DBInspectionActivity"
             android:exported="false"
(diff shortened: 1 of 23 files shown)


JakeDot:
Using the authorize live api button gives me a 500 error. Did you use a secret key not yet configured in CI?

baiti:
> Using the authorize live api button gives me a 500 error. Did you use a secret key not yet configured in CI?

Yes, this needs to be resolved by putting keys into the CI, without it, it won't run.

moving-bits:
> > Using the authorize live api button gives me a 500 error. Did you use a secret key not yet configured in CI?
> 
> Yes, this needs to be resolved by putting keys into the CI, without it, it won't run.

IIRC those keys are personalized and will be given out by GC only after signing some NDA / compliance agreement.

If we would decide to add GC API to c:geo (which would need quite some discussion with GC first), we need to think of a way to not expose that key - eg., it must not become part of the APK.

baiti:
I have iteratively fixed the issues surfaced by co-pilot's review, knowing that there might be leftovers. At this time I believe it makes no sense to resolve them all. This can be done later, should we ever decide to do anything serious with this PoC. It would be perfectly ok with me if the team decided (a) yes, let's pursue this but (b) throw away that PoC and build it new, for real, and by those who understand this codebase much much better than I do.

A word on the keys issue: I have deliberately not included any such keys anywhere. Should we ever want to pursue this, it is technically possible to handle the keys in compliance with the T's & C's. How this can be done we have discussed already years ago in various threads.

Let me highlight again what this PR is all about:

1. A PoC to show what it means to drive the app via the API. I wanted to see this primarily for myself to understand the implications, whether it would fit into cgeo's concepts, the structure, the effort. I wanted to see results on performance and some other metrics.

3. I wanted to find out what can be accomplished with today's AI capabilities. In this case I have used Claude Code but could also have done it with Google's Antigrativy (which is reportedly even better). And BTW: I did not let this thing run loose on the code base, I deliberately instructed it to understand the coding style, the coding pattern, the checkstyle requirements, the codacity. To what extent it was able to obey to these orders, well, this needs to be judged by the team reviewing this code.

Issue #16360, opened 2024-11-17
make "all offline caches" mode available on map (in addition to "live")


ztNFny:
map in c:geo - at least the old one, for Unified i'm not 100% sure - has 3 modes:
- display only the selected object (caches, cache, zone, waypoint) (Transient)
- Live: Display all offline stored caches + query the online services
- Offline: Display all offline stored caches
"Offline" mode can be experienced by mapping a single cache, enabling live and disabling it again. Map will then show all offline stored caches.

I'd request the "live" button to be enhanced with a long-tap action to not just toggle live on/off but to explicitly enable "offline" mode. As this third map mode is not something that most users are even aware about I'd say "hiding" it behind a long-tap action would be fine.

Usecase:
- Circumstances: I do all my caching with offline data
- Goal: I'm mainly doing a multi but in case I walk past a cache on the route I of course want to find it.
- In the multi I map the next waypoint
- On map I enable "live" to see other caches as well
- This however now sends queries to the services, which a) cause traffic I don't want, b) load cache information I don't want, e.g. if there's a cache in my local list with corrected coords but online they aren't corrected then that cache jumps to the original coords.

SammysHP:
It is either online, meaning that c:geo also queries online source in addition to locally stored caches, or offline, meaning that it won't send any network requests. What is shown in the offline mode depends on the situation:

- single cache: open the map for a single cache (and its waypoints), online is disabled.
- list: shows all caches of a list, online is disabled.
- live map with online disabled: this shows the result of the live map, but without asking the servers for more caches.

This means that "live/offline" is kind of a separate state not related to what is shown on the map. It simply enables or disable network requests.

Now, how to show all offline stored caches? Open the "All" list and show it on the map, or go to the live map and select the "All" list from the drop down list at the top.

ztNFny:
> Now, how to show all offline stored caches? Open the "All" list and show it on the map, or go to the live map and select the "All" list from the drop down list at the top.

I know that this is a way - however that way doesn't match the usecase at all.
To use the wording you used in your list of the 3 possibilities (which afaict is just a rewording of mine) I want an option to enable "live map with online disabled" from all map views, in addition to the existing "live map with online enabled" aka Live button

ztNFny:
> This means that "live/offline" is kind of a separate state not related to what is shown on the map. It simply enables or disable network requests.

[...]

PR #17883, opened 2026-02-20
Fix #17882: restore long click on enable live

### LINKED ISSUE #17882: [Nightly] Long click on "Live-Mode" does not work any more
### Describe your problem!

UnifiedMap: Long click on "Enable Live" does not work as before 

### How to reproduce?

Select a list
Open Map for the list
Long click on "enable live"

Tooltip is shown

All offline stored caches should be displayed (see PR #16753)

### Reproducible

Yes

### c:geo Version

2026.02.20-NB-29e4aeb

### System information

```text

```

### Additional Information

_No response_

murggel:
## Description

Restore behaviour of long click on enable live, old implementation probably does not work with new toolbar (PR #17775) -> move it into post-handler (see working menu_select_mapview)

## Related issues

Fix #17882
rel #16753
rel #16360

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -1498,10 +1498,10 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
             final View liveButton = findViewById(R.id.menu_map_live);
             if (liveButton != null) {
                 liveButton.setOnLongClickListener(v -> {
-                    // viewModel.mapType = UnifiedMapType.getPlainMapWithTarget(viewModel.mapType); // switch to PLAIN mode
-                    //  viewModel.transientIsLiveEnabled.setValue(false);
+                    viewModel.mapType = UnifiedMapType.getPlainMapWithTarget(viewModel.mapType); // switch to PLAIN mode
+                    viewModel.transientIsLiveEnabled.setValue(false);
                     Settings.setLiveMap(false);
-                    // reloadCachesAndWaypoints();
+                    reloadCachesAndWaypoints();
                     return true;
                 });
             }


main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
```diff
@@ -1504,6 +1494,17 @@ private void initializeMapViewLongClick() {
...
                 });
             }
+
+            final View liveButton = findViewById(R.id.menu_map_live);
+            if (liveButton != null) {
+                liveButton.setOnLongClickListener(v -> {
+                    // viewModel.mapType = UnifiedMapType.getPlainMapWithTarget(viewModel.mapType); // switch to PLAIN mode
+                    //  viewModel.transientIsLiveEnabled.setValue(false);
+                    Settings.setLiveMap(false);
+                    // reloadCachesAndWaypoints();
```
moving-bits: This will change c:geo's behavior on long-tapping the button:

While the old behavior only _temporarily_ switches to showing all stored caches without enabling live mode, the new behavior would disable live mode.

Also: Will just disabling live mode (which might not have been enabled at that point at all) lead to reloading the stored caches?

IIRC, the main use case for long-tapping on "enable/disable live mode" button was this:
- Map a stored list
  - This shows a map without live map, displays only caches from the selected list
- If you then want to add all stored caches, without enabling live mode, you can do so by long-tapping that button

I cannot test the PR currently, but from looking at the code I think this won't work any more with this PR.

murggel: Sorry, I missed to uncomment the code again, that was just for testing purpose if the code would be reached...

Will change it as soon as I have access to my computer.

murggel:
```suggestion
                    viewModel.mapType = UnifiedMapType.getPlainMapWithTarget(viewModel.mapType); // switch to PLAIN mode
                    viewModel.transientIsLiveEnabled.setValue(false);
                    Settings.setLiveMap(false);
                    reloadCachesAndWaypoints();
```

moving-bits:
retest this please

PR #17889, opened 2026-02-21
Reserve space for navigation area in settings (fix #17878)

### LINKED ISSUE #17878: Edge2Edge: Settings go behind Android's navigation bar
### Describe your problem!

My screen size is "perfect" in the way that in e. g. ``Backup / Restore`` I can no longer change the value of ``Automatic backup``.  ;)

### How to reproduce?

Just open settings and have a look at Android's nav bar ...

... it's pretty transparent, but there's no way to access the entries behind it.

There should be an offset to the nav bar, so that scrolling brings the contents above it.

### Reproducible

Yes

### c:geo Version

2026.02.19-NB-c4a23aa

### System information

```text

```

### Additional Information

_No response_

moving-bits:


moving-bits:
Failing test (`testGetOCCache`) is unrelated

PR #17891, opened 2026-02-21
Support coloring statusbar (rel. to #17775)

moving-bits:
## Description
Colors statusbar the same color as actionbar.
(For CacheDetailActivity, this will be a color depending on cache type, if enabled in settings.)
Fixes the remaining open issue of #17775.

This is achieved by:
- using inset.top as additional padding to toolbar, which needs to be increased in height accordingly
- by this, toolbar's background color will be used for top padding (= statusbar area) as well
- finally, set inset.top=0, as the added padding will fill the space
- spacer will get a similar adjustment

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
@@ -10,7 +10,6 @@ import cgeo.geocaching.utils.Log;
 import cgeo.geocaching.utils.MapMarkerUtils;
 import cgeo.geocaching.utils.TextUtils;
 
-import android.content.res.Configuration;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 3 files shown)


moving-bits:
Failing tests (`testSearchTrackable` and `testGetOCCache`) are unrelated

murggel:
I would color the status bar for all actionbar-activities. For the non-cachedetail-activity it would be the "normal" color of the actionbar

moving-bits:
> I would color the status bar for all actionbar-activities. For the non-cachedetail-activity it would be the "normal" color of the actionbar

implemented for `AbstractActionBarActivity` and its descendants

murggel:
@moving-bits I can provide a PR for the correct text-color of the status bar, if you want or just post the code here...
Don't know, if I can push directly into you PR...

moving-bits:
You can either add commits to this PR or give me instructions what to add in my PR, so that I can add it, whatever you prefer.

murggel:
> You can either add commits to this PR or give me instructions what to add in my PR, so that I can add it, whatever you prefer.

It's not much, just changing in `ActionBarUtils.setSystemBarAppearance`   `windowInsetsController.setAppearanceLightStatusBars(isLightSkin);` to `windowInsetsController.setAppearanceLightStatusBars(false);` should be enough

moving-bits:
Thanks, I've added it to the PR

murggel:
Thanks for diving into it

PR #17894, opened 2026-02-22
Refactor openwig code to Java 8, some readability and other optimizations

JakeDot:
Modernises the copied openWIG/Kahlua engine code to Java 8 idioms and converts all `*Lib` dispatch classes into enums, eliminating manual index bookkeeping and replacing raw `int` screen constants with a type-safe enum.

Adds type safety to raw collections.

Inlines many one line interface implementations to lambdas.

### Java 8 modernisation
- **Lambdas** — `JavaFunction`/`Runnable` anonymous inner classes → lambdas; `@FunctionalInterface` added to `JavaFunction`
- **Collections** — `Vector` → `ArrayList`/`List`; `Hashtable` → `HashMap`/`Map`; `BackgroundRunner` queue: `Vector` → `ArrayDeque` with `synchronized pollTask()` to preserve thread safety without holding the lock during execution
- **Loops** — `Enumeration`+`while` → `entrySet()` for-each; indexed `elementAt()` loops → enhanced for-loops
- **Boxing** — `new Double(x)` → `xd` literals or `(double)` casts; `new Integer(x)` → `Integer.valueOf(x)`
- **Misc** — `StringBuffer` → `StringBuilder`; star imports → explicit imports

### `UI.Screen` enum
Replaces 6 raw `int` constants on the `UI` interface with a nested `Screen` enum. `ordinal()` provides the Lua protocol values (declaration order = 0–5); `fromId(int)` does a bounds-checked `values()[id]` array lookup.

```java
// Before
public static final int MAINSCREEN = 0;
public void showScreen(int screenId, EventTable details);

// After (nested inside UI interface)
enum Screen { MAINSCREEN, DETAILSCREEN, INVENTORYSCREEN, ITEMSCREEN, LOCATIONSCREEN, TASKSCREEN; … }
public void showScreen(Screen screen, EventTable details);
```

`WherigoThingType` field changed from `int wherigoScreenId` to `UI.Screen screen`; `THING` uses `null` instead of the sentinel `-1`.

### `*Lib` classes → enums
`WherigoLib`, `OsLib`, `CoroutineLib`, `TableLib`, and `BaseLib` converted from classes with numeric index fields and parallel arrays to `enum implements JavaFunction`. Each constant's `ordinal()` is its Lua protocol position — no separate id field or `names[]` array needed. `call()` switches directly on `this`.

```java
// Before: class with int index, names[] array, switch(index)
// After:
enum WherigoLib implements JavaFunction {
    ZONE_GETDISTANCE, ZONE_CONTAINSPOINT, … ;
    @Override public int call(LuaCallFrame callFrame, int nArguments) {
        switch (this) {
            case ZONE_GETDISTANCE: …
        }
    }
}
```

### DIFF
--- a/main/src/main/java/cgeo/geocaching/wherigo/kahlua/stdlib/StringLib.java
+++ b/main/src/main/java/cgeo/geocaching/wherigo/kahlua/stdlib/StringLib.java
@@ -713,7 +713,7 @@ public enum StringLib implements JavaFunction {
         int offset = ii - 1;
         for (int i = 0; i < nReturns; i++) {
             char c = s.charAt(offset + i);
-            callFrame.set(i, new Double((double) c));
+            callFrame.set(i, (double) c);
         }
         return nReturns;
     }
@@ -797,7 +797,7 @@ public enum StringLib implements JavaFunction {
             Object[] caps = new String[level];
             for (int i = 0; i < level; i++) {
                 if (capture[i].len == CAP_POSITION) {
-                    caps[i] = new Double(src_init.length() - capture[i].init.length() + 1);
+                    caps[i] = (double) (src_init.length() - capture[i].init.length() + 1);
                 } else {
                     caps[i] = capture[i].init.getString().substring(0, capture[i].len);
                 }
@@ -906,7 +906,7 @@ public enum StringLib implements JavaFunction {
             if (l == CAP_UNFINISHED) {
                 throw new IllegalStateException("unfinished capture");
             } else if (l == CAP_POSITION) {
-                Double res = new Double(ms.src_init.length() - ms.capture[i].init.length() + 1);
+                final Double res = Double.valueOf(ms.src_init.length() - ms.capture[i].init.length() + 1);
                 ms.callFrame.push(res);
                 return res;
             } else {
@@ -983,7 +983,7 @@ public enum StringLib implements JavaFunction {
                 ms.level = 0;
                 if ( ( res = match ( ms, s1, p ) ) != null ) {
                     if ( find ) {
-                        return callFrame.push(new Double(s.length () - s1.length () + 1), new Double(s.length () - res.length ())) +
+                        return callFrame.push((double) (s.length() - s1.length() + 1), (double) (s.length() - res.length())) +
                         pushCaptures( ms, null, null );
                     } else {
                         return pushCaptures( ms, s1, res );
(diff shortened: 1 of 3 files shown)


main/src/main/java/cgeo/geocaching/wherigo/WherigoGame.java
```diff
@@ -459,7 +438,31 @@ public void showScreen(final UI.Screen screen, final EventTable details) {
...
+    private void showMainOrListScreen(final UI.Screen screen) {
+        //close open dialog if any
+        WherigoDialogManager.get().clear();
+        //special: also close open thing dialogs which might have been opened directly
+        WherigoThingDialogProvider.closeAllThingDialogs();
 
+        //jump to main screen: if we are already displaying the main screen then do nothing
+        final Activity currentActivity = CgeoApplication.getInstance().getCurrentForegroundActivity();
+        if (currentActivity instanceof WherigoActivity) {
+            return;
```
bekuno: this is not handled in the calling method

main/src/main/java/cgeo/geocaching/wherigo/WherigoGame.java
```diff
@@ -459,7 +438,31 @@ public void showScreen(final UI.Screen screen, final EventTable details) {
...
+        }
+        //if we can jump to main screen then do it
+        if (currentActivity != null) {
+            WherigoActivity.start(currentActivity, false);
+            return;
```
bekuno: here the same

main/src/main/java/cgeo/geocaching/wherigo/WherigoGame.java
bekuno: WTF?
I marked the "return" statement. This was already there before the method was extracted. So, the meaning was another in the code before.
@JakeDot Mabe it is easier to do it by your hand instead of a AI which loses the context.

main/src/main/java/cgeo/geocaching/wherigo/WherigoGame.java
bekuno: Also here we have a return statement.
Please check/correct it in the same manner as suggested before.

main/src/main/java/cgeo/geocaching/wherigo/kahlua/stdlib/BaseLib.java
```diff
@@ -38,268 +38,210 @@ of this software and associated documentation files (the "Software"), to deal
 import java.util.Locale;
 import java.util.function.Function;
```
JakeDot: @copilot Remove ordinal comments here and in other enums

JakeDot:
### OsLib date/time

All Calendar/Date/TimeZone replaced with java.time (`ZonedDateTime, ZoneId, Instant, WeekFields`)
Hard-coded day/month name arrays (`SHORT_DAY_NAMES, LONG_DAY_NAMES, SHORT_MONTH_NAMES, LONG_MONTH_NAMES`) removed; replaced with `DayOfWeek/Month.getDisplayName(TextStyle, Locale.ROOT)`

PR #17898, opened 2026-02-24
use local timezone (of device, not event) for calendar entries

ztNFny:
Use the devices local time for adding calendar entries instead of GMT.
Before this when you're adding an calendar entry for an event in Germany (GMT+2) at 12:00 the calendar would show 10:00 (GMT) which is correct but confusing, with this it'll show 12:00 GMT+2.

It doesn't change anything about the event date / timezone detection as that info simply isn't available, so adding calendar entires for events in another timezone is just as broken as it was before ;-)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ concurrency:
 on:
   push:
   pull_request_target:
-    types: [labeled, synchronize]
+    types: [opened, labeled, synchronize, reopened]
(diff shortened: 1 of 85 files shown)


ztNFny:
retest this please

ztNFny:
failing tests are unrelated

ztNFny:
retest this please

ztNFny:
retest this please

Issue #17904, opened 2026-02-25
CI tests for all are not running without ok-to-test


bekuno:
For all user now the tests are guarded with the `okto-test` label.

see also https://github.com/cgeo/cgeo/issues/17871#issuecomment-3957461254

kumy:
Until last week, All jobs were running twice, once for the branch push, and once for the PR. I noticed that last week, and pushed a commit to the skip the PR one only.
I checked the linked PR, I can see the skipped, but I can also click "see successful", and there I see the other set of CI jobs.
(I m on mobile and cannot easily attach screenshots sorry).
@bekuno Is this just an UI issue, or I mis understood the issue?

kumy:
To note, I also did Another commit to completely skip the integration tests (check style still run) in case the PR does not modify cgeo code (eg only when files in .github/) are updated

kumy:
He are the screenshots

We see the skipped jobs first
![Screenshot_20260226-070743.png](https://github.com/user-attachments/assets/07802984-4c47-4ee6-b23e-72654820c302)

But 7passed 7skipped
![Screenshot_20260226-070743_1.png](https://github.com/user-attachments/assets/650bf582-3d6d-4fb6-9d30-27fa83947b2f)

On the details, we see both sets
![Screenshot_20260226-070847.png](https://github.com/user-attachments/assets/33f7fbad-1e87-489a-b0da-6f9920a97038)

kumy:
And now I see the issue title ; without ok to test ci is not running. This should be the case only for pr from external repo, and should not be necessary for branch PR, i ll check that!

eddiemuc:
> And now I see the issue title ; without ok to test ci is not running. This should be the case only for pr from external repo, and should not be necessary for branch PR, i ll check that!

The vast majority of PRs is from external repo, not from internal branches. Not sure if we need special handling for the few branch PRs we have

kumy:
I checked the linked PR, and the source is from external repo, so yes the PR must require the label to the CI.

I will see if I can detect if the pr author has write permission on the repo, and *auto label the PR*. We already do such check to check for the secrets access

kumy:
@eddiemuc The separate handling is mainly for security issue (this is logged in the initial (still open) issue about the ci work, let me find that)

kumy:
@eddiemuc see around this message https://github.com/cgeo/cgeo/issues/15608#issuecomment-2057870244

eddiemuc:
[...]

PR #17905, opened 2026-02-26
ci: Automatically add label ok-to-test from contributors repo fork

kumy:
## Description

As discussed in #17904, here is a little addition that should be able to add the `ok-to-test` label for PR from c:geo contributors external repo fork.

## Related issues

Relates #17904

## Additional context

x-ref: https://github.com/cgeo/cgeo/issues/17904#issuecomment-3964474059

> > For the eat your own food, really? What I propose is half the way, you need the label to be set, but the step already present to check if you really have access to the secrets will also add itself the label ok to test. It s just a few lines...

> In my experience having a normal workflow for "common" users and a special way for "core" users leads to decay of the "normal" workflow because it is seldom used. We want (need) to make our system atteactive for development by non-core-users, we depend on them

/cc @eddiemuc 

Note: this cannot be tested directly in this repo without merge first

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -106,6 +106,14 @@ jobs:
           echo "::error::${{ github.triggering_actor }} does not have permissions on this repo.\nCurrent permission level is ${{ steps.checkAccess.outputs.user-permission }}\nJob originally triggered by ${{ github.actor }}"
           exit 1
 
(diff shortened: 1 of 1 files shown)


bekuno:
Let's test it.

bekuno:
@kumy I created a test PR #17928 after merging this PR. Please check if it worked as expected.

PR #17915, opened 2026-03-01
Prevent keyboard from overlapping certain fields (fix #17842)

### LINKED ISSUE #17842: Edge2edge: multiline textfields goes behind keyboard
Writing a long text in the personal note field it'll grow and expand behind the keyboard, meaning you can't see what you're writing.

Image

moving-bits:
## Description
Prevents waypoint user note textedit field and cache log textfield from overlapping with keyboard by applying padding to those input fields, that take keyboard size (and system bars) into account.


moving-bits:
retest this please

PR #17934, opened 2026-03-09
complete Simplified Chinese translations

grill-glitch:
complete Simplified Chinese translation


SammysHP:
Just in case you haven't figured it out yet: Our translations are managed through Crowdin and translation pull requests can't be accepted here. Please take a look here:

https://crowdin.com/project/cgeo

PR #17942, opened 2026-03-11
Keep parameters for user-defined tile providers

moving-bits:
## Description
A user on support wants to use a user-defined tile provider that requires additional URL parameters (like API key and language selection), which is currently not supported by c:geo.

This PR extends user-defined tile providers to keep query parameters in the configured tileprovider string.


bekuno:
retest this please

Issue #17935, opened 2026-03-09
Images from Dropbox not visible (User-Agent blocking)


Eselter:
### Describe your problem!

Images included in description from Dropbox are not visible.

### How to reproduce?

1. Open description of cache GCBKD9Q

### Actual result after these steps?

It display error image not actual image from Dropbox (postapo PC):
<img width="1080" height="2107" alt="Image" src="https://github.com/user-attachments/assets/3b7720f7-4021-4547-80d7-1f71046690a9" />

### Expected result after these steps?

You should see image from Dropbox (postapo PC):
<img width="1080" height="1085" alt="Image" src="https://github.com/user-attachments/assets/130b147d-4e97-437f-ae26-c336f95aaf0f" />

### Reproducible

Yes

### c:geo Version

2026.02.18

### System information

```text
## System information

c:geo version: 2026.02.18

Datetime: 10 marca 00:27

Device:
-------
- Device type: SM-S931B (pa1qxeea, samsung)
- Available processors: 8
- Android version: 16
- Android build: BP2A.250605.031.A3.S931BXXS8BZB5
- Screen resolution: 1080x2105px (360x702dp)
- Pixel density: 3.0
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 26.08.34 (260400-876566425)
- Memory:  Available:2,8 GB, Total:10,9 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v11, Count:196
- Set language: pl_PL (system default)
- System date format: d.MM.y
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: 2 wrz 2024, 01:14
- Last auto backup: 4 mar, 16:03
- Routing mode: Walk (shortest.brf / trekking.brf / car-eco.brf / (brak) / (brak))
- Map mode: UnifiedMap / live / OSM multi-threading: off
- Map: Google: Satelita
  - Id: cgeo.geocaching.unifiedmap.tileproviders.GoogleSatelliteSource
  - Attrs: 
[...]

PR #17943, opened 2026-03-11
Update User-agent

moving-bits:
## Description
Cherry-picked the commit from #17941 and targeted it to `release`


bekuno:
retest this please

moving-bits:
Even if it may not resolve #17935 in all constellations, it should not hurt, either, therefore let's merge it.

PR #17928, opened 2026-03-08
DNM: test for #17904 / #17905

bekuno:

PR #17953, opened 2026-03-15
ci: add trigger on pull_request_target: opened/reopened

kumy:
## Description

Add trigger on PR open/reopen so the task to add the label if the user is authorized is triggered

## Related issues

#17928 #17905 #17904 

## Additional context

Due to the `pull_request_target` the code need to land in `master` to be tested :/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ concurrency:
 on:
   push:
   pull_request_target:
-    types: [labeled, synchronize]
+    types: [opened, labeled, synchronize, reopened]
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -51,6 +51,7 @@ jobs:
       (
         github.event.pull_request.head.repo.full_name != github.repository &&
         (
+          github.event.action == 'opened' ||
```
bekuno: Also for action `reopend`?

kumy: ho, ok!, good catch 👀

bekuno:
Let's test it.

PR #17959, opened 2026-03-18
Fix #17958: Use color-selector for disabling actionbar-items

### LINKED ISSUE #17958: [Nigthly] disabled items in ActionBar cannot be distinguished from the enabled ones
Disabled items in ActionBar cannot be distinguished from the enabled ones.

Possible related to changes in the ActionBar

Image

Image

murggel:
Tried to find a solution with copilot (for #17958) - but I am not convinced, but maybe as a start / discussion point, it is good / bad enough

Disabled menu items (text + icons) are now visually distinguishable from enabled ones.

Root cause: The toolbar was switched to MaterialToolbar/MaterialComponents, which no longer automatically dims disabled menu item text and icons.

Fix:
- New ColorStateList color_text_selector: uses colorText with alpha=0.6 for state_enabled=false; referenced in text_default style (via android:itemTextAppearance) and appToolbarPopup theme (android:textColor + android:textColorPrimary)
- New ColorStateList action_bar_item_text_selector / menu_icon_tint_selector: same pattern for toolbar action icons and overflow menu icons respectively
- MenuUtils.tintMenuIcons: replaced setTint(int) with getColorForState() using an empty int[] for the disabled state (not -state_enabled, which does not match ColorStateList selectors) to resolve the correct dimmed color per item.isEnabled()
- MenuUtils.setEnabled: calls tintSingleIcon() immediately after setEnabled() so that action bar icons are updated at once, without waiting for the next onPrepareOptionsMenu cycle
- New android:actionMenuTextColor in tool_bar style for action button text dimming

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/MenuUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/MenuUtils.java
@@ -3,13 +3,15 @@ package cgeo.geocaching.utils;
 import cgeo.geocaching.R;
 
 import android.annotation.SuppressLint;
-import android.content.res.Resources;
+import android.content.Context;
+import android.content.res.ColorStateList;
 import android.graphics.drawable.Drawable;
 import android.view.Menu;
 import android.view.MenuItem;
 
 import androidx.appcompat.view.menu.MenuBuilder;
 import androidx.appcompat.view.menu.MenuItemImpl;
+import androidx.core.content.ContextCompat;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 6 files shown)


moving-bits:
I have not cross-checked all places, but at least changing so many colors to the new color selector seems suspicious to me.

Having tortured Gemini with something like 20+ questions finally something useful came out of it. 🤷 
See #17960 for an alternative and more local approach.

murggel:
Thanks - I will cross-check tomorrow

moving-bits:
Superseded by #17960, therefore closing here

PR #17960, opened 2026-03-18
Set colors for disabled menu state (fix #17958)

### LINKED ISSUE #17958: [Nigthly] disabled items in ActionBar cannot be distinguished from the enabled ones

moving-bits:
## Description
Fixes coloring of disabled menu items.

light mode:
image.image

dark mode:
image.image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -768,7 +768,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
         }
 
         MenuUtils.enableIconsInOverflowMenu(menu);
-        MenuUtils.tintToolbarAndOverflowIcons(menu);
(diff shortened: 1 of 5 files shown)


murggel:
Looks good, but the entries in the ActionBar are updated only when the menu is opened

grafik

grafik

moving-bits:
Good catch, thanks! Hadn't thought that we could have items visible on the toolbar already in a disabled state. I will extend the PR accordingly.

moving-bits:
I've changed the approach to extend the existing method `MenuUtils.tintMenuIcons()` to differentiate between enabled/disabled state + to also color the menu item's text (instead of hooking into `AbstractActionBarActivity.onMenuOpened()`, which gets only called for the overflow menu popup).

murggel:
Works for me

moving-bits:
Thanks for cross-checking. Let's put it in the nightlies, then.

PR #17989, opened 2026-04-06
rel to #17987: add DavidV Offline Translator as external translator

eddiemuc:
rel to #17987: add DavidV Offline Translator as external translator

Translator as of now only available via F-Droid. Still it seems worth including because it seems to be the only current purely offline translator accessible via standard Android integration (using Intent.PROCESS_TEXT).

F-Droid: https://f-droid.org/packages/dev.davidv.translator/
Source: https://github.com/DavidVentura/offline-translator


eddiemuc:
retest this please

moving-bits:
Looks like we are getting those "Could not create ADB Bridge" errors again. I'll try to contact @mucek4.

bekuno:
As we will have no beta cycle in the next time, I propose to add this also to the release branch.

eddiemuc:
> As we will have no beta cycle in the next time, I propose to add this also to the release branch.

The whole support for external translators is not in release branch I'm afraid. We are in dire need of a new beta cycle

PR #18000, opened 2026-04-09
Fix wherigo folder duplication on SAF downloads

### LINKED ISSUE #17995: [Wherigo / SAF] Downloading cartridge creates duplicate "wherigo (n)" folders instead of using existing directory
From support [Ticket #890951](https://cgeosupport.droescher.eu/scp/tickets.php?id=10164)

### Description
When a user downloads a cartridge, the file is not saved in the existing `wherigo` directory. Instead, a new directory named `wherigo (n)` is created each time (with `n` incrementing for every download).

### Steps to Reproduce
1. Trigger a cartridge download
2. Observe the target directory

### Expected Behavior
The cartridge file should be saved in the existing `wherigo` folder if it already exists.

### Actual Behavior
A new folder `wherigo (n)` is created for each download instead of reusing the existing directory.

### Impact
- Clutters the file system with multiple duplicate folders  
- Cartridge is not found by c:geo

### Additional Context
- also other folders are affected, but user didn't notice any other unusual behavior.
  - I am waiting for answer regarding field-notes and gpx-ex-/import
  - I am waiting for extended debug-log
- System seems to be rooted
  - Device type: OnePlus Nord (OnePlusNord, OnePlus)
  - Available processors: 8
  - Android version: 12
  - Android build: e_avicii-user 12 SQ3A.220705.004 eng.root.20260312.135323

utafrali:
## Description
Some SAF storage providers have stale caches and rename new directories if the name already exists (e.g. "wherigo (1)" instead of "wherigo"). This was creating duplicate folders instead of reusing the existing one.

Split out the directory search logic and added a check after creation. If the provider renamed it unexpectedly, we delete the misnamed directory and re-query to get the existing one.

## Related issues
Fixes #17995

## Additional context
N/A

### DIFF
--- a/main/src/main/java/cgeo/geocaching/storage/DocumentContentAccessor.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DocumentContentAccessor.java
@@ -360,22 +360,39 @@ class DocumentContentAccessor extends AbstractContentAccessor {
         return dir;
     }
 
-    private Uri findCreateSubdirectory(final Uri dirUri, final String dirName, final boolean createIfNotExisting) throws IOException {
+    private Uri findSubdirectoryInQuery(final Uri dirUri, final String dirName) throws IOException {
         final List<Uri> result = queryDir(dirUri, new String[]{DocumentsContract.Document.COLUMN_DOCUMENT_ID, DocumentsContract.Document.COLUMN_DISPLAY_NAME, DocumentsContract.Document.COLUMN_MIME_TYPE}, c -> {
             if (dirName.equals(c.getString(1)) && DocumentsContract.Document.MIME_TYPE_DIR.equals(c.getString(2))) {
                 return DocumentsContract.buildDocumentUriUsingTree(dirUri, c.getString(0));
             }
             return null;
-
         });
         for (Uri uri : result) {
             if (uri != null) {
                 return uri;
             }
         }
+        return null;
+    }
+
+    private Uri findCreateSubdirectory(final Uri dirUri, final String dirName, final boolean createIfNotExisting) throws IOException {
+        final Uri found = findSubdirectoryInQuery(dirUri, dirName);
+        if (found != null) {
+            return found;
+        }
         if (createIfNotExisting) {
             try {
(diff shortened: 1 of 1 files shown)


bekuno:
ok to test

main/src/main/java/cgeo/geocaching/storage/DocumentContentAccessor.java
```diff
@@ -360,22 +360,39 @@ private Uri getSubdirUri(@NonNull final Uri baseUri, @NonNull final List<String>
...
             try {
-                return DocumentsContract.createDocument(getContext().getContentResolver(), dirUri, DocumentsContract.Document.MIME_TYPE_DIR, dirName);
+                final Uri created = DocumentsContract.createDocument(getContext().getContentResolver(), dirUri, DocumentsContract.Document.MIME_TYPE_DIR, dirName);
+                if (created != null) {
+                    // Some providers (stale cache) rename the new dir (e.g. "wherigo (1)") when the real one already exists.
+                    // If that happened, delete the duplicate and re-query to return the existing dir.
+                    final ContentStorage.FileInformation fi = getFileInfo(created, null);
+                    if (fi != null && !dirName.equals(fi.name)) {
+                        DocumentsContract.deleteDocument(getContext().getContentResolver(), created);
+                        return findSubdirectoryInQuery(dirUri, dirName);
```
eddiemuc, line 392: When we are here, then this means that the same query returned null in line 379. 
Why should this return something else than null?

utafrali:
Good point - that section's a retry with a different cache source, so it might succeed where the previous query failed. I can clarify the logic in a code comment if that'd help though

eddiemuc:
> Good point - that section's a retry with a different cache source, so it might succeed where the previous query failed. I can clarify the logic in a code comment if that'd help though

The line in question calls the same method as the other line with the same parameters. Why would this second call go to a different cache source? Is a cache-refresh triggered by the create/delete logic between the two lines in question?

Did you test that your PR indeed fixes the problem? On which device or emulator?

eddiemuc:
Questions to @utafrali remain unanswered since 5 months. Assuming this PR is stale. It has conflicts too

Issue #7811, opened 2019-08-26
Fetch coords from map via button in coordinate dialog


Lineflyer:
From support mail (transcribed from spanish):

**Is your feature request related to a problem?**
I want to create a waypoint with coordinates selected from the map

**Describe the solution you'd like**
I learned about the possibility to open the map of a cache and tap for creating a new waypoint. This works, however it requires a lot of clicking and its not visible to users, who do not know this feature.

**Describe alternatives you've considered**
I suggest to add another button to the dialog where I can enter coordinates when creating a waypoint (currently: My coordinates, Cache coordinates, Clear coordinates) which reads "Select from map". This shall open the map and let me set a needle by clicking the map.

Dam297:
Can I contribute and have the issue assigned to me (I'm new)?

moving-bits:
> Can I contribute and have the issue assigned to me (I'm new)?

Sure, contributions are very welcomed. I've assigned you to this issue.

You may want to read our [README](https://github.com/cgeo/cgeo/blob/master/README.md), [Coding Conventions](https://github.com/cgeo/cgeo/wiki/Coding-Conventions) and maybe [c:geo and git](https://github.com/cgeo/cgeo/wiki/Working-on-c%3Ageo-for-git-beginners) for further info on how to contribute.

murggel:
I am glad and you are very welcome to contribute, but in the meantime we have the option to create a waypoint from long click on the map...

![Screenshot_20260301_175032_cgeo.jpg](https://github.com/user-attachments/assets/387110e9-5c0b-47f0-8d72-9328248b0bee)


So is this additional option in waypoint dialog still necessary?

Dam297:
So is this issue still valid?

Lineflyer:
I guess it is still a valid request as this is about a button in the coordinate dialog to pick the coords from the map (see description in initial post). 
IMHO this would make the feature more accessible to users.

ztNFny:
> I am glad and you are very welcome to contribute, but in the meantime we have the option to create a waypoint from long click on the map...
> 
> ![Screenshot_20260301_175032_cgeo.jpg](https://github.com/user-attachments/assets/387110e9-5c0b-47f0-8d72-9328248b0bee)
> 
> 
> So is this additional option in waypoint dialog still necessary? 

Imho the current solution fully covers the request. We shouldn't have multiple UIs for doing the same thing, that's just bad design.

SammysHP:
[...]

PR #18011, opened 2026-04-13
Select coords from map (Issue 7811)

Dam297:
## Description
Add new feature:
Fetch coords from map via button in coordinate dialog

## Related issues

#7811

### DIFF
--- a/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
@@ -872,7 +872,7 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
         super.onActivityResult(requestCode, resultCode, data);
 
         if (requestCode == UnifiedMapType.REQUEST_CODE_GET_COORDS && resultCode == RESULT_OK && data != null) {
-            Geopoint gp = new Geopoint(StringUtils.defaultString(data.getStringExtra("coords")));
+            final Geopoint gp = new Geopoint(StringUtils.defaultString(data.getStringExtra("coords")));
             if (listener != null) {
                 listener.showCoordinatesInputDialog(gp, cache);
             }
--- a/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
@@ -31,6 +31,7 @@ import cgeo.geocaching.ui.WeakReferenceHandler;
 import cgeo.geocaching.ui.dialog.CoordinateInputDialog;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
+import cgeo.geocaching.unifiedmap.UnifiedMapType;
 import cgeo.geocaching.utils.AndroidRxUtils;
 import cgeo.geocaching.utils.ClipboardUtils;
 import cgeo.geocaching.utils.CommonUtils;
@@ -213,6 +214,8 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
         }
     }
 
+    private static CoordDialogListener listener;
+
     private void nonEditable(final TextInputLayout textLayout, final TextView textView) {
         textLayout.setEndIconMode(TextInputLayout.END_ICON_NONE);
         textView.setKeyListener(null);
@@ -247,7 +250,8 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
             setTitle(LocalizationUtils.getString(R.string.waypoint_edit_title));
         }
 
-        binding.buttonLatLongitude.setOnClickListener(new CoordDialogListener());
+        listener = new CoordDialogListener();
+        binding.buttonLatLongitude.setOnClickListener(listener);
 
         final List<String> wayPointTypes = new ArrayList<>();
         for (final WaypointType wpt : WaypointType.ALL_TYPES_EXCEPT_OWN_AND_ORIGINAL) {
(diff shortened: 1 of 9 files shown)


bekuno:
Ok to test

ztNFny:
Could you please describe the value of having this vs the existing implementation of
- map cache
- long-tap on map
- "create waypoint"

This seems to do exactly the same but adding additional UI elements.

murggel:
@Dam297 sorry, I didn't want to update your branch, I clicked the wrong button on my phone, just wanted to" load more items"

murggel:
@Dam297  Can you add some screenshots, how the workflow inside the map would look like?

Dam297:
> @Dam297 Can you add some screenshots, how the workflow inside the map would look like?

As it was written before: it makes the feature more accessible to users.

Current1: Make waypoint -> Coords dialog -> Set coordinates... (if user knows coordinates or have them copied)

Current2: Make waypoint -> Coords dialog ->Back to map ->On map: long tap -> Remember or copy coordinates -> Back to make waypoint -> Coords dialog -> Set coordinates

New version: Make waypoint -> Coords dialog -> Select on map (button) -> On map: long tap -> Confirm coordinates (Copy coordinates) -> It returns auto back to coords dialog -> Set coordinates

If wanted I can make screenshots

murggel:
> > @Dam297 Can you add some screenshots, how the workflow inside the map would look like?
> ...
> New version: Make waypoint -> Coords dialog -> Select on map (button) -> On map: long tap -> Confirm coordinates (Copy coordinates) -> It returns auto back to coords dialog -> Set coordinates
> 
> If wanted I can make screenshots

Yes please, especially how / where the new button fit in the dialog and how the confirmation of the coordinates in the map after long tap will look like (dialog/menu/which text.. ) 
Thanks in advance

moving-bits:
Thanks for your contribution, very welcomed.

I've added some screenshots to be able to better explain some issues I have with the current implementation:

|Coord edit|long tap menu|confirmation|
|---|---|---|
|image|image|image|

- First screenshot: The menu entry for "Select from map" should get a different icon, maybe the map symbol? (Already available as icon in c:geo)
- Second screenshot: Why are you using a popup menu just for showing a "select coordinates" entry? I propose to skip this dialog in this specific mode and directly jump to the next dialog on long-tap on map. (This would also eliminate the, IMHO, ugly "menu-hacking" code.)
- Third screenshot: Why do you chose a manual "copy coordinates" way of implementation? I propose to change the action buttons in this dialog to "Cancel" and "Ok" when in this mode, close the dialog on cancel, but automatically transfer the selected coordinates back into the coordinate edit dialog on "ok", without having to manually copy&paste the coords.

murggel:
I am currently not sure, how this will work together, but at least I want to mention it for reference: 
In PR #18023 there is also a implementation regarding coordinates and map, there it is more the other way round to show the coordinates on map.

Dam297:
I'll look into it, but a little bit later.

moving-bits:
Setting "WIP / work in progress" label for now, as there are some open items for this PR

Dam297:
After thinking about it
>  The menu entry for "Select from map" should get a different icon, maybe the map symbol? (Already available as icon in c:geo)

This can be done.



> Why are you using a popup menu just for showing a "select coordinates" entry? I propose to skip this dialog in this specific mode and directly jump to the next dialog on long-tap on map. (This would also eliminate the, IMHO, ugly "menu-hacking" code.)

This can be done. 

> Why do you chose a manual "copy coordinates" way of implementation? I propose to change the action buttons in this dialog to "Cancel" and "Ok" when in this mode, close the dialog on cancel, but automatically transfer the selected coordinates back into the coordinate edit dialog on "ok", without having to manually copy&paste the coords.

Originally I also wanted to automatically transfer the coordinates. But now it is implemented like this: After clicking the "Select from map" button, a new activity with the map is started. After confirming (copying) the selected coordinates, the activity is ended and returned back. I did not know how to check if it returns (so that the coordinates are automatically inserted) without polling. If you can help me or advise me how to implement it as you wrote, I will implement it.

moving-bits:
> > Why do you chose a manual "copy coordinates" way of implementation? I propose to change the action buttons in this dialog to "Cancel" and "Ok" when in this mode, close the dialog on cancel, but automatically transfer the selected coordinates back into the coordinate edit dialog on "ok", without having to manually copy&paste the coords.
> 
> Originally I also wanted to automatically transfer the coordinates. But now it is implemented like this: After clicking the "Select from map" button, a new activity with the map is started. After confirming (copying) the selected coordinates, the activity is ended and returned back. I did not know how to check if it returns (so that the coordinates are automatically inserted) without polling. If you can help me or advise me how to implement it as you wrote, I will implement it.

This can be achieved using `startActivityForResult()` instead of `startActivity()`, and retrieving the result with `onActivityResult()` in the calling activity. You need to define a specific request code to assign and retrieve the correct result. See `CacheListActivity.startListSelection()` for an example, using `REQUEST_CODE_IMPORT_PQ` as request code.

(There are also more modern ways of achieving this, using "Activity Result APIs", see https://developer.android.com/training/basics/intents/result, but we have not migrated c:geo to that approach. It would be up to you to choose whether to go with the old way, finding code examples for it in c:geo, or using the new way, but having to do a bit more of implementation work initially.)

Dam297:
Added new fixed version

moving-bits (approved):
LGTM, thanks @Dam297 for this contribution.

Small notice: Concept of starting a UnifiedMapActivity is slightly different (should have mentioned this earlier, sorry) - instead of adding additional parameters that determine how the map should behave, if parameter is present, we use a different UnifiedMapType and select actions based on this. - But this is no blocker, and I can fix it in a separate step.

PR #18012, opened 2026-04-13
Trigger update after language switch

### LINKED ISSUE #9556: Language switch incomplete or failing in logging activity
Originally posted by @MagpieFourtyTwo in https://github.com/cgeo/cgeo/issues/9532#issuecomment-747509253:

Current situation: Android set to "de", c:geo set to "en", which seemed to work pretty well over the last two days.

While caching this morning (thus have opened c:geo now and then to navigate to the [P], walk to the coords, look for position and stuff) I finally found and logged the cache, saved the log offline and happily went away ... while walking back to [P] s/th came to my mind which should be mentioned in the log, so I switched on and tried to reopen the log for edit.  But on short click on the text I just got the popup you get when clicking somebody else's log, which has of course no entry for "edit". Strange. Only way to edit it was, to again click on ``Log Visit``.

Then I recognized that the log was titled ``Dein gespeichertes Log`` - which did not really match my language setting, which was still set to ``en``, as I cross checked right after. After resetting it to ``de`` and afterwards to ``en`` again, the log was properly titled with ``Your saved log`` and even the edit worked again as usual.

Conclusion: There seem to be some scenarios, where the language switch does not properly resp. not *fully* work, as my log title was in Android's language (de), while e. g. the ``Details`` page and c:geo's main screen where still in c:geo's language (en).

Wild guess: The decision, if after tap on a log either the menu popup appears or edit will start, depends on a string compare of the log title ... and of course fails in cases of Babylonian language confusion?  ;)

### LINKED ISSUE #18006: Wrong language for routing-profile-name after switching language
> image
> 
> only that the description line (marked in red) showed "(keine)" two times.
> 
> "(keine)" would be the equivalent to "no user-selected routing file", in which case the symbols "1" and "2" (and their description) should not be shown at all. I guess it's some timing issue when switching language, string "(keine)" compared to some translation, therefore being unequal, therefore buttons + description being shown. But if my theory is correct (haven't yet checked in code), this would be a (minor) bug in map quick settings, not in this PR.
> 

Correct, this is already an existing behavior. 
Will look into it separately

_Originally posted by @murggel in https://github.com/cgeo/cgeo/issues/17871#issuecomment-4187721118_

murggel:
## Description

This PR fixes the remaining issues from language-switch:
* recreate Settings-Activity to apply language directly
* removes (now) unnecessary res-parameter

## Related issues

fix #9556

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActivity.java
@@ -1,15 +1,14 @@
 package cgeo.geocaching.activity;
 
-import cgeo.geocaching.CgeoApplication;
 import cgeo.geocaching.R;
 import cgeo.geocaching.models.CalculatedCoordinate;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Waypoint;
 import cgeo.geocaching.service.GeocacheChangedBroadcastReceiver;
+import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.utils.ActionBarUtils;
-import cgeo.geocaching.utils.ApplicationSettings;
 import cgeo.geocaching.utils.EditUtils;
 import cgeo.geocaching.utils.LifecycleAwareBroadcastReceiver;
 import cgeo.geocaching.utils.LocalizationUtils;
@@ -20,7 +19,7 @@ import cgeo.geocaching.utils.html.HtmlUtils;
 
 import android.content.Context;
 import android.content.Intent;
-import android.content.res.Resources;
+import android.content.res.Configuration;
 import android.os.Bundle;
 import android.util.AndroidRuntimeException;
 import android.util.Pair;
@@ -50,8 +49,6 @@ import io.reactivex.rxjava3.disposables.Disposable;
 
 public abstract class AbstractActivity extends AppCompatActivity implements IAbstractActivity {
 
-    protected CgeoApplication app = null;
-    protected Resources res = null;
     private final CompositeDisposable resumeDisposable = new CompositeDisposable();
 
     private final String logToken = "[" + this.getClass().getName() + "]";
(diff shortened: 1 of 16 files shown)


main/src/main/java/cgeo/geocaching/activity/AbstractActivity.java
```diff
@@ -125,10 +124,21 @@ public void invalidateOptionsMenuCompatible() {
         ActivityMixin.invalidateOptionsMenu(this);
     }
 
+    @Override
+    protected void attachBaseContext(final Context newBase) {
```
murggel, line 128: `setLocale` before calling `onCreate`. This will ensure, that the correct language will be taken for the activity. This makes the settings-preference translated after language switch

moving-bits:
There is some strange effect on recreating the activity, an additional space appearing at the top:

|settings before language switch|settings after language switch|stays on returning to main screen|
|---|---|---|
|image|image|image|

This is very likely unrelated to the PR itself, but gets triggered by recreation of activity.

murggel:
> There is some strange effect on recreating the activity, an additional space appearing at the top:
> This is very likely unrelated to the PR itself, but gets triggered by recreation of activity.

Unfortunately I can't reproduce it in the emulator -  @moving-bits  on which device / did that occur?

moving-bits:
> Unfortunately I can't reproduce it in the emulator - @moving-bits on which device / did that occur?

Some emulated Pixel device, but can't say which model and API level out of my head. I can retest with a physical device, but not today.

moving-bits:
Have not been able to reproduce it yet, so maybe just give it a try?

murggel:
> Have not been able to reproduce it yet, so maybe just give it a try?

@moving-bits  Wasn't able to reproduce it yet or didn't you find the time to reproduce?

moving-bits:
> > Have not been able to reproduce it yet, so maybe just give it a try?
> 
> @moving-bits Wasn't able to reproduce it yet or didn't you find the time to reproduce?

Both 😄 - I have retried on two emulated devices, but was not able to reproduce it there. But I have several development environments (Linux/Windows + different computers) and I have not checked them all (and can't remember which device I was working on when I made the screenshots above).

murggel:
> > > Have not been able to reproduce it yet, so maybe just give it a try?
> > 
> > @moving-bits Wasn't able to reproduce it yet or didn't you find the time to reproduce?
> 
> Both 😄 - I have retried on two emulated devices, but was not able to reproduce it there. But I have several development environments (Linux/Windows + different computers) and I have not checked them all (and can't remember which device I was working on when I made the screenshots above).

Then I would suggest to try it in the nightly and adapt if you found the failing one...

PR #18015, opened 2026-04-14
Initialize imperial units setting on clean install

### LINKED ISSUE #18014: clean install is using imperial units
Just did a clean install on a second device and routing distance info on map was in feet. The "use imperial units" setting was disabled. Toggling it twice fixed it. Seems the default is wrong somewhere.

utafrali:
## Description
On clean installs the imperial units preference wasn't getting initialized during the migration. This left it unset, causing the default to fall back to imperial. Added the initialization so it respects the locale default on first boot.

## Related issues
Fixes #18014

## Additional context
N/A

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -307,6 +307,7 @@ public class Settings {
         final SharedPreferences prefsV0 = CgeoApplication.getInstance().getSharedPreferences(preferencesNameV0, Context.MODE_PRIVATE);
         if (currentVersion == 0 && prefsV0.getAll().isEmpty()) {
             final Editor e = sharedPrefs.edit();
+            e.putBoolean(getKey(R.string.pref_units_imperial), useImperialUnitsByDefault());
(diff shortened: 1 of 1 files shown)


ztNFny:
If we take this approach (initializing the value on first start then all other references to useImperialUnitsByDefault() should be removed as they'd never be called.

utafrali:
Done, pushed the fix.

moving-bits (approved):
According to `Settings.useImperialUnitsByDefault()` it returns `true` for `Locale.getDefault().getCountry()` returning `US`, `LR` or `MM`.
Preference layout `preferences_appearance.xml` defaults to `false` for this setting

Therefore startup behavior would be:
- on fresh install, no settings exist yet, so `migrateSettings()` is mostly skipped
- when you open preferences first, settings key `units` does not exist, so the default from layout file is taken (`false`) for initializing the checkbox
- but as soon as you open a map (and are using one of the locales from above), a value of `true` gets returned by `useImperialUnits()` (but not written to preferences)

Therefore persisting the identified default locale on first initialization LGTM.

moving-bits:
Thanks, @utafrali, for your contribution, very welcomed!

PR #18019, opened 2026-04-15
Save log after image is selected

### LINKED ISSUE #18017: adding a user image doesn't store the cache
Usuall edit-actions on cache trigger a save of the cache, but for adding images this doesn't happen and they might get lost easily on data cleanup.

utafrali:
## Description
When adding an image to a cache log, the imageListFragment processes the selection but the log wasn't being saved. Added saveLog() in onActivityResult to persist the changes.

## Related issues
Fixes #18017

## Additional context
N/A

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
@@ -503,6 +503,7 @@ public class LogCacheActivity extends AbstractLoggingActivity implements LoaderM
     protected void onActivityResult(final int requestCode, final int resultCode, final Intent data) {
         super.onActivityResult(requestCode, resultCode, data);  // call super to make lint happy
         imageListFragment.onParentActivityResult(requestCode, resultCode, data);
+        saveLog();
     }
 
     @Override


eddiemuc:
It is wantes behavior that the log is not saved on returning from image selection. This PR would change this wanted behavior.

 Imagine the user starts editing an offline log, changes maybe the text, then adds an image. But then he cancela the edit. Then the old text must remain. With this PR the old text would be gone in this workflow.

In case of "cancel" a previously added log inage must he deleted in order to not waste space. This logic should be in place already. (Haven't checked in current code state though)

murggel:
In which cases does the log and its images get not saved correctly? 

AFAIU In the mentioned issue the cache should be saved on adding a image to the cache, no log is affected.

eddiemuc:
Let me reformulate using an exemplary situation:

Prerequisite:
* A cache has an existing offline log with:
  * Logtext: TFTC
  * Type: "note"
  * Two attached images A and B with captions "Image A" and "Image B" and both in portrait mode

Steps:
* User opens the offline log for edit
* User changes Logtext to "Thanks a LOT for this cache!"
* User changes Logtype to "Found"
* User deleted image A
* User changes orientation of image B to "landscape" and changes caption text to "Beautiful Image"
* User adds a new image C with caption "A second beautiful image"
* IMPORTANT: User now hits "CANCEL" (does not want to save the log changes

Question: what should now be the situation? 

My assumption is:
* Log text should still be "TFTC" (state from before)
* Log type should still be "note" (state from before)
* Log should still have two images A and B attached, both still in portrait mode and with captions "Image A" and "Image B"
* On disk there should be the images A and B stored, both in portrait mode. Image C should NOT be stored

Current code in master should be such that my assumption is true.

If this PR us merged then I assume the situation would be:
* Log text is now "Thanks a LOT for this cache!" 
* Log type is now "found" 
* Log has now two images B and C attached, A is gone. B is in landscape and has caption "Beautiful image"
* On disk three images are stored (A, B and C)

murggel:
And AFAIU the mentioned issue in the PR handles images of a cache and not images of a log

moving-bits:
To my understanding, this PR is about adding a user-defined image to a cache's gallery:
- open cache
- swipe to gallery
- select "add image"
- chose an image from image file picker
- close cache listing => issue #18017: image does not get saved

Does this interaction (adding a user-defined image to cache gallery) interfere with the procedure for adding a log image, causing this PR to have side-effects?

eddiemuc:
> To my understanding, this PR is about adding a user-defined image to a cache's gallery

Although that might be the topic of zhe referenced issue, this PR changes log (image) handling. It even says so in the title

moving-bits:
> > To my understanding, this PR is about adding a user-defined image to a cache's gallery
> 
> Although that might be the topic of zhe referenced issue, this PR changes log (image) handling. It even says so in the title

Sorry, my fault - was on a mobile and had not looked up the code change.

So I agree, current code change does not resolve the original issue, therefore adding WIP label.

@utafrali 
Please update this PR to actually fix the original issue, without side-effects on logging. Thanks.

moving-bits:
Superseded by #18071, therefore closing here

PR #18021, opened 2026-04-16
Fix #15519: configure length of image caption and description

### LINKED ISSUE #15519: Log image description: restrict to length of 125 characters
Support ticket 475964

When logging, c:geo allows unrestricted length for image description while GC.com only allows 125 chara

When trying to send the log it fails. I reduced the number of characters in my text and it sent without a problem. I think c:geo should also restrict this to 125 characters so that the send doesn’t fail!

murggel:
## Description

Adapt max length of image caption and description depending on connector:
|Connector|caption|description|
| --- | --- | --- |
| GC | 50 | 125 |
|Other| 50 | 250|

## Related issues

Fix #15519


moving-bits (approved):
LGTM

PR #18025, opened 2026-04-19
limit height of waypoint infosheet (fix #17991)

### LINKED ISSUE #17991: Waypoint quicklook "slide up" too far
### Describe your problem!

(example) single adventure questions on the map. Tap the 🧩. The quick view slides in. Now, please try to compass, route, navigate or use the 3-dot menu to compass, route or navigate ("other", "drive", "walk", "bike") to that specific location. I can't, because the menu bar slides up too high.

When selecting a multi(-waypoint), the quick view stops in time to use the menu. 

### How to reproduce?

Adventure slide looks like so:
Image 

But it should look like this multi:

Image

Only on adventures it seems impossible to use the menu to navigate to a next waypoint? 

Any and all waypoints should habe the option to navigate or use compass. 

### Reproducible

Yes

### c:geo Version

2026.04.04-NB-917db64

### System information

```text

```

### Additional Information

_No response_

moving-bits:
## Description
This PR fixes #17991 by limiting the maximum height of the cache/waypoint info sheets.

To not confuse swipe handling for (possibly long) waypoint popups, the "swipe up to open cache" action gets restricted to cache popups only (as suggested by @fm-sys).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java
@@ -18,6 +18,7 @@ import android.content.Intent;
 import android.content.res.Resources;
 import android.os.Handler;
 import android.os.Looper;
+import android.util.DisplayMetrics;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.ViewTreeObserver;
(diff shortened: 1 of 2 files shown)


fm-sys:
Hi @moving-bits,
It's been quite a while since I was working with the c:geo codebase and especially with bottom sheets. I see several easy possibilities:
- Truncate the length of the text field
- Remove the swipe up action for waypoints (it's unclear anyway whether the user wants to see the corresponding cache or waypoint)


Right now, the bottom sheet consists of two fragments. While swiping up, the content get's faded out and the `SwipeToOpenFragment` gets visible:

https://github.com/cgeo/cgeo/blob/243b673f54a0662156ff10369234a82284453c05/main/src/main/java/cgeo/geocaching/activity/AbstractNavigationBarMapActivity.java#L118-L135

Using a (Nested-)ScrollView is probably the way to go, as you already tried. IIRC there are some methods to influence whether the scrollview or the parent should be scrolled. We would like to want a behavior like "first scroll locally, then the parent". 

For normal scroll view, you can do something like this (e.g. only disallow intercept when not fully scrolled to the bottom):

```java
scrollView.setOnTouchListener((v, event) -> {
    v.getParent().requestDisallowInterceptTouchEvent(true);
    return false;
});
```

Unfortunately I don't have the time to look further into it right now. (honestly this looks like something one could spend hours on. Maybe there is a quick and easy fix, maybe rewriting the whole thing altogether would have been faster at the end... 🤷‍♀️)

moving-bits:
Thanks, @fm-sys, for the pointer. I've found a working solution:
- Removed the "swipe up to open cache" behavior for waypoint popups
- Switch `ScrollView` to `NestedScrollView` for waypoint popups (to enable scroll content down as well)

PR #18043, opened 2026-04-24
Fix #18006 use correct localization for list entry <none> / <new>

### LINKED ISSUE #18006: Wrong language for routing-profile-name after switching language

murggel:
## Description

* Don't use localized "none" as key for routing-profil
* Localize list-entry "new" on demand

## Related issues

Fix #18006

### DIFF
--- a/main/src/main/java/cgeo/geocaching/list/AbstractList.java
+++ b/main/src/main/java/cgeo/geocaching/list/AbstractList.java
@@ -4,18 +4,26 @@ import android.util.SparseArray;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 
 public abstract class AbstractList {
 
     public final int id;
     @NonNull
     public final String title;
+    @StringRes
+    protected final int titleResourceId;
     public final int markerId;
     private static final SparseArray<AbstractList> LISTS = new SparseArray<>();
 
-    public AbstractList(final int id, @NonNull final String title, @NonNull final int markerId) {
+    public AbstractList(final int id, @NonNull final String title, final int markerId) {
+        this(id, title, 0, markerId);
+    }
+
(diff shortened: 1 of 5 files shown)


moving-bits (approved):
Have done a short smoke test with switching several profiles, LGTM

moving-bits:
retest this please

murggel:
> Have done a short smoke test with switching several profiles, LGTM

Can you also check an update-scenario?

moving-bits:
> > Have done a short smoke test with switching several profiles, LGTM
> 
> Can you also check an update-scenario?

Seems to work as well. My test case was:
- c:geo set to English
- set profile 2 to "(none)"
- switch c:geo to German
- set profile 1 to "(keine)"

Settings and map will look like this then:

|Settings|Map|
|---|---|
|image|image|

(Map quick settings should not show the "2" button, as it is basically the same as "(keine)".)

Then I installed this PR and opened settings + map - both look fine now: Settings shows both entries in localized "(keine)", and map quick settings no longer display the "2" button.

On switching language to "Spanish", settings show both entries as "(ninguno)", and map quick settings still do not show the "2" button.

LGTM

moving-bits (approved):

PR #18060, opened 2026-05-03
Update built-in BRouter engine to v1.7.9

moving-bits:
## Description
- Updates built-in BRouter engine to v1.7.9
- Prints BRouter version string in credentials


bekuno:
The reverse geocoder test failure is unrelated.
The results by google are defective. I will create a PR to correct this the next days.

bekuno:
> The reverse geocoder test failure is unrelated. The results by google are defective. I will create a PR to correct this the next days.

Fixed by #18066 (target to release)

Issue #17499, opened 2025-09-28
Manual login for Captcha resolving is missing OK button


Lineflyer:
### Describe your problem!

While abroad I first time had the case, that the login failed due to Captcha requirement.

I tried the "Manual loing" button, which lead me to a web browser window which for the web login part worked but had problems:
1. The text on top sayed I should tap the "OK" button after I logged in, but did not show any OK button
2. The view does not seem fit for edge-to-edge, as the top of it was shown under my system status bar

Maybe 1. + 2. have the same root cause....I cannot judge as I did not use that feature before

Did some screenshots but can only add them after my trip.

### How to reproduce?

See above

### Reproducible

Unclear

### c:geo Version

2025.09.26-NB

### System information

```text
Samsung S24, Android 15
```

Lineflyer:
Assigning myself to add screenshots

moving-bits:
PR #17508 should fix this

ztNFny:
> PR [#17508](https://github.com/cgeo/cgeo/pull/17508) should fix this

@moving-bits any idea why this was never merged to release? stumbled upn that as part of https://github.com/cgeo/cgeo/pull/18069

moving-bits:
> > PR [#17508](https://github.com/cgeo/cgeo/pull/17508) should fix this
> 
> [@moving-bits](https://github.com/moving-bits) any idea why this was never merged to release? stumbled upn that as part of [#18069](https://github.com/cgeo/cgeo/pull/18069)

Probably because all changes for edge2edge are on `master` only, not part of `release`

ztNFny:
> > > PR [#17508](https://github.com/cgeo/cgeo/pull/17508) should fix this
> > 
> > 
> > [@moving-bits](https://github.com/moving-bits) any idea why this was never merged to release? stumbled upn that as part of [#18069](https://github.com/cgeo/cgeo/pull/18069)
> 
> Probably because all changes for edge2edge are on `master` only, not part of `release`

oops, is release really that out of date? sorry, didn't think to check that as it just seemed so unplausible... that's what you get for always using master ...

moving-bits:
yes, it's probably the longest period in c:geo's history without a new feature release, thanks to "edge2edge"...

PR #18064, opened 2026-05-05
Wherigo login fix (master)

ztNFny:


ztNFny:
https://github.com/cgeo/cgeo/pull/18069 is targeted against release, that's probably the better target

ztNFny:
keeping this PR against master as there are bigger differences in the base branches (master vs release) that'd cause problems (see https://github.com/cgeo/cgeo/issues/17499#issuecomment-4410031049), this one already has the good state against master

ztNFny:
retest this please

moving-bits:
retest this please

murggel:
I tested it succesfully in master and release.
As you stated, sometimes you have to do a second try, because the login wasn't fast enough. 


Also it was not really obvious, what I have to click and confirm. Maybe change to "Please wait ..., **follow the instructions to** log into your account..."

ztNFny:
The text was reused from GC login. Maybe make it clearer by adding "complete the login process until you see the Wherigo dashboard"?

Timeout due to slowness isn't something i experienced, do you have a scenario?

murggel:
> The text was reused from GC login. Maybe make it clearer by adding "complete the login process until you see the Wherigo dashboard"?
> 

If you reused it, changing to something which mentions wherigo is probably not possible, unless you create a new text, which is in my opinion not necessary. Don't know which page will be shown for GC login...

Maybe something "and complete the login process"?

> Timeout due to slowness isn't something i experienced, do you have a scenario?

Not reproducible... I am not sure whether it was slowness or ... but trying to download a second time was successful without another login-procedure. 
So I think we can ignore that

ztNFny:
Changing the text wouldn't be a problem, i'll look into it later. (And if i don't find the time then it's at least not a blocker for this PR)

ztNFny:
Text updated

moving-bits:
retest this please

PR #18066, opened 2026-05-06
Update testAndroidGeocoder

bekuno:
The test lookup result is now incorrect.
So, let's use another address (it is a nearby street).


bekuno:
retest this please

PR #18071, opened 2026-05-08
ensure cache is stored before adding a user image, fixes #18017

### LINKED ISSUE #18017: adding a user image doesn't store the cache

ztNFny:
ensure cache is stored on clicking the "add image" buttons in gallery. This will also store the cache in case adding the image is canceled, but the same behavior exists for adding a log, so should be ok


murggel:
Should we then also delete the images when removing the cache from the device (when deleting orphaned data)?

PR #18086, opened 2026-05-11
Rel to #17842: Refine edge-to-edge inset handling

baiti:
# Analysis: Edge-to-Edge UI Overlap Bug

## Problem Description
In `CacheDetailActivity` (personal note), `LogCacheActivity`, and `EditWaypointActivity`, the editable text area is vertically enlarged downwards into the keyboard area when more text is entered than fits. This makes the text unreachable.

## Root Cause
The issue is caused by a conflict in how Window Insets (specifically the IME/Keyboard insets) are handled across the application.

1. **Global Handling in `AbstractActivity`**:
   The base class `AbstractActivity` applies padding to the view with ID `activity_content`. It uses `WindowInsetsCompat.Type.ime()`, meaning it automatically adds bottom padding to the root container when the keyboard appears.

2. **Surgical Handling via `ViewUtils.preventKeyboardOverlap`**:
   Commit `ae60e494b` (PR #17915) introduced `ViewUtils.preventKeyboardOverlap`, which adds *another* `OnApplyWindowInsetsListener` to the specific `EditText` views. This listener also adds padding based on the IME insets.

3. **Double Padding**:
   When both are active, the layout receives double the necessary padding. In a `ScrollView` (used in `LogCacheActivity` and `EditWaypointActivity`), the root is already padded by the activity, and then the internal view is padded again. This "pushes" the view incorrectly or prevents the `ScrollView` from correctly calculating its scrollable area, leading to the "disappearing text" behavior.

4. **Fullscreen Dialogs**:
   `EditNoteDialog` (used for personal notes in `CacheDetailActivity`) inherits from `AbstractFullscreenDialog`, which has its own `applyEdge2Edge` logic that manually sets padding on the root view. If this view contains an `EditText` that also tries to handle insets, similar conflicts occur.

## Current Status (as of May 11, 2026 - Iteration 3)
- **Branch**: `fix/edge-to-edge-overlap`
- **What Works**:
    - **Personal Note**: Works perfectly. It uses `AbstractFullscreenDialog` which applies padding to its root view in `applyEdge2Edge`.
- **What Partially Works**:
    - **Logging & Waypoint Notes**: The "double-padding" conflict is gone. With `clipToPadding="false"`, the text can now be scrolled up into view using cursor keys, but it still initially extends underneath the keyboard. This indicates the `ScrollView` is not automatically resizing its viewport to match the keyboard's top edge.
- **Implemented Changes**:
    - Removed redundant `ViewUtils.preventKeyboardOverlap` calls.
    - Added `adjustResize` to Manifest for Logging/Waypoints.
    - Added `clipToPadding="false"` to `ScrollView`s in Logging/Waypoints.
    - Reverted `WindowInsetsCompat.CONSUMED` to avoid "floating keyboard" issue.

## Refined Strategy
The success of the **Personal Note** is the key. It uses a `DialogFragment` that manually applies system bar + IME insets as padding to its entire view.
`AbstractActivity` tries to do something similar by padding `activity_content`, but `activity_content` is often nested inside a `activity_content_wrapper` (FrameLayout) which is inside a `LinearLayout` with a `Toolbar`.

**Next Steps**:
1.  **Compare View Hierarchies**: Analyze why `AbstractFullscreenDialog`'s padding results in a resize/scroll effect that works, while `AbstractActivity`'s padding on `activity_content` only results in a padded scroll area that requires manual scrolling.
2.  **Align AbstractActivity**: Potentially move the inset listener in `AbstractActivity` to a higher-level view or adjust how the `ScrollView` reacts to the padding to ensure the focused `EditText` is always kept above the keyboard automatically.

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -241,7 +241,7 @@
         <activity
             android:name=".EditWaypointActivity"
             android:label="@string/waypoint_edit_title"
-            android:windowSoftInputMode="stateHidden" >
+            android:windowSoftInputMode="stateHidden|adjustResize" >
         </activity>
         <activity
             android:name=".NavigateAnyPointActivity"
@@ -454,7 +454,8 @@
             />
         <activity
             android:name=".log.LogCacheActivity"
-            android:label="@string/log_new_log" >
+            android:label="@string/log_new_log"
+            android:windowSoftInputMode="adjustResize" >
         </activity>
         <activity
             android:name=".log.LogTrackableActivity"
--- a/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
+++ b/main/src/main/java/cgeo/geocaching/EditWaypointActivity.java
@@ -285,8 +285,6 @@ public class EditWaypointActivity extends AbstractActionBarActivity implements C
             binding.noteLayout.setVisibility(View.GONE);
             updateCoordinates(preprojectedCoords);
         }
-
-        ViewUtils.preventKeyboardOverlap(binding.userNote);
     }
 
     private void setCoordsModificationVisibility(final IConnector con) {
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActivity.java
@@ -166,15 +166,10 @@ public abstract class AbstractActivity extends AppCompatActivity implements IAbs
         windowInsetsController.setSystemBarsBehavior(WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE);
         //apply edge2edge to activity content view
         ViewCompat.setOnApplyWindowInsetsListener(currentWindow.getDecorView(), (v, windowInsets) -> {
-            final View activityContent = v.findViewById(R.id.activity_content);
-            if (activityContent == null) {
-                Log.w("edge2edge: activityContent not found in " + this);
-            } else {
-                //calculate and set the activity_content's insets
-                this.currentWindowInsets = windowInsets.getInsets(DEFAULT_INSETS);
-                //trigger insets recalculation
-                refreshActivityContentInsets();
-            }
+            //calculate and set the activity_content's insets
+            this.currentWindowInsets = windowInsets.getInsets(DEFAULT_INSETS);
+            //trigger insets recalculation
+            refreshActivityContentInsets();
             return windowInsets;
         });
 
@@ -188,16 +183,22 @@ public abstract class AbstractActivity extends AppCompatActivity implements IAbs
             //method was called before insets were set
             return;
         }
-        final View activityContent = getWindow() == null || getWindow().getDecorView() == null ? null :
-            getWindow().getDecorView().findViewById(R.id.activity_content);
-        if (activityContent == null) {
+        final View decorView = getWindow() == null ? null : getWindow().getDecorView();
+        if (decorView == null) {
+            return;
+        }
+        final View activityContent = decorView.findViewById(R.id.activity_content);
+        final View activityWrapper = decorView.findViewById(R.id.activity_content_wrapper);
+        final View root = activityWrapper != null ? activityWrapper : activityContent;
+
+        if (root == null) {
             return;
         }
 
         //let subclasses modify insets according to their needs
         final Insets insets = calculateInsetsForActivityContent(this.currentWindowInsets);
         //apply final insets to activity content
-        activityContent.setPadding(
+        root.setPadding(
                 insets.left < 0 ? this.currentWindowInsets.left : insets.left,
                 insets.top < 0 ? this.currentWindowInsets.top : insets.top,
                 insets.right < 0 ? this.currentWindowInsets.right : insets.right,
--- a/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogCacheActivity.java
@@ -256,7 +256,6 @@ public class LogCacheActivity extends AbstractLoggingActivity implements LoaderM
             }
         }
         inventoryAdapter.putActions(lastSavedState.inventoryActions);
-        ViewUtils.preventKeyboardOverlap(binding.log);
         refreshGui();
 
         requestKeyboardForLogging();
--- a/main/src/main/java/cgeo/geocaching/ui/ViewUtils.java
+++ b/main/src/main/java/cgeo/geocaching/ui/ViewUtils.java
@@ -67,11 +67,8 @@ import androidx.annotation.StyleRes;
 import androidx.annotation.StyleableRes;
 import androidx.appcompat.widget.TooltipCompat;
 import androidx.core.content.res.ResourcesCompat;
-import androidx.core.graphics.Insets;
 import androidx.core.graphics.drawable.DrawableCompat;
 import androidx.core.text.util.LinkifyCompat;
-import androidx.core.view.ViewCompat;
-import androidx.core.view.WindowInsetsCompat;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -898,14 +895,6 @@ public class ViewUtils {
         final Drawable circularIcon = IndeterminateDrawable.createCircularDrawable(button.getContext(), spec);
 
         return enable -> mButton.setIcon(enable ? circularIcon : originalIcon);
-    }
-
-    public static void preventKeyboardOverlap(final View view) {
-        ViewCompat.setOnApplyWindowInsetsListener(view, (v, windowInsets) -> {
-            final Insets newInsets = windowInsets.getInsets(WindowInsetsCompat.Type.ime() | WindowInsetsCompat.Type.systemBars());
-            v.setPadding(v.getPaddingLeft(), v.getPaddingTop(), v.getPaddingRight(), newInsets.bottom);
-            return windowInsets;
-        });
     }
 
 }
--- a/main/src/main/res/layout/editwaypoint_activity.xml
+++ b/main/src/main/res/layout/editwaypoint_activity.xml
@@ -9,18 +9,6 @@
     android:orientation="vertical"
     tools:context=".EditWaypointActivity">
 
-    <!-- Coordinates -->
-
-    <Button
-        android:id="@+id/buttonLatLongitude"
-        style="@style/button_full_double"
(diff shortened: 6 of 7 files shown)


baiti:
Made some additional changes:
   1. Layout Refactor: logcache_activity.xml and editwaypoint_activity.xml now use a LinearLayout with a fixed header (Type/Date/Coords) and a weighted ScrollView. This ensures the core
      action buttons stay pinned to the Toolbar while the text area resizes.
   2. Inset Logic: AbstractActivity now applies padding to the activity_content_wrapper exactly once, resolving the double-padding conflict.
   3. Manifest: Explicitly set adjustResize for the affected activities to ensure standard keyboard interaction.
   4. Consolidated Fix: AbstractFullscreenDialog also now correctly consumes insets to maintain consistency across dialogs and activities.

Perhaps they way it is working now is in fact acceptable, can even be released in the upcoming beta?

baiti:
Perhaps the following discussion w/ Gemini helps:

**baiti:**
 > from how you understand the UI/UX now: Is this something we could ship? Do you understand why personal cache notes behaves different, in a way that the frame never grows out of view?

**Gemini:**
In its current state, I believe it is much safer to ship than the original bug, as the critical "unreachable text" issue (overlap) is resolved. However, from a senior UI/UX perspective, it isn't "perfect" yet because of the difference in how the text areas handle growth.

  Why Personal Note behaves differently (and better):
  The Personal Note dialog uses a "Rigid Frame" layout:
   * Technically: It has no root ScrollView. The TextInputLayout is given a layout_weight="1".
   * The Result: When the keyboard appears, the dialog view is padded, and the TextInputLayout is forced to shrink to fit the exact space remaining between the top toolbar and the keyboard. It cannot grow "higher"; it is physically constrained. Any extra text is handled by the EditText's own internal scrollbar.

  Why Logging and Waypoints behave "elastic":
  These activities use an "Elastic Viewport" layout:
   * Technically: They have many fields, so they must have a root ScrollView to allow the user to reach the bottom fields (Inventory, Waypoint visited, etc.).
   * The Result: Because the EditText is inside a ScrollView, the ScrollView allows its content to be as tall as it wants to be. As you type more lines, the EditText grows. The ScrollView then tries to "keep up" with the cursor by scrolling down. This is why it feels like the text area is growing "higher and higher" and pushing the top of your message out of the viewport.

  Can we ship it?
  Yes, but with a note to the team.
   * Pros: It's a functional fix. The user can always reach their text and see what they are typing. The "floating keyboard" and "hidden text" bugs are gone.
   * Cons: It doesn't feel as "contained" as the Personal Note dialog.

  How to achieve the "Personal Note" behavior in Activities:
  To get the exact same behavior, we would have to:
   1. Give the TextInputLayout a fixed maximum height (e.g., 200dp) so it stops growing at some point and starts scrolling internally.
   2. However, this can feel clunky in a long form like Waypoints.

  Final Verdict: I would ship this branch for review. It moves the needle from "Broken" to "Functional but Elastic." The team can then decide if they want to apply specific maxHeight constraints to those individual fields to mimic the dialog's rigidity.

moving-bits:
I'm not at my PC currently, therefore cannot test the PR, but wanted to share some piece of info:

There is a fundamental difference between "personal cache note" (PCN) and the other two, which explains, why PCN edit is working easily, whereas the other are not: PCN edit is a full-screen view, which just needs to set the inset for edge2edge mode, no further scrolling involved within that page. In contrast to this, the other two fields are part of a longer list of elements, because of which the whole page is made scrollable. As the edit field can grow/shrink, Android might not know where to apply the scrolling gesture to, and how to apply the insets to (and in which direction to grow/shrink the input field) - this is a much more difficult seting than PCN edit.

Before throwing an AI to it, you may want to apply a `git revert` of my PR #17915 to the source for a "fresh" start without my only-partially working attempt of fixing the issue.

baiti:
> 
> Before throwing an AI to it, you may want to apply a `git revert` of my PR #17915 to the source for a "fresh" start without my only-partially working attempt of fixing the issue.

# Re: reverting PR #17915

PR #17915 made exactly three changes:

1. Added `ViewUtils.preventKeyboardOverlap()` in `ViewUtils.java`
2. Called it on `binding.userNote` in `EditWaypointActivity`
3. Called it on `binding.log` in `LogCacheActivity`

All three have now been addressed: the call-sites (2 and 3) were removed as part of the layout restructuring, and the now-dead `preventKeyboardOverlap()` method itself (1) has since been removed along with its orphaned imports.

**A `git revert` of #17915 is therefore not needed** — the branch already goes well beyond a plain revert. It restructures both `logcache_activity.xml` and `editwaypoint_activity.xml` and refines inset handling in `AbstractActivity`, which is the actual fix rather than the padding-listener approach from #17915.

From testing, this appears to be working correctly and should be considered the final fix for the issue. Please review when you get to it.

moving-bits:
Sorry, @baiti: On re-reading my last post I noticed it sounded not the way I had intended it to. Generally I just wanted to give some context why PCN vs. the other two places behave differently. My last sentence was NOT meant to be an accusation at all, but just a note that removing my PR might give AI a better (cleaner) start. I had not recognized at that time, that AI had already removed all parts of it in this PR. Sorry again!

moving-bits (commented):
I found time today to do a bit of smoke testing. Some observations:
- The two places look fine now, in general, when starting with empty text fields or just short texts - text area will grow then, keyboard stays below.
- It does not yet fully work when reentering a text field that has a long waypoint note/log text in it already - text field will be partially hidden beneath keyboard and will not resize. But I regard this as a minor issue, since the page as well as the text fields content is scrollable, thus the end of the text is reachable.
- I do not understand why the first line of buttons gets fixed. They are not special, IMHO - the important buttons are part of the toolbar.

As for the other activities / dialogs I have opened only some of them, but no full cross-check for unwanted side-effects that might have been introduced by adapting the insets logic. No problems came up so far, though.

All tests done on a physical device running Android 15.

baiti:
> Sorry, @baiti: On re-reading my last post I noticed it sounded not the way I had intended it to. Generally I just wanted to give some context why PCN vs. the other two places behave differently. My last sentence was NOT meant to be an accusation at all, but just a note that removing my PR might give AI a better (cleaner) start. I had not recognized at that time, that AI had already removed all parts of it in this PR. Sorry again!

No offense taken at all! I had actually started with reverting that PR, I worked a precise `task.md` file and wanted analysis from Google's Gemini to see what it can do. Later I reviewed Gemini's work with Claude and ultimately came up with that PR as it is now. It may not be the perfect and final solution but I would say it is at least not gating the upcoming Beta, or in other words, it is usable the way it is now, would you agree? Feel free to improve that PR as you see fit. You can also check out that branch, fix it, close my PR and create a new one.

eddiemuc:
I reviewed the PR throughout and I think the approach and execution is fine. I vote for merging this PR and observe in nightlies. It is definitely an improvement. See below for two findings which might need discussion.

(Overall AI claims that our edge-2-edge-approach is not the cleanest, but to clean this up would be a very big thing in redesigning UI so I wouldn't approach this path right nowe and rather try to get master shipped into beta fast)

**Finding 1: PR introduces fixed buttons on top. Is this intended/wanted?**
The PR wraps editwaypintview and logview into a new LinearLayout with scrollview being a child of it to get edge-2-edge under control. That approach seems fine. However, it choses NOT to include some top-buttons into the scrollview, which leads to those buttons being "fixed" to the top of screen when scrolling. This is the coordinate button in editwaypointview and the type/date-buttons on logview. Example screenshot below. --> is this intended and accepted?

image


**Finding 2: Codebase could be simplified by removing R.id.activity_content_wrapper in favor of activity_content** 
The PR correctly points out (and fixes) that usage of R.id.activity_content_wrapper in toolbar instead of R.id.activity_content was not handled. PR decided to add handling of activity_content_wrapper also in AbstractActivity. But could it be an alternative to get rid of activity_content_wrapper altogether? Seems this is only used in Toolbar 

@moving-bits: is this needed/wanted for some reason?

eddiemuc:
Update: I edited my previous response heavily

moving-bits:
Thanks @eddiemuc for the review.
- I would prefer not to have those buttons fixed - they are in no way "special", IMHO, the relevant buttons are part of the toolbar.
- The `activity_content_wrapper` was named as such, as `AbstractActionBarActivity` creates a full wrapper around an existing layout if it could not find the toolbar. The current page´s layout is put into the container named `activity_content_wrapper`. This avoids possible name clashes if the current page already has `activity_content` defined - but I'm not sure if our program flow actually allows that at all. => So we might rename that container to `activity_content` and test-drive it in the nightlies.

eddiemuc:
> Thanks @eddiemuc for the review.
> 
> * I would prefer not to have those buttons fixed - they are in no way "special", IMHO, the relevant buttons are part of the toolbar.
> * The `activity_content_wrapper` was named as such, as `AbstractActionBarActivity` creates a full wrapper around an existing layout if it could not find the toolbar. The current page´s layout is put into the container named `activity_content_wrapper`. This avoids possible name clashes if the current page already has `activity_content` defined - but I'm not sure if our program flow actually allows that at all. => So we might rename that container to `activity_content` and test-drive it in the nightlies.

@baiti should I tackle this and commit to your branch at baiti:fix/edge-to-edge-overlap or would you prefer to handle this yourself? Another approach would be to merge your PR and immediately start a second one to tackle this

baiti:
> @baiti should I tackle this and commit to your branch at baiti:fix/edge-to-edge-overlap or would you prefer to handle this yourself? Another approach would be to merge your PR and immediately start a second one to tackle this

pick the workflow that is easiest for you and does the job. Feel free to commit into my branch. Didn't know that this is even possible but it is absolutely fine if this works ;)

eddiemuc:
> > @baiti should I tackle this and commit to your branch at baiti:fix/edge-to-edge-overlap or would you prefer to handle this yourself? Another approach would be to merge your PR and immediately start a second one to tackle this
> 
> pick the workflow that is easiest for you and does the job. Feel free to commit into my branch. Didn't know that this is even possible but it is absolutely fine if this works ;)

I just pushed a second commit to your branch. If PR quality check fails now, that's on me :-)

I changed layout as wanted, but I left the activity_content_wrapper as is. I think this is ok even for prod.

eddiemuc:
CI seems very broken on this one...

eddiemuc:
Looks green when you go to codacy server and default is green. 
Ok to merge?

moving-bits:
> Ok to merge?

Short smoke test looks good, let's try it in the nightlies

PR #18091, opened 2026-05-12
Different Wherigo-icons in CacheDetail and Popup

murggel:
## Description

After removing the external wherigo-player, we have now 2 different icons for starting the internal wherigo-player
|CacheDetail (colored) | Popup-Fragment (black/white)|
| --- | --- |
|grafik|grafik|

Which one should we use? I am not sure, how we decided...


bekuno:
I would stay with b/w icon on both places (as it is in the released version for the internal WIG player).
It looks more consistent to the other icons.
Or, use a blue symbol with tracparent background (the blue needs then to be compatible to the light/dark designs).
Maybe reorder the icons in the popup. I propose: hint - note - player. This is more logic from the cache point of view starting with the cache (More details) button.
Btw., the player icon in the light design has white foreground (also in the released c:geo app).

moving-bits:
As c:geo's getting a bit more colorful meanwhile, I would prefer the colored version, actually. It fits nicely with the Wherigo's cache symbol and the type filter symbol for Wherigos.

We do have a colored button for lab adventures as well (which also starts a separate player, though not so tightly integrated in this case).

But in the end, it's mostly a matter of taste.

bekuno:
Let's use a poll:
- thumbs up      = blue forground
- thumbs down = blue background
- hooray            = black & white

murggel:
> Btw., the player icon in the light design has white foreground (also in the released c:geo app).

That's where I came from, wanted to adapt the style and saw the difference

murggel:
> Let's use a poll:
> 
> As c:geo's getting a bit more colorful meanwhile, I would prefer the colored version, actually. It fits nicely with the Wherigo's cache symbol and the type filter symbol for Wherigos.


We have more places where the icon is shown (in the NavigationBar, the ActionBar,...) - if we want to use the colored version we should probably use it there as well...

murggel:
> Maybe reorder the icons in the popup. I propose: hint - note - player. This is more logic from the cache point of view starting with the cache (More details) button.

I can provide some screenshots - I am not sure if we really should change the order - at least I am used to have the hint at the most right and the note at the most left side - and the optional stuff in between...

grafik

murggel:
> We do have a colored button for lab adventures as well (which also starts a separate player, though not so tightly integrated in this case).
> 

And that is, what for me the color stands for -> external tool, so I vote for the b/w

murggel:
There is the chirp-button missing in the popup

grafik

 - shall I add it?
 - 
grafik

murggel:
Ready for review - I added the chirp.
The wherigo is only possible for fully loaded caches , since the guid is searched inside the description

bekuno:
ok to test

main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
```diff
@@ -0,0 +1,268 @@
...
+                Settings.setLastDisplayedList(listId);
+                if (cacheDetailActivity != null) {
+                    cacheDetailActivity.setNeedsRefresh();
+                }
+                CacheListActivity.startActivityOffline(view.getContext());
+            }
+        }, start, builder.length(), Spanned.SPAN_EXCLUSIVE_EXCLUSIVE);
+    }
+
+    public static void updateWherigoBox(final Geocache cache, final Activity activity, @Nullable final View wherigoView, @Nullable final TextView wherigoText, @Nullable final Button wherigoButton) {
```
moving-bits: You cold make the last parameter @NonNull and save one null check inside, as all callers are setting a valid layout element here. (This is applicable for all three `updateXXXBox` methods)

main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
```diff
@@ -0,0 +1,268 @@
...
+        final List<String> wherigoGuis = WherigoUtils.getWherigoGuids(cache);
+        final boolean isEnabled = !wherigoGuis.isEmpty();
+        if (null != wherigoView) {
```
moving-bits: You could shorten this by calling `ViewUtils.setVisibility(wherigoView, isEnabled ? View.VISIBLE : View.GONE)`. (Applicable for all three `updateXXXBox` methods)

moving-bits (approved):
Did a quick smoke test with save and unsaved Wherigos & Labs, works as expected. Just have two minor coding style suggestions (non-blocking).

murggel:
> Ready for review - I added the chirp. The wherigo is only possible for fully loaded caches , since the guid is searched inside the description

We can show the button always for wherigos and fetch the data on demand, but I would do that in a separate PR

main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
murggel: Done, and I changed the order

main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
murggel: I changed ViewUtils.setText to using TextParam.id().applyTo(view)

moving-bits:
@murggel 
PR has merge conflicts meanwhile 🤷, can you resolve it, please?

murggel:
> @murggel 
> PR has merge conflicts meanwhile 🤷, can you resolve it, please?

I am on a hiking trip, will have a look on sunday or monday

murggel:
> > @murggel
> > PR has merge conflicts meanwhile 🤷, can you resolve it, please?
> 
> I am on a hiking trip, will have a look on sunday or monday

@moving-bits  resolved

moving-bits:
Thanks for resolving, let's get it in the nightlies then

ztNFny:
@murggel is there a reason you skipped the call to CacheInfoBoxes.updateAttributes ? Otherwise I'd add that as part of https://github.com/cgeo/cgeo/pull/18430 (need to touch a lot of code around it and while doing that stumbled upon that unused method)

murggel:
> @murggel is there a reason you skipped the call to CacheInfoBoxes.updateAttributes ? Otherwise I'd add that as part of #18430 (need to touch a lot of code around it and while doing that stumbled upon that unused method)

No - I guess that was due to revoling of merge-conflicts - the identical method in CacheDetailActivity.DetailsViewCreator.updateAttribues is used instead - so you can remove that method and call instead the one of CacheInfoBoxes.
Or if you want - I can provide a PR for that right now...

ztNFny:
Already done - i'd prefer to include it in mine to avoid conflicts.

Issue #18092, opened 2026-05-13
List (Parent) Names: Colons get multiplied (and some more glitches)


MagpieFourtyTwo:
### Describe your problem!

While saving a cache offline using ``(+) <Create New List>`` and in the process selecting an already existing ``Parent List``, much to my surprise nevertheless a new ``Parent List`` was created, with just the new list in it.

The strange thing:  The new list name contained not just one. but *two* colons, like in ``Parent::List``.

Maybe the colon is not removed from the found Parent names, as even the items of the ``Parent list``'s dropdown still have the separating colon at the end - which was not the case in former times, IIRC. And then another colon is added in the process of generating the new list's name ...?

BTW, there are two more glitches in the ``Parent list``'s dropdown which perhaps can be corrected in the same go:
1. When, in the process of renaming a list prefix, you open the ``Parent list``'s dropdown (assumed you have more than 5 parent lists), the lower parts of the dropdown list go behind the keyboard.  You now have to close the keyboard first (which additionally closes the dropdown) and re-open the dropdown list to get access to the lower items (this has already been corrected for the list creation dialog, but over here it happens again).
2. The dropdown list contains two entries, which do the same: ``<none>`` and ``      `` (i. e. an empty line) - IMHO one of these can safely be removed, as they give the absolut same result.

However, what makes this double-colon-bug so nasty is, that it's kind of hard to get rid of the second colon. In order to do so, you ...
1. have to enter a list
2. select ``Menu`` => ``Manage lists`` => ``Rename list prefix``
3. choose the one with doubled colon 
4. in ``Rename to`` you must remove *JUST ONE* colon (i. e. leave one seemingly unlogical colon at the end)
5. then press [OK]
6. and press another [OK] for the ``Rename list prefix`` warning dialog

After that, the list is still displayed as ``Parent:List`` in the Header, but once you open the list selection dropdown, you see that it is correctly assigned now.

[...]

PR #18094, opened 2026-05-14
Fix some list naming issues (rel. to #18092)

moving-bits:
## Description
Fixes many (but not all) things mentioned in #18092:
- removes leading/trailing spaces and colons on saving list name
- removes multiple consecutive colons on saving list name
- (same for setting new list prefix)
- adds unit tests for list naming handling
- closes keyboard on leaving "list prefix to" field
- returns list of list prefixes without trailing colons

## Additional context
Untouched by this PR is to enable "OK" button on just changing selected parent list.


moving-bits:
"Could not create ADB bridge" again, I've notified @mucek4...

moving-bits:
retest this please

PR #18097, opened 2026-05-16
Emulatorhelper: helper script to start/stop/restart/fix-dns emulators

eddiemuc:
Emulatorhelper: helper script to start/stop/restart/fix-dns emulators

In case it is wanted: helper commands to deal with local emulators (and especially the damn networking problems I always have)

source it, then call "emu_help" for instructions


eddiemuc:
failing ci is unrelated

PR #16616, opened 2025-02-02
POC GPX multi parser

moving-bits:
## Description
This is for discussion only, no real PR yet.

We have quite a few issues dealing with GPX import. A major issue is the missing handling of multiple items of different types within a single GPX file, eg. a list of geocaches together with a track or route.

This PR attempts to draft a possible solution:
- parse geocaches, tracks and routes in parallel
- depending on the result, offer multiple options
  - single cache: import to list and open cache
  - list of caches: import to list and open list
  - track: add to map's track display
  - route: replace individual route (after confirmation)
- some of those steps may be automated depending on context, eg: when importing from within a list, import to this list and stay in this list

So basically:
- combine all of today's GPX parsers into a single one
- add some user interface to handle multiple inputs

Code is a very basic draft only, just identifying the opening tags, but no handling of attributes etc. yet - it's for demonstration purposes only.

Code gets triggered when tapping on a GPX file in the device's file manager and selecting c:geo to handle it.
(All c:geo-internal GPX import functions are still routed to the existing code.)

Question to the team: Do we want to pursue further into that direction?

## Related issues
This would deal at least with the following issues: #14329, #15364, #16187, #15827, #8626, #6710

## Additional context
see also the related [project for GPX-related issues](https://github.com/orgs/cgeo/projects/6/views/2)

PR #18096, opened 2026-05-16
Copilot: suggested instruction improvements

eddiemuc:
Copilot: suggested instruction improvements

PR #18105, opened 2026-05-16
Experiment: WIP UnifiedGPXParser

moving-bits:
## Description
I've played a bit with Claude today and chose the "UnifiedGPXParser" as a project (see #16616 for a manual approach).

I started with some general explanations (different types of GPX, expected return format etc.), then let it generate the first parsing framework, then dived into track parsing + tests for it, and finally steered Claude into direction of Geocache parsing.

This draft is the current result, completely untested yet, not reviewed in detail, not adjusted manually at all. Very likely much functionality is missing yet, and there is currently no integration into c:geo at all - the parser is standalone currently and nowhere integrated. (All of this is reserved for later steps.)

I just wanted to share this preliminary state for some early feedback - is this approach reasonable enough to pursue further?

## Related issues
Several, are linked in the earlier approach of #16616.

## Additional context
This is an early draft, not even a WIP yet.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:github.com)",
+      "WebFetch(domain:raw.githubusercontent.com)",
+      "Bash(findstr -i \"route track rte trk\")"
+    ]
+  }
+}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,9 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "WebFetch(domain:github.com)",
-      "WebFetch(domain:raw.githubusercontent.com)",
-      "Bash(findstr -i \"route track rte trk\")"
-    ]
-  }
-}
--- a/.gitignore
+++ b/.gitignore
@@ -49,4 +49,7 @@ main/gradle.properties
 # ignore AS layout captures
 captures/
 
+# ignore Claude internal files
+.claude/
+
 main/usage.txt
--- a/main/src/main/java/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/main/java/cgeo/geocaching/files/GPXParser.java
@@ -182,7 +182,7 @@ abstract class GPXParser extends FileParser {
         version = versionIn;
     }
 
-    static Date parseDate(final String inputUntrimmed) throws ParseException {
+    public static Date parseDate(final String inputUntrimmed) throws ParseException {
         // remove milliseconds to reduce number of needed patterns
         final MatcherWrapper matcher = new MatcherWrapper(PATTERN_MILLISECONDS, inputUntrimmed.trim());
         final String input = matcher.replaceFirst("");
--- a/main/src/main/java/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/main/java/cgeo/geocaching/files/GPXParser.java
@@ -182,7 +182,7 @@ abstract class GPXParser extends FileParser {
         version = versionIn;
     }
 
-    public static Date parseDate(final String inputUntrimmed) throws ParseException {
+    static Date parseDate(final String inputUntrimmed) throws ParseException {
         // remove milliseconds to reduce number of needed patterns
         final MatcherWrapper matcher = new MatcherWrapper(PATTERN_MILLISECONDS, inputUntrimmed.trim());
         final String input = matcher.replaceFirst("");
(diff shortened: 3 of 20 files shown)


eddiemuc:
I absolutley love it! 

Few notes:
* All classes containing results (eg Result, OrphanedWaypoint etc) should be made its own classes in own files (not parser-subclasses) because they will be used around other parts of c:geo
* The "Track" and "Waypoint" Subparser-classes in turn will to my understanding NOT be used outside of the parser (correct?) So they should be made inaccessible
* you checked in your claude local settings, that was probably not intended
* According to GPX-spec, both Routes (rteType/rtept) and Tracks (trktype/trksegtype/trkpt) contain Waypoints (wptType). "Translated" to our use case this means they could be full-blown geocaches or geocache-waypoints. Although this is not a common use case, a generic parser should parse these things as full-blown waypoints first. Conversion to Route/Track should then be done depending on the use case (helper might be part of Result object though). See https://www.topografix.com/GPX/1/1

A note to claude: I find if (!) you can use the superior Claude models (specifically Opus as of now) in Copilot there is no notable difference any more between Claude and GitHub Copilot AI quality. That being said, I am aware that Opus is not accessible to us through the usual Pro/Pro+ models :-(

moving-bits:
Thanks for your feedback. I had explicitly instructed Claude to create a stand-alone parser for the time being and leave integration for later, as I did not want to mangle our existing classes in this early stage yet, but maybe we are coming closer to interlocking the code.

I'll try to find some time throughout the week to continue work on this. Maybe we can collect some instructions here that I should feed into the AI for improving the PR? Feel free to edit the list below and add/update/remove instructions.

- [ ] Please move the UnifiedGPXParser classes into a separate folder "unifiedgpxparser", subfolder of "files". Each class should get their own file. Make parser-internal classes package-private, but extract classes required for integration into c:geo into their own files.

Additionally I should update `CLAUDE.md` to better adapt to our coding styles etc. 🤷

eddiemuc:
Everything fine here, I am excited on how well your experiment goes :-)

Some thoughts:

**Agent instruction structure**
As I feel we aim for a multi-AI-supporting framework right now (at least claude and Github copilot), I suggest to change the structure of AI-instructions a bit:
* move the full instruction set from (now) .github/copilot-instructions.md to (new) AGENTS.md in root of project
* create two files  .github/copilot-instructions.md and CLAUDE.md which however are just thin wrappers pointing to AGENTS.md

wrappers can be easily generated using AI :-) If you want to go this way, please kindly consider my changes in https://github.com/cgeo/cgeo/pull/18096 first

**Requirements for Agents**
you want to collect instructions to feed into the AI for improving the PR.
My experience (mainly in work where I have free access to higher-level models) is that this can best be done by AI itself :-)

For example, if you agree with my suggestions you can ask something like the following:

```
Here is a set of improvements I would like to make on the parser:
* ...A
* ...B
* ... C (eg copy my set here or have own requirements)
Please analyze these requests, ask any questions you might have on them and suggest improvements or changes to them. Then write a revised structured requirement document to file xy.md
```

I get amazingly good results with this approach. Often the AI asks me things that I would never have thought of myself.

moving-bits:
> Please analyze these requests, ask any questions you might have on them and suggest improvements or changes to them. Then write a revised structured requirement document to file xy.md

This might be something we put in `AGENTS.md` to establish this as a general way of interaction.

eddiemuc:
> > Please analyze these requests, ask any questions you might have on them and suggest improvements or changes to them. Then write a revised structured requirement document to file xy.md
> 
> This might be something we put in `AGENTS.md` to establish this as a general way of interaction.

Well I have it in my personal (local) agent preferences :-)

eddiemuc:
@moving-bits do you want to continue working on this PR solely, or would it be fine (for me) to add on top of it? (No promises :-) I might have some time next couple of days)

moving-bits:
> @moving-bits do you want to continue working on this PR solely, or would it be fine (for me) to add on top of it? (No promises :-) I might have some time next couple of days)

Would be very welcomed if you want to continue work on this 👍 - haven't found myself time yet for further looks into it after having created it (with AI).

We could merge it into `master`, as the code is completely separated yet (on purpose), so no existing functionality should be broken by it. Integration into today's code is probably last step, as we will have quite different workflows then. We might also think about having it in parallel for a while, with a config switch to select one or the other.

eddiemuc:
Merge would be fine for me

moving-bits:
> Merge would be fine for me

Ok, let's merge and continue work on `master` then

eddiemuc:
I might (...) have time this week to work on the gpx topic again. Can't promise anything unfortunately :-(

@moving-bits would that be ok for you?

moving-bits:
> I might (...) have time this week to work on the gpx topic again. Can't promise anything unfortunately :-(
> 
> @moving-bits would that be ok for you?

Sure, please go ahead, very welcomed!
Feel free to continue on top of this PR, tear half of it down or restart from scratch, whatever is more suitable.

PR #18119, opened 2026-05-20
Fine-tuning for "select coords from map" (rel. to #18011)

moving-bits:
## Description
- Add UMTT_SelectCoords as new UnifiedMapType
- On selecting coords from map, use current GPS position for new waypoints, and waypoint's position otherwise


moving-bits:
retest this please

Issue #17610, opened 2025-12-10
[Nightly] APK file size doubled


MagpieFourtyTwo:
### Describe your problem!

Since around December 7th, the size of the downloaded APK file has nearly doubled from it's already impressive 42 MB to a whopping size of ~80 MB. How come?

### How to reproduce?

Just have a look at the file behind https://download.cgeo.org/cgeo-nightly.apk ... ;)

### Actual result after these steps?

Download size of 78 MB (10.12.25) and up to 82,14 MB (08.12.25).

### Expected result after these steps?

Download of 41 (or even better the former 14) MB.  ;)

### Reproducible

Yes

### c:geo Version

2025.12.10-NB-91b50dd

### System information

```text

```

bekuno:
This happened after increasing minSdk:
<img width="829" height="306" alt="Image" src="https://github.com/user-attachments/assets/3fcbbb61-bd2a-414f-b991-53b875f61296" />
Before:
<img width="747" height="178" alt="Image" src="https://github.com/user-attachments/assets/b1d42c1f-0407-4632-bafb-29fefbaed7df" />

bekuno:
The diff of both apk shows:
<img width="627" height="328" alt="Image" src="https://github.com/user-attachments/assets/fa741726-25d9-4293-8c23-95f6417b59c7" />

I found this:
https://stackoverflow.com/questions/68186329/changing-minsdkversion-from-16-to-26-increased-release-apk-size-from-17-to-39-mb with reference to https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs .
In short: Native libraries (*.so) are now uncompressed in the *apk file. This enables faster installation and smaller incremental downloads from the Google Play Store.
We can configure compression as before the minSDK change for non release apk.

MagpieFourtyTwo:
Sounds good - esp. as incremental updates are not an option for nightlies anyway.

bekuno:
@MagpieFourtyTwo  Should be fixed in the next nightly.

MagpieFourtyTwo:
Thx!

MagpieFourtyTwo:
APK size suddenly increased again from 44,58 MB on May, 18th (2026.05.18-B-6c3d951) to 61,98 MB (2026.05.19-NB-f7c0b6e).
For any obvious reason?

SammysHP:
Probably the new offline translation we're trying as an alternative to the previous one.

MagpieFourtyTwo:
[...]

PR #18122, opened 2026-05-22
Remove text split for HTML in offline translation (rel. to #17610)

moving-bits:
## Description
- Removes text splitting by HTML segments (see lines 161-164), as it's no longer needed when using Bergamot translator
- Additionally removes some unused methods in `TranslatorUtils.java`

## Examples for cache GC6W5KE
|French (original)|French -> English|French -> German|
|---|---|---|
|image|image|image|


SammysHP:
This changes the behavior for both Google/mlkit and Bergamot? That would make make it hard to compare with the previous mlkit translation.

moving-bits:
> This changes the behavior for both Google/mlkit and Bergamot?

Yes

> That would make make it hard to compare with the previous mlkit translation.

You would need two different installations anyway, as Bergamot has a higher precendence than ML Kit, so for texts that Bergamot translates, you won't ever see a ML Kit translation in the same version.

Thus either install current release (ML Kit only) + Developer version (Bergamot) on the same device, or current release (ML Kit only) + Nightly or Developer version (Bergamot) on a separate device, if you want to compare it. (Or use emulator to force either Bergamot or ML Kit translation.)

eddiemuc:
As long as we haven't drxided what happens to mlkit, i propose a different approach:
- add method "boolean supportsNativeHtml()" to ITranslatorAccessor interface
- implement it to return true for bergamot and false for mlkit
- use method during translation to determine how to handle html

Alternatively, remove mlkit completely from code already

moving-bits:
Reasoning behind it was to have a small & quick change to check how Bergamot translation quality changes if we remove the HTML split. But I agree, it would be more consistent to remove ML Kit translation completely, as it is "hidden" by Bergamot in most cases anyway. And nothing that a `git revert` couldn't fix easily, should we decide to stay with ML Kit instead if Bergamot.

I have adapted the PR accordingly.

Issue #4752, opened 2015-03-21
Detecting own opencaching caches fails when not logged in


rsudev:
To reproduce:
- have one of your own opencaching caches stored offline on some list
- switch of all network
- ensure you are logged out (kill c:geo)
- open a map of the list
result:
the cache has the disc symbol (stored) not the star symbol (own)

if you keep connected, the star is shown instead

Lineflyer:
For me it seems like saved own OC-caches are never marked as own.
I have a list of my owned caches and the marker is missing independently of network is active or not.

rsudev:
Does the main screen show a succesful login to oc.de in this case?
Caches are generally not marked as own. Ownership is detected dynamically by compring your user name and the real owner name.

Natureshadow:
I see owned OC caches not being recognised as owned even when I am usccessfully logged in.

PR #18125, opened 2026-05-23
Fix filtering for own caches

### LINKED ISSUE #6628: OC caches not detected as "owned" until fully loaded
Not sure if this is the same issue as #4752 or related, but OC caches in search results (and probably map as well) are not detected as "owned". The status is detected correctly after the cache details were loaded.

murggel:
## Description

* Use StatusGeocacheFilter "Own" for searching for own caches from SearchActivity.
* Improve StatusGeocacheFilter "Own" to handle multiple connectors correctly.
* For Okapi-Search without geo-/distance-information, don't limit the radius for the search, only the limit of returned caches

## Related issues

rel #6628 
rel #4752

## Additional context

Implementation correct behaviour for multi-connector-filter is then not so urgent any more, but nevertheless probably useful (PR #18002 , PR #17812

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/oc/OCApiLiveConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/oc/OCApiLiveConnector.java
@@ -160,7 +160,7 @@ public class OCApiLiveConnector extends OCApiConnector implements ISearchByViewP
     @Override
     public boolean isOwner(@NonNull final Geocache cache) {
         final String userName = getUserName();
-        return StringUtils.isNotEmpty(userName) && Strings.CS.equals(cache.getOwnerDisplayName(), userName);
+        return StringUtils.isNotEmpty(userName) && Strings.CI.equals(cache.getOwnerDisplayName(), userName);
     }
 
     @Override
--- a/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
+++ b/main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
@@ -364,25 +364,21 @@ final class OkapiClient {
 
         final List<BaseGeocacheFilter> filters = filter.getAndChainIfPossible();
 
-        final Geopoint searchCoords;
-        final String radius;
-
-        final DistanceGeocacheFilter df = GeocacheFilter.findInChain(filters, DistanceGeocacheFilter.class);
-        if (df != null) {
-            searchCoords = df.getEffectiveCoordinate();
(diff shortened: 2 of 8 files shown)


murggel:
I have to check, if I must add the STATUS-Filter to the SuConnector-capability now

main/src/main/java/cgeo/geocaching/connector/oc/OCApiLiveConnector.java
```diff
@@ -160,7 +160,7 @@ public boolean login() {
     @Override
     public boolean isOwner(@NonNull final Geocache cache) {
         final String userName = getUserName();
-        return StringUtils.isNotEmpty(userName) && Strings.CS.equals(cache.getOwnerDisplayName(), userName);
+        return StringUtils.isNotEmpty(userName) && Strings.CI.equals(cache.getOwnerDisplayName(), userName);
```
eddiemuc, line 163: What is CI vs CS?

main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
```diff
@@ -364,26 +364,22 @@ private static SearchResult retrieveCaches(@NonNull final OCApiConnector connect
...
+        final Parameters params = new Parameters("search_method", METHOD_SEARCH_ALL);
+        final Map<String, String> valueMap = new LinkedHashMap<>();
+        valueMap.put("limit", "" + take);
+        valueMap.put("offset", "" + skip);
 
         final DistanceGeocacheFilter df = GeocacheFilter.findInChain(filters, DistanceGeocacheFilter.class);
         if (df != null) {
-            searchCoords = df.getEffectiveCoordinate();
-            radius = df.getMaxRangeValue() == null ? DEFAULT_RADIUS : "" + (df.getMaxRangeValue().intValue() * 1000);
+            final String radius = df.getMaxRangeValue() == null ? DEFAULT_RADIUS : "" + (df.getMaxRangeValue().intValue() * 1000);
```
eddiemuc, line 375: Is this a side-fix for something else?

eddiemuc (approved):
Lgtm, questions are only out of interest

main/src/main/java/cgeo/geocaching/connector/oc/OCApiLiveConnector.java
murggel: > What is CI vs CS?

case-insensitive vs case-sensitive

main/src/main/java/cgeo/geocaching/connector/oc/OkapiClient.java
murggel: Yes and no. It's kind of a side fix, but otherwise only own caches within the radius would be returned

PR #18127, opened 2026-05-25
Only enqueue online caches in CacheDownloaderService (fix #18093)

### LINKED ISSUE #18093: "Update caches" should count only those from online connectors
_Originally posted by @dcjkfgdjhd in [#18087](https://github.com/cgeo/cgeo/issues/18087#issuecomment-4433155531):_

> When I create a list containing one cache from GC.com and one custom cache I created in c:geo itself,
> if I then select "update all caches",
> c:geo's notification popup says that it tries to update 2 caches (even though the custom cache obviously cannot be updated),
> and at the end it says "2 caches updated".
> 
> This looks like a bug. Custom caches should simply be ignored during the update process.

moving-bits:
## Description
- Only enqueues caches from online connector to `CacheDownloaderService`
- Shows a short notification if offline-only caches were requested

## Test case
- Create a list which has at least one user-defined cache
- Select "caches -> refresh all" for this list
- Before: Downloader notification shows one failing update
- Now: Short toast notification + correct system notification

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/IConnector.java
+++ b/main/src/main/java/cgeo/geocaching/connector/IConnector.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.connector;
 
 import cgeo.geocaching.connector.capability.ICredentials;
 import cgeo.geocaching.connector.capability.ILogin;
+import cgeo.geocaching.connector.capability.ISearchByGeocode;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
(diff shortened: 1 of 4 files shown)


eddiemuc:
Imho each connector needs to provide information whether it is "online" or "offline". Thus it is not a capability (like eg SeachByFilter) and doesn't need its own marker interface.

You could replace IOnline with a new method eg "boolean hasOnlineSource()" for IConnector with by default returning true, and override it in offline connectors with returning false.

moving-bits:
That's debatable, IMHO, the capability would be to be able to update specific cache data via online connection - is it that different from searching capabilities etc.? But I can live with a feature flag as well.

eddiemuc:
Tbh I am not a fan of marker interfaces in gener5. But thats a personal style preference and of course not mandatory for c:geo. Still I would argue that "online" is the default for connectors and "offline" is the exception.

Anyway, that's nitpicking. Of course I am fine with leaving everything as is.

moving-bits:
I've change it to a `hasOnlineSource()` approach which default to `true` (if connector supports `searchByGeocode()`)

Issue #16034, opened 2024-09-04
Show heart symbol on caches with many favorite points


eddiemuc:
Support Ticket #564792
Version 2024.09.01

A user on support requests that in c:geo caches with man favorite points are labelled with a heart symbol in the map. This feature is already present in the official groundspeak app:
![Screenshot_2024-09-04-16-51-51-14_7dcc949bcca61d8fb3afc2d82550efab.jpg](https://github.com/user-attachments/assets/a9572118-9285-4153-aa4a-f7962242f5c8)

ztNFny:
You can already achieve quite similar results using filters.

eddiemuc:
> You can already achieve quite similar results using filters.

Not really. You can filter for favorite points (but only in advanced mode), but you can't show both caches with and without high fav point count where the ones with a high fav count are marked.

You could use list symbols but this would require pre-filtering and storing the high fav caches upfront. This would not work on live map with not-stored caches.

Also all of the above would be something for experts only, not for the common user.

eddiemuc:
This could for example be solved by introducing dynamic cache symbols: user would be allowed to configure a symbol and a filter. For all caches where the filter is true, the symbol is shown. User can configure and store multiple such Symbol-Filter-combinations and activate/deactivate them. And we could supply some preconfigured symvol-filters eg for high fav count

ztNFny:
Turning a filter on and off shouldn't be a feature only advanced users can use 😉

Permanent filter icons is something we discussed before and I'd like that - I'd propose to just use the existing list icon places (those already handle multiple icons applied to the same cache)

murggel:
+1 From support [Ticket #663893](https://cgeosupport.droescher.eu/scp/tickets.php?id=10345) for showing heart symbol on map for favorite count > x

moving-bits:
To keep things simple, we could simply use one of the existing symbol positions to show a small heart symbol for caches having > x Favos. We could use top-center for it (giving it preference over "stored" symbol, if user has configured T/D-display on the map).

The threshhold "x" could be either fix (100?) or user-defined.

Is "number of favos" available for stored caches only, or also for live data?

murggel:
> The threshhold "x" could be either fix (100?) or user-defined.

I would say user-defined with 100(?) as default 

> 
> Is "number of favos" available for stored caches only, or also for live data?

It is shown in cache-popup, so it should be available

ztNFny:
[...]

Issue #12363, opened 2021-12-19
fully customizable cache icons/markers


ztNFny:
### Tell us your idea!

Cache icons display a lot of information already and new ideas for what to show are plenty (e.g. #7911 (Bettercacher) or #12321 (DT marker)
The importance of that information isn't universal but differs from user to user. We already have some options to cater to that (e.g. the "double-sized smiley" setting), but those are limited.

### What solution would you suggest?

A fully customizable cache icon where the user can decide himself, which information is shown where.
The idea is to display a page listing the 7 positions on the icon for markers (top-left, top-center, top-right, center-left, center-right, bottom-left, bottom-right) and for each being able to add elements to that section - similar to the filter dialog. Those sections can then be dragged & dropped to indicate the importance of the marker.

E.g. to implement the current "top-left" icon it'd be
- Owned
- Log status

Possible UI mockup
![image](https://user-images.githubusercontent.com/1258173/146685471-06d6651d-0cbc-4760-84b8-520068bc9081.png)

### Additional Information

Information that currently may be shown on the cache icon:
- center-center: cache type / double-sized smiley / custom cache icon (emoji)
- top-left: Owned / found / offline log
- top-middle: stored (if DT enabled)
- top-right: DT / stored
- center-left/right: list marker
- bottom-left: user note
- bottom-right: corrected coordinates / has final

fm-sys:
I'm not sure if is really necessary to make everything customizable. But anyway, this brings up the interesting idea of "conditional list markers" (markers, which will be displayed dynamically if a specific filter condition applies for the individual cache). This would have the benefit, that you could also mark caches which are not yet stored offline.

ztNFny:
now that we have BetterCacher support (#7911) it'd be nice to also have a marker for those - but I feel like we can't add one without the icon overflowing. So a BC marker implementation should go in hand with the options to customize markers

ztNFny:
> this brings up the interesting idea of "conditional list markers" (markers, which will be displayed dynamically if a specific filter condition applies for the individual cache). This would have the benefit, that you could also mark caches which are not yet stored offline.

That's an idea I also had - associating markers with filters (instead of just lists). I have many lists that exist only for the purpose of assigning markers to those caches (e.g. missing for D/T, high favorite percentage, ..) that could instead be done as a filter.

MagpieFourtyTwo:
[...]

PR #18133, opened 2026-05-26
rel to #12363 and #16034: introduce conditional cache markers

eddiemuc:
rel to #12363 and #16034: introduce conditional cache markers

This PR introduces the concept of conditional cache markers. With this:
* it implements a part of #12363 (not the part to configure what to display where though)
* it is a possible solution for #16034: users can configure "fav points" as conditional cache marker now
* it introduced a potentially powerful new concept to create more dynamic cache markers

Feature introduces a new settings "Conditional cache markers":
image

Upon tap the user may add/delete/edit rules containing a cache marker and a filter condition on which to apply it:
image

This filter shows in map as follows (note that the heart symbol is only on two of the three tradis since only two of them have >= 200 fav points):
image

Implementation note: for now, if any conditional marker applies to a cache, it is displayed in place of the "list markers" (left and right side of cache). Logic is now that conditional markers have precedence over list markers. Eg if a cache has 1 conditional marker and 2 list markers, then the conditional marker is displayed left and the first list marker right. If a cache has 2 conditional markers then those are displayed (no list markers). In screenshot above, note that there are two caches which are also part of a list with blue ribbon symbol (a tradi and a lab). Both also have one condition marker which is why the list symbol is displayed on the right.
--> this placement ruling can be discussed of course

### DIFF
--- a/main/AndroidManifest.xml
+++ b/main/AndroidManifest.xml
@@ -859,6 +859,11 @@
         </activity>
 
         <activity
+            android:name=".settings.ConditionalCacheMarkersActivity"
+            android:label="@string/conditional_marker_activity_title" >
+        </activity>
+
+        <activity
             android:name=".connector.trackable.GeokretyAuthorizationActivity"
             android:configChanges="keyboardHidden|orientation|screenSize"
             android:launchMode="singleTask"
--- a/main/src/main/java/cgeo/geocaching/models/ConditionalCacheMarker.java
+++ b/main/src/main/java/cgeo/geocaching/models/ConditionalCacheMarker.java
@@ -1,0 +1,97 @@
+package cgeo.geocaching.models;
+
+import cgeo.geocaching.filters.core.GeocacheFilter;
+import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.utils.EmojiUtils;
+import cgeo.geocaching.utils.JsonUtils;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Represents a single conditional cache marker rule: a filter paired with an emoji marker.
+ * When the filter matches a cache, the marker is shown on that cache's icon.
+ * This class is immutable — create a new instance to change values.
+ */
+public class ConditionalCacheMarker {
+
+    private static final String JSON_KEY_MARKER = "marker";
+    private static final String JSON_KEY_FILTER = "filter";
+
+    private final int markerId;
+    @Nullable private final GeocacheFilter filter;
+
+    public ConditionalCacheMarker(final int markerId, @Nullable final GeocacheFilter filter) {
+        this.markerId = markerId;
+        this.filter = filter;
+    }
+
+    public int getMarkerId() {
+        return markerId;
+    }
+
+    @Nullable
+    public GeocacheFilter getFilter() {
+        return filter;
+    }
+
+    /** Serialises this rule to a JSON object node. A null filter is stored as an absent field */
+    @NonNull
+    public ObjectNode toJson() {
+        final ObjectNode node = JsonUtils.createObjectNode();
+        JsonUtils.setInt(node, JSON_KEY_MARKER, markerId);
+        // null filter stored as absent field via JsonUtils.setText null-handling
+        JsonUtils.setText(node, JSON_KEY_FILTER, filter != null ? filter.toConfig() : null);
+        return node;
+    }
+
+    /** Deserialises a rule from a JSON node */
+    @NonNull
+    public static ConditionalCacheMarker fromJson(@NonNull final JsonNode node) {
+        final int markerId = JsonUtils.getInt(node, JSON_KEY_MARKER, EmojiUtils.NO_EMOJI);
+        final String filterConfigStr = JsonUtils.getText(node, JSON_KEY_FILTER, null);
+        final GeocacheFilter filter = filterConfigStr != null
+                ? GeocacheFilter.createFromConfig(filterConfigStr)
+                : null;
+        return new ConditionalCacheMarker(markerId, filter);
+    }
+
+    /**
+     * Returns the list of conditional marker IDs (emoji code-points) that apply to the given cache.
+     * Returns an empty list if cache is null.
+     */
+    @NonNull
+    public static List<Integer> getMarkersForCache(@Nullable final Geocache cache) {
+        return getMarkersForRules(cache, Settings.getConditionalCacheMarkers());
+    }
+
+    /**
+     * Package-private helper for testing: evaluates rules against a cache without reading from Settings.
+     * Rules are evaluated top-to-bottom. Rules with markerId == EmojiUtils.NO_EMOJI are skipped.
+     * A null or empty (non-filtering) filter matches all caches. Returns an empty list if cache is null.
+     */
+    @NonNull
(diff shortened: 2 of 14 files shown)


ztNFny:
cool, thanks for this implementation!

About the wording/accessibility: Maybe call it "overlay" instead of "marker"? Marker feels like the internal term, not sure users understand that. But then again we're also using it for like D/T overlays ... Not sure :-D

I'd rather have the UI in a place that allows for more dynamic changes (like the filter button in toolbar), but that's of course something to be discussed and could easily be extended upon this PR

eddiemuc:
> About the wording/accessibility: Maybe call it "overlay" instead of "marker"? Marker feels like the internal term, not sure users understand that. But then again we're also using it for like D/T overlays ... Not sure :-D

Also for "list marker":


Not for cachea though there it is called "cache icon". 
Maybe something to discuss separately

eddiemuc:
Let's tr it out in nightlies.

ztNFny:
my GUI idea would be like adding it in filter dialog:
image

and similar to the filter list one where you can select filters as overlays (might be integrated in the filter list in case that one supports some mode of switching between filter (single-select) and overlay (multi-select):
image

eddiemuc:
I am not following. Do you mean you want to make markers assignable only to name-saved filters? I find it a bit confusing to mix the general-purpose filter dialog with assigning markers. Also, how to handle priority of different conditional markers?

ztNFny:
using filters in a similar way to how filters are used (quick assignment in map/lists) would have the advantage of having it quite dynamic, allowing users to quickly select which overlays to use.
e.g. "i want to do a cache tour but along that tour highlight special caches"
And re-using filters would have the advantage that pretty much everything is already there.

whereas the current approach is for "constant" markers, like "100+ FPs"

eddiemuc:
Understood. What about adding a new option in map menu (maybe even in quick settings) to open conditional filters?

moving-bits:
I found the time to play a bit with the new conditional filters - cool feature, thanks for this addition. Two small remarks after my short test:
- We can skip the "map content & behavior - conditional cache markers" setting, IMHO. Calling this from the map is much more intuitive for adding/removing "on the fly".
- Feature request: How about adding the same marker coloring to cache lists as well?

ztNFny:
> - Feature request: How about adding the same marker coloring to cache lists as well?

Lists already support adding markers (list icon), that's where the inspiration for this came from. Unless i misunderstood you again?

eddiemuc:
> - We can skip the "map content & behavior - conditional cache markers" setting, IMHO. Calling this from the map is much more intuitive for adding/removing "on the fly".

I agree that adding it in map makes sense. Still, conditional markers are not map-specific but apply globally (everywhere where cache/waypoint icons are displayed eg also in stored lists or search results). This is why I think there should be a setting. Why the setting ia placed in "map content & behavior" is debatable. I placed it there because the D/T marker is there too, although it is also a global thingy. We can move both 3g to "Appearance"?

> - Feature request: How about adding the same marker coloring to cache lists as well?

Conditional markers should already be displayed on cache symbols in stored lists (haven't tested it there though). Or do you want to display a shortcut to conditional marker edit thete as well.

In case this isn't clear: conditional markers are currently implemented as a c:geo-global feature, not just for specific screens like map or stored lists or others.

eddiemuc:
You know, we could take this thing one step further and convert it into true "dynamic lists". That would be a "stored list", but instead of assigning caches to and from it the caches belonging to the list are defined by a filter. The dynamic list would show up in "stored list" selection (would only show offline caches there of course), and by assigning a list marker to such a list we again have conditional cache markers. Those markers would then also show up for online caches.

That would integrate conditional markera nicely with our list concept.

Probably something for later? :-)

moving-bits:
> > We can skip the "map content & behavior - conditional cache markers" setting, IMHO. Calling this from the map is much more intuitive for adding/removing "on the fly".

> I agree that adding it in map makes sense. Still, conditional markers are not map-specific but apply globally (everywhere where cache/waypoint icons are displayed eg also in stored lists or search results). This is why I think there should be a setting. Why the setting ia placed in "map content & behavior" is debatable. I placed it there because the D/T marker is there too, although it is also a global thingy. We can move both 3g to "Appearance"?

IMHO there is a difference between the two: "D/T marker on icons" very likely is a "set once, then forget" setting, whereas you might want to create or remove dynamic markers way more frequently, which is why having a direct access from current context (here: map menu) is pretty handy. But there's no need to duplicate that setting to settings as well. But... there is my feature request:

> > * Feature request: How about adding the same marker coloring to cache lists as well?
> 
> Conditional markers should already be displayed on cache symbols in stored lists (haven't tested it there though).

I guess I had selected the wrong list in my short test yesterday, so that my filter ("> 400 FP") did not catch. On retesting today, this works as expected.

Which leaves only the second part of this request:

> Or do you want to display a shortcut to conditional marker edit thete as well.

Yes, `CacheListActivity` should show the same shortcut menu entry that got introduced to map view. (And then we won't need the global setting anymore, IMHO).

moving-bits:
> > * Feature request: How about adding the same marker coloring to cache lists as well?
> 
> Lists already support adding markers (list icon), that's where the inspiration for this came from. Unless i misunderstood you again?

Yes, but that's a static assignment to the list, whereas the new feature is dynamically adding markers to caches across all lists when a certain condition is met. As such those two features complement each other. And, as I have just learned, it's already working within cache lists, just the quick access button in CacheListActivity's menu is missing.

How to handle if both dynamic marker and list markers are set? This could be a case for #12363, by which you could configure assigned positions and priorities of markers.

ztNFny:
> How to handle if both dynamic marker and list markers are set?

We already have that situation with caches in multiple lists, there are 2 marker slots, additional ones are dropped.

moving-bits:
> > How to handle if both dynamic marker and list markers are set?
> 
> We already have that situation with caches in multiple lists, there are 2 marker slots, additional ones are dropped.

What I meant: With #12363 we should make this configurable: Icontype lists per position should be have a priority top-down, and each marker that has already been displayed in another position is out of the loop for all other positions. Which means that positions itself need to get a priority/order.

eddiemuc:
> > > How to handle if both dynamic marker and list markers are set?
> > 
> > We already have that situation with caches in multiple lists, there are 2 marker slots, additional ones are dropped.
> 
> What I meant: With #12363 we should make this configurable: Icontype lists per position should be have a priority top-down, and each marker that has already been displayed in another position is out of the loop for all other positions. Which means that positions itself need to get a priority/order.

True, but not a new thing introduced with conditional markers. Already before that, imagine a cache is assigned to three lists with markers. Then one is assigned to left side (first priority), one to the right side (second priority) and one not at all (no more slots). Only thing I did is to add conditional markers BEFORE any list markers wrt priority in order of the conditional markers.

#12363 should probably allow to set slots with other prios as well (eg "Marker Prio 3" and "Marker Prio 4") for people who want more markers (conditional or list) on their caches.

eddiemuc:
> IMHO there is a difference between the two: "D/T marker on icons" very likely is a "set once, then forget" setting, whereas you might want to create or remove dynamic markers way more frequently, 

I think thats making assumptions. I for one could use a quick setting to turn on/off D/T. Contrary I don't see myself turning on/off conditional markers frequently tbh. I am still fine with adding shortcuts for other cachers use cases though...

> > > * Feature request: How about adding the same marker coloring to cache lists as 
> > Or do you want to display a shortcut to conditional marker edit thete as well.
> 
> Yes, `CacheListActivity` should show the same shortcut menu entry that got introduced to map view. (And then we won't need the global setting anymore, IMHO).

... but I don't see why not to display it in settings as well. Is this a great issue for you?

What about online search? You want a shortcut there as well?

moving-bits:
> ... but I don't see why not to display it in settings as well. Is this a great issue for you?

No, not a big issue, more a reflex: Because we have so many settings already, I'd like to avoid duplication between global settings and specific contexts, when possible.

PR #18130, opened 2026-05-26
ci: Fix nodejs deprecation warning

kumy:
## Description

Replace usage of GHA `actions-ecosystem/action-add-labels@v1` by `gh` cli call.

## Related issues

#18129

## Additional context

Code fix suggestion from https://github.com/actions-ecosystem/action-add-labels/issues/459#issuecomment-1906355780

PR #18136, opened 2026-05-26
DNM - only a test

bekuno:
Test for #18130.


bekuno:
@kumy Now all tests including "check-secrets" are skipped.

kumy:
Strange I will have a look

Issue #17479, opened 2025-09-21
Coloured text rendering as 'Black on black'


alan666notb:
### Describe your problem!

We've has this problem before, I thought it was fixed but it seems to be back.

Coloured text, in this case red text on a normal white background, now seems to rendering as black-on-black and is completely unreadable.

### How to reproduce?

1. Load GCBD4K7 - Church Micro 15134...Chipstead
2. Scroll down to the text just above the hint
3. View the cache page on a web browser and compare.

### Actual result after these steps?

The text in c:geo is unreadable - black-on-black

### Expected result after these steps?

Red text on a white background.
Or at least something readable.

### Reproducible

Yes

### c:geo Version

2025.08.31

### System information

```text

```

moving-bits:
I just tried with release 2025.08.31 and it works fine over here - text is shown as black on white background (dark theme). 

On further analysis, this seems to depend on (three dot menu) -> advanced -> change description style: If set to "Default w/o contrast fix", text is indeed black on black.

alan666notb:
I didn't know about that menu!

But it seems to work the other way for me...
I had it set to 'Default' - Black-on-black
When I set it to "Default w/o contrast fix", the text appeared.

moving-bits:
> But it seems to work the other way for me..

Are you using c:geo in light theme?
Then this seems to be working the other way around - seems to be an edge case w/r to contrast fix on/off.

alan666notb:
> > But it seems to work the other way for me..
> 
> Are you using c:geo in light theme? Then this seems to be working the other way around - seems to be an edge case w/r to contrast fix on/off.

Yes, I'm using light theme.
Does that setting have to be set individually for every cache?

eddiemuc:
> > > But it seems to work the other way for me..
> > 
> > Are you using c:geo in light theme? Then this seems to be working the other way around - seems to be an edge case w/r to contrast fix on/off.
> 
> Yes, I'm using light theme.
> Does that setting have to be set individually for every cache?

The setting is not persisted. It will go back to "Default" every time you (re)open  cache details.

The contrast algorithm seems to fail for this particular cache / text color. Has to be analyzed. 

We could additionally make "default w/o contrast fix" the new default setting when in light mode? After all, the contrast fix is primarily meant for dark mode (because background of c:geo is black while cache owners optimize their listing for white bg)

alan666notb:
> We could additionally make "default w/o contrast fix" the new default setting when in light mode?

[...]

PR #18139, opened 2026-05-27
Issue 17479 - change ordering of spans to have correct colours

Dam297:
## Description

Change and fixed ordering of spans in description of cache, to have text with same colors as on website

## Related issues

#17479


bekuno:
ok to test

PR #18145, opened 2026-05-28
Fix navigation interception in GeocacheFilterActivity with deferred ActivityMixin helpers

eddiemuc:
## Description

`GeocacheFilterActivity`'s confirmation dialog was non-functional: `onBackPressed()` called `super` unconditionally (closing immediately), the Toolbar ← arrow bypassed `onSupportNavigateUp()`, and `menu_item_cancel` skipped the dialog entirely.

- **`ActivityMixin`**: adds `registerBackNavigationInterceptor()` for `AppCompatActivity`, wires both `OnBackPressedDispatcher` and nav-up interception, passes deferred navigation `Runnable`s to resume intercepted back/up actions after confirmation, and provides `isActivityValid()` overloads for both `Activity` and `WeakReference`
- **Simplified interceptor API**: the navigation interceptor always handles both back press and nav-up, removing any boolean configuration and aligning the helper with its actual usage
- **WeakReference-safe continuation**: the returned navigation `Runnable`s resolve the activity through `WeakReference` before continuing navigation, avoiding leaks while keeping delayed confirmation handling simple
- **`GeocacheFilterActivity`**: replaces deprecated `onBackPressed()` handling with the new interceptor flow so back press, Toolbar ←, cancel menu item, and confirmed discard all route through the same unsaved-changes navigation path; the cancel menu item and home menu item both route through `onSupportNavigateUp()`
- **Confirmed navigation handling**: resumes the original back/up action by executing the deferred navigation callback from the confirmation dialog
- **Nav-up fallback behavior**: `registerBackNavigationInterceptor()` reports whether navigate-up was actually concluded, allowing `GeocacheFilterActivity.onSupportNavigateUp()` to fall back to `ActivityMixin.navigateUp(this)` and then `super.onSupportNavigateUp()` when the interceptor does not handle the action
- **No interference between `navigateUp` and interceptor**: `ActivityMixin.navigateUp()` is the conclude-navigation step invoked only after the interceptor has approved navigation; it does not call `onSupportNavigateUp()`, so there is no feedback loop or double-navigation risk

## Additional context

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
+++ b/main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
@@ -228,7 +228,8 @@ public final class ActivityMixin {
      *                    (e.g. from a dialog confirmation).
      *                    Return {@code true} to STOP navigation, {@code false} to CONTINUE.
      * @return A BooleanSupplier to be called from {@code onSupportNavigateUp()}. Returns {@code true}
-     *         when navigation was handled or intercepted, {@code false} if the activity is gone.
+     *         when navigation was handled or intercepted, {@code false} if the activity is gone or
+     *         navigate up could not be concluded.
      */
     public static BooleanSupplier registerBackNavigationInterceptor(
             @NonNull final AppCompatActivity activity,
@@ -267,16 +268,18 @@ public final class ActivityMixin {
             if (!isActivityValid(weakActivity)) {
                 return false;
             }
+            final boolean[] navigationHandled = new boolean[] {false};
             final Runnable continueNavigateUp = () -> {
                 final AppCompatActivity currentActivity = weakActivity.get();
                 if (isActivityValid(currentActivity)) {
-                    navigateUp(currentActivity);
+                    navigationHandled[0] = navigateUp(currentActivity);
                 }
             };
-            if (!interceptor.test(continueNavigateUp)) {
+            final boolean stop = interceptor.test(continueNavigateUp);
+            if (!stop) {
                 continueNavigateUp.run();
             }
-            return true;
+            return stop || navigationHandled[0];
         };
     }
 }
--- a/main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
+++ b/main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
@@ -220,75 +220,63 @@ public final class ActivityMixin {
     }
 
     /**
-     * Registers an interceptor for back press and/or navigate up on the given activity.
+     * Registers an interceptor for back press and navigate up on the given activity.
      *
-     * @param activity       The activity to register on. Held via WeakReference to avoid memory leaks.
-     * @param interceptBack  Whether to intercept the system back press.
-     * @param interceptNavUp Whether to intercept navigate up (Toolbar ← arrow). Use the returned
-     *                       BooleanSupplier in {@code onSupportNavigateUp()}.
-     * @param interceptor    Called when back/navUp is triggered.
-     *                       Parameter: a Runnable to conclude the intercepted navigation later
-     *                       (e.g. from a dialog confirmation).
-     *                       Return {@code true} to STOP navigation, {@code false} to CONTINUE.
+     * @param activity    The activity to register on. Held via WeakReference to avoid memory leaks.
+     * @param interceptor Called when back/navUp is triggered.
+     *                    Parameter: a Runnable to conclude the intercepted navigation later
+     *                    (e.g. from a dialog confirmation).
+     *                    Return {@code true} to STOP navigation, {@code false} to CONTINUE.
      * @return A BooleanSupplier to be called from {@code onSupportNavigateUp()}. Returns {@code true}
-     *         when navigation was handled or intercepted, {@code false} if interceptNavUp is false
-     *         or the activity is gone.
+     *         when navigation was handled or intercepted, {@code false} if the activity is gone.
      */
-    public static BooleanSupplier registerNavigationInterceptor(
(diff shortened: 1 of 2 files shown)


main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
```diff
@@ -205,4 +210,87 @@ public static void requireActivity(final @Nullable Activity activity, final Acti
...
+     *                       Return {@code true} to STOP navigation, {@code false} to CONTINUE.
+     * @return A BooleanSupplier to be called from {@code onSupportNavigateUp()}. Returns {@code true}
+     *         when navigation was intercepted (stopped), {@code false} to continue (call
+     *         {@code super.onSupportNavigateUp()}). Returns {@code false} immediately if
+     *         interceptNavUp is false or the activity is gone.
+     *         Note: the caller is responsible for invoking {@code super.onSupportNavigateUp()} when
+     *         the supplier returns {@code false}, as super calls cannot be made from outside the class.
+     */
+    public static BooleanSupplier registerNavigationInterceptor(
+            @NonNull final AppCompatActivity activity,
```
eddiemuc, line 236: For whole PR: replace AppCompatActivity with Activity (if possible?)

main/src/main/java/cgeo/geocaching/activity/ActivityMixin.java
```diff
@@ -205,4 +210,87 @@ public static void requireActivity(final @Nullable Activity activity, final Acti
...
+            @NonNull final Predicate<Boolean> interceptor) {
+
+        final WeakReference<AppCompatActivity> weakActivity = new WeakReference<>(activity);
+
+        if (interceptBack) {
+            final OnBackPressedCallback callback = new OnBackPressedCallback(true) {
+                @Override
+                public void handleOnBackPressed() {
+                    final AppCompatActivity a = weakActivity.get();
+                    if (a == null || a.isFinishing() || a.isDestroyed()) {
```
eddiemuc: Factor out a static method isActivityValid() for this check and use it everywhere in this PR where it makes sense

main/src/main/java/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java
```diff
@@ -366,17 +369,25 @@ private void finishWithResult() {
...
+     *
+     * @param isNavigateUp true if triggered by the Up arrow, false if by back press or Cancel menu item
+     * @return true to STOP navigation (a confirmation dialog was shown), false to CONTINUE
+     */
+    private boolean onNavigationIntercepted(final boolean isNavigateUp) {
         final GeocacheFilter newFilter = getFilterFromView();
-        final boolean filterWasChanged = !originalFilterConfig.equals(newFilter.toConfig());
+        final boolean filterWasChanged = originalFilterConfig != null && !originalFilterConfig.equals(newFilter.toConfig());
         if (filterWasChanged) {
             SimpleDialog.of(this).setTitle(R.string.confirm_unsaved_changes_title).setMessage(R.string.confirm_discard_changes).confirm(this::finish);
```
eddiemuc: This still uses "finish", I would have expected this to use the new triggernav helper. Why doesn't it do that?

eddiemuc (commented):
@copilot please check my comments. Please comment back, fix or make suggestions

eddiemuc:
@copilot :
Theres a "wrong input order" problem in activitymixin

 i dont like the complexity of current approach esp introduced staric map in activitymixin. What about this:
Instead of passing a boolean to the function y we pass a runnable which, if executed, would conclud3 the navigation step. Runnable is designed to be executed at a later point in time (eg after the user confirmed a dialog). Would that make code leas xomplex

eddiemuc:
@copilot :
* wrong import order in ActivityMixin still needs fixing
* ActivityMixin.triggerNavigation method doesn't make sense any more, please remove and call "backpress" in GeocachFilterActivity directly
* remove support for "Activity" for registerNavigationInterceptor, just support for AppCompatActivity is enough. This should also allow you to remove helper method for backpress in ActivityMixin
* add variant for ActivityMixin.isActivityValid() to pass a WeakReference<Activity> and do same validity check. 
* returned runnable to continue navigation should also use Weakreference of activity so no memory is leaked.

eddiemuc:
@copilot: are you able to squash all commits of this pr into one with message: 
Fix broken backpress handling of geocachefilteractivity, create general purpose navigation interceptor

eddiemuc:
@copilot :
- rename register method to "registerBackNavigationInterceptor"
- remove two booleans to decide what to intercept, always assume both is true
- in geocachfilteractivity, the home.menunentry should mimic nav up, not back press. Can you make it call the nav up instead of backpress please?

eddiemuc:
@copilot:
- on home in geocachfilteractivity: use activitymixin.navigateUp
- onsupportnavigateup in geocachfilteractivity : shouldn't we call and return super methos in case the navhandler returns false?

eddiemuc:
@copilot lot: I noticed that ActivityMixin.navigateUp calls finish under certain circumstances and does lots of stuff. How does this interfere with the back/nav-interceptor logic the current PR is executing? Will calls to ActivityMixin.navigateUp correctly get intercepted by our new method? Can the behaviour of navigateUp cause any problems eg loops? Can you investigage and probably advise on how to deal with this?

eddiemuc:
@copilot+claude-sonnet-4.6 please work on my previous comment

eddiemuc:
tested + works, lets merge here

eddiemuc:
Damn I forgot to squash commits before merging PR. Very sorry...

Can anyone tell me why github doesn't provide auto-squash when using rebase-merge? This is so inconvenient.

Issue #18142, opened 2026-05-27
conditional marker feedback


ztNFny:
Hey @eddiemuc , first of thanks a lot for this implementation. Starting this ticket to collect and discuss some feedback.

1. Using a stored filter should use the filter icon instead of needing to assign a separate icon
2. the "smiley" as placeholder for an unassigned icon is not clear. Should probably be a "greyed out" X or something similar
3. Long-tap on "Add rule" should show stored filters dialog (like long-tap on filters does in map activity)
4. long-tap on the icon in map toolbar should toggle the conditional marker display on/off
5. smiley as icon is bad as it's undescriptive - i'll look into finding something else

ztNFny:
icon suggestions (unrefined, posting them to get some feedback which direction would work)
- star on circle
<img width="445" height="444" alt="Image" src="https://github.com/user-attachments/assets/48b72f8a-5221-4894-96dd-4fc36afc8cdc" />

- marker
<img width="269" height="266" alt="Image" src="https://github.com/user-attachments/assets/43780cfa-eac0-4917-a763-930dbed9a43f" />

- label
<img width="261" height="262" alt="Image" src="https://github.com/user-attachments/assets/b1244c1c-c218-4da0-b2f8-41d35ebdcaff" />

moving-bits:
Thanks for the proposals. I haven't checked the new feature yet, so I'm not sure if the following proposal applies:

If the new feature is about applying markers based on filters, one could combine our filter symbol (funnel) with the star from above's first proposal.

The other two symbols (the "mike" (marker) and "label") look too generic, IMHO.

A different approach could be to use a blank cache marker symbol (in black & white, outline only) and place a large star symbol in it, similar to Google Font's ["Star" symbol in "Outline" version](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:stars:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=badge&icon.size=24&icon.color=%231f1f1f).

ztNFny:
> A different approach could be to use a blank cache marker symbol (in black & white, outline only) and place a large star symbol in it, similar to Google Font's ["Star" symbol in "Outline" version](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:stars:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=badge&icon.size=24&icon.color=%231f1f1f).

toolbar icons are single-color - so that's basically the above "star on circle"

[...]

PR #18147, opened 2026-05-28
dedicated "conditional marker" icon

ztNFny:
implements a new icon for conditional marker
Image

and an option to disable/pause the display of them

related to https://github.com/cgeo/cgeo/issues/18142


ztNFny:
@eddiemuc let's merge this regardless of future improvements to the feature

main/src/main/java/cgeo/geocaching/settings/ConditionalCacheMarkersActivity.java
```diff
@@ -191,6 +191,7 @@ public boolean onCreateOptionsMenu(final Menu menu) {
     public boolean onOptionsItemSelected(final MenuItem item) {
         final int itemId = item.getItemId();
         if (itemId == R.id.menu_item_save) {
+            Settings.setConditionalCacheMarkersEnabled(true);
```
eddiemuc, line 194: Why should ever save enable markers?

main/src/main/java/cgeo/geocaching/settings/ConditionalCacheMarkersActivity.java
```diff
@@ -199,6 +200,10 @@ public boolean onOptionsItemSelected(final MenuItem item) {
         } else if (itemId == android.R.id.home) {
             finish();
             return true;
+        } else if (itemId == R.id.menu_item_disable) {
```
eddiemuc, line 203: Shouldn't this rather be a toggle?

eddiemuc (commented):
See comments

main/src/main/java/cgeo/geocaching/settings/ConditionalCacheMarkersActivity.java
ztNFny: I made it so that pause disables it, OK saves + enable - because you wouldn't want to save to then realize nothing is showing as you forgot to un-disable them

main/src/main/java/cgeo/geocaching/settings/ConditionalCacheMarkersActivity.java
ztNFny: That's the global whether markers are shown.

eddiemuc:
I am not totally convinced of this approach, but let's merge and see how it behaves in nightlies.
(Also I am working on a bigger redesign anyway)

ztNFny:
The "main" toggle is in map toolbar - but only if there's room for 4 buttons there. So the disable button in that dialog is kind of a workaround for smaller screens.

PR #18164, opened 2026-05-31
Unify named filters and conditional markers

eddiemuc:
This PR will unify named filters and conditional markers.


eddiemuc:
Named filters will become first class entities. They will be stored in database and be editable in their own screen:
image

A named filter has a name, a filter and an associated marker (optional). A named filter can be activated/deactivated to be shown as conditional marker symbol on caches. Named filters have an order which defines their priority for marker display.

Named filters can be quick-chosen as filters in map view and cache list view by long-clicking filter icon:'
image

Named filters can be accessed via icon from map view or overflow item in cache list view. Long click on that icon allows quick enabling/disabling of conditional marking:
image

In GeocacheFilter-Activity, old "name filter management" section is removed. Instead there are three more icons in advanced mode:
image
from left to right:
* add new filter criterion (same as before)
* quick-select named filter (result is shown in screenshot)
* save current filter configuration as named filter
* open named filter dialog

In GeocacheDetails, matching named filters are shown:
image

Existing named filters and conditional markers are migrated

eddiemuc:
@moving-bits and I removed access to named filters in Settings completely

eddiemuc:
Note: I realize this is a large PR. Nevertheless I would like to merge it rather soon because it blocks other work (esp due to database migration) and will soon lead to conflicts due to involved refactorings. Thus if there are no major concerns I will soon proceed (GUI layouting, icons, texts and stuff can be fixed later - I am aware that current design might not be the final one)

ztNFny:
@eddiemuc unfortunately this PR has major issues:
- many filters were transformed wrong, they now have no filter criteria at all. I still have to figure out which filters exactly - as unfortunately they are now gone. But I think it's all filters "cache is in list" or "cache is in other named filter"
- this broke hiearchical view in "Select filter" dialog
image

ztNFny:
The filters that failed to convert are:
- Stored Lists: multiple lists selected
- Named filter (multiple of them with OR connect)
- Last details updated: > 30 days

eddiemuc:
> The filters that failed to convert are:
> 
> * Stored Lists: multiple lists selected
> * Named filter (multiple of them with OR connect)
> * Last details updated: > 30 days

@ztNFny I can confirm the problems with stored lists + named filters. I will provide a fix for those (and extend named filter filter to allow multiple named filters while I am at it).

However I cannot reproduce the problem with "last details updated".

ztNFny:
I extracted the filter from DB and it seems broken 😱 It was a relative filter >30 days, but the offsets are both stored as -1
I guess it actually got broken at some earlier time and i didn't notice till now ...

`{"name":"♥️ Interesting:⏳ Interesting & Last refresh >30d","advanced":true,"inconclusive":false,"tree":{"type":"AND","children":[{"type":"stored_list","config":{"values":["339","356","471","136","169","91","60","48"]}},{"type":"stored_since","config":{"relative":false,"minOffset":-1,"maxOffset":-1}}]}}`


Did you also happen to look at the broken hierarchy yet? I saw in your screenshot that it's supposed to be working, so I assume that's a migration issue as well?

eddiemuc:
> Did you also happen to look at the broken hierarchy yet? I saw in your screenshot that it's supposed to be working, so I assume that's a migration issue as well?

That at least is an easy fix: I used two colons instead of one as separater. WIll fix it.

Issue #8402, opened 2020-06-01
RejectedExecutionException


Lineflyer:
From Google Play flagged as new for beta version 2020.05.31-RC
1 report, Android 9

```
io.reactivex.rxjava3.exceptions.UndeliverableException: 
 
  at io.reactivex.rxjava3.plugins.RxJavaPlugins.onError (RxJavaPlugins.java:367)
 
  at io.reactivex.rxjava3.internal.schedulers.NewThreadWorker.scheduleActual (NewThreadWorker.java:153)
 
  at io.reactivex.rxjava3.internal.schedulers.NewThreadWorker.schedule (NewThreadWorker.java:50)
 
  at io.reactivex.rxjava3.core.Scheduler$Worker$PeriodicTask.run (Scheduler.java:535)
 
  at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.run (ScheduledRunnable.java:65)
 
  at io.reactivex.rxjava3.internal.schedulers.ScheduledRunnable.call (ScheduledRunnable.java:56)
 
  at java.util.concurrent.FutureTask.run (FutureTask.java:266)
 
  at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run (ScheduledThreadPoolExecutor.java:301)
 
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
 
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
 
  at java.lang.Thread.run (Thread.java:764)
Caused by: java.util.concurrent.RejectedExecutionException: 
 
  at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution (ThreadPoolExecutor.java:2085)
 
  at java.util.concurrent.ThreadPoolExecutor.reject (ThreadPoolExecutor.java:848)
 
  at java.util.concurrent.ScheduledThreadPoolExecutor.delayedExecute (ScheduledThreadPoolExecutor.java:334)
 
  at java.util.concurrent.ScheduledThreadPoolExecutor.schedule (ScheduledThreadPoolExecutor.java:579)
 
  at io.reactivex.rxjava3.internal.schedulers.NewThreadWorker.scheduleActual (NewThreadWorker.java:146)
```

bekuno:
Maybe this point is helpful:
[...]

PR #18166, opened 2026-05-31
Add RxJava error handler

moving-bits:
## Description
Adds a global error handler for RxJava to prevent app crashes such as #8402 (and probably more, haven't cross-checked the whole issue list, but there are quite a few of them in Play Store logs).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CgeoApplication.java
+++ b/main/src/main/java/cgeo/geocaching/CgeoApplication.java
@@ -29,13 +29,17 @@ import android.os.UserManager;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import java.io.IOException;
 import java.lang.reflect.Method;
 import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/CgeoApplication.java
```diff
@@ -176,6 +170,31 @@ public void onCreate() {
...
+
+            // Handle specific known cases
+            if (cause instanceof IOException || cause instanceof InterruptedException || cause instanceof java.util.concurrent.RejectedExecutionException) {
+                // Fine: network/file disruption, some blocking code was cancelled, or Scheduler/ThreadPool was shut down before the stream finished
+            } else if (cause instanceof NullPointerException || cause instanceof IllegalArgumentException) {
+                // That's likely a bug in the application
+                Objects.requireNonNull(Thread.currentThread().getUncaughtExceptionHandler()).uncaughtException(Thread.currentThread(), cause);
+            } else if (cause instanceof IllegalStateException) {
+                // That's a bug in RxJava or the app code
+                Objects.requireNonNull(Thread.currentThread().getUncaughtExceptionHandler()).uncaughtException(Thread.currentThread(), cause);
```
bekuno: Is this wanted to have the same content?

moving-bits: yes, it's basically one "fine to ignore", one "rethrow" and one "log others" branch. Have consolidated the source

PR #18173, opened 2026-06-03
log translation fixes

ztNFny:
- Fix an issue that caused only every 2nd log to get translated
- add translation info to translated logs
image


ztNFny:
The failing tests are actually caused by https://github.com/cgeo/cgeo/commit/f32f9e3f028850fcf00d8b847cf48bb07b03021f
not sure why they are failing here and not there ..

moving-bits (approved):
LGTM, thanks for fixing

moving-bits:
Stumbled upon a bug I had not noticed during my quick smoke test:

- open a log view with more logs than fit on one page
- translate two or so of them
- scroll page down

When recycling log view due to scrolling, the newly appearing logs are already marked as translated, even though they are not.

@ztNFny 
Can you cross-check, please?

ztNFny:
> Stumbled upon a bug I had not noticed during my quick smoke test:
> 
> * open a log view with more logs than fit on one page
> * translate two or so of them
> * scroll page down
> 
> When recycling log view due to scrolling, the newly appearing logs are already marked as translated, even though they are not.
> 
> @ztNFny Can you cross-check, please?

added to https://github.com/cgeo/cgeo/pull/18179

PR #18179, opened 2026-06-04
don't collapse logs on translation

ztNFny:
remember that a log text was expanded and don't auto-collapse it again on text change (= translation)


ztNFny:
failing tests are unrelated

Issue #18150, opened 2026-05-28
offline translate (bergamot) issues


ztNFny:
- [x] Some strings still refer to Google, e.g. translating logs - see PR #18153
- [x] is the download package size 20-40mb still valid? - seems so, see https://github.com/cgeo/cgeo/issues/18150#issuecomment-4579500076
- [x] translating logs (other than the first per cache) requires to perform the action twice, first time nothing happens - see PR #18173
- [ ] translating logs might be slow, show a spinner while translating
- [x] with mlkit removed is there a cleanup routine in place to delete the downloaded mlkit files from internal storage? - see PR #18154
- [ ] c:geo crashes when trying to translate a log entry, when language pair is not yet available locally
- [ ] Downloading language files has no indication that something is happening in the background. No download manager notification, no "running downloads" entries, 
- [ ] language selection has no indication which files are already scheduled for download (or available offline already).
- [x] c:geo crashes when source and target languages are identical - see PR #18152
- [x] attribution in listing ("translated by Bergamot") still required? - removed with PR #18181
- [x] add attribution to about:contributors page - see PR #18183

SammysHP:
> translating logs requires to perform the action twice, first time nothing happens

Works for me the first time.

---

Downloading language files has no indication that something is happening in the background. No download manager notification, no "running downloads" entries, even opening the language selection has no indication which files are already scheduled for download.

moving-bits:
> * is the download package size 20-40mb still valid?

en/de is app. 20 MB compressed / 30 MB uncompressed, see files in `<Context.getDataDir()>/bergamot/*`

moving-bits:
> > translating logs requires to perform the action twice, first time nothing happens
> 
> Works for me the first time.

When translating several log entries, I sometimes experienced this problem as well, but not for all log entries. Maybe some internal state needs to be reset after having completed a translation, or maybe that state should be log-specific, but is "global"?

moving-bits:
> with mlkit removed is there a cleanup routine in place to delete the downloaded mlkit files from internal storage?

where did ML Kit store its translation models to?
Is it a separate folder, that could be deleted completely? 
Or, if not: What are the relevant file masks to be taken into account?

ztNFny:
/data/data/cgeo.geocaching/no_backup/com.google.mlkit.translate.models/ had subfolders with the actual models
/data/data/cgeo.geocaching/no_backup/com.google.mlkit.InstallationId
[...]

PR #18182, opened 2026-06-04
Shorten log translation hint

moving-bits:
## Description
Makes log translation notice less prominent. Tap on hint still triggers restoring original text.

image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/log/LogsViewCreator.java
+++ b/main/src/main/java/cgeo/geocaching/log/LogsViewCreator.java
@@ -27,7 +27,6 @@ import cgeo.geocaching.utils.html.UnknownTagsHandler;
 import cgeo.geocaching.utils.offlinetranslate.TranslateAccessor;
 
 import android.os.Bundle;
-import android.text.SpannableStringBuilder;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -206,7 +205,7 @@ public abstract class LogsViewCreator extends TabbedViewPagerFragment<LogsPageBi
                     final OfflineTranslateUtils.Status translationStatus = getTranslationStatus(log);
                     if (translationStatus.isTranslated()) {
                         translationStatus.setNotTranslated();
-                        holder.binding.logTranslateHint.setVisibility(View.GONE);
+                        holder.binding.logTranslateNote.setVisibility(View.GONE);
                         fillViewHolder(null, holder, log);
                     } else {
                         final String logText = HtmlUtils.extractText(log.log);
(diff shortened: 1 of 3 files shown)


ztNFny:
you removed what imho is the most relevant part of the message: that the text was translated and is not the original

beyond that, i don't see any reason to have a completly different UI in logs than what we have in description

ztNFny:
and please use fair screenshots, I feel you're intentionally misrepresenting the problem 🙁 Noone would ever translate several logs that obviously contain no content at all. A more realistic screenshot has already been posted in https://github.com/cgeo/cgeo/issues/18150#issuecomment-4620232388

moving-bits:
> you removed what imho is the most relevant part of the message: that the text was translated and is not the original

not at all, I just compressed that info: If there's a flag + the text "Tap to restore original", it's translated, otherwise it's the original text. (And not to forget: The user has translated that very text probably just seconds ago by opening the log entry's context menu + selecting the "translate offline" option.)

> beyond that, i don't see any reason to have a completly different UI in logs than what we have in description

Cache listing has one text, typically a large one, whereas log view contains of several texts, mostly shorter ones (except for special caches).

> and please use fair screenshots, I feel you're intentionally misrepresenting the problem 🙁 Noone would ever translate several logs that obviously contain no content at all.

Ten logs or more would be a bit unrealistic, agreed, but having 5 - 7 potentially relevant logs is not uncommon for me. I use translation feature mostly in countries I'm not capable of reading the local language, so even shorter texts of one or two lines may be relevant to me in finding a cache, if the hiding location is tricky.

ztNFny:
> Cache listing has one text, typically a large one, whereas log view contains of several texts, mostly shorter ones (except for special caches).

you're picking tourist tradis there - obviously they have only short texts, but that's not representative. And it's especially not representative of logs that'll be translated.
Looking at Project-GC stats average log-length is around 50 words

> Ten logs or more would be a bit unrealistic, agreed, but having 5 - 7 potentially relevant logs is not uncommon for me. I use translation feature mostly in countries I'm not capable of reading the local language, so even shorter texts of one or two lines may be relevant to me in finding a cache, if the hiding location is tricky.

all the logs in your original screenshot very obviously contain no relevant information - and that's obvious before even translating them.

GC6WZ22 is a random cache somewhere in the middle of France i picked for testing. This is basically a view of all logs, with all of the short ones - similar to the ones in your screenshot obviously not worth a second look:
image
image
image
Considering ALL logs that cache has I never had more than 2 worth translating on any screen.

moving-bits:
I remain unconvinced, but I’m not going to fight for it in this case, it’s not important enough to me here.

PR #18185, opened 2026-06-04
fix #18164: fix migration problems for unifying named filters

eddiemuc:
fix #18164: fix migration problems for unifying named filters

### DIFF
--- a/main/src/androidTest/java/cgeo/geocaching/filters/GeocacheFilterTestUtils.java
+++ b/main/src/androidTest/java/cgeo/geocaching/filters/GeocacheFilterTestUtils.java
@@ -42,7 +42,7 @@ public class GeocacheFilterTestUtils {
     }
 
     public static boolean testSingleLogic(final Geocache cache, final IGeocacheFilter filter, final Boolean expectedResult) {
-        assertThat(filter.filter(cache)).as("LOGIC: Wrong test result for filter: " + filter.getConfig() + "/cache: " + cache).isEqualTo(expectedResult);
+        assertThat(filter.filter(cache)).as("LOGIC: Wrong test result for filter: " + filter.getJsonConfig() + "/cache: " + cache).isEqualTo(expectedResult);
         return true;
     }
 
--- a/main/src/androidTest/java/cgeo/geocaching/filters/NamedFilterGeocacheFilterTest.java
+++ b/main/src/androidTest/java/cgeo/geocaching/filters/NamedFilterGeocacheFilterTest.java
@@ -12,6 +12,7 @@ import cgeo.geocaching.utils.functions.Action1;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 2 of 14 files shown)


ztNFny:
migration and hierarchy are working now,  thanks!

looking at the filter list what's the purpose of the "marker" icon next to each filter?

ztNFny:
> looking at the filter list what's the purpose of the "marker" icon next to each filter?

nevermind, that's the icon assigned to each marker which is of course blank for all those migrated.

The "Named filters" management dialog should show the same drawable for unassigned icon, right now it shows a "X" which seems like a delete/close button. Or, as the title is "Named fitlers" and not "named markers" I suggest using the filter icon there as default

ztNFny:
on thing that got lost and I suggest to restore it: Loading a named filter in the filter dialog. This is right now only possible using long-tap on the filter icon on map but not from filter activity. makes it impossible to quickly create a filter based on an existing named filter

ztNFny:
and another bug (sorry for the spam): assigned emoj is too large in the filters/markers list

eddiemuc:
> The "Named filters" management dialog should show the same drawable for unassigned icon, right now it shows a "X"

I think this was suggested earlier (when it was still a "conditional marker" screen), but can be changed of course.

> on thing that got lost and I suggest to restore it: Loading a named filter in the filter dialog.

That's still possible: click the second button (right to the "add criteria" button). Granted, the symbol is not ideal, I used the "add/open file" icon for now. But I didn't know what else to use. Any ideas?

> assigned emoj is too large in the filters/markers list

Sure that's easy to change

ztNFny:
> That's still possible: click the second button (right to the "add criteria" button). Granted, the symbol is not ideal, I used the "add/open file" icon for now. But I didn't know what else to use. Any ideas?

that's adding a named filter as filter-criteria (same as doing add -> named filter), not loading the named filter. as it's duplicating an existing functionality + the "load" is missing i suggest to turn it into a "load" and create a transient filter from that criateria

eddiemuc:
> > That's still possible: click the second button (right to the "add criteria" button). Granted, the symbol is not ideal, I used the "add/open file" icon for now. But I didn't know what else to use. Any ideas?
> 
> that's adding a named filter as filter-criteria (same as doing add -> named filter), not loading the named filter. as it's duplicating an existing functionality + the "load" is missing i suggest to turn it into a "load" and create a transient filter from that criateria

You want to use named filters as templates rather than as references? Thats ok we can change the behavior. What would be a good button icon for that functionality?

ztNFny:
> > > That's still possible: click the second button (right to the "add criteria" button). Granted, the symbol is not ideal, I used the "add/open file" icon for now. But I didn't know what else to use. Any ideas?
> > 
> > that's adding a named filter as filter-criteria (same as doing add -> named filter), not loading the named filter. as it's duplicating an existing functionality + the "load" is missing i suggest to turn it into a "load" and create a transient filter from that criateria
> 
> You want to use named filters as templates rather than as references? Thats ok we can change the behavior. What would be a good button icon for that functionality?

Yes - loading a saved filter and then slightly modifying it is a common scenario.
What was the icon to load a stored filter before this change? Should be that - don't change a running system 😉

eddiemuc:
See https://github.com/cgeo/cgeo/pull/18194

Issue #18108, opened 2026-05-18
App not responding on program start


moving-bits:
Reported by user on support for current release 2026.05.15, ticket [293774](https://cgeosupport.droescher.eu/scp/tickets.php?id=10375):

On each program start the user gets the "app not responding" warning from Android. Tapping "wait" makes everything working fine (seemingly), but the error returns after a short while.

One possible cause: The user's database is about 2.4G in size.

Do we have anything during app startup that runs in the foreground and gets blocked by database access / database initialization on program startup?

Anything else running in foreground during startup and taken such a long time that it causes app hang?

--- 

Update 2026-06-06:
Meanwhile the user sent two videos showing the delay. One is from directly opening the map, the other from mapping a cache list. So it seems to be starting the map causing the delay, not starting the program. (At least not alone.) 
A possible cause for this is addressed by #18199 by preloading the list of tileproviders in background during startup instead of when opening the map. The user in question has more than 40 offline maps.

eddiemuc:
If there are longer blockers in main thead, they should be logged by LooperLogger. It would be worth having a logfile for this user.

moving-bits:
> It would be worth having a logfile for this user.

That's part of the linked support ticket.
Haven't checked yet, though, whether it contains such log entries.

moving-bits:
Here's a log extract for LooperLogger:

```
05-18 18:24:09.061  8381  8381 I cgeo    : [main] LooperLogger: started for Thread: main
05-18 18:24:09.063  8381  8381 I cgeo    : [main] [TranslationModelManager]: Supported languages: [de, hi, pt, lt, hr, lv, ht, hu, uk, id, ur, mk, af, mr, ms, el, mt, en, eo, is, it, zh, es, et, ar, vi, ja, fa, ro, nl, no, be, fi, ru, bg, bn, fr, ka, sk, sl, ga, sq, ca, kn, sv, ko, sw, gl, ta, gu, cs, te, th, cy, tl, pl, da, he, tr]
05-18 18:24:09.063  8381  8381 I cgeo    : [main] [CtxLog]CGeoApplication.onCreate:END (179ms)
05-18 18:24:09.072 28721 28762 I Kumiho-Kumiho: getPackageName: cgeo.geocaching
05-18 18:24:09.074 28721 28762 I Kumiho-Kumiho: getPackageName: cgeo.geocaching
[...]

PR #18186, opened 2026-06-04
Fix chunking controller

moving-bits:
## Description
While analyzing for #18108 I've stumbled about our chunking controller: First chunk is of (up to) defined size (900), but then there's only one follow-up chunk with all the rest. (And as I was testing a list of 17k caches, second chunk was nearly the whole list.)


eddiemuc (approved):
good catch, thanks a lot!

PR #18193, opened 2026-06-05
Emoji

ztNFny:
- replace internal limited emoji picker with one that offers all emojis, including search, categories and recently used
- migration of old int-based emoji storage to string-based (for combined emojis, e.g. colors, flags, ..)
- The 26 colored circles previously available are being converted to the 9 circle emojis with an "as close as possible" mapping
- the number unicode chars are only available via search, not selectable in the list as the list of emojis can't be extended but the search results are custom

The change in storage from int to String is the reason this PR touches a lot of files, the main change is in EmojiUtils

image


ztNFny:
@eddiemuc @moving-bits this PR should be ready from my side. But due to the sheer size of it, affecting many parts of cgeo (cache, lists, filters) and also requiring quite some data migration (which is something that is hard to undo once broken) I'd appreciate a review from you - time permits of course - and of course everyone else is also welcome!

main/src/main/java/cgeo/geocaching/list/StoredList.java
```diff
@@ -58,10 +58,13 @@ public final class StoredList extends AbstractList {
     public static final StoredList TEMPORARY_LIST = new StoredList(TEMPORARY_LIST_ID, "<temporary>", EmojiUtils.NO_EMOJI, true, 0); // Never displayed
     public static final int STANDARD_LIST_ID = 1;
     public final boolean preventAskForDeletion;
+    /** emoji assigned to this list as its marker, or null/empty for "none" */
+    @Nullable public final String emojiMarker;
```
eddiemuc, line 60: "markerId" (int) is still a member of AbstractList. I think its ok to move the marker concept from AbstractList to StoredList, but still it needs to be removed in AbstractList in that case.

Why not call it "marker" or "markerId" (not "emojiMarker")?

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -1980,8 +1983,8 @@ public void onUpgrade(final SQLiteDatabase db, final int oldVersion, final int n
                             if (markersRoot != null && markersRoot.isArray()) {
                                 int j = 0;
                                 for (final com.fasterxml.jackson.databind.JsonNode child : markersRoot) {
-                                    final int markerId = cgeo.geocaching.utils.JsonUtils.getInt(child, "markerId",
```
eddiemuc, line 1982: Don't change past dbVersion migrations. You have to assume that this code part has already run on some installations and thus have to do your migration step based on whatever previous migration steps left over.

Looking at your migration code towards version 110 I believe you can just not touch code here. Your migration already seems complete without it

main/src/main/java/cgeo/geocaching/ui/ImageParam.java
```diff
@@ -62,6 +64,17 @@ public static ImageParam emoji(final int emojiSymbol, final int emojiSizeInDp) {
         return new ImageParam(-1, emojiSymbol, Math.max(2, emojiSizeInDp) , null);
     }
 
+    /**
+     * create from emoji string (as returned by the emoji picker / stored as cache or list marker)
+     */
+    public static ImageParam emoji(@Nullable final String emojiStr) {
```
eddiemuc, line 59: The "emoji" overrides taking an int as parameter should be removed from ImageParam. If they are still used that would be suspicious.

main/src/main/java/cgeo/geocaching/ui/ImageParam.java
```diff
@@ -36,6 +36,8 @@ public class ImageParam {
     @DrawableRes
     private final int drawableId;
     private final int emojiSymbol;
+    @Nullable
+    private final String emojiStr;
```
eddiemuc: Member "int emojiSymbol" should be removed. If it is still used that is suspicious

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -3,276 +3,213 @@
...
+    private static final String VARIATION_SELECTOR_EMOJI_STR = new String(Character.toChars(VariationSelectorEmoji));
 
-    // list of emojis supported by the EmojiPopup
-    // should ideally be supported by the Android API level we have set as minimum (currently API 21 = Android 5),
-    // but starting with API 23 (Android 6) the app automatically filters out characters not supported by their fonts
-    // for a list of supported Unicode standards by API level see https://developer.android.com/guide/topics/resources/internationalization
-    // for characters by Unicode version see https://unicode.org/emoji/charts-5.0/full-emoji-list.html (v5.0 - preferred compatibility standard based on our minAPI level)
-    // The newest emoji standard can be found here: https://unicode.org/emoji/charts/full-emoji-list.html
-
+    // commonly used built-in emojis (single codepoints) referenced from various places in the app
```
eddiemuc, line 89: Shouldn't those constants be replaced with their String equivalent?

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -3,276 +3,213 @@
...
+     *
+     * @param currentValue the currently assigned emoji, or empty/null for "none assigned"; pass {@code null} together
+     *                     with {@code bulkMode=true} for the "assign to multiple caches" case where there is no single
+     *                     current value
+     * @param bulkMode     {@code true} when assigning to several caches at once (no single current value); shows the
+     *                     generic marker icon as preview and always offers the reset-to-default action
+     * @param cache        optional cache used to render the type marker as default/preview icon
+     * @param setNewCacheIcon callback receiving the selected emoji String, or the empty string when reset to default
+     */
+    public static void selectEmojiPopup(final Context context, @Nullable final String currentValue, final boolean bulkMode, @Nullable final Geocache cache, final Action1<String> setNewCacheIcon) {
```
eddiemuc: You mentioned some problems using the picker in an AlertDialog, so I assume the rather complex looking dialog logic in this method is necessary for the picker to be displayed correctly.

main/src/main/res/layout/dialog_title_button_button.xml
```diff
@@ -2,7 +2,8 @@
 <RelativeLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content">
+    android:layout_height="wrap_content"
+    android:minHeight="@dimen/emoji_picker_title_height">
```
eddiemuc, line 6: The layout "dialog_title_button_button" is not specific to emoji picker, yet it has a emoji-picker-specific dimension assigned. That seems strange

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -3,276 +3,213 @@
...
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Set;
 
+import com.google.android.material.bottomsheet.BottomSheetBehavior;
+import com.google.android.material.bottomsheet.BottomSheetDialog;
 import com.google.android.material.button.MaterialButton;
+import com.google.common.util.concurrent.Futures;
```
eddiemuc, line 64: There seems to be problems with setup of guava. Jenkins reports compile failure, and I have them locally too:
Duplicate class com.google.common.util.concurrent.ListenableFuture found in modules guava-26.0-jre.jar -> guava-26.0-jre (com.google.guava:guava:26.0-jre) and listenablefuture-1.0.jar -> listenablefuture-1.0 (com.google.guava:listenablefuture:1.0)

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -296,139 +233,273 @@ public int getSpanSize(final int position) {
...
-                tt = itemView.findViewById(R.id.info_title);
-                tv = itemView.findViewById(R.id.info_text);
-                iv = itemView.findViewById(R.id.info_drawable);
-                sep = itemView.findViewById(R.id.separator);
             }
         }
+    }
+
+    /** @return true if the given emoji String represents "no emoji" (null or empty) */
+    public static boolean isNoEmoji(@Nullable final String emoji) {
```
eddiemuc: I have seen this method's return value negated many times ("if (!isNoEmoji(something))"). There should  really be a counterpart "isEmoji()"-method. Double-negations are very hard to read.

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -3,276 +3,213 @@
...
+     * Opens a dialog hosting the androidx EmojiPicker so the user can select an emoji for a cache icon / list marker /
+     * filter marker.
+     *
+     * @param currentValue the currently assigned emoji, or empty/null for "none assigned"; pass {@code null} together
+     *                     with {@code bulkMode=true} for the "assign to multiple caches" case where there is no single
+     *                     current value
+     * @param bulkMode     {@code true} when assigning to several caches at once (no single current value); shows the
+     *                     generic marker icon as preview and always offers the reset-to-default action
+     * @param cache        optional cache used to render the type marker as default/preview icon
+     * @param setNewCacheIcon callback receiving the selected emoji String, or the empty string when reset to default
```
eddiemuc: SHould ne named not after a specific use case but generic (eg "selectAction")

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
```diff
@@ -3,276 +3,213 @@
...
-            }
-        }
+    /**
+     * Opens a dialog hosting the androidx EmojiPicker so the user can select an emoji for a cache icon / list marker /
+     * filter marker.
+     *
+     * @param currentValue the currently assigned emoji, or empty/null for "none assigned"; pass {@code null} together
+     *                     with {@code bulkMode=true} for the "assign to multiple caches" case where there is no single
+     *                     current value
+     * @param bulkMode     {@code true} when assigning to several caches at once (no single current value); shows the
```
eddiemuc: I can't compile and run code, so I have to deduct from code alone what the neaming of "buldMode" really is.

In its current form it indicates that the method contains use-case-specific logic ("bulk mode" is something very specific to assigning one marker to multiple geocaches), it is unnecessary that the method knows about this.

If the comment correctly describes the behaviour of the paramter then it canbe renamed to "forceResetAction" (and handled as such). The "generic marker" preview can assumingly be triggered by whether cache is null or not, so bulkMode is not needed for that.

eddiemuc (commented):
See comments

eddiemuc:
I am a big fan of getting a full-blown emoji picker into c:geo, so I support this change. 

One big concern back when we put in the "old" emoji picker was to only use emojis that will be supported across all Android versions/systems (as far as I can remember). I am assuming with this picker this would no longer be the case, correct? Is that a problem?

I saw in code that you had to introduce Futures to integrate the picker. Unfortunately this makes my build fail with some futurelistener-duplicate exceptions (I also see them in Jenkins). This also leads to not being able to check the dialog itself on emulator. Thus I kind of had to omit a more deep review of the EmojiUtils class itself which was probably the most tricky one. The migration parts, as far as I can tell from code, look fine to me - see my comments on some details.

Thanks a lot for your work!

ztNFny:
> I saw in code that you had to introduce Futures to integrate the picker. Unfortunately this makes my build fail with some futurelistener-duplicate exceptions (I also see them in Jenkins). 

build works fine here. what could be the issue there?

> One big concern back when we put in the "old" emoji picker was to only use emojis that will be supported across all Android versions/systems (as far as I can remember). I am assuming with this picker this would no longer be the case, correct? Is that a problem?

I'm not aware of that concern, at least it wasn't considered when we expanded the list.

eddiemuc:
> > I saw in code that you had to introduce Futures to integrate the picker. Unfortunately this makes my build fail with some futurelistener-duplicate exceptions (I also see them in Jenkins).
> 
> build works fine here. what could be the issue there?

Seems to be some guava issue. We have a stone-old "workaround" (?) for some problem in the past in gradle which might be related to that: 
https://github.com/cgeo/cgeo/blob/a0a7b4f41da4da08205e1a11b02857d6c5b1ab95/build.gradle#L99

This was well before my time, anyone knows what this is about?

moving-bits:
> > I saw in code that you had to introduce Futures to integrate the picker. Unfortunately this makes my build fail with some futurelistener-duplicate exceptions (I also see them in Jenkins).
> 
> build works fine here. what could be the issue there?

I had similar build problems in the beginning, but was able to resolve them by adding this to our main `build.gradle`:

```
implementation 'com.google.guava:guava:33.6.0-android'
```

and removing those lines from the other `build.gradle` file:

```
// new versions of checkstyle use guava 27, which has troublesome dependencies: https://groups.google.com/forum/#!topic/guava-announce/Km82fZG68Sw
force 'com.google.guava:guava:26.0-jre'
```

> > One big concern back when we put in the "old" emoji picker was to only use emojis that will be supported across all Android versions/systems (as far as I can remember). I am assuming with this picker this would no longer be the case, correct? Is that a problem?
> 
> I'm not aware of that concern, at least it wasn't considered when we expanded the list.

We've added some safeguard back then to skip unsupported codepoints, need to look up that in code.

ztNFny:
> I had similar build problems in the beginning, but was able to resolve them by adding this to our main build.gradle:\n\nimplementation 'com.google.guava:guava:33.6.0-android'\nand removing those lines from the other build.gradle file

Will look into it. Strange that my Android studio which is using gradle as build system doesn't have these problems...

moving-bits:
> We've added some safeguard back then to skip unsupported codepoints, need to look up that in code.

Found the code: This was part of `EmojiUtils.selectEmojiPopup()`, `checker.hasGlyph()`

moving-bits:
Freshly migrated database, where are those items in recently used list coming from / what do they represent?

image

Additionally, those items are truncated.

moving-bits:
Proposal:

The new picker has a nice integrated search feature. Currently it's icon + some underlined space, labelled "Search". This requires space that could be used to display more category items.

Proposal: Reduce search feature to magnifier icon initially.
This gives category items more space, and as search field expands anyway, it's no loss of functionality.

ztNFny:
> Freshly migrated database, where are those items in recently used list coming from / what do they represent?
> 
> image
> 
> Additionally, those items are truncated.

Do you only have them in recent or also other places? Might be i missed a migration for recent

ztNFny:
> Proposal:
> 
> The new picker has a nice integrated search feature. Currently it's icon + some underlined space, labelled "Search". This requires space that could be used to display more category items.
> 
> Proposal: Reduce search feature to magnifier icon initially.
> This gives category items more space, and as search field expands anyway, it's no loss of functionality.

I tried that, but then search looks just like any other category and people might not realize it's a search. And it makes it harder to hit that area. 
For me with the emoji selector i 95% use search or recent, rarely scrolling through the list, so icd like to keep the search accessible.

I also looked at what some other apps are doing and the current way seems to be a common one, with the next popular being a fool width search field at the top and categories at the bottom.

moving-bits:
> Do you only have them in recent or also other places? Might be i missed a migration for recent

So far I have seen it in "recent" only, not as an actual cache icon. But I need to check in more detail - I'm currently working with a test database of 90k caches, not my own.

ztNFny:
Ok. Anyway take your time, i won't be able to look into anything before tomorrow.

moving-bits:
> > Proposal: Reduce search feature to magnifier icon initially.
> 
> I tried that, but then search looks just like any other category and people might not realize it's a search. And it makes it harder to hit that area.

Here's some quick mockup that might resolve this issue:

image

Make magnifier symbol "button sized" + add a divider. (Exact spacings, size & divider color TBD)

Additionally, an ellipsis right of the currently displayed category icons might be a good indicator that there are more of them to the right.

Edit: Or, if "Search" is just an icon initially, it could be placed as an actual button to the left of the "current icon" button (but still expanding to a full line search in its current place, not in the header).

image

image

moving-bits:
When scrolling through the list, the current category is highlighted. This highlighting switches to next (or previous) category icon according to the currently displayed part of the list - ok.

Finding: If it switches to a category icon currently not visible (due to horizontal scrolling of category icon list), you have no (visible) highlight indicator anymore.

image

Category icon list should make sure the currently highlighted icon is visible.
This should be fixed independently from the result of above's discussion about search indicator.

ztNFny:
> > > Proposal: Reduce search feature to magnifier icon initially.
> > 
> > I tried that, but then search looks just like any other category and people might not realize it's a search. And it makes it harder to hit that area.
> 
> Here's some quick mockup that might resolve this issue:
> 
> image
> 
> Make magnifier symbol "button sized" + add a divider. (Exact spacings, size & divider color TBD)
> 
> Additionally, an ellipsis right of the currently displayed category icons might be a good indicator that there are more of them to the right.
> 
> Edit: Or, if "Search" is just an icon initially, it could be placed as an actual button to the left of the "current icon" button (but still expanding to a full line search in its current place, not in the header).

As I mentioned that seriously dimishes accessibility of the search field - which is the most important in that line. And it'd give priority to the least-important, the category selector which is rarely used (most apps actually completely hide in on scrolling).

I spent quite some time looking into different options and this is the one i settled on. Second option would be a full-size search and the categories below that, similar to how Signal does ot.

moving-bits:
> > Make magnifier symbol "button sized" + add a divider. (Exact spacings, size & divider color TBD)
> 
> As I mentioned that seriously dimishes accessibility of the search field 

Can't see why, if the symbol is enlarged a bit (as proposed) + separated visually from the category item.

Additionally, I had updated my [item](https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638054438) with a second proposal, initiating search by a regular button, see screenshots above.

> which is the most important in that line.  And it'd give priority to the least-important, the category selector which is rarely used (most apps actually completely hide in on scrolling).

That's a matter of personal preferences. I use category selection regularly, both in c:geo and in other apps. And I guess it will be more frequently with having so many more icons.

Please don't get me wrong, being able to search is a great addition and I would like to have it accessible as well, but it's not the only reasonable way to access the icons.

(BTW: "Hiding categories" is done differently in apps, Signal does hide category icons, WhatsApp does not. Just two random examples.)

> I  spent quite some time looking into different options 

I assumed so, and it's highly appreciated!

It's just that we have learned over time, that people are using c:geo (as any other app) in quite different ways, and we should never assume that the way we use it is the only way to go. Even if we finally decide on a single way of implementation, at least in our discussions we should allow to different perspectives.

moving-bits:
Separate finding: Is there a way to exit search function without having to redo from scratch?

image

Currently I don't see a way to exit search function other than using back button, which also exits the whole selection. Being able to return to the full list without having to restart the whole selection would be helpful.

Additionally, the typical "x" at the end of the input field should be added to clear the current input.

moving-bits:
Separate finding: Downward compatibility is broken

After having done some tests with the emoji picker version, I continued work on a different task, being based on DB version 106. Startup failed with this error:

```
[main] DataStore.init: unable to open database for R/W
android.database.sqlite.SQLiteException: Can't downgrade database from version 110 to 106: 110 is not downward compatible
at cgeo.geocaching.storage.DataStore$DbHelper.onDowngrade(DataStore.java:1255)
at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:414)
at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:316)
at cgeo.geocaching.storage.DataStore.initAndCheck(DataStore.java:958)
at cgeo.geocaching.storage.DataStore.init(DataStore.java:932)
at cgeo.geocaching.storage.DataStore.lambda$getStoredLists$72(DataStore.java:4615)
at cgeo.geocaching.storage.DataStore$$ExternalSyntheticLambda87.get(D8$$SyntheticClass:0)
at cgeo.geocaching.storage.DataStore.withAccessLock(DataStore.java:1022)
at cgeo.geocaching.storage.DataStore.getStoredLists(DataStore.java:4613)
at cgeo.geocaching.storage.DataStore.getLists(DataStore.java:4608)
at cgeo.geocaching.CgeoApplication.onCreate(CgeoApplication.java:158)...
```

ztNFny:
> Separate finding: Downward compatibility is broken
> 
> After having done some tests with the emoji picker version, I continued work on a different task, being based on DB version 106. Startup failed with this error:

Well obviously a change in DB schema breaks downwards compatibility?! Or do you want to have a migration routine to convert back and up between old and new picker? Why would anyone downgrade?

moving-bits:
> Well obviously a change in DB schema breaks downwards compatibility?! Or do you want to have a migration routine to convert back and up between old and new picker? Why would anyone downgrade?

Most users won't downgrade ever, except some when switching from nightly or beta version back to a release version. But for us developers it's essential to not break downward compatibility (as otherwise switching between tasks means having to prepare separate emulators or do fresh reinstall).

@eddiemuc back then took great efforts to enable downward compatibility, not by adding downgrade routines, but by letting c:geo know which versions are downward compatible (there is some constant in `DataStore.java` for it) + by (re)writing upgrade statements in a way that they do not fail when returning from a downgraded version back to a newer version. (eg. "CREATE TABLE _IF NOT EXISTS_ ...", but also for adding fields. Therefore an upgrade should not remove fields no longer required (only mark them in field list as such), and never change type of an existing field.

ztNFny:
I'm quite sure that was already the case before, after the list marker changes i tried downgrading but had the incompatible message.
This PR doesn't remove the old data - though I'm not sure how the old code would handle new icons.

moving-bits:
> I'm quite sure that was already the case before, after the list marker changes i tried downgrading but had the incompatible message. This PR doesn't remove the old data - though I'm not sure how the old code would handle new icons.

see `DataStore.DBVERSIONS_DOWNWARD_COMPATIBLE`

Old code will use the int-based emoji field in caches, probably showing no new emoji for those items changed later.
LRU list will break, probably, as will filters, but that would be acceptable in such a downgrade situation, IMHO.

main/src/main/java/cgeo/geocaching/storage/DataStore.java
```diff
@@ -265,7 +266,7 @@ public static String getNameFor(final int id) {
     private static final CacheCache cacheCache = new CacheCache();
     private static volatile SQLiteDatabase database = null;
     private static final ReentrantReadWriteLock databaseLock = new ReentrantReadWriteLock();
-    private static final int dbVersion = 109;
+    private static final int dbVersion = 110;
```
eddiemuc, line 271: 110 must be added to DBVERSIONS_DOWNWARD_COMPATIBLE list which is defined somewhere around line 290

eddiemuc:
> > Well obviously a change in DB schema breaks downwards compatibility?! Or do you want to have a migration routine to convert back and up between old and new picker? Why would anyone downgrade?
> 
> Most users won't downgrade ever, except some when switching from nightly or beta version back to a release version. But for us developers it's essential to not break downward compatibility (as otherwise switching between tasks means having to prepare separate emulators or do fresh reinstall).
> 
> @eddiemuc back then took great efforts to enable downward compatibility, not by adding downgrade routines, but by letting c:geo know which versions are downward compatible (there is some constant in `DataStore.java` for it) + by (re)writing upgrade statements in a way that they do not fail when returning from a downgraded version back to a newer version. (eg. "CREATE TABLE _IF NOT EXISTS_ ...", but also for adding fields. Therefore an upgrade should not remove fields no longer required (only mark them in field list as such), and never change type of an existing field.

I checked, indeed the new version is not in the list. I also overlooked this in my code review yesterday.
Fix is easy though: just need to add 110 to DataStore.DBVERSIONS_DOWNWARD_COMPATIBLE. The changes itself are downward compatible imho (judged from code review only).

eddiemuc:
> I'm quite sure that was already the case before, after the list marker changes i tried downgrading but had the incompatible message. 

For which db version did that happen? (It is part of the error message)

> This PR doesn't remove the old data - though I'm not sure how the old code would handle new icons.

When db version is downgraded always together with the code version that relies on that downgraded db version. So any code changes done in the downgraded version would write into the db the same way as before. When db is upgraded again, it would be along with the code version relying on that version. upgrade migration routines would be executed again and thus should migrate the settings from the "old" version to the "new".
By putting a dbVersion into the DBVERSIONS_DOWNWARD_COMPATIBLE set, you basically confirm that your migration routines can handle this typ of situations.
For the vast majority of migrations this is the case (-> adding a new DB column which allows null values and probably fill that new column with values from "old" db structures ). Don't remember a db version change in the past where this was not possible. 
The downside, of course, is that we accumulate some non-used columns in DB over time. Deleting columns can never be a downward compatible operation of course.

eddiemuc:
@ztNFny will you continue to work on this PR (alone) or would you appreciate contributions to it? (-> pushing commits to your branch)

ztNFny:
> @ztNFny will you continue to work on this PR (alone) or would you appreciate contributions to it? (-> pushing commits to your branch)

i will look at it again once i'm back at a PC ;-) If you want to push changes in the meantime go for it - I also see there's some conflicts with named filter stuff of yours in the meantime

eddiemuc:
> > @ztNFny will you continue to work on this PR (alone) or would you appreciate contributions to it? (-> pushing commits to your branch)
> 
> i will look at it again once i'm back at a PC ;-) If you want to push changes in the meantime go for it - I also see there's some conflicts with named filter stuff of yours in the meantime

OK, I am locally working on some of the code review findings + of course the guava thingy. Please give me a hint before you start working so we don't do duplicate work.

ztNFny:
> > > @ztNFny will you continue to work on this PR (alone) or would you appreciate contributions to it? (-> pushing commits to your branch)
> > 
> > i will look at it again once i'm back at a PC ;-) If you want to push changes in the meantime go for it - I also see there's some conflicts with named filter stuff of yours in the meantime
> 
> OK, I am locally working on some of the code review findings + of course the guava thingy. Please give me a hint before you start working so we don't do duplicate work. 

I think I have some uncommitted changes locally, give me 30 min

ztNFny:
Guava i didn't look at yet

ztNFny:
> > > @ztNFny will you continue to work on this PR (alone) or would you appreciate contributions to it? (-> pushing commits to your branch)
> > 
> > 
> > i will look at it again once i'm back at a PC ;-) If you want to push changes in the meantime go for it - I also see there's some conflicts with named filter stuff of yours in the meantime
> 
> OK, I am locally working on some of the code review findings + of course the guava thingy. Please give me a hint before you start working so we don't do duplicate work.

@eddiemuc pushed

eddiemuc:
I pushed some changes that should resolve migration + compile issues + some comments.

To resolve conflicts I would have to rebase/force-push. Do you want me to do that on your branch?

ztNFny:
Go ahead, not working on it right now

eddiemuc:
> Go ahead, not working on it right now

Can I squash too?

main/src/main/java/cgeo/geocaching/list/StoredList.java
eddiemuc: solved in c7e721dd864702619dc9c02ca996333d5f39b0c0

main/src/main/java/cgeo/geocaching/storage/DataStore.java
eddiemuc: assuing that no one will migrate to any dbVersion from 107 to 109 after this PR is merged, this change can stay.

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
eddiemuc: solved with c7e721dd864702619dc9c02ca996333d5f39b0c0

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
eddiemuc: solved with c7e721dd864702619dc9c02ca996333d5f39b0c0

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
eddiemuc: solved with c7e721dd864702619dc9c02ca996333d5f39b0c0

main/src/main/java/cgeo/geocaching/utils/EmojiUtils.java
eddiemuc: solved

main/src/main/res/layout/dialog_title_button_button.xml
eddiemuc: nitpick

eddiemuc:
The PR in its current state would be fine for me to merge. Open points are:
* commits should be squashed before merge
* not sure if discussion about GUI is resolved between @ztNFny and @moving-bits ?

When continuing to work on this, please note that I force-pushed both my commit and the whole branch several times towards remote!

eddiemuc:
retest this please

moving-bits:
> * not sure if discussion about GUI is resolved between @ztNFny and @moving-bits ?

Not yet. Downgradability and Guava issues seem to be resolved with latest commit, but the following findings seem to be open (haven't cross-checked the latest build, though):

- Search should be reduced to icon initially, then expand to full-line search input field (two different proposals, see https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638054438)
- Search input field should support clear action ("x" at the end of the input line) (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727)
- Search should be cancellable without having to restart selection from scratch (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727)
- Highlighted category should always be visible (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638081980)

ztNFny:
> Search should be reduced to icon initially, then expand to full-line search input field (two different proposals, see https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638054438)

I'm still highly opposed to that idea. Not a single app out there is taking that approach while all (that i checked) are using one of the 2 approaches i suggested (small search in same line as categories (as currently implemented) or full-line search). Can we please a) trust that those "big guys" who actually have budget to do UX research and decided on those implementations know what they are doing and b) not confuse our users with an unfamilar implementation?

imho this PR is good to go - the 3 remaining UI issues are minor and can be addressed later.

@eddiemuc thanks for your support in the cleanup and fixing!

moving-bits:
Sorry, but I have to contradict here:

I don't see the usefulness of an empty 1/3 line besides the search icon, which expands anyway before becoming an input field, while at the same time making the user having to scroll for the category icons. (And I don't know of a wide-spread app using a combination of search field + scrolling category icons in such a way.)

Either make it just a search icon, expanding to a full line search when being tapped on - which would leave enough space for category icons to not have to scroll, or use a separate full line for search (icon + input field) and give the category icons a full line.

ztNFny:
> And I don't know of a wide-spread app using a combination of search field + scrolling category icons in such a way.

e.g. Gboard - the most popular keyboard application out there.

ztNFny:
> Either make it just a search icon, expanding to a full line search when being tapped on - which would leave enough space for category icons to not have to scroll, or use a separate full line for search (icon + input field) and give the category icons a full line.

I did suggest 2 full-width lines, hiding on scroll and you completly ignored that ...

moving-bits:
> I did suggest 2 full-width lines, hiding on scroll and you completly ignored that ...

can you calm down a bit, please? This aggressiveness in tone is tedious (at least it sound aggressive to me).

I'm neither trying to nag you nor do I deny the way you are using the app - I just point out that users are using c:geo in many different ways and have different priorities (as we have in this case, but in others not) - and I'm trying to find a a solution that fits for such different ways of interaction with the app.

Coming back to the UI question:

We are using the "icon + expand to input line on tap" in other places already (list selection, home screen), but if a two line solution is a possible compromise, then I would vote for it.

ztNFny:
> and I'm trying to find a a solution that fits for such different ways of interaction with the app.

Doesn't feel like it at all! I provided reasons why imho the icon-only suggestion is bad in UX terms, gave examples of other apps and how they are doing it and already spent quite some time looking into different UX concepts before even creating this PR, then spent even more time making it actually work. And then you come and say again and again "but it needs to be icon only" without even acknowledging, let alone adressing, any of those arguments.
So yes, I'm highly annoyed by your "but i want it that way", "throw away hours of your work because i want exactly that" attitude here!

moving-bits:
I won't dive into this discussion any further. Maybe you can re-read our respective comments, but change perspectives this time, and think of how I might feel about your reactions to my comments and thoughts.

I'm out of this PR.

ztNFny:
- Replaced the search+categories line with 2 full-width lines and made them scroll
- made the layout more similar to other bottomsheets

image
image
image

ztNFny:
@eddiemuc could you please check that my last force-push didn't mess up things? I rebased so had to force, but somehow the order of commits is different now and i'm worried something got lost.

bekuno:
Error on Jenkins: 
`emoji_picker_dialog.xml:7: error: resource layout/emoji_picker_title (aka cgeo.geocaching.developer:layout/emoji_picker_title) not found.`

ztNFny:
> Error on Jenkins: `emoji_picker_dialog.xml:7: error: resource layout/emoji_picker_title (aka cgeo.geocaching.developer:layout/emoji_picker_title) not found.`

oops, forgot a git add

ztNFny:
There's still an issue with search results being smaller and dimmed on a real device (but not in emulator).

eddiemuc:
> @eddiemuc could you please check that my last force-push didn't mess up things? I rebased so had to force, but somehow the order of commits is different now and i'm worried something got lost.

Tested locally and coudn't find any problems. "My" commit is now second-to-last and a new commit of you is before it. I compared commit order with a previously created backup branch. Naturally all commits have new hashes now and are based on a new root (-> due to rebase + forced-push) but according to the git comments the order is still same as before.

eddiemuc:
you might want to do a squash to clean up PR commit history a bit

ztNFny:
> you might want to do a squash to clean up PR commit history a bit

done. should be ready to go now. Further optimizations can be done in future PRs, let's try this in nightlies to get some feedback.

bekuno:
retest this please

bekuno:
@ztNFny Please check the lint issues. Your PR has 20 new, see https://ci.cgeo.org/job/cgeo-CI_PR-build/15770/androidLintResult/new/. (66 minus `values-*`)

eddiemuc:
After the discussion above I would feel very uncomfortable merging this PR w/o checking back on the points made by @moving-bits 

> - Search should be reduced to icon initially, then expand to full-line search input field (two different proposals, see https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638054438)

This seemed to be the difficult point. Is this solved in a way that is acceptable to both @moving-bits and @ztNFny ? Or can it be adjusted later to such a point?

> - Search input field should support clear action ("x" at the end of the input line) (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727)
> - Search should be cancellable without having to restart selection from scratch (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727)
> - Highlighted category should always be visible (https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638081980)

Those three issues seem to be straightforward. Are those solved?

(I am currently unable to check myself I am afraid)

ztNFny:
> Those three issues seem to be straightforward. Are those solved?

yes

Looking into the linter warnings now, this is mostly removed stuff from the old implementation so should be easy to clean.

ztNFny:
Fixed a bug in the migration logic in regards to filling the Recent Emojis and made the "recently used" limit dynamic so that it's always 2 lines, no matter the screen size

moving-bits:
> > * Search should be reduced to icon initially, then expand to full-line search input field (two different proposals, see [Emoji #18193 (comment)](https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638054438))
> 
> This seemed to be the difficult point. Is this solved in a way that is acceptable to both @moving-bits and @ztNFny ?

Yes from my side, we have a "two line solution" now, search input line stays visible & category icons no longer require scrolling

> > * Search input field should support clear action ("x" at the end of the input line) ([Emoji #18193 (comment)](https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727))
> > * Search should be cancellable without having to restart selection from scratch ([Emoji #18193 (comment)](https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638137727))
> > * Highlighted category should always be visible ([Emoji #18193 (comment)](https://github.com/cgeo/cgeo/pull/18193#issuecomment-4638081980))
> Those three issues seem to be straightforward. Are those solved?

All of those are either implemented meanwhile or no longer necessary (as category icons do no longer scroll). Thanks for the adaptions and fixes.

Found two small nitpicks (see marked with a red border; non-blocking, can be fixed later):

image

- if search field gets focus, the input/divider line gets accent color. This should revert to default color as soon as search field loses focus (eg. by tapping on one of the category items)
- there should be a small blank space between category icon highlight marker and the actual emojis

eddiemuc:
Thanks for the feedback. So content-wise the PR would be merge-ready. Unfortunately it seems to have gathered new codacy issues and merge conflicts meanwhile.

ztNFny:
rebased - the changes in CacheDetailsActivity caused quite a lot of conflicts :-(

eddiemuc:
I am pretty sure this will conflict with PR #18242 of mine. It would only be fair if I solve these conflicts this time, so I would merge here. Ok?

bekuno:
Please have a look at the remaining 5 lint issues.
https://ci.cgeo.org/job/cgeo-CI_PR-build/15819/androidLintResult/new/

ztNFny:
> Please have a look at the remaining 5 lint issues.
> https://ci.cgeo.org/job/cgeo-CI_PR-build/15819/androidLintResult/new/
> 

All 5 are false-positives

.idea/checkstyle-idea.xml
bekuno, line 1: Please revert the change on this file. It is not relevant to your PR.

moving-bits:
On a quick glance, those lint messages fall into one of three categories:
- The PR overrides strings and dimensions that are defined in emojipicker library. If this is intended, the string/dimension should be marked with `tools:override="true"`, eg: `@string/emoji_category_people`
- Unused resources, eg: `R.dimen.emoji_picker_header_icon_width`, `R.dimen.emoji_picker_header_icon_height`
- Errors of type ` StringFormatCount`: We seem to have changed a string from "containing variables" to "no longer containing variables" (not necessarily in this PR, haven't checked), but some translations have not yet been updated accordingly. When removing parameters from an already translated string, the string resource should always get a new name to avoid such errors.

The first two error categories are easy to fix (and should be fixed before merging). If the errors from category three are not caused by this PR, they should be fixed separately.

ztNFny:
> On a quick glance, those lint messages fall into one of three categories:
> 
> * The PR overrides strings and dimensions that are defined in emojipicker library. If this is intended, the string/dimension should be marked with `tools:override="true"`, eg: `@string/emoji_category_people`

All of these strings are in translation-files which are not controlled by this PR - the strings have been removed from values/strings.xml

> * Unused resources, eg: `R.dimen.emoji_picker_header_icon_width`, `R.dimen.emoji_picker_header_icon_height`

These are in fact used - not explicitly in c:geo but in the library below - I added  tools:override="true"

> * Errors of type ` StringFormatCount`: We seem to have changed a string from "containing variables" to "no longer containing variables" (not necessarily in this PR, haven't checked), but some translations have not yet been updated accordingly. 

Not part of this PR

moving-bits:
> > * Errors of type ` StringFormatCount`: We seem to have changed a string from "containing variables" to "no longer containing variables" (not necessarily in this PR, haven't checked), but some translations have not yet been updated accordingly.
> 
> Not part of this PR

Will be fixed with #18244

moving-bits:
> > On a quick glance, those lint messages fall into one of three categories:
> > 
> > * The PR overrides strings and dimensions that are defined in emojipicker library. If this is intended, the string/dimension should be marked with `tools:override="true"`, eg: `@string/emoji_category_people`
> 
> All of these strings are in translation-files which are not controlled by this PR - the strings have been removed from values/strings.xml

That's an interesting issue. So back then when we implemented the old emoji picker, we have used string resources with the same names as emojipicker library does. No problem back then, but now with including the library we have name conflicts even though those strings got removed - probably due to some dangling translations. - If that's the cause, it will be resolved automatically with next Crowdin syncs after merging this PR.

> > * Unused resources, eg: `R.dimen.emoji_picker_header_icon_width`, `R.dimen.emoji_picker_header_icon_height`
> 
> These are in fact used - not explicitly in c:geo but in the library below - I added tools:override="true"

Can you push that change, please? Thx.

ztNFny:
> That's an interesting issue. So back then when we implemented the old emoji picker, we have used string resources with the same names as emojipicker library does. No problem back then, but now with including the library we have name conflicts even though those strings got removed - probably due to some dangling translations. - If that's the cause, it will be resolved automatically with next Crowdin syncs after merging this PR.

The old translations should be pretty much identical to the new strings in the library, so won't be an issue. For english I only noticed because it used different casing.

> Can you push that change, please? Thx.

oops, pushed the wrong branch earlier ...

eddiemuc:
1:0!

eddiemuc:
@ztNFny as expected, meging https://github.com/cgeo/cgeo/pull/18242 causes conflicts in your PR.
I am volunteering to fix remaining issues (checkstyle-config + maybe lint), resolving conflicts, squashing + merging on your PR if you want. Just say the word.

ztNFny:
> @ztNFny as expected, meging https://github.com/cgeo/cgeo/pull/18242 causes conflicts in your PR.
> I am volunteering to fix remaining issues (checkstyle-config + maybe lint), resolving conflicts, squashing + merging on your PR if you want. Just say the word.

Sure, go ahead. Lint + checkstyle is all done though?

eddiemuc:
> > @ztNFny as expected, meging https://github.com/cgeo/cgeo/pull/18242 causes conflicts in your PR.
> > I am volunteering to fix remaining issues (checkstyle-config + maybe lint), resolving conflicts, squashing + merging on your PR if you want. Just say the word.
> 
> Sure, go ahead. Lint + checkstyle is all done though?

Lint is assumingly handled by @moving-bits separate PR, but checkstyle-idea is still part of this PRs change-set

eddiemuc:
all done. If CI is green I will merge

Issue #18196, opened 2026-06-06
Mapping a cache list triggers loadCachesFromGeocodes four times


moving-bits:
## Observations

While analyzing #18108 I noticed that our cache loading implementation for the map triggers a full reload of cache set four times initially, when mapping a cache list.

### Test case:
- Open a cache list with many caches (17k+ in my case)
- Map this list

### Results:
- `DataStore.loadCachesFromGeocodes` gets triggered four times for the whole set of geocodes (some of the calls with slight variations in amount)
- (after some time, `DataStore.loadCachesFromGeocodes` gets called again, with set size 1, seemingly for every single cache (probably just those being visible in the current viewport, haven't cross-checked), but that's not in the focus of this issue, so I stripped that in instrumentation)

### Further observations:
- switching map type (Google => OSM or vice versa) triggers reload only once (seemingly), but does so again after a couple of seconds (and seems to stay at that, then). Same for switching within tileprovider (eg. Google Map => Google Satellite, or between different OSM online providers)
- Number of initial calls can be reduced to two, when guarding the initial calls in `onMapReadyTasks()` and `onResume()` to fire only once initially, see below
- Just opening the map via bottom navigation (not selecting any list along the way) does trigger `loadCachesFromGeocodes` only once, even without the safeguard. But if you select a specific list (long tap on bottom navigation item), it triggers `loadCachesFromGeocodes` twice - this again can be reduced to one by the safeguard below.

**Why does mapping a list from `CacheListActivity` lead to calling `loadCachesFromGeocodes` twice as frequently, and how can it be avoided?**

### DataStore

Add the following lines to the beginning of `loadCachesFromGeocodes`:
```
final int size = geocodes.size();
if (size > 1) {
    Log.e("loadCachesFromGeocodes, size=" + size);
}
```

### UnifiedMapActivity

To reduce the initial calls from four to two, add the following lines:

- Add class member `private boolean skipResumeReloadOnce = false;`
- Add this at the end of `onCreate`: `skipResumeReloadOnce = true;`
- Add switch to `onResume`:
  ```
  if (skipResumeReloadOnce) {
      skipResumeReloadOnce = false;
  } else {
      reloadCachesAndWaypoints();
  }
  ```

PR #18197, opened 2026-06-06
Skip UnifiedMap loading caches twice initially (rel. to 18108)

moving-bits:
## Description
See observations in https://github.com/cgeo/cgeo/issues/18196#issue-4602479489


moving-bits:
This area has changed meanwhile due to the new filtering, so let's close here for now.

Issue #16596, opened 2025-01-26
Event date filter string overflow


SammysHP:
### Describe your problem!

When adding an event date filter, the German (and possibly other) translation overflows.

![Image](https://github.com/user-attachments/assets/74150452-8819-4f7b-a361-a7ccc0dfbf3a)

### How to reproduce?

Add event date filter.

### Actual result after these steps?

Overflow.

### Expected result after these steps?

Something without an overflow.

### Reproducible

Yes

### c:geo Version

2025.01.26-NB-d042942

### System information

```text

```

ztNFny:
this is not really a bug with c:geo but with the german translation. In english the string is short enough, so no need for a general change. If languages don't find a short version they should resort to some other option.

SammysHP:
In my opinion the actual problem is that long text strings aren't suitable as labels for such inputs. Even "past" could be a problem on smaller screens.

ztNFny:
> In my opinion the actual problem is that long text strings aren't suitable as labels for such inputs. Even "past" could be a problem on smaller screens.

This could happen no matter how short strings are. the only realistic general solution I see is leaving out annotations on the axis dynamically

PR #18198, opened 2026-06-06
Fix issue: #16596 by leaving out label which don't fit

Dam297:
## Description

Fix issue: #16596 by leaving out label which don't fit
Update ViewUtils so that text (e.g. label in event date) is not displayed if it does not fit

## Related issues

 #16596


bekuno:
Ok to test

eddiemuc:
Hiding non-fitting labels doesn't seem like a good solution to me. Instead either the labels have to be changed/translated to something that fits or the available lable space has to be extended by some clever layout change.

Thus I vote not to merge this PR.

moving-bits:
Thanks, @Dam297, for your contribution. 
In this case, we won't merge it, though, due to the reasons mentioned [above](https://github.com/cgeo/cgeo/pull/18198#issuecomment-4639483808), which also lists some possible approaches.

PR #18202, opened 2026-06-06
Make CacheCache size dynamic (rel. to #18108)

moving-bits:
Allow user-configured size of CacheCache for a more responsive system (CacheListActivity etc.) with large cache lists.

Opt-in feature for now. Default size of CacheCache is 2000.

Define a new preference key named `cachecache_size` of type `int` and assign either "0" for dynamic cache size allocation, or a value in between the given boundaries (2k - 50k) for a fixed size.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -32,6 +32,7 @@ import cgeo.geocaching.sensors.DirectionData;
 import cgeo.geocaching.sensors.MagnetometerAndAccelerometerProvider;
 import cgeo.geocaching.sensors.OrientationProvider;
 import cgeo.geocaching.sensors.RotationProvider;
+import cgeo.geocaching.storage.CacheCache;
 import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.storage.LocalStorage;
 import cgeo.geocaching.storage.PersistableFolder;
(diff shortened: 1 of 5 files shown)


moving-bits:
retest this please

moving-bits:
retest this please

Lineflyer:
@moving-bits 
Does this change the default behavior (reading "dynamic" makes me think this is a new feature) or is it just a development add-on to make it configurable?

What is the reason to have such preference_key but no UI for it. Is is just for debugging purposes?

moving-bits:
> @moving-bits Does this change the default behavior (reading "dynamic" makes me think this is a new feature) or is it just a development add-on to make it configurable?

It's implemented to not change the default behavior, default size is 2000.

> What is the reason to have such preference_key but no UI for it. Is is just for debugging purposes?

This is done intentionally, for the time being. It's an experimental feature to get better support for large lists. In my tests on a physical device, loading cache lists with some 18000 entries got way faster, like 1.5s instead of 60+ seconds (need to look up the exact times in my recordings), starting with opening the list for a second time, due to increased cache size. With this PR, one can set the cache size to a fixed number of entries (eg. 15000 instead of default 2000), or make let it be dynamic (number of entries calculated from available memory, use "0" value in config for this).

Making it a hidden key gives the team the opportunity to test the behavior extensively for hidden side-effects or other downsides (especially the dynamic calculation) without breaking end-user experience. If everything is fine-tuned and running, we can easily add a UI option for it.

PR #18207, opened 2026-06-07
Parse multiple occurrences of attributes block (fix #18189)

### LINKED ISSUE #18189: Attribute parsing fails if description already contains attributes
### Describe your problem!

If a cache description contains an attribute, c:geo displays this instead of the "real" attribute list.

### How to reproduce?

1. in c:geo open a cache listing with an attribute in the description (e. g. [GCB2ZTQ](https://coord.info/GCB2ZTQ))
2. remember attributes (over here: >10km only) 
3. open cache in browser
4. check attributes (Challenge, Power trail, Stealth, Winter)

in c:geo only one attribute is shown (the one noted in the description), although the listing has 4 attributes

c:geo should show the "real" attributes from the attributes section

### Reproducible

Yes

### c:geo Version

2026.05.27

### System information

```text

```

### Additional Information

_No response_

moving-bits:
## Description
c:geo parses only the first attributes block it finds. If a listing text contains an extra one (like GCB2ZTQ from the related issue), it will parse the one from the listing text, not the actual cache attributes.

This PR extends c:geo to parse all attributes blocks it finds, discarding all but the last found block (in the assumption that all previous blocks are user-supplied via listing text, and therefore to be ignored for cache attributes).

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
@@ -427,25 +427,29 @@ public final class GCParser {
         // cache attributes
         try {
             final List<String> attributes = new ArrayList<>();
-            final String attributesPre = TextUtils.getMatch(page, GCConstants.PATTERN_ATTRIBUTES, true, null);
-            if (attributesPre != null) {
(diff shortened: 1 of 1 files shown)


moving-bits:
retest this please

PR #18210, opened 2026-06-07
rel to #18150: remove foss build (after mlkit was removed in favor of bergamot)

eddiemuc:
rel to #18150: remove foss build (after mlkit was removed in favor of bergamot)

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -198,10 +198,6 @@ android {
             resources.srcDirs = ['src/androidTest/res']
             res.srcDirs = ['src/androidTest/res']
         }
-        // resources specific to flavor "basic".
-        basic {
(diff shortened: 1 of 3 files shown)


JakeDot:
Maybe related: https://github.com/cgeo/cgeo/issues/18298

JakeDot:
Should we really remove foss build? Isn't it the source for f-droid?

moving-bits:
> Should we really remove foss build? Isn't it the source for f-droid?

Before adding ML Kit translation, there was no separate "foss" build, f-droid used our regular release build. On publishing the next _feature_ release, we will return to that procedure (to avoid having to maintain two separate build objects).

eddiemuc:
seems that foss build shall stay. Also PR has conflicts now. Closing here

PR #18212, opened 2026-06-07
Calculated Coordinates: remove support for legacy format

eddiemuc:
Calculated Coordinates: remove support for legacy format

For over 3 years, the not-so-new-any-more Calcluated Coordinates / Variables are live. It is time to remove support for old legacy coordinate format. Any calculated waypoint which was ever created using the old calculator and accessed within last 3 years is already migrated.


SammysHP:
I guess there are many users out there who archive all their finds in c:geo and use it to track additional information, like calculated coordinates. Some of them have way over 50k caches stored in c:geo.

Is it possible to migrate all caches with the old format automatically or at least back it up in some way?

eddiemuc:
> Is it possible to migrate all caches with the old format automatically or at least back it up in some way?

It is not meaningfully possible to make this some sort of database migration routine, the logic is far too complaex. I guess a "Backup" could be made by saving both an old database and an old c:geo version APK. But that's definitely not something for normal users.

We can keep support for this legacy stuff some more time I guess, but at some point there needs to be a desupport date. 
I don't think that anybody (including myself) understands this old format any more. If 3 years is not enough, how many years are?

SammysHP:
I don't think it's about time, it's about loosing data. Can we add all information, that would be lost, to the personal note? Or to the notes of the waypoint.

eddiemuc:
> I don't think it's about time, it's about loosing data. Can we add all information, that would be lost, to the personal note? Or to the notes of the waypoint.

The idea was to give users a grace period (3 years) to save their data. We can leave the legacy stuff in if this doesn't seem enough time.

eddiemuc:
seems that legacy support for calculated coordinates shall stay indefinitely. closing here

PR #18216, opened 2026-06-08
Show favpoints-text-box only for found caches

murggel:
Favorite points can only be awarded for caches that have been found. So don't show for caches, which you haven't found

grafik

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1787,9 +1787,15 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             } else {
                 favoriteLine.layout.setVisibility(View.GONE);
             }
+
             final boolean supportsFavoritePoints = cache.supportsFavoritePoints();
             binding.favpointBox.setVisibility(supportsFavoritePoints ? View.VISIBLE : View.GONE);
             if (!supportsFavoritePoints) {
+                return;
+            }
+            
+            // Add/remove to Favorites is only possible if the cache has been found
+            if (!cache.isFound()) {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
```diff
@@ -1802,12 +1808,6 @@ private void updateFavPointBox(final CacheDetailActivity activity) {
                 binding.removeFromFavpoint.setVisibility(View.GONE);
                 binding.favpointText.setText(R.string.cache_favpoint_not_on);
             }
-
-            // Add/remove to Favorites is only possible if the cache has been found
-            if (!cache.isFound()) {
```
moving-bits, line 1807: Shouldn't this be the `else` branch of `cache.isFavorite()`?

(You could have logged a cache as found, without marking it as favorite at that time, but then later on decide to assign it a favo. This seems no longer possible when this block gets removed. But as `found()` prerequisite is already checked, this leaves this block being just the `!cache.isFavorite()` part = `else` branch.)

moving-bits (changes requested):

main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
murggel: I think, I quite don't get your point. Maybe you mix found and favorite state?

The old logic:
Cache is favorite ->  show addToFavpoint  and favpointText, hide removeFromFavpoint
Cache is not favorite -> hide addToFavpoint, show favpointText and removeFromFavpoint
After this, if cache is not found -> hide addToFavpoint  and removeFromFavpoint, (but don't touches favpointText , so it will be shown)

That leaves only the favpointText visible for a cache, which is not found (don't matter, if it is a favorite).

The new logic just skips the question, if the cache is a favorite and hides all three information (addToFavpoint , removeFromFavpoint and favpointText 


But for more clearity, I will skip the return and put it around the logic for the three information

murggel: Or we use a different text - something like "Only found caches can be added to favorites."

moving-bits: sorry, I had not noticed until now that GitHub was "hiding" some lines from me - the actual `else` branch. For me it looked like some `if` branch, and the possible `else` branch having been deleted. My bad.

Image

moving-bits (approved):
I withdraw the original review, see comment

PR #18235, opened 2026-06-12
Follow-up for PR #18091: Download for wherigo-guids on demand

murggel:
## Description

Show button to start a wherigo, even if the guid is not available yet (due to not fully loaded)
Fetch / load guids on demand / when clicking the button

## Related issues

PR #18091


ztNFny:
This seems to be missing handling of the case that no WIG is linked? Or is this already handled by existing code being called?

murggel:
> This seems to be missing handling of the case that no WIG is linked? Or is this already handled by existing code being called?

Good point - I guess this is not handled yet, but will check and add a corresponding message then

murggel:
> > This seems to be missing handling of the case that no WIG is linked? Or is this already handled by existing code being called?
> 
> Good point - I guess this is not handled yet, but will check and add a corresponding message then

In old implementation the button in cache-detail (and cache-popup) was not shown in that case. 

For a cache of type WHERIGO the button will be always shown now, even if no cartridge-guid is in the listing. This is necessary to show the button for wherigos in the popup (due to to light cache-info the listing is not available and will be lazy-loaded by clicking the wherigo-button). A message will be shown now, if no guid can be found in the listing then.
Same implementation for cache-detail, so if no cartridge is found in the listing, button will be shown nevertheless and also a message is shown.
I could change that, so no button will be shown for cache-detail in case of missing guid. But I like the idea of seeing the button for wherigos...

eddiemuc:
Caching guidelines state:
"A Wherigo cache requires a Wherigo cartridge to find a cache container with a logbook. The cartridge must be hosted on Wherigo.com and the cache description must include a link to the cartridge."
https://www.geocaching.com/play/guidelines

So there should be very few wherigo listings without a cartridge link.

murggel:
retest this please

murggel:
> Caching guidelines state: "A Wherigo cache requires a Wherigo cartridge to find a cache container with a logbook. The cartridge must be hosted on Wherigo.com and the cache description must include a link to the cartridge." https://www.geocaching.com/play/guidelines
> 
> So there should be very few wherigo listings without a cartridge link.

So I kept the dialog in for those rare cases ;-)

JakeDot:
> So there should be very few wherigo listings without a cartridge link.

It seems that rule isn't always enforced, I have two "Mystery" wigs that require solving a riddle before the link is given.

https://coord.info/GC7N54F

murggel:
> > So there should be very few wherigo listings without a cartridge link.
> 
> It seems that rule isn't always enforced, I have two "Mystery" wigs that require solving a riddle before the link is given.
> 
> https://coord.info/GC7N54F

Ok, I will test the PR against this WIG

murggel:
> GC7N54F

If the listing of a WIG (contrary to geocaching.com's guidelines) does not contain a link, the following message will be displayed -  I think that is sufficent
grafikgrafik

murggel:
I would like to give it a try in the nightlies - if no one objects, I will merge this in the next days

eddiemuc:
Go for it

murggel:
retest this please

PR #18239, opened 2026-06-13
Fix button states on list rename (fix #18231)

### LINKED ISSUE #18231: Changing parent list does not enable button
### Describe your problem!

If you want to change the parent list via the "manage lists → rename current list" menu, the "rename" button is only enabled when you change the name, but not when changing other inputs in the dialog.

### How to reproduce?

Open a child list, open menu, manage lists, rename current list, then try changing the parent list.

Rename button to apply changes is disabled. You have to change the name to enable it.

Any change on the dialog should enable the button.

### Reproducible

Yes

### c:geo Version

2026.06.12-NB-cd9289a

### System information

```text

```

### Additional Information

_No response_

moving-bits:
- Update "OK" button enabled state also on parent listname change
- Block "OK" if listname is empty or "Create new parent list" is selected + new parent listname is empty
- Trim values before checking


murggel:
No new bug, but can you remove the ":" from the parent list?
If you select a different one, the ":" is also not shown


grafik

moving-bits:
> No new bug, but can you remove the ":" from the parent list?

have added this to the PR

PR #18258, opened 2026-06-18
fix TB logging (website change)

ztNFny:
adjust to website change

@eddiemuc Code for generating/parsing the requests is different between your Log and my TBLog implementations. This should be harmonized, but I'd rather get this fix out asap and then make it nice later.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
@@ -443,18 +443,30 @@ public class GCLogAPI {
             }
         }
 
-        //URL: https://www.geocaching.com/api/live/v1/logs/TBxyz/trackableLog
-        try (GCWebTrackableLogResponse response = websiteReq().uri("/api/live/v1/logs/" + tbCode + "/trackableLog")
+        final ObjectNode inner = JsonUtils.createObjectNode();
+        inner.put("referenceCode", trackableLog.geocode);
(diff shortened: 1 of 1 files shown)


eddiemuc:
> @eddiemuc Code for generating/parsing the requests is different between your Log and my TBLog implementations. This should be harmonized, but I'd rather get this fix out asap and then make it nice later.

Agree. Let's merge this. Thanks for the fix!

ztNFny:
@eddiemuc  fyi, working on a generic Trpc-wrapper now - to avoid you starting the same work in parallel. Will assign you for review once done.

eddiemuc:
> @eddiemuc  fyi, working on a generic Trpc-wrapper now - to avoid you starting the same work in parallel. Will assign you for review once done.

Sure go ahead.

Note that I used the current logging problem also to switch from an Object-serialization-approach to a JsonNode-oriented approach. But only for the logging methods which were failing. So its currently mixed. If you work in this area you should decide for one approach and change to it completely.

ztNFny:
> > @eddiemuc  fyi, working on a generic Trpc-wrapper now - to avoid you starting the same work in parallel. Will assign you for review once done.
> 
> Sure go ahead.
> 
> Note that I used the current logging problem also to switch from an Object-serialization-approach to a JsonNode-oriented approach. But only for the logging methods which were failing. So its currently mixed. If you work in this area you should decide for one approach and change to it completely.

As we'll probably have to switch everything to the new APIs step-by-step I'd suggest to do that then

Issue #1786, opened 2012-06-14
Batch submit of offline logs to website


Lineflyer:
As  #71 was closed to have separate issues for each point I am opening this issue:

 Batch submit of offline logs to website (not via field-notes).

Submitting the logs by the normal way has the advantage, that
1. The TB activities can be submitted as well
2. The gcvote rating can be sumitted as well
3. The [NUMBER] implementation could be adapted to work normal in this case
4. There is no need to manually "merge" the field-notes on the website

andrixnet:
Regarding trackables, for each cache log, perform the default set trackable action (do nothing, or visited) by loading the existing trackables if necessary and then submitting the actual log(s).

Remember that TB logs go together with cache log, but GK logs are independent.

MagpieFourtyTwo:
Just stumbled upon this thread, cause I in fact just wanted to open a FR with exactly this ... 😎

Although I personally do not really need it (cause I either log one by one online (with GC little helper's help), and if there are bulk logs, I do these with GSAK), but on every single event I visited lately, at least one cacher expressed the desire for a function to batch upload offline logs. They would love to be able to do it with c:geo, cause over there it could be pretty straight forward, does not need a PC - and moreover most of them are not familiar with GSAK (and not really nerdy enough to fiddle with it's peculiarities).

IOW: Use cases are a lot, and users seem to desire it. 😉

So I worked through this thread, and found (just) a few main "problems":
- TB logging 
  - not a real problem, IMHO, not even GSAK supports (an easy way) to log TB action per cache
  - moreover taking, placing and grabbing a TB does not happen that often (compared to the number of "normal" logs), thus I would not call TB actions a deal breaker
- macros/variables like [NUMBER] etc.
  - if we decide that a previously written offline log is mandatory for batch logging, variables like [NUMBER], [OWNER] or alike should not be a problem at all, as these already been replaced in the (manually created) log
- no more ...  😉 ... if you know some, feel free to extend this list ...

TMALSS: I add a +5 (most probably it was even more) to this FR. 🙃

MagpieFourtyTwo:
Post scriptum to even cover the IMHO really seldom case 
- "TB logging action w/o network coverage" or
- "TB actions will not be saved along with offline log"

I would suggest to let "post all offline logs" be a filter dependent function, i. e. just logs of these caches will be posted, which are visible in the currently active filter, as well as in ``Select mode`` just post logs of selected caches.

This way it would be pretty easy to even handle these rare scenario:
[...]

Issue #18249, opened 2026-06-17
Logs not posting (Cache- and TB-Log)


campdave:
### Describe your problem!

Any issues causing logs not to send?

When I click on send log the message below appears.

I have internet connection. Cgeo loads the cache information, the option to award favourite, trackables in inventory. Failure is after the send log button is pressed.

### How to reproduce?

<img width="1146" height="786" alt="Image" src="https://github.com/user-attachments/assets/fa52ed75-9bee-4353-88d0-7e5658ff7175" />

### Actual result after these steps?

Screenshot above

### Expected result after these steps?

Log posted.

### Reproducible

Yes

### c:geo Version

2026.06.04

### System information

```text

```

SammysHP:
Also reported by two users at geoclub.de: https://www.geoclub.de/forum/t/c-geo-aktuelle-bugs.67870/post-1429990

campdave:
System info as requested 

```
## System information

c:geo version: 2026.06.04

Datetime: 17 June at 07:23

Device:
-------
- Device type: motorola edge 60 (scout_g_sysesnp, motorola)
- Available processors: 8
- Android version: 16
- Android build: W1VCS36H.14-20-19-4
- Screen resolution: 1220x2449px (434x871dp)
- Pixel density: 2.8125
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 26.20.31 (260400-919905943)
- Memory:  Available:3.6 GB, Total:11.2 GB, Threshold: 216.0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v11, Count:121
- Set language: en_GB (system default)
- System date format: dd/MM/y
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: never
- Last auto backup: 16 Jun, 08:03
- Routing mode: Walk (shortest.brf / trekking.brf / car-eco.brf / (none) / (none))
- Map mode: UnifiedMap / live / OSM multi-threading: off
- Map: OpenStreetMap.org
[...]

PR #18260, opened 2026-06-19
Batch posting of offline logs

ShiralynDev:
## Description

Adds a "post offline logs" button in "manage caches" in the "stored" menu. This allows batch posting of offline caches.

## Related issues

#1786 

## Additional context

I do not have much java experience so much is based on other functions and might not always be optimal. There is a wait in the posting loop to avoid rate limiting. This time is just a guess and is not tested for I just wanted to be safe when testing the feature.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -20,6 +20,7 @@ import cgeo.geocaching.command.RenameListCommand;
 import cgeo.geocaching.command.SetCacheIconCommand;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.IConnector;
+import cgeo.geocaching.connector.LogResult;
 import cgeo.geocaching.connector.gc.BookmarkListActivity;
 import cgeo.geocaching.connector.gc.BookmarkUtils;
 import cgeo.geocaching.connector.gc.PocketQueryListActivity;
@@ -122,6 +123,7 @@ import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
+import androidx.appcompat.app.AlertDialog;
 import androidx.core.graphics.Insets;
 import androidx.core.util.Consumer;
 import androidx.loader.app.LoaderManager;
@@ -137,6 +139,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -962,7 +965,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
     private void postOfflineLogs(final Collection<Geocache> caches) {
         SimpleDialog.of(this).setTitle(R.string.caches_post_offlinelogs).setMessage(R.string.caches_post_offlinelogs_message).setButtons(SimpleDialog.ButtonTextSet.YES_NO).confirm(() -> {
             progress.show(CacheListActivity.this, null, LocalizationUtils.getString(R.string.caches_post_offlinelogs_progress), true, postOfflineLogsHandler.disposeMessage());
-            postOfflineLogs(postOfflineLogsHandler, caches);
+            postOfflineLogs(this, postOfflineLogsHandler, caches);
         });
     }
 
@@ -1462,7 +1465,7 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
         });
     }
 
-    private static void postOfflineLogs(final Handler handler, final Collection<Geocache> selectedCaches) {
+    private static void postOfflineLogs(final Context context, final Handler handler, final Collection<Geocache> selectedCaches) {
         Schedulers.io().scheduleDirect(() -> {
 
             final List<Geocache> sortedCaches = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


SammysHP:
I haven't looked at the changes in detail, but a few suggestions/questions:

- Is it possible to select a subset of caches? Does it use all offline logs or just logs from the current list?
- How does it handle trackables?
- How does it handle errors? Does it stop or does it skip a log?
- In which order are logs posts? Chronologically?
- Groundspeak appears to implement a method to post multiple logs with a single request. Maybe this new method should be used, although I don't know if the API is stable yet.
- One argument against this method was that we didn't want to encourage logging with a default text. Would it make sense to force opening the logging activity for each cache at least once? Would have to be tracked in the database probably and not a reliable or user friendly approach.

ShiralynDev:
* I think it should use selection
* It does do stuff with inventory for the post log so it might work, it did work with special logs such as reporting faults with the cache
* I believe it skips logs on error
* From my logs it seems to not really have any pattern in the order of posting logs
* I could look in to multiple logs method if you could tell me where to look for these methods / this api
* Could you explain what you mean by "encourage logging with a default text"? I implemented this since the logging was broken whilst I was geocaching and wanted to upload all my saves. Those logs were still written without using a default text. Is there some default text draft feature I don't know of?

ShiralynDev:
Had a hard time testing since I had to upload logs and those had to be real logs of course. Is there any better method of testing against a dummy server or is it best to just comment out the sending and play around with fake logs?

ztNFny:
> I implemented this since the logging was broken whilst I was geocaching and wanted to upload all my saves. 

That actually sounds like you want to upload field notes as drafts? That's posibble with "Export field notes"

SammysHP:
@ShiralynDev 

> I think it should use selection

Well, I suggest verifying that then.

> It does do stuff with inventory for the post log so it might work, it did work with special logs such as reporting faults with the cache

So that means it uses the action chosen while doing the offline log?

> I believe it skips logs on error

If it is happening in the foreground, I suggest to pause the queue and ask the user how to proceed. It's quite possible that the user wants all logs in their correct sequence (e.g. for distance counting TBs).

> From my logs it seems to not really have any pattern in the order of posting logs

I suggest to use the timestamp to sort them before submitting the logs. Same reason as above.

> I could look in to multiple logs method if you could tell me where to look for these methods / this api

See #18249 and associated pull request for details. As I said, maybe not stable yet.

> Could you explain what you mean by "encourage logging with a default text"? I implemented this since the logging was broken whilst I was geocaching and wanted to upload all my saves. Those logs were still written without using a default text. Is there some default text draft feature I don't know of?

c:geo support two things: 1) Setting a default template that is inserted in the log automatically and 2) quick offline log option that allows you to log a cache (offline) with a single click. So this new feature would make it quite easy to mark every cache as found and eventually submit all logs with a default text in the evening.

> Had a hard time testing since I had to upload logs and those had to be real logs of course. Is there any better method of testing against a dummy server or is it best to just comment out the sending and play around with fake logs?

If your changes don't depend on the server communication, I would suggest to disable the actual posting step temporarily. Unfortunately we don't have access to a staging environment to test such features.

Also I've noticed that you included translated strings in your commit. Please remove them and apply them later via crowdin, otherwise they may get lost.

ShiralynDev:
I do have a hard time testing trackables since I don't have any.
I did note that when uploading my caches the problem date would be the postage date and not the date of the log. I'm guessing this might be a geocaching.com limitation since all I do is call the normal createLogTaskLogic function.

Making a todo list to keep track.

- [x] Use selection for caches
- [ ] Verify trackable 
- [x] Pause for log post error
- [x] Post order based on timestamp
- [ ] Look into batch api

ShiralynDev:
I'm quite unsure on how optimal the code is for the dialog but I tried my best.

ShiralynDev:
Holy I really messed up

ShiralynDev:
Ill try to fix the git history...

eddiemuc:
Is there a relation between this PR and https://github.com/cgeo/cgeo/pull/18271?

ztNFny:
I have my concerns about this one. In the past we actively decided against batch-logging - had that changed @cgeo/team ?

If we were to introduce batch-logging i think we need a full workflow to set log texts and TB actions, similar to what GSAK has.

While out caching you seldomly write the full log (and if you do why not just send it?). So for batch-logging there needs to be an option to later apply a log template to all caches - and probably that one isn't a template as we have them today but a text that you write for that specific bunch of caches.
Then for TBs we need to be able to specify the actions per cache. As TB actions are only performed at part of the online log I don't see how that could work with the current abilities at all.

ShiralynDev:
> While out caching you seldomly write the full log (and if you do why not just send it?).

If you lack internet connection or if the c:geo posting is broken.

baiti:
since this is flagged with `feedback required`, here is my feedback: use offline logging, which gives  you a nice fieldnotes record, export it, check "upload to geocaching.com" and you are all set for logging it in a batch on the website, or even better in GSAK. I think this isn't a function that should be even attempted in cgeo ... just my thoughts, just my feedback.

murggel:
> > While out caching you seldomly write the full log (and if you do why not just send it?).
> 
> If you lack internet connection or if the c:geo posting is broken.

I mostly use offline logs while out in the field to mark which I found and the order. Sometimes I already write some notes, sometimes not.
And normally I am not aware of the trackables.
And I usually don't use fieldnotes to write the logs at the computer ..
So for me it would be essential, to have a kind of preview for the logs where I can select, which I want to send and which I want/need to "rework".

ztNFny:
> if the c:geo posting is broken.

that's not something that regulary happens and imho we shouldn't have a feature that just covers another feature being broken

PR #18266, opened 2026-06-20
fix TB log deletion, implement generic TrpcWrapper

ztNFny:
Implements generic classes for the new GC Trpc API request and response bodies.

Creating a body:
final TrpcRequestBody requestBody = new TrpcRequestBody(logId).with("reasonText", reason);
-> {"0":{"referenceCode":"XXX","reasonText":reason}} 

Parsing and using a response:
//Response: [{"result":{"data":{"logReferenceCode":"TLxxx",...}}}]
final TrpcResponseBody response = TrpcResponseBody.of(websiteReq().uri("/api/live/v1/trpc/web.logs.deleteGeocacheLog")....
final String logReferenceCode = response.getText("logReferenceCode");
if (logReferenceCode == null) {

and of course there's at least 1 place (so far) where they break their own "rules" by using "refCode" isntead of "referenceCode" 🙁

### DIFF
--- a/main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
+++ b/main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
@@ -29,8 +29,10 @@ import androidx.annotation.Nullable;
 import androidx.annotation.WorkerThread;
 
 import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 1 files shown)


main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
```diff
@@ -612,4 +598,69 @@ private static String getCsrfToken() {
                 .blockingGet();
         return StringUtils.isBlank(csrfResponse.csrfToken) ? null : csrfResponse.csrfToken;
     }
+
+    // Helper classes for building and parsing JSONs for the TRPC batch endpoints
+    static class TrpcResponseBody {
```
eddiemuc, line 601: This class encapsulates a tree model (JsonNode)....

main/src/main/java/cgeo/geocaching/connector/gc/GCLogAPI.java
```diff
@@ -612,4 +598,69 @@ private static String getCsrfToken() {
...
+            return JsonUtils.getText(data, field, null);
+        }
+
+        @Override
+        public String toString() {
+            return JsonUtils.nodeToString(data);
+        }
+    }
+
+    static class TrpcRequestBody {
```
eddiemuc, line 630: ...while this class uses data binding (serialization directly from Java Object).

eddiemuc (commented):
It looks a bit like you couldn't decide whether to use Data Binding (using serialization/deserialization directly into Java Objects) or Tree Model (using JsonNode and such). Because you use one for request and another for response. I think it would be better to use the same for both. 
(My preference would be tree model, but that's just a personal favorite)

ztNFny:
I kinda prefer annotations and used the JsonNode stuff just to expand where you had converted it already. Did it all with json now - in the end all that matters is that the next few Trpc changes are faster ;-)

eddiemuc:
> I kinda prefer annotations and used the JsonNode stuff just to expand where you had converted it already. Did it all with json now - in the end all that matters is that the next few Trpc changes are faster ;-)

I usdually prefer annotations too and did the whole GCLogAPI stuff with it before. But when I needed to emergency-change this to trpc under time pressure in my lunch break, I noticed that this is much more complex to adapt than when it would have been done with json tree / JsonNode. given our unique situation of "suddenly-changing-APIs-without-documentation" I now think working with trees gives us more flexibility.

ztNFny:
The flexibility in annotations was done with an any-getter/setter. But i don't care which way - and it's even easier to change this later

PR #18268, opened 2026-06-20
Collapse groups in "Delete offline data" dialog by default

### LINKED ISSUE #16688: Collapse groups in "Delete offline data" dialog by default
Extend SimpleDialog with a "setGroupsCollapsedByDefault()" method to show some dialogs with many lines (like the "Delete offline data") in a cleaner state by default.

ztNFny:
Provide the option in SimpleDialog to collapse long lists by default and apply it to the delete offline data dialog
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
+++ b/main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
@@ -461,6 +461,9 @@ public class SimpleItemListView extends LinearLayout {
     }
 
     private boolean isGroupExpanded(final Object group) {
+        if (group == null) {
+            return true;
+        }
         return !model.getGroupingOptions().getReducedGroups().contains(group);
     }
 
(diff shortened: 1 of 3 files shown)


main/src/main/java/cgeo/geocaching/ui/SimpleItemListView.java
```diff
@@ -461,6 +461,12 @@ private void triggerRefreshList() {
     }
 
     private boolean isGroupExpanded(final Object group) {
+        if (group == null) {
+            return true;
+        }
+        if (model.getGroupingOptions().isGroupsCollapsedByDefault()) {
```
eddiemuc: If I understand correctly, you "revert" the meaning of a group being part of the "reduced groups" set: if the new boolean is set then being part of that group means it should be EXPANDED (not reduced). Since reduced group set is empty initially, this creates the desired effect.

But wouldn't that collide with the feature to safe/restore last reduced/expanded group state?

I think it would be better to simply add all existing groups into the "reduced group" set when flag is set, but only if no reduced group set was stored previously.

ztNFny: the dialog where this is used doesn't remember the states - but true, it could. So I've implemented it as 2 options now: Collapsed by default / forced.

moving-bits:
retest this please

PR #18270, opened 2026-06-20
fix problem with version rollover on older builds (eg when switching between release and master)

eddiemuc:
fix problem with version rollover on older builds (eg when switching between release and master)

While working/testing named filter migration problem I installed release on emulator, created some named filters, then switched to master installed/started it. To my surprise AS claims that the app version on emulator (build from release) is newer than the one I want to install (build from master). Investigating this lead to a gradle configuration cache problem where the "current date" is cached. This PR should fix that problem


eddiemuc:
didn't happen again and PR has conflicts not. Closing here

PR #18275, opened 2026-06-21
Fix cache/waypoint infosheet opening delayed (fix #17450)

### LINKED ISSUE #17450: Bottom sheet on map opens with delay
### Describe your problem!

While out in the field with my Android 15 device today I keep having the problem, that on tap on a cache or waypoint the bottom sheet only opened "a bit" and remained there for 1-3 seconds before finally opening normally as it should.
It feels a bit like "it waits" for the next cyclic map update (e.g. change of position).

It does not always happen but in approx. 50% of caases.

### How to reproduce?

Not 100% reproducible, my scenario was:
- Use live map
- Have GPX track loaded on map
- Used single tap to remove header 
- Scroll the map and tap on a cache or waypoint shown

Looks like that for 1-3 seconds until finally continues to scroll up
Image

Normal full size bottom sheet to be shown immediately

### Reproducible

Unclear

### c:geo Version

2025.09-07-NB-136c430

### System information

```text
Android 15
Samsung S24
```

### Additional Information

_No response_

moving-bits:


moving-bits:
Let's try it in the nightlies.

@Lineflyer
Can you test this on your device with next nightly? Thanks

Lineflyer:
@moving-bits 
Will do for sure!

PR #18278, opened 2026-06-21
Fix #18232: Rename prefix: Prefill with parent of current list

### LINKED ISSUE #18232: Rename list prefix menu item not disabled
### Describe your problem!

Once the "rename list prefix" menu item got enabled by accessing a list with a list prefix, the menu item is never disabled again.

### How to reproduce?

1. Open list with list prefix
2. Go to another list without list prefix
3. Open menu → manage lists

There is an item "rename list prefix". When selected, it opens a dialog allowing you to rename the first available prefix.

Menu item should be disabled if the current list does not have a prefix.

### Reproducible

Yes

### c:geo Version

2026.06.12-NB-cd9289a

### System information

```text

```

### Additional Information

_No response_

murggel:
## Description

* Disable if name did not change
* Rename menu-entry from "Rename list prefix" to "Rename parent lists"
* Prefill with parent of current list
* Make it also available for standard lists

## Related issues

Fix #18232

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheListActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheListActivity.java
@@ -629,6 +629,8 @@ public class CacheListActivity extends AbstractListActivity implements FilteredA
             setMenuItemLabel(menu, R.id.menu_drop_caches, R.string.caches_remove_selected, R.string.caches_remove_all, checkedCount);
             MenuUtils.setVisibleEnabled(menu, R.id.menu_drop_caches_all_lists, isHistory || containsStoredCaches, !isEmpty);
             setMenuItemLabel(menu, R.id.menu_drop_caches_all_lists, R.string.caches_remove_selected_completely, R.string.caches_remove_all_completely, checkedCount);
+            MenuUtils.setVisibleEnabled(menu, R.id.menu_remove_from_other_lists, isOffline && listId != PseudoList.ALL_LIST.id, !isEmpty);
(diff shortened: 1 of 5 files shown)


moving-bits:
Some findings:
- Do you want to name the menu item "Rename parent list" (as mentioned in opening post) or "Rename parent lists" (plural, as currently in code)?
- If you change the "rename to" value, but change it back to the original value, "OK" button stays enabled.
- When coming from a standard list (eg: "Stored"), first list prefix is selected (ok), and if I change this to another prefix, "rename to" value is adapted as well (ok). But in that case "OK" button also gets enabled (not ok, "rename to" does not differ from "rename from" yet directly after changing list prefix)

murggel:
> Some findings:
> 
>     * Do you want to name the menu item "Rename parent list" (as mentioned in opening post) or "Rename parent lists" (plural, as currently in code)?

I want the plural form - will change it in opening post

>     * If you change the "rename to" value, but change it back to the original value, "OK" button stays enabled.
> 
>     * When coming from a standard list (eg: "Stored"), first list prefix is selected (ok), and if I change this to another prefix, "rename to" value is adapted as well (ok). But in that case "OK" button also gets enabled (not ok, "rename to" does not differ from "rename from" yet directly after changing list prefix)

Will look into it

murggel:
> > ```
> > * If you change the "rename to" value, but change it back to the original value, "OK" button stays enabled.
> > 
> > * When coming from a standard list (eg: "Stored"), first list prefix is selected (ok), and if I change this to another prefix, "rename to" value is adapted as well (ok). But in that case "OK" button also gets enabled (not ok, "rename to" does not differ from "rename from" yet directly after changing list prefix)
> > ```
> 
> Will look into it

-> fixed (separator was the culprit)

moving-bits (approved):
Did a short retest, LGTM now, thanks for fixing

PR #18286, opened 2026-06-23
name-dialog parent

ztNFny:
2. add "parent" field to filter name dialog, unify implementation with StoredLists


eddiemuc:
I am not really an expert on this whole "parent list" thing. Maybe something for @moving-bits ?
The group icon thingy is fine of course.

One thing to note: all named filters are held in memory the whole time. I don't know how many named filters people will use, I hope it is not in the >100 range...

ztNFny:
@murggel noticed you're working on parent renaming in https://github.com/cgeo/cgeo/pull/18278, so maybe you could review the parent-part of this PR?

murggel:
> @murggel noticed you're working on parent renaming in https://github.com/cgeo/cgeo/pull/18278, so maybe you could review the parent-part of this PR?

Sure, then it's worth if I add some utility-methods for handling of parent-stuff...

murggel:
I hope I will find some time this evening

murggel:
> I hope I will find some time this evening

Looks good to me and I didn't find any unwanted behaviour, just one minor glitch:

|method|new|old|
| --- | --- | --- |
|Rename list: label "list name" is missing|grafik | grafik|
|Save filter: label "filter name" would be helpful| grafik | grafik|

murggel:
> > I hope I will find some time this evening
> 
> Looks good to me and I didn't find any unwanted behaviour
> 

@moving-bits I didn't test it with old list / filter names. Don't know, if there should be some special cases considered, maybe you know?

moving-bits:
I did a short smoke test, focussing on "rename list" functionality. Looks good to me, generally - thanks for integrating parent item functionality into our standard dialog system.

Two minor findings, though, probably related, while testing renaming lists:
- Open "rename list": As soon as you enter a single letter in the "parent list" field you can no longer select one of the available parent lists (even though selection arrow is still shown)
- Creating a new parent lists had a separate "menu" entry "<Create new parent list>" in the past, which, when selected, introduced a separate input box for the new parent list name. This seems to be superseded by reusing the parent list selection field also as an input field.
  - Is this change intended & is it understandable for the user?
  - If this change is intended, having a "x" symbol at the end of the input field would be useful when switching "select list item" to "edit" mode. (And at least string `list_create_parent` could be removed then, haven't checked for possible other candidates.)

ztNFny:
> * Open "rename list": As soon as you enter a single letter in the "parent list" field you can no longer select one of the available parent lists (even though selection arrow is still shown)

typing a letter first filters the list - so typing "a" will still show the lists "a", "aaaa" in the dropdown.

> If this change is intended, having a "x" symbol at the end of the input field

it's still a dropdown

> Is this change intended & is it understandable for the user?

imho (otherwise wouldn't have changed it ;-)) it's clearer than the old - i didn't understand why the user first has to select in an input field for "parent" why he has to select "new", then in a 2nd input specify that new - instead of just specifying the old/new.

moving-bits:
> > * Open "rename list": As soon as you enter a single letter in the "parent list" field you can no longer select one of the available parent lists (even though selection arrow is still shown)
> 
> typing a letter first filters the list - so typing "a" will still show the lists "a", "aaaa" in the dropdown.
> 
> > If this change is intended, having a "x" symbol at the end of the input field
> 
> it's still a dropdown


I guess the combination of the two (filter + dropdown) is the culprit:
- As long as you have not entered anything manually in the "parent list" input field, it works as an (unfiltered) dropdown, and all available parent lists are shown.
- But as soon as you have entered even a single character, it acts as a filtered list. Even if you remove the new character(s), it will act in filtered mode and only ever return to unfiltered mode if you remove all characters - which isn't easy, as there is no "x".

I don't have a quick solution for this at hand, but I think current behavior (in this place) is confusing for the user.

Some ideas:
Possible partial mitigation: Let it return to "unfiltered mode" as soon as the entered text is the same as initially? (Hitting the dropdown icon would then show all entries again).
Or let the input field become it's own entry in the dropdown list (item can be edited further, when selected)?
Or differentiate between filter input and input field for the new parent (like the old implementation did; not necessarily to be done in the same way, though).

ztNFny:
> Even if you remove the new character(s), it will act in filtered mode and only ever return to unfiltered mode if you remove all characters

Not sure what you mean by this? If you have filters "aaa", "ab", "ba" and type "a" only the first 2 will show, you type "aa" -> only "aaa" is shown, you delete the second "a" and "aaa", "ab" will be shown again

> which isn't easy, as there is no "x".

every keyboard ever had a backspace key ;-)

> but I think current behavior (in this place) is confusing for the user.

It shouldn't be...? If you don't want to create a new one then simply don't type a new string - then there's also no need to go back. Like why would you type "c" if you wanted to select one of the existing "aaa", "ab", "ba"? It can't happen accidentally either as dropdown and keyboard are far from each other.

murggel:
> I guess the combination of the two (filter + dropdown) is the culprit:
> - As long as you have not entered anything manually in the "parent list" input field, it works as an (unfiltered) dropdown, and all available parent lists are shown.
> - But as soon as you have entered even a single character, it acts as a filtered list. Even if you remove the new character(s), it will act in filtered mode and only ever return to unfiltered mode if you remove all characters - which isn't easy, as there is no "x".
> 

I stumbled upon that behavior as well, but I think in real usage it would be an edge case. Whether I select directly from the drop-down (filtered or not) or (if there are no filtered entries) I write the new name. 
To be honest for the latter case I would like to have the "x" out of convenience - in most of our input-fields we have one, and for filtering lists we also have the possibility to clear the input wit one click (I guess those are implemented via a dialog and not a real dropdown).
So if it's not much effort to change that, I would appreciate it, but it's not mandatory

ztNFny:
Changed the "new parent" layout:
image
image

ztNFny:
I've unfortunately had quite a lot of conflicts with https://github.com/cgeo/cgeo/pull/18278 as substantial parts of the code touched by that PR are now gone. I think everything should now be covered, but a double-check would be appreciated @murggel

main/src/main/java/cgeo/geocaching/list/StoredList.java
```diff
@@ -483,51 +409,52 @@ public void promptForParentListRename(final int listId, final Runnable runAfterR
...
-
-            parentList.setVisibility(View.VISIBLE);
-            parentList.setHint(R.string.rename_from);
-            parentListView.setText(parentName);
-            parentListView.setAdapter(new ArrayAdapter<>(activity, R.layout.createlist_item, hierarchies));
-            parentListView.addTextChangedListener(ViewUtils.createSimpleWatcher(s -> {
-                ((EditText) menu.findViewById(R.id.title)).setText(s);
+            fromView.setInputType(InputType.TYPE_NULL);
+            fromView.setAdapter(new ArrayAdapter<>(activity, android.R.layout.simple_dropdown_item_1line, hierarchies));
+            fromView.setText(hierarchies.get(0), false);
```
murggel:
```suggestion
            final StoredList list = DataStore.getList(listId);
            final String parentName = StringUtils.defaultIfEmpty(getGroupFromList(list, null), hierarchies.get(0));
            fromView.setText(parentName, false);
```

If a list is selected, the values should be prefilled with the parent of the current list.

main/src/main/java/cgeo/geocaching/list/StoredList.java
```diff
@@ -483,51 +409,52 @@ public void promptForParentListRename(final int listId, final Runnable runAfterR
...
+
+            viewBinding.inputNameLayout.setHint(activity.getString(R.string.rename_to));
+            toView.setText(hierarchies.get(0));
```
murggel:
```suggestion
            toView.setText(parentName);
```

If a list is selected, the values should be prefilled with the parent of the current list.

murggel:
> I've unfortunately had quite a lot of conflicts with #18278 as substantial parts of the code touched by that PR are now gone. I think everything should now be covered, but a double-check would be appreciated @murggel

Sorry - missed to mark the PR as WIP ...

I have two change-suggestions - see review comments...

murggel (changes requested):

murggel:
Merge conflicts seem successfully resolved.

main/src/main/java/cgeo/geocaching/ui/dialog/SimpleDialog.java
```diff
@@ -700,6 +721,162 @@ public void input(final InputOptions options, final Consumer<String> okayListene
...
+        if (input == null) {
+            return "";
+        }
+        return Arrays.stream(input.split(":", -1))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.joining(":"));
+    }
+
+    public void inputWithParent(final InputOptions nameOptions, @NonNull final List<String> parentChoices, final Consumer<String> okayListener) {
```
eddiemuc: Imho the "parent selection" logic is too specific to go into SimpleDialog. I suggest to put it in a util class (eg hierarchicallistutils or such).

Simpledialog offers "setCustomView" to extend any dialog with a custom view. This can be used for the GUI part. This can be used also from a utils class since it is public (on purpose).

ztNFny: sorry, but i can't figure out how to achieve what you're asking for - how you intend SimpleDialog to be extended. Claude failed as well ...

I'll split "in named filters dialog use the first child items icon for the group (same as in StoredLists)" into a separate PR so this simple fix can get merged quickly.
Then the new filter implementation will still be missing the parent selection part, but that can be worked around by manually specifying the name with ":" as separator. If someone wants to take it up as a "clean" implemention as intended in SimpleDialog please go for it.

eddiemuc: My Claude doesn't have problems understanding this :-)
It is about creating clearer responsibilities:
* SimpleDialog remains a generic Dialog Builder without any business knowledge
* specific Dialogs are constructed using it, but in other classes (could be eg Dialogs, could be a new class eg "HierarchicalDialogHelper")

But I don't inisist on this. This is community code after all. I am not blocking the change.

ztNFny: I guess your prompts are better than mine then - probably as you already understand quite clearly what you want to do. 
Let's take your Claudes implementation than rather than this one.

eddiemuc: I asked Claude for an evaluation, I didn't carry out implementation.

As I said earlier, the list-creation-with-parent thing is not really m area kf expertise, I would rather leave this PR to the experts

murggel:
> I hope I will find some time this wee to have a look at the workflow and give some feedback

I think from users perspective it is not quite obvious to type a colon to create sublists. 
So I would vote for the new design.
From my point of view, bBoth approaches (PR #18286 and PR #18317) seem to struggle with distinguishing between defining a new parent and selecting an existing one - so why not opening a dialog when clicking on the "+"-button to enter the new name and preselect it then in the drop down. 

Maybe it has technical reasons, because the pulldown is filled with only existing lists?

eddiemuc:
> I think from users perspective it is not quite obvious to type a colon to create sublists. 
> ...why not opening a dialog when clicking on the "+"-button to enter the new name and preselect it then in the drop down. 

Let's assume that users don't want to type colons. Let's further assume that a user maintains a three-level list hierarchy eg distinguishing caches by "needed eta" and "preferable season" with sub-categories. User has a list hierarchy like this:

- preferable season
   - spring 
      - March
      - April
      - May
   - summer
      - June
      - July 
- needed eta
   - climbing gear
      - rope
      - crampon
   - fishing rod
      - the long one
      - the short one 

(I don't have a real-world example, I am not using parent lists myself :-))

He gets a parent selection list like this:

```
preferable season
preferable season:spring 
preferable season:summer
needed eta
needed eta:climbing gear
needed eta:fishing rot
```

He wants to add a new needed eta "ladder" with lists "big" and small"

With the current approach he would:
- select "preferable season"
- select +, gets a new text field 
- types "ladder" in the text field
- types the list name "big" and saves

For the second list he would be able to select "needed eta:ladder" and just type "small"

--> no typing of a colon needed

With the proposed approach as well as the approach in the other PR, the user has to type "needed eta:ladder" in the "new parent" text field, thus typing a colon. Also the user has to type the root parent name again (needed eta) and be careful not to mistype there.

This is true for all new parents which are not on the root level.

> Maybe it has technical reasons, because the pulldown is filled with only existing lists? 

No that's doable. Little bit more complex but no blocker

eddiemuc:
Updated my comment

ztNFny:
> think from users perspective it is not quite obvious to type a colon to create sublists.

Is a 2nd level of lists really needed though? Is anyone using that??

> From my point of view, bBoth approaches (PR #18286 and PR #18317) seem to struggle with distinguishing between defining a new parent and selecting an existing one

Did you see the last implementation there? Can't really be clearer than that imho ... An input field for "new parent" and nothing else on the screen.

https://github.com/cgeo/cgeo/pull/18286#issuecomment-4830948126

murggel:
> With the proposed approach as well as the approach in the other PR, the user has to type "needed eta:ladder" in the "new parent" text field, thus typing a colon. Also the user has to type the root parent name again (needed eta) and be careful not to mistype there.
> 
> This is true for all new parents which are not on the root level.
> 
> 


Fair, maybe I missed that 🙈

murggel:
> > think from users perspective it is not quite obvious to type a colon to create sublists.
> 
> Is a 2nd level of lists really needed though? Is anyone using that??

I am playing around with deeper hierarchical lists, so maybe yes...

> > From my point of view, bBoth approaches (PR #18286 and PR #18317) seem to struggle with distinguishing between defining a new parent and selecting an existing one
> 
> Did you see the last implementation there? Can't really be clearer than that imho ... An input field for "new parent" and nothing else on the screen.
> 
> https://github.com/cgeo/cgeo/pull/18286#issuecomment-4830948126


Yes I saw that, but changing between selecting an existing list and creating a new is not really obvious and creating a new sublist also is not really clear to me. 

My approach was to just move the input field in an extra dialog to have the possibility to show the new hierarchical list together with the existing ones

murggel:
> > With the proposed approach as well as the approach in the other PR, the user has to type "needed eta:ladder" in the "new parent" text field, thus typing a colon. Also the user has to type the root parent name again (needed eta) and be careful not to mistype there.
> > 
> > This is true for all new parents which are not on the root level.
> > 
> > 
> 
> 
> Fair, maybe I missed that 🙈
> 

Hm - but if that is used as sublist to the current selected one, no colon is needed  - or am I missing something?

My brain is a little slow this week...

eddiemuc:
> Hm - but if that is used as sublist to the current selected one, no colon is needed  - or am I missing something?

The problem arises only when you want to create/move a list under a new parent where the new pare t shall not be on root level but under an existing parent.

ztNFny:
Can we please first align on whether n-level lists (vs 2-level) is even a wanted feature? (To have it in the GUI - functional-wise it's technically already possible since forever with manual using colon).
Most of the discussion is around implementing a nice GUI for n-level - not having that would greatly simplify the requirements for that dialog.

Imho more than 2 levels of lists is not useful, handling it on that small screen is cumbersome, navigating lists takes time. It's a lot easier to work with more 1st level folders than creating 2nd level folders - let alone 3rd+

murggel:
For me there is no reason to not allow n-level-hierarchical lists and hence to restrict the user.
So maybe we have to discuss, if we want a nice UI for that or just live with typing the colon.

murggel:
Assuming that you're going to stone me, but maybe we also have to discuss, if we want to find a different UI / unify the workflow for creating a new list then - this is currently an entry in the list-selection, as "new parent list" is an entry in the parent-list-selection.
So why changing it in one place but not in the other....

ztNFny:
> Assuming that you're going to stone me, but maybe we also have to discuss, if we want to find a different UI / unify the workflow for creating a new list then - this is currently an entry in the list-selection, as "new parent list" is an entry in the parent-list-selection.
> So why changing it there but not here?

That's the whole point of this PR - to have 1 unified UI for lists/filters - and any future place where parents could be needed.

ztNFny:
> For me there is no reason to not allow n-level-hierarchical lists and hence to restrict the user.

That wasn't the question 😉
> So maybe we have to discuss, if we want a nice UI for that or just live with typing the colon.

Exactly that is what I asked - do we need a UI for CREATING multi-nested lists/filters.

murggel:
But this is not touched in both PRs, correct?

eddiemuc:
> But this is not touched in both PRs, correct?
> 
> 

Correct, this is not touched. I assume your earlier comment referneced this? (and not something in the "create/rename list/filter item" dialogs)

murggel:
> > But this is not touched in both PRs, correct?
> > 
> > 
> 
> Correct, this is not touched. I assume your earlier comment referneced this? (and not something in the "create/rename list/filter item" dialogs)
> 
> 

Correct, that I was referring when talking about the workflow for creating a new list

ztNFny:
ok, misunderstood you then. So yes, this PR refines the workflow for creating a new list/filter - it doesn't change where this workflow is started from. And to be honest I wouldn't want to put this into the same PR, seeing as how complex it already is.

ztNFny:
If we start talking about separating "Create new list" then we should also talk about the "Select all" and "Select visible" pseudo-entries - but I'd really like to separate that discussion.

murggel:
Fine for me to put it in a separate PR, but maybe it is worth to discuss it (the create new entry) together with this topic, for me it seems quite related - it's about creating new entries in addition to selecting existing ones...


The discussion about the select-entries may be a different topic, but when discussing about changing the workflow / UI, why not considering those as well. 
Implementation is a different part

ztNFny:
fwiw: We should add the "icon" selection to the dialog because it's used in both places that currently offer parent selection and it makes perfect sense to define that when creating the list/filter

ztNFny:
> fwiw: We should add the "icon" selection to the dialog because it's used in both places that currently offer parent selection and it makes perfect sense to define that when creating the list/filter

added it to this PR:
image

The concern by eddiemuc about code being in the SimpleDialg instead of a utility class aren't yet resolved - and I'm still stuck there.
Maybe we can finish the UI discussion here, then implement it "clean" in eddies branch?

moving-bits:
@ztNFny 
I guess this PR is superseded by #18430?

PR #18300, opened 2026-07-01
Persist metered network checkbox in download dialog (fix #18299)

### LINKED ISSUE #18299: Make "download over metered connection" persistent
### Describe your problem!

I don't have WiFi, ever, so having to check the checkmark every time is error-prone as if left unchecked, the downloads pile up unable to ever complete.

I ask for a setting or change in behaviour to make the choice persist over different download dialog instances.

Image

### How to reproduce?

Wait for update or trigger manually a download of a map, theme, ... 

Metered connection has to be selected every time

Connection status should be remembered 

### Reproducible

Yes

### c:geo Version

nightly

JakeDot:
Add a "Remember metered network choice" setting (off by default) in offline data preferences. When enabled, the "allow download over metered network" checkbox in the download dialogs is pre-filled with the last chosen value and updated whenever a download is confirmed.

Claude-Session: https://claude.ai/code/session_01Bz9xV22zmC3D9EZ7ygy5SX

## Related issues
[Make "download over metered connection" persistent #18299](https://github.com/cgeo/cgeo/issues/18299)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/downloader/DownloaderUtils.java
+++ b/main/src/main/java/cgeo/geocaching/downloader/DownloaderUtils.java
@@ -175,10 +175,12 @@ public class DownloaderUtils {
         builder.setView(binding.getRoot());
         binding.downloadInfo1.setText(LocalizationUtils.getString(R.string.download_confirmation, StringUtils.isNotBlank(additionalInfo) ? additionalInfo + "\n\n" : "", filename, "\n\n" + LocalizationUtils.getString(R.string.download_warning) + (StringUtils.isNotBlank(sizeInfo) ? "\n\n" + sizeInfo : "")));
         binding.downloadInfo2.setVisibility(View.GONE);
+        binding.allowMeteredNetwork.setChecked(Settings.getDownloadAllowMeteredNetwork());
 
         builder
                 .setPositiveButton(android.R.string.ok, (dialog, which) -> {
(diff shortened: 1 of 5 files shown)


bekuno:
Retest this please

JakeDot:
Fixed Gemini CR comment

https://github.com/JakeDot/cgeo/pull/282#discussion_r3503793752

ztNFny:
I don't think this should be an extra setting, there's already too many of them.

How about this in terms of UI: In the download dialog show 2 checkboxes:
1. Download over metered network
2. Default to downloading over metered network

2nd is disabled +unchecked by default and only enabled if you check the 1st it gets enabled. If the user then checks it the default is stored as a setting. Unchecking box 1 OR 2 in any future dialog also resets the setting.

moving-bits:
I'd also vote for not adding extra UI elements for persisting the setting.

My proposal would be to simply remember the last state of the toggle (default = unchecked), similar to what we do with map live mode.

JakeDot:
I removed the settings GUI and implemented a pure memory for the checkbox.

JakeDot:
Could this be merged to master for me to test?

ztNFny:
> Could this be merged to master for me to test? 

You should test every change with local builds, before making the PR.

JakeDot:
> You should test every change with local builds, before making the PR.

Got it thanks 👍🏼

moving-bits (approved):
Did a quick cross-check, LGTM

PR #18307, opened 2026-07-05
Add pseudo bookmark list for own unpublished caches

JakeDot:
Adds an "Own Unpublished Caches" entry to the Bookmark Lists screen that scrapes my/geocaches.aspx?archived=y instead of the real bookmark-list API, letting users pull in their own not-yet-published (or de-published) geocaches as a cache list.

Claude-Session: https://claude.ai/code/session_01LG5LEBM12ancuVYYTzz9R1


moving-bits:
retest this please

Issue #18120, opened 2026-05-21
Ignore list


novw:
(edit: typos) 

### Tell us your idea!

I have recently come to understand the joys of the "ignore" list. C:geo understands the importance of that list. Unfortunately adventure caches do not have the option to ignore them. There are a couple of adventures (in Amsterdam) I would really love to ignore. Is an internal c:geo ignore list feasible? 

### What solution would you suggest?

Instead of "caches around / open in browser / share cache / translate" maybe add an "ignore cache" option in the quick view?

ztNFny:
You can already add them to a list and then create a filter to NOT show caches on that list.

novw:
> You can already add them to a list and then create a filter to NOT show caches on that list.

Actually I think not. The "stored list" filter has no negative setting. So showing all caches not on a list is impossible. So I can not filter to show all caches (including live). All caches would have to be downloaded in a seperate list. I can set "inverse filter" but than for me to see all caches NOT found, I need to set that filter to show "found". Another example : to show opencaching caches I need to check all boxes _except_ opencaching... "inverse filter" is famous for *all* filter selections. That's a "break your brain" situation.

eddiemuc:
> > You can already add them to a list and then create a filter to NOT show caches on that list.
> 
> Actually I think not. The "stored list" filter has no negative setting. So showing all caches not on a list is impossible. So I can not filter to show all caches (including live). All caches would have to be downloaded in a seperate list. I can set "inverse filter" but than for me to see all caches NOT found, I need to set that filter to show "found". Another example : to show opencaching caches I need to check all boxes _except_ opencaching... "inverse filter" is famous for *all* filter selections. That's a "break your brain" situation. 

You can use a nested filter. In that nested filter you can configure a "Stored list" filter with you personal ignore list and invert it.
(Nested filters allow arbitrarily complex nested filter constructions)

But I agree that for ignoring caches an easier-to-use concept would be meaningful.

novw:
2026.05.22-NB-4912239 : I have a few ("real") caches on the official online ignore list, yet they show up on c:geo map?

JakeDot:
The official ignore list is not the same as the user-defined individual ignore list.

novw:
[...]

PR #18310, opened 2026-07-07
Ignore list as a (pseudo) c:geo list 

JakeDot:
The GC.com ignore feature previously only supported one-way "ignore" (via the legacy ignore.aspx endpoint), with no way to see or undo it from within c:geo. The online ignore list is actually an auto-created bookmark list ("Ignore Listing"), reachable through the same GC.com Lists API c:geo already uses for regular bookmark lists.

This adds a reserved local list (StoredList.IGNORE_LIST_ID) that mirrors
that bookmark list:

1.    On first opening the Ignore list, its content is fetched via the GC    Lists API and stored locally (IgnoreListUtils.syncOnlineIgnoreListIfNeeded).
2.    Ignoring/un-ignoring a cache (from the cache detail menu, from bulk    list actions, or by removing an entry from the Ignore list itself)    calls through to the connector and keeps the local list in sync. GCConnector now implements removeFromIgnorelist and the new IIgnoreListCapability, backed by GCParser bookmark-list add/remove    and a new Network.deleteJsonRequest(headers, body) overload.

Co-Authored-By: Claude Sonnet 5 noreply@anthropic.com
Claude-Session: https://claude.ai/code/session_01Lg5ZLLRqK8CCrcQmHNwnQr

<!-- Examples: - Fixes a NULL check i


JakeDot:
See https://github.com/JakeDot/cgeo/pull/291

JakeDot:
Addresses some of https://github.com/cgeo/cgeo/issues/18120

JakeDot:
retest this please

PR #18316, opened 2026-07-11
filter list group icons

ztNFny:
in named filters dialog use the first child items icon for the group (same as in StoredLists)
image

### DIFF
--- a/main/src/main/java/cgeo/geocaching/filters/FilterUtils.java
+++ b/main/src/main/java/cgeo/geocaching/filters/FilterUtils.java
@@ -152,7 +152,7 @@ public class FilterUtils {
                 }
                 return TextParam.text(name);
             }, (f, gi) -> f.getName(), null)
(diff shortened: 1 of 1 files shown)


Lineflyer:
Might be a stupid question, but in your screenshot above the second group does in fact not use the icon of the first child. Does it?

ztNFny:
> Might be a stupid question, but in your screenshot above the second group does in fact not use the icon of the first child. Does it?

It does - "b:a" has no icon assigned, so both filter and folder use the default icon. Same behaviour as for lists.

Lineflyer:
Ah, I wasn't aware, that this star/circle icon is for filtering. Need to get used to this new filter UI.

ztNFny:
> Ah, I wasn't aware, that this star/circle icon is for filtering. Need to get used to this new filter UI.

We maybe should change the icon back to a filter again, as we did for the labels.

eddiemuc:
> > Ah, I wasn't aware, that this star/circle icon is for filtering. Need to get used to this new filter UI.
> 
> We maybe should change the icon back to a filter again, as we did for the labels.

I changed the ixon in todays pr

moving-bits:
Any objections to merging content-wise?

@ztNFny 
As we have branched for next feature release meanwhile, can you retarget this to `release`, please?

PR #17498, opened 2025-09-28
Allow creation of new parent list in new list dialog (fix #17190)

moving-bits:
## Description
Allow creation of a new parent list in "New list" / "Rename list" dialogs:

image.image.image.image

PR #18317, opened 2026-07-12
redesign parent list select for reuse in namedfilters: approach eddiemuc

eddiemuc:
redesign parent list select for reuse in namedfilters: approach eddiemuc

Ok, I couldn't keep my fingers from it. This is my own approach to handle the "create/rename list" parent thingies. Idea is to provide it in a way that it can be reused for other places where hierarchical lists are used, esp for namedfilters.

The approach is far from ready! This is to gather feedback whether I should continue this or abort.

Key class is the new HierarchicalListHelper to provide helpers for entities using hierarchical lists 

Features:
* reusable hierarchical list helpers (ready for usage in storedlists and named filters)
* allows renaming/creating with selecting existing parent or add a new parent
* allows edit/creation of markers in same dialog

Create dialog initial:
image

Parent selection:
image

Add parent ("plus" button) + add marker:
image

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -31,3 +31,6 @@ android.useAndroidX=true
 
 # For local instrumented tests: leave original c:geo installation intact on device
 android.injected.androidTest.leaveApksInstalledAfterRun=true
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
--- a/main/src/main/java/cgeo/geocaching/list/StoredList.java
+++ b/main/src/main/java/cgeo/geocaching/list/StoredList.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.activity.Keyboard;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
+import cgeo.geocaching.ui.HierarchicalListHelper;
 import cgeo.geocaching.ui.ImageParam;
 import cgeo.geocaching.ui.SimpleItemListModel;
 import cgeo.geocaching.ui.TextParam;
@@ -42,6 +43,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -94,6 +96,38 @@ public final class StoredList extends AbstractList {
         private final WeakReference<Activity> activityRef;
 
         public static final String GROUP_SEPARATOR = ":";
+
+        private static final HierarchicalListHelper.HierarchyAccessor<AbstractList> HIERARCHY_ACCESSOR = new HierarchicalListHelper.HierarchyAccessor<AbstractList>() {
+
+
+            @NonNull
+            @Override
+            public String getFullName(final AbstractList item) {
+                return item.getTitle();
+            }
+
+            @Override
+            public TextParam getDisplayText(final AbstractList item, final String simpleName) {
+                if (item instanceof StoredList) {
+                    return TextParam.text(simpleName + " [" + ((StoredList) item).count + "]");
+                }
+                return TextParam.text(item.getTitleAndCount());
+            }
+
+            @Override
+            public boolean supportsMarkers() {
+                return true;
+            }
+
+            @Nullable
+            @Override
+            public String getMarker(final AbstractList item) {
+                if (item instanceof StoredList) {
+                    return ((StoredList) item).emojiMarker;
+                }
+                return null;
+            }
+        };
 
         public UserInterface(@NonNull final Activity activity) {
             this.activityRef = new WeakReference<>(activity);
@@ -226,7 +260,7 @@ public final class StoredList extends AbstractList {
             return CommonUtils.getNullHandlingComparator((p1, p2) -> collator.compare(getSortString(p1, selectedIds), getSortString(p2, selectedIds)), true);
         }
 
-        private String getSortString(final Object item, final Set<Integer> selectedIds) {
+        private static String getSortString(final Object item, final Set<Integer> selectedIds) {
             //Stored list sorting:
             // A. If displayed: "Create new"
             // B. If displayed: "Stored",
@@ -332,13 +366,13 @@ public final class StoredList extends AbstractList {
 
         public void promptForListCreation(@NonNull final Action1<Integer> runAfterwards, final String newListName) {
             // We need to update the list cache by creating a new StoredList object here.
-            handleListNameInput(newListName, R.string.list_dialog_create_title, R.string.list_dialog_create, listName -> {
+            handleListNameInput(null, R.string.list_dialog_create_title, R.string.list_dialog_create, (listName, marker) -> {
                 final Activity activity = activityRef.get();
                 if (activity == null) {
                     return;
                 }
                 final int newId = DataStore.createList(listName);
-                new StoredList(newId, listName, EmojiUtils.NO_EMOJI, false, 0);
+                new StoredList(newId, listName, marker, false, 0);
 
                 if (newId >= DataStore.customListIdOffset) {
                     runAfterwards.call(newId);
@@ -350,13 +384,14 @@ public final class StoredList extends AbstractList {
 
         public void promptForListCreation(@NonNull final Action1<Set<Integer>> runAfterwards, final Set<Integer> selectedLists, final String newListName) {
             // We need to update the list cache by creating a new StoredList object here.
-            handleListNameInput(newListName, R.string.list_dialog_create_title, R.string.list_dialog_create, listName -> {
+            handleListNameInput(null, R.string.list_dialog_create_title, R.string.list_dialog_create, (listName, marker) -> {
                 final Activity activity = activityRef.get();
                 if (activity == null) {
                     return;
                 }
                 final int newId = DataStore.createList(listName);
-                new StoredList(newId, listName, EmojiUtils.NO_EMOJI, false, 0);
+                DataStore.setListEmoji(newId, marker);
+                new StoredList(newId, listName, marker, false, 0);
 
                 if (newId >= DataStore.customListIdOffset) {
                     selectedLists.remove(PseudoList.NEW_LIST.id);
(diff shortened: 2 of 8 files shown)


ztNFny:
Thx for taking a stab at this - sure looks cleaner.

> Add parent ("plus" button) + add marker: image

imho the "delete"-icon is wrong here as it doesn't exist yet. And I find the 2 fields (existing list + new list) confusing. In my implementation it's a toggle between existing and new so that always only 1 of them is shown.

eddiemuc:
> Thx for taking a stab at this - sure looks cleaner.
> 

> imho the "delete"-icon is wrong here as it doesn't exist yet. And I find the 2 fields (existing list + new list) confusing. In my implementation it's a toggle between existing and new so that always only 1 of them is shown.

I assumed the intention of the original is that user can select an existing (sub)group and can then create another subgroup in it in which she then places the new list/filter.

ztNFny:
> > Thx for taking a stab at this - sure looks cleaner.
> > 
> 
> > imho the "delete"-icon is wrong here as it doesn't exist yet. And I find the 2 fields (existing list + new list) confusing. In my implementation it's a toggle between existing and new so that always only 1 of them is shown.
> 
> I assumed the intention of the original is that user can select an existing (sub)group and can then create another subgroup in it in which she then places the new list/filter.
> 
> 

No, it's that you either select an existing or create a new parent. There's no GUI for 3+ level lists.

eddiemuc:
> > > Thx for taking a stab at this - sure looks cleaner.
> > > 
> > 
> > > imho the "delete"-icon is wrong here as it doesn't exist yet. And I find the 2 fields (existing list + new list) confusing. In my implementation it's a toggle between existing and new so that always only 1 of them is shown.
> > 
> > I assumed the intention of the original is that user can select an existing (sub)group and can then create another subgroup in it in which she then places the new list/filter.
> > 
> > 
> 
> No, it's that you either select an existing or create a new parent. There's no GUI for 3+ level lists.

Then that's a new feature of my PR too :-)
Anyway, can be discussed if this is wanted

eddiemuc:
> > Thx for taking a stab at this - sure looks cleaner.
> 
> > imho the "delete"-icon is wrong here as it doesn't exist yet. And I find the 2 fields (existing list + new list) confusing. In my implementation it's a toggle between existing and new so that always only 1 of them is shown.
> 
> I assumed the intention of the original is that user can select an existing (sub)group and can then create another subgroup in it in which she then places the new list/filter.

To make it more visible that the "new parent" can only be used one time I introduced disabled icon buttons (I missed this for a long time now anyway):

image

The screenshotted dialog would create a list `tiefer:noch tiefer:und noch tiefer:Lieblingsliste`

ztNFny:
So both parent fields are active and used in combination? That's a deviation from the old UI where only 1 field was used - if "new parent" was filled only that one was used.
As probably not many people use multi-level lists I'm not sure we should provide the UI for that - and it could lead to accidental multi-level creations.

Curious to hear what others think about this.

eddiemuc:
> So both parent fields are active and used in combination? That's a deviation from the old UI where only 1 field was used ...

 I assume you mean current release (master states are not in circulation except for nightly testers, I consider master as "in development"): this doesn't even have a "new parent" field, so this is completely new. Still it already allowed creation of subparents under existing parents (only not in a separate textfield):



> Curious to hear what others think about this.

Same. Who introduced the separate "new parent" field?

ztNFny:
We've long had a parent-dialog for lists, I'm referring to that. That's also why my PR unified lists and filters.

You could of course always define multiple levels manually using the separators - and imho that's good enough.

murggel:
I hope I will find some time this wee to have a look at the workflow and give some feedback

eddiemuc:
> We've long had a parent-dialog for lists, I'm referring to that.

But we are talking about the "new parent" extra text field. That one is only in dev.

> You could of course always define multiple levels manually using the separators - and imho that's good enough.

I would understand this gui logic. But why was an extra "new parent" field introduced at all in that case? (Prod state is good enough for this) Only reason I can think of is to keep the user from typing a colon.

ztNFny:
> > We've long had a parent-dialog for lists, I'm referring to that.
> 
> But we are talking about the "new parent" extra text field. That one is only in dev.

New parent field is also in the old implementation, it's shown when you select "create new" from the dropdown.

> > You could of course always define multiple levels manually using the separators - and imho that's good enough.
> 
> I would understand this gui logic. But why was an extra "new parent" field introduced at all in that case? (Prod state is good enough for this) Only reason I can think of is to keep the user from typing a colon.

Again the field isn't new, the only thing i changed in my implementation is that i replaced the dropdown entry with an icon - an icon that when clicked hides the dropdown and shows a text input field. I did that because imho the old "item in dropdown" is not a nice UI.
So both old implantations didn't allow creating multi-level structures, only yours does.

eddiemuc:
> New parent field is also in the old implementation, it's shown when you select "create new" from the dropdown.

There's no "create new" entry in the released prod state. A not-yet-released state can't be "old". It is just in the test workflow, not even in beta. It is fair to discuss whether this should go out to our end users. 

> > I would understand this gui logic. But why was an extra "new parent" field introduced at all in that case? (Prod state is good enough for this) Only reason I can think of is to keep the user from typing a colon.
> 
> Again the field isn't new, the only thing i changed in my implementation is that i replaced the dropdown entry with an icon - an icon that when clicked hides the dropdown and shows a text input field. I did that because imho the old "item in dropdown" is not a nice UI. So both old implantations didn't allow creating multi-level structures, only yours does.

I can't easily test the master state, so I am asking: when one selects "create new list" (or clicks the icon in your PR) then I assume the text field for "add new parent" appears. 
Questions:
* what happens now with the "parent list select" field? Does it disappear? If yes, how can I come back to it
* if it does not disappear: what happens now if I select another list there? Does then the "add new parent list" field disappear?
* if both fields (parent list selector and add-new-parent-list field) are visible at the same time, I select a parent list (not "create new") and enter something in the add-new-parent-list field and then click "create": what is then created? Is the selected list silently dropped in that case?

ztNFny:
> when one selects "create new list" (or clicks the icon in your PR) then I assume the text field for "add new parent" appears.

This screenshot from my PR should answer all the questions: https://github.com/cgeo/cgeo/pull/18286#issuecomment-4830948126
Only 1 field is ever shown, the button toggles between them.

> if both fields (parent list selector and add-new-parent-list field) are visible at the same time, I select a parent list (not "create new") and enter something in the add-new-parent-list field and then click "create": what is then created? Is the selected list silently dropped in that case?

In the old (current master) implementation if "new" is filled then new is used, else the dropdown value. Which I find bad UI, which is why I changed it.

> There's no "create new" entry in the released prod state. A not-yet-released state can't be "old". It is just in the test workflow, not even in beta. It is fair to discuss whether this should go out to our end users.

It was implemented in October, 9 months back: https://github.com/cgeo/cgeo/pull/17498
Is release really lagging behind by almost a year???

eddiemuc:
> This screenshot from my PR should answer all the questions: https://github.com/cgeo/cgeo/pull/18286#issuecomment-4830948126
> Only 1 field is ever shown, the button toggles between them.

Thanks

> In the old (current master) implementation if "new" is filled then new is used, else the dropdown value. Which I find bad UI, which is why I changed it.

Agree

> > There's no "create new" entry in the released prod state. A not-yet-released state can't be "old". It is just in the test workflow, not even in beta. It is fair to discuss whether this should go out to our end users.
> 
> It was implemented in October, 9 months back: https://github.com/cgeo/cgeo/pull/17498
> Is release really lagging behind by almost a year???

Yes it is :-(

murggel:
> I hope I will find some time this wee to have a look at the workflow and give some feedback

I think from users perspective it is not quite obvious to type a colon to create sublists. 
So I would vote for the new design.
From my point of view, bBoth approaches (PR #18286 and PR #18317) seem to struggle with distinguishing between defining a new parent and selecting an existing one - so why not opening a dialog when clicking on the "+"-button to enter the new name and preselect it then in the drop down. 

Maybe it has technical reasons, because the pulldown is filled with only existing lists?

moving-bits:
Finally found time to write a few lines regarding the "parent lists" feature. I'll try to summarize the thoughts I had when implementing a "create parent list" feature.

Let's start with the following lists:
- Countries:Germany
- Vacation

There are a couple of different use-cases (with regard to hierarchical lists) when creating new lists:
- (a) Add a new sublist for an already existing parent list, eg. "Countries:Austria"
- (b) Add a sublist to a non-parent list, eg. "Vacation:Place A", "Vacation:Place B" etc.
- (c) Add a new hierarchy, eg. "Attributes:Climbing"

(Same would be true for renaming lists. And there may be more use-cases.)

(a) probably is the most frequently used use-case. It can be achieved with current release 2026.06.19 already by selecting "Countries" from "parent list" selector.

(b) and (c) currently (released version) requires you to know of and use the ":" syntax, which makes this functionality sort of invisible for most users, even though it's probably the most important use-case, as this is how a new hierarchy can be created initially. That's why I added the "create new parent list" entry to "parent list" selector (an UI which @ztNFny improved upon).

Showing the non-parent lists in the selector would make use-case (b) even easier, but we need to find a good UI for it to avoid confusion (eg: parent lists should probably stay above the list of non-parent lists to make them more accessible).

Creating multi-level hierarchical lists is a less-frequently used case, IMHO, probably mostly for power users with many, many lists. It can be achieved already today by renaming step-by-step, adding new levels with each step. Or, if the user knows of ":" syntax, by using this.
I would not object such a feature for multi-level-list creation, but only so, if the UI for it stays very clean and does not make creating "default" (= simple parent/child) list creation more complicated

So my vote would be:
- We need a "create new parent" functionality (as we have currently on `master`, but not yet released as feature release).
- UI improvement by @ztNFny would be fine for me on a quick glance.
- Having it technically separated for reusability with filters etc. seems to be the logical next step.
- Having non-parent lists selectable as new parent would be a nice add-on.
- Creating multi-level parent lists is not necessary, IMHO, but I would not object, if we can integrate it in a way that is non-blocking for the use-cases above.

eddiemuc:
From discussion here and in the other PR (https://github.com/cgeo/cgeo/pull/18286) I conclude that multilevel-hierarchy-support is not prioritized. Being a somewhat "pure" mind myself (:-)) my opinion on this is that we 
* either support this --> then we allow multilevel hierarchies in general AND consider this in all our gui including create/rename
* or explicitely disallow it --> then we need to disallow such hierarchical creation in general (even with the colon sign) and of course dont offer it in create/rename either

Anyway, if multihierarchical support is not wanted I guess we go with PR https://github.com/cgeo/cgeo/pull/18286 and abandon this one?

ztNFny:
I wouldn't disallow it, just not provide an UI for the creation. So if the user does "a:a:a" manually it should be shown as 3 levels - and currently is.

The problem with my PR is that you rightfully remarked it's ugly in adding stuff to SimpleDialog and should be in a utility method. UI wise I'd think mine would be sufficient.

moving-bits:
Multi-level parent lists are fine for me as well, I'm just struggling with a feasible UI for it. It should not interfere with creating the more common parent/sublist pairs, and the question remains, how many parent levels you want to support UI-wise? The ":" syntax allows an "unlimited" number of levels, how many do you want to support explicitly in the UI?

I agree with @ztNFny, UI of his PR would be fine for me, but, as you rightfully pointed out, code-wise integration could be improved upon.

eddiemuc:
> Multi-level parent lists are fine for me as well, I'm just struggling with a feasible UI for it. It should not interfere with creating the more common parent/sublist pairs, and the question remains, how many parent levels you want to support UI-wise? The ":" syntax allows an "unlimited" number of levels, how many do you want to support explicitly in the UI?

This PR represents my suggestion for exactly that. The GUI proposed here would support inifinite sublevels without having to ever type a colon for the user

> I agree with @ztNFny, UI of his PR would be fine for me, but, as you rightfully pointed out, code-wise integration could be improved upon.

There's no restriciton using "my" PR as a (new) base for "@ztNFny"'s GUI, if this is the approach you want to follow. Some work would be left to do though, eg I didn't do the integration into "named filters" yet

ztNFny:
Considering the ongoing beta-phase and that we should deliver something for a release, let's be a bit pragmatic about this:
1. Unify UI between lists and filters
2. Enable 1-level list creation for both
3. Tune multi-level-lists

1 and 2 we already have, only 3 seems to require some discussion. So lets do 1+2 in a solid implementation (which would be approximately my PR with @eddiemuc code improvements) quickly - this would be a state that we can ship as release. Then let's discuss - and i'd really say discuss = in one of our calls - the requirements and concepts for multi-level. If we can make that in 1 release good, if it's only in the next also fine - and if we decide not to do it at all also fine. But let's not block the required 1+2 by 3.

murggel:
> Considering the ongoing beta-phase and that we should deliver something for a release, let's be a bit pragmatic about this:
> 1. Unify UI between lists and filters
> 2. Enable 1-level list creation for both
> 3. Tune multi-level-lists
> 
> 1 and 2 we already have, only 3 seems to require some discussion. So lets do 1+2 in a solid implementation (which would be approximately my PR with @eddiemuc code improvements) quickly - this would be a state that we can ship as release. Then let's discuss - and i'd really say discuss = in one of our calls - the requirements and concepts for multi-level. If we can make that in 1 release good, if it's only in the next also fine - and if we decide not to do it at all also fine. But let's not block the required 1+2 by 3.

Fine for me, and for multi-level support, I will overthink my workflow, maybe it will also work for me with the PR of @ztNFny

moving-bits:
@eddiemuc 
I guess this PR is superseded by #18430 as well?

eddiemuc:
> @eddiemuc 
> I guess this PR is superseded by #18430 as well?

I don't know, have to check in new beta

ztNFny:
Superceded

PR #18354, opened 2026-07-23
rel to #8713: add parser for trailing-hemisphere decimal coordinates …

### LINKED ISSUE #8713: Support certain coordinate format for import from clipboard
I just now received the second user claiming, that a certain coordinate format is no (or no longer?) detected as valid coords from clipboard into c:geo.

This format is used by the app "mapy.cz", which is rather popular in Czech Republic.

The format is:
`51.3481209N, 7.0445565E`

Scenario:
- Have such coords contained in your clipboard
- Create new waypoint
- Open coordinate input

Result:
- Import from clipboard not shown

Expected:
- Import from clipboard should be shown to import these coord format

Remark:
Both users claimed, that it worked before the current beta 2020.07.26-RC, however I tested with 2020.07.02 but I does not work there as well. Maybe it used to work before? Therefore labeling as "Feature Request" and "Bug" for now.

Question:
Is it the same logic here as also for the coordinate parsing from personal note or are these two functions separated codewise?

cylorak:
…(mapy.cz)

Adds a parser for decimal-degree coordinates with a trailing hemisphere letter (e.g. mapy.cz 50.3481209N, 7.0445565E), including variants with a space before the letter and an optional degree sign.

This makes such coordinates recognized for clipboard import in the coordinate input dialog (hasClipboardCoordinates() → GeopointParser.parse()).

Adds unit test test8713 and a changelog entry.

Fixes #8713.

Remark for reviewer: Targeting master because this is in fact a new feature (the issue additionally has a Bug label)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/location/GeopointParser.java
+++ b/main/src/main/java/cgeo/geocaching/location/GeopointParser.java
@@ -27,7 +27,7 @@ public class GeopointParser {
     private static final Pattern PATTERN_BAD_BLANK_DOT = Pattern.compile("(\\d)\\. ([-+]?\\d{2,})");
     private static final Pattern PATTERN_BAD_BLANK_FOR_DEG_COMMA_COMMA_PARSER = Pattern.compile("([-+]?\\d{1,3},\\d+), ([-+]?\\d{1,3},\\d+)");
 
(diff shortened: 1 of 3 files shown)


moving-bits:
@cgeo/team 
Any objections to merge this one? 
~(Otherwise will do so tomorrow, resolving merge conflict on included changelog entry along the way)~ some fixes required first

moving-bits:
ok to test

moving-bits:
retest this please

moving-bits:
@cylorak 
There is one failing test, `testEditModifiedCoordinates` in `CGeoParserTest` - can you cross-check, please?

And, while being at it, please remove changes to `changelog_base.md` - we'll do that after merging, otherwise this is a frequent source of merge conflicts.

PR #18356, opened 2026-07-24
inject cookies into GC session to simplify wherigo login

ztNFny:
Wherigo login is now done by an interactive signin to geocaching.com, then granting wherigo.com an oauth token.
To ease that login workflow inject the c:geo session cookies for geocaching.com into the webview to skip the login step in most cases


ztNFny:
rebased on master as this should go into the current beta

PR #18361, opened 2026-07-24
Add fetching and storing of "special" bookmark lists (fix #18362)

### LINKED ISSUE #18362: Add Support for special lists
### Tell us your idea!

geocaching.com exposes a user's ignore list and favorites as virtual lists at /plan/lists/ignored and /plan/lists/favorites. Unlike regular bookmark lists these have no GUID and aren't served by the bookmark list API and not available in c:geo

Integration of the special lists with a new JSON parser under BookmarkListActivity

### Is there an alternative?

_No response_

### Additional Information

PR #18361

JakeDot:
geocaching.com exposes a user's ignore list and favorites as virtual lists at /plan/lists/ignored and /plan/lists/favorites. Unlike regular bookmark lists these have no GUID and aren't served by the bookmark list API, so their geocache data is scraped from the JSON embedded in the page's __NEXT_DATA__ script tag instead.

- GCParser.searchBySpecialList() fetches and parses these pages, reusing the existing bookmark-list geocache parsing logic
- GCList gains a specialList flag so GCListLoader/GCConnector route these through the new fetcher instead of the bookmark list API
- BookmarkListActivity now lists "Favorites" and "Ignore list" as fixed entries alongside the user's real bookmark lists, using the existing preview/download/offline-storage flow


main/src/main/java/cgeo/geocaching/connector/gc/GCParser.java
```diff
@@ -746,46 +746,126 @@ public static SearchResult searchByBookmarkList(final IConnector con, final Stri
...
 
             return searchResult;
         } catch (final Exception e) {
-            Log.e("GCParser.searchByBookmarkLists: error parsing html page", e);
+            Log.e("GCParser.searchBySpecialList: error parsing page", e);
             return null;
         }
     }
 
+    //method is only used in AndroidTests
```
ztNFny, line 850: test methods should be in test classes

PR #18364, opened 2026-07-24
Fix stacked progress dialogs when deleting multiple logs (fixes #18279)

### LINKED ISSUE #18279: deleting multiple logs the "delete log" dialog appears multiple times
When you delete multiple logs of 1 cache the progress dialog gets shown X times for the Xth log being deleted.
So 1. log all good, 2. log it shows 2 progress dialogs on top of each other, 3. 3 and so on.
Didn't notice any functional problems but couldn't debug it yet, so assuming it's only visual.

JakeDot:
LogsViewCreator created a new LogActivityHelper on every delete-log click. Since the underlying delete WorkerTask is a global singleton keyed by id, each new LogActivityHelper registered another ProgressDialogFeature observer on it without ever disposing the previous ones (the hosting activity never gets destroyed between clicks). As a result, deleting the Nth log caused N progress dialogs to appear.

Reuse a single LogActivityHelper instance for the lifetime of the activity, matching the pattern already used in LogCacheActivity and LogTrackableActivity.


moving-bits:
retest this please

moving-bits:
Retest this please

Issue #18058, opened 2026-05-03
Add a "cache placement" view to the map


bitlux:
### Tell us your idea!

The "Hide a Geocache" page on geocaching.com is inadequate for hiding caches for one main reason: physical stages with edited coords are not shown in the conflicts map. For example, if I know a puzzle final is at N 40 25.123 W 100 10.100 and have edited its coords to this value, it will still not show up on the map. I think it would be great if c:geo provided a feature that corrected for this shortcoming.

Specifically, the live map could have a view or option to shade/highlight the area within 528 feet of every traditional cache and physical waypoint of a multi, puzzle, letterbox, or Wherigo cache whose coordinates have been edited by the user. (I specify physical waypoint because there is no distance requirement for virtual waypoints).

eddiemuc:
I believe what you ask for already exists: the "circle" view in nap quick settings:

![Screenshot_2026-05-03-08-19-10-95_a916ceb0c8a82619faeb76fe6b8ae997.jpg](https://github.com/user-attachments/assets/9eab7199-5f1d-4919-b2df-cb6f4f2b9756)

![Screenshot_2026-05-03-08-17-19-05_a916ceb0c8a82619faeb76fe6b8ae997.jpg](https://github.com/user-attachments/assets/ea9d7a49-93e4-459b-9f5f-25ba8cbf6c30)

Please reopen if this is not what you mean?

bitlux:
@eddiemuc Thanks for pointing that out! That is almost exactly what I was asking for. The two possible enhancements I would suggest are:

1. If a puzzle cache has "challenge" in the name, *do* draw a circle around the virtual coordinates, because challenge caches ARE at the posted coordinates.
2. I'm not sure this one is possible. For hidden, physical stages of caches I own (puzzle finals, multi stages, etc.), but I have *not* edited the coordinates, can those waypoints be shown as well?

ztNFny:
> If a puzzle cache has "challenge" in the name, do draw a circle around the virtual coordinates, because challenge caches ARE at the posted coordinates.

Not necessarily. New challenges require a final waypoint - which may or may not be at the listing coords. Old challenges don't even require the coords to be listed at all.
So drawing a circle around the listing coords wouldn't be correct.

giangi1986:
What's missing, in my opinion (as per the app and the official website), is an input menu that lets you know whether you're in the red circles by using your real or fake location.
I'd really like a dedicated menu.
The fake location section is currently missing.

[...]

PR #18366, opened 2026-07-24
Support Unpublished Caches

JakeDot:
## Description

Adds an "Own Unpublished Caches" feature that lets users see, sync, and act on the caches they own that aren't live on geocaching.com yet.

- `PseudoList.OWN_UNPUBLISHED_LIST` — a real pseudo-list alongside `<All Caches>`/`<History>` in the Lists picker. Opening it (`OwnUnpublishedGeocacheListLoader`) syncs from `my/geocaches.aspx?archived=y` and stores membership via `DataStore.syncOwnUnpublishedCachesList`.
- `OwnUnpublishedCachesActivity` — a dedicated `CacheListActivity` subclass fixed to that pseudo-list, with a "New draft" action (opens `geocaching.com/play/hide`) and a "Show on live map" action.
- `OwnUnpublishedCachesLayer` — overlays these caches as clickable markers on the live map (reuses the existing cache-marker click-to-detail plumbing).
- `UnpublishedGeocache` — a `Geocache` subclass (reflection-based field copy, since `Geocache` has no copy constructor and ~65 fields to keep in sync by hand). `GCParser.searchOwnUnpublishedGeocaches()` now loads results as this type.
- `Geocache.getStatusCodes()` — an extension point for subclasses to report additional per-cache status codes on top of the regular archived/disabled/etc. flags; wired into `CacheDetailsCreator.addCacheState()`, the method shared by both the full cache-detail screen and the map's cache popup sheet.
- `UnpublishedGeocache` distinguishes the two real geocaching.com states for a not-yet-live cache: `StatusCode.UNPUBLISHED` (still in the owner's editing phase) vs. `StatusCode.AWAITING_PUBLICATION` (submitted, waiting on a reviewer) — determined by checking the cache's logs for a `SUBMIT_FOR_REVIEW` not followed by a reviewer's `TEMP_DISABLE_LISTING` (which sends it back to editing).
- A secrets-free debug-APK build workflow (`.github/workflows/build-apk.yml`), since the existing `tests.yml` pipeline requires org secrets not available on this fork. It also publishes the built APK to an `apks` branch for easy retrieval.

## Related issues

Related to the discussion on placement-conflict visibility in cgeo/cgeo#18058.

## Additional context

Every commit has been verified via the CI build in this fork (compiles cleanly) plus static bytecode inspection (`aapt`/`apksigner`/dex string scans) confirming the classes and logic actually landed in the built APK. Runtime UI behavior has not been verified on a real device/emulator yet.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LG5LEBM12ancuVYYTzz9R1)_

### DIFF
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -1,0 +1,41 @@
+name: Build debug APK
+
+# Secrets-free debug APK build for testing feature branches on forks
+# where the org secrets required by tests.yml are not available.
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - claude/unpublished-caches-bookmarklist-7d8mj7
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: temurin
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v5
+        with:
+          cache-read-only: true
+
+      - name: Build debug APK
+        run: ./gradlew assembleBasicDebug
+
+      - name: Upload APK
+        uses: actions/upload-artifact@v7
+        with:
+          name: cgeo-basic-debug-apk
+          path: main/build/outputs/apk/basic/debug/cgeo*.apk
+          if-no-files-found: error
--- a/.github/workflows/build-apk.yml
+++ b/.github/workflows/build-apk.yml
@@ -10,7 +10,7 @@ on:
       - claude/unpublished-caches-bookmarklist-7d8mj7
 
 permissions:
-  contents: read
+  contents: write
 
 jobs:
   build:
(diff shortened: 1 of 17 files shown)


JakeDot:
retest this please

ztNFny:
this implementation is quite heavy implementation-wise for a relatively simple and rarely used functionality. If it was implemented more like e.g. PQs then I'd be more open to it, but right now it's adding a lot of overhead.

E.g. i don't understand why it requires a dedicated activity and a dedicated map layer when it only implements what's already there:

> Dedicated screen for {@link PseudoList#OWN_UNPUBLISHED_LIST}: the normal offline cache list, plus a shortcut
> * to start a new hide on geocaching.com and a shortcut to see these caches overlaid on the live map
normal list + create custom cache - which is already in any list + viewing lists on map - already in any list

> Overwrites the membership of {@link PseudoList#OWN_UNPUBLISHED_LIST} with the given caches. Unlike {@link #addToList},     * this bypasses the {@link AbstractList#isConcrete()} check, since that pseudo list is intentionally kept in sync     * with an external source (geocaching.com) rather than edited by the user like a normal list.

Why should this list be different than any other list??

> UnpublishedGeocache.java‎

Just use the existing custom cache type?

> build-apk.yml

doesn't belong in here at all. If you have special needs for your own pipelines please keep them there.

moving-bits:
@JakeDot 
Can you describe in a few sentences what this PR is meant to be, eg. what actions do you want to support for unpublished caches?

I have not checked this PR yet, but here my 2 cents in terms of "managing unpublished caches from within c:geo":
- I vote for not recreating the "create new geocache" functionality of geocaching.com
- As soon as a geocache is published for review, it's accessible for c:geo by entering the GC code
- So the only missing piece (in my opinion) would be to read and interpret the list of "caches marked as ready for review" and load those into c:geo. As @ztNFny mentioned, this should be done in a similar way to what we have for PQ/Bookmark lists (as far as applicable technically). We probably don't need a "special internal list" for it, a regular one should do. (Yet-unpublished caches will be shown as "archived" in c:geo, so published ones look different and are easily distinguishable, this alone should make handling/this PR quite easier.)

Issue #18304, opened 2026-07-01
Map zooms to world on resume


murggel:
Reported on support, ticket [146640](https://cgeosupport.droescher.eu/scp/tickets.php?id=10532)

Using Google satellite map:

It happens many times when i am near cache. Map shows only one cache when i am near of it.

I continue walking and close my phone screen with power button and i put phone to pocket.

Next time when i open phone screen with power button and fingerprint map is zoomed totally out and i see whole world.

I removed cgeo-app, shut down phone for couple of minutes. Then i install cgeo-app again from google play store. Nothing happens. Maybe it goes more bad. 

Problem occurs several times, but not every time when i have shut down my screen.

My phone screen is turned off only for couple of minutes. I think that only 1 minute is enough for this map zoom problem.

murggel:
I already informed the user about waypoints with far away coordinates

murggel:
Feedback from user: it seems to happen only with Google satellite (Offline maps are ok) and when moving (while being at home it seems also not to happen)

Lineflyer:
I can almost reproduce this behavior with Google Sat map on `master`:
Once I try to map a list of caches in a certain region/city while having Google Sat as my map provider, it will randomly open "some other region".  For example: List of caches around Barcelona will result in map of western Africa (zoomed out 1000km per unit). List of caches in whole Bavaria opens a zoomed out world map.

My guess would be a coding error while calculating or handing over the bounding box to be shown for Google Satellite (unless you can't reproduce).

moving-bits:
Haven't cross-checked in code yet, but bounding box calculation should be done either globally in `UnifiedMapActivity` (then there should be no differences between GM and OSM) or by the underlying abstract tileprovider class (then it should affect all GM variants, not GM Sat only). Can you cross-check whether the othet GM variants are affected as well?

We have a relatively recent fix regarding bounding box and GM, but that got published with 2026.05.21 - maybe that fix did not cover all relevant places?

murggel:
> Haven't cross-checked in code yet, but bounding box calculation should be done either globally in `UnifiedMapActivity` (then there should be no differences between GM and OSM) or by the underlying abstract tileprovider class (then it should affect all GM variants, not GM Sat only). Can you cross-check whether the othet GM variants are affected as well?
> 
> We have a relatively recent fix regarding bounding box and GM, but that got published with 2026.05.21 - maybe that fix did not cover all relevant places?

[...]

PR #18367, opened 2026-07-24
rel #18304:  postpone animation in GoogleMapsFragment.zoomToBounds

murggel:
## Description

postpone animation in GoogleMapsFragment.zoomToBounds

Proposal from AI

probably fixes the random zoom in  #18304

Since I can't reproduce it clearly, difficult to test for me

## Related issues

rel #18304

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsFragment.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsFragment.java
@@ -216,10 +216,21 @@ public class GoogleMapsFragment extends AbstractMapFragment implements OnMapRead
     @Override
     public void zoomToBounds(final Viewport bounds) {
         if (mMap != null) {
-            final CameraUpdate cu = CameraUpdateFactory.newLatLngBounds(new LatLngBounds(
-                    new LatLng(bounds.bottomLeft.getLatitude(), bounds.bottomLeft.getLongitude()),
-                    new LatLng(bounds.topRight.getLatitude(), bounds.topRight.getLongitude())), 50);
-            mMap.animateCamera(cu);
+            final View view = requireView();
+            if (view.getWidth() == 0 || view.getHeight() == 0) {
+                //See Bug #14948: w/o map width/height the bounds can't be calculated
+                // -> postpone animation to later on UI thread (where map width/height will be set)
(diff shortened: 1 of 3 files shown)


moving-bits (commented):
Haven't had time yet to check the implications of this change (or to debug the original issue), just wanted to mention that all that you are changing here are the allowed minimum/maximum values for those tile providers, not actual zooming behavior.
I'm not sure, whether those minimum/maximum values are actually checked against by `UnifiedMapActivity` anymore (the old map implementation did at some place), but either way, it should not affect bounding box calculation anyway. Additionally, the values you are using here, are invalid values, maximum should not be smaller than minimum, and supported maximum is something around low-twenties (would need to look those values up in Mapsforge code), and I do not know how Mapsforge library reacts to invalid values.

murggel:
Similar to PR #18502

moving-bits:
I've done some testing with the user-provided data I used for #18502 - I was not able to make this PR here work reliably. Here's what I did:
- I have added `Log.e()` calls to `setBearing` and `zoomToBounds` (as those two were the culprit in the other issue).
- On a quick glance, it looked ok. First mapping after opening the cache list did work.
- Interestingly, the `getWidth() == 0 || getHeight() == 0` branch never got logged, which fits into the observation, that the `mMap.setOnMapLoadedCallback()` gets logged first, before any `setBearing` calls.
- But when I then moved the map manually to some other place, closed the map (returning to cache list), then mapped this list again, the resulting position was random most of the time.
- Additionally, map was focussing on N0/W0 area first frequently, only then positioning to a different place.

murggel:
> I've done some testing with the user-provided data I used for #18502 - I was not able to make this PR here work reliably. Here's what I did:
> - I have added `Log.e()` calls to `setBearing` and `zoomToBounds` (as those two were the culprit in the other issue).
> - On a quick glance, it looked ok. First mapping after opening the cache list did work.
> - Interestingly, the `getWidth() == 0 || getHeight() == 0` branch never got logged, which fits into the observation, that the `mMap.setOnMapLoadedCallback()` gets logged first, before any `setBearing` calls.
> - But when I then moved the map manually to some other place, closed the map (returning to cache list), then mapped this list again, the resulting position was random most of the time.
> - Additionally, map was focussing on N0/W0 area first frequently, only then positioning to a different place.
> 

Thanks for testing, so let's close this in favor of #18502

PR #18372, opened 2026-07-26
Fix persistent cookie expiration restoration

Konano:
## Description

Preserve the expiration time of persistent cookies when saving and restoring the cookie store.

Previously, restored cookies lost their original expiration timestamp. This could cause them to be treated as session cookies and disappear from persistent storage after a subsequent update or restart.

This change:

- Stores each persistent cookie's `expiresAt` timestamp.
- Restores cookies with their original expiration time.
- Remains compatible with cookies stored using the legacy format.
- Correctly handles cookie values containing `=` characters.


Konano:
move to `release` branch.

PR #18374, opened 2026-07-26
Fix persistent cookie expiration restoration

Konano:
## Description

Preserve the expiration time of persistent cookies when saving and restoring the cookie store.

Previously, restored cookies lost their original expiration timestamp. This could cause them to be treated as session cookies and disappear from persistent storage after a subsequent update or restart.

This change:

- Stores each persistent cookie's `expiresAt` timestamp.
- Restores cookies with their original expiration time.
- Remains compatible with cookies stored using the legacy format.
- Correctly handles cookie values containing `=` characters.

> Sorry for submitting two incorrect PRs Orz

### DIFF
--- a/main/src/main/java/cgeo/geocaching/network/Cookies.java
+++ b/main/src/main/java/cgeo/geocaching/network/Cookies.java
@@ -106,10 +106,25 @@ public final class Cookies {
             final String oldCookies = Settings.getPersistentCookies();
             if (oldCookies != null) {
                 for (final String cookie : StringUtils.split(oldCookies, ';')) {
-                    final String[] split = StringUtils.split(cookie, "=", 3);
-                    if (split.length == 3) {
+                    final int nameSeparator = cookie.indexOf('=');
+                    final int expiresAtSeparator = cookie.lastIndexOf('=');
+                    long expiresAt = Long.MAX_VALUE;
+                    String cookieWithoutExpiresAt = cookie;
+                    if (expiresAtSeparator > nameSeparator) {
                         try {
-                            addCookie(new Builder().name(split[0]).value(split[1]).domain(split[2]).build());
+                            expiresAt = Long.parseLong(cookie.substring(expiresAtSeparator + 1));
+                            cookieWithoutExpiresAt = cookie.substring(0, expiresAtSeparator);
+                        } catch (final NumberFormatException ignored) {
+                            // Legacy cookie without expiresAt
+                        }
+                    }
+                    final int domainSeparator = cookieWithoutExpiresAt.lastIndexOf('=');
+                    if (nameSeparator > 0 && domainSeparator > nameSeparator) {
+                        final String name = cookieWithoutExpiresAt.substring(0, nameSeparator);
+                        final String value = cookieWithoutExpiresAt.substring(nameSeparator + 1, domainSeparator);
+                        final String domain = cookieWithoutExpiresAt.substring(domainSeparator + 1);
+                        try {
+                            addCookie(new Builder().name(name).value(value).domain(domain).expiresAt(expiresAt).build());
                         } catch (final RuntimeException ignored) {
                             // ignore
                         }
@@ -129,6 +144,8 @@ public final class Cookies {
                 persistentCookies.append(cookie.value());
                 persistentCookies.append('=');
                 persistentCookies.append(cookie.domain());
+                persistentCookies.append('=');
+                persistentCookies.append(cookie.expiresAt());
                 persistentCookies.append(';');
             }
             Settings.setPersistentCookies(persistentCookies.toString());


Konano:
Can anyone take a look? @moving-bits @ztNFny

ztNFny:
For what is restoring the cookie even required at all? For services the login data is stored and a re-login can be performed at any time.

Konano:
> For what is restoring the cookie even required at all? For services the login data is stored and a re-login can be performed at any time.

@ztNFny Thanks for taking a look. Restoring the cookie is needed to reuse an existing, still-valid server-side session. Stored username/password credentials should be the fallback when that session is missing or invalid, rather than being submitted again on every app restart.

This PR does not introduce cookie restoration; c:geo already deliberately saves persistent cookies and calls `Cookies.restoreCookies()` during application startup. It fixes a broken round trip in that existing mechanism:

1. The server returns `gspkauth` with an expiration time, so OkHttp marks it as persistent and c:geo stores it.
2. On the next startup, c:geo restores only its name, value and domain. Since `expiresAt` is omitted, OkHttp rebuilds it as `persistent=false`.
3. When another cookie is received, `dumpCookieStore()` runs again. It only writes cookies where `persistent()==true`, so the restored `gspkauth` is removed from persistent storage.
4. On a later startup there is no authentication cookie left, and c:geo has to POST the stored credentials again.

I verified this with debug logging. Before the fix, the sequence was:

```
SAVE gspkauth: persistent=true
RESTORE gspkauth: persistent=false
DUMP: gspkauth omitted, stored cookie data becomes empty
next cold start: GCLogin.postCredentials()
```

With this PR, the original expiration timestamp and therefore the persistent state survive the round trip. On repeated cold starts, c:geo reports `Already logged in` and does not call `postCredentials()`.

Re-login with stored credentials still works as before when the cookie is missing or rejected. The change merely avoids unnecessary authentication requests while an existing session can still be reused.

ztNFny:
Cookie lifetime is quite short and a backup old, so it wouldn't be valid anymore in most cases. And we anyway perform a re-login on every start.
The whole cookie restoration can probably be removed completely without any negative impact, reducing code complexity.

Konano:
Ah, I see your point. I looked into this because I frequently encountered CAPTCHAs during automatic re-login, which was quite annoying, and from what I observed, the geocaching.com authentication cookie has a fairly long lifetime.

ztNFny:
There might be some potential to improve the cookie storage - I agree that it seems login is triggered "too often".
Cookies are handled via Android Webview, might be some config options there if you want to look into it.

In any case this PR wouldn't improve that situation - and as mentioned - imho we could remove the whole cookie restore. Btw - do we export cookies without the "include credentials" option? That'd be a bug as it's leaking info - and with the option it's not needed as the app can just re-login.

SammysHP:
> Cookie lifetime is quite short and a backup old

This isn't about backups and we should keep cookies as long as their specified lifetime. Using a previous login session reduces server and network load and improves overall app responsiveness.

ztNFny:
> > Cookie lifetime is quite short and a backup old
> 
> This isn't about backups and we should keep cookies as long as their specified lifetime. Using a previous login session reduces server and network load and improves overall app responsiveness.

Going by the name I assumed it was related to restoring backups - confusing naming there.
So I took the time to dig through the code: It's called on every app start and restores the setting "cookiejar".
However looking at my devices the setting "cookiejar" only contains the jwt cookie from geocaching.com - which is useless. After a fresh start of c:geo it has "gpskauth" which would be good - but after running the app for a while it's gone and only jwt is there. So there's definitly a big bug in that saving code.

-> As I now learned that the "restore" is "app start" then yes, this PR makes sense. However we also need to adress the issues that a) the gpskauth cookie often isn't stored, b) cookies that shouldn't be stored are stored (jwt), c) evaluate which cookies from which services actually need to be stored (OC?)

(The other place the method to write that setting is used is for the WIG login - and that cookie is very short-lived - it probably shouldn't be stored at all (that's code I implemented, will remove it there).)

ztNFny:
And is this active restoration even required? Android Webviews cookie store should handle that between sessions ... It might be that our actual problem is having this method in the first place?!

Konano:
Using WebView's `CookieManager` as the persistent backend might indeed be a cleaner solution. However, OkHttp does not share WebView cookies automatically; we would need a `CookieJar` adapter that writes response cookies through `CookieManager.setCookie()` and reads request cookies through `getCookie(url)`.

That would be a larger refactoring with some details to handle, such as asynchronous writes, flushing, and the existing calls.

Konano:
I think two separate cookie stores are being conflated here. Normal GC login and network requests use OkHttp with `Cookies.cookieJar`; Android WebView's `CookieManager` is only used for interactive/manual login, after which the authentication cookie is copied into the OkHttp jar. Therefore WebView persistence cannot restore the cookie jar used by `Network` on app startup.

The disappearance of `gspkauth` you observed is exactly what this PR fixes: after restoration it currently becomes non-persistent, so a later cookie-store dump removes it.

You are also right about settings export: `pref_cookiejar` is currently missing from `getSensitivePreferenceKeys()`. That seems to be a separate existing bug and should be fixed as well.

moving-bits:
@ztNFny 
Are all questions clarified, ok to merge then?

moving-bits:
> You are also right about settings export: `pref_cookiejar` is currently missing from `getSensitivePreferenceKeys()`. That seems to be a separate existing bug and should be fixed as well.

fixed with #18479

moving-bits:
Discussion in today's dev meeting:
- We want to test this on `master` branch first (cherry pick or retarget before merging)
- To get the expected result, we need to adapt our login process as well, as we login unconditionally currently

PR #18413, opened 2026-08-03
Integrate Pebble map service, watchapp, and UI

caco3:


main/AndroidManifest.xml
```diff
@@ -14,6 +14,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION" android:minSdkVersion="34" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
```
eddiemuc, line 17: Is this mandatory for the feature to work? We will.most likely not get this approved by Google Playstore

SammysHP:
Where's the issue where this change was discussed? What about the description of the pull request?

caco3:
Please calm down!

As you can see, this PR is still marked as draft, thus it is not ready and does not need to be reviewed!
It is a PoC I am working on. I am very aware that it will very unlikely pass a review or even the Google Play Store checks!

The goal of this PoC was to see if I could show part of the c:geo map on a Pebble watch. This would be handy when hiking to see which direction to take on a crossing without having to take out the phone every time. My PoC works, but is rater slow. And I had to learn that a background task needs to run to render the map periodically, even when the phone got locked. This is very intrusive. I know that https://github.com/caco3/pebble-navigation has access to direction and distance from c:geo even the phone is locked. But there seems to be a difference between a background task to provide just some numbers and a background tasks which must render maps.

SammysHP:
> As you can see, this PR is still marked as draft, thus it is not ready and does not need to be reviewed!

Sorry, I didn't want to offend you. By opening a pull request, you basically say: "Hey, look at this code, that I'm working on. What do you think? Any comments?" The draft state doesn't matter. If you don't want feedback, don't open a pull request. But once you open a PR, expect comments, especially when the PR does not contain any description and doesn't link against an existing issue. ;)

Issue #18407, opened 2026-08-01
New Filter UI (filters vs. advanced filters vs. markers)


Lineflyer:
While using current RC I still stumble upon the "new" filter features, and seemingly I still don't fully understand the difference.

The main issue for me is, that you now have to (overflow) menu items called `Filter`. The first one being the well known filter screen with its funnel icon, the second the new filter mechanism with an icon looking like a funnel with something (?) next to it. 

First I thought the new filtering feature is for using it as a list/map marker, but somehow both interact. If I e.g. store a traditional filter using advanced mode, it pops up in the other filtering menu.

Somehow I even managed (not sure how), that I ended up with the filter no longer filtering my live map but showing the "to be filtered" caches with a faded out cache icon marked with some other icon.

Long story short:
IMHO the UI is not very intuitive in explaining the user how it is correctly used.

ztNFny:
For the icon there's https://github.com/cgeo/cgeo/issues/18282

We have 2 clearly separate features: filter and conditioal markers.
Them sharing the same rule definitions is perfectly fine and shouldn't be changed.
Imho the only problem is the dialog "manage filters", which is used for both managing filters as well as selecting which markers to apply. I don't use that screen at all, instead using only the markers list (long-tap on the icon) to apply one - with that usage everything makes perfect sense.

So my suggestion would be to call the second "filter" button "markers" and have it open a screen to apply them. That screen - like the filter dialog - can then contain a "manage" button that goes to a dialog to edit them.

eddiemuc:
Imho we can also completely remove the link to the svaed filters/markers from map screen. In the beginning I suggested to (only) make this screen available via settings, but that wasn't well received.

Imho we could also remove the hidden long-tap-function to activate/deactivate markers.  I don't believe that users really need this as much as we probably think.

ztNFny:
> Imho we could also remove the hidden long-tap-function to activate/deactivate markers.  I don't believe that users really need this as much as we probably think.

For me that one is the ONLY part of the functionality that's needed.

eddiemuc:
> > Imho we could also remove the hidden long-tap-function to activate/deactivate markers.  I don't believe that users really need this as much as we probably think.
> 
> For me that one is the ONLY part of the functionality that's needed.

Correct. Your claim is currently the only reason this feature is accessible in such a prominent way in the map. But  since @Lineflyer is asking for a reevaluation of the UX maybe there are other ways satisfying both your personal needs and the issue adressed by him.

[...]

PR #18430, opened 2026-08-04
markers vs filters

ztNFny:
This is an FINISHED draft to support the discussions in https://github.com/cgeo/cgeo/issues/18407

# What's done:
### "Marker" toolbar button
- Make "Marker" button open marker selection list
- "Manage" leads to a dialog where markers/filters can be defined
- Dialog displays only those filters with icon defined - only those make sense to apply as markers
image
image

### Filter management
- Make filter management (load/save) available in basic mode
image

# What's still left to do:
- Reworking filter management dialog to be only about managing it - or even completly remove it in favor of the load/save buttons in Filters dialog
- Store "priority" (for marker) as a property of the filter
- ~~add icon-assignment to the filter save dialog~~ -> https://github.com/cgeo/cgeo/pull/18286
- behaviour of "Markers" toolbar button if no markers are defined
- finalize naming (e.g. Markers vs. overlays vs ...)
- no easy way to clear applied markers (as the neutral button is "manage", not "clear")


ztNFny:
- save (list/filter): SimpleDialog -> own class
- filter load/save buttons in basic mode
- naming: "marker"
- order activity for markers -> priority becomes a normal integer --- base: ManageListAdapter

ztNFny:
all the points we discussed in last weeks call should be done.
- unified "new" dialog for list/filters
- filter management back in filter dialog, removed the new "named filters" activity
- new marker priority activity
- marker menu opens list to enable them
- filter management also in basic mode

If there are comments/remarks/suggestions I'll try to address them this week, but latest Thursday. After that I'd appreciate a merge and fixes in follow-up PRs.

ztNFny:
retest this please

moving-bits:
build/main/src/main/java/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java:253: error: cannot find symbol
09:30:33            final NamedFilter namedFilter = NamedFilter.overwrite(name, getFilterFromView());
09:30:33                                                       ^
09:30:33      symbol:   method overwrite(String,GeocacheFilter)
09:30:33      location: class NamedFilter

ztNFny:
> build/main/src/main/java/cgeo/geocaching/filters/gui/GeocacheFilterActivity.java:253: error: cannot find symbol 09:30:33 final NamedFilter namedFilter = NamedFilter.overwrite(name, getFilterFromView()); 09:30:33 ^ 09:30:33 symbol: method overwrite(String,GeocacheFilter) 09:30:33 location: class NamedFilter

oops, that got introduced on the last rebase. Unfortunately this PR has now grown quite a bit larger than I like :-(

moving-bits:
There is a failing test:

19:06:49  NamedFilterTest > testAddNewSortsAlphabetically FAILED
19:06:49      java.lang.NullPointerException at NamedFilterTest.java:59
19:06:56  
19:06:56  902 tests completed, 1 failed, 28 skipped

moving-bits:
Thanks for the development so far. I could do only a very short test today, will try to continue tomorrow.

One question, though: Is there a way to rename a filter? So far I found only "save as", which will create a new filter, giving me a warning dialog + leaving me with two copies.

Other (possible) finding:
- Create a filter and store it as named filter, wg. "tradis". => Filter is now available as marker
- Change the filter settings on the map and confirm dialog, but do not save the filter again.
- To me it looks like the transient map filter reflects the new settings (ok), but the filter edit dialog still shows the name of the filter this transient setting is based on. This is good when trying to update a filter, but may cause confusion otherwise. Have no solution for it yet, though.

ztNFny:
>  Is there a way to rename a filter?

No, only delete + save (and there previously wasn't such an option either. For lists it's also only in the seperate list management submenu. As noone requested that feature in many years I don't think it's an issue.)

> Other (possible) finding:

true - on map this is indicated with "*" (unchanged from previous implementation) - that part is missing in filter dialog.

I was actually thinking about a "saved" indicator in filter dialog that gets reset to "unsaved" state as soon as you modify any setting - but implementing that requires LOTS of listeners and I reverted it.

ztNFny:
> * , but the filter edit dialog still shows the name of the filter this transient setting is based on. This is good when trying to update a filter, but may cause confusion otherwise. Have no solution for it yet, though.

UI adjusted to indicate the unstored filter. This is only updated on opening the filter dialog, NOT live when modifying a filter.
image

moving-bits:
> > Is there a way to rename a filter?
> 
> No, only delete + save (and there previously wasn't such an option either. For lists it's also only in the seperate list management submenu. As noone requested that feature in many years I don't think it's an issue.)

That's sad, especially when being able to create filters for markers + sort them in hierarchical lists.

Use case: Have a handful of filters for markers, then, after a while, you want to organize them by putting them into a hierarchy. You would have to save each filter again, confirm the warning dialog, then later manually delete all "old"/non hierarchical filters.

Proposal: We already detect if you want to save an existing filter under a different name. Add a "rename" option here:
image
- Ok: Save copy of filter under a different name
- Cancel: Cancel saving filter
- Rename: Rename existing filter

ztNFny:
> Proposal: We already detect if you want to save an existing filter under a different name. Add a "rename" option here:

Good idea.. Feel free to add this once this PR is merged.

eddiemuc:
> > Proposal: We already detect if you want to save an existing filter under a different name. Add a "rename" option here:
> 
> Good idea.. Feel free to add this once this PR is merged.

If renaming should be outside of this PR then maybe https://github.com/cgeo/cgeo/issues/18493 should be reopened.

Technically not allowing renaming is a regression because it is possible in current beta release. So we either have to dexide that rename is not necessary or consider this a bug report.

eddiemuc:
Is deleting of filters possible?

ztNFny:
We don't have rename for filters in release. If it's a wanted feature then please feel free to add it - i won't as i don't have time, as announced.

ztNFny:
> Is deleting of filters possible?

Sure!

eddiemuc:
Proposal: in the "load" filter selection we could add an "edit" button behind each filter. If clicked, a new dialog opens where the filter name can be changed. This dialog also has a delete option



Such action icon buttons are already supported natively by SimpleDialog.

eddiemuc:
> > Is deleting of filters possible?
> 
> Sure!

How?

ztNFny:
> Proposal: in the "load" filter selection we could add an "edit" button behind each filter. If clicked, a new dialog opens where the filter name can be changed. This dialog also has a delete option
> 
> 
> 
> Such action icon buttons are already supported natively by SimpleDialog.

The filter list in release has such buttons for delete. I removed (actually just didn't re-add it, was removed by eddiemuc before) the delete button from there as it a) reduces the room available for actual list names (more important with nested lists) and b) increases the height of each list item.

ztNFny:
> > > Is deleting of filters possible?
> > 
> > Sure!
> 
> How?



Pretty much the same as in your named filter dialog 😉

moving-bits:
> > Proposal: We already detect if you want to save an existing filter under a different name. Add a "rename" option here:
> 
> Good idea.. Feel free to add this once this PR is merged.

Will look into it (as time permits). Reopening #18493.

ztNFny:
To be clear, the goal of this PR is to get filters vs. markers to a state where we don't get users asking "what's this second filter button and why doesn't it work" + unify the "new" dialog between filters and lists.

There's probably a ton of UI improvements (all feature requests against the current release) that can be done on top of this, e.g. 
- in list management the "assign list marker" and "rename list" should be replaced with the new InputWithParentDialog
- in filter management a "rename parent" should be added (maybe this should even be inside InputWithParentDialog and unified with list management)
- unify "rename list/filter" - probably also in InputWithParentDialog

eddiemuc:
> > > > Is deleting of filters possible?
> > > 
> > > Sure!
> > 
> > How?
> 
> 
> 
> Pretty much the same as in your named filter dialog 😉

This deletes the store filter? I thought it would simply delete the filter settings used (in case of the screenshot the offline-filter), not the persistes named filter settings. 

How does one now deletes the filter settings (eg for offline or live) without deleting the named saved filter?

ztNFny:
> How does one now deletes the filter settings (eg for offline or live) without deleting the named saved filter?

Same as before, the third button in toolbar
image
Actually that's not a delete but a "reset/clear filter settings" - and with the delete icon right below it I thought a different icon to differentiate them was in order. If the "reset" icon doesn't work maybe "clear" would be an option
https://github.com/cgeo/cgeo/blob/master/main/src/main/res/drawable/ic_menu_clear_playlist.xml 
image

eddiemuc:
Back to the "rename" topic: we could replace the trash can symbol.here with an "edit" symbol. Click opens a dialog where user can change name. Dialog has the usual "ok" and "cancel" and a third option "delete" to delete the saved filter.

ztNFny:
> Back to the "rename" topic: we could replace the trash can symbol.here with an "edit" symbol. Click opens a dialog where user can change name. Dialog has the usual "ok" and "cancel" and a third option "delete" to delete the saved filter.

Could be an option to keep the UI minimal - and that dialog could integrate the rename parent as well.

Question is whether by that "delete" would be too hidden. You already didn't find that delete button, would you find a delete button hidden behind an Edit button?

eddiemuc:
> > Back to the "rename" topic: we could replace the trash can symbol.here with an "edit" symbol. Click opens a dialog where user can change name. Dialog has the usual "ok" and "cancel" and a third option "delete" to delete the saved filter.
> 
> Could be an option to keep the UI minimal - and that dialog could integrate the rename parent as well.
> 
> Question is whether by that "delete" would be too hidden. You already didn't find that delete button, would you find a delete button hidden behind an Edit button?

Few ideas:
* design a new "edit/delete§ icon (somehow merging the pen and trashcan icons into one)
* have two icon buttons (edit and delete) at the place. But that takes space away from the line of course
* have a three-dot-symbol which, if tapped, expands into a. overflow dropdown with "edit" and "delete" option

moving-bits:
> * have a three-dot-symbol which, if tapped, expands into a. overflow dropdown with "edit" and "delete" option

Out of the mentioned options I would vote for the third option, as we have this in other places already (eg.: Route/Tracks quick settings) - that's probably easier to understand than another handcrafted button. 

Would "Edit" be more than a "Rename" option? If not, I suggest to use "Rename" as menu entry.

Maybe make the dedicated "Save" button just that - "Save" using existing name, no renaming, no parent list selection, no overwrite warning etc., and offer "Rename", "Save as" and "Delete" options via overflow menu. (With "Save as" and "Delete" allowing to select existing and/or create new parent filter hierarchy levels.)

ztNFny:
> Maybe make the dedicated "Save" button just that - "Save" using existing name, no renaming, no parent list selection, no overwrite warning etc., and offer "Rename", "Save as" and "Delete" options via overflow menu. 

I wouldn't do that as it's quite the deviation from the existing behaviour and users have for many years been used tonit NOT overwriting. There should at least be a overwrite warning, point to an "Save as new" button that's elsewhere (e.g. in the overflow menu).

The overflow menu could then have
- Rename / Change icon
- Delete
- Save as new
- Rename parent

ztNFny:
Judging by the discussions there's for sure some room for improvement in the filter management area. But it seems there's nothing fundamentally wrong - so can this PR be merged then? I'll anyway not be able to add further changes.

moving-bits:
@cgeo/team:
If no one objects I plan to merge this PR tonight (not before 9pm) to be able to collaborate on this issue.

PR #18436, opened 2026-08-06
fix #6631: keep offline log verbatim, trim only when publishing

### LINKED ISSUE #6631: Offline loggs and signiture
I am encountering the following problem when using one-click offline log. 
I have set up a signiture for myself. The signiture contains a text that is preceeded by 3 newlines.
When I log normally, online, the signiture is entered into the log properly, and as the cursor is placed at the beginning of the field, my text and signiture are properly formatted to begin with.

When I save a one-click offline log, a log entry is saved on my device. It already contains the signiture and no other text. I click the log entry and the online logging form is shown.
The problem is that the first 3 newlines from my signiture are gone. I have to place the cursor at the beginning of the text and tyype my newlines, then place the cursor again at the beginning of the field to enter my log text.

IMO the newlines should not be removed (as described) and cursor be placed at the very beginning of the text entry field, even if it has text content.

Tested with c:geo 2017.05.22 on 3 different devices.

cylorak:
### What
Closes #6631
One-click offline logs stored the signature trimmed, so leading blank lines were lost.

### How
Store offline drafts verbatim (setLog override in OfflineLogEntry) and move trimming to the publishing
boundaries via a shared LogUtils.trimForPublishing helper — once in the upload path
(createLogTaskLogic, covering all connectors) and in the field-note export. Parsed server logs unchanged.

### Tests
LogUtilsTest (helper) + OfflineLogEntryLogTextTest (verbatim storage). Local gates green:
testBasicDebug, assembleBasicDebug, checkstyle.

## Additional context
Prepared with AI assistance and reviewed.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/export/FieldNotes.java
+++ b/main/src/main/java/cgeo/geocaching/export/FieldNotes.java
@@ -2,6 +2,7 @@ package cgeo.geocaching.export;
 
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
+import cgeo.geocaching.log.LogUtils;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.storage.ContentStorage;
 import cgeo.geocaching.storage.Folder;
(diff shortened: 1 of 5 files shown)


moving-bits:
ok to test

moving-bits:
Retest this please

PR #18441, opened 2026-08-07
fix #16147: correct heading hash handling when editing a sent log

### LINKED ISSUE #16147: Log Edit: heading hash not properly recognized
From support  Ticket #666955
 https://cgeosupport.droescher.eu/scp/tickets.php?id=8645
Version (2024.09.18)

When editing a previously created and sent log entry, hash symbols used to mark a line as heading get not treated correctly.

To reproduce:
- write a new log entry for a cache which contains heading hashes. 
- send log entry to server 
- refresh cache (and thus the log entry)
- open the log entry for edit 

For initial log: `#Test#`
This produces: `Test`
(Hashes are removed) 

For initial log:
`#Test#
More log text`

This produces:
`# Test #
More log text`
(Spaces are added after/before hash)

cylorak:
Closes #16147

### What
Editing a previously sent log corrupted markdown headings. geocaching.com returns a real heading as
`<h1>Test</h1>`, and `GCLoggingManager.convertLogTextToEditableText` converted the **closing** tag to a
leading space + trailing hash (`</h1>` → `" #"`), so the heading round-tripped as `# Test #` and kept
growing on every cycle of edit → send → edit.

### How
Drop the closing-hash conversions: `</h1>|</h2>|</h3>` → `""`. ATX-style markdown headings do not use
closing hashes, so `<h1>Test</h1>` → `# Test` and the conversion becomes idempotent. Minimal fix
(3 closing-tag lines + a one-line clarifying comment).

### Tests
No new test scaffolding (kept minimal). Verified locally: `assembleBasicDebug` + `checkstyle` green,
and an on-device before/after check (heading shows `# Test`, not `# Test #`).

### Notes
- The historical lone-`#Test#` → `Test` case (from 2024) was a gc.com-side rendering issue and no longer
  reproduces; this targets the genuine c:geo heading round-trip bug.
- No changelog entry (team adds on merge). No user-facing strings changed (no Crowdin impact).

## Additional context
- Prepared with AI assistance and reviewed before submission.


moving-bits:
ok to test

ztNFny:
according to the documentation on GC the old implementation is in fact the correct one??
image

ztNFny:
Tested it with a real log. This PR would actually break log formatting!

image

@cylorak contributions are welcome, but this (and many other comments) seem to be pure AI generated. Obviously this contribution wasn't tested before submitting it 🙁

cylorak:
Thanks @ztNFny for your counter-examples. `# Text #` being the documented and correct GC markup
means that dropping the closing hash was the wrong direction; I get the point.

One correction though: I tested this by myself and manually — I edited a real cache log on
geocaching.com and checked the before/after for a single heading (`# Test`), which is the exact case
like in the original issue. What I did not cover were your two scenarios — consecutive headings and a
heading followed by text — and those are where I see the round-trip actually breaks (heading-merge, and the
trailing space that makes GC re-parse the heading as a `<p>`). So the fault is incomplete test
coverage; fair criticism, and I take it.

On the AI point: yes, I use AI assistance, but I did the analysis and did the on-device testing
myself — for me, the fault here was scope, not blind automation. I'll continue the way-forward on
#16147.

ztNFny:
> One correction though: I tested this by myself and manually — I edited a real cache log on
geocaching.com and checked the before/after for a single heading (# Test), which is the exact case

your PR already breaks the original scenario - "# test" is not rendered correctly - which is why can only assume it wasn't tested.

The analysis I did in the issue actually shows that the original issue is invalid - the one that this PR tries to adress. But instead we have some other ones there.

PR #18442, opened 2026-08-08
Add bitmap cache for GoogleV2GeoItemLayer (rel. to #18404)

moving-bits:
## Description
First part of https://github.com/cgeo/cgeo/issues/18404#issuecomment-5218883836, related to Google Maps marker cache. Created with AI.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/GoogleV2GeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/GoogleV2GeoItemLayer.java
@@ -13,7 +13,6 @@ import cgeo.geocaching.utils.Log;
 
 import android.content.res.Resources;
 import android.graphics.Point;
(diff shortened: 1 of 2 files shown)


eddiemuc (approved):
LGTM.

This fixes the Google Maps part of #18404: markers using the same source bitmap now reuse a BitmapDescriptor instead of allocating a new descriptor (and corresponding bitmap copy) for every marker. This is intentionally a partial fix for the broader OOM investigation.

The 256-entry limit is a sensible bound against unbounded cache growth. If more than 256 distinct Bitmap instances are used, older entries are evicted and a later use creates a descriptor again; this reduces reuse only and does not affect already-rendered markers or rendering correctness.

Keeping this cache local to the Google Maps integration is the right choice. Although both map integrations key by source bitmap identity, their cached values and lifecycles differ: Google Maps manages BitmapDescriptors, while mapsforge bitmaps require explicit reference counting.

moving-bits:
thanks for cross-checking the two GeoItemLayer related PR

PR #18443, opened 2026-08-08
Share and release marker bitmaps in mapsforge layer (rel. to #18404)

moving-bits:
## Description
Next part of https://github.com/cgeo/cgeo/issues/18404#issuecomment-5218883836, related to Mapsforge (non-VTM) marker cache. Created with AI.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeV6GeoItemLayer.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/geoitemlayer/MapsforgeV6GeoItemLayer.java
@@ -18,7 +18,9 @@ import androidx.annotation.ColorInt;
 import androidx.annotation.NonNull;
 
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


eddiemuc (approved):
LGTM.

This fixes the mapsforge marker-bitmap part of #18404: converted mapsforge bitmaps are shared per source bitmap instead of being copied once per marker. The accompanying change to destroy removed child layers is essential as well, because it lets Marker.onDestroy() release each marker's bitmap reference. This is intentionally a partial fix for the broader OOM investigation.

The 256-entry limit is a sensible bound on retained reusable bitmap conversions. If a map uses more than 256 distinct source Bitmap instances, older cache entries are evicted and later use of such an image requires another conversion. Existing markers remain correct because they retain their own reference; eviction only limits reuse.

Keeping this cache specific to mapsforge is appropriate. Its cached bitmap values use explicit reference counting, unlike Google Maps' BitmapDescriptors, so a shared cross-provider cache would make ownership and cleanup less clear.

Also good that the removed layers are destroyed outside layerLock, avoiding calls into foreign lifecycle code while holding the group-layer lock.

PR #18454, opened 2026-08-09
fix #17105: render numbered and unnumbered lists manually

### LINKED ISSUE #17105: Ordered HTML list not displayed as such
### Describe your problem!

The cache listing of GCGVDQ contains an **ordered** HTML list which is displayed as **unordered** list in c:geo. The order list starts with at  "Station 1", see HTML source code (and attached screenshot).

Image

### How to reproduce?

1. Load the listing of GCGVDQ in c:geo.
2. Switch to the description tab.

An **unordered** list is displayed, see screenshot.

Image

An **ordered** list should be displayed.

### Reproducible

Yes

### c:geo Version

2025.06.15-RC

### System information

```text

```

### Additional Information

_No response_

eddiemuc:
fix #17105: render numbered and unnumbered lists manually

This fix REPLACES all ol, ul and li tags with a custom tag and then handles all three custom tags in the unknowntagshandler. Supports nested lists, also when they are mixed (numbered/unnumbered). Does NOT support attributes of the list tags.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/html/HtmlUtils.java
+++ b/main/src/main/java/cgeo/geocaching/utils/html/HtmlUtils.java
@@ -37,8 +37,12 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Locale;
+import java.util.Stack;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Strings;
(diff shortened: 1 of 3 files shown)


eddiemuc:
failing test is unrelated

moving-bits:
Found a bug, see GC7KPHM, missing indentation of list elements:

|Display in c:geo|Display on webpage|
|---|---|
|image|image|

Source code is like this:
```
<ul>
<li> ... <br /> ... </li>
<li> ... <br /> ... </li>
<li> ... <br /> ... </li>
</ul>
```

eddiemuc:
After looking into source code of Android's Html class I fear it would be much more complex to consider all additional styling infos considered there for own list rendering. So probably my approach here was in vain. I close for now.

PR #18478, opened 2026-08-14
Adapt build.gradle to new versioncode scheme

moving-bits:
## Description
Implements the necessary `build.gradle` changes for our new versionCode scheme
(See https://github.com/orgs/cgeo/discussions/113)

Edit: Starts with "X=9" for nightly builds and switches "year + 1000" to "year + 1" for nighlies (as suggested in https://github.com/orgs/cgeo/discussions/113#discussioncomment-18020574), to get the new numbering scheme to our CI.
If we want to stay with "Year + 1000", it can still be activated later. (This would be upgrade-compatible, but not the other way around.)


moving-bits:
retest this please

moving-bits:
FYI: Won't merge today due to current CI problems

PR #18479, opened 2026-08-14
Mark cookieJar as sensitive preference (rel. to #18374)

moving-bits:
## Description
As mentioned in https://github.com/cgeo/cgeo/pull/18374#issuecomment-5200934460


JakeDot:
LGTM

PR #18482, opened 2026-08-15
Mass export of Events to Google Calendar

cano64:
## Description

This new feature adds a menu option to the Map screen to Export all future event caches on the map page to the Google Calendar. A new calendar permission is requested when this feature is used.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -31,3 +31,6 @@ android.useAndroidX=true
 
 # For local instrumented tests: leave original c:geo installation intact on device
 android.injected.androidTest.leaveApksInstalledAfterRun=true
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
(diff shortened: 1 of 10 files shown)


eddiemuc:
Before starting with any review for this:
is this a wanted feature?

SammysHP:
Looks like a lot of code for a feature with a very specific use case nobody requested so far. Even with events almost every day in my area currently, I only export individual events to the calendar that I expect to attend.

cano64:
Yes, I wanted this feature, because I didn't feel like adding the events manually one by one to the calendar.  That's why I implemented it. You can pick manually which events you want  before export.

eddiemuc:
Technical note:
currently, we use an intent to add an event to calendar. There's unfortunately no way to add multiple events via intent. This is why this PR introduces calendar API. Which in turn requires (dynamic) calendar read and write permission.

eddiemuc:
Question: does this PR really only support google calendar or also other calendar apps?

ztNFny:
I dislike the extra permission for this corner-case. 

Would it be an option to write a .ics file with the events? This can the easily be imported into any calendar.

Also the UI should be removed - filtering can be done in the list or the user can use select mode, no need for an extra UI there.

moving-bits:
FYI: The selection mechanism mentioned above is part of cache list view. When using that, you can trigger an action for either all caches in that list (if none is selected) or for all selected caches. The action would then filter for event caches and create a temporary `.ics` file.

If you want to automate things further, you can create the `.ics` file and then share it via an Android intent. For this, you won't need any extra UI, you won't need any extra permissions, and you address any calender app the user might have installed, not just Google Calendar.

Setting WIP label for now.

PR #18483, opened 2026-08-15
Add WebView listing description style (fix #5986)

### LINKED ISSUE #5986: webview of cache

JakeDot:
## Description
This PR adds a new listing description style that renders listing HTML in a `WebView` instead of the existing HTML parsing + `Spannable` pipeline. Existing styles remain unchanged and continue to use the current text rendering path.

- **Style surface**
  - Added `HtmlStyle.WEBVIEW` and exposed it in the existing description-style selector.

- **Description UI**
  - Added `@id/description_webview` to the description page layout.
  - Switched visibility between `TextView` and `WebView` based on selected style.

- **Rendering path split**
  - For `WEBVIEW`, bypasses `HtmlUtils.renderHtml(...)` / `Spannable` creation and loads raw listing HTML via `loadDataWithBaseURL(...)`.
  - Kept markdown/internal-connector behavior on the existing markdown path.
  - Preserved existing behavior for all non-WebView styles.

- **WebView behavior**
  - Configured conservative settings (no JS, no file/content access).
  - Routed clicked links out through `ShareUtils.openUrl(...)`.

```java
if (descriptionStyle == HtmlStyle.WEBVIEW) {
    return new Pair<>(descriptionText, false);
}
```

## Related issues
https://github.com/cgeo/cgeo/issues/17105#issuecomment-3021792865
https://github.com/cgeo/cgeo/issues/5986

## Additional context
The new style is additive and opt-in; users must explicitly select it from the “change description style” dialog.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -144,6 +144,10 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.View.OnLongClickListener;
 import android.view.ViewGroup;
+import android.webkit.WebResourceRequest;
+import android.webkit.WebSettings;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
 import android.widget.LinearLayout;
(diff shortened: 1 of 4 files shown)


eddiemuc:
I would be willing to give this a shot. It's probably not working completely yet, but its only in the nightlies and must be actively activated. So it can be tried out in nightlies without risk.

Ok to merge?

eddiemuc:
Retest this please

JakeDot:
retest this please

moving-bits:
@JakeDot 

Lint reports an error:

```
18:47:00  Lint found 1 error and 99 warnings. First failure:
18:47:00  src/main/res/layout/cachedetail_description_page.xml:31: Error: Placing a <WebView> in a parent element that uses a wrap_content layout_height can lead to subtle bugs; use match_parent instead [WebViewLayout]
18:47:00          <WebView
18:47:00           ~~~~~~~
18:47:00      src/main/res/layout/cachedetail_description_page.xml:10: wrap_content here may not work well with WebView below
18:47:00  
18:47:00  
18:47:00  > Task :main:lintBasicDebug FAILED
18:47:00  Lint found 1 error, 99 warnings. First failure:
18:47:00  
18:47:00  src/main/res/layout/cachedetail_description_page.xml:31: Error: Placing a <WebView> in a parent element that uses a wrap_content layout_height can lead to subtle bugs; use match_parent instead [WebViewLayout]
18:47:00          <WebView
18:47:00           ~~~~~~~
18:47:00      src/main/res/layout/cachedetail_description_page.xml:10: wrap_content here may not work well with WebView below
18:47:00  
18:47:00     Explanation for issues of type "WebViewLayout":
18:47:00     The WebView implementation has certain performance optimizations which will
18:47:00     not work correctly if the parent view is using wrap_content rather than
18:47:00     match_parent. This can lead to subtle UI bugs.
```

When applying a fix for this, please ensure it does not break the other description styles.

PR #18486, opened 2026-08-16
fix #18485: guard audiomanager against start-play-failure of mediaplayer

### LINKED ISSUE #18485: IllegalStateException in audio player (Wherigo)
Reported by Play Store stats for current beta 2026.08.11-RC:

```
Exception java.lang.IllegalStateException:
  at android.media.MediaPlayer.getDuration
  at cgeo.geocaching.utils.AudioManager.$r8$lambda$ejHbcNB-ZEtTKpkOSpKytA0nNpM (SourceFile:76)
  at cgeo.geocaching.utils.AudioManager$$ExternalSyntheticLambda4.run (SourceFile)
  at cgeo.geocaching.utils.AudioManager.$r8$lambda$LAjrSbL4IYdRgigcOAB8nN_H8ek (SourceFile:223)
  at cgeo.geocaching.utils.AudioManager$$ExternalSyntheticLambda12.run (SourceFile)
  at io.reactivex.rxjava3.internal.schedulers.ScheduledDirectTask.call (SourceFile:38)
  at io.reactivex.rxjava3.internal.schedulers.ScheduledDirectTask.call (SourceFile:25)
  at java.util.concurrent.FutureTask.run (FutureTask.java:328)
  at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run (ScheduledThreadPoolExecutor.java:323)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1100)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:624)
  at java.lang.Thread.run (Thread.java:1572)
```

Wherigo Player seems to be the only calling place for that method, therefore setting Wherigo label.

### Additional info

Google AI gives the following additional info:

An IllegalStateException in android.media.MediaPlayer.getDuration() occurs because you are calling the method while the player is in an invalid lifecycle state. According to the official [Android Developers MediaPlayer Documentation](https://developer.android.com/reference/android/media/MediaPlayer), getDuration() can only be safely called when the player is in one of the following states: Prepared, Started, Paused, PlaybackCompleted, or Stopped.If you call it immediately after initializing the player or setting a data source without waiting for it to prepare, the internal state machine throws this error.

❌ Common Mistakes That Trigger the Error
- Calling it right after setDataSource(): The media hasn't been loaded or analyzed yet.
- Calling it in the Idle or Initialized states: The player does not yet know the file length.
- Calling it after release() or reset(): The player is completely uninitialized or destroyed.

How to Fix It
1. Use the Asynchronous Preparation Listener (Recommended)

Never assume the media is ready immediately. Use an OnPreparedListener to ensure the track details are completely loaded into memory before fetching the duration.

```
MediaPlayer mediaPlayer = new MediaPlayer();
mediaPlayer.setDataSource(audioUrl);

// Listen for the "Prepared" state before calling getDuration
mediaPlayer.setOnPreparedListener(new MediaPlayer.OnPreparedListener() {
    @Override
    public void onPrepared(MediaPlayer mp) {
        // Safe to call getDuration() here!
        int duration = mp.getDuration(); 
        mp.start();
    }
});

// Prepare asynchronously so you don't block the main UI thread
mediaPlayer.prepareAsync(); 
```

2. Wrap It in a State-Check Helper

If your UI updates dynamically (e.g., via a seek bar timer), wrap the duration calculation inside a check or a try-catch block to handle edge-case user interactions like rapid clicking.

```
public int getSafeDuration(MediaPlayer mp) {
    try {
        if (mp != null) {
            return mp.getDuration();
        }
    } catch (IllegalStateException e) {
        // Log error or fallback to a default value
        Log.e("MediaPlayer", "Player was not in a valid state to get duration.");
    }
    return 0;
}
```

eddiemuc:
fix #18485: guard audiomanager against start-play-failure of mediaplayer

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/AudioManager.java
+++ b/main/src/main/java/cgeo/geocaching/utils/AudioManager.java
@@ -56,7 +56,7 @@ public class AudioManager {
                     }
                 }
             }, 0, 500);
-            startPlayer(this.player, this.uri, this.mute, success -> {
+            final boolean started = startPlayer(this.player, this.uri, this.mute, success -> {
                 synchronized (mutex) {
                     if (songId.get() != thisSongId) {
                         //this is from another song and can be ignored
@@ -73,10 +73,12 @@ public class AudioManager {
                     }
                 }
             });
-            this.duration = player.getDuration();
-            this.timePassedBeforeLastStart = 0;
-            this.timeLastStartTimestamp = System.currentTimeMillis();
-            setState(State.PLAYING, false);
+            if (started) {
+                this.duration = getDuration(this.player);
+                this.timePassedBeforeLastStart = 0;
+                this.timeLastStartTimestamp = System.currentTimeMillis();
+                setState(State.PLAYING, false);
+            }
         });
     }
 
@@ -153,7 +155,7 @@ public class AudioManager {
                 } else {
                     if (player.isPlaying() && (state == State.COMPLETED || state == State.STOPPED)) {
                         //when we set "COMPLETED" status while player is still playing (when duration is reached)
-                        //then it can happen we come here and player is not paused. Thus we ensure this here
+                        //then it can happen we come here and player is not paused. Thus, we ensure this here
                         this.player.pause();
                     }
                     this.player.seekTo(0);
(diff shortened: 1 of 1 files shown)


eddiemuc:
Jenkins seems to be down. 

All the GitHub Actions are skipped, is this an error?

JakeDot:
@eddiemuc could it be the case that the player now silently (no pun intended, still present) doesn't play some audio and the user has no way of knowing without opening the error logs (which big standard wig users won't know how to do?

If that is the case and also in other cases, I would ask you to create toasts for any playback errors.

eddiemuc:
> @eddiemuc could it be the case that the player now silently (no pun intended, still present) doesn't play some audio and the user has no way of knowing without opening the error logs (which big standard wig users won't know how to do?
> 
> If that is the case and also in other cases, I would ask you to create toasts for any playback errors. 

The player is stateful and has a state error. In case it is in that state this is visually indicated by displaying the music play status in red (see method getUserDisplayableShortState() in audiomanager).

Toasts are not ideal. Because when something is wrong and the player triggers music many times (some cartridges do) then each time a new toast would be created. That would be very annoying. Hence the visual clue as described above instead.

JakeDot:
Got it, right! 🤓

eddiemuc:
Retest this please

eddiemuc:
Retest this please

PR #18433, opened 2026-08-05
Fix live map search crash for empty filters

Konano:
## Description

Fix a crash during online live map searches when all active basic filters simplify to `ALWAYS_TRUE`.

`GeocacheFilter.getAndChainIfPossible()` previously returned `ConstantGeocacheFilter.ALWAYS_TRUE` as a basic filter. `GCMap` then attempted to process it based on its filter type, which is `null`, causing a `NullPointerException`.

This PR:

- Omits `ALWAYS_TRUE` from connector filter chains because it imposes no search restriction.
- Preserves `ALWAYS_FALSE` as a marker indicating that the online search can be skipped.
- Handles `ALWAYS_FALSE` in the GC, OC, SU, and AL connectors.
- Adds regression tests for both constant-filter cases.

## Related issues

None.

## Additional context

The issue was observed when moving or zooming the live map: cached offline entries remained visible, while the online refresh failed with a `NullPointerException`.

Unit tests were added, but they were not run locally because no Java runtime is currently available on the development machine.

## Log

```
08-06 00:23:40.551 26636  6644 W cgeo    : java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Enum.ordinal()' on a null object reference
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.gc.GCMap.fillForBasicFilter(SourceFile:140)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.gc.GCMap.createSearchForFilter(SourceFile:130)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.gc.GCMap.createSearchForFilter(SourceFile:112)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.gc.GCMap.searchByViewport(SourceFile:63)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.gc.GCConnector.searchByViewport(SourceFile:280)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.ConnectorFactory.$r8$lambda$G9xy3SlRYFdDKiRX1lRFx4dNliE(SourceFile:363)
08-06 00:23:40.551 26636  6644 W cgeo    : 	at cgeo.geocaching.connector.ConnectorFactory$$ExternalSyntheticLambda7.apply(SourceFile:0)
[...]

Issue #18400, opened 2026-07-30
[2026.07.27-RC] No caches are loaded on map


murggel:
Reported on support, tickets [#689683](https://cgeosupport.droescher.eu/scp/tickets.php?id=10643), [#400669](https://cgeosupport.droescher.eu/scp/tickets.php?id=10649), [#866358](https://cgeosupport.droescher.eu/scp/tickets.php?id=10642), [#324084](https://cgeosupport.droescher.eu/scp/tickets.php?id=10668), [#202511](https://cgeosupport.droescher.eu/scp/tickets.php?id=10676), [#754327](https://cgeosupport.droescher.eu/scp/tickets.php?id=10670), [270043](https://cgeosupport.droescher.eu/scp/tickets.php?id=10679)

We have at least 3 tickets on support, where user claim, that only stored caches are shown on the online map. They tried to recognize, clear filters etc. With some filters it seem to work...

Since last update, I am not able to load cache on the map.
By default, just the labcaches are loaded.
No filter is applied. 

If I apply a filter to select only Traditional cache, they are loaded. (Idem with mystery, not tested for other types)
If I remove all filter, just previously loaded caches are displayed. 

## System information

c:geo version: 2026.07.27-RC

Datetime: 29 juillet à 12:23

Device:
-------
- Device type: SM-S931B (pa1qxeea, samsung)
- Available processors: 8
- Android version: 16
- Android build: BP4A.251205.006.S931BXXSBCZG3
- Screen resolution: 1080x2340px (360x780dp)
- Pixel density: 3.0
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 26.26.34 (260400-945364269)
- Memory: c:geo:629,8 MB, Available:3,1 GB, Total:10,9 GB, Threshold: 216,0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v11, Count:145
- Set language: fr_FR (system default)
- System date format: dd/MM/y
- Time zone: GMT 02:00
- Translator(external): GOOGLE/appPackage:com.google.android.apps.translate(av=true)
- Debug mode active: no
[...]

PR #18488, opened 2026-08-16
rel to #18400, replacing #18433: prevent map crash due to missing constant filter type, optimizes server queries

eddiemuc:
rel to #18400, replacing #18433: prevent map crash due to missing constant filter type, optimizes server queries


eddiemuc:
Retest this please

moving-bits:
Do I understand it correctly, that this PR supersedes #18433, so that #18433 can be closed unmerged?

Issue #15884, opened 2024-07-18
Custom parameters for user-defined tile provider URL


IiroP:
### Tell us your idea!

The newly added support for user-defined tile provider URLs (#15714) is a great start. However, the current implementation is somewhat limiting as it only supports defining the base URL. Some tile servers (e.g. Trailmap.fi) require an API token as query parameter (`?key=XXX`). 

### What solution would you suggest?

I can see at least two potential solutions for this problem:
1) New setting field for URL parameters that are appended to calculated URL (i.e. user inputs `https://tile.openstreetmap.org` to one field and `key=XXX` to another)
2) Allowing the user to type full URL with X/Y/Z wildcards (e.g. `https://tile.openstreetmap.org/{z}/{x}/{y}.png?key=XXX`)

Option 2 is more versatile, as it would also allow the user to change image format etc.

fm-sys:
I would also vote for option 2

wasu-code:
Yes, some maps can have additional parameters - both query parameters (mostly API keys) and path parametes (map themes).

So it would be perfect to provide map URL in format of:  
```url
http://tile{server|stripe}.example.org/cycling/{z}/{x}/{y}.png?key={apikey}&showPOI=never
```

`{z}` - Zoom

`{x}`,`{y}` - Tile number

`{stripe}` - Cycle server number (1..4)

`{server}` - Cycle server char (a,b..c)

`{quadkey}` - [Quadkey numbering system](https://learn.microsoft.com/en-us/bingmaps/articles/bing-maps-tile-system#tile-coordinates-and-quadkeys)

`{ln}` - Map language (en,it,fr...)

It's also worth noticing that different maps can have different **max zoom**

moving-bits:
For documentation purposes: Our nightly versions support customizing the Uri, supporting parameters `{X}`, `{Y}` and `{Z}`, but you can add your own api key etc. to the address. 
For migration purposes: If c:geo does not find both `{X}` and `{Y}` parameters, it automatically adds `/{Z}/{X}/{Y}.png` to the given address (as before the recent change).

PR #18494, opened 2026-08-18
Add support for multiple custom online maps

fabiancz:
## Description

Adds support for multiple custom online maps using XYZ map tiles.

Users can add any online map provider that exposes its map tiles through an HTTP or HTTPS URL containing the `{X}`, `{Y}`, and `{Z}` placeholders, for example:

`https://tiles.example.com/{Z}/{X}/{Y}.png`

Custom map sources can be added, edited, and removed in the map source settings. URL templates may also contain query parameters, including API keys. Lowercase `{x}`, `{y}`, and `{z}` placeholders are accepted as well.

The configured maps are available in the map source selection and work with both Mapsforge map implementations.

This PR also:

- migrates the existing user-defined tile provider to the new custom maps configuration
- validates and normalizes XYZ URL templates
- adds tests for URL validation, formatting, and legacy configuration migration

## Related issues

No related issue.

## Additional context

Custom online maps use temporary in-memory caching only. Users are responsible for complying with the terms of use of their selected map tile providers.

Feature implemented with help of Codex 5.6 Sol.

Screenshot_20260818_215428_cgeo
Screenshot_20260818_214952_cgeo
Screenshot_20260818_215035_cgeo


fabiancz:
I’ve rebased my feature branch onto the latest master and fixed import order, so Static Code Analysis is up to standards. Is there anything I can do on my side to prevent the CI pipelines from failing?

Do you like this feature when you user can add any URL of online map tiles, which supports x,y,z variables?

bekuno:
@fabiancz 
Please have a look at #15884 and #15714.
Doesn't the existing featue solve your request?
Or, what is your PR compared to the existing feature?

SammysHP:
@bekuno The PR adds support for more than one custom tile provider. Currently users are limited to one provider at the same time.

fabiancz:
@bekuno I admit I completely overlooked this feature.
Yes my PR adds support for more than one custom tile provider and naming option. Also it is case insensitive when user inserts URL with `{x} {y} {z}` or `{X} {Y} {Z}`. It looks like it's problem in current implementation.

I can remove this PR and create new one, which will extend current User defined maps feature to support more than one provider and have custom name.
In current implementation of User defined maps there is only support for Mapsforge. My code supports also VTM. I can fix it too.
What do you think?

Thanks for your review 👍

moving-bits:
> In current implementation of User defined maps there is only support for Mapsforge. My code supports also VTM. I can fix it too. What do you think?

FYI: User-defined maps are implemented for both Mapsforge and VTM, see `unifiedmap/tileproviders/UserDefinedMapsforgeOnlineSource.java` and `unifiedmap/tileproviders/UserDefinedMapsforgeVTMOnlineSource.java`. In case this is not working for you this might be a bug.

magma1447:
I searched the issues for this feature and didn't find anyone requesting it, so since I
wanted it myself I went ahead and wrote a patch. Afterwards I realised I should have
checked the pull requests as well, and there I found this PR. Funny timing — just a week
before I started on mine.

Yours is the more complete implementation, so this is explicitly not a competing PR and I
won't be opening one. I'm commenting because of the question you asked that hasn't been
answered yet:

> I can remove this PR and create new one, which will extend current User defined maps
> feature to support more than one provider and have custom name. What do you think?

That happens to be the variant I built, so if the maintainers prefer that direction there
is already a working reference to look at:

https://github.com/cgeo/cgeo/compare/master...magma1447:cgeo:multiple-user-defined-tile-providers

**The difference is replace vs extend.** Yours adds `CustomMapsforge*Source` plus a new
`pref_customMaps` key and retires the old classes and preference. Mine keeps
`UserDefinedMapsforge*Source` and the existing `userDefinedTileProviderUri` key, which now
holds a JSON list — a legacy bare Uri is still understood and converted on read, so there
is no migration step and existing settings backups keep working unchanged.

**One detail that took me a while to spot**, and which may be worth a look in your version
too: the provider id for user-defined providers currently comes out as `<class>:null`,
because `AbstractMapsforge(VTM)TileProvider.getId()` appends
`mapUri.getLastPathSegment()` while the constructor has already rewritten `mapUri` down to
`scheme://host`. So the moment there is more than one entry they collide in
`TileProviderFactory`'s map. That id is also what gets persisted into `pref_tileprovider`,
`pref_previous_tileprovider` and `pref_tileprovider_hidden`.

I gave every entry its own key, and let the entry migrated from the legacy format keep the
key `"null"` on purpose, so its id stays byte-identical to the old one and an existing map
source selection survives the upgrade without anything being rewritten. You solve the same
problem from the other side, by rewriting the prefs during migration — reading the diff, it
looks like `pref_tileprovider` and `pref_previous_tileprovider` are covered but
`pref_tileprovider_hidden` isn't. Might be worth checking whether a previously hidden
user-defined provider leaves a stale id behind there. I could easily be misreading the
diff, though.

**Things yours has that mine doesn't:** URL validation and normalisation, case-insensitive
`{x}/{y}/{z}`, and unit tests. Mine has no tests for the settings parsing at all, which is
a genuine gap on my side.

Mine builds clean against `checkstyle.xml` (10.26.1), `lintBasicDebug` and the unit test
suite, and I've run it on my phone with two providers configured plus an upgrade from a
previously configured single provider.

Happy to hand over anything useful from it, rebase it onto yours, or simply close the tab
and let yours land — whichever the maintainers prefer. Mostly I just didn't want two people
solving this twice without knowing about each other.

Per CONTRIBUTING.md: my version was written with AI assistance (Claude).

moving-bits:
Thanks, @fabiancz and @magma1447 for your contributions and thoughts about this issue, very welcomed!

I have no preferences for one or the other proposal and haven't checked them in detail code-wise, but from reading your statements it looks like a combination of the two might be the ideal solution. Some aspects & feedbacks:
- Guessing from the screenshots of this PR, UI-integration seems fine to me
- introducing a new preference (and migrating the data from the old preference) would be the preferred way in this case. Please keep in mind, that we have users that sometimes (have to) switch between nightly versions and regular versions, so it's a good thing to keep preferences compatible or introduce new ones, if they cannot be kept compatible (as it is here).
- individual tile provider ids is a very valid point - we need that for the reasons stated by @magma1447 
- URL validation and normalisation, case-insensitive {x}/{y}/{z}, and unit tests are all very welcomed parts
- green checks for Checkstyle and lint are a must-have

magma1447:
@fabiancz It seems like we thought about different different details here. Do you want to continue your PR? I won't do anything until you say I should. The PR was yours to start with. If you prefer, I can make a try to take over and push a merge of our implementations.

moving-bits:
superseded by #18586

PR #18502, opened 2026-08-21
Avoid resetting GM to N0/W0 on mapping list

moving-bits:
## Description
We had a user on support stating that every time the user maps a list of caches, position on Google Maps gets reset to N0/W0.
With the help of using the user's database I was finally able to track this down and fix it:
Culprit was that internal map position was still at 0/0 when we tried to zoom to the viewport required for mapping the list. Map animation got cancelled, though, due to setting a new bearing from sensor data. Thus the new map viewport never got set but always returned to (or stayed at) N0/W0.

Might be related to / having the same cause as #18304, but as I cannot reproduce that, I leave that issue untouched.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsFragment.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsFragment.java
@@ -48,6 +48,7 @@ public class GoogleMapsFragment extends AbstractMapFragment implements OnMapRead
     private LatLngBounds lastBounds = null;
 
     private boolean mapIsCurrentlyMoving;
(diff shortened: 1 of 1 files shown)


murggel:
Similar to PR #18367

moving-bits:
> Similar to PR #18367

Thanks, hadn't looked into the technical details of that update.
I'm not sure, but will it ever settle to update the map while new bearing data comes floating in from the sensor?

murggel:
> > Similar to PR #18367
> 
> Thanks, hadn't looked into the technical details of that update.
> I'm not sure, but will it ever settle to update the map while new bearing data comes floating in from the sensor?

To be honest - I don't know, have to check that. That was a proposal from AI...

PR #18515, opened 2026-08-25
fix #18500: re-add display of filters in cache details, make it configurable

### LINKED ISSUE #18500: (2026.08.11-RC) deactivated markers shouldn't show up in Cache-Detail
From support ticket [#754373](https://cgeosupport.droescher.eu/scp/tickets.php?id=10771)

We have a feedback to the markers.
User wants, that deactivated markers doesn't show up in Cache-Detail.
She assumes, that computing the matching markers is slow (it seems, she has a bunch of them, many deactivated and only used for special cases)

Original from user:

> Und noch ein Feedback zu den neuen Markern. Großartiges Feature!  Ich habe viele Filter, einige davon sind aber Hilfsfilter für verschachtelte Filter (z.b. Negationen), andere einfach aus Bequemlichkeit nicht gelöscht, falls ich sie mal wieder benutzen möchte. Diese habe ich zwar in der Filtermarkierung-Übersicht deaktiviert, aber wenn ich Caches aufrufe, sehe ich die unter Passende Filter ausgegraut. z.b. "x_ohne_geloggte_Caches", oder "alles ohne Labs" oder sowas, weil sie ja praktisch auf fast alle Caches zutreffen. Die sollten dort niemals angezeigt werden. 
> 
> Ich habe das Gefühl, dass er an diesen neuen Markierungsmöglichkeiten richtig herumzurechnen hat, die Verzögerungen sind teilweise deutlich. Deshalb wäre es gut, wenn man Filter, die man eh nicht braucht, herausnehmen könnte, ohne sie zu löschen.

eddiemuc:
fix #18500: re-add display of filters in cache details, make it configurable

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1260,7 +1260,7 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             @Override
             protected void onFinished() {
-                CacheInfoBoxes.updateCacheLists(CacheDetailActivity.this.findViewById(R.id.offline_lists), cache, null);
+                CacheInfoBoxes.updateCacheListsAndMatchingFilters(CacheDetailActivity.this.findViewById(R.id.offline_lists), cache, null);
             }
         }.execute();
     }
@@ -1423,10 +1423,10 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
             // cache attributes
             CacheInfoBoxes.updateAttributes(cache, binding.attributesText, binding.attributesGrid, activity);
 
-            // list
+            // list and matching filters
             CacheInfoBoxes.updateOfflineBox(binding.getRoot(), cache, new RefreshCacheClickListener(), new DropCacheClickListener(),
                     new StoreCacheClickListener(), null, new MoveCacheClickListener(), new StoreCacheClickListener());
-            CacheInfoBoxes.updateCacheLists(binding.getRoot(), cache, activity);
+            CacheInfoBoxes.updateCacheListsAndMatchingFilters(binding.getRoot(), cache, activity);
 
             // watchlist
             binding.addToWatchlist.setOnClickListener(new AddToWatchlistClickListener());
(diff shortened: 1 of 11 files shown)


eddiemuc:
Settings:
image

Cache Details:
image

ztNFny:
is it intentional that this added the markers also to CachePopup, not just to details? As we don't even display lists in popup I would have expected that markers aren't show there either

murggel:
> is it intentional that this added the markers also to CachePopup, not just to details? As we don't even display lists in popup I would have expected that markers aren't show there either

But we do show lists in popup, or is that also new in nightly?

ztNFny:
Oops ... I was actually blind there... Was only looking at an unstored cache (that I was sure I had stored ... maybe there's a bug there, need to check).
Somehow the markers section feels more disturbing visually which is what irked me. Maybe because of the visual gap in the buttons and the extra line between the sections?
image

btw, there's an "s" missing in "Keine paSsenden Filter"

murggel:
> btw, there's an "s" missing in "Keine paSsenden Filter"

The typo should be already fixed in crowdin

murggel:
> Oops ... I was actually blind there... Was only looking at an unstored cache (that I was sure I had stored ... maybe there's a bug there, need to check). Somehow the markers section feels more disturbing visually which is what irked me. Maybe because of the visual gap in the buttons and the extra line between the sections? image
> 

Shall we add a separator between the marker and the "Detail"-Button or remove the line between the lists and marker?

I would vote for adding a separator, otherwise the lists and marker are too close

ztNFny:
I'd actually vote for removing the marker info from popup (not from details). We have a limited set of the most important info there. A) markers are not that "important", b) the section consumes quite a lot of space

ztNFny:
Or if it is wanted to stay: hide it unless it actually has content (active markers) to show.

murggel:
> I'd actually vote for removing the marker info from popup (not from details). We have a limited set of the most important info there. A) markers are not that "important", b) the section consumes quite a lot of space

Good point, maybe they are not so important there, because you come from map and have the information via the icon there probably 

I will remove it

See PR #18607

PR #18529, opened 2026-08-29
Add parsing and creation of stored list identifiers in personal notes

eddiemuc:
Add parsing and creation of stored list identifiers in personal notes

Enhancing parsing and creating waypoint infos from/to PNs was one of my very first tasks at c:geo back then. This was extended years later with parsing/creating of variable information.

Even back then I wanted to add also parsing/creating of lists that the cache is associated with. Merging this PR would provide that functionaly. There might be an issue for this, not entirely sure.

Like prefix `@` is used for waypoints and `$` is used for variables, this PR will introduce `#` as prefix for list names

AI was used to create this PR initially, but I have reviewed it entirely and modified it by hand.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -2915,12 +2915,18 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
                 userModifiedWaypoints.add(w);
             }
         }
-        if (userModifiedWaypoints.isEmpty() && cache.getVariables().isEmpty()) {
+        final List<String> listNames = new ArrayList<>();
+        for (final StoredList list : DataStore.getLists()) {
+            if (list.id != StoredList.STANDARD_LIST_ID && cache.getLists().contains(list.id)) {
+                listNames.add(list.getTitle());
+            }
+        }
+        if (userModifiedWaypoints.isEmpty() && cache.getVariables().isEmpty() && listNames.isEmpty()) {
             showShortToast(LocalizationUtils.getString(R.string.cache_personal_note_storewaypoints_nowaypoints));
             return;
         }
 
-        final String newNote = CacheArtefactParser.putParseableWaypointsInText(note, userModifiedWaypoints, cache.getVariables());
+        final String newNote = CacheArtefactParser.putParseableWaypointsInText(note, userModifiedWaypoints, cache.getVariables(), listNames);
         setNewPersonalNote(newNote);
         showShortToast(R.string.cache_personal_note_storewaypoints_success);
     }
--- a/main/src/main/java/cgeo/geocaching/files/GPXParser.java
+++ b/main/src/main/java/cgeo/geocaching/files/GPXParser.java
@@ -20,6 +20,7 @@ import cgeo.geocaching.list.StoredList;
 import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.log.LogEntry;
 import cgeo.geocaching.log.LogType;
+import cgeo.geocaching.models.CacheArtefactParser;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.models.Trackable;
 import cgeo.geocaching.models.Waypoint;
@@ -273,6 +274,10 @@ abstract class GPXParser extends FileParser {
                     }
                     // modify cache depending on the use case/connector
                     afterParsing(cache);
+
+                    //invalidate list tags in PN to avoid list creation
+                    cache.setPersonalNote(
+                        CacheArtefactParser.markListsAsIgnored(cache.getPersonalNote(), null));
 
                     // finally store the cache in the database
                     result.add(geocode);
--- a/main/src/main/java/cgeo/geocaching/models/CacheArtefactParser.java
+++ b/main/src/main/java/cgeo/geocaching/models/CacheArtefactParser.java
@@ -6,6 +6,8 @@ import cgeo.geocaching.location.Geopoint;
 import cgeo.geocaching.location.GeopointFormatter;
 import cgeo.geocaching.location.GeopointParser;
 import cgeo.geocaching.location.GeopointWrapper;
+import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.storage.DataStore;
 import cgeo.geocaching.utils.TextUtils;
 import cgeo.geocaching.utils.formulas.FormulaUtils;
 import cgeo.geocaching.utils.formulas.VariableList;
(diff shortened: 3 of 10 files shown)


eddiemuc:
Example for a cache in three lists red, blue and blume:
image

Is this a feature that should be merged? I am waiting a few days for feedback before proceeding.

murggel:
I am unsure, how this affects removing a cache from a list. Do I have then first activate "prevent waypoint from note" (which we should then rename) to avoid that it is automatically readded (as it is currently the case for deleted waypoints) or is the list removed automatically from the note on removing the cache from the list? 
Or we have to think of a new workflow when parsing of note is done and / or  removing and changing of information which is stored in the note.

Or am I wrong / to complicated and there is a simpler way which I don't know?

eddiemuc:
> I am unsure, how this affects removing a cache from a list. Do I have then first activate "prevent waypoint from note" (which we should then rename) to avoid that it is automatically readded (as it is currently the case for deleted waypoints) or is the list removed automatically from the note on removing the cache from the list? 

That's a valid point. As of now it would be like with waypoints (and as you fear).

When removing a waypoint today, my workflow is:
- delete that waypoint from PN (using "clear" option and/or deleting manual in text)
- delete waypoint 
- if wanted: recreate PN info

Same would be needed for lists too.

I can extend the PR to auto-remove any matching hash-patterns from PN on list removal. But since this was never requested for waypoints (where this would be more complex btw): is this wanted?

eddiemuc:
Idea:
- make the parser ignore lists which are prefixed with `#-`
- when a cache is removed from a list, all `#` entries marching the list are made into `#-`

This would keep changes to user PN at list removal to a minimum

murggel:
One more thought:
 I would like to have a (separate) setting, for me it's not necessary to store the lists in PN and it maybe uses too many characters and the PN gets truncated on upload.

murggel:
And even one more thought:
When importing an exported cache with those note on a different device (e.g. from my cache mait), there could be an option to ignore the lists in PN, maybe she has a different management

eddiemuc:
> Idea:
> 
> * make the parser ignore lists which are prefixed with `#-`
> * when a cache is removed from a list, all `#` entries marching the list are made into `#-`
> 
> This would keep changes to user PN at list removal to a minimum

Done:
image

eddiemuc:
> One more thought: I would like to have a (separate) setting, for me it's not necessary to store the lists in PN and it maybe uses too many characters and the PN gets truncated on upload.

Done:
image
(the third one is the new one)

eddiemuc:
> And even one more thought: When importing an exported cache with those note on a different device (e.g. from my cache mait), there could be an option to ignore the lists in PN, maybe she has a different management

Done: on import, personal notes are modified such that contained lists are marked as ignored (for parser)

murggel:
> > Idea:
> > 
> > * make the parser ignore lists which are prefixed with `#-`
> > * when a cache is removed from a list, all `#` entries marching the list are made into `#-`
> > 
> > This would keep changes to user PN at list removal to a minimum
> 
> Done: image

What happens, if the `#-red` is inside the `{c:geo}-block`? Does that also work?

eddiemuc:
> What happens, if the #-red is inside the {c:geo}-block? Does that also work?

Same as for waypoints + variables: 
every scan sxans the whole note (inside and outside the block) and applies the same behavior to everything.
This also applies to the #- syntax: on lost remove (and GPX import) it is applied to all existing list tags (inside+outside) and on scan it is recognized also both inside and outside.

eddiemuc:
Failing tests are unrelated

eddiemuc:
Thanks @murggel for all the construtive feedback.
I believe I adressed all concerns. Especially the feature can be deactivated in settings. 

I will wait for more feedback until tomorrow morning. If none is given I will merge.

moving-bits:
I have not tested this PR, but from reading the notes above I have one question and one small nit-pick:
- Why do we react on list names outside the {cgeo} block at all?
  - IMHO "adding to list" is different from "creating a waypoint", as for a waypoint you need to write a very specific text format (you would not do that accidentally), whereas for the new list feature you just need to add a "#" in front of some text (which is a common symbol for uncommenting things)
  - I suggest to restrict parsing of list names to {cgeo} block
- Nit pick: I would put "Disable waypoint extraction" and "Disable list extraction" one after another, then "Overwrite PCN" (or "Overwrite PCN" as first, then the two others)

eddiemuc:
> - Why do we react on list names outside the {cgeo} block at all?... for the new list feature you just need to add a "#" in front of some text ... I suggest to restrict parsing of list names to {cgeo} block

I get the point but that would make the feature unusable for my personal use case (managing lists purely in PN, like I do already for waypoints and variables (because the dialog way for all three is way to complicated for my personal taste). I also never use the copy-cache-info-to-pn-thingy.

Suggestion: only parse lines for list names which start with double-hash `##`?

> - Nit pick: I would put "Disable waypoint extraction" and "Disable list extraction" one after another, then "Overwrite PCN" (or "Overwrite PCN" as first, then the two others)

Sure, no problem

bekuno:
Is this a fix or a new feature?
If new, it should be target to master. We have enough to do with optimizing the new marker feature.

murggel:
> > - Why do we react on list names outside the {cgeo} block at all?... for the new list feature you just need to add a "#" in front of some text ... I suggest to restrict parsing of list names to {cgeo} block
> 
> I get the point but that would make the feature unusable for my personal use case (managing lists purely in PN, like I do already for waypoints and variables (because the dialog way for all three is way to complicated for my personal taste). I also never use the copy-cache-info-to-pn-thingy.
> 
> Suggestion: only parse lines for list names which start with double-hash `##`?
> 
 
How about `#+` / `#-` I guess, that combination does not occur often ( `#` / `##` maybe used also as bullet list)

moving-bits:
> How about `#+` / `#-` I guess, that combination does not occur often ( `#` / `##` maybe used also as bullet list)

That might work, should be a combination that happens less frequently by accident.

eddiemuc:
> Is this a fix or a new feature?
> If new, it should be target to master. We have enough to do with optimizing the new marker feature.

Fair point. In this case I would abandon the PR though. Not worth for me waiting until the release AFTER the current beta

moving-bits:
No need to abandon that PR, IMHO - we can still put it into the first release after the upcoming feature release. So still target it to `release`, but merge only after feature release got published. 

We just should not implement new features on `release` branch currently, one day before the deadline by Play Store hits us (\*), and with already so many loose ends in dynamic markers and hierarchical list naming features.

(\*) We won't be able to publish a new bugfix release on top of current release 2026.08.05 starting tomorrow, until we have increased targetAPI to 35 (which for us currently means publishing `release` branch as feature release).

PR #18538, opened 2026-08-30
fix #18504: Use cursor.getColumnIndex in fileInfoFromCursor

### LINKED ISSUE #18504: GPX default name on import (Dropbox)
Reported by user on support [Ticket #487053](https://cgeosupport.droescher.eu/scp/tickets.php?id=10789):

"When I am importing a GPX file to save in a list via dropbox - I select open with "c:geo GPX import" from the list of options in dropbox.  C:geo will open up and show the "Import GPX to List" page.  I then select option and it shows me the new list window.  The list name dialog box always defaults to "application/octet-stream" for the list name.  I have to backspace to delete this and then type in the name of the list I want to create.
Question: can this list name just come up blank so I don't always have to backspace over the "application/octet-stream" text? "

---

Don't have Dropbox installed in my test environment and can't reproduce it with default file explorer:
- export a GPX from c:geo
- open the device's default file explorer and tap on the exported file
- select "open with c:geo" (GPX importer)
- select "Create new list" in c:geo's "Import GPX to list" dialog
- then the correct filename is being displayed in that dialog for me

Anyone having Dropbox installed can try to reproduce this?

murggel:
## Description

Getting fileInfo from Cursor throwed an exception when invoked from an dropbox-uri.
This PR gets the values from cursor via fetching the columns instead of relying on fixed order

## Related issues

fix #18504

### DIFF
--- a/main/src/main/java/cgeo/geocaching/storage/DocumentContentAccessor.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DocumentContentAccessor.java
@@ -217,25 +217,27 @@ class DocumentContentAccessor extends AbstractContentAccessor {
 
         //using dir.listFiles() is FAR too slow. Thus we have to create an explicit query
 
-        return queryDir(dir, new String[]{
-                DocumentsContract.Document.COLUMN_DOCUMENT_ID,
-                DocumentsContract.Document.COLUMN_DISPLAY_NAME,
-                DocumentsContract.Document.COLUMN_MIME_TYPE,
-                DocumentsContract.Document.COLUMN_LAST_MODIFIED,
-                DocumentsContract.Document.COLUMN_SIZE,
-
-        }, c -> fileInfoFromCursor(c, null, dir, folder));
+        return queryDir(dir, FILE_INFO_COLUMNS, c -> fileInfoFromCursor(c, null, dir, folder));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


murggel:
@eddiemuc  I am unsure about the default-values...

eddiemuc (approved):
Fix seems straightforward and fine to me.
Don't know about valid default values tbh. Maybe reject whole entry if column is missing? Or return information like name "error: not found"? For lastmodified maybe use current timestamp?

murggel:
> For lastmodified maybe use current timestamp?

Hm - so if not available it will work like "update not necessary" and not "update may be helpful", so that would kind of a saver way then....

> Fix seems straightforward and fine to me. Don't know about valid default values tbh. Maybe reject whole entry if column is missing? 

Problem with that is, that the entry for the dropBox doesn't have a value for COLUMN_DOCUMENT_ID (which is not really necessary there, because of the existing docUri). 

So maybe I go for a first version only to check the documenId and leave the other values untouched.

murggel:
@eddiemuc  I pushed a version with access to the columns only when its necessary.
This still fixes the issue and does not have unwanted side-effects

eddiemuc (approved):
Lgtm

murggel:
Retest this please

PR #18551, opened 2026-08-31
Move matching filter next to button

murggel:
## Description

This PR rearrange the listed filters next to the respective button to make them more distinguishable from the lists

| current | new | popup |
| --- | --- | --- |
| grafik|grafik |grafik|

### DIFF
--- a/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
+++ b/main/src/main/java/cgeo/geocaching/CacheDetailActivity.java
@@ -1260,7 +1260,8 @@ public class CacheDetailActivity extends TabbedViewPagerActivity
 
             @Override
             protected void onFinished() {
-                CacheInfoBoxes.updateCacheListsAndMatchingFilters(CacheDetailActivity.this.findViewById(R.id.offline_lists), cache, null);
+                CacheInfoBoxes.updateCacheListsAndMatchingFilters(CacheDetailActivity.this.findViewById(R.id.cache_offlinebox),
+                        cache, CacheDetailActivity.this);
             }
         }.execute();
     }
--- a/main/src/main/java/cgeo/geocaching/filters/FilterUtils.java
+++ b/main/src/main/java/cgeo/geocaching/filters/FilterUtils.java
@@ -74,7 +74,7 @@ public class FilterUtils {
     }
 
     /** opens a dialog to activate/deactivate named filter markers */
-    public static void openDialogActivateMarkers(final Activity context) {
+    public static void openDialogActivateMarkers(final Context context) {
         final List<NamedFilter> filters = NamedFilter.getAllWithIcons();
 
         if (filters.isEmpty()) {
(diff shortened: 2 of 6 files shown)


eddiemuc:
I am fine with merging this PR. Just one note:

I thought about implementing it like this PR does first. i didn't because the current text info already merges two unrelated infos: time of last update and assigned lists. So i thought I merge a third info (matching filters) as well. Stricktly speaking we would need to separate also "time of last update" and related "refresh"  and "delete" buttons into its own line as well if we follow the approach of this PR.

murggel:
> I am fine with merging this PR. Just one note:
> 
> I thought about implementing it like this PR does first. i didn't because the current text info already merges two unrelated infos: time of last update and assigned lists. So i thought I merge a third info (matching filters) as well. Stricktly speaking we would need to separate also "time of last update" and related "refresh"  and "delete" buttons into its own line as well if we follow the approach of this PR.

Understood, but since the information of "time of the last update" is somehow related to lists and the filters are a complete separate topic, I wouldn't care

eddiemuc:
> Understood, but since the information of \"time of the last update\" is somehow related to lists and the filters are a complete separate topic, I wouldn't care

In the context of cache details, I see the filters as "dynamic lists" (where caches are assigned/unassigned not by putting them into the list explicitly but implicitly by filter definition). In that sense it is imho related to "static" lists.

But you have a point: in contrast to static lists, filters/dynamic-lists may also apply to non-stored caches. 

So yeah, I am fine with this PR.

murggel:
I am just stumbling about the naming of the dialog title for activating markers.
It looks like the list selection dialog (and have "select markers" as caption), but in fact activate markers. 
When started from CacheDetail and "selecting" a marker, it may not appear in the filter-list, which is correct, but may be irritating first.
I would propose to rename the title of that dialog to "activate marker".

bekuno:
I see also a problem with the names here.
The PR headline says "marker", the description says "Filter", in the images the "marker" symbol is used, but "Filter" as text.

murggel:
> I see also a problem with the names here.
> The PR headline says "marker", the description says "Filter", in the images the "marker" symbol is used, but "Filter" as text.

I guess, that is, because (named) filters are used as marker and depending on the settings, not only markers are listed here.

murggel:
> > I see also a problem with the names here.
> > The PR headline says "marker", the description says "Filter", in the images the "marker" symbol is used, but "Filter" as text.
> 
> I guess, that is, because (named) filters are used as marker and depending on the settings, not only markers are listed here.
> 

We could change the text depending on that setting to filter  resp. marker

eddiemuc:
It is exactly as @murggel suspected: 
In cache details, named filters are used as something else than just markers (more as dynamic lists). This is why I choose the name "filter" for related GUI (not marker). Introducing a new term like "Dynamic lists" felt too confusing to me.

As for the marker icon on the button: I am not too happy about it, but it felt wrong to use a different symbol than used in map and lists. 
Edit: Using a "filter" icon would probably even more confusing. I tried to pitch a different symbol back then (a filter with a save icon as well as a star icon), but that was rejected last team meeting (granted, Copilot didn't do a great job creating it).

Maybe simply change the second option text from "Only if marker is active" to "Only filters  activated as marker"?

eddiemuc:
I just saw that the original request suggested wording "Show only filters with an active marker in the listing". Would be fine for me too. See https://github.com/cgeo/cgeo/issues/18500#issuecomment-5407596577

There was also a suggestion to have a separate selection for filters used in Cache Details (so one "activation flag" per filter for marker usage and another one for dynamic list display in cache details) here: https://github.com/cgeo/cgeo/issues/18500#issuecomment-5378390591 Felt like too much complexitiy to me at the time.

moving-bits:
I'm also in favor of this change. Just one question: Can we print a "Filters: none matching" (or something like that) if there are no filters with icons? Otherwise we have a huge blank space with just that single button on the right and nothing else in that line.

eddiemuc:
> I'm also in favor of this change. Just one question: Can we print a "Filters: none matching" (or something like that) if there are no filters with icons? 

Currently, it will print something similar in that case:
Filters: -

murggel:
> > I'm also in favor of this change. Just one question: Can we print a "Filters: none matching" (or something like that) if there are no filters with icons? 
> 
> Currently, it will print something similar in that case:
> Filters: -
> 
> 

It seems that currently nothing is shown, but I will check and add "Filters: -" if it is not already there

murggel:
> > I'm also in favor of this change. Just one question: Can we print a "Filters: none matching" (or something like that) if there are no filters with icons? 
> 
> Currently, it will print something similar in that case:
> Filters: -
> 
> 


For lists "Not available offline" is shown, so maybe we show "No matching filters" ?

grafik grafik

PR #18555, opened 2026-09-01
Fix #18520: make emoji drawable copies + fix: dont refresh cachelist on constant coords

### LINKED ISSUE #18520: markers: Changing marker priority (leading to a switch from left to right) is not applied until after restart of c:geo
Initial situation: couple of markers defined with given priority (green first, red second)

Image

Image

--> everything is fine

Now changing priority (move up the "red" ones to the top):

Image

--> icons stay as they are (green left, red right). 

Image

This is wrong, change of priority must lead to red left, green right):

After c:geo restart it is displayed correctly

Image

This effect ONLY occurs when change of priority leads to same markers (but different order). If change of priority leads to displaying other icons, then the update is done correctly.

eddiemuc:
rel to #18520: dont refresh cachelist when coordinates are not changing

while investigating #18520 on an emulator with nonchanging GPS coordinate, I noticed that the cache list is regularly refreshed (and markers for all caches are re-requested from mapmarkerutils, thus I presume they are also redrawn). This is caused by regular location update checks. This PR introduces a check so the cache list is not redrawn when location has not changed (or changed insignificantly, meaning less than 1m, conservatively calculated)

### DIFF
--- a/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
+++ b/main/src/main/java/cgeo/geocaching/activity/AbstractActionBarActivity.java
@@ -3,6 +3,7 @@ package cgeo.geocaching.activity;
 import cgeo.geocaching.R;
 import cgeo.geocaching.enumerations.CacheListType;
 import cgeo.geocaching.enumerations.CacheType;
+import cgeo.geocaching.maps.CacheMarker;
 import cgeo.geocaching.models.Geocache;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.ui.ViewUtils;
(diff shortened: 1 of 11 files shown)


eddiemuc:
PR now fixes #18520 and is ready for review/merge

eddiemuc:
Note to self:
- call mutate on drawable copy!
- place that logic already into EmojiUtils! It might cause same kind of problems elsewhere otherwise 
- maybe drawable -helper in ViewUtils?

eddiemuc:
> Note to self:
> 
> * call mutate on drawable copy!
> * place that logic already into EmojiUtils! It might cause same kind of problems elsewhere otherwise

done

Issue #18138, opened 2026-05-27
c:geo creates two logs (found and not-found)


eddiemuc:
Ticket https://cgeosupport.droescher.eu/scp/tickets.php?id=10315
Version 2026.05.21

A user claims that in areas with poor reception c:geo creates two logs, one found and one not-found. I can't reproduce this, does anyone know about such a problem?

When you create a FOUND log in areas with poor reception, c:geo
automatically creates two logs: one FOUND and one NOT FOUND.
If you were attaching a photo (e.g., a signed logbook), that photo
will only appear in the NOT FOUND LOG.
So, each time, you have to go to the "NOT FOUND" LOG, save the photo,
delete the "NOT FOUND" LOG, go to the "FOUND" LOG, and add the newly
saved photo...
It's quite laborious and frustrating...
And if you happen to be in an area with poor reception and create a
lot of logs, it becomes time-consuming... if you even notice that the
system created a double log.

MagpieFourtyTwo:
Have seen such a weird double log in a cache I visited the other day, too.
But I didn't pay any attention to, because the user who logged it doesn't seem to be the most tech-savvy anyway.
And in fact I don't even know if he uses c:geo at all ... but if you want to investigate further, I can send you the GC-Code.

eddiemuc:
> ... but if you want to investigate further, I can send you the GC-Code.

If you can easily retrieve this code (without searching in your history for a long time) then yes. Although I doubt it will ahed more light into this. I really can't think of an way such an unintentional double-log can happen

ztNFny:
I know such situations with Found+Note. That happens when you think the log wasn't sent, then log it again (manually) but as there's already a found log log type will be auto-changed to Note in the Log dialog - but the user will not notice it and send the log again.
I don't know of any scenarios where this happens automatically and wouldn't even be able to think of one where it changes to DNF.

murggel:
+1 from Support [Ticket #198906](https://cgeosupport.droescher.eu/scp/tickets.php?id=10421)

murggel:
Some more feedback from user 

I have had this happen a few more times.  I think it is always wherebtherebis a poor connection.  I also thinknit is when I am attaching a photo.  I think the found it log posts without the photo, then cgeo thinks it didnt post, so prompts me to save offline.  Then it looks like dnf since found it is no longer an option.   Then when you try to post the saved log you cannot edit to found it since there actually is one posted.

ztNFny:
[...]

PR #18558, opened 2026-09-02
rel #18138: fix wrong log-type

murggel:
## Description

On updateing available log-types, keep current log type if still available, 
otherwise set to NOTE or first available log type

## Related issues

rel #18138

## Additional context

The fix does not deal with the fact, that two logs are send to the service, but maybe because of the different log-type a second log was send, which in a found case the server would have rejected...


murggel:
Superseeded by #18591

PR #18561, opened 2026-09-03
fix #16206: diacritic-insensitive top search bar

### LINKED ISSUE #16206: Top search bar does not collate
### Describe your problem!

At the top of the c:geo main screen there is a search bar. 
One can quickly search for caches using it. 

However, one has to be precise in typing regarding regional character variations. 

### How to reproduce?

Have any number of caches stored whose titles contain regional characters. 
Type a search using the basic latin character set. 

search-collating

As seen here, the initial partial name matched up to the "ş" character. (which to make matters more complicated, for Romanian has 2 variants, visually almost identical, though one has been declared "legacy"). 

A usual database search, this including most web apps the user is accustomed to, will support collation and the returned results will contain items with equivalent characters. 
In the example given, the search for `la dus` should also return `la duş`. 

### Reproducible

Yes

### c:geo Version

2024.09.18

### System information

_No response_

### Additional Information

_No response_

cylorak:
**What**
The top search bar now matches diacritic- and case-insensitively across **cache names**, **cache owner names** and **trackable names**, e.g. `la dus` finds `la duş`/`la duș`, `strasse` finds `straße`, `zurich` finds `Zürich`. Closes #16206.

**How**
- New `TextUtils.foldDiacritics` (Unicode NFD + strip combining marks, plus an explicit map for atomic letters without decomposition: ß→ss, ø→o, æ→ae, œ→oe, ł→l, đ/ð→d, þ→th, ħ→h, ŧ→t, ŋ→n, ĸ→k, ſ→s, ı→i; both Romanian ş variants U+015F/U+0219 → s) and `containsIgnoreCaseAndDiacritics`.
- New normalized shadow columns `name_normalized` + `owner_normalized` (cg_caches) and `title_normalized` (cg_trackables), filled on write.
- `DataStore.findCaches()` matches cache name and owner, `findTrackables()` matches trackable name — both query the normalized columns, with a raw-`LIKE` fallback for rows not yet migrated so search stays correct during migration.
- Existing rows are filled by a throttled, chunked background task started from `MainActivity` (duty-cycle delay between chunks, power-save aware), showing a Toast on completion.
- @moving-bits : Just let me know if I should provide a complete conversion table if needed. 

**Tests**
- Pure unit test `TextUtilsDiacriticsTest` (combining diacritics, atomic letters, casing incl. capital ẞ, Vietnamese horn letters, dotted-İ, both ş variants, `containsIgnoreCaseAndDiacritics`).
- Frontend verified on the emulator (screenshots below).

**Screenshots**

Cache **name** match `!NameDemo Þórsmörk` **and** owner match `!OwnerDemo Iceland` for the ASCII query `thorsmork` (owner not shown in the drop-down):
01_thorsmork

Opening `!OwnerDemo Iceland` reveals the owner `Þórsmörk` — i.e. the query matched via the owner name:
02_ownerdemo_detail

Both Romanian ş variants (U+015F and U+0219) plus the œ ligature, for the original report's query `la dus`:
03_la_dus

Trackable **name** matches for `zurich` (trackable icon, TB codes):
05_zurich_tb

**Notes**
- DB version 112 is marked downward-compatible (nullable columns via `createColumnIfNotExists`, re-runnable — like 108/109).
Remark: after a downgrade an older version won't maintain the `*_normalized` columns, so a cache edited there keeps a stale folded value until re-saved (the NULL-driven migration only refills NULLs). 
Let me know to drop 112 from `DBVERSIONS_DOWNWARD_COMPATIBLE` if you'd rather block downgrades.
- AI-assisted, reviewed and tested by me.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/MainActivity.java
+++ b/main/src/main/java/cgeo/geocaching/MainActivity.java
@@ -512,6 +512,8 @@ public class MainActivity extends AbstractNavigationBarActivity {
             cLog.add("perm");
 
             init();
+
+            DataStore.migrateNormalizedColumnsIfNeeded(this);
         }
 
         if (Log.isEnabled(Log.LogLevel.DEBUG)) {
--- a/main/src/main/java/cgeo/geocaching/storage/DataStore.java
+++ b/main/src/main/java/cgeo/geocaching/storage/DataStore.java
@@ -5,6 +5,7 @@ import cgeo.geocaching.DBInspectionActivity;
 import cgeo.geocaching.Intents;
 import cgeo.geocaching.R;
 import cgeo.geocaching.SearchResult;
+import cgeo.geocaching.activity.ActivityMixin;
 import cgeo.geocaching.connector.ConnectorFactory;
 import cgeo.geocaching.connector.capability.ILogin;
 import cgeo.geocaching.connector.internal.InternalConnector;
@@ -91,6 +92,7 @@ import android.database.sqlite.SQLiteOpenHelper;
 import android.database.sqlite.SQLiteStatement;
 import android.location.Location;
 import android.net.Uri;
+import android.os.PowerManager;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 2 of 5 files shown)


moving-bits:
ok to test

PR #18565, opened 2026-09-03
Fix Kotlin metadata version mismatch in gradle build

bekuno:
On some Tasks `:main:lintAnalyzeBasicDebug*` we got errors like: 
`e: file:///home/jenkins/.gradle/caches/8.14.5/transforms/94ec4645bb1424a1995ad3e6f840a2ad/transformed/chartLib-5.2.4/jars/classes.jar!/META-INF/chartLib.kotlin_module Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.4.0, expected version is 2.2.0.`

So we need to configure the kotlin-d8-r8-version to a newer level until AGP/Gradle is migrated to version 9.x.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,9 +2,13 @@
 buildscript {
     repositories {
         google()
+        maven { url 'https://storage.googleapis.com/r8-releases/raw' }
     }
     dependencies {
         // these dependencies are used by gradle plugins, not by our projects
+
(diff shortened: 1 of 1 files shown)


bekuno:
retest this please

bekuno:
retest this please

bekuno:
replaced by #18567

PR #18567, opened 2026-09-04
Fix Kotlin metadata version mismatch in gradle build

bekuno:
The prepared solution in #18565 does not work as expected. Therefore another solution: downgrade chartlib and downgrade kotlin-stdlib strategy to 2.2.10.


bekuno:
PR is based to release, as the described problem is also there.

PR #18571, opened 2026-09-05
Quick fix to raise targetSDK to 36

moving-bits:
## Description
- sets targetSDK to 36
- opts out of the new [predictive back gestures](https://developer.android.com/about/versions/16/behavior-changes-16#predictive-back), which are enabled by default on Android 16+ devices (API 36)


bekuno:
Also opt out for "Device form factors" behaviour change https://developer.android.com/about/versions/16/behavior-changes-16#large-screens-form-factors ?

For this we can set a property, see
https://developer.android.com/about/versions/16/behavior-changes-16#temporary-opt-out

PR #18575, opened 2026-09-06
Fix #18573: fix overlapping TextViews in offline box

### LINKED ISSUE #18573: Lists overlapping Filters
### Describe your problem!

With the latest Beta (2026.09.05-RC) the shortly gone filters section inside the cache detail page came back.

If a cache is assigned to several lists, this section is overlapping with the new markers section:

Image

### How to reproduce?

Just add a cache to several lists

See explanation and screenshot above

Avoid overlapping by moving the markers section basing on the needed space of all list entries. 

### Reproducible

Yes

### c:geo Version

2026.09.05-RC

### System information

```text

```

### Additional Information

_No response_

murggel:
## Description

 Fix overlapping TextViews in offline box, reuse separate / existing layout for named filter

## Related issues

Fix #18573

## Additional context

Disadvantage of own text-box is now, that it will take more space in cases with multiple / longer lists

grafik

### DIFF
--- a/main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
+++ b/main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
@@ -204,12 +204,14 @@ public class CacheInfoBoxes {
         offlineLists.setMovementMethod(LinkMovementMethod.getInstance());
         offlineLists.setVisibility(builder.length() > 0 ? View.VISIBLE : View.GONE);
 
-        final View marker = view.findViewById(R.id.marker_button);
-        if (marker != null) {
+        final View markerBox = view.findViewById(R.id.namedfilter_box);
+        if (markerBox != null) {
             final Context context = view.getContext();
             final boolean enableMarkerButton = context != null && Settings.getNamedFilterDisplayMode() != Settings.NamedFilterDisplayMode.NONE;
-            marker.setVisibility(enableMarkerButton ? View.VISIBLE : View.GONE);
-            if (enableMarkerButton) {
+            markerBox.setVisibility(enableMarkerButton ? View.VISIBLE : View.GONE);
+            
+            final View marker = view.findViewById(R.id.marker_button);
+            if (marker != null && enableMarkerButton) {
                 FilterUtils.registerFilterActivateDeactivateButton(context, marker);
             }
         }
--- a/main/src/main/res/layout/cache_namedfilterbox.xml
+++ b/main/src/main/res/layout/cache_namedfilterbox.xml
@@ -1,33 +1,41 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@+id/namedfilter_box"
-    android:layout_width="fill_parent"
+    android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:orientation="vertical"
-    android:visibility="gone">
+    android:baselineAligned="false"
+    android:orientation="vertical">
 
-    <View
-        style="@style/separator_horizontal"
-        android:layout_marginBottom="9dp"
-        android:layout_marginTop="9dp" />
+    <LinearLayout style="@style/separator_horizontal_layout">
 
+        <View style="@style/separator_horizontal" />
+    </LinearLayout>
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:baselineAligned="false"
+        android:orientation="horizontal">
     <TextView
-        android:id="@+id/namedfilter_text"
-        android:layout_width="fill_parent"
+        android:id="@+id/marker_text"
+        android:layout_width="0dp"
         android:layout_height="wrap_content"
-        android:layout_gravity="left"
+        android:layout_weight="1"
         android:layout_marginLeft="6dip"
-        android:layout_marginRight="6dip"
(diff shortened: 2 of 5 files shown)


murggel:
Disadvantage of own text-box is now, that it will take more space in cases with multiple / longer lists

grafik



Maybe we revert to the previous state and add only an empty line to make it more visible. But then the text is not aligned to the marker-button

moving-bits:
I have consulted AI for it, proposal is to use a ConstraintLayout for the buttons and their text:

(Have not tested it yet, though, I'm not at my dev machine currently)

```
<androidx.constraintlayout.widget.ConstraintLayout 
    xmlns:android="http://android.com"
    xmlns:app="http://android.com"
    android:padding="16dp">

    <!-- FIRST BUTTON (Top Right) -->
    <com.google.android.material.button.MaterialButton
        android:id="@+id/button1"
        android:layout_width="wrap_content"
        android:text="Button 1"
        app:layout_constraintTop_toTopOf="parent"
        app:layout_constraintEnd_toEndOf="parent" />

    <!-- FIRST TEXT (Left of Button 1, Top-Aligned) -->
    <!-- app:layout_constraintBottom_to_top of text2 prevents it from overlapping text2 -->
    <TextView
        android:id="@+id/text1"
        android:layout_width="0dp"
        android:text="This is the first explanatory text. It can sometimes get longer than one line."
        android:layout_marginEnd="16dp"
        app:layout_constraintTop_toTopOf="@+id/button1"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintEnd_toStartOf="@+id/button1"
        app:layout_constraintBottom_toTopOf="@+id/text2"
        app:layout_constraintVertical_bias="0.0"
        app:layout_constraintVertical_chainStyle="packed" />

    <!-- SECOND BUTTON (Below Button 1, Fixed Position) -->
    <com.google.android.material.button.MaterialButton
        android:id="@+id/button2"
        android:layout_width="wrap_content"
        android:text="Button 2"
        android:layout_marginTop="16dp"
        app:layout_constraintTop_toBottomOf="@+id/button1"
        app:layout_constraintEnd_toEndOf="parent" />

    <!-- SECOND TEXT (Left of Button 2, Top-Aligned by default) -->
    <TextView
        android:id="@+id/text2"
        android:layout_width="0dp"
        android:text="This is the second explanatory text."
        android:layout_marginEnd="16dp"
        android:layout_marginTop="16dp"
        app:layout_constraintTop_toTopOf="@+id/button2"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintEnd_toStartOf="@+id/button2"
        app:layout_constraintBottom_toBottomOf="parent" />

</androidx.constraintlayout.widget.ConstraintLayout>
```

"Why this works:
- Fixed Buttons: button2 is constrained directly to the bottom of button1. No matter how long text1 gets, it cannot push button2 down because button2 completely ignores the text views.
- Top-Alignment: text1 is top-aligned to button1, and text2 is top-aligned to button2 using app:layout_constraintTop_toTopOf.
- The "Push" Behavior: text1 and text2 form a vertical chain. Because text1 has app:layout_constraintBottom_toTopOf="@+id/text2", if text1 grows too long, its bottom edge will naturally push text2 downward, safely moving it away from the overlapping text area while leaving button2 exactly where it is."

murggel:
I don't think implementation would be the problem.
Question is, what do we want...

Do we want 
* have the "Filter text" aligned to the button (but then we have the gap between the buttons and hence use more room as necessary 
* compact layout but then the list of filter is not aligned to the button

Will provide screenshots later

murggel:
Ah or is your approach a combination? 
If room, then fill the space with the filter-list until it's bottom aligned? 
Only if it needs more space then move the bottom?

moving-bits:
> Ah or is your approach a combination? If room, then fill the space with the filter-list until it's bottom aligned? Only if it needs more space then move the bottom?

Yes: top-align texts and related buttons, if possible, but don't move filter button down, if list of lists gets too long

murggel:
> > Ah or is your approach a combination? If room, then fill the space with the filter-list until it's bottom aligned? Only if it needs more space then move the bottom?
> 
> Yes: top-align texts and related buttons, if possible, but don't move filter button down, if list of lists gets too long

But move button down if it gets longer than the bottom of the button?
Otherwise I would put it in the same box...

murggel:
I will try it with the constraint layout - thanks for the hint

murggel:
With constraint layouts - Not perfect, but not so bad

| Scenario | Image | Remark|
| --- | --- | --- |
| short list - short filters |grafik<p><p>grafik | ok|
| long list - short filters |grafik| should be bottom aligned, but hard to achieve|
| short list - long filters|grafik| should be bottom aligned, but ok|
|long list - long filters |grafik| ok|

murggel:
I will try other models now, starting with that base...

bekuno:
Looking at all the options, I think it’s much better to use a little more space to clearly separate the two sections than to save a bit of space at the expense of readability. I would therefore prefer a solution like the last option (long list – long filters) in all situations.
I would also generally justify the text and icon horizontally. That would fit in best with other entries ("No matching filter", watchlist, Adventure). And also justify other below (e.g. Chirp)

eddiemuc:
I see two options:

Option 1: we feel that filters and lists are unrelated:
-> then we should place each in their own "box" (with a line inbetween like eg between attributes and lists).

Option 2: we feel that filters and lists are related:
-> then we should revert to the original state (put filters under lists as text)

murggel:
> I see two options:
> 
> Option 1: we feel that filters and lists are unrelated:
> -> then we should place each in their own "box" (with a line inbetween like eg between attributes and lists).
> 
> Option 2: we feel that filters and lists are related:
> -> then we should revert to the original state (put filters under lists as text)

I feel list and filters are not related...

So I would vote for a separate box with own alignment

murggel:
Own box with separator

grafik

murggel:
Then I would give it a try in the nightly and merge this evening

PR #18576, opened 2026-09-06
Update database inspection view

moving-bits:
## Description
Updates the built-in database inspection view to support multi-column sorting, better filtering etc.

### DIFF
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -367,7 +367,7 @@ dependencies {
     implementation 'net.sf.geographiclib:GeographicLib-Java:2.1'
 
     // DB inspection (new library)
-    implementation 'com.github.moving-bits:grid:-SNAPSHOT'
+    implementation 'com.github.moving-bits:grid:7fccf69583'
 
     // Jackson XML processing
     String jacksonVersion = '2.21.6'
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -35,6 +35,7 @@ android {
     compileSdk = 36
 
     compileOptions {
+        // support new Java language features on older devices
         coreLibraryDesugaringEnabled = true
         sourceCompatibility = JavaVersion.VERSION_17
         targetCompatibility = JavaVersion.VERSION_17
@@ -366,7 +367,7 @@ dependencies {
     implementation 'net.sf.geographiclib:GeographicLib-Java:2.1'
 
     // DB inspection (new library)
-    implementation 'com.github.moving-bits:grid:12e7675c6e'
+    implementation 'com.github.moving-bits:grid:-SNAPSHOT'
 
     // Jackson XML processing
     String jacksonVersion = '2.21.6'
--- a/main/build.gradle
+++ b/main/build.gradle
@@ -35,12 +35,9 @@ android {
     compileSdk = 36
 
     compileOptions {
-        // https://developer.android.com/studio/write/java8-support
-        // enable Java 8 APIs like Streams on older Android versions
         coreLibraryDesugaringEnabled = true
-        // use the diamond operator, lambdas and some other Java 8 language features in Android
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
     }
 
     defaultConfig {
@@ -368,9 +365,8 @@ dependencies {
     // GeographicLib
     implementation 'net.sf.geographiclib:GeographicLib-Java:2.1'
 
-    // DB inspection
-    implementation 'com.github.moving-bits:datatables:v0.1.0'
-    implementation 'com.github.moving-bits:dbinspection:v0.1.2'
+    // DB inspection (new library)
+    implementation 'com.github.moving-bits:grid:12e7675c6e'
 
     // Jackson XML processing
     String jacksonVersion = '2.21.6'
--- a/main/src/main/java/cgeo/geocaching/DBInspectionActivity.java
+++ b/main/src/main/java/cgeo/geocaching/DBInspectionActivity.java
@@ -4,10 +4,13 @@ import cgeo.geocaching.activity.AbstractActionBarActivity;
 import cgeo.geocaching.databinding.DbinspectionActivityBinding;
 import cgeo.geocaching.settings.Settings;
 import cgeo.geocaching.storage.DataStore;
+import cgeo.geocaching.storage.extension.OneTimeDialogs;
+import cgeo.geocaching.ui.SimpleItemListModel;
 import cgeo.geocaching.ui.TextParam;
 import cgeo.geocaching.ui.dialog.Dialogs;
 import cgeo.geocaching.ui.dialog.SimpleDialog;
 import cgeo.geocaching.utils.LocalizationUtils;
+import cgeo.geocaching.utils.Log;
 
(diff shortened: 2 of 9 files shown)


main/build.gradle
```diff
@@ -35,12 +35,9 @@ android {
     compileSdk = 36
 
     compileOptions {
-        // https://developer.android.com/studio/write/java8-support
-        // enable Java 8 APIs like Streams on older Android versions
```
bekuno, line 39: Keep the comments, please

main/build.gradle
```diff
@@ -35,12 +35,9 @@ android {
...
         coreLibraryDesugaringEnabled = true
-        // use the diamond operator, lambdas and some other Java 8 language features in Android
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
```
bekuno, line 41: Is the switch to Java 17 now necessary?

bekuno (changes requested):

main/build.gradle
moving-bits: I'm using some constructs that Java 1.8 does not support, eg. record classes and switch expressions. Both are "semantic sugar", but make things easier... Accordings to the AI I had consulted before doing that switch, it should not harm our installations, as R8 will "desugar" it during compile. MinSdk 26 is supported (even MinSdk 24 would be), according to Gemini.

With regard to our development environments, the change should not cause any trouble, as Java 17 is supported since AS Flamingo or so, and even mandatory minimum runtime environment for more recent AS versions.

main/build.gradle
moving-bits: I had removed them as they refer to our upgrade to Java 8 quite some time ago and no longer apply to Java 17, but I could upgrade them to Java 17, if requested.

bekuno: It refers to the line `coreLibraryDesugaringEnabled = true` which we need further on.
A comment for this schould be there. Probable something shorter (one liner), e.g.
`// enable newer Java API functions on older Android versions`

main/build.gradle
bekuno: Ok.

main/build.gradle
moving-bits: is updated meanwhile

eddiemuc:
Since this PR changed the source/target to Java 17: does this mean we can use Java 17 elements now in c:geo code?
(-> I add this to discussion topics of dev meeting)

moving-bits:
> does this mean we can use Java 17 elements now in c:geo code?

yes (at least language constructs, but not all library functions)

desugaring lib will ensure that it works on older Android systems

---

Here's some more detailed AI-generated answer:

<html><body>
<html><head></head><body>
With minSdkVersion 26 (Android 8.0) and coreLibraryDesugaring enabled, your app's compilation will process smoothly, but you cannot use the entire Java 17 surface area. The restrictions break down into two categories: [1, 2] 

## 1. Java Language Features (Syntax)
The Android build system (D8/R8) compiles Java 17 bytecode down to Android Dex bytecode. Most modern syntax features work flawlessly on API 26 because they are purely compiled down to structural equivalents: [3, 4, 5] 

* 
* Fully Supported: Switch expressions (Java 14), Text Blocks (Java 15), instanceof Pattern Matching (Java 16), and Sealed Classes (Java 17).
* Conditionally Supported: Java Records (Java 16) require coreLibraryDesugaring version 2.0.0 or higher. If you use an older 1.x.x desugaring library, Records will crash on devices below Android 14 (API 34) due to the missing runtime class java.lang.Record. [1, 3, 6] 
* 

## 2. Java APIs (The Standard Library Trap)
This is where you will face the strict limitations. Setting targetCompatibility = JavaVersion.VERSION_17 allows your IDE to auto-complete newer Java 11-17 methods, but desugaring does not backport all of them. [1, 7] 
The [[Android Studio API Support Table](https://developer.android.com/studio/write/java8-support-table)](https://developer.android.com/studio/write/java8-support-table) dictates what works. When using desugar_jdk_libs:2.0.+ on API 26: [6, 8] 

| Java Package / API | Status on Android API 26+ |
|---|---|
| java.time (Date/Time) | Fully Supported (Natively present in API 26). |
| java.util.stream (Streams) | Fully Supported (Desugared backport available). |
| java.net.http.HttpClient (Java 11 Http) | Not Supported. It is entirely absent from Android. You must use OkHttp. |
| New String / Collection Methods | Partially Supported. Methods like List.of() or String.isBlank() work, but several specialized utility methods added between Java 11 and 17 are not desugared. |
| java.nio.file (NIO Path APIs) | Partially Supported. Requires explicitly upgrading to the heavier desugar_jdk_libs_nio artifact to avoid runtime crashes. |

## How to protect your app from runtime crashes
To ensure you don't accidentally call an unsupported Java 17 API that slips past the compiler: [8, 9] 

   1. Rely on Android Lint: Android Studio will automatically warn you with an error ("API defined in library desugaring but not supported by minimum SDK") if you attempt to use an invalid API. Never suppress or ignore these specific Lint warnings.
   2. Use the modern artifact: Ensure your build.gradle file utilizes the modern 2.x desugaring dependency:
   
   dependencies {
       coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.4' // Or latest 2.x version
   }
   
   [6, 9] 

Would you like me to check if a specific Java 11-17 class or method you intend to use is safe to deploy under your current configuration?

[1] [[https://developer.android.com](https://developer.android.com/build/jdks)](https://developer.android.com/build/jdks)
[2] [[https://dev.to](https://dev.to/tkuenneth/java-on-android-less-sugar-more-fun-b57)](https://dev.to/tkuenneth/java-on-android-less-sugar-more-fun-b57)
[3] [[https://github.com](https://github.com/mozilla/rhino/issues/2394)](https://github.com/mozilla/rhino/issues/2394)
[4] [[https://abdulmughnialfikri.medium.com](https://abdulmughnialfikri.medium.com/desugaring-in-android-using-modern-java-features-on-older-devices-f8667db6556c)](https://abdulmughnialfikri.medium.com/desugaring-in-android-using-modern-java-features-on-older-devices-f8667db6556c)
[5] [[https://github.com](https://github.com/adaptech-cz/Tesseract4Android/issues/63)](https://github.com/adaptech-cz/Tesseract4Android/issues/63)
[6] [[https://medium.com](https://medium.com/@sivavishnu0705/desugaring-in-android-the-magic-potion-for-modern-kotlin-on-every-device-900cd3c20b56)](https://medium.com/@sivavishnu0705/desugaring-in-android-the-magic-potion-for-modern-kotlin-on-every-device-900cd3c20b56)
[7] [[https://blog.devgenius.io](https://blog.devgenius.io/unlocking-modern-java-features-in-android-with-corelibrarydesugaringenabled-c8cd37698a13)](https://blog.devgenius.io/unlocking-modern-java-features-in-android-with-corelibrarydesugaringenabled-c8cd37698a13)
[8] [[https://developer.android.com](https://developer.android.com/studio/write/java8-support-table)](https://developer.android.com/studio/write/java8-support-table)
[9] [[https://github.com](https://github.com/microsoftgraph/msgraph-sdk-java-core/issues/252)](https://github.com/microsoftgraph/msgraph-sdk-java-core/issues/252)

bekuno:
`coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5` is already integrated.

PR #18582, opened 2026-09-08
Introducing a new unified gpx parser

eddiemuc:
Introducing a new unified gpx parser.

```
/**
 * Unified, namespace-tolerant "business element" pull-parser for GPX files relevant to Geocaching. Reacts on
 * Geocaches, Waypoints, (Named) Coordinates, Routes and Tracks instead of raw XML elements, via
 * {@link IGPXParseHooks}.
 * <br>
 * <ul>
 *     <li>Handles GPX 1.0 and GPX 1.1 (and mixtures thereof) in one unified code path (no version-specific subclasses).</li>
 *     <li>Does not rely on namespaces being correctly declared; elements/attributes are identified by local
 *     name only (namespace prefixes, if any, are simply stripped).</li>
 *     <li>Self-contained: will only use information from GPX file(s) in created objects.</li>
 *     <li>Multi-file-capability: multiple calls to "parse" may be used to scan related GPX files (eg from a ZIP file)</li>
 *     <li>Handling of enhancements such as GSAP, Groundspeak and c:geo's own tags (via dedicated methods).</li>
 * </ul>
 * <br>
 * {@link #parse} can be called repeatedly, in any order, on the same instance to handle multi-GPX-file-situations (eg from ZIP file).
 * Session-scoped state (esp the cross-file name→geocode index) is <em>not</em> cleared automatically between calls,
 * so resolution works cross-file. Call {@link #reset()} to explicitly start a new, unrelated session on a reused instance.
 */
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,7 @@ captures/
 # ignore Claude internal files
 .claude/
 
+# allow for temp local filees
+temp/
+
 main/usage.txt
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,6 +3,7 @@ org.gradle.daemon=true
 
 # build addons in parallel to main app
 org.gradle.parallel=true
+org.gradle.tooling.parallel=true
 
 # Use more memory to benefit from in-process dexing
 # Use the parallel garbage collector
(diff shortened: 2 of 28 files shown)


eddiemuc:
This is obviously WIP

murggel:
Correct handling of utf8 (#18589) is fixed in PR #18592, but should not have any impact on this PR, except for some tests maybe

PR #18583, opened 2026-09-09
Fix HealthScoreGeocacheFilter: dont include null values

eddiemuc:
Fix HealthScoreGeocacheFilter: dont include null values

The bug made the feature pretty much unusable because every unknown cache (= not stored offline or no health score calc running) was marked as "unhealthy". Fixed with this PR. I had to extend NumberFilter capabilities for this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,5 @@ captures/
 .claude/
 
 main/usage.txt
+
+temp/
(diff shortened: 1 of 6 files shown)


eddiemuc:
retest this please

eddiemuc:
Retest this please

PR #18586, opened 2026-09-10
Support multiple user-defined tile providers

magma1447:
## Description

Replaces the single user-defined tile provider with a list, so several can be configured,
named and switched between. They are managed under *Settings → Map Sources* in the same
way as the log templates: one row per provider with an inline delete icon, and an "add"
button below.

**This supersedes #18494 by @fabiancz.** More on that below - the short version is that
this started from that pull request's review and takes what came out of it.

Details worth calling out:

- **An existing provider carries over.** Somebody who has configured the single
  user-defined provider today will find it in the list after updating, still selected as
  their map source if it was. Nothing is lost and nothing has to be re-entered.
- The list is stored in a preference of its own. As long as that new preference is empty,
  the old one is read instead and its provider shown as the single entry in the list. The
  old preference is never written to or cleared, so downgrading to a version without this
  feature still finds the provider it knows about.
- The entry carried over that way keeps a fixed key, so the tile provider id derived from
  it stays byte-identical to the one used before. That id is what `pref_tileprovider`,
  `pref_previous_tileprovider` and `pref_tileprovider_hidden` store, so none of them needs
  rewriting and the map source selection survives untouched.
- Deriving the id from the map Uri alone is not enough once there is more than one
  provider, since several may share a host - `getId()` returned `<class>:null` for every
  user-defined provider, so they would collide in `TileProviderFactory`.
- Uris are validated and their placeholders normalised, accepting `{x}`, `{y}` and `{z}` in
  either case. A bare `scheme://host` stays valid, as before, and gets the default tile
  path appended.
- Saving validates the Uri by actually fetching one tile from it, so a typo in the host or
  path is caught at once instead of showing up later as a map that stays empty. The check
  is advisory: if it fails, the reason is shown and the provider can still be saved anyway,
  since a server may cover only some regions or be temporarily unavailable.
- **Sharing a provider with somebody else.** Copying one puts a single line of plain text
  on the clipboard - the name and the Uri - which can be sent through any chat client or
  mail. At the other end, pasting it into either field of the add dialog is recognised and
  fills in both. A bare Uri without a name works as well, so a link found on a website can
  be pasted straight in. The text stays readable, so the recipient can see what they are
  about to add before adding it.
- Unit tests cover the url handling, the shareable form and the stored json.

### Screenshots

Screenshot_20260910-140300

Screenshot_20260910-140335

The configured providers appear for both renderers - here with a raster one in use, and
with its VTM counterpart:

<table>
<tr>
<td>
Screenshot_20260910-140425
</td>
<td>
Screenshot_20260910-140440
</td>
</tr>
</table>

## Related issues

- Supersedes #18494 — *Add support for multiple custom online maps*.
- #15884 — *Custom parameters for user-defined tile provider URL*, and #15714, which added
  the single user-defined provider this builds on.

## Additional context

**On superseding #18494.** That pull request asked for exactly this feature and its review
produced the requirements this implementation follows: a new preference with the data
migrated from the old one rather than the old key being reused, individual tile provider
ids, url validation and normalisation with case-insensitive placeholders, unit tests, and
green Checkstyle and lint. @fabiancz also raised the replace-versus-extend question that
was never resolved, and @moving-bits suggested that a combination of the two proposals
would be ideal. This pull request is that combination. The question of how to proceed was
left open on that thread and has had no reply there since 2026-08-31.

None of this is meant to push @fabiancz's work aside - the shape of the settings UI, the
url validation and the idea of unit testing it all come from there, and the credit for
them is his. What this adds on top is the preference migration that keeps older versions
working, the id handling, and the pieces shared with the sibling pull requests below. If
you would rather have #18494 with these changes applied to it instead, that is fine by me
and I am happy to help get it there.

rejected on its own:

- this one, user-defined tile providers (#18586)
- user-defined tile overlays (#18587)
- a grayscale option for the map (#18588)

This one and the overlay pull request share `cgeo.geocaching.unifiedmap.tiles` - url
handling, the shareable text form and the tile check - and a block of shared strings. Those
files are byte-identical in both branches, so whichever is merged first, the other merges
them without a conflict. Ordering of the two settings categories is the one thing that
needs a hand: whichever lands second should place its category after the one already there.

**Testing**: built and run on a Pixel (Android 17, API 37) with several providers
configured. The upgrade from a single provider was verified by reading the id derivation
rather than on the device: the old id was `<class>:null`, because the map Uri is rewritten
to `scheme://host` and has no last path segment, and the migrated entry reproduces exactly
that string. Checkstyle, lint and the unit test suite pass locally.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -41,6 +41,7 @@ import cgeo.geocaching.ui.AvatarUtils;
 import cgeo.geocaching.ui.notifications.Notifications;
 import cgeo.geocaching.unifiedmap.UnifiedMapType;
 import cgeo.geocaching.unifiedmap.tileproviders.AbstractTileProvider;
+import cgeo.geocaching.unifiedmap.tileproviders.PrefUserDefinedTileProvider;
 import cgeo.geocaching.unifiedmap.tileproviders.TileProviderFactory;
 import cgeo.geocaching.utils.FileUtils;
 import cgeo.geocaching.utils.LocalizationUtils;
@@ -1342,9 +1343,60 @@ public class Settings {
         return sharedPrefs.getStringSet(getKey(R.string.pref_tileprovider_hidden), empty);
     }
 
-    @Nullable
-    public static String getUserDefinedTileProviderUri() {
-        return getString(R.string.pref_userDefinedTileProviderUri, null);
+    /**
+     * User-defined tile providers, stored as a JSON list.
+     * Understands the legacy format as well, where the very same preference held a single, bare tile provider Uri.
+     */
+    /**
+     * The user-defined tile providers.
+     *
+     * <p>Falls back to the single provider of earlier versions while the new preference has not
+     * been written yet. Nothing is migrated eagerly and the old preference is never touched, so a
+     * version without this feature keeps working after a downgrade.</p>
+     */
+    @NonNull
+    public static List<PrefUserDefinedTileProvider> getUserDefinedTileProviders() {
+        final String value = StringUtils.trimToEmpty(getString(R.string.pref_userDefinedTileProviders, null));
+        // an empty *string*, not an empty list: a list that has been written is at least "[]", so
+        // this is only true while the preference has never been written. Emptying the list keeps
+        // it empty and does not bring the legacy provider back.
+        if (StringUtils.isEmpty(value)) {
+            final String legacyUri = StringUtils.trimToNull(getString(R.string.old_pref_userDefinedTileProviderUri, null));
+            return legacyUri == null ? new ArrayList<>() : new ArrayList<>(Collections.singletonList(
+                    // keeps LEGACY_KEY, so the derived tile provider id stays what it was and neither the
+                    // selected map source nor the list of hidden map sources needs rewriting
+                    new PrefUserDefinedTileProvider(PrefUserDefinedTileProvider.LEGACY_KEY, null, legacyUri)));
+        }
+        try {
+            return MAPPER.readValue(value, new TypeReference<List<PrefUserDefinedTileProvider>>() {
+            });
+        } catch (JsonProcessingException e) {
+            Log.e("Failure parsing user-defined tile providers: " + e.getMessage());
+            return new ArrayList<>();
+        }
+    }
+
+    /**
+     * Adds, updates or (if the given provider has no Uri) removes a user-defined tile provider.
+     */
+    public static void putUserDefinedTileProvider(final PrefUserDefinedTileProvider provider) {
+        final List<PrefUserDefinedTileProvider> providers = getUserDefinedTileProviders();
+        final int index = providers.indexOf(provider);
+        if (StringUtils.isBlank(provider.getUri())) {
+            if (index == -1) {
+                return;
+            }
+            providers.remove(index);
+        } else if (index == -1) {
+            providers.add(provider);
+        } else {
+            providers.set(index, provider);
+        }
+        try {
+            putString(R.string.pref_userDefinedTileProviders, MAPPER.writeValueAsString(providers));
+        } catch (JsonProcessingException e) {
+            Log.e("Failure writing user-defined tile providers: " + e.getMessage());
+        }
     }
 
     /** json array holding the configured tile overlays, see cgeo.geocaching.unifiedmap.overlays.TileOverlays */
--- a/main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
+++ b/main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
@@ -1,0 +1,197 @@
+package cgeo.geocaching.settings;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.Keyboard;
+import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.ui.dialog.Dialogs;
+import cgeo.geocaching.ui.dialog.SimpleDialog;
+import cgeo.geocaching.unifiedmap.tileproviders.PrefUserDefinedTileProvider;
+import cgeo.geocaching.unifiedmap.tiles.TileUrlTester;
+import cgeo.geocaching.unifiedmap.tiles.TileUrlUtils;
+import cgeo.geocaching.utils.AndroidRxUtils;
+import cgeo.geocaching.utils.ClipboardUtils;
+import cgeo.geocaching.utils.LocalizationUtils;
+
+import android.content.ClipData;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.widget.Button;
+import android.widget.EditText;
+
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.AlertDialog;
+import androidx.core.view.ViewCompat;
+import androidx.preference.Preference;
+import androidx.preference.PreferenceViewHolder;
+
+import com.google.android.material.button.MaterialButton;
+import com.google.android.material.textfield.TextInputLayout;
+import org.apache.commons.lang3.StringUtils;
+
+/** Preference to add / edit / remove a single user-defined tile provider */
+public class UserDefinedTileProviderPreference extends Preference {
+
+    public UserDefinedTileProviderPreference(final Context context) {
+        super(context);
+        setWidgetLayoutResource(R.layout.preference_copy_delete_buttons);
+    }
+
+    public UserDefinedTileProviderPreference(final Context context, final AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public UserDefinedTileProviderPreference(final Context context, final AttributeSet attrs, final int defStyle) {
+        super(context, attrs, defStyle);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull final PreferenceViewHolder holder) {
+        super.onBindViewHolder(holder);
+        setOnPreferenceClickListener(preference -> {
+            launchEditDialog();
+            return false;
+        });
+        final MaterialButton copyButton = (MaterialButton) holder.findViewById(R.id.copyview);
+        copyButton.setIconResource(R.drawable.ic_menu_copy);
+        copyButton.setOnClickListener(v -> {
+            final PrefUserDefinedTileProvider provider = getProvider();
+            if (provider != null) {
+                ClipboardUtils.copyToClipboard(provider.toShareableText());
+                ViewUtils.showShortToast(getContext(), R.string.settings_tileUrl_copied);
+            }
+        });
+
+        final MaterialButton button = (MaterialButton) holder.findViewById(R.id.deleteview);
+        button.setIconResource(R.drawable.ic_menu_delete);
+        button.setOnClickListener(v -> SimpleDialog.ofContext(getContext())
+                .setTitle(R.string.settings_userDefinedTileProvider)
+                .setMessage(R.string.settings_userDefinedTileProvider_remove_confirm)
+                .confirm(() -> {
+                    // a provider without Uri gets removed
+                    Settings.putUserDefinedTileProvider(new PrefUserDefinedTileProvider(getKey(), null, null));
+                    callChangeListener(null);
+                }));
+    }
+
+    public void launchEditDialog() {
+        final View v = LayoutInflater.from(getContext()).inflate(R.layout.userdefined_tileprovider_preference_dialog, null);
+        final TextInputLayout uriLayout = v.findViewById(R.id.editLayout);
+        final EditText editTitle = v.findViewById(R.id.title);
+        final EditText editUri = v.findViewById(R.id.edit);
+
+        final PrefUserDefinedTileProvider provider = getProvider();
+        final boolean isNew = provider == null;
+        if (!isNew) {
+            editTitle.setText(provider.getName());
+            editUri.setText(provider.getUri());
+        }
+        Dialogs.moveCursorToEnd(editUri);
+        Keyboard.show(getContext(), isNew ? editTitle : editUri);
+
+        final AlertDialog dialog = Dialogs.newBuilder(getContext())
+                .setView(v)
+                .setTitle(isNew ? R.string.settings_userDefinedTileProvider_addnew : R.string.settings_userDefinedTileProvider_edit)
+                // listeners are attached after show(), so that neither an invalid Uri nor a test
+                // run dismisses the dialog and discards what was typed
+                .setPositiveButton(android.R.string.ok, null)
+                .setNeutralButton(R.string.settings_tileUrl_test, null)
+                .setNegativeButton(android.R.string.cancel, (d, which) -> d.dismiss())
+                .show();
+
+        dialog.getButton(DialogInterface.BUTTON_POSITIVE).setOnClickListener(v2 -> save(dialog, uriLayout, editTitle, editUri));
+        dialog.getButton(DialogInterface.BUTTON_NEUTRAL).setOnClickListener(v2 -> test(dialog, uriLayout, editUri));
+        acceptSharedProviderOnPaste(editTitle, uriLayout, editTitle, editUri);
+        acceptSharedProviderOnPaste(editUri, uriLayout, editTitle, editUri);
+    }
+
+    /**
+     * Lets a shared tile provider be pasted into either field, filling both.
+     *
+     * <p>Hooks the paste action rather than reading the clipboard by itself: from Android 13 on,
+     * reading it unprompted shows a system notice and is restricted to the focused app, while
+     * handling what the user pastes needs no permission and stays silent.</p>
+     */
+    private void acceptSharedProviderOnPaste(@NonNull final EditText field, @NonNull final TextInputLayout uriLayout,
+                                             @NonNull final EditText editTitle, @NonNull final EditText editUri) {
+        ViewCompat.setOnReceiveContentListener(field, new String[]{"text/*"}, (view, payload) -> {
+            final ClipData clip = payload.getClip();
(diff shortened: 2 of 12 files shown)


eddiemuc:
Ok to test

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -1342,9 +1397,57 @@ public static Set<String> getHideTileproviders() {
...
+                return;
+            }
+            providers.remove(index);
+        } else if (index == -1) {
+            providers.add(provider);
+        } else {
+            providers.set(index, provider);
+        }
+        try {
+            putString(R.string.pref_userDefinedTileProviders, MAPPER.writeValueAsString(providers));
```
eddiemuc, line 1396: Migration:
if I understand code correctly, then when user has a legacy tile provider, then adds a new one (now has two tile providers), then deletes BOTH of them --> then the legacy tile provider will "magically" reappear (because the legacy property is still in the settings and will be then used by "getUserDefiledTileProvider()").
Is that correct? I think you should delete the legacy setting here (when you first write again the new setting where it is contained in)

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -1342,9 +1397,57 @@ public static Set<String> getHideTileproviders() {
...
+     * The user-defined tile providers.
+     *
+     * <p>Falls back to the single provider of earlier versions while the new preference has not
+     * been written yet. Nothing is migrated eagerly and the old preference is never touched, so a
+     * version without this feature keeps working after a downgrade.</p>
+     */
+    @NonNull
+    public static List<PrefUserDefinedTileProvider> getUserDefinedTileProviders() {
+        final String value = StringUtils.trimToEmpty(getString(R.string.pref_userDefinedTileProviders, null));
+        if (StringUtils.isEmpty(value)) {
```
eddiemuc, line 1363: Why not always adding legacy property to the returned list (if present)?

main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/UserDefinedTileProviderHelper.java
```diff
@@ -0,0 +1,33 @@
...
+import cgeo.geocaching.settings.Settings;
+import cgeo.geocaching.unifiedmap.tiles.TileUrl;
+import cgeo.geocaching.utils.LocalizationUtils;
+
+import android.net.Uri;
+
+import org.apache.commons.lang3.StringUtils;
+
+/** Common helpers for the user-defined tile providers (Mapsforge and VTM variant) */
+public class UserDefinedTileProviderHelper {
```
eddiemuc: It feels like all the methods in this class could be member methods of class PrefUserDefinedTileProvider  instead. Then we won't need the extra class

main/src/main/java/cgeo/geocaching/settings/Settings.java
```diff
@@ -267,6 +267,61 @@ public String toString() {
         }
     }
 
+    public static class PrefUserDefinedTileProvider {
```
eddiemuc: Class is used outside of settings Settings, it should not be a subclass of Settings but its own class (maybe place in new "tileprovider" package

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlShare.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * The text form of a single configured tile source, for passing one to somebody else through a
+ * chat message.
+ *
+ * <p>Deliberately plain text rather than json: several chat clients replace straight quotes with
+ * typographic ones, which would silently break json at the receiving end, and a plain line stays
+ * readable to whoever has to decide whether to trust the link.</p>
+ */
+public final class TileUrlShare {
```
eddiemuc, line 18: This class seems about sharing "PrefUserDefinedTileProvider" instances (not TileUrls). I believe its methods should be integrated into "PrefUserDefinedTileProvider" class (something like "toShareableText" and "fromShareableText", same coding style as with json config)

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlTester.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Fetches a single tile, from around the last map position so that the test hits an area the
+     * user actually looks at rather than a low zoom level many overlays do not serve at all.
+     *
+     * <p>Blocking - must not be called on the UI thread.</p>
+     *
+     * @return {@code null} if a tile was served, otherwise a short description of what went wrong
+     */
+    @Nullable
+    public static String test(@NonNull final String urlTemplate) {
```
eddiemuc, line 51: Why not use a "TileUrl" instance as parameter?

main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
```diff
@@ -0,0 +1,181 @@
...
+        final Settings.PrefUserDefinedTileProvider provider = getProvider();
+        final boolean isNew = provider == null;
+        if (!isNew) {
+            editTitle.setText(provider.getName());
+            editUri.setText(provider.getUri());
+        }
+        Dialogs.moveCursorToEnd(editUri);
+        Keyboard.show(getContext(), isNew ? editTitle : editUri);
+
+        final AlertDialog dialog = Dialogs.newBuilder(getContext())
```
eddiemuc, line 94: You might consider using SimpleDialog instead of constructing own AlertDialog. But that's not mandatory.

main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+        uriLayout.setError(null);
+
+        // a syntactically valid Uri can still be wrong, which shows up only as a map that stays
+        // empty - so fetch one real tile before storing it
+        final Button okButton = dialog.getButton(DialogInterface.BUTTON_POSITIVE);
+        okButton.setEnabled(false);
+        uriLayout.setHelperText(LocalizationUtils.getString(R.string.settings_tileUrl_testing));
+        AndroidRxUtils.andThenOnUi(AndroidRxUtils.networkScheduler,
+                () -> TileUrlTester.test(TileUrl.withDefaultTilePath(uri)),
```
eddiemuc: Url testing should not be mandatory for saving config IMHO because it requires internet access. 
Instead, there should be an optional "test" button in edit dialog for a configuration.

eddiemuc (commented):
lgtm overall. Few comments. Nothing major

eddiemuc:
@magma1447 Many thanks for your contribution! Few comments on code, nothing major. Necessary checks are green (default + Codacy).

main/src/main/java/cgeo/geocaching/settings/Settings.java
magma1447: 7ab2b4130 - Move PrefUserDefinedTileProvider out of Settings

Done. Put it in the existing `unifiedmap.tileproviders` package rather than a new one,
since the providers built from it and the helper operating on it already live there.

main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/UserDefinedTileProviderHelper.java
magma1447: 43a66bd61 - Fold UserDefinedTileProviderHelper into the provider class

Done, the helper class is gone. `getName` became `getDisplayName`, since the plain name is
already the stored field it falls back from.

One thing worth flagging: both had to be marked `@JsonIgnore`. Jackson would otherwise have
written `displayName` and `configured` into the stored json, and a later read would have
failed on the unknown properties. Added a test pinning the stored form to the three real
fields, as nothing else would have caught that before somebody's settings stopped loading.

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlShare.java
magma1447: 6572fe259 - Give the provider class its own shareable text form

Added `toShareableText()` and `fromShareableText()` on the class, so sharing reads like the
json config does and the paste handler now works with a provider instead of a two-element
`String[]`. That part was an improvement, thanks.

I did keep `TileUrlShare` itself, though, as the text format rather than as something
callers touch. The reason is not visible from this PR: the tile overlays in #18587 share a
name and a url in exactly the same way, from a class that has no `PrefUserDefinedTileProvider`
anywhere near it, and `unifiedmap/tiles` is byte-identical in both branches so that whichever
you merge first the other needs no changes there. Putting the format inside the provider
class would mean a second copy of it on the overlay side.

Say the word if you would rather have this PR judged on its own and the format inlined - it
is a small class and I can do that instead, it just costs the sharing between the two.

Also dropped `isShareable()` while in there, nothing but its own test called it.

main/src/main/java/cgeo/geocaching/settings/Settings.java
magma1447: aa072abf5 - Say why an empty string means the list was never written

I do not think it does, and I think the line reads as if it would - which is worth fixing
either way, so I have added a comment there.

Walking the scenario through: with a legacy provider and nothing else, the new preference
has never been written, so the fallback supplies it as the single entry. Adding a second one
writes both to the new preference. Deleting both writes `[]` to it. On the next read the
stored value is `"[]"`, which is not an empty *string*, so the fallback is not taken and the
list stays empty.

So it hinges on `StringUtils.isEmpty` being asked about the stored text rather than about the
list - a list that has ever been written is at least `[]`. That is easy to read as "the list
is empty", which I suspect is what this comment is really about, hence the comment in the
code now.

`putUserDefinedTileProvider` is the only writer and always writes json, so there is no path
that stores an empty string.

main/src/main/java/cgeo/geocaching/settings/Settings.java
magma1447: (no commit - proposing to keep the current behaviour)

Two reasons, and the first is not mine: @moving-bits asked for exactly this on the
predecessor pull request, https://github.com/cgeo/cgeo/pull/18494#issuecomment-5467468858 -
new preference, data migrated from the old one, old one kept readable because people move
between nightly and regular builds. Leaving `userDefinedTileProviderUri` alone is what makes
that work: a build without this feature still finds the provider it knows about.

The second is that the two preferences do not hold the same thing. The new one stores a name
and a uri per provider; the old one is a bare uri with nowhere to put a name. So an entry that
kept living in the old key would have its name in the list and its uri in the old string - one
entry split across two keys.

The coherent version of that is to write the first entry's uri back to the old key on every
save, and honestly that is not much code, and whoever eventually drops legacy support would
find it sitting right there. It would not be my own first choice - it means a second writer
for data the new preference already holds, to keep a key that is meant to fade away - but if
you would rather have it that way, say so and I will add it.

What it does instead is seed the list from the old value once. The first save writes that entry
into the new preference with everything else, so from then on it is an ordinary entry that can
be renamed or deleted like any other, and the old key is simply left as it was.

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlTester.java
magma1447: (no commit yet - checking I have understood you before starting)

There is no instance to pass at the moment: `TileUrl` is a utility class, private constructor
and static methods only. Which I suspect is the actual point of your comment - the name
promises a type and the class is not one.

If that is what you mean, I agree, and I think it has become worth doing rather than just
nicer. Until now there was one user-defined url in the whole app; this pull request turns that
into a list, and the overlays in #18587 add another kind of the same thing. `test()` takes a
bare `String` today and re-validates it with `TileUrl.isValid`, purely because a caller could
hand it something that never went through `normalize`. A `TileUrl.of(...)` that returns null
for anything invalid would remove that check and the "invalid url" result along with it.

Before starting I would rather be sure, because it is the biggest of the changes in this round:
seven call sites here, about ten in #18587, and the 19 tests in `TileUrlTest` rewritten. Since
`unifiedmap/tiles` is byte-identical in both branches I would push the same change there at the
same time to keep it that way. It is all in files these pull requests add, so nothing existing
gets disturbed.

Happy to do it - just confirm that is what you meant and I will get started.

main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
magma1447: 3c4d1d449 - Test a tile url on request instead of before every save

Agreed, done. Saving no longer touches the network at all; the check sits behind a "Test"
button in the edit dialog and reports either way without storing anything. The failure text
lost its "Save it anyway?", since there is nothing left to decide.

Worth knowing that #18587 does the same thing in its overlay dialog, and the `settings_tileUrl_*`
strings are shared between the two branches. I have pushed the matching change there so the two
dialogs behave the same.

Screenshot_20260911-102255

main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
magma1447: (no commit - taking you up on "not mandatory", but happy to be overruled)

Fair question, and the preference this replaces did use it: `TextPreference` calls
`SimpleDialog.input(...)`. That convenience is for a single free-text value though, and this
dialog has two fields, a custom layout with an explanatory footer, and now a test button that
must not close it.

`SimpleDialog` can still do it, it just comes out more indirect rather than less. `confirm()`
only creates a neutral button when passed a neutral `Runnable`, and any runnable it is passed is
run and then dismissed - so a test button that leaves the dialog open means setting the neutral
button separately, putting all three buttons' behaviour in `setButtonClickAction`, and calling
`confirm(null, null, null, null)`. Disabling the test button while the request runs then needs
`getDialog().getButton(...)`, the `AlertDialog` underneath, anyway. Against two plain click
listeners today, I did not think that was a trade worth making.

For what it is worth, this was modelled on the log templates editor under logging settings -
the rows with an edit dialog behind each one, and a trailing button to add another. That is
also where the dialog construction comes from, so the two match rather than this one having
gone its own way.

If you would rather have `SimpleDialog` regardless, say so and I will convert this and the
overlay dialog in #18587 together so those two stay alike.

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlShare.java
eddiemuc: fine for me

main/src/main/java/cgeo/geocaching/unifiedmap/tiles/TileUrlTester.java
eddiemuc: Indeed I overlooked that TileUrl is a Utils class :-) Sorry, that was my fault.

Maybe simply rename "TileUrl" to "TileUrlUtils"?

main/src/main/java/cgeo/geocaching/settings/UserDefinedTileProviderPreference.java
eddiemuc: not mandatory, so leaving as is is fine with me

eddiemuc:
@magma1447 all fine for me. Regarding TileUrl you may consider renaming to TileUrlUtils (or even merge into UrlUtils).

Your changes produces some minor Codacy-findings:
main/src/main/java/cgeo/geocaching/unifiedmap/tileproviders/PrefUserDefinedTileProvider.java#L15
Extra separation in import group before 'com.fasterxml.jackson.annotation.JsonCreator'

eddiemuc:
Ah and could you please squash commits into 1?

magma1447:
All three done and pushed.

`TileUrl` is now `TileUrlUtils`. I left it as its own class rather than merging into `UrlUtils`, since there is no `UrlUtils` in the codebase to merge into - happy to be pointed at the one you had in mind if I have missed it.

The Codacy finding was mine: I had put a blank line between the commons and jackson imports, which are the same group. Fixed, and I checked the rest of the files this adds for the same thing.

Squashed to a single commit. #18587 too, since the same applies there.


A friendly reminder while both are open: #18586 and #18587 will not both merge cleanly. They add a settings category to the same screen and strings to the same file, so whichever goes in second will conflict in four files - none of them the shared tile classes. Say the word once the first one is merged and I will rebase the other onto it, so you do not have to resolve anything yourself.

eddiemuc:
Both have still xodacy issues I'm afraid (sorry, I know from own experience that this is unnerving...)

magma1447:
> Both have still xodacy issues I'm afraid (sorry, I know from own experience that this is unnerving...)

It's fixed - and you have nothing to be sorry about. Sloppy on my part, as simple as that. The tooling is there, it just has to be used properly.

I am just amazed over your short response time to be honest. And my brain feels a bit fried today. I have been working on three different things in parallel more or less. But I think all three of them (including this) is done now.

eddiemuc:
> I am just amazed over your short response time to be honest. 

I am doing this via mobile phone standing in line in the supermarket (hence the typos :-)

First PR is merged. As you predicted this ine has conflicts now.

magma1447:
> I am doing this via mobile phone standing in line in the supermarket (hence the typos :-)
> 
> First PR is merged. As you predicted this ine has conflicts now.

Do you want me to fix (rebase) this branch, and push again? I am very happy to do it.

(I am on it now, then I'll afk for an hour)

eddiemuc:
> > I am doing this via mobile phone standing in line in the supermarket (hence the typos :-)
> > 
> > First PR is merged. As you predicted this ine has conflicts now.
> 
> Do you want me to fix (rebase) this branch, and push again? I am very happy to do it.
> 
> (I am on it now, then I'll afk for an hour)

Yes please 

(I could do these things too eg squashing or conflict resolution but then your commit would be wrongly labeled as mine)

Issue #3954, opened 2014-06-06
Add support for (custom) WMS (map)data sources


Megachip:
So user can add his own (online) maps.

MrYaz:
What is the status of this issue?
Is c:geo still using Google Maps API v1?

I see a great use for this feature in c:geo since google maps are lacking a lot of details in Sweden.
We have a lot of better maps that I would like to be able to use in c:geo.
Eniro.se, Hitta.se, Ekonomiska kartan to name a few.

I actually believe that the lack of custom maps in c:geo might soon result in a lot of users migrating to Locus, Cachebird, Cachesense to name a few...

Too bad if that happened...

rsudev:
We still use google maps api v1, as the shift to v2 is an undertaking nobody had the resources for yet. Unfortunately there is no way to do this stepwise incremental (at least I see none), which would make it more manageable.

matej116:
I'm now working on Google maps api v2, so far I got working map, showing user location and caches, but it's a huge change of the map source architecture and abstraction. I am going to commit it and create pull request as soon as it will have implemented all the functionality which v1 has. But it would still probably require rethinking of the architechture.

LeXeRLV:
Other offline and online maps supported by cgeo will be the best update for
it.
Without them usually need to use Locus..

Lineflyer:
We are meanwhile using GMaps API v2 and it sounds like the prerequisite for such feature is now there.
Therefore I keep this open.

sdennler:
This would be very nice. I would love to use Swissimage and LIDAR images: https://josm.openstreetmap.de/wiki/Maps/Switzerland

Issue #1286, opened 2012-03-15
Support of eniro maps


Lineflyer:
Feature request from FB (don't know if it is technically possible):

Is it possible to get Eniro streetmap in C:geo, in Neongeo is with this URL: http//map.eniro.com/geowebcache/service/tms1.0.0/map/[z]/[x]/[iy].png

The small islands, creeks och "beachline" is missing or very deformed, looks fine in Eniro streetmap, even wetlands are missing in google.

Live in sweden

ExTechOp:
As far as I know, the idea of removing tile-based rendering support from mapsforge has not progressed since 2012, so there is really no reason for **not** to merge this into the main c:geo codebase. Could this still happen?

SammysHP:
Merge what? That test branch is four years old (sometimes it is hard to merge something after just one week) and does not exist anymore.

ExTechOp:
Yeah, obviously that instance is long gone, but IMHO the idea of having configurable tile servers should be resurreceted. For example, the National Land Survey of Finland has excellent freely available maps, and there is an unofficial tile server. Are you still there, **rsudev**?

SammysHP:
Sure, he is one of our main developers and is currently working on a mapsforge 0.6 implementation.

Souldrinker:
+1 on this request. When I used Neongeo back in the day I often used the Eniro aerial view which seemed more detailed than google satelites and for example today I just solved a myst that is located in an area where Eniro have paths marked on their maps, but Google/OSM doesn't, so this would be appreciated.

Since NeonGeo stopped being developed like three years ago and it doesn't support hi-res screens in a good way I stopped using that in favor of c:geo. I saw on this thread: http://www.geocaching.se/forum/viewtopic.php?f=66&t=18952, that also CacheSense supports custom map url's, but I haven't used that either for a while since it didn't seem to read in all caches consistently and neither those or the official apps gets close to c:geo in other regards, so having support for custom tile servers in c:geo would be great.

PR #18587, opened 2026-09-10
Add user-defined tile overlays

magma1447:
## Description

Adds support for user-defined tile overlays: raster tile layers drawn *on top* of the
current map source, rather than replacing it.

Until now c:geo can only swap the base map. A user-defined tile provider takes over the
whole map, so raster data meant to complement a map — sea marks, hiking and cycling
routes, piste maps, national ortho layers — cannot be shown together with it. The only
transparent raster layer so far is hillshading, which is fed from local DEM files and is
not configurable.

What this adds:

- A new **User-defined overlays** group under *Settings → Map Sources*, listing the
  configured overlays one per row with an inline delete icon and an "add" button below,
  following the layout of the log templates.
- Each overlay is switched on and off individually in the map's layer selection, directly
  below the map sources. Several overlays can be active at once, and they are independent
  of the map source in use.
- Tiles are expected to carry their own transparency, so no alpha blending is applied.
- The name is optional; when left empty the host of the url is shown instead. It is
  derived on read rather than stored, so correcting a typo in the url also corrects a name
  that was never explicitly set. Names need not be unique.

Details worth calling out:

- Url templates accept `{x}/{y}/{z}` in either case and are normalised to the upper case
  spelling used elsewhere in the tile provider classes. Query parameters are preserved,
  which api keys usually depend on. Only `http`/`https` are accepted. As with the existing
  user-defined tile provider, a url without placeholders is accepted and gets the default
  tile path appended.
- Configuration and on/off state live in two separate new preferences, so toggling an
  overlay on the map never rewrites the configuration. Both are new keys, nothing is
  migrated, and older versions simply ignore them.
- Every overlay gets a key of its own that stays stable across renames, so the enabled
  state cannot be confused between entries.
- Implemented for both the Mapsforge and the VTM renderer. A new z-index between the
  buildings and the labels slot places overlays above the map and its hillshading, but
  below street labels and all cache and route items.
- Saving validates the Uri by actually fetching one tile from it, so a typo in the host or
  path is caught at once instead of showing up later as an overlay that silently draws
  nothing. The check is advisory: if it fails, the reason is shown and the overlay can
  still be saved anyway, since a server may cover only some regions or be temporarily
  unavailable. What it can and cannot catch is described further down.
- **Sharing an overlay with somebody else.** Copying one puts a single line of plain text
  on the clipboard - the name and the Uri - which can be sent through any chat client or
  mail. At the other end, pasting it into either field of the add dialog is recognised and
  fills in both. A bare Uri without a name works as well, so a link found on a website can
  be pasted straight in. The text stays readable, so the recipient can see what they are
  about to add before adding it.
- Unit tests cover the url handling, the shareable form and the stored json.

### Screenshots

Screenshot_20260910-140909

Screenshot_20260910-141012

The same view with nothing enabled, with one overlay, and with two - showing that several
can be active at once and that they compose over whatever map source is underneath:

<table>
<tr>
<td>
Screenshot_20260910-141044
</td>
<td>
Screenshot_20260910-141049
</td>
<td>
Screenshot_20260910-141057
</td>
</tr>
<tr>
<td align="center"><em>no overlay</em></td>
<td align="center"><em>sea marks</em></td>
<td align="center"><em>sea marks + turf zones</em></td>
</tr>
</table>

Both of them in the layer selection, below the map sources, each with its own checkbox:

Screenshot_20260910-141315

The sea marks come from openseamap.org and the zones from a Turf tile server of my own,
which I plan to make generally available in a week or two. Neither is shipped with c:geo or
suggested anywhere in the app - they are just what I had configured while testing.

## Related issues

No open issue asks for this directly, as far as I can find. The nearest requests - #3954
(*custom WMS map data sources*) and #1286 (*eniro maps*) - are about replacing the base map
rather than drawing on top of it, and are answered by the sibling pull request for
user-defined tile providers instead.

- #18494 — *Add support for multiple custom online maps*. Related but separate: that one is
  about base maps, this one about overlays. See the note on the sibling pull requests below.

## Additional context

**One deliberate divergence from the log templates layout it otherwise follows:** the edit
dialog keeps a Cancel button. The log templates dialog has none and relies on tapping
outside to dismiss, which we thought worth diverging from rather than copying.

Overlays are their own preference category rather than a row inside *Map Data*. We tried
a de-emphasised sub-heading first, on the grounds that overlays are one aspect of map data
rather than an unrelated topic, but the section divider above it still read as "new
section" and contradicted the muted title. A normal category makes both cues agree and
matches the rest of the app.

**On how many entries the layer selection can take.** The overlays are listed inline there,
one checkbox each, which is fine for a handful. If you expect people to configure more than
that and would rather keep the menu short, they could just as well go behind a submenu, the
way *Manage offline data* does - either always, or only once there is more than one. Happy
to change it; it is a small change and I have no strong preference.

rejected on its own:

- this one, user-defined tile overlays (#18587)
- user-defined tile providers, superseding #18494 (#18586)
- a grayscale option for the map (#18588)

The two tile pull requests share `cgeo.geocaching.unifiedmap.tiles` - url handling, the
shareable text form and the tile check - and a block of shared strings. Those files are
byte-identical in both branches, so whichever is merged first, the other merges them
without a conflict. Ordering of the two settings categories is the one thing that needs a
hand: whichever lands second should place its category after the one already there.

~~**Google Maps renderer is not implemented.** Overlays work on the Mapsforge and VTM
fragments; selecting a Google map source swaps in `GoogleMapsFragment`, which would need
its own `addTileOverlay` call. It is a small addition, but `maps_api2_key` is empty in the
repository so Google sources never register in a locally built APK, and I have no way to
test it. Happy to add it blind if you would rather have it in this PR than as a follow-up.~~

**Attribution and usage conditions** are the user's responsibility, as with the existing
user-defined tile provider. No overlay is shipped as a default or suggested in the UI.

**What the url check can and cannot catch.** It fetches one tile at zoom 12 around the last
map position, or around Geocaching HQ when no map position has been stored yet.

It catches an unreachable or misspelled host, TLS failures, HTTP error codes, and error
pages served as html. It does not catch a wrong path on a server that answers unknown
paths with a default tile - openseamap.org for instance returns the same small transparent
png for `/seamark/12/2253/1204.pn`, `/seamarkXX/...` and `/nonsense/...` as it does for a
genuinely empty area, so no check could tell those apart. Conversely a server that covers
only one region, or that 404s where it has no data, can fail the check for a perfectly
good url - hence saving anyway.

**Testing**: built and run on a Pixel (Android 17, API 37), against OpenSeaMap seamarks and
Waymarked Trails hiking tiles. Verified on both the Mapsforge and the VTM renderer:
overlays render with correct transparency over the base map, toggle on and off, survive
map source changes, and persist across restarts. LeakCanary run over toggle and
map-source-change cycles on both renderers. Checkstyle, lint and the unit test suite pass
locally.

**AI assistance**: this contribution was written with AI assistance (Claude Opus 5). I
have reviewed and tested everything submitted here.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -1347,6 +1347,32 @@ public class Settings {
         return getString(R.string.pref_userDefinedTileProviderUri, null);
     }
 
+    /** json array holding the configured tile overlays, see cgeo.geocaching.unifiedmap.overlays.TileOverlays */
+    @Nullable
+    public static String getTileOverlaysConfig() {
+        return getString(R.string.pref_tileOverlays, null);
+    }
+
+    public static void setTileOverlaysConfig(@Nullable final String config) {
+        putString(R.string.pref_tileOverlays, config);
+    }
+
+    /** keys of the tile overlays currently switched on in the map view */
+    public static Set<String> getTileOverlaysEnabled() {
+        final Set<String> empty = Collections.emptySet();
+        if (sharedPrefs == null) {
+            return empty;
+        }
+        return sharedPrefs.getStringSet(getKey(R.string.pref_tileOverlaysEnabled), empty);
+    }
+
+    public static void setTileOverlaysEnabled(final Set<String> keys) {
+        if (sharedPrefs == null) {
+            return;
+        }
+        sharedPrefs.edit().putStringSet(getKey(R.string.pref_tileOverlaysEnabled), keys).apply();
+    }
+
     public static void setMapLanguage(@Nullable final String language) {
         putString(R.string.pref_mapLanguage, StringUtils.isBlank(language) ? "" : language);
     }
--- a/main/src/main/java/cgeo/geocaching/settings/TileOverlayPreference.java
+++ b/main/src/main/java/cgeo/geocaching/settings/TileOverlayPreference.java
@@ -1,0 +1,192 @@
+package cgeo.geocaching.settings;
+
+import cgeo.geocaching.R;
+import cgeo.geocaching.activity.Keyboard;
+import cgeo.geocaching.databinding.TileOverlayEditDialogBinding;
+import cgeo.geocaching.ui.ViewUtils;
+import cgeo.geocaching.ui.dialog.Dialogs;
+import cgeo.geocaching.ui.dialog.SimpleDialog;
+import cgeo.geocaching.unifiedmap.overlays.TileOverlay;
+import cgeo.geocaching.unifiedmap.overlays.TileOverlays;
+import cgeo.geocaching.unifiedmap.tiles.TileUrlShare;
+import cgeo.geocaching.unifiedmap.tiles.TileUrlTester;
+import cgeo.geocaching.unifiedmap.tiles.TileUrlUtils;
+import cgeo.geocaching.utils.AndroidRxUtils;
+import cgeo.geocaching.utils.ClipboardUtils;
+import cgeo.geocaching.utils.LocalizationUtils;
+
+import android.content.ClipData;
+import android.content.Context;
+import android.content.DialogInterface;
+import android.view.LayoutInflater;
+import android.widget.Button;
+import android.widget.EditText;
+
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.AlertDialog;
+import androidx.core.view.ViewCompat;
+import androidx.preference.Preference;
+import androidx.preference.PreferenceViewHolder;
+
+import com.google.android.material.button.MaterialButton;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * One configured tile overlay, shown as a row in the settings.
+ *
+ * <p>Follows the layout of the log templates: tapping the row opens the edit dialog, the trailing
+ * icon deletes the entry. Whether an overlay is actually drawn is decided in the map's layer
+ * selection, not here.</p>
+ */
+public class TileOverlayPreference extends Preference {
+
+    private final TileOverlay overlay;
+    private final Runnable onChanged;
+
+    public TileOverlayPreference(final Context context, @NonNull final TileOverlay overlay, @NonNull final Runnable onChanged) {
+        super(context);
+        this.overlay = overlay;
+        this.onChanged = onChanged;
+
+        setKey(overlay.getKey());
+        setTitle(overlay.getDisplayName());
+        setSummary(overlay.getUrl());
+        setIconSpaceReserved(false);
+        setPersistent(false);
+        setWidgetLayoutResource(R.layout.preference_copy_delete_buttons);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull final PreferenceViewHolder holder) {
+        super.onBindViewHolder(holder);
+        setOnPreferenceClickListener(preference -> {
+            showEditDialog();
+            return true;
+        });
+        final MaterialButton copyButton = (MaterialButton) holder.findViewById(R.id.copyview);
+        copyButton.setIconResource(R.drawable.ic_menu_copy);
+        copyButton.setOnClickListener(v -> {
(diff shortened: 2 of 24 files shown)


perwestling:
This is a nice feature! Have been doing Geocaching and Turf since 2010, and have not found any good way to integrate them.

Did talk to Turf team over 10 years about possibility to export GPS coordinates from zones.
And have also tried to export turfzones and import them into a cachetur.no trip.
But have not gotten any of these to work.

For me, when travelling, I primarily focus on Geocaching, using c:geo. If I remember I might start the app and check for turf zones, but I often forget. Having them in c:geo map would increase the possibility of me remembering.

This should also make it possible to save battery, as I can some times use just one app (c:geo) for both games.

Have no views on the PR as such, as this is the first time I am into this repository. But it is good to know it exists, so if I need some feature I can look into the repository and maybe create a PR for it? Who knows?

magma1447:
Since @eddiemuc hinted me that I can get my own Google map keys I have done so. I will look into making sure this works with the Google maps renderer as well. When its solved I will update the branch.

magma1447:
Add code to make this supported on Google maps as well.

eddiemuc:

Ok to test

eddiemuc (approved):
lgtm

eddiemuc:
Code lgtm, necessary checks are green (default + Codacy). 

@magma1447 Many thanks for your contribution! Ready to merge from your perspective?

magma1447:
> @magma1447 Many thanks for your contribution! Ready to merge from your perspective?

This one is also affected by some of the feedback in #18586. They share a lot of the new code. I'll try to work through all of the open threads during the day today, and those that are relevant for this PR will be corrected in this branch as well.

magma1447:
Yes, ready now.

This carries the parts of the #18586 feedback that apply here: the url test is behind a "Test" button in the overlay dialog as well, so saving no longer waits for the network, and `TileUrl` is renamed to `TileUrlUtils`. Squashed to one commit like the other one.

The `unifiedmap/tiles` classes and the `settings_tileUrl_*` strings are byte-identical between the two branches, so whichever you merge first, the other needs no changes there.


A friendly reminder while both are open: #18586 and #18587 will not both merge cleanly. They add a settings category to the same screen and strings to the same file, so whichever goes in second will conflict in four files - none of them the shared tile classes. Say the word once the first one is merged and I will rebase the other onto it, so you do not have to resolve anything yourself.

magma1447:
Another example of what this can do. County borders according to Project-GC. The tile layer isn't released yet, I was just testing it out.

Screenshot_20260911-205959

moving-bits:
> Another example of what this can do. County borders according to Project-GC. The tile layer isn't released yet, I was just testing it out.

FYI: This can be achieved by loading a GeoJSON as well. There are several websites out there providing GeoJSON files with political boundaries of different levels (eg.: country, county, ...).

magma1447:
> FYI: This can be achieved by loading a GeoJSON as well. There are several websites out there providing GeoJSON files with political boundaries of different levels (eg.: country, county, ...).

But not online linked right? I found issue #13793 referencing the concept. It's not an obvious way to do it, but if one knows it, I guess it works.

Personally I would rather see an option that auto-updates the data. Instead of, as a user, have to update it regularly.

This wasn't my primary goal when I sent this PR, if it was I would have written the code in Project-GC first, so that I knew that part worked (fast enough). But I do think that it's quite a user-friendly way to get the administrative borders for most of the world, the way that Project-GC and Challenges measures it.

PS! If there is a way to add an url for geojson data, which I have missed, feel free to hint me to where.

moving-bits:
Yeah, I think both approaches have their merits:

The "border layer" has the advantage of being up to date (if the data source is), the GeoJSON variant is offline-capable.

I did not want to imply that one was is better or the other, just wanted to point out that there are multiple ways to achieve this goal in c:geo. Which way works better for a user depends on that user's specific needs.

Anyway, thanks for your recent additions to the map - this opens up another range of possibilities.

(PS: There's no way currently to auto-update GeoJSON, KML or GPX track/route files for display on map, that's all offline-only.)

PR #18588, opened 2026-09-10
Add a grayscale option for the map

magma1447:
## Description

Adds an on/off entry to the map's layer selection that renders the map itself in grey,
while caches, waypoints and routes keep their colours.

Geocache icons are hard to pick out when zoomed out over a large area: the map's own
colours compete with them, and a list of caches spread over a region turns into clutter.
Draining the colour from the map alone makes the icons the only coloured thing on screen,
while the roads, coastline and forests are still there to tell you where you are looking.

How it works:

- Vector maps go through the render theme, so their colours are replaced by way of a
  `ThemeCallback`.
- Raster tiles arrive already rendered and have no theme, so their decoded bitmaps are
  converted instead, through a tile source that plugs its own decoder into the data source.

Both hooks belong to VTM, so the entry is only shown while a VTM map source is in use -
the same way hillshading and the background map entries are only shown where they apply.

### Screenshots

The option sits in the map's layer selection, below the map sources:

Screenshot_20260910-135223

The same area zoomed out, before and after - same caches, same map, colour drained from the
map only:

<table>
<tr>
<td>
Screenshot_20260910-135213
</td>
<td>
Screenshot_20260910-135227
</td>
</tr>
<tr>
<td align="center"><em>off</em></td>
<td align="center"><em>on</em></td>
</tr>
</table>

## Related issues

- Solves the same problem as #17436 (*Add "empty" map*), which describes wanting "an
  overview of a list with scattered caches" without "being distracted by a map". To me,
  personally, this is a much better solution to that problem: an empty map removes the
  clutter but also removes every clue about where you are, while a grey map keeps the
  geography and only takes away what was competing with the icons. The two do not conflict
  and can live side by side in the list.

## Additional context

This is not a new idea - GCDroid has had it for years and I have used it there for exactly
this, on zoomed-out maps of an area I am about to go caching in. It works well in practice,
which is why I wanted it in c:geo.

rejected on its own:

- this one, the grayscale option (#18588)
- user-defined tile overlays (#18587)
- user-defined tile providers, superseding #18494 (#18586)

This one shares no code with the other two; it only sits next to them in the same menu.

**Not implemented for the Mapsforge renderer.** Mapsforge has an equivalent `ThemeCallback`
for vector maps, so that half would be straightforward, but its raster tiles are decoded
inside the `AndroidGraphicFactory` singleton and there is no per-layer hook to convert them
without touching a factory the whole app shares. Rather than offer the option and have it
work for only some map sources, it is shown only where it works completely. Happy to look
at the Mapsforge side separately if you would like it there too.

**Testing**: built and run on a Pixel (Android 17, API 37), on VTM with both an online
raster source and an offline vector map, and with "UnifiedMap variants" set to VTM and to
Mapsforge + VTM. The map turns grey, cache icons keep their colours, and the toggle works
in both directions. Checkstyle, lint and the unit test suite pass locally.

### DIFF
--- a/main/src/main/java/cgeo/geocaching/settings/Settings.java
+++ b/main/src/main/java/cgeo/geocaching/settings/Settings.java
@@ -2639,6 +2639,14 @@ public class Settings {
         return !getBoolean(R.string.pref_mapScaleOnly, true);
     }
 
+    public static boolean getMapGrayscale() {
+        return getBoolean(R.string.pref_mapGrayscale, false);
+    }
+
+    public static void setMapGrayscale(final boolean value) {
+        putBoolean(R.string.pref_mapGrayscale, value);
+    }
+
     public static boolean getMapShadingShowLayer() {
         return getBoolean(R.string.pref_maphillshading_show_layer, true);
     }
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/UnifiedMapActivity.java
@@ -1092,6 +1092,10 @@ public class UnifiedMapActivity extends AbstractNavigationBarMapActivity impleme
                         mapFragment.getViewport().filter(vmCaches));
                 CacheListActivity.startActivityMap(this, new SearchResult(caches));
             }
+        } else if (id == R.id.menu_grayscale) {
+            Settings.setMapGrayscale(!Settings.getMapGrayscale());
+            item.setChecked(Settings.getMapGrayscale());
+            changeMapSource(mapFragment.currentTileProvider);
         } else if (id == R.id.menu_hillshading) {
             Settings.setMapShadingShowLayer(!Settings.getMapShadingShowLayer());
             item.setChecked(Settings.getMapShadingShowLayer());
--- a/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsThemeHelper.java
+++ b/main/src/main/java/cgeo/geocaching/unifiedmap/googlemaps/GoogleMapsThemeHelper.java
@@ -240,9 +240,13 @@ class GoogleMapsThemeHelper {
                 style = addStyler(style, option, true);
             }
         }
-        //5. apply map style constructed above to Google Map
+        //5. drain the colour from whatever the steps above produced, if the map is set to grayscale
+        if (Settings.getMapGrayscale()) {
+            style = addGrayscaleStyler(style);
+        }
+        //6. apply map style constructed above to Google Map
         map.setMapStyle(style == null ? null : new MapStyleOptions(JsonUtils.nodeToString(style)));
-        //6. adjust scaleDrawer as needed
+        //7. adjust scaleDrawer as needed
         if (scaleDrawer != null) {
             scaleDrawer.setNeedsInvertedColors(theme.needsInvertedColors);
         }
@@ -272,6 +276,32 @@ class GoogleMapsThemeHelper {
      *     ]
      *   }
      */
+    /**
+     * Appends a styler that takes all colour out of the map, whatever the theme before it asked for.
+     * Carries no featureType, so it applies to everything the style covers. Markers, routes and the
+     * tile overlays are not part of the map style and keep their colours.
+     * <p>
+     * Creates JSON like following example:
+     * {
+     *     "stylers": [
+     *       {
+     *         "saturation": -100
+     *       }
+     *     ]
+     *   }
+     */
+    private static ArrayNode addGrayscaleStyler(final ArrayNode inputNode) {
+        final ArrayNode node = inputNode == null ? JsonUtils.createArrayNode() : inputNode;
+        final ObjectNode saturationNode = JsonUtils.createObjectNode();
+        JsonUtils.setInt(saturationNode, "saturation", -100);
+        final ArrayNode stylers = JsonUtils.createArrayNode();
+        stylers.add(saturationNode);
+        final ObjectNode grayscaleNode = JsonUtils.createObjectNode();
+        JsonUtils.set(grayscaleNode, "stylers", stylers);
+        node.add(grayscaleNode);
+        return node;
+    }
+
     private static JsonNode createVisibilityOnStylerForFeatureType(@Nullable final String featureType, final boolean on) {
         final ObjectNode node = JsonUtils.createObjectNode();
         if (featureType != null) {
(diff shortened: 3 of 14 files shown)


SammysHP:
I like this idea a lot, but wonder if it is possible to achieve the same result with some kind of shader overlay that is applied to the whole map layer just before the marker layer is rendered. If it requires changes in VTM and Mapsforge, that should be fine. Bonus points if it can also be used together with Google Maps.

magma1447:
> I like this idea a lot, but wonder if it is possible to achieve the same result with some kind of shader overlay that is applied to the whole map layer just before the marker layer is rendered. If it requires changes in VTM and Mapsforge, that should be fine. Bonus points if it can also be used together with Google Maps.

Thanks - and I had the same thought, so I went looking at what each renderer would need.
The summary is that a "grey the map layer before the markers" pass is doable, but it turns
into three different mechanisms rather than one, and they are not equally expensive.

**Google Maps: easier than a shader, and it already has the plumbing.**
`GoogleMapsThemeHelper` loads `MapStyleOptions` from the JSON in `res/raw` and already
builds `stylers` entries programmatically. A `saturation: -100` styler across the feature
types greys the map and leaves markers alone, with no rendering changes at all. So the
bonus points look like the cheapest part of this to write - but not for me to verify:
`maps_api2_key` is empty in the repository, so Google map sources never register in a build
made here and I cannot see the result. I can write it, but somebody with a key would have
to confirm it actually looks right before it is trusted.

**VTM: possible, and I do not think it needs changes in VTM.** `Layer` exposes a
`LayerRenderer` with `render(GLViewport)`, so a layer inserted between the map layers and
the marker layers could copy the framebuffer into a texture and draw it back through a
desaturating shader. I have not written it, so I cannot promise it behaves - and it does
mean a full-screen copy every frame, which is a real cost on weaker devices for something
that is only a display preference.

**Mapsforge: the awkward one.** It draws through an Android `Canvas`, so the equivalent is
compositing with `BlendMode.SATURATION` over the map layers before the markers are drawn.
That is API 29+, and minSdk here is 26, so 26-28 would need a separate path through an
offscreen bitmap.

Mapsforge is also the renderer my current approach cannot cover for raster tiles, for an
unrelated reason: those tiles are decoded inside `AndroidGraphicFactory.INSTANCE`, which
the whole app shares, and there is no per-layer hook to convert them on the way through.
That is a property of the library rather than of the Android version, so no minSdk bump
would help it. Your idea would cover that case, which mine cannot.

So the trade, as far as I can see it: what is in this PR is small and cheap but VTM-only,
while the compositing approach covers Mapsforge raster and Google as well, at the cost of
GL code, a per-frame copy on VTM, and an API-level fallback on Mapsforge.

One small thing in favour of the narrow version: because the entry is only shown where it
works, anyone not on VTM never sees it at all - the same way hillshading and the background
map entries are hidden where they do not apply. Nobody is offered an option that does
nothing, and the layer menu does not grow for them. That is a consolation rather than an
argument, though; covering all three renderers would plainly be better.

I am happy to try the compositing version for VTM and Mapsforge - either replacing what is
here or as a follow-up once this is in, whichever you prefer. For Google I would rather
write it only if one of you can check it on a build with a key, or leave that part to
somebody who can.

If you would rather see the compositing version, it would help to know whether the
per-frame framebuffer copy on VTM is acceptable to you - that is the part I would not want
to discover you object to only after building it.

eddiemuc:
For Mapsforge **Offline** maps, grayscale can be achieved pretty simple using a ThemeCallback. Eg:

```
        mMapView.getModel().displayModel.setThemeCallback(new ThemeCallbackAdapter() {

            @Override
            public int getColor(final RenderInstruction renderInstruction, final int originColor) {
                final int a = (originColor >> 24) & 0xFF;
                final int r = (originColor >> 16) & 0xFF;
                final int g = (originColor >> 8) & 0xFF;
                final int b = originColor & 0xFF;

                // Calculate luminance using standard NTSC/BT.601 weights
                final int luminance = (int) (0.299 * r + 0.587 * g + 0.114 * b);

                // Recompose color with original alpha channel
                return (a << 24) | (luminance << 16) | (luminance << 8) | luminance;
            }
        });
```

@magma1447 you can easily use Google Maps by creating your own API Key. See https://github.com/cgeo/cgeo#api-keys

magma1447:
@eddiemuc Thanks, that works - pushed. Mapsforge offline maps are greyed the same way the VTM ones
are now, so the entry is offered for every VTM source plus the offline Mapsforge ones.

I did check whether the same callback could reach raster tiles too, since `ThemeCallback`
has a `getBitmap` on it, but that one is only called from `XmlUtils` when the theme loads
its own symbols - downloaded tiles never pass through it. So Mapsforge online sources are
still the one case this cannot cover, for the reason above rather than a missing hook I had
overlooked.

Nothing had to be done about stale tiles: `reapplyMapTheme` already purges the cache on
every theme (re-)apply, and toggling goes through `changeMapSource`.

One difference worth knowing, in case you would rather have it the other way: on Mapsforge
the theme's own symbols keep their colours, because I only replaced the colours and not
`getBitmap`. VTM has no equivalent hook, so leaving it alone is what makes the two
renderers look the same. Overriding it on Mapsforge is a couple of lines if you prefer the
symbols greyed as well.

And thanks for the pointer about the API key - I have one now. That also let me add tile
overlay support for Google Maps over in the other pull request. For grayscale, Google would
be a `saturation: -100` styler in `GoogleMapsThemeHelper` rather than anything to do with
themes or tiles; happy to add it here, or leave it out if you would rather keep this one to
the two Mapsforge renderers.

magma1447:
@SammysHP Added it, so Google Maps is greyed as well now.

It turned out to need even less than I described: `setCurrentThemeOnMap` already assembles
the style array, so this is one more styler appended at the end - `saturation: -100` with no
featureType, after the theme and its feature visibilities, so nothing can override it.
Markers, routes and tile overlays are not part of the map style and keep their colours.

The plain map only, though: satellite and terrain ignore a map style, the same reason the
theme options are hidden for them, so the entry is not offered there either.

That leaves the option covering every VTM source, the offline Mapsforge ones, and Google
Maps. The two gaps are Mapsforge raster and the Google imagery types, both because the
renderer gives nothing to hook into rather than because of anything left undone here.

Tested on a Pixel with a key of my own.

eddiemuc:

Ok to test

eddiemuc (approved):
lgtm

eddiemuc:
Code lgtm, necessary checks are green (default + Codacy). 

@magma1447 Many thanks for your contribution! Ready to merge from your perspective?

magma1447:
@eddiemuc  Yes, ready to merge from my side. Thanks for the quick review.

PR #18596, opened 2026-09-13
Abstract progress handler for batch-processing

ztNFny:
- create an abstract implementation for batch-processing of caches
-- should be used as basis for all mass-online-actions to provide feedback to the user
-- parameters: title, description, counter with progress bar, cancel button
- migrate the existing "download cache" service to it
- convert coord upload and personal note upload from blocking dialog
- convert watch/unwatch from silent background task

image


eddiemuc:
- possible to have fewer notifications? Eg only one by task type? Means: when a "add to watchlist" batch starts and a notification fot "add to watchlist" is already present (either still running or finished), can the new task be "integrated" into this notification? 
- would there be also notifications for single cache updates? If yes, could we please have a preference to deactivate this?
- what happens when user has deactivated notifications?

ztNFny:
> possible to have fewer notifications? Eg only one by task type? 

Sure - i didn't change anything about the behaviour here but for sure we can enqueue the 2nd+ job

> would there be also notifications for single cache updates? 

It's only implemented in CacheList for now, for single cache imho we should use (as is now) the progress bar

> what happens when user has deactivated notifications?

Dunno ... Same thing as before 😉 Download has been like that forever

ztNFny:
> possible to have fewer notifications? Eg only one by task type?

Implemented.
E.g. while a refresh operation for 20 caches is running you start a second one for 10 caches it'll be "downloading X/30" instead of 2 ops with "X/20" and "X/10"

PR #18598, opened 2026-09-13
show "add filter criteria" button also in basic mode

ztNFny:
Shows the "Add filter criteria" button also in Basic filtering mode.
Makes it easier to access advanced mode (removes 1 tap) and makes it easier to discover.
Adding a filter criteria then automatically switched to Advanced mode.

image


eddiemuc:
The basic mode was once introduced for users not interested and maybe even confused by our advanced filter options. I don't think we should make the basic view more complex (imho even "filter saving" options don't belong in basic view).

How about instead introducing a preference to start filter view in advanced mode instead of basic?

ztNFny:
> How about instead introducing a preference to start filter view in advanced mode instead of basic?

Or just remember the last state.

Reason i didn't suggest that it's then always starting with an empty view - but the filters in basic provide a good starting point for more advanced filters - so it'll also be additional taps.

I don't have strong feelings about this one - did add it to my builds and wanted to check if the idea makes sense for the main one.

PR #18600, opened 2026-09-13
Hide empty attrib section

murggel:
## Description

While working on #18573, I noticed that the “attrib box" appears empty even when the cache has no attributes

## Additional context

grafik

grafik


main/src/main/java/cgeo/geocaching/utils/CacheInfoBoxes.java
```diff
@@ -157,12 +158,15 @@ public static void updateAttributes(final Geocache cache, final TextView attribu
             }
         }
 
+        TextView attributesText = attributesView.findViewById(R.id.attributes_text);
```
murggel:
```suggestion
        final TextView attributesText = attributesView.findViewById(R.id.attributes_text);
```

murggel (commented):
Fix codacy

PR #18609, opened 2026-09-16
Map marker performance improvements

ztNFny:
After my recent experience with slow map drawing I did an extensive (AI-supported) review session of MapMarkerUtil.
The result is a number of changes that -in theory- should improve performance:
- Cache icons on Geocaches in memory
- Use a custom HashCodeBuilder
- Cache marker results
- Cap cache size
- Introduce bitmap cache for Mapsforge

Detailed descriptions of each change in the commits - I recommend to review them individually rather than the whole PR - I kept the commit messages quite detailed there.
I already went through several cycles of manual and tool-assisted review, but further reviews would be appreciated.

However in a real-device test I couldn't notice any difference in performance 😱 A test class (included) reduces runtime from 3:34 to 2:57 - not that significant.

Question is if we want to keep any of these changes? Imho many of them still make sense - why they don't have the expected impact is unclear to me ... or it means that we actually have bottlenecks in other places and MapMarkerUtil doesn't get around to making that much of an impact?


ztNFny:
Here's an analysis why the performance impact is so small:

### The dominant cost (createCacheMarker) is identical before and after                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                           
  Bitmap allocation + layered-drawable construction + Canvas draw is the expensive operation (~1–10 ms per unique marker on device). The 630 unique markers are created on cold exactly once — the optimization doesn't change when a cache miss occurs. Cold ≈ dirty ≈ several seconds; warm ≈ tens of ms 
  — but that gap comes from the overlaysCache that already existed before any of these commits.                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                           
### The benchmark measures the wrong layer                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                           
  The test calls getCacheMarker directly and unconditionally for all 5,000 caches on every pass. In the real app, GeoItemsLayer (commit 77ff238b65) skips getCacheMarker entirely for stable caches — that fast path is where most of the cycles are actually saved. The benchmark hardwires the slow path 
  every single pass, so it never captures the main optimization.                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                           
### The getCacheMarker-internal dirty flag is secondary                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                           
  Within getCacheMarker, skipping the 22-field hash saves ~5 µs per cache. Across 5,000 caches that's ~25 ms. Against the cold-pass cost (bitmap rendering for ~630 unique marker shapes, which dominates at 1–10 ms each), 25 ms is small signal-to-noise. Warm vs. dirty differences are measurable but  
  easily buried in JIT and GC noise on a single run.                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                           
### On a real device with 5,000 caches, Mapsforge rendering is the bottleneck                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                           
  Even after every optimization fires — fast path skips getCacheMarker for stable caches, backendBitmap avoids repeat convertToBitmap calls — the Mapsforge pipeline still has to blit 5,000 bitmap markers to the canvas every time the viewport changes. That's an O(n) draw cost that none of these     
  commits touch. If it runs in 100–200 ms, shaving 25 ms off marker computation is invisible.

ztNFny:
I'd propose to keep these 3 changes, regardless if they produce significant results as they should at least improve performance in the long term (memory optimizations). I'll also look into further improvements on createCacheMarker() 

### Replace HashCodeBuilder with inline hash                                                                                                                                                                                                                                                    
  The previous HashCodeBuilder constructed a new object on every getCacheMarker call. At 5,000 caches × 4 observer ticks/sec this produced ~20,000 short-lived heap allocations per second. The inline polynomial hash is stack-allocated and generates zero garbage. This is a genuine GC pressure  reduction that manifests as reduced jank during sustained live-map panning, even if it doesn't shorten any individual call.                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                           
### Cap overlaysCache at 2,000 entries with LRU                                                                                                                                                                                                                                                 
  Before this, the overlaysCache was unbounded. With 5,000 caches of varying visual state it could grow to several thousand entries. Memory pressure from an oversized cache causes GC pauses and competes with Mapsforge's tile buffers. The cap bounds the marker cache to ~28 MB and prevents that      
  growth. This is a correctness fix for memory, with an indirect performance benefit.                                                                                                                                                                                                                      

### GeoItemsLayer O(1) fast path for stable caches 
This is the most structurally significant change. Before it, getCacheMarker and layer.put() were called for all N caches on every viewModel.caches observer notification, regardless of whether anything had changed. The fast path skips both calls for any cache whose visual state, download-pending  
  flag, and coordinates are all unchanged since the last tick. On a stable viewport this reduces per-tick work from O(N × getCacheMarker_cost) to O(N × three-field comparison)

eddiemuc:
I agree with keeping the thrree supposed changes, but it seems the PR contains more than those?

main/src/main/java/cgeo/geocaching/utils/HashCode.java
```diff
@@ -0,0 +1,70 @@
...
+ * threads or between start() and toHashCode() calls on the same thread.
+ *
+ * WARNING — reentrancy: append(Object) calls v.hashCode() on the appended
+ * value. If that hashCode() itself calls HashCode.start(), the ThreadLocal
+ * builder is silently reset mid-computation, corrupting the outer hash.
+ * Only pass objects whose hashCode() is known NOT to use HashCode (e.g.
+ * String, enums, standard JDK collections of strings/primitives). For all
+ * other types, extract the hash first: append(myObj.hashCode()).
+ */
+public final class HashCode {
```
eddiemuc, line 24: Wouldn't it be easier to just use Objects.hash() instead of creating this new class?

ztNFny:
> I agree with keeping the thrree supposed changes, but it seems the PR contains more than those?

I implemented a number of changes, each of which should (in theory) improve performance - that's the 5 commits.
As performance improvement isn't as expected the question is whether we nevertheless want to keep any/all of these 5 - basically the (slight) performance vs. code complexity evaluation.

main/src/main/java/cgeo/geocaching/utils/HashCode.java
ztNFny: Objects.hash() requires 1 object to hash - we don't have that object?!

eddiemuc: I don't mean Objects.hashCode().

Objects.hash takes multiple objects:

ztNFny: Ah, sorry ... Objects.hash() is actually worse than the originally used HashCodeBuilder as it created a temp object (heap allocation, same as HashCodeBuilder) + boxing of each value.

The manual approach used in HashCode is actually Java best-practice for high-performance hash calculation (Effective Java, Joshua Bloch). Not something I invented. I trust the gurus on that one.

ztNFny: HashCode is basically HashCodeBuilder without some guardrails that are added in HashCodeBuilder, especially in regards to recursive hash building - which is something that's not done/needed here.
That's why I added an explicit warning to HashCode - it shouldn't be used a a general replacement of HashCodeBuilder but only controlled situations where performance is important and you know that you definitly won't run into this problem.

 * WARNING — reentrancy: append(Object) calls v.hashCode() on the appended
 * value. If that hashCode() itself calls HashCode.start(), the ThreadLocal
 * builder is silently reset mid-computation, corrupting the outer hash.
 * Only pass objects whose hashCode() is known NOT to use HashCode (e.g.
 * String, enums, standard JDK collections of strings/primitives). For all
 * other types, extract the hash first: append(myObj.hashCode()).

ztNFny:
I've set up some benchmark scenarios now (should have started with that ...) and no matter what I do: Performance doesn't actually change in any measurable way. So all of the fixes above do improve performance slightly - but none of them are on the critical path.
If we want to keep any of them it'd be for "nice code"/"better pattern" only.

The real bottlenecks are in the map renderers (especially VTM which is known to be bad and [No map](https://github.com/cgeo/cgeo/issues/18612) which is just catastrophic)

ztNFny:
One of the changes in this PR is causing problems with updating markers. As the whole bunch brings no performance improvement I'll just close the whole thing without further investigation.
If one of the changes should still be included it can easily be cherry-picked.

Issue #18623, opened 2026-09-20
"Show as list" behavior changed


moving-bits:
Reported by user on support (ticket [501411](https://cgeosupport.droescher.eu/scp/tickets.php?id=10887)) for current beta 2026.09.19-RC and reproducible for me:

"I am in a list of caches.
I 'click' on 'Show on map' icon in the top bar.
The map opens and displays the caches from the current list.
Fine.

I zoom on the map, to allow a narrower geographical selection.
I then open the menu (3 dots) and select 'Show as list'.

The list shows all caches from the original list, NOT the caches displayed on the zoomed section of the map.
I would like the 'Show as list' feature to show only the caches selected on the zoomed map.

This issue is a regression.
In earlier Releases it used to work as I describe."

eddiemuc:
That might have been me. But I honestly thought up to now that "show as list" should open the list, not the caches currently shown...

moving-bits:
I did some additional testing:

My personal use-case for "Show as list" mostly is
- enable live map
- zoom to the area relevant to me
- "show as list" and store only the caches in the relevant area

This scenario still DOES work in current beta release.

The scenario described by the user still may be useful, though, if you 
- receive a GPX file for a larger area, 
- import it to a list, 
- map that list, 
- and want to update only caches of an area relevant for you.

SammysHP:
I often use that feature to sort stored caches into other lists depending on the area.

ztNFny:
Confirmed - I just ran into that today. "Show as list" always showed the current viewport as list. For switching to the stored list we have nav bar, so I'd vote for reverting to the previous behaviour.
Usecase for "show as list" for the current viewport is that a lot of cache-management functionality is not available from map menus but only from list menus.

eddiemuc:
By all means, lets revert this. Its a small change.

Is there any way switching between map and list (both directions) where the shown list is taken over?

ztNFny:
> Is there any way switching between map and list (both directions) where the shown list is taken over?

It was discussed back when we introduced bottom-nav whether switching from lists to map should map the list or show live map. Decision was to show live map and have the separate "map" button in lists toolbar.
That switching the selected list in map view and then switching to lists doesn't keep the selected list irked me quite a few times already, but it's somewhat consistent with the above decision - I'd welcome such a change though

PR #18634, opened 2026-09-21
revert "Show as list" behaviour (fixes #18623)

ztNFny:
"Show as list" should show only the caches in current viewport, fixes https://github.com/cgeo/cgeo/issues/18623


eddiemuc (approved):
Thanks a lot!

Issue #18622, opened 2026-09-20
List pickers: Scroll to active list + Inconsistent behavior CacheList vs. Map


moving-bits:
Reported by user on support (ticket [501411](https://cgeosupport.droescher.eu/scp/tickets.php?id=10887)) for current beta 2026.09.19-RC:

User has a long list of cache lists. When changing lists, our new picker always starts at the top, thus you may have to scroll a lot, when you want to select a list adjacent to the currently active one, if that one is farther down the list.

Our old list picker used radio buttons. The currently active entry  was always visible on opening the dialog (and clearly marked), whereas the new dialog always opens at the top.

The issue is reproducible for me.

Would it be possible to ensure that the currently active list is always visible on opening the list picker + maybe highlight it visually in some way?

Additional finding:
- Scenario described above applies to `CacheListActivity` and to first opening in `UnifiedMapActivity`.
- After having selected a new list on the map, the picker there switches to a variant still using radio buttons. -> Interesting feature, but not very consistent 😄.

Maybe we should simply use the one with radio buttons in both locations, and always?

---

Original report by the user:

"I am in a list of caches, and I want to navigate to another list.
I 'click' in the top bar, to open the 'Pick a list' object.

I would like the 'Pick a list' object to open positioned at the level of the current list.
And therefore also to display the full name of the current List.
It currently opens at the top.
To navigate to a List adjacent to the one I am in, I have to scroll down the full list.
This is very cumbersome when performing Lists maintenance, rendering it almost unusable.

This issue is a regression. In earlier Releases it used to work as I describe."

ztNFny:
I'd prefer the non-radio version as it's visually cleaner. However if the radio is the only simple way to fix list position I'm fine with that.

ztNFny:
> * After having selected a new list on the map, the picker there switches to a variant still using radio buttons. -> Interesting feature, but not very consistent 😄.

caused by 955b6fc8

@eddiemuc I don't understand why you made this particular change, could you please check? Looking at the code it seems to be intended to pass the filter if set, but that also works fine without that change.
https://github.com/cgeo/cgeo/commit/955b6fc8e22ea5091e0eb7a43aae8c63af9f51f3#diff-7fdc4e9bfb9829a6455e9cb29dd7b5ae48dff06209e6baed41a919d2ece83ad8R108

I'd propose to revert that line to `}, false, PseudoList.NEW_LIST.id));`

eddiemuc:
[...]

PR #18635, opened 2026-09-21
scroll to selected list on map/lists view

ztNFny:
This causes the list to scroll to the active item but without the radio buttons being visible.

No visual distinction for the active item is implemented for now, but could of course be done.

Goal is to unify the list selection GUI in different places as raised in https://github.com/cgeo/cgeo/issues/18622. We have 3 places where a list is selected for display: 1. in Map/List activity in the toolbar, 2. long-tap on Stored in bottom nav, 3. long-tap on Map in bottom nav.
Only the first is guaranteed to have an "active list" information as the others can be called from anywhere. So displaying radio buttons with an "active list" doesn't make sense semantically.


main/src/main/java/cgeo/geocaching/ui/CacheListActionBarChooser.java
```diff
@@ -100,12 +100,12 @@ private void refreshActionBarTitle() {
             actionBar.setDisplayShowTitleEnabled(false);
             actionBar.setDisplayShowCustomEnabled(true);
             actionBar.getCustomView().setOnClickListener(v -> new StoredList.UserInterface(context).promptForListSelection(R.string.list_title, selectedListId -> {
-                if (selectedListId != listId || namedFilterId >= 0) {
+                if (selectedListId != listId) {
```
eddiemuc: Please leave namedfilterid check, it is needed when named filter is clicked in cachedetails (it is 0 in all other cases anyway).

main/src/main/java/cgeo/geocaching/ui/CacheListActionBarChooser.java
```diff
@@ -100,12 +100,12 @@ private void refreshActionBarTitle() {
...
                     this.onSelectAction.call(selectedListId);
                     this.listId = selectedListId;
                     refreshActionBarTitle();
                 }
-            }, false, Collections.singleton(PseudoList.NEW_LIST.id), namedFilterId >= 0 ? -1 : listId, null));
+            }, false, Collections.singleton(PseudoList.NEW_LIST.id), listId, null));
```
eddiemuc: Same as above

main/src/main/java/cgeo/geocaching/ui/SimpleItemListModel.java
```diff
@@ -68,7 +68,8 @@ public class SimpleItemListModel<T> {
 
 
     /** Supported display modes for choosing items from a list */
-    public enum ChoiceMode { SINGLE_PLAIN, SINGLE_RADIO, MULTI_CHECKBOX }
+    /** Like SINGLE_PLAIN but allows pre-marking an active item (via setSelectedItems) for scroll anchoring without any visual indicator */
+    public enum ChoiceMode { SINGLE_PLAIN, SINGLE_RADIO, SINGLE_ACTIVE, MULTI_CHECKBOX }
```
eddiemuc: I am surprised that it is enough to add and use the new enum without doing any logic with it. Does this really do the trick?

ztNFny: yes, because the "set item as active" is in the base implementation and that's the key. But that lead me to the realization ... even that change isn't needed - all that's needed is setting the ChoiceMode to plain!
At least if we don't want another way of highlighting the active item (e.g. by coloring it)

main/src/main/java/cgeo/geocaching/ui/CacheListActionBarChooser.java
ztNFny: I don't get what you want that code to do ...
It's doing "if a named filter is selected" pass "no list selected" to the list selection menu - so no list is selected and thus scrolling to the list is broken.
So that's actually an important part of the fix.

eddiemuc: The namedfilterid doesn't refer to a named filter selected as filter, but a named filter selected as primary list to display (not affecting the offline filter). It should not be relevant for the bug. Does the scrolling bug still occur when you leave the old code here?

ztNFny: Scrolling is broken with namedFilterId in place. With no filter selected namedFilterId is 0 - did you mean to make the check "> 0" instead of ">= 0"?

> but a named filter selected as primary list to display (not affecting the offline filter)

How can you list/map a filter??

eddiemuc: > Scrolling is broken with namedFilterId in place. With no filter selected namedFilterId is 0 - did you mean to make the check "> 0" instead of ">= 0"?

You're right, that's the bug. Should be > 0.
 
> > but a named filter selected as primary list to display (not affecting the offline filter)
> 
> How can you list/map a filter??

Its just a dynamic list in this context. Try it.

ztNFny: > Its just a dynamic list in this context. Try it.

How?! I have no idea on how you intend that to be used/accessed.

eddiemuc: Click any filter listed in cache details

ztNFny: I had already wondered why the markers in CacheDetails now display as links - but had not yet clicked them.
So that's basically the "All" pseudo-list with a filter applied. Wouldn't it actually be easier to implement (and display) it like that instead of generating a new filter-pseudo-list?

(For me and my usage of c:geo that feature is not useful, just as I don't see any use of the "All" list. I have all my found caches stored, those in my homezone but also plenty of lists with interesting caches country- and world-wide. So using All with or without filter is just a random bunch of stuff).

eddiemuc: I did it like that first, but then I realized that it deletes/overwrites the offline filter this way. Didn't want that

ztNFny:
restest this please

moving-bits:
retest this please

ztNFny:
failing test is unrelated (the waypoint-parser issue)

moving-bits:
yeah, CI seems to be more stable currently, last test results were both like 20+/- failing tests, but now it's back to just waypoint parser

moving-bits:
This is a nice improvement, thanks for implementing.

> No visual distinction for the active item is implemented for now, but could of course be done.

Having a visual indicator would be nice, indeed. Two ideas for it:
- A small triangle to the left of the list icon, point towards the list entry
- or a light gray background block, full-width

ztNFny:
If we want an indicator i'd just go for our highlight color for the text instead of introducing a new convention of highlighting items (e.g. dark background)

ztNFny:
But imho no highlight is needed - the selected list is visible in the app screen behind the dialog - which is why i didn't implement one. And also apparently noone missed it since we removed the radio buttons.

eddiemuc:
If a highlight is implemented (and I don't say it should) then I suggest to replace the current "radio button" mode of SimpleDialog completely with this new highlight option. It would then be used everywhere where we use radio buttons in select lists today.

moving-bits:
Highlighting is not needed if you have just a handful of lists, but as soon as there are ten or more entries, a visual indicator would be very useful, IMHO. And even more, if you have so many entries (lists, filters, ...) that the list starts scrolling.

Using highlight color would be fine, IMHO:

|without highlighting|with highlighting|
|---|---|
|image|image|

moving-bits:
I propose to add visual indicator via highlighting first, to have it in the new release, and if we want to remove radio button mode, then do that in a step after the release (to not block the release).

Screenshots above are just visual mockups. I can try to update SimpleDialog if you give me a pointer where to do it @eddiemuc

eddiemuc:
> I propose to add visual indicator via highlighting first, to have it in the new release, and if we want to remove radio button mode, then do that in a step after the release (to not block the release).
> 
> Screenshots above are just visual mockups. I can try to update SimpleDialog if you give me a pointer where to do it @eddiemuc

A line item in simpledialog is represented by `simpleitemlist_item_view.xm`l. You will find there the different components one can have in such a line (a checkbox, a radiobutton etc) which are then made visible or gone depending on what's needed by the logic in `SimpleItemListView.ItemListViewHolder.fillData()`

Issue #18637, opened 2026-09-21
Description not rendering


campdave:
### Describe your problem!

When I click on a cache on the map, the cache loads the hint but not the description.

Same behaviour on both my phone and tablet.

<img width="1220" height="2712" alt="Image" src="https://github.com/user-attachments/assets/7a95c958-9209-4e27-89c6-0641ae15c54a" />

### How to reproduce?

Open map
Click on cache
Swipe to description, nothing there 

### Reproducible

Yes

### c:geo Version

2026.08.05

### System information

```text
## System information

c:geo version: 2026.08.05

Datetime: 21 September at 18:22

Device:
-------
- Device type: motorola edge 60 (scout_g_sysesnp, motorola)
- Available processors: 8
- Android version: 16
- Android build: W1VCS36H.14-20-19-7
- Screen resolution: 1220x2449px (434x871dp)
- Pixel density: 2.8125
- System font scale: 1.0 / used scale: 1.0
- Sailfish OS detected: false
- Google Play services: enabled - 26.34.36 (260400-981326859)
- Memory:  Available:3.4 GB, Total:11.2 GB, Threshold: 216.0 MB, low:false

Sensor and location:
-------
- Low power mode: inactive
- Compass capabilities: yes
- Rotation vector sensor: present
- Orientation sensor: present
- Magnetometer & Accelerometer sensor: present
- Direction sensor used: rotation vector

Program settings:
-------
- Settings: v11, Count:126
- Set language: en_GB (system default)
- System date format: dd/MM/y
- Time zone: GMT+01:00
- Debug mode active: no
- Log Settings: debug=false, minLevel=WARN, minAddCallerInfo=NONE, addCallerInfoMaxDepth=8, throwOnError=false, transactionSizes=false
- Last manual backup: never
- Last auto backup: 6 Sept, 16:17
- Routing mode: Walk (shortest.brf / trekking.brf / car-eco.brf / (none) / (none))
- Map mode: UnifiedMap / live / OSM multi-threading: off
- Map: OpenStreetMap.org
  - Id: cgeo.geocaching.unifiedmap.tileproviders.OsmOrgSource:null
  - Attrs: © OpenStreetMap contributors
  - Theme: none

Filters:
-------
- Hide waypoints: -
[...]

PR #18638, opened 2026-09-21
fix GC description rendering (fixes #18637)

ztNFny:
adjust to website change


ztNFny:
I'm looking into the additional failures to due website changes (waypoints), updating coming up.

bekuno:
probably the waipoint test is now broken by the error or the fix;
~~~
23:51:32  cgeo.geocaching.connector.gc.WaypointsTest > testDownloadWaypoints[android-30(AVD) - 11] [31mFAILED [0m
23:51:32  	java.lang.AssertionError:
23:51:32  	Expected size: 15 but was: 8 in:
23:51:33  
~~~

ztNFny:
> probably the waipoint test is now broken by the error or the fix;
> ~~~
> 23:51:32  cgeo.geocaching.connector.gc.WaypointsTest > testDownloadWaypoints[android-30(AVD) - 11] [31mFAILED [0m
> 23:51:32  	java.lang.AssertionError:
> 23:51:32  	Expected size: 15 but was: 8 in:
> 23:51:33  
> ~~~

Parsing of GC3KE70 is indeed still buggy, only 8 of 15 WPs do show up 😨. However other multis are fine with all WPs available. Will check again...
The whole WP parser is really not nice, but i didn't want to do a bigger rewrite for an emergency fix, but maybe we should...

ztNFny:
See https://github.com/cgeo/cgeo/pull/18639 for a follow-up
From all I can tell that cache should have been broken before already, but without the old html it's impossible to tell what exactly changed here.

